PR #16637, opened 2021-03-12
force check if newref needed if project create fail

frederic34:


eldy:
If it fails for another reason than a duplicate ref (for example a trigger, or a mandatory extrafields). Does is will generate also another ref ?

frederic34:
It just in case of two person trying to create project with same number, the new ref will not change if there is no prject created

PR #16639, opened 2021-03-13
Update list.php wrong number presentation in stock

fbosman:
Changed the presentation of the stock to show decimals and thousands separator. changes price2num to price and use MAIN_MAX_DECIMALS_STOCK

# Instructions
*Please:*
- *only keep the "Fix", "Close" or "New" section*
- *replace the bracket enclosed textswith meaningful informations*

# Fix #[*issue_number Short description*]
[*Long description*]

# Close #[*issue_number Short description*]
[*Long description*]

# New [*Short description*]
[*Long description*]

### DIFF
--- a/htdocs/product/list.php
+++ b/htdocs/product/list.php
@@ -1644,7 +1644,7 @@ if ($resql) {
 					print img_warning($langs->trans("StockLowerThanLimit", $obj->seuil_stock_alerte)).' ';
 				}
 				$decimals = !isset($conf->global->MAIN_MAX_DECIMALS_STOCK) ? 5 : $conf->global->MAIN_MAX_DECIMALS_STOCK;
-				print price($product_static->stock_reel, 0, $langs->defaultlang,1,-1,$decimals);
+				print price($product_static->stock_reel, 0, $langs->defaultlang, 1, -1, $decimals);
 				//print price2num($product_static->stock_reel, 'MS');
 			}
 			print '</td>';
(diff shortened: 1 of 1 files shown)


htdocs/product/list.php
```diff
@@ -1643,7 +1643,9 @@
 				if ($obj->seuil_stock_alerte != '' && $product_static->stock_reel < (float) $obj->seuil_stock_alerte) {
 					print img_warning($langs->trans("StockLowerThanLimit", $obj->seuil_stock_alerte)).' ';
 				}
-				print price2num($product_static->stock_reel, 'MS');
+				$decimals = !isset($conf->global->MAIN_MAX_DECIMALS_STOCK) ? 5 : $conf->global->MAIN_MAX_DECIMALS_STOCK;
```
eldy, line 1646: Using the "MS" of price2num is already the "shared code" to round an amount of stock.
Why not just adding price to format amount into locale format.

print price(price2num($product_static->stock_reel, 'MS'));

fbosman:
As far as I can see this code in price2num disables all formatting:
//print "TT".$amount.'<br>';

		// Always make replace because each math function (like round) replace
		// with local values and we want a number that has a SQL string format x.y
		if ($thousand != ',' && $thousand != '.') {
			$amount = str_replace(',', '.', $amount); // To accept 2 notations for french users
		}
		$amount = str_replace(' ', '', $amount); // To avoid spaces
		$amount = str_replace($thousand, '', $amount); // Replace of thousand before replace of dec to avoid pb if thousand is .
		$amount = str_replace($dec, '.', $amount);

eldy:
Yes, goal of price2num() is to clean a value in memory (no formatting, but rounding)
And goal of price() is to format for output on screen into the correct user language/localization.
That's why price(price2num(   , 'MS')) should be use to both round and format.

PR #16648, opened 2021-03-13
fix php8 fatal error

frederic34:


htdocs/main.inc.php
```diff
@@ -48,6 +48,18 @@
 	}
 }
 
+if (!function_exists('is_countable')) {
```
eldy, line 51: main is loaded only in web context when memory.lib can be used in cli context too.
The is_countable function should be moved into master, shouldn't it ?

frederic34: I moved function in master.inc.php

PR #16673, opened 2021-03-14
fix php8 warnings

frederic34:


htdocs/fichinter/card-rec.php
```diff
@@ -102,13 +104,13 @@
 
 // Create predefined intervention
 if ($action == 'add') {
-	if (!GETPOST('titre')) {
+	if (!GETPOSTISSET('titre')) {
```
eldy, line 107: GETPOSTISSET seems not enough GETPOST must also return a value.
What's wrong with GETPOST('titre') ?
If post is not defined, GETPOST should return an empty value without warning too.

PR #16695, opened 2021-03-15
Update hook_document.php

dolibit-ut:
add   $help_url = 'EN:Hooks_system|FR:Système_de_Hooks';

delete  //$help_url='EN:Module_Third_Parties|FR:Module_Tiers|ES:Empresas';


eldy:
This file is not related to hooks but to zapier module (the name use is not a good choice).
Doc should be linked to this url https://wiki.dolibarr.org/index.php?title=Module_Zapier

PR #16705, opened 2021-03-15
Update card.php

dolibit-ut:
$helpurl = "EN:Module_Banks_and_Cash";


dolibit-ut:
sometimes  $helpurl  is used
and sometimes it is  $help_url

what do we want to use by default?

eldy:
When several way of development is found, we must choose the one found into template /modulebuilder (the reference), so $help_url.

PR #16759, opened 2021-03-17
Fix set const at init

atm-quentin:
# Fix set const at init

### DIFF
--- a/htdocs/core/class/conf.class.php
+++ b/htdocs/core/class/conf.class.php
@@ -481,22 +481,7 @@ class Conf
 			unset($this->global->PROJECT_USE_SEARCH_TO_SELECT);
 		}
 
-		if (!empty($this->productbatch->enabled))
-		{
-			$this->global->STOCK_CALCULATE_ON_BILL = 0;
-			$this->global->STOCK_CALCULATE_ON_VALIDATE_ORDER = 0;
-			$this->global->STOCK_CALCULATE_ON_SHIPMENT = 1;
-			$this->global->STOCK_CALCULATE_ON_SHIPMENT_CLOSE = 0;
-			$this->global->STOCK_CALCULATE_ON_SUPPLIER_BILL = 0;
-			$this->global->STOCK_CALCULATE_ON_SUPPLIER_VALIDATE_ORDER = 0;
-			if (empty($this->reception->enabled)) {
-				$this->global->STOCK_CALCULATE_ON_SUPPLIER_DISPATCH_ORDER = 1;
-			}
-			else {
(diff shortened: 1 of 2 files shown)


htdocs/core/modules/modProductBatch.class.php
```diff
@@ -120,6 +120,20 @@ public function init($options = '')
...
 		$sql = array();
 
+		dolibarr_set_const($db, 'STOCK_CALCULATE_ON_BILL', 0);
+		dolibarr_set_const($db, 'STOCK_CALCULATE_ON_VALIDATE_ORDER', 0);
+		dolibarr_set_const($db, 'STOCK_CALCULATE_ON_SHIPMENT', 1);
+		dolibarr_set_const($db, 'STOCK_CALCULATE_ON_SHIPMENT_CLOSE', 0);
+		dolibarr_set_const($db, 'STOCK_CALCULATE_ON_SUPPLIER_BILL', 0);
+		dolibarr_set_const($db, 'STOCK_CALCULATE_ON_SUPPLIER_VALIDATE_ORDER', 0);
+
+        if (empty($this->reception->enabled)) {
```
hregis, line 130: @atm-quentin  this is "$conf->reception->enabled" here !
but this check will not be executed if the "reception" module is activated or deactivated after the "productbatch" module !!

atm-quentin: right but same thing with cashdeks after

htdocs/core/class/conf.class.php
```diff
@@ -481,22 +481,7 @@ public function setValues($db)
 			unset($this->global->PROJECT_USE_SEARCH_TO_SELECT);
 		}
 
-		if (!empty($this->productbatch->enabled))
```
eldy, line 484: Currently, goal of this code is because, batch number management is not compatible with some rules of stock autoincrease/decrease. So we force it the rules when module is enabled whatever is setup.
if you move this as init, forcing will not work. We need to be sure to have this setup when module ProductBatch is on (any other options does not work).

atm-quentin: I can't unset this const if i want to manage stock movement otherwise. It would be better to not allow checking other conf if it's not handle, but allow to disable const

eldy: Which option for example ?
This is on purpose that we are not able to unset the conf that are forced when module productbatch is enabled. It is because the application itself is not compatible and bugged if we try to use other value than the one forced when productbatch is enabled.
Solution to allow to set this option is to make application working correctly with batch number and the option set. Currently it is not.

PR #16770, opened 2021-03-17
Update llx_20_c_departements.sql

dolibit-ut:
Questions:

- at some countries we have fields 'cheflieu' and 'tncc' - on others not
  (what are the meaning of cheflieu, tncc, ncc .. what are the effects)

- at some countries is a state active set - on others not
  (is it useful to set or to delete at all)


eldy:
The field "cheflieu" contains the name of the main town of the state.
tncc is a code to say what must be the pronoun to use when we use the name of town in a sentence (it is used by French language only)

PR #16771, opened 2021-03-17
fix php8 warnings

frederic34:


dolibit-ut:
Hi, why we go from 'project' back to 'projet' ?

I think we should change all 'projet' terms to 'project'.

frederic34:
@dolibit-ut thank's for comment

PR #16776, opened 2021-03-17
fix php8 warnings

frederic34:


htdocs/core/boxes/box_graph_nb_tickets_type.php
```diff
@@ -128,6 +128,11 @@ public function loadBox($max = 5)
 			}
 			$dataseries = array();
 			$data = array();
+			$data['COM'] = 0;
+			$data['HELP'] = 0;
+			$data['ISSUE'] = 0;
+			$data['REQUEST'] = 0;
+			$data['OTHER'] = 0;
```
eldy, line 135: Do we really have to init array with hard coded value.
Just after we make a select to fill dynamically $data to avoid to know hard coded values.

frederic34: image
some value may not initialised

eldy: I think we should instead fix the line 145 by testing if key exists in array and if not setting 0

PR #16791, opened 2021-03-18
FIX: fix checkbox displayed according to module project setup parameters - work in progress

Robin-Yann:
fix checkbox displayed according to module project setup parameters - work in progress


eldy:
I don't understand what is the bug you try to fix here. Can you describe how to reproduce the bug ?

Robin-Yann:
This is a display bug on the project page when disabling the module parameters:
* Utiliser les projets pour suivre les affaires / opportunités
* Utiliser les projets pour suivre les tâches et/ou saisir du temps consommé (feuilles de temps)

Depending on the page, the title of the “Usage” line remained displayed as well as the “Invoice time spent” checkbox remained displayed while the options are unchecked.
Editing is not necessary in the Task and Time pages, it seemed good to keep the same code everywhere.

PR #16620, opened 2021-03-11
Define Masks for Batch Products Equipments

altairis-tof:
First commit to add a setup option to define masks for batch products, one mask for batch products managed by lots and one mask for serial numbers.
Next step will be to use this setup when adding new batch products in the database.

@eldy some people insist to be able to define batch masks per products, arguing they use different numbering lot/sn management depending of the kind of products. i convinced them to first use standard dolibarr masks management and that we will see to code it later into core or into spécific module. 

What do you think of it @eldy ?

PR #16797, opened 2021-03-18
add serials from product card

altairis-tof:
add a button on product card to directly got to the stock correction to add equipments to the product


htdocs/product/card.php
```diff
@@ -2282,6 +2282,10 @@
 					print '<a class="butAction" href="'.$_SERVER["PHP_SELF"].'?action=clone&amp;id='.$object->id.'">'.$langs->trans("ToClone").'</a>';
 				}
 			}
+
+			if ( ! empty($conf->productbatch->enabled) && $object->hasBatch() ) {
```
eldy, line 2286: Because it is not an action that has effect on the workflow of the product, it is not consistent to have this link here.
I suggest to replace this with a link on line into the tab Stock, just after "Last movement", having a new line "Last lot / serial" with link to list and a button + to add one new lot batch (your new link)
image

eldy, line 2286: This feature is related to stock. And this link/button already exists on tab "Stock".

altatof: it was a demand from users to have a "shortcut" on the product tab to avoid multiple clicks.
And in the logic of the users, equipments are more linked to the product than to stock.
@lmag

fappels: @altatof maybe add to dolitools ;)

atm-maxime: Yes shortcuts is a good idea, but everybody wants his own personal shortcut. A company without stock mgmt but with multi-language will want a shortcut to add a product translation. Another who only deals with purchase will want a shortcut to add a supplier price...
So a global shortcut system giving access to actions directly from the tabs for example would be nice ?
And in your code you didn't check the user permission on stock.

altatof: hello @atm-maxime 
_So a global shortcut system giving access to actions directly from the tabs for example would be nice ?_
Very good idea ! How do you imagine that ?

eldy: I agree with @maxime.
There is a ton of user request to have shortcut depending of what they used the most. But we can't add all combinations of desired shortcuts in application everywhere. Position of feature must remains at the best logical place (and to do a stock movement, the best place is the "Stock" tab, not the general information tab.
Currently, feature to add your own shortcut is an external module that add link by using the hook.

atm-maxime: > How do you imagine that ?

Currently we have static tabs everywhere and action buttons at the bottom of each page. If we add on every tab a link (or on right click?) showing available actions this would be helpful I think.

lmag: I am not sure I understood your proposals exactly.
The idea of ​​being able to customize all the tabs seems practical to me. 
I will review with @altatof the demand for stocks because it seems to me that @eldy is right (stocks in stock) ! In case we will upload screens to be sure we are talking of the same thing.

altairis-tof:
Thanks everybody for the comments. I understand your arguments about shortcuts.
Anyway, i will PR a new tab with list of equipments for one product, with a Add button at the edge respecting the dolibarr UX chart.
This button will not lead to the stock correction but to the equipment creation card.
I guess it will do the job.
I close this PR.

altairis-tof:
@eldy please dont forget to merge #16620

PR #16801, opened 2021-03-18
fix php8 warning

frederic34:


frederic34:
when calling user->getNomUrl() need to have statut defined

htdocs/core/ajax/ajaxdirtree.php
```diff
@@ -288,6 +288,7 @@
 			print '<td class="center">';
 			$userstatic->id = $val['fk_user_c'];
 			$userstatic->lastname = $val['login_c'];
+			$userstatic->lastname = $val['statut_c'];
```
hregis, line 291: @frederic34    $userstatic->statut

frederic34: @hregis thank's

frederic34: comments are always welcome

PR #16818, opened 2021-03-18
Code translation

daraelmin:
Just a code translation to make it more readable

I wonder if we should transfer the "statistics" part into a function in the adherentstats.class.php file.

### DIFF
--- a/htdocs/adherents/index.php
+++ b/htdocs/adherents/index.php
@@ -193,11 +193,11 @@ if ($conf->use_javascript_ajax) {
 	$boxgraph .='<tr class="liste_titre"><th colspan="2">'.$langs->trans("Statistics").'</th></tr>';
 	$boxgraph .='<tr><td class="center" colspan="2">';
 
-	$SommeA = 0;
-	$SommeB = 0;
+	$SumToValidate = 0;
+	$SumValidated = 0;
 
-	$SommeC = 0;
-	$SommeD = 0;
+	$SumUpToDate = 0;
+	$SumResiliated = 0;
 	$total = 0;
 	$dataval = array();
 	$i = 0;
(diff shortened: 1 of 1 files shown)


frederic34:
@eldy I made box to replace fixed box, but the old box are back (screenshot with firefox)
image

frederic34:
@daraelmin I moved this code in box_member_xxxx

eldy:
> @eldy I made box to replace fixed box, but the old box are back (screenshot with firefox)
> image

@frederic34 I do not experience any trouble on my side with develop branch:
image

I think the new widgets where not migrated. You should add this into migration script to have them declared (replace box_graph_ticket_by_severity with name of new widget)
-- VMYSQL4.1 INSERT INTO llx_boxes_def (file, entity) SELECT  'box_graph_ticket_by_severity.php', 1 FROM DUAL WHERE NOT EXISTS (SELECT * FROM llx_boxes_def WHERE file = 'box_graph_ticket_by_severity.php' AND entity = 1);

If migrated, i think I will have 2 boxes for the new boxes so we can comment/remove hardcoded code for the 2 others. May be i restored them when making test and found them disappeared.

frederic34:
this is not enough ?  https://github.com/Dolibarr/dolibarr/blob/5ecef1bad35f25bc6210e15d76872990c6d3d639/htdocs/core/modules/modAdherent.class.php#L197-L200

daraelmin:
> this is not enough ?
> 
> https://github.com/Dolibarr/dolibarr/blob/5ecef1bad35f25bc6210e15d76872990c6d3d639/htdocs/core/modules/modAdherent.class.php#L197-L200

I'm mot sure that the name "box_member_subscriptions_by_year" is correct as there's no "DISTINCT" on member (one member can have more than one subscription).  It should be named : "box_subscriptions_by_year"

I problematize this question because thereafter, it would seem interesting to me to propose a class "subscriptionstats". To distinguish member statistics from subscription statistics

daraelmin:
> @daraelmin I moved this code in box_member_xxxx

@frederic34 thank you. I will see how you did in order to do it myself next time rather than "ordering"

PR #16845, opened 2021-03-19
FIX avoid php8 warning with multicompany

hregis:


htdocs/core/ajax/pingresult.php
```diff
@@ -67,14 +67,14 @@
 if ($action == 'firstpingok') {
 	// Note: pings are per installed instances / entity.
 	// Once this constants are set, no more ping will be tried (except if we add parameter &forceping=1 on URL). So we can say this are 'first' ping.
-	dolibarr_set_const($db, 'MAIN_FIRST_PING_OK_DATE', dol_print_date($now, 'dayhourlog', 'gmt'));
-	dolibarr_set_const($db, 'MAIN_FIRST_PING_OK_ID', $hash_unique_id);
+	dolibarr_set_const($db, 'MAIN_FIRST_PING_OK_DATE', dol_print_date($now, 'dayhourlog', 'gmt'), 'chaine', 0, '', $conf->entity);
```
eldy, line 70: The MAIN_FIRST_PING_OK_DATE and other are parameters global for all entities. entities must not be provided on parameters

hregis: @eldy ok for all entities i change my PR to force entity 0

PR #16846, opened 2021-03-19
NEW auto notification with templated emails

ptibogxiv:
This PR need some changes: generate pdf directly with some class

### DIFF
--- a/htdocs/core/class/notify.class.php
+++ b/htdocs/core/class/notify.class.php
@@ -3,6 +3,7 @@
  * Copyright (C) 2004-2011 Laurent Destailleur  <eldy@users.sourceforge.net>
  * Copyright (C) 2014	   Juanjo Menent		<jmenent@2byte.es>
  * Copyright (C) 2018 	   Philippe Grand		<philippe.grand@atoo-net.com>
+ * Copyright (C) 2021 	   Thibault FOUCART		<support@ptibogxiv.net>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


htdocs/core/class/notify.class.php
```diff
@@ -442,122 +443,149 @@ public function send($notifcode, $object, $filename_list = array(), $mimetype_li
 
 						switch ($notifcode) {
 							case 'BILL_VALIDATE':
-								$link = '<a href="'.$urlwithroot.'/compta/facture/card.php?facid='.$object->id.'">'.$newref.'</a>';
-								$dir_output = $conf->facture->dir_output;
+								$link = '<a href="'.$urlwithroot.'/compta/facture/card.php?facid='.$object->id.'&entity='.$object->entity.'">'.$newref.'</a>';
```
eldy, line 446: Why adding entity into url ? Entity should be found from the object id itself and providing on url is not reliable.

ptibogxiv: because with multicompany and not fullsharing mode; basic notification send link but always to object in first entity and display error.

htdocs/core/class/notify.class.php
```diff
@@ -24,7 +25,7 @@
  *      \brief      File of class to manage notifications
  */
 require_once DOL_DOCUMENT_ROOT.'/core/class/CMailFile.class.php';
-
+require_once DOL_DOCUMENT_ROOT.'/core/lib/functions.lib.php';
```
eldy, line 28: Why do you need to include functions.lib.php ?
A dolibarr class should be loaded into a dolibarr context, so main or master file have been already loaded and they automatically include this file already.

ptibogxiv: @eldy indeed. I have delete it

PR #16854, opened 2021-03-19
get conf global function

frederic34:


atm-maxime:
Does this mean we stop using $conf->global everywhere?

frederic34:
It's shorter than (empty($conf>global->MY_CONST) ? '' : $conf->global->MY_CONST) and because php8 doesn't like undeclared variable

PR #16891, opened 2021-03-20
NEW TakePOS cashiers and waiters

andreubisquerra:
Cashiers and waiters in TakePOS using user module

Why use the user table?
Because both use the same fields and cashier also can login to Dolibarr

What is the difference between a user and a cashier/waiter?
1. A cashier can have simple passwords or even no password. I commit it in the future.
2. A cashier sometimes may not be able to login to Dolibarr.
3. It is common to use a user to login to the ERP and then use several cashiers at the point of sale.
4. More

If is approved I will continue with more commits with cashiers

cashiers

### DIFF
--- a/htdocs/install/mysql/migration/13.0.0-14.0.0.sql
+++ b/htdocs/install/mysql/migration/13.0.0-14.0.0.sql
@@ -274,6 +274,8 @@ ALTER TABLE llx_salary CHANGE paye paye smallint default 0 NOT NULL;
 DELETE FROM llx_boxes WHERE box_id IN (SELECT rowid FROM llx_boxes_def WHERE file IN ('box_graph_ticket_by_severity', 'box_ticket_by_severity.php', 'box_nb_ticket_last_x_days.php', 'box_nb_tickets_type.php', 'box_new_vs_close_ticket.php'));
 DELETE FROM llx_boxes_def WHERE file IN ('box_graph_ticket_by_severity', 'box_ticket_by_severity.php', 'box_nb_ticket_last_x_days.php', 'box_nb_tickets_type.php', 'box_new_vs_close_ticket.php');
 
+ALTER TABLE llx_user ADD pos_cashier smallint DEFAULT 0;
+
 -- VMYSQL4.1 INSERT INTO llx_boxes_def (file, entity) SELECT  'box_graph_ticket_by_severity.php', 1 FROM DUAL WHERE NOT EXISTS (SELECT * FROM llx_boxes_def WHERE file = 'box_graph_ticket_by_severity.php' AND entity = 1);
 -- VMYSQL4.1 INSERT INTO llx_boxes_def (file, entity) SELECT  'box_graph_nb_ticket_last_x_days.php', 1 FROM DUAL WHERE NOT EXISTS (SELECT * FROM llx_boxes_def WHERE file = 'box_graph_nb_ticket_last_x_days.php' AND entity = 1);
 -- VMYSQL4.1 INSERT INTO llx_boxes_def (file, entity) SELECT  'box_graph_nb_tickets_type.php', 1 FROM DUAL WHERE NOT EXISTS (SELECT * FROM llx_boxes_def WHERE file = 'box_graph_nb_tickets_type.php' AND entity = 1);
--- a/htdocs/install/mysql/tables/llx_user.sql
+++ b/htdocs/install/mysql/tables/llx_user.sql
@@ -108,5 +108,6 @@ create table llx_user
   import_key        varchar(14),				-- import key
   default_range     integer,
   default_c_exp_tax_cat     integer,
-  fk_warehouse      integer						-- default warehouse os user
+  fk_warehouse      integer,					-- default warehouse os user
+  pos_cashier       smallint DEFAULT 0			-- user has cashier or waiter profile
 )ENGINE=innodb;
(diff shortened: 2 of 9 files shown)


daraelmin:
I vote for this enhencement as I have 22 cashier using their smartphone on 20 differents location

eldy:
A cashier is just a user that is in the user group of cashier. And this group of user only has the permission on takepos only.
If we add a field pos_cashier, we must add also a ton of other colon  pos_manager, user_bookkeeping,  etc..
Goal of permission is just to provide different features and permission for different people profile.

Having a user logged on ERP with one name and then using alternative profiles (for different cashier) is not compliant with modern security rules. It may means that the application will no more be compliant with some ISO standard, with some company policies, and even may become not legal in some countries (trackability and following good practices of security become more and more important with Saas expansion).

daraelmin:
the problem is that the cashiers (and I imagine the waiters) are responsible for their cash and the sums they carry. how could we distribute the amounts handled by each of them?

At the moment we are doing this parallel eb check on an excel board, based on usage by time slot (they don't work all 22 at the same time), but it is unsatisfactory.

daraelmin:
An alternative solution could be to allocate a specific account per user-cashier which would allow to trace only the movements that a user entered and which would allow to display the balance (what the waiter's purse contains) at the end of the day, week, month or year.

eldy:
The use/cashier is already tracked for each sale. A cashier can start/stop a cash fence when he want (when he start and when he stop its working session for example).

daraelmin:
Thanks for the answer. So I'll try to use as you described and maybe the "fix" for this feature should be on an enhecement of the wiki on TakePOS

andreubisquerra:
@eldy do you think that all the legal and security problems could be solved creating a cashiers table completely independent of the users?
Thanks to @daraelmin for your opinion

eldy:
A cashier is a physical people right ? So it must be a user. It must NOT be a different table.
Cashier is already to user in top right of interface.
If you want to switch to another cashier, to make a record into application or POS, it means you want to switch user.Switching need to enter a login and pass otherwise everybody can record every data for anybody. And if goal is to switch quickly,we must introduce a feature to switch user quickly but we must not introduce a second table of users (even if role of these users differs, role must be managed by permission groups and not by a second table since users are the same).

thegeettanna:
@andreubisquerra, 

The underlying concept of being able to associate another user to the TakePOS Ticket is something that can be used in a lot of ways. 
I am going to put up our use case below:

Based in India, our organizations has users in different departments like Marketing, Accounts, Support, etc.
So our Marketing team members bring in customers - (Access to commercial proposals & Sales Order Creation), but all our customer invoices are handled by our accounts department. That on the customer invoice module link the sales person who brings in that sale. This in return is similar to adding the author of sales order as sales person on customer invoice (extra field in invoices module). We use them to generate custom reports where the amount of sales invoices from each Marketing member is seen. 
Report Sales Person - This Month

Now, when a customer walks in, we use TakePOS for making an sale, but as TakePOS generates an invoice, our accounts team handle this part, but the sale is closed by one of our Marketing People (the sales person in this case). So we want to be able to associate the sales person to the invoice (specifically in the ef.salesperson extra field added in the invoices module). This way, we can consistently keep using the TakePOS for invoice generation for walk in customers. 

I am not sure how this fits into the larger picture. But hope this gives an insight into the uses of TakePOS and its user connection can help even the non-resturants and bar companies.

eldy:
I understand the need. But i dont understand why you need to add a new table of users for that. Everything you mention seems to be already possible with the current user table and existing links mechanism.
It is possible to add an option that show a combo with list of marketing people (list of users filtered on a user group or on a user tag for example or a full list) and the link this user to invoice during the sale so you can see it on tab "Contact/address of the invoice and make any reporting you need based on this association.

thegeettanna:
@eldy, 
I understand that it can be achieved with the current links mechanism. 
All I need is an option directly inside TakePOS, to link a user from the users current table to an extra field created in invoice module (called "salesperson"). 

That is what we actually require. I believe Andreu & myself were not on the same page and thus you can find two different set of thoughts on how to achieve it. But our application is as above mentioned. 
Again, @andreubisquerra might be working on a different set of customization as well. I am not sure.

eldy:
What is the definition of a "salesperson" of an invoice. The user that start to create the invoice, the user that modifiy it, the user that valid the invoice or the user that close the invoice ?

For storing all this information, you already have the fields :
fk_user_author		integer,								-- user making creation
 fk_user_modif         integer,                                -- user making last change
 fk_user_valid			integer,								-- user validating
 fk_user_closing          integer,                            -- user closing

And if it is nothing of this but just "another type of contact of invoice", you can use the system to link an invoice with a contact (must be used for any other type of contact).

thegeettanna:
I understand that there are a lot of ways to link a user to an invoice but it is not possible to link a user to an invoice when on create page. Due to which after each invoice is created, one has to go into the contact/addresses tab and then link a user to it. Since we have two accountants that make more than a hundred invoices, it is necessary for us to find a solution that doesn't require going into every invoice, because it is hard to track ones which are not linked. 

So instead we created an extra field called salesperson:
image
Now we can select an user directly when creating an invoice. 

But this is unavailable on TakePOS Module. We want to be able to link a salesperson directly when making an invoice from takepos. Preferably on the same extrafield that we created in invoice module, so our reports don't need to be customized anymore. In the end, TakePOS is creating an invoice, all we are looking for is an option to add the invoice extra fields on front end of the TakePOS Module.

andreubisquerra:
Solved with external module [TakePOS Cashiers](https://www.dolistore.com/en/modules/1397-TakePOS-Cashiers-13.html) and the @AXeL-dev help.
@eldy we will speak about this feature with time in next DevCamp (I hope soon)

PR #16900, opened 2021-03-21
NEW generate pdf when validating for send it before triggers

ptibogxiv:
changes need for #16846  (need generate PDF before calling triggers)

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -2571,13 +2571,14 @@ class Facture extends CommonInvoice
 	 * Object must have lines loaded with fetch_lines
 	 *
 	 * @param	User	$user           Object user that validate
-	 * @param   string	$force_number	Reference to force on invoice
+	 * @param string	$force_number	Reference to force on invoice
 	 * @param	int		$idwarehouse	Id of warehouse to use for stock decrease if option to decreasenon stock is on (0=no decrease)
 	 * @param	int		$notrigger		1=Does not execute triggers, 0= execute triggers
 	 * @param	int		$batch_rule		0=do not decrement batch, else batch rule to use, 1=take in batches ordered by sellby and eatby dates
+	 * @param	array		$pdf		array of values to generate pdf
 	 * @return	int						<0 if KO, 0=Nothing done because invoice is not a draft, >0 if OK
 	 */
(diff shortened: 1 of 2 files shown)


ptibogxiv:
@eldy when this PR will be OK for you; i will push others for others objects

htdocs/compta/facture/class/facture.class.php
```diff
@@ -2808,6 +2809,26 @@ public function validate($user, $force_number = '', $idwarehouse = 0, $notrigger
 				}
 			}
 
+			// Generate PDF
```
eldy, line 2812: I am not sure it is a good idea to merge 2 operations that are so different;
- validate is a low level database change
- generation of PDF is a heavy IO process (not necessarly done when validate is inside a batch that make other transformations after the validate for example).

So we must keep   validate()   and   generateDocument()   as   2 different operations not necessarly linked.

But we can add an option into the validate method to add the action so we will have both actions in same method if we provide parameter to 1, and we can still ignore document generation if we need. Why not just adding a test on

if (!empty($pdf)) {

This is our option.
Also please rename $pdf into $arraytobuilddoc   (doc is not always a PDF)

eldy: Then generation of doc is an operation not related to database, so it must be done completely at end, so after the trigger  BILL_VALIDATE that is still inside the database context.

ptibogxiv: the mistake with this that when validating no pdf or doc are joined with templated email because it was generated after.

all modern ERP/CRM can joined pdf or doc immediately  with notification in templated emails

ptibogxiv: @eldy have you some idea for notifications with doc (pdf or odt) 
the mistake is for validation because doc generation is after....

eldy: You can join pdf and doc immediatly. This is inside generateDocument(). For example, the module concatpdf can join an external pdf each time a pdf is modified.
For notification, we can imagine that the notification check the doc is up to date and update it if not but only when it needs to attach the file to it (so the notification can call the generateDocument() for example in case it need it).

ptibogxiv: good idea. I push a new PR

PR #16929, opened 2021-03-21
NEW: Add company info in dropdown login menu

aspangaro:
Useful for secretaries !

image

image

image


htdocs/main.inc.php
```diff
@@ -2138,6 +2153,17 @@ function top_menu_user($hideloginname = 0, $urllogout = '')
 	                $("#topmenu-login-dropdown").toggleClass("open");
 	            });
 
+	            $(".copyToClipboard").on("click", function() {
```
eldy, line 2156: To add feature for a copy/paste, there is already an implemented solution:
instead of 
print $myvalue
just do
print showValueWithClipboardCPButton($myvalue);

aspangaro: Thanks @eldy 

It's done

PR #16932, opened 2021-03-22
NEW Add native compression in rest apis

AlexisLaurier:
# If the user has some declared data compression ability and the server is able to do it, we send compressed data for API.

### DIFF
--- a/htdocs/api/index.php
+++ b/htdocs/api/index.php
@@ -315,19 +315,17 @@ Luracast\Restler\Defaults::$returnResponse = true;
 //print $api->r->handle();
 
 $result = $api->r->handle();
-if(isset($_SERVER['HTTP_ACCEPT_ENCODING'])) {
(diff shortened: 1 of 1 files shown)


htdocs/api/index.php
```diff
@@ -310,9 +311,22 @@
 // Call API (we suppose we found it).
 // The handle will use the file api/temp/routes.php to get data to run the API. If the file exists and the entry for API is not found, it will return 404.
 
-//Luracast\Restler\Defaults::$returnResponse = true;
+Luracast\Restler\Defaults::$returnResponse = true;
```
eldy, line 314: I suggest to introduce here (and later into section that compress result) an option to be able to disable the compression (we have trouble in past with it).
So to add a 
if (empty($conf->global->API_DISABLE_COMPRESSION)) {

So if user has a trouble with this new native compression enabled by default, we have a quick setup to restore old behaviour.

AlexisLaurier: @eldy I've added management of API_DISABLE_COMPRESSION const to be able to return to previous behavior

PR #16935, opened 2021-03-22
Update llx_c_forme_juridique.sql

dolibit-ut:
add at Germany:  'eG - eingetragene Genossenschaft'

moved belgium

formatting header


dolibit-ut:
Can we always delete the old ACTIVE fields?

eldy:
> Can we always delete the old ACTIVE fields?

Because the default value of field is 1, yes, we can remove the field if the entry must be loaded as enabled. But sometimes some entries are loaded "disabled by default" (because not common to use but loaded to save time for users that need it. He can just enable the line from dictionary setup)

dolibit-ut:
Okay, if we have lines with an active value '0' , then we need to keep the entry.

PR #16943, opened 2021-03-23
Develop fix keep values vat create form

atm-lena:
# Fix keep values vat create form

Keep values in VAT create form

### DIFF
--- a/htdocs/compta/paiement_vat.php
+++ b/htdocs/compta/paiement_vat.php
@@ -249,7 +249,7 @@ if ($action == 'create')
 
 	print '<tr>';
 	print '<td class="tdtop">'.$langs->trans("Comments").'</td>';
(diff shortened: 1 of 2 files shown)


htdocs/compta/paiement_vat.php
```diff
@@ -277,7 +277,7 @@
 			} */
 			$remaintopay = $objp->amount - $sumpaid;
 			print '<input type=hidden class="sum_remain" name="'.$nameRemain.'" value="'.$remaintopay.'">';
-			print '<input type="text" class="right width100" name="'.$namef.'" id="'.$namef.'" value="'.$remaintopay.'">';
+			print '<input type="text" class="right width100" name="'.$namef.'" id="'.$namef.'" value="'.GETPOST('amount_'.$objp->id, 'intcomma').'">';
```
eldy, line 280: The first time we entry the page, value will not be prefilled.
May be you should make here :
GETPOSTISSET($namef) ? GETPOST($namef, 'alpha') : remaintopay

PR #16949, opened 2021-03-24
Fix #13024 wrong line is used for reception

### LINKED ISSUE #13024: Order receptions: ErrorWarehouseRequiredIntoReceptionLine
# Bug
When receiving an order, we're encountering this error: "ErrorWarehouseRequiredIntoReceptionLine". It wasn't present in v10;  seems to be a regression.

## Environment
- **Version**: 11
- **OS**: Debian 9
- **Web server**: Apache2
- **PHP**: 7.3
- **Database**: MariaDB
- **URL(s)**: /reception/card.php?originid=xx&origin=supplierorder

Should work as the Warehouse is specified. Currently, triggers the "ErrorWarehouseRequiredIntoReceptionLine" error

Create any order and try to receive products from it.

[*Files*]

Capture d’écran 2020-02-04 à 14 30 50

d3v2a:
# Fix #13024
Line tested for fk_product is the wrong line

### DIFF
--- a/htdocs/reception/card.php
+++ b/htdocs/reception/card.php
@@ -317,10 +317,15 @@ if (empty($reshook)) {
 			//var_dump($_POST);exit;
 			for ($i = 1; $i <= $num; $i++) {
 				$lineToTest = '';
+				$lineId = GETPOST($idl, 'int');
 				foreach ($objectsrc->lines as $linesrc) {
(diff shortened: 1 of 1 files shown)


vincentjdc:
That also seems to work for version 13.0.x. Is there any reason why this fix is not also applied to that version ?

PR #16955, opened 2021-03-24
FIX : wrong dol_get_fiche_end()

atm-gauthier:
Hi @eldy, i did here a little correction with a duplicated dol_get_fiche_end

I also have a question about VAT payments and salary payments :

I saw that for salary, we can use an arrow to fill remain amount to pay (like invoices for example), and for vat remain amount to pay is automatically filled.
Wouldn't it be better to uniformize that ?
Do you prefer use this arrow as well as all payment screens of dolibarr or not ?

image

image

### DIFF
--- a/htdocs/salaries/paiement_salary.php
+++ b/htdocs/salaries/paiement_salary.php
@@ -174,7 +174,7 @@ if ($action == 'create') {
 	print '<input type="hidden" name="chid" value="'.$chid.'">';
 	print '<input type="hidden" name="action" value="add_payment">';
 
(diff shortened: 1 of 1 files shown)


eldy:
For vat payment, in 90% of time we pay the amount of payment of the vat (and there is always 1 and only 1 vat declaration to pay). So we can fill automatically the amount.
It is the same for salary. We can fill automatically amount to save time.

For invoice, we can fill automatically but is not always what we expect. It may have a lot of invoice, and payment are often partial so it is better to keep as it is now

PR #16956, opened 2021-03-24
FIX : old query using tva object for payments list

atm-gauthier:


htdocs/compta/charges/index.php
```diff
@@ -271,20 +273,21 @@
 
 	print load_fiche_titre($langs->trans("VATPayments").($year ? ' ('.$langs->trans("Year").' '.$year.')' : ''), '', '');
 
-	$sql = "SELECT pv.rowid, pv.amount, pv.label, pv.datev as dm, pv.fk_bank,";
+	$sql = "SELECT ptva.rowid, pv.rowid as id_tva, pv.amount as amount_tva, ptva.amount, pv.label, pv.datev as dm, ptva.datep as date_payment, ptva.fk_bank,";
```
eldy, line 276: It seems you replace the report that show list of vat declaration + amount paid with list of amount paid only. (vat declaration without payment disappears from report).
Is looks better to replace the title of page "Payment of vat" into "Vat declarations and payments"

atm-gauthier: @eldy i understand, then i replaced in my last commit "INNER JOIN" by a "LEFT JOIN", so now we can see all vat declarations even if there are no payments, and now we get a similar view to social charges view.
What do you think about that ?

image

PR #16975, opened 2021-03-25
Input multiple serial numbers at once

altairis-tof:
This allow to input multiple serial numbers in dispatch, stock correction and stock transfer for product managed by SN.
We also introduced a setup defined separator (default is ";").


lmag:
That ok for me @eldy

htdocs/fourn/commande/dispatch.php
```diff
@@ -886,20 +933,24 @@
 
 							print '<tr class="oddeven" name="'.$type.$suffix.'">';
 							print '<td>';
-							print '<input name="fk_commandefourndet'.$suffix.'" type="hidden" value="'.$objp->rowid.'">';
-							print '<input name="product_batch'.$suffix.'" type="hidden" value="'.$objp->fk_product.'">';
+							print '<input name="fk_commandefourndet'.$suffix.'" type="hidden" value="'.$objp->rowid.'" />';
```
eldy, line 936: Do not use xml ending syntax, even if html5 is able to accept it, using pure html is better

altairis-tof:
ok it's fixed.

htdocs/fourn/commande/dispatch.php
eldy: For entry of product/serial/qty for reception (and shipment), a more powerful way to manage multiple serial input is planned.
If you take a look at the page to enter a reception, you will see a picto "split" to split a line into several lines so we are able from this screen to dispatch several lot, qty, warehouse target.
For mass input we must keep this philosophy. What is planned is:
Having a button "Mass input" that open a popup. This popup will have one textarea empty by default. We must be able to enter (by copy/paste for example from an excel sheet) a content that is
LOT1
LOT2
LOT3
or 
LOT1;QTY1
LOT2;QTY2
LOT3;QTY3
or 
LOT1;QTY1;WAREHOUSEID1
LOT2;QTY2;WAREHOUSEID1
LOT3;QTY3;WAREHOUSEID1

When clicking on "Close popup", it will automatically cal the js method addDispatchLine() for each line entered, the same way it is done when we click on the "split" picto and the entered value into line is filled by js into this new line.
We must proceed like this for several reason:
- We will be able to use the mass werehouse selection even on lot entered massively, keeping ability to use different warehouses.
- Mass entry will share the same interface than manual entry with all the powerfull of manual entrie.
- We will be able to fill the popup textarea field by using a barcode reader by scanning several barcode in a serial process (it will make LOT1, LOT2, LOT3 appearing with still the ability to modify, fix manually while keeping scanning)
- We will be able also to load a file into this popup so to make mass entry o serial from a file; a third way to enter data, but with always using the same interface.
- Such an interface can be used at a lot of other place.

For the page Stock transfer and Stock movement, there is currently no picto and no js function but we can image to show the picto just after the field "lot serial". But your first change to use a textarea instead of input is 

Also can you try to split this PR is smaller pieces (atomic PR), it is easier to discus on each feature on problem on some changes does not block the others.

altatof: ok i will look at all this... and will also make smaller PRs.
Thanks

lmag: This is the discussion we had with @altatof . We plan to do this version later. We will see that.

lmag: LOT1;QTY1;WAREHOUSEID1
LOT2;QTY2;WAREHOUSEID2
LOT3;QTY3;WAREHOUSEIDX
Ou
SerialNumber;1;WAREHOUSEID1

lmag: We will propose a Workflow with screenshots

htdocs/fourn/commande/dispatch.php
```diff
@@ -465,11 +490,33 @@
 					$errors = $mouv->errors;
 					$error++;
 				} else {
-					$mouv->origin = &$object;
-					$result = $mouv->reception($user, $product, $supplierorderdispatch->fk_entrepot, $supplierorderdispatch->qty, $price, $comment, $eatby, $sellby, $batch);
-					if ($result < 0) {
-						$errors = $mouv->errors;
+					// if serial number management, delete before recreate (to avoid problem of already existing product_lot)
```
eldy, line 493: I don't understand why deleting ProductLot is required.
A record is entered into the table the first time we see the lot. But then we are able to make other move using them lot with no need to delete the record.

htdocs/product/class/product.class.php
```diff
@@ -4913,30 +4913,59 @@ public function correct_stock($user, $id_entrepot, $nbpiece, $movement, $label =
 	 * @param  int	    $disablestockchangeforsubproduct	Disable stock change for sub-products of kit (usefull only if product is a subproduct)
 	 * @return int                      <0 if KO, >0 if OK
 	 */
-	public function correct_stock_batch($user, $id_entrepot, $nbpiece, $movement, $label = '', $price = 0, $dlc = '', $dluo = '', $lot = '', $inventorycode = '', $origin_element = '', $origin_id = null, $disablestockchangeforsubproduct = 0)
+	public function correct_stock_batch($user, $id_entrepot, $nbpiece, $movement, $label = '', $price = 0, $dlc = '', $dluo = '', $batch = '', $inventorycode = '', $origin_element = '', $origin_id = null, $disablestockchangeforsubproduct = 0)
```
eldy, line 4916: This function is a low level function and must remains as it is.
It is the caller that must split the batch number if it is a string with several values and then call this function several times for each batch number.

htdocs/product/stock/tpl/stockcorrection.tpl.php
```diff
@@ -78,8 +78,8 @@ function init_price()
 	}
 	print img_picto('', 'stock').$formproduct->selectWarehouses($ident, 'id_entrepot', 'warehouseopen,warehouseinternal', 1, 0, 0, '', 0, 0, null, 'minwidth100');
 	print ' &nbsp; <select class="button buttongen" name="mouvement" id="mouvement">';
-	print '<option value="0">'.$langs->trans("Add").'</option>';
-	print '<option value="1"'.(GETPOST('mouvement') ? ' selected="selected"' : '').'>'.$langs->trans("Delete").'</option>';
+	print '<option value="3">'.$langs->trans("Add").'</option>';
```
eldy, line 81: Why renaming 0 into 3.
This seems to break compatibility/feature with the page that include the tpl and manage the "correct_stock" action that need 0 or 1 to know if it is a + or - stock movement.

altairis-tof:
i close this PR and start to make smaller ones.

htdocs/fourn/commande/dispatch.php
lmag: image

lmag: I think we need more information like : 
WAREHOUSEIDX,Qty,Lot-orserialnumber;DMD;DLC;Prix;Projet;libellé;codemvt

PR #16977, opened 2021-03-25
NEW: Check Thirdparty name

milenmk:
# New [*Check Thirdparty name*]

When adding a new client, check if a Third Party name already exists and throw an error if so.


atm-maxime:
🤔 two thirdparties can have the same name...

htdocs/societe/card.php
```diff
@@ -406,6 +406,20 @@
 			setEventMessages($langs->trans("ErrorFieldRequired", $langs->transnoentities("No_Email")), null, 'errors');
 		}
 
+		if ($action == 'add') {
+			$thirdparty = GETPOST('name');
+			$sql = 'SELECT nom FROM '.MAIN_DB_PREFIX.'societe WHERE nom="'.$db->escape($thirdparty).'"';
```
eldy, line 411: It is frequent that 2 companies has the same name. 
Currenlty this is allowed into Dolibar but we may add an option to forget this.
You can add a test here 
if (! empty($conf->global->THIRDPARTY_NAME_MUST_BE_UNIQUE))

daraelmin: What about individual ?  How would you manage namesake of humain being ?
How would you manage multiple business units, parent and child society with sans name ?

milenmk: It is possible to add a filter to this, like 

`
if (! empty($conf->global->THIRDPARTY_NAME_MUST_BE_UNIQUE))
`

BB2A-Anthony:
It's interesing add option

And if existe proposed to create or not

It's interesing for product

eldy:

PR #16982, opened 2021-03-27
NEW: It's time to remove Fax

aspangaro:
I think it's time to say good bye to the fax 👋


atm-maxime:
Agreed, but maybe a bit hard to delete everywhere, some irreducible users might still use it...

frederic34:
It's time to be able to add multiple phone numbers with type (fax, mobile, pers, pro etc...)

aspangaro:
Yes @frederic34 like your social network system maybe ?

frederic34:
Look at https://github.com/Dolibarr/dolibarr/pull/16987 and comments are welcome

BB2A-Anthony:
Why not rename fax to phone_bis or other

And 100%  agre with @frederic34

daraelmin:
In switzerland, faxes are still used, in particular by lawyers, trustees and administrations, because it is the only way (other than by post or personal delivery) to transmit a signed contract or mandate which has the value of proof in the event of dispute.

Issue #15831, opened 2020-12-29
Feature Request : CDAV (carddav) for adherent and/or third-party


daraelmin:
# Feature Request
The CDAV module should offer the possibility to sync adherent and third party.

## Use case
Association are more often in contact with their adherent than external companies. A possible workaround could be to offer the sync carddav with the third party.
It could be useful to use as a WhatsApp diffusion list for association
For large association it is very useful to have the contact sync on the phone to know who is calling.

## Suggested implementation
In CDAV

dolibit-ut:
You mean CardDAV at the (sabre) DAV Module.
I think there is such a module from befox - if you look for befox/cdav.

daraelmin:
Ok it seems the befox/cdav is doing the job. Why this module isn't include in the core ?

BB2A-Anthony:
> Ok it seems the befox/cdav is doing the job. Why this module isn't include in the core ?

Yes it's very more for dolibarr if it is include in the core

soerpoers:
same for me – i think carddav/caldav is useful to sync with most platforms and keep up to date…

Bagu:
> > Ok it seems the befox/cdav is doing the job. Why this module isn't include in the core ?
> 
> Yes it's very more for dolibarr if it is include in the core

+1 here

zunami:
I think it is a pity that this ticket was closed. 

I am of the same opinion. Why is this not included in the dolibarr core system? the times are becoming more modern how to communicate and you need this in everyday life where you check your contacts.

please add cdav to the dolibarr core system

PR #16987, opened 2021-03-27
add new db tables to store emails and phones numbers

frederic34:
@eldy @aspangaro @atm-maxime @hregis @dolibit-ut @daraelmin @FHenry @altatof @ptibogxiv and any other interested
I propose to migrate all emails and phones numbers into separate table, so members, socpeoples, thirdparties and users can have multiple emails and phones (without loosing old old :fax: numbers, but still used)
Thank's in advance for comments

### DIFF
--- a/htdocs/install/mysql/data/llx_c_type_contacting_object.sql
+++ b/htdocs/install/mysql/data/llx_c_type_contacting_object.sql
@@ -19,7 +19,7 @@
 -- de l'install et tous les sigles '--' sont supprimés.
 --
 
--- type_emails
+-- type_contacting_object
 
-INSERT INTO llx_c_type_emails (code, label, active, position) VALUES ('business', 'BusinessEmail', 1, 10);
-INSERT INTO llx_c_type_emails (code, label, active, position) VALUES ('personal', 'PersonalEmail', 1, 20);
+INSERT INTO llx_c_type_contacting_object (code, label, active, position) VALUES ('BUSINESS', 'BusinessContactingObject', 1, 10);
+INSERT INTO llx_c_type_contacting_object (code, label, active, position) VALUES ('PRIVATE', 'PrivateContactingObject', 1, 20);
--- a/htdocs/install/mysql/data/llx_c_type_emails.sql
+++ b/htdocs/install/mysql/data/llx_c_type_emails.sql
@@ -1,0 +1,25 @@
+--
+-- This program is free software; you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as published by
+-- the Free Software Foundation; either version 3 of the License, or
+-- (at your option) any later version.
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU General Public License for more details.
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program. If not, see <http://www.gnu.org/licenses/>.
+--
+--
+
+--
+-- Ne pas placer de commentaire en fin de ligne, ce fichier est parsé lors
+-- de l'install et tous les sigles '--' sont supprimés.
+--
+
+-- type_emails
+
+INSERT INTO llx_c_type_emails (code, label, active, position) VALUES ('business', 'BusinessEmail', 1, 10);
+INSERT INTO llx_c_type_emails (code, label, active, position) VALUES ('personal', 'PersonalEmail', 1, 20);
--- a/htdocs/install/mysql/data/llx_c_type_phones.sql
+++ b/htdocs/install/mysql/data/llx_c_type_phones.sql
@@ -21,7 +21,7 @@
 
 -- type_phones
 
-INSERT INTO llx_c_type_emails (code, label, active, position) VALUES ('PHONE', 'PhoneStd', 1, 10);
-INSERT INTO llx_c_type_emails (code, label, active, position) VALUES ('PHONEMOBILE', 'PhoneMobile', 1, 20);
-INSERT INTO llx_c_type_emails (code, label, active, position) VALUES ('PHONEFAX', 'PhoneFax', 1, 30);
-INSERT INTO llx_c_type_emails (code, label, active, position) VALUES ('PHONEPAGER', 'PhonePager', 1, 40);
+INSERT INTO llx_c_type_phones (code, label, active, position) VALUES ('PHONE', 'PhoneStd', 1, 10);
+INSERT INTO llx_c_type_phones (code, label, active, position) VALUES ('PHONEMOBILE', 'PhoneMobile', 1, 20);
+INSERT INTO llx_c_type_phones (code, label, active, position) VALUES ('PHONEFAX', 'PhoneFax', 1, 30);
+INSERT INTO llx_c_type_phones (code, label, active, position) VALUES ('PHONEPAGER', 'PhonePager', 1, 40);
--- a/htdocs/install/mysql/data/llx_c_type_phones.sql
+++ b/htdocs/install/mysql/data/llx_c_type_phones.sql
@@ -1,0 +1,27 @@
+--
+-- This program is free software; you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as published by
+-- the Free Software Foundation; either version 3 of the License, or
+-- (at your option) any later version.
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU General Public License for more details.
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program. If not, see <http://www.gnu.org/licenses/>.
+--
+--
+
+--
+-- Ne pas placer de commentaire en fin de ligne, ce fichier est parsé lors
+-- de l'install et tous les sigles '--' sont supprimés.
+--
+
+-- type_phones
+
+INSERT INTO llx_c_type_emails (code, label, active, position) VALUES ('PHONE', 'PhoneStd', 1, 10);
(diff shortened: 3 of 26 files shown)


frederic34:
a phone can be std, fax mobile etc, can be businness or private and can be default phone to use
an email can be business or private and can be default email to use

hregis:
@frederic34 wouahou... It's very complicated! Why just one table with type of element?

ptibogxiv:
@frederic34 can you include fax too (it's a good PR to disabled fax but keep it when needed)

hregis:
And type of media?

BB2A-Anthony:
It's very good proposal I follow you 100%

You need an email and a phone by default it's on for read on head

frederic34:
> @frederic34 wouahou... It's very complicated! Why just one table with type of element?

That's why I'm asking everybody, but for foreign keys and speed I think it's better, but I may be wrong

> @frederic34 can you include fax too (it's a good PR to disabled fax but keep it when needed)

fax are still there, I set a dictionary of phones types

> And type of media?

can you develop ?

> You need an email and a phone by default it's on for read on head

yes it's in priority column(sort of rank), by example for user asking new password or for preferred phone number etc

hregis:
One table with type of media (tel, fax, e-mail, etc...) and with elements (user, customer, adherent, etc...)

daraelmin:
Great Idea... But what a hard work !

May I suggest to :
- add postal address, social network and website also 
- remove dupplicated column (adherent, third party, contact, user, etc.)

frederic34:
> One table with type of media (tel, fax, e-mail, etc...) and with elements (user, customer, adherent, etc...)

so in commonobject a function for all

ptibogxiv:
It will be si good to include country origin for phone/fax number too ie: fr -> +33

hregis:
Hard work with all tables (llx_member_email, llx_member_phone, etc...) just one table with "type" (tel, fax, etc..) and "element" (user, customer, member, etc...) is very "souple" for me 😉

frederic34:
hum hum one table to store

- emails
- phones
- address
- website url
- ...

define for each object a schema and store them in json format

daraelmin:
Why in json and not in carddav?

atm-maxime:
Seems a good idea. I agree one table is enough.
But isn't it hard work for few use case ? Usually people, members, users have one email and one mobile phone. But it would be more compliant with most address books.

daraelmin:
> Seems a good idea. I agree one table is enough.
> But isn't it hard work for few use case ? Usually people, members, users have one email and one mobile phone. But it would be more compliant with most address books.

When an association has employee or elected member as employee, you can different profil for the same person dependong if you re looking for the member (adherent), the employee (user) or the client (thirdparty).

For Union (syndicat), the member has also a professional profil to store.

fappels:
One phonenumber can be linked to more telecom service (Signal, whatsapp, telegram, ...).

atm-maxime:
> When an association has employee or elected member as employee, you can different profil for the same person dependong if you re looking for the member (adherent), the employee (user) or the client (thirdparty).

Agreed, but you already can have different email and phone on the user card (employee), member card and customer card. And your example is about foundations, which doesn't represent all Dolibarr users. I'm ok with the functionality because it's more compliant with current standards, just pointing out that the need doesn't concern the user majority I think.

altairis-tof:
hello, i also vote for one single table. 
@daraelmin also suggest carddav; it seems to be interesting; maybe @daraelmin can tell us more about it ?

daraelmin:
@altatof 
In fact... I'm ignorant about this standart.
But I need to sync my adherent with an nextcloud instance.
Here is a feature request linked to carddav  #15831 and the module suggested is here https://github.com/Befox/cdav

BB2A-Anthony:
I am ready to help in the dev if idea is validated by @eldy .
I think you have to be careful with API reset

frederic34:
how to search in db if you store info in carddav format?

daraelmin:
> how to search in db if you store info in carddav format?

Right.

daraelmin:
From what I read, I imagine a really hard reenginering. Have you already tried to schematize?

It seems to me, that we can inspirate ourselves from the extrafields management. However I wonder if the query consultation performance will not be too impacted ... otherwise the big stake will be to "fix" some default value to guarantee functionality in the modules.

To me it appears we are talking of something like this

##sql side
An "entity" table with the base fields and an "entity property" table with relatively few columns each of which is indexed.

something like:
ENTITY (*id_entity*, company_name, gender, firstname, lastname, typeent, parent_entity_id, avatar_img)
ENTITY_PROPERTIES (*fk_id_entity, property*, value, default_in_module)
and a value diconnaire for "property".

The entity key is the only thing register in the member, company, user and contact tables and we delete the other values of type " profil". 

##php side

default fields mandatory for module fonctionnality, but selectable among the values already registered for the entity

Admin user should add mandatory fields or tell if default field are mandatory.

a possibility to add new property (within property_dictionnary) in the profile of the entity ( in a dedicated page or in each module with a core include or maybe both)

##js side
Ajax request for adding property and save them on update

##Impact
Performance ?
Vcard
Module retro compatibilities
Migration
Extrafield
Anything else ?

BB2A-Anthony:
For migrate slowly whay

Add fonctionality and not touch the fonctionality is in inplace.  The number and mail this in place are not touching.  

For migration

Add number and mail in new table (set by default) dont touch in old emplacement if next time edit , modify on old emplacement and new table

You understant?

daraelmin:
Ok. 
I assume you have a plan, so maybe I can help doing part of it. 
Just tell me wich steps you are thinking and on witch part you need volonteer. I'll give you a hand if I can.

EDIT : Sorry I've found the plan here : #16989

eldy:
What is goal of adding a table compared to adding extrafields.
Moving horizontal data into a vertical storage leads to a lot of problem (how can be extract data in excel, how do we import, what about the search for a part of a field or end with or start with, what about merging thirdparties, finding duplicates, etc...)
Currently users does not see how data is stored. They just enter their data because fields for multiple phone, fax, etc, already exists.

What is the use case (a real life situation) that can justify that we must make architecture more complex that currently ? Does this happen frequently that we need to store more than 3 phone numbers for 1 thirdparties, knowing that, if we have more phone number, we must manage this by the "alternative contact record" (so contact/address).
For members, we also have 3 fields for phone but we can image to add a type "Pro, Perso, Mobile Pro, Mobile Person Fax, ..." to each of this 3 fields for a more accurate storage.
But using a vertical storage for just phone (when there is already a feature for that) seems a gaz engine for a very very small gain (if fact i don't really see the gain for end users compared to adding a simple "type of number" for each field). This was legitimate for social networks, because there is several hundred of social networks, and each user use a different family of social networks, but looks strange to me for just phone numbers.

daraelmin:
> What is goal of adding a table compared to adding extrafields.
> Moving horizontal data into a vertical storage leads to a lot of problem (how can be extract data in excel, how do we import, what about the search for a part of a field or end with or start with, what about merging thirdparties, finding duplicates, etc...)
> Currently users does not see how data is stored. They just enter their data because fields for multiple phone, fax, etc, already exists.
> 
> What is the use case (a real life situation) that can justify that we must make architecture more complex that currently ? Does this happen frequently that we need to store more than 3 phone numbers for 1 thirdparties, knowing that, if we have more phone number, we must manage this by the "alternative contact record" (so contact/address).
> For members, we also have 3 fields for phone but we can image to add a type "Pro, Perso, Mobile Pro, Mobile Person Fax, ..." to each of this 3 fields for a more accurate storage.
> But using a vertical storage for just phone (when there is already a feature for that) seems a gaz engine for a very very small gain (if fact i don't really see the gain for end users compared to adding a simple "type of number" for each field). This was legitimate for social networks, because there is several hundred of social networks, and each user use a different family of social networks, but looks strange to me for just phone numbers.
> 

To me, the real benefit of having a separate table is to refer from all module and table to the same data. That may simplify the sync, export and import.

daraelmin:
I have some use case : when user is employee, we should use his personnal adress when sending him his contract or salary but we must use our professionnal adress when he contact client or prospect or to impress visit card
With multy company... The same employee could have multiple adresses or phones...

eldy:
For personal data, there is already fields to store data (like birthday, or personal phone, personal address of a user is already the field address, zip and country of the user card).
Currently we don't have more, because there is no module yet that need to manage more. 
This will be the goal of module HRM and it may introduce more fields dedicated for this module (the fields this modules need and that differs from other existing data) but because this data remains related to user/employee, it must be on the dedicated table of user. There is no need for a vertical table to store horizontal data.
My question was "a use case that show we must store data in a vertical table instead of an horizontal table)". The use case you give show we can perfectly work with current table (we are speaking about current phone numbers of a user), even if we currently miss some fields that will be managed by the module HRM in the future. 
if an employee has several address, for its employment coordinates, only the main address will be necessary for contracts (history may be required by HRM module but in such a case, it will be the role of the module HRM to introduce a table for history purpose). If a user absolutely needs to store 2 current adresses for the same employee, it is still possible to add extra fields. For most common need (and the core must cover the common need) current features (existing fields + extra fields) seems to provide the needs

philippe-opendsi:
Hello, I will add one need when using in multicompany mode, a user can have different email adress depending on the entity he is working on.

hregis:
@philippe-opendsi  i can add this feature in the module

atm-maxime:
> Hello, I will add one need when using in multicompany mode, a user can have different email adress depending on the entity he is working on.

I asked myself the same and ended up to create a different user account. This is not the same user that works in 2 different entities.

daraelmin:
Wouldn't it be more coherent to remove the redundant columns across the member / company / employee / user / contact / etc tables to group them into one (legal_entitey) which all the modules would refer to?
a single table with a single extrafield per legal entity
the management of name, email and telephone changes would thus be facilitated by centralization as well as searches.

When someone gets married, their name must be changed and checked in 3 to 6 tables ... same for a move, a change of operator or email address

um recent example:
One member changed her name after her marriage. we changed the name in adherent first ... But the thirdparty (client) hasn't be affected. we have therefore been modifying the name in the thirdparty... the operation causes a synchro and has modify ... the name of the company which employs our member ....

I'm not sure adding tables is relevant if we don't remove their modular assignment.

frederic34:
add next comments on issue.

PR #16988, opened 2021-03-27
Fix #16899 New stuff related to excluded members forgotten during transfer to box graph

daraelmin:
# Fix missing new excluded member in box_members_by_type
when migrating from adherents/index.php, it seems that the new status was forgotten
# Fix use English for coding in box_members_by_type
Translation Somme A to SumToValidate, etc.
# Fix use a plain red disc instead of a read circle to list excluded member
It's more logical to use the plain disc in the graph (as we do for the out-of-date witch should be an orange circle but we use a plain orange disc)
### ACTUALLY
image
### WITH PR 
image

### DIFF
--- a/htdocs/adherents/class/adherent.class.php
+++ b/htdocs/adherents/class/adherent.class.php
@@ -2280,7 +2280,7 @@ class Adherent extends CommonObject
 			$labelStatus = $langs->trans("MemberStatusResiliated");
 			$labelStatusShort = $langs->trans("MemberStatusResiliatedShort");
 		} elseif ($status == -2) {
-			$statusType = 'status8';
+			$statusType = 'status10';
 			$labelStatus = $langs->trans("MemberStatusExcluded");
 			$labelStatusShort = $langs->trans("MemberStatusExcludedShort");
 		}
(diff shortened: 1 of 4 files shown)


eldy:
Note that in dolibarr color convention, a full plain color means the object is "live", and if color is red or orange, it means there is something to do or urgent to do.
Here when a member is excluded, the member become a "dead" record we don't want to deal with anymore (it's the past and we want to forget it). For this reason, showing a highly visible alert (plain color) is not a good choice (there is no urgent action to process for this user).

daraelmin:
> Note that in dolibarr color convention, a full plain color means the object is "live", and if color is red or orange, it means there is something to do or urgent to do.
> Here when a member is excluded, the member become a "dead" record we don't want to deal with anymore (it's the past and we want to forget it). For this reason, showing a highly visible alert (plain color) is not a good choice (there is no urgent action to process for this user).

OK, but it's not really readble in graph.

By the way, why an orange circle for out-of-date and a plain one for waiting for subscription ? To me it is the same 

(And just a word to be accurate we don't want to forget the excluded, because it s important not to deal anymore with it)

PR #16990, opened 2021-03-28
Close #15830 Add setup option ADHERENT_DEFAULT_CREATE_THIRDPARTY

### LINKED ISSUE #15830: Feature Request : setup option ADHERENT_DEFAULT_CREATE_THIRDPARTY
# Feature Request
Add a (hidden) option ADHERENT_DEFAULT_CREATE_THIRDPARTY

## Use case
When you use the adherent module it is common to use a third party linked to sell or renew subscription.
It would be very user-friendly if we could automatically create this third party when the adherent is validated just as it is possible to create a contact for an individual third party (THIRDPARTY_DEFAULT_CREATE_CONTACT).

daraelmin:
# Close #15830 Add setup option ADHERENT_DEFAULT_CREATE_THIRDPARTY

### DIFF
--- a/htdocs/societe/class/societe.class.php
+++ b/htdocs/societe/class/societe.class.php
@@ -3795,7 +3795,8 @@ class Societe extends CommonObject
 		$this->client = 1; // A member is a customer by default
 		$this->code_client = ($customercode ? $customercode : -1);
 		$this->code_fournisseur = -1;
-		$this->typent_id = ($member->morphy == 'phy' ? 8 : 0); // The type of thirdparty is private (individual) if the member is human (phy)
+		$this->typent_code = ($member->morphy == 'phy' ? 'TE_PRIVATE' : 0);  
+		$this->typent_id = dol_getIdFromCode($this->db, $this->typent_code, 'c_typent', 'id', 'code'); 
 
 		$this->db->begin();
 
--- a/htdocs/societe/class/societe.class.php
+++ b/htdocs/societe/class/societe.class.php
@@ -3796,23 +3796,23 @@ class Societe extends CommonObject
 		$this->code_client = ($customercode ? $customercode : -1);
 		$this->code_fournisseur = -1;
 		$this->typent_id = ($member->morphy == 'phy' ? 8 : 0); // The type of thirdparty is private (individual) if the member is human (phy)
-		
+
 		$this->db->begin();
 
 		// Cree et positionne $this->id
 		$result = $this->create($user);
-		
+
 		if ($result >= 0) {
 			// Auto-create contact on thirdparty creation
-		    if (!empty($conf->global->THIRDPARTY_DEFAULT_CREATE_CONTACT)) {
+			if (!empty($conf->global->THIRDPARTY_DEFAULT_CREATE_CONTACT)) {
 				// Fill fields needed by contact
-		        $this->name_bis = $member->lastname;
+				$this->name_bis = $member->lastname;
 				$this->firstname = $member->firstname;
 				$this->civility_id = $member->civility_id;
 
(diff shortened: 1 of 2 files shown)


htdocs/societe/class/societe.class.php
```diff
@@ -3793,11 +3795,13 @@ public function create_from_member(Adherent $member, $socname = '', $socalias =
 		$this->client = 1; // A member is a customer by default
 		$this->code_client = ($customercode ? $customercode : -1);
 		$this->code_fournisseur = -1;
+		$this->typent_id = ($member->morphy == 'phy' ? 8 : 0); // The type of thirdparty is private (individual) if the member is human (phy)
```
eldy, line 3798: You must not use "8" here but get the id from the code 'TE_PRIVATE'. You can use method getIdFromCode... for that.

daraelmin: I didn't find the method you are talking about.
Should I use the dol_function like this (from societe/card.php)
`$object->typent_code = dol_getIdFromCode($db, $object->typent_id, 'c_typent', 'id', 'code'); `
or maybe the societe method : setThirdpartyType
`$this->setThirdpartyType(8);`

eldy: Yes, it's dol_getIdFromCode

daraelmin:
@eldy 
Hope I write it the way you were expecting.
Do you think it is possible to sync the change (update) between adherent, third-party and contact ? A
And would you approve that as a basic feature or an hidden option feature ?

htdocs/societe/class/societe.class.php
```diff
@@ -3793,11 +3795,14 @@ public function create_from_member(Adherent $member, $socname = '', $socalias =
 		$this->client = 1; // A member is a customer by default
 		$this->code_client = ($customercode ? $customercode : -1);
 		$this->code_fournisseur = -1;
+		$this->typent_code = ($member->morphy == 'phy' ? 'TE_PRIVATE' : 0);
+		$this->typent_id = dol_getIdFromCode($this->db, $this->typent_code, 'c_typent', 'id', 'code');
```
eldy, line 3799: You can call the dol_getIdFromCode only if code is 'TE_PRIVATE' to avoid a useless sql access if empty.

eldy:
There is no relation between a member and an "alternative contacts/adresses" so no need for sync.
I think there is (it should have) already a sync during update between "member" - "thirdparty"  and  "user" (the 3 object may be linked), at least for data like address. If you find a data not sync tell me for which object and from which object...

htdocs/societe/class/societe.class.php
daraelmin: Done

daraelmin:
> There is no relation between a member and an "alternative contacts/adresses" so no need for sync.

When THIRDPARTY_DEFAULT_CREATE_CONTACT is set to 1, contact, that should be linked to thirdparty is created. But you're right it is not.

Now with this feature, when an adherent is created, a third-party is created and a contact. 

So if I want to sync thirdparty and the default contact created. This way I'm hoping syncing contact with member.

Why ? 

Because, we want to give access to some smartphone to adress book and to update member directly from there.
So I'm trying to get access to member on cellphones (we have 700 active members) with DAVx, via bfox module Cdav.

> I think there is (it should have) already a sync during update between "member" - "thirdparty"  and  "user" (the 3 object may be linked), at least for data like address. If you find a data not sync tell me for which object and from which object...
> 
In V14, I have done the two tests below
When changing the name of a third party (in case of misspelling for example) linked to a member, it doesn't change the name of the member. 
When setting an parent thirdparty it doesnt affect a company to the member.

PR #16991, opened 2021-03-29
fix : when thridpartry lis is not filtered ans do a massaction filter changes

FHenry:
On fundation dolibarr go to
 -  thirdparty list
 - check somes ilen selected delete or affect tag mass action
 => When page is reloaded the list is not the same because filter search_sales is -1 and not empty and force link with sales man


FHenry:
REvienw in live better fix in other pr

PR #16996, opened 2021-03-29
llx_product batch_mask definition

altairis-tof:
Begin to implement batch mask per product.
Next steps will be to : 
- add a "batch mask per product" option in product lot setup (admin card)
- manage batch_mask in create and update functions of product class
- ability to input and update batch mask in product card if "batch mask per product" is on in setup

And also the use of the mask in the mo production tab (mask per product or other setup defined masks if any)

### DIFF
--- a/htdocs/install/mysql/tables/llx_product.sql
+++ b/htdocs/install/mysql/tables/llx_product.sql
@@ -51,7 +51,7 @@ create table llx_product
   recuperableonly               integer NOT NULL DEFAULT '0',       -- French NPR VAT
   localtax1_tx                  double(6,3)  DEFAULT 0,
   localtax1_type                varchar(10)  NOT NULL DEFAULT '0',
-  localtax2_tx                  double(6,3)  DEFAULT 0, 
+  localtax2_tx                  double(6,3)  DEFAULT 0,
   localtax2_type                varchar(10)  NOT NULL DEFAULT '0',
   fk_user_author                integer DEFAULT NULL,               -- user making creation
   fk_user_modif                 integer,                            -- user making last change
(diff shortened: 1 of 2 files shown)


htdocs/install/mysql/tables/llx_product.sql
```diff
@@ -51,14 +51,15 @@ create table llx_product
...
-  localtax2_tx                  double(6,3)  DEFAULT 0, 
+  localtax2_tx                  double(6,3)  DEFAULT 0,
   localtax2_type                varchar(10)  NOT NULL DEFAULT '0',
   fk_user_author                integer DEFAULT NULL,               -- user making creation
   fk_user_modif                 integer,                            -- user making last change
   tosell                        tinyint      DEFAULT 1,             -- Product you sell
   tobuy                         tinyint      DEFAULT 1,             -- Product you buy
   onportal                      tinyint      DEFAULT 0,	            -- If it is a product you sell and you want to sell it on portal (module website must be on)
   tobatch                       tinyint      DEFAULT 0 NOT NULL,    -- Is it a product that need a batch management (eat-by or lot management)
+  batch mask			        varchar(32),                        -- If the product has batch feature, you may want to use a batch mask per product
```
eldy, line 62: No spaces are allowed into a field name. Can you use
batch_mask instead of batch mask

altairis-tof: of course, it's a bad typing mismatch !

altairis-tof: @eldy : fixed with https://github.com/Dolibarr/dolibarr/pull/16996/commits/0adbffd8645457abc155169f9461f8a0b0cd36d1

PR #17005, opened 2021-03-29
Update llx_zapier_hook.key.sql

dolibit-ut:
- Part of /path
- Authors (Contributors/Editors)
- License

http://  ->  https://


eldy:
Copyright and licence must remain in same section in same block comment to be correctly grouped/expanded by ide.

PR #17011, opened 2021-03-29
NEW (re)generate PDF for notification emails (if not existing)

ptibogxiv:
new PR of  #16900 
with reload object for correct (refrence) directory for pdf storage and correct ref with substitution in templated mail

### DIFF
--- a/htdocs/core/class/notify.class.php
+++ b/htdocs/core/class/notify.class.php
@@ -561,7 +561,7 @@ class Notify
 								$link = '<a href="'.$urlwithroot.'/comm/action/card.php?id='.$object->id.'&entity='.$object->entity.'">'.$newref.'</a>';
 								$dir_output = $conf->agenda->dir_output;
 								$object_type = 'action';
-								$labeltouse = $conf->global->ACTION_CREATE_TEMPLATE;
+								//$labeltouse = $conf->global->ACTION_CREATE_TEMPLATE;
 								//$defaultmodelpdf = $conf->global->ACTION_ADDON_PDF;
 								$mesg = $outputlangs->transnoentitiesnoconv("EMailTextActionAdded", $link);
 								break;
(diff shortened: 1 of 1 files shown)


htdocs/core/class/notify.class.php
```diff
@@ -577,30 +577,58 @@ public function send($notifcode, $object, $filename_list = array(), $mimetype_li
...
+								$newlang = '';
+
+								if ($conf->global->MAIN_MULTILANGS && empty($newlang)) {
+										  $newlang = $object->thirdparty->default_lang;
+								}
+								if (!empty($newlang)) {
+									$outputlangs = new Translate("", $conf);
+									$outputlangs->setDefaultLang($newlang);
+								}
+								$modelpdf = !empty($object->modelpdf)?$object->modelpdf:$conf->global->FACTURE_ADDON_PDF;
```
eldy, line 591: Doc will not always be an invoice here. So use of $conf->global->FACTURE_ADDON_PDF must be done only if object->element is 'facture'
You must also deal with the case modelpdf is empty. You can just exclude generation of doc in such a case.

ptibogxiv: fix done

htdocs/core/class/notify.class.php
```diff
@@ -577,30 +588,58 @@ public function send($notifcode, $object, $filename_list = array(), $mimetype_li
 						$pdf_path = $dir_output."/".$ref.".pdf";
 						if (!dol_is_file($pdf_path)) {
 							// We can't add PDF as it is not generated yet.
-							$filepdf = '';
-						} else {
+							if (empty($conf->global->MAIN_DISABLE_PDF_AUTOUPDATE)) {
```
eldy, line 591: I would add a test on method_exists($object, 'generateDocument')  too here.

ptibogxiv: done !

PR #17015, opened 2021-03-30
FIX: Add a sub table to manage accountancy on product/company

aspangaro:


eldy:
Already fixed by another PR

PR #17050, opened 2021-03-30
Update subscription.php

dolibit-ut:
$_POST  ->  $GETPOST
lines 413 .. 417

// Clean some POST vars
		if (!$error) {
			$GETPOST("subscription",'alpha') = '';
			$GETPOST("accountid",'int') = '';
			$GETPOST("operation",'alphanohtml') = '';
			$GETPOST("label",'alpha') = '';
			$GETPOST("num_chq",'alphanohtml') = '';

### DIFF
--- a/htdocs/adherents/subscription.php
+++ b/htdocs/adherents/subscription.php
@@ -410,11 +410,11 @@ if ($user->rights->adherent->cotisation->creer && $action == 'subscription' && !
 
 		// Clean some POST vars
 		if (!$error) {
-			$GETPOST("subscription",'alpha') = '';
-			$GETPOST("accountid",'int') = '';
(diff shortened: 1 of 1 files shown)


htdocs/adherents/subscription.php
```diff
@@ -410,11 +410,11 @@
 
 		// Clean some POST vars
 		if (!$error) {
-			$_POST["subscription"] = '';
-			$_POST["accountid"] = '';
-			$_POST["operation"] = '';
-			$_POST["label"] = '';
-			$_POST["num_chq"] = '';
+			$GETPOST("subscription", 'alpha') = '';
```
eldy, line 413: For affectation, we must keep $_POST[...

PR #17065, opened 2021-04-01
Debug feature to import stock movement from csv file

Hystepik:
Debug of feature to import stock movement from csv file in massstockmove.php

### DIFF
--- a/htdocs/product/stock/massstockmove.php
+++ b/htdocs/product/stock/massstockmove.php
@@ -315,7 +315,7 @@ if ($action == 'importCSV' && !empty($user->rights->stock->mouvement->creer)) {
 		$importcsv->import_open_file($fullpath);
 		$labelsrecord = $importcsv->import_read_record();
 
-		if ($nblinesrecord <= 1) {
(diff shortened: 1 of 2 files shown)


htdocs/product/stock/massstockmove.php
```diff
@@ -329,28 +329,46 @@
 					continue;
 				}
 				//var_dump($data);
-
+				$formproduct = new FormProduct($db);
```
eldy, line 332: Each new allocate more memory. If you load 10000 line, you may have an out of memory.
Because we only need the instance to make a fetch (that erase old data), i think you can move the 4 new Xxx() here just before the while of the loop.

htdocs/product/stock/massstockmove.php
```diff
@@ -645,3 +678,30 @@
 // End of page
 llxFooter();
 $db->close();
+
+/**
+ * Verify if $haystack startswith $needle
+ * @param String $haystack string to test
```
eldy, line 684: Add one empty line after the description and before the first line of param.

PR #17071, opened 2021-04-01
fix: Right ID defenition in module builder

FHenry:
fix: Right ID defenition in module builder


FHenry:
@eldy I'm not agree with your changes..

With my fix (concat Module id . $r) :
Module id = 436301
Permission should be 4363011,4363012,....,43630110,43630111

With your fix (concat Module id + $r + 1) :
With your fix Perms ID it will be 436301+1+1=436303, 436301+2+1=436304
Let create another module with ID  436302
With your fix Perms ID 436302+1+1=436303 => Conflict with ID perms of first module...

And this fix must be in 13.0 and not in develop.

PR #17074, opened 2021-04-02
Fix 13 right definition id in modulebuilder

FHenry:
@eldy I'm not agree with your changes on #17071 

With my fix (concat Module id and $r) :
Module id = 436301
Permission ID will be 4363011,4363012,....,43630110,43630111

With your fix (addition Module id + $r + 1) :
With your fix Perms ID it will be 436301+1+1=436303, 436301+2+1=436304
Let create another module with ID 436302
With your fix Perms ID 436302+1+1=436303 => Conflict with ID perms of first module...

And this fix must be in 13.0 and not in develop.

### DIFF
--- a/htdocs/modulebuilder/template/core/modules/modMyModule.class.php
+++ b/htdocs/modulebuilder/template/core/modules/modMyModule.class.php
@@ -266,17 +266,17 @@ class modMyModule extends DolibarrModules
 		/* BEGIN MODULEBUILDER PERMISSIONS */
 		$this->rights[$r][0] = $this->numero . $r; // Permission id (must not be already used)
 		$this->rights[$r][1] = 'Read objects of MyModule'; // Permission label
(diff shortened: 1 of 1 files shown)


eldy:
The other modules (except first one) into core use another rule
With module id = 12300
then permission id are 12301, 12302, ...
So permissions id are in the range of module.
That's why each module id has an id 10 more than previous id. Not very good because this generated conflict.
If next module is 12310 and there is more than 10 permission in previous module => BUG

But your suggestion looks not perfect too.
If module is 436300 and permission is 1 with module 43630 and permission 10 we will have also a collision.
I suggest this instead:

$this->numero . sprintf("%02d", $r)

So each module can have 99 permission (and not only 10).
Because this is a change in architecture rules, this must be fixed into develop (we keep old rule for v13).

FHenry:
Close by #17195

PR #17095, opened 2021-04-02
Revert "Create index.html"

eldy:
Reverts Dolibarr/dolibarr#17088


eldy:
Dir scripts is not inside webroot so index.html is useless

PR #17097, opened 2021-04-03
Create Partnership Module

NextGestion:
Create Partnership Module


NextGestion:
Fix Errors

eldy:
Can you removed all files that we should not have into core and files with bad name like:
- files with a name with "--" into the name
- files with "object_" into name
- hidden files (.gitattributes, .editoconfig, .gitignore), we already have them into root of core.

eldy:
There is errors reported by stickler-ci (Click on "details" to see them).
For example, you must not use 
`var $var`
into class
but 
`public $var`
Thanks to fix

NextGestion:
Yes, we use "public $var".
**"stickler-ci"** detect a comment :

/**
	 * **@var string** ID to identify managed object.
	 */

NextGestion:
> There is errors reported by stickler-ci (Click on "details" to see them).
> For example, you must not use
> `var $var`
> into class
> but
> `public $var`
> Thanks to fix

Yes, we use "public $var".
"stickler-ci" detect a comment :

/**
* @var string ID to identify managed object.
*/

eldy:
> > There is errors reported by stickler-ci (Click on "details" to see them).
> > For example, you must not use
> > `var $var`
> > into class
> > but
> > `public $var`
> > Thanks to fix
> 
> Yes, we use "public $var".
> "stickler-ci" detect a comment :
> 
> /**
> 
> * @var string ID to identify managed object.
>   */

I'm afraid sticker has right.
I still has a lot of 
var $xxx
into code.
For example into class PartnershipUtils

Note that a PR can't be merged as long as there is error on CI checks.

htdocs/partnership/class/partnershiputils.class.php
```diff
@@ -0,0 +1,82 @@
...
+require_once DOL_DOCUMENT_ROOT.'/core/lib/files.lib.php';
+dol_include_once('partnership/lib/partnership.lib.php');
+
+
+/**
+ *	Class with cron tasks of Partnership module
+ */
+class PartnershipUtils
+{
+	var $db;							//!< To store db handler
```
eldy, line 37: Wrong declaration

NextGestion: Ok we fixed it.

Issue #17094, opened 2021-04-02
Bug attachment field does not appear in the emailing module


daraelmin:
# Bug
Bug: attachment field does not appear in the emailing module

## Environment
- **Version**: 13+
- **PHP**: 7.4
- **Database**: MySQL 5.7
- **URL(s)**: htdocs/comm/mailing/card.php?leftmenu=mailing&action=create

## Expected behavior
When creating a new emailing, the field attachement should appear to be able to attach file for the mailing.

## Atual behavior
The attachement field doesn't not appear when creating the emailing but when editing it with ckcreator it appears and is usable.
See below

### 1) create a new emailing (the field doesn't mot appear)
![IMG_20210402_221744](https://user-images.githubusercontent.com/8067223/113450981-4ea13d00-9401-11eb-80d5-ba3286d823d8.jpg)

### 2) edit with ckeditor (the field appear)
![IMG_20210402_221626](https://user-images.githubusercontent.com/8067223/113450997-57920e80-9401-11eb-88ae-f3965d77828b.jpg)

daraelmin:
This is not solved in 14.0

PR #17099, opened 2021-04-03
Use GETPOST instead of $_POST and use card.php instead of PHP_SELF

daraelmin:
# Fix use of $_POST and $_GET
# Fix use of PHP_SELF 
# Fix code style


daraelmin:
@eldy I wasn't sure for line 726 (old 801) and 737 (old 813) why the script is setting $GET['action'] so  I comment it just for memory in case of regression
I have cleaned the code cause I wanted to solve this #17094 and then I began to remove code stale errors.... So I'll fix the bug once the style is ok

PR #17101, opened 2021-04-03
Use GETPOST instead of $_POST and use card.php instead of PHP_SELF

daraelmin:
# Fix usage of constant
Fix use of $_POST and $_GET
Fix use of PHP_SELF

# Fix Code Style
Fix "{" on empty lines
Fix to much totally empty lines

@eldy 
I have cleaned the code cause I wanted to solve this #17094 and then I began to remove code stale errors.... So I'll fix the bug once the style is ok
I wasn't sure for line 730 (old 801) and 741 (old 813) why the script is setting $GET['action'] to "... .so I comment it just for memory in case of regression.


eldy:
Because the PR contains code style fixing and not just bug fix, can you push it on develop ?
For branch 13, only the minimal change to fix a bug must be pushed.

Issue #16628, opened 2021-03-12
Compatibility phpv8


eldy:
Install php v8, check that dolibarr gui is working correctly without warning. Fixing any reported error and warning by navigating in each screen.

ksar-ksar:
Hi,

Need to check on develop or V13 or both ?

eldy:
On develop only.

frederic34:
PHP Warning:  Undefined array key "dol_entity" in /home/fred/netlogic/dolibarr/htdocs/main.inc.php on line 788

frederic34:
PHP Warning: Undefined array key "dol_entity" in dolibarr/htdocs/main.inc.php on line 788
PHP Warning:  Undefined array key "dol_screenwidth" in dolibarr/htdocs/main.inc.php on line 1996
PHP Warning:  Undefined array key "dol_screenheight" in dolibarr/htdocs/main.inc.php on line 1996
PHP Warning:  Undefined array key "dol_dst" in dolibarr/htdocs/main.inc.php on line 1987
PHP Warning:  Undefined array key "dol_tz_string" in dolibarr/htdocs/main.inc.php on line 1989

frederic34:
PHP Fatal error: Uncaught TypeError: count(): Argument #1 ($var) must be of type Countable|array, string given in dolibarr/htdocs/core/lib/memory.lib.php:96\
https://github.com/Dolibarr/dolibarr/pull/16648

frederic34:
![image](https://user-images.githubusercontent.com/3624836/111078933-17252c00-84f8-11eb-9804-ef257fee081d.png)

PR #17103, opened 2021-04-03
Fix compatibility phpv8

TheGivrix:
# Fix #16628 *Compatibility phpv8*
*Fixed errors to get a first validation.*


htdocs/admin/company.php
```diff
@@ -666,7 +666,7 @@
 print "</tr>\n";
 
 print '<tr class="oddeven"><td><label for="SOCIETE_FISCAL_MONTH_START">'.$langs->trans("FiscalMonthStart").'</label></td><td>';
-print $formother->select_month($conf->global->SOCIETE_FISCAL_MONTH_START, 'SOCIETE_FISCAL_MONTH_START', 0, 1, 'maxwidth100').'</td></tr>';
+print (isset($conf->global->SOCIETE_FISCAL_MONTH_START) ? $formother->select_month($conf->global->SOCIETE_FISCAL_MONTH_START, 'SOCIETE_FISCAL_MONTH_START', 0, 1, 'maxwidth100') : '').'</td></tr>';
```
eldy, line 669: If $conf->global->SOCIETE_FISCAL_MONTH_START is not set, the method $formother->select_month to select the month will not be called at all.

htdocs/admin/delais.php
```diff
@@ -289,7 +289,7 @@
 print '<br>';
 
 
-if ($conf->global->MAIN_DISABLE_METEO != 1) {
+if (isset($conf->global->MAIN_DISABLE_METEO) && $conf->global->MAIN_DISABLE_METEO != 1) {
```
eldy, line 292: This test is not similar with previous one.
For example, if   $conf->global->MAIN_DISABLE_METEO  is not set, old test was true, new one is false

htdocs/admin/mails.php
```diff
@@ -48,7 +48,7 @@
 	'__DOL_MAIN_URL_ROOT__'=>DOL_MAIN_URL_ROOT,
 	'__ID__' => 'RecipientIdRecord',
 	//'__EMAIL__' => 'RecipientEMail',				// Done into actions_sendmails
-	'__CHECK_READ__' => (is_object($object) && is_object($object->thirdparty)) ? '<img src="'.DOL_MAIN_URL_ROOT.'/public/emailing/mailing-read.php?tag='.$object->thirdparty->tag.'&securitykey='.urlencode($conf->global->MAILING_EMAIL_UNSUBSCRIBE_KEY).'" width="1" height="1" style="width:1px;height:1px" border="0"/>' : '',
+	'__CHECK_READ__' => (isset($object->thirdparty) && is_object($object) && is_object($object->thirdparty)) ? '<img src="'.DOL_MAIN_URL_ROOT.'/public/emailing/mailing-read.php?tag='.$object->thirdparty->tag.'&securitykey='.urlencode($conf->global->MAILING_EMAIL_UNSUBSCRIBE_KEY).'" width="1" height="1" style="width:1px;height:1px" border="0"/>' : '',
```
eldy, line 51: If $object i snot an object then object->thirdparty does not exists and will return an error, this test must be after the test is_object($object)

htdocs/admin/mails.php
```diff
@@ -554,7 +554,7 @@ function initfields()
 	print '<tr class="liste_titre"><td class="titlefieldmiddle">'.$langs->trans("Parameter").'</td><td>'.$langs->trans("Value").'</td></tr>';
 
 	// Disable
-	print '<tr class="oddeven"><td>'.$langs->trans("MAIN_DISABLE_ALL_MAILS").'</td><td>'.yn($conf->global->MAIN_DISABLE_ALL_MAILS);
+	print '<tr class="oddeven"><td>'.$langs->trans("MAIN_DISABLE_ALL_MAILS").'</td><td>'.(isset($conf->global->MAIN_DISABLE_ALL_MAILS) ? yn($conf->global->MAIN_DISABLE_ALL_MAILS) : '');
```
eldy, line 557: Result is not similar

htdocs/admin/mails.php
```diff
@@ -737,26 +737,28 @@ function initfields()
...
-		} elseif ($conf->global->MAIN_MAIL_DEFAULT_FROMTYPE === 'company') {
-			print $langs->trans('CompanyEmail').' '.dol_escape_htmltag('<'.$mysoc->email.'>');
-		} else {
-			$id = preg_replace('/senderprofile_/', '', $conf->global->MAIN_MAIL_DEFAULT_FROMTYPE);
-			if ($id > 0) {
-				include_once DOL_DOCUMENT_ROOT.'/core/class/emailsenderprofile.class.php';
-				$emailsenderprofile = new EmailSenderProfile($db);
-				$emailsenderprofile->fetch($id);
-				print $emailsenderprofile->label.' '.dol_escape_htmltag('<'.$emailsenderprofile->email.'>');
+		if (isset($conf->global->MAIN_MAIL_DEFAULT_FROMTYPE)) {
```
eldy, line 740: This test is not similar, It will never reach the else.

htdocs/admin/mails.php
```diff
@@ -776,7 +778,7 @@ function initfields()
 		print '</td></tr>';
 
 		//Add user to select destinaries list
-		print '<tr class="oddeven"><td>'.$langs->trans("MAIN_MAIL_ENABLED_USER_DEST_SELECT").'</td><td>'.yn($conf->global->MAIN_MAIL_ENABLED_USER_DEST_SELECT).'</td></tr>';
+		print '<tr class="oddeven"><td>'.$langs->trans("MAIN_MAIL_ENABLED_USER_DEST_SELECT").'</td><td>'.(isset($conf->global->MAIN_MAIL_ENABLED_USER_DEST_SELECT) ? yn($conf->global->MAIN_MAIL_ENABLED_USER_DEST_SELECT) : '').'</td></tr>';
```
eldy, line 781: Result is not similar, the yn is not called in some cases.

htdocs/core/boxes/box_dolibarr_state_board.php
```diff
@@ -66,7 +66,7 @@ public function __construct($db, $param = '')
 			$this->enabled = 0; // disabled by this option
 		}
 
-		$this->hidden = !($user->rights->societe->lire && empty($user->socid));
+		$this->hidden = !(isset($user->rights->societe->lire) && empty($user->socid));
 	}
```
eldy, line 70: Result is not similar if $user->rights->societe->lire = 0

htdocs/core/lib/admin.lib.php
```diff
@@ -1885,14 +1885,14 @@ function email_admin_prepare_head()
 		$head[$h][2] = 'common';
 		$h++;
 
-		if ($conf->mailing->enabled) {
+		if (isset($conf->mailing->enabled)) {
```
eldy, line 1888: Result is not similar if $conf->mailing->enabled = 0

htdocs/core/modules/modDataPolicy.class.php
```diff
@@ -132,7 +132,7 @@ public function __construct($db)
 			array('DATAPOLICY_ADHERENT', 'chaine', '', $langs->trans('NUMBER_MONTH_BEFORE_DELETION'), 0),
 		);
 
-		$country = explode(":", $conf->global->MAIN_INFO_SOCIETE_COUNTRY);
+		if (isset($conf->global->MAIN_INFO_SOCIETE_COUNTRY)) $country = explode(":", $conf->global->MAIN_INFO_SOCIETE_COUNTRY);
```
eldy, line 135: if $conf->global->MAIN_INFO_SOCIETE_COUNTRY is not defined, the variable $country is undefined when it should be defined to empty array.

htdocs/admin/mails.php
Givriz: What is print after <td> in older version, when $conf->global->MAIN_DISABLE_ALL_MAILS is not set ?

htdocs/admin/mails.php
Givriz: How can I preg_replace if $conf->global->MAIN_MAIL_DEFAULT_FROMTYPE isn't set ?
I guess I have to initialise it, but how sould I do ?

htdocs/admin/mails.php
eldy: With old code, if  $conf->global->MAIN_DISABLE_ALL_MAILS  is not set, result is
print ...<td>.yn(unset value);

And the function yn is executed. With your change, the function is not executed.

htdocs/admin/mails.php
eldy: When a value is not set and PHP need to process it, PHP autoconvert it into 0 or ''
So here, if we want to change without knowing what the code do, we should replace the 
`if ($conf->global->MAIN_MAIL_DEFAULT_FROMTYPE === 'robot') {`
into
`if (isset($conf->global->MAIN_MAIL_DEFAULT_FROMTYPE) && $conf->global->MAIN_MAIL_DEFAULT_FROMTYPE === 'robot') {`

Givriz: Ok, got it for if condition. But then, we will reach the else. How preg_replace() will manage an unset value ?
Sould we do this :
preg_replace('/senderprofile_/', '', isset($conf->global->MAIN_MAIL_DEFAULT_FROMTYPE) ? $conf->global->MAIN_MAIL_DEFAULT_FROMTYPE : null);

htdocs/admin/mails.php
```diff
@@ -737,14 +737,14 @@ function initfields()
 
 		print '<tr class="oddeven"><td>'.$langs->trans('MAIN_MAIL_DEFAULT_FROMTYPE').'</td>';
 		print '<td>';
-		if ($conf->global->MAIN_MAIL_DEFAULT_FROMTYPE === 'robot') {
+		if (!empty($conf->global->MAIN_MAIL_DEFAULT_FROMTYPE) === 'robot') {
```
eldy, line 740: !empty($conf->global->MAIN_MAIL_DEFAULT_FROMTYPE) is a boolean. Comparing with 'robot' will always be false.

PR #17107, opened 2021-04-03
Create index.html

dolibit-ut:


eldy:
script dir is outside the web root so adding index.html is useless here.

PR #17110, opened 2021-04-04
FIX: Field amount is not deprecated

aspangaro:


htdocs/accountancy/class/bookkeeping.class.php
```diff
@@ -135,7 +135,6 @@ class BookKeeping extends CommonObject
 
 	/**
 	 * @var float FEC:Amount (Not necessary)
-	 * @deprecated No more used
```
eldy, line 138: The amount is same value than debit or credit right ?
What is rule of sign for this field ? For example for
Common invoice:    debit= ?   credit = ?    amount= ?
Credit note :          debit= ?   credit = ?    amount= ?

aspangaro: amount is abs() of debit/credit fields.
Sign is registered in 'sens' field with two possibilities (Sens of the account:  0: Debit, 1: Credit)

Technically, the field is not necessary (amount and sens (Direction)) but it is not depreciated because during the FEC export to ensure maximum compatibility, everyone displays it

eldy: So we don't need to store it. If field is required by the FEC format, we must just forge the value from the rule you describe to fill the FEC file.

PR #17111, opened 2021-04-04
Accountancy - Add a constant to enable fk_docdet register

aspangaro:
Required for some accounting-based modules


eldy:
Note that introducing this will generate a lot of trouble into reports and will corrupt database (by generating duplicate entries.
As long as it is hidden it is not a problem, but we should absolutely not enable it in production).

PR #17121, opened 2021-04-05
NEW Add support for Friday as a holiday

0x00FE:
# New [Add Support for Friday as a Holiday]
[Currently only Saturday and Sunday are supported for holiday calculation, this adds Friday as an option as well. The variables have to be added in Setup -> Other Setup "MAIN_NON_WORKING_DAYS_INCLUDE_FRIDAY", "MAIN_NON_WORKING_DAYS_INCLUDE_SATURDAY", "MAIN_NON_WORKING_DAYS_INCLUDE_SUNDAY" with 1 for yes and 0 for no. Currently Saturday and Sunday seems to be considered holidays by default.]

### DIFF
--- a/htdocs/core/lib/date.lib.php
+++ b/htdocs/core/lib/date.lib.php
@@ -687,8 +687,8 @@ function num_public_holiday($timestampStart, $timestampEnd, $country_code = '', 
 		$country_code = $mysoc->country_code;
 	}
 	if ($includefriday < 0) {
-                $includefriday = (isset($conf->global->MAIN_NON_WORKING_DAYS_INCLUDE_FRIDAY) ? $conf->global->MAIN_NON_WORKING_DAYS_INCLUDE_FRIDAY : 1);
-        }
+				$includefriday = (isset($conf->global->MAIN_NON_WORKING_DAYS_INCLUDE_FRIDAY) ? $conf->global->MAIN_NON_WORKING_DAYS_INCLUDE_FRIDAY : 1);
+	}
 	if ($includesaturday < 0) {
 		$includesaturday = (isset($conf->global->MAIN_NON_WORKING_DAYS_INCLUDE_SATURDAY) ? $conf->global->MAIN_NON_WORKING_DAYS_INCLUDE_SATURDAY : 1);
 	}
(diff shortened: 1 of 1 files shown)


0x00FE:
The modified function adds Friday as a parameter before the existing Saturday and Sunday variables to maintain days order.

`function num_public_holiday($timestampStart, $timestampEnd, $country_code = '', $lastday = 0, $includefriday = -1, $includesaturday = -1, $includesunday = -1)`

The 1st four variables seems to be required while the last two "$includesaturday" and "$includesunday" are optional and as per the code are only passed to the function during build tests in file "DateLibTest.php"

The continous-integration fails at line 238 of the file at the following test:

```
$result=num_public_holiday($date1, $date2, 'FR', 1, 1, 1);
$this->assertEquals(2, $result, 'NumPublicHoliday for FR');   // 1 opened day, 2 closed days
```

Here, with the new proposed function "$includefriday" is the first optional variable and in this assertion it will lead to 3 holidays since the default behavior is to have Saturday and Sunday as holidays as per code.


"$includefriday = -1" could be moved to the end of the function declaration to pass the build tests, but it will not maintain logical order of the days for the function if it is used later.
Since the function does not use any of those optional variables outside the scope of testing, I can't see this new proposal breaking any compatibility, the testing would need to be adjusted.

htdocs/core/lib/date.lib.php
```diff
@@ -658,20 +658,21 @@ function getGMTEasterDatetime($year)
...
+ *  @param	string			$country_code		Country code
+ *  @param	int			$lastday		Last day is included, 0: no, 1:yes
+ *  @param	int                     $includefriday		Include friday as non working day (-1=use setup, 0=no, 1=yes)
+ *  @param	int			$includesaturday	Include saturday as non working day (-1=use setup, 0=no, 1=yes)
+ *  @param	int			$includesunday		Include sunday as non working day (-1=use setup, 0=no, 1=yes)
+ *  @return   	int|string					Number of non working days or error message string if error
  *  @see num_between_day(), num_open_day()
  */
-function num_public_holiday($timestampStart, $timestampEnd, $country_code = '', $lastday = 0, $includesaturday = -1, $includesunday = -1)
+function num_public_holiday($timestampStart, $timestampEnd, $country_code = '', $lastday = 0, $includefriday = -1, $includesaturday = -1, $includesunday = -1)
```
eldy, line 675: friday is before saturday, however, to keep backward compatibility, when adding a new parameter, it must be added at end of function. Can you move it at end ?

PR #17139, opened 2021-04-06
FIX: Fix when add or modify ressources on events with duplicate entries

kkhelifa-opendsi:
Fix when add or modify ressources on events with duplicate entries

### DIFF
--- a/htdocs/comm/action/class/actioncomm.class.php
+++ b/htdocs/comm/action/class/actioncomm.class.php
@@ -590,9 +590,11 @@ class ActionComm extends CommonObject
 
 						$resql = $this->db->query($sql);
 						if (!$resql) {
-							$error++;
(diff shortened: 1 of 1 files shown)


htdocs/comm/action/class/actioncomm.class.php
```diff
@@ -590,9 +590,11 @@ public function create(User $user, $notrigger = 0)
 
 						$resql = $this->db->query($sql);
 						if (!$resql) {
-							$error++;
-							dol_syslog('Error to process userassigned: '.$this->db->lasterror(), LOG_ERR);
-							$this->errors[] = $this->db->lasterror();
+							if ($this->db->errno() != 'DB_ERROR_RECORD_ALREADY_EXISTS') {
```
eldy, line 593: We should process any error, even duplicate as an error, because with some databases like postgresql, it generates a rollback, even if we return no error.
So we must fix differently by avoiding to insert when it is already inserted.

PR #17141, opened 2021-04-06
FIX: Force event type to AC_EMAIL when action contain SENTBYMAIL

kkhelifa-opendsi:
Force event type to AC_EMAIL when 'action' contain SENTBYMAIL
and add new case of set 'societeforaction' when 'object->thirdparty' is defined

### DIFF
--- a/htdocs/core/triggers/interface_50_modAgenda_ActionsAuto.class.php
+++ b/htdocs/core/triggers/interface_50_modAgenda_ActionsAuto.class.php
@@ -942,6 +942,8 @@ class InterfaceActionsAuto extends DolibarrTriggers
 			$societeforaction->fetch($object->socid);
 		} elseif (isset($object->fk_soc) && $object->fk_soc > 0) {
 			$societeforaction->fetch($object->fk_soc);
+		} elseif (isset($object->thirdparty->id) && $object->thirdparty->id > 0) {
+			$societeforaction = $object->thirdparty;
 		}
 
 		$projectid = isset($object->fk_project) ? $object->fk_project : 0;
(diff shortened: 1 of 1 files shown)


htdocs/core/triggers/interface_50_modAgenda_ActionsAuto.class.php
```diff
@@ -942,6 +942,8 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
 			$societeforaction->fetch($object->socid);
 		} elseif (isset($object->fk_soc) && $object->fk_soc > 0) {
 			$societeforaction->fetch($object->fk_soc);
+		} elseif (isset($object->thirdparty->id) && $object->thirdparty->id > 0) {
```
eldy, line 945: If $object->thirdparty is not defined, it will crash on php8. Can you also add a test
if (isset($object->thirdparty->id) ?

htdocs/core/triggers/interface_50_modAgenda_ActionsAuto.class.php
```diff
@@ -958,11 +960,14 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
...
 
+		$type_code = '';
+		if (strpos($action, 'SENTBYMAIL')) $type_code = 'AC_EMAIL_OUT';
+
 		// Insertion action
 		require_once DOL_DOCUMENT_ROOT.'/comm/action/class/actioncomm.class.php';
 		$actioncomm = new ActionComm($this->db);
-		$actioncomm->type_code   = $object->actiontypecode; // Type of event ('AC_OTH', 'AC_OTH_AUTO', 'AC_XXX'...)
-		$actioncomm->code        = 'AC_'.$action;
+		$actioncomm->type_code   = !empty($type_code) ? $type_code : $object->actiontypecode; // Type of event ('AC_OTH', 'AC_OTH_AUTO', 'AC_XXX'...)
```
eldy, line 969: The caller should be able to define the value by setting $object->actiontypecode
So this value must used in priority and only after type_code can be used.
Which value do you have for $object->actiontypecode when action is *SENTBYMAIL* ?

htdocs/core/triggers/interface_50_modAgenda_ActionsAuto.class.php
```diff
@@ -94,7 +94,9 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
 
 		$langs->load("agenda");
 
-		if (empty($object->actiontypecode)) {
+		if (strpos($action, 'SENTBYMAIL') && (empty($object->actiontypecode) || $object->actiontypecode == 'AC_OTH_AUTO')) {
+			$object->actiontypecode = 'AC_EMAIL';
```
eldy, line 98: This generates a problem.
It break the rule:
If action is an automatic action, we use a typecode that has a type 'systemauto', if it is a manual action we use a typecode that has a type 'system'. Here we force to AC_EMAIL so manual action whatever is nature of event, so even for automatic action.

What value to you have for $action and in for which situation do you need to fix this ?

eldy: As discussed during devcamp, i cancel the PR.

Issue #16167, opened 2021-02-04
Mass action propal accept and workflo


BB2A-Anthony:
# Bug
If execute mass action accept propal the and workflo is activate.

The workflo is not exécute and no event add

PR #17143, opened 2021-04-06
Workflo mass action propal sign or NotSigned and notification

BB2A-Anthony:
# Fix #[*#16167 Mass action propal accept and workflo*]

Resolved the problems to triggers and notifications

# Close #[#16167]


BB2A-Anthony:
@eldy PR not resolv the problème?

eldy:
Thanks for pointing to correction. Due to another PR that rename the method, i propagate the fix manually.
Can you check it is fixed on your side with the other commits merged ?

BB2A-Anthony:
> Thanks for pointing to correction. Due to another PR that rename the method, i propagate the fix manually.
> Can you check it is fixed on your side with the other commits merged ?

@eldy #17143

PR #17148, opened 2021-04-07
Fix empty string WEBSITE_OTHERLANG (better fix)

indelog:
This is better fix for my previous pr : #17144 whit better description.

If post WEBSITE_OTHERLANG with empty value or with value like this
'en,,sv' or 'en,sv,' we can get an otherlang with empty value. This can lead
unexpected behaviour when we use an element of otherlang to set the
subdirectory name to store pages for a sub lang.

### DIFF
--- a/htdocs/core/lib/website2.lib.php
+++ b/htdocs/core/lib/website2.lib.php
@@ -104,10 +104,12 @@ function dolSavePageAlias($filealias, $object, $objectpage)
 	}
 	// Save also alias into all language subdirectories if it is a main language
 	elseif (empty($objectpage->lang) || !in_array($objectpage->lang, explode(',', $object->otherlang))) {
-		if (empty($conf->global->WEBSITE_DISABLE_MAIN_LANGUAGE_INTO_LANGSUBDIR) && !empty($object->otherlang)) {
+		if (empty($conf->global->WEBSITE_DISABLE_MAIN_LANGUAGE_INTO_LANGSUBDIR)) {
 			$dirname = dirname($filealias);
 			$filename = basename($filealias);
 			foreach (explode(',', $object->otherlang) as $sublang) {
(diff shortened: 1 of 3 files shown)


indelog:
Not merge it now. I'll push a fix.

htdocs/core/lib/website2.lib.php
```diff
@@ -104,10 +104,12 @@ function dolSavePageAlias($filealias, $object, $objectpage)
 	}
 	// Save also alias into all language subdirectories if it is a main language
 	elseif (empty($objectpage->lang) || !in_array($objectpage->lang, explode(',', $object->otherlang))) {
-		if (empty($conf->global->WEBSITE_DISABLE_MAIN_LANGUAGE_INTO_LANGSUBDIR) && !empty($object->otherlang)) {
+		if (empty($conf->global->WEBSITE_DISABLE_MAIN_LANGUAGE_INTO_LANGSUBDIR)) {
```
eldy, line 107: It is strange to remove the test on "&& !empty($object->otherlang)"
If $object->otherlang is empty, then we won't execute the loop late based on $object->otherlang (it means we have nothing to do)
I think you can keep this test.

indelog: I removed it because finally test if `$object->otherlang` is empty is not a reliable mean to avoid to have an empty `$subdir` in the loop bellow. We can have `$object->otherlang` with value like this : `es,,en` or `es,en,` (may be an user input error), if is the case the array returned by `explode(',', $object->otherlang)` will result get an element with an empty string which will result to get a `$subdir` with empty string in the for loop even if `$object->otherlang` is not empty. I figured that testing if `$object->otherlang` is empty will give the a false security idea so i removed it. If you think is better to keep it i will re-add it.

PR #17154, opened 2021-04-07
NEW : Search contact in select or multiselect with AJAX support

kkhelifa-opendsi:
Search contact in select or multiselect with AJAX support


htdocs/core/class/html.form.class.php
```diff
@@ -1572,126 +1572,262 @@ public function select_remises($selected, $htmlname, $filter, $socid, $maxvalue
...
+	 * @param	array	$ajaxparameters			Additionnals parameters for ajax request in ajax_combobox()
+	 * @param	array	$ajaxoptions			Options for ajax_autocompleter // Not used for this moment todo add ajaxoptions to ajax_combobox()
+	 * @return	int								<0 if KO, Nb of contact in list if OK
+	 * @deprected								You can use selectcontacts directly (warning order of param was changed)
 	 */
-	public function select_contacts($socid, $selected = '', $htmlname = 'contactid', $showempty = 0, $exclude = '', $limitto = '', $showfunction = 0, $moreclass = '', $showsoc = 0, $forcecombo = 0, $events = array(), $options_only = false, $moreparam = '', $htmlid = '')
+	public function select_contacts($socid, $selected = '', $htmlname = 'contactid', $showempty = 0, $exclude = '', $limitto = '', $showfunction = 0, $moreclass = '', $showsoc = 0, $forcecombo = 0, $events = array(), $options_only = false, $moreparam = '', $htmlid = '', $multiple = false, $disableifempty = 0, $new_ajax = false, $thirdparty_htmlname = '', $limit = 20, $hidelabel = 1, $ajaxurl = DOL_URL_ROOT . '/contact/ajax/contact.php', $ajaxparameters = array(), $ajaxoptions = array())
```
eldy, line 1601: Is having a dynamic default value based on a runtime var as a default paramater in declaration working on all cases/PHP ? (here $ajaxurl = DOL_URL_ROOT...)
May be you can set default value to '' and into function set it to DOL_URL_ROOT ... if '' ?
if may work, but it looks strange to me.

daraelmin: From the manual, since PHP 5.6:

"Example 6 Using non-scalar types as default values (...)
The default value must be a constant expression, not (for example) a variable, a class member or a function call."

So, it seems correct to me to use constant as default value in function.

https://www.php.net/manual/en/functions.arguments.php#functions.arguments.default

daraelmin: But that was not always true in PHP 5.4

eldy: However, this make still code declaration not easy to understand. this may also obfuscate bugs on some Continuous integration like scrutinizer.
Can you replace with $ajaxurl = '' and in code of function make
```
if (empty($ajaxurl)) {
   $ajaxurl = DOL_URL_ROOT . '/contact/ajax/contact.php';
}
```

htdocs/contact/ajax/contact.php
```diff
@@ -0,0 +1,113 @@
...
+$id=GETPOST('id', 'int');
+$socid=GETPOST('socid', 'int');
+$showempty=GETPOST('showempty', 'int');
+$showfunction=GETPOST('showfunction', 'int');
+$showsoc=GETPOST('showsoc', 'int');
+$exclude=GETPOST('exclude', 'array');
+$page=GETPOST('page', 'int');
+$limit=GETPOST('limit', 'int');
+
+
```
eldy, line 46: It seems anybody even without permission can scan and download all the contact database with this new file.
Can you 
- force $socid if user->socid is set (for external users)
- test that user as permission to read thirdparties
- Also if the user has not the permission "Read every thirdparty", a join must be added into the new selectcontactlist to limit the select to contact linked to thirdparty that as the user as sale representative so he can get only list of contact of thirdparties he is allowed to read.

eldy:

PR #17162, opened 2021-04-07
13.0-fix16323

### LINKED ISSUE #16323: {company_name} in ODT not filled
# Bug
In the proposal ODT gneration the fild {company_name} will be empty

## Environment
- **Version**: [*Affected Dolibarr version(s)*] 13.0.0
- **OS**: [*Server OS type and version*] Ubuntu 20
- **Web server**: [*Webserver type and version*] apache

glu000:
# Fix #16323
[*Long description*]

This is a new bug in V13 in php-scripts htdocs/core/modules/xxx/doc/doc_generic_xxx_odt.modules.php

The following line makes no sense, because !isset OR !empty is always true:

if ($usecontact && ($object->contact->fk_soc != $object->thirdparty->id && (!isset($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT) || !empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT))))

Also the following line makes no sense

$socobject = $object->contact;

because $socobject should be a Societe object and not contact

### DIFF
--- a/htdocs/core/modules/commande/doc/doc_generic_order_odt.modules.php
+++ b/htdocs/core/modules/commande/doc/doc_generic_order_odt.modules.php
@@ -305,8 +305,10 @@ class doc_generic_order_odt extends ModelePDFCommandes
 				$contactobject = null;
 				if (!empty($usecontact))
 				{
-					if ($usecontact && ($object->contact->fk_soc != $object->thirdparty->id && (!isset($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT) || !empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT)))) {
-						$socobject = $object->contact;
+					if ($usecontact && ($object->contact->fk_soc != $object->thirdparty->id && (!empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT)))) {
+						$object->contact->fetch_thirdparty();
+						$socobject = $object->contact->thirdparty;
+						$contactobject = $object->contact;
 					} else {
 						$socobject = $object->thirdparty;
 						// if we have a CUSTOMER contact and we dont use as recipient we store the contact object for later use
--- a/htdocs/core/modules/commande/doc/doc_generic_order_odt.modules.php
+++ b/htdocs/core/modules/commande/doc/doc_generic_order_odt.modules.php
@@ -305,7 +305,7 @@ class doc_generic_order_odt extends ModelePDFCommandes
 				$contactobject = null;
 				if (!empty($usecontact))
 				{
-					if ($usecontact && ($object->contact->fk_soc != $object->thirdparty->id && (!empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT)))) {
+					if ($usecontact && ($object->contact->fk_soc != $object->thirdparty->id && (isset($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT) && !empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT)))) {
 						$object->contact->fetch_thirdparty();
 						$socobject = $object->contact->thirdparty;
 						$contactobject = $object->contact;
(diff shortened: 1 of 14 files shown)


htdocs/core/modules/commande/doc/doc_generic_order_odt.modules.php
```diff
@@ -305,8 +305,10 @@ public function write_file($object, $outputlangs, $srctemplatepath, $hidedetails
 				$contactobject = null;
 				if (!empty($usecontact))
 				{
-					if ($usecontact && ($object->contact->fk_soc != $object->thirdparty->id && (!isset($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT) || !empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT)))) {
-						$socobject = $object->contact;
+					if ($usecontact && ($object->contact->fk_soc != $object->thirdparty->id && (!empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT)))) {
```
eldy, line 308: Th isset was to avoid warning with php8 when param is not set

htdocs/core/modules/commande/doc/doc_generic_order_odt.modules.php
```diff
@@ -305,8 +305,10 @@ public function write_file($object, $outputlangs, $srctemplatepath, $hidedetails
...
+						$object->contact->fetch_thirdparty();
```
eldy, line 309: Goal of option MAIN_USE_COMPANY_NAME_OF_CONTACT is to replace the company information with the contact.
So we must keep 
$socobject = $object->contact;
The object of company is replaced with the object of contact.

The behaviour you introduce is already managed by the "else".

glu000: I fully understand the purpose of MAIN_USE_COMPANY_NAME_OF_CONTACT but the current implementation just does not work, because later the function function get_substitutionarray_thirdparty is called with this object and this function expects a thirdparty object (and not a contact object). I've tested it and with my implementation it works in both cases (MAIN_USE_COMPANY_NAME_OF_CONTACT is set or not set)

htdocs/core/modules/commande/doc/doc_generic_order_odt.modules.php
glu000: I see. But it makes no sense to me. Because if the param is not set then !isset becomes true

If the param is set, then !empty becomes true.

So the expression is always true

glu000: ... so probably the solution would be

(isset($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT) && !empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT))

eldy: Yes, correct code should be 
`(isset($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT) && !empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT))`

glu000: PR is updated

htdocs/core/modules/commande/doc/doc_generic_order_odt.modules.php
eldy: Sorry i gave you bad information. Test should be
`if ($usecontact && ($object->contact->fk_soc != $object->thirdparty->id && (!isset($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT) && !empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT)))) {`
So like it is in old version.
What we want is to have MAIN_USE_COMPANY_NAME_OF_CONTACT the default value.
so if unset, we must act as if it is set to MAIN_USE_COMPANY_NAME_OF_CONTACT=1
Only case MAIN_USE_COMPANY_NAME_OF_CONTACT set to 0 will switch in the else.
Really Sorry. Can you fix my error too ?

eldy: Still trouble with the test 
`if ($usecontact && ($object->contact->fk_soc != $object->thirdparty->id && (isset($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT) && !empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT)))) {`
Old code looks good.

PR #17166, opened 2021-04-07
Product masks

altairis-noe:
# New option in product_batch core module to have custom masks by product
masks


altairis-tof:
@eldy : @altairis-noe starts to work on batch numbering mask management per product. others PR to come very soon.

eldy:
Welcome @altairis-noe 
You will see i am the nightmare of developers !  But once PR is validated, it's a dream.

I would like to understand this new option. On you screen shot it appears, you can have both lot_advanced and the new optoin selected.
What if you have lot_standard and the new option ?
My understanding is that this new option has a sense only if we choose "advanced". 
if i am right, the option should not be a fourth selector but a selector inside the "lot_advanced" area, just under or before the button "Modify". 
So the code to manage this switch should be into the file of the lot_advanced selector file and not into the setup page (if we remove the lot_standard file, the line of numbering rule lot_advanced must disappear but also this new option.

altairis-tof:
Comment deleted

altairis-tof:
After second thought, we are ok with what you said.
@altairis-noe will update the PR.

altairis-noe:
Hi there, and thanks @eldy for the welcome !
I adjusted so the product mask option is available only if the _advanced modes is active. I hope you’ll find it to your liking.
masks

lmag:
On détecte une once d'humanité chez @eldy "Welcome @altairis-noe
You will see i am the nightmare of developers ! But once PR is validated, it's a dream." :)
La PR est testé ce jour c'est ok pour moi en test

eldy:
Can you confirm this ?
"If the option is set, when we will generate a lot nmber, we will not use the mask defined into the setup page but the mask defined into product page in priority".

If i am right, having the option visible only if "lot_advanced" is enabled is not enough.
It still means the code related to setup the numbering handler is outside the handler file.
Code to select option should be moved inside the lot_advanced handler and should appear on the same line than lot_advanced line. The select box should be in the area where there is the mask and save button and not into column "Status" of numbering module.

altairis-noe:
Hello again,
I will check with @altairis-tof and relocate the option, if you think that’s how it should be done. I am not sure how you see the “select box”. The idea is that if we have selected lot_advanced, we can add this option so that each product can have its own mask or use the one defined as lot_advanced. I don’t know if I am crystal clear…

lmag:
"If the option is set, when we will generate a lot nmber, we will not use the mask defined into the setup page but the mask defined into product page in priority".

* Yes

"If i am right, having the option visible only if "lot_advanced" is enabled is not enough."

* We talked about adding a tooltip to define where it is managed

"It still means the code related to setup the numbering handler is outside the handler file.
Code to select option should be moved inside the lot_advanced handler and should appear on the same line than lot_advanced line. The select box should be in the area where there is the mask and save button and not into column "Status" of numbering module."

* Knowing that the management of batch and serial numbers is specific to each product. We put the box at the level of each product.

image

altairis-noe:
So I updated the PR with checkboxes relocated in the mod files. When we check the option, if necessary, the _advanced mode is automatically selected. Tell me if something doesn’t add up
Screenshot_2021-04-12 Configuration du module lot série + multicompany

eldy:
👍🏻 Perfect !

PR #17171, opened 2021-04-07
Create index.html

dolibit-ut:


eldy:
Not sure it is a good idea to add index.html everywhere into.
Why not remove them everywhere ?
Why keeping them ? It is not a problem to see the content of a dir. This information is already public on github !

daraelmin:
It could be more efficient to use htaccess 'deny from all' command

dolibit-ut:
> Not sure it is a good idea to add index.html everywhere into.
> Why not remove them everywhere ?
> Why keeping them ? It is not a problem to see the content of a dir. This information is already public on github !


I think it makes a difference to read the Dolibarr dirs/files from Github
or to scan the files on a real productive webserver.

Not all hosts use a /dolibarr/ path - 
so you dont know what you are scanning until you see the files.

And maybe not everyone has an up-to-date repository 
so it can be useful to hide this.

A htaccess is possibly a good alternative.

eldy:
A htaccess is surely a better alternative.

PR #17182, opened 2021-04-08
add serials separator configuration

altairis-tof:
add a config option to define separator that will be used later in multple serials input options (cf old PR https://github.com/Dolibarr/dolibarr/pull/16975)
Configuration_du_module_lot_série_+_multicompany_-_2021-04-08_10 33 24

### DIFF
--- a/htdocs/core/modules/modProductBatch.class.php
+++ b/htdocs/core/modules/modProductBatch.class.php
@@ -77,7 +77,6 @@ class modProductBatch extends DolibarrModules
 		$this->langfiles = array("productbatch");
 
 		// Constants
-		// Constants
 		$this->const = array();
 		$r = 0;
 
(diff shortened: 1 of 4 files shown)


eldy:
I don't think there is a , or ; or space as part of a serial number.
So i don't think we need a setup here.
Why not just consider any char  [,;\s] as separator for list of serial ?

altairis-tof:
how do you explode a string without specifying a specific separator ?

eldy:
`$myarray  =   preg_split('/[,;\s]+/',  $thestringtoexplode);`

altairis-tof:
ok for me ! thanks for the tips. 
i close the PR.

BebZ:
how do you explode or "consume" quantities on each batch ? 
if you enter in dispatch 3 batch numbers :  
AAA;BBB;CCC and a qty of 30 
it will add 10AAA 10BBB and 10CCC  
or 30 of each in stock ?
 or do you plan a verification if there is a batch separator, so there must be as many batch numbers as qties ?
so for qty 30 you will have (mandatory) to input 30 batch numbers ? : 
AAA;BBB;CCC...AAE ?

Thanks

lmag:
With this ultime solution it will work

LOT1;QTY1;WAREHOUSEID1
LOT2;QTY2;WAREHOUSEID2
LOT3;QTY3;WAREHOUSEIDX

PR #17184, opened 2021-04-08
Manage stock movement for unique serial numbers

altairis-tof:
add unicity check on serial number input and check qty not greater than batch in stock before remove.
(from older PR #16975)


htdocs/product/stock/class/mouvementstock.class.php
```diff
@@ -256,6 +256,36 @@ public function _create($user, $fk_product, $entrepot_id, $qty, $type, $price =
 				return -2;
 			}
 
+
+			// check unicity for input serial numbered equipments ( different for lots managed products)
+			if ( $product->status_batch == 2 && ($type == 0 || $type == 3) )
```
eldy, line 261: Why moving this test at beginning.
Testing at end of code (when every action were done, like triggers for example) is more secured.

altairis-tof: it seemed to me it is a gain of performance to  first check the passed parameter before doing other checks into database.

altairis-tof: @eldy what do you think of my argument for performance ? is it ok for you or do i have to put the code below ?

eldy: Performance is important but it will lead to a lot of other problem. But integrity too.
And checking integrity of data at begin of transaction means all the rest of transaction can break integrity.
And integrity has priority versus performance. So test at end (where it is) must remain.

altairis-tof:
@eldy it should be ok now; i moved the SN unicity check at the good place, before the qty check for negative transfer (and i removed my own qty check for that case that was thus useless !).

altairis-tof:
I note some difficulties with the "batch" word, especially now that we have to differentiate lot managed products and SN managed products... maybe a more generic word like "equipment" would do the job ?

also some interogations subsists around the product_lot and product_batch tables; i think it should be more clear if they were called product_lot and product_lot_movements for instance.
and i do not understand the eatby and sellby presence in the product_batch table since it is already defined in product_lot.

BebZ:
Please attention to backward / future compatibility... 
Modifying table names with data inside is a nightmare for products in stock...
I also notice you want to modify "batch" or other important words that are use in many variables, no ?

So it's like rewriting all the module, in next version, it won't be usable by people having it installed in v10...1...12...13 ?
And how can we answer their future questions...

Also Eldy has the "manufacturing" module based on "batch" and "product_lot" "product_batch" tables I guess, or is it planned to populate these tables or future different ones ?

Thanks

altairis-tof:
@BebZ i totally agree with all your argument; i just note that the words are source of confusion; maybe we can let the database as is and just be more meanigful on screens ?

altairis-tof:
@eldy c'est tout bon maintenant !

lmag:
@eldy that ok for me

htdocs/product/stock/class/mouvementstock.class.php
```diff
@@ -348,7 +357,7 @@ public function _create($user, $fk_product, $entrepot_id, $qty, $type, $price =
 						$i++;
 					}
 				}
-				else   // If not found, we add record
+				elseif ($type == 0 || $type == 3)   // If not found, we add record
```
eldy, line 360: This break the autorepair feature.
When we have in stock a record with batch number X, when we change stock of X (and we must be able since it is into database), we refuse because the record into Productlot is not correctly synchronized.
We have no reason to add a condition here. If a record is in stock with a lot X, we must create to fix consistency.
This will generate a dead lock situation.

lmag: @altairis-tof

altairis-tof: @lmag i saw it but did not understand what to do...
@eldy do i have to just remove this code or do it another way ? i'm pulling my hair out ;-)

altairis-tof: @eldy ok i have removed the test. you can merge (i hope)

htdocs/product/stock/class/mouvementstock.class.php
```diff
@@ -272,6 +272,15 @@ public function _create($user, $fk_product, $entrepot_id, $qty, $type, $price =
 				$i = 0;
 				if ($num > 0)
 				{
+					if ( $product->status_batch == 2 && ($type == 0 || $type == 3) ) {
```
eldy, line 275: The table product_lot is a memory table to store properties of a lot. Not to store we have this lot in stock or not.
Having a record into this table should not block any stock movement on this lot.
We may have a record into this table (as a memory because lot did exists in the past) and have a quantity in stock of 0. So trying to create a stock movement of +1 is legitimate and should not be blocked.
Only test on final quantity to be sure we have 0 or 1 at end of transaction should block.

htdocs/product/stock/class/mouvementstock.class.php
```diff
@@ -379,6 +388,17 @@ public function _create($user, $fk_product, $entrepot_id, $qty, $type, $price =
 		$movestock = 0;
 		if ($product->type != Product::TYPE_SERVICE || !empty($conf->global->STOCK_SUPPORTS_SERVICES)) $movestock = 1;
 
+		// check unicity for serial numbered equipments ( different for lots managed products)
```
eldy, line 391: If we have a corruption of data and qty in stock is 10 and we want to fix to get 1 (because it is a unique serial) for example after a direct import or because we changed the status of batch from 1 to 2 later, this test will make not possible to fix situation.
Must be removed.

htdocs/product/stock/class/mouvementstock.class.php
```diff
@@ -568,32 +588,14 @@ public function _create($user, $fk_product, $entrepot_id, $qty, $type, $price =
 			// Update detail stock for batch product
 			if (!$error && !empty($conf->productbatch->enabled) && $product->hasbatch() && !$skip_batch)
 			{
-				// check unicity for serial numbered equipments ( different for lots managed products)
-				if ( $product->status_batch == 2 && $qty > 0 )
-				{
-					if ( $this->getBatchCount($fk_product, $batch) > 0 )
```
eldy, line 574: Why removing this test. It is the only test that gurarantee unicity into database.
PS: This test should be moved ate very end of transaction, so at line 642 (after the triggers).

htdocs/product/stock/class/mouvementstock.class.php
altairis-tof: i agree

altairis-tof: but where do you check the serail number is not already used for the product ?

htdocs/product/stock/class/mouvementstock.class.php
altairis-tof: ok i do it as for the previous, but we will have to have those tests elsewhere for dispatch, stock transfer or correction, bom, apis...

htdocs/product/stock/class/mouvementstock.class.php
altairis-tof: @eldy ok for me, latest commit should solve all your points

altairis-tof:
@eldy hello, i think red tag "blocking" is no more relevant; otherwise tell me what to do more please.

lmag:
Is everything ok for you @eldy

eldy:
I have made another PR to set at the good place the test on  
if ( $product->status_batch == 2) { this->getBatchCount(
So i think we can delete this PR

altairis-tof:
ok thanks

PR #17189, opened 2021-04-08
add fk_project to warehouse table

altairis-tof:
The aim of this PR is to be able to link one or several warehouses to a project.
The objective is to trace products and especially equipments in order to know for example where are the equipments with a specific lot number, but also trace the life of the equipment (in stock, at the customer, in reparation, ...)
@lmag


lmag:
@altairis-tof  that ok for me

PR #17204, opened 2021-04-09
Close #17193 : Move button added with PR 17161

### LINKED ISSUE #17193: Move button added with PR #17161
Complete PR #17161 
Move the button added on the website module menu to generate .webp images into the page of module GED for manual directories, just after the picto + (new directory) and refresh on this screenshot.
image

Having the feature here will allow to use it also in the GED, and because this screen is also the screen used for the "media" button into website module, it will be automatically available for a website too.
Also it looks more natural to have this feature into the image media manager since its a a feature that interact on images.

Hystepik:
Move on button to change images in webp in filemanager.tpl.php


eldy:
Are you sure the transformation is done on "current selected dir" only ?
You probably need to add some js script to process only the current dir. To be sure, you should show value of relative directory into the confirm box.
Also, when we are into a subdir and we click on button, we should remains on same open dir instead of going back to root.

PR #17206, opened 2021-04-09
Fix use json when storing array constant

daraelmin:
Fix use json  when storing array constant
When setting an array as a constant, it should store it using JSON encode / decode to be sure.


eldy:
I planned to depreciate completely the type to store constant so only string must be stored. So it is role of the function that call the method to encode/decode using the encoding/decoding of its choice to store the data.

Issue #17192, opened 2021-04-08
Confusing Bookmarks in Calendar - Possible error


victorcardim:
# Bug
[*Confusing Bookmarks in Calendar *]

## Environment
- **Version**: [*13.0.2*]
- **OS**: [*Win 10 Pro 64bits*]
- **Web server**: [*Apache 2.4.46*]
- **PHP**: [*7.4.16*]
- **Database**: [*MariaDB 10.5.8*]
- **URL(s)**: [*/dolibarr/comm/action/*]

## Expected and actual behavior
[*
**Possible problem encountered:**

When performing day to day routines in Dolibarr (ex: Create user, make budget, create order), I noticed that Dolibarr, despite correctly registering the date and time in events, its schedule shows a “delay” of one day, in the my case.

**Reporting in another way:**

Today, 06/04/2021 at 08:42 AM, Brazilian time, I created a new third party (client) and when consulting the agenda the event was automatically inserted on 04/05/2021. Not only the creation of a new third party but everything that was done in the system today (creating budgets, orders, among others).*]

## [Attached files](https://help.github.com/articles/issue-attachments) (Screenshots, screencasts, dolibarr.log, debugging informations…)
[*
![LocationParametersDolibarr](https://user-images.githubusercontent.com/82184733/114105363-fbb50300-98a2-11eb-9b3b-f01b1222f496.jpg)
![ViewMoreInfoUserDolibarr](https://user-images.githubusercontent.com/82184733/114105364-fce63000-98a2-11eb-808d-909b07ad54e3.jpg)
*]

PS: This matter was discussed at the international Dolibarr Forum.

For
follow the speech visit the link

[https://www.dolibarr.org/forum/t/confusing-bookmarks-in-calendar/20517](url)

PR #17211, opened 2021-04-10
Fix: #16452 Dates in perweek.ph are shifted by one

### LINKED ISSUE #16452: Per week activity displays holidays on the wrong date.
# Bug
Using a timezone other than GMT breaks holidays on `/projet/activity/perweek.php`. The holidays will be off by one day. In our example, using the timezone 'Europe/Berlin` yields the software calculating independence day on July 5th.

## Environment
- **Version**: 13.0.0
- **OS**: CentOS 7.0
- **Web server**: NGinx
- **PHP**: 7.3.21
- **Database**: MariaDB 10.4.6
- **URL(s)**: `/projet/activity/perweek.php`

When setting the server timezone to Europe/Berlin creating independence day on July 4th, the overview should indicate that July 4th is a holiday. This works for Jan, 6th and any other date we tested.

The system displays July 5th

Holidays are all offset in our experience.

## Our solution
I am not submitting a pull, since I do not know whether this is the right way to fix this, but replacing the line 99 in `perweek.php` with 

```
$firstdaytoshow = dol_mktime(0, 0, 0, $first_month, $first_day, $first_year, true);
```

Would yield the system behaving properly. So I assume there's an issue with the fact that `dol_mktime()` uses server timezone by default but the block to calculate whether the day is a holiday does not.

Csardelacal:
# Fix https://github.com/Dolibarr/dolibarr/issues/16452

When opening `projet/activity/perweek.php?leftmenu=tasks` dates
are shifted by 1 day to the future / past depending on the timezone of
the machine hosting the code.

When changing the calculation to UTC times, the issue goes away.

### DIFF
--- a/htdocs/core/lib/date.lib.php
+++ b/htdocs/core/lib/date.lib.php
@@ -680,7 +680,7 @@ function num_public_holiday($timestampStart, $timestampEnd, $country_code = '', 
 	/**
 	 * This bit checks that the date we received is actually a GMT date. It doesn't
 	 * matter what date we're checking against, since the UTC clock does not have
-	 * daylight savings, but it's important that the timestamp represents a UTC 
+	 * daylight savings, but it's important that the timestamp represents a UTC
 	 * midnight time.
 	 */
 	if (((dol_mktime(0, 0, 0, 10, 10, 2020, true) - $timestampStart) % 86400) != 0) {
--- a/htdocs/core/lib/date.lib.php
+++ b/htdocs/core/lib/date.lib.php
@@ -677,6 +677,16 @@ function num_public_holiday($timestampStart, $timestampEnd, $country_code = '', 
 
 	$nbFerie = 0;
 
+	/**
+	 * This bit checks that the date we received is actually a GMT date. It doesn't
+	 * matter what date we're checking against, since the UTC clock does not have
+	 * daylight savings, but it's important that the timestamp represents a UTC 
+	 * midnight time.
+	 */
+	if (((dol_mktime(0, 0, 0, 10, 10, 2020, true) - $timestampStart) % 86400) != 0) {
+		die('Error: Dates must be GMT dates');
+	}
+
 	// Check to ensure we use correct parameters
 	if ((($timestampEnd - $timestampStart) % 86400) != 0) {
 		return 'Error Dates must use same hours and must be GMT dates';
--- a/htdocs/projet/activity/perweek.php
+++ b/htdocs/projet/activity/perweek.php
@@ -98,7 +98,8 @@ $next_month = $next['month'];
 $next_day   = $next['day'];
 
 // Define firstdaytoshow and lastdaytoshow (warning: lastdaytoshow is last second to show + 1)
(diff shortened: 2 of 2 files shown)


eldy:
I can't reproduce on current develop.
What is the time of your PHP (param date.timezone) ?
What is the timezone of your server (linux OS) ?

Csardelacal:
Everything seems fine to me. The server is in Germany and should be fine.

```
[root@web ~]# date +"%Z %z"
CEST +0200
[root@web ~]# php -r "echo date_default_timezone_get(); echo PHP_EOL;"
Europe/Berlin
[root@web ~]#  php -r "echo ini_get('date.timezone'); echo PHP_EOL;"
Europe/Berlin
```

eldy:
Still can't reproduce with these timezone. Can you paste a screenshot that show where is the offset ?

Csardelacal:
I'd have to ask the client for the screenshot. I'm not at the premise right now. But the issue is that the week overview displays stuff like "Easter Monday" on a Tuesday. But if the screenshot helps I'll be glad to procure it.

eldy:
Check if FIX of #17192 solve this trouble. If not, yes please send a screenshot (after applying the patch).
What do you mean with "week overview displays stuff like "Easter Monday" on a Tuesday." ?

Csardelacal:
Sorry, the issue seems completely unrelated. Here's a screenshot of our `projet/perweek` view and it clearly shows that "Easter Monday" is on a Tuesday. Which is funny, but not correct.

Screenshot from 2021-04-13 13-57-54

The issue you linked seems unrelated to the specific issue we're running into. And, after applying it, it's still not fixing the issue.

We have reproduced the issue on the demo server: https://demo.dolibarr.org/projet/activity/perweek.php?year=2021&month=6&day=7&search_usertoprocessid=2 . Here Mondays are off and Sundays are working days.

My guess is that `date.lib.php` expects a UTC timestamp to be passed to `num_public_holiday` in line 674 but the value for `$firstdaytoshow` in `perweek.php` is calculated using the local timezone. Which leads to the shift. In our testing this affects all holidays, both with fixed dates and with calculated dates.

**Note**: This pull request lead to different issues, so I would advise against merging it. It seems that the issue is rooted deeper in the way dolibarr handles timestamps, and that any attempt to fix this issue will lead to collateral issues.

Csardelacal:
I updated this PR with the fix that resolved the issue on our machine and seems to have the least collateral damages. This addresses two issues:

 1. `num_public_holiday` now checks the date it receives is actually a UTC timestamp at midnight. Preventing bad function calls.
 2. `perweek` calculates the offset in seconds between the timestamps that represent UTC and local midnight of each date. And then applies this to the `num_public_holiday`

I am unsure why travis is failing here. The issue it's failing with seems completely unrelated to the changes we made.

eldy:
What is your Dolibarr version ?

Csardelacal:
That would be 13.0.1

eldy:
Can you try with 13.0.4. i think it was already solved.

eldy:

PR #17215, opened 2021-04-11
Work on split module fournisseur

aspangaro:


htdocs/admin/stock.php
```diff
@@ -356,7 +356,7 @@
 	print '<tr class="oddeven">';
 	print '<td>'.$langs->trans("ReStockOnDispatchOrder").'</td>';
 	print '<td class="right">';
-	if (!empty($conf->fournisseur->enabled)) {
+	if (!empty($conf->fournisseur->enabled) && empty($conf->global->MAIN_USE_NEW_SUPPLIERMOD) || !empty($conf->supplier_order->enabled)) {
```
eldy, line 359: Mix of && and || without () is surely a bug.

aspangaro: Sorry, Resolve!

PR #17220, opened 2021-04-11
Update llx_10_c_regions.sql

dolibit-ut:
add new: Denmark


dolibit-ut:
please check denmark country id=80

PR #17222, opened 2021-04-12
FIX Write right on document

aspangaro:


htdocs/contact/document.php
```diff
@@ -168,7 +168,7 @@
 	$permission = $user->rights->societe->contact->creer;
 	$permtoedit = $user->rights->societe->contact->creer;
 	$param = '&id='.$object->id;
-	include DOL_DOCUMENT_ROOT.'/core/tpl/document_actions_post_headers.tpl.php';
+	include_once DOL_DOCUMENT_ROOT.'/core/tpl/document_actions_post_headers.tpl.php';
```
eldy, line 171: Never use include_once on a .tpl file.   Goal of tpl is to include code inline.
If you use include_once, php may decide to not include the php code because it was already loaded.

htdocs/don/document.php
```diff
@@ -182,7 +182,7 @@
 	print dol_get_fiche_end();
 
 	$modulepart = 'don';
-	$permission = $user->rights->don->lire;
+	$permission = $user->rights->don->creer;
```
eldy, line 185: permission may be to read and permtoedit to edit. So should be lire, shouldn't it ?

htdocs/contact/document.php
aspangaro: Hum.... No comment... I just added a PR that modifies all the files that include this tpl and that were all, I say all in include_once

htdocs/don/document.php
aspangaro: No, permission is here for a old compatibility and all document file, we have permission = permtoadd = creer and in file /core/tpl/document_actions_post_headers.tpl.php, we have this part : 

image

eldy:
Oh my god the include_once !!!

PR #17223, opened 2021-04-12
FIX : UPDATE query must not use alias for postgres compatibility

atm-gauthier:


atm-gauthier:
hi @eldy , i saw that if i insert lines with rowid (for example with llx_salary insert query in migration 13=>14 script) we need to use repair.php script with rebuild_sequences=confirmed url param.
is it possible to indicate somewhere to dolibarr that this method must be automatically run when we use migration script ?
Does Dolibarr already do it in develop ?

eldy:
In fact, if there is no bug into dolibarr, we should never call this method.
But i found 1 case where there is a bug.
When we load the chart of account of a country, we do insert having the rowid explicitely in the sql string. Postgresinsert it and does not increase the sequence. We should never make insert with rowid explictely named into the sql, but here we had to do it for historical reason to manage the multicompany. 
And may be the case of migration.
A fix to avoid to call the repair should be to add 
* for the load of chart of account, just after the run_sql, where $db->type = 'psql' to call the stock proc that repair the sequence 
* and do the same into the migration sql file by adding a line at end
-- VPGSQL8.2  call the proc stock to repair sequence

Goal is to avoid to call this when on pgsql (with mysql, there is no point).

PR #17225, opened 2021-04-12
NEW : Added of user groups for user exports

atm-adrien:
### FIX : Added of user groups for user exports

- Added of userGroup field on user exports


eldy:
Can you push it on develop ?

PR #17229, opened 2021-04-12
Update bom_card.php

dolibit-ut:
$title = $langs->trans('BOM');


dolibit-ut:
is sometime  trans('BOM')  and  trans('BillOfMaterials')  here used for the same content?

PR #17248, opened 2021-04-13
Batch mask product

altairis-noe:
# Add custom masks for products serial / lot numbers
This uses the new batch_mask db field to manage serial or lot numbers per product. It adds a field in the product card (visible if the option is selected in the product_batch module as in PR #17166 )
maskincard
maskincardedit
maskincardhidden


htdocs/product/card.php
```diff
@@ -1083,10 +1089,47 @@
 
 		// Batch number management
 		if (!empty($conf->productbatch->enabled)) {
-			print '<tr><td>'.$langs->trans("ManageLotSerial").'</td><td colspan="3">';
+			print '<tr><td>'.$langs->trans("ManageLotSerial").'</td><td>';
```
eldy, line 1092: All the lines before are using 4 columns.
You can remove the colspan to allow space for the td of batch number selection. But you must also print the <td></td> if we are no in the case the selection must be shown (so add a ampty td into the else for example)

htdocs/product/card.php
```diff
@@ -307,6 +307,9 @@
 			$object->status             	 = GETPOST('statut');
 			$object->status_buy            = GETPOST('statut_buy');
 			$object->status_batch = GETPOST('status_batch');
+			if ($object->status_batch !== 0) {
```
eldy, line 310: All the lines before are using 4 columns. Just do in all cases.
$object->batch_mask = GETPOST('batch_mask');
If status_batch is 0 then $object->batch_mask will have no effect "later" but we can store it if a value is set (in case of we switch the status_batch from 1 to 0 to 1, for example, it is always useless to keep data stored.

htdocs/product/card.php
```diff
@@ -475,6 +478,9 @@
 				$object->status                 = GETPOST('statut', 'int');
 				$object->status_buy             = GETPOST('statut_buy', 'int');
 				$object->status_batch = GETPOST('status_batch', 'aZ09');
+				if ($object->status_batch !== 0) {
+					$object->batch_mask = GETPOST('batch_mask', 'alpha');
```
eldy, line 482: Same. Just use  =  with no condition.
Condition should appears when we will need to know if we must use or not batch_mask.

altairis-noe:
I think I am on par with what you asked. I’d wish to point out that this PR will be useless until #17166 is merged. I think it is ready.

htdocs/product/card.php
```diff
@@ -1083,10 +1085,49 @@
...
+											$("#field_mask, #mask_option").toggleClass("hideobject");
+											$("#batch_mask_input").val("'.$inherited_mask_sn.'");
+										}
+									})
+								})
+							</script>';
+				} else {
+					print '<td colspan="2"></td>';
+				}
+			}
```
eldy, line 1128: the print '<td colspan="2"></td>'; should also be into this other else.

altairis-noe: I am not sure what you mean. What print am I missing ? I actually have put two "else", for add and edit.

altairis-tof: @altairis-noe he is talking about the else of "if ($status_batch !== '0') {" since you removed the colspan above.

altairis-noe: thanks @altairis-tof, I think I have it now !

eldy:
i have merged the #17166, but now there is a small conflict on language files that github does not allow me to fix. Can you fix it please ?

altairis-tof:
ok @eldy i've just resolve the conflict. you can merge now.

altairis-noe:
Hey there, @eldy, I think it is ready for merging, if this is OK for you !

PR #17252, opened 2021-04-13
# Fix # Label on select

BB2A-Anthony:

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -1962,7 +1962,7 @@ class Form
 					if (!is_numeric($show_empty)) {
 						$textforempty = $show_empty;
 					}
-					$out .= '<option class="optiongrey" value="'.($show_empty < 0 ? $show_empty : -1).'"'.((empty($selected) || in_array(-1, $selected)) ? ' selected' : '').'>'.$textforempty.'</option>'."\n";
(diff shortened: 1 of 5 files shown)


htdocs/core/class/html.form.class.php
```diff
@@ -1962,7 +1962,7 @@ public function select_dolusers($selected = '', $htmlname = 'userid', $show_empt
 					if (!is_numeric($show_empty)) {
 						$textforempty = $show_empty;
 					}
-					$out .= '<option class="optiongrey" value="'.($show_empty < 0 ? $show_empty : -1).'"'.((empty($selected) || in_array(-1, $selected)) ? ' selected' : '').'>'.$textforempty.'</option>'."\n";
+					$out .= '<option class="optiongrey" value="'.($show_empty <= 0 ? $show_empty : -1).'"'.((empty($selected) || in_array(-1, $selected)) ? ' selected' : '').'>'.$textforempty.'</option>'."\n";
```
eldy, line 1965: By convention (in Dolibarr) on list that list available id, when we want an empty value, we use the value -1 as the empty value. This is because sometimes, we have 0 as possible value for id (even if this is not the case for list of users).

bb2a: > By convention (in Dolibarr) on list that list available id, when we want an empty value, we use the value -1 as the empty value. This is because sometimes, we have 0 as possible value for id (even if this is not the case for list of users).

Ok @eldy i search a solution to display select wording in empty value for look feel and user frendly

BB2A-Anthony:
@eldy ok for me it's bug i supose.
image

On black mode

image

PR #17255, opened 2021-04-13
Compatibility phpv8

TheGivrix:
# Fix #16628 *Compatibility phpv8*
*Continuing development on pvpv8 compatibility*


htdocs/admin/mails_templates.php
```diff
@@ -254,7 +254,7 @@
 	$action = 'list';
 	$massaction = '';
 }
-if (!GETPOST('confirmmassaction', 'alpha') && $massaction != 'presend' && $massaction != 'confirm_presend') {
+if (!GETPOST('confirmmassaction', 'alpha') && !empty($massaction) && $massaction != 'presend' && $massaction != 'confirm_presend') {
```
eldy, line 257: This has a different result.
If massaction is not set, test is now false. Before it was true.

PR #17264, opened 2021-04-14
ADD Picto mass actions in lists

BB2A-Anthony:


eldy:
Goog.
However i will remove the `&ensp;` because a better practice is to use the 'class="pictofixedwidth"' as third parameter of img_picto. To the space between picto and text is managed by css.

PR #17266, opened 2021-04-14
# New Add picto in translate

BB2A-Anthony:
# New [*Add picto in translate*]

For simplify code add picto befor a translate

Add look feel


eldy:
The translation utility class should be reserved to translation and only translation (I think "presentation feature" are not at the good place into a translation utility method).
What is the goal of this ?

PR #17274, opened 2021-04-14
Project warehouse

altairis-noe:
# Link warehouses to projects
This will make possible to add a project to a warehouse using the fk_project field added in PR #17189. In this PR, we can link them in the creation and edition of a new warehouse, and it will display in the warehouse card, where it can be modified.
warehouse_new_project
warehouse_edit_project
warehouse_card_project
warehouse_card_edit_project


altairis-tof:
ok for me

PR #17276, opened 2021-04-14
FIX : Rebuild sequence for postgres only after query "INSERT INTO llx_salary(rowid..."

atm-gauthier:


atm-gauthier:
@eldy is it what you thought about for postgres rebuild sequence ?
(https://github.com/Dolibarr/dolibarr/pull/17223)

atm-gauthier:
@eldy i tried this migration script with rebuild sequence query with a postgres db and i confirm you that it corrects my problem

atm-gauthier:
@eldy could you say me if this PR is ok for you ?

PR #17281, opened 2021-04-15
Warehouses in project

altairis-noe:
# View and manage warehouses in project overview
This complements PR #17274 and adds warehouse management in project overview
project_warehouse


eldy:
To avoid to overload interface with warehouse in project (not a common case), can you the option like done for other object:

image

PR #17295, opened 2021-04-16
FIX : type link extrafield case for advanced target emailing

atm-gauthier:


htdocs/comm/mailing/class/advtargetemailing.class.php
```diff
@@ -640,6 +640,10 @@ public function query_thirdparty($arrayquery)
 						if ($arrayquery['options_'.$key]!=''){
 							$sqlwhere[]= " (te.".$key." = ".$arrayquery['options_'.$key].")";
 						}
+					} elseif ($extrafields->attributes[$elementtype]['type'][$key] == 'link') {
+						if ($arrayquery['options_'.$key] > 0){
+                                                        $sqlwhere[]= " (te.".$key." = ".$arrayquery['options_'.$key].")";
```
eldy, line 645: What is type of $arrayquery['options_'.$key]
If int add the cast (int) to avoid sql injection
If string, add ' ' and $this->db->escape()

atm-gauthier:
@eldy i did the correction

PR #17300, opened 2021-04-16
Update Supplier_Proposal\contact.php

fr69400:
actions may have been cancelled by triggers. This is not a system error


fr69400:
> actions may have been cancelled by triggers. This is not a system error

PR #17304, opened 2021-04-17
Batch : Expired Members & Dolibarr Backlink

NextGestion:
# Cancel Partnership For Expired Members
We must have a cron job that cancel status of partnership when subscription is expired + x days (x is a parameter of the cron job). This cron task must be supplied by module. If option 'PARTNERSHIP_IS_MANAGED_FOR' = 'thirdparty', this cron job does nothing.

# Warning of partnership if Dolibarr backlink not found on partner website
Batch not enabled by default. Every week, check that website contains a link to dolibarr.(org/fr/es) Web site. The url to check is a regex set as a parameter of module. If regex not found into the website of partner, batch increase a counter check fail (field count_last_url_check_error into table llx_partnership), if found, reset value to 0. If counter > 2, send email to partner (template email mut be defined in setup). If counter > 4, set status of partnership to canceled (set also reason of partnership cancelation).


NextGestion:
Bonjour @eldy , Merci de nous fournir plus d'informations concernant l'erreur déclaré par "**continuous-integration/travis-ci/pr**".

PR #17306, opened 2021-04-18
NEW Notification email for notification module can use an email template

ptibogxiv:
new version with hidden option for automatic emails to contacts/thirdparty

 from https://github.com/Dolibarr/dolibarr/pull/17011 and https://github.com/Dolibarr/dolibarr/pull/16900


ptibogxiv:
fix directory of file
avoid parasite emails with actions

PR #17314, opened 2021-04-19
NEW mcurrency rate 

atm-jpb:
## Context
Currently multicurrency logs past exchange rates when you "update" the exchange rate of a currency, but it doesn't enable you to correct past exchange rates or to date an exchange rate when you add it (the date is automatically set to the current time).

## This PR
This PR introduces some fix and behaviors to multicurrency list like : 

- empty form create by default
- edit inline 
- filter on date start-end
- filter select currency 

-  The page is accessible via the administration module.
multi-11


htdocs/multicurrency/multicurrency_rate.php
```diff
@@ -537,56 +481,66 @@
...
+
+		// USER REQUEST UPDATE FOR THIS LINE
+		if ($action == "updateRate" && $obj->rowid == $id_rate_selected) {
+			//  var_dump($obj);
+			$form = new Form($db);
+			print '<td>' . $form->selectDate($obj->date_sync, 'dateinput', 0, 0, 1).'</td>';
+
+
+			print '<td>' . $form->selectMultiCurrency($obj->code, 'multicurrency_code', 1, " code != '".$conf->currency."'", true) . '</td>';
+			print ' <td><input type="number" min ="0" step="any" class="minwidth200" name="rateinput" value="' . dol_escape_htmltag($obj->rate) . '"></td>';
```
eldy, line 493: type="number" : Does this mean the rate is an integer.
This HTML5 type must be avoided because it means the check on number syntax and decimal separator depends on the browser language and does not follow Dolibarr setup (that may be different).

edupulpillo:
The sync is not working in 14.0.2, and always is returned an empty error, after trying to sync with currencly layer
image

PR #17322, opened 2021-04-19
product and product_lot db fields

altairis-noe:
# Adds traceability fields for products and product_lots

this will add six fields in llx_product_lots : 
- eol_date -> end of life
- manufacturing_date -> when was it made
- scrapping_date -> when was it « trashed »
- commissionning_date -> when was it first used
- control_periodicity -> how often do we control it
- fk_status -> link to a dictionary associating it to its current state (available, in repair, out of order…)

and two fields in llx_product :
- life_duration -> how long is it supposed to work
- control_periodicity -> it may look as a duplicate from llx_product_lots but is actually necessary for legal reasons


htdocs/install/mysql/migration/13.0.0-14.0.0.sql
```diff
@@ -373,6 +373,15 @@ UPDATE llx_propal SET date_signature = date_cloture WHERE date_signature IS NULL
 
 
 ALTER TABLE llx_product ADD COLUMN batch_mask VARCHAR(32) NULL;
+ALTER TABLE llx_product ADD COLUMN life_duration VARCHAR(6);
```
eldy, line 376: Can you rename "life_duration" into "lifetime"
"control_periodicity" into "inspection_frequency"

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
```diff
@@ -373,6 +373,15 @@ UPDATE llx_propal SET date_signature = date_cloture WHERE date_signature IS NULL
...
+ALTER TABLE llx_product ADD COLUMN control_periodicity VARCHAR(6);
+
+ALTER TABLE llx_product_lot ADD COLUMN eol_date datetime;
+ALTER TABLE llx_product_lot ADD COLUMN manufacturing_date datetime;
+ALTER TABLE llx_product_lot ADD COLUMN scrapping_date datetime;
```
eldy, line 381: What is goal of this date ? What is translation in french ?

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
```diff
@@ -373,6 +373,15 @@ UPDATE llx_propal SET date_signature = date_cloture WHERE date_signature IS NULL
...
+ALTER TABLE llx_product_lot ADD COLUMN commissionning_date datetime;
```
eldy, line 382: Same question. What is goal of this date ? What is translation in french ?

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
```diff
@@ -373,6 +373,15 @@ UPDATE llx_propal SET date_signature = date_cloture WHERE date_signature IS NULL
...
+ALTER TABLE llx_product_lot ADD COLUMN control_periodicity VARCHAR(6);
```
eldy, line 383: Can you rename in 
qc_frequency   (Quality control Frequency)
I suppose this is what you mean.

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
altairis-noe: sure thing, I’m on it

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
altairis-noe: scrapping_date = mise au rebut

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
altairis-noe: commisionning = mise en service

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
altairis-noe: that’s exactly it means, I’ll do it

altairis-noe:
Hey there @eldy. I don’t know why travis-ci is in error here, it spots an error in some ja_JA lang file which I swear I did not touch. Do I have to merge a more recent develop branch into this one to get rid of it ?
Would you like me to rename those fields which meanings are not obvious. Would `destruction_date` be preferable than `scrapping_date` and `first_use_date` better than `commissioning_date` ?

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
eldy: Ok for scrapping_date, we can keep it as it is.

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
eldy: For which use cas do you need this date. What do you want to store in this date ?

lmag: There are many possibilities. We could offer them in the settings:
-Date of purchase
-Date of receipt (from buying or manufacturing)
-Shipping date
Or leave the field free

lmag: Initially we can do it in mass then by product in a second time

altairis-noe:
I changed the references to the "status" dictionnary to fk_qcstatus, in order to reflect changes in PR #17358

altairis-noe:
On second thoughts, the fk_qcstatus appears irrelevant in product_lot table, so I removed it from the PR

altairis-tof:
@eldy : this PR seems ok.

altairis-noe:
Well, that qcstatus is tricky. Turns out, in order to be useful, it has his place in the stock_mouvement table.

fappels:
> Well, that qcstatus is tricky. Turns out, in order to be useful, it has his place in the stock_mouvement table.

Think in both, one for current status and other for status history.

altairis-tof:
@fappels in a lot of 10 equipments, it is possible to have 4 KO and 6 OK so it is not possible to store it in produt_batch (if i havefinally understood the meaning ot this table)

fappels:
Indeed if you want to use the status for a lot of 10pcs product with 4 KO and 6 OK and not for unique equipment and use Product batch you need to put KO in other warehouse then the OK. Or maybe add llx_product_batch_qcstatus table with fk_product_batch.

altairis-tof:
@fappels yes i agree, there are 2 stock movements with a different qcstatus on each. and if the equipment is managed by unique SN (not a "lot") then you have the qcstatus in the latest stock movement.
yes another table will be useful to know exactly the current statuses in a lot... but i prefer @eldy agreement before coding again and again (even if the coder is @altairis-noe and not me)

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
eldy: Having such a date on the lot is not relevant.
If you buy 10 of Lot A the 1st and 10 of Lot A the 2nd, you can't store a date into the table that store the unique lot A. 
Suggested date are information related to a stock movement, or a sale, or a shipment, but is not not intrasec of a lot. I remove it for the moment so we can validate other fields.

eldy:
> scrapping_date

scrapping date is good.

eldy:
The quality control is not something on the stock movement level.
If you move a quantity of 10 and quality control say later that 6 ar OK and 4 are KO, there is no relation with stock movement.

If you want to flag a quality control for product, this must be store into the product_lot table (this is goal of such a table):

Into table lot, if you decide to flag a lot as bad (in such a case, it means, if a qty of 4 is KO for lot A, it is KO for every product with lot A, so for the other 6 quantity too).
If you want to make quantity control on a per unit product, if you have 10 product of lot A, it means you will get 6 product Lot A-OK and 4 products Lot A-KO.   Lot A-OK is a different lot than Lot A-KO (we want to make a distinction between two). This means we split a lot into 2 (Lot A sub-lot OK and the Lot A sub lot KO). This is rare (in most use case, goal of lot is to scrap any product with the lot if one is wrong). But some use cases exists to validate on a per unit. In such a case, this is managed by adding a prefix to a lot number to get a longer lot number. For example  Lot A-1 and Lot A-2.  A1 will be flagged as the good lot, A2 as the bad lot.
If Lot A is a serial (qty is always 1), you already have the minimum quantity so it will never be splitted in A1 and A2, A will be Good or BAD with no need of renaming the lot number.  But status is also stored on the lot level.

In conclusion, for serial or for lot, control quality is managed the same way, it is a flag/status on the lot table.

htdocs/install/mysql/tables/llx_product_lot.sql
```diff
@@ -18,15 +18,20 @@
...
+  sellby              date DEFAULT NULL,          -- Sellby date
+  datec               datetime,
+  tms                 timestamp DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
+  fk_user_creat       integer,
+  fk_user_modif       integer,
+  eol_date            datetime,                   -- End of life
+  manufacturing_date  datetime,
+  scrapping_date      datetime,                   -- Put to disposal
+  commissionning_date datetime,                   -- First use
+  qc_frequency        int,                        -- Quality control periodicity
```
eldy, line 35: Are you sure this is on product_lot ? This looks better on product.

htdocs/install/mysql/tables/llx_stock_mouvement.sql
```diff
@@ -37,5 +37,6 @@ create table llx_stock_mouvement
   fk_origin       integer,
   origintype      varchar(32),
   model_pdf       varchar(255),
-  fk_projet       integer NOT NULL DEFAULT 0
+  fk_projet       integer NOT NULL DEFAULT 0,
+  fk_qcstatus     VARCHAR(16)
```
eldy, line 41: A status of control quality must be store on the product_lot.

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
lmag: I agree

htdocs/install/mysql/tables/llx_stock_mouvement.sql
altairis-noe: The problem here is that parts of a same lot can have a different qc_status. You can have 200 broken pieces in a 500 pieces lot. If it doesn’t fit in the stock_movement table, we shall consider a new table.

lmag: I agree with @eldy  :
"The quality control is not something on the stock movement level."
This step can be carried out during other actions which could be:

- Remove from stock
- Re-invoice the customer
- etc ...

As such it seems to me that the initial quantity for the lot is missing
For example these quantities can be limited <1000 part for the "marque NF"

I think that could be a very good idea : 
"we shall consider a new table."
@eldy 
Do we offer you something like
Creation of a Quality Control table : 
-Numero lot
-Statut
-Quantité (par statut/par lot)

htdocs/install/mysql/tables/llx_product_lot.sql
lmag: The qc_frequency is normally on the product.
However, this can change over time depending on the products and regulations.
This is the reason why we added it on product and product_lot
@eldy

eldy: This means there is a default value at a time t (this should be in product). 
When a new lot arrives, we can store the value so we keep information of what qc frequency was when lot appears the first time.
Then in future we may decide to change the value. This will change the default value (so value on product and new lot).
So qc_frequency on lot is a copy of default value from product. So before having this field in product_lot, we should have it first on product.

Here we have several fields for different features need. This make difficult to validate because some things are good and some not.
I recommand to make one PR for fields for different cases:
field on product_lot for quality control so qc_frequency and fk_qcstatus
another for traceability so  eol_date, manufacturing_date, scrapping_date

htdocs/install/mysql/tables/llx_stock_mouvement.sql
eldy: @altairis-noe If you have 2 different lot with 2 different status, it can't be the same lot. Because, by definition a lot is the "smallest group of products that share the same properties". In such a cas the lot A will become Lot A and Lot A-Bad so 2 different lot.

altairis-noe: I’ll have to ask @lmag for his opinion, here, because I don’t understand lots that way. Products are sold in different quantities sharing an identifying lot number. If we create a lot each time part of lot is damaged or unavailable (even temporarily), I fear we lose pretty quick efficient traceability of the initial lot.

htdocs/install/mysql/tables/llx_product_lot.sql
altairis-noe: > Here we have several fields for different features need. This make difficult to validate because some things are good and some not. 

I hear you and will try to be more cautious in the content of my future PRs (I’m still new in Dolibarr world).

> So before having this field in product_lot, we should have it first on product.

That’s why I put it in `product` *and* `product_lot` in the current PR

htdocs/install/mysql/tables/llx_product_batch.sql
```diff
@@ -26,6 +26,6 @@ CREATE TABLE llx_product_batch (
   sellby datetime DEFAULT NULL,			-- deprecated. should not be used here but should be stored into a table llx_product_lot
   batch varchar(128) NOT NULL,
   qty double NOT NULL DEFAULT 0,
-  import_key varchar(14) DEFAULT NULL
+  import_key varchar(14) DEFAULT NULL,
```
frederic34, line 29: no comma here

htdocs/install/mysql/tables/llx_stock_mouvement.sql
altairis-noe: Anyhow, that qcstatus question in a tricky one, and we will not resolve it here, I take it out of the PR

htdocs/install/mysql/tables/llx_product_batch.sql
altairis-noe: thx, I corrected it

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
```diff
@@ -373,6 +373,16 @@ UPDATE llx_propal SET date_signature = date_cloture WHERE date_signature IS NULL
 
 
 ALTER TABLE llx_product ADD COLUMN batch_mask VARCHAR(32) NULL;
+ALTER TABLE llx_product ADD COLUMN lifetime VARCHAR(6);
+ALTER TABLE llx_product ADD COLUMN qc_frequency INT;
```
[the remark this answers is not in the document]
altairis-noe: I corrected it, thx

htdocs/install/mysql/tables/llx_product_lot.sql
lmag: @altairis-noe I agree with @eldy 
"qc_frequency on lot is a copy of default value from product. So before having this field in product_lot, we should have it first on product."

frederic34:
@altairis-noe db fields have been added by @eldy

htdocs/install/mysql/tables/llx_product_lot.sql
altairis-noe: And I think it is what I have done. Unless @eldy means I should have two PRs, one for each table ?

lmag:
I completely agree with @eldy @frederic34 
"This means there is a default value at a time t (this should be in product).
When a new lot arrives, we can store the value so we keep information of what qc frequency was when lot appears the first time.
Then in future we may decide to change the value. This will change the default value (so value on product and new lot).
So qc_frequency on lot is a copy of default value from product. So before having this field in product_lot, we should have it first on product.

Here we have several fields for different features need. This make difficult to validate because some things are good and some not.
I recommand to make one PR for fields for different cases:
field on product_lot for quality control so qc_frequency and fk_qcstatus
another for traceability so eol_date, manufacturing_date, scrapping_date"

Everything is ok in product
**But it missing fields in product_lot**

- commissionning_date datetime,                   -- First use
- qc_frequency        int,                        -- Quality control periodicity

eldy:
Waiting for more maturity on #17322 and quality control frequency field into product before adding it into product_lot (that is a more detailed case of the generic case with field on product).

eldy:
Fields seems to be added the  fields seems to have been added according to frederic comment.
So i think we can close this PR.

PR #17342, opened 2021-04-19
New substitution key online member payment link in mailing

daraelmin:


daraelmin:
@eldy I dont know why this code does not pass the security test, but this substitution key would be very usefull to recall late members. 
Thx in advance for your revew

daraelmin:
Wip still not functionnal

PR #17343, opened 2021-04-20
NEW : collapse batch detail

atm-gauthier:


atm-gauthier:
@eldy travis failed but error doesn't deal with what i added

htdocs/fichinter/list.php
```diff
@@ -288,7 +288,7 @@
 	$sql .= " AND s.rowid = sc.fk_soc AND sc.fk_user = ".$user->id;
 }
 if ($socid) {
-	$sql .= " AND s.rowid = ".((int) $socid);
+	$sql .= " AND s.rowid = ".$socid;
```
eldy, line 291: Why removing the cast (it is a protection we should have when forging a sql with a test = on a non string data.

atm-gauthier: Hi @eldy
 it's a mistake, I will restore the cast on tomorrow ;)

atm-gauthier: @eldy it's ok now

atm-gauthier:
@eldy everything's ok for you in this PR ?

PR #17366, opened 2021-04-21
Fix #17318 false renew date

### LINKED ISSUE #17318: Bug: In renew form subscription validity date are wrong registered for the subscription
# Bug: In renew form subscription validity date are wrong registered for the subscription

## Environment
- **Version**: 13.0.2
- **URL(s)**: public/payment/newpayment.php

When the member is verx late (more than one year in my case) the renewal form set the subscription for the first unpaid period in llx_subscritpion but set the validity membership validity to the next and period in llx_adherent (end of present year in my case). So the subscription appears expired, but the member is up to date.

I'm expecting that the subscription date should be for the present period as the member won't ever pay for a period outdated.

1) set a member late more than duration 
2) use renew form and pay subscription
3) see result

subscription3

daraelmin:
#Fix #17318 false renew date when late for more the one year
#Fix excluded membre must ne validated by user action

### DIFF
--- a/htdocs/public/payment/paymentok.php
+++ b/htdocs/public/payment/paymentok.php
@@ -394,7 +394,7 @@ if ($ispaymentok) {
 
 			// Do action only if $FinalPaymentAmt is set (session variable is cleaned after this page to avoid duplicate actions when page is POST a second time)
 			if (!empty($FinalPaymentAmt) && $paymentTypeId > 0) {
-				$result = $object->status < $object::STATUS_DRAFT ? -1 : $object->validate($user); // if membre is excluded (status == -2) the new validation is not possible
+				$result = ($object->status == $object::STATUS_EXCLUDED) ? -1 : $object->validate($user); // if membre is excluded (status == -2) the new validation is not possible
 				if ($result < 0 || empty($object->datevalid)) {
 					$error++;
 					$errmsg = $object->error;
--- a/htdocs/public/payment/paymentok.php
+++ b/htdocs/public/payment/paymentok.php
@@ -394,7 +394,7 @@ if ($ispaymentok) {
 
 			// Do action only if $FinalPaymentAmt is set (session variable is cleaned after this page to avoid duplicate actions when page is POST a second time)
 			if (!empty($FinalPaymentAmt) && $paymentTypeId > 0) {
-				$result = $object->status == -2 ? -1 : $object->validate($user); // if membre is excluded (status == -2) the new validation is not possible
+				$result = $object->status == $object::STATUS_RESILIATE ? -1 : $object->validate($user); // if membre is excluded (status == -2) the new validation is not possible
 				if ($result < 0 || empty($object->datevalid)) {
 					$error++;
 					$errmsg = $object->error;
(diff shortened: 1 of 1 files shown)


htdocs/public/payment/paymentok.php
```diff
@@ -412,7 +412,13 @@
 
 				$datesubend = null;
 				if ($datesubscription && $defaultdelay && $defaultdelayunit) {
-					$datesubend = dol_time_plus_duree(dol_time_plus_duree($datesubscription, $defaultdelay, $defaultdelayunit), -1, 'd');
+					$datesubend = dol_time_plus_duree($datesubscription, $defaultdelay, $defaultdelayunit);
+					// the new end date of subscription must be in futur
```
eldy, line 416: Not necessarly. When a subscription is done and is was expired since 10 days for example, the subscription for the year must restart from the end date of previous subscription, so 10 days ago.

htdocs/public/payment/paymentok.php
```diff
@@ -394,7 +394,7 @@
 
 			// Do action only if $FinalPaymentAmt is set (session variable is cleaned after this page to avoid duplicate actions when page is POST a second time)
 			if (!empty($FinalPaymentAmt) && $paymentTypeId > 0) {
-				$result = $object->validate($user);
+				$result = $object->status == -2 ? -1 : $object->validate($user); // if membre is excluded (status == -2) the new validation is not possible
```
eldy, line 397: Can you replace the -2 with $object::STATUS_RESILIATE ?

htdocs/public/payment/paymentok.php
daraelmin: Yes but the end date should be in future.

That is the purpose of the while as it loop on the end sub date.

htdocs/public/payment/paymentok.php
daraelmin: Only the excluded member should not be able to reopen their account with public payment. 

Why a resiliated member couldn't change his mind and reactivate his account ?

htdocs/public/payment/paymentok.php
eldy: I see.
Why 
while (dol_time_plus_duree($datesubend, -1, 'd') < $now) {
and not
while ($datesubend < $now) {
removing 1 day into a look that add a delay that may be 1 day to (even if commonly it is 1 year) may generate an infinite loop

daraelmin: Much better. Thx

daraelmin: > I see.
> Why 
> while (dol_time_plus_duree($datesubend, -1, 'd') < $now) {
> and not
> while ($datesubend < $now) {
> removing 1 day into a look that add a delay that may be 1 day to (even if commonly it is 1 year) may generate an infinite loop

I'll change that, I used this cause it was the old way to define the end date

htdocs/public/payment/paymentok.php
daraelmin: Your commit is much better thx

PR #17367, opened 2021-04-22
Update llx_10_c_regions.sql

dolibit-ut:
Peru
Hungary


dolibit-ut:
something is not correct at Andorra:
id country=18   ≠  fk_pays  56

PR #17368, opened 2021-04-22
block purging of security logs

cfoellmann:
This add the conf `MAIN_DISABLE_SECURITY_EVENTS_PURGE`
when set to `1` admins are not able to purge the security related events

### DIFF
--- a/htdocs/admin/tools/listevents.php
+++ b/htdocs/admin/tools/listevents.php
@@ -138,7 +138,7 @@ if (GETPOST('button_removefilter_x', 'alpha') || GETPOST('button_removefilter.x'
 }
 
 // Purge audit events
-if ($action == 'confirm_purge' && $confirm == 'yes' && $user->admin) {
+if ($action == 'confirm_purge' && $confirm == 'yes' && $user->admin && empty($conf->global->MAIN_DISABLE_SECURITY_EVENTS_PURGE)) {
 	$error = 0;
 
 	$db->begin();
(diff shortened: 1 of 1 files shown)


daraelmin (commented):
But admin can update the constant to 0

cfoellmann:
maybe go for a parameter in the conf.php?

daraelmin:
In my opinion, you'd better define a user role "webadmin" who's got all permission but is not an admin.

cfoellmann:
but a non-admin can not get the security events.
purging seems too invasive to me and should be avoided.

daraelmin:
What do you think of a new permission on the purging security log for non admin ?

And did you try the module "Archives / Logs Unalterable" ? I'm not sure of the result.

cfoellmann:
the module "Archives / Logs Unalterable" logs different events. When I got time I might look into extending that module
But the logs of security events can be crucial when you have a breach or a rough admin

eldy:
Currently, only an admin user can purge the security logs.
So what is a goal to set an option when the admin user can modify this option ?

cfoellmann:
I understand what you guys mean.

Maybe I should add a conf.php variable for this?
concerning this it would be great to modify this in staging environments. @eldy : How does `$this->global->LOCAL_CONSTS_FILES` work`? That way we could overwrite conf.php values in different environments!?

eldy:
LOCAL_CONSTS_FILES should not be used as it is a serious security hole.
You can add a global var inside common conf.php file, for example $dolibarr_disable_audit_purge and reuse it if defined to lock the feature you want to lock

cfoellmann:
how about removing the LOCAL_CONSTS_FILES ?

I will add my purging block as a conf.php parameter.

@eldy I am missing the WordPress ability to overwrite any option by hook.
My example issue is `MAIN_MAIL_FORCE_SENDTO`. I copy the complete DB to staging and need to set the mail address new everytime. (Will set up a script to edit the db but a function for the conf.php would be great to overwrite const items

PR #17369, opened 2021-04-22
allow proper filters for listUsersForGroup

cfoellmann:
when you use the `function listUsersForGroup($excludefilter = '', $mode = 0)` there is no access to the users data to be used in the parameter `$excludefilter`
this change loads the complete user row to exclude users via the parameter

### DIFF
--- a/htdocs/user/class/usergroup.class.php
+++ b/htdocs/user/class/usergroup.class.php
@@ -237,7 +237,7 @@ class UserGroup extends CommonObject
 
 		$ret = array();
 
-		$sql = "SELECT u.rowid";
+		$sql = "SELECT u.*";
 		if (!empty($this->id)) {
 			$sql .= ", ug.entity as usergroup_entity";
 		}


eldy:
The exclude filter is used to forge the WHERE of the sql. there is no need to have it into the list of fields in the SELECT.

cfoellmann:
Please say that I am wrong but in my "experience":
The WHERE part can only compare to data that is in the SELECT data?!

Example:
When I do this:
`listUsersForGroup('u.statut = 1', 1)` it did still gave back deactivated users.

Is there a possibility to do this comparison without the change above?

eldy:
No a select can make operation on any fields even if not into the list of return fields (at least into a WHERE, not into a HAVING).
Doing listUsersForGroup('u.statut = 1', 1)  should return only users with statut = 1.
What sql request do you have in log ?

cfoellmann:
@eldy your are completely right. my issue must have been somewhere else

PR #17370, opened 2021-04-22
NEW Make public bookmarks editable by admin users only

cfoellmann:
this fixes issues with the bookmarks module

public bookmarks added by admins can now only be added, edited and removed by admins


daraelmin (changes requested):
The code is not functionnal as it throw a parse error.
You have at least two missing openning "{" 

One on line 272
One on line 275

in list.php

cfoellmann:
@daraelmin thanks.
did the fixes

PR #17372, opened 2021-04-22
FIX: invoice PDF generation after payment

marc-dll:
- FIX: payment creation: re-generate invoice PDF with correct display options
- FIX: payment validation: invoices PDF were no re-generated, make it with correct display options (this is an hidden option)


marc-dll:
Just for fun, the second issue dates back to 2007 : the TODO in this commit was never done... until today :D : https://github.com/Dolibarr/dolibarr/commit/a4cff6c06304aa21a9a597b8fc204fcebd6086d7

PR #17383, opened 2021-04-23
FIX : In migration script, we set fk_account of tva and salary with bank line fk_account

atm-gauthier:


htdocs/install/mysql/migration/13.0.0-14.0.0.sql
```diff
@@ -205,6 +205,8 @@ INSERT INTO llx_payment_vat (rowid, fk_tva, datec, datep, amount, fk_typepaiemen
 --ALTER TABLE llx_tva DROP COLUMN fk_bank;
 
 ALTER TABLE llx_tva ALTER COLUMN paye SET DEFAULT 0;
+-- VMYSQL4.3 UPDATE llx_tva tva INNER JOIN llx_bank b ON (b.rowid = tva.fk_bank) SET tva.fk_account = b.fk_account WHERE tva.fk_account IS NULL;
+-- VPGSQL8.2 UPDATE llx_tva tva SET fk_account = b.fk_account FROM llx_bank b WHERE b.rowid = tva.fk_bank AND tva.fk_account IS NULL;
```
eldy, line 209: Do you mean UPDATE llx_tva as tva  (it is better to have the as explicitely mentionned)

atm-gauthier: Hi @eldy, fk_bank doesn't exists for new install in 14.0 (when it will be released), but it's not yet deleted for instances which migrate from dolibarr < 14.0 to dolibarr 14.0 and for these instances my queries fill object fk_account only where fk_bank link exists (new 14.0 tva and salary objects don't use fk_bank field then they will never be updated by my queries).
And during the migration, 1 vat = 1 vat payment and 1 salary = 1 salary payment, we'll only be able to set several payments for 1 object after migration

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
```diff
@@ -205,6 +205,8 @@ INSERT INTO llx_payment_vat (rowid, fk_tva, datec, datep, amount, fk_typepaiemen
 --ALTER TABLE llx_tva DROP COLUMN fk_bank;
 
 ALTER TABLE llx_tva ALTER COLUMN paye SET DEFAULT 0;
+-- VMYSQL4.3 UPDATE llx_tva as tva INNER JOIN llx_bank b ON (b.rowid = tva.fk_bank) SET tva.fk_account = b.fk_account WHERE tva.fk_account IS NULL;
```
eldy, line 208: Column tva.fk_bank does not exists anymore in v14

atm-gauthier: @eldy OK, I thought it could be interesting but no problem, it's not very important

PR #17385, opened 2021-04-23
Compatibility phpv8

TheGivrix:
*Continuous integration for phpv8*


htdocs/admin/tools/listevents.php
```diff
@@ -245,55 +245,55 @@
 	if ($limit > 0 && $limit != $conf->liste_limit) {
 		$param .= '&limit='.urlencode($limit);
 	}
-	if ($optioncss != '') {
+	if (!empty($optioncss) && $optioncss != '') {
 		$param .= '&optioncss='.urlencode($optioncss);
 	}
-	if ($search_code) {
+	if (!empty($search_code)) {
```
eldy, line 251: Did you get an error on this ?
Because $search_code is initializer at begin with 
$search_code = GETPOST('search_code');
So value should always be defined and you should not get any warning, even with PHP8

Givriz: I didn't got an error, but souldn't we do it anyway ? It cost nothing

eldy: if ($search_code) {
is easier to read than 
if (!empty($search_code)) {
Two negation is always a pain for developer. So it is better to keep if ($search_code) {

Also, there is ton of case of var, if we must protect any test on any var, it is an endless work. So to see the end of the tunnel, it is better to focus only on line that generate a warning with php8. This reduce also the rate of conflict for other PR and make impact analysis faster when we need to make diff between 2 versions.

PR #17387, opened 2021-04-23
Fix Name on filter select close #17357

### LINKED ISSUE #17357: BUG No label name on filter select
# Bug
[*No label name on filter select*]

For user frendly it's good to read the label name

image

On black mode

image

Dolibarr version : 14

BB2A-Anthony:
# Fix #17357[*Name on filter select*]

# Close #17357[*BUG No label name on filter select*]


eldy:
Seems already fixed by another PR

BB2A-Anthony:
> Seems already fixed by another PR
@eldy where is the PR because i have always the bug

PR #17393, opened 2021-04-25
Close/CLOSE #17390 Products API, GET /products without variants

### LINKED ISSUE #17390: Rest API, get products without variant products
# Feature Request
When I am acceesing the get route for /products, the response I get is the both 'basic' products and it's variations as different products without any connection between them. 
I think that this is what happening and I didn't miss something.

Is there any way to access the 'basic' products without variations being in the same list? Or access a list of products with it's variations inside the products object.

If  there isn't this kind of functionlality, i think this would be a good feature for the products api..

## Suggested implementation
This can be implemented as a filter to /products

kastoras:
When accessing get /products, using $without_variants parameter to true (the deafult value is false), the api return only products and no variants.

# Close #17390
When calling Rest API get products, I am getting all products & variants

# New, Get only Products without variants
I added one new parameter $without_variants when calling GET /products. When this parameter is true I added a new SQL statment ($sql .= ' AND t.rowid in (SELECT distinct fk_product_parent FROM '.MAIN_DB_PREFIX.'product_attribute_combination')
Using this sql condition, the inner select query will gather all product IDs that act as a parrent product for variant products.
Finaly using t.rowid IN in the above IDs the api returns only products and not variants.

### DIFF
--- a/htdocs/product/class/api_products.class.php
+++ b/htdocs/product/class/api_products.class.php
@@ -166,10 +166,10 @@ class Products extends DolibarrApi
 	 * @param  int    $category   			Use this param to filter list by category
 	 * @param  string $sqlfilters 			Other criteria to filter answers separated by a comma. Syntax example "(t.tobuy:=:0) and (t.tosell:=:1)"
 	 * @param  bool   $ids_only   			Return only IDs of product instead of all properties (faster, above all if list is long)
-	 * @param  bool   $without_variants   	Return only parent products, not variants
+	 * @param  int    $variant_filter   	Use this param to filter list (0 = all, 1=products without variants, 2=parent of variants, 3=variants only)
 	 * @return array                		Array of product objects
 	 */
-	public function index($sortfield = "t.ref", $sortorder = 'ASC', $limit = 100, $page = 0, $mode = 0, $category = 0, $sqlfilters = '', $ids_only = false, $without_variants = false)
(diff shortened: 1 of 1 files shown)


htdocs/product/class/api_products.class.php
```diff
@@ -186,6 +187,11 @@ public function index($sortfield = "t.ref", $sortorder = 'ASC', $limit = 100, $p
 			$sql .= ", ".MAIN_DB_PREFIX."categorie_product as c";
 		}
 		$sql .= ' WHERE t.entity IN ('.getEntity('product').')';
+
+		if ($without_variants) {
+			$sql .= ' AND t.rowid in (select distinct fk_product_parent from '.MAIN_DB_PREFIX.'product_attribute_combination)';
```
eldy, line 192: Some products are A) common products, some are B) parent of variant and some are C) variant.
Here adding the test will return all products B (right).
Why not but it means label of parameter is wrong (without_variant means A for me).
If what you want if really B, i suggest to rename parameter into "$parent_of_variant"

Or better, rename it into "$variant_filter" with 2 values (0 = all, 1=A only, 2=B only, 3=C only). Need more change. First suggestion correction will be enough.

kastoras: Good point. I will make the change and push it again.

kastoras:
I have implemented this using sub-queries. I could do this using joins if this is more preferable.

PR #17394, opened 2021-04-25
New Add option for doctemplate

BB2A-Anthony:
# New [*Add option for doctemplate*]
[*Long description*]

Add option for doctemplate for user friendly

If is good for you i cuntinue for the other options


eldy:
Because those option is only for the PDF templates, and not ODT or other external templates, i would move this tab into the section Home - Setup - PDF.
I would introduce a tab "Common option" and a tab "Proposals", and then other tabs for other modules...
Can you move the tab ?

BB2A-Anthony:
Yes i look

BB2A-Anthony:
> Because those option is only for the PDF templates, and not ODT or other external templates, i would move this tab into the section Home - Setup - PDF.
> I would introduce a tab "Common option" and a tab "Proposals", and then other tabs for other modules...
> Can you move the tab ?

@eldy It's ok for me in PR #17402 i close this

PR #17400, opened 2021-04-26
FIX Supplier invoices (deposit, multicurrency in lines)

kkhelifa-opendsi:
FIX Supplier invoices (deposit, multicurrency in lines)


htdocs/fourn/class/paiementfourn.class.php
```diff
@@ -230,7 +230,66 @@ public function create($user, $closepaidinvoices = 0, $thirdparty = null)
...
+
+										if ($error) {
+											setEventMessages($discount->error, $discount->errors, 'errors');
+											$error++;
+										}
+									}
+
+									// Set invoice to paid
+									if (!$error) {
+										$result = $invoice->set_paid($user, '', '');
```
eldy, line 287: Old code was using setPaid() 
set_paid seems deprecated. Can you check ?

PR #17401, opened 2021-04-26
New #14826 Multicompany PMP per entity

ibuiv:
# New PMP/AWP per product AND per entity

Issue [#14826](https://github.com/Dolibarr/dolibarr/issues/14826)

General Idea : Calculate the PMP for each entity and each product instead of each product. 
## Actual Behaviour (Classic)
PMP is calculated by product when a stock movement is created. All the entities can update this value. It's a problem if the entities are different legal entities.

## Updated Behaviour (If MultiCompany is activated with this new option implemented)
PMP is calculated and depends on a couple entity/product. So, each entity has its own PMP for each product.
 - **Warning** : If several entities supply the same warehouse. The total PMP of the warehouse may be different for each entity.
 - **Solution** : --COMPLICATED-- Must create fields 'owner' for each produc in the warehouse => may create a lot of identical lines and the dolibarr functions to get stock must be updated

 - On Dolibarr Core : 
    - product.class.php : fetch() gets the pmp per entity if not empty instead of classic pmp
    - entrepot.class.php : nb_products() gets the pmp per entity if not empty instead of classic pmp
    - product/stock/card.php : the SQL query is updated to take in consideration separated PMP

> ### Goes with updates on **Multicompany core code** *(Merge Request done on iNodBox framagit)*
>  - Added new SQL Table `llx_entity_product_pmp` (Table + keys created on `init`)
>  - If init while `llx_entity_product_pmp` Table doesn't exists or is empty, call `updatePMP()` cloning all existing PMPs to the new table for each product and entity.
>  - Constant added to inform Dolibarr Product Class and Entrepot Class that the PMP is in another table `MULTICOMPANY_PMP_PER_ENTITY_ENABLED`
>  - Created another trigger file with lesser priority than the existing one (to separate the ones dedicated to the PMP)
>    - Duplicated the PMP calculation method on the `STOCK_MOVEMENT` trigger to stock the result on the new table.
>    - Trigger added on MULTICOMPANY_CREATE to add PMPs for each product for the entity newly created
>


ibuiv:
**SOLUTION FOUND?** (Job in progress)

Travis CI returns 
CodingPhpTest::testPHP
Found non sanitized string in building of a IN or NOT IN sql request product/class/product.class.php - Bad.

How Can I fix this test error ? 

I think the error comes from : 


```
$visibleWarehousesEntities = $conf->entity;
if (isset($mc->sharings['stock']) && !empty($mc->sharings['stock'])) {
      $visibleWarehousesEntities .= "," . implode(",", $mc->sharings['stock']);
}

...

if ($separatedStock) {
	$sql .= " AND sp.fk_entrepot IN (
		      SELECT rowid
		      FROM ".MAIN_DB_PREFIX."entrepot	WHERE entity IN (" . $visibleWarehousesEntities	."))";
}

the source of $visibleWarehousesEntities is from $conf ans $mc (variable from MultiCompany Module)

```

eldy:
entity IN (" . $visibleWarehousesEntities	.")
must be replace with 
entity IN (" .$this->db->sanitize( $visibleWarehousesEntities)	.")

htdocs/product/class/product.class.php
```diff
@@ -2181,12 +2181,45 @@ public function fetch($id = '', $ref = '', $ref_ext = '', $barcode = '', $ignore
 		} else {
 			$sql .= " pa.accountancy_code_buy, pa.accountancy_code_buy_intra, pa.accountancy_code_buy_export, pa.accountancy_code_sell, pa.accountancy_code_sell_intra, pa.accountancy_code_sell_export,";
 		}
-		$sql .= " p.stock,p.pmp, p.datec, p.tms, p.import_key, p.entity, p.desiredstock, p.tobatch, p.batch_mask, p.fk_unit,";
+		//For MultiCompany PMP per entity
+		$separatedEntityPMP = false;
+		if (!empty($conf->global->MULTICOMPANY_PRODUCT_SHARING_ENABLED) && !empty($conf->global->MULTICOMPANY_PMP_PER_ENTITY_ENABLED)) {
+			$checkPMPPerEntity = $this->db->query("SELECT pmp FROM " . MAIN_DB_PREFIX . "entity_product_pmp WHERE fk_product  = ".((int) $id)." AND entity = ".(int) $conf->entity);
```
eldy, line 2187: The table entity_product_pmp does not exists.
You use the llx_product_perentity that exists but must enhance it by adding the pmp field.

htdocs/product/class/product.class.php
```diff
@@ -2181,12 +2181,45 @@ public function fetch($id = '', $ref = '', $ref_ext = '', $barcode = '', $ignore
...
+			if ($this->db->num_rows($checkPMPPerEntity)>0) {
+				$separatedEntityPMP = true;
+			}
+		}
+		//For MultiCompany Stocks Sharings stock_reel includes only stocks shared with this entity
+		$separatedStock = false;
+		if (!empty($conf->global->MULTICOMPANY_STOCK_SHARING_ENABLED)) {
```
eldy, line 2194: To enhance readability of code, can you merge all new code neede by this new feature into the same
if (!empty($conf->global->MULTICOMPANY_STOCK_SHARING_ENABLED)) {

}

ibuiv: sanitize worked
I will work on all your demands
Thanks for your remarks

ibuiv: updates asked done
but Travis throws a test error for a file and a method I didn't touch ... 

1) CodingPhpTest::testPHP
Found a $this->db->idate to forge a sql request without quotes around this date field compta/facture/class/facture.class.php

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
```diff
@@ -307,7 +307,8 @@ create table llx_product_perentity
     accountancy_code_sell_export  varchar(32),                        -- Selling accountancy code for vat export
     accountancy_code_buy          varchar(32),                        -- Buying accountancy code
     accountancy_code_buy_intra    varchar(32),                        -- Buying accountancy code for vat intracommunity
-    accountancy_code_buy_export   varchar(32)                  		-- Buying accountancy code for vat import
+    accountancy_code_buy_export   varchar(32),                     		-- Buying accountancy code for vat import
+    pmp double(24,8)
```
eldy, line 311: The field must also be added into the llx_product_perentity.sql file so it will be availble for new installation too.

ibuiv:
Fixed sql 13.0.0 -> 14.0.0 instruction, `ADD COLUMN` instead of just `ADD`

the error is always there ... 
CodingPhpTest::testPHP
Found a $this->db->idate to forge a sql request without quotes around this date field compta/facture/class/facture.class.php

don't know why ... any idea ?

ibuiv:
Should be OK now

PR #17403, opened 2021-04-26
New/NEW Product Variants API, add variant stock to response

kastoras:
When acessing product variants via GET API /products/{id}/variants, Include stock_warehouse data to the response


htdocs/product/class/api_products.class.php
```diff
@@ -1574,6 +1575,11 @@ public function getVariants($id)
 			$prodc2vp = new ProductCombination2ValuePair($this->db);
 			$combinations[$key]->attributes = $prodc2vp->fetchByFkCombination((int) $combination->id);
 			$combinations[$key] = $this->_cleanObjectDatas($combinations[$key]);
+
+			$productModel = new Product($this->db);
+			$productModel->fetch((int) $combination->fk_product_child);
+			$productModel->load_stock();
```
eldy, line 1581: Because we want API to be a low level and this method may generate a high level of subrequest, it is better to add an option to do this.
For example, other api have a parameter call $includestockdata set to 0 by default. And load of stock is done only if this param is set to 1. Can you add this ?

kastoras: Yes, of course. I will make that change.

kastoras:
I have add the suggested fix.

PR #17404, opened 2021-04-26
New Add option for doctemplate

BB2A-Anthony:
# New [*Add option for doctemplate*]
[*Long description*]
Add option for doctemplate for user friendly

If is good for you i cuntinue for the other options

@eldy its ok for me close your demande #17394


htdocs/core/lib/pdf.lib.php
```diff
@@ -35,6 +36,85 @@
...
+	$head[$h][2] = 'general';
+	$h++;
+
+	// Show more tabs from modules
+	// Entries must be declared in modules descriptor with line
+	// $this->tabs = array('entity:+tabname:Title:@mymodule:/mymodule/mypage.php?id=__ID__');   to add new tab
+	// $this->tabs = array('entity:-tabname:Title:@mymodule:/mymodule/mypage.php?id=__ID__');   to remove a tab
+	complete_head_from_modules($conf, $langs, null, $head, $h, 'pdf_admin');
+
+	$head[$h][0] = DOL_URL_ROOT.'/admin/pdf_project.php';
```
eldy, line 62: You should add a condition here if module is on for each tab. For example:
if (!empty($conf->projet->enabled)) {

BB2A-Anthony:
> You should add a condition here if module is on for each tab. For example:
if (!empty($conf->projet->enabled)) {
 

@eldy it's ok for me

PR #17405, opened 2021-04-27
Fix #17219 : complement of PR 17161

### LINKED ISSUE #17219: Complement of PR 17161
Are you sure the transformation is done on "current selected dir" only ?
You probably need to add some js script to process only the current dir. To be sure, you should show value of relative directory into the confirm box.
Also, when we are into a subdir and we click on button, we should remains on same open dir instead of going back to root.

Hystepik:
Fix of possibility to create webp in a directory using js


htdocs/core/tpl/filemanager.tpl.php
```diff
@@ -87,10 +87,27 @@
...
 	print img_picto('', 'images', '', false, 0, 0, '', 'size15x flip marginrightonly');
 	print '</a>';
 }
+if ($permtoadd && $module == 'ecm') {	// If on file manager medias in ecm
+	print '<a id="generateimgwebp" href="'.$_SERVER["PHP_SELF"].'?action=confirmconvertimgwebp" class="inline-block valignmiddle toolbarbutton paddingtop" title="'.dol_escape_htmltag($langs->trans("GenerateImgWebp")).'">';
+	print img_picto('', 'images', '', false, 0, 0, '', 'size15x flip marginrightonly');
+	print '</a>';
+}
+print "<script>
+$(\"#generateimgwebp\").on(\"click\",function(){
```
eldy, line 100: To hep debug when async js is done, you should add a log into console to help user that try to track what happen understand. For exemple here:
console.log("We click to generate webp image, we set current dir into hidden vars");

htdocs/core/tpl/filemanager.tpl.php
```diff
@@ -139,22 +156,30 @@
...
+		$formquestion['website']=array('type'=>'hidden', 'value'=>$website->ref, 'name'=>'website');
+	}
+	print $form->formconfirm($_SERVER["PHP_SELF"], $langs->trans('ConfirmImgWebpCreation'), $langs->trans('ConfirmGenerateImgWebp', $object->ref), 'convertimgwebp', $formquestion, "yes", 1);
 	$action = 'file_manager';
 }
 
 if ($action == 'convertimgwebp' && $permtoadd) {
 	if ($module == 'medias') {
-		$imagefolder = $conf->website->dir_output.'/'.$websitekey.'/medias/image/'.$websitekey.'/';
+		$imagefolder = $conf->website->dir_output.'/'.$websitekey.'/medias/'.GETPOST('section_dir', 'alpha');
```
eldy, line 173: Because data can come from user input, to avoid to forge bad url, use dol_sanitizePathName to clean GETPOST('section_dir', 'alpha')

PR #17406, opened 2021-04-27
FIX : Change status values taken for load_stats_commande_fournisseur()

atm-adrien:
# FIX : Change status value taken in load_stats_commande_fournisseur() function

The status validated and approved is taken in the function, but at this status, the order is not placed so it should not be part of the supplier ordered virtual stock

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -4479,7 +4479,11 @@ class Product extends CommonObject
 		}
 		if (! empty($conf->fournisseur->enabled))
 		{
-			$result=$this->load_stats_commande_fournisseur(0, '3,4', 1);
+			if (!empty($conf->global->SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK)){
+				$result=$this->load_stats_commande_fournisseur(0, $conf->global->SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK, 1);
+			} else {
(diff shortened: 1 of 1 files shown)


eldy:
Agree with status = 1 (validated but not yet approved).
Bu i think status = 2 should remains. If purchase order has been approved, even if not yet ordered to customer, it is into the pipe and we are sure that some units will be received. We know that it will be sent to a supplier and will be received later because it has been approved. So i think we can include it into virtual stock. Don't you think so ?

atm-adrien:
I don't quite agree with that. From my perspective, virtual stock is all physical stock minus quantities in sales orders placed but not yet shipped, plus quantities in supplier orders that have been placed but not yet received.
I don't think an approved order necessarily involves receiving units, and we can't know when it will be placed. And this is problematic, in the event that the order is not placed until later (for some reason) or canceled, the stock would be  increased, and users might believe during a time that the products are arriving, when it is not really the case. This might cause lates, or errors.
I think the user should only see the stock increment when the order is placed (status = 3) and when we are sure the products will be shipped.

dolibit-ut:
I believe that depending on the company and the concepts, both variants can be correct.
But in general, i would also say that after the supplier order has been placed, the status should be set to 
order accepted & ready for dispatch (=3) before the stock is added to the virtual warehouse.

eldy:
to get best of both we can replace '1,2,3,4' with $conf->global->SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK

atm-adrien:
Is this good for you ?

PR #17409, opened 2021-04-27
Update supplier_proposal.class.php

fr69400:
Some function doesn't allways return value.
Some unitialized variables

The line bellow is incorrect : $error is allways 0. 
if (!$error && (preg_match('/^[\(]?PROV/i', $this->ref) || empty($this->ref)))

May cause trouble if $soc cant be fetched, then $ref can't be get correct value => need to test $soc before

### DIFF
--- a/htdocs/supplier_proposal/class/supplier_proposal.class.php
+++ b/htdocs/supplier_proposal/class/supplier_proposal.class.php
@@ -1404,7 +1404,7 @@ class SupplierProposal extends CommonObject
 			// Numbering module definition
 			$soc = new Societe($this->db);
 			$soc->fetch($this->socid);
-			
+
 			if ($soc < 0) return -1;
 
 			// Define new ref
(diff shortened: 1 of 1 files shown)


htdocs/supplier_proposal/class/supplier_proposal.class.php
```diff
@@ -1403,8 +1405,10 @@ public function valid($user, $notrigger = 0)
 			$soc = new Societe($this->db);
 			$soc->fetch($this->socid);
 
+			if ($soc < 0) return -1;
```
eldy, line 1408: $soc is an object, it can' be compared to < 0 
May be you want to do 
```
$result = $soc->fetch($this->socid);
if ($result < 0) return -1;
```

fr69400: Oh... Yes, of course !

PR #17411, opened 2021-04-27
Fix wrong set of constant

daraelmin:
# Fix commit 035828ba511d3a77fd43a29c29d083f0a80b7179
@eldy I assume that it is for historical reason but as i van see draft status = -1 and rediliated status = 0, so this fix is here for backward compatibility


eldy:
You're right, historical reason. So constant are welcome to avoid such errors.

daraelmin:
> You're right, historical reason. So constant are welcome to avoid such errors.

I have add some commit in order to use constant

PR #17415, opened 2021-04-28
Close #17271 : Replacement of 2 graphes in /compta/index.php

### LINKED ISSUE #17271: Replace the 2 graph on page /compta/index.php?
The 2 graph invoice status of customer and  invoice status of supplier on /compta/index.php are useless because with time, the number of invoice closed or paid increases and we see only on topic.
Goal is to "comment" existing code for graph (not remove)
And then replace the 2 graph with the following graph (so only invoice, not with type "credit note" and with status validated, status billed to no):
- Invoice to be paid (due date <  30 days)
- Invoice to be paid (due date < 15 days)
- Invoice to be paid (due date < current date)
- Invoice to be paid (due date >= current date)
- Invoice to be paid (due date > 15 days)
- Invoice to be paid (due date > 30 days)

Same for supplier

Hystepik:
Modify /compta/index.php by commenting 2graphes and by creating new graph for customers invoices and supplier invoices in /core/lib/invoice.lib.php


htdocs/core/lib/invoice.lib.php
```diff
@@ -456,6 +456,121 @@ function getPurchaseInvoicePieChart($socid = 0)
...
+ * @return string A HTML table that contains a pie chart of customers or supplier invoices
+ */
+function getNumberInvoicesPieChart($mode)
+{
+	global $conf, $db, $langs, $user;
+	if (!empty($conf->facture->enabled) && !empty($user->rights->facture->lire)) {
+		include DOL_DOCUMENT_ROOT.'/theme/'.$conf->theme.'/theme_vars.inc.php';
+		$langs->load("boxes");
+		$tmpinvoice = new Facture($db);
+		$sql = "SELECT f.rowid, f.ref, f.fk_statut as status, f.type, f.total_ht, f.total_tva, f.total_ttc, f.paye, f.datef";
```
eldy, line 471: You must use the "due date" of invoice and not the "date" of invoice.

Also imagine there is 1 million of record into database, your alogrithm will return 1 million of record when all what we need is a count. You must something like that: (you must always make the most work as possible by the database engine to reduce network and memory overload).
So use something like that
```
select  
sum( $db->ifsql ( 'date_ech < '.now - 30 days.', 1, 0)) as nbbefore30, 
sum( $db->ifsql ( 'date_ech < '.now - 15 days.', 1, 0)) as nbbefore15, 

etc...
```
You will get immediatly a count for each range in one record answer.

Hystepik:
image

PR #17421, opened 2021-04-28
Look and feel supplier bill tabs 

defrance:


htdocs/fourn/facture/contact.php
```diff
@@ -174,71 +175,83 @@
...
-		if ($object->type == FactureFournisseur::TYPE_REPLACEMENT) {
+		if ($object->type == FactureFournisseur::TYPE_REPLACEMENT)
+		{
 			$facreplaced = new FactureFournisseur($db);
 			$facreplaced->fetch($object->fk_facture_source);
 			print ' ('.$langs->transnoentities("ReplaceInvoice", $facreplaced->getNomUrl(1)).')';
 		}
-		if ($object->type == FactureFournisseur::TYPE_CREDIT_NOTE) {
+		if ($object->type == FactureFournisseur::TYPE_CREDIT_NOTE)
+		{
```
eldy, line 192: Must use the PHPCS style sheet defined in  dev/setup/codesniffer/ruleset.xml 
{ must be on same line for a if (PSR 12)

defrance (commented):
corrections made

htdocs/fourn/facture/contact.php
```diff
@@ -215,30 +214,40 @@
 		}
 		print '</td></tr>';
 
+
 		// Label
-		print '<tr><td>'.$form->editfieldkey("Label", 'label', $object->label, $object, 0).'</td><td>';
-		print $form->editfieldval("Label", 'label', $object->label, $object, 0);
+		print '<tr><td>'.$langs->trans('Label').'</td>';
```
eldy, line 219: Why removing the use of ->editfieldkey
In a future all fields in form may use this so we will be able to make "inline edition".

defrance: on a undedicaced tabs i find that it's not a good way to have fields updatable
replace all the display fields by inline edition on dolibarr is an very huge work for small usage by the way
why not continue to keep it simple?

eldy:
Even if this PR goal is to fix a bug, it seems it introduces more problems or regressions than it solves, so request is closed.

PR #17430, opened 2021-04-29
Update llx_10_c_regions.sql

dolibit-ut:
Hungary - sorting & regions in hungarian


dolibit-ut:
Is it useful to add a column for english name/description and use this also for translation into other languages?
Or we make translations based on region_id ? 
But a translation at least into english would be useful to make readable for a wider range of users.

eldy:
Currently there is no system to translate region. so we keep them into the language of the country they are related to

PR #17453, opened 2021-05-01
New/NEW Product api datec tms filter

kastoras:
New Handle date value to sqlfilter

When I was trying to pass a date time value to sqlfilter (datec:<:2021-04-25 10:12:18) I was getting a SQL injection error. With the change I am suggesting, I can parse the datetime to timestamp and use it to the sqlfilter((datec:<:1616861893) and (tms:<:1616861893)).


eldy:
With v14, i do not experience  sql injection error. But the filter using hours fails. i pushed a better fix.

PR #17458, opened 2021-05-02
Create en_GB

dolibit-ut:


eldy:
With module builder, only 1 language file is created. Any other language file can be created from the modulebuilder.

PR #17461, opened 2021-05-02
Create mymodule.lang

dolibit-ut:


eldy:
No language needed into templates. They are create dynamically by module builder

PR #17469, opened 2021-05-03
Fix show table of pdf payments repports in file order

indelog:
Use scandir() rather opendir() and readdir() to get a sorted list of files.

### DIFF
--- a/htdocs/compta/paiement/rapport.php
+++ b/htdocs/compta/paiement/rapport.php
@@ -150,7 +150,7 @@ if ($year)
 		foreach ($files as $f) {
 			$tfile = $dir.'/'.$year.'/'.$f;
 			$relativepath = $year.'/'.$f;
-			if (is_file($tfile)) {
+			if (is_file($tfile) && preg_match('/^payment/i', $f)) {
 				print '<tr class="oddeven">';
 				print '<td><a data-ajax="false" href="'.DOL_URL_ROOT.'/document.php?modulepart=facture_paiement&amp;file='.urlencode($relativepath).'">'.img_pdf().' '.$f.'</a>'.$formfile->showPreview($f, 'facture_paiement', $relativepath, 0).'</td>';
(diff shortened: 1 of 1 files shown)


htdocs/compta/paiement/rapport.php
```diff
@@ -147,22 +146,17 @@
 		print '<td class="right">'.$langs->trans("Date").'</td>';
 		print '</tr>';
 
-		if (is_resource($handle))
-		{
-			while (($file = readdir($handle)) !== false)
-			{
-				if (preg_match('/^payment/i', $file))
```
eldy, line 154: This filter is removed ? Why ?

indelog: It is a forget. It is fixed.

htdocs/compta/paiement/rapport.php
```diff
@@ -147,22 +146,17 @@
...
-				{
-					$tfile = $dir.'/'.$year.'/'.$file;
-					$relativepath = $year.'/'.$file;
-					print '<tr class="oddeven">';
-					print '<td><a data-ajax="false" href="'.DOL_URL_ROOT.'/document.php?modulepart=facture_paiement&amp;file='.urlencode($relativepath).'">'.img_pdf().' '.$file.'</a>'.$formfile->showPreview($file, 'facture_paiement', $relativepath, 0).'</td>';
-					print '<td class="right">'.dol_print_size(dol_filesize($tfile)).'</td>';
-					print '<td class="right">'.dol_print_date(dol_filemtime($tfile), "dayhour").'</td>';
-					print '</tr>';
-				}
+		$files = (scandir($dir.'/'.$year));
```
eldy, line 149: What is goal of scandir compared to readdir ?
If goal is to make code simplest, a better solution is to use dol_dir_list to get all files into an array.

indelog: My initial goal was only to get a sorted report list. But since use dol_dir_list() can greatly simplify the code as much use it.

PR #17470, opened 2021-05-03
Create EORI.info

dolibit-ut:


eldy:
Can you move file into dev/resources/iso-normes

dolibit-ut:
Okay. But i can not see how to change path now. Can you move it?
Otherwise we close this and i create it new.

PS: EORI is not really a ISO-Norm - maybe we could later rename iso-normes to normes & standards

PR #17471, opened 2021-05-03
WIP NEW - Prepare database for work on type transaction

aspangaro:
Related to this post : https://www.dolibarr.fr/forum/t/regles-de-gestion-comptabilite-analytique/36500?u=aspangaro-opendsi

First Step : Data model
Second Step : Add dictionary
Third Step : Code


eldy:
If we factorize the search of the product for accounting transfer into a function, then we will be able to add specific rules to get an account according to rules.
If would be possible to overwrite return value from a hook of an external module or from a table rule in the core. So we just need a table rule.

The rules to get the accounting code will be able to rely on existing fields and can be extrafields. We don't need to add any more columns for this. The rule can rely on any field.
If we have a rule that say "Do not take the accounting account of product but take in priority the accounting code A123 when extrafield 'type of transaction' is A", we've got it.
This provide a more powerful feature. Any user will be able to define its overwrite rule even if it is not a "type of transaction" (this is very very specific to one need, too specific).
So we need a table of rule (that is able to be based on any column of invoicedet or invoicedet_extra, no need to add any more columns into invoice table invoicedet).

altairis-tof:
@eldy i wonder how the rules will be defined by the users ( i mean to keep it simple for the user), and also how they will be stored in the database; do you think about a data model for this rules storage ?

eldy:
For example a combo list of fields (for example a list with value "country", "product", "type of transaction", "myotherextrafield2", for your case you choose "type of transaction") + an input field value (for example "location", it can also be a combo list that show possible values) + an input field with accounting to use (for example you enter A123, it can also be a combo of existing accounitnig account) then a button "Add rule". So it is just like a dictionary. But we can imagine in a future a more advanced interface where we can choose combined conditions.

altairis-tof:
@eldy : we are ok for managing rules as you said and add a hook in the imminent factorized getAccountXXX function. (@aspangaro ?)
we just need to be sure if you are ok or not with that transaction type dictionnary.
I think it's ok since you refer to if in your latest comment (`for your case you choose "type of transaction"`) but we still have a doubt since this PR is still not merged ;) 
CC : @lmag

lmag:
@eldy @aspangaro 
Is it ok for dictionary integration ?

eldy:
As explained, there is no need to add a new dictionary for a so particular use case. Extrafields are enough to allow everybody to add its own properties to classify its invoice lines. 
We can't add a specific dictionary table for each very specific need when a generic solution already exists.
An enhanced feature to defined product accounting number must be based on a table of rules to define the accounting numbers based on properties of invoice, lines or extrafields of invoices so the need here can be addressed and everybody can make its own advanced rule for product account number assignement whatever is its way to categorize its lines.

PR #17481, opened 2021-05-03
[NEW] PDF Add options

BB2A-Anthony:


eldy:
Sorry, to avoid to have to many tabs with too low options, i moved all tabs with specific options for PDF into one tab called "other" (pdf_other.php) with one section for each module. 
Can you move your option addition into this new file ?

BB2A-Anthony:
> Sorry, to avoid to have to many tabs with too low options, i moved all tabs with specific options for PDF into one tab called "other" (pdf_other.php) with one section for each module.
> Can you move your option addition into this new file ?

@eldy It's done

BB2A-Anthony:
@eldy it's ok for me

eldy:
A lot of options added here are currently hidden because they are not yet ready or stable or simply not a good idea to be supplied for end users. Because there is a lot, i suggest to make a different PR for each option so it is easier to discus case by case what we can do to make the option completely available.

bb2a:
@eldy Done it's ok for me

bb2a:
@eldy conflict resolved

BB2A-Anthony:
Done it's ok for me

@eldy conflict resolved

eldy:
Other PR were introduced to add each option one by one and few of them were already merged so i closed this one.

PR #17489, opened 2021-05-04
Fix  #17475 : Add of List-Unsubscribe in mass emailing

### LINKED ISSUE #17475: Support of "List-Unsubscribe in mass emailing
Into the mass emailing module, 
when we add have the key `__UNSUBSCRIBE__` into the message, this adds a link to message to unsusbcribe.
Add also the smtp header 
"List-Unsubscribe ... with the same url"

Hystepik:
Add of List-Unsubscribe in mass email header


htdocs/core/lib/functions.lib.php
```diff
@@ -7312,9 +7313,12 @@ function make_substitutions($text, $substitutionarray, $outputlangs = null, $con
 			} else {
 				$value = dol_nl2br("$value");
 			}
-
 			$text = str_replace("$key", "$value", $text); // We must keep the " to work when value is 123.5 for example
 		}
+		if ($key == '__UNSUBSCRIBE__' && (isset($moreinheader))) {
```
eldy, line 7318: The make_substitutions() function is a technical low level method (to make a repacement of strings). We should not find code specific to one feature.

So, you must forge the string of the header
`List-Unsubscribe: <__UNSUBSCRIBE__>`
into the mailing/card.php code (because this is where you need it) and call the make_substitution on it to avec it ready to send it to the new CMailFile

Note that the mailing can also be sent using the command line script/emailings/mailing-send.php, so same action must be one inside too.

Note also that

scripts/emailings/mailing-send.php
```diff
@@ -268,6 +268,13 @@
 
 						$substitutionisok = true;
 
+						$moreinheader = '';
+						if (preg_match('/__UNSUBSCRIBE__/', $message)) {
+							$moreinheader = make_substitutions('__UNSUBSCRIBE__', $substitutionarray);
```
eldy, line 273: Why not just, in 2 lines
```
$moreinheader = "List-Unsubscribe: <__UNSUBSCRIBE__>\n";
$moreinheader = make_substitutions($moreinheader, $substitutionarray);
```

Hystepik: Yes,I did check and the link correctly shown

PR #17497, opened 2021-05-05
Fix #17491 : date extrafield filter works well now

### LINKED ISSUE #17491: Regression - Date extrafield filter in list doesn't work anymore
# Bug
filter extrafield type date in list thirdparty doesn't work anymore.
It's a regression from 12.0.X

## Environment
- **Version**: 13.0.2
- **OS**: debian9
- **Web server**: apache2
- **PHP**: 7.0.33-0+deb9u10
- **Database**: MariaDB 5.5.5-10.1.48-MariaDB-0+deb9u2
- **URL(s)**: /societe/list.php

filter should work properly

add date extrafield in thirdparty type DATE.
fill this field on multiple thirdparty
go to /societe/list.php
try to filter on the date (write it or use date picker)
nothing is returned

Hystepik:
Edit of core/tpl/extrafields_list_search_sql.tpl.php to use only Date (and not time) when extrafield of type  date is used

### DIFF
--- a/htdocs/core/tpl/extrafields_list_search_sql.tpl.php
+++ b/htdocs/core/tpl/extrafields_list_search_sql.tpl.php
@@ -23,9 +23,9 @@ if (!empty($extrafieldsobjectkey) && !empty($search_array_options) && is_array($
 
 		if ($crit != '' && in_array($typ, array('date', 'datetime', 'timestamp')))
 		{
-			if($typ == 'date'){
+			if ($typ == 'date'){
 				$sql .= " AND ".$extrafieldsobjectprefix.$tmpkey." = DATE('".$db->idate($crit)."')";
-			}else{
+			}else {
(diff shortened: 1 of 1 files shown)


htdocs/core/tpl/extrafields_list_search_sql.tpl.php
```diff
@@ -23,7 +23,11 @@
 
 		if ($crit != '' && in_array($typ, array('date', 'datetime', 'timestamp')))
 		{
-			$sql .= " AND ".$extrafieldsobjectprefix.$tmpkey." = '".$db->idate($crit)."'";
+			if ($typ == 'date'){
+				$sql .= " AND ".$extrafieldsobjectprefix.$tmpkey." = DATE('".$db->idate($crit)."')";
```
eldy, line 27: DATE function on database level must not be used : It does not share the same timezone as the PHP timezone that is mostly used by the main code.
So you must convert the $crit timestamp into a value where hour, minute and seconds are 0.
For this, you can replace  $crit  with  dol_get_first_hour($crit)

PR #17503, opened 2021-05-05
 Fix #17501 : Undownloable attached file 

### LINKED ISSUE #17501: Substitutions in GETPOST make some attached files undownloadable
# Bug
`GETPOST()` makes substitutions when the value is passed by `GET` and contains `__SOMETHING__` (where SOMETHING can be any upper-case string and may contain numbers and one underscore).
When it does not find a candidate for substitution, the whole part is just deleted from the returned value.

This is a problem with attached files because the name of the file can contain `__SOMETHING__` and will be correctly stored, but when you want to retrieve it through `/document.php`, the file name is passed by GET and the unwanted substitution happens, causing an error (`ErrorFileDoesNotExists`).

This substitution behavior may cause other problems, but this is the issue one of our clients reported.

## Environment
- **Version**: reported in 9.0, probably still present in 13.0
- **OS**: any
- **Web server**: any
- **PHP**: any
- **Database**: any

1. Create a empty file named `test__TEST__.txt` and attach it to a product (let's call the product "myproduct" for instance). The file is attached correctly.
2. Click on the link to download the file → you get *ErrorFileDoesNotExists myproduct/myproduct-test.txt* because it is looking for `myproduct-test.txt` instead of` myproduct-test__TEST__.txt`.

Hystepik:
Fix of a bug on GETPOST which impact versions 9.0 to develop.
The bug is when you upload a file with \_\_SOMETHING\_\_ it is impossible to download or delete


atm-maxime:
I think you can ask for this fix in v 11.
And the issue is also present when you rename a file and add __xxx__ in the file name :)

Hystepik:
Thanks I hadn’t thought of that

PR #17511, opened 2021-05-06
FIX Product accountancy affectation with product_perentity activated

aspangaro:


htdocs/accountancy/admin/productaccount.php
```diff
@@ -189,7 +189,7 @@
 					$db->begin();
 
 					if (!empty($conf->global->MAIN_PRODUCT_PERENTITY_SHARED)) {
-						$sql = "INSERT INTO ".MAIN_DB_PREFIX."product_perentity (fk_product, entity, '".$db->escape($accountancy_field_name)."')";
+						$sql = "INSERT INTO ".MAIN_DB_PREFIX."product_perentity (fk_product, entity, ".$db->escape($accountancy_field_name).")";
```
eldy, line 192: If you use $db->escape( it means it is a string. In such a cas, the ' ' must be present into the SQL string.
If value is an int, you must use instead   (int) $var  when forging the sql request.

htdocs/accountancy/admin/productaccount.php
```diff
@@ -189,7 +189,7 @@
...
 						$sql .= " VALUES (".((int) $productid).", ".((int) $conf->entity).", '".$db->escape($accounting->account_number)."')";
 						$sql .= " ON DUPLICATE KEY UPDATE ".$accountancy_field_name." = '".$db->escape($accounting->account_number)."'";
```
eldy, line 194: ON DUPLICATE KEY UPDATE is not portable.
Must be replaced with a select test to find record and the insert or update depending of if found or not.

PR #17521, opened 2021-05-06
Add option pdf global "AutoUpdate"

BB2A-Anthony:


eldy:
This feature is designed to make some debug or bypass some troubles in some technical error case.
What is the goal to have it available for end users ?

BB2A-Anthony:
Disable for more perform

eldy:
There is no reported performance problem reported for the moment due to document generation so this option does not seems to be useful enough to become an end user option. It generates a big degradation of user experience (with pdf that are corrupted compared to data) for a small performance enhancement that is not enough requested.

PR #17553, opened 2021-05-09
Fix Title company card

aspangaro:


aspangaro:
For travis error : https://github.com/Dolibarr/dolibarr/pull/17557

PR #17566, opened 2021-05-10
Fix bad test on THEME_ELDY_ENABLE_PERSONALIZED

daraelmin:


eldy:
Can't find how you reproduce the bug ?

daraelmin:
@eldy i was not able to personalize  darktheme

daraelmin:
I have experimented problems with darkmode whith theme_eldy_enable_personnalized

In some case some parameter are set to "brigth" by default even in darkmode. I had to search long time where cause it was not in theme_vars.inc.php file.

But it is a workaround that doesnt fix the bug. I ll try something else.

PR #17577, opened 2021-05-11
add thirdparty customer code to substitutionarray

p-payet:
# New add thirdparty customer code to substitution array for Emailing
Add thirdparty customer code, and thirdparty customer code without 'C' prefix, to substitution array for Emailing


htdocs/comm/mailing/card.php
```diff
@@ -201,6 +201,13 @@
 
 						// Array of possible substitutions (See also file mailing-send.php that should manage same substitutions)
 						$substitutionarray['__ID__'] = $obj->source_id;
+						if ($obj->source_type == "thirdparty") {
+							$thirdpartystatic = new Societe($db);
```
eldy, line 205: The new can be set outside of the loop.

PR #17512, opened 2021-05-06
Stock product card traceability fields

altairis-noe:
# Traceability fields in product card and stock product lot card
stock_product_traceability_fields_card
stock_product_traceability_fields_create
stock_product_traceability_fields_edit
stock_product_traceability_fields_lot
Using the fields created in PR #17322, this adds editable fields in the productlot card and fields in the product card. Poke @altairis-tof

PR #17582, opened 2021-05-11
Update llx_product.sql

frederic34:
https://github.com/Dolibarr/dolibarr/pull/17512
https://github.com/Dolibarr/dolibarr/pull/17322


frederic34:
`$sql .= ", lifetime = ".($this->lifetime > 0 ? (int) $this->lifetime : 'null');`
`$sql .= ", qc_frequency = ".($this->qc_frequency > 0 ? (int) $this->qc_frequency : 'null');`

PR #17587, opened 2021-05-12
fix: when only supplier module is Active top menu should be displayed

FHenry:
fix: when only supplier module is active (no customer, no propal, no customer order) top menu "Sales" should be displayed to acces to supplier order


atm-maxime:
Hi.
Shouldn't you push this fix on 11 branch ?

moreauftheobald:
> Hi.
> Shouldn't you push this fix on 11 branch ?

No, on develop branch it will be ok ... ;)

PR #17589, opened 2021-05-12
Update llx_00_c_country.sql

dolibit-ut:


dolibit-ut:
Is it useful to add a column 'label_nativ'  (or  'label_local' or 'label_national' or 'label_original') 
for the designation in the national language to show always national and english terms?

eldy:
We don't need label_nativ because the value in each language for each country is already stored into file dict.lang (value in database is just for information, not visible on interface)

PR #17602, opened 2021-05-13
FIX: Impossible to add multiple localtax2

fmarcet:
In Spain we handle various localtax2 rates. The nomenclature used is to separate them by the symbol ':'. With the parameter 'nohtml' in the GETPOST it did not work correctly since only the first rate was saved. So I have added a specific case for the localtax2.


eldy:
With v13, the GETPOST(.., 'alpha') is now same than GETPOST(..., 'alphanohtml').
So does it works on v13 too ?

PR #17604, opened 2021-05-14
NEW Split Sale in TakePOS

andreubisquerra:
Important new functionality in TakePOS. Now you can separate a sale to make a sale per customer, for example very useful in a restaurant to charge per person or to change only one person to other table.

split


eldy:
github does not allow me to fix the conflict from interface. Can you solve it and push pr again ?

htdocs/takepos/css/split.css.php
```diff
@@ -0,0 +1,116 @@
+<?php
```
eldy, line 1: Why adding a new css file just for this. The less css file we are the best it is. Can't you just merge the content into a section into an exsiting one, for example pos.css.php

PR #17610, opened 2021-05-14
Fix error in MouvementStockTest

Hystepik:
Fix of failure of MouvementStockTest by removing some fields in update product.class and productlot.class


fappels:
Did you run latest DB migration?

Hystepik:
Sorry I did forget to make the migration and it works

PR #17616, opened 2021-05-15
Fix LDAP synchronisation for User class when using posixAccount objectclass

piernov:
Bugs present in 11.0, 12.0, 13.0 and develop since the introduction of the feature in #11868 .
**PR based off the 11.0 branch but will have to be merged in newer branches as well.**

Prevents from actually using the posixAccount objectclass when using Dolibarr to LDAP User synchronization since LDAP will refuse the request with missing attributes.

# Fix default gid 65534 for User in LDAP

posixAccount objectclass (intended to be used with the uid/gid/homedir fields) requires a gid.
Always set a gid in LDAP for the User class even if the user does not belong to any group (e.g., when creating a new User).
By default 65534 which corresponds to the nobody group on major distributions (incl. Debian).

# Fix use login for User homedir in LDAP

posixAccount objectclass (intended to be used with the uid/gid/homedir fields) requires a homedir.
Always set a homedir in LDAP for the User class even if user does not have a firstname by using the login instead.
Additionally on Linux the login is typically used for the home directory rather than the first name.
It also avoids having accentuated or other special characters (commonly found in names) in the home directory path.


eldy:
Because this may have side effect on current usage, i validate the fix for the next major version only, so v14.

PR #17617, opened 2021-05-15
Fix: properly set gid for Group LDAP synchro

piernov:
Bugs present in 11.0, 12.0, 13.0 and develop since the introduction of the feature in #11868 .
**PR based off the 11.0 branch but will have to be merged in newer branches as well.**

#  Fix: properly set gid for Group LDAP synchro 

Even if the gidNumber field was set in the configuration, a gid different from the Dolibarr's one was generated and stored into LDAP.
Now the gidNumber is properly set to Dolibarr's gid if gidNumber field is set in the LDAP module configuration.

### DIFF
--- a/htdocs/core/triggers/interface_50_modLdap_Ldapsynchro.class.php
+++ b/htdocs/core/triggers/interface_50_modLdap_Ldapsynchro.class.php
@@ -282,8 +282,8 @@ class InterfaceLdapsynchro extends DolibarrTriggers
 					$info=$object->_load_ldap_info();
 					$dn=$object->_load_ldap_dn($info);
 
-					// Get a gid number for objectclass PosixGroup
-					if (in_array('posixGroup', $info['objectclass'])) {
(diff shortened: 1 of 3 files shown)


francoisPE:
I don't know why, but, in 13.0.3 (different from 13.0.2) I can't set group gidnumber from Dolibarr. It refuses...

When I test with web interface test, log says
Jun 29 23:03:04 ns1 slapd[227260]: conn=24383 op=2 ADD dn="cn=DOLIBARR GROUP SPECIMEN,ou=groups,dc=client,dc=fr"
Jun 29 23:03:04 ns1 slapd[227260]: Entry (cn=DOLIBARR GROUP SPECIMEN,ou=groups,dc=client,dc=fr), attribute 'gidNumber' not allowed
Jun 29 23:03:04 ns1 slapd[227260]: conn=24383 op=2 RESULT tag=105 err=65 text=attribute 'gidNumber' not allowed

PR #17618, opened 2021-05-15
Fix add/del user to group modifies LDAP group

piernov:
Bug present at least in 12.0, 13.0 and develop but it has probably been there for longer.

_**v2:**_
**PR based off the 12.0 branch but will have to be merged in newer branches as well.**

# Fix add/del user to group modifies LDAP group 

Adding or removing a user from a group modifies the user object on Dolibarr's side.
In LDAP however, members of a group are stored in the group itself.
Therefore group must be updated after adding/removing a user from it.
Update group in LDAP with new list of users at the end of USER_MODIFY trigger.

~~_**v1:**_~~
~~**PR based off the 12.0 branch (to at least have the correct `USERGROUP_MODIFY` trigger) but will have to be merged in newer branches as well.**~~

# ~~Fix add/del user to group trigger USERGROUP_MODIFY~~

~~Adding or removing a user from a group modifies the user object on Dolibarr's side.~~
~~In LDAP however, members of a group are stored in the group itself.~~
~~Therefore group must be updated after adding/removing a user from it.~~
~~Trigger the USERGROUP_MODIFY event from SetInGroup/RemoveFromGroup to that effect.~~

~~**Warning**: I don't know if there are side effects to calling the trigger here. I also don't know if it is the proper way to solve this problem.~~

### DIFF
--- a/htdocs/core/triggers/interface_50_modLdap_Ldapsynchro.class.php
+++ b/htdocs/core/triggers/interface_50_modLdap_Ldapsynchro.class.php
@@ -126,6 +126,52 @@ class InterfaceLdapsynchro extends DolibarrTriggers
 					$newparent = $object->_load_ldap_dn($info, 1);
 
 					$result = $ldap->update($dn, $info, $user, $olddn, $newrdn, $newparent);
+
+					if ($result > 0 && !empty($object->context['newgroupid'])) {      // We are in context of adding a new group to user
+						$usergroup = new Usergroup($this->db);
+
+						$usergroup->fetch($object->context['newgroupid']);
+
+						$oldinfo = $usergroup->_load_ldap_info();
+						$olddn = $usergroup->_load_ldap_dn($oldinfo);
+
+						// Verify if entry exist
+						$container = $usergroup->_load_ldap_dn($oldinfo, 1);
+						$search = "(".$usergroup->_load_ldap_dn($oldinfo, 2).")";
+						$records = $ldap->search($container, $search);
+						if (count($records) && $records['count'] == 0)
+						{
+							$olddn = '';
+						}
+
+						$info = $usergroup->_load_ldap_info(); // Contains all members, included the new one (insert already done before trigger call)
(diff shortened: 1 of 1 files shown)


atm-maxime:
Hi ! Thanks for contributing in Dolibarr and welcome !
For this one let's see what @eldy says, but maybe new triggers like USER_ADD_IN_GROUP / USER_REMOVE_FROM_GROUP would be better here as that's neither a user modification, nor a group modification.

eldy:
In very old versions of dolibarr (< v6), we used a trigger called USER_SETINGROUP andd USER_REMOVEFROMGROUP (code is still present and commented into htdocs/core/triggers/interface_50_modLdap_Ldapsynchro.class.php).
However, they were removed to better follow the rule: 1 Trigger for 1 CRUD events (so Create, Read, Update or Delete).

So for this need, you must keep only one trigger call, the call of USER_MODIFY (no trigger on group must be done because nothing is changed on group, we change something on links between user and groups, and choice for that was to include this into the event of modification of USER).

What you must do is modifying the code inside the trigger USER_MODIFY into the trigger of Ldap module, so into the file htdocs/core/triggers/interface_50_modLdap_Ldapsynchro.class.php.

Just add a test into USER_MODIFY part of the LDAP trigger:
```
if !empty($this->context['newgroupid']) {      // We are in context of adding a new group to user
  // call the same code that is inside the USERGROUP_MODIFY part of trigger.
}
if !empty($this->context['oldgroupid']) {      // We are in context of remobing a group from user
  // call the same code that is inside the USERGROUP_MODIFY part of trigger.
}
```

Can you fix that way ?

piernov:
Thanks, that makes sense.
I updated the PR by removing the previous changes and instead adapting the `USER_MODIFY` trigger with the code from the `USER_SETINGROUP` trigger, for both cases of adding and removing a group from a user.
The issue is that it creates a bunch of duplicate lines of code.
Each trigger handling already has some similar duplicate code so I am not sure if it is ok or not.

PR #17630, opened 2021-05-17
Fix: user rest api rights

DorianVabre:
We did not check in 12 and above, so we advise you to see especially for new endpoints.

### DIFF
--- a/htdocs/user/class/api_users.class.php
+++ b/htdocs/user/class/api_users.class.php
@@ -209,7 +209,7 @@ class Users extends DolibarrApi
     {
 	    // check user authorization
 	    if(! DolibarrApiAccess::$user->rights->user->creer && !DolibarrApiAccess::$user->admin) {
-	       throw new RestException(401, "User creation not allowed");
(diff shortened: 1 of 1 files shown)


eldy:
Permission to read user is DolibarrApiAccess::$user->rights->user->user->lire, no need to be super admin. A user with permission to read or create a user is allowed to read or create users.

DorianVabre:
> Permission to read user is DolibarrApiAccess::$user->rights->user->user->lire, no need to be super admin. A user with permission to read or create a user is allowed to read or create users.

Sorry, we did not explain clearly what we (I checked with @marc-dll about this issue) were doing.

The User API nodes don't check whether the user is an admin like user or permission pages do.
This causes an admin not to be able to read users from the API, so we modified the checks done by the API class.
Being super-admin or admin is the same in this case.

From the interface, an admin cannot add rights to any admin including himself (and by rights, I mean an entry in the table llx_user_rights). That's why we have to check if the user is admin or not in this case.

PR #17632, opened 2021-05-17
add missing printFieldListHaving hook

p-payet:
# Fix # Add missing printFieldListHaving hook for contact list
Add missing printFieldListHaving hook in htdocs/contrat/list.php


htdocs/contrat/list.php
```diff
@@ -274,6 +274,10 @@
 	elseif ($search_op2df == '>=') $sql .= " HAVING MIN(".$db->ifsql("cd.statut=4", "cd.date_fin_validite", "null").") >= '".$db->idate(dol_get_first_day($search_dfyear, $search_dfmonth, false))."'";
 	else $sql .= " HAVING MIN(".$db->ifsql("cd.statut=4", "cd.date_fin_validite", "null").") <= '".$db->idate(dol_get_last_day($search_dfyear, $search_dfmonth, false))."' AND MIN(".$db->ifsql("cd.statut=4", "cd.date_fin_validite", "null").") >= '".$db->idate(dol_get_first_day($search_dfyear, $search_dfmonth, false))."'";
 }
+// Add HAVING from hooks
+$parameters = array();
+$reshook = $hookmanager->executeHooks('printFieldListHaving', $parameters, $object); // Note that $action and $object may have been modified by hook
+$sql .= $search_dfyear > 0 && $search_op2df ? $hookmanager->resPrint : (' HAVING 1=1 ' . $hookmanager->resPrint);
```
eldy, line 280: A more standard way to implement hook is to test $reshook and execute the code into "if ($search_dfyear > 0 && $search_op2df)" only if reshook is 0. If value is 1, $sql is concatenated with $hookmanager->resPrint
So the hook can replace (if return is 1) or just complete (if return is 0). The hook decide.

PR #17639, opened 2021-05-17
look and feel : Author in lists

BB2A-Anthony:
image


atm-maxime:
Fetch in lists are forbidden, it leads to performance issues. If your list displays hundreds of records

BB2A-Anthony:
> Fetch in lists are forbidden, it leads to performance issues. If your list displays hundreds of records

@atm-maxime ok it's done

eldy:
Fetch in list is in fact "not recommended", not completely forbidden. Sometimes, we have no choice (in such a case, we can add a cache to avoid multiple similar choices). But in this case, we have the choice, so it is better to have data retreived into the select. Thanks for having fixed this.

PR #17642, opened 2021-05-17
Compatibility phpv8

TheGivrix:
*I fixed errors and navigated across pages. But I didn't see any other error. Seems like it's over*


htdocs/admin/system/modules.php
```diff
@@ -277,7 +277,7 @@
 }
 
 // Fields from hook
-$parameters = array('arrayfields'=>$arrayfields, 'param'=>$param, 'sortfield'=>$sortfield, 'sortorder'=>$sortorder);
+$parameters = array('arrayfields'=>$arrayfields, 'param'=>empty($param) ? '' : $param, 'sortfield'=>$sortfield, 'sortorder'=>$sortorder);
```
eldy, line 280: Using test on empty is not always the best way to fix this. Here warning appears because we use a var that is not initialized. A best fix is to add initialization, so add a
`$param = '';`

Another good solution here is to simply remove this param. Because we provide a parameter that does not exists and is never defined into page, so we have no reason to propagate it into the $parameters array.

So here i would simply do:
`$parameters = array('arrayfields'=>$arrayfields, 'sortfield'=>$sortfield, 'sortorder'=>$sortorder);`

I prefer first solution (intialize $param) because it matches the template (the modulebuilder skeleton example) and in future we may add value inside $param

htdocs/admin/tools/listevents.php
```diff
@@ -245,7 +245,7 @@
 	if ($limit > 0 && $limit != $conf->liste_limit) {
 		$param .= '&limit='.urlencode($limit);
 	}
-	if ($optioncss != '') {
+	if (!empty($optioncss)) {
```
eldy, line 248: Same here. Following the template htdocs/modulebuilder/template/myobject_list.php
a better fix would be to initialize the $optioncss like it is into the example skeleton so adding at top of page
`$optioncss = GETPOST('optioncss', 'aZ'); // Option for the css output (always '' except when 'print')`

PR #17651, opened 2021-05-18
FIX filter by supplier on replenish

lvessiller-opendsi:
FIX filter by supplier on replenish
- when you want to replenish, and you filter by supplier, the stock of product can be computed several times if the product has several buying prices (same supplier)
- so when you filter on "alerts only" this product can disappear of theses results whereas the same product has a stock alert (in stock tab of product card)

**In replenish with no filter**
image

**In replenish filter by supplier (no result)**
image

**In stock tab of product card**
image


eldy:
Thanks for analysis and suggestion.
I pushed another way to fix that i think should have a better execution plan (so should run faster). 
Method seems also better because what we want is just a condition and not more data from other table so having the test into the where seems better than adding a join.
Also it solves a missing condition on entity on the table product_fournisseur_price.
Thanks to reopen if it does not work for you.

Issue #15707, opened 2020-12-08
Rest API tickets: PUT tickets doesn't written in DB


matDOTviguier:
# Bug
No API POST on tickets have no effect in Database at all.

## Environment
Seem to be environment independant

## Expected and actual behavior
Expected, a new row in ticket table. Actual behavior, none. Get a 500 error.

## Steps to reproduce the behavior
Explore the API once connected, send the data by : 

{
    "subject": "S1",
    "message": "M1"
  }

```
2020-12-08 16:19:34 INFO    127.0.0.1   --- Access to /api/index.php/tickets - action=, massaction=
2020-12-08 16:19:34 DEBUG   127.0.0.1   sql=SELECT transkey, transvalue FROM llx_overwrite_trans where lang='fr_FR' OR lang IS NULL AND entity IN (0, 1) ORDER BY lang DESC
2020-12-08 16:19:34 INFO    127.0.0.1   Load a dedicated API file moduleobject=tickets moduledirforclass=ticket
2020-12-08 16:19:34 INFO    127.0.0.1   Search api file /ticket/class/api_tickets.class.php => dir_part_file=/var/www/html/www.site.fr/htdocs/ticket/class/api_tickets.class.php classname=Tickets
2020-12-08 16:19:34 DEBUG   127.0.0.1   sql=SELECT u.login, u.datec, u.api_key,  u.tms as date_modification, u.entity FROM llx_user as u WHERE u.api_key = '_KEY_SNIPPED_'
[...]

PR #17665, opened 2021-05-18
Fix issue when loading external libs (pb with autoloader conflict and luracast)

frederic34:
Discussion here https://github.com/Luracast/Restler/pull/515
FR Forum here https://www.dolibarr.fr/forum/t/autoloader-et-librairie-externe/36722

### DIFF
--- a/htdocs/includes/restler/framework/Luracast/Restler/AutoLoader.php
+++ b/htdocs/includes/restler/framework/Luracast/Restler/AutoLoader.php
@@ -263,7 +263,7 @@ class AutoLoader
      * @return bool false unless className now exists
      */
     private function loadLastResort($className, $loader = null) {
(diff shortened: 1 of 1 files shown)


frederic34:
https://github.com/Dolibarr/dolibarr/issues/15707

In V10
https://github.com/Dolibarr/dolibarr/blob/2637f9af6334651da3c98b6d698fab8d374d530f/htdocs/includes/restler/framework/Luracast/Restler/AutoLoader.php#L265-L272

eldy:
Still don"t understand why this fix (using SORT_REGULAR instead of SORT_STRING) fixes something. Can you explain ?
This may make one class file used in priority of another, but this may generate a conflict in the other side the the other that passed after in the sort order.

frederic34:
@eldy I also got the problem using api and swiftmailer, by example with ticket create with api

PR #17669, opened 2021-05-19
NEW: autocomplete account group when editing an account card

atm-florianm:
# NEW: autocomplete for accounting account groups
This is just a `<datalist>` for the `<input>` of the "pcg_type" (it uses suggestions from the dictionary of account groups).


aspangaro:
Sorry but you are mixing two different features.

Account classes (pcg_type): https://www.compta-facile.com/plan-des-comptes-en-comptabilite/

and

The categories of an account to create a custom report (In France: un compte de résultat / bilan / solde intermédiaires de gestion).

The best solution would be to create a new dictionary for the account classes (pcg_type) like llx_c_account_classes but this will require updating all available charts of accounts...

atm-florianm:
Thank you for your reply. It seems I was mislead by existing data (in the client database I used, the existing values for `pcg_type` happened to match those in the account classes dictionary, which made me believe they were the same).

I will need some time to read up on the matter (thanks for the link).

Maybe instead of a dictionary we could simply autosuggest existing values, changing the SQL query to this (simplified) and adapting the code accordingly ?
```sql
SELECT DISTINCT pcg_type FROM llx_accounting_account
```

aspangaro:
@atm-florianm You're right, it's a better approch with autosuggest. No need to update all chart of accounts already exist.

eldy:
+1 for autosuggesting from existing values

atm-florianm:
I have reworked the proposed addition to only autosuggest values that already exist in the currently selected chart of accounts.

PR #17677, opened 2021-05-19
NEW import force checkbox in import files

lvessiller-opendsi:
NEW import force checkbox in import files
- add a checkbox to "force import data" even if wa have errors

image

This checkbox must be checked to force insert data even if we have errors.
It can be usefull to import a lot of datas with several lines in errors (we know but we don't want to fix all in import file)


atm-maxime:
What happens if there's an error on a line, is it totally ignored ? If not and the error is about a required field ?
I think a more visible warning should be present to explain what happens when data are forced to be imported

dolibit-ut:
Hello, if i understand correctly, all data will then be imported, even the incorrect data - right?
But maybe it would make more sense to skip the lines with errors - if that is possible.

eldy:
What mean "Force if error" ? You try to make insert despite  of this ? 
or does it means the label of option is "Force import by discarding errors" ?

lvessiller-opendsi:
Ok you're right I will review this to skip line in errors

lvessiller-opendsi:
@eldy 
I fixed it, and you insert only lines with no error.
And if there is at least one error : the after import request is'nt launched

htdocs/imports/import.php
```diff
@@ -1958,7 +1985,10 @@
 
 	//var_dump($array_match_file_to_database);
 
-	$db->begin();
+	$run_after_import = true;
+	if (empty($import_force)) {
+		$db->begin();
```
eldy, line 1990: begin and commit/rollback seems not correctly balanced.
If we have this condition on begin, we should also found it for both the rollback and the commit.

htdocs/imports/import.php
```diff
@@ -2003,29 +2044,35 @@
...
 				}
 			}
-		}
 
-		if (!$error) {
-			$db->commit(); // We can commit if no errors.
-		} else {
-			$db->rollback();
+			if (!$error) {
+				$db->commit(); // We can commit if no errors.
```
eldy, line 2072: commit and rollback are inside a condition where begin is inside another path

lvessiller-opendsi:
@eldy 
I reviewed this pull request and you can force to import datas only if there isn't sql to run after import.

lvessiller-opendsi:
@eldy 
I reviewed this pull request, could you see if it can be integrated ?

eldy:
When you say 
"I reviewed this pull request and you can force to import datas only if there isn't sql to run after import."
Do you mean
"I reviewed this pull request and you can force to import datas only if there isn't sql error to run after import" => In such a case why having this option
or
"I reviewed this pull request and you can force to import datas only if there is at least one sql error to run after import" => In such a case the process can continue and import success line and discard errors.

lvessiller-opendsi:
@eldy
In some cases you can have SQL requests to run after the import and in this case you keep the last method to import datas without the force option. And if you don't have SQL resquests to run after the import, you have an option to force this import : so you import only succes lines and discard errors.

eldy:
Sorry i still don't understand the goal of this option.
If a line generates a sql error during insert, whatever are the option you choosed, it is not possible to import it.
Can you provide an example of error you want to bypass with this option ?

lvessiller-opendsi:
@eldy 
For example, you want to import sales representatives in companies. You have a lot of lines in CSV file (serveral thousands).
If link between sale reprensentative and company exists, you got this kind of error : 
 Ligne 51
    > Duplicate entry '11219-5' for key 'llx_societe_commerciaux.uk_societe_commerciaux'

And if you have a lot of errors like this, you got this final message :
'Il y a encore 2852 autres lignes en erreur mais leur affichage a été limité.'
And you can't import lines with no error.

I would like to import CSV lines with no error and assign the sale representative to the company.

eldy:
So example here is to manage duplicate record.
When you import a duplicate it probably means you want to update instead of insert. Why don't you enable the property "Field for update if already found", so you will be able to avoid errors even if record is twice.

Also note that if you have an error " Duplicate entry '11219-5' for key '", the commit will fails with some database setup, and it will even always fails with pgsql (default setup is that a transaction is always canceled when a sql error occurs, even if you make a commit at end) making this PR not working.

PR #17703, opened 2021-05-22
No try to translate label field in dictionary for new

tuxgasy:
In honorific titles, the label field is hidden when the user interface is in french (and some other languages).

### DIFF
--- a/htdocs/admin/dict.php
+++ b/htdocs/admin/dict.php
@@ -2015,13 +2015,13 @@ function fieldList($fieldlist, $obj = '', $tabname = '', $context = '')
 			print '<td class="'.$classtd.'">';
 			$transfound = 0;
 			$transkey = '';
-			if (in_array($fieldlist[$field], array('label', 'libelle')) and !empty($obj->code))		// For label
+			if (in_array($fieldlist[$field], array('label', 'libelle')))		// For label
 			{
 				// Special case for labels
-				if ($tabname == MAIN_DB_PREFIX.'c_civility') {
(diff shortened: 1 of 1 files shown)


htdocs/admin/dict.php
```diff
@@ -2015,7 +2015,7 @@ function fieldList($fieldlist, $obj = '', $tabname = '', $context = '')
 			print '<td class="'.$classtd.'">';
 			$transfound = 0;
 			$transkey = '';
-			if (in_array($fieldlist[$field], array('label', 'libelle')))		// For label
+			if (in_array($fieldlist[$field], array('label', 'libelle')) and !empty($obj->code))		// For label
 			{
 				// Special case for labels
 				if ($tabname == MAIN_DB_PREFIX.'c_civility') {
```
eldy, line 2021: I think test 
`&& !empty($obj->code)`
should be added here instead.

tuxgasy: Yes but all label translations are based on `$obj->code`. So we can have the same issue with other dictionary.

tuxgasy: @eldy do I have to review my PR ?

eldy: Yes, but for some other $obj->code may simply not exists.
So test should be added on the 2 that are concerned, so the 2 with the if ($tabname == ...

tuxgasy: Ok, I changed it.

PR #17710, opened 2021-05-25
Fix #17650  : fix of test PdfDocTest

### LINKED ISSUE #17650: Error on test PdfDocTest
# Bug
When I run `PdfDocTest::testPdfDocGetLineDesc` test in my localhost it returns:
```PdfDocTest::testPdfDocGetLineDesc A product with ref PIDRESS must exists into database```

## Environment
- **Version**: [*14 develop branch*]
- **OS**: [*Windows 10*]
- **Web server**: [*xampp - apache*]
- **PHP**: [*7.4*]
- **Database**: [*MariaDB*]

### PR DESCRIPTION
**Pull request by Hystepik**
Fix of errors in test PdfDocTest

### DIFF
--- a/test/phpunit/PdfDocTest.php
+++ b/test/phpunit/PdfDocTest.php
@@ -141,13 +141,14 @@ class PdfDocTest extends PHPUnit\Framework\TestCase
 		$db=$this->savdb;
 
 		$localproduct=new Product($this->savdb);
-		$result = $localproduct->fetch(0, 'PIDRESS');
+		$localproduct->initAsSpecimen();
+		$result = $localproduct->fetch(0, 'PRODUCT_SPEC');
 		if ($result < 0) {
-			print "\n".__METHOD__." Failed to make the fetch of product PIDRESS. ".$localproduct->error; die(1);
+			print "\n".__METHOD__." Failed to make the fetch of product PRODUCT_SPEC. ".$localproduct->error; die(1);
 		}
 		$product_id = $localproduct->id;
-		if ($product_id <= 0) {
-			print "\n".__METHOD__." A product with ref PIDRESS must exists into database"; die(1);
+		if ($product_id < 0) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: test/phpunit/PdfDocTest.php [hunk 1]
```diff
@@ -141,13 +141,14 @@ public function testPdfDocGetLineDesc()
 		$db=$this->savdb;
 
 		$localproduct=new Product($this->savdb);
-		$result = $localproduct->fetch(0, 'PIDRESS');
+		$localproduct->initAsSpecimen();
```
eldy, line 144: Why a initAsSpecimen() to set dummy properties if just after there is a fetch to load real data from database ?

Hystepik: Sorry I tried something but forgot to delete it

test/phpunit/PdfDocTest.php
```diff
@@ -141,13 +141,13 @@ public function testPdfDocGetLineDesc()
...
-		$result = $localproduct->fetch(0, 'PIDRESS');
+		$result = $localproduct->fetch(0, 'PRODUCT_SPEC');
 		if ($result < 0) {
-			print "\n".__METHOD__." Failed to make the fetch of product PIDRESS. ".$localproduct->error; die(1);
+			print "\n".__METHOD__." Failed to make the fetch of product PRODUCT_SPEC. ".$localproduct->error; die(1);
 		}
 		$product_id = $localproduct->id;
-		if ($product_id <= 0) {
-			print "\n".__METHOD__." A product with ref PIDRESS must exists into database"; die(1);
+		if ($product_id < 0) {
```
eldy, line 149: Here the goal of test is to ensure we have some data into database that is ready to make the phpunit test.
If product PIDRES does not exists it returns an error.
If you test < 0, you will not have error it it does not exists and the test later with goel to check that the REF or product is correctly inserted into the label is not done.

Here there is no really fix to do except asking the user that run the test to have the correct data ready in its database (or enhance the phpunit test by adding the product before making the test)

PR #17712, opened 2021-05-25
Fix #17648 : Filter extrafield not working in contact/address

### LINKED ISSUE #17648: Filter extrafield type date in list contact/address earlier than 01/01/1970 doesn't work
# Bug
filter extrafield type date in list contact/address older than 31/12/1969 doesn't work.

## Environment
Version: 13.0.2
OS: AWS Linux 2
Web server: NGINX 1.18.0
PHP: 7.3.27
Database: MariaDB 10.5.9
Dolibarr > Setup > Display > Default language: English (united Kingdom) (en_GB)
URL(s): contact/list.php

Date filter should work properly even filter earlier date than 01/01/1970.

add date extrafield in contact/address type DATE.
fill this field on multiple contact/address
go to /contact/list.php
try to filter on the date earlier than 01/01/1970  (write it or use date picker)
nothing is returned
Filter is working for date after 01/01/1970.

Hystepik:
Fixing an bug where extrafields where not saved in contact/address section and which induce a bug in extrafield filter

### DIFF
--- a/htdocs/contact/card.php
+++ b/htdocs/contact/card.php
@@ -301,6 +301,38 @@ if (empty($reshook)) {
 		}
 	}
 
+	// Update extrafields
+	if ($action == "update_extras" && !empty($permissiontoadd) || $action == 'update' && empty($cancel) && $user->rights->societe->contact->creer) {
+		$object->fetch(GETPOST('id', 'int'));
+
+		foreach ($object->array_options as $attributekey => $value) {
+			if (GETPOSTISSET($attributekey)) {
+				if (GETPOSTISSET($attributekey.'day') && GETPOSTISSET($attributekey.'month') && GETPOSTISSET($attributekey.'year')) {
+					// This is properties of a date
+					$object->array_options[$attributekey] = dol_mktime(GETPOST($attributekey.'hour', 'int'), GETPOST($attributekey.'min', 'int'), GETPOST($attributekey.'sec', 'int'), GETPOST($attributekey.'month', 'int'), GETPOST($attributekey.'day', 'int'), GETPOST($attributekey.'year', 'int'));
+					//var_dump(dol_print_date($object->array_options['options_'.$attributekey]));exit;
+				} else {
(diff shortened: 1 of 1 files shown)


htdocs/contact/card.php
```diff
@@ -105,6 +106,8 @@
 if ($reshook < 0) {
 	setEventMessages($hookmanager->error, $hookmanager->errors, 'errors');
 }
+// Actions cancel, add, update, delete or clone
+include DOL_DOCUMENT_ROOT.'/core/actions_addupdatedelete.inc.php';
```
eldy, line 110: Because we already have the code of action "inline" into the contact/card.php file, why adding the 
`include DOL_DOCUMENT_ROOT.'/core/actions_addupdatedelete.inc.php';`
We must choose one or the other.
Because we have specific code, we should keep working with the exisiing code
`if ($action == 'add' && $user->rights->societe->contact->creer) {`

htdocs/contact/card.php
```diff
@@ -441,7 +429,7 @@
 
 					$object->old_lastname = '';
 					$object->old_firstname = '';
-					$action = 'view';
+					$action = 'update_extras';
```
eldy, line 432: I think here you set update_extras to make the update of product in 2 steps. Firs you update the product and then later, in the if $action == "update_extras" you will update extra fields (so you reuse the controller code added to manage the update of extrafields alone). But there is a problem. If backtopage is defined, there is a redirect before we reach the part that update the update_extras. Also this means we update the extrafields into another transaction than the main properties. Also it means that the trigger inside the update of product used by external module will be run on a product that is not yet completely updated.
For this reason, you must keep the code juste after the comment
// Fill array 'array_options' with data from add form
So extrafields is filled and the call of  $object->update  will both update comm properties and extrafields (in same transaction, including trigger)

I did not test but i think your add 
if ($action == "update_extras" && !empty($permissiontoadd)) {
should be enough to allow edition of extrafields when we update them alone.

PR #17732, opened 2021-05-26
FIX: postgresql: rebuild sequences at install/upgrade/repair 

marc-dll:
Contrary to MySQL, PostgreSQL doesn't recompute a `SERIAL` sequence (equivalent to `AUTO_INCREMENT`) when an `INSERT` query declares it explicitly. This happens for example in some dictionary tables, like in the script `htdocs/install/mysql/data/llx_c_chargesociales.sql`.

As a consequence causes multiple `DB_ERROR_RECORD_ALREADY_EXISTS` (as many as the first available unused value) when `INSERT`ing in such tables after the offset (the sequence advances even when the record already exists).

The necessary function to correctly "reboot" the sequence was created but never called (except in v14). It should be, at least at install, upgrade and repair.

If this is validated, I will propose subsequent PRs for the migration script in version 12 and later.

### DIFF
--- a/htdocs/install/mysql/data/llx_zzz_install_end.sql
+++ b/htdocs/install/mysql/data/llx_zzz_install_end.sql
@@ -1,0 +1,21 @@
+-- Copyright (C) 2021      Marc de Lima Lucio   <68746600+marc-dll@users.noreply.github.com>
+--
+-- This program is free software; you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as published by
+-- the Free Software Foundation; either version 3 of the License, or
+-- (at your option) any later version.
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


eldy:
Because this touch on structure of data, it is better to include this in develop branch. I tried but there is a bug conflict with develop. Can you help me to fix conflict ?

marc-dll:
This has no effect on the data structure (no new table / column / index / function / trigger), it only changes the data (the current value of the serial sequence), so I don't understand why this should only be included in develop.

This is needed particularly at the end of the migration script for all currently maintained versions.

eldy:
The sequence (their values) are part of the structure. Changing them may generate critical issues in case of bug or regression. A complete validation must be done for this, so with a beta that validate non regression.
Also this touch to migration and there is no migration into a minor version, they are validated and approved on real databases only during major versions. So this must be introduced into a beta (so develop).

FHenry:
@marc-dll could you rebase on 13.0 ?

eldy:
dol_util_rebuild_sequences() has been introduced into the repair.php tool with mode=rebuild_sequences

PR #17740, opened 2021-05-27
Update llx_c_email_templates.sql

dolibit-ut:
- formatting
- removed *tms* at eventorganization


dolibit-ut:
think tms is not useful for template

dolibit-ut:
Is the field 'enabled' only necessary when assigning the template to a module?

eldy:
> Is the field 'enabled' only necessary when assigning the template to a module?

enabled field can be used for a lot of purpose. It decides when when the template must be ignored as if it does not exists.

PR #17741, opened 2021-05-27
Update llx_c_chargesociales.sql

dolibit-ut:
idea:  fk_pays at first  - id as second (based on fk_pays)


dolibit-ut:
my idea: fk_pays at first - id as second (based on fk_pays)
agree?
why are fk_pays values here in ' ' touching marks?

eldy:
Why not for id of country first.
No reason to have the quote on id of country. We can have 2 instead of '2'

PR #17747, opened 2021-05-27
fix: fixing mess done with partially merge of #17322

FHenry:
fix: fixing mess done with partially merge of #17322

Migration 13.0.0_14.0.0 is now clean with column creation from llx_product.sql and llx_product_lot.sql


htdocs/install/mysql/migration/13.0.0-14.0.0.sql
```diff
@@ -72,6 +72,8 @@ INSERT INTO llx_accounting_system (fk_country, pcg_version, label, active) VALUE
 ALTER TABLE llx_product_lot ADD COLUMN eol_date datetime NULL;
 ALTER TABLE llx_product_lot ADD COLUMN manufacturing_date datetime NULL;
 ALTER TABLE llx_product_lot ADD COLUMN scrapping_date datetime NULL;
+ALTER TABLE llx_product_lot ADD COLUMN commissionning_date datetime NULL;
```
eldy, line 75: This fields has been voluntary discarded (its use and position in this table songs wrong)

htdocs/install/mysql/migration/13.0.0-14.0.0.sql
```diff
@@ -72,6 +72,8 @@ INSERT INTO llx_accounting_system (fk_country, pcg_version, label, active) VALUE
...
+ALTER TABLE llx_product_lot ADD COLUMN qc_frequency int;
```
eldy, line 76: Same here, for the moment this field must be first into llx_product.

FHenry: The problem is that you add it into llx_product.sql but not in migration files

eldy: Yes they are, line 423

PR #17750, opened 2021-05-27
Fix #17743 newpayement source membersubscription is member

### LINKED ISSUE #17743: Bug regression membersubscription payment link
# Bug regression membersubscription payement link doesn't work with unik key

@eldy  I'm mot sure why but I've got a "Bad value for key." now when I set  PAYMENT_SECURITY_TOKEN_UNIQUE=1. 

If PAYMENT_SECURITY_TOKEN_UNIQUE is set to 0 all works fine. I assume it comes from this commit: https://github.com/Dolibarr/dolibarr/commit/2caaa52e855e36576d5fbb0059f5e1e82a546ce7

It seems ok if the source in url is "member" and not "membersubscription" but for retro compatibility it should accept both.

daraelmin:
Fix #17743 newpayement source membersubscription is not valid anymore

@eldy I'm not sure, but the bug may still occure with external module referring to membersubscription instead of member.

### DIFF
--- a/htdocs/core/lib/payments.lib.php
+++ b/htdocs/core/lib/payments.lib.php
@@ -158,7 +158,7 @@ function getValidOnlinePaymentMethods($paymentmethod = '')
 /**
  * Return string with full online payment Url
  *
- * @param   string	$type		Type of URL ('free', 'order', 'invoice', 'contractline', 'membersubscription' ...)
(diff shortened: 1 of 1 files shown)


eldy:
The code has been modified to be able to deal both 'member', and 'membersubscription' for backward compatibility (I think)

PR #17759, opened 2021-05-28
Fix inventory/inventory.php

Hystepik:
Fix inventory/inventory.php of negative values in realqty field


htdocs/product/inventory/inventory.php
```diff
@@ -251,7 +254,10 @@
 			$error++;
 			setEventMessages($langs->trans("ErrorFieldRequired", $langs->transnoentitiesnoconv("Product")), null, 'errors');
 		}
-
+		if (GETPOST('qtytoadd') < 0) {
```
eldy, line 257: Due to some countries that use special char for decimal separator, GETPOST('qtytoadd') can be negative but detected as 0 when a string. So you must make the test on 
if (price2num(GETPOST('qtytoadd'), 'MS') < 0) {

PR #17764, opened 2021-05-28
Fix of a bug on mass mailing

Hystepik:
Fix of a bug on mass mailing
When send a mass mailing the status in view isn't updated so it's still on validated status and not send status


htdocs/comm/mailing/card.php
```diff
@@ -677,6 +677,7 @@
 $form = new Form($db);
 $htmlother = new FormOther($db);
 
+$object->fetch($id);
```
eldy, line 680: Because the fetch is already done at begining, it would be better to make this fetch only in the case of we need a refresh/reload.

PR #17768, opened 2021-05-29
BUG Customer without VAT in TakePOS

andreubisquerra:


fappels (commented):
lib functions get_default_tva() and get_default_npr() are made for this

htdocs/takepos/invoice.php
```diff
@@ -503,6 +503,9 @@ function fail($message)
 	$localtax1_tx = get_localtax($tva_tx, 1, $customer, $mysoc, $tva_npr);
 	$localtax2_tx = get_localtax($tva_tx, 2, $customer, $mysoc, $tva_npr);
 
+	// Customer without VAT
+	if ($customer->tva_assuj==0) $tva_tx=0;
```
eldy, line 507: If customer X has ->tva_assuj, it means this thirdparty X will not include VAT when he will make an invoice (so it is designed for use with vendor invoices).

PR #17778, opened 2021-05-31
Fix: Do a strict comparison when check if stock is enough

tuxgasy:
# Fix bug

If a product with 2 batchs like 59 and 59c, ship the product with batch 59c will fail if the stock is not enough for 59.

### DIFF
--- a/htdocs/product/stock/class/mouvementstock.class.php
+++ b/htdocs/product/stock/class/mouvementstock.class.php
@@ -355,7 +355,7 @@ class MouvementStock extends CommonObject
 				$qtyisnotenough = 0;
 				foreach ($product->stock_warehouse[$entrepot_id]->detail_batch as $batchcursor => $prodbatch)
 				{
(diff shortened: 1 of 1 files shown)


htdocs/product/stock/class/mouvementstock.class.php
```diff
@@ -355,7 +355,7 @@ public function _create($user, $fk_product, $entrepot_id, $qty, $type, $price =
 				$qtyisnotenough = 0;
 				foreach ($product->stock_warehouse[$entrepot_id]->detail_batch as $batchcursor => $prodbatch)
 				{
-					if ($batch != $batchcursor) continue;
+					if ($batch !== $batchcursor) continue; // Do a strict comparison because $batchcursar can be an integer
```
eldy, line 358: What do you thin of this fix that seems better (we want a string comparison whatever is type and not a type comparison):

if ((string) $batch !== (string) $batchcursor)

PR #17784, opened 2021-06-01
NEW/FIX : Allow auto picking for child warehouse and taking into account of the quantities already affected in list

john-botella:
This is new feature that fixes Dolibarr behavior when creating a shipment.
When a shipment contains the same product several times (on several lines) then the quantities displayed on the next lines do not take into account the quantities already assigned (batch number included)

When a warehouse is selected, also take into account children warehouses.

----------------------------------------------------------

Voici une nouvelle fonctionnalité qui corrige un comportement de Dolibarr sur le formulaire lors de la création d'une expédition.
En effet lors de la création d'une expédition contenant plusieurs fois un même produits (sur plusieurs lignes) alors les quantities affichées sur les lignes d'après ne prennent pas en compte les quantités déjà affectées (n° de lot inclus)

Lorsqu'un entrepôt est sélectionné, prendre en compte aussi les entrepôts enfants.

### DIFF
--- a/htdocs/expedition/card.php
+++ b/htdocs/expedition/card.php
@@ -1021,10 +1021,9 @@ if ($action == 'create') {
 			$object->loadExpeditions();
 
 
-            $alreadyQtyBatchSetted = $alreadyQtySetted = array();
+			$alreadyQtyBatchSetted = $alreadyQtySetted = array();
 
-            if ($numAsked)
-            {
+			if ($numAsked) {
 				print '<tr class="liste_titre">';
 				print '<td>'.$langs->trans("Description").'</td>';
 				print '<td class="center">'.$langs->trans("QtyOrdered").'</td>';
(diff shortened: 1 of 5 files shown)


fappels:
Hello @atm-john , maybe better to move sorts into product->load_stock() with new parameter to ask for sort?

john-botella:
Hi @fappels, good idea.

eldy:
What do you mean with "When a shipment contains the same product several times (on several lines) then the quantities displayed on the next lines do not take into account the quantities already assigned (batch number included)". Can you put a screenshot ?

john-botella:
@eldy I created an exemple :

This kind of situation usually happens on the documents of construction companies because they groups their lines by block of work. It is then common to find lines common to each block.

Without patch we can see qty dispatch mistake :
image

With patch all qty are correctly dispatched : 
image

a simple screen for the tooltip :
image

PR #17785, opened 2021-06-01
Update for eldy theme modernize

BB2A-Anthony:
### **Befor**

image

image

image

### **After**
image

image

image


daraelmin:
Please use a variable to add a new personnalized color . 
It is a simpler way to add pour preference without changing the whole look and feel for everyone and it is better to be able to manage darkmode.

BB2A-Anthony:
> Please use a variable to add a new personnalized color .
> It ils a simpler way to add pour preference withput changing the whole look ans feel d'or everyone and it ils better to ne able to manage darkmode.

Ok good option
it can give something original

daraelmin:
Thanks
Sorry for all the typos in my first comment, I wrote it from my phone with spelling correction enable... for french...

eldy:
Can you provide a screenshot of result with default theme eldy with no custom setup. what happen with the shadow (with the default theme/setup eldy, there is no difference between box and background, so adding a shadow may look strange. Do you understand what i mean ?

PR #17789, opened 2021-06-01
NEW Add field date from/to in customer payment list

aspangaro:
image


htdocs/compta/paiement/list.php
```diff
@@ -211,10 +209,15 @@
 	}
 
 	// Search criteria
-	$sql .= dolSqlDateFilter("p.datep", $day, $month, $year);
 	if ($search_ref) {
 		$sql .= natural_search('p.ref', $search_ref);
 	}
+	if ($search_date_start) {
```
eldy, line 215: We must always keep the detail of the 3 component (day, month, year) for search component so we shoul dhave
if ($search_date_startday) {

if ($search_date_startmonth) {
..

htdocs/compta/paiement/list.php
```diff
@@ -275,6 +278,8 @@
 }
 $param .= (GETPOST("orphelins") ? "&orphelins=1" : '');
 $param .= ($search_ref ? "&search_ref=".urlencode($search_ref) : '');
+$param .= ($search_date_start ? "&search_date_start=".urlencode($search_date_start) : '');
```
eldy, line 281: We must always keep the detail of the 3 component (day, month, year) when forgin url params for search component so we should have
if ($search_date_startday) {

if ($search_date_startmonth) {
..

aspangaro: Already fix

PR #17790, opened 2021-06-01
NEW Add field date from/to in supplier payment list

aspangaro:


htdocs/fourn/paiement/list.php
```diff
@@ -254,14 +259,11 @@
 if ($search_ref) {
 	$param .= '&search_ref='.urlencode($search_ref);
 }
-if ($saerch_day) {
-	$param .= '&search_day='.urlencode($search_day);
-}
-if ($saerch_month) {
-	$param .= '&search_month='.urlencode($search_month);
+if ($search_date_start) {
```
eldy, line 262: We must always keep the detail of the 3 component (day, month, year) for search component so we shoul dhave
if ($search_date_startday) {

if ($search_date_startmonth) {
..

Issue #14154, opened 2020-06-30
Dolibarr 12.0.0 (Sourceforge) - Unable to rename group


truefriend-cz:
# Bug
Unable to rename group

## Environment
- **Version**: 12.0.0 from Sourceforge
- **OS**: Windows
- **Web server**: Apache
- **PHP**: 7.4
- **Database**: MySQL

## Expected and actual behavior
**Expected:** after renaming the group, it should appear in the group list under the new name
**Actual:** The group can be renamed, but when I view the list of groups, it is still under the old name, and you can see in other places that the group has not been renamed

## Steps to reproduce the behavior
1. Home -> Users & Groups -> New group
2. Name new group "Test rnd4200" and click on CREATE GROUP button and assign one user to this group (emplyoment)
3. Home -> Users & Groups -> List of groups
4. Click on "Test rnd4200" name group
5. Click on button MODIFY
6. Name of group change to "Test rnd654" and click on button SAVE
7. In popup notify show box "Group modified" and on page is new grup name (Test rnd654)
8. Home -> Users & Groups -> List of groups
9. The renamed group has the old name and the group was not renamed
Other places in Dolibarr also show the old name of the group.

wyattmcgehee:
I'm on 12.0.1, just upgraded from 10.x, I have the same problem.

Cannot rename an existing group.

After saving, there is a confirmation popup, acknowledging the save, and the Group Card looks like it has been updated. As soon as the Group Card is reloaded, the name is reverted.

Looks like this is a side-effect of #13861 / 86e571323355b8aae91407e4521a1c6b5005398a

Looking at a current MySQL dump, my 'llx_usergroup' table has a 'nom' key, but not a 'name' key, which means we both probably still need this:
https://github.com/Dolibarr/dolibarr/blob/8cd4636de7ee5e59f117bb43a5f00beb8b787fa1/htdocs/user/group/card.php#L208

I haven't figured out why
https://github.com/Dolibarr/dolibarr/blob/8cd4636de7ee5e59f117bb43a5f00beb8b787fa1/htdocs/user/group/card.php#L133
hasn't seem to have affected the add group function, while the same line of code flubs the update group function.

dpriskorn:
This should be fixed in the next major version and nom should be renamed by the migration script IMO.

ksar-ksar:
Hello,

Could you check on V13 ?

gmilad:
Hello,

[...]

PR #17793, opened 2021-06-02
update ->nom field if different from ->name

atm-greg:
# Fix Impossible update for group label
On usergroup card, we can modify label and description of a group.

When the update is launched, the value of the label input populate the "name" field
The object calls on update_common, which is based on "fields" array field.
There is no "name" field in this array. Just "nom", so it won't update with the user entry...


atm-maxime:
Fix #14154

PR #17794, opened 2021-06-02
Fix jointure cabyprod

atm-quentin:
# Fix join cabyprod
join of categ and where gives a very long time to show (wrong) results

### DIFF
--- a/htdocs/compta/stats/cabyprodserv.php
+++ b/htdocs/compta/stats/cabyprodserv.php
@@ -263,11 +263,11 @@ if ($modecompta == 'CREANCES-DETTES')
 			}
 		}
 
-		$sql .= " AND (p.rowid IN ";
-		$sql .= " (SELECT fk_product FROM ".MAIN_DB_PREFIX."categorie_product cp WHERE ";
+		$sql .= " AND ";
+
 		if ($subcat) $sql .= "cp.fk_categorie IN (".$listofcatsql.")";
 		else $sql .= "cp.fk_categorie = ".$selected_cat;
-		$sql .= "))";
+
 	}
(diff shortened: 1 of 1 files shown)


htdocs/compta/stats/cabyprodserv.php
```diff
@@ -232,7 +232,7 @@
 	if ($selected_cat === -2) {	// Without any category
 		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."categorie_product as cp ON p.rowid = cp.fk_product";
 	} elseif ($selected_cat) { 	// Into a specific category
-		$sql .= ", ".MAIN_DB_PREFIX."categorie as c, ".MAIN_DB_PREFIX."categorie_product as cp";
+		$sql .= " INNER JOIN ".MAIN_DB_PREFIX."categorie_product as cp ON (cp.fk_product = p.rowid) INNER JOIN ".MAIN_DB_PREFIX."categorie as c ON (cp.fk_categorie = c.rowid)";
```
eldy, line 235: I think we should completely removed this because this join can generate duplicates record that will be counted twice into the SUM and later just keep the test line 266
```
$sql .= " AND (p.rowid IN ";
		$sql .= " (SELECT fk_product FROM ".MAIN_DB_PREFIX."categorie_product cp WHERE ";
		if ($subcat) $sql .= "cp.fk_categorie IN (".$listofcatsql.")";
		else $sql .= "cp.fk_categorie = ".$selected_cat;
		$sql .= "))";
```
So even if product is in several category when we search with $subcal set, we found and return product/line only once.

atm-quentin: good idea i do that

atm-quentin:
done

PR #17801, opened 2021-06-02
NEW : Auto create invoice for payment done on an order with online payment link of order

### LINKED ISSUE #17762: Missing all process after payment with online link on order
# Bug

The payment link on the order is natively available in dolibarr
Payment is possible on this link
Once the payment has been made, no action is taken.
It missing at least :

- Creating an invoice
- the allocation of the payment to this invoice.

## Environment
- **Version**: All
- **OS**: All
- **Web server**: All
- **PHP**: All
- **Database**: All
- **URL(s)**: All

Show online payment if the invoice module is active.
After payment of an order

- Creation of the invoice
- Assignment of payment on the billing (link to the order)

nicolas-eoxia:
# Fix #17762


FHenry:
@nicolas-eoxia Could you add a test on order/card.php, display payment link if invoice module is enabled (there is already the test for stripe/paypal/payBox) ?

lmag:
Top thank guys we will do it 
@nicolas-eoxia

htdocs/commande/card.php
```diff
@@ -2633,7 +2633,7 @@
 			$somethingshown = $form->showLinkedObjectBlock($object, $linktoelem, $compatibleImportElementsList);
 
 			// Show online payment link
-			$useonlinepayment = (!empty($conf->paypal->enabled) || !empty($conf->stripe->enabled) || !empty($conf->paybox->enabled));
+			$useonlinepayment = ((!empty($conf->paypal->enabled) || !empty($conf->stripe->enabled) || !empty($conf->paybox->enabled)) && !empty($conf->facture->enabled));
```
eldy, line 2636: If we don't use the invoice module, we still should be able to record payment.
Test 
if !empty($conf->facture->enabled)   must be done inside the  paymentok.php file and is the condition to know if we must auto-create the invoice and payment or not.

FHenry:
@nicolas-eoxia @lmag could you fix conflict and check the review ?

nicolas-eoxia:
@eldy @FHenry It's fine for me, it seems fine to you.

PR #17817, opened 2021-06-03
FIX : 13.0 massaction bill order

ghost:
# FIX
(This fix is acting when we want to make a single bill for a third).

**We can now create one bill for several orders of the same third**

- Traductions added
- Add of an input which expect the ref_suppliar (required) to create a bill from at least two orders.
- After the creation of the bill, list is sorted as at the begging, by billed: no

_HAPPY READING_

### DIFF
--- a/htdocs/fourn/class/fournisseur.facture.class.php
+++ b/htdocs/fourn/class/fournisseur.facture.class.php
@@ -398,7 +398,7 @@ class FactureFournisseur extends CommonInvoice
 		$sql .= ")";
 		$sql .= " VALUES (";
 		$sql .= "'(PROV)'";
-		$sql .= ", ".($this->ref_supplier ? "'".$this->db->escape($this->ref_supplier)."'" : "null");
+		$sql .= ", '".$this->db->escape($this->ref_supplier)."'";
 		$sql .= ", ".$conf->entity;
 		$sql .= ", '".$this->db->escape($this->type)."'";
 		$sql .= ", '".$this->db->escape(isset($this->label) ? $this->label : (isset($this->libelle) ? $this->libelle : ''))."'";
--- a/htdocs/fourn/class/fournisseur.facture.class.php
+++ b/htdocs/fourn/class/fournisseur.facture.class.php
@@ -398,7 +398,7 @@ class FactureFournisseur extends CommonInvoice
 		$sql .= ")";
 		$sql .= " VALUES (";
 		$sql .= "'(PROV)'";
-		$sql .= ", '".$this->db->escape($this->ref_supplier)."'";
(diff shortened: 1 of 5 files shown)


htdocs/install/mysql/migration/12.0.0-13.0.0.sql
```diff
@@ -585,3 +585,5 @@ insert into llx_c_actioncomm (id, code, type, libelle, module, active, position)
 
 ALTER TABLE llx_export_model MODIFY COLUMN type varchar(64);
 
+
+ALTER TABLE llx_facture_fourn CHANGE ref_supplier ref_supplier VARCHAR(180) NULL;
```
eldy, line 589: To increase size of a column, use instead 
ALTER TABLE llx_table MODIFY COLUMN name varchar(180);

htdocs/fourn/commande/list.php
```diff
@@ -737,8 +743,27 @@
...
+		print '</td>';
+		print '<td>';
+		print '<input name="refsupplier" type="text" id="refsupplier">';
+		print '</td>';
+		print '</tr>';
 		print '</table>';
-
+		?>
+		<script>
+			window.addEventListener('load', function () {
```
eldy, line 759: Js code to check if page is loaded must be done with jquery:
$( document ).ready()

htdocs/fourn/class/fournisseur.facture.class.php
```diff
@@ -398,7 +398,7 @@ public function create($user)
 		$sql .= ")";
 		$sql .= " VALUES (";
 		$sql .= "'(PROV)'";
-		$sql .= ", '".$this->db->escape($this->ref_supplier)."'";
+		$sql .= ", ".($this->ref_supplier ? "'".$this->db->escape($this->ref_supplier)."'" : "null");
```
eldy, line 401: Should not be required, we want ref_supplier being not null.

htdocs/fourn/commande/list.php
```diff
@@ -737,8 +743,29 @@
 		print $form->selectyesno('validate_invoices', 1, 1);
 		print '</td>';
 		print '</tr>';
+		print '<tr class="fieldrequired" style="display:none">';
+		print '<td>';
+		print $langs->trans('RefSupplier');
+		print '</td>';
+		print '<td>';
+		print '<input name="refsupplier" type="text" id="refsupplier">';
```
eldy, line 751: when using a mass action, if you select a lot of supplier orders from different vendors, when we will create invoices, we will have 1 supplier_ref for each different vendor. So how can we enter here the value into a single field ?
Asking the field may be interesting if we select "one bill by third" (you manage this, good), but also if the supplier is the same for all selected purchase order.

PR #17840, opened 2021-06-07
Fix multiple order linked to invoice

atm-quentin:
# Fix multiple order linked to invoice wasn't display on pdf_crabe

If multiple order is link to an invoice, it's concatenated with public note.
I did the same thing with shipment.

If only one link, display like it was


eldy:
PR into develop (it change the behaviour) is better.

PR #17841, opened 2021-06-07
Compatibility phpv8

TheGivrix:


htdocs/admin/dict.php
```diff
@@ -2221,7 +2221,7 @@ function fieldList($fieldlist, $obj = '', $tabname = '', $context = '')
 
 	foreach ($fieldlist as $field => $value) {
 		if ($value == 'entity') {
-			$withentity = $obj->{$value};
+			$withentity = empty($obj->{$value}) ? '' : $obj->{$value};
```
eldy, line 2224: If $obj->{$value} is defined and is 0, the change will have for result
$withentity = '';
So it is a "not compatible" change.
May be a !isset is better.

htdocs/core/lib/admin.lib.php
```diff
@@ -1466,10 +1466,7 @@ function complete_elementList_with_modules(&$elementList)
 
 						// We discard modules according to features level (PS: if module is activated we always show it)
 						$const_name = 'MAIN_MODULE_'.strtoupper(preg_replace('/^mod/i', '', get_class($objMod)));
-						if ($objMod->version == 'development' && $conf->global->MAIN_FEATURES_LEVEL < 2 && !$conf->global->$const_name) {
-							$modulequalified = 0;
-						}
-						if ($objMod->version == 'experimental' && $conf->global->MAIN_FEATURES_LEVEL < 1 && !$conf->global->$const_name) {
+						if ($objMod->version == 'development' && getDolGlobalInt('MAIN_FEATURES_LEVEL') < 2 && empty($conf->global->$const_name) || $objMod->version == 'experimental' && getDolGlobalInt('MAIN_FEATURES_LEVEL') < 1 && empty($conf->global->$const_name)) {
```
eldy, line 1469: Mixing || and && make the condition being difficult to understand and probably wrong.
Old code was easier to understand.
We test first complex condition to set modulequalified = 0, then we test another one to make same result. 
We are sure here we do a or. With the new code the or is wrong because it is evaluated because the last && making all test bugged.

eldy:
Even if this PR goal is to fix a bug, it seems it introduces more problems than it solves, so request is closed.

PR #17845, opened 2021-06-08
FIX : need to add payment sum to getlibstatus function in object linked block (bug still exists in 13.0)

atm-gauthier:


htdocs/compta/facture/tpl/linkedobjectblock.tpl.php
```diff
@@ -85,7 +85,7 @@
     	}
 	}
     print '</td>';
-    print '<td class="linkedcol-statut right">'.$objectlink->getLibStatut(3).'</td>';
+    print '<td class="linkedcol-statut right">'.$objectlink->getLibStatut(3, $objectlink->getSommePaiement()).'</td>';
```
eldy, line 88: I think $objectlink can be an object of any type and ->getSommePaiement does not exists for mots of them

atm-gauthier: @eldy you're right, i added a method_exists() and did the same thing for supplier invoices

PR #17846, opened 2021-06-08
avoid fatal error

frederic34:


htdocs/compta/facture/card.php
```diff
@@ -3208,7 +3208,7 @@
 				$facids = $facturestatic->list_qualified_avoir_invoices($soc->id);
 				if ($facids < 0)
 				{
-					dol_print_error($db, $facturestatic);
+					dol_print_error($db, $facturestatic->error, $facturestatic->errors);
 					exit;
```
frederic34, line 3212: html_entity_decode does not support object

PR #17847, opened 2021-06-08
db->sanitize is removing spaces

frederic34:
it breaks the sql by removing spaces

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -4164,7 +4164,7 @@ class Facture extends CommonInvoice
 			$sqlSit .= " AND fs.fk_statut in (".self::STATUS_VALIDATED.",".self::STATUS_CLOSED.")";
 			$sqlSit .= " GROUP BY fs.situation_cycle_ref";
 			$sqlSit .= " ORDER BY fs.situation_counter";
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/class/facture.class.php
```diff
@@ -4164,7 +4164,7 @@ public function list_qualified_avoir_invoices($socid = 0)
 			$sqlSit .= " AND fs.fk_statut in (".self::STATUS_VALIDATED.",".self::STATUS_CLOSED.")";
 			$sqlSit .= " GROUP BY fs.situation_cycle_ref";
 			$sqlSit .= " ORDER BY fs.situation_counter";
-			$sql .= " AND ( f.type != ".self::TYPE_SITUATION." OR f.rowid IN (".$this->db->sanitize($sqlSit).") )"; // Type non 5 si facture non avoir
+			$sql .= " AND ( f.type != ".self::TYPE_SITUATION." OR f.rowid IN (".$this->db->escape($sqlSit).") )"; // Type non 5 si facture non avoir
```
eldy, line 4167: Inside a IN (...), we must use ->sanitize
->escape if for  = '...'

frederic34: it happens with INVOICE_USE_SITUATION_CREDIT_NOTE

eldy: I see. The sanitize is dedicated to put a list of id into a IN.
Here the IN contains a subrequest, so we don't need sanitize, neither we need escape.
So i pushed a better fix. Thanks to test it.

PR #17850, opened 2021-06-08
Adding hooks on product stat in order to allow to expand stat on othe…

moreauftheobald:
Adding hook to htdocs/core/lib/product.lib.php /  show_stats_for_company In order to allow to extends stat to other module

### DIFF
--- a/htdocs/core/lib/product.lib.php
+++ b/htdocs/core/lib/product.lib.php
@@ -361,8 +361,9 @@ function product_lot_admin_prepare_head()
 function show_stats_for_company($product, $socid)
 {
 	global $conf, $langs, $user, $db, $hookmanager;
+
 	$form = new Form($db);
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/product.lib.php
```diff
@@ -360,9 +360,9 @@ function product_lot_admin_prepare_head()
  */
 function show_stats_for_company($product, $socid)
 {
-	global $conf, $langs, $user, $db;
+	global $conf, $langs, $user, $db, $hookmanager;
 	$form = new Form($db);
-
+	$hookmanager->initHooks(array('productstats'));
```
eldy, line 365: the inithook must be done by the parent page (the main php file that include all other). So a init should have already be done when we enter the show_stats_for_company(), so i removed this line.

PR #17859, opened 2021-06-09
FIX Missing link to language file

aspangaro:
image


daraelmin:
Isn't it a bad name ? I mean, it should be "DateOfBirth" instead "DateToBirth"

PR #17862, opened 2021-06-09
FIX : auto rounding on "_FORMATED_" tags

atm-gauthier:


htdocs/core/lib/functions.lib.php
```diff
@@ -6453,11 +6453,11 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
...
-		$substitutionarray['__AMOUNT_EXCL_TAX_FORMATED__'] = is_object($object) ? ($object->total_ht ? price($object->total_ht, 0, $outputlangs, 0, 0, -1, $conf->currency) : null) : '';
-		$substitutionarray['__AMOUNT_VAT_FORMATED__']      = is_object($object) ? ($object->total_vat ? price($object->total_vat, 0, $outputlangs, 0, 0, -1, $conf->currency) : ($object->total_tva ? price($object->total_tva, 0, $outputlangs, 0, 0, -1, $conf->currency) : null)) : '';
-		if ($onlykey != 2 || $mysoc->useLocalTax(1)) $substitutionarray['__AMOUNT_TAX2_FORMATED__']     = is_object($object) ? ($object->total_localtax1 ? price($object->total_localtax1, 0, $outputlangs, 0, 0, -1, $conf->currency) : null) : '';
-		if ($onlykey != 2 || $mysoc->useLocalTax(2)) $substitutionarray['__AMOUNT_TAX3_FORMATED__']     = is_object($object) ? ($object->total_localtax2 ? price($object->total_localtax2, 0, $outputlangs, 0, 0, -1, $conf->currency) : null) : '';
+		$substitutionarray['__AMOUNT_FORMATED__']          = is_object($object) ? ($object->total_ttc ? price($object->total_ttc, 0, $outputlangs, 0, -1, -1, $conf->currency) : null) : '';
```
eldy, line 6456: The value inside $object->total_ttc should already have the correct accuracy, so it should not be required to round (rounding avoid to see the real value). 
Do you have an example of use case that need this. Which value did you get/see on PDF and what was expected ?

atm-gauthier: Hi @eldy, if i have an invoice with only one decimal > 0 :
image

I get this kind of result :
image

for me i should get price like x xxx,xx no matter how many decimal > 0 my invoice price is

PR #17865, opened 2021-06-09
New develop autocomplete accounting account groups

atm-florianm:
# NEW: autocomplete accounts classes based on existing values

My [previous PR](https://github.com/Dolibarr/dolibarr/pull/17669) was closed even though I had updated it to take into account input from the reviewers.

I believe it was closed by mistake, based on the `PR not qualified` tag. I think the tag no longer applies since the PR was updated according to suggestions.


aspangaro:
Ping @eldy

PR #17867, opened 2021-06-09
FIX clone invoice with acount line

lvessiller-opendsi:
FIX clone invoice with acount line
- when you have an invoice with an acount line and you want to clone it : you got a blank page

1) Create an invoice with an acount line like this ;
image

2) Click to clone
image

3) Then click yes and you got a blank page

- with this "apache" error : "Got error 'PHP message: PHP Fatal error:  Uncaught Error: __clone method called on non-object in ..."

### DIFF
--- a/htdocs/core/triggers/interface_90_modSociete_ContactRoles.class.php
+++ b/htdocs/core/triggers/interface_90_modSociete_ContactRoles.class.php
@@ -84,11 +84,7 @@ class InterfaceContactRoles extends DolibarrTriggers
 				if (is_array($TContact) && ! empty($TContact)) {
 					$TContactAlreadyLinked = array();
 					if ($object->id > 0) {
-						$cloneFrom = dol_clone($object, 1);
-
-						if (! empty($cloneFrom->id)) {
-							$TContactAlreadyLinked = array_merge($cloneFrom->liste_contact(- 1, 'external'), $cloneFrom->liste_contact(- 1, 'internal'));
-						}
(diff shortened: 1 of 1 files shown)


htdocs/core/triggers/interface_90_modSociete_ContactRoles.class.php
```diff
@@ -84,11 +84,7 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
 				if (is_array($TContact) && ! empty($TContact)) {
 					$TContactAlreadyLinked = array();
 					if ($object->id > 0) {
-						$cloneFrom = dol_clone($object, 1);
-
-						if (! empty($cloneFrom->id)) {
-							$TContactAlreadyLinked = array_merge($cloneFrom->liste_contact(- 1, 'external'), $cloneFrom->liste_contact(- 1, 'internal'));
-						}
+						$TContactAlreadyLinked = array_merge($object->liste_contact(- 1, 'external'), $object->liste_contact(- 1, 'internal'));
```
eldy, line 87: Can't see why this fix is fixing something. Can you point me ?

lvessiller-opendsi: To clone this "object"  (dol_clone()) generates an error when you have an "deposit" line in the origin invoice.
So I replaced '$cloneFrom" with the "$object" to avoid this problem.
It seems not to be needed to clone "$object" to merge contact lists.

eldy: You mean dol_clone() return an error according to content of $object ?
This looks very strange. Which error do you have ?

PR #17873, opened 2021-06-10
NEW when multiple order linked to facture, show list into note.

atm-quentin:
# Fix multiple order linked to facture
https://github.com/Dolibarr/dolibarr/pull/17840


atm-quentin:
@eldy

atm-maxime:
> @eldy

PR postponed label means the PR will be merged in develop once the new branch (14) has been created.

htdocs/core/lib/pdf.lib.php
```diff
@@ -2262,7 +2262,7 @@ function pdf_getTotalQty($object, $type, $outputlangs)
  * 	@param	Translate	$outputlangs	Object lang for output
  * 	@return	array                       Linked objects
  */
-function pdf_getLinkedObjects($object, $outputlangs)
+function pdf_getLinkedObjects(&$object, $outputlangs)
```
eldy, line 2265: Why do you need the  &  ? No reason to modify object in a read only method.

atm-quentin: because i use $object->note_public to print order ref if there're too much linked orders

atm-quentin: @eldy

htdocs/core/lib/pdf.lib.php
```diff
@@ -2284,7 +2284,17 @@ function pdf_getLinkedObjects($object, $outputlangs)
 			}
 		} elseif ($objecttype == 'commande' || $objecttype == 'supplier_order') {
 			$outputlangs->load('orders');
-			foreach ($objects as $elementobject) {
+
+			if (count($objects) > 1) {
+				$object->note_public .= '<br/>'.$outputlangs->transnoentities("RefOrder").' : <br/>';
```
eldy, line 2289: You must use <br> that is html5 tag and not <br/> that is xml tag (even if <br/> is accepted by browsers).
Also, you must use dol_concatdesc to add text to a note_public (because not can be a full text or html text and concatenation need different action when you add a html text on a non html text. dol_concatdesc do it for you.

atm-quentin: done

atm-quentin:
@eldy

atm-quentin:
@eldy is it ok for you ?

atm-quentin:
@eldy

PR #17877, opened 2021-06-10
new: add qc_frequency on product lot finish #17322

FHenry:
new: add qc_frequency on product lot finish #17322


htdocs/install/mysql/migration/13.0.0-14.0.0.sql
```diff
@@ -79,6 +79,7 @@ DELETE FROM llx_user_param where param = 'MAIN_THEME' and value in ('auguria', '
 ALTER TABLE llx_product_lot ADD COLUMN eol_date datetime NULL;
 ALTER TABLE llx_product_lot ADD COLUMN manufacturing_date datetime NULL;
 ALTER TABLE llx_product_lot ADD COLUMN scrapping_date datetime NULL;
+ALTER TABLE llx_product_lot ADD COLUMN qc_frequency integer DEFAULT NULL;
```
eldy, line 82: This field is inside llx_product

FHenry: Yes and as you said on previous pr, it can be defaulted on product and different on lot

FHenry:
@eldy I've updated the PR to change migration file to impect, now in v15 migration SQL file

eldy:
Before going further away on this topic, i would like to have a summary of what is already working around this feature in current develop/v15.
Can you make a summary of what we can already do around this feature without the patch ? Is there something I can test ? Where ? I need to understand current existing feature before validating more. But where are current feature around this for the moment ?

eldy:
This PR introduce a qc_frequency in lot but before we should have it on product.
I can't find the feature  currently in v15 and develop v16 on how to manage the qc_frequency for products. Did i miss something ?

PR #17883, opened 2021-06-10
NEW: holiday: handle monthly balance updates with cronjob

marc-dll:


htdocs/holiday/define_holiday.php
```diff
@@ -193,10 +193,6 @@
 
 
 $typeleaves = $holiday->getTypes(1, 1);
-$result = $holiday->updateBalance(); // Create users into table holiday if they don't exists. TODO Remove this whif we use field into table user.
```
eldy, line 196: A lot of users are hosting Dolibarr on server where cron tasks are not available. So we can add a cron but we must keep also update from web logging and having it disabled if cron module is on + cron job has status "scheduled".

marc-dll: @eldy Done

PR #17887, opened 2021-06-10
NEW: leave requests: add field into type dictionary to block request if balance is negative

marc-dll:


htdocs/holiday/class/holiday.class.php
```diff
@@ -259,6 +259,17 @@ public function create($user, $notrigger = 0)
 			$this->error = "ErrorBadParameterFkType"; return -1;
 		}
 
+		$checkBalance = getDictvalue(MAIN_DB_PREFIX.'c_holiday_types', 'block_if_negative', $this->fk_type);
+
+		if ($checkBalance > 0) {
```
eldy, line 264: We must check when we validate only. 
But when we create, we must check only if status is not draft.

marc-dll:
@eldy PR updated (finally!) following your remarks, and with conflicts solved

marc-dll:
@eldy Conflicts solved again ;)

marc-dll:
@eldy Conflicts solved another time. Can you please make a feedback if you consider it should not be merged, please ?

PR #17899, opened 2021-06-11
Update stockcorrection.tpl

kkerleo:
Batch id was missing


eldy:
Can you describe how we can reproduce the trouble ?

eldy:

PR #17953, opened 2021-06-16
FIX : Tracking number label shouldn't be displaied on "espadon" model…

atm-gauthier:
… if shipping tracking number is empty

### DIFF
--- a/htdocs/core/modules/expedition/doc/pdf_espadon.modules.php
+++ b/htdocs/core/modules/expedition/doc/pdf_espadon.modules.php
@@ -364,7 +364,11 @@ class pdf_espadon extends ModelePdfExpedition
 					$tab_top_alt = $tab_top;
 
 					$pdf->SetFont('', 'B', $default_font_size - 2);
-					$pdf->writeHTMLCell(60, 4, $this->posxdesc - 1, $tab_top - 1, $outputlangs->transnoentities("TrackingNumber")." : ".$object->tracking_number, 0, 1, false, true, 'L');
+					if(!empty($object->tracking_number)) {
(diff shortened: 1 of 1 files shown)


htdocs/core/modules/expedition/doc/pdf_espadon.modules.php
```diff
@@ -364,7 +364,11 @@ public function write_file($object, $outputlangs, $srctemplatepath = '', $hidede
 					$tab_top_alt = $tab_top;
 
 					$pdf->SetFont('', 'B', $default_font_size - 2);
-					$pdf->writeHTMLCell(60, 4, $this->posxdesc - 1, $tab_top - 1, $outputlangs->transnoentities("TrackingNumber")." : ".$object->tracking_number, 0, 1, false, true, 'L');
+					if(!empty($object->tracking_number)) {
+						$pdf->writeHTMLCell(60, 4, $this->posxdesc - 1, $tab_top - 1, $outputlangs->transnoentities("TrackingNumber")." : ".$object->tracking_number, 0, 1, false, true, 'L');
+					} else {
+						$pdf->writeHTMLCell(60, 4, $this->posxdesc - 1, $tab_top - 5, '', 0, 1, false, true, 'L');
```
eldy, line 370: Why keeping this line (it prints empty string) ?

atm-gauthier: Hi @eldy, i got a strange pdf display when removed the line, but you're right, keeping this line was not the right way to keep a correct display, then i added "if(!empty($object->tracking_number))" before "$tab_top_alt = $pdf->GetY();" now everything seems ok

PR #17968, opened 2021-06-17
Make some product fields hidable

supergeorg:
# New: Add some more parameters to hide fields in the product card

This commit allows to hide some more fields in the product card (e.g. you can hide SIZE but not WEIGHT).
To allow the product card to be customized a bit better, following new variables can be set to hide the corresponding fields:

- PRODUCT_DISABLE_PUBLIC_URL
- PRODUCT_DISABLE_STOCK_LEVELS
- PRODUCT_DISABLE_NATURE
- PRODUCT_DISABLE_WEIGHT
- PRODUCT_DISABLE_PRICES
- PRODUCT_DISABLE_ACCOUNTING

The documentation has to be adapted accordingly.

### DIFF
--- a/htdocs/product/card.php
+++ b/htdocs/product/card.php
@@ -1236,13 +1236,13 @@ if (is_object($objcanvas) && $objcanvas->displayCanvasExists($action)) {
 		}
 
 		if ($type != 1) {	// Nature, Weight and volume only applies to products and not to services
-            if (empty($conf->global->PRODUCT_DISABLE_NATURE)) {
-                // Nature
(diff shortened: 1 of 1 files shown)


eldy:
Thanks for contribution.
However, using 1 constant for each field to show/hide is not a good choice for future. We should use instead a generic feature into menu 
Value/Features/Default values.
We can add a tab  "Hide fields", so we can choose any field and add a js script code to hide all field named into this setup (like we do for the tab "Fields with the focus"). So this feature will work for any form, any field, any page.

aljawaid:
> 
> The documentation has to be adapted accordingly.

Done 
Setup Other https://wiki.dolibarr.org/index.php?title=Setup_Other

PR #17265, opened 2021-04-14
New possibility to define batch in draft shipment

atm-quentin:
# New possibility to define batch in draft shipment to use it as a preparation voucher

Give possibility to define later (on draft shipment) batch in a new tab.

1st : new line "Every warehouse" when converting order to shipment to define which product will be defined later.
2nd : Validation is blocked while every batch isnt define
3rd : you can define it in the new tab

image

PR #17978, opened 2021-06-18
NEW : modify a shipping draft

altairis-tof:
Add a button "Modify" on shipping draft and manage it.

### DIFF
--- a/htdocs/expedition/card.php
+++ b/htdocs/expedition/card.php
@@ -1060,7 +1060,7 @@ if ($action == 'create' || $action == 'edit') {
 			print '<table class="noborder centpercent">';
 
 			// Load shipments already done for same order
-			$object->loadExpeditions($action == 'edit'?'1':null); // dont load draft shippings if edit so we can show all lines
+			$object->loadExpeditions($action == 'edit' ? 1 : -1); // dont load draft shippings if edit so we can show all lines
 
 			if ($numAsked) {
 				print '<tr class="liste_titre">';
--- a/htdocs/expedition/card.php
+++ b/htdocs/expedition/card.php
@@ -1060,7 +1060,7 @@ if ($action == 'create' || $action == 'edit') {
 			print '<table class="noborder centpercent">';
 
 			// Load shipments already done for same order
-			$object->loadExpeditions($action == 'edit'); // dont load draft shippings if edit so we can show all lines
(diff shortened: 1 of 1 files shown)


daraelmin (commented):
Isn't $db->begin() needed to create the $objectsrc object on line 228?

altairis-tof:
@eldy did you view this PR ?

altairis-tof:
some sreen samples : 
- add an Update button on draft shipment : 
Expédition_-_2021-08-03_13 50 06
- then the draft editable shipment : 
Expédition_-_2021-08-03_13 49 35

BebZ:
Wow, so there are 2 PR in "conflict", maybe that's why Eldy did not merge both of them yet ? 
PR of Quentin is making a **draft** shipment possible without batch +1
and PR of Tof is to be "fully" edit-able (but the PR of Quentin makes it also possible) +0 +0
PR of Tof is not a new "tab" it seems completely integrated in the existing screens +1

... I did not try/test both PRs...

altairis-tof:
hello @eldy it would be nice to make a decision about this PR with the label "pending analysis" since september 2021 ;-)
i think it is useful to allow to edit a shipping.

eldy:
I did not received a lot of feedback, except @Bebz with point of views about the different PRs for draft shipment edition.
PR of Quentin #17265 will probably be merged before because having another tab will open the road to other feature, like allowing edition of barcode at any step.

altairis-tof:
i have made lot of improvements that are not yet included in this PR... i still do not understand why you dont agree with an edit form for shipments...

eldy:
> i have made lot of improvements that are not yet included in this PR... i still do not understand why you dont agree with an edit form for shipments...

Never say i disagree. The point is how, using which solution of which PR. It appears that for future we need a dedicated page to edit draft shipment lot/serial/warehouses. Must be a dedicated new TAB on the shipment. Main tab will remain the summary of the shipment (current view). The new Tab is the editor of shipments lot/ serial /warehouse.
This is what is suggested by https://github.com/Dolibarr/dolibarr/pull/17265 but PR is nor mergeable for the moment.
If you have another PR that add such a tab to edit lot/serial/warehouse of a draft shipment, can you paste the number here as a comment ?
Note: A good page to make such a tab we can use as example is the page /fourn/commande/dispatch.php?id=xxx

eldy:
The feature to allow edit of shipment is available in v18, so i close this PR.

Issue #17058, opened 2021-03-31
REST API supplierorders error on creation with lines array


guiguichupote:
# Bug
When creating a supplier order, the json lines sent to api are parsed as an array of arrays (json_decode is called with param associative to true, so objects are parsed as associative arrays). 
But in the create method in fournisseur.commande.class.php file, it expects array of lines to be an array of objects.
Is it a bug or a configuration to be setted?

eldy:
You should paste the paylog you used as json string pushed into your API...

guiguichupote:
In fact there is no method to add line in supplier order.
I think we can add it 


```
  /**
       * Add a line to given supplier order
       *
       * @param int   $id             Id of supplier order to update
       * @param array $request_data   supplier order line data
       *
       * @url	POST {id}/lines
       *
       * @return int|bool
       */
      public function postLine($id, $request_data = null)
      {
          if (!DolibarrApiAccess::$user->rights->fournisseur->commande->creer) {
              throw new RestException(401);
          }

          $result = $this->order->fetch($id);
          if (!$result) {
			throw new RestException(404, 'Supplier order not found');
		}

		if (!DolibarrApi::_checkAccessToResource('fournisseur', $this->order->id, 'commande_fournisseur', 'commande')) {
			throw new RestException(401, 'Access not allowed for login '.DolibarrApiAccess::$user->login);
		}

        $updateRes = $this->order->addline(
            $request_data['desc'],
            $request_data['subprice'],
            $request_data['qty'],
            $request_data['tva_tx'],
            $request_data['localtax1_tx'],
            $request_data['localtax2_tx'],
            $request_data['fk_product'],
            0,
            $request_data['ref_fourn'], // $this->lines[$i]->ref_fourn comes from field ref into table of lines. Value may ba a ref that does not exists anymore, so we first try with value of product
            $request_data['remise_percent'],
            'HT',
            0,
            $request_data['product_type'],
            $request_data['info_bits'],
            false,
            $request_data['date_start'],
            $request_data['date_end'],
            $request_data['array_options'],
            $request_data['fk_unit']
        );

        if ($updateRes < 0) {
            throw new RestException(400, 'Unable to insert the new line. Check your inputs. '.$this->order->error);
        }

        return $updateRes;
    }

    /**
     * Deletes a line of a given supplier order
     *
     * @param int   $id             Id of supplier order
     * @param int   $lineid 		Id of the line to delete
     *
[...]

Issue #14764, opened 2020-09-17
REST orders/supplierorders: Support adding products without specifying prices


dpriskorn:
# Feature Request
currently Dolibarr defaults to empty prices if no price is specified during adding of supplier_order_line

## Use case
I'm importing an order from a supplier and I got the quantity and product id, that should be enough to add a meaningful line to the order. Dolibarr can fetch the supplier price form the database and error if none found.

## Suggested implementation
default to fetching price from database instead of NULL

## Suggested steps
edit the api-file and implement getting prices there
or add it in the dolibarr function called from the api-file

PR #17983, opened 2021-06-19
Add what REST API cannot do to README

dpriskorn:
Update readme with details about what the REST API cannot do


README.md
```diff
@@ -219,6 +220,17 @@ These are features that Dolibarr does **not** yet fully support:
 - Payroll module
 - No native embedded Webmail, but you can send email to contacts in Dolibarr with e.g. offers, invoices, etc.
 - Dolibarr can't do coffee (yet)
+- The REST API currently cannot:
+  - Update a supplier order 
+  - Get tickets
```
eldy, line 225: Seems available in develop branch

README.md
```diff
@@ -219,6 +220,17 @@ These are features that Dolibarr does **not** yet fully support:
...
+  - Add new ledger entry
+  - Output extrafields when getting orders
+  - Get virtual products (aka lots)
+  - Add special taxcodes to a line on a supplier invoice
+  - Get or post shipments
+  - Do updating or inserting of extrafields on any objects
```
eldy, line 231: Seems available in develop branch

README.md
```diff
@@ -219,6 +220,17 @@ These are features that Dolibarr does **not** yet fully support:
 - Payroll module
 - No native embedded Webmail, but you can send email to contacts in Dolibarr with e.g. offers, invoices, etc.
 - Dolibarr can't do coffee (yet)
+- The REST API currently cannot:
+  - Update a supplier order 
+  - Get tickets
+  - Add new ledger entry
+  - Output extrafields when getting orders
```
eldy, line 227: Seems available in develop branch

README.md
```diff
@@ -219,6 +220,17 @@ These are features that Dolibarr does **not** yet fully support:
...
+  - Get virtual products (aka lots)
+  - Add special taxcodes to a line on a supplier invoice
+  - Get or post shipments
+  - Do updating or inserting of extrafields on any objects
+  - Add an image to a product
```
eldy, line 232: Seems available in develop branch

README.md
```diff
@@ -219,6 +220,17 @@ These are features that Dolibarr does **not** yet fully support:
...
+  - Add multiprices to products
```
eldy, line 233: Seems available in develop branch

dpriskorn:
Here is another one :)
#18071
Maybe this can be done by updating the order?
https://github.com/Dolibarr/dolibarr/blob/1a63a379cc447c3ddc07dd1c9c715db9efe11141/htdocs/fourn/class/api_supplier_orders.class.php#L255
I would prefer a separate method and endpoint like "validate" though.

dpriskorn:
Here is another one:
 #17058
ksar posted a solution in a comment

dpriskorn:
Here is another one:
If prices are already present it should be optional to specify them IMO:
 #14764
this is labeled as a feature request, but IMO it is probably a bug because the user (me) expect Dolibarr to use the data it already has instead of defaulting to NULL.

README.md
dpriskorn: Where? I found nothing in https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/product/class/api_products.class.php

README.md
dpriskorn: Where? I found nothing in https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/product/class/api_products.class.php

README.md
dpriskorn: Correct! Confirmed working on v13 also on get /orders/{id}

README.md
dpriskorn: This line is too vague so I'll delete it

README.md
dpriskorn: Correct! https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/ticket/class/api_tickets.class.php
I'll remove it :)

dpriskorn:
@eldy is it ready to merge now?

README.md
eldy: A product image is sending a document image. You must use api document

PR #17987, opened 2021-06-20
Fix 17986 error on comparing integer and string in strict sql

daraelmin:
@eldy 
I'm not sure that is the best solution, but it may be important to fix it before 14.0.0, so I try the faster way.

### DIFF
--- a/htdocs/adherents/class/adherentstats.class.php
+++ b/htdocs/adherents/class/adherentstats.class.php
@@ -92,7 +92,7 @@ class AdherentStats extends Stats
 		$sql = "SELECT date_format(p.dateadh,'%m') as dm, count(*)";
 		$sql .= " FROM ".$this->from;
 		//if (!$user->rights->societe->client->voir && !$user->socid) $sql.= ", ".MAIN_DB_PREFIX."societe_commerciaux as sc";
-		$sql .= " WHERE ".dolSqlDateFilter('p.dateadh', 0, 0, $year, 1);
+		$sql .= " WHERE ".dolSqlDateFilter('p.dateadh', 0, 0, (int) $year, 1);
 		$sql .= " AND ".$this->where;
 		$sql .= " GROUP BY dm";
 		$sql .= $this->db->order('dm', 'DESC');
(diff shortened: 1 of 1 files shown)


daamien:
This is not the proper solution.

It's best to use `YEAR()` or `EXTRACT( YEAR FROM ...)` instead of `DATE_FORMAT()`

```php
$sql .= " WHERE YEAR(p.dateadh) = ". (int) $year;
```

If you want to do a numeric comparison, just convert both values in `INT`. No need to use strings for that.

daraelmin:
May be this one is better.

eldy:
Usage of date function on SQL side (native but also provided by dolibarr like date_format) must be avoided because it will not manage correctly the timezone of user (that is defined on Dolibarr) and the timezone of server (that is stored in php setup). Also it is slower than using a true comparison on date (such comparison allow to use the index on date when using a test on YEAR will not use it)
Here the good practice is to use the function
$sql .= "WHERE ".dolSqlDateFilter('p.dateadh', 0, 0, $year, 1);

daraelmin:
> Usage of date function on SQL side (native but also provided by dolibarr like date_format) must be avoided because it will not manage correctly the timezone of user (that is defined on Dolibarr) and the timezone of server (that is stored in php setup). Also it is slower than using a true comparison on date (such comparison allow to use the index on date when using a test on YEAR will not use it)
> Here the good practice is to use the function
> $sql .= "WHERE ".dolSqlDateFilter('p.dateadh', 0, 0, $year, 1);

@eldy
Ok. I've done the changes. Thank you for your review.

However, I wonder if we shouldn't change this kind of statement to  :
```php
$sql = "SELECT date_format(p.dateadh,'%m') as dm, count(*)";
````

eldy:
> > Usage of date function on SQL side (native but also provided by dolibarr like date_format) must be avoided because it will not manage correctly the timezone of user (that is defined on Dolibarr) and the timezone of server (that is stored in php setup). Also it is slower than using a true comparison on date (such comparison allow to use the index on date when using a test on YEAR will not use it)
> > Here the good practice is to use the function
> > $sql .= "WHERE ".dolSqlDateFilter('p.dateadh', 0, 0, $year, 1);
> 
> @eldy
> Ok. I've done the changes. Thank you for your review.
> 
> However, I wonder if we shouldn't change this kind of statement to :
> 
> ```
> $sql = "SELECT date_format(p.dateadh,'%m') as dm, count(*)";
> ```

No, here the date_format is used to extract substring to get the Month (it is used as a substr method). We can keep it. The dolSqlDateFilter is designed to be used to forge the WHERE of sql.

daamien:
Thanks

PR #17992, opened 2021-06-20
fix warnings

frederic34:


htdocs/admin/user.php
```diff
@@ -268,7 +268,7 @@
 
 								// Defaut
 								print '<td class="center">';
-								if ($conf->global->USER_ADDON_PDF == $name) {
+								if (getDolGlobalString('USER_ADDON_PDF_ODT') == $name) {
```
frederic34, line 271: here it fix php warning and set default doc model

PR #17998, opened 2021-06-21
FIX: Reload values of the specified entity when change entity in the execution of cron jobs

kkhelifa-opendsi:
Reload values of the specified entity when change entity in the execution of cron jobs


htdocs/core/class/conf.class.php
```diff
@@ -161,6 +174,58 @@ public function __construct()
 	 */
 	public function setValues($db)
 	{
+		// Unset all old modules values
+		if (!empty($this->modules)) {
```
eldy, line 178: The whole $this->modules is completely erased few lines after with
$this->modules = array();
So i think we can remove this portion of code, doesn't it ?

htdocs/core/class/conf.class.php
```diff
@@ -161,6 +174,58 @@ public function __construct()
...
+			foreach ($this->modules as $m) {
+				if (isset($this->$m)) unset($this->$m);
+			}
+		}
+
+		// Properly declare multi-modules objects.
```
eldy, line 184: According to code comment and existing code, it seems that only 
		$this->file = new stdClass();
		$this->db = new stdClass();
are require to be set before the setValues is called.

So we should retrieve here all what is initialized into the constructor (except ->file and ->db).

What if you also add this ? Is this working on your side ?
		$this->mycompany = new stdClass();
		$this->admin = new stdClass();
		$this->user	= new stdClass();
		$this->browser = new stdClass();
		$this->medias = new stdClass();
		$this->cache = array();

philippe-opendsi:
ping @aspangaro

aspangaro:
Ping @eldy

PR #17999, opened 2021-06-21
FIX #17996: Avoid implicit type casts for llx_adherent_type.subscription

### LINKED ISSUE #17996: [Adherents]  llx_adherent_type.subscription is a VARCHAR
Another type mismatch in the "Adherents" module, similar to #17986 

in adherents/list.php , there are 3 `varchar = integer` comparisons like this

```php
$sql .= " AND (datefin IS NULL OR t.subscription = 0)";
```

See https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/adherents/list.php#L366

Again MySQL allows this kind of comparisons but PostgreSQL does not. By the way, the default value of the subscription column is "yes" which would lead to some unexpected behaviours because for MySQL the comparision `'yes' = 1` is false.

Anayway the numeric value should be quoted like this

```php
$sql .= " AND (datefin IS NULL OR t.subscription = '0')";
```

daamien:
Closes #17996

### DIFF
--- a/htdocs/adherents/list.php
+++ b/htdocs/adherents/list.php
@@ -363,13 +363,13 @@ if ($search_type > 0) {
 	$sql .= " AND t.rowid=".((int) $search_type);
 }
 if ($search_filter == 'withoutsubscription') {
-	$sql .= " AND (datefin IS NULL OR t.subscription = 0)";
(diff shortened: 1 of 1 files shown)


daamien:
Thanks !

PR #18005, opened 2021-06-22
FIX: holidays card: more standard error handling

noec764:


eldy:
What is the number of the bug that is fixed (how to reproduce) ?
Is it only a code quality PR ? If yes, you must push it into develop branch ?

noec764:
I don't know the number of the bug fixed.
In holidays card module some errors had non dolibarr standard handling with a blank page displayed when they occurred. 
I'm an intern from Scopen corporation.

eldy:
So it seems this is quality code enhancement and not a fix for a declared bug, so you must push your PR into the develop branch...

PR #18008, opened 2021-06-23
fix date in extrafields list search sql

p-payet:
# Fix # date in extrafields list search sql
Fix date in htdocs/core/tpl/extrafields_list_search_sql.tpl.php to match to the mysql date format


eldy:
Seems already fixed in current 13.0 branch

eldy:

PR #18019, opened 2021-06-24
NEW: make it easier to set the `keyword`, `keywords` and `description` attributes of an ecm file object

atm-florianm:
# New facilitate setting keyword(s) and description of ECM file
This PR extends the feature introduced with [that PR](https://github.com/Dolibarr/dolibarr/pull/15019/files) to enable modules to set the `keyword`, `description` and `keywords` attributes of `llx_ecm_file` entries more easily.


htdocs/ecm/class/ecmfiles.class.php
```diff
@@ -288,6 +288,7 @@ public function create(User $user, $notrigger = false)
 		$sql .= 'fullpath_orig,';
 		$sql .= 'description,';
 		$sql .= 'keywords,';
+		$sql .= 'keyword,';
```
eldy, line 291: This field does not exists. Field to store keywords of file is "keywords"

atm-florianm: It is true that the class EcmFiles doesn't reference the `keyword` field, but the field does exist in the database as per `install/mysql/migration/5.0.0-6.0.0.sql` (`ALTER TABLE llx_ecm_files ADD COLUMN keyword varchar(750);`)

The field `keyword` is also referenced in `commonobject.class.php` in `commonGenerateDocument()`.

Since it is not explicitly deprecated, I assumed it should be accounted for too.

eldy: The field does not exists on any new installation. So we must not have it into database request.
I have fixed the mention of ->keyword in code (it was also generated warning with php8)

atm-florianm: Thanks for the explanation, I have updated the PR accordingly.

atm-florianm:
Hi, just making sure you saw that the PR was fixed and the initial discussion resolved.

htdocs/core/lib/files.lib.php
```diff
@@ -1789,7 +1789,9 @@ function addFileIntoDatabaseIndex($dir, $file, $fullpathorig = '', $mode = 'uplo
 
 		if (is_object($object) && $object->id > 0) {
 			$ecmfile->src_object_id = $object->id;
-			$ecmfile->src_object_type = $object->table_element;
+			if (isset($object->table_element)) $ecmfile->src_object_type = $object->table_element;
```
eldy, line 1792: Setting src_object_id wihout setting src_object_type is useless.
For this reason table_element mus always be set. And if not set, we should report an error instead of allowing to record a data that we won't be able to reuse later.

atm-florianm: Thanks for your reply, I updated the PR.

PR #18031, opened 2021-06-25
Fix #18029 : modification of regular expression in function.lib

### LINKED ISSUE #18029: Getting email with "This is a message with multiple parts in MIME format."
# Bug
When i send a quote i receive this message "This is a message with multiple parts in MIME format." and no attachment in the email.

## Environment
- **Version**: 13.0.3

## [Attached files]
error

Hystepik:
Fix to make ' an impossible text in GETPOST('','restricthtml')


test/phpunit/SecurityTest.php
```diff
@@ -91,7 +91,7 @@ public function __construct()
      *
      * @return void
      */
-    public static function setUpBeforeClass()
+    public static function setUpBeforeClass():void
```
frederic34, line 94: travis check in develop branch are also done with php5.6

Hystepik: my bad I forgot to remove them

eldy:
Can you confirm that this PR fix the bug #18029 ? Can't understand why.
Also you comment let me thing that with the change you can't now include the char  '   inside the text ?
So a'b is no more a valid text with restricthtml ?

Hystepik:
First yes a'b isn't a valid text with restricthtml because in 14.0 it isn't too. Actually this bug is fix because if you cannot include the char ' in the text the message and you doesn't close it i don't know why but the MIME part was considered as a string so it would display in the mail.

eldy:
I see. 
It seems you fixes this by making a backport of a change in develop/v14. Right ? If yes, think to mention this into your PR desc or comment, because if it was already merged into another version, it means it is already "validated", so it may save time for the analysis...

Hystepik:
Yes it was a backport of a change in develop/14 sorry for the misunderstanding.
I next backport I will mention it in my PR.

eldy:
However i still don't understand why this fix fixes the #18029 . Because the GETPOST with restricthtml is called only when we submit data but not for generating the email content. Did you succeed in reproducing the bug ? How ? How did you find the trouble was in GETPOST restrcithtml ?

eldy:
PR seems to not be related to #18029

PR #18033, opened 2021-06-26
variable was created but not used

grandoc:


htdocs/adherents/list.php
```diff
@@ -591,7 +591,7 @@
 if ($user->rights->adherent->creer) {
 	$arrayofmassactions['close'] = img_picto('', 'close_title', 'class="pictofixedwidth"').$langs->trans("Resiliate");
 }
-if ($user->rights->adherent->supprimer) {
+if ($permissiontodelete) {
```
eldy, line 594: It seems this introduce a bug because the variable is declared into a if so is not always declared.

PR #18045, opened 2021-06-28
better count of order shipments

altairis-tof:
Count order shipments upon element_element instead of expeditiondet.

This prepares next PR that will improve shipping management.

This allow to count empty shipments... (not possible for the moment, but will be possible if we add option to (update and) delete shipping lines.

### DIFF
--- a/htdocs/commande/class/commande.class.php
+++ b/htdocs/commande/class/commande.class.php
@@ -2175,13 +2175,12 @@ class Commande extends CommonOrder
 	{
 		$nb = 0;
 
-		$sql = 'SELECT COUNT(DISTINCT ed.fk_expedition) as nb';
-		$sql .= ' FROM '.MAIN_DB_PREFIX.'expeditiondet as ed,';
-		$sql .= ' '.MAIN_DB_PREFIX.'commandedet as cd';
-		$sql .= ' WHERE';
-		$sql .= ' ed.fk_origin_line = cd.rowid';
-		$sql .= ' AND cd.fk_commande ='.$this->id;
-		//print $sql;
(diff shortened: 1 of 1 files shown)


htdocs/commande/class/commande.class.php
```diff
@@ -2175,13 +2175,12 @@ public function getNbOfShipments()
...
 		$nb = 0;
 
-		$sql = 'SELECT COUNT(DISTINCT ed.fk_expedition) as nb';
-		$sql .= ' FROM '.MAIN_DB_PREFIX.'expeditiondet as ed,';
-		$sql .= ' '.MAIN_DB_PREFIX.'commandedet as cd';
-		$sql .= ' WHERE';
-		$sql .= ' ed.fk_origin_line = cd.rowid';
-		$sql .= ' AND cd.fk_commande ='.$this->id;
-		//print $sql;
+		$sql = "SELECT COUNT(DISTINCT ee.rowid) as nb";
```
eldy, line 2178: Does not look to be a good idea. The method is called in corelation with loadExpeditions() that also get the shipment generated from the order.
Table element_element only report object that are linked together manually but link between order and shipment is stronger than that. Each lne must also be linked.

altairis-tof: @eldy i agree with you but how can we imagine to count possible empty shippings (no lines inside) ?
the use case it to display right number of shipments in order shipping tah header.
also this function is only called by order.lib.php so i do not see the link with loadExpeditions()

eldy: What is a shipment with no line inside ? This should be possible only if shipment is a draft. And a draft shipment should not be counted since the definition of a draft is something that does not exists yet.

eldy:

PR #18051, opened 2021-06-29
NEW generate intervention from time spent

altairis-tof:
Allow to generate intervention card from project time spent.
image

image


htdocs/core/class/html.form.class.php
```diff
@@ -8741,6 +8741,147 @@ public function selectExpense($selected = '', $htmlname = 'fk_c_type_fees', $use
...
+	 *  @param	int		$discard_closed Discard closed projects (0=Keep,1=hide completely,2=Disable)
+	 *  @param	int		$forcefocus		Force focus on field (works with javascript only)
+	 *  @param	int		$disabled		Disabled
+	 *  @param	string	$morecss        More css added to the select component
+	 *  @param	string	$projectsListId ''=Automatic filter on project allowed. List of id=Filter on project ids.
+	 *  @param	string	$showproject	'all' = Show project info, ''=Hide project info
+	 *  @param	User	$usertofilter	User object to use for filtering
+	 *	@return int         			Nbr of project if OK, <0 if KO
+	 */
+	public function selectIntervention($socid = -1, $selected = '', $htmlname = 'interid', $maxlength = 24, $option_only = 0, $show_empty = '1', $discard_closed = 0, $forcefocus = 0, $disabled = 0, $morecss = 'maxwidth500', $projectsListId = '', $showproject = 'all', $usertofilter = null)
```
eldy, line 8762: To try to reduce the size of html.form.class.php (keeping only method common to all features), can you move this method into html.formintervention.class.php
Also the method select_intervention seems to already exists. You should extend it instead of creating a new one.

altairis-tof: @eldy i've just modify this PR to use already existing select_interventions() function. Thanks for the advice !

PR #18053, opened 2021-06-29
Fix php8 warning

fappels:
Fix warning php 8 in product ajax.
Fix some proposal and order php 8 warning.
Invoice php 8 issue and warnings. Seems if ("" < 0) is true in php 8, already fixed in other commit, but think still better to make float from sting.
Fix some core php8 warnings. …
Fixing php 8 warnings seems also to solve some situations like a payment condition removed from dictionary can not be selected.


htdocs/compta/facture/card.php
```diff
@@ -2362,20 +2367,21 @@
 		$line = new FactureLigne($db);
 		$line->fetch(GETPOST('lineid', 'int'));
 		$percent = $line->get_prev_progress($object->id);
+		$progress = floatval(GETPOST('progress', 'int'));
```
eldy, line 2370: if progress is an int, you must use
$progress = (int) (GETPOST('progress', 'int'));
If it is a real, you must use
$progress = floatval(GETPOST('progress', 'alpha'));    // alpha to not sanitize the decimal separator

PR #18056, opened 2021-06-29
FIX Accountancy - Clean virtual zero on the import

aspangaro:


htdocs/core/modules/import/import_csv.modules.php
```diff
@@ -623,6 +623,8 @@ public function import_insert($arrayrecord, $array_match_file_to_database, $obji
 									}
 								} elseif ($objimport->array_import_convertvalue[0][$val]['rule'] == 'numeric') {
 									$newval = price2num($newval);
+								} elseif ($objimport->array_import_convertvalue[0][$val]['rule'] == 'accountingaccount') {
```
eldy, line 626: Always trimming ? Even when option ACCOUNTING_MANAGE_ZERO is set ?

PR #18060, opened 2021-06-30
FIX deposit can create credit note in payment conf

atm-arnaud:
# Fix
The problem comes from FACTURE_DEPOSITS_ARE_JUST_PAYMENTS conf. In this case we can't allowed an user to create a credit note on a deposit. This is not good because Dolibarr will decrease our sales revenue stats.

AC 5000€ not count in sales revenue
AV -5000 count in sales revenue

This is not a good method and we have to delete the credit note button in this case

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -5260,7 +5260,7 @@ elseif ($id > 0 || !empty($ref))
 			}
 
 			// Create a credit note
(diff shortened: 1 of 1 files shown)


eldy:
Note: FACTURE_DEPOSITS_ARE_JUST_PAYMENTS is a deprecated option that was introduced for backward compatibility with an old Dolibarr behaviour that is not the good way to work. It should NEVER be used anymore. By definition, using this means your data will always be wrong somewhere because it means deposit are managed in a way that it is not the way a deposit must be managed.

PR #18064, opened 2021-07-01
FIX : Update field "billed" on reopening expedition

atm-adrien:
### FIX : Update field "billed" on reopening expedition

When we reopen an expedition, the button "Cloturer" disapppear because of the condition on the "billed" field. So when it happen we can't close this expédition.

- Update the "billed" field at the same time as the "statut" field.

### DIFF
--- a/htdocs/expedition/class/expedition.class.php
+++ b/htdocs/expedition/class/expedition.class.php
@@ -2448,7 +2448,7 @@ class Expedition extends CommonObject
 
 		$oldbilled = $this->billed;
 
-		$sql = 'UPDATE '.MAIN_DB_PREFIX.'expedition SET fk_statut=1';
+		$sql = 'UPDATE '.MAIN_DB_PREFIX.'expedition SET fk_statut=1, billed=0';
 		$sql .= ' WHERE rowid = '.$this->id.' AND fk_statut > 0';
 
 		$resql = $this->db->query($sql);


eldy:
Are you using the hidden option WORKFLOW_BILL_ON_SHIPMENT ?
According to this doc https://wiki.dolibarr.org/index.php?title=Module_Shipments#In_standard_configuration, there is 2 different way of working...

atm-adrien:
Looks like I'm not using this hidden option : WORKFLOW_BILL_ON_SHIPMENT
But I think we should be able to close an expedition no matter what the situation and so no matter what the way of working

eldy:
Sure, we 

> Looks like I'm not using this hidden option : WORKFLOW_BILL_ON_SHIPMENT
> But I think we should be able to close an expedition no matter what the situation and so no matter what the way of working

So your are in standard mode. Yes we should be able to reopen. I updated the doc 
https://wiki.dolibarr.org/index.php?title=Module_Shipments#In_standard_configuration
to show the status change we want to implement.
However, i see no reason to change the "billed" status when we switch from "closed/shipped" to "to ship". May be this is the condition on the visibility of button that is wrong.

atm-adrien:
But just under the "UPDATE" SQL request, you change the value of billed ? Is that an error too ? 
image

Should I make another PR where I change the condition on the visibility of button that is wrong ?

atm-adrien:
@eldy ?

eldy:
Don't understand why we change the status to "To bill" when we reopen a shipment... Can describe a use case ?

PR #18066, opened 2021-07-02
Fix #18063 : search filters lost "Back to list"

### LINKED ISSUE #18063: Search filters are lost when using link "Back to list"
When clicking the "Back to list" link, all the previous filter values are gone, despite the "save_lastsearch_values=1" in URL.

This bug is active for any kind of list, but in this report I will show reproduce steps for Purchase orders.

1. Click "Awaiting reception" to get a list of which PO still has not arrived.

2. In the list, observe the filters "Ordered / Partially received" which are automatically added. This is correct. Then I click order KO2021-0017 to get the details of it.

3. Within the order I verify all values are okay. Then I want to go back to check the second order, so I click "Back to list" and I expect filters "Ordered / Partially received" are still selected.

4. The filters are gone and all purchase orders are therefore shown. This is not expected behaviour.

Dolibarr 12.0.5.

Hystepik:
Fix error where search filters where lost when using "Back to list" link
reuse of module builder part

### DIFF
--- a/htdocs/fourn/commande/list.php
+++ b/htdocs/fourn/commande/list.php
@@ -187,6 +187,18 @@ $arrayfields = dol_sort_array($arrayfields, 'position');
 
 $error = 0;
 
+// Initialize array of search criterias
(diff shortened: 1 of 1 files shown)


Hystepik:
not finished

PR #18068, opened 2021-07-02
fix: payment ok should trigger PAYMENTONLINE_PAYMENT_OK even on custom object

FHenry:
With custom object (other than invoice, etc...), $object is never set, so trigger is not called (stdClass do not implement call_triggers).
But we need for custom object to register payment and this is the only way to catch it


FHenry:
@nicolas-eoxia it's in the box

FHenry:
@evarisk-theo @nicolas-eoxia you can publish your module

nicolas-eoxia:
Thank you very much  @FHenry

PR #18081, opened 2021-07-02
fix: #19428 ticket substitution

lmag:
# Fix # Fix ticket substitution


lmag:
@FHenry @nicolas-eoxia
I am not sure but will see !

htdocs/core/lib/functions.lib.php
```diff
@@ -6848,6 +6848,21 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 				/*$substitutionarray['__MEMBER_NOTE_PUBLIC__'] = '__MEMBER_NOTE_PUBLIC__';
 				$substitutionarray['__MEMBER_NOTE_PRIVATE__'] = '__MEMBER_NOTE_PRIVATE__';*/
 			}
+			// add variables subtitutions ticket
+			if (!empty($conf->ticket->enabled) && (!is_object($object) || $object->element == 'ticket')) {
+				$substitutionarray['__TICKET_TRACKID__'] = '__TICKET_TRACKID__';
+				$substitutionarray['__TICKET_REF__'] = '__TICKET_REF__';
```
eldy, line 6854: To match other object, key for ref must be  '__REF__' 

__TICKET_XXX__ must be used for fields that are specific to ticket and not common to all objects.

So __USER_CREATE__ can also be managed has a common substitution key.

lmag:
Done @eldy

PR #18086, opened 2021-07-04
FIX #18085: Migration 13 to 14 : Explicit cast when changing the type…

### LINKED ISSUE #18085: [13.0.0-14.0.0.sql] column "pos" cannot be cast automatically to type integer
The 13-to-14 SQL migration script fails on PostgreSQL with the following error

```
Request 158 sql='ALTER TABLE llx_c_ticket_category MODIFY COLUMN pos integer DEFAULT 0 NOT NULL;'
--
Erreur DB_ERROR_42804: ALTER TABLE llx_c_ticket_category MODIFY COLUMN pos integer DEFAULT 0 NOT NULL;ERROR: 42804: column "pos" cannot be cast automatically to type integer HINT: You might need to specify "USING pos::integer". LOCATION: ATPrepAlterColumnType, tablecmds.c:9803
```

This is related to the Type change of the `llx_c_ticket_category.pos` column here

https://github.com/Dolibarr/dolibarr/blob/14.0/htdocs/install/mysql/migration/13.0.0-14.0.0.sql#L431

Currently the SQL request that is run against PostgreSQL is looks like this:

```sql
ALTER TABLE llx_c_ticket_category ALTER COLUMN pos TYPE INTEGER;  
```

This is not correct because PostgreSQL will not automatically convert VARCHAR to INTEGER

The correct way to change the type of the column is like this:

```sql
ALTER TABLE llx_c_ticket_category ALTER COLUMN pos TYPE INTEGER USING pos::INTEGER;  
```

daamien:
closes #18085

### DIFF
--- a/htdocs/install/mysql/migration/13.0.0-14.0.0.sql
+++ b/htdocs/install/mysql/migration/13.0.0-14.0.0.sql
@@ -428,6 +428,10 @@ create table llx_eventorganization_conferenceorboothattendee_extrafields
 ALTER TABLE llx_eventorganization_conferenceorboothattendee_extrafields ADD INDEX idx_conferenceorboothattendee_fk_object(fk_object);
 
 ALTER TABLE llx_c_ticket_category ADD COLUMN public integer DEFAULT 0;
+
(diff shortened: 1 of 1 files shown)


daamien:
If accepted this patch should be pushed to the v14 branch

eldy:
Cherry picked in v14

PR #18089, opened 2021-07-05
add filter

atm-jpb:
# new 
Addition of a "Validated + in progress" filter on the list of sales orders.

### DIFF
--- a/htdocs/commande/list.php
+++ b/htdocs/commande/list.php
@@ -514,7 +514,7 @@ if ($search_status <> '') {
 		$sql .= ' AND ((c.fk_statut IN (1,2)) OR (c.fk_statut = 3 AND c.facture = 0))'; // validated, in process or closed but not billed
 	}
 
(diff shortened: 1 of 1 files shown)


htdocs/commande/list.php
```diff
@@ -513,6 +513,9 @@
 		//$sql.= ' AND c.facture = 0'; // invoice not created
 		$sql .= ' AND ((c.fk_statut IN (1,2)) OR (c.fk_statut = 3 AND c.facture = 0))'; // validated, in process or closed but not billed
 	}
+	if ($search_status == -4) {	// To bill
```
eldy, line 516: The property status "To bill" is no more a part of the status. It is a different non correlated property. There is already a filter. An order can have any status with to bill to 1 like it can has every status with to bill to 0

atm-jpb: yes it's a wrong comment. better is  " to validate and in progress"

PR #18090, opened 2021-07-05
FIX Accountancy - Register accountancy_code for thirdparty & disable completely "product account on thirdparty"

aspangaro:


htdocs/accountancy/customer/index.php
```diff
@@ -130,34 +130,50 @@
...
-	if (!empty($conf->global->MAIN_COMPANY_PERENTITY_SHARED)) {
-		$sql .= " spe.accountancy_code_sell as company_code_sell";
-	} else {
-		$sql .= " s.accountancy_code_sell as company_code_sell";
+	$sql .= " s.tva_intra";
+	if (!empty($conf->global->ACCOUNTANCY_USE_PRODUCT_ACCOUNT_ON_THIRDPARTY)) {
+		if (!empty($conf->global->MAIN_COMPANY_PERENTITY_SHARED)) {
+			$sql .= ", spe.accountancy_code_sell as company_code_sell";
+		} else {
+			$sql .= ", s.accountancy_code_sell as company_code_sell";
```
eldy, line 145: i think this
$sql .= ", s.accountancy_code_sell as company_code_sell";
must not be in this else but also in the else of the 
if (!empty($conf->global->ACCOUNTANCY_USE_PRODUCT_ACCOUNT_ON_THIRDPARTY)) {

right or wrong ?

aspangaro: Wrong !

If we have ACCOUNTANCY_USE_PRODUCT_ACCOUNT_ON_THIRDPARTY enabled.
We look if we have MAIN_COMPANY_PERENTITY_SHARED enable.
If ok go table llx_societe_perentity. If no, use table llx_societe.

If ACCOUNTANCY_USE_PRODUCT_ACCOUNT_ON_THIRDPARTY is not present, we don't need this information.

PR #18092, opened 2021-07-05
Fix php8 warning

fappels:
on home page, in inventory module and on not set field keys


htdocs/product/inventory/card.php
```diff
@@ -180,7 +183,7 @@
 	if ($backtopage) {
 		print '<input type="hidden" name="backtopage" value="'.$backtopage.'">';
 	}
-	if ($backtopageforcancel) {
+	if (isset($backtopageforcancel)) {
```
eldy, line 186: In top of page we may have
$backtopageforcancel = GETPOST('backtopageforcancel')
In such a case, the var is defined, empty bu defined.
A test if (!empty())  seems better.

fappels: changed to !empty

PR #18099, opened 2021-07-06
FIX reading MAIN_USE_COMPANY_NAME_OF_CONTACT

st31ny:
The logic to evaluate MAIN_USE_COMPANY_NAME_OF_CONTACT was mostly
broken. The option became active when the variable was unset or not
empty, however, checking for non-emptyness should be enough.

### DIFF
--- a/htdocs/core/modules/bom/doc/doc_generic_bom_odt.modules.php
+++ b/htdocs/core/modules/bom/doc/doc_generic_bom_odt.modules.php
@@ -296,7 +296,7 @@ class doc_generic_bom_odt extends ModelePDFBom
 				$contactobject = null;
 				if (!empty($usecontact)) {
 					// On peut utiliser le nom de la societe du contact
-					if ($usecontact && ($object->contact->fk_soc != $object->thirdparty->id && (!isset($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT) || !empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT)))) {
+					if ($usecontact && ($object->contact->fk_soc != $object->thirdparty->id && !empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT))) {
 						$socobject = $object->contact;
 					} else {
 						$socobject = $object->thirdparty;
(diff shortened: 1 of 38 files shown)


htdocs/core/modules/bom/doc/doc_generic_bom_odt.modules.php
```diff
@@ -296,7 +296,7 @@ public function write_file($object, $outputlangs, $srctemplatepath, $hidedetails
 				$contactobject = null;
 				if (!empty($usecontact)) {
 					// On peut utiliser le nom de la societe du contact
-					if ($usecontact && ($object->contact->fk_soc != $object->thirdparty->id && (!isset($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT) || !empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT)))) {
+					if ($usecontact && ($object->contact->fk_soc != $object->thirdparty->id && !empty($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT))) {
```
eldy, line 299: If MAIN_USE_COMPANY_NAME_OF_CONTACT is not set, we want the same behaviour than having the value set to 1. So old code seems good to me.

st31ny: Oh, I see, but how can I disable that option then?
If I set MAIN_USE_COMPANY_NAME_OF_CONTACT to an empty value, it is deleted immediately, thus it is not set anymore and the option is active. If I set it to something non-empty, the option is active, too. So effectively, it is always active...

eldy: You can set the parameter to '0' into home - setup - other

st31ny: Yeah, sure, but then the variable is not empty but has the value `"0"`, hasn't it?

st31ny: Ah, I just checked the manual, I wasn't aware that PHP considers `"0"` the same as `false` and thus empty... Sorry for the noise then.

PR #18102, opened 2021-07-07
FIX display of extrafields on shipping lines

altairis-tof:
if MAIN_VIEW_LINE_NUMBER constant is activated the extrafields label goes in the Position column with wordbreak class on td, so it looks like this :  
image

same screenshot after fix : 
image


altairis-tof:
@eldy i've just made a better fix after you merged !

altairis-tof:
https://github.com/Dolibarr/dolibarr/pull/18112

PR #18105, opened 2021-07-07
Fix : multi module hook incompatibility

john-botella:
If this hook is used by  2 or more modules and returns 1 for both, then $head will contain data twice  because HookManager makes an array_merge

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -7752,13 +7752,14 @@ function complete_head_from_modules($conf, $langs, $object, &$head, &$h, $type, 
 	// No need to make a return $head. Var is modified as a reference
 	if (!empty($hookmanager))
 	{
-		$parameters = array('object' => $object, 'mode' => $mode, 'head' => $head);
+		$parameters = array('object' => $object, 'mode' => $mode, 'head' =>& $head);
 		$reshook = $hookmanager->executeHooks('completeTabsHead', $parameters);
 		if ($reshook > 0)
 		{
 			$head = $hookmanager->resArray;
-			$h = count($head);
 		}
+
+		$h = count($head);
 	}
 }
 


atm-florianm (approved):
À merger lorsque la PR cœur l’aura été

eldy:
If 2 hooks returns 1, it means they both want to not have anybody else defining the head and be the only one for that.
So it means there is a business/feature incompatibility. What is the rule in such case ?
My opinion is that we must keep ONLY the value of the last one (or first one). And the other (that is not compatible) has no effect. Are you ok with this rule ?

john-botella:
> So it means there is a business/feature incompatibility. What is the rule in such case ?

In my case these modules just use this hook to add a new tab with a badge for counting their elements (like files tab does). These two modules don't have to be incompatibles.

> If 2 hooks returns 1, it means they both want to not have anybody else defining the head and be the only one for that.

These both modules returns 1 because currently it's the only way to change  $head, but with this patch they will return 0 without having to force overriding, they just add their modifications they dont need to erase modifications made by others modules.
I agree with the fact : _if a module return 1 so he want to rules them all_,  and this patch allow it but keep  possibility to make little changes before by others modules too if they dont return 1.

> My opinion is that we must keep ONLY the value of the last one (or first one). And the other (that is not compatible) has no effect.

keep the last one is good only if the value given to him was already changed by others modules to be sure the last one can do what it want and why not overriding.
For me, modules works with each other with hooks, not individually 

You can see an example of this hook usage in class/actions_webpassword.class.php of module webpassword

eldy:
I pushed a different fix. 
@atm-john @atm-florianm Can you try it on your case and tell if it works. Ps: Module hooks must return 0 to have entries concatenated. When 1 is returned, it replaces completely standard behaviour (so standard entries).

john-botella:
I saw the new code. Seems ok for me.

PR #18107, opened 2021-07-07
add filter by thirdparty ids for thirdparty API

p-payet:
# Fix # Add filter by thirdparty ids for thirdparty API
Add filter by thirdparty ids for thirdparty API in htdocs/societe/class/api_thirdparties.class.php index function.


eldy:
Why not using the generic property  sqlfilters  for that ?

PR #18109, opened 2021-07-08
FIX : don't display custom masks if they are not used

altairis-noe:
While testing v14, it occurred to us that the custom mask field was not properly hidden in product card edit.

Here when the lot product mask is inactive
hiddenField
Here when the serial number product mask is activated
visibleField

### DIFF
--- a/htdocs/product/card.php
+++ b/htdocs/product/card.php
@@ -1636,14 +1636,14 @@ if (is_object($objcanvas) && $objcanvas->displayCanvasExists($action)) {
 						$tooltip .= '<br>'.$langs->trans("GenericMaskCodes4a", $langs->transnoentities("Batch"), $langs->transnoentities("Batch"));
 						$tooltip .= '<br>'.$langs->trans("GenericMaskCodes5");
 						print '<tr><td id="mask_option">'.$langs->trans("ManageLotMask").'</td>';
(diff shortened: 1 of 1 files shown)


htdocs/product/card.php
```diff
@@ -1649,29 +1649,37 @@
...
-							$("#field_mask, #mask_option").addClass("hideobject");
-							var preselect = $("#status_batch option:selected");
-							if (preselect !== "0") {
-								$("#field_mask, #mask_option").toggleClass("hideobject");
-							}
-							$("#status_batch").on("change", function () {
-								console.log("We change batch status");
+							$("#field_mask").parent().addClass("hideobject");
+							var preselect = document.getElementById("status_batch");';
+						if ($conf->global->PRODUCTBATCH_SN_USE_PRODUCT_MASKS)
```
eldy: From v14, use 
if (getDolGlobalString('PRODUCTBATCH_SN_USE_PRODUCT_MASKS'))
instead of 
if ($conf->global->PRODUCTBATCH_SN_USE_PRODUCT_MASKS)
So it will avoid errors with php8

altairis-noe: sure thing. I suppose every $conf->global is concerned, even when it is not an if condition ?

eldy: Yes. But this will be fixed step by step, commit after commit. This is a long term task.

altairis-noe: Done here, I didn’t plan on refurbishing the all product card, indeed, just did it in the section I was modifying in this particular PR

PR #18112, opened 2021-07-08
better fix for extrafields

altairis-tof:
better fix for https://github.com/Dolibarr/dolibarr/pull/18102#event-4995386237


htdocs/core/class/commonobject.class.php
```diff
@@ -7492,10 +7492,8 @@ public function showOptionals($extrafields, $mode = 'view', $params = null, $key
 
 						if ($display_type == 'card') {
 							$out .= '<tr '.($html_id ? 'id="'.$html_id.'" ' : '').$csstyle.' class="valuefieldcreate '.$class.$this->element.'_extras_'.$key.' trextrafields_collapse'.$extrafields_collapse_num.(!empty($this->id)?'_'.$this->id:'').'" '.$domData.' >';
-							if ( ! empty($conf->global->MAIN_VIEW_LINE_NUMBER) ) {
-								$out .= '<td></td>';
-							}
-							$out .= '<td class="wordbreak';
+							if ( ! empty($conf->global->MAIN_VIEW_LINE_NUMBER) ) $out .= '<td class="';
+							else $out .= '<td class="wordbreak';
```
eldy, line 7496: What was wrong with previous code ?

altairis-tof:
it was adding a td before any card extrafield.
the real problem is that shipping lines are using "card" parameter when calling showoptionals function.

eldy:
> showoptionals

May be this is that the trouble. It should use "line" instead of "card". What happen if you fix by switching into "line" ?

altairis-tof:
the "line" parameter generate a div and not a tr; so the div goes above the shipping lines table and not below the shipping line itself.

altairis-tof:
@eldy i think this fix is important

eldy:
I have some reason to think this may be fixed into more recent version. Does this bug still occurs with last stable version ?
If not, can you close ticket ?
If yes, can you confirm the last stable version you reproduced the bug with ?

altairis-tof:
@eldy yes the bug is still relevant (copac) and this PR fix it.

PR #18122, opened 2021-07-09
FIX : still one if too many for custom masks in product card

altairis-noe:
After further testing, my last PR #18109 yesterday was incomplete. An "if" would prevent displaying the options as we want them if we edited a product previously set as not needing a serial or lot number. This fixes it.

I don’t add screenshots because it is pretty much the same as yesterday.


eldy:
A better fix was pushed.

PR #18125, opened 2021-07-09
NEW add hookmanager on note pages

fr69400:


htdocs/adherents/note.php
```diff
@@ -68,15 +68,20 @@
...
+
 // Security check
 $result = restrictedArea($user, 'adherent', $object->id, '', '', 'socid', 'rowid', 0);
 
-
 /*
  * Actions
  */
+$reshook = $hookmanager->executeHooks('doActions', array(), $object, $action); // Note that $action and $object may have been modified by some hooks
+if ($reshook < 0)
```
eldy, line 80: To match our PSR and checkstyle rules, can you replace with
if ($reshook < 0) {
...
}

The { must be mentionned, even if there is only one instruction into if.

htdocs/adherents/note.php
```diff
@@ -68,15 +68,20 @@
...
+	setEventMessages($hookmanager->error, $hookmanager->errors, 'errors');
 
-include DOL_DOCUMENT_ROOT.'/core/actions_setnotes.inc.php'; // Must be include, not include_once
+if (empty($reshook))
```
eldy, line 83: Same point.

PR #18145, opened 2021-07-13
FIX : method_exists needs object at first param

atm-gauthier:


atm-maxime:
The if statement seems useless, same code is in if and else...

eldy:
> The if statement seems useless, same code is in if and else...

@atm-maxime The case of the method name differs. In the doubt, i will keep it.

PR #18159, opened 2021-07-16
New : Language support for knowledge record 

Hystepik:
Support of a language field to give the language of the question and the solution


htdocs/install/mysql/migration/13.0.0-14.0.0.sql
```diff
@@ -596,3 +596,4 @@ create table llx_onlinesignature
   pathoffile				varchar(255)
 )ENGINE=innodb;
 
+ALTER TABLE llx_knowledgemanagement_knowledgerecord ADD default_lang varchar(6)
```
eldy, line 599: Add the ; at end so a new line added by another PR or a merge will not generate syntax error;

htdocs/install/mysql/tables/llx_knowledgemanagement_knowledgerecord.sql
```diff
@@ -21,6 +21,7 @@ CREATE TABLE llx_knowledgemanagement_knowledgerecord(
 	date_creation datetime NOT NULL, 
 	tms timestamp, 
 	last_main_doc varchar(255), 
+	default_lang varchar(6),
```
eldy, line 24: Rename the field "lang" instead of "default_lang" because it is the language of article not a default that can be overwrite.

htdocs/knowledgemanagement/knowledgerecord_list.php
```diff
@@ -117,6 +118,15 @@
 			'position'=>$val['position'],
 			'help'=> isset($val['help']) ? $val['help'] : ''
 		);
+	} elseif ($key = 'lang') {
```
eldy, line 121: Include this test into the 
if (!empty($val['visible'])) {
and do the content of the if (!empty...) into the else.

htdocs/knowledgemanagement/knowledgerecord_card.php
```diff
@@ -332,6 +353,15 @@
 	//unset($object->fields['fk_soc']);					// Hide field already shown in banner
 	include DOL_DOCUMENT_ROOT.'/core/tpl/commonfields_view.tpl.php';
 
+	//Language of question/response
```
eldy, line 356: Can you include this into the commonfields_view.tpl.php so every module that use a 'lang' field will benefit the language selector/beautifier.

htdocs/knowledgemanagement/knowledgerecord_card.php
```diff
@@ -185,6 +192,13 @@
 	// Common attributes
 	include DOL_DOCUMENT_ROOT.'/core/tpl/commonfields_add.tpl.php';
 
+	//Language of question/response
```
eldy, line 195: Can you include this into the commonfields_add.tpl.php so every module that use a 'lang' field (name reserved) will benefit the language selector/beautifier.

htdocs/knowledgemanagement/class/knowledgerecord.class.php
```diff
@@ -115,6 +115,7 @@ class KnowledgeRecord extends CommonObject
 		//'url' => array('type'=>'varchar(255)', 'label'=>'URL', 'enabled'=>'1', 'position'=>55, 'notnull'=>0, 'visible'=>-1, 'csslist'=>'tdoverflow200', 'help'=>'UrlForInfoPage'),
 		'fk_c_ticket_category' => array('type'=>'integer:CTicketCategory:ticket/class/cticketcategory.class.php', 'label'=>'GroupOfTicket', 'enabled'=>'$conf->ticket->enabled', 'position'=>512, 'notnull'=>0, 'visible'=>-1, 'help'=>'YouCanLinkArticleToATicketCategory'),
 		'status' => array('type'=>'integer', 'label'=>'Status', 'enabled'=>'1', 'position'=>1000, 'notnull'=>1, 'visible'=>1, 'default'=>0, 'index'=>1, 'arrayofkeyval'=>array('0'=>'Draft', '1'=>'Valid'),),
+		'lang' => array('type'=>'varchar(6)', 'label'=>'Language', 'enabled'=>'1', 'position'=>51, 'notnull'=>0, 'visible'=>0,),
```
eldy, line 118: try to restore visible to 1 because this property may be used for other need than the test into the .tpl.php file.
Because we want field visible, we should keep it to 1.

htdocs/knowledgemanagement/knowledgerecord_card.php
```diff
@@ -225,6 +239,13 @@
 	// Common attributes
 	include DOL_DOCUMENT_ROOT.'/core/tpl/commonfields_edit.tpl.php';
 
+	//Language of question/response
```
eldy, line 242: Same.
Can you include this into the commonfields_edit.tpl.php so every module that use a 'lang' field will benefit the language selector/beautifier.

PR #18163, opened 2021-07-16
NEW: Separate read-only and adding time on projects

FHenry:
Issue : #18162


eldy:
Can you provide an example of use case for this need ?

FHenry:
@eldy Input spend time should be not the same as read task time. External user, for exemple, should be able see task time, but input spent time should different.

FHenry:
@eldy Could you have a look ?

htdocs/core/modules/modProjet.class.php
```diff
@@ -202,6 +202,12 @@ public function __construct($db)
 		$this->rights[$r][4] = 'all';
 		$this->rights[$r][5] = 'supprimer';
 
+		$r++;
+		$this->rights[$r][0] = 145; // id de la permission
+		$this->rights[$r][1] = "Can enter time consumed on assigned tasks (timesheet)"; // libelle de la permission
```
eldy, line 207: So the label of permission 41 and permission 142 should be modified. They already bring permission to enter time.
If you move this into a dedicated permission, you must update any use + any label of permission 41 but also 142

PR #18167, opened 2021-07-16
FIX: line extrafields are inoperative in dispatch cards even when they exist

atm-florianm:
## Issue
Currently, supplier order dispatch lines can have extrafields, but they are not displayed and not editable using the standard interface.

## Fix
This fix tries to address the issue by enabling the user to edit the extrafields when creating a new dispatch object. Their value can be pre-filled from HTTP query parameters or (but I didn't test this case) using the value from the origin supplier order line if it has an extrafield with the same name.

## Further fix
We may also want to fix the "reception" tab of the supplier order so that the extrafield values can be pre-set. There is also a minor bug on that page (cloned date pickers don't work when you dispatch one supplier order line to more than one warehouse) which this PR doesn't fix. However, this PR should be fine on its own.

### DIFF
--- a/htdocs/reception/card.php
+++ b/htdocs/reception/card.php
@@ -1183,10 +1183,10 @@ if ($action == 'create')
 					$srcLine->fetch_optionals(); // fetch extrafields also available in orderline
 					$line->fetch_optionals();
 
-					$line->array_options = array_merge($line->array_options, $srcLine->array_options);
 					if (empty($line->array_options) && !empty($dispatchLines[$indiceAsked]['array_options'])) {
 						$line->array_options = $dispatchLines[$indiceAsked]['array_options'];
 					}
+					$line->array_options = array_merge($line->array_options, $srcLine->array_options);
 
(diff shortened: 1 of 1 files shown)


htdocs/reception/card.php
```diff
@@ -1164,6 +1184,9 @@
 					$line->fetch_optionals();
 
 					$line->array_options = array_merge($line->array_options, $srcLine->array_options);
+					if (empty($line->array_options) && !empty($dispatchLines[$indiceAsked]['array_options'])) {
+						$line->array_options = $dispatchLines[$indiceAsked]['array_options'];
```
eldy, line 1188: So if there is some extrafields submited into form, it will erase completely all $line->array_options ?
May be we can make this 3 new lines before the line 1186 instead of after ?
So submited data are added,  and extrafields coming from $srcLine are also used (current behavior).

atm-florianm: You’re right, I moved the lines according to your suggestion.

PR #18182, opened 2021-07-20
NEW : field and behavior in product object

atm-jpb:
# NEW addition of a mandatory period field in the product object when it is of type service.

Message to the user on the need to enter a start date and an end date on a service when creating / validating an invoice, commercial proposal or  sales order. 
This action is not blocking in the process of confirmation

the field in product (service type)
service-2

when we add a service to propal/invoice or commande  we add an underline on date start and end. (non blocking situiation) 
add product underline

same thing on update line product
update_product_date_underline

on description lines, we have now a warning on service with mandatory_Period checked and no date entered. 
propal-line-warning

if we have a service with mandatoryPreriod checked and no date entered (or just one)  we inform the user.
the situation is not blocking for validation
confirmMessage


htdocs/core/tpl/objectline_create.tpl.php
```diff
@@ -675,13 +676,26 @@ function checkFreeLine(e, npRate)
 	}
 	?>
 
+<?php if ($this->table_element_line != 'commande_fournisseurdet' || $this->table_element_line == 'facture_fourn_det') { ?>
```
eldy, line 679: This test looks strange
!=   ||   ==
Can you check

atm-jpb: yep  ;) i removed the second one. thx

atm-arnaud:
@eldy Is it good for you ?

atm-jpb:
@eldy made an udpate

htdocs/core/tpl/objectline_create.tpl.php
```diff
@@ -718,6 +752,7 @@ function(data) {
 							if (!empty($conf->global->FCKEDITOR_ENABLE_DETAILS)) { ?>
 						if (typeof CKEDITOR == "object" && typeof CKEDITOR.instances != "undefined")
 						{
+							var editor = CKEDITOR.instances['dp_desc'];
```
eldy, line 755: Duplication of line is required ?

atm-jpb:
@eldy i made the change. it's up to you

htdocs/core/tpl/objectline_create.tpl.php
atm-jpb: rmeove duplicated line

atm-jpb:
@eldy i made the change you asked before, can you give me your feedback on it ?

htdocs/core/tpl/objectline_view.tpl.php
```diff
@@ -182,6 +182,22 @@
 		if ($line->date_start || $line->date_end) {
 			print '<div class="clearboth nowraponall opacitymedium">'.get_date_range($line->date_start, $line->date_end, $format).'</div>';
 		}
+		if (!$line->date_start || !$line->date_end) {
+			// show warning under line
+			// we need to fetch product associated to line for some test
+			if ($object->element != 'order_supplier') {
```
eldy, line 188: Why test 
if ($object->element != 'order_supplier') {
and not 
supplier_proposal or and any other possible values ?
It is a generic class and lot of external module also use this tpl file;

May be we should invert the test to make the field mandatory when it is in a list of known object (proposal, sale order and invoices only) ?

atm-jpb: made changes

atm-jpb:
@eldy i made the change on your feedback. can you see if it's ok for you ?

PR #18202, opened 2021-07-22
Compatibility phpv8

TheGivrix:


htdocs/societe/list.php
```diff
@@ -551,46 +551,46 @@
 if ($search_id > 0) {
 	$sql .= natural_search("s.rowid", $search_id, 1);
 }
-if ($search_nom) {
+if (!empty($search_nom)) {
```
eldy, line 554: Did you get warning due to this line ?
$search_nom seems to be always defined (set at top of file) so i don't see why you should need this here.

htdocs/societe/index.php
```diff
@@ -367,7 +367,7 @@
 $boxlist .= '</div>'."\n";
 
 $boxlist .= '<div class="secondcolumn fichehalfright boxhalfright" id="boxhalfright">';
-$boxlist .= $lastmodified;
+$boxlist .= empty($lastmodified) ? '' : $lastmodified;
```
eldy, line 370: $lastmodified  is previously set and sometimes not.  
A better fix would be to initialise to '' at top of file so it is always defined and 
$boxlist .= $lastmodified;
will be enough.

eldy:

PR #18213, opened 2021-07-23
Compatibility phpv8

TheGivrix:


fappels:
Think php8 fix also accepted for version 14

eldy:

PR #18217, opened 2021-07-25
FIX: Salary - link 'salary' of bank_url  isn't managed in module bank & accountancy

aspangaro:
Return to type 'user' in bank_url to resolve some problems.

In bank entries :
image

In accountancy before transaction inscription :
image

### DIFF
--- a/htdocs/salaries/card.php
+++ b/htdocs/salaries/card.php
@@ -226,7 +226,8 @@ if ($action == 'add' && empty($cancel)) {
 			$paiement->amounts      = array($object->id=>$amount); // Tableau de montant
 			$paiement->paiementtype = $type_payment;
 			$paiement->num_payment  = GETPOST("num_payment", 'alphanohtml');
-			$paiement->note = GETPOST("note", 'none');
+			$paiement->note			= GETPOST("note", 'none');
+			$paiement->fk_user		= GETPOST("fk_user", 'int') > 0 ? GETPOST("fk_user", "int") : 0;
 
 			if (!$error) {
 				$paymentid = $paiement->create($user, (int) GETPOST('closepaidsalary'));
--- a/htdocs/salaries/class/paymentsalary.class.php
+++ b/htdocs/salaries/class/paymentsalary.class.php
@@ -572,8 +572,6 @@ class PaymentSalary extends CommonObject
 				// Add link 'user' in bank_url between user and bank transaction
 				if (!$error) {
 					foreach ($this->amounts as $key => $value) {
-						dol_syslog(get_class($this) . '::Insertuser ' . $key);
-
 						if ($mode == 'payment_salary') {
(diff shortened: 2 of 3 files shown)


htdocs/salaries/card.php
```diff
@@ -226,7 +226,8 @@
 			$paiement->amounts      = array($object->id=>$amount); // Tableau de montant
 			$paiement->paiementtype = $type_payment;
 			$paiement->num_payment  = GETPOST("num_payment", 'alphanohtml');
-			$paiement->note = GETPOST("note", 'none');
+			$paiement->note			= GETPOST("note", 'none');
+			$paiement->fk_user		= GETPOST("fk_user", 'int') > 0 ? GETPOST("fk_user", "int") : 0;
```
eldy, line 230: There is no property fk_user on a payment of salary. Payment of salary is done on a salary and the user paid is fk_user on salary object.

htdocs/salaries/class/paymentsalary.class.php
```diff
@@ -536,14 +569,26 @@ public function addPaymentToBank($user, $mode, $label, $accountid, $emetteur_nom
 					}
 				}
 
-				// Add link 'company' in bank_url between invoice and bank transaction (for each invoice concerned by payment)
-				$linkaddedforthirdparty = array();
-				foreach ($this->amounts as $key => $value) {
-					if ($mode == 'payment_salary') {
-						$salary = new Salary($this->db);
-						$salary->fetch($key);
-						$result = $acc->add_url_line($bank_line_id, $salary->id, DOL_URL_ROOT.'/salaries/card.php?id=', '('.$salary->label.')', 'salary');
```
eldy, line 545: We should keep the link with the salary. Later we will be able to guess the user from this object.

aspangaro: I'm not ok with you. In fact, you have alrealdy a link to payment salary in first column and the column Thirdparty in page bankentries is for thirdparty/user... not to have a link to salary. It's the same functionality than expense report module

image

htdocs/salaries/card.php
aspangaro: I'm agree that fk_user has moved to llx_salary in v14 but I needed fk_user for $paiement->addPaymentToBank() (line 242) to give employee id in bank_url.

At the base, with type salary, it's doesn't work too...

eldy: If you have the payment (payment id of full object), you can guess the salary so the employee.
Note: setting non existing properties is nonmore allowed in php8.

htdocs/salaries/class/paymentsalary.class.php
eldy: You're right.
Let's add the add_url_line for user and removed the one on salary.
I pushed the fix manually because i introduced also other fixes.

PR #18224, opened 2021-07-26
FIX: Missing column Date validation in ledger & subledger

aspangaro:


htdocs/accountancy/bookkeeping/listbyaccount.php
```diff
@@ -144,6 +152,7 @@
 	't.credit'=>array('label'=>$langs->trans("Credit"), 'checked'=>1),
 	't.lettering_code'=>array('label'=>$langs->trans("LetteringCode"), 'checked'=>1),
 	't.date_export'=>array('label'=>$langs->trans("DateExport"), 'checked'=>1),
+	't.date_validated'=>array('label'=>$langs->trans("DateValidation"), 'checked'=>1),
```
eldy, line 155: What is this date ? Is it for a feature of lettering enabled by option ACCOUNTING_ENABLE_LETTERING or by something else ?

aspangaro: No, this does not concern the lettering of accounts. It is related to the FEC format, it is requested to validate the accounting entries and to block them in deletion / modification. Without this functionality, a FEC format can be totally validated

image

eldy: My question was Where and when can we set this date into application ? On which page ? I can't find. Do we have to set an option or beeing in dev lide to be able to set this date ?

aspangaro: This date are set when you export on journal or when you use accoutancy > close menu

PR #18234, opened 2021-07-26
fix for php 8

ptibogxiv:


eldy:
I suggest to return '' instead of null when not defined so token is always a string (no side effect on code that is using it)

ptibogxiv:
> I suggest to return '' instead of null when not defined so token is always a string (no side effect on code that is using it)

done !

PR #18244, opened 2021-07-27
Fix php debug bar for php 8.0

Hystepik:


htdocs/includes/symfony/var-dumper/Cloner/VarCloner.php
```diff
@@ -291,7 +291,7 @@ private static function initHashMask()
             if (!empty($frame['line'])) {
                 ob_start();
                 debug_zval_dump($obj);
-                self::$hashMask = substr(ob_get_clean(), 17);
+                self::$hashMask = strlen(substr(ob_get_clean(), 17));
             }
```
eldy, line 295: strlen will always return 17. Is it the expected change ?
What decides you to fix like this ?

Hystepik: Yes, I agree with you strlen always return 17, the problem was with the string. Actually,when we remove the strlen() self::$hashMask is a string and the next line (line:295) always encounter a fatal error which crash the debugbar. I decided to make self::$hashMask an int to prevent the crash. But it seems that it isn't a great fix because the hash isn't that random anymore.

eldy: This file is part of an external lib of symphony project. Did you report bug to original project ? Did you search into original project if a fix already exists for php 8 ?

ksar-ksar: In package V3.2 they just put `(int)` to cast to int

https://github.com/symfony/var-dumper/blob/3.2/Cloner/VarCloner.php#L320

Hystepik:
I tried to update var_dumper to 3.2 with the repo for a better fix

PR #18247, opened 2021-07-27
Fix a bug in ecm

Hystepik:
Fix a bug in ecm module where files were not displayed.


eldy:
Pr was pushed in develop ? Is bug also present in v14 branch ? If yes, can you push in v14 branch ?

Hystepik:
Yes the bug was on v14 too. Look #18270

PR #18248, opened 2021-07-27
NEW add agenda auto creation on supplier_order cancellation

cfoellmann:
This creates the automatic agenda entry for cancellation of supplier orders

makes everything more consistent.


cfoellmann:
@eldy I am not sure if the sql update is correct or even in the correct file

eldy:
Yes, update is ok and into correct file. But you must also add an entry into file llx_c_action_trigger.sql (for first install)

cfoellmann:
@eldy merge conflicts solved.

PR #18253, opened 2021-07-27
FIX manage event in select_company with AJAX search option on

kkhelifa-opendsi:
FIX manage event in select_company with AJAX search option on


htdocs/core/lib/ajax.lib.php
```diff
@@ -516,22 +537,19 @@ function(response) {
 							if (response.num) {
 								var selecthtml_str = response.value;
 								var selecthtml_dom=$.parseHTML(selecthtml_str);
-								if (typeof(selecthtml_dom[0][0]) !== \'undefined\') {
```
eldy, line 519: Why removing this test ?

PR #18282, opened 2021-07-29
FIX: Manage credit note on situation invoice for calculate margin

aspangaro:
When you realize a credit invoice on situation invoice, margin are badly calculate. Had a way to calculate.

Type SITUATION (5) move to Type CREDIT_INVOICE (2)...


eldy:
|| and && are mixed without parebthesis. Result seems to not be the one expected on some php versions.

aspangaro:
Review ok. Ping @eldy

PR #18283, opened 2021-07-29
NEW : Stocks transfers objects management

atm-gauthier:
This module was created to save stock transfers information into a new object, and generate a PDF transfer sheet which contains adresses of source warehouse and destination warehouse and products details (quantity, product source location, product destination location, PMP, etc...)

image
image
image
image
image
image
image
image
image


atm-maxime:
At least a bit of a description to explain what this new module does. And/or screenshots.
Beggining by creating a feature request explaining what is needed would be a good way too...

BebZ:
Hello, 

That's a great module to keep track of stock movements ! 
Lot/série or Batch/DLUO field is a simple "input" or a dropdown to select existing ones in stocks ? 

image


Thanks

atm-gauthier:
Hi @BebZ, thank you for your comment ! ;)
for this first version it is a simple input as well as when you click on "Correct stock" on product stock tab, but you're right, in the future it should be nice to have an autocompletion for batch.

image

BebZ:
thanks ;) ! 
True that in "core" dolibarr, no dropdown or autocompletion... would it be hard to do ?
I think one of us in our company could make it :)
is there a specific function to implement #CheckExistingBatchInStock or smthg ?

eldy:
We can have a new process to record stock transfer. I will use the name of "Stock transaction" that is a "group of several stock movements" between 2 warehouses.

Currently, there is already 4 things to make group of stock movements:
- A) The shipment object
- B) The reception object
- C) The inventory object
- D) And for things that are not related to this, for a transfer from warehouse A to B, we have the stock movement table that contains the field "Inv./Mov. code" that is the key that group unit stock movement together to represent a Stock transaction of several movements". The feature to record such a group of movement is in the menu "Mass stock transfer" can also be used and can also be recorded manually with unit stock movement using the same inventory/code

So this module add another way to group stock movement. But the feature is very similar to the "Mass stock movement" feature. The difference i see is that we can "prepare a mass stock transfer" to do it really later (and attach document). The current feature "mass stock transfer" is a quick/realtime stock transfer. We can imagine that we generate a PDF from the  "Inv./Mov. code" too with no need of new table and no need of new module on a transaction (the PDF will be the same).

So this module must be presented as a "advanced stock transfer" feature compared to the existing one. The name of such a module is important to avoid to have user trying to use it when it is no necessary. I suggest to summarize the feature of this module with this name:
   "Multi-Step mass stock transfer"     (multi-step because we can prepare first, then complete, then finally validate)
And the name of object managed may be 
  "Prepared Stock Transfer"

Also I suggest to have the feature just under the menu entry "Mass stock transfer" with the menu name "Mass stock transfer (multi-step)". So user will see immediatly there is 2 solutions and understand immediatly it is a similar feature but the second one provides advanced options and workflow.

This is some question about current implementation:
- When you record the stock movements, do you fill fk_origin and origintype with id of this new object into the table llx_stock_mouvement ?
- Which value do you use for the "Inv/Mov. code" in llx_stock_movement ?

atm-gauthier:
Hi @eldy, of course when stock movements are done with my module, origin id and origin type are set into llx_stock_movement, but I let Inv/Mov. code empty for the moment.

I confirm that the principal use of this module is to prepare stock transfers (with dates of actions), to do movements in 2 times, and to generate transfer sheet which allows logistics officer to know what he will have to move, where he will find each item, and where he will have to move them.
Several of our customers asked us this module, then I created it and we thought lots of dolibarr users could like this too.

So what what would you like me to do now?
You talked about not using new tables. But current simple mass stock transfer feature don't save lines until movements are done (only stored in session if I'm right), then it's not possible to use the same method with my module, I do need to save lines in a new table (to be able to prepare and to see them when we want before the real actions), as well as all commercial object in dolibarr.

eldy:
So it looks ok for me. For the moment, this is what you can do:
- Fill a value for "Inv/Mov. code" (I suggest to use a string YYYYMMDDHHMMSS). A movement code must ALWAYS be provided for each record into llx_stock_movement as it is a key in several screen to see grouped movements.
- Add menu entry "Mass stock transfer (multi-step)" just under the existing menu entry "Mass stock transfer"
- Solve conflict of PR.
- Check there is no Travis CI error
I will validate the PR once all of this is ok.

atm-maxime:
My comments :
Not sure about the usefulness of the "Inv/Mov. code" as long as all movements are linked to the stock transfer object, but it can be added in the form confirm as the movement label is asked and pre-filled
Regarding the menu, I think a separated menu (Stock transfer > New, List > Draft, Validated, Closed) having a link "New quick transfer" still going on the current screen would be better. And of course keeping the current menu if the new module is not active.

atm-gauthier:
Hi @eldy, hi @atm-maxime 
I filled inventory code like @atm-maxime said, i think it's a good way :
image

I also had to change origintype of llx_stock_movement from varchar(32) to varchar(64) to be able to save complete string "StockTransfer@product/stock/stocktransfer" (necessary for dolibarr to be able to show origin object if existing)

For the menu, what do i finally have to do ? did you see @atm-maxime comment ?

To finish, travis seems wrong about 2 last feedbacks, parameters are well filled

atm-gauthier:
Hi @eldy, do you have new feedback about my PR or everything's ok?

atm-gauthier:
@eldy did you check my PR ?

atm-gauthier:
@eldy all travis checks are ok now with this PR, and stickler too, i don't know why it's red...

atm-gauthier:
@eldy are you going to merge this module or is there still something you want me to modify ?

atm-thibaultf:
@eldy could you tell us if this module can be merged or if we need to modify something ?

cltvchannel:
@eldy Will be great if this module will be released in v15.

atm-maxime:
@eldy what about this PR ?

eldy:
@atm-gauthier or @atm-thibaultf 
Can you help me to fix the conflict of this PR ?  I plan to have this PR into v16

atm-gauthier:
Hi @eldy, i just merged develop into my branch

htdocs/install/mysql/tables/llx_stocktransfer_stocktransfer.key.sql
```diff
@@ -0,0 +1,29 @@
+-- Copyright (C) ---Put here your own copyright and developer email---
```
eldy, line 1: With v16, new tablesare now installed only when module is enabled, so you must:
- remove any code to create table into migration (no more need).
- rename your table llx_aaa_bbb.sql  into llx_aaa_bbb-modulename.sql
- you must include the lines at begin of method public function init($options = '') in you rmodule descriptor file :
 		$result = $this->_load_tables('/install/mysql/', 'mymodule');
		if ($result < 0) {
			return -1; // Do not activate module if error 'not allowed' returned when loading module SQL queries (the _load_table run sql with run_sql with the error allowed parameter set to 'default')
		}

So all sql files with -mymodule into its name are executed during first module activation.

atm-gauthier: Hi @eldy, i just did it, it seems to work

atm-maxime:
Can you try to add "// phpcs:enable" in functions having "// phpcs:disable" ? I think that's why travis is angry

atm-gauthier:
@atm-maxime @eldy i tried to add missing phpcs:enable in the pdf_eagle.modules.php file and travis is always angry.
I really don't know what to do now...

atm-gauthier:
Congrats @atm-maxime :D 
@eldy now everything's ok with this PR.

lmag:
👍 👍 👍

PR #18285, opened 2021-07-29
add missing trigger ORDER_SUPPLIER_DISPATCH 

cfoellmann:
add missing trigger ORDER_SUPPLIER_DISPATCH to fire when a supplier order is received.

### DIFF
--- a/htdocs/fourn/commande/card.php
+++ b/htdocs/fourn/commande/card.php
@@ -1073,6 +1073,8 @@ if (empty($reshook)) {
 				if ($result > 0) {
 					$langs->load("deliveries");
 					setEventMessages($langs->trans("DeliveryStateSaved"), null);
+
+					$trigger = $object->call_trigger('ORDER_SUPPLIER_DISPATCH', $user);
 					$action = '';
 				} elseif ($result == -3) {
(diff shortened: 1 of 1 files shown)


fappels:
Did you try trigger 'ORDER_SUPPLIER_SUBMIT'?

cfoellmann:
> Did you try trigger 'ORDER_SUPPLIER_SUBMIT'?

My description above was false (changed now). My trigger is NOT for when the order is sent to the supplier BUT for when the products arrive,
We send a notification to the creator of the supplier_order to tell that the items have arrived.

Should be merge because there really is no trigger firing at that point

fappels:
$object->Livraison also calls trigger ORDER_SUPPLIER_RECEIVE
ORDER_SUPPLIER_DISPATCH trigger is called in dispatch.php but do not understand why. Just before $object->calcAndSetStatusDispatch also is called. This one also calls ORDER_SUPPLIER_RECEIVE. So think ORDER_SUPPLIER_DISPATCH is obsolete.

cfoellmann:
Thanks. You are totally right.

I don't know why I didn't see this.

Issue #18292, opened 2021-07-30
Missing Table in database : llx_product_lang_extrafields


lmag:
# Feature Request
Missing table llx_product_lang_extrafields

## Use case
Uses traduction with extrafields

## Suggested steps
First : Just add table
Second : Do code missing in product class

lmag:
@FHenry @nicolas-eoxia

PR #18293, opened 2021-07-30
NEW: add table product langs extrafields #18292

FHenry:
NEW: add table product langs extrafields #18292


eldy:
What is goal of this table. I can't guess by reading the fields of table.

FHenry:
@eldy Goal is to add extrafields on product lang table to allow better management of translation for item extrafields. For exemple in the Dolibarr CMS if we want to develop e-commerce site, we should need more than just label and description translatable, and it can depend on each customer. There is the hidden field "other", but it is better to have an extrafields table like all other object in Dolibarr

eldy:
Do you mean you want to store n values (1 value per language) for the extrafields of poducts ?
Do you have example of such data / use case ?

lmag:
The practical case consists in making the organization of WordPress, [WPML](https://wpml.org/fr/), WPshop and Dolibarr homogeneous.
By default the relationship between Dolibarr and WPShop is :
1 Post(Id1 de WP) = 1 produit/service(Id1 de doli)

with WPML
1 Post(ID1 de WP) FR
-1 Post ID2 de WP pour EN
-1 Post ID3 de WP pour US
-1 Post ID4 de WP pour IT
-Etc

We plan to keep this organization by adding the data of the WordPress translation POST ID in the extrafields table like this1 Post(ID1) FR = 			1 produit/service(Id1) FR
-1 Post ID2 pour EN				-La traduction pour EN liée par l'ID2 de WP
-1 Post ID3 pour US				-La traduction pour US liée par l'ID3 de WP
-1 Post ID4 pour IT				        -La traduction pour IT liée par l'ID4 de WP
-Etc

lmag:
@eldy @nicolas-eoxia

FHenry:
@eldy Yes that's it.
Extrafield of product_langs for exemple "name_bis" could have a translation in many language.
Even if "name_bis" is not a product attribute, it can be a product translatable attribute.

I work on another branch [WIP branch](https://github.com/FHenry/dolibarr/tree/dev_product_langs_extrafields_corechange) that allow to edit them, but I prefer to first add db change and then core change

eldy:
Do you know there is already a table called llx_object_lang dedicated to store alternative values for other language.
The product does not use this table yet, but in future, we should because it allows to support translation for anyfield any table (it is already working correctly for exemple on company when option PDF_USE_ALSO_LANGUAGE_CODE is on).
We can image storing also alternate value for extrafields (for example when field property is "options_xxx")

But well, ideally, the best is first migrate translation of product into this table...

PR #18297, opened 2021-07-30
NEW add hook 'menuLeftMenuItems' to filter the leftmenu items

cfoellmann:
add hook 'menu_LeftMenuArray' so we can manipulate the `leftmenu` from any module

Usage example:

```php
function menuLeftMenuItems( $parameters, &$object, &$action, $hookmanager ) {

	    // remove "activities" item by url value
	    $key1 = array_search("/projet/activity/perweek.php?leftmenu=tasks", array_column($object, 'url'));
	    unset($object[$key1]);

	    $hookmanager->results = $object;

	    return 1; // 0 = add / 1 = replace

    }
```

### DIFF
--- a/htdocs/core/class/hookmanager.class.php
+++ b/htdocs/core/class/hookmanager.class.php
@@ -193,6 +193,7 @@ class HookManager
 				'getFormatedSupplierRef',
 				'getIdProfUrl',
 				'getInputIdProf',
+				'menu_LeftMenuItems',
 				'moveUploadedFile',
 				'moreHtmlStatus',
 				'pdf_build_address',
--- a/htdocs/core/class/hookmanager.class.php
+++ b/htdocs/core/class/hookmanager.class.php
@@ -193,7 +193,7 @@ class HookManager
 				'getFormatedSupplierRef',
 				'getIdProfUrl',
 				'getInputIdProf',
-				'menu_LeftMenuItems',
+				'menuLeftMenuItems',
 				'moveUploadedFile',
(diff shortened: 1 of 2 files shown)


eldy:
Because the hook name is also name of a method and we want method names to follow camelcase syntax, can you rename hook into menuLeftMenuItems (without _)

frederic34:
@cfoellmann perhaps you can add sample code in modulebuilder template

cfoellmann:
> @cfoellmann perhaps you can add sample code in modulebuilder template

good idea. How should do that? Commented out?

cfoellmann:
@frederic34 @eldy 
Can we add "all" hooks to the wiki?
Should we add one page with all hooks or a category with pages for each hook?

@eldy can you decide and create a draft? So I can start with my hook. Also will start documenting the hooks I have already used

eldy:
@cfoellmann You can start from page
https://wiki.dolibarr.org/index.php?title=Hooks_system#List_of_available_Contexts_in_Dolibarr
Where list of hooks are listed (not complete) and make each hookname a link to another page. On this page, you can show an example HelloWork on how to implement the hook.

cfoellmann:
@eldy I am a litte lost using mediawiki.
We need a category `Hooks` and collect all hooks there + links to detail pages for each hook. Can you start that? So I can add hooks after your example

cfoellmann:
@eldy 
I am not sure what I did here or what I am doing wrong now.
The unset does not work (anymore) 😢

PR #18305, opened 2021-08-02
Fix fatal error with php 8.0 holiday

Hystepik:
Fix fatal error with php 8.0 on holiday module due to a string in abs() function


htdocs/core/lib/date.lib.php
```diff
@@ -137,11 +137,11 @@ function dol_time_plus_duree($time, $duration_value, $duration_unit)
 	$deltastring = 'P';
 
 	if ($duration_value > 0) {
-		$deltastring .= abs($duration_value);
+		$deltastring .= abs((int) $duration_value);
```
eldy, line 140: duration_value and duration_unit should be already int according to method doc.
If not, it would be better to fix the caller. Here the fix may "hide" a problem into caller so i prefer to keep current code as it is to alert when a call is not done with correct parameters.

PR #18313, opened 2021-08-02
Fix variants issues: decimal separator not always parsed right + fatal on variant list

atm-florianm:
# Fix fatal on variant list + decimal sep. parsing
The fatal error occurs on `htdocs/variants/list.php` because class `Product` is redeclared (using `require_once` instead of `require` fixes it).

The other error occurs when you save a variant with a non-percentage variation (box unchecked) using a decimal separator other than ".": the value is not saved properly. Applying `price2num` from the moment the HTTP parameter value is assigned to its variable fixes the problem.


eldy:
I pushed a better fix. Thanks to try it

PR #18316, opened 2021-08-03
NEW : skip accept/refuse process for proposals

altairis-tof:
with this constant, on draft proposals, validate button will put the proposal to signed status and close button to unsigned status.


eldy:
To avoid confusion with other feature "Validate" means "No more draft", can you check that if the option PROPAL_SKIP_ACCEPT_REFUSE is on, the button "Validate" that set to signed has label "Validate and sign" and the button Close is renamed into "Set refused and close"  (so it is clear what are behind button compared to common process).

altairis-tof:
@eldy this one is ok now except travis complaining about regions table 
image

altairis-tof:
@eldy this PR is ready for merge.

PR #18321, opened 2021-08-03
NEW : add constant PROPAL_DONT_KEEP_SHIPPING_INFOS_ON_CLONING

altairis-tof:
do not copy availability id and planned shippiing date when cloning a proposal

### DIFF
--- a/htdocs/comm/propal/class/propal.class.php
+++ b/htdocs/comm/propal/class/propal.class.php
@@ -1378,6 +1378,10 @@ class Propal extends CommonObject
 			$object->note_private = '';
 			$object->note_public = '';
 		}
+		if ( ! empty($conf->global->PROPAL_DONT_KEEP_SHIPPING_INFOS_ON_CLONING)) {
+			$object->availability_id = '';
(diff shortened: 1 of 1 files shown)


eldy:
When we make a cloning, a confirmation popup is shown. On some other features, you can check option of what you clone on not in this popup. I would prefer to have this feature managed such this way instead of having a hidden constant for that.

altairis-tof:
@eldy i am not sure that people that clone 50 proposals or more per day will appreciate to have to think about checking or unchecking the options at each time.
and furthermore i do not think their boss will appreciate that they can check or uncheck options that he decided for the company workflow.

altairis-tof:
@eldy I agree with you; but are you ok if we add a hidden constant to define default behaviour ? 
@altairis-benjamin please follow this discussion linked with #18322 and #18329

eldy:
When we have an easy way to provide a feature that will do the job better than a hidden option AND this other way is not a huge work, we should not include the hidden option but introduce only the best way (to avoid the inflation of hidden options).

PR #18322, opened 2021-08-03
NEW : add constant PROPAL_DONT_KEEP_INTERNAL_CONTACTS_ON_CLONING

altairis-tof:
add new constant to not copy internal contacts when cloning a proposal

### DIFF
--- a/htdocs/comm/propal/class/propal.class.php
+++ b/htdocs/comm/propal/class/propal.class.php
@@ -1387,7 +1387,7 @@ class Propal extends CommonObject
 			$error++;
 		}
 
-		if (!$error) {
+		if ( ! $error && empty($conf->global->PROPAL_DONT_KEEP_INTERNAL_CONTACTS_ON_CLONING)) {
 			// copy internal contacts
 			if ($object->copy_linked_contact($this, 'internal') < 0) {
 				$error++;


eldy:
When we make a cloning, a confirmation popup is shown. On some other features, you can check option of what you clone on not in this popup. I would prefer to have this feature managed this way (with a checkbox "Clone also contacts") instead of having a hidden constant for that.

altairis-tof:
@eldy i am not sure that people that clone 50 proposals or more per day will appreciate to have to think about checking or unchecking the options at each time.
and furthermore i do not think their boss will appreciate that they can check or uncheck options that he decided for the company workflow.

eldy:
They won't have. the checkbox can be automatically set or not according to an option. But users keep choice and must not go into the admin setup if he want to change the behaviour for one particular case.

altairis-tof:
@altairis-benjamin thanks to look at this one and to check if it is not already coded in recent releases and develop branch. same here #18329 for orders and #18321 for shippings

eldy:
For gui consistency with other object and options on cloning, this must be implemented by suggested a checkbox in the confirm popup. A hidden option can exist to have the checkbox pre-selected or not.

PR #18327, opened 2021-08-03
NEW: add constant ORDER_ALLOW_DRAFT_CANCEL

altairis-tof:
allow to cancel a draft order


eldy:
By definition a draft is an object that is not yet finished to be edited/created. It can always be deleted because it does not exists yet.
What is the goal of allowing to cancel a draft. By definition the draft does not exists yet, why not just deleting it.

altairis-tof:
I will ask the company that made me do this long time ago

altairis-tof:
Hello @eldy ; here is the answer (in french) of my customer. To a better understanding, he use Tiaris's ecommerce module to get orders from his Prestashop ;-)

> A mes yeux, le bouton ‘annuler’ existant est très utile, notamment pour les commandes draft et il sert à laisser une trace des commandes (surtout les web) qui ont été créées et même payées, avant d’être annulées/remboursées ou le montant versé laissé en crédit dans le compte du client.
Si la commande est tout simplement supprimée ça laisse un trou dans le futur suivi du crédit restant ou montant remboursé, une procédure assez fréquente dans la gestion et la modification des commandes où les délais de livraison sont trop longs pour le client, le produit est finalement obsolète et/ou le client change d’avis et demande que nous remplacions des articles dans sa commande

eldy:
So if the order was already paid, it means the order has no more the status "Draft" but the status "Validated" (If the order has been set into an ecommerce platform, the order exists in real life, so the order must have status "validated" and even "ordered" into Dolibarr. 
The use case described here is to be able to set a validated order (that exists) to status "Cancel", not to set a draft order to cancel.
You can't have any hole into numerotation with a draft because a draft has no number. It does not exists yet, so no number was yet allocated. The ref is assigned only during validation step.
The answer provided by the customer demonstrates that he don't need the PR.

altairis-tof:
in this use case, eshop orders are sometimes converted into drafts in Dolibarr if they are missing informations or something else that need human validation.
The hole in numerotation is forbidden for invoices but not for orders.
No problem if you dont want this PR, it will remain as specific into my customer's dolibarr.

dpriskorn:
I'm closing this

altairis-tof:
Why closing this ?

altairis-tof:
@dpriskorn i mean : it's an hidden optional feature that can be useful in the case described by my customer, and surely for others dolibarr users, so why dont you want to integrate it in the core ? if you do so, every dolibarr instance needing this feature will need to have it as specific and to be checked at every dolibarr version change !

dpriskorn:
If I understood Eldy correct it does not make sense to cancel a draft in the current workflow of an order. You delete it instead.
I closed it because it did not seem clear how this could be of value.
I did not understand the french text, so I might have missed something 🙃. Would you mind adding a machine translation?

dpriskorn:
> eshop orders are sometimes converted into drafts in Dolibarr if they are missing informations or something else that need human validation.

↑ Which eshop are you talking about here?

altairis-tof:
Sometimes you have e-commerce websites linked with Dolibarr...

altairis-tof:
You just should not close PRs. It is @eldy and other Yodas job.

BebZ:
> > eshop orders are sometimes converted into drafts in Dolibarr if they are missing informations or something else that need human validation.
> 
> ↑ Which eshop are you talking about here?

For us we have Magento 2 synching draft orders and sometimes payment doesn't reach us, so instead of keeping a draft or validated and erase/cancel order (that makes statistics false), we could need it too

dpriskorn:
> You just should not close PRs. It is @eldy and other Yodas job.

Oh, sorry for that. I'll refrain from doing that going forward.

altairis-tof:
i will resolve conflicts next week i think

altairis-tof:
@eldy conflicts fixed but Travis also complaining 
image

BebZ:
I was reading again, 
and it makes sense to "cancel a draft" and not "erase it" 
for shops wanting to keep an eye and make stats 
on customers making an "order", and then they forget to make the payment.
So we "erase" it at the moment, but we have approx a thousand customers with an "empty" account, no draft, no cancelled order, which looks silly, because we could have kept the information in Dolibarr about their "interest" (the "cancelled" order)

eldy:
There is no reason to cancel a draft. A draft is something that does not exist yet.
So if you don't want it, just delete it. If you want to keep it but cancel it, it means you want to validate it (draft means "does not exist" so "want to keep" means validate = not being draft = keep it). And if you don't want to send the order, just cancel the validated order. Don't forget "validated" does not means "approved", it just means "draft finished, keep the object that is into a non existing state by moving it into a final state to keep it definitely".

A shortcut may exists to make it in one step "Validate + Cancel" but this will just create more confusion in feature to just save 1 click.

eldy:
See comment of June 12, 2022 for closing reason.

PR #18331, opened 2021-08-03
FIX: multicurrency: fields in discount unitialized when creating deposit

marc-dll:
Causes all kinds of mayhem in said deposit when converting to discount, etc.


FHenry (approved):

marc-dll:
@eldy : Can you please make a feedback as to why you would consider this PR should not be merged ?

PR #18337, opened 2021-08-04
Fix : we only want employee users for holidays

atm-maxime:
Without this, holiday were managed for all users, even external...

### DIFF
--- a/htdocs/holiday/class/holiday.class.php
+++ b/htdocs/holiday/class/holiday.class.php
@@ -1630,6 +1630,7 @@ class Holiday extends CommonObject
 					$sql .= " WHERE u.entity IN (".getEntity('user').")";
 				}
 				$sql .= " AND u.statut > 0";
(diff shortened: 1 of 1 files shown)


htdocs/holiday/class/holiday.class.php
```diff
@@ -1630,6 +1630,7 @@ public function fetchUsers($stringlist = true, $type = true, $filters = '')
 					$sql .= " WHERE u.entity IN (".getEntity('user').")";
 				}
 				$sql .= " AND u.statut > 0";
+				$sql .= " AND u.employee = 1"; // We only want employee users for holidays
```
eldy, line 1633: Why not the boss can't also create holiday to have its record validated and visible (even if I agree he validates himself)

atm-maxime: This fix is mostly about the holiday counters. The boss has no counter or if so, he must be considered as an employee. This doesn't prevent the boss to record it's holidays and have it shared on the company planning.

PR #18343, opened 2021-08-05
fix translation string for action generation on email sent

cfoellmann:
fixes the malformed title for the agenda item generated when sending an item via email


cfoellmann:
my change was incomplete that is right.
But your change does not makes it really good.

I gave it some thought again and my change would be better but we need to change the translation `MailSentBy` to:

```
MailSentBy = MailSentBy=Email sent by % to %
```
and the equivalent for the other languages. @eldy what do you think?

PR #18344, opened 2021-08-05
Fix lang error on hrm/index.php

Hystepik:
Fix error where typeleaves was not translated in hrm/index.php


cfoellmann:
Works!
@Hystepik please send to branch `14.0` since this is a fix for a current issue that should be fixed in all running versions

PR #18347, opened 2021-08-06
Fix: Change French word to English in English array.

erikvanberkum:
Status filter on /mrp/mo_list.php shows French word **Brouillon**, this is replaced with the English word **Draft** in the array.

Screenshot_2021-08-06 134428


cfoellmann:
Works!

cfoellmann:
@erikvanberkum @eldy 
The language works but the filter is not working for me.
Before and after this change. There seems to be some other issues here for the function itself

PR #18365, opened 2021-08-09
FIX #18361 Eat-By vs Sell-By: French mixup

### LINKED ISSUE #18361: Eat By vs Sell By Date mixup in French
# Bug
The French translations for `SellByDate` and `EatByDate` should be swapped. At first I thought the issue lay with the code (cf. [this closed PR](https://github.com/Dolibarr/dolibarr/pull/18311/files)) but I now think it is purely a translation problem.

Current translations:

| Key | en_US | fr_FR |
|----|-------|-------|
| SellByDate | Sell-by date | DLC |
| EatByDate | Eat-by date | DMD/DLUO |

Correct translations:

| Key | en_US | fr_FR |
|----|-------|-------|
| SellByDate | Sell-by date | DDM/DLUO |
| EatByDate | Eat-by date | DLC |

Note that, as far as I can tell, other languages (I checked de_DE and es_ES) use correct translations.

The legal meaning of DMD/DLUO matches the meaning of "Sell-by date" more closely than DLC does. Actually, I found [DDM](https://agriculture.gouv.fr/comment-conserver-les-aliments-dlc-ddm-quelle-difference), not "DMD".

While "Eat-by date" doesn't seem to have a precise legal meaning in US-American English, the idea it conveys matches that of "DLC" in French (the legal phrase used in US-American English is "[Use-by date](https://www.fda.gov/consumers/consumer-updates/confused-date-labels-packaged-foods)", but "Eat-by" is close enough).

For companies that use Dolibarr in a multilingual environment, this mistranslation could have serious consequences, but for companies that use only French, fixing this problem could also cause problems (after upgrading DLC would become DMD/DLUO and vice versa).

atm-florianm:
# FIX #18361 Eat-By vs Sell-By: French translation mixup

This PR includes the following:
+ renaming of all variable names, http query parameter names and array keys containing "dlc" or "dluo" with "eatby" and "sellby" to avoid confusion. Note: most of the time, the variable names were correct (they did not match the incorrect translations).
+ slight refactoring of 2 blocks
+ replacing some `$_POST` with `GETPOST`
+ swap fr_FR translations for eat-by and sell-by

# Warning
If accepted, this may require users who only use Dolibarr in French to also swap the values of database fields for `eatby` and `sellby` (because they will have entered the values based on the wrong translations).

However, if not accepted, the consequences could be more serious, especially for companies whose users use different languages (for instance, someone using Dolibarr in French enters a value as a "DLC" for a batch, and later an English-speaking user from the same company puts products from that batch in a delivery and that date becomes the "Sell By Date" → then a customer might buy that product at the date when it was deemed unfit for consumption).


cfoellmann:
I will test this asap

eldy:
Renaming var is a "re-engineering task". It must be done into develop branch only (it will cause too many trouble during merge if we do so in maintenance version. Only minimum code change must be done into maintenance version).
Can you push into dev or make a simple fix without re-engineering ?

atm-florianm:
I get your point; as soon as I find time to do it, I will split this PR to separate the variable renaming, which may be fit for the `develop` branch, and the actual fix, which should go into 13.0.

htdocs/fourn/commande/dispatch.php
```diff
@@ -306,8 +306,8 @@
 			$pu = 'pu_'.$reg[1].'_'.$reg[2];
 			$fk_commandefourndet = 'fk_commandefourndet_'.$reg[1].'_'.$reg[2];
 			$lot = 'lot_number_'.$reg[1].'_'.$reg[2];
-			$dDLUO = dol_mktime(12, 0, 0, $_POST['dluo_'.$reg[1].'_'.$reg[2].'month'], $_POST['dluo_'.$reg[1].'_'.$reg[2].'day'], $_POST['dluo_'.$reg[1].'_'.$reg[2].'year']);
-			$dDLC = dol_mktime(12, 0, 0, $_POST['dlc_'.$reg[1].'_'.$reg[2].'month'], $_POST['dlc_'.$reg[1].'_'.$reg[2].'day'], $_POST['dlc_'.$reg[1].'_'.$reg[2].'year']);
+			$dSellBy = dol_mktime(12, 0, 0, GETPOST('dluo_' . $reg[1] . '_' . $reg[2] . 'month', 'int'), GETPOST('dluo_' . $reg[1] . '_' . $reg[2] . 'day', 'int'), GETPOST('dluo_' . $reg[1] . '_' . $reg[2] . 'year', 'int'));
```
eldy, line 309: dlc is the sellby
dluo is the eatby

eldy:
I close PR because 2 other commits were done to fix this. See comments in #18361

Also values and translation seems good. Except in file reception.php that was bugged before fixes related into #18361

PR #18366, opened 2021-08-09
FIX compatibility postgre sql

atm-quentin:
# Fix compatibility postgre sql

When you filter on a select, an error occurs because two quotes are used, but for varchar search you have to use simple quote

### DIFF
--- a/htdocs/core/tpl/extrafields_list_search_sql.tpl.php
+++ b/htdocs/core/tpl/extrafields_list_search_sql.tpl.php
@@ -41,7 +41,7 @@ if (! empty($extrafieldsobjectkey) && ! empty($search_array_options) && is_array
 			if (in_array($typ, array('chkbxlst', 'checkbox'))) $mode_search=4;	                            // Search on a multiselect field with sql type = text
 			if (is_array($crit)) $crit = implode(' ', $crit); // natural_search() expects a string
 			elseif ($typ === 'select' and is_string($crit) and strpos($crit, ' ') === false) {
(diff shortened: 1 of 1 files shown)


htdocs/core/tpl/extrafields_list_search_sql.tpl.php
```diff
@@ -41,7 +41,7 @@
 			if (in_array($typ, array('chkbxlst', 'checkbox'))) $mode_search=4;	                            // Search on a multiselect field with sql type = text
 			if (is_array($crit)) $crit = implode(' ', $crit); // natural_search() expects a string
 			elseif ($typ === 'select' and is_string($crit) and strpos($crit, ' ') === false) {
-				$sql .= ' AND (' . $extrafieldsobjectprefix.$tmpkey . ' = \'' . $db->escape($crit) . '\')';
+				$sql .= " AND (" . $extrafieldsobjectprefix.$tmpkey . " = '" . $db->escape($crit) . "')";
```
l00ptr, line 44: Hello,

This modification breaks compat with PostgreSQL. We can't use double quote to compare strings, for more information : https://www.postgresql.org/docs/12/sql-syntax-lexical.html

Best regards,
L.

atm-maxime: This fix exactly what you say. Before it was using double quotes, after it uses simple quotes.

l00ptr: sorry for the mistake... I rode the commit log too fast.

PR #18369, opened 2021-08-09
[DRAFT] Add sub total and title

BB2A-Anthony:
[EN]
I would like to add in the corp the possibility of adding titles and subtotals in the documents. Do you think the best option is to add free line types or buttons at the bottom?

[FR]
J'aimerais ajouter dans le corp la possibilité d'ajouter des titres et sous-total dans les documents. Penssez vous que la meilleur option est de rejouter des types de ligne libre ou des bouttons dans le bas?

128700072-9c276e98-453d-4936-9a17-5f3d5a9ca6ab

### DIFF
--- a/htdocs/comm/propal/card.php
+++ b/htdocs/comm/propal/card.php
@@ -2509,6 +2509,22 @@ if ($action == 'create') {
 																									   // modified by hook
 		if (empty($reshook)) {
 			if ($action != 'editline') {
+				// Add title line in propal
+				if ($object->statut == Propal::STATUS_DRAFT&& $usercancreate) {
+					if ($usercanvalidate) {
+						print '<a class="butAction" href="'.$_SERVER["PHP_SELF"].'?id='.$object->id.'&amp;action=addtitleline">'.$langs->trans('AddTitleLine').'</a>';
+					} else {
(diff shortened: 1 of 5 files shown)


cfoellmann:
I am currently testing https://github.com/ATM-Consulting/dolibarr_module_subtotal but we use `rank` column and the lines between the products/services also get numbers and should be excluded from the `rank`

atm-maxime:
If you want to discuss a feature request, start by opening an issue before coding.
I'd say that a line type is a good choice but there's a lot of work to check that everywhere lines are manipulated we check for type 0 or 1 (to avoid adding a subtotal to the main total for example)

cfoellmann:
I am with @atm-maxime 
You should open an issue an discuss what you want to achieve. Your idea is not a simple task.
Look at the atm module it is not just a few lines of code

BB2A-Anthony:
At first it is not to go as far as the atm module

Issues #18380

erikvanberkum:
Just my personal opinion:
add free line types

eldy:
Also related to #18369 or #21675?

BB2A-Anthony:
@atm-maxime can you watch this #18380 ?

PR #18373, opened 2021-08-09
[NEW] PDF Document ref in title option

BB2A-Anthony:
# New [*PDF Document ref in title*]
image


eldy:
To avoid to have an inflation of option, i suggest to have this one being the default behaviour (no option at all, and having the new ref always after the title of document). This save one line and having an option to have it just under does not worth the case. We don't need compatibility for this. Can you make the new behaviour the default and remove all old code to have it on a dedicated line ?

BB2A-Anthony:
> To avoid to have an inflation of option, i suggest to have this one being the default behaviour (no option at all, and having the new ref always after the title of document). This save one line and having an option to have it just under does not worth the case. We don't need compatibility for this. Can you make the new behaviour the default and remove all old code to have it on a dedicated line ?

@eldy it's done

PR #18374, opened 2021-08-09
NEW Add option to hide customer code on PDF documents

BB2A-Anthony:
# New [*PDF not display customer code*]


atm-maxime:
Usually, const names start with "MAIN_".
And if you add an option like this, you should apply it to all standard PDF documents, not just azure.
And I think there's a function "yn ()" that displays a yes/no select

BB2A-Anthony:
> Usually, const names start with "MAIN_".
> And if you add an option like this, you should apply it to all standard PDF documents, not just azure.
> And I think there's a function "yn ()" that displays a yes/no select

Ok i have change const

You speek all standard PDF documents it's

- Project
- Propal
- Order
- Invoice
- Ticket
- Inter

Others?

I use yes / no as for the other options

htdocs/admin/pdf.php
```diff
@@ -371,6 +373,18 @@
 //}
 print '</td></tr>';
 
+//
+
+print '<tr class="oddeven"><td>'.$langs->trans("DOC_SHOW_CUSTOMER_CODE");
```
eldy, line 378: Can you invert the option (the default value must be as actual, not set and showing the customer ref on document). And setting must hide.
So can you rename option 'DOC_SHOW_CUSTOMER_CODE' into 'MAIN_DOC_HIDE_CUSTOMER_CODE'

bb2a: It's ok for me

BB2A-Anthony:
@eldy it's done

PR #18382, opened 2021-08-10
add global constants to suppress unneeded pdfs

cfoellmann:
'SUPPLIER_ORDER_DISABLE_PDF_AUTOUPDATE' and 'SUPPLIER_INVOICE_DISABLE_PDF_AUTOUPDATE' to suppress unneeded pdfs

PDFs for supplier_order and supplier_invoice are non-sense for us. It generates unneeded volumes of pdfs that use up space and backup cost.
Since we cannot use arrays in global constants for now this is a normal solution

### DIFF
--- a/htdocs/fourn/commande/card.php
+++ b/htdocs/fourn/commande/card.php
@@ -618,7 +618,7 @@ if (empty($reshook)) {
 			$ret = $object->fetch($object->id); // Reload to get new records
 
 			// Define output language
-			if (empty($conf->global->MAIN_DISABLE_PDF_AUTOUPDATE)) {
+			if (empty($conf->global->MAIN_DISABLE_PDF_AUTOUPDATE) && empty($conf->global->SUPPLIER_ORDER_DISABLE_PDF_AUTOUPDATE)) {
 				$outputlangs = $langs;
 				$newlang = '';
 				if ($conf->global->MAIN_MULTILANGS && empty($newlang) && GETPOST('lang_id', 'aZ09')) {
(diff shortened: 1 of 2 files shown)


BB2A-Anthony:
Why not add the param in modules configuration?

cfoellmann:
I could add an option for this if that is a big plus for some

cfoellmann:
@eldy how about introducing MAIN_DISABLE_PDF_AUTOUPDATE as the first constant that holds an array (serialized)?

```php
array('all') // migrate on update from `1`
array('supplier_order','supplier_invoice', )
```

cfoellmann:
or we use json_encode + json_decode which seems to be more performant

BB2A-Anthony:
> I could add an option for this if that is a big plus for some

This option can be interesting for others who do not have to think about their problems. Add an info giving the benefits

eldy:
If pdf are non sense for you, just do not enable pdf template and pdf will not be generated.
If a pdf is generated, it means you set a default pdf into setup of module supplier. By default there is no pdf set and so no pdf generated. Current setup can already manage this.

cfoellmann:
@eldy I understand what you mean and that would be great to not need another global constant BUT: I __cannot__ deactivate all pdf templates or uncheck the default option. ☹️

cfoellmann:
Some more testing:

1. deactivate all templates but 1
2. removed `COMMANDE_SUPPLIER_ADDON_PDF` from the DB table (cannot be unchecked on the admin page)

-> works for 'supplier_invoice'
-> does __not__ work for 'supplier_order'

eldy:
If it does not work for 'supplier_order', it means the feature for optional pdf generation is not yet supported for supplier order.
Because it is already supported for supplier_invoice, we must implement the feature for no pdf selection/generation for supplier_order the same way we did for supplier_invoice.

cfoellmann:
see https://github.com/Dolibarr/dolibarr/pull/18520
I am not sure how to get it to work correctly

PR #18384, opened 2021-08-10
FIX: shipping validation workflow: 'ORDER_CLOSE' trigger called from wrong object

marc-dll:


marc-dll:
Sorry, duplicate of #18385

PR #18388, opened 2021-08-10
fix : Undefined property: stdClass:: in C:\wamp64\www\dolibarr-dev\ht…

grandoc:
…docs\main.inc.php on line 2019


htdocs/main.inc.php
```diff
@@ -2016,24 +2016,24 @@ function top_menu_user($hideloginname = 0, $urllogout = '')
 	$dropdownBody .= '<div id="topmenulogincompanyinfo" >';
 
 	if ($langs->transcountry("ProfId1", $mysoc->country_code) != '-') {
-		$dropdownBody .= '<br><b>'.$langs->transcountry("ProfId1", $mysoc->country_code).'</b>: <span>'.showValueWithClipboardCPButton($conf->global->MAIN_INFO_SIREN).'</span>';
+		$dropdownBody .= '<br><b>'.$langs->transcountry("ProfId1", $mysoc->country_code).'</b>: <span>'.showValueWithClipboardCPButton(!empty($conf->global->MAIN_INFO_SIREN)).'</span>';
```
eldy, line 2019: Instead of showing the value, you show true or false !
You can fix this by replacing 
$conf->global->XXXX with getDolGlobalString($conf->global->XXXX)

PR #18407, opened 2021-08-12
FIX warning in supplier order list (status filter is multiselect)

atm-florianm:
# FIX
If you use the status filter in the supplier order list, you get a warning
> `preg_match()` expects parameter 2 to be string, array given in `/path/to/dolibarr/htdocs/core/lib/functions.lib.php` on line 559

This is because the status filter is a multiselect, which causes the post value to be an array. Since the `GETPOST` check parameter is `intcomma`, the check implies passing the value (the array) to a `preg_match()`.

This fix only avoids running `preg_match` on the array for the `intcomma` check. Maybe there could be a better fix, but it seems that other type checks do the very same.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -555,6 +555,7 @@ function GETPOST($paramname, $check = 'alphanohtml', $method = 0, $filter = null
 			if (!is_numeric($out)) { $out = ''; }
 			break;
 		case 'intcomma':
+			if (is_array($out)) break;
 			if (preg_match('/[^0-9,-]+/i', $out)) $out = '';
 			break;
 		case 'alpha':


htdocs/core/lib/functions.lib.php
```diff
@@ -555,6 +555,7 @@ function GETPOST($paramname, $check = 'alphanohtml', $method = 0, $filter = null
 			if (!is_numeric($out)) { $out = ''; }
 			break;
 		case 'intcomma':
+			if (is_array($out)) break;
```
eldy, line 558: If you come here with $out that is an array and $check that is 'intcomma', there is a bug into the caller.
You must fix the caller to call the GETPOST with param 'array'
On recent version, it must be param  'array:intcomma'

atm-florianm: While I agree that the caller should ideally not call GETPOST with `intcomma` as the check parameter on an array, I am merely being consistent with how the other check parameter values are treated (like `san_alpha`, `aZ`, `aZ09`, `aZ09comma`).

eldy: You're right that other case are also not the good way of working. This is why it was changed into v13 with check array:typetocheck
But hare we are in a case to fix and even if other case are not the expected target, when we fix we must follow as much as possible the expected target for 2 reasons: Better fix is better than fix that follow same error than other cases and also it will generated less regression when it will be merged with higher version that has already fixed the trouble globally.
Your PR is ok (i validate it), but my comment was a hope to have a fix more complete ;-)

atm-florianm: I agree. I thought this was the "easy" fix, but even if it means more work, it is better.

It will probably take some time to find all calls to `GETPOST` where the parameter can be an array and isn't checked as such, but we can do it little by little.

PR #18409, opened 2021-08-13
NEW : Allows product ref generation after the form is validated with PRODUCT_GENERATE_REF_AFTER_FORM

vincentjdc:
Adds global variable `PRODUCT_GENERATE_REF_AFTER_FORM` to allow product ref generation after the form validation

# New : Adds the possibility to generate a product reference after the form validation
We use a product ref generator that takes into account the value of a mandatory extra field of the product.
In that case, we dont want the ref input displayed in the creation form (nor the ref generated at that moment) and we want the reference to be generated after the form is validated and the object populated with the extra fields values.

So I use a global variable `PRODUCT_GENERATE_REF_AFTER_FORM` and if its value is not empty, the ref input field is hidden and the ref module is called just before the creation of the object in the database.


eldy:
Because this introduce a new feature/workflow, can you push this PR into develop branch (v13 is open for bug fix only) ?

eldy:
Can you solve conflict fo this change when pushed into develop branch ?

vincentjdc:
Conflict solved

PR #18410, opened 2021-08-13
FIX: e-mail of applicant is not transferred to the new user when creating a user from an applicant card

atm-florianm:
# FIX

cf. [forum thread](https://www.dolibarr.fr/forum/t/module-recrutement-pb-creation-compte-depuis-fiche-candidat/37535/4)

When you create a new Dolibarr user from an applicant card, the creation fails with the message "Email …@… already exists."
Transferring the e-mail field from the applicant object to the user object fixes it.


atm-maxime:
I'm not sure this is the best fix. The applicant email is its personal email. So it will never become its company email...

PR #18413, opened 2021-08-14
Fix: MO status selector draft showing all values instead of showing o…

erikvanberkum:
Currently, when status selector draft is selected in Manufacturing Orders, all values are listed (Draft, Validated, In Progress, Produced and Canceled). The pull request fixes this issue by only showing the Draft MO's.

This bug is also present in bom_list.php and workstation_list.php when this pull request is accepted, I will create a pull request for both bom_list and workstation_list. 

This issue was referenced @cfoellmann in https://github.com/Dolibarr/dolibarr/pull/18347#issue-705190299


cfoellmann:
maybe "we" should just jump to refactoring the status search fields to the proper multiselect like https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/fourn/commande/list.php#L1162 !?

cfoellmann:
+1 for this PR. It is the fix for the issue at hand

eldy:
I made a more complete fix, including bom_list.php and workstation_list.php

PR #18417, opened 2021-08-15
FIX #18389 Accountancy - Bug on LDcompta10 export for supplier invoice

### LINKED ISSUE #18389: LDCompta10 export issue - accountancyexport.class.php
# Bug
The LDCompta 10 export is bad.

Code Nature Pièce (CNPI) must be AC for customer asset (avoir) and FC for supplier_invoice. The old calculation was if montant <0 then AC, else FC. It was GOOD. Now, with a debit / credit substraction, it doesn't work.

## Environment
- **Version**: 13.0.0.4
- **OS**: Windows 2019
- **Web server**: DoliWamp
- **PHP**: N/A
- **Database**: N/A
- **URL(s)**: /accountancy/bookkeeping/list.php?action=export_file&&search_date_startmonth=1&search_date_startday=1&search_date_startyear=2021&search_date_endmonth=12&search_date_endday=31&search_date_endyear=2021

The LDCompta10 technical description explain she need FC line for supplier invoice  and AC for customer asset (avoir) but with the "new" algorithm, we have FC for "positive" line and AC for negative with the calculation debit-credit <0 then AC, else FC

export bookeeping for LDCOMPTA10

Must reuse the montant < 0 use before in the accountancyexport.class.php

aspangaro:
Even if the amount field is depreciated, we still need it to detect a credit easily. A simple calculation (debit - credit) is not enough to know the type of situation.

Example : 

Supplier invoice
image

Credit supplier invoice
image

Only information in amount field define the sens of the operation... Please, it is necessary to keep the amount field

### DIFF
--- a/htdocs/accountancy/class/accountancyexport.class.php
+++ b/htdocs/accountancy/class/accountancyexport.class.php
@@ -1462,13 +1462,13 @@ class AccountancyExport
 			print $date_lim_reglement.$separator;
 			// CNPI
 			if ($line->doc_type == 'supplier_invoice') {
-				if (($line->debit - $line->credit) < 0) {
+				if (($line->amount) < 0) {
 					$nature_piece = 'AF';
 				} else {
 					$nature_piece = 'FF';
 				}
 			} elseif ($line->doc_type == 'customer_invoice') {
-				if (($line->debit - $line->credit) < 0) {
+				if (($line->amount) < 0) {
 					$nature_piece = 'AC';
 				} else {
 					$nature_piece = 'FC';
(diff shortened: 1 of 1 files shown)


GuillaumeMARQUES:
Absolutely, we need one way to identify transaction way.... Amount is mandatory...

htdocs/accountancy/class/accountancyexport.class.php
```diff
@@ -1460,13 +1462,13 @@ public function exportLDCompta10($objectLines)
 			print $date_lim_reglement.$separator;
 			// CNPI
 			if ($line->doc_type == 'supplier_invoice') {
-				if (($line->debit - $line->credit) > 0) {
+				if (($line->amount) < 0) {
```
eldy, line 1465: line->amount is not always filled. It was the field used in past but was replaced with debit and credit and is now not always filled.
Can you try to find a fix that use $line->debit, $line->credit and $line->sens

PR #18424, opened 2021-08-16
Fix php warning

GurvanKervern:
Fix/PHP8 warnings

Removing PHP8 warnings by adding tests and/or values to a bunch of variables


cfoellmann:
please remove the comments about the php8 fixing. This makes not sense to anyone reading the code afterwards

GurvanKervern:
> please remove the comments about the php8 fixing. This makes not sense to anyone reading the code afterwards

Some of those files include 6-7 others, and are included in dozens more, so I was afraid of unforeseen effects and thought it would be nice to leave a comment for another programmer hunting a potential bug. I thought they would be easy enough to remove at a later date/in future commits.

I can remove them right now but shouldn't I wait to confirm this doesn't cause any bug downstream for anyone?

eldy:
You must remove code comment. The track to know the fix was done for fixing php8 (reason of fix) must be recorded in github PR label.

htdocs/societe/list.php
```diff
@@ -672,6 +672,8 @@
 }
 $sql .= $hookmanager->resPrint;
 // Add GroupBy from hooks
+//removing PHP8 warnings
```
eldy, line 675: Comment to remove

htdocs/supplier_proposal/list.php
```diff
@@ -202,6 +202,9 @@
 	setEventMessages($hookmanager->error, $hookmanager->errors, 'errors');
 }
 
+//to remove PHP8 warnings down the line
```
eldy, line 205: Comment to remove

htdocs/societe/list.php
```diff
@@ -672,6 +672,8 @@
...
+$all, = isset($all,) ? $all, : '';
```
eldy, line 676: Syntax error here. I suggest to fix by replacing the
$parameters = array('all' => $all, 'fieldstosearchall' => $fieldstosearchall);
into
$parameters = array('fieldstosearchall' => $fieldstosearchall);
just after.
The $all is never defined, so providing this parameter is useless.

htdocs/contrat/services_list.php
```diff
@@ -412,18 +412,28 @@
 if (!empty($filter_opcloture) && $filter_opcloture != -1) {
 	$param .= '&amp;filter_opcloture='.urlencode($filter_opcloture);
 }
+//removing PHP warnings
+$filter_dateouvertureprevue = isset($filter_dateouvertureprevue) ? $filter_dateouvertureprevue : '' ;
```
eldy, line 416: A better fix would be to make
$filter_dateouvertureprevue = '';
$filter_date1 = '';
$filter_date2 = '';
$filter_opcloture = '';

at line 293 (so just before the code that may change the value to something else).

htdocs/societe/list.php
GurvanKervern: Ok

htdocs/contrat/services_list.php
GurvanKervern: Wouldn't that overwrite the variables if they were defined in including or included files?

htdocs/comm/action/index.php
```diff
@@ -528,7 +528,7 @@
 if ($user->rights->agenda->myactions->create || $user->rights->agenda->allactions->create) {
 	$tmpforcreatebutton = dol_getdate(dol_now(), true);
 
-	$newparam .= '&month='.str_pad($month, 2, "0", STR_PAD_LEFT).'&year='.$tmpforcreatebutton['year'];
+	$newparam = 'month='.str_pad($month, 2, "0", STR_PAD_LEFT).'&year='.$tmpforcreatebutton['year'];
```
fappels, line 531: You need to keep &month because is put after &backtopage

GurvanKervern: not when the variable is undefined, but now both cases are covered :)

htdocs/contrat/services_list.php
```diff
@@ -467,14 +479,15 @@
 
 print_barre_liste($title, $page, $_SERVER["PHP_SELF"], $param, $sortfield, $sortorder, $massactionbutton, $num, $nbtotalofrecords, 'contract', 0, '', '', $limit);
 
-if ($sall) {
+if (isset($sall)) {
```
eldy, line 482: if ($sall) 
is not similar to 
if (isset($sall)) {
If $sall = '', the isset is true but the if ($all) is true only if value is something not null, not empty, not 0.

Use instead 
if (!empty($sall)) {

htdocs/comm/action/index.php
```diff
@@ -528,7 +528,11 @@
 if ($user->rights->agenda->myactions->create || $user->rights->agenda->allactions->create) {
 	$tmpforcreatebutton = dol_getdate(dol_now(), true);
 
-	$newparam .= '&month='.str_pad($month, 2, "0", STR_PAD_LEFT).'&year='.$tmpforcreatebutton['year'];
+	if (isset($newparam)) {
+		$newparam .= '&month='.str_pad($month, 2, "0", STR_PAD_LEFT).'&year='.$tmpforcreatebutton['year'];
```
eldy, line 532: An easier fix is to keep only one line of code and just add before the if the init of var
$newparam = '';

GurvanKervern: From what I understand we can't : @fappels mentioned $newparam sometimes contains something for him, so we can't set it, erasing whatever was in prior.

GurvanKervern: Also, wouldn't the amperstand at the beginning of the query string cause problems? (like index.php?**&**month...)

fappels: I was confused by meaning of 'newparam'. Now have found it's: 'Parameters on current URL' so parameter for backtopage and not after backtopage. 
Think best solution is to put 
$newparam = ''; //Parameters on current URL 
at beginning of script to avoid to have 'null' send to function show_day_events.
url like /index.php?&month... seems not to give any issue and shows correct month.

GurvanKervern: Ok

htdocs/contrat/services_list.php
eldy: You must know one thing: 
Into a main page, the only var that are (shoud be) defined into external include files are the main global var:
$mysoc (object that store your company), 
$conf (storing all the setup), 
$user (the current logged user), 
$hookmanager (a utility class), 
$lang (the language to use for output). 
So if var is not this one, it is missing an init.   Only *.tpl.php file can use other vars that are not initialized because they are template file included into main pages.

GurvanKervern: Cool, thanks :)

PR #18426, opened 2021-08-16
FIX Accountancy - SQL error when insert a manuel transaction

aspangaro:
image


htdocs/accountancy/class/bookkeeping.class.php
```diff
@@ -729,7 +729,10 @@ public function fetch($id, $ref = null, $mode = '')
 		$sql .= " t.journal_label,";
 		$sql .= " t.piece_num,";
 		$sql .= " t.date_creation,";
-		$sql .= " t.date_export,";
+		// In llx_accounting_bookkeeping_tmp, field date_export doesn't exist
+		if ($mode <> "_tmp") {
```
eldy, line 733: <> is strange. Do you mean  !=   ?

PR #18427, opened 2021-08-17
Enable getSumDepositsUsed for Supplier Deposit Invoice

henrynopo:
When FACTURE_DEPOSITS_ARE_JUST_PAYMENTS is enabled, getSumDepositsUsed shall be enabled for Supplier Deposit Invoice in order to fetch the correct amount of deposit amount applied as payment of Supplier Invoice.

### DIFF
--- a/htdocs/core/class/commoninvoice.class.php
+++ b/htdocs/core/class/commoninvoice.class.php
@@ -159,11 +159,6 @@ abstract class CommonInvoice extends CommonObject
 	 */
 	public function getSumDepositsUsed($multicurrency = 0)
 	{
-		if ($this->element == 'facture_fourn' || $this->element == 'invoice_supplier') {
-			// TODO
-			return 0.0;
-		}
-
 		require_once DOL_DOCUMENT_ROOT.'/core/class/discount.class.php';
 
 		$discountstatic = new DiscountAbsolute($this->db);


eldy:
Mode FACTURE_DEPOSITS_ARE_JUST_PAYMENTS must no more be used. It is an very old abandoned mode that does not respect the laws and rules on accounting. It has been abandoned and must not be used.

henrynopo:
> Mode FACTURE_DEPOSITS_ARE_JUST_PAYMENTS must no more be used. It is an very old abandoned mode that does not respect the laws and rules on accounting. It has been abandoned and must not be used.

What you meant is for EU, if I understand well. Should keep Mode FACTURE_DEPOSITS_ARE_JUST_PAYMENTS valid because this is allowable in different countries.

eldy:
For all country, this option has no more sense. All countries fight against false turnover. Using this feature means that each time you make a deposit, the amount invoiced is counted twice: in the deposit and later int the final invoice after (because the deposit is not deducted of the amount of final invoice). This is surely not legal in all countries in the world.

What is you use case so you need this ? Why not doing a real deposit if you need to manage deposit ?

henrynopo:
> For all country, this option has no more sense. All countries fight against false turnover. Using this feature means that each time you make a deposit, the amount invoiced is counted twice: in the deposit and later int the final invoice after (because the deposit is not deducted of the amount of final invoice). This is surely not legal in all countries in the world.
> 
> What is you use case so you need this ? Why not doing a real deposit if you need to manage deposit ?

Accounting-wise, there are different accounts for deposit and final invoice. The counting of deposits can be easily solved as below:
- deposit or prepayment paid to Supplier is an asset (I believe it can be classified under Asset / Deposit received). Whenever the order is cancelled, one has the right to claim back the amount.
- deposit or prepayment received from the customer is a liability (possible to classify under Liabilities / Deposit paid). Whenever the order is cancelled, one has the duty to pay back the amount.
- Deposit can be directly counted as sales/cost if it is for example none-refundable or very unlikely refundable.
- The sales / expense (cost) will be counted only if the products are delivered / received. 
- the rules of accounting for services (for example IT programming / consultancy) are different from physical goods (products).

Accordingly, based on the same rules, the final invoice amount will be counted into Sales / Cost and the deposits shall not be counted.

The double-counting of deposit amount with the final invoice amount is currently a serious bug of Dolibarr current versions. Therefore, every time I have to retrieve manually the database and do my sales/cost calculations by the rules above. Actually, my company prefers FACTURE_DEPOSITS_ARE_JUST_PAYMENTS, because taking deposit directly into turnover is actually false claim.

Lastly, it is understood that Dolibarr intends to be international / global. Then this function must be maintained and the bug shall be resolved (I believe not very complicated: just need to omit the deposit amounts from turnover.).

eldy:
Due to violation violation of laws and of accounting rules, the option FACTURE_DEPOSITS_ARE_JUST_PAYMENTS may be completely removed in a future (this option never worked correctly and has been deprecated a long time ago, it was kept for compatibility only).

You should describe what is your bug without this option if you think there is a bug (providing data and screenshots) because there is several thousand of bookkeepers that use Dolibarr for double entry accounting and none of them has declared any bug (that I saw) related to deposits (except the bookkeepers that are were using the option FACTURE_DEPOSITS_ARE_JUST_PAYMENTS to on of course). Please create a ticket with an example of the data you are using to describe your bug if you think you have one.

If final invoice is counted into sales and deposit not, it means it is not a "deposit invoice" but just a pre-payment (you can call it "deposit" instead of "pre-payment" but the "deposit invoice" in Dolibarr was designed to be a real "deposit invoice"). May be your need is something else.

henrynopo:
Hi @eldy . After a lot of searching, I understand a little bit more about your concern.

In fact, we are talking about different meaning or practice of "deposit".

In housing rental, for example, deposit is more "security deposit" or a pledge and is contractually not allowable to be used to offset the last rental payment. Deposit is also not taxable and is not recognised as income or cost. 

Sometimes, deposit may refer to "down payment", which is a milestone payment, therefore, taxable and recoginaised as revenue / cost. In this case, I agree with your principle that FACTURE_DEPOSITS_ARE_JUST_PAYMENTS is not applicable and instead, Standard Invoice should be applied. As you mentioned double-entry accounting, I activate it in my system but have not actually applied it yet. Are you concerning with that this accounting function has no differentiation of deposit invoice from standard invoice, which is why it results in false declaration of turnover etc.?

In our business, i.e. trading (operated internationally), deposit refers to "advance payment" or "prepayment", which are subject to contract. According to our accountant and my professional friends, these payments are not recognised as revenue / cost, so not taxable. As I explained  previously, our deposit (more correctly, advance payment) paid and received are accounted into assets and liabilities (the likely account number for UK Base Charte of Account: 165 & 275). These are not in revenue / cost accounts and are non taxable. Accordingly, for annual report, deposits are not reported in Profit & Loss Statement (basis for tax declaration) but in Balance Sheet.

The product delivery is arranged only after the reception of advance payment, and the Tax Invoice (standard invoice) is generated then. Therefore, for tax declaration or annual report, the tax invoice is accountable and deposit invoice is not. FACTURE_DEPOSITS_ARE_JUST_PAYMENTS  allows us to generate the Tax Invoice correctly. Without this function, the deposit payment is shown as part of the product list and its amount is deducted from the tax invoice amount, resulting in complicated customs clearance (international trading cases), not acceptable at all for some customs.

In addition, Tax Invoice is generated later than the advance payments and most of our tax invoices (either supplier or customer) have different amounts from the orders because the exact shipped quantities (net weight) are always slightly different from the order. It is therefore not practical to record the advance payments directly under any Standard Invoice copied from the orders, which needs modification later (Standard Invoices should be modified only in case of exception).

In a word, we are very satisfied with the function FACTURE_DEPOSITS_ARE_JUST_PAYMENTS, especially when Deposit Invoice is now open to supplier side. It fits perfectly with our trading context. Most importantly, it allows us to do accounting easily and correctly and it is the correct function to generate correct Tax Invoice for our customers. The current coding has no better invoice / payment alternative.

By understanding your concern, I think that its application is subject to business context. If it is a very serious concern, please prohibit it for EU companies, but please maintain it for international end-users like us. I guess that the tax issue may require correction(s) in the Accounting module.

My humble suggestion is that: the deposit invoice can be renamed to avoid confusion / wrong application, and it should be counted differently from standard invoice. In our business, our Deposit Invoice is actually just a Notice of Advance Payment. Without generating any deposit invoice (either for customer or supplier), we are unable to record correctly the respective payments. Unfortunately, the current coding has no other way to prepare and record such advance payment, convertible to payment of the formal Tax Invoice. As we have enabled FACTURE_DEPOSITS_ARE_JUST_PAYMENTS, the deposit invoice amount should also be 0 in the total amount listed under each order once it is converted into discount. Same for credit note.

image

After all, long texting may not be efficient enough and we can speak if necessary. Your re-consideration will be much appreciated.

PR #12875, opened 2020-01-17
FIX: new document line: default cost price selection can never be the product cost price if there is a PMP

ATM-Marc:

PR #18431, opened 2021-08-17
FIX only keep pmp price if set in margin setup

lvessiller-opendsi:
FIX only keep pmp price if set in margin setup
- pmp price was selected by default even if it was'nt set in margin setup

### DIFF
--- a/htdocs/core/tpl/objectline_create.tpl.php
+++ b/htdocs/core/tpl/objectline_create.tpl.php
@@ -698,7 +698,7 @@ if (!empty($usemargins) && $user->rights->margins->creer)
 						{
 							// If margin is calculated on PMP, we set it by defaut (but only if value is not 0)
 							console.log("id="+this.id+"-price="+this.price);
-							if ('pmp' == defaultbuyprice || 'costprice' == defaultbuyprice)
+							if ('pmp' == defaultbuyprice)
 							{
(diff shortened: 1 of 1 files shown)


eldy:
Don't understand the fix. What do you mean with "keep pmp price" ?

lvessiller-opendsi:
@eldy 
If "costprice" is set in margin setup, we have "pmp" price selected by default in prices select list when you add a product.
For example in order card, when you want to add a product and "pmp" and "cost" prices are not null, you have "pmp" price selected by default instead of having "cost" price.

eldy:
If "costprice" is set in margin setup, it means: "Autoselect the costprice if defined, else pmp else best vendor price else 0". To summarize, it measn "costprice first, but not only costprice". The goal is to have always something set the most accurate as possible so margin error is lower as possible.

Are you sure a costprice was set on the product ? if not, having the PMP set as a fallback is the expected behaviour.

lvessiller-opendsi:
@eldy 
Example : 
- you set "costprice" in margin setup
- you have a product with "costprice" and "pmp" defined
- you add a product line and "pmp" price was selected by default instead of "costprice"

lvessiller-opendsi:
@eldy 
You're right, it was fixed in "FIX #12875" and costprice is seleced by default even if pmp price is defined.

PR #18432, opened 2021-08-17
Adding translation of extrafield attribute if defined

henrynopo:
In case of multilingual production, it is necessary to allow translation of the extrafield, both the lable and attribute, especially for pdf generation. If the extrafield attribute is predefined in the translation, this update will allow it be shown in different languages in the pdf file.

### DIFF
--- a/htdocs/core/class/commondocgenerator.class.php
+++ b/htdocs/core/class/commondocgenerator.class.php
@@ -1341,7 +1341,7 @@ abstract class CommonDocGenerator
 
 				$field = new stdClass();
 				$field->rank = intval($extrafields->attributes[$object->table_element]['pos'][$key]);
-				$field->content = $this->getExtrafieldContent($object, $key);
+				$field->content = $outputlangs->transnoentities($this->getExtrafieldContent($object, $key));
 				$field->label = $outputlangs->transnoentities($label);
 				$field->type = $extrafields->attributes[$object->table_element]['type'][$key];
 


eldy:
Imagine the value of the extrafield is a code, for example  DAY, when showing on pdf, you will show "Jour" in french.
And if the value is a text, for example "this is my custom text", the translation will made nothing because the translation can only translate predefined keys. So in both cases, making a translation here will be a bad case.
Do you have an exemple where you need a translation on the value entered by the user ?

henrynopo:
> Imagine the value of the extrafield is a code, for example DAY, when showing on pdf, you will show "Jour" in french.
> And if the value is a text, for example "this is my custom text", the translation will made nothing because the translation can only translate predefined keys. So in both cases, making a translation here will be a bad case.
> Do you have an exemple where you need a translation on the value entered by the user ?

In sales order, I have added extrafield Select from List (e.g. 1,Seller; 2,Buyer) in English. However, the sales order shall be generated in another language, e.g. Chinese. As of now, it seems no way for me to achieve the objective, except to predefined the respective translation for the extrafield list, and tweak the extrafield content in the process of pdf generation.

eldy:
Right. We must implement a generic feature for this.  I suggest to change the following way:

1) Replace $field->content = $this->getExtrafieldContent($object, $key);
with $field->content = $this->getExtrafieldContent($object, $key, $outputlangs);
I have just commited this part.

2) Inside the getExtrafieldContent() method, make the outputlang->trans('XXX'), but only if text has the syntax `'__(XXX)__' `(this is the syntax in dolibarr to say we want the value XXX translated and that XXX is not a string to show as is but a translation key).
You can make a PR for this part. So in the definition of extrafields, users can now use` '1,__(Seller)__; 2,__(Buyer)__; 3,A fixed string that must not be translated'`

henrynopo:
Great. Thanks a lot.

PR #18435, opened 2021-08-18
Added China into the list of country using State / province in address

henrynopo:
Amended as subject.

In addition, I noticed that the coding seems to allow bi-lingual records of Towns: array_languages['town'][$extralangcode]. Where should this language-specific Town list be saved? in the corresponding language folder or in the database?

If the Town can be saved in bilingual, it would be perfect to allow State/Province also be shown in bi-lingual. Something similiar might be:
$state = ($extralangcode ? $object->array_languages['state'][$extralangcode] : (empty($object->state) ? '' : $object->state));

Same question is: where to save this language-specific list of states?

In addition, the function pdf_build_address does not specify any $extralangcode when calling the function in discussion here (dol_format_address). Is this $extralangcode useful or shall it be further replaced by $outputlangs?

An example of application for the above cases: in Dolibarr basic database, the list of states for Greece is all in Grecque language. How should I know its Latin state names both in card (for example Third-Party / Contact) and in pdf (for example order / invoice / shipment etc)?

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -2130,7 +2130,7 @@ function dol_format_address($object, $withcountry = 0, $sep = "\n", $outputlangs
 	global $conf, $langs;
 
 	$ret = '';
-	$countriesusingstate = array('AU', 'CA', 'US', 'IN', 'GB', 'ES', 'UK', 'TR'); // See also MAIN_FORCE_STATE_INTO_ADDRESS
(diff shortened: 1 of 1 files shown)


eldy:
Currently the name of a state can be store into one language only

henrynopo:
> Currently the name of a state can be store into one language only

How about the name of Town? can it be bilingual (English/Latin and local language)?

Working in bilingual business environment, I would suggest that town / state names in local language be saved in a dedicate database column (namely "local language name" or "local name") and the English/Latin name remains as the current "nom". Alternatively, the local names can be stored in the respective language folder with files namely state.lang and town.lang.

Is the current coding ready for any of the options above, or for a more complicated scenario?

PR #18445, opened 2021-08-19
Fix: TechnicalID displayed 2 times in both MO and BOM the PR hides 1.

### LINKED ISSUE #18444: TechnicalID displayed 2 times in both bom and mo card
When enabled constant, ```MAIN_SHOW_TECHNICAL_ID``` the technical ID is displayed double in both MO and BOM.
```
/bom/bom_card.php?id=
/mrp/mo_card.php?id=
```

dolibarr_pr_remove_technical_ID

erikvanberkum:
Fix: #18444 TechnicalID displayed 2 times in both MO and BOM the PR hides 1

picture explains it better than words:

dolibarr_pr_remove_technical_ID


htdocs/bom/class/bom.class.php
```diff
@@ -94,7 +94,7 @@ class BOM extends CommonObject
 	 * @var array  Array with all fields and their property. Do not use it as a static var. It may be modified by constructor.
 	 */
 	public $fields = array(
-		'rowid' => array('type'=>'integer', 'label'=>'TechnicalID', 'enabled'=>1, 'visible'=>-1, 'position'=>1, 'notnull'=>1, 'index'=>1, 'comment'=>"Id",),
+		'rowid' => array('type'=>'integer', 'label'=>'TechnicalID', 'enabled'=>1, 'visible'=>0, 'position'=>1, 'notnull'=>1, 'index'=>1, 'comment'=>"Id",),
```
eldy, line 97: In module builder, we use -2 for the value 'visible' for 'rowid'. Can you try with it instead of 0 ?
diff is that with -2, property should also be visible in list (with 0, it is never visible)

PR #18447, opened 2021-08-20
NEW: Add $object to 'printFieldListValue' hookmanager

milenmk:
NEW: Adding this option give the ability to pass commande options as object for hooks (like $object->id etc.). Otherwise it is not possible.


htdocs/commande/list.php
```diff
@@ -1779,7 +1779,8 @@
 		include DOL_DOCUMENT_ROOT.'/core/tpl/extrafields_list_print_fields.tpl.php';
 		// Fields from hook
 		$parameters = array('arrayfields'=>$arrayfields, 'obj'=>$obj, 'i'=>$i, 'totalarray'=>&$totalarray);
-		$reshook = $hookmanager->executeHooks('printFieldListValue', $parameters); // Note that $action and $object may have been modified by hook
+		$object = $generic_commande;
+		$reshook = $hookmanager->executeHooks('printFieldListValue', $parameters, $object); // Note that $action and $object may have been modified by hook
```
eldy, line 1783: Here $object is not a complete object. Providing it as $object parameter let think it is. All data of $object = $generic_commande should already be inside $obj, shouldn't it ?

PR #18448, opened 2021-08-20
improve forms - redirect after submission

cfoellmann:
redirect forms to appropriate target after submission + clear all POST to prevent resubmission errors

when a form (e.g. new product) is sent
1. redirect to product cardWITH `?id=`
2. cleared all $_POST to allow users to refresh the page without issue

### DIFF
--- a/htdocs/adherents/card.php
+++ b/htdocs/adherents/card.php
@@ -603,11 +603,14 @@ if (empty($reshook)) {
 					setEventMessages($object->error, $object->errors, 'errors');
 				}
 			}
-			$page = DOL_URL_ROOT.'/adherents/card.php?rowid='.$object->id;
-			header("Location: ".$page);
-			exit;
 		}
-		$action = ($result < 0 || !$error) ?  '' : 'create';
+
+		if ($result > 0 && !$error) {
+			header("Location: ".$_SERVER["PHP_SELF"].'?rowid='.$object->id);
+			exit;
+		} else {
+			$action = 'create';
+		}
 	}
 
 	if ($user->rights->adherent->supprimer && $action == 'confirm_delete' && $confirm == 'yes') {
--- a/htdocs/adherents/card.php
+++ b/htdocs/adherents/card.php
@@ -603,6 +603,9 @@ if (empty($reshook)) {
 					setEventMessages($object->error, $object->errors, 'errors');
 				}
 			}
+			$page = DOL_URL_ROOT.'/adherents/card.php?rowid='.$object->id;
+			header("Location: ".$page);
+			exit;
 		}
 		$action = ($result < 0 || !$error) ?  '' : 'create';
 	}
--- a/htdocs/projet/tasks/time.php
+++ b/htdocs/projet/tasks/time.php
@@ -208,9 +208,8 @@ if ($action == 'addtimespent' && $user->rights->projet->lire) {
 				}
 			}
 			$page = DOL_URL_ROOT . '/projet/tasks/time.php?withproject=1&projectid='.$projectid;
-			header( "Location: ".$page );
+			header("Location: ".$page);
 			exit;
-
 		}
 	} else {
 		if (empty($id)) {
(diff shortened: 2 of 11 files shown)


htdocs/adherents/card.php
```diff
@@ -603,6 +603,9 @@
 					setEventMessages($object->error, $object->errors, 'errors');
 				}
 			}
+			$page = DOL_URL_ROOT.'/adherents/card.php?rowid='.$object->id;
```
eldy, line 606: If there is an error in this action, we must not make a redirect, we must stay on the form to create. So we must force  $action = 'create' and not make the redirect

cfoellmann: let me look into that

htdocs/projet/tasks/time.php
```diff
@@ -207,6 +207,9 @@
 					$error++;
 				}
 			}
+			$page = DOL_URL_ROOT . '/projet/tasks/time.php?withproject=1&projectid='.$projectid;
```
eldy, line 210: Same: redirect must be done only if create was a success. Other case shoul dbe alredy managed by having the $action set to the action to redo (for example here $action = 'createtime'; line 186)

cfoellmann: let me review the PR once over before you comment on all steps

htdocs/adherents/card.php
cfoellmann: Would this be ok?

```php
if ($result < 0 || !$error) {
	$action =  '';
	$page = DOL_URL_ROOT.'/adherents/card.php?rowid='.$object->id;
	header("Location: ".$page);
	exit;
} else {
	$action = 'create';
}
```
at the end?

AND is this better?:

```php
header("Location: ".$_SERVER["PHP_SELF"].'?rowid='.$object->id);
```

eldy: If ($result < 0
it means there was an error too, isn't it ? In such a case, we should not make the redirect. redirect should be done on success only.

eldy: May be

eldy:
If you look into the code of modulebuilder, you will find the "generic" method to manage Cancel button.
It should work for all cases. I pushed the example with adherents/card.php
 This is the way:


If the page card.php has the 
	include DOL_DOCUMENT_ROOT.'/core/actions_addupdatedelete.inc.php'


Then you can add this (in a perfect world):

```
if (empty($reshook)) {

	$backurlforlist = DOL_URL_ROOT.'/societe/list.php';

	if (empty($backtopage) || ($cancel && empty($id))) {
		if (empty($backtopage) || ($cancel && strpos($backtopage, '__ID__'))) {
			if (empty($id) && (($action != 'add' && $action != 'create') || $cancel)) {
				$backtopage = $backurlforlist;
			} else {
				$backtopage = DOL_URL_ROOT.'/societe/card.php?id='.((!empty($id) && $id > 0) ? $id : '__ID__');
			}
		}
	}

...
```

If the page card.php has NOT the 
	include DOL_DOCUMENT_ROOT.'/core/actions_addupdatedelete.inc.php'

then it should be in a perfect world:

```

if (empty($reshook)) {

	$backurlforlist = DOL_URL_ROOT.'/societe/list.php';

	if (empty($backtopage) || ($cancel && empty($id))) {
		if (empty($backtopage) || ($cancel && strpos($backtopage, '__ID__'))) {
			if (empty($id) && (($action != 'add' && $action != 'create') || $cancel)) {
				$backtopage = $backurlforlist;
			} else {
				$backtopage = DOL_URL_ROOT.'/societe/card.php?id='.((!empty($id) && $id > 0) ? $id : '__ID__');
			}
		}
	}

	if ($cancel) {
		if (!empty($backtopageforcancel)) {
			header("Location: ".$backtopageforcancel);
			exit;
		} elseif (!empty($backtopage)) {
			header("Location: ".$backtopage);
			exit;
		}
		$action = '';
	}
```

eldy:
I have processed most cases. So i close this ticket. If you find more not managed you can ping me.

PR #18451, opened 2021-08-20
improve forms - buttonsSaveCreate()

### LINKED ISSUE #3756: Feature Request: class "button-save", "button-cancel" ... instead of
```
print '<input type="submit" class="button" value="' . $langs->trans("Save") . '">';
```

Proposed change would result in:

```
print '<input type="submit" class="button-save">';
```

and would allow for colourizing buttons in one place only (eg. include once  buttons.php).

cfoellmann:
the function buttonsSaveCreate() streamlines "all" Save/Cancel buttons on all forms.

This change is massive but well tested.
**BUT it removes all occurrences of `javascript:history.back(-1)` and requires other changes to the handling of `$cancel`!!**

@eldy + others: please give me feedback on the way I did this.
When the function is ok for you I would add the changes to handle `$cancel` and submission of theses forms properly.

Example: (societe/card.php)
```php
if (empty($reshook)) {
	if ($cancel) {
		if (!empty($backtopage)) {
			header("Location: ".$backtopage);
			exit;
		} elseif ($action == 'update') {
			if (!$socid) {
				$page = DOL_URL_ROOT.'/societe/index.php?leftmenu=thirdparties';
			} else {
				$page = DOL_URL_ROOT.'/societe/card.php?socid='.$socid;
			}
			header("Location: ".$page);
			exit;
		}
	}
}
```

Close #3756

### DIFF
--- a/htdocs/adherents/type.php
+++ b/htdocs/adherents/type.php
@@ -399,9 +399,8 @@ if ($action == 'create') {
 	print dol_get_fiche_end();
 
 	print '<div class="center">';
-	print '<input type="submit" name="button" class="button" value="'.$langs->trans("Add").'">';
-	print '&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;';
-	print '<input type="submit" name="cancel" class="button button-cancel" value="'.$langs->trans("Cancel").'" onclick="history.go(-1)" />';
+	print '<input type="submit" class="button button-save" name="save" value="'.$langs->trans('Add').'">';
+	print '<input type="submit" class="button button-cancel" name="cancel" value="'.$langs->trans("Cancel").'">';
 	print '</div>';
 
 	print "</form>\n";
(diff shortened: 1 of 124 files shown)


htdocs/adherents/type.php
```diff
@@ -399,9 +399,8 @@
 	print dol_get_fiche_end();
 
 	print '<div class="center">';
-	print '<input type="submit" name="button" class="button" value="'.$langs->trans("Add").'">';
-	print '&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;';
-	print '<input type="submit" name="cancel" class="button button-cancel" value="'.$langs->trans("Cancel").'" onclick="history.go(-1)" />';
+	print '<input type="submit" class="button button-save" name="save" value="'.$langs->trans('Add').'">';
```
eldy, line 402: If my understanding is good, we can replace this portion of code with

print $form->buttonsSaveCancel();

eldy:
It's ok.
Having a $form->buttonsSaveCancel is a good factorizing action.
Let's go for fixing the cancel action. My understanding is that only cancel whith no $backtopage provided is impacted. Lets' try as much as possible to provide a backtopage parameter when opening the page with the form to cancel.

PR #18462, opened 2021-08-22
fix : Undefined property: stdClass:: in C:\wamp64\www\dolibarr-140\ht…

grandoc:
…docs\comm\propal\card.php on line 125

and fixe : Attempt to read property "creer" on null in C:\wamp64\www\dolibarr-140\htdocs\comm\propal\card.php on line 125

### DIFF
--- a/htdocs/comm/propal/card.php
+++ b/htdocs/comm/propal/card.php
@@ -121,9 +121,7 @@ $usercansend = (empty($conf->global->MAIN_USE_ADVANCED_PERMS) || $user->rights->
 $usercancreateorder = $user->rights->commande->creer;
 $usercancreateinvoice = $user->rights->facture->creer;
 $usercancreatecontract = $user->rights->contrat->creer;
-if (!empty($conf->ficheinter->enabled)) {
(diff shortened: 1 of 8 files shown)


aspangaro:
Maybe in branch v14 ?

grandoc:
No it's a v15 alpha version
I am testing php 8 compliance

htdocs/comm/propal/card.php
```diff
@@ -121,7 +121,9 @@
 $usercancreateorder = $user->rights->commande->creer;
 $usercancreateinvoice = $user->rights->facture->creer;
 $usercancreatecontract = $user->rights->contrat->creer;
-$usercancreateintervention = $user->rights->ficheinter->creer;
+if (!empty($conf->ficheinter->enabled)) {
```
eldy, line 124: Using such condition at each permission test will lead to a lot of useless code. Also here the usercancreateintervention is not more initialized in some cases.
So i added a method user->hasRight(). You can now replace:

$usercancreateintervention = $user->rights->ficheinter->creer;
with 
$usercancreateintervention = $user->hasRight('ficheinter', 'creer');

PR #18468, opened 2021-08-23
NEW invoice fetch not found syslog debug level instead of error

lvessiller-opendsi:
In invoice "fetch" method when a record was not found we had an error in syslog.
This not an error but a debug information.
And a free "SQL" resource was added .
And a syslog with "SQL" error was added.


lvessiller-opendsi:
Replaced by a FIX in https://github.com/Dolibarr/dolibarr/pull/18475

PR #18474, opened 2021-08-24
Fix Shipment Bug for edit product lines and Add button for modification

henrynopo:
This bug exists since the first day I use the module Shipment from v13.0.0. (Please ignore the other files which should not be in this topic but I don't know how to move it out from here).

1. htdocs//expedition/card.php

In my case, no product batch or stock is pre-defined but the products are predefined (product ref exists). Therefore, whenever the shipment is created, editing the product line(s) and the respective extrafields always fail even though the shipment is in draft status. This fix (expedition/card.php) added the missing case for both editline and updateline.

2. htdocs/langs/en_US/sendings.lang

In addition, I have added a button for Shipment with status Validated. It will allow back to draft status for further editing.

3. Incorrect translation Key

The current translation for ValidateSending (confirm validation) seems incorrect and should be corrected. However, I failed to find its language file.

### DIFF
--- a/htdocs/commande/card.php
+++ b/htdocs/commande/card.php
@@ -800,6 +800,11 @@ if (empty($reshook)) {
 				if (!empty($price_ht) || $price_ht === '0') {
 					$pu_ht = price2num($price_ht, 'MU');
 					$pu_ttc = price2num($pu_ht * (1 + ($tmpvat / 100)), 'MU');
+					//added test to record multicurrency price while the default price is 0.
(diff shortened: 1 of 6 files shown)


atm-maxime:
Hi. Thanks for your fix !
We can't ignore some files, your pull request should contain only the fix. And since you say the issue is present in v 13, I suggest to start from the 13.0 branch, apply the fix and make a pull request from your branch to the core 13.0.

PR #18475, opened 2021-08-24
FIX invoice fetch not found syslog debug level instead of error

lvessiller-opendsi:
FIX invoice fetch not found syslog debug level instead of error

In invoice "fetch" method when a record was not found we had an error in syslog.
This not an error but a debug information.
And a free "SQL" resource was added .
And a syslog with "SQL" error was added.

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -1735,12 +1735,11 @@ class Facture extends CommonInvoice
 
 				return 1;
 			} else {
-				dol_syslog(__METHOD__ . ' Invoice with id=' . $rowid . ' or ref=' . $ref . ' or ref_ext=' . $ref_ext . ' not found', LOG_DEBUG);
+				dol_syslog(__METHOD__ . ' Invoice with id=' . $rowid . ' or ref=' . $ref . ' or ref_ext=' . $ref_ext . ' not found', LOG_WARNING);
 				return 0;
 			}
 		} else {
 			$this->error = $this->db->lasterror();
-			dol_syslog(__METHOD__ . ' Error : ' . $this->error, LOG_ERR);
 			return -1;
 		}
 	}
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -1625,10 +1625,10 @@ class Facture extends CommonInvoice
 		}
 
 		dol_syslog(get_class($this)."::fetch", LOG_DEBUG);
-		$result = $this->db->query($sql);
-		if ($result) {
-			if ($this->db->num_rows($result)) {
-				$obj = $this->db->fetch_object($result);
+		$resql = $this->db->query($sql);
+		if ($resql) {
+			if ($this->db->num_rows($resql)) {
+				$obj = $this->db->fetch_object($resql);
 
 				$this->id = $obj->rowid;
 				$this->entity = $obj->entity;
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/class/facture.class.php
```diff
@@ -1730,14 +1730,17 @@ public function fetch($rowid, $ref = '', $ref_ext = '', $notused = '', $fetch_si
...
 			} else {
-				$this->error = 'Invoice with id='.$rowid.' or ref='.$ref.' or ref_ext='.$ref_ext.' not found';
-				dol_syslog(get_class($this)."::fetch Error ".$this->error, LOG_ERR);
+				dol_syslog(__METHOD__ . ' Invoice with id=' . $rowid . ' or ref=' . $ref . ' or ref_ext=' . $ref_ext . ' not found', LOG_DEBUG);
 				return 0;
 			}
 		} else {
-			$this->error = $this->db->error();
+			$this->error = $this->db->lasterror();
+			dol_syslog(__METHOD__ . ' Error : ' . $this->error, LOG_ERR);
```
eldy, line 1743: There is no need to add a dol_syslog here. When a sql error occurs, there is already automaticaly a dol_syslog to output the error.

htdocs/compta/facture/class/facture.class.php
```diff
@@ -1730,14 +1730,17 @@ public function fetch($rowid, $ref = '', $ref_ext = '', $notused = '', $fetch_si
 					$this->error = $this->db->error();
 					return -3;
 				}
+
+				$this->db->free($resql);
+
 				return 1;
 			} else {
-				$this->error = 'Invoice with id='.$rowid.' or ref='.$ref.' or ref_ext='.$ref_ext.' not found';
```
eldy, line 1735: Why removing the set of ->error. It is required to return the reason of the fetch not found to the caller.

htdocs/compta/facture/class/facture.class.php
```diff
@@ -1730,14 +1730,17 @@ public function fetch($rowid, $ref = '', $ref_ext = '', $notused = '', $fetch_si
...
-				dol_syslog(get_class($this)."::fetch Error ".$this->error, LOG_ERR);
```
eldy, line 1736: May be you can just replace with LOG_WARNING or LOG_DEBUG (In most cases, if we do a fetch, we are using an id and not founding the record is not the expected behaviour)

lvessiller-opendsi: Ok I replaced with LOG_WARNING in syslog when no record was found

htdocs/compta/facture/class/facture.class.php
lvessiller-opendsi: @eldy
I don't understand why to keep an error in this class if it's not considered like an error (returning 0).
I think it's in the view (card) you show or not a message with a forbidden access.

htdocs/compta/facture/class/facture.class.php
lvessiller-opendsi: Thanks for your remark, I removed this line.

htdocs/compta/facture/class/facture.class.php
eldy: Only > 0 is a success
If you have an id and want to fetch it, and you do not found it. This is not the expected behaviour. It is not a technical "error" (< 0) but error is not just technical error, it is not expected result.

PR #18491, opened 2021-08-25
Fix : conditiion always wrong when on following link

daraelmin:
when members follow link, the oldtype is always equals to newtype, so this condition is alway wrong


daraelmin (commented):
@eldy to test this you should ad amount in adherent type

PR #18495, opened 2021-08-25
add lines title column position "Title" if activated

cfoellmann:
Add the column title for "Position" to the lines table header


eldy:
This column is of tne only 1 o 2 chars.
Using a label "Position" on title may increase the column size to show in information that is not really interesting (user already see th 1, 2, 3, ... to understand it is a number and we need to save space for a good readability of highly loaded screens.

PR #18496, opened 2021-08-25
Contact on different thirdparties with parent

leninrivas:


htdocs/core/class/html.form.class.php
```diff
@@ -1666,6 +1666,15 @@ public function selectcontacts($socid, $selected = '', $htmlname = 'contactid',
 		if ($socid > 0 || $socid == -1) {
 			$sql .= " AND sp.fk_soc = ".((int) $socid);
 		}
+		// Show contact of parent
+    	if (!empty($conf->global->CONTACT_ON_DIFFERENT_THIRDPARTIES)) {
+    		require_once DOL_DOCUMENT_ROOT.'/societe/class/societe.class.php';
+			$soc = new Societe($db);
+			$result = $soc->fetch($socid);
+			if ($result > 0 && !empty($soc->parent)) {
+				$sql .= ' OR sp.fk_soc = '.$soc->parent;
```
eldy, line 1675: Surely a bug here. You add a OR be before you may have a AND and after another AND.
You must introduce parenthesis on test for fk_soc to avoid unexepected effects.
Also can you confirm that option CONTACT_ON_DIFFERENT_THIRDPARTIES means in fact
CONTACT_SHOW_ALSO_PARENT_COMPANY_CONTACTS (i mean it show also contact of parent company but no other thirdpary ?).
If yes can you use this name CONTACT_SHOW_ALSO_PARENT_COMPANY_CONTACTS

leninrivas: Hi. Is right. Fixed

leninrivas (commented):
Recomendations applied

htdocs/core/class/html.form.class.php
```diff
@@ -1664,8 +1664,27 @@ public function selectcontacts($socid, $selected = '', $htmlname = 'contactid',
...
+		// Show contacts of thirdparties with same parent
+    	if (!empty($conf->global->CONTACT_SHOW_ALSO_COMPANY_CONTACTS_WITH_SAME_PARENT)) {
+    		require_once DOL_DOCUMENT_ROOT.'/societe/class/societe.class.php';
+			$soc = new Societe($db);
+			$result = $soc->fetch($socid);
+			if ($result > 0 && !empty($soc->parent)) {
+				$sql .= ' OR sp.fk_soc IN (SELECT rowid FROM '.MAIN_DB_PREFIX.'societe WHERE parent = '.$soc->parent.')';
+			}
+    	}
+		$sql .= ')';
```
eldy, line 1687: A parenthesis is closed in any case, but it is open only in one case. This is surely a bug.
Also can you check your IDE to set tabs to be tabs instead of spaces ? This generated trouble on merging and code analysis.

leninrivas: I was testing an IDE

PR #18497, opened 2021-08-25
NEW: date filter on project list and task list From->To

FHenry:
NEW: date filter on project list and task list From->To


aspangaro:
#18492 for task list was integrated

PR #18498, opened 2021-08-25
Undef vars2

GurvanKervern:
Fix undefined variables, indexes and properties

To remove PHP8 warnings


htdocs/core/menus/standard/eldy.lib.php
```diff
@@ -85,9 +85,9 @@ function print_eldy_menu($db, $atarget, $type_user, &$tabMenu, &$menu, $noout =
 		'position' => 10,
 		'id' => $id,
 		'idsel' => 'home',
-		'classname' =>  $classname = ($_SESSION["mainmenu"] && $_SESSION["mainmenu"] == "home") ? 'class="tmenusel"' : 'class="tmenu"',
+		'classname' =>  $classname = (!isset($_SESSION["mainmenu"]) && $_SESSION["mainmenu"] == "home") ? 'class="tmenusel"' : 'class="tmenu"',
```
eldy, line 88: I think test here should be 
!empty($_SESSION["mainmenu"])
instead of 
!isset

htdocs/comm/action/peruser.php
```diff
@@ -489,6 +490,9 @@
 	$newcardbutton .= dolGetButtonTitle($langs->trans("AddAction"), '', 'fa fa-plus-circle', DOL_URL_ROOT.'/comm/action/card.php?action=create&datep='.sprintf("%04d%02d%02d", $tmpforcreatebutton['year'], $tmpforcreatebutton['mon'], $tmpforcreatebutton['mday']).$hourminsec.'&backtopage='.urlencode($_SERVER["PHP_SELF"].($newparam ? '?'.$newparam : '')));
 }
 
+if (empty($num)) {
```
eldy, line 493: Juste set $num = '';
We are in a global page so if you can't find the var defined before, it means it is never set so we can just init it here.

htdocs/core/menus/standard/eldy.lib.php
GurvanKervern: Done

htdocs/comm/action/peruser.php
GurvanKervern: Ok, modified it too

htdocs/core/menus/standard/eldy.lib.php
eldy: Same on line 88

GurvanKervern: Oops, my bad, I changed all occurences

PR #18501, opened 2021-08-25
Fix Sales Order - add product line - always fail to input multicurrency price if product price is by default 0

henrynopo:
For my case, the product selling price is not pre-defined so the system set it at 0 by default.

When multicurrency is enabled, the system does not allow me the input the price in currency when a predefined product is added, resulting in price at 0. The price can be updated later when the product line is edit mode.


eldy:
Too many files (> 2000) modified. PR bugged so closed.

PR #18503, opened 2021-08-25
NEW : Sub bill of materials

atm-adrien:
# NEW : Sub bill of materials 

Addition of sub-BOM : 
- Addition of a BOM select on "add line" page that show all the existant BOM for the product selected (in the product select)
- Addition of a column for BOM in the list 
- Addition of +/- to show/hide the composition of sub-BOM
- Addition of options that permit to show/hide all lines that can be it
- Multiply sub-BOM lines quantity/costs by their parents products quantity
- Set total cost for products (that have a sub-BOM) by sum of sub-BOM lines costs

image

image


eldy:
Having 2 different columns for BOM and product makes tihng not clear (i spent to much time to understand how this will be used).  For the feature of cascading BOM (Cascading BOM is a feature to add a BOM into another BOM), I suggest to makes this clearer like this:

Merge the 2 colums into one to show for example on the same column "Description" of this BOM A, this:
Line 1  "Product P1"
Line 2  "BOM B or Product P2"    (what is important here is the "or")
It is clear that generating a MO later for BOM A will lead to 2 different choices:
If we click the checkbox "Generate Cascading MO": it will generated a MO B (to make product P2) and then a MO A (to make PF from P1 + P2, MO "A" will have status pending until MO "B" is done).
If we click the checkbox "Generate only parent MO": it will generated a MO with Product P1 and P2

And into the add form, we can also have
"The combo for product"
or 
"The combo for BOM"
instead of selecting a BOM after having selected a product.

Can you change the presentation this way ?


PS: Also in a future version, if a BOM contains the parent product of a variant, it means "ask during MO generation which variant of product to include into MO"

atm-adrien:
I admit that I do not understand how the merger of the two columns makes it clearer. Indeed, this could affect the readability and therefore the clarity of the table.
In addition, the table seems rather clear to me in the current state, indeed:
During the selection, it is not possible to select a BOM when none is associated with the selected product.
Once added, the rows appear in the table as you can see. The BOM column corresponds to the BOM associated with the product of the row. Why do you want to delete this column, which is only present to signify that the product is associated with a BOM and that you can see the composition of this BOM?

Also, several things that I have trouble understanding:

- For the addition form, we will therefore have two "selects", one for the selection of the product to add, and one for the selection of the associated BOM. Should it be possible to select only a BOM (without selecting a product)?
- For the display of rows in the table, when a BOM has been associated with a product (via the "select" of BOM), in what form should this be displayed: "Product or BOM" or " BOM "

eldy:
> I admit that I do not understand how the merger of the two columns makes it clearer. Indeed, this could affect the readability and therefore the clarity of the table.
> In addition, the table seems rather clear to me in the current state, indeed:
> During the selection, it is not possible to select a BOM when none is associated with the selected product.
> Once added, the rows appear in the table as you can see. The BOM column corresponds to the BOM associated with the product of the row. Why do you want to delete this column, which is only present to signify that the product is associated with a BOM and that you can see the composition of this BOM?
> 
> Also, several things that I have trouble understanding:
> 
> * For the addition form, we will therefore have two "selects", one for the selection of the product to add, and one for the selection of the associated BOM. Should it be possible to select only a BOM (without selecting a product)?

We should be able to select:
- A BOM only (we store the id of BOM into the line, no product)
- A product only (we store the product into the line, no BOM id)
It means we present the combo to select a BOM and under the combo to select a product. But user will select only into one.

When we will generate the MO, if this is a product, we include the product in list to consume. If this is a BOM (cascading BOM), we will let the user to include products of this BOM or to generate another a sub MO.

> * For the display of rows in the table, when a BOM has been associated with a product (via the "select" of BOM), in what form should this be displayed: "Product or BOM" or " BOM "
- You should show  "Product PREFXXXX or BOM BOREFZZZZ"  (the text "or" must be visible).

atm-thibaultf:
Hi @eldy,
Just to be sure that we understand well what you are saying, we design a mockup for cascading BOM here below :
Cascading BOM

For now we manage only one level to expand the first child BOM.

Also, here below you can find a mockup for the form to create a MO from this kind of BOM. Each line of the BOM which contains another BOM should propose a checkbox to decide if we want to generate a child MO. If we don't check this checkbox we have the actual behavior. Also we would like to add the stock (real and virtual) to this table like shown below :
Cascading MO

@eldy could you tell us if this is ok for you ?

atm-adrien:
@eldy

atm-thibaultf:
Hello @eldy,

Thank you for your feedback. In fact we need to be sure that the last mockup I have done is ok for you. The code of the actual PR doesn't include the last interface yet. If you could tell us if the screenshots are ok for you, we can change the PR to fit this interface.

Thank you in advance for your answer.
Thibault

eldy:
Yes, code seems good but you have to fix:
* interface.php should be inside a subdir "ajax" and not "script" and should have same first lines that existing files into ajax subdirs (also name could be change to represent the goal of file).
* CI must be ok.

atm-maxime:
:clap: :clap: :clap:

edupulpillo:
Great Job!!!!
One suggestion: What about integrate a tree as is in Tags/Categories to surf in the multilevel BOM list? (Also in projects-multilevel tasks will be useful)
image

eldy:
I made some fix on last submission to make the code stable again.
However, i added a TODO into the code to complete and fix it.
The component to add must be stored into a file 
html.formmrp.class.php
and the method that return the select may be called   ->selectBom()    
This match the way other modules are built. Then if you decide to use a simple select or an ajax call to return the list of BOM, this choice must be implemented inside the selectBom() like you have for selectProduct() or select_company()

Issue #13863, opened 2020-05-10
TakePOS sugestion


DATUC:
Hi Team just a suggestion about the new interface on V12 concerning the finder

![POS](https://user-images.githubusercontent.com/26550824/81503112-bf1b0900-92d9-11ea-9dad-3ab430fd7ee5.png)

DATUC:
Came we have a clean interface like
![POS](https://user-images.githubusercontent.com/26550824/81503368-25ecf200-92db-11ea-9a82-215cf5358075.png)

andreubisquerra:
The finder is located on the top in Dolibarr 12. Similar to your picture. I hope you like it.
![takepos12big](https://user-images.githubusercontent.com/28452516/82345960-2a5d8d00-99f6-11ea-86fe-e698f88a1b0f.JPG)

DATUC:
Hi @andreubisquerra i have done the taste on V12 and i wanted to figure out the fact that the actual position of the finder is not good comparing that to mamy other POS solutions and also the user appreciation, the store where i use to do the demo on testing enviroment and all user ask if ther is a possibilitie to change the finder position.
Excuse my poor English

DATUC:
Hi the request is still waited it concer the finder barr position on the interface

altairis-tof:
a solution would be to have movable blocks by drag&drop...

altairis-tof:
we have also made a PR to hide categories since they are not always usable (no category or  two many categories) : https://github.com/Dolibarr/dolibarr/pull/19527
another one to show category description  : https://github.com/Dolibarr/dolibarr/pull/19525
and another to show SKUs (product references) : https://github.com/Dolibarr/dolibarr/pull/19521
to come : category search PR and sell only product with stock PR
(PRs by @altairis-melina)

andreubisquerra:
I am working on an improved search engine which also autocompletes with a list of products. I will be ready shortly.

altairis-tof:
> I am working on an improved search engine which also autocompletes with a list of products. I will be ready shortly.

thanks for the information @andreubisquerra ;-) 

What about Melina Pull requests then ? it always take lot of time to push on develop branch so i hope it was not useless.

andreubisquerra:
Melina Pull requests is aproved for me!
Thank you @altairis-tof & @altairis-melina

eldy:
@andreubisquerra Can you mention ID of ticket you validated for Melina Pull requests (this may save me time) ;-)

[...]

PR #18505, opened 2021-08-25
FIX: printOriginLinesList: backport PR #13863 to v12

marc-dll:
I worked on nearly the same change proposed in PR #13863 before seeing. However, I would like to propose to backport it in v12.

The behaviour changed had been modified way back (6 years ago, commit https://github.com/Dolibarr/dolibarr/commit/990fbf3ef45a06577264f7ce6630964acdc9269c for version 3.9) for the equivalent `printObjectLine` hook, which is great as it enabled us to completely change the structure of the documents.

Not being able to do the same in the origin line list seems incoherent to me, it should have been made the same way from the beginning, that is why i'm proposing in this PR for version 12.0 instead of develop.

More over, I did some more modifications than in the PR#13863

The changes in detail :
- Initialize `$hookmanager` if it is not the case (should not happen anyway)
- expose hook `printOriginObjectLineTitle` to modify the origin lines table header

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -4459,7 +4459,6 @@ abstract class CommonObject
 		$i = 0;
 
 		if (!empty($this->lines)) {
-
 			foreach ($this->lines as $line) {
 				$reshook = null;
 
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -4429,37 +4429,49 @@ abstract class CommonObject
 	{
 		global $langs, $hookmanager, $conf, $form;
 
-		print '<tr class="liste_titre">';
-		print '<td>'.$langs->trans('Ref').'</td>';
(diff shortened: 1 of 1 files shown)


marc-dll:
I would like to add that this doesn't break compatibility for existing hook calls : they only have an effect on a restricted set of lines today and the will ignore the new parameters passed.

eldy:
I prefer not validate backport on old stable version (except for security or major bugs). it will difficult to explain community why some PR are validated and other not. If i do it for one, i must do it for all suggested back-port and this lead to non understandable release policy.
I recommand to main your own branch based on v12 if you need this.

fappels:
@marc-dll 
I made workaround like this:
```
// workaround to have hook 'printOriginObjectLine'
if (!empty($objectsrc->lines)) {
	foreach ($objectsrc->lines as &$line) {
		$line->product_type = 9;
		$line->special_code = 1;
	}
}

$objectsrc->printOriginLinesList('', $selectedLines);
```

PR #18507, opened 2021-08-26
Customer Invoice - Multicurrency Remaining to Pay (updated)

henrynopo:
Updated the new function to show remainder to pay in multicurrency.


htdocs/compta/facture/card.php
```diff
@@ -5122,6 +5128,21 @@ function js_recalculate_revenuestamp(){
 		}
 		print ' :</td>';
 		print '<td class="right'.($resteapayeraffiche ? ' amountremaintopayback' : (' '.$cssforamountpaymentcomplete)).'">'.price($sign * $resteapayeraffiche).'</td>';
+
+		// Remainder to pay back Multicurrency
+		if ($object->multicurrency_code != $conf->currency || $object->multicurrency_tx != 1) {
+			print '<tr><td colspan="'.$nbcols.'" class="right">';
```
eldy, line 5134: 2 lines before we have a closing </td>, so we should not have an open tr here, or we must add a closing tr before opening a new one.

PR #18509, opened 2021-08-26
NEW IHM - Include new options to manage dashboard

aspangaro:
Add a tab to manage dashboard in ihm

If option "Disable totally dashboard" is activated, option disable block of the dashboard one by one is not show

image


eldy:
Just to be sure. Are you speaking about the widget of the open/living objects (like in this screenshot) ?
image

aspangaro:
Hello @eldy, yes it's to manage this table like the screenshot. 

Maybe you want to rename in workboard than dashboard ?

eldy:
What do you think of 
"Thumbs of Open Objects",    "Vignette des objets ouverts" (FR)    for the title of the tab ? The goal of all this thumbs is to represent "living/open document" (so object that need action in present or in future).

altairis-tof:
hello i think this PR is great !
I did not look at the code but will there be any hook to replace this dashboard with anything else ?

PR #18511, opened 2021-08-26
Add button to modify the validated shipment (back to draft)

henrynopo:
Suggested as above.

The current translation for ValidateSending (confirm validation) seems incorrect and should be corrected. However, I failed to find it in the en_US language file.


eldy:
The key ValidateSending is inside file stocks.lang
For the moment i suspend any PR on workflow of shipment because there is a lot of PR not compatible each other that suggest changes like this one and they are using different philosophy. I must exchange will all core dev to decide with change to keep (during next devcamp this winter).

PR #18520, opened 2021-08-27
allow supplier order to not generate pdfs

cfoellmann:
If you __deselect__ the "standard" for pdf templates on supplier invoice it gets not generated automatically.
This does not work for supplier orders.

If anyone can see why please reply.

- [x] add ability to deselect standard on templates for supplier order

### DIFF
--- a/htdocs/admin/supplier_order.php
+++ b/htdocs/admin/supplier_order.php
@@ -132,6 +132,8 @@ if ($action == 'updateMask') {
 	if ($ret > 0) {
 		$ret = addDocumentModel($value, $type, $label, $scandir);
 	}
+} elseif ($action == 'unsetdoc') {
+	dolibarr_del_const($db, "COMMANDE_SUPPLIER_ADDON_PDF", $conf->entity);
 } elseif ($action == 'setmod') {
 	// TODO Verifier si module numerotation choisi peut etre active
 	// par appel methode canBeActivated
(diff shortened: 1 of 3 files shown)


henrynopo:
I would prefer not to automatically generate pdf for supplier order, because, for our business, suppliers will issue sales orders directly and we need not issue any pdf PO.

eldy:
@henrynopo This is why templates may not be selected "as default template". So no document will be generated by default. Same behavior than for supplier invoice (currently working on invoice side)

henrynopo:
> @henrynopo This is why templates may not be selected "as default template". So no document will be generated by default. Same behavior than for supplier invoice (currently working)

@eldy The existing module Vendor / Purchase Order auto selects at least one document model and does not allow deactivating all. I retried in V14.0.0 and am unable to deselect the last model.

cfoellmann:
> > @henrynopo This is why templates may not be selected "as default template". So no document will be generated by default. Same behavior than for supplier invoice (currently working)
> 
> 
> 
> @eldy The existing module Vendor / Purchase Order auto selects at least one document model and does not allow deactivating all. I retried in V14.0.0 and am unable to deselect the last model.

This is why I started this fix.
It allows to deselect default but it still autogenerates. 
Happy for any ideas @eldy

cfoellmann:
I cannot see why supplier orders PDFs are still being generated with the default option deselected

henrynopo:
> I cannot see why supplier orders PDFs are still being generated with the default option deselect

I have noticed the same issue.

cfoellmann:
only deselect for now. not generating should work but does not

PR #18530, opened 2021-08-27
add hidden option to display date and duration in invoice generation from task time spent

FHenry:
Add hidden option to display date and duration in invoice generation from task time spent


eldy:
A checkbox into the confirm popup to select the mode to generate invoice instead of using the $conf->global->PROJECT_TIME_SPENT_INTO_INVOICE_ADD_TIME_DT would be welcome.

PR #18534, opened 2021-08-28
Revert "[NEW] PDF Document ref in title option"

eldy:
Reverts Dolibarr/dolibarr#18373

The ref of project was overlapping the ref of doc on proposal.
The ref of doc were concatenated at the wrong place on invoice.
So i had to revert the change.


eldy:
I pushed a fixed implementation in invoice pdf with this commit
c188ef0438259048e1eb74e2b3952bf21b13e417

PR #18535, opened 2021-08-28
move version info to dolibarr info in user dropdown

### LINKED ISSUE #18418: move "version-hover" to user dropdown
move this 
image

to here
image

What do you think?

cfoellmann:
CLOSE #18418


eldy:
Removing the version from the main screenshot was already tested in the past (with version 11 i think). It brings a lot of trouble making very difficult for support team and integrator to get quickly the information from screenshots. We had to revert it and restore the version into main screen

PR #18550, opened 2021-08-30
Fix #18531 : fix not applicable error

### LINKED ISSUE #18531: Is not posible to update the Status  an event in agenda as "Not aplicable" instead "to do"
# Bug
Is not posible to update the Status  an event in agenda as "Not aplicable" instead "to do"
This means that every upodated event will be transformed ina a task "to do" even is not a task.

## Environment
- **Version**: 14.0.1
- **OS**: Linux64
- **Web server**: Apache/2.4.46  OpenSSL/1.1.1g mod_jk/1.2.41 mod_bwlimited/1.4 mod_fcgid/2.3.9 Phusion_Passenger/5.3.7
- **PHP**: 7.4.7
- **Database**: MySQL or MariaDB 5.6.29
- **URL(s)**: https://mycompany.com/comm/action/card.php?action=edit&id=3698

## Expected and actual behaviour
Now is possible to create manual events in the agenda to register the content of a Telephone call for example, and register the event with Status as "non applicable", that is very convenient since is not a task to do, is just the registration of an event.
But unfortunately, if you modify the event to update any data, then the status is switched to a type "to do", and then the event is converted as a task to do, even is not a task.

## Steps to reproduce the behaviour
- Create a manual event in the agenda with status/percentage type as "not applicable"
- Modify/update the event after creation
- Observe that is not possible to save it with the status/percentage type as "not applicable" any more.

image

Hystepik:
fix not applicable error in agenda

### DIFF
--- a/htdocs/comm/action/card.php
+++ b/htdocs/comm/action/card.php
@@ -471,7 +471,7 @@ if (empty($reshook) && $action == 'update') {
 		$apmin = GETPOST('apmin', 'int');
 		$p2hour = GETPOST('p2hour', 'int');
 		$p2min = GETPOST('p2min', 'int');
-		$percentage = in_array(GETPOST('status'), array(-1, 100)) ?GETPOST('status') : (in_array(GETPOST('complete'), array(-1, 100)) ?GETPOST('complete') : GETPOST("percentage")); // If status is -1 or 100, percentage is not defined and we must use status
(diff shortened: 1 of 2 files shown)


eldy:
The fix seems good, but it changes the key of the component form_select_status_action(). This may breaks compatibility with some external code that use it.
I commited a way to fix this without changing the component with this commit. Also it reduce the number of GETPOST.
443827ff42a04b62715cf69426fc899bc75968bb

PR #18554, opened 2021-08-30
NEW : Allow drag and drop on kits

atm-adrien:
### NEW: Drag & Drop on virtual products
**Addition of the drag & drop on "List of products / services constituting this virtual product" in the "Kits" tab of a product:**

- Added a functional drag and drop system to reorder the list
- Adding a column for drag and drop
- Adding the "Rank" column to the display
- Addition of the "Rank" field in the database (table: llx_product_association)

image


htdocs/product/composition/card.php
```diff
@@ -407,17 +423,20 @@
...
+					print '</td>';
 					print '</tr>'."\n";
 				} else {
 					$hide = '';
 					if (empty($conf->global->PRODUCT_SHOW_SUB_SUB_PRODUCTS)) {
 						$hide = ' hideobject'; // By default, we do not show this. It makes screen very difficult to understand
 					}
 
-					print '<tr class="oddeven'.$hide.'" id="sub-'.$value['id_parent'].'">';
+					print '<tr class="oddeven'.$hide.'" id="sub" parent_id="'.$value['id_parent'].'">';
```
eldy, line 435: The value for id="sub" means you will hame several time the same id on several lines. This is not allowed.
Why removing 
id="sub-'.$value['id_parent'].'">';
?

atm-adrien: When the configuration PRODUCT_SHOW_SUB_SUB_PRODUCTS is not activated : 
image

When  the configuration PRODUCT_SHOW_SUB_SUB_PRODUCTS is activated:
image

Regardless of the configuration, the sub-lines are still created, with the "hidden" state or not.
As you can see, if we leave the code as it is (keeping the parent id for the sub-lines), then these lines will be taken into account when calculating the ranks and this is a problem

eldy: May be the code is working. But you introduce a bug in html validation.
There is several time the same id into the html page. Where id can have onl one value.

atm-adrien: Is it really usefull to have an id for these lines when they have an attribute parent-id ? 
Can I remove it ? that will avoid having several times id = "sub" on different lines and that will not impact the functioning of the development

BebZ:
Best option if it is html allowed is to have the same hierarchy on the sub and sub subs and sub sub subs....

Because on some child , you will have also an order, I guess... so if possible you take the order of the sub and sub sub and you display like this to keep a good consistency : 

parent = 1
sub 1.1
sub 1.2
sub 1.3
parent 2
parent 3
sub 3.1
sub sub 3.1.1
sub sub 3.1.2
sub sub sub 3.1.2.1
sub sub sub 3.1.2.2
etc...



But I don't know if it is "possible"

htdocs/product/composition/card.php
eldy: Why removing it ? Why not keeping it with a unique name ?

atm-adrien: Because when we have an integer id for these sub-line, these will be taken for the rank count.
As you can see, on pictures above, sub-lines take ranks and disorder all the list.

Issue #18155, opened 2021-07-15
Finding Products


jbarc128:
In some combobox, like BOM product adding or Stock Masive transfer, Dolibarr can't find items. It seems that it only finds the first 250 items if it start with the same character.

There is no issues with other combobox like List of BOM.

Environment
- Version: 13.0.4
- PHP: 7.4.20
- Database: mysql 8.0.25

Problem adding product on BOM

![image](https://user-images.githubusercontent.com/68402255/125826252-31bd3c75-c6ef-4fd9-862f-1eb096af8b5c.png)

Working on List of BOM

![image](https://user-images.githubusercontent.com/68402255/125826405-6641022b-0c3c-44c3-b59b-b6030b3899ee.png)

erikvanberkum:
jbarc128, I am having the same issue with the BOM list in 13.0.2, but I am able to get a list of exactly 1000 items. If you get a list of 250 items, there must be a setting different in your and my setup where it's getting the max 250 or 1000 items from.

erikvanberkum:
There is a limit in the following query that causes the issue

```
SELECT  p.rowid, p.ref, p.label, p.description, p.barcode, p.fk_country, p.fk_product_type, p.price, p.price_ttc, p.price_base_type, p.tva_tx, p.duration, p.fk_price_expression, IF(p.stock IS NULL,0,p.stock) AS stock, pl.label as label_translated FROM llx_product as p LEFT JOIN llx_product_lang as pl ON pl.fk_product = p.rowid AND pl.lang='en_US' WHERE p.entity IN (1) ORDER BY p.ref ASC LIMIT 1000 
```

erikvanberkum:
This issue still persists in version 14.0.0

erikvanberkum:
A temporary workaround to get a full list again is to modify /htdocs/bom/tpl/objectline_create.tpl.php

around line 101-106

change 2 times `$conf->product->limit_size` to `$conf->product->limit_size=0`


```
	if (!empty($conf->global->ENTREPOT_EXTRA_STATUS)) {
		// hide products in closed warehouse, but show products for internal transfer  after filtertype 
		$form->select_produits(GETPOST('idprod', 'int'), 'idprod', $filtertype, $conf->product->limit_size=0, $buyer->price_level, $statustoshow, 2, '', 1, array(), $buyer->id, '1', 0, 'maxwidth500', 0, 'warehouseopen,warehouseinternal', GETPOST('combinations', 'array'));
[...]

PR #18556, opened 2021-08-30
Fix: select products in BOM, when adding new item the $conf->product-…

erikvanberkum:
Currently, when you want to add a new item to a BOM, not all items are loaded as `$conf->product->limit_size `is blocking to load all products.  This PR removes the limitation. #18155


eldy:
The limit is not the  $conf->product->limit_size but the setup. 
Which value did you set into option "Maximum number of products to show in combo select lists (0=no limit)" ?
Is this value the default value ?

erikvanberkum:
Not a bug wrong setting in the product admin now its working:

image

PR #18561, opened 2021-08-31
Include core massaction in reception list to allow using doMassActions hook

vincentjdc:
# New - *call doMassActions hook in reception list*
The hook doMassActions is now called when doing a mass action in the reception list

### DIFF
--- a/htdocs/reception/list.php
+++ b/htdocs/reception/list.php
@@ -40,7 +40,10 @@ $socid = GETPOST('socid', 'int');
 $massaction = GETPOST('massaction', 'alpha');
 $toselect = GETPOST('toselect', 'array');
 
-$diroutputmassaction = $conf->reception->dir_output.'/temp/massgeneration/'.$user->id;
+// Security check
+$receptionid = GETPOST('id', 'int');
+if ($user->socid) $socid = $user->socid;
+$result = restrictedArea($user, 'reception', $receptionid, '');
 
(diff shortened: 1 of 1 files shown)


htdocs/reception/list.php
```diff
@@ -166,6 +169,10 @@
 }
 
 if (empty($reshook)) {
+	$uploaddir = $conf->reception->dir_output.'/temp/massgeneration/'.$user->id;
+	$objectclass = 'Reception';
+	include DOL_DOCUMENT_ROOT.'/core/actions_massactions.inc.php';
```
eldy, line 174: There is this test into this file
	if (empty($diroutputmassaction)) {
		dol_print_error(null, 'include of actions_massactions.inc.php is done but var $diroutputmassaction was not defined');
		exit;
	}

So why having remove the

diroutputmassaction = $conf->reception->dir_output.'/temp/massgeneration/'.$user->id;

vincentjdc: You're right. I did not noticed that `$diroutputmassaction` was used in actions_massactions.inc.php. I tested it with custom massaction (that was not `builddoc`) so I did not have that error.

`$uploaddir` could be the same as `$diroutputmassaction` ?

eldy: $uploaddir is the root of module. So
$uploaddir = $conf->reception->dir_output;
$diroutputmassaction is a temporary sub-dir of this directory.

eldy:

PR #18570, opened 2021-09-01
Fix error of no button to classify credit note partially paid

henrynopo:
Just discovered this bug. As the remainder of payment is negative for credit note, the current coding does not work to classify credit note as partially paid, same for both supplier and customer sides.

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -4043,12 +4043,17 @@ if ($action == 'create') {
 
 		$formconfirm = $form->formconfirm($_SERVER["PHP_SELF"].'?facid='.$object->id, $langs->trans('UnvalidateBill'), $text, 'confirm_modif', $formquestion, "yes", 1);
 	}
+	
+	$sign = 1;
+	if ($object->type == Facture::TYPE_CREDIT_NOTE) {
+		$sign = - 1;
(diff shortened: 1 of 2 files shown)


eldy:
Can you describe howto reproduce the bug? May be you can create a case on demo.dolibarr.org and give the ref.

henrynopo:
> Can you describe howto reproduce the bug? May be you can create a case on demo.dolibarr.org and give the ref.

This is how I dicovered this bug (v14.0.1):
- Supplier order cancelled. 
- So there was a credit note generated on the Supplier linked to the cancelled order
- Supplier refunded the Credit Note amount (e.g. USD10,000). However, I received only USD 99,800 for which the difference was due to intermediary bank's charge.
- Therefore, I entered a payment of USD 99,800 received on the credit note. There should be a USD 200 as the remainder. At this time, there is no button to classify the credit note as partially paid.
- I have not tried whether this might happen on credit note to customers since we always refund the billed amount.

As coded, the credit note amounts should be negative and the remainder USD 200 is also calculated as "-200", which means the credit note is not fully paid. However, the current coding is to be justified for invoices which amounts are always positive, and the same for their remainders in case of partial payment.

That's why I proposed to update the coding the amend the issue.

PR #18571, opened 2021-09-01
Invoice remainder to pay in multicurrency - updated the calculation

henrynopo:
Having found out the existing functions to fetch the various amounts in multicurrency (invoice, discounts/credits, amount paid), I have updated the coding for the newly proposed remainder to pay / pay back in multicurrency.

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -5075,6 +5075,11 @@ if ($action == 'create') {
 
 		// Remainder to pay Multicurrency
 		if ($object->multicurrency_code != $conf->currency || $object->multicurrency_tx != 1) {
+			$multicurrency_payment = $invoice->getSommePaiement(1);
+			$multicurrency_creditnotes = $invoice->getSumCreditNotesUsed(1);
+			$multicurrency_deposits = $invoice->getSumDepositsUsed(1);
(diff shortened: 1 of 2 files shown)


htdocs/compta/facture/card.php
```diff
@@ -5129,6 +5134,11 @@ function js_recalculate_revenuestamp(){
 
 		// Remainder to pay back Multicurrency
 		if ($object->multicurrency_code != $conf->currency || $object->multicurrency_tx != 1) {
+			$multicurrency_payment = $invoice->getSommePaiement(1);
```
eldy, line 5137: There is no reason to make a lot of subrequest to get the remain to pay in multicurrency.
We already have the remain to pay and the rate. 
So remain to pay in currency should be the rate * remain to pay in main currency.

henrynopo: Due to the forex conversion and the rounding, it happens very often that the factual remainder is 0 in the original currency but 0.01 in the main currency. The below shows an example after I applied the new coding, otherwise it will results in 0.01 in both currencies. To supplement, in my case 1 local dollar = 0.75 USD. You can double check, the billed amount rounds up and resulted in the excess of 0.01 local currency.

image

PR #18575, opened 2021-09-01
Update facture.class.php

nicolasb827:
copy ref_client from facrec
substitute keys into ref_client. (eg: __INVOICE_YEAR__ etc...)

# New ref_client is not copied from facture_rec
When creating invoices from recurring invoices, original ref_client field is left blank on new invoice.
This fix allow ref_client to be copied, dans substituted to create date aware ref_client field values.
This allow recurring invoices like:
 [Company] Invoice __INVOICE_YEAR__-__INVOICE_MONTH__ 

to be instantiated for each month like:
 [Company] Invoice 2021-09


eldy:
The PR seems empty. May be change was already done by another PR into dev branch by someone else ?

htdocs/compta/facture/class/facture.class.php
```diff
@@ -527,6 +527,11 @@ public function create(User $user, $notrigger = 0, $forceduedate = 0)
 			$this->mode_reglement_id = GETPOST('mode_reglement_id', 'int') > 0 ? ((int) GETPOST('mode_reglement_id', 'int')) : $_facrec->mode_reglement_id;
 			$this->fk_account        = GETPOST('fk_account') > 0 ? ((int) GETPOST('fk_account')) : $_facrec->fk_account;
 
+			// if ref_client was not set by caller, use the facref title
+			if (empty($this->ref_client)) {
+				$this->ref_client =  trim($_facrec->title);
```
eldy, line 532: The ref_client is the reference of the invoice on the client side (so often know by the client). Why storing the label of the recuring invoice inside this field ?

eldy:
Even if this PR goal is to fix a bug, it seems it introduces more problems than it solves, so request is closed.

PR #18576, opened 2021-09-01
Update ChangeLog

dolibit-ut:


eldy:
The changelog for v14.0 contains the list of changes for v14.0 that has been released so content can't evolves. So you must not add the change done into branch v14 after the release. It makes the changelog wrong. The changelog will be generated when doing the v14.0.2 and all fixes will be added when the changelog generation for 14.0.1 to 14.0.2 will be executed.

PR #18593, opened 2021-09-03
[NEW] MAJ tcpdi parse library

kkhelifa-opendsi:
[NEW] Update tcpdi parse library

### DIFF
--- a/htdocs/includes/tcpdi/tcpdi_parser.php
+++ b/htdocs/includes/tcpdi/tcpdi_parser.php
@@ -1,9 +1,9 @@
 <?php
 //============================================================+
 // File name   : tcpdi_parser.php
-// Version     : 1.0
+// Version     : 1.1
 // Begin       : 2013-09-25
-// Last Update : 2013-09-25
+// Last Update : 2016-05-03
 // Author      : Paul Nicholls - https://github.com/pauln
 // License     : GNU-LGPL v3 (https://www.gnu.org/copyleft/lesser.html)
 //
@@ -44,7 +44,7 @@
  * This is a PHP class for parsing PDF documents.<br>
  * @author Paul Nicholls
  * @author Nicola Asuni
- * @version 1.0
+ * @version 1.1
  */
 
 // include class for decoding filters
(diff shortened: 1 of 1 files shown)


htdocs/includes/tcpdi/tcpdi_parser.php
```diff
@@ -484,7 +484,7 @@ protected function decodeXrefStream($startxref, $xref=array()) {
 			$v = $sarr[$key];
 			if (($key == '/Type') AND ($v[0] == PDF_TYPE_TOKEN AND ($v[1] == 'XRef'))) {
 				$valid_crs = true;
-			} elseif (($key == '/Index') AND ($v[0] == PDF_TYPE_ARRAY AND count($v[1]) >= 2)) {
+            } elseif (($key == '/Index') AND ($v[0] == PDF_TYPE_ARRAY AND count($v[1] >= 2))) {
```
eldy, line 487: This looks strange
AND count($v[1] >= 2))) {
Old code
AND count($v[1]) >= 2)) {

Is it a fix on your own or official project patch ?

eldy:
Is it a fix on your own or official project patch ? If project patch, where does he come from ?

htdocs/includes/tcpdi/tcpdi_parser.php
kkhelifa-opendsi: it's come from here https://github.com/pauln/tcpdi_parser

eldy: The old code was a fix when making lib compatible with php8. So i just think this fix was not included into the version at pauln. But i think we must keep it.
Can you try to also add this fix 
AND count($v[1]) >= 2)) {
And tell if your use of tcpdfi still works for the case you tested ?

eldy:
@kkhelifa-opendsi 
Can you try to also add this fix
`} elseif (($key == '/Index') AND ($v[0] == PDF_TYPE_ARRAY AND count($v[1]) >= 2)) {`   (old code)
instead of 
`} elseif (($key == '/Index') AND ($v[0] == PDF_TYPE_ARRAY AND count($v[1] >= 2))) {`   (new one with change on parenthesis for the count)
And tell if your use of tcpdfi still works for the case you tested ?

eldy:
I think tcpdf was updated since , so i close the pr.

PR #18596, opened 2021-09-03
NEW Add forced exit of the code if the merged PDF is corrupted, print and log the error

kkhelifa-opendsi:
NEW Add forced exit of the code if the merged PDF is corrupted, print and log the error


eldy:

PR #18598, opened 2021-09-04
Increase the width of column for General Set-ups - Update dolibarr.php

henrynopo:
Some names of the general set-ups are too long and are cut off by the limited width 300px.

Suggest to double it (here updated). Or it would be considerable to take out the limit of width.


eldy:
Can you provide a screenshot before/after ?

henrynopo:
Before change
image

After change
image

The general set-up THIRDPARTY_CAN_HAVE_CATEGORY_EVEN_IF_NOT_CUSTOMER_PROSPECT_SUPPLIER is the longest and 600px is just nice to show it completely.

Issue #17963, opened 2021-06-17
Add a "billable" link to the Order widget


bos4711:
The "Purchase order" widget has a link to "Open" orders and "Awaiting reception" orders. It would be nice if the Order widget could have a link to Billing/Payment->Billable orders.

![Untitled](https://user-images.githubusercontent.com/11036018/122397735-89832d80-cf79-11eb-8a36-e0d9d871306e.png)

atm-maxime:
So it will be the count of all customer orders (a least with validated status) where the "billed" flag is "no". And it would be called "billable orders".

bos4711:
Yes, where the link goes to "/commande/list.php?leftmenu=orders&search_status=-3&billed=0&contextpage=billableorders".

atm-maxime:
Are you sure you want to have orders that are not yet delivered ? My guess is that a billable order is a delivered order not yet billed.

bos4711:
Your guess is correct.

PR #18600, opened 2021-09-04
New : add a dashboard showing number of orders to bill #17963

atm-maxime:
Add a new value to the customer order Dashboard showing the number of orders to bill.
Currently it's all delivered orders that are not flagged as billed. For me there's no delay notion, once an order is delivered, is has to be billed.


atm-maxime:
image

htdocs/commande/class/commande.class.php
```diff
@@ -3512,18 +3512,39 @@ public function load_board($user)
 		}
 		$sql .= $clause." c.entity IN (".getEntity('commande').")";
 		//$sql.= " AND c.fk_statut IN (1,2,3) AND c.facture = 0";
-		$sql .= " AND ((c.fk_statut IN (".self::STATUS_VALIDATED.",".self::STATUS_SHIPMENTONPROCESS.")) OR (c.fk_statut = ".self::STATUS_CLOSED." AND c.facture = 0))"; // If status is 2 and facture=1, it must be selected
+		if ($mode == 'toship') {
+			// An order to ship is an open order (validated or in progress)
+			$sql .= " AND c.fk_statut IN (" . self::STATUS_VALIDATED . "," . self::STATUS_SHIPMENTONPROCESS . ")";
+		}
+		if ($mode == 'tobill') {
```
eldy, line 3519: The status "to bill" (the order was not billed) is no more saved into the status of the order. Because some company need to have the order "to bill" whatever is the status of shipment (it is a different process) so we must test on 
"facture" = 0 or not to know if it is billed or not.

However some want to the information "To bill" when order is shipped. And some want the information "to bill" as soon as order has been created. Due to 2 different use case, the information "to bill" was not visible on the workboard (because in half case, it is wrong because it depends on the use case).


So i suggest to add an option into module "Order" to ask when the order must be billed:
- As soon as validated   => test "to bill" will be done on  c.facture = 0";
- When an order as been shipped   => test "to bill" will be done the way you suggest  c.fk_statut IN (" . self::STATUS_CLOSED . ") AND c.facture = 0";

BebZ:
Thanks, my 2 cents 

- What can be the behaviour when an order is "partially shipped" and the first shipment is invoiced and closed ?
do you have a "partially billed" status of orders ?

As soon as validated => count for "to bill" will be done after a filter on c.facture = 0";

When an order as been shipped => count for "to bill" will be done the way you suggest c.fk_statut IN (" . self::STATUS_CLOSED . ") AND c.facture = 0";

htdocs/commande/class/commande.class.php
dpriskorn: @atm-maxime I agree with eldy. Would you be willing to rework the pull request so the user get a new setting "When to bill" -> "After validated order" or "After validated shipping sent"?

eldy: @atm-maxime Can you change the test on  "facture = 1"  or if you prefer closing the PR ?

atm-maxime: @eldy I can update this PR. I think adding an option in the order module, just to display this workboard indicator will mislead the user in thinking that the option changes the workflow on orders.
What I suggest is to change the workboard like this :

TODAY :
image
(note that the link on the "open" orders is wrong, it uses the filter showing closed orders also, I'll fix this too)

AFTER :
image
À livrer => status IN (1,2)
À facturer => status IN (1,2,3) AND facture = 0
Livrée / À facturer => status = 3 AND facture = 0

What do you think ?

atm-maxime: @eldy ?

eldy: > @eldy I can update this PR. I think adding an option in the order module, just to display this workboard indicator will mislead the user in thinking that the option changes the workflow on orders. What I suggest is to change the workboard like this :
> 
> TODAY : image (note that the link on the "open" orders is wrong, it uses the filter showing closed orders also, I'll fix this too)
> 
> AFTER : image À livrer => status IN (1,2) À facturer => status IN (1,2,3) AND facture = 0 Livrée / À facturer => status = 3 AND facture = 0
> 
> What do you think ?

Looks good to me.

eldy:
@BebZ The case you describe is a different use case. It means in your example that you invoice not the order but the shipment. For this, we should have a button "Create invoice" on a shipment to generate the invoice from a shipment file instead of an order and billing status will be followed from the shipment. For this cas, we should have another widget with "Shipments open and Shipments to bill".

htdocs/commande/class/commande.class.php
atm-maxime: Done !

atm-maxime:
@eldy

PR #18602, opened 2021-09-04
Manufacturing order pdf template

mschamp:
File template based on cornas to generate pdf file from manufacturing orders


eldy:
I have rename the template.

PR #18611, opened 2021-09-05
Feature Request : Add an option to postpone notifications (after ref and PDF generation)

### LINKED ISSUE #18588: Feature Request: Invoice notification after PDF generation
# Feature Request
Send notification *after* the PDF generation on Invoice.

## Use case
When using template invoice to auto generate invoice, the notification is send to billing contact before the PDF is generated.

They are two issues:

1. The mail template use the "(PROVxx)" as invoice reference for recurring invoice. 
2. They are no attached PDF in the notification mail or the attached file is the draft invoice.

The notification send to billing contact may be send after the invoice reference generation and PDF generation.

This feature is very useful for invoice template generation to immediately send generated invoice by mail.

## Work to be implemented
Email is send by trigger system. The same trigger which verify invoice can be created. The next step after triggers is generating the invoice reference and sending the mail.

Two proposals :
1. Use the same trigger system for invoice ref generation and PDF generation.
2. Use another trigger after the ref is assigned on invoice and PDF generated.

The first proposition will let invoicing system to be more modulable. Developers will be able to create trigger between ref and PDF generation to change invoice reference or a trigger after PDF generation to send it on external storage.

The second proposition use less structure modification. It's more easy to implement.

bontiv:
# Close #18588

I add a new setup option MAIN_POSTPONE_TRIGGER. It postpone trigger execution after the invoice is validated and generate the PDF if MAIN_DISABLE_PDF_AUTOUPDATE is empty.

Postpone triggers and generate PDF during validation have many improvements:

1. The notification module can send final PDF in the notification.
2. Recurring factures with invoice template and notification module enable send PDF and use the good invoice reference in the notification.
3. Others developers can use this option to use postponed triggers for sending the generated PDF on external storage (OneDrive / Google Drive / iCloud) or make action on external customer panel.

## Why not postponing all trigger?
Trigger may be used by modules to verify if invoice can be validate or change / add informations on invoice. In this cas, postponing trigger may brake somme modules.

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -2876,11 +2876,10 @@ class Facture extends CommonInvoice
 				// POSTPONED TRIGGER
 				// In case of postpone PDF and triggers
 				if (!empty($conf->global->MAIN_POSTPONE_TRIGGER)) {
-
 					$this->statut = self::STATUS_VALIDATED;
 					$this->status = self::STATUS_VALIDATED;
 					$this->brouillon = 0;
-					$this->date_validation = $now;	
+					$this->date_validation = $now;
 
 					if (empty($conf->global->MAIN_DISABLE_PDF_AUTOUPDATE)) {
 						$outputlangs = $langs;
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -16,7 +16,8 @@
  * Copyright (C) 2013      Florian Henry         <florian.henry@open-concept.pro>
  * Copyright (C) 2016      Ferran Marcet         <fmarcet@2byte.es>
  * Copyright (C) 2018      Alexandre Spangaro    <aspangaro@open-dsi.fr>
- * Copyright (C) 2018      Nicolas ZABOURI        <info@inovea-conseil.com>
+ * Copyright (C) 2018      Nicolas ZABOURI       <info@inovea-conseil.com>
+ * Copyright (C) 2021      Remi BONNET           <prog.bontiv@gmail.com>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -2826,7 +2827,7 @@ class Facture extends CommonInvoice
 			}
 
 			// Trigger calls
-			if (!$error && !$notrigger) {
+			if (empty($conf->global->MAIN_POSTPONE_TRIGGER) && !$error && !$notrigger) {
 				// Call trigger
 				$result = $this->call_trigger('BILL_VALIDATE', $user);
 				if ($result < 0) {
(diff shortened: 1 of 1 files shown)


BebZ:
Hello, how in practice it is "postponed" ? 

We should enter a value in seconds in the hidden option : MAIN_POSTPONE_TRIGGER ? 

So with your commit, customer invoices are impacted, but we can copy this behaviour to other objects like Customer orders / supplier orders / commercial propositions ?

Thanks for this little but very clever tweak

bontiv:
Hello,

It's the execution of triggers which is postponed. The default behavior for validation is:

1. Generate reference
2. Execute invoice validation
  a. Verify if invoice can be validated
  b. Run triggers
  c. Change the invoice to validate state
3. Generate the PDF

Some modules use triggers to make some actions on invoice validation like notification module which send mail when an invoice is validated. The problem is about when triggers is executed in the stack. Its execution occur before the invoice is in validated state. The notification is more "This draft invoice is about to be validated" than "You have a new validated invoice".

When you set MAIN_POSTPONE_TRIGGER, the trigger is raise after the invoice state change to validate. The new validation execution stack is:

1. Generate reference
2. Execute invoice validation
  a. Verify if invoice can be validated
  b. Change the invoice to validate state
  c. Generate PDF if not auto generation is not disabled
  d. Run triggers

Triggers is raise after the invoice is validated and PDF generated. The notification is now "This new invoice is validated" instead of "This draft will be validated".

eldy:
This break the idea of "triggers" (an event that is included into the same transation to make all or nothing).
Also imagine that the validate() method is included into another more global transaction, example the Batch XXX that do

- Begin transaction
- Read invoice
- Validate invoice
- Read another object
- Update another object
- Commit if no error or rollback.

It means the commit or rollback will be done after your post-ponned trigger and the problem will be the same (if error, transaction will be rollbacked after your trigger).

There is no real solution to be sure that something is done after things are effectively commited, because only the process that had initiated the transaction and close it know this. For web pages, this can be done by adding code in the hook "beforeBodyClose" but it won't be executed when method is called by an API or another external tool.

And your solution will work only for the case of the validate method is called with no open transaction yet. In such a case, if you really need to have transaction commited before doing your job, you can use current trigger, check the nb/level of opened transaction, close them by a commit your self by your trigger, then do the job and then reopen a transaction at same level than it was before you closed it.

bontiv:
Hi,

Ok, I understand what do you mean. But it's a real problem to have notification trigged before the commit too. If an error is raised, the notification sent cannot be rollback.

In other hand, it make no sens to send a notification before having the document reference and PDF.

Using beforeBodyClose may not be the best idea. I don't think we can know witch documents is created or modified during the page generation.

Another solution may be another trigger or hook after the ref is set (line 2872) for notifications, PDF generation, saving document in another secure location (Dropbox / OneDrive / Box / Google Drive / etc..), sending document with another canal than mails.

I will change my PR to have the trigger in the same transaction level as the first trigger and rollback directory rename if an error occur during trigger.

eldy:
I still don't understand the goal of this PR.
It seems you try to modify the position of the trigger so you can execute code after the PDF file has been generated and only once the full transaction is finished. So notification is sent only if everything is ok.
If you need to execute code "only if everything is ok", trigger is not the good choice (a trigger must always be part of the transaction, a lot of modules rely on this so we can't change this or the data integrity will ne be guaranteed anymore). 

Using a hook is a better choice. There is a lot of hook everywhere in application.
In you use case, it seems the hook  afterPDFCreation would be a good choice (it is called after the generation of document so executed after the database transaction has been validated)

PR #18617, opened 2021-09-06
FIXED BUG #18589 about Project tasks linked file access denied

alsoft10:
# Instructions
#18589 linked files with the task is not working. And it is fixed. Hopefully it will be merged.

### DIFF
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -1004,6 +1004,7 @@ function dolCheckVirus($src_file)
 }
 
 
+
 /**
  *	Make control on an uploaded file from an GUI page and move it to final destination.
  * 	If there is errors (virus found, antivir in error, bad filename), file is not moved.
@@ -1839,6 +1840,7 @@ function deleteFilesIntoDatabaseIndex($dir, $file, $mode = 'uploaded')
 
 	$db->begin();
 
+
 	$rel_dir = preg_replace('/^'.preg_quote(DOL_DATA_ROOT, '/').'/', '', $dir);
 
 	$filename = basename($file);
@@ -2293,7 +2295,7 @@ function dol_check_secure_access_document($modulepart, $original_file, $entity, 
 
 	// Find the subdirectory name as the reference. For exemple original_file='10/myfile.pdf' -> refname='10'
 	if (empty($refname)) {
-		$refname = basename(dirname($original_file)."/");
+		$refname = basename(dirname($original_file, 2)."/");
 	}
 
 	// Define possible keys to use for permission check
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/files.lib.php
```diff
@@ -2293,7 +2295,7 @@ function dol_check_secure_access_document($modulepart, $original_file, $entity,
 
 	// Find the subdirectory name as the reference. For exemple original_file='10/myfile.pdf' -> refname='10'
 	if (empty($refname)) {
-		$refname = basename(dirname($original_file)."/");
+		$refname = basename(dirname($original_file, 2)."/");
```
eldy, line 2298: If you add ,2 parameter it means that for original_file = a/b/c, dirname(original_file, 2)  will return 
a
But here we want 
b
Old code seems good.

alsoft10: I partially agree with you as I am not experienced as you are in Dolibarr. But here...
a= Project Reference
b= Task Reference inside the above project
c= Filename

Dolibarr creates directory for each and every project to store linked files in it. Now tasks files are stored new directory inside the project directory. So you must know the **_a_**, before you go to **_b_**. I have fixed it and it is completely working in my system pretty good. I will suggest to check once again. Thank you.

JoAnMaFe:
18617

In v14.0.2 it doesn't work for me either.
Access denied.
You try to access a page, area or functionality of a disabled module or without an authenticated session or not allowed to your user. 
Permissions for this user can be assigned by Dolibarr administrator via menu Start-> Users. 

In addition, thumbnails of events are not visible, they can be opened, but a blank icon is displayed.

edupulpillo:
Same here.
The problem persist in the version 14.0.2, with old files attached in a task, and also new files attached in v14.0.2

How can be requested to "reopen" an issue in GIthub/Dolibarr ?

Euroiranian:
Hi, 
I just updated to V 14.0.2 and problem persists. 

Anyone following this issue????? 

Thanks.

htdocs/core/lib/files.lib.php
Euroiranian: Good morning,

Dolibarr V 14.0.2 latest update already installed and problem persisting.

I've also changed code following your instructions but still facing the same problem. The most surprising is that depending in system, problem is persisting or not. I mean, I tested 2 desktop PCs, 3 mobile/tablets and all of them working fine, but from laptops (Lenovo & MacBook Air) we are facing problems, no matter from which country do we try or the navigator we use (Chrome or Edge). I don't know if it makes sense, but seems that problems are generated by laptops??!!

Best regards and grateful, hope you could help us. Thanks.

JoAnMaFe:
https://github.com/Dolibarr/dolibarr/issues/18716 """"Solved""""

edupulpillo:
In 14.0.3 the problem persist, returning a problem of permissions, even using the admin user with all the rights enabled:

Access denied.
You try to access to a page, area or feature of a disabled module or without being in an authenticated session or that is not allowed to your user.

Current login: admin
Permission for this login can be defined by your Dolibarr administrator from menu Home->Users.

edupulpillo:
The problem persist in 14.0.5

edupulpillo:
Problem solved with commit  5d5a219 explained in #19770 
Please close this issue

PR #18618, opened 2021-09-06
Fix url template API login

PyroShape:


htdocs/api/admin/index.php
```diff
@@ -140,7 +140,7 @@
 // Show message
 $message = '';
 //$url = $urlwithroot.'/api/index.php/login?login=<strong>auserlogin</strong>&password=<strong>thepassword</strong>[&reset=1]';
-$url = $urlwithroot.'/api/index.php/login?login=auserlogin&password=thepassword[&reset=1]';
+$url = $urlwithroot.'/api/index.php/login?login=auserlogin&password=thepassword&reset=1';
```
eldy, line 143: We don't want to have the reset=1 enabled by default. If we click on link by error, we will erase (reset) the api login.
It is an option, this is why it is inside [ ]

PyroShape: What about removing the reset parameter ?
That way the link will working but without any key loss if a user click on it.

eldy: The goal of the page that show the link is to show the link + all possible parameters so user can choose which link he want.
Also not that the api using login and pass should never be used. Good solution to use api is to use the token.
A good way to fix this is to completely reomve the link (because we should not use it, we keep it for compatibility) and replace it with a message to say that to use API, we must first set a token on the user and the use this token.

PR #18620, opened 2021-09-06
FIX Product accountancy affectation with product_perentity activated

aspangaro:


htdocs/accountancy/admin/productaccount.php
```diff
@@ -189,9 +189,16 @@
 					$db->begin();
 
 					if (!empty($conf->global->MAIN_PRODUCT_PERENTITY_SHARED)) {
-						$sql = "INSERT INTO ".MAIN_DB_PREFIX."product_perentity (fk_product, entity, '".$db->escape($accountancy_field_name)."')";
-						$sql .= " VALUES (".((int) $productid).", ".((int) $conf->entity).", '".$db->escape($accounting->account_number)."')";
-						$sql .= " ON DUPLICATE KEY UPDATE ".$accountancy_field_name." = '".$db->escape($accounting->account_number)."'";
+						if ($this->db->type == 'pgsql') {
```
eldy, line 192: $this is not defined here. Should make a fatal error.

aspangaro: Fix Eldy

htdocs/accountancy/admin/productaccount.php
```diff
@@ -189,9 +189,16 @@
...
-						$sql .= " ON DUPLICATE KEY UPDATE ".$accountancy_field_name." = '".$db->escape($accounting->account_number)."'";
+						if ($db->type == 'pgsql') {
+							$sql = "INSERT INTO ".MAIN_DB_PREFIX."product_perentity (fk_product, entity, " . $db->escape($accountancy_field_name) . ")";
+							$sql .= " VALUES (" . ((int) $productid) . ", " . ((int) $conf->entity) . ", \"" . $db->escape($accounting->account_number) . "\")";
+							$sql .= " ON CONFLICT (fk_product, entity)";
+							$sql .= " DO UPDATE SET " . $db->escape($accountancy_field_name)." = \"" . $db->escape($accounting->account_number) . "\"";
+						} else {
+							$sql = "INSERT INTO " . MAIN_DB_PREFIX . "product_perentity (fk_product, entity, " . $db->escape($accountancy_field_name) . ")";
+							$sql .= " VALUES (" . ((int) $productid) . ", " . ((int) $conf->entity) . ", '" . $db->escape($accounting->account_number) . "')";
+							$sql .= " ON DUPLICATE KEY UPDATE " . $db->escape($accountancy_field_name) . " = '" . $db->escape($accounting->account_number) . "'";
```
eldy, line 200: Using non common/standard sql is not allowed in dolibarr as it is non portable.
I should not have validated the previous code that contains the "ON DUPLICATE KEY UPDATE".
What do you try to do ?
Insert if it does not exists and ignore if it already exists ?
-> For such case, solution is select and insert if not found
or ALWAYS replace value with new one
-> For such case, solution is delete (delete may delete nothing) + insert

PR #18624, opened 2021-09-06
Close #18584 : Code enhancement of MO 

### LINKED ISSUE #18584: Fix / enhancement on MO
When you create a MO with no BOM, only the product to manufacture is known. But after creation, the section product to build is empty. We must have at least one line with the product.
If we choosed "disassemble MO", the first initialized line must be into the left section "to consume"

Capture du 2021-09-02 17-26-52

Also when the right section is empty, we should have a link "Add new line to produce" like the link "Add new line to consume" on the left.

Hystepik:
Modification of Mo code to have product in consume or produce when creating without Bom

### DIFF
--- a/htdocs/mrp/mo_production.php
+++ b/htdocs/mrp/mo_production.php
@@ -975,7 +975,9 @@ if ($object->id > 0 && (empty($action) || ($action != 'edit' && $action != 'crea
 		}
 		$newlinetext = '';
 		if ($object->status != $object::STATUS_PRODUCED && $object->status != $object::STATUS_CANCELED && $action != 'consumeorproduce' && $action != 'consumeandproduceall') {
-			$newlinetext = '<a href="'.$_SERVER["PHP_SELF"].'?id='.$object->id.'&action=addproduceline">'.$langs->trans("AddNewProduceLines").'</a>';
+			if ($nblinetoproduce == 0 || $object->mrptype == 1) {
+				$newlinetext = '<a href="'.$_SERVER["PHP_SELF"].'?id='.$object->id.'&action=addproduceline">'.$langs->trans("AddNewProduceLines").'</a>';
+			}
 		}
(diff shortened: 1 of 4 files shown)


htdocs/mrp/mo_production.php
```diff
@@ -962,7 +967,17 @@
...
-		print load_fiche_titre($langs->trans('Production'), '', '');
+		$nblinetoproduce = 0;
+		foreach ($object->lines as $line) {
+			if ($line->role == 'toproduce') {
+				$nblinetoproduce++;
+			}
+		}
+		$newlinetext = '';
+		if ($object->status != $object::STATUS_PRODUCED && $object->status != $object::STATUS_CANCELED && $action != 'consumeorproduce' && $action != 'consumeandproduceall') {
+			$newlinetext = '<a href="'.$_SERVER["PHP_SELF"].'?id='.$object->id.'&action=addproduceline">'.$langs->trans("AddNewProduceLines").'</a>';
```
eldy, line 978: I think you forgot the test 
`if ($nblinetoproduce == 0) {`
here. So the link to add a product to product is suggested only if section is empty (no need for the moment if already filled)
Am i right ?

Hystepik: Ye but the problem with this is that in Disassemble mode we can only add 1 line of production maybe 
``` 
if ($nblinetoproduce == 0 || $object->mrptype == 1) {
```
could be a good fix which permit to have more than one production in Disassemble mode and only 1 production in Manufacturing

eldy: Correct. Go this way.

WebAuxilium: Hello
Thanks it's works
Therefore in tab mrp/mo_card.php,  we already enter the product to produce.
So why in mrp/mo_production.php, we must must add this product.
It will be better to transfert the product to produce from mo_card to mo_production.php.
Stef

Issue #18268, opened 2021-07-28
Banking : ICS not well implemented - Can't make SEPA paiement


warnerbryce:
# Bug
In Dolibarr v14 i can't make a SEPA Prelevement (WithdrawlsReceipt), an error is thrown : You don't have ICS in your Bank Account
But the ICS number is entered in the good section in the banking details.

## Environment
- **Version**: 14.0.0
- **OS**: Ubuntu Server 20.04
- **Web server**: Nginx
- **PHP**: Php7.4
- **Database**: MariaDB 10.3.29

## Expected and actual behavior
Click on "Créer fichier de prélèvement automatique" and you get your xml file

## Steps to reproduce the behavior
Configure a Bank Account with a ICS number
Configure a ThirdParty with an IBAN and generate a SEPAmandate
Make an invoice for this ThirdParty with a SEPA paiement
Go into Banking module and make an xml file to send to your bank.

Analysing the SQL database something is missing when you enter the ICS number in your bank account. 
in llx_bank_account you have two columns with ICS : 
ics and ics_transfer
When you enter your ICS number and save, the database get this number and write it in ICS but nothing is written in ics_transfer. 
For testing i copied the ICS number in ics_transfer and then it works.

bontiv:
https://github.com/Dolibarr/dolibarr/blob/6c19524fd76dbb4c88353f4dea74ce3eeec7edc0/htdocs/compta/prelevement/create.php#L98

Change to
 ```
  if (empty($bank->ics_transfer)) {
  ```
  
  Workaroud without code, enable the module named "Payment by bank tranfer". You will see a field "ICS Transfer" in the bank account. Copy and past your ICS on both fields (ICS and ICS transfer). You will be able to create a SEPA.
 
 French explanation:
 Le problème vient d'une erreur dans la séparation de l'ICS entre pour le virements pour les fournisseurs et l'ICS pour les prélèvement. La solution est de ne PAS vérifier la présence de l'ICS pour virement lors d'un prélèvement ou d'activer le module pour les virements et remplir le champs ICS pour virement.

bontiv:
They are also a bug here:
https://github.com/Dolibarr/dolibarr/blob/6c19524fd76dbb4c88353f4dea74ce3eeec7edc0/htdocs/compta/prelevement/card.php#L331

Same in line 335 and 338.

[...]

PR #18625, opened 2021-09-06
Fix : unable to create a direct debit if ics_transfer is empty

jyhere:
Fix : unable to create a direct debit if ics and ics_transfer are empty even though ics_transfer is only used for bank transfer.


bontiv:
This PR also fix #18268. 
Thanks!

PR #18632, opened 2021-09-06
Manage price_level with membership

ptibogxiv:
Manage / change price_level if thirdparties with membertypes of linked member
->first PR modify SQL

### DIFF
--- a/htdocs/install/mysql/migration/14.0.0-15.0.0.sql
+++ b/htdocs/install/mysql/migration/14.0.0-15.0.0.sql
@@ -68,5 +68,6 @@ ALTER TABLE llx_mrp_mo ADD COLUMN mrptype integer DEFAULT 0;
 
 DELETE FROM llx_menu WHERE type = 'top' AND module = 'cashdesk' AND mainmenu = 'cashdesk';
 
-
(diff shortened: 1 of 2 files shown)


eldy:
Is the use case this one : 
Members will be able to buy products or services and you want to set n different prices on a product ?

ptibogxiv:
> Is the use case this one :
> Members will be able to buy products or services and you want to set n different prices on a product ?

non just manage the change of price level of the thirdparties linked to a member ie: membertype 1= price level 1, membertype 2 = price level 2, membertype 3 = price level = 1; with big fundation, we can forget changing the price level of  a thirdparty when changing membertype (i will add option or setting for disabled automatic change if somebody don't want changing pricelevel when changing memebertype)
buying products of service ara only for thirparties

eldy:
So it means you decided for your use case that there is n prices for product, and for membertype, there is also n member type.
But this is not the case for most users. A user may have 5 price level (according to company size for example) and have 2 member types only (vip and standard). 
This change seems to be a feature that is very specific to a choice of implementing features for one use case. It would be better, it you need such interaction to manage this into a trigger of an external module (the module "Update price level of thirdparties from the membertype". This module will be responsible of maintaining the matching between price level and id of member type).

PR #18637, opened 2021-09-07
FIX multicompany transverse mode compatibility

hregis:


htdocs/user/class/user.class.php
```diff
@@ -3298,14 +3298,21 @@ public static function replaceThirdparty(DoliDB $db, $origin_id, $dest_id)
 	public function load_state_board()
 	{
 		// phpcs:enable
+	    global $conf;
 
 		$this->nb = array();
 
-		$sql = "SELECT count(u.rowid) as nb";
+		$sql = "SELECT DISTINCT count(u.rowid) as nb";
```
eldy, line 3305: What means a DISTINCT COUNT() ?
You means   COUNT(DISTINCT  ) ?

hregis: thank you

PR #18641, opened 2021-09-07
14.0 fix takepos payment methods

### LINKED ISSUE #18635: Regression: Takepos issue with cash method and dictionnaries
# Bug
There is a mess in payment methods. First, in the dictionaries, changing a payment method's label has no visible effect (for any payment method):

Peek 07-09-2021 17-30

Also, it has no visible effect on TakePos UI. Whatever is set in the dictionaries, the button always shows "Espèce" as the label.

Worst, for "Espèces" if the label value is not set to "cash", takepos will not accept the payment, complaining about bank account not set for that method. ([also seen in the forum](https://www.dolibarr.fr/forum/t/impossible-de-valider-un-paiement-en-especes/33868/2))

image

This was notices after upgrading from version 12 to 13, and it is still an issue in 14.0.1

I looked at the code, and it looks messy to me with many things like:

```
if ($paycode == 'LIQ') $paycode = 'CASH';
if ($paycode == 'CB')  $paycode = 'CB';
if ($paycode == 'CHQ') $paycode = 'CHEQUE';
```
```
	if ($paycode == 'LIQ') {
		$paycode = 'cash';
```
```
$paycode = $pay;
if ($pay == 'cash')   $paycode = 'LIQ'; // For backward compatibility
if ($pay == 'card')   $paycode = 'CB'; // For backward compatibility
if ($pay == 'cheque') $paycode = 'CHQ'; // For backward compatibility
```

## Environment
- **Version**: 14.0.1, 13.0.4, develop
- **OS**: Ubuntu
- **Web server**: Apache, any version
- **PHP**: 7.4
- **Database**: MySQL
- **URL(s)**: /htdocs/admin/dict.php?id=13 , /htdocs/takepos/index.php?idmenu=186&mainmenu=takepos&leftmenu=

Well, a label is a label, and it has a "code" to abstract its usage. The set labels should be used for payment method's labels in the UI, and changing it _shouldn't_ affect core functionalities.

loxK:
# Fix #18635

- Only check takepos configuration if a terminal is chosen, to prevent error
- Fix cash payment in some cases. Make pay requests language independent and rely on payment method code.

 Has to be backported to 13.0

### DIFF
--- a/htdocs/takepos/index.php
+++ b/htdocs/takepos/index.php
@@ -978,36 +978,39 @@ if (empty($conf->global->TAKEPOS_HIDE_HEAD_BAR)) {
 <?php
 
 // TakePOS setup check
-$sql = "SELECT code, libelle FROM ".MAIN_DB_PREFIX."c_paiement";
-$sql .= " WHERE entity IN (".getEntity('c_paiement').")";
-$sql .= " AND active = 1";
-$sql .= " ORDER BY libelle";
+if ( isset( $_SESSION["takeposterminal"] ) && $_SESSION["takeposterminal"] ) {
+	$sql = "SELECT code, libelle FROM " . MAIN_DB_PREFIX . "c_paiement";
+	$sql .= " WHERE entity IN (" . getEntity( 'c_paiement' ) . ")";
+	$sql .= " AND active = 1";
+	$sql .= " ORDER BY libelle";
 
-$resql = $db->query($sql);
-$paiementsModes = array();
-if ($resql) {
-	while ($obj = $db->fetch_object($resql)) {
-		$paycode = $obj->code;
-		if ($paycode == 'LIQ') {
-			$paycode = 'CASH';
-		}
-		if ($paycode == 'CHQ') {
-			$paycode = 'CHEQUE';
-		}
+	$resql          = $db->query( $sql );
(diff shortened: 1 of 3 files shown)


htdocs/takepos/index.php
```diff
@@ -978,36 +978,39 @@ function WeighingScale(){
...
+			}
 		}
+	}
 
-		$constantforkey = "CASHDESK_ID_BANKACCOUNT_".$paycode.$_SESSION["takeposterminal"];
-		//var_dump($constantforkey.' '.$conf->global->$constantforkey);
-		if (!empty($conf->global->$constantforkey) && $conf->global->$constantforkey > 0) {
-			array_push($paiementsModes, $obj);
-		}
+	if (empty($paiementsModes)) {
```
eldy, line 1007: I think this test must be moved outside the parent "if".
So it is still able to report error when $_SESSION["takeposterminal"] is empty, isn't it ?

loxK: I don't think so. Let me explain.
When you first come to TakePos, $_SESSION["takeposterminal"] isn't set, and you get the terminal selector that, when selected, set the terminal in session. Without that change, an error shows when the terminal selector is shown.

PR #18644, opened 2021-09-07
TakePos - fast product search

loxK:
# New Greatly enhances product search

- Add temporization when searching products: Adds a temporization after each typed letter, giving the user 700ms to type next letter before sending the search request. It prevents hammering the server for each typed letter and "results jumping" effect, which is error-prone.

- Set proper image directly when searching products: Bypasses genimg/index.php that either redirects to viewimage.php or to nophoto.png. That was making lots of request to the server and excessive load on the database server, that queues them slowing down the UI.

On my test environment, it gives nearly instantaneous search results.


eldy:
Only blocking bug fixes are allowed into branch already released

PR #18648, opened 2021-09-07
NEW : mass action delete, no more break if at least one object has child

atm-gauthier:


htdocs/core/actions_massactions.inc.php
```diff
@@ -1328,15 +1328,18 @@
...
-		$db->commit();
-	} else {
-		$db->rollback();
+	// Message for elements well deleted
+	if ($nbok > 1) {
+		setEventMessages($langs->trans("RecordsDeleted", $nbok), null, 'mesgs');
+	} elseif (!empty($nbok)) {
+		setEventMessages($langs->trans("RecordsDeleted", '1'), null, 'mesgs');
+	}
+	$db->commit();
```
eldy, line 1337: If an error occured during a massaction, we must not make a commit. For example a trigger or an external module may have failed in doing a complementary action like deleting into a synchronized system, in such case, we must not validate the deletion into dolibarr.

atm-gauthier:
Hi @eldy, i understand this this coment, but if deleteCommon returns 0, it means that there is at least one object linked, and this is not an error but a warning for me.
Then in my last commit i did 3 case, one for at least a deleteCommon result < 0 (then we break and rollback as well as already done), one one for deleteCommon result == 0 (then we set warning event message and test other objects), and one for deleteCommon result > 0 (we delete object and test other objects, and we commit we no error, or rollback if at least one error).
What do you think about this new version ?

eldy:
Looks good now.

htdocs/core/actions_massactions.inc.php
```diff
@@ -1335,8 +1335,8 @@
 		// Message for elements well deleted
 		if ($nbok > 1) {
 			setEventMessages($langs->trans("RecordsDeleted", $nbok), null, 'mesgs');
-		} elseif (!empty($nbok)) {
-			setEventMessages($langs->trans("RecordsDeleted", '1'), null, 'mesgs');
+		} elseif ($nbok == 1) {
+			setEventMessages($langs->trans("RecordDeleted"), null, 'mesgs');
 		}
```
atm-gauthier, line 1340: @eldy, thank you for merge.

But i used "Record**s**Deleted" trans for only 1 element too voluntarily.
Because if we get 2 messages, one warning for linked object, and one for success deletions, i think it's more clear to get something like :

image

than something like that : 

image

What's your opinion about that ?

PR #18672, opened 2021-09-10
NEW Survey - Add a clickable link in email send

aspangaro:


htdocs/opensurvey/fonctions.php
```diff
@@ -198,6 +198,7 @@ function getUrlSondage($id, $admin = false)
 	} else {
 		$url = get_server_name().'/public/studs.php?sondage='.$id;
 	}
+	$url = '<a href="'.$url.'">'.$url.'</a>';
```
eldy, line 201: I think the getUrlSondage must continue to return the brut url. Formatting is bette when done by the one that get the return of this method.
Only the caller know how must be forged the href, with css or target to add.

aspangaro: Done @eldy

PR #18694, opened 2021-09-13
Fix  ENDOFMONTH for payment terms

atm-greg:
# Fix End of month payment terms must be applied first
When we select a payment term like 45DENDOFMONTH, the system was calculating the day shift first and then the ENDOFMONTH clause.
E.x. : 
facturation date : 2021-09-13
+45 days : 2021-10-28
ENDOFMONTH : 2021-10-31

To be correct the date must be calculated in a different order
facturation date : 2021-09-13
ENDOFMONTH : 2021-09-30
+45 days : 2021-11-14 => good value

### DIFF
--- a/htdocs/core/class/commoninvoice.class.php
+++ b/htdocs/core/class/commoninvoice.class.php
@@ -649,23 +649,13 @@ abstract class CommonInvoice extends CommonObject
 		// 1 : application of the "end of the month" rule
 		elseif ($cdr_type == 1)
 		{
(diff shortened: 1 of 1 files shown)


eldy:
I don't think problem is in the code but into th dictionary table setup.
Currently the rule ENDOFMONTH works like this:
date +delta, then end of month, then +offset.
The rule you present means delta = 0 and offset = 45
Currently into setup we have delta = 45 and offset = 0

But the rule is able to manage the 2 cases. By default we just had setup first case.

eldy:

PR #18697, opened 2021-09-13
FIX : test must be === and not == (this correction has already been d…

atm-gauthier:
…one on commit c35c15e but disapperead...)


eldy:
Thanks duplicate PR received at sametime

PR #18702, opened 2021-09-13
Close #18692 : Possibility to move directories in ecm

### LINKED ISSUE #18692: Can move a directory into ecm module
Go on module Document, into "manual tree".
Edit a directory to reach the page
ecm/dir_card.php?action=edit&module=ecm&section=...
Replace the input field that allows to change the name of the field with 2 lines:
Parent directory: here a combi to select parent dir
Dir name: here the input field to change the name of directory.

Action on save should be similar except that the move of old directory into new one isbnot always inside same dir.

So we will be able to organize data by changing dir position.

Hystepik:
Possibility to move directories in ecm module in edit mode

### DIFF
--- a/htdocs/ecm/class/htmlecm.form.class.php
+++ b/htdocs/ecm/class/htmlecm.form.class.php
@@ -55,10 +55,10 @@ class FormEcm
 	 *  @param	int		$selected    		Id of preselected section
 	 *  @param  string	$select_name		Name of HTML select component
 	 *  @param	string	$module				Module ('ecm', 'medias', ...)
-	 *  @param	array	$ignore_id			Arroy of id to ignore
+	 *  @param	array	$ids_to_ignore		Array of id to ignore
 	 *  @return	string						String with HTML select
 	 */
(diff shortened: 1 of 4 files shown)


htdocs/ecm/class/htmlecm.form.class.php
```diff
@@ -55,16 +55,20 @@ public function __construct($db)
 	 *  @param	int		$selected    		Id of preselected section
 	 *  @param  string	$select_name		Name of HTML select component
 	 *  @param	string	$module				Module ('ecm', 'medias', ...)
+	 *  @param	array	$ignore_id			Arroy of id to ignore
 	 *  @return	string						String with HTML select
 	 */
-	public function selectAllSections($selected = 0, $select_name = '', $module = 'ecm')
+	public function selectAllSections($selected = 0, $select_name = '', $module = 'ecm', $ignore_id = array())
```
eldy, line 61: Can you rename the var 
$ignore_id
into
$ids_to_ignore

Because var ending with _id are used for a numeric ID so this create a confusion because here it is an array of several.

Hystepik: It's done

PR #18723, opened 2021-09-15
Explicit date_creation instead of date

PyroShape:


htdocs/fourn/class/fournisseur.commande.class.php
```diff
@@ -111,7 +111,7 @@ class CommandeFournisseur extends CommonOrder
 
 	public $socid;
 	public $fourn_id;
-	public $date;
+	public $date_creation;
```
eldy, line 114: We must keep the field $date (date of order) that may be different than date_creation.
I suggest to add public $date_creation instead of just replacing date.

htdocs/fourn/class/fournisseur.commande.class.php
```diff
@@ -375,7 +375,7 @@ public function fetch($id, $ref = '')
 			$this->total_localtax1		= $obj->localtax1;
 			$this->total_localtax2		= $obj->localtax2;
 			$this->total_ttc			= $obj->total_ttc;
-			$this->date = $this->db->jdate($obj->date_creation);
+			$this->date_creation = $this->db->jdate($obj->date_creation);
```
eldy, line 378: Keep also $this->date even if value will be same than date_creation.

htdocs/fourn/class/fournisseur.commande.class.php
```diff
@@ -1282,7 +1282,7 @@ public function create($user, $notrigger = 0)
 		$now = dol_now();
 
 		// set tmp vars
-		$date = ($this->date_commande ? $this->date_commande : $this->date); // in case of date is set
+		$date = ($this->date_commande ? $this->date_commande : $this->date_creation); // in case of date is set
```
eldy, line 1285: Should stay ->date

htdocs/fourn/class/fournisseur.commande.class.php
```diff
@@ -2847,9 +2847,9 @@ public function initAsSpecimen()
 		$this->ref = 'SPECIMEN';
 		$this->specimen = 1;
 		$this->socid = 1;
-		$this->date = $now;
+		$this->date_creation = $now;
```
eldy, line 2850: Keep also ->date, date_creation must be added not replacing

PyroShape:
Hello Eldy.
I don't understand why do you want to keep $date. It's in fact the $date_creation of the order, the date of order is $date_commande (or am I missing something ?)

eldy:
date_commande is the deprecated name. The name that we must use now to get the date of order is  ->date  (were are removing name in french and we also aligned all date of object -proposal,invoice,order- with name "date").
So we should now have the ->date property disappear (this is the real property name that must be used as date of order).

PyroShape:
Ok so if a date_commande exist, date should be the date_commande, instead the date is the date_creation.
Is that correct ?

eldy:
> Ok so if a date_commande exist, date should be the date_commande, instead the date is the date_creation. Is that correct ?

I merged PR. But note that date = date_command should always be set.

PR #18724, opened 2021-09-15
Fix ldap passw bug with activedirectory

ibuiv:
Fix to avoid LDAP Error: 53 (Unwilling to perform) on Samba4 AD

# Fix Bug on LDAP Samba4 AD when creating or updating password
Format was wrong creating LDAP Error: 53 (Unwilling to perform)


ibuiv:
My mastery of Git and GitHub is not perfect. 
My purpose was to fix this bug in 14.0
But, if I switch Base Branch, this PR includes 1400+ Files changed ... 
My bad

eldy:
Reason is that you must create another branch on your dev env from branch 14.0 instead of develop and make the change on this new env. When pushing your PR, it will be targeted into 14.
Don't worry, i will cherry-pick the commit into v14

PR #18729, opened 2021-09-16
NEW Expose bank account holder, label and number

esion-net:
NEW Expose bank account holder, label and number


FHenry:
@esion-net can only be PR on develop branch even if it seems "normal to have it" or "little change", all new feature are only in develop branch.

esion-net:
I know, but it really should have been implemented earlier as this fields are vital for a proper invoice. I pushed it to all branches from 12.0 upwards (we are using 12.0). It also already got merged in 13.0, 14.0 and develop. Please reconsider. 
#18730  #18732 #18733

PR #18737, opened 2021-09-16
FIX: task time: can't filter by user with pgsql + show error message

marc-dll:
natural_search() can't work on interger fields in pgsql because of mixing types. In this case, it's not needed anyway.


htdocs/projet/tasks/time.php
```diff
@@ -1058,7 +1058,7 @@
 		if ($search_note) $sql .= natural_search('t.note', $search_note);
 		if ($search_task_ref) $sql .= natural_search('pt.ref', $search_task_ref);
 		if ($search_task_label) $sql .= natural_search('pt.label', $search_task_label);
-		if ($search_user > 0) $sql .= natural_search('t.fk_user', $search_user);
+		if ($search_user > 0) $sql .= ' AND t.fk_user = ' . intval($search_user);
```
eldy, line 1061: Can you try instead 
natural_search('t.fk_user', $search_user, 2);

marc-dll: Done

Issue #18741, opened 2021-09-17
Barcode missing in product_lot


lmag:
# Feature Request
The bar code and the QR code are missing which should be linked to the lot numbers and serial numbers
product/stock/productlot_card.php

PR #18742, opened 2021-09-17
NEW: Add code codebar column on serial/lot structure

FHenry:
Start dev on #18741

### DIFF
--- a/htdocs/install/mysql/migration/14.0.0-15.0.0.sql
+++ b/htdocs/install/mysql/migration/14.0.0-15.0.0.sql
@@ -91,5 +91,8 @@ ALTER TABLE llx_mrp_mo ADD COLUMN mrptype integer DEFAULT 0;
 
 DELETE FROM llx_menu WHERE type = 'top' AND module = 'cashdesk' AND mainmenu = 'cashdesk';
 
-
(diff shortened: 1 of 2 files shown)


BebZ:
Hello, 

If you add a barcode on serial/lot will it be only for internal/which purpose ? (For Manufacturing, this is really useful i understand)

But for logisitcs : there are many places where barcodes should/would be used. 

For suppliers prices : there is a recent field called "packaging" / "emballage" in french that would permit to register different levels of barcodes : we can record barcodes of the units / boxes / cartons / pallets and set with "emballage" how many selling units are inside a box / carton / pallet...

For reception/dispatch : we can record barcodes of the units / boxes / cartons / pallets
For picking : we can scan the barcode of units / boxes / cartons / pallets
For "checking" and shipping : we also scan the same barcodes.

Those are the "logistics" barcodes. 

Batch barcode is "what" : a sub-units barcode that enters the "logisitcs" ones 
or an independent barcode that is not to manage any "logistic" ?

BebZ:
?? 

Batch barcode is "what" : a sub-units barcode that enters the "logisitcs" ones
or an independent barcode that is not to manage any "logistic" ?

PR #18756, opened 2021-09-18
Show inventory in table toConsume

mschamp:
# New: add inventory in table toConsume
Add a column with the reel inventory to the production screen of manufacturing orders.
This gives the user an indication if production is possible.
Also add a warning if the inventory is to low.


htdocs/mrp/mo_production.php
```diff
@@ -149,19 +149,14 @@
...
 		$moline->fk_mo = $object->id;
 		$moline->qty = GETPOST('qtytoadd', 'int'); ;
 		$moline->fk_product = GETPOST('productidtoadd', 'int');
-		if (GETPOST('addconsumelinebutton')) {
-			$moline->role = 'toconsume';
-		} else {
-			$moline->role = 'toproduce';
-		}
+		$moline->role = 'toconsume';
 		$moline->origin_type = 'free'; // free consume line
```
hregis: Can you fix this please

$line->qty-$alreadyconsumed

Thanks

mschamp: Excuse me, I'm quite new to git and php.
I don't really understand what you mean that I have to fix.
I thought the '$line->qty' is the total number of products required and the '$alreadyconsumed' the number of products you already used in a partial production. The difference between those 2 is what is still needed to produce the other products and finish the manufacturing order.
If you can explain a little more what the problem is, I will fix it.

PR #18779, opened 2021-09-21
FIX GETPOST type check for the note of the discount on thridparty card

kkhelifa-opendsi:
FIX GETPOST type check for the note of the discount on thridparty card

### DIFF
--- a/htdocs/comm/remise.php
+++ b/htdocs/comm/remise.php
@@ -56,9 +56,9 @@ if (GETPOST('action', 'aZ09') == 'setremise') {
 	$discount_type = GETPOST('discount_type', 'int');
 
 	if (!empty($discount_type)) {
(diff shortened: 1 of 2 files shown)


htdocs/comm/remise.php
```diff
@@ -56,9 +56,9 @@
 	$discount_type = GETPOST('discount_type', 'int');
 
 	if (!empty($discount_type)) {
-		$result = $object->set_remise_supplier(price2num(GETPOST("remise")), GETPOST("note", "alphanohtml"), $user);
+		$result = $object->set_remise_supplier(price2num(GETPOST("remise")), GETPOST("note", "san_alpha"), $user);
```
eldy, line 59: The field note here seems to be a "one line text". So recommended value is "alphanohtml".
Why do you need "san_alpha" ?

PR #18781, opened 2021-09-21
NEW: minimum php version is now 7.3

FHenry:
As preview version of PHP is not supported anymore https://www.php.net/supported-versions.php we need to set minimum php version to 7.3

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
 		"vendor-dir" : "htdocs/includes"
 	},
 	"require" : {
-		"php" : ">=5.6.0",
+		"php" : ">=7.3.0",
 		"ext-curl" : "*",
 		"ckeditor/ckeditor" : "4.12.1",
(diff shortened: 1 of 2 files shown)


eldy:
50% of our users. We can't betray them.

FHenry:
https://p.datadoghq.eu/sb/wtaqftsy891utpai-a1a17be3de5b82189abab454e5c90bfb?tv_mode=true

50% of the users ? Are you sure  ?
image

eldy:
This is stat of "new" installation. I was speaking about still live installations.
Just take a look at new installation that were done just in june 2020 (so users that are live today if we imagine an ERP is kept at least 1 year):
image
We are around 63%. 
And we must also include all users that was installed before (that is much higher than new users that did the installation in last 12 month, but we don't have statistics for installation before june 2020). It is difficult to say how many has stopped to use it during the year so this is why i rounded to 50% but it is probably more than that.

PR #18783, opened 2021-09-21
NEW : add showdocument hook

fr69400:
May now overwrite showdocument or change option before call this method


htdocs/adherents/card.php
```diff
@@ -2018,7 +2018,11 @@ function initfieldrequired() {
 			$genallowed = $user->rights->adherent->lire;
 			$delallowed = $user->rights->adherent->creer;
 
-			print $formfile->showdocuments('member', $filename, $filedir, $urlsource, $genallowed, $delallowed, $object->model_pdf, 1, 0, 0, 28, 0, '', '', '', (empty($object->default_lang) ? '' : $object->default_lang), '', $object);
+			$parameters = array();
+			$reshook = $hookmanager->executeHooks('showdocuments', $parameters, $object, $action);
```
eldy, line 2022: Why not including the hook inside the showdocument instead of everywhere before the call ?

PR #18784, opened 2021-09-21
FIX check if $categories is not empty

hregis:


htdocs/core/actions_addupdatedelete.inc.php
```diff
@@ -236,7 +236,9 @@
 
 		if ($conf->categorie->enabled) {
 			$categories = GETPOST('categories', 'array');
-			$object->setCategories($categories);
+			if (!empty($categories)) {
```
eldy, line 239: If you want to remove all categories, the array will be empty. But removal will fails if we don't call the setCategories to save this.

PR #18795, opened 2021-09-22
NEW #18770 : Can enter the buying price (for Weighted Average Price update) on lines during the reception step

### LINKED ISSUE #18770: Can enter the buying price on line of each reception during reception process.
When creating a reception (reception module, see screenshot), if option for "Stock increment" is set with option
$conf->global->STOCK_CALCULATE_ON_RECEPTION or $conf->global->STOCK_CALCULATE_ON_RECEPTION_CLOSE

we must have a field "buying price" on each line (between "qty to dispatch" and "warehouse").

Capture du 2021-09-20 20-18-12

This field will be pre-filled using the amount of the purchase order. User should be able to edit it to another value. 
When saving/creating the reception, data will be stored into field "cost_price" into table llx_commande_fournisseur_dispatch (so the table to store lines of receptions, field must be added).

Then when we make the stock change (on reception validation if STOCK_CALCULATE_ON_RECEPTION is on or on reception closing when STOCK_CALCULATE_ON_RECEPTION_CLOSE is on), we must use this price when we call the 
$mouvS->reception
instead of the price of line in purchase order.

So the Weighted Average Price (updated when calling $mouvS->reception) may be updated using a different amount than the purchase order price, at user choice.

Hystepik:
Can enter the buying price on line of each reception during reception process
image


htdocs/reception/card.php
```diff
@@ -364,8 +365,11 @@
 					$sellby = GETPOST($sellby, 'alpha');
 					$eatbydate = str_replace('/', '-', $eatby);
 					$sellbydate = str_replace('/', '-', $sellby);
-
-					$ret = $object->addline($entrepot_id, GETPOST($idl, 'int'), GETPOST($qty, 'int'), $array_options[$i], GETPOST($comment, 'alpha'), strtotime($eatbydate), strtotime($sellbydate), GETPOST($batch, 'alpha'));
+					if (!empty($conf->global->STOCK_CALCULATE_ON_RECEPTION) || !empty($conf->global->STOCK_CALCULATE_ON_RECEPTION_CLOSE)) {
+						$ret = $object->addline($entrepot_id, GETPOST($idl, 'int'), GETPOST($qty, 'int'), $array_options[$i], GETPOST($comment, 'alpha'), strtotime($eatbydate), strtotime($sellbydate), GETPOST($batch, 'alpha'), GETPOST($cost_price, 'double'));
```
eldy, line 369: GETPOST($cost_price, 'double')  is a string enter by end user (using its language format for decimal and thousand separator).
To be sure it is a real amount (expected by function), you must include  
price2num(GETPOST($cost_price, 'double'), 'MU')

htdocs/reception/class/reception.class.php
```diff
@@ -562,6 +562,9 @@ public function valid($user, $notrigger = 0)
 			$sql = "SELECT cd.fk_product, cd.subprice,";
 			$sql .= " ed.rowid, ed.qty, ed.fk_entrepot,";
 			$sql .= " ed.eatby, ed.sellby, ed.batch";
+			if (!empty($conf->global->STOCK_CALCULATE_ON_RECEPTION || $conf->global->STOCK_CALCULATE_ON_RECEPTION_CLOSE)) {
```
eldy, line 565: Because here it is a select, and the field already exists, to make code simpler, there is no need for the if.
You can always have the field into list of return value. Later, we will use the condition to exploit or not the data.

PR #18797, opened 2021-09-22
FIX: Warehouse project fk

altairis-noe:
We should not be authorized to delete projects if warehouses are still linked to them. This constraint fixes that.


eldy:
The project on a warehouse is not a mandatory information.
So if we try to delete a project and the warehouse is linked to this project, the warehouse must just become linked to no project.

PR #18802, opened 2021-09-22
Close: #18801 Jan-30 + 1m IS NOT Mar-2

### LINKED ISSUE #18801: Change the behavior of PHP over dateinterval when the result of this function is Feb 29 (in non-leap years), 30 or Feb 31 (php returns March 1, 2 or 3 respectively)
# Feature Request
Give the option to change the behavior of PHP with dateinterval when the result of this function is Feb 29 (non-leap years), 30 or Feb 31 (php returns March 1, 2 or 3 respectively)

## Use case
If $ time = 2021-01-(29,30 or 31)

We want dol_plus_duree($time, 1, 'm') to result in 2021-02-28 instead of 2021-03-(01, 02 or 03)

More info: https://bugs.php.net/bug.php?id=74013

## Suggested implementation
add a CONST and modify the dol_time_plus_duree function

javieralapps4up:
Close: #18801 related to avoid this behaviour https://bugs.php.net/bug.php?id=74013

# Close #18801 related to avoid this behaviour https://bugs.php.net/bug.php?id=74013

I think we need the option that dol_time_plus_duree (+ 1m) returns 2021-02-28 instead of 2021-03-02 (example with 2021-01-30)

### DIFF
--- a/htdocs/core/lib/date.lib.php
+++ b/htdocs/core/lib/date.lib.php
@@ -181,9 +181,9 @@ function dol_time_plus_duree($time, $duration_value, $duration_unit)
 		$newtimetotalmonths = (($newtimeyear * 12) + $newtimemonth);
 
 		if ($monthsexpected < $newtimetotalmonths) {
-			$newtimehours = dol_print_date($newtime, '%m');
-			$newtimemins = dol_print_date($newtime, '%m');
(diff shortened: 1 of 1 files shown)


javieralapps4up (commented):
Lines 184, 185 and 186 with mistakes, must be %h %i and %s. I edited the file, but i dont see the changes

PR #18810, opened 2021-09-24
FIX Accountancy - Missing specific filename for export on format FEC2

aspangaro:


htdocs/accountancy/tpl/export_journal.tpl.php
```diff
@@ -38,7 +38,9 @@
 include_once DOL_DOCUMENT_ROOT.'/accountancy/class/accountancyexport.class.php';
 $accountancyexport = new AccountancyExport($db);
 
-if ($accountancyexport->getFormatCode($formatexportset) == $accountancyexport::$EXPORT_TYPE_FEC && $type_export == "general_ledger") { // Specific filename for FEC model export into the general ledger
+// Specific filename for FEC model export into the general ledger
+if ($accountancyexport->getFormatCode($formatexportset) == ($accountancyexport::$EXPORT_TYPE_FEC || $accountancyexport::$EXPORT_TYPE_FEC2)
```
eldy, line 42: Surely a bug here 
$accountancyexport::$EXPORT_TYPE_FEC || $accountancyexport::$EXPORT_TYPE_FEC2
is a boolean.
$accountancyexport->getFormatCode($formatexportset) is a string

aspangaro: You're right ! Maybe for that what functionality doesn't work for a long while.

I repare it simply for FEC, FEC2, CIEL (with constant ACCOUNTING_EXPORT_XIMPORT_FORCE_FILENAME at 1)

Issue #18820, opened 2021-09-25
Import Export Accounting Level 3 values on ThirdParty


josett225:
# Instructions
*This is a template to help you report good issues. You may use [Github Markdown](https://help.github.com/articles/getting-started-with-writing-and-formatting-on-github/) syntax to format your issue report.*
*Please:*
- *replace the bracket enclosed texts with meaningful information*
- *remove any unused sub-section*

# Feature Request
Add in ThirdParty import and Export the 2 Accounting Level 3 values (as ExtraFields behavior) when activating ACCOUNTANCY_USE_PRODUCT_ACCOUNT_ON_THIRDPARTY via setup 

## Use case
This feature could be awesome when migrating from another ERP to Dolibarr or when activating the level 3 accounting feature a mass action is required. This is helpful to avoid to manually go on each thirdparty to add the right Buy/Sale Accounting number.

## Suggested implementation
[*Verbose description*]

## Suggested steps
[*List of tasks to achieve goal*]

PR #18821, opened 2021-09-25
NEW #18820 Accountancy - Add product account in import/export of thirdparty (level 3)

aspangaro:
image


josett225:
Hi @aspangaro 

Many thanks for the quick update of this feature request. Do you know if this could be back port to version 14?

Regards

Jose

eldy:
As it is a feature still not officially available for every user (still experimental and protected by a hidden constant) and the code change is inside code that is protected by this hidden constant, it has no effect on official code, so it is ok to cherry-pick the commit into v14 since it has no effect. I did the cherry-pick.

aspangaro:
Thanks @eldy

PR #18850, opened 2021-09-29
New : integration gpec, source code

atm-gauthier:


atm-gauthier:
Hi @eldy, i also did hrm_position > hrm_job_user adaptations on source code but i didn't rename concerned object and files because it would be a very long job.
Is it ok for you ?

htdocs/hrm/lib/hrm.lib.php
```diff
@@ -0,0 +1,69 @@
...
+ * \ingroup hr
+ * \brief   Library files with common functions for Workstation
+ */
+
+/**
+ * Prepare admin pages header
+ *
+ * @return array
+ */
+function hrmAdminPrepareHead()
```
eldy, line 33: We are loosing the tab that link to the existing page to setup the "department" of a company:
"admin_establishment.php" ?

atm-gauthier: Hi @eldy, I will have a look at this on Monday. Have a nice week-end.

atm-gauthier: Hi @eldy, i just restore link to hrm establihsment tab

atm-gauthier:
Hi @eldy, i have a problem with travis : 

I get this feedback : Found non escaped string in building of a sql request (case 3) in hrm/class/evaluation.class.php: $sql .= ' AND ('.implode(' '.$filtermode.' ', $sqlwhere).')';

But if i add $this->db->escape on value i get an error because for my request in skill_tab.php i need to use fetch_all with objecttype='job'

AND for postgresql compatibility, i needed to use simple quotes and not double quotes.

Then if i add $this->db->escape on $sqlwhere i will get something like : 

image

and it breaks my query...

What can i do ?

It seems that some travis feedback are not up to date with what modulebuilder creates in class

eldy:
Replace  
$sql .= ' AND ('.implode(' '.$filtermode.' ', $sqlwhere).')';
with
$sql .= " AND (".implode(" ".$filtermode." ", $sqlwhere).")";

To forge sql request we must use " for php string and ' is used for sql string. To allow detection of sql injection we must follow strictly this rule.

atm-gauthier:
@eldy i'm working to correct this and the same problem in a lot of file.
But i think module builder should be corrected too because a lot of files i'm fixing come from module builder use.

atm-gauthier:
@eldy you have a parse error in develop : 

image

It is blocking for travis build

atm-gauthier:
i did a pr https://github.com/Dolibarr/dolibarr/pull/18918

atm-gauthier:
Hello @eldy, now all checks have passed for this PR.
You will see that besides hrm files, i had to add some modifications on several standard files, to help you i list them here :
htdocs/core/lib/company.lib.php
htdocs/user/class/user.class.php

If you have some questions don't hesitate

PR #18862, opened 2021-09-30
NEW: expense report: show link to already existing overlapping expense report

marc-dll:


htdocs/expensereport/class/expensereport.class.php
```diff
@@ -2195,9 +2195,10 @@ public function deleteline($rowid, $fuser = '')
 	 * @param   User       $fuser          User
 	 * @param   integer    $date_debut     Start date
 	 * @param   integer    $date_fin       End date
+	 * @param   bool       $return_id      True to return ID of existing expense report
 	 * @return  int                        <0 if KO, >0 if OK
 	 */
-	public function periode_existe($fuser, $date_debut, $date_fin)
+	public function periode_existe($fuser, $date_debut, $date_fin, $return_id = false)
```
eldy, line 2201: I don't think we need to add a param.
Returning the id does not break the rule > 0 is ok so we can take the risk to return the id all the times by default.

marc-dll: @eldy Done ;)

PR #18863, opened 2021-09-30
NEW: expense reports: make project field required on lines

marc-dll:


htdocs/core/modules/modProjet.class.php
```diff
@@ -330,6 +330,10 @@ public function init($options = '')
 	{
 		global $conf, $langs;
 
+		if (dolibarr_del_const($this->db, 'EXPENSEREPORT_PROJECT_IS_REQUIRED', $conf->entity) < 0) {
```
eldy, line 333: What is goal of this ?
When disabling and enabling a module, we should retrieve the setup before disabling.

marc-dll: @eldy I guess I was just sleep deprived :/ Fixed

PR #18865, opened 2021-09-30
Agenda events visibility

bomuux:
# Fix
All users should be able to filter event types and to see auto events in agenda views. Currently read-only agenda events users can't filter, and don't see auto events.


eldy:
The variable $canedit, means can change the filters according to documentation in top of method
$canedit    Can edit filter fields
So this breaks the future. The fix should probably be done in the calling method that must set this parameter to 1

bomuux:
ok, understood. A problem remains to me, i'll explain it in a new issue. I close this one.

PR #18866, opened 2021-09-30
FIX: postgresql compatibility with camelcase fields

marc-dll:
This was partly fixed in PR #18094 where I explained why that happend, but it never made it to the develop then...

The rest of it is a regression introduced in commit https://github.com/scopen-coop/dolibarr_customer/commit/82f74ceb38d2ed8ea673df74eeca7de6977ac51e

Tested both in MySQL and PostgreSQL

### DIFF
--- a/htdocs/admin/dict.php
+++ b/htdocs/admin/dict.php
@@ -223,7 +223,7 @@ $tabsql[24] = "SELECT rowid   as rowid, code, label, active FROM ".MAIN_DB_PREFI
 $tabsql[25] = "SELECT rowid   as rowid, code, label, active, module FROM ".MAIN_DB_PREFIX."c_type_container as t WHERE t.entity IN (".getEntity('c_type_container').")";
 //$tabsql[26]= "SELECT rowid   as rowid, code, label, short_label, active FROM ".MAIN_DB_PREFIX."c_units";
 $tabsql[27] = "SELECT id      as rowid, code, libelle, picto, active FROM ".MAIN_DB_PREFIX."c_stcomm";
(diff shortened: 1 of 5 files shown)


htdocs/admin/dict.php
```diff
@@ -223,7 +223,7 @@
 $tabsql[25] = "SELECT rowid   as rowid, code, label, active, module FROM ".MAIN_DB_PREFIX."c_type_container as t WHERE t.entity IN (".getEntity('c_type_container').")";
 //$tabsql[26]= "SELECT rowid   as rowid, code, label, short_label, active FROM ".MAIN_DB_PREFIX."c_units";
 $tabsql[27] = "SELECT id      as rowid, code, libelle, picto, active FROM ".MAIN_DB_PREFIX."c_stcomm";
-$tabsql[28] = "SELECT h.rowid as rowid, h.code, h.label, h.affect, h.delay, h.newByMonth, h.fk_country as country_id, c.code as country_code, c.label as country, h.active FROM ".MAIN_DB_PREFIX."c_holiday_types as h LEFT JOIN ".MAIN_DB_PREFIX."c_country as c ON h.fk_country=c.rowid";
```
eldy, line 226: In the sql instruction to create the table, the field is name "newByMonth" in v14 so we must named it the same way in code.
An enhancement can be done to rename it into "newbymonth" (all columns should be lower case), but this means we must also rename the field into database. So this must be done into v14.

marc-dll: @eldy Done, it was the only column with uppercase characters in its name, we should be ok from then on !

Issue #18688, opened 2021-09-12
Canceled purchase orders should have a own list on overview (Projects)


braito4:
# Feature Request
Normally in purchase orders on overview (Project), we have the list of purchase orders that we ordered, and we use de total price to check if all the payments have been sent.
When you cancel a order, the order stills in the list (we still need this in order to kwon if a supplier have been contact), but in this list is not useful.

## Suggested implementation
In Overview (Project) create a list that If it's exist a cancel order/facture... a new list appears whit this orders ("List of cancel purchase orders related to the project"), instead the the approved ones

PR #18867, opened 2021-09-30
NEW Added a list of canceled purchase orders

polkiko:
Added a new list in project overview "List of canceled purchase orders realated to the project" that includes all the canceled purchase orders, not showing them on the list of purchase orders that already exists.

Implements #18688

### DIFF
--- a/htdocs/core/tpl/card_presend.tpl.php
+++ b/htdocs/core/tpl/card_presend.tpl.php
@@ -207,7 +207,9 @@ if ($action == 'presend') {
 	}
 
 	// Make substitution in email content
-	$formmail->setSubstitFromObject($object, $langs);
+	if ($object) {
+		$formmail->setSubstitFromObject($object, $langs);
+	}
 	$substitutionarray = getCommonSubstitutionArray($outputlangs, 0, $arrayoffamiliestoexclude, $object);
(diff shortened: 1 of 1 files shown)


eldy:
A better enhancement is to modify the existing list to include also canceled orders to avoid to have too many list for same object.

polkiko:
> A better enhancement is to modify the existing list to include also canceled orders to avoid to have too many list for same object.

I agree, there are many tables. A possible solution is to show a show/hide button with canceled purchase orders. This is what I propose in my commit.

image
image

Another solution that would be more convenient, would be to apply filters on the columns of the tables, but this is a bigger improvement.

eldy:
What is hurting me is to have a second table for this, i would prefer to have the list into the same table, clicking on the "Hide/show canceled" will use js to hide/show the lines of canceled orders. 
We can imagine a lot of other case of filters like this and we can't imagine to have one array for each filter type.

polkiko:
> What is hurting me is to have a second table for this, i would prefer to have the list into the same table, clicking on the "Hide/show canceled" will use js to hide/show the lines of canceled orders. We can imagine a lot of other case of filters like this and we can't imagine to have one array for each filter type.

Once again, I agree with you. Now the canceled orders do not add up to the total amount, and can be seen by clicking on the "show/hide" button:
image
image

It would be necessary to finish the code correctly (js/css)

PR #18872, opened 2021-10-01
Add trigger 'MRP_MO_DELETE'

MarBuA:
Hi, I add a new trigger for MO module.
If you thing it may be interesting I would appreciate you incorporating it into de original Dolibarr project.
Thank you.

# New: Add trigger 'MRP_MO_DELETE'
Add a new trigger for a MO Delete action.

### DIFF
--- a/htdocs/mrp/class/mo.class.php
+++ b/htdocs/mrp/class/mo.class.php
@@ -735,7 +735,10 @@ class Mo extends CommonObject
 	 */
 	public function delete(User $user, $notrigger = false)
 	{
+		$result = $this->call_trigger('MRP_MO_DELETE', $user);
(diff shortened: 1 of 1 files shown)


htdocs/mrp/class/mo.class.php
```diff
@@ -735,7 +735,10 @@ public function updateProduction(User $user, $notrigger = true)
 	 */
 	public function delete(User $user, $notrigger = false)
 	{
+		$result = $this->call_trigger('MRP_MO_DELETE', $user);
```
eldy, line 738: A trigger is already automatically included into the deleteCommon() method.
Name is MO_DELETE

PR #18901, opened 2021-10-04
Fix #18890 Fix disabled button if sum real qty is 0

### LINKED ISSUE #18890: Cannot book inventory counting where all lines are zero quantity
# Bug
We have virtual warehouses for transit, quarantine, etc. When performing counting, it's possible we count down discrepancies to zero (i.e. the warehouse is empty). If we try to do this using the inventory counting, we cannot use the "generate movements and close" button if all of the "real qty" values are zero. 

## Environment
- **Version**: 14.0.2
- **OS**: Debian 4.19
- **Web server**: Apache
- **PHP**: 7.3.29-1
- **Database**: MySQL 8.0.23
- **URL(s)**: /product/inventory/list.php?leftmenu=stock_inventories

It should be possible to count down all quantities to 0 (= empty warehouse) when all "real qty" values have been explicitly set to zero. However, when setting values to zero, the system evaluates this the same as having nothing entered in these fields at all.

 1. Create a counting journal for a particular warehouse with stock (negative or otherwise)
 2. Set all "real qty" fields to 0
 3. Save
 4. Result: "Generate movements and close" button remains unavailable

xgerhard:
# Fix #18890 Fix disabled button if sum real qty is 0
Instead of counting the sum of the 'real qty' input, only disable the button if no input has been given, or if all input is the same as the expected qty.

'$object->status' check seems unnecessary here since the button is only there when the status is 'STATUS_VALIDATED'.

### DIFF
--- a/htdocs/product/inventory/inventory.php
+++ b/htdocs/product/inventory/inventory.php
@@ -801,7 +801,7 @@ if ($object->id > 0) {
 	print '</div>';
 
 	// Call method to disable the button if no qty entered yet for inventory
-	if (!$hasinput) {
+	if ($object->status != $object::STATUS_VALIDATED || !$hasinput) {
 		print '<script type="text/javascript" language="javascript">
 					jQuery(document).ready(function() {
 						disablebuttonmakemovementandclose();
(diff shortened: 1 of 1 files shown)


xgerhard:
Hey, thanks for merging the PR! Just wondering why the status check was added back, did I miss something?
The button #idbuttonmakemovementandclose seems to only be there when the object has status 'validated': https://github.com/xgerhard/dolibarr/blob/ba1590758267c010f0dcc10275af39f7a3e5b698/htdocs/product/inventory/inventory.php#L480-L482

PR #18905, opened 2021-10-05
Fix creataing invoice originating for custom module object

indelog:
This fix the invoice create form when we provide `origin` element in
url. The actual create form overide the value for origin field with
`$objectsrc->element`. This make creation of invoice originating from a custom
module module impossible because we need the module part to instantiate
the class for the custom module object.

This also modify Common Object::fetch Object Linked() to append the
module part for `$sourcetype` and `$targettype` if it not set and object
have module property and it not in core module and 
`CommonObject::add_object_linked()` to append the module part if
object has `$module` property and is not in core modules.

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -3731,7 +3731,7 @@ if ($action == 'create') {
 		print '<input type="hidden" name="total"    value="'.$objectsrc->total_ttc.'">'."\n";
 		print '<input type="hidden" name="tva"      value="'.$objectsrc->total_tva.'">'."\n";
 		// The lines below override the parameters set by GET or POST for origin value and also be present in $origin and $originid variables (it make creation of invoice originating from an external module object impossible because for get an external module class we also need to know the name of the module in addition of the name of the element).
-        // This input fields already printed above at the <form> begin.
+		// This input fields already printed above at the <form> begin.
 		//print '<input type="hidden" name="origin"   value="'.$objectsrc->element.'">';
 		//print '<input type="hidden" name="originid" value="'.$objectsrc->id.'">';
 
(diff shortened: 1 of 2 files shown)


htdocs/compta/facture/card.php
```diff
@@ -3730,8 +3730,10 @@ function setRadioForTypeOfIncoice() {
 		print '<input type="hidden" name="amount"   value="'.$objectsrc->total_ht.'">'."\n";
 		print '<input type="hidden" name="total"    value="'.$objectsrc->total_ttc.'">'."\n";
 		print '<input type="hidden" name="tva"      value="'.$objectsrc->total_tva.'">'."\n";
-		print '<input type="hidden" name="origin"   value="'.$objectsrc->element.'">';
-		print '<input type="hidden" name="originid" value="'.$objectsrc->id.'">';
+		// The lines below override the parameters set by GET or POST for origin value and also be present in $origin and $originid variables (it make creation of invoice originating from an external module object impossible because for get an external module class we also need to know the name of the module in addition of the name of the element).
```
eldy, line 3733: If we remove this, we will lose them when submitting form and there is an error.
If origin and originid are provided on url, 
$objectsrc->id and $objectsrc->element should have the value of value provided.
So it overwrites yes, but with the same value to guarantee the data is propagated to the post.

indelog:
I correct my first approach to customize sourcetype and targettype in methods
`CommonObject::showLinkToObjectBlock()` and CommonObject::fetchObjectLinked().
I thinked to use the `$coremodule` to determine if we must add the module
part to objecject element (wich is needed for get class path when the object
provide form a custom module).

It' a bad idea because it do not take care of other cusom modules and it make
the core code more complicated to maintain because we must take care of
updating the `$coremodule` value on each method which this variable is present
if we add `$module` property in object owned by per-existing core module.

Rather do this we can simply add a hook `setLinkedObjectSourceTargetType` for
allow developer of custom module to easily override sourcetype and targettype
if needed without impacting the core behavior.

_Example of hook usage :_

```
public function setLinkedObjectSourceTargetType($parameters, &$object, &$action, $hookmamager)
{
    if ($parameters['sourcetype'] == 'myobject') {
        $this->results['sourcetype'] = 'mymodule_myobject';
    }
    if ($parameters['targettype'] == 'myobject') {
        $this->results['targettype'] = 'mymodule_myobject';
    }
    return 1;
}
```

Issue #18443, opened 2021-08-19
The option "overwrite file if exist" doesn't work in DMS/ECM, manual tree.


edupulpillo:
# Bug
The option "overwrite file if exist" doesn't work in DMS/ECM, manual tree.

## Environment
- **Version**: *14.0.0*
- **OS**: *Linux64*
- **Web server**: *Apache/2.4.46 *
- **PHP**: *7.4.7*
- **Database**: *MySQL or MariaDB 5.6.29*
- **URL(s)**: *http://mycompany.com/ecm/index.php?uploadform=1*

## Expected and actual behavior
*The file should be replaced, with a warning message indicating "the file was replaced for a new version.*

## Steps to reproduce the behavior
*Verbose description*
Upload a file in the maual tree, and observe that selecting the checkbox "overwrite file if exist" does not work,

## Screenshot:
![image](https://user-images.githubusercontent.com/7486607/130030696-107508f1-b2a9-4006-b282-1c681e2ec8bd.png)

PR #18906, opened 2021-10-05
Fix PR #18845 and #18843 (error on src_objet_type from emc files)

indelog:
The PR #18845 and #18843 modify `CommonObject::commonGenerateDocument()`
and emc auto index page to use a `'@'.$object->module` suffix appened to `
`@object->table_element` when object has `$object->module` property. It's a
mistake (when originating object is referenced in stock movement table,
the module part is indicated after a `@` suffix and i thought it was a convention for
specify module part but in here `_` must be used).

Before the PR #18445 and #18443, `$object->table_element` was used to
set `src_object_type` in `llx_ecm_file`, normally the tables are named
like `$object->module.'_'.$object->element` but for random reason it
possible to name them otherwise (objects has specific `table_element`
property for that). So in CommonObject::commonGenerateDocument(), if the
object has a property `$this->module` use
`$this->module.'_'.$this->element` seem more reliable than use
`$this->table_element`.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -5297,7 +5297,7 @@ abstract class CommonObject
 								$ecmfile->gen_or_uploaded = 'generated';
 								$ecmfile->description = ''; // indexed content
 								$ecmfile->keywords = ''; // keyword content
-								$ecmfile->src_object_type = $this->table_element.(empty($this->module) ? '' : '@'.$this->module);
+								$ecmfile->src_object_type = (empty($this->module) ? '' : $this->module.'_').$this->element;
 								$ecmfile->src_object_id   = $this->id;
 
 								$result = $ecmfile->create($user);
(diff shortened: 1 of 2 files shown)


eldy:
I reverted the PR #18845 and #18843.
However, what's wrong with table_name. From v14, value of ->table_name in an external module should be  "module_table" so it should be enough as a value for modulepart.

indelog:
It'right. But i thinked this is not the purpose of `$object->table_name` to be
used for this (all others places in the code where we need to get module part
added to the element name of an object we use `$object->module` and
`$object->element`.  It seem to me it may be than possible for some unknown
reason than a developer need to name a table for an object with anoter name
than `mymodule_myobject` (it not my case), i just think it more safer way and
semantically more correct to use `$object->module.'_'.$object->element` rather
`$object->table_element` to specify `src_object_type` of an ecm file.

eldy:

PR #18920, opened 2021-10-07
Close #18479 added dropdown function.

### LINKED ISSUE #18479: PROJECT: Create new object button
If you activate the global `PROJECT_SHOW_CREATE_OBJECT_BUTTON` you get a huge amount of buttons on the project card:

image

Maybe nicer would be something looking like this:
image

Example code: https://www.w3schools.com/howto/howto_js_dropdown.asp

But it should work like the quickadd or bookmark dropdowns

oliolo:
# Close #18479 added dropdown function.
Following the feature request #18479, we added a dropdown function in htdocs\projet.php\card.php to not have all the buttons visible at the same time. We also added CSS code at the end of both htdocs\theme\eldy\dropdown.inc.php and htdocs\theme\md\dropdown.inc.php to assist in the above change.


aspangaro:
Thanks, can we have a picture ?

oliolo:
dropdown

eldy:
Idea to group all "Create" actions looks good.
However a button with an arrow would be nicer, like here
image

Do you think you can reach this effect with css change ?

cfoellmann:
I am with @eldy. Should be a little bit nicer. I am no css designer but this would stay in the UI for a long time.
Adding a style for the whole system to use would be preferable

eldy:
This dev has been merged and enhanced manually into develop branch for v18

PR #18925, opened 2021-10-08
FIX reload user lang

lvessiller-opendsi:
FIX reload user lang
- load user configuration profile before trying to load user lang (not already defined)


eldy:
It's hacktoberfest time, you can create your profile here to participate: https://hacktoberfest.digitalocean.com/profile

PR #18931, opened 2021-10-08
FIX: consistent UX when calling a tab from the invoice card with empty ref/id

atm-florianm:
# Fix
Currently, if you call a tab from the invoice card without providing an invoice ID or ref, the user experience is inconsistent: for most pages the user will get a blank page, for documents, an error message, and for pages with modules implementing hooks, it might even lead to fatals.

Example: https://demo.dolibarr.org/compta/facture/card.php?facid=

This PR tries to make it more consistent. For future Dolibarr versions, it might be a good idea to create functions to handle this, but I thought it was out of scope for a fix.


eldy:
It's hacktoberfest time, you can create your profile here to participate: https://hacktoberfest.digitalocean.com/profile

PR #18940, opened 2021-10-10
NEW : info of number of product still to dispatched in a supplier order

PyroShape:
When you are in the tab "Commande fournisseur" you can now see if you have some product to receive and dispatch.
With this it's more user friendly to see if there is still an action to do in the dispatch product of the supplier order.

For example here you have still 2 product to receive and dispatch from your supplier order.
image

Then, you have 1.
image

### DIFF
--- a/htdocs/core/lib/fourn.lib.php
+++ b/htdocs/core/lib/fourn.lib.php
@@ -162,14 +162,17 @@ function ordersupplier_prepare_head($object)
 
 		//If dispach process running we add the number of item to dispatch into the head
 		if ($object->statut == '3' OR $object->statut == '4' OR $object->statut == '5') {
-			$lines = $object->fetch_lines();
+			$object->fetch_lines();
+			$nbLinesOrdered = count($object->lines);
 			$dispachedLines = $object->getDispachedLines(1);
+			$nbDispachedLines = count($dispachedLines);
+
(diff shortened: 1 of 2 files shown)


eldy:
To help understand, can you paste screenshot before / after to see what the feature brings ?

PyroShape:
> 
> 
> To help understand, can you paste screenshot before / after to see what the feature brings ?

Of course, I have updated the PR description : tell me if you need more information.

eldy:
Thanks for screenshot. I understand now. 

Also, this must follow the same rule than for shipment. And currently for shipments, the badge show the number of shipments done (like it is for reception done for receptions). So if you change here to replace number of reception with a qty, you must also replace on shipment side (it must represents same things).

However, on other tabs, the number represent the number of documents or a qty so it would be easier to understand if we follow this rule too (here your patch suggests to show the number of lines). We should not introduce a third unit. So i suggest to keep the number of documents (nb of receptions documents already done), or to show the remain of quantity to receive, but in suche a cas we must show both the number that remains to receive and the total to receive.
So if remain to receive is 2 (when total to receive is 10), can you instead put in the badge "8/10"   (8 received on 10 to received). So when we have 1 number (like most tabs), this is a count of record, when there is 2 numbers like x/y (like for the production module), it is a quantity already done / qty recorded.
When nothing has been received you can have   "0/10"

PR #17819, opened 2021-06-04
NEW : rewrite bind accountancy code finding on sales invoice

FHenry:
- Bind function is now a method of AccountingAccount class
 - Can be hooked by a hookmanager call
 - TODO : Do the same things on supplier side

Issue #18859, opened 2021-09-30
Update supplier part with new way AccountToBind method (next of #17819)


FHenry:
Next of #17819

For supplier side, you can reuse the getAccountingCodeToBind(). You can use $facture->element to knwo if it is a supplier or seller invoice so into getAccountingCodeToBind, instead of using
$conf->global->ACCOUNTING_SERVICE_SOLD_ACCOUNT and ACCOUNTING_PRODUCT_SOLD_ACCOUNT
you will be able to use
$conf->global->ACCOUNTING_SERVICE_BUY_ACCOUNT and ACCOUNTING_PRODUCT_BUY_ACCOUNT

Also for supplier side, the param

$accoutinAccountArray = array(
			'dom'=>$objp->aarowid,
			'intra'=>$objp->aarowid_intra,
			'export'=>$objp->aarowid_export,
			'thirdparty' =>$objp->aarowid_thirdparty);

Will surely be

$accoutinAccountArray = array(
			'dom'=>$objp->aarowid,
			'intra'=>$objp->aarowid_intra,
			'export'=>$objp->aarowid_export,
);

without the 'thirdparty' case filled.

PR #18942, opened 2021-10-11
NEW Accountancy - rewrite bind accountancy code finding on purchase invoice

aspangaro:
& uniformize

Refer to #17819 

Supplier side
Bind function is now a method of AccountingAccount class
Can be hooked by a hookmanager call


FHenry:
Related to #18859

FHenry:
Close because here #19110 19110

PR #18945, opened 2021-10-11
FIX : Remove not complete order from the virtual stock

PyroShape:
Order at statut '2' (Approved but not yet ordered) should not be add in the virtual stock.
At this stage the added stock is more "hypotetical" than "virtual".
(By the way the statut '2' is used bellow if we decide to "include draft for virtual").


eldy:
Currently definition of virtual stock is stock that take into account all open document. Status 2 is an open purchase order so it is included.
This is why there is option SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK so you can modify yourself the list of status id (because if you change here you must change at other place too. using this var will change at any place where we want to exclude the status 2).

PR #18957, opened 2021-10-12
FIX payment style and html5 tags

lvessiller-opendsi:
FIX payment style and html5 tags

- in terminal view, when you enter in payment page, you had :

TerminalReglement

- display "total" labels and values in center of the terminal :

image

And I removed "center" and "font" tag that are not supported in HTML5.
And I fixed account select when "TAKEPOS_CAN_FORCE_BANK_ACCOUNT_DURING_PAYMENT" was enabled.

### DIFF
--- a/htdocs/takepos/css/pos.css.php
+++ b/htdocs/takepos/css/pos.css.php
@@ -320,12 +320,11 @@ table.postablelines tr td {
 
 div.paymentbordline
 {
-	width:calc(50% - 16px);
-	background-color:#888;
+	background-color: #888;
 	border-radius: 8px;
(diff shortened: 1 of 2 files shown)


eldy:
The problem is that sometimes (if my memory is good) we have a lot of other fields (already paid, bank account, may be more, etc..) and we won't to have the height of this section being too high to be sure to never have the vertical scroll appears.
So choice to set this on 2 columns were done to address this.

lvessiller-opendsi:
@eldy 
I understand tou remarks and I corrected it to have 2 columns.

PR #18831, opened 2021-09-27
NEW Can select existing batch numbers for a combo list (in mrp consumtion)

SteiniGreenwork:
# New load existing batch numbers for item  within mrp consumtion
While consuming items within mrp, existing batch numbers for the corresponing item are loaded into select field

PR #18959, opened 2021-10-12
NEW: select from existing serial/lot number when produce an Manufacturing Order

FHenry:
Before
image
You can input serial/lot that do not exists and it can be source of error when produce

Now
image

When consume item that have Lot/Serial, Dolibarr suggests existing Serial/Lot


eldy:
We must still be able to enter a lot that does not exists (inventory is not always correct and we must be able to continue production despite of this).
Suggesting existing lot is good to reduce the error.
But another way to address this has been implemented and we should use this method everywhere:
Everwhere there is an input text for lot/serial, you can call just after the method
selectLot('xxx')
with 'xxx' = Name inside attribute "list" of the input text field

FHenry:
With a system like now, errors will occurs when input serial num, it even by design as it's a free input, and in an ERP it's a real problem to not to have quality in the data. So stock will be wrong all the time.....
With this feature, if you have a error in stock, first you fix it and then you produce, like that you provide more quality in alll company. 
By the way, it's the same design as in Shippement module, as you said you would be able to ship non existing serial , it's a non-sense

atm-maxime:
#18831 ?

FHenry:
Ok, I didn't notice this PR an feature, that better than an input, I'll do another PR to explain it's a search input, add quantity available in select list and a split feature

PR #18960, opened 2021-10-12
Add an option to forbid changing the project of a supplier order

vincentjdc:
# New Add the option to forbid project change of a supplier 

Add the option (global var `FORBID_EDIT_PROJECT_ON_SUPPLIER_ORDER`) to forbid project change of a supplier order if :

- Order is validated
- Order is in draft but already has a reference that is not `(PROV*)`.

This is useful when the supplier order ref generator uses the project ref. 

The option can be activated in the supplier module configuration.


eldy:
Can you rename var into 
SUPPLIER_ORDER_FORBID_EDIT_PROJECT
(must start with name of project)

vincentjdc:
> Can you rename var into SUPPLIER_ORDER_FORBID_EDIT_PROJECT (must start with name of module)

Done :)

PR #18968, opened 2021-10-13
fix: in case of csrfattack error 500 =>  dol_syslog is not define yet, function.lib.php is include latter

FHenry:
in all way, process die, but investigate why error 500 is lost of time, I prefer print and die


eldy:
I made a different fix so we can still have the log (we need it for fail2ban)

PR #18974, opened 2021-10-13
Fix ExtraFields::setOptionalsFromPost() check enabled

indelog:
`ExtraFields::setOptionalsFromPost()` check if extrafield is enabled or
not with `list` property. Shouldn't that be the property `enabled` instead ?


eldy:
Yes, we should use both 'enabled' that is used to say if property is enabled by a module activation or not.
And 'list' that is used for the visibility (hidden, visible).
So i pushed a more complete fix, but into develop due to side effect that this change may have.

PR #18975, opened 2021-10-13
FIX Accountancy - If deposit invoice is used, force binding in deposit accounting account to solve transaction

aspangaro:
Before, only deposit line from final invoice are put to accounting deposit account. 

Also alls lines from original deposit invoice must be binded into accounting deposit account. 

image


aspangaro:
This is what we have to achieve : (In next PR, we also add subledger account in specific deposit line to help when we have a lot of deposit)

image

PR #19001, opened 2021-10-18
FIX : keyval filters on lists don't work

atm-gauthier:


atm-gauthier:
Hi @eldy, i did this update and actually it works with this method !

PR #19013, opened 2021-10-19
NEW: Ticket status "COMLPETED"

milenmk:
# New [*Ticket status "COMLPETED"*]

When a user is assigned a ticket and he completes the job, further to set the progress to 100%, he can also mark the ticket as  COMPLETED. 

Thus, it will be easier for the assigner/supervisor to see that his problem is solved (or task is completed) and close the ticket as closed. This can be seen as a duplicate to 100% progress (and in fact, it is), but not everyone has the progress column active on the list view, while the status is active by default.

# NOTE
Theme files "badges.inc.php" were changed to give the new status a badge, which looks like as shown in the 
Screenshot_1
attached file.


eldy:
This introduce duplicating of storing data. We already have the information into database. If we want to change the way it is viewed, this must be done by changing output (for example the status badge can take into account the percentage) and not by adding more data.
Also, note that once a task is complete, a good practice is to set status to "Waiting reporter feedback" so he know a check is expected and he can close the ticket.

milenmk:
> This introduce duplicating of storing data. We already have the information into database. If we want to change the way it is viewed, this must be done by changing output (for example the status badge can take into account the percentage) and not by adding more data. Also, note that once a task is complete, a good practice is to set status to "Waiting reporter feedback" so he know a check is expected and he can close the ticket.

I.e. better approach will be to fetch the "progress" column from the database, and if it is 100 the output will be "Waiting reporter feedback" without introducing new status in class, correct?

But, this way filtering via "Status" won't be possible, because we need "$newarrayofstatus[] = Ticket::STATUS_???;" in the list view and definition in class.

I can change the naming from "COMPLETED" to "Waiting reporter feedback", while changes must remain as they are now.

eldy:
Fetching the 100 from database and use it to change the look of the status "In progress" is good.  For example by changing label into "In progres (completed)"
However, the status "Waiting reporter feedback" must remain as a different status (it already exists as an independant status and it is used to mark that the ticket is now on the requester side).

milenmk:
> In progres (completed)

Just look through the code again. And everything about the status, both in list.php and card.php is connected to "public function LibStatut". So even if the header template is modified, the above will be possible only for  "card.php?action=view" and in this case, it is pretty useless.

htdocs/ticket/class/ticket.class.php
```diff
@@ -218,6 +218,7 @@ class Ticket extends CommonObject
 	const STATUS_IN_PROGRESS = 3;
 	const STATUS_NEED_MORE_INFO = 5;
 	const STATUS_WAITING = 7;			// on hold
+	const STATUS_COMPLETED = 6;			// Completed by the user who has been assigned to the ticket. Waiting to be closed as solved.
```
eldy, line 221: No need for this status. There is already an information in database to say action is complete (STATUS_IN_PROGRESS + progression = 100 should be enough to show status "Completed").
The progression has been added into the label of status.

PR #19024, opened 2021-10-19
Fix adhtype amount as default on new subscription

daraelmin:


daraelmin:
Ok for me, I've tested just now.

PR #19035, opened 2021-10-20
Fix default delay for subscription is not used

daraelmin:
defaultdelay and defaultdelayunit were defined at line 253 but not used.
TODO : add a global constant in admin panel of thé module to adjust these variable.

### DIFF
--- a/htdocs/adherents/subscription.php
+++ b/htdocs/adherents/subscription.php
@@ -77,7 +77,7 @@ $extrafields->fetch_name_optionals_label($object->table_element);
 
 $errmsg = '';
 
-// TODO add a global constant in admin panel of the adhetent module to adjust default delay. 
+// TODO add a global constant in admin panel of the adhetent module to adjust default delay.
 $defaultdelay = 1;
 $defaultdelayunit = 'y';
 
--- a/htdocs/adherents/subscription.php
+++ b/htdocs/adherents/subscription.php
@@ -77,6 +77,7 @@ $extrafields->fetch_name_optionals_label($object->table_element);
 
 $errmsg = '';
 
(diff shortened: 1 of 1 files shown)


eldy:
They are not used because we want now to keep this field empty by default (to reduce number of field to enter or read).
The delay by default of 1year should be defined after clicking on button create. It is not the case for you ?

daraelmin:
I'm not sure I understood the expected process.

in v14, when I click on create, the "end date subscrption" field is empty, the software does not ask me to enter a time limit (see screenshot).

However, the default delay exists and is used at creation time (if the end date is left blank (line 253).

It is not clear to the user who filled out the membership form. Either the field should appears with a deadline proposal, or it must remain hidden.

By the way, an empty field should be able to be entered for associations which have a subscription valid for life (without expiration date). I'll correct my PR in this sense (by removing line 253).

daraelmin:
The empty value of the subscription end date is currently not foreseen. In the Adherent-> subscription function, having an empty date, or null, results in a default duration of one year. The database do not accept the null value, so it is not possible to introduce this quickly.

In the event that my proposal to modify this treatment is validated, I think that it would take a separate PR. I will be offering this PR as a new feature rather than a FIX.

eldy:
With your fix it means that if you want another value than the start and end date suggested,you must change both date.
Currently you set only the start date (and you enter end date ONLY if end date must not be the default date).
If we want a start date that differs, 
I think we should keep this behaviour. May be we can just add a note after the end date like this
"Keep this field empty to use the default period of X month"

daraelmin:
Ok. 
But the note should use the var $defaultdelay and $defaultdelay unit in case we change this (with constant or hook)

PR #19079, opened 2021-10-21
FIX #19064 - access forbidden when editing other user's password

### LINKED ISSUE #19064: Access forbidden when the password of other users is changed
# Bug
Access forbidden when the password of other users is changed

## Environment
- **Version**: 14.0.2
- **OS**: All
- **Web server**: All
- **PHP**: 7.4.13
- **Database**: MariaDB 10.6.4
- **URL(s)**: /htdocs/user/card.php?id=xx

User with lire and password (user) perms, but no creer.

When this user saves or cancels the edition of the password of another user, he is sent to the prohibited page

atm-florianm:
# FIX #19064 
When a user does not have permission to create other users but has permission to edit other users' passwords, when they do so, they land on a rejection page.

Since this fix touches core security features, I wouldn't mind extra scrutiny to make sure it doesn't introduce any new vulnerability.

### DIFF
--- a/htdocs/core/lib/security.lib.php
+++ b/htdocs/core/lib/security.lib.php
@@ -411,11 +411,11 @@ function restrictedArea($user, $features, $objectid = 0, $tableandshare = '', $f
 				}
 			} elseif (!empty($feature2)) {														// This is for permissions on one level
 				foreach ($feature2 as $subfeature) {
-					if ($subfeature == 'user' && $user->id == $objectid && $user->rights->user->self->creer) {
-						continue; // User can edit its own card
-					}
-					if ($subfeature == 'user' && $user->id == $objectid && $user->rights->user->self->password) {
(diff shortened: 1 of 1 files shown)


eldy:
What happen after your fix if:
User has only permission to edit password of others and user try to edit the firstname ?

atm-florianm:
> What happen after your fix if: User has only permission to edit password of others and user try to edit the firstname ?

I just tested this scenario:
1) (as admin) I gave user A the right to edit other users
2) (as user A) I opened the edit form for user B and changed the first name
3) (as admin) I removed permission from user A
4) (as user A) I submitted the form

Result:
No forbidden access, but also no first name change for user B (the modification was not saved, which is what we expect with respect to permissions).

atm-florianm:
Closing the PR: I realize my PR is a duplicate of #19065 (though fixes differ slightly).

PR #19087, opened 2021-10-22
fix: bad comma SQL request in modulebuilder template list 

FHenry:
fix source bug of #19081


frederic34:
in develop https://github.com/Dolibarr/dolibarr/pull/19088

Issue #19091, opened 2021-10-22
Link object to category


evarisk-theo:
It is possible to link thirdparties, customers, suppliers and others to a category with this form but it's not possible for objects as members, contacts, projects and others

![image](https://user-images.githubusercontent.com/68421590/138430490-033200e6-f59d-4f9c-b124-2881ff891444.png)

PR #19116, opened 2021-10-22
Fix link object to categories

evarisk-theo:
#19091

add quick link between categories and account, user, contact, adherent & project.

couldn't add this functionality to actioncomm & website page :
-  actioncomm : the rowid database field is named id so an error occurs when getting linked objects to the category
- website_page : database doesn't have an entity column


eldy:
Can you provide a screenshot of before/after the change ?

evarisk-theo:
Of course. This code add the quick link to an other object form to a category card.

before : 
image

after : 
image

PR #19121, opened 2021-10-22
fix multiple warnings

frederic34:


frederic34:
@atm-maxime @FHenry need glasses to control

I replace !user->rights->xxxx by empty($user->rights->xxxx)

Issue #19111, opened 2021-10-22
wrong redirection after contact creation


evarisk-theo:
![image](https://user-images.githubusercontent.com/68421590/138470411-e3e85a8c-0af0-4673-b755-4c47e7bea742.png)

tiaris:
I can not reproduce it with 14.0.3. Can you give more details ? Configuration?

evarisk-theo:
This bug happens only on 15.0.0 alpha

PR #19136, opened 2021-10-23
fix: redirection after creation contact #19111

nicolas-eoxia:
#19111


eldy:
Note: To have the issue automatically close when PR is merged, you must put
Fix #19111 
instead of just 
#19111 
into the description.

PR #19144, opened 2021-10-23
FIX :  dom declaration

john-botella:
A little dom fix, to be able to structure menus and sections for theming improvement and follow html standards.


kevin-reo:
More info for HTML Standard : 
- https://developer.mozilla.org/fr/docs/Web/HTML/Element/nav
- https://developer.mozilla.org/fr/docs/Web/HTML/Element/header
:)

PR #19147, opened 2021-10-23
BUG Can't add products when load draft invoice from history

### LINKED ISSUE #18935: BUG When load a draft invoice from history I can't add products
# Instructions
Only load history invoice unclosed and add products

# Bug
When load a draft invoice from history I can't add products

## Environment
All versions

andreubisquerra:
FIX https://github.com/Dolibarr/dolibarr/issues/18935

### DIFF
--- a/htdocs/compta/facture/list.php
+++ b/htdocs/compta/facture/list.php
@@ -1719,7 +1719,13 @@ if ($resql) {
 
 			print '<tr class="oddeven"';
 			if ($contextpage == 'poslist') {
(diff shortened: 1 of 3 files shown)


andreubisquerra:
Fixed in https://github.com/Dolibarr/dolibarr/commit/85cc5851efd3667e7daacaa33ca94a403a46fe43#diff-9a5edd954ab65b86539a2b7ba4dea0daa2709f76d81a7afd31534ff046d36e52

PR #19186, opened 2021-10-24
Fix excluded member display in takepos

ptibogxiv:


daraelmin:
Hi,
It may me more readable if you use constant Adherent::STATUS_EXCLUDED instead of -2

PR #19189, opened 2021-10-25
Fix - add product line - always fail to input multicurrency price if product price is by default 0 (supplier invoice, sales order &  invoice)

henrynopo:
For my cases, the product selling price is not pre-defined in Dolibarr and the system set it at 0 by default for selling prices and NULL for purchase.

image

image

As multicurrency is enabled, the current coding will not register the U.P. in order currency manually input while adding a predefined product in supplier invoice / sales order / customer invoice. 

It means that when a pre-defined product is added, a U.P. main currency can be registered but if a user tries to input the U.P. (currency) (let's say 1$) then the product lines upon added will show a price of 0 in both U.P. of main currency and multicurrency and the user has to re-edit the product line in order to update the multicurrency price to correct value.

With the proposed fix, multi-currency U.P. can be input / registered successfully when a new product line is added into supplier invoice / sales order / customer invoice.

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -2101,7 +2101,12 @@ if (empty($reshook)) {
 
 				// if price ht was forced (ie: from gui when calculated by margin rate and cost price). TODO Why this ?
 				if (!empty($price_ht) || $price_ht === '0') {
-					$pu_ht = price2num($price_ht, 'MU');
+					if($price_ht === '0' && $price_ht_devise != '') { //added test to record multicurrency price while the default price is 0.
+						$pu_ht_devise = price2num($price_ht_devise, 'MU');
+						$pu_ht = price2num(GETPOST('price_ht', 'MU'));
+					} else {
+						$pu_ht = price2num($price_ht, 'MU');
+					}
 					$pu_ttc = price2num($pu_ht * (1 + ($tmpvat / 100)), 'MU');
 				} elseif ($tmpvat != $tmpprodvat) {
 					// On reevalue prix selon taux tva car taux tva transaction peut etre different
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -2101,7 +2101,7 @@ if (empty($reshook)) {
 
 				// if price ht was forced (ie: from gui when calculated by margin rate and cost price). TODO Why this ?
 				if (!empty($price_ht) || $price_ht === '0') {
-					if($price_ht === '0' && $price_ht_devise != '') { //added test to record multicurrency price while the default price is 0.
+					if ($price_ht === '0' && $price_ht_devise != '') { //added test to record multicurrency price while the default price is 0.
 						$pu_ht_devise = price2num($price_ht_devise, 'MU');
 						$pu_ht = price2num(GETPOST('price_ht', 'MU'));
 					} else {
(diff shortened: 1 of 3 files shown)


htdocs/compta/facture/card.php
```diff
@@ -2101,7 +2101,12 @@
 
 				// if price ht was forced (ie: from gui when calculated by margin rate and cost price). TODO Why this ?
 				if (!empty($price_ht) || $price_ht === '0') {
-					$pu_ht = price2num($price_ht, 'MU');
+					if ($price_ht === '0' && $price_ht_devise != '') { //added test to record multicurrency price while the default price is 0.
+						$pu_ht_devise = price2num($price_ht_devise, 'MU');
```
eldy, line 2105: price_ht_devise here is value price2num(GETPOST('multicurrency_subprice'), 'MU', 2);
So if you also set 
$pu_ht = price2num(GETPOST('price_ht', 'MU'));
So both price_ht_devise and pu_ht are set according to value coming from use input. If may be non consistent values. For example, we may have pu_ht = 1, multicurrency_subprice = 2 and a rate of 1.5.
To be sure data are not corrupted, we must:
- If data was entered in main currency, value for foreign currency must be calculated from it.
- If data was entered in foreign currency, value for main currency must be calculated from it.
We should not have both value coming from user input.

henrynopo: Sorry Laurent, it seems that I don't fully understand your suggestion or the existing coding logic.

As explained in my case, we are using a multicurrency function and of course, the price is input only for multicurrency. Each of my colleagues (users) can see that the system calculates the price in the main currency (pu_ht) by itself.

The key issue is that, when one tries to add a line of pre-defined product or service into the invoices or sales order, the existing coding does not record the value of multicurrency price and resulted in 0 since there is no input for pu_ht (and by default it is 0 for the selling price of pre-defined products or services). Otherwise he / she has to manually calculate the pu_ht and input it while adding the lines in concern; and alternatively, he or she must re-edit this line in order to update the multi-currency price with correct value.

I will appreciate very much if you can suggest a fix, more appropriate if any, to this bug.

eldy: You must decide what is the master value used to saved data.
If it is the multicurrency_subprice, then the pu_ht must be calculated using the currency rate.
If it is the pu_ht, then the multicurrency_subprice must be calculated using the currency rate.
But inside the "if" line 2104, you have no calculation. Both value pu_ht and multicurrency_subprice are coming from a user input. Is my explanation better ?

henrynopo: Hi Laurent. I see what you meant.

The newly added "if" condition intended only to fix the issue I had raised, which means that there is no pre-defined price (both pu_ht and pu_ht_devise equals to 0) and there is no user input for multicurrency_price_ht.

With regards to your concern, I have tried with my fixed coding. If I input subprice for both price_ht and multicurrency_price_ht, the newly added "if" codnition is not satisfied so the calculation will fall to the "else" condtion and take price_ht. The multicurrency_price_ht is calculated according to the predefined exchange rate.

Of course, you may have better ideas how to improve this portion of coding. For example, as a non-programmer, I did not find how to activate the dynamic display of the price_ht or multicurrency_price_ht when a user is inputting price for the other one when adding a product line (this function is activated when editting the product line).

PR #19197, opened 2021-10-25
fix: #19009

### LINKED ISSUE #19009: Contract / Style for the new line part seems messed up
# Bug
When adding a new line to a contract, the style for the new line part seems messed up

## Environment
- **Version**: 14.0.2
- **OS**: Ubuntu 20.04.1
- **Web server**: Apache 2.4.41
- **PHP**: 7.1.33
- **Database**: MariaDB 5.5.5-10.3.31
- **URL(s)**: /contrat/card.php

Expected: Labels floating left, inputs floating right
Actual: Labels floating left, inputs directly next to them

1. Add new line to any contract

Tried clearing cache, Ctrl+F5, using other browsers, etc... No difference.

[*Files*
### Before (v12.0.4)
Before (v12.0.4)
### After (v14.0.2)
After 14.0.2
]

FHenry:
Try to fix #19009


FHenry:
image

htdocs/core/boxes/box_factures_fourn_imp.php
```diff
@@ -101,6 +101,7 @@ public function loadBox($max = 5)
 			$sql .= " AND f.entity = ".$conf->entity;
 			$sql .= " AND f.paye = 0";
 			$sql .= " AND fk_statut = 1";
+			$sql .= " AND f.date_lim_reglement IS NOT NULL";
```
eldy, line 104: If an unpaid invoice has no dead line, it is always an unpaid invoice, isn't it ?
So why discarding it ?

FHenry: @eldy Yes trus, the real bug (warning php) was fix in dol_print_date

eldy:
Problem is that with your fix all input data are under the label instead of being on right of label. Did you try this
```
<dl>
<dt>Title of extrafiels</dt>
<dd>Value</dd>
<dt>Title of extrafiels 2</dt>
<dd>Value 2</dd>
</dl>
```

See https://onextrapixel.com/how-to-use-dl-dt-and-dd-html-tags-to-list-data-vs-table-list-data/

FHenry:
@eldy In this case it is the method showOutputField the should be change and not showOptionals.
CAnnot be done in 14.0, I prefer try to update it in develop. But change a output from showOutputField can really be tricky regarding every where it's use

PR #19208, opened 2021-10-26
FIX: Can't create new attendee

Quentin-Seekness:
Field date_subscription is mandatory when creating a new attendee.


frederic34:
@FHenry

htdocs/public/eventorganization/attendee_new.php
```diff
@@ -268,6 +268,7 @@ function llxFooterVierge()
 		} else {
 			// Need to create a confattendee
 			$confattendee->date_creation = dol_now();
+			$confattendee->date_subscription = dol_now();
```
eldy, line 271: The date_subscription is not a mandatory field. It is field when we record a payment (a different step than recording an attendee, coming after).

PR #19210, opened 2021-10-26
FIX Crash DateTime::setTimestamp()

fr69400:
FIX Crash PHP Warning:  DateTime::setTimestamp() expects parameter 1 to be int, string given in D:\Professionnel\dolibarr\htdocs\core\lib\functions.lib.php on line 2291

Bug description :
See https://www.dolibarr.fr/forum/t/mise-a-jour-vers-v14-0-3/38173
https://www.dolibarr.org/forum/t/datetime-settimestamp-expect-int/21339

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -2288,6 +2288,12 @@ function dol_strftime($fmt, $ts = false, $is_gmt = false)
 function dol_print_date($time, $format = '', $tzoutput = 'auto', $outputlangs = '', $encodetooutput = false)
 {
 	global $conf, $langs;
+	
+	// If date undefined or "", we return ""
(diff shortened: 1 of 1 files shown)


frederic34:
to submit in branch 14

frederic34:
https://github.com/Dolibarr/dolibarr/pull/19219
https://github.com/Dolibarr/dolibarr/pull/19230

fr69400:
These solutions doesn't work correctly... Example : When you have empty date in database (example : no validation date), you get with :
#19219 :  you display in fact the current date... So wrong data 
#19230 :  you display 01-01-1970 because '' is casted into 0

fappels:
@eldy, can you also include your fix in V14.
Empty datetime extrafield give warning, also for php 7.4
Product Computer XP4523 - Card 10-31-2021 12-17-52 PM

PR #19219, opened 2021-10-27
fix: warning dol_print_date setTimeStamp when time empty

FHenry:
fix: warning dol_print_date setTimeStamp when time empty


hregis:
@frederic34  yes you right ! 👍

PR #19222, opened 2021-10-27
Update list.php

oscim:
list contrat for thirdparty is not closed

# Instructions
*Please:*

[*Long description*]

[*Long description*]

[*Long description*]


FHenry:
@oscim is it not the case for other document list (propal, order, invoice, customer/supplier) ? shoud we add a "morefilter" for it ?

eldy:
Every filter used to restrict a list visible to end user must also be a filter "visible" to end-user (user must know why list is not complete and giving user a chance to see everything).
For the moment there is not available filter "Status of thirdparty (open or closed)", so adding this will lead to a blocking situation. User that need to see all contracts (even for closed thirdparties will have no more chance to be able to do it).
A better fix would be to add a filter field "Status of thirdparty". I would not use a $morehtml filter as suggested by @FHenry but a simple new column (we already have alot of information about the Thirdparty as filter, we can add one)
Note also that once we have such filter, it is also easy to define the default value of this filter using the generic filter feature.

eldy:

eldy:
Even if this PR goal is to fix a bug or add a feature, it seems it introduces more problems than it solves, so request is closed.

PR #19225, opened 2021-10-27
NEW: thirdparty: be able to define default contacts from other thirdparties

marc-dll:
This uses the UI template included in most of the "Contacts/Addresses" tabs.

No adaptation was necessary to the trigger code that does add default contacts to documents at creation.


marc-dll:
@eldy I just pushed a bugfix. Some feedback on this PR would be appreciated :)

eldy:
Can you test on v16, i think this has been implemented into v16.

eldy:
It seems this PR was set with "Discussion" or "PR to fix" tag or "Need more information", but no answer nor correction was provided since a long time. So request is automatically closed.

PR #19230, opened 2021-10-28
FIX DateTime::setTimestamp parameter must be an integer

hregis:


frederic34:
https://github.com/Dolibarr/dolibarr/pull/19219

Issue #19193, opened 2021-10-25
set defaults for form fields - not working consistently 


cfoellmann:
# Bug
example:
the `target` select for a bookmark is broken by this change https://github.com/Dolibarr/dolibarr/commit/687eb536aee1eb8a71d0077abc0ab1de20118a55 again. 
After my PR https://github.com/Dolibarr/dolibarr/pull/18349 it worked although I do not see why the change above does not work properly.

we need to think about moving the creation of forms to a class that needs to be extended like the `class CommonObject`

## Environment
- **Version**: develop 15.x

eldy:
I can't reproduce the trouble (on develop). Can you describe how you reproduce it ?

PR #19231, opened 2021-10-28
FIX: User defined default values ignored when using GETPOSTISSET as a condition to use GETPOST or hard coded default value

noec764:
GETPOSTISSET and GETPOST are designed to be used in conjonction but their behaviour was not aligned. 
This caused bugs with default values on Input value on Createform, Form Order or Search Filter that were not handled by GETPOSTISSET. Therefore default values were ignored in those cases. Fixes : #19193

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -241,36 +241,39 @@ function dol_shutdown()
 }
 
 /**
- * Return true if we are in a context of submitting a parameter
+ * Return true if we are in a context of submitting a parameter, or if there is a default value associated to this parameter,
+ * false otherwise.
  *
  * @param 	string	$paramname		Name or parameter to test
- * @return 	boolean					True if we have just submit a POST or GET request with the parameter provided (even if param is empty)
+ * @return 	boolean					True if we have just submit a POST or GET request with the parameter provided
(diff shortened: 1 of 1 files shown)


eldy:
Seems already fixed in develop.
Waiting for feedbacks on how to reproduce the bug on develop if not.

noec764:
I checked in develop, I did not find a correction for this problem, Indeed, the GETPOST function always performs more checks than its counterpart GETPOSTISSET.
For example if we have a default value for a specific field but nothing in $_POST, GETPOSTISSET returns false while GETPOST returns the default value.
So If we check if a value exist like:
```
$var = GETPOSTISSET('var') ? GETPOST('var', 'alpha') : '';
```
or 
```
$var =  empty(GETPOST('var', 'alpha')) ? GETPOST('var', 'alpha')  : '';
```
We will not get the same value in all case. 
@eldy

eldy:
> GETPOSTISSET

Goal of GETPOSTISSET is to say "yes, a value was submitted by the form, use it in priority before the other value". So it does not take into account the default value. So this looks to be it seems implementation is correct.
May be what you want to do is just 
$var = GETPOST('var', 'alpha')  

Can you describe the bug you are facing and how to reproduce it ?

PR #19234, opened 2021-10-28
FIX show connected server if we have a primary and secondary ldap server

hregis:


hregis:
@aspangaro Hello, can merge this PR please, i am blocking to continue, thank you

htdocs/adherents/ldap.php
```diff
@@ -209,7 +209,6 @@
 	}
 
 	$ldap->unbind();
-	$ldap->close();
```
eldy, line 212: Why removing the ldap->close() ?

hregis: @eldy this is an alias of unbind in php doc... Same function.

PR #19244, opened 2021-10-29
FIX missing service access right

hregis:


aspangaro:
Thanks @hregis 

Maybe propose this fix in older version if necessary than only develop.

hregis:
@aspangaro 
this line of code exists only in develp branch

PR #19248, opened 2021-10-29
Fix: Import don't work if extrafield is separator

fmarcet:
When importing, it also tries to insert extrafield 'separator' which has no column in table 'element'_extrafields


atm-maxime:
Thanks for your fix !
When you do a fix in 12.0, no need to do it again in 13 and 14. After merging your fix, the 12 branch will be merged in 13, and then 13 in 14, and 14 in develop. This is how your fix is propagated to all these versions. You can close your 2 other pull requests.

fmarcet:
Hello @atm-maxime , sorry if it was not the best way. But in version 13 and 14 there are imports that don't exist in version 12. That is why I made a different pull request for versions 13 and 14.

PR #19251, opened 2021-10-29
Fix: Import don't work if extrafield is separator

fmarcet:
When importing, it also tries to insert extrafield 'separator' which has no column in table 'element'_extrafields


eldy:
Already pushed into v12, so closed

fmarcet:
Hello @eldy , sorry if it was not the best way. But in version 14 there are imports that do not exist in version 12 or 13: bom. That is the reason why I made a different pull request for version 14.

PR #10796, opened 2019-03-11
NEW : cron arguments might come from php CLI (argv vs getpost)

altairis-tof:

Issue #8927, opened 2018-06-06
Cron job doesn't work


manukieli:
# Bug
The URL to execute cronjob work.
But when I schedule a task with the command, they don't run.
With Curl or Wget command to shedule the cron to the URL, they return <b>Userlogin is required</b>.
Or if I use command to execute they do nothing.

## Environment
- **Version**: 7.0.2
- **OS**: Centos 7.5
- **Web server**: Apache 2.4.6
- **PHP**: 7.1
- **Database**: MariaDB 5.6
- **URL(s)**: [*Affected URL(s)*]

## Expected and actual behavior
I'm using this command and they return this:
`Output from command /usr/bin/php /home/mydomain.com/public_html/scripts/cron/cron_run_jobs.php 96r9Br5izhGJK88xP40qYQ4GTuBE9fgk admin > /home/mydomain.com/public_html/documents/cron_run_jobs.php.log`

Execute this url to the browser, they work well:
`http://mydomain.com/htdocs/public/cron/cron_run_jobs.php?securitykey=96r9Br5izhGJK88xP40qYQ4GTuBE9fgk&userlogin=admin`

loxK:
Same here. Executing from command line:

```
# sudo -H -u www-data /srv/dolibarr/scripts/cron/cron_run_jobs.php [hash] [user]
***** cron_run_jobs.php (7.0.1) pid=32461 ***** userlogin=[user] ***** 1533171287 *****
cron_run_jobs.php cronjobid: 16
```

Ion00:
Same here with Dolibarr 8.0, upgraded from 6.0. Execute cronjobs from web explorer works fine, but execute cronjobs from cron does not work, this is the error messate:

Securitykey is required. Check setup of cron jobs module.

Note: Securitykey is already seted in cron jobs module setup

delcroip:
I just proposed a PR to fix it.

delcroip:
Hi,
Apparently we are executing the wrong file because there is another /cron_run_jobs.php  in Dolibarr/script/cron and this one takes the cli parameters into account.

Personally I'd rather have a single file that is used for the CLI and the CURL/WEB but it's not eldy's point of view:
https://github.com/Dolibarr/dolibarr/pull/9395

br

edupulpillo:
Maybe a good idea to have only one filename, but doing that the script be able to work in both modes CLI and CGI, taking care of the right parameters.

[...]

PR #19258, opened 2021-10-30
FIX #10796 #8927 Accept both CLI argv and CGI modes in /script/cron_run_jobs.php

edupulpillo:

### DIFF
--- a/scripts/cron/cron_run_jobs.php
+++ b/scripts/cron/cron_run_jobs.php
@@ -25,48 +25,68 @@
  * \brief Execute pendings jobs
  */
 
-if (!defined('NOTOKENRENEWAL')) define('NOTOKENRENEWAL', '1'); // Disables token renewal
-if (!defined('NOREQUIREMENU'))  define('NOREQUIREMENU', '1');
-if (!defined('NOREQUIREHTML'))  define('NOREQUIREHTML', '1');
-if (!defined('NOREQUIREAJAX'))  define('NOREQUIREAJAX', '1');
-if (!defined('NOLOGIN'))        define('NOLOGIN', '1');
-
+if (!defined('NOTOKENRENEWAL')) {
+	define('NOTOKENRENEWAL', '1'); // Disables token renewal
+}
+if (!defined('NOREQUIREMENU')) {
+	define('NOREQUIREMENU', '1');
+}
+if (!defined('NOREQUIREHTML')) {
+	define('NOREQUIREHTML', '1');
+}
+if (!defined('NOREQUIREAJAX')) {
+	define('NOREQUIREAJAX', '1');
+}
+if (!defined('NOLOGIN')) {
+	define('NOLOGIN', '1');
+}
+if (!defined('NOSESSION')) {
+	define('NOSESSION', '1');
+}
 
 $sapi_type = php_sapi_name();
 $script_file = basename(__FILE__);
 $path = __DIR__.'/';
 
 // Error if Web mode
-if (substr($sapi_type, 0, 3) == 'cgi') {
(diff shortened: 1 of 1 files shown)


edupulpillo (commented):
This PR should be applied also to v13 and v14. Is my first commit here, kindly help me to drive me to perform a PR in several branches. In this case i would like to PR in the versions: v12, v13, v14 and develop.

eldy:
No need to push a PR in several branch. If you push it for v12, it will be merged few days after into v13 , v14... until develop.

eldy:
Because the file is into the dir "scripts", it means it can be executed by CLI only. Only dire htdocs is reachable by web access.
You already have a page to access cron by url into  public/cron/cron_run_jobs_by_url.php

edupulpillo:
Thank you for the info!.
So then we have 4 versions of the script to launch scheduled jobs, with almost the same code, as you warn already inside the comment in sourcecode as "// TODO Duplicate code", except the first part that control the calling parameters:
1. /scripts/cron/cron_run_jobs (The only one supposed to accept to use CLI mode)
2. /public/cron/cron_run_jobs.php
3. /public/cron/cron_run_jobs_by_url.php
4. /public/cron/cron_run_jobs_bycgi.php

So if you agree, i can work to try to merge it in only one version of php script **cron_run_jobs.php** taking care of all the calling cases and applying the values and call controls accordingly the way it was invoked. 
Then we will have an easier code maintenance with only one file, instead using 4 different versions of files for cli, cgi, url. and script, providing more simplicity in the code, as you warn already in one comment inside as "// TODO Duplicate code".

The controls for the parameters could be:
1. If is detected that cron_run_jobs.php was launched from **/scripts/cron/** then
     1.1  if CLI mode detected, take securitykey  and userlogin from the passed arguments of the command line and check that are valid.
     1.2  if CGI mode detected, take securitykey from $conf->global->CRON_KEY and $userlogin = "firstadmin", (this help a lot in shared hostings, because now the security key and loginname can be seen with ps, basically is the PR i did, because **is the only way to make it works in my shared hosting and never php cli mode is detected**) 
2.  If is detected   that cron_run_jobs.php was launched **from any different path than /scripts** then:
     2.1  if CLI mode detected, return an error
     2.2 if CGI mode detected, get the parameters from GETPOST and check that are valid

Doing this way, also the same script can works fine in shared hosting environments with CPANEL that even is specified a CLI mode, is detected as cgi inside the script, and also works fine in external url request mode.

To avoid to have to files duplicated in **/scripts/cron/cron_run_jobs** and **/public/cron/cron_run_jobs.php**, maybe one can be a symbolic link to the another one or just that one file have inside only an include to the another one, to have only one version of the code.

At the moment the solution seems is already working pretty good in my instance of Dolibarr. :-)

eldy:
We only have 2 version of scripts
/scripts/cron/cron_run_jobs (The one used by CLI)
/public/cron/cron_run_jobs_by_url.php  (The one for used from CGI)

We still need to have this 2 different scripts. They are used in different context with different security rules.
We may have similar portion of code inside that may be factorized but not so much.

edupulpillo:
Thanks for the update and clarifications!. Then this pull request in
/scripts/cron/cron_run_jobs (The only one supposed to accept to use CLI mode) allows now to also accept to be calles in CGI mode instead CLI mode only, allowing to be invoked by cron configurations in shared hosting services that do not allow CLI mode, that without his update never works with CPANEL.
This feature do not create any problem in security rules, since is not under public_html, so only add the feature tp add both CLI and CGI modes.

eldy:
When you call a CLI script with CPANEL, are you able to set an environment variable ?
If you add a test  $_ENV['DOL_ACCEPT_CLI_AS_CGI'] to accept to run script as a cgi is a better option.

eldy:

PR #19259, opened 2021-10-30
FIX #10796 #8927 Accept both CLI argv and CGI modes in /script/cron_run_jobs.php

edupulpillo:


edupulpillo:
This PR should be applied to branches  v12, v13, v14 and develop. Is my first commit here.
Hoping i did in the expected way. Kindly drive me to perform a PR in several branches at the same time.

edupulpillo (commented):
Added missing parenthesis in conditions to check if master.inc.php exist

eldy:
Already pushed into v12, so closed here.

Issue #18996, opened 2021-10-17
Rewrite category/tag module


atm-maxime:
# Feature Request
Rewrite the category/tag module to make it easier to use and allow any element to be tagged

## Use case
Currently each element/tag association is stored in a dedicated table. Which means that when we want to use rag on a new element (see recent KM module), we must create and use a new table. Having a single table to store this will make it easier to expand the tag system on any element (ECM, tasks, proposals, ...)

## Suggested steps
- [x] Create a table "element_tag" (rowid, fk_category, fk_element)
- [ ] Migrate data from the old tables (societe_category, contact_category, ...) to the new table
- [ ] Rewrite the category class to use the new table
- [ ] Move the tag management menu to the "Tools" main menu to manage all tags from there instead of in all menu
- [ ] Add a module_part "tags" to define if a module uses the tag system

## To go further
Ideas to discuss to improve the tag module :
- Allow to create a tag while creating/editing an object
- ...

lmag:
@evarisk-theo @nicolas-eoxia

evarisk-theo:
I think we should add a element_type column to this table to make this table generic and usable for all objects

edupulpillo:
Great approach

JoAnMaFe:
Certainly, this new approach could simplify the use of categories, but it could result in a very large "single table".

atm-maxime:
> I think we should add a element_type column to this table to make this table generic and usable for all objects

I think this is not necessary because this information is already in llx_categorie with the column type which correspond to the element. But maybe improve this to store element (ie : "societe") instead of a numeric type.

FHenry:
@atm-maxime Did you notice that llx_element_tag was created in install/mysql/migration/3.2.0-3.3.0.sql ? It's only in this migration file.... Should it be remove/rename it in install/mysql/migration/15.0.0-16.0.0.sql before CREATE TABLE llx_element_tag for people that come from 3.2 (few cases sure, but can exists). Did you target to finish Task on this issue for 16.0 ? or later ?

lmag:
@atm-maxime are you interested in working on it : Rewrite the category class to use the new table ?

atm-maxime:
> @atm-maxime are you interested in working on it : Rewrite the category class to use the new table ?

@lmag Interested? Always, that's why I started it 😉. But for now I don't have time to continue the task. Maybe in Nantes :)

[...]

PR #19262, opened 2021-10-31
New : prepare future element_tag table

atm-maxime:
SQL structure for future element_tag table
See #18996


htdocs/install/mysql/tables/llx_element_tag.key.sql
```diff
@@ -0,0 +1,23 @@
...
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU General Public License for more details.
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program. If not, see <https://www.gnu.org/licenses/>.
+--
+-- ============================================================================
+
+ALTER TABLE llx_element_tag ADD PRIMARY KEY pk_element_tag (fk_categorie, fk_socpeople);
```
eldy, line 19: column fk_socpeople does not exists.

atm-maxime:
I checked with other tables (element_contact or element_resources) and the column name is "element_id". Do you want me to replace fk_element by element_id ?

htdocs/install/mysql/tables/llx_element_tag.key.sql
```diff
@@ -0,0 +1,23 @@
...
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU General Public License for more details.
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program. If not, see <https://www.gnu.org/licenses/>.
+--
+-- ============================================================================
+
+ALTER TABLE llx_element_tag ADD PRIMARY KEY pk_element_tag (fk_categorie, fk_element);
```
eldy, line 19: Primary key must be on any table on rowid so
rowid           	integer AUTO_INCREMENT PRIMARY KEY,
So we have a working autoincrement.
To have then a unique value on the couple, add an index like this
ALTER TABLE llx_element_tag ADD UNIQUE INDEX idx_element_tag_uk (fk_category, fk_element);

htdocs/install/mysql/tables/llx_element_tag.key.sql
```diff
@@ -0,0 +1,23 @@
...
+ALTER TABLE llx_element_tag ADD INDEX idx_element_tag_fk_categorie (fk_categorie);
```
eldy, line 20: Because you already have an index on couple   fk_categorie, fk_element, an index on fk_categorie alone is useless (fk_categorie is already the begining of index on the couple)

atm-maxime:
Ok done ! I know now that the rowid column must always be present. It's not like this currently in llx_categorie_xxx tables, that's why I made the mistake.
Ok for the unique key replacing the index.

**Should I leave _fk_element_ or do you prefer _element_id_ ?**

PR #19275, opened 2021-11-01
fix error

frederic34:
image


htdocs/core/modules/printing/modules_printing.php
```diff
@@ -69,7 +69,7 @@ public static function listDrivers($db, $maxfilenamelength = 0)
 		$listoffiles = array();
 		$dirmodels = array_merge(array('/core/modules/printing/'), (array) $conf->modules_parts['printing']);
 		foreach ($dirmodels as $dir) {
-			$tmpfiles = dol_dir_list(dol_buildpath($dir, 0), 'all', 0, '\modules.php', '', 'name', SORT_ASC, 0);
+			$tmpfiles = dol_dir_list(dol_buildpath($dir, 0), 'all', 0, '.modules.php', '', 'name', SORT_ASC, 0);
```
eldy, line 72: I think it should be
'\/modules.php'
don't you think so ?

frederic34:
tested with php7 and php8

htdocs/core/modules/printing/modules_printing.php
frederic34: we need to find `/core/modules/printing/`abcdef`.modules.php`

Issue #2737, opened 2015-04-30
Log for edit bank account's thirdparty information


braito4:
the edition of the bank account information belonged to a third party should have a log.
In case this changes cause a issue later

PR #19281, opened 2021-11-02
NEW Thirdparty - Add advanced right to add / edit payment information

aspangaro:
Add an advanced right to anticipe this type of scam : 

Information (French) : https://www.quechoisir.org/actualite-arnaque-en-ligne-le-faux-rib-fait-irruption-dans-les-boites-mail-n94380/

Before : 

image

After without advanced right : 

image

image


braito4:
Also a log of who made the changes should be great:
https://github.com/Dolibarr/dolibarr/issues/2737

aspangaro:
Good idea @braito4 , I'll see what I can do

htdocs/core/modules/modSociete.class.php
```diff
@@ -197,6 +197,13 @@ public function __construct($db)
 		$this->rights[$r][3] = 0; // La permission est-elle une permission par defaut
 		$this->rights[$r][4] = 'export';
 
+		$r++;
+		$this->rights[$r][0] = 130;
+		$this->rights[$r][1] = 'Modify thirdparty information payment';
+		$this->rights[$r][3] = 0;
+		$this->rights[$r][4] = 'addupdate_thirdparty_payment_information';      // Visible if option MAIN_USE_ADVANCED_PERMS is on
```
eldy, line 204: The key must end with   "_advance" when we add and advanced permission.
I suggest   "thirdparty_paymentinformation_advance"    (the addupdate is useless, is is already into the "write" of the next field).

eldy:
Having a log may be implemented with a trigger of agenda module (that track business event action) of payment update to add a line to save old and new value into an automatic event.

PR #19284, opened 2021-11-02
Supplier Invoice Payment - Re-order the "Payment amount, original currency" to be together with the columns of original currency

henrynopo:
The current coding put the "Payment amount, original currency" at the 1st column on the right, away from the columns of original currency.

This PR re-adjust its order and group all the columns of "original currency" together. This will harmonise the interface of payment page for supplier invoice and customer invoice. The result is shown as below:
image


htdocs/fourn/facture/paiement.php
```diff
@@ -555,14 +555,12 @@ function callForResult(imgId)
 							print '<td class="right">'.$langs->trans('MulticurrencyAmountTTC').'</td>';
 							print '<td class="right">'.$langs->trans('MulticurrencyAlreadyPaid').'</td>';
 							print '<td class="right">'.$langs->trans('MulticurrencyRemainderToPay').'</td>';
+							print '<td class="center">'.$langs->trans('MulticurrencyPaymentAmount').'</td>';
```
eldy, line 558: You should use add here
`if (!empty($conf->multicurrency->enabled)) {`
If module is off, we should not get any feature relative to multicurrency

PR #19293, opened 2021-11-02
CLOSE #19292 Add access right and Product member to read stock availability

### LINKED ISSUE #19292: Stock: Add access right and Product member to read stock availability (or lack thereof)
# Feature Request
Add an access right and two fields in the Product class to allow an user to see stock availability when he doesn't have stock reading permission, but he does have stock availability reading permission.

## Use case
This would allow an external user (for example) to see through the web UI if a product is available or not without giving him too much information about real stock numbers. This would allow the same through the REST API, with further freedom on use case (displaying or not, etc).

lainwir3d:
# CLOSE #19292
Add an access right and two fields in the Product class to allow an user to see stock availability when he doesn't have stock reading permission, but he does have stock availability reading permission.

### DIFF
--- a/htdocs/langs/en_US/stocks.lang
+++ b/htdocs/langs/en_US/stocks.lang
@@ -185,6 +185,7 @@ inventoryReadPermission=View inventories
 inventoryWritePermission=Update inventories
 inventoryValidatePermission=Validate inventory
 inventoryDeletePermission=Delete inventory
+stockAvailabilityReadPermission=See stock availability
 inventoryTitle=Inventory
 inventoryListTitle=Inventories
 inventoryListEmpty=No inventory in progress
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -2997,7 +2997,7 @@ class Form
 
 					unset($tmpproduct);
 				}
-			}else if (!empty($user->rights->stock->availability->read)) {
+			} elseif (!empty($user->rights->stock->availability->read)) {
 				if ($objp->stock > 0) {
 					$opt .= ' - '.$langs->trans("Stock").': '.$langs->trans("Available");
 					$outval .= ' - <span class="product_line_stock_ok">'.$langs->transnoentities("Stock").': '.$langs->trans("Available");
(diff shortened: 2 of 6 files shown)


htdocs/langs/en_US/stocks.lang
```diff
@@ -185,6 +185,7 @@ inventoryReadPermission=View inventories
 inventoryWritePermission=Update inventories
 inventoryValidatePermission=Validate inventory
 inventoryDeletePermission=Delete inventory
+stockAvailabilityReadPermission=See stock availability
```
eldy, line 188: What is the use case to have a user able to read the stock but without knowing stock availability of a product ?

lainwir3d: True, an user with stock read availability should have its availability reading right checked automatically. Is there a way to do that? I guess through a trigger?

But the inverse is not true, I want some external user to be able to tell if there is some stock available without being able to have the exact stock numbers (as its private). 

The real use case behind this, is by using the REST API, I want to prevent product ordering from a customer (identified as an external user) by hiding the "add to cart" if there is no stock available, but I don't want a clever user to make the REST API call on its own using its key and see the stock numbers other than "available". The product should still be displayed though for later reference.

Do you think there might be a better way around this?

Cheers

lainwir3d: As an example:

image

lainwir3d: As an example:

image

lainwir3d: As an example:

image

eldy:
"I want some external user to be able to tell if there is some stock available without being able to have the exact stock numbers (as its private)."
If the user need to add n products in cart, he must know if there is n available. And if there is an api to know that, user can also know what is exact stock. He just have to call the api several time by increasing n and when api say there is not enough, we know we have the real stock.

So such permission would be useless into core. I recommend to develop your own api if you need a special number of stock being reported that is not the real stock.

lainwir3d:
That's true that if he truly is "smart", he'll find a way to find the real stock values, but this will still stop "classic" users from finding out (though they probably won't even know what a REST API is...)

I still find this would be useful, but the access right is confusing into core, I'll develop my own module.

Can I put this in a "computed" extrafields? Never used that kind of extrafield. Would something like a boolean extrafield with computed field as "$object->stock > 0" work?

Cheers.

eldy:
Extrafields with  "$object->stock > 0"  would probably not work, because  "$object->stock" is not a real data (it is a dernormalized value for performance purpose). The real value of stock is the sum of qty found into table of stock.

PR #19304, opened 2021-11-03
 Fix default enddate subscription must be in futur

daraelmin:
Currently, the first proposed date for a new subscription is the date immediately following the end of the last contribution. However, the proposed subscription may already have expired if a member has not renewed for longer than the default period. 
My proposition is to use the first day of the present year, as the "Now link" is already enabled.

### DIFF
--- a/htdocs/adherents/subscription.php
+++ b/htdocs/adherents/subscription.php
@@ -943,8 +943,10 @@ if ($rowid > 0) {
 		}
 		if (!$datefrom) {
 			$datefrom = $object->datevalid;
-			if ($object->datefin > 0) {
-				$datefrom = dol_time_plus_duree($object->datefin, 1, 'd');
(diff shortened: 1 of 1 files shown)


htdocs/adherents/subscription.php
```diff
@@ -943,8 +943,10 @@
 		}
 		if (!$datefrom) {
 			$datefrom = $object->datevalid;
-			if ($object->datefin > 0) {
+			if ($object->datefin > 0 && dol_now() < dol_time_plus_duree(dol_time_plus_duree($object->datefin, 2 * $defaultdelay, $defaultdelayunit), -1, 'd')) {
```
eldy, line 946: I think you should do
 if ($object->datefin > 0 && dol_time_plus_duree($object->datefin, $defaultdelay, $defaultdelayunit) < dol_now()) {

daraelmin: Right ! 

Done.

daraelmin:
Change done as you requested

PR #19314, opened 2021-11-04
redirect to view screen on element update

cfoellmann:
This fixes the page you get redirected to when updating an element.
The resulting url should be the card + id parameter of the element and a clean $_POST

### DIFF
--- a/htdocs/core/actions_addupdatedelete.inc.php
+++ b/htdocs/core/actions_addupdatedelete.inc.php
@@ -254,7 +254,10 @@ if ($action == 'update' && !empty($permissiontoadd)) {
 	if (!$error) {
 		$result = $object->update($user);
 		if ($result > 0) {
-			$action = 'view';
+			$urltogo = $backtopage ? str_replace('__ID__', $result, $backtopage) : $backurlforlist;
+			$urltogo = preg_replace('/--IDFORBACKTOPAGE--/', $object->id, $urltogo); // New method to autoselect project after a New on another form object creation
+			header("Location: ".$urltogo);
+			exit;
 		} else {
 			$error++;
 			// Creation KO


htdocs/core/actions_addupdatedelete.inc.php
```diff
@@ -253,7 +253,10 @@
 	if (!$error) {
 		$result = $object->update($user);
 		if ($result > 0) {
-			$action = 'view';
+			$urltogo = $backtopage ? str_replace('__ID__', $result, $backtopage) : $backurlforlist;
```
eldy, line 256: In common cases, we should not have to make a redirect after an update.
The $object should be already up to date (due to ->update), so output the page with action = view should be enough.
In which use case,do you need a redirect ?

cfoellmann: without the header redirect the $_POST is not cleared. That way a user can not refresh the page or move back and forward without issue

eldy: For which element, which url update can you reproduce this ?

cfoellmann: /projet/card.php?id=13 -> Edit -> Cancel -> /projet/card.php?id=13 | OK
/projet/card.php?id=13 -> Edit -> Save -> /projet/card.php | NOK

eldy:
You say 
/projet/card.php?id=13 -> Edit -> Save -> /projet/card.php | NOK
I can't reproduce any trouble. Can you tell me which data is not correctly updated/visible after an Edit / Save of a project ?

cfoellmann:
Problem is that the GET parameter gets lost.

eldy:
The id is not propagated as GET but should be as POST.  So for the GETPOST('id', 'int'), it should be the same.
I still did not succeed in finding any bug on this part. Can you describe how you reproduced the bug and what is wrong ?

eldy:
I can't reproduce.
This bug or the feature request seems to be fixed or included into last recent stable version, so I close the issue. If not thanks to reopen.

Issue #19320, opened 2021-11-05
Redirect after connection in other page then index page


nicolas-eoxia:
# Feature Request
It would be interesting to be redirected to any page when connecting to Dolibarr.

PR #19321, opened 2021-11-05
NEW : redirect after connection in login page #19320 

nicolas-eoxia:
#19320


hregis:
@nicolas-eoxia hello, use "switchentity" instead "entity" to avoid conflict, thanks

$parameters = array('switchentity' => GETPOST('switchentity', 'int'));

PR #19324, opened 2021-11-05
FIX #15273 : Provide correct remain to pay and already payed

### LINKED ISSUE #15273: {object_remain_to_pay_locale} / {object_remain_to_pay} doesn't propose remain to pay
# Instructions
To reproduce use odt template and {object_remain_to_pay_locale} tag

# Bug
I use an odt template to generate invoice. I use {object_remain_to_pay_locale} to display remain to pay. But it looks to sum {remain_to_pay} with {object_total_ttc}.
Should be a difference not a sum.

## Environment
- **Version**: 10.0.7
- **OS**: Debian
- **Web server**: apache 2.4
- **PHP**: 
- **Database**: Mariadb
- **URL(s)**: Not relevant

camlafit:
Provide a correct remain to pay and already payed value in each invoice. 
sumcreditnote and sumdeposit are negative value then we must minus them from each total.

Before :
Capture d’écran de 2021-11-05 18-32-01

After : 
Capture d’écran de 2021-11-05 18-30-48

### DIFF
--- a/htdocs/core/class/commondocgenerator.class.php
+++ b/htdocs/core/class/commondocgenerator.class.php
@@ -408,8 +408,8 @@ abstract class CommonDocGenerator
 			$sumpayed = $object->getSommePaiement();
 			$sumdeposit = $object->getSumDepositsUsed();
 			$sumcreditnote = $object->getSumCreditNotesUsed();
-			$already_payed_all = $sumpayed + $sumdeposit + $sumcreditnote;
-			$remain_to_pay = $sumpayed - $sumdeposit - $sumcreditnote;
+			$already_payed_all = $sumpayed - $sumdeposit - $sumcreditnote;
+			$remain_to_pay = $sumpayed + $sumdeposit + $sumcreditnote;
 
 			if ($object->fk_account > 0) {
 				require_once DOL_DOCUMENT_ROOT.'/compta/bank/class/account.class.php';


htdocs/core/class/commondocgenerator.class.php
```diff
@@ -408,8 +408,8 @@ public function get_substitutionarray_object($object, $outputlangs, $array_key =
 			$sumpayed = $object->getSommePaiement();
 			$sumdeposit = $object->getSumDepositsUsed();
 			$sumcreditnote = $object->getSumCreditNotesUsed();
-			$already_payed_all = $sumpayed + $sumdeposit + $sumcreditnote;
-			$remain_to_pay = $sumpayed - $sumdeposit - $sumcreditnote;
+			$already_payed_all = $sumpayed - $sumdeposit - $sumcreditnote;
```
eldy, line 411: The already_payed_all should be the sum of all. So old code seems good.
Can you reproduce the bug on demo.dolibarr.org ?

camlafit: Hello

Behavior is different between my instance and demo.
In my demo "remise" is set a standard payment but in demo is set as a negative service

Demo : new service row with negative amount
Capture d’écran de 2021-11-08 09-36-06

Prod : new payment line to report acompte invoice

eldy: You probably enabled the option FACTURE_DEPOSITS_ARE_JUST_PAYMENTS ?
This option is an old an bugged implementation that should not be used. There is ton of bug with it and it is compliant with legal rules of accountancy.

camlafit:
Hello

Indeed we use this feature : FACTURE_DEPOSITS_ARE_JUST_PAYMENTS
Then if is no more maintained how to get a similar behavior ?

Have a nice day

PR #19333, opened 2021-11-08
Load product data optional fields to the line

BitKFu:
# Optional Fields
This fix enables to use optional fields from product in invoices or offers at the line.

### DIFF
--- a/htdocs/core/class/commondocgenerator.class.php
+++ b/htdocs/core/class/commondocgenerator.class.php
@@ -559,6 +559,15 @@ abstract class CommonDocGenerator
 			$resarray = $this->fill_substitutionarray_with_extrafields($object, $resarray, $extrafields, $array_key, $outputlangs);
 		}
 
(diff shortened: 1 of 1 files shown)


htdocs/core/class/commondocgenerator.class.php
```diff
@@ -559,6 +559,14 @@ public function get_substitutionarray_object($object, $outputlangs, $array_key =
 			$resarray = $this->fill_substitutionarray_with_extrafields($object, $resarray, $extrafields, $array_key, $outputlangs);
 		}
 
+		// Load product data optional fields to the line -> enables to use "line_options_{extrafield}"
```
eldy, line 562: $line here is a random value so not reliable (it as value of the last line defined by loop on line 519).

BitKFu: Yes, I see.
Thanks for the reply. 

I finally found out that the functionality was already implemented in get_substitutionarray_lines (line 669).

You can close the pull request.

PR #19335, opened 2021-11-08
FIX - There is no way to set some PDF config variable + Incomplete \n conversion if HTML tag are used

fr69400:
1°) PDF Lib use global config variable  PDF_BOLD_PRODUCT_LABEL and PDF_BOLD_PRODUCT_REF_AND_PERIOD but there is no way to set theses variable on config page
2°) Incomplete \n conversion if HTML tag are used in PDF
3°) Fix some warning (array is required but null is given)


htdocs/core/lib/pdf.lib.php
```diff
@@ -1553,6 +1553,7 @@ function pdf_getlinedesc($object, $i, $outputlangs, $hideref = 0, $hidedesc = 0,
 
 	// Now we convert \n into br
 	if (dol_textishtml($libelleproduitservice)) {
+		$libelleproduitservice = str_replace("\n", '<br/>', $libelleproduitservice);
```
eldy, line 1556: If your text $libelleproduitservice  is a html text, the \n into it is not a new line, but something that must not appears on PDF.
If $libelleproduitservice has a HTML format, and need a new line into content, the <br> must be inside the content $libelleproduitservice.  You can't have some part of var that is html and some that are not. All must be non html or all must use the html syntax.

fr69400: Not sure...

**First Example :**
1°) Create a product with a multilined description
image

The description use \n

2°) Create a propal for example with this product and generate the pdf
image

Its'ok because the is no HTML tag

3°) Then set the variable PDF_BOLD_PRODUCT_LABEL and generate again the PDF

image

The result is bad, because Product label is bolded width <b></b> but product description still use \n

**The second one :**
Set the variable SHOW_SUBPRODUCT_REF_IN_PDF
The result may change if you bold or not a line ... bescause in the pdf lib use \n in the code

image

Same things with the variable MAIN_DOCUMENTS_DESCRIPTION_FIRST. The result is different if you bold something or not



The other way i see to solve this problem is to replace all valid \n char to __N__ 
For exempla you have 

image


$desc must be checked.... Because it may use valid \n

image

The same kind of things must be done in some other lines of the code

fr69400: I have updated code to force valid \n to be replaced by __N__

eldy:
This is reason wyh PDF_BOLD_PRODUCT_LABEL is not a supported variable for the moment.
It introduces html tag into a string making a string moved from text to html without converting also the \n into a br. The bug is inside the PDF_BOLD_PRODUCT_LABEL management.

PR #19342, opened 2021-11-09
Fix ldap issues with activedirectory (Samba 4 AD)

ibuiv:
# FIX Some LDAP issues with activedirectory

Mode : dolibarr2ldap with Samba 4 AD (type : activedirectory)
- Change userAccountControl when Enabling or Disabling User
- Sync userAccountControl when User sync 
- Store password if user is modified or Password is generated. (Config LDAP Module -> User Mapping -> Password Not Encrypted -> unicodePwd
- Add warning when User sync : MUST change Password
- Unset member entry if empty while creating usergroup

WARNINGS : 
- Add a new entry in ldap.lang for fr_FR and en_US only
- Commit 8361ea8e074f2af704702e9e8c1e951d9d9b36b9 was not attended (a file added for eventorganization module). Do not know how to remove this commit.


htdocs/user/group/ldap.php
```diff
@@ -75,6 +75,12 @@
 		$dn = $object->_load_ldap_dn($info);
 		$olddn = $dn; // We can say that old dn = dn as we force synchro
 
+		if ($ldap->serverType == "activedirectory") {
+			$info['sAMAccountName'] = $object->name;
+		}
+
+		unset($info['member']);
```
eldy, line 82: Are you sure we must unset this all the times or only if ($ldap->serverType == "activedirectory") {

ibuiv: The line was not intended to be in this pull request.
Removed it. 
Thanks for the review and sorry for the time loss.

eldy:
@hregis I think you use activedirectory. Do you think it's good to introduce code with test on hard coded value like
`$info['userAccountControl'] = 512; //Account enabled`

PR #19345, opened 2021-11-09
FIX lost superadmin grade after edit user card

hregis:


htdocs/user/card.php
```diff
@@ -450,11 +450,15 @@
 				// Do we update also ->entity ?
 				if (!empty($conf->multicompany->enabled)) {	// If multicompany is not enabled, we never update the entity of a user.
 					if (!empty($conf->global->MULTICOMPANY_TRANSVERSE_MODE)) {
-						$object->entity = 1; // all users are in master entity
+						if (GETPOST('superadmin', 'int')) {
+							$object->entity = 0;
```
eldy, line 454: Relying on a property provided as a parameter to decide if an object must be into an admin entity or not means anybody, even not superadmin can set the superadmin flag to any user.

hregis: @eldy 
I check this in the update function of the class user

hregis:
@eldy
This correction is very important because a superadmin that edits its user profile or any superadmin profile downgrades the profile to a simple administrator!

htdocs/user/card.php
```diff
@@ -448,19 +448,15 @@
 				$object->lang = GETPOST('default_lang', 'aZ09');
 
 				// Do we update also ->entity ?
-				if (!empty($conf->multicompany->enabled)) {	// If multicompany is not enabled, we never update the entity of a user.
-					if (!empty($conf->global->MULTICOMPANY_TRANSVERSE_MODE)) {
-						$object->entity = 1; // all users are in master entity
+				if (!empty($conf->multicompany->enabled && $user->entity == 0 && !empty($user->admin))) {	// If multicompany is not enabled, we never update the entity of a user.
+					if (GETPOST('superadmin', 'int')) {
```
eldy, line 452: We should not rely on a value from GETPOST to force the entity.
According to your comment "// If multicompany is not enabled, we never update the entity of a user.", if this is true, why no just commenting the line 
$object->entity = 0;
If we must never update the field, we must just not set the value when we are in such a case.

PR #19352, opened 2021-11-11
improve addline + updateline form submission

cfoellmann:
- after addline = success -> clear $_POST + redirect to <url>#addline (jump to addline form again)

- after updateline = success -> clear $_POST + keep redirect to updated line

### DIFF
--- a/htdocs/comm/propal/card.php
+++ b/htdocs/comm/propal/card.php
@@ -1163,38 +1163,9 @@ if (empty($reshook)) {
 						$object->generateDocument($object->model_pdf, $outputlangs, $hidedetails, $hidedesc, $hideref);
 					}
 
-					unset($_POST['prod_entry_mode']);
-
-					unset($_POST['qty']);
-					unset($_POST['type']);
-					unset($_POST['remise_percent']);
-					unset($_POST['price_ht']);
-					unset($_POST['multicurrency_price_ht']);
-					unset($_POST['price_ttc']);
-					unset($_POST['tva_tx']);
-					unset($_POST['product_ref']);
-					unset($_POST['product_label']);
-					unset($_POST['product_desc']);
-					unset($_POST['fournprice']);
-					unset($_POST['buying_price']);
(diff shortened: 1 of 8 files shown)


htdocs/comm/propal/card.php
```diff
@@ -1163,38 +1163,9 @@
...
-					unset($_POST['date_startday']);
-					unset($_POST['date_startmonth']);
-					unset($_POST['date_startyear']);
-					unset($_POST['date_endhour']);
-					unset($_POST['date_endmin']);
-					unset($_POST['date_endsec']);
-					unset($_POST['date_endday']);
-					unset($_POST['date_endmonth']);
-					unset($_POST['date_endyear']);
+					$urltogo = preg_replace('/--IDFORBACKTOPAGE--/', $object->id, $urltogo) . '#addline';
```
eldy, line 1166: I am not a big fan of making redirect when it is not absolutely necessary. It increase the "flash" effect on slow servers, making a bad experience for end user.
Why not instead, introducing a function unsetPostAfterInsertLine() ?

cfoellmann:
in the current state of the code the page does NOT jump down to the addline form after submission! Bad UX

Are you sure that this header location does create an additional pageload? I am pretty sure that it does not. Otherwise the unset of POST would not work for the user!

eldy:
Yes, i am sure that adding a header with a redirect create an overload, because it returns a result to the browser to say to the browser to ask another page. So there is on answer and one request more on the network.

To jump to the addline form (and retreive same position than before submit), we must use a generic feature:
- add the class "reposition" on the CSS of the button (so page position is restored at same place that before clicking "submit". It is better than using the #abc (position differs a little bit from position of button).
- It works magically if the css was added on a href link. But if it was added on a button of a form, you must also check that a hidden field called "page_y" is inside the form and it will work too.
`print '<input type="hidden" name="page_y" value="" />';`

eldy:

cfoellmann:
I will try to validate and create a new PR for anything that is still needed

Issue #19360, opened 2021-11-11
NEW Stripe Payment Element


ptibogxiv:
Replace actual dolibarr"s stripe payment integration by this

https://stripe.com/docs/payments/payment-element

in order to support more gateways

colabware-app:
Hey @ptibogxiv, it’s hard finding open-source contributors to build your features for free. By [submitting your request through Colabware](https://colabware.co/project/1637415975874x610736382815988200), harness the power of crowdfunding to encourage developers, whilst enabling others to support your vision. Please do let us know if you have any questions, and I hope we can assist with your project!

PR #19362, opened 2021-11-12
NEW stripe element with more gateways

ptibogxiv:
cf #https://github.com/Dolibarr/dolibarr/issues/19360

more doc here
https://stripe.com/docs/payments/payment-element

actually functionnal but car be optimized with less fields and better design

### DIFF
--- a/htdocs/public/payment/newpayment.php
+++ b/htdocs/public/payment/newpayment.php
@@ -3,7 +3,7 @@
  * Copyright (C) 2006-2017	Laurent Destailleur		<eldy@users.sourceforge.net>
  * Copyright (C) 2009-2012	Regis Houssin			<regis.houssin@inodbox.com>
  * Copyright (C) 2018	    Juanjo Menent			<jmenent@2byte.es>
- * Copyright (C) 2018-2019	Thibault FOUCART	    <support@ptibogxiv.net>
+ * Copyright (C) 2018-2021	Thibault FOUCART	    <support@ptibogxiv.net>
  * Copyright (C) 2021		Waël Almoman	    	<info@almoman.com>
  * Copyright (C) 2021		Dorian Vabre			<dorian.vabre@gmail.com>
  *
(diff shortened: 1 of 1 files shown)


htdocs/public/payment/newpayment.php
```diff
@@ -2423,49 +2422,28 @@
...
 
-					stripe.handleCardPayment(
-					clientSecret, cardElement, {
+					stripe.confirmPayment({
+		  elements,confirmParams: {
+		  return_url: '<?php echo $urlok; ?>',
 						payment_method_data: {
 							billing_details: {
-								name: cardholderName.value
+								name: 'test'
```
eldy, line 2446: This looks strange. Does this means we do not propagate the name of credit card owner anymore ?

ptibogxiv: @eldy nope! The New element  form don’t use it

eldy:
I modified your PR so to enable your new feature you must set 
STRIPE_USE_INTENT_WITH_AUTOMATIC_CONFIRMATION to 2 instead of 1
So the old code when STRIPE_USE_INTENT_WITH_AUTOMATIC_CONFIRMATION = 1 is still available.
And I keep this old code as the default value because the new one seems to not work correctly. I am investigating on it...

ptibogxiv:
llok an eye on https://github.com/Dolibarr/dolibarr/pull/19444 i fix incompatibility between gateways...

eldy:
This is not a problem in Stripe nor in TakePOS, but a problem in the new code implementation:
Take the new code, set STRIPE_USE_INTENT_WITH_AUTOMATIC_CONFIRMATION to 2 to use the new Stripe system. Then try to make a payment of a simple invoice using the link suggested on the invoice:
With old method (STRIPE_USE_INTENT_WITH_AUTOMATIC_CONFIRMATION = 1), after a payment is done the page newpayment.php was called with parameter action=charge and this page was sets sessions variables and then make a redirect to ko or ko page.
With the new one (STRIPE_USE_INTENT_WITH_AUTOMATIC_CONFIRMATION = 2), once a payment is done, the page  ok is called directly without the object sessions required by the paymentok page (data are not set into session like it is into the action=charge when STRIPE_USE_INTENT_WITH_AUTOMATIC_CONFIRMATION=1), so the paymentok page does nothing instead of recording the payment and other actions. So payment is validated into Stripe but not into Dolibarr !

eldy:
I pushed a fix. I let you test that everything is working as expected...

PR #19370, opened 2021-11-12
ADD new stripe provider in dictionnary

ptibogxiv:
For  better informations in Dolibarr and on object/pdf (order, invoice....)


htdocs/install/mysql/migration/14.0.0-15.0.0.sql
```diff
@@ -448,3 +448,8 @@ ALTER TABLE llx_socpeople DROP COLUMN googleplus;
 ALTER TABLE llx_socpeople DROP COLUMN youtube;
 ALTER TABLE llx_socpeople DROP COLUMN whatsapp;
 
+INSERT INTO llx_c_paiement (id,code,libelle,type,active) values (100, 'KLA', 'Klarna', 1, 1);
```
eldy, line 451: If you add it into migration, you must also add it into the mysql/data/llx_c_paiement.sql file
Also, the id must not be the same value for each entry. Use 100, 101, 102...

ptibogxiv: @eldy fix done

PR #19393, opened 2021-11-16
NEW: Add option MAIN_ADD_LINE_AT_POSITION to choose position of line during insert

atm-gauthier:
When MAIN_VIEW_LINE_NUMBER is enabled, we can now choose in which position we want to add a new line (available on propal, order, invoice, propal fourn, order fourn, invoice fourn)

We select position :
image

And then it's added on selected line
image

By default, lines are added at the end as well as current workflow :
image


atm-gauthier:
Hi @eldy, for my feature the "for" loop seems to be the most appropriate but i get errors like this :

image

What's your opinion about that ?

atm-maxime:
Suggestions :
Use an input type number instead of a drop-down, because if your document has 100 lines, a drop-down will be hard to use.
The keyword OnLine for the translation might be wrongly understood, use AddLineOnPosition instead.

atm-gauthier:
@atm-maxime i agree with your proposition, i'm just waiting for @eldy feedback.
If he is ok with this feature, i will replace select by input number

atm-gauthier:
Hi @eldy, i finally did what @atm-maxime said, it's now an input number :

image

htdocs/comm/propal/class/propal.class.php
```diff
@@ -734,6 +735,10 @@ public function addline($desc, $pu_ht, $qty, $txtva, $txlocaltax1 = 0.0, $txloca
 				// Reorder if child line
 				if (!empty($fk_parent_line)) {
 					$this->line_order(true, 'DESC');
+				} elseif ($ranktouse > 0 && $ranktouse <= count($this->lines)) { // Update all rank of all other lines
+					for ($ii = $ranktouse; $ii <= count($this->lines); $ii++) {
```
frederic34, line 739: `$linecount = count($this->lines);
for ($ii = $ranktouse; $ii <= $linecount; $ii++) {`

atm-gauthier: @frederic34 you were right, it's ok now with my PR, thank you very much ! :)

atm-gauthier:
I did a new version with input number :

image

htdocs/core/tpl/objectline_create.tpl.php
```diff
@@ -309,6 +310,11 @@
 			echo '<input type="hidden" name="pbq" id="pbq" value="">';
 			echo '</span>';
 		}
+
+		if (!empty($conf->global->MAIN_VIEW_LINE_NUMBER)) {
```
eldy, line 314: I would suggest to use another constant like
MAIN_ADD_LINE_NUMBER
and keep MAIN_VIEW_LINE_NUMBER for current behaviour (read only the column)

atm-gauthier: Hi @eldy, i did it and created hidden conf "MAIN_ADD_LINE_AT_POSITION"

atm-gauthier:
@eldy i did modification you asked, can you merge this PR ?
I often have to re-merge develop in my branch... ;)

PR #19395, opened 2021-11-16
NEW show selected invoice only on paiement form

lvessiller-opendsi:
NEW show selected invoice only on paiement form
- show only one invoice if "fac_only" parameter is set
- keep the default behavour in other cases


lvessiller-opendsi:
I close this request because functionally it's not used.
We have to see all invoices linked to a third party even if there is a filter on the "facid" field.

PR #19396, opened 2021-11-16
fix: #ZENDSI fix margin calc with credit note

FHenry:
fix: #ZENDSI fix margin calc with credit note

### DIFF
--- a/htdocs/accountancy/customer/index.php
+++ b/htdocs/accountancy/customer/index.php
@@ -526,13 +526,14 @@ if ($conf->global->MAIN_FEATURES_LEVEL > 0) { // This part of code looks strange
 			if ($j > 12) {
 				$j -= 12;
 			}
-			$sql .= ' SUM(CASE WHEN
-							MONTH(f.datef)='.$j.'
-						THEN (CASE WHEN	fd.total_ht < 0
-							  THEN (-1 * (abs(fd.total_ht) - (fd.buy_price_ht * fd.qty * (fd.situation_percent / 100))))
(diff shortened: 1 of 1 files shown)


htdocs/accountancy/customer/index.php
```diff
@@ -520,16 +520,22 @@
...
 		print '<td width="60" class="right"><b>'.$langs->trans("Total").'</b></td></tr>';
-
 		$sql = "SELECT '".$db->escape($langs->trans("Vide"))."' AS marge,";
 		for ($i = 1; $i <= 12; $i++) {
 			$j = $i + ($conf->global->SOCIETE_FISCAL_MONTH_START ? $conf->global->SOCIETE_FISCAL_MONTH_START : 1) - 1;
 			if ($j > 12) {
 				$j -= 12;
 			}
-			$sql .= "  SUM(".$db->ifsql('MONTH(f.datef)='.$j, '(fd.total_ht-(fd.qty * fd.buy_price_ht))', '0').") AS month".str_pad($j, 2, '0', STR_PAD_LEFT).",";
+			$sql .= ' SUM(CASE WHEN
```
eldy, line 529: I have fixed the mysql driver to use the "CASE WHEN" when we do a ->ifsql instead of so proprietary form "IF..."
So now, you should be able to implement this request calling the 2 ifsql :

`"SUM(".$db->ifsql('MONTH(f.datef)='.$j, .$db->ifsql('fd.total_ht <0', ...`

Can you try to restore the use of ->ifsql instead of hardcoded CASE ...

FHenry: @eldy Done

PR #19399, opened 2021-11-17
NEW : allow specific tpl for stock correction and stock transfer

altairis-tof:
dol_include_once for tpl files


altairis-tof:
dol_include_once not working for tpl

PR #19402, opened 2021-11-17
FIX : reorder all lines when delete document line everywhere (4 documents were missing)

atm-gauthier:


htdocs/commande/card.php
```diff
@@ -216,6 +216,8 @@
 		$result = $object->deleteline($user, $lineid);
 		if ($result > 0)
 		{
+			// reorder lines
+			$object->line_order(true);
```
eldy, line 220: After a deletion, sort sort criteria should remains good. 
Why do you need this ?

atm-gauthier: Hi @eldy, it's about following PR : https://github.com/Dolibarr/dolibarr/pull/19393
For this feature to work properly, we must have no break in lines rank numerotation.
Without "$object->line_order(true);" we can get rank list like 1,2,3,5,6,7 after line 4 has been deleted, then for my feature i always need to get rank list like 1,2,3,4,5,6 after line 4 deletion.

atm-gauthier: And to be more precise, i just added the same code that is is already present in supplier propsal and propal cards !

image

PR #19414, opened 2021-11-18
NEW: Extend substitution tokens for email templates

DrBarns:
# NEW [*Extend substitution tokens for email templates to match better with existing ODT substitions*]
Added Token: __AMOUNT_REMAIN__ (amount to pay reduced by payments, deposits and credit notes)
Added Tokens: __AMOUNT_REMAIN_FORMATED__, __AMOUNT_REMAIN_LOCALE__ (same as __AMOUNT_REMAIN__ but localized)*]

This allows users to use the "correct" stipulated amount of money for a given invoice, order, etc... 
It uses the same statements as the ODT substituion for **object_remain_to_pay** and **object_remain_to_pay_locale**.


DrBarns:
**Travis CI - Failure**

Dont know to satisfy the following Travis CI test: 
"Error Failed to load doc generator with modelpaths=core/modules/usergroup/doc/ - modele=templatenamethadoesnotexist"

I just forked dolibarr and used the Netbeans 12 editor to add my code, pushed by branch and did a pull, where is the culprit?!

htdocs/core/lib/functions.lib.php
```diff
@@ -7294,10 +7294,20 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
...
+			$sumdeposit = $object->getSumDepositsUsed();
+			$sumcreditnote = $object->getSumCreditNotesUsed();
+			$already_payed_all = $sumpayed + $sumdeposit + $sumcreditnote;
+		}
+
 		$substitutionarray['__DATE_YMD__']        = is_object($object) ? (isset($object->date) ? dol_print_date($object->date, 'day', 0, $outputlangs) : null) : '';
 		$substitutionarray['__DATE_DUE_YMD__']    = is_object($object) ? (isset($object->date_lim_reglement) ? dol_print_date($object->date_lim_reglement, 'day', 0, $outputlangs) : null) : '';
 
 		$substitutionarray['__AMOUNT__']          = is_object($object) ? $object->total_ttc : '';
+		$substitutionarray['__AMOUNT_FORMATED__']          = is_object($object) ? ($object->total_ttc ? price($object->total_ttc, 0, $outputlangs, 0, 0, -1, $conf->currency) : null) : '';
```
eldy, line 7308: If price is "formated", you should use -1 as parameter 5th parameter

eldy: The error on travis may come because you make a 
$object->getSommePaiement();
$object->get...
So you call method that does not always exists for some objects. 
Try to test that method exists on object before calling them.

DrBarns: > If price is "formated", you should use -1 as parameter 5th parameter

I've copied the line from version 13 working dolibarr installation.  I'm sorry not using the most current version, fixed in the commit. 

> The error on travis may come because you make a $object->getSommePaiement(); $object->get... So you call method that does not always exists for some objects. Try to test that $object->element == 'facture' before calling them so you are sure methods exists.

Used "instanceof" to better support polymorphism, fixed in the commit

DrBarns:
Fixed the code as requested , thank you for the hint.

PR #19417, opened 2021-11-19
FIX : Supplier order stats 

atm-adrien:
## FIX : Supplier order stats 

There was an error when filter on state (with the multiselect)


htdocs/commande/stats/index.php
```diff
@@ -46,7 +46,9 @@
 	accessforbidden();
 }
 
-$object_status = GETPOST('object_status', 'intcomma');
+$object_status = GETPOST('object_status', 'array');
```
eldy, line 49: object_status does not seem to be an array.
I do not experience trouble on page of stats of order, even when using the filter on status

atm-adrien: Done @eldy

PR #19424, opened 2021-11-19
FIX: user employee tab: offset in open days messes up holiday length calculation

marc-dll:


eldy:
I think this bug is already fixed but differently on develop branch. Can you reproduce the bug on develop ? If yes, can you describe your timezone and data you enter to reproduce it on dev ?

marc-dll:
Still the same in develop. I created the following leave request :
Capture d’écran de 2021-12-10 17-42-15

As you can see, as 14/07 is an off day in France, the calculation of 3 days is correct.

The display is OK in the leave request list :
Capture d’écran de 2021-12-10 17-43-44


But in the HR/bank tab, I see the following :
Capture d’écran de 2021-12-10 17-44-00

In Settings > PHP Info, my timezone section is the following :
Capture d’écran de 2021-12-10 17-44-12

PR #19425, opened 2021-11-19
FIX: project timesheet: public holidays offset by 1 day

marc-dll:
+ small unused code cleanup


eldy:
Can you describe how to reproduce the pb ? Is is still bugged into develop (a lot of change were done related to this in v14/develop) ?

marc-dll:
Just tested in latest develop and it is still bugged.

At first, you must have the const `MAIN_DEFAULT_WORKING_DAYS` defined to a correct value (it is empty by default). Otherwise, as pointed out in #19420 , every day will be displayed as a weekend day, in grey (td with class `weekend`), whether or not it is an offday.

Then, for example, offdays will be in yellow but offset (when there are not true weekend days as Christmas and New Year's will be soon).

PR #19426, opened 2021-11-19
NEW: make unicity for user assigned(responsible) on ticket from anyways

FHenry:
NEW: make unicity for user assigned(responsible) on ticket from anyways


eldy:
Can you describe me more the goal of this new feature ? I don't understand by reading the code...

FHenry:
Actually you have two way to assign user on ticket, one into card and antoher into Contact tabs. The goal of this feature is to set internal contact on same value as user assign card field whatever you decide to use Contact (internal) tabs or use assign card field

eldy:
I see. I did some test on v16:  
- When you assign from the main ticket card, it automatically also assign contact on "contact-tab".
- But when you assign from the contact-tab, it does not change nothing on the ticket card.
So I validate PR to see if this fixes the second point...

eldy:
Trouble seems partially fixed after the merge.
If i remove an assignement from the tab "Contact", it is not removed on the ticket card.
Also if i assign 2 guys form contact tab (one contributor, one use rassigned), if i modified use rassigned from the ticket, all are removed, even if the other assignement has the type "Contributor" and not "User assigned". We should removed only link for "User assigned".
Can you check @FHenry the PR merge ?

PR #19444, opened 2021-11-21
NEW support of Stripe Terminal with Takepos

ptibogxiv:
With availability in Europe (France, UK, Ireland...) 
Only support smart readers ie javascript and not only iOs and Android 

@eldy @andreubisquerra


ptibogxiv:
PR functionnal with virtual test reader include in stripe SDK and demo card include in code

ptibogxiv:
Terminal ordered with test card and shipping planned for this week
further development in few days

ptibogxiv:
Just need more informations when error and loading

ptibogxiv:
@eldy @andreubisquerra now is fully fonctionnal! test with Stripe smart reader  bbpos wisepos E

I introduce some notifications for following the actions and errors ie pin error or card declined (can be used by other code)

meuchels:
I added this to my configuration and it works swell but the payment screen "pay.php" still shows "Credit Card" as a payment type. I believe credit card still has to be active for this to work but the extra button on the GUI in TakePOS is confusing.

htdocs/takepos/pay.php
```diff
@@ -51,11 +53,60 @@
...
+	});
+}
+</script>
+<?php }
 
 /*
  * View
  */
 
+if (!empty($conf->stripe->enabled) && (empty($conf->global->STRIPE_LIVE) || GETPOST('forcesandbox', 'alpha'))) {
```
eldy, line 106: If "CASHDESK_ID_BANKACCOUNT_STRIPETERMINAL".$terminaltouse  is not set, it means we didn't configure to use stripe for payment on takepos. Right ?
If right, can you add a condition here to avoid warning when not required.

htdocs/takepos/pay.php
```diff
@@ -314,7 +521,11 @@ function ValidateSumup() {
 	),
 );
 $numpad = $conf->global->TAKEPOS_NUMPAD;
-
+if (!empty($conf->stripe->enabled) && !empty($conf->global->STRIPE_CARD_PRESENT)) {
```
eldy, line 524: Same here: If "CASHDESK_ID_BANKACCOUNT_STRIPETERMINAL".$terminaltouse is not set, we should not output the message.

htdocs/takepos/pay.php
```diff
@@ -421,8 +632,18 @@ function ValidateSumup() {
 	$i = $i + 1;
 }
 
-$keyforsumupbank = "CASHDESK_ID_BANKACCOUNT_SUMUP".$_SESSION["takeposterminal"];
+if (!empty($conf->stripe->enabled) && !empty($conf->global->STRIPE_CARD_PRESENT)) {
```
eldy, line 635: Same. Must be used only if stripe payment enabled for the current terminal $_SESSION["takeposterminal"]

htdocs/takepos/pay.php
eldy: @ptibogxiv Can you check this ?

PR #19446, opened 2021-11-22
FIX empty account selected

inoveaconseil:
In a few case we have a bad type and the test return false, so we cast the variable.


htdocs/core/class/html.formaccounting.class.php
```diff
@@ -329,7 +330,7 @@ public function select_account($selectid, $htmlname = 'account', $showempty = 0,
 					}
 					// Remember guy's we store in database llx_facturedet the rowid of accounting_account and not the account_number
 					// Because same account_number can be share between different accounting_system and do have the same meaning
-					if ($selectid != '' && $selectid == $select_value_in) {
+					if ($selectid != "" && intval($selectid) == intval($select_value_in)) {
```
eldy, line 333: obj->account_number  so selectid can be a string so it should be better to make
`if ($selectid != "" && ((string) $selectid == (string) $select_value_in))`

PR #19459, opened 2021-11-23
NEW can merge products

frederic34:


frederic34:
@eldy any update here ?

Issue #19458, opened 2021-11-23
When you enable a module, store its current version number


atm-florianm:
# What
When a module is initialized, Dolibarr should store its current version number in a standardized way. Some modules already do that, but not all and not all in the same way: it is not automatic and we think it would be a good idea to extend this to all modules.

New methods in the `DolibarrModule` class would also help modules perform some upgrade actions conditioned by the current (stored) version and the target version (the version for which the upgrade action is relevant).

# Why
There are many expected benefits:
- the `init()` method of an external module would be able to enable / disable settings or perform some actions depending on the installed version, and not re-do the same action upon module re-enabling when it is no longer required.
- Dolibarr would be able to warn the user that a module needs to be disabled/re-enabled.
- SQL files could be run only when needed (like column creation, deletion, etc.) without running all of them every time, enabling better error reporting (right now it is not always possible to tell if a query fails because it is not necessary, such as a column that already exists, or if it fails because it has some error).

PR #19461, opened 2021-11-23
keep info on latest activation version of module

frederic34:
image


frederic34:
https://github.com/Dolibarr/dolibarr/issues/19458

atm-maxime:
@atm-florianm @atm-john

PR #19469, opened 2021-11-24
enable societe.class.php in coding test

frederic34:


test/phpunit/CodingPhpTest.php
```diff
@@ -213,12 +213,11 @@ public function testPHP()
 					'multicurrency.class.php',
 					'productbatch.class.php',
 					'reception.class.php',
-					'societe.class.php'
```
eldy, line 216: 👍🏻

PR #19488, opened 2021-11-25
FIX stocks increments on deleting order if $conf->global->STOCK_CALCULATE_ON_VALIDATE_ORDER

ibuiv:
# FIX stocks increments on deleting order if $conf->global->STOCK_CALCULATE_ON_VALIDATE_ORDER == 1

When Deleting an order, if $conf->global->STOCK_CALCULATE_ON_VALIDATE_ORDER == 1, ask for the warehouse to increment the stock if the order was validated before being deleted

### DIFF
--- a/htdocs/commande/card.php
+++ b/htdocs/commande/card.php
@@ -217,7 +217,7 @@ if (empty($reshook)) {
 			}
 		}
 		if (!$error) {
-			$result = $object->delete($user,0,$idwarehouse);
+			$result = $object->delete($user, 0, $idwarehouse);
 			if ($result > 0) {
 				header('Location: list.php?restore_lastsearch_values=1');
 				exit;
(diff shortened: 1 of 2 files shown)


eldy:
I dont' think it is a good idea to make stock change on a deletion.
I think stock change is already done when we do the opposite action of validation, so going back to draft.
But deletion is not the opposite of validation. It means we want to remove completely any data from database (in most cases, this is to make purge and product may have been moved somewhere else since. Also, a lot of use cases won't need that, it will depend on a lot of cases if we want to move stock back or not. Best is to let user decide but in most cases, if it need stock back, it will be a return, so an order with negative value, or a go back to draft. Rarely we will want to make both a stock movement and have the source of it not beeing trackable by having it completely deleted).

Currently there is no deletion that make stock change, whatever are the option for stock increase/decrease because deletion is not a business event but a purge event.

ibuiv:
I think it is too easy to delete a validated order (which may changed the stocks and also be the origin of stock movements) without considering the consequences. 
Once deleted, it's impossible to track why this stock changed because the origin is not here. 
 
Maybe a big warning to add to inform the user this action may cause a stock mismatch. Or deny the possibility of delete on order which is the origin of any stock moves ? 

My quick solution was to remove the permissions of order deletions. 
But I think it's logic to delete a non validated order (to purge for example).

What do you think ? 
Thanks again for all your time. 

P.S. : I spoke to several users of Dolibarr in Nantes, and when I described this issue, they were agree with the non attended and non logical behaviour for this action with this configuration.

PR #19503, opened 2021-11-27
Update main.lang

KlaasOl:
Did not complete earlier work.

Date format in Dutch corrections


eldy:
Can you check also that fix is done into transifex (otherwise, change will be lost after next transifex sync for nl_NL)

PR #19509, opened 2021-11-28
FIX Html conversion check with PDF_BOLD_PRODUCT_LABEL

fr69400:
Fix  PDF_BOLD_PRODUCT_LABEL
Fix PDF_BOLD_PRODUCT_REF_AND_PERIOD
Add the definition on admin/pdf.php

### DIFF
--- a/htdocs/core/lib/pdf.lib.php
+++ b/htdocs/core/lib/pdf.lib.php
@@ -1363,7 +1363,7 @@ function pdf_getlinedesc($object, $i, $outputlangs, $hideref = 0, $hidedesc = 0,
 		// Description short of product line
 		$libelleproduitservice = $label;
 		if (!empty($libelleproduitservice) && !empty($conf->global->PDF_BOLD_PRODUCT_LABEL)) {
-			if (!dol_textishtml($libelleproduitservice)){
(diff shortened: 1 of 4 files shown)


htdocs/core/lib/pdf.lib.php
```diff
@@ -1363,7 +1363,9 @@ function pdf_getlinedesc($object, $i, $outputlangs, $hideref = 0, $hidedesc = 0,
 		// Description short of product line
 		$libelleproduitservice = $label;
 		if (!empty($libelleproduitservice) && !empty($conf->global->PDF_BOLD_PRODUCT_LABEL)) {
-			// This part of code is bugged. It introduces a HTML tag making the label a html string but without converting \n into br if it was a full text non html string before.
+			if (!dol_textishtml($libelleproduitservice)) {
+				$libelleproduitservice = str_replace("\n", '__N__', $libelleproduitservice);
```
eldy, line 1367: Why not just replacing directly into   <br>  instead of  __N__ ?

fr69400: Why not...

I update the code to replace directly into br

PR #19517, opened 2021-11-30
Create supplier proposal improve

1ocate:
If supplier have own currency should always choice.
Sometimes user did'nt see currency.
So if supplier have own currency automatic change select currency is better good way.


1ocate:
have issue

PR #19520, opened 2021-11-30
Fix bug within contact tabs of invoice

### LINKED ISSUE #19519: Can't edit contacts related to an invoice
# Bug
We can't edit contact related to an invoice; the invoice is never fetched from the database within this script and we directly fall under "ErrorRecordNotFound". It's probably because without action into the URL we directly go to the view part, and it starts with this piece of code :
```
if (empty($object->id)) {
        llxHeader();
        $langs->load('errors');
        echo '<div class="error">'.$langs->trans("ErrorRecordNotFound").'</div>';
        llxFooter();
        exit;
}
```

https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/compta/facture/contact.php#L103
and 
https://github.com/Dolibarr/dolibarr/blob/13.0/htdocs/compta/facture/contact.php#L111

## Environment
-  13.0
-  Debian 10
- Apache
- PHP 7.3.31-1~deb10u1
-  PostgreSQL 12
- /compta/facture/contact.php?facid=4395

be able to edit contact related to a invoice

We fall directly under the "Enregistrement non trouvé." / ErrorRecordNotFound invoice is never fetched

Just visit the contact tab of an invoice with Dolibarr 13 and probably 14

l00ptr:
# Fix #19519

Contact tab in invoice can't work. we never get the invoice from database
object->id stay null so we always fall under the exit (and ErrorRecordNotFound).

This commit tries to fix this issue by fetching the object directly after the creation
of "Facture" object before those line :

```
if (empty($object->id)) {
        llxHeader();
        $langs->load('errors');
        echo '<div class="error">'.$langs->trans("ErrorRecordNotFound").'</div>';
        llxFooter();
        exit;
}

```

### DIFF
--- a/htdocs/compta/facture/contact.php
+++ b/htdocs/compta/facture/contact.php
@@ -52,7 +52,7 @@ if ($user->socid) {
 $object = new Facture($db);
 // Load object
 if ($id > 0 || !empty($ref)) {
-	$ret = $object->fetch($id, $ref, '', '', $conf->global->INVOICE_USE_SITUATION);
+		$object->fetch($id, $ref, '', '', $conf->global->INVOICE_USE_SITUATION);
 }
 
 $result = restrictedArea($user, 'facture', $object->id);


htdocs/compta/facture/contact.php
```diff
@@ -52,7 +52,8 @@
 $object = new Facture($db);
 // Load object
 if ($id > 0 || !empty($ref)) {
-	$ret = $object->fetch($id, $ref, '', '', $conf->global->INVOICE_USE_SITUATION);
+		$object->fetch($id);
```
eldy, line 55: Why removing the other parameters likee $ref ?

l00ptr: indeed after reading the doc; i probably need to readd them (i added those line by copy / pasting some code from another script not using them.

So let's readd them. So i propably shoud remove the fetch_thirdparty just bellow.

l00ptr: fixed thx for your review eldy

l00ptr:
We have the same problem here : https://github.com/Dolibarr/dolibarr/blob/ca8dab3/htdocs/compta/facture/info.php#L46

@eldy should i fix it into another MR ? What's the best way to fix this ? I think we should fetch the data before the if and exit part, what do you think about it ?

PR #19533, opened 2021-12-01
CLOSE #19531 Product REST API: allow list / search method to retrieve stock data.

### LINKED ISSUE #19531: Product REST API: allow list / search API to retrieve stock data at the same time.
# Feature Request
Allow to optionally get stock information directly from the lookup product REST API (index method).

## Use case
This would avoid multiple request, one to list product and one for each product to get stock information.

lainwir3d:
# CLOSE #19531
Allow to optionally get stock information directly from the lookup product REST API (index method), to avoid requesting each product afterward.


lainwir3d:
I don't get why one of the checks is wrong.

lainwir3d:
Seems unrelated to me.

PR #19553, opened 2021-12-06
Add hookmanager in function pdf_pagefoot of pdf.lib.php

habot-it:
Customize PDF footers with HTML.
(Only on the last page with free text)


htdocs/core/lib/pdf.lib.php
```diff
@@ -1143,50 +1143,82 @@ function pdf_pagefoot(&$pdf, $outputlangs, $paramfreetext, $fromcompany, $marge_
...
 	}
 
-	$marginwithfooter = $marge_basse + $freetextheight + (!empty($line1) ? 3 : 0) + (!empty($line2) ? 3 : 0) + (!empty($line3) ? 3 : 0) + (!empty($line4) ? 3 : 0);
-	$posy = $marginwithfooter + 0;
+	// For customize footer
+	if (is_object($hookmanager)) {
+		$parameters = array('line1' => $line1, 'line2' => $line2, 'line3' => $line3, 'line4' => $line4, 'outputlangs'=>$outputlangs);
+		$action = '';
+		$hookmanager->executeHooks('pdf_pagefoot', $parameters, $object, $action); // Note that $action and $object may have been modified by some hooks
+		if (!empty($hookmanager->resPrint) and $hidefreetext == 0) {
```
eldy, line 1151: Use && instead of "and" for code standardization.

PR #19558, opened 2021-12-06
Remove illegal characters in SEPA XML

yaslama:
# NEW|New Remove illegal characters in SEPA XML

### DIFF
--- a/htdocs/compta/prelevement/class/bonprelevement.class.php
+++ b/htdocs/compta/prelevement/class/bonprelevement.class.php
@@ -121,6 +121,26 @@ class BonPrelevement extends CommonObject
 		$this->fetched = 0;
 	}
 
+	/**
+	*	Remove any illegal characters for SEPA
+	*
+	*	@param	string	$str			String to remove illegal chars
+	*	@return string   	       		Cleaned string
+	*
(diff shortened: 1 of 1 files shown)


htdocs/compta/prelevement/class/bonprelevement.class.php
```diff
@@ -121,6 +121,26 @@ public function __construct($db)
...
 	}
 
+	/**
+	*	Remove any illegal characters for SEPA
+	*
+	*	@param	string	$str			String to remove illegal chars
+	*	@return string   	       		Cleaned string
+	*
+	*/
+	public static function removeIllegalChars($str)
```
eldy, line 131: For my information, which character did you met that make the xml not valid ?

yaslama: We received this error from the CIC: `Cet élément est invalide - La valeur '<valeur>' est invalide d'après son type de donnée 'Max70Text' - La vérification de la contrainte de pattern a échoué.
(Non conforme à la norme SEPA et/ou ISO20022)` when the string contained `&`, `"`, `<`, `_`,`Ç`.

PR #19560, opened 2021-12-06
FIX: separator extrafields display

marc-dll:
Wrong default colspan (should be 1 to cover both columns when incrementing at `showOptionals()` call) + bad colspan on user card.

I know this code has been modified on develop by #19124 , but it was only to remove a warning and not to fix the display problem.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -7096,18 +7096,17 @@ abstract class CommonObject
 					$langs->load($extrafields->attributes[$this->table_element]['langfile'][$key]);
 				}
 
-				$colspan = '';
+				$colspan = 1;
 				if (is_array($params) && count($params) > 0) {
-					if (array_key_exists('cols', $params)) {
-						$colspan = $params['cols'];
+					if (array_key_exists('cols', $params) && is_numeric($params['cols'])) {
+						$colspan = intval($params['cols']);
 					}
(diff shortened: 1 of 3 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -7096,18 +7096,17 @@ public function showOptionals($extrafields, $mode = 'view', $params = null, $key
 					$langs->load($extrafields->attributes[$this->table_element]['langfile'][$key]);
 				}
 
-				$colspan = '';
+				$colspan = 1;
```
eldy, line 7099: It seems to be a backport.  On new code this is   $colspan = 0;

marc-dll: As I said above, there is a PHP warning bug and a display bug. The warning bug has been fixed in develop by the PR I referred to above. But it is not the case for the display bug no : setting $colspan to 0 by default provokes a separator with attribute colspan="1". It therefore doesn't expand to the full width of most of the cards, that have 2 columns.

Call the first part a backport if you want, but the second part is fixed nowhere, needs the first part to be fixed, and anyway, in the first place, the correct way to fix the first part should have been to fix it in v12.0.

eldy: So my understanding is that there is still a bug into develop (only 1 on 2 were fixed) ?
Can you describe how we can reproduce this second bug on develop ?

eldy:
Closed after too long time pening without answer.

PR #19571, opened 2021-12-07
multiple serial numbers in dispatch and reception

altairis-noe:
# NEW|Manage serial numbers in receptions
If a product uses unique serial numbers, we replace the input for the batch number by a textarea in which we can put as many serial numbers as we need.
211203SN-Multiples-receptions
Those serial numbers are then processed in the reception card. In addition, we have a link to the batch/lot card in the reception card, which can be quite useful
211203SN-Multiples-reception


fappels:
Not for this PR, but related:
~~We should also have new stock module option 'stock movement on reception create'. Because check unique is done on stock movement, so we see not unique error only on reception validation.~~
@altairis-tof Noticed reception card first add all line in memory, so my remark is useless. This PR already solves the issue for check duplicate SN, you now can easy correct or remove duplicate SN.

altairis-tof:
> Not for this PR, but related: We should also have new stock module option 'stock movement on reception create'. Because check unique is done on stock movement, so we see not unique error only on reception validation.

@fappels : why do you want to create a reception without stock movement ? unless i miusunderstood...
anyway it would be better to open a new issue because it is already so difficult to have a PR dealed with,  that it is not safe to add other discussions inside ;-)

1ocate:
This is little bit different topic.. but wanna talk with all..

Case Shipment(expedition) create including free type item.

But Reception.. only can item from Product.

This is bug or Reception only can use for Product register item?

Thank you All.

eldy:
Another interface to enter lot/serial (unique or not) is planned (based on the new tool "scan or mass input of serial") in progress in the inventory feature. This may be in conflict with this, so for the moment i suspend this PR.

altairis-tof:
Hello @eldy we tried to use the interface in progress but it is unusable in current state, especially in dispatches. @altairis-noe will tell you more about this but it's like the js code is too much imbricated with inventory code and features.

altairis-noe:
Hello @eldy,
as @altairis-tof said, in its current state, we can’t use the interface in progress which is strictly connected with inventories. For example, it is linked to a form (and js) in inventory which will be submitted on exit, which is incompatible with dispatch.
On a side note, I don’t think having one line per SN in dispatch is practical if we have to treat a big delivery.
I will happily participate to a global approach on serial/lots management, if there is a place for such discussion ;)

DATUC:
> Another interface to enter lot/serial (unique or not) is planned (based on the new tool "scan or mass input of serial") in progress in the inventory feature. This may be in conflict with this, so for the moment i suspend this PR.

Hi @eldy any update about **"scan or mass input of serial"** feature ???...

eldy:
Note: Url concerned is /fourn/commande/dispatch.php?id=...
TODO: We should add the interface to enter serial/lot into this page.

eldy:
Too many changes since or on the page.
Also combo of lot must be enabled if product need lot or serial, not only serial.

PR #19573, opened 2021-12-08
FIX: Backport compatibility for formconfirm() function

kkhelifa-opendsi:
FIX: Backport compatibility for formconfirm() function

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -4810,7 +4810,7 @@ class Form
 						$moreonecolumn .= '</div>'."\n";
 					} elseif ($input['type'] == 'hidden') {
 						// Do nothing more, already added by a previous loop
-					} else {
+					} elseif (!empty($input['type'])) {
 						$more .= 'Error type '.$input['type'].' for the confirm box is not a supported type';
 					}
 				}


htdocs/core/class/html.form.class.php
```diff
@@ -4810,7 +4810,7 @@ public function formconfirm($page, $title, $question, $action, $formquestion = '
 						$moreonecolumn .= '</div>'."\n";
 					} elseif ($input['type'] == 'hidden') {
 						// Do nothing more, already added by a previous loop
-					} else {
+					} elseif (!empty($input['type'])) {
```
eldy, line 4813: This will hide when a input field is used without a type defined. So this will hide a development bug. Why hiding such cases ?

kkhelifa-opendsi: For this case who is not a bug (it's used in formconfirm() function, ...) :
$formquestion[] = array('type' => 'other', 'name' => 'xxxxhour', 'label' => $langs->trans("Date"), 'value' => $form->selectDate($date, 'xxxx', 1, 1, 0, '', 0));
$formquestion[] = array('name'=>$input['name'].'min');

eldy: Sorry i didn't understood your example. Can you give me example of file and line where you have a line
$formquestion[] = array('name'=>'xxxxhour');   
without the type parameter ?

kkhelifa-opendsi: > $formquestion[] = array('type' => 'other', 'name' => 'duration_planned', 'label' => $langs->trans("DurationPlanned"), 'value' => $form->select_duration('duration_planned', $duration_planned, 0, 'text', 0, 1)),
> $formquestion[] = array('name' => 'duration_plannedhour'),
> $formquestion[] = array('name' => 'duration_plannedmin'),

it's just a sample if the 'value' have multiple input this case also occurs
we need to provide parameters all the input identifier used to send by GET in the url
if the form question is send by POST and not by GET this discussion would be useless

kkhelifa-opendsi:
I don't see this code has been added with this change
>					if (is_array($input) && isset($input['name'])) {
>						if (strpos($input['name'], ',') > 0) {
>							$inputok = array_merge($inputok, explode(',', $input['name']));
>						} else {
>							array_push($inputok, $input['name']);
>						}
>					}
So i close this issue even if we lose the compatibility of the old versions before v13 ...

PR #19577, opened 2021-12-09
Fix extrafield type timestamp and UK company lookup URL

arunasr:
Fixes handling of timestamp extrafields.
Fixes UK company lookup URL.


htdocs/core/class/commonobject.class.php
```diff
@@ -5936,6 +5936,8 @@ public function insertExtraFields($trigger = '', $userused = null)
 						break;
 					case 'date':
 					case 'datetime':
+					case 'timestam': //the type name may be truncated to 8 chars
```
eldy, line 5939: According to possible value of extrafields type, the timestam and timesamp does not exists
image

arunasr: They are possible from module; that's how I've got into trouble. The expiration timestamp wasn't being saved.
`
$extrafields->addExtraField('hmrc_token_tms', "HMRC OAuth2 token timestamp", 'timestamp', 1, 0, 'user', 0, 0, '', '', 0, '', 0, 0, '', '', 'vat@vat', $conf->vat->enabled),
`

eldy: So, question, is why did you use "timestamp" and not "datehour" for your extrafield addition ?

arunasr: My understanding is, timestamp is absolute time in UTC, while DATETIME is local time in given timezone.
I need an absolute timestamp (using time() function) when the token was retrieved, it has nothing to do with the timezone in which it happened.
I will probably use just a long int after all. But I do not think that current handling of a valid DB field type is correct.
This is only a suggestion, ignore if you don't like it.

arunasr: Note that currently the table llx_extra trims all names to 8 chars, so we end up with 'timestam' without a 'p'

eldy: No. It is different.
On mysql, timestamp is the same than datehour. But there is 2 difference:
If you use them into your own request, there is no difference. They are both using the same timezone, absolutely the same (the client that generate the sql decide if he inject local or gmt timezone. Into the request it is always just a string 'YYYY-MM-DD HH:MM:SS'). But the "timestamp" has a special flag on mysql to say, if i don't update the field myself, and i update another column, please update also yourself the field. So it is an automatic updated field. But this has a big problem. The database decide to update the field itself. And the database uses its own timezeone that may differs than the timezone of the application, leading to offset appearing and different behaviour depending on timezone of database.
We should not have result that depends on how a timezone of a database is set so using timestamp in Dolibarr was an error. It was cool due to the autoupdate field, but this information is not reliable. Also note the if using another database, the behaviour of mysql does not applies, so you have another behaviour.
So well, for extrafields, you should never use timestamp (result will be unpredicable). Only datetime. Data in database is always saved with the timezone of the php server. But there is not yet option into etrafields to decide if we want to show the date in the user timezone or in the server timezone. This is a missing feature but it can't be adressed using the timezome that must be not be used for user interaction.

arunasr: It is not important for me anymore. Let's drop it.
I use JSON in a TEXT field in my module now.

PR #19590, opened 2021-12-09
New : Empty case on extrafields selects

ghost:
# NEW  | Empty filter case on select extrafields 

The "selectEmpty" filter will allow to bypass the other filters to see only the lines with NULL values.
I took example from Tag/Categ from Product/Service.


htdocs/core/class/extrafields.class.php
```diff
@@ -1065,6 +1065,7 @@ public function showInputField($key, $value, $moreparam = '', $keysuffix = '', $
 
 			$out .= '<select class="flat '.$morecss.' maxwidthonsmartphone" name="'.$keyprefix.$key.$keysuffix.'" id="'.$keyprefix.$key.$keysuffix.'" '.($moreparam ? $moreparam : '').'>';
 			$out .= '<option value="0">&nbsp;</option>';
+			$out .= '<option value="-2">- '. $langs->trans("SelectEmpty") . ' -</option>';
```
atm-florianm, line 1068: I think the "select empty" option should not appear when you are on a card (modifying the value of the extrafield), but only for list filters.

atm-gregr: Well spotted ! I fixed it

ghost:
@eldy 
I didn't edit the file which is problematic for Travis. (cf img)
image

eldy:
> @eldy I didn't edit the file which is problematic for Travis. (cf img) image

The coding standard rules say: When you forge an SQL request, you must use the " to make the string and use the ' as a SQL string.
So here the ' HAVING 1=1 ' should be " HAVING 1=1 ".

PR #19594, opened 2021-12-10
FIX - unasigned variable $action given to hook printOriginObjectLine

fr69400:


fappels:
Think also 'printOriginObjectSubLine' should be 'printOriginObjectLine'. Used to be so in previous versions.

fr69400:
Yes, i dont know why the hook as been renamed...
It may be use to print lines not sublines...

But change.log file say : 

Following changes may create regressions for some external modules, but were necessary to make Dolibarr better:
* Update hook 'printOriginObjectLine', removed check on product type and special code. Need now reshook.

In fact, when you look at the original code, you can see the the original hook can't work properly. The condition before the call was very strange

fappels:
I did the update mentioned in Changelog, but some other patches were done after. Think 'printOriginObjectSubLine' is some copy paste error. I have send comment to originator but no reply yet. Think you can patch back to printOriginObjectLine in your PR and see what @eldy decides.

eldy:
Don't remember why the hook were renamed. i see no reason, so let's go back with name of v14.

PR #19600, opened 2021-12-13
FIX line extrafields title display error.

1ocate:
I found out line extrafield title display error when browser resize small.

Screen Shot 2021-12-13 at 2 44 15 PM

After this commit.

Screen Shot 2021-12-13 at 4 24 51 PM

I think no need class "wordbreak" for line extrafield title.

Please review and comment.

Thank you.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -6972,19 +6972,19 @@ abstract class CommonObject
 						if (!empty($conf->global->MAIN_VIEW_LINE_NUMBER) && $action == 'view') {
 							$out .= '<td></td>';
 						}
-						$out .= '<td class="wordbreak';
+						$out .= '<td';
 						//$out .= "titlefield";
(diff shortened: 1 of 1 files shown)


eldy:
The problem here looks different.
We should not have a long text into the column that is dedicated to the position. The text should start in second column.

Do you have same trouble on develop ?

1ocate:
@eldy Thank you for review develop no problem.
Screen Shot 2021-12-17 at 10 57 16 AM

But also found out develop also have class "wordbreak" however this is  second column and using div tag.

I think this problem only 13.0.
Because extrafields title same column with number column(fisrt column) and fallow size that(that column css no min-width). 
So I think this situation use that better remove "wordbreak".


Anyway when remove that class "wordbreak" from develop no have  problem(for me). 
But maybe newer version have but last version remove also wired.

Thank you.

eldy:
This bug or the feature request seems to be fixed or included into last recent stable version, so I close the issue.

PR #19605, opened 2021-12-13
Fix : add  info for acces right  on compta/ bank / info

atm-jpb:
# FIX
due to the new way on acces with fonction restrictedArea()
ex : _$result = restrictedArea($user, 'banque', $fieldvalue, 'bank_account', '', '', $fieldtype);_

Establish the link with the $parenttableforentity variable to manage access right


atm-maxime:
@atm-jpb since this change, we can't access the card of a bank entry. From the bank entry list of a bank account, simply click on the bank entry ref (compta/bank/line.php?rowid=18)

PR #19609, opened 2021-12-14
FIX supplier_proposal when line update price U.P change 0

1ocate:
When use multi currency and add line supplier_proposal it works ok.
Screen Shot 2021-12-14 at 2 08 54 PM

But update price U.P value not expected change.
Screen Shot 2021-12-14 at 2 09 22 PM
Screen Shot 2021-12-14 at 2 09 42 PM

I found out addline function and updateline function different. 
And match and code clean.

This problem also 14, 13 version. if need I will PR again (Let me know.)

Please check.
Thank you all.

### DIFF
--- a/htdocs/supplier_proposal/class/supplier_proposal.class.php
+++ b/htdocs/supplier_proposal/class/supplier_proposal.class.php
@@ -725,7 +725,9 @@ class SupplierProposal extends CommonObject
 			$total_ttc = $tabprice[2];
 			$total_localtax1 = $tabprice[9];
 			$total_localtax2 = $tabprice[10];
-			$pu = $pu_ht = $tabprice[3];
+			$pu_ht = $tabprice[3];
+			$pu_tva = $tabprice[4];
+			$pu_ttc = $tabprice[5];
 
 			// MultiCurrency
 			$multicurrency_total_ht  = $tabprice[16];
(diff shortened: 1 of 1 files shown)


htdocs/supplier_proposal/class/supplier_proposal.class.php
```diff
@@ -518,10 +518,6 @@ public function addline($desc, $pu_ht, $qty, $txtva, $txlocaltax1 = 0, $txlocalt
 				$txtva = preg_replace('/\s*\(.*\)/', '', $txtva); // Remove code into vatrate.
 			}
 
-			if (!empty($conf->multicurrency->enabled) && $pu_ht_devise > 0) {
```
eldy, line 521: I think we should keep this, because if the 2 value are provided (pu and pu_ht_devise), we want to have the  pu_ht_devise used in priority for prices calculation. So we need to force pu to zero for calcul_price_total.
May be you can keep only the line added later
`$pu = $pu_ht = $tabprice[3] fields ?`

1ocate: @eldy Thank you review.
You're right. 
I already fix it please check it.

Thank you.

1ocate:
Please check again thank you

PR #19610, opened 2021-12-14
Fix korean word and translate

1ocate:
maybe south korean user only me but I will keep change.

If have error please check.

Thank you.


1ocate:
I will use this.

PR #19622, opened 2021-12-15
Add 13.0 margin in paiement/list.php

1ocate:
This requirement my co worker.

He want see margin from "paiement/card.php"
This margin is needed to calculate the marketing commission that is usually given for each order, when 1 pay to many invoice.

Screen Shot 2021-12-15 at 1 21 50 PM

Please check and review please. 
Thank you all.

### DIFF
--- a/htdocs/compta/paiement/card.php
+++ b/htdocs/compta/paiement/card.php
@@ -382,7 +382,7 @@ if ($resql)
 	print '<td class="right">'.$langs->trans('PayedByThisPayment').'</td>';
 	//Add Margin
 	if (!empty($conf->margin->enabled)) {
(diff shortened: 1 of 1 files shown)


eldy:
The margin is not something related to a payment but related to an invoice. SO it would be better to have the column just after the invoice ref (between invoice and company).
Also you must push this into develop branch (v13 is froozen).

1ocate:
@eldy Thank you again review code and give comment.

I apply your comment and PR again. (#19642)

I want put margin column between invoice and company, but seem bit wired
so put after company.

If have problem please give comment again.
Screen Shot 2021-12-17 at 2 33 17 PM

eldy:
closed as duplicate of #19642

PR #19623, opened 2021-12-15
NEW : Add constant to change price HT in TakePos

altairis-melina:
Add constant TAKEPOS_CHANGE_PRICE_HT

### DIFF
--- a/htdocs/takepos/invoice.php
+++ b/htdocs/takepos/invoice.php
@@ -680,6 +680,8 @@ if ($action == "updateprice") {
 			} else {
 				if (empty($user->rights->takepos->editlines) || (empty($user->rights->takepos->editorderedlines) && $line->special_code == "4")) {
 					dol_htmloutput_errors($langs->trans("NotEnoughPermissions", "TakePos"), null, 1);
+				} elseif ($conf->global->TAKEPOS_CHANGE_PRICE_HT == 1) {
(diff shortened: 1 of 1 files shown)


htdocs/takepos/invoice.php
```diff
@@ -680,6 +680,8 @@ function fail($message)
 			} else {
 				if (empty($user->rights->takepos->editlines) || (empty($user->rights->takepos->editorderedlines) && $line->special_code == "4")) {
 					dol_htmloutput_errors($langs->trans("NotEnoughPermissions", "TakePos"), null, 1);
+				} elseif ($conf->global->TAKEPOS_CHANGE_PRICE_HT == 1) {
```
eldy, line 683: Try to use   getDolGlobalString('TAKEPOS_CHANGE_PRICE_HT')  instead of  $conf->global-TAKEPOS_CHANGE_PRICE_HT
so it will be compatible with php8.

altairis-tof: hello @eldy do youmean that each time we use $conf->global in the core and external modules, we have to call this function instead ???

eldy: Yes, this is the new way to read configurations from develop/v15.

altairis-tof:
@altairis-melina : please make the requested change so the PR may be merged.

altairis-tof:
@eldy Melina made the requested change; i think the PR is ok for merge now. Thanks

PR #19626, opened 2021-12-15
FIX FATAL ERROR : missing library

atm-kevin:
# FIX 

Fatal error when entering a public note on a proposal

146176221-5ded8da5-940a-4e29-92a1-907d4030c4ff

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -40,6 +40,7 @@
  */
 
 include_once DOL_DOCUMENT_ROOT.'/core/lib/json.lib.php';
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -40,6 +40,7 @@
  */
 
 include_once DOL_DOCUMENT_ROOT.'/core/lib/json.lib.php';
+include_once DOL_DOCUMENT_ROOT.'/core/lib/signature.lib.php';
```
eldy, line 43: The functions.lib.php is a file used everywhere.
signature should not be included here or it will be loaded everywhere.

PR #19629, opened 2021-12-15
NEW - Look feel card

BB2A-Anthony:
# NEW|New [*Look feel card*]
I propose a loock more puch for me. We must add the const THEME_ELDY_RADIUS

image

image

image

image


john-botella:
Use css var instead, this will limit usage of php for css file, will allow mainteners to keep clear css code.
One day I wish i have enough time to create an advance theme configurator in Dolibarr, so usage of css var is highly recomanded (first i have to finish parameter class and an helper class for list display management :-) )

PS : look and feel are personal, for me, this design remind me old web design like 2010

PR #19644, opened 2021-12-17
FIX: manage multientity in ticket public page by querystring (#Evarisk)

FHenry:
FIX: manage multientity in ticket public page by querystring


lmag:
Merci !

PR #19655, opened 2021-12-20
use any available fontawesome icon in picto

cfoellmann:
This allows to use all fontawesome icons available without the need for mapping all icons in the img_picto() function.

Usage:
`fontawesome_envelope-open-text_fas_red` (delimiter is `_` )
if you want to use fontaweseome icons: fontawesome_icon-slug_style_color_size


aspangaro:
@cfoellmann Seems good for me, I try for the moment the fontawesome library v6b3 in pro version with my Oblyon template. The only one problem for the moment is the weight of the library. ~x3/x4 > 100 Mo

PR #19664, opened 2021-12-20
FIX : Amount isn't rounded

atm-quentin:
# FIX Amount isn't rounded in substition


atm-maxime:
I think what you're looking for is `__AMOUNT_FORMATED__` 😉

atm-quentin:
thanks :)

Issue #19648, opened 2021-12-18
Changing the name order for a thirdparty from a member.


Guerno:
# Instructions
When creating a third party from a member, I think it would be better to use the last name first and then the first name after.

In fact, when the third party's accounting account number is subsequently created with the DIGITARIA tool, we will find ourselves less often with duplicates.

## Suggested implementation
I think the changes are very simple in the file:
htdocs/adherents/card.php
* line 596: replace: $fullname = $object->getFullName($langs);
  * by: $fullname = $object->lastname.' '.$object->firstname;
* line 1476: replace: $fullname = $object->getFullName($langs);
  * by: $fullname = $object->lastname.' '.$object->firstname;

---
I'm not comfortable with GitHib, hope to do things the right way.
cordially.

daraelmin:
There is already an option to decide the common order of firstname/lastname (into home - setup - display).

PR #19670, opened 2021-12-21
FIX : #19648 Change name order when creating a third party from a member

jocker132:
FIX : #19648 Change name order when creating a third party from a member

The update was conducted on the file adherents/card.php in order to send the last name then the first name to third party creation and avoid to have duplicates as requested in the issue


daraelmin:
I think this should be an option (setup - other) not the default way.

jocker132:
You are right i'll work on it

jocker132:
Please @eldy how can i provide DML database script. IS there any migration scripts under dolibarr htdocs folder to use?
I want to add a parameter under llx_const table.

eldy (commented):
There is already an option to decide the common order of firstname/lastname (into home - setup - display).
You can get the string using the getFullName() of object adherent (member) that should take this option into account.

jocker132:
> There is already an option to decide the common order of firstname/lastname (into home - setup - display). You can get the string using the getFullName() of object adherent (member) that should takebthisboption into account.

So in this case we should close the issue and advise the user to change the settings from Home>>Setup>>Display since the actual code on 12.0 aleady calls getFullName.

image

eldy:
Closed as requested

PR #19684, opened 2021-12-23
2022.4 invoice

SylvainLegrand:
Minimum prices with customer prices
Load the right bank account on creation card
Show the IBAN well formed
Load all the needed information on list and indexes


eldy:

PR #19701, opened 2021-12-28
Fix #19667 : bankentries wrong SQL parenthesis + missing filter

### LINKED ISSUE #19667: bank entries search result
If I search a Third-party/User in the bank entries, all the results are duplicate 3 times.
/compta/bank/bankentries_list.php

Screenshot 2021-12-21 at 17 31 53

atm-maxime:
SQL parenthesis was misplaced so the "OR" was wrongly done.
And search_thirdparty var was wrong


Silocan:
I've got the same problem with dolibarr v13.0.4 and this fix doesn't work

PR #19703, opened 2021-12-28
Fix #19662 : mysql8 user creation in 2 steps

### LINKED ISSUE #19662: Cannot grant user privileges on dolibarr database when using MySQL 8
# Bug
When using MySQL 8.0, Dolibarr install scripts cannot grant the dolibarr user the privileges it needs on the dolibarr database 

## Environment
- **Version**: 14.0.4
- **OS**: Ubuntu 20.04
- **Web server**: Apache 2.4.41
- **PHP**: 7.4.3
- **Database**: MySQL 8.0.19
- **URL(s)**: [*Affected URL(s)*]

After filling in database name and database user credentials in the web-based installer and selecting "Create database" and "Create user account or grant user account permission on the Dolibarr database", the installer should do that -  "Create database" and "Create user account or grant user account permission on the Dolibarr database".

Instead, it returns the error:

> Creation user : dolibarr	Error: DB_ERROR_SYNTAX You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'IDENTIFIED BY 'dolibarr007'' at line 1. Connected user not allowed to GRANT ALL PRIVILEGES ON dolibarr.* TO 'dolibarr'@'localhost' IDENTIFIED BY '*****'

## Reasons

In MySQL 8.0, you can no longer create a user and grant privileges all within the same command. The user needs to be created first, then privileges can be granted to that user.

In MySQL 5.7, you could do both in one step, which is what Dolibarr is trying to do here (in dolibarr/htdocs/core/db/mysqli.class.php):
```
mysql> GRANT ALL PRIVILEGES ON dolibarr.* TO 'dolibarr-user'@'localhost' IDENTIFIED BY 'dolibar-user-password';
```

In MySQL 8.0, you have to do it in two steps:
```
mysql> CREATE USER 'dolibarr-user'@'localhost' IDENTIFIED BY 'dolibar-user-password';
mysql> GRANT ALL ON dolibarr.* TO 'dolibarr-user'@'localhost';
```
See:
[Grant command in MySQL 8.0](https://dev.mysql.com/doc/refman/8.0/en/grant.html)
[Grant command in MySQL 5.7](https://dev.mysql.com/doc/refman/5.7/en/grant.html)

## Fix

The current code (in dolibarr/htdocs/core/db/mysqli.class.php) first creates the user, but without allocating it a password (lines 942 and 955) then uses the now deprecated grant syntax to grant privileges and allocate password at the same time (line 958). Line 962 then provides an error code corresponding to the command in line 958, in case that command fails.

What we need is to first create the user, allocating it a password straight away, then grant the user privileges on the database. 

I tried the the following changes to lines 942, 955, 958 and 962:

### Line 942
Replace
`$sql = "CREATE USER '".$this->escape($dolibarr_main_db_user)."'";`
by
`$sql = "CREATE USER '".$this->escape($dolibarr_main_db_user)."'@'".$this->escape($dolibarr_main_db_host)."' IDENTIFIED BY '".$this->escape($dolibarr_main_db_pass)."'";`

### Line 955
Replace
`$sql = "CREATE USER '".$this->escape($dolibarr_main_db_user)."'@'localhost'";`
by
`$sql = "CREATE USER '".$this->escape($dolibarr_main_db_user)."'@'".$this->escape($dolibarr_main_db_host)."' IDENTIFIED BY '".$this->escape($dolibarr_main_db_pass)."'";`

### Line 958
Replace
`$sql = "GRANT ALL PRIVILEGES ON ".$this->escape($dolibarr_main_db_name).".* TO '".$this->escape($dolibarr_main_db_user)."'@'".$this->escape($dolibarr_main_db_host)."' IDENTIFIED BY '".$this->escape($dolibarr_main_db_pass)."'";`
by
`$sql = "GRANT ALL PRIVILEGES ON ".$this->escape($dolibarr_main_db_name).".* TO '".$this->escape($dolibarr_main_db_user)."'@'".$this->escape($dolibarr_main_db_host)."'";`

### Line 962
Replace
`$this->error = "Connected user not allowed to GRANT ALL PRIVILEGES ON ".$this->escape($dolibarr_main_db_name).".* TO '".$this->escape($dolibarr_main_db_user)."'@'".$this->escape($dolibarr_main_db_host)."'  IDENTIFIED BY '*****'";`
by
`$this->error = "Connected user not allowed to GRANT ALL PRIVILEGES ON ".$this->escape($dolibarr_main_db_name).".* TO '".$this->escape($dolibarr_main_db_user)."'@'".$this->escape($dolibarr_main_db_host)."'";`

I tried that change twice on two different virtual machines running MySQL 8.0, and it worked fine.
I have not tried it with MySQL 5.7 though.

atm-maxime:
@antoineguichard can you check please ?

### DIFF
--- a/htdocs/core/db/mysqli.class.php
+++ b/htdocs/core/db/mysqli.class.php
@@ -939,7 +939,7 @@ class DoliDBMysqli extends DoliDB
 	public function DDLCreateUser($dolibarr_main_db_host, $dolibarr_main_db_user, $dolibarr_main_db_pass, $dolibarr_main_db_name)
 	{
 		// phpcs:enable
-		$sql = "CREATE USER '".$this->escape($dolibarr_main_db_user)."'@'".$this->escape($dolibarr_main_db_host)."' IDENTIFIED BY '".$this->escape($dolibarr_main_db_pass)."'";
+		$sql = "CREATE USER '".$this->escape($dolibarr_main_db_user)."' IDENTIFIED BY '".$this->escape($dolibarr_main_db_pass)."'";
 		dol_syslog(get_class($this)."::DDLCreateUser", LOG_DEBUG); // No sql to avoid password in log
 		$resql = $this->query($sql);
 		if (!$resql) {
--- a/htdocs/core/db/mysqli.class.php
+++ b/htdocs/core/db/mysqli.class.php
@@ -939,7 +939,7 @@ class DoliDBMysqli extends DoliDB
 	public function DDLCreateUser($dolibarr_main_db_host, $dolibarr_main_db_user, $dolibarr_main_db_pass, $dolibarr_main_db_name)
 	{
 		// phpcs:enable
-		$sql = "CREATE USER '".$this->escape($dolibarr_main_db_user)."'";
+		$sql = "CREATE USER '".$this->escape($dolibarr_main_db_user)."'@'".$this->escape($dolibarr_main_db_host)."' IDENTIFIED BY '".$this->escape($dolibarr_main_db_pass)."'";
 		dol_syslog(get_class($this)."::DDLCreateUser", LOG_DEBUG); // No sql to avoid password in log
 		$resql = $this->query($sql);
 		if (!$resql) {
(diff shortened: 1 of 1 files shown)


eldy:
Do you remember what was the request that hang with mysql8 and current code ?

atm-maxime:
Check out the issue I mention #19662 it's well explained by @antoineguichard

antoineguichard:
I have not been able to test it further than what I reported in https://github.com/Dolibarr/dolibarr/issues/19662#issuecomment-1003186501 as I have not managed to build a VM combining both MYSQL 5.1 (for which support ended 2013-12) and PHP 5.6 (first released 2014-08).

htdocs/core/db/mysqli.class.php
```diff
@@ -939,7 +939,7 @@ public function DDLDropField($table, $field_name)
 	public function DDLCreateUser($dolibarr_main_db_host, $dolibarr_main_db_user, $dolibarr_main_db_pass, $dolibarr_main_db_name)
 	{
 		// phpcs:enable
-		$sql = "CREATE USER '".$this->escape($dolibarr_main_db_user)."'";
+		$sql = "CREATE USER '".$this->escape($dolibarr_main_db_user)."'@'".$this->escape($dolibarr_main_db_host)."' IDENTIFIED BY '".$this->escape($dolibarr_main_db_pass)."'";
```
eldy, line 942: The dolibarr_main_db_host is the host of the web server. In some case, the database mayy be into another one.
So i think we should not add the '@'".$this->escape($dolibarr_main_db_host)."' here, only add the IDENTIFIED BY so database will use the default value for domain name (like actually).

PR #19732, opened 2022-01-04
FIX : allow create supplier order from other external modules

defrance:
we use the same origin feature like other element who allow create supplier order from external call module


defrance (commented):
idem for create and add

PR #19737, opened 2022-01-04
NEW Add possibility with constant MAIN_LOGIN_BADCHARUNAUTHORIZED to define bad character unauthorized into login name

aspangaro:
With a migration on v14, have a problem with login contains '@'. In fact, some customer use email address like a login and it's not possible.

Add a constant to define /[,<>"\']/ to autorize character '@'


eldy:
Note that you can login with email address even if you set a login that is not an email address. There is no need to use a login that is email for that.

PR #19755, opened 2022-01-06
Add hooks

vincentjdc:
# NEW|New Add hooks to a series of pages

This PR adds hooks to allow adding selected, joins and where conditions to several SQL queries.

### DIFF
--- a/htdocs/commande/class/commandestats.class.php
+++ b/htdocs/commande/class/commandestats.class.php
@@ -61,8 +61,9 @@ class CommandeStats extends Stats
 	 * @param   int		$userid     Id user for filter (creation user)
 	 * @param	int		$typentid   Id typent of thirdpary for filter
 	 * @param	int		$categid    Id category of thirdpary for filter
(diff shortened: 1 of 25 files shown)


htdocs/commande/class/commandestats.class.php
```diff
@@ -61,8 +61,9 @@ class CommandeStats extends Stats
 	 * @param   int		$userid     Id user for filter (creation user)
 	 * @param	int		$typentid   Id typent of thirdpary for filter
 	 * @param	int		$categid    Id category of thirdpary for filter
+	 * @param	Hookmanager		$hookmanager    A hook manager for more filtering
```
eldy, line 64: A choice has been done in architecture to have the $hookmanager being a global variable like it is for $db, $conf, $langs and $mysoc.
So you must use it by doing global $hookmanager  instead of propagating it in parameters everywhere you need it. so please remove the param and modify the
global $user, $conf;
into
global $user, $conf, $hookmanager;

Issue #19763, opened 2022-01-07
ADD default Service on Task Time


lmag:
# Instructions
ADD FK_product Default Service in projet_task_time

# Feature Request
Adding this key allows you to retrieve it and automatically manage invoicing

## Use case
Billing of time spent by service automatically

## Suggested implementation
Step 1 : ADD FK_product in BDD
![image](https://user-images.githubusercontent.com/3288514/148534475-868c2ac7-5d94-48e6-8702-4ca49681f914.png)

Step 2 : Add the option to recover the service linked to the time spent line in the invoice creation screen
![image](https://user-images.githubusercontent.com/3288514/148535092-54190567-ae9b-4949-9d75-8e3810dfd4b8.png)

lmag:
@nicolas-eoxia @evarisk-theo @FHenry

lmag:
STEAP 1 : https://github.com/Dolibarr/dolibarr/pull/19764

lmag:
STEAP 2 : In progress : https://github.com/Dolibarr/dolibarr/pull/19765

FHenry:
Add mass action to apply product on time spend
Change label on Service to Invoice with Service to Invoice (when empty)

When Invoice use product of line

PR #19764, opened 2022-01-07
NEW: #19763 DB change only - Can invoice task time per different services

FHenry:
NEW: #19763 DB change only - Can invoice task time per different services


lmag:
👍

eldy:
Merge can't be done by github. Can you resubmit the pr after merge by develop ?

FHenry:
@eldy done, but as I can't(don't want) commit wihout local git hook phpCSsnifer is ok, I have to put comment into Wordpress OAuth class also

eldy:
> @eldy done, but as I can't(don't want) commit wihout local git hook phpCSsnifer is ok, I have to put comment into Wordpress OAuth class also

When you have modified files into includes dir, you can make a commit for this file with
git commit --no-verify
so it bypass phpcs for it, then you commit the rest of files as usual.

Issue #18606, opened 2021-09-04
v14.0.0. - 14.0.1 and 14.0.2 The tasks show the list of attachments, but do not allow to open or download them.


JoAnMaFe:
Since the update to **V14.0.0 -14.0.1 y 14.0.2** 
the tasks show the list of attachments, but do not allow to open or download them, displaying the following message
Access denied.
You are trying to access a page, area or functionality of a disabled module or without an authenticated or unauthorised session for your user.
Current login: administrator Permissions for this user can be
permissions for this user can be assigned by Dolibarr administrator through menu Start-> Users.
With previous versions (13.04) this problem did not exist.
In the end I discovered that while everything was working OK. in 13.04 when upgrading to v 14.0.0 the files attached to the tasks cannot be downloaded or displayed.
It is possible that many users do not add attachments to their tasks but if their tasks are based on JPG or PDF photo documentation, they will not be able to use it and will not be able to support their customers.  

# Bug
[the tasks show the list of attachments, but do not allow to open or download them]

## Environment
- **Version**: [14.0.0 - 14.0.1 and 14.0.2]
- **OS**: [W10 x64]
- **Web server**: [Apache/2.4.48]
- **PHP**: [7.4.23]
- **Database**: [10.5.4-MariaDB]
- **URL(s)**: [Localhost]

## Expected and actual behavior
[Restore viewing and downloading of task attachments.]

## Steps to reproduce the behavior
[*Verbose description*]

## [Attached files](https://help.github.com/articles/issue-attachments) (Screenshots, screencasts, dolibarr.log, debugging informations…)
[*Files*]

JoAnMaFe:
In v14.0.2 it doesn't work for me either.
Access denied.
You try to access a page, area or functionality of a disabled module or without an authenticated session or not allowed to your user. 
Permissions for this user can be assigned by Dolibarr administrator via menu Start-> Users. 

In addition, thumbnails of events are not visible, they can be opened, but a blank icon is displayed.

edupulpillo:
The problem persist in 14.0.3
Trying to access to any attached file to the task using the link is returning a problem of permissions, even using the admin user with all the rights enabled. The only way to be able to access to the file attached is generate a public access link, and use it.

This is the error message returned:

> Access denied.
> You try to access to a page, area or feature of a disabled module or without being in an authenticated session or that is not allowed to your user.
> Current login: admin
> Permission for this login can be defined by your Dolibarr administrator from menu Home->Users.

[...]

PR #19770, opened 2022-01-08
14.0 FIX #18716 , #18606 , #18589 , #18617 permissions problem trying to open a file attached to a task

### LINKED ISSUE #18716: v14.0.0. - 14.0.1 y 14.0.2 Las archivos adjuntos asociados a las tareas, no permiten abrirlos ni descargarlos. Solucionado para mi
The files attached to the task appear in the attachment list, but you cannot open or download them. # 18617
alsoft10 I propose a solution # 18589 but in my case it did not work, with all the files attached, some of them did but others remained the same.
There was another problem with the mini images of the attached files in the events, they could be opened, but it did not allow the preview of the thumbnail.
In any case, I thank you for cluing me in on the location of the problem. (which worked for me).
I replaced the file files.lib.php located in Apache24/htdocs/dolibarr/htdocs/core/lib with the same file, but from a version 13.0.3 that I have in a production system.
For the moment it works correctly, although not being a programmer, I don't know if it will eventually generate another problem.

## Environment
- **Version**:14.0.0  14.0.1 - 14.0.2
- **OS**: W10
- **Database**: MarisDB

See if it can be officially fixed in 14.0.3

edupulpillo:
The permissions problem trying to open a file attached to a task is mentioned in multiple Issues. With this small update described below, can be fixed and closed #18716 , #18606 , #18589 , #18617, #19770

Replace **projet&project** by **projet_task&project_task** in the line number 2665 of the file  /htdocs/dolibarr/htdocs/core/lib/files.lib.php in version 14.0.5
This small update fix the problem.

### DIFF
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -2660,7 +2660,7 @@ function dol_check_secure_access_document($modulepart, $original_file, $entity, 
 				include_once DOL_DOCUMENT_ROOT.'/projet/class/task.class.php';
 				$tmptask = new Task($db);
 				$tmptask->fetch('', $refname);
-				$accessallowed = checkUserAccessToObject($user, array('projet_task'), $tmptask->id, 'projet&project', '', '', 'rowid', '');
+				$accessallowed = checkUserAccessToObject($user, array('projet_task'), $tmptask->id, 'projet_task&project_task', '', '', 'rowid', '');
 			}
 		}
 		$original_file = $conf->projet->dir_output.'/'.$original_file;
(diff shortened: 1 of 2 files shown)


atm-maxime:
Thanks for your fix, I'll test it. But there's also a modification of the cron_run_job file, why ?

edupulpillo:
> Thanks for your fix, I'll test it. But there's also a modification of the cron_run_job file, why ?

Sorry, i mistake in the  commit, since it should be two different commits.  
There should be two commits:

1. One for the files attached to the task, taht you already have accepted, 
2. Another related the cron jobs to allow in scripts cron Jobs use CLI and CGI mode, since the sharing hostings and Cpanel always call it as cgi/we instead cli mode. I mention this problem about cron job here:

#19258

eldy:
Can you split the PR in 2 PR ? I have applied the fix manually to be able to have a separated PR.
There is no need to make the script/cron_run_job file running as a cgi, because there is already a version of this script to be ran as cgi into htdocs/public/cron/cron_run_jobs_by_url.php

PR #19776, opened 2022-01-08
Updates to (very)old replacement code in invoices

mikee2:
# FIX|Fix #1 Fix wrong replacemetns
Old code did not do the replacement correctly but I was not able to manage the way to update this repo. This code fixes all problems and does the replacement the right way.

# NEW|New New function to get the date relative to the invoiced date
Formerly if you [re]generated  an invoice the replacements were [re]calculated relative to the date when the generation was taking place. This changed the content of the invoice voiding it.

Now the replacements use the date the original invoice as created not the date where you are re-generating it (this occurs, for example, when paying it). So the invoice content does not change reletive to the original one.

I have used the code of V14.0.0 (that is the one that I am running) to include my code. This is the reason that there are so many changes in odf.php because the file in github was very old.

Bear with me cause I am not very knowledgeable with this platform.


eldy:
Can you describe what is wrong currently with an example to reproduce the bug ?

mikee2:
The old code used the system date in the replacements as I did not know too much about the inners of dolibarr. The side effect of this is that whenever you did re-generate the invoice (or when it was automatically re-generated when paid for example) the dates in the replaced fields changed if you did not make the change at the same date (day, month or year) that was initially used to generate the first copy of the invoice.

Of course this is not correct: once generated an invoice should be immutable. But you can always push de generate button or any internal process may generate the invoice again at a later time. So the code has to guarantee that if the invoice was already generated the replacements would be the same as the original. To achieve that the proposed code now takes the reference date from the invoice itself and not from the system. As the invoice date is already a filled field when we reach the replacemtn code, the result is always the same no matter how many times you generate it or how far from the initial date you repeat the process.

htdocs/includes/odtphp/Segment.php
```diff
@@ -169,29 +169,84 @@ public function merge()
     *
     * Miguel Erill 09/04/2017
     *
+    * * 01/02/2019 - Changed the replacement mode to adapt to new way of doing things
+    *
+    * + 27/03/2019 - Added a function in odf.php to retrieve odf variables. Use it here to 
+    *                get the actual invoice creation date so that date functions are relative
+    *                to invoice dates not to invoice generation dates as before. Now you can 
+    *                (re)generate the invoice when you need and the substitutions will not change
```
eldy, line 177: Thinks are not clear for me.
When you generate an invoice, you take the odt template and replace the tag {object_date} into date with the format of the output language selected (for exemple in spanish format if user ask to generate doc in spanish). Right ? And if do this the 1st we should have __CURRENTDAY__ = 1
If you generate a second time the invoice, into italian, then the doc will reuse the same template and will replace the {object_date} into this template (the template did not change) to build another doc into italian format. Right ? And if we do this the 2nd, we should have __CURRENTDAY__ = 2. Right ?
So what is goal of your patch ?

mikee2: Hi Eldy.
Up to now the date was obtained from the system every time you generated the invoice. This is not correct: the date of the invoice MUST be the same no matter how many times you generete it. If you want to generate a different invoice (cause you have issued it at the wrong day) then delete the existing one and generate a new one.

I see no point generating the same invoice with different invoice dates. Besides, this date change was unexpected and undesired: the moment you paid an invoice it was [re]generated by the system in the background (do not understand why) with the date of payment, some times months after, spoiling the orginal invoice. I had to rollback the date of the server just to fix them.

This function get the date from the invoice itself, be new of existing, so the invoice does not change over time no matter how many times you generete it and when you generate it. Of course only if you do not delete the original one.

Hope this clarifies the need of that function.

mikee2: Hi Eldy.

You are right, {object_date} has the inmutable invoice date but the fact is that it is not available in the 'segment' file (or I was not able to find it) so I created the new function in the 'odf' file to retrieve it. It is the only modification to that file.

In the meantime you also are right that __CURRENTDAY__ should give the value relative to the current date to be consistent, not the one of the invoice. As I want, and need, the one on the invoice I have created a new block where 'CURRENT' has been changed by 'INVOICE' to build equivalents to the others (now relative to the current date) but relative to the invoice date.

Hope this clarifies the used tags and validate the changes.

eldy: This is still not clear.
If you have the tag {object_date}, wherever it is in template, in header or in lines, it will be replaced by date of invoice.
This is done into od->parse() method, the code make the replacement everywhere into the ->contentXml
So even if {object_date} is inside a segment (a line of invoice), replacement is done with the date of invoice.

mikee2: I'll try to clarify. Please correct me if I am wrong.

The tag {object_date} can be used, wherever you want, in the template as you say. But the use that I am pursuing is not on the template itself but on the lines the user writes inside the invoice. If you use the existing tag you will get the full date and I am in need of getting only the day, the month (be in number or in textual form), the year, etc. This is the reason of this macro replacement. I need to tell the customer about next mont/year ending of his contract renewal, and things like that that I did not know how to do before I implemented this macro replacement.

Can this be achieved with the existing tags and without the proposed code?

eldy: I see
This looks a similar need tha we used to build invoice notes from a template invoice. We made replacement of such data (previous month, next year, ... compared to the invoice date).
But i don't think it is a good think to but a solution that use specific code of Dolibarr inside the odf class (it should remaine an independant lib).
I pushed a suggestion to address your need. For the moment i have hard coded the new var you miss, but in a future we shoud (at the same place where i added the code), 1) scan the template to found tags and the format to output it, add the tag into the substitution array so any property will be supported with no need to hard code the list of tags.

mikee2: Ok. So you suggest that instead of using my tags __SOMETHING__ I better use {object_WHATEVER}, right?

In this case we could dismisss, when this alternative is available, all my code (both the one in odf and segment) as useless and thin a bit those source files.

What I miss in your list is {object_DAY} for the invoice day (I need it because of licenses expiring dates when they are the same as the issued invoice for example).

Do you have any idea of when will them be available so that I adjust all my invoices?

And it will be good to document that somewhere because I had no idea they existed and they are quite useful.  This is not a complain only a humble suggestion.

Thanks for all.

eldy: You can make another PR To add into htdocs/core/modules/facture/doc/doc_generic_invoice_odt.modules.php
the {object_DAY} 
As this is PR merged into develop, it should be released with v16 so in beta in April.

eldy:

PR #19785, opened 2022-01-10
Module assets - Totally review - Palier2 FPC21

aspangaro:


htdocs/install/mysql/tables/llx_asset.sql
```diff
@@ -15,20 +15,37 @@
...
+	date_start					date 			NOT NULL,
+	qty							real			DEFAULT 1 NOT NULL,	-- qty asset
+	acquisition_type			smallint		DEFAULT 0 NOT NULL,	-- new, occasion
+	asset_type					smallint		DEFAULT 0 NOT NULL,	-- intangible, tangible, in progress, financial
+	not_depreciated				integer(1)		DEFAULT 0,
+
+	disposal_date				date,
+	disposal_amount_ht			double(24,8),
+	fk_disposal_type			integer,
+	disposal_depreciated		integer(1)		DEFAULT 0,
```
eldy, line 38: integer(1) is not allowed into sql (pb of partability with mysql version)
use just
integer

eldy:
As mentionned during last devcamp, and into file CONTRIBUTE.md, any PR that bring new structure on tables must be sent into a separated PR that the PR that bring the php code.
So the database new structure is first validated before any code dev is sarted. So can you splist your PR in 2 to first submit one PR of database major addition/changes (the .sqlfiles so) ?
Also, can you rename the new sql files related to module asset into llx_asset_xxxx.sql  into   llx_asset_xxxx-asset.sql   (with v15 and v16+, having a prefix  -yyy.sql  will allow to have the table created during the first module yyy activation instead of always during installation.

aspangaro:
PR replaced by #20058 & #20059

PR #19786, opened 2022-01-11
2022.5 remainings

SylvainLegrand:
For orders shipped partially, the dispatchable order indicator does not take into account shipments already made.
The checking of the quantities in stock must be done only on the remainders!


aspangaro:
@SylvainLegrand Problem of merge from Easya.

Remaining on customer order is a new or fix PR ?

PR #19788, opened 2022-01-11
FIX user actions rights when multi-company transverse mode is enabled

lvessiller-opendsi:
FIX user actions rights when mulit-company transverse mode is enabled

You can't modify your own password or your own information (user card) if MULTICOMPANY_TRANSVERSE_MODE is enabled and your user is logged in the second entity for example.

### DIFF
--- a/htdocs/user/card.php
+++ b/htdocs/user/card.php
@@ -1905,7 +1905,7 @@ else
 					}
 				}
 
-				if ($caneditfield && (empty($conf->multicompany->enabled) || !$user->entity || ($object->entity == $conf->entity) || ($conf->global->MULTICOMPANY_TRANSVERSE_MODE && $conf->entity == 1)))
+				if ($caneditfield && (empty($conf->multicompany->enabled) || !$user->entity || ($object->entity == $conf->entity) || ($conf->global->MULTICOMPANY_TRANSVERSE_MODE && $object->entity == 1)))
 				{
 					if (!empty($conf->global->MAIN_ONLY_LOGIN_ALLOWED))
 					{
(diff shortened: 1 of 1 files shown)


htdocs/user/card.php
```diff
@@ -1905,7 +1905,7 @@
 					}
 				}
 
-				if ($caneditfield && (empty($conf->multicompany->enabled) || !$user->entity || ($object->entity == $conf->entity) || ($conf->global->MULTICOMPANY_TRANSVERSE_MODE && $conf->entity == 1)))
+				if ($caneditfield && (empty($conf->multicompany->enabled) || !$user->entity || ($object->entity == $conf->entity) || ($conf->global->MULTICOMPANY_TRANSVERSE_MODE && $object->entity == 1)))
```
eldy, line 1908: This replace a feature with another.
Old version was: if MULTICOMPANY_TRANSVERSE_MODE is on, we can modify any user whatever is its entity if we are connected on entity 1.
New version means: if MULTICOMPANY_TRANSVERSE_MODE is on, we can modify the user that was created into entity 1, even if we are connected onto another entity.
The definition of MULTICOMPANY_TRANSVERSE_MODE is still not clear for the moment (it was never documented on the page https://wiki.dolibarr.org/index.php?title=Module_MultiCompany, so i don't know which behaviour we want to have with this option so i don't know if i can validate or not). @hregis can you update the wiki page to define the rule on permission that brings this option ?

lvessiller-opendsi:
@hregis 
Do you think this fix is correct ?

lvessiller-opendsi:
@hregis
Could you check this fix ?

mverove:
Hello,
Any update on this ?
Thank you in advance.

philippe-opendsi:
Hi
@eldy in transverse mode, all the user are created in the master entity (1). It is not possible to create a user in other entities... 
User who have acces to any other entitie than 1 can not update their information even thy have the rights to do that. And sometimes they don' have access to the primary entity

hregis:
@philippe-opendsi @lvessiller-opendsi

I see that this weekend and send a feedback here

hregis:
@lvessiller-opendsi @mverove @philippe-opendsi @eldy 

Hello, MULTICOMPANY_TRANSVERSE_MODE activates the option of centralizing users in the main entity. This constant could be renamed because now it only has this functionality. Historically the code did not allow modifying a user card from a different entity. This is no longer the case with "$object->entity". This PR is OK !
 Regarding the creation of users from another entity does not have much interest since only a superadmin can do it. A simple administrator will not see the created user until the user is linked to a group and one or more entities.

lvessiller-opendsi:
@eldy 
Could you see to merge this PR ?

mverove:
Hi,

Is it possible to merge this PR please ?

Thank you.

lvessiller-opendsi:
@eldy 
Could you merge it ?

PR #19823, opened 2022-01-12
FIX Solving non-blocking bug in user login

harrywinner:
When using the function **check_user_password_dolibarr** to authenticate a user, a call is made to the function **dol_verifyHash** which requires a pre-calculated hash. This hash is not available with programmatically-created never-used accounts. Hence, in suce cases we receive the following warning:
_Notice: Trying to access array offset on value of type null in .../htdocs/core/lib/security.lib.php on line 156_
This fix solves that bug by avoiding that unnecessary call

PS: This is my first PR ever, to project I use a lot. Starting small. Thanks for your time building this.


eldy:
Welcome as a new Admiral among the Dolibarr army:   https://wiki.dolibarr.org/index.php/Dolibarr_Project

PR #19824, opened 2022-01-12
FIX : Fatal error on admin.lib

ksar-ksar:
Fatal error: Uncaught Error: Undefined constant” warning "in … / httpd.www / db / htdocs / core / lib / admin.lib.php: 655 Stack 
trace: 
    0 … / httpd.www / db / htdocs / admin / modules.php (501): modules_prepare_head (1, 75) 
    1 {main} thrown in … / httpd.www / db / htdocs / core / lib / admin.lib.php on line 655


ksar-ksar:
Already corrected by @frederic34  #19820

PR #19825, opened 2022-01-13
fix : preg_replace_callback by preg_replace

defrance:
FIX : error on login page

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -828,7 +828,7 @@ function checkVal($out = '', $check = 'alphanohtml', $filter = null, $options = 
 				// We replace chars from a/A to z/Z encoded with numeric HTML entities with the real char so we won't loose the chars at the next step (preg_replace).
 				// No need to use a loop here, this step is not to sanitize (this is done at next step, this is to try to save chars, even if they are
 				// using a non coventionnel way to be encoded, to not have them sanitized just after)
-				$out = preg_replace_callback('/&#(x?[0-9][0-9a-f]+;?)/i', 'realCharForNumericEntities', $out);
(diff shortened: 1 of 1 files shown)


frederic34:
realCharForNumericEntities is the callback function (in main.inc.php), it is not a replacement string

defrance:
Oups, you're right, 
in fact the realCharForNumericEntities function is present in main.inc.php and sometime when a GETPOST is call they are not include (then fail)
more infos here : https://www.dolibarr.fr/forum/t/preg-replace-callback/38836

htdocs/core/lib/functions.lib.php
```diff
@@ -828,7 +828,7 @@ function checkVal($out = '', $check = 'alphanohtml', $filter = null, $options =
 				// We replace chars from a/A to z/Z encoded with numeric HTML entities with the real char so we won't loose the chars at the next step (preg_replace).
 				// No need to use a loop here, this step is not to sanitize (this is done at next step, this is to try to save chars, even if they are
 				// using a non coventionnel way to be encoded, to not have them sanitized just after)
-				$out = preg_replace_callback('/&#(x?[0-9][0-9a-f]+;?)/i', 'realCharForNumericEntities', $out);
+				$out = preg_replace('/&#(x?[0-9][0-9a-f]+;?)/i', 'realCharForNumericEntities', $out);
```
eldy, line 831: preg_replace_callback and preg_replace
preg_replace replace string with the value 'realCharForNumericEntities'
preg_replace_callback repalce string with the result of the function realCharForNumericEntities()

defrance: more info on the french forum link

eldy:
I replaced with
$out = preg_replace_callback('/&#(x?[0-9][0-9a-f]+;?)/i', function ($m) {
					return realCharForNumericEntities($m); }, $out);
According to php doc, it should be the same. But well, if this form work when not the other one, as it is the same, let's use the one without warning.

PR #19826, opened 2022-01-13
FIX : several adaptations rank => rankorder and langs correction for HRM

atm-gauthier:


atm-gauthier:
Hi @eldy, do you know why i don't see same rights for GRH in 15.0 and in develop even if modHRM file is identical ?

15.0 :
image

develop :
image

eldy:
Try to pull then disable and reenable module. You should get this (note that translation in french is not yet good)
image

atm-gauthier:
Hi, i just deleted my database and re created one in doli 15.0, and i always get this :

image

I don't know why...

PR #19829, opened 2022-01-13
FIX search by user in supplier order list

lvessiller-opendsi:
FIX search by user in supplier order list
- if supplier orders are linked to user contacts and you want sarch by user in supplier list you have no results

No fileds are selected in "element_contact" and "c_type_contact" tables so I only add an EXISTS request to improve SQL performance

### DIFF
--- a/htdocs/fourn/commande/list.php
+++ b/htdocs/fourn/commande/list.php
@@ -561,7 +561,7 @@ if ($search_user > 0) {
 	$sql .= " SELECT ec.rowid ";
 	$sql .= " FROM " . MAIN_DB_PREFIX . "element_contact as ec";
 	$sql .= " INNER JOIN " . MAIN_DB_PREFIX . "c_type_contact as tc ON tc.rowid = ec.fk_c_type_contact";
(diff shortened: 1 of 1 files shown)


htdocs/fourn/commande/list.php
```diff
@@ -561,7 +556,15 @@
 if ($search_type_thirdparty) $sql .= " AND s.fk_typent IN (".$db->escape($search_type_thirdparty).')';
 if ($search_company) $sql .= natural_search('s.nom', $search_company);
 if ($search_sale > 0) $sql .= " AND s.rowid = sc.fk_soc AND sc.fk_user = ".$db->escape($search_sale);
-if ($search_user > 0) $sql .= " AND ec.fk_c_type_contact = tc.rowid AND tc.element='supplier_order' AND tc.source='internal' AND ec.element_id = cf.rowid AND ec.fk_socpeople = ".$db->escape($search_user);
+if ($search_user > 0) {
```
eldy, line 559: 👍🏻 
All requests should use this way to filter on child tables so we are able to manage filter without changing the parent request just by adding a WHERE condition (and this will help a working in progress feature that is the advanced filter tool).

PR #19835, opened 2022-01-14
FIX: Don't blindly unset api object notes

### LINKED ISSUE #19833: Member subscription notes are unavailable via api
# Bug
Member subscription notes are unavailable via api

## Environment
- **Version**: 14.0.4
- **OS**: Linux
- **Web server**: Apache 2
- **PHP**: 7.4
- **Database**: MariaDB 10.5
- **URL(s)**: /api/index.php/subscriptions/{id}

Expected behavior: the `note` subscription field, available in member subscription form, should also be available via api.

Actuel behavior: it is not :)
```
{
  "datec": 1234567890,
  "datem": 1234567890,
  "dateh": 1234567890,
  "datef": 1234567890,
  "fk_type": "1",
  "fk_adherent": "1234",
  "amount": "123.00000000",
  "fk_bank": null,
  "id": "12345",
  "entity": null,
  "import_key": null,
  "array_options": [],
  "array_languages": null,
  "linkedObjectsIds": null,
  "canvas": null,
  "fk_project": null,
  "contact_id": null,
  "user": null,
  "origin": null,
  "origin_id": null,
  "ref": "12345",
  "ref_ext": null,
  "statut": null,
  "status": null,
  "country_id": null,
  "country_code": null,
  "state_id": null,
  "region_id": null,
  "barcode_type": null,
  "barcode_type_coder": null,
  "mode_reglement_id": null,
  "cond_reglement_id": null,
  "demand_reason_id": null,
  "transport_mode_id": null,
  "shipping_method_id": null,
  "model_pdf": null,
  "last_main_doc": null,
  "fk_account": null,
  "note_public": null,
  "note_private": null,
  "total_ht": null,
  "total_tva": null,
  "total_localtax1": null,
  "total_localtax2": null,
  "total_ttc": null,
  "lines": null,
  "name": null,
  "lastname": null,
  "firstname": null,
  "civility_id": null,
  "date_creation": null,
  "date_validation": null,
  "date_modification": null,
  "specimen": 0
}
```
By the way, there a lot of useless null fields over there, and i notice `note_public`  and `note_private`.

I also notice that the `DolibarrApi::_cleanObjectDatas` method blindly remove the `note` field arguing that `note_public` and `note_private` are now used: https://github.com/Dolibarr/dolibarr/blob/cb45f79ab07a155ab24f03a1d6ba8c3f44792431/htdocs/api/class/api.class.php#L136

Problem is member subscription don't have such fields :)

Add a note to a member subscription, and accces it via api

nervo:
# FIX|Fix #19833

`DolibarrApi` class is extended by 32 classes.

17 of them already unset their object notes in their extended `_cleanObjectDatas` methods. These are:

- AgendaEvents
- Boms
- Categories
- Contacts
- Donations
- ExpenseReports
- Invoices
- Mos
- Orders
- Products
- Projects
- Proposals
- Shipments
- StockMovements
- Tasks
- Thirdparties
- Tickets

Take, for instance `AgendaEvents`:  https://github.com/Dolibarr/dolibarr/blob/1ae01ba2fcb9adeb762b812a70cb62c0421560e5/htdocs/comm/action/class/api_agendaevents.class.php#L359

For various reasons, 3 of them are irrelevants:

- Documents
- MyModuleApi
- Setup

2 of them has no `note` fields at all:

- Warehouses
- ZapierApi

1 one them has only a `note_public` field:

- BankAccounts

6 of them have both `note_public` and `note_private` fields:

- Contracts
- Interventions
- Members
- SupplierInvoices
- SupplierOrders
- Supplierproposals

1 of them have both `note` and `note_public` fields:

- Users

2 of them have *only* a `note` field:

- MembersTypes
- Subscriptions

By blindly unset object notes in parent method, the last two classes (`MembersTypes` and `Subscriptions`) lost their absolutely relevant `note` field when exposed via api.
This unsetting should happen on the children level, aka. extended `_cleanObjectDatas` method.

In facts, that's already happen in the first 17 classes group :)
In the other class groups, only `Users` *SHOULD*  *MAYBE*  (up to you) unset it, so that `MembersTypes` and `Subscriptions` could keep their `note` field.

### DIFF
--- a/htdocs/api/class/api.class.php
+++ b/htdocs/api/class/api.class.php
@@ -133,7 +133,6 @@ class DolibarrApi
 
 		unset($object->mode_reglement);		// We use mode_reglement_id now
 		unset($object->cond_reglement);		// We use cond_reglement_id now
-		unset($object->note);				// We use note_public or note_private now
 		unset($object->contact);			// We use contact_id now
 		unset($object->thirdparty);			// We use thirdparty_id or fk_soc or socid now
 


eldy:
Every object must now use note_public and / or note_private   (the property 'note' has been abandoned because it was not clear if it was for a public or private use).
Some of them may use only note_private and note_public only and not both. 
Into database, the field may have been named "note" for historical reason but we should start to rename it also with the good name note_public or note_private. 
But all objects must now use the property ->note_private or ->note_public and no more ->note (fix on database should be done later). I completed v15 with some patches to follow this direction. Property ->note must never be used anymore due to the confusion developer has because of the 2 different meanings it has.
This is why we unset always the 
unset($object->note); // alreaydy into note_private 
Developer must never use it anymore. So we must keep this. Instead if you miss the field note in api output, you can fix by setting the note_private or note_public property.

nervo:
I'm sorry @eldy , i don't get it :(
What do you mean by "you can fix by setting the note_private or note_public property" ? We have a lot of member subscriptions in database with a `note` field, and the subscription form itself use the `note` field. What exactly are we supposed to do ?

PR #19845, opened 2022-01-15
FIX: Stop removing all double quotes from requests

nervo:
# FIX
Stop removing all double quotes from requests

Because just... wtf ?!?

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -774,10 +774,9 @@ function checkVal($out = '', $check = 'alphanohtml', $filter = null, $options = 
 					// Remove html tags
 					$out = dol_string_nohtmltag($out, 0);
 					// Remove also other dangerous string sequences
-					// '"' is dangerous because param in url can close the href= or src= and add javascript functions.
 					// '../' or '..\' is dangerous because it allows dir transversals
 					// Note &#38, '&#0000038', '&#x26'... is a simple char like '&' alone but there is no reason to accept such way to encode input data.
-					$out = str_ireplace(array('&#38', '&#0000038', '&#x26', '&quot', '&#34', '&#0000034', '&#x22', '"', '&#47', '&#0000047', '&#92', '&#0000092', '&#x2F', '../', '..\\'), '', $out);
+					$out = str_ireplace(array('&#38', '&#0000038', '&#x26', '&quot', '&#34', '&#0000034', '&#x22', '&#47', '&#0000047', '&#92', '&#0000092', '&#x2F', '../', '..\\'), '', $out);
 				} while ($oldstringtoclean != $out);
 				// keep lines feed
 			}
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -774,10 +774,9 @@ function checkVal($out = '', $check = 'alphanohtml', $filter = null, $options =
 					// Remove html tags
 					$out = dol_string_nohtmltag($out, 0);
 					// Remove also other dangerous string sequences
-					// '"' is dangerous because param in url can close the href= or src= and add javascript functions.
 					// '../' or '..\' is dangerous because it allows dir transversals
 					// Note &#38, '&#0000038', '&#x26'... is a simple char like '&' alone but there is no reason to accept such way to encode input data.
-					$out = str_ireplace(array('&#38', '&#0000038', '&#x26', '&quot', '&#34', '&#0000034', '&#x22', '"', '&#47', '&#0000047', '&#92', '&#0000092', '&#x2F', '../', '..\\'), '', $out);
+					$out = str_ireplace(array('&#38', '&#0000038', '&#x26', '&quot', '&#34', '&#0000034', '&#x22', '&#47', '&#0000047', '&#92', '&#0000092', '&#x2F', '../', '..\\'), '', $out);
```
eldy, line 779: This introduce a serious security hole allowing to make sql injection and getting all content of database by any user.
For which use case do you need this ?

nervo: Well, as far as i know, sql injection has nothing to do with double quotes, and you can do it with simple quote :) In php world, you must just take care of escaping the input data with `mysql_real_escape_string`, or forge your request with pdo.
Anyway, my use case is having double quotes somewhere in fields, like names, comments, ... It's no so unusual :)

A peculiar use case is also having json fields, where `{"foo":"bar"}` become `{foo:bar}`...

eldy: Sure, sql or html injection can be done with simple or double quotes (depending on context) but there is some protection for the simple quotes (into the WAF and sanitizing security features) that is strict but that we can't use (for technical reason for the double quotes). Also mysql_real_escape_string is often not enough, a lot of advanced technique allows to by pass it. So we must sanitize the data has much as possible in the entry (so into the GETPOST). Reports of our hacker team working on the bug bounty program show it was really easy when allowing the double quotes.
If you need to enter a json field into a string, it means the field in the code must not be retreived with 
GETPOST('var', 'alphanohtml')  
but must use another parameter
like 
GETPOST('var', 'restricthtml')   for example.
A better enhancement would be to introduce a
GETPOST('var', 'json')
This method does not exist yet but could be introduced if necessary to validate the json is a valid json format and retreive the data with sanitization optimized for json.
There is not so much case that need the " when the parameter on the GETPOST is 'alphanohtml' (meaning we need a simple string). The case that realy need it use another parameter. And for the other case, this is a constraint that allow us to have HTML and SQL injection really more secured (without this protection, bypassing WAF or escapement later are too easy).

nervo: Hm, sound interesting, but a bit messy :) I will give `GETPOST('var', 'restricthtml')` a try!

PR #19848, opened 2022-01-15
Update emailcollector.class.php

randallmoraes:
Enable to init hook to emailcollectorcard, when docollect is executed by cron task.

[*When the method docollect is called by cron task, wont execute init hook because the (!is_object($hookmanager) return false, with this update both at execute docollect or call doCollectOneCollector, enable to call the hook... *]


eldy:
I pushed a different fix c46963410138ca70271fefa41d90ee31834510a9. Can you check this other fix is ok for you ?

PR #19856, opened 2022-01-17
Fix clone invoice

SylvainLegrand:
When you clone an invoice with external contacts, these contacts are copied to the new invoice even if you change third party.


eldy (commented):
Why did you remove the file testvirus.txt ?

SylvainLegrand:
I had to do this to clone the repository, otherwise my local Github desktop couldn't open it.

eldy:
I backported a fix done in v15 to make this file no more detected as virus (like coqueluche vaccin !)
So you should have no problem anymore with v14 clone too.
Can you repush the PR after merging ?

PR #19857, opened 2022-01-17
getTasksArray() fetch_optionals to populate ExtraFields

cfoellmann:
when using `$tasks->getTasksArray()` ExtraFields are not accessable in the returned object. Is the standard for other methods based on CommonObject

This populates `->array_options` with the ExtraFields


eldy:
Loading fetch_optionnal is the default for all fetch method.
But getTaskArray() is used to load massively data for different use cases. In such situation, we should optimize data access. Can you instead add a parameter $loadextrra into getTaskArray.

cfoellmann:
> Loading fetch_optionnal is the default for all fetch method. But getTaskArray() is used to load massively data for different use cases. In such situation, we should optimize data access. Can you instead add a parameter $loadextrra into getTaskArray.

Good idea!

cfoellmann:
@eldy I implemented your suggestion

PR #19863, opened 2022-01-17
FIX : right expensereport for admin

john-botella:
# FIX

On create expense report card,  the admin user with all rights in Dolibarr can't select users not in his hierarchy 
it's possible by enabling MAIN_USE_ADVANCED_PERMS conf and $user->rights->expensereport->writeall_advance right but activating advance rights to unlock possibility to an admin with ALL RIGHTS to create expense report for all usesr isn't logical.

### DIFF
--- a/htdocs/expensereport/card.php
+++ b/htdocs/expensereport/card.php
@@ -236,7 +236,8 @@ if (empty($reshook)) {
 		}
 		if (!$error) {
 			if (empty($conf->global->MAIN_USE_ADVANCED_PERMS) || empty($user->rights->expensereport->writeall_advance)) {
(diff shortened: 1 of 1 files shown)


eldy:
We had the same pb for leave request. And another way to fix this has been implemented for 15.0
The permission writeall_advance has been migrated as writeall. so now you can create a leave for someone else if you have this permission.
We must implement the expense report the same way, so i suggest to introduce writeall permission for expense report too. So all hr modules have same level of permission ( read / create is for oursself + hieerarchy, and readall / writeall is for user out of hierarchy).

john-botella:
@eldy thanks

PR #15802, opened 2020-12-22
FIX: Accountancy - Retire entire opening balance feature

aspangaro:
In French accounting, this functionality is dangerous, it takes all the entries and adds all the accounts without time and class limits (Class 6 and 7 accounts ???) and does not take into account the "a-nouveau" journal.

Please remove from Dolibarr for the moment

PR #19867, opened 2022-01-18
FIX opening balance unsafe code #15802

maboelfotoh:
# FIX #15802 safer implementation of opening balance code
This implementation of the opening balance feature utilizes fetchAllBalance() along with its filter parameters.


eldy:
@aspangaro Can you validate this PR ? Seems good for me because it used a hidden option. I don't understand goal of this option but well, as long as it is a hidden option, this is not a problem...

aspangaro:
@eldy I don't understand also this modification. I try it and I will confimed if this modification is good or not

maboelfotoh:
@eldy @aspangaro When I first implemented the 'opening balance' addition to balance.php, I put the SQL statements in balance.php, and the code potentially did not respect all filters selected by the user.

This PR removes the SQL from balance.php, and reuses the logic that already exists in bookkeeping.class.php, albeit with slight modification, in order to find the opening balance.

This a more efficient, cleaner implementation that respects the filters selected by the user.

eldy:
@maboelfotoh So this is a bug fix or code cleaning ?
If it fixes a bug, you can remove the hidden option and directly push the corrected version... (we don't need an option to say if optionwithbugfixed is on then...). This has a sens only if if have   if optionwithapossiblebug is on then ...

maboelfotoh:
@eldy @aspangaro It is both a bug fix and code cleaning.
This is to address the issues mentioned in [https://github.com/Dolibarr/dolibarr/pull/15802](https://github.com/Dolibarr/dolibarr/pull/15802). Here's a recap below:

@aspangaro noted the following when I first introduced this feature:
"In French accounting, this functionality is dangerous, it takes all the entries and adds all the accounts without time and class limits (Class 6 and 7 accounts ???) and does not take into account the "a-nouveau" journal.

Please remove from Dolibarr for the moment"

@eldy suggested that instead there would be a hidden option instead of removing the feature, which @aspangaro then added.

I have no issues with the hidden option introduced by @aspangaro. If you are comfortable with the feature and agree that it is no longer unsafe, then I would be more than happy to remove the hidden constant ACCOUNTANCY_SHOW_OPENING_BALANCE that was introduced by @aspangaro. Or at least change its default value to enabled. Otherwise, it is sufficient to have this update merged.

eldy:
The pb in submitting a PR with both a bug fix and code cleaning is that
code cleaning makes difficult to find what is code to just fix the bug so what is code to test to validate the fix. Also code cleaning should never be merged (can be posted but merge is delayed) during a beta period. And when both are in same PR, this is not possible. Can you resubmit the PR in 2.  1 for the bug fix only (must include the minimal code to remove the trouble), and one for the code cleaning.

maboelfotoh:
@eldy What I meant by 'code cleaning' is removal of the unsafe direct SQL statements in balance.php (the 'bug' being addressed), which has become dead code as a result of the fix (function calls from balance.php to code implemented in bookkeeping.class.php).

With that said, if you still think that this should be resubmitted as 2 PRs, kindly let me know.

eldy:
Ok, my understanding is the pr was a refactoring of code and this fixes a bug as a side effect. Because refactoring often introduce bugs too and create pb during merge with higher version , even if there is a good side effect, it should be pushed into develop.
Currently too many changes were done since pr submit and it is not no mere mergeable into develop branch.
Also may be trouble is already fixed with otherem changes done since.

maboelfotoh:
Alright, no worries. This lends itself to a new PR. Thank you.

PR #19884, opened 2022-01-20
Check if the lines in create is type object (like supplier invoice)

smil2k:
# Instructions

# FIX|Fix #[*#17058 REST API supplierorders error on creation with lines array*]

The api_supplier_order contained an object cast before calling the internals. This has been removed, thus the attributes in the array were not visible.

Applied the same method as in api_supplier_invoice.


eldy:
👍🏻

smil2k:
@eldy I'm new here, do I need to do anything to have it in develop also?

PR #19679, opened 2021-12-23
Feature Update #19678

alsoft10:
#19678 Now Dolibarr can give theme developer to write custom JavaScript code only applicable for theme. That the code scope will be inside the particular theme where it is written.

Issue #19678, opened 2021-12-23
Advanced Theme Customization ~ Allow writing JavsScript code inside the theme itself


alsoft10:
# Feature Request
Currently Dolibarr allow web developers to create new theme and customization of it. But current theme customization is very basic level. As you know theme is related to the Skin of any software. So theme development and design are done most of the time by dedicated web and graphic designers. If we want to design highly interactive theme, then JS code is very important inside the theme. Obviously that JS code should not conflict with main engine JavaScript code. When a particular Theme is activated only JS code related to that theme will only be activated (if it present). If the JS code is not present for that theme., no problem, stay cool and relaxed. My suggestion is given below to achieve this.

## Suggested implementation
To give Dolibarr developers more flexibility on theme design, we must let them to be able to add a JS file inside the theme root directory. That JS file will contain advanced level of user activity. They may add it if want or they may not add. To achieve this we have to follow the following steps.

## Suggested steps

1. STEP 1: I have created a Dolibarr constant, **_ALLOW_THEME_JS_**. This constant will act as extra layer of security. So if end user wants to activate JavaScript for a particular theme, irrespective of JS file is present or not inside of the theme directory, they must set to _**ALLOW_THEME_JS=1**_. Else they can set to ALLOW_THEME_JS=0 to disable unwanted execution of JS code.
2. STEP 2: Now user are allowed to run theme specific JS code by enabling ALLOW_THEME_JS=1. Now we need to create a JS file inside theme root directory with the same name as the theme. That is if you are working with theme **_eldy_**, JS file name will be **_eldy.js_**. Now Dolibarr engine can easily add that JS file automatically for us.
3. STEP 3: Now you can write any theme specific JS code at your file **_eldy.js_** and it will directly applied on the skin.

## Suggested Code
`//If you want to load custom javascript file from your selected theme directory
		if (!empty($conf->global->ALLOW_THEME_JS)) { 
			$theme_js = dol_buildpath('/theme/'.$conf->theme.'/'.$conf->theme.'.js', 0);
			if (file_exists($theme_js)) {
				print '<script src="'.DOL_URL_ROOT.'/theme/'.$conf->theme.'/'.$conf->theme.'.js'.($ext ? '?'.$ext : '').'"></script>'."\n";
			}
		}`

alsoft10:
@eldy I am actively waiting for your feedback on this. Because this feature will make Dolibarr more powerful.

PR #19885, opened 2022-01-20
PR REQUEST #19679

alsoft10:
#19678 Now Dolibarr can give theme developer to write custom JavaScript code only applicable for theme. That the code scope will be inside the particular theme where it is written.


alsoft10:
@eldy I have created PR on develop branch as per your request.

PR #19897, opened 2022-01-20
fix docker build setup

smil2k:
# FIX|Fix #[*19896: Fix xdebug in docker build *]

### DIFF
--- a/build/docker/Dockerfile
+++ b/build/docker/Dockerfile
@@ -40,8 +40,11 @@ RUN chmod +x /usr/local/bin/docker-run.sh
 RUN pecl install xdebug && docker-php-ext-enable xdebug
 RUN echo 'zend_extension="/usr/local/lib/php/extensions/no-debug-non-zts-20180731/xdebug.so"' >> ${PHP_INI_DIR}/php.ini
 RUN echo 'xdebug.mode=debug' >> ${PHP_INI_DIR}/php.ini
+RUN echo 'xdebug.start_with_request=yes' >> ${PHP_INI_DIR}/php.ini
 RUN echo 'xdebug.client_host=host.docker.internal' >> ${PHP_INI_DIR}/php.ini
 RUN echo 'xdebug.client_port=9003' >> ${PHP_INI_DIR}/php.ini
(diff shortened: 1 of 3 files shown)


build/docker/Dockerfile
```diff
@@ -39,15 +39,10 @@ RUN chmod +x /usr/local/bin/docker-run.sh
...
 RUN pecl install xdebug && docker-php-ext-enable xdebug
 RUN echo 'zend_extension="/usr/local/lib/php/extensions/no-debug-non-zts-20180731/xdebug.so"' >> ${PHP_INI_DIR}/php.ini
-RUN echo 'xdebug.remote_autostart=1' >> ${PHP_INI_DIR}/php.ini
-RUN echo 'xdebug.remote_enable=1' >> ${PHP_INI_DIR}/php.ini
-RUN echo 'xdebug.default_enable=1' >> ${PHP_INI_DIR}/php.ini
-#RUN echo 'xdebug.remote_host=docker.host' >> ${PHP_INI_DIR}/php.ini
-RUN echo 'xdebug.remote_port=9000' >> ${PHP_INI_DIR}/php.ini
-RUN echo 'xdebug.remote_connect_back=1' >> ${PHP_INI_DIR}/php.ini
-RUN echo 'xdebug.profiler_enable=0' >> ${PHP_INI_DIR}/php.ini
-RUN echo 'xdebug.remote_log="/tmp/xdebug.log"' >> ${PHP_INI_DIR}/php.ini
```
eldy, line 49: Do we really need to remove this line (same question for other removed line) ?

smil2k: I just used the settings suggested by IDE. 

Now i've checked the following guide: https://xdebug.org/docs/upgrade_guide:
autostart =  start_with_request
connect_back = discover_client_host

The following three is not strictly needed:
1) I've left the host.docker.internal address for safety, in case discovery does not work.
2) I've commented out the xdebug.log because it is a really chatty log for debugging xdebug problems. 
3) Added idekey because it is needed for netbeans, but ignored by eclipse. 

I would keep these if it is ok with you (unless somebody has a problem with it).

eldy: Thanks for feedback. We will make a try with your version.

PR #19909, opened 2022-01-24
FIX : Restricted area on bank card

ghost:
# FIX
Missing `$feature` which cause restricted area on the card of bank

image

### DIFF
--- a/htdocs/core/lib/security.lib.php
+++ b/htdocs/core/lib/security.lib.php
@@ -615,9 +615,9 @@ function checkUserAccessToObject($user, array $featuresarray, $objectid = 0, $ta
 		if ($feature == 'task') {
 			$feature = 'projet_task';
 		}
-        if ($feature == 'banque') {
-            $feature = 'fk_account@bank_account';
-        }
+		if ($feature == 'banque') {
+			$feature = 'fk_account@bank_account';
+		}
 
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/security.lib.php
```diff
@@ -615,6 +615,9 @@ function checkUserAccessToObject($user, array $featuresarray, $objectid = 0, $ta
 		if ($feature == 'task') {
 			$feature = 'projet_task';
 		}
+		if ($feature == 'banque') {
```
eldy, line 618: Not sure it is good.
When feature = banque it means that object should be a Account so e should check access rowid with a field rowid an not fk_account.

We must force the $feature = 'fk_account@bank_account';  when object is a line of llx_bank, but in such a case, the $feature should not be banque (already used for bank_account table) but something else.
Can you provide the url of the page that fails ?

atm-gregr: The URL is Dolibarr/compta/bank/card.php

eldy: I don't see any trouble with current code in current v14 branch.
When param id=3, I get this sql to know if the checkUserAccessToObject() method
SELECT COUNT(dbt.rowid) as nb FROM llx_bank_account as dbt WHERE dbt.rowid IN (3) AND dbt.entity IN (1)
And this is what is expected.
What do you have without your PR ?

atm-gregr: I think that I did a mistake

PR #19919, opened 2022-01-25
NEW stock filter in reassort lists

lvessiller-opendsi:
NEW stock filter in reassort lists and add "Without category" in category filter

- in product stock list
image

- in poduct batch stock list
image

And add filter on "-- without category -- "


htdocs/product/reassort.php
```diff
@@ -323,32 +343,6 @@
 		print '</div>';
 	}
 
-	$param = '';
```
eldy, line 326: Are you sure we can remove this ? Don't we need it for the 
print_liste_field_titre("Ref", $_SERVER["PHP_SELF"], "p.ref", '', $param, "", $sortfield, $sortorder);
later ?

lvessiller-opendsi:
@eldy 
Could you check and merge this pull request ?

PR #19923, opened 2022-01-25
NEW - Add param color button action

BB2A-Anthony:
# NEW|New [*Add param color button action*]
[*Long description*]


htdocs/theme/md/style.css.php
```diff
@@ -439,7 +444,7 @@
 	margin-top:1px;
 }
 #mainbody input.button:not(.buttongen):not(.bordertransp), #mainbody a.button:not(.buttongen):not(.bordertransp) {
-	background: var(--butactionbg);
+	background: rgb(<?php echo $butactionbg ?>);
```
eldy, line 447: You must keep the css var
--butactionbg
and make the 
<?php echo $butactionbg ?>
where the css var is defined, so at one place only in the section  :root {   of the css file.

PR #19932, opened 2022-01-26
FIX : Replace restricthtml to nohtml in export .

atm-jpb:
# Instructions
Export 
when we export a date field with a filter like <20220101
the getpost restricthtml test does not return the value of the field.
we replace it with nohtml


eldy:
Can you try  'alphawithlgt'   instead of   'nohtml'   ?

atm-jpb:
ok for me. made changes

PR #19941, opened 2022-01-26
NEW - Pdf propal sender address under logo

BB2A-Anthony:
# NEW|New [*Pdf propal sender address under logo*]

128733770-32c3b12d-c765-4953-9e24-30c7910b72d5


eldy:
We may introduce tons of such options.
But we must do it only if it bring a real gain for end user. Otherwise we will have thousands of option making the end user lost.
I don't think this features brings a too important difference compared to default feature to justify to be added in the core. The core must contains "default and minimal" feature. Adding too many options is a dangerous direction for the core. How many users ABSOLUTELY claimed this feature. If we are less than 2%, we should not add it as a default feature. 

So I recommend to address this need of enhancement with an external module like it is done for example with the module "UltimatePDF":  it brings its own PDF template + a page with tons of options managed by this tempate to change positions of things.

BB2A-Anthony:
> We may introduce tons of such options. But we must do it only if it bring a real gain for end user. Otherwise we will have thousands of option making the end user lost. I don't think this features brings a too important difference compared to default feature to justify to be added in the core. The core must contains "default and minimal" feature. Adding too many options is a dangerous direction for the core. How many users ABSOLUTELY claimed this feature. If we are less than 2%, we should not add it as a default feature.
> 
> So I recommend to address this need of enhancement with an external module like it is done for example with the module "UltimatePDF": it brings its own PDF template + a page with tons of options managed by this tempate to change positions of things.

OK i understand 
@eldy In this case, rather put it by default for a more usual and conventional presentation

PR #19946, opened 2022-01-26
NEW add datem and type parameters to API to create movements

### LINKED ISSUE #19944: Movement API does not provide a possiblity to add the date of movement
### Feature Request

Initial upload of movelements is not possible, because the value date cannot be specified on the API

### Use case

- Initial upload of past data requires the date of movement. 
- Eventual synchronization from our source of movement might happened in the past.

### Suggested implementation

REST api should add a datem attribute with a date.

### Suggested steps

_No response_

smil2k:
# Close
#19944 - add possibility to specify movement date


smil2k:
Thank you for fixing the type parameter

PR #19962, opened 2022-01-28
FIX pay value in TakePOS

andreubisquerra:
After a pay values commit we need fix this value


eldy:
Do we have to backport this into v15 branch ?

andreubisquerra:
Yes, branch 15 is also affected, although it is not a serious problem, it is recommended.

PR #19966, opened 2022-01-30
Fix "Warning: This option make TCPDF library being crazy sometime and content on image disappeared!"

habot-it:


htdocs/core/lib/pdf.lib.php
```diff
@@ -713,6 +713,7 @@ function pdf_pagehead(&$pdf, $outputlangs, $page_height)
 		if (file_exists($filepath)) {
 			$pdf->SetAutoPageBreak(0, 0); // Disable auto pagebreak before adding image
 			$pdf->Image($filepath, (isset($conf->global->MAIN_USE_BACKGROUND_ON_PDF_X) ? $conf->global->MAIN_USE_BACKGROUND_ON_PDF_X : 0), (isset($conf->global->MAIN_USE_BACKGROUND_ON_PDF_Y) ? $conf->global->MAIN_USE_BACKGROUND_ON_PDF_Y : 0), 0, $page_height);
+			$pdf->SetPageMark(); // Fix "Warning: This option make TCPDF library being crazy sometime and content on image disappeared!"
```
eldy, line 716: What is the goal of adding this instruction if adding it makes the image disappeared ?
Is it to avoid the image disappear ? If yes, why adding a comment to warn the developer that if make image disappeared ?

habot-it: Hello @eldy,
On this page: https://wiki.dolibarr.org/index.php/Setup_Other
Note that the MAIN_USE_BACKGROUND_ON_PDF option sometimes has unwanted behavior.
This entry fixes the problem.
The image is well placed as a background on all pages.
(All multicells are displayed correctly)

eldy: So the comment you put into the code is not to war that the instruction introduce a bug but to say what it fixes ? If yes, it is in fact the comment that we should have into the label of the PR but not into the code.
So can you remove the comment from code ?  (comment in code are instructions to explain current code or give idea for future not to explain what was behaviour before we add code. This can already be followed with git change and commit log)

PR #19977, opened 2022-02-02
Add column "fk_parent_line" to llx_mrp_mo

atm-lena:
# NEW|New [*Add column "fk_parent_line" to llx_mrp_mo*]

Data to develop new feature "Manufacturing Order Child". We would like two things :

**1.** A specific link between Manufacturing Order and Manufacturing Order line parent for futur developments
**2.**  "Manufacturing Order line Parent" in card and list information

I have two proposals : 

**1. Create 2 columns "fk_parent_line" and "fk_parent" in llx_mrp_mo**

We have a specific link between line and child for futur developments.
Manufacturing Order Parent displays like a field without specific behaviour.

**2. Create 1 column "fk_parent_line" and develop the display of Manufacturing Order Parent**

We have a specific link between line and child for futur developments.
Manufacturing Order Parent diplays with a specific behaviour.

Which proposition do you think is better @eldy  ? For now, I applied solution 2 in merge request.

### DIFF
--- a/htdocs/install/mysql/migration/15.0.0-16.0.0.sql
+++ b/htdocs/install/mysql/migration/15.0.0-16.0.0.sql
@@ -94,3 +94,5 @@ CREATE TABLE llx_stock_mouvement_extrafields (
     fk_object integer NOT NULL,
     import_key varchar(14)
 )ENGINE=innodb;
+
+ALTER TABLE llx_mrp_mo ADD COLUMN fk_parent_line integer;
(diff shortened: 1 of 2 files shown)


eldy:
In scenario 2, the field fk_parent_line would be a reference/link to which table and column ?

atm-lena:
For both scenario : "fk_parent_line" will be an column in table llx_mrp_mo reference/link to "rowid" column of llx_mrp_production table for each line with "role" value "toconsume".

eldy:
Let's try with solution 2.

PR #19979, opened 2022-02-02
fix bad array index for multicurrency

atm-greg:
# FIX bad index for multicurrency arrays
When converting a supplier credit note into reduc, the multicurrency amounts are lost because of the generation of a key that is potentially different from $line->tva_tx for multicurrency arrays.

This key must be the same as amounts arrays to be sure to get the good multicurrency amounts when creating the discount.

Consequence : when multicurrency is enabled and a credit note is applied on a supplier invoice, the remainToPay is calculated on the multicurrency amounts which as 0.

ksnip_20220202-154610

P.S. : excuse my bad english.

### DIFF
--- a/htdocs/fourn/facture/card.php
+++ b/htdocs/fourn/facture/card.php
@@ -527,13 +527,14 @@ if (empty($reshook)) {
 			$i = 0;
 			foreach ($object->lines as $line) {
 				if ($line->product_type < 9 && $line->total_ht != 0) { // Remove lines with product_type greater than or equal to 9 and no need to create discount if amount is null
+					$keyforvatrate = $line->tva_tx.($line->vat_src_code ? ' ('.$line->vat_src_code.')' : '');
 
-					$amount_ht[$line->tva_tx] += $line->total_ht;
-					$amount_tva[$line->tva_tx] += $line->total_tva;
-					$amount_ttc[$line->tva_tx] += $line->total_ttc;
-					$multicurrency_amount_ht[$line->tva_tx] += $line->multicurrency_total_ht;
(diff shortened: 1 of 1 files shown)


htdocs/fourn/facture/card.php
```diff
@@ -527,14 +527,13 @@
...
 				if ($line->product_type < 9 && $line->total_ht != 0) { // Remove lines with product_type greater than or equal to 9 and no need to create discount if amount is null
-					$keyforvatrate = $line->tva_tx.($line->vat_src_code ? ' ('.$line->vat_src_code.')' : '');
 
 					$amount_ht[$line->tva_tx] += $line->total_ht;
 					$amount_tva[$line->tva_tx] += $line->total_tva;
 					$amount_ttc[$line->tva_tx] += $line->total_ttc;
-					$multicurrency_amount_ht[$keyforvatrate] += $line->multicurrency_total_ht;
-					$multicurrency_amount_tva[$keyforvatrate] += $line->multicurrency_total_tva;
-					$multicurrency_amount_ttc[$keyforvatrate] += $line->multicurrency_total_ttc;
+					$multicurrency_amount_ht[$line->tva_tx] += $line->multicurrency_total_ht;
```
eldy, line 534: If i compare with the customer side, it seems the fix must be done into the opisite direction:
Replacing  $line->tva_tx  with  $keyforvatrate  into  $amount_xxx[]  instead of changing the multicurrency_amount_ht

atm-greg: @eldy changes have been done according to your recommendations

PR #19984, opened 2022-02-03
Fix: prevent multiple counting of physical stock when product assigned to multiple categories

### LINKED ISSUE #19983: Invalid physical stock in product stock list when assigned to multiple categories
### Bug

When a product is assigned to multiple categories, the physical stock in the product list is being multiplicated.

### Environment Version

14.0.4

### Environment OS

Windows, Linux

### Environment Web server

Apache, NGINX

### Environment PHP

7.4.3

### Environment Database

MariaDB 10.3.25

### Environment URL(s)

/product/reassort.php

physical stock should show the correct value

Assign multiple categories to a product and look at product list

### Attached files

_No response_

code-chicken:
# |Fix #19983]
Invalid physical stock in product stock list when assigned to multiple categories

Added "distinct" keyword to SUM function to prevent counting the physical amount multiple times

### DIFF
--- a/htdocs/product/reassort.php
+++ b/htdocs/product/reassort.php
@@ -131,7 +131,7 @@ $htmlother = new FormOther($db);
 $sql = 'SELECT p.rowid, p.ref, p.label, p.barcode, p.price, p.price_ttc, p.price_base_type, p.entity,';
 $sql .= ' p.fk_product_type, p.tms as datem,';
 $sql .= ' p.duration, p.tosell as statut, p.tobuy, p.seuil_stock_alerte, p.desiredstock,';
-$sql .= ' SUM(s.reel) as stock_physique';
+$sql .= ' SUM(distinct s.reel) as stock_physique';
 if (!empty($conf->global->PRODUCT_USE_UNITS)) {
 	$sql .= ', u.short_label as unit_short';
 }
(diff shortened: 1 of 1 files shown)


htdocs/product/reassort.php
```diff
@@ -131,7 +131,7 @@
 $sql = 'SELECT p.rowid, p.ref, p.label, p.barcode, p.price, p.price_ttc, p.price_base_type, p.entity,';
 $sql .= ' p.fk_product_type, p.tms as datem,';
 $sql .= ' p.duration, p.tosell as statut, p.tobuy, p.seuil_stock_alerte, p.desiredstock,';
-$sql .= ' SUM(s.reel) as stock_physique';
+$sql .= ' SUM(distinct s.reel) as stock_physique';
```
eldy, line 134: If there is two lines with stock 2 (2 in warehouse A and 2 in warehouse B), you will get 2 instead of 4

You should have no problem of duplicate because of link with category because there is a test 
AND cp.fk_categorie = id category
So record link with product id and category id is always unique.
May be your trouble is the one fixed by this commit 15dca77330e3412e7fb445a20f3aecbf32c84eac ?

code-chicken: Hi Eldy,

I just installed a clean copy of Dolibarr 14.0.5, created two categories, one product and one warehouse.
I assigned 10 units to the product for this warehouse.
I assigned both categories to this product.
When I click on the search symbol in /product/reassort.php shows 20 units, exactly the double amount.
So this is indeed a bug which has to be solved.

image^
image
image

eldy: Can you paste the sql you have in the list ?  Did you apply the fix of https://github.com/Dolibarr/dolibarr/commit/15dca77330e3412e7fb445a20f3aecbf32c84eac ?

code-chicken: Thank you. The problem was solved with your patch.

PR #19994, opened 2022-02-04
fix : duration in hour and min (not in second)

defrance:
FIX ; use convertSecondToTime for totalise duration


htdocs/fichinter/list.php
```diff
@@ -741,7 +741,7 @@
 			if (!$i) {
 				$totalarray['pos'][$totalarray['nbfield']] = 'fd.duree';
 			}
-			$totalarray['val']['fd.duree'] += $obj->duree;
+			$totalarray['val']['fd.duree'] += convertSecondToTime($obj->duree, 'allhourmin');
```
eldy, line 744: $totalarray['val']['fd.duree'] is a value in second. You can't add a string (result of convertSecondToTime) to and int.
The convertSecondToTime must be used at output of value.

defrance: oki propose a second way with a type and a specific print by type
maybe use another tip as "count" ans "sum" to diplay some other value in totalisation list

PR #19997, opened 2022-02-05
Add options in function pdf_pagefoot of pdf.lib.php

habot-it:
PDF_FOOTER_BACKGROUND_COLOR:
This option inserts a colored banner on the footer.
The free text area is not affected.
Example value (blue): 6,65,184

PDF_FOOTER_TEXT_COLOR:
This option changes the color of the footer text as well as the free text.
The color of the free text can be changed in the WYSIWYG editor.
Example value (white): 255,255,255

PDF_FOOTER_TOP_MARGIN:
This option changes the padding between the footer and the free text.
Default value : 1

### DIFF
--- a/htdocs/core/lib/pdf.lib.php
+++ b/htdocs/core/lib/pdf.lib.php
@@ -1154,8 +1154,6 @@ function pdf_pagefoot(&$pdf, $outputlangs, $paramfreetext, $fromcompany, $marge_
 		}
 	}
 
-	$pdf->SetAutoPageBreak(0, 0); // Disable auto pagebreak
-
 	// For customize footer
 	if (is_object($hookmanager)) {
 		$parameters = array('line1' => $line1, 'line2' => $line2, 'line3' => $line3, 'line4' => $line4, 'outputlangs'=>$outputlangs);
@@ -1171,7 +1169,9 @@ function pdf_pagefoot(&$pdf, $outputlangs, $paramfreetext, $fromcompany, $marge_
 			// Option for footer background color (without freetext zone)
 			if (!empty($conf->global->PDF_FOOTER_BACKGROUND_COLOR)) {
 				list($r, $g, $b) = sscanf($conf->global->PDF_FOOTER_BACKGROUND_COLOR, '%d, %d, %d');
+				$pdf->SetAutoPageBreak(0, 0); // Disable auto pagebreak
 				$pdf->Rect(0, $dims['hk'] - $posy + $freetextheight, $dims['wk'] + 1, $marginwithfooter + 1, 'F', '', $fill_color = array($r, $g, $b));
+				$pdf->SetAutoPageBreak(1, 0); // Restore pagebreak
 			}
 
 			if ($line) {	// Free text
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/pdf.lib.php
```diff
@@ -1149,6 +1154,8 @@ function pdf_pagefoot(&$pdf, $outputlangs, $paramfreetext, $fromcompany, $marge_
 		}
 	}
 
+	$pdf->SetAutoPageBreak(0, 0); // Disable auto pagebreak
```
eldy, line 1157: If you disable pagebreak here, all content output after that are too long (for example the free text or content from pdf_pagefoot hook) may not generate the required pagebreak. Right ?

habot-it: Yes, pagebreak is disable for hook and default.

eldy: So it means we introduce a bug when PDF_ALLOW_HTML_FOR_FREE_TEXT has a high content ?

habot-it: Personally, when my free text or custom footer is too high. I modify the values of MAIN_PDF_FREETEXT_HEIGHT and MAIN_PDF_MARGIN_BOTTOM.

If you want, I can add an option : PDF_FOOTER_DISABLE_PAGEBREAK ?

habot-it: Hello @eldy,
What do you think of my changes ?
Thank

htdocs/core/lib/pdf.lib.php
```diff
@@ -1172,17 +1185,35 @@ function pdf_pagefoot(&$pdf, $outputlangs, $paramfreetext, $fromcompany, $marge_
...
+			}
+
+			// Option for set top margin height of footer after freetext
+			if (!empty($conf->global->PDF_FOOTER_TOP_MARGIN) || ($conf->global->PDF_FOOTER_TOP_MARGIN === '0')) {
+				$posy -= $conf->global->PDF_FOOTER_TOP_MARGIN;
+			} else { $posy--; }
+
+			if ($conf->global->PDF_FOOTER_DISABLE_PAGEBREAK === '1') { $pdf->SetAutoPageBreak(0, 0); } // Option for disable auto pagebreak
+			$pdf->writeHTMLCell($pdf->page_largeur - $pdf->margin_left - $pdf->margin_right, $mycustomfooterheight, $dims['lm'], $dims['hk'] - $posy, dol_htmlentitiesbr($mycustomfooter, 1, 'UTF-8', 0));
+			$pdf->SetAutoPageBreak(1, 0); // Restore pagebreak
```
eldy, line 1201: Because you disabled pagebreak only 
`if ($conf->global->PDF_FOOTER_DISABLE_PAGEBREAK === '1') {`
you must alos enable it 
if ($conf->global->PDF_FOOTER_DISABLE_PAGEBREAK === '1') {

habot-it: Hello @eldy like this ?
Thank

PR #20001, opened 2022-02-06
FIX Propal - Missing field import_key in import propal line

aspangaro:
Additional correction to #20000 

image


eldy:
Already pushed into v14

PR #19952, opened 2022-01-27
NEW : optional display warning icons on ticket list

thomas-Ngr:
Optionnaly displays warnings on ticket list when delays are configured.

Introduces two new options :

    TICKET_DELAY_BEFORE_FIRST_RESPONSE : tickets should be answered before this delay (hours)
    TICKET_DELAY_SINCE_LAST_RESPONSE : when answered, tickets should have another answer before this delay (hours).

PR #20013, opened 2022-02-07
New : ticket add last_msg_sent field

thomas-Ngr:
On ticket database, add a 'last_msg_date' field that is updated each time a message is sent from the ticket interface to the customer.

This will allow to display warnings for tickets that have not been answered, see PR #19952

### DIFF
--- a/htdocs/core/triggers/interface_50_modTicket_TicketEmail.class.php
+++ b/htdocs/core/triggers/interface_50_modTicket_TicketEmail.class.php
@@ -279,7 +279,10 @@ class InterfaceTicketEmail extends DolibarrTriggers
 						if ($mailfile->error) {
 							dol_syslog($mailfile->error, LOG_DEBUG);
 						} else {
-								  $result = $mailfile->sendfile();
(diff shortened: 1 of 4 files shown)


htdocs/core/triggers/interface_50_modTicket_TicketEmail.class.php
```diff
@@ -279,7 +279,10 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
 						if ($mailfile->error) {
 							dol_syslog($mailfile->error, LOG_DEBUG);
 						} else {
-								  $result = $mailfile->sendfile();
+							$result = $mailfile->sendfile();
+							// update last_msg_sent date
```
eldy, line 283: May be you should add a test on result on ->sendfile to update only if email sending is ok ?

thomas-Ngr: Thanks, I also added a test on ticket->sendTicketMessageByEmail() .

PR #20015, opened 2022-02-07
fix: if capital is not numeric just display input value

FHenry:
In company Setting set Capital to "Variable", generate a PDF (Quotation, Orde, Invoice) in the footer there is 
"Capital de" with nothing. 

$tmpamounttoshow is a return of price2num call,if it's not a number the function return "".

### DIFF
--- a/htdocs/core/lib/pdf.lib.php
+++ b/htdocs/core/lib/pdf.lib.php
@@ -1064,7 +1064,7 @@ function pdf_pagefoot(&$pdf, $outputlangs, $paramfreetext, $fromcompany, $marge_
 		if (is_numeric($tmpamounttoshow) && $tmpamounttoshow > 0) {
 			$line3 .= ($line3 ? " - " : "").$outputlangs->transnoentities("CapitalOf", price($tmpamounttoshow, 0, $outputlangs, 0, 0, 0, $conf->currency));
 		} else {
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/pdf.lib.php
```diff
@@ -1063,8 +1063,8 @@ function pdf_pagefoot(&$pdf, $outputlangs, $paramfreetext, $fromcompany, $marge_
 		$tmpamounttoshow = price2num($fromcompany->capital); // This field is a free string
 		if (is_numeric($tmpamounttoshow) && $tmpamounttoshow > 0) {
 			$line3 .= ($line3 ? " - " : "").$outputlangs->transnoentities("CapitalOf", price($tmpamounttoshow, 0, $outputlangs, 0, 0, 0, $conf->currency));
-		} else {
-			$line3 .= ($line3 ? " - " : "").$outputlangs->transnoentities("CapitalOf", $tmpamounttoshow, $outputlangs);
+		} elseif (!empty($fromcompany->capital)) {
+			$line3 .= ($line3 ? " - " : "").$outputlangs->transnoentities("CapitalOf", $fromcompany->capital, $outputlangs);
```
eldy, line 1067: If capital stored in table llx_societe is a float number.
So to avoid to have Capital: 123.0000 (with decimals to 0), we must use the price2num.
However, you can add also price() to format it in correct language.

PR #20016, opened 2022-02-07
fix: in some case index do not exists and return an error

FHenry:
fix: in some case index do not exists and return an error


htdocs/core/class/extrafields.class.php
```diff
@@ -699,7 +699,7 @@ public function update($attrname, $label, $type, $length, $elementtype, $unique
 					if ($unique) {
 						$sql = "ALTER TABLE ".MAIN_DB_PREFIX.$table." ADD UNIQUE INDEX uk_".$table."_".$attrname." (".$attrname.")";
 					} else {
-						$sql = "ALTER TABLE ".MAIN_DB_PREFIX.$table." DROP INDEX uk_".$table."_".$attrname;
+						$sql = "ALTER TABLE ".MAIN_DB_PREFIX.$table." DROP INDEX IF EXISTS uk_".$table."_".$attrname;
```
eldy, line 702: The result $resql of the query is not tested. So having this error should not be a trouble.
Also are your sure this sql is compatible with postgresql ?

FHenry: Yes it is compatible : https://www.postgresql.org/docs/9.3/sql-dropindex.html
and if Dolibarr do not test it, mysql/pgsql return error and for those who monitor sql error from database engine it can be false error warning.
By the you'r right I've just add the test result

PR #20018, opened 2022-02-08
fix : extrafield of line was scratch into the number column

grandoc:


eldy:
On which page/url do you have a trouble ?

grandoc:
on fichinter/card.php if you use an extrafield of line, and the numbering of lines is on. The PR I have made face the issue.

eldy:
Do we still need this PR ? I think bug has been fixed differently ?

PR #20019, opened 2022-02-08
New MO child

atm-lena:
# NEW|New Create MO Child

@eldy  I have a problem.

I would like to display MO parent on MO List and I use getNomURL in MOLine (give MO parent link, no MOLine link). 
If field type is a link, we need showcombobox value, but MO Line object dont' have it (fk_parent_line is a MO object field).

image

I see two choices : 

1. Create a new type InPutField that allows to have an input field parent/child link

2.  Go back on the data base decision  (https://github.com/Dolibarr/dolibarr/pull/19977 )  and add two column : fk_parent on llx_mrp_mo and fk_parent_line on llx_mrp_production to display with current functions

If we choose solution 1, creation of new input field type and MO parent display will be done in second time. 
But if we choose solution 2, I need to do another process to create MO child. 

That’s why I need an answer at this point.

Thank you !

P.S : don't merge this PR, it's not over

### DIFF
--- a/htdocs/mrp/class/mo.class.php
+++ b/htdocs/mrp/class/mo.class.php
@@ -1216,7 +1216,7 @@ class Mo extends CommonObject
 		$objectline = new MoLine($this->db);
 
 		$TFilters = array('customsql'=>'fk_mo = '.((int) $this->id));
-		if(!empty($rolefilter)) $TFilters['role'] = $rolefilter;
+		if (!empty($rolefilter)) $TFilters['role'] = $rolefilter;
 		$result = $objectline->fetchAll('ASC', 'position', 0, 0, $TFilters);
 
 		if (is_numeric($result)) {
@@ -1343,7 +1343,7 @@ class Mo extends CommonObject
 		print '<td class="center">'.$langs->trans('DisableStockChange').'</td>';
 		print '<td class="center">'.$langs->trans('MoChildGenerate').'</td>';
 		//print '<td class="center">'.$form->showCheckAddButtons('checkforselect', 1).'</td>';
-//		print '<td class="center"></td>';
+		//      print '<td class="center"></td>';
 		print '</tr>';
 		$i = 0;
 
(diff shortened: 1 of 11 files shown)


htdocs/mrp/class/mo.class.php
```diff
@@ -1648,4 +1667,92 @@ public function delete(User $user, $notrigger = false)
...
+		global $conf, $langs, $hookmanager;
+
+		if (!empty($conf->dol_no_mouse_hover)) {
+			$notooltip = 1; // Force disable tooltips
+		}
+
+		$result = '';
+
+		$moparent = new Mo($this->db);
+		$moparent->fetch($this->fk_mo);
```
eldy, line 1692: fetch must be done only if $this->fk_mo > 0

htdocs/mrp/class/mo.class.php
```diff
@@ -1648,4 +1667,92 @@ public function delete(User $user, $notrigger = false)
...
+	 *  Return a link to the object card (with optionaly the picto)
+	 *
+	 *  @param  int     $withpicto                  Include picto in link (0=No picto, 1=Include picto into link, 2=Only picto)
+	 *  @param  string  $option                     On what the link point to ('nolink', '', 'production', ...)
+	 *  @param  int     $notooltip                  1=Disable tooltip
+	 *  @param  string  $morecss                    Add more css on link
+	 *  @param  int     $save_lastsearch_value      -1=Auto, 0=No save of lastsearch_values when clicking, 1=Save lastsearch_values whenclicking
+	 *  @return	string                              String with URL
+	 */
+	public function getNomUrl($withpicto = 0, $option = '', $notooltip = 0, $morecss = '', $save_lastsearch_value = -1)
```
eldy, line 1681: Here the method getNomUrl is inside the MoLine. So it MUST return the ref of a line.
But a MO line has no ref, except its id.
You don't need to add such a method here. 
You mus tuse the existing method getNomUrl that exists on MO object to show a link to a MO.

The fk_parent_line must have a type that describe it is a foreign key to a MO object.

PR #20027, opened 2022-02-09
FIX: Edit card not permitted when invoice is validated and INVOICE_CAN_ALWAYS_BE_EDITED is set to 0

kkhelifa-opendsi:
FIX: Edit card not permitted when invoice is validated and INVOICE_CAN_ALWAYS_BE_EDITED is set to 0

Fix of wrong commit on PR#19675


htdocs/compta/facture/card.php
```diff
@@ -122,6 +122,7 @@
 $usercansend = (empty($conf->global->MAIN_USE_ADVANCED_PERMS) || $user->rights->facture->invoice_advance->send);
 $usercanreopen = (empty($conf->global->MAIN_USE_ADVANCED_PERMS) || $user->rights->facture->invoice_advance->reopen);
 $usercanunvalidate = ((empty($conf->global->MAIN_USE_ADVANCED_PERMS) && !empty($usercancreate)) || (!empty($conf->global->MAIN_USE_ADVANCED_PERMS) && !empty($user->rights->facture->invoice_advance->unvalidate)));
+$usercanmodify = $usercancreate && ($object->statut == Facture::STATUS_DRAFT || ($object->statut == Facture::STATUS_VALIDATED && !empty($conf->global->INVOICE_CAN_ALWAYS_BE_EDITED)));
```
eldy, line 125: Option INVOICE_CAN_ALWAYS_BE_EDITED means "Can edit the invoice even if a first payment was started".
If no payment was done yet, it seems we can already correct an error into an invoice (what is expected). Can you describe how you reproduce the bug ?

warnerbryce: For reproducing this bug, we have to set INVOICE_CAN_ALWAYS_BE_EDITED to 0 and having Advanced Perms enabled MAIN_USE_ADVANCED_PERMS to 1
You can set the perms for a user to avoid : RE-open an Invoice and RE-open an Invoice already paid
Then we have to validate an Invoice, we can no longer re-open it but we can change some settings : Payment Terms, Payment Due on, Payment Type and then generate a new PDF with this moded params... That shouldn't be allowed. 
That's why this PR address this. 
With this fix, you can no longer modify this params when the invoice is validated (except if you have the right to)

eldy: There is a bad interpretation of the meaning of "INVOICE_CAN_ALWAYS_BE_EDITED".
It does not mean can always be edited. I have renamed it into "INVOICE_CAN_BE_EDITED_EVEN_IF_PAYMENT_DONE" so now the goal of this option is clearer.

Currently, there is no reason to block modification of project, payment term, etc... since it allows to fix typing errors with no harmful effects.

PR #20028, opened 2022-02-09
Change of SQL request when fetching CommandeFournisseurLigne

noec764:
When a CommandeFounisseurDet had more than one entry in llx_product_supplier_price the sql query selected the supplier order twice. I changed this behavior to select the llx_product_vendor_price with the closest lower quantity.


frederic34:
image

PR #20034, opened 2022-02-10
New needs net tab on bom

atm-quentin:
# NEW|New needs net tab on bom

On the standard BOM form, a Net Requirements tab has been added to display all the products / services required for the BOM and its sub-BOMs, with no level limit. The products / services will be grouped by products / services, the table will display :
- The quantity needed for the BOM and its sub-BOMs
- The real stock
- The virtual stock

A button Display in tree structure will allow you to display the complete tree structure of the BOM and its sub-BOMs without level limit.

image
image


eldy:
Can you replace the button "Display in a tree structure" and "Back to" that is at position for "Actions" button with 2 pictos just over the table that is position for "presentation" (Picto Flat view /  Hierarchic view) like it is for the list of users. So it will match look and feel. See example:
image

atm-quentin:
Yes but we are in a card there, where should i display those pictos ?

eldy:
> Yes but we are in a card there, where should i display those pictos ?

Just above the table, like it is for the task in a project card:
image

atm-quentin:
I did it : 
image

Is it ok for you ?

atm-quentin:
(i can add a hidden conf if necessary)

atm-quentin:
@eldy

PR #20036, opened 2022-02-10
NEW Automatically add the CUSTOMER contact

dolibarr95:
# NEW Add email recipient from object contact's.

When user wants  to send an email from an object card (order...) it's select the contact "CUSTOMER" (llx_c_type_contact) from the contacts list of the object.

see https://www.dolibarr.fr/forum/t/envoi-email-commande-client/39153

### DIFF
--- a/htdocs/core/tpl/card_presend.tpl.php
+++ b/htdocs/core/tpl/card_presend.tpl.php
@@ -269,10 +269,10 @@ if ($action == 'presend') {
 
 	$contactarr = array();
 	$contactarr = $tmpobject->liste_contact(-1, 'external');
-	
-	if(!GETPOSTISSET("receiver")){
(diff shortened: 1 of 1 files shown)


dolibarr95:
Hello
What about adding a constant something like 
```
MAILING_EMAIL_AUTOSEL_RECIPIENT
```
and use this modification only if the constant is defined ?

Tks for reading

eldy:
> Hello What about adding a constant something like
> 
> ```
> MAILING_EMAIL_AUTOSEL_RECIPIENT
> ```
> 
> and use this modification only if the constant is defined ?
> 
> Tks for reading

Yes adding a hidden contant MAILING_EMAIL_AUTOSEL_RECIPIENT to enable this behaviour seems better.

eldy:

PR #20056, opened 2022-02-13
FIX Disable customer type by default if type prospect/customer is disabled

aspangaro:
# Instructions

1. Define customer type by default on prospect/customer

image

2. Disable customer type prospect/customer

image

Now, create new customer with menu entry : 

image

Solution : 
Force to value "Customer" (1) if type customer/prospect in closed.


aspangaro:
Ping @eldy Fix always available

PR #20058, opened 2022-02-14
NEW Asset module - New SQL structure - FPC21

aspangaro:


hregis:
Hello @aspangaro 
the good declaration of timestamp (tms)
`tms timestamp DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP`

aspangaro:
Thanks, done @hregis

hregis:
@aspangaro  again in sql update file ;-)

aspangaro:
Arggg ! Done @hregis 🥇 Thanks

eldy:
Can you make me a summary of goal of each new table (with a example of data). Adding this info and the example into the header comment of each sql file would help to keep the information for future. Can you add it ?

eldy:
@aspangaro Can you complete the description of the new tables (with a example of data) to help me understand the PR ?

htdocs/install/mysql/tables/llx_asset_accountancy_codes_economic-asset.sql
```diff
@@ -0,0 +1,39 @@
...
+-- Table to store the configuration of the accounting accounts of a fixed asset for economic status
+--
+-- Data example:
+-- INSERT INTO `llx_asset_accountancy_codes_economic` (`rowid`, `fk_asset`, `fk_asset_model`, `asset`, `depreciation_asset`, `depreciation_expense`, `value_asset_sold`, `receivable_on_assignment`, `proceeds_from_sales`, `vat_collected`, `vat_deductible`, `tms`, `fk_user_modif`) VALUES
+-- (2, 1, NULL, '2183', '2818', '68112', '675', '465', '775', '44571', '44562', '2022-01-18 14:20:20', 1);
+
+CREATE TABLE llx_asset_accountancy_codes_economic(
+    rowid						integer			AUTO_INCREMENT PRIMARY KEY NOT NULL,
+    fk_asset					integer,
+    fk_asset_model				integer,
```
eldy, line 27: My understanding is that each "asset" as a type into "asset_model".
So you have a link with fk_asset_model into table fk_asset.

The table llx_asset_accountancy_codes_economic has record linked to an asset.
So we already have an inherited link to the asset model through the asset. The model is a property of the asset.
So we should not have fk_asset_model here.
Should be the same for other tables.

aspangaro: No, not quite.

An asset can be created with or without a model.

If you create a template, you will be able to set up accounting accounts for the economic & fiscal status in this template.
When you create an asset, you choose this model and the accounting accounts of the model are included in this asset because not everyone knows all the accounting accounts to correctly account for an asset.
In the table llx_asset_accountancy_codes_economic, you will have a row for the model and another for the asset created.

If you create an asset without a template, you can set all the accounting accounts on or off.
In the table llx_asset_accountancy_codes_economic, you will have a row for the asset created.

In summary, models & assets both use the tables llx_asset_accountancy_codes_economic & llx_asset_accountancy_codes_fiscal to store their information. If it's a model, fk_asset_model is populated. If it is an asset, only fk_asset is filled.

image

I correct some code for model here: #20984

aspangaro:
Ping @eldy 

Please consider this PR, it's the SQL structure of the module, the code part that has been integrated can't work without its base

htdocs/install/mysql/tables/llx_asset_accountancy_codes_economic-asset.sql
eldy: so let's start with your summary
"If it's a model, fk_asset_model is populated. If it is an asset, only fk_asset is filled."
So it means the table is used to store accountancy code for 2 differents object. Why not if property are very similar.

So into file llx_asset_accountancy_code_economic.sql, you have this comment:

`-- Table to store the configuration of the accounting accounts of a fixed asset for economic status`

If my understanding is good, can we replace it with

```
-- Table to store the configuration of the accounting accounts of a fixed asset for economic status (fk_asset will be filled and fk_asset_model will be null)
-- or to store the configuration of the accounting accounts for templates of asset (fk_asset_model will be filled and fk_asset will be null)
```

Can you confirm ?

htdocs/install/mysql/data/llx_c_asset_disposal_type.sql
```diff
@@ -12,6 +12,14 @@
...
 -- along with this program. If not, see <https://www.gnu.org/licenses/>.
+--
+--
 
+--
+-- Do not include comments at end of line, this file is parsed during install and string '--' are removed.
+--
 
-ALTER TABLE llx_asset_type_extrafields ADD INDEX idx_asset_type_extrafields (fk_object);
+INSERT INTO `llx_c_asset_disposal_type` (`rowid`, `entity`, `code`, `label`, `active`) VALUES
```
eldy, line 22: Remove the ` that is not portable SQL from your sql request...

htdocs/install/mysql/data/llx_c_asset_disposal_type.sql
```diff
@@ -12,6 +12,14 @@
...
+(1, 1, 'C', 'Sale', 1),
```
eldy, line 23: To be compatible with the error management system if install and migration, you must not used the "mysql compact version of sql" but 3 full real sql requests. So split this line in 3 lines.

Also, we must not include the rowid into the insert (not compatible with multicompany for example). The system will use the automatic increment system. We did it in past for some dictionary objects and it is still a problem. So can you instead do:
INSERT INTO llx_c_asset_disposal_type (entity, code, label, active) VALUES (1, 'C', 'Sale', 1);
INSERT INTO llx_c_asset_disposal_type (entity, code, label, active) VALUES...;
INSERT INTO llx_c_asset_disposal_type (entity, code, label, active) VALUES...;

aspangaro:
I have updated the comment in the 'llx_asset_accountancy_codes_economic-asset.sql' file and updated the sql data file.

I confirm that in the solution we designed, we do have two objects using the same data within the 'llx_asset_accountancy_codes_economic' table and that we will never have the two elements (fk_asset & fk_asset_model) differentiating them from filled in at the same time.

Thank you for your consideration of this PR

PR #20064, opened 2022-02-14
Fix uniformize pdf logo behaviour for supplier orders

atm-greg:
# FIX uniformize pdf logo behaviour
Supplier order pdf models get large logo in every pdf generation.
It's a double problem if you do not use SUPPLIER_ORDER_NO_DIRECT_APPROVE => 2 generations on approval => 40 sec to validate and approve...

Applied behaviour is the same as client side.


atm-greg:
@eldy i think that it is an important optimisation.
If you can check it.

PR #20070, opened 2022-02-15
add hooks

vincentjdc:
# NEW|New Add some hooks
Add some hooks

### DIFF
--- a/htdocs/comm/index.php
+++ b/htdocs/comm/index.php
@@ -314,26 +314,33 @@ if (!empty($conf->supplier_proposal->enabled) && $user->rights->supplier_proposa
  * Draft customer orders
  */
 if (!empty($conf->commande->enabled) && $user->rights->commande->lire) {
+	$tables = [ 'c' => 'commande', 's' => 'societe' ];
 	$sql = "SELECT c.rowid, c.ref, c.ref_client, c.total_ht, c.total_tva, c.total_ttc, c.fk_statut as status";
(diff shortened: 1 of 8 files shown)


htdocs/comm/index.php
```diff
@@ -314,26 +314,33 @@
...
 	}
+	$parameters = array('type' => 'customer_draft_orders', 'tables' => $tables);
+	$hookmanager->executeHooks('printFieldListFrom', $parameters, $object);
+	$sql .= $hookmanager->resPrint;
 	$sql .= " WHERE c.entity IN (".getEntity($orderstatic->element).")";
 	$sql .= " AND c.fk_statut = ".Commande::STATUS_DRAFT;
-	$sql .= " AND c.fk_soc = s.rowid";
 	if (empty($user->rights->societe->client->voir) && !$socid) {
-		$sql .= " AND s.rowid = sc.fk_soc AND sc.fk_user = ".((int) $user->id);
+		$sql .= " AND sc.fk_user = ".((int) $user->id);
```
eldy, line 336: Because you change the join from a simple ","  into a inner join, you muwt also move the condition after the ON  on the INNER JOIN so it must be moved on line 327
$sql .= " INNER JOIN ".MAIN_DB_PREFIX."societe_commerciaux as sc ON s.rowid = sc.fk_soc AND sc.fk_user = ".((int) $user->id);

eldy:

PR #20083, opened 2022-02-16
FIX : HRM | Vacant position option

ghost:
# FIX | Vacant Position Option on HRM

Allows you to create vacant positions:

Create positions in prevention of future recruitments (without fill user) . Allows to have a quick view of job vacancy can be interesting

This option is materialized via a checkbox:
image
If it is checked, select2 of users is disabled:
image


ghost:
@eldy fixed

PR #20098, opened 2022-02-17
Create MO Child

atm-lena:
# NEW|New Create MO Child

From a BOM, add possibility to create MO Child.

Create MO and Children : 
image

MO Children display in Linked Object of Mo Parent : 
image

MO Parent displays in MO Child : 
image

MO Parent in MO List : 

image


eldy:
On first screenshot, we show a checkbox for product3. For this case, we should not have any checkbox, isn't it ?

atm-lena:
> On first screenshot, we show a checkbox for product3. For this case, we should not have any checkbox, isn't it ?

Indeed, I removed input for line with only product and no bom

eldy:
Can you rename title "Objet liés" in screenshot into "Child MO" ?
Also can you confirm that when Mo a is parent of Mo b, it means we must first make Mo b and AFTER Mo a ?

atm-lena:
> Can you rename title "Objet liés" in screenshot into "Child MO" ? Also can you confirm that when Mo a is parent of Mo b, it means we must first make Mo b and AFTER Mo a ?

For now, we can create MO parent and MO Child only from interface. We can't create a MO parent / child manually.

atm-lena:
@eldy ?

atm-maxime:
@eldy I reviewed this with @atm-lena and I can confirm that all child MO must be finished to be able to finish the parent MO. Although there's no restriction for the moment.

atm-thibaultf:
Indeed, it can be a restriction we could add in the future with a specific configuration. But if it's ok for you @eldy , is it possible to merge this feature in the present state ?

eldy:
> Can you rename title "Objet liés" in screenshot into "Child MO" ?
> Also can you confirm that when Mo a is parent of Mo b, it means we must first make Mo b and AFTER Mo a ?

I try to understand what is the direction of a "Child MO". 
When you are on MO "M" and you see in the section "Object linked" a MO "A" and "B". What does it means ?
That you must first make MO A and B and the resulted products will be used into MO M.
Or does it means you must first make MO M and then you can make the MO A and B using the product generated by MO M ?

atm-thibaultf:
> I try to understand what is the direction of a "Child MO".
> When you are on MO "M" and you see in the section "Object linked" a MO "A" and "B". What does it means ?
> That you must first make MO A and B and the resulted products will be used into MO M.
> Or does it means you must first make MO M and then you can make the MO A and B using the product generated by MO M ?

I confirm that we think about this behaviour : 
When you are on MO "M" and you see in the section "Object linked" a MO "A" and "B", it means that you must first make MO A and B and the resulted products will be used into MO M.

MO "M" is the parent MO of MO "A" and "B". 
A MO can have only one parent but several children (from cascading BOM).

PR #20109, opened 2022-02-18
FIX Accountancy - On manual transaction with multicompany - Function transformTransaction can validate & delete transaction of an other entity FPC21+

aspangaro:
Hard to reproduce.

Begin a manual transaction on general ledger on entity 2 without validate the transaction. Go to entity 1 et begin a manual transaction and validate it.

If the first transaction of entity 2 is not balanced, it's resolved with an incompressible error but if it's balanced on entity 2, you validate the transaction on entity 2 from entity 1 et delete operation of entity 1...

Function transformTransaction() needed to be restricted on the multi-entity


hregis:
@aspangaro @eldy

`$sql .= " WHERE aa.entity = " . ((int) $conf->entity); // Do not use getEntity for accounting features`

equal with : 
`
$sql .= " WHERE aa.entity = " . (int) $conf->entity; // Do not use getEntity for accounting features`

"extra" parentheses are superfluous and useless !

PR #20110, opened 2022-02-18
FIX: api for sqlfilter with operator NOT LIKE

fappels:


fappels:
~~Thanks for fixing @eldy, possible to have this in 14? Know somebody who's e-commerce app using rest api started failing after upgrade Dolibarr form 13 to 14.~~ 
Sorry, see also in 14, thanks. Strange github does not show this in commit.

PR #20120, opened 2022-02-18
Add ticket page in user personal landpage

RomainDeschamps:
# NEW [*Add ticket page in user personal landpage*]

image


hregis:
@RomainDeschamps @eldy
Maybe a generic function for the next release 16 for include modules with a parameter?

RomainDeschamps:
Would be great. But just can't do it :-)

PR #20121, opened 2022-02-19
FIX error with php8

ptibogxiv:


frederic34:
@ptibogxiv is_countable was added in php 7.3 and we still support php 5.6

ptibogxiv:
> @ptibogxiv is_countable was added in php 7.3 and we still support php 5.6

I will change my fix for support old php

PR #20145, opened 2022-02-22
FIX: function dolGetButtonAction change title if html not set

atm-GregM:
On dolGetButtonAction function, add new attribute for title if html is empty.

before
image

with fix
image


atm-john:
Ok for me

john-botella (approved):
OK for me

PR #20151, opened 2022-02-23
FIX Invoice - When you create an invoice and defined a thirdparty, fk_account is not retrieved from company card FPC21+

aspangaro:
If you are defined a bank account in company card, try to create an invoice. Bank account is not retrieved.

image

image

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -2898,6 +2898,9 @@ if ($action == 'create') {
 			if (empty($mode_reglement_id)) {
 				$mode_reglement_id = $soc->mode_reglement_id;
 			}
+			if (empty($fk_account)) {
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/card.php
```diff
@@ -3005,6 +3008,11 @@
 		$mode_reglement_id = GETPOST("mode_reglement_id", 'int');
 	}
 
+	// when bank account is empty (means not override by payment mode form a other object, like third-party), try to use default value
+	if (empty($fk_account)) {
```
eldy, line 3012: i think you should replace this test and line with
$fk_account = GETPOSTISSET("fk_account") ? GETPOST("fk_account", 'int') : $fk_account;
Because priority is always the value coming from the POST form and then other default values.

PR #20164, opened 2022-02-23
NEW dol_uncompress() supports extensions (.gz, .bz2, .zstd). Only .zip was supported before.

Hystepik:
The function dol_uncompress can now uncompress :
- tar.gz files
- .gz files
- .bz2 files
- .tar.bz2 files

### DIFF
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -2085,9 +2085,6 @@ function dol_uncompress($inputfile, $outputdir)
 {
 	global $conf, $langs, $db;
 
-	include_once DOL_DOCUMENT_ROOT."/core/class/utils.class.php";
-	$utils = new Utils($db);
-	
 	$fileinfo = pathinfo($inputfile);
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/files.lib.php
```diff
@@ -2083,65 +2083,97 @@ function dol_compress_file($inputfile, $outputfile, $mode = "gz", &$errorstring
...
-				$zip->extractTo($outputdir.'/', array($zip->getNameIndex($i)));
-			}
 
-			$zip->close();
-			return array();
+		return array('error'=>'ErrNoZipEngine');
+	} elseif ($fileinfo["extension"] == "gz" || $fileinfo["extension"] == "bz2") {
+		$extension = pathinfo($fileinfo["filename"], PATHINFO_EXTENSION);
+		if ($extension == "tar") {
+			$cmd = "tar -C ".$outputdir." -xvf ".$fileinfo["dirname"]."/".$fileinfo["basename"];
```
eldy, line 2152: When forging  a $cmd string, can you use escapeshellcmd() on parameters for security purpose

PR #20177, opened 2022-02-24
Load fk_bank when fetching loan and prefill payment card with predefi…

### LINKED ISSUE #20176: Loan payment card missing predefined values
### Bug

When entering a loan payment term, the card is not prefilled with payment mode, nor bank account although these informations are present in the database for the loan and schedule.

### Environment Version

14.0.5

### Environment OS

CentOS 8

### Environment Web server

nginx

### Environment PHP

7.4

### Environment Database

mariadb 10.3

### Environment URL(s)

/loan/payment/payment.php

The form should be prefilled withe known values

_No response_

### Attached files

_No response_

simicar29:
…ned values for payment type and bank account

FIX #20176

### DIFF
--- a/htdocs/loan/payment/payment.php
+++ b/htdocs/loan/payment/payment.php
@@ -293,7 +293,7 @@ if ($action == 'create') {
 		print '</tr>';
 
 		print '<tr><td class="fieldrequired">'.$langs->trans("PaymentMode").'</td><td colspan="2">';
-		$form->select_types_paiements(GETPOSTISSET("paymenttype") ? GETPOST("paymenttype", 'alphanohtml') : $line->fk_typepayment, "paymenttype");
+		$form->select_types_paiements(GETPOSTISSET("paymenttype") ? GETPOST("paymenttype", 'alphanohtml') : (!empty($line) ? $line->fk_typepayment : ''), "paymenttype");
 		print "</td>\n";
 		print '</tr>';
 
--- a/htdocs/loan/payment/payment.php
+++ b/htdocs/loan/payment/payment.php
@@ -293,14 +293,14 @@ if ($action == 'create') {
 		print '</tr>';
 
 		print '<tr><td class="fieldrequired">'.$langs->trans("PaymentMode").'</td><td colspan="2">';
-		$form->select_types_paiements(GETPOSTISSET("paymenttype") ? GETPOST("paymenttype", 'alphanohtml') : $loan->fk_typepayment, "paymenttype");
+		$form->select_types_paiements(GETPOSTISSET("paymenttype") ? GETPOST("paymenttype", 'alphanohtml') : $line->fk_typepayment, "paymenttype");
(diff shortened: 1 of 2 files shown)


htdocs/loan/payment/payment.php
```diff
@@ -293,14 +293,14 @@
 		print '</tr>';
 
 		print '<tr><td class="fieldrequired">'.$langs->trans("PaymentMode").'</td><td colspan="2">';
-		$form->select_types_paiements(GETPOSTISSET("paymenttype") ? GETPOST("paymenttype", 'alphanohtml') : $loan->fk_typepayment, "paymenttype");
+		$form->select_types_paiements(GETPOSTISSET("paymenttype") ? GETPOST("paymenttype", 'alphanohtml') : $line->fk_typepayment, "paymenttype");
```
eldy, line 296: Object $line seems not defined here (at least in some cases).

simicar29:
You are right, I added a test on $line (!empty($line)).

htdocs/loan/payment/payment.php
```diff
@@ -293,14 +293,14 @@
 		print '</tr>';
 
 		print '<tr><td class="fieldrequired">'.$langs->trans("PaymentMode").'</td><td colspan="2">';
-		$form->select_types_paiements(GETPOSTISSET("paymenttype") ? GETPOST("paymenttype", 'alphanohtml') : $loan->fk_typepayment, "paymenttype");
+		$form->select_types_paiements(GETPOSTISSET("paymenttype") ? GETPOST("paymenttype", 'alphanohtml') : (!empty($line) ? $line->fk_typepayment : ''), "paymenttype");
```
eldy, line 296: Why not just using ' '   ?
$line just does not exists or has no sens on this line. What is $line on a field that is at loan level ?
It seems we dont have fk_typepayment store on loan, so may be we should completely remove this field off the form creation ?

simicar29: When calling payment.php, il will seek the last unpaid scheduled term and fill $line with the related values (lines 54 to 84). So in practice, $line will be defined most of the time you want to enter a payment. Payment type could have been defined as the loan level, but it is not unreasonable to have it at the schedule level either. In any case it is supposed to be a mandatory information when registering payments.

eldy: So you must use a new var $lastpaymenttypeoflines initialized before the loop
and set it into the loop, so when you are outside the loop this var still has a meaning.
Then use it into select_types_paiements
On line 296 $line is potentially a var that should does not exists anymore since it was defined inside the loop (so var is local to the scope of the loop only). Having it still defined outside of a loop is a side effect but code must not rely on side effect.

simicar29: I am not sure I get your point here. Should'nt the empty test return true if it is not defined ? By the way, the same test is performed 4 other times in that script.

PR #20186, opened 2022-02-25
FIX : Add ID into formproject method

atm-kevin:
# FIX

Add an ID into formproject method to be able to target afterwards


htdocs/core/class/html.form.class.php
```diff
@@ -5199,13 +5199,15 @@ public function form_project($page, $socid, $selected = '', $htmlname = 'project
 			$out .= '<input type="submit" class="button smallpaddingimp" value="'.$langs->trans("Modify").'">';
 			$out .= '</form>';
 		} else {
+            $out .= '<span id="project_head_block">';
```
eldy, line 5202: We can have only 1 id per value into a html page but form_project can be called several times.
So:
use class="..."
or
use id= but include the htmlname into the id name.

PR #20193, opened 2022-02-25
NEW : accoutancy : subleger label input automatic filling when selecting subleger account in select2 list

gdsoftdev:
subleger label input automatic filling when selecting subleger account in select2 list


eldy:
PR has been canceled. It was not pushed on develop branch.
Thanks to resubmit it on develop.

PR #20196, opened 2022-02-25
FIX - Add last_main_doc in table societe	

BB2A-Anthony:
# FIX|Fix #[*Add last_main_doc in table societe*]
[*Long description*]


eldy:
Does it generate a bug into v15 ?
If yes, which one ?
If not, can you push into develop ?

BB2A-Anthony:
@eldy  it's done

PR #20198, opened 2022-02-25
NEW : accoutancy : subleger label input automatic filling when selecting subleger account in select2 list

gdsoftdev:
accoutancy : subleger label input automatic filling when selecting subleger account in select2 list


aspangaro:
Interesting !

Issue #20137, opened 2022-02-21
NEW - Head contact


BB2A-Anthony:
### Feature Request

I suggeste to create a link to a contact (head contact) like head quarter.

And a option to update automaticly a contact childs.

And other option delete head contact -> delete all childs contact

### Use case

I think this alleviates the problem visibly often encountered 1 contact belongs to one and only one third party.

### Suggested steps

Create link in database -> fk_head_contact
In card contact list of childs contact

eldy:
Do you mean having a contact that is parent of other ?

BB2A-Anthony:
Yes this it

eldy:
Why not. 

It may be used to manage a hierarchy of contact into a given thirdparty.
But trying to use 1 contact for different companies is a very very bad idea. When the same physical contact is in 2 different company it is not the same role, also often not the same contact phone, address, etc;;, so it must be also remain 2 different address record.
Also this will bring a lot of problem (permission problem, pb with external contact, pb for import, ...) . This feature was already implemented in past and we have to remove it. It is not a so requested feature and when it is, it appears that it is to represent the same firstname and lastname only and role is different so keeping 2 different record was better and we revert to feature 1 contact for 2 thirdparties. A "link" without constraint may be a better implementation but it must be a hidden feature for the moment for 2 reasons:
- Some similar problem we got in past will reappears. 
- This feature is clearly not a feature that is requested by a lot of user (and when it is, 99.99% of time, it appears at the first need to change the phone, desk, email, in company one and not in company 2 that it was a bad idea).

You must first send a PR with database structure changes before submitting php code.

BB2A-Anthony:
@eldy i understand i have create a PR for database structure

BB2A-Anthony:
up

eldy:
PR merged manually.

PR #20201, opened 2022-02-25
NEW - Manage a hierarchy of contacts into customer of supplier companies

BB2A-Anthony:
# NEW|New [*Link contact*]

Next time other PR for clone contact and affect link.
If this PR merged.

Link issue #20137

First PR #23147


atm-maxime:
I'm not sure I understand your goal here. How defining a parent contact on a contact will help to tell that a contact is related to several third-parties ?

BB2A-Anthony:
> I'm not sure I understand your goal here. How defining a parent contact on a contact will help to tell that a contact is related to several third-parties ?

@atm-maxime the goal is multiple 

- know that the contact exists in several third-party and potentially different coordinates to contact him.
- Thereafter to be able to propose an update of all the contacts when one of them is modified to have up-to-date data on all the linked contacts. (An option on a linked contact that should not auto-update as the information should remain different)

eldy:
By design, we don't want to have one contact to be linked to several companies in Dolibarr.
If this is the case, this is 2 different roles, and in such situation it means it is 2 different contacts (even if in real life, it is at time T the same person). We already tried to include such a feature in the past to have one contact on several thirdparties, but this bring more problems (permissions problems, conflict wit the external user feature, need to split contact because when in 2 different companies, the level of knwoledge on 2 contact must differs). Finally it was forgotten.

But is your feature related to make 1 contact on several companies or is it to build a hierarchy of contacts into 1 company ?

BB2A-Anthony:
@eldy For me i look the build a hierarchy of contacts into 1 company

bb2a:
> By design, we don't want to have one contact to be linked to several companies in Dolibarr.
> If this is the case, this is 2 different roles, and in such situation it means it is 2 different contacts (even if in real life, it is a time T the same person). We already trie to include such a feature in the feature to have on contact on several thirdparties, but this bring more problems (permissions problems, conflict wit the external user feature, need to split contact because when in 2 different companies, the level of knwoladge on 2 contact must differs). Finally it was forgotten.
> 
> But is your feature related to make 1 contact on several companies or is it to build a hierarchy of contacts into 1 company ?

@eldy For me i look the build a hierarchy of contacts into 1 company

eldy:
Field was called "parent" into llx_soc but it is an exception. We like to have fields that points to a table fk_.. and if it is the same table to have it name   fk_parent.
Can you name it like this ?
Also the default behaviour should be the current behaviour (only one level of contacts) because it will be more commonly used like this, so can you revert option DISABLE_PARENT_CONTACT into  ENABLE_PARENT_CONTACT ?

BB2A-Anthony:
@eldy it's done

eldy:
Do you mean having one contact among all the contact of thirdparties that are a more powerfull contact ?
What is goal of having such a more important contact than others ?

BB2A-Anthony:
> Do you mean having one contact among all the contact of thirdparties that are a more powerfull contact ? What is goal of having such a more important contact than others ?

@eldy 
It is a parent contact.

We cannot have a contact for several third parties for various issues, I understand.

The idea is to have a parent contact (referrer) which allows several things.

To know that it is the same person and that it interacts in several tiers.

The update of all the contacts attached with an option of course (not global, but on the contact) so that the other contacts remain up to date without being returned on each card. (Future PR)

you understand me

French:

C'est un contact parent.

Nous ne pouvons pas avoir un contact pour plusieurs tiers pour diverses problématiques, je conçois.

L'idée est d'avoir un contact parent (référent) ce qui permet plusieurs choses.

De savoir qu'il s'agit de la même personne et quelle interagis dans plusieurs tiers.

La mise à jour de tous les contacts rattaché avec une option bien sûr (pas global, mais sur le contact) pour que les autres contacts restent à jour sans retourné sur chaque fiche. (Futur PR)

Vous me comprenez?

htdocs/contact/card.php
```diff
@@ -1589,6 +1624,11 @@ function init_check_no_email(input) {
 
 		if ($action != 'presend') {
 			print '<div class="fichecenter"><div class="fichehalfleft">';
+			// Childs list
+			if (empty($conf->global->SOCIETE_DISABLE_CONTACT_CHILDS)) {
```
eldy, line 1628: Is it SOCIETE_DISABLE_CONTACT_CHILDS or SOCIETE_ENABLE_PARENT_CONTACT  ?

htdocs/societe/admin/societe.php
```diff
@@ -236,6 +237,46 @@
...
 
+// Disable Parent company
+if ($action == "setdisableParentCompagny") {
+	$val = GETPOST('value', 'int');
+	$res = dolibarr_set_const($db, "SOCIETE_DISABLE_PARENTCOMPANY", $val, 'yesno', 0, '', $conf->entity);
+
+	if (!($res > 0)) {
+		$error++;
+	} else {
+		$res = dolibarr_set_const($db, "SOCIETE_DISABLE_SUBSIDIARIES", $val, 'yesno', 0, '', $conf->entity);
```
eldy, line 248: We should modify a second option when changing value of a first one. I think you can remove this. A fix was done in v16 so SOCIETE_DISABLE_SUBSIDIARIES is used only if SOCIETE_DISABLE_PARENTCOMPANY is set so you should no more need to modify it when changing SOCIETE_DISABLE_PARENTCOMPANY

bb2a: @eldy Ok i delete

BB2A-Anthony:
It's done you are right it is clearer no need for two options

htdocs/contact/class/contact.class.php
```diff
@@ -113,6 +114,7 @@ class Contact extends CommonObject
 		'priv' =>array('type'=>'smallint(6)', 'label'=>'ContactVisibility', 'enabled'=>1, 'visible'=>1, 'notnull'=>1, 'position'=>175),
 		'fk_stcommcontact' =>array('type'=>'integer', 'label'=>'ProspectStatus', 'enabled'=>1, 'visible'=>-1, 'notnull'=>1, 'position'=>220),
 		'fk_prospectlevel' =>array('type'=>'varchar(12)', 'label'=>'ProspectLevel', 'enabled'=>1, 'visible'=>-1, 'position'=>255),
+		'fk_parent' =>array('type'=>'integer', 'label'=>'ParentContact', 'enabled'=>1, 'visible'=>-1, 'position'=>260),
```
eldy, line 117: I think we should have
'enabled'=>getDolGlobalInt('SOCIETE_ENABLE_PARENT_CONTACT')
so field will be visible only when option is on.

bb2a: OK it's done

eldy:
I think I understood but i think such a feature bring a lot of complexity in database relations for too few users that need it. I think such a feature should be provided by an external module instead providing its own link table.

PR #20212, opened 2022-02-27
Add 2 hooks formObjectOptions in payment

avadnc:
# New: Add 2 hooks formObjectOptions in payment
In México and other countries in LATAM when someone create payment, we need insert the exchange amount in foreing currency. Dolibarr only allows change the price of exchange in invoice card, but we need enter this data in the payment card.

### DIFF
--- a/htdocs/admin/multicurrency.php
+++ b/htdocs/admin/multicurrency.php
@@ -24,9 +24,9 @@
 
 // Dolibarr environment
 require '../main.inc.php';
-require_once DOL_DOCUMENT_ROOT.'/core/lib/admin.lib.php';
-require_once DOL_DOCUMENT_ROOT.'/core/lib/multicurrency.lib.php';
-require_once DOL_DOCUMENT_ROOT.'/multicurrency/class/multicurrency.class.php';
+require_once DOL_DOCUMENT_ROOT . '/core/lib/admin.lib.php';
+require_once DOL_DOCUMENT_ROOT . '/core/lib/multicurrency.lib.php';
+require_once DOL_DOCUMENT_ROOT . '/multicurrency/class/multicurrency.class.php';
 
 
 // Load translation files required by the page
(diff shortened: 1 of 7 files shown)


htdocs/admin/multicurrency.php
```diff
@@ -198,10 +198,24 @@
...
+// when your currency is less than foreign currency
+
+print '<tr class="oddeven">';
+print '<td>' . $langs->transnoentitiesnoconv('MULTICURRENCY_LESS_THAN_FOREIGN') . '</td>';
+print '<td class="center">';
+if ($conf->use_javascript_ajax) {
+	print ajax_constantonoff('MULTICURRENCY_LESS_THAN_FOREIGN');
+} else {
+	$arrval = array('0' => $langs->trans("No"), '1' => $langs->trans("Yes"));
+	print $form->selectarray("MULTICURRENCY_LESS_THAN_FOREIGN", $arrval, $conf->global->MULTICURRENCY_LESS_THAN_FOREIGN);
```
eldy, line 210: What is the goal of this new feature ?

eldy:
New features must be pushed in develop branch.
Also the PR contains 2 differents thing.  One to add hooks, one to add an option MULTICURRENCY_LESS_THAN_FOREIGN, you must make 1 PR per atomic change.

avadnc:
The objective is to be able to give the user the freedom to be able to add a textbox or a select, and thus be able to issue a payment receipt. For example, dolibarr currently takes the exchange rate of the invoice or the one that has been entered in the multicurrency module, but in some countries, companies receive payments where the exchange rate is different from the two mentioned, as is the case of Mexico.
In addition to that, for the development of an electronic invoicing module, it is very convenient to give the user the possibility of adding additional objects in the payment, so as not to have to duplicate data or generate redundant screens, when the information required by the treasury can be entered from the payment.

avadnc:
Respect that MULTICURRENCY_LESS_THAN_FOREIGN i will create another pull request. in this case the goal is allow users in latam insert the change currency like goverment publish in Official Bulletins like Mexico when publish relationship USD or EUR to MXN for example: 01/03/2022 USD 20.5272 (MXN)

eldy:
> Respect that MULTICURRENCY_LESS_THAN_FOREIGN i will create another pull request. in this case the goal is allow users in latam insert the change currency like goverment publish in Official Bulletins like Mexico when publish relationship USD or EUR to MXN for example: 01/03/2022 USD 20.5272 (MXN)

Do you know that in multicurrency module setup, there is a tab/page to enter/setup currency rates.

eldy:
Closed, pr was not pushed in the correct branch.

PR #20238, opened 2022-03-01
Add PDF option display date

BB2A-Anthony:
# NEW|New [*Add PDF option display date*]
[*Long description*]


htdocs/admin/pdf.php
```diff
@@ -486,8 +489,18 @@
 	print '</td></tr>';
 }
 
-//
+// Display date
+print '<tr class="oddeven"><td>'.$langs->trans("MAIN_PDF_DATE_TEXT");
+print '</td><td>';
+if ($conf->use_javascript_ajax) {
+	print ajax_constantonoff('MAIN_PDF_DATE_TEXT');
```
eldy, line 496: This feature is not yet completely implemented (a lot of PF templates does not implement it), so we can't enable it as a visible stable feature since it just does not work on a lot of cases.

bb2a: I understand. i have add comment in option title. like other options who is not available in all PDF

Issue #20234, opened 2022-03-01
Introduce CheckInvoiceDate feature as a core functionnality


thomas-Ngr:
### Feature Request

[CheckInvoiceDate](https://www.dolistore.com/en/modules/1334-Invoice-dates-check.html) module allows to check an invoice date when validating a draft invoice. This is useful for some companies. However, this module seems not to be maintained.

I suggest to add this functionality in the core, available as an option of Facture module.

### Use case

The user wants to be sure that validated invoices of same type follow one another, by ref number and by date.

### Suggested implementation

Considering new option  INVOICE_CHECK_POSTERIOR_DATE  is activated :

- when validating a draft invoice, we verify that the current invoice date is posterior to the last invoice **with the same invoice type**. If not, an error should be displayed and block the invoice validation.

- when validating an invoice **that is being modified** (name is not PROV* but FA*), the validation should pass even if the current invoice date is anterior to the last invoice found for same invoice type.

- by "posterior date", we mean "the **day** is posterior or equal". We won't check for **hour** posteriority.

- this check is made on single invoice validation (compta/facture/card.php) and mass validation (compta/facture/list.php)

- ~~when mass validating invoices, an error on this check on a single invoice should not prevent other invoices to be validated (should not block the whole validation lot). A warning should be displayed with the appropriate invoice reference (PROV*)~~ (this would need to change the whole mass validation process)

- when mass validating invoices, a sorting should be made by invoice date before validating invoices. This prevents conflicts due to wrong invoice order.

### Suggested steps

- introduce option INVOICE_CHECK_POSTERIOR_DATE in Facture module and Facture admin panel (part "Invoices and credit notes numbering model")

- create a method willBeLastOfSameType() for Facture object

- add the check on facture/card.php
- add the check on facture/list.php

atm-maxime:
Hi. Thanks for contributing.
Did you try to contact the module editor to ask him about his module ?

thomas-Ngr:
Hi !

No. I just sent them an email.

However, the PR https://github.com/Dolibarr/dolibarr/pull/20247 I made inbetween (draft for now) adds this check on mass validation, which seems not to be part of the original module functionnality. So  I would rather merge this PR.

Moreover, I did not use the original module code.

thomas-Ngr:
Since I did not have any answer after one week, and considering:
- the PR does not use any code from checkInvoicedate external module
- the PR adds functionnality (invoice date check on mass validation)

I think this proposition can be merged.

PR #20247, opened 2022-03-02
NEW: check invoice date

thomas-Ngr:
NEW check invoice date

Introduces an option to check invoices dates on validation.
When activated, validating an invoice will be forbidden if its date is anterior to the date of last invoice of same type.

This introduces  INVOICE_CHECK_POSTERIOR_DATE constant.

Issue https://github.com/Dolibarr/dolibarr/issues/20234

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -305,7 +305,7 @@ if (empty($reshook)) {
 		if (!empty($conf->global-> INVOICE_CHECK_POSTERIOR_DATE)) {
 			$last_of_type = $object->willBeLastOfSameType();
 			if (empty($object->date_validation) && !$last_of_type[0]) {
-				setEventMessages($langs->transnoentities("ErrorInvoiceIsNotLastOfSameType", $object->ref , dol_print_date($object->date, 'day'), dol_print_date($last_of_type[1], 'day')), null, 'errors');
+				setEventMessages($langs->transnoentities("ErrorInvoiceIsNotLastOfSameType", $object->ref, dol_print_date($object->date, 'day'), dol_print_date($last_of_type[1], 'day')), null, 'errors');
 				header('Location: '.$_SERVER["PHP_SELF"].'?facid='.$id);
 				exit;
 			}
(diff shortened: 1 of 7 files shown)


htdocs/compta/facture/class/facture.class.php
```diff
@@ -5123,6 +5130,37 @@ public function sendEmailsRemindersOnInvoiceDueDate($nbdays = 0, $paymentmode =
...
 	}
+
+	/**
+	 * See if current invoice date is posterior to the last invoice date among validated invoices of same type.
+	 * @return boolean
+	 */
+	public function willBeLastOfSameType()
+	{
+		// get date of last validated invoices of same type
+		$sql  = 'SELECT datef';
```
eldy, line 5141: For sql forging, please use ", the simple ' is reserved for string inside the sql (so string in sql format).
This will allow some of our security tool to works better against sql injection.

thomas-Ngr: Thanks @eldy , I corrected it. I think the PR can be merged.

mverove:
Hello,

Is it possible to merge this PR please ?

Thank you in advance.

htdocs/compta/facture/card.php
```diff
@@ -302,6 +302,15 @@
 		// Validation
 		$object->fetch($id);
 
+		if (!empty($conf->global-> INVOICE_CHECK_POSTERIOR_DATE)) {
+			$last_of_type = $object->willBeLastOfSameType();
+			if (empty($object->date_validation) && !$last_of_type[0]) {
+				setEventMessages($langs->transnoentities("ErrorInvoiceIsNotLastOfSameType", $object->ref, dol_print_date($object->date, 'day'), dol_print_date($last_of_type[1], 'day')), null, 'errors');
+				header('Location: '.$_SERVER["PHP_SELF"].'?facid='.$id);
```
eldy, line 309: Why making a redirect here ? and not just 
$action = '';
like done on other tests after ?

thomas-Ngr: I wanted to stop rendering the page to prevent the validation. But you are right, using `$action = ''` has the same effect. I just pushed a new commit with this modification.

Issue #19310, opened 2021-11-04
Attached files : option to display links into the "attached files" section on a card


atm-maxime:
# Feature Request
On any card, we have an "attached files" section on the bottom left, displaying the attached files linked to the card. It would be great to have an option to also display "links" associated to the card.

## Suggested implementation
Add a global option "Show attached links directly on cards"

atm-maxime:
Still wanted

PR #20248, opened 2022-03-02
New #19310 display links documents on card

atm-GregM:
# new #19310
add links documents on card view
add possibility to delete link on card view

image

### DIFF
--- a/htdocs/core/actions_builddoc.inc.php
+++ b/htdocs/core/actions_builddoc.inc.php
@@ -178,7 +178,6 @@ if ($action == 'confirm_remove_link') {
 	require_once DOL_DOCUMENT_ROOT.'/core/class/link.class.php';
 	$link = new Link($db);
 	$res = $link->fetch(GETPOST('linkid', 'int'));
-	var_dump($res, $link->error, $link->errors);
 	if ($res > 0) {
 		$link->delete($user);
 	} else {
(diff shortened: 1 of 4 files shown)


john-botella:
Can you add font awesome link icon before link label ?
```<i class="fa fa-link"></i>```

atm-GregM:
add font awesome external-link icon

image

atm-GregM:
Dissociated files and links
image

add None message
image

eldy:
This means we add all the content of page "Files and linked files" to the main home page.
Trying to move all information of all tabs on the first main page is a dangerous direction. It save one click when we need to see the linked file, but we introduce also more scrolling (a larger page even when we don't need this information making more cons than pro). It is also a very bad user experience to have an information managed in read at one place and managed in write (upload) at another place.
I would prefer this steps:
A) - First rewrite the page documents, so the action "Upload a file" or "Add a link" is just a simple picto (+) at end of line title of arrays of joined files and at end of line title of line of array of attached file (using js that make the upload component appears).
B) - After that, we will have only 2 tables to manage, 1 for the joined files and 1 for the linked files. 
C) - Once this will be a truth, we will be able to completely remove the tab "Linked files" into the main card. So we keep the rule read and edit at same place. We lost space on the main tab due to a new section but we saved some due to one tab that disappear (so balance is neutral instead of being negative). But i am not sure this step will remain possible in future because a lot of feature may appears to manage files, upload, etc... But at least it will be easier to move into the main tab if it is still a good choice to do this. 
So we must first focus on A).

atm-maxime:
@atm-GregM Following discussion with @eldy, can you go back on the non separated version and merge conflicts ?

eldy:
Closed, another implementation has been suggested

Issue #18434, opened 2021-08-18
Employee management improvements


atm-maxime:
# Feature Request
Change the way data of an employee are managed, add new data and permissions about these datas

## Use case
When we want to manage informations about our employees (personal informations), these informations are shared to everyone who can read a user card. We'd like to add new permissions to deal with personal datas and move all datas in the employee tab

## Suggested implementation
- [x] Add 2 new permissions in the user module : read employee and write employee
- [ ] Only display the employee tab if the user is flagged as an employee
- [ ] Move these datas in the employee tab : Supervisor, Force expense report validator, Force leave request validator, Job position, Average hourly rate, Average daily rate, Salary, Hours worked (per week), Employment date, Date of birth, Accounting account
- [x] Add new fields in employee tab : ref (i.e. employee number, used on the payslips), national registration number (i.e. social security number)

cfoellmann:
I am completely with you on this.
Do you need help with anything or do you have the time to get this ready for 15.x?

cfoellmann:
We already have the HR+bank tab.
This should be the employee data tab!?

atm-adrien:
https://github.com/Dolibarr/dolibarr/pull/20249

PR #20249, opened 2022-03-02
New employee management improvement #18434

atm-steve:
Add read and write rights for user as employee, add two new fields ref_employee and national register number in RH and Bank tab, move Accounting account field to RH and Bank tab.


htdocs/install/mysql/tables/llx_user.sql
```diff
@@ -108,5 +108,7 @@ create table llx_user
   import_key        varchar(14),				-- import key
   default_range     integer,
   default_c_exp_tax_cat     integer,
+  employee_number   varchar(50),
```
atm-maxime, line 111: Should be ref_employee

htdocs/user/bank.php
```diff
@@ -502,6 +502,25 @@
...
 	}
 
+	// Accountancy code
+	if (!empty($conf->accounting->enabled)) {
+		print '<tr><td>'.$langs->trans("AccountancyCode").'</td>';
+		print '<td>'.$object->accountancy_code.'</td></tr>';
+	}
+
+	// Employee Number
+	if (!empty($conf->accounting->enabled)) {
```
atm-maxime, line 512: No need to test for accounting enabled

hregis: Hello @atm-maxime 

to see more broadly, if we accept to save the accounting data while the module is not active,
must we save the data of the other modules while they are not activated?
Should we keep a history or make sure to have the means to recreate this history?

htdocs/core/modules/modHRM.class.php
```diff
@@ -249,6 +249,22 @@ public function __construct($db)
 		$this->rights[$r][4] = 'compare_advance';
 		$this->rights[$r][5] = 'read'; // In php code, permission will be checked by test if ($user->rights->hrm->compare_advance->read)
 		$r++;
+
+		// Read employee
+		$this->rights[$r][0] = 4031; // Permission id (must not be already used)
+		$this->rights[$r][1] = 'Read employee'; // Permission label
```
eldy, line 255: An employee is a user with object. 
And we already have a permission to read users and a permission for each type of object (leave, expense report, salary, etc...).
Adding such a permission create a lot of business rule conflicts into the code and on some security checks.
i think the permission you want to add is not to define if we can see an employee or not but to decide if we can see some kind of information (like personal information) ?

atm-steve: These two new rights is about the personal informations fields only.

eldy: So can you please rename the label of permission so things are clear. They were not for me and i have to go into the code to understand the permission was to view/read personal information. But end users are not able to do this analyze so having a better label of permission will avoid interrogation.

atm-maxime:
🥳

PR #20251, opened 2022-03-02
FIX : Add hooks on several project tabs

atm-kevin:
# FIX

Add initHooks method on several project tabs


htdocs/projet/document.php
```diff
@@ -32,6 +32,7 @@
 
 // Load translation files required by the page
 $langs->loadLangs(array('projects', 'other'));
+$hookmanager->initHooks(array('projetcarddocument'));
```
frederic34, line 35: proje**_c_**tcarddocument instead projetcarddocument ?

PR #20255, opened 2022-03-03
NEW option update prices on proposal cloning

lvessiller-opendsi:
NEW option update prices on proposal cloning

Add a check box in dialog box before cloning a proposal :
image

Rules on price :
- if only one price : get product price by default
- if serveral price levels : get price linked to thirdparty price level (const PRODUIT_MULTIPRICES)
- if one price for each customer : get price of selected company before cloning the prosposal (const PRODUIT_CUSTOMER_PRICES)

Rules on VAT : 
- get default VAT with the seller and the new buyer
- change it with product prices rules defined in product setup (PRODUIT_MULTIPRICES and PRODUIT_CUSTOMER_PRICES)

Rules on discount : 
- get by default the selected company discount


eldy:
Can you rename the label "Update prices" with "Update prices with current known prices"
Also what is default value of checkbox ?  To be compliant with "clone" verb, i suggest to have it not checked by default.

lvessiller-opendsi:
@eldy
I understand your suggest and I replace "Update prices" with "Update prices with current known prices".
I also add a const "PROPOSAL_CLONE_UPDATE_PRICES".
So by default the checkbox is unchecked and prices aren't updated with known prices.

This paramater can be set in propsal setup : 
image

eldy:
PR can't be merged, can you resubmit it after a pull ?

lvessiller-opendsi:
@eldy 
I resolved all conflicts after a pull.
Could you see to merge it ?

PR #20256, opened 2022-03-03
FIX: project creation prevented if PROJECTLEADER contact role renamed, de-activated or deleted

marc-dll:


htdocs/projet/card.php
```diff
@@ -174,10 +174,11 @@
...
 			{
 				// Add myself as project leader
-				$typeofcontact = 'PROJECTLEADER';	// TODO If use rename this code in dictionary, the add_contact will generate an error.
+				$typeofcontact = 'PROJECTLEADER';
 				$result = $object->add_contact($user->id, $typeofcontact, 'internal');
-				if ($result < 0)
-				{
+
+				// -3 means type not found (PROJECTLEADER renamed, de-activated or deleted), so don't prevent creation if it has been the case
+				if ($result < 0 && $result != -3) {
```
eldy, line 181: If we don't prevent creation, it means the project will be created without any contact to it. So the user that create the project is not a contact and he can't see/edit the project he has just created.
A specific message to ask to restore the key (that must be seen as a reserved key for the default contact type of the contact of a project) would be better.

Issue #20161, opened 2022-02-23
Tools / Import / Contacts - field skype does not exist anymore


rycks:
### Bug

Hello all,
there is a bug in import tool on v15 due to the fact that skype (and others) fields does not exists anymore in socpeople table ...

### Environment Version

15.0.0

### Environment OS

GNU/Linux

### Environment Web server

Apache

### Environment PHP

7.4

### Environment Database

MariaDB

### Environment URL(s)

/imports/import.php

### Steps to reproduce the behavior

Download XLS template, add one line, try to import that file -> error

### Attached files

![image](https://user-images.githubusercontent.com/1468823/155306189-bda42557-c7d1-4015-b45b-97563e2b33c4.png)

PR #20264, opened 2022-03-04
Fix: #20161 Can't import thirds. Column Skype doesn't exist

fmarcet:
In version 15 the 'skype' column has been removed but is trying to import it


hregis:
@fmarcet "skype" field or other social networks fields is deprecated, the new field is : 

` socialnetworks           text DEFAULT NULL,                           -- json with socialnetworks`

use with "llx_c_socialnetworks" dictionnary

fmarcet:
Hi @hregis, yes I know. But it was still in the list of fields to import in the third parties, so it was impossible to import third parties because the skype column does not exist.

hregis:
@fmarcet this field is no longer created on new installations

aspangaro:
#20161

PR #20275, opened 2022-03-04
Proposal : allow use of FQDN in REST API IPS allow list #20254

### LINKED ISSUE #20254: Allow to filter REST API access on dynamic dns FQDN
### Feature Request

May allow dynamic DNS domain for REST API access.

### Use case

Some dev is working on REST API on a remote Dolibarr test server.

For now, the Dolibarr admin may allow access by adding the dev IP in api/admin/index.php .

As most devs dont have fixed IPS, This must be done at each IP change by some admin  or admin access must be given to devs to allow him/her to change its IP.

Allow the use of some dynamic FQDN (machine.ddns.net for example) may save time for everybody. 

### Suggested implementation

_No response_

### Suggested steps

_No response_

eldk:
# Fix #20254
Add use of FQDN in REST API allowed client list


htdocs/api/index.php
```diff
@@ -172,6 +172,22 @@
...
+	// allow the use of FQDN for dynamic dns access
+	$count = count($allowedip);
+	for ($i = 0; $i < $count; $i++) {
+		// if submitted host is not an IP but some FQDN
+		if (!preg_match('/^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$/', $allowedip[$i])
+			&& preg_match('/^[a-zA-Z0-9-]*\.[a-zA-Z0-9-]*\.[a-zA-Z0-9-]*\.$/', $allowedip[$i])) {
+			// set the IP of the host
+			$allowedip[$i] = gethostbyname($allowedip[$i]);
+			//  if submitted host is not an IP removes it
+		} elseif (!preg_match('/^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$/', $allowedip[$i])) {
```
eldy, line 185: i think we should remove this elseif.
Ip entered may also be an ipv6.
And cleaning the array $allowedip[$i] seems useless because value is compared later to ipv4 and ipv6 so if value is wrong, it will just not match any test later.

eldk: It's OK for me. I'll do it in next days.
Thanks

eldy: @eldk Time to make the suggested minor fix so i can validate the PR ?

eldy:

PR #20290, opened 2022-03-07
Fix #17317 - Add default label for online subscription

### LINKED ISSUE #17317: Bug: Subscription label is wrong when member pay online on renew and new form
# Bug: Subscription label is wrong when member pay online on renew and new form 

## Environment
- **Version**: 13.0.2
- **URL(s)**: htdocs/public/payment/newpayment.php

Label is set on facture and in llx_subscription to something like "Online subscription AAAA-MM-DD HH:MM:SS using stripe from XXX.XX.XXX.XX (IP member address) - Transaction ID = pi_xxxxxxxxxxxxxxxxxxxxx"
It should be set to as the default label for subscription : trans("Subscription") AAAA

rowid_subscription2

daraelmin:
#Fix #17317 Subscription label is wrong when member pay online on renew and new form
Add default label for subscription before payment info if specified in module setup

### DIFF
--- a/htdocs/public/payment/paymentok.php
+++ b/htdocs/public/payment/paymentok.php
@@ -447,7 +447,7 @@ if ($ispaymentok) {
 				$amount = $FinalPaymentAmt;
 				// add default label for subscription before paymnetinfo if specified in module setup
 				$label = (empty($conf->global->MEMBER_NO_DEFAULT_LABEL)) ? $langs->trans("Subscription").' '.dol_print_date(($datefrom ? $datefrom :  dol_now()), "%Y") . ' - ': '';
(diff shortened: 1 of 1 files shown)


htdocs/public/payment/paymentok.php
```diff
@@ -445,6 +445,8 @@
 
 				$paymentdate = $now;
 				$amount = $FinalPaymentAmt;
+				// add default label for subscription before paymnetinfo if specified in module setup
+				$label = (empty($conf->global->MEMBER_NO_DEFAULT_LABEL)) ? $langs->trans("Subscription").' '.dol_print_date(($datefrom ? $datefrom :  dol_now()), "%Y") . ' - ': '';
```
eldy, line 449: The $label is replaced on line just after. So this line seems useless.

PR #20293, opened 2022-03-07
NEW Availibility dictionnary new column unit and number

atm-quentin:
image


htdocs/install/mysql/migration/15.0.0-16.0.0.sql
```diff
@@ -276,3 +276,13 @@ UPDATE llx_c_ticket_type SET use_default=1 WHERE code='OTHER' AND NOT EXISTS(SEL
 
 UPDATE llx_c_country SET eec=0 WHERE eec IS NULL;
 ALTER TABLE llx_c_country MODIFY COLUMN eec tinyint DEFAULT 0 NOT NULL;
+
+ALTER TABLE llx_c_availability ADD COLUMN type_duration char(1);
+ALTER TABLE llx_c_availability ADD COLUMN number real DEFAULT 0;
+
+delete from llx_c_availability WHERE rowid IN (1,2,3,4,5);
```
eldy, line 283: Do not include instruction that may destroy customization of user into a migration. Instead,
just keep the insert (there is a protection of duplicate due to the unique key on code) and after, for each insert, make an update to update the new field when code = the code.
Do not rely on rowid that may be different on each installation.

atm-quentin: I made some modifications, are you ok with that ?

atm-maxime:
I think "number" is an incorrect name for a SQL column name. Did you check if it's a reserved word ?

atm-quentin:
i renamed it into qty

htdocs/install/mysql/migration/15.0.0-16.0.0.sql
```diff
@@ -279,4 +279,11 @@ ALTER TABLE llx_propal ADD last_main_doc VARCHAR(255) NULL AFTER model_pdf;
 
 UPDATE llx_c_country SET eec=0 WHERE eec IS NULL;
 ALTER TABLE llx_c_country MODIFY COLUMN eec tinyint DEFAULT 0 NOT NULL;
-
+ALTER TABLE llx_c_availability ADD COLUMN type_duration char(1);
```
eldy, line 282: Do not use 'char', use always varchar, even for 1 char.

htdocs/install/mysql/tables/llx_c_availability.sql
```diff
@@ -23,6 +23,8 @@ create table llx_c_availability
 	rowid		integer	 	AUTO_INCREMENT PRIMARY KEY,
 	code		varchar(30) NOT NULL,
 	label		varchar(128) NOT NULL,
+	type_duration	char(1),
```
eldy, line 26: Do not use 'char', use always varchar, even for 1 char.

htdocs/install/mysql/migration/15.0.0-16.0.0.sql
atm-quentin: i did it

htdocs/install/mysql/tables/llx_c_availability.sql
atm-quentin: i did it

PR #20298, opened 2022-03-08
Add paging to inventory card

atm-lena:
# NEW|New
Add paging to inventory card

image

+ I deleted "disablebuttonmakemovementandclose()" function : the button worked even in "disabel" mode. Now, button is always visible but a message was added to the pop-in "No stock movement will be made on products that do not have an actual quantity entered.".


eldy:
Removing the disablebuttonmakemovementandclose may create a big confusion :
It was here to helping and force user to understand that he must first enter a quantity and click on "Save at bottom" BEFORE being able to record the movement. Each time the user modify a quantity without saving it first, the button become grey again, so the workflow is forced to firstly save the page.
Without that, the user can modify data on page, do not save its change and click on Generate movement and because it was not saved before, the movements are not the one requested. This happen for example when:
A- clicking on button Replace quantities with expected quantities
B- clear all quantities
C- entering or modifiyng manually a quantity.
This 3 events change value on the form but are not saved. 
You added a protection to save value when we navigate, good, but what if we change on page using method A, B or C and we immediatly click on "Generate movements" without changing page first to save data ? And pagination can occurs also when changing the limit of element, not only when clicking on previous/next. If you click on Generate movement without clicking on Save first, the recorded movement are wrong (they does not match the value modified).

I suppose your removed this protection because it was not compatible with the navigation ? (PS: the protection was not working correctly but I have fixed it for 15.0.1).
Nevermind, if you want to remove the disablebuttonmakemovementandclose, we must first enhance the page so the 3 events (A-B-C) that modifiy the quantities are saved in real time in table so when we launch the "generate movement" it uses the correct value.

So i suggest to first make a PR that solve this by updating the table, juste after doing event A, B and C, then it will be safe to remove the protection disablebuttonmakemovementandclose() and it will be possible to introduce pagination.

atm-thibaultf:
Hi @eldy,

For user experience, I think it should be better to add the save functionnality on the button to make stock movements and close.
We keep the "Save" button at the bottom for user who want to save their work without finishing the process of the inventory.

So for me, we should have :
- The "Save" button which save the modification of the page. When we switch to another page, the save is automatic.
- The "Generate stock movement and close" button, always active, which save first and make stock movement and close.

What do you think @eldy ?

eldy:
> Hi @eldy,
> 
> For user experience, I think it should be better to add the save functionality on the button to make stock movements and close. We keep the "Save" button at the bottom for user who want to save their work without finishing the process of the inventory.
> 
> So for me, we should have :
> 
> * The "Save" button which save the modification of the page. When we switch to another page, the save is automatic.
> * The "Generate stock movement and close" button, always active, which save first and make stock movement and close.
> 
> What do you think @eldy ?

If the button "Generate stock movement and close" do a Save first, it is ok for me to have the button always active.
But there is an other case you must manage with the pagination: it is when we change the number of element on page (from 10 to 20 for example). In this case, you must also introduce an auto save like with next / previous.

atm-lena:
Done @eldy

eldy:
What happen if you enter qty in some lines and the change the combo box with the max nb of record from 10 to 5 for example ?
Do you also save data ?

atm-lena:
> What happen if you enter qty in some lines and the change the combo box with the max nb of record from 10 to 5 for example ? Do you also save data ?

Yes data is saved

atm-lena:
> > What happen if you enter qty in some lines and the change the combo box with the max nb of record from 10 to 5 for example ? Do you also save data ?
> 
> Yes data is saved

@eldy

PR #20299, opened 2022-03-08
FIX check mandatory thirdparty fields for mass action

lvessiller-opendsi:
FIX check mandatory thirdparty fields for mass action
- when you validate several invoices from mass action, mandatory fields rules in company setup were not checked


lvessiller-opendsi:
@eldy
Could you see to merge this fix ?

lvessiller-opendsi:
@eldy 
I corrected it with all of yours comments

PR #20302, opened 2022-03-08
FIX: better error management at product selling price update

marc-dll:


htdocs/product/price.php
```diff
@@ -163,11 +163,15 @@
 			setEventMessages($object->error, $object->errors, 'errors');
 		}
 
-		if ($error)
-		{
+		if (!$error) {
```
marc-dll, line 166: Question mark there : I reversed the condition as it made more sense as far as I could understand, but it would be best you checked if it's OK

PR #20307, opened 2022-03-08
NEW - Record count badge in custom modules tab

BB2A-Anthony:
# NEW|New [*Record count badge in custom modules tab*]

New params in label construct tab in description module
Label,object,class,fonction

Function return

$nbrec

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -8750,7 +8750,7 @@ function getLanguageCodeFromCountryCode($countrycode)
  */
 function complete_head_from_modules($conf, $langs, $object, &$head, &$h, $type, $mode = 'add')
 {
-	global $hookmanager;
+	global $hookmanager, $db;
 
 	if (isset($conf->modules_parts['tabs'][$type]) && is_array($conf->modules_parts['tabs'][$type])) {
 		foreach ($conf->modules_parts['tabs'][$type] as $value) {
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -8771,7 +8771,19 @@ function complete_head_from_modules($conf, $langs, $object, &$head, &$h, $type,
 							complete_substitutions_array($substitutionarray, $langs, $object, array('needforkey'=>$values[2]));
 							$label = make_substitutions($reg[1], $substitutionarray);
 						} else {
-							$label = $langs->trans($values[2]);
+							$labeltemp = explode(',', $values[2]);
+							if (is_array($labeltemp)) {
```
eldy, line 8775: labeltemp is result of explode so it will always be an array ?  (even if there is one element)

htdocs/core/lib/functions.lib.php
```diff
@@ -8771,7 +8771,19 @@ function complete_head_from_modules($conf, $langs, $object, &$head, &$h, $type,
...
+								$label = $langs->trans($labeltemp[0]);
+								if (!empty($labeltemp[1]) && is_object($object) && !empty($object->id)) {
+									dol_include_once($labeltemp[2]);
+									$obj = new $labeltemp[1]($db);
+									$function = $labeltemp[3];
+									$nbrec = $obj->$function($object->id, $obj);
```
eldy, line 8781: If $funtion is not set or does not exists, it will make a fatal error

frederic34:
you can add a badge in the hook
https://github.com/Dolibarr/dolibarr/blob/f75049d0e85af5dd960ac3179c9eb3baf32b6336/htdocs/modulebuilder/template/class/actions_mymodule.class.php#L306-L355

htdocs/core/lib/functions.lib.php
bb2a: I resolv problem

htdocs/core/lib/functions.lib.php
bb2a: Yes i enderstand 'if (is_array($labeltemp)) {' is always true i change

bb2a: Done

PR #20308, opened 2022-03-08
Fix do not modify entity on group edit

frederic34:
https://www.dolibarr.fr/forum/t/etoile-globalgroup-dans-les-groupes-dutilisateurs-quelle-signification-quel-usage/38772/26

### DIFF
--- a/htdocs/user/group/card.php
+++ b/htdocs/user/group/card.php
@@ -217,8 +217,8 @@ if (empty($reshook)) {
 
 			if (!empty($conf->multicompany->enabled) && !empty($conf->global->MULTICOMPANY_TRANSVERSE_MODE)) {
 				$object->entity = 0;
-			} else {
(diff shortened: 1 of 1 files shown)


htdocs/user/group/card.php
```diff
@@ -218,7 +218,7 @@
 			if (!empty($conf->multicompany->enabled) && !empty($conf->global->MULTICOMPANY_TRANSVERSE_MODE)) {
 				$object->entity = 0;
 			} else {
-				$object->entity = GETPOST("entity");
+				$object->entity = GETPOSTISSET("entity") ? GETPOST("entity") : $conf->entity;
```
eldy, line 221: If we don't want to modify it, we should have
$object->entity = GETPOSTISSET("entity") ? GETPOST("entity") : $object->entity;
instead of
$object->entity = GETPOSTISSET("entity") ? GETPOST("entity") : $conf->entity;
??

PR #20309, opened 2022-03-09
Restoring missing VATIsNotUsed label

ymollard:
Restores new label introduced by https://github.com/Dolibarr/dolibarr/commit/7be9a98ef03e770f912f5d00392ed675740333ed 
Overwritten by a transiflex sync because of its absence in the en_US source.


eldy:
Already merged into v15 so will be available into dev automatically soon

PR #20314, opened 2022-03-09
fetchLines() in bom class : fetch child bom

atm-lena:
# FIX|Fix #
We need to fetch bom child of the bom line if it exists when we use "fetchlines()" for a BOM.

### DIFF
--- a/htdocs/bom/class/bom.class.php
+++ b/htdocs/bom/class/bom.class.php
@@ -398,11 +398,10 @@ class BOM extends CommonObject
 		$result = $this->fetchLinesCommon();
 
 		//associate to each line childbom if exist
-		foreach($this->lines as $line) {
+		foreach ($this->lines as $line) {
 			$bom_child = new BOM($this->db);
 			$res = $bom_child->fetch($line->fk_bom_child);
 			if ($res > 0) $line->childBom[] = $bom_child;
-
 		}
 
 		return $result;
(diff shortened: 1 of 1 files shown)


htdocs/bom/class/bom.class.php
```diff
@@ -396,6 +396,16 @@ public function fetchLines()
 		$this->lines = array();
 
 		$result = $this->fetchLinesCommon();
+
+		//associate to each line childbom if exist
+		if (!empty($this->lines)) {
+			foreach ($this->lines as $line) {
+				$bom_child = new BOM($this->db);
+				$res = $bom_child->fetch($line->fk_bom_child);
```
eldy, line 404: fetch and the init of new BOM must be done only if  $line->fk_bom_child is set

atm-lena: @eldy Done

htdocs/bom/class/bom.class.php
```diff
@@ -396,6 +396,18 @@ public function fetchLines()
 		$this->lines = array();
 
 		$result = $this->fetchLinesCommon();
+
+		//associate to each line childbom if exist
+		if (!empty($this->lines)) {
+			foreach ($this->lines as $line) {
+				if($line->fk_bom_child > 0) {
+					$bom_child = new BOM($this->db);
+					$res = $bom_child->fetch($line->fk_bom_child);
```
eldy, line 405: There is a risk of infinite loop if in database a bom A has child B that has child A...
Do you really need have a subrequest for child to load "immediatly" the child.
Can't you load it only when "you need it" ?

eldy:
Closed. Opened for too long. Missing answer.

PR #20318, opened 2022-03-10
Vietnam date format, VAT rates

Achicken7301:
# UPDATE|Update [*date format in Vietnam*]

In Vietnam, dates follow the "day month year" order. All-numeric dates can be written as:[[1]](https://en.wikipedia.org/wiki/Date_and_time_notation_in_Vietnam#cite_note-1)

d/m/yyyy (9/1/2021)
d-m-yyyy (9-1-2021)
dd/mm/yyyy (09/01/2021)
dd-mm-yyyy (09-01-2021)


eldy:
vat rates have been merged.
However for language date format, change will be lost if you don't fix this on the translation platform so on https://www.transifex.com/dolibarr-association (i have done it)

PR #20335, opened 2022-03-12
Update upgrade2.php

dolibit-ut:
add/translate some english comments

### DIFF
--- a/htdocs/install/upgrade2.php
+++ b/htdocs/install/upgrade2.php
@@ -15,7 +15,7 @@
  */
 
 
-/* License 
+/* License
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


htdocs/install/upgrade2.php
```diff
@@ -1,10 +1,21 @@
 <?php
+
+/*
```
eldy, line 3: i don't want to start this king of change and prefer to keep the header of file using the same structure everywhere so 
copyright, licence, comment. In 1 block.
Then doc for doxygen. This must remains in 1 block after. 
Some tools may rely on this and this does not bring a not enough enhancement to justify a hugh work (if we do for one file, we must do for all others and this is a very high work with to gain).

dolibit-ut: okay agree

PR #20336, opened 2022-03-12
Update llx_adherent.sql

dolibit-ut:


daraelmin (commented):
You should add thé excluded status in comment:
-2 excluded / exclu

dolibit-ut:
okay

PR #20354, opened 2022-03-14
FIX: Stock move on shipment close and reception close disabled for lot/serial type

fappels:
List of stock movements - multicompany 3-14-2022 6-08-32 PM


BebZ:
Hello Francis ! 
Does it change behaviour when we use stock movements on reception and shipments only ? 
Because it may add a double move when we close shipment and movements have already been done when validating shipment, no ?

fappels:
Hello Hubert,
This patch only allows to select the 'move on close' when lot/serial enabled. You still can't choose both.
Before you could only select move on validate and the move on close was disabled when lot/serial option enabled.
Best is to move on validate, but code to move lot/serial is available in setClosed method and maybe some use case could need this way of flow.

PR #20355, opened 2022-03-14
NEW - Option autocopy for user send mail

BB2A-Anthony:
# NEW|New [*Option autocopy for user send mail*]
[*Long description*]


BB2A-Anthony:
@eldy look my other PR please

eldy:
This feature seem to be the same than the one achieved with option MAIN_MAIL_AUTOCOPY_TO, except that any user can ask to be in a bulk copy without asking the administrator (only admin can set the option MAIN_MAIL_AUTOCOPY_TO).
I am not sure this is a good idea as BCC is hidden and must be reserved to a corporate policy and not to end user choice.

BB2A-Anthony:
> This feature seem to be the same than the one achieved with option MAIN_MAIL_AUTOCOPY_TO, except that any user can ask to be in a bulk copy without asking the administrator (only admin can set the option MAIN_MAIL_AUTOCOPY_TO). I am not sure this is a good idea as BCC is hidden and must be reserved to a corporate policy and not to end user choice.

Thank @eldy

Why it's not a good idea for a user to autocopy their own emails

if I understand MAIN_MAIL_AUTOCOPY_TO automatically copy all messages, not just those sent by the user

The goal is to have a trace in the user's mailbox

eldy:
Did you tried this
Set the constant
MAIN_MAIL_AUTOCOPY_TO   to value   "`__USER_EMAIL__`"    (with develop)    
It should have the same effect (each email sent goes into the user mailbox) but it is centralized by admin.

BB2A-Anthony:
> Did you tried this Set the constant MAIN_MAIL_AUTOCOPY_TO to value "`__USER_EMAIL__`" (with develop) It should have the same effect (each email sent goes into the user mailbox) but it is centralized by admin.
Yes is goog. It would be nice to be able to configure it per user

PR #20356, opened 2022-03-15
NEW Accountancy - Add subledger account when we generate a transaction with a deposit invoice FPC21+

aspangaro:


htdocs/accountancy/journal/sellsjournal.php
```diff
@@ -412,8 +412,13 @@
 					$bookkeeping->fk_docdet = 0; // Useless, can be several lines that are source of this record to add
 					$bookkeeping->thirdparty_code = $companystatic->code_client;
 
-					$bookkeeping->subledger_account = '';
-					$bookkeeping->subledger_label = '';
+					if ($k == $conf->global->ACCOUNTING_ACCOUNT_CUSTOMER_DEPOSIT) {
```
eldy, line 415: You must check also that $conf->global->ACCOUNTING_ACCOUNT_CUSTOMER_DEPOSIT is defined for the if for php8 compatibility

aspangaro: Thanks @eldy Resolve

PR #20359, opened 2022-03-15
fix: prededefault setup values for public field (visibility) works

FHenry:
Into admin/defaultvalues.php
image

Test on GETPOSTISSET do not check predefault values, use of GETPOST instead make it works


eldy:
We are in create mode, so $object->public does not exists, so we can just have GETPOST('public').
i pushed a different fix

FHenry:
thank @eldy to fix it in 15.0.... the bug is declare in 14.0. I'll do a PR to fix in 14.0 in your way

PR #20365, opened 2022-03-15
Fix : keep row order into invoices created from massactions

gdsoftdev:
# Fix : keep row order into invoices created from massactions
when creating invoices from receptions or client orders, row order is not kept if you choose to create one bill by third.
lines are mixed between commands or receptions

### DIFF
--- a/htdocs/core/actions_massactions.inc.php
+++ b/htdocs/core/actions_massactions.inc.php
@@ -635,6 +635,7 @@ if ($massaction == 'confirm_createbills') {   // Create bills from orders.
 
 	$TFact = array();
 	$TFactThird = array();
+	$TFactThirdNbLines = array();
 
 	$nb_bills_created = 0;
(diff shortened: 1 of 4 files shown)


eldy:
When we create 1 invoice for several orders with choice 1 invoice per thirdparty, we want to have into invoice
All lines of order 1, then all lines of order 2.
Do you confirm that goal of your patch is to achieve this or is it to have another order (which one in this case) ?

gdsoftdev:
> When we create 1 invoice for several orders with choice 1 invoice per thirdparty, we want to have into invoice All lines of order 1, then all lines of order 2. Do you confirm that goal of your patch is to achieve this or is it to have another order (which one in this case) ?

Yes that's the goal.
all lines of order 1 with lines ordered like in order 1, then all lines of order 2 with lines ordered like in order 2

PR #20372, opened 2022-03-16
FIX : in result by custom accounting account groups report, if we add…

atm-gauthier:
… 1 to param "onlysimplestring", a formula like GRP1+GRP2 never works because "+" char is matched by the regex in the test "if ($onlysimplestring == '1')"

Before my correction :
image

After my correction :
image

Maybe the right correction should be to correct the regex ? i'm not sure but in all cases there is a correction to do

### DIFF
--- a/htdocs/compta/resultat/result.php
+++ b/htdocs/compta/resultat/result.php
@@ -334,7 +334,7 @@ if ($modecompta == 'CREANCES-DETTES') {
 
 				//var_dump($result);
 				//$r = $AccCat->calculate($result);
-				$r = dol_eval($result, 1, 1, '1');
+				$r = dol_eval($result, 1, 1, 0);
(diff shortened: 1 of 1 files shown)


atm-gauthier:
@eldy I don't understand travis feedback :( 
image

eldy:
Travis has become crazy. Not related to your commit.
I suggest to fix differently by adding the "+" char into the allowed chars for the mode simple. As i pushed. Can you check it fixes you trouble.

atm-gauthier:
@eldy i confirm it works thank you, be we need to allow other operators (division, multiplication)
Here's my PR for that : https://github.com/Dolibarr/dolibarr/pull/20380
Can you check if it's ok ? I’m not very comfortable with regex

PR #20382, opened 2022-03-16
NEW: Cash control flexible period in TakePOS

andreubisquerra:
Currently in TakePOS the closures are natural days. But this is a limitation. It must be possible to carry out several closures in one day. Or even if a restaurant closes at 1am these sales must be left in the interval

### DIFF
--- a/htdocs/compta/cashcontrol/report.php
+++ b/htdocs/compta/cashcontrol/report.php
@@ -120,7 +120,7 @@ $sql.=" OR b.fk_account = ".((int) $conf->global->CASHDESK_ID_BANKACCOUNT_CB);
 $sql.=" OR b.fk_account = ".((int) $conf->global->CASHDESK_ID_BANKACCOUNT_CHEQUE);
 $sql.=")";
 */
-$sql = "SELECT f.rowid as facid, f.ref, f.datef as do, pf.amount as amount, b.fk_account as bankid, cp.code";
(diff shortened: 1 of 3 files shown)


htdocs/compta/cashcontrol/report.php
```diff
@@ -135,6 +135,8 @@
 	dol_print_error('Value for key = '.$key.' not supported');
 	exit;
 }*/
+if ($object->date_valid==null) $object->date_valid=dol_now();
+$sql .= " AND f.date_closing BETWEEN '".$db->idate($object->date_creation)."' AND '".$db->idate($object->date_valid)."'";
```
eldy, line 139: Don't understand this condition.
Process is creation -> validation -> closing.
What is goal of testing if invoice was closed before being validated ?

eldy, line 139: Don't you think we should use a  date of payment instead of f.date_closing ?
If on an invoice, a payment is done the morning, next one the second day, invoice should appears twice and first payment amount on 1 day and second payment amount on second day.

htdocs/compta/cashcontrol/report.php
```diff
@@ -135,6 +135,8 @@
...
 if ($syear && !$smonth) {
 	$sql .= " AND datef BETWEEN '".$db->idate(dol_get_first_day($syear, 1))."' AND '".$db->idate(dol_get_last_day($syear, 12))."'";
```
eldy, line 141: Same question, may be we should include the date of payment here instead of the date of invoice ? What do you think ?

PR #20387, opened 2022-03-17
FIX : multicurrency variable naming issue

ghost:
# FIX

When you try to discount from a deposit using multicurrency, the deducted amount is not taken into account.

diff


eldy:
I pushed a different fix because in v14 we renamed all amount into total_ttc, total_tva, total_ht.
Can you check it works ?

ghost:
It works ! Thank you !

PR #20388, opened 2022-03-17
Fix missing lang trad

john-botella:


aspangaro:
Hi @atm-john !

No file for other language than en_US. All other languages are managed inTransifex. You can remove translation in fr_FR/other.lang

frederic34:
it will be overwrited by next transifex sync

PR #20390, opened 2022-03-17
NEW : Add nb_open_day on llx_holiday for store complex compute

defrance:
# NEW  Add nb_open_day on llx_holiday for store complex compute
the main goal is to store the number of holiday open day who is not store on llx_holiday table
this reduce the number of time who made this compute (on list)
and allow to made some reporting and stat on this with sql query

### DIFF
--- a/htdocs/holiday/card.php
+++ b/htdocs/holiday/card.php
@@ -263,6 +263,7 @@ if (empty($reshook)) {
 				$object->date_debut = $date_debut;
 				$object->date_fin = $date_fin;
 				$object->halfday = $halfday;
+				$object->nb_open_day = $nbopenedday;
 
 				$result = $object->create($user);
 				if ($result <= 0) {
@@ -384,6 +385,7 @@ if (empty($reshook)) {
 				$object->date_fin = $date_fin;
 				$object->fk_validator = $approverid;
 				$object->halfday = $halfday;
+				$object->nb_open_day = $nbopenedday;
 
 				// Update
 				$verif = $object->update($user);
--- a/htdocs/holiday/card.php
+++ b/htdocs/holiday/card.php
@@ -573,7 +573,7 @@ if (empty($reshook)) {
 			// If no SQL error, we redirect to the request form
 			if (!$error) {
 				// Calculcate number of days consummed
-				$nbopenedday = num_open_day($object->date_debut_gmt, $object->date_fin_gmt, 0, 1, $object->halfday);
+				$nbopenedday = $object->nb_open_day;
 				$soldeActuel = $object->getCpforUser($object->fk_user, $object->fk_type);
 				$newSolde = ($soldeActuel - $nbopenedday);
 				$label = $langs->transnoentitiesnoconv("Holidays").' - '.$object->ref;
@@ -792,7 +792,7 @@ if (empty($reshook)) {
 				}
 
 				// Calculcate number of days consummed
-				$nbopenedday = num_open_day($object->date_debut_gmt, $object->date_fin_gmt, 0, 1, $object->halfday);
+				$nbopenedday = $object->nb_open_day;
 
 				$soldeActuel = $object->getCpforUser($object->fk_user, $object->fk_type);
 				$newSolde = ($soldeActuel + $nbopenedday);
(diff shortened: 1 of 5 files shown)


htdocs/install/mysql/migration/15.0.0-16.0.0.sql
```diff
@@ -30,6 +30,10 @@
 -- -- VPGSQL8.2 SELECT dol_util_rebuild_sequences();
 
 
+
+ALTER TABLE llx_holiday ADD COLUMN nb_open_day double(24,8)     DEFAULT 0,
```
frederic34, line 34: @defrance replace comma by semicolon at the end of line

frederic34:
what do you think of that
```
		if (empty($this->fk_type) || !is_numeric($this->fk_type) || $this->fk_type < 0) {
			$this->error = "ErrorBadParameterFkType"; return -1;
		}

		$nbopenday = num_open_day(
			$this->date_debut,
			$this->date_fin,
			0,
			1,
			$this->halfday
		);
		// num_open_day returns string in case of error
		if (is_string($nbopenday)) {
			$this->errors[] = $nbopenday;
			return -1;
		}
		// Insert request
		$sql = "INSERT INTO ".MAIN_DB_PREFIX."holiday(";
		$sql .= "ref,";

```

defrance:
> what do you think of that
> 
> ```
> 		if (empty($this->fk_type) || !is_numeric($this->fk_type) || $this->fk_type < 0) {
> 			$this->error = "ErrorBadParameterFkType"; return -1;
> 		}
> 
> 		$nbopenday = num_open_day(
> 			$this->date_debut,
> 			$this->date_fin,
> 			0,
> 			1,
> 			$this->halfday
> 		);
> 		// num_open_day returns string in case of error
> 		if (is_string($nbopenday)) {
> 			$this->errors[] = $nbopenday;
> 			return -1;
> 		}
> 		// Insert request
> 		$sql = "INSERT INTO ".MAIN_DB_PREFIX."holiday(";
> 		$sql .= "ref,";
> ```

think the error control is a better way but nor sur to bypass the use of
                  $this->db->jdate($this->date_debut, 1)
function

frederic34:
this is not working
image

and for old data, there are no consumed days, you have to do something to migrate old data

defrance (commented):
seem jdate work wrong on class and function nb_open_day already call on card.php page, so we use it and tranfert the good value on class

htdocs/holiday/card.php
```diff
@@ -792,7 +794,7 @@
 				}
 
 				// Calculcate number of days consummed
-				$nbopenedday = num_open_day($object->date_debut_gmt, $object->date_fin_gmt, 0, 1, $object->halfday);
+				$nbopenedday = $object->nb_open_day;
```
eldy, line 797: If you remove the call here to num_open_day(), we should retreive the call somewhere else. But i can't find it anywhere in your new code...

htdocs/install/mysql/migration/15.0.0-16.0.0.sql
```diff
@@ -30,6 +30,10 @@
 -- -- VPGSQL8.2 SELECT dol_util_rebuild_sequences();
 
 
+
+ALTER TABLE llx_holiday ADD COLUMN nb_open_day double(24,8)     DEFAULT 0;
```
eldy, line 34: According to announcement done during last devcamp, change into the structure of tables must be pushed into a different PR, before the PR that change the PHP code.

defrance (commented):
msyql pr ok then, what the next?

eldy:
> msyql pr ok then, what the next?

next is to send a PR to exploit this field. 
Default mode (old behaviour) is to not use this property. It is dynamically calculated.
I suppose you want to have this field being modified manually so you can push a PR with a hidden constant that add this field in edit/input mode. A next step will be to have it autofill by default with the calculated value but keeping the input field being editable. So this need a new PR so i close this one.

defrance:
You suppose wrong, this field is not editable, it's calculed during create or update that all (have you read the initial explain?), not on each read and display this value...
another important way is to have the good value without use php function, just by a query
my developpent is good, i don't understand why you have one more time made a tyranic close #hangry

Issue #20391, opened 2022-03-18
Ticket : option TICKET_AUTO_CREATE_FICHINTER_CREATE does not have any effect from public interface


thomas-Ngr:
### Bug

The `ticket` module has a hidden option TICKET_AUTO_CREATE_FICHINTER_CREATE that creates a PROV ficheinter when a ticket is created.

However, this option is only used in ticket/card.php. As a consequence, it is not active when a ticket is created from the public interface.

## Current behavior: 
when the option TICKET_AUTO_CREATE_FICHINTER_CREATE is set:
- when creating a ticket from the backend, a PROV ficheinter is created
- when creating a ticket from public interface, no ficheinter is created and we have to do it manually.

## Expected behavior:
when the option TICKET_AUTO_CREATE_FICHINTER_CREATE is set:
- when creating a ticket from the backend, a PROV ficheinter is created
- when creating a ticket from public interface, a PROV ficheinter is created

## Proposition
1. Since this option creates a workflow between Ticket and Ficheinter module, it should be part of the Workflow module.
2. Since we want it to be active anytime a ticket is created, it should be executed by triggers in TICKET_CREATE event.
3. Since it uses the contract linked to the ticket, it should be placed after the automatic link ticket->contract (option TICKET_AUTO_ASSIGN_CONTRACT_CREATE)

PR #20225, opened 2022-02-28
NEW: add a workflow to auto link contract on a ticket

thomas-Ngr:
Fix: Ticket: automatic link to contract.
This is a rewrite for PR #19992 to focus on the main functionality.

This enhances the former TICKET_AUTO_ASSIGN_CONTRACT_CREATE option, which:
- would link contract if the thirdparty had exactly one contract
- was in the Ticket module and not in the Workflow module

These are the changes made:

- rename constant TICKET_AUTO_ASSIGN_CONTRACT_CREATE to WORKFLOW_TICKET_LINK_CONTRACT
- treatments are made in Workflow module
- create options in Workflow admin panel (/admin/workflow.php)
- create constant TICKET_PRODUCT_CATEGORY and adapt ticket admin panel (/admin/ticket.php)
- add option WORKFLOW_TICKET_USE_PARENT_COMPANY_CONTRACTS to search among contracts of parent companies to link them to the ticket if none is found in initial thirdparty.

PR #20392, opened 2022-03-18
Fix: ticket/workflow: automatically create ficheinter

thomas-Ngr:
On ticket module, the option TICKET_AUTO_CREATE_FICHINTER_CREATE does not have any effect on tickets created from ~~backend~~ public interface. See issue #20391

This PR:
- fixes this issue by moving the ficheinter creation from ticket/card.php to triggers
- moves the option from the Ticket module to the Workflow module
- should be merged after https://github.com/Dolibarr/dolibarr/pull/20225 that deals with ticket->contract automatic link

### DIFF
--- a/htdocs/admin/workflow.php
+++ b/htdocs/admin/workflow.php
@@ -72,10 +72,10 @@ $workflowcodes = array(
 		'picto'=>'bill'
 	),
 	'WORKFLOW_TICKET_CREATE_INTERVENTION' => array (
-        'family'=>'create',
-        'position'=>25,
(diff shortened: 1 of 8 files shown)


htdocs/admin/ticket.php
```diff
@@ -219,7 +228,7 @@
 
 $dirmodels = array_merge(array('/'), (array) $conf->modules_parts['models']);
 
-$form = new Form($db);
+$form = new FormCategory($db);
```
eldy, line 231: $form var is commonly used for   $form = new Form($db);
So can you rename the var into   $formcategory  ?

PR #20421, opened 2022-03-21
FIX: replacing to dolGetButtonAction on Delivery card

atm-GregM:
# FIX| replacing to dolGetButtonAction on Delivery card


eldy:
Note that & alone is commonly used everywhere else and works great.
The important is not have both.

PR #20429, opened 2022-03-22
Fix : filemanagement in ecm and medias

Hystepik:
- Fix bug in directory add where "add in" field was empty
- Fix bug where backtopage in filemanager.tpl.php didn't had section and section_dir
- Fix php 8.0 warnings in html.formfile.class.php and filemanager.tpl.php


htdocs/core/tpl/filemanager.tpl.php
```diff
@@ -51,7 +51,7 @@
 	$permtoupload = ($user->rights->mailing->creer || $user->rights->website->write);
 	$showroot = 1;
 }
-
+$section = 0;
```
eldy, line 54: if ! isset

Issue #20418, opened 2022-03-21
Strings not being replaced with translation


gillerby:
### Bug

e.g. in attachment you can see it says "setupdescription1" where the proper string in english should go.

My user's language is english.

### Environment Version

14.0.5 (continues in 15.0.1)

### Environment PHP

7.4

### Environment Database

mysql

### Attached files

![image](https://user-images.githubusercontent.com/36923120/159267939-962bf0e6-52cb-4163-9d3e-622b9c0b385b.png)
![image](https://user-images.githubusercontent.com/36923120/159267972-2f9a2aa9-4b76-45ff-9931-fd5de5732160.png)

atm-lena:
I can’t reproduce the problem

PR #20441, opened 2022-03-23
Issue #20418 : load errors translation

atm-lena:
# FIX|Fix #Issue #20418 : load errors translation

Index Page

### DIFF
--- a/htdocs/index.php
+++ b/htdocs/index.php
@@ -37,6 +37,8 @@ $action = GETPOST('action', 'aZ09');
 
 $hookmanager->initHooks(array('index'));
 
+$langs->load('errors');
+
 
 /*
  * Actions


htdocs/index.php
```diff
@@ -37,6 +37,8 @@
 
 $hookmanager->initHooks(array('index'));
 
+$langs->load('errors');
```
eldy, line 40: Loading of error file should be done only on error case.
the lang file error should not be loaded in common usage. What is the key missing ?

atm-lena: Key translation "WarningLockFileDoesNotExists" is not loading. I found this key in errors.lang file

eldy: So add the $langs->load('errors');
just before the line that use the key translation (so file will be loaded only when required instead of at each call of page).

atm-lena: After looking back on the subject, I notice that  $langs->load('errors') juste before the line that use the key translation is already here. 
I no longer reproduce the problem. I close PR and note my comments on issue.

PR #20456, opened 2022-03-24
FIX multicompany sharings compatibility

hregis:


htdocs/core/class/commonobject.class.php
```diff
@@ -4399,23 +4400,32 @@ public function isObjectUsed($id = 0)
...
 				if ($obj->nb > 0) {
 					$langs->load("errors");
 					//print 'Found into table '.$table.', type '.$langs->transnoentitiesnoconv($elementname).', haschild='.$haschild;
 					$haschild += $obj->nb;
-					if (is_numeric($elementname)) {	// old usage
-						$this->errors[] = $langs->transnoentities("ErrorRecordHasAtLeastOneChildOfType", method_exists($this, 'getNomUrl') ? $this->getNomUrl() : $this->ref, $table);
-					} else // new usage: $elementname=Translation key
-					{
-						$this->errors[] = $langs->transnoentities("ErrorRecordHasAtLeastOneChildOfType", method_exists($this, 'getNomUrl') ? $this->getNomUrl() : $this->ref, $langs->transnoentitiesnoconv($elementname));
+					if (is_numeric($element) || empty($element['name'])) {	// old usage
```
eldy, line 4425: This create a compatibility problem. In most current cases, $element will be a string.
When $element is not a numeric (vey old usage) and is a string (old usage after your PR), $table is not the name of the object.
You should keep 3 cases.
if (is_numeric    for very old usage

elseif (!empty(     for old usage, with a message that will $element as the string value instead of $table in the error message

else your new usage

hregis: @eldy ok i see, and now with thix fix ?

htdocs/product/class/product.class.php
```diff
@@ -64,13 +64,13 @@ class Product extends CommonObject
...
 	 */
 	protected $childtables = array(
-		'supplier_proposaldet',
-		'propaldet',
-		'commandedet',
-		'facturedet',
-		'contratdet',
-		'facture_fourn_det',
-		'commande_fournisseurdet'
+		'supplier_proposaldet' => array('parent' => 'supplier_proposal', 'parentkey' => 'fk_supplier_proposal'),
```
eldy, line 67: May be we should also now set a property 'name' here ?  This is the new usage you introduce and the error message is using $element['name']  into its content.

hregis: @eldy it's fixed

htdocs/core/class/commonobject.class.php
```diff
@@ -4374,9 +4374,10 @@ public function getSpecialCode($lineid)
 	 *  Check is done into this->childtables. There is no check into llx_element_element.
 	 *
 	 *  @param	int		$id			Force id of object
+	 *  @param	int		$entity		Force entity to check
 	 *  @return	int					<0 if KO, 0 if not used, >0 if already used
 	 */
-	public function isObjectUsed($id = 0)
+	public function isObjectUsed($id = 0, $entity = 0)
```
atm-greg, line 4380: this method is used in object::delete methods
Restrict the search in an entity allows the user to delete an object in use in entity 3 when verifing entity 2 (in which it is not used)

hregis: @atm-greg @eldy 
"entity" parameter  is used only with Multicompany module !
if this poses a problem, it is only in Multicompany that you will have to see to correct it.

PR #20031, opened 2022-02-09
NEW barcode rule to insert product in takepos

lvessiller-opendsi:
NEW barcode rule to insert product in takepos

- add a barcode rule in takepos setup (barcode plugin must be enabled)

image

- then when you search a product with a specific barcode in a terminal

image

it will split serch term with barcode rule into : 
- product reference
- quantity (units)
- quantity (decimals)

and after you got a new invoice line with qty defined from barcode rule

image

PR #20458, opened 2022-03-24
FIX search bracode rule and avoid undefined quantity on search in tak…

lvessiller-opendsi:
FIX search bracode rule and avoid undefined quantity on search in takepos

This is a fix for PR #20031 :
- search from barcode rule only if barcode is different of term posted in form
- qty is set to 1 by default in result of this search

### DIFF
--- a/htdocs/takepos/ajax/ajax.php
+++ b/htdocs/takepos/ajax/ajax.php
@@ -142,7 +142,7 @@ if ($action == 'getProducts') {
 			}
 
 			if (isset($barcode_value_list['ref'])) {
-				//search product from reference
+				// search product from reference
 				$sql  = "SELECT rowid, ref, label, tosell, tobuy, barcode, price";
 				$sql .= " FROM " . $db->prefix() . "product as p";
 				$sql .= " WHERE entity IN (" . getEntity('product') . ")";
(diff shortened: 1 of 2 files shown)


htdocs/takepos/ajax/ajax.php
```diff
@@ -151,6 +151,7 @@
 					$sql .= " AND EXISTS (SELECT cp.fk_product FROM " . $db->prefix() . "categorie_product as cp WHERE cp.fk_product = p.rowid AND cp.fk_categorie IN (".$db->sanitize($filteroncategids)."))";
 				}
 				$sql .= " AND tosell = 1";
+				$sql .= " AND (barcode IS NULL OR barcode != '" . $db->escape($term) . "')";
```
eldy, line 154: If the product PR1 has a barcode BBB, and with search on "PR", the result will return an empty string when we want to get the product ?
This was the goal of the line
$sql .= " AND ref = '" . $db->escape($barcode_value_list['ref']) . "'";
If you add this new line, the search will always fails when searching on a barcode_rules_list that requests a search on a ref.
When doing a 
if (isset($barcode_value_list['ref'])) {
my understanding is that we want a search on ref whatever is barcode.

lvessiller-opendsi: @eldy 
No, this part of code it's only used when you have a barcorde rule to insert product with a quantity (such as product whose barcode depends on weight)
So you can't set a barcode in this kind of product or have only one barcode for only one unit of product.
And the search is only done when the number of characters in length of term matches with the number of characters in barcode rule "($barcode_length == $barcode_char_nb)"

I kept the old method of search (see at bottom of this code).

PR #20459, opened 2022-03-24
NEW Accountancy - New widget to show unbalanced entries

aspangaro:
image


htdocs/core/boxes/box_accountancy_unbalanced_entries.php
```diff
@@ -0,0 +1,179 @@
...
+		global $user, $langs, $conf;
+
+		include_once DOL_DOCUMENT_ROOT.'/accountancy/class/bookkeeping.class.php';
+
+		$bookkeepingstatic = new BookKeeping($this->db);
+
+		$this->info_box_head = array('text' => $langs->trans("BoxTitleUnbalancedEntries", $max));
+
+		if ($user->rights->accounting->mouvements->lire) {
+			$sql = "SELECT DISTINCT b.piece_num";
```
eldy, line 83: Seems not valid sql.
You have a group by that does not contains all fields in the select. Also there is no count or min or max function.
I think you want the SUM(b.debit) and the SUM(b.credit) into the select too ?

aspangaro: @eldy Thanks for your return.

Yes, we seems to have a problem with databases who have a "only full group by" option activated but when I try to list all field in group by, my request doesn't work.

What I want is to have a list of unbalanced accounting entries. Only one line with the piece_num as long as the sum of the debits and credits of the accounting entry are not equal.

Normally, it is not possible to have unbalanced entries in Dolibarr but on imports or advanced manipulations, we can meet unbalances so it will allow to have a widget that will detect the errors to help to solve them.

eldy: So this is what you must do to get only 1 line:
remove the distint on the b.piece_num,
remove the line 
$sql .= ", b.doc_date as date_movement";
$sql .= ", b.label_operation";
$sql .= ", b.montant as amount";
$sql .= ", b.code_journal";
and add instead add
$sql .= " SUM(b.debit), SUM(b.credit)"
You should get 1 line for each piece_num with the debit and credit on same line, and only when debit != credit.

PR #20468, opened 2022-03-25
FIX - Case of invoice situation at 100% with a credit note

kkhelifa-opendsi:
Set situation_final to 0 if is a credit note and the invoice source is a invoice situation (case when invoice situation is at 100%) when the credit note is validated
 So we can continue to create new invoice situation


htdocs/compta/facture/class/facture.class.php
```diff
@@ -2848,6 +2848,25 @@ public function validate($user, $force_number = '', $idwarehouse = 0, $notrigger
 				}
 			}
 
+			/*
+			 * Set situation_final to 0 if is a credit note and the invoice source is a invoice situation (case when invoice situation is at 100%)
+			 * So we can continue to create new invoice situation
+			 */
+			if (!$error && $this->type == self::TYPE_CREDIT_NOTE && $this->fk_facture_source > 0) {
```
eldy, line 2855: Any credit note has a fk_facture_source defined. But this does not means it is a situation invoice and does not means its parent source is a situation invoice.
We have no reason to force a setFinal, if parent invoice is not a situation invoice. May be you can introduce a test ?
Also, we can also check that situation_invoice is not already 0 so we make the change only if value is not correct (so not 0).

kkhelifa-opendsi: It's modified

lvessiller-opendsi:
@eldy 
Could you see to merge it ?

PR #20469, opened 2022-03-25
FIX : with situation invoice, total_ht is total_ht - total_ht prev situation line

atm-gauthier:


htdocs/accountancy/customer/lines.php
```diff
@@ -190,7 +190,13 @@
  * Customer Invoice lines
  */
 $sql = "SELECT f.rowid as facid, f.ref as ref, f.type, f.datef, f.ref_client,";
-$sql .= " fd.rowid, fd.description, fd.product_type as line_type, fd.total_ht, fd.total_tva, fd.tva_tx, fd.vat_src_code, fd.total_ttc,";
+$sql .= " fd.rowid, fd.description, fd.product_type as line_type, fd.total_tva, fd.tva_tx, fd.vat_src_code, fd.total_ttc,";
+$sql .= " CASE WHEN l_prev_situation_line.rowid IS NOT NULL
```
eldy, line 194: This is more complex than that.
On situation invoices, the total also depends on type. Some part of code are storing the total, other are storing the delta with previous one. it depends also if it is a credit note or standard invoice (bad choice done into the structure)
For this reason, the situation invoice as it is encoded will not be released as it need a complete rewriting. Same reason why I won't validate PR that makes the core code more complex for situation invoices (this feature is not an official feature and won't be as long as a rewriting using a more clever way to store its data has been implemented).

atm-gauthier: @eldy i understand, i close this PR

PR #20478, opened 2022-03-29
fix #20477: add online_payment_url on orders API requests

### LINKED ISSUE #20477: API: add online_payment_url for orders and invoices
### Feature Request

I can't find any API entry to get online_payment_url for an order or invoice ?

### Use case

Get details of an order via API and make a link to online payment

### Suggested implementation

Here is a "solution" for orders, if it's a good idea i could send a patch / MR, let me know.

htdocs/commande/class/api_orders.class.php near line 234

```PHP
            while ($i < $min) {
                $obj = $this->db->fetch_object($result);
                $commande_static = new Commande($this->db);
                if ($commande_static->fetch($obj->rowid)) {
                    // Add external contacts ids
                    $commande_static->contacts_ids = $commande_static->liste_contact(-1, 'external', 1);

                    //erics : add payment link
                    require_once DOL_DOCUMENT_ROOT.'/core/lib/payments.lib.php';
                    $commande_static->online_payment_url = getOnlinePaymentUrl(0, 'order', $commande_static->ref);

                    $obj_ret[] = $this->_cleanObjectDatas($commande_static);
                }
                $i++;
            }
```

### Suggested steps

_No response_

rycks:
# Fix #20477

That patch add a new online_payment_url attribute on orders API requests to be able to make a link to online payment if needed

### DIFF
--- a/htdocs/commande/class/api_orders.class.php
+++ b/htdocs/commande/class/api_orders.class.php
@@ -135,6 +135,11 @@ class Orders extends DolibarrApi
 		// Add external contacts ids
 		$this->commande->contacts_ids = $this->commande->liste_contact(-1, 'external', $contact_list);
 		$this->commande->fetchObjectLinked();
+
+		// Add online_payment_url, cf #20477
+		require_once DOL_DOCUMENT_ROOT.'/core/lib/payments.lib.php';
+		$this->commande->online_payment_url = getOnlinePaymentUrl(0, 'order', $this->commande->ref);
+
 		return $this->_cleanObjectDatas($this->commande);
 	}
 
--- a/htdocs/commande/class/api_orders.class.php
+++ b/htdocs/commande/class/api_orders.class.php
@@ -230,6 +230,10 @@ class Orders extends DolibarrApi
 				if ($commande_static->fetch($obj->rowid)) {
 					// Add external contacts ids
 					$commande_static->contacts_ids = $commande_static->liste_contact(-1, 'external', 1);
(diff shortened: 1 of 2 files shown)


htdocs/commande/class/api_orders.class.php
```diff
@@ -230,6 +235,10 @@ public function index($sortfield = "t.rowid", $sortorder = 'ASC', $limit = 100,
 				if ($commande_static->fetch($obj->rowid)) {
 					// Add external contacts ids
 					$commande_static->contacts_ids = $commande_static->liste_contact(-1, 'external', 1);
+					// Add online_payment_url, cf #20477
+					require_once DOL_DOCUMENT_ROOT.'/core/lib/payments.lib.php';
+					$commande_static->online_payment_url = getOnlinePaymentUrl(0, 'order', $commande_static->ref);
```
eldy, line 240: If you add this property, you must also declare the property into the file commande.class.php

rycks: @eldy hmmm ok ... but I'm afraid it does not cause bugs in other parts that would be based on the list of properties to do a treatment or display, right? 

For example in that sort of code:

```
foreach ($object->fields as $key => $val) {
```

So i don't know if there are other parts of code to check before commit commande.class.php ?

eldy: I was just speaking about adding:
public $online_payment_url;
into top of header. No need to ass it into ->fields.

PR #20483, opened 2022-03-29
FIX avoid warning for missing cookie samesite params (and more secure)

hregis:


eldy:
This cookies contains data for screenoutput and no sensitive information. There is no reason to have them secured. Also having a different name if secured looks strange.
When do you get the warning ?

hregis:
@eldy 
i know but it's just for the "green" status ! ;-)
for example, check "set-cookie" : https://securityheaders.com/?q=demo.dolibarr.org&hide=on&followRedirects=on
https://www.sjoerdlangkemper.nl/2017/02/09/cookie-prefixes/

PR #20485, opened 2022-03-29
FIX : ref_client doesn't exists on supplier invoice, then ref_fourn needs to have a default value when we want to bill several supplier orders

atm-gauthier:


htdocs/fourn/commande/list.php
```diff
@@ -291,7 +291,7 @@
 				$objecttmp->fk_project = $cmd->fk_project;
 				$objecttmp->multicurrency_code = $cmd->multicurrency_code;
 				if (empty($createbills_onebythird)) {
-					$objecttmp->ref_client = $cmd->ref_client;
+					$objecttmp->ref_supplier = !empty($cmd->ref_supplier) ? $cmd->ref_supplier : dol_print_date(dol_now(), '%Y%m%d%H%M%S');
```
eldy, line 294: If we set a generated value as default value, it would be better if this value is the same for all the transaction.
So can you create the default ref outside of the loop ?

atm-gauthier: @eldy in fact i was wrong, we do need to have a different default ref supplier because this field needs to be unique in supplier invoices.
I just did a new commit which seems better.

htdocs/fourn/commande/list.php
```diff
@@ -291,7 +293,7 @@
 				$objecttmp->fk_project = $cmd->fk_project;
 				$objecttmp->multicurrency_code = $cmd->multicurrency_code;
 				if (empty($createbills_onebythird)) {
-					$objecttmp->ref_client = $cmd->ref_client;
+					$objecttmp->ref_supplier = !empty($cmd->ref_supplier) ? $cmd->ref_supplier : $default_ref_supplier+1;
```
eldy, line 296: This will works only for 2 values.
The 3rd one will have  $default_ref_supplier+1; that will be same than the 2nd one.

atm-gauthier: @eldy of course, sorry i probably was asleep while coding this morning...

PR #20491, opened 2022-03-30
Change class btn login

BB2A-Anthony:
[*Long description*]

@eldy Please check my others PR thank


eldy:
For the moment the butAction css is reserved for button that are into the Action bar.
We must keep this separation of css key so if a theme need different rendering, we can.

PR #20494, opened 2022-03-30
FIX Customer invoice - Credit invoice can be equal to zero

aspangaro:


htdocs/compta/facture/card.php
```diff
@@ -286,8 +286,8 @@
 
 		// On verifie signe facture
 		if ($object->type == Facture::TYPE_CREDIT_NOTE) {
-			// Si avoir, le signe doit etre negatif
-			if ($object->total_ht >= 0) {
+			// If credit invoice, the sign must be negative or equal to zero
```
eldy, line 289: Do you mean we can make a credit note of 0 euros ?

eldy:
Do you confirm that we can make a credit note of 0 euros ? this looks strange to me.

PR #20495, opened 2022-03-30
FIX : regression select extrafield search + fix function name

john-botella:
SEE #20366

## Instructions
In lists, extrafields of type 'select' become à simple select but in past we were allowed to search for a simple text (for key only) but it's allow users to make multi search by playing with key names.

## FIX
- Allow multiselect for extrafield type select.
- Add new GETPOSTISARRAY function
- Fix also function naming by deprecating function checkVal


htdocs/core/tpl/extrafields_list_search_sql.tpl.php
```diff
@@ -59,13 +59,14 @@
...
+			if (in_array($typ, array('chkbxlst', 'checkbox', 'select'))) {
 				$mode_search = 4; // Search on a multiselect field with sql type = text
 			}
 			if (is_array($crit)) {
 				$crit = implode(' ', $crit); // natural_search() expects a string
-			} elseif ($typ === 'select' and is_string($crit) and strpos($crit, ' ') === false) {
-				$sql .= " AND (".$extrafieldsobjectprefix.$tmpkey." = '".$db->escape($crit)."')";
+			} elseif ($typ === 'select' and is_string($crit) and strpos($crit, ',') === false) {
+				$critSelect = implode("','", array_map(array($db, 'escape'), explode(',', $crit)));
+				$sql .= " AND (".$extrafieldsobjectprefix.$tmpkey." IN ('".$critSelect."') )";
```
eldy, line 69: When you have a sql forging a IN, the var $critSelect must be escaped with 
`$db->sanitize($critSelect, 1)`

atm-john: I already do a 
```array_map(array($db, 'escape'), explode(',', $crit))```

atm-john: and make a $db->sanitize after could sanitize an escape char so $db->sanitize could maybe unsecure query in this case, no ?

PR #20498, opened 2022-03-30
Avoid grouping lines with ordered lines

andreubisquerra:
It makes no sense to group a new line inside line already sent to the kitchen. If it is grouped, it is not printed in the kitchen and different prices or notes cannot be applied.

### DIFF
--- a/htdocs/takepos/invoice.php
+++ b/htdocs/takepos/invoice.php
@@ -535,7 +535,7 @@ if ($action == "addline") {
 	}
 
 	$idoflineadded = 0;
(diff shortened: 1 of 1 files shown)


eldy:
To not forgot your comment, why not adding it as comment into code, under the if, you can add the

`// It makes no sense to group a new line inside line already sent to the kitchen. If it is grouped, it is not printed in the kitchen and different prices or notes cannot be applied.
`

andreubisquerra:
Done!

PR #20517, opened 2022-04-01
FIX 16.0 - copy `title` over `ref` on FactureFournisseurRec object

atm-florianm:
# FIX copy `ref`-like field over `ref`

`FactureFournisseurRec`, like other document template objects (`FactureRec`, `FichinterRec`), has no `ref` field.

The technical "substitute" of the `ref` property the `titre` property. However, some functions, e.g. `getNomUrl()`, expect a `ref` property (the visible part of the returned link will be empty if `ref` is not set).

This PR ensures the value of the `titre` property is copied over `ref` whenever it is set by a method of the class, so that methods of the object (or methods of the CommonObject) expecting `ref` will work as expected.

The `table_ref_field` property is used so that a (hypothetical) renaming of the `field` property will require less code rewriting.

### DIFF
--- a/htdocs/fourn/class/fournisseur.facture-rec.class.php
+++ b/htdocs/fourn/class/fournisseur.facture-rec.class.php
@@ -242,7 +242,7 @@ class FactureFournisseurRec extends CommonInvoice
 
 		// Clean parameters
 		$this->titre = empty($this->titre) ? '' : $this->titre;
-        $this->ref = $this->{$this->table_ref_field};
+		$this->ref = $this->{$this->table_ref_field};
 		$this->ref_supplier = empty($this->ref_supplier) ? '' : $this->ref_supplier;
 		$this->usenewprice = empty($this->usenewprice) ? 0 : $this->usenewprice;
 		$this->suspended = empty($this->suspended) ? 0 : $this->suspended;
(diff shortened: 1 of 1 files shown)


htdocs/fourn/class/fournisseur.facture-rec.class.php
```diff
@@ -242,6 +242,7 @@ public function create($user, $facFournId, $notrigger = 0)
 
 		// Clean parameters
 		$this->titre = empty($this->titre) ? '' : $this->titre;
+		$this->ref = $this->{$this->table_ref_field};
```
eldy, line 245: This syntax is controversial and may be deprecated. 
Prefer to use
```
$keyforref = $this->table_ref_field;
$this->ref = $this->$keyforref;

```

atm-florianm: > This syntax is controversial

I haven't been able to find controversy regarding this particular syntax… I found an RFC for deprecating "${}" inside quotes and the discussion included a debate on the whole concept of variable variable (which would also include `$this->$keyforref`), but the debate was off-topic.

If you have more info on the possible upcoming deprecation of brackets for variable variables, I am interested.

PR #20523, opened 2022-04-01
NEW 16.0 - proposal for a prefix-based consistency mechanism for trigger names

atm-florianm:
# NEW - consistency for trigger names

### Rationale
Currently it is up to programmers to ensure that trigger names are consistent, and this is not always easy, which leads to inconsistency.

This PR proposes a mechanism based on a class constant (that can be overridden by child classes of `CommonObject`).

A basic check is performed to make sure the prefix is not added twice to make it easier to transition gradually (i.e. class by class) from the current code to the proposed mechanism. It will change nothing for classes that don't override `TRIGGER_PREFIX`.

It will also not help fixing inconsistencies in suffixes (such as `MYOBJECT_MODIFY` vs `MYOBJECT_UPDATE`).

### Example:
```php
class MyObject extends CommonObject {
    const TRIGGER_PREFIX = 'MYOBJECT';
    /* … */
    public function delete($user, notrigger = false) {
        // some code here
        $this->call_trigger('DELETE'); // instead of 'MYOBJECT_DELETE'
    }
}
```
### Warning
While exploring the feasibility of this feature, I came across some calls to `call_trigger()` that might cause problems if the class of the object in question had a `TRIGGER_PREFIX` (for instance in `actions_sendmails.inc.php`) but I think it wouldn't be too difficult to fix).


eldy:
So you replace trigger name into DAO file
 "MYOBJECT_DELETE"
into 
"DELETE" + a prefix
also both stored into DAO file.

I do not really see the gain. But i see the lost: Making an impact analysis into code to find all calls of a trigger name by a grep into code will no more be possible. Even if this is not the expected way, a trigger may be called outside of the DAO.
I would prefer to keep the key   "MYOBJECT_DELETE" but we can add a control to check that the trigger key start with "MYOBJECT_".  So you have your consistency check and we keep a key that can be easily found for impact analysis or debug in code.

atm-florianm:
> I do not really see the gain.

The gain would be that, by adding a prefix automatically, it would make it harder to use inconsistent prefixes.

> But i see the lost

I agree that analysis would be made more difficult, so I am happy with your alternative proposal (add a control to check that the trigger key starts with "MYOBJECT_").

I have updated my PR but I am not sure whether this should be an exception (maybe just a dol_syslog? or a dol_shutdown with a message?).

htdocs/core/class/commonobject.class.php
```diff
@@ -5602,7 +5603,9 @@ public function call_trigger($triggerName, $user)
 	{
 		// phpcs:enable
 		global $langs, $conf;
-
+		if (!empty(self::TRIGGER_PREFIX) && strpos($triggerName, self::TRIGGER_PREFIX . '_') !== 0) {
+			throw new Exception('The trigger "' . $triggerName . '" does not start with "' . self::TRIGGER_PREFIX . '_" as required.');
```
eldy, line 5607: What happen when you do a "throw" outside a try catch context ?

atm-florianm: In a standard apache context, I think this causes a fatal error.

atm-florianm: I changed it accordingly.

PR #20532, opened 2022-04-04
FIX - Fix the adding of lines in the create invoice functions

kkhelifa-opendsi:
FIX - Fix the adding of lines in the create invoice functions

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -800,6 +800,10 @@ class Facture extends CommonInvoice
 							$fk_parent_line = 0;
 						}
 
+						// Complete vat rate with code
+						$vatrate = $newinvoiceline->tva_tx;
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/class/facture.class.php
```diff
@@ -810,7 +814,37 @@ public function create(User $user, $notrigger = 0, $forceduedate = 0)
 							$newinvoiceline->fk_remise_except = $discountId;
 						}
 
-						$result = $newinvoiceline->insert();
+						$result = $this->addline(
```
eldy, line 817: The future is to abandon the method addline that is stored into object and instead use the method   ->insert of the objectline.
So old code seems nearer the target.
Can you try your fix keeping the "->insert" ?

philippe-opendsi: Hello
What about all the calculation made in addline ?

eldy: Hum, right.
So no choice, let's go with addline().

PR #20534, opened 2022-04-04
avoid fetch_objectlinked for each page of the document

atm-greg:
# FIX optimize PDF Generation
Avoid fetch_objectlinked for each page.
the _pagehead method of each PDF model calls "pdf_writeLinkedObjects" which calls "pdf_getLinkedObjects" which make a $object->fetchObjectLinked() on each call.

It induces that we loose several seconds on each printed page for something already done.

the fix consist in flaging the object to know that the full load of its objectLinked have already been done.

Ex. : Situation invoice whith 600 lines generate 25 pages for the PDF
Without this fix => 34 seconds
With the fix => 3 seconds


htdocs/core/lib/pdf.lib.php
```diff
@@ -2282,7 +2282,9 @@ function pdf_getLinkedObjects(&$object, $outputlangs)
 
 	$linkedobjects = array();
 
-	$object->fetchObjectLinked();
+	if (empty($object->linkedObjectsFullLoaded)) {
```
eldy, line 2285: What you do is implementing cache.
There is 2 way to manage a cache: 100% by the caller. In this case, nothing must be modified into the $object. 

Example with v15, a cache can be stored into variable $conf->cache that is dedicated for that.
So you can replace this with
```
if (!isset($this->cache['fetchObjectLinked_'.$object->element.'_'.$object->id])) {
              $object->fetchObjectLinked();
              $this->cache['fetchObjectLinked_'.$object->element.'_'.$object->id] = $object->linkedObjects;
} else {
             $object->linkedObjects =  $this->cache['fetchObjectLinked_'.$object->element.'_'.$object->id];
}
```
This is solution 1.

Solution 2 is that cache is managed 100% into the called method fetchObjectLinked
In such a case the test 
if (empty($object->linkedObjectsFullLoaded)) {
must not be here but inside the  $object->fetchObjectLinked();

For this situation, i would prefer solution 2.

atm-greg: Thank you for the tip.
I enjoy learning something to be a better developer.
I  implemented the second solution

atm-greg:
@eldy conflicts solved

PR #20536, opened 2022-04-04
add class name on tds  expensereport rule card

atm-jpb:
# new  : add class name on td

### DIFF
--- a/htdocs/admin/expensereport_rules.php
+++ b/htdocs/admin/expensereport_rules.php
@@ -185,14 +185,14 @@ if ($action != 'edit') {
 
 	echo '<table class="noborder centpercent">';
 
-	echo '<tr class="liste_titre headerforminput">';
-	echo '<th class="headerlinecolapplyto">' . $langs->trans('ExpenseReportApplyTo') . '</th>';
-	echo '<th class="headerlinecoltype">' . $langs->trans('Type') . '</th>';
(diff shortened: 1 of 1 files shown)


htdocs/admin/expensereport_rules.php
```diff
@@ -185,31 +185,31 @@
...
-	echo '<tr class="liste_titre">';
-	echo '<th>' . $langs->trans('ExpenseReportApplyTo') . '</th>';
-	echo '<th>' . $langs->trans('Type') . '</th>';
-	echo '<th>' . $langs->trans('ExpenseReportLimitOn') . '</th>';
-	echo '<th>' . $langs->trans('ExpenseReportDateStart') . '</th>';
-	echo '<th>' . $langs->trans('ExpenseReportDateEnd') . '</th>';
-	echo '<th>' . $langs->trans('ExpenseReportLimitAmount') . '</th>';
-	echo '<th>' . $langs->trans('ExpenseReportRestrictive') . '</th>';
+	echo '<tr class="liste_titre headerforminput">';
+	echo '<th class="headerlinecolapplyto">' . $langs->trans('ExpenseReportApplyTo') . '</th>';
```
eldy, line 189: To match other use, i suggest to use the same css for the same column, so 
linecoltype
instead of
headerlinecoltype

We already have the th vs td to make the difference, and we also have the css 'headerforminput' (i suggest to rename it 'headerexpensereportrules') on the tr too.

atm-jpb: agree with that. made the changes

PR #20538, opened 2022-04-04
NEW propagate demand reason in invoice card

lvessiller-opendsi:
NEW propagate demand reason in invoice card

- add "fk_input_reason" in invoice table

- add "Input reason" in invoice card (create and update forms)
image

- the "input reason" is pre-filled with the value of source object (proposal or order)


htdocs/install/mysql/migration/15.0.0-16.0.0.sql
```diff
@@ -304,3 +304,6 @@ UPDATE llx_c_availability SET type_duration = 'w', qty = 4 WHERE code = 'AV_4W';
 ALTER TABLE llx_boxes_def ADD COLUMN fk_user integer DEFAULT 0 NOT NULL;
 
 ALTER TABLE llx_contratdet ADD COLUMN rang integer DEFAULT 0 AFTER info_bits;
+
+ALTER TABLE llx_facture ADD COLUMN fk_input_reason integer NULL DEFAULT NULL AFTER last_main_doc;
```
eldy, line 308: Having this stored on an invoice looks strange to me. 
what is it for ? 
Knowing the way the user contact us ? if yes, it means this must be stored into proposal or order (and it seems it is already stored on this level).

lvessiller-opendsi: @eldy 
If ever you create invoices without any source document, it's interesting to retrieve this information.

lvessiller-opendsi:
@eldy 
I resolved conflicts.

eldy:
If ever you create invoices without any source document, it's interesting to retrieve this information.
But If ever you create invoices without any source document there is no reason to propagate the information from the source document (it does not exists).
End user that need this info because it uses invoice without using proposal or order, can still create an extrafield.
I see no reason to break the normal form of the database to store an information already store on the original document.

lvessiller-opendsi:
@eldy
And if ever you create invoices both with AND without source document, it is a pity to have to input this information on the invoice though it could be propagated.
Data analysis would then be made on invoices only instead of having to navigate between different documents and tables

lvessiller-opendsi:
@eldy 
I need this field in invoice as I explained to you.
Could you see if you can merge it ?

lvessiller-opendsi:
@eldy 
Could you merge it ?

lvessiller-opendsi:
@eldy 
I resolved all conflcts.

lvessiller-opendsi:
@eldy 

lvessiller-opendsi:
@eldy 

lvessiller-opendsi:
@eldy 
I resolved all conflicts and I also moved "create bills" mass action in order list (due to a pull from develop branch").
Please, could you see to merge it before having others conflicts ?

lvessiller-opendsi:
@eldy

eldy:
> If ever you create invoices without any source document, it's interesting to retrieve this information.
> But If ever you create invoices without any source document there is no reason to propagate the information from the source document (it does not exists).
> End user that need this info because it uses invoice without using proposal or order, can still create an extrafield.
> I see no reason to break the normal form of the database to store an information already store on the original document.

eldy:
> @eldy And if ever you create invoices both with AND without source document, it is a pity to have to input this information on the invoice though it could be propagated. Data analysis would then be made on invoices only instead of having to navigate between different documents and tables

Duplicating data and storing data at the wrong place in hope to make analysis easier for end user is surely not a good solution.
The data structure must remain clean. If analysis is difficult, the solution is in providing a feature that make reporting but we must not solve this need by dirtying the database structure.
A) If only invoice is used, the need can be address by an extrafield.
B) If you use order, you probably want to analyze source of order by reading the order
If you are not in cas A or B, this is a not common case and we must not break the normal form of the database to solve a need that is not the common case. There is still some uncommon situation that will need external modules or features to be solved and we must take care to not include everything in the core if the meaning or way is doubtful.

lvessiller-opendsi:
@eldy 
I understand your advice but you have the same propagation between customer proposals and orders.
Do you prefer, not to have the same demand reason between order and invoice ? (without propagation)

philippe-opendsi:
> B) If you use order, you probably want to analyze source of order by reading the order
I agree, but it means that the user is not able to have this information directly in Dolibarr because this information is not availaible on the invoice list, you have no standard export to have this information.
As said by @lvessiller-opendsi this propagation exit between proposals and order. Not having it on invoice make Dolibarr seems inconsistent for user.
What about user having both case A and B ?

eldy:
The propagation between between proposals and orders is not really a propagation :
- You can use the proposal without order.
- You can also use the order without proposal.
And because the "reason for proposal" is a field of proposal and the reason for order is a field of an order, the field is found in both table proposal and order. In real life, field will be filled only when making proposal, and if proposal are not used, will be filled only into order (the reason why field is also into order).
Sometimes users are using both proposal and orders. For such cases, to avoid to enter duplicate data, we copy the field from proposal into order, because when both proposal and order are used, we don't know where users decide to store this information.
But there is no reason to have such a field into the database of the table invoice. It is a sale representative information so information is not know by invoice manager (when it is by proposal or order manager). As it is a sale tracking information, it will be filled in the proposal or into the order (depending of end user choice) but not into invoice.

Of course, technically, we can duplicate all the column of proposal or order into the table invoice structure so we are able to propagate all information of proposal or order into invoices but this is not a good practice to get a good integrity of data and have data value consistent.

PR #20539, opened 2022-04-04
Bug: Cannot set Thirdparty customer accountancy code to an empty string

bomuux:
FIX
Cannot change Thirdparty customer accountancy code to an empty string, due to property name deprecation (code_compta to code_compta_client).


bomuux:
Sorry, bug not resolved by this PR

PR #20541, opened 2022-04-04
fix: Bad dirthday displayed date (d-1) in the Bithday Widget

FHenry:
User bithday date set to 05/04/2022 into user card
image

Displayed in the box 04/04/2022
image

https://www.dolibarr.fr/forum/t/date-anniversaire-erreur-1-jour/39717/3

### DIFF
--- a/htdocs/core/boxes/box_birthdays.php
+++ b/htdocs/core/boxes/box_birthdays.php
@@ -119,7 +119,7 @@ class box_birthdays extends ModeleBoxes
 
 					$this->info_box_contents[$line][] = array(
 						'td' => 'class="center nowraponall"',
-						'text' => dol_print_date($dateb, "day").' - '.$age.' '.$langs->trans('DurationYears')
+						'text' => dol_print_date($dateb, "day", 'tzserver').' - '.$age.' '.$langs->trans('DurationYears')
 					);
 
 					/*$this->info_box_contents[$line][] = array(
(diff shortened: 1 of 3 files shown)


htdocs/core/boxes/box_birthdays.php
```diff
@@ -119,7 +119,7 @@ public function loadBox($max = 20)
 
 					$this->info_box_contents[$line][] = array(
 						'td' => 'class="center nowraponall"',
-						'text' => dol_print_date($dateb, "day", 'gmt').' - '.$age.' '.$langs->trans('DurationYears')
+						'text' => dol_print_date($dateb, "day").' - '.$age.' '.$langs->trans('DurationYears')
```
eldy, line 122: If you remove the 'gmt' here it means the date shown depends on the timezone.
But a birthdate must be the same whatever is the position on earth.
So for me trouble is in the function that store the birthdate, probably using this->db->idate() instead of this->db->idate(..., 'gmt').
So we store in gmt and we read in gmt.

FHenry: If I try to put $this->db->idate(..., 'gmt') in update user.class.php method the result is still on day before input date (from selectDate Input). If I try to add gmt to selectDate from user/card.php result is the same. (even worth each time I edit save the card the birthdate is -1 days


I've swtich to all tzserver, like that the date is related to dolibarr server..
It work like that but I'm not sure if this what you want

PR #20542, opened 2022-04-05
FIX 16.0 - supplier invoice card creates two invoices instead of one

atm-florianm:
# FIX double supplier invoice creation from template

The conditions of the two `if`s overlap, causing the supplier invoice to be created twice (once from the block handling creation from templates and once from the block handling standard creation).

I also renamed a variable (`$fac_rec` -> `$fac_recid`) and used it instead of GETPOST
 + changed some conditions to not just test empty but also negative values for `$fac_recid` in case the `<select>`'s empty value changes
 + fixed a missing supplier ref assignment

### DIFF
--- a/htdocs/fourn/facture/card.php
+++ b/htdocs/fourn/facture/card.php
@@ -76,7 +76,7 @@ $lineid		= GETPOST('lineid', 'int');
 $projectid = GETPOST('projectid', 'int');
 $origin		= GETPOST('origin', 'alpha');
 $originid = GETPOST('originid', 'int');
-$fac_rec = GETPOST('fac_rec', 'int');
+$fac_recid = GETPOST('fac_rec', 'int');
 
 // PDF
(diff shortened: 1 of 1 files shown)


htdocs/fourn/facture/card.php
```diff
@@ -929,7 +929,7 @@
 		}
 
 		// Standard invoice or Deposit invoice, not from a Predefined template invoice
-		if (GETPOST('type') == FactureFournisseur::TYPE_STANDARD || GETPOST('type') == FactureFournisseur::TYPE_DEPOSIT && GETPOST('fac_rec') <= 0) {
```
atm-florianm, line 932: On this line I didn't just reorder the conditions but I added missing parentheses

htdocs/fourn/facture/card.php
```diff
@@ -885,7 +885,7 @@
 		}
 
 		// Standard invoice or Deposit invoice, created from a Predefined template invoice
-		if ((GETPOST('type') == FactureFournisseur::TYPE_STANDARD || GETPOST('type') == FactureFournisseur::TYPE_DEPOSIT) && GETPOST('fac_rec', 'int') > 0) {
```
atm-florianm, line 888: On this line it is simply a reordering of the conditions + using the variable instead of re-calling GETPOST

PR #20550, opened 2022-04-05
Fix 16.0 deprecate amount field on facture fourn?

atm-florianm:
# NEW: deprecate `amount` field + FIX SQL error on update

I couldn't figure what `amount` field of supplier invoices is used for; I believe it is obsolete (if it is not, maybe this is an opportunity to document it).

I also noticed this field (as well as fk_project) causes an SQL error when it is set as an empty string (because `isset` will return true if the field value is `''`).


eldy:
Yes field "amount" and "remise" are not more used since a long time (on all objects). We use "total_..." instead and "remise_percent" (on line level only).
We can clean code and sql (i started with a first PR).

PR #20551, opened 2022-04-05
NEW display errors in a message box after generating documents

lvessiller-opendsi:
NEW display errors in a message box after generating documents

When errors occur on generating a document :

- before errors were displayed with "dol_print_error" method : 
GeneratePrintError1

- now errors are displayed in a message box : 
GeneratePrintError2

And errors are logged in log file

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -5380,16 +5380,19 @@ abstract class CommonObject
 					return 1;
 				} else {
 					$outputlangs->charset_output = $sav_charset_output;
-					dol_print_error($this->db, "Error generating document for ".__CLASS__.". Error: ".$obj->error, $obj->errors);
+					dol_syslog("Error generating document for ".__CLASS__.". Error: ".$obj->error, LOG_ERR);
+					setEventMessages($obj->error, $obj->errors, 'errors');
 					return -1;
 				}
 			} else {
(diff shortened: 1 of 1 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -5380,16 +5380,19 @@ protected function commonGenerateDocument($modelspath, $modele, $outputlangs, $h
 					return 1;
 				} else {
 					$outputlangs->charset_output = $sav_charset_output;
-					dol_print_error($this->db, "Error generating document for ".__CLASS__.". Error: ".$obj->error, $obj->errors);
+					dol_syslog("Error generating document for ".__CLASS__.". Error: ".$obj->error, LOG_ERR);
+					setEventMessages($obj->error, $obj->errors, 'errors');
```
eldy, line 5384: The method commonGenerateDocument is not a method into a GUI context (not a page). It is also used in batch context.
So the setEventMessages must not be used here. Can you try to replace with 
```
$this->error = $obj->error
$this->errors = $obj->errors

```

lvessiller-opendsi: @eldy 
You're right.
I replaced it.

lvessiller-opendsi:
@eldy
I removed "setEventMessage" as you suggested to me.
I think you can merge it.

PR #20557, opened 2022-04-06
NEW Supplier order - Introduce constant ORDER_SUPPLIER_METHOD_BY_DEFAULT to define order method by default

aspangaro:
image

image


eldy:
There is already a generic feature to set the default value of a combo. Like hidden option it is feature designed for integrators but i don't think we need a hidden option for each field now that we have a generic solution. Why not trying to have it compatible with this page ?
For end user, we can imagine to add a "default for sale" and "default for purchase" column into the dictionnary.

Also if we want a more friendly feature, it should be implemented by added a column bydefault in dictionnary.

PR #20561, opened 2022-04-06
change restrictedarera call on pages

atm-jpb:
# fix  : 

impossible to read the payment transfer due to a bad definition of the object even if the rights are defined for the user

1) added paymentbybanktransfer feature in security lib.
2) modified calls to restrictedArea
3) replacing the conf->module call with the correct module name


eldy:
Which bug does this fix ?

atm-jpb:
added context

atm-jpb:
added the bug description  in comment

PR #20564, opened 2022-04-06
NEW : Reverse movement product consumption

atm-kevin:
# NEW

Addition of a trash icon on the lines of consumed products.
If I delete a product line it does the reverse stock movement and the line is removed.
If I delete the product global line and there are consumed product lines, the product global line is removed and reverse stock movements of all consumed lines are performed.


eldy:
Can you paste a screenshot that show the new rendering of the page, and check that there is a tooltip when you put the mous on the icon to explain the rule so for example "Delete line and revert stock" in the title of the link so user understand that action will not only remove the line but also change the stock ?

atm-kevin:
revert_stock_movement_global


revert_stock_mov_detail_line

htdocs/mrp/class/mo.class.php
```diff
@@ -750,12 +750,79 @@ public function delete(User $user, $notrigger = false)
...
+		$langs->load('stocks');
+
 		if ($this->status < 0) {
 			$this->error = 'ErrorDeleteLineNotAllowedByObjectStatus';
 			return -2;
 		}
 
-		return $this->deleteLineCommon($user, $idline, $notrigger);
+		$productstatic = new Product($this->db);
+		$fk_movement = GETPOST('fk_movement', 'int');
```
atm-maxime, line 762: You must not use GETPOST in a class

Humml87:
Hi,
this PR has some Bugs.
The Bug issue for this is: #28533 
Best regards
Christian

PR #20571, opened 2022-04-06
Update llx_c_paiement.sql

dolibit-ut:
add PayPal


eldy:
I renamed code to PPL, it sound more "Paypal" than PYL.
Also i added a number after 100 (before 100 is "real" payment mode and after are payment services that can suggest different payment modes).

dolibit-ut:
okay

PR #20572, opened 2022-04-06
NEW - Right for stats orders

BB2A-Anthony:
# NEW|New [*Right for stats orders*]
[*Long description*]


htdocs/core/modules/modCommande.class.php
```diff
@@ -178,7 +178,13 @@ public function __construct($db)
 		$this->rights[$r][4] = 'commande';
 		$this->rights[$r][5] = 'export';
 
-
+		$r++;
```
eldy, line 181: Does it has a sens to add a permission to read stats that differs than permission to export. Because if a user has permission to export and not to read stats, he can export and read stats from excel !
So this is a permission that does not bring any security. And a security features that does not provide any security at all should be avoided.
what do you think of adding a hidden constant:
MAIN_NEED_EXPORT_PERMISSION_TO_READ_STATISTICS
By defaut, the read permission is enough, if this option is set, the export permission will be required.

bb2a: For me it would be rather the opposite export allows to externalize the data.
If I have the right to internal statistics I can export

BB2A-Anthony:
> Does it has a sens to add a permission to read stats that differs than permission to export. Because if a user has permission to export and not to read stats, he can export and read stats from excel !
So this is a permission that does not bring any security. And a security features that does not provide any security at all should be avoided.
What do you think of adding a hidden constant:
MAIN_NEED_EXPORT_PERMISSION_TO_READ_STATISTICS
By defaut, the read permission is enough, if this option is set, the export permission will be required. 

@eldy I changed the permissions as you suggested

Issue #20562, opened 2022-04-06
Bad rights for "Rectrutement" module


atm-adrien:
### Bug

When we tried to access to recruitment card without the thirdparty rights "Consulter les tiers (sociétés) liés à l'utilisateur" and "Étendre l'accès à tous les tiers ET leurs objets...", we have not the access.
For me, thirparty's rights are not linked to recruitment module, so why sould we have these rights to access to recruitment module screen ? 

After vérification the problem's origtin is in the checkUserAccessToObject function (security.lib.php) : 
- we go though the condition : elseif (!in_array($feature, $nocheck)) [1]
- In this condition, we go though the condition : elseif (!empty($conf->societe->enabled) && !$user->rights->societe->client->voir)  [2] but needed param are not filled in the function call ($dbt_keyfield [3])

What can we do ? 
Filling needed params ? 

we can also make the condition : elseif (!empty($conf->multicompany->enabled)) go back up before the condition : elseif (!empty($conf->societe->enabled) && !$user->rights->societe->client->voir) in order to make the first one a priority [4]

![image](https://user-images.githubusercontent.com/67913809/161949044-6e1e98b4-f732-4308-96ff-65631710e1b7.png)

### Environment Version

14.0

### Steps to reproduce the behavior

Create an user with all the recruitment rights but without thirparties right 
Try to go on job card in recruitment module

PR #20575, opened 2022-04-07
Fix fill parameter for recruitment 14.0

atm-adrien:
# FIX : Recruitment access problem  

Same problem : https://github.com/Dolibarr/dolibarr/issues/20562

There is a problem on the recruitment access right.
En fact, with all recruitments rights but without any thirparty richts, we can't access to the recruitment cards.

It is due to the function checkUserAccessToObject
To correct it, I modified the function to take into account the case where the user does not have any thirparty permissions.

Is this correctoin good ?
Is there another way to correct this ?

### DIFF
--- a/htdocs/core/lib/security.lib.php
+++ b/htdocs/core/lib/security.lib.php
@@ -766,13 +766,13 @@ function checkUserAccessToObject($user, array $featuresarray, $objectid = 0, $ta
 					if (empty($dbt_keyfield)) {
 						dol_print_error('', 'Param dbt_keyfield is required but not defined');
 					}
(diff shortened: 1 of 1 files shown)


eldy:
I don't hink it is a good fix :
I agree we must have no link with thirdparty for the recruitment module. So this means we must not enter into the test that check the link with the company so we must not enter into test line 763
Instead we must enter into the test line 634 (like we do for the other RH module expensereport and leave). Can you test by just adding the recruitement module into the array $check ?

atm-adrien:
I'm goin to create a new PR with your fix to branch 14.0

PR #20587, opened 2022-04-07
FIX: holiday/leave requests: write status change emails in HTML

marc-dll:


htdocs/holiday/card.php
```diff
@@ -449,10 +449,9 @@
 				$subject = $societeName." - ".$langs->transnoentitiesnoconv("HolidaysToValidate");
 
 				// Content
-				$message = $langs->transnoentitiesnoconv("Hello")." ".$destinataire->firstname.",\n";
-				$message .= "\n";
+				$message = "<p>".$langs->transnoentitiesnoconv("Hello")." ".$destinataire->firstname.",</p>\n";
```
eldy, line 452: If you include <p> tags, it means the content will be html, so new line should be <br> and not \n.
Are you sure the rendering of email is correct with this change ?

marc-dll:
I kept the `\n` new lines on purpose for ease of debugging (breathing space when `echo`ing or `var_dump`ing). Handled correctly by the MUA, the `<p>` tags are treated as paragraphs with a small margin. I tested all those mails in Roundcube and they were rendered fine.

Capture d’écran de 2022-04-11 12-09-06

PR #20591, opened 2022-04-08
FIX - Right close order

BB2A-Anthony:
# FIX|Fix #[*Right close order*]
It is not possible that a user can cancel an order if he only has the right to close

### DIFF
--- a/htdocs/commande/card.php
+++ b/htdocs/commande/card.php
@@ -2628,7 +2628,7 @@ if ($action == 'create' && $usercancreate) {
 				}
 
 				// Cancel order
-				if ($object->statut == Commande::STATUS_VALIDATED && (!empty($usercanclose) || !empty($usercancancel))) {
+				if ($object->statut == Commande::STATUS_VALIDATED && !empty($usercancancel)) {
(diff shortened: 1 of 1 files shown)


htdocs/commande/card.php
```diff
@@ -2628,7 +2628,7 @@
 				}
 
 				// Cancel order
-				if ($object->statut == Commande::STATUS_VALIDATED && (!empty($usercanclose) || !empty($usercancancel))) {
+				if ($object->statut == Commande::STATUS_VALIDATED && !empty($usercancancel)) {
```
eldy, line 2631: Why removing the   !empty($usercanclose) 
It is a or with the condition   !empty($usercancancel)
so only one of this 2 conditions should be necessary, as expected.

bb2a: If a user does not have the right to cancel why can he do it if he has the right to close?

Security error!

We do not give the right to delete but he can do it!

bb2a: Why can a user who has the right to close cancel when there is an authorization defined for this?

I do not understand! For me it's a security error!

Issue #20582, opened 2022-04-07
Authorship recognition at ticket API messaging


alsoft10:
### Feature Request

Currently in Dolibarr 15.0.1, Dolibarr provides very simple ticket messaging via it's API. Generally API client is connected using an API key. The SuperAdmin key is generally used for API communication. Now here is the lack of features in the ticket API communication.

1. If a website user, third party or customer creates a ticket using API, Dolibarr can't recognize them. It simply logged SuperAdmin as the ticket author. But the author should be the user or customer.
2. If a website user, third party or customer  reply to a ticket with the message, again SuperAdmin is taken as the author of the message. While the message author would be the website user, customer or third party.

The above said features is not available in the DOLIBARR Ticket API. I think it's a very basic required features for a ticket based messaging system.

alsoft10:
@eldy I am actively waiting for feedback in this matter.

priojk:
What kind of feedback do you need? Whether it's ok to submit a pull request? Since it's quite straight forward, I would say, just go ahead.

However, since @pratushraj already created something and mentioned your issue - does this suit your needs? Could be a helpful input. Note: the pull request is based on 15.0, but since it's a new feature and not a bug, I guess it will rather go into the develop branch.

alsoft10:
@priojk you are correct. @pratushraj's issue will be fit for the purpose. Pratushraj already created the pull request for this feature request. You can check that at #20598.

PR #20598, opened 2022-04-09
Authorship recognition at ticket API messaging #20582

pratushraj:
# Instructions
*Dolibarr provides simple tickets Creation and messaging via its API. Generally, the API client is connected using an API key. The SuperAdmin key is generally used for API communication. Now here is the lack of features in the ticket API communication*

*You can simply do that to add "authorid": "0" for creating the ticket or sending the new message from the customer side.*

# FIX|Fix #[*20582 Authorship recognition at ticket API messaging*]


eldy:
This may create security trouble.
Any user can use the api to create a ticket by hidding it was created by himself. What is the goal ?
If goal is to avoid to send an email after ticket creation, a new param into API would be a better choice.

pratushraj:
@eldy  the goal of this is to make the ticket API communication more advanced. We want to integrate ticket API communication with another software having the customers. Now customer will directly reply to the ticket from the software. And the software will send the message to the Dolibarr system using the API. Now as the Dolibarr mistakenly makes author of the reply to the API owner, we can't track the user who actually is the author.

eldy:
There is no need to add a property authorid for that.
There is already the property  "fk_user_create"

Did you try to set   fk_user_create  into api call so when the api call ticket->create, the fk_user_create is set to this id ?

pratushraj:
Yes, I tried. But not working with the property "fk_user_create". Because the property "fk_user_create" always takes the value of the user Id. That's why I add the property "authorid".

eldy:
I checked the code in v16. I see no reason why setting fk_user_create into api call won't work. Did you try with the develop branch ?

pratushraj:
Yes, I Also checked in develop branch, that "fk_user_create" not working because "fk_user_create" took the value of the user related to the API key. That's why I introduced the new property called "authorid".

Issue #11335, opened 2019-06-13
Tax 2 disappears when converting a purchase order into a supplier invoice


nanoxp:
# Bug
*Tax 2 disappears when converting a purchase order into a supplier invoice*

## Environment
- **Version**: [*9.0.3*]
- **OS**: [*Ubuntu 18.04.2 lts*]
- **Web server**: [*Apache 2.4.29*]
- **PHP**: [*7.2.19*]
- **Database**: [*MySQL 5.7.26 *]

## Expected and actual behavior
*When converting a purchase order into a supplier invoice, taxes 1 and 2 taxes defined in the purchase order must be respected*

## Steps to reproduce the behavior
Dolibarr was set up to operate a second tax and the VAT tax rate + IEPS8 was created in the VAT Dictionary as shown below:
![imagen](https://user-images.githubusercontent.com/24461406/59453021-9c842580-8dd4-11e9-86dd-b885a70146b0.png)
When generating a Purchase order where we add an item with this type of VAT (VAT + IEPS8) we see that if it correctly shows Tax 2 of 8%:
![imagen](https://user-images.githubusercontent.com/24461406/59453072-b9b8f400-8dd4-11e9-9aa1-a5f710274437.png)
Validate and approve the purchase order to proceed to create invoice as shown below:
![imagen](https://user-images.githubusercontent.com/24461406/59453094-c2112f00-8dd4-11e9-9a99-6283a366acee.png)
We observe in the previous image that if it correctly indicates the type of tax and amount tax 2, but when clicking "Create draft" and creating the supplier invoice, we see that tax 2 does not appear anymore or in the totals as in the line of each article :
![imagen](https://user-images.githubusercontent.com/24461406/59453114-cdfcf100-8dd4-11e9-8d38-463d3837df57.png)

# Feature Request
When converting a purchase order into a supplier invoice, the taxes of 1 and 2 taxes defined in each article of the purchase order must be respected.

nanoxp:
I have applied the aforementioned correction and the problem persists with my report # 11335

nanoxp:
Hello,

 

I applied the aforementioned correction in via 8afffc8 and the problem was not corrected, follow the report # 11335

 

Thanks for the support

Edgar Ruiz

 

De: Laurent Destailleur [mailto:notifications@github.com] 
Enviado el: lunes, 17 de junio de 2019 09:01 a. m.
Para: Dolibarr/dolibarr <dolibarr@noreply.github.com>
[...]

Issue #6471, opened 2017-02-27
Bug: Tax 2 not calculated on suppliers orders invoices


mrsarge:
# Bug
In suppliers orders, second tax is applied in the order and draft invoice, but not on final invoice. everything is fine on customers orders.

## Environment
- **Version**: [4.0.4] - [8.0.4]
- **OS**: [Ubuntu 16.04.1]
- **Web server**: [Apache 2.4.18]
- **PHP**: [PHP 5.6.31]
- **Database**: [MYSQL 5.7.24]

## Expected and actual behavior
Tax 2 amount should be shown on invoice and total due should include it.  

## Steps to reproduce the behavior
Create a new supplier third party with two tax.  
Create new supplier order for this third party.  
Add product or service with sales tax.  
Validate and approve order. Tax, Tax 2 and Total should be ok.   
Create invoice.  
Tax 2 line is not shown, but Total include it.  
Create draft.  
Tax 2 is 0.00 $ and Total doesn't include it.  
The invoice PDF also has a 0.00 Tax 2 line and Total doesn't include it either.

## Screenshots  
**Order page and PDF:**
![order](https://user-images.githubusercontent.com/6961825/30011418-2839dbac-9105-11e7-9733-0f9deab29d33.png)
![30011420-2a502324-9105-11e7-9e91-af815832bd52](https://user-images.githubusercontent.com/6961825/31409808-1c08f1f4-addb-11e7-8ffa-661f881c1928.png)

**Invoice page and PDF:**
![invoice](https://user-images.githubusercontent.com/6961825/30011429-377e8900-9105-11e7-9404-838cf41dded7.png)
![invoice_pdf](https://user-images.githubusercontent.com/6961825/30011430-39270976-9105-11e7-890b-0a5ef350b2b8.png)

mrsarge:
Just upgraded to 5.0.2 and issue is still there.

technoartista:
How to enable a second and 3rd tax for the already added third party . I can see it allows me to edit for new 3rd party but not allowing me to edit for existing 3rd party .

mrsarge:
Issue still present in 6.0.0.  Maybe similar to issue #7298

mrsarge:
Issue still present in 6.0.2.

mrsarge:
Partially fixed in Dolibarr 6.0.3 (same for 6.0.4).

The total on the invoice page and pdf (validated or not) is still wrong (second tax missing). 

[...]

PR #20600, opened 2022-04-10
FIX Tax 2 is lost when order is invoiced

javierybar:
Tax 2 is lost when an order is invoiced from mass actions.
Reported in:
https://github.com/Dolibarr/dolibarr/issues/11335
https://github.com/Dolibarr/dolibarr/issues/6471

I have observed this issue happening from a lot of years and in the current version it is still happening.


htdocs/core/actions_massactions.inc.php
```diff
@@ -657,6 +657,7 @@
 		} else {
 			// If we want one invoice per order or if there is no first invoice yet for this thirdparty.
 			$objecttmp->socid = $cmd->socid;
+			$objecttmp->fetch_thirdparty();
```
eldy, line 660: I validated PR but modified it after by changing this into 
$objecttmp->thirdparty = $cmd->thirdparty;

Can you confirm it still works as expected ?

PR #20603, opened 2022-04-11
FIX move multicompany tables in Multicompany module (since v16)

hregis:


eldy:
If you move tables out of core, it also means we must move all code that mention theses tables out of core.
tables structure definition and code using this structure must be on the same place.
Is this possible to remove also all the code that is using those tables ?

hregis:
@eldy 
Yes I will look to move the code in the module. Moreover, the code is not optimized for all scenarios.

eldy:
Because code that exploit the tables are still inside the core, we will keep the tables inside the core too.

PR #20608, opened 2022-04-11
Fix: Direct download don't work with multicompany

fmarcet:
If the document has been generated, for example, in entity 2, we cannot download it because by default it always considers that we are in entity 1.

### DIFF
--- a/htdocs/document.php
+++ b/htdocs/document.php
@@ -5,6 +5,7 @@
  * Copyright (C) 2005-2012 Regis Houssin        <regis.houssin@inodbox.com>
  * Copyright (C) 2010	   Pierre Morin         <pierre.morin@auguria.net>
  * Copyright (C) 2010	   Juanjo Menent        <jmenent@2byte.es>
+ * Copyright (C) 2022	   Ferran Marcet        <fmarcet@2byte.es>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


eldy:
When you request api with multicompany request is done by default on entity 1.
Did you try by adding into the HTTP header of your request the 
DOLAPIENTITY: 2
?

fmarcet:
Hello @eldy, I haven't tried it through the API. The erroneous operation that I have seen is, for example, when trying to download the PDF of the budget at the moment in which you are going to do the online signature. The link for that document is something like this: ../htdocs/document.php?hashp=qS22K67ncUVWP8c7To92b1KzkZDoi7pN No information about the document entity is generated and the only data we have is the hashp. That is why I had thought that it was a good solution to load the entity at the same time as the document data is loaded from the hashp.

hregis:
@fmarcet @eldy 
you have already the "entity" field value of an ecm file !

hregis:
@fmarcet @eldy this PR is good for me !

eldy:
> qS22K67ncUVWP8c7To92b1KzkZDoi7pN

The hash is already a unique id. 
And the id in which we make the request is sent into DOLAPIENTITY so the code can check after finding the file with the hash that the  entity of object is same than the entity request. If not security error, if yes, we return the file.
API should say in which entity we want to work with DOLAPIENTITY, similar to the entity in session when using GUI. If entity is a parameter on url, we have a security trouble somewhere (like if a user can overwrite is logged entity just by providing it on url).

PR #20612, opened 2022-04-12
NEW: ModuileBuilder - Add tabs view in module builder

creekorful:
When working with the module builder, I figured out that the tabs were not shown, I have added a basic page providing read access over the module tabs.

image


eldy:
As this is a new feature, can you push your PR into develop branch ? (stable version accepts only bug fixes).

creekorful:
@eldy Done. Thanks for the clarification 👍🏻

PR #20614, opened 2022-04-12
FIX : Missing entity on adding new VAT

atm-kevin:


htdocs/compta/tva/class/tva.class.php
```diff
@@ -136,6 +137,7 @@ public function create($user)
 		$sql .= "fk_user_creat,";
 		$sql .= "fk_user_modif";
 		$sql .= ") VALUES (";
+		$sql .= " '".$conf->entity."',";
```
eldy, line 140: Remove the ' ' in sql and use
$sql .= " ".((int) $conf->entity.", ";

PR #20615, opened 2022-04-12
NEW product categories filter on inventory list

lvessiller-opendsi:
NEW product categories filter on inventory list

- add product categories filter with "AND" or "OR" operator in this list : 
image


eldy:
Your last suggestion of presentation seems good (at least i understand it).
So let's go like this. However, in a near future (i hope), we will integrate a new widget for filter and this filter for categories may not be ported immediatly in the new tool. So to avoid to have the filter provided to user and then remove whennew filter widget will be published. So can you introduce a hidden constant so this will be visible only for people that ask it immediatly and it avoids to provide a feature removed at next version. It will be better to have it visible definitely when search widget will be ok for such cases too.lter and this filter for categories may not be ported immediatly in the new tool. So to avoid to have the filter provided to user and then remove whennew filter widget will be published. So can you introduce a hidden constant so this will be visible only for people that ask it immediatly and it avoids to provide a feature removed at next version. It will be better to have it visible definitely when search widget will be ok for such cases too.

lvessiller-opendsi:
@eldy 
I added an hidden const "MAIN_SEARCH_CATEGORY_PRODUCT_ON_LISTS" as you suggested.
I'm not sure I understood everything, could you chek it's enough and could you merge it ?

aspangaro:
image

lvessiller-opendsi:
@eldy 
All you suggested was realised, could you merge it ?

PR #20618, opened 2022-04-12
NEW : add linkedObjectBlock TPL for Tasks

atm-GregM:
# NEW : add linkedObjectBlock TPL for Tasks

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -8369,7 +8369,7 @@ class Form
 		global $conf, $langs, $hookmanager;
 		global $bc, $action;
 
-		$conf->project = (object)array("enabled"=>true);
+		$conf->project = (object) array("enabled"=>true);
 		$object->fetchObjectLinked();
 
 		// Bypass the default method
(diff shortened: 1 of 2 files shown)


htdocs/core/class/html.form.class.php
```diff
@@ -8369,6 +8369,7 @@ public function showLinkedObjectBlock($object, $morehtmlright = '', $compatibleI
 		global $conf, $langs, $hookmanager;
 		global $bc, $action;
 
+		$conf->project = (object)array("enabled"=>true);
```
eldy, line 8372: Need 1 space after the (object) and before array.

htdocs/core/class/html.form.class.php
```diff
@@ -8369,6 +8369,7 @@ public function showLinkedObjectBlock($object, $morehtmlright = '', $compatibleI
 		global $conf, $langs, $hookmanager;
 		global $bc, $action;
 
+		$conf->project = (object) array("enabled"=>true);
```
eldy, line 8372: The project is an object managed differently in Dolibarr. It is linked to other object with the field fk_project. So using "wild" links like suggested by the feature to link object each other will create a duplicate way to make project link making confusion and unconsistency into data. 
Please introduce a hidden constant for this because this will lead to very dangerous situation.

eldy:
Waiting suggested implementation for too long time so i close pr. Thanks to open a new one if feature still required.

PR #20622, opened 2022-04-13
FIX a user can not change permissions of admins

hregis:


eldy:
Currently the definition of permission "$user->rights->user->user->creer" is
Can create and edit permission of all users
So this change will break the specifications.

hregis:
@eldy 
if you only give these rights to a user, he will not be able to edit an admin card, 
but he can modify the rights of admins!
it is not normal ! only an admin can modify another admins ! right ?

```
Consulter les autres utilisateurs 
Consulter les permissions des autres utilisateurs (Autorisation disponible en mode Autorisations avancées uniquement) |  
Créer/modifier les utilisateurs internes/externes et leurs permissions
```

hregis:
@eldy 
Going a little deeper, there are some advanced permissions that are not respected!

hregis:
@eldy 
for example with the rights below, a user must be able to see the rights assigned to him without being able to modify them. But this is not the case, the user can modify his rights!

Capture d’écran 2022-04-13 à 15 42 59

eldy:
Yes, there is a lof of bugs into the implementation of advanced permission that does not respect the specifications so what the user expect. We must fix this.
But this is not a reason to introduce more bugs.

If you give the right to read and not modify a user to X,  yes X can read only and not modify an admin. this is what is expected.
And if you give X the right to modify permissions to everyone, yes X can modify permission of an admin. this is what is expected.
So i don't see any bugs. The code must always match the specification. Here behaviour looks strange because user did a strange setup. But the code must do what the setup asks. Currently I don't see any problem if a user X can modify the permission of any user if we gave him the permission to modify the permission of any user.

PR #20625, opened 2022-04-13
NEW Can use products categories to make inventory

lvessiller-opendsi:
NEW products categories in inventory card

- add products categories field in inventory card
image

- and in create and update forms (with multi-selecton categories)
image

- so you can start an inventory for one or several categories
image


eldy:
Can you confirm that the feature here is to define categories INSTEAD of one given product so inventory will be done on all product into one of the categories ?
If yes, can you rename the filter into  "or products with tags/category"   instead of   "tags of products"

lvessiller-opendsi:
@eldy 
I confirm it's a feature to define products categories instead of having only one given product.
I resolved conflicts and I put the suggested label on products categories filter.

lvessiller-opendsi:
@eldy 
Could you merge it ?

thomas-Ngr:
Hi @eldy, it seems this PR is ready, could you have a look and merge it ? Thanks in advance.

lvessiller-opendsi:
@eldy 
I don't know why I got a travis error.
Could you chek travis failure ?

lvessiller-opendsi:
@eldy

htdocs/install/mysql/tables/llx_inventory-stock.sql
```diff
@@ -28,8 +28,9 @@ CREATE TABLE llx_inventory
   fk_user_modif integer,                    -- user making last change
   fk_user_valid integer,                    -- valideur de la fiche
   fk_warehouse integer DEFAULT NULL, 
-  fk_product integer DEFAULT NULL, 
-  status integer DEFAULT 0, 
+  fk_product integer DEFAULT NULL,
+  fk_categories_product varchar(255) DEFAULT NULL, -- product categories id separated by comma
```
eldy, line 32: Because syntax fk_... is reserved for foreign key that point to a primary key, can you rename the field into just "categories_product".
PS: it is recommended to push PR on SQL structure change into a different PR so structure can be quickly validated and merged before the code is done.

lvessiller-opendsi:
@eldy 
I replace "fk_categories_product" with "categories_product" as you suggeested to me.
I resolve all conflicts and I will separate SQL changes for next PR.
Could you merge it ?

lvessiller-opendsi:
@eldy 
I resolved all conflicts and apply all your suggests, could you merge it ?

eldy:
Your last suggestion of presentation seems good (at least i understand it).
So let's go like this. However, in a near future (i hope), we will integrate a new widget for fiYour last suggestion of presentation seems good (at least i understand it).
So let's go like this. However, in a near future (i hope), we will integrate a new widget for filter and this filter for categories may not be ported immediatly in the new tool. So to avoid to have the filter provided to user and then remove whennew filter widget will be published. So can you introduce a hidden constant so this will be visible only for people that ask it immediatly and it avoids to provide a feature removed at next version. It will be better to have it visible definitely when search widget will be ok for such cases too.lter and this filter for categories may not be ported immediatly in the new tool. So to avoid to have the filter provided to user and then remove whennew filter widget will be published. So can you introduce a hidden constant so this will be visible only for people that ask it immediatly and it avoids to provide a feature removed at next version. It will be better to have it visible definitely when search widget will be ok for such cases too.

lvessiller-opendsi:
@eldy 
A travis error still exists but I don't know why ?

lvessiller-opendsi:
@eldy 
Could you check why I always have this travis error ?

lvessiller-opendsi:
@eldy 
Could you merge it ?

PR #19649, opened 2021-12-18
NEW FontAwesome - Add constant to define your own directory and font family

aspangaro:
1. Add constants : 

- MAIN_FONTAWESOME_DIRECTORY : to define your own directory (after DOL_URL_ROOT | "/theme/common/fontawesome-5" by default). If you use specific fontawesome kit or an another version (v6) or for custom template module

- MAIN_FONTAWESOME_FAMILY : to define font family (v5, v6, Free, Pro, etc | "Font Awesome 5 Free" by default)

- MAIN_FONTAWESOME_ICON_STYLE : to define icon style (fas, far, fal, fat, fad | 'fas' by default) 

2. Also : Deprecated compatibility with font awesome 4

PR #20630, opened 2022-04-14
NEW Update fontawesome library to 6.11

aspangaro:
Awesome compatibility with fontawesome 5.13 (FA 4 is deprecated). New free icons, method, etc

For the other distribution of dolibarr with a pro version of font awesome, please accept this PR :
https://github.com/Dolibarr/dolibarr/pull/19649

Have a nice day


eldy:
Switching completely with no transitionnel step is not a good idea.
You must implement code that continue to use the font awesome 5 and a hidden option that allow to switch to v6.
So developers will be able to validate during at least one transitional version that their module is compatible with font awesome v6 by switching an option in Dolibarr v16 with no risk to break nothing, before we can remove v5.
Like we did in Dolibarr v10 for transition from v14 to v15 where both solutions were possible

hregis:
@aspangaro @eldy 
The developers do with two versions per year... so a library that changes every 2 years!!! ;-)))

aspangaro:
@eldy  Ok for transitional step to add font awesome 6 library (even if it is fully compatible with version 5) but in this case, accept the other PR. #19649 

this will allow you to define anything you want, including future versions of font awesome or the compacted library licensed to professionals.

PR #20631, opened 2022-04-14
NEW Move setup logo to font awesome

aspangaro:
Before :

image

After : 

image


eldy:
font icon are good things for very small picto.
Here the image is to illustrate a page, we can image to replace it with a high quality clipart, we loose this ability.
Why replacing an illustration with a low quality font ? SVG is more recommended for such need.

aspangaro:
it was for the coherence of the interface of dolibarr, we have font awesome everywhere and there, we have a drawing which is there only to furnish... So we might as well remove it...

PR #20635, opened 2022-04-14
Fix : add of unique(entity,label) in llx_takepos_floor_tables table

Hystepik:
add of unique couple (entity,label) in llx_takepos_floor_tables table to prevent having same label of table in the same entity


hregis:
@eldy 
Sometimes you validate blind!
the bug is at the end of the road! ;-)

PR #20642, opened 2022-04-14
FIX: Call of printOriginObjectLine hook

kkhelifa-opendsi:
FIX: Call of printOriginObjectLine hook

This test is useless because we can't manage the children lines case with hook


eldy:
Merged into dev because it change behaviour of hook and can break compatibility of some external modules

PR #20646, opened 2022-04-15
Picto mass action  presetcommercial

BB2A-Anthony:
# FIX|Fix #[*Picto mass action  presetcommercial*]
[*Long description*]


aspangaro:
Already correct

PR #20649, opened 2022-04-15
fix: CREATE TABLE llx_element_tag without migration script 

FHenry:
Following: #18996

**History**
In migration dolibarr/htdocs/install/mysql/migration/3.2.0-3.3.0.sql the is a create table llx_element_tag. But there is no llx_element_tag.sql files

**Then** 
#19262 create table with llx_element_tag.sql and llx_element_tag.key.sql files without adding them into dolibarr/htdocs/install/mysql/migration/14.0.0-15.0.0.sql

This PR will just make it OK

### DIFF
--- a/htdocs/install/mysql/migration/14.0.0-15.0.0.sql
+++ b/htdocs/install/mysql/migration/14.0.0-15.0.0.sql
@@ -509,3 +509,16 @@ INSERT INTO llx_c_action_trigger (code,label,description,elementtype,rang) value
 
 -- VMYSQL4.3 ALTER TABLE llx_user MODIFY COLUMN fk_soc integer NULL;
 -- VPGSQL8.2 ALTER TABLE llx_user ALTER COLUMN fk_soc DROP NOT NULL;
+
+DROP TABLE IF EXISTS llx_element_tag; -- in migration 3.2.0 to 3.3.0 there is a element_tag table creation that is notin create table
+CREATE TABLE llx_element_tag
+(
+    rowid integer AUTO_INCREMENT PRIMARY KEY,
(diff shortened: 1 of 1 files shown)


htdocs/install/mysql/migration/14.0.0-15.0.0.sql
```diff
@@ -510,6 +510,20 @@ INSERT INTO llx_c_action_trigger (code,label,description,elementtype,rang) value
 -- VMYSQL4.3 ALTER TABLE llx_user MODIFY COLUMN fk_soc integer NULL;
 -- VPGSQL8.2 ALTER TABLE llx_user ALTER COLUMN fk_soc DROP NOT NULL;
 
+-- In migration 3.2.0 to 3.3.0 there is a element_tag table creation that is not in create table file
+DROP TABLE IF EXISTS llx_element_tag;
```
eldy, line 514: We should never drop a table into a migration, except if table is no more used into code. This means destruction loss of data for user.
Simple creation is enough, the migration wizard already ignore the error if table already exists.

FHenry: What do with people that have dolibarr since 3.2 with the table created from migration 3.2 to 3.3 but with a different structures than the one created here? 
Your right, the script is always executed on migration so drip isn't a good idea here anyway

eldy: > What do with people that have dolibarr since 3.2 with the table created from migration 3.2 to 3.3 but with a different structures than the one created here? 
In such a case, we must migrate by adding the field we need.

Issue #11125, opened 2019-05-04
[Feature Request] Move checkbox column as first column


eldy:
Move the checkbox on all list at begin as first column instead of last one.
All pages lists.php must be processed, but this can be done page after page.

benyaminl:
Just asking, does it need an option in settings so it have backward compatibility or change it all the way so the first check position become standard?

fuhraih:
I am starting this feature request. It is a good training for understanding how the screens are working.

fuhraih:
New PR with only this feature included

BB2A-Anthony:
Why not put it already in module builder?

BB2A-Anthony:
Is include in V16.

Module builder and Thirdpartie

Option is MAIN_CHECKBOX_LEFT_COLUMN

PR #20650, opened 2022-04-15
NEW - Move checkbox column as first column on Thirdparty list

BB2A-Anthony:
# NEW|New [*Move checkbox column as first column on Thirdparty list*]

A first to issue #11125


eldy:
I suggest to have (for the moment), by default still on right and starting implementing to be able to have it on left.
Can you confirm that after the PR, the "default" on old installation is still on right (we will revert the option that allow to choose when main screen will be compatible) ?
Can you also provide a screenshot to see result when on left ?

BB2A-Anthony:
> I suggest to have (for the moment), by default still on right and starting implementing to be able to have it on left. Can you confirm that after the PR, the "default" on old installation is still on right (we will revert the option that allow to choose when main screen will be compatible) ? Can you also provide a screenshot to see result when on left ?

I add imediatly option MAIN_CHECKBOX_LEFT_COLUMN .

image

image

BB2A-Anthony:
@eldy done

PR #20655, opened 2022-04-15
Move checkbox column as first column on modulebuilder

BB2A-Anthony:
# NEW|New [*Move checkbox column as first column on modulebuilder*]

Issue #11125

### DIFF
--- a/htdocs/modulebuilder/template/myobject_list.php
+++ b/htdocs/modulebuilder/template/myobject_list.php
@@ -536,7 +536,7 @@ if (!empty($moreforfilter)) {
 }
 
 $varpage = empty($contextpage) ? $_SERVER["PHP_SELF"] : $contextpage;
-if (empty($conf->global->MAIN_CHECKBOX_LEFT_COLUMN)) {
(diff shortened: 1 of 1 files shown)


eldy:
Can you provide a screenshot of result of a page with this change ?

htdocs/modulebuilder/template/myobject_list.php
```diff
@@ -536,8 +536,12 @@
 }
 
 $varpage = empty($contextpage) ? $_SERVER["PHP_SELF"] : $contextpage;
-$selectedfields = $form->multiSelectArrayWithCheckbox('selectedfields', $arrayfields, $varpage); // This also change content of $arrayfields
-$selectedfields .= (count($arrayofmassactions) ? $form->showCheckAddButtons('checkforselect', 1) : '');
+if (empty($conf->global->MAIN_CHECKBOX_LEFT_COLUMN)) {
```
eldy, line 539: Should be if !empty
Default is current behaviour and option move column on left.

bb2a: Yes it's done

Issue #20652, opened 2022-04-15
ADD EVENT, FROM, FILTER on recurring invoice model


lmag:
### Feature Request

- [x] ADD filter on FAC_REC_SOURCE compta/facture/list.php #20656 
- [ ] ADD Event on recurring invoice model
- [ ] ADD Event create invoice model
- [ ] ADD Event create invoice based on this modele with the rowID of the invoice genered
- [ ] ADD Event modify by
- [ ] ADD Tab with list filter by billing modèle in compta/facture/card-rec.p

PR #20656, opened 2022-04-15
NEW: Add column template invoice in invoice list

FHenry:
NEW: Add column template invoice in invoice list
issue #20652


htdocs/compta/facture/list.php
```diff
@@ -1500,6 +1512,12 @@
 		print '<td class="liste_titre">';
 		print '</td>';
 	}
+	if (!empty($arrayfields['f.fk_fac_rec_source']['checked'])) {
+		// Template Invoice
+		print '<td class="liste_titre maxwidthonsmartphone right">';
+		$form->selectInvoiceRec($search_fac_rec_source, 'search_fac_rec_source');
```
eldy, line 1518: I think that using a combo here is not a good idea.
If combo is a complete list of element, it must be reserved to element that does not increase with time. But for invoice and recurring invoice, we may have more and more... Or we must allow a method that allow to fill the list after clicking on 1 or 2 keys (like we do for product, thirdparty, ...).

To avoid to develop this a better choice would be to simply offer an input field to allow to search on a string.

FHenry: Ok, done, Input text rather select select combobox

PR #20658, opened 2022-04-15
NEW #20653 edit discount pourcentage for all lines in one shot

### LINKED ISSUE #20653: Can edit the percentage discount of lines in one shot
### Feature Request

Currently when editing an invoice, we can set the vat of all lines in one action using the picto "edit" on the title of the line.
image

The request is to be able to do the same for the discount % on lines (the column after the column qty) on the screenshot.

### Use case

_No response_

### Suggested implementation

_No response_

### Suggested steps

_No response_

Hystepik:
Can edit the percentage discount of lines in one shot for propal, sales order and invoice


DomiMacMan (commented):
Allows any discount percentage regardless of the minimum sale price of the items.
The problem still exists on V17.01 : http://demo.dolibarr.org/public/demo/

PR #20666, opened 2022-04-18
NEW Add an option in GUI to show a Quick add button into top menu bar

aspangaro:
image


altairis-tof:
good idea ! :+1: 
dont forget hooks in quick add list display ;)

PR #20675, opened 2022-04-19
NEW Accountancy - Balance modify target of getNomUrl - go to ledger instead of accounting account admin page

aspangaro:
image

image


eldy:
Currently, everywhere in dolibarr, when we click on a link that show the reference of an object, we reach the page of the file of the object.
Here you would like to switch to a list pre-filtered. 
To not break a generic feature, I suggest to introduce instead, a clickable picto at end of label coming from img_picto('', 'line')   or   img_picot('', 'filter')

aspangaro:
Ok, thanks for your return. 

The logical functioning of dolibarr is different from that of the accounting software. I will play with the pictos to have a similar system

PR #20676, opened 2022-04-20
NEW : add parent product column on list

defrance:
#NEW : if show_child products is enabled, add a column who contains the parent product


htdocs/product/list.php
```diff
@@ -413,8 +415,9 @@
 	$sql .= ' p.fk_unit, cu.label as cu_label,';
 }
 $sql .= ' MIN(pfp.unitprice) as minsellprice';
-if (!empty($conf->variants->enabled) && (!empty($conf->global->PRODUIT_ATTRIBUTES_HIDECHILD) && !$show_childproducts)) {
-	$sql .= ', pac.rowid prod_comb_id';
+if (!empty($conf->variants->enabled) && (!empty($conf->global->PRODUIT_ATTRIBUTES_HIDECHILD) )) {
```
eldy, line 418: The condition here to add column in request is not consistent with the condition where field pac.fk_product_parent is shown into $arrayfields
Do you really need to remove the test on $show_childproducts ?

PR #20677, opened 2022-04-20
NEW : Add support of different update keys on another tables in a import

kkhelifa-opendsi:
NEW : Add support of different update keys on another tables in a import


eldy:
What do you mean with different update keys ?
Can you provide an example of use case ?

eldy:
@kkhelifa-opendsi can you provide an example of this feature ?

eldy:

PR #20684, opened 2022-04-21
fix: Ticket extrafields were not kept during triggers on TICKET_CREAT…

thomas-Ngr:
fix : Ticket extrafields values are not kept when sending an email

~~the $object->fetch() resets the $object->array_options property.
We must save it and replace in the newly fetched object.~~

This is due to extrafields being saved *after* triggers call on Ticket::create() method.

### DIFF
--- a/htdocs/ticket/class/ticket.class.php
+++ b/htdocs/ticket/class/ticket.class.php
@@ -487,15 +487,6 @@ class Ticket extends CommonObject
 
 			if (!$error) {
 				$this->id = $this->db->last_insert_id(MAIN_DB_PREFIX."ticket");
-
-				if (!$notrigger) {
-					// Call trigger
-					$result = $this->call_trigger('TICKET_CREATE', $user);
-					if ($result < 0) {
-						$error++;
-					}
-					// End call triggers
-				}
 			}
 
 			if (!$error && ! empty($conf->global->TICKET_ADD_AUTHOR_AS_CONTACT)) {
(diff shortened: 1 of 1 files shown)


htdocs/core/triggers/interface_50_modTicket_TicketEmail.class.php
```diff
@@ -399,9 +399,11 @@ private function composeAndSendCustomerMessage($sendto, $base_subject, $body, $s
 			$result = $mailfile->sendfile();
 			if ($result) {
 				// update last_msg_sent date
-				$object->fetch($object->id);
+				$save_extrafields = $object->array_options;
```
eldy, line 402: The fetch just after contains the load of the ->array_options so i don't see the need to save it.

thomas-Ngr: The problem source is in Ticket::create() method. In this method, triggers are called (l.493) for ticket creation *before* extrafields values are registered (l.510).

Here, on a ticket creation :
- at line 402, `$object->array_options` is set , but extrafields values are not registered in the DB yet.
- at line 403, calling `$object->fetch($object->id)` replaces  `$object->array_options` by an empty array.
- When returning in Ticket::create() method, `$object->array_options` is empty and extrafields values can't be saved.
- As a consequence, saving and restoring `array_options` property is necessary on ticket creation.

We need  `$object->fetch($object->id)` at line 403 when a ticket is being closed, otherwise the `$object->update()` at line 405 will restore its state to the former state.

thomas-Ngr: Another solution would be in Ticket::create() to move the ->insertExtrafields() call before the ->call_trigger(). What do you prefer ?

eldy: move the ->insertExtrafields() call before the ->call_trigger() !!!
triggers on create must always be called at end. The createCommon() used for modules build with modulebuilder rules (see this method for example on good sequence).

thomas-Ngr: Thanks !

PR #20686, opened 2022-04-22
NEW Option MAIL_MASS_ACTION_ADD_LAST_IF_MAIN_DOC_NOT_FOUND to send last document in mass mailing action

lvessiller-opendsi:
NEW send last document in mass mailing action

- when you want to send by mail several documents (proposals, orders, invoices, etc), you select lines from the related list and click on mailing mass action button :
image

- in this case the included file is only the file whose name match with the object reference and extension is "pdf" (ex : FA2204-0016.pdf)

- with this PR the file to include is :
1. the file whose name match with the object reference and extension is "pdf" (ex : FA2204-0016.pdf) as above
2. if file in "1." is not found, we try to include the file whose path is defined by the "last_main_doc" (the last generated file for the object) property of the object (so we can have others files names such as "Test-FA2204-0016.pdf")

### DIFF
--- a/htdocs/core/actions_massactions.inc.php
+++ b/htdocs/core/actions_massactions.inc.php
@@ -322,7 +322,7 @@ if (!$error && $massaction == 'confirm_presend') {
 						'name' => $filename,
 						'path' => $filepath,
 					);
-					if (!empty($objectobj->last_main_doc)) {
+					if (!empty($conf->global->MAIL_MASS_ACTION_ADD_LAST_IF_MAIN_DOC_NOT_FOUND) && !empty($objectobj->last_main_doc)) {
 						$file_check_list[] = array(
 							'name' => basename($objectobj->last_main_doc),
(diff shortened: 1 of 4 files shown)


eldy:
This may be dangerous because we may send documents that are not the document of invoice: 
Imagine a user has 2 template: crabe to build the main invoice and "doc-for-invoice-collector" (to send a pdf to a partner that need to supervise invoicing just after generating the invoice).

When making a mass action to send documents to all customers, we may send the "doc-for-invoice-collector" that was dedicated for the partner and not to the customer. Right ?
I have no good solution to suggest but may be we can include a hidden option to enable this feature. also note that screen say about the checkbox "Send the main document". if you also send the last generated if the main doc is not available yet, it means the label should be "Send the main document or the last generated one if not found". Right ?

lvessiller-opendsi:
@eldy 
You're right and I add a const to have the main document or the last generated file the main doc is not found.

PR #20691, opened 2022-04-22
Mailing language filter

### LINKED ISSUE #20690: Add lang filter on mailing module
### Feature Request

If you install dolibarr as CRM/More in multilanguage company and multi language customers you need to be able to send mailing and add filters on customer language ...

The first part is #20651 

And now i'm on email tool :)

Could be a duplicate of #12208 ... not really duplicate because #12208 is about multi-categories filter but #12208 is speaking about a filter with language ... so

### Use case

_No response_

### Suggested implementation

_No response_

### Suggested steps

_No response_

rycks:
# FIX|Fix #20690 

you could now choose target language for your mailings


rycks:
sorry i push a branch with 2 fixes :-( i will come back later with better stuff !

PR #20692, opened 2022-04-22
NEW language support for more emailing target selectors

### LINKED ISSUE #20690: Add lang filter on mailing module

rycks:
# FIX|Fix #20690 

Add a language filter for companies and people on Mailing module

For people previous fix #20688  is required


isamuse2:
Works fine with v12.0.4
Please note that with the set up MAIN_LANGUAGES_ALLOWED set to fr_FR,en_GB only French appear (both for contacts and third parties)

PR #20702, opened 2022-04-24
FIX avoid access forbidden when ref is numeric only

hregis:


ozit:
Thanks @hregis , I will wait the @eldy 's review before closing this issue.
I test your modification on my test environment : it works

Arnaud-Dorthe:
It happens also when creating a new bank account with a numerical "Réf.".
Let me please reopen this bug.

PR #20742, opened 2022-04-28
Create a credit note with corresponding serie

leninrivas:
# NEW|New [*Credit note with corresponding serie*]
[*Create a credit note with corresponding serie if use module 2series*]


atm-maxime:
I think you can't do this. 2series is an external module so you can't add code related to this module in Dolibarr core code.

leninrivas:
I understand. Thank :)

PR #20745, opened 2022-04-29
NEW Accountancy - Add filter interval on amount on ledger

aspangaro:
image


josett225:
hi @aspangaro 
After sharing internally and with several accounting managers, this feature is awesome as it helps to track issues and to accelerate the manual lettering in reducing the scope and accelerating when typing. e.g. value debit or credit 1001,98 you can type 1000 and 1002 faster without even remember the values and you get all at once.

eldy:
@josett225 Note that in dolibarr, you can already filter in an interval by using the syntax  "> 1000 < 1002"  into the amount field.

josett225:
@eldy Many thanks for this valuable information as I am discovering it and it works fine.  In the meantime, the idea is to filter in both debit and credit at the same time. So adding either an "or" feature between debit and credit or a unique amount with your interval syntax feature or @aspangaro solution could be great  : to find values in a interval and/or to accelerate lettering.

aspangaro:
@josett225 I close this useless PR, I try another formulation for this feature.

PR #20764, opened 2022-05-02
NEW field required on mandatory fields

lvessiller-opendsi:
NEW field required on mandatory fields

- when you setup a field as required : 
image

- you have the mandatory field in bold (add "fieldrequired" class) :
image

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -8990,6 +8990,9 @@ function printCommonFooter($zone = 'private')
 								print 'jQuery("select[name=\''.$paramkey.'\']").prop(\'required\',true);'."\n";
 								print 'jQuery("select[name=\''.$paramkey.'\'] option[value=\'-1\']").prop(\'value\', \'\');'."\n";
 								print 'jQuery("select[name=\''.$paramkey.'\'] option[value=\'0\']").prop(\'value\', \'\');'."\n";
+
+								// Add 'field required' class on closest td for all input elements : input, textarea and select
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -8990,6 +8990,9 @@ function printCommonFooter($zone = 'private')
 								print 'jQuery("select[name=\''.$paramkey.'\']").prop(\'required\',true);'."\n";
 								print 'jQuery("select[name=\''.$paramkey.'\'] option[value=\'-1\']").prop(\'value\', \'\');'."\n";
 								print 'jQuery("select[name=\''.$paramkey.'\'] option[value=\'0\']").prop(\'value\', \'\');'."\n";
+
+								// Add 'field required' class on closest td for all input elements : input, textarea and select
+								print 'jQuery(":input[name=\'' . $paramkey . '\']").closest("tr").find("td:first").addClass("fieldrequired");' . "\n";
```
eldy, line 8995: Note for info:
To have a more robust solution in future, a rule on syntax has been introduced in v16 for field in form. It is
<tr class="field_xxxx">
<td class="titlefieldcreate">...</td>
<td class="valuefieldcreate">input field</td>

where xxx is the name of the input field. This is supported by page generated with modulebuilder v15 and extrafields in v16. In future, fields will be flagged by this css everywhere...

lvessiller-opendsi: @eldy 
It's seems to be interesting but I don't understant where do you put "xxx input field" ?
In the setup page of mandatory fields ?

PR #20765, opened 2022-05-02
better translation for ToPayOn

altairis-noe:
because « à payer sur XX/XX/XXXX » is not good french, I suggest « à payer le XX/XX/XXXX »


frederic34:
@altairis-noe translation are done in transifex, any change here will be overwritten

https://www.transifex.com/

eldy:
PR on language files are useless, files are erased by transifex sync (except en_US).

PR #20767, opened 2022-05-03
FIX can remove all "module_parts" when module is disabled

hregis:


hregis:
@eldy 
i don't see where is th Travis error !!

htdocs/core/modules/DolibarrModules.class.php
```diff
@@ -2244,26 +2244,17 @@ public function delete_module_parts()
...
-				if (is_array($value) && isset($value['entity'])) {
-					$entity = $value['entity'];
-				}
-
-				$sql = "DELETE FROM ".MAIN_DB_PREFIX."const";
-				$sql .= " WHERE ".$this->db->decrypt('name')." LIKE '".$this->db->escape($this->const_name)."_".strtoupper($key)."'";
-				$sql .= " AND entity = ".((int) $entity);
+		$sql = "DELETE FROM ".MAIN_DB_PREFIX."const";
+		$sql .= " WHERE ".$this->db->decrypt('name')." LIKE '".$this->db->escape($this->const_name)."_%'"; // remove all MAIN_MODULE_MYMODULE_*
+		$sql .= " AND entity IN (0, ".((int) $conf->entity); // remove global and current entity
```
eldy, line 2250: This means when you disable a module in entity 3, you delete conf that may be required by all entities (if they are on 0).
May be that we can add entity 0 only if conf->entity = 1 
So when acting on entity 3, stay on entity 3 only ?

hregis: @eldy 
I corrected, on the other hand if we use "entity" with a value of 0, in general there are modules that we cannot deactivate in the child entities!

hregis:
@eldy 
I don't understand why I have a Travis error on the "init" of the Agenda module!?

atm-maxime:
I think that removing all const related to a module is very dangerous. If I install a module, complete the setup, then reactivate the module, I've lost all my setup...

htdocs/core/modules/DolibarrModules.class.php
```diff
@@ -2246,27 +2246,21 @@ public function delete_module_parts()
...
-				// If entity is defined
-				if (is_array($value) && isset($value['entity'])) {
-					$entity = $value['entity'];
-				}
-
-				$sql = "DELETE FROM ".MAIN_DB_PREFIX."const";
-				$sql .= " WHERE ".$this->db->decrypt('name')." LIKE '".$this->db->escape($this->const_name)."_".strtoupper($key)."'";
-				$sql .= " AND entity = ".((int) $entity);
+		$sql = "DELETE FROM ".MAIN_DB_PREFIX."const";
+		$sql .= " WHERE ".$this->db->decrypt('name')." LIKE '".$this->db->escape($this->const_name)."_%'"; // remove all MAIN_MODULE_MYMODULE_*
```
eldy, line 2251: This means we don't delete only record stored to save module_part array, but all setup of modules ?
The goal to delete only module_parts is to deleted only constants that are used to enable interactions on code but does not delete all constants that contains data setup. So when we re-enable the module setup is still available.

PR #20768, opened 2022-05-03
FIX : fatal error for $db  usage in tpl

john-botella:
in /core/tpl/extrafields_list_search_sql.tpl.php

there is usage of $db method but $db not defined

### DIFF
--- a/htdocs/projet/class/task.class.php
+++ b/htdocs/projet/class/task.class.php
@@ -918,6 +918,7 @@ class Task extends CommonObjectLine
 		// Add where from extra fields
 		$extrafieldsobjectkey = 'projet_task';
 		$extrafieldsobjectprefix = 'efpt.';
+		global $db; // needed for extrafields_list_search_sql.tpl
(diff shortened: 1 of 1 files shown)


htdocs/projet/class/task.class.php
```diff
@@ -918,6 +918,7 @@ public function getTasksArray($usert = null, $userp = null, $projectid = 0, $soc
 		// Add where from extra fields
 		$extrafieldsobjectkey = 'projet_task';
 		$extrafieldsobjectprefix = 'efpt.';
+		global $db; // needed for extrafields_list_search_sql.tpl
```
eldy, line 921: Inside a method of a class the valid database handler is stored inside  $this->db.
Did you try to do
$db = $this->db;      // Because the tpl after is using $db variable.

atm-john: yes but I'm afraid by a $db replacement or strange behavior if we include a php file with global $db;
for me $db is reseved as  global var and I prefert not use it as a local variable to prevent misstake.

In fact I have no preference for this fix, because for me use a  function or a class meto retunring an objet std is more accurate than a tpl like this.

PR #20785, opened 2022-05-05
FIX Accountancy - Add an option to disable use of sub-account on deposit line FPC21+

aspangaro:
If we use thirdparty account (class 4 in France) to stock deposit account, it's ok if you have sub-account but if you work on income account (class 7 in France), it is not advisable to use sub-accounts.

Wrong version : 

image

Good version :

image

To solve problem with all users, Propose to integrate an option in default account admin : 

image


aspangaro:
Who is the problem with stickler, I've no information...

eldy:
Can't you click on "Details" on the stickler line, then "View more details on Stickler CI" ?

PR #20792, opened 2022-05-06
CMailFile class evolution purpose

Libaud:
An evolution of CMailFile class would be interesting and very useful for users.

- First : There's no solution to integrate the same CSS style (in code or external files) on html mail. An additional parameter on company configuration would be the best way. This would be loaded on CMailFile class on Build CSS method.
- Second : The signature for user how expedite the mail doesn't use on HTML mail. Then what can be defined in the record of it.


eldy:
Something seems wrong with the PR (1500+ files modified).
Can you redo it from develop branch ?

hregis:
@Libaud
Can you resend a clean PR please?
I close this...

PR #20798, opened 2022-05-06
FIX : Trigger on commonObject setProject

atm-kevin:
Add trigger in commonObject setProject function


atm-kevin:
@eldy how should i do to resolve this Travis issue ?
I don't understand why the changes I made would have an impact on the file WebservicesProductsTest.php.

eldy:
In the log of travis pass, we've got this
Warning: Declaration of Ticket::setProject($projectid) should be compatible with CommonObject::setProject($projectid, $notrigger = 0) in /home/travis/build/Dolibarr/dolibarr/htdocs/ticket/class/ticket.class.php on line <i>2023</i>

So you must :  
Solution 1:  update the signature of method  Ticket::setProject
Solution 2: completely remove the method  Ticket::setProject  because ticket inherit of CommonObject so method is already defined into parent.

Solution 2 is better.

atm-kevin:
Thank you @eldy, i finally applied the second solution.

PR #20802, opened 2022-05-07
New supplier invoice model : just add a stamp on PDF

rycks:
# NEW : add a stamp on supplier invoice PDF

That new model add only a stamp on supplier PDF with internal dolibarr reference (for example "SI2205-0075") : so you don't have to take your pen and write that ref on paper when you make your paperwork (save time on accounting job).

Example: here is a screenshot of supplier invoice pushed in dolibarr

image

And here is the result with paprika model

image

Maybe you can find more informations on french forum : https://www.dolibarr.fr/forum/t/facture-fournisseur-compta-numerotation-perte-de-temps-et-automatisation/40032

### DIFF
--- a/htdocs/core/modules/supplier_invoice/doc/pdf_paprika.modules.php
+++ b/htdocs/core/modules/supplier_invoice/doc/pdf_paprika.modules.php
@@ -292,10 +292,10 @@ class pdf_paprika extends ModelePDFSuppliersInvoices
 				foreach (glob($dir."/".$objectref."-*.pdf") as $otherpdf) {
 					$otherpdfindir = $otherpdf;
 				}
-                if (null !== $otherpdfindir) {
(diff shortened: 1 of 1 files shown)


eldy:
I don't understand.
Where is coming the content of the PDF. IF i upload 3 docs linked to the purchase invoice. When i choose paprika, what are the PDF used as template for the content ?

rycks:
forget that PR :-)

daraelmin:
> I don't understand.
> Where is coming the content of the PDF. IF i upload 3 docs linked to the purchase invoice. When i choose paprika, what are the PDF used as template for the content ?

I think it is linked to the external module scaninvoice

rycks:
forget that PR, it's embedded into ScanInvoices free plugin

PR #20807, opened 2022-05-09
line deleted when updated after rapproched

mapiolca:
Fix an issue that deletes line when we update data after its rappro.


mapiolca:
Yes, you just need to uncheck the box, Apply , change your data, and re-check the box, and apply

PR #20811, opened 2022-05-09
add hidden option SUPPLIER_ORDER_ALLOW_NEGATIVE_QTY_FOR_SUPPLIER_ORDE…

ibuiv:
# NEW add hidden option SUPPLIER_ORDER_ALLOW_NEGATIVE_QTY_FOR_SUPPLIER_ORDER_RETURN

Consider you can send via order supplier (for return or retrocession)
If this option is activated, the StockMouvement method called will be livraison() instead of reception() if the qty < 0
It will help you to manage expeditions (the qty will be automatically set to all instead of 0 if qty < 0)


ibuiv:
@eldy : As asked, I ask for a new pull request (refers to PR #17360)

PR #20812, opened 2022-05-10
FIX: formconfirm: pass action even when confirm is no

creekorful:
Obviously, one may question whether this is really a bugfix or a new feature (behavior change), but I'm kinda confident that this is not a breaking change for people who were relying on confirm=no without the action.

This change is really helpful if you want to handle the refusal of a specific action and do specific treatment later on.


eldy:
To be sure and take no risk and because this is not to fix a visible bug, i merge this into develop for v16

PR #20819, opened 2022-05-10
FIX : user creation from admin user

atm-adrien:
### NEW : User creation with multicompany 

If multicompany and the config MULTICOMPANY_TRANSVERSE_MODE enabled, when an admin user tries to create another user, the admin user can't access to created user card. 

There was an error message just after the creation, when the redirection / user tries to access to the created user card.

### DIFF
--- a/htdocs/core/lib/security.lib.php
+++ b/htdocs/core/lib/security.lib.php
@@ -644,7 +644,7 @@ function checkUserAccessToObject($user, array $featuresarray, $objectid = 0, $ta
 						$sql .= " WHERE dbt.".$dbt_select." IN (".$db->sanitize($objectid, 1).")";
 						$sql .= " AND ((ug.fk_user = dbt.rowid";
 						$sql .= " AND ug.entity IN (".getEntity('usergroup')."))";
-						$sql .= " OR dbt.entity = 0)"; // Show always superadmin
+						$sql .= " OR dbt.entity IN (0, ".$conf->entity."))"; // Show always superadmin
 					}
 				} else {
 					$sql .= " WHERE dbt.".$dbt_select." IN (".$db->sanitize($objectid, 1).")";


hregis:
@atm-adrien
If MULTICOMPANY_TRANSVERSE_MODE is enabled, only the superadmin can create a user in entity 1

hregis:
@atm-adrien can you explain the context of your problem?

atm-adrien:
Thank you for your prompt response ! 


An admin (non-superadmin) user attempted to create a new user. The creator user, with creation rights, succeeded in creating the new user, but could not access his card or see him in the list.

I think that there are 2 possibles solutions : 
- Make admin user able to create and see the other users on the main entity 
- Remove user create form access for admin user and authorize it only for superAdmin

hregis:
@atm-adrien
With `MULTICOMPANY_TRANSVERSE_MODE` a user must be associated with a group and one or more entities to define which entities he belongs to. This is why it is necessary to limit the creation of users to superadmins.

atm-adrien:
Ok, so in this case, the best option is to "Remove user create form access for admin user and authorize it only for superAdmin" ?

eldy:
I agree with @atm-adrien , there is a problem with MULTICOMPANY_TRANSVERSE_MODE (in fact there is a lof of trouble with this option). A user is created and just after creation the user disappear, even for the user that has created it, because the created user is linked to no group. This is not the only problem with this mode. This is just one example.
The best solution i see to solve this choking situation (at least the case of user not visible just after creation) is to add protection for MULTICOMPANY_TRANSVERSE_MODE so when we create the user we must also automatically add it into at least one group (for example a groupe "Everyone",  even if the group has no permissions at all).

hregis:
@atm-adrien @atm-adrien

No, because not only must it be associated with a group, but also with one or more entities! And associating it with entity 1 by default is not recommended because this user may not have the rights to access this entity! When user management is centralized on the main entity, only a superadmin can create a user and assign him access rights to entities.

hregis:
@atm-adrien @eldy

A simple admin can modify the configuration of the entities to which he has access, but he cannot see all the groups or all the entities if he himself is limited! You understand ?

atm-adrien:
@eldy

eldy:
> @atm-adrien @eldy
> 
> A simple admin can modify the configuration of the entities to which he has access, but he cannot see all the groups or all the entities if he himself is limited! You understand ?

This is not the point.
If a user X create a user Y, the user Y must be visible for the user X just after the creation step !!! If not this is a bug. And currently this bug is confirmed. However, the way it is fixed is wrong.

hregis:
@eldy @atm-adrien 
in a mode where each entity manages its own users there is no problem, each entity reacts like a simple Dolibarr.
In a mode where the main entity manages users (manages access), a link is made with a group of users and one or more entities to which users have the right to access... It would be difficult for a " simple user" or a "simple admin" who does not have access to everything to be able to decide on the access rights of a "simple user"...

hregis:
@atm-adrien @eldy 

the superadmin was originally created for this...

hregis:
@eldy 
keep-calm-i-am-superadmin

atm-adrien:
> in a mode where each entity manages its own users there is no problem, each entity reacts like a simple Dolibarr.
> In a mode where the main entity manages users (manages access), a link is made with a group of users and one or more entities to which users have the right to access... It would be difficult for a " simple user" or a "simple admin" who does not have access to everything to be able to decide on the access rights of a "simple user"...

@eldy @hregis 

Yes, but today, a simple admin (who can't manage user groups and access to entities) has the capacity of creating user , bu he can't see them just after the creation. I think it's a problem.
In fact, an user creator must be able to see its created users, at least just after their creation.

I think it is necessary to make a decision on this subject : 
- Make admin user able to create and see the other users on the main entity (@hregis is against this one)
- Remove user creation rights for admin user and authorize it only for superAdmin (the best solution in my point of view)

hregis:
@atm-adrien
It's ready in the last commit in v15 branch (and develop)
Just a superadmin can create users and groups in entity 1 when users/groups management is centralized in entity 1.

eldy:
According to message of Regis, a protection was added, making this PR obsolete so i close it.

PR #20833, opened 2022-05-11
Fix missing array managements for form setup class

john-botella:
fix some missing elements

@eldy this fix can be integrate in V15 ?


eldy:
No, only bugfix are qualified for v15 (features not completely implemented are not bugfixes)

PR #20866, opened 2022-05-12
NEW create third-party with contact if not found on pulic ticket

lvessiller-opendsi:
NEW create third-party with contact if not found on pulic ticket

1. Setup
- add a parameter on ticket setup (public interface tab) : 
image

When this parameter "Create third-party with contact if not found from given email" is enable, the above parameter "Email must be linked to a contact to create a ticket" is disabled and "captcha" security is also enabled to avoid to have unwanted insertions into database from bots.

2. On public interface, this above parameter will add some contatc and company fields if no contact found from given email : 
image

- contact lastname (required)
- contact firstname (required)
- company name (not required)
- contact phone (not required)

And when you create the ticket, a third-party is created whose name is company name if put in this form or contact lastname + contact firstname if company name is not put in this form.
And a contact is created linked to this company and with email, lastname, firstname and phone put in this form.
And the created ticket is linked to this new company and this new contact.

3. On public interface, this above parameter will show the contact found from given email : 
image

We keep the same behavior for the ticket creation as the one used until now


lvessiller-opendsi:
@eldy 
I got a travis error but it not seems not to be occured by this feature.
Could you check ?

PR #20880, opened 2022-05-17
Fix #20878 : improper use of dol_move()

### LINKED ISSUE #20878: Improper use of dol_move() to rename Directory; should be @rename()
### Bug

In file /projet/class/task.class.php within public function update() the wrong function is called at line 482 in V15.0.1.

Line 482 (V15.0.1) should be changed from ...
$res = dol_move($olddir, $newdir);
to become ...
$res = @rename($olddir, $newdir);

image
 

### Environment Version

15.0.1 and earlier

### Environment OS

Linux

### Environment Web server

Apache

### Environment PHP

7.4.29

### Environment Database

Mysql 8.0.23

### Environment URL(s)

mydomain/projet/tasks/task.php

When modifying a Task (action= update) renaming the Ref of the Task requires a corresponding change of the directory name where the generated documents are stored on the server.

Instead, when changing the Ref of the Task an error is returned saying "Failed to rename directory".

image
(Note: in the image you will see that we have changed the necessary language file so that "Tasks" are called "Work-Orders" for our car-restoration application. Our language file change has no relevance to the error.)

1. Create a Task with a Ref. 
2. Generate a PDF document for that Task to ensure that a sub-directory is created for that Task in the corresponding Project Directory.
3. Try to rename the Ref and the error message is returned.

### Attached files

_No response_

Hystepik:
Fix rename of folder when changing task ref + add file rename process

### DIFF
--- a/htdocs/projet/class/task.class.php
+++ b/htdocs/projet/class/task.class.php
@@ -470,7 +470,7 @@ class Task extends CommonObjectLine
 		}
 
 		if (!$error && (is_object($this->oldcopy) && $this->oldcopy->ref !== $this->ref)) {
-			// We remove directory
+			// We rename directory
 			if ($conf->projet->dir_output) {
 				$project = new Project($this->db);
 				$project->fetch($this->fk_project);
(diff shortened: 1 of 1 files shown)


htdocs/projet/class/task.class.php
```diff
@@ -479,13 +479,36 @@ function ($allTasksCompleted, $task) {
 				$newdir = $conf->projet->dir_output.'/'.dol_sanitizeFileName($project->ref).'/'.dol_sanitizeFileName($this->ref);
 				if (file_exists($olddir)) {
 					include_once DOL_DOCUMENT_ROOT.'/core/lib/files.lib.php';
-					$res = dol_move($olddir, $newdir);
+					$res = @rename($olddir, $newdir);
```
eldy, line 482: Why replacing dol_move with rename ?

Hystepik: Because the sourcefile of dol_move() cannot be a directory
@param string $srcfile Source file (can't be a directory. use native php @rename() to move a directory)

eldy: Thanks to introduce a dol_move_dir() including renaming of ref.

PR #20932, opened 2022-05-20
 fix #20817 : user->id moved to user->contact_id

### LINKED ISSUE #20817: socpeople and $user->id is not the same
### Bug

It seems to be a bug, for example in project.class.php

```
			$sql .= " AND ( p.public = 1";
			$sql .= " OR ( ec.fk_c_type_contact IN (".$this->db->sanitize(join(',', array_keys($listofprojectcontacttype))).")";
			$sql .= " AND ec.fk_socpeople = ".((int) $user->id).")";
			$sql .= " )";
```

and the problem is that i have an external user with user->id=22 and there is no ec.fk_socpeople=22 because his socpeople id is 12 !

image

image

That "bug" is present in some files, for the moment i've seen it in :

- project.class.php
- box_tasks.php
- project.lib.php
- task.class.php

### Environment Version

_No response_

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

rycks:
# FIX #20817 : user->id moved to user->contact_id

Some SQL requests use user->id as target instead of user->contact_id

### DIFF
--- a/htdocs/core/boxes/box_task.php
+++ b/htdocs/core/boxes/box_task.php
@@ -154,10 +154,10 @@ class box_task extends ModeleBoxes
 			$sql .= " JOIN ".MAIN_DB_PREFIX."projet as p ON (pt.fk_projet = p.rowid)";
 
 			if ($filterValue === 'im_task_contact') {
-				$sql .= " JOIN ".MAIN_DB_PREFIX."element_contact as ec ON (ec.element_id = pt.rowid AND ec.fk_socpeople = ".((int) $user->id).")";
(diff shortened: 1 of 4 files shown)


eldy:
The name fk_socpeople is not a good choice of name because sometimes it store user id,sometime contact id. (The obect $user is sometime a User sometimes a Contact)
So this add more problem than it fixes the bug.

PR #20942, opened 2022-05-21
FIX gmstrftime for PHP 8.1

inoveaconseil:


htdocs/includes/adodbtime/adodb-time.inc.php
```diff
@@ -1361,7 +1361,7 @@ function adodb_strftime($fmt, $ts=false,$is_gmt=false)
 	if (!defined('ADODB_TEST_DATES')) {
 		if ((abs($ts) <= 0x7FFFFFFF)) { // check if number in 32-bit signed range
 			if (!defined('ADODB_NO_NEGATIVE_TS') || $ts >= 0) // if windows, must be +ve integer
-				return ($is_gmt)? @gmstrftime($fmt,$ts): @strftime($fmt,$ts);
+				return ($is_gmt)? @gmdate($fmt,$ts): @date($fmt,$ts);
```
eldy, line 1364: $fmt  is  for example    "%d %m %y"
Format is https://www.php.net/manual/fr/function.strftime.php
But format for gmdate is
Format is https://www.php.net/manual/fr/datetime.format.php
So changing gmstrftime into gmdate is not enough

eldy: We need to support this tags
%Y %m %d %H %M %S
but also %a %A %b %B

Also what happen if the string contains a Z at end like with this format
"%Y%m%d%H%M%SZ"

PR #20957, opened 2022-05-21
fix : The property chid does not seem to exist on PaymentSocialContri…

grandoc:
…bution


htdocs/compta/paiement_charge.php
```diff
@@ -291,7 +291,7 @@
 		print "</tr>\n";
 		$total += $objp->total;
 		$total_ttc += $objp->total_ttc;
-		$totalrecu += $objp->am;
+		$totalrecu += $objp->amount;
```
eldy, line 294: The property "amount" seems to not exists into $objp

PR #20960, opened 2022-05-21
Close #20930 NEW Use ajax for state loading after country change

### LINKED ISSUE #20930: User card : state not related to country
### Bug

While creating or editing a user card, the state dropdown field is not refreshed after selecting a country. All states from all countries are available.
On thirdparty card there's a refresh after country selection

### Environment Version

at least 13.0

### Environment OS

all

### Environment Web server

all

### Environment PHP

all

### Environment Database

all

### Environment URL(s)

http://dolibarr.local/user/card.php?leftmenu=users&action=create

The state dropdown should refresh to show only states related to the selected country

Go on the user creation form, select a country in the list, check the state dropdown content

### Attached files

image

homer8173:
combo country / state


homer8173:
isn't it time to implement it ?

htdocs/core/ajax/ziptown.php
```diff
@@ -141,6 +141,9 @@
 	}
 
 	echo json_encode($return_arr);
+} elseif (isset($_GET['country_codeid'])) {
```
eldy, line 144: Must use here GETPOSTISSET

PR #20966, opened 2022-05-21
fix price2numjs in case of very small amount

SylvainLegrand:


htdocs/core/js/lib_head.js.php
```diff
@@ -1128,6 +1128,7 @@ function price2numjs(amount) {
 	amount = amount.replace(dec, '.');
 	//console.log("amount before="+amount+" rouding="+rounding)
 	var res = Math.round10(amount, - rounding);
+	if (res == 0 && amount > 0)	res = amount;	// in case of very small amount
```
eldy, line 1131: Do you have an example of "small" value that is wrong ? And what is the value of rounding.

PR #20968, opened 2022-05-21
New : function dol_move_dir in htdocs/core/lib/files.lib.php

Hystepik:
Move a directory into another name
when merged change dol_move() in htdocs/projet/class/task.class.php line 482 to fix #20880


eldy:
Please provide also a phpunit test to test the method for different cases.

PR #20972, opened 2022-05-21
New:Proposition of Object Oriented Scripts Structure applied to propales

bafbes:
New:Proposition of Object Oriented Scripts Structure applied to propales


bafbes:
There is an associated module in https://github.com/bafbes/cardmodifdemo  that shows how to use this new structrure more easily to perform a simple action that is impossible to do with the present structure of card pages unless we do the modification through JS. 

**Other Comments :**
1.Next to /comm/propal/card.php page which has been greatly modified, there is also 
1.1 abstractions.class.php which defines the abstract class AbstractActions class that should be used as a parent class for a class Action and that is contains the previous contents of /comm/propal/card.php and its included files.
1.2 security.class.php that contains the previous functions of security.lib.php as methods of a new class Security so that we may change the file associated to this class int the init hook if we need to change functionalities previously defined in security.lib.php

2.This approach can be continued to all other card and list files in Dolibarr by replacing their code as in the case of propal cards and by replacing all lib files by their equivalent class files. 

3.Next to the advantages of full and easy functionality control by modules, this method of script construction is much clearer as scripts are much shorter and clearer than the originals. Furthermore, functionalities in the new created classes can later be embedded in other classes in the aim to minimize code quality and volume.

bafbes:
**New comments :**
1.The new hook "init" serves to redefine all needed files included in the script and the definition of all objects used giving control to change all the code of the script except the code included by the call to main.inc.php
2.In the exposed code, we can notice that there are two types of classes:
 2.1. Classes created from lib files: These are classes for which a single method will generally be used within the script. To change the code of the called function, we only need to redefine the file containing the class within the init hook by redefining the class and the called method. When the method called calls an other functions in the same file, the call to the function can be replaced by that of a derived class which  overloads only the function called initially. The latter will then call the second function in the parent class. Note that the inheritance mechanism for the new Security class has not been fully implemented in the PR. This work needs to be done later in another PR.
 2.2. Classes derived from abstract classes: These are classes exporting several functions within the same script. They were derived from abstract classes to be able to override the contents of any method in the derived class without having to redeclare all the functions of the class. This inheritance mechanism can also be improved to be even more efficient.

eldy:
This PR contains several different suggestion of refactoring.
This PR must be split into different PR with one type of suggestion in one PR so it will be possible to discuss for the pro and cons of each suggestion change. It will also allow to validate them step by step reducing regression risks.

bafbes:
Il seems that the example module link was erroneous. I just fixed it to https://github.com/bafbes/cardmodifdemo.  All this refactorings were necessary to get the propal card page overloading a method to do the work. That is, prohibit modifying the associated project if the propal is not in draft state and with no lines inside.

Without these improvements, we would need to whether redirect the page through JS or integrate the change directly in Dolibarr core.

Anyway, the more important things for this goal are :

1. The hook "init" that controls which files are loaded by the page and the declaration of the global objects.
2. The rest of the page must consist only of methods of already declared objects. This permits to overload this methods through the call to the init hook.

Furthermore, objects whose methods are used in many points in the page need to be derived from an abstract class by overloading only the method that is changed by the module and avoiding the redeclaration of other methods, while other objects are replaced by new ones with the redeclaration of the only used method or exceptionally methods.

eldy:
You say "Il seems that the example module link was erroneous.". So you mean there is a bug. Can your report it as an issue ?

Also for change of architecture, the new target is to use modulebuilder not enhance the old structure used by propal/card.php. So please suggest, your enhancement of architecture into the template /modulebuilder/template/
Backporting the reference /modulebuilder/template/ on old pages is another task once the template is validated.

bafbes:
It is not an issue, There were just a little error that prevented the module to work that I fixed immediately before posting the message.

PR #20988, opened 2022-05-22
fix : The property typent_code does not seem to exist on Contact

grandoc:


htdocs/contact/class/contact.class.php
```diff
@@ -162,6 +162,11 @@ class Contact extends CommonObject
 	 */
 	public $state_code;
 
+	/**
+	 * @var string // Code of typent
+	 */
+	public $typent_code;
+
```
eldy, line 169: What is a "typent_code" on a contact ?

eldy:
Pr closed, missing answer for qualification

PR #20992, opened 2022-05-23
NEW : Display Datedeliveryplanned and aviability in same time

defrance:
NEW 
on proposal Datedeliveryplanned and aviability are two differents notion
we need sometime to display it the same time


eldy:
Search/filter on categories must now be implemented using the sql EXISTS and NOT EXISTS (so we can search on several categories, search on no category defined, and filter can be added into the WHERE only with no impact on the select and the FROM, so this is compatible with advanced filter tool.
See examples into projet/list.php

eldy:
Failed to solve conflict so i close ticket.

PR #20999, opened 2022-05-23
NEW : add coefficient on unit  ht   line  in expenseReport card

atm-jpb:
# NEW| add coef on input ht line expenseReport

In expense report card. 
when choosing the type of mileage expense report, automatically add the coefficient relating to the qty
it 'will provide  the right calculation on add action via object
to discuss (this version will not take offset on calculation )

### DIFF
--- a/htdocs/expensereport/card.php
+++ b/htdocs/expensereport/card.php
@@ -2604,7 +2604,7 @@ if ($action == 'create') {
                     let tax_cat = jQuery("#select_fk_c_exp_tax_cat").find(":selected").val();
                     let tva = jQuery("#vatrate").find(":selected").val();
                     let qty = jQuery("#input_qty").val();
-                    console.log(qty);
+
 
 
 					let path = "'.dol_buildpath("/expensereport/ajax/ajaxik.php", 1) .'";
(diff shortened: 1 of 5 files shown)


john-botella:
Set as draft instead of WIP

htdocs/expensereport/card.php
```diff
@@ -2559,7 +2559,7 @@
 
 				// Quantity
 				print '<td class="right inputqty">';
-				print '<input type="text" min="0" class="right maxwidth50" name="qty" value="'.dol_escape_htmltag($qty ? $qty : 1).'">'; // We must be able to enter decimal qty
+				print '<input type="text" min="0" class="right maxwidth50" id="input_qty" name="qty" value="'.dol_escape_htmltag($qty ? $qty : 1).'">'; // We must be able to enter decimal qty
```
eldy, line 2562: A field with id="input_qty" has already be added line 2300.
an html id must be used only once per page.

atm-jpb: maybe i'm wrong but this two ids declaration never appear at the same moment.
$action != 'editline'  in one case and $action == 'editline' for the other

atm-jpb: i made the changes on my last commit.

PR #21010, opened 2022-05-24
New :add explicit right on read draft leave

atm-jpb:
# New  
creation of a right to explicitly read leave requests with draft status

(which could give rise to a fine adjustment : 
I can see the requests and subordinate to the status validate and/or draft)


atm-maxime:
I can't see the use case where this new permission is needed. If I have permission to see leave, I can see them no matter what status...

eldy:
The draft status is not really a status.
It here just to say that object does not exist yet, until validated.

We must not have any permission to rule the visibility of draft status. Permission rule must be the same than the one for a validated object. Note: do not make a confustion between validated holiday that just mean, i finished to enter data on my holiday and approved holiday (meaning it has been validated by the approver) that is a different status.

Issue #20272, opened 2022-03-04
Accountancy - Create multiples transaction in one screen (each in one line)


daraelmin:
### Feature Request

Currently, you have to go through 3 screens to add an accounting line to the journal:

1. Create the document in the journal - htdocs/accountancy/bookkeeping/card.php?action=create
2. Create debit account movement (Cash) - htdocs/accountancy/bookkeeping/card.php
3. Create the movement of the credit account (Bank) - htdocs/accountancy/bookkeeping/card.php

Normally, when the accountant embarks on the logging of several operations in the accounting journal, he should be able to do everything on one line, ideally by adding this one just after the visible ones.

### Use case

In most case, The accountant arrives to record the depreciation of furniture, computer equipment, real estate, etc.

It goes to the global journal page, which looks like a table.
He adds a line (directly on the journal) noting the item number, the specific journal concerned (other, bank, etc.), indicates the debit account number, the credit account number, (ev. the account auxiliary), the label, the debit sum, the credit sum.

It adds a line and does the same for the next operation.

When the same coin or transaction is allocated to several accounts, he has the option of adding an empty debit or credit account (or dash) with an empty sum or dash.

He validates. If the balance is good, everything is recorded, otherwise it must read again to balance correctly.

### Suggested implementation

The idea being to present this in a readable form at first glance and to avoid duplicate lines with just the account and the amount that changes.

Here is an example of this type of simplified presentation (the box highlighted in blue allows the addition of a new line starting from the date) :

![image](https://user-images.githubusercontent.com/8067223/156781896-e8fc2466-b98d-4136-ac53-de666ae35837.png)

### Suggested steps

1. Consolidate all transaction registration screens into a single screen.
2. Add a create multiple transaction action on the page
3. With Adapt the presentation page of the global journal (and why not specific journals) so that it can via javascript add lines then send them in bulk.

daraelmin:
Still working on this one

PR #21016, opened 2022-05-24
New Accountancy - transcation creation and mouvement in one single screen

daraelmin:
# NEW Accountancy - transaction creation and movement in one single screen

First step of #20272 - Consolidate all transaction registration screens into a single screen
When pressing on "+" in bookkeeping/list.php 
We can register the whole transaction.
By default, the piece name is the next rowid and the Journal by default is Other.
Hope I'm on the right way


daraelmin:
Hi @aspangaro,
Could you give a try and tell me what you think about this PR ?
Cheers

eldy:
What do you mean in creation of transaction in one single screen. For me the page to create/edit a transaction is already on 1 page ?

daraelmin:
Hi @eldy,

Actually, when clicking on the "+" button on general ledger you go to this screen 



Then you click save and you go to this screen :



Then you click add to add the second line to balance account (this is treated in my second PR).


So, if you have many mouvement it can be long. 

With this PR, you click "+" button and you can register your mouvement and first transaction on one screen.




The next step is to have both credit and debit transaction ready to be stored (witch is treated in my next PR)

daraelmin:
But, we changed the bookkeeping/card.php since I submitted my PR. So it may be better if I close this one, update it and combine it with the second.

I'll try to tomorrow

daraelmin:
These changes are integrated here : #21217

PR #21034, opened 2022-05-25
NEW shipment kits with dispatcher (SQL)

lvessiller-opendsi:
NEW shipment kits with dispatcher (SQL). Code following is #21088
- sql install


eldy:
Can you describe the goal of batch (provide example of data) ?

lvessiller-opendsi:
@eldy 
You can see more information in PR #21088

lvessiller-opendsi:
@eldy 
All conflicts resolved, could you merge it ?

lvessiller-opendsi:
@eldy 
I don't understand travis error.
Could you check and tell me what's wrong ?

lvessiller-opendsi:
@eldy
I don't understand travis error.
Could you check and tell me what's wrong ?

eldy:
I did a comparison with how the reception works. Add a new table will make things too different than reception.

For reception, detail per batch/warehouse is stored into the table llx_commande_fournisseur_dispatch.sql (that should have been named llx_receptiondet_batch).

So for shipment, we should also store detail of couple batch/warehouse into 1 table, so the existing llx_expeditiondet_batch.   

This means we should instead add the field fk_warehouse into llx_expeditiondet_batch (to replace fk_origin_stock that is not enough accurate/appropriate for the need)

eldy:

lvessiller-opendsi:
@eldy 
I think you refer to this PR : https://github.com/Dolibarr/dolibarr/pull/24717, is it right ?
I tried it and I see only a new "dispatch" tab on shipment card.
You can access to this "dispatch" tab only when the shipment was created from an order.
So it's to update the shipment and not to create.

The shipment lines has already been dispatched when you click the "Shipment" button from order card : 
image

So you have to dispatch lines before with the old functionality.
And then you can update dispatched lines with the new functionality given by the PR 24717.
So I don't see the way to dispatch virtual products like I implemented in this PR.

eldy:
The dispatching (reception and shipment) is now managed by a dedicated tab into shipment and reception object.
So if a feature to make dispatching of virtual product must be implemented, it should be into this new tab.

htdocs/install/mysql/tables/llx_product_association.sql
```diff
@@ -23,7 +23,6 @@ create table llx_product_association
   fk_product_pere       integer NOT NULL DEFAULT 0, -- id du produit maitre
   fk_product_fils       integer NOT NULL DEFAULT 0, -- id du sous-produit
   qty                   double NULL,
-  incdec                integer DEFAULT 1, -- when set to 1 changing stock of product will change stock of linked product too
```
eldy, line 26: Why removing this. It breaks a future to allow user to decide if a line of a kit must have stock impacted if not (default is 1 but some use case need 0 for example when it contains a produt we don't want to follow stock)

lvessiller-opendsi: According to me,
in a virtual product the parent is never incremented or decremented, it's only on children you can increment or decrement. You increment only children on reception and you sell a virtual product that contain some others products and you decrement only children and not the virtual product.

eldy: > According to me, in a virtual product the parent is never incremented or decremented, it's only on children you can increment or decrement. You increment only children on reception and you sell a virtual product that contain some others products and you decrement only children and not the virtual product.

Yes, And this field incdec is necessary to know if we increment or not the children products. Currently we can't set the property "must be stock managed" at product level. So we can use this to have virtual products that contains both children product that need to be increment or decremented and other that won't change (because we don't want to follow consumption for this product).

eldy:
PR seems replaced with https://github.com/Dolibarr/dolibarr/pull/26519 so i close this one.

PR #21038, opened 2022-05-25
Update members.lang

nealjoos:


daraelmin:
Translation should be modified with transifex (https://www.transifex.com/dolibarr-association/) not in github.

nealjoos:
Oh sorry, I submitted a membership application for the Transifex team.

PR #21051, opened 2022-05-29
Fix API knowledgemanagement

ptibogxiv:
@eldy i don't find where i can fix the second error with "/api/index.php/categories/13/objects?type=knowledgemanagement"
i try  update all const but not functionnal...
image


ptibogxiv:
@eldy yes the table exist on my Dolibarr

PR #21068, opened 2022-05-31
FIX remove special product lines on discount create

lvessiller-opendsi:
FIX remove special product lines on discount create
- avoid from creating a discount (got from invoice deposit payment) when we have lines with a special product
- otherwise amounts were added twice (one time for special product line and other time in components lines)

### DIFF
--- a/htdocs/compta/paiement/class/paiement.class.php
+++ b/htdocs/compta/paiement/class/paiement.class.php
@@ -354,7 +354,7 @@ class Paiement extends CommonObject
 										// Loop on each vat rate
 										$i = 0;
 										foreach ($invoice->lines as $line) {
-											if ($line->total_ht != 0) {    // no need to create discount if amount is null
+											if ($line->product_type != 9 && $line->total_ht != 0) {    // no need to create discount if special product or amount is null
 												$amount_ht[$line->tva_tx] += $line->total_ht;
 												$amount_tva[$line->tva_tx] += $line->total_tva;
 												$amount_ttc[$line->tva_tx] += $line->total_ttc;


htdocs/compta/paiement/class/paiement.class.php
```diff
@@ -354,7 +354,7 @@ public function create($user, $closepaidinvoices = 0, $thirdparty = null)
 										// Loop on each vat rate
 										$i = 0;
 										foreach ($invoice->lines as $line) {
-											if ($line->total_ht != 0) {    // no need to create discount if amount is null
+											if ($line->product_type != 9 && $line->total_ht != 0) {    // no need to create discount if special product or amount is null
```
eldy, line 357: Do you mean you have a line with
product_type = 9 and also total_ht that is not 0.

We can use product_type to identify special line, but we should keep the rule "total of lines of invoices in database must match total of invoice in database".  So if a special line exists and must not be included into total, a better practice would be to have total to 0 for such line;

Can you describe the use case / module that add such a line ?

lvessiller-opendsi: Yes, I have a line with "product_type" = 9 and "total_ht" is not 0.
For example with the "Ouvrage" module, you can create a deposit invoice with these lines : 
image

So when you put a payment, the amount for third-party discount is double counted
- one time in the first line (work line)
- and an other time in others lines (components lines of the work line)

eldy:
This looks like a bug into "Ouvrage" that is corrupting the database.
If a line is a line of type 9, and must not be included into total of invoice, the total_xxx in line must be 0 into database. Amount should be calculated and output dynamically on screen if required.  See example on module "Sub-total" that add a line with type 9 with the subtotal, but amount remain 0 so total in llx_facturedet remains consistent with total in llx_facture.
"Data consistency" has priority on "Ease of development"

lvessiller-opendsi:
@eldy 
I understand your point of view for "Data consistency" but "Ouvrage" and "SubTotal" are two differents modules with differents functionalities. 
In case of "SubTotal", you don't need to have Total for its differents levels of titles.
In case of "Ouvrage", you have lines with total for parent line and total of each components lines. All amounts are needed not only total HT and it will reduce performances to compute dynamically all these amounts.
How can I do in this case ? What do you suggest ?
Maybe, do you prefer having an another "product_type" value or check the "special_code" value ?

eldy:
> In case of "Ouvrage", you have lines with total for parent line and total of each components lines. All amounts are needed not only total HT and it will reduce performances to compute dynamically all these amounts.

Sure, but data consistency has also priority on performance. using another product_type or using special_code is not the point. The point is that total in llx_facture must always match the total in llx_facturedet.
If you need to write virtual lines for your module and you also nee to store an amount, on this virtual line that will break this rule, it means you must not store your amount into the common field but into another field. For example, an extrafield.

inoveaconseil:
@eldy 
I'm surprised at the answer, if the possibility of creating lines without taking into account the total (product_type=9). Consistency should be full on all Dolibarr features whether the total_ht is filled or not.
In addition, the PR does not disturb the discount system.

eldy:
> @eldy 
> I'm surprised at the answer, if the possibility of creating lines without taking into account the total (product_type=9). Consistency should be full on all Dolibarr features whether the total_ht is filled or not.
> In addition, the PR does not disturb the discount system.

Sure, consistency of data should be everywhere. If there is an exception somewhere, we may propagate the exception everywhere so exception become "a rule" and no more "an exception". 
Currently test on product_type != 9 is only used everywhere to force fk_parent_line to 0.
And it seems that, currently, total of lines match total of object so i see no reason to introduce regression to break this good thing.

PR #21070, opened 2022-05-31
FIX : Classify the reception

atm-adrien:
## FIX : Classify the reception

Classify the reception "All products received" when all products have been received

### DIFF
--- a/htdocs/reception/class/reception.class.php
+++ b/htdocs/reception/class/reception.class.php
@@ -633,8 +633,25 @@ class Reception extends CommonObject
 			}
 		}
 
-		// Change status of order to "reception in process"
-		$ret = $this->setStatut(4, $this->origin_id, 'commande_fournisseur');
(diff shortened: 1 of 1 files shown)


eldy:
The feature seems to be already implemented into last branch.
Can you describe why the bug is blocking the user ?

atm-adrien:
Yes, the feature is already implemented into last branch.
But in version 14.0, this bug is problematic. Indeed, when we receive all the products of a supplier order thank in a reception, the status of the supplier order does not change. So the user think that the order hasn't been fully receipt.

I think it is a problem, and this backport correct this.

eldy:
This does not look to be a blocking bug, but a feature of "autoclose" not implemented (this feature of autoclose was never implemented in previous version). Backport of bugs must be done only on very critical and blocking bugs to avoid infinite fight against regressions.

PR #21074, opened 2022-05-31
NEW recurring invoices : init hooks

gdsoftdev:
init hooks like in invoice card to have the same result as if we have created the invoices manually

### DIFF
--- a/htdocs/compta/facture/class/facture-rec.class.php
+++ b/htdocs/compta/facture/class/facture-rec.class.php
@@ -1286,6 +1286,10 @@ class FactureRec extends CommonInvoice
 			$sql .= ' AND rowid = '.((int) $restrictioninvoiceid);
 		}
 		$sql .= $this->db->order('entity', 'ASC');
+
+		//init hooks like in card to have the same result as if we have created the invoices manually 
+		$hookmanager->initHooks(array('invoicecard','globalcard'));
+
 		//print $sql;exit;
 		$parameters = array(
 			'restrictioninvoiceid' => $restrictioninvoiceid,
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/class/facture-rec.class.php
```diff
@@ -1286,6 +1286,10 @@ public function createRecurringInvoices($restrictioninvoiceid = 0, $forcevalidat
 			$sql .= ' AND rowid = '.((int) $restrictioninvoiceid);
 		}
 		$sql .= $this->db->order('entity', 'ASC');
+
+		//init hooks like in card to have the same result as if we have created the invoices manually
+		$hookmanager->initHooks(array('invoicecard','globalcard'));
```
eldy, line 1291: The context of a hook (param of initHooks) should not be set inside a CRUD method but by the caller (that know what is the context).
This means the beforeCreationOfRecurringInvoices can be created in different context.
If you need to have your hook working the same way for each context call, you must make your method working for each context set by caller.

gdsoftdev: You are right @eldy  but I give you an example : when recurring invoices are created by cron job, the generated documents are not the same as if we have created it manually because of modules (like subtotal, autoliquidationtva, sitfac, ...) hooking on invoicecard context.
What would you do in those cases ? All modules have to hook on an other context ? Is there a better place where we could init hook 'invoicecard' context ?
Thanks

gdsoftdev: So the context "cli" is a new context that module developers need to know and all modules concerning invoices must be modified to use this context.
The problem is that as long as the modules are not corrected, the documents will not output the same result by the card as by the cron (like subtotal or autoliquidationtva)
I'll let my modifications on my clients installation until then.
Thanks

eldy:
Closed. 
Context for code ran from cron is "cli" so any hook is able to make a different code if hook as called from cli cron or from a page.

PR #21081, opened 2022-06-01
New split all button on mo

atm-quentin:
# New split all button on mo

Adding a button to split all line with qty 1 by one click

https://user-images.githubusercontent.com/21307529/171349747-033a6a74-5f52-4133-a51b-b9efcf989c64.mp4


eldy:
I don't think the button "split in n lines" should be on the line that present the remain qty to consume/product.
We already have a button to split the line to enter different lot on the line prefilled. It currently split qty n into  1 and n-1.
If you need a feature to split  qty n into n*1,  is is just a different rule of splitting, so it should be on same level (same place) that existing button (on left for example).
It looks more clear to me, not for you ?

PR #21085, opened 2022-06-01
NEW Add a BOM tab on product card

atm-orian:
# NEW: Add a BOM tab on product card
When you visit a product and click on the BOM tab, it shows the list of BOM linked to a product

### DIFF
--- a/htdocs/bom/bom_card.php
+++ b/htdocs/bom/bom_card.php
@@ -116,13 +116,13 @@ if (empty($reshook)) {
 		}
 	}
 
-    if ($cancel) {
-        if (!empty($backtopage)) {
-            header("Location: ".$backtopage);
-            exit;
-        }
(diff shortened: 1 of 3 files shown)


eldy:
We already have this information on the tab "Referring object" of a product where we can see list of all object linked to a product. There is a line for BOM like any other object that you can click to see the detail.

atm-orian:
Hello, the Bom tab is now available in hidden conf. Moreover, this tab is different from the "referring object" tab because we have filter, sorting, paging and a create button.

eldy:
Even with a hidden conf, we introduce duplicated code and duplicated feature. Imagine if we add a hidden conf + a page for a tab for order, invoice, MO, etc...
I would prefer to see the interesting addition you did like filter, sorting, paging and create into the existing page.

PR #21092, opened 2022-06-01
New : Empty case on extrafields selects

ghost:
# NEW : Empty filter case on select extrafields

The "selectEmpty" filter will allow to bypass the other filters to see only the lines with NULL values.
I took example from Tag/Categ from Product/Service.

This PR replaces [this PR](https://github.com/Dolibarr/dolibarr/pull/19590)
Feature redeveloped to adapt to @atm-john's improvement

### DIFF
--- a/htdocs/core/tpl/extrafields_list_search_sql.tpl.php
+++ b/htdocs/core/tpl/extrafields_list_search_sql.tpl.php
@@ -67,8 +67,8 @@ if (!empty($extrafieldsobjectkey) && !empty($search_array_options) && is_array($
 			} elseif ($typ === 'select' and is_string($crit)) {
                 $critArray = explode(',', $crit);
                 if(in_array(-2, $critArray)) {
-                    array_shift($critArray);
-                    $critArray[] = 0;
+                    $key = array_search(-2, $critArray);
+                    $critArray[$key] = 0;
                 }
                 $critSelect = "'".implode("','", array_map(array($db, 'escape'), $critArray))."'";
 				$sql .= " AND (COALESCE(".$extrafieldsobjectprefix.$tmpkey.", 0) IN (".$db->sanitize($critSelect, 1).") )";
--- a/htdocs/core/tpl/extrafields_list_search_sql.tpl.php
+++ b/htdocs/core/tpl/extrafields_list_search_sql.tpl.php
@@ -65,12 +65,12 @@ if (!empty($extrafieldsobjectkey) && !empty($search_array_options) && is_array($
 			if (is_array($crit)) {
 				$crit = implode(' ', $crit); // natural_search() expects a string
 			} elseif ($typ === 'select' and is_string($crit)) {
-                $critArray = explode(',', $crit);
-                if(in_array(-2, $critArray)) {
-                    $key = array_search(-2, $critArray);
-                    $critArray[$key] = 0;
-                }
-                $critSelect = "'".implode("','", array_map(array($db, 'escape'), $critArray))."'";
+				$critArray = explode(',', $crit);
+				if(in_array(-2, $critArray)) {
+					$key = array_search(-2, $critArray);
+					$critArray[$key] = 0;
+				}
+				$critSelect = "'".implode("','", array_map(array($db, 'escape'), $critArray))."'";
 				$sql .= " AND (COALESCE(".$extrafieldsobjectprefix.$tmpkey.", 0) IN (".$db->sanitize($critSelect, 1).") )";
 				continue;
 			}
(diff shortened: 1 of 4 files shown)


htdocs/core/tpl/extrafields_list_search_sql.tpl.php
```diff
@@ -64,9 +64,14 @@
 			}
 			if (is_array($crit)) {
 				$crit = implode(' ', $crit); // natural_search() expects a string
-			} elseif ($typ === 'select' and is_string($crit) and strpos($crit, ',') === false) {
-				$critSelect = "'".implode("','", array_map(array($db, 'escape'), explode(',', $crit)))."'";
-				$sql .= " AND (".$extrafieldsobjectprefix.$tmpkey." IN (".$db->sanitize($critSelect, 1).") )";
+			} elseif ($typ === 'select' and is_string($crit)) {
```
eldy, line 67: Can you provide an example of value for 
$crit
not working for old code and that need this new code ?

atm-gregr: The part of the condition that says `and strpos($crit, ',') === false`, so it doesn't run the block if $crit contains a comma, but the block itself expects $crit to contain commas (`explode(',', $crit)`).

So if the user selects "empty value" and "3" for instance, `$crit` will have the value `-2,3`. By removing the comma exclusion condition, it allows us to show the elements where the extrafield's value is either empty or 3.

eldy: So I think what we anted to do was 
`and strpos($crit, ',') !== false`
instead of 
`and strpos($crit, ',') === false`

htdocs/core/tpl/extrafields_list_search_sql.tpl.php
```diff
@@ -64,9 +64,14 @@
...
+				$critArray = explode(',', $crit);
+				if(in_array(-2, $critArray)) {
+					$key = array_search(-2, $critArray);
+					$critArray[$key] = 0;
+				}
+				$critSelect = "'".implode("','", array_map(array($db, 'escape'), $critArray))."'";
+				$sql .= " AND (COALESCE(".$extrafieldsobjectprefix.$tmpkey.", 0) IN (".$db->sanitize($critSelect, 1).") )";
```
eldy, line 74: Why do you need the coalesce here ? According to documentation when using a SELECT extrafields:

"La liste doit être de la forme clef,valeur (où la clé ne peut être '0')
par exemple :
1,valeur1
2,valeur2
3,valeur3
...
"

so value must never be 0 for keys on select extrafields (such case will never be clear because sometime it means a specific value so we must make selected only if value is 0 and sometime it means unknown/undefined so must be selected when value is null. This is why 0 must never be used in sub-list of select extrafields.
If this coalesce fix a case, i think it will also break other cases.

htdocs/core/tpl/extrafields_list_search_sql.tpl.php
atm-florianm: Yes, I think this was the intention, but if we add this condition, it means we need more code to handle the specific case when the user selects only one value in the filter. I think the code in the PR handles both cases (unique value or multiple values) fine.

htdocs/core/tpl/extrafields_list_search_sql.tpl.php
eldy: > The way I understand it, without COALESCE, rows where the value of the extrafield is `NULL` would not be included when the user filters on the "empty" value. If 0 is indeed not allowed as a regular value for SELECT extrafields, there should be no conflict of usage (if the extrafield value is 0, it is equivalent to empty, no?)

Yes, if value is 0, the COALESCE is useless but it breaks the ability to use the index for database.

What do you mean with filtering on empty value ? For select, empty value are also not allowed. Can you provide one example ?

atm-florianm: > COALESCE […] breaks the ability to use the index for database engine in all other cases.

Extrafields are not indexed by default, but I agree that sometimes we create indexes on extrafields to speed up queries. If this is problematic, we can certainly find an alternative.

> What do you mean with filtering on empty value ? Do you mean filtering on "Value not yet defined" ?

The idea is: in the list, you select « empty value » in the extrafield's filter and it shows only records where this extrafield is NULL.

> If you means this how do you get in this situation to have an extrafield select and no value defined ?

When you create an extrafield, you can choose if it is mandatory or not. If the extrafield is not mandatory, you can create records where the value of the extrafield in question will be NULL. And, as you mentioned, you can also create an extrafield after having created records of its object, so those will also be NULL.

eldy: So if goal is to have a choice of a filter for end user "Value not defined", we should see a new sql filter line
$sql .= " AND field is null";
No need for coalesce.

eldy:

PR #21096, opened 2022-06-01
Fix deposit management on accountancy supplier index

SylvainLegrand:


htdocs/accountancy/supplier/index.php
```diff
@@ -299,10 +299,14 @@
...
+if (!empty($conf->global->FACTURE_DEPOSITS_ARE_JUST_PAYMENTS)) {
+	$sql .= " AND ff.type IN (".FactureFournisseur::TYPE_STANDARD.",".FactureFournisseur::TYPE_REPLACEMENT.",".FactureFournisseur::TYPE_CREDIT_NOTE.")";
+} else {
+	$sql .= " AND ff.type IN (".FactureFournisseur::TYPE_STANDARD.",".FactureFournisseur::TYPE_REPLACEMENT.",".FactureFournisseur::TYPE_CREDIT_NOTE.",".FactureFournisseur::TYPE_DEPOSIT.")";
+}
 $sql .= " GROUP BY ffd.fk_code_ventilation,aa.account_number,aa.label";
-$sql .= ' ORDER BY aa.account_number';
 
-dol_syslog('htdocs/accountancy/supplier/index.php');
+dol_syslog('htdocs/accountancy/supplier/index.php sql='.$sql, LOG_DEBUG);
```
eldy, line 309: No need to add sql='.$sql   here, the sql is logged just after always into the db->query

htdocs/accountancy/supplier/index.php
```diff
@@ -299,10 +299,14 @@
 $sql .= "  AND ffd.product_type <= 2";
 $sql .= " AND ff.entity IN (".getEntity('facture_fourn', 0).")"; // We don't share object for accountancy
 $sql .= " AND aa.account_number IS NULL";
+if (!empty($conf->global->FACTURE_DEPOSITS_ARE_JUST_PAYMENTS)) {
+	$sql .= " AND ff.type IN (".FactureFournisseur::TYPE_STANDARD.",".FactureFournisseur::TYPE_REPLACEMENT.",".FactureFournisseur::TYPE_CREDIT_NOTE.")";
+} else {
+	$sql .= " AND ff.type IN (".FactureFournisseur::TYPE_STANDARD.",".FactureFournisseur::TYPE_REPLACEMENT.",".FactureFournisseur::TYPE_CREDIT_NOTE.",".FactureFournisseur::TYPE_DEPOSIT.")";
```
eldy, line 305: The default behaviour is to use all invoices. So no need for a filter here.

htdocs/accountancy/supplier/index.php
```diff
@@ -299,10 +299,14 @@
 $sql .= "  AND ffd.product_type <= 2";
 $sql .= " AND ff.entity IN (".getEntity('facture_fourn', 0).")"; // We don't share object for accountancy
 $sql .= " AND aa.account_number IS NULL";
+if (!empty($conf->global->FACTURE_DEPOSITS_ARE_JUST_PAYMENTS)) {
```
eldy, line 302: Currently FACTURE_DEPOSITS_ARE_JUST_PAYMENTS is a deprecated and abandonned option for sales invoice only.
Because it has been deprecated, it was not implemented for supplier invoice. But if we introduce it, we should name the option differently, for exemple FACTURE_SUPPLIER_DEPOSITS_ARE_JUST_PAYMENTS

PR #21100, opened 2022-06-01
php V8 warning

defrance:


eldy:
Fixed in v18 with isModEnabled(), so i close this pr

PR #21108, opened 2022-06-02
fix: actioncomm.code have to be updated with correct c_actioncomm code (like in create)

FHenry:
fix: actioncomm.code have to be updated with correct c_actioncomm code (like in create)

### DIFF
--- a/htdocs/comm/action/class/actioncomm.class.php
+++ b/htdocs/comm/action/class/actioncomm.class.php
@@ -1127,7 +1127,7 @@ class ActionComm extends CommonObject
 		$sql = "UPDATE ".MAIN_DB_PREFIX."actioncomm ";
 		$sql .= " SET percent = '".$this->db->escape($this->percentage)."'";
 		if ($this->type_id > 0) {
-			$sql .= ", fk_action = ".(int)$this->type_id;
+			$sql .= ", fk_action = ".(int) $this->type_id;
 			if (empty($this->type_code)) {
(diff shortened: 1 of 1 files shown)


htdocs/comm/action/class/actioncomm.class.php
```diff
@@ -1128,6 +1128,12 @@ public function update(User $user, $notrigger = 0)
 		$sql .= " SET percent = '".$this->db->escape($this->percentage)."'";
 		if ($this->type_id > 0) {
 			$sql .= ", fk_action = '".$this->db->escape($this->type_id)."'";
+			$cactioncomm = new CActionComm($this->db);
```
eldy, line 1131: into object ActionComm we have 2 code:
->type_code that is the code found into parent table llx_c_actioncomm
->code that is a free text (for example "AC_MRP_MO_VALIDATE", "AC_ORDER_SUPPLIER_RECEIVE", ...)
For compatibility purpose we set the ->code with value found into code of llx_c_actioncomm when no value is provided (but this is rare). But it is a different data than type_code that is the code of the type of event (manual or automatic).

So, if we update it into "update" (and we should be able to do it), we should to it with property $this->code, and we must use the value from CActionComm only if $this->code is empty (not provided).

FHenry:
@eldy ok I've change it, use always $this->type_code on update, but if fk_action>0 and empty(  $this->type_code) then we set $this->type_code with code from CActionComm

PR #21115, opened 2022-06-02
emailcollector : hookmanager is null

rycks:
# FIX hookmanager is null

just a missing global call ...

```
PHP Fatal error:  Uncaught Error: Call to a member function initHooks() on null in /var/www/dolibarr-15/htdocs/emailcollector/class/emailcollector.class.php
```


rycks:
haaaaaaa one more bullet in my foot ! please wait i pushed a bad commit in that pull request !
i don't know how to delete a commit into that pull request ....

Issue #21116, opened 2022-06-02
Enhance addmoduletoeamailcollectorjoinpiece hook to pass Attachment(s)


rycks:
### Feature Request

if i would to capture attachment(s) from addmoduletoeamailcollectorjoinpiece hook i need something more in function call

### Use case

In ScanInvoices for example i would like to implement a solution with email collector, take pdf files and put in the ScanInvoices queue...

### Suggested implementation

a PR will come, please wait a minute :)

rycks:
Maybe linked to #18049

rycks:
Hmmm what do you think about the next need and add $operation and not only $data ?
```
$reshook = $hookmanager->executeHooks('addmoduletoeamailcollectorjoinpiece', $parameters, $data, $operation);
```
because we could need informations like $operation['actionparam'] into plugins who handle that hook ...
if you are ok i will put a second commit into that PR
thanks

PR #21117, opened 2022-06-02
new: add data to addmoduletoeamailcollectorjoinpiece hook

rycks:
# New #21116 

i would like to be able to handle attachment files by a plugin (scaninvoices for example) and for that i need more data on the hooked function

### DIFF
--- a/htdocs/emailcollector/class/emailcollector.class.php
+++ b/htdocs/emailcollector/class/emailcollector.class.php
@@ -1904,6 +1904,7 @@ class EmailCollector extends CommonObject
 						}
 					} elseif ($operation['type'] == 'recordjoinpiece') {
 						$pj = getAttachments($imapemail, $connection);
+						$data = array();
 						foreach ($pj as $key => $val) {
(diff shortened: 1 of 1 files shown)


htdocs/emailcollector/class/emailcollector.class.php
```diff
@@ -1904,6 +1904,7 @@ public function doCollectOneCollector()
 						}
 					} elseif ($operation['type'] == 'recordjoinpiece') {
 						$pj = getAttachments($imapemail, $connection);
+						$data = array();
```
eldy, line 1907: In method  executeHooks, the third param where $data is used is declared as
@param		Object	$object			Object to use hooks on
What happen if you set here
$data = new stdClass();
instead of 
$data = array();
Does is solve your need ? if yes it would be better...

rycks: Ok no problem, $data could store a list of attachment as a array or as an object with list of entries

PR #21121, opened 2022-06-03
FIX php8.1 compatibility

hregis:


htdocs/core/modules/syslog/mod_syslog_file.php
```diff
@@ -173,7 +173,7 @@ public function export($content, $suffixinfilename = '')
 				$this->lastTime = $now;
 			}
 
-			$message = strftime("%Y-%m-%d %H:%M:%S", time()).$delay." ".sprintf("%-7s", $logLevels[$content['level']])." ".sprintf("%-15s", $content['ip'])." ".($this->ident > 0 ?str_pad('', $this->ident, ' ') : '').$content['message'];
+			$message = date("Y-m-d H:i:s", time()).$delay." ".sprintf("%-7s", $logLevels[$content['level']])." ".sprintf("%-15s", $content['ip'])." ".($this->ident > 0 ?str_pad('', $this->ident, ' ') : '').$content['message'];
```
eldy, line 176: date() must be avoid because the problem of timezone is not clear. Use instead 
dol_print_date(dol_now(), "standard")

htdocs/core/lib/functions.lib.php
```diff
@@ -2761,7 +2761,7 @@ function dol_mktime($hour, $minute, $second, $month, $day, $year, $gm = 'auto',
 	}
 	//var_dump($localtz);
 	//var_dump($year.'-'.$month.'-'.$day.'-'.$hour.'-'.$minute);
-	$dt = new DateTime(null, $localtz);
+	$dt = new DateTime('', $localtz);
```
eldy, line 2764: '' seems not valid.  Is it ok with  'now'  (suggested by php doc)  ?

hregis: @eldy this is the warning : 
datetime::__construct(): passing null to parameter #1 ($datetime) of type string is deprecated

PR #21128, opened 2022-06-03
fix : MAIN_DB_PREFIX is already into $tabname[$id]

grandoc:


htdocs/admin/dict.php
```diff
@@ -851,7 +851,7 @@
 			if ($tabrowid[$id]) {
 				// Get free id for insert
 				$newid = 0;
-				$sql = "SELECT MAX(".$tabrowid[$id].") as newid FROM ".MAIN_DB_PREFIX.$tabname[$id];
+				$sql = "SELECT MAX(" . $tabrowid[$id] . ") as newid FROM " . $tabname[$id];
```
eldy, line 854: MAIN_DB_PREFIX should not be into $tabname.

PR #21134, opened 2022-06-05
FIX #20444

### LINKED ISSUE #20444: V15.0.1 Bug : Customer Currency is not updated
### Bug

Updated to V15.1, found a bug of customer currency is not updated in customer profile.

Please refer to the attachment picture for details

### Environment Version

15.0.1

### Environment OS

Ubuntu 18.04

### Environment Web server

Apache2.4

### Environment PHP

7.4

### Environment Database

MySQL

### Environment URL(s)

https;//www.dolibarr-URL.com/societe/card.php?leftmenu=customers&action=create&type=c

Expected : The currency and value should change as per the customer currency.  
Actual : The currency and value remains as per company currency.

1. Setup > Modules > Multi currency and add a new currency and add exchange value
2. create a new customer and select the new currency as customer currency
3. Enter the value of "Capital"
4. You will find that the value is not changed as per the customer currency but remains as per company currency.

### Attached files

image

fboitel:
FIX #20444

### DIFF
--- a/htdocs/societe/card.php
+++ b/htdocs/societe/card.php
@@ -1139,6 +1139,11 @@ if (is_object($objcanvas) && $objcanvas->displayCanvasExists($action)) {
 			$object->country = $tmparray['label'];
 		}
 		$object->forme_juridique_code = GETPOST('forme_juridique_code');
+
+		// We set multicurrency_code if enabled
+		if (!empty($conf->multicurrency->enabled)) {
+			$object->multicurrency_code = GETPOST('multicurrency_code') ? GETPOST('multicurrency_code') : $conf->currency;
+		}
 		/* Show create form */
 
 		$linkback = "";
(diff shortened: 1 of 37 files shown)


htdocs/societe/card.php
```diff
@@ -1261,6 +1266,16 @@ function init_check_no_email(input) {
                      });';
 				print '</script>'."\n";
 			}
+			if (!empty($conf->multicurrency->enabled)) {
```
eldy, line 1269: Using js here has a default. 
If we have filled manually some other field, and have not finished, the form will be submitted before we have finished to fill it.
I think we can completely remove the js part and have the currency modification visible only after the first submit.

htdocs/societe/card.php
```diff
@@ -1656,7 +1674,7 @@ function init_check_no_email(input) {
 			print '<tr>';
 			print '<td>'.$form->editfieldkey('Currency', 'multicurrency_code', '', $object, 0).'</td>';
 			print '<td colspan="3" class="maxwidthonsmartphone">';
-			print $form->selectMultiCurrency(($object->multicurrency_code ? $object->multicurrency_code : $conf->currency), 'multicurrency_code', 1);
+			print $form->selectMultiCurrency((GETPOSTISSET('multicurrency_code') ? GETPOST('multicurrency_code') : $object->multicurrency_code), 'multicurrency_code', 1);
```
eldy, line 1677: Introducing the GETPOSTISSET seems good. But for the case test is false, we fallback on 
($object->multicurrency_code)
 where we should fallback on 
($object->multicurrency_code ? $object->multicurrency_code : $conf->currency)    -  like in old code
so if field was empty (never initialized in db and $object->multicurrency_code is null), we fallback on the global currency setup.

eldy (changes requested):
This test line 2783 is not enough: Imagine we never set any specific currency for the company.
I suggest to replace it with
if (!empty($conf->multicurrency->enabled) && !empty($object->multicurrency_code)) {

PR #21136, opened 2022-06-06
HRM experimental to stable for 16.0

atm-arnaud:
We already debugged this module and he is ready to be used in production

### DIFF
--- a/htdocs/core/modules/modHRM.class.php
+++ b/htdocs/core/modules/modHRM.class.php
@@ -56,7 +56,7 @@ class modHRM extends DolibarrModules
 		// Module description, used if translation string 'ModuleXXXDesc' not found (where XXX is value of numeric property 'numero' of module)
 		$this->description = "HRM";
 		// Possible values for version are: 'development', 'experimental', 'dolibarr' or version
-		$this->version = 'experimental';
+		$this->version = 'dolibarr';
 		// Key used in llx_const table to save module status enabled/disabled (where MYMODULE is value of property name of module in uppercase)
 		$this->const_name = 'MAIN_MODULE_'.strtoupper($this->name);
 		// Name of image file used for this module.


atm-maxime:
@eldy what do you mean by postponed ? Do you plan to have it stable in v16 ? Do we need to do something on our side ?

atm-arnaud:
@eldy the aim is to change the version for 16.0 (already 1 version experimental)

eldy:
There is still a lot of trouble i want to fix before making it stable.
- Don't understand what is the field "Ctrl" 
- Can't link a skill into a job (error saying it is already linked but no link visible).
- sql error when doing a comparison on a job position.
- sql error when modifying an evaluation 
- I did no succeed in adding a line into an evaluation ? How can we do this ?

I am currently working of the translation and label of term to have same value, translation key when we speak about the same object.

Currently module is not yet stable enough for me to be stable in v16. Can you check my posted point ?

atm-arnaud:
@eldy 

- Don't understand what is the field "Ctrl"
This is "deplacement" (i don't understand why the translation is Ctrl). Like : Is this a job with business travels ?

- Can't link a skill into a job (error saying it is already linked but no link visible).
That's not normal with the interface. (it doesn't use linkedobject table one the card. We have a special interface in the card)

- sql error when doing a comparison on a job position.
We have to fix that

- sql error when modifying an evaluation
We have to fix that

- I did no succeed in adding a line into an evaluation ? How can we do this ?
There is no method to add a skill into an evaluation. The goal is to reach jobs with skills. Next evaluate an user for a job (autoimplement skills linked to a job)

eldy:
>   There is no method to add a skill into an evaluation. The goal is to reach jobs with skills. Next evaluate an user for a job (autoimplement skills linked to a job)

Sorry i used wrong term. This is my problem. I did'nt succeed in linked a skill to a job on this tab. When i click on "Add", a message says skill is already linked to job position. Even if not.

image

Note: For the Crtl, is renamed the translation key into "NeedBusinessTravels", so it is clear for this point now.

atm-arnaud:
@eldy we are currently working on. Will be good in 1 week. We will link PR to this one

eldy:
Has been set to stable into v17

PR #21141, opened 2022-06-07
NEW: Send Notification Without File when E-mail Template doesn't attach file

almisbahinfo:
When we create an Email Template that doesn't require main file to be attached, we still send notification with an attached file, eg: when we send a notification when validating proposal it send proposal with attached file with mention NOT VALIDATED.
I added OR condition on checking pdf_path.

### DIFF
--- a/htdocs/core/class/notify.class.php
+++ b/htdocs/core/class/notify.class.php
@@ -615,7 +615,7 @@ class Notify
 
 						$ref = dol_sanitizeFileName($newref);
 						$pdf_path = $dir_output."/".$ref.".pdf";
-						if (!dol_is_file($pdf_path)) {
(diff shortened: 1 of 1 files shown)


htdocs/core/class/notify.class.php
```diff
@@ -615,7 +615,7 @@ public function send($notifcode, $object, $filename_list = array(), $mimetype_li
 
 						$ref = dol_sanitizeFileName($newref);
 						$pdf_path = $dir_output."/".$ref.".pdf";
-						if (!dol_is_file($pdf_path)) {
+						if (!dol_is_file($pdf_path)||!$arraydefaultmessage->joinfiles) {
```
eldy, line 618: $arraydefaultmessage may still be null here (it depends on context). And if null, $arraydefaultmessage->joinfiles will generate a fatal error. So test must be enhanced to support this case...

PR #21143, opened 2022-06-07
FIX #20696

### LINKED ISSUE #20696: Creation date of transaction in bookkeeping is sometimes in future
### Bug

When creating a new transaction in accounting at 22:00 (GMT+01:00), the creation date is set to tomorrow (future) while the document date is set to today (correct).

### Environment Version

14.0.5

### Environment URL(s)

htdocs/accountancy/bookkeeping/card.php

Creation date should be today till 23h59

Go to bookkeeping
Try to create a new entry at 22h
see result

### Attached files

image

fboitel:
FIX #20696


eldy:
Date conversion from string must be done just after the reading of database to have the data in memory with the correct date format.

PR #21172, opened 2022-06-09
Look & Feel - Align amount to right on card

aspangaro:
Several clients have asked us to attach the amounts to the right side of the cards for clarity and to ensure that the numbers line up properly

image

image

To avoid wasting time if this PR is refused, I will propose the modifications on the other cards if the principle is ok


eldy:
Seems good because in v16 all section with amount have been moved to be alone (no other fields except amount in the right, so changing alignement should not be a problem, we should have amount only). But the main reason was:
Merging the 6 lines when we used the multi-currency into 3 lines:
"Montant HT                       X €                    Y $"
"Montant TVA                     X2 €                  Y2 $"
"Montant TTC                    X3 €                  Y3 $"

But we can start by making the right align 
Can you complete PR with other main object ?

aspangaro:
@eldy The same modification on supplier invoice case. (This is the only case with the principle of recalculation of VAT)

I am not for merging the lines of amounts with their amount in another currency, it will be less clear in my opinion

Before

image

After

image

PR #21176, opened 2022-06-09
add var_dump check

frederic34:


frederic34:
image

eldy:
The 3 warning reported are expected print_r (because it is tools to output debugging info).
How can be avoid this false positive ?

frederic34:
@eldy files added in exclude list

PR #21184, opened 2022-06-09
FIX #21094 

### LINKED ISSUE #21094: Function "getChildsArbo" from Product did not give all sub products from a kit
### Bug

Example in the screenshot bellow 

Function "getChildsArbo" doesn't give the right result when the following case occurs 
- Product_1 : Has Product_3 and Product_2  as child 
- Product_2 : Has Product_3 and Product_4 as child 

Current result of the function is : 
Product_1
-> product_3
-> product_2
--> product_4

instead of 
Product_1
-> product_3
-> product_2
--> product_3
--> product_4

In the tree, we do not see the right amount of quantity because product_3 can be part of product_1 ans part of product 2

How to patch : 
**/product/class/product.class.php - Line 4688**
Remove "continue statement" 
`if (!empty($alreadyfound[$rec['rowid']])) {
					dol_syslog(get_class($this).'::getChildsArbo the product id='.$rec['rowid'].' was already found at a higher level in tree. We discard to avoid infinite loop', LOG_WARNING);
					continue;}`

### Environment Version

14.0.4

### Environment OS

_No response_

### Environment Web server

Apache

### Environment PHP

7.4

### Environment Database

Mysql

### Environment URL(s)

https://XXXX.com/expedition/shipment.php?id=XXX

_No response_

_No response_

### Attached files

Image_1_06_22_21_05_collée

Quatadah:
FIX #21094 
commenting a line "continue;" => resolves the bug

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -4681,7 +4681,7 @@ class Product extends CommonObject
 			while ($rec = $this->db->fetch_array($res)) {
 				if (!empty($alreadyfound[$rec['rowid']])) {
 					dol_syslog(get_class($this).'::getChildsArbo the product id='.$rec['rowid'].' was already found at a higher level in tree. We discard to avoid infinite loop', LOG_WARNING);
-					continue;
+					//continue;
 				}
 				$alreadyfound[$rec['rowid']] = 1;
 				$prods[$rec['rowid']] = array(
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -4647,9 +4647,10 @@ class Product extends CommonObject
 	 * @param  int $id             		Id of product to search childs of
 	 * @param  int $firstlevelonly 		Return only direct child
 	 * @param  int $level          		Level of recursing call (start to 1)
+	 * @param  array $parents   	    Array of all parents of $id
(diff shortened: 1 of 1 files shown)


htdocs/product/class/product.class.php
```diff
@@ -4681,7 +4681,7 @@ public function getChildsArbo($id, $firstlevelonly = 0, $level = 1)
 			while ($rec = $this->db->fetch_array($res)) {
 				if (!empty($alreadyfound[$rec['rowid']])) {
 					dol_syslog(get_class($this).'::getChildsArbo the product id='.$rec['rowid'].' was already found at a higher level in tree. We discard to avoid infinite loop', LOG_WARNING);
-					continue;
+					//continue;
```
eldy, line 4684: If we remove this, you remove an important protection.
If A is linked to B and B to C and C to A, this will generates problem. In such a case, it means data are corrupted in database, because we dont want to have a relation parent-child with data with a child that has child that is already a parent of him.
So if we have in database  A -> B -> C -> A, we cant to return only 
A -> B -> C   (supposing that link C -> A is an error in database, it should not happen but if this happen, the function must deal it)

eldy:
It works but, too many sql...
To rewrite differently by adding a paramater that propagate the hierachy to child...

PR #21192, opened 2022-06-09
Add old fields on llx_element_tag 

defrance:
if we create the llx_element_tag table again, I suggest that we use the initial structure of the table (before its deletion in 2017)
For information, this table is not used except in one of my modules (and I bring the old structure since 2017)
In short, either we add the table with the old structure, or we do not add it with bullshit fields

### DIFF
--- a/htdocs/install/mysql/migration/15.0.0-16.0.0.sql
+++ b/htdocs/install/mysql/migration/15.0.0-16.0.0.sql
@@ -51,7 +51,10 @@
 -- VPGSQL8.2 ALTER TABLE llx_c_payment_term ALTER COLUMN rowid SET DEFAULT nextval('llx_c_payment_term_rowid_seq');
 -- VPGSQL8.2 SELECT setval('llx_c_payment_term_rowid_seq', MAX(rowid)) FROM llx_c_payment_term;
 
-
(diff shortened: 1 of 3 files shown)


eldy:
The goal of table llx_element_tag is to replace all tables llx_categories_xxx   (with one table).
It is here to allow dev to start to work on the work to introduce a generic table for link between an object and llx_categorie table.
So the existing fields seems enough.

defrance:
my problem is that this table is already used by one of my modules sold in nearly 200 copies, based on the old structure of the table (the one deleted in V14)
the name "llx_element_tag" does not seem to me suitable for the intended use, it seems more logical to me to use llx_element_category
--- 
mon problème c'est que cette table est déjà utilisé par l'un de mes modules commercialisé à près de 200 exemplaires, se basant sur l'ancienne structure de la table (celle supprimé en V14)
le nom "llx_element_tag" ne me semble pas adapté à l'usage qu'il en est prévu, il me semble plus logique d'employer llx_element_categorie

defrance:
considering this PR
https://github.com/Dolibarr/dolibarr/pull/21321
this doesn't seem relevant to me anymore.

Issue #20226, opened 2022-02-28
Bug: Adding pre-payment invoice to supplier's invoice does not reduce the amount


hasse-basse:
I've made a number of pre-payments to a supplier and created accordingly "factures d'accompte". I've now recived one invoice and then I chosed one of the availble prepayments, of 339.20€ to pay pay it.

![1](https://user-images.githubusercontent.com/24247236/156037759-03e8ada6-17a3-43e1-b036-584d86cb6a71.jpg)

As a result is it is written on the right side of the page as per below

![2](https://user-images.githubusercontent.com/24247236/156038174-aecbda46-c963-4bd6-ade9-500d62587d00.jpg)

Acompte FDA200100     **339.20**
Facturé                           **339.20**
Reste à payer                  **339.20**

In other words, the debt is not reduced, but available prepayment amount has been reduced width 339.20 and the invoice remain unpaid. In other words, the money has gone up in thin air.

The correct behaviour would have been that Rest à payer should have been at 0 and the invoice should have had the status Payé (Paid).

Acompte FDA200100     **339.20**
Facturé                           **339.20**
Reste à payer                  **0**

## Environment
- **Version**: 15.0.0
- **OS**: Linux
- **Web server**: Apache
- **PHP**: fpm-fcgi 7.3.33-1+0~20211119.91+debian10~1.gbp618351
- **Database**: MySQL or MariaDB 5.5.5-10.3.31-MariaDB-0+deb10u1
- **URL(s)**: /azert/fourn/facture/card.php?facid=519

## Expected and actual behavior
[*Verbose description*]

## Steps to reproduce the behavior
[*Verbose description*]

## [Attached files](https://help.github.com/articles/issue-attachments) (Screenshots, screencasts, dolibarr.log, debugging informations…)
[*Files*]

hregis:
@hasse-basse do you have externals modules?

hasse-basse:
CONCAT PDF 6.0.2
JALONS 9.0+2.3.0 (seem to be your module!) I can see it is too old! Does version 14 work with Dolibarr 15?
Ultimat PDF an old version. I don't use it so I will take it away

I've tried to turn all of them off, but it didn' help!

hasse-basse:
I just tried it on another invoice and it worked as designed  si I don't know what is going on. Maybe it still was about one of the modules as suggested.

I close the issue.

hasse-basse:
[...]

Issue #19724, opened 2022-01-02
14.0.4 - Bug on Deposit Invoice Supplier For Final Supplier Invoice


Sylvain089:
# Bug
When generating a Deposit Supplier Invoice (After paid this invoice), and applying this "deposit" invoice on a Final Invoice Supplier, the amount of this invoice is not applied to "remain to pay" of the Final supplier invoice.

## Environment
- **Version**: 14.0.4
- **OS**: Linux (NAS Synology), although irrelevant
- **Web server**: Apache, although irrelevant
- **PHP**: 7.4.18, although irrelevant
- **Database**: MariaDB 10, although irrelevant
- **URL(s)**: htdocs/fourn/facture/card.php?facid=23&save_lastsearch_values=1

## Expected and actual behavior
The remain to pay is not affected when applying a Deposit Supplier Invoice to a Final supplier invoice. 

## Steps to reproduce the behavior
- Create a Deposit Supplier Invoice with the TTC amount : 300 €
- Paid this Deposit Supplier Invoice
- Create a Final Supplier Invoice with the TTC amount : 900 €
- Validate the final supplier invoice
- Apply the Deposit Supplier Invoice (The invoice number and the total amount appear, on the right position)
- The Deposit amount is not apply on the remain to pay, the amount keeps : 900 €

You can fix this issue by : 
This error is due to a bug in the file : htdocs/core/class/commoninvoice.class.php : 
To fix this issue : The lines 162 to 164 must be commented to avoid return "0.0" for the function "public function getSumDepositsUsed($multicurrency = 0)"

## [Attached files](https://help.github.com/articles/issue-attachments) (Screenshots, screencasts, dolibarr.log, debugging informations…)
Subject in English Forum : 
https://www.dolibarr.org/forum/t/report-bug-supplier-deposit-invoice/21015/2

josett225:
Hi there,

This issue is still  in Dolibarr v14.0.5.
Also strangely, we cannot do it anymore in V15 and V16 as the Deposit can be used only as a discount line added to the supplier invoice. Anyway to fix the issue as explain above and if possible understand the changes done in V15 and V16 
See screenshots for v14 and v16 and V14 with the fix above.
Many thanks

V14 with no fix
<img width="626" alt="Screenshot at May 20 19-55-53 Deposit issue" src="https://user-images.githubusercontent.com/57623859/169586899-a5bc0bba-0390-483d-8434-e1ffcd6253e8.png">
V16
[...]

PR #21194, opened 2022-06-09
Fix supplier deposit can not be used to pay supplier invoice

SylvainLegrand:


SylvainLegrand:
When a supplier requests a deposit, it is paid and transformed into an exceptional discount.
When you use this discount to pay all or part of the supplier's invoice, it is not deducted from the amount remaining due on the invoice.

eldy:
A deposit of 10 euros must not be deducted from the remain to pay of an invoice of 100 euros but must be deducted from the total on the remaining invoice (so second invoice must be 90 euros with a remain to pay of 90 euros, not 100 euros with already paid 10 euros).
Because a deposit is an invoice and is already counted into the turnover and must not be counted twice, you must have 1 invoice of 10 for the deposit and 1 invoice of 90 euros for the second invoice. So a deposit must be included as a item of -10 in the second invoice so invoice is 90 euros and not as a payment of 10 euros on an invoice of 100.

josett225:
Hi @eldy and @SylvainLegrand 

There is already 2 opened issues (if I am not wrong), could you please review the following content :
https://github.com/Dolibarr/dolibarr/issues/19724
https://github.com/Dolibarr/dolibarr/issues/20226

I already discussed with accounting expert and it seems both solutions could work. 
Main issue in Edly's behavior is the fact that the displayed margins are wrong as the deposit is used as a product line. See issue #19724. All invoices are either in 707 or 607 account classes.
In the meantime Sylvain's behavior works well as the Deposit is used like a payment after correcting the code in v14  (in V15 the setup has also to changed with FACTURE_DEPOSITS_ARE_JUST_PAYMENTS set to 1). Deposit invoices are in 409 or 419 accounting classes and final invoice is 707 or 607. 

As I am not expert as you are,  I let you comeback  with your thoughts/directions.

Best Regards

Jose

eldy:
currently, there is no known problem of margin shown when using the default solution (so when FACTURE_DEPOSITS_ARE_JUST_PAYMENTS is 1) :  Don't forget that the deposit is an invoice, so you have a part of the margin in this invoice too, and another part of the margin in the final invoice. You must take into account the 2 invoice to get a good idea of the complete margin. 
If you think you experience problem in margin in default mode, thanks to paste a view of invoice (lines and margin section) for both the deposit and the finale invoice.

When using the other mode (when FACTURE_DEPOSITS_ARE_JUST_PAYMENTS is set to 1), may be there is margin trouble, this mode is deprecated. A lot of accounting feature can't work correctly in this mode. Note that to get a similar behaviour on purchase/supplier side, the option will be FACTURE_SUPPLIER_DEPOSITS_ARE_JUST_PAYMENTS but this option is introduced in v16 only (also a hidden feature because not the recommended way to work).

For the way to make accounting on deposit, having the deposit moved first into 409 or 419 for the deposit of 10 euros instead of 401 and 411 and then transferred later into 401 or 411 does not depends on the fact you use FACTURE_DEPOSITS_ARE_JUST_PAYMENTS to on or not. 
Currently, with a minimal setup, Dolibarr transfers the first 10 euros of deposit invoices directly in 401 or 411 and the remain of final invoice of 90 into 401 or 411 too, for a total of 100. The feature to transfer "firstly" in 409 and 419 and then moving this into 401 or 411 at the same moment we add the final 90, is implemented only if you set a "Default deposit account" to 419 into menu Home - Accounting - Setup - Default accounts). So, the use of 419 instead of 411 for deposit depends on this setup and not on deprecated option FACTURE_DEPOSITS_ARE_JUST_PAYMENTS. It is available for sale invoice only. Also note that you have to manually move amount from 419 to 411 (i think but i am not sure, i never use this setup).

PR #21196, opened 2022-06-10
NEW Accountancy - Manage customer retained warranty FPC21+

aspangaro:
Retained warranty is hidden by the constant INVOICE_USE_RETAINED_WARRANTY


htdocs/accountancy/admin/defaultaccounts.php
```diff
@@ -84,6 +84,9 @@
 if (!empty($conf->banque->enabled)) {
 	$list_account[] = 'ACCOUNTING_ACCOUNT_TRANSFER_CASH';
 }
+if (!empty($conf->global->INVOICE_USE_RETAINED_WARRANTY)) {
```
eldy, line 87: Can you explain in few sentences what is a retained warranty and the mechanism we need to manage into accountancy ?  (a simple example with data/value is often enough).

PR #21200, opened 2022-06-10
Save usermodid to updatePercent

leninrivas:
# FIX|Fix *save id of the user who updates the event from the list*


htdocs/comm/action/class/actioncomm.class.php
```diff
@@ -2515,12 +2515,13 @@ public function sendEmailsReminder()
 	 * @param int		$percent	The new percent value for the event
 	 * @return int					1 when update of the event was suscessfull, otherwise -1
 	 */
-	public function updatePercent($id, $percent)
+	public function updatePercent($id, $percent, $usermodid)
```
eldy, line 2518: For compatibility, add  $usermodid = 0 into function declaration.
Also, add a line into the doc comment.

leninrivas: Updated. Thank you

PR #21222, opened 2022-06-12
Fix php8 compatibility

ptibogxiv:


htdocs/user/bank.php
```diff
@@ -446,7 +446,7 @@
 	print "</tr>\n";
 
 	// Date of birth
-	if ($user->rights->hrm->read_personal_information->read || $user->rights->hrm->write_personal_information->write) {
+	if (isset($user->rights->hrm->read_personal_information->read) || isset($user->rights->hrm->write_personal_information->write)) {
```
eldy, line 449: if isset($user->rights->hrm->read_personal_information->read)
has not the same result than 
if $user->rights->hrm->read_personal_information->read
If var is set to 0, first is true, second false

You can user 
$user->hasRight('hrm', 'read_personal_information', 'read')

ptibogxiv: @eldy i update the PR

PR #21223, opened 2022-06-12
Fix#20444

### LINKED ISSUE #20444: V15.0.1 Bug : Customer Currency is not updated

fboitel:
FIX #20444

### DIFF
--- a/htdocs/accountancy/bookkeeping/card.php
+++ b/htdocs/accountancy/bookkeeping/card.php
@@ -540,7 +540,7 @@ if ($action == 'create') {
 		print '</td>';
 		print '</tr>';
 
-		// Date document creation
+		// Date document export
 		print '<tr>';
 		print '<td class="titlefield">'.$langs->trans("DateExport").'</td>';
 		print '<td>';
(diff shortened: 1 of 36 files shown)


eldy:
Note; in your comment use
"FIX #20444"
instead of 
"FIX#20444"
with a space so parsing tools executed on merging PR can detect the key "FIX" correctly. I fixed manually.

PR #21226, opened 2022-06-12
Fix php8 compatibility

ptibogxiv:


eldy:
Because this is a change into an external library, you must add a note on delta added into file  dev/dolibarr_changes.txt

ptibogxiv:
> Because this is a change into an external library, you must add a note on delta added into file dev/dolibarr_changes.txt

it's done

PR #21227, opened 2022-06-12
NEW support of Stripe Terminal with Takepos

ptibogxiv:
#19444 new PR
A recent change on Stripe could update with a better code later (less javascript)

### DIFF
--- a/htdocs/stripe/admin/stripe.php
+++ b/htdocs/stripe/admin/stripe.php
@@ -104,6 +104,10 @@ if ($action == 'setvalue' && $user->admin) {
 	if (!$result > 0) {
 		$error++;
 	}
+	$result = dolibarr_set_const($db, "STRIPE_LOCATION", GETPOST('STRIPE_LOCATION', 'alpha'), 'chaine', 0, '', $conf->entity);
+	if (!$result > 0) {
+		$error++;
+	}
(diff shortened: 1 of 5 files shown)


htdocs/stripe/admin/stripe.php
```diff
@@ -104,6 +104,10 @@
 	if (!$result > 0) {
 		$error++;
 	}
+	$result = dolibarr_set_const($db, "STRIPE_LOCATION", GETPOST('STRIPE_LOCATION', 'alpha'), 'chaine', 0, '', $conf->entity);
```
eldy, line 107: What is Stripe Location ?

eldy, line 107: You should add  
if (GETPOSTISSET('STRIPE_LOCATION'))
so when switching from debug and not debug, updating setup will not clear setup.

ptibogxiv: @eldy done today :)

eldy:
@ptibogxiv Can you write a chapter on this wiki page https://wiki.dolibarr.org/index.php/Module_Stripe_FR
to explain how we can use the Stripe TPE ?
Setup to do on stripe side and dolibarr side. How to test it.

PR #21232, opened 2022-06-13
php V8 warning

defrance:


htdocs/admin/modulehelp.php
```diff
@@ -266,7 +267,7 @@
 	$text .= '<br><span class="opacitymedium">'.$langs->trans("LastActivationIP").':</span> ';
 	$text .= $ip;
 }
-$lastactivationversion = (empty($tmp['lastactivationversion']) ? '' : $tmp['lastactivationversion']);
+$lastactivationversion = (empty($tmp['lastactivationversion']) ? '' : $tmp['lastactivationversion'][0]);
```
eldy, line 270: Use of $tmp['lastactivationversion'][0]   looks strange.
We store a string into
$tmp['lastactivationversion']
For example in line 1007 of DolibarrModules.php
Can you check ?

PR #21244, opened 2022-06-13
BUG FIX Class FichinterLigne does not have "date" attribute

fboitel:


eldy:
We started to standardize fields so any main date of objects is called ->date (we keep ->datexxx to keep compatibility)
If you search on ->date into fichinter dir, you will see some references to   ...->date, so if you remove the line, you will break features.
So i suggest instead to declare the property ->date and mark the datei as deprecated.

PR #21247, opened 2022-06-13
BUG FIX: evaluate() doesn't cover the case where $expr is empty

Quatadah:
# FIX
The evaluate() function in evalmath.class.php doesn't cover the case where $expr is empty, + adding unit test.

### DIFF
--- a/htdocs/core/class/evalmath.class.php
+++ b/htdocs/core/class/evalmath.class.php
@@ -144,6 +144,10 @@ class EvalMath
 	 */
 	public function evaluate($expr)
 	{
+		if (empty($expr)) {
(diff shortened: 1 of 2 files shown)


htdocs/core/class/evalmath.class.php
```diff
@@ -374,7 +374,7 @@ private function nfx($expr)
 	 */
 	private function pfx($tokens, $vars = array())
 	{
-		if ($tokens == false) {
+		if (empty($tokens)) {
```
eldy, line 377: Because you change code of an utility function, you must also add a phpunit test for the class
EvalMath.

PR #21249, opened 2022-06-13
BUG FIX Class Salary does not contain attribute salary but amount instead

fboitel:


htdocs/salaries/card.php
```diff
@@ -232,7 +232,7 @@
 	// Set user current salary as ref salary for the payment
 	$fuser = new User($db);
 	$fuser->fetch(GETPOST("fk_user", "int"));
-	$object->salary = $fuser->salary;
+	$object->amount = $fuser->salary;
```
eldy, line 235: The amount is used to store the amount of the salary of the month.
The field salary is used to store what is the default salary of user. Sometimes, we pay a salary of the month with a different value than the expected value. So you must keep object->salary = $fuser->salary   (fuser->salary is default salary of a user). But you must declare the property ->salary.

eldy:
@fboitel check in your IDE that the option  "remove spaces at end of line" is set to on.

PR #21255, opened 2022-06-14
PHP 8 in social contribution

aspangaro:
Lot of PHP8 errors found in social contribution module


htdocs/compta/sociales/payments.php
```diff
@@ -327,13 +333,19 @@
...
 			$accountstatic->iban = $obj->iban;
 			$accountstatic->bic = $obj->bic;
 			$accountstatic->currency_code = $langs->trans("Currency".$obj->currency_code);
 			$accountstatic->clos = $obj->clos;
+
+			if (!empty($conf->accounting->enabled)) {
+				$accountstatic->account_number = $obj->account_number;
+
+				$accountingjournal = new AccountingJournal($db);
+				$accountingjournal->fetch($obj->fk_accountancy_journal);
```
eldy, line 346: For performance reason, we must not include a subrequest inside a loop.
If you have a list of 100 request, it means you will make 101 requests instead of 1.
There is 2 solutions:
A) Getting the child property with a join in a select
B) Using a cache inside the loop so when we request the $accountingjournal->fetch for an id $obj->fk_accountancy_journal that we already had in a previous pass, we reuse the object in memory.

Because the number of journal is low, the solution B seems good.
This is an example found with a product found into list fourn/commande/dispatch.php


				```
		if (empty($conf->cache['product'][$objp->fk_product])) {
							$tmpproduct = new Product($db);
							$tmpproduct->fetch($objp->fk_product);
							$conf->cache['product'][$objp->fk_product] = $tmpproduct;
						} else {
							$tmpproduct = $conf->cache['product'][$objp->fk_product];
						}
```

aspangaro: I prefer the first method... I can't do it with the cache function

PR #21259, opened 2022-06-14
BUG FIX : changing function name to the existing one

Quatadah:
changing from getNbByMonth() to _getNbByMonth() because it's the one that exists.
However, this function still needs the parameter $sql.

### DIFF
--- a/htdocs/core/class/stats.class.php
+++ b/htdocs/core/class/stats.class.php
@@ -86,7 +86,7 @@ abstract class Stats
 				$year = $year - 1;
 			}
 			while ($year <= $endyear) {
-				$datay[$year] = $this->getNbByMonth($year, $format);
+				$datay[$year] = $this->_getNbByMonth($year, $format);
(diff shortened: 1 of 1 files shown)


htdocs/core/class/stats.class.php
```diff
@@ -86,7 +86,7 @@ public function getNbByMonthWithPrevYear($endyear, $startyear, $cachedelay = 0,
 				$year = $year - 1;
 			}
 			while ($year <= $endyear) {
-				$datay[$year] = $this->getNbByMonth($year, $format);
+				$datay[$year] = $this->_getNbByMonth($year, $format);
```
eldy, line 89: The method getNbByMonthWithPrevYear is never used by an instance. This class is abtracts. There is another class that inherit it and that is used to call getNbByMonthWithPrevYear().
And the getNbByMonth() is a method of this parent class.
So i think current code is good.

PR #21260, opened 2022-06-14
BUG FIX : changing function name to existing one (2)

Quatadah:
changing from getAverageByMonth() to _getAverageByMonth() because it's the one that exists.
However, this function still needs the parameter `$sql`.

### DIFF
--- a/htdocs/core/class/stats.class.php
+++ b/htdocs/core/class/stats.class.php
@@ -220,6 +220,13 @@ abstract class Stats
 	}
 
 	/**
+	 * @param	int		$year			year number
+	 * @param	int 	$format			0=Label of abscissa is a translated text, 1=Label of abscissa is month number, 2=Label of abscissa is first letter of month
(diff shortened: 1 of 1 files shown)


eldy:
Same than #21259

htdocs/core/class/stats.class.php
```diff
@@ -219,6 +219,13 @@ public function getAmountByMonthWithPrevYear($endyear, $startyear, $cachedelay =
 		return $data;
 	}
 
+	/**
+	 * @param	int		$year			year number
+	 * @param	int 	$format			0=Label of abscissa is a translated text, 1=Label of abscissa is month number, 2=Label of abscissa is first letter of month
+	 * @return 	int						value
+	 */
+	protected abstract function getAverageByMonth($year, $format = 0);
```
eldy, line 227: Signature/parameetrs of the added abstract method seems wrong...

PR #21261, opened 2022-06-14
BUG FIX : changing function name to the exisiting one (3)

Quatadah:
changing from getAmountByMonth() to _getAmountByMonth() because it's the one that exists.
However, this function still needs the parameter `$sql`.


eldy:
Same than  #21259

PR #21263, opened 2022-06-14
BUG FIX FetchLines does not exist on class emailcollector action but FetchLinesComon does

fboitel:


htdocs/emailcollector/class/emailcollectoraction.class.php
```diff
@@ -272,7 +272,7 @@ public function fetch($id, $ref = null)
 	{
 		$result = $this->fetchCommon($id, $ref);
 		if ($result > 0 && !empty($this->table_element_line)) {
-			$this->fetchLines();
+			$this->fetchLinesCommon();
```
eldy, line 275: Property table_element_line does not exists (is not set) in class on emailcollectoraction.class.php
So test will always be false and we never call the fetchLines.    The emailcollectoraction class has not sublines.
So it would be better to comment the line 274, 275 and 276

PR #21264, opened 2022-06-14
BUG FIX Attribute datee doesn't exist on class Task

fboitel:


htdocs/projet/class/task.class.php
```diff
@@ -2304,7 +2304,7 @@ public function hasDelay()
 
 		$now = dol_now();
 
-		$datetouse = ($this->date_end > 0) ? $this->date_end : ((isset($this->datee) && $this->datee > 0) ? $this->datee : 0);
+		$datetouse = ($this->date_end > 0) ? $this->date_end : 0;
```
eldy, line 2307: We can find a lot of reference to ->datee into code. This means this property is still in used.
So for backward compatibility, it is better to declare the property datee and set it with status "@deprecated". All properties datee may be removed in same step but later (such actions are never done during a beta but during the alpha period).

PR #21265, opened 2022-06-14
NEW: helper functions for dates + small demo case

atm-florianm:
# NEW helper functions for dates

Adds 2 functions: `GETPOSTDATE()` and `buildParamDate()`. These functions could help make code more readable and remove potential for typos (when we have to use the same prefix 3 or 6 times, we sometimes make typos or copy-paste bloopers).

**As an illustration**, I included in this PR a modification to the list of supplier invoices to show how the functions can make code more readable and less prone to small copy-paste errors.

I don't expect the PR to be merged as is (there are probably things that need improving or options to take into account, for instance I am wondering if it would be better if `buildParamDate` took a timestamp as a parameter instead of calling `GETPOST`). However I feel confident that, once it is ready, it could improve programmer experience with Dolibarr :slightly_smiling_face: .

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -11110,3 +11110,34 @@ function dolForgeCriteriaCallback($matches)
 
 	return $db->escape($operand).' '.$db->escape($operator)." ".$tmpescaped;
 }
+
+/**
+ * Helper function that combines values of a dolibarr DatePicker (such as Form::selectDate) for year, month, day (and
+ * optionally hour, minute, second) fields to return a timestamp.
+ *
+ * @param string $prefix Prefix used to build the date selector (for instance using Form::selectDate)
+ * @param bool $useHourTime If true, will also include hour, minute, second values from the HTTP request
+ * @param string $gm Passed to dol_mktime
+ * @return int|string  Date as a timestamp, '' or false if error
+ */
+function GETPOSTDATE($prefix, $useHourTime = false, $gm = 'auto') {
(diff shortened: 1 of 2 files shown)


eldy:
I have some stickler error. Do you have enabled the git hook in your IDE ? See https://wiki.dolibarr.org/index.php?title=Environment_and_development_tools#Git_Hooks

atm-florianm:
> Do you have enabled the git hook in your IDE ? See https://wiki.dolibarr.org/index.php?title=Environment_and_development_tools#Git_Hooks

Sorry, I had enabled it last year, but I made the mistake to make it default for every new git repo (including non-Dolibarr projects which have other rules) so I disabled it… long story short, I have re-enabled it for my Dolibarr v16 project so I should no longer bother you with formatting issues from now on :-) [note to myself: I should also re-enable it for all my other Dolibarr projects]

atm-florianm:
I set the PR to Draft because after discussing with John, I feel that the function parameters will need to be reworked a bit to make sure that the function's parameters don't grow over time as we need more specific options.

atm-florianm:
I set the PR back to "ready for review". I talked to John and we think that, should the need for more flags arise, we could add an associative array parameter called `$flags` or `$options` for better control over how conversions are made (same principle as `$params` in `dolGetButtonAction`), but it is not currently necessary.

eldy:
I have a merge conflict on github, can you repush pr ?

atm-florianm:
I have opened a new PR #23483 with the same content, up to date with the current develop.

PR #21275, opened 2022-06-14
FIX : loadExpedition() called on wrong object

jyhere:


htdocs/expedition/card.php
```diff
@@ -1045,7 +1045,7 @@
 			print '<table class="noborder centpercent">';
 
 			// Load shipments already done for same order
-			$object->loadExpeditions();
+			$objectorder->loadExpeditions();
```
eldy, line 1048: $object is defined on line 846 with $object = new $classname($db);
and $classname is "Commande" so $object seems good to execute loadExpeditions().
This may be a false warning of your IDE or do you experience bug ?

PR #21276, opened 2022-06-14
FIX bug :The properties product_tosell & product_tobuy do not exist on OrderLine

badre-agtaib:
# FIX bug :The properties product_tosell & product_tobuy do not exist on OrderLine
The properties product_tosell & product_tobuy are related to ExpeditionLine Class

### DIFF
--- a/htdocs/commande/class/commande.class.php
+++ b/htdocs/commande/class/commande.class.php
@@ -2060,7 +2060,7 @@ class Commande extends CommonOrder
 		$sql .= ' l.total_ht, l.total_ttc, l.total_tva, l.total_localtax1, l.total_localtax2, l.date_start, l.date_end,';
 		$sql .= ' l.fk_unit,';
 		$sql .= ' l.fk_multicurrency, l.multicurrency_code, l.multicurrency_subprice, l.multicurrency_total_ht, l.multicurrency_total_tva, l.multicurrency_total_ttc,';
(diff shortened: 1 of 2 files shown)


htdocs/commande/class/commande.class.php
```diff
@@ -2060,7 +2060,7 @@ public function fetch_lines($only_product = 0, $loadalsotranslation = 0)
 		$sql .= ' l.total_ht, l.total_ttc, l.total_tva, l.total_localtax1, l.total_localtax2, l.date_start, l.date_end,';
 		$sql .= ' l.fk_unit,';
 		$sql .= ' l.fk_multicurrency, l.multicurrency_code, l.multicurrency_subprice, l.multicurrency_total_ht, l.multicurrency_total_tva, l.multicurrency_total_ttc,';
-		$sql .= ' p.ref as product_ref, p.description as product_desc, p.fk_product_type, p.label as product_label, p.tosell as product_tosell, p.tobuy as product_tobuy, p.tobatch as product_tobatch, p.barcode as product_barcode,';
```
eldy, line 2063: We want to get this properties so we will avoid sql access to get them when we need them. They are the status of product so we need them often, everywhere we show a product ref or its tooltip.
So we must keep them.
A better fix would be to declare the property at the same place where ->product_label and ->product_desc are defined.

PR #21282, opened 2022-06-15
NEW Fetch product by ID with consideration of entities

almisbahinfo:
When we fetch a product by ID it return all related objects in all entities. We added a rule to take entities in consideration.

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -2338,7 +2338,7 @@ class Product extends CommonObject
 		}
 
 		if ($id) {
-			$sql .= " WHERE p.rowid = ".((int) $id);
+			$sql .= " WHERE p.rowid = ".((int) $id)." AND p.entity IN (".getEntity($this->element).")";
 		} else {
 			$sql .= " WHERE p.entity IN (".getEntity($this->element).")";
 			if ($ref) {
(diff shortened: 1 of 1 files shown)


htdocs/product/class/product.class.php
```diff
@@ -2339,6 +2339,7 @@ public function fetch($id = '', $ref = '', $ref_ext = '', $barcode = '', $ignore
 
 		if ($id) {
 			$sql .= " WHERE p.rowid = ".((int) $id);
+			$sql .= " AND p.entity IN (".getEntity($this->element).")";
```
eldy, line 2342: When we make a fetch using the id, it means we want a give product.
We don't need to add test on entity. If you add this it means we won't be able to use the fetch to get a product when we are in a context where we must ignore entity (for example, when we use the fetch to get product for migration purpose into a batch, of into a tool to move a product from one entity into another one).
What is the motivation of adding the test ?

almisbahinfo: It's appeared when we want to check if a product exists by id, but product must live in the entity in which we are connected.
example: we want to import an orders list and order lines related to this order with fk_product in a given entity.

hregis: You can't have the same id on different entities, filtering by entity will only bring errors.

almisbahinfo:
I see what do you want to say, thank you anyway.

PR #21286, opened 2022-06-15
BUG FIX aray implicitly converted to boolean

fboitel:


htdocs/opensurvey/card.php
```diff
@@ -404,7 +404,7 @@
 // Comment list
 $comments = $object->getComments();
 
-if ($comments) {
+if (!empty($comments)) {
```
eldy, line 407: Note: Because, just after there is a "foreach", a better test would be
`if (!empty($comments)) && is_array($comments)) {`

PR #21287, opened 2022-06-15
BUG FIX : Adding an abstract method "updateline" to an abstract class "CommonObject"

Quatadah:


htdocs/core/class/commonobject.class.php
```diff
@@ -2417,6 +2417,37 @@ public function setMulticurrencyCode($code)
...
+	 *	@param      int			$date_start       	Start date of the line
+	 *	@param      int			$date_end         	End date of the line
+	 *  @param		array		$array_options		extrafields array
+	 * 	@param 		string		$fk_unit 			Code of the unit to use. Null to use the default one
+	 * 	@param		double		$pu_ht_devise		Unit price in currency
+	 * 	@param		int			$notrigger			disable line update trigger
+	 * @param       integer $rang   line rank
+	 *  @return     int     		        		0 if OK, <0 if KO
+	 */
+	public abstract function updateline($rowid, $pu, $qty, $remise_percent, $txtva, $txlocaltax1 = 0.0, $txlocaltax2 = 0.0, $desc = '', $price_base_type = 'HT', $info_bits = 0, $special_code = 0, $fk_parent_line = 0, $skip_update_total = 0, $fk_fournprice = 0, $pa_ht = 0, $label = '', $type = 0, $date_start = '', $date_end = '', $array_options = 0, $fk_unit = null, $pu_ht_devise = 0, $notrigger = 0, $rang = 0);
```
eldy, line 2449: A search on 
function updateline(
show that on current code, we have a lot of call to local method updateline of class that implement CommonObject, but with different paramaters.
So adding this abstract method make a lot of existing code in error.

So i think best solution for the moment is to not add this method.

PR #21288, opened 2022-06-15
BUG FIX : Adding an abstract method 'delete' to the abstract class 'CommonObject'

Quatadah:


eldy:
Take a look at the error reported by travis CI. Adding the abstract method need also to fix the old call that was calling the method with the bad number of parameters...

PR #21291, opened 2022-06-15
BUG FIX : Temporary transformation (from string to array) so it can be traversable

Quatadah:
$data is a string variable, we cannot traverse it using `foreach`. The solution here was meant to transform it temporarily to an array `$dataArray`.


htdocs/core/class/extrafields.class.php
```diff
@@ -1268,7 +1268,8 @@ public function showInputField($key, $value, $moreparam = '', $keysuffix = '', $
 					require_once DOL_DOCUMENT_ROOT.'/categories/class/categorie.class.php';
 					$data = $form->select_all_categories(Categorie::$MAP_ID_TO_CODE[$InfoFieldList[5]], '', 'parent', 64, $InfoFieldList[6], 1, 1);
 					$out .= '<option value="0">&nbsp;</option>';
-					foreach ($data as $data_key => $data_value) {
+					$dataArray = str_split($data); // as $data is not traversable as it's a string
```
eldy, line 1271: Do you have an exemple of value $data that you get and that need to be converted into array ?
I think if you got a non array here, it means we should not make the loop at all.
So i would fix this like this
if (is_array($dataArray)) {

}

PR #21297, opened 2022-06-16
NEW add attached file in presend email form of thirdparty card

lvessiller-opendsi:
NEW add attached file in presend email form of thirdparty card

When you set an email template and you want to add an attached file : 
image

When you apply this template in presend email form of thirparty card, you have the last generated file automatically attached : 
image


eldy:
Currently the field "Join file" into the email template means "Join the main doc of the document".
So for an invoice it is the PDF of invoice, for an order the PDF of an order, etc.. The code is using dol_most_recent_file() but with the ref into the name of the file searched (second parameter is preg_quote($ref, '/').'[^\-]+'), so we are sure to take the main document and not other attached files. The tooltip on "Join file" call it is the "main" document.

This main doc file may be different than the last attached file.
For a third-parties, there is no main document. Imagine template is "Send a remind to customer for being late" and last document that was generated is a document generated for your lawyer or accountant with information about the customer. You will send this document to the customer. This feature might be dangerous.

So I suggest to introduce another value, for example value 2 instead of 1 to say 
1=join main doc if applicable  (so works for document that has a "main" doc, like invoice, order, etc..)
2=join last uploaded file.

So it is possible for user to choose and understand the rule the is behind the choice of the document that is attached.

lvessiller-opendsi:
@eldy 
I corrected it and I only attach the main document as you suggested in step 1, and  I keep the same behavour used in others cards.
For "2. join last uploaded file" value, do you suggest to introduce a new parameter in all email templates ?
Could you merge it at first ?

lvessiller-opendsi:
@eldy 
I corrected travis error, could you check and merge it ?

eldy:
> @eldy I corrected it and I only attach the main document as you suggested in step 1, and I keep the same behavour used in others cards. For "2. join last uploaded file" value, do you suggest to introduce a new parameter in all email templates ? Could you merge it at first ?

No, not introducing a new parameter, just using the same one, but instead of value 1, user will be able to enter value 2 instead.

PR #21303, opened 2022-06-17
BUG FIX Some attributes were missing on class ActionsCardProduct

fboitel:


eldy:
@fboitel Take a look at my fix 3179fd13d8d328ddabbaf76bef7062ae3da6639b
The property we want to use are in fact the one defined into this->object set on line 8 with $this->object = $tmpobject;  (and tmpobject is an object loaded just before).
For information, the injection of the dependency is done into htdocs/product/card.php at line 161 when calling getCanvas()
It is in this file that the class (that the code after will depends on) is loaded dynamically, so we say that the code in
htdocs/product/card.php   depends on  product/canvas/XXX/actions_card_XXX.class.php  but we don't know it at dev time. We know it only at line 156 of htdocs/product/card.php (after knowing that the var $canvas = XXX).

PR #21306, opened 2022-06-17
BUG FIX attribute total_tva set deprecated declaration of total_vat o…

fboitel:


eldy:
Name of property total_tva was previously using a lot of different name. So we recently standardize the name into total_tva.
Because this was done recently we will still wait one or 2 version before starting the migration of total_tva into the english name total_vat. So i will close this PR, this will be done by a old core team member in v18 or +

PR #21317, opened 2022-06-18
Filter accounting reports by projects

ymollard:
# NEW|New Allow accounting reports to filter output by project
When **comptabilite** or **accounting** module is activated, a new button in the project card allows to export an accounting report with attached files.

### DIFF
--- a/htdocs/compta/accounting-files.php
+++ b/htdocs/compta/accounting-files.php
@@ -47,6 +47,11 @@ require_once DOL_DOCUMENT_ROOT.'/compta/bank/class/paymentvarious.class.php';
 require_once DOL_DOCUMENT_ROOT.'/core/class/html.formfile.class.php';
 require_once DOL_DOCUMENT_ROOT.'/loan/class/paymentloan.class.php';
 
+if (!empty($conf->project->enabled)) {
+	require_once DOL_DOCUMENT_ROOT.'/projet/class/project.class.php';
+	require_once DOL_DOCUMENT_ROOT.'/core/class/html.formprojet.class.php';
+}
+
 // Constant to define payment sens
 const PAY_DEBIT = 0;
 const PAY_CREDIT = 1;
@@ -64,6 +69,7 @@ $date_stopMonth = GETPOST('date_stopmonth', 'int');
 $date_stopYear = GETPOST('date_stopyear', 'int');
 $date_stop = dol_mktime(23, 59, 59, $date_stopMonth, $date_stopDay, $date_stopYear, 'tzuserrel');
 $action = GETPOST('action', 'aZ09');
+$projectid = (GETPOST('projectid', 'int') ? GETPOST('projectid', 'int') : 0);
 
 // Initialize technical object to manage hooks of page. Note that conf->hooks_modules contains array of hook context
 $hookmanager->initHooks(array('comptafileslist', 'globallist'));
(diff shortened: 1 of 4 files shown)


htdocs/compta/accounting-files.php
```diff
@@ -175,6 +181,7 @@
 			$sql .= " WHERE datef between ".$wheretail;
 			$sql .= " AND t.entity IN (".$db->sanitize($entity == 1 ? '0,1' : $entity).')';
 			$sql .= " AND t.fk_statut <> ".Facture::STATUS_DRAFT;
+			if(!empty($projectid)) $sql .= " AND fk_projet = ".$db->sanitize($projectid);
```
eldy, line 184: When the param in sql is an int, the sanitization is done like this:
`if(!empty($projectid)) $sql .= " AND fk_projet = ".((int) $projectid);`
db->sanitize is used to sanitize a string by escaping the ' so not enough to avoid sql injection if the param is not inside ' '.

htdocs/compta/accounting-files.php
```diff
@@ -240,9 +251,10 @@
 			$sql .= " FROM ".MAIN_DB_PREFIX."payment_various as t";
 			$sql .= " WHERE datep between ".$wheretail;
 			$sql .= " AND t.entity IN (".$db->sanitize($entity == 1 ? '0,1' : $entity).')';
+			if(!empty($projectid)) $sql .= " AND fk_projet = ".$db->sanitize($projectid);
 		}
 		// Loan payments
-		if (GETPOST('selectloanspayment') && !empty($listofchoices['selectloanspayment']['perms'])) {
+		if (GETPOST('selectloanspayment') && !empty($listofchoices['selectloanspayment']['perms']) and empty($projectid)) {
```
eldy, line 257: For code standardization, use the "&&" for the and in PHP. We use the "AND" for the and of SQL.

htdocs/compta/accounting-files.php
```diff
@@ -460,7 +472,19 @@
 	$log .= ','.$langs->transnoentitiesnoconv("Country");
 	$log .= ','.$langs->transnoentitiesnoconv("VATIntra");
 	$log .= ','.$langs->transnoentitiesnoconv("Sens")."\n";
-	$zipname = $dirfortmpfile.'/'.dol_print_date($date_start, 'dayrfc', 'tzuserrel')."-".dol_print_date($date_stop, 'dayrfc', 'tzuserrel').'_export.zip';
+	$zipname = $dirfortmpfile.'/'.dol_print_date($date_start, 'dayrfc', 'tzuserrel')."-".dol_print_date($date_stop, 'dayrfc', 'tzuserrel');
+	if(!empty($projectid)) {
+		$sql = 'SELECT t.ref FROM '.MAIN_DB_PREFIX.'projet as t';
```
eldy, line 477: A better solution would be do to :

$project = new Project($db);
$project->fetch($projectid);
if ($project->ref) {
    $zipname .= '_'.$project->ref;
}

So there is no sql at all inside the php code.

htdocs/compta/accounting-files.php
ymollard: Oops, the AND is a reflex from a pythonista :D
Fixed!

ymollard:
This PR is fixed @eldy

PR #21321, opened 2022-06-20
NEW : Create llx_element_categorie.sql

defrance:
The goal of table llx_element_categorie is to replace all tables llx_categories_xxx (with one table).
the llx_element_tag table initially intended for this does not have the correct nomenclature to express its real use (and was used for other things before)
more informations here : https://github.com/Dolibarr/dolibarr/pull/21192

# Instructions
*Please:*

[*Long description*]

[*Long description*]

[*Long description*]


htdocs/install/mysql/tables/llx_element_categorie.sql
```diff
@@ -0,0 +1,25 @@
...
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU General Public License for more details.
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program. If not, see <https://www.gnu.org/licenses/>.
+--
+-- ============================================================================
+
+create table llx_element_categorie
```
eldy, line 19: I suppose goal is to replace the llx_element_tags that was introduced for generic storing of link between categories/tag and object. 
This is possible because the llx_element_tags is not yer used by core code. But in such a case, can you include in same PR, the removal of files that create table llx_element_tags so we won't keep duplicate tables.

defrance: I don't prefer to delete this table (llx_element_tags) which has been used by one of my modules for more than 7 years...
Adding its deletion in the installation of dolibarr would result in my customers losing their data present in it
The llx_element_tags table has never been used in native dolibarr modules (this is why it was removed with V12 (and then added with another structure in V14).
I already did a PR putting the structure of llx_element_tags back to its old structure (which is the format I use).

Another solution would be to condition its deletion only if it is empty...
--
Je ne préfère pas  supprimer cette table (llx_element_tags) qui est utilisée par l'un de mes modules depuis plus de 7 ans...
Ajouter sa suppression dans l'installation de dolibarr aurait pour conséquence de faire  perdre à mes clients leurs données présente dans celle-ci
La table llx_element_tags n'a jamais été utilisé dans les modules natifs de dolibarr (c'est pour cette raison qu'elle avait été supprimé avec la V12 (et rajouté ensuite avec une autre structure dans la V14).
J'ai déjà fait un PR remettant la structure de llx_element_tags dans son ancienne structure (qui est le format que j'utilise).

Une autre solution serai de conditionner sa suppression uniquement si elle  vide...

defrance: oki, i suppress it from tables folders

htdocs/install/mysql/tables/llx_element_tag.key.sql
```diff
@@ -1,21 +0,0 @@
--- ============================================================================
```
eldy, line 1: It seems this file is removed instead of being renamed like it is for llx_element_tag.

defrance: in fact i forgot to create llx_element_categorie.key.sql, it's done

Issue #21320, opened 2022-06-20
Add a select-box as an alternative to textbox when entering "customer reference" on orders / invoices / etc


bos4711:
### Feature Request

When we create sales orders, we add customer name as reference for 99% of the time. Instead of typing the names all the time it would be neat to just use a select-box as all customer names are added in “Contacts” for the third-party.

### Use case

_No response_

### Suggested implementation

My own quick-hack in htdocs/commande/card.php just to show the idea:

```
(near line 1590)
// Reference client                                                                                                                                                                                                                                                
print '<tr><td>'.$langs->trans('RefCustomer').'</td><td>';
$ref_test_foo = "Select reference:" . $form->selectcontacts($soc->id, $contactid, 'ref_client', 1, $srccontactslist, '', 0);
if (!empty($conf->global->MAIN_USE_PROPAL_REFCLIENT_FOR_ORDER) && !empty($origin) && !empty($originid)) {
    print '<input type="text" name="ref_client" value="'.$ref_client.'">' . $ref_test_foo . '</td>';
} else {
    print '<input type="text" name="ref_client" value="'.GETPOST('ref_client').'">' . $ref_test_foo . '</td>';
}
print '</tr>';
```

![2022-06-19_195252](https://user-images.githubusercontent.com/11036018/174535456-f29bb072-dd6b-467d-8847-a8fe983aa188.png)
![2022-06-19_195332](https://user-images.githubusercontent.com/11036018/174535457-2e557f52-2c24-4eb6-88df-b3edc876ab26.png)

(The problem here is that $form->selectcontacts() returns an integer-id for the contact instead of the name, but the id is added to "ref_client" which shows that this idea is doable).

### Suggested steps

_No response_

PR #21330, opened 2022-06-20
NEW #21320 

fboitel:
Possibility in a new order to select quickly the client name as client reference


eldy:
We arleady have a customer field and a section to set contact of the customer available as a common feature. So why adding this information into the customer reference of order. Because the data of client name is already recorded into other fields (as a contact of order), i suggest to keep the field empty instead.

PR #21331, opened 2022-06-20
Errors in Calculation of margins on lines of credit notes when total_ht is 0

deevando:
Hello, it is the first time that I upload modifications that I think can solve problems, if there is something that I do not do correctly, you can tell me. Thanks.
Everything you send is based on the program for the Spanish version (Spain)

**Errors in Calculation of margins**

If the type of invoice is a credit note, we must use "total_ht<=0" in the conditional
If the type of invoice is normal, we must use the conditional “total_ht<0”
We have an invoice with the following structure:
Sell (total_ht)    Purchase     Margin
      15                  10                5
       0                  10              -10
     30                  10              20
Total                Total            Total
    45                  30               15

When making the credit note it should look like this:
Sell(total_ht)    Purchase     Margin
     -15                  10               -5
        0                  10                10
     -30                 10               -20
 Total              Total             Total
    -45                 30                -15

If we use in credit note “total_ht<0”
total_ht < 0  True —>   -1 * (abs(-15) - (10 * 1)) = -1 * (15 - 10) = -5  
**total_ht < 0  Falso—> 0 - (10 * 1)    = – 10  (error)**

if it is a credit note{
 if total_ht<= 0 True —> -1 * (abs(0) - (10 * 1)) = -1 * -10 = 10
}
total_ht < 0  True —> -1 * (abs(-30) - (10 * 1)) = -1 * (30-10) = -1 * 20 = -20

If we want total_ht < 0 False—> 0 - (10 * 1) = – 10 to be correct, we have to validate if it is a credit note and if it is, apply “total_ht <=0)” so that the result is correct.
A product line can have a sale price of 0 and in the credit invoice, 0 , generates an error in the calculation.

If in the credit invoice the product line has a sales amount of 0 and cost X gave a negative margin

**Error margin rectified invoices**

There is also a problem in the calculation of margins that are doubled when there is a normal invoice and we make a corrected invoice and the corrected invoice is not added.

$sql .= " AND f.fk_status <> 3"; //Do not calculate the margin of rectified - replaced invoices

### DIFF
--- a/htdocs/accountancy/customer/index.php
+++ b/htdocs/accountancy/customer/index.php
@@ -377,6 +377,7 @@ if (!empty($conf->global->ACCOUNTING_DATE_START_BINDING)) {
 	$sql .= " AND f.datef >= '".$db->idate($conf->global->ACCOUNTING_DATE_START_BINDING)."'";
 }
 $sql .= " AND f.fk_statut > 0";
+$sql .= " AND f.fk_statut <> 3"; //Do not calculate the margin of rectified - replaced invoices
 $sql .= " AND fd.product_type <= 2";
 $sql .= " AND f.entity IN (".getEntity('invoice', 0).")"; // We don't share object for accountancy
 $sql .= " AND aa.account_number IS NULL";
@@ -494,6 +495,7 @@ if (!empty($conf->global->ACCOUNTING_DATE_START_BINDING)) {
 }
 $sql .= " AND f.entity IN (".getEntity('invoice', 0).")"; // We don't share object for accountancy
 $sql .= " AND f.fk_statut > 0";
+$sql .= " AND f.fk_statut <> 3"; //Do not calculate the margin of rectified - replaced invoices
 $sql .= " AND fd.product_type <= 2";
 if (!empty($conf->global->FACTURE_DEPOSITS_ARE_JUST_PAYMENTS)) {
 	$sql .= " AND f.type IN (".Facture::TYPE_STANDARD.", ".Facture::TYPE_REPLACEMENT.", ".Facture::TYPE_CREDIT_NOTE.", ".Facture::TYPE_SITUATION.")";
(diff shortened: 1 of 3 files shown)


htdocs/accountancy/customer/index.php
```diff
@@ -661,6 +671,7 @@
 		}
 		$sql .= " AND f.entity IN (".getEntity('invoice', 0).")"; // We don't share object for accountancy
 		$sql .= " AND f.fk_statut > 0";
+		$sql .= " AND f.fk_statut <> 3"; //Do not calculate the margin of rectified - replaced invoices
```
eldy, line 674: Instead of removing replacement invoice, we should remove the "replaced" invoice ? so  
`AND (close_code IS NULL OR close_code != Facture::CLOSECODE_REPLACED)`

deevando: Sorry if my English is not very good, I'm Spanish,

 
In "index.php?leftmenu=accountancy_dispatch_customer&mainmenu=accountancy" 
we can see accounting invoices and margins. If there is a replaced invoice, the two are added in margins when only the new invoice that has replaced the erroneous one should be taken into account. In Spain, if there is an error on an invoice, there is up to 4 years to be able to make a corrected invoice.

"When should they be issued?

The rectification will be made as soon as the person obliged to issue the invoice is aware of the circumstances that require its issuance, provided that four years have not elapsed from the moment in which the tax was accrued or, as the case may be, from the date on which that the circumstances regulated in article 80 of Law 37/1992 have occurred, which give rise to the modification of the VAT tax base." 

This is the link in of correctives invoice in spain:
https://sede.agenciatributaria.gob.es/Sede/iva/facturacion-registro/facturacion-iva/facturas-rectificativas.html

deevando: Its ok.  Works correctly

                $sql .= " AND f.entity IN (".getEntity('invoice', 0).")"; // We don't share object for accountancy
		$sql .= " AND f.fk_statut > 0";
		//$sql .= " AND f.fk_statut <> 3"; //Do not calculate the margin of rectified - replaced invoices
		$sql .= " AND (f.close_code IS NULL OR f.close_code != '".Facture::CLOSECODE_REPLACED."')";
		$sql .= " AND fd.product_type <= 2";

When the pull request is closed, does it mean that it is not taken into account and that it is incorrect? because the calculation of margins works correctly with the previous changes.
Or should I send another with this latest modification? Thanks.

deevando:
This is an example how margin is not working correctly in dolibarr 15.0.2. 
The modificacions do it in the pull request are working in my dolibarr installation correctly but was closed. 
Only i had to change line 101 with this one.

if ($object->element == 'facture' && $object->type == $object::TYPE_CREDIT_NOTE) {

Captura de Pantalla 2022-07-08 a las 23 28 39

This is an invoice with unit price 0 and only one product has all the total price. 
Here you can see that margin is correct. 

Captura de Pantalla 2022-07-08 a las 23 18 18

This is the credit invoice and here calculations are incorrect. cause 0 value in price make fault in the calcules.

Margin and total buying price are incorrect. 
Captura de Pantalla 2022-07-08 a las 23 19 24


Captura de Pantalla 2022-07-08 a las 23 20 07

In the invoce margin are: 145,34 € and total buy price: 852,09€
In the credit note margin are: -1695,50€ when must be -145,34€  and total buy price is 698,07€ and must be 852,09€

This error is important!

Please check againg the modifications that is working correctly in margin calculate with credit note in all situations and the replace invoice was resolved too in the avoid duplicate margin in the file was upload in accountancy/customer/index.php

Thanks.

PR #21332, opened 2022-06-21
FIX: Missing hook in select_produits_fournisseurs_list

noec764:
Add Hook in _select_produits_fournisseurs_list_ in order to add specific text in product selector of _fourn/commande/card.php_

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -3387,8 +3387,6 @@ class Form
 				$opt .= $optlabel;
 				$outval .= $outvallabel;
 
-				$hookmanager->initHooks(array('ordersuppliercard'));
-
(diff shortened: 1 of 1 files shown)


htdocs/core/class/html.form.class.php
```diff
@@ -3387,6 +3387,14 @@ public function select_produits_fournisseurs_list($socid, $selected = '', $htmln
 				$opt .= $optlabel;
 				$outval .= $outvallabel;
 
+				$hookmanager->initHooks(array('ordersuppliercard'));
```
eldy, line 3390: The initHooks must be done by the main page. 
So when select_produits_fournisseurs_list is called by page a, context is a, when called by page b, context os b.

noec764:
I will redo this PR in order to test the result of the hook and allow more freedom when it comes to possible actions in the hook.

PR #21334, opened 2022-06-21
clone button on tasks

Quatadah:
Is there any changes I can add to improve the clone button functionnality ?
+ I made changes projects.lang files but I didn't push them. should I ?


htdocs/projet/tasks/task.php
```diff
@@ -619,9 +660,24 @@
 		$reshook = $hookmanager->executeHooks('addMoreActionsButtons', $parameters, $object, $action); // Note that $action and $object may have been
 		// modified by hook
 		if (empty($reshook)) {
+			// Cancel
+			if ($cancel) {
```
eldy, line 664: Can you confirm we need this delete ?  If yes, can you describe the use case (steps in GUI when you reach this code) ?

atm-maxime:
Sweet feature, thanks !

htdocs/projet/tasks/task.php
Quatadah: The way the code is implemented, this block is needed, as the user may cancel the cloning of the task.
Way to get to this block: 
- Click on a task
- Clone it
- Instead of saving the modifications, try to cancel. If we didn't had that block, the task would still have be cloned.

PR #21345, opened 2022-06-22
New invoice rec default to yes

atm-steve:
Add a conf to set default select value to yes on invoice rec


htdocs/core/tpl/objectline_create.tpl.php
```diff
@@ -348,6 +348,10 @@
 		if (!empty($conf->service->enabled) && ($object->element == 'facturerec' || $object->element == 'invoice_supplier_rec')) {
 			echo '<div class="divlinefordates"><br>';
 			echo $langs->trans('AutoFillDateFrom').' ';
+			if ($conf->global->INVOICE_REC_DATE_TO_YES) {
```
frederic34, line 351: `if (!empty($conf->global->INVOICE_REC_DATE_TO_YES)) {`
if not defined, it will make warning with php8

PR #21348, opened 2022-06-22
FIX: 15 fix graph ficheinter status

FHenry:
All status are not displayed, and code was over complicated for simple data array

DataSeries
/var/www/html/fichinter/index.php:135:
array (size=4)
  0 => 
    array (size=2)
      0 => string 'Brouillon' (length=9)
      1 => int 195
  1 => 
    array (size=2)
      0 => string 'Validé' (length=7)
      1 => int 620
  2 => 
    array (size=2)
      0 => string 'Clos' (length=4)
      1 => int 32243
  3 => 
    array (size=2)
      0 => string 'Done' (length=4)
      1 => int 0

Before
image

After
image

### DIFF
--- a/htdocs/fichinter/index.php
+++ b/htdocs/fichinter/index.php
@@ -98,7 +98,6 @@ if ($resql) {
 	// -1=Canceled, 0=Draft, 1=Validated, 2=Accepted/On process, 3=Closed (Sent/Received, billed or not)
 	if ($num>0) {
 		while ($row = $db->fetch_row($resql)) {
-
 			if (!isset($vals[$row[1]])) {
 				$vals[$row[1]] = 0;
 			}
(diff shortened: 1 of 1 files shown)


FHenry:
May be we can remove Status 2 (label "Close" but Fichinter::STATUS_BILLED) because most intervention will be at this status.
If this graph have to show current operating indicators (like "only what we have to do"), so we remove also Status 3 (label "Done" but Fichinter::STATUS_CLOSED) from this graph. Or we display all of them properly

By the way Status 3 (label "Done" Fichinter::STATUS_CLOSED) is not used by ficheinter card unless hidden option FICHINTER_CLASSIFY_BILLED So should we display this status base on this conf ?

eldy:
We should never show the last status (when workflow is finished). So we should exclude STATUS_CLOSED
The STATUS_BILLED is an intermediate status between VALIDATED and CLOSED, we must show it only of FICHINTER_CLASSIFY_BILLED is set.

See https://wiki.dolibarr.org/index.php/Module_Fiches_d%27interventions

FHenry:
Ok better like that

image


and with FICHINTER_CLASSIFY_BILLED

image

PR #21354, opened 2022-06-22
FIX #21204

### LINKED ISSUE #21204: Add option so that Trigger PROPAL_CLOSE_SIGNED in modWorkflow creates an order only if there is none
### Feature Request

Add on option, so that when workflow module is enabled and we sign a propal, it checks if an order is already associated with the propal before creating one

### Use case

When I have to change things in an already signed propal, I reopen it, change things, and put it back signed, but in that case, if an order already exists and module workflow is activated with option WORKFLOW_PROPAL_AUTOCREATE_ORDER , a new order is created, and I don't want this.

### Suggested implementation

Add an option (for example WORKFLOW_PROPAL_AUTOCREATE_ORDER_IFNOTEXISTS) in module workflow setup and change a few things in Trigger (in file interface_20_modWorkflow_WorkflowManager.class.php) to check if an order is already associated with propal, and so not create another.

### Suggested steps

_No response_

badre-agtaib:
# NEW #21204

### DIFF
--- a/htdocs/admin/workflow.php
+++ b/htdocs/admin/workflow.php
@@ -65,6 +65,12 @@ $workflowcodes = array(
 		'enabled'=>(!empty($conf->propal->enabled) && !empty($conf->commande->enabled)),
 		'picto'=>'order'
 	),
+	'WORKFLOW_PROPAL_NOTCREATE_ORDER_IFEXISTS'=>array(
+		'family'=>'create',
+		'position'=>15,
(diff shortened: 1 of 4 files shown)


eldy:
I suggest a different implementation.
We do not add another option. We keep the old option but we enhance it:  If 1 link to an order already exists, we do not recreate the order, instead we show a warning "An order was already open linked to this proposal, so no other order was created automatically".

PR #21363, opened 2022-06-23
adding dropdown menu in case of a lot of buttons in project

Quatadah:
Transforming a lot of buttons into a dropdown menu.
image

### DIFF
--- a/htdocs/projet/card.php
+++ b/htdocs/projet/card.php
@@ -1318,46 +1318,52 @@ if ($action == 'create' && $user->rights->projet->creer) {
 
 			// Add button to create objects from project
 			if (!empty($conf->global->PROJECT_SHOW_CREATE_OBJECT_BUTTON)) {
+				print'<div class="dropdown-holder">';
+				print'<a class="dropdown-toggle butAction">'.$langs->trans("Create").'</a>';
+				print'<div class="dropdown-content">';
 				if (!empty($conf->propal->enabled) && $user->rights->propal->creer) {
 					$langs->load("propal");
-					print dolGetButtonAction('', $langs->trans('AddProp'), 'default', DOL_URL_ROOT.'/comm/propal/card.php?action=create&amp;projectid='.$object->id.'&amp;socid='.$object->socid, '');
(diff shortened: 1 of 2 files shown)


htdocs/projet/card.php
```diff
@@ -1318,46 +1318,52 @@ function change_percent()
 
 			// Add button to create objects from project
 			if (!empty($conf->global->PROJECT_SHOW_CREATE_OBJECT_BUTTON)) {
+				print'<div class="dropdown-holder">';
+				print'<a class="dropdown-toggle butAction">'.$langs->trans("Create").'</a>';
+				print'<div class="dropdown-content">';
 				if (!empty($conf->propal->enabled) && $user->rights->propal->creer) {
 					$langs->load("propal");
-					print dolGetButtonAction('', $langs->trans('AddProp'), 'default', DOL_URL_ROOT.'/comm/propal/card.php?action=create&amp;projectid='.$object->id.'&amp;socid='.$object->socid, '');
+					print '<a class="dropdown-item" href="' . DOL_URL_ROOT.'/comm/propal/card.php?action=create&amp;projectid='.$object->id.'&amp;socid='.$object->socid . '">' . $langs->trans('AddProp') . '</a>';
```
eldy, line 1326: We should keep the call to dolGetButtonAction() because it brings a central point to output a button.
It seems you need to add the dropdown-item for the feature to work. Why not adding this class in a parameter of the dolGetButtonAction. There is already the 6th parameter that can be used for this.

Quatadah: Yes you're right maybe I will just add a default parameter in that function so it can changes the class of anchor element. I didn't want to make changes in that function in the beginning but this way I think it's better !

eldy: On screenshot, we can see the space between button "Create" and "Clone" is larger than other spaces. 
Can you also check this point ?

Quatadah: Sure

Quatadah: I've done this, however, the toggling doesn't work anymore. I don't know if the css have changed, I've tried to change it too, I searched on the documentation, I'm on eldy theme, I tried changing the styles on theme/eldy but nothing changes.

eldy:
Note: Using css is better than style

PR #21373, opened 2022-06-25
NEW Enable all types of fees in expense reports

ymollard:
# NEW Enable all types of fees in expense reports
Some fee types were not enabled by default and no UI setting allowed to enable them

### DIFF
--- a/htdocs/install/mysql/data/llx_c_type_fees.sql
+++ b/htdocs/install/mysql/data/llx_c_type_fees.sql
@@ -35,7 +35,14 @@
 insert into llx_c_type_fees (code,label,active) values ('TF_OTHER',    'Other',  1);
 insert into llx_c_type_fees (code,label,active) values ('TF_TRIP',     'Transportation',   1);
 insert into llx_c_type_fees (code,label,active) values ('TF_LUNCH',    'Lunch',  1);
-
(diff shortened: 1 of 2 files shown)


atm-maxime:
Hi. I think there are doubles. Fuel, hotel, péage already exists (ex_fue, ex_hot, ex_toll...)

ymollard:
Yes, however they start with a different prefix `EX_` and are separated by a blank line from the other `TF_`. Does it mean anything?
Also most `EX_` are not active by default and can't be activated from the UI. Should be active more of them by default?

atm-maxime:
What do you mean "can't be activated from the UI" ? Differences between prefixes is not a problem.

ymollard:
> What do you mean "can't be activated from the UI" ?

There is no button so that a regular user activate them (e.g. in the module settings).

atm-maxime:
All data such as expense report types (of payment methods, currencies, ...) are managed in the dictionaries section, not in a module setup. From there you can activate/add/edit whatever value you want

ymollard:
Hum, ok.
In that case this PR has no reason to be, and i guess that I should instead edit the user doc of the HRM module to tell them how to enable the other types of expenses that are deactivated by default.

PR #21394, opened 2022-06-27
fix page titles

cfoellmann:
A lot of pages do not have a <head><title> describing the content of the page.

My change generates titles in this format:

Example `Proposal`

- new item: `$langs->trans("NewPropal");`
- list items: `$langs->trans('ListOfProposals');`
- card: `$object->ref." - ".$langs->trans('Card');`
- item tabs (example info.php): `$object->ref." - ".$langs->trans('Info');`


eldy:
Can you resend the PR on develop branch ?

PR #21401, opened 2022-06-28
NEW hook "changeHelpURL" to modify target of the help button

cfoellmann:
Use the hook `changeHelpURL` to modify the help_url parameter.

Example action:

```php
public function changeHelpURL( $parameters, &$object, &$action, $hookmanager ) {

	$context = $parameters['context'];
	
	switch ($context) {

		case 'main:mrpindex':
			$changed_url = 'https://doku.something.local/mrp';
		break;

		case 'main:index':
		default:
			$changed_url = 'https://doku.something.local/';
		break;
	}

	$hookmanager->resPrint = $changed_url;

	return 1;

}
```


eldy:
Don't forget to add NEW or FIX at begin of label of PR so line will appears into the generated ChangeLog file at release.

cfoellmann:
@eldy just renaming the PR title like this?

eldy:
> @eldy just renaming the PR title like this?

Yes

PR #21404, opened 2022-06-28
FIX - Edit propal date

BB2A-Anthony:
# FIX|Fix #[*Edit propal date*]

### DIFF
--- a/htdocs/comm/propal/class/propal.class.php
+++ b/htdocs/comm/propal/class/propal.class.php
@@ -2060,7 +2060,9 @@ class Propal extends CommonObject
 
 			$this->db->begin();
 
-			$sql = "UPDATE ".MAIN_DB_PREFIX."propal SET datep = '".$this->db->idate($date)."'";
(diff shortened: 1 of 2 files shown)


htdocs/comm/propal/class/propal.class.php
```diff
@@ -2060,7 +2060,9 @@ public function set_date($user, $date, $notrigger = 0)
 
 			$this->db->begin();
 
-			$sql = "UPDATE ".MAIN_DB_PREFIX."propal SET datep = '".$this->db->idate($date)."'";
+			$fin_validite = $date + ($this->duree_validite * 24 * 3600);
```
eldy, line 2063: If $this->duree_validite is not set and in database fin_validite is empty, we must keep field empty (unknown) in database.

BB2A-Anthony:
@eldy it's not good for you?

BB2A-Anthony:
@eldy it's not good for you?

PR #21406, opened 2022-06-28
NEW API for module reception

Quatadah:
I copied api_shipments.php file into api_receptions.php file and changed all Expedition object and fields into Reception Object.
As both objects inherits from the same Abstract class they have the same interface, and this should work.

### DIFF
--- a/htdocs/reception/class/api_receptions.class.php
+++ b/htdocs/reception/class/api_receptions.class.php
@@ -71,7 +71,7 @@ class Receptions extends DolibarrApi
 
 		$result = $this->reception->fetch($id);
 		if (!$result) {
-			throw new RestException(404, 'Shipment not found');
+			throw new RestException(404, 'Reception not found');
 		}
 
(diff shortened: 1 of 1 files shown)


eldy:
Don't forget to add the key  "NEW ..."   or "FIX ..." into the label of your PR. So the PR will be printed into the generated ChangeLog automatically (internal rule and tool of Dolibarr project). Use lowercase New and Fix if you don't want to have the PR visible into the ChangeLog.

Quatadah:
done

htdocs/reception/class/api_receptions.class.php
```diff
@@ -0,0 +1,721 @@
...
+	 * Validate fields before create or update object
+	 *
+	 * @param   array           $data   Array with data to verify
+	 * @return  array
+	 * @throws  RestException
+	 */
+	private function _validate($data)
+	{
+		$reception = array();
+		foreach (Shipments::$FIELDS as $field) {
```
carl0zen, line 713: this line is incorrect, we should replace this with Receptions::, did this and the endpoint worked

Issue #21399, opened 2022-06-28
Enhance function build_rssfile() to add image tag in RSS stream


eldy:
### Feature Request

Enhance function build_rssfile() to add image tag in RSS stream
Param $events_array can be a simple array (we can add a key "image" to provide the image and output it into RSS if defined)
or an array of object WebsitePage (in such case, we already have a property image into each object).

To test a rss, just create a website from Dolibarr CMS, create a page with type "blog" and call page 
http://myserver/wrapper.php?rss=1

eldy:
First implementation try available into v17

PR #21408, opened 2022-06-29
NEW #21399

badre-agtaib:
# NEW #21399

### DIFF
--- a/htdocs/core/lib/xcal.lib.php
+++ b/htdocs/core/lib/xcal.lib.php
@@ -301,7 +301,7 @@ function build_calfile($format, $title, $desc, $events_array, $outputfile)
  *  @param      string	$format             "rss"
  *  @param      string	$title              Title of export
  *  @param      string	$desc               Description of export
- *  @param      array	$events_array       Array of events ("uid","startdate","summary","url","desc","author","category") or Array of WebsitePage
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/xcal.lib.php
```diff
@@ -404,6 +407,10 @@ function build_rssfile($format, $title, $desc, $events_array, $outputfile, $filt
 				fwrite($fichier, "<category><![CDATA[".$category."]]></category>\n");
 				fwrite($fichier, "<description><![CDATA[");
 
+				if (!empty($image)) {
+					fwrite($fichier, '<p><img class="center" src="'.$image.'"/></p>');
```
eldy, line 411: When parameter is coming from $event_array['image'], we can say it is a public image url.
But when parameters even_array is list of WebsitePage, the property "image" is just a relative path. For example "image/template-corporate/background_dolibarr.jpg"
So the link will not works.  You should forge the url of image with 
$website->virtualhost.'/medias/'$image;   

$website is a website object parent of $websitepage.
$website->virtualhost contains for example:  https://www.dolicloud.com

badre-agtaib:
Capture d’écran du 2022-07-04 17-30-42
Capture d’écran du 2022-07-04 17-31-31

Issue #21095, opened 2022-06-01
No thirdparty created while using the public member creation form


daraelmin:
### Bug

When new member create a new adherent with the public form and pay his or her subscription (by stripe), no thirdparty is created, even if the hidden option ADHERENT_DEFAULT_CREATE_THIRDPARTY is set to 1.

So there's a facture and there's a bank transaction but no thirdparty linked. So it may be problematic for accountacy.

### Environment Version

14.0.5

### Environment PHP

PHP 7.4

### Environment Database

MySQL

### Environment URL(s)

htdocs/public/members/new.php

### Expected and actual behavior
New memeber is created
Payement is OK
Subscription is registered
Bank transaction is notified
But there is no thirdparty linked to the operation since there is no thirdparty created and linked to the member

PR #21412, opened 2022-06-29
Membership rework

ymollard:
# Membership rework

I am proposing a rework of the membership ("adherent") self-subscription on several topics:

# NEW "any amount" memberships ("montant libre") are now available per-type
Former parameter `MEMBER_NEWFORM_EDITAMOUNT` still enables custom amounts for all membership types.

# NEW Visitor auto-subscription form now displays available membership types as a table for easiness of reading
image

# NEW Visitors can no longer edit amounts of 0€ at the payment stage in `payment/newpayment.php` (only for membership subscriptions)
If authorized, custom amounts must have been chosen first in `members/new.php`. It clarifies the logic of subscription: first you edit the amount, then you pay if needed.


ymollard:
I am waiting for a first review of this and then plan to pursue:
* Allow to **re**new existing subscriptions rather than telling the visitor to contact the association in that case (probably through a link sent via e-mail to identify the corresponding subscription) 
* Allow to upload profile picture instead of requesting a link to an online picture (Since there is no captcha and this form is public, is there a risk in allowing visitors to upload a picture here?)
* Send membership cards by e-mail after a membership validation/payment (only the invoice is currently sent)
* Do not show Paypal/Stripe buttons if amount is 0€

eldy:
Looks good to me.
However, the addition of the new column must be added into file
htdocs/install/mysql/migration/16.0.0-17.0.0.sql      (create the file if it does not exists)
because no new feature will be added into v16

ymollard:
> the addition of the new column must be added into file htdocs/install/mysql/migration/16.0.0-17.0.0.sql (create the file if it does not exists)

Done @eldy

daraelmin:
Hi @ymollard,
As you're working on this page (and what a great job! Thank you !), i'm wondering if you can check this issue: https://github.com/Dolibarr/dolibarr/issues/21095

### BUG 21095 - No thirdparty created while using the public member creation form

PR #21421, opened 2022-06-30
backport fetchObjectLinked optimisation for pdf

atm-greg:
# FIX backport optimize PDF Generation
Avoid fetch_objectlinked for each page.
the _pagehead method of each PDF model calls "pdf_writeLinkedObjects" which calls "pdf_getLinkedObjects" which make a $object->fetchObjectLinked() on each call.

It induces that we loose several seconds on each printed page for something already done.

the fix consist in flaging the object to know that the full load of its objectLinked have already been done.

Ex. : Situation invoice whith 600 lines generate 25 pages for the PDF
Without this fix => 34 seconds
With the fix => 3 seconds

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -119,6 +119,11 @@ abstract class CommonObject
 	public $linkedObjects;
 
 	/**
+￼	 * @var boolean		Array of boolean with object id as key and value as true if linkedObjects full loaded. Loaded by ->fetchObjectLinked. Important for pdf generation time reduction.
+￼	 */
+	public $linkedObjectsFullLoaded = array();
+
+	/**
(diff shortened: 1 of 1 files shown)


atm-greg:
Sorry @eldy.
I made a mistake when Requesting PR #20534.
The fix was for 15.0 version.
I implemented your fix for static objects too in this one.

eldy:
Hi @atm-greg. 
Because it is a performance optimization, it won't be included in v15. Only bug fixed are allowed. This is to reduce regression.
For example this change has bring some regressions when it was merged into develop. It has been fixed since. I let you see the current code of fetchObjectLinked() in develop to see what has been changed to solve them.
A "if" has been changed an a new parameter has been introduced to disable the cache when we don't load object because having the cache working in all cases was not compatible with a lot of part of existing code that need fresh data when calling fetchObjectLinked().

PR #21431, opened 2022-07-02
NEW : Recurrent events on agenda v1

fboitel:


htdocs/comm/action/card.php
```diff
@@ -1038,44 +1076,26 @@ function setdatefields()
 		print img_picto($langs->trans("Recurrence"), 'recurring', 'class="paddingright2"');
 		print '<input type="hidden" name="recurid" value="'.$object->recurid.'">';
 		$selectedrecurrulefreq = 'no';
-		$selectedrecurrulebymonthday = '';
-		$selectedrecurrulebyday = '';
-		if ($object->recurrule && preg_match('/FREQ=([A-Z]+)/i', $object->recurrule, $reg)) {
```
eldy, line 1043: We must keep this sophisticated syntax to store the frequency. Because this is the standard to store a recurence into an event and is compatible with future feature that need that will need more complex rules (for exemple 1 event every 2 weeks instead of every week).
So object->recurrule need to follow the syntax FREQ=xxx...  (you may find spec into ical format standard)

fboitel:
I just submitted a new version, but its not completely ended, some tests are still missing, like checking that the day of the week is less than or equal to 7, etc...

Issue #21442, opened 2022-07-04
Enhancement of module builder init


eldy:
### Feature Request

On page to create a new module
![image](https://user-images.githubusercontent.com/883887/177136483-db934532-9ee8-4fff-bdfd-4989ccd698fc.png)

Ask also:
Name of company (preffiled with $mysoc->name)
URL of company (prefille with $mysoc->url)
Version (prefilled with 1.0.0)
The number ID of module (prefilled with 500000)
The familly (suggest a combo list with 
```
                                <option value="hr">Gestion des Ressources Humaines (RH)</option>
                                <option value="crm">Gestion de la relation client (GRC)</option>
                                <option value="srm">Gestion de la relation fournisseur (GRF)</option>
                                <option value="financial">Modules financiers (Compta/trésorerie)</option>
                                <option value="products">Gestion des Produits/Services (PM)</option>
                                <option value="projects">Projets/Travail collaboratif</option>
                                <option value="ecm">Gestion électronique de documents (GED)</option>
                                <option value="technic">Outils multi-modules</option>
                                <option value="portal">Sites web et autres applications frontales</option>
                                <option value="interface">Interfaces avec des systèmes externes</option>
                                <option value="base">Système</option>
                                <option value="other" selected="">Autre</option>
```
Picto ID for module  (prefilled with value   "generic")

And use this values to replace hard coded value inside the modXXX.class.php file generated.

lamrani002:
@eldy i guess that issue was already resolved https://github.com/Dolibarr/dolibarr/pull/21455

eldy:
Yes i close the ticket

PR #21455, opened 2022-07-06
NEW #21442 :Enhancement of module builder init

badre-agtaib:
# NEW #21442 Enhancement of module builder init
Capture d’écran du 2022-07-06 08-05-13
Capture d’écran du 2022-07-06 08-03-57


badre-agtaib:
this last commit creates a bug, the state of the code now is not working because we still use modulefamily and idpicto in index.php
@eldy

eldy:
> this last commit creates a bug, the state of the code now is not working because we still use modulefamily and idpicto in index.php @eldy

Sorry i didn't understand. Can you provide an example of what's wrong ?

Issue #21444, opened 2022-07-04
Count elements in some lists


lvessiller-opendsi:
### Bug

In some lists, the new way to count elements can occur an SQL error : 
- this new way uses "preg_replace" method with a regexp "^SELECT[a-zA-Z0-9\._\s\(\),=<>\:\-\']+\sFROM\s"
- and replace all fields in select with a "COUNT(rowid)" SQL request

In an external module you can use the hook  "printFieldListFrom()", to add a subrequest with "LEFT JOIN (SELECT fields FROM table)", and an SQL error occurs when you show this list and you got a blank page.

This new way to build the count SQL request in list, seems to be a good idea to improve memory performances, however it also occurs errors with external modules.

### Environment Version

14.0

### Environment PHP

7.3

### Environment Database

Mysql

eldy:
To help me understand the issue #21444 can you provide the full SQL request you have when using your external module. String before and after.

lvessiller-opendsi:
@eldy

SQL before
----------------
SELECT f.rowid AS id,
       f.ref,
       f.ref_client,
       f.fk_soc,
       f.type,
       f.note_private,
       f.note_public,
       f.increment,
       f.fk_mode_reglement,
       f.fk_cond_reglement,
       f.total_ht,
       f.total_tva,
       f.total_ttc,
       f.localtax1 AS total_localtax1,
       f.localtax2 AS total_localtax2,
       f.fk_user_author,
       f.fk_multicurrency,
       f.multicurrency_code,
       f.multicurrency_tx,
       f.multicurrency_total_ht,
       f.multicurrency_total_tva AS multicurrency_total_vat,
       f.multicurrency_total_ttc,
       f.datef,
       f.date_valid,
       f.date_lim_reglement AS datelimite,
       f.module_source,
       f.pos_source,
       f.paye AS paye,
       f.fk_statut,
       f.close_code,
       f.datec AS date_creation,
       f.tms AS date_update,
       f.date_closing AS date_closing,
       f.retained_warranty,
       f.retained_warranty_date_limit,
       f.situation_final,
       f.situation_cycle_ref,
       f.situation_counter,
       s.rowid AS socid,
       s.nom AS name,
       s.name_alias AS ALIAS,
       s.email,
       s.phone,
       s.fax,
       s.address,
       s.town,
       s.zip,
       s.fk_pays,
       s.client,
       s.fournisseur,
       s.code_client,
       s.code_fournisseur,
[...]

PR #21459, opened 2022-07-06
FIX count elements in invoice list (Issue #21444)

lvessiller-opendsi:
FIX count elements in invoice list (Issue #21444)


lvessiller-opendsi:
@eldy 
Why this PR was revert ?
Did you see any problem ?

Issue #4482, opened 2016-01-23
Feature request: simplify mysql backup wizard


marcosgdf:
Show only a simple wizard to create a backup with the most common options and hide under an "advanced options" menu the rest of the options.

elarifr:
the backup tool should be splitted in two parts
- one admin tool for settings 
- and a user tool that run backup

this is what i try to do in extrabackup module for documents folder, htdocs folder and also mysql db.

settings are managed in admin tool by an admin who has rights 
https://www.dropbox.com/s/8v4lyzyyuluv8yv/2016-03-06%2014_16_22-Greenshot.png?dl=0

then a user with no admin privilege but who have only module rights can run the backup but not download the zip
https://www.dropbox.com/s/fucp6crbbtq2qni/2016-03-06%2014_16_38-Greenshot.png?dl=0

This should also allow to run backups by cron jobs

eldy:
In first step, adding js to hide/show advanced option on page htdocs/admin/tools/dolibarr_export.php   would be enough

eldy:
Has been implemented into v17

PR #21473, opened 2022-07-08
NEW #4482 adding js to hide/show advanced option

badre-agtaib:
# NEW #4482 adding js to hide/show advanced option


eldy:
@badre-agtaib The change works but the look and feel is not very nice.
Button is not aligned and when switching, there is a flash effect where the left section change. i recommend to complete with a PR to move the button and the "advanced option" section under the section "Export method"

PR #21479, opened 2022-07-08
FIX count elements in invoice list (Issue #21444)

lvessiller-opendsi:
FIX count elements in invoice list (Issue #21444)

### DIFF
--- a/htdocs/compta/facture/list.php
+++ b/htdocs/compta/facture/list.php
@@ -480,42 +480,47 @@ $facturestatic = new Facture($db);
 $formcompany = new FormCompany($db);
 $companystatic = new Societe($db);
 
-$sql = 'SELECT';
+$sql_select = "SELECT";
+$sql_select_count = "SELECT COUNT(f.rowid)";
 if ($sall || $search_product_category > 0 || $search_user > 0) {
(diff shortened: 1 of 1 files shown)


lvessiller-opendsi:
@eldy 
Why PR #21459 was revert ?
Did you see any problem ?

eldy:
> @eldy Why PR #21459 was revert ? Did you see any problem ?

Yes I got SQL error when using a defaut Dolibarr setup. So we still have a problem when we used external module, but i prefered to revert because common use without external module was broken (SQL syntax errors with some strict mysql setup, and bad counting with others).

To help me understand the issue #21444 can you provide the full SQL request you have when using your external module. String before and after.

lvessiller-opendsi:
@eldy 
The SQL request remains the same at final.
Could you send me your SQL request to understand where I forgot to concatenate the SQL request  ?

lvessiller-opendsi:
@eldy 
I checked with a default Dolibar 14.0 but I did'nt get any error.
Could you send me your SQL error to understand my mistake ?

lvessiller-opendsi:
Fixed in Devcamp

PR #21481, opened 2022-07-08
Create a 3rd party from the self-subscription membership form

### LINKED ISSUE #21095: No thirdparty created while using the public member creation form

ymollard:
# FIX #21095 Create a 3rd party from the self-subscription membership form
Only when ADHERENT_DEFAULT_CREATE_THIRDPARTY = 1


daraelmin:
Thx a lot !!!

daraelmin:
Hi @ymollard,

I think pour patch should ne pushed un 14.0 not in develop. That may ne thé reason of thé falling intéegation test.

Cheers

ymollard:
OK. I created another PR for v14: https://github.com/Dolibarr/dolibarr/pull/21574

PR #21497, opened 2022-07-12
another set of empty($action) fixes

cfoellmann:
Extends #21469 

some more blockers for card.php without $action

### DIFF
--- a/htdocs/bom/bom_card.php
+++ b/htdocs/bom/bom_card.php
@@ -67,10 +67,6 @@ foreach ($object->fields as $key => $val) {
 	}
 }
 
-if (empty($action) && empty($id) && empty($ref)) {
-	$action = 'view';
-}
-
 // Load object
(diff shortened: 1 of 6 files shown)


eldy:
Note: I reverted #21469 because it introduced some regression going on the create form when we should not (for example on navigation).
What do you think of fixing diferently, for example by adding this just after the $action = GETPOST('action', 'aZ09');
`if (empty($action)) $action = 'view'; `

cfoellmann:
setting to `view` would kill the card.php also.
When there is nothing to view why load a view screen?

cfoellmann:
I realize what the problem is. What makes it hard is that all card.php files are somewhat different. We definitely need a class to replace the functionality of card.phps

```
if (empty($action) && empty($id) && empty($ref)) {
    $action = 'create';
}
```

if it defaults to 'create' it might do the trick

cfoellmann:
I will do another PR and collect all fixes until it is complete

PR #21499, opened 2022-07-12
FIX load product stock in inventory lines

lvessiller-opendsi:
FIX load product stock in inventory lines
- fix real stock when a product is added in an started inventory and when we generate movements and close an inventory

First you begin an inventory with theses lines : 
image

You add a product and the real stock is empty (no value whereas the real sotck is 20) : 
image

And when you "Generate movements and close" this inventory, you got an incrementation in product stock and so you got 30 in real stock instead of having 10 (in this example for PROD4).

### DIFF
--- a/htdocs/product/inventory/inventory.php
+++ b/htdocs/product/inventory/inventory.php
@@ -103,6 +103,8 @@ if ($cancel) {
 	$action = '';
 }
 
+$error = 0;
+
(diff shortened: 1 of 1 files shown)


eldy:
I can't reproduce on v16. Is it a backport of a fix done into v16 or an original implementation in v14 only ?

lvessiller-opendsi:
@eldy 
It's a backport of a fix done into v16.

PR #21500, opened 2022-07-12
fix establishment object UI display

cfoellmann:
display the location label otherwise there is just a number displayed for the user without any information


eldy (commented):
Oops

cfoellmann:
What do you mean?
I did add an extra field from the class and it just displays the house icon + the number

PR #21502, opened 2022-07-13
NEW hook formConfirm on action comm card

lvessiller-opendsi:
NEW hook formConfirm on action comm card
- add "formConfirm" hook for external modules
- fix undefined "formquestion" variable

### DIFF
--- a/htdocs/comm/action/card.php
+++ b/htdocs/comm/action/card.php
@@ -1903,12 +1903,21 @@ if ($id > 0) {
 	} else {
 		print dol_get_fiche_head($head, 'card', $langs->trans("Action"), -1, 'action');
 
+		$formconfirm = '';
+
 		// Clone event
 		if ($action == 'clone') {
-			$formconfirm = $form->formconfirm($_SERVER["PHP_SELF"].'?id='.GETPOST('id'), $langs->trans('ToClone'), $langs->trans('ConfirmCloneEvent', $object->label), 'confirm_clone', $formquestion, 'yes', 1);
-
-			print $formconfirm;
+			$formconfirm = $form->formconfirm($_SERVER["PHP_SELF"].'?id='.GETPOST('id'), $langs->trans('ToClone'), $langs->trans('ConfirmCloneEvent', $object->label), 'confirm_clone', array(), 'yes', 1);
 		}
+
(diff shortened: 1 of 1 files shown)


htdocs/comm/action/card.php
```diff
@@ -1903,13 +1903,22 @@ function init_repeat()
...
 		if ($action == 'clone') {
-			$formconfirm = $form->formconfirm($_SERVER["PHP_SELF"].'?id='.GETPOST('id'), $langs->trans('ToClone'), $langs->trans('ConfirmCloneEvent', $object->label), 'confirm_clone', $formquestion, 'yes', 1);
-
-			print $formconfirm;
+			$formconfirm = $form->formconfirm($_SERVER["PHP_SELF"].'?id='.GETPOST('id'), $langs->trans('ToClone'), $langs->trans('ConfirmCloneEvent', $object->label), 'confirm_clone', array(), 'yes', 1);
 		}
 
+		// Call Hook formConfirm
+		$parameters = array();
+		$reshook = $hookmanager->executeHooks('formConfirm', $parameters, $object, $action); // Note that $action and $object may have been modified by hook
```
eldy, line 1915: Why don't you add the executeHooks  inside  $form->formconfirm() method ?

lvessiller-opendsi: @eldy 
The formConfirm hook is always implemented in cards.
It's dangerous now to move this hook in formconfirm() method.
If we do it, we should remove all formconfirm hooks in cards.
Do you think we can change this way to use formConfirm hooks ?

eldy: I see.
Yes, in alpha phase we can take some risk. And having it implemented in all card looks a bad practice so if  we can implement it differently in a better way, we should.
Risk is limited to external module, and adding a mention into the WARNING section in the "ChangeLog" can reduce seriously bad impact it may have.
Is it a huge work to use the suggested new implementation instead of old one ?

I can validate the PR because it match the way code is currently developped, but if you can submit a version with a better place for trigger, it would be better (despite risk of regression on some external module)...

eldy: Ok, thanks. So let keep as it is.

PR #21503, opened 2022-07-13
NEW : Stock transfer on products deserialization

atm-adrien:
## NEW : Stock transfer on products deserialization

When an user deserialize a product, the serialized stock stay in place. This can be a problem.

To avoid this problem, on the deserialization of a product, for each serialized stock mouvements :
- Clone of the serialized stock mouvement into a global stock mouvement (without batch)
- Creation of a reverse serialized stock mouvements to cancel the serialized stock mouvement 
- Displaying an alert on deserialization

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -1221,6 +1221,41 @@ class Product extends CommonObject
 					}
 				}
 
+				if (!$this->hasbatch() && $this->oldcopy->hasbatch()){
+					// Selection of all product stock mouvements that contains batchs
+					$sql = 'SELECT * FROM '.MAIN_DB_PREFIX.'stock_mouvement sm';
+					$sql .= ' WHERE fk_product = '.$this->id;
+					$sql .= ' AND batch IS NOT NULL';
+
+					$resql = $this->db->query($sql);
+					if ($resql) {
+						while ($obj = $this->db->fetch_object($resql)) {
+							$value = $obj->value;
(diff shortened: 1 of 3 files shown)


BebZ:
Ow ! Good idea ! And if it is a mistake, and doesn't want to deserialize ? is there a reverse way ? 
or display a very big WARNING : DO YOU WANT TO CANCEL ALL SERIALIZED MOVEMENTS ?

htdocs/product/class/product.class.php
```diff
@@ -1221,6 +1221,41 @@ public function update($id, $user, $notrigger = false, $action = 'update', $upda
 					}
 				}
 
+				if (!$this->hasbatch() && $this->oldcopy->hasbatch()) {
+					// Selection of all product stock mouvements that contains batchs
+					$sql = 'SELECT * FROM '.MAIN_DB_PREFIX.'stock_mouvement sm';
```
eldy, line 1226: Why do you analyze stock movement table ?
If we want to clean the stock for each serial number, we should loop on the table of stock for serial (so llx_product_batch) instead ?  The table stock_movement is not the reflect of the stock and depending on how the stock was initialized may not contains all delta transactions from 0 bu may be only from a certain date. To be sure to have the real stock of each lot to clean, the reliable table is llx_product_batch.

atm-gauthier:
@atm-adrien i think you can close this PR, i did @eldy's corrections in this new one : https://github.com/Dolibarr/dolibarr/pull/21691

PR #21505, opened 2022-07-13
FIX: the date mask on products did not work properly

noec764:
The date mask on products did not work properly. I changed the mask to take into account the date in the SQL query used to find the next value.

### DIFF
--- a/htdocs/core/lib/functions2.lib.php
+++ b/htdocs/core/lib/functions2.lib.php
@@ -1256,12 +1256,14 @@ function get_next_value($db, $mask, $table, $field, $where = '', $objsoc = '', $
 	$maskLike = dol_string_nospecial($mask);
 	$maskLike = str_replace("%", "_", $maskLike);
 
+	$now = new DateTime();
+	$maskLike = preg_replace('/\{yyyy\}/i', $now->format('Y'), $maskLike);
+	$maskLike = preg_replace('/\{yy\}/i', $now->format('y'), $maskLike);
+	$maskLike = preg_replace('/\{y\}/i',  substr($now->format('y'),1), $maskLike);
+	$maskLike = preg_replace('/\{mm\}/i', $now->format('m'), $maskLike);
+	$maskLike = preg_replace('/\{dd\}/i', $now->format('d'), $maskLike);
+
 	// Replace protected special codes with matching number of _ as wild card caracter
-	$maskLike = preg_replace('/\{yyyy\}/i', '____', $maskLike);
-	$maskLike = preg_replace('/\{yy\}/i', '__', $maskLike);
-	$maskLike = preg_replace('/\{y\}/i', '_', $maskLike);
-	$maskLike = preg_replace('/\{mm\}/i', '__', $maskLike);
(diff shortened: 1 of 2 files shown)


htdocs/core/lib/functions2.lib.php
```diff
@@ -1256,12 +1259,13 @@ function get_next_value($db, $mask, $table, $field, $where = '', $objsoc = '', $
 	$maskLike = dol_string_nospecial($mask);
 	$maskLike = str_replace("%", "_", $maskLike);
 
+	$maskLike = preg_replace('/\{yyyy\}/i', $now->format('Y'), $maskLike);
```
eldy, line 1262: We must replace with "_" because the number must not depends on the date, so if we have a mask
{000}{yyyy} 
and record in database
0042022
0032023
we must get a maskLike that is "{000}____" with 4 _ where _ means any char in sql (we don't want to take into account the year but number of chars must match)

eldy: When you says "don't work". Can you describe the bug and the way to data to reproduce it.
Also it seems after your change all phpunit test done on the numbering system fails when they were ok before.

noec764: I have a product in my database with the name PRLVT_24h but my mask for the new product to create is P{yy}{000000}.
So when get_next_value is called, this product is found by sql search with $maskLike = P________ 
with the value VT_24h in $counter. Then $counter becomes 0 (because it is not numerical) and the next_value for the product name becomes P22000001 even though it already exists in my database.

noec764: The failed unit Test is ```test/phpunit/NumberingModulesTest.php :: 170 ``` 
		
``` 
$localobject2->date=dol_mktime(12, 0, 0, 1, 1, 1916);	// we use following year for second invoice (there is no reset into mask)
$numbering=new mod_facture_mercure();
$result=$numbering->getNextValue($mysoc, $localobject2, 'last');
$this->assertEquals('1915-0001', $result, "Test to get last value with param 'last'");
```
I don't understand why I should find 1915-0001 with the last parameter because there is no other invoice this year, I expect to get 1916-0001.

eldy: > I have a product in my database with the name PRLVT_24h but my mask for the new product to create is P{yy}{000000}. So when get_next_value is called, this product is found by sql search with $maskLike = P________ with the value VT_24h in $counter. Then $counter becomes 0 (because it is not numerical) and the next_value for the product name becomes P22000001 even though it already exists in my database.

So this is the expected behaviour. Every old value that does not match the mask must not be taken into account to get the next value of counter.
So your next value for counter with a mask of 
P{yy}{000000}
if you have only one article that is PRLVT_24h
must be
P22000001

noec764: I am agree but it is currently not the case. I have  PRLVT_24h and another P22000004 but the next value is  P22000001. 
But I think I understand why this happened I will do another PR.

PR #21506, opened 2022-07-13
NEW events list with color

lvessiller-opendsi:
NEW event list with color
- setup colors for past, current and future events (in agenda setup)
image

- show colors on events list
image

_Remark : you can keep empty colors if you don't want to color this events list_


eldy:
We already use 3 colors on events 
- one depending on type of event (visible on the font awesome picto)
- one depending on the owner of event (visible in view calendar on left bar of the event).
- one for the status (visible on the status picto)
image

Introducing a 4th level of color will make setup of colors very difficult to understand. Also a CSS was already introduced called "cal_past" to use a different css on a past event. But the template decided to show the same way (in old version, it was greyd instead of black like Google calendar do for past event).

Before introducing a 4th level of color i woul prefer to wait that the 3 other level of colors are completely implemented (if we use it into a calendar view, we must also use it into list view and the color for owner is not yet implemented into list).
Also, to visualize past event to do, we already have the couple status + warning late if not done.

Using a color on line make not possible to keep the same signalitic everywhere. Why not instead using a rule to change color of the date only. For example, when date start is in past, we set it in grey (opacitymedium), when in future, we keep it black, same for date end. If the 2 dates are grey, if means it is past, only date start, we are in range, none, event is in future.

lvessiller-opendsi:
@eldy 
I added colors for owners and replace background color on lines with background color on start date for past, current and future events : 
image

lvessiller-opendsi:
@eldy 
I think all has been done for this event list.
Could you merge it ?

lvessiller-opendsi:
@eldy 

lvessiller-opendsi:
@eldy 
Could you merge it ?

lvessiller-opendsi:
@eldy 
Could you merge it or tell me if there is something wrong ?

lvessiller-opendsi:
@eldy 
What about this PR ?

PR #21514, opened 2022-07-15
NEW Add search on member nature morphy

dbeniamine:
# New Add a selector to search on morphy on member list

This MR adds a selector to show only moral / physical members on member list

image


eldy:
Can you confirm that default selected value is "nothing" into this combo list, so by default there is no filter for this key ?

dbeniamine:
> Can you confirm that default selected value is "nothing" into this combo list, so by default there is no filter for this key ?

Yes there are no filters by default

PR #21522, opened 2022-07-18
NEW: hook to modify supplier product html select

marc-dll:


htdocs/core/class/html.form.class.php
```diff
@@ -3277,6 +3277,14 @@ public function select_produits_fournisseurs_list($socid, $selected = '', $htmln
 		if (!empty($filtre)) {
 			$sql .= " ".$filtre;
 		}
+
+		if (! is_object($hookmanager)) {
```
eldy, line 3281: We are in a method used to forge html context.
So a such method should be called when a hook already exists and a context too. We should never have the context of use set by a low level method but must be set by the caller.
In which call and situation do you have $hookmanager not defined ?

marc-dll: I did not encounter such a case, I just put that to be safe, but I understand it is superfluous, I will remove it.

PR #21531, opened 2022-07-19
Fix bad condition to enable search_morphy

dbeniamine:
# FIX Bad condition to enable search_morphy

There was an error in #21514 : the search_morphy sql was always inserted.

This commit changes the condition to that enable the sql to make sure it is only used when the user requests for a filter on the member's nature


eldy:
Fix seems good but the PR is in conflict so can't be merged. Can you resubmit it ?

dbeniamine:
> Fix seems good but the PR is in conflict so can't be merged. Can you resubmit it ?

Done

eldy:
It was never possible to reproduce, so i close pr ( i suppose it is fixed)

PR #21533, opened 2022-07-20
FIX: the date mask on products did not work properly

noec764:
A PR related to #21505

### DIFF
--- a/htdocs/core/lib/functions2.lib.php
+++ b/htdocs/core/lib/functions2.lib.php
@@ -1278,7 +1278,7 @@ function get_next_value($db, $mask, $table, $field, $where = '', $objsoc = '', $
 
 	// Get counter in database
 	$counter = 0;
-	$sql = "SELECT MAX(".$sqlstring.") as val";
+	$sql = "SELECT MAX(CAST(".$sqlstring." AS UNSIGNED)) as val";
 	$sql .= " FROM ".MAIN_DB_PREFIX.$table;
 	$sql .= " WHERE ".$field." LIKE '".$db->escape($maskLike)."'";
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions2.lib.php
```diff
@@ -1278,7 +1278,7 @@ function get_next_value($db, $mask, $table, $field, $where = '', $objsoc = '', $
 
 	// Get counter in database
 	$counter = 0;
-	$sql = "SELECT MAX(".$sqlstring.") as val";
+	$sql = "SELECT MAX(CAST(".$sqlstring." AS UNSIGNED)) as val";
```
eldy, line 1281: I see the idea.  But does this SQL will be compatible with postgresql ?

htdocs/core/lib/functions2.lib.php
```diff
@@ -1278,10 +1278,12 @@ function get_next_value($db, $mask, $table, $field, $where = '', $objsoc = '', $
 
 	// Get counter in database
 	$counter = 0;
-	$sql = "SELECT MAX(".$sqlstring.") as val";
+
+	$sql = "SELECT ".$sqlstring." as val";
```
eldy, line 1282: If we have 1 millions of record, this may return 1 million of lines that then will be parsed line by line.

noec764: I already try to do this in the commit 053fb76 but Travis CI failed.

noec764: @eldy What can I do for this PR ?

eldy:
Closed. Better solution provided in #22819

PR #21535, opened 2022-07-20
NEW : Function getGlobalConst - Avoid PHP8 Warnings

ghost:
This function will be used to avoid PHP8 warnings by checking if the configuration exists and falling back to a default value if not.
I don't know where to put this portion of code, I find that the function.lib is rather suitable for the moment. We have started to count our modules and have already been very often confronted with this type of warning. I therefore think that this function can be very useful for the community in view of the PHP changes that will occur in the upcoming months.


eldy:
We already have the method getDolGlobalString() that is dedicated for this need.

PR #21539, opened 2022-07-20
Add hook to receiptprinter class

catrielr:
Add hook's to add aditionals tag for external modules.

# Instructions
*Please:*

[*Long description*]

[*Long description*]

[*Long description*]

### DIFF
--- a/htdocs/core/class/dolreceiptprinter.class.php
+++ b/htdocs/core/class/dolreceiptprinter.class.php
@@ -579,7 +579,7 @@ class dolReceiptPrinter extends Printer
 		global $conf, $mysoc, $langs, $user;
 		global $hookmanager;
 		$error = 0;
-		
+
 		$hookmanager->initHooks(array('receiptprinter'));
 		$ret = $this->loadTemplate($templateid);
 
--- a/htdocs/core/class/dolreceiptprinter.class.php
+++ b/htdocs/core/class/dolreceiptprinter.class.php
@@ -578,6 +578,9 @@ class dolReceiptPrinter extends Printer
 	{
 		global $conf, $mysoc, $langs, $user;
 		$error = 0;
+		require_once DOL_DOCUMENT_ROOT.'/core/class/hookmanager.class.php';
+		$hookmanager = new HookManager($this->db);
+		$hookmanager->initHooks(array('receiptprinter'));
 		$ret = $this->loadTemplate($templateid);
 
 		// tags a remplacer par leur valeur avant de parser (dol_value_xxx)
@@ -620,6 +623,15 @@ class dolReceiptPrinter extends Printer
 		$this->template = str_replace('{dol_value_vendor_firstname}', $user->firstname, $this->template);
 		$this->template = str_replace('{dol_value_vendor_lastname}', $user->lastname, $this->template);
 		$this->template = str_replace('{dol_value_vendor_mail}', $user->email, $this->template);
+
+		$parameters = array('object'=>$object);
+		$action = '';
+		$reshook = $hookmanager->executeHooks('sendToPrinterBefore', $parameters, $this, $action); // Note that $action and $object may have been modified by some hooks
+		if ($reshook < 0) {
+			$this->error = "Error in hook dolReceiptPrinter sendToPrinterBefore ".$reshook;
+			dol_syslog("dolReceiptPrinter::sendToPrinter: error=".$this->error, LOG_ERR);
+			return $reshook;
+		}
 
 		// parse template
 		$this->template = str_replace("{", "<", $this->template);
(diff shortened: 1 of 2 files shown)


catrielr:
update code

htdocs/core/class/dolreceiptprinter.class.php
```diff
@@ -577,7 +577,10 @@ public function sendTestToPrinter($printerid)
 	public function sendToPrinter($object, $templateid, $printerid)
 	{
 		global $conf, $mysoc, $langs, $user;
+		global $hookmanager;
 		$error = 0;
+
+		$hookmanager->initHooks(array('receiptprinter'));
```
eldy, line 583: You should be able to remove this line too. 
Can you try ?
You should be able to use the context "main" that is initialized at begin of each page with
`$hookmanager->initHooks(array('main'));`

catrielr: Eldy, if I don't do the $hookmanager->initHooks(array('receiptprinter')); does not call the hook.

eldy: To have the hook called, did you declared into your module the context "main" (because you want the method sendToPrinterBefore to be called whatever is the context).

`'hooks' => array('data' => array("main")...`

The reenabled module to take change into account.

catrielr:
Eldy, sorry I had not understood you yesterday.
The problem with leaving it in the "main" context is that it runs on all pages, and should only run when a print is sent from the receiptprinter module. So to make a new context, and from the external module to be able to execute specific code for that context. That's why I do the $hookmanager->initHooks(array('receiptprinter'));

htdocs/core/class/dolreceiptprinter.class.php
catrielr: Eldy, sorry I had not understood you yesterday.
The problem with leaving it in the "main" context is that it runs on all pages, and should only run when a print is sent from the receiptprinter module. So to make a new context, and from the external module to be able to execute specific code for that context. That's why I do the $hookmanager->initHooks(array('receiptprinter'));

htdocs/core/class/dolreceiptprinter.class.php
```diff
@@ -578,6 +578,9 @@ public function sendToPrinter($object, $templateid, $printerid)
 	{
 		global $conf, $mysoc, $langs, $user;
 		$error = 0;
+		require_once DOL_DOCUMENT_ROOT.'/core/class/hookmanager.class.php';
+		$hookmanager = new HookManager($this->db);
```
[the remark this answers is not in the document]
catrielr: Eldy, is the same case of: [https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/core/modules/facture/doc/pdf_sponge.modules.php](url) 

`		// Add pdfgeneration hook
				if (!is_object($hookmanager)) {
					include_once DOL_DOCUMENT_ROOT.'/core/class/hookmanager.class.php';
					$hookmanager = new HookManager($this->db);
				}
				$hookmanager->initHooks(array('pdfgeneration'));
				$parameters = array('file'=>$file, 'object'=>$object, 'outputlangs'=>$outputlangs);
				global $action;
				$reshook = $hookmanager->executeHooks('beforePDFCreation', $parameters, $object, $action); // Note that $action and $object may have been modified by some hooks
`

htdocs/core/class/dolreceiptprinter.class.php
eldy: But here you create a context to restrict the call to method.
Call of context is to define in which situation/page we may call the method and the method decide at which place in code we lake the hook.

With the context main, the hook will be called each time the method sendToPrinter() will be called.
Should be the same than when you add a new context, shouldn't it ? Except that with hook main, the limitation is done by the coupe context/method name, in your case, having a limit done by context and by method name has the same result making the couple "context / method" useless.

catrielr: It's the other way around, the sendprinter() method calls the hook only for external modules that have:
```

$this->module_parts = array(
             'hooks'=> array(
                              'data' =>array( 'receiptprinter'),
```

It would be the same as what happens with this in pdf_sponge.modules.php
```
$hookmanager->initHooks(array('pdfgeneration'));
$parameters = array('file'=>$file, 'object'=>$object, 'outputlangs'=>$outputlangs);
global$action;
$reshook = $hookmanager->executeHooks('beforePDFCreation', $parameters, $object, $action); // Note that $action and $object may have been modified by some hooks

```
Since the sendprint() method is the one that shapes the ticket

I hope it is understood. I have it running in development.
And with this hook I can even add ESC commands to the printer that are not the ones defaulted by the sendprinter() method.

eldy: PDF generation is a offline process, that's why a context exists especially for the xxxPDFCreation.

But for sendToPrinter, this function is already called into existing contet.
For example  "takeposproductsearch" and "takeposinvoice". So you should just init the hook to support this 2 context.

$this->module_parts = array( 'hooks'=> array( 'data' =>array( 'takeposproductsearch', 'takeposinvoice'),

and you should get the same result.

PR #21551, opened 2022-07-21
Fix #21449

### LINKED ISSUE #21449: Price request status in listing says Open, what ever the real status is
### Bug

Price request status in listing says Open, what ever the real status is

### Environment Version

15.0.2

### Environment OS

Debian 10

### Environment Web server

Apache2 

### Environment PHP

fpm-fcgi 7.3.33-1+020211119.91+debian101.gbp618351

### Environment Database

MySQL or MariaDB 5.5.5-10.3.34-MariaDB-0+deb10u1

### Environment URL(s)

/supplier_proposal/list.php?leftmenu=supplier_proposals

When the status of a price request is changed to Accepted, Refused or Closed that should be reflected in the listing, but it is not. In the listing everything is marked as Ouvert (Open) as can be seen on the attached picture. But if the filtering fuction above the colmun is used, it will filter correctly, but the marking is still wrong; everthing is marked as Open.

Create a price request, play around with with its status and see what appears in the listing.

### Attached files

status

emheyarssi:
Fix #21449

### DIFF
--- a/htdocs/supplier_proposal/list.php
+++ b/htdocs/supplier_proposal/list.php
@@ -302,7 +302,7 @@ $sql .= " p.rowid as project_id, p.ref as project_ref,";
 if (empty($user->rights->societe->client->voir) && !$socid) {
 	$sql .= " sc.fk_soc, sc.fk_user,";
 }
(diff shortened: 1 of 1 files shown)


emheyarssi:
I think I should make a pull request to branch 15.0

emheyarssi:
See new pull request #21557

PR #21555, opened 2022-07-22
fix: display message only if change (with emballage on)

FHenry:
fix: display message only if change (with emballage on)


eldy:
@FHenry Don't forget that for FIX, the branch 16.0 is open. 
I have cherry-picked this commit into 16.0.

FHenry:
My bad, I want to put it 15

Issue #14753, opened 2020-09-16
Recruitment module - non english code


cfoellmann:
recruitment is still in the early stages but there are again file names with french terminology.
what is `candidature`? it that the person applying that is called `candidate`?

ksar-ksar:
Hi,

Sorry but this is an english word : https://www.collinsdictionary.com/dictionary/english/candidature

cfoellmann:
But that does not make sense here.
on the pages "called" `candidature` you track your candidates.

eldy:
With the object candidature, we track candidature, not candidate (it is nearly the same but candidature is a "candidate request for 1 given job position", if 1 candidate apply for 2 job positions, we will create 2 record).
But term "candidacy" may be a better translation ?

cfoellmann:
I am not a professional translator but have done a lot of translations.
For me `candidates` are people looking to get voted or accepted into a political or honorary position.

For a job the people are `applicants` and their request to get the job is an `application`

atm-maxime:
I agree that a better translation would be application (candidacy) and applicant (candidate)

atm-florianm:
I completely agree with @cfoellman regarding the terminology, but since this issue seems to be about file names and/or variable names (not `.lang` files), it might be too late to change them (some external modules may already be relying on the misleading names, and fixing them might break compatibilty).

[edit]
Sorry, I didn't get that this was about the new v13.0 feature… I'll have a look then; please forget my previous comment.

cfoellmann:
I think that this an alpha feature so it should be good to change this before going a step further.
as much as i like the speed new features are coming, older features as mostly not really usable because of some crucial part missing.

There should be a big break with new features for v14 and only make this an improvement release. Adding needed features to existing parts and polishing up the code.

atm-florianm:
> I think that this an alpha feature so it should be good to change this before going a step further.

Yes, I realized that a minute ago. It should be possible to fix this terminology issue, except for strings that get inserted into the database (such as `recruitmentcandidature`). Actually, even for those, a migration script could be provided so that users who already use the feature can update their data.

atm-florianm:
I’m having second thoughts about this mass renaming… it turns out to be harder than I figured.

cfoellmann:
That is what I thought but should be done,
Moving from the French expressions to ENG is hard too. But it makes the project more open to international audiences

cfoellmann:
[...]

PR #21563, opened 2022-07-22
Fix recruitment

### LINKED ISSUE #18279: Recruitment mask (customized number) did not work
# Bug
The customized numbers in the Recruitment module did not work

## Environment
- **Version**: Dolibarr 14.0.0

cfoellmann:
# FIX
- numbering for jobs
- advanced numbering for applications

CLOSE #18279

Question is:

can we fix the naming?
"Candidature" makes no sense!! It is just wrong. It is an application.

All constants and parameters are far too long all over the place. It is a pain to debug or extend this.

I am open to do the work and change this bad naming as discussed here #14753


eldy:
Yes Candidature is not the best term. "Application" is the good one.
However, this term should never appears to end user. Only for dev.

You can make PR to change this but with very small changes and atomic. For example, changing only the name of table and having code still ok. Another one for changing only the name of class without changing table. In the order of your choice but only one PR for on reason (table, or class or constants, or ..., not a global PR for all renaming).

What about this PR ? Seems to be in draft status. Can i merge it ?

cfoellmann:
@eldy Thanks for being open for the naming change.

This PR is still in the works. I want to fix "everything" that is currently broken in the module.
There are still a lot of bugs

htdocs/recruitment/class/recruitmentjobposition.class.php
```diff
@@ -105,13 +105,13 @@ class RecruitmentJobPosition extends CommonObject
...
 		'fk_establishment' => array('type'=>'integer:Establishment:hrm/class/establishment.class.php', 'label'=>'Establishment', 'enabled'=>'$conf->hrm->enabled', 'position'=>56, 'notnull'=>0, 'visible'=>-1, 'foreignkey'=>'establishment.rowid',),
 		'fk_soc' => array('type'=>'integer:Societe:societe/class/societe.class.php:1:status=1 AND entity IN (__SHARED_ENTITIES__)', 'label'=>'WorkPlace', 'enabled'=>'$conf->societe->enabled', 'position'=>57, 'notnull'=>-1, 'visible'=>-1, 'css'=>'maxwidth500', 'index'=>1, 'help'=>"IfJobIsLocatedAtAPartner", 'picto'=>'company'),
 		'date_planned' => array('type'=>'date', 'label'=>'DateExpected', 'enabled'=>'1', 'position'=>60, 'notnull'=>0, 'visible'=>1,),
-		'remuneration_suggested' => array('type'=>'varchar(255)', 'label'=>'Remuneration', 'enabled'=>'1', 'position'=>62, 'notnull'=>0, 'visible'=>1,),
+		'remuneration_suggested' => array('type'=>'integer', 'label'=>'Remuneration', 'enabled'=>'1', 'position'=>62, 'notnull'=>0, 'visible'=>1,),
```
eldy, line 114: We must keep a free text here.
If may be a text like "Depends on ..." or "To negociate..."

cfoellmann: all other renumerations_ are int with char it is not comparable

cfoellmann: I will remove for now

PR #21565, opened 2022-07-24
Fix #21536

### LINKED ISSUE #21536: API Invoice Delete error
### Bug

In file /compta/facture/class/api_invoices.class.php in function delete on line 669 a test is made if $result is lower than 0 in order to check if invoice has been deleted
> 
		$result = $this->invoice->delete(DolibarrApiAccess::$user);
		if ($result < 0) {
			throw new RestException(500, 'Error when deleting invoice');
		}

But, in file /compta/facture/class/facture.class.php in function delete in line 2582 value 0 is returned if invoice is not erasable
> 
		// Test to avoid invoice deletion (allowed if draft)
		$result = $this->is_erasable();

		if ($result <= 0) {
			return 0;
		}

The check in API should be lower or equal to 0

### Environment Version

_No response_

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

emheyarssi:
API Invoice Delete doesn't raise exception in some case

### DIFF
--- a/htdocs/compta/facture/class/api_invoices.class.php
+++ b/htdocs/compta/facture/class/api_invoices.class.php
@@ -666,7 +666,7 @@ class Invoices extends DolibarrApi
 		}
 
 		$result = $this->invoice->delete(DolibarrApiAccess::$user);
-		if ($result < 0) {
+		if ($result <= 0) {
 			throw new RestException(500, 'Error when deleting invoice');
 		}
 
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/class/api_invoices.class.php
```diff
@@ -666,7 +666,7 @@ public function delete($id)
 		}
 
 		$result = $this->invoice->delete(DolibarrApiAccess::$user);
-		if ($result < 0) {
+		if ($result <= 0) {
```
eldy, line 669: The case 0 is not a technical error so it must not return the Exception with code 500
Add a if to return a
throw new RestException(403, 'Invoice not erasable');

PR #21570, opened 2022-07-25
FIX : Yogosha report 11874 js injection in stockmovement.php

Hystepik:
Fix a security issue in /product/stock/stockmovement.php

### DIFF
--- a/htdocs/product/stock/massstockmove.php
+++ b/htdocs/product/stock/massstockmove.php
@@ -336,11 +336,11 @@ if ($action == 'importCSV' && !empty($user->rights->stock->mouvement->creer)) {
 					continue;
 				}
 				//var_dump($data);
-				$tmp_id_sw = $data[$i][0]['val'];
-				$tmp_id_tw = $data[$i][1]['val'];
-				$tmp_id_product = $data[$i][2]['val'];
-				$tmp_qty = $data[$i][3]['val'];
-				$tmp_batch = $data[$i][4]['val'];
+				$tmp_id_sw = dol_escape_htmltag($data[$i][0]['val']);
+				$tmp_id_tw = dol_escape_htmltag($data[$i][1]['val']);
+				$tmp_id_product = dol_escape_htmltag($data[$i][2]['val']);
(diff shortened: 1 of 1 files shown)


htdocs/product/stock/massstockmove.php
```diff
@@ -336,11 +336,11 @@
 					continue;
 				}
 				//var_dump($data);
-				$tmp_id_sw = $data[$i][0]['val'];
-				$tmp_id_tw = $data[$i][1]['val'];
-				$tmp_id_product = $data[$i][2]['val'];
-				$tmp_qty = $data[$i][3]['val'];
-				$tmp_batch = $data[$i][4]['val'];
+				$tmp_id_sw = dol_escape_htmltag($data[$i][0]['val']);
```
eldy, line 339: No. In memory, we must have date as a brute data.
The escapement must be done when we need it. So if we need to output the data for sql, we will escape using db->escape, if we need to use it for output on screen, we will use dol_escape_htmltag but this will be done during the print of the html content.
Imagine that the data $tmp_batch is used later for a test or to be written into a file, you will write a value with html output escapement when you don't want.

Data in memory must be brut and utf8 encoded.
Escapement must be done when we need it (the later as possible).

htdocs/product/stock/massstockmove.php
```diff
@@ -336,11 +336,11 @@
...
+				$tmp_id_tw = dol_escape_htmltag($data[$i][1]['val']);
+				$tmp_id_product = dol_escape_htmltag($data[$i][2]['val']);
```
eldy, line 341: dol_escape_htmltag is used to sanitize string. When we are sure data is an int, we must use (int) or (float)
So tmp_id_product =  (int) $data[$i][2]['val']

htdocs/product/stock/massstockmove.php
```diff
@@ -336,11 +336,11 @@
...
+				$tmp_qty = dol_escape_htmltag($data[$i][3]['val']);
```
eldy, line 342: Here we cas use
$tmp_qty = price2num((float) $data[$i][3]['val'], 'MS)

PR #21574, opened 2022-07-25
FIX: Create a 3rd party from the self-subscription membership form

### LINKED ISSUE #21095: No thirdparty created while using the public member creation form

ymollard:
# FIX https://github.com/Dolibarr/dolibarr/issues/21095 Create a 3rd party from the self-subscription membership form

Only when ADHERENT_DEFAULT_CREATE_THIRDPARTY = 1

### DIFF
--- a/htdocs/societe/class/societe.class.php
+++ b/htdocs/societe/class/societe.class.php
@@ -3907,12 +3907,16 @@ class Societe extends CommonObject
 		$fullname = $member->getFullName($langs);
 
 		if ($member->morphy == 'mor') {
-			$socname = $member->company? $member->company : $member->societe;
+			if (empty($socname)) {
+				$socname = $member->company? $member->company : $member->societe;
+			}
 			if (!empty($fullname) && empty($socalias)) {
 				$socalias = $fullname;
 			}
 		} elseif (empty($socname) && $member->morphy == 'phy') {
-			$socname = $fullname;
+			if (empty($socname)) {
+				$socname = $fullname;
+			}
 			if (!empty($member->company) && empty($socalias)) {
 				$socalias = $member->company;
 			}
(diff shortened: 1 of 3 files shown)


daraelmin:
Thx

htdocs/societe/class/societe.class.php
```diff
@@ -3904,12 +3904,21 @@ public function create_from_member(Adherent $member, $socname = '', $socalias =
 		global $conf, $user, $langs;
 
 		dol_syslog(get_class($this)."::create_from_member", LOG_DEBUG);
+		$fullname = $member->getFullName($langs);
 
-		$name = $socname ? $socname : $member->societe;
-		if (empty($name)) {
-			$name = $member->getFullName($langs);
+		if ($member->morphy == 'mor') {
+			$socname = $member->company? $member->company : $member->societe;
```
eldy, line 3910: If $socname or $socalias was forced by caller, here you erased the value.
I suggest to implement the logic to define socname and socalias in this method only if input var are not forced by caller, for compatibility and to keep the use of existing parameters.
So you should add a test` if (empty($socname))`

htdocs/societe/class/societe.class.php
```diff
@@ -3904,12 +3904,21 @@ public function create_from_member(Adherent $member, $socname = '', $socalias =
...
+			if (!empty($fullname) && empty($socalias)) {
+				$socalias = $fullname;
+			}
+		} elseif (empty($socname) && $member->morphy == 'phy') {
+			$socname = $fullname;
```
eldy, line 3915: Same here: erase $socname if not already forced.

PR #21577, opened 2022-07-26
FIX: Added international date format to es-HN

axellpadilla:
# FIX Dates in ISO format: yyyy-MM-dd for es-HN
Dates in format: yyyy-MM-dd, previous format can cause some serious conflicts on a country that uses dd/MM/yyyy instead of MM/dd/yyyy so you must set it as dd/MM/yyyy
Also not that translation for no en_US languages mus tbe done into transifex https://www.transifex.com/dolibarr-association/dolibarr/translate/#es_HN/ and not as a PR in github


eldy:
According to this page, date format in Honduras is  dd/MM/yyy
https://en.wikipedia.org/wiki/Date_format_by_country
So why changing. The date format depend on language not in country. If you enter it in your language, it will be correctly formated when reading it into another language whatever is the format.

PR #21584, opened 2022-07-26
NEW: Adding template 01

Quatadah:
#NEW

Adding new template.


eldy:
The dir with image files is empty into the zip
Did you name your images into template with  image/xxxx    with xxxx that is exatly the same name than the website key ?

Quatadah:
This template has no images on purpose. I can add them though if you wish.

eldy:
It seems the zip fils was still generated with an old version of Dolibarr.
The end of the sql insert into website_pages.sql  into the zip is
`"...>__N__', '', '');"`
and you should have
`"...>__N__', '', 0);"`
if you generate the zip with the last version of Dolibarr.

Can you regenerated it ?

PR #21585, opened 2022-07-26
NEW: Adding template 03

Quatadah:
#NEW 
template style 03


eldy:
Note that the system is using a menu based on external javascript lib. 
The template must not depends on an external lib that may disappear soon. You can solve this by embedding the js into the template (you can put a js file into media directory js/mywebsite).
Also, it seems the template is adding a css on class * breaking the css style of Dolibarr. We should remove this.

Quatadah:
which external lib are you talking about ? because dolibarr adds some of them by default. And I just checked, the css has no styling on *. Can you tell me which one exactly ?

PR #21595, opened 2022-07-27
Fix : missing migrations in 13.0.0-14.0.0

Hystepik:
missing migrations in 13.0.0-14.0.0 to add missing columns in a certain case

### DIFF
--- a/htdocs/install/mysql/migration/13.0.0-14.0.0.sql
+++ b/htdocs/install/mysql/migration/13.0.0-14.0.0.sql
@@ -88,6 +88,11 @@ ALTER TABLE llx_product_fournisseur_price ADD COLUMN packaging real DEFAULT NULL
 -- VMYSQL4.3 ALTER TABLE llx_product_fournisseur_price MODIFY COLUMN packaging real DEFAULT NULL;
 -- VPGSQL8.2 ALTER TABLE llx_product_fournisseur_price MODIFY COLUMN packaging real DEFAULT NULL USING packaging::real;
 
+ALTER TABLE llx_user ADD COLUMN datelastpassvalidation datetime;
+ALTER TABLE llx_user ADD COLUMN datestartvalidity datetime;
+ALTER TABLE llx_user ADD COLUMN idpers1 varchar(128);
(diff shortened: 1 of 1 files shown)


htdocs/install/mysql/migration/13.0.0-14.0.0.sql
```diff
@@ -88,6 +88,11 @@ ALTER TABLE llx_product_fournisseur_price ADD COLUMN packaging real DEFAULT NULL
 -- VMYSQL4.3 ALTER TABLE llx_product_fournisseur_price MODIFY COLUMN packaging real DEFAULT NULL;
 -- VPGSQL8.2 ALTER TABLE llx_product_fournisseur_price MODIFY COLUMN packaging real DEFAULT NULL USING packaging::real;
 
+ALTER TABLE llx_user ADD COLUMN datelastpassvalidation datetime;
+ALTER TABLE llx_user ADD COLUMN datestartvalidity datetime;
+ALTER TABLE llx_user ADD COLUMN idpers1 varchar(128);
```
eldy, line 93: It seems field idpers1, 2 and 3 are errors. I can't find any code in Dolibarr that uses them.
Did you got a sql error somewhere due to this fields ?
If not we should drop them into v17

Hystepik: This fields was in 12.0.0-13.0.0.sql and I hadn't them in my database in 13 so I had them in the migration 13.0.0-14.0.0

eldy: Sure but do we really need them. It seems no. So i think the error was to have them into migration file 12.0.0-13.0.0.sql because they are used nowhere fields.

Hystepik: Indeed I did not find any in v14 nor development

eldy: So I choosed a different way to fix this:
I included the drop of the non expected columns into the file migration for v17.

PR #21597, opened 2022-07-27
FIX : VAT rate recalculation

atm-adrien:
## FIX : VAT rate recalculation

Whne we clone an invoice for another thirparty that don't come from the same country than the first thirdprty, the VAT rate can change.
So, I think, we must recalculate the vat rate on the clone

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -1275,6 +1275,13 @@ class Facture extends CommonInvoice
 			$object->lines[$i]->ref_ext = ''; // Do not clone ref_ext
 		}
 
+		if($objFrom->socid != $object->socid){
+			global $mysoc;
+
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/class/facture.class.php
```diff
@@ -1275,6 +1275,13 @@ public function createFromClone(User $user, $fromid = 0)
 			$object->lines[$i]->ref_ext = ''; // Do not clone ref_ext
 		}
 
+		if ($objFrom->socid != $object->socid) {
```
eldy, line 1278: Original idea of clone was "We make a similar copy".
This means we probably need a feature instead. When we clone, we can have into the confirmation request a checkox
"Modify the VAT rate to use the default rate depending on the new thirdparty".

I suggest to provide a PR with such an enhancement so we can keep the rule "Clone is a mirror except for option chosen" consistent everywhere

PR #21600, opened 2022-07-27
FIX: EventOrganization Missing databases entries and labels

ymollard:
# FIX|Missing databases entries for EventOrganization
This was preventing the module to work with the initdemo data.

# FIX|Added en_US labels for EventOrganization

# FIX|Propagate errors to forms
$thing->errors was sometimes not read


htdocs/core/class/commonobject.class.php
```diff
@@ -8675,7 +8675,7 @@ public function isDuration($info)
 	public function isInt($info)
 	{
 		if (is_array($info)) {
-			if (isset($info['type']) && ($info['type'] == 'int' || preg_match('/^integer/i', $info['type']))) {
```
ymollard, line 8678: Not sure of the fixes in `commonobject.class.php`.

Reason of theses fixes: Inserting value 0 to a MySQL column with type "smallint" was being inserted as `null`, causing the insertion to fail.

PR #21601, opened 2022-07-28
Module Tickets - modification select

FranckInfans:
The purpose of this proposal is to create a constant 'TICKET_FORCE_DISPLAY_EMPTY_OPTION' which allows to add an empty line to the "Type of request", "Ticket categorization" and "Severity" selectors.
This ensures that the user will make a real choice and not leave the first option as default

Le but de cette proposition est de créer une constante 'TICKET_FORCE_DISPLAY_EMPTY_OPTION' qui permet de d'ajouter une ligne vide aux selecteurs "Type de demande", "Catégorisation du ticket" et "Sévérité".
Cela permet de s'assurer que l'utilisateur fera un vrai choix et ne laissera pas la première option par défaut


eldy:
You're right. The choice must be done explicitely and there is no reason to have choice selected on first elem (this is not compliant with dolibarr look and feel rules).
However, i don't think we need a hidden option for this.
You can set the empty choice as the behaviour by default for everyone.
Then if end user wants a default value, he can use the setup - default values feature for this.

Can you resubmit by removing option and replace it with hard coded "1" directly ?

PR #21609, opened 2022-07-28
NEW add tax message on takepos menu

ptibogxiv:


ptibogxiv:
for legal reasons, some country need more informations to display

eldy:
Still travis errors

eldy:
@ptibogxiv can you fix the syntax error reported by travis ?

eldy:

PR #21615, opened 2022-07-28
NEW - FTP lib

BB2A-Anthony:
# NEW|New [*NEW - FTP lib and lang file for FTP*]

The goal is to be able to use the functions in something else like sending backups

And enrich it afterwards


htdocs/langs/en_US/ftp.lang
```diff
@@ -0,0 +1,15 @@
+# FTP
```
eldy, line 1: For several reason (performance purpose, disk optimization translation admin), we create a new language file only if we think we will rich 1000 lines. I dont think module ftp need a dedicated file multiplicated n times pers language dir for just few entries.
Please keep them into "other" that is dedicated for language files that are too small to justify its own language files

bb2a: OK i understand i change

@eldy it's done

PR #21632, opened 2022-08-01
NEW Can send emails with smtp and oauth2 authentication

fboitel:
### Configuration (with gmail)

First you need to setup the following constant:
Screenshot from 2022-08-01 11-18-54

Then you will need to create activate the OAUTH module and create a token (We have named it 'mails' here).

Finally, you should configure  admin/mails as following: 
Screenshot from 2022-08-01 11-17-53


eldy:
Good for a first version.

I now suggest an enhancement on the setup page :
Under the line SMTP IP, we have SMTP Password.
I suggest to keep this line and add a radiobox
[x] Use a password   or   [ ] Use a OAUth token
When we select Use a password, we have the input box to enter the password (like currently).
When we select the OAuth token, we make the new combo box visible to select the OAuth entry (with a message "You must create an OAUTH entry first" if not done yet).

fboitel:
**New proposition with a radiobox:**
Screenshot from 2022-08-03 19-59-57
Screenshot from 2022-08-03 19-59-44

hregis:
@fboitel  "Google" instead "GOOGLE" is less noisy ! ;-))
Thank you for your work...

fboitel:
> 

Indeed !  That's patched, thanks for the feedback!

htdocs/admin/mails.php
```diff
@@ -613,11 +698,27 @@ function initfields()
 			print '<tr class="oddeven"><td>'.$langs->trans("MAIN_MAIL_SMTPS_ID").'</td><td>'.$conf->global->MAIN_MAIL_SMTPS_ID.'</td></tr>';
 		}
 
+		// AUTH method
+		if (isset($conf->global->MAIN_MAIL_SENDMODE) && in_array($conf->global->MAIN_MAIL_SENDMODE, array('smtps'))) {
```
frederic34, line 702: `if (getDolGlobalString('MAIN_MAIL_SENDMODE') == 'smtps') {`

PR #21633, opened 2022-08-01
FIX : VAT rate recalculation

atm-jpb:
# FIX set a checkbox on clone formquestion
When we clone an invoice for another thirparty that don't come from the same country than the first thirdprty, the VAT rate can change.
So, I think, we must recalculate the vat rate on the clone

Capture d’écran de 2022-08-01 16-33-24


eldy:
@atm-jpb 
I add to revert the PR because it was submitted and merged into v15 branch.
Can you resubmit it into "develop" please. Github does not allow me to move it once merged.

PR #21635, opened 2022-08-02
NEW : conf TIMESPENT_ALWAYS_UPDATE_THM, when it's on we always check …

atm-gauthier:
…current thm of user to update it in task time line


atm-gauthier:
@eldy i don't understand why stickler is angry

image

PR #21638, opened 2022-08-02
FIX : Datepicker year interval

atm-kevin:
# FIX

The standard datepicker does not display on year's date select more than 10 years ago and 10 years in the future

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -6230,8 +6230,8 @@ class Form
 				} elseif ($usecalendar == 'jquery') {
 					if (!$disabled) {
 						// Output javascript for datepicker
-						$minYear = $conf->global->MIN_YEAR_SELECT_DATE ? $conf->global->MIN_YEAR_SELECT_DATE : (date('Y') - 10);
-						$maxYear = $conf->global->MAX_YEAR_SELECT_DATE ? $conf->global->MAX_YEAR_SELECT_DATE : (date('Y') + 10);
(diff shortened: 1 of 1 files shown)


htdocs/core/class/html.form.class.php
```diff
@@ -6230,11 +6230,15 @@ public function selectDate($set_time = '', $prefix = 're', $h = 0, $m = 0, $empt
 				} elseif ($usecalendar == 'jquery') {
 					if (!$disabled) {
 						// Output javascript for datepicker
+						$minYear = $conf->global->MIN_YEAR_SELECT_DATE ? $conf->global->MIN_YEAR_SELECT_DATE : (date('Y') - 10);
```
eldy, line 6233: What happen if we want to use the component to select a birthday in 1980 for example ?

atm-kevin: The current problem is that we have to click several times in a row on the drop-down list to reach a date in the past or into the future.

What do you think would be the best solution to solve this problem?
(default setting of -100 / +100 if the value of the "setup other" is not filled or do you think at another solution?)

eldy: Yes, having a default value of +-100 may be enough. It is still possible to enter the data manually if it  out of this range.

atm-kevin: Thank you for your answer.

Yes, it is still possible to enter the data manually.
I did the modification for the +-100 default value.

PR #21646, opened 2022-08-03
NEW Add DoS security on GETPOST with img Yogosha report 11963

Hystepik:
Add a constant to prevent DoS attack on public pages

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -923,6 +923,10 @@ function sanitizeVal($out = '', $check = 'alphanohtml', $filter = null, $options
 						// Remove the trick added to solve pb with text without parent tag
 						$out = preg_replace('/^<div class="tricktoremove">/', '', $out);
 						$out = preg_replace('/<\/div>$/', '', $out);
+						preg_match_all('/(<img)/', $out, $reg);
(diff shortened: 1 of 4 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -954,6 +954,11 @@ function sanitizeVal($out = '', $check = 'alphanohtml', $filter = null, $options
 
 				// Restore entity &apos; into &#39; (restricthtml is for html content so we can use html entity)
 				$out = preg_replace('/&apos;/i', "&#39;", $out);
+
+				preg_match_all('/(<img)/', $out, $reg);
```
eldy, line 958: Must be preg_match_all('/(<img)/i', $out, $reg)    to be case insensitive

htdocs/core/lib/functions.lib.php
```diff
@@ -954,6 +954,11 @@ function sanitizeVal($out = '', $check = 'alphanohtml', $filter = null, $options
...
+				if (count($reg[0]) > (getDolGlobalInt("MAIN_SECURITY_MAX_IMG_IN_HTML_CONTENT") ? getDolGlobalInt("MAIN_SECURITY_MAX_IMG_IN_HTML_CONTENT") : 1000)) {
```
eldy, line 959: Why not just
`if (count($reg[0]) > (getDolGlobalInt("MAIN_SECURITY_MAX_IMG_IN_HTML_CONTENT", 1000))) {`

PR #21647, opened 2022-08-03
NEW : add service status filter on list

defrance:


htdocs/contrat/list.php
```diff
@@ -774,224 +784,229 @@
 
 while ($i < min($num, $limit)) {
 	$obj = $db->fetch_object($resql);
+	if (($obj->nb_initial > 0 && in_array("1", $search_status))
```
eldy, line 787: Adding the filtering inside the loop make the pagination system not working correctly.
Did you try to add the filter directly into the request using
$sql .= " HAVING nb_initial > 0"
or 
$sql .= " HAVING SUM(".$db->ifsql("cd.statut=0", 1, 0).') > 0"     (condition must be same than the way it is written into the select)

defrance:
what should I do so that the corrections made are taken into account and integrated again? I have the impression that she remains abandoned afterwards
-- 
que dois-je faire pour que les corrections apportés soient de nouveau prise en compte et intégrée? J'ai l'impression qu'elle reste à l'abandon ensuite

htdocs/contrat/list.php
```diff
@@ -108,7 +109,8 @@
 $staticcontratligne = new ContratLigne($db);
 
 if ($search_status == '') {
-	$search_status = 1;
+	// if nothing selected, we select all
+	$search_status = array('1', '2', '3', '4');
```
eldy, line 113: If no filter is selected, we must not include a test or for all values but you must just have no criteria in the sql at all

eldy:

PR #21665, opened 2022-08-05
Fix layout of Incomes-Expenses + Claims-Debts accounting tables

ymollard:
Tested with Incomes-Expenses + Claims-Debts but not tested with $modecompta = "BOOKKEEPING"

### DIFF
--- a/htdocs/compta/resultat/clientfourn.php
+++ b/htdocs/compta/resultat/clientfourn.php
@@ -249,7 +249,7 @@ if ($date_endyear) {
 }
 
 print '<table class="noborder centpercent">';
-print '<tr class="liste_titre">'; 
+print '<tr class="liste_titre">';
 if ($modecompta != 'CREANCES-DETTES') {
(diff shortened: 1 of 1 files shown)


htdocs/compta/resultat/clientfourn.php
```diff
@@ -250,6 +250,9 @@
 
 print '<table class="noborder centpercent">';
 print '<tr class="liste_titre">';
+if ($modecompta != 'CREANCES-DETTES') {
+	print_liste_field_titre('');  // Make 4 columns in total whatever $modecompta is
```
eldy, line 254: We can target to have same nb of columns for both "CREANCES-DETTES" and other modes, but it would be better if the fix is done on line 265 by adding a 
```
else {
print_liste_field_titre(''); // Make 4 columns in total whatever $modecompta is
}
```

So each print with a condition is counterbalanced by a print if condition has the opposite value. Would be better for code readability.

PR #21667, opened 2022-08-05
Fix : Spam db public page yogosha report 11813

Hystepik:
Add support of ip field in public ticket creation + add security to prevent spam on public page


htdocs/public/ticket/create_ticket.php
```diff
@@ -231,6 +232,21 @@
 		$object->type_code = GETPOST("type_code", 'aZ09');
 		$object->category_code = GETPOST("category_code", 'aZ09');
 		$object->severity_code = GETPOST("severity_code", 'aZ09');
+		$object->ip = (empty($_SERVER['REMOTE_ADDR']) ? 'unknown' : $_SERVER['REMOTE_ADDR']);
```
eldy, line 235: Getting ip with $_SERVER['REMOTE_ADDR'] is not reliable (may be modified by some proxy).
You must use 
$object->id = getUserRemoteIP()

PR #21670, opened 2022-08-05
FIX #21668 Disassemble BOM is not handling properly the Weighted average price

### LINKED ISSUE #21668: BUG: Disassemble BOM is not handling properly the Weighted average price
### Bug

In a normal manufacturing MO the sum of all product costs (WAP or cost price or buying price) is being used as stock movement cost for the final product of the MO. This works wonderful and you get WAP for manufactured products. 

In the case you are disassemle a product with WAP then you go from 1 input product to lot of output products, each of the output producs like screws, glue, parts will increase in stock but the cost of the stock movement would be the one of the input product that is a very high one. This will cause suddenly all of the raw materials to have a very high WAP

WAP(PMP) Weighted average price

### Environment Version

15

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

The expected behavior would be to prepopulate all the fields of cost with the latest WAP of the raw materials. The user anyway if it has right to change the cost has right to override this cost 

To reproduce just create and MO of type disassemble and make sure the input product has cost (either WAP, cost price, buying price)

### Attached files

dissambly issue

elsp1991:
In a normal manufacturing MO the sum of all product costs (WAP or cost price or buying price) is being used as stock movement cost for the final product of the MO. This works wonderful and you get WAP for manufactured products.

In the case you are disassemle a product with WAP then you go from 1 input product to lot of output products, each of the output producs like screws, glue, parts will increase in stock but the cost of the stock movement would be the one of the input product that is a very high one. This will cause suddenly all of the raw materials to have a very high WAP

This fix is ignoring the price of the input product in case of MO type disassemble

### DIFF
--- a/htdocs/mrp/mo_production.php
+++ b/htdocs/mrp/mo_production.php
@@ -1169,7 +1169,12 @@ if ($object->id > 0 && (empty($action) || ($action != 'edit' && $action != 'crea
 					print '<td class="right">'.$line->qty.'</td>';
 					if ($permissiontoupdatecost) {
 						// Defined $manufacturingcost
+						if ($object->mrptype==0) {  // manufacturing
 						$manufacturingcost = $bomcost;
+						}
+						else { 						// disassemble
+							$manufacturingcost = NULL;
+						}
 						if (empty($manufacturingcost)) {
 							$manufacturingcost = price2num($tmpproduct->cost_price, 'MU');
 						}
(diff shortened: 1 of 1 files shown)


eldy:
Fix is good.
Because it is a bug that may corrupt data, i merge it manually (because other old bugs appears on this page when testing your fix) into a lower branch (will be into v16)
See commit 427541beaa3199dbca97b46eca73684f143baa2a that fix the bug and other not reported in look and feel.

PR #21674, opened 2022-08-05
NEW : Factoring and Fix external module elementType limit

john-botella:
# FIX
Module builder or external module element type 32 char limitation

# NEW
Factoring element_element fetch object mechanism to be used in more conditions

create new statics methods 
- getElementType
- extractElementTypeInfosFromString
- getObjectByElementType
- getObjectFromCache

~~Add new class ElementProperties to allow developers to define elements properties for old modules or class override ;-)~~


htdocs/core/class/commonobject.class.php
```diff
@@ -3874,6 +3872,277 @@ public function add_object_linked($origin = null, $origin_id = null, $f_user = n
 		}
 	}
 
+	/**
+	 * return element type string formated like element_element target_type and target_type
```
atm-john, line 3876:
```suggestion
	 * return element type string formated like element_element target_type and source_type
```

htdocs/core/class/commonobject.class.php
```diff
@@ -3874,6 +3872,277 @@ public function add_object_linked($origin = null, $origin_id = null, $f_user = n
...
+	 * Usage : to easily fetch element form element_element with cache
+	 *
+	 * @param DoliDB $db the DoliDB
+	 * @param string $elementType    Type of object ('invoice', 'order', 'expedition_bon', 'myobject@mymodule', ...)
+	 * @param bool   $elementId      Id of element to provide if fetch is needed
+	 * @param int    $maxCacheByType max number of cached element by type
+	 * @return CommonObject object of $elementType, fetched by $elementId
+	 * @throws Exception
+	 */
+	public static function getObjectByElementType($db, $elementType, $elementId = false, $maxCacheByType = 10)
```
atm-john, line 3901: use fetchObjectByElement

htdocs/install/mysql/tables/llx_element_type.sql
```diff
@@ -0,0 +1,33 @@
...
+-- ============================================================================
+
+create table llx_element_type
+(
+  rowid           	integer AUTO_INCREMENT PRIMARY KEY,  
+  fk_element		integer NOT NULL,
+  element_type		varchar(64) NOT NULL,
+  element		    varchar(32) NOT NULL,
+  module_name		varchar(31) NOT NULL,
+  class_path		text NOT NULL,
```
atm-florianm, line 30: D'un point de vue sémantique, je m'attends à ce que `path` contienne un chemin complet (y compris le nom de fichier)
```suggestion
  class_dir		text NOT NULL,
```

htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -77,3 +77,6 @@ ALTER TABLE llx_recruitment_recruitmentcandidature ADD email_date datetime after
 ALTER TABLE llx_ticket ADD email_date datetime after email_msgid;
 
 INSERT INTO llx_const (name, entity, value, type, visible) VALUES ('MAIN_SECURITY_MAX_IMG_IN_HTML_CONTENT', 1, 1000, 'int', 0);
+
+ALTER TABLE llx_element_element CHANGE sourcetype sourcetype VARCHAR(64) NOT NULL;
```
eldy, line 81: To change the type of a field, use syntax
ALTER TABLE llx_table MODIFY COLUMN name varchar(60);
(see header of migration file).
Use for CHANGE is to rename a column.

this rule is for better compatiblity and impact analysis searches.

htdocs/install/mysql/tables/llx_element_type.sql
```diff
@@ -0,0 +1,33 @@
...
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program. If not, see <https://www.gnu.org/licenses/>.
+--
+-- ============================================================================
+-- Table used for relations between elements of different types:
+-- invoice-propal, propal-order, etc...
+-- ============================================================================
+
+create table llx_element_type
```
eldy, line 23: What is the goal of this table ?
The direction we must go is to follow syntax rules (with hack to manage exception that we should remove version after version), instead of adding tables for conventions purposes

atm-john: I working on a element type descriptor storable in database.
It will be usable for getElementProperties (in my TODO)
it's goal is not to create link but describe an element like getElementProperties does.

I named it llx_element_type because in Dolibarr the var I saw was named like this, but we can change this table name as llx_element_properties or an other name more appropriate...

htdocs/install/mysql/migration/16.0.0-17.0.0.sql
atm-john, line 81:
```suggestion
ALTER TABLE llx_element_element MODIFY COLUMN sourcetype VARCHAR(64) NOT NULL;
```

htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -77,3 +77,6 @@ ALTER TABLE llx_recruitment_recruitmentcandidature ADD email_date datetime after
...
+ALTER TABLE llx_element_element CHANGE targettype targettype VARCHAR(64) NOT NULL;
```
atm-john, line 82:
```suggestion
ALTER TABLE llx_element_element MODIFY COLUMN targettype VARCHAR(64) NOT NULL;
```

htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -79,8 +79,14 @@ ALTER TABLE llx_product ADD COLUMN sell_or_eat_by_mandatory tinyint DEFAULT 0 NO
 
 ALTER TABLE llx_recruitment_recruitmentcandidature ADD email_date datetime after email_msgid;
 
+INSERT INTO llx_const (name, entity, value, type, visible) VALUES ('MAIN_SECURITY_MAX_IMG_IN_HTML_CONTENT', 1, 1000, 'int', 0);
```
atm-john, line 82:
```suggestion
```
TODO : investigate why this line disappear in develop

htdocs/core/class/commonobject.class.php
```diff
@@ -3884,6 +3882,19 @@ public function add_object_linked($origin = null, $origin_id = null, $f_user = n
...
 	}
 
+	/**
+	 * return element type string formated like element_element target_type and source_type
+	 * @return string
+	 */
+	public function getElementType()
+	{
+		// Elements of the core modules which have `$module` property but may to which we don't want to prefix module part to the element name for finding the linked object in llx_element_element.
+		// It's because an entry for this element may be exist in llx_element_element before this modification (version <=14.2) and ave named only with their element name in fk_source or fk_target.
```
atm-florianm, line 3892:
```suggestion
		// Elements of the core modules having a `$module` property but for which we may not not want to prefix the element name with the module name for finding the linked object in llx_element_element.
		// It's because existing llx_element_element entries inserted prior to this modification (version <=14.2) may already use the element name alone in fk_source or fk_target (without the module name prefix).
```

htdocs/core/class/elementproperties.class.php
```diff
@@ -0,0 +1,324 @@
...
+	 * @param DoliDB $db the doliDb instance
+	 */
+	public function __construct($db)
+	{
+		$this->db = $db;
+	}
+
+
+	/**
+	 * Load object in memory from the database
```
atm-florianm, line 92:
```suggestion
	 * Load object in memory from the database using either the ID, element type or a custom filter
	 * (parameters are mutually exclusive)
```

htdocs/core/lib/functions.lib.php
```diff
@@ -10831,35 +10832,101 @@ function getElementProperties($element_type)
 		'classfile' => $classfile,
 		'classname' => $classname
 	);
+
+
+
+	if (!class_exists('ElementProperties')) { require_once DOL_DOCUMENT_ROOT . '/core/class/elementproperties.class.php'; }
```
eldy, line 10838: This does not go in the good direction.
We try to have generic code working with "code convention". Introducing sql for a such often used code will allow developer to break code convention.
We should be able to have something working completely with code convention.

htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -113,6 +113,25 @@ ALTER TABLE llx_ticket ADD email_date datetime after email_msgid;
 
 ALTER TABLE llx_cronjob ADD COLUMN pid integer;
 
+
+ALTER TABLE llx_element_element MODIFY COLUMN sourcetype VARCHAR(64) NOT NULL;
+ALTER TABLE llx_element_element MODIFY COLUMN targettype VARCHAR(64) NOT NULL;
+ALTER TABLE llx_c_type_contact MODIFY COLUMN element VARCHAR(64) NOT NULL;
+
+CREATE TABLE llx_element_properties
```
eldy, line 121: Oh my god no !
We work since several year to standardize code to have element = module_name = class_dir = class_name = class_file.
We should never need such a table.

john-botella:
So ?

john-botella:
@eldy ?

eldy:
Sorry John, can you help me to fix conflict ?
Also i have changes into .sql file of migration. Is it done on purpose or this remains from an other PR ?

eldy:
@atm-john can you help me to fix conflict ?

john-botella:
I'll make a new FIX but with several pull requests.
Don't close, I'll do it after the pull requests are created.

PR #21681, opened 2022-08-07
FIX : suppress borring line

defrance:
duplicate "printTabsHead" hook call if present


eldy:
We have a dol_fiche_end() so we must have a dol_fiche_head()

eldy:
the change will break the balance between dol_get_fiche_head() and dol_get_fiche_end(). Musr have 1 end for each head
so seems wrong so i close the PR

PR #21683, opened 2022-08-07
Update import_xlsx.modules.php

### LINKED ISSUE #21474: Error importing products with additional fields
### Bug

When importing products through xlsx file, if this product has additional fields and they are included in the file, it gives the following error in the database.

```
SELECT rowid FROM llx_product_extrafieldsWHERE fk_object = 3940
[DB_ERROR_SYNTAX] You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '= 3940' at line 1
```

I think the error is that in the query, the table name and the WHERE statement are pasted without space.

### Environment Version

15.0.2

### Environment OS

Linux

### Environment Web server

_No response_

### Environment PHP

7.4.29(php-fpm)

### Environment Database

Mysql 5.7

### Environment URL(s)

_No response_

_No response_

1. Add to product one or two additional fields in the modules section.
2. Save a new xlsx product import file. 
3. Fill it with data, including values for the extra fields.
4. Launch product import

### Attached files

_No response_

glu000:
# FIX|Fix #21474[*Error importing products with additional fields*]
When importing products through xlsx file, if this product has additional fields and they are included in the file, it gives the following error in the database.
```
SELECT rowid FROM llx_product_extrafieldsWHERE fk_object = 3940
[DB_ERROR_SYNTAX] You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '= 3940' at line 1
```

### DIFF
--- a/htdocs/core/modules/import/import_xlsx.modules.php
+++ b/htdocs/core/modules/import/import_xlsx.modules.php
@@ -863,7 +863,7 @@ class ImportXlsx extends ModeleImports
 								if (empty($keyfield)) {
 									$keyfield = 'rowid';
 								}
-								$sqlSelect .= "WHERE " . $keyfield . " = " .((int) $lastinsertid);
+								$sqlSelect .= " WHERE " . $keyfield . " = " .((int) $lastinsertid);
 
 								$resql = $this->db->query($sqlSelect);
 								if ($resql) {


glu000:
I don't understand why stickler-ci failed. What do I need to change?

PR #21691, opened 2022-08-09
NEW Can switch product batch management to no management.

atm-gauthier:


atm-gauthier:
Hi @eldy, in this PR i did corrections you asked to @atm-adrien (old PR : https://github.com/Dolibarr/dolibarr/pull/21503).
Now we check llx_product_batch table and loop on results to move batch stock into global stock.

htdocs/product/class/product.class.php
```diff
@@ -1221,6 +1221,41 @@ public function update($id, $user, $notrigger = false, $action = 'update', $upda
...
+					$resql = $this->db->query($sql);
+					if ($resql) {
+						while ($obj = $this->db->fetch_object($resql)) {
+							$value = $obj->qty;
+							$fk_entrepot = $obj->fk_entrepot;
+							$price = 0;
+							$dlc = '';
+							$dluo = '';
+							$batch = $obj->batch;
+							$inventorycode = dol_print_date(dol_now(), '%Y%m%d%H%M%S');
```
eldy, line 1239: Because all the change here is done into 1 similar purpose, all stock movement must have the same inventorycode, so can you move the init of this var before the loop ?

atm-gauthier:
Hi @eldy, you can see here what's added :

First i we stop using serial numbers or batch, this message appears :
image

Second thing if our product status batch was batch, we'll no longer be able to choose serial number, because Dolibarr doesn't know in which serial number dispatch each quantity :
image

eldy:
Ok for first point.

For the second point, 
"Second thing if our product status batch was batch, we'll no longer be able to choose serial number, because Dolibarr doesn't know in which serial number dispatch each quantity"
We should not block the selection of "Use serial". Showing a warning like this should be enough:  "If you currently have a quantity higher or equal to 2 for the product, switching to this choice means you will still have a product with different object with the same serial (when you want it to be unique). The duplicate will remain until an inventory or a manual stock movement to fix this is done."

Switching first to "no" choice, to "delete" existing serial seems is a bad idea as we will loose information. For example we may want to convert lot A into Lot A-1, Lot A-2, ...  and lot B into Lot B into Lot B-1, Lot B-2.
Can you just show a warning whit no data deletion (cleaning of data will be done later by inventory or manual stock move) ?
So only a warning must be shown. This is not a problem for dolibarr there is a protection to disallow any stock movement if product is not unique, except if we make a movement to make the serial unique. So warning is enough.
Can you change this way ?
Also this change the behaviour, so thanks to push it on develop.

htdocs/product/card.php
```diff
@@ -1944,7 +1944,45 @@
 				if ($object->isProduct() || !empty($conf->global->STOCK_SUPPORTS_SERVICES)) {
 					print '<tr><td>'.$langs->trans("ManageLotSerial").'</td><td>';
 					$statutarray = array('0' => $langs->trans("ProductStatusNotOnBatch"), '1' => $langs->trans("ProductStatusOnBatch"), '2' => $langs->trans("ProductStatusOnSerial"));
-					print $form->selectarray('status_batch', $statutarray, (GETPOSTISSET('status_batch') ? GETPOST('status_batch') : $object->status_batch));
+
+					print $form->selectarray('status_batch', $statutarray, $object->status_batch);
```
eldy, line 1948: On error, we should not lose the user initial delection, so can you restore the test on GETPOSTISSET in current line of code ?

PR #21694, opened 2022-08-09
FIX - Edit propal date

BB2A-Anthony:
# FIX|Fix #[Edit propal date]


htdocs/comm/propal/class/propal.class.php
```diff
@@ -2060,8 +2060,11 @@ public function set_date($user, $date, $notrigger = 0)
 
 			$this->db->begin();
 
-			$sql = "UPDATE ".MAIN_DB_PREFIX."propal SET datep = '".$this->db->idate($date)."'";
-			$sql .= " WHERE rowid = ".((int) $this->id)." AND fk_statut = ".self::STATUS_DRAFT;
+			if (isset($this->duree_validite) && !empty($this->fin_validite)) {
```
eldy, line 2063: set_date is a low level method:
So if we call it, we must update the field.
conditions must be moved on caller side, they will depends on context of this caller page.

bb2a: @eldy moved conditions on caller side

htdocs/comm/propal/card.php
```diff
@@ -315,6 +315,9 @@
 		}
 
 		if (!$error) {
+			if (isset($object->duree_validite) && !empty($object->fin_validite)) {
+				$object->fin_validite = $datep + ($object->duree_validite * 24 * 3600);
```
eldy, line 319: So you add code to avoid bad sync between duration of validity and end of validity. Why not.
But imagine, we initialy create a proposal the 1st of january with a validity duration of 10 days. So the end of validity is set during creation at 10th of january. Dolibarr is using this date for alerts.
Imagine the customer ask more delay. So the saler increase the end of validity date to 30th of january. But here, the code seems to overwrite this 30th of january to 10th of january. Am i wrong ?

bb2a: yes it is for a change of date. @eldy for you what would be the best solution?  To update the validity date.  take back the gap between the two dates?  or put options or offer to update the date?

htdocs/comm/propal/class/propal.class.php
eldy: Oops problem in PR, 474 files are modified...

bb2a: @eldy sorry manipulation error I create another PR #21839

BB2A-Anthony:
PR https://github.com/Dolibarr/dolibarr/pull/21839

PR #21695, opened 2022-08-09
fix doc_generic_recruitmentjobposition_odt

frederic34:


cfoellmann:
@frederic34 can you have a look at the output dir of odt output files?
Some issue is open about it.
I think you might have a better and faster look at this

cfoellmann:
@frederic34 sorry for my last post. Just saw it on the desktop

PR #21699, opened 2022-08-10
update code for PHP8 compliance

grandoc:


htdocs/accountancy/admin/fiscalyear_card.php
```diff
@@ -34,7 +34,7 @@
 if ($user->socid > 0) {
 	accessforbidden();
 }
-if (empty($user->rights->accounting->fiscalyear->write)) {
+if (empty($user->hasRight('accounting', 'fiscalyear', 'write'))) {
```
eldy, line 37: Do not use empty on $user->hasRight...  
Use  
`if (!$user->hasRight('accounting', 'fiscalyear', 'write')) {`

htdocs/accountancy/admin/fiscalyear_info.php
```diff
@@ -33,7 +33,7 @@
 if ($user->socid > 0) {
 	accessforbidden();
 }
-if (empty($user->hasRight('accounting', 'fiscalyear', 'write'))) {
+if ($user->hasRight('accounting', 'fiscalyear', 'write')) {
```
fappels, line 36: Should be !

PR #21700, opened 2022-08-10
NEW add oauth2 for swiftmailer

frederic34:


frederic34:
and of course to enable additional scope foor google oauth
image

PR #21710, opened 2022-08-10
NEW make sell or eat by mandatory in product lot

lvessiller-opendsi:
NEW make sell or eat by mandatory in product lot (see PR #21709 for SQL install)

- Add a "make sell-by or eat-by date" mandatory in a product card when product batch is enabled
image

If sell-by or eat-by is mandatory then, we have these fields required when we create or update or product batch :
1) On create
image

2) On update
image

3) And an error message is shown if one of these fields are empty
image

4) We also check these mandatory fields on stock correction
image

5) And check on dispatch from supplier order
image

6) And when reception is enabled we also have this check
image


eldy:
According to screen 4, the DLC and DLUO fields are mandatory.
But we want to reduce the quantity of field to enter. And if the Lot is already filled and the lot already exists, entering DLC/DLUO is useless because we can retreive them from the lot table. Right ?
Do you check that DLC/DLUO is mandatroy after trying to retreive them from the lot (if it exists) and if you block if not found or if not previously defined (what we should do), or do you block as soon as no data are entered on form screen 4 (bad) ?
If no data are entered, we must first try to retreive them from existing table. And if not found or not defined, we can through an error.  And if some data were entered on form, we must check that value found into existing table match the entered values.

I think this check is already implemented into the low level stock move method (called _create). 
It seems you make the check of DLC/DLUO mandatory into each pages. But the DLC/DLUO are really know and completely guessed only into _create method.
So check about field defined or not must also be done only into this method. So call of checkSellOrEatByMandatoryFromProductIdAndDates() should be done only into _create().
The rule should be
DLC/DLUO must not be mandatory if a serial is defined (whatever are options because options are "mandatory to record a movement", not "mandatory to increase you the time to enter data"), but when the move will be recorded into _create(), if mandatory status is set, then after the code that tried to retreived the value and validate them, we must throw an error if they are still empty.

htdocs/core/class/html.form.class.php
```diff
@@ -201,9 +201,10 @@ public function editfieldkey($text, $htmlname, $preselected, $object, $perm, $ty
 	 * @param	string	$formatfunc		Call a specific function to output field in view mode (For example: 'dol_print_email')
 	 * @param	string	$paramid		Key of parameter for id ('id', 'socid')
 	 * @param	string	$gm				'auto' or 'tzuser' or 'tzserver' (when $typeofdata is a date)
+	 * @param	string	$editaction		[=''] use GETPOST default action or set action to edit mode
 	 * @return  string					HTML edit field
 	 */
-	public function editfieldval($text, $htmlname, $value, $object, $perm, $typeofdata = 'string', $editvalue = '', $extObject = null, $custommsg = null, $moreparam = '', $notabletag = 0, $formatfunc = '', $paramid = 'id', $gm = 'auto')
+	public function editfieldval($text, $htmlname, $value, $object, $perm, $typeofdata = 'string', $editvalue = '', $extObject = null, $custommsg = null, $moreparam = '', $notabletag = 0, $formatfunc = '', $paramid = 'id', $gm = 'auto', $editaction = '')
```
eldy, line 207: Why do you need to be able to force the editaction ?

lvessiller-opendsi: @eldy 
When an error occurs the action value is reset and is different of GET / POST action value.
For example in "product lot" card, you can edit "sell by" field and when you click on "Modify" button the GET / POST action is "setsellby" and if an error occurs we want to edit again "sell by" field and show the error message. So the action value is "editsellby" and the GET / POST action value is "setsellby".

lvessiller-opendsi:
> According to screen 4, the DLC and DLUO fields are mandatory. But we want to reduce the quantity of field to enter. And if the Lot is already filled and the lot already exists, entering DLC/DLUO is useless because we can retreive them from the lot table. Right ? Do you check that DLC/DLUO is mandatroy after trying to retreive them from the lot (if it exists) and if you block if not found or if not previously defined (what we should do), or do you block as soon as no data are entered on form screen 4 (bad) ? If no data are entered, we must first try to retreive them from existing table. And if not found or not defined, we can through an error. And if some data were entered on form, we must check that value found into existing table match the entered values.
> 
> I think this check is already implemented into the low level stock move method (called _create). It seems you make the check of DLC/DLUO mandatory into each pages. But the DLC/DLUO are really know and completely guessed only into _create method. So check about field defined or not must also be done only into this method. So call of checkSellOrEatByMandatoryFromProductIdAndDates() should be done only into _create(). The rule should be DLC/DLUO must not be mandatory if a serial is defined (whatever are options because options are "mandatory to record a movement", not "mandatory to increase you the time to enter data"), but when the move will be recorded into _create(), if mandatory status is set, then after the code that tried to retreived the value and validate them, we must throw an error if they are still empty.

For screen 4), DLC or DLUO is not mandatory to put a value, the product lot is automatically found when we put "lot or serial number" in this form. And if you put a wrong DLUO you got an error message : 
image
image

If you don't put DLC or DLUO, the stock is corrected like in this example : 
image
image

And if you got a new lot number, the DLC or DLUO match the rule defined for the product. If the DLC is mandatory you got an error of field required : 
image
image

lvessiller-opendsi:
@eldy 
Have you any suggestion ?
Could you merge it ?

lvessiller-opendsi:
@eldy 
Why this PR is still under discussion ?

lvessiller-opendsi:
@eldy
What about this PR ?

eldy:
On lot and update creation, it is ok. 
However this field must not be mandatory on stock creation. Because we must be able to make a stock creation or transfer by just entering the lot number.
Because we retreive the dates from the lot.
So error must be reported only if 
Field not filled + we can't retreive the date from lot.
Is it what was implemented ?

The control should be implemented into the stock movement method only, once the date has been filled by user or retreived by system so we are sure all movement are done with a date when option date mandatory is defined and we keep agility to make stock using lot (the main goal of using lot nb)

lvessiller-opendsi:
@eldy
I improved this functionality : 
- sell-by and eat-by date are fill automatically in this form when a batch number is found

lvessiller-opendsi:
@eldy
All conflicts resolved

htdocs/product/class/product.class.php
```diff
@@ -2437,7 +2491,7 @@ public function fetch($id = '', $ref = '', $ref_ext = '', $barcode = '', $ignore
 			} else {
 				$sql .= " p.pmp,";
 			}
-			$sql .= " p.datec, p.tms, p.import_key, p.entity, p.desiredstock, p.tobatch, p.batch_mask, p.fk_unit,";
+			$sql .= " p.datec, p.tms, p.import_key, p.entity, p.desiredstock, p.tobatch, p.sell_or_eat_by_mandatory, p.batch_mask, p.fk_unit,";
```
eldy, line 2494: Since version 15, all new features that need change into database structure must be solitted in 2: one PR for sql changes so database change can be validated, and then one PR for code that implement/use the database.

htdocs/core/class/html.form.class.php
```diff
@@ -233,7 +234,10 @@ public function editfieldval($text, $htmlname, $value, $object, $perm, $typeofda
 		if (!empty($conf->global->MAIN_USE_JQUERY_JEDITABLE) && !preg_match('/^select;|day|datepicker|dayhour|datehourpicker/', $typeofdata)) { // TODO add jquery timepicker and support select
 			$ret .= $this->editInPlace($object, $value, $htmlname, $perm, $typeofdata, $editvalue, $extObject, $custommsg);
 		} else {
-			$editmode = (GETPOST('action', 'aZ09') == 'edit'.$htmlname);
+			if ($editaction == '') {
```
eldy, line 237: What is goal/use case of this framework enhancement ? Also because it is a framework enhancemt, can you push it on a separate PR ?

htdocs/product/class/product.class.php
lvessiller-opendsi: And it's in a separate PR #21709 you merged

htdocs/core/class/html.form.class.php
lvessiller-opendsi: @eldy 
I removed the enhancement part as you suggested to me.
I'm going to add a separate PR for editfieldval() method.

lvessiller-opendsi: Separate this PR into PR #23825

lvessiller-opendsi (commented):
resolve conflicts

eldy:
This PR seems too large, and there is still a lot of trouble to explain. The good part of the PR are blocked by the bad part. And discussing about how to make bad part is difficult until the good parts are not merged. 
So i suggest to refactor this PR to split it.

First i suggest to start with a simple PR to introduce the code that check the mandatory of the eat /sell date, so the call to  checkSellOrEatByMandatoryFromProductIdAndDates()
This call/check must be done into the create of stock movement only.  Because this is the only centralized point that can guarantee that the rule for mandatory eat by or sell is followed for each movement whatever is the way we make stock movement (GUI, API, ...). 
An error will be returned if mandatory option not matched. 

Then we will be able to enhance the GUI to report alerts before the technical error occurs. But firstly we should have the control done at the good place.

lvessiller-opendsi:
@eldy 
I have already split this PR in two PRs.
I didn't understand how can I split again this PR.

lvessiller-opendsi:
see PR #24911

PR #21714, opened 2022-08-11
token was not refreshing

frederic34:


frederic34:
@fboitel do you have the same problem with smtps ?

fboitel:
What do you mean by "Not refreshing"? Is it about the expiration of the token or you mean the token doesn't update on mail module when you update it?

frederic34:
If you just send mail, token is not updating when expired

image

with changes made here, it auto update (for swiftmailer)

fboitel:
I get it, indeed, with smtp you have to regenerate manually the token once it expires. I will update this shortly, inspiring by you solution. Thanks for the work !

PR #21722, opened 2022-08-11
NEW show SellBy/EatBy dates for each batch product in shipment card

lvessiller-opendsi:
NEW show SellBy/EatBy dates for each batch product in shipment card
image


htdocs/expedition/card.php
```diff
@@ -1526,6 +1526,12 @@
 										} else {
 											print 'TableLotIncompleteRunRepairWithParamStandardEqualConfirmed';
 										}
+										if (empty($conf->global->PRODUCT_DISABLE_SELLBY)) {
+											print ' - '.$langs->trans("SellByDate").': '.dol_print_date($dbatch->sellby, "day");
```
eldy, line 1530: Sometime the sellby and eatby date are used only for few products and others just use a lot.
Is showing the sellby and eatby title always for each lot even if value is empty required ?
What do you think of adding a test 
`& !empty($dbatch->sellby)`

lvessiller-opendsi: @eldy 
You're right it will be better, so I updated this PR

lvessiller-opendsi:
@eldy 
I have taken into consideration your advice, could you check and merge it ?

PR #21732, opened 2022-08-12
PR improve function library (ergonomics)

SylvainLegrand:


eldy:
Can you split the PR in atomic PR.
One to fix the bug of the phone number for madagascar,  one to add the feature of the new substitution variables,

SylvainLegrand:
Ok in a few minutes.

PR #21735, opened 2022-08-12
FIX : onlineSign.php bad hash construction

Deefaze:
dol_verifyHash delcaration does not match signature.lib hash construction

### DIFF
--- a/htdocs/core/ajax/onlineSign.php
+++ b/htdocs/core/ajax/onlineSign.php
@@ -72,7 +72,7 @@ if ($type == 'proposal') {
 	$securekeyseed = isset($conf->global->PROPOSAL_ONLINE_SIGNATURE_SECURITY_TOKEN) ? $conf->global->PROPOSAL_ONLINE_SIGNATURE_SECURITY_TOKEN : '';
 }
 
-if (empty($SECUREKEY) || !dol_verifyHash($securekeyseed.$type.$ref, $SECUREKEY, '0')) {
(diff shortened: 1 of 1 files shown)


htdocs/core/ajax/onlineSign.php
```diff
@@ -72,7 +72,7 @@
 	$securekeyseed = isset($conf->global->PROPOSAL_ONLINE_SIGNATURE_SECURITY_TOKEN) ? $conf->global->PROPOSAL_ONLINE_SIGNATURE_SECURITY_TOKEN : '';
 }
 
-if (empty($SECUREKEY) || !dol_verifyHash($securekeyseed.$type.$ref, $SECUREKEY, '0')) {
+if (empty($SECUREKEY) || !dol_verifyHash($securekeyseed.$type.$ref.$entity, $SECUREKEY, '0')) {
```
eldy, line 75: I see this line into signature.lib.php:
`$out .= '&securekey='.dol_hash($securekeyseed.$type.$ref, '0');`
So code seems to match.

eldy: I pushed several fixes that fix a lot of other trouble
3d73f210fb8c1e691b68e74afbb8877c1930b7b4
57fea5151d3d25c21111da838ca9a9ef0d437083
61739cc3383b6e7d52b377c92c30f8edffa30d97
a0d56b6db764736c1e6b1bc0cc2b799f1ffc7654

PR #21740, opened 2022-08-13
php V8 warning

defrance:


htdocs/societe/admin/societe.php
```diff
@@ -713,7 +713,7 @@
 	$i++;
 }
 
-if ($conf->accounting->enabled) {
+if (!empty($conf->accounting->enabled)) {
```
eldy, line 716: Note that you can also use now   
`if (isModEnabled('accounting')) {`

PR #21757, opened 2022-08-14
Improve extrafields substitutions

SylvainLegrand:


htdocs/core/lib/functions.lib.php
```diff
@@ -7145,7 +7145,7 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 				if ($object->fetch_optionals() > 0) {
 					if (is_array($extrafields->attributes[$object->table_element]['label']) && count($extrafields->attributes[$object->table_element]['label']) > 0) {
 						foreach ($extrafields->attributes[$object->table_element]['label'] as $key => $label) {
-							$substitutionarray['__EXTRAFIELD_'.strtoupper($key).'__'] = $object->array_options['options_'.$key];
+							$substitutionarray['__EXTRAFIELD_'.strtoupper($key).'__'] = $extrafields->showOutputField($key, $object->array_options['options_'.$key], '', $object->table_element);
```
eldy, line 7148: The showOutputField is a function dedicated to format a date to be output into a HTML content.
But $substitutionarray['__EXTRAFIELD_'.strtoupper($key).'__'] is a variable that is dedicated to contains a brute and not formated value. So we should not apply such a method on the data we want to store into the memory array.

SylvainLegrand:
This function (showOutputField) is already dedicated to formatting many different types of values such as dates but also int, double, boolean, mail, url, phone, price, etc....
In emails, for instance, we may need to use extra fields to print urls, email addresses, formatted phone numbers, etc...

eldy:
> This function (showOutputField) is already dedicated to formatting many different types of values such as dates but also int, double, boolean, mail, url, phone, price, etc.... In emails, for instance, we may need to use extra fields to print urls, email addresses, formatted phone numbers, etc...


> getCommonSubstitutionArray

There is some case where formatting is legitmage (for example when using dol_print_phone()). But in most cases, we should not (for example some html output formating will return html li tags and we need a brut value (the substitution array may be used by any component, including batch process that need not transformed data). So i suggest to manage the case we need a formatting by adding a 
if (type == 'phone') {
$substitutionarray['__EXTRAFIELD_'.strtoupper($key).'__'] = dol_print_phone($object->array_options['options_'.$key]);
}

htdocs/core/lib/functions.lib.php
SylvainLegrand: Now extrafields can receive many different data: checkbox, picklist, dates, price, boolean, etc...
More and more customers like to use extrafields associated with a substitution variable to create an email and/or a PDF and use these substitution variables even in a product description.
With this PR we avoid duplicating control over the data in order to make it readable and usable.

SylvainLegrand:
Now extrafields can receive many different data: checkbox, picklist, dates, price, boolean, etc...
More and more customers like to use extrafields associated with a substitution variable to create an email and/or a PDF and use these substitution variables even in a product description.
With this PR we avoid duplicating control over the data in order to make it readable and usable.

PR #21765, opened 2022-08-16
NEW : add predicted_cost and real_cost on mrp_mo object

atm-jpb:
# NEW | Feature  add predicted_cost and real_cost on mrp_mo object

Add predicted cost and actual cost functionality. 
Add a caculateCostLines function in the mrp_mo object 
Add  a function that returns the unit price of a product 

Capture d’écran de 2022-08-16 17-45-05

we can see a  service condition in  calculatecostLines not implemented. i think this part can be handle after some other pr  in pipe that handle this particular point.


htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -84,3 +84,8 @@ ALTER TABLE llx_societe ADD last_main_doc VARCHAR(255) NULL AFTER model_pdf;
 ALTER TABLE llx_ticket ADD COLUMN ip varchar(250);
 
 ALTER TABLE llx_ticket ADD email_date datetime after email_msgid;
+
+
+ALTER TABLE llx_mrp_mo ADD COLUMN predicted_cost double(24, 8) DEFAULT 0 AFTER qty;
+
+ALTER TABLE llx_mrp_mo ADD COLUMN real_cost double(24, 8) DEFAULT 0 AFTER qty;
```
eldy, line 91: I suppose that
real cost is the "observed price" once production is finished ?

But what is predicted_cost ? Is it an estimation done by end user or a calculation ?

atm-jpb: it's the qty with the flag tocosume in mrp_production

atm-jpb: and it's a  calculation

PR #21766, opened 2022-08-16
Imap by oauth2 v1

fboitel:
# note
This is a first version of email collector using oauth2. This proposition use the external librairy webklex/php-map.

The set up required is the following : 
Screenshot from 2022-08-16 17-47-34
Screenshot from 2022-08-16 17-46-15


eldy:
This is how to solve the travis error:
Into file .travis.yml add after "--exclude htdocs/includes/webmozart" the entry
"--exclude htdocs/includes/webklex"

Also can you try to remove non required files into htdocs/includes/webklex
For example, we found some dirs used for the develop of the external lib.  We only need the external lib files and the licence file to retreive source of this lib but don't need all files used to generate of develop it. We do this only for our own  Dolibarr code.
For example, you can remove "php-unit". I am pretty sur we don't need it.
There is also a lot of subdir into the dir vendor installed by this fucking tool of composer making the list of files and depencies stupidly high. You can at least remove all dir installed for dev purpose... The less subdir not used we have, the better we are...

PR #21750, opened 2022-08-14
NEW: Added "Get lines and Post lines from BOM" at the REST Service

Humml87:
It´s my first small Pull Request in GIT->OpenSource. Please give me an feedback, if it´s necessary to optimize something.

PR #21776, opened 2022-08-17
Clean Code: Remove duplicate code form bom_card.php

Humml87:
With the PR #21750 the function "addLine" from BOM is available. This functions include all the code with necessary to validate, create, recalculate...


Humml87:
Is closed with PR #21804.

PR #21778, opened 2022-08-17
PR for smarter footer and nicer payment page

SylvainLegrand:


eldy:
Can you provide a screenshot of "after the PR" to understand the change ?

SylvainLegrand:
Formulaire de paiement 22-08-2022 14-47-43

eldy:
What is the difference with old code except we are missing the top header ?

Issue #21578, opened 2022-07-26
HRM > Recruitment


cfoellmann:
Fixing up the recruitment module for `17.0` to get out of alpha status.

## Bugs:

- [ ] Recruitment Module : warning #20470
- [x] broken email field `appplication` (fixed #21563)
- [x] numbering `jobs` (fixed #21563)
- [ ] broken establishment field (SEE:  #20464)
- [x] wrong odf output path -- wrong output dir in odt file generation for recruitment #19449
- [ ] Bug Recruitment Modul ... referrer false #18488
- [ ] API endpoints broken
- [ ] Application > Rating not existing
- [ ] Actions are not being recorded
- [ ] add to GDPR deletion schedule
- [ ] enable to set `fk_user_recruiter` on position create form (defaults feature)

## Features:
- [x] advanced numbering `application` (fixed #21563)

## Cleanup (naming):

cfoellmann:
will have a look at #18434

frederic34:
https://github.com/Dolibarr/dolibarr/pull/21695

cfoellmann:
@eldy is using `xyz_card.php` on the top level of modules a hard must now?
for me it seems "more elegant" like this
`recruitment/application/card.php`
`recruitment/application/list.php`
`recruitment/job/card.php`
`recruitment/job/list.php`

cfoellmann:
@eldy how do I need to change the `status` field in the class RecruitmentCandidature to "just" not show it on creation card?
Setting a different status than draft has negative consequences.

eldy:
Yes 
mymodule/myobject/file  is more elegant. But we try to increase the use and compatibility with module builder and 
mymodule/myobject_file   is easier (at least currently) to manage in our low-code / no-code point of view. I don't want to break this before a long time, the modulebuilder needs more stabilisation to be done before doing refactoring.

PR #21796, opened 2022-08-18
Recruitment improvements

cfoellmann:
- rename recruitmentindex.php -> index.php
- fix some UI strings in en_US
- fix pages titles
- mail form sender fetch from constant, recruiter_email field or recruiter (user)

Ref #21578


htdocs/recruitment/recruitmentcandidature_card.php
```diff
@@ -257,13 +257,13 @@
...
-$title = $langs->trans("RecruitmentCandidature");
-$help_url = '';
-llxHeader('', $title, $help_url);
-
 
 // Part to create
 if ($action == 'create') {
+	$title = $langs->trans('NewCandidature');
+	$help_url = '';
+	llxHeader('', $title, $help_url);
```
eldy, line 265: The llxHeader must always be at top (so line 260) because it contains a lot of security features like adding HTTP headers and we must be sure that it is executed for all "if" cases.
A lot of security bug were introduces with previous moves of llxHeader and i started to revert them so we always have a llxHeader at begin of the "/* View" part of a page. This avoid to forgot it in some if cases and avoid to have to repeat it for each if later.

If the title change, we should add the if line 260.

cfoellmann: did move the function up to the top code level.

cfoellmann:
@eldy what do you think about this starting point?
And this renaming https://github.com/Dolibarr/dolibarr/issues/21578#issuecomment-1217751133 ?

PR #21798, opened 2022-08-19
Work for #21780 Add pid field to Cronjob class and store PID on job execution

### LINKED ISSUE #21780: Clean unfinished cronjob
### Feature Request

Clean job processing state when the job is killed and not reported to Dolibarr.

### Use case

When a cron gets killed by the time limit of PHP, a server shutdown or more, the current job gets stuck in the processing state.

This processing state avoids another run of the job until its unset.

Dolibarr should clean unfinished jobs to avoid skipping killed tasks.

### Suggested implementation

Add a `pid` column to the `llx_cronjob` table, this is set at the start of the job with the value `processing` set to `1`.

Create a cleanup job to run every 5 minutes, this search for processing job and if the PID didn't exist in the system, reset the value `processing` to `0` and set `lastresult` to mark the task in error.
If a PID matches, a security should be added to verify if it's not a process having the same ID after the server reboot.

Finally, fire a trigger to allow third party module to clean other parameter if needed (for example: modules that set locking constant).

### Suggested steps

- [x] Store con job PID in the job row
- [x] Cleanup unfinished job
- [x] Set the job last result as an error
- [ ] Fire a trigger to allow third party modules to clean up

fcharlaix-opendsi:
# Work for #21780 Add pid field to Cronjob class and store PID on job execution

Add the new pid column to the Cronjob class internal value and functions

Store the current PID on job execution during run_jobs

I use the CLOSE keyword, but I end up closing the unfinished initial feature request. Should I use another keyword or none ?

### DIFF
--- a/htdocs/cron/class/cronjob.class.php
+++ b/htdocs/cron/class/cronjob.class.php
@@ -718,7 +718,7 @@ class Cronjob extends CommonObject
 		}
 
 		if (empty($this->pid)) {
-			$this->pid = NULL;
+			$this->pid = null;
 		}
 
 		// Check parameters
@@ -939,7 +939,7 @@ class Cronjob extends CommonObject
 		$this->frequency = '';
 		$this->status = 0;
 		$this->processing = 0;
-		$this->pid = NULL;
+		$this->pid = null;
 		$this->fk_user_author = 0;
 		$this->fk_user_mod = 0;
 		$this->note_private = '';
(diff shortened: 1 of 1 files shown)


eldy:
> # CLOSE #21780 Add pid field to Cronjob class and store PID on job execution
> Add the new pid column to the Cronjob class internal value and functions
> 
> Store the current PID on job execution during run_jobs
> 
> I use the CLOSE keyword, but I end up closing the unfinished initial feature request. Should I use another keyword or none ?

If the feature request is not completely finish, you can use no keyword and just a sentence like "Work for #21780" or "Next setp for #21780".   Only PR Starting with CLOSE, FIX, NEW will be retained to be included into the ChangeLog.

PR #21819, opened 2022-08-20
Fix: left and top menu can't create. Now the checks are correct.

Humml87:
Now you can create a left menu.


dolibit-ut:
Hello, we can create left menues in Dolibarr 14 & 15 - without this change. 
IMHO this change does not really make sense, to say if menu is not left to generate a error for left menu.  ?

Humml87:
Hello, thank you for the feedback. In the new code the checks are working with GETPOST (in the old code with $_POST). Now i added on the GETPOST the Parameter "Methode 3" to get the post value.

PR #21821, opened 2022-08-21
NEW Accountancy - Manage supplier deposit with specific account FPC21+

aspangaro:
image

### DIFF
--- a/htdocs/accountancy/admin/defaultaccounts.php
+++ b/htdocs/accountancy/admin/defaultaccounts.php
@@ -128,6 +128,11 @@ if ($action == 'update') {
 	}
 
 	$constname = 'ACCOUNTING_ACCOUNT_CUSTOMER_DEPOSIT';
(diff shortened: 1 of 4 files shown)


htdocs/accountancy/admin/defaultaccounts.php
```diff
@@ -128,6 +128,7 @@
 	}
 
 	$constname = 'ACCOUNTING_ACCOUNT_CUSTOMER_DEPOSIT';
+	$constname = 'ACCOUNTING_ACCOUNT_SUPPLIER_DEPOSIT';
```
eldy, line 131: This erase previous constant ACCOUNTING_ACCOUNT_CUSTOMER_DEPOSIT. Do we forgot it ?

aspangaro: Fixed. We need the two constants eventually

PR #21839, opened 2022-08-22
FIX - Edit propal date

BB2A-Anthony:
# FIX|Fix #[*FIX - Edit propal date*]

#21694

### DIFF
--- a/htdocs/comm/propal/class/propal.class.php
+++ b/htdocs/comm/propal/class/propal.class.php
@@ -2066,7 +2066,7 @@ class Propal extends CommonObject
 
 			$this->db->begin();
 
-			$sql = "UPDATE ".MAIN_DB_PREFIX."propal SET datep = '".$this->db->idate($date)."'";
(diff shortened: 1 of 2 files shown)


htdocs/comm/propal/class/propal.class.php
```diff
@@ -2066,7 +2066,7 @@ public function set_date($user, $date, $notrigger = 0)
 
 			$this->db->begin();
 
-			$sql = "UPDATE ".MAIN_DB_PREFIX."propal SET datep = '".$this->db->idate($date)."'";
+			$sql = "UPDATE ".MAIN_DB_PREFIX."propal SET datep = '".$this->db->idate($date)."',fin_validite = '".$this->db->idate($this->fin_validite)."'";
```
eldy, line 2069: In all other set_date method, only the main date of object is modified.
So i suggest to not modify the "fin_validite" date here but move this inside caller, just after the ->set_date by doing a complementary update.

bb2a: @eldy OK i have changed

BB2A-Anthony:
@eldy ok for you?

PR #21846, opened 2022-08-23
php V8 warning, import max size not used

defrance:


htdocs/imports/import.php
```diff
@@ -624,9 +625,9 @@
 	print '<div class="marginbottomonly">';
 	$maxfilesizearray = getMaxFileSizeArray();
 	$maxmin = $maxfilesizearray['maxmin'];
-	if ($maxmin > 0) {
-		$texte .= '<input type="hidden" name="MAX_FILE_SIZE" value="'.($maxmin * 1024).'">';	// MAX_FILE_SIZE must precede the field type=file
-	}
+	//if ($maxmin > 0) {
```
eldy, line 628: Why removing ? This is used by the browser to return some warnings/alerts

defrance: 1 - $texte var is not well filled  ( $texte .= instead $texte =)
2 - $texte var is not used

maybe replace it by a print or removing, i'm not sure but for the moment this is not used (so we can remove it)

eldy: So $texte .= ... must be replaced with print ...

PR #21850, opened 2022-08-23
Move up the Parent Company field in the create form

habot-it:
Thanks in advance ;)

Best regards,
Alban


eldy:
What is goal ?
We try to keep at top the most frequent field used. 
Parent company is not so often used. Also we try to keep same position of field in create form that we have in the view page.

habot-it:
Hello @eldy,

My apologies for this incomplete pull request...

Thank you for your hard work !
Best regards,
Alban

PR #21855, opened 2022-08-23
add categories type for TASKs

dolibit-ut:
first step to add categories for tasks


htdocs/categories/class/categorie.class.php
```diff
@@ -151,7 +155,8 @@ class Categorie extends CommonObject
 		'actioncomm' => 'ActionComm',
 		'website_page' => 'WebsitePage',
 		'ticket' => 'Ticket',
-		'knowledgemanagement' => 'KnowledgeRecord'
+		'knowledgemanagement' => 'KnowledgeRecord',
+		'task' => 'Task';
```
frederic34, line 159: @dolibit-ut syntax error here

frederic34: @dolibit-ut can you fix the syntax error please ?

eldy:
@dolibit-ut can you fix conflict ?

eldy:

PR #21861, opened 2022-08-24
FIX php8 compatibility

grandoc:


htdocs/admin/const.php
```diff
@@ -185,12 +185,12 @@
 print '<div class="div-table-responsive-no-min">';
 print '<table class="noborder centpercent">';
 print '<tr class="liste_titre">';
-print getTitleFieldOfList('Name', 0, $_SERVER['PHP_SELF'], 'name', '', $param, '', $sortfield, $sortorder, '')."\n";
+print getTitleFieldOfList('Name', 0, $_SERVER['PHP_SELF'], 'name', '', $param, '', $sortfield, $sortorder, '') . "\n";
```
eldy, line 188: Note that PSR12 say we must not add spaces before/after the concat symbol .
You should find a setup in your IDE to disable this...

PR #21868, opened 2022-08-24
target the right anchor on POST to addline form

cfoellmann:
On submission of the addline form the resulting URL should target the `#addline` anchor (css id) added to the tpl

That way after adding a new line the form gets focused again

### DIFF
--- a/htdocs/comm/propal/card.php
+++ b/htdocs/comm/propal/card.php
@@ -2704,7 +2704,14 @@ if ($action == 'create') {
 	global $inputalsopricewithtax;
 	$inputalsopricewithtax = 1;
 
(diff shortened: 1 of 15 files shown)


eldy:
Is it to have the page at the same position after submitting a form ? If yes, we have abandonned the use of anchor. Instead we use the css "reposition" on a link and input button so page remain at same position.

cfoellmann:
Then it is not working what you have tried to implement.
The anchor #line_* is present so it is just using the same technique.

Something working needs to be implemented. My users are going crazy on items with a lot of lines

cfoellmann:
I think a class for pages is overdue.
Fixes or changing stuff All over the place is soo hard to get right completely if not impossible.

An extendable class for pages is the way to got. Still old school but a step into the right direction

eldy:
Can you describe the process to reproduce the trouble ?
I tested on invoice on develop version, i have no trouble, after submit a new line to add on invoice, screen and form remains at same position

cfoellmann:
I am sorry for the trouble I am causing from time to time.
It is a problem with a module I should not have tested on my dev test environment

PR #21869, opened 2022-08-24
Imap by oauth2 attachments

fboitel:
When creating ticket or project for collected mails, attachments are saved automatically in the corresponding file in htdocs/documents to be able to retreive them when consutling tickets/project in question.


htdocs/emailcollector/class/emailcollector.class.php
```diff
@@ -2440,6 +2442,20 @@ public function doCollectOneCollector()
 										$errorforactions++;
 										$this->error = 'Failed to create project: '.$langs->trans($projecttocreate->error);
 										$this->errors = $projecttocreate->errors;
+									} else {
+										if ($attachments) {
+											$destdir = $conf->project->dir_output.'/'.$projecttocreate->ref;
+											if (!dol_is_dir($destdir)) {
+												dol_mkdir($destdir);
+											}
+											if (!empty($conf->global->MAIN_IMAP_USE_PHPIMAP) && $this->acces_type == 1) {
```
eldy, line 2451: No more need to do
`if (!empty($conf->global->MAIN_IMAP_USE_PHPIMAP) && $this->acces_type == 1) {`
You can just do
`if (!empty($conf->global->MAIN_IMAP_USE_PHPIMAP)) {`

Because when MAIN_IMAP_USE_PHPIMAP,  acces_type = 0 and acces_type = 1 are both implemented (I added the case acces_type = 0 recently) and they are both working correctly the same way.

PR #21910, opened 2022-08-26
Membership fixes

ymollard:
# FIX | Allow sorting non-integer values as numeric values
Allows to sort membership lists by their numerical ID (previously, string sorting applied, so after 1 we got 10)

# FIX | Fix $backtopage after member deletion
Go back to the list of members instead of an invalid card (because deleted)

# FIX | Fixed duration add for leap years + corresponding unit test
Convert 1y to 12m so that the `$ruleforendofmonth` also apply to additions made in years.

# FIX | Skip payment if amount = 0 
Subscriptions with amount = 0 will be recorded as drafts for later validation instead.

### DIFF
--- a/htdocs/adherents/list.php
+++ b/htdocs/adherents/list.php
@@ -464,7 +464,7 @@ if (empty($conf->global->MAIN_DISABLE_FULL_SCANLIST)) {
 }
 
 // Complete request and execute it with limit
-$sql .= $db->order($sortfield, $sortorder);
+$sql .= $db->order($sortfield, $sortorder, $sortfield == "d.ref");
 if ($limit) {
 	$sql .= $db->plimit($limit + 1, $offset);
 }
--- a/htdocs/core/db/DoliDB.class.php
+++ b/htdocs/core/db/DoliDB.class.php
@@ -264,7 +264,7 @@ abstract class DoliDB implements Database
 	 * @param   bool 		$isnumeric 		Sort order must be interpreted as a numeric value (integer)
 	 * @return	string						String to provide syntax of a sort sql string
 	 */
-	public function order($sortfield = null, $sortorder = null, $isnumeric=false)
+	public function order($sortfield = null, $sortorder = null, $isnumeric = false)
 	{
 		if (!empty($sortfield)) {
 			$oldsortorder = '';
@@ -279,9 +279,9 @@ abstract class DoliDB implements Database
 					$return .= ', ';
 				}
 
-				if($isnumeric) $return .= 'CAST(';
+				if ($isnumeric) $return .= 'CAST(';
 				$return .= preg_replace('/[^0-9a-z_\.]/i', '', $val); // Add field
-				if($isnumeric) $return .= ' AS UNSIGNED)';
+				if ($isnumeric) $return .= ' AS UNSIGNED)';
 
 				$tmpsortorder = (empty($orders[$i]) ? '' : trim($orders[$i]));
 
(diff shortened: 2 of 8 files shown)


htdocs/adherents/list.php
```diff
@@ -464,7 +464,7 @@
 }
 
 // Complete request and execute it with limit
-$sql .= $db->order($sortfield, $sortorder);
+$sql .= $db->order($sortfield, $sortorder, $sortfield == "d.ref");
```
eldy, line 467: You mean 
$sortfield = "d.ref"
?

htdocs/core/db/DoliDB.class.php
```diff
@@ -278,7 +279,9 @@ public function order($sortfield = null, $sortorder = null)
 					$return .= ', ';
 				}
 
+				if ($isnumeric) $return .= 'CAST(';
```
eldy, line 282: Using CAST UNSIGNED may not be compatible with other database like pgsql.
What is the goal of this ?

htdocs/core/lib/date.lib.php
```diff
@@ -122,6 +122,12 @@ function dol_time_plus_duree($time, $duration_value, $duration_unit, $ruleforend
 {
 	global $conf;
 
+	if ($ruleforendofmonth == 1 && $duration_unit == 'y') {
```
eldy, line 125: duration_unit = y seems supported line 160.
Do you have an example of case that is wrong with current code so we can add it into phpunit test cases ?

htdocs/adherents/list.php
ymollard: No, we're passing a boolean as a parameter of a function, that is true only if $sortfield == "d.ref", false otherwise.

htdocs/core/db/DoliDB.class.php
ymollard: Is INTEGER compatible with MySQL? I get:
```
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'INTEGER) DESC LIMIT 26' at line 1
```

eldy: I do not experience trouble on mariadb with 
`select CAST("123" AS INTEGER) from llx_const;`
Which version of mysql do you have ?

ymollard: MySQL 8.0.30
CAST AS INTEGER looks [supported by MariaDB](https://mariadb.com/kb/en/cast/) but [not MySQL](https://dev.mysql.com/doc/refman/8.0/en/cast-functions.html#function_cast).
Should we make a difference according to the RDBMS with INTEGER for everything but MySQL?

frederic34:
Member Refs now have a numbering module which can be ordered alphabetically.

eldy:
According to @frederic34 feedback, this pr is now depecated, so i close it.

ymollard:
> Member Refs now have a numbering module which can be ordered alphabetically.

@frederic34 could you please elaborate this topic? In which commit has it been fixed?

frederic34:
https://github.com/Dolibarr/dolibarr/commit/690562c22ef7d1e9d49054f21d60bf3663fb7789

ymollard:
Thank you @frederic34 Does that commit fix the sortering order of members?

Here is a view from the member view of 18.0.4, it still shows an incorrect sorting order:
image

frederic34:
what is your setup in the module member ?

ymollard:
What setting are you looking for @frederic34?
image

frederic34:
image
with mod_simple you can have 0001, 0002, 0003 etc
you can adjust the left pad with MEMBER_MOD_SIMPLE_LPAD (size of the ref, here 4)

ymollard:
Would activating "Simple" by default be a good idea?

Activating the Simple numbering this way still does not class members properly:
image


PS: in our first member, rowid is 131 but ref is 1. I am missing something here.

frederic34:
add MEMBER_MOD_SIMPLE_LPAD=3 if you want 001, 002 in configuration/divers

PR #21915, opened 2022-08-27
NEW : use badge for limittoshow tabs instead parenthesis

defrance:


htdocs/core/lib/functions.lib.php
```diff
@@ -1950,7 +1950,7 @@ function dol_get_fiche_head($links = array(), $active = '', $title = '', $notab
 			$tabsname = str_replace("@", "", $picto);
 		}
 		$out .= '<div id="moretabs'.$tabsname.'" class="inline-block tabsElem valignmiddle">';
-		$out .= '<div class="tab"><a href="#" class="tab moretab inline-block tabunactive"><span class="hideonsmartphone">'.$langs->trans("More").'</span>... ('.$nbintab.')</a></div>'; // Do not use "reposition" class in the "More".
+		$out .= '<div class="tab tabunactive"><a href="#" class="tab moretab inline-block"><span class="hideonsmartphone">'.$langs->trans("More").'</span><span class="badge marginleftonlyshort">'.$nbintab.'</span></a></div>'; // Do not use "reposition" class in the "More".
```
eldy, line 1953: The css "badge" is currently used to highlight a quantity of a data in database (related to text before). Here we want a number of menu entries. So i suggest to introduce another css
For example  "badgemenu" instead of "badge".

defrance: I quite agree with the idea of using two different styles of badge because it does not correspond to the same thing.
However after analyzing other examples of tabs with a grouping, I see that they do not use a totalization and I wonder if we should not simply do without this function

eldy:
A good practice looks like using the label "More" or "..." (better on small screen and verbose language) followed by the down arrow like in this examples:
https://experience.sap.com/fiori-design-web/icontabbar/

PR #21929, opened 2022-08-27
Update modSociete.class.php

dolibit-ut:
'thirdparty_customer_advance'


eldy:
If we change into definition of descriptor file, we must also change:
- into table where the key is already stored (column perms or subperms llx_rights_def) so by sql code into migration script
- the existing code that use this permission

PR #21942, opened 2022-08-29
NEW manage no email with thirdparties (better for GDPR)

ptibogxiv:


htdocs/core/modules/mailings/thirdparties.modules.php
```diff
@@ -226,7 +226,7 @@ public function getNbOfRecipients($sql = '')
 		$sql .= " FROM ".MAIN_DB_PREFIX."societe as s";
 		$sql .= " WHERE s.email <> ''";
 		$sql .= " AND s.entity IN (".getEntity('societe').")";
-
+		$sql .= " AND (SELECT count(*) FROM ".MAIN_DB_PREFIX."mailing_unsubscribe WHERE email = s.email) = 0";
```
eldy, line 229: This force database engine to make a heavy subquery for each line.Not good on large database.
Can you try this instead :
$sql .= " AND EXISTS (SELECT rowid FROM ".MAIN_DB_PREFIX."mailing_unsubscribe as mu WHERE mu.email = s.email and mu.entity = ".((int) $conf->entity).")";
The explain plan should show that the database make an index search on email instead of a subquery.

ptibogxiv: @eldy functionnal but just invert it !

i pushes a commit in this PR with this

PR #21943, opened 2022-08-29
fix #21941 call_trigger

### LINKED ISSUE #21941: workflow : public online sign does not call PROPAL_CLOSE_SIGNED trigger
### Bug

You have to enable workflow module AND activate "auto create order when propal is signed".

Then if you clic on dolibarr backend button "validate" propal it make an order -> perfect.

BUT if you send to your customer the online "sign" link -> the trigger is not run then workflow does not work.

### Environment Version

_No response_

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

rycks:
# FIX #21941 : call trigger on public propal sign

### DIFF
--- a/htdocs/public/onlinesign/newonlinesign.php
+++ b/htdocs/public/onlinesign/newonlinesign.php
@@ -468,7 +468,6 @@ if ($action == "dosign" && empty($cancel)) {
 			print '<br>';
 			if ($message == 'refused') {
 				print '<span class="ok">'.$langs->trans("PropalRefused").'</span>';
-
 			} else {
 				print '<span class="warning">'.$langs->trans("PropalAlreadyRefused").'</span>';
 			}
--- a/htdocs/public/onlinesign/newonlinesign.php
+++ b/htdocs/public/onlinesign/newonlinesign.php
@@ -168,6 +168,15 @@ if ($action == 'confirm_refusepropal' && $confirm == 'yes') {
 
 		$message = 'refused';
 		setEventMessages("PropalRefused", null, 'warnings');
+		if (method_exists($object, 'call_trigger')) {
+			//customer is not a user !?! so could we use same user as validation ?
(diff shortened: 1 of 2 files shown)


htdocs/public/onlinesign/newonlinesign.php
```diff
@@ -452,13 +452,34 @@
 			print '<br>';
 			if ($message == 'signed') {
 				print '<span class="ok">'.$langs->trans("PropalSigned").'</span>';
+				//fix #21941 run trigger
+				if (method_exists($object, 'call_trigger')) {
```
eldy, line 456: Why calling the method into the public page and not into the method that set the status so into 
core/ajax/onlineSign.php

rycks: why .... hmmm because i have to make a choice 🔀  but it seems to be the bad one i've made, sorry 👯

rycks: so ? would you prefer that i switch that code from newonlinesign.php to ajax/onlineSign.php ?

eldy: Yes trigger must be nearest the code that justify its execution.

rycks: yes sir ! i will do it now

rycks: ok, i've moved one trigger to ajax/onlineSign.php but for the other (refused) there is no other place than newonlinesign.php
... sorry :)

commit & PR update will become in some sec / min depends on my git-power !

rycks: PR is clean and ready :)

PR #21946, opened 2022-08-29
Update websiteaccount_card.php

dolibit-ut:


htdocs/website/websiteaccount_card.php
```diff
@@ -43,7 +43,7 @@
 $object = new SocieteAccount($db);
 $extrafields = new ExtraFields($db);
 $diroutputmassaction = $conf->website->dir_output.'/temp/massgeneration/'.$user->id;
-$hookmanager->initHooks(array('websiteaccountcard')); // Note that conf->hooks_modules contains array
+$hookmanager->initHooks(array('websiteaccountcard'));   //  Note that conf->hooks_modules contains array
```
eldy, line 46: Note that adding a space is useless because tool to sanitize and standardize code and a lot of IDE remove them to keep only 1 space. So this change will probably be reverted by another PR done by someone else.

htdocs/website/websiteaccount_card.php
```diff
@@ -32,9 +32,9 @@
 $langs->loadLangs(array("website", "other"));
 
 // Get parameters
-$id = GETPOST('id', 'int');
+$id         = GETPOST('id', 'int');
```
eldy, line 35: Note that adding a space is useless because tool to sanitize and standardize code and a lot of IDE remove them to keep only 1 space. So this change will probably be reverted by another PR done by someone else.

dolibit-ut: unterstood
but that is a pity, because it allows better & faster recognition of patterns and errors

PR #21983, opened 2022-08-30
jump to create if $action empty

cfoellmann:
@eldy maybe this is an option to tackle the empty $action issue!?

### DIFF
--- a/htdocs/fourn/facture/card.php
+++ b/htdocs/fourn/facture/card.php
@@ -1901,8 +1901,1265 @@ if ($action == 'create') {
 $help_url = 'EN:Module_Suppliers_Invoices|FR:Module_Fournisseurs_Factures|ES:Módulo_Facturas_de_proveedores|DE:Modul_Lieferantenrechnungen';
 llxHeader('', $title, $help_url);
 
+// View or edit mode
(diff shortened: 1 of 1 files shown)


eldy:
What is the trouble with empty $action ?   action is defined 
- when there is an action to do. If not, it is just a read.
- or to ask a page in a different mode than a simple view, for example to show it in mode create or edit.

cfoellmann:
@eldy problem is if there is NO parameter in the url for a card.php it results in a white page.

eldy:
> @eldy problem is if there is NO parameter in the url for a card.php it results in a white page.

Yes.
According to module builder, a card page return the file if an id=X or ref=Y is provided, return a creation page if action=create is provided and return an empty page if page is called without parameter (should not happen in app context, and we do the less possible if the action=create is not provided).

cfoellmann:
I understand you but a user should never see a whitescreen
How does he know what happened?

eldy:
> I understand you but a user should never see a whitescreen How does he know what happened?

He should never see it. The only case where he will see it if an attempt to hack app by removing parameters or a bug inside app. Am i wrong ?

PR #21989, opened 2022-08-30
NEW : Column Montant HT on Expedition list

ghost:
# NEW | New [*Addition of the column Montant HT on the list of expeditions]
[*The column values ​​are calculated in relation to the sum of the product between the price of the service associated with the line and the quantity to be shipped*]

### DIFF
--- a/htdocs/expedition/list.php
+++ b/htdocs/expedition/list.php
@@ -411,7 +411,13 @@ $parameters = array();
 $reshook = $hookmanager->executeHooks('printFieldListWhere', $parameters); // Note that $action and $object may have been modified by hook
 $sql .= $hookmanager->resPrint;
 
-$sql .= ' GROUP BY e.rowid';
+$sql .= ' GROUP BY e.rowid, e.ref, e.ref_customer, e.date_expedition, e.weight, e.weight_units, ';
+$sql .= 'e.date_delivery , e.fk_statut, e.billed, e.tracking_number, e.fk_shipping_method, ';
+$sql .= 'l.date_delivery, s.rowid, s.nom, s.town, s.zip, s.fk_pays, s.client, s.code_client, ';
+$sql .= 'typent.code, state.code_departement, state.nom, e.date_creation, e.tms, u.login';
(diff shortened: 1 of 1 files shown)


htdocs/expedition/list.php
```diff
@@ -260,7 +263,7 @@
 if ($sall || $search_product_category > 0 || $search_user > 0) {
 	$sql = 'SELECT DISTINCT';
 }
-$sql .= " e.rowid, e.ref, e.ref_customer, e.date_expedition as date_expedition, e.weight, e.weight_units, e.date_delivery as delivery_date, e.fk_statut, e.billed, e.tracking_number, e.fk_shipping_method,";
+$sql .= " e.rowid, e.ref, e.ref_customer, e.date_expedition as date_expedition, e.weight, SUM(pd.subprice*ed.qty) as MontantHT, e.weight_units, e.date_delivery as delivery_date, e.fk_statut, e.billed, e.tracking_number, e.fk_shipping_method,";
```
eldy, line 266: Not so easy.
The SUM here is not compatible with rest of sql because when 
$sall, $search_product_category or $search_user is defined, lines are duplicated and then filtered by a distinct. This make the sum counting several times the amount in line (distinct is done after the sum).
Before being able to add a join on product line grouping by a sum, we must first obtain a sql request than do not use any more the distinct.
This can be done by a PR that filter on category doing a 
`AND EXISTS (SELECT rowid from .MAIN_DB_PREFIX.categorie_societe as cs ON s.rowid = cs.fk_soc AND cp.fk_categorie = ".((int) $search_product_category))`
instead of introducing a join that duplicate record.
With this, we have no more reason to include a distinct to get the list of shipment.

For the search on user, the distinct is useless because the join does not generate duplication of lines.

For the search on sall, the distinct beome useless too because of the group by that you introduce.

eldy:
Need to wait that filters that was implemented by doing a left join (like filter on categories or users) are replaced by a simple condition into the where to allow the use of the aggregate function required by this PR without side effect.

eldy:
A lot of changes were done in v20 to make the sql request simpler. But this make this PR no more valid with the new code so i close the ticket.

PR #21992, opened 2022-08-30
NEW : Product supplier price: autofill default supplier VAT

thomas-Ngr:
# NEW Product supplier price: autofill default supplier VAT

On product supplier price creation, when one chooses a supplier, the default VAT rate is not updated.
To reproduce:
- create a supplier with tva_assuj = 0
- navigate to any product, supplier price ( htdocs/product/fournisseurs.php?id=XXX&action=add_price )
- chose the supplier -> the VAT stays to default (20).

This PR adds an AJAX call to update the VAT rate that matches the rules for the supplier.

### DIFF
--- a/htdocs/product/fournisseurs.php
+++ b/htdocs/product/fournisseurs.php
@@ -506,6 +506,43 @@ if ($id > 0 || $ref) {
 							print '</a>';
 						}
 					}
+					print '<script type="text/javascript">
+					$(document).ready(function () {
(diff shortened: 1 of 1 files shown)


thomas-Ngr:
Whe have the correct vat rate for the supplier when we request core/ajax/vatrates.php.

core/ajax/vatrates.php returns an HTML select list, with an option 'selected' on the correct vat rate option.
The JS code extracts the correct value to automatically fill the vat input field.

PR #22010, opened 2022-09-01
Add the VAT calculation mechanism according to the 2 possible modes f…

SylvainLegrand:


eldy:
When using a computed accountancy the only way to have reliable accountancy at a line level is to make rounding on line level.
For supplier, we don't control the supplier and some of them are still making accounting using excel sheet allowing them to make rounding on final total.  That is why we provide the alternative method to match the bad practice of supplier.
However, if a user is using Dolibarr, we know it is using a numeric accounting, and in such a case we must never allow to make rounding of final total because we know such way of calculation will give wrong result and bad consistency in total.
So providing such a feature to the dolibarr user is clearly introducing a bug. If a customer is not ready to make rounding on line level, it means he is not ready to make numeric recording of invoices, so not ready to use Dolibarr. Only the mode rounding on line can give the accurate storing of data that Dolibarr needs to have an accountancy and report reliable.

philippe-opendsi:
Hi @eldy we have a use case when order comes from an e-commerce software. We need to reflect in Dolibarr'invoices the price paid by the custommer "online".  Without such option, we need to adjust the unit price and pray to have the same ttc amount in Dolibarr....

PR #22011, opened 2022-09-01
Add an option to manage input size for customer / supplier references

SylvainLegrand:


eldy:
What is the goal ?  To be able to define a maximum length of the customer ref ?

SylvainLegrand:
The purpose is only ergonomics.
Some users of Dolibarr ask me for that because sometimes they need to input some long customer reference and they can't read it easily when they write it.

PR #22019, opened 2022-09-01
Fix multicompany compatibility

hregis:


htdocs/core/class/commondocgenerator.class.php
```diff
@@ -171,7 +171,7 @@ public function getSubstitutionarrayMember($member, $outputlangs)
 	public function get_substitutionarray_mysoc($mysoc, $outputlangs)
 	{
 		// phpcs:enable
-		global $conf;
+		global $conf, $object;
```
eldy, line 174: get_substitutionarray_mysoc is a method to make replacement of var that depends on the company and only the company, so $mysoc.
I see no reason here to have a global var $object. No external object should have effect on properties of a company.

May be you need   $mysoc->entity ?

hregis: @eldy 
With multicompany, if you activate the sharing of commercial proposals for example, and you want to generate an ODT of a commercial proposal of an entity from another entity, the "mycompany_logo" tag does not work !

hregis:
@eldy ok, i add a substitutions library in mutlicompany module

PR #22020, opened 2022-09-01
FIX : Recursively retrieves all parent bom in the tree that leads to …

atm-gauthier:
…the $bom_id bom to check if we're allowed to add bom as bomline

### DIFF
--- a/htdocs/bom/class/bom.class.php
+++ b/htdocs/bom/class/bom.class.php
@@ -1147,6 +1147,34 @@ class BOM extends CommonObject
 			}
 		}
 	}
+
+	/**
+	 * Recursively retrieves all parent bom in the tree that leads to the $bom_id bom
+	 *
+	 * @param 	array	$TParentBom		We put all found parent bom in $TParentBom
+	 * @param 	int		$bom_id			ID of bom from which we want to get parent bom ids
+	 * @return 	void
+	 */
+	public function getParentBomTreeRecursive(&$TParentBom, $bom_id) {
+
+		$sql = 'SELECT l.fk_bom, b.label
+				FROM '.MAIN_DB_PREFIX.'bom_bomline l
+				INNER JOIN '.MAIN_DB_PREFIX.$this->table_element.' b ON b.rowid = l.fk_bom
+				WHERE fk_bom_child = '.$bom_id;
+
+		$resql = $this->db->query($sql);
+		if(!empty($resql)) {
+
+			while($res = $this->db->fetch_object($resql)) {
+
+				$TParentBom[$res->fk_bom] = $res->fk_bom;
+				$this->getParentBomTreeRecursive($TParentBom, $res->fk_bom);
+
+			}
+
+		}
+
+	}
 }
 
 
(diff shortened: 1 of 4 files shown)


htdocs/bom/class/bom.class.php
```diff
@@ -1147,6 +1147,30 @@ public function getNetNeedsTree(&$TNetNeeds = array(), $qty = 0, $level = 0)
...
+	 * @param 	int		$bom_id			ID of bom from which we want to get parent bom ids
+	 * @return 	void
+	 */
+	public function getParentBomTreeRecursive(&$TParentBom, $bom_id)
+	{
+
+		$sql = 'SELECT l.fk_bom, b.label
+				FROM '.MAIN_DB_PREFIX.'bom_bomline l
+				INNER JOIN '.MAIN_DB_PREFIX.$this->table_element.' b ON b.rowid = l.fk_bom
+				WHERE fk_bom_child = '.$bom_id;
```
frederic34, line 1164: @atm-gauthier add int cast here

htdocs/bom/class/bom.class.php
```diff
@@ -1147,6 +1147,30 @@ public function getNetNeedsTree(&$TNetNeeds = array(), $qty = 0, $level = 0)
...
+	{
+
+		$sql = 'SELECT l.fk_bom, b.label
+				FROM '.MAIN_DB_PREFIX.'bom_bomline l
+				INNER JOIN '.MAIN_DB_PREFIX.$this->table_element.' b ON b.rowid = l.fk_bom
+				WHERE fk_bom_child = '.((int) $bom_id);
+
+		$resql = $this->db->query($sql);
+		if (!empty($resql)) {
+			while ($res = $this->db->fetch_object($resql)) {
```
eldy, line 1168: Recursive methods always need a protection against infinite loop.
Imagine data was loaded with a child thas has a child that is its parent, you fall into an infinite loop.
You must add a counter that is increased +1 at each call and we stop the recursion when counter > 1000 (we know the depth of BOM will never be 1000).

atm-gauthier: Hi @eldy, i just did it

htdocs/bom/class/bom.class.php
```diff
@@ -1147,6 +1147,38 @@ public function getNetNeedsTree(&$TNetNeeds = array(), $qty = 0, $level = 0)
...
+	 * @param 	int		$bom_id			ID of bom from which we want to get parent bom ids
+	 * @param 	int		$level		Protection against infinite loop
+	 * @return 	void
+	 */
+	public function getParentBomTreeRecursive(&$TParentBom, $bom_id = '', $level = 1)
+	{
+
+		// Protection against infinite loop
+		if ($level > 1000) {
+			return 0;
```
eldy, line 1164: Not compatible with doc saying   "return void"
Just do "return" or better, change return code in doc into   @return  int

atm-gauthier:
@eldy stickler gives an error but no information about it

eldy:
> @eldy stickler gives an error but no information about it

Don't worry about sticker. Only Travis is important.

atm-gauthier:
@eldy OK, is everything OK now with this pr for you?

PR #22027, opened 2022-09-01
Update webhookindex.php

dolibit-ut:


dolibit-ut:
should we rename file to index.php or webhook.php ?

PR #22042, opened 2022-09-02
FIX: bad bookmark save

### LINKED ISSUE #22040: Integration of filters in bookmarks
### Bug

Take into account variables that pass in POST in the url

### Environment Version

All Dolibarr

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

Sample on projet : 
Step 1) Filter on list
image
Step 2) Add bookmark
image
Step 3) Select bookmark
image

### Attached files

_No response_

FHenry:
fix #22040

### DIFF
--- a/htdocs/bookmarks/bookmarks.lib.php
+++ b/htdocs/bookmarks/bookmarks.lib.php
@@ -36,28 +36,39 @@ function printDropdownBookmarksList()
 
 	$langs->load("bookmarks");
 
+	$authorized_var=array('page','limit','optioncss','contextpage');
 	$url = $_SERVER["PHP_SELF"];
-
+	$url_param=array();
 	if (!empty($_SERVER["QUERY_STRING"])) {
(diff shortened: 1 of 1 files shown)


htdocs/bookmarks/bookmarks.lib.php
```diff
@@ -36,28 +36,39 @@ function printDropdownBookmarksList()
 
 	$langs->load("bookmarks");
 
+	$authorized_var=array('page','limit','optioncss','contextpage');
```
eldy, line 39: ok to add the param limit, optioncss and contextpage as param to include into the bookmark url to match a post.
However, i don't think "page" is a good idea. When changing data, the content of page is changing too. When we add the result of a post in bookmarks it means we want the result of some filters. It looks strange to me to say "i want to bookmark result when searching on abc but i want to go by default on the page 2 of the result".
Don't you think so ?

FHenry: You'r right in Dolibarr Core this param are never set in post, but maybe so external module can play with it. No Use case for now, just a guess

PR #22043, opened 2022-09-02
Update bom_card.php

dolibit-ut:


htdocs/bom/bom_card.php
```diff
@@ -335,7 +335,7 @@
 		}
 
 		$text = $langs->trans('ConfirmValidateBom', $numref);
-		/*if (!empty($conf->notification->enabled))
+		/*if (! empty($conf->notification->enabled))
```
eldy, line 338: According to PSR 12 (https://www.php-fig.org/psr/psr-12/) that we try to follow (except for tab indentation), we should not have spaces after a unitary operator.

PR #22049, opened 2022-09-02
NEW Can edit VAT in 1 step on all lines on purchase objects FPC21+

aspangaro:
Add pen to edit VAT on all lines like sells cycle

image


htdocs/fourn/commande/card.php
```diff
@@ -405,7 +405,16 @@
 	}
 
 	// Add a product line
-	if ($action == 'addline' && $usercancreate) {
+	if ($action == 'addline' && GETPOST('submitforalllines', 'aZ09') && GETPOST('vatforalllines', 'alpha')) {
```
eldy, line 408: We are missing a test on permission here.
May be    && $usercancreate

aspangaro: Corrected

aspangaro: Ok, fixed

PR #22059, opened 2022-09-03
NEW private and public note on user.php

ptibogxiv:
use same code on Dolibarr

### DIFF
--- a/htdocs/user/note.php
+++ b/htdocs/user/note.php
@@ -40,7 +40,7 @@ $object->fetch($id, '', '', 1);
 $object->getrights();
 
 // If user is not user read and no permission to read other users, we stop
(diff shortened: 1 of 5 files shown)


htdocs/user/note.php
```diff
@@ -59,37 +63,23 @@
 /*
  * Actions
  */
-
 $parameters = array('id'=>$socid);
-$reshook = $hookmanager->executeHooks('doActions', $parameters, $object, $action); // Note that $action and $object may have been modified by some hooks
+$reshook = $hookmanager->executeHooks('doActions', array(), $object, $action); // Note that $action and $object may have been modified by some hooks
```
eldy, line 67: Why removing $parameters with empty array() ?

PR #22066, opened 2022-09-03
Fix #22063 Ajax with limit of results.

### LINKED ISSUE #22063: Bug: Product not found in combobox if more then 1000 products -> or 1000 results on ajax
### Bug

Only 1000 Products are loaded in the combobox. Only the loaded products or services can filtering and selecting by the list.

### Environment Version

17.0.0-alpha

### Environment OS

Windows 10

### Environment Web server

xampp

### Environment PHP

PHP 7.4

### Environment Database

mariaDB

### Environment URL(s)

_No response_

Refresh articles and services in the combobox on each char typing.

_No response_

### Attached files

image

Humml87:
Limits the results of a ajax call.

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -2305,7 +2305,7 @@ class Form
 				}
 			}
 			// mode=1 means customers products
-			$urloption = 'htmlname='.$htmlname.'&outjson=1&price_level='.$price_level.'&type='.$filtertype.'&mode=1&status='.$status.'&status_purchase='.$status_purchase.'&finished='.$finished.'&hidepriceinlabel='.$hidepriceinlabel.'&warehousestatus='.$warehouseStatus;
+			$urloption = 'htmlname='.$htmlname.'&outjson=1&price_level='.$price_level.'&type='.$filtertype.'&mode=1&status='.$status.'&status_purchase='.$status_purchase.'&finished='.$finished.'&hidepriceinlabel='.$hidepriceinlabel.'&warehousestatus='.$warehouseStatus.($limit > 0 ? '&limit='.$limit : '');
 			//Price by customer
 			if (!empty($conf->global->PRODUIT_CUSTOMER_PRICES) && !empty($socid)) {
 				$urloption .= '&socid='.$socid;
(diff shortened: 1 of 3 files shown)


Humml87:
@eldy Please review this PR. I need this code to clean some wrong behaviors in dolibarr. Thank you!

htdocs/core/class/html.form.class.php
```diff
@@ -2305,7 +2305,7 @@ public function select_produits($selected = '', $htmlname = 'productid', $filter
 				}
 			}
 			// mode=1 means customers products
-			$urloption = 'htmlname='.$htmlname.'&outjson=1&price_level='.$price_level.'&type='.$filtertype.'&mode=1&status='.$status.'&status_purchase='.$status_purchase.'&finished='.$finished.'&hidepriceinlabel='.$hidepriceinlabel.'&warehousestatus='.$warehouseStatus;
+			$urloption = 'htmlname='.$htmlname.'&outjson=1&price_level='.$price_level.'&type='.$filtertype.'&mode=1&status='.$status.'&status_purchase='.$status_purchase.'&finished='.$finished.'&hidepriceinlabel='.$hidepriceinlabel.'&warehousestatus='.$warehouseStatus.($limit > 0 ? '&limit='.$limit : '');
```
eldy, line 2308: We are in case of PRODUIT_USE_SEARCH_TO_SELECT is on (so in case we want the limit managed by the first chars we enter and not by a setup). So we should not have parameter limit here.

htdocs/product/admin/product.php
```diff
@@ -653,12 +653,11 @@
...
-if (empty($conf->global->PRODUIT_USE_SEARCH_TO_SELECT)) {
-	print '<tr class="oddeven">';
-	print '<td>'.$langs->trans("NumberOfProductShowInSelect").'</td>';
-	print '<td class="right"><input size="3" type="text" class="flat" name="value_PRODUIT_LIMIT_SIZE" value="'.$conf->global->PRODUIT_LIMIT_SIZE.'"></td>';
-	print '</tr>';
-}
+// Limit of search results
+print '<tr class="oddeven">';
+print '<td>'.$langs->trans("NumberOfProductShowInSelect").'</td>';
+print '<td class="right"><input size="3" type="text" class="flat" name="value_PRODUIT_LIMIT_SIZE" value="'.$conf->global->PRODUIT_LIMIT_SIZE.'"></td>';
```
eldy, line 659: The goal of option PRODUIT_USE_SEARCH_TO_SELECT is to have limited by the first char we entered instead of a setup limit.
If you make the option PRODUIT_LIMIT_SIZE working in both case (when PRODUIT_USE_SEARCH_TO_SELECT is on and off), you must also check that the default value for limit when PRODUIT_USE_SEARCH_TO_SELECT is on, remains to value "0" (no limit). So we keep the expected behaviour:
"if we choosed to have a limit managed by the char entered instead of a setup, we must not have by default a list restricted".

So into conf.class.php we should change
```
if (!isset($this->global->PRODUIT_LIMIT_SIZE)) {
				$this->global->PRODUIT_LIMIT_SIZE = 1000;
}

```
into
```
if (!isset($this->global->PRODUIT_LIMIT_SIZE)) {
  if (empty($conf->global->PRODUIT_USE_SEARCH_TO_SELECT)) {
				  $this->global->PRODUIT_LIMIT_SIZE = 1000;
  } else {
				  $this->global->PRODUIT_LIMIT_SIZE = 0;
  }
}
```

So, with a default setup, when we never set any limit value, if we choose "limit by entered keys", the combo list will be complete and not truncated (we experience a lot of issue of end users asking why the list is not "complete" so it must not be truncated by default when using ajax mode).

eldy:
Closed, draft since too long delay

PR #22067, opened 2022-09-03
Bug #22064 effectif should be 51-100 and 101-500 (not 100-500)

mikygee:
Bug #22064
Effectif should not overlap.
51-100 and 100-500 is not correct
51-100 and 101-500 is correct

### DIFF
--- a/htdocs/product/fournisseurs.php
+++ b/htdocs/product/fournisseurs.php
@@ -506,6 +506,43 @@ if ($id > 0 || $ref) {
 							print '</a>';
 						}
 					}
+					print '<script type="text/javascript">
+					$(document).ready(function () {
+						$("#search_id_fourn").change(load_vat)
+						console.log("Requesting default VAT rate for the supplier...")
+					});
+					function load_vat() {
+						// get soc id
+						let socid = $("#id_fourn")[0].value
+	
+						// load available VAT rates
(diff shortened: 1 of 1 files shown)


htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -144,3 +144,5 @@ CREATE TABLE llx_bank_extrafields
 ALTER TABLE llx_bank_extrafields ADD INDEX idx_bank_extrafields (fk_object);
 
 ALTER TABLE llx_user CHANGE COLUMN note note_private text;
+
+UPDATE llx_c_effectif SET code='EF101-500', libelle='101 - 500' WHERE id=5;
```
eldy, line 148: UPDATE llx_c_effectif SET code='EF101-500', libelle='101 - 500' WHERE code='EF100-500'
seems more secured.

mikygee: I'm not very used to git, and I don't know how how to modify a pull request.
So after some attempts, I pushed another pull request taking account your comment.

mikygee: Hello,
I managed to make the changes here
https://github.com/Dolibarr/dolibarr/pull/22076/files

PR #22072, opened 2022-09-04
Use isModEnabled

aspangaro:


eldy:
Because the v16 should be released soon, i will not merge any mass update on code into it.
Thanks to push such PR into v17.
Target is:   v16 "works with php8 with warning", v17 "works without warning"

Issue #22048, opened 2022-09-02
Notes for lots/serials 


lkotsoni:
### Feature Request

Possibility to add notes in the Productlot module for each serial or batch number.

PR #22085, opened 2022-09-05
NEW #22048 Added notes (public/private) to productlot module

lkotsoni:


htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -142,3 +142,6 @@ CREATE TABLE llx_bank_extrafields
 )ENGINE=innodb;
 
 ALTER TABLE llx_bank_extrafields ADD INDEX idx_bank_extrafields (fk_object);
+
+ALTER TABLE llx_product_lot ADD COLUMN note_public text DEFAULT NULL after batch;
```
eldy, line 146: Can you also update the file   htdocs/install/mysql/table/llx_product_lot.sql
so fields will exists for new installation too...

PR #22087, opened 2022-09-05
check for errors in elephant examples

frederic34:
https://github.com/Dolibarr/dolibarr/issues/22079


ksar-ksar:
Hello @frederic34 , I have tried something in #22083

PR #22098, opened 2022-09-06
Fix : Ci error on user class

Hystepik:
Fix Ci errors for user class


aspangaro:
Already correct with PR #22090

PR #22101, opened 2022-09-06
Update conferenceorbooth_card.php

dolibit-ut:


dolibit-ut:
can we rename all files 'conferenceorbooth' into  'eventorganization' or shorter 'eventorg' 
(or 'convention')

PR #22123, opened 2022-09-07
NEW payment default values when supplier order created from reception

lvessiller-opendsi:
NEW payment default values when supplier order created from reception

**Before**
1) Create a supplier order
image

2) Create a reception from this supplier order
image

3) Create a supplier invoice from this reception
image

The payment terms and conditions are not propagated.

**In this feature**
1) Create supplier order from reception card
Elements are propagate when we create a supplier invoice from a reception :
- payment terms and conditions
- bank account id
- percent ans absolute discount
- transport mode id
All theses elements come from supplier order linked to the reception.

For other elements :
- project id
- pulic and private notes
- extra fields
Are set from reception object (source object of supplier order)

2) Create from "create bills" mass action
When we use "create bills" mass action from reception list we also propagate elements like in case 1 (see above).

### DIFF
--- a/htdocs/fourn/facture/card.php
+++ b/htdocs/fourn/facture/card.php
@@ -2050,10 +2050,9 @@ if ($action == 'create') {
 				$currency_tx = $objectsrc->multicurrency_tx;
 			}
 		}
-
-		$dateinvoice = empty($conf->global->MAIN_AUTOFILL_DATE) ?-1 : '';
+		
 		$datetmp = dol_mktime(12, 0, 0, GETPOST('remonth', 'int'), GETPOST('reday', 'int'), GETPOST('reyear', 'int'));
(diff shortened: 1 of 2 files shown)


lvessiller-opendsi:
@eldy 
Could you see and merge it ?

htdocs/fourn/facture/card.php
```diff
@@ -1973,6 +2051,7 @@
 			}
 		}
 
+		$dateinvoice = empty($conf->global->MAIN_AUTOFILL_DATE) ?-1 : '';
```
eldy, line 2054: Do we need this line ? It seems overwritten by line 2056

lvessiller-opendsi:
@eldy
What do you plan for this PR ?
Is it postponed ?

kkhelifa-opendsi:
@eldy
What do you plan for this PR ?

PR #22127, opened 2022-09-07
NEW select2 component can search on several strings.

frederic34:


frederic34:
image

BEFORE
image
AFTER
image

PR #22135, opened 2022-09-07
Update card.php

dolibit-ut:
Line 792:  $prodcustprice->fetch_all   =>  $prodcustprice->fetchAll


eldy:
PR must be atomic. Group all code change related to renaming into the same PR.

PR #22157, opened 2022-09-08
Update invoice.php

dolibit-ut:


rycks:
hello
there is only comments, spaces and update some lines order in that commit ... i'm right ?

dolibit-ut:
yes

PR #22172, opened 2022-09-09
Fix : missing column in establishement sql

Hystepik:
Fix missing label column in llx_establishement which c reate a sql error on admin_establishement


htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -149,3 +149,4 @@ UPDATE llx_c_effectif SET code='EF101-500', libelle='101 - 500' WHERE code='EF10
 
 ALTER TABLE llx_rights_def ADD COLUMN tms timestamp DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP;
 
+ALTER TABLE llx_establishment ADD COLUMN label varchar(255) NOT NULL AFTER entity;
```
eldy, line 152: I think this was just a bad naming of column.
It may be "name".
What do you think of renaming "name" into "label" (and update code to use "label" instead) ?

Hystepik: After another check no code is using the "name" column, so renaming the column is better

Hystepik: After another check no code is using the "name" column, so renaming the column is better

htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -149,7 +149,8 @@ UPDATE llx_c_effectif SET code='EF101-500', libelle='101 - 500' WHERE code='EF10
 
 ALTER TABLE llx_rights_def ADD COLUMN tms timestamp DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP;
 
-ALTER TABLE llx_establishment ADD COLUMN label varchar(255) NOT NULL AFTER entity;
+UPDATE llx_establishment SET name='' WHERE name IS NULL;
+ALTER TABLE llx_establishment CHANGE name label varchar(255) NOT NULL;
```
Hystepik, line 153: This may be a better fix

PR #22181, opened 2022-09-10
Dolibit patch require once

dolibit-ut:


htdocs/core/modules/modAccounting.class.php
```diff
@@ -25,7 +25,9 @@
  * \ingroup		Double entry accounting
  * \brief		Module to activate the double entry accounting module
  */
-include_once DOL_DOCUMENT_ROOT.'/core/modules/DolibarrModules.class.php';
+
+require_once DOL_DOCUMENT_ROOT.'/core/modules/DolibarrModules.class.php';
+
```
eldy, line 31: include_once seems better because the modXXX files are loaded dynamically by the page to edit module. if a module us corrupted because of a missing include file, the page return fatal error, with no chance to try to disable the module.

PR #22182, opened 2022-09-10
sample deprecated function

frederic34:


frederic34:
@eldy what do you think is missing ?

eldy:
> @eldy what do you think is missing ?

May be just a note into the CHANGELOG file to say  "All functions fetch_all() are deprecated for naming consitency, use fetchAll() instead".

PR #22237, opened 2022-09-12
NEW : Upgrade multiples versions in one upgrade

Hystepik:
This new function permit to upgarde dolibarr version not only 
from x.y.z to x+1.y.z but also to x.y.z to x+n.y.z

In summary, it is therefore possible to upgrade the version of dolibarr to any higher version in one go.

### DIFF
--- a/htdocs/install/upgrade2.php
+++ b/htdocs/install/upgrade2.php
@@ -245,11 +245,12 @@ if (!GETPOST('action', 'aZ09') || preg_match('/upgrade/i', GETPOST('action', 'aZ
 
 			$versiontoarray = explode('.', $versionto);
 			$versionranarray = explode('.', DOL_VERSION);
+			$versionfromarray = explode('.', $versionfrom);
 
 
 			$afterversionarray = explode('.', '2.0.0');
 			$beforeversionarray = explode('.', '2.7.9');
-			if (versioncompare($versiontoarray, $afterversionarray) >= 0 && versioncompare($versiontoarray, $beforeversionarray) <= 0) {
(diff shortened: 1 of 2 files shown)


htdocs/install/upgrade2.php
```diff
@@ -337,7 +338,7 @@
 			// Script for 2.9
 			$afterversionarray = explode('.', '2.8.9');
 			$beforeversionarray = explode('.', '2.9.9');
-			if (versioncompare($versiontoarray, $afterversionarray) >= 0 && versioncompare($versiontoarray, $beforeversionarray) <= 0) {
+			if ((versioncompare($versiontoarray, $afterversionarray) >= 0 && versioncompare($versiontoarray, $beforeversionarray) <= 0) || (versioncompare($versionfromarray, $afterversionarray) <= 0 && versioncompare($versiontoarray, $beforeversionarray) >= 0)) {
```
eldy, line 341: Each function called may use structure of data that are migrated by the upgrade.php script.
So to migrate from x to x+2 we currently do 
upgrade to x+1
upgrade2 to x+1
upgrade to x+2
upgrade2 to x+2

With you change you do directly
upgrade2 to x+2
But when do the upgrade to x+2 is done ? Before ? It means we do
upgrade to x+2
upgrade2 to x+2 
?

Hystepik: Yes we do upgrade x+1 _ upgrade x+2 _ upgrade x+n _ upgrade2 x+1 _upgrade2 x+2 _ upgrade2 x+n

PR #22239, opened 2022-09-12
add to produce status on products

homer8173:
implement ajax on off


htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -55,6 +55,9 @@ ALTER TABLE llx_user DROP COLUMN idpers3;
 
 -- v17
 
+
+ALTER TABLE llx_product ADD COLUMN toproduce tinyint(4) AFTER tobuy;
```
eldy, line 59: What is the goal / use case ?  If a product is to produce, we can just create a BOM on it.
If goal is just to say it is a manufactured product, we already have the field "nature of product".

eldy:
Missing answer since a long time, so closed.

homer8173:
Hi @eldy,  

I’d like to reopen the discussion regarding the **"To Produce"** status, as it remains a crucial feature for companies managing production.  

Currently, Dolibarr allows marking a product as **To Buy** or **To Sell**, which helps manage purchasing and sales availability. However, production management also requires a way to control whether a product can still be manufactured. The existing "nature of product" field and BOM structure are not sufficient for this purpose.  

## Why is the "To Produce" status necessary?  

### 1. Preventing Obsolete Production  
- Some products are no longer manufactured, either due to discontinued components, regulatory changes, or strategic decisions.  
- Without a **"To Produce"** status, users can still create Manufacturing Orders (MO) for obsolete products, leading to errors and unnecessary costs.  

### 2. Maintaining Historical BOMs Without Allowing New Production  
- A company may need to **keep the BOM of a product for reference** (e.g., for spare parts or documentation) while ensuring no new MO can be created.  
- The **current system does not allow locking the MO creation** without fully deleting or modifying the BOM.  

### 3. Consistency with "To Buy" and "To Sell" Logic  
- Just as products can be marked as **no longer purchasable or sellable**, manufactured products should also have a similar control mechanism.  
- This aligns with how businesses phase out products while keeping necessary data.  

## Proposed Implementation  

- A **"To Produce"** checkbox or status field in the product configuration.  
- If **unchecked**, Dolibarr should **prevent new Manufacturing Orders (MO) from being created** for that product.  
- The BOM remains accessible for reference, ensuring traceability without accidental production.  

This feature is **especially critical for large companies managing complex production workflows**, as manually checking each product before creating an MO is inefficient and prone to mistakes.  

I believe this addition would benefit all users handling production and improve Dolibarr’s manufacturing management capabilities. Could we revisit this discussion?  

Best regards,  
[@homer8173]

eldy:
Looks ok  to have a status "toproduce".
However pr with addition of sql structure should be in an isolated pr first...

eldy:

PR #22240, opened 2022-09-12
FIX :  Rich text is not diplayed

ghost:
# FIX 

Rich text is not displayed correctly on the tasks list
image
I used the code from the card to fix it 
`nl2br($object->description)` 
image

### DIFF
--- a/htdocs/projet/tasks/list.php
+++ b/htdocs/projet/tasks/list.php
@@ -988,7 +988,7 @@ while ($i < min($num, $limit)) {
 		// Description
 		if (!empty($arrayfields['t.description']['checked'])) {
 			print '<td>';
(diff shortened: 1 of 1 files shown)


htdocs/projet/tasks/list.php
```diff
@@ -988,7 +988,7 @@
 		// Description
 		if (!empty($arrayfields['t.description']['checked'])) {
 			print '<td>';
-			print dol_escape_htmltag($object->description);
+			print nl2br($object->description);
```
eldy, line 991: For content that can have several lines into a list, i recommend to use instead
dolGetFirstLineOfText($object->description, 5)
to restrict to 5 first lines only.

PR #22242, opened 2022-09-12
Fix scrutinizer 3

fboitel:


htdocs/expensereport/class/expensereport_ik.class.php
```diff
@@ -126,6 +126,19 @@ public function fetch($id, $ref = null)
 		return $result;
 	}
 
+		/**
+	 * Load object lines in memory from the database
+	 *
+	 * @return int         <0 if KO, 0 if not found, >0 if OK
+	 */
+	public function fetchLines()
+	{
```
frederic34, line 135: there is no table_element_line you don't need to create fetchLines but you can remove the call at line 124

PR #22247, opened 2022-09-13
[FIX] Get shared document when the element is shared between multiple entities

kkhelifa-opendsi:
[FIX] Get shared document when the element is shared between multiple entities

### DIFF
--- a/htdocs/document.php
+++ b/htdocs/document.php
@@ -168,7 +168,7 @@ if (!empty($hashp)) {
 			$conf->entity = $entity;
 			$conf->setValues($db);
 		}
-		$fullpath_original_file = $ecmfile->fullpath_orig;
+		$fullpath_original_file = DOL_DATA_ROOT . '/' . $ecmfile->filepath . '/' . $ecmfile->filename;
 	} else {
 		$langs->load("errors");
 		accessforbidden($langs->trans("ErrorFileNotFoundWithSharedLink"), 0, 0, 1);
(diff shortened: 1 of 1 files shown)


lvessiller-opendsi (approved):

htdocs/document.php
```diff
@@ -218,7 +219,9 @@ function llxFooter()
 $check_access = dol_check_secure_access_document($modulepart, $original_file, $entity, $user, $refname);
 $accessallowed              = $check_access['accessallowed'];
 $sqlprotectagainstexternals = $check_access['sqlprotectagainstexternals'];
-$fullpath_original_file     = $check_access['original_file']; // $fullpath_original_file is now a full path name
+if (empty($fullpath_original_file)) {
+	$fullpath_original_file     = $check_access['original_file'];
```
eldy, line 223: $check_access['original_file']  is a path on server side.
$ecmfile->fullpath_orig  is the name that the file has on local computer before uploading it. 

There is no relation between both so i see no reason to have an affect on 2 var that has different meanings.

kkhelifa-opendsi:
Can you merge please ?
I have fixed your request.

kkhelifa-opendsi:
Can you merge please ?
I have fixed your request.

htdocs/document.php
```diff
@@ -168,6 +168,7 @@ function llxFooter()
 			$conf->entity = $entity;
 			$conf->setValues($db);
 		}
+		$fullpath_original_file = DOL_DATA_ROOT . '/' . $ecmfile->filepath . '/' . $ecmfile->filename;
```
eldy, line 171: The original_path is an information on original filename.
Not used for the moment, but must not be forged to contains another value than the one provided into the form by end user.

PR #22252, opened 2022-09-13
FIX some upgrade sql errors

hregis:


hregis:
@eldy 
- user fetch error "unknown note_private field" when upgrade2 function migrate_user_photospath2() is used with v17+
- constraint create failed

PR #22255, opened 2022-09-13
FIX : amount on shipments for linked objects

ghost:
# FIX : Amount on shipments for linked objects

Recently we can link shipments on commercial documents thanks to this [PR](https://github.com/ATM-Consulting/dolibarr/pull/306). 
This fix captitalizes on [this](https://github.com/ATM-Consulting/dolibarr/pull/338) PR.

It fixes the amount before taxes for shipments on commercial documents via linked objects because shimpents don't have the field `total_ht` so it uses the parent order.

Before the fix : 
image
After the fix :
image

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -8693,7 +8693,7 @@ class Form
 					'enabled'=>isModEnabled('expedition'),
 					'perms'=>1,
 					'label'=>'LinkToExpedition',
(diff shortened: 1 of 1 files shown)


htdocs/core/class/html.form.class.php
```diff
@@ -8825,7 +8826,7 @@ public function showLinkToObjectBlock($object, $restrictlinksto = array(), $excl
 							$form = new Form($this->db);
 							print $form->textwithpicto('', $langs->trans("InformationOnLinkToContract")).' ';
 						}
-						print '<span class="amount">'.price($objp->total_ht).'</span>';
+						print '<span class="amount">'.price($objp->montantHT).'</span>';
```
eldy, line 8829: There is no field  montantHT.  Field is named  total_ht

eldy:
No fix suggested after feedback since a long time, so pr is closed

PR #22267, opened 2022-09-14
strftime is deprecated in php8.1

frederic34:


htdocs/compta/prelevement/class/bonprelevement.class.php
```diff
@@ -766,8 +766,8 @@ public function create($banque = 0, $agence = 0, $mode = 'real', $format = 'ALL'
 			$datetimeprev = $executiondate;
 		}
 
-		$month = strftime("%m", $datetimeprev);
-		$year = strftime("%Y", $datetimeprev);
+		$month = date("%m", $datetimeprev);
```
eldy, line 769: date should not be used because result depends on php timezone setup. It is not clear. Can you use instead dol_print_date(dol_now() with 3rd param 'tzuserrel' or 'gmt' .

htdocs/compta/prelevement/class/bonprelevement.class.php
```diff
@@ -760,14 +760,14 @@ public function create($banque = 0, $agence = 0, $mode = 'real', $format = 'ALL'
...
-		$datetimeprev = time();
+		$datetimeprev = dol_now('gmt');
 		//Choice the date of the execution direct debit
 		if (!empty($executiondate)) {
 			$datetimeprev = $executiondate;
 		}
 
-		$month = strftime("%m", $datetimeprev);
-		$year = strftime("%Y", $datetimeprev);
+		$month = dol_print_date("%m", $datetimeprev, 'gmt');
```
eldy, line 769: Parameters are in wrong order.

PR #22278, opened 2022-09-15
FIX Accountancy - Review reconcile function (separate manuel & auto reconcile) FPC21

aspangaro:
Based on customer return

Add possibility to reconcile operation : 
- On manuel & auto mode directly on subledger (not only on transfert)
- On the same journal (Example : One customer invoice and one credit note)
- Reconcile with a manuel entry & generated entry (coming from a module)


htdocs/accountancy/class/lettering.class.php
```diff
@@ -478,271 +503,353 @@ public function bookkeepingLettering($bookkeeping_ids, $type = 'customer_invoice
...
-		$sql .= " WHERE ab.doc_type = 'bank'";
-		//	$sql .= " AND ab.subledger_account != ''";
-		//	$sql .= " AND ab.numero_compte = '" . $this->db->escape($account_number) . "'";
-		$sql .= " AND bu.type = '" . $this->db->escape($bank_url_type) . "'";
-		if (!empty($bookkeeping_ids)) $sql .= " AND ab.rowid IN (" . $this->db->sanitize(implode(',', $bookkeeping_ids)) . ")";
-
-		dol_syslog(__METHOD__ . " - Get all payment id from bank lines", LOG_DEBUG);
+		if (!empty($bookkeeping_ids)) {
+			// Get all bookkeeping lines of piece number
+			$sql .= " LEFT JOIN (";
```
eldy, line 523: Because the select does not contains any field of the second table accounting_bookkeeping, the left join on a select distinct is not a good sql.
You should instead use the syntax
AND EXISTS (select rowid from accounting_bookkeeping as pn WHERE ... AND pn.piece_num = ab.piece_num)

aspangaro: We have updated the request to work with EXISTS.

Thanks for your return @eldy

eldy:
@aspangaro Can you help me to solve conflict on this PR ?

PR #22284, opened 2022-09-15
Fix : datetime extrafield behaviour

Hystepik:
Fix datetime input field
Before when inputting a date with blank time on end input the time was set to 00:00
Now when inputting a date with blank time on end input the time was set to 23:59


htdocs/core/class/extrafields.class.php
```diff
@@ -2088,9 +2088,12 @@ public function getOptionalsFromPost($extrafieldsobjectkey, $keyprefix = '', $ke
 					$dateparamname_end   = $keysuffix . 'options_' . $key . $keyprefix . '_end';
 					if (GETPOSTISSET($dateparamname_start . 'year') && GETPOSTISSET($dateparamname_end . 'year')) {
 						// values provided as a date pair (start date + end date), each date being broken down as year, month, day, etc.
+						$dateparamname_end_hour = GETPOST($dateparamname_end . 'hour', 'int') !='-1' ? GETPOST($dateparamname_end . 'hour', 'int') : '23';
```
eldy, line 2091: What if empty string ''  (nothing entered by end user) ?

Hystepik: If nothing is entered then the field remain empty
If a date is entered then hour and min is 23:59
If date and hour is  entered min is 59

PR #22291, opened 2022-09-15
FIX : Display the real_PMP on inventory when its value is equal to 0

atm-adrien:
## FIX : Display the real_PMP on inventory when its value is equal to 0

When we setted the real_PMP to 0 and then save it, the expected_PMP took the real_PMP place.

### DIFF
--- a/htdocs/product/inventory/inventory.php
+++ b/htdocs/product/inventory/inventory.php
@@ -1086,7 +1086,7 @@ if ($object->id > 0) {
 					print '<td class="right">';
 
 
-					if (! empty($obj->pmp_real) || $obj->pmp_real == 0) $pmp_real = $obj->pmp_real;
+					if (! empty($obj->pmp_real) || (string) $obj->pmp_real === '0') $pmp_real = $obj->pmp_real;
 					else $pmp_real = $product_static->pmp;
 					$pmp_valuation_real = $pmp_real * $qty_view;
 					print '<input type="text" class="maxwidth75 right realpmp'.$obj->fk_product.'" name="realpmp_'.$obj->rowid.'" id="id_'.$obj->rowid.'_input_pmp" value="'.price2num($pmp_real).'">';
--- a/htdocs/product/inventory/inventory.php
+++ b/htdocs/product/inventory/inventory.php
@@ -1086,7 +1086,7 @@ if ($object->id > 0) {
 					print '<td class="right">';
 
 
-					if (! empty($obj->pmp_real)) $pmp_real = $obj->pmp_real;
+					if ($obj->pmp_real >= 0) $pmp_real = $obj->pmp_real;
 					else $pmp_real = $product_static->pmp;
 					$pmp_valuation_real = $pmp_real * $qty_view;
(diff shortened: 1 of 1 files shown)


htdocs/product/inventory/inventory.php
```diff
@@ -1086,7 +1086,7 @@ function barcodeserialforproduct(tabproduct,index,element,barcodeproductqty,sele
 					print '<td class="right">';
 
 
-					if (! empty($obj->pmp_real)) $pmp_real = $obj->pmp_real;
+					if ($obj->pmp_real >= 0) $pmp_real = $obj->pmp_real;
```
eldy, line 1089: This test seems to be always true. When is it false ?

eldy (approved):

htdocs/product/inventory/inventory.php
atm-adrien: You're right, this condition seems to always be true. But, on the inventory lines, the pmp is not displayed when it is equal to 0. However, it seems important to display this information (rather than leaving the information empty)

atm-adrien: What should I do ? remove the test that seems to be always true ?

atm-adrien: @eldy

eldy: > You're right, this condition seems to always be true. But, on the inventory lines, the pmp is not displayed when it is equal to 0. However, it seems important to display this information (rather than leaving the information empty)

Yes when value is 0 in database, it means value was never initialized or defined. So we don't have to show it to keep only non null value so screen is cleaner to show only data that has a value.

atm-adrien: Here I'm talking about entering the PMP (real PMP) as 0.
When I do this, the test prevents this data from being recorded (which then shows the expected PMP)

atm-adrien: @eldy

atm-adrien: @eldy

atm-adrien: @eldy

eldy: If goal is to make a difference between PMP entered with value 0 with no PMP entered, then you can test if value is '' or '0'.

atm-adrien: Done @eldy

htdocs/product/inventory/inventory.php
```diff
@@ -1086,7 +1086,7 @@ function barcodeserialforproduct(tabproduct,index,element,barcodeproductqty,sele
 					print '<td class="right">';
 
 
-					if (! empty($obj->pmp_real)) $pmp_real = $obj->pmp_real;
+					if (! empty($obj->pmp_real) || $obj->pmp_real == 0) $pmp_real = $obj->pmp_real;
```
eldy, line 1089: If ( $obj->pmp_real) is null  (never set), $obj->pmp_real == 0 will return true when we want false.
May be you want to do
`if (! empty($obj->pmp_real) || (string) $obj->pmp_real === '0')`

Issue #22298, opened 2022-09-15
Future VAT and salary payment not seen in future bank entries


apio-sys:
### Feature Request

When looking at future bank entries (be it negative or positive) you can have a quick idea of your near future cash status. This includes invoices expected to be paid by customers or supplier invoices to be paid according to their due time entered. Also future fiscal/insurance expenses are included here. But as far as I can seen it doesn't include any pending salaries to be paid nor any outstanding VAT to be paid even though they are entered as pending payment. It would be great to include those registered expenses also in the future bank balance (provided the correct bank account is entered in such future payments of course). Maybe a minor issue but a great help to get better insight in your cash-flow.

### Use case

Enter a future payment in either salaries or upcoming VAT payments connected to the correct bank account and you'll notice that in the bank account future entries, they are not taken into consideration.

### Suggested implementation

Implement those entries in the same manner as do supplier invoices (ie. negative towards the bank account) or as with tax and social charges which are already taken into consideration in the bank forecast.

apio-sys:
Thanks for correcting this so quickly however I can't see it yet in the current version (I'm using 16.0.3) or in the changelog. How to now in which version it will be included?

Kind regards,

J.

apio-sys:
Any idea on implementation in prod? Sorry to ask double questions maybe but I find it hard to track what is fixed in which version. I understand there are loads of fixes and improvements to track on each release so not an easy job. I can help if needed in one way or another.

Let me know your thoughts,

Kind regards & happy hols,

Joris.

apio-sys:
Noticed this is now effective in 17.0.0. Many thanks, works great and is really a help in forecasting cash which I think is a help for a lot of users. It is for me at least so thanks.

PR #22325, opened 2022-09-19
NEW #22298 Bank - Add salaries & vat in tab planned entries FPC21+

aspangaro:
Example : 

image


apio-sys:
Maybe a stupid question, but when can we see this in production? Can't see this in latest stable (Im' running 16.0.3). Been through the changelogs as well and no note of it (normal since it isn't in prod yet). How can one easily now which fix is pished to prod on which version?

Kind regards,

Joris.

apio-sys:
Noticed this is now effective in 17.0.0. Many thanks, works great and is really a help in forecasting cash which I think is a help for a lot of users. It is for me at least so thanks.

PR #22331, opened 2022-09-20
Error Trad

atm-lena:
# FIX|Fix #
Error traduction "BOMTotalCostService"


frederic34:
@atm-lena did you check translation in transifex ?

atm-lena:
> @atm-lena did you check translation in transifex ?

I didn't find this translation on transifex. Maybe because translation's key is not defined in en_us file ? I updated pull request in this sense

PR #22361, opened 2022-09-22
Close #20215 : export label of extrafields

### LINKED ISSUE #20215: Fonction Export d'un attribut complémentaire
### Feature Request

When we use a complementary attributes of type select with this for example
1,abc
2,def
3,ghi

When we export the field, we found into the CSV/excel file the value 1, 2 or 3.
I suggest an enhancement: When we export an extrafield with type "select", we should also export the label into another field that we can call "nameofetrfield_label" on first line. The column for label of retrafield "select" can be added at end of all other columns.

### Use case

_No response_

### Suggested implementation

_No response_

### Suggested steps

_No response_

Hystepik:
When EXPORT_LABEL_FOR_SELECT is set
Create columns in export for each select type extrafields to store select label the name of the column is selectname_label

### DIFF
--- a/htdocs/core/modules/export/export_csv.modules.php
+++ b/htdocs/core/modules/export/export_csv.modules.php
@@ -220,12 +220,21 @@ class ExportCsv extends ModeleExports
 		} else {
 			$outputlangs->charset_output = 'ISO-8859-1';
 		}
+		$selectlabel = array();
 
 		foreach ($array_selected_sorted as $code => $value) {
 			$newvalue = $outputlangs->transnoentities($array_export_fields_label[$code]); // newvalue is now $outputlangs->charset_output encoded
 			$newvalue = $this->csvClean($newvalue, $outputlangs->charset_output);
 
 			fwrite($this->handle, $newvalue.$this->separator);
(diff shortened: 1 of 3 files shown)


htdocs/core/modules/export/export_csv.modules.php
```diff
@@ -279,14 +288,22 @@ public function write_record($array_selected_sorted, $objp, $outputlangs, $array
 			$newvalue = $this->csvClean($newvalue, $outputlangs->charset_output);
 
 			if (preg_match('/^Select:/i', $typefield) && $typefield = substr($typefield, 7)) {
-				$array = json_decode($typefield, true);
-				$array = $array['options'];
-				$newvalue = $array[$newvalue];
+				$array = jsonOrUnserialize($typefield);
```
eldy, line 291: jsonOrUnserialize has been introduct to start migration from storing data using "serialize" (not secured) into storing data using "json". Here old code is already using json, so jsonOrUnserialize seems not necessary, we shoul be able to keep json_decode (infeature we want only json_decode ans json_encode instead of serialize, unserialize).
What happen if we keep 
`$array = json_decode($typefield, true);`

Issue #22326, opened 2022-09-19
Can't modifify expense report line / Impossible de modifier une ligne de note de frais


ghost:
### Bug

Hello,
After update to Dolibarr 16.0.0, i'm not able to modifiy expense report line.
I've to create a new one and delete the bad line.
Best regards,

************************************************************

Bonjour,
Après la mise à jour de Dolibarr en version 16.0.0, je ne suis plus ne mesure de modifier une ligne de note de frais.
Je dois créer une nouvelle ligne et supprimer la mauvaise.
Cordialement,

### Environment Version

16.0.0

### Environment OS

Linux webd031.cluster014.gra.hosting.ovh.net 4.14.253-ovh-vps-grsec-zfs-classid #1 SMP Thu Oct 28 11:15:35 CEST 2021 x86_64

### Environment Web server

Apache

### Environment PHP

5.6.40

### Environment Database

MySQL or MariaDB 5.7.39-log

### Environment URL(s)

Confidential

### Expected and actual behavior

When i modifiy, price, tax or text, modifications needs to be save

### Steps to reproduce the behavior

modifiy, price, tax or text and save it.
Save action doesn't take effect.

thibdrev:
Hi,
I have the same strange behaviour on a new fresh V16.0.0.
If you try to modify an existing expense line, Dolibarr acts like if you wanted to create a new one.
Workaround : click on the "modify" icon, modify your information and validate. This creates a new expense line and you can therefore delete the original one.

hregis:
@thibdrev @CoXwEn 
fix with https://github.com/Dolibarr/dolibarr/pull/22372
thanks

thibdrev:
@hregis Thanks!

mathieu-rossignol:
Exactly the same problem ! Thank you for reporting @CoXwEn , thank you for fixing @hregis !

PR #22366, opened 2022-09-22
FIX issue #22326

atm-jpb:
# FIX issue #22326
the return of the trigger returns 0 in case of success.
we must test at least the result >= 0 (other test can be == 0 ) tell me what you think about it.


hregis:
@atm-jpb better is check only with $error
https://github.com/Dolibarr/dolibarr/pull/22372

atm-jpb:
@hregis yes i think you'r right. i made the change

hregis:
@atm-jpb see my PR https://github.com/Dolibarr/dolibarr/pull/22372

PR #22368, opened 2022-09-22
New link to object block function

SylvainLegrand:
New function to link an object to an object of the same type

### DIFF
--- a/htdocs/comm/propal/card.php
+++ b/htdocs/comm/propal/card.php
@@ -2967,7 +2967,7 @@ if ($action == 'create') {
 		print $formfile->showdocuments('propal', $objref, $filedir, $urlsource, $genallowed, $delallowed, $object->model_pdf, 1, 0, 0, 28, 0, '', 0, '', $soc->default_lang, '', $object);
 
 		// Show links to link elements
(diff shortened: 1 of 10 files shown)


htdocs/comm/propal/card.php
```diff
@@ -2967,7 +2967,7 @@
 		print $formfile->showdocuments('propal', $objref, $filedir, $urlsource, $genallowed, $delallowed, $object->model_pdf, 1, 0, 0, 28, 0, '', 0, '', $soc->default_lang, '', $object);
 
 		// Show links to link elements
-		$linktoelem = $form->showLinkToObjectBlock($object, null, array('propal'));
```
eldy, line 2970: Instead of disabling the param saying which kind of element it is at each call, i suggest instead to keep caller unchanged and add into the method showLinkToObjectBlock a test on the hidden option MAIN_ALLOW_LINK_BETWEEN_ALL_OBJ_TYPE (that will allow link to same type of element), if on, you reset the array $excludelinksto.
Linking between same type may bring troubles, so it is better to start slowly by allowing it with a hidden option.

PR #22369, opened 2022-09-23
FIX #22263 Accountancy - Account number expected in place of a rowid on parent account on export FPC21+

### LINKED ISSUE #22263: Bug: parent account output should be the parent account number and not the rowid
### Bug

Version 15.0.3

Hello,

When I export plan comptable
Compta2

The parent account data is the row id and not the parent account number
image
image

However when I'm making the export, it's clear that the expected output is the parent account number (620000000 or 62)
image

Also if that output is corrected, the import data should be adapted accordingly.

Regards

### Environment Version

15.0.3

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

aspangaro:
Before

image

After

image


aspangaro:
#22263

PR #22383, opened 2022-09-24
FIX #22382 Error on length of supplier reference

### LINKED ISSUE #22382: Error on length of supplier reference
### Bug

You can not enter a product supplier ref with more than 30 caracters
Often this refs are longer
Field in data base is a varchar 128 since v 15

### Environment Version

15.0

### Environment OS

All

### Environment Web server

All

### Environment PHP

all

### Environment Database

MariaDb

### Environment URL(s)

product/fournisseurs.php?id=37&action=add_price

It shoud be possible to enter lo,ger references
Control should not be hardcoded in code or should be based on varcharlength in database 

try to create a supplier price with a ref longer that 30 cars

### Attached files

_No response_

tiaris:
# FIX #22382 Error on length of supplier reference
Length control hardcoded for 30 caracs

### DIFF
--- a/htdocs/product/fournisseurs.php
+++ b/htdocs/product/fournisseurs.php
@@ -508,9 +508,9 @@ if ($id > 0 || $ref) {
 				print '<tr><td class="fieldrequired">'.$langs->trans("SupplierRef").'</td><td>';
 				if ($rowid) {
 					print '<input type="hidden" name="ref_fourn_old" value="'.$object->ref_supplier.'">';
-					print '<input class="flat width150" maxlength="30" name="ref_fourn" value="'.$object->ref_supplier.'">';
+					print '<input class="flat width150" name="ref_fourn" value="'.$object->ref_supplier.'">';
 				} else {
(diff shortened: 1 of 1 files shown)


htdocs/product/fournisseurs.php
```diff
@@ -508,9 +508,9 @@
 				print '<tr><td class="fieldrequired">'.$langs->trans("SupplierRef").'</td><td>';
 				if ($rowid) {
 					print '<input type="hidden" name="ref_fourn_old" value="'.$object->ref_supplier.'">';
-					print '<input class="flat width150" maxlength="30" name="ref_fourn" value="'.$object->ref_supplier.'">';
+					print '<input class="flat width150" name="ref_fourn" value="'.$object->ref_supplier.'">';
```
eldy, line 511: The length in database is 128, so why not changing maxlength="30" into maxlength="128" ?

tiaris: May be somebody needs a longer field and change in DB to varchar(256), and we will have the same problem

tiaris: Another solution will be to get the value of the varchar in DB, and put a maxlength=$value. But is it usefull ?

PR #22387, opened 2022-09-24
FIX  #22384 SKR03 account numbers, parent accounts

### LINKED ISSUE #22384: SKR03 - account numbers not 4 digits and parent accounts wrong
### Bug

In the preconfigured German accounting chart SKR03,  account numbers shorter than four digits do not have leading 0s (i.e. not following the DATEV SKR03 definition) and parent accounts are partially defined incorrectly (having set the parent account numbers instead of the row id of the parent account).

This is part of issue #22383, pull request will be provided.

priojk:
- account numbers following the DATEV standard: four digits with leading zeros
- removed incorrect parent accounts, all parent accounts set as empty


hregis:
@priojk you break "account_parent" value !

eldy:
Why did you reset value for field account_parent to 0 ?  No need for hierarchy in the chart of account ?

priojk:
Hi @eldy and @hregis, the plan was to intentionally set them all to '0' like e.g. in the Austrian (_at) version of the chart of accounts, until this can be sorted out in a better way. Currently, most parent accounts in the file were just wrong, since they reference the account number instead of the row id of the parent account. It's the same problem like in the SKR04, where for this reason most lines have been temporarily commented out.

Technically, there is no direct account hierarchy in the German charts of accounts, rather groupings. But a hierarchy will be useful in the future for the creation of a balance sheet and a profit & loss statement. BUT these are following a special hierarchy according to law. These hierarchies will need to be added as additional accounts which then act as the parents of the current SKR03 accounts.

aspangaro:
The rowids you want to take are already taken. Take a new range of 2000 accounts in the wiki and base it on that, so you can then declare the rowid of your parent accounts. 

https://wiki.dolibarr.org/index.php?title=Chart_of_accounts_id

It is necessary to update this page, we agree, it misses many ranges...

priojk:
I think there is probably not a row id conflict, although it looks like this on the first sight:

If I understand the concept right,  there is a special (parsable) comment in the sql file for each country, which gets parsed and processed by the file account.php, before the accounts are inserted into the table. 

-- Descriptif des plans comptables DE SKR03 + SKR04
-- ADD 500000 to rowid # Do no remove this comment --

This value gets added as an offset to the row ids stated in the lines of the sql file, meaning that when there is written row id 1, this is then row id 500001.

Note that there is an additional offset added for multi company support of 100000000.

So in the end, if there is given a row id of 1 for the German SKR03 chart of accounts, this will end up as row id 100500001 in the database.

priojk:
Sorry, something went wrong with the update of my pull request. 
Plan was to start with the leading zeros of the account numbers shortar than four digits (plus improving the formatting by removing unneded empty lines).
Will create a new pull request and reference this one..

Issue #16476, opened 2021-02-27
Regenerate PDF produces wrong language


bos4711:
Dolibarr 12.0.4, multilanguage support, en_GB as main language but all customers have sv_SE as default language (to get swedish on invoices etc). When manually clicking "generate" under invoice, "Swedish" is selected and PDF goes in Swedish.

However, when selecting multiple invoices (all which are set to "swedish") and clicking Regenerate PDF, all those PDF are in en_GB language. That is, the regenerate function ignores the third party language settings and uses the system wide language setting instead.

This is incorrect, I believe it should respect the third party language setting.

![2021-02-27_083629](https://user-images.githubusercontent.com/11036018/109380937-b383e780-78d7-11eb-804d-6abd98038006.png)

eldy:
Can you try with 12.0.5 ?

bos4711:
Sorry for the delay.

I just tried it in Dolibarr v12.0.5 and it is the same issue. All PDF:s get regenerated in en_GB despite they are set to "swedish" in the invoice settings.

bos4711:
Bug is confirmed in 14.0.3 as well. Third parties has default language Swedish (sv_SE), all invoice PDF have language sv_SE, yet mass-action "Re-generate PDF" produces en_US output.

I have been digging in `actions_massactions.inc.php` (in the `if (!$error && $massaction == 'generate_doc' && $permissiontoread)` tree) for an hour now but I cannot track down the error. Will try again tomorrow.

bos4711:
Now I've had more time to investigate (in the stable 14.0 branch). In `actions_massactions.inc.php`, on the if-case on line 1346, is where I found the regeneration occurs. In this if-branch, there are four new ifs:

    if ($conf->global->MAIN_MULTILANGS && empty($newlang) && GETPOST('lang_id', 'aZ09')) {
      $newlang = GETPOST('lang_id', 'aZ09');
    }
    if ($conf->global->MAIN_MULTILANGS && empty($newlang) && isset($objecttmp->thirdparty->default_lang)) {
      $newlang = $objecttmp->thirdparty->default_lang; // for proposal, order, invoice, ...
    }
    if ($conf->global->MAIN_MULTILANGS && empty($newlang) && isset($objecttmp->default_lang)) {
      $newlang = $objecttmp->default_lang; // for thirdparty
    }
    if (!empty($newlang)) {
      $outputlangs = new Translate("", $conf);
[...]

PR #22392, opened 2022-09-24
FIX: #16476 on massaction the pdf generation is not using the t…

gschratzer:
on massaction the pdf generation is not using the thirdparty language settings
see also issue # #22392

### DIFF
--- a/htdocs/core/actions_massactions.inc.php
+++ b/htdocs/core/actions_massactions.inc.php
@@ -1388,7 +1388,6 @@ if (!$error && $massaction == 'generate_doc' && $permissiontoread) {
 	foreach ($toselect as $toselectid) {
 		$result = $objecttmp->fetch($toselectid);
 		if ($result > 0) {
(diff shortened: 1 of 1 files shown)


htdocs/core/actions_massactions.inc.php
```diff
@@ -1388,6 +1388,7 @@
 	foreach ($toselect as $toselectid) {
 		$result = $objecttmp->fetch($toselectid);
 		if ($result > 0) {
+			$objecttmp->fetch_thirdparty(); //load lang from thirdparty
```
eldy, line 1391: I suggest to load the thirdparty only if we need it so inside the test line 1398.
what do you think ?

gschratzer: Unfortunately it does not enter the IF

gschratzer: @eldy I did a second version, what do you think

PR #22398, opened 2022-09-25
FIX Accountancy - Some manuals operations are displayed in subledger FPC21+

aspangaro:
When you add a manual operation in accountancy, subsidiary account can be empty, not only NULL


htdocs/accountancy/class/bookkeeping.class.php
```diff
@@ -892,6 +892,7 @@ public function fetchAllByAccount($sortorder = '', $sortfield = '', $limit = 0,
 		// Affichage par compte comptable
 		if (!empty($option)) {
 			$sql .= ' AND t.subledger_account IS NOT NULL';
+			$sql .= ' AND t.subledger_account != ""';
```
eldy, line 895: String for sql must always be '    when you can use both   ' or  " for php string
So use instead
`$sql .= " AND t.subledger_account != ''";`

PR #22399, opened 2022-09-25
FIX #21543

### LINKED ISSUE #21543: Vendor VAT rate on service/product is not respected on invoice
### Bug

For shipping ("Fraktbeställning"), we have set up that the vendor Postnord has got 0% tax on this particular product. No other vendors have a link to this product. When adding the product to a vendor invoice, the tax setting is not respected and the invoice says 25% tax on this product, while it should be 0% (since we set that up).

### Environment Version

Dolibarr 15.0.0

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

Expected value on invoice is 0% tax for this product / vendor.

_No response_

### Attached files

service_1
service_2

Ants-In:
FIX #21543 products and services VAT was not showing up automatically in vendor order*]
while making an vendor order/purchase order, the pre defined products and services VAT was not coming in the order.

### DIFF
--- a/htdocs/core/tpl/objectline_create.tpl.php
+++ b/htdocs/core/tpl/objectline_create.tpl.php
@@ -952,12 +952,14 @@ if (!empty($usemargins) && $user->rights->margins->creer) {
 			var discount = parseFloat($('option:selected', this).attr('data-discount'));
 			if (isNaN(discount)) { discount = parseFloat(jQuery('#idprodfournprice').attr('data-discount'));}
 
-			/* var tva_tx = $('option:selected', this).data('tvatx'); */
+			var tva_tx = $('option:selected', this).data('tvatx'); /* FIX21543 */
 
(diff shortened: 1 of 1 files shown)


htdocs/core/tpl/objectline_create.tpl.php
```diff
@@ -952,13 +952,15 @@ function(data) {
 			var discount = parseFloat($('option:selected', this).attr('data-discount'));
 			if (isNaN(discount)) { discount = parseFloat(jQuery('#idprodfournprice').attr('data-discount'));}
 
-			/* var tva_tx = $('option:selected', this).data('tvatx'); */
+			var tva_tx = $('option:selected', this).data('tvatx'); /* FIX21543 */
```
eldy, line 955: Do not include the comment of the fix into code. This information is already tracked by git.

Ants-In: thank you for the correction. i will be careful next time.

PR #22409, opened 2022-09-26
#FIX tempalte 03 css

Quatadah:
Template 03 Contact page is now centred, additionally all the css is under the .bodywebsite class.


hregis:
@Quatadah where is the repository of templates? Thanks

Quatadah:
@hregis htdocs/install/doctemplates/websites/*
is that the repository you are asking for ?

hregis:
@Quatadah no the source of zip

Quatadah:
@hregis Only the zip is exported to the github repository, but if you create a new website and you import the zip of the template you want, you can see its source code.

hregis:
If I need to fix or add features in the template?

Quatadah:
@hregis you can create a website on the websites module, import the zip of the template you want to edit, and then you can make the modifications, export the new zip and replace it in htdocs/install/doctemplates/websites/

hregis:
@Quatadah it's not user friendly ! you can't manage the history of changes

eldy:
> @Quatadah it's not user friendly ! you can't manage the history of changes

@hregis currently, we have no better solution for website templates because a website template is "inside" database (files are just cache). We should enhance the "Import website" process to be able to make an import from a "directory" instead of a "zip" (the directory would be the template unzipped). In such a case, we would be able to store the website template inside a subdir of htdocs/install/doctemplates/websites/ . At least we will have txt file.
and we will generate the zip (to save spaces and being able to distribute easily templates) only for the release.

PR #22412, opened 2022-09-26
#FIX Dolicloud template04

Quatadah:
Template04 css is fixed.


eldy:
Your PR contains files from template01. So this generate conflicts with other PR because the PR contains files already pushed and merged into another PR.
Can you make a different branch between 01 and 04 and push PR as 2 different branches so each PR will be independant and not in conflict with previous one ?

Quatadah:
done !

eldy:
Merged manually

PR #22421, opened 2022-09-26
Update trips.lang

dolibit-ut:
we have trip.lang & new expensereports.lang 
should we merge and make a migration to expensereports.lang (or shorter expense/expreport) or stay at trip.lang ?


eldy:
> expensereports

@dolibit-ut 
Moving all content of trips.lang file into expensereports.lang and removing trips.lang would be a good thing...

dolibit-ut:
okay

PR #22423, opened 2022-09-26
Update llx_adherent.sql

dolibit-ut:
add membersince date


htdocs/install/mysql/tables/llx_adherent.sql
```diff
@@ -63,25 +67,23 @@ create table llx_adherent
   --youtube          varchar(255),                -- deprecated
   --whatsapp         varchar(255),                -- deprecated
 
-  phone            varchar(30),
-  phone_perso      varchar(30),
-  phone_mobile     varchar(30),
   birth            date,                          -- birthday
+  membersince      date,                          -- member since date
```
eldy, line 71: We don't use such a property.

dolibit-ut: I will cheque / review.
It's about a manual field to store this info also about membership time before using Dolibarr
and maybe to say someone has a 10 year anniversary etc.

eldy:
It seems we can already get the information from the subscription database so i close the ticket.

PR #22428, opened 2022-09-27
FIXME sql error

hregis:


htdocs/accountancy/class/accountingjournal.class.php
```diff
@@ -441,24 +441,29 @@ public function getAssetData(User $user, $type = 'view', $date_start = null, $da
...
-		}
+		}*/
 		$sql .= "SELECT ad.fk_asset AS rowid, a.ref AS asset_ref, a.label AS asset_label, a.acquisition_value_ht AS asset_acquisition_value_ht";
 		$sql .= ", a.disposal_date AS asset_disposal_date, a.disposal_amount_ht AS asset_disposal_amount_ht, a.disposal_subject_to_vat AS asset_disposal_subject_to_vat";
 		$sql .= ", ad.rowid AS depreciation_id, ad.depreciation_mode, ad.ref AS depreciation_ref, ad.depreciation_date, ad.depreciation_ht, ad.accountancy_code_debit, ad.accountancy_code_credit";
 		$sql .= " FROM " . MAIN_DB_PREFIX . "asset_depreciation as ad";
 		$sql .= " LEFT JOIN " . MAIN_DB_PREFIX . "asset as a ON a.rowid = ad.fk_asset";
 		if ($in_bookkeeping == 'already' || $in_bookkeeping == 'notyet') {
-			$sql .= " LEFT JOIN in_accounting_bookkeeping as iab ON iab.fk_docdet = ad.rowid";
+			$sql .= " LEFT JOIN " . MAIN_DB_PREFIX . "accounting_bookkeeping as iab ON iab.fk_docdet = ad.rowid";
```
eldy, line 466: The old code was creating a temporary table so each fk_docdet is unique so the
LEFT JOIN will join with an empty record (if no fk_docdet found with value ad.rowid) or will return just one line.
With your change, if there is several record accounting_bookkeeping with fk_docdet = ad.rowid, that the result will return several lines.
So sql is not similar so i don't kown it if is what we want.

I think we must completely remove the ...WITH in_accounting_bookkeeping... but also the join on ...accounting_bookkeeping... and instead in the where we add
```
if ($in_bookkeeping == 'already')
$sql .= " AND EXISTS (select fk_docdet from accountin_bookkeeping as iab WHERE iab.fk_docdet = ad.rowid)";
} else if ($in_bookkeeping == 'notyet') {
$sql .= " AND NOT EXISTS (select fk_docdet from accountin_bookkeeping as iab WHERE iab.fk_docdet = ad.rowid)";
}

```

hregis: @eldy i have this "WITH" error with Mysql 5.7.34 but not with MariaDB 10.4.25 !

`You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'in_accounting_bookkeeping(fk_docdet) AS ( SELECT DISTINCT fk_docdet FROM llx_acc' at line 1`

hregis: @eldy 
i send a new fix, it's ok for you ?

PR #22430, opened 2022-09-27
Fix : php 8.1 warnings

Hystepik:
Fix php 8.1 warnings

### DIFF
--- a/htdocs/comm/propal/list.php
+++ b/htdocs/comm/propal/list.php
@@ -1900,7 +1900,11 @@ if ($resql) {
 			if (!$i) {
 				$totalarray['pos'][$totalarray['nbfield']] = 'p.total_ht';
 			}
-			$totalarray['val']['p.total_ht'] += $obj->total_ht;
+			if (empty($totalarray['val']['p.total_ht'])) {
+				$totalarray['val']['p.total_ht'] = $obj->total_ht;
+			} else {
+				$totalarray['val']['p.total_ht'] += $obj->total_ht;
+			}
 		}
 		// Amount VAT
 		if (!empty($arrayfields['p.total_tva']['checked'])) {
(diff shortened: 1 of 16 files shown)


htdocs/comm/propal/list.php
```diff
@@ -1911,7 +1915,11 @@
 			if (!$i) {
 				$totalarray['pos'][$totalarray['nbfield']] = 'p.total_tva';
 			}
-			$totalarray['val']['p.total_tva'] += $obj->total_tva;
+			if (empty($totalarray['val']['p.total_tva'])) {
```
frederic34, line 1918: you can define `$totalarray['val']['p.total_tva'] = 0` at line 1589 and also the other

htdocs/compta/tva/class/tva.class.php
```diff
@@ -345,8 +345,8 @@ public function fetch($id, $user = null)
 				$this->fk_user_creat = $obj->fk_user_creat;
 				$this->fk_user_modif = $obj->fk_user_modif;
 				$this->fk_account = $obj->fk_account;
-				$this->fk_type = $obj->fk_type;
-				$this->rappro  = $obj->rappro;
+				$this->fk_type = !empty($obj->fk_type) ? $obj->fk_type : "";
+				$this->rappro  = !empty($obj->fk_type) ? $obj->rappro : "";
```
frederic34, line 349: typo here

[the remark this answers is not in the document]
eldy: $obj->rappro ?? ""  is still a very obscur syntax.
What if value is '0' and 0 ?
For readability using old syntax looks better.

eldy: > typo here

Thanks

PR #22452, opened 2022-09-28
NEW try to use invoice instead facture

hregis:
@eldy for compatibility during renaming facture with invoice ?


eldy:
Why not. Good idea. Because class invoice has been modified since pr, i close the request but i agree to experiment this in develop.

PR #22453, opened 2022-09-28
FIX #22440 : Lang Key "UserAccountNotDefined" is missing

### LINKED ISSUE #22440: Translation key "UserAccountNotDefined" seems to be missing somewhere
### Bug

The translation key "UserAccountNotDefined" called by `/htdocs/accountancy/journal/expensereportsjournal.php` at line 641 seems to be missing :
- it is shown as "UserAccountNotDefined" on the Expense Report Journal (see attached)
- this key doesn't seem to exist [in Transifex](https://www.transifex.com/dolibarr-association/dolibarr/translate/#fr_FR/$?q=key%3AUserAccountNotDefined)

as seen in the expense report journal page
in the code

### Environment Version

16.0.0

### Environment OS

Linux

### Environment Web server

Apache

### Environment PHP

7.4.30

### Environment Database

### Environment URL(s)

/htdocs/accountancy/journal/expensereportsjournal.php

* Expected behavior : 
    * It should say "Compte utilisateur non défini" (or something similar in French)
* Actual behavior: 
    * It says "UserAccountNotDefined" (probably because the corresponding sentence is not translated yet... or if the key wasn't found)

I'm not sure on how to reproduce this 😮 as it happened when I was playing with the accountability module and creating our structure. I thought it only was a translation missing that would be corrected via transifex, and didn't memorise the actions I just performed to get there.
Hopefully it won't matter in this particular case.

Anyway, if needed, I basically did this on a fresh install:
1. Activate modules "Expense Report" and "Accountability" (modAccounting),
2. Create expense reports and validate them,
3. Link the expense reports to the journals,
4. Try to generate the expense report journal.

### Attached files

_No response_

ksar-ksar:
# Fix #22440

@aspangaro let me know if it's the good place to put it.
Trans is needed here : https://github.com/Dolibarr/dolibarr/blob/16.0/htdocs/accountancy/journal/expensereportsjournal.php#L641

### DIFF
--- a/htdocs/langs/en_US/accountancy.lang
+++ b/htdocs/langs/en_US/accountancy.lang
@@ -58,6 +58,7 @@ MainAccountForSuppliersNotDefined=Main accounting account for vendors not define
 MainAccountForUsersNotDefined=Main accounting account for users not defined in setup
 MainAccountForVatPaymentNotDefined=Main accounting account for VAT payment not defined in setup
 MainAccountForSubscriptionPaymentNotDefined=Main accounting account for subscription payment not defined in setup
+UserAccountNotDefined=Accounting account for user not defined in setup
 
 AccountancyArea=Accounting area
 AccountancyAreaDescIntro=Usage of the accountancy module is done in several step:


aspangaro:
@ksar-ksar If language key is missing in en_US, it's the good way.

Afterwards, the language files are synchronized on Transifex and go to wait for translation on the other languages and will come back down at the next sync.

ksar-ksar:
Hello @aspangaro, yes the key is missing in en_us.
It's just you seems to have a way of arranging the key in the file, not sure to be at the right place within the file

PR #22454, opened 2022-09-29
FIX avoid override initial message ($this instead $object)

hregis:


hregis:
@eldy i restart stickler-ci and it's ok

hregis:
@eldy this PR is ok

Issue #22424, opened 2022-09-26
Implement the signature feature on pdf of contract


eldy:
### Feature Request

Provide the same feature we have to make online signature of commercial proposal but for contract:
- A setup page in setup of module contract to enable the online signature
- A modification of the public/onlinesign/onlinesign.php page to suggest the signature on a contract

PR #22458, opened 2022-09-29
NEW #22424 online signature for contracts

fboitel:
NEW #22424 Online signature for contracts

It's now possible to sign online PDF contract. 

What you need to do is only activate the online sign on the contract setup page. The link will automatically appear on the contract card screen, if this one is validated.


fparrar:
Hi! On the contract setup page doesn't appear the online sign option. By setup page do you mean the 'config module page'?

PR #22465, opened 2022-09-30
NEW move htmlPrintOnlineFooter in their own lib

lvessiller-opendsi:
NEW move htmlPrintOnlineFooter in their own lib
- htmlPrintOnlinePaymentFooter() was on ticket public interface 
and now we use htmlPrintOnlineTicketFooter() and this method is in ticket lib
- a new method is introduced htmlPrintOnlineCompanyFooter() in company lib

So on ticket public interface we use htmlPrintOnlineTickteFooter() and not htmlPrintOnlinePaymentFooter() for : 
- ticket list
- ticket view
- ticket create
- ticket index

In ticket_create source code, I replaced : 
- htmlPrintOnlinePaymentFooter($mysoc, $langs, 1, $suffix, $object);
by htmlPrintOnlineTicketFooter($mysoc, $langs);
- because paramater "1, $suffix and $object" are only used for payment public interface

### DIFF
--- a/htdocs/core/lib/company.lib.php
+++ b/htdocs/core/lib/company.lib.php
@@ -2156,3 +2156,66 @@ function addMailingEventTypeSQL($actioncode, $objcon, $filterobj)
 		return $sql2;
 	}
 }
+
+/**
+ * Show footer of company in HTML pages
+ *
+ * @param   Societe		$fromcompany	Third party
+ * @param   Translate	$langs			Output language
+ * @return 	string[]	[line1, line2]
+ */
+function htmlPrintOnlineCompanyFooter($fromcompany, $langs)
+{
+	global $conf;
+
+	// Juridical status
+	$line1 = "";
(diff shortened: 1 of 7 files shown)


eldy:
So you split one mutualized method (used for all public pages) into 2 methods (one for tocket public page and one for others) ?
The method was introduced to have a common method for public pages for footer.
Why not introducing a param to manage what can be the difference between the 2.

lvessiller-opendsi:
@eldy 
If I understand, your idea is : 
- to have one common method "htmlPrintOnlineFooter()" with parameter for the difference between ticket public page and others public pages
- rename htmlPrintOnlinePaymentFooter() method with htmlPrintOnlineFooter() and move it in a common lib ?

But which lib can I use to put htmlPrintOnlineFooter() method ?

lvessiller-opendsi:
@eldy 
I replaced it with PR #22526

Issue #22469, opened 2022-10-01
Bug: Espèce should be written with a S => Espèces in the dictionnary


mikygee:
### Bug

Version: 17.0.0-alpha

Hello,

Espèce should be written with a S => Espèces in the dictionnary

Surce: Larousse
https://www.larousse.fr/dictionnaires/francais/esp%C3%A8ce/31030

![image](https://user-images.githubusercontent.com/2222550/193431576-ee979f46-5658-4410-8b84-0f61dd1a32a3.png)

### Environment Version

17.0.0-alpha

ksar-ksar:
Hello,

The change of language files must be done in transifex : https://www.transifex.com/dolibarr-association/dolibarr/dashboard/

ksar-ksar:
Can we close it ?

mikygee:
I don't know, changes on transifex need an account that should be validated.

mikygee:
Hello,
Maybe you can do this change on transifex if you have the rights to do so.

mikygee:
Not sure how to validate the translation.
![image](https://user-images.githubusercontent.com/2222550/203188955-fbe8cb95-cb01-4e60-a72c-beae5d60a25f.png)

mikygee:
Please reopen.
In 18.0.3 it's still badly written.
I tried to change it in transifex I can't.
Why don't you do it instead of closing the ticket ?
Regards

ksar-ksar:
Can you explain where in Dolibarr ? Because it has been corrected in bills.

All the translation issues are not be reported here but on transifex

mikygee:
Yes of course. It's in dictionnaries.
![image](https://github.com/Dolibarr/dolibarr/assets/2222550/e37e0ec9-83b6-4ba4-bc53-578a7ece053e)

I can't make the correction on transifex.

PR #22470, opened 2022-10-01
Correction bug #22469 Especes is only written with plurial

mikygee:
Espèce should be written with a S => Espèces in the dictionnary

Surce: Larousse
https://www.larousse.fr/dictionnaires/francais/esp%C3%A8ce/31030


eldy:
Editing language file with github is useless. Files are erased by transifex synch. Thanks to fix this in source, so https://www.transifex.com/dolibarr-association/dolibarr/dashboard/

PR #22475, opened 2022-10-03
FIX : Propal's negative quantities

atm-kevin:
Add a global constant to allow a negative quantity in propal's line


eldy:
Can you provide an example of use case ?

PR #22484, opened 2022-10-04
NEW set thirdparty type with company modify trigger

lvessiller-opendsi:
NEW set thirdparty type with company modify trigger
- add a trigger "COMPANY_MODIFY" when thirdparty type is direcly set from card instead of using "Modify" button
- uses setValueFrom() method to set the third-party type

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -2072,7 +2072,7 @@ abstract class CommonObject
 		if ($resql) {
 			if ($trigkey) {
 				// call trigger with updated object values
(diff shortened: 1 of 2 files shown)


lvessiller-opendsi:
@eldy 
I used setValueFrom() method with the trigger "COMPANY_MODIFY" but I'm not sure it's the good way to do.

eldy:
> @eldy I used setValueFrom() method with the trigger "COMPANY_MODIFY" but I'm not sure it's the good way to do.

Yes, using the generic method setValueFrom is the recommended generic method instead of having x duplicate method for each field. Trigger name is good too.

PR #22485, opened 2022-10-04
NEW|New Add Option on LDAP module to create Separate Groups in LDAP with Multicompany (dolibarr2ldap)

ibuiv:
# NEW|New Add Option on LDAP module to create Separate Groups in LDAP with Multicompany (dolibarr2ldap)
## New option
Ldap Multicompany Separate Groups Option added on LDAP Groups admin page.

## New Behaviour
With this option enabled, Dolibarr will create on the LDAP one group for each combination Group/Entity
**For example** : GroupA with two entities will create GroupA-1 and GroupA-2 in the LDAP. Users will only be members of the specific group for their entity/entities.
To see what's going on, a little button was added on the LDAP Tab on the Group Card.

## Improve Compatibility (Active Directory)
A little enhancement has been added to improve compatibility with Active Directory (sAMAccountName has been set to CN name And if no members in the group, the user Administrator is added)

### Details
ADD parameter $entity for Usergroup::_load_ldap_info : Default -1. If given, it will return the infos for groupname-$entity
ADD const LDAP_GROUP_MULTICOMPANY_SEPARATE (SelectYesNo on LDAP Admin Group Tab)
ADD 'entity' field in context array of $user for the trigger USER_MODIFY
ADD 2 fields in admin.lang (en_US) LDAPMultiCompanySeparateGroups & LDAPMultiCompanySeparateGroupsExample
ADD $info['sAMAccountName'] in return for Usergroup::_load_ldap_info()
ADD Administrator user if $info['member'] is empty for Usergroup::_load_ldap_info()


hregis:
@ibuiv maybe push this fonctionnality in Multiompany module and add hooks if needed ?

ibuiv:
@hregis 
Why not ? 
I will work on this soon. 
What is your opinion for this idea ?

eldy:
Yes, because multicompany features are  better managed by multicompany module, it is better to have it into multicompany module (if this is possible). So i close the PR.

PR #22496, opened 2022-10-06
new: on CronJob, sendEmailsRemindersOnInvoiceDueDate use billing contact email if exists

FHenry:
In invoicie list, the mass action send email use billing contact of invoice if it exists, the CronJob sendEmailsRemindersOnInvoiceDueDate should use the billing contact email


eldy:
What if there is 2 external billing contact ? 
Only first or last one seems addressed ? Same for mass action (only first or last one) ? 
May be we can forge the "to" with   "email1, email2, ..."

FHenry:
> What if there is 2 external billing contact ? Only first or last one seems addressed ? Same for mass action (only first or last one) ? May be we can forge the "to" with "email1, email2, ..."

@eldy it's already done for Send Invoice by mail from list in dolibarr/htdocs/core/actions_massactions.inc.php line 280
```

foreach ($listofobjectcontacts[$objectid] as $contactemailid => $contactemailemail) {
	$emails_to_sends[] = $objectobj->thirdparty->contact_get_property($contactemailid, 'email');
	if (!in_array($contactemailid, $contactidtosend)) {
		$contactidtosend[] = $contactemailid;
	}
}
```

Else it's reviewed for cronjob

PR #22499, opened 2022-10-06
fix: display correct id of thirdparty in Error of sendEmailsRemindersOnInvoiceDueDate

FHenry:
fix: display correct id of thirdparty in Error of sendEmailsRemindersOnInvoiceDueDate


htdocs/compta/facture/class/facture.class.php
```diff
@@ -5143,11 +5143,11 @@ public function sendEmailsRemindersOnInvoiceDueDate($nbdays = 0, $paymentmode =
 							if (!empty($recipient->email)) {
 								$to = $recipient->email;
 							} else {
-								$errormesg = "Failed to send remind to thirdparty id=".$tmpinvoice->fk_soc.". No email defined for user.";
+								$errormesg = "Failed to send remind to thirdparty id=".$tmpinvoice->thirdparty->id.". No email defined for user.";
```
eldy, line 5146: $tmpinvoice->thirdparty->id is value once we succeed in loading.
May be we should use   $tmpinvoice->socid  ?

PR #22502, opened 2022-10-06
fix #22501

### LINKED ISSUE #22501: Add NAF/APE formatting option to dol_print_profids()
### Feature Request

In `dolibarr/htdocs/core/lib/functions.lib.php` we already have the function `dol_print_profids()` that provides formatting for French options :
- SIREN (`$profIDtype == "ProfId1"`)
- SIRET (`$profIDtype == "ProfId2"`)
- TVA Intracommunautaire (`$profIDtype == "VAT"`)

https://github.com/Dolibarr/dolibarr/blob/c513a767f3e5e55ecd8a5cf39f45ea13510cca78/htdocs/core/lib/functions.lib.php#L3151-L3167

I propose to add an additional option for "NAF/APE".

As explained on [economie.gouv.fr](https://www.economie.gouv.fr/entreprises/activite-entreprise-code-ape-code-naf#), this code is composed of 4 digits and 1 letter.
And formatted as "12.34A" (with a point after the 2nd digit).

### Use case

_No response_

### Suggested implementation

1. Add an option to `dol_print_profids()` for cases where `$id == 3`, corresponding to NAF/APE.
2. Call the function with a point as separator (last argument : `$separ='.'` ). 

### Suggested steps

_No response_

thibdrev:
1. Add a formatting option for NAF/APE code in `dol_print_profids()` in `htdocs/core/lib/functions.lib.php`.
2. Specifically call this function with a point as separator for NAF/APE code in `htdocs/societe/card.php` (otherwise the default separator is a non-breaking space `&nbsp;`).


htdocs/societe/card.php
```diff
@@ -2820,7 +2820,14 @@ function init_check_no_email(input) {
 				print '<tr>';
 				print '<td>'.$idprof.'</td><td>';
 				$key = 'idprof'.$i;
-				print dol_print_profids($object->$key, 'ProfId'.$i, $object->country_code, 1);
+
+				if ($i == 3) {
```
eldy, line 2824: Other countries may not use the . as separator for their profid3.
So i think forcing to . must be done into dol_print_profids for FR and i = 3 only

thibdrev: Correct. That's what I intended first. But then I saw the `dol_print_profids()` signature allowing to chose the separator and thought "well, we'll use the `$separ` argument then, it's here for a reason". 

However I think it is misleading allowing deciding over the formatting in two different places.

Especially if this `$separ`argument is forced in some cases. 
For eg. one could call `dol_print_profids($object->$key, 'ProfId3', 'FR', 1, '-')` thinking this would replace the "." by a "-". But that wouldn't.

What is the benefit in keeping the `$separ` argument in the function definition?

eldy: I don't know/remeber why we have the param separ. I see no reasin to have it. Logic should be inside the method. Did you analyse in current if this param was used by some callers ? If not we can remove it.

thibdrev: In the develop branch, a `grep -rl "dol_print_profids("` returns 3 files having this method: 
- `htdocs/societe/card.php`
- `htdocs/main.inc.php`
- `htdocs/core/lib/functions.lib.php`

And none calls this method using the `$separ` parameter:
- `htdocs/societe/card.php` => 2 calls : 
https://github.com/Dolibarr/dolibarr/blob/787c838e43f49e105701416e18728fb5ebca87e6/htdocs/societe/card.php#L2823
https://github.com/Dolibarr/dolibarr/blob/787c838e43f49e105701416e18728fb5ebca87e6/htdocs/societe/card.php#L2938
- `htdocs/main.inc.php` => 7 calls:
https://github.com/Dolibarr/dolibarr/blob/787c838e43f49e105701416e18728fb5ebca87e6/htdocs/main.inc.php#L2175-L2193

thibdrev:
By the way I think the `$separ` parameter probably comes from a similar method named `dol_print_phone()`.
https://github.com/Dolibarr/dolibarr/blob/2fedf5f3a587a75a5180e10cda31ba71f89ce76b/htdocs/core/lib/functions.lib.php#L3260-L3274

PR #22506, opened 2022-10-06
FIX You can not set the value to empty

hregis:
@eldy for example when using a multiple choice list, the total deselection of choices is not saved


hregis:
@eldy same problem in 15, 16 and develop

eldy:
I can't reproduce on v15.
For example with extrafields  chkbxlst   with value  c_typent:libelle:id::active=1  like in example
image
I can unselect all.

Are you using chkbxlst or checkbox type ?

hregis:
@eldy yes with chkbxlst and checkbox, try with extrafield for propal lines by example

hregis:
@eldy  the problem is with extrafields chkbxlst and checkbox for object lines

eldy:
@hregis Is this trouble still present into v18 ?

hregis:
@eldy yes the problem still present in v18 with checkboxes (multichoice) only with extrafields in object lines

PR #22513, opened 2022-10-06
fix #22511

### LINKED ISSUE #22511: Missing "PredefinedMailContentThirdparty" in other.lang
### Bug

From a Third Party card, when you click on "Send email", the content of the message is pre-filled with `PredefinedMailContentThirdparty`, which seems to be a variable not translated.
image

### Environment Version

16.0.0

### Environment OS

Linux builder.o2switch.net 4.18.0-348.12.2.lve.el7h.x86_64

### Environment Web server

Apache

### Environment PHP

7.4.30

### Environment Database

### Environment URL(s)

/htdocs/societe/card.php

Expected behavior : the message is prefilled with something such as "Dear Mr, Mrs, [...] Best regards."
Actual behavior : the message is prefilled with `PredefinedMailContentThirdparty`.

1. Go to a third party card and click on "Send Email".
2. Observe that the message is prefilled with `PredefinedMailContentThirdparty`.

A quick `grep -rl "PredefinedMailContentThirdparty"`shows that this variable is only located in `htdocs/core/class/html.formmail.class.php`:
https://github.com/Dolibarr/dolibarr/blob/be10cdee9567dfa62893ac65bf5bd7d992e4ada7/htdocs/core/class/html.formmail.class.php#L1398-L1399

Since other similar variables (such as `PredefinedMailContentSendActionComm` for example) are located in the translation files `langs/AA_AA/other.lang` ; my guess is that the variable `PredefinedMailContentThirdparty` is missing in `langs/en_US/other.lang`.

### Attached files

_No response_

thibdrev:
Introduce PredefinedMailContentThirdparty in other.lang

### DIFF
--- a/htdocs/langs/en_US/other.lang
+++ b/htdocs/langs/en_US/other.lang
@@ -103,6 +103,7 @@ PredefinedMailContentSendShipping=__(Hello)__\n\nPlease find shipping __REF__ at
 PredefinedMailContentSendFichInter=__(Hello)__\n\nPlease find intervention __REF__ attached\n\n\n__(Sincerely)__\n\n__USER_SIGNATURE__
 PredefinedMailContentLink=You can click on the link below to make your payment if it is not already done.\n\n%s\n\n
 PredefinedMailContentGeneric=__(Hello)__\n\n\n__(Sincerely)__\n\n__USER_SIGNATURE__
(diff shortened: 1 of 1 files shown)


eldy:
I pushed a different fix because we want now to remove all keys PredefinedMailContent... (and keep only one PredefinedMailContentGeneric). Because now templates is managed by emails template into database so no more need for this in language files.

PR #22515, opened 2022-10-06
Fix php error in takepos

ptibogxiv:


htdocs/takepos/invoice.php
```diff
@@ -239,20 +239,20 @@ function fail($message)
 			$error++;
 			dol_syslog('Sale without lines');
 			dol_htmloutput_errors($langs->trans("NoLinesToBill", "TakePos"), null, 1);
-		} elseif (isModEnabled('stock') && $conf->global->$constantforkey != "1") {
-			$savconst = $conf->global->STOCK_CALCULATE_ON_BILL;
-			$conf->global->STOCK_CALCULATE_ON_BILL = 1;
+		} elseif (isModEnabled('stock') && getDolGlobalInt($constantforkey) != "1") {
+			$savconst =  getDolGlobalInt('STOCK_CALCULATE_ON_BILL');
+			//getDolGlobalInt('STOCK_CALCULATE_ON_BILL') = 1;
```
eldy, line 244: If you remove this it means when option for constantforkey = CASHDESK_NO_DECREASE_STOCKsession is 0, or unset, so the default value, you won't have stock decrease. And we want a stock decrease by default on sales done from TakePOS.

PR #22518, opened 2022-10-07
FIX remove db object for avoid error with postgresql

hregis:


htdocs/variants/ajax/get_attribute_values.php
```diff
@@ -77,6 +77,11 @@
 
 $res = $prodattrval->fetchAllByProductAttribute($id);
```
eldy, line 78: I think the problem is that we use fetchAllByProductAttribute($id); that returns an array of instantiated object when we just need and array of 4 properties.
Goals is to store the array in db so we should not manage objects but simple data.
Why not modifying fetchAllByProductAttribute($id);
and replace line 293
$tmp = new ProductAttributeValue($this->db);
into
$tmp = new stdClass();
We already do this into ProductCombination.class.php line 666 for example.

hregis: @eldy i send a better fix for backward compatibility

PR #22524, opened 2022-10-07
update code towards php8 compliance

grandoc:


htdocs/core/modules/action/rapport.class.php
```diff
@@ -158,7 +158,7 @@ public function write_file($socid = 0, $catid = 0, $outputlangs = '')
 
 			$pdf = pdf_getInstance($this->format);
 			$heightforinfotot = 50; // Height reserved to output the info and total part
-			$heightforfreetext = (isset($conf->global->MAIN_PDF_FREETEXT_HEIGHT) ? $conf->global->MAIN_PDF_FREETEXT_HEIGHT : 5); // Height reserved to output the free text on last page
+			$heightforfreetext = getDolGlobalInt('MAIN_PDF_FREETEXT_HEIGHT', 5); // Height reserved to output the free text on last page
```
eldy, line 161: Not really the same.
If value is set to 0, old code return 0, yours return 5

PR #22526, opened 2022-10-07
NEW move htmlPrintOnlineFooter in their own lib

lvessiller-opendsi:
NEW move htmlPrintOnlineFooter in their own lib

htmlPrintOnlinePaymentFooter() was on ticket public interface
and now we use a common method htmlPrintOnlineFooter() and this method is in company lib

So on public interface we use htmlPrintOnlineFooter() :

In public interfaces, I replaced :

- htmlPrintOnlinePaymentFooter() by htmlPrintOnlineFooter()
- I introduced a new parameter "footerPosition" to set footer position (see PR #21778) for smarter footer and nicer payment page

It replace PR #22465


lvessiller-opendsi:
@eldy
I introduced a new parameter to fix footer at bottom for public pages.
Nethertheless, in some cases we have pages with scroll bar. 
And in this case the style fixed bottom is not a good parameter because the footer is behidn the content.
For example in public ticket list : 
- when we have less than ten tickets, we not have the scroll bar and the footer is at the bottom of the page (it's good)
- when we have a scroll bar to see the full list, the footer is behind this list

How can I do to imporve it ? 
Have you any idea ?

eldy:
Troubles is still not clear to me.
It seems you will be present at Dolibarr devcamp Nantes 2022. May be we can exchange around this ticket during it ?

lvessiller-opendsi:
@eldy 
Yes, it will be easier to show you what's wrong with this interface and it will be nice to discuss with you at the "devcamp".

lvessiller-opendsi:
Replaced with #22960

Issue #22500, opened 2022-10-06
Make setup of module member easier.


eldy:
### Feature Request

On page to setup member, the 2 last section msut be modifier to we have only 1 button "Modify" per array instead of 1 per line. So 1 form per array
htdocs/adherents/admin/member.php

PR #22528, opened 2022-10-08
NEW #22500: member module set up made easier

fboitel:
NEW #22500


htdocs/adherents/admin/member.php
```diff
@@ -195,10 +225,10 @@
 
 print '<form action="'.$_SERVER["PHP_SELF"].'" method="POST">';
 print '<input type="hidden" name="token" value="'.newToken().'">';
-print '<input type="hidden" name="action" value="updateall">';
+print '<input type="hidden" name="action" value="updatemainoptions">';
 
 
-// Mains options
+/** Main options */
```
eldy, line 231: When log comment is on 1 line only, keep syntax // for php comments

PR #22536, opened 2022-10-10
#22535

### LINKED ISSUE #22535: Supplier order: wrong price format when selecting a product
### Bug

When creating a new line in a supplier order and selecting a pre-defined product with a supplier price, the format of the price is always using "." as decimal char instead of correct local setting. This is a) an optical problem and b) leads to follow-up problems in following case which is typical German setting:

* local setting decimal char = ','
* local setting thousands char = '.'
* pre-defined price has 3 digits after comma

Let's make an example with a price of 1.553. The correct German view would be 1,553 (because decimal character is ',') When selecting the product with the pre-defined price then the line is saved as 1553, because when saving the line the input is assumed to be in correct local format and the "." is interpreted as thousands char.

### Environment Version

16

### Environment OS

n/a

### Environment Web server

n/a

### Environment PHP

7.4

### Environment Database

na

### Environment URL(s)

/fourn/commande/card.php

The correct local price setting should be used when selecting a pre-defined product

* Set local settings to German
* Create a product
* Create a supplier price for that product with 3 digits after comma (e.g. 1.553)
* Create a new supplier order
* Create a new line and select the pre-defined product
* Save the line

The unit price will be 1553 instead of 1.553

### Attached files

_No response_

glu000:
# FIX|Fix #22535
Supplier order: Local setting of price format is ignored when adding a pre-defined product

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -3553,9 +3553,9 @@ class Form
 						'value'=>$outref,
 						'label'=>$outval,
 						'qty'=>$outqty,
-						'price_qty_ht'=>price2num($objp->fprice, 'MU'),	// Keep higher resolution for price for the min qty
-						'price_unit_ht'=>price2num($objp->unitprice, 'MU'),	// This is used to fill the Unit Price
-						'price_ht'=>price2num($objp->unitprice, 'MU'),		// This is used to fill the Unit Price (for compatibility)
+						'price_qty_ht'=>price($objp->fprice),		// Keep higher resolution for price for the min qty -> local format settings need to be considered
+						'price_unit_ht'=>price($objp->unitprice),	// This is used to fill the Unit Price -> therefore local format settings need to be considered
+						'price_ht'=>price($objp->unitprice),		// This is used to fill the Unit Price (for compatibility) -> local format settings need to be considered
 						'tva_tx'=>$objp->tva_tx,
 						'default_vat_code'=>$objp->default_vat_code,
 						'discount'=>$outdiscount,


htdocs/core/class/html.form.class.php
```diff
@@ -3553,9 +3553,9 @@ public function select_produits_fournisseurs_list($socid, $selected = '', $htmln
 						'value'=>$outref,
 						'label'=>$outval,
 						'qty'=>$outqty,
-						'price_qty_ht'=>price2num($objp->fprice, 'MU'),	// Keep higher resolution for price for the min qty
-						'price_unit_ht'=>price2num($objp->unitprice, 'MU'),	// This is used to fill the Unit Price
-						'price_ht'=>price2num($objp->unitprice, 'MU'),		// This is used to fill the Unit Price (for compatibility)
+						'price_qty_ht'=>price($objp->fprice),		// Keep higher resolution for price for the min qty -> local format settings need to be considered
```
eldy, line 3556: What if you use   
`price(price2num($objp->fprice, 'MU'))`
So we keep the rounding on MU = Unit Amount and we return data in local string format.
Is this will be ok for you ?

eldy:
Thanks for good description of bug and analysis and for bug fix.
I don't like to have variables that sometimes contains value for something and sometimes contains for other.
We often need also price_ht as a numeric non formatted value so using it to contains a formated value sometimes is source of troubles.
So to avoid confusion, i added another variable   price_ht_locale  and the combo selector will use it instead of price_ht.
It was more complex to fix, but makes code cleaner for future.
Can you please check my suggested fix works ?

glu000:
Thanks for improving the fix - makes absolutely sense. I tested it and it works perfectly fine

PR #22538, opened 2022-10-10
FIX missing rename field "name" to "label"

hregis:
@eldy this the same error after v13


hregis:
@eldy The Travis error is not related to my PR

PR #22543, opened 2022-10-11
FIX add short civility

atm-jpb:
# FIX : add short civility
In some languages, it is used civilityMME and Civility MR with diminutives in French for example one can have Madame et Monsieur and M. aisin que Mme.
this fix deals with this distinction by adding two keys in dict.lang that can be used by the languages ​​concerned.

### DIFF
--- a/htdocs/langs/en_US/dict.lang
+++ b/htdocs/langs/en_US/dict.lang
@@ -250,7 +250,9 @@ CountryMF=Saint Martin
 
 ##### Civilities #####
 CivilityMME=Mrs.
+CivilityMMEShort=CivilityMMEShort
 CivilityMR=Mr.
+CivilityMRShort=CivilityMRShort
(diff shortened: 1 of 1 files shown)


htdocs/langs/en_US/dict.lang
```diff
@@ -250,7 +250,9 @@ CountryMF=Saint Martin
 
 ##### Civilities #####
 CivilityMME=Mrs.
+CivilityMMEShort=CivilityMMEShort
```
eldy, line 253: This line looks strange. After the = we should have a translated text (so a human readable text).

atm-jpb: i agree  , this is only used for other languages , i'll change it with the same as Mrs.

atm-jpb:
@eldy done

PR #22548, opened 2022-10-12
# NEW|New [Adding new bank account doesn't work]

turkluka89:
# NEW|New [Adding new bank account doesn't work]
When you want to add first bank account, you can't, because app returns you are not allowed. The problem is when checking authorization on a table that doesn't have any rows count always returns 0. I edited sql select to add if there is any rows in table. I hope it helps. This is my first request.


hregis:
@turkluka89 you can't merge develop to 16 branch

turkluka89:
> @turkluka89 you can't merge develop to 16 branch

how can I commit then. I wanted to commit to develop but i couldn't.

hregis:
@turkluka89 maybe a conflict, try to update your develop branch (git pull)

turkluka89:
> 



> When you want to add first bank account, you can't, because app returns you are not allowed. The problem is when checking authorization on a table that doesn't have any rows count always returns 0. I edited sql select to add if there is any rows in table. I hope it helps. This is my first request.

image
I get this error on online editor.

PR #22549, opened 2022-10-12
Develop

turkluka89:
# NEW|New [Adding new bank account doesn't work]
When you want to add first bank account, you can't, because app returns you are not allowed. The problem is when checking authorization on a table that doesn't have any rows count always returns 0. I edited sql select to add if there is any rows in table. I hope it helps. This is my first request.


hregis:
@turkluka89 sorry but 16 to develop is not correct. Create a new branch with your develop branch

hregis:
@turkluka89 i can't reproduce your problem with 16 or develop branch !

turkluka89:
How can I show you?

hregis:
@turkluka89 have you externals modules ?

turkluka89:
I have installed dolibarr and added the module Banks & Cash. In this module
I wanted to add a new account as a super admin. Then I click add. After I
put all the data in form I click the button to create an account. Then I
get an authorization error. When I inserted a new line in a database, then
I could add a new account.

PR #22573, opened 2022-10-13
Close #22571 : order by could be set

### LINKED ISSUE #22571: sellist and Sortfield: Sortfield is not used
### Bug

For example if i have a model with that sort of definition:

```
'docpdf => array('type'=>"sellist:document_model:libelle:rowid::nom IS NOT NULL AND nom NOT LIKE 'ODT_%':type",'label'=>'PDFDocument', 'enabled'=>'1', 'position'=>120, 'notnull'=>-1, 'visible'=>1,),
```

The SQL request build was

```
SELECT rowid as rowid, libelle FROM llx_document_model WHERE nom IS NOT NULL AND nom NOT LIKE 'ODT_%' ORDER BY libelle
```

so last ":type" is not used as ORDER BY close like documentation seems to says:

```
'sellist:TableName:LabelFieldName[:KeyFieldName[:KeyFieldParent[:Filter[:Sortfield]]]]',
```

### Environment Version

15

If you confirm that bug i could make a PR to fix it

rycks:
# Fix #22571 sellist with order by option


htdocs/core/class/commonobject.class.php
```diff
@@ -6144,7 +6145,13 @@ public function showInputField($val, $key, $value, $moreparam = '', $keysuffix =
 				$sql .= $sqlwhere;
 				//print $sql;
 
-				$sql .= ' ORDER BY '.implode(', ', $fields_label);
+				//fix #22571 : order by could be set
+				//remember 'sellist:TableName:LabelFieldName[:KeyFieldName[:KeyFieldParent[:Filter[:Sortfield]]]]',
+				if(isset($InfoFieldList[5]) && $InfoFieldList[5] != "") {
```
eldy, line 6150: For extrafield that must follow same rule than commonfield, the field 5 and 6 were already reserved.
Is it possible to add the sort as field at position 7 ?

rycks: ho sorry for the delay, let me try it of course !

rycks: @eldy updated

PR #22576, opened 2022-10-14
Fix #22529 Third Party - Creation or modification - Only keep letters or numbers when pasting French professional IDs

### LINKED ISSUE #22529: Automatically remove spaces and specific characters when pasting SIREN, SIRET or NAF  in its <input> field.
### Feature Request

When creating or modifying a Third Party card, we shall automatically remove spaces and specific characters when copy-pasting a SIREN, SIRET or NAF in its `<input>` field.

### Use case

When you create or modify a Third Party from `/societe/card.php`, if you copy-paste a SIREN, SIRET or NAF which is already formatted with spaces or dots, then it will be truncated at the end since the `<input>` field is limited respectively to 9, 14 or 5 characters.

For example:
| What you copy-paste            | What you get    | What you want          |
|--------------------------------|-------------------|------------------------|
| SIREN `123 123 123`             | `123 123 1`          | `123123123`              |
| SIRET `123 123 123 12345`   | `123 123 123 12` | `12312312312345`    |
| NAF `70.22Z`                         | `70.22`                 | `7022Z`                     |

### Suggested implementation

Dolibarr might already have a JavaScript function doing that and defined somewhere.

Otherwise we've got this example https://stackoverflow.com/questions/11943924/javascript-removing-spaces-on-paste, but we need to slightly adjust the regex as currently it  solely keeps digits while we want both digits and letters.

This JS behaviour shall be activated only when the input field requires it. 
Apparently the input fields are generated here:
https://github.com/Dolibarr/dolibarr/blob/5fc1ce5491cd7312932004c542c2d5325b95e1ca/htdocs/core/class/html.formcompany.class.php#L918-L930

### Suggested steps

This is beyond my coding abilities, but shall someone give it a try, I'd be glad to see how this can be done :smile:.

thibdrev:
When creating or modifying a French Third Party, automatically remove non-alphanumeric characters when pasting SIREN, SIRET, NAF or TVA intracommunautaire in their input field.

### DIFF
--- a/htdocs/societe/card.php
+++ b/htdocs/societe/card.php
@@ -1026,6 +1026,33 @@ if (is_object($objcanvas) && $objcanvas->displayCanvasExists($action)) {
 	// -----------------------------------------
 	// When used in standard mode
 	// -----------------------------------------
+	
+	$jscodespecifictofrenchprofid = "\n".'
+		<!-- BEGIN JavaScript code specific to French third parties only. -->
+		<script type="text/javascript">
+            jQuery(document).ready(function() {
+				
+				/** 
+				 * Clean pasted text from any non-alphanumeric character.
+				 * ("ex: "123 123 123" => "123123123", "20.90Z" => "2090Z" or "FR87 520 339 938" => "FR87520339938")	
+				 */
+				function onlyAlphaNum(e) {
+					e.preventDefault();
+					var pastedText = "";
+					var pastedText = e.originalEvent.clipboardData.getData("text/plain");
+					var cleanText = pastedText.replace(/[^a-z0-9]/gi, "");
+					$(this).val(cleanText);
+				}
+
+				// Apply to the following input fields:
+				$("#idprof1").on("paste", onlyAlphaNum);	// SIREN
+				$("#idprof2").on("paste", onlyAlphaNum);	// SIRET
+				$("#idprof3").on("paste", onlyAlphaNum);	// NAF or APE
+				$("#intra_vat").on("paste", onlyAlphaNum);	// TVA intracommunautaire
+			})
+		</script>
+		<!-- END JavaScript code specific to French third parties only. -->'."\n";
+		
 	if ($action == 'create') {
 		/*
 		 *  Creation
@@ -1201,6 +1228,11 @@ if (is_object($objcanvas) && $objcanvas->displayCanvasExists($action)) {
 			$object->country_code = $tmparray['code'];
 			$object->country = $tmparray['label'];
 		}
+		
+		if ($object->country_code == 'FR') {
+			print $jscodespecifictofrenchprofid;
+		}
+		
 		$object->forme_juridique_code = GETPOST('forme_juridique_code');
 
 		// We set multicurrency_code if enabled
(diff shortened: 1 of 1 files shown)


hregis:
@thibdrev we can't put this in the third party card, imagine if we have to manage all the countries it will become undrinkable!

thibdrev:
By the way I think the `$separ` parameter probably comes from a similar method named `dol_print_phone()`.
https://github.com/Dolibarr/dolibarr/blob/2fedf5f3a587a75a5180e10cda31ba71f89ce76b/htdocs/core/lib/functions.lib.php#L3260-L3274

thibdrev:
Thanks Regis. I'm not a big fan either of how it's currently incorportated.
Any idea where this should go instead?

eldy:
Fields are "varchar" because we must stay open to a lot of use case. Sometime we need to store here a comment after the string. Or sometimes the label of fields is renamed to store something else.
Having to restrictive check generate often a lot of frustration by end users that can't enter what they would like to enter. Also a lot of data are entered by other input than screen (for example, by import or by API). That's why adding business rules on screen side is not efficient.
Reporting a warning when string has not the correct format (according to a predefined format) seems better (rules are managed on server side where business code is located, so in php side).
If you really need a cut and paste cleaning specific to France, implemented on screen side, an external module seems better .

thibdrev:
The original idea isn't to add yet another business rule, but to facilitate the user experience.

However I now see that the code I've proposed can be misleading and still does more than required: I've used a regex keeping alphanum only, instead of a method only deleting spaces or points. I thought this was cleaner (but still global enough compared to regex keeping only figures or applying exact patterns for each case).

I'll try to better explain the origin of the identified need. It came from this use case:
1. A new user tests Dolibarr, starts using it, and part of the first things he does, creates third parties.
2. He gathers the third parties professional ids and maybe copy-paste them from their websites.
3. But when copy-pasting already formatted SIREN (such as "123 123 123") in its Dolibarr, he finds out that Dolibarr automatically truncates them to "123 123 1" (with two missing digits). Therefore he needs to fill in the IDs again manually.
4. And this goes on and happens again for SIRET, NAF, TVA and the other 3rd parties.

My point is: today V16 currently presents a potential irritant for the user, linked precisely to existing business rules on client side: the input max length attribute. 
And since creating a third party is probably one of the first things a user does when testing Dolibarr, I thought "yeah, maybe we should improve this UX, in order to facilitate the adoption of Dolibarr. Because this does not provide a good feeling."

Now, I am not a dev, nor a UX expert, so I leave it to you as to what's best for Dolibarr. :+1:

thibdrev:
@eldy @hregis Here is a proposal to progress on this topic.
It is limited to { idprof1 + FR } in order to first check if the overall idea is the right one.

- I propose to release the constraints on client side by deleting (or increasing) the `maxlength=` attribute of the idprof1 `input` field. This would allow copy-pasting valid values such as `123 123 123` or `123-123-123`. The code is here:
https://github.com/Dolibarr/dolibarr/blob/5fc1ce5491cd7312932004c542c2d5325b95e1ca/htdocs/core/class/html.formcompany.class.php#L911-L929

- Good news: the value recorded in the database will still be checked correctly by `id_prof_check()` because this method cleans the data beforehand: https://github.com/Dolibarr/dolibarr/blob/5fc1ce5491cd7312932004c542c2d5325b95e1ca/htdocs/societe/class/societe.class.php#L3630-L3666

- On the other hand the `dol_print_profid()` method will have to be slightly adjusted because it currently solely checks for `dol_strlen()` and doesn't clean the data beforehand. Resulting in not formatting valid values such as `123 123123`. https://github.com/Dolibarr/dolibarr/blob/5fc1ce5491cd7312932004c542c2d5325b95e1ca/htdocs/core/lib/functions.lib.php#L3151-L3164

- Lastly, if and only if `id_prof_check()` is true, we could also decide to directly record the "cleaned" value in the database.

What do you think?

thibdrev:
Since it is better to manage the rules on server side, maybe we shall simply delete the `maxlength="n"` in the SIREN, SIRET and NAF `<input>` tags.
That would allow copy-pasting SIREN and SIRET with spaces in between numbers, and NAF codes with a dot in it.
And the existing rules on server side would still apply.

PR #22583, opened 2022-10-14
new: add information of line detail origin and source into stock mouvement

FHenry:
fix: refactor stock mvt on expedition into one method (better than duplicate code into valid and setCloture (with actuel difference...))
new: add element source and origin info into stockmvt attribute (that are not saved so no harm), but can be very useful into trigger STOCK_MOVEMENT to identify the real origin (Order line ID and Exepedition line ID, .... other) of the mvt stock


altairis-noe:
@eldy did you have time to review this one. It would be a fine enhancement. And, yes, I would need this in the project I am working on.

htdocs/expedition/class/expedition.class.php
```diff
@@ -2243,6 +2114,95 @@ public function setClosed()
...
+					$qty = $obj->edbqty;
+				}
+				if ($qty <= 0) {
+					continue;
+				}
+				dol_syslog(get_class($this) . "::valid movement index " . $i . " ed.rowid=" . $obj->rowid . " edb.rowid=" . $obj->edbrowid);
+
+				$mouvS = new MouvementStock($this->db);
+				$mouvS->origin = &$this;
+				$mouvS->setOrigin($this->element, $this->id, $obj->cdid, $obj->edid);
```
altairis-noe, line 2165: @eldy as you can see here, line $lineIdObjectSrc would be the command line id and $lineIdObjectOrigin would the expedition line id, so I would say origin_line_id for  $lineIdObjectOrigin and maybe origin_src_line_id ?

altairis-noe:
@FHenry, I hope I wasn’t mistaken in my answer to @eldy. What do you think ?

FHenry:
> @FHenry, I hope I wasn’t mistaken in my answer to @eldy. What do you think ?

That's exactly the goal. As we do not have batch or serials on orders lines, but only on shipping lines, it can be really useful to have the information about "order line source" at this step

htdocs/expedition/class/expedition.class.php
eldy: This does not match the setOrigin function parameters.
Inthis function, you have 2 new parameters: 1 to store the id of line and one to store the type of what is the id type (so a string, for example 'order_line', 'shipment_line', ...)

altairis-noe:
Hello, @eldy, what can be done to have this added to Dolibarr ?

htdocs/expedition/class/expedition.class.php
FHenry: The method setOrigin  in MouvementStock class had now this parameters (with this PR). 
That's the goal give more information into Stock mouvement SetOrigin to be more acccurate on why there is a mouvement. The new parameters are defaulted to '', it's harmless for retrocompatibility

altairis-noe:
Nothing yet, @eldy ? I wish we had an answer, because some blocking issue in the COPAC project depends on it. cc @lmag

altairis-noe:
Hello there, still very anxious to know where this one is going, @eldy

altairis-noe:
Happy new year, @eldy : did your thinking on this topic progressed

FHenry:
@eldy I came back on this old PR, sorry to not update it.

Is there something wrong about merging it ?

PR #22584, opened 2022-10-14
FIX: #22581

### LINKED ISSUE #22581: Shipping multiple products that have unique serial numbers
### Bug

When shipping product with serial number We can ship a quantity greater than 1

- [ ] Step 1 : Solve #22582 
- [ ] Step 2 : add the link between shipment lines and stock movements

_No response_

### Attached files

_No response_

FHenry:
FIX: #22581

### DIFF
--- a/htdocs/expedition/card.php
+++ b/htdocs/expedition/card.php
@@ -215,6 +215,8 @@ if (empty($reshook)) {
 		$object->size_units = GETPOST('size_units', 'int');
 		$object->weight_units = GETPOST('weight_units', 'int');
 
+		$product = new Product($db);
+
 		// We will loop on each line of the original document to complete the shipping object with various info and quantity to deliver
 		$classname = ucfirst($object->origin);
 		$objectsrc = new $classname($db);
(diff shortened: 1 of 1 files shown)


lmag:
Missing number lot/serial in message ?

eldy:
@FHenry Can you help me to merge this into develop ?

lmag:
@altairis-noe

FHenry:
@eldy  Why did you change merge branch to develop, it's a bug into from 14.0 not a feature ?

eldy:
> @eldy  Why did you change merge branch to develop, it's a bug into from 14.0 not a feature ?

Nearlly.
It is a control that is not implemented. So it can be viewed as a bug but it is also something not yet coded. Also it will introduce a restriction making some action no more possible when it was before. Yes this is in hope to fix a bug but it also change the behaviour of what app accept. This is not a blocking bug and it has also a high risk of introducing regression (because it introduce new restrictions that didn't exists before). So such changes can be qualified as bug or as feature not completely implemented. Don't forget, in a stable, we try to fix "blocking or very annoying" bugs in priority and avoid others, other enhancements must be in develop. Before it has a high risk of regression. The fact that there is a lot of conflict with current code and merging looks difficult is a good proof of this level of risk, make me think it is better to qualify this a feature not yet completely implemented and trying to merge it in develop so we can benefit of the long beta phase if this has side bad effect.

lmag:
In this case are you ok to validate this feature in development? If we adapt it?
@altairis-noe

altairis-arre:
it's seems to be a real problem to be able to send several times a unique SN no ?
it's a bug, not a new feature.

FHenry:
@eldy There a PR on develop branch

eldy:
Merged in develop.

PR #22589, opened 2022-10-14
FIX: Fix the function get_origin() on MouvementStock class for general case

kkhelifa-opendsi:
FIX: Fix the function get_origin() on MouvementStock class for general case


htdocs/product/stock/class/mouvementstock.class.php
```diff
@@ -976,7 +976,7 @@ public function get_origin($origin_id, $origin_type)
 				if ($origin_type) {
 					// Separate originetype with "@" : left part is class name, right part is module name
 					$origin_type_array = explode('@', $origin_type);
-					$classname = ucfirst($origin_type_array[0]);
+					$classname = strtolower($origin_type_array[0]);
```
eldy, line 979: Are you sure ? Class names usually start with uppercase.

Issue #22594, opened 2022-10-14
On page to setup company, add a new line for VAT ID (see screenshot)


eldy:
### Feature Request

On page to setup the module "Third Party", add a new line for VAT ID on column  "Must be unique" (and only this one). No choice on the 2 others.
Implement the check like done, at same place of code, than for other options
![image](https://user-images.githubusercontent.com/883887/195951871-5ad808e2-8ad4-45b1-8e10-ef5058549b0a.png)

PR #22599, opened 2022-10-15
NEW #22594: can chose if VAT ID is unique or not for third parties

fboitel:
NEW #22594


eldy:
You must also implement the check that value "vat_intra" is unique when creating or updating a thirdparty (when option $conf->global->SOCIETE_VAT_INTRA_UNIQUE is on.
You can find code to change by searching on "SOCIETE_EMAIL_UNIQUE" that is optiion to check if email is unique to see where there is some checks to implement.

PR #22607, opened 2022-10-17
NEW: Change filter type on tickets list to multiselect

kkhelifa-opendsi:
NEW: Change filter type on tickets list to multiselect


htdocs/core/class/html.formticket.class.php
```diff
@@ -730,9 +732,9 @@ public function selectTypesTickets($selected = '', $htmlname = 'tickettype', $fi
 				}
 
 				// If text is selected, we compare with code, otherwise with id
-				if (preg_match('/[a-z]/i', $selected) && $selected == $arraytypes['code']) {
+				if (in_array($arraytypes['code'], $selected)) {
```
eldy, line 735: $arraytypes['code'] is a string

PR #22617, opened 2022-10-18
FIX : add loadRoleMode on getlinearray

atm-jpb:
# FIX :  add loadRoleMode on getlinearray

in the case of a project deletion, the associated tasks are never deleted in the database


atm-jpb:
@eldy

PR #22635, opened 2022-10-20
FIX 16.0 select_bom filter on defined products

atm-steve:
# FIX 16.0 select_bom function in html.form.class.php
Add parameter to define wich is the source product. Enable to display only the BOM(s) used by a defined product when user search for a BOM in list


atm-steve:
Missing too much fix

PR #22659, opened 2022-10-23
Use default e-mail address for the from field of expense reports related e-mails

ymollard:
# FIX Use default e-mail address for the from field of expense reports related e-mails

Recent increase of SPF and DKIM requirements in GAFAM mail providers cause Dolibarr e-mails to be classified as spam or even silently rejected. The use of a SPF-DKIM-authorized address should fix this. This fix changes the sender of the expense reports related e-mails for the default address instead of the address of the user address

# NEW Add help picto to remind users to respect SPF/DKIM in FROM e-mails

### DIFF
--- a/htdocs/admin/mailing.php
+++ b/htdocs/admin/mailing.php
@@ -136,7 +136,8 @@ print '<td>'.$langs->trans("Example").'</td>';
 print "</tr>\n";
 
 print '<tr class="oddeven"><td>';
-print $langs->trans("MailingEMailFrom").'</td><td>';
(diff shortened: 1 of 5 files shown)


ymollard:
@eldy: should we also push this to other supported major versions? (What are they: 15.x,14.x?)

I'm not much literate about how spam filters work, but if they learn from production e-mails, they may learn that the Dolibarr templates and/or users' web servers are spam content because of frequent SPF failures, so I would push that PR in any production branch that could prevent this learning to happen (this is maybe too late already).

PR #22668, opened 2022-10-23
FIX compatibility with multidimensional array input name

hregis:
@eldy please, it's very important for branch 14, 15 , 16 and develop

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -4903,8 +4903,8 @@ class Form
                          			if ($("input[name=\'" + inputname + "\']").attr("type") == "radio") {
 										inputvalue = $("input[name=\'" + inputname + "\']:checked").val();
 									} else {
-                         		    	if ($("#" + inputname).attr("type") == "checkbox") { more = ":checked"; }
-                         				inputvalue = $("#" + inputname + more).val();
+                         		    	if ($("input[name=\'" + inputname + "\']").attr("type") == "checkbox") { more = ":checked"; }
+                         				inputvalue = $("input[name=\'" + inputname + "\']" + more).val();
 									}
                          			if (typeof inputvalue == "undefined") { inputvalue=""; }
 									console.log("check inputname="+inputname+" inputvalue="+inputvalue);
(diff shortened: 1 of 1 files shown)


eldy:
Can you describe the bug and how to reproduce it ?

hregis:
@eldy if you would use multiple field to send an array
for example :
`<input name="invoice[$invoice->id][lines][$line->id]" value="">`

htdocs/core/class/html.form.class.php
```diff
@@ -4903,8 +4903,8 @@ public function formconfirm($page, $title, $question, $action, $formquestion = '
                          			if ($("input[name=\'" + inputname + "\']").attr("type") == "radio") {
 										inputvalue = $("input[name=\'" + inputname + "\']:checked").val();
 									} else {
-                         		    	if ($("#" + inputname).attr("type") == "checkbox") { more = ":checked"; }
-                         				inputvalue = $("#" + inputname + more).val();
+                         		    	if ($("input[name=\'" + inputname + "\']").attr("type") == "checkbox") { more = ":checked"; }
```
altairis-tof, line 4906: @hregis Be careful ! select boxes may also be used in formconfirm and this will not work !
You should write $("[name=xxx]") without the input tag.

altairis-tof (changes requested):

htdocs/core/class/html.form.class.php
hregis: @altairis-tof 
and if your remove "input" in `"input[name=\'" + inputname + "\']" `?

```
if ($("[name=\'" + inputname + "\']").attr("type") == "checkbox") { more = ":checked"; }
inputvalue = $("[name=\'" + inputname + "\']" + more).val();
```

altairis-tof: yes it works; dont forget to do the same on each line you changed ;)

altairis-tof: i see you did it ! perfect !

hregis:
@eldy i send a new fix to work with all situation

eldy:
Do you mean you have 2 input fields ?
If yes, each input fields must have its own id. For example <input name="invoice_'.$invoice->id.'_lines_'.$line->id" value="">
Still not succeed in reproducing the bug. On which url do you have it ?

hregis:
@eldy 
the "formconfirm()" function only takes into account the "name" of the element and not its "id", so my PR checks on the "name" instead of the "id"! The result will be the same since the "id" potentially has the same value as the "name"! no?

hregis:
@eldy 
or then in a second step, it is necessary to review the "formconfirm" function in order to be able to define an "id" different from the "name" for the proposed response fields? But, for the moment, this solution solves the problem.

eldy:
Using name instead of id introduce other troubles. 
PR may introduce more pb that fixes,, so i prefer to close pr.
Also i also failed to reproduce the bug.

PR #22670, opened 2022-10-24
FIX can not set prospect status "Do not contact"

hregis:


htdocs/societe/class/societe.class.php
```diff
@@ -1479,7 +1479,7 @@ public function update($id, $user = '', $call_trigger = 1, $allowmodcodeclient =
 
 			$sql .= ",fk_effectif = ".($this->effectif_id > 0 ? ((int) $this->effectif_id) : "null");
 			if (isset($this->stcomm_id)) {
-				$sql .= ",fk_stcomm=".($this->stcomm_id > 0 ? ((int) $this->stcomm_id) : "0");
+				$sql .= ",fk_stcomm=".($this->stcomm_id >= -1 ? ((int) $this->stcomm_id) : "0");
```
eldy, line 1482: This test is strange, it seems it will ba always true.

hregis: @eldy it is especially that it would not have been necessary to put the value -1 for status "Do not contact", it would have been better to put 9 !

hregis: @eldy i send a better fix

htdocs/societe/class/societe.class.php
```diff
@@ -1479,7 +1479,7 @@ public function update($id, $user = '', $call_trigger = 1, $allowmodcodeclient =
 
 			$sql .= ",fk_effectif = ".($this->effectif_id > 0 ? ((int) $this->effectif_id) : "null");
 			if (isset($this->stcomm_id)) {
-				$sql .= ",fk_stcomm=".($this->stcomm_id > 0 ? ((int) $this->stcomm_id) : "0");
+				$sql .= ",fk_stcomm=".(($this->stcomm_id > 0 || $this->stcomm_id = -1) ? ((int) $this->stcomm_id) : "0");
```
eldy, line 1482: You mean == and not = ?
May be you replace with just    ...fk_stcomm=".((int) $this->stcomm_id);    ?

hregis: @eldy yes you are right, by dint of fixing bugs we get confused!
it's fixed

PR #22672, opened 2022-10-24
ADD  - Option Hide Column On PDF

BB2A-Anthony:


htdocs/admin/pdf.php
```diff
@@ -454,6 +454,13 @@
 	print $form->selectyesno('MAIN_GENERATE_DOCUMENTS_WITHOUT_VAT', (!empty($conf->global->MAIN_GENERATE_DOCUMENTS_WITHOUT_VAT)) ? $conf->global->MAIN_GENERATE_DOCUMENTS_WITHOUT_VAT : 0, 1);
 }
 print '</td></tr>';
+print '<tr class="oddeven"><td>'.$langs->trans("HideColumnVATInformationOnPDF").'</td><td';
```
eldy, line 457: We already have option MAIN_GENERATE_DOCUMENTS_WITHOUT_VAT
What is difference. Only  MAIN_GENERATE_DOCUMENTS_WITHOUT_VAT seems usefull.
Do you have any use case that need MAIN_GENERATE_DOCUMENTS_WITHOUT_VAT_COLUMN and not MAIN_GENERATE_DOCUMENTS_WITHOUT_VAT ?

BB2A-Anthony:
> We already have option MAIN_GENERATE_DOCUMENTS_WITHOUT_VAT
What is difference. Only  MAIN_GENERATE_DOCUMENTS_WITHOUT_VAT seems usefull.
Do you have any use case that need MAIN_GENERATE_DOCUMENTS_WITHOUT_VAT_COLUMN and not MAIN_GENERATE_DOCUMENTS_WITHOUT_VAT ? 

If I want to hide the column and display the total VAT in the resume

PR #22679, opened 2022-10-26
Fix #22669 : Prospect status can be set to 'Do not contact'

### LINKED ISSUE #22669: Prospect Status Cannot be Set to 'Do not contact'
### Bug

In the third parties list it's not possible to set the prospect status to 'Do not contact' (at least since version 6). Instead it switches to 'Never contacted'.  
  
I've found the issue to be inside '\societe\class\societe.class.php', on line: 

`$sql .= ",fk_stcomm=".($this->stcomm_id > 0 ? $this->stcomm_id : "0");`

We must check for a prospect status ID ranging from '-1' to '3' to account for all possible status IDs, not only for a positive number. Changing the comparison to '>= -1' solved the problem for me. Otherwise, if the 'Do not contact' link ist clicked, always '0' (= 'Never contacted') will be returned.

I've reported this years ago in the English forum, but the post has been deleted shortly after. I guess it was the wrong place, so I'm giving it another try here.

### Environment Version

16.0.1 (and older)

### Environment OS

Linux

### Environment Web server

_No response_

### Environment PHP

8 (and older)

### Environment Database

MySQL 5.7

### Environment URL(s)

/societe/list.php

_No response_

Click on 'Do not contact'.

### Attached files

_No response_

Hystepik:
Fix societe.class.php code to add the possibility to set to 'Do not contact' prospect status

### DIFF
--- a/htdocs/societe/class/societe.class.php
+++ b/htdocs/societe/class/societe.class.php
@@ -1452,7 +1452,7 @@ class Societe extends CommonObject
 
 			$sql .= ",fk_effectif = ".($this->effectif_id > 0 ? ((int) $this->effectif_id) : "null");
 			if (isset($this->stcomm_id)) {
-				$sql .= ",fk_stcomm=".($this->stcomm_id > 0 ? ((int) $this->stcomm_id) : "0");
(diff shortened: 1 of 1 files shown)


hregis:
Hello @Hystepik it's already fixed https://github.com/Dolibarr/dolibarr/pull/22670

PR #22687, opened 2022-10-27
15 fix propal right

FHenry:
module name is "propal" and test 

line 724 
if (!in_array($module, $conf->modules)) {
		return 0;
}

is on "propale"


FHenry:
@eldy 
Could merge 15.0 into 16.0 ? This bug also exists into 16.0

PR #22688, opened 2022-10-27
FIX: check if contact is active before creating associated ticket

altairis-noe:
If we use the public interface to create tickets, we check if the reporter’s email is known from Dolibarr and associate the ticket to the first contact whose email checks out. But we forget to ensure that said contact is currently active in Dolibarr. 
This corrects that.


eldy:
I made some minor changes so if all contacts are disabled, we will get the error (with your version, it seems that we take the first active but if no active exists, we remain on the first non active).
Can you check my modified version works ?

altairis-noe:
OK, @eldy, I tested and corrected it. If the first address checked out, it would not be considered 😉

PR #22689, opened 2022-10-27
NEW Make option PDF_INCLUDE_ALIAS_IN_THIRDPARTY_NAME visible

BB2A-Anthony:
# NEW|New [*Compléte function pdfBuildThirdpartyName to use company name of contact*]

@eldy please check my others PR

### DIFF
--- a/htdocs/admin/pdf.php
+++ b/htdocs/admin/pdf.php
@@ -174,9 +174,6 @@ if ($action == 'update') {
 	if (GETPOSTISSET('PDF_INCLUDE_ALIAS_IN_THIRDPARTY_NAME')) {
 		dolibarr_set_const($db, "PDF_INCLUDE_ALIAS_IN_THIRDPARTY_NAME", GETPOST('PDF_INCLUDE_ALIAS_IN_THIRDPARTY_NAME', 'alpha'), 'chaine', 0, '', $conf->entity);
 	}
-	if (GETPOSTISSET('MAIN_USE_COMPANY_NAME_OF_CONTACT')) {
-		dolibarr_set_const($db, "MAIN_USE_COMPANY_NAME_OF_CONTACT", GETPOST('MAIN_USE_COMPANY_NAME_OF_CONTACT', 'alpha'), 'chaine', 0, '', $conf->entity);
-	}
 	if (GETPOSTISSET('PDF_USE_A')) {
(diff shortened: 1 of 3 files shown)


htdocs/admin/pdf.php
```diff
@@ -399,6 +401,21 @@
...
 print '</td></tr>';
 
+// Show alias in thirdparty name
+print '<tr class="oddeven"><td>'.$langs->trans("PDF_INCLUDE_ALIAS_IN_THIRDPARTY_NAME").'</td><td>';
+if ($conf->use_javascript_ajax) {
+	$arrval = array('0' => $langs->trans("No"), '1' => $langs->trans("THIRDPARTY_ALIAS"), '2' => $langs->trans("ALIAS_THIRDPARTY"));
+	print $form->selectarray("PDF_INCLUDE_ALIAS_IN_THIRDPARTY_NAME", $arrval, getDolGlobalInt('PDF_INCLUDE_ALIAS_IN_THIRDPARTY_NAME'));
+}
+
+print '<tr class="oddeven"><td>'.$langs->trans("MAIN_USE_COMPANY_NAME_OF_CONTACT").'</td><td>';
```
eldy, line 411: The default behaviour is already tha same as MAIN_USE_COMPANY_NAME_OF_CONTACT = 1
So to avoid non usefull options inflation, we don't need to have this option available into setup, the default behaviour is already to have it enabled.

bb2a: ok it's done

htdocs/core/lib/pdf.lib.php
```diff
@@ -400,9 +400,16 @@ function pdfBuildThirdpartyName($thirdparty, Translate $outputlangs, $includeali
 			}
 		}
 	} elseif ($thirdparty instanceof Contact) {
-		$socname = $thirdparty->socname;
-		if (($includealias || getDolGlobalInt('PDF_INCLUDE_ALIAS_IN_THIRDPARTY_NAME')) && !empty($thirdparty->name_alias)) {
-			// TODO PDF_INCLUDE_ALIAS_IN_THIRDPARTY_NAME not completely implemented
+		if ($thirdparty->socid > 0) {
+			$thirdparty->fetch_thirdparty();
```
eldy, line 404: You need this fetch to get thirdparty->thirdparty->name_alias.
But we need it only it 
`if (($includealias || getDolGlobalInt('PDF_INCLUDE_ALIAS_IN_THIRDPARTY_NAME'))`
So to avoid useless fetch, can you move it after the if ?

PR #22704, opened 2022-10-28
New constant warhouse internal not sell

leninrivas:
# NEW | New Constant ENTREPOT_WAREHOUSEINTERNAL_NOT_SELL
To no show stock of products in internal warehouse


csteh:
Production (MRP) is not interpreted as internal mouvement

Issue #22626, opened 2022-10-19
Add sorting filter on date creation and date update on lists


eldy:
### Feature Request

We can filter on most field but not date creation and date modification for list of thirdparties, contracts and projects.
Suggest a filter with 2 dates (from / until) for each date.
See filter on date start/end on project list for example.

PR #22707, opened 2022-10-29
NEW #22626 date filter thirdparties contracts projects

fboitel:
NEW #22626


eldy:
Can you resubmit after solving merging conflicts ?

Issue #22676, opened 2022-10-25
Add a mass action on product "Increase/Decrease price"


eldy:
### Feature Request

Add a mass action on product list to update the price of a product

### Suggested implementation

You can take example on mass action "set tag" to see how to ask a parameter after selectiong the mass action.
So you ask, after selecting "increase/decrease customer price", the new rate (prefilled with +0%).
So user enter the percent (exemple 10) and all products have their price modified. 30 become 33, 100 become 110...

PR #22710, opened 2022-10-30
NEW #22676 : massaction for updating product prices

fboitel:
NEW #22676


atm-maxime:
Nice feature ! Thanks for contributing.
Maybe a round() of the calculation is necessary to avoid many decimal numbers. Maybe as an option ? I often have customers who want to round to the unit. E.g. 125 € + 3% = 129 €

fboitel:
Thank you for the feedback ! Maybe I should instead put an option before confirming the mass action to choose how we want to round it? Because rounding it up to the unit can be a problem for low prices

eldy:
Prices are already rounded using the setup of user to have a unit price. But i think maxime request is to have a choice after the rate to decide to round differently just for this update. You may add this in a second step: Ask the nb of decimal for rounding and use it before calling updatePrice. This is not a pb for small amount because user can choose the product. User is still able to maje the update in 2 steps if there is 2 different rounding rule to apply.

goldenwt:
Hello,

Please where is this implemented? i have updated the files with the solution files in my instance, but cant find where the bulk update pricing is. Is there a place it is activated?

Pls can someone guide?

ReemaZa:
Hello, 
When I perform a mass action to update the product price, the price base type of the product is removed. I believe this is due to $obj being used instead of $object when passing the price_base_type to the updatePrice function. 
By changing $obj->price_base_type to $object->price_base_type in both the if statement and the arguments of the updatePrice function it works correctly.

`if (!$error && ($action == 'updateprice' && $confirm == 'yes') && $permissiontoadd) {
	$db->begin();
	if (GETPOSTISSET('pricevariation')) {
		$pricepercentage=GETPOST('pricevariation', 'int');
		if ($pricepercentage == 0) {
			setEventMessages($langs->trans("RecordsModified", 0), null);
		} else {
				//var_dump($contcats);exit;
				if ($result > 0) {
					if (**$obj->price_base_type** == 'TTC') {
						$newprice = $object->price_ttc * (100 + $pricepercentage) / 100;
						$minprice = $object->price_min_ttc;
					} else {
						$newprice = $object->price * (100 + $pricepercentage) / 100;
						$minprice = $object->price_min;
					}
					$res = $object->updatePrice($newprice, **$obj->price_base_type**, $user, $object->tva_tx, $minprice, 0, $object->tva_npr, 0, 0, array(), $object->default_vat_code);
					if ($res > 0) {
						$nbok++;
					} else {
					}
				} else {
					$error++;
					break;
				}
			}
		}
	}

	if (!$error) {
		if ($nbok > 0) {
			setEventMessages($langs->trans("RecordsModified", $nbok), null);
		}
		$db->commit();
		$toselect=array();
	} else {
		$db->rollback();
	}
}`

Another thing is that when the price of a product with variants is updated using the mass action, the prices of the product's variants are not updated. However, when we update the price of a product with variants in the usual way on the product price page (i mean without using mass action), the prices of its variants are updated. This is because, in the price.php file, after the updatePrice function is called on the product, the update function is also called, which handles updating the product's variants.

PR #22719, opened 2022-10-31
really fetch all product data and prices info

frederic34:
Product list do not display dynamic price


htdocs/product/list.php
```diff
@@ -1317,43 +1318,43 @@
...
-		$product_static->weight = $obj->weight;
-		$product_static->weight_units = $obj->weight_units;
-		$product_static->volume = $obj->volume;
-		$product_static->volume_units = $obj->volume_units;
-		$product_static->surface = $obj->surface;
-		$product_static->surface_units = $obj->surface_units;
-		if (!empty($conf->global->PRODUCT_USE_UNITS)) {
-			$product_static->fk_unit = $obj->fk_unit;
-		}
+		$product_static->fetch($obj->rowid);
```
eldy, line 1321: Using a fetch guarantee that the fetch is complete, however, if we have a list of 500 lines, it means we will make 501 request to show the page instead of 1.
Such change makes code cleaner and easier but make application no more usable on large databases.

frederic34: @eldy what do you think of adding hook there ?

PR #22728, opened 2022-11-02
FIX Not download share link

leninrivas:


htdocs/core/lib/files.lib.php
```diff
@@ -2913,14 +2913,15 @@ function dol_check_secure_access_document($modulepart, $original_file, $entity,
 		if ($fuser->rights->expedition->{$lire} || preg_match('/^specimen/i', $original_file)) {
 			$accessallowed = 1;
 		}
-		$original_file = $conf->expedition->dir_output."/".(strpos('sending/', $original_file) === 0 ? '' : 'sending/').$original_file;
-		//$original_file = $conf->expedition->dir_output."/".$original_file;
+		//$original_file = $conf->expedition->dir_output."/".(strpos('sending/', $original_file) === 0 ? '' : 'sending/').$original_file;
+		$original_file = $conf->expedition->dir_output."/".$original_file;
```
eldy, line 2917: I don't understand the need to change this.
What is the value of $original_file you have that fails ?

leninrivas: Hi. It's duplicated sending.
sending/sending/ un Expedition and sending/receipt in delivery.

leninrivas: Its fix download from document.php?hashp= ... but crash download PDF from module. :(
This is the $fullpath_original_file: expedition/**sending/**receipt/T001-00000011/T001-00000011.pdf
That word is the problem.
In expedition is:
This is the $fullpath_original_file: expedition/**sending/**sending/EXP01-00000011/EXP01-00000011.pdf

leninrivas:
Fixed in #24345. Closed

PR #22730, opened 2022-11-02
add same translation in many entities

frederic34:


hregis:
@frederic34 @eldy 
 I think for translations it should work the same as for extrafields. A choice yes or no on sharing this translation between entity during creation and modification, so the value of the "entity" field would be 0 for all entities or the id of the entity that creates the translation. What do you think ?

frederic34:
If i want Devis instead Proposal on entities 1 and 3 but not on 2 it's not possible due to index unique

eldy:
> @frederic34 @eldy I think for translations it should work the same as for extrafields. A choice yes or no on sharing this translation between entity during creation and modification, so the value of the "entity" field would be 0 for all entities or the id of the entity that creates the translation. What do you think ?

I confirm.  Value 0 should mean "all entities". x mean entity x only.

Issue #22740, opened 2022-11-03
Feature Request: OpenID Connect authentication


jeritiana:
There is already an "OpenID" support in Dolibarr, but it is actually OpenID version 2.0 (with the use of Yadis, XRDS document and such) and it is not well documented.

The "OpenID" protocol in today's world rather means OpenID Connect (third generation of OpenID), supported by Auth0, Keycloak, Google etc. Read here: https://openid.net/connect/.

This is a very simple implementation of the OpenID Connect Authorization flow (w/o PKCE) for Dolibarr. Read here: https://developer.okta.com/docs/guides/implement-grant-type/authcode/main/#grant-type-flow

This implementation focuses on providing the authentication with **zero** modification on the Dolibarr codebase. The integration only consists of copy-pasting a new file + application configuration, before going further.

Coming along with PR #22741. If you want to try, see how to sections below.

## Use case

Leverage federated identity which is the way to go. Whether to use SAML vs OpenID https://github.com/Dolibarr/dolibarr/issues/6600 is another story. In the meantime, Dolibarr may support both and it should be up to the end user to choose depending on his constraints.

### OpenID Server configuration

1. Check that grant type `Authorization code` is enabled.
2. Allowed callback URLs: `https://dolibarr.domain.com`
3. Allowed logout URLs: `https://dolibarr.domain.com`

A local URL like `http://localhost:1234` can also be used for testing.

### Note about Dolibarr user login

On the OpenID user profiles, there must be a claim (= user attribute) matching the Dolibarr user login, otherwise Dolibarr will say that the user could not be found in its database. Use this claim in `MAIN_AUTHENTICATION_OIDC_LOGIN_CLAIM`.

E.g.:
Bob used to log in into Dolibarr with login `bob42` and password `********`.
With OpenID Connect, Bob now uses his OpenID credentials to log in, for e.g. e-mail `bob@domain.com` and password `**********`. But Dolibarr still expects Bob's internal login `bob42` to recognize him for authentication, his user permissions etc. We must then be able to retrieve this login from Bob's profile on the Identity Provider.

If this login cannot be retrieved from the Identity Provider for some reason, the Dolibarr database must be edited and all user logins have to be updated to an existing claim on the OpenID profile (e.g.: `email`, or `nickname`). Hint: in the Dolibarr database, see table `llx_user`, column `login`.

### On the Dolibarr host

[...]

PR #22741, opened 2022-11-03
NEW #22740 Add OpenID Connect impl

jeritiana:
See context, use case and implementation details in #22740

### DIFF
--- a/htdocs/core/login/functions_openid_connect.php
+++ b/htdocs/core/login/functions_openid_connect.php
@@ -34,7 +34,7 @@ include_once DOL_DOCUMENT_ROOT.'/core/lib/geturl.lib.php';
  */
 function check_user_password_openid_connect($usertotest, $passwordtotest, $entitytotest)
 {
-	global $conf;
+	global $db, $conf, $langs;
 
 	dol_syslog("functions_openid_connect::check_user_password_openid_connect");
 
--- a/htdocs/core/login/functions_openid_connect.php
+++ b/htdocs/core/login/functions_openid_connect.php
@@ -76,7 +76,7 @@ function check_user_password_openid_connect($usertotest, $passwordtotest, $entit
 				$login_claim = $conf->global->MAIN_AUTHENTICATION_OIDC_LOGIN_CLAIM;
 			}
 
-			if (array_key_exists($login_claim, $userinfo_content)) {
+			if (property_exists($userinfo_content, $login_claim)) {
 				// Success: retrieve claim to return to Dolibarr as login
 				$login = $userinfo_content->$login_claim;
 				$_SESSION["dol_loginmesg"] = "";
--- a/htdocs/core/login/functions_openid_connect.php
+++ b/htdocs/core/login/functions_openid_connect.php
@@ -78,8 +78,33 @@ function check_user_password_openid_connect($usertotest, $passwordtotest, $entit
 
 			if (property_exists($userinfo_content, $login_claim)) {
 				// Success: retrieve claim to return to Dolibarr as login
-				$login = $userinfo_content->$login_claim;
-				$_SESSION["dol_loginmesg"] = "";
+				$sql = 'SELECT login, entity, datestartvalidity, dateendvalidity';
+				$sql .= ' FROM '.MAIN_DB_PREFIX.'user';
+				$sql .= " WHERE login = '".$userinfo_content->$login_claim."'";
+				$sql .= ' AND entity IN (0,'.(array_key_exists('dol_entity', $_SESSION) ? ((int) $_SESSION["dol_entity"]) : 1).')';
+	
+				dol_syslog("functions_openid::check_user_password_openid", LOG_DEBUG);
+				$resql = $db->query($sql);
+				if ($resql) {
+					$obj = $db->fetch_object($resql);
+					if ($obj) {
+						$now = dol_now();
+						if ($obj->datestartvalidity && $db->jdate($obj->datestartvalidity) > $now) {
+							// Load translation files required by the page
+							$langs->loadLangs(array('main', 'errors'));
+							$_SESSION["dol_loginmesg"] = $langs->transnoentitiesnoconv("ErrorLoginDateValidity");
+							return '--bad-login-validity--';
+						}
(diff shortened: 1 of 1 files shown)


hregis:
@jeritiana Thank you for this contribution, but on the other hand when do you check the existence of the user in Dolibarr?

jeritiana:
@hregis I don't check the existence of the user inside this module (or at least I didn't think I had to) because it is already handled somewhere else downstream. From my understanding of the [authentication documentation](https://wiki.dolibarr.org/index.php?title=Authentication), the goal of the login modules is to check the validity of a couple user/password. In the case of OpenID, this is delegated to the OpenID Identity Provider.
Which brings us to the following sequence:
1. The user provides his credentials login/password to the OpenID provider
2. This module checks from the OpenID provider if the authentication went well
3. If the authentication fails: the error message is displayed (e.g. Bad login/password)
  If the authentication succeeds on the OpenID provider: this module returns the login value to the calling function.
  Later, if the user (i.e the returned login) does **not** exist in Dolibarr, the error message "Unable to find user xxx" is displayed.

hregis:
@jeritiana yes but you don't check the existence of the user in Dolibarr !

hregis:
@jeritiana you can check the existing openid login method https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/core/login/functions_openid.php

jeritiana:
@hregis Okay I see what you mean. Isn't this already the purpose of [main.inc.php#L869](https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/main.inc.php#L869), leading to "User not found, connexion refused" and `ErrorCantLoadUserFromDolibarrDatabase`? i.e. check the existence of the user in Dolibarr?

Or should we also add this check inside an authentication module?

hregis:
@jeritiana in my opinion it is better to check if the user exists and is "active" before querying OpenID. In addition it will be necessary to add compatibility with Multicompany, this I can take care of.

hregis:
@jeritiana and why not have improved the OpenID login method already existing in Dolibarr?

jeritiana:
> @jeritiana in my opinion it is better to check if the user exists and is "active" before querying OpenID. In addition it will be necessary to add compatibility with Multicompany, this I can take care of.

Sure enough, let's try that. I will let you know of any progression. Thanks for taking care of the Multicompany compatibility.

> @jeritiana and why not have improved the OpenID login method already existing in Dolibarr?

Because the already existing login method is fully featured for OpenID 2.0. Modifying would be a breaking change for Dolibarr and there would be impacts on users, compared to adding a new next gen OpenID Connect support.

battosai30:
Line 49 there is an error I think : 
(GETPOSTISSET['code']) 
should be
(GETPOSTISSET('code'))

htdocs/core/login/functions_openid_connect.php
```diff
@@ -0,0 +1,114 @@
...
+			'grant_type'    => 'authorization_code',
+			'client_id'     => $conf->global->MAIN_AUTHENTICATION_OIDC_CLIENT_ID,
+			'client_secret' => $conf->global->MAIN_AUTHENTICATION_OIDC_CLIENT_SECRET,
+			'code'          => $auth_code,
+						'redirect_uri'  => $conf->global->MAIN_AUTHENTICATION_OIDC_REDIRECT_URL
+		];
+
+		$token_response = getURLContent($conf->global->MAIN_AUTHENTICATION_OIDC_TOKEN_URL, 'POST', http_build_query($auth_param));
+		$token_content = json_decode($token_response['content']);
+		dol_syslog("functions_openid_connect::check_user_password_openid_connect /token=".print_r($token_response, true), LOG_DEBUG);
```
battosai30, line 64: $token_response = getURLContent($conf->global->MAIN_AUTHENTICATION_OIDC_TOKEN_URL, 'POST', http_build_query($auth_param),1, array(),array('http', 'https'),2);
=> allow local identity provider (in my case, I use virtual machines on the same network so 192.168.xx.xx and it's blocked)

FlorentPoinsaut (changes requested):

htdocs/core/login/functions_openid_connect.php
```diff
@@ -0,0 +1,114 @@
...
+
+		if ($token_content->access_token) {
```
FlorentPoinsaut, line 66:
```suggestion
		if (property_exists($token_content, 'access_token')) {
```

htdocs/core/login/functions_openid_connect.php
```diff
@@ -0,0 +1,114 @@
...
+ * If test is ko, reason must be filled into $_SESSION["dol_loginmesg"]
+ *
+ * @param	string	$usertotest		Login
+ * @param	string	$passwordtotest	Password
+ * @param   int		$entitytotest   Number of instance (always 1 if module multicompany not enabled)
+ * @return	string					Login if OK, '' if KO
+ */
+function check_user_password_openid_connect($usertotest, $passwordtotest, $entitytotest)
+{
+	global $conf;
```
FlorentPoinsaut, line 37:
```suggestion
	global $db, $conf, $langs;
```

FlorentPoinsaut (changes requested):
Check if the user exists as in the original openid method

FlorentPoinsaut:
Hi @jeritiana 
Do you want to continue to maintain this PR or would you prefer that I create a new one?
Thank you in advance,

jeritiana:
Hi @FlorentPoinsaut , I will have a look and integrate the suggested changes. Thanks for the ping!

philippe-opendsi:
Hi, 
For the impatient, we have put this PR in a module available here. 
https://git.open-dsi.fr/dolibarr-extension/openidconnect
Working with us with a Lemon-LDAP
ping @eldy 
Regards

eldy:
Thanks @jeritiana 
Yes, i confirm as you suggest in the issue ticket that parameters may come from the conf.php as alternative instead of database. But this can be done in a next step (for the moment oauth2 ids are into database...
Goal is to have all param related to security in a layer executed before the code (managing the database access) and so safe from any database readability.

PR #22753, opened 2022-11-04
Fix: Problem with ODT generation on lines

### LINKED ISSUE #16184: Probleme with line ODT generation
# Instructions
*This is a template to help you report good issues. You may use [Github Markdown](https://help.github.com/articles/getting-started-with-writing-and-formatting-on-github/) syntax to format your issue report.*
*Please:*
- *remove any unused sub-section*

# Bug
Generation invoices ODT

## Environment
- Dolibarr 13: 13
- Debian 
- Nginx
- 7.4
- Maria DB

It generates line with html balises like this 
Capture

When I generate an odt file

thomas-Ngr:
# FIX #16184 Problem with ODT generation on lines

In #14974, a line has been introduced to convert '&' chars to htmlentities. This broke the creation of ODTs with lines containing HTML. This PR fixes this for the tags that are managed in the code (br, b, strong, i, u, s, em, sub, sup, span in _replaceHtmlWithOdtTag() ).

There was many problems:

1. Lines description with html are already passed through htmlspecialchars(). Doing it another time in convertVarToOdf() will make the converted chars appear in the ODT. as mentioned in #16184.
2. Passing the values through htmlspecialchars() 'erases' the html tags. `$this->_hasHtmlTag($value)` will always be `false`. So, we have to re-set HTML closing and opening brackets.
3. HTML tags have to be converted to XML tags. But only some tags are dealt with. So we have to strip_tags() the values, authorizing only br, b, strong, i, u, s, em, sub, sup, span.
4. We should keep line breaks in the ODT file. This keeps a minimal formatting for lists in the ODT file. For example, this is an HTML `<ul>` with new code : 
image

____

edit : clarifications

We have this line description in an invoice. It has been generated with WYSISWYG editor, but manual HTML should have the same behavior:
image

With current ODT generation, the output displays HTML tags. One should reproduce it each time they create an ODT with lines description that contain HTML:
image

This is due to:
- description are passed through htmlspecialchars() before being saved in the DB
- #14974 introduced another  htmlspecialchars() to fix special chars `&` in thirdparty name
- So, descriptions are passed two times through htmlspecialchars().

We have to:
- decode descriptions from database
- deal with html->xml tags conversion
- strip HTML tags that are not managed for safety and odt readability
- keep htmlspecialchars() for normal text

This PR fixes this issue and introduces a unit test to expose and lock expected behavior.
Now when generating an ODT we have correct tags.
As you can see, this is a fix: lists are not managed.

image


htdocs/includes/odtphp/odf.php
```diff
@@ -156,7 +157,21 @@ public function setVars($key, $value, $encode = true, $charset = 'ISO-8859')
      */
 	public function convertVarToOdf($value, $encode = true, $charset = 'ISO-8859')
 	{
-		$value = $encode ? htmlspecialchars($value) : $value;
+		if ($encode) {
```
eldy, line 160: It is not possible to understand if this change is good or not without samples in input and expected value to compare in output.
So can you introduce a phpunit test that validate that convertVarToOdf do the right thing ?
How can we reproduce the bug ?
Do you have example of value input and output that is wrong currently ? the source ticket is not clear and it seems each change to fix a trouble break another one :-(

thomas-Ngr: Thanks for the advice. Writing a test helped me fix my original PR.
I edited the PR message for clarifications.

thomas-Ngr:
~~I just encountered a case that I didn't foresee :  HTML heading tags such as `<h3>` still appear in the odt.
It might come from _hasHtmlTags() that would not return true on heading tags.
Please do not merge for now.~~

thomas-Ngr:
> I just encountered a case that I didn't foresee : HTML heading tags such as `<h3>` still appear in the odt. It might come from _hasHtmlTags() that would not return true on heading tags. Please do not merge for now.

I have fixed this issue in the two last commits.

thomas-Ngr:
PR is old. Re-creating it.

PR #22756, opened 2022-11-04
FIX: Reload page on supplier change and keeping the values

kkhelifa-opendsi:
FIX: Reload page on supplier change and keeping the values of the input already set in the form


eldy:
Ergonomy enhancement can't be added into already released stable version.
Can you push it on develop ?

PR #22757, opened 2022-11-04
FIX: Fix the prospecting level column on contact list

kkhelifa-opendsi:
Fix the prospecting level column on contact list by normalize the field namein the table with the name defined in the fields list of the contact class


eldy (commented):
Reengeneering and changing database structure is not allowed into an already released version. Can you push on develop ?

PR #22772, opened 2022-11-06
NEW Accountancy - Model Digitaria - Add a way to clean some words when you generate thirdparty accounting account FPC22

aspangaro:
image

image


aspangaro:
https://github.com/Easya-Solutions/dolibarr/pull/185

PR #22777, opened 2022-11-07
increase size of email field

defrance:
we use 2 other col on the right to extends the size of fields


htdocs/societe/card.php
```diff
@@ -2346,7 +2346,7 @@ function init_supplier_categ() {
 
 			// EMail
 			print '<tr><td>'.$form->editfieldkey('EMail', 'email', GETPOST('email', 'alpha'), $object, 0, 'string', '', (!empty($conf->global->SOCIETE_EMAIL_MANDATORY))).'</td>';
-			print '<td>';
+			print '<td colspan="3">';
```
eldy, line 2349: It seems we have 4 columns per tr. If you add colspan=3 here, we will use 6 columns

frederic34:
image

eldy:
CI error since a too long time, so ticket closed.

PR #22778, opened 2022-11-07
FIX syntax error and add a constant for not to delete unassigned contact

hregis:
@eldy All internal contacts were deleted when reassigning a ticket. I also added a constant so as not to delete the old assigned user from the ticket contacts. This bug is also present in v15, v16, develop

### DIFF
--- a/htdocs/ticket/card.php
+++ b/htdocs/ticket/card.php
@@ -381,20 +381,18 @@ if (empty($reshook)) {
 
 		if (!$error) {    // Update list of contacts
 			// Si déjà un user assigné on le supprime des contacts
-			if ($useroriginassign > 0) {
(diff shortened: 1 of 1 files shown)


htdocs/ticket/card.php
```diff
@@ -381,20 +381,18 @@
...
-					}
-					{
 						//print "user à effacer : ".$useroriginassign;
 						$object->delete_contact($contact['rowid']);
 					}
 				}
 			}
 
-			if ($usertoassign > 0) {
+			if ($usertoassign > 0 && $usertoassign != $useroriginassign) {
```
eldy, line 395: At line 384, we force the deletion of link contact for $useroriginassign, so we can make the add_contact at no risk to insert it twice. But if we add test  $usertoassign != $useroriginassign  and we assign to them, we first delete and then we do not recreate it loosing the contact, right ?

eldy:

PR #22782, opened 2022-11-07
NEW Reception - Add a real reception date  rather than an unmodifiable date_create

aspangaro:
- Add a real date reception field (Already present in database...)
- Link object based on this date
- Stock movement based also on this date, not on date create. if you are late for your data entry, this will cause quite large stock differences and without any possibility to modify it

image

image


eldy:
Having a date stored into database to use it as date for real reception does not seems to be a good idea. This was already the goal of the field "Date planned for reception". But because this is not accurate to enter a date before knowing what is the real date, we currently use another date:
- If stock change is setup to be done on reception validation, then we use the date of validation. 
- If stock change is planned to be on reception closing, then we use the date of closing.
If this 2 dates are not accurate (because for exemple we record the validation or the closing 2 days after), it means we must ask the date on the popup of confirmation of this 2 steps.
image

And we must use this date (setup by default to current date) as the date for real stock change.

josett225:
Hi there,
Not sure I understood all in the last comments, but the life of reception in logistics is not always that someone is behind a PC to enter data at the time the real reception is made. Therefore having this feature is a must have for several companies but could be delivered behind a CONST as I understand both scenarios.
The reception date has also sometimes to be changed for error reasons due to high traffic or human mistahes. Stock date movement has to follow the changes. This is important to stock date value calculation.

PR #22786, opened 2022-11-08
FIX wrong stock list with multicompany and without stock sharing

hregis:
@eldy this bug affect v14, v15, v16 and develop

### DIFF
--- a/htdocs/product/reassort.php
+++ b/htdocs/product/reassort.php
@@ -141,7 +141,7 @@ $reshook = $hookmanager->executeHooks('printFieldListSelect', $parameters); // N
 $sql .= $hookmanager->resPrint;
 $sql .= ' FROM '.MAIN_DB_PREFIX.'product as p';
 $sql .= ' LEFT JOIN '.MAIN_DB_PREFIX.'product_stock as s ON p.rowid = s.fk_product';
-$sql .= ' RIGHT JOIN '.MAIN_DB_PREFIX.'entrepot as e ON s.fk_entrepot = e.rowid AND e.entity IN ('.getEntity('entrepot').')';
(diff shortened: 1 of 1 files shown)


htdocs/product/reassort.php
```diff
@@ -141,7 +141,7 @@
 $sql .= $hookmanager->resPrint;
 $sql .= ' FROM '.MAIN_DB_PREFIX.'product as p';
 $sql .= ' LEFT JOIN '.MAIN_DB_PREFIX.'product_stock as s ON p.rowid = s.fk_product';
-$sql .= ' LEFT JOIN '.MAIN_DB_PREFIX.'entrepot as e ON s.fk_entrepot = e.rowid AND e.entity IN ('.getEntity('entrepot').')';
+$sql .= ' RIGHT JOIN '.MAIN_DB_PREFIX.'entrepot as e ON s.fk_entrepot = e.rowid AND e.entity IN ('.getEntity('stock').')';
```
eldy, line 144: Why a RIGHT JOIN. We already have a LEFT JOIN so we take all product whateever is link into product_stock, so adding a RIGHT JOIN here will means we also have another master table. We can't have 2 master table in same SQL as this lead to unpredicable results.

hregis: @eldy INNER JOIN ? or move `e.entity IN ('.getEntity('stock').')` after the WHERE ?

hregis: @eldy i replace with INNER JOIN

eldy: If you use a inner join, if the product has not stock, the line will not be returned.

hregis: @eldy "reassort" with no stock (entrepot)  it's possible ? :-)

PR #22796, opened 2022-11-09
FIX: Invoice status was never shown in draft or validated status in getNomUrl popup

noec764:
Capture d’écran de 2022-11-09 11-11-03

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -1,4 +1,4 @@
-<?php
+1<?php
 /* Copyright (C) 2002-2007  Rodolphe Quiedeville    <rodolphe@quiedeville.org>
  * Copyright (C) 2004-2013  Laurent Destailleur     <eldy@users.sourceforge.net>
  * Copyright (C) 2004       Sebastien Di Cintio     <sdicintio@ressource-toi.org>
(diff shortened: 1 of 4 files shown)


htdocs/compta/facture/class/facture.class.php
```diff
@@ -1,4 +1,4 @@
-<?php
+1<?php
```
eldy, line 1: Typo error here

htdocs/compta/facture/class/facture.class.php
```diff
@@ -1520,8 +1520,12 @@ public function getNomUrl($withpicto = 0, $option = '', $max = 0, $short = 0, $m
 			if ($this->type == self::TYPE_SITUATION) {
 				$label = img_picto('', $picto).' <u class="paddingrightonly">'.$langs->transnoentitiesnoconv("InvoiceSituation").'</u>';
 			}
-			if (isset($this->statut) && isset($this->alreadypaid)) {
-				$label .= ' '.$this->getLibStatut(5, $this->alreadypaid);
+			if (isset($this->statut)) {
+				if (!isset($this->alreadypaid)) {
+					$this->alreadypaid = $this->getSommePaiement();
```
eldy, line 1525: Having a call of getSommePaiement that make heavy database access inside a method dedicated to output formatting is  a problem. The pro is that it solves trouble everywhere. The cons is that this need a high cost in performance, and does not match good practices: no database access for formatting output.
Can you try to make the getSommePaiement on the caller side (when on caller side, we often have already the data or we will be able later to use a cache for each invoice, so cost will be paid only if required).

noec764:
Now Facture::alreadypaid need to be filled before static call to Facture::getNomurl(), it was already done on /facture/list.php and facture/index.php but maybe I forgot to some static Facture object

noec764:
@eldy What should I do about this  PR ?

noec764:
@eldy Can you check this PR again ?

noec764:
@eldy

noec764:
@eldy Can you check this PR again please ?

eldy:
We should not have subrequest $this->getSommePaiement() into low level fetch() methods (only subrequest to get extrafields, and lines for objects, like to get invoice lines for invoices is allowed into fetch).
With v18, another method to show the popup has been introduced with option MAIN_ENABLE_AJAX_TOOLTIP
With v19, this option will be on by default.
With this option, an ajax call is done when we show the popup, so database access is done only when required so this may be the solution to solve this bug without the need to make a heavy SQL inside the core low level Read method fetch(). Can you check it is ok in v18 with this option on ?

eldy:

Issue #6835, opened 2017-05-11
Feature Request: expand/collapse user permission module


dolibarr95:
# Feature Request
Hi

I think it could be nice if in the page : 
- dolibarr/user/perms.php?id=45

All the lines of a module are hidden by default. And display if the user click on the module name.
Actually the page perms.php can be very long (depends of number of modules installed).

## Suggested implementation
Add a toggle() jquery to the table.

## Suggested steps

- [ ] give a attribute name to each <tr> lines (that's contains rights information of a module ) of the table and set "display:none;" style to theses lines.
The name could be something like this : ID_LINE where ID is the id of the module.
- [ ] ad an onclik .toggle() event to each <tr> lines (that's contains name of the module) of the table.
It could be something like this : $( "tr[name=250_LINE]" ).toggle()

Tks

dolibarr95:
at line 327 of the perms.php file : 
```
$style = '';
		if (isset($obj->module) && ($oldmod <> $obj->module))
		{
			$oldmod = $obj->module;
			$var = !$var;

			// Rupture detectee, on recupere objMod
			$objMod=$modules[$obj->module];
			$picto=($objMod->picto?$objMod->picto:'generic');
			
        	if ($caneditperms && (empty($objMod->rights_admin_allowed) || empty($object->admin)))
        	{
        		// On affiche ligne pour modifier droits
        		print '<tr '. $bc[$var].' onclick="$( \'tr[name='.$obj->module.'_LINE]\' ).toggle();" style="cursor:pointer">';
        		print '<td class="maxwidthonsmartphone tdoverflowonsmartphone">'.img_object('',$picto).' '.$objMod->getName();
        		print '<a name="'.$objMod->getName().'"></a></td>';
        		print '<td align="center" class="nowrap">';
        		print '<a class="reposition" title="'.dol_escape_htmltag($langs->trans("All")).'" alt="'.dol_escape_htmltag($langs->trans("All")).'" href="perms.php?id='.$object->id.'&amp;action=addrights&amp;entity='.$entity.'&amp;module='.$obj->module.'">'.$langs->trans("All")."</a>";
        		print '/';
[...]

PR #22801, opened 2022-11-09
New : expand/collapse user permission module 

Hystepik:
Create a way to expand / collapse user permission as in #6835
With 1 row expanded :
image

Without row expanded:
image


eldy:
How do you expand a section (by clicking on the line title) ? May be you can add the picto expand/collapse like found into page Tags/categories of menu Product ?

Hystepik:
You can click on the line title or click on the top button Show all permission rows to expand. Yes this can be a good thing

Hystepik:
New look:
image
image

PR #22803, opened 2022-11-09
Fix issue scrutinizer htdocs/core/lib/functionsnumtoword.lib.php 

lamrani002:
# FIX|Fix  issue in The property default does not exist on Translate 
 change property default by function in Translate::getDefaultLang(1)
link to issue : https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/core/lib/functionsnumtoword.lib.php?selectedSeverities%5B0%5D=10&orderField=path&order=asc&fileId=htdocs/core/lib/functions.lib.php&honorSelectedPaths=0

### DIFF
--- a/htdocs/core/lib/functionsnumtoword.lib.php
+++ b/htdocs/core/lib/functionsnumtoword.lib.php
@@ -187,7 +187,7 @@ function dolNumberToWord($numero, $langs, $numorcurrency = 'number')
 
 	/*In dolibarr 3.6.2 (my current version) doesn't have $langs->default and
 	in case exist why ask $lang like a parameter?*/
-	if (((is_object($langs) && $langs->getDefaultLang(1) == 'es_MX') || (!is_object($langs) && $langs == 'es_MX')) && $numorcurrency == 'currency') {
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functionsnumtoword.lib.php
```diff
@@ -187,7 +187,7 @@ function dolNumberToWord($numero, $langs, $numorcurrency = 'number')
 
 	/*In dolibarr 3.6.2 (my current version) doesn't have $langs->default and
 	in case exist why ask $lang like a parameter?*/
-	if (((is_object($langs) && $langs->default == 'es_MX') || (!is_object($langs) && $langs == 'es_MX')) && $numorcurrency == 'currency') {
+	if (((is_object($langs) && $langs->getDefaultLang(1) == 'es_MX') || (!is_object($langs) && $langs == 'es_MX')) && $numorcurrency == 'currency') {
```
eldy, line 190: You replace ->default with getDefaultLang(1)
I think it should be getDefaultLang(0)
According to dof of getDefaultLang(), if using parameter 1, result will return a lang code on 2 char, but comparison is done with es_MX so we need the lang code on 5 chars, so with param 0 and not 1

PR #22804, opened 2022-11-09
Fix issue Scrutinizer in htdocs/core/lib/json.lib.php 

lamrani002:
# FIX|Fix issue Scrutinizer 
change equal operator to identical operator.
**link to issue** : https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/core/lib/json.lib.php?selectedSeverities%5B0%5D=10&orderField=path&order=asc&fileId=htdocs/core/lib/functionsnumtoword.lib.php&honorSelectedPaths=0

### DIFF
--- a/htdocs/core/lib/functionsnumtoword.lib.php
+++ b/htdocs/core/lib/functionsnumtoword.lib.php
@@ -187,7 +187,7 @@ function dolNumberToWord($numero, $langs, $numorcurrency = 'number')
 
 	/*In dolibarr 3.6.2 (my current version) doesn't have $langs->default and
 	in case exist why ask $lang like a parameter?*/
-	if (((is_object($langs) && $langs->default == 'es_MX') || (!is_object($langs) && $langs == 'es_MX')) && $numorcurrency == 'currency') {
+	if (((is_object($langs) && $langs->getDefaultLang(1) == 'es_MX') || (!is_object($langs) && $langs == 'es_MX')) && $numorcurrency == 'currency') {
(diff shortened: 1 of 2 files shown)


htdocs/core/lib/functionsnumtoword.lib.php
```diff
@@ -187,7 +187,7 @@ function dolNumberToWord($numero, $langs, $numorcurrency = 'number')
 
 	/*In dolibarr 3.6.2 (my current version) doesn't have $langs->default and
 	in case exist why ask $lang like a parameter?*/
-	if (((is_object($langs) && $langs->default == 'es_MX') || (!is_object($langs) && $langs == 'es_MX')) && $numorcurrency == 'currency') {
+	if (((is_object($langs) && $langs->getDefaultLang(1) == 'es_MX') || (!is_object($langs) && $langs == 'es_MX')) && $numorcurrency == 'currency') {
```
eldy, line 190: You replace ->default with getDefaultLang(1)
I think it should be getDefaultLang(0)
According to dof of getDefaultLang(), if using parameter 1, result will return a lang code on 2 char, but comparison is done with es_MX so we need the lang code on 5 chars.

lamrani002: effectivement du coup j'ai juste ajouter un commentaire car l'operateur === est déja résolu

lamrani002 (commented):
si le probléme est l'operateur === il est déja fixé

PR #22817, opened 2022-11-10
Fix issue Scrutinizer in core/modules/project/mod_project_universal.php

lamrani002:
# FIX|Fix issue by declare $db in class 
link to issue : https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/core/modules/project/mod_project_universal.php?selectedSeverities%5B0%5D=10&orderField=path&order=asc&fileId=htdocs/core/modules/project/doc/doc_generic_project_odt.modules.php&honorSelectedPaths=0


htdocs/core/lib/project.lib.php
```diff
@@ -1124,9 +1124,9 @@ function projectLinesPerAction(&$inc, $parent, $fuser, $lines, &$level, &$projec
 			$projectstatic->ref = $lines[$i]->project_ref;
 			$projectstatic->title = $lines[$i]->project_label;
 			$projectstatic->public = $lines[$i]->public;
-			$projectstatic->status = $lines[$i]->project_status;
+			$projectstatic->status = $lines[$i]->project->status;
```
eldy, line 1127: When checking if this is a good fix, i discovered that the method projectLinesPerAction() is called nowhere in Dolibarr. So this is dead code. So i pushed a PR that removed this method.
This means your PR is no in conflcit.
Thanks to :
- reopen your branch of this fix
- pull last version of develop branch
- fix the conflict (method removed against method modified -> good version must be method removed)
- resubmit the PR so the change into the other file can be published

lamrani002: done

PR #22819, opened 2022-11-10
FIX: GetNextValue() adding regexsql() function

noec764:
[Related To #21533](https://github.com/Dolibarr/dolibarr/pull/21533/files)

I realized that "CAST %s AS INTEGER" can't be used in postgre because in case of non integer value it will raise an error
I decide to implements the sql function REGEXP to be more more general.


eldy:
Introducing new methods can be done into an already published version.
Thanks to submit PR into develop branch.

noec764:
@eldy 
Could you look at this PR again ?

PR #22832, opened 2022-11-13
NEW - Shipping address to invoice

BB2A-Anthony:
# NEW|New [*Shipping address to invoice*]

Add shipping address to invoice

If a delivery address is defined, it is displayed.
If a billing address is defined, I consider that the delivery address is the one on which the invoice was made.

image


StephaneBRT:
very useful feature. thank you very much.

eldy:
Great. 
The template crabe is deprecated and will be replaced soon by template sponge. So can you add also the same change into template sponge ?    Both templates are similar so change should be cut and paste (i hope).
Because this is legal need, this new feature could be merged despite the freeze.

BB2A-Anthony:
> Great. 
> 
> The template crabe is deprecated and will be replaced soon by template sponge. So can you add also the same change into template sponge ?    Both templates are similar so change should be cut and paste (i hope).
> 
> Because this is legal need, this new feature could be merged despite the freeze.

In my case, I only display it if it is different from the billing address! Should there always be a delivery address even if it is the same?

Dans mon cas je ne l'affiche que si elle est différente de l'adresse de facturation! Faut il qu'il y ai toujours une adresse de livraison même si elle est la même?

BB2A-Anthony:
> Great. The template crabe is deprecated and will be replaced soon by template sponge. So can you add also the same change into template sponge ? Both templates are similar so change should be cut and paste (i hope). Because this is legal need, this new feature could be merged despite the freeze.

@eldy it's done and I improved the code

htdocs/admin/pdf_other.php
```diff
@@ -69,6 +69,10 @@
 		dolibarr_set_const($db, "INVOICE_ADD_SWISS_QR_CODE", GETPOST("INVOICE_ADD_SWISS_QR_CODE", 'int'), 'chaine', 0, '', $conf->entity);
 		dolibarr_del_const($db, "INVOICE_ADD_ZATCA_QR_CODE", $conf->entity);
 	}
+	if (GETPOSTISSET('DOC_SHOW_SHIPPING_ADDRESS')) {
```
eldy, line 72: Do you mean DOC_SHOW_SHIPPING_ADDRESS instead of DOC_SHOW_SHIPPING_ADDRESS ?

bb2a: I have a change @eldy

hasse-basse:
As a clarification, at least in France, it is not a legal requirement to mention the delivery address on invoices when conducting B2B sales. This requirement is typically applicable when selling to consumers (B2C). Therefore, it would be beneficial to have a checkbox option on the invoice card, allowing to choose whether or not to include the delivery address for each specific invoice. Or maybe it could be controlled in the settings of what kind of client it is? Remeber that in B2B sales there can be many delivery adresses on one invoice.

PR #22838, opened 2022-11-14
New - trigger online sign

Brice-VINCENT:
# NEW|New [*Trigger Online Sign*]
[*Tool to send an email when signing an online quote.*]

### DIFF
--- a/htdocs/core/ajax/onlineSign.php
+++ b/htdocs/core/ajax/onlineSign.php
@@ -200,6 +200,16 @@ if ($action == "importSignature") {
 
 				dol_syslog(__METHOD__, LOG_DEBUG);
 				$resql = $db->query($sql);
+				//Trigger propal signe en ligne
+				if (method_exists($object, 'call_trigger')) {
+					$user = new User($db);
+					$user->fetch($object->user_valid_id);
+					$result = $object->call_trigger('PROPAL_CLOSE_SIGNED_ONLINE', $user);
+					if ($result < 0) {
+						$error++;
+					}}
+					//End of modification
+	
 				if (!$resql) {
 					$error++;
 				} else {
--- a/htdocs/core/ajax/onlineSign.php
+++ b/htdocs/core/ajax/onlineSign.php
@@ -209,7 +209,7 @@ if ($action == "importSignature") {
 						$error++;
 					}}
 					//End of modification
-	
+
 				if (!$resql) {
 					$error++;
 				} else {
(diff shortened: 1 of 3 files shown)


htdocs/core/ajax/onlineSign.php
```diff
@@ -200,6 +200,16 @@
 
 				dol_syslog(__METHOD__, LOG_DEBUG);
 				$resql = $db->query($sql);
+				//Trigger propal signe en ligne
+				if (method_exists($object, 'call_trigger')) {
+					$user = new User($db);
+					$user->fetch($object->user_valid_id);
+					$result = $object->call_trigger('PROPAL_CLOSE_SIGNED_ONLINE', $user);
```
eldy, line 207: There is no need to duplicate the trigger PROPAL_CLOSE_SIGNED into a variant PROPAL_CLOSE_SIGNED_ONLINE.
What you can do is just add the line
`$object->context = array('closedfromonlinesignature' => 'closedfromonlinesignature')`
before the 
$result = $object->call_trigger('PROPAL_CLOSE_SIGNED', $user); line 227

htdocs/public/onlinesign/newonlinesign.php
```diff
@@ -193,6 +193,15 @@
 	}
 
 	$object->fetch(0, $ref);
+	//Trigger propal signe en ligne
```
eldy, line 196: I pushed a commit to add the context on $object so this trigger call is useless.

htdocs/core/triggers/interface_70_modpropale_MailPropalSigne.class.php
```diff
@@ -0,0 +1,236 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+/**
+ *  \file       htdocs/core/triggers/interface_70_modpropale.MailPropalSigne.class.php
```
eldy, line 19: Rename your trigger interface_50_modPropale_Mail.class.php and use the trigger code PROPAL_CLOSE_SIGNED
and send email when $object->context contains the key closedfromonlinesignature

BB2A-Anthony:
Hi its good job 

if the proposal is refused it also works?

eldy:
> Hi its good job
> 
> if the proposal is refused it also works?

I think so, the context has also been added when trigger PROPAL_CLOSE_REFUSED is called
```
		$object->context = array('closedfromonlinesignature' => 'closedfromonlinesignature');
		$result = $object->call_trigger('PROPAL_CLOSE_REFUSED', $user);

```

eldy:
An other solution was implemented differently so PR is closed.

BB2A-Anthony:
What the solution

PR #22839, opened 2022-11-14
Fix #11555 : Message display on adherent public

### LINKED ISSUE #11555: Extrafields: no message displayed if a required extrafield is missing in the members autosubscription formular
# Bug
When validating the members autosubscription formular (public/members/new.php) with missing required extrafields, no message is displayed, the formular is not validated (back to top of page) but a subscription request email is sent.

## Environment
- **Version**: 10.0.0
- **Web server**: Apache

The autosubscription formular (public/members/new.php) does not have the same behavior for required Dolibarr standard fields and for the extrafields concerning the error messages: when a required Dolibarr standard field is missing, a message is displayed in red but when a required extrafield is missing, no error message is displayed but the formular is back to top, the data are not validated but a subscription request is sent by email.
It is necessary to show an error message the same way as the standard Dolibarr fields if a required extrafield is missng. No email should be sent until the data are not validated and saved in the database.

Use the standard autosubscription formular ((public/members/new.php) with mandatory extrafields.

Hystepik:
Fix an error where object was trying to be created when a required extrafields was not set and there was no messages for the required empty extrafield

### DIFF
--- a/htdocs/public/members/new.php
+++ b/htdocs/public/members/new.php
@@ -289,127 +289,129 @@ if (empty($reshook) && $action == 'add') {
 			$error++;
 		}
 
-		$result = $adh->create($user);
-		if ($result > 0) {
-			require_once DOL_DOCUMENT_ROOT.'/core/class/CMailFile.class.php';
-			$object = $adh;
+		if (!$error) {
+			$result = $adh->create($user);
+			if ($result > 0) {
+				require_once DOL_DOCUMENT_ROOT.'/core/class/CMailFile.class.php';
+				$object = $adh;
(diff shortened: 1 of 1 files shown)


htdocs/public/members/new.php
```diff
@@ -469,6 +471,7 @@ function llxFooterVierge()
 print '</div>';
 
 dol_htmloutput_errors($errmsg);
+dol_htmloutput_events();
```
eldy, line 474: If we add dol_htmloutput_events(), do we still need dol_htmloutput_errors() ?

PR #22841, opened 2022-11-14
NEW Add abbility of ODT support to supplier invoices

artemChernitsov:
# New [*Add ability of ODT support to supplier invoices*]
Right now Supplier Invoices usually gets PDF from Supplier Order, but for most users it doesn't applicable. So, right now users can add custom ODT templates into system.


htdocs/core/modules/modFournisseur.class.php
```diff
@@ -95,14 +95,25 @@ public function __construct($db)
 		$this->const[$r][4] = 0;
 		$r++;
 
-		/* For supplier invoice, we must not have default pdf template on. In most cases, we need to join PDF from supplier, not have a document generated.
+		/* OLD For supplier invoice, we must not have default pdf template on. In most cases, we need to join PDF from supplier, not have a document generated.
+		   NEW Uncomment to add ability to generate PDF for Supplier Invoices which generated not from Order.
+		*/
```
eldy, line 100: I think we must keep the constant INVOICE_SUPPLIER_ADDON_PDF not enabled by default.
So by default no generation of PDF is done for supplier invoice.

artemChernitsov: I can describe better cases, where I need PDF generation directly from Supplier Invoice. 
In my business model I have third parties (vendors), and I have not any orders/purchases etc., but I want to create Vendor invoice for these third parties and to have ability generate PDF for this invoice (ofc with custom ODT templates)

eldy: Yes i mean keep INVOICE_SUPPLIER_ADDON_PDF empty by default. If you need it, you can enable the template. But default must match most common use, so no doc generation at all.

PR #22842, opened 2022-11-14
Fix issue in htdocs/don/payment/payment.php

lamrani002:
# FIX|Fix scrutinizer in htdocs/don/payment/payment.php
 By changin datepai by datep and paymenttype by fk_typepayment and accountid by fk_account  

**link to issue :** https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/don/payment/payment.php?selectedSeverities%5B0%5D=10&selectedSeverities%5B1%5D=5&orderField=path&order=asc&fileId=htdocs/don/class/paymentdonation.class.php&honorSelectedPaths=0


htdocs/don/payment/payment.php
```diff
@@ -192,14 +192,14 @@
 	print '</tr>';
 
 	print '<tr><td class="fieldrequired">'.$langs->trans("PaymentMode").'</td><td colspan="2">';
-	$form->select_types_paiements(GETPOSTISSET("paymenttype") ? GETPOST("paymenttype") : $object->paymenttype, "paymenttype");
+	$form->select_types_paiements(GETPOSTISSET("paymenttype") ? GETPOST("paymenttype") : $object->fk_typepayment, "paymenttype");
```
eldy, line 195: There was a confusion (and there is still a confusion) between paymenttype and fk_typepayment that are 2 variables used for same thing when fk_typepayment should contain the ID (rowid) of type of payment and paymenttype should contains the code (but still contains an ID sometime).
So this change is sensitive due to old code not correctly written.
So to avoid regression, can you tell me if you succeed in creating a payment on a donation after making this change ?

htdocs/don/payment/payment.php
```diff
@@ -192,14 +192,14 @@
...
 	print "</td>\n";
 	print '</tr>';
 
 	print '<tr>';
 	print '<td class="fieldrequired">'.$langs->trans('AccountToCredit').'</td>';
 	print '<td colspan="2">';
-	$form->select_comptes(GETPOSTISSET("accountid") ? GETPOST("accountid") : $object->accountid, "accountid", 0, '', 2); // Show open bank account list
+	$form->select_comptes(GETPOSTISSET("accountid") ? GETPOST("accountid") : $object->fk_account, "accountid", 0, '', 2); // Show open bank account list
```
eldy, line 202: $object seems to be the class new Don()
For this class, there is no bank account into database, into the fetch and into screen to create and edit.
So i think the fix is different and we should replace the $object->accountid with "0"  (there is no default bank account stored into a donation that we can reuse here as a default value for the combobox).

lamrani002: Ok i see i thought the first param of the function should be required and contain a value, now i understand it can have also a default value but why "0" in place " " ?

eldy: Yes, having 0 instead of $object->accountid that is undef seems the good way to fix this case.

lamrani002: ok i will fix it

PR #22843, opened 2022-11-14
Fix issue  scrutinizer in htdocs/expedition/list.php

lamrani002:
# FIX|Fix issue by changing attribute array_fields to array_options 
**link to issue :** https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/expedition/list.php?selectedSeverities%5B0%5D=10&selectedSeverities%5B1%5D=5&orderField=path&order=asc&fileId=htdocs/expedition/class/expedition.class.php&honorSelectedPaths=0

### DIFF
--- a/htdocs/expedition/list.php
+++ b/htdocs/expedition/list.php
@@ -149,7 +149,7 @@ $arrayfields = array(
 // Extra fields
 include DOL_DOCUMENT_ROOT.'/core/tpl/extrafields_list_array_fields.tpl.php';
 
-$object->fields = dol_sort_array($object->fields, 'position');
+$object->array_options = dol_sort_array($object->array_options, 'position');
 $arrayfields = dol_sort_array($arrayfields, 'position');
 
 


htdocs/expedition/list.php
```diff
@@ -149,7 +149,7 @@
 // Extra fields
 include DOL_DOCUMENT_ROOT.'/core/tpl/extrafields_list_array_fields.tpl.php';
 
-$object->fields = dol_sort_array($object->fields, 'position');
+$object->array_options = dol_sort_array($object->array_options, 'position');
```
eldy, line 152: According to our template code into   modulebuilder/template/myobject_list.php, old code seems good.
Problem here is that the $object = Expedition does not has the ->$fields array.
In a future all objects must have this array.
So i pushed a different fix by adding just 
	public $fields = array();
into Expedition class. In a future we will have to replace it with the ->fields that describe all fields of the object.

PR #22858, opened 2022-11-15
fix: always display security warning on home page

FHenry:
For all user and admin MAIN_REMOVE_INSTALL_WARNING shouldn't exists. We are no more in CareBear world

Bad Dolibarr installation should always trigger a warning


FHenry:
if merge remove MAIN_REMOVE_INSTALL_WARNING  from https://wiki.dolibarr.org/index.php?title=Setup_Other

hregis:
@FHenry and this https://github.com/Dolibarr/dolibarr/pull/22862/files

Zulgrib:
It shouldn't be possible to hide from administrators the install.lock warning.
I don't know if this warning is shown to non-admin users, if this is the case, the constant MAIN_REMOVE_INSTALL_WARNING should only affect non-admin users.

Also, on first install (not upgrade), it should ask if we want to create the install.lock file. If accepted, "MAIN_ALWAYS_CREATE_LOCK_AFTER_LAST_UPGRADE" should be set to 1 and the install.lock file created.

PR #22862, opened 2022-11-16
FIX SECURITY don't create an admin if an admin already exists

hregis:
@eldy @FHenry URGENT for 14, 15, 16 and dev !!

### DIFF
--- a/htdocs/user/class/user.class.php
+++ b/htdocs/user/class/user.class.php
@@ -1318,8 +1318,10 @@ class User extends CommonObject
 				$this->db->free($resqltochecklogin);
 			}
 		}
-		if ($this->email !== '') {
+
+		if (!empty($this->email)) {
 			$sqltochecklogin = "SELECT COUNT(*) as nb FROM ".MAIN_DB_PREFIX."user WHERE entity IN (".$this->db->sanitize((int) $this->entity).", 0) AND email = '".$this->db->escape($this->email)."'";
+			echo $sqltochecklogin;
 			$resqltochecklogin = $this->db->query($sqltochecklogin);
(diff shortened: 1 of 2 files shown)


htdocs/user/class/user.class.php
```diff
@@ -1318,8 +1318,10 @@ public function create($user, $notrigger = 0)
 				$this->db->free($resqltochecklogin);
 			}
 		}
-		if ($this->email !== '') {
+
+		if (!empty($this->email)) {
 			$sqltochecklogin = "SELECT COUNT(*) as nb FROM ".MAIN_DB_PREFIX."user WHERE entity IN (".$this->db->sanitize((int) $this->entity).", 0) AND email = '".$this->db->escape($this->email)."'";
+			echo $sqltochecklogin;
```
altairis-tof, line 1324: @hregis je crois que tu as laissé un message de debug ;-)

hregis: @altairis-tof yep ! merci :-)

eldy:
The goal of the installer that is also a reinstaller is to create and to recreate an admin user when installation is not complete or hangs before end.

I have some doubt that it is a good idea to block the feature to be able to complete/finish correctly an installation. A dolibarr must not be used in production if installation is not completed.
i will let feedback if there is other idea (may be we can find better):
- Creating the install.lock by default (there is already an option for that, we may invert it) ?
- Disabling the ability to remove the warning saying the installation is not finished ? like suggested in #22858

altairis-tof:
> * Disabling the ability to remove the warning saying the installation is not finished ?

hello @eldy , i think that @FHenry did a PR for this : #22858

Issue #22846, opened 2022-11-15
Impossible to put notes on product


ns-info90:
### Bug

Unable to put notes on product 
the edit button (pen) is missing

### Environment Version

16.0.2

### Environment OS

Linux

### Environment PHP

7.4

### Environment Database

MariaDB

### Environment URL(s)

/product/note.php?id=

JoAnMaFe:
Still happening in 16.03 with W10

inoveaconseil:
We can't reproduce in 16.0.3

ns-info90:
I reproduced the problem in 16.0.3

ns-info90:
The correction was made in the version present in Github but not in the last version delivered in 16.0.3 
Can you release a 16.0.4?
Thanks

PR #22881, opened 2022-11-17
Fix issue #22846

lamrani002:
# FIX|Fix issue Impossible to put notes on product #22846
incoherence  "product" in place of "produit " in note.php and notes.tpl.php 
link to issue : https://github.com/Dolibarr/dolibarr/issues/22846

### DIFF
--- a/htdocs/product/note.php
+++ b/htdocs/product/note.php
@@ -54,13 +54,13 @@ $permissionnote = $user->rights->produit->creer; // Used by the include of actio
 
 if ($object->id > 0) {
 	if ($object->type == $object::TYPE_PRODUCT) {
-		restrictedArea($user, 'produit', $object->id, 'product&product', '', '');
+		restrictedArea($user, 'product', $object->id, 'product&product', '', '');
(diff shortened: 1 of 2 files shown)


htdocs/product/note.php
```diff
@@ -54,13 +54,13 @@
 
 if ($object->id > 0) {
 	if ($object->type == $object::TYPE_PRODUCT) {
-		restrictedArea($user, 'produit', $object->id, 'product&product', '', '');
+		restrictedArea($user, 'product', $object->id, 'product&product', '', '');
```
eldy, line 57: inside the code of restrictedArea, we found a line that transform "product" into "produit".
So i don't think this will change something. 
Can you reproduce the bug without the change and do you experience a fix after the change ?

lamrani002: yes you are write it hasn't changed anything , i changed because if the "else" was executed

daxitsolutions:
Tested, OK after patch on Dolibarr v16.0.3

PR #22882, opened 2022-11-17
Fix new method for fast scanlist for subselects.

fappels:
Issue on sall= and have also trouble with hook, subselects in printFieldListSelect not removed.

### DIFF
--- a/htdocs/product/list.php
+++ b/htdocs/product/list.php
@@ -595,11 +595,11 @@ $sql .= $hookmanager->resPrint;
 
 $nbtotalofrecords = '';
 if (!getDolGlobalInt('MAIN_DISABLE_FULL_SCANLIST')) {
-	if (!getDolGlobalInt('MAIN_ENABLE_FAST_SCANLIST')) {
+	if (getDolGlobalInt('MAIN_FEATURES_LEVEL') < 2) {
 		$result = $db->query($sql);
 		$nbtotalofrecords = $db->num_rows($result);
 	} else {
-		/* Experimental: The fast and low memory method to get and count full list converts the sql into a sql count */
+		/* Develop: The fast and low memory method to get and count full list converts the sql into a sql count */
 		$sqlforcount = preg_replace('/^SELECT[a-zA-Z0-9\._\s\(\),=<>\:\-\']+\sFROM/Ui', 'SELECT COUNT(*) as nbtotalofrecords FROM', $sql);
 		$sqlforcount = preg_replace('/'.preg_quote($linktopfp, '/').'/', '', $sqlforcount);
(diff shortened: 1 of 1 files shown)


htdocs/product/list.php
```diff
@@ -594,20 +594,22 @@
...
-	$sqlforcount = preg_replace('/'.preg_quote($linktopfp, '/').'/', '', $sqlforcount);
-	$sqlforcount = preg_replace('/GROUP BY .*$/', '', $sqlforcount);
-	$resql = $db->query($sqlforcount);
-	if ($resql) {
-		$objforcount = $db->fetch_object($resql);
-		$nbtotalofrecords = $objforcount->nbtotalofrecords;
+if (!getDolGlobalInt('MAIN_DISABLE_FULL_SCANLIST')) {
+	if (getDolGlobalInt('MAIN_FEATURES_LEVEL') < 2) {
+		$result = $db->query($sql);
+		$nbtotalofrecords = $db->num_rows($result);
```
eldy, line 600: This method must be abandonned. It download all the database into php memory.
Can you describe the trouble you have instead insto a ticket, and put the sql you get after this if ?

fappels: Thanks.
For the subselect in hook I have this SQL issue:
`SELECT COUNT(*) as nbtotalofrecords FROM llx_product_stock WHERE fk_product = p.rowid AND fk_entrepot = 20) as warehouse_stock FROM llx_product as p LEFT JOIN llx_product_extrafields as ef on (p.rowid = ef.fk_object) LEFT JOIN llx_product_lang as pl ON pl.fk_product = p.rowid AND pl.lang = 'fr_FR' LEFT JOIN llx_c_units cu ON cu.rowid = p.fk_unit WHERE p.entity IN (1) AND p.fk_product_type <> 1`

Part from hook which is not removed from select = `FROM llx_product_stock WHERE fk_product = p.rowid AND fk_entrepot = 20) as warehouse_stock`

htdocs/product/list.php
```diff
@@ -594,20 +594,22 @@
...
 	} else {
-		dol_print_error($db);
+		/* Develop: The fast and low memory method to get and count full list converts the sql into a sql count */
+		$sqlforcount = preg_replace('/^SELECT[a-zA-Z0-9\._\s\(\),=<>\:\-\']+\sFROM/Ui', 'SELECT COUNT(*) as nbtotalofrecords FROM', $sql);
+		$sqlforcount = preg_replace('/'.preg_quote($linktopfp, '/').'/', '', $sqlforcount);
+		$sqlforcount = preg_replace('/GROUP BY .*$/', '', $sqlforcount);
```
fappels, line 605: @eldy Can you check this?

PR #22889, opened 2022-11-17
NEW add footer hook accountancy

noec764:


eldy:
Can you send PR on develop (after solving conflict with develop)

PR #22892, opened 2022-11-17
FIX: Set extrafield value of multiselect type

kkhelifa-opendsi:
Set extrafield value of multiselect type

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -5988,6 +5988,12 @@ abstract class CommonObject
 							dol_syslog('Error bad setup of extrafield', LOG_WARNING);
 						}
 						break;
+					case 'checkbox':
+					case 'chkbxlst':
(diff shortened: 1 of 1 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -5988,6 +5988,12 @@ public function insertExtraFields($trigger = '', $userused = null)
 							dol_syslog('Error bad setup of extrafield', LOG_WARNING);
 						}
 						break;
+					case 'checkbox':
+					case 'chkbxlst':
+						if (is_array($this->array_options[$key])) {
+							$new_array_options[$key] = implode(',', $this->array_options[$key]);
```
eldy, line 5994: And what happen if it is not array($this->array_options[$key])). I think you miss the 
else
`$new_array_options[$key] = $this->array_options[$key]`

kkhelifa-opendsi:
It's fixed

Issue #22879, opened 2022-11-16
HTTP ERROR 500 when accessing Shipments -> Statistics


oerweber:
### Bug

Hello,

I have a fresh installation of Dolibarr 16.0.1 on Debian 11.5, php 7.4.30 and mariadb 10.5.15.

When I access Shipments -> Statistics (Url: http://xxx/expedition/stats/index.php?leftmenu=sendings) 

I get an 

```
This page isn’t working
xxxxxxxxxxx is currently unable to handle this request.
HTTP ERROR 500
```

### Environment Version

16.0.1

### Environment OS

Debian 11.5

### Environment Web server

nginx 1.18.0-6.1+deb11u2

### Environment PHP

7.4

### Environment Database

mariadb 10.5.15

### Environment URL(s)

http://xxx/expedition/stats/index.php?leftmenu=sendings) 

### Expected and actual behavior

Expetected:

Statistics page should show up

Actual:

This page isn’t working
xxx is currently unable to handle this request.
HTTP ERROR 500

### Steps to reproduce the behavior

Click on Menu -> Shipments -> Statistics

### Attached files

/var/log/nginx/error.log

2022/11/16 18:26:19 [error] 114530#114530: *181 FastCGI sent in stderr: "PHP message: PHP Fatal error:  **Class ExpeditionStats contains 2 abstract methods and must therefore be declared abstract or implement the remaining methods** (Stats::getAmountByMonth, Stats::getAverageByMonth) in /var/www/html/dolibarr-16.0.1/htdocs/expedition/class/expeditionstats.class.php on line 35" while reading response header from upstream, client: x.x.x.x, server: xxx, request: "GET /expedition/stats/index.php?leftmenu=sendings HTTP/1.1", upstream: "fastcgi://unix:/run/php/php7.4-fpm.sock:", host: "xxx", referrer: "http://xxx/expedition/list.php?leftmenu=sendings"

Hystepik:
Issue closed because merged

PR #22897, opened 2022-11-18
Fix issue #22879 by adding 2 methods in expedition/class/expedition.class.php

lamrani002:
# FIX|Fix  issue #22879  "HTTP ERROR 500 when accessing Shipments -> Statistics "
2 abstracts methods missed in "expeditionstats" Class, even if we don't use them it is necessary to implement them.


eldy:
I pushed a more complete fix 309a7ccf34b310aa350a4bdb87fa3ab3662d8554
The method is added to solve the error on missing methof on abstract herited class, but i also added the full implementation of code (by copy/paste of the same method into other objects)

PR #22907, opened 2022-11-20
FIX #19537 - List sorted by ref in adherents list

### LINKED ISSUE #19537: List sorted by ref in adherents module
see this post "Tri par réf dans liste adhérent"

https://www.dolibarr.fr/forum/t/tri-par-ref-dans-liste-adherent/38454
image

daraelmin:


daraelmin:
With this PR, all the "ref" fields, even in extrafields, are sorted "naturally" whether the field contains only letters, only numbers or an alphanumeric mix.
Firstly, I was unsure if this should be an optional parameter of the function or the default workflow, but since in all case ref can be alphanumeric this doesn't matter. And moreover it works in all cases with integer fields at the link of varchar. 
This should work on PostgreSQL, MariaDB and MySQL.

htdocs/core/db/DoliDB.class.php
```diff
@@ -247,7 +247,12 @@ public function order($sortfield = null, $sortorder = null)
 				if (!$return) $return .= ' ORDER BY ';
 				else $return .= ', ';
 
-				$return .= preg_replace('/[^0-9a-z_\.]/i', '', $val); // Add field
+				if (strpos($val, 'ref') !== false ){
+					$return  .= "lpad(".preg_replace('/[^0-9a-z_\.]/i', '', $val).", 30, '0')";
```
eldy, line 251: Adding a lpad function in a sql search string breaks completely the ability of database to use index .
Also when a sort is done on ref it is done using natural_search so sort should already use a natural search order.

daraelmin: Hi, 
I understand that sorting with lpad is not efficient but I canot think to a better way of sorting this new field.

Currently, the natural search order may work when looking for something (meaning using the search parameter), but the sort order is wrongly displayed since it displays:
1 - adh.rowid_1
10 - adh.rowid_10
100 - adh.rowid_100
2 - adh.rowid_2
20 - adh.rowid_20
3 - adh.rowid_3
4 - adh.rowid_4

See discussion here :
https://www.dolibarr.fr/forum/t/tri-par-ref-anormal-dans-liste-adherents/41390

I can only see three ways to sort the list properly:
1) with lpad, but this leads to a loss of efficiency
2) sort on rowid, but it only works if string ref equals integer rowid
3) add leading zeros when saving in the database (but can we do something so impactful in a bug fix?)

I'vd chose the first one cause it is not thé sort ordre by default, so the lpad fonction may not be call to often, but, for real it should be the second one witch is the best... but the ref field was created with the aim of no longer being equal to rowid so...

eldy: If order give
1 - adh.rowid_1
10 - adh.rowid_10
100 - adh.rowid_100
2 - adh.rowid_2
20 - adh.rowid_20
3 - adh.rowid_3
4 - adh.rowid_4

It means result is an alphabetical search, as expected. We want a sort on string for ref, not a sort on numeric. Don't forget that the ref is a string and will be later generated by a numbering rule like other modules. 
It is temporarily a numeric for the transition, but value will soon be MEMyymm-99999
There is no real solution for natural sorting using sql, but this need should disappear when ref numbering will be implemented.

PR #22912, opened 2022-11-21
add hook in productlot tooltip

frederic34:


htdocs/product/stock/product.php
```diff
@@ -1132,6 +1132,7 @@
 				}
 				foreach ($details as $pdluo) {
 					$product_lot_static->id = $pdluo->lotid;
+					$product_lot_static->fk_product = $pdluo->fk_product_stock;
```
eldy, line 1135: fk_product_stock is a rowid in into llx_product_stock
but fk_product is an id of product . Are you sure ?

frederic34: i remove and will check later

frederic34:
@eldy

frederic34:
@eldy

PR #22913, opened 2022-11-21
NEW : Now we can edit amount on vat and salaries clone action

atm-gauthier:


htdocs/salaries/card.php
```diff
@@ -380,6 +380,10 @@
 		$object->paye = 0;
 		$object->id = $object->ref = null;
 
+		if (GETPOST('amount', 'alphanohtml')) {
```
eldy, line 383: Must be GETPOSTISSET instead of GETPOST so we can set amount to 0

htdocs/compta/tva/card.php
```diff
@@ -333,6 +333,10 @@
 		$object->id = $object->ref = null;
 		$object->paye = 0;
 
+		if (GETPOST('amount', 'alphanohtml')) {
```
eldy, line 336: Must be GETPOSTISSET instead of GETPOST so we can set amount to 0

atm-gauthier: Hi @eldy, on vat and salaries amount is mandatory, why do you want to be able to set it to 0 ?
image

atm-gauthier: I used the same test than social charges, but in this one there is a "$object->check();" before create, then if amount is null, clone action fails

atm-gauthier: @eldy did you see my answer ?

PR #22924, opened 2022-11-22
FIX add applicant name

cfoellmann:
add applicant name to application header


htdocs/recruitment/recruitmentcandidature_agenda.php
```diff
@@ -143,6 +143,7 @@
 	$linkback = '<a href="'.dol_buildpath('/recruitment/recruitmentcandidature_list.php', 1).'?restore_lastsearch_values=1'.(!empty($socid) ? '&socid='.$socid : '').'">'.$langs->trans("BackToList").'</a>';
 
 	$morehtmlref = '<div class="refidno">';
+	$morehtmlref.= $object->getFullName('', 1);
```
eldy, line 146: If we add the full name here, we don't need the name/firstname among the table. You can disable them by setting
_'visible'=>3_
instead of 
`'visible'=>1`
for the $fields['lastname'] and $fields['firstname']

cfoellmann: I am new to what is currently allow internally for application since GDPR was introduced but the (3) makes the list look pretty empty.

cfoellmann: although name should not matter on applications finding a candidate might be harder without the name columns in the list

PR #22928, opened 2022-11-22
Fix(scrutinizer) 

KiVoilaJe:
https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/don/class/paymentdonation.class.php?selectedLabels%5B0%5D=9&selectedSeverities%5B0%5D=10&orderField=lastFound&order=desc&honorSelectedPaths=0


htdocs/don/class/paymentdonation.class.php
```diff
@@ -130,7 +130,7 @@ public function create($user, $notrigger = false)
 		$now = dol_now();
 
 		// Validate parameters
-		if (!$this->datepaid) {
+		if (!$this->datep) {
```
eldy, line 133: Line 192, we use this->datepaid.
So if we change here, we must also change on line 192

Hystepik:
@eldy The fix is good for me. I do not see datepaid use anymore

PR #22934, opened 2022-11-22
NEW introduce (get/set)DolGlobalArray()

cfoellmann:
# NEW
Introduce the functions `getDolGlobalArray()` and `setDolGlobalArray()`

use these functions to save an array to the GLOBAL. Array will be json encoded on save and json decoded on get


eldy (commented):
Use case/need of this functions seems very low and seem to replace only json_encode and json_decode.
I may reconsider the PR in future if need increase.

PR #22945, opened 2022-11-23
NEW - add hook 'llxFooter'

cfoellmann:
# NEW
add hook `llxFooter`

@eldy I have one mayor problem:
I can not get the hook to append the output from multiple modules.

I have 2 modules with these in the actions file (independently it works fine):

Actions file 1:
```php
public function llxFooter( $parameters, &$object, &$action, $hookmanager ) {

	$output = '<code><pre>';
	$output .= print_r($parameters, true);
	$output .= '</pre></code>';

	$hookmanager->resPrint = $output;

	return 0; // 0 = add / 1 = replace

} // END llxFooter()
```

Actions file 2:
```php
public function llxFooter( $parameters, &$object, &$action, $hookmanager ) {

	$hookmanager->resPrint = '<div style="width: 100%; height: 100px; background: azure; float: left;">A</div>';

	return 0; // 0 = add / 1 = replace

}
```

The HTML from both modules should be added together and output on the/all pages.
but it does not.


eldy:
Looks ok. PR is in draft mode to change.

cfoellmann:
@eldy Can you help me how to use "a hook" to keep output from all modules that use the hook?

cfoellmann:
@eldy since this is non-invasive it may get into 17.x?

cfoellmann:
can  be merged into 18.x alpha

Issue #22742, opened 2022-11-03
Include MathPHP's finance fuctions in Dolibarr


atm-florianm:
### Feature Request

Including [MathPHP](https://github.com/markrogoyski/math-php#finance), a mathematics library, to make some financial (among others) calculations more flexible and more reliable.

### Use case

I am currently working on improvements to the Loan module. It currently uses its own calculation formula to compute installments including compounded interest. As far as I can tell, the formulas are correct, but they use implicit parameters (hard-coded constants that should be parameters). The features I am working on require more flexibility in those formulas, but the current formula is not documented and therefore hard to modify without risking introducing errors.

Current implementation in Dolibarr:
```php
    function calc_mens($capital, $rate, $nbterm)
	{
		$result='';

		if (!empty($capital) && !empty($rate) && !empty($nbterm)) {
			$result = ($capital*($rate/12))/(1-pow((1+($rate/12)),($nbterm*-1)));
		}

		return $result;
	}
```
The calculation is correct (as far as my testing goes), but it will only work with an annual proportional rate for monthly installments in arrear. I am trying to enable (for instance) quarterly installments in advance (the rate will still be annual + proportional with the features I am trying to introduce, but it will be explicit to avoid any confusion).

I could implement my own calculation function, but I think there are clear benefits if we use MathPHP for those calculations:
1. very generic functions
   1. no implicit / hidden / hard-coded parameters
   2. no dependency to a specific business object, can be used anywhere in the project
   3. when possible, functions have the same name and signature as their counterpart in common spreadsheet software (Excel / LibreOffice Calc) → programmers who are already familiar with spreadsheet (personally I am not) will have less difficulty understanding code that uses those functions
2. specialized and unit tested: I believe it will do a better job than me if I implement my own formulas (I am not a trained expert in financial mathematics)
4. whenever possible, function parameters are standard PHP types (arrays, ints, floats…): no need to learn unnecessary new APIs
5. well-documented (not perfect, but it sometimes provides sources + detailed explanations)
6. feature-rich
7. not too heavy (less than 2 MB without unit tests and `.git`, 5MB if you include them)
8. if necessary we can choose not to include the whole library, but only the parts we need (for instance, the financial part (`Finance.php`), which is what I need today, is completely independant of the rest of the lib. `Finance.php` only weighs 27 kB
9. the license (MIT) is compatible with Dolibarr's GNU GPL

PR #22956, opened 2022-11-24
NEW : Library including math and financial functions

ATM-Nicolas:
# NEW library including math and financial functions
Related to the issue #22742 

These financial functions are useful for the loan module


atm-maxime:
@eldy can you tell us why this PR is postponed since november ? Can this be included in v18 ?

eldy:
It seems the PR include 378 files just to bring few mathematical functions.
This include a dependency that we can/must avoid. Why not just extracting the mathematical function we need.

Using complex solution for simple needs is not a good direction to take. To be resilient to time change, to be able to follown php and external library cycle of life, we should avoid to add external lib when we can get or build easily our own code to do the same ( or extract the small code we need from a large code base)

Trying to include o ly the Finance.php is a better direction even if this need to break the source external lib to make our own version.

PR #22961, opened 2022-11-25
FIX #22920

### LINKED ISSUE #22920: Attached files and documents in products
### Bug

If you add more as 9 image to a product under linked files, then you can't sort the attached files any more, like you wanted.

### Environment Version

16.0.2

### Environment OS

Linux

### Environment Web server

Apache

### Environment PHP

7.4

### Environment Database

Mysql

### Environment URL(s)

_No response_

_No response_

By more as 9 image, the sorting function is not working proper any more. 

### Attached files

_No response_

inoveaconseil:
Change the default value because the default sort of dir list is position_name so an int


eldy:
dol_sort_array is a generic function that is used at different place. 
The default value was set to keep compatibility. So i am not sure changing this is a good idea. Why don't you modify the caller to switch this parameter if you need it differently ?

eldy:
Ticket waiting for response since too long delay, so closed.

PR #22977, opened 2022-11-25
Fix bank statement can't use links in document tab

### LINKED ISSUE #13255: Cannot upload a bank statement / Impossible de téléverser un relevé bancaire
Dolibarr 11.0.1

Uploading a PDF bank statement does not work.
Moreover, by linking URL of the PDF bank statement, this statement is linked to all statements in Dolibarr.

-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-

Le téléversement d'un relevé bancaire PDF ne fonctionne pas.
De plus, en liant l'URL du relevé bancaire PDF, ce relevé se trouve lié à tous les relevés dans Dolibarr.

atm-maxime:
The table links allows only elements having an id. A bank statement is not a real object, so it doesn't have an id. We can't add links to bank statement for now

Fix #13255

### DIFF
--- a/htdocs/core/tpl/document_actions_post_headers.tpl.php
+++ b/htdocs/core/tpl/document_actions_post_headers.tpl.php
@@ -121,7 +121,8 @@ $formfile->form_attach_new_file(
 	$object,
 	'',
 	1,
-	$savingdocmask
+	$savingdocmask,
+	0
 );
 
 // List of document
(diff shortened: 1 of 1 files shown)


htdocs/core/tpl/document_actions_post_headers.tpl.php
```diff
@@ -147,7 +148,3 @@
 );
 
 print "<br>";
-
-//List of links
-$formfile->listOfLinks($object, $permission, $action, GETPOST('linkid', 'int'), $param);
```
eldy, line 152: You remove links on all documents pages for everybody  ?

atm-maxime:
oups

PR #22990, opened 2022-11-25
FIX#20987

### LINKED ISSUE #20987: Expedition not set as billed in database
### Bug

When an invoice is created from a shipment (expedition), the column "billed" in llx_expediton remains with a value 0 instead of 1

### Environment Version

14.0.4, 15.0.2

### Environment OS

Linux

### Environment Web server

Apache

### Environment PHP

7.4

### Environment Database

_No response_

### Environment URL(s)

_No response_

When an invoice is created from the expedition card, the column value of llx_expedition.billed should be changed to 1

Create order, create a shipment from order, create invoice from shipment.

### Attached files

_No response_

inoveaconseil:
FIX #20987

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -733,14 +733,14 @@ class Facture extends CommonInvoice
 							$this->error = $this->db->lasterror();
 							$error++;
 						}
-						else{
-							if($origin == "shipping"){
+						else {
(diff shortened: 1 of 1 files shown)


eldy:
Waiting feedback of #20987 
Feature is inter-module feature so status change may be already supported by the workflow module.

eldy:
This feature has been implemented into workflow module

PR #22993, opened 2022-11-25
NEW Bank module not enabled by default in TakePOS

andreubisquerra:
TakePOS can work without the bank module from https://github.com/Dolibarr/dolibarr/pull/21135 Many users do not use the functionality of the bank module in TakePOS. Why force all users to configure all these options? Many users get confused to configure TakePOS because many things have to be created and configured.


eldy:
Better to make setup easier than disabling bank

PR #23001, opened 2022-11-25
FIX #19400

### LINKED ISSUE #19400: The bank account specified into the third-parties for paiements is not usable into Commercial proposal pdf module
For Dolibarr 14.0.3 and olders releases:

When generating a pdf with the module pdf_azur_aster.modules.php (for example) the default bank account specified for the paiement into the third-party is never printed.

Expected:
The default bank account for the third party shoud be printed at the bottom of the PDF into "Commercial proposal" even if the default bank account is specified

The var "$object->fk_account" is null even if the bank is specified:

So i added this piece of code (after the line "if (empty($object->mode_reglement_code) || $object->mode_reglement_code == 'VIR') {"

        $sql ="select fk_account from llx_societe where rowid=".$object->socid;
        $resql=$object->db->query($sql);

        if ($resql)
        {
          $obj=$object->db->fetch_object($resql);
          if (!empty($obj->fk_account))
            $object->fk_account=$obj->fk_account;
        }

inoveaconseil:
FIX #19400

### DIFF
--- a/htdocs/core/modules/propale/doc/pdf_azur.modules.php
+++ b/htdocs/core/modules/propale/doc/pdf_azur.modules.php
@@ -1009,11 +1009,20 @@ class pdf_azur extends ModelePDFPropales
 
 			// If payment mode not forced or forced to VIR, show payment with BAN
 			if (empty($object->mode_reglement_code) || $object->mode_reglement_code == 'VIR') {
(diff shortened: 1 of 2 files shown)


eldy:
The bank account to show on a document is the bank account defined on document.
The bank account defined on thirdparty level is used to autofill the value of the document but on PDF, we must used the value defined for document and only this value. If we set no specific bank account on document, it means we don't want specific bank account on document so we not force it to another value.

PR #23002, opened 2022-11-25
FIX: update product with stock to lot or serial type.

fappels:


htdocs/expedition/card.php
```diff
@@ -63,7 +63,7 @@
 }
 
 // Load translation files required by the page
-$langs->loadLangs(array("sendings", "companies", "bills", 'deliveries', 'orders', 'stocks', 'other', 'propal'));
+$langs->loadLangs(array("sendings", "companies", "bills", 'deliveries', 'orders', 'stocks', 'other', 'propal', 'errors'));
```
eldy, line 66: Errors are "uncommon case" . We want the file error.log that will become larger and larger being loaded only if we need it instead of at each page call. Can you load it instead just before the lang->trans that need error sentence ?

fappels:
On second thought, creating default serial number not possible in real life. Also when set by accident on one million stock, will create one million serial record.

fappels:
Shifted to 18

fappels:
Some fix and shifted to 19

htdocs/product/class/product.class.php
```diff
@@ -1167,14 +1168,34 @@ public function update($id, $user, $notrigger = false, $action = 'update', $upda
 					// Quantities in batch details are not same as stock quantity,
 					// so we add a default batch record to complete and get same qty in parent and child table
 					if ($ObjW->real <> $qty_batch) {
-						$ObjBatch = new Productbatch($this->db);
-						$ObjBatch->batch = $valueforundefinedlot;
-						$ObjBatch->qty = ($ObjW->real - $qty_batch);
-						$ObjBatch->fk_product_stock = $ObjW->id;
-
-						if ($ObjBatch->create($user, 1) < 0) {
+						if ($this->status_batch == 2) {
```
eldy, line 1171: With old code, it was possible to fix a corrupted qty for product with serial. So after it was possible to record a stock correction to get a qty of 1.
If we block here, it means we can't solve a pb of corrupted data for such product ?

eldy:
> On second thought, creating default serial number not possible in real life. Also when set by accident on one million stock, will create one million serial record.

Yes but goal here is to solve a situation of corrupted data (should also not happen in real life). It is better to have consistency of data (with a bad value that we can modify after) than corrupted data (with no way to fix because if try to fix one data, the other one will be corrupted as they differs)

fappels:
@eldy I now allow update for data consistency, but show warning in stock card and stocks by lot/serial list to do a stock correction.

PR #23010, opened 2022-11-26
New rib from thirdparty

inoveaconseil:
Add RIB from thirdparty on order pdf


eldy:
The bank account to show on doc mus not be the default value of company but the value set on order level (more powerful).

The option to set the bank on order is something like BANK_ASK_PAYMENT_DURING_...

PR #23014, opened 2022-11-26
Suggestion for style tree and scss folder

kevin-reo:


htdocs/core/menus/standard/eldy.lib.php
```diff
@@ -728,20 +728,20 @@ function print_left_eldy_menu($db, $menu_array_before, $menu_array_after, &$tabM
 
 	if (is_array($moredata) && !empty($moredata['searchform'])) {	// searchform can contains select2 code or link to show old search form or link to switch on search page
 		print "\n";
-		print "<!-- Begin SearchForm -->\n";
-		print '<div id="blockvmenusearch" class="blockvmenusearch">'."\n";
+		//print "<!-- Begin SearchForm -->\n";
```
eldy, line 731: What is goal of removing this line ?

krikrak: It’s just a suggestion for clean html in production.
[Idea] > Maybe comments should only be visible in debug mod (if debug : display comments in dom; else: none;) ?

eldy: The goal of such comment is to be able when viewed in production (after a bug report for example) to quickly find the code on development sources.
They are on purpose on production to make bug fixes and impact analysis from a browser very fast.

htdocs/core/menus/standard/eldy.lib.php
```diff
@@ -87,7 +87,7 @@ function print_eldy_menu($db, $atarget, $type_user, &$tabMenu, &$menu, $noout =
 		'id' => $id,
 		'idsel' => 'home',
 		'classname' =>  $classname = (!empty($_SESSION["mainmenu"]) && $_SESSION["mainmenu"] == "home") ? 'class="tmenusel"' : 'class="tmenu"',
-		'prefix' => '<span class="fa fa-home fa-fw paddingright"></span>',
+		'prefix' => '<span class="fa fa-home fa-fw "></span>',
```
eldy, line 90: Why removing paddingright ?

htdocs/core/menus/standard/eldy.lib.php
```diff
@@ -971,13 +969,13 @@ function print_left_eldy_menu($db, $menu_array_before, $menu_array_after, &$tabM
...
-
-			print '<!-- Process menu entry with mainmenu='.$menu_array[$i]['mainmenu'].', leftmenu='.$menu_array[$i]['leftmenu'].', level='.$menu_array[$i]['level'].' enabled='.$menu_array[$i]['enabled'].', position='.$menu_array[$i]['position'].' -->'."\n";
+			// For debug only ? (add global option) ?
+			//print '<!-- Process menu entry with mainmenu='.$menu_array[$i]['mainmenu'].', leftmenu='.$menu_array[$i]['leftmenu'].', level='.$menu_array[$i]['level'].' enabled='.$menu_array[$i]['enabled'].', position='.$menu_array[$i]['position'].' -->'."\n";
 
 			// Menu level 0
 			if ($menu_array[$i]['level'] == 0) {
 				if ($menu_array[$i]['enabled']) {     // Enabled so visible
-					print '<div class="menu_titre">'.$tabstring;
+					print '<li class="menu_titre">'.$tabstring;
```
eldy, line 978: Using li instead of div for menu is a good idea.
However, we must follow the rule 1 PR = 1 feature purpose. Can you make a new PR with only this goal in mind into a separate atomic PR.

htdocs/theme/eldy/global.inc.php
```diff
@@ -3096,23 +3079,27 @@
...
 
 div.vmenu, td.vmenu {
-	margin-<?php print $right; ?>: 2px;
 	position: relative;
 	float: left;
-	padding: 0px;
-	padding-bottom: 0px;
-	padding-top: 1px;
-	width: 190px;
+	width: 200px;
```
eldy, line 3084: If you made some changes into CSS, does it change visual effect ? If yes, can you provide a screenshot of after (on desktop and smartphone) ?

krikrak: Okay I see, thanks!
I am working on a taskboard to reference IHM integration method,
let me try something and I’ll come back to you :)

htdocs/core/menus/standard/eldy.lib.php
krikrak: Yes, I add new scss folder for isolate "section/layout/objects" styles. 
I create the scss folder for test new structure and new object segmentation (wip:)
_In this case, the icon is a part of the object named "list-item", at least that is the future intention..._
> what do you think ?
> I didn't ask you, is there a working folder to compile eldy style ?

PR #23025, opened 2022-11-28
Fix issue #23023 

### LINKED ISSUE #23023: 'off' parameters are erased during a module update
### Bug

When disabling, then enabling a module (for example during an update), the on/off constants set on 'off' by the user are re-set on 'on' if it is the default parameter.

### Environment Version

tested on v14 and v16.

### Environment OS

any

### Environment Web server

any

### Environment PHP

any

### Environment Database

any

### Environment URL(s)

_No response_

current behavior: When disabling, then enabling a module (for example during an update), the on/off constants that were previously set to 'off' by the user are re-set on 'on' if 'on' is the default parameter.

expected behavior: if a parameter has been set to 'OFF' by a user, it should stay on 'OFF' when re-enabling the module, even if the default parameter is 'ON'.

This behavior can be found on WYSIWYG editor, but **it will also happen on almost any internal or external module where a default value is set to 1**.

1. On a fresh install, we enable WYSIWYG editor. We have the **default** setup:
image

2. we disable one parameter:
image

3. We disable/enable the module
4. The modified parameter has been overridden:
image

___

Explanations
-----------------

This behavior is due to using `dolibarr_del_const()` when setting a parameter to 'OFF'. The const is deleted from table llx_const. When re-enabling the module, an insertion is done because the const does not appear in the table.

When setting a parameter to 'OFF', the const value should be set to '0'.

On most modules (ex. WYSIWYG editor) the fix must be made on the admin page.
If using ajax_constantonoff, the fix must also be made on core/ajax/constantonoff.php file.

### Attached files

_No response_

lamrani002:
# FIX|Fix #23023 issue erase values during a module update 
replace dolibarr_del_const() by dolibarr_set_const() to maintain the setting of module before disable.

### DIFF
--- a/htdocs/admin/fckeditor.php
+++ b/htdocs/admin/fckeditor.php
@@ -100,7 +100,7 @@ foreach ($modules as $const => $desc) {
 		exit;
 	}
 	if ($action == 'disable_'.strtolower($const)) {
-		dolibarr_set_const($db, "FCKEDITOR_ENABLE_".$const, 0);
+		dolibarr_set_const($db, "FCKEDITOR_ENABLE_".$const, "0", 'chaine', 0, '', $conf->entity);
 		header("Location: ".$_SERVER["PHP_SELF"]);
 		exit;
 	}
--- a/htdocs/admin/fckeditor.php
+++ b/htdocs/admin/fckeditor.php
@@ -100,7 +100,7 @@ foreach ($modules as $const => $desc) {
 		exit;
 	}
 	if ($action == 'disable_'.strtolower($const)) {
-		dolibarr_del_const($db, "FCKEDITOR_ENABLE_".$const, $conf->entity);
+		dolibarr_set_const($db, "FCKEDITOR_ENABLE_".$const, 0);
 		header("Location: ".$_SERVER["PHP_SELF"]);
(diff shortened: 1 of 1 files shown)


htdocs/admin/fckeditor.php
```diff
@@ -100,7 +100,7 @@
 		exit;
 	}
 	if ($action == 'disable_'.strtolower($const)) {
-		dolibarr_del_const($db, "FCKEDITOR_ENABLE_".$const, $conf->entity);
+		dolibarr_set_const($db, "FCKEDITOR_ENABLE_".$const, 0);
```
eldy, line 103: You must keep the parameter $conf->entity into the dolibarr_set_const
If not, the constant will always be saved into the main company (this is used by the multicompany module to use 1 dolibarr to manage n companies).

lamrani002: yes absolutely , i fix it

thomas-Ngr:
As the author of the related issue, I must point out that this PR fixes the issue only partially (that is, only for WYSYWIG editor that was used as an _example_). It is a good thing to start with, however it should not close the issue. I plan to fix this further.

PR #23037, opened 2022-11-29
Fix(scrutinizer) paymentexpensereport

KiVoilaJe:
https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/expensereport/class/paymentexpensereport.class.php?selectedLabels%5B0%5D=9&selectedSeverities%5B0%5D=10&orderField=lastFound&order=desc&honorSelectedPaths=0

### DIFF
--- a/htdocs/expensereport/class/paymentexpensereport.class.php
+++ b/htdocs/expensereport/class/paymentexpensereport.class.php
@@ -113,7 +113,7 @@ class PaymentExpenseReport extends CommonObject
 		$now = dol_now();
 
 		// Validate parameters
-		if (!$this->datepaid) {
+		if (!$this->datep) {
 			$this->error = 'ErrorBadValueForParameterCreatePaymentExpenseReport';
 			return -1;
 		}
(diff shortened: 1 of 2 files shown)


htdocs/expensereport/class/paymentexpensereport.class.php
```diff
@@ -113,7 +113,7 @@ public function create($user)
 		$now = dol_now();
 
 		// Validate parameters
-		if (!$this->datepaid) {
+		if (!$this->datep) {
```
eldy, line 116: If a make sa search on ->datepaid in dir expensereport i see found a 
$payment->datepaid       = $datepaid;
just before a 
$payment->create()
This means the property ->datepaid is set before calling the create, so if you change the test here, you will probably get an error.

There is surely a problem with datepaid / datep   (both my be used). But of you change at some place, you must also analyse who use the modified code and check if code after the change still works. Can you check this (by making an expense report and then creating a payment) to validate your change does not break something ?

Hystepik:
@eldy The fix is good for me. I do not see datepaid use anymore

xbloq:
I can't pay donations V17 I have a php error with htdocs/don/class/paymentdonation.class.php

Undefined property: PaymentDonation::$datepaid in /var/www/html/dolibarr/htdocs/don/class/paymentdonation.class.php on line 133

changing (3) datepaid by datep in  paymentdonation.class.php I can paid the donation. I don't see any error after the change

PR #23038, opened 2022-11-29
Fix(Scrutinizer) payout

KiVoilaJe:
https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/stripe/payout.php?selectedLabels%5B0%5D=9&selectedSeverities%5B0%5D=10&orderField=lastFound&order=desc&honorSelectedPaths=0

### DIFF
--- a/htdocs/societe/class/societe.class.php
+++ b/htdocs/societe/class/societe.class.php
@@ -784,16 +784,6 @@ class Societe extends CommonObject
 	 */
 	public $multicurrency_code;
 
-	/**
-	 * @var string login
-	 */
-	public $login;
-
-	/**
-	 * @var string admin
-	 */
-	public $admin;
-
 	// Fields loaded by fetchPartnerships()
 
 	public $partnerships = array();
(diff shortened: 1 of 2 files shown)


htdocs/societe/class/societe.class.php
```diff
@@ -784,6 +784,15 @@ class Societe extends CommonObject
 	 */
 	public $multicurrency_code;
 
+	/**
+	 * @var string login
+	 */
+	public $login;
```
eldy, line 790: There is no reason to have a field ->login on a company. This is not into database.
Same for property ->admin

htdocs/stripe/payout.php
```diff
@@ -154,15 +154,8 @@
 			{
 				$memberstatic->id = 0;
 			}*/
-
 			$societestatic->fetch($charge->metadata->idcustomer);
 			$societestatic->id = $charge->metadata->idcustomer;
```
eldy, line 158: Just before, we have 
$societestatic->fetch()
It means object is loaded with all its property including the ->id set to $charge->metadata->idcustomer that is the id received as parameter to make the fetch.
So you can also remove this line   $societestatic->id = $charge->metadata->idcustomer;

KiVoilaJe:
Tout a déjà étaient fix

PR #23040, opened 2022-11-29
Fix(scrutinizer) list

KiVoilaJe:
https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/compta/facture/list.php?selectedLabels%5B0%5D=9&selectedSeverities%5B0%5D=10&orderField=lastFound&order=desc&honorSelectedPaths=0

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -171,6 +171,8 @@ class Facture extends CommonInvoice
 	public $total_ttc;
 	public $revenuestamp;
 
+	public $resteapayer;
+
 	/**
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/class/facture.class.php
```diff
@@ -199,6 +201,7 @@ class Facture extends CommonInvoice
 	public $fk_facture_source;
 	public $linked_objects = array();
 
+	public $date_valid;
```
eldy, line 204: There is already a field ->date_validation that is used frequently.
So i think problem is where the code is using ->date_valid (should be date_validation) and not in the missing property.

PR #23043, opened 2022-11-29
Fix #22494 Bad error message and trans missing (radio button)

### LINKED ISSUE #22494: Radio button empty value gives ErrorBadValueForParamNotAString
### Bug

I try to make a Radio button in "Attributs supplémentaries (commandes)" but when there is no value chosen this message appear: "ErrorBadValueForParamNotAString"

### Environment Version

16

### Environment OS

Debian 11

### Environment Web server

Apache

### Environment PHP

7.4.30

### Environment Database

Maria 10.5.15

### Environment URL(s)

_No response_

When making a Radio button list in "Attributs supplémentaries (commandes)" it is not possible to leave an empty value as it gives the message "ErrorBadValueForParamNotAString" both in the listing and on the card, as can bee seen on the 2 attached images.

I assume it should be possible to leave an empty value?

The Radio button list looks like this

1,Reçu
2,Confirmé
3,Refusé

I've temporarly solved it by checking Required and added one more row like this

1,Pas d'réponse
2,Reçu
3,Confirmé
4,Refusé

_No response_

### Attached files

listing
card

ibuiv:
# Fix #22494 Bad error message and trans missing (radio button)

On commande/card if an radio button extrafield has a bad default value (or none), the displayed error message was "ErrorBadValueForParamNotAString"
It is now the translation for "ErrorNoValueForRadioType" (Example : Les valeurs de la liste d'options doivent être renseignées in French version)

### DIFF
--- a/htdocs/commande/card.php
+++ b/htdocs/commande/card.php
@@ -14,7 +14,6 @@
  * Copyright (C) 2015       Jean-François Ferry		<jfefe@aternatik.fr>
  * Copyright (C) 2018-2021  Frédéric France         <frederic.france@netlogic.fr>
  * Copyright (C) 2022	    Gauthier VERDOL     <gauthier.verdol@atm-consulting.fr>
- * Copyright (C) 2022	    Antonin MARCHAL     	<antonin@letempledujeu.fr>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


htdocs/commande/card.php
```diff
@@ -60,7 +61,7 @@
 }
 
 // Load translation files required by the page
-$langs->loadLangs(array('orders', 'sendings', 'companies', 'bills', 'propal', 'deliveries', 'products', 'other'));
+$langs->loadLangs(array('orders', 'sendings', 'companies', 'bills', 'propal', 'deliveries', 'products', 'other', 'errors'));
```
eldy, line 64: The goal of the errors lang file is to contains error message that are loaded only when required.
Can you move the load of lang file
$langs->load("errors");
just before the code that need the error keys, so before the line 
$value = $langs->trans('ErrorNoValueForRadioType');

ibuiv: ok done

PR #23046, opened 2022-11-29
FIX add "cats_by_id" var to not loose category id in array key

hregis:


htdocs/categories/class/categorie.class.php
```diff
@@ -1176,22 +1188,24 @@ public function get_full_arbo($type, $markafterid = 0, $include = 0)
...
-					unset($this->cats[$key]);
+					unset($categories[$key]);
 				}
 			}
 		}
 
 		dol_syslog(get_class($this)."::get_full_arbo dol_sort_array", LOG_DEBUG);
-		$this->cats = dol_sort_array($this->cats, 'fulllabel', 'asc', true, false);
+		$this->cats_sortby_id = $categories;
+		$this->cats_sortby_fulllabel =  dol_sort_array($categories, 'fulllabel', 'asc', true, false);
```
eldy, line 1203: Why storing the same array into 2 different variables ? You can retreive at any time the second one doing a dol_sort_array on 'fulllabel',  'id' or any other field of your choice.

hregis: @eldy yes but if you use the return of "get_full_arbo" the value is already "order by fulllabel" with "dol_sort_array()" function, but the "key" of array before "dol_sort_array()" is ID of category, and after is an incremental (0,1,2,...)
it's not possible to compare` $this->cats[$catory_id]` !!!! The "key" is wrong ! ;-)

hregis: @eldy it's not the same array !!

eldy: After call of this function, you get an array $var->cats
The caller can sort it if he wants with
dol_sort_array($var->cats, 'fulllabel', 'asc', true, false);
To have it sorted on id
dol_sort_array($var->cats, 'id', 'asc', true, false);

Yes, the id of entry is not category id but a simple index, but you got the property id into ->cats['id']. So to avoid to do the sort each time 2 times for each call, even if we don't need it, it is better to have this done by the caller (the caller can also retreive the id), and keep only 1 sort that is the sort required for the purpose of the method: getting a tree sorted according to its branches.

PR #23047, opened 2022-11-29
Fix(scrutinizer) Info

KiVoilaJe:
https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/loan/info.php?selectedLabels%5B0%5D=9&selectedSeverities%5B0%5D=10&orderField=lastFound&order=desc&honorSelectedPaths=0

### DIFF
--- a/htdocs/loan/class/loan.class.php
+++ b/htdocs/loan/class/loan.class.php
@@ -66,6 +66,7 @@ class Loan extends CommonObject
 	public $account_capital;
 	public $account_insurance;
 	public $account_interest;
+	public $total_paid;
 
 	/**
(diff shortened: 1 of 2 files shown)


htdocs/loan/class/loan.class.php
```diff
@@ -66,6 +66,7 @@ class Loan extends CommonObject
 	public $account_capital;
 	public $account_insurance;
 	public $account_interest;
+	public $total_paid;
```
eldy, line 69: This field does not exists into database so no reason to have it into class.
Poblem may be in parent code that use it, probably useless.

PR #23048, opened 2022-11-29
Fix(scrutinizer) Price

KiVoilaJe:
https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/societe/price.php?selectedLabels%5B0%5D=9&selectedSeverities%5B0%5D=10&orderField=lastFound&order=desc&honorSelectedPaths=0


htdocs/societe/class/societe.class.php
```diff
@@ -478,6 +478,12 @@ class Societe extends CommonObject
 	 */
 	public $tva_intra;
 
+	public $tva_npr;
```
eldy, line 481: There is no reason to have such property on the societe class. There is no such fields in database.

Hystepik:
@eldy ok for me

PR #23049, opened 2022-11-29
Fix(scrutinizer) Contact

KiVoilaJe:
https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/resource/contact.php?selectedLabels%5B0%5D=9&selectedSeverities%5B0%5D=10&orderField=lastFound&order=desc&honorSelectedPaths=0

### DIFF
--- a/htdocs/resource/class/dolresource.class.php
+++ b/htdocs/resource/class/dolresource.class.php
@@ -52,11 +52,6 @@ class Dolresource extends CommonObject
 	public $mandatory;
 
 	/**
-	 * @var integer
-	 */
-	public $socid;
-
-	/**
 	 * @var int ID
 	 */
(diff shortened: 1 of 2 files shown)


htdocs/resource/class/dolresource.class.php
```diff
@@ -51,6 +51,11 @@ class Dolresource extends CommonObject
 	public $busy;
 	public $mandatory;
 
+	/**
+	 * @var integer
+	 */
+	public $socid;
```
eldy, line 57: If i look at the record of a dolresource, i don't see field company (socid).
So i don't think the problem is that the declaration is missing. Problem may be on the other side where the field is used and where it should not.

image

PR #23050, opened 2022-11-30
Fix user off constants not saved

### LINKED ISSUE #23023: 'off' parameters are erased during a module update

thomas-Ngr:
# FIX #23023 : module parameters set to OFF are deleted
On module config, when user sets parameters to OFF, the related constants are deleted from DB through dolibarr_del_const(). When re-enabling a module, if the module has default parameters set to 'ON', these parameters override the users choice.
We should save OFF parameters as set to 0 instead of deleting the llx_const line.

(Started in develop on PR https://github.com/Dolibarr/dolibarr/pull/23025)

### DIFF
--- a/htdocs/admin/agenda.php
+++ b/htdocs/admin/agenda.php
@@ -90,7 +90,7 @@ if ($action == "save" && empty($cancel)) {
 		$keyparam = 'MAIN_AGENDA_ACTIONAUTO_'.$trigger['code'];
 		//print "param=".$param." - ".$_POST[$param];
 		if ($search_event === '' || preg_match('/'.preg_quote($search_event, '/').'/i', $keyparam)) {
-			$res = dolibarr_set_const($db, $keyparam, (GETPOST($keyparam, 'alpha') ?GETPOST($keyparam, 'alpha') : ''), 'chaine', 0, '', $conf->entity);
+			$res = dolibarr_set_const($db, $keyparam, (GETPOST($keyparam, 'alpha') ?GETPOST($keyparam, 'alpha') : '0'), 'chaine', 0, '', $conf->entity);
 			if (!($res > 0)) {
 				$error++;
 			}
--- a/htdocs/admin/confexped.php
+++ b/htdocs/admin/confexped.php
@@ -70,7 +70,7 @@ if ($action == 'activate_delivery') {
 	header("Location: confexped.php");
 	exit;
 } elseif ($action == 'disable_delivery') {
-	dolibarr_del_const($db, "MAIN_SUBMODULE_DELIVERY", $conf->entity);
+	dolibarr_set_const($db, "MAIN_SUBMODULE_DELIVERY", 0, 'chaine', 0, '', $conf->entity);
 	header("Location: confexped.php");
 	exit;
 }
--- a/htdocs/admin/confexped.php
+++ b/htdocs/admin/confexped.php
@@ -70,7 +70,7 @@ if ($action == 'activate_delivery') {
 	header("Location: confexped.php");
 	exit;
 } elseif ($action == 'disable_delivery') {
-	dolibarr_set_const($db, "MAIN_SUBMODULE_DELIVERY", 0, 'chaine', 0, '', $conf->entity);
+	dolibarr_del_const($db, "MAIN_SUBMODULE_DELIVERY", $conf->entity);
 	header("Location: confexped.php");
 	exit;
 }
--- a/htdocs/admin/ecm.php
+++ b/htdocs/admin/ecm.php
@@ -49,7 +49,7 @@ if (preg_match('/set_([a-z0-9_\-]+)/i', $action, $reg)) {
 
 if (preg_match('/del_([a-z0-9_\-]+)/i', $action, $reg)) {
 	$code = $reg[1];
-	if (dolibarr_del_const($db, $code, $conf->entity) > 0) {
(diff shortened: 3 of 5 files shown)


eldy:
In most cases, when a parameter is set to 0, we want to have it removed complete from llx_const.
Because framework say  if not found, it means 0.
We need to keep the entry to 0 only for very rare case that are not correctly implemented because a default value is set during the module activation instead of being defined into  conf.class.php

htdocs/core/ajax/constantonoff.php
```diff
@@ -71,7 +71,8 @@
 		if ($action == 'set') {
 			dolibarr_set_const($db, $name, $value, 'chaine', 0, '', $entity);
 		} elseif ($action == 'del') {
-			dolibarr_del_const($db, $name, $entity);
+			// We should save the constant as 0 instead of deleting the constant.
+			dolibarr_set_const($db, $name, 0, 'chaine', 0, '', $entity);
```
thomas-Ngr, line 75: I keep setting the value to '0' here because:
- lots of external modules will use this script
- the core module stock.php has a default const STOCK_ALLOW_NEGATIVE_TRANSFER set on 1, and uses ajax/constantonoff.php for this const.

htdocs/admin/agenda.php
```diff
@@ -90,7 +90,7 @@
 		$keyparam = 'MAIN_AGENDA_ACTIONAUTO_'.$trigger['code'];
 		//print "param=".$param." - ".$_POST[$param];
 		if ($search_event === '' || preg_match('/'.preg_quote($search_event, '/').'/i', $keyparam)) {
-			$res = dolibarr_set_const($db, $keyparam, (GETPOST($keyparam, 'alpha') ?GETPOST($keyparam, 'alpha') : ''), 'chaine', 0, '', $conf->entity);
+			$res = dolibarr_set_const($db, $keyparam, (GETPOST($keyparam, 'alpha') ?GETPOST($keyparam, 'alpha') : '0'), 'chaine', 0, '', $conf->entity);
```
thomas-Ngr, line 93: Here we have to set the const to '0' instead of '', because dolibarr_set_const will delete the const if it is ''.

thomas-Ngr:
Hi @eldy, I have only kept these changes on modules where there are default parameters set to 1.

thomas-Ngr:
Hi @eldy, is there any other issue this this PR ?

htdocs/admin/confexped.php
```diff
@@ -70,7 +70,7 @@
 	header("Location: confexped.php");
 	exit;
 } elseif ($action == 'disable_delivery') {
-	dolibarr_del_const($db, "MAIN_SUBMODULE_DELIVERY", $conf->entity);
+	dolibarr_set_const($db, "MAIN_SUBMODULE_DELIVERY", 0, 'chaine', 0, '', $conf->entity);
```
eldy, line 73: I see no reason to not remive this const completely

htdocs/admin/ecm.php
```diff
@@ -49,7 +49,7 @@
 
 if (preg_match('/del_([a-z0-9_\-]+)/i', $action, $reg)) {
 	$code = $reg[1];
-	if (dolibarr_del_const($db, $code, $conf->entity) > 0) {
```
eldy, line 52: Here $code was ECM_AUTO_TREE_ENABLED.
But i pushed recenty a pr to revert the option in ECM_AUTO_TREE_HIDDEN so now the iption is no more created during module activation (we must avoid this).
So we don't need anymore to replace this dol_delete with dol_set....

thomas-Ngr: >  i pushed recenty a pr to revert the option in ECM_AUTO_TREE_HIDDEN

You are right, you did on `develop` branch . 
This fix points to 14.0 branch. Do you want me to switch to `develop` ? There is no problem on my side.

htdocs/admin/confexped.php
thomas-Ngr: You are right, I reverted this commit.

thomas-Ngr: You are right, I reverted the related commit.

thomas-Ngr:
Pr done to branch develop :  #23769
Closing this one.

PR #23051, opened 2022-11-30
FIX missing type of category during create action (for hook)

hregis:


htdocs/categories/card.php
```diff
@@ -81,6 +81,7 @@
 }
 
 $object = new Categorie($db);
+$object->type = $type; // for hooks
```
eldy, line 84: This line seems already present at line 140

eldy:
This bug or the feature request seems to be fixed or included into last recent stable version, so I close the issue.

PR #23055, opened 2022-11-30
Jyhere patch showoptionals hook params

jyhere:
Allows to use some params like colspan in the showOptionals hook.


eldy:
I pushed a more complete change so other params are also propagated.

PR #23066, opened 2022-12-01
FIX: propal and order stats broken on Tag+User(retricted customer list)

FHenry:
If user is restricted to his/her customer (user perms) on propal stats (comm/propal/stats/index.php) and Order stat (commande/stats/index.php) 
Stats filterd by Tags, User created, Status and Year give SQL error

`Modules/Applications: syslog, user, export, import, stock, workflow, concatpdf, categorie, bookmark, listincsv, projet, notification, blockedlog, propal, expedition, tax, product, definednotes, fckeditor, accounting, agenda, banque, cron, commande, expensereport, facture, fournisseur, holiday, margin, salaries, service, societe, attachments
Type gestionnaire de base de données: mysqli
Requête dernier accès en base en erreur: SELECT date_format(p.datep,'%m') as dm, COUNT(*) as nb FROM llx_propal as p, llx_societe_commerciaux as sc LEFT JOIN llx_categorie_societe as cs ON cs.fk_soc = p.fk_soc LEFT JOIN llx_categorie as c ON c.rowid = cs.fk_categorie WHERE p.datep BETWEEN '2020-01-01 00:00:00' AND '2020-12-31 23:59:59' AND p.entity IN (1) AND p.fk_soc = sc.fk_soc AND sc.fk_user = 4 AND fk_user_author = 4 AND c.rowid = 4 AND p.fk_statut IN (2) GROUP BY dm ORDER BY dm DESC
Code retour dernier accès en base en erreur: DB_ERROR_NOSUCHFIELD
Information sur le dernier accès en base en erreur: Unknown column 'p.fk_soc' in 'on clause'`

SQL Query must be re-write with inner join rather than ","+where

### DIFF
--- a/htdocs/comm/propal/class/propalestats.class.php
+++ b/htdocs/comm/propal/class/propalestats.class.php
@@ -93,9 +93,6 @@ class PropaleStats extends Stats
 		}
 		//$this->where.= " AND p.fk_soc = s.rowid AND p.entity = ".$conf->entity;
 		$this->where .= ($this->where ? ' AND ' : '')."p.entity IN (".getEntity('propal').")";
-		if (!$user->rights->societe->client->voir && !$this->socid) {
-			$this->where .= " AND p.fk_soc = sc.fk_soc AND sc.fk_user = ".((int) $user->id);
-		}
 		if ($this->socid) {
 			$this->where .= " AND p.fk_soc = ".((int) $this->socid);
 		}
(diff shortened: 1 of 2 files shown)


eldy:
Good direction is to do like we did for the page using nothing into the from and using a where EXISTS in condition. But this will be implemented in develop.

PR #23067, opened 2022-12-01
NEW filter on entity in import

lvessiller-opendsi:
NEW filter on entity in import
- add a SQL filter for entity in all import from CSV or XLSX file

### DIFF
--- a/htdocs/core/modules/import/import_xlsx.modules.php
+++ b/htdocs/core/modules/import/import_xlsx.modules.php
@@ -916,6 +916,10 @@ class ImportXlsx extends ModeleImports
 										$filters[] = $col.' = '.$data[$key];
 									}
 								}
+								if (!empty($tablewithentity_cache[$tablename])) {
+									$where[] = "entity = ".((int) $conf->entity);
+									$filters[] = "entity = ".((int) $conf->entity);
+								}
 								$sqlSelect .= " WHERE " . implode(' AND ', $where);
 
 								$resql = $this->db->query($sqlSelect);
@@ -952,6 +956,10 @@ class ImportXlsx extends ModeleImports
 									$keyfield = 'rowid';
 								}
 								$sqlSelect .= " WHERE ".$keyfield." = ".((int) $lastinsertid);
+
+								if (!empty($tablewithentity_cache[$tablename])) {
+									$sqlSelect .= " AND entity = ".((int) $conf->entity);
+								}
 
 								$resql = $this->db->query($sqlSelect);
 								if ($resql) {
@@ -996,6 +1004,10 @@ class ImportXlsx extends ModeleImports
 
 								if ($is_table_category_link) {
 									$sqlend = " WHERE " . implode(' AND ', $where);
+								}
+
+								if (!empty($tablewithentity_cache[$tablename])) {
+									$sqlend .= " AND entity = ".((int) $conf->entity);
 								}
 
 								$sql = $sqlstart . $sqlend;
--- a/htdocs/core/modules/import/import_xlsx.modules.php
+++ b/htdocs/core/modules/import/import_xlsx.modules.php
@@ -106,6 +106,8 @@ class ImportXlsx extends ModeleImports
 	public function __construct($db, $datatoimport)
 	{
 		global $conf, $langs;
+
+		parent::__construct();
 		$this->db = $db;
 
 		// this is used as an extension from the example file code, so we have to put xlsx here !!!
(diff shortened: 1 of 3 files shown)


eldy:
Currently import is done on entity where you are when making import.
Does this means you can, when on entity 2, import some data that are inserted in entity 1 ?

lvessiller-opendsi:
> Currently import si done on entity where you are when making import. Does this means you can, when on entity 2, import some data that are inserted in entity 1 ?

No it was a filter on SQL requests only when we update data on multi-entity. 
I will check if it's the good way to do when multi-entity is enabled.

lvessiller-opendsi:
@eldy 
To reproduce this problem : 
1) Create a product with ref "PRODMULTI1" in entity A with a description "Product for entity A"
2) Switch on entity B : there is no products
3) Try to import a product with reference is also "PRODMULTI1" and a description "Product for entity B" and set "ref" as update key this import
4) When you import on entity B, you have an update instead of having an insert
**So when you come back to the entity A you will see the description "Product for entity B" for product whose ref is "PRODMULTI1"**

This PR will fix this last behaviour. 
Nevertheless, how to do if products or other objects are shared between entities ?

lvessiller-opendsi:
@eldy 
It also works well with other shared objects between entities

htdocs/core/modules/import/import_xlsx.modules.php
```diff
@@ -916,6 +916,10 @@ public function import_insert($arrayrecord, $array_match_file_to_database, $obji
 										$filters[] = $col.' = '.$data[$key];
 									}
 								}
+								if (!empty($tablewithentity_cache[$tablename])) {
+									$where[] = "entity = ".((int) $conf->entity);
```
eldy, line 920: Ok, i see.
So I suggest to replace this line an dline 921 with
`$where[] = "entity IN (".getEntity($tablename).")";`
So when on session with entity 2 and 3 shared, if we find product PRODMULTI on 2 when on 3, update will work.
And if there is 2 products, one in 2 and same with same ref in 3, we will found 2 record and it will return the expected error (line 933) saying we don't know which record to update.

lvessiller-opendsi: @eldy 
Are you sure that it's always getEntity($tablename) with $tablename as parameter ?
In getEntity() method the first parameter is the object element and not table element.
And what about lines 961 and 1010 ? Should I also use "IN getEntity($tablename)" instead of "=" ?

hregis: @eldy @lvessiller-opendsi 
this is not the "tablename" but this is the "element" name

lvessiller-opendsi: @eldy @hregis 
But how can I get the element name from table name ? (element is not defined in the declaration of import)
The getEntity with $tablename can solve the most of cases.

lvessiller-opendsi: @hregis 
But how can I get the element name from table name ?

hregis: @lvessiller-opendsi @eldy 
the value of $tablename would be for example "llx_product"? 
and "getEntity()" does not handle table names.

So either we add the name of the element in the description of the import, eg:

```
$this->import_tables_array[$r] = array('p'=>MAIN_DB_PREFIX.'product');
$this->import_element_array[$r] = array('p'=>'product');

$this->import_tables_array[$r] = array('s'=>MAIN_DB_PREFIX.'societe');
$this->import_element_array[$r] = array('s'=>'thirdparty');
```

or I add a correspondence in the Multicompany module?

`array('llx_product' => 'product', 'llx_societe' => 'thirdparty');`

lvessiller-opendsi: @hregis @eldy 
I added a list to map table to element in model import class.
So I use this map list to filter with getEntity() method in import.
Could you check if it's a good way to implement it ?
I think I put all shareable tables for multi-company : could you check map list ? (hregis)

hregis: @lvessiller-opendsi 
it's possible to call you by phone ? send me an email please, thanks

lvessiller-opendsi: @hregis 
Yes, I sent you an email with my phone number

hregis: @eldy it's ok for me, thanks

htdocs/core/modules/import/modules_import.php
```diff
@@ -75,15 +75,54 @@ class ModeleImports
...
+	 * @var	array	Element mapping from table name
+	 */
+	public static $mapTableToElement = array(
+		'actioncomm' => 'agenda',
+		'adherent' => 'member',
+		'adherent_type' => 'member_type',
+		'bank_account' => 'bank_account',
+		'categorie' => 'category',
+		'commande' => 'commande',
+		'commande_fournisseur' => 'commande_fournisseur',
```
eldy, line 88: Do we need record with key that has same value than value in this table ? My understanding is that this array is used to be able to get element when it is not same than table_element ? So it is a table to manage exceptions in bad old naming exception. if i'm right, can't we remove entries with key = value ?

lvessiller-opendsi: @eldy
I think you're right and I removed all keys are the same as table element.
And I modified method "getElementFromTableWithPrefix" to retrun "table element" as default and the mapped element if set.

lvessiller-opendsi:
@eldy 
What about this PR ?

kkhelifa-opendsi:
@eldy
What about this PR ?

PR #23070, opened 2022-12-01
NEW substitute date now in email collector

lvessiller-opendsi:
NEW substitute date now in email collector

### DIFF
--- a/htdocs/emailcollector/class/emailcollector.class.php
+++ b/htdocs/emailcollector/class/emailcollector.class.php
@@ -989,6 +989,7 @@ class EmailCollector extends CommonObject
 								}
 							}
 						}
+						$substitutionarray['__DATE_NOW__'] = dol_now();
 						//var_dump($substitutionarray);
 						dol_syslog(var_export($substitutionarray, true));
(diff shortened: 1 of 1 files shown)


htdocs/emailcollector/class/emailcollector.class.php
```diff
@@ -989,6 +989,7 @@ private function overwritePropertiesOfObject(&$object, $actionparam, $messagetex
 								}
 							}
 						}
+						$substitutionarray['__DATE_NOW__'] = dol_now();
```
eldy, line 992: Why not calling getCommonSubstitutionArray() so the substitution array will contains all other entries. The dol_now will be available with __NOW_TMS__

lvessiller-opendsi: Yes I will do as you suggest to me

lvessiller-opendsi: @eldy 
I used getCommonSubstitutionArray function as you suggested and removed __DATE_NOW__ substitute key.

Issue #23080, opened 2022-12-02
Improvement on signature WorkFlow


lmag:
#  Add functions

- [x] Add signatures to SEPA
- [ ] Add signatures to orders https://github.com/Dolibarr/dolibarr/issues/28871
- [ ] Signature on deliveries
- [ ] Add the function of deleting signatures automatically after generation of pdf
- [ ] Print IP of signature on document (option)

# DataModel improvement : onlinesignature.sql
Version 20.X missing fields
- [ ] SHA of document
- [x] IP
- [ ] Migrate (online_sign_ip, online_sign_name) to onlinesignature
- [ ] Keep place of signature by document

# CRUD Improvement
- [ ] Do CRUD
- [ ] Add Extrafield on this table

# Document modifications should
- [ ] Delete signature after document modification
- [ ] Delete information of IP sign

rycks:
Suggestion : make a list of "what is what" ... that is my main problem for the moment with uptosign !

For example:
- humain draw "like paper signature" :
- document timestamp & seal with eidas conformance certificate : 
- document sha fingerprint :
- .../...

Just because i don't really know what "Add signatures to SEPA" means for example.

In uptosign i just have only two concepts :
- seal is for timesamp & certificate seal on a document.
- sign is for a document where people draw a human "paper sign" + SMS check + at the end of the process global document seal

but even if there is only two concept i need to store more informations, i call it "operation", an operation could be a SMS send with code, a human draw etc. when i started that project everything was "signature" :-/

PR #23081, opened 2022-12-02
fix: propal when back to draft delete ip and name of web sign

FHenry:
If a propal is Sign on Web public link and reset to draft we should not keep sign IP and user name information


FHenry:
Task of #23080

PR #23082, opened 2022-12-02
Fix: SQL_ERROR_NO_SUCH_COLUMN

hector-07:
Debugging on version 17, encountered a sql error that column cannot be found. get_product_vat_for_country() is called by the tpl core/tpl/objectline_create.tpl.php
All this being inside compta/facture/card.php


eldy:
You must call the page  /install/upgrade.php to have your database up to date with v17

PR #23100, opened 2022-12-04
fix #23099 : pass massaction to hook instead of action

### LINKED ISSUE #23099: Error on doPreMassActions hook : missing massaction parameter
### Bug

here is the hookmanager call :

$reshook = $hookmanager->executeHooks('doPreMassActions', $parameters, $object, $action);

but $action is not set ... i think (and PR soon) it could be $massaction

### Environment Version

_No response_

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

rycks:
# Fix #23099 

massaction is not send to hookmanager, then we can't catch and implement that hook !


rycks:
sorry for the two commit, after first commit i just have a look to doMassActions implementation then i change my fix to the other one to be more close to doMassActions stuff

rycks:
and an other one to be compatible with develop branch

PR #22254, opened 2022-09-13
Add Extrafields management bom line

atm-lena:
# FIX|Fix #
Add extrafield management for bom line

PR #23104, opened 2022-12-05
Extrafields BomLine Management

atm-lena:
# NEW|New Extrafields BomLine Management

PR https://github.com/Dolibarr/dolibarr/pull/22254 for 17.0


daxitsolutions (commented):
Be carefull, "as-is" this code will not be compliant with v16.0.x 
Tested and imported on a v16.0.3, some new argument needs to be patched too, 
Example : 
@param	string	$fk_unit				Unit
Unknown column 'l.fk_unit'

PR #23106, opened 2022-12-05
FIX INVOICE_CREDIT_NOTE_STANDALONE

lvessiller-opendsi:
FIX INVOICE_CREDIT_NOTE_STANDALONE

### DIFF
--- a/htdocs/fourn/class/fournisseur.facture.class.php
+++ b/htdocs/fourn/class/fournisseur.facture.class.php
@@ -592,7 +592,7 @@ class FactureFournisseur extends CommonInvoice
 		$sql .= ", ".(int) $this->fk_multicurrency;
 		$sql .= ", '".$this->db->escape($this->multicurrency_code)."'";
 		$sql .= ", ".(double) $this->multicurrency_tx;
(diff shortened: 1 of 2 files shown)


htdocs/fourn/class/fournisseur.facture.class.php
```diff
@@ -592,7 +592,7 @@ public function create($user)
 		$sql .= ", ".(int) $this->fk_multicurrency;
 		$sql .= ", '".$this->db->escape($this->multicurrency_code)."'";
 		$sql .= ", ".(double) $this->multicurrency_tx;
-		$sql .= ", ".(isset($this->fk_facture_source) ? $this->fk_facture_source : "NULL");
+		$sql .= ", ".($this->fk_facture_source ? "'".$this->db->escape($this->fk_facture_source)."'" : "null");
```
eldy, line 595: this->fk_facture_source is an int so the good way to forge sql is
$sql .= ", ".($this->fk_facture_source ? ((int) $this->db->escape($this->fk_facture_source)) : "null");

lvessiller-opendsi: @eldy 
I did it

Issue #23094, opened 2022-12-04
Separate expedition billing status from main statuses and add more shipping statuses


milenmk:
### Feature Request

As discussed [HERE](https://github.com/Dolibarr/dolibarr/issues/20987#issuecomment-1336352533), it will be good to separate expedition billing status from shipment status.

### Use case

The reasons for this are, that now when an invoice is created from a shipment (expedition), the column "billed" in llx_expediton remains with a value of 0 instead of 1. Sometimes, there is a workflow, where you create a shipment, then create an invoice and validate this invoice. But the items are NOT shipped yet. So, at a later stage, when the warehouse ship the items they classify the shipment as "Closed (sent)". I.e. you can have an invoice when the shipment is not yet processed. Thus, when you create an invoice, the shipment should be set as billed but NOT closed.

### Suggested implementation

Make following changes:

1. When an invoice is created, no matter the status of the shipment, it should be changed to billed, like "Validated - billed" or, to avoid too many statuses that can be confusing, add a separate field (not editable by user) named  "Billed" in view and list mode

2. Add more Shipment statuses like "Awaiting shipment", "Shipped", "Delivered"

3. This is not very often, but it will be good to have also status "Returned", or "Returned from customer". It could be a wrong address or wrong items but mistakes happens. For this status, it also will be good upon clicking "Set as returned" to have a pop-up to enter the reason for return for future reference.

milenmk:
@eldy could you assign this to me, please?

milenmk:
I am trying to create a pull request for the changes I've made, but I keep getting Travis CI error:

`1) CodingPhpTest::testPHP
Found non escaped string in building of a sql request (case 3) in expedition/card.php: $sql .= ' AND obj.fk_' . $db->escap - Bad.
Failed asserting that false is true.`

and the line in question is `$sql .= ' AND obj.fk_' . $origin . ' = ' . ((int)$origin_id);`

PR #23107, opened 2022-12-05
NEW|New [#23094]

milenmk:
# NEW|New [#23094]
Separate expedition billing status from main statuses and add more shipping statuses


milenmk:
Some thinking. As individuals and small companies may not require more detailed info about expeditions status, should I introduce a new constant like "EXPEDITION_USE_MORE_STATUSES" and show the new buttons only if this constant is set to 1?

The button for closing the shipment is still visible after validation, but once one of the new statuses is applied, now it disappears.

PR #23109, opened 2022-12-05
FIX data too long in ticket message

lvessiller-opendsi:
FIX data too long in ticket message
- limit message field to 65 000 characters

### DIFF
--- a/htdocs/ticket/class/ticket.class.php
+++ b/htdocs/ticket/class/ticket.class.php
@@ -381,7 +381,11 @@ class Ticket extends CommonObject
 
 		if (isset($this->message)) {
 			$this->message = trim($this->message);
-			if (dol_strlen($this->message) > 65000) $this->message = dol_substr($this->message, 0, 65000, 'UTF-8') . '...';
+			if (dol_strlen($this->message) > 65000) {
+				$this->errors[] = 'ErrorFieldTooLong';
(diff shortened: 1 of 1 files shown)


htdocs/ticket/class/ticket.class.php
```diff
@@ -381,6 +381,7 @@ private function verify()
 
 		if (isset($this->message)) {
 			$this->message = trim($this->message);
+			if (dol_strlen($this->message) > 65000) $this->message = dol_substr($this->message, 0, 65000, 'UTF-8') . '...';
```
eldy, line 384: Truncating may generate not valid html. Why doing
```
$this->errors[] = 'ErrorFieldTooLarge';
dol_syslog(get_class($this)."::create error -1 ref null", LOG_ERR);
$result = -1;
```

eldy: I will also change size of field into mediumtext instead of text

lvessiller-opendsi: @eldy 
I did what you suggested to me.
The translate key "ErrorFieldTooLarge" doesn't exist
So, I replaced "ErrorFieldTooLarge" with "ErrorFieldTooLong".
And the mediumtext size had already been changed in your commit : https://github.com/Dolibarr/dolibarr/commit/60133805ee1b2a8b27d9a3b75ccaa2ed901e08bb

Issue #23115, opened 2022-12-06
Linked files in the Project tasks show access denied


pratushraj:
### Bug

Under Project & Collaboration module, if you directly access the Task List and select any task. After selecting the task, go to the Linked Files tab and see the list of Attached files and documents, there you see the thumbnail of the image files is not shown.

### Environment Version

16.0.3

### Environment PHP

8.1.10

### Environment Database

mariaDB

### Environment URL(s)

https://[[HOSTNAME]]/projet/tasks/document.php?id=1&withproject=1&uploadform=1

### Attached files

![Untitled](https://user-images.githubusercontent.com/82569386/205830920-3ff62a6e-785e-4213-8a9c-09cefa70fb6f.png)

PR #23116, opened 2022-12-06
BUG FIXED : Linked files in the Project tasks show access denied #23115

pratushraj:
Now, the linked file in the project task thumbnail issue is fixed, Please check my fix regarding issue #23115.
Here in the code, variable $refname doesn't give the value of the task ref.no.


eldy:
Thanks @pratushraj 
I pushed a different fix that looks better

PR #23123, opened 2022-12-06
Fix : ficheinter cards and stat

Hystepik:
- Add backtopage in card-rec for cancel in create
- fix php 8.1 warnings

### DIFF
--- a/htdocs/fichinter/card-rec.php
+++ b/htdocs/fichinter/card-rec.php
@@ -153,7 +153,7 @@ if ($action == 'add') {
 			setEventMessages($langs->transnoentities("ErrorFieldRequired", $langs->trans("Date")), null, 'errors');
 			$action = "create";
 			$error++;
-		}else {
+		} else {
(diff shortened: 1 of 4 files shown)


htdocs/fichinter/card-rec.php
```diff
@@ -763,22 +788,22 @@
...
-		if ($search_societe) {
+		if (!empty($search_societe)) {
 			$sql .= natural_search('s.nom', $search_societe);
 		}
-		if ($search_frequency == '1') {
+		if (!empty($search_frequency) && $search_frequency == '1') {
 			$sql .= ' AND f.frequency > 0';
 		}
-		if ($search_frequency == '0') {
+		if (!empty($search_frequency) && $search_frequency == '0') {
```
eldy, line 806: This test will always be wrong.   empty('0') return true 

Try
`if (isset($search_frequency) && (string) $search_frequency == '0') {`

PR #23130, opened 2022-12-06
NEW|New [#23094]

milenmk:
Add more shipment statuses (In Process > Prepared (stock decrease) > Shipped > Delivered > Returned (stock increase))

Reopen is available at any stage, unless the status is Returned, to avoid a double stock increase(return).

More shipment statuses are visible and available if, in Setup>Other, the constant `EXPEDITION_USE_MORE_STATUSES` is set to 1


eldy:

PR #23131, opened 2022-12-06
NEW|New [Add link to other parameters]

milenmk:
On Setup > Other, add a link on top of the page to https://wiki.dolibarr.org/index.php?title=Setup_Other

### DIFF
--- a/htdocs/admin/const.php
+++ b/htdocs/admin/const.php
@@ -175,6 +175,8 @@ print load_fiche_titre($langs->trans("OtherSetup"), '', 'title_setup');
 
 print '<span class="opacitymedium">'.$langs->trans("ConstDesc")."</span><br>\n";
 print "<br>\n";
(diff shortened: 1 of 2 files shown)


htdocs/admin/const.php
```diff
@@ -175,6 +175,8 @@
 
 print '<span class="opacitymedium">'.$langs->trans("ConstDesc")."</span><br>\n";
 print "<br>\n";
+print '<span class="opacitymedium">'. $langs->trans('ConstLink') . " <a href='https://wiki.dolibarr.org/index.php?title=Setup_Other' target='_blank'>https://wiki.dolibarr.org/index.php?title=Setup_Other</a></span><br>\n";
```
eldy, line 178: The list of options on this page are dangerous and not stable parameters. So it looks a bad idea to offer a visibility of this parameters for end users.

milenmk: I see a lot of topics in the forum (English version) that can be solved by using one or some of the parameters. Therefore, I get the impression that almost none deeply involved in Dolibarr knows them and many of them can be helpful for everyday use.

Maybe add an additional warning?

eldy: Don't trust all message on forum. It may solve a need sometimes but this is often by introducing more troubles.
So unstable options must not appears as public options.

PR #23141, opened 2022-12-07
FIX : mail templates - $userlocalbrowser false

atm-lena:
# FIX|Fix #mail templates - $userlocalbrowser false

We can't search in local files pictures to add in mail templates.

**Now :** 
image

**With FIX :** 

image

### DIFF
--- a/htdocs/admin/mails_templates.php
+++ b/htdocs/admin/mails_templates.php
@@ -740,7 +740,7 @@ if ($action == 'view') {
 			if (empty($conf->global->FCKEDITOR_ENABLE_MAIL)) {
 				$okforextended = false;
 			}
-			$doleditor = new DolEditor($tmpfieldlist, (!empty($obj->{$tmpfieldlist}) ? $obj->{$tmpfieldlist} : ''), '', 180, 'dolibarr_mailings', 'In', 0, false, $okforextended, ROWS_4, '90%');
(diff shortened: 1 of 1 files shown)


eldy:
For email, we don't want to be able to link to local email here. Because local files are local (and need a login/pass to be readable) so can not be read from an email. So it is on purpose this is disabled for email and email templates.

atm-thibaultf:
@eldy : I don't understand, it is possible to add a local image in the user signature and it works fine. The image is uploaded in the folder **/documents/medias/image/** and it seems ok to diplay the image for the recipient of the email without login/pass. Why is it not possible to use the same way on email template ?

eldy:
After you change, what is the url you got into email content into the html source of the received email after <img src=... ?

atm-thibaultf:
The URL of the image when added into the user signature is : `URL_DOLIBARR/viewimage.php?modulepart=medias&entity=1&file=image/nom_image.jpg`
I tested it with Gmail and Google change the URL to use a temporary URL, so it's not really relevant in Gmail because tu URL received is : ``
But it seems to work even with others messaging tools.

eldy:
Thanks for analysis. Trouble is morecomplex. I pushed today a fix into v17 related to this by introducing a method to know if we can or not accept the filemaneger on ckeditor because there is some cases, it won't work.
As it needs a non minor change i pushed it in v17 instead.

PR #23143, opened 2022-12-07
NEW|New [Bulk delete Project tasks]

milenmk:
NEW|New [Bulk delete Project tasks]

Mass action to delete multiple tasks from the tab Tasks of Project

Screenshot_1

Screenshot_2

Screenshot_3


htdocs/core/class/html.form.class.php
```diff
@@ -7286,7 +7286,7 @@ public function selectProjects($selected = '', $htmlname = 'projectid', $filtert
 			$out .= $this->selectProjectsList($selected, $htmlname, $filtertype, $limit, $status, 0, $socid, $showempty, $forcecombo, $morecss);
 		}
 
-		if (empty($nooutput)) print $out;
```
eldy, line 7289: If nooutput is not set (0 = empty), we want output on screen, so we want print $out

PR #23144, opened 2022-12-07
Fix(scrutinizer)

KiVoilaJe:
https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/expedition/class/expedition.class.php?selectedLabels%5B0%5D=9&selectedSeverities%5B0%5D=10&orderField=lastFound&order=desc&honorSelectedPaths=0

I can't find the error for the 3 var. I push and get the right info

### DIFF
--- a/htdocs/expedition/class/expedition.class.php
+++ b/htdocs/expedition/class/expedition.class.php
@@ -54,6 +54,11 @@ class Expedition extends CommonObject
 	use CommonIncoterm;
 
 	/**
+	 * @var string code
+	 */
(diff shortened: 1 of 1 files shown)


eldy:
srutinizer say the property   ->code must be into Expedition because it is inside class CommonIncoterm. But looking in code CommonIncoterm show there is no property ->code. So this is clearly a bad advice of scrutinizer.  I forced the "ignore error" of this warning in scrutinizer setup.

PR #23145, opened 2022-12-07
fix(scrutinizer)

KiVoilaJe:
https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/reception/stats/month.php?selectedLabels%5B0%5D=9&selectedSeverities%5B0%5D=10&orderField=lastFound&order=desc&honorSelectedPaths=0

### DIFF
--- a/htdocs/reception/stats/month.php
+++ b/htdocs/reception/stats/month.php
@@ -49,9 +49,8 @@ $HEIGHT = DolGraph::getDefaultGraphSizeForStats('height');
 $mesg = '';
 
 print load_fiche_titre($langs->trans("StatisticsOfReceptions").' '.GETPOST("year", 'int'), $mesg);
-
-$stats = new ReceptionStats($db);
-$data = $stats->getNbReceptionByMonth(GETPOST("year", 'int'));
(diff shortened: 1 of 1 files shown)


htdocs/reception/stats/month.php
```diff
@@ -49,9 +49,8 @@
 $mesg = '';
 
 print load_fiche_titre($langs->trans("StatisticsOfReceptions").' '.GETPOST("year", 'int'), $mesg);
-
-$stats = new ReceptionStats($db);
-$data = $stats->getNbReceptionByMonth(GETPOST("year", 'int'));
+$stats = new ReceptionStats($db, $socid, '', ($userid > 0 ? $userid : 0));
```
eldy, line 52: You introduce the parameter $socid.
Yes, the method need it. But it is not defined previously (at least not always).
You must add
$socid = GETPOST('socid', 'int') at line 32 because the id of the company come from a parameter into the form.

Same for $userid.  Where is this var set ? It miss
$userid = GETPOST('userid', 'int')  
that also come as parameter on page:

image

PR #23147, opened 2022-12-07
Head contact database structure

BB2A-Anthony:
Issue #20137

[*Long description*]

[*Long description*]

[*Long description*]


htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -389,3 +389,4 @@ ALTER TABLE llx_opensurvey_comments ADD COLUMN date_creation datetime NULL;
 
 ALTER TABLE llx_c_tva ADD COLUMN use_default tinyint DEFAULT 0;
 
+ALTER TABLE llx_socpeople ADD COLUMN fk_parent integer DEFAULT 0 NOT NULL AFTER fk_stcommcontact;
```
eldy, line 392: Why not instead :
ALTER TABLE llx_socpeople ADD COLUMN fk_parent integer NULL AFTER fk_stcommcontact;
?
Value fk_parent will be filled only when a parent exists and if we add an index on field, it will consume only for fk_parent defined.

bb2a: Ok i enderstand it's done

htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -389,3 +389,5 @@ ALTER TABLE llx_opensurvey_comments ADD COLUMN date_creation datetime NULL;
 
 ALTER TABLE llx_c_tva ADD COLUMN use_default tinyint DEFAULT 0;
 
+ALTER TABLE llx_socpeople ADD COLUMN update_parent integer AFTER fk_stcommcontact;
```
eldy, line 392: What is the goal of thefield update_parent ?

eldy:
I have merged th addition of field fk_parent into develop branch. But not the field update_parent, because the choice to delete or not child when we delete a contact should/can be a question on the user interface, so seems not required on database.

Issue #23075, opened 2022-12-01
Wrong invoice amount when billing Project task spent time


milenmk:
### Bug

When creating an invoice for project tasks time spent, the amount in the invoice is wrong.

### Environment Version

14-16

### Expected and actual behavior

When the task is for 10 hours and the user pay rate is 15 currency/h, the invoice line should be for QTY = 10 and U.P net = 15.

In reality, the invoice QTY is correct, but the U.P net is not the hourly rate, but the total for the task (i.e. 150)

### Steps to reproduce the behavior

Create a project, create a task, add time spent to this task, and then bill it.

![1](https://user-images.githubusercontent.com/4587305/205131290-09b97066-b0c0-4559-902a-ec8c3ec48f8d.png)
![2](https://user-images.githubusercontent.com/4587305/205131304-e52121f7-d96d-4a1b-b211-c43fdd31fe04.png)
![3](https://user-images.githubusercontent.com/4587305/205131313-0977fd8a-5d5c-4213-aec6-3b5703adf623.png)

milenmk:
P.S.: If in file "/projet/tasks/time.php" line 465 is uncommented and changed to `$pu_ht = $fuser->thm;` the invoice is correct.

eldy:
It seems fix was already pushed into v16. So it will be merged up in higher version (17, develop, ) automatically. No need to repush the pr in develop

PR #23158, opened 2022-12-08
FIX|Fix [#23075]

milenmk:
Wrong invoice unit price when billing Project task spent time

Instead of using the hourly rate for a user, the total time amount is put as a unit price

### DIFF
--- a/htdocs/projet/tasks/time.php
+++ b/htdocs/projet/tasks/time.php
@@ -589,6 +589,7 @@ if ($action == 'confirm_generateinvoice') {
 				foreach ($arrayoftasks as $timespent_id => $value) {
 					$userid = $value['user'];
 					//$pu_ht = $value['timespent'] * $fuser->thm;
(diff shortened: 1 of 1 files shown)


htdocs/projet/tasks/time.php
```diff
@@ -589,6 +589,7 @@
 				foreach ($arrayoftasks as $timespent_id => $value) {
 					$userid = $value['user'];
 					//$pu_ht = $value['timespent'] * $fuser->thm;
+					$pu_ht = $fuser->thm;
```
eldy, line 592: The price per hour of the user may change with time.
And total of timespent may have been calculated with different rate that change with time. Each time line may have been entered with a different hourly rate.
Di you try this instead 
$pu_ht = price2num($value['totalvaluetodivideby3600'] / 3600, 'MU');

eldy:
Duplicate of #23159

milenmk:
#23159 is for v16, this one is for v17

PR #23159, opened 2022-12-08
FIX|Fix [#23075]

milenmk:
Wrong invoice unit price when billing Project task spent time

Instead of using the hourly rate for a user, the total time amount is put as a unit price

### DIFF
--- a/htdocs/projet/tasks/time.php
+++ b/htdocs/projet/tasks/time.php
@@ -462,7 +462,7 @@ if ($action == 'confirm_generateinvoice') {
 
 				foreach ($arrayoftasks as $userid => $value) {
 					$fuser->fetch($userid);
-					//$pu_ht = $value['timespent'] * $fuser->thm;
+					$pu_ht = $fuser->thm;
 					$username = $fuser->getFullName($langs);
 
 					// Define qty per hour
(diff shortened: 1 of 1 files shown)


eldy:
The price per hour of the user may change with time.
And total of timespent may have been calculated with different rate that change with time. Each time line may have been entered with a different hourly rate.
Di you try this instead
$pu_ht = price2num($value['totalvaluetodivideby3600'] / 3600, 'MU');

milenmk:
> $pu_ht = price2num($value['totalvaluetodivideby3600'] / 3600, 'MU');

With this code, I've tested a project with 2 different tasks. Task one has a time spent record for 2 hours, and the second task has a time record for 3 hours. Total is 5 hours i.e. 100 EUR. When generating an invoice with the option "one line for each time spent declaration", the result is 2 rows with Qty =2 and Qty = 3 (as expected) but again, the U.P.net is the total for the line. Same result when both time records are for the same task.

I.e. nothing is changed, regarding the bug.

As for the potential hourly rate value change:
1. If the user's hourly value is changed, already recorded time spent records do not change their value.
2. Newly added time spent records take the new value
3

Screenshot_1

milenmk:
Did a few more tests, and `$pu_ht = $value['totalvaluetodivideby3600'] / $value['timespent'];` also provides correct U.P.Net. Should I hange both commits with this code?

eldy:
If you set 
$pu_ht = $fuser->thm;
the unit price will be the same for all lines. When it should be the one stored on each timespent line (column "thm" of table llx_projet_task_time).

PR #23166, opened 2022-12-09
Delete MO :  object's links must be deleted

atm-lena:
# FIX|Fix #
When we delete an mo, we must deleted all links with others elements.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -9472,6 +9472,12 @@ abstract class CommonObject
 			}
 		}
 
+		// Delete linked object
+		$res = $this->deleteObjectLinked();
+		if ($res < 0) {
+			$error++;
+		}
+
 		// Commit or rollback
 		if ($error) {
 			$this->db->rollback();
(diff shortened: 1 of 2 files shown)


htdocs/mrp/class/mo.class.php
```diff
@@ -744,8 +744,31 @@ public function updateProduction(User $user, $notrigger = true)
...
 	public function delete(User $user, $notrigger = false)
 	{
-		return $this->deleteCommon($user, $notrigger);
-		//return $this->deleteCommon($user, $notrigger, 1);
+		$this->db->begin();
+
+		$error = 0;
+
+		// Delete linked object
+		$res = $this->deleteObjectLinked();
```
eldy, line 752: Did you think of adding the deleteObjectLinked into the method deleteCommon ?

htdocs/core/class/commonobject.class.php
```diff
@@ -9472,6 +9472,12 @@ public function deleteCommon(User $user, $notrigger = false, $forcechilddeletion
 			}
 		}
 
+		// Delete linked object
+		$res = $this->deleteObjectLinked();
```
eldy, line 9476: This block should be before the deleteExteafields because order must be the oposite if create. Create should be
Insert object
Insert extrafields
Insert links
Insert other 
Call trigger

So delete should be
Call trigger
Delete other
Delete links (the part you added)
Delete extrafields
Delete object

So just move just before the delete of extrafields...

PR #23171, opened 2022-12-09
NEW inventory without virtual products (kits)

lvessiller-opendsi:
NEW inventory without virtual products (kits)
- remove virtual products in inventory, only keep children
- it's in accordance with #21088

**Before**
When you start an inventory for a warehouse, you could have some virtual products :
- PANIER1 is a virtual product
image

- MINIPANIERGARNI is a virtual product (sub-component of PANIER1)
image

And in the inventory :
image

**After**
In the inventory :
image

So we keep only final sub-components of virtual products (kits)


lvessiller-opendsi:
@eldy 
What do you think about this PR ?

PR #23173, opened 2022-12-09
NEW Various payment - List - Add massaction to delete

aspangaro:
image


eldy:
@aspangaro pr still in draft ?

PR #23177, opened 2022-12-10
v17.0: Wrong invoice amount when billing project task spent time

milenmk:
FIX|Fix [#23075]

v17.0: Wrong invoice amount when billing project task spent time

### DIFF
--- a/htdocs/projet/tasks/time.php
+++ b/htdocs/projet/tasks/time.php
@@ -600,6 +600,7 @@ if ($action == 'confirm_generateinvoice') {
 					}
 					*/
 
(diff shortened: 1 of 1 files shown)


milenmk:
Cannot pass the Travis CI as it gives an error unrelated to the commit:

```
1) CodingPhpTest::testPHP
Found a forged SQL string that mix on same line the use of ' for PHP string and PHP variables into file intracommreport/list.php Use " to forge PHP string like this: $sql = "SELECT ".$myvar...
```

PR is made on a fresh branch created from Dolibarr > develop

PR #23181, opened 2022-12-11
NEW Allow also not latin characters in {tttt} tag

sonikf:
# NEW|New [*Allow also non latin characters in {tttt} tag*]
[*Allow also non latin characters in {tttt} tag in my case Greek to use as a mask*]


eldy (commented):
A better solution is to use dol_substr()
Can you test and resubmit PR with that ?

Issue #23136, opened 2022-12-07
Error on creating variant with the API


Franckapik:
### Bug

I try to make this api POST request to add variant of a product : 

```
curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' --header 'DOLAPIKEY: myapikey' -d '{ \ 
   "weight_impact": 0, \ 
   "price_impact": 0, \ 
   "price_impact_is_percent": true, \ 
   "features": [ \ 
     "1 => 1" \ 
   ] \ 
 }' 'https://quadratikfr.saas2.doliondemand.fr/api/index.php/products/1/variants'
```

I have this error : 

```
{
  "error": {
    "code": 401,
    "message": "Unauthorized"
  },
  "debug": {
    "source": "api_products.class.php:1741 at call stage",
    "stages": {
      "success": [
        "get",
        "route",
        "negotiate",
        "authenticate",
        "validate"
      ],
      "failure": [
        "call",
        "message"
      ]
    }
  }
}
```

**Is this a syntax error or missing data or finally a bug from this endpoint api ?** I have tested POST a product and it's working but this i have 401 error with variant request. We can't know more on the error report. 

Talking about POST product, do i miss documentation about the required data to send or is there any information missing in the proposed model on the explorer page ? (only  "request_data": ["string"] ? ) I suppose that the model should be too long probably but we don't really know what key:value we can use for it ... 

### Environment Version

16.0.0

### Environment OS

Linux Debian 4.19.67-2+deb10u2 (2019-11-11) x86_64

### Environment Web server

Apache/2.4.38 (Debian)

### Environment PHP

7.3.31-1~deb10u1

### Environment Database

5.5.5-10.3.36-MariaDB-0+deb10u1

### Expected and actual behavior

I would like to add a new variant by using the api. I have an 401 error. I don't know if i am wrong on the syntax or if there is a bug on the variant POST endpoint of the API. I have tested this on my server with dolibarr 16.0.0 and on online demo with 16.0.3 . Error 401 or 404.

### Steps to reproduce the behavior

- have all permissions on the superadmin user
- go to the api explorer
- paste the doliapikey and click explore
- go to [POST] /products/{id}/variants
- complete the id of the supposed parent product and add this object : 

{
  "weight_impact": 0,
  "price_impact": 0,
  "price_impact_is_percent": true,
  "features": [
    "1=>1"
  ]
}

### Attached files

[...]

PR #23182, opened 2022-12-11
FIX: confusing error codes in products API

atm-florianm:
# FIX confusing error messages in the products API

cf. issue https://github.com/Dolibarr/dolibarr/issues/23136

The HTTP response codes can be confusing because 401 is for requests by unauthorized users. In the issue above, the user is authorized but provided wrong parameter values.

[edit] Should I move the PR up one or two versions (v14)? I did it in v16 just because it was more convenient.

### DIFF
--- a/htdocs/product/class/api_products.class.php
+++ b/htdocs/product/class/api_products.class.php
@@ -1727,8 +1727,12 @@ class Products extends DolibarrApi
 			throw new RestException(401);
 		}
 
-		if (empty($id) || empty($features) || !is_array($features)) {
-			throw new RestException(401);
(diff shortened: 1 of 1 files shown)


eldy:
We can't change a result code of an api into a stable version.
If it was not clear it must remain not clear. Not clear us better than a compatibility regression of a maintenance release.
You must it in develop.

atm-florianm:
> We can't change a result code of an api into a stable version. If it was not clear it must remain not clear. Not clear us better than a compatibility regression of a maintenance release. You must it in develop.

I changed the target branch to `develop`. I believe it is mergeable.

PR #23198, opened 2022-12-12
Update llx_20_c_departements.sql

dolibit-ut:
removed 'active' from UK


eldy:
What is goal of this change ?

dolibit-ut:
to transfer all data into a uniform form

PR #23203, opened 2022-12-13
do not delete reception if not draft

frederic34:


eldy:
Sometimes, we just want to delete record to clean data. When we have delete permission, we are able to delete record for other object. Why not for receptions. By definition there is no need of a permission to delete a draft (draft means "does not exists yet", so permission to delete means delete when not draft.

frederic34:
@eldy if we delete a validated reception, stock is not decreased. To decrease stock we need to go back draft before.
Deleteting validated reception give stock errors.

eldy:
> @eldy if we delete a validated reception, stock is not decreased. To decrease stock we need to go back draft before.
> Deleteting validated reception give stock errors.

It does not give stock error, the delete reception exist to be able to delete a reception without changing the stock. So this is what we want . Why changing the stock when we just want to clear a list of old reception ? This the the reason why it needs a particular permission. Removing a reception is not the same than returning product (a feature not yet implemented). Only returning product,  canceling a reception, must modify the stock.
However, we can add a warning into the confirmation message to say stock won't be affected.

Issue #23202, opened 2022-12-13
Editing an existing third party - Parent company is removed on save


mattcole-42:
### Bug

When editing an existing third party the parent company is removed when saved, as the edit form has no field for parent company.

Version used Dolibarr 16.0.3

### Expected and actual behavior

Expected
The non edited fields including parent company would remain untouched

Actual:
the third party was saved, however the parent company was removed.

### Steps to reproduce the behavior

Pre-Requisite:
Third party exists with a parent company linked.

Steps:
Open Third Party
select modify
change any field
save

milenmk:
Confirmed. I'll make a fix.

Same bug is on v17 also.

PR #23204, opened 2022-12-13
v16 Parent company field missing when editing a third party

milenmk:
FIX|Fix [#23202]

Fix for V16

Parent company field missing when editing a third party


eldy:
Merged in v17.

milenmk:
This is for v16. Bug is the same.

PR #23207, opened 2022-12-13
Fix v14 Darkmode input disabled color and bgcolor

daraelmin:
Text in input is not readable when input box is disabled.
Input box are not delimited in darkmode
 
before PR
image

after PR
image


eldy:
The darkmode was still exoerimental in v14. Do you experience trouble with v17 ? If yes, can you fix in v17 branch ?

daraelmin:
Hi @eldy ,
As it's just a display problem, not a real fix of feature, I think it could be corrected for v14, v15, v16 and v17, but you're the boss, I'll do an other PR for v17.
Cheers

PR #23217, opened 2022-12-14
16.0 FIX Bom Calculate Price

atm-lena:
# FIX|Fix #

- Price of sub bom didn't display
- Lag Case Table
- Prices format

### DIFF
--- a/htdocs/bom/tpl/objectline_view.tpl.php
+++ b/htdocs/bom/tpl/objectline_view.tpl.php
@@ -239,13 +239,13 @@ if ($resql) {
 		// Cost
 		if (!empty($sub_bom->id)) {
 			$sub_bom->calculateCosts();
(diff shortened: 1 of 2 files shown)


htdocs/bom/tpl/objectline_view.tpl.php
```diff
@@ -239,13 +247,13 @@
 		// Cost
 		if (!empty($sub_bom->id)) {
 			$sub_bom->calculateCosts();
-			print '<td class="linecolcost nowrap right" id="sub_bom_cost_'.$sub_bom_line->id.'"><span class="amount">'.price($sub_bom->total_cost * $sub_bom_line->qty * $line->qty).'</span></td>';
+			print '<td class="linecolcost nowrap right" id="sub_bom_cost_'.$sub_bom_line->id.'"><span class="amount">'.price2num($sub_bom->total_cost * $sub_bom_line->qty * $line->qty, 'MT').'</span></td>';
```
eldy, line 250: You must keep the price.
The price2num goal is to round a real number (in php memory)
But when we want to ouput the value to match end user localization for a price, we must use price.
So here i think it should be
price(price2num(...

atm-lena:
@eldy fixed

PR #23225, opened 2022-12-14
FIX move llx_user_perentity table in dolibarr core

hregis:


htdocs/install/mysql/tables/llx_user_perentity-multicompany.sql
```diff
@@ -0,0 +1,33 @@
...
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU General Public License for more details.
+--
+-- You should have received a copy of the GNU General Public License
+-- along with this program; if not, write to the Free Software
+-- Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA 02111-1307, USA.
+--
+-- ===========================================================================
+
+create table llx_user_perentity
```
eldy, line 20: Currently a user and its properties are already linked to 1 entity. Also permission system are also on a user. Having one user with different properties for n entities (what you introduce) seems not compatible with n user for one entity (what we need for permission system and security to work).
Adding such a feature will bring a very high complexity and a lot of troubles into the core, some without solution, for the need of 1 external module.
For the moment the permission system for multicompany is still not clear and is source of a lot of trouble to know which data can be seen or not. I prefer to not add more complexity on users management in the core and let the trouble this will generate out of Dolibarr core, so inside external modules.

hregis: @eldy 
I wanted to include it in the core like llx_product_perentity and llx_societe_perentity so that it is fully integrated, for Dolibarr and for external modules that support Multicompany. Otherwise it is unmanageable for the moment, because there is too much code that we cannot customize and we would have inconsistencies.

eldy:
Table llx_product_perentity and llx_societe_perentity were introduced with a suffix -multicompany.sql so tables are created only hen module multicompany is enabled.

PR #23237, opened 2022-12-15
FIX : "read" right of hrm evaluations must allow to see only user eva…

atm-gauthier:
…luations and those of subordinates, readall is another right + several fixes


eldy:
Module was experimental in v16, no need to push fixes. Can you push it in v17 ?

atm-gauthier:
Hi @eldy, i just did merge conflits

atm-gauthier:
@eldy can you say me if it's ok for this pr ?

PR #23251, opened 2022-12-19
I think it should be $obj->rowid

mannysoft:
I think it should be `$obj->rowid` because when its just `$obj->id` then it goes nowhere.


eldy:
Right. Thanks for the fix

PR #23254, opened 2022-12-19
NEW Kanban view for list of adherents 

lamrani002:
# NEW mode view "kanban" for list of adherents by adding function getKanbanview() in class adherent


htdocs/core/lib/project.lib.php
```diff
@@ -1064,6 +1064,239 @@ function projectLinesa(&$inc, $parent, &$lines, &$level, $var, $showproject, &$t
 }
 
 
+/**
```
eldy, line 1067: We should not have this file modified into this PR

PR #23261, opened 2022-12-20
FIX: trigger LINEORDER_DELETE, delete extrafields should be moved after delete trigger.

fappels:


htdocs/commande/class/commande.class.php
```diff
@@ -4331,23 +4331,22 @@ public function delete(User $user, $notrigger = 0)
 		dol_syslog("OrderLine::delete", LOG_DEBUG);
 		$resql = $this->db->query($sql);
 		if ($resql) {
-			// Remove extrafields
-			if (!$error) {
-				$this->id = $this->rowid;
-				$result = $this->deleteExtraFields();
+			if (!$error && !$notrigger) {
```
eldy, line 4334: In fact, it even should be before the DELETE of commandedet
In create, we do create line, then extrafields, then trigger. 
So in delete we should have the reverse order. Don't you think so ?

fappels: @eldy Moving delete extrafield after trigger (reverse order) is what this fix does. On create you add extrafield and trigger, on delete you should trigger and delete extrafield. If you block delete in trigger for some reason to disallow delete, you loose you extrafield data. I compared with invoice delete line trigger where already delete extrafield after trigger is implemented.

PR #23263, opened 2022-12-20
New mode view for adherent type List

lamrani002:
# NEW mode kanban for adherent type list 
NB: add function getKanbanView() in adherent_type class


htdocs/core/lib/project.lib.php
```diff
@@ -1064,6 +1064,239 @@ function projectLinesa(&$inc, $parent, &$lines, &$level, $var, $showproject, &$t
 }
 
 
+/**
```
eldy, line 1067: Pb. This file should not be included into this PR

PR #23265, opened 2022-12-20
NEW : notify also the contributor affected to a ticket if a new message public is post

kkhelifa-opendsi:
notify also the contributor affected to a ticket if a new message public is post (add global TICKET_PUBLIC_NOTIFICATION_NEW_MESSAGE_ALSO_CONTRIBUTOR)


BB2A-Anthony:
Hi good job

Why not add the param in module setup page

eldy:
> Hi good job
> 
> Why not add the param in module setup page

I don't think we need a new param. Now when a notif email has to be sent, it will be always to all contacts/contributors of ticket. I think this has been set as degault behaviour during the merge.

PR #23272, opened 2022-12-20
New mode view for contact adress list

lamrani002:
# NEW|New mode view "kanban" to contact adress 
in function getKanbanview() : if **socid** is null we show phone Number else we show information about societe with getNomUrl() function .


htdocs/core/lib/project.lib.php
```diff
@@ -1064,6 +1064,239 @@ function projectLinesa(&$inc, $parent, &$lines, &$level, $var, $showproject, &$t
 }
 
 
+/**
```
eldy, line 1067: Something wrong here, we should not have this file modified in this PR

PR #23276, opened 2022-12-20
StockMvt Class : dol_syslog add level LOG_ERR

FHenry:
StockMvt Class : dol_syslog add level LOG_ERR


htdocs/product/stock/class/mouvementstock.class.php
```diff
@@ -259,7 +259,7 @@ public function _create($user, $fk_product, $entrepot_id, $qty, $type, $price =
 			if (empty($batch)) {
 				$langs->load("errors");
 				$this->errors[] = $langs->transnoentitiesnoconv("ErrorTryToMakeMoveOnProductRequiringBatchData", $product->ref);
-				dol_syslog("Try to make a movement of a product with status_batch on without any batch data");
+				dol_syslog("Try to make a movement of a product with status_batch on without any batch data", LOG_ERR);
```
eldy, line 262: LOG_ERR means to the admin user:  There is a critical pb in app. But here it seems it is just bad data entry by end user. So we should return error to user, but not on system log ?

PR #23283, opened 2022-12-20
NEW :  Uniformize member stats in boxes and index

daraelmin:
One request per graph instead of four or five

Befor PR

image

After PR

image


htdocs/adherents/class/adherentstats.class.php
```diff
@@ -181,4 +182,56 @@ public function getAllByYear()
...
+	 * 	@return		array                   Array with
+	 */
+	public function countMembersByTypeAndStatus()
+	{
+		global $user;
+
+		$now = dol_now();
+
+		$sql = "SELECT COALESCE(d.fk_adherent_type, 'total') as fk_adherent_type, t.libelle as label";
+		$sql .= ", COUNT(CASE WHEN d.statut = ".Adherent::STATUS_DRAFT." THEN 'members_draft' ELSE NULL END) as members_draft";
```
eldy, line 199: Do not use CASE WHEN into sql, use instead  $db->ifsql method that will generate the sql using the if with the syntax according to database manager.

htdocs/adherents/class/adherentstats.class.php
```diff
@@ -181,4 +182,56 @@ public function getAllByYear()
...
+		$sql .= ", COUNT(CASE WHEN d.statut = ".Adherent::STATUS_VALIDATED."  AND (d.datefin IS NULL AND t.subscription = '1') THEN 'members_pending' ELSE NULL END) as members_pending";
+		$sql .= ", COUNT(CASE WHEN d.statut = ".Adherent::STATUS_VALIDATED."  AND (d.datefin >= '".$this->db->idate($now)."' OR t.subscription = 0) THEN 'members_uptodate' ELSE NULL END) as members_uptodate";
+		$sql .= ", COUNT(CASE WHEN d.statut = ".Adherent::STATUS_VALIDATED."  AND (d.datefin < '".$this->db->idate($now)."' AND t.subscription = 1) THEN 'members_expired' ELSE NULL END) as members_expired";
+		$sql .= ", COUNT(CASE WHEN d.statut = ".Adherent::STATUS_EXCLUDED." THEN 'members_excluded' ELSE NULL END) as members_excluded";
+		$sql .= ", COUNT(CASE WHEN d.statut = ".Adherent::STATUS_RESILIATED."  THEN 'members_resiliated' ELSE NULL END) as members_resiliated";
+		$sql .= " FROM ".MAIN_DB_PREFIX."adherent as d , ".MAIN_DB_PREFIX."adherent_type as t";
```
eldy, line 205: If a type has no member at all, it seems this request will return nothing when need need a count per type, so if there is no member at all for this type, we need a result into the sql request. May be we need to convert the join into left join ?

htdocs/adherents/class/adherentstats.class.php
```diff
@@ -181,4 +182,56 @@ public function getAllByYear()
...
+		$sql .= " WHERE t.rowid = d.fk_adherent_type";
+		$sql .= " AND d.entity IN (" . getEntity('adherent') . ")";
+		$sql .= " AND t.entity IN (".getEntity('member_type').")";
+		$sql .= " AND t.statut = 1";
+		$sql .= " GROUP BY d.fk_adherent_type";
+		$sql .= " WITH ROLLUP;";
```
eldy, line 211: This instruction must be avoided :  Not managed the same way by different databases.
Also the use of ROLLUP break the purity of data. Intermediary aggregation can easily be done using PHP instead keeping the array of data no corrupted by artificially inserted data and keep the array result pure.

htdocs/adherents/class/adherentstats.class.php
daraelmin: Done

htdocs/adherents/class/adherentstats.class.php
daraelmin: Done

htdocs/adherents/class/adherentstats.class.php
daraelmin: Done

daraelmin:
Fixed as suggested
Add a new column "Total by status" in box
Change label to match the graph and the left menu

daraelmin:
@eldy hope it's better now

daraelmin:
With all last modification (Add total, limit in time, and add period in title)

image

PR #23294, opened 2022-12-21
FIX : get multicurrency infos of propal when create order from propal…

atm-gauthier:
… with "WORKFLOW_PROPAL_AUTOCREATE_ORDER" conf

### DIFF
--- a/htdocs/commande/class/commande.class.php
+++ b/htdocs/commande/class/commande.class.php
@@ -1355,7 +1355,7 @@ class Commande extends CommonOrder
 		$this->origin = $object->element;
 		$this->origin_id = $object->id;
 
-                // Multicurrency (test on $this->multicurrency_tx because we should take the default rate only if not using origin rate)
+				// Multicurrency (test on $this->multicurrency_tx because we should take the default rate only if not using origin rate)
 		if (!empty($conf->multicurrency->enabled)) {
 			if (!empty($object->multicurrency_code)) {
 				$this->multicurrency_code = $object->multicurrency_code;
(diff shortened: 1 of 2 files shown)


atm-gauthier:
Hi @eldy, i don't understand this message in travis error :
image

eldy:
> Hi @eldy, i don't understand this message in travis error :
> image
> 

Don't understand too. May be the use of list(...) = ...
Avoid use of list() =.
Try to storz into a tmp array first $tmparray = ...
Then use tmparray[0], tmparray[1]... if defined.

atm-gauthier:
@eldy i have the same problem after modification :(

eldy:
i found the trouble. It is this fuc... composer. When phpunit is installer with composer 2.5 instead of 2.4, even if phpunit is same version, result differs (composer add stupid autoload code).

atm-gauthier:
@eldy ok, do you know if i can do something about this ? Or are you going to merge anyway ?

frederic34:
@atm-gauthier can you try to comment this line : https://github.com/Dolibarr/dolibarr/blob/fc466b1ce30b526b7feb6df60602bcd34f0db5b2/htdocs/includes/restler/framework/Luracast/Restler/AutoLoader.php#L78
I got similar problem using dolibarr api and other autoloader, this line unregister other autoloaders to keep only restler autoloader (function name is thereCanBeOnlyOne)

atm-gauthier:
@frederic34 thank you for your answer but as i don't know this file, i prefer ask @eldy if this is what i should do ?

eldy:
@atm-gauthier nothing to do on your side, i have fixed this by forcing composer version used to install phpunit in travis setup.

PR #23295, opened 2022-12-21
Fix(scrutinizer)

KiVoilaJe:
https://scrutinizer-ci.com/g/Dolibarr/dolibarr/issues/develop/files/htdocs/public/project/new.php?selectedLabels%5B0%5D=9&selectedSeverities%5B0%5D=10&orderField=lastFound&order=desc&honorSelectedPaths=0


htdocs/public/project/new.php
```diff
@@ -338,50 +338,50 @@ function llxFooterVierge()
...
-					complete_substitutions_array($substitutionarray, $outputlangs, $object);
-					$subjecttosend = make_substitutions($subject, $substitutionarray, $outputlangs);
-					$texttosend = make_substitutions($msg, $substitutionarray, $outputlangs);
-
-					if ($subjecttosend && $texttosend) {
-						$moreinheader = 'X-Dolibarr-Info: send_an_email by public/lead/new.php'."\r\n";
-
-						$result = $object->send_an_email($texttosend, $subjecttosend, array(), array(), array(), "", "", 0, -1, '', $moreinheader);
-					}
+				// if ($object->email) {
```
eldy, line 341: It seems we have en email when a user post the form.
So the email is inside  GETPOST("email")
Why not just using it to send the email ?

PR #23296, opened 2022-12-21
 FIX : Product list in setup.php in new Module

IC-Florian:


htdocs/modulebuilder/template/admin/setup.php
```diff
@@ -55,6 +55,7 @@
 // Libraries
 require_once DOL_DOCUMENT_ROOT."/core/lib/admin.lib.php";
 require_once '../lib/mymodule.lib.php';
+require_once DOL_DOCUMENT_ROOT.'/product/class/product.class.php';
```
eldy: There is no reason to have loading this class here.
Code is generic code that does not depends on module. May be you must introduce the require at a different place nearest than code that need it...

PR #23307, opened 2022-12-21
Fix v17 White bg-color for chart and badge in darkmode

daraelmin:
use transparency instead of white background

Befor PR
image

After PR
image


eldy:
And if you are not using the dark mode ? Still ok ?

daraelmin:
Hi,

Seems OK to me, but, I've only tested on two browser (Bromite and Firefox). I can try on more browser this morning, but the color is set to transparent, so... 

IMG_20221222_072646

Cheers

PR #23316, opened 2022-12-22
Fix: Remove Replenishment with "0" Value on "to order"

Humml87:
Now a row for a "replenishmet" with a value "0" on "To Order" not creating.

As example a 0 can a result from "STOCK_REPLENISH_ADD_CHECKBOX_INCLUDE_DRAFT_ORDER" with active checkbox "&draftchecked".


htdocs/product/stock/replenish.php
```diff
@@ -875,6 +875,8 @@
 			$stocktobuywarehouse = max(max($desiredstockwarehouse, $alertstockwarehouse) - $stockwarehouse, 0); //ordered is already in $stock in virtual mode
 		}
 
+		$toOrder = ((!empty($conf->global->STOCK_ALLOW_ADD_LIMIT_STOCK_BY_WAREHOUSE) && $fk_entrepot > 0) > 0 ? $stocktobuywarehouse : $stocktobuy);
```
eldy, line 878: (!empty($conf->global->STOCK_ALLOW_ADD_LIMIT_STOCK_BY_WAREHOUSE) && $fk_entrepot > 0) is a boolen.
But you compare it with > 0 
Surely a bug.

Humml87: It was a c&p fail, sorry. I changed this now to a clean condition.

Humml87:
I make a review on this PR. I think it is better to filter the useless replenishments with the initial SQL query.
Currently we have also the problem that the "Replensishment Counter" not correct. See the following screenshot.
image
The query include the services also, if the option "STOCK_SUPPORTS_SERVICES" not set. So the counter is not correct.
image

eldy:
I can't reproduce the trouble. Do you know how to create data to reproduce the bug ?

Humml87:
For the first Problem "0 Value":
Step 1:
image
Step 2:
image
Step 3:
image

Humml87:
For the second Problem "Wrong Counter" (option "STOCK_SUPPORTS_SERVICES" **not** set):
Step 1:
image
Step 2:
image
Step 3:
image
Step 4:
image

Humml87:
@eldy two examples to reproduce the troubles.

Humml87:
@eldy can you merge/review? I like to start with the next Issues (#23446) which builds on this PR.
Thank you!

eldy:
> For the first Problem "0 Value":
> Step 1:
> image
> Step 2:
> image
> Step 3:
> image
> 

This is the expected behaviour to have 0 here. The suggested value to order is based on the "desired stock", not on the "limit for alert". Limit for alert is just to say when we want a warning. Desired stock is to say value to reach for reassort. On your example, value of desired stock is not defined, so suggested amount to order is 0 (no target defined = no suggestion to reach target)

Humml87:
We talk about different situations.
In the shown example, i have set the option **"include also draft orderes"**. So the proposal to replenish the product is not right.
I hope i can show you the problem with a few pictures:
image
image

image

The "Desired Stock" is not important on this. If the option "include also draft orders" is enabled then we should pretend the order is ordered.

> _**"On your example, value of desired stock is not defined, so suggested amount to order is 0 (no target defined = no suggestion to reach target)"**_

=> Currently, if no "Desired Stock" is set, so the replenishment make a proposal to stock up to "Limit for alert". I think this is correct.

Best regards

eldy:
The "limit for alert" must be used on for alert.
The "desired stock" is used for reassort and only reassort.
So if Desired stock is not defined, the expected behaviour is to suggest 0. End user has defined "No target" for stock reassort so no suggestion during reassort must be done.
Including draft order or not will and must not change this. This is the user choice to not have reassort target to not include product into reassort process. So event if there is draft order, if "desired stock" is empty, no order will be suggested. 
Including draft order may change the quantity to reach the target if a target has been defined.

I did not succeed in reproducing a use case where this expected behaviour is not working.

Humml87:
> So if Desired stock is not defined, the expected behaviour is to suggest 0. End user has defined "No target" for stock reassort so no suggestion during reassort must be done.

The Problem with replenishment lines with a "0" value is the workflow to create a order look like this:
image
The User must always disable all "0" value lines to create an order. And the clarity with many of "0" value lines is not good (we have hundreds of this). I think no User needs a replenishment line with "0".

Best regards Christian

eldy:
The message "Some orders haven't been created because of too-lo quantities" is not related with the line to order at 0.
When "to order" is 0 (for example when desired stock for reassort is empty on product), nothiing shoul dbe created for this line.
The message seems related to other line. dolibarr try to create orders but you probably have the option to disallow creation of order if quantity is not enough (In setup of module "Stock")

Humml87:
> This is the expected behaviour to have 0 here.

Hi @eldy, sorry for disturb you for this topic again.
I become some feedback from my purchasing department. please view the following screenshot. We have in this case 199 open replenishment's. Much replenishment's of there are ordered already, but still in the list.
It's really uncomfortable to create orders for the replenishment's, where "To order" is grather then 0.

image
I hope you can understand our problem.

What do you think of the idea to implement a new setting in the module "Stock" to hide lines with "To order" are 0?
image

Best regards
Christian

eldy:
Thanks for your screenshot. Things are clearer...
So we need something to reduce the list. There may be other reason to have suggested value of 0, but for your need the noise is coming from product already ordered that does not need more order.
So I think the label of option could be 
"Hide lines with a stock already higher than the expected qty"
So you can filter lines with virtual stock > desired stock.
I don't think this must be into the setup but this option, that is filter, should be on the replenishment page.

PR #23321, opened 2022-12-23
NEW Hook in product title pop-up

PaulLepont:
# NEW|New Hook into getNomUrl function in product.class.php

Add new hook in this function to catch and update data displaying.

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -4990,7 +4990,6 @@ class Product extends CommonObject
 			$tmpphoto = $this->show_photos('product', $conf->product->multidir_output[$this->entity], 1, 1, 0, 0, 0, 80);
 			if ($this->nbphoto > 0) {
 				$contentinfo['picture'] = '<div class="photointooltip">'.$tmpphoto.'</div>';
-
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


eldy:
We already have the hook getNomUrl into the method getNomUrl() to allow a module to customize the tooltip.
Even, if you hook makes works for module developer easier, we can't add 1 hook every ten lines of code. Having 1 hook for 1 rendering content should be enough.
Structuring data to make reuse by a hook is interesting but this must be implemented into the existing hook 
getNomUrl to avoid to duplicate hooks call for the same need.

PaulLepont:
Hello Eldy,

Yes, this hook has already been very useful to us, the problem was that to add elements in the tooltips is not a problem, it was rather to remove them.

But no worries.

Thank you for the answer.

eldy:
> Hello Eldy,
> 
> Yes, this hook has already been very useful to us, the problem was that to add elements in the tooltips is not a problem, it was rather to remove them.
> 
> But no worries.
> 
> Thank you for the answer.

Why not setting your contentinfo array as you did (without changing existing code) , and just propagate it into the parameters array to the existing hook getNomUrl() ?
So, you can get structured data to reforge the output string in getNomUrl ?

PR #23322, opened 2022-12-23
FIX : sender holiday email cannot be MAIN_MAIL_EMAIL_FROM with multi …

atm-gauthier:
…smtp

### DIFF
--- a/htdocs/holiday/card.php
+++ b/htdocs/holiday/card.php
@@ -445,8 +445,8 @@ if (empty($reshook)) {
 				// From
 				$expediteur = new User($db);
 				$expediteur->fetch($object->fk_user);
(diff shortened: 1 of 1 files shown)


atm-gauthier:
@eldy i think it can never work with MAIN_MAIL_EMAIL_FROM with multi smtp, we get an error like "a@soc.fr not allowed to send as contact@soc.fr"

eldy:
The email sent is an automatic email. With a generic content from a template.
So sent by dolibarr computer. It must be sent from the email for automated emails so noreply@mycompany.com
What is goal of having the user as sender for an automated email ?
Note that a lot of user have an account and no email defined.

Using multi smtp is a bad practice but for the case it is used, we may add a double check to use the user email only if user email is defined and if the multismtp is on

eldy:

PR #23324, opened 2022-12-23
add optional setup to upload tos files

frederic34:
image


eldy:
Can you explain me how to use fileupload.class.php ? Can't find usage in dolibarr core.
What is its goal and how to use it ? May be adding a proto page into public/test/test_upload.php ?
@Hystepik is working on a solution for a better experience upload, may be this file can become part of solution ? What is it, how it work ?

frederic34:
fileupload.class.php is not changed (whitespaces changes only)
the goal of this change is to upload a pdf (terms of service) and to be able to specify the language of the uploaded file

PR #23327, opened 2022-12-23
WIP add nb of active/inactive tag with CATEGORY_SHOW_COUNTS

daraelmin:
Befor PR
image

After PR
image

### DIFF
--- a/htdocs/categories/class/categorie.class.php
+++ b/htdocs/categories/class/categorie.class.php
@@ -2085,12 +2085,13 @@ class Categorie extends CommonObject
 	}
 
 	/**
-	 *      
 	 *
-	 *      @return array 
+	 *
+	 *      @return array
 	 */
-	public function getLibStatutObject($status_id, $type = 'product', $mode=0, $moreparam=array()) {
-	
+	public function getLibStatutObject($status_id, $type = 'product', $mode = 0, $moreparam = array())
+	{
+
 		global $user;
 
 		$objs = array();
--- a/htdocs/categories/class/categorie.class.php
+++ b/htdocs/categories/class/categorie.class.php
@@ -119,6 +119,7 @@ class Categorie extends CommonObject
 		'supplier' => 'soc',
 		'contact'  => 'socpeople',
 		'bank_account' => 'account',
+		//'bank_line'  => 'fk_account',
 	);
 
 	/**
@@ -130,6 +131,7 @@ class Categorie extends CommonObject
 		'customer' => 'societe',
 		'supplier' => 'fournisseur',
 		'bank_account'=> 'account',
+		//'bank_line'  => 'bank',
 	);
 
 	/**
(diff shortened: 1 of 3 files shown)


htdocs/categories/class/categorie.class.php
```diff
@@ -856,7 +856,7 @@ public function getObjectsInCateg($type, $onlyids = 0, $limit = 0, $offset = 0,
 		$classnameforobj = $this->MAP_OBJ_CLASS[$type];
 		$obj = new $classnameforobj($this->db);
 
-		$sql = "SELECT c.fk_".(empty($this->MAP_CAT_FK[$type]) ? $type : $this->MAP_CAT_FK[$type]);
+		$sql = "SELECT c.fk_".(empty($this->MAP_CAT_FK[$type]) ? $type : $this->MAP_CAT_FK[$type]).", o.*";
```
eldy, line 859: Usage of * for fields is forbidden.
Please use explicit column name

htdocs/categories/index.php
```diff
@@ -177,6 +177,22 @@
 // Define data (format for treeview)
 $data = array();
 $data[] = array('rowid'=>0, 'fk_menu'=>-1, 'title'=>"racine", 'mainmenu'=>'', 'leftmenu'=>'', 'fk_mainmenu'=>'', 'fk_leftmenu'=>'');
+
+if (!empty($conf->global->CATEGORY_SHOW_COUNTS)) {
+	$entry = '<table class="nobordernopadding centpercent">';
+	$entry .=  '<tr class="" >';
+	$entry .=  '<td class=*left* >'.$langs->trans("Label").'</td>';
+	$entry .=  '<td class="left" width="40px">'.$langs->trans("Active").'</td>';
```
eldy, line 185: What mean a nb of "active" ?

htdocs/categories/class/categorie.class.php
daraelmin: As there is no real standard between all type of objetc that can be categorised to define the status of the object, I 've used o.* to avoir SQL error. 
For the moment, I've found "statut", " "status" and "clos" in dôme object, but I have renonced trying to map all table, since this function may be used by external module...

htdocs/categories/index.php
daraelmin: Depending of élément, a statut or a status can have different values, how far i know dolibarr, only the object with a status more than 0 are open (active), instead of draft, resiliated or closed

for me it makes no sense to count objects that have been closed for years or are still at the draft stage. Even if the tag is still on the object.

As exemple, we use tag to know in witch section is one member. Some of them have resiliated for 10 years. When we display the count of member per section, it make no sens to count these member.

But, as the function should be able to read in all object on witch we can use a tag, I was not sur if there was a way to return the status type for each différent situation and translation.

PR #23335, opened 2022-12-25
FIX #23334 Price equivalent for minimum price check

### LINKED ISSUE #23334: Modify/add a line in an invoice with a price in foreign currency
### Bug

When modifying/adding a line in an invoice with a price in foreign currency, php code fails

in file htdocs/commande/card.php (line 947 -add- and 1072 -update-)
```
if (((!empty($conf->global->MAIN_USE_ADVANCED_PERMS) && empty($user->rights->produit->ignore_price_min_advance)) || empty($conf->global->MAIN_USE_ADVANCED_PERMS)) && (!empty($price_min) && (price2num($pu_ht) * (1 - price2num($remise_percent) / 100) < price2num($price_min)))) {
                                $mesg = $langs->trans("CantBeLessThanMinPrice", price(price2num($price_min, 'MU'), 0, $langs, 0, 0, - 1, $conf->currency));
```
Condition fails because $pu_ht is not defined

A solution would be to define $pu_ht_equi that would be $pu_ht if defined and $pu_ht_devise if defined
My problem is I don't see where is currency rate

Many thanks

### Environment Version

16.0.1

### Environment OS

ubuntu

### Environment Web server

apache 2.4

### Environment PHP

php8.0

### Environment Database

mariadb

### Environment URL(s)

htdocs/commande/card.php (line 947 -add- and 1072 -update-)

_No response_

_No response_

### Attached files

_No response_

Amael-PE:
### FIX #23334 Price equivalent for minimum price check
Implement a price equivalent when dealing with minimum prices and foreign currencies. 

**Bug**
As reported in #23334, if a product line is saved filling a foreign currency field and a minimum price is set for the service, the local currency field is null, and minimum price check raise an error. Indeed, the check is done on local currency field, which does not exist.

**Solution**
To avoid this, I create a price equivalent. This is the price in local currency, take as it or computed from the foreign currency price.
This price is then used in the check, so it can make the check work and fix the bug.

**Fix**
Fix added in:
- `command/card.php` for add and update ;
- `comm/propal/card.php` for add and update.
Might need to add it somewhere else... but I didn't find another place.


eldy:
Even if this is a bug, this may change the behaviour on current working process.
So can you publish this PR into the develop or v17 branch ?

Amael-PE:
For sure !

Amael-PE:
@eldy I just finished the merge and passed the tests.
Okay for you ?

Issue #23364, opened 2022-12-29
Incorrect Virtual Stock - Not impacted by Draft Orders


GuillaumeMARQUES:
### Bug

In stock management, the virtual stock is not impacted by draft orders.

### Environment Version

16.0.3

### Environment OS

WIndows 2012 & 2019 / Debian /

### Environment Web server

IIS / Apache

### Environment PHP

No impact

### Environment Database

No impact

PR #23366, opened 2022-12-29
Update product.class.php change to correct BUG #23364

GuillaumeMARQUES:
Correct virtual stock calculation as explained in #23364

# FIX|Fix #[*23364 Virtual Stock calculation not ok*]
[*Virtual Stock is not calculated correctly with draft order informations
I add the draft order quantity decreasing the virtual stock. *]

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -5517,6 +5517,7 @@ class Product extends CommonObject
 		$stock_sending_client = 0;
 		$stock_reception_fournisseur = 0;
 		$stock_inproduction = 0;
+		$stock_commande_draft_client = 0;
 
 		//dol_syslog("load_virtual_stock");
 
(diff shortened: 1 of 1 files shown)


josett225:
Hi There

Not sure we understand why the virtual stock has to be by default impacted by a draft order.
In our side I have several customers with e-commerce or shops that could be impacted negatively by this behavior. They are preparing Orders in draft mode until a final confirmation is received or to mass advance orders preparation or for sales testing purpose or by error. In this case the virtual stock cannot be linked with draft orders as it has a direct negative business impact. 
I believe a solution could be to add a const to keep it optional. 
Other option is to use the hook feature. 
Let us know if we miss understood this request. 
Regards.

eldy:
A "draft" order is a draft that dos not exist yet. As if it was not yet entered into the system.
So it must have no effect anywhere on counting. In dolibarr, draft means "does not exists yet, still in temporary state, will exists as soon as status is "validated", that mean draft is finished)
If you really need to have draft order included, you can add a constant that define the list if status that will be included for virtual stock calculation.
We used SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK for this purpose for purchase order so i suggest to use ORDER_STATUS_FOR_VIRTUAL_STOCK for sales order. Value of this constant will be list of status, for example "1,2,3" to use in the where on the sql filter.

GuillaumeMARQUES:
So when we choose a "STOCK_CALCULATE_ON_VALIDATE_ORDER" could you explain me the code on product class:

$this->stock_theorique += 0;

And I don't understand the difference between physical and virtual stock...

eldy:
Real stock is stock you have NOW.
Virtual stock is stock you will have when all open orders, purchase orders, manufacturing order, shipment will be processed. So it is a vision in future.

PR #23369, opened 2022-12-29
Fix: Can't show MO if no thirdparty is defined.

Humml87:
Now the "MO" are showing. Also the thirdparty is editable.
image


htdocs/mrp/mo_card.php
```diff
@@ -183,7 +183,23 @@
...
+
+		$res = $db->query($sql);
+
+		if (!$res) {
+			$db->rollback();
+		} else {
+			$db->commit();
+		}
+
+		header('Location: '.$_SERVER['PHP_SELF'].'?id='.$object->id);
```
eldy, line 202: After a header, we must always have an exit();

PR #23380, opened 2022-12-30
FIx v14 graph stat subscrition by month

daraelmin:
The date of creation seems more important than the date of start of the subscription

Befor PR
image

After PR
image


eldy:
The statistics is for the subscription not for members, so expected date is the date of subscription.

daraelmin:
I don'tunderstand your comment since dateadh and datec concernant both the subscription and not the membre.

dateadh is the first date of validity of the membership
datec is the date of création witch is the payement date of the fee (there is no subscription status as draft or unpaid)

So, if the subscription is due as long as the member does not terminate and/or according to the date of signature of the contract and/or a date fixed at the commitment (the majority of associations that I know), or more simply in the event of automatic renewal (expired subscription) datec (creation of subscription = date of payment) is more what we need than dateadh (start date of membership).

eldy:
Yes datadh is a date when the subscription starts.
And report is also a report about the list of subscription by date. This is why we expect dateadh to be used. We want to see stats of subscriptuon, when subscription start.
In your case it seems you start you subscription always in january ? So in such a case a report on date of creation is better, but this is a different need, so must be an option.

daraelmin:
The real question to ask is :

In which case dateadh gives really relevant info (different from datec) ? 

Never, I think, because :
- either the membership is renewed (online, most often) from the end date of the last contribution and therefore does not change from one year to another;
- either the membership begins with the day of creation of the contribution and, then, the date of creation is equal to the date of beginning of membership;
- either membership leaves on a fixed date (as in my case) and then, what may be interesting, is the date of creation.

but hey, I'll suggest something with option soon.

PR #23382, opened 2022-12-31
expected type object found null

grandoc:


htdocs/main.inc.php
```diff
@@ -1078,7 +1078,9 @@ function analyseVarsForSqlAndScriptsInjection(&$var, $type)
 
 		dol_syslog("This is a new started user session. _SESSION['dol_login']=".$_SESSION["dol_login"]." Session id=".session_id());
 
-		$db->begin();
+		if (is_object($db)) {
```
eldy, line 1081: In which case do you reach this point with no $db defined ?

PR #23390, opened 2023-01-02
New search on time spent duration range

atm-gauthier:


eldy:
Can you paste a screenshot after the change ?

atm-gauthier:
@eldy yes !

image

PR #23403, opened 2023-01-03
Fix v14 fatal error undefined fct lenght_accounta

daraelmin:
See https://www.dolibarr.fr/forum/t/erreur-fatal-paiements-divers-v15-0-3/41914


htdocs/compta/bank/various_payment/list.php
```diff
@@ -660,6 +660,7 @@
 
 		// Accounting subledger account
 		if ($arrayfields['subledger']['checked']) {
+			require_once(DOL_DOCUMENT_ROOT.'/core/lib/accounting.lib.php';
```
eldy, line 663: Pb in parenrhesis.

PR #23412, opened 2023-01-04
 FIX #23408 Option to display tasks files in project doc page

### LINKED ISSUE #23408: Option to display tasks files in project doc page
### Feature Request

In project documents page, only the docs directly linked to the project are listed

Sometimes (a client asked it to me ...)  it would be interesting to display in this page also the files linked to the tasks of the project

### Use case

_No response_

### Suggested implementation

It's very simple to do 
Add an optionnal const PROJECT_LIST_RECURSIVE_FILES

line ~#124 of /htdocs/projet/document.php
```
// Files list constructor
$filearray = dol_dir_list($upload_dir, "files", 0, '', '(\.meta|_preview.*\.png)$', $sortfield, (strtolower($sortorder) == 'desc' ?SORT_DESC:SORT_ASC), 1);
```
to be replaced by
`$filearray = dol_dir_list($upload_dir, "files", !empty($conf->global->PROJECT_LIST_RECURSIVE_FILES), '', '(\.meta|_preview.*\.png)$', $sortfield, (strtolower($sortorder) == 'desc' ?SORT_DESC:SORT_ASC), 1);
`

### Suggested steps

_No response_

vmaury:
FIX #23408 Option to display tasks files in project doc page
Feature Request

In project documents page, only the docs directly linked to the project are listed

Sometimes (a client asked it to me ...) it would be interesting to display in this page also the files linked to the tasks of the proje

### DIFF
--- a/htdocs/societe/list.php
+++ b/htdocs/societe/list.php
@@ -1826,7 +1826,7 @@ while ($i < $imaxinloop) {
 
 		if (!empty($arrayfields['s.fk_prospectlevel']['checked'])) {
 			// Prospect level
-			print '<td class="center">';
+			print '<td class="center nowraponall">';
 			print $companystatic->getLibProspLevel();
 			print "</td>";
 			if (!$i) {
@@ -1836,7 +1836,7 @@ while ($i < $imaxinloop) {
 
 		if (!empty($arrayfields['s.fk_stcomm']['checked'])) {
 			// Prospect status
-			print '<td class="center nowrap">';
+			print '<td class="center nowraponall">';
 
 			$prospectid = $obj->rowid;
 			$statusprospect = $obj->stcomm_id;


htdocs/core/class/extrafields.class.php
```diff
@@ -1254,7 +1254,7 @@ public function showInputField($key, $value, $moreparam = '', $keysuffix = '', $
 					dol_syslog(get_class($this).'::showInputField type=sellist', LOG_DEBUG);
 					$resql = $this->db->query($sql);
 					if ($resql) {
-						$out .= '<option value="0">&nbsp;</option>';
+						if (!$required) $out .= '<option value="0">&nbsp;</option>';
```
eldy, line 1257: If tou don't add the zmpty value when field is required, it means the system will auto select the first value in list. This will save one click, avoinding user to select a value, but will being a lot of problem of data not correctly set or corrupted because value was set randomly (first found) instead of explicitely selected by end user.
This is why we must keep the empty value. The mandatory status will be blocked later if user does not set explicitely a value.

PR #23416, opened 2023-01-04
Fix: #23415 REST BOM API - It's able to have a wrong value on the 'ref' Field

### LINKED ISSUE #23415: REST BOM API - It's able to have a wrong value on the 'ref' Field
### Bug

It's able to have a wrong value on the 'ref' Field by adding or updating a BOM.

### Environment Version

18

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

PHP 8.1.10

### Environment Database

_No response_

### Environment URL(s)

_No response_

**Actual behavior / Problems:**
- **If you update a BOM with the API you can change the the status from DRAFT to a other Status, with the same ref e.g. '(PROV54)'**
- Currently the 'ref' Field is required. So you can not use the Dolibarr NextNum Generator.

**Expected behavior**
- If no 'ref' is set on create and the status is 0 (DRAFT) then create a ref '(PROF...)'.
- If no 'ref' is set on create and the status is grater then > 0, then create a 'ref' with number generator.
- If 'ref' is set, use the value from the REST-Call.

_No response_

### Attached files

_No response_

Humml87:
Fix: #23415

### DIFF
--- a/htdocs/bom/class/api_boms.class.php
+++ b/htdocs/bom/class/api_boms.class.php
@@ -554,8 +554,8 @@ class Boms extends DolibarrApi
 			throw new RestException(400, "Wrong naming scheme '(PROV%)' is only allowed on 'DRAFT' status. For automatic increment use 'auto' on the 'ref' field.");
 		}
 
-		if ($this->bom->ref == 'auto'){
-			if (empty($this->bom->id) && $this->bom->status == 0){
+		if ($this->bom->ref == 'auto') {
+			if (empty($this->bom->id) && $this->bom->status == 0) {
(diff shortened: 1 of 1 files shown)


htdocs/bom/class/api_boms.class.php
```diff
@@ -536,4 +541,19 @@ private function _validate($data)
 		}
 		return $myobject;
 	}
+
+	/**
+	 * Validate the ref field and get the next Number if it's necessary.
+	 *
+	 * @return void
+	 */
+	private function CheckRefNumbering(): void
```
eldy, line 550: Rename method CheckRefNumbering into checkRefNumbering

Humml87:
@eldy I think the PR is fine now?

PR #23440, opened 2023-01-07
NEW Accountancy - Add an option to disable autolettering when transferring to accounting

aspangaro:
Auto lettering is enabled by default when transferring to accounting. For some needs, add possibility to disable temporarily

### DIFF
--- a/htdocs/accountancy/admin/index.php
+++ b/htdocs/accountancy/admin/index.php
@@ -261,6 +261,20 @@ if ($action == 'setenablelettering') {
 	}
 }
 
+if ($action == 'setenableautolettering') {
+	$setenableautolettering = GETPOST('value', 'int');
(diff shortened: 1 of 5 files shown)


aspangaro:
Ping @josett225

josett225:
Many Thanks @aspangaro 
I will backport it to v14.0.5

htdocs/accountancy/admin/index.php
```diff
@@ -499,6 +514,19 @@
 }
 print '</tr>';
 
+print '<tr class="oddeven">';
```
eldy, line 517: I think this part of setup must be visible only if leterring is on ? If lettering is off, autolettering option has non sense, right ?

aspangaro: You're right @eldy, I push a condition to see option

PR #23444, opened 2023-01-07
Fix php 8 errors

ptibogxiv:


htdocs/core/actions_sendmails.inc.php
```diff
@@ -185,12 +185,12 @@
 		}
 
 		$tmparray = array();
-		if (trim($_POST['sendto'])) {
+		if (isset($_POST['sendto'])) {
```
eldy, line 188: I think isset is not enough. We must enter the if if value us not empty too   Why not using

 if GETPOST

ptibogxiv: i post a fix

eldy:
Has been replaced by another PR

PR #23459, opened 2023-01-08
Fix #23108: Set NULL on fk_warehouse if stockmovement is empty.

### LINKED ISSUE #23108: Bug: Consuming a Service or a Product with activated flag "Stock change disabled" in a MO
### Bug

If consuming a Service or a Product with activated flag "Stock change disabled" in a MO, a error was thrown.

> Cannot add or update a child row: a foreign key constraint fails (`dolibarr_develop`.`llx_mrp_production`, CONSTRAINT `fk_mrp_production_stock_movement` FOREIGN KEY (`fk_stock_movement`) REFERENCES `llx_stock_mouvement` (`rowid`))

Also the Services must split from the "Consumption" to "Consumption Services" similar to BOM View. It´s not right to see the same fields how "Stock" or "Warehouse" as a Product.
image

### Environment Version

17.0.0-beta

### Environment OS

Windows 10

### Environment Web server

xampp

### Environment PHP

PHP 8.1.10

### Environment Database

mariaDB

### Environment URL(s)

_No response_

_No response_

image

### Attached files

_No response_

Humml87:
Fix #23108

### DIFF
--- a/htdocs/mrp/mo_production.php
+++ b/htdocs/mrp/mo_production.php
@@ -1,5 +1,6 @@
 <?php
-/* Copyright (C) 2019-2020 Laurent Destailleur  <eldy@users.sourceforge.net>
+/* Copyright (C) 2019-2020 	Laurent Destailleur  <eldy@users.sourceforge.net>
+/* Copyright (C) 2023 		Christian Humpel  <christian.humpel@gmail.com>
  *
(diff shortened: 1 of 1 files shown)


Humml87:
@eldy i think this you can also pull in branch v17?

PR #23460, opened 2023-01-09
Fix EAN13 prefix available for internal use

Max13:
[GS1](https://www.gs1.org/prefixes) only allows prefixes 040-049 for internal use within a company, while the often mistakenly 020-029 prefixes are restricted.


Max13:
Hi @eldy, as it's a "minor" bug not affecting usability but potentially causing collision with existing barcodes, should I also commit a fix on N-1 and N-2 releases ?

eldy:
> Hi @eldy, as it's a "minor" bug not affecting usability but potentially causing collision with existing barcodes, should I also commit a fix on N-1 and N-2 releases ?

No. Even if situation is not good as it is in n-1 and n-2, versions are already used this way. Here the fix change the behaviour (numbering will be different), and must not change how app works when providing a maintenance release, so we must consider this bug as "won't be fixed" into n-1 and -2. It may solve pb but may bring other. In such a case, status quo is prefered

PR #23463, opened 2023-01-09
 NEW Add supplier in getNomURL and getNomURL in linked supplier order

BenjaminFlr:
NEW

Capture d’écran 2023-01-09 à 10 34 27

I added the supplier name in getNomURL for supplier orders and also added getNomUrl in the tpl for linked supplier order.


htdocs/fourn/class/fournisseur.commande.class.php
```diff
@@ -837,6 +838,9 @@ public function getNomUrl($withpicto = 0, $option = '', $notooltip = 0, $save_la
 
 		$label = '';
 
+		$fourn = new Societe($this->db);
```
eldy, line 841: A getNomUrl method should not contains any subrequests. The method is often used to output long list and introducing a subrequest into a formating method will lead to very bad performances.
You can add the supplier, but the information must already be saved inside the $object (so into $object->thirdparty->name must be set by caller)

BenjaminFlr:
Is it better this way @eldy ?

htdocs/fourn/class/fournisseur.commande.class.php
```diff
@@ -382,12 +384,16 @@ public function fetch($id, $ref = '')
 				return 0;
 			}
 
+			$thirdparty = new Societe($this->db);
+			$thirdparty->fetch($obj->fk_soc);
```
eldy, line 388: The fetch of child data must not be introduced into the fetch of an object.
You must call 

$object->fetch_thirdparties()
just after the $object->fetch()

So only the caller that need to have thirdparties loaded will load them.

xakan:
Hi @eldy, just made the changes. Is it better now? Thanks

BenjaminFlr:
Hi @eldy, I just made the changes. Is it better now? Thanks

eldy:
No the fetch_thirdparty has been included insude the getNomUrl.
No database access is allowed into this method getNomUrl that is a rendering method that must forge html output from data already set into the properties of the object. Loading properties must be done by the caller. If object is loaded once by a fetch, the fetch_thirdparty can be called just after. If object is generatded inside a loop fir a list, properties must be loaded by a join or left join in the sql request of the list page.

BenjaminFlr:
Hi @eldy , thanks for these explanations. I now get the thirdparty and its information from the commonobject.class.php, and not in the getNomURL.

BenjaminFlr:
Hi @eldy , any update here?

htdocs/core/class/commonobject.class.php
```diff
@@ -4129,6 +4130,9 @@ public function fetchObjectLinked($sourceid = null, $sourcetype = '', $targetid
 									$object = new $classname($this->db);
 									$ret = $object->fetch($objectid);
 									if ($ret >= 0) {
+										if (method_exists($object, 'fetch_thirdparty')) {
+											$object->fetch_thirdparty();
```
eldy, line 4134: If you add the fetch_thirdparty here, it means all the code calling the fetchObjectLinked will generate an extra load of data even for caller that don't need it. Please move the fetch_thirdparty at a higher level (so the caller of fetchObjectLinked or the caller of caller, ...)

frederic34:
@altairis-benjamin look at this PR https://github.com/Dolibarr/dolibarr/pull/24198

BenjaminFlr:
I don't see where the fetch_thirdparty could be loaded, as the object is fetched just two lines above.
Would a constant (like MAIN_FETCH_OBJECTLINKED_ADD_FETCH_THIRDPARTY) be a solution here?

PR #23478, opened 2023-01-10
NEW VAT - Add type

aspangaro:
Republish this PRs after a merge problem : (See for discussion / pictures)

https://github.com/Dolibarr/dolibarr/pull/22823


htdocs/product/price.php
```diff
@@ -141,11 +141,12 @@
 			// We look into database using code (we can't use get_localtax() because it depends on buyer that is not known). Same in create product.
 			$vatratecode = $reg[1];
 			// Get record from code
-			$sql = "SELECT t.rowid, t.code, t.recuperableonly, t.localtax1, t.localtax2, t.localtax1_type, t.localtax2_type";
+			$sql = "SELECT t.rowid, t.type_vat, t.code, t.recuperableonly, t.localtax1, t.localtax2, t.localtax1_type, t.localtax2_type";
 			$sql .= " FROM ".MAIN_DB_PREFIX."c_tva as t, ".MAIN_DB_PREFIX."c_country as c";
 			$sql .= " WHERE t.fk_pays = c.rowid AND c.code = '".$db->escape($mysoc->country_code)."'";
 			$sql .= " AND t.taux = ".((float) $tva_tx)." AND t.active = 1";
 			$sql .= " AND t.code = '".$db->escape($vatratecode)."'";
+			$sql .= " AND t.type_vat = 1";
```
eldy, line 149: Why this test ?
We alreayd have the rate+code that make the vat rate unique.

aspangaro: Because we only want the type 1 VAT rates, i.e. the sales type VAT rates.

I can remove but we are on the page of customer prices so we only display the VAT rates of the sales type, it is useless to display those of the purchases.

You sell at 20%, you don't want the 10%, 5.5%, 2.1% rates to be displayed.

aspangaro: @eldy I proposed to display type 0 (VAT All type) + type 1 (VAT Sales type)

htdocs/core/class/html.form.class.php
```diff
@@ -6468,7 +6474,7 @@ public function load_tva($htmlname = 'tauxtva', $selectedrate = '', $societe_ven
 		}
 
 		// Now we get list
-		$num = $this->load_cache_vatrates($code_country); // If no vat defined, return -1 with message into this->error
+		$num = $this->load_cache_vatrates($code_country, $type_vat); // If no vat defined, return -1 with message into this->error
```
eldy, line 6477: This load the database into a cache array. We should not filter on type here. Filter should be done later after lin 6532.
If you filter here, when you call the load_tva a second time on same page with a different value for type, you will get result of first call.

PR #23481, opened 2023-01-10
FIX 15.0 missing transactions / triggers

atm-florianm:
- missing sql transaction and trigger in CompanyBankAccount::update()
- missing sql transaction in CompanyBankAccount::create()
- missing override of fields `element` and `table_element` (the values inherited from Account are misleading)

Note: after adding the transaction in `create()`, I did not remove the transactions already present in the processing of action `add` in `paymentmodes.php` because a failure to set the UMR number (line 310) should also roll back the creation.

I did not analyze deeply enough, but maybe a next step would be to set the UMR directly in the `create()` method so that third party bank information created from elsewhere (API etc.) would also benefit from the automatic UMR number ?

### DIFF
--- a/htdocs/societe/class/companybankaccount.class.php
+++ b/htdocs/societe/class/companybankaccount.class.php
@@ -35,6 +35,17 @@ class CompanyBankAccount extends Account
 {
 	public $socid;
 
+	/**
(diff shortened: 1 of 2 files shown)


eldy:
Merged manually into v17

Issue #23493, opened 2023-01-10
Contacts List - show Birthday


dolibit-ut:
### Feature Request

Contacts list add option to show field Birthday

htdocs/contact/list.php

PR #23496, opened 2023-01-10
Update contact.class.php

dolibit-ut:
- 'Remove Roles' part was twice 
- translations and comments

### DIFF
--- a/htdocs/contact/class/contact.class.php
+++ b/htdocs/contact/class/contact.class.php
@@ -167,7 +167,10 @@ class Contact extends CommonObject
 	 */
 	public $state;
 
-	public $poste; // Position
+	/**
+	 * @var string // Job Position
+	 */
+	public $poste;
 
 	/**
 	 * @var int Thirdparty ID
@@ -176,7 +179,7 @@ class Contact extends CommonObject
 	public $fk_soc;		// both socid and fk_soc are used
 
 	/**
-	 * @var int 0=inactive, 1=active
+	 * @var int  0=inactive, 1=active // Status
 	 */
 	public $statut;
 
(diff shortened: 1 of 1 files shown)


dolibit-ut:
I want to change birthday visible state from 3 to 1 

'birthday' =>array('type'=>'date', 'label'=>'Birthday', 'enabled'=>1, 'visible'=>3, 'position'=>80),

see  #23493 

Is there anything that speaks against this?

htdocs/contact/class/contact.class.php
```diff
@@ -167,7 +167,10 @@ class Contact extends CommonObject
 	 */
 	public $state;
 
-	public $poste; // Position
+	/**
+	 * @var string // Job Position
```
eldy, line 171: Avoid use of // when inside /*  */

htdocs/contact/class/contact.class.php
```diff
@@ -176,7 +179,7 @@ class Contact extends CommonObject
 	public $fk_soc;		// both socid and fk_soc are used
 
 	/**
-	 * @var int 0=inactive, 1=active
+	 * @var int  0=inactive, 1=active // Status
```
eldy, line 182: Avoid use of // when inside /*  */

htdocs/contact/class/contact.class.php
```diff
@@ -1283,18 +1295,6 @@ public function delete($notrigger = 0)
 			}
 		}
 
-		if (!$error) {
-			// Remove Roles
-			$sql = "DELETE FROM ".MAIN_DB_PREFIX."societe_contacts WHERE fk_socpeople = ".((int) $this->id);
```
eldy, line 1288: Why removing this ?
If we delete the contact, we must also delete it into this table, right ?

htdocs/contact/class/contact.class.php
dolibit-ut: okay

PR #23497, opened 2023-01-10
Update ChangeLog

dolibit-ut:
Sorting


dolibit-ut:
is this right ?:
***** ChangeLog for 16.0.4 compared to 16.0.2 *****   

or should be
***** ChangeLog for 16.0.4 compared to 16.0.3 *****   
?

PR #23498, opened 2023-01-10
Update 16.0.0-17.0.0.sql

dolibit-ut:
v17 Sorting

ALTER TABLE llx_emailcollector_emailcollector ADD COLUMN port varchar(10) DEFAULT '993';   was twice

### DIFF
--- a/htdocs/install/mysql/migration/16.0.0-17.0.0.sql
+++ b/htdocs/install/mysql/migration/16.0.0-17.0.0.sql
@@ -56,24 +56,25 @@ UPDATE llx_c_actioncomm SET type = 'system' WHERE code = 'AC_OTH';
 
 ALTER TABLE llx_opensurvey_user_studs MODIFY reponses VARCHAR(200) NOT NULL;
 
+
 -- v17
 
 ALTER TABLE llx_mailing_cibles MODIFY COLUMN source_type varchar(32); 
-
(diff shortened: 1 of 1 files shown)


dolibit-ut:
any reason why  `ADD COLUMN ip`  is  `var(250)`   and not var(255) ?

eldy:
> any reason why  `ADD COLUMN ip`  is  `var(250)`   and not var(255) ?

I don't think so.

eldy (commented):
Editing a file used for migration to v16 to v17 is not allowed for version higher than v17. It is only in branch v17.
However, the v17 beta is nearly finished, so no change in this file will not be done except if it fixes a blocking bug.

dolibit-ut:
Okay. I have only sorted the order and so I noticed that this line is duplicated:

ALTER TABLE llx_emailcollector_emailcollector ADD COLUMN port varchar(10) DEFAULT '993';

This should be deleted.

For the future a better order/sorting is recommended.

eldy:
> Okay. I have only sorted the order and so I noticed that this line is duplicated:
> 
> ALTER TABLE llx_emailcollector_emailcollector ADD COLUMN port varchar(10) DEFAULT '993';
> 
> This should be deleted.
> 
> For the future a better order/sorting is recommended.

Don't forget that even a simple sort generates trouble: when doing merge of maintenance releases, changing sort order of lines in files generates troubles and conflict that lead to errors (line forgotten or corrupted) when trying to solve conflict.

PR #23500, opened 2023-01-10
fix properties

frederic34:
`		if (!empty($this->trackid)) { // TODO Always use trackid (ticXXX) instead of track_id (abcd123)
			$keytoavoidconflict = '-'.$this->trackid;
		} else {
			$keytoavoidconflict = empty($this->track_id) ? '' : '-'.$this->track_id;
		}`

### DIFF
--- a/htdocs/core/class/html.formticket.class.php
+++ b/htdocs/core/class/html.formticket.class.php
@@ -50,7 +50,6 @@ class FormTicket
 
 	/**
 	 * @var string	The track_id of the ticket. Used also for the $keytoavoidconflict to name session vars to upload files.
-	 * @deprecated
 	 */
 	public $track_id;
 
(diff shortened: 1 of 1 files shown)


htdocs/core/class/html.formticket.class.php
```diff
@@ -50,9 +50,15 @@ class FormTicket
...
 	/**
 	 * @var string	The track_id of the ticket. Used also for the $keytoavoidconflict to name session vars to upload files.
+	 * @deprecated
 	 */
 	public $track_id;
 
+	/**
+	 * @var string trackid
+	 */
+	public $trackid;
```
eldy, line 60: track_id and trackid ae 2 different vars.
track_id is a column of ticket containing a second id in format of a hash of ticket. 
trackid is the key used for the tracking of emails sents

PR #23508, opened 2023-01-11
fix warnings

frederic34:
->title doesn't exist


htdocs/core/class/commonobject.class.php
```diff
@@ -490,6 +522,18 @@ abstract class CommonObject
 	 */
 	public $name;
 
+	/**
+	 * @var string The name alias
+	 */
+	public $name_alias;
```
eldy, line 528: Thus field is needed only for Societe object, why adding it to all classes ?

eldy: Right but this is a hack to manage the case of Societe objet. So i think a better place would be into the Societe object only. May be a test  if property_exists would be welcome.

Issue #21426, opened 2022-07-01
During the export of a CSV, allow the user to choose between file format ISO-8859-1 or UTF8


eldy:
### Feature Request

There is already an option called EXPORT_CSV_FORCE_CHARSET into the tool that generate the CSV file.
Instead of using a fixed hidden option, we must ask the user with a combo of 2 choices;
UTF-8
ISO-8859-1

For the CSV, ask ask also for the separator: Ask for 
,
;
Must be preselected to the constant EXPORT_CSV_SEPARATOR_TO_USE

PR #23509, opened 2023-01-11
NEW Add possibility to choose format #21426

KiVoilaJe:
NEW|New feature(#21426) Added in the export module the possibility to choose between utf-8 and iso format


htdocs/core/modules/export/export_csvutf8.modules.php
```diff
@@ -0,0 +1,379 @@
...
+
+require_once DOL_DOCUMENT_ROOT.'/core/modules/export/modules_export.php';
+
+// avoid timeout for big export
+set_time_limit(0);
+
+/**
+ *	Class to build export files with format CSV
+ */
+class ExportCsvUtf8 extends ModeleExports
```
eldy, line 32: It seems this class is a copy of the class for iso but with just few lines changed.
So why not don't you make an inherit 
 class ExportCsvUtf8 extends ExportCsv
And implement into class only properties or method with change for utf8 ?

Hystepik:
Still travis error. Try to move the set of the CHARSET... into an surcharging method write_record, ... instead of into constructor.

Hystepik:
Unit test to fix

Hystepik:
@eldy seams ok

PR #23521, opened 2023-01-11
ajout d'un hook pour ajouter des landingpage dans la liste de user_ihm

comaiteseb:
# NEW Adding a hook addToLandingPageList in userihm context
This new hook is used to add new items in the list of possible landing page in the user params ihm.

### DIFF
--- a/htdocs/user/param_ihm.php
+++ b/htdocs/user/param_ihm.php
@@ -216,10 +216,10 @@ if (isModEnabled('ticket')) {
 }
 
 // Hook for insertion new items in the List of possible landing pages
-$reshook = $hookmanager->executeHooks('addToLandingPageList',$tmparray,$object); 
+$reshook = $hookmanager->executeHooks('addToLandingPageList', $tmparray, $object);
 if ($reshook < 0) {
 	setEventMessages($hookmanager->error, $hookmanager->errors, 'errors');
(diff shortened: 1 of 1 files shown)


htdocs/user/param_ihm.php
```diff
@@ -215,6 +215,17 @@
 	$tmparray['ticket/list.php?mainmenu=ticket&leftmenu='] = 'Tickets';
 }
 
+// Hook for insertion new items in the List of possible landing pages
+$reshook = $hookmanager->executeHooks('addToLandingPageList', $tmparray, $object);
+if ($reshook < 0) {
+	setEventMessages($hookmanager->error, $hookmanager->errors, 'errors');
+} elseif ($reshook === 0) {
```
eldy, line 222: You must make a test if 0 or > 0
If > 0, $tmparray=$hookmanager->resArray;
So it replace completely the default tmparray.
If == 0, $tmparray=array_merge($tmparray, $hookmanager->resArray);
So if 0, list is just completed, if >0 it is fully replaced.
Can you enhance this way ?

comaiteseb: I made the correction. 
Is adding a hook considered as a new feature? (wish I could add it on v16 and v17)

PR #23549, opened 2023-01-13
Fix #23548

### LINKED ISSUE #23548: [ Invoice ] Credits can be null
### Feature Request

As Invoice, credits can be null.

### Use case

For exemple, in our case. 
Society create 0€ invoice to move stock, sometime it's necessary for stock correction to create a credits 0€ to adapt stock mouvement.

### Suggested implementation

Modify file https://github.com/Dolibarr/dolibarr/blob/a074c89960b506e60aa2f7b33c5c8717342a6bae/htdocs/compta/facture/card.php#L325

remove equals for result  
`if ($object->total_ht > 0) {`

### Suggested steps

I will create a PR linked if you think this could be usefull for others.

pierrejego:
Invoice/Credit can be null

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -321,8 +321,9 @@ if (empty($reshook)) {
 
 		// On verifie signe facture
 		if ($object->type == Facture::TYPE_CREDIT_NOTE) {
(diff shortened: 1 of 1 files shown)


eldy:
Closed, no answer to the issue.

PR #23550, opened 2023-01-13
NEW kanban for post employees list

lamrani002:
# NEW|New mode view "kanban" for post employee
by adding a getkanbanView in class position


htdocs/hrm/position_list.php
```diff
@@ -602,13 +605,30 @@
...
 			print '<div class="box-flex-container">';
 		}
-		// Output Kanban
-		print $object->getKanbanView('');
+		// get info needed
+
+		$object->date_start = $obj->date_start;
+		$object->date_end = $obj->date_end;
+		$job = new job($db);
+		$job->fetch($obj->fk_job);
```
eldy, line 619: We are inside a loop in a lis (that may be very long). In suche situation , we must never introduce call to a method that make sql subrequests.

If you have 500 lines in screen, it means you will make 501 SQL request instead of 1 increasing time to show the page by a factor  500 !
You must include all properties you need inside the main sql of the list so you have them already known in the result of the fetch. All you have to do is to make the set  userstatic->xxx = $obj->xxx

lamrani002: ok i see i will fix it

htdocs/hrm/position_list.php
```diff
@@ -229,7 +234,7 @@
 $reshook = $hookmanager->executeHooks('printFieldListSelect', $parameters, $object); // Note that $action and $object may have been modified by hook
 $sql .= preg_replace('/^,/', '', $hookmanager->resPrint);
 $sql = preg_replace('/,\s*$/', '', $sql);
-$sql .= " FROM ".MAIN_DB_PREFIX.$object->table_element." as t";
+$sql .= " FROM ".MAIN_DB_PREFIX.$object->table_element." as t, ".MAIN_DB_PREFIX.$userstatic->table_element." as u,".MAIN_DB_PREFIX."hrm_job as j";
```
eldy, line 237: I am not sure, but i think the user is not mandatory, so use a LEFT JOIN instead of strict join on fk_user.

htdocs/hrm/position_list.php
```diff
@@ -218,6 +219,10 @@
 // --------------------------------------------------------------------
 $sql = 'SELECT ';
 $sql .= $object->getFieldList('t');
+$sql .= ',';
+$sql .= $userstatic->getFieldList('u');
+$sql .= ',u.email, u.statut';
```
Hystepik, line 224: to remove because it's in getFieldList

lamrani002: the fields in class user doesn't have an email and statut

PR #23570, opened 2023-01-14
Update ticket.php

dolibit-ut:
Layout - Info column info column standardized - size, headers
& also at *Document templates* as last column, before preview


htdocs/admin/ticket.php
```diff
@@ -358,7 +361,7 @@
 
 
 
-/*
+/********************************
```
eldy, line 364: /*** must be avoid.
Use only 
/*   for any comment
or
/**     for comment on method / class

PR #23571, opened 2023-01-15
Update errors.lang

dolibit-ut:
sort Warnings


eldy:
Waiting for a response or fix for too long, so closed

eldy:
Waiting for a response for too long, so closed

PR #23579, opened 2023-01-16
Update mo_movements.php

dolibit-ut:


eldy:
This PR does not provide fix, neither new feature ? Mass code change for only structure has annoying side effect when trying to merge old version.

PR #23587, opened 2023-01-17
Set MAIN_DO_NOT_SANITIZE_PRODUCT_REF to not sanitize product ref and label on database entry (storage).

marcusdeangabriel:
# NEW
This is a request for enhancement to set MAIN_DO_NOT_SANITIZE_PRODUCT_REF to not sanitize product ref and label values on database entry (storage).

Sanitizing product ref and label entries before storage into the database is not the correct policy for us and possibly others.  For example, we have suppliers who use slashes (/) in references and who have label information with double quotes (") to represent inches.  For instance, stripping this information on database entry yields PDF documents such as supplier orders that have incorrect information from the point of view of our supplier.

We propose a global constant MAIN_DO_NOT_SANITIZE_PRODUCT_REF  to disable this sanitation of product ref and label values on database entry which changes to

- `htdocs/product/card.php`
- `htdocs/product/class/product.class.php`

Our experience indicates that not sanitizing product ref and label values on database entry is not a problem

- Over a year ago, we imported initially via SQL scripts non-sanitized product ref and label information into our production environment without incident of usage since the import.
- We examined Dolibarr code with for example `find htdocs -name \*.php | xargs fgrep -i sanitize | egrep "\$ref|->ref" | egrep -v "dol_sanitizeFileName|sanitizeVal"` indicating that the community for example fetches product ref and label entries, sanitizes them, and then uses them for filesystem pathnames.

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -532,7 +532,7 @@ class Product extends CommonObject
 	{
 		global $conf;
 
-		if (!empty($conf->global->MAIN_DO_NOT_SANITIZE_PRODUCT_REF)) {
+		if (!empty($conf->global->MAIN_SECURITY_ALLOW_UNSECURED_REF_LABELS)) {
 			$this->ref = trim($this->ref);
 		} else {
 			$this->ref = dol_sanitizeFileName(stripslashes($this->ref));
@@ -568,7 +568,7 @@ class Product extends CommonObject
 		$error = 0;
 
 		// Clean parameters
-		if (!empty($conf->global->MAIN_DO_NOT_SANITIZE_PRODUCT_REF)) {
+		if (!empty($conf->global->MAIN_SECURITY_ALLOW_UNSECURED_REF_LABELS)) {
 			$this->ref = trim($this->ref);
 		} else {
 			$this->ref = dol_sanitizeFileName(dol_string_nospecial(trim($this->ref)));
@@ -967,7 +967,7 @@ class Product extends CommonObject
 		}
 
 		// Clean parameters
-		if (!empty($conf->global->MAIN_DO_NOT_SANITIZE_PRODUCT_REF)) {
+		if (!empty($conf->global->MAIN_SECURITY_ALLOW_UNSECURED_REF_LABELS)) {
 			$this->ref = trim($this->ref);
 		} else {
 			$this->ref = dol_string_nospecial(trim($this->ref));
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -530,7 +530,13 @@ class Product extends CommonObject
 	 */
 	public function check()
 	{
-		$this->ref = dol_sanitizeFileName(stripslashes($this->ref));
+        global $conf;
+
+		if (!empty($conf->global->MAIN_DO_NOT_SANITIZE_PRODUCT_REF)) {
+		    $this->ref = trim($this->ref);
+		} else {
+		    $this->ref = dol_sanitizeFileName(stripslashes($this->ref));
+		}
 
 		$err = 0;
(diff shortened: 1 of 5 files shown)


eldy:
Removing sanitization of values leans introduce security holes. There is already an option called MAIN_SECURITY_ALLOW_UNSECURED_LABELS_WITH_HTML
My undertsanding is that you want a different option that may be called
MAIN_SECURITY_ALLOW_UNSECURED_REF_LABELS ?
Can you rename your option with this name so user using it will understand that his instance is no more secured due to easy use  of " that can be used for sql injection.
Also this option must be pushed into develop (v16 was already released a long time ago).

htdocs/product/class/product.class.php
```diff
@@ -562,7 +568,11 @@ public function create($user, $notrigger = 0)
 		$error = 0;
 
 		// Clean parameters
-		$this->ref = dol_sanitizeFileName(dol_string_nospecial(trim($this->ref)));
+		if (!empty($conf->global->MAIN_DO_NOT_SANITIZE_PRODUCT_REF)) {
```
eldy, line 571: Removing sanitization of ref means you can get a / into it. Ref is also used to be the directory name where files are stired. This will break a lot of features.
If you need to store a non sanitized ref to store a value with dangerous characters, you should add an complementary fields called ref_not_sanitized to store it for example.

marcusdeangabriel: > Removing sanitization of ref means you can get a / into it.

En effet, nous avons beaucoup, beaucoup de ref de produit avec / depuis plus qu'un an.  Il y a plusieurs points :

- Nous avons importer tous ces / lorsqu'on a basculé via des scripts SQL à Dolibarr en production sans observer des problèmes d'utilisation.
- Nous avons enlevé sanitization pour des nouvelles créations de produit il y a an sans observer des problèmes.
- Nous avons examiné le code pour des utilisations des ref de produit et nous observons que les programmeurs utilisent les fonctions dol_sanitizeFileName ou sanitizeVal  lorsqu'ils obtiennent un ref de produit de la base pour des noms de fichiers ou dossiers.

> This will break a lot of features.

Apparemment non. Sérieusement, nous avons cherché et nous n'avons trouvé aucun problème.

En ce qui concerne la suggestion MAIN_SECURITY_ALLOW_UNSECURED_REF_LABELS, j'avoue qu'il faut que je réfléchisse avant que je puisse répondre.

marcusdeangabriel: Bonjour,

Avec commit [1e17c43](https://github.com/marcusdeangabriel/no-sanitize-prod-ref/commit/1e17c43e992f66585f11a710d4f03e200234ad9d), j'ai pris en compte votre conseil d'utiilser MAIN_SECURITY_ALLOW_UNSECURED_REF_LABELS au lieu de MAIN_DO_NOT_SANITIZE_PRODUCT_REF.

J'ai corrigé la documentation de la fonction `GETPOST` dans `functions.lib.php`.  `nohtml` laisse passer `"` et `../` mais il était documenté comme `alphanohtml` qui enlève `"` et `../` ce qui m'a dérouté.

marcusdeangabriel: I was preparing newer pull request than this current and and commited [7acef58](https://github.com/marcusdeangabriel/dolibarr/commit/7acef585ff2da7ca272516e4ab91326dbd9d39f2), which provoked the commit [3c556f9](https://github.com/Dolibarr/dolibarr/pull/23587/commits/3c556f9fc060bbf320cbe3541a5c5f426b72bb4d), by mistake.  These two commits should not be included for the current pull request.

marcusdeangabriel:
> Also this option must be pushed into develop (v16 was already released a long time ago).
I will do this.  (My bad for not reading the entire reply.)

PR #23598, opened 2023-01-17
Fix : workflow on bill validate don't close propal when there is a deposit

gdsoftdev:
On bill validate, workflow compares invoice amount and propal amount.
If a deposit is applied on invoice, propal and invoice don't have the same amount and propal is not closed.
We need to list every invoices linked to propals to have the real amount of invoices and compare to propal.

### DIFF
--- a/htdocs/core/triggers/interface_20_modWorkflow_WorkflowManager.class.php
+++ b/htdocs/core/triggers/interface_20_modWorkflow_WorkflowManager.class.php
@@ -168,9 +168,8 @@ class InterfaceWorkflowManager extends DolibarrTriggers
 							$element->fetchObjectLinked('', '', null, $object->element);
 							dol_syslog("Nb invoices linked to propal " . $element->id . " : " . sizeof($element->linkedObjects));
 							if (!empty($element->linkedObjects)) {
-								foreach($element->linkedObjects['facture'] as $invoice) {
-									if (($invoice->statut == Facture::STATUS_VALIDATED || $invoice->statut == Facture::STATUS_CLOSED) && !in_array($invoice->id, $invoiceIds)) 
-									{
+								foreach ($element->linkedObjects['facture'] as $invoice) {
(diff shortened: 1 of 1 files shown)


eldy:
The problem seems more complex. Sometimes (often), we generate 1 invoices for n proposals.
So imagine we have

Proposal A - 100
Proposal B - 100
Proposal C - 100
Invoice X deposit is 25 (for 25 of C) 
Invoice Y is 50 (for 25 A + 25 B)
Invoice Z is 100 (for 50 A + 50 C)

When paying Z of 50, you will find that A is linked to Y and Z, so total of invoices linked to A = 100 that is also amount of A so A is closed, but only 75 is paid for A after paying Z. And invoice A will be closed when it should not.

With current code, when total is same, we are sure we can close, if it differs there is some cases we should close and we don't do it.  But it seems better to forget to close an invoice that should be closed than closing an invoice that should not.

gdsoftdev:
I understand your case. My clients don't often generate invoice for n proposals. They have a proposal for a client/project, make a deposit invoice to start the work and then a balance invoice. So we could limit the proposal closing as I did by checking that there is only one linked proposal. What do you think ?

eldy:
> I understand your case. My clients don't often generate invoice for n proposals. They have a proposal for a client/project, make a deposit invoice to start the work and then a balance invoice. So we could limit the proposal closing as I did by checking that there is only one linked proposal. What do you think ?

This is better, however:
- as any new change in the behavior, it must be pushed on develop branch
- because this is specific to a way a user uses the app and the feature can't be generalized to anybody (or it will introduce a bug), this must be implemented by adding a hidden option   
```
if (getDolGlobalString("WORKFLOW_INVOICE_CLASSIFY_BILLED_PROPAL_IF_ONE_LINK_EVEN_NO_MATCHING_AMOUNT")) {
   // all code specific to manage this at the same place into this if, it overwrite previous decision/test code
}
```

eldy:

PR #23604, opened 2023-01-17
Update products.lang

Quentin-Seekness:
More accurate translation as the Azur pdf model is an old one


eldy:
Changing translation for non us fikes is no useless because files are erased by transifex sync. Please fix language in transifex.com/dolibarr

PR #23618, opened 2023-01-18
FIX : Empty FormSetup emailTemplate type IF empty fieldvalue

atm-GregM:
# FIX -  Empty FormSetup emailTemplate type IF empty fieldvalue

Bug : An email template is selected when the conf value is empty (View mode)


atm-GregM:
@eldy

PR #23628, opened 2023-01-18
deprecate ref_client

frederic34:


htdocs/core/class/commonobject.class.php
```diff
@@ -235,6 +235,11 @@ abstract class CommonObject
 	 */
 	public $ref;
 
+	/**
+	 * @var string The object's old reference before update
+	 */
+	public $oldref;
```
eldy, line 241: May be we can set this as deprecated because now, we should try to store old value using ->oldcopy=old object clone ?

PR #23638, opened 2023-01-19
Clean Societe class code

frederic34:


htdocs/societe/class/companypaymentmode.class.php
```diff
@@ -255,8 +267,7 @@ public function createFromClone(User $user, $fromid)
 
 		// Clear fields
 		$object->ref = "copy_of_".$object->ref;
-		$object->title = $langs->trans("CopyOf")." ".$object->title;
```
eldy, line 258: May be we should use label instead of title ?

htdocs/societe/class/societe.class.php
```diff
@@ -462,6 +485,12 @@ class Societe extends CommonObject
 	 */
 	public $idprof6;
 
+	/**
+	 * Object of company
+	 * @var string
+	 */
+	public $object;
```
eldy, line 492: Objet of company inside company ? Strange

frederic34: https://github.com/Dolibarr/dolibarr/blob/dd4a9a94c7fbb8daa745e3a2ff880f3871be8206/htdocs/societe/class/societe.class.php#L4271

frederic34: https://github.com/Dolibarr/dolibarr/commit/b6e333876ed5fd371e8e51aae8a52585d1bd506c

eldy: Hum i see it is not the "object" in POO sense but into the meaning "Reason/Goal of the company".

PR #23642, opened 2023-01-20
NEW: use account address in sepa mandate

thomas-Ngr:
# NEW use account address in sepa mandate

In SEPA mandate pdf model, the account owner and address are the ones of the thirdparty.
However, users are allowed to define an owner and an address for the account, that are different from the name and address of the thirdparty.

This PR introduces option PRELEVEMENT_PREFER_ACCOUNT_OWNER_AND_ADDRESS that displays account owner + account address instead of thirdparty name + thirdparty address.


eldy:
We don't need to add an option for this. Rules should be:
If we have a precise information on bank account, we must use this in priority. If not, we use a more  generictm value of third party.
To fight against Options inflations, can you simplify the pr 
 this way (so remove the pafe setup of option)

thomas-Ngr:
@eldy I have removed the options.

thomas-Ngr:
@eldy could you have a look at the changes ?

PR #23649, opened 2023-01-21
FIX : Update main.inc.php

cmfpmatik:
Fix everywhere Warning error of PHP and log it in specific file.

### DIFF
--- a/htdocs/main.inc.php
+++ b/htdocs/main.inc.php
@@ -37,16 +37,18 @@
  */
 
 // Log WARNING to trace_warningphp.log and not display it to user
-error_reporting( E_ALL ^ E_NOTICE );
-@set_error_handler("__error_handler" );
-function __error_handler($errno, $errstr, $errfile, $errline){
-    //echo "<pre><br/>ERREUR/WARNING : $errno, $errstr, $errfile, $errline<br/></pre>"; // for display it to user
-   trace("ERREUR/WARNING: $errno, $errstr, $errfile, $errline" );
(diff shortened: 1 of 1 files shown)


eldy:
Note that we already have a tool to track (differently) php warning: enable module debugbar.

cmfpmatik:
> Note that we already have a tool to track (differently) php warning: enable module debugbar.

This fix is not a tool to track php warning, just to isn't displaying them. (safety recommendation)

cmfpmatik:
what's happen about this ?

htdocs/main.inc.php
```diff
@@ -36,6 +36,21 @@
...
+error_reporting(E_ALL ^ E_NOTICE);
+@set_error_handler("__error_handler");
+function __error_handler($errno, $errstr, $errfile, $errline)
+{
+	//echo "<pre><br/>ERREUR/WARNING : $errno, $errstr, $errfile, $errline<br/></pre>"; // for display it to user
+	trace("ERREUR/WARNING: $errno, $errstr, $errfile, $errline");
+}
+function trace($toTrace)
+{
+	$f = fopen("trace_warningphp.log", "a+");
```
eldy, line 49: On a secured installation, the application has no permission to write into a file into htdocs directory, so this won't work.

Also, if creating log, rotation of log must be managed.

Also hidding errors to users with no choice of user is not a good practice. user must choose if he want to not see important information he should see.

You don't need this to report errors onto a log file. You can already use the module "Log debug" but also the module "Debug bar".

Issue #23449, opened 2023-01-08
Discussion / Logical Bug: Limited duration for Services and also for Products


Humml87:
### Feature Request

Can also set a limited duration for Products.

### Use case

In some (in our company in the most cases) business events its necessary to deliver also products to a different date.

### Suggested implementation

![image](https://user-images.githubusercontent.com/78662388/211195541-9a4f06fe-24b6-459e-af87-43feb00ff99e.png)

### Suggested steps

Variant 1:
Remove all if-statements where is requesting "its a service...".

Variant 2:
Integrate a configuration variable to say "PRODUCT_LIMITED_DURATION_FOR_PRODUCTS"

---
@eldy: What do you think is the best implementation?

dolibit-ut:
Hello, this "Limited Duration" feature is actually for subscription services 
and not for a targeted delivery date (If I understand your intention right).

But i think it can is also be useful for products - 
e.g. to map seasonal products, product changes or serials/lots or whatever else.

In my opinion it could be the 'simple' variant 1.

Humml87:
Hi, thank you for the feedback.

> (If I understand your intention right)

You right.

> In my opinion it could be the 'simple' variant 1.

i think this variant 1 is good. I can't found problems for this integration.
Then i do it ;-)

PR #23655, opened 2023-01-22
New #23449 Limited duration (deliverytarget) for services and also for products too

Humml87:
See Feature Request: #23449

Impression:
image
image

### DIFF
--- a/htdocs/core/tpl/objectline_create.tpl.php
+++ b/htdocs/core/tpl/objectline_create.tpl.php
@@ -479,49 +479,49 @@ if ($nolinesbefore) {
 
 <?php
 	print '<tr id="trlinefordates" class="oddeven">'."\n";
-	if (!empty($conf->global->MAIN_VIEW_LINE_NUMBER)) {
-		print '<td></td>';
-	}
+if (!empty($conf->global->MAIN_VIEW_LINE_NUMBER)) {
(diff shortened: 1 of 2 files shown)


eldy:
This looks strange. The differencebetween a product and a service is that when there is a limit duration, the oroduct is in fact a service. So why having limit duration on product ?

Humml87:
Hi @eldy, 
as example, you have a customer order with different products. The customer need the products on different dates.
Or:
You order some products, but you like the products on different dates.

It's better to rename 'Limited duration' to 'Delivery time' if it's a product? 
I hope you understand my mind. 
Best regards!

hregis:
@Humml87 you can add extrafields !

Humml87:
> @Humml87 you can add extrafields !

Hi, I think it is not a good idea. As example we create a report for expected sales volume, then you need the values in the same column.

Please let me know, wath the foreseeable problem of this PR is.

hregis:
@Humml87 you can create a module with custom "/core/tpl/objectline.*.tpl.php" templates

Humml87:
Hi eldy, hi hregis,
now i understand your problem with this PR. Joachim from the german community helping me to understood this.
@hregis thanks for the tips.
Best regard!

PR #23657, opened 2023-01-23
ADD ref_customer for propal in project overview page

iouston:
In propal object ref_customer is ref_client so i added this case


eldy:
A PR was suggested by @frederic34 to depreciate/replace ref_client with ref_customer. I will probly merge it in develop making your PR obsolete...

iouston:
The best solution is that the names are the same between the objects, so it's fine like that.
Thanks

PR #23668, opened 2023-01-23
Protection against deletion of ref_employee and national_registration…

comaiteseb:
# FIX
Protection against deletion of ref_employee and national_registration_number while fields are not presents in the user tab
When you set ref_employee and/or national_registration_number in the RH & bank tab if you edit the user tab and save, the 2 fields are erased.
There is two other solution to solve this issue. The first is to set the two fields on the user tab, the second, delete the two lines in the actions "add" and "edit".
I don't know wich is the best solution. 
My solution is only a temporary patch in wait of a  decision.

### DIFF
--- a/htdocs/user/card.php
+++ b/htdocs/user/card.php
@@ -420,13 +420,13 @@ if (empty($reshook)) {
 				/*
 				* Protection against deletion of ref_employee while the field is not present in the user tab
 				*/
-				if (GETPOSTISSET("ref_employee")){
+				if (GETPOSTISSET("ref_employee")) {
 					$object->ref_employee = GETPOST("ref_employee", 'alphanohtml');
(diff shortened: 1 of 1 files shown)


eldy:
> # FIX
> Protection against deletion of ref_employee and national_registration_number while fields are not presents in the user tab
> When you set ref_employee and/or national_registration_number in the RH & bank tab if you edit the user tab and save, the 2 fields are erased.
> There is two other solution to solve this issue. The first is to set the two fields on the user tab, the second, delete the two lines in the actions "add" and "edit".
> I don't know wich is the best solution. 
> My solution is only a temporary patch in wait of a  decision.

The good one is the one you suggest: using GETPOSTISSET to overwrite or not.

PR #23670, opened 2023-01-24
Encase database name in backticks

### LINKED ISSUE #23643: Can't repair database if database name has - (hyphen)
### Bug

https://github.com/Dolibarr/dolibarr/blob/7a13ca65b503a8161b73f67d1d47e7a42bbcacd5/htdocs/core/db/mysqli.class.php#L708

image

Change

```sql
$sql = "SHOW TABLES FROM ".$tmpdatabase." ".$like.";";
```

to

```sql
$sql = "SHOW TABLES FROM `".$tmpdatabase."` ".$like.";";
```

Create a database with hyphen or simply try the SQL `SHOW TABLES FROM test-test;` in MySQL.

Daviid-P:
# FIX #23643 Encase database name in backticks
If the database name contains a hyphen `-` then the SQL query `SHOW TABLES FROM database-name` fails.

### DIFF
--- a/htdocs/core/db/mysqli.class.php
+++ b/htdocs/core/db/mysqli.class.php
@@ -678,7 +678,7 @@ class DoliDBMysqli extends DoliDB
 		}
 		$tmpdatabase = preg_replace('/[^a-z0-9\.\-\_]/i', '', $database);
 
-		$sql = "SHOW TABLES FROM ".$tmpdatabase." ".$like.";";
+		$sql = "SHOW TABLES FROM `".$tmpdatabase."` ".$like.";";
 		//print $sql;
 		$result = $this->query($sql);
 		if ($result) {


htdocs/core/db/mysqli.class.php
```diff
@@ -678,7 +678,7 @@ public function DDLListTables($database, $table = '')
 		}
 		$tmpdatabase = preg_replace('/[^a-z0-9\.\-\_]/i', '', $database);
 
-		$sql = "SHOW TABLES FROM ".$tmpdatabase." ".$like.";";
+		$sql = "SHOW TABLES FROM `".$tmpdatabase."` ".$like.";";
```
eldy, line 681: On line 679, we clean database name to remove - char 
So i am surprised your fix work. If thee is a - into name (we try to avoid this in install process), the show tables will use table name without the -

Daviid-P: The regex is removing anything that's not a to z, 0 to 9, period, hyphen or underscore (case insensitive).
Also, I don't see where in the install process the - or other bad characters are stripped or checked.

The GETPOST filter aZ09 also checks with the same regex.

```php
$db_name = GETPOST('db_name', 'aZ09') ?GETPOST('db_name', 'aZ09') : (empty($argv[10]) ? '' : $argv[10]);
```

```php
case 'aZ09':
  if (!is_array($out)) {
    $out = trim($out);
      if (preg_match('/[^a-z0-9_\-\.]+/i', $out)) {
          $out = '';
      }
  }
  break;
```

Also, in `DDLCreateDb()` we do use backticks:

```php
public function DDLCreateDb($database, $charset = '', $collation = '', $owner = '')
{
	// phpcs:enable
	if (empty($charset)) {
		$charset = $this->forcecharset;
	}
	if (empty($collation)) {
		$collation = $this->forcecollate;
	}

	// ALTER DATABASE dolibarr_db DEFAULT CHARACTER SET latin DEFAULT COLLATE latin1_swedish_ci
	$sql = "CREATE DATABASE `".$this->escape($database)."`";
	$sql .= " DEFAULT CHARACTER SET `".$this->escape($charset)."` DEFAULT COLLATE `".$this->escape($collation)."`";

	dol_syslog($sql, LOG_DEBUG);
	$ret = $this->query($sql);
	if (!$ret) {
		// We try again for compatibility with Mysql < 4.1.1
		$sql = "CREATE DATABASE `".$this->escape($database)."`";
		dol_syslog($sql, LOG_DEBUG);
		$ret = $this->query($sql);
	}
	return $ret;
}
```

Daviid-P: image

PR #23677, opened 2023-01-24
FIX #23676 desired and minimum stock displayed on replenish

### LINKED ISSUE #23676: Replenish - displayed info is good but does not seem logic because of treatment
### Bug

image

If a warehouse is specified to prepare the replenish list, the desired stocks and stocks minimum displayed may indicate 0 when it is no specified for this warehouse in particular. 
Although, the values specified on the product card are used instead (good behaviour)

I think, the value must be displayed in another color with a tooltip which explain the fact that this is the value of the product card because no one was specified for the warehouse

### Environment Version

17 and older

### Environment OS

Linux

### Environment Web server

Linux Debian

### Environment PHP

php 7.4

### Environment Database

_No response_

### Environment URL(s)

htdocs/product/stock/replenish.php

# Expected 
When no value is specified to stock minimum or desired stock for a warehouse in particular, indicates the value used in another color instead of 0. (The user may misunderstand the reason of this line appears)

# Actual
The values are 0, but the line appears if the default value matches the condition.

Fresh Dolibarr
Activate
- Product
- Supplier
- Stock

Create a product with desired / minimum stock to 10 for example
Create a warehouse
Go to Stock Configuration and allows Desired and Minimum Stock per warehouse-product pair

Go to Product -> Warehouse -> Replenish and choose a warehouse before filter

The line appears even if you have 0 in stock and 0 in desired stock column

P.S. : I will manage this bug soon if nobody is faster ;)

### Attached files

_No response_

ibuiv:
# FIX|Fix #23676
In the replenish section (Product, Stock, Supplier with the possibility to specify pair product-warehouse desired and minimum stocks.
0 was displayed when the value has not been set for the warehouse in the filter. 
Users may misunderstand why this product was included anyway in this list.
If it's not set, the default value will appear with an information icon which explain why this value has been displayed.


eldy:
When option STOCK_ALLOW_ADD_LIMIT_STOCK_BY_WAREHOUSE is set, if we did no set any value for the warehouse, we must suppose that value is zero (we don't want to replenish this warehouse).
If you choose the desired value for product, if there is 100 warehouses, you will replenish 100 x the desired quantity.
So it seems better to me to apply the rule:  No explicit instruction by user -> no action.

So showing a warning is good, but we should also show value "zero" to explain user we won't replenish it. So i modified the PR in this way...

PR #23680, opened 2023-01-24
Fix bad results on replenish when filtering for a warehouse

ibuiv:
Bad name for getEntity multicompany sharing parameter The stock compared to prepare the list is now considering only products in the chosen warehouse

# FIX|Fix Wrong Behaviour on replenish while filtering on a warehouse
The list of Warehouse IDs in case of multicompany uses was false. Wrong parameter.
To prepare the list, the stock we will verify are NOW only the one into the chose warehouse.

### DIFF
--- a/htdocs/product/stock/replenish.php
+++ b/htdocs/product/stock/replenish.php
@@ -6,7 +6,7 @@
  * Copyright (C) 2016		ATM Consulting		<support@atm-consulting.fr>
  * Copyright (C) 2019       Frédéric France         <frederic.france@netlogic.fr>
  * Copyright (C) 2021		Ferran Marcet		<fmarcet@2byte.es>
- * Copyright (C) 2021		Antonin MARCHAL		<antonin@letempledujeu.fr>
+ * Copyright (C) 2021-2023		Antonin MARCHAL		<antonin@letempledujeu.fr>
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


Humml87:
Hi ibuiv, your PR has a conflict with my PR #23316. It's is easier to merge first the PR#23316 then integrate your PR.
@eldy please check this.

goldenwt:
Hi @eldy ,

Well done with your team on Dolibarr, you guys are doing a great job and honestly, i am proud of the work everyone of you is doing to make Dolibarr the preferred solution to businesses.

I pls have an issue with the replenishment feature, though it seemed as if from the feedback it has been addressed, but i have not seen the effect.

The products i expect to show on the replenishment report doesnt show despite setting the stock alert and desired stock. In fact once the list gets to 38, it doesnt exceed despite the fact that we have products due for restock. We have tried filtering according to warehouse and open filtering - no selection of warehouses to see if there is a difference, but none

Is there something i am not doing or settings not right from the code point of view??

(!empty($conf->global->STOCK_ALLOW_ADD_LIMIT_STOCK_BY_WAREHOUSE) && $fk_entrepot > 0) is a boolen.
But you compare it with > 0, where can i make this correction and with was am i to replace the code to get the right result or output

htdocs/product/stock/replenish.php
```diff
@@ -480,45 +480,85 @@
 	}
 
 	$sql .= ' HAVING (';
-	$sql .= " (".$sqldesiredtock." >= 0 AND (".$sqldesiredtock." > SUM(".$db->ifsql("s.reel IS NULL", "0", "s.reel").')';
+	if (!empty($conf->global->STOCK_ALLOW_ADD_LIMIT_STOCK_BY_WAREHOUSE) && $fk_entrepot > 0) {
+		$sql .= " (" . $sqldesiredtock . " >= 0 AND (" . $sqldesiredtock . " > SUM(" . $db->ifsql("s.reel IS NULL OR s.fk_entrepot <> " . $fk_entrepot, "0", "s.reel") . ')';
```
eldy, line 484: Muse use 
`$sql .= " (" . $sqldesiredtock . " >= 0 AND (" . $sqldesiredtock . " > SUM(" . $db->ifsql("s.reel IS NULL OR s.fk_entrepot <> " . (int) $fk_entrepot, "0", "s.reel") . ')';`

htdocs/product/stock/replenish.php
```diff
@@ -480,45 +480,85 @@
 	}
 
 	$sql .= ' HAVING (';
-	$sql .= " (".$sqldesiredtock." >= 0 AND (".$sqldesiredtock." > SUM(".$db->ifsql("s.reel IS NULL", "0", "s.reel").')';
+	if (!empty($conf->global->STOCK_ALLOW_ADD_LIMIT_STOCK_BY_WAREHOUSE) && $fk_entrepot > 0) {
```
eldy, line 483: Must replace 
if (!empty($conf->global->STOCK_ALLOW_ADD_LIMIT_STOCK_BY_WAREHOUSE)...
with
if (getDolGlobalString('STOCK_ALLOW_ADD_LIMIT_STOCK_BY_WAREHOUSE')...

eldy:

PR #23693, opened 2023-01-25
NEW - Do not select a default recipient

BB2A-Anthony:
# NEW|New [*Do not select a default recipient*]
[*To avoid recipient errors, it is essential to have the possibility of deactivating the preselection of the email recipient even if there is a single choice.*]


htdocs/core/class/html.formmail.class.php
```diff
@@ -1076,10 +1078,14 @@ public function getHtmlForTo()
 					$tmparray[$key] = dol_htmlentities($tmparray[$key], null, 'UTF-8', true);
 				}
 
-				$withtoselected = GETPOST("receiver", 'array'); // Array of selected value
+				if (getDolGlobalInt('MAIN_MAIL_NO_WITH_TO_SELECTED')) {
+					$withtoselected = '';
```
eldy, line 1082: This means we will always erase the recipients, even for the case, we have entered a list of recipient and submit the form and we have an error. We will loose the selection ?

PR #23700, opened 2023-01-26
Limitation of rights to modify HR informations on their own profile

comaiteseb:
# FIX 
Limitation of rights to modify HR informations (salaries, daily rate, hourly rate, position held, date of entry, etc.) on their own profile so that a simple employee can access it without however being able to modify it
A user with the right user->self->creer but not having the right user->user->creer will not be able to modify his position, his salary, his tjm and thm as well as his date of entry.

### DIFF
--- a/htdocs/core/modules/modTakePos.class.php
+++ b/htdocs/core/modules/modTakePos.class.php
@@ -267,7 +267,7 @@ class modTakePos extends DolibarrModules
 	 */
 	public function init($options = '')
 	{
-		global $conf, $db, $langs, $user;
+		global $conf, $db, $langs, $user, $mysoc;
 		$langs->load("cashdesk");
 
 		dolibarr_set_const($db, "TAKEPOS_PRINT_METHOD", "browser", 'chaine', 0, '', $conf->entity);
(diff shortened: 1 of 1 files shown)


htdocs/user/card.php
```diff
@@ -2791,7 +2792,11 @@
 				// Salary
 				print '<tr><td>'.$langs->trans("Salary").'</td>';
 				print '<td>';
-				print img_picto('', 'salary', 'class="pictofixedwidth paddingright"').'<input size="8" type="text" name="salary" value="'.price2num(GETPOST('salary') ?GETPOST('salary') : $object->salary).'">';
+				if ($caneditsensitivedata) {
```
eldy, line 2795: The salary is not a data on the control of the user. So i think it should be managed by the permission on the salary or rh module (generally owned by HR people). Not by a permission on users (generally owned by admin).

comaiteseb: what do you think about using user->right->salaries->write for the salary field and user->rights->hrm->write_personal_information->write for the others (job, weekly hours, tjm, thm,dateemployment and dateemploymentend) ?

comaiteseb: @eldy is this Ok for you ?

PR #23703, opened 2023-01-26
NEW : payment on several expense report (code integration)

atm-gauthier:


atm-gauthier:
Like SQL integration PR (https://github.com/Dolibarr/dolibarr/pull/23702) i get an error in testExpenseReportOther() function because dev is split in 2 PR

PR #23707, opened 2023-01-26
Allow a user to modify his IBAN with the only right user->self->creer

comaiteseb:
# FIX
Allow a user to modify his IBAN with the only right user->self->creer
To avoid hazardous results on existing installations, I added a parameter named USER_CANEDIT_IBAN in the setup of the user module


aspangaro:
A last question : Why use an option when you can use an advanced right?

aspangaro:
And for security reasons, it is not advised to give access to this information directly to the user because a hacking of a user account would allow to insert another potentially erroneous RIB to which in the future we can base ourselves to make SEPA transfer files in Dolibarr. 

It is better that a third person has access to modify the RIB but not the user directly or with registration of an event/notification and transmission of an email in case of modification.

comaiteseb:
I hear the security issue, but it is the same for any user including administrators... 
It is therefore necessary to send a warning by email and record the change in the events (which should have been done originally)
I will add notifications and event logging in a next comit

In my opinion, a user without the right "Create/modify own user information" should not be able to modify his IBAN. However, with an advanced right, he could have the right to modify his IBAN without necessarily having the right to modify his own information.
Even if in the code it is feasible, I fear that in terms of understanding this in the management of rights it is less intuitive.

htdocs/admin/user.php
```diff
@@ -184,6 +184,25 @@
 }
 print '</td></tr>';
 
+// user can edit is IBAN in rh tab if he is authorized to edit himself his informations
+
+print '<tr class="oddeven">';
```
eldy, line 189: Permissions to do or not must not be managed by an option. This make things unclear for the permission administrator.

Because IBAN is an information provided by the user. If the user has the permission  user->self->creer , he can be allowed ot modify IBAN.
"user->self->creer" means you can modify all information related to you so using this can be enough.

Sure, having a warning sending an email to warn the change is a good thing, but this can be another PR...

htdocs/user/bank.php
```diff
@@ -862,8 +862,8 @@
 }
 
 // Edit
-if ($id && ($action == 'edit' || $action == 'create') && $user->rights->user->user->creer) {
-	$title = $langs->trans("User");
+if ($id && ($action == 'edit' || $action == 'create') && ($permissiontoaddbankaccount || $user->rights->user->user->creer)) {
+	$title = $langs->trans("User");echo "ici";
```
eldy, line 866: Debug log to remove

comaiteseb:
remove the option USER_CANEDIT_IBAN
create a new function in user class to send mail to user personal_email 
add a trigger USER_RIBMODIFY to log as security events, in agenda and send a mail with the previous user new function.

htdocs/langs/fr_FR/users.lang
```diff
@@ -129,3 +129,6 @@ IPLastLogin=IP dernière connexion
 IPPreviousLogin=Connexion précédente IP
 ShowAllPerms=Afficher toutes les lignes d'autorisation
 HideAllPerms=Masquer toutes les lignes d'autorisation
+EventUserRibModifiedHTML=Vos coordonnées bancaires viennent d'être modifiées. Si vous n'êtes pas à l'origine de cette modification, veuillez rapidement en informer votre administrateur.
```
frederic34, line 132: add translations for en_US, they are used to create all other translations

htdocs/core/class/events.class.php
```diff
@@ -103,6 +103,7 @@ class Events // extends CommonObject
 		array('id'=>'USERGROUP_CREATE', 'test'=>1),
 		array('id'=>'USERGROUP_MODIFY', 'test'=>1),
 		array('id'=>'USERGROUP_DELETE', 'test'=>1),
+		array('id'=>'USER_RIBMODIFY', 'test'=>1),
```
frederic34, line 106: RIB is a french term, prefer to use english term like IBAN

htdocs/user/class/user.class.php
```diff
@@ -2487,6 +2487,56 @@ public function send_password($user, $password = '', $changelater = 0)
 		}
 	}
 
+	// phpcs:disable PEAR.NamingConventions.ValidFunctionName.ScopeNotCamelCaps
```
frederic34, line 2490: do not use phpcs:disable to introduce new not CamelCaps code

htdocs/core/class/events.class.php
comaiteseb: Done, but in a great part of code the term RIB is used... If you think is a good  idea I can explore the whole source code to transform RIB in IBAN

htdocs/user/class/user.class.php
comaiteseb: Done

htdocs/langs/fr_FR/users.lang
comaiteseb: Done

htdocs/user/class/user.class.php
```diff
@@ -2487,6 +2487,55 @@ public function send_password($user, $password = '', $changelater = 0)
...
+	 *  Send an email to the personnal user adress
+	 *
+	 *  @param	User	$user           Object user that send the email (not the user we send to)
+	 *  @param	string	$subject        Mail subject
+	 *	@param	string  $html 			Mail body in html
+	 *  @return int 		            < 0 si erreur, > 0 si ok
+	 */
+	public function sendPersonalEmail($user, $subject, $html)
+	{
+		// phpcs:enable
```
frederic34, line 2500: remove also phpcs:enable

comaiteseb: done

PR #23723, opened 2023-01-29
Update commonobject.class.php

dcoupin:
# NEW|New [*possibility to add order*]
[*possibility to add order in showInputField*]


htdocs/core/class/commonobject.class.php
```diff
@@ -6776,8 +6776,12 @@ public function showInputField($val, $key, $value, $moreparam = '', $keysuffix =
 				}
 				$sql .= $sqlwhere;
 				//print $sql;
-
-				$sql .= ' ORDER BY '.implode(', ', $fields_label);
+				//permet de modifier les order
+				if (($InfoFieldList[5])) {
+					$sql.= ' ORDER BY '.$InfoFieldList[5];
```
eldy, line 6781: Can you also add a comment on line 6727 to document this new parameter ?

PR #23728, opened 2023-01-30
This allows the hook getFormMail to automatically add email recipients

marcusdeangabriel:
# NEW|
This is a request for enhancement (RFE).  This change allows the hook getFormMail to automatically add email recipients as desired when a user clicks on the button SEND MAIL (language key `SendMail`).  The function `getFormMail ` is the only one that calls the function `getHtmlForTo`, so one uses the hook `getFormMail ` to set `$this->withtoselected` via `$object->withtoselected` in the hook `getHtmlForTo` to automatically add email recipients as desired.

### DIFF
--- a/htdocs/core/class/html.formmail.class.php
+++ b/htdocs/core/class/html.formmail.class.php
@@ -1082,11 +1082,11 @@ class FormMail extends Form
 					$withtoselected = array_keys($tmparray);
 				}
 
-                // The function getFormMail is the only one that calls this one, so one uses the hook
-                // getFormMail to set $this->withtoselected via $object->withtoselected in the hook
-                if (!empty($this->withtoselected)) {
-                    $withtoselected = array_keys($this->withtoselected);
-                }
+				// The function getFormMail is the only one that calls this one, so one uses the hook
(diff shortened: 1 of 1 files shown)


htdocs/core/class/html.formmail.class.php
```diff
@@ -1082,6 +1082,12 @@ public function getHtmlForTo()
 					$withtoselected = array_keys($tmparray);
 				}
 
+				// The function getFormMail is the only one that calls this one, so one uses the hook
+				// getFormMail to set $this->withtoselected via $object->withtoselected in the hook
+				if (!empty($this->withtoselected)) {
```
eldy, line 1087: Why don't you set the ->withto  ?  it seems it is the one dedicated to set  $withtoselected to your own value

marcusdeangabriel:
The `->withto` field fills the recipient field and hides the following and/or drop-down list which we did not want.  Thus my choice above.  Your question forced me to examine the dolibarr code once again and the` ->withtofree` field does what we want.  So I do not need to modify `core/class/html.formmail.class.php` in order to make our hook work.

A pull is not necessary and I will close this pull request.

Thanks for the assist.

PR #23736, opened 2023-01-31
Fix : stripe sepa payment process

Hystepik:
Debug stripe sepa payment process
Move the update of invoice to payed when getting a post payment_intent.succeeded from stripe api


htdocs/core/class/commoninvoice.class.php
```diff
@@ -1186,126 +1194,11 @@ public function makeStripeSepaRequest($fuser, $did = 0, $type = 'direct-debit',
...
-											$paiement->date = $now;
-											if ($currencyCodeType == $conf->currency) {
-												$paiement->amounts = [$this->id => $amounttopay];   // Array with all payments dispatching with invoice id
-											} else {
-												$paiement->multicurrency_amounts = [$this->id => $amounttopay];   // Array with all payments dispatching
-
-												$postactionmessages[] = 'Payment was done in a different currency than currency expected of company';
-												$ispostactionok = -1;
-												// Not yet supported, so error
+											$result =$this->setStatut($this::STATUS_PROCESSING);
```
eldy, line 1197: The fact that there is something "processing" must not be store into the invoice but into the llx_prelevement_bon (that must be created where there is the // TODO Create a prelevement_bon)
Because 1 invoice can have 2 request: 1 done, 1 pending, and a remain to pay not started. So no need to add a status on invoice. We already know there is some request in the pipe with the table llx_prelevement_request (where is sotre the invoice and the amount) linked to llx_prelevement_bon (where is stored the status).

htdocs/core/class/commoninvoice.class.php
```diff
@@ -1563,13 +1456,13 @@ public function makeStripeSepaRequest($fuser, $did = 0, $type = 'direct-debit',
 					$idtransferfile = 0;
 
 					// Update the direct debit payment request of the processed invoice to save the id of the prelevement_bon
-					$sql = "UPDATE".MAIN_DB_PREFIX."prelevement_demande SET";
+					$sql = "UPDATE ".MAIN_DB_PREFIX."prelevement_demande SET";
```
eldy, line 1459: You must introduce the creation of processing request 4 lines before this one at the // TODO Create a prelevement_bon...

eldy:
Note: I was in need for the modification of the sepaStripe function to use the new API instead of old one. 
So i merged the suggested change manually by cut and paste, then i made minor enhancement. You can remove this part from your PR now it is inside the core.

htdocs/compta/prelevement/class/bonprelevement.class.php
```diff
@@ -106,7 +106,7 @@ class BonPrelevement extends CommonObject
 	 */
 	public function __construct($db)
 	{
-		global $conf, $langs;
+		global $conf, $langs, $db;
```
eldy, line 109: Why do you add $db as global, you receive it as parameter ?

PR #23741, opened 2023-01-31
Fix MRP (GPAO) create truncatured manufacturing order

### LINKED ISSUE #23739: MRP (GPAO) create truncatured manufacturing order when item inside BOM is qty 0
### Bug

When i create a manufacturing order using a BOM with one item qty=0, the manufacturing order contain only items before the one with qty=0.

### Environment Version

Dolibarr 17.0 (branch 17.0 commit c1c21846ab0730826a3123d743a385f77365d921)

### Environment OS

Debian GNU/Linux 11 (bullseye)

### Environment Web server

nginx/1.18.0

### Environment PHP

7.4.33

### Environment Database

MariaDB 5.5.5-10.5.18-MariaDB-0+deb11u1

### Environment URL(s)

https://test.pwln.eu/dolibarr/mrp/mo_card.php?id=1

### Expected behavior :
All items listed in the BOM with at leaste qty>0 must be present in the manufacturing order.
Capture d’écran du 2023-01-31 17-13-58
Capture d’écran du 2023-01-31 17-14-43

### Actual behavior:
Only the items before the one with qty=0 are present in the manufacturing order.
Capture d’écran du 2023-01-31 14-30-46
Capture d’écran du 2023-01-31 17-12-32

### create a BOM with 9 items like this one :
Capture d’écran du 2023-01-31 14-30-53

### create a Manufacturing Order with this BOM
Capture d’écran du 2023-01-31 18-46-44

### Attached files

_No response_

jn-powerlan:
# Fix #23739 
At the creation of a Manufacturing Order if BOM containing item with qty=0 then skip them instead of breaking the loop.


htdocs/mrp/class/mo.class.php
```diff
@@ -706,7 +706,7 @@ public function updateProduction(User $user, $notrigger = true)
 							if ($moline->qty <= 0) {
 								$error++;
 								$this->error = "BadValueForquantityToConsume";
-								break;
+								continue;
```
eldy, line 709: We have the $error++, so we should return an error code with the goal to cancel everything.
We want nothing happen if error. The the continue should be useless.
I thin the error is later when we make a commit instead of a rollback when the return code is < 0

jn-powerlan (commented):
There is two problems to fix : 

1. When a bomline qty is 0, it should not throw an error.
2. When the calculated value of qty to consume is less than zero it should throw an error and stop the creation of the manufacturing order.

I think my solution for the first problem is ok.
But for the second i'm not sure, it's a proposal, any help is welcome.

eldy:
> There is two problems to fix :
> 
> 1. When a bomline qty is 0, it should not throw an error.

Having a bomline qty to 0 looks strange. If we have this there is probably an error to fix before. If we have 0, it means we don't need the line at all. doesn't it ? Do you have an exemple of use cas where bom qty line can be 0 ? Why not removing the line completely if we have zero ?

> 2. When the calculated value of qty to consume is less than zero it should throw an error and stop the creation of the manufacturing order.

As soon as there is 1 error, even if the rest is good, we must cancel everything and do nothing so until the error is fixed, nothing happen.

> 
> I think my solution for the first problem is ok. But for the second i'm not sure, it's a proposal, any help is welcome.

jn-powerlan:
> > There is two problems to fix :
> > 
> > 1. When a bomline qty is 0, it should not throw an error.
> 
> Having a bomline qty to 0 looks strange. If we have this there is probably an error to fix before. If we have 0, it means we don't need the line at all. doesn't it ? Do you have an exemple of use cas where bom qty line can be 0 ? Why not removing the line completely if we have zero ?
> 
> > 2. When the calculated value of qty to consume is less than zero it should throw an error and stop the creation of the manufacturing order.
> 
> As soon as there is 1 error, even if the rest is good, we must cancel everything and do nothing so until the error is fixed, nothing happen.
> 
> > I think my solution for the first problem is ok. But for the second i'm not sure, it's a proposal, any help is welcome.

In that case, having a BOM line with a quantity of 0 would mean that a certain component is not needed for that particular variant of the product. This is a common use case in electronic BOMs where different product variants share the same PCB but require different components.

In this situation, it would be appropriate to modify the code to handle BOM lines with a quantity of 0 without throwing an error. Instead, the line should be skipped and not included in the manufacturing order.

For the second problem, your proposal of throwing an error and stopping the creation of the manufacturing order when the calculated value of the qty to consume is less than zero is correct. It ensures that the manufacturing order is not created with invalid or impossible quantities.

htdocs/mrp/class/mo.class.php
```diff
@@ -706,6 +709,7 @@ public function updateProduction(User $user, $notrigger = true)
 							if ($moline->qty <= 0) {
 								$error++;
 								$this->error = "BadValueForquantityToConsume";
+								dol_print_error(0, $langs->trans("BadValueForquantityToConsume", $moline->qty));
```
eldy, line 712: The mo class can be executed in any context. dol_print_error is allowed into web page context only
On error, class must return a negative value after setting ->error property.

PR #23745, opened 2023-02-01
FIX: Homogenise handling of null extrafield value in create/update

Kazimir42:
In *insertExtraFields* method the following check is made to check whether an extra field value is NULL:

https://github.com/Dolibarr/dolibarr/blob/5395a4ab4b6d9db3c1ecb4fdbf871ea3ff9cad31/htdocs/core/class/commonobject.class.php#L6028

In *updateExtraField* method the following check is made to check whether an extra field value is NULL:

https://github.com/Dolibarr/dolibarr/blob/5395a4ab4b6d9db3c1ecb4fdbf871ea3ff9cad31/htdocs/core/class/commonobject.class.php#L6353

---

Because of that when a user is trying to select empty value from extra field of type ‘link’ in the propal page Dolibarr tried to insert an empty value which cause an SQL error.

image

The behaviour works in other page because *insertExtraFields* is used instead of *updateExtraField* .

These checks were not coherent so I homogenise them in order to ensure same behaviour.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -6350,7 +6350,7 @@ abstract class CommonObject
 			}
 
 			if ($linealreadyfound) {
-				if ($this->array_options["options_".$key] === null) {
+				if ($this->array_options["options_".$key] === null || $this->array_options["options_".$key] === "") {
 					$sql = "UPDATE ".MAIN_DB_PREFIX.$this->table_element."_extrafields SET ".$key." = null";
 				} else {
 					$sql = "UPDATE ".MAIN_DB_PREFIX.$this->table_element."_extrafields SET ".$key." = '".$this->db->escape($this->array_options["options_".$key])."'";
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -6304,6 +6304,11 @@ abstract class CommonObject
 						$this->array_options["options_".$key] = null;
 					}
 					break;
+				case 'link':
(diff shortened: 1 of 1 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -6350,7 +6350,7 @@ public function updateExtraField($key, $trigger = null, $userused = null)
 			}
 
 			if ($linealreadyfound) {
-				if ($this->array_options["options_".$key] === null) {
+				if ($this->array_options["options_".$key] === null || $this->array_options["options_".$key] === "") {
```
eldy, line 6353: Imagine we have an extrafield of type string and we want to set an empty string inside it. We can't anymore.
I think we must separate to case: The case the extrafield is type "link" and the other.
May be we can add this few lines before in the switch that sanitize date:

```
case 'link':
   if ($this->array_options["options_".$key] === "") {
    $this->array_options["options_".$key] = null
  }
```

Can you check/test this way ?

Kazimir42: @eldy This would work but the behavior is not coherent between the 2 methods, in *insertExtraFields* an empty link is represented by an empty string and with this changes in *updateExtraField* an empty link will be represented by a null value. What do you think?

eldy: An empty link should be a null value in database (when there is no link) to avoid foreign key errors during insert or update.

Kazimir42: Okay thanks,

@eldy It's done

creekorful:
@eldy Any news on this one?

PR #23754, opened 2023-02-01
Fix progress less than previous percent on situation invoice

SylvainLegrand:


eldy:

SylvainLegrand:
On situation n°1, the progression is set at 80%.
Then you create the following situation invoice (n°2).
You should not set the progress to less than 80%!

htdocs/compta/facture/card.php
```diff
@@ -2473,12 +2473,12 @@
 				setEventMessages($mesg, null, 'warnings');
 				$error++;
 				$result = -1;
-			} elseif ($progress < $percent) {
-				$mesg = '<div class="warning">'.$langs->trans("CantBeLessThanMinPercent").'</div>';
-				setEventMessages($mesg, null, 'warnings');
-				$error++;
-				$result = -1;
 			}
+		} elseif ($progress < $percent) {
```
eldy, line 2477: This is strange because you move a test about situation invoice out of a if that is related to situation invoice.
This means this test will be executed even for non situation invoice ?

SylvainLegrand: I do not understand your question !
The progress column is only displayed on the situation invoice!
Also, for now, the test is done only for situation credit note.

eldy: When an invoice is not a situation invoice, i suppose that invoice with $object->situation_cycle_ref = 0 or invoice with $object->situation_cycle_ref is not defined.

Here, when  $object->situation_cycle_ref = 0, you make a test ($progress < $percent)   but $progress has a meaning only for situation invoice.

PR #23758, opened 2023-02-02
NEW Can modify margin rates in offers like VAT rates.

marcusdeangabriel:
# NEW
This is a request for enhancement to permit users to modify all of the margins of the products in an offer at once just as one can do for the VAT of the products in an offer.

A toggle to enable or disable this feature, `PROPALE_MODIFY_MARGIN_RATES`, was added to the settings of the offers module.  If enabled, a margin rates column is visible and a modification pencil like that for the VAT is visible and usable if the user has the right to create offers.  Note that the margins rates column is visible when `PROPALE_MODIFY_MARGIN_RATES` is true even if `DISPLAY_MARGIN_RATES` of the margins module is false.  If `DISPLAY_MARGIN_RATES` is true and `PROPALE_MODIFY_MARGIN_RATES`  is false, the margin rates column is visible but margins cannot be modified.

For example,
PROPALE_MODIFY_MARGIN_RATES


htdocs/admin/propal.php
```diff
@@ -615,6 +615,19 @@
 print "</tr>";
 
 
+print '<tr class="oddeven">';
+print '<td>'.$langs->trans("ModifyMarginRates").'</td>';
+print '<td class="center">';
+if ($conf->use_javascript_ajax) {
+	print ajax_constantonoff('PROPALE_MODIFY_MARGIN_RATES');
```
eldy, line 622: I don't think we need an option PROPALE_MODIFY_MARGIN_RATES for this as it is an "addition" feature (so low risk to break existing).
Can you remove this option and remove condition PROPALE_MODIFY_MARGIN_RATES (we imagine this condition is always true).

marcusdeangabriel: As suggested, removed PROPALE_MODIFY_MARGIN_RATES as not needed.

marcusdeangabriel:
I have review my work, and for me, this is ready for review.

PR #23761, opened 2023-02-02
Add options in function pdf_pagefoot & pdf_pagehead of pdf.lib.php

habot-it:
Hello @eldy,

In pdf_pagehead function :
- Add MAIN_USE_BACKGROUND_ON_PDF_ALPHA option for change opacity of the background
- SetPageMark moved to fixe the same problem of MAIN_USE_BACKGROUND_ON_PDF but for MAIN_ADD_PDF_BACKGROUND too

In pdf_pagefoot function :
- Add PDF_FOOTER_HIDDEN to hide all footer (page number will no hidden)
- Add PDF_FOOTER_PAGE_NUMBER_X & PDF_FOOTER_PAGE_NUMBER_Y to change XY position of the page number

Best regards,
Alban


htdocs/core/lib/pdf.lib.php
```diff
@@ -724,11 +724,13 @@ function pdf_pagehead(&$pdf, $outputlangs, $page_height)
 		$filepath = $conf->mycompany->dir_output.'/logos/'.$conf->global->MAIN_USE_BACKGROUND_ON_PDF;
 		if (file_exists($filepath)) {
 			$pdf->SetAutoPageBreak(0, 0); // Disable auto pagebreak before adding image
+			if (getDolGlobalString('MAIN_USE_BACKGROUND_ON_PDF_ALPHA')) { $pdf->SetAlpha($conf->global->MAIN_USE_BACKGROUND_ON_PDF_ALPHA); } // Option for change opacity of background
 			$pdf->Image($filepath, (isset($conf->global->MAIN_USE_BACKGROUND_ON_PDF_X) ? $conf->global->MAIN_USE_BACKGROUND_ON_PDF_X : 0), (isset($conf->global->MAIN_USE_BACKGROUND_ON_PDF_Y) ? $conf->global->MAIN_USE_BACKGROUND_ON_PDF_Y : 0), 0, $page_height);
-			$pdf->SetPageMark(); // This option avoid to have the images missing on some pages
+			if (getDolGlobalString('MAIN_USE_BACKGROUND_ON_PDF_ALPHA')) { $pdf->SetAlpha(1); }
```
eldy, line 729: There was an important non on the line $pdf->SetPageMark(); 
Do you really need to remove this line ? Doing this make the code when MAIN_USE_BACKGROUND_ON_PDF_ALPHA is off to be different than old one.

htdocs/core/lib/pdf.lib.php
```diff
@@ -724,11 +724,13 @@ function pdf_pagehead(&$pdf, $outputlangs, $page_height)
...
 			$pdf->SetAutoPageBreak(1, 0); // Restore pagebreak
 		}
 	}
+	$pdf->SetPageMark(); // This option avoid to have the images missing on some pages
```
eldy, line 733: Why moving this outside of the if (!empty($conf->global->MAIN_USE_BACKGROUND_ON_PDF)
If this test is wrong nothing is modified, so do we really need to do a SetPageMark ?

habot-it: Hello @eldy,
I deleted my change.

Yes, we need $pdf->SetPageMark(); otherwise the image is not really considered as a background image : the layers are mixed between the background image, the text, the frames, etc

Best regards,
Alban

habot-it: Revision 2 : I confirm that when we use a PDF as a background image, we have the same problem as with MAIN_USE_BACKGROUND_ON_PDF

habot-it:
Merci @eldy ;)

PR #22960, opened 2022-11-25
NEW move htmlPrintOnlineFooter in their own lib

lvessiller-opendsi:
NEW public and payment interface with footer at bottom

**Before**
Public interface

- Index
image

- In list (a page with scroll-bar)
image

Payment interface

image

**After**
Public interface

- Index
image

- In list (a page with scroll-bar)
image

Payment interface

image

The method "htmlPrintOnlinePaymentFooter" is reviewed and use a new method "htmlOnlineCompanyFooter" to get lines of company for footer.
This method "htmlOnlineCompanyFooter" is now introduced for all public footer to get lines of company for footer.
As the header, the footer is also separated.

PR #23777, opened 2023-02-03
Ticket public pages : reformat footer

thomas-Ngr:
# NEW Cleaner footer for ticket public pages

This is a cleaner version of PR https://github.com/Dolibarr/dolibarr/pull/22960 by @lvessiller-opendsi , focused on ticket interface.

This PR:
- Reformats htmlPrintOnlinePaymentFooter() in payment.lib.php and creates a function htmlPrintOnlineCompanyFooter() in company.lib.php. Now, one can display the company informations in the footer without displaying payment information.
- Cleans the structure of public ticket pages, using meaningful tags header, main, footer. Using only div tags did lead to a bug where one div was not properly closed, leading to wrong display.

### Before 
- in public interface
image

- in list, with few elements (the footer is not at the bottom of the page)
image

### After 
- in public interface
image

- in lists, with few elements
image

### DIFF
--- a/htdocs/core/lib/company.lib.php
+++ b/htdocs/core/lib/company.lib.php
@@ -2264,3 +2264,76 @@ function addMailingEventTypeSQL($actioncode, $objcon, $filterobj)
 		return $sql2;
 	}
 }
+
+/**
+ * Show footer of company in HTML pages
+ *
+ * @param   Societe		$fromcompany	Third party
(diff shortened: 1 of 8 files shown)


eldy:
Due to other PR in conflict, this PR was only partially merged, manually.
But one feature is original PR is still missing. The footer is not forced on bottom of page. 
I think it is better to open another PR for this.

PR #23778, opened 2023-02-03
NEW Link to input facture source on payment when there is a lot of invoice

leninrivas:
# NEW|New [*Link to input facture source*]
[*Link to input facture source, useful when there are many invoices*]


eldy:
Can you provide a screenshot showing the addition ?

leninrivas:
> Can you provide a screenshot showing the addition ?

Good day.
this screenshoot
screenshoot_when_manyinvoices

eldy:
OK, so my understanding is that this link is when there is a high number of lines ? May be to avoid user to ask what is goal of the link (as i did), we can show it only if there is more than 10 links for example ?

leninrivas:
> OK, so my understanding is that this link is when there is a high number of lines ? May be to avoid user to ask what is goal of the link (as i did), we can show it only if there is more than 10 links for example ?

Yes, then with a constant for example FACTURE_PAYMENTS_SHOW_LINK_TO_INPUT_ORIGIN_IS_MORE_THAN, and this for default 10.
let's proceed then.

leninrivas:
> OK, so my understanding is that this link is when there is a high number of lines ? May be to avoid user to ask what is goal of the link (as i did), we can show it only if there is more than 10 links for example ?

That is it

PR #23782, opened 2023-02-04
FIX Bug Estimated Value stockatdate.php

josett225:
# FIX # Bug Estimated Stock at date value
In stockatdate page, the estimatedvalue is wrong. In fact the calculated value in the SQL request is the current stock value and not the stock at date value.
P.S : I changed in code estimatedvalue to currentvalue and kept it for future use (other PR) as it is in interesting to get a comparison between stockatdate value to currentstock value.


eldy:
It seems bug was fixed inside v17 but in a different manner.
Can you check to backport the way it was fixed in v17 (using same name as possible for the lines that are modified) to avoid conflict ?

josett225:
Hi @eldy
Sorry that's right I found the fix in v17 (version I did not tested initially as still in beta if I am not wrong).
I can of course apply the same change in my fix but I have 2 concerns here :

- One is the fact estimatedvalue in the SQL request is not an estimated value. It is the current stock value.
- Second : I reused the $estimatedvalue variable later in the code to calculate the right estimated stock value once and to avoid to do 3 times the same calculation as seen in V17. The code is slow so removing a bit a process could help.

Don't you think It could be better to clean v14 and above to avoid the 2 concerns above?

I will do anyway what you will recommend.

eldy:
To avoid conflict and regression the maximum we can, and keep code analysis of past changes easier as possible, changes on old version must be lower as possible to fix the bug for the end user.
If you have improvement (renaming variable, or reuse), it is better to push this into develop branch.

htdocs/product/stock/stockatdate.php
```diff
@@ -245,12 +245,12 @@
 
 $sql = 'SELECT p.rowid, p.ref, p.label, p.description, p.price, p.pmp,';
 $sql .= ' p.price_ttc, p.price_base_type, p.fk_product_type, p.desiredstock, p.seuil_stock_alerte,';
-$sql .= ' p.tms as datem, p.duration, p.tobuy, p.stock, ';
+$sql .= ' p.tms as datem, p.duration, p.tobuy, p.stock, p.pmp, ';
```
eldy, line 248: Field p.pmp seems already added 2 lines before...

josett225: Hi @eldy 
I will review the code.

josett225: Hi @Eldy
Sorry for having been so late.

In version 14.0, I just reviewed the latest code and there is no p.pmp in line 246 (or 245 to be exact).
I will anyway make the same change as seen in v17 or even v19 if you want.

I will also submit  a new PR on develop to change the estimatedvalue to currentvalue in the SQL request to clear the code.
Let me know your decision.

josett225:
I will close this PR as it has been replaced by #26479 for v14 and #26481 for V19

PR #23802, opened 2023-02-06
 Fix #23799 External users are not able to create events

### LINKED ISSUE #23799: External users are not able to create events
### Bug

External users are not able to create events, even if they have the permission 

### Environment Version

V16.x

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

/comm/action/card.php

_No response_

_No response_

### Attached files

_No response_

glu000:
FIX|Fix #23799 External users are not able to create events
External users are not able to create events


htdocs/comm/action/card.php
```diff
@@ -142,9 +142,6 @@
 $TDurationTypes = array('y'=>$langs->trans('Years'), 'm'=>$langs->trans('Month'), 'w'=>$langs->trans('Weeks'), 'd'=>$langs->trans('Days'), 'h'=>$langs->trans('Hours'), 'i'=>$langs->trans('Minutes'));
 
 $result = restrictedArea($user, 'agenda', $object->id, 'actioncomm&societe', 'myactions|allactions', 'fk_soc', 'id');
-if ($user->socid && $socid) {
```
eldy, line 145: External user must be able to create events if event is linked to the thirdparty of external user.
If you remove this external user will be able to create event without the link to its thirdparty.

glu000: Hi Eldy,

I tested this PR and figured out, that it is not possible to create events for other thirdparties. When creating a new event (comm/action/card.php?action=create) it is not possible to select a another thirdparty then his own one. This is because the function select_thirdparty_list is used to print the combobox which is limiting the SQL query to socid of the user (see file html.form.class.php line#1373).

The problem of keeping the line
`$result = restrictedArea($user, 'societe', $socid);`
is, that it is not possible to add events to other elements like proposals, customer orders, etc. Th use case is, that an external user can access all orders of his company and should be able to leave comments by adding an event (which is eg a note)

eldy: Do you mean that
restrictedArea($user, 'societe', $socid);
is false when user is an external user, even when  $user->socid = $socid ?

glu000: Yes, restrictedArea is false because it checks if the user has one of this rights while feature is "societe" (line 482 in security.lib.php V16):

'''
$user->rights->$feature->creer
$user->rights->$feature->write
$user->rights->$feature->create
'''

As the external user has none of these rights (which is correct, because otherwise it would mean the user has write rights to thirdparties) the result of the function is false. Please keep in mind that the use case is, that we want to add an event to an element (eg proposal). Of course the user has the right $user->rights->agenda->myactions->create which is necessary to create events.

eldy: So may be we can replace the restrictedArea with a test 
If (socid != User->id) access_forbidden() ?

glu000: I think you ment
```
if ($socid != $user->socid) accessforbidden();
```
This would work. However earlier in ```comm/action/card.php``` on line 93 there is
```
if ($user->socid) {
	$socid = $user->socid;
}
```
So we should replace this part by
```
if ($user->socid) {
	if ($socid != $user->socid) accessforbidden();
}
```
I tested it and it worked.

Please tell me if this would be OK and I will open an new PR

eldy: Yes i meant
`if ($socid != $user->socid) accessforbidden();`
You corrected version looks ok for me...

PR #23804, opened 2023-02-06
FIX attached files and send by mail using trackid from ticket form

lvessiller-opendsi:
FIX attached files and send by mail using trackid from ticket form

**To reproduce**
- attach a file from new ticket form
image
- and click on create
- then no files attached on ticket card
image

**After**
- with this fix an attached file is now on ticket card
image

### DIFF
--- a/htdocs/core/class/html.formticket.class.php
+++ b/htdocs/core/class/html.formticket.class.php
@@ -49,10 +49,10 @@ class FormTicket
 
 	public $track_id;
 
-    /**
-     * @var string 		Email $trackid. Used also for the $keytoavoidconflict to name session vars to upload files.
-     */
-    public $trackid;
+	/**
+	 * @var string 		Email $trackid. Used also for the $keytoavoidconflict to name session vars to upload files.
+	 */
+	public $trackid;
 
 	/**
 	 * @var int ID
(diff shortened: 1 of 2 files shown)


htdocs/core/class/html.formticket.class.php
```diff
@@ -883,7 +889,11 @@ public function clear_attached_files()
 			dol_delete_dir_recursive($upload_dir);
 		}
 
-		$keytoavoidconflict = empty($this->trackid) ? '' : '-'.$this->trackid; // this->trackid must be defined
+		if (!empty($this->trackid)) { // TODO Always use trackid (ticXXX) instead of track_id (abcd123)
+			$keytoavoidconflict = '-'.$this->trackid;
```
eldy, line 893: trackid and track_id are 2 differents fields.
Only  ->trackid must be used for $keytoavoidconflict

lvessiller-opendsi: @eldy 
You're right. 
I corrected it and I only use track_id for key to avoid conflict.

htdocs/ticket/card.php
```diff
@@ -191,6 +191,9 @@
 
 			$object->fk_project = $projectid;
 
+			// File transfer
+			$object->copyFilesForTicket();
```
eldy, line 195: What is the reason to move this before the database action ?
Doing it after allow to avoid to make changes on disks if there is an error on database (due to if !$error test) so nothing is done.
If you do it before, even if there is an error that need to cancel action, file system will be altered by the copyFilesForTicket.  Is there is particular reason ?

lvessiller-opendsi: @eldy 
I put "copyFilesForTicket" before "create" in order to have attached files on ticket card.
If you put after and attach a file, the "trackid" won't be the same and nothing will be attached on ticket card.

eldy: > @eldy I put "copyFilesForTicket" before "create" in order to have attached files on ticket card. If you put after and attach a file, the "trackid" won't be the same and nothing will be attached on ticket card.

I experience this opposite.
copyFilesForTicket() has the line code
`$formmail->trackid = 'tic'.$this->id;`
So this->id must be defined. And to be defined, the record must be created into database.

lvessiller-opendsi: "trackid" is null at begin, when you add the file in the create form.
And if you put "cpyFilesForTicket" after create the "trackid" is "tic".$this->id and $this->id is set.
So the attached file is not found and not put on ticket card.
Perhaps, it will be better to cut the copyFilesForTicket() into separated methods.

eldy: I pushed a different fix...

PR #23820, opened 2023-02-07
FIX substitute project variables in document with free text

lvessiller-opendsi:
FIX substitute project variables in document with free text

**To reproduce**
When you have free text in invoice for example :
image

The generated document don't substitute project variables : 
image

**After**
I moved fetch project function in "getCommonSubstitutionArray()" in order to be able to substitute project variables (not only for mass action)
So mass action and documents use this common method and the free text is now substituted : 
image


htdocs/core/lib/functions.lib.php
```diff
@@ -6990,6 +6990,11 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 				$substitutionarray['__RECEPTIONTRACKNUMURL__'] = 'Shipping tracking url';
 			}
 		} else {
+			// can substitute variables for project
+			if (!empty($conf->projet->enabled) && method_exists($object, 'fetch_project') && is_null($object->project)) {
```
eldy, line 6994: Because method fetch_project is inside the commonobject, it will exists for nearly all objects, even the one that does not have project on them. This means we will make a fetch_object on object where this does not exists.
I would include also a test if property fk_project exists on object to avoid this.
Note the "subtle" difference because:  "feature was never available or not completely implemented for a given use case" is not similar to a "bug". To avoid risk and side effect, when the limit is border-line, i prefer to this such change into develop.

lvessiller-opendsi: @eldy 
I don't understand your point.
The property "fk_project" is also in commonobject and method "fetch_project" return 0 if "fk_project" is empty.

eldy: I have think of this need for a long time. And i think we must implement this differently:  
The method getCommonSubstitutionArray is called very frequently with different context and need.
And we can start to introduct subrequest (fetch_projet) each time we need a new substitution data. In future we will need fetch_thirdparty, fetch_contact, xxx. This means we will make a ton of SQL request a lot of time.
So to avoid this i think a different way to implement your need.
You can add a substitution key  
$substitutionarray[__XXX__@lazyload]  = 'path:class:method'

Then inside the method make_substitutions(), we add a loop to scan $substitutionarray:
For each key ending with '@lazyload', we extract the substitution key '__XXX__' and we check inside the $text (the 1st parameter of  make_substitutions), the the string __XXX__ exists.
If no, we replace with ''.
If yes, we can make the 
```
include_once $path;
$tmpobj = new $class($db);
$valuetouseforsubstitution = $tmpobj->$method(__XXX__)
```

And make the replacement of "__XXX__@lazyload" with valuetouseforsubstitution

So the SQL access or even any complex and heavy code to get a subsitution key value is done if and only if needed.

lvessiller-opendsi: @eldy 
If you put "__XXX___@lazyload" instead of "__XXX__", the final user will have to change all templates with the new "__XXX___@lazyload" key in their email templates or documents. How will he know if he needs to add or not "@lazyload" ?

Before calling the fetch_project method, it checks that project hasn't been loaded yet. So the "fetch_project" method (and SQL) is called only one time in all cases.

Furthermore, when you call 'getCommonSubstitutionArray()', you want to make substitution and call immediately after  "make_substitutions" method. So you got the sames SQL requests.

lvessiller-opendsi:
@eldy 
I improve the lazy load code to substitute project variables in "make_substitutions" function.
I use a memory of each object loaded and use fetch method only one time.

lvessiller-opendsi:
@eldy 
What about this PR ?

eldy:
Moving substitution into the lazymode instead of alwaysbfetching is a good thing.
However PR has been pushed into branch v14 ?

lvessiller-opendsi:
> Moving substitution into the lazymode instead of alwaysbfetching is a good thing. However PR has been pushed into branch v14 when lazymode seems available only in v17 or 18 ?

Lazy load not exist, it's a new concept.
Do you prefer I switch branch into develop ?

lvessiller-opendsi:
@eldy
What about this PR ? 
Do you prefer I switch branch into develop ?

lvessiller-opendsi:
@eldy 
PR #25656 seems to fix project substitution variable, doesn't it ?
Do you think I can do a "fetch_project" on invoice card ? at the begining (after fetch object) ?

eldy:
> @eldy 
> PR #25656 seems to fix project substitution variable, doesn't it ?
> Do you think I can do a "fetch_project" on invoice card ? at the begining (after fetch object) ?

Yes, pr 25656 has fixed for the substitution  from sale order card.
If not for invoice, you can do the same in 14.0 branch

Using the lazy mode solution is also welcome (should work all the time we need it), but must be into develop.

lvessiller-opendsi:
@eldy
I pushed lazy load functionality on #26451

PR #23824, opened 2023-02-07
fix supplier reload page 

thomas-Ngr:
# Fix : On fourn/commande/card.php creation, supplier values were not used

When const RELOAD_PAGE_ON_SUPPLIER_CHANGE is set, the page is reloaded on supplier change.

When reloading the page on supplier change, we want to use the supplier parameters for cond_reglement_id and mode_reglement_id. This was the behavior before PR #22756. In #22756, the parameter &socid has been forgotten.

Another bug is that on a blank creation page, the value of cond_reglement_id is always set. On page reload, the value is always set, so it is not replaced by the wanted value for supplier.

example : 

before : 

1. virgin creation page : cond_reglement_id is always set
before-1

2. after supplier selection : the supplier values have not been used
before-2

after :

1.  virgin creation page : cond_reglement_id is empty
after-1

2. after supplier selection : the supplier values have been used 
after-2

NOTE : **This works only if MAIN_DEFAULT_PAYMENT_TERM_ID is not set.** If MAIN_DEFAULT_PAYMENT_TERM_ID is set, the parameter cond_reglement_id will always be set on the first page. Since it is sent as a POST parameter, it will override the supplier value.

### DIFF
--- a/htdocs/fourn/commande/card.php
+++ b/htdocs/fourn/commande/card.php
@@ -1702,6 +1702,7 @@ if ($action == 'create') {
 						console.log("We have changed the company - Reload page");
 						// reload page
 						$("input[name=action]").val("create");
+						$("input[name=socid]").val($("#socid").val());
 						$("form[name=add]").submit();
 					});
 				});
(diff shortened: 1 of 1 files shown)


eldy:
RELOAD_PAGE_ON_SUPPLIER_CHANGE has been renamed into RELOAD_PAGE_ON_SUPPLIER_CHANGE_DISABLED into develop so page is always reload with a POST (no field should be lost)

eldy:
PR no more required, so i close it.

PR #23833, opened 2023-02-08
FIX: various PHP8 compatibility fixes

marc-dll:
Most of those issues also appear on v17.


eldy:
Target of v15 is php 7.4 max.
Can you push on v17 ?

marc-dll:
Why ? There are no changes in this PR that are not compatible with PHP 5.6 to 7.4. Merging in 15.0 has no negitive impact whatsoever, on the contrary, it makes use of cleaner code using `getDolGlobal*()` functions.

I just tested with php 5.6 and everything's fine !

eldy:
Because target of v15 is php 7.4 max.
So there is no reason to increase the number of commits, increase the number of lines modifying making impact analysis more difficult after merged in higher version , increasing nb of conflicts, etc...
There is an important rule to keep the project maintainable: the minimum and ONLY the minimum must be done into maintenance version and we have no reason to add code for php 8 in v15 because v15 will never be compatible with php8. We must not try to make code to enhance system compatibility of a maintenance version, otherwise it is not a maintenance commit.
History proves that EVERY commit, even the one we are COMPLETELY sure there is no impact, have impact at a moment in the project life.

htdocs/admin/facture.php
```diff
@@ -364,7 +364,7 @@
 							if ($conf->global->FACTURE_ADDON == $file || $conf->global->FACTURE_ADDON.'.php' == $file) {
 								print img_picto($langs->trans("Activated"), 'switch_on');
 							} else {
-								print '<a class="reposition" href="'.$_SERVER["PHP_SELF"].'?action=setmod&token='.newToken().'&value='.preg_replace('/\.php$/', '', $file).'&scan_dir='.$module->scandir.'&label='.urlencode($module->name).'" alt="'.$langs->trans("Default").'">'.img_picto($langs->trans("Disabled"), 'switch_off').'</a>';
```
eldy, line 367: Why removing scan_dir= param ?
Is it done on purpose ?
It is used for ODT generator when a var to store paths for ODT files is required.

marc-dll: @eldy The occurrences I removed are in the context of numbering modules, where this property doesn't exist, and not ODT/PDF modules. I guess it was there because of a bad copy/paste.

PR #23836, opened 2023-02-08
Allow a user to modify his IBAN with the only right user->self->creer

comaiteseb:
FIX

Allow a user to modify his IBAN with the only right user->self->creer
Add a mail notification and an event log on IBAN edition

CF closed PR https://github.com/Dolibarr/dolibarr/pull/23707 with to many conflicts to resolve

### DIFF
--- a/htdocs/core/triggers/interface_20_all_Logevents.class.php
+++ b/htdocs/core/triggers/interface_20_all_Logevents.class.php
@@ -158,6 +158,13 @@ class InterfaceLogevents extends DolibarrTriggers
 			// Initialisation donnees (date,duree,texte,desc)
 			$text = $langs->transnoentities("GroupDeleted", $object->name);
 			$desc = $langs->transnoentities("GroupDeleted", $object->name);
+		} elseif ($action == 'USER_IBANMODIFY') {
+			dol_syslog("Trigger '".$this->name."' for action '$action' launched by ".__FILE__.". id=".$object->id);
+			$langs->load("users");
+			// Initialisation donnees (date,duree,texte,desc)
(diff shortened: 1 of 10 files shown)


comaiteseb:
@eldy you forget me since jan 26th ?

htdocs/core/triggers/interface_20_all_Logevents.class.php
```diff
@@ -158,6 +158,13 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
 			// Initialisation donnees (date,duree,texte,desc)
 			$text = $langs->transnoentities("GroupDeleted", $object->name);
 			$desc = $langs->transnoentities("GroupDeleted", $object->name);
+		} elseif ($action == 'USER_IBANMODIFY') {
```
eldy, line 161: We try to reduce the nb of trigger events. Must match the CRUD. So Create , Modify, Delete.
Here changing the ban for a user is doing an update of the UserBankAccount. So you must use the trigger into the update event of UserBankAccount so USERBANKACCOUNT_MODIFY if you need to generate a particular track in agenda.

eldy:
The PR contains 2 features. Must be split (adding event on a trigger, and changing permission). Non atomic PR are not allowed. Must be splitted in 2.
For Trigger; We try to reduce the nb of trigger events. Must match the CRUD. So Create , Modify, Delete.
Here changing the ban for a user is doing an update of the UserBankAccount. So you must use the trigger into the update event of UserBankAccount so USERBANKACCOUNT_MODIFY if you need to generate a particular track in agenda.

PR #23839, opened 2023-02-08
Add hooks to select_produits_list

marcusdeangabriel:
# NEW
This is a request for enhancement to add hooks to the functions `select_produits_list` and `constructProductListOption` so as to be able to add search information per line in the drop down list of _Predefined products/services to sell_ (language key `PredefinedProductsAndServicesToSell`).

In the function `select_produits_list`, we add two hooks `selectProductsListSelect` and `selectProductsListFrom` modelled after the already existing hook `selectProductsListWhere`, that is, these hooks only add but do not over-ride.

In the function `constructProductListOption`, we add the hook `constructProductListOption` to either over-ride the string with options or to simply add to it.

For example, if available, we add per line of options, a custom (extra) field that we have defined and the supplier reference.  In the image below, that would be _Gizmo_ and then _1614111_ or _9876543210_:
Predefined products, services to sell


frederic34:
@marcusdeangabriel do not forget to fix stickler-ci

marcusdeangabriel:
Stickler-ci fixed.  Good to go for review.

htdocs/core/class/html.form.class.php
```diff
@@ -2542,6 +2542,10 @@ public function select_produits_list($selected = '', $htmlname = 'productid', $f
 		global $langs, $conf;
 		global $hookmanager;
 
+		if (empty($hookmanager->hooks)) {
+			$hookmanager->initHooks(array('commonobject'));
```
eldy, line 2546: context of hook must be initialized by the parent page or by the parent CLI script so context is the same from the start until the end of code.
We must avoid init contact on low level functions.

marcusdeangabriel: As I recall, every card worked as expected except for offers (propale).  Upon examining the `$hookmanager` that I received from the propale card, there was a context, but the hook was empty which is why I ran `initHooks` when the hook in `$hookmanager` was empty.  I bases the above code on code that I found in the file.

Late next week, I can look at why the propale card is not giving me a hook whereupon I can fix it and remove the above could if you would prefer that solution.

marcusdeangabriel:
The error belongs to me.  To activate my hook, I typed `propqlcard` instead of `propalcard` which took me a while to see.  Thus, the offending code block `if (empty(->hooks)) { ... }` is not necessary and has been removed: [Removed not necessary 'if (empty(->hooks)) { ... } block.](https://github.com/Dolibarr/dolibarr/pull/23839/commits/c9369415a226b681783f5b906f783b0a41d05249).

htdocs/core/class/html.form.class.php
eldy: Yes, please, we must find why a proposal card doe not set the hook when you call the select_produits_list().
It should.

marcusdeangabriel:
I need help.  I do not know why this error

`FormTest::testSelectProduitsList`
`Undefined array key "REQUEST_URI"`

is occurring when it was not before, that is, it was working, then it was not, with no change in code on my branch, no change by me. only merges from develop to my branch.

marcusdeangabriel:
I found the error.  It was my error but I am not sure how I introduced it.  Be that as it may, the PR is now good to merge.  Sorry about that.

PR #23843, opened 2023-02-08
Fix (qodana)

KiVoilaJe:
fix function comment wrong return

### DIFF
--- a/htdocs/core/class/fileupload.class.php
+++ b/htdocs/core/class/fileupload.class.php
@@ -488,7 +488,6 @@ class FileUpload
 	 */
 	public function post()
 	{
-		var_dump("here");exit;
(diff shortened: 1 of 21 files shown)


htdocs/core/class/fileupload.class.php
```diff
@@ -489,7 +489,7 @@ public function get()
 	public function post()
 	{
 		if (isset($_REQUEST['_method']) && $_REQUEST['_method'] === 'DELETE') {
-			return $this->delete();
+			echo $this->delete();
```
eldy, line 492: If you remove the return here, it meas, once the action detete has been done, we output the result code on a page, but the code continue and we do the rest of code.
We must keep a return here to stop the action.

PR #23854, opened 2023-02-09
fix : Warning: Trying to access array offset on value of type null in…

grandoc:
… /home/httpd/vhosts/aflac.fr/domains/dev.aflac.fr/httpdocs/core/class/commondocgenerator.class.php on line 1108

### DIFF
--- a/htdocs/core/class/commondocgenerator.class.php
+++ b/htdocs/core/class/commondocgenerator.class.php
@@ -1105,8 +1105,10 @@ abstract class CommonDocGenerator
 	 */
 	public function getColumnContentXStart($colKey)
 	{
-		$colDef = $this->cols[$colKey];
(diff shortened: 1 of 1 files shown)


htdocs/core/class/commondocgenerator.class.php
```diff
@@ -1105,8 +1105,10 @@ public function getColumnContentWidth($colKey)
 	 */
 	public function getColumnContentXStart($colKey)
 	{
-		$colDef = $this->cols[$colKey];
-		return isset($colDef['xStartPos']) ? $colDef['xStartPos'] + $colDef['content']['padding'][3] : null;
+		if (isset($this->cols[$colKey]) && !is_null($this->cols[$colKey])) {
```
eldy, line 1108: if condition is false, the function never return any value.

PR #23860, opened 2023-02-09
Fix: default values would not work for fk_account.

thomas-Ngr:
# Fix : Default value is not used for fk_account on an invoice creation

**Current behavior:**
On page admin/defaultvalues.php we set this parameter :
image

Problem : it is not used on the compta/facture/card.php page (Compte bancaire).
image

But once we select a thirdparty, the value is set to the account configured on the thirdparty (here CIC).
image

And if the thirdparty does not have an account configured, the account stays blank. The default value is never used.

**Expected behavior:**

When default value is set for fk_account:
- on a virgin page, fk_account should not be set.
- If the page has been reloaded with parameter &socid:
  - if fk_account is defined for thirdparty, use this value
  - else, use the default value by using GETPOST() function.

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -2981,7 +2981,9 @@ if ($action == 'create') {
 	}
 
 	// when bank account is empty (means not override by payment mode form a other object, like third-party), try to use default value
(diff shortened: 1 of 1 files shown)


thomas-Ngr:
Thanks @eldy !

PR #23866, opened 2023-02-10
NEW: Invoice time from task, make task note better display in invoiceline

FHenry:
image

Before
image

After
image


FHenry:
fix: #23866

htdocs/projet/tasks/time.php
```diff
@@ -587,7 +587,12 @@
 
 					$arrayoftasks[$object->timespent_id]['timespent'] = $object->timespent_duration;
 					$arrayoftasks[$object->timespent_id]['totalvaluetodivideby3600'] = $object->timespent_duration * $object->timespent_thm;
-					$arrayoftasks[$object->timespent_id]['note'] = $ftask->ref.' - '.$ftask->label.' - '.$username.($object->timespent_note ? ' - '.$object->timespent_note : '');		// TODO Add user name in note
+					$arrayoftasks[$object->timespent_id]['note'] = $ftask->ref.' - '.$ftask->label.' - '.$username;		// TODO Add user name in note
+					if (!empty($conf->fckeditor->enabled) && !empty($conf->global->FCKEDITOR_ENABLE_DETAILS) && !empty($object->timespent_note)) {
```
eldy, line 591: FCKEDITOR_ENABLE_DETAILS on or off is the current state. it may differs from the state of option when data was recorded into database.
For this reason, you must use 
dol_concatdesc() to concat strings without regarding value of FCKEDITOR_ENABLE_DETAILS and module.
so result will be good even if timespent note was recorded hen option differs of the task note.

FHenry: `timespent note` is never with rich text editor (for now).

The goal is to put the  `timespent note`  text after the line label.
With dol_concatdesc it put all on the same line and remove the line returns on `timespent note` 

`$arrayoftasks[$object->timespent_id]['note']=dol_concatdesc($arrayoftasks[$object->timespent_id]['note'],$object->timespent_note);`

will result with 
`TK1607-0004 - Project preparation phase A - Alice Adminson aaaa bbbb ccc`
instead of 
```
TK1607-0004 - Project preparation phase A - Alice Adminson
aaaa
bbbb
ccc
```

eldy: No dol_concatdesc does not concat on same line.
It concats on a new line taking care for you if content is text or html.

FHenry: in this case there is a bug in dol_concatdesc
The new commit fix it

eldy:
Your change introduce a bug into dol_concatdesc as reported by phpunit (regression in dol_concatdesc).
I do not find any troubles with the current version of dol_concatdesc. For which entry parameters do you have a wrong output according to you ? And what is the output you expect ? (please report value as a var_dump result and not as a text string so we can see if there is \n or not if result is a full string)

FHenry:
@eldy 
I've review the code, and use dol_concatdesc properly without change it

PHPCS (base on dolibarr ruleset) make lot's of change in space in time.php to check without changes 
https://github.com/Dolibarr/dolibarr/pull/23866/files?w=1

eldy:
> @eldy 
> I've review the code, and use dol_concatdesc properly without change it
> 
> PHPCS (base on dolibarr ruleset) make lot's of change in space in time.php to check without changes 
> https://github.com/Dolibarr/dolibarr/pull/23866/files?w=1
> 

@FHenry There is an option on github to ignore spaces when comparing change file. But only on desktop web version.

FHenry:
@eldy this is an old pr, but all flags are green, I've made the change regarding your review. 
I let you check, thank's

htdocs/projet/tasks/time.php
```diff
@@ -1047,12 +1043,12 @@
 
 			// Description
 			print '<td class="titlefield tdtop">'.$langs->trans("Description").'</td><td>';
-			print dol_htmlentitiesbr($projectstatic->description);
+			print nl2br($projectstatic->description);
```
eldy, line 1046: Why replacing with nl2br ?
description is text of html, and to ouput this on screen we should use
dol_htmlentitiesbr

FHenry: True, fixed

FHenry:
@eldy this have all test OK. You've put a label discuss, may be you want to have more information about it ?

Issue #20521, opened 2022-04-01
BUG regression from v14 - Conversion of time to days and hours for display


priojk:
### Bug

**The problem:**

I noticed this regression on the intervention card and pdf, but it probably is also present in different places:

**A formatted duration which contains days and hours is abbreviating the expression for day in an "unconventional" way.**

Example: Duration of 27 hours ist displayed as **3 d. 03:00** (in English) or **3 t. 03:00** (in German).

At least in German, t. is no valid abbreviation for Tag (sg.) or Tage (plural), but also d. in English seems not to be a usual abbreviation.

**Analysis:**

Trying to change the expression "t." via translation was not possible, this is no word from transifex.

The output for the duration is created with the following function: 
`convertSecondToTime($object->duration, 'all', $conf->global->MAIN_DURATION_OF_WORKDAY)`
which is defined in core/lib/date.lib.php

beginning from line 280, there is the following code:

```
                    if ($sDay > 0) {
			$dayTranslate = $langs->trans("Day");
			if ($sDay > 1) {
				$dayTranslate = $langs->trans("Days");
			}
			$sTime .= $sDay.' '.strtolower(dol_substr($dayTranslate, 0, 1)).'. ';
		}
```

So, the word for day is correctly translated and also correctly handle in singular and plural form (1 Day, 2 Days).

But then, the word gets downcased, and just the first letter taken, followed by '.'.

This explains, how e.g. **3 Tage 03:00** ends up as **3 t. 03:00**

The line downcasing and abbreviating the word for day got introduced around 13 months ago with the commit message "Look and feel v14" by @eldy.

I assume, there are locations in the UI, where the space is limited, so there is a good reason to make the output shorter.

Unfortunately, in particular the intervention pdf stating the spent times is customer-facing, so some improvement would be nice.

Since I currently don't know where the shorter version is required, I am unsure, if we need two output variants from convertSecondToTime or if we could just revert the change.

Would be good, if @eldy could advise.

PR #23877, opened 2023-02-10
Fix: #20521 removed downcasing and abbreviation of already translated term for day

priojk:
Best is to see the long description in the issue #20521.

Brief version:
Word for day gets translated into target language.
But then it got downcased, the first character taken and a . added.

This is unfortunately not the way, the word for Day gets abbreviated in most languages.

If a one letter version is really needed somewhere (could not find a place so far), it would be better to create a dedicated translation key and translate it in other languages as needed.

### DIFF
--- a/htdocs/core/lib/date.lib.php
+++ b/htdocs/core/lib/date.lib.php
@@ -286,7 +286,7 @@ function convertSecondToTime($iSecond, $format = 'all', $lengthOfDay = 86400, $l
 			if ($sDay > 1) {
 				$dayTranslate = $langs->trans("Days");
 			}
-			$sTime .= $sDay.' '.strtolower(dol_substr($dayTranslate, 0, 1)).'. ';
+			$sTime .= $sDay.' '.$dayTranslate.' ';
 		}
 
 		if ($format == 'all') {


priojk:
Ok, the Travis CI fails exactly because of the suggested correction. 

@eldy would be good if you could advise how to proceed.
Currently, instead of "Tage" in German, the date is using "t.", which makes no sense in this language and cannot be corrected via Transifex.

There was 1 failure:
1) DateLibTest::testConvertSecondToTime
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'1 d.'
+'1 Day'
/home/travis/build/Dolibarr/dolibarr/test/phpunit/DateLibTest.php:347

htdocs/core/lib/date.lib.php
```diff
@@ -286,7 +286,7 @@ function convertSecondToTime($iSecond, $format = 'all', $lengthOfDay = 86400, $l
 			if ($sDay > 1) {
 				$dayTranslate = $langs->trans("Days");
 			}
-			$sTime .= $sDay.' '.strtolower(dol_substr($dayTranslate, 0, 1)).'. ';
+			$sTime .= $sDay.' '.$dayTranslate.' ';
```
eldy, line 289: Hi we already have the translation for first letter of day into language file, into product.lang we have for english
d=d
for french
d=j

I suppose you need d=T for german ?  So i suggest to fix with 

$langs->load("products");
$sTime .= $sDay.' '.$langs->trans("d").' ';

Then to fix the german translation for d into transifex...

priojk: Hi,
that's a perfect solution. Actually, the one letter abbreviation for Day in German is d=d (coming from dies in Latin for whatever reason...). I just checked in Transifex and the translation was already ok.

eldy:
So i think i can close this PR. Fix done manually.

PR #23889, opened 2023-02-12
New: added help link for page in German

priojk:


priojk:
Need to also add language detection for DE, will create a new pull request.

PR #23893, opened 2023-02-13
doc

frederic34:


htdocs/asset/class/asset.class.php
```diff
@@ -157,6 +157,11 @@ class Asset extends CommonObject
 	public $status;
 	public $user_cloture_id;
 
+	/**
+	 * @var Asset object oldcopy
+	 */
```
eldy, line 162: The $oldcopy is not already into rhe CommonObject class ?

frederic34: yes but as CommonObject

frederic34: image
these properties exist on Asset but not on CommonObject

PR #23896, opened 2023-02-13
FIX problem when search boolean (Yes/No) extrafields with value 0 (No)

hregis:


hregis:
@eldy same in v15 to develop

htdocs/core/tpl/extrafields_list_search_input.tpl.php
```diff
@@ -47,7 +47,7 @@
 					if (in_array($typeofextrafield, array('link', 'sellist', 'text', 'html'))) {
 						$morecss = 'maxwidth200';
 					}
-					echo $extrafields->showInputField($key, (empty($search_array_options[$search_options_pattern.$tmpkey]) ? '' : $search_array_options[$search_options_pattern.$tmpkey]), '', '', $search_options_pattern, $morecss, 0, $extrafieldsobjectkey, 1);
+					echo $extrafields->showInputField($key, ((empty($search_array_options[$search_options_pattern.$tmpkey]) && $search_array_options[$search_options_pattern.$tmpkey] != 0) ? '' : $search_array_options[$search_options_pattern.$tmpkey]), '', '', $search_options_pattern, $morecss, 0, $extrafieldsobjectkey, 1);
```
eldy, line 50: According to line 47, we are not in the case of bolean valu here but in the case of html, text, link or sellist. 
Did i miss something ?

hregis: @eldy boolean use this showInputField !

hregis: @eldy create a boolean for propal, go to list and search "No", the value "No" is not kept, with my correction the value "No" remains following the search

hregis:
@eldy ping

eldy (commented):
Seems fixed in recent version, but in a different way. To avoid pb in merging, i will close this pr. If we need a backport, using the same fix than the one done for recent version must be used.

PR #23897, opened 2023-02-13
NEW inc.php: handle parameters from argv

alexandre-janniaux:
# NEW|New inc.php: handle parameters from argv

This commit adds support for --help and --config to provide a different config file when running upgrade.php from commandline. The argv array is patched afterwards as if the options were never given so that it stays transparent for the code path that are not aware of those arguments.

The intention behind is incrementally move the usage of argc/argv to this location and force help/usage from there, in particular for upgrading.

The rationale for the --config addition is to be able to provide a different path for conf.php when multiple dolibarr instance are using the same readonly htdocs folder, which is already possible by modifying the `include_path` from PHP for the htdocs/ directory but not for the htdocs/install directory since relative paths are used to fetch the config file. Since the use-case is to upgrade/migrate a Dolibarr instance from CLI, it makes sense to be able to select for which instance (and database parameters) the upgrade should take place.

---

Hi, this is my first merge request to the project. I've tried to discuss this point on the Discord but we can discuss here or I can first create an issue if needed. I wasn't sure whether I should create the issue first since it would have included multiple point on upgrade.php CLI usage and maybe that's not the target for the project? However, if it was preferred, I'll create new issues before sending more patches on this topic.

### DIFF
--- a/htdocs/install/inc.php
+++ b/htdocs/install/inc.php
@@ -6,6 +6,7 @@
  * Copyright (C) 2012       Marcos García           <marcosgdf@gmail.com>
  * Copyright (C) 2016       Raphaël Doursenaud      <rdoursenaud@gpcsolutions.fr>
  * Copyright (C) 2021       Charlene Benke      	<charlene@patas-monkey.com>
+ * Copyright (C) 2023       Alexandre Janniaux      <alexandre.janniaux@gmail.com>
 *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


alexandre-janniaux:
Note: maybe this PR requires more testing, in particular in scenario where update is not done from the CLI to ensure it breaks nothing. I'm trying to see how this could be integrated in the Travis CI but if it's too hard, I'll have specific tests for that in my pipeline too.

Also, don't hesitate to postpone for after 17.0 if needed. I currently apply the patch on the version I need during deployment and it applies back to 15.0 up to 16.0.4 so it's not an issue for me.

alexandre-janniaux:
I realized I was missing some points from the developer instructions, now fixed after some PHP tooling learning:
 - copyright line has been added.
 - I changed string substitution by string concatenation in `usage()`.
 - The new `usage()` function now has documentation and doesn't trigger any php-codesniffer warning.
 - I adjusted the commit (and git repo) to use my personal e-mail.

If this gets accepted, I guess the next step will consist of providing more features for format (html vs plain) to provide console reports during upgrade, and check that a failed upgrade triggers a failure in the script.

htdocs/install/inc.php
```diff
@@ -70,6 +71,67 @@
...
+	echo "      Provide a different conf.php file to use.\n";
+	echo "\n";
+	echo "  -h, --help:\n";
+	echo "      Display this help message.\n";
+}
+
+if (php_sapi_name() === "cli") {
+	$rest_index = 0;
+	$opts = getopt($short_options, $long_options, $rest_index);
+	if ($rest_index < $argc - 2) {
```
eldy, line 109: There is another case that can be used when upgrade2 is ran as a CLI programm. this is the syntax:

upgrade2.php  0.0.0  0.0.0  MAIN_MODULE_XXX,MAIN_MODULE_YYY,...
So 3 paramaters and $argv[3] is used in such a case.
Can you fix your test to allow this case ?

alexandre-janniaux: Sure, I'll fix that and check.

In the mean time, do think it's ok to keep the cli parsing there? What I think about was dedicating a cli.php which would be including upgrade.php and upgrade.php2 depending on what's needed, allowing cli.php to be in scripts/ instead, but that's currently a bit convoluted because of how inc.php is provided for the upgrade scripts. I'm not sure where I should discuss this, maybe next DevCamp?

eldy: CLI parsing inside the inc.php is a fist good step.

Moving it into a  script/dolibarr-tools.php may be possible. But it should be a wrapper that execute the most as possible existing codes and not contains logical business code, just a launcher to call others.

alexandre-janniaux: Ok, done with a few fixes (short options) and changes (moving options down to allow parsing -h directly). Do I need to add more tests for this in travis?

PR #23899, opened 2023-02-13
doc

frederic34:


htdocs/fourn/class/fournisseur.facture.class.php
```diff
@@ -243,6 +243,12 @@ class FactureFournisseur extends CommonInvoice
 	 */
 	public $lines = array();
 
+	/**
+	 * Invoice line
+	 * @var SupplierInvoiceLine
+	 */
+	public $line;
```
eldy, line 250: Can't find where we are using the property ->line ?
We should not need it.

frederic34: $this->line = $this->lines[$i];

https://github.com/Dolibarr/dolibarr/blob/d8e59ab19c838d4110ad36cf099ef2be465ef0f0/htdocs/fourn/class/fournisseur.facture.class.php#L1827

eldy: It seems it was a bad coding, because it was set and seems not used after, except to do unset($this->line) few lines after.

PR #23905, opened 2023-02-14
FIX id null generates errors

BenjaminFlr:
FIX

If the id in array is null, it generates an error with SQL request on rowid.

### DIFF
--- a/htdocs/categories/class/categorie.class.php
+++ b/htdocs/categories/class/categorie.class.php
@@ -1464,7 +1464,7 @@ class Categorie extends CommonObject
 		$ways = array();
 
 		$parents = $this->get_meres();
-		if (is_array($parents)) {
+		if ((is_array($parents)) && ($parents['id'] > 0)) {
 			foreach ($parents as $parent) {
 				$allways = $parent->get_all_ways();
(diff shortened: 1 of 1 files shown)


htdocs/categories/class/categorie.class.php
```diff
@@ -1464,7 +1464,7 @@ public function get_all_ways()
 		$ways = array();
 
 		$parents = $this->get_meres();
-		if (is_array($parents)) {
+		if ((is_array($parents)) && ($parents['id'] > 0)) {
```
eldy, line 1467: $parents['id'] seems to never exists.
into get_meres, return value is filed with  $parents[] = $cat;
So index is an autogenerated number.

altairis-benjamin: The case encountered seems to be due to an external module. Can we let this conversation open, so I'll be able to confirm or not in the following days ? If it's actually due to the module, I'll close this PR.

eldy: A module that affect the content returned by get_meres() ?
If yes, a better change would be to output an error if $parents content is not in the good format so if a module make something wrong, we will immediately see it.

PR #23906, opened 2023-02-14
Fix (qodana) 

KiVoilaJe:
Fix (qodana) Incompatible return type


htdocs/install/upgrade2.php
```diff
@@ -4222,7 +4222,7 @@ function migrate_delete_old_dir($db, $langs, $conf)
  * @param	Conf		$conf			Object conf
  * @param	array		$listofmodule	List of modules, like array('MODULE_KEY_NAME'=>', $reloadmode)
  * @param   int         $force          1=Reload module even if not already loaded
- * @return	int							<0 if KO, >0 if OK
+ * @return	int|void					<0 if KO, >0 if OK
```
Hystepik, line 4225: The problem that trigger the codding error here is that the first return of the function is empty (return;) line 4230.
Maybe replace return; by return 0; could be a better fix so if the function does something then >0 for success , <0 for error and if nothing done then 0

htdocs/contact/class/contact.class.php
```diff
@@ -1868,7 +1868,7 @@ public function getContactRoles($element = '')
 	 * Updates all roles (default contact for companies) according to values inside the ->roles array.
 	 * This is called by update of contact.
 	 *
-	 * @return float|int
+	 * @return void|int
```
Hystepik, line 1871: Here the first return is empty maybe you could return 0 to specify that nothing was done. This done you can remove void return

Hystepik:
@eldy seams ok to me

PR #23914, opened 2023-02-15
fix : Warning: Undefined variable  in /home/httpd/vhosts/aflac.fr/dom…

grandoc:
…ains/dev.aflac.fr/httpdocs/fichinter/list.php on line 364

### DIFF
--- a/htdocs/fichinter/list.php
+++ b/htdocs/fichinter/list.php
@@ -361,7 +361,7 @@ $num = $db->num_rows($resql);
 
 
 // Direct jump if only one record found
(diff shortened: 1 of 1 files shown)


htdocs/fichinter/list.php
```diff
@@ -361,7 +361,7 @@
 
 
 // Direct jump if only one record found
-if ($num == 1 && !empty($conf->global->MAIN_SEARCH_DIRECT_OPEN_IF_ONLY_ONE) && $search_all && !$page) {
+if ($num == 1 && !empty($conf->global->MAIN_SEARCH_DIRECT_OPEN_IF_ONLY_ONE) && isset($search_all) && !$page) {
```
eldy, line 364: Result is not the same.
Use instead !empty($search_all)

PR #23921, opened 2023-02-15
New : add a new ajax file to catch flowjs process

Hystepik:
Add the server side program to catch flowjs client request


htdocs/core/ajax/flowjs-server.php
```diff
@@ -0,0 +1,161 @@
...
+}
+// If there is no need to load and show top and left menu
+if (!defined("NOLOGIN")) {
+	define("NOLOGIN", '1');
+}
+//if (! defined('NOREQUIRETRAN'))  define('NOREQUIRETRAN','1');    // Required to know date format for dol_print_date
+
+// Load Dolibarr environment
+require '../../main.inc.php';
+require '../../core/lib/files.lib.php';
```
eldy, line 46: Only link to the main.inc must be relative and using require.

Once the main is loaded, we must use absolut link using DOL_DOCUMENT_ROOT.'/core/lib/files.lib.php'; and you must use require_once

PR #23926, opened 2023-02-15
move log after variable updates to get good informations in logs

rycks:
between 15.0 and 16.0 a lot of code was moved down in that file ... then syslog call is now set too early, for example we don't have information about bcc !

### DIFF
--- a/htdocs/core/class/CMailFile.class.php
+++ b/htdocs/core/class/CMailFile.class.php
@@ -142,7 +142,7 @@ class CMailFile
 	{
 		global $conf, $dolibarr_main_data_root, $user;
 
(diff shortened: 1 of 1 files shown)


eldy:
This depends on the goal of the log. Here the goal is to see the start of the method call and to know what is value of  entry parameters. So, when we found the line in a production system, we can follow the code un function, knowing all the starting values.
Moving too low, when we exit the method before reaching the log, we have no log at all, and we even don't know that the method was called. Also if values in log are not the one we expect, we have no chance to know why. We must keep the log at begin to track entry point, but may be we can add more to track change into values...

rycks:
Just one example:
- In v15 that logs are after addr_bcc set
- but In v16+ $addr_bcc will be always empty if we keep syslog call at the same place

rycks:
Here is an alternative : put logs at real start of the method then a new syslog with what could be updated by code during that part of code

PR #23927, opened 2023-02-15
Add tags for Orders

frederic34:
this a postponed fixed PR from Pr/florent poinsaut/22849


FHenry:
Shouldn't we migrate tag in "One table" system like "element_element" ?
The tag system growing version form version and migrate it will be more work in the future.
We've discuss this point with eldy at a devcamp and he was open to a rewrite of the feature, but I agree right now it's easier to add table than rewrite all the mechanics and i'm still not sure about  what is the best design regarding the data quantity and database engine.
On other ERP there is both solutions but the "One table" system seems to have a the up trend.
May be at the next devcamp a Workshop around this subject, even in Devcamp we prefer to debug and not rewrite all mechanics and create more bugs

hregis:
@FHenry yes one table for all element is more simple ! then you have to think about the indexes, will it be more efficient in terms of sql queries when there is a lot of data?!

hregis:
@frederic34 @FHenry maybe have a performance to transpose element name with element id for a better indexes ?

hregis:
@FHenry or maybe change Mysql with NoSQL ?? oh...that reminds me of something! ;-)

PR #23950, opened 2023-02-17
Add ISO 3166-1 alpha2 alpha3 and numeric codes to llx_c_country

Daviid-P:
# NEW|New Add ISO 3166-1 Codes to llx_c_country

The numeric code is missing so I decided to add `alpha2_code`, `alpha3_code` and `numeric_code` columns to the llx_c_country table.

I realize that `code` and `code_iso` are most likely alpha2 and alpha3 but changing the names to be more clear would be a breaking change, hence why I just added new columns.

Any advice is welcome.


Daviid-P:
BTW Kosovo is missing in my PR since it's not recognized by ONU and not included in ISO 3166-1.

llx_c_country
---
| rowid | code | code_iso | label | active | favorite | eec | alpha2_code | alpha3_code | numeric_code | 
| ---: | --- | --- | --- | ---: | ---: | ---: | --- | --- | ---: | 
| 247 | XK | XKX | Kosovo | 1 | 0 | 0 |  |  | 0 |

htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -396,3 +396,257 @@ ALTER TABLE llx_c_tva ADD COLUMN use_default tinyint DEFAULT 0;
 
 ALTER TABLE llx_commande_fournisseurdet MODIFY COLUMN ref varchar(128);
 ALTER TABLE llx_facture_fourn_det MODIFY COLUMN ref varchar(128);
+
+ALTER TABLE llx_c_country ADD COLUMN alpha2_code CHAR(2) NOT NULL;
```
eldy, line 400: the alpha2 is already present, it is the field "code"
the alpha2 is also present, it is the field "code_iso"

Daviid-P: Personally, I took it from wikipedia.
I believe the "official" list can, unfortunately, be purchased here https://www.iso.org/publication/PUB500001.html

As for code and code_iso, as I mentioned alpha2 and alpha3 would be more specific but renaming the columns would be a breaking change, right?

Also having a code_iso makes it seem as if the other code isn't ISO, which I guess it's only true for Kosovo.

eldy: Why are they more specific ?
code on 2 and 3 are iso fields so they are standardized so there is no reason to have 2 official codes for the same country.
For country not defined by iso, we can still store the value inside this existing field.

Daviid-P: If both are already iso why is only one of them identified as `_iso`?

Why are they more specific? Because thats what the ISO calls them, "alpha2" and "alpha3".

There is no reason to have 2 official codes, if it was up to me I'd also remove code and code_iso, but since some people might be using them in their own code/modules I did not delete them.

Ideally people would read the changelog and update their code from code to alpha_2_code and code_iso to alpha_3_code, but many will not.

A country being recognized by the ONU was just a comment as for why Kosovo was left empty, apparently although not official it's still in ISO 3166-1 as "user-assigned codes"

eldy: So your goal is only to rename the column "code" into "alpha_2" and "code_iso" into "alpha_3" ?
Why changing the name of this column ? What is the enhancement ?

Daviid-P: - *So your goal is only to rename the column "code" into "alpha_2" and "code_iso" into "alpha_3" ?*
And also to add the numeric code.

- *Why changing the name of this 2 columns?*
`code` is a column name, `code_iso` is a column name that makes it seems as if the previous `code` column is not ISO even though it is (because one is specifically named `_iso` while the other is not), **to make it more clear** I decided to rename them to `alpha_2_code` and `alpha_3_code`.

- *For which purpose ?*
Because I think it makes it clear that both columns are standard ISO defined values and we also use the names they use in the ISO 3166-1 specification.


- *Why not just add the numeric column and rename the other columns?*
Because I think there could be some people that have SQL sentences in their own code that directly access the columns `code` and `code_iso` so I didn't want to break their code. Since this is not a change that will show them a notice of deprecation like with class properties and good IDEs I decided to leave them.

eldy: So the conclusion is that we duplicate data into different column just to have a name of a column more clear.
Why   "alpha_3_code" is more clear than  "code_iso" that is a varchar of 3 chars ?
And what is goal of numeric code ?  Is it a standard value ?

Daviid-P: - *Why "alpha3_code" is more clear than "code_iso" that has the type varchar of 3 chars ?*
Because it's what the ISO 3166-1 call the 3 letter code.

- *And what is goal of the numeric code ?*
It's used by some other accounting programs (personally I had problem when trying to upload my clients to [Contasol](https://www.sdelsol.com/programa-contabilidad-contasol/) and they told me they needed the numeric code from ISO 3166-1, then I found that Dolibarr did not have this in llx_c_country)

- *Is it a standard value defined somewhere ? If yes, where ?*
It's defined in ISO 3166-1 just like Alpha 2 and Alpha 3 codes.


We break the best practices of never duplicating data in the database structure into a different column to not break other peoples modules in case they use the `code` and `code_iso` columns directly. At least until developers have had  sufficient time to update their code.

But if you want me to I can change the PR with:
```sql
ALTER TABLE llx_c_country RENAME COLUMN code TO alpha2_code;
ALTER TABLE llx_c_country RENAME COLUMN code_iso TO alpha3_code;
ALTER TABLE llx_c_country ADD COLUMN numeric_code INT NOT NULL;


UPDATE llx_c_country SET numeric_code = 4 WHERE alpha3_code = "AFG";
UPDATE llx_c_country SET numeric_code = 248 WHERE alpha3_code = "ALA";
UPDATE llx_c_country SET numeric_code = 8 WHERE alpha3_code = "ALB";
.
.
.
.
.
```

If what you're asking is where you can find the official documents for ISO 3166-1 then I can't help you, anything official I can find is behind a paywall.

https://www.iso.org/standard/72482.html 150$ just for country codes
https://www.iso.org/publication/PUB500001.html 324$ for Country code, Country subdivision code and Code for formerly used names of countries.

Technically there's no reason to believe anything published by someone else but...

https://github.com/lukes/ISO-3166-Countries-with-Regional-Codes/blob/master/all/all.csv
https://gitlab.com/pasosdeJesus/division-politica/-/blob/main/PAISES_ISO_2022-ampliado.csv
https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes

Other than that maybe some university close to you has access


Okay, I kept searching and found this https://www.iso.org/obp/ui/#search/code/ not sure what's the difference between this freely available page and https://www.iso.org/standard/72482.html

eldy: OK. So let's go to add the numeric code. 
However type varchar(3) seems better. According to iso it is a numeric on 3 chars and the zero at begin seems expected.
However, we can keep code and code_iso. ISO say code are alpha2 or 3 to say it is an alpha on 2 or 3. This is the type of the data, it does not means we must have the same name. code and code_iso are currently used by code so we must keep them, we have not enough reason to rename them.
Can you modify the PR, to just add a third field called  "code_numeric" with type varchar(3) ?

Daviid-P: Ok, left only numeric code and changed it to CHAR(3) with leading zeros.
Also changed to develop

eldy:
I have merged the PR manually to solve 2 pb:
- PR has structure change so must be pushed into develop
- PR was supported migration but not first install, so i also copy the update into install sql file.

PR #23951, opened 2023-02-17
NEW add bookmarks in selectable landing pages for users

altairis-noe:
# NEW: Allow users to select internal bookmarks as landing pages
image

### DIFF
--- a/htdocs/user/param_ihm.php
+++ b/htdocs/user/param_ihm.php
@@ -214,6 +214,28 @@ if (isModEnabled('agenda')) {
 if (isModEnabled('ticket')) {
 	$tmparray['ticket/list.php?mainmenu=ticket&leftmenu='] = 'Tickets';
 }
+// add bookmarks to available landing pages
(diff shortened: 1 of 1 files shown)


htdocs/user/param_ihm.php
```diff
@@ -214,6 +214,28 @@
...
+	$sql .= " WHERE 1=1";
+	$sql .= " AND b.entity IN (".getEntity('bookmark').")";
+	$sql .= " AND b.url NOT LIKE 'http%'";
+	if (!$object->admin) {
+		$sql .= " AND (b.fk_user = ".((int) $object->id)." OR b.fk_user is NULL OR b.fk_user = 0)";
+	}
+	$resql = $object->db->query($sql);
+	if ($resql) {
+		$i = 0;
+		while ($i < $object->db->num_rows($resql)) {
```
eldy, line 231: Can you move the $object->db->num_rows($resql) out of the while. Do it before and sav it into $num_rows then use num_rows into the while.

PR #23959, opened 2023-02-19
fix : Warning: Undefined property: FactureFournisseur:: in /home/http…

grandoc:
…d/vhosts/aflac.fr/domains/dev.aflac.fr/httpdocs/fourn/class/fournisseur.facture.class.php on line 2212


htdocs/fourn/class/fournisseur.facture.class.php
```diff
@@ -229,6 +229,7 @@ class FactureFournisseur extends CommonInvoice
 
 	public $mode_reglement_id;
 	public $mode_reglement_code;
+	public $special_code;
```
eldy, line 232: There is no column special code on supplier invoice. So we should not have this property.

PR #23984, opened 2023-02-21
Fix LDAP Authentification failure with error 500

Yanonix:
# FIX LDAP Authentification failure with error 500
This PR fix the LDAP Authentification when an username/password couple doesn't exist in LDAP and throws an error 500.

This bug was fixed in 6a52b96 (PR: #19726) , but was broken with massive rename of `$ldap->close()` to `$ldap->unbind()` in f497492

It is happening because there are too many closing LDAP connection calls. I'm sorry, I doesn't have the time to fix the source of the problem (with removing `$ldap->unlink()` spares) and don't want to break anything else. That's why I have only patched the closing function instead.

It will be nice to have this fix in 16.0.x version.

Thank you ;)

### DIFF
--- a/htdocs/core/class/ldap.class.php
+++ b/htdocs/core/class/ldap.class.php
@@ -396,7 +396,7 @@ class Ldap
 	public function unbind()
 	{
 		$this->result = true;
-		if (is_resource($this->connection)) {
+		if (is_resource($this->connection) || is_object($this->connection)) {
 			$this->result = @ldap_unbind($this->connection);
 		}
 		if ($this->result) {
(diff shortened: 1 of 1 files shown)


htdocs/core/class/ldap.class.php
```diff
@@ -401,7 +396,7 @@ public function bindauth($bindDn, $pass)
 	public function unbind()
 	{
 		$this->result = true;
```
eldy, line 398: this->connection can be a resource or (php8.1) a LDAP\Connection
Code must be compatible for both case. The new code seems ko with php 8.1. May be you can do 
`if (is_resource($this->connection) || is_object($this->connection)) {`

PR #23986, opened 2023-02-21
FIX - payment mode and payment term in search list

BB2A-Anthony:
# FIX|Fix #[*payment mode and payment term for search list*]
[*Long description*]
if const 'MAIN_DEFAULT_PAYMENT_TYPE_ID' or 'MAIN_DEFAULT_PAYMENT_TERM_ID' is set select is never empty for payment method and payment terms in the search list.


eldy:
I think the problem is not into current code of page but into the option MAIN_DEFAULT_PAYMENT_TYPE_ID and MAIN_DEFAULT_PAYMENT_TERM_ID that are badly implemented. they are implemented into a presentation method meaning that default will overwrite selection each time we show the selectbox.
Your fix will work but it is a hack on each page the use the selectbox making code not standard compared to rest of code.
This 2 variables should be removed and instead we should use the generic feature of dolibarr to set default values.
Did you try to set the default value using the menu Home - Setup - Default value ?

BB2A-Anthony:
> I think the problem is not into current code of page but into the option MAIN_DEFAULT_PAYMENT_TYPE_ID and MAIN_DEFAULT_PAYMENT_TERM_ID that are badly implemented. they are implemented into a presentation method meaning that default will overwrite selection each time we show the selectbox. Your fix will work be it is a hack on each page the use the selectbox making code not standard compared to rest of code. This 2 variables should be removed and instead we should use the generic feature of dolibarr to set default values. Did you try to set the default value using the menu Home - Setup - Default value ?

I tested and it is not conclusive to select availability_id, cond_reglement_id, mode_reglement_id,shipping_method_id,demand_reason_id

eldy:
> availability_id

Why ? For which URL of list did you try and what happen or not happen ?

BB2A-Anthony:
If I put this parameter in the default values it is not taken into account when creating an order

image

PR #23991, opened 2023-02-21
fix #23980 : max length for RUM

### LINKED ISSUE #23980: Génération du RUM dépasse la taille de la BDD
### Bug

Lors de la génération d'un RUM pour les prélèvements, le RUM généré est trop long (>32 car)

### Environment Version

16.0.3

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

Créer un client avec un code client long Exemple : PERFORMANCECONSEIL
Saisir un compte bancaire avec génération du RUM 
`Data too long for column 'rum' at row 1`

### Attached files

_No response_

rycks:
# FIX #23980 

Shrink customer account name in case of max length reached

### DIFF
--- a/htdocs/compta/prelevement/class/bonprelevement.class.php
+++ b/htdocs/compta/prelevement/class/bonprelevement.class.php
@@ -1697,8 +1697,18 @@ class BonPrelevement extends CommonObject
 	public static function buildRumNumber($row_code_client, $row_datec, $row_drum)
 	{
 		global $langs;
+		$maxRumLength = 34 - 2; // look at iban_prefix in account.class.php
 		$pre = substr(dol_string_nospecial(dol_string_unaccent($langs->transnoentitiesnoconv('RUM'))), 0, 3); // Must always be on 3 char ('RUM' or 'UMR'. This is a protection against bad translation)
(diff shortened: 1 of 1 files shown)


htdocs/compta/prelevement/class/bonprelevement.class.php
```diff
@@ -1697,8 +1697,18 @@ public function generate($format = 'ALL', $executiondate = '', $type = 'direct-d
 	public static function buildRumNumber($row_code_client, $row_datec, $row_drum)
 	{
 		global $langs;
+		$maxRumLength = 34 - 2; // look at iban_prefix in account.class.php
```
eldy, line 1700: All what e need for a rum is a random number.
I think for a simple random number adding so many condition and code is useless.
in v17, i replaced with

`return $pre.'-'.dol_print_date($row_datec, 'dayhourlogsmall').'-'.dol_trunc($row_drum.($row_code_client ? '-'.$row_code_client : ''), 15, 'right', 'UTF-8', 1);
`

PR #23994, opened 2023-02-22
NEW : Add new structure for invoice situation - GIF Situation

aspangaro:
@eldy 

Propose new structure to stock invoice situation with amount cumulated.

Next step when the structure will be validated : 
- move field total_ht, total_tva, total_ttc with cumulated amount to the new field
- Stock amount on total_ht, total_tva, total_ttc not cumulated

### DIFF
--- a/htdocs/install/mysql/migration/17.0.0-18.0.0.sql
+++ b/htdocs/install/mysql/migration/17.0.0-18.0.0.sql
@@ -82,4 +82,13 @@ ALTER TABLE llx_supplier_proposal ADD INDEX idx_supplier_proposal_fk_user_valid 
 ALTER TABLE llx_supplier_proposal ADD INDEX idx_supplier_proposal_fk_projet (fk_projet);
 ALTER TABLE llx_supplier_proposal ADD INDEX idx_supplier_proposal_fk_account(fk_account);
 
+ALTER TABLE llx_facturedet ADD COLUMN situation_percent_cumulated real DEFAULT 100 AFTER fk_prev_id;
+ALTER TABLE llx_facturedet ADD COLUMN situation_total_ht_cumulated double(24,8) AFTER situation_percent_cumulated;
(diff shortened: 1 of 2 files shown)


htdocs/install/mysql/migration/17.0.0-18.0.0.sql
```diff
@@ -82,4 +82,13 @@ ALTER TABLE llx_supplier_proposal ADD INDEX idx_supplier_proposal_fk_user_valid
 ALTER TABLE llx_supplier_proposal ADD INDEX idx_supplier_proposal_fk_projet (fk_projet);
 ALTER TABLE llx_supplier_proposal ADD INDEX idx_supplier_proposal_fk_account(fk_account);
 
+ALTER TABLE llx_facturedet ADD COLUMN situation_percent_cumulated real DEFAULT 100 AFTER fk_prev_id;
+ALTER TABLE llx_facturedet ADD COLUMN situation_total_ht_cumulated double(24,8) AFTER situation_percent_cumulated;
```
eldy, line 86: We don't need this information.

We already have a chain between each situation invoice. The chain is build with existing fields
situation_cycle_ref    and    situation_counter

And the "cumulated amount" is the sum of each invoice into the same chain (with situation_counter lower). Adding a field to store is a duplication of info and will generate troubles.

htdocs/install/mysql/tables/llx_facturedet.sql
```diff
@@ -65,8 +65,14 @@ create table llx_facturedet
   
   situation_percent real DEFAULT 100, 								-- % progression of lines invoicing
   fk_prev_id        integer, 										-- id of the line in the previous situation
+  situation_percent_cumulated real DEFAULT 100,                     -- Invoice situation percent cumulated
```
eldy, line 68: Same remark than for invoice.
On invlice line we have a field    
fk_prev_id
It is the link to the previous invoice line.  total cumulated is sum of all lines.
getCumulatedAmountForPreviousSituationLine()  on invoice line

PR #24018, opened 2023-02-23
New : Test of big upload on test_forms

Hystepik:
+ Add test of big upload on public/test/test_forms.php
+ fix upload process

### DIFF
--- a/htdocs/core/ajax/flowjs-server.php
+++ b/htdocs/core/ajax/flowjs-server.php
@@ -35,9 +35,6 @@ if (!defined('NOREQUIREAJAX')) {
 if (!defined('NOREQUIRESOC')) {
 	define('NOREQUIRESOC', '1');
 }
-if (!defined('NOCSRFCHECK')) {
-	define('NOCSRFCHECK', '1');
-}
 // If there is no need to load and show top and left menu
 if (!defined("NOLOGIN")) {
 	define("NOLOGIN", '1');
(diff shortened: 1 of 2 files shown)


htdocs/public/test/test_forms.php
```diff
@@ -105,6 +105,70 @@
 $arrayselected = array(1, 3);
 print $form->multiselectarray('testmulti', $array, $arrayselected, '', 0, 'minwidth100');
 
+print '<br><br>'."\n";
+
+// Test6a: Upload of big files
+print "Test 6a: Upload of big files<br>\n";
+print "The file will be uploaded in the directory: documents/test_form/temp/<br>\n";
```
eldy, line 112: You can use  documents/test/temp/   that is already used by other components

htdocs/core/ajax/flowjs-server.php
```diff
@@ -35,6 +35,9 @@
 if (!defined('NOREQUIRESOC')) {
 	define('NOREQUIRESOC', '1');
 }
+if (!defined('NOCSRFCHECK')) {
```
eldy, line 38: If you disable the CSRF protection, it means a hacker can provide a link that will force your to execute this component. Becasue this is a write component (it create a file on disk), this is dangerous.
We should be able to have it working with NOCSRFCHECK to on.
Did you include the   token=currentToken as parameter by the caller ?

frederic34:
@Hystepik is it possible to add a check on $module, like isModuleEnabled($module) ?

Hystepik:
@frederic34  $module is only used to make an upload to a specific document part.
Exemple: $module = test -> the file will be  upload in documents/test/temp/...
But yes I can do it

PR #24019, opened 2023-02-23
fix : Warning: Undefined property: Conf:: in /home/httpd/vhosts/aflac…

grandoc:
….fr/domains/dev.aflac.fr/httpdocs/contrat/card.php on line 1652

### DIFF
--- a/htdocs/contrat/card.php
+++ b/htdocs/contrat/card.php
@@ -1649,9 +1649,9 @@ if ($action == 'create') {
 					if ($objp->subprice >= 0) {
 						$colspan = 6;
 
-						if ($conf->margin->enabled && getDolGlobalString('PRODUCT_USE_UNITS')) {
+						if (isModEnabled('margin') && getDolGlobalString('PRODUCT_USE_UNITS')) {
 							$colspan = 8;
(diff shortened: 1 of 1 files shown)


Daviid-P:
Isn't `isModEnabled(xxxxx)` just using `$conf->xxxxxx->enabled` under the hood? How does that change fix 

`Undefined property: Conf:: in /home/httpd/vhosts/aflacfr/domains/dev.aflac.fr/httpdocs/contrat/card.php on line 1652`

???

eldy:
> Isn't `isModEnabled(xxxxx)` just using `$conf->xxxxxx->enabled` under the hood? How does that change fix
> 
> `Undefined property: Conf:: in /home/httpd/vhosts/aflacfr/domains/dev.aflac.fr/httpdocs/contrat/card.php on line 1652`
> 
> ???

$conf->margin->enabled   is the old syntax

isModEnabled('margin')   is the new syntax to use.  Currently the same, but the new one does not generate warnings on PHP8.2 when $conf->margin not defined and it will allow to introduce more features in a future inside the isModEnabled function.

PR #24022, opened 2023-02-23
NEW Make it possible to select hours and minutes in form_confirm

Daviid-P:
I'm not sure where to put this.

Is this a FIX because there should have always been the possibility to select hours and minutes?

Or is it a NEW since technically it is a new addition?

It would work by adding `hours` and `minutes` to the array of inputs passed to `formconfirm()` in the `$formquestion` argument.
```php
$array_inputs = array(
  array(
    'label' => "Select a date and time:",
    'type' => "date",
    'value' => $default_value,
    'name' => 'last_day_of_work',
    'hours' => 1,
    'minutes' => 1,
  )
);
```

`hours` and `minutes` end up as the `$h` and `$m` for `select_date()`

```php
/*
*	@param	int			$h				1 or 2=Show also hours (2=hours on a new line), -1 has same effect but hour and minutes are prefilled with 23:59 if date is empty, 3 show hour always empty
*	@param	int			$m				1=Show also minutes, -1 has same effect but hour and minutes are prefilled with 23:59 if date is empty, 3 show minutes always empty
*/
```


eldy:
NEW
Take into consideration this for maintenance version:  Did the feature working before ? If no, it is a not completely implemented / not finished feature. So fix must be into develop branch.
If blocking or feature was working correctly into a previous version: Bug.

So thanks to push on develop (if feature not already available)

Daviid-P:
What would be the best way to do it?

I admit I personally first tried 'datetime' as type.

```php
//<<<<<<< add_hours_minutes_to_form_confirm
		$h = $input['hours'] ?: 0;
		$m = $input['minutes'] ?: 0;
		$more .= $this->selectDate($input['value'], $input['name'], $h, $m, 0, '', 1, $addnowlink);
//=======
		$more .= $this->selectDate($input['value'], $input['name'], ($input['type'] == 'datetime' ? 1 : 0), ($input['type'] == 'datetime' ? 1 : 0), 0, '', 1, $addnowlink);
//>>>>>>> develop
```

My solution let's you choose the values of `$h` and `$m` and doesn't contraint you to 0 or 1

eldy:
Solution 1 seems better
```
$h = $input['hours'] ?: 0;
$m = $input['minutes'] ?: 0;
$more .= $this->selectDate($input['value'], $input['name'], $h, $m, 0, '', 1, $addnowlink);
```

Daviid-P:
Fixed!

PR #24033, opened 2023-02-24
pagination works with any font

frederic34:


htdocs/core/lib/pdf.lib.php
```diff
@@ -1287,11 +1288,14 @@ function pdf_pagefoot(&$pdf, $outputlangs, $paramfreetext, $fromcompany, $marge_
 		}
 	}
 	// Show page nb only on iso languages (so default Helvetica font)
-	if (strtolower(pdf_getPDFFont($outputlangs)) == 'helvetica') {
+	// if (strtolower(pdf_getPDFFont($outputlangs)) == 'helvetica') {
```
eldy, line 1291: The comment seems to say there is pb with exotic languages.
Did you test for example, greek, chinese, japanse, arab ?

frederic34: https://github.com/Dolibarr/dolibarr/blob/42696954f3c2003eba94dca6e814759397f85077/htdocs/includes/tecnickcom/tcpdf/tcpdf.php#L13965-L13971
drawing this with $pdf->multicell doesn't give something good

actual code (with remove of check if font is helvetica) with dejavusans

image

with proposed changes

image

you need to use `$pdf->PageNo()` and `$pdf->getNumPages()` if you want to draw with `$pdf->MultiCell`
image

PR #24035, opened 2023-02-24
NEW - See the documents of my subordinates

BB2A-Anthony:
# NEW|New [*See the documents of my subordinates*]
This function is essential in my opinion


htdocs/comm/propal/list.php
```diff
@@ -629,8 +629,13 @@
 
 $sql .= ' WHERE p.fk_soc = s.rowid';
 $sql .= ' AND p.entity IN ('.getEntity('propal').')';
-if (empty($user->rights->societe->client->voir) && !$socid) { //restriction
-	$sql .= " AND s.rowid = sc.fk_soc AND sc.fk_user = ".((int) $user->id);
+if (!$user->hasRight('societe', 'client', 'voir') && !$socid) { //restriction
+	$sql .= " AND s.rowid = sc.fk_soc AND (sc.fk_user = ".((int) $user->id);
+	$userschilds = $user->getAllChildIds();
```
eldy, line 634: the getAllChildIds return the hierachy and the permission management on hierarchy is supported only for the modules HR (expense report and holiday).

The link with "sales represnetatives" is planned to have a rewrite (because it is often managed by a left join, when we need a WHERE only).  This is a postponed delay, and i suggest to wait this task is completed (and we will try to include your need) before adding the feature to manage the hierarchy into link with sale representatives...

BB2A-Anthony: @eldy not return only id tu subordonate

eldy: Link with sales representative to make filter on sale representative has been rewrite with no join in v19

eldy:
Link with sales representative to make filter on sale representative has been rewrite with no join in v19.
May be you can make now another try for this feature ?

PR #24041, opened 2023-02-24
FIX backtopage url in event card

lvessiller-opendsi:
FIX backtopage url in event card :
- HTTP referer is like an external link and contains full url with protocol sush as "http://" or "https://"
- this url is not allow in GETPOST method and the protocol with "//" are removed
- see commented line 648 in core/lib/functions.lib.php 
`// We remove schema*// to remove external URL`

**To reproduce**
- go to the event tab on product card
image
- then click on "+" button to create an event
image
- and click on cancel button of the event card in create mode
You will appear a page with error 404 for file not found
image

I removed the HTTP_REFER in backtopage url in order to avoid this case, but I don't know if it's the good way to do.
Maybe, we shouldn't use "backtopage=1" in event tabs.

### DIFF
--- a/htdocs/comm/action/card.php
+++ b/htdocs/comm/action/card.php
@@ -948,7 +948,7 @@ if ($action == 'create') {
 	print '<input type="hidden" name="donotclearsession" value="1">';
 	print '<input type="hidden" name="page_y" value="">';
 	if ($backtopage) {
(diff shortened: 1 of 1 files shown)


eldy:
I am not sure of the side effect, so i merged it manually into v17.

PR #24043, opened 2023-02-24
Fix member nature display in new.php

ymollard:


htdocs/public/members/new.php
```diff
@@ -539,6 +539,8 @@ function initmorphy()
 // Moral/Physic attribute
 $morphys["phy"] = $langs->trans("Physical");
 $morphys["mor"] = $langs->trans("Moral");
+print '<tr class="morphy"><td class="titlefield">'.$langs->trans('MemberNature').' <span style="color: red">*</span></td><td>'."\n";
```
eldy, line 542: This is duplicate with following line...

ymollard: Good catch, I missed the deletion of the second line.
That line must appear both in the if and else, otherwise the table is broken.

htdocs/public/members/new.php
```diff
@@ -539,8 +539,9 @@ function initmorphy()
...
```
eldy, line 542: Nearly. Because the ts is now open outside the if, you must also move the closing of tr in and td outside.

ymollard: Yes. The good answer was actually much simpler: there were 2 prints of the member nature when morphy was forced. With my last commit morphy is only hidden when forced :p

PR #24049, opened 2023-02-25
v17 Merge problem - Restore #22762 - Accountancy - Better warning for situation invoice in index

aspangaro:
https://github.com/Dolibarr/dolibarr/pull/22762 Patch already present in v15-v16, not in v17

Before : widget and tutorial are lost

image

After : 

image

### DIFF
--- a/htdocs/accountancy/index.php
+++ b/htdocs/accountancy/index.php
@@ -115,11 +115,6 @@ if (isModEnabled('accounting')) {
 		print "<br>";
 	}
 
-	if (!empty($conf->global->INVOICE_USE_SITUATION) && $conf->global->INVOICE_USE_SITUATION == 1) {
-		print info_admin($langs->trans("SorryThisModuleIsNotCompatibleWithTheExperimentalFeatureOfSituationInvoices"));
-		print "<br>";
-	}
-
 	print '<div class="'.($helpisexpanded ? '' : 'hideobject').'" id="idfaq">'; // hideobject is to start hidden
 	print "<br>\n";
 	print '<span class="opacitymedium">'.$langs->trans("AccountancyAreaDescIntro")."</span><br>\n";
--- a/htdocs/accountancy/index.php
+++ b/htdocs/accountancy/index.php
@@ -83,12 +83,7 @@ $help_url = 'EN:Module_Double_Entry_Accounting#Setup';
 
 llxHeader('', $langs->trans("AccountancyArea"), $help_url);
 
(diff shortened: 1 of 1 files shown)


htdocs/accountancy/index.php
```diff
@@ -83,12 +83,7 @@
 
 llxHeader('', $langs->trans("AccountancyArea"), $help_url);
 
-if (!empty($conf->global->INVOICE_USE_SITUATION) && $conf->global->INVOICE_USE_SITUATION == 1) {
-	print load_fiche_titre($langs->trans("AccountancyArea"), '', 'accountancy');
-
-	print '<span class="opacitymedium">'.$langs->trans("SorryThisModuleIsNotCompatibleWithTheExperimentalFeatureOfSituationInvoices")."</span>\n";
-	print "<br>";
-} elseif (isModEnabled('accounting')) {
+if (isModEnabled('accounting')) {
```
eldy, line 86: The goal of the protection is to not use the module accountancy when the not stable feature situation invoice is enabled. Because results are wrong.
So we should not remove this.
We can introduce a constant 
ACCOUNTANCY_ALLOWED_WITH_BUGGED_SITUATION_INVOICES
to set to 1 if we want to force the accountant when situation invoice is on, but we should not allow it.

aspangaro: It's juste a better display of the message for not lost tutorial and widget when we use the not stable feature situation invoice.

Removing the manual when you need it most is weird and blocking the module because you use a hidden feature is not a solution either.

By the way, everything is functional in accounting for status invoices apart from the statistics displayed in the binding pages.

Moreover, you have validated this patch in the old versions and now you change your mind...

eldy: I can't see all the side effect of a feature when i validate a pr, sometimes i discover the problem after. 
I don't understand how result can be good in accountancy. Amount of a situation invoice for 100 final (and 2 invoices of 50), is 100 instead in database instead of 50, so dispatching into accountancy will also dispatch 100. So result is necesserly wrong, not just on the binding page but everywhere. I do not succeed in having a correct result on my side.

We should not provide to end users a feature that can't work. If we did it in the past, it was an error (and error happens), but continuing to do it is a second error.
Saying "we must continue to let user having problems because we let them have problem before is not a good idea".
This is not "changing mind". It is "discovering an critical error was done and fixing it".


For the moment i still try to understand how data can be good in accounting, i always bad value.

PR #24051, opened 2023-02-25
Add External Ref for Cheques Deposit

adriennsm:
# Add External Ref for Cheques Deposit

image

https://www.dolibarr.fr/forum/t/reference-externe-remise-de-cheques/20469/2

### DIFF
--- a/htdocs/compta/paiement/cheque/card.php
+++ b/htdocs/compta/paiement/cheque/card.php
@@ -575,7 +575,6 @@ if ($action == 'new') {
 	print '</tr>';
 
 	// External ref
-	/* Ext ref are not visible field on standard usage
 	print '<tr><td>';
 
 	print '<table class="nobordernopadding" width="100%"><tr><td>';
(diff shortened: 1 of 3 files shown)


eldy:
The definition of "External ref" is : 
"it is in database but never used by dolibarr, never read and never modified by dolibarr".  It is here for need of external app, not for Dolibarr.
If you need a visible field, with possible edition, it means it is not the ref_ext. What would be the purpose of the field you need ?

adriennsm:
Thanks @eldy.
The need is to have a field to indicate the cheque deposit form reference, as provided by the bank.
Does this mean that a new field must be created?

eldy:
> Thanks @eldy. The need is to have a field to indicate the cheque deposit form reference, as provided by the bank. Does this mean that a new field must be created?

Yes, you must first push a PR to change db structure to add a new field. It could have been  "ref_bank", but it may be a better idea to add a field "label" and user is free to enter anything of its choice. So what do you think of just a field "label" (empty by default and free text). So it can also be used for other purpose for people that want to store something else that an ref coming from the bank.

eldy:
@adriennsm I have added the field  "label" into llx_bordereau_cheque so the road is open to implement edition of an input text label at the cheque deposit slip level.

adriennsm:
Thanks @eldy. Sorry for my late reply.
This is perfect with the new field "label".
I try to propose a PR, in a few weeks, to add this field in the cheque deposit UI.

PR #24054, opened 2023-02-26
FIX Accountancy - Journal - Ajust terms by default like only sell, purchases, etc

aspangaro:
By default with a fresh install of v17 when we create data of journal, the term must be simple. Please simplifiy terms based on translation.

And if users do not rename the journal labels in Accounting > Configuration > Journals, the default terms are too long and with brackets. This will cause problems for some software and will be problematic during the FEC control, which could lead to a refusal of the latter.

By default :  

image

Move to : (After transifex synchronisation)

image

### DIFF
--- a/htdocs/accountancy/journal/sellsjournal.php
+++ b/htdocs/accountancy/journal/sellsjournal.php
@@ -4,7 +4,7 @@
  * Copyright (C) 2011       Juanjo Menent           <jmenent@2byte.es>
  * Copyright (C) 2012       Regis Houssin           <regis.houssin@inodbox.com>
  * Copyright (C) 2013       Christophe Battarel     <christophe.battarel@altairis.fr>
- * Copyright (C) 2013-2021  Alexandre Spangaro      <aspangaro@open-dsi.fr>
+ * Copyright (C) 2013-2023  Alexandre Spangaro      <aspangaro@open-dsi.fr>
  * Copyright (C) 2013-2016  Florian Henry           <florian.henry@open-concept.pro>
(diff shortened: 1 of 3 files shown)


eldy:
We must keep the label as it is for english language, because short terms means nothing in some countries.
After feedback for bookkeepers from australia and usa, the old label were not valid for a lot of english speaking countries. So english version was modified to means something.
Also the label allows 128 char longn and parenthesis/brackets are allowed.
If having a free label of 128 char here including parenthesis generates a trouble later (during FEC export) it means we have a bug, but problem is not into the default preloade values. May be we can remove text into parenthesis/brackets during the FEC export ?
Because label is a free text, if parenthesis/bracket is a problem, removing them into the preloaded value, will not be enough: If users are able to add parenthesis, users will add them and we will face a bug for them, whatever are the preloaded value...
Why not adding this during FEC export :
`$texttoexport = preg_replace('/\(.*\)/', '', $texttoexport);`
What is the maxium lenth allowed by FEC ?

aspangaro:
There is no length limit for the label in the FEC format, the separator being the tab.

The problem is that it is not necessary to declare "journal" in a column of data for... journals. You lose characters and most French accounting software does not support 128 characters, they store this type of information in as few as 16-32 characters and do not support parentheses. Last time I had an import error because the software thought it was an injection. It truncated the journal name between one of the two parentheses (the first open, not the second).

I just want a simple name, what you added for the USA, the Australians does not mean anything in French accounting, I had to rectify everything in the French translation at the level of Transifex, I will do the same for the journal codes. 

Even journal codes can be limited to 2 characters...

PR #24066, opened 2023-02-28
doc

frederic34:


htdocs/comm/action/class/actioncommreminder.class.php
```diff
@@ -72,7 +72,6 @@ class ActionCommReminder extends CommonObject
 	 *  'showoncombobox' if field must be shown into the label of combobox
 	 */
 
-	// BEGIN MODULEBUILDER PROPERTIES
```
eldy, line 75: Can you keep the Modulebuilder tag. We may need it in future...

PR #24069, opened 2023-02-28
Ignore localtaxes if localtax type zero

### LINKED ISSUE #24068: Ignore taxes if `localtax1_type` or `localtax2_type` is 0
https://github.com/Dolibarr/dolibarr/blob/49a403b4a578d88ba163a68939f92e472bbcf631/htdocs/core/lib/functions.lib.php#L6167-L6188

In the case when you have 
image

Or maybe you want one option where localtax is added before VAT and one option where it's added after.

The select will return 2 rows, in my case `$obj->localtax2` ends up being 0 because that row is the first one returned.

Daviid-P:
# FIX|Fix #24068 Ignore localtaxes if localtax type zero
The SQL query should ignore the rows where the "Use Local Tax" select in the dictionary page was set to "No" (localtax1_type or localtax2_type is 0 on DB)


Daviid-P:
Mmmmmh...
I think I messed up?
https://github.com/Daviid-P/dolibarr/tree/ignore_localtaxes_if_localtax_type_zero
This should only be 1 commit.

PR #24070, opened 2023-02-28
Ignore localtaxes if localtax1_type or localtax2_type is 0

Daviid-P:
FIX|Fix https://github.com/Dolibarr/dolibarr/issues/24068 Ignore localtaxes if localtax type zero

The SQL query should ignore the rows where the "Use Local Tax" select in the dictionary page was set to "No" (localtax1_type or localtax2_type is 0 on DB)

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -5688,22 +5688,20 @@ function isOnlyOneLocalTax($local)
 function get_localtax_by_third($local)
 {
 	global $db, $mysoc;
-	$sql = "SELECT t.localtax1_type, t.localtax2_type, t.localtax1, t.localtax2 ";
+
+	$sql  = " SELECT t.localtax$local as localtax";
 	$sql .= " FROM ".MAIN_DB_PREFIX."c_tva as t inner join ".MAIN_DB_PREFIX."c_country as c ON c.rowid=t.fk_pays";
 	$sql .= " WHERE c.code = '".$db->escape($mysoc->country_code)."' AND t.active = 1 AND t.taux=(";
 	$sql .= "  SELECT max(tt.taux) FROM ".MAIN_DB_PREFIX."c_tva as tt inner join ".MAIN_DB_PREFIX."c_country as c ON c.rowid=tt.fk_pays";
 	$sql .= "  WHERE c.code = '".$db->escape($mysoc->country_code)."' AND tt.active = 1";
-	$sql .= "  )";
+	$sql .= "  ) ";
+	$sql .= " AND t.localtax${local}_type > 0";
+	$sql .= " ORDER BY rowid DESC";
 
 	$resql = $db->query($sql);
 	if ($resql) {
-		while ($obj = $db->fetch_object($resql)) {
-			if ($local == 1 && $obj->localtax1_type > 0) {
-				return $obj->localtax1;
-			} elseif ($local == 2 && $obj->localtax2_type > 0) {
-				return $obj->localtax2;
-			}
-		}
+		$obj = $db->fetch_object($resql);
+		return $obj->localtax;
 	}
 
 	return 0;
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -5697,11 +5697,12 @@ function get_localtax_by_third($local)
 
 	$resql = $db->query($sql);
 	if ($resql) {
-		$obj = $db->fetch_object($resql);
-		if ($local == 1) {
-			return $obj->localtax1;
-		} elseif ($local == 2) {
-			return $obj->localtax2;
+		while ($obj = $db->fetch_object($resql)) {
```
eldy, line 5700: Do we really need this while ? We should have only one line.
Which distribution of data do you have into llx_c_tva that need this look ?

Daviid-P: llx_c_tva
---
| rowid | fk_pays | CODE | taux | localtax1 | localtax1_type | localtax2 | localtax2_type | note | 
| ---: | ---: | --- | ---: | --- | --- | --- | --- | --- | 
| 44 | 4 | IVA21 | 21 | 0 | 0 | 0 | 0 | IVA21 STANDARD | 
| 2464 | 4 | IVA21+IRPF | 21 | 5.2 | 3 | -19:-15:-7 | 5 | IVA21 + IRPF VARIOS | 

Right now the select would return these 2 rows, and return at the first one so that's why I added the while loop.
I though it was easier this way, would it be better to add a condition in the `WHERE` to filter out `localtax#_type` based on `$local`

Or maybe this would be more readable?
```php
function get_localtax_by_third($local)
{
	global $db, $mysoc;

	$sql = "SELECT t.localtax$local as localtax";
	$sql .= " FROM ".MAIN_DB_PREFIX."c_tva as t inner join ".MAIN_DB_PREFIX."c_country as c ON c.rowid=t.fk_pays";
	$sql .= " WHERE c.code = '".$db->escape($mysoc->country_code)."' AND t.active = 1 AND t.taux=(";
	$sql .= "  SELECT max(tt.taux) FROM ".MAIN_DB_PREFIX."c_tva as tt inner join ".MAIN_DB_PREFIX."c_country as c ON c.rowid=tt.fk_pays";
	$sql .= "  WHERE c.code = '".$db->escape($mysoc->country_code)."' AND tt.active = 1";
	$sql .= "  ) ";
	$sql .= "  AND t.localtax${local}_type > 0";

	$resql = $db->query($sql);
	if ($resql) {
		return $obj->localtax;
	}

	return 0;
}

Since $local is alway 1 or 2 and we can only get 1 value we can use the variable to build the correct SQL query.
```

Daviid-P: You mean like this?

```php
function get_localtax_by_third($local)
{
	global $db, $mysoc;

	$sql = "SELECT t.localtax$local as localtax";
	$sql .= " FROM ".MAIN_DB_PREFIX."c_tva as t inner join ".MAIN_DB_PREFIX."c_country as c ON c.rowid=t.fk_pays";
	$sql .= " WHERE c.code = '".$db->escape($mysoc->country_code)."' AND t.active = 1 AND t.taux=(";
	$sql .= "  SELECT max(tt.taux) FROM ".MAIN_DB_PREFIX."c_tva as tt inner join ".MAIN_DB_PREFIX."c_country as c ON c.rowid=tt.fk_pays";
	$sql .= "  WHERE c.code = '".$db->escape($mysoc->country_code)."' AND tt.active = 1";
	$sql .= "  ) ";
	$sql .= "  AND t.localtax${local}_type > 0";
	$sql .= "  ORDER BY t.localtax${local} DESC";

	$resql = $db->query($sql);
	if ($resql) {
		return $obj->localtax;
	}

	return 0;
}
```

I'm not sure in other countries' cases, but in Spain since localtax2 is negative 0 would still be the first returned row.

Did you mean order desc by localtax_type?

eldy: Yes i was thinking of something like that. The order by so result is predictive.
And because in spain, value may be negative, you finaly must keep the
`if ($local == 1 && $obj->localtax1_type > 0) {`

PR #24074, opened 2023-02-28
FIX 16.0 total line for extrafields on list if totalizable

atm-GregM:
FIX 16.0 total line for extrafields on list if totalizable


atm-GregM:
@eldy

atm-GregM:
Hello @eldy, can you check this Pull request please ?

PR #24077, opened 2023-02-28
clean code

frederic34:
add doc
remove unused code


htdocs/workstation/class/workstation.class.php
```diff
@@ -94,7 +93,6 @@ class Workstation extends CommonObject
 	 *  Note: To have value dynamic, you can set value to 0 in definition and edit the value on the fly into the constructor.
 	 */
 
-	// BEGIN MODULEBUILDER PROPERTIES
```
eldy, line 97: We must keep tags   // ... MODULEBUILDER ...
because in a future we will be able to use modulebuilder to edit also modules of the core. And this tags are necessary for modulebuilder to work.

frederic34: @eldy tags restored

Issue #24001, opened 2023-02-22
`FileUpload` class can't be used without an object


Daviid-P:
### Bug

The class constructor has all arguments set to null by default
https://github.com/Dolibarr/dolibarr/blob/78b0ce6b24c611ae35f4b021abde30238679253e/htdocs/core/class/fileupload.class.php#L44

However actually using it like that leads to
```
$dir_output = null;
$pathname = null;
$filename = null;
$classname = null;
```

all of which have to be not null and related to valid dolibarr classes for the code to work.
https://github.com/Dolibarr/dolibarr/blob/78b0ce6b24c611ae35f4b021abde30238679253e/htdocs/core/class/fileupload.class.php#L107-L117

Is it possible to upload files without needing to use or assign a class to that file?

### Environment Version

16.0.3

Hystepik:
It seems to be fixed so I close the issue.
Feel free to reopen if not.

PR #24079, opened 2023-03-01
Allow FileUpload to be used when $fk_element and $element are null

Daviid-P:
# NEW #24001 Allow to upload files not linked to a class
This is a personal problem I had, I wanted to use FileUpload.class.php to upload a file in a custom module but since $fk_element and $element were null it failed when trying to instantiate a class with an empty $classname variable.

Since we don't create the upload_dir from a class then we must check either that the supplied options contains a valid upload_dir or that after the `overrideUploadOptions` hook `$this->options['upload_dir']` contains a valid upload_dir, so I added checks at the end of the constructor, lmk if that's not necessary.


eldy:
Note:  this file is not used by Dolibarr so will probably be removed soon...

Daviid-P:
> Note: this file is not used by Dolibarr so will probably be removed soon...

Really? What would be the correct way to upload a file then?

Issue #23995, opened 2023-02-22
email send to closed contact


MaximilienR-easya:
### Feature Request

to not send the email and to signal the user that there is a closed contact

### Use case

We have cases with customers where contacts linked to tickets, propals, etc, end up being closed along the way and therefore remains in the contacts despite the fact that they have been closed and end up receiving emails when they should no longer receive them.

### Suggested implementation

don't send the email to the closed contact and see a warning or an error signaling the user who within the contacts are closed and that their email was not send.

I'm working on it but i would like your opinion and if it is subject to an option.

PR #24081, opened 2023-03-01
NEW suggested contacts when sending email ticket does not include closed contacts anymore by defaut

MaximilienR-easya:
# New [*feature request #23995 with PR #24076*]
This PR and the PR #24076 are here To respond to the feature request #23995.
We have added a function to filter the closed contact from the list of contact before sending the email and added a warning to the user to inform him if a client is closed, who it is and that he was not send the email.


htdocs/ticket/class/ticket.class.php
```diff
@@ -2812,6 +2812,7 @@ public function sendTicketMessageByEmail($subject, $message, $send_internal_cc =
 
 		$from = $conf->global->TICKET_NOTIFICATION_EMAIL_FROM;
 		$is_sent = false;
+		$array_receiver = $this->removeClosedContact($array_receiver);
```
eldy, line 2815: I don't think we must remove after having loaded :
We should modify the  getInfosTicketInternalContact and add a parameter  $onlyopen = 1 for example.
Then we call line 2805    getInfosTicketInternalContact(1)   
Warning is also useless because warning should be used when 
- something is wrong and it is the common life to have contact disabled. 
- an action is expected to fix
Here we should just not include them. No need to alert user to make an action.

PR #24087, opened 2023-03-02
better situation invoice title

rycks:
#FIX missing situation invoice step number in default sponge pdf

Please have a look on dolibarr french forum for more details: https://www.dolibarr.fr/forum/t/affichage-du-numero-de-situation-dans-les-factures-pdf/42372/

before:

image

after:

image


eldy:
To save spaces, in v17 all title of document were moved from
```
Title
Ref: XXX

```
into 
`Title XXX`

But your change go into the oposite direction. May be you can make:
Facture de situation n°4 - 00006206
To restore the goal to have title and ref on 1 line ?

Also can you push on develop as this does not fix a critical bug.

Last point, because this is an invoice, the term "Invoice" must appear.

PR #24094, opened 2023-03-02
FIX: custom accountancy categories: postgresql data type errors

marc-dll:


htdocs/accountancy/class/accountancycategory.class.php
```diff
@@ -743,9 +743,9 @@ public function getSumDebitCredit($cpt, $date_start, $date_end, $sens, $thirdpar
 				}
 				$listofaccount .= "'".$cptcursor."'";
 			}
-			$sql .= " AND t.numero_compte IN (".$this->db->sanitize($listofaccount).")";
```
eldy, line 746: Why not just fixing with
`$sql .= " AND t.numero_compte IN (".$this->db->sanitize($listofaccount, 1).")";`
Is it enough for you ?

marc-dll: @eldy Yes, it works and it's simpler. Thank you.

PR #24103, opened 2023-03-03
Autofocus field working on cards

atm-greg:
# FIX autofocus on cards
Make autofocus working on cards using configuration of focused fields

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -8617,7 +8617,7 @@ function printCommonFooter($zone = 'private')
 			}
 
 			// Management of focus and mandatory for fields
-			if ($action == 'create' || $action == 'edit' || (empty($action) && (preg_match('/new\.php/', $_SERVER["PHP_SELF"])))) {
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -9591,7 +9591,7 @@ function printCommonFooter($zone = 'private')
 			}
 
 			// Management of focus and mandatory for fields
-			if ($action == 'create' || $action == 'edit' || (empty($action) && (preg_match('/new\.php/', $_SERVER["PHP_SELF"])))) {
+			if ($action == 'create' || $action == 'edit' || (empty($action) && (preg_match('/new\.php/', $_SERVER["PHP_SELF"]))) || ((empty($action) || $action == 'addline') && (preg_match('/card\.php/', $_SERVER["PHP_SELF"])))) {
```
eldy, line 9594: The focus on card should be already managed by the  $action == 'create' || $action == 'edit' 
Adding a case   empty($action)   may result in focus forced in a lot of case we don't need.

atm-greg: This is for lines form. We need this to focus on search_idprod.
When you are on a draft propal, the action is empty and after adding a line, the action is 'addline'.
Have you some suggestion to make it another way

eldy: May be we can add 'addline'  ?
`if ($action == 'create' || $action == 'edit' || $action == 'addline') {`

eldy: No better idea.
Lets do like that and we will see side effects.

PR #24105, opened 2023-03-03
fix: remove invoice line id and invoice on time spend when delete line

FHenry:
fix: remove invoice line id and invoice on time spend when delete line

And refactor SQL into line->delete better than in invoice->deleteline.

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -4230,19 +4230,6 @@ class Facture extends CommonInvoice
 
 		$this->db->begin();
 
-		// Free discount linked to invoice line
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/class/facture.class.php
```diff
@@ -4230,19 +4230,6 @@ public function deleteline($rowid, $id = 0)
 
 		$this->db->begin();
 
-		// Free discount linked to invoice line
```
eldy, line 4233: I don't understand why you remove this ?

FHenry: Because I move it to delete method

FHenry:
@eldy  I've move the delete sql into Line class delete and add the task_time fix in the method too, it's harmfull because the Line method is called by the head class.

Issue #24102, opened 2023-03-03
Manual addition of invoice lines on time spent in tasks


lmag:
### Feature Request

Allow to add billing in past time

### Use case

Time spent invoices can be made without the relation to time in task.
It must be possible to assign them afterwards in order to allow a future calculation of ROI on the time invoiced.

### Suggested implementation

![image](https://user-images.githubusercontent.com/3288514/222674692-6c9ca4b4-7133-42ce-83f4-d094b55263fd.png)

![image](https://user-images.githubusercontent.com/3288514/222675993-92177bb1-0f55-40d0-a64d-40b96cd288c6.png)

### Suggested steps

- [ ] Add select invoice
- [ ] Add ligne invoice
- [ ] Add informations of facture line (Quantity, Amount)

lmag:
Whaou !
![image](https://user-images.githubusercontent.com/3288514/222705865-e22a6597-be00-4a3b-8704-2f497423c850.png)
#24110

lmag:
Whaou !  - 2
![image](https://user-images.githubusercontent.com/3288514/222707211-ba00fe58-01f2-4911-aa2e-5361f1e33ec6.png)

PR #24110, opened 2023-03-03
 feat: update invoice and invoice line on time spend #24102 

FHenry:
feat: update invoice and invoice line on time spend #24102

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -10027,9 +10027,10 @@ class Form
 	 *  @param	string	$projectsListId ''=Automatic filter on project allowed. List of id=Filter on project ids.
 	 *  @param	string	$showproject	'all' = Show project info, ''=Hide project info
 	 *  @param	User	$usertofilter	User object to use for filtering
+	 *  @param	int		$status			Invoice status filter (-1 for no filter)
 	 *	@return int         			Nbr of project if OK, <0 if KO
 	 */
-	public function selectInvoice($socid = -1, $selected = '', $htmlname = 'invoiceid', $maxlength = 24, $option_only = 0, $show_empty = '1', $discard_closed = 0, $forcefocus = 0, $disabled = 0, $morecss = 'maxwidth500', $projectsListId = '', $showproject = 'all', $usertofilter = null)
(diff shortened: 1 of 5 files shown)


htdocs/projet/tasks/time.php
```diff
@@ -309,6 +309,7 @@
 			}
 			$object->timespent_fk_user = GETPOST("userid_line", 'int');
 			$object->timespent_fk_product = GETPOST("fk_product", 'int');
+			$object->timespent_fk_product = GETPOST("fk_product", 'int');
```
eldy, line 312: Duplicate line.
May be instead ?
```
$object->timespent_invoiceid = GETPOST("invoiceid", 'int');
$object->timespent_invoicelineid = GETPOST("invoicelineid", 'int');
```

htdocs/core/class/html.form.class.php
```diff
@@ -10154,13 +10201,11 @@ public function selectInvoice($socid = -1, $selected = '', $htmlname = 'invoicei
...
-			print $out;
+			return $out;
 
 			$this->db->free($resql);
-			return $num;
 		} else {
 			dol_print_error($this->db);
-			return -1;
 		}
 	}
```
eldy, line 10210: Must add a return ''; 
so function will always return a string and change the doc of function from @return int into @return string

htdocs/core/class/html.formprojet.class.php
```diff
@@ -744,8 +747,132 @@ public function selectOpportunityStatus($htmlname, $preselected = '-1', $showemp
...
+	 *  Output a combo list with invoices and lines qualified for a project
+	 *
+	 * @param int $selectedInvoiceId Id invoice preselected
+	 * @param int $selectedLineId Id invoice line preselected
+	 * @param string $htmlNameInvoice Name of HTML select for Invoice
+	 * @param int $htmlNameInvoiceLine Name of HTML select for Invoice Line
+	 * @param string $morecss More css added to the select component
+	 * @param array $filters Array of filters
+	 * @param int $lineOnly return only option for line
+	 * @return int                    Nbr of project if OK, <0 if KO
```
eldy, line 765: @return string

PR #24114, opened 2023-03-03
New : expand/collapse groups permission module

Hystepik:
Same feature as Pull Request [22801](https://github.com/Dolibarr/dolibarr/pull/22801) but for the group permission menu

### DIFF
--- a/htdocs/user/group/perms.php
+++ b/htdocs/user/group/perms.php
@@ -284,7 +284,7 @@ if ($object->id > 0) {
 		$i = 0;$j = 0;
 		$oldmod = '';
 
-		$cookietohidegroup = (empty($_COOKIE["DOLUSER_PERMS_HIDE_GRP_GRP"]) ? '' : preg_replace('/^,/', '', $_COOKIE["DOLUSER_PERMS_HIDE_GRP_GRP"]));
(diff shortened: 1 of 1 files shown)


eldy:
👍🏻

htdocs/user/group/perms.php
```diff
@@ -275,9 +281,12 @@
 	$result = $db->query($sql);
 	if ($result) {
 		$num = $db->num_rows($result);
-		$i = 0;
+		$i = 0;$j = 0;
 		$oldmod = '';
 
+		$cookietohidegroup = (empty($_COOKIE["DOLUSER_PERMS_HIDE_GRP_GRP"]) ? '' : preg_replace('/^,/', '', $_COOKIE["DOLUSER_PERMS_HIDE_GRP_GRP"]));
```
eldy, line 287: For individual permission we use DOLUSER_PERMS_HIDE_GRP
Bu i think we can use the same cookie.  If a user want to hide some module and show other to see permission, the choice will be the same if he is on individual permission or group permission.
Can't we use   DOLUSER_PERMS_HIDE_GRP  as well ?

Hystepik: Yes it makes sense

Issue #21548, opened 2022-07-21
TimeSpent : object standardization


atm-maxime:
### Feature Request

The goal is to have a real technical object to manage time spent, so in the future we can use it on tickets or mo in addition of tasks.

### Use case

I'd like to enter times I spent regarding a ticket or a manufacturing order

### Suggested implementation

Creation of a TimeSpent object and use of it in the project/task module/class

### Suggested steps

- [ ] Rename llx_projet_task_time to llx_element_time
- [ ] Rename column fk_task to fk_element and add column element_type
- [ ] Fill the element_type column with the value 'task' for existing data
- [ ] Replace current usage of time spent by the new object (card, class, export, ...)

atm-maxime:
@eldy could you tell us if such a standardization is ok for you so we get started on it ?

JoAnMaFe:
the idea of a technical object to track, has long been demanded in Dolibarr, in fact some of us do this process by calling our project with the serial number of the technical object to be monitored and tracking from the tasks or tickets associated with it.
I think it would not be enough to change the name of some tables and fields and could disrupt some ideas settled in the projects.
It is possible that I do not understand your approach, but in my point of view, this new function should be somehow associated to the new equipment module (Assets).

atm-maxime:
Our goal is to standardize the current timespent management, which only works with project tasks. After this standardization, we will be able to use timespent on tickets or manufacturing orders for example.

I don't see how timespent can be related to "the new equipment module (Assets)." which to my knowledge is an asset management regarding accountancy.

algonomp:
Would it be possible for you guys to generalize it even more? In my case i would use it to record how much time my employees spent cleaning someones house.

My case: Customers sign a cleaning contract, it generates monthly invoices with a fixed cost amount and the cleaners get payed via the hours they spent cleaning. I don't use the Projekt/Leads module for this, just the contract and invoicing modul. It would be overkill to use the projekt modul and create a projekt for every customer.

atm-maxime:
> Would it be possible for you guys to generalize it even more? In my case i would use it to record how much time my employees spent cleaning someones house.
> 
> My case: Customers sign a cleaning contract, it generates monthly invoices with a fixed cost amount and the cleaners get payed via the hours they spent cleaning. I don't use the Projekt/Leads module for this, just the contract and invoicing modul. It would be overkill to use the projekt modul and create a projekt for every customer.

[...]

PR #24116, opened 2023-03-03
NEW : table llx_element_time to store time spent on several elements (mo, ticket...) SQL integration

atm-gauthier:


atm-gauthier:
Hi @eldy, this PR deals with issue https://github.com/Dolibarr/dolibarr/issues/21548
This is the table we want to create, I await your opinion before starting work on the code and migration part of the current data

eldy:
Why creating a second table that represent the same information than into llx_projet_task_time
It is better to extend existing llx_projet_task_time by adding the 

fk_element integer NOT NULL,
elementtype varchar(32) NOT NULL

with a default value to 'task'.
The name of table will not match real need llx_projet_task_time instead of llx_element_time, but it is better to have a bad name on a table than having 2 duplicates tables for the same purpose.

atm-gauthier:
@eldy What i want to do is move data from llx_projet_task_time to llx_element_time and not keeping 2 tables.
Then i could add the 2 fields and rename the table, What do you think about that ?
(If you are OK with that please just say it to me and don't merge it now, i Will work on that feature and add all necessary code next weeks)

atm-gauthier:
@eldy i did complete sql update.
Can you check it and say me if you're ok with that ?
If it's ok don't merge now, just say it to me and i will work on php code

atm-gauthier:
Hi @eldy, this is the code integration PR : https://github.com/Dolibarr/dolibarr/pull/24158

atm-gauthier:
travis error doesn't seem to deal with what i did :
image

atm-gauthier:
I moved everything into https://github.com/Dolibarr/dolibarr/pull/24158

PR #24119, opened 2023-03-03
FIX impossible to edit user if you are admin

comaiteseb:
FIX

Line 132 surcharged the line 116 even if you ard admin ou user->user->write. So I was impossible to update user card if it was not yours...


eldy:
Closed. Already pushed into v17.

PR #24123, opened 2023-03-04
Remove unused code

BenjaminFlr:
Some code was in an `if (0)` condition, I just removed that unused part to clean the code.


eldy:
We need to keep it. The if (0) was added because the action behind is not yet implemented but it is code waiting the action behind is implemented...

PR #24124, opened 2023-03-05
php v8 warning

defrance:
On install, after first user create; we have a warning error from user class line 2042 : dol_verifyHash( $chain = 'admin123', $hash = NULL, $type = ??? )


htdocs/core/lib/security.lib.php
```diff
@@ -260,7 +261,9 @@ function dol_verifyHash($chain, $hash, $type = '0')
 	global $conf;
 
 	if ($type == '0' && !empty($conf->global->MAIN_SECURITY_HASH_ALGO) && $conf->global->MAIN_SECURITY_HASH_ALGO == 'password_hash' && function_exists('password_verify')) {
-		if ($hash[0] == '$') {
+		if (empty($hash) {
```
frederic34, line 264: missing closing parenthesis

defrance (commented):
oupsi, correct change

eldy:
Fixed by #24157

PR #24132, opened 2023-03-06
FIX 14.0 dissociate invalid & valid email

atm-GregM:
FIX dissociate invalid & valid email

-Check if sento is valid
- return at the end a setEventmessages with email was send & another message for email invalid


eldy:
This change too much the behaviour of application, must be pushed into develop.

PR #24135, opened 2023-03-06
Fix default value access

atm-jpb:
# Instructions

# FIX|Fix #Fix default value access
In the selects the test getpostisset does not give the possibility to the defaultValue to take control.( sub process in getpost function)

in function.lib.php , the GETPOST function tests the presence of the conf: MAIN_ENABLE_DEFAULT_VALUES

this part of the code will check if there is a parameter overload and replace the value with that of the user.
here for example, in propal/card.php
if we use in our tests,
 GETPOSTISSET && GETPOST in the same if  or ternary test, we cannot override an empty parameter. this pr fixes the problem.
this type of test should be corrected in all existing files.


atm-jpb:
@eldi stickler seem to be in two states  ...

frederic34:
already done in develop https://github.com/Dolibarr/dolibarr/commit/ae83802f387298eab2ec65cdeaa794e0a287ce90 can you look at this to avoid merge conflicts

atm-jpb:
oh yes , but  seem to be a fix bug for me ;)  not a new behavior. 
And it's not the same think. in fact, you don't fix the loading defaultValue overwriting process existing in GETPOST method.
your GETPOSTISSET avoid it completly. 
ref:  functions.lib.php   :  if (!empty($conf->global->MAIN_ENABLE_DEFAULT_VALUES)) in getpost function

frederic34:
_you don't fix the loading defaultValue_ => you can test in version develop, it works, because I keep the getpostisset

I have tested your fix, if thirdparty have default values set in thirdparty card, they are overwrited by default values config

frederic34:
`$paymentTermId = (GETPOSTISSET('cond_reglement_id') && GETPOST('cond_reglement_id', 'int') != 0) ? GETPOST('cond_reglement_id', 'int') : (!empty($soc->cond_reglement_id) ? $soc->cond_reglement_id : GETPOST('cond_reglement_id', 'int'));`

the last getpost take care of default values config

atm-jpb:
yes you right  didn't see the last sentence ;) .

PR #24136, opened 2023-03-06
NEW - Add origin info when create a product batch when created from a movement stock

kkhelifa-opendsi:
Add origin info when create a product batch when created from a movement stock


thomas-Ngr:
I explain the comment : When we create a product batch or a manufacturing order, we want to know if it was created during a stock movement.

This PR adds this feature

PR #24141, opened 2023-03-06
NEW: include limit in bookmark url params

altairis-noe:
# NEW: include limit in bookmark url params
When we choose to “Add current page to bookmarks", we should keep the limit parameter from a list, or the bookmark will not be the *same* page.

### DIFF
--- a/htdocs/bookmarks/bookmarks.lib.php
+++ b/htdocs/bookmarks/bookmarks.lib.php
@@ -63,7 +63,7 @@ function printDropdownBookmarksList()
 	}
 	if (!empty($_POST) && is_array($_POST)) {
 		foreach ($_POST as $key => $val) {
-			if ((preg_match('/^search_/', $key) || in_array($key, $authorized_var))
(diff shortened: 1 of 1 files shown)


htdocs/bookmarks/bookmarks.lib.php
```diff
@@ -63,7 +63,7 @@ function printDropdownBookmarksList()
 	}
 	if (!empty($_POST) && is_array($_POST)) {
 		foreach ($_POST as $key => $val) {
-			if ((preg_match('/^search_/', $key) || in_array($key, $authorized_var))
+			if ((preg_match('/^search_|^limit$/', $key) || in_array($key, $authorized_var))
```
eldy, line 66: The string 'limit' is already inside the array 
 $authorized_var
so the test should already be true, with no need of this addition. I miss something....

altairis-noe:
I didn’t see that, indeed. I had to add this in a previous version of Dolibarr (v14) and didn’t see that evolution. Sorry for the noise…

PR #24134, opened 2023-03-06
remove duplicate

frederic34:
added here https://github.com/Dolibarr/dolibarr/commit/e12d10158c8bee720b190e8b7eeda39732d3a8da

image

image

Issue #24127, opened 2023-03-05
Multi select code_client


wdammak:
https://github.com/Dolibarr/dolibarr/blob/3c96c42f32fbd93813cdd8c1da5cfe41f8d41b60/htdocs/compta/facture/list.php#L579

s.code_client as code_client,...........,s.code_client

PR #24142, opened 2023-03-06
fix #23943 : duplicate code_client

rycks:
#Fix 23943

There was a duplicate code_client in my prev contrib !!!! Thanks @frederic34 !


frederic34:
https://github.com/Dolibarr/dolibarr/issues/24127
https://github.com/Dolibarr/dolibarr/pull/24134

PR #24151, opened 2023-03-07
FIX : Set the fields authorid to see my own events

atm-adrien:
## FIX : Set the fields authorid to see my own events

Set the fields authorid to see my own events when the permission "$user->rights->agenda->allactions->read" isn't set for me


hregis:
@atm-adrien it's already fixed
https://github.com/Dolibarr/dolibarr/pull/24075

PR #24157, opened 2023-03-08
FIX: installation superadmin creation: PHP 8 warning

marc-dll:


frederic34:
related to https://github.com/Dolibarr/dolibarr/pull/24124

eldy:
Fixed by #24157 24157

PR #24162, opened 2023-03-08
FIX dol_textishtml() function

atm-lena:
# FIX|Fix #

For example, if $msg parameter of dol_textishtml() function is "<i class="fa fa-refresh" ></i>Refresh" : function return false.

Th function checks that $msg has the <i> tag but ignores the case with, for example, a class is added before the closing tag.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -7483,7 +7483,7 @@ function dol_textishtml($msg, $option = 0)
 			return true;
 		} elseif (preg_match('/<\/textarea/i', $msg)) {
 			return true;
-		} elseif (preg_match('/<(b|em|i|u)(\s*[^>]+)?>/i', $msg)) {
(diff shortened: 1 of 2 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -7483,7 +7483,7 @@ function dol_textishtml($msg, $option = 0)
 			return true;
 		} elseif (preg_match('/<\/textarea/i', $msg)) {
 			return true;
-		} elseif (preg_match('/<(b|em|i|u)>/i', $msg)) {
+		} elseif (preg_match('/<(b|em|i|u)[^>]*>/i', $msg)) {
```
eldy, line 7486: May be we can protect better with
preg_match('/<(b|em|i|u)(\s*[^>]+)?>/i', $msg)
so nothing after b em i u 
or   a space and something no > until the >
Can you check this ?
Also, you should be able to complete the phpunit tests cases with an example that was not correctly detected and he is after.

atm-lena: You're right !
Is this test ok ?

PR #24164, opened 2023-03-08
use a fonction to get multidir output

frederic34:


htdocs/commande/class/commande.class.php
```diff
@@ -3960,6 +3960,7 @@ public function initAsSpecimen()
 		$this->id = 0;
 		$this->ref = 'SPECIMEN';
 		$this->specimen = 1;
+		$this->entity = $conf->entity;
```
frederic34, line 3963: this is to add entity to specimen

htdocs/core/lib/functions.lib.php
```diff
@@ -71,6 +71,27 @@ function utf8_decode($elements)
...
+ * @param CommonObject $object Dolibarr common object
+ * @param string $module override object element, by example mycompany instead societe
+ * @since Dolibarr V18
+ * @return string|void
+ */
+function getMultidirOutput($object, $module = '')
+{
+	global $conf;
+	if (!is_object($object)) {
+		return;
```
eldy, line 86: The function is used by doing 
$var = function()
So we should not have void as return. May be we can return null or '' ?

hregis:
@frederic34 @eldy maybe to have the possibility to use $conf->entity instead of return null ?

return $conf->$module->multidir_output[$conf->entity];

PR #24174, opened 2023-03-09
Some fields of a supplier order are always editable after a validation

### LINKED ISSUE #24173: Some fields of a supplier order are always editable after a validation
### Bug

Some fields of a supplier order are always editable after a validation.

### Environment Version

16.0.3

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

These fields should only be editable in the draft state.

Create a supplier order and validate it.

### Attached files

image

FlorentPoinsaut:
# FIX #24173
Some fields of a supplier order are always editable after a validation


eldy (commented):
Some field for example the payment condition may evolve with time, even once purchase order is no more a draft and just validated.
Some may even be modified after being sent to supplier, like the date of delivery, when rhe supplier call to say the delivery will be delayed...

eldy:
So test should be done on permission only and not on invoice status.

PR #24182, opened 2023-03-10
Fix: stripe ipn.php for payment_intent.succeeded

Hystepik:
Fix webhook system of stripe to update dolibarr db after event payment_intent.succeeded send to public/stripe/ipn.php


htdocs/public/stripe/ipn.php
```diff
@@ -315,157 +315,178 @@
...
+	if ($result) {
+		$obj = $db->fetch_object($result);
+		if ($obj) {
+			$invoice_id = $obj->fk_facture;
+			$prelevement_bons_id = $obj->fk_prelevement_bons;
+			$payment_amount = $obj->amount;
+			$paymentTypeId = $obj->type;
+		}
+	} else {
+		$error = $this->db->lasterror();
```
eldy, line 348: Fatal error here.

htdocs/public/stripe/ipn.php
```diff
@@ -315,157 +315,178 @@
 	$db->commit();
 } elseif ($event->type == 'payment_intent.succeeded') {		// Called when making payment with PaymentIntent method ($conf->global->STRIPE_USE_NEW_CHECKOUT is on).
 	dol_syslog("object = ".var_export($event->data, true));
-
-	/* TODO LMR We must retreive the invoice and payment amount from the id = ext_payment_id into llx_prelevement_demande
-	include_once DOL_DOCUMENT_ROOT . '/compta/paiement/class/paiement.class.php';
+	include_once DOL_DOCUMENT_ROOT . '/compta/paiement/class/paiement.class.php'; //TOTEST
+	global $stripearrayofkeysbyenv;
+	$errors = 0;
```
eldy, line 320: If it is a counter or code (so not an array), use variable $error   ($errors is used to store an array of error message)

eldy:
What happen if we imagine that stripe send the message 2 times. Do we create the payment two times ? Can you identify into the code where is the protection to avoid problem when this occurs...

htdocs/public/stripe/ipn.php
```diff
@@ -315,157 +315,203 @@
...
-				$obj = $db->fetch_object($result);
-				$idbon = $obj->idbon;
+			dol_syslog("The payment has been created for invoice id " . $invoice_id);
+		}
+		if (!$error && isModEnabled('banque')) {
+			$ispaymentdone = 0;
+			$sql = "SELECT p.id, p.fk_bank FROM llx_paiement as p";
+			$sql .= " WHERE p.ext_payment_id = '".$db->escape($paiement->ext_payment_id)."'";
+			$sql .= " AND p.ext_payment_site = '".$db->escape($paiement->ext_payment_site)."'";
+			$sql .= " AND p.fk_bank <> '0'";
```
eldy, line 420: int compare with 0 will result in error.

PR #24184, opened 2023-03-10
New events about mail show the mail information (receivers, sender, Cc, etc)

MaximilienR-easya:
# New
Adds the ability to see email information in the event page: 
- sender
- receiver
- Cc (only shown if there is any Cc)
- subject

### DIFF
--- a/htdocs/comm/action/card.php
+++ b/htdocs/comm/action/card.php
@@ -2404,7 +2404,7 @@ if ($id > 0) {
 		}
 
 		//mail information
-		if (!empty($object->email_msgid) && $conf->global->COMM_ACTION_MAIL_DESCRIPTION) {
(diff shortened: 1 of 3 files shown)


htdocs/comm/action/card.php
```diff
@@ -2403,6 +2403,20 @@ function setdatefields()
 			print '</td></tr>';
 		}
 
+		//mail information
+		if (!empty($object->email_msgid) && $conf->global->COMM_ACTION_MAIL_DESCRIPTION) {
```
eldy, line 2407: All this information are short. So we can show them all the time when email_msgid is set.
No need for option inflation COMM_ACTION_MAIL_DESCRIPTION. You can just add the lines after the 
`if (!empty($object->email_msgid)`

PR #24200, opened 2023-03-13
fix : Fatal error: Uncaught TypeError: strlen(): Argument #1 () must …

grandoc:
…be of type string, array given in /home/httpd/vhosts/aflac.fr/domains/dev.aflac.fr/httpdocs/core/class/extrafields.class.php on line 728

### DIFF
--- a/htdocs/core/class/extrafields.class.php
+++ b/htdocs/core/class/extrafields.class.php
@@ -725,6 +725,8 @@ class ExtraFields
 
 			if (is_array($param) && count($param) > 0) {
 				$params = serialize($param);
+			} elseif (is_array($param)) {
+				$params = '';
 			} elseif (strlen($param) > 0) {
 				$params = trim($param);
 			} else {


Daviid-P:
Without knowing how $params is later retrieved from DB and used I think the correct fix would be 
`if (is_array($param)) {`
instead of 
`if (is_array($param) && count($param) > 0) `

This serializes `serialize([])` and stores `a:0:{}` in DB

If the array wasn't empty it would be serialized and I assume at some point, retireved and unserialized, so there's no point of storing an empty string instead of a serialized empty array, not like it takes much space.

```php
if (is_array($param)) {
    $params = serialize($param);
} elseif (strlen($param) > 0) {
    $params = trim($param);
} else {
    $params = '';
}
```

Now, since if we fail the first condition we don't have an array and if we fail the second condition we have an empty string, there's no need for the else since we already have a ''

UNLESS there's some extreme cases in which for some reason $param could be something else like an integer, an object, a boolean, etc...

So we can remove the else
```php
if (is_array($param)) {
    $params = serialize($param);
} elseif (strlen($param) > 0) {
    $params = trim($param);
}
```

grandoc:
Your suggestion seems correct, as long as the assumptions you've made are true. The goal appears to be to store an empty value in the database as a serialized empty array instead of an empty string. To achieve this, you suggest modifying the first condition to check whether $param is an array without checking whether it contains elements, and then removing the final else statement because if the first two conditions aren't met, $params should be an empty string.
This approach seems reasonable because it allows for consistent storage of empty values with the rest of the data in the database. However, it's important to note that this approach will only work if the data stored in the database is also retrieved and processed consistently with this approach.
Additionally, you note that it's possible that $param could contain a data type other than an array or a string. In this case, the function could generate an error. Therefore, it's important to check that $param is of an acceptable type before applying either of the conditions.
Overall, your solution seems like a good way to correct the initial error, after tested to ensure it works correctly with the rest of the code.

Daviid-P:
image

Bruh

image

lol, lmao even

grandoc:
It can help to answer quickly :))

Daviid-P:
> It can help to answer quickly :))

I'd rather have a 👍or a simple "I agree", if anyone gets angry because of short answers they shouldn't try to help. 🤷‍♂️
Personally I don't care, if you agree just say that, if you don't just say you don't (with an explanation if possible, maybe you just don't like the alternative, that's fine too)

grandoc:
It wasn't an angry response, it was a lasy response 
for me my correct was just working
Regards

Daviid-P:
> It wasn't an angry response, it was a lasy response for me my correct was just working Regards

Yeah, I didn't take it as an angry response.

PR #24204, opened 2023-03-14
FIX: Correct Evaluation for extrafields

noec764:
In V17 the comment of last dol_eval() attribute specify that '0' is used for computed property of extrafields. 
So I understand that it should be used here. 
For example if an extrafields list property is 
``` 
(! empty($object->array_options["options_other_extrafield"]) ? 5 : 2)
```
If the third attribute was not set to '0', an error was returned by dol_eval() because '['  and ']'  was not accepted char

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -7743,7 +7743,7 @@ abstract class CommonObject
 					// Test on 'enabled' ('enabled' is different than 'list' = 'visibility')
 					$enabled = 1;
 					if ($enabled && isset($extrafields->attributes[$this->table_element]['enabled'][$key])) {
-						$enabled = dol_eval($extrafields->attributes[$this->table_element]['enabled'][$key], 1, 1, '1');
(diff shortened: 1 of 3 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -7743,20 +7743,20 @@ public function showOptionals($extrafields, $mode = 'view', $params = null, $key
 					// Test on 'enabled' ('enabled' is different than 'list' = 'visibility')
 					$enabled = 1;
 					if ($enabled && isset($extrafields->attributes[$this->table_element]['enabled'][$key])) {
-						$enabled = dol_eval($extrafields->attributes[$this->table_element]['enabled'][$key], 1, 1, '1');
+						$enabled = dol_eval($extrafields->attributes[$this->table_element]['enabled'][$key], 1, 1, '0');
```
eldy, line 7746: Param '0' should be used only for 
$extrafields->attributes[$this->table_element]['computed']
not for
$extrafields->attributes[$this->table_element]['enabled']

PR #24205, opened 2023-03-14
Improve SQL Query performances

BenjaminFlr:
Instead of starting from products table and load a lot of data, we query from commande_fournisseur_dispatch, with less data to load.

### DIFF
--- a/htdocs/fourn/class/fournisseur.commande.class.php
+++ b/htdocs/fourn/class/fournisseur.commande.class.php
@@ -2403,14 +2403,21 @@ class CommandeFournisseur extends CommonOrder
 		$ret = array();
 
 		// List of already dispatched lines
-		$sql = "SELECT p.ref, p.label,";
+		$sql = "SELECT p.rowid as pid, p.ref, p.label,";
 		$sql .= " e.rowid as warehouse_id, e.ref as entrepot,";
-		$sql .= " cfd.rowid as dispatchedlineid, cfd.fk_product, cfd.qty, cfd.eatby, cfd.sellby, cfd.batch, cfd.comment, cfd.status, cfd.fk_commandefourndet";
-		$sql .= " FROM ".MAIN_DB_PREFIX."product as p,";
-		$sql .= " ".MAIN_DB_PREFIX."commande_fournisseur_dispatch as cfd";
+		$sql .= " cfd.rowid as dispatchlineid, cfd.fk_product, cfd.qty, cfd.eatby, cfd.sellby, cfd.batch, cfd.comment, cfd.status, cfd.datec";
+		$sql .= " ,cd.rowid, cd.subprice";
+		if ($conf->reception->enabled) {
+			$sql .= " ,cfd.fk_reception, r.date_delivery";
+		}
+		$sql .= " FROM ".MAIN_DB_PREFIX."commande_fournisseur_dispatch as cfd";
+		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."product as p ON cfd.fk_product = p.rowid";
+		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."commande_fournisseurdet as cd ON cd.rowid = cfd.fk_commandefourndet";
 		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."entrepot as e ON cfd.fk_entrepot = e.rowid";
-		$sql .= " WHERE cfd.fk_commande = ".((int) $this->id);
-		$sql .= " AND cfd.fk_product = p.rowid";
+		if ($conf->reception->enabled) {
+			$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."reception as r ON cfd.fk_reception = r.rowid";
+		}
+		$sql .= " WHERE cfd.fk_commande = ".((int) $object->id);
 		if ($status >= 0) {
 			$sql .= " AND cfd.status = ".((int) $status);
 		}
--- a/htdocs/fourn/class/fournisseur.commande.class.php
+++ b/htdocs/fourn/class/fournisseur.commande.class.php
@@ -2411,9 +2411,9 @@ class CommandeFournisseur extends CommonOrder
 		}
 		$sql .= " FROM ".MAIN_DB_PREFIX."commande_fournisseur_dispatch as cfd";
 		$sql .= " INNER JOIN ".MAIN_DB_PREFIX."product as p ON cfd.fk_product = p.rowid";
-		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."commande_fournisseurdet as cd ON cd.rowid = cfd.fk_commandefourndet";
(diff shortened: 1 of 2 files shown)


htdocs/fourn/commande/dispatch.php
```diff
@@ -1128,16 +1128,15 @@
 	if ($conf->reception->enabled) {
 		$sql .= " ,cfd.fk_reception, r.date_delivery";
 	}
-	$sql .= " FROM ".MAIN_DB_PREFIX."product as p,";
-	$sql .= " ".MAIN_DB_PREFIX."commande_fournisseur_dispatch as cfd";
+	$sql .= " FROM ".MAIN_DB_PREFIX."commande_fournisseur_dispatch as cfd";
+	$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."product as p ON cfd.fk_product = p.rowid";
```
eldy, line 1132: Why adding a LEFT JOIN here and not INNER JOIN ?  
cfd.fk_product should always be defined to a value existing in product table.

htdocs/fourn/class/fournisseur.commande.class.php
```diff
@@ -2403,14 +2403,21 @@ public function getDispachedLines($status = -1)
...
-		$sql .= " cfd.rowid as dispatchedlineid, cfd.fk_product, cfd.qty, cfd.eatby, cfd.sellby, cfd.batch, cfd.comment, cfd.status, cfd.fk_commandefourndet";
-		$sql .= " FROM ".MAIN_DB_PREFIX."product as p,";
-		$sql .= " ".MAIN_DB_PREFIX."commande_fournisseur_dispatch as cfd";
+		$sql .= " cfd.rowid as dispatchlineid, cfd.fk_product, cfd.qty, cfd.eatby, cfd.sellby, cfd.batch, cfd.comment, cfd.status, cfd.datec";
+		$sql .= " ,cd.rowid, cd.subprice";
+		if ($conf->reception->enabled) {
+			$sql .= " ,cfd.fk_reception, r.date_delivery";
+		}
+		$sql .= " FROM ".MAIN_DB_PREFIX."commande_fournisseur_dispatch as cfd";
+		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."product as p ON cfd.fk_product = p.rowid";
```
eldy, line 2414: Using a LEFT JOIN instead of INNER JOIN increase the cost of the execution. Why not INNER JOIN ?

htdocs/fourn/class/fournisseur.commande.class.php
```diff
@@ -2403,14 +2403,21 @@ public function getDispachedLines($status = -1)
...
 
 		// List of already dispatched lines
-		$sql = "SELECT p.ref, p.label,";
+		$sql = "SELECT p.rowid as pid, p.ref, p.label,";
 		$sql .= " e.rowid as warehouse_id, e.ref as entrepot,";
-		$sql .= " cfd.rowid as dispatchedlineid, cfd.fk_product, cfd.qty, cfd.eatby, cfd.sellby, cfd.batch, cfd.comment, cfd.status, cfd.fk_commandefourndet";
-		$sql .= " FROM ".MAIN_DB_PREFIX."product as p,";
-		$sql .= " ".MAIN_DB_PREFIX."commande_fournisseur_dispatch as cfd";
+		$sql .= " cfd.rowid as dispatchlineid, cfd.fk_product, cfd.qty, cfd.eatby, cfd.sellby, cfd.batch, cfd.comment, cfd.status, cfd.datec";
+		$sql .= " ,cd.rowid, cd.subprice";
```
eldy, line 2409: This 2 new fields seems to not be used.

htdocs/fourn/class/fournisseur.commande.class.php
altairis-benjamin: My bad, the left join was part of a specific development, but INNER JOIN is better right now.

htdocs/fourn/class/fournisseur.commande.class.php
altairis-benjamin: Same answer as for the first point: due to a specific development.

htdocs/fourn/class/fournisseur.commande.class.php
```diff
@@ -2403,14 +2403,20 @@ public function getDispachedLines($status = -1)
...
-		$sql .= " cfd.rowid as dispatchedlineid, cfd.fk_product, cfd.qty, cfd.eatby, cfd.sellby, cfd.batch, cfd.comment, cfd.status, cfd.fk_commandefourndet";
-		$sql .= " FROM ".MAIN_DB_PREFIX."product as p,";
-		$sql .= " ".MAIN_DB_PREFIX."commande_fournisseur_dispatch as cfd";
+		$sql .= " cfd.rowid as dispatchlineid, cfd.fk_product, cfd.qty, cfd.eatby, cfd.sellby, cfd.batch, cfd.comment, cfd.status, cfd.datec";
+		if ($conf->reception->enabled) {
+			$sql .= " ,cfd.fk_reception, r.date_delivery";
+		}
+		$sql .= " FROM ".MAIN_DB_PREFIX."commande_fournisseur_dispatch as cfd";
+		$sql .= " INNER JOIN ".MAIN_DB_PREFIX."product as p ON cfd.fk_product = p.rowid";
+		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."commande_fournisseurdet as cd ON cd.rowid = cfd.fk_commandefourndet";
```
eldy, line 2414: It seems we don't need this join on table cd in common case. We need it only `if ($conf->reception->enabled) {`
so can you move it into the condition ?

htdocs/fourn/commande/dispatch.php
eldy: LEFT to change into INNER

htdocs/fourn/class/fournisseur.commande.class.php
```diff
@@ -2403,14 +2403,20 @@ public function getDispachedLines($status = -1)
 		$ret = array();
 
 		// List of already dispatched lines
-		$sql = "SELECT p.ref, p.label,";
+		$sql = "SELECT p.rowid as pid, p.ref, p.label,";
 		$sql .= " e.rowid as warehouse_id, e.ref as entrepot,";
-		$sql .= " cfd.rowid as dispatchedlineid, cfd.fk_product, cfd.qty, cfd.eatby, cfd.sellby, cfd.batch, cfd.comment, cfd.status, cfd.fk_commandefourndet";
-		$sql .= " FROM ".MAIN_DB_PREFIX."product as p,";
-		$sql .= " ".MAIN_DB_PREFIX."commande_fournisseur_dispatch as cfd";
+		$sql .= " cfd.rowid as dispatchlineid, cfd.fk_product, cfd.qty, cfd.eatby, cfd.sellby, cfd.batch, cfd.comment, cfd.status, cfd.datec";
```
eldy, line 2408: You remove the field cfd.fk_commandefourndet and replace it with cfd.datec.
But late, we do
`'orderlineid' => $objp->fk_commandefourndet`
So this will introduce a bug

PR #24218, opened 2023-03-15
NEW function in modulebuilder.lib for rewriting all permissions

lamrani002:
# NEW|New function for rewriting all permissions after every update or delete or add

### DIFF
--- a/htdocs/core/lib/modulebuilder.lib.php
+++ b/htdocs/core/lib/modulebuilder.lib.php
@@ -416,3 +416,71 @@ function rebuildObjectSql($destdir, $module, $objectname, $newmask, $readdir = '
 
 	return $error ? -1 : 1;
 }
+
+/**
+ * Rewriting all permissions after updating file
+ * @param string         $file         file with path
+ * @param string         $objectname   name of the permission object
+ * @param string         $right        permission of object
+ * @param string         $rightUpdated permission of object updated
+ * @param int            $action       0 for delete, 1 for add, 2 fo update
+ * @return void
+ */
+function rewritingAllPermissions($file, $objectname, $right, $rightUpdated, $action)
+{
+	global $langs;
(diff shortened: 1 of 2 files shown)


htdocs/core/lib/modulebuilder.lib.php
```diff
@@ -416,3 +416,70 @@ function rebuildObjectSql($destdir, $module, $objectname, $newmask, $readdir = '
...
+	$error = 0;
+	if (empty($right)) {
+		$error++;
+	}
+	if ($action == 2 && empty($rightUpdated)) {
+		$error++;
+	}
+	if (!$error) {
+		// Open the file and read line by line
+		$handle = fopen($file, "r");
```
Hystepik, line 444: Here we prefer to use a file_get_contents() for security purpose

lamrani002: ok , i see

htdocs/core/lib/modulebuilder.lib.php
```diff
@@ -416,3 +416,70 @@ function rebuildObjectSql($destdir, $module, $objectname, $newmask, $readdir = '
...
+					}
+					$lines[] = $line;
+					$i++;
+				}
+				break;
+			}
+			$i++;
+		}
+		$allContent = implode("", $lines);
+		if (str_contains($allContent, $right)) {
```
Hystepik, line 467: A better way to make this is to reforge entirely the /* BEGIN MODULEBUILDER PERMISSIONS / / END MODULEBUILDER PERMISSIONS */.
This way we can modify permission even if user as modified the file

lamrani002: I make some update if you can check and tell me if that what we need

Hystepik:
@eldy seams ok to me

PR #24220, opened 2023-03-15
fix missing include

frederic34:


htdocs/comm/propal/agenda.php
```diff
@@ -246,6 +246,7 @@
 		$filters['search_rowid'] = $search_rowid;
 
 		// TODO Replace this with same code than into list.php
+		require_once DOL_DOCUMENT_ROOT.'/core/lib/company.lib.php';
 		show_actions_done($conf, $langs, $db, $object, null, 0, $actioncode, '', $filters, $sortfield, $sortorder, property_exists($object, 'module') ? $object->module : '');
```
eldy, line 250: Is it for a need inside show_actions_done...  ?
If yes, may be we should move the include into  show_actions_done  ?

frederic34: I have show_actions_done unknown funtion

PR #24225, opened 2023-03-15
FIX FILTER_VALIDATE_EMAIL is not a string and limit to POST

hregis:


hregis:
@eldy ready to merge

htdocs/core/class/html.formticket.class.php
```diff
@@ -49,10 +49,15 @@ class FormTicket
 	public $db;
 
 	/**
-	 * @var string	The track_id of the ticket. Used also for the $keytoavoidconflict to name session vars to upload files.
+	 * @deprecated
```
eldy, line 52: No the comment was good.
track_id is not trackid.
They are 2 different fields used for 2 different purposes.

htdocs/public/ticket/ajax/ajax.php
```diff
@@ -65,6 +66,12 @@
 	httponly_accessforbidden('Option TICKET_CREATE_THIRD_PARTY_WITH_CONTACT_IF_NOT_EXIST of module ticket is not enabled');
 }
 
+if (empty($token) || $token !== currentToken()) {
```
eldy, line 69: Because the page that is calling the ajax is a public page, using a token is useless because the token will be also public.
Goal of token is to avoid to allow CSRF attack so allow an attacker to execute the page into the session of another user. But here this is a public component and any body can execute it so what is goal of blocking the use into the session of another user when you are already allowed to execute it in your own session.

htdocs/core/class/html.formticket.class.php
```diff
@@ -264,23 +266,26 @@ public function showForm($withdolfichehead = 0, $mode = 'edit', $public = 0, Con
...
 								},
 								function(response) {
+									var response = jQuery.parseJSON(response);
 									if (response.error) {
                                         jQuery("#contact_search_result").html("<span class=\"error\">"+response.error+"</span>");
 									} else {
                                         var contact_list = response.contacts;
-										if (contact_list.length == 1) {
+
+										if (typeof(contact_list) !== "undefined" && contact_list.length) {
```
eldy, line 283: What happen if the response return 2 record ?

hregis: @eldy if 2 records with the same email address it displays the first, it's not good I presume? :-)

hregis: @eldy if nothing is displayed when it finds more than 1, there will be the possibility of creating a new contact at the risk of creating one at each request!

hregis: @eldy I just tested with "contact_list.length == 1" and indeed with each ticket request the user creates a new third party with the same name!

hregis: @eldy to have this type of functionality it would be preferable to have a customer account access and a customer account request!

eldy: > @eldy to have this type of functionality it would be preferable to have a customer account access and a customer account request!

Sure. Autofill data based on a user input with outside of a login session will always means "no good duplicate management" and also "no security". It means you can guess and retreive data by entering random entry. 

On other pages, you must first enter   "A Public Tracking ID" and an "Email" and email must match the email of the ticket of the entered public tracking id to be into a valid session and see data. It should be the same for the call of this page. If we want to retreive a contact, search should not be done with email only but with couple email + a tracking id of a ticket previously entered.
This is why this feature `"if ($with_contact) {" ` is disabled by default.

eldy:
I try to close this ticket. But i don't understand the initial trouble. Can you explain what is the initial problem this PR try to fix ?

PR #24238, opened 2023-03-16
FIX #24138 Fix box_birthdays SQL for postgres

### LINKED ISSUE #24138: Migration issue 15.0.4 -> 17.0.0: SQL error on "birthday for this month"
### Bug

After migrating from 15.0.4 to 17.0.0, the following error happens in the birthday panel (in french "anniversaire de ce mois").

```
 ERROR: 0A000: invalid UNION/INTERSECT/EXCEPT ORDER BY clause LINE 1: ...te_format( u.dateemployment, '%c') = '3' ORDER BY DAY(datea)... ^ DETAIL: Only result column names can be used, not expressions or functions. HINT: Add the expression/function to every SELECT, or move the UNION into a FROM clause. LOCATION: transformSetOperationStmt, analyze.c:1776 sql=SELECT u.rowid, u.firstname, u.lastname, u.birth as datea, 'birth' as typea,…
```

There were no errors in 15.0.4.

### Environment Version

17.0.0

### Environment OS

Debian bullseye

### Environment Web server

Nginx

### Environment PHP

PHP 7.4

### Environment Database

Postgres Scaleway

### Environment URL(s)

_No response_

Expected : no birthday this month

Actual behaviour: SQL error message

Install a 15.0.4 instance, upgrade to 17.0.0 with `php upgrade.php / upgrade2.php 15.0.4 17.0.0`.

I can produce such instance in a few days to ensure it can be reproduced from scratch, and share database anyway.

### Attached files

In 15.0.4:

image

In 17.0.0:

image

lastmikoi:
# FIX|Fix #24138 
PR #21631's implementation causes issues with PostgreSQL and potentially
other stricter RDBMSes, indeed doing an ORDER BY using a function is not
supported.

In this fix we're computing the day of each user birthday and
employment day in the SELECT columns, and use those columns as a sorting
key.

### DIFF
--- a/htdocs/core/boxes/box_birthdays.php
+++ b/htdocs/core/boxes/box_birthdays.php
@@ -85,19 +85,19 @@ class box_birthdays extends ModeleBoxes
 		if ($user->rights->user->user->lire) {
 			$tmparray = dol_getdate(dol_now(), true);
 
-			$sql = "SELECT u.rowid, u.firstname, u.lastname, u.birth as datea, 'birth' as typea, u.email, u.statut as status";
+			$sql = "SELECT u.rowid, u.firstname, u.lastname, u.birth as datea, date_format(u.birth, '%d') as daya, 'birth' as typea, u.email, u.statut as status";
 			$sql .= " FROM ".MAIN_DB_PREFIX."user as u";
 			$sql .= " WHERE u.entity IN (".getEntity('user').")";
 			$sql .= " AND u.statut = 1";
 			$sql .= dolSqlDateFilter('u.birth', 0, $tmparray['mon'], 0);
 			$sql .= ' UNION ';
-			$sql .= "SELECT u.rowid, u.firstname, u.lastname, u.dateemployment as datea, 'employment' as typea, u.email, u.statut as status";
+			$sql .= "SELECT u.rowid, u.firstname, u.lastname, u.dateemployment as datea, date_format(u.dateemployment, '%d') as daya, 'employment' as typea, u.email, u.statut as status";
 			$sql .= " FROM ".MAIN_DB_PREFIX."user as u";
 			$sql .= " WHERE u.entity IN (".getEntity('user').")";
 			$sql .= " AND u.statut = 1";
 			$sql .= dolSqlDateFilter('u.dateemployment', 0, $tmparray['mon'], 0);
 
-			$sql .= " ORDER BY DAY(datea) ASC";
+			$sql .= " ORDER BY daya ASC";
 
 			dol_syslog(get_class($this)."::loadBox", LOG_DEBUG);
 			$result = $this->db->query($sql);


htdocs/core/boxes/box_birthdays.php
```diff
@@ -85,19 +85,19 @@ public function loadBox($max = 20)
 		if ($user->rights->user->user->lire) {
 			$tmparray = dol_getdate(dol_now(), true);
 
-			$sql = "SELECT u.rowid, u.firstname, u.lastname, u.birth as datea, 'birth' as typea, u.email, u.statut as status";
+			$sql = "SELECT u.rowid, u.firstname, u.lastname, u.birth as datea, date_format(u.birth, '%d') as daya, 'birth' as typea, u.email, u.statut as status";
```
eldy, line 88: Why do you need the date_format() ? Using this make the database not able to use the index.
Why not just 
u.birth as daya ?

htdocs/core/boxes/box_birthdays.php
```diff
@@ -85,19 +85,19 @@ public function loadBox($max = 20)
...
 			$sql .= ' UNION ';
-			$sql .= "SELECT u.rowid, u.firstname, u.lastname, u.dateemployment as datea, 'employment' as typea, u.email, u.statut as status";
+			$sql .= "SELECT u.rowid, u.firstname, u.lastname, u.dateemployment as datea, date_format(u.dateemployment, '%d') as daya, 'employment' as typea, u.email, u.statut as status";
 			$sql .= " FROM ".MAIN_DB_PREFIX."user as u";
 			$sql .= " WHERE u.entity IN (".getEntity('user').")";
 			$sql .= " AND u.statut = 1";
 			$sql .= dolSqlDateFilter('u.dateemployment', 0, $tmparray['mon'], 0);
 
-			$sql .= " ORDER BY DAY(datea) ASC";
+			$sql .= " ORDER BY daya ASC";
```
eldy, line 100: Why not just doing ORDER BY datea ?
Adding a field "date_format(u.birth, '%d') as daya" to make an order means the database will not be able to use the index.

lastmikoi: That was my initial intuition as well (see https://github.com/Dolibarr/dolibarr/commit/8be998a9066d5f5c9d3282f106f8c9f81e833357) but after asking the original author (https://github.com/Dolibarr/dolibarr/issues/24138#issuecomment-1472074229) he mentioned that the birthday box needs to specifically sort on the birth day, while sorting on `datea` would sort people born in earlier years first, even if their actual day of birth is later.

lastmikoi: As far as I can tell, and I may mistaken, DAY(datea) on MySQL would also need to apply a SQL function over the whole index, then sort that result. So if we both want to fix SQL strictness *and* improve performance by sorting on an indexed field, then we probably should add an index to the schema that contains the day number of both the birthday and the employment date.

Changing the schema is however beyond my current capabilities and familiarity with the code, but I can give it a try if you deem it necessary !

lastmikoi:
After a bit more digging, it appears that using a SQL function to `ORDER BY` isn't what's causing the issue, it's only in the context of a `UNION`, which explains why this issue doesn't impact `box_birthdays_members.php` and why the issue only appeared when the `box_birthdays` has been enriched with employment "birth"days

eldy:
> box_birthdays_members

So can you paste the SQL error you have with postgresql ?

lastmikoi:
> > box_birthdays_members
> 
> So can you paste the SQL error you have with postgresql ?

The error is detailed in the original bug report https://github.com/Dolibarr/dolibarr/issues/24138#issuecomment-1472065386

Here is it copy/pasted:

```sql
SELECT u.rowid, u.firstname, u.lastname, u.birth as datea, 'birth' as typea, u.email, u.statut as status FROM llx_user as u WHERE u.entity IN (0,1) AND u.statut = 1 AND  date_format( u.birth, '%c') = '3' UNION SELECT u.rowid, u.firstname, u.lastname, u.dateemployment as datea, 'employment' as typea, u.email, u.statut as status FROM llx_user as u WHERE u.entity IN (0,1) AND u.statut = 1 AND  date_format( u.dateemployment, '%c') = '3' ORDER BY DAY(datea) ASC;
ERROR:  invalid UNION/INTERSECT/EXCEPT ORDER BY clause
LINE 1: ...te_format( u.dateemployment, '%c') = '3' ORDER BY DAY(datea)...
                                                             ^
DETAIL:  Only result column names can be used, not expressions or functions.
HINT:  Add the expression/function to every SELECT, or move the UNION into a FROM clause.
```

eldy:
> > > box_birthdays_members
> > 
> > 
> > So can you paste the SQL error you have with postgresql ?
> 
> The error is detailed in the original bug report [#24138 (comment)](https://github.com/Dolibarr/dolibarr/issues/24138#issuecomment-1472065386)
> 
> Here is it copy/pasted:
> 
> ```sql
> SELECT u.rowid, u.firstname, u.lastname, u.birth as datea, 'birth' as typea, u.email, u.statut as status FROM llx_user as u WHERE u.entity IN (0,1) AND u.statut = 1 AND  date_format( u.birth, '%c') = '3' UNION SELECT u.rowid, u.firstname, u.lastname, u.dateemployment as datea, 'employment' as typea, u.email, u.statut as status FROM llx_user as u WHERE u.entity IN (0,1) AND u.statut = 1 AND  date_format( u.dateemployment, '%c') = '3' ORDER BY DAY(datea) ASC;
> ERROR:  invalid UNION/INTERSECT/EXCEPT ORDER BY clause
> LINE 1: ...te_format( u.dateemployment, '%c') = '3' ORDER BY DAY(datea)...
>                                                              ^
> DETAIL:  Only result column names can be used, not expressions or functions.
> HINT:  Add the expression/function to every SELECT, or move the UNION into a FROM clause.
> ```

I see, thanks for clarification. Let's go with your fix suggestion.

lastmikoi:
Thanks for the merge !

PR #24245, opened 2023-03-17
FIX FILTER_VALIDATE_EMAIL param is not a string

hregis:


lvessiller-opendsi:
@hregis
You're right it's an integer
@eldy 
It's the correct way to fix it otherwise email is null even if we have a valid email address in get param.

lvessiller-opendsi (approved):

PR #24246, opened 2023-03-17
FIX : Fix case when the value of a extrafields of the type 'boolean', 'select' or other have an option with a value equal to '0'.

kkhelifa-opendsi:
Fix case when the value of a extrafields of the type 'boolean', 'select' or other have an option with a value equal to '0'.

It's not reselected when the filter is set.


eldy:
Can you describe the setup you do to reproduce this bug ?

kkhelifa-opendsi:
Make a extrafields of type 'boolean' and try to filter with the 'no' choice on the list.
The filter is not kept (but the filter is applied)

PR #24252, opened 2023-03-17
Invoices - Add valid SwissQR code on Crabe & Sponge 

OdyX:
Heavily inspired by @benvia's  https://github.com/sinabe/dolibarr-swissqr, here comes an integration of valid SwissQr bottom-page invoices in both Crabe and Sponge PDF invoice models.

Most of the code that matters is:
* in `./htdocs/core/modules/facture` for the actual QR bill insertion;
* in `./htdocs/includes` to add the required dependencies;
* in `./htdocs/includes/sprain/swiss-qr-bill/autoload.php` to get autoload the above.

I have heavily tested that when not activated, the generated invoices don't get changed at all!

Here's what an example Sponge-generated invoice can look like;
[IN2303-0001.pdf](https://github.com/Dolibarr/dolibarr/files/11002927/IN2303-0001.pdf)

I'm happy and eager to work on that branch to merge any needed changes!


OdyX:
@eldy any hints here? How could we update that PR to make it find its way towards a Dolibarr version ?

eldy:
The PR add +2000 new files ?
This looks strange when we are already able to draw on PDF and generate barcode image into PDF. Another side effect of this fu... composer tool.
composer is a very bad practice that brings a ton of files we don't need, so a ton of security vulnerabilities, and a ton of trouble when we need to upgrade. 
Can you check to reduce this and integrate only the minimum files we need ?

OdyX:
Thank you for your initial feedback.

As you can see from the [PR diff](https://github.com/Dolibarr/dolibarr/pull/24252/files), after excluding includes/ and language updates, there are only 4 files that are updated; and these are the ones which need a review:
* `htdocs/admin/pdf_other.php`
* `htdocs/core/modules/facture/doc/pdf_crabe.modules.php`
* `htdocs/core/modules/facture/doc/pdf_sponge.modules.php`
* `htdocs/core/modules/facture/modules_facture.php`
* `htdocs/includes/sprain/swiss-qr-bill/autoload.php` (specifically to auto-load these dependencies)

I also made sure to only include the dependencies in a single commit https://github.com/Dolibarr/dolibarr/pull/24252/commits/b28d8a07a7e3a42cddefe3ec3e32f9678ca71905

For the rest, I will of course **try** to investigate to reduce the number of included dependencies to the strict minimum. As you write, there's already code to write to PDFs, but the goal of that PR was precisely to rely on an existing library doing most of the work for SwissQR specifically, in a shared and standard way. Having the all the minutiae of the "legally required" formatting of SwissQRs in a library is a great way to **not have** to maintain that specifically in Dolibarr.

Finally, I see from issues and PRs that you have taken a firm stance against relying on `composer` to maintain the projects' dependencies, so I will not use this PR to try changing your mind; I will solely focus on trying to get this acceptable for you so that it can be merged, as it's really going to be _very_ critical for Swiss businesses to adopt Dolibarr; a nice-looking SwissQR at the bottom of invoices is basically a requirement since September 2022.

eldy:
The PR has error on continuous integration. This is a side effect of composer (that is used by travis automatically to  install other versions of some components). Here it installed a second version of jquery (some sub-libraries need a version of jquery, other need another one, but to avoid trouble we want to deal with only 1 jquery version in the project, so, to avoid problem due to duplicated version of jquery files, a phpunit test is done to ensure we don't have such duplicates). This makes the Continuous Test fails due to the duplicate.

Fix should be easy: Just remove the composer.json of the added component or renamed it into composer.json.disabled so we still have the file and a simple rename allow to re-use composer. Doing this should be enough to avoid a composer update that reinstall another version of jquery.

The 5 modified core files are ok.

The risk on security/maintenance/code conflict is on the + 2 000 other files.  Past experiences using intensive composer shows we introduce 1 critical security vulnerability for every 100 files added by composer. Here we add + 2000 files !  This is why after doing a composer update to update a lib, and after patching the lib with the known bug fix defined into dolibarr_changes.txt file (because composer erases the important bug fixes we did to library as most are not propagated into official project or propagated too late by project owner), we remove all the non necessary files when possible. 
For example with php-imap, that has not a lot of dependencies, by just "deleting files of all non used library", we saved 3 MB of files/code for the same result/feature. We removed also a ton of vulnerability. For information each vulnerability cost us an average of 100 $ per vulnerability with our private bug bounty campaigns).

OdyX:
Thanks for your answer, and for providing a pathway to merge this.

> The PR has error on continuous integration. This is a side effect of composer (that is used by travis automatically to install other versions of some components). Here it installed a second version of jquery (some sub-libraries need a version of jquery, other need another one, but to avoid trouble we want to deal with only 1 jquery version in the project, so, to avoid problem due to duplicated version of jquery files, a phpunit test is done to ensure we don't have such duplicates). This makes the Continuous Test fails due to the duplicate.
> 
> Fix should be easy: Just remove the composer.json of the added component or renamed it into composer.json.disabled so we still have the file and a simple rename allow to re-use composer. Doing this should be enough to avoid a composer update that reinstall another version of jquery.

Fair enough. I clearly understand the requirement to only have and ship one version of each library; especially of jquery.

I have now renamed the `composer.json` from the diff to `composer.json.disabled`

> The 5 modified core files are ok.

Great, good news. 

> The risk on security/maintenance/code conflict is on the + 2 000 other files. Past experiences using intensive composer shows we introduce 1 critical security vulnerability for every 100 files added by composer.

By this, you're saying that 1% of the open-source PHP ecosystem's files ship "critical security vulnerabilities" in recent versions, as shipped by composer. I don't have numbers, but that seems _enormous_ (if you had said "unexpected behaviours", I'd clearly agree, but using "critical security vulnerability" (each with a CVE number?) seems like a slight exaggeration.

> Here we add + 2000 files ! This is why after doing a composer update to update a lib, and after patching the lib with the known bug fix defined into dolibarr_changes.txt file (because composer erases the important bug fixes we did to library as most are not propagated into official project or propagated too late by project owner), we remove all the non necessary files when possible.

If that is a pathway you're interested to take, I would **highly recommend** taking an alternative route, for the sake of the future maintainability of Dolibarr: I would clearly **embrace composer fully, but use it to its full power**. Let me extend a bit.

I read from the various PRs and issues (including some really old ones) that the project experiences issues with composer, that range from version mixing issues, to unexpected changes in new versions, to the difficulty of integrating patches and updates to dependencies, etc. And that from that, the project's taking the alternative route of embedding the dependencies in its own code repository, (and that's the important point) **under its own responsibility**. As of today, 78M / 2000+ files, that are directly committed in the project. By doing so, the tie to the upstream project is cut, versioning history is lost, version tracking becomes manual, upstream fixes become harder to integrate etc. Of course, this makes it much more direct to patch; and makes Dolibarr easier to ship (unzip the archive, done). But responsibility has shifted; if a critical vulnerability is discovered in `sabre/sabre`, then Dolibarr has to 1: get informed / find out; 2: go find the fix by itself 3: manually apply it in its code copy (while maintaining the additional patches) 4: inform its own users about the vulnerability and its fixing. It's a PITA.

composer, **when used correctly, and to its full powers** is **designed** to help a project like Dolibarr with the maintenance of its dependencies
* by having **one and only one** `composer.json` for Dolibarr, you can ensure library unicity (no more duplicated jquery) and compatibility (no update can break other components' dependencies)
* by using **strong versioning constraints** in the `composer.json` file, you can make sure no unexpected versions get installed; that no upgrade goes unnoticed.
* by **shipping the composer.lock file**, either directly in history, or for release artifacts, you can ensure that all users of the releases get _exactly_ the intended dependencies, in their precise versions.
* by **referencing external repositories** (see [this doc](https://medium.com/swlh/using-your-own-forks-with-composer-699358db05d9), for example), you can maintain patched forks of upstream projects as your needs evolve; before upstream merges a patch, or on the long-term if there's upstream disagreement.

The latter point is the most important I think, so let me rephrase it: it is **possible** for Dolibarr to **both** maintain patched forks **and** rely on composer. Actually, it's going to be **much cleaner** as upstream history will be kept, and the Dolibarr code source smaller and cleaner.

Bref. It's really crystal clear to me (and, I'm sure, large parts of the general PHP ecosystem) that the project would be in a **much better position** with a good usage (including lots of expert-level usage) of composer for dependencies' maintenance, than it is now. This very PR is a good example of that; it makes a lot of sense to rely on a specialized and specific Swiss QR library, that is kept external from the project, rather than custom-implementing this internally. That library is likely to evolve with legal or cultural requirements as time goes, and having to manually replicate that in a code copy in Dolibarr is just wasting the precious and costly developer time that could be better spent elsewhere (or merely reviewing version updates in a composer.json/.lock pair).

OdyX:
Added a fix for when the SwissQR wasn't enabled, also rebased on `develop`.

OdyX:
Fixed another rebase typo, and more file removals

eldy:
I merged the PR but i experienced some troubles.
It seems feature is not compatible with a lot of PHP version. This is a common problem when using composer. It can't work on most environment.
So i modified the setup "Enable switzerland barcode" to accept 3 values instead of 2. 
Value No, Value yes (barcode in top like in v17) and value Bottom (using the new feature that is experimental since it does not work yet). I fixed some bugs but more test for your are welcome due to the changes i did to restore the old working version keeping the new working.

OdyX:
Yay, thanks for merging. The QR code still works for me fine, at least. Thanks for your changes too.

benvia:
Thank @Ody for your integration of the swiss qr bill.

This is a partial integration as the dependencies are not merged into Dolibarr.

@eldy The question is, should we add the dependencies to Dolibarr and add complexity for maintenance, or create a separate open source module specifically for this need, or even for other needs of Switzerland ?

eldy:
> Thank @Ody for your integration of the swiss qr bill.
> 
> This is a partial integration as the dependencies are not merged into Dolibarr.
> 
> @eldy The question is, should we add the dependencies to Dolibarr and add complexity for maintenance, or create a separate open source module specifically for this need, or even for other needs of Switzerland ?

As the dependecy does not follow same prerequisites than Dolibarr and is a very large set of files, the best is to provide an exyernal module that add the depebdency and activate the feature.

dazu89:
Hi there! Under Dolibarr 21.0.1, PHP 8.3 and Mysql 8.4 this feature can be activated under Modules → Invoices → Setup → Invoice documents models → Swiss QR invoice. 

However, I noticed three problems:

1. In German language, the column width is to narrow, such that that the table's header text gets broken over one line and overlaps into the next row of the table.
2. The payment slip uses the ›Company name‹ instead of the bank ›Account owner name‹ which may not co-inside.
3. Payment slip can only be generated on a separate page?

eldy:
> Hi there! Under Dolibarr 21.0.1, PHP 8.3 and Mysql 8.4 this feature can be activated under Modules → Invoices → Setup → Invoice documents models → Swiss QR invoice. 
> 
> However, I noticed three problems:
> 
> 1. In German language, the column width is to narrow, such that that the table's header text gets broken over one line and overlaps into the next row of the table.
> 2. The payment slip uses the ›Company name‹ instead of the bank ›Account owner name‹ which may not co-inside.
> 3. Payment slip can only be generated on a separate page?

Can you provide a screenshot that show the trouble ?

benvia:
Hello,

I'm now ready with a working beta version of  the Swiss QR invoice module. For now it support only the Sponge invoice model.

Minimum requirements is PHP 8.1 and Dolibarr 18.
Currently tested up to Dolibarr 21

You can find the module on the v2 branch of the [Sinabe SwissQR repository](https://github.com/sinabe/dolibarr-swissqr)

Feedbacks are welcome.

If this is an acceptable solution, I will submit a pull request to remove the code that has become obsolete in the Sponge template and modules_facture.php.

gagarine:
@OdyX I opened an issue to discuss the composer situation ->  https://github.com/Dolibarr/dolibarr/issues/35108

benvia:
The current Dolibarr implementation of Swiss QR invoice will stop working on 22.11.2025.
A structured postal address will be required after this date (https://www.six-group.com/dam/download/banking-services/standardization/qr-bill/ig-qr-bill-v2.3-en.pdf see 4.3.1).

gagarine:
@benvia Their is a foot note: 

> For the time being, the specification of the house number (element "BldgNbOrAdrLine2") in
> the element "StrtNmOrAdrLine1" is allowed and will not be rejected when the order is
> placed. When creating the QR-bill, however, it is important to ensure that the debtor's
> address is complete and correct, despite this tolerance. This data must be fully available
> and recognisable to the system when the cash is deposited at the counter, otherwise
> processing of the QR-bill may be rejected.

Say that, it's the ISO 20022 standards that push for that. Dolibar should **absolutely** store all adresses in a structural way because the full Swift system will comply and enforce this sooner than later https://www.swift.com/standards/iso-20022/iso-20022-faqs/brief-introduction-postal-address-field .

benvia:
@gagarine it's why I created a working module for Swiss QR including the required change for address. By the way the module also support reference (BESR-ID).

There are other reasons for using an external module:
- For the moment, this is the fastest and most reliable way to make Swiss QR invoices work.
- No need to backport changes to multiple versions of Dolibarr.
- Does not overload all Dolibarr installations worldwide just for Switzerland.
- Allows us to avoid having to keep up with Dolibarr's release schedule to make improvements, changes, or update of dependencies.
- Follow recommendation of @eldy

sonikf:
Hi @benvia @gagarine 
I'm from Greece. 
If my company or any European company invoice a Swiss company do we have to use Swiss QR?

benvia:
@sonikf No, you don't need the Swiss QR invoice.
It's not mandatory in Switzerland either. It makes payment easier by automatically entering data using the QR code.
Payment using an IBAN is still possible.

sonikf:
Thank you @benvia

This repo created by @eldy https://github.com/Dolibarr/dolibarr-community-modules 
for modules that can't be embedded into the core allows for discoverability and ease of installation, so maybe you could ask for inclusion!

eldy:
> Thank you @benvia
> 
> This repo created by @eldy https://github.com/Dolibarr/dolibarr-community-modules 
> for modules that can't be embedded into the core allows for discoverability and ease of installation, so maybe you could ask for inclusion!
> 
> 
> 

May be a good idea to include your work inncommunity repo. Even if php dependzncies does not match dolibarr one, it could be accepted inside.
If change compared to existing code is small and module can work with current core with minor change, push in core may be ok too. Rukes for core are juste more strict.

gagarine:
@benvia thanks, it was not clear to me from your GitHub repository that your module is more modern and aims to replace the implementation in Dolibarr. I will check your repository then.

However, I still strongly believe that all address storage in Dolibarr needs to be adjusted to support Swift structured addresses. This should not be the responsibility of an external module.

PR #24253, opened 2023-03-17
FIX #24240 Dolibarr V17.0.0 PHP8 fatal error

### LINKED ISSUE #24240: Dolibarr V17.0.0 PHP8 fatal error
### Bug

When I add a product on an existing recurant invoice it fail with error 500:

PHP Fatal error: Uncaught TypeError: Unsupported operand types: string / int in /var/www/html/compta/facture/card-rec.php

With V17.0.0 and PHP 7.3.3 no problem

### Environment Version

17.0.0

### Environment OS

linux

### Environment Web server

apache

### Environment PHP

8.0

### Environment Database

mysql

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

simnandez:


htdocs/compta/facture/card-rec.php
```diff
@@ -637,7 +637,7 @@
 				$info_bits |= 0x01;
 			}
 
-			if ($usercanproductignorepricemin && (!empty($price_min) && (price2num($pu_ht) * (1 - price2num($remise_percent) / 100) < price2num($price_min)))) {
+			if ($usercanproductignorepricemin && (!empty($price_min) && (price2num($pu_ht) * (1 - price2num((float) $remise_percent) / 100) < price2num($price_min)))) {
```
eldy, line 640: What was the value of $remise_percent that cause a fatal error ?
price2num should be able ta accept a string to convert it into a float. And value may be a string that can't be converted by (float) like '1 234,56'.

simnandez: Although a discount percentage value I don't think it will reach to "1 234.56", you are right, and I have generated a more appropriate solution

PR #24255, opened 2023-03-18
FIX return $conf->entity if $object->entity is null

hregis:


hregis:
@frederic34 @eldy 

https://github.com/Dolibarr/dolibarr/pull/24164

htdocs/core/lib/functions.lib.php
```diff
@@ -82,15 +82,15 @@ function utf8_decode($elements)
 function getMultidirOutput($object, $module = '')
 {
 	global $conf;
-	if (!is_object($object)) {
+	if (!is_object($object) && empty($module)) {
```
eldy, line 85: In most cases, $module is empty and is guessed from $object. Idea of @frederic34 is to have only $object as parameter ($module is for special cases and will be always empty)

PR #24264, opened 2023-03-18
Fix #24260: Fix of hook in parseExpression for dynamic prices 

### LINKED ISSUE #24260: DynamicPrices: Hook in function parseExpression does not use returned values
### Bug

In the module DynamicPrices the function parseExpression translates a formula for a dynamic price into a value.

A hook exists to modify this function. However, when the function that is called by the hook is finished, the value of the call of the hook (reshook) is returned as the value of parseExpression instead of the result that was calculated by the hooked function (returned in resArray). This is in particular a problem, if the hook is used to replace the existing code (reshook > 0), but the return value is a  0. The existing code will then be executed and its result returned, which is not the intent..

Since the hook cannot modify the relevant input values, as they are not in the object but in parameters (values, expression), it can effectively only be used to replace the existing code, but with the return problem, in the current state, the hook is useless.

### Environment Version

17

Hook shall either replace the existing function and return a result like the existing function (i.e. the value of the parsed expression) (reshook > 0) or it should preprocess and modify some of the input values of the existing function, before it is executed (reshook == 0).

priojk:
The previously existing hook was not working, as in case of successful parsing the expression, the return value was always 1, not the calculated value of the expression.

Corrected the bug so that the hook now can either replace the existing function or it can be used to preprocess the input values.

There was a typo in the name of the hook (doDynamiPrice instead of doDynami**c**Price, c was missing). This correction is a somewhat breaking change, but since the hook was not functional anyways, I think it's reasonable to correct it as well.

### DIFF
--- a/htdocs/product/dynamic_price/class/price_parser.class.php
+++ b/htdocs/product/dynamic_price/class/price_parser.class.php
@@ -127,25 +127,13 @@ class PriceParser
 		global $user;
 		global $hookmanager;
 		$action = 'PARSEEXPRESSION';
-		$reshook = $hookmanager->executeHooks('doDynamicPrice', array(
-			'expression' =>$expression,
-			'product' => $product,
-			'values' => $values
-			), $this, $action);
-
-		if ($reshook > 0) {
-			// this function is replaced by hook, >0 also for incorrect expressions
-			return $hookmanager->resArray['return'];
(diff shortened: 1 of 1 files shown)


priojk:
Need to add better handling for hookres == 0.

PR #24268, opened 2023-03-19
Fix #24267: loading the extrafields of the related product in dynamic price expression

### LINKED ISSUE #24267: Bug: Dynamic price / parseExpression ist not loading the extrafields of product
### Bug

In the price expression for dynamic prices, extrafields of the product should be usable via #extrafield_nameofthefield#, but currently they are not loaded correctly.

### Environment Version

17

All extrafields of the product shall be usable as variables in the price expression.

priojk:
In the price expression of dynamic prices, the extrafields of the product were not loaded correctly. Amended the routine to fetch the extrafileds/optionals.

### DIFF
--- a/htdocs/product/dynamic_price/class/price_parser.class.php
+++ b/htdocs/product/dynamic_price/class/price_parser.class.php
@@ -156,11 +156,10 @@ class PriceParser
 
 		//Retrieve all extrafield for product and add it to values
 		$extrafields = new ExtraFields($this->db);
+        $extralabels = $extrafields->fetch_name_optionals_label($product->table_element);
 		$product->fetch_optionals();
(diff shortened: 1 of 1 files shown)


htdocs/product/dynamic_price/class/price_parser.class.php
```diff
@@ -156,11 +156,10 @@ public function parseExpression($product, $expression, $values)
 
 		//Retrieve all extrafield for product and add it to values
 		$extrafields = new ExtraFields($this->db);
+		$extralabels = $extrafields->fetch_name_optionals_label($product->table_element);
```
eldy, line 159: I made a little change (pushed after this PR) because it is better to use the already loaded $extrafields object if available.

PR #24272, opened 2023-03-19
ECM/GED : dol_copy same params than dol_move

rycks:
dol_move can call antivirus and index file into database, but dol_copy not ... here is a patch to add that options to dol_copy

### DIFF
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -707,12 +707,14 @@ function dolReplaceInFile($srcfile, $arrayreplacement, $destfile = '', $newmask 
  * @param	string	$destfile			Destination file (can't be a directory)
  * @param	int		$newmask			Mask for new file (0 by default means $conf->global->MAIN_UMASK). Example: '0666'
  * @param 	int		$overwriteifexists	Overwrite file if exists (1 by default)
+ * @param   int     $testvirus          Do an antivirus test. Move is canceled if a virus is found.
+ * @param	int		$indexdatabase		Index new file into database.
  * @return	int							<0 if error, 0 if nothing done (dest file already exists and overwriteifexists=0), >0 if OK
  * @see		dol_delete_file() dolCopyDir()
  */
-function dol_copy($srcfile, $destfile, $newmask = 0, $overwriteifexists = 1)
+function dol_copy($srcfile, $destfile, $newmask = 0, $overwriteifexists = 1, $testvirus = 0, $indexdatabase = 1)
 {
-	global $conf;
+	global $conf, $db, $user;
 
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/files.lib.php
```diff
@@ -754,7 +767,64 @@ function dol_copy($srcfile, $destfile, $newmask = 0, $overwriteifexists = 1)
...
+
+			if ($resultecm > 0) {
+				$result = true;
+			} else {
+				$result = false;
+			}
+		}
+	}
+
+	return $result;
```
eldy, line 827: result must be int (3 different values possible)

htdocs/core/lib/files.lib.php
```diff
@@ -754,7 +767,64 @@ function dol_copy($srcfile, $destfile, $newmask = 0, $overwriteifexists = 1)
...
+			if ($resultecm < 0) {
+				dol_syslog("Create ECM file ok", LOG_DEBUG);
+				setEventMessages($ecmfile->error, $ecmfile->errors, 'warnings');
+			} else {
+				dol_syslog("Create ECM file error", LOG_DEBUG);
+				setEventMessages($ecmfile->error, $ecmfile->errors, 'warnings');
+			}
+
+			if ($resultecm > 0) {
+				$result = true;
```
eldy, line 820: result must be int (3 different values possible)

htdocs/core/lib/files.lib.php
```diff
@@ -707,12 +707,14 @@ function dolReplaceInFile($srcfile, $arrayreplacement, $destfile = '', $newmask
  * @param	string	$destfile			Destination file (can't be a directory)
  * @param	int		$newmask			Mask for new file (0 by default means $conf->global->MAIN_UMASK). Example: '0666'
  * @param 	int		$overwriteifexists	Overwrite file if exists (1 by default)
+ * @param   int     $testvirus          Do an antivirus test. Move is canceled if a virus is found.
+ * @param	int		$indexdatabase		Index new file into database.
  * @return	int							<0 if error, 0 if nothing done (dest file already exists and overwriteifexists=0), >0 if OK
  * @see		dol_delete_file() dolCopyDir()
  */
-function dol_copy($srcfile, $destfile, $newmask = 0, $overwriteifexists = 1)
+function dol_copy($srcfile, $destfile, $newmask = 0, $overwriteifexists = 1, $testvirus = 0, $indexdatabase = 1)
```
eldy, line 715: Having index database to 1 by default was an error in dol_move. Because this param is not used yet and default behaviour is to not have indexing, i suggest to introduct the param with the default value  indexdatabase = 0  here.

PR #24278, opened 2023-03-20
Bugfix fetch_optionals before the line update to not lose the invisible fields

MaximilienR-easya:
# Fix
Fix a problem where I was losing my invisible extra fields data every time I was updating a line inside fichinter card

Now we fetch the extra fields before doing an array_merge to update the EF changed and not lose the invisible EF data

### DIFF
--- a/htdocs/fichinter/card.php
+++ b/htdocs/fichinter/card.php
@@ -605,7 +605,8 @@ if (empty($reshook)) {
 		// Extrafields
 		$extrafields->fetch_name_optionals_label($object->table_element_line);
 		$array_options = $extrafields->getOptionalsFromPost($object->table_element_line);
-		$objectline->array_options = $array_options;
+		$objectline->fetch_optionals();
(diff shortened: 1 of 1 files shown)


htdocs/fichinter/card.php
```diff
@@ -605,7 +605,8 @@
 		// Extrafields
 		$extrafields->fetch_name_optionals_label($object->table_element_line);
 		$array_options = $extrafields->getOptionalsFromPost($object->table_element_line);
-		$objectline->array_options = $array_options;
+		$objectline->fetch_optionals();
```
eldy, line 608: The fetch_optionals() has already be done into the 
if ($objectline->fetch($lineid)

eldy: You're right. Because every fetch() should contains the fetch_optionals(), I pushed a different fix that include the fetch_optional inside the fetch (like for other objects).

PR #24284, opened 2023-03-20
fix : Warning: Undefined property: stdClass:: in /home/httpd/vhosts/a…

grandoc:
…flac.fr/domains/dev.aflac.fr/httpdocs/admin/dict.php on line 2456


htdocs/admin/dict.php
```diff
@@ -2355,7 +2355,7 @@ function fieldList($fieldlist, $obj = '', $tabname = '', $context = '')
 	$withentity = '';
 
 	foreach ($fieldlist as $field => $value) {
-		if ($value == 'entity') {
+		if (isset($entity) && $value == 'entity') {
```
eldy, line 2358: isset($entity) will always be false.

PR #24285, opened 2023-03-20
New : uploadfile drag and drop on fourn invoice card

Hystepik:
Add an upload file to fourn invoice card
This add a drag and drop zone on card with a independent module

image
image


eldy:
Add also support for modulebuider and enable by default.

PR #24287, opened 2023-03-20
NEW functionality delete table of property from documentation

lamrani002:
# NEW|New function for delete table of properties in documentation if was generated


htdocs/core/lib/modulebuilder.lib.php
```diff
@@ -620,3 +620,48 @@ function writePropsInAsciiDoc($file, $objectname, $destfile)
...
+ * @param  string  $file         file or path
+ * @param  string  $objectname   name of object wants to deleted
+ * @return void
+ */
+function deletePropsFromDoc($file, $objectname)
+{
+
+	$start = "== Table of fields and their properties for object *".ucfirst($objectname)."* : ";
+	$end = "== end table for object ".ucfirst($objectname);
+	$string = getFromFile($file, $start, $end);
```
Hystepik, line 664: Here use fileGetContents($file) to get the file in a string
Then do a preg_replace() with a regex that starts with $start and ends with $end and that takes everything in beetween the 2, and replace the match.
The call looks like : preg_replace($regex, '', $stringtomodify)

lamrani002: ok, it's done

Hystepik:
@lamrani002 Conflict to solve

PR #24322, opened 2023-03-22
FIX : missing column default workstation

atm-gauthier:


htdocs/product/list.php
```diff
@@ -409,7 +411,10 @@
 
 $sql = 'SELECT p.rowid, p.ref, p.label, p.fk_product_type, p.barcode, p.price, p.tva_tx, p.price_ttc, p.price_base_type, p.entity,';
 $sql .= ' p.fk_product_type, p.duration, p.finished, p.tosell, p.tobuy, p.seuil_stock_alerte, p.desiredstock,';
-$sql .= ' p.tobatch,';
+$sql .= ' p.tobatch, ';
+if (isModEnabled('workstation')) {
+	$sql .= ' p.fk_default_workstation, ws.status as status_workstation, ws.ref as ref_workstation, ';
```
eldy, line 416: Do you set your mysql config to "strict" ? See https://wiki.dolibarr.org/index.php?title=Langages_et_normes#Sp.C3.A9cificit.C3.A9s_MySQL
I think if you are set in this recommended mode, you should get a SQL error with this change as the new fields are not inside the group by

htdocs/product/list.php
```diff
@@ -1665,6 +1682,23 @@
 		}
 	}
 
+	// Default Workstation
+	if (!empty($arrayfields['p.fk_default_worksation']['checked'])) {
+		print '<td align="center">';
+		if (!empty($obj->fk_default_workstation)) {
+			$static_ws = new Workstation($db);
```
eldy, line 1689: Move the new out of the loop, at begin of file, and reuse same object on each loop pass.

PR #24330, opened 2023-03-23
Fix odt generation

### LINKED ISSUE #16184: Probleme with line ODT generation

thomas-Ngr:
# FIX #16184 Problem with ODT generation on lines

This is a reboot of PR  #22753 targeted to develop branch. See there for initial discussion (almost no discussion).

In #14974, a line has been introduced to convert '&' chars to htmlentities. This broke the creation of ODTs with lines containing HTML. This PR fixes this for the tags that are managed in the code (br, b, strong, i, u, s, em, sub, sup, span in _replaceHtmlWithOdtTag() ).

There was many problems:

1. Lines description with html are already passed through htmlspecialchars(). Doing it another time in convertVarToOdf() will make the converted chars appear in the ODT. as mentioned in #16184.
2. Passing the values through htmlspecialchars() 'erases' the html tags. `$this->_hasHtmlTag($value)` will always be `false`. So, we have to re-set HTML closing and opening brackets.
3. HTML tags have to be converted to XML tags. But only some tags are dealt with. So we have to strip_tags() the values, authorizing only br, b, strong, i, u, s, em, sub, sup, span.
4. We should keep line breaks in the ODT file. This keeps a minimal formatting for lists in the ODT file. For example, this is an HTML `<ul>` with new code : 
image

____

edit : clarifications

We have this line description in an invoice. It has been generated with WYSISWYG editor, but manual HTML should have the same behavior:
image

With current ODT generation, the output displays HTML tags. One should reproduce it each time they create an ODT with lines description that contain HTML:
image

This is due to:
- description are passed through htmlspecialchars() before being saved in the DB
- #14974 introduced another  htmlspecialchars() to fix special chars `&` in thirdparty name
- So, descriptions are passed two times through htmlspecialchars().

We have to:
- decode descriptions from database
- deal with html->xml tags conversion
- strip HTML tags that are not managed for safety and odt readability
- keep htmlspecialchars() for normal text

This PR fixes this issue and introduces a unit test to expose and lock expected behavior.
The test separates PHP7.3 behavior from PHP7.4 behvaior, because the PHP function strip_tags() behavior changes on PHP7.4. 
Now when generating an ODT we have correct tags.
As you can see, this is a fix: lists are not managed.

image


htdocs/includes/odtphp/odf.php
```diff
@@ -161,12 +165,17 @@ public function setVars($key, $value, $encode = true, $charset = 'ISO-8859')
...
+		$value = html_entity_decode($value, ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401);
+
+		// fix breaklines...
+		$value = str_replace("<br>", "<br />", $value);
 		$convertedValue = $value;
 
 		// Check if the value includes html tags
 		if ($this->_hasHtmlTag($value) === true) {
+			// Note: allowing many tags is supported on PHP >7.4 . For older versions, this will strip all HTML tags.
+			$value = strip_tags($value, ['<br>', '<strong>', '<b>', '<i>', '<em>', '<u>', '<s>', '<sub>', '<sup>', '<span>']);
```
eldy, line 177: If instead of using an array parameter
['<br>', '<strong>', '<b>', '<i>', '<em>', '<u>', '<s>', '<sub>', '<sup>', '<span>']
you give the paramater as a string
'<br><strong><b><i><em><u><s><sub><sup><span>'
you will have same result for all php version. Giving array works only with PHP 7.4 but giving a string works on all version. Can you try this ?

thomas-Ngr: @eldy thanks for the hint ! I had a wrong understanding of the second parameter of strip_tags before php7.3. I added the changes.

jccs1953:
The problem still exists in the base install of  17.0.1 and it still sends html tags to the line when printing an ODT generated file ... I have moved a copy of the old odtphp folder ( 11.0.4 ) to the new version( 17.0.1 ) and this seems to have fixed the issue - use the file below to replace the folder ( make a backup of the current folder on the system or  rename it then upload this file in its place you can find it in the root of your install in the folder /includes
[odtphp.zip](https://github.com/Dolibarr/dolibarr/files/11212968/odtphp.zip)

Selection_188

This is a screenshot of the line generation using the Old file from 11.0.4

Thatoo:
The problem can't be fixed in v 17.0.1 as the fixe (https://github.com/Dolibarr/dolibarr/commit/fa0528a5d76294376d1b95fca1eeec30a21f020d) has apparently only been merged in develop branch 3 weeks ago and hasn't been merged in v 17 then.
Hopefully, it will been merged in v 17.0.2 or v 18.

PR #24340, opened 2023-03-24
FIX #24201 upload of external module fails to copy from incorrectly generated temp source dir

### LINKED ISSUE #24201: v17.0.0 upload of external module fails to copy from incorrectly generated temp source dir
### Bug

Since 17.0.0 it is not possible to install an external module using the provided form because the system fails to copy from an unexisting source path.

Not related to file permission, but bad source file path.

Capture d’écran de 2023-03-13 20-42-25

### Environment Version

17.0.0

### Environment OS

Linux

### Environment Web server

Apache

### Environment PHP

7.4, 8.1

### Environment Database

MySQL

### Environment URL(s)

/admin/modules.php

**Expected**

Uploading a custom module shall copy it to /htdocs/custom/ folder and install the module.

**Actual**

System is trying to copy a source folder that does not exzist, therefore breaking the installation of external module.
In detail, the system append a _htdocs_ folder in the temporary path of the uploaded module, that does not exist on disk, creating the error.

Capture d’écran de 2023-03-13 20-35-29

Upload any external module using the form in /admin/modules.php .

### Attached files

Capture d’écran de 2023-03-13 20-35-29

Capture d’écran de 2023-03-13 20-37-05

simnandez:


htdocs/admin/modules.php
```diff
@@ -230,13 +230,12 @@
 					// Now we install the module
 					if (!$error) {
 						@dol_delete_dir_recursive($dirins.'/'.$modulenameval); // delete the target directory
-						$submodulenamedir = $conf->admin->dir_temp.'/'.$tmpdir.'/htdocs/'.$modulenameval;
-						dol_syslog("We copy now directory ".$submodulenamedir." into target dir ".$dirins.'/'.$modulenameval);
-						$result = dolCopyDir($submodulenamedir, $dirins.'/'.$modulenameval, '0444', 1);
+						dol_syslog("We copy now directory ".$modulenamedir." into target dir ".$dirins.'/'.$modulenameval);
+						$result = dolCopyDir($modulenamedir, $dirins.'/'.$modulenameval, '0444', 1);
```
eldy, line 234: We are in a loop that process $modulenameval
`foreach ($modulenamearrays as $modulenameval) {`
If you use $modulenamedir, because it has always the same value, you always copy the same dir.
If metapackage.conf contains
a
b
c
we must copy dir a into a , then b into b, then c into c.
With your change, source always has value a

simnandez: You're right, I did not take into account the metapackages, the change has been made and tested.

PR #24346, opened 2023-03-27
FIX : Position field for events types was not working

atm-adrien:
## FIX : Position field for events types was not working

The "Position" field for events types was not working. Indeed, the order in event type lists was not the same as the order created by the position field.


htdocs/comm/action/class/cactioncomm.class.php
```diff
@@ -172,7 +172,7 @@ public function liste_array($active = '', $idorcode = 'id', $excludetype = '', $
 		if ($morefilter) {
 			$sql .= " AND ".$morefilter;
 		}
-		$sql .= " ORDER BY type, position, module";
+		$sql .= " ORDER BY position";
```
eldy, line 175: The position was introduced to allow to sort into combo list of types but this combo is grouped by type
image

This is why we have a group by type (to avoid to a a type into the wrong family)

atm-adrien:
@eldy I change my fix to create a title for the user inserted events types

htdocs/langs/en_US/commercial.lang
```diff
@@ -67,6 +67,7 @@ ActionAC_OTH=Other
 ActionAC_OTH_AUTO=Other auto
 ActionAC_MANUAL=Manually inserted events
 ActionAC_AUTO=Automatically inserted events
+ActionAC_USER=User inserted events
```
eldy, line 70: The "User inserted events" already exists: it is the "Manually inserted events". It meas the event was created by the user directly (standard calendar).
"Automatically inserted events" is events created by server for tracking purpose.

atm-adrien: I think we have to make a difference between "Manually inserted events" types which are created by the system and "User inserted events" types which are created by the user directly in the event type dictionary .

So I think it's important to have both "Manually inserted events" and "User inserted events" in the list of types

atm-thibaultf: @eldy the point is, when we want to add new type of event in the dictionary, we can only select "user" for the type. "Manually inserted event" have a type "system", that's the reason we are not able to manage the position of the new type created in the list of preexisting type of event. To do that, we suggest the three solutions below :
- Best option : Allow to select the type "system" when we add value in the dictionary. So the new value added is insert in the "Manually inserted event" and we can manage the position in this category.
- Keep the type "user" but the two types "user" and "system" should appear below the separator "Manually inserted events" and the position of the two types should be common
- Create a new separator for "Custom type event" for the type "user" in the dictionary. So new values added in the dictionary appear below this separator and we can manage position of them separately from the "Manually inserted events"

@eldy what do you think is the best option ?

eldy: You mention 
"Best option : Allow to select the type "system" when we add value in the dictionary. So the new value added is insert in the "Manually inserted event" and we can manage the position in this **category."

But if the entry is entered into dictionary by the end user, it means the event will be then inserted manually in calendar (because no code manage it). So in this cas the type must be "user".
And if there is a module that create automatically from a trigger or something else the entry, then the entry in dictionary must be entered during module activation. In such a case, the type can be "system".

Note: for module "managing event", the option 3 is used. But your module can choose option 1. It has no sense to enter into dictionary a type "system" if after, the event is recorded into agenda manually by end user.

atm-thibaultf: @eldy ok so we can do the second option : Keep the type "user" but the two types "user" and "system" should appear below the separator "Manually inserted events" and the position of the two types should be common
Is it ok for you @eldy ?

atm-quentin: @eldy

atm-thibaultf: @eldy

eldy: > @eldy ok so we can do the second option : Keep the type "user" but the two types "user" and "system" should appear below the separator "Manually inserted events" and the position of the two types should be common
> Is it ok for you @eldy ?

I still don't understand what's wrong. I do not experience any trouble on my side and behaviour is as expected.
Can you describe using screenshots the bug you try to fix ?

atm-thibaultf: The problem appears when you create a new entry in the event type dictionary. Let's me give an example below.

If I create 3 new types in the dictionary (Test_10, Test_23, Test_60) like this : 
image

In the type filter in the agenda list (action/list.php) I would like to have this 3 new types below the "Manually inserted event" and order in the good position. So normally, I should have this in the combo box filter :

- Manually inserted event
      - Test 10 (Position 10)
      - Appel téléphonique (Position 20)
      - Test 23 (Position 23)
      - Réception d'email (Position 50)
      - Test 60 (Position 60)
- Automatically inserted event
      - Autre auto (Position 210)
      - Autre (Position 220) 

But currently, I have this in the filter :
image

The problem is that the new types created appear below all the other independently of the position. We can see that only the event with the type "system" already existing in Dolibarr appear below the "Manually inserted event" separator. The new types created have the type "user" and seem to not be considered to appear in the "Manually inserted event" separator.

That's why we propose one of this 3 solutions :
- Solution 1 : Keep the type "user" but the two types "user" and "system" should appear below the separator "Manually inserted events" and the position of the two types should be common
- Solution 2 : Allow to select the type "system" when we add value in the dictionary. So the new value added is insert in the "Manually inserted event" and we can manage the position in this category.
- Solution 3 : Create a new separator for "Custom type event" or "User inserted events" for the type "user" in the dictionary. So new values added in the dictionary (with type "user" appear below this separator and we can manage position of them separately from the "Manually inserted events"

atm-thibaultf: @eldy

atm-thibaultf: @eldy

atm-maxime:
@atm-thibaultf @atm-adrien 
After presenting this to @eldy, we agreed that the solution 1 is the good one. The "Manually inserted events" section should display both "system" and "user" events type, ordered by position. So we'll have :

Manually inserted event
- Test 10 (Position 10)
- Appel téléphonique (Position 20)
- Test 23 (Position 23)
- Réception d'email (Position 50)
- Test 60 (Position 60)

Automatically inserted event
- Autre auto (Position 210)
- Autre (Position 220)

atm-adrien:
@eldy

PR #24347, opened 2023-03-27
FIX : Add bookmark with search fields that are arrays (backport 41572…

atm-gauthier:
…63cb898f1847cfcfc22dee6007c01b13a4d)


atm-gauthier:
@eldy this fix is present in v17 but missing in v15 and v16

PR #24348, opened 2023-03-27
NEW new feature (select prospect status)

KiVoilaJe:
Change of the way to modify the status of the prospect by replacing it with a combobox.

Creation of a selectStatus function which allows to create a select with the different status, which then makes an ajax request to allow to modify the status without reload the page. In this function, we can see an if ($mode) which allows to call the js script only once by calling the SelectStatus function with the parameter ($mode = js).


htdocs/core/ajax/ajaxstatusprospect.php
```diff
@@ -0,0 +1,70 @@
...
+$idstatus = GETPOST('id', 'int');
+$idprospect = GETPOST('prospectid', 'int');
+$action = GETPOST('action', 'aZ09');
+
+if ($action === "updatestatusprospect") {
+	$response = '';
+
+	$sql  = "UPDATE ".MAIN_DB_PREFIX."societe SET ";
+	$sql .= "fk_stcomm=".$db->escape($idstatus);
+	$sql .= " WHERE rowid = ".$db->escape($idprospect);
```
Hystepik, line 57: same as previous comment

htdocs/core/ajax/ajaxstatusprospect.php
```diff
@@ -0,0 +1,70 @@
...
+if (!defined('NOREQUIRESOC'))   define('NOREQUIRESOC', '1');
+// If there is no need to load and show top and left menu
+if (!defined("NOLOGIN")) {
+	define("NOLOGIN", '1');
+}
+
+// Load Dolibarr environment
+require '../../main.inc.php';
+
+
```
eldy, line 41: Missing 
```
// Security check
if ($user->socid > 0) {
	unset($action);
	$socid = $user->socid;
}
$result = restrictedArea($user, 'societe', $object->id, '&societe');

```

htdocs/core/ajax/ajaxstatusprospect.php
```diff
@@ -0,0 +1,70 @@
...
+}
+if (!defined('NOREQUIREMENU'))  define('NOREQUIREMENU', '1');
+if (!defined('NOREQUIREHTML'))  define('NOREQUIREHTML', '1');
+if (!defined('NOREQUIREAJAX'))  define('NOREQUIREAJAX', '1');
+if (!defined('NOREQUIRESOC'))   define('NOREQUIRESOC', '1');
+// If there is no need to load and show top and left menu
+if (!defined("NOLOGIN")) {
+	define("NOLOGIN", '1');
+}
+
```
eldy, line 37: Add recuperation of paramaetr here

htdocs/core/ajax/ajaxstatusprospect.php
```diff
@@ -0,0 +1,70 @@
...
+ * View
+ */
+
+top_httphead();
+
+$idstatus = GETPOST('id', 'int');
+$idprospect = GETPOST('prospectid', 'int');
+$action = GETPOST('action', 'aZ09');
+
+if ($action === "updatestatusprospect") {
```
eldy, line 52: if ($action === "updatestatusprospect" && $permisstiontoupdate) {

eldy:
Add screenshot...

FliyFly:
Here is an example of the select

KiVoilaJe:
Here is an example of the selct: 
Capture d’écran du 2023-05-09 18-38-45

Hystepik:
@eldy ok for me

PR #24352, opened 2023-03-28
FIX #24352 Wrong emailing delay in prompt in mailing-send.php

battosai30:
# FIX|Fix #[*24352 Wrong emailing delay in prompt*]

The delay between email announced in prompt is not right : when the delay is called, usleep() takes useconds as argument and MAILING_DELAY is multiplied by 1000000, but prompt information doesn't have to as it's admitted that is seconds

### DIFF
--- a/scripts/emailings/mailing-send.php
+++ b/scripts/emailings/mailing-send.php
@@ -87,7 +87,7 @@ if (!isModEnabled('mailing')) {
 print "***** ".$script_file." (".$version.") pid=".dol_getmypid()." *****\n";
 
 if (!empty($conf->global->MAILING_DELAY)) {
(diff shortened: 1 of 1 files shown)


eldy:
I don't see trouble.
The MAILING_DELAY is the number of second (between 10 and 0.001).
But usleep need usecond. So we need to multiplicate by 1000000

battosai30:
usleep needs usec. MAILING_DELAY  is second.
 So if you `echo 'delay between email is ' . MAILING_DELAY * 100000 . ' seconds' ` it's wrong because you set for example MAILING_DELAY=60 you get "delay between email is 6000000 seconds" instead of "delay between email is 60 seconds"

This fix doesn't correct usleep() part, just the message prompt.

eldy:
Thanks. i see. Trouble and fix was just into the log message not into executed code (i was focused on this).

PR #24360, opened 2023-03-28
Fix #24256: default values cannot be set for some forms due to inconsistent action names 

### LINKED ISSUE #24256: Enhancement for product variants
### Feature Request

Hi Contributors,

Upon creation of the product variants, you can select multiple variants at one run however when hitting Create it only considers the last chosen Variant attribute. It would be so much more quick and convenient if the system would consider all of the selected attributes and would create them all together.

Thanks a lot!

image
 

### Use case

Less time consuming when many combinations must be created.

### Suggested implementation

_No response_

### Suggested steps

_No response_

priojk:
As described in more detail in the issue: Default values for forms are only applied, if the name of the action starts with 'create'.

However, action names have gotten a bit inconsistent:
- The form for creating a new buyer price for a product uses the action '**add**_price'
- The form to confirm an order incl. selection of order method uses the action '**make**order'

Changing the action names in a consistent way might break other parts.

Therefore, my proposal to extend the pattern matching to also cover alternative names for 'create'.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -20,6 +20,7 @@
  * Copyright (C) 2022       Anthony Berton	         	<anthony.berton@bb2a.fr>
  * Copyright (C) 2022       Ferran Marcet           	<fmarcet@2byte.es>
  * Copyright (C) 2022       Charlene Benke           	<charlene@patas-monkey.com>
+ * Copyright (C) 2023		Joachim Kueter				<git-jk@bloxera.com>
  *
  * This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


priojk:
Travis CI Error seems unrelated to the changes in the PR:

PHP Fatal error:  Uncaught Error: **Failed opening required '/home/travis/build/Dolibarr/dolibarr/test/phpunit/WebsiteTest.php**' (include_path='.:/home/travis/.phpenv/versions/8.1.12/share/pear') in home/travis/build/Dolibarr/dolibarr/test/phpunit/AllTests.php:279

priojk:
Should have gone to branch 17.0

PR #24361, opened 2023-03-28
FIX : No usage of the function updateProduction in the update function

atm-adrien:
## FIX : No usage of the function updateProduction in the update function

This function deletes the lines contained in the MO concerned to then add them according to the lines of the associated BOM.
I can't really see the point?

Moreover, when I create an MO without any linked BOM, add lines and then modify data of the MO concerned, it deletes all the lines that I have previously created. This is due to the use of the updateProduction function in the MO update.

### DIFF
--- a/htdocs/mrp/class/mo.class.php
+++ b/htdocs/mrp/class/mo.class.php
@@ -638,21 +638,21 @@ class Mo extends CommonObject
 
 		$oldQty = $this->oldQty;
 		$newQty = $this->qty;
-		if($newQty != $oldQty) {
+		if ($newQty != $oldQty) {
 			$sql = 'SELECT rowid FROM ' . MAIN_DB_PREFIX . 'mrp_production WHERE fk_mo = ' . $this->id;
 			$resql = $this->db->query($sql);
-			if($resql) {
+			if ($resql) {
 				while ($obj = $this->db->fetch_object($resql)) {
 					$moLine = new MoLine($this->db);
 					$res = $moLine->fetch($obj->rowid);
-					if(!$res) $error++;
+					if (!$res) $error++;
 
-					if($moLine->role == 'toconsume' || $moLine->role == 'toproduce') {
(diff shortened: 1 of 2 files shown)


htdocs/mrp/class/mo.class.php
```diff
@@ -612,11 +612,6 @@ public function update(User $user, $notrigger = false)
 			$error++;
 		}
 
-		$result = $this->updateProduction($user, $notrigger);
```
eldy, line 615: I think usage of this method here is done on purpose.
However, inside this method it seems to miss code under the comment 

// TODO Check that production has not started. If yes, we stop here.

There is nothing in this TODO. We call the method when we are in draft mode, but we should do nothing if production has started.

atm-adrien: I admit that I don't understand the need to delete the MO lines to add them later (without the changes made before the MO update).
Indeed, I understand the use of an update function for MO lines: It is possible to modify the quantity of an OF in "Draft" status, its lines must therefore be modified accordingly. However, this does not require deleting the rows and then adding them again.

Moreover, as I said above, when I create an MO without an associated BOM, add lines to it and then modify the MO (the label for example), all the lines previously entered are deleted. And it has nothing to do with whether production has started or not. This is due to the fact that we delete all the lines of the MO to add that of the associated BOM, but what happens to MOs that do not have associated BOMs?

I think the solution is to use a real MO rows update function (like it's done for all other objects in Dolibarr)

I'm developing a fix to correct that

atm-adrien:
Travis say that there is an error, but the showed error does not correspond to my feature :
image

atm-adrien:
@eldy

atm-adrien:
Hi @eldy ! 

We really need to fix this bug for version 17, preferably before version 18 is released.

Do you have the possibility to merge this PR?

Please let me know if there is any problem.

Issue #24356, opened 2023-03-28
Bug: default values cannot be set for some forms due to inconsistent action names


priojk:
### Bug

In the form to add a new supplier price to a product (product/fournisseurs.php?action=add_price) it is not possible to set default values for entry fields with the standard mechanisms.

The reason is that default values will only be applied to forms that have an action containing either 'create' or 'presend' (see function GETPOST in functions.lib.php).

Unfortunately, the action to create a new supplier price is called add_price instead of create_price, thus not containing create and therefore disabling the default values for this form, although it has all mechanisms included to apply default values.

There is a similar problem in the form fourn/commande/card.php?action=makeorder to confirm a new supplier order. It would e.g. be helpful, if a default value for the order method could be defined for users always choosing the same method. However, this has no effect, since the action here is called makeorder instead of create_order.

### Environment Version
 v17

### Options for fixing the bug

In order to fix this bug, we need to decide, if we either change the action names in a consistent way. This might possibly break some hooks etc. but would be the cleanest approach. 

Alternatively, we could change the check within GETPOST and add more flexibility by allowing create/add/make and maybe some more terms in an action to provide default values.

priojk:
Closed by #24369 (develop) and #24363 (v17)

PR #24363, opened 2023-03-28
Fix #24256: default values cannot be set for some forms due to inconsistent action names 

### LINKED ISSUE #24256: Enhancement for product variants

priojk:
As described in more detail in the [issue #24256](https://github.com/Dolibarr/dolibarr/issues/24356): Default values for forms are only applied, if the name of the action starts with 'create'.

However, action names have gotten a bit inconsistent:

- The form for creating a new buyer price for a product uses the action '**add**_price'
- The form to confirm an order incl. selection of order method uses the action '**make**order'

Changing the action names in a consistent way might break other parts.

Therefore, my proposal to extend the pattern matching to also cover alternative names for 'create'.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -588,7 +588,7 @@ function GETPOST($paramname, $check = 'alphanohtml', $method = 0, $filter = null
 				}
 			}
 			if (!empty($conf->global->MAIN_ENABLE_DEFAULT_VALUES)) {
-				if (!empty($_GET['action']) && (preg_match('/^create/', $_GET['action']) || preg_match('/^presend/', $_GET['action'])) && !isset($_GET[$paramname]) && !isset($_POST[$paramname])) {
+				if (!empty($_GET['action']) && (preg_match('/^create|^add|^make/', $_GET['action']) || preg_match('/^presend/', $_GET['action'])) && !isset($_GET[$paramname]) && !isset($_POST[$paramname])) {
 					// Now search in setup to overwrite default values
 					if (!empty($user->default_values)) {		// $user->default_values defined from menu 'Setup - Default values'
 						if (isset($user->default_values[$relativepathstring]['createform'])) {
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -588,7 +588,7 @@ function GETPOST($paramname, $check = 'alphanohtml', $method = 0, $filter = null
 				}
 			}
 			if (!empty($conf->global->MAIN_ENABLE_DEFAULT_VALUES)) {
-				if (!empty($_GET['action']) && (preg_match('/^create/', $_GET['action']) || preg_match('/^presend/', $_GET['action'])) && !isset($_GET[$paramname]) && !isset($_POST[$paramname])) {
+				if (!empty($_GET['action']) && (preg_match('/^create|^add|^make/', $_GET['action']) || preg_match('/^presend/', $_GET['action'])) && !isset($_GET[$paramname]) && !isset($_POST[$paramname])) {
```
eldy, line 591: Adding the ^add will bring bad side effect because 'add( is action after posting a form shown with create. And default values must be triggered only on the create page. After a post of add, the value to keep is the value entered by user on form. May be we can add ^add_price
Adding ^make seems ok.

However, even if renaming action to have consistency may have effect, for a long term goal i think we should take the risk to do this into "develop" branch. Moving to consistency is a long road action and this is a new step we should dare into develop...

priojk: Very good point about the impact on the 'add' action - I missed that.

I now changed it to add_price as a short term solution for v17 and will create another patch for develop to change to create_price instead.

PR #24368, opened 2023-03-29
Fix naming of create action for adding buying prices

priojk:
This is related and a follow up to PR #24363 

The mechanism for inserting default values into forms can only work, if the action to create a form starts with 'create'. Some forms do not follow this rule and therefore defined default values for these forms do not get inserted.

This fix changes the wording of the action for the create form to add new product buying prices of suppliers and reverts the related temporary fix for v17 (allowing add_price for the action).


priojk:
Needed to rebase.

PR #24380, opened 2023-03-29
forceduedate is an option of the function

rycks:
# FIX forceduedate is a function arg

That's so strange to get that here because $forceduedate is an argument of the function and some lines after there is a if(empty($forceduedate .../...


htdocs/compta/facture/class/facture.class.php
rycks, line 1: please have a look line 582 :

		// Define due date if not already defined
		if (empty($forceduedate)) {
			$duedate = $this->calculate_date_lim_reglement();

PR #24385, opened 2023-03-29
fix substitution for online sign contract

frederic34:


julbroa:
I have added the following lines as an experiment;
if (is_object($object) && $object->element == 'contrat') {
					require_once DOL_DOCUMENT_ROOT.'/core/lib/signature.lib.php';
					$substitutionarray['__ONLINE_SIGN_CONTRACT_URL__'] = getOnlineSignatureUrl(0, 'contrat', $object->ref);

but this is the result;
image

Jozmichne:
for some reason I cannot get it to work, I am on version 17.0.1 and it looks like the corrections that you have listed above are already implemented but its still broken.

julbroa:
> for some reason I cannot get it to work, I am on version 17.0.1 and it looks like the corrections that you have listed above are already implemented but its still broken.

no, my corrections is bad, the correct correction is that: https://github.com/Dolibarr/dolibarr/commit/5494b70ce062e8b9b081bc82fce759246d8544c9

Jozmichne:
> > for some reason I cannot get it to work, I am on version 17.0.1 and it looks like the corrections that you have listed above are already implemented but its still broken.
> 
> no, my corrections is bad, the correct correction is that: [5494b70](https://github.com/Dolibarr/dolibarr/commit/5494b70ce062e8b9b081bc82fce759246d8544c9)

Sorry for the delayed response, I actually found the file and downloaded it then replaced it into my installation and it worked, thank you for your reply.. the first time around I just copied the corrections listed in the solution above and for some reason they weren't applied correctly. I am new to php so I probably missed something while applying the adjustments..

PR #24386, opened 2023-03-30
NEW: Add global PRODUCT_DONOTSHOW_PRODUCTWITHNOSUPPLIERPRICE

kkhelifa-opendsi:
NEW: Add global PRODUCT_DONOTSHOW_PRODUCTWITHNOSUPPLIERPRICE for do not show the product without supplier price instead of show all product and disabled these products

Case when $alsoproductwithnosupplierprice is equal at 0 for the select_produits_fournisseurs_list() function

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -3355,6 +3355,9 @@ class Form
 			$sql .= " LEFT JOIN " . $this->db->prefix() . "c_units u ON u.rowid = p.fk_unit";
 		}
 		$sql .= " WHERE p.entity IN (" . getEntity('product') . ")";
+		if (empty($alsoproductwithnosupplierprice) && !empty($conf->global->PRODUCT_DONOTSHOW_PRODUCTWITHNOSUPPLIERPRICE)) {
+			$sql .= " AND pfp.rowid IS NOT NULL";
+		}
 		if ($statut != -1) {
 			$sql .= " AND p.tobuy = " . ((int) $statut);
 		}


htdocs/core/class/html.form.class.php
```diff
@@ -3355,6 +3355,9 @@ public function select_produits_fournisseurs_list($socid, $selected = '', $htmln
 			$sql .= " LEFT JOIN " . $this->db->prefix() . "c_units u ON u.rowid = p.fk_unit";
 		}
 		$sql .= " WHERE p.entity IN (" . getEntity('product') . ")";
+		if (empty($alsoproductwithnosupplierprice) && !empty($conf->global->PRODUCT_DONOTSHOW_PRODUCTWITHNOSUPPLIERPRICE)) {
```
eldy, line 3358: Can you have a look at option 
SUPPLIER_INVOICE_WITH_PREDEFINED_PRICES_ONLY, 
SUPPLIER_ORDER_WITH_PREDEFINED_PRICES_ONLY
SUPPLIER_PROPOSAL_WITH_PREDEFINED_PRICES_ONLY
it seems to do the same but is more accurate because can be set differently for proposal and invoice

kkhelifa-opendsi: Yes when you set these global you have all the products in the select with all products who have not a predefined price disabled
So this option dont get the product who have not a predefined price if these options is set : SUPPLIER_INVOICE_WITH_PREDEFINED_PRICES_ONLY,
SUPPLIER_ORDER_WITH_PREDEFINED_PRICES_ONLY,
SUPPLIER_PROPOSAL_WITH_PREDEFINED_PRICES_ONLY
Maybe my option is superflu and we dont show these product if these options is set so i can remove " && !empty($conf->global->PRODUCT_DONOTSHOW_PRODUCTWITHNOSUPPLIERPRICE)"

eldy: Sorry i did not understood your conclusion. Do you still need the PR ?

kkhelifa-opendsi: Yes, I still need it :

- using SUPLPIER_INVOICE_XXX the lines are **disabled** if there is no supplier price
- using PRODUCT_DONOTSHOW_PRODUCTWITHNOSUPPLIERPRICE the lines are **not shown** if there is no supplier price

eldy: > * SUPLPIER_INVOICE_XXX

So instead of introducing a new option, it would be better to play with the value of SUPPLIER_XXX_WITH_PREDEFINED_PRICES_ONLY,

value 1 : lines are disabled and a there is a test added on minimum qty and disable the autofill of qty with the minimum.
value 2 : the default behaviour
value 3: your behaviour, lines are removed (not visible).

eldy: @kkhelifa-opendsi If you still need this change, can you implement using the already existing option
SUPPLIER_INVOICE_WITH_PREDEFINED_PRICES_ONLY = 1
or
SUPPLIER_ORDER_WITH_PREDEFINED_PRICES_ONLY = 1
...

eldy:
@kkhelifa-opendsi If you still need this change, can you implement using the already existing option
SUPPLIER_INVOICE_WITH_PREDEFINED_PRICES_ONLY = 1
or
SUPPLIER_ORDER_WITH_PREDEFINED_PRICES_ONLY = 1
...

eldy:

PR #24387, opened 2023-03-30
NEW: Add option on the MRP module for set by default the remaining quantities to be consumed

kkhelifa-opendsi:
Add option on the MRP module for set by default the remaining quantities to be consumed when you click on the button "Consume or Produce"


eldy:
Don't understand the goal of this PR. We already have 2 button on the MRP.
One to have the value autofill with 0
And the other one to have the value autofilled with the remaining quantity.
Why not just using the second button instead of adding an option so the first button do the same thing than the second one (did i miss somethiing) ?

kkhelifa-opendsi:
Because we want the possibility of having all the preset quantities to be consumed without the preset quantities to be produced
It's only a options because it's depend of the preference of the customer
This avoids that the customer has to enter the quantity consumed of 100 lines of products for example in this case

eldy:
Can you provide a use case where all products are set as consumed but no production is done ? (exemple)
This happen often when consumption is not completed (because final product is not finished), but seems uncommon when all consumption is completed. There is no step after, into current process, what is next step in your current process once all product are consumed and production is not yet finished. I imagine that the end user have to re-click on "production" to enter the produced quantity ? Is it done on a next step ? next day ?

kkhelifa-opendsi:
In the case we have, there are two actors that make the production :
- the first **prepares** the materials (consumes the products but do not produce the new item)
- the second **builds** (produces) the new product using the materials prepared by the first one

erikvanberkum:
> In the case we have, there are two actors that make the production :
> 
> - the first **prepares** the materials (consumes the products but do not produce the new item)
> 
> - the second **builds** (produces) the new product using the materials prepared by the first one 

Nice option that a product to produce is in 2 stages:

1. Produceable 
2. Produced

eldy:
Due to conflit i can merge online, i added the option manually.
I did not add the setup page because i think we can do better now to fight against the feature available by setup only. 

In v20+, we can now group several different actions in the same button, so we could imagine to replace the button "Consume or produce" into a combo button with the 3 choices:

"Consume or product"
"Consume only"
"Produce only"

the first choice is same as now, the second is the feature you implemented with your option, the third one will be similar to the one you did but on the production side (only production qty is prefilled).

We can keep the button "Consume and produce All" as a separated button because it is a button used by a lot of use case that does not need production in two steps to save clicks. We can also have it into the grouped button.

PR #24389, opened 2023-03-30
FIX: Fix compare on float when test on stock dont allow negative transfert

kkhelifa-opendsi:
Fix compare on float when test on stock dont allow negative transfert (case when a stock is show at 2.20 but equal to 2.199999 in memory)


htdocs/product/stock/class/mouvementstock.class.php
```diff
@@ -419,7 +419,7 @@ public function _create($user, $fk_product, $entrepot_id, $qty, $type, $price =
 					return -8;
 				}
 			} else {
-				if (empty($product->stock_warehouse[$entrepot_id]->real) || $product->stock_warehouse[$entrepot_id]->real < abs($qty)) {
+				if (empty($product->stock_warehouse[$entrepot_id]->real) || price2num($product->stock_warehouse[$entrepot_id]->real) < abs($qty)) {
```
eldy, line 422: $product->stock_warehouse[$entrepot_id]->real)  is a data in memory, so it should already be numeric.
The price2num should be move higher when filling the variable.

kkhelifa-opendsi: you get a float type for $product->stock_warehouse[$entrepot_id]->real so in memory you can have 2.1999999999999999999999999999999999 even if it's equal to 2.2 in reality it's a common problem in php when you compare float/double.
Maybe using price2num() is not the correct function but it's work well

eldy: If you have 2.199999999999999999 in memory instead of 2.2, it means you didn't use the price2num with parameter 'MS' or 'MT' when you retreive the data from the end user to store it in memory. The price2num must be use when you convert the data coming from user or to fix a calculation.
I confirm that the price2cum must be used when the variable $product->stock_warehouse[$entrepot_id]->real is set and not when the variable is consumed.

kkhelifa-opendsi: It's done.
I move the price2num() when the variable $product->stock_warehouse[$entrepot_id]->real is set

PR #24391, opened 2023-03-30
fix display price supplier list

frederic34:
BEFORE
image

AFTER
image

first column is date of creation
tms column is date of last price


htdocs/product/fournisseurs.php
```diff
@@ -954,7 +954,7 @@ function update_price_from_multicurrency() {
 					'pfp.barcode'=>array('label'=>$langs->trans("BarcodeValue"), 'enabled' => isModEnabled('barcode'), 'checked'=>0, 'position'=>16),
 					'pfp.packaging'=>array('label'=>$langs->trans("PackagingForThisProduct"), 'enabled' => getDolGlobalInt('PRODUCT_USE_SUPPLIER_PACKAGING'), 'checked'=>0, 'position'=>17),
 					'pfp.status'=>array('label'=>$langs->trans("Status"), 'enabled' => 1, 'checked'=>0, 'position'=>40),
-					'pfp.tms'=>array('label'=>$langs->trans("DateModification"), 'enabled' => isModEnabled('barcode'), 'checked'=>1, 'position'=>50),
+					'pfp.tms'=>array('label'=>$langs->trans("AppliedPricesFrom"), 'enabled' => isModEnabled('barcode'), 'checked'=>1, 'position'=>50),
```
eldy, line 957: The AppliedPricesFrom means from when the new price entry is used. And it is used as soon as the entry is created into databases. If we update the line (for example the barcode or availability), this should not change the AppliedPricesFrom. If entry as created the 1st of january, even if we have several changes into the line, the price ill have effect from the 1st of january.

PR #24397, opened 2023-03-30
Improve margin calculation by updating buy price

SylvainLegrand:
In most cases, users use the cost price to calculate the margin from the best supplier price.
So, when they clone a proposal, they want to update the selling price and the buying price if the supplier's price list changes.
In this case, you can have a quick and precise view of the margin regardless of changes in customer or supplier prices.

### DIFF
--- a/htdocs/comm/propal/class/propal.class.php
+++ b/htdocs/comm/propal/class/propal.class.php
@@ -1455,6 +1455,12 @@ class Propal extends CommonObject
 							$line->subprice = $pu_ht;
 							$line->tva_tx = $tva_tx;
 							$line->remise_percent = $remise_percent;
+							// Update also cost price
+							$subpricetmp = isset($conf->global->ForceBuyingPriceIfNull) && $conf->global->ForceBuyingPriceIfNull > 0 ? 0 : $line->subprice;
+							$result = $object->defineBuyPrice($subpricetmp, $line->remise_percent, $line->fk_product);
+							if ($result > 0) {
+								$line->pa_ht = $result;
+							}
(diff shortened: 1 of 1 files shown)


htdocs/comm/propal/class/propal.class.php
```diff
@@ -1455,6 +1455,12 @@ public function createFromClone(User $user, $socid = 0, $forceentity = null, $up
 							$line->subprice = $pu_ht;
 							$line->tva_tx = $tva_tx;
 							$line->remise_percent = $remise_percent;
+							// Update also cost price
+							$subpricetmp = isset($conf->global->ForceBuyingPriceIfNull) && $conf->global->ForceBuyingPriceIfNull > 0 ? 0 : $line->subprice;
+							$result = $object->defineBuyPrice($subpricetmp, $line->remise_percent, $line->fk_product);
```
eldy, line 1460: We should never have data that is silently updated into database without the consent of the user.

htdocs/comm/propal/class/propal.class.php
```diff
@@ -1455,6 +1455,12 @@ public function createFromClone(User $user, $socid = 0, $forceentity = null, $up
 							$line->subprice = $pu_ht;
 							$line->tva_tx = $tva_tx;
 							$line->remise_percent = $remise_percent;
+							// Update also cost price
```
eldy, line 1458: When we make a clone, the data of the new object must match the value of the src.

eldy:
Feature seems implemented with #24397 in a better way, so i close this one.

PR #24417, opened 2023-04-02
FIX Force constant ACCOUNTING_REEXPORT to value 0 by default to avoid confusion

aspangaro:
See this post forum, [FR] https://www.dolibarr.fr/forum/t/comptabilite-export-attention-sur-la-v17/42845

### DIFF
--- a/htdocs/install/mysql/migration/16.0.0-17.0.0.sql
+++ b/htdocs/install/mysql/migration/16.0.0-17.0.0.sql
@@ -403,4 +403,5 @@ ALTER TABLE llx_projet ADD COLUMN extraparams varchar(255);
 
 DELETE FROM llx_const WHERE name = 'TICKET_CREATE_THIRD_PARTY_WITH_CONTACT_IF_NOT_EXIST';
 
-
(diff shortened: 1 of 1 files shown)


eldy:
This is a filter to defined the content of the list.
And a filter must never be enabled by default. This is a look and feel behaviour of Dolibarr. Filters that restrict information on screen must be applied by the end user, on purpose.
Having a filter enabled by default means some data are hidden and just disappears. This generates a lot of ticket "Help, my data has disappeared !", "Can you restore my database, the data were lost", ...   This is why in Dolibarr we never hide data without the consent of end user. This is the end user that decides to hide or not information into its list using the filters when on a list. Once a choice has been done a first time, the filter remains, but at least, it is the user that had decided to hide data and not the machine making behaviour consistent everywhere.

aspangaro:
See #24438 for better integration.

aspangaro:
@eldy 

In view of what has been done on v17, the first feedbacks are mixed... 

Nevertheless, everyone agrees on the fact that it is necessary to separate the export from the journals page and the second PR goes in this direction.

Please accept PR #24438 for better integration

The v17 will be a special case...

PR #24426, opened 2023-04-03
FIX Propagate extrafields from supplier order to reception

aspangaro:
Wrong object source. $recept = $objectsrc

Thanks @lvessiller-opendsi


lvessiller-opendsi:
You're welcome @aspangaro 
$object is not use in create mode and "extrafields" were loaded in $recept variable and not in $object

PR #24428, opened 2023-04-03
Limit the list of invoice lines to link in accounting

comaiteseb:
FIX
Limit the list of invoice lines to be linked in accounting to lines whose quantity and amount are greater than 0


aspangaro:
Hello,

Three points :
- Why act on quantities? Only "l.total_ht<>0" is enough
- And in other page ? (Customer, Supplier, Expense report) The index page will return a number of operations waiting to be processed even if it is zero because it tries to link operations that have no indication on l.fk_code_ventilation
- This is not a bug, if it is accepted, it should be proposed this improvement on the branch develop

PR #24438, opened 2023-04-05
NEW: Accountancy - Add specific page to export accounting data

aspangaro:
.... rather than the journals page

To avoid misunderstandings, the accounting data visualization function is separated from the export function. 

1/ The journals page resumes its original function of displaying the accounting data without the possibility of direct export. Similarly, the export dates are no longer displayed by default.

So we have a General Ledger view and a Journals view like in other software of this type

Example:
image

2/ The export page has the same structure as before but with a clear message of why the data is no longer displayed when it is noted as exported.

a/ By default, when you arrive on the page

image

b/ With button reexport activated

image

This will be clearer and more understandable as the export was not possible from the general ledger or the balance.


aspangaro:
Travis return a build construct problem. Not my code...

aspangaro:
Please @eldy 

All the feedback from users, customers are positives. Everyone agrees that it is the best solution (see the forum) to dedicate a page to the export of the accounting and to restore the journal page as it is expected rather than the solution that you have proposed in v17. 

See this post forum, [FR] https://www.dolibarr.fr/forum/t/comptabilite-export-attention-sur-la-v17/42845

Thank you in advance for taking this PR into consideration

PR #24441, opened 2023-04-05
FIX : Disable the product selection list 

atm-adrien:
## FIX : Disable the product selection list 

On the MO creation screen :
Users can select a BOM (it will set the linked product on the product selection list) and then changge the linked product even if that does not correspond to the selected BOM.

I think we must control that, and make impossible product change when the BOM is set

### DIFF
--- a/htdocs/mrp/mo_card.php
+++ b/htdocs/mrp/mo_card.php
@@ -284,6 +284,7 @@ if ($action == 'create') {
 	?>
 	<script>
 		 $(document).ready(function () {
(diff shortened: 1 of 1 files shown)


atm-adrien:
Travis say that there is an error, but the showed error does not correspond to my feature : 
image

eldy:
There is a use case that need to be able to change product.
Imagine you have variant products (red bike, blue bike, green bike,...). To avoid to create n bom for each variant, we create a bom for the parent product (so the bike).
Then we create mo and we change the product to choose the good product.
I think we can find other use cases that need to be able to modify generated product.

PR #24443, opened 2023-04-05
NEW Send a notification email when ticket assigned

BenjaminFlr:
NEW
Introducing a new constant TICKET_NOTIFY_CUSTOMER_TICKET_ASSIGNED to send an email to the thirdparty contacts when the ticket is assigned to somebody and wasn't assigned before, in order to prevent the customer from receiving an email for each assigned change.


BenjaminFlr:
Hi @eldy , what do you think about this feature?

PR #24456, opened 2023-04-06
Select Company in create form of a fichinter : must display only clients

atm-lena:
# FIX|Fix #

Select Company in create form of a fichinter : must display only clients


eldy:
Why only clients ?
We can also make intervention at supplier or at prospect not yet client ( for example a visit to transform them as client)

PR #24461, opened 2023-04-07
fix : Warning: Undefined array key other_linked_objects in /home/http…

grandoc:
…d/vhosts/aflac.fr/domains/dev.aflac.fr/httpdocs/compta/facture/card.php on line 1470


htdocs/compta/facture/card-rec.php
```diff
@@ -1186,6 +1187,7 @@
 		}
 
 		// Call Hook formConfirm
+		$formconfirm = '';
```
eldy, line 1190: You erase the formconfirm that is set few lines before

grandoc:
But it was commented, and so make an error

htdocs/compta/facture/card-rec.php
grandoc: I have fixed it in an other pull request

eldy: So i close this PR

PR #24462, opened 2023-04-07
FIX : We should always be able to filter on action type no matter if …

atm-gauthier:
…we can't read actions of others

### DIFF
--- a/htdocs/core/lib/agenda.lib.php
+++ b/htdocs/core/lib/agenda.lib.php
@@ -69,18 +69,17 @@ function print_actions_filter($form, $canedit, $status, $year, $month, $day, $sh
 	}
 	print '<input type="hidden" name="search_showbirthday" value="'.((int) $showbirthday).'">';
 
+	print '<div class="divsearchfield">';
+	// Type
+	$multiselect = 0;
+	if (!empty($conf->global->MAIN_ENABLE_MULTISELECT_TYPE)) {     // We use an option here because it adds bugs when used on agenda page "peruser" and "list"
+		$multiselect = (!empty($conf->global->AGENDA_USE_EVENT_TYPE));
+	}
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/agenda.lib.php
```diff
@@ -69,18 +69,17 @@ function print_actions_filter($form, $canedit, $status, $year, $month, $day, $sh
...
+	// Type
+	$multiselect = 0;
+	if (!empty($conf->global->MAIN_ENABLE_MULTISELECT_TYPE)) {     // We use an option here because it adds bugs when used on agenda page "peruser" and "list"
+		$multiselect = (!empty($conf->global->AGENDA_USE_EVENT_TYPE));
+	}
+	print img_picto($langs->trans("ActionType"), 'square', 'class="fawidth30 inline-block" style="color: #ddd;"');
+	print $formactions->select_type_actions($actioncode, "search_actioncode", $excludetype, (empty($conf->global->AGENDA_USE_EVENT_TYPE) ? 1 : -1), 0, $multiselect, 0, 'maxwidth500');
+	print '</div>';
 
+	if ($canedit) {
```
eldy, line 82: The test $canedit means  can edit the filter fields (if unset, filters are forced and not modifed).
If you move the code of filter out of this if, this option will not work anymore.

atm-gauthier: I need this fix for actioncomm list.
Currently if we don't have permission to see actions of others we can't filter on a specific type, for me it makes no sens, then i moved the "if ($canedit)" later because i believe type filter should always be available

atm-adrien: @eldy

atm-quentin:
@eldy

atm-thibaultf:
@eldy

PR #24477, opened 2023-04-10
FIX: Societe List - More explicite translation on label of ProfId

aspangaro:
Could we translate the idprof into something understandable ?

Before

image

After

image


eldy:
When you are on a list of companies,the meaning of each column depends on each line.
For first line, the column Idprof1 is "SIREN" because country of thirdparty is france.
On next line, the Idprof1 will we "Prof number" because country is belgium.
etc...
This is the reason why the label is only provided in the tooltip after "Id Prof 1"

PR #24481, opened 2023-04-11
FIX: action buttons: reset params for every button

marc-dll:
When the first buttons have a title (for example, modifiy button on an invoice when the invoice is dispatched into accountancy), it is not reset for every button.


eldy:
Is this still a bug on v17 ?
I can't reproduce or may be i did not understand the trouble. Can you explain with a screenshot ?

marc-dll:
@eldy : when a button generated using `dolGetButtonAction()` is forbidden with a help text on hover, the following buttons also have the same hovering text. This is because variable `$params` is declared once and never reset. It's also the case in v17.

On the following screenshot, the pointer is over the "Saisir le paiement" button.

Capture d’écran de 2023-04-24 12-25-17

Sorry if it wasn't clearer the first time.

Issue #23076, opened 2022-12-01
Extend Dolibarr by adding more invoice types and more masks-series.


sonikf:
### Feature Request

Extend Dolibarr by adding more masks-series and more invoice types to help boost the adoption of Dolibarr in Greece

### Use case

For more than a year it is compulsory for everyone issuing an invoice to upload it to the Greek finance ministry’s database called Mydata. Had not been for a community member who stepped up and built a module, Dolibarr would have become nearly useless for Greece.
It’s also compulsory to create a different type of invoice(with different name and mask-series) for each use case. So far there are 14 types of invoices, 3 types of credit notes and 3 other types for specific use cases, total is 20. Real madness! Wait there is more, each type can have infinite number of masks!  
So far we are barely coping by: 
a. Using the {tttt} method which is not obvious for a new user, it does not support Greek letters, it’s hackish and counter-intuitive for our use case and negates the professional feel of Dolibarr.
b. Renaming the deposit invoice which is really hackish for obvious reasons.
See screen shot from the invoice type and mask-series setup page of the Greek government invoicing app designed to support sme’s and freelancers.
![1](https://user-images.githubusercontent.com/93765174/205138733-d04dd095-8d9a-4290-a0af-19007c6c8e38.png)

For little over a month now I’m testing mercure with  3 extra masks and invoice types with minor code changes and no bugs so far(see screen shots), which is a half measure by itself because I’ve only added the 3 other major types needed and it doesn’t solve the mask-series predicament.
Although my experience in php is basic I'm willing to put in the work but your help and guidance is needed
Thanks for all the great work!
![2](https://user-images.githubusercontent.com/93765174/205145651-7f4654a5-a782-47b1-a571-048efee511f3.png)
![3](https://user-images.githubusercontent.com/93765174/205145667-35da42a3-7bfb-4621-9b0c-8cf5c44d8189.png)

PR #24483, opened 2023-04-11
The deposit and replacement invoice types are not allowed in Greece.

sonikf:


htdocs/master.inc.php
```diff
@@ -233,7 +233,14 @@ function is_countable($c)
...
 		$conf->global->INVOICE_DISABLE_REPLACEMENT = 1;
 	}
-
+	if ($mysoc->country_code == 'GR' && !isset($conf->global->INVOICE_DISABLE_REPLACEMENT)) {
+		// The replacement invoice type is not allowed in Greece.
+		$conf->global->INVOICE_DISABLE_REPLACEMENT = 1;
+	}
+	if ($mysoc->country_code == 'GR' && !isset($conf->global->INVOICE_DISABLE_DEPOSIT)) {
+		// The deposit invoice type is not allowed in Greece.
+		$conf->global->INVOICE_DISABLE_DEPOSIT = 1;
```
eldy, line 242: The replacement invoice is not allowed into a lof of country.
However it is more strange for deposit invoice (also known as down payment).
When a company receive a down payment, before the final amount is know. What is the process in greece. No invoice is generated ?

sonikf: For down payments(advance) there is no legal obligation to issue an invoice but instead for cash payments only we issue a simple receipt referencing the client. In bank payments the bank statement is considered as receipt because payment is made to a "professional account".

For more context please read this [issue](https://github.com/Dolibarr/dolibarr/issues/23076) 

So in case of an audit things get messy

Since then the solution we've came up with to make dolibarr compliant with Greek legislation is by using the following modules:
* https://www.dolistore.com/en/modules/1834-CTPT.html    
it creates the mask using {tttt} 

* https://github.com/evansgl/dolibarr-mydata                        
it uploads invoice data to ministry database

* https://github.com/TechnickSysAid/mask4greece              
it generates pdf retrieving data from mydata module plus other requirements 

Any suggestion or insight as to how we may proceed would be greatly appreciated
Thank you

eldy (commented):
Ly understanding is that you need for Greece, the module :
Mydata
Mask4greece. 

What is name of mask4greece. Mask4greece or external_access ?
Also, if mask4greece use data inside module mydata to generate pdf, it depends on it ? Why a different module than mydata ?
May be 1 module called mydata4greece would be easier to understand ?

About the module cpt, because it uses the type of thirdpary (so change it at each invoice generation), i have introduced a new colum "subtype" into invoice. 
Do you think we are going on the good way with this column ? If yes, we can introduce into the invoice form, a "if greece, then show the combo with list of invoice type for greece". Then we will need a new dictionary table to define/list the table of subtypes.

sonikf:
@eldy 

You recently introduced category of operations, my hope was that is similar to our use case but unfortunately is not, what is a mention for you for us is another invoice type so we have 
sale invoice
sale invoice eu
sale invoice other countries
service invoice
service invoice eu
service invoice other countries
and so on.. also cash register receipts are also invoice types
sale receipt
service receipt....

> Ly understanding is that you need for Greece, the module : Mydata Mask4greece.

Yes also ctpt 
ctpt is useful for situations like
1. I have three shops same vat id so shop A will have mask A, shop B mask B....
2. I have big client so i assign mask "BC" for him
3. Because there is no replacement invoice in our legislation when i make a mistake in invoice with mask A for example i have to issue credit note with mask Α and new invoice with mask B
4. .... endless possibilities the system is f**ked up

> What is name of mask4greece. Mask4greece or external_access ?
It's mask4greece. I made pr that fixes it.

>  Also, if mask4greece use data inside module mydata to generate pdf, it depends on it ? Why a different module than mydata ? May be 1 module called mydata4greece would be easier to understand ?

I agree but different maintainers. I have to contact maintainers first.
mask4greece is fork of mercure with only invoice and credit note and ctpt for mask and pdf forks of sponge, cyan, espadon...
it use invoice type from mydata so for Greek customers is "sale invoice A-0001" and for eu and other the default "invoice A-0001"
Another example is an invoice - delivery type combination where delivery date and vehicle registration should be in the invoice pdf.
As soon as possible i'll commit upstream all changes maybe useful for other countries such as page number, analysis per vat percent..

> About the module cpt, because it uses the type of thirdpary (so change it at each invoice generation), i have introduced a new colum "subtype" into invoice. Do you think we are going on the good way with this column ? If yes, we can introduce into the invoice form, a "if greece, then show the combo with list of invoice type for greece". Then we will need a new dictionary table to define/list the table of subtypes.

Yes, anything that could be in the core is better.
Please explain how that will work with ctpt and invoice type. Would it be like 
| Subtype  | mask | country|
| ----------------------  | ------------- | ------------- |
| Sale invoice  | A | GR |
| Sale invoice| B |  GR |
|  Service invoice | A |  GR | 

Thanks so much for taking the time!
I will commit any changes you instruct such as dictionary and so on

eldy:
The field subtype is now in database (develop branch). The dictionnary is also available. You will see it if country is greece. 
For the moment, I did not include the column mask into the table but i will include it.
I have no idea on how it will works, my idea is just to store information according to what it seems we need. Goal is to have database structure ready as soon as possible to allow development of screen after. I won't be able to work on screens or numbering or pdf modules, but i can speed up changes for database structure.

sonikf:
@eldy 
Thank you very much, i've found them.
The combo box idea seems nice.
I will inform the Greek community and as soon as the database structure is there, we will start working!

Issue #24482, opened 2023-04-11
Improvements to product list requested by customer


vdegrandpre:
### Feature Request

The feature request consist of two things :

1. Product list shall display stock from specified warehouse only when a global variable STOCK_CONSUMPTION_FROM_MANUFACTURING_WAREHOUSE is configured and that a warehouse is specified. Otherwise, stock quantity from all warehouses are displayed.
2. When the slow option STOCK_SHOW_VIRTUAL_STOCK_IN_PRODUCTS_COMBO is active and the global variable PRODUCT_STOCK_LIST_SHOW_VIRTUAL_WITH_NO_PHYSICAL is active, the product line shall be displayed as stock-OK.

### Use case

For operational planning, customer need to have only the stock from the selected warehouse visible, as well as all virtual stocks that are pending, in the product list. In case the product have virtual stock but no physical stock, the customer need to see the product line as stock-OK.

### Suggested implementation

Use the global variables PRODUCT_STOCK_LIST_SHOW_VIRTUAL_WITH_NO_PHYSICAL and STOCK_CONSUMPTION_FROM_MANUFACTURING_WAREHOUSE together with the STOCK_SHOW_VIRTUAL_STOCK_IN_PRODUCTS_COMBO to acheive the requested feature.

### Suggested steps

- A change in core/class/html.form.class.php is necessary to modify select_produits_list() function to take account of warehouseId in select, as well as displaying the product line stock-OK
- A change in core/tpl/objectline_create.tpl.php is necessary to call the select_produits_list function with warehouse_id
- A change in mrp/mo_production.php to call the select_produits_list function with warehouse_id

PR #24484, opened 2023-04-11
NEW Show stock only from selected warehouse in product list and sh…

vdegrandpre:
…ow line as stock-OK where virtual stock but no physical stock

# NEW|New #24482 [Improvements to product list requested by customer](https://github.com/Dolibarr/dolibarr/issues/24482)

1. *Show stock quantity of selected warehouse only in customer order and MO*

- Using global variable STOCK_CONSUMPTION_FROM_MANUFACTURING_WAREHOUSE and an associated warehouse in manufacturing order or in customer order product lists
- In Customer order product list, global ENTREPOT_EXTRA_STATUS need also to be set.

2. *Show product line as stock-OK when there are virtual stock but no physical stock*

- Using global variable PRODUCT_STOCK_LIST_SHOW_VIRTUAL_WITH_NO_PHYSICAL along with STOCK_SHOW_VIRTUAL_STOCK_IN_PRODUCTS_COMBO to show product lines as stock-OK in manufacturing order (add production item) product list and customer order defined product list.


eldy:
It seems this PR implements 2 features.
One to allow selection of n warehouses in filter.
Another one to implement option STOCK_CONSUMPTION_FROM_MANUFACTURING_WAREHOUSE.
If i am right, can you split PR in 2 ?

vdegrandpre:
Will split in two and create new PR accordingly. So closing this one.

PR #24498, opened 2023-04-12
FIX: The minimum price in proposals is not checked correctly

fmarcet:
If we add a product to a budget whose client does not apply VAT (for example, a European client), it does not correctly calculate the minimum price. Since the VAT is 0, it considers that the HT and TTC price is the same. But in the minimum prices VAT does apply, with what you are really checking is the minimum price with VAT and the sale price of the product without VAT.

I don't know if the solution I propose is the best, but I have seen that, for example, only the minimum price without VAT is checked on invoices. So I've done something similar here in budgets.


htdocs/comm/propal/card.php
```diff
@@ -1238,7 +1238,7 @@
 
 			// Prepare a price equivalent for minimum price check
 			$pu_equivalent = $pu_ht;
-			$pu_equivalent_ttc = $pu_ttc;
+			//$pu_equivalent_ttc = $pu_ttc;
```
eldy, line 1241: I see no reason to disable this line as this is just an assignation.
If as a business rule, we decide to check minimum ONLY for price without vat, we should disable the code that do the comparison but not this assignation.

PR #24518, opened 2023-04-15
 NEW : Update card.php fichinter - Total duration in hours

cmfpmatik:
This code generates second row of a table to display the total duration of an employee, one in time format (DD HH:MM) and the other in hour format only (HH).

The first row (original code) displays the title "TotalDurationTime" using the "trans" method of the $langs object to translate the string. The total time is converted to HH:MM:SS format using the "convertSecondToTime" function, which takes into account the daily work time duration specified in the global variable $conf->global->MAIN_DURATION_OF_WORKDAY.

The second row displays the title "TotalDurationHour" using the "trans" method of the $langs object to translate the string. The total duration is converted to hours using the "convertDurationtoHour" function, specifying "s" as the time unit (seconds).

# NEW|New [Total duration also in hours]
When selling time tickets or time in batches, the conversion to number of days makes no sense.
Add second row of a table to display the total duration of an employee, one in time format (DD HH:MM) and the other in hour format only (HH).

### DIFF
--- a/htdocs/fichinter/card.php
+++ b/htdocs/fichinter/card.php
@@ -9,6 +9,7 @@
  * Copyright (C) 2015-2016  Abbes Bahfir            <bafbes@gmail.com>
  * Copyright (C) 2018-2022 	Philippe Grand       	<philippe.grand@atoo-net.com>
  * Copyright (C) 2020       Frédéric France         <frederic.france@netlogic.fr>
+ * Copyright (C) 2023       Benjamin Grembi         <benjamin@oarces.fr>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


eldy:
Can you provide a screenshot of the page that is modified ?

cmfpmatik:
Screenshot_20230418_183219

eldy:
Both fields represent the same information, only the unit differs. So i suggest to keep only one line: The original line and you just add the value in the second unit on the same line
For example
Duration:         2 d 8:00   =   56 h

cmfpmatik:
If you want, but it's not the same information.
The first line can be converted to "working day", the second line is only used to charge something per hour...

But, by default, 1d corresponds to 24h, so it's not really useful from a bisness point of view (maybe in dictatorship if the employee works hard all the time?)
But if you have specified a daily working time, 1d can be equal to 8h.

eldy:
Both lines shoul be here to charge.
If working day is 8h, and total is 18h, you should get 2d 02:00 so 2 days to charge + 2 hours. Don't you get this ?
First is in unit "day" if you want to charge using the daily rate, second is unit hour if you want to charge using hourly rate.
Both information on screen comes from the same field in same table.

cmfpmatik:
by default, 1d = 24h. You have to do know and do custom conf to change that (wihtout any detail information about it anywhere) and nobody charge like that.
The same information in the same database can be used differently.
For better UX design, you need to separate all information with different interest.
Sometimes billing is by hours, other by day or other things.
Developer logic is not user logic.

If you have contract with mention of hours, you must have invoice in hours...

But whatever, the correction was made as you want.

PR #24529, opened 2023-04-17
fix : Warning: Undefined property: stdClass:: in /home/httpd/vhosts/a…

grandoc:
…flac.fr/domains/dev.aflac.fr/httpdocs/paypal/admin/paypal.php on line 190


grandoc:
Bad fixing, should be with getDolGlobalString()

PR #24541, opened 2023-04-18
FIX for empty shipping

altairis-tof:
SHIPMENT_GETS_ALL_ORDER_PRODUCTS constant was not always tested so it was impossible to have all lines empty in shipping before receiving products


altairis-tof:
did not understood why second commit has not updated this PR... so i've done another #24542

PR #24544, opened 2023-04-18
FIX: status filter on order list

jyhere:
Combinated status -4 is not present (Validated + Sent) , and -2 status is renamed to "To Process"

### DIFF
--- a/htdocs/commande/list.php
+++ b/htdocs/commande/list.php
@@ -511,7 +511,7 @@ if (empty($reshook)) {
 							if ($result > 0) {
 								$lineid = $result;
 								if (!empty($createbills_onebythird)) //increment rang to keep order
-									$TFactThirdNbLines[$rcp->socid]++;
+									$TFactThirdNbLines[$cmd->socid]++;
 							} else {
 								$lineid = 0;
 								$error++;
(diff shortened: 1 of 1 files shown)


htdocs/commande/list.php
```diff
@@ -1741,8 +1741,9 @@
 			Commande::STATUS_VALIDATED=>$langs->trans("StatusOrderValidated"),
 			Commande::STATUS_SHIPMENTONPROCESS=>$langs->trans("StatusOrderSentShort"),
 			Commande::STATUS_CLOSED=>$langs->trans("StatusOrderDelivered"),
+			-4=>$langs->trans("StatusOrderValidatedShort").'+'.$langs->trans("StatusOrderSentShort"),
 			-3=>$langs->trans("StatusOrderValidatedShort").'+'.$langs->trans("StatusOrderSentShort").'+'.$langs->trans("StatusOrderDelivered"),
-			-2=>$langs->trans("StatusOrderValidatedShort").'+'.$langs->trans("StatusOrderSentShort"),
+			-2=>$langs->trans("StatusOrderToProcessShort"),
```
eldy, line 1746: Filter executed when using -2 is c.fk_statut IN (1,2) 
So we make a filter on 2 different status. So we should see the 2 status on the combo list.

jyhere: My bad, did not see the sql condition changed from V16... closing the request. :)

PR #24547, opened 2023-04-18
property and doc

frederic34:


alexandre-janniaux:
Hi, maybe usage of `$this->line` can be removed instead to only use a local variable, since the `$this->line` value is never read back.

PR #24553, opened 2023-04-19
New mo bom add line default workstation

atm-lena:
# NEW|New 

Following https://github.com/Dolibarr/dolibarr/pull/24538, step 2 :
- Add default workstation of service in fk_default_workstation of line when we add line for mo and bom or when we create a bom
- Display fk_default_workstation of line first

### DIFF
--- a/htdocs/bom/bom_card.php
+++ b/htdocs/bom/bom_card.php
@@ -171,7 +171,7 @@ if (empty($reshook)) {
 		$efficiency = price2num(GETPOST('efficiency', 'alpha'));
 		$fk_unit = GETPOST('fk_unit', 'alphanohtml');
 
-		if(!empty($idprod) && $conf->workstation->enabled){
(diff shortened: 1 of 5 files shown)


htdocs/bom/bom_card.php
```diff
@@ -170,6 +170,13 @@
 		$disable_stock_change = GETPOST('disable_stock_change', 'int');
 		$efficiency = price2num(GETPOST('efficiency', 'alpha'));
 		$fk_unit = GETPOST('fk_unit', 'alphanohtml');
+
+		if (!empty($idprod) && $conf->workstation->enabled) {
```
eldy, line 174: From v17, try to use isModEnabled('workstation') instead of conf->global ...

atm-lena: Oh yes, sorry.. thank you !

PR #24557, opened 2023-04-20
FIX display project tasks list columns

atm-steve:
Fix columns in project tasks list. Add extrafields in columns list add empty space on total line.


atm-steve:
@eldy Can't fix stickler, please check the fix

eldy:
You IDE is probably not configured with the dolibarr syntax rules.
Indentation must use tabulation instead of spaces.
Just set your IDE to use tab instead of space, at least for the dolibarr project, then open and resave the same page. The IDE should save the file correctly.

atm-jpb:
@eldy

atm-greg:
@eldy IDE problem solved

john-botella:
@eldy IDE problem solved

PR #24560, opened 2023-04-20
NEW: Multicurrency REST API to create, update, delete, update rate...

alexboin:
# FIX: Invalid SQL syntax in Currency update request
In the update currency query, a comma was missing between the list of fields, leading to an invalid SQL query.

# NEW: Add permissions for currencies and their rates (and their french translations)
In view of adding routes to the REST API to control currencies and their exchange rates, 3 permissions have been added to the Multicurrency module:
- the right to consult the currencies and their exchange rates
- the right to create or modify a currency and its exchange rates
- the right to remove a currency or its exchange rates

French translations have also been added.

# NEW : Add endpoints to the Multicurrency REST API
Routes have been added to the REST API of the Multicurrency module, allowing to:
- retrieve a currency and its last exchange rate:
   - by its id
   - by its code
- retrieve the list of exchange rates for a currency
- update the exchange rate of a currency
- create a currency (with an optional initial exchange rate)
- modify a currency (name and code)
- delete a currency


htdocs/langs/fr_FR/admin.lang
```diff
@@ -1004,6 +1004,9 @@ Permission23001=Voir les travaux planifiés
 Permission23002=Créer/Modifier des travaux planifiées
 Permission23003=Effacer travail planifié
 Permission23004=Exécuter travail planifié
+Permission40001=Consulter les devises et leurs taux de change
```
eldy, line 1007: Can you add the new translation entries into the file in en_US file instead of fr_FR ?
Other language files, including french, will be automaticaly included into transifex platform for the translation team. Sync en_US to transifex is done once a week, and sync all translated languages to Dolibarr is done before beta and before each release.

alexboin: Hello,
I added the translations entries in the en_US file as requested, as well as fixed the stickler-ci continuous integration test error.

PR #24572, opened 2023-04-23
FIX: Propagate correct origin/origin_id when creating order from proposal

creekorful:
When creating an order from a proposal using `\Commande::createFromProposal` the original line information were lost, rendering any trigger on `LINEORDER_INSERT` useless.

This PR add the original line information to the order line which in conjunction with `MAIN_CREATEFROM_KEEP_LINE_ORIGIN_INFORMATION` allows to receive original line information in a `LINEORDER_INSERT` trigger.

### DIFF
--- a/htdocs/commande/class/commande.class.php
+++ b/htdocs/commande/class/commande.class.php
@@ -1325,6 +1325,9 @@ class Commande extends CommonOrder
 			$line->marge_tx			= $marginInfos[1];
 			$line->marque_tx		= $marginInfos[2];
 
+			$line->origin           = $object->element;
+			$line->origin_id        = $object->lines[$i]->id;
+
 			// get extrafields from original line
 			$object->lines[$i]->fetch_optionals();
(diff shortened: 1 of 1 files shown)


htdocs/commande/class/commande.class.php
```diff
@@ -1325,6 +1325,9 @@ public function createFromProposal($object, User $user)
 			$line->marge_tx			= $marginInfos[1];
 			$line->marque_tx		= $marginInfos[2];
 
+			$line->origin           = $object->element;
```
creekorful, line 1328: One question tho: as per the documentation I can read about `MAIN_CREATEFROM_KEEP_LINE_ORIGIN_INFORMATION` it looks like the element I should be passing here should be `$object->lines[$i]->element` and no the parent element. 

However in many triggers I'm using, when handling line it's always the parent element (propal, ...) instead of the line element (propaldet, ...) who I am receiving.

PR #24574, opened 2023-04-23
NEW Filter product list on warehouse ID

vdegrandpre:
# NEW|New for #24482 - Part 1/2

Using global variable STOCK_CONSUMPTION_FROM_MANUFACTURING_WAREHOUSE and an associated warehouse in manufacturing order or in customer order product lists. Note : In Customer order product list, global ENTREPOT_EXTRA_STATUS need also to be set.

Allow product list to filter by Warehouse ID. Only available in Manufacturing Order and Customer Order product lists.


vdegrandpre:
Salut @eldy , ce PR était bloqué sur une fusion à effectuer depuis un petit bout de temps. J'ai corrigé le tout pour intégration. Est-ce possible de m'indiquer si et quand ce sera intégré, à peu près, car j'ai d'autres modifs à soumettre et il faut y aller un PR à la fois. Merci !

htdocs/core/class/html.form.class.php
```diff
@@ -2630,6 +2631,9 @@ public function select_produits_list($selected = '', $htmlname = 'productid', $f
 			if (preg_match('/warehouseinternal/', $warehouseStatus)) {
 				$warehouseStatusArray[] = Entrepot::STATUS_OPEN_INTERNAL;
 			}
+			if ( is_numeric($warehouseStatus) ) {
```
eldy, line 2634: Can you also update the comment of the method to add that thisbparameter can also be the numeric id of a given warehouse ?

vdegrandpre: I agree.

vdegrandpre: @eldy Trying to implement that extra parameter, I think there is something wrong in the function calls on lines 269 and 271 of /core/tpl/objectline_create.tpl.php as it seem there is an unwanted parameter (the 10th is an array() that does not match any parameter).

Can you validate and confirm I can remove that parameter and add the warehouseId parameter at the end of the function call ?

vdegrandpre: Précision : calling select_produits() of $form

vdegrandpre: @eldy , should I file a separate bug for that and fix it before pushing the pull request ?

eldy: > @eldy Trying to implement that extra parameter, I think there is something wrong in the function calls on lines 269 and 271 of /core/tpl/objectline_create.tpl.php as it seem there is an unwanted parameter (the 10th is an array() that does not match any parameter).
> 
> Can you validate and confirm I can remove that parameter and add the warehouseId parameter at the end of the function call ?

You can't remove parameter at position 10 that is an array, because it may be used by external module. Seems to be always null or array() for calls found into dolibarr core.
It also may be filled in a future (the parameter is exploited inside the function at least in 1 place)

So i suggest to add a parameter $warehouseIds with id of warehouse to filter on a given warehouse.

PR #24582, opened 2023-04-24
PHP8 compatibility

MaximilienR-easya:
# Fix 
PHP8 Fix some undefined property warning
undefined property

undefined variables2

and
return type deprecated

return type deprecated

### DIFF
--- a/htdocs/includes/stripe/stripe-php/lib/Collection.php
+++ b/htdocs/includes/stripe/stripe-php/lib/Collection.php
@@ -2,6 +2,8 @@
 
 namespace Stripe;
 
+use Traversable;
+
 /**
  * Class Collection.
  *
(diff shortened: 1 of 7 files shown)


htdocs/includes/stripe/stripe-php/lib/Collection.php
```diff
@@ -2,6 +2,8 @@
 
 namespace Stripe;
 
+use Traversable;
```
eldy, line 5: The includes dir is reserved for external lib.
We do not update files of external dir. Instead we must update the fill lib files with a new version provided by the maintener of the app.  Upgrade of external lib must be into a separated PR. 
Can you repush the PR without the changes into Stripe ?

MaximilienR-easya: The stripe modification has been removed

PR #24583, opened 2023-04-24
Prevent PHP Warning in error log file, when the generated doc file does not exists. 

BitKFu:
This is the case when we create a PDF file out of a ODT file and automatically delete the ODT.
The indexed file will then be the ODT, which can't be found on drive (because it has been deleted when converting to PDF) and therefore creates an error log entry.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -5671,7 +5671,7 @@ abstract class CommonObject
 					$update_main_doc_field = 1;
 				}
 
-				// Check that the file exists, before indexing it. 
+				// Check that the file exists, before indexing it.
 				// Hint: It does not exist, if we create a PDF and auto delete the ODT File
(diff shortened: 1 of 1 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -5671,7 +5671,11 @@ protected function commonGenerateDocument($modelspath, $modele, $outputlangs, $h
 					$update_main_doc_field = 1;
 				}
 
-				$this->indexFile($destfull, $update_main_doc_field);
+				// Check that the file exists, before indexing it.
+				// Hint: It does not exist, if we create a PDF and auto delete the ODT File
+				if (file_exists($destfull)) {
```
eldy, line 5676: Does this is ok if we use dol_is_file instead of file_exists ?

BitKFu: I changed it

PR #24592, opened 2023-04-25
FIX define a different CRON_KEY by entity

hregis:


htdocs/cron/admin/cron.php
```diff
@@ -44,7 +44,7 @@
 
 	$db->begin();
 
-	$i += dolibarr_set_const($db, 'CRON_KEY', GETPOST("CRON_KEY"), 'chaine', 0, '', 0);
+	$i += dolibarr_set_const($db, 'CRON_KEY', GETPOST("CRON_KEY"), 'chaine', 0, '', $conf->entity);
```
eldy, line 47: Don't understand how this can work because into the linux crontab, we have only one entry to execute the cron taks. Then the CLI script process each entity separately but the triggering is common to the dolibarr installation.

eldy:

PR #24593, opened 2023-04-25
FIX : reopen supplier order after line delete if needed

altairis-tof:
if the supplier order has status "All products received" but you delete a dispatch line, the status should be set to "Partially received".


htdocs/fourn/commande/card.php
```diff
@@ -857,8 +857,19 @@
 		}
 
 		if (!$error) {
-			$db->commit();
+			// reopen order if necessary
+			if ($object->status == CommandeFournisseur::STATUS_RECEIVED_COMPLETELY) {
+				if ($object->setStatus($user, CommandeFournisseur::STATUS_RECEIVED_PARTIALLY) < 0) {
+					$db->rollback();
```
eldy, line 863: Making a rollback here means we will make duplicate rollback (because it is also done few lines later.

altairis-tof: Thanks @eldy ; it is fixed now.

eldy (approved):

PR #24595, opened 2023-04-25
fix warning PHP8.1

frederic34:


htdocs/core/lib/company.lib.php
```diff
@@ -1040,7 +1040,7 @@ function show_contacts($conf, $langs, $db, $object, $backtopage = '', $showuserl
 					'label'=>$extrafields->attributes[$contactstatic->table_element]['label'][$key],
 					'checked'=>(($extrafields->attributes[$contactstatic->table_element]['list'][$key] < 0) ? 0 : 1),
 					'position'=>1000 + $extrafields->attributes[$contactstatic->table_element]['pos'][$key],
-					'enabled'=>(abs($extrafields->attributes[$contactstatic->table_element]['list'][$key]) != 3 && $extrafields->attributes[$contactstatic->table_element]['perms'][$key]));
+					'enabled'=>(abs((int) $extrafields->attributes[$contactstatic->table_element]['list'][$key]) != 3 && $extrafields->attributes[$contactstatic->table_element]['perms'][$key]));
```
eldy, line 1043: May be we should use dol_eval() here for the case the "list" property is a condition ?

frederic34: I took this as reference : 
https://github.com/Dolibarr/dolibarr/blob/5b90302bf85cec4f1d809da65a11e3bff25651ac/htdocs/core/tpl/extrafields_list_array_fields.tpl.php#L26-L28

PR #24601, opened 2023-04-26
FIX : update import_key when import update an object

atm-lena:
# FIX|Fix #
Import : update import_key even if the object is existing (not only on creation, but also on update)


eldy:
Goal of impot_key is to be able to rollback a large import campaign when several steps are done like when files are run, then updated from other sources, or rerun after source file error.
If we want to cancel completely the campaign, currently, we can rely on import_key.
With this pr, i am afraid we won't be able anymore. But i see the need: tracking record modified by 1 particular step of the campaign.
We add 1 capability but we remove also another one. So my mind balance between merging and not...

atm-lena:
@eldy  Maybe add an other column for import's update information ? For the next version of Dolibarr
I think it's relevant to note import's update somewhere.

PR #24604, opened 2023-04-26
"FIX" - API reception _validate function 

VMR-Global-Solutions:
Change made in _private function _validate_ in reception/class/api_receptions.class.php:712
replaced line 715
_foreach (Shipments::$FIELDS as $field) {_ => _foreach (Receptions::$FIELDS as $field) {_
[api_receptions.class.zip](https://github.com/Dolibarr/dolibarr/files/11331620/api_receptions.class.zip)


eldy:
The PR seems to contains only commits from other dev but i can't find any code change from your own. Can you give the commit number of your changes ?

VMR-Global-Solutions:
I've submit an other PR.

PR #24624, opened 2023-04-27
Fix conf.class.php: add missing syslog property

alexandre-janniaux:
# Fix conf.class.php: fix dynamic property warning

Fix a deprecated dynamic property warning:

    Deprecated: Creation of dynamic property Conf::$syslog is
    deprecated in dolibarr/htdocs/install/inc.php on line 441


eldy:
It seems trouble was fixed previously by the complete removal of all ->syslog assignement in other PR.
Do you still have the trouble on develop ?

alexandre-janniaux:
@eldy I adapted the patch to what's left of the issue now, thanks for checking this first!

PR #24625, opened 2023-04-27
Doc for commonobject.class.php

frederic34:


eldy:
Merged manually to solve conflict.

PR #24635, opened 2023-04-28
API add / fix filter on extrafields in index()

VMR-Global-Solutions:
Add / change in api_xxx.class.php files the ability to filter on extrafields when missing in the public functions index() by changing the SQL query with the pattern 
" FROM ".MAIN_DB_PREFIX"xxxx AS t LEFT JOIN ".MAIN_DB_PREFIX."xxxx_extrafields AS ef ON (ef.fk_object = t.rowid)

### DIFF
--- a/htdocs/adherents/class/api_members.class.php
+++ b/htdocs/adherents/class/api_members.class.php
@@ -222,7 +222,7 @@ class Members extends DolibarrApi
 		}
 
 		$sql = "SELECT t.rowid";
-		$sql .= " FROM ".MAIN_DB_PREFIX."adherent as t";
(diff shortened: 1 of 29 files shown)


JonBendtsen:
In case anyone else ever needs to understand how to correlate

- extrafield which is called Complementary attributes in the module configuration, and array_options in the json, and `__EXTRAFIELD_` when sending emails.
- Complementary attributes as defined under a module, in this case the product module
- sqlfilters in the API
- keys in the json exported by the API

Screenshot of definition of an Complementary attribute
image

So as you can see I have one defined which I call `service_type` and which values I allow it to have.


**sqlfilters**
In the API under the `sqlfilters` field I would write this
```
(ef.service_type:like:'Workshop')
```

**JSON output**
In the resulting JSON output I would get something like this
```
    "array_options": {
      "options_service_type": "Workshop",
```


You can also combine multiple queries in the `sqlfilters` field like this
```
(t.tosell:=:1) and (ef.service_type:like:'Workshop')
```

PR #24646, opened 2023-04-28
link clients mail/card to ticket messages

MaximilienR-easya:
# New
We can now see a link to the contact card or just the email (if not found) of the contact who send a message via the public interface.
Inside the public interface, we do not have a link to the contact card, just the name or email (if not found).

### DIFF
--- a/htdocs/contact/class/contact.class.php
+++ b/htdocs/contact/class/contact.class.php
@@ -1535,7 +1535,7 @@ class Contact extends CommonObject
 			$linkend = '</a>';
 		}
 
-		$result .= $linkstart;
+		$result .= (($option == 'nolink') ? '' : $linkstart);
 		if ($withpicto) {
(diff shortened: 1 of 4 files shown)


eldy:
Can you provide a screenshot ?

eldy:
Can you provide a screenshot ?

MaximilienR-easya:
In this screenshot, TEST Jean and MANIQUANT are clients who have sent messages via the public ticket interface and this is how we see it in the list of messages on the public page
Public interface message side

Here we see the same case as the first screenshot but from the ticket card messages list, here we have a link to the card of the contact present on their name
ticket card interface

At this one, we have a case with a client who is not found (in case the option to permit ticket from people who are not client in the options)
User not found public side
User not found show email

MaximilienR-easya:
@eldy I have provided some screenshot

PR #24655, opened 2023-05-01
NEW handle contract document (list, build and upload) in documents Rest API

romain-neil:
# NEW API: upload contract document
Add the ability to upload contract via the API


romain-neil:
Tested and working in dolibarr v17.0.0

Issue #23331, opened 2022-12-25
Nested projects


mschamp:
### Feature Request

For a project that consists of multiple sub projects, it would be a nice feature to be able to assign a 'parent' project.
When loading the parent project, the card and overview screen give a total overview including the status of the child projects.
This way, more structure can be added.

JoAnMaFe:
This would certainly be a very interesting feature.
It should be accompanied by appropriate GANTT diagrams.

edupulpillo:
the feature is not present in v 19.0.0

JonBendtsen:
In case anyone wonders how to enable this feature, then see this comment

https://github.com/Dolibarr/dolibarr/pull/24856#issuecomment-2213513998

PR #24656, opened 2023-05-01
 #23331 Add parent project field to database

mc2contributor:
Provide support for sub-projects, as requested in issue #23331, "Nested Projects".
This adds the required database field.


mc2contributor:
Should the new field for the parent project ID be 'parent', 'fk_projet', or something else?

eldy:
> Should the new field for the parent project ID be 'parent', 'fk_projet', or something else?

Field should be   fk_project

Note that this represent a very important work due to credentials on projets and tasks. That may need a long time because such a feature is stable so to not break existing features, i recommend to implement all code (once this PR on structure is ok with name of field) with a condition 
` if (getDolGlobalInt('PROJECT_SUPPORT_SUB_PROJECT'))`

htdocs/install/mysql/migration/17.0.0-18.0.0.sql
```diff
@@ -411,3 +411,4 @@ CREATE TABLE llx_c_invoice_subtype (
 
 ALTER TABLE llx_c_invoice_subtype ADD UNIQUE INDEX uk_c_invoice_subtype (entity, code);
 
+ALTER TABLE llx_fk_project ADD COLUMN parent int DEFAULT NULL;
```
eldy, line 414: There is no table llx_fk_project. fk_project must be field name not table name

mc2contributor: Whoops! Thanks for catching that.

mc2contributor:
Added "Status" column to child projects list to more closely fit the requirements of the feature request.

eldy:
According to Contributing rules. Changes on database structure must be done in a separated and standalone PR.
Code that use the new structure into another one.

mc2contributor:
PR revised to include only the database structure change.

PR #24038, opened 2023-02-24
Better management for sale representative link / filter

defrance:
if we use classic join, we don't have the possibility to left join other table...

# Instructions
*This is a template to help you make good pull requests. You may use [Github Markdown](https://help.github.com/articles/getting-started-with-writing-and-formatting-on-github/) syntax to format your issue report.*
*Please:*
- *only keep the "FIX", "CLOSE" or "NEW" section* (use uppercase to have the PR appears into the ChangeLog, lowercase will not appears)
- *follow the project [contributing guidelines](/.github/CONTRIBUTING.md)*
- *replace the bracket enclosed texts with meaningful information*

# FIX|Fix #[*issue_number Short description*]
[*Long description*]

# CLOSE|Close #[*issue_number Short description*]
[*Long description*]

# NEW|New [*Short description*]
[*Long description*]

PR #24664, opened 2023-05-02
Better management for sale representative link

defrance:
replace #24038  PR with specificated way for filtering
my fears of performance with respect to the "left join" function are unfounded (which was the case with an "IN SELECT")

### DIFF
--- a/htdocs/societe/list.php
+++ b/htdocs/societe/list.php
@@ -13,6 +13,7 @@
  * Copyright (C) 2020       Open-Dsi         		<support@open-dsi.fr>
  * Copyright (C) 2021		Frédéric France			<frederic.france@netlogic.fr>
  * Copyright (C) 2022		Anthony Berton			<anthony.berton@bb2a.fr>
+ * Copyright (C) 2023		Charlene Benke			<charlene@patas-monkey.com>
  *
  * This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


htdocs/societe/list.php
```diff
@@ -512,33 +513,39 @@
...
+	$sql .= " AND EXISTS (SELECT 1 FROM  ".MAIN_DB_PREFIX."societe_commerciaux as sc";
+
+	//if (empty($user->rights->societe->client->voir) && (empty($conf->global->MAIN_USE_ADVANCED_PERMS) || empty($user->rights->societe->client->readallthirdparties_advance)) && !$socid)	$sql .= " AND s.rowid = sc.fk_soc AND sc.fk_user = ".((int) $user->id);
+	if (empty($user->rights->societe->client->voir) && !$socid) {
+		$sql .= " AND s.rowid = sc.fk_soc AND sc.fk_user = ".((int) $user->id);
+	}
+	if ($search_sale && $search_sale != '-1' && $search_sale != '-2') {
+		$sql .= " AND s.rowid = sc.fk_soc"; // Join for the needed table to filter by sale
+	}
+	if ($search_sale == -2) {
```
eldy, line 541: For the search for record with no sales representative, we must use a NOT EXISTS instead of a EXISTS into a select that will always return nothing.
Did you try the suggested code into comment of PR #24038 ?

PR #24669, opened 2023-05-03
FIX - Remove warning on lettering - Impossible to write in ledger

aspangaro:
When auto-lettering is enabled and you try to transfer in ledger, you can see this error while the transaction is balanced : 

image

Fix also wrong colspan in sells journal


aspangaro:
#24835 New PR for v16

aspangaro:
#24836

PR #24675, opened 2023-05-03
FIX parameters array_merge + global activation

inoveaconseil:


eldy:
The PR introduce the new feature CaRE_GRANTCHILD
Can you describe the goal of thisnoption and push it into develop instead if 13.0...

inoveaconseil:
The code was incorrect and with this parameter, we can manage the argument. If you want manage the rank of all the child.

htdocs/core/class/commonobject.class.php
```diff
@@ -2671,7 +2673,7 @@ public function getChildrenOfLine($id, $includealltree = 0)
 			if ($this->db->num_rows($resql) > 0) {
 				while ($row = $this->db->fetch_row($resql)) {
 					$rows[] = $row[0];
-					if (!empty($includealltree)) $rows = array_merge($rows, $this->getChildrenOfLine($row[0]), $includealltree);
+					if (!empty($includealltree)) $rows = array_merge($rows, $this->getChildrenOfLine($row[0]));
```
eldy, line 2676: Why removing the feature $includealltree ?

eldy:

PR #24676, opened 2023-05-03
NEW Add default supplier order input method

FlorentPoinsaut:
# NEW Add default supplier order input method

Ability to set a default supplier order input method with the conf `SUPPLIER_ORDER_DEFAULT_INPUT_METHOD`.

### DIFF
--- a/htdocs/fourn/commande/card.php
+++ b/htdocs/fourn/commande/card.php
@@ -2650,7 +2650,13 @@ if ($action == 'create') {
 
 			// Force mandatory order method
 			print '<tr><td class="fieldrequired">'.$langs->trans("OrderMode").'</td><td>';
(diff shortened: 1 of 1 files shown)


priojk:
Hi @FlorentPoinsaut,

actually, for v18 this problem is already solved in a different way: The root cause was issue #24356, which is now solved with PR #24369. And there is now even a patch for v17 (I think in v17.0.1).

Long story short: it is now possible to use Dolibarr's standard mechanism for preselecting an option (Home -> Settings -> Default Values), in particular also for order methods.

FlorentPoinsaut:
Hi @priojk ,

Oh, this is wonderful! Thank you for your work and your explanations!

PR #24680, opened 2023-05-03
fix typo

priojk:
compta/facture is a customer invoice, not a supplier invoice

Since this hook has just recently been introduced by myself, I hope it's still ok to correct this typo to avoid future confusion.

### DIFF
--- a/htdocs/accountancy/admin/subaccount.php
+++ b/htdocs/accountancy/admin/subaccount.php
@@ -130,7 +130,7 @@ llxHeader('', $title, $help_url);
 
 
 // Customer
-$sql = "SELECT sa.rowid, sa.nom as label, sa.code_compta as subaccount, '1' as type, sa.entity";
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/prelevement.php
```diff
@@ -695,7 +695,7 @@
 
 	// Hook to change amount for other reasons, e.g. apply cash discount for payment before agreed date
 	$parameters = array('remaintopay' => $resteapayer);
-	$reshook = $hookmanager->executeHooks('finalizeAmountOfSupplierInvoice', $parameters, $object, $action); // Note that $action and $object may have been modified by some hooks
+	$reshook = $hookmanager->executeHooks('finalizeAmountOfCustomerInvoice', $parameters, $object, $action); // Note that $action and $object may have been modified by some hooks
```
eldy, line 698: Depending if $type is 'bank-transfert' or not, the page is used commonly for vendor or supplier invoice. 
So i suggest name finalizeAmountOfInvoice

PR #24681, opened 2023-05-03
New: add hooks to apply cash discount agreements

priojk:
The remaining amount of an invoice to be paid can depend on a cash discount agreement. I.e. if the invoice gets paid within a certain timeframe which is shorter than the due date, an agreed percentage can be deducted from the invoice total. This is known as cash discount or escompte in FR or Skonto in DE. E.g. due date is 30 days from invoice date, but if it gets paid within 7 days, 3% can be deducted from the total.

The added hooks allow to add cash discount handling in a separate module for countries, where this applies.


htdocs/compta/facture/class/api_invoices.class.php
```diff
@@ -1414,6 +1415,20 @@ public function addPayment($id, $datepaye, $paymentid, $closepaidinvoices, $acco
 		$totaldeposits = $this->invoice->getSumDepositsUsed();
 		$resteapayer = price2num($this->invoice->total_ttc - $totalpaid - $totalcreditnotes - $totaldeposits, 'MT');
 
+		// hook to finalize the remaining amount, considering e.g. cash discount agreements
+		$parameters = array('$totalpaid'=>$totalpaid, '$totalcreditnotes'=>$totalcreditnotes, '$totaldeposits'=>$totaldeposits, 'remaintopay'=>$resteapayer);
+		$action = 'API_CUSTOMER_INVOICE';
```
eldy, line 1420: $action is commonly used with value
'' or 'read' for no action done, 'delete' or 'add', or 'update'.
Having its value to API_... Looks strange.
Do you really need this value. Can't we have it set to a common alteady used value ?

priojk: Thanks for the comment. $action is not needed in the hooked function, therefore I now omitted it. What I did was replicating some (bad) example I saw somewhere else in the code, where a non existing $action was created this way for a hook.

In the long run I think it makes sense to add a new member function getRemainToPay() like the existing  getSommePaiement(), getSumCreditNotesUsed() or getSumDepositsUsed() functions, since the calculation ot the remainToPay happens in many places. So all the new hooks could then go into one place within the member function and overall the code would become a bit cleaner. I would keep this in mind for the time after v18.

priojk: P.S. the travis CI problem seems to be unrelated to my code

priojk:
@eldy I think the hooks should be ok now. Is there a chance to have them added in v18?

priojk:
The Travis error seems unrelated to the PR.

eldy:
Your PR is added a hook to modify the remain to pay and replace the remain to pay calculated with 
```
		$this->invoice->totalpaid = $this->invoice->getSommePaiement();
		$this->invoice->totalcreditnotes = $this->invoice->getSumCreditNotesUsed();
		$this->invoice->totaldeposits = $this->invoice->getSumDepositsUsed();
		$this->invoice->remaintopay = price2num($this->invoice->total_ttc - $this->invoice->totalpaid - $this->invoice->totalcreditnotes - $this->invoice->totaldeposits, 'MT');
```

But this means we can modify with an external module the remain to pay globally, but what if we need to modify only the amount credit note or deposits. Also if we modify using your new hook, how to know where is part of the modification:  into the down payment, into the credit note or into the not yet paid ?
Can't you move the hook into the good of the 3 method getSommePaiement, getSumCreditNotesUsed or getSumDepositsUsed ?
Which amount among this 3 method do you need to modify ?

Daviid-P:
I agree that one of these `getSommePaiement`, `getSumCreditNotesUsed` or `getSumDepositsUsed` should be modified instead of adding hooks to all pages that call the functions.

Perhaps even a 4th function that allows to add discounts to the payment without modifying `total_ht`, `total_tva` and `total_ttc` for accounting. For cases when there's some kind of government aid where they pay a part of the total invoice, so the seller get's the full amount and the client is paying less (These aids usually end up in a specific accounting account).

Or maybe that's not a 4th function but a different kind of InvoiceLine?

And while doing this someone might consider checking https://github.com/Dolibarr/dolibarr/issues/26153

priojk:
Thanks for chiming in. I am still checking about the best way forward.

What I am trying to address here is the "time variable" end amount of the invoice: At least in Germany and in Austria it is a common business practice to formally agree an a "Skonto" while closing a deal. The Skonto agreement describes that if the invoice is paid within an agreed number of days after issuing, an agreed percentage gets deducted from the total amount of the  invoice. I am not 100% sure about the details, but I think something like this exists in France as well.

From a legal/tax perspective this acts as a "time variable" total amount **and** VAT amount of the invoice. There is no need to state any discounted amount on the invoice, except for the generic Skonto agreement as part of the payment terms (e.g. printed on the invoice: payable net within 14 days, 3% Skonto if paid within 5 days). 

And **this nevertheless does indeed affect the total_tva owed to the authorities, but not the printed VAT amount**: VAT owed to the tax authorities is automatically by the agreed Skonto percentage lower than the one printed on the invoice, if payment is made on time.

So for this scenario, an additional invoice line is unfortunately no option. From en engineering point of view, this is a mess and reminds me of stored procedures in a DB, but it is as it is and we cannot change the business practices... 

I will check, if this could be achieved by using a hook within $this->invoice->getSumCreditNotesUsed(); but it must be avoided that the Skonto discount is affecting the final, printed total amount and VAT amount, because this must definitely be stated on the invoice without the Skonto-deductions.

With the existing hooks, it was already achievable to create a Skonto-Module for most scenarios. What is currently missing are just some places/mass actions, but I will see how to address these best.

fappels:
@priojk Did you check the feature close invoice as paid partial. On close you can choose as paid before due date.

priojk:
Hi Francis, yes, I know this feature. But companies that process many invoices need more support from their ERP-system:

For instance: If I enter a payment for an invoice, the invoice gets auto-closed. With Skonto it remains in the partially paid state until someone closes it manually using the feature you mentioned.

Also entering payments is more cumbersome. The full amount can get entered with the click of a symbol. Amounts with Skonto deducted need to be entered manually etc. That's why I developed a dedicated module for a customer to assist with these things. Just a screenshot as an example: The  reduced amount can now also be entered with a click on the symbol (the one on the right side, also clearly showing the conditions). It still needs some manual choices, since companies need to be flexible with time limits and rounding "errors", i.e. larger customers sometimes stretch the rules a bit and need to get through with it. 

And even more important: With this module, the invoice gets also auto-closed like with full payments.

image

It's just that in some areas, this support is still missing (direct debit payments, some mass actions etc.) and I try to find the best way.

fappels:
Hello Joachim, thanks for clarifying. Indeed standard feature only for an occasional discount and it does not register a full payment. Success in finding the best way.

eldy:
@priojk I merge the request. However, i had to remove the part into the API because of conflicts.
May be you can repush another PR with them on a recent version of develop branch.
Note also that for hookmanager in api, you don't have anymore to init the context. A context "api" is init for all api.

PR #24683, opened 2023-05-04
Add warning on subaccount list for prospect

SylvainLegrand:
In accounting administration, if you check the list of sub-accounts, nothing tells you if a customer is a prospect.

It's a problem if a customer with unposted invoices is turned into a prospect by mistake.
The registration of invoices in accounting is not complete, exports and account balances are false.
This PR makes it possible to very quickly visualize the parameter error.

### DIFF
--- a/htdocs/accountancy/admin/subaccount.php
+++ b/htdocs/accountancy/admin/subaccount.php
@@ -130,7 +130,7 @@ llxHeader('', $title, $help_url);
 
 
 // Customer
-$sql = "SELECT sa.rowid, sa.nom as label, sa.code_compta as subaccount, '1' as type, sa.entity";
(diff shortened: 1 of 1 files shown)


htdocs/accountancy/admin/subaccount.php
```diff
@@ -393,7 +393,7 @@
 		// Subaccount label
 		if (!empty($arrayfields['label']['checked'])) {
 			print "<td>";
-			print $obj->label;
+			print $obj->nature == 2 ? '<span style = "color: red;"><b>'.$obj->label.' ('.$langs->trans("	Prospect").')</b></span>' : $obj->label;
```
eldy, line 396: Parameter into langs->trans contains a non expected space.

SylvainLegrand: Thanks, I fixed that

PR #24691, opened 2023-05-05
feat: Upgrade default PDF models to the 'new' ones (eproved since 4 dolibarr versions from now)

FHenry:
Time to remove forget azur, muscadet, crabe, einstein, new PDF model are new standard of developpement.


htdocs/install/mysql/migration/17.0.0-18.0.0.sql
```diff
@@ -411,3 +411,16 @@ CREATE TABLE llx_c_invoice_subtype (
 
 ALTER TABLE llx_c_invoice_subtype ADD UNIQUE INDEX uk_c_invoice_subtype (entity, code);
 
+-- Upgrade default PDF models to the 'new' ones (eproved since 4 dolibarr versions from now)
+UPDATE llx_const SET value="eratosthene" WHERE name="COMMANDE_ADDON_PDF" and value="einstein";
+DELETE FROM llx_document_model WHERE name="einstein" AND type="order";
```
eldy, line 416: Seems the good moment to switch to new template as default template. But do we really need to delete old template completely ? When we generate an invoice pdf using model x, the x is saved into invoice in model_pdf field. So, when we regeneate the doc automatically, the same template is used. Will this continue to work if we remove entry from table llx_document_model ?

FHenry: Normally yes because if model_pdf is empty or not found it default loaded by DOC_ADDON_PDF const

eldy: > Normally yes because if model_pdf is empty or not found it default loaded by DOC_ADDON_PDF const

But this means it will change the template with the new one, when initzial generation was using old one ? I think we should avoid this at least into a transitionnal step ( i mean not removing old template declaration).

Also can you help me solving conflict ?

FHenry: That's true, it was by design, that is a way to force this new use, but I understand that is may too pushy, I remove the delete part. But I think I forget the init part on first module activation. I'll check that

FHenry: Ok, done for the last ones

PR #24693, opened 2023-05-05
New/save planned workload service of bom in seconds

atm-lena:
# NEW|New 
When a duration is entered on the services of an MO, it is recorded in seconds. When a duration is entered on the services of a BOM, the unit changes according to the one entered. It's inconsistent.

1. We reproduce the same data processing and the same display as on an MO for the duration of the services.
2. We delete the "Unit" column which is no longer useful
3. We add a migration query to convert all existing data into seconds (only concerns services)

Next step in a next PR: give the possibility to enter seconds


eldy:
If you remove the unit in bom, when you edit a Bom to change the delay of service, the user will see for example 86400 s, instead of 1 day onscreen ? Or 24h ? How do you decide how data is rendered on screen ?

atm-lena:
@eldy  For the moment, there is not yet a configuration of working hours for a day, so we have chosen to always display the hours :
image

eldy:
Currently we are able to defined the unit when setting a BOM
image

I don't think ther eis a consistency problem:
- A BOM is a DEFINITION. Into a definition, we are able to say in which unit we enter the defintion.
- Inside the MO, we don't need the original defintion we need a value for calculation. So second seems good.

I did not find any bug (in last version) related to this. So i suppose your change is only for consistency ?

atm-lena:
@eldy 

> Currently we are able to defined the unit when setting a BOM image
> 
> I don't think ther eis a consistency problem:
> 
> * A BOM is a DEFINITION. Into a definition, we are able to say in which unit we enter the defintion.
> * Inside the MO, we don't need the original defintion we need a value for calculation. So second seems good.
> 
> I did not find any bug (in last version) related to this. So i suppose your change is only for consistency ?

Yes it's for consistency of data. In qty column we have several types of data (seconds, minutes, hours, days, months, years). We must have only seconds save.

atm-lena:
@eldy

eldy:
I don't see trouble here
A BOM is a reference document to be used later into a MO.
The reference must contains the unit because it is a "user input". If we store in second we are not able to restore the unit to user, and user must control the unit of its BOM.
However, once converted into a MO, this is converted into second because we need a "calculated" duration: value is a result of a calculation (from BOM quantity, duration and unit). So unit for output will be chosen automatically during rendering output, according to the value (according to the result of calculation, we can show result in second, hour, day...)

atm-thibaultf:
Hello @eldy ,

In fact, it's important to have consistency with units used in BOM and MO because we don't know how to convert the quantity of the BOM in the MO. For example, if in a BOM, I have a quantity of 1.2 days on a service line, how can we know how to convert this quantity to hours and seconds in MO ? We can't assume to use a 24h conversion by default because 1 day of work could be 7h, 8h, 10h...depending of the company. The problem is worst with week, month or year (how can we convert 1 year in hours and seconds ?!)

So we really need to have the same setup between BOM an MO and to be able to set the quantity of a service in hours and seconds in BOMs.

eldy:
You must make a difference between a data (duration) entered and managed by end user and a data manged and calculated by the machine.
For the first case, you must not ask a user to enter date in number of second, you lust ask end user to enter a duration using a unit and a number. So for example 2 days, 5 hours. This is a "end user" duration.

For the second one, for a culculated duration, as this is a data manged by the machine, the unit must be international unit, so seconds. It during output that we will decide in which unit we wil show it.


I don't understand why you say "we don't know how to convert a quantity into BOM into MO". For me, we can. We know the quantity and unit into BOM, that is the reference to generate the MO, so we know the duration in seconds. If we store directly in seconds in BOM, it does not solve the last problem you mention. May be we have stored seconds of working hours and may be we have stored seconds of full days.
This problem just means we are missing one data on service level. So when you setup a seevice to say it is 500 euros and service is 1 day, so 500 euros per day. Is it for working hours or for delay ?
Currently, it is implemented as "delay" when we make a conversion BOM to MO. If user need a service with duration of 1 day that is in fact only 7 hours, it is the job of end user to enter that the service is defined as 7 hours instead of 1 day.
This is why, for services, like for BOM, we also enter duration using number + a unit (it is a user data, so must be enter using this way).
The way the end user enter data does not change the way we are able to answer the question full time or working time. This trouble is currently managed by changing the reference unit for the moment, but cmsolving it differently needs another data.

PR #24694, opened 2023-05-05
CLOSE #24658 Allow custom rowid when adding timespent

### LINKED ISSUE #24658: Need ability to set rowid when adding via API
### Feature Request

The REST API uses AUTO_INCREMENT to set the rowid for new records, without allowing the caller to choose their own rowid.  It should be possible to use either method: rowid specified by caller, or AUTO_INCREMENT.  The call should fail if the rowid already exists.

### Use case

The simplest way of keeping our timesheet system in sync with Dolibarr is for both to use the same IDs for each timespent record.

### Suggested implementation

_No response_

### Suggested steps

_No response_

mc2contributor:
CLOSE #24658 Allow custom rowid when adding timespent
Allows developers to override rowid AUTO_INCREMENT value when creating any business object, though the immediate use case is for timespent records belonging to Task objects, where this option is exposed via the REST API.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -9176,7 +9176,6 @@ abstract class CommonObject
 		if (array_key_exists('fk_user_creat', $fieldvalues) && !($fieldvalues['fk_user_creat'] > 0)) {
 			$fieldvalues['fk_user_creat'] = $user->id;
 		}
-		unset($fieldvalues['rowid']); // The field 'rowid' is reserved field name for autoincrement field so we don't need it into insert.
 		if (array_key_exists('ref', $fieldvalues)) {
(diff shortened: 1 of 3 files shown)


eldy:
What is goal of being able to force the rowid during insert ?

mc2contributor:
The goal is to make it easier to keep Dolibarr in sync with our external timesheet system.  When a time record is added to our external system, we want to use the same id in Dolibarr, so that we don't need an additional table to track the correlation between the two systems.  (We've configured Dolibarr not to allow editing of time records in the UI, making that part read-only for now.)

eldy:
Having the ID forced by an external system may generate troubles, during update, backup, import or maintenance actions.
For such need, you can use the field "ref_ext" in table llx_element_time (ll_project_task_time in old version).

The risk of collision if a second module try also to force the id with its own value is the same then forcing the ID yourself, but the risk of collision with dolibarr core feature and maintenance and import operations become zero.

PR #24697, opened 2023-05-05
FIX | Prevent adding Free line when "MAIN_DISABLE_FREE_LINES" activated

NextGestion:
FIX | Prevent adding Free line when the option"**MAIN_DISABLE_FREE_LINES**" is activated (Object line in Proposal, invoice, order, contract)

image
image


NextGestion:
Hello @eldy ,

Can you provide us more details on this "**failing check**" ?

eldy:
> Hello @eldy ,
> 
> Can you provide us more details on this "**failing check**" ?

Click on the error. Then click on "the build", then View Log, you should get more info at bottom of log...

NextGestion:
Hello @eldy ,

Ok thanks, we will continue debugging on this.

For your information, the **same problem** exist for these last **Pull Requestes** : 
image
image

Thank you for your time

NextGestion:
Hello @eldy ,

We have fixed the issues. 

Can you check this **PR**, and see if it can be merged?

Thanks

htdocs/fourn/facture/card-rec.php
```diff
@@ -428,8 +428,13 @@
 			$tva_tx = (GETPOST('tva_tx', 'alpha') ? GETPOST('tva_tx', 'alpha') : 0);
 			$ref_fournisseur = (GETPOSTISSET('fourn_ref') ? GETPOST('fourn_ref', 'restricthtml') : '');
 		} else {
-			$idprod = GETPOST('idprod', 'int');
+			$idprod = GETPOST('idprodfournprice', 'int');
```
eldy, line 431: Note that idprodfournprice may be
- an int when it is a supplier price id
- a string 'idprod_xxx' where xxx is the id of product (when there is no price defined on the product).

So if you make the GETPOST with int param, for the case idprod_xxx, you will get zero. 
Few lines later there is a regex that is used so we got into idprod an int for both cases.
Currently the GETPOST is not used as it is overwritten later but with your change it is used bybthe new if so you should take into account the case value is idprod_xxx

NextGestion:
Hello @eldy ,

Ok, we canceled this modification.

Can you check this **PR**, and see if it can be merged?

Thanks

PR #24698, opened 2023-05-05
FIX: Inventory lines overwrite when changing page

bomuux:
# FIX
When inventory lines are displayed on several pages, they are updated at each page change from the page selector.

The reason is that the form is always posted with action=updateinventorylines assuming that only the Save button post the form : this is not the case, changing page also post the form (javascript) thus wrongly applying an update, sometimes resulting in data overwriting.

This is a quick fix, the problem will reappear if one add more action (input) buttons to the form.

### DIFF
--- a/htdocs/accountancy/admin/subaccount.php
+++ b/htdocs/accountancy/admin/subaccount.php
@@ -130,7 +130,7 @@ llxHeader('', $title, $help_url);
 
 
 // Customer
-$sql = "SELECT sa.rowid, sa.nom as label, sa.code_compta as subaccount, '1' as type, sa.entity";
+$sql = "SELECT sa.rowid, sa.nom as label, sa.code_compta as subaccount, '1' as type, sa.entity, sa.client as nature";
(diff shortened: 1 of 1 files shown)


htdocs/product/inventory/inventory.php
```diff
@@ -88,6 +88,11 @@
 	$action = 'view';
 }
 
+// Prevent updating lines when not posted by SAVE input button
+if (!GETPOSTISSET('save') && $action == 'updateinventorylines') {
```
eldy, line 92: So this means that:
You edit the first page of inventory by changing quantities, you switch tonthe next page to edit the quantities for next page and you loose all your change of previous page if you didn't click the save before changing page ?

bomuux: No, it is not the problem. And btw my PR is not the best answer to it - it corrects the bug but have to click on save button each time before changing page, which i thought was the normal behaviour :(

I close this PR and submitted another: #24709

PR #24703, opened 2023-05-07
NEW test/phpunit: skip tests instead of die() when requirements are not met

alexandre-janniaux:
# NEW test/phpunit: skip tests instead of die() when requirements are not met

The whole MR revolves around making tests that require modules that are not present fallible instead of die() the whole testsuite at construct time.

It also adds the --fail-on-skipped flag in the CI to ensure the behaviour stays the same and the CI coverage stays the largest possible, and fixes a few inconsistencies in the testsuite in some of the modified files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -479,7 +479,7 @@ script:
   echo "Unit testing"
   # Ensure we catch errors. Set this to +e instead of -e if you want to go to the end to see dolibarr.log file.
   set -e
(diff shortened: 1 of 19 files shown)


alexandre-janniaux:
Seems like --fail-on-skipped doesn't work before 9.4.3... https://github.com/sebastianbergmann/phpunit/commit/150b727fe6bffb8e58360ca72057f7807a302e13

eldy:
Due to a lot of other changes done on several PR, this one is deprecated or at least no more mergeable, so i close it.

PR #24706, opened 2023-05-07
FIX test/phpunit: add $name to __construct()

alexandre-janniaux:
# FIX test/phpunit: add $name to __construct()

Fix the following errors when running recent phpunit:

    Message:  Too few arguments to function PHPUnit\Framework\TestCase::__construct(), 0 passed in
    dolibarr/test/phpunit/DateLibTzFranceTest.php on line 63 and exactly
    1 expected Location: phar:///usr/share/webapps/bin/phpunit.phar/phpunit/Framework/TestCase.php:265

Indeed, the old constructor had an optional $name='' parameter but the new constructor reads like this:

    public function __construct(string $name)

and the parameter is now mandatory.


alexandre-janniaux:
Fixed the modulebuilder test also.

PR #24717, opened 2023-05-09
New : Dispatch card on expedition object

Hystepik:
Add a new card on expedition object to permit a better edition of the object, like the dispatch.php page that is on object fourn/commande

image

Expedition needs to be on draft to edit


htdocs/expedition/class/expedition.class.php
```diff
@@ -979,7 +982,7 @@ public function addline_batch($dbatch, $array_options = 0)
 					$linebatch->qty = $value['q'];
 					$tab[] = $linebatch;
 
-					if ($conf->global->STOCK_MUST_BE_ENOUGH_FOR_SHIPMENT) {
+					if (!empty(getDolUserString("STOCK_MUST_BE_ENOUGH_FOR_SHIPMENT", '0'))) {
```
eldy, line 985: No need for empty around getDolUserString

PR #24719, opened 2023-05-09
FIX broken feature : send private message by email

BenjaminFlr:


BenjaminFlr:
@eldy : Blocking evolution prevent users from sending a private message as email from a ticket.

eldy:

BenjaminFlr:
When we click on "Add a private message" on a ticket's card, we can't select both "Private message" and "Send by mail". We can only select one of the two.

hregis:
@eldy add a constant TICKET_SEND_PRIVATE_EMAIL to have possibility to check "private message" and "send email" for the action code "TICKET_MSG_PRIVATE_SENTBYMAIL", only send email for the internal contact linked with a ticket

PR #24736, opened 2023-05-10
end accounting period

MaximilienR-easya:
# Fix
the end accounting period should not be the fiscal month start but the month just before it


aspangaro:
Thanks @MaximilienR-easya , it's an old bug never fixed

PR #24739, opened 2023-05-11
FIX: Task events not displayed

noec764:
Task `elementtype` is 'project_task' and not 'task' and in `llx_projet_task` there is no column name fk_soc so thirdparty is never registered when project_task event is registered.

Without those elements, getAction sql query did not fetch well task's events.

### DIFF
--- a/htdocs/projet/tasks/task.php
+++ b/htdocs/projet/tasks/task.php
@@ -644,8 +644,7 @@ if ($id > 0 || !empty($ref)) {
 		// List of actions on element
 		include_once DOL_DOCUMENT_ROOT.'/core/class/html.formactions.class.php';
 		$formactions = new FormActions($db);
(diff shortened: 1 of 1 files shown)


htdocs/projet/tasks/task.php
```diff
@@ -644,8 +644,7 @@
 		// List of actions on element
 		include_once DOL_DOCUMENT_ROOT.'/core/class/html.formactions.class.php';
 		$formactions = new FormActions($db);
-		$defaultthirdpartyid = $socid > 0 ? $socid : $object->project->socid;
-		$formactions->showactions($object, 'task', $defaultthirdpartyid, 1, '', 10, 'withproject='.$withproject);
+		$formactions->showactions($object, 'project_task', 0, 1, '', 10, 'withproject='.$withproject);
```
eldy, line 647: 2nd param was changed in project_taks. OK
But third was changed into 0. Is it done on purpose ?

PR #24740, opened 2023-05-11
can use isModEnabled("module") in extrafield enabled

frederic34:
if you create an extrafield with enabled property like `isModEnabled("module")` it is broken on extrafield edit

### DIFF
--- a/htdocs/core/actions_extrafields.inc.php
+++ b/htdocs/core/actions_extrafields.inc.php
@@ -367,7 +367,7 @@ if ($action == 'update') {
 					$computedvalue,
 					(GETPOST('entitycurrentorall', 'alpha') ? 0 : ''),
 					GETPOST('langfile'),
-					GETPOST('enabled', 'restricthtml'),
+					GETPOST('enabled', 'nohtml'),
 					(GETPOST('totalizable', 'alpha') ? 1 : 0),
 					GETPOST('printable', 'alpha'),
 					array('css' => $css, 'cssview' => $cssview, 'csslist' => $csslist)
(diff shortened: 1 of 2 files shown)


htdocs/core/tpl/admin_extrafields_edit.tpl.php
```diff
@@ -143,7 +143,7 @@ function init_typeoffields(type)
 <input type="hidden" name="attrname" value="<?php echo $attrname; ?>">
 <input type="hidden" name="action" value="update">
 <input type="hidden" name="rowid" value="<?php echo (empty($rowid) ? '' : $rowid) ?>">
-<input type="hidden" name="enabled" value="<?php echo $extrafields->attributes[$elementtype]['enabled'][$attrname]; ?>">
+<input type="hidden" name="enabled" value="<?php echo htmlspecialchars($extrafields->attributes[$elementtype]['enabled'][$attrname]); ?>">
```
eldy, line 146: We should not use htmlspecialchars
Rest of dolibarr code is using 
`dol_escape_htmltag($extrafields->attributes[$elementtype]['enabled'][$attrname]) `
Does it work for you with this ?

htdocs/core/actions_extrafields.inc.php
```diff
@@ -367,7 +367,7 @@
 					$computedvalue,
 					(GETPOST('entitycurrentorall', 'alpha') ? 0 : ''),
 					GETPOST('langfile'),
-					GETPOST('enabled', 'alpha'),
+					GETPOST('enabled', 'restricthtml'),
```
eldy, line 370: Seems not enough protected. Can you try with 'nohtml' instead of 'restricthtml' ?

frederic34: I made test, it's ok

PR #24746, opened 2023-05-11
fix: html better than text because product label can contain html

FHenry:
fix: html better than text because product label can contain html


htdocs/takepos/index.php
```diff
@@ -335,7 +335,7 @@ function LoadProducts(position, issubcat) {
 				var titlestring = <?php echo $titlestring; ?>;
 				<?php if (!$conf->global->TAKEPOS_HIDE_PRODUCT_IMAGES) {
 					echo '$("#prodivdesc"+ishow).show();';
-					echo '$("#prodesc"+ishow).text(data[parseInt(idata)][\'label\']);';
+					echo '$("#prodesc"+ishow).html(data[parseInt(idata)][\'label\']);';
```
eldy, line 338: Label of product is a short string to quickly identify the name. Currently it is like a ref. I am not sure the label is a good place so store rich content ?

PR #24747, opened 2023-05-11
FIX: Dolibarrize DolibarrModules::insert_cronjobs() method

noec764:
In V17 I noticed that **datenextrun** is a mandatory field for Cron jobs to run for the first time, so I checked and it was not possible to add this field with the module descriptor because `DolibarrModules::insert_cronjobs()` does not take this field into account when creating a cron job via its SQL query. So I decided to remove this SQL query and use the `cronjob::create()` method for better handling of all fields.


noec764:
The change applied by the merge of 17.0 can be problematic. 
```php
//$sql .= " WHERE module_name = '" . $this->db->escape(empty($this->rights_class) ? strtolower($this->name) : $this->rights_class) . "'";
$sql .= " WHERE label = '".$this->db->escape($label)."'";
```
Indeed as label can be set by external modules, some problems can occurs with accents for example. 
I think it is safer to compare with module / class / method than label

noec764:
@eldy What should I do about this issue ?

noec764:
@eldy Can you check this PR again

PR #24759, opened 2023-05-13
add errors-to for swiftmailer

frederic34:


htdocs/core/class/CMailFile.class.php
```diff
@@ -557,6 +557,14 @@ public function __construct($subject, $to, $from, $msg, $filename_list = array()
 				}
 			}
 
+			if (!empty($this->errors_to)) {
+				try {
+					$this->message->setReturnPath($this->getArrayAddress($this->errors_to));
```
eldy, line 562: It seems you force return path to errorsto ?
Errorsto is email when error appears.
Return path is email to send email to when user make reply, so should be different.
Return path is an optionnal field.
If not defined, the reply to is used i think, so do we really need no to force it to a errorsto email (replyto seems a better replacement for returnpath) ?

PR #24760, opened 2023-05-13
NEW Can restore product in stock when deleting the supplier invoice

NextGestion:
# FIX| Products not removed from stock when deleting the supplier invoice

1- https://www.dolibarr.fr/forum/t/produits-non-mouvemente-lors-de-la-suppression-de-la-facture-fournisseur/43208

2- https://www.dolibarr.fr/forum/t/probleme-de-mouvement-stock-lorsque-je-repasse-la-facture-fournisseur-en-brouillant-ou-je-la-supprime-avec-la-quantite-du-stock-produit-zero/43262


NextGestion:
Hello @eldy,

Can you check this **PR**, and see if it can be merged?

Thank you

htdocs/fourn/class/fournisseur.facture.class.php
```diff
@@ -1848,6 +1848,11 @@ public function validate($user, $force_number = '', $idwarehouse = 0, $notrigger
 							$result = $mouvP->reception($user, $this->lines[$i]->fk_product, $idwarehouse, $this->lines[$i]->qty, $up_ht_disc, $langs->trans("InvoiceValidatedInDolibarr", $num));
 						}
 						if ($result < 0) {
+							if (count($mouvP->errors)) {
+								setEventMessages(null, $mouvP->errors, 'errors');
```
eldy, line 1852: We are inside a method than can be called out of user interface context, for example by a batch.
So we should not have setEventMessage inside such method.
Instead, you must set error message inside
this->error
And return a negative value
Then the caller canndecide to use setEvent... Or doing something else depending on user context.

htdocs/user/card.php
```diff
@@ -772,7 +772,19 @@
 	$formproduct = new FormProduct($db);
 }
 
-if ($object->id > 0) {
+if (GETPOSTINT('id') > 0) {
```
eldy, line 775: The fetch is using $id so the test should be on $id

NextGestion: Hello @eldy,

We forced to use "**GETPOSTINT('id')**" instead of "**$id**".

Because there is a test at top that checks if the "**$id**" is empty :

`if (empty($id)) {`
	`$id = $user->id;`
`}`

image

The condition "($id > 0)" will be always true. So, when we create a user, the page title will be the name of the current user and this is wrong.

We have changed the structure now. Could you check it from your side?

Best regards

htdocs/fourn/class/fournisseur.facture.class.php
NextGestion: Hello @eldy ,

We have taken your comments into consideration.

Can you check this **PR**, and see if it can be merged?

Thanks

NextGestion:
Hello @eldy ,

Can you check this PR, and see if it can be merged?

Thanks

eldy:
The PR contains code change related to stock modification when deleting a shipment(such behaviour is not always expected). When we want to revert stock, shipment is canceled or set back to draft. When we delzte, we do it for purge purpose. In such a case, stock modification is not expected.
I suggest to add a question during action to ask if stock must be reverted or not.

The PR contains also changes into user card not related to this.

Merging PR that contains 2 different purposes is not possible. Only atomic PR can be merged. Thanks to split into different PR.

josett225:
Hi @eldy 
As this seems to be a normal behavior, do you think this could be done on all the version below Develop?
Thanks

NextGestion:
Hello @eldy,

We have completed processing your feedback.

Can you check this **PR**, and see if it can be merged?

Thank you

NextGestion:
Hello @eldy,

Can you check this PR [#24760](https://github.com/Dolibarr/dolibarr/pull/24760) and see if it can be merged?

Thank you

PR #24764, opened 2023-05-14
fix : Warning: A non-numeric value encountered in /home/httpd/vhosts/…

grandoc:
…aflac.fr/domains/dev.aflac.fr/httpdocs/core/lib/functions.lib.php on line 2713

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -2562,7 +2562,15 @@ function dol_print_date($time, $format = '', $tzoutput = 'auto', $outputlangs = 
 
 	// Clean parameters
 	$to_gmt = false;
-	$offsettz = $offsetdst = 0;
+	if (!is_numeric($time)) {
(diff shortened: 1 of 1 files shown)


eldy:
Don't understand this fix. Can you paste full error message you have ?

grandoc:
It was the full message. I copied-passed it as it was

htdocs/core/lib/functions.lib.php
```diff
@@ -2562,7 +2562,15 @@ function dol_print_date($time, $format = '', $tzoutput = 'auto', $outputlangs =
 
 	// Clean parameters
 	$to_gmt = false;
-	$offsettz = $offsetdst = 0;
+	if (!is_numeric($time)) {
+		$time = 0;
+	}
+	if (!is_numeric($offsettz)) {
```
eldy, line 2568: This test seems useless because var is never defined yet here.

PR #24776, opened 2023-05-15
FIX : No HTML interpretation in ODT PDF models

atm-adrien:
## FIX : No HTML interpretation in ODT PDF models

There is currently no interpretation of HTML in ODT type document templates.
I'm not really sure if the method I'm using to fix this is the correct one. If not, do you have any idea how to fix it?

### DIFF
--- a/htdocs/includes/odtphp/odf.php
+++ b/htdocs/includes/odtphp/odf.php
@@ -161,12 +161,13 @@ class Odf
 	 */
 	public function convertVarToOdf($value, $encode = true, $charset = 'ISO-8859')
 	{
+		$value = dol_html_entity_decode($value, ENT_COMPAT);
 		$value = $encode ? htmlspecialchars($value) : $value;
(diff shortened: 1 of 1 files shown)


eldy:
Can you provide an example of value you have in entry and the expected value expected as output ?

atm-adrien:
Here is, the object line : 
(we expect this on the ODT PDF)
image

And here is the object line in the ODT PDF :
image

atm-adrien:
@eldy is it okay for you ?

atm-adrien:
@eldy

eldy:
My question was 
Can you provide an example of value you have in entry OF THE FUNCTION convertVarToOdf and the expected value expected as output of the function ?
It seems you fix a bug in this method, so this means there is currently an output return that does not match the zxpected output. So my question is what is this entry value. We need this to add a Unit test to guarantee the fix correct, definitive, and bug will not reappear when updating the lib.

PR #24777, opened 2023-05-15
FIX mask next value sql string for external modules

altairis-tof:
get real next value upon mask in case of ref is completed by external module in database

### DIFF
--- a/htdocs/core/lib/functions2.lib.php
+++ b/htdocs/core/lib/functions2.lib.php
@@ -1340,7 +1340,7 @@ function get_next_value($db, $mask, $table, $field, $where = '', $objsoc = '', $
 		$ref = '';
 		$sql = "SELECT ".$field." as ref";
 		$sql .= " FROM ".MAIN_DB_PREFIX.$table;
-		$sql .= " WHERE ".$field." LIKE '".$db->escape($maskLike)."'";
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions2.lib.php
```diff
@@ -1280,7 +1280,7 @@ function get_next_value($db, $mask, $table, $field, $where = '', $objsoc = '', $
 	$counter = 0;
 	$sql = "SELECT MAX(".$sqlstring.") as val";
 	$sql .= " FROM ".MAIN_DB_PREFIX.$table;
-	$sql .= " WHERE ".$field." LIKE '".$db->escape($maskLike)."'";
+	$sql .= " WHERE ".$field." LIKE '".$db->escape($maskLike)."%'";
```
eldy, line 1283: If user change its numbering rule, for example adding this will inteoduce a bug because search may find string that should not match when exact search is done.
So i recommand to add a constant SEARCH_FOR_NEXT_VAL_ON_START_ONLY to add this.
Also this is not a bug into v15, it is an enhancement to allow external module needs so thus must be pushed into develop.

altairis-tof:
replaced by https://github.com/Dolibarr/dolibarr/pull/24781

PR #24788, opened 2023-05-16
New attributes on supplier proposal card

KiVoilaJe:
Edit attributes on supplier proposal card

Modification by changing the table when multicurrency is on, with the amounts, passing from 8 lines to 5.
example :
Amount HT 1,00€ $1,00


KiVoilaJe:
Capture d’écran du 2023-05-17 00-10-42

PR #24792, opened 2023-05-17
fix :  entity field missing in llx_paiementcharge

atm-jpb:
the  checkUserAccessToObject function check if multicompany is activated and handle with entity field in llx_paiementcharge table. when some customer have this module activated,  the function return false and finally  
the restrictedArea method too (sql error ).

i suggest to add this field in 17.0 version

### DIFF
--- a/htdocs/install/mysql/migration/16.0.0-17.0.0.sql
+++ b/htdocs/install/mysql/migration/16.0.0-17.0.0.sql
@@ -65,6 +65,7 @@ ALTER TABLE llx_overwrite_trans ADD UNIQUE INDEX uk_overwrite_trans(lang, transk
 ALTER TABLE llx_mailing_cibles MODIFY COLUMN source_type varchar(32); 
 
 ALTER TABLE llx_actioncomm ADD INDEX idx_actioncomm_percent (percent);
+ALTER TABLE llx_paiementcharge ADD COLUMN entity integer NOT NULL DEFAULT 1 AFTER rowid;
 
 UPDATE llx_c_paiement SET code = 'BANCON' WHERE code = 'BAN' AND libelle = 'Bancontact';
 
--- a/htdocs/install/mysql/tables/llx_paiementcharge.sql
+++ b/htdocs/install/mysql/tables/llx_paiementcharge.sql
@@ -20,6 +20,7 @@ create table llx_paiementcharge
 (
   rowid           integer AUTO_INCREMENT PRIMARY KEY,
   fk_charge       integer,
+  entity		  integer DEFAULT 1,
   datec           datetime,           -- date de creation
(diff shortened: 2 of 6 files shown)


htdocs/install/mysql/tables/llx_paiementcharge.sql
```diff
@@ -19,7 +19,9 @@
 create table llx_paiementcharge
 (
   rowid           integer AUTO_INCREMENT PRIMARY KEY,
+  entity		  integer DEFAULT 1,
```
eldy, line 22: We should not need a field entity here, because a payment of a social contribution is linked to a social contribution.
And the social contribution is already inside one company (so the field entity is already inside the table llx_chargesociales)
And if a social contribution is inside entity X, the payment must also be into entity X. So we are alreay able to know the entity of a payement using the entity of its parent social contribution.

atm-jpb:
i have 2 solutions : 

1 ) i can possibly add a param to  checkUserAccessToObject to avoid  the check entity
2) create a specific test on paiement_sc   in  checkUserAccessToObject (must create a sql join on chargesociales table.

Can you  propose  your view  if this two proposals  are not in range ?

atm-jpb:
@eldy

atm-greg:
@eldy You didn't answer to my teammate

atm-jpb:
@eldy

eldy:
No database change is allowed into an already released and migrated version.

Also the payment of a social controbution is a child of social contribution, its entity is the entity if its parent.
So for a payment of a social contribution, we must check entity of social contribution.

atm-quentin:
i made an other PR https://github.com/Dolibarr/dolibarr/pull/26034 is it ok for you with this ?

htdocs/install/mysql/migration/16.0.0-17.0.0.sql
```diff
@@ -65,6 +65,7 @@ ALTER TABLE llx_overwrite_trans ADD UNIQUE INDEX uk_overwrite_trans(lang, transk
 ALTER TABLE llx_mailing_cibles MODIFY COLUMN source_type varchar(32); 
 
 ALTER TABLE llx_actioncomm ADD INDEX idx_actioncomm_percent (percent);
+ALTER TABLE llx_paiementcharge ADD COLUMN entity integer NOT NULL DEFAULT 1 AFTER rowid;
```
eldy, line 68: No database structure change is allowed into an already released version. We can return in the past. What was released is already released.
Also there is no reason to have entity field on a payment table. A payment is always in entity of its parent object.

atm-quentin: please check https://github.com/Dolibarr/dolibarr/pull/26034

PR #24800, opened 2023-05-17
NEW - Hook formmail

BB2A-Anthony:
# NEW|New [*Hook formmail*]
[*Long description*]


htdocs/core/tpl/card_presend.tpl.php
```diff
@@ -24,7 +24,7 @@
  *
  * $trackid must be defined
  * $modelmail
- * $defaulttopic
+ * $defaulttopicFromEmail
```
eldy, line 27: Why renaming old var name in comment ?

htdocs/core/tpl/card_presend.tpl.php
```diff
@@ -383,6 +383,12 @@
 	$formmail->param['returnurl'] = $_SERVER["PHP_SELF"].'?id='.$object->id;
 	$formmail->param['fileinit'] = array($file);
 
+	$parameters = array();
+	$reshook = $hookmanager->executeHooks('FromEmail', $parameters, $formmail);
```
eldy, line 387: It seems you already have the getFormMail hook inside the 
$formmail->get_form();
called 2 lines after.

PR #24801, opened 2023-05-17
fix : Warning: Undefined array key 186 in /home/httpd/vhosts/aflac.fr…

grandoc:
…/domains/dev.aflac.fr/httpdocs/compta/paiement/class/paiement.class.php on line 279

### DIFF
--- a/htdocs/compta/paiement/class/paiement.class.php
+++ b/htdocs/compta/paiement/class/paiement.class.php
@@ -289,10 +289,10 @@ class Paiement extends CommonObject
 				$this->error = 'ErrorYouTryToPayInvoicesWithDifferentCurrenciesInSamePayment';
 				return -1;
 			}
-		
+
 			$totalamount_converted += $value_converted;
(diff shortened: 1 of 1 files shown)


htdocs/compta/paiement/class/paiement.class.php
```diff
@@ -276,17 +276,23 @@ public function create($user, $closepaidinvoices = 0, $thirdparty = null)
 				return -1;
 			}
 			if (empty($currencyofpayment)) {
-				$currencyofpayment = $this->multicurrency_code[$key];
+				if (array_key_exists($key, $this->multicurrency_code)) {
+					$currencyofpayment = $this->multicurrency_code[$key];
+				} else {
+					// Handle the case where the key does not exist in the array
```
eldy, line 282: If $this->multicurrency_code[$key] is not defined, then currencyofpayment is set to null with old code and continue.
With new one, the code is finished.

Code must be similar after and before.

PR #24802, opened 2023-05-17
NEW : bigger size for display of customer name

altairis-tof:
This is for better view of current customer
image


altairis-tof:
i close this one also; i will manage it in specific css

PR #24803, opened 2023-05-17
FIX : fix hook return test

altairis-tof:
better test after hook

### DIFF
--- a/htdocs/takepos/invoice.php
+++ b/htdocs/takepos/invoice.php
@@ -564,7 +564,7 @@ if (empty($reshook)) {
 			if ($reshook < 0) setEventMessages($hookmanager->error, $hookmanager->errors, 'errors');
 
 
-			if (!empty($reshook)) {
+			if (!empty($hookmanager->resArray)) {
 				$line = $hookmanager->resArray;
 			}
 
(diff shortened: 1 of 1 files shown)


htdocs/takepos/invoice.php
```diff
@@ -564,14 +564,12 @@ function fail($message)
 			if ($reshook < 0) setEventMessages($hookmanager->error, $hookmanager->errors, 'errors');
 
 
-			if (empty($reshook)) {
-				if (!empty($hookmanager->resArray)) {
-					$line = $hookmanager->resArray;
-				}
-
-				$idoflineadded = $invoice->addline($line['description'], $line['price'], $qty, $line['tva_tx'], $line['localtax1_tx'], $line['localtax2_tx'], $idproduct, $line['remise_percent'], '', 0, 0, 0, '', $price_base_type, $line['price_ttc'], $prod->type, -1, 0, '', 0, (!empty($parent_line)) ? $parent_line : '', $line['fk_fournprice'], $line['pa_ht'], '', $line['array_options'], 100, '', null, 0);
+			if (!empty($hookmanager->resArray)) {
```
eldy, line 567: We must always test on 
`if (empty($reshook)) {`
Because $reshook says if we must replace standard code (standard code disabled) or not (executed).

altairis-tof:
sorry @eldy i made an error; i close

PR #24805, opened 2023-05-17
NEW : TakePOS - differenciate invoice discount and line discount

altairis-tof:
it is better for users to have two different labels for invoice and lines discount

### DIFF
--- a/htdocs/takepos/index.php
+++ b/htdocs/takepos/index.php
@@ -202,7 +202,7 @@ function ClearSearch() {
 	$("#search").val('');
 	$("#qty").html("<?php echo $langs->trans("Qty"); ?>").removeClass('clicked');
 	$("#price").html("<?php echo $langs->trans("Price"); ?>").removeClass('clicked');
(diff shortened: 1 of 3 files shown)


htdocs/takepos/index.php
```diff
@@ -202,7 +202,7 @@ function ClearSearch() {
 	$("#search").val('');
 	$("#qty").html("<?php echo $langs->trans("Qty"); ?>").removeClass('clicked');
 	$("#price").html("<?php echo $langs->trans("Price"); ?>").removeClass('clicked');
-	$("#reduction").html("<?php echo $langs->trans("ReductionShort"); ?>").removeClass('clicked');
+	$("#reduction").html("<?php echo $langs->trans("LineDiscount"); ?>").removeClass('clicked');
```
eldy, line 205: We have 2 keys for discount label
Reduction
ReductionShort
The ReductionShort is used for the case we have very few spaces. For example in english, it is translated with "Disc."
For responsive reability, we must keep the use of a key and a keyShort when we know space is very low (in title of tab columns for example)

altairis-tof: ok i have put the short translations; ok for me.

PR #24828, opened 2023-05-19
NEW : TakePOS - force to choose a customer on new sale

altairis-tof:


htdocs/takepos/invoice.php
```diff
@@ -1143,6 +1143,8 @@ function SetNote() {
 	$s = $langs->trans("Customer");
 	if ($invoice->id > 0 && ($invoice->socid != getDolGlobalString($constforcompanyid))) {
 		$s = $soc->name;
+	} else {
+		print '$("#idcustomer").val("");';
```
eldy, line 1147: In most use cases for POS, the customer is a generic customer so a customer not modified, for such use case, removing selectedx customer seems to add a new step to select the customer if we reset it ?

altairis-tof: ok with you , i introduce now a new constant TAKEPOS_FORCE_CUSTOMER_SELECT

goldenwt:
I have setup TakePOS and we are very much okay.

However, we get invoice drafts from it.

When you select a customer name, upon making Product selection or edit price or qty, the selected customer name defaults back to the default customer name selected during setup in the TakePOS cpnfiguration.

The system ends up recording the default customer name selected as against the real customer name for the transaction(It creates the original customer transaction as draft, while it captures the default customer entry as validated invoice)

Dont know what to do about it.

Could it be a bug?

altairis-tof:
> I have setup TakePOS and we are very much okay.
> 
> However, we get invoice drafts from it.
> 
> When you select a customer name, upon making Product selection or edit price or qty, the selected customer name defaults back to the default customer name selected during setup in the TakePOS cpnfiguration.
> 
> The system ends up recording the default customer name selected as against the real customer name for the transaction(It creates the original customer transaction as draft, while it captures the default customer entry as validated invoice)
> 
> Dont know what to do about it.
> 
> Could it be a bug?

hello @goldenwt , does your comment report with this PR ?

htdocs/takepos/invoice.php
eldy: When in default use case, when
`if ($invoice->id > 0 && ($invoice->socid != getDolGlobalString($constforcompanyid))) {`
was true, we did the print and then execut line 1275.
After your PR, code on line 1275 is no more executed for the default use case. Should be executed with default use case, like in old version, right ?

altairis-tof: ok @eldy i restore the default use case.

PR #24830, opened 2023-05-19
NEW : TakePOS - add constant to define payment terms if payment delayed

altairis-tof:
NEW add TAKEPOS_DELAYED_TERMS constant to define payment term id for delayed payments

### DIFF
--- a/htdocs/takepos/invoice.php
+++ b/htdocs/takepos/invoice.php
@@ -285,7 +285,7 @@ if (empty($reshook)) {
 					$payment->create($user);
 					$payment->addPaymentToBank($user, 'payment', '(CustomerInvoicePayment)', $bankaccount, '', '');
 					$remaintopay = $invoice->getRemainToPay(); // Recalculate remain to pay after the payment is recorded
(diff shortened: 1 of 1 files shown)


htdocs/takepos/invoice.php
```diff
@@ -285,6 +285,8 @@ function fail($message)
 					$payment->create($user);
 					$payment->addPaymentToBank($user, 'payment', '(CustomerInvoicePayment)', $bankaccount, '', '');
 					$remaintopay = $invoice->getRemainToPay(); // Recalculate remain to pay after the payment is recorded
+				} elseif (!empty(getDolGlobalInt("TAKEPOS_DELAYED_TERMS"))) {
```
eldy, line 288: `elseif (!empty(getDolGlobalInt("TAKEPOS_DELAYED_TERMS"))) {`
must be replaced with
`elseif (getDolGlobalInt("TAKEPOS_DELAYED_TERMS")) {`
empty is not allowed of the result of a function

altairis-tof: @eldy fixed

PR #24832, opened 2023-05-19
NEW:  TakePOS - manage permissions on terminals

altairis-tof:


htdocs/takepos/index.php
```diff
@@ -54,11 +54,26 @@
 $setterminal = GETPOST('setterminal', 'int');
 $setcurrency = GETPOST('setcurrency', 'aZ09');
 
+// get user authorized terminals
+$nb_auth_terms = 0;
+$numterminals = max(1, getDolGlobalInt("TAKEPOS_NUM_TERMINALS"));
+for ($i = 1; $i <= $numterminals; $i++) {
+	if ($user->rights->takepos->{'access_takepos_' . $i}) {
```
eldy, line 61: Use user->hasRight to be compatible with php8.2

altairis-tof:
@eldy is it possible to merge this PR that allow to define permissions on TakePos terminal ? this is a feature often asked by users (last example : https://www.dolibarr.fr/forum/t/attribuer-une-caisse-a-un-utilisateur-par-defaut/45456)

eldy:
On your PR, i have no permission to update the base branch. So no way for me to solve conflict or fix CTI (don't know where is the difference between other contributors, where it is possible to fix a PR manually from github).

altairis-tof:
> On your PR, i have no permission to update the base branch. So no way for me to solve conflict or fix CTI (don't know where is the difference between other contributors, where it is possible to fix a PR manually from github).

@eldy i have just invited you to have wirte access on the fork; tell me if it is ok please.

htdocs/takepos/admin/setup.php
```diff
@@ -95,6 +95,13 @@
 	}
 
 	if (!$error) {
+		// add terminal permissions
+		require_once DOL_DOCUMENT_ROOT . '/core/modules/modTakePos.class.php';
```
eldy, line 99: Using 1 permission to assign a different user on each terminal is not always the desired behaviour. Small shops would prefer to have one global permission as currently. So can you introduce first a hidden option to switch in mode 1 permission per terminal. 
When stable we will be able later to offer the option on takepos setup...

eldy (changes requested):
Using 1 permission to assign a different user on each terminal is not always the desired behaviour. Small shops would prefer to have one global permission as currently. So can you introduce first a hidden option to switch in mode "1 permission per terminal".
When stable we will be able later to offer the option on takepos setup...

htdocs/takepos/admin/setup.php
altairis-tof: ok @eldy i introduce TAKEPOS_USE_TERMINAL_PERMISSIONS

eldy:

altairis-tof:
dommage

PR #24843, opened 2023-05-22
FIX svg image not displaying with Chromium (Chrome, Brave, etc...)

hregis:


eldy:
I can't reproduce the problem. Can you describe how to reproduce.
Note that image of company is also used inside pdf and odt, so file format is limited to png jpg and jpeg for this reason.
Are you speaking about the image set into home - setup - company ? The squarre image or the other one ?

hregis:
@eldy it concerned the default image "dolibarr_logo.svg" on a fresh install, I replace it with my own logo in svg, but apparently it was my image that was not converted correctly! sorry

PR #24846, opened 2023-05-22
NEW -  Add terms of sale on documents

BB2A-Anthony:
# NEW|New [*Add terms of sale on documents*]
[*Long description*]


eldy:
Can you provide screenshots to show what the pr brings....

BB2A-Anthony:
> 
@eldy 
not screenshots after the doc PDF

attached

[test_PR2302-00001.pdf](https://github.com/Dolibarr/dolibarr/files/11572803/test_PR2302-00001.pdf)

eldy:
I see.
However i think the setup page to upload the pdf term condition, should be on the menu 
Setup - pdf - other and not into menu - company. We must avoid to have the setup of 1 feature splitted into 2 different pages.
Can you fix this ?

Also there is currently a bug into the tcpdi library that remove html link (they are no more clicable). Don't know how to fix it, not so important but i think a better solution to include terms condition would be to provide a very large textarea html link so user can paste its terms in the zone as full text or html content.
So it will be ediable with the rich text editor and can be added into pdf without loosing links. This feature will probably replace the one you suggest as soon as i receive it....

BB2A-Anthony:
> I see.
However i think the setup page to upload the pdf term condition, should be on the menu 
Setup - pdf - other and not into menu - company. We must avoid to have the setup of 1 feature splitted into 2 different pages.
Can you fix this ?

> Also there is currently a bug into the tcpdi library that remove html link (they are no more clicable). Don't know how to fix it, not so important but i think a better solution to include terms condition would be to provide a very large textarea html link so user can paste its terms in the zone as full text or htm content.
So it will be ediable with the rich text editor and can be added into pdf without loosing links. This feature will probably replace the one you suggest as soon as i receive it.... 

I understand what you mean, I added it here in order to be able to use them in other places than in pdf files, as a link that can be sent or a link on the electronic signature page, etc.

where do you see the text box for the general conditions?

sonikf:
Hi @BB2A-Anthony 
Excellent contribution!
If you don't mind could you please rename termstosale and TERMSTOSALE to termsofsale and TERMSOFSALE

BB2A-Anthony:
> Hi @BB2A-Anthony Excellent contribution! If you don't mind could you please rename termstosale and TERMSTOSALE to termsofsale and TERMSOFSALE

@sonikf It's done

eldy:
> > I see.
> However i think the setup page to upload the pdf term condition, should be on the menu 
> Setup - pdf - other and not into menu - company. We must avoid to have the setup of 1 feature splitted into 2 different pages.
> Can you fix this ?
> 
> > Also there is currently a bug into the tcpdi library that remove html link (they are no more clicable). Don't know how to fix it, not so important but i think a better solution to include terms condition would be to provide a very large textarea html link so user can paste its terms in the zone as full text or htm content.
> So it will be ediable with the rich text editor and can be added into pdf without loosing links. This feature will probably replace the one you suggest as soon as i receive it.... 
> 
> I understand what you mean, I added it here in order to be able to use them in other places than in pdf files, as a link that can be sent or a link on the electronic signature page, etc.
> 
> where do you see the text box for the general conditions?

It should be Setup - pdf - other.
I understand your idea to prepare use for other purpose that pdf but for moment implemzntation is for pdf only, so in pdf setup is a good place. May be in a future the setup - pdf will be renamed into "setup - pdf and generated documents". Another idea is that the same setup can be accessible at different places so it is always at the place expected by user.
But we will see later if we need this...

BB2A-Anthony:
> > > I see.
> > > However i think the setup page to upload the pdf term condition, should be on the menu
> > > Setup - pdf - other and not into menu - company. We must avoid to have the setup of 1 feature splitted into 2 different pages.
> > > Can you fix this ?
> > 
> > 
> > > Also there is currently a bug into the tcpdi library that remove html link (they are no more clicable). Don't know how to fix it, not so important but i think a better solution to include terms condition would be to provide a very large textarea html link so user can paste its terms in the zone as full text or htm content.
> > > So it will be ediable with the rich text editor and can be added into pdf without loosing links. This feature will probably replace the one you suggest as soon as i receive it....
> > 
> > 
> > I understand what you mean, I added it here in order to be able to use them in other places than in pdf files, as a link that can be sent or a link on the electronic signature page, etc.
> > where do you see the text box for the general conditions?
> 
> It should be Setup - pdf - other. I understand your idea to prepare use for other purpose that pdf but for moment implemzntation is for pdf only, so in pdf setup is a good place. May be in a future the setup - pdf will be renamed into "setup - pdf and generated documents". Another idea is that the same setup can be accessible at different places sobit is always at the place zxpected by user. But we will see later if we need this...

@eldy ok i have changed it's done

BB2A-Anthony:
Up

BB2A-Anthony:
@eldy Ok for you?

eldy:
@BB2A-Anthony Can you help me to fix conflicts ?

BB2A-Anthony:
> @BB2A-Anthony Can you help me to fix conflicts ?

@eldy it's done

PR #24856, opened 2023-05-23
CLOSE #23331 Add support for parent projects

### LINKED ISSUE #23331: Nested projects

mc2contributor:
Provide support for sub-projects, as requested in issue https://github.com/Dolibarr/dolibarr/issues/23331, "Nested Projects".  This adds a parent project field to the project edit page, and shows parent and child projects on the project information page.


eldy:
Because the code is inside an explicit condition that is a hidden conf, ther me is no trouble to merge it. However, how do you plan to manage permission on project. When a user ha permission on p2 but not on p1 thatvis parent of p2 ?

edupulpillo:
The option is not present yet in v 19.0.0

Zimputer:
To enable this feature, set the following variable:
PROJECT_ENABLE_SUB_PROJECT = 1
This can be done under Home->Setup->Other Setup.

PR #24858, opened 2023-05-23
NEW: Add Encryption method and NoRSH configuration on email collector

noec764:
In the case of multiple emailcollectors being deployed at the same time, it may be useful to specify an encryption method and norsh parameter for each one.

Added the encryption method field, previously the encryption method was SSL by default and could only be changed with a global configuration. Also if  conf  `MAIN_IMAP_USE_PHPIMAP`  is ON, SSL was always used. 

Same for the noRSH configuration, a parameter existed but was never used.


htdocs/emailcollector/class/emailcollector.class.php
```diff
@@ -1191,7 +1192,7 @@ public function doCollectOneCollector($mode = 0)
 				$client = $cm->make([
 					'host'           => $this->host,
 					'port'           => $this->port,
-					'encryption'     => 'ssl',
+					'encryption'     => !empty($this->imap_encryption) ? $this->imap_encryption : false,
```
eldy, line 1195: Property $imap_encryption must also be declared into the emailcollector.class.php to avoid undefined property error.

FHenry:
@noec764 
Travis failed 
There was 1 failure:

1) EmailCollectorTest::testEmailCollectorCollect

Error in getConnectStringIMAP 

Failed asserting that two strings are equal.

--- Expected

+++ Actual

@@ @@

-'{localhost:993/service=imap/ssl/novalidate-cert}'

+'{localhost:993/service=imap/novalidate-cert}'

/home/travis/build/Dolibarr/dolibarr/test/phpunit/EmailCollectorTest.php:207

noec764:
@FHenry 
The default values defined in the $object->fields array were not taken into account when creating a specimen.

In my case, I set the value 'ssl' for imap_encryption but when the specimen was created, no value was assigned.

This is my fix: [PR 24867](https://github.com/Dolibarr/dolibarr/pull/24867)

PR #24862, opened 2023-05-23
new update for function generate_doc

lamrani002:
# NEW|New update for generate doc when delete object and rewrite properties


htdocs/core/lib/modulebuilder.lib.php
```diff
@@ -663,8 +663,7 @@ function writePropsInAsciiDoc($file, $objectname, $destfile)
 		$string = trim($string, "'");
 		$string = rtrim($string, ",");
 
-		$array = [];
-		eval("\$array = [$string];");
+		$array = eval("return [$string];");
```
eldy, line 666: Try to use dol_eval()

lamrani002: with do_eval it's not the result I want, but I've taken the necessary steps to work with eval by cleaning the string and checking if it's an array

eldy: Try 
dol_eval("return $string;", 1);
and better
dol_eval("$string");

lamrani002: do_eval not work for me so i found other way to do it

dev/initdemo/mysqldump_dolibarr_17.sql.zip
eldy, line 1: To remove

PR #24871, opened 2023-05-24
New msg attachment linked to event

MaximilienR-easya:
# New
Now the linked files inside ticket messages are linked to the event of the message and not to the ticket.
With this change, we can see the attached files to the message they are linked inside the message history

test card side

this PR is needed for another PR [*#24872*]


MaximilienR-easya:
@eldy

htdocs/core/lib/files.lib.php
```diff
@@ -378,18 +378,30 @@ function completeFileArrayWithDatabaseInfo(&$filearray, $relativedir)
...
-				$ecmfile->filename = $filename;
-				$ecmfile->label = md5_file(dol_osencode($filearray[$key]['fullname'])); // $destfile is a full path to file
-				$ecmfile->fullpath_orig = $filearray[$key]['fullname'];
-				$ecmfile->gen_or_uploaded = 'unknown';
-				$ecmfile->description = ''; // indexed content
-				$ecmfile->keywords = ''; // keyword content
-				$result = $ecmfile->create($user);
-				if ($result < 0) {
-					setEventMessages($ecmfile->error, $ecmfile->errors, 'warnings');
+				// verify if it's existing already
```
eldy, line 381: When we are in this part, it means "We found a file called not indexed into database."
So we can add it. Why doing a fetch ? We should already no if it exists or not with the the if (!$found) line 363

htdocs/core/class/html.formfile.class.php
```diff
@@ -1298,7 +1298,15 @@ public function list_of_documents($filearray, $object, $modulepart, $param = '',
 					if (array_key_exists('rowid', $filearray[$key]) && $filearray[$key]['rowid'] > 0) {
 						$lastrowid = $filearray[$key]['rowid'];
 					}
-					$filepath = $relativepath.$file['name'];
+					//$filepath = $file['level1name'].'/'.$file['name'];
+
+					$filepath = preg_replace('/^'.preg_quote(DOL_DATA_ROOT, '/').'/', '', $file['fullname']);
```
eldy, line 1303: Why using the fullname. We should use the short 'name' concatenated to a relativepath (relative path must be a variable set by the server and must not depends on a data coming from the end user as a parameter).
So we must keep use of $relativepath and never $file['fullname']

PR #24875, opened 2023-05-25
NEW : Add hook in user bank page

ATM-NicolasV:
# NEW|New [*Add hook in user bank page*]
[*Simple PR to add hook un user module on bank.php*]


.gitignore
```diff
@@ -4,6 +4,7 @@ htdocs/conf/conf.php.sav
 htdocs/conf/conf.php.back
 documents/
 custom2/
+htdocs/custom
```
eldy, line 7: We want to keep the custom dir with its readme as part of code to be sure it exists when we install with a git clone. So it must not be excluded by gitignore. However you can include your module dir into the gitignore into custom.

ATM-NicolasV: I fix it, thx for tour review

PR #24880, opened 2023-05-25
FIX PHP 8 Warning in subscription list

daraelmin:


htdocs/core/tpl/list_print_total.tpl.php
```diff
@@ -13,7 +13,7 @@
 	while ($i < $totalarray['nbfield']) {
 		$i++;
 		if (!empty($totalarray['pos'][$i])) {
-			switch ($totalarray['type'][$i]) {
+			switch ($totalarray['pos'][$i]) {
```
eldy, line 16: The different cases are different types. But you replace the variable of type with position ? Why ?

PR #24881, opened 2023-05-25
NEW ADD: custom compute for exports

jyhere:
NEW : add custom compute feature for exports

I needed a way to compute some values in my module export, like the remainToPay on invoice export... Unfortunately I realized  it was hardcoded it the export.class.php, with some other specific cases. So i used a similar way to import compute fields. Here's an example how to define a custom computed field in your export (i used the array_export_special prop) :

`		$this->export_special_array[$r] = [
			'none.encours' => [
				'rule' => 'compute',
				'classfile' => 'mymodule/lib/ExportLibrary.class.php',
				'class' => 'ExportLibrary',
				'method' => 'calculateMyCustomValue',
				'method_params' => [
					'rowid'
				]
			]
		];
`

### DIFF
--- a/htdocs/exports/class/export.class.php
+++ b/htdocs/exports/class/export.class.php
@@ -706,6 +706,44 @@ class Export
 									$remaintopay = $tmpobjforcomputecall->getRemainToPay();
 								}
 								$obj->$alias = $remaintopay;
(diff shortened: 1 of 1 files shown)


frederic34:
@eldy can you review ?

PR #24882, opened 2023-05-25
FIX error expedition qty 0

BenjaminFlr:
When `SHIPMENT_GETS_ALL_ORDER_PRODUCTS` is enabled, if we have a qty set to 0 (because already sent or unavailable in current stock), it generates an error on creating the shipment, as `GETPOST($qty, 'int')` doesn't get any value when the field is disabled, whereas GETPOSTINT generates an integer.

It's linked to the following error:
https://github.com/Dolibarr/dolibarr/blob/a47f59c5adbca3c8bd0a3266fd0dd3fc1cb8853f/htdocs/expedition/class/expedition.class.php#L2705

### DIFF
--- a/htdocs/expedition/card.php
+++ b/htdocs/expedition/card.php
@@ -357,7 +357,7 @@ if (empty($reshook)) {
 								$entrepot_id = 0;
 							}
 
-							$ret = $object->addline($entrepot_id, GETPOST($idl, 'int'), GETPOST($qty, 'int'), $array_options[$i]);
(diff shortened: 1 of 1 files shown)


atm-maxime:
@altairis-benjamin this needs an urgent FIX. You can't use `GETPOSTINT` to replace `GETPOST($qty, 'int')` or you loose any float value. For now users can't do a shipment with a decimal quantity. I am checking how to revert this and still fix your original issue

PR #24883, opened 2023-05-26
Add new substitution

SylvainLegrand:
Customer needs for email and SMS.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -8210,21 +8210,21 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 
 		if ($object->element =='facture') {
 			if (is_object($object)) {
-				$totalpaye			= $object->getSommePaiement();
-				$totalcreditnotes	= $object->getSumCreditNotesUsed();
-				$totaldeposits		= $object->getSumDepositsUsed();
-				$resteapayer		= price2num($object->total_ttc - $totalpaye - $totalcreditnotes - $totaldeposits, 'MT');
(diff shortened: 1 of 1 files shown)


kkhelifa-opendsi:
@eldy

htdocs/core/lib/functions.lib.php
```diff
@@ -8170,8 +8170,11 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 
 			if (is_object($object) && $object->element == 'action') {
 				$substitutionarray['__EVENT_LABEL__'] = $object->label;
+				$substitutionarray['__EVENT_TYPE__'] = $outputlangs->trans("Action".$object->type_code);
 				$substitutionarray['__EVENT_DATE__'] = dol_print_date($object->datep, '%A %d %b %Y');
 				$substitutionarray['__EVENT_TIME__'] = dol_print_date($object->datep, '%H:%M:%S');
+				$substitutionarray['__EVENT_DATE_SHORT__'] = dol_print_date($object->datep, '%d/%m/%y');
```
eldy, line 8176: The format of dol_print_date must be 'day' for a fotmated string in correct language.

SylvainLegrand: @eldy 
I do it.

Greetings,
Sylvain

eldy:
The substitution key for events has been added manually.
However, the addition for invoice has not be included because the substitution method is a low level method that can be called thousand of time in the same page. It must not include database access.

Instead a lazy mode must be implemented for the need of some substitution keys that need database accesses (see discussion into https://github.com/Dolibarr/dolibarr/pull/23820), so i close this PR.

PR #24886, opened 2023-05-26
fix: php warning on sabre lib

FHenry:
fix: php warning on sabre lib

What to do, this version of sabre is really old, we must in next Dolibarr version put the last ones 
https://github.com/sabre-io/vobject/releases/tag/4.5.3
is for php 7.4 minimum and dolibarr develop is php 7.0 minimum...

#24888

### DIFF
--- a/htdocs/includes/sabre/sabre/vobject/lib/Component/VCard.php
+++ b/htdocs/includes/sabre/sabre/vobject/lib/Component/VCard.php
@@ -506,7 +506,7 @@ class VCard extends VObject\Document {
                 switch ($property->name) {
 
                     case 'VERSION':
-                        continue;
+                        continue 2;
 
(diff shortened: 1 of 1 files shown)


lmag:
lastest version for php 7.0.0 : https://github.com/sabre-io/vobject/releases/tag/4.2.2

eldy:
We can push version [Release 4.5.1](https://github.com/sabre-io/vobject/releases/tag/4.5.1) that is ok for php 7.2 to 8.2 into Dolibarr or develop or 8.0 branch (qualified for php 7.1 to 8.2). Then setting a "php_min" to 7.2 into the desriptor of module Dav to explain module it does not work on php 7.1 to avoid error seems to be the larger common combination we can have between dolibarr and sabre.

PR #24889, opened 2023-05-26
FIX: Set update keys set in module descriptor by default in multiselectarray

noec764:
To prevent SQL import error if user don't fill the multiselectarray, I think set modules values by default is a better option.


eldy:
The update often needs several fields to filter on the unique key. If you add only 1 fields, result may be worse than no fields.
May be we must introduce a property to say what are the default value to use as preselected update keys...

PR #24892, opened 2023-05-26
FIX: String was compared to an integer

noec764:
Here $key is a string so $key - 1 throw a PHP error
```
 Uncaught TypeError: Unsupported operand types: string - int
```

### DIFF
--- a/htdocs/core/modules/import/import_csv.modules.php
+++ b/htdocs/core/modules/import/import_csv.modules.php
@@ -809,7 +809,7 @@ class ImportCsv extends ModeleImports
 										break;
 									}
 									$classinstance = new $class($this->db);
-									$res = call_user_func_array(array($classinstance, $method), array(&$arrayrecord, $listfields, ($key - 1)));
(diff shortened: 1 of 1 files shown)


htdocs/core/modules/import/import_csv.modules.php
```diff
@@ -809,7 +809,7 @@ public function import_insert($arrayrecord, $array_match_file_to_database, $obji
 										break;
 									}
 									$classinstance = new $class($this->db);
-									$res = call_user_func_array(array($classinstance, $method), array(&$arrayrecord, $listfields, ($key - 1)));
+									$res = call_user_func_array(array($classinstance, $method), array(&$arrayrecord, $listfields, $key));
```
eldy, line 812: This looks strange. Everywhere in the rest of the code of this file, we use $key -1.
Exemple: line 578, line 587, line 746, line 764, ...
so the key to get the value inside the &arrayrecord is ($key - 1).

noec764: @eldy I change as you said

PR #24895, opened 2023-05-26
NEW: Add triggers on import success

noec764:
Adding a trigger at the end of import process can be useful

### DIFF
--- a/htdocs/core/modules/import/import_csv.modules.php
+++ b/htdocs/core/modules/import/import_csv.modules.php
@@ -809,7 +809,7 @@ class ImportCsv extends ModeleImports
 										break;
 									}
 									$classinstance = new $class($this->db);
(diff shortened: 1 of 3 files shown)


hregis:
@noec764 @eldy triggers or hooks here ?

noec764:
@hregis If a custom process needs to be run after an import (e.g. linking one object to another after it has been created, as could be done with a standard creation trigger), there is currently no way to run it.

eldy:
trigger is reserved for CRUD BUSINESS event.
See https://wiki.dolibarr.org/index.php?title=Triggers
and https://wiki.dolibarr.org/index.php?title=Hooks_system
should use a hook here

noec764:
@eldy @hregis I've replaced my triggers with hooks in the import page, although I'm not sure why inserting items into the database isn't a CRUD event.

PR #24897, opened 2023-05-26
FIX: Error handling for computed values on import

noec764:
There is currently no way of catching errors for computed values if they occur in method


noec764:
@eldy Can you check this PR again ?

Issue #14522, opened 2020-08-21
Captcha Dolibarr


victorcardim-zz:
### Meio Ambiente

- Dolibarr 12.0.2
- Windows NT 10.0 compilação 19041 (Windows 10)
- Apache / 2.4.41 (Win64) PHP / 7.3.12
- 5.5-10.4.10-MariaDB 

Captcha is not differentiating upper and lower case characters. 

_This flaw was also identified by another user in the English Forum._
[https://www.dolibarr.org/forum/t/captcha-dolibarr/19436](url)

erikvanberkum:
I can confirm that on my installation:

- Dolibarr 12.0.2
- Ubuntu 18.04
- Apache
- mysql 

have the same issue

eldy:
The goal of a captcha is not like a password. A password must be case sensitive to increase the number of combination. There is not need to be case sensitive for a captchca.
Goal of a captcha is to force the system that submit the form to prove it is not a robot and that he was able to understand "humanly" the captcha. And proving that analysis of image was done by a human does not need to use case sensitive information when submitting the value, the value itself is enough (we don’t mind the number of possible combination for that, this is the goal of password). Think to the captcha of google that is the “Do you see a car on image ?”. We don’t mind if the answer if “Yes” or “yes”, all thing we need is to get yes or no.

victorcardim-zz:
Hello!

Thank you for your answer and clarification.

PR #24901, opened 2023-05-28
activate lowercase and uppercase in captcha

abdabd1988:
# Fix Error : Activate Upper Case of the CAPTCHA

### DIFF
--- a/htdocs/main.inc.php
+++ b/htdocs/main.inc.php
@@ -738,7 +738,7 @@ if (!defined('NOLOGIN')) {
 		// Verification security graphic code
 		if ($test && GETPOST("username", "alpha", 2) && !empty($conf->global->MAIN_SECURITY_ENABLECAPTCHA) && !isset($_SESSION['dol_bypass_antispam'])) {
 			$sessionkey = 'dol_antispam_value';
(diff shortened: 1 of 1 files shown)


priojk:
For a Captcha, as opposed to a password, there is no need to exactly match the case.

Please see also the discussion in [issue #14522 ](https://github.com/Dolibarr/dolibarr/issues/14522)

htdocs/main.inc.php
```diff
@@ -738,7 +738,7 @@ function analyseVarsForSqlAndScriptsInjection(&$var, $type)
 		// Verification security graphic code
 		if ($test && GETPOST("username", "alpha", 2) && !empty($conf->global->MAIN_SECURITY_ENABLECAPTCHA) && !isset($_SESSION['dol_bypass_antispam'])) {
 			$sessionkey = 'dol_antispam_value';
-			$ok = (array_key_exists($sessionkey, $_SESSION) === true && (strtolower($_SESSION[$sessionkey]) === strtolower(GETPOST('code', 'restricthtml'))));
+			$ok = (array_key_exists($sessionkey, $_SESSION) === true && $_SESSION[$sessionkey] === GETPOST('code', 'restricthtml'));
```
eldy, line 741: Goal of captchca is to add a layer test difficult tobread for non human, not to increase the entropy of possible combination (this is the goal of password). For such needs, beeing case sensitive is useless.

PR #24906, opened 2023-05-29
Fix undefined variable/property/index errors

mc2contributor:
Fix a number of undefined variable/property/index errors that appear in PHP 8.

### DIFF
--- a/htdocs/projet/admin/project.php
+++ b/htdocs/projet/admin/project.php
@@ -788,7 +788,7 @@ if (!$conf->use_javascript_ajax) {
 		'2'=>$langs->trans("Yes").' ('.$langs->trans("NumberOfKeyToSearch", 2).')',
 		'3'=>$langs->trans("Yes").' ('.$langs->trans("NumberOfKeyToSearch", 3).')',
 	);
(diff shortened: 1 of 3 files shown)


htdocs/projet/admin/project.php
```diff
@@ -788,7 +788,7 @@
 		'2'=>$langs->trans("Yes").' ('.$langs->trans("NumberOfKeyToSearch", 2).')',
 		'3'=>$langs->trans("Yes").' ('.$langs->trans("NumberOfKeyToSearch", 3).')',
 	);
-	print $form->selectarray("activate_PROJECT_USE_SEARCH_TO_SELECT", $arrval, $conf->global->PROJECT_USE_SEARCH_TO_SELECT);
+	print $form->selectarray("activate_PROJECT_USE_SEARCH_TO_SELECT", $arrval, $conf->global->PROJECT_USE_SEARCH_TO_SELECT ?? null);
```
eldy, line 791: For getting constant of setup, a better fix would be to use
getDolGlobalString('PROJECT_USE_SEARCH_TO_SELECT')

eldy:

PR #24907, opened 2023-05-29
Fix undefined and invalid access issues

mc2contributor:
Fix the following errors in htdocs/projet/element.php that show up in PHP 8:

line 1167: Undefined array key 1
line 1195: Attempt to read property "id" on null
line 1242: Attempt to read property "multidir_output" on null
line 1242: Trying to access array offset on value of type null
line 1242: Undefined property: Conf::$invoice_supplier
line 1242: Undefined property: Conf::$order_supplier
line 1242: Undefined property: Conf::$project_task
line 1492: Undefined variable $total_time


htdocs/projet/element.php
```diff
@@ -1164,7 +1164,7 @@
 			for ($i = 0; $i < $num; $i++) {
 				$tmp = explode('_', $elementarray[$i]);
 				$idofelement = $tmp[0];
-				$idofelementuser = $tmp[1];
+				$idofelementuser = $tmp[1] ?? null;
```
eldy, line 1167: If you include this code it means at next line, you may make a fetch with null parameters. If value is null, there is no reason to make a fetch
Can you describe a step to reproduce the warning ?

mc2contributor: `$elementarray` comes from `get_element_list()`, and is formatted according to line 865 of `htdocs/projet/class/project.class.php`, which makes the `'_'.$obj->fk_user` suffix optional:

```
    $elements[$i] = $obj->rowid.(empty($obj->fk_user) ? '' : '_'.$obj->fk_user);
```

We're working with invoices, orders, and tasks that are imported from another system by a script, and therefore don't have fk_user set.

PR #24909, opened 2023-05-29
Filter child projects out of project list.

mc2contributor:
With the addition of sub-projects, it is good to be able to shorten the list and see only the top-level projects.  This adds a checkbox to omit sub-projects from the project list.


eldy:
The feature of subproject will probably need a very long time to be stable due to a lot of side effects. For this reason any code related to subproject must be inside a condition if getDolGlobalInt('PROJECT_ENABLE_SUB_PROJECT')

PR #24912, opened 2023-05-30
NEW: agenda per user use quarter hour split instead of half hour split

FHenry:
feat: agenda per user use quarter hour split instead of half hour split


eldy:
Can you copy a screenshot of result when there is for example one event each quarter of hour ?

FHenry:
With
image


Without
image

FHenry:
@eldy I've implement this feature in custom page for my customer, so if this feature is not relevant in Dolibarr Core, I can close this PR. Just let me know.

PR #24931, opened 2023-06-01
NEW: add field 'ref' (of supplier product) for import supplier PO lines

kev-73:
# NEW
Currently when import supplier PO lines, the database field 'ref' (table "llx_commande_fournisseurdet") for supplier product reference is not available.
This short change allow to fill all fields of supplier PO line (displayed on supplier order web page).
image


htdocs/core/modules/modFournisseur.class.php
```diff
@@ -814,6 +814,7 @@ public function __construct($db)
 			'cd.fk_commande'    => 'PurchaseOrder*',
 			'cd.fk_parent_line' => 'ParentLine',
 			'cd.fk_product'     => 'IdProduct',
+			'cd.ref'     		=> 'Ref',
```
eldy, line 817: Can you use the label SupplierRef instead of Ref to be consistent in the field name with other screens...

kev-73: Yes, I didn't read correctly the file to use the good label...
By checking this, I saw that in the export part of file for the supplier order lines, it is the "RefSupplier" label which is used for the supplier product references instead of "SupplierRef". Would it be useful to modify this label as well ?

image

image

I updated my modification in this way.

PR #24934, opened 2023-06-01
FIX : Add/Update required modules for thirdparties (by categ) & partnership

atm-adrien:
## FIX : Add/Update required modules for thirdparties (by categ) & partnership

There were some errors the mailing pages because of some disabled modules.

### DIFF
--- a/htdocs/core/modules/mailings/thirdparties.modules.php
+++ b/htdocs/core/modules/mailings/thirdparties.modules.php
@@ -29,7 +29,7 @@ class mailing_thirdparties extends MailingTargets
 	public $desc = "Third parties (by categories)";
 	public $require_admin = 0;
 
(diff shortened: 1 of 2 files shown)


htdocs/core/modules/mailings/thirdparties.modules.php
```diff
@@ -29,7 +29,7 @@ class mailing_thirdparties extends MailingTargets
 	public $desc = "Third parties (by categories)";
 	public $require_admin = 0;
 
-	public $require_module = array("societe"); // This module allows to select by categories must be also enabled if category module is not activated
+	public $require_module = array("societe", "category"); // This module allows to select by categories must be also enabled if category module is not activated
```
eldy, line 32: When category module is not enabled, the selector must still be available. Only the filter on categories must be disabled.

atm-adrien: Indeed you're right, I corrected that

PR #24938, opened 2023-06-01
fix: display budget field on load project

FHenry:
On edit when only "Manage Task" is enable the budget is hide because "Task invoice" is not check

### DIFF
--- a/htdocs/projet/card.php
+++ b/htdocs/projet/card.php
@@ -771,7 +771,7 @@ if ($action == 'create' && $user->rights->projet->creer) {
 	}
 
 	// Budget
-	print '<tr class="classusetask classusebilltime"><td>'.$langs->trans("Budget").'</td>';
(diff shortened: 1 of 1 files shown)


htdocs/projet/card.php
```diff
@@ -1023,7 +1023,7 @@ function set_usage_task() {
 
 					function set_usage_bill_time() {
 						console.log("set_usage_bill_time");
-						if (jQuery("#usage_bill_time").prop("checked")) {
+						if (jQuery("#usage_bill_time").prop("checked") || jQuery("#usage_task").prop("checked") ) {
```
eldy, line 1026: Why adding the condition on usage task ?
To create bill, the project role we need is usage_bill_time.

FHenry: Because if you just have task usage checked, the budget field is hidden as it have classusebilltime class

PR #24943, opened 2023-06-02
NEW: Add missing code for THIRDPARTY_PROPAGATE_EXTRAFIELDS_TO_PROPAL on propal card

kkhelifa-opendsi:
Add missing code for THIRDPARTY_PROPAGATE_EXTRAFIELDS_TO_PROPAL on propal card

### DIFF
--- a/htdocs/comm/propal/card.php
+++ b/htdocs/comm/propal/card.php
@@ -1998,7 +1998,27 @@ if ($action == 'create') {
 	}
 
 	// Other attributes
-	include DOL_DOCUMENT_ROOT.'/core/tpl/extrafields_add.tpl.php';
+	$parameters = array();
+	if (!empty($origin) && !empty($originid) && is_object($objectsrc)) {
+		$parameters['objectsrc'] =  $objectsrc;
+	}
+	$parameters['socid'] = $socid;
+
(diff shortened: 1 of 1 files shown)


htdocs/comm/propal/card.php
```diff
@@ -1998,7 +1998,27 @@
 	}
 
 	// Other attributes
-	include DOL_DOCUMENT_ROOT.'/core/tpl/extrafields_add.tpl.php';
```
eldy, line 2001: You remove the geneic code with specific one ?
We try currently to go on the oposite direction...

eldy:
the following commit 137fdd28ac1a6cdabb60acdfea97a4b05acdd413 was done so you can use the generic code to achieve your need.
Just set the variable thirdpartytopropagateextrafieldsfrom if option is on before calling the extrafields_add.tpl.php.

eldy:
@kkhelifa-opendsi if you still need this change, can you implement it using the generic feature, it should be possible to get it without duplicating the common code now...

thomas-Ngr:
I implemented it the right way in following PR : https://github.com/Dolibarr/dolibarr/pull/36810

@kkhelifa-opendsi can you close this PR ?

PR #24965, opened 2023-06-06
NEW VAT rates support multientities

aspangaro:
Manage entity on VAT rate.
When we have two entities with different country, it's really needed

#19430


htdocs/install/mysql/migration/18.0.0-19.0.0.sql
```diff
@@ -33,7 +32,20 @@
...
+
+
+
+
+-- v19
+
+-- VAT multientity
+-- VMYSQL4.1 DROP INDEX uk_c_tva_id on llx_c_tva;
+-- VPGSQL8.2 DROP INDEX uk_c_tva_id;
+ALTER TABLE llx_c_tva DROP INDEX uk_c_tva_id;
```
eldy, line 45: Drop of index is already done by the 2 previous lines (mysql and postgresql need  2 different instruction as they are not compatible for drop). This one will fails on pgsql i think. You can remove it.

aspangaro: Thanks for the review - Fix done

PR #24966, opened 2023-06-06
Update list.php

Je2fb:
Modification of the SQL request to filter all orders which are not closed (delivered with invoice) and not canceled, this request will give a better view on orders

### DIFF
--- a/htdocs/commande/list.php
+++ b/htdocs/commande/list.php
@@ -900,7 +900,8 @@ if ($search_status <> '') {
 	if ($search_status == -3) {	// "validated + in process + delivered"
 		//$sql.= ' AND c.fk_statut in (1,2,3)';
 		//$sql.= ' AND c.facture = 0'; // invoice not created
-		$sql .= ' AND (c.fk_statut IN (1,2,3))'; // validated, in process or closed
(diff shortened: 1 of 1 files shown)


eldy:
Filter on order status should filter only on filter status. If a filter on invoice status is required or not, you should add a filter on invoice status that is a different field.
Mixing 2 fields cindition into same field create a confusion on the meaning of the filter.

Je2fb:
From the main page when you click on order status, the list need to keep your attention so it can be waiting for shipment, waiting for delivery or invoice to do. 

With the actual filter, you have a complete list of all order closed (deliver and invoiced). 

I understand it was not the best way, do you think we can add a new status? deliver?

eldy:
Can you paste a screenshot of the page where you click on order that reach to a non expected filter ?

Je2fb:
When you click order in the home page:
image

You will get all order already invoiced and deliver:
image

I think the main problem is the status closed, for my view the order is closed when the order is delivered and you did the invoice.

atm-maxime:
Hi ! I did some modifications in #18600, I think this ends up with the same solution as yours

eldy:
Closed. Should be fixed by #18600

PR #24968, opened 2023-06-06
add hook in constructor

frederic34:
by example, to be able to modify object->fields properties


frederic34:
@eldy still postponed ?

eldy:
Allowing a module to interact during the constructor seems a very dangerous geature that can lead to unreversible broken installation.
What can be the goal for an external module to modify the field array. Adding new fields ?

frederic34:
OK will find another way

PR #24981, opened 2023-06-07
PHP 8.2 compat

frederic34:


eldy (commented):
We should not need this with v18.
Why do you need it ?

frederic34:
not really needed, just for ask if it will help transition in php version

PR #24992, opened 2023-06-08
NEW: Hooks tab in debugbar

jyhere:
Add a hook tab in the debug bar looking like this : 

image


frederic34:
https://github.com/Dolibarr/dolibarr/pull/25028

htdocs/core/class/hookmanager.class.php
```diff
@@ -155,6 +160,26 @@ public function initHooks($arraycontext)
 	 */
 	public function executeHooks($method, $parameters = array(), &$object = '', &$action = '')
 	{
+        global $debugbar;
+		if (is_object($debugbar) && get_class($debugbar) === 'DolibarrDebugBar') {
```
eldy, line 164: Do you use this test ti detect "if debugbar module is enabled" ?
If yes can you try with just
If (isModEnabled('debugbar')) {

PR #24996, opened 2023-06-08
FIX #24756 Total sum on annual report bank

### LINKED ISSUE #24756: Report in Module Bank
### Bug

In version 17.0.1 (using php 8.0) the sums in monthly report in the module "bank" are incorrect. I found a problem in the file compta/bank/annuel.php. In line 199 and 200 the annual sums for "sorties" and "entrees" are set to zero every month in the loop over years. So as a final result I only see the sum of December.
Instead of 
```
for ($annee = $year_start; $annee <= $year_end; $annee++) {
$totsorties[$annee] = 0;
$totentrees[$annee] = 0;

```
it should be like this:
```
$totsorties[$annee] = 0;|
$totentrees[$annee] = 0;|
for ($annee = $year_start; $annee <= $year_end; $annee++) {
```

I'm not a developer, so maybe someone can fix this for future releases.

thx, Peter

### Environment Version

17.0.1

### Environment OS

synology disk station - linux

### Environment Web server

nginx

### Environment PHP

8.0

### Environment Database

MariaDB 10

### Environment URL(s)

_No response_

explained in Bug-Description.

In the menu Bank open the monthly report. The sum for the year is incorrect.

### Attached files

_No response_

aspangaro:
Before : 

image

After : 

image

### DIFF
--- a/htdocs/compta/bank/annuel.php
+++ b/htdocs/compta/bank/annuel.php
@@ -196,9 +196,10 @@ for ($mois = 1; $mois < 13; $mois++) {
 	print '<tr class="oddeven">';
 	print "<td>".dol_print_date(dol_mktime(1, 1, 1, $mois, 1, 2000), "%B")."</td>";
 
-	$totsorties[$annee] = 0;
-	$totentrees[$annee] = 0;
 	for ($annee = $year_start; $annee <= $year_end; $annee++) {
+		$totsorties[$annee] = isset($totsorties[$annee]) ? $totsorties[$annee] : 0;
+		$totentrees[$annee] = isset($totentrees[$annee]) ? $totentrees[$annee] : 0;
+
 		$case = sprintf("%04s-%02s", $annee, $mois);
 
 		print '<td class="right" width="10%">&nbsp;';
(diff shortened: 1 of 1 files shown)


htdocs/compta/bank/annuel.php
```diff
@@ -195,10 +195,10 @@
 for ($mois = 1; $mois < 13; $mois++) {
 	print '<tr class="oddeven">';
 	print "<td>".dol_print_date(dol_mktime(1, 1, 1, $mois, 1, 2000), "%B")."</td>";
-	for ($annee = $year_start; $annee <= $year_end; $annee++) {
-		$totsorties[$annee] = 0;
-		$totentrees[$annee] = 0;
 
+	$totsorties[$annee] = 0;
```
eldy, line 199: You can't do $totsorties[$annee] = 0 before the loop because $annee is not define before the loop.
May be you can move the init of $totsorties before the for $month by doing a 
        for ($annee = $year_start; $annee <= $year_end; $annee++) {
$totsorties[$annee] = 0

But before the loop on month...

PR #25012, opened 2023-06-08
fix SQL dump of 17

FHenry:
fix SQL dump of 17


yledoare:
ERROR 1136 (21S01) at line 1979: Column count doesn't match value count at row 1

for Dolibarr:develop

PR #25013, opened 2023-06-08
Add preference option to toggle support for sub-projects

mc2contributor:
# NEW Add preference option to toggle support for sub-projects
Several merge requests have been made to support sub-projects.  This adds a toggle switch to the settings for the Projects module.


eldy (commented):
Because this feature will probably need a long time to be stable (due to alot of side effect onba lot of screen or permission management), the option will remain as a hidden feature as long as implementation is not completed everywhere. Dolibarr policy is to make a feature visible only when it is fully 100 implemented.

PR #25028, opened 2023-06-09
add triggers collector for debugbar

frederic34:
image


htdocs/core/class/interfaces.class.php
```diff
@@ -68,6 +68,28 @@ public function __construct($db)
 	public function run_triggers($action, $object, $user, $langs, $conf)
 	{
 		// phpcs:enable
+		global $debugbar;
+
+		if (is_object($debugbar) && get_class($debugbar) === 'DolibarrDebugBar') {
+			if (!is_array($_SESSION['triggersHistory'])) {
```
eldy, line 74: Storing data that are debug and temporary to the access page is annoying.
Can't we store this instead into an already existing global var, like $conf->cache['triggerstack'] ?

frederic34: @eldy using a variable cause lost information due to header reload

PR #25031, opened 2023-06-09
NEW Default customer, category and product when enable TakePOS

andreubisquerra:
Easy setup for TakePOS. Create a customer when TakePOS is enabled. It checks if TakePOS has already been configured to avoid recreating the customer if it has already been created.


htdocs/core/modules/modTakePos.class.php
```diff
@@ -266,9 +267,45 @@ public function __construct($db)
...
-		global $conf, $db;
+		global $conf, $db, $langs, $user;
+		$langs->load("cashdesk");
 
 		dolibarr_set_const($db, "TAKEPOS_PRINT_METHOD", "browser", 'chaine', 0, '', $conf->entity);
+		
+		//Default customer for Point of sale
+		if (empty(getDolGlobalInt('CASHDESK_ID_THIRDPARTY1'))) {
+			$societe = new Societe($db);
+			$societe->nom = $langs->trans("DefaultPOSThirdLabel");
```
eldy, line 278: What if you disable and reenable the module. You must check the objectvdoes not existvyet by a fetch to avoid to try to create it a second time

htdocs/core/modules/modTakePos.class.php
```diff
@@ -266,9 +267,45 @@ public function __construct($db)
...
+		//Create category if not exists
+		$categories = new Categorie($this->db);
+		$cate_arbo = $categories->get_full_arbo('product', 0, 1);
+		if (is_array($cate_arbo)) {
+			if (!count($cate_arbo)) {
+				$category = new Categorie($db);
+				$category->label = $langs->trans("DefaultPOSCatLabel");
+				$category->type = Categorie::TYPE_PRODUCT;
+				$result = $category->create($user);
+				if ($result > 0) {
```
eldy, line 297: I understand addition of a Default Category and Thirdparty, but what is the goald of this product ?

htdocs/langs/en_US/cashdesk.lang
```diff
@@ -145,3 +145,6 @@ ShowProductReference=Show reference of products
 UsePriceHT=Use price excl. taxes and not price incl. taxes
 TerminalName=Terminal %s
 TerminalNameDesc=Terminal name
+DefaultPOSThirdLabel=TakePOS sales
```
eldy, line 148: I suggest to name it "Default customer for POS"

htdocs/core/modules/modTakePos.class.php
```diff
@@ -266,9 +267,45 @@ public function __construct($db)
...
+		global $conf, $db, $langs, $user;
+		$langs->load("cashdesk");
 
 		dolibarr_set_const($db, "TAKEPOS_PRINT_METHOD", "browser", 'chaine', 0, '', $conf->entity);
+		
+		//Default customer for Point of sale
+		if (empty(getDolGlobalInt('CASHDESK_ID_THIRDPARTY1'))) {
+			$societe = new Societe($db);
+			$societe->nom = $langs->trans("DefaultPOSThirdLabel");
+			$societe->client = 1;
```
eldy, line 279: You should also set property note_private = 'Default customer automaticaly created by Point Of Sale module activation. Can be used as the default customer in the Point Of Sale setup. Can also be edited or removed.'

htdocs/langs/en_US/cashdesk.lang
```diff
@@ -145,3 +145,6 @@ ShowProductReference=Show reference of products
...
+DefaultPOSCatLabel=Point of sale
```
eldy, line 149: I suggest to name this category like this
"Category for all products in Point Of Sale"

htdocs/core/modules/modTakePos.class.php
andreubisquerra: When reactivated the second time, as the variable CASHDESK_ID_THIRDPARTY1 is already assigned, it will not be created again. It is only created the first time when the CASHDESK_ID_THIRDPARTY1 variable has never been initialized.

htdocs/core/modules/modTakePos.class.php
andreubisquerra: For TakePOS to work you must have created at least one product. If there are no products created, TakePOS says it needs at least one product to work. A generic product may be interesting to be able to use TakePOS without configuration. Then if the user wants to change the name or create more products, they are welcome.

htdocs/langs/en_US/cashdesk.lang
andreubisquerra: Approved for me

htdocs/core/modules/modTakePos.class.php
andreubisquerra: Approved for me

htdocs/langs/en_US/cashdesk.lang
andreubisquerra: Approved for me

PR #25037, opened 2023-06-09
NEW Ask to create cash account if there is no cash account in TakePOS

andreubisquerra:


htdocs/takepos/index.php
```diff
@@ -938,6 +956,10 @@ function WeighingScale(){
 	if ($_SESSION["takeposterminal"] == "") {
 		print "ModalBox('ModalTerminal');";
 	}
+	
+	if (empty(getDolGlobalInt('CASHDESK_ID_BANKACCOUNT_CASH'.$_SESSION["takeposterminal"]))) {
```
eldy, line 960: What if we do not accept cash ? 
Can't we add the warning on the pay php page to show a warning picto and make the payment mode greyed if payment mode bank account not defined ?

eldy: There is 2 point:
1)
We can ask to create automatically a cash account, like we can create automatically the category for products in POS.
But all automatic creation must not be done during the usage of the POS (the user using the POS may not have the permission and we should not bypass permission, alors this is related to setup). So this must be done at the first initial activation of the module using a generic label (there is even no need to ask, we can just create default record - category-bank account- with default label if there does not exists). And setup can be automatically set to this values if not yet defined to something.
This must be introduced into the init method of the TakePOS module.
2) Then we must deal when setup was modified and is not complete, with un understandable message. This can be done into the takepos/index.php but we should not have setup change nor creation done into this page as this page use used with low permission (by takepos simple user and action that must be under control of admin must not be done by a user with low privileges). So we should only provide advices on setup but automatics setup/creation of object must be done into case 1 only (so module Activation).

PR #25043, opened 2023-06-10
NEW Change default background style of company logo

Kazimir42:
Hi, I think it should be better :

Before :
image

After :
image

### DIFF
--- a/htdocs/theme/eldy/global.inc.php
+++ b/htdocs/theme/eldy/global.inc.php
@@ -2072,10 +2072,6 @@ div.blockvmenulogo
 	max-width: 100px;
 	vertical-align: middle;
 }
-.backgroundforcompanylogo {
-	background-color: rgba(255,255,255,0.2);
-	border-radius: 4px;
-}
 .menulogocontainer img.mycompany {
 	object-fit: contain;
 	width: inherit;
--- a/htdocs/theme/md/style.css.php
+++ b/htdocs/theme/md/style.css.php
@@ -3409,10 +3409,6 @@ div.blockvmenulogo
 	max-width: 100px;
 	vertical-align: middle;
 }
-.backgroundforcompanylogo {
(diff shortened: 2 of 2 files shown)


eldy:
Let's explain why we got a small background style different.
When logo is jpg with no transparence, this does not change nothing, we don't see the background.
But when the company upload a logo png with a transparency. When the logo has a color near the background color of the top menu, we can't see it any more
This happen frequently because, it is very common that the admin customize the color of the top banner with the color of the company. And logo is also the color of the company. So the logo is no more visible.
For this reason an opacity has been added on the background of logo so whatever is the situation, the logo is always visible clearly.
A better solution would be to detect if we need to add this opacity or not, but adding it, all the time, was a quick solution that work in most cases (even if we can do nicer when logo does not share colors with the top menu background, like in your exemple).

John-H-Smith:
As an admin, my expectations on the behavior of the logo would be as described in the PR.
I would expect that the logo is added with a fully transparent background. This is how most of the open souce softwares are handling it - so I am expecting it here as well. The admin user should be aware that the logo isn't visible if it has the same color as the menu bar.
You could add a setting to to add such a opacity to catch the case as described above.

eldy:
Ok to comment the background.color.
But can we keep the border radius ?

John-H-Smith:
Why would you? 
Some companies might have a logo without a padding and hard corners. 
This could make the logo be different and then it wouldn't match the corporate's designs. 
But change my mind - what benefit do you see on adding a border radius?

Kazimir42:
I agree, I think it's better not to change the design of the company's logo.

eldy:
In v20 and + the logo is not visible by default (we need more space for menu bar and logo output often generates a bad visual effect dur to color combination). So this point is now less important because by default, there is no problem.
So i validate the PR.

Issue #22744, opened 2022-11-03
Bank Journal - Auxiliary account - "NotDefined" is returned instead of a clear value


thibdrev:
### Bug

In the Auxiliary Account column of the Bank Journal, `NotDefined` is returned instead of a default value:
![image](https://user-images.githubusercontent.com/7697736/199775335-16f20e2f-cc0c-4d92-9ff6-5303e9d273fd.png)

### Environment Version

16.0.1

### Environment OS

Linux builder.o2switch.net 4.18.0-348.12.2.lve.el7h.x86_64

### Environment Web server

Apache

### Environment PHP

7.4.30

### Environment Database

MySQL or MariaDB 5.5.5-10.5.17-MariaDB

### Environment URL(s)

/htdocs/accountancy/journal/bankjournal.php?id_journal=3

### Expected and actual behavior

* Expected behavior: 
    * A clear message such as: "Compte auxiliaire non défini ou tiers ou utilisateur inconnu. Nous utiliserons **421USER**"
* Actual behavior:
    * An unclear message: "Compte auxiliaire non défini ou tiers ou utilisateur inconnu. Nous utiliserons NotDefined"

### Steps to reproduce the behavior

At first I thought this was linked to the user accountancy account not being defined before I made the payment.
But I checked the following on a test instance: create a user and fill in its accountancy account > create an expense report, validate it and pay it. And I still get this behaviour.

The translated message is located here on Transifex:
https://www.transifex.com/dolibarr-association/dolibarr/translate/#fr_FR/accountancy/165994858
`Compte auxiliaire non défini ou tiers ou utilisateur inconnu. Nous utiliserons %s`

It is linked to the translation key `ThirdpartyAccountNotDefinedOrThirdPartyUnknown` which is used here:
https://github.com/Dolibarr/dolibarr/blob/57feb24f4acd82c25fe28c8706d902335d70220f/htdocs/accountancy/journal/bankjournal.php#L1287

So I'm guessing the `$tabcompany[$key]['code_compta']` is the `NotDefined` value.

ouitec:
Hello,

Exactly in the same situation. We've upgraded to 16.0.3 and we still have the issue, when do you think you can fix this issue ?

We also have this issue on the modules "Paiements/Salaires" :

[...]

PR #25060, opened 2023-06-10
Update bankjournal.php

EchoLoGeek:
#22744 @aspangaro ?


aspangaro:
Thanks @EchoLoGeek but I don't think this fix works. The problem seems to go deeper than that. I'll check and get back to you.

PR #25074, opened 2023-06-11
FIX shipment gets all order products

BenjaminFlr:
Regarding the constant `SHIPMENT_GETS_ALL_ORDER_PRODUCTS`, if we manually set the quantity to 0 during the shipment creation from an order, the line isn't added to the expedition, whereas the ones without stock are.


BenjaminFlr:
I missed the existing fix. Closing the PR.

PR #25090, opened 2023-06-13
Fix : deprecated warnings on various pages

Hystepik:
fix some warnings on deprecated use of some functions on some pages


htdocs/core/modules/project/mod_project_simple.php
```diff
@@ -150,7 +150,8 @@ public function getNextValue($objsoc, $project)
 		$date = (empty($project->date_c) ? dol_now() : $project->date_c);
 
 		//$yymm = strftime("%y%m",time());
-		$yymm = strftime("%y%m", $date);
+		//$yymm = strftime("%y%m", $date);
+		$yymm = date("%y%m", $date);
```
eldy, line 154: Can you use dol_print_date instead of date ? (date is not clear about timezone behaviour)

PR #25097, opened 2023-06-14
FIX Update fk_user_assign in DB during new ticket creation [23838]

FredBl79:
# FIX|Fix #[23838]
New ticket creation form
execute assignuser when a new ticket is created in order to update the row in the db with fk_user_assign value. This is missing today when you assign an user directly during ticket form creation. Works already well when ticket is already existing.

### DIFF
--- a/htdocs/ticket/class/ticket.class.php
+++ b/htdocs/ticket/class/ticket.class.php
@@ -500,8 +500,13 @@ class Ticket extends CommonObject
 				if ($this->add_contact($this->fk_user_assign, 'SUPPORTTEC', 'internal') < 0) {
 					$error++;
 				}
-			}
+				$ret = $this->assignUser($user, $this->fk_user_assign);
+				if ($ret < 0) {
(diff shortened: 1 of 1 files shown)


htdocs/ticket/class/ticket.class.php
```diff
@@ -500,8 +500,13 @@ public function create($user, $notrigger = 0)
 				if ($this->add_contact($this->fk_user_assign, 'SUPPORTTEC', 'internal') < 0) {
 					$error++;
 				}
-			}
+				$ret = $this->assignUser($user, $this->fk_user_assign);
```
eldy, line 503: What happen if fk_user_assign is not set or empty (case by default) ?

FredBl79: Just go deeper, I can simplify the code, the issue is that trigger TICKET_ASSIGNED is not checked in function create. Will change and come back to you

PR #25113, opened 2023-06-15
FIX : Link in knowledge records

atm-adrien:
## FIX : Link in knowledge records

The field "answer" was displayed by the dolEditor object.
But this must be used only for edit and create view otherwise the links can not work.

Before : 
image

After : 
image


eldy:
Because the solution of a km article is the main article. It was designed on purpose to have the message on all the width of the page. Thus allow the show the article that can be a full web page content.

atm-adrien:
Ok, but, with this solution, no link entered in the article will be functional.
Do you have a solution that I could implement to solve this problem?

atm-maxime:
Maybe just display the answer on the full width of the screen not only half left.
Another inconvenient of the ckeditor is that the height is limited and displays a scrollbar

atm-adrien:
I created a new PR that refers to this one: https://github.com/Dolibarr/dolibarr/pull/25173

atm-adrien:
I created a new PR that refers to this one: https://github.com/Dolibarr/dolibarr/pull/25173

atm-maxime:
@eldy what is the point of having this textearea displayed in readonly with ckeditor ?

Issue #22467, opened 2022-10-01
UNIQUE index for _extrafields table


fgarciapatino:
### Bug

facturedet extrafield table allows multuple fk_object records

### Environment Version

14.0.2

### Environment OS

LAMP

### Environment Web server

Apache/2.4.6 (CentOS) OpenSSL/1.0.2k-fips PHP/7.4.21

### Environment PHP

7.4.21

### Environment Database

5.5.5-10.2.39-MariaDB

### Expected and actual behavior

Expected: only one record for foreign key
Actual: multiple records for the same fk 

### Steps to reproduce the behavior

insert into llx_facturedet_extrafields (fk_object, XXX) values ([Valid FK], [Valid value]);
-- And again
insert into llx_facturedet_extrafields (fk_object, XXX) values ([Valid FK], [Valid value]);

vmaury:
I confirm !

mc2contributor:
This is a problem for all `_extrafields` tables, except for `ged_files` and `societe`.  The concept of extra fields, as I understand it, is to add fields to the base table without modifying the base table, so there should be a one-to-one correlation.

PR #25116, opened 2023-06-15
Use foreign keys to link extrafields tables, fixes #22117 and #22467

### LINKED ISSUE #22117: Indexes on extrafields should be UNIQUE
### Bug

The database model seems to be not clear on extrafields tables. Here is what i have on a dolibarr 14 : idx_propal_extrafields is an index but not unique.

image

The result is you CAN (sql accept) have more than one row with different extrafields for a propal :-( and dolibarr database could become dirty. 

The relation between main table and extrafield table is 1.1 isn't it ?

### Environment Version

14

If idx_propal_extrafields was unique we could use UPSERT requests to Insert Or Updade massive data with raw SQL requests like that

`
INSERT INTO llx_propal_extrafields(fk_object,informations,representant) VALUES (2019 ,'','CR') ON DUPLICATE KEY UPDATE fk_object=2019, informations = 'xxxxxxxxx', representant = 'xxxxxxxx'
`

mc2contributor:
# CLOSE #22117 and #22467
`extrafields` tables should have a one-to-one correlation with the tables they add fields to.  This requires _at least_ the use of unique indexes, but using foreign keys with `ON DELETE CASCADE` is better, ensuring that tables stay in sync.

Use of unique indexes prevents the need for additional code to check for existing records.  Now I can simply do:

```
$extras['fk_object'] = $rowid;
$fields = implode(',',array_keys($extras));                                                                    
$values = implode(',',array_map(function($s) use($db) {return "'".$db->escape($s)."'";}, $extras));            
$updates = 'values('.implode('),values(', array_keys($extras)).')';                                            
$q = "INSERT INTO d_projet_task_extrafields ($fields) VALUES ($values) ON DUPLICATE KEY UPDATE $updates";
```

whereas before making this change, the following was required:

```
$q = "SELECT 1 FROM d_projet_task_extrafields WHERE fk_object=$rowid LIMIT 1";
$res = $db->query($q);
if (!$res) $exists = 0;
else {
    $row = $db->fetch_row($res);
    $exists = $row && intval($row[0]);
}
if ($exists) {
    $set = [];
    foreach ($extras as $k=>$v) {
        $set = "$k='".$db->escape($v)."'";
    }
    $setSql = implode(',',$set);
    $q = "UPDATE d_projet_task_extrafields SET $setSql WHERE fk_object=$rowid ORDER BY rowid LIMIT 1";
}
else {
    $extras['fk_object'] = $rowid;
    $fields = implode(',',array_keys($extras));
    $values = implode(',',array_map(function($s) use($db) {return "'".$db->escape($s)."'";}, $extras));
    $q = "INSERT INTO d_projet_task_extrafields ($fields) VALUES ($values)";
}
```


eldy:
Such changes generates a lot of troubles. It breaks the sql rules defined on this page :
https://wiki.dolibarr.org/index.php?title=Language_and_development_rules#SQL_Coding_rules

mc2contributor:
Revised to use `UNIQUE INDEX` without foreign keys.

mc2contributor:
I just realized I could have kept the foreign keys, as long as I don't use 'ON CASCADE DELETE', is that right?  I'll change this back to using foreign keys.

htdocs/install/mysql/migration/17.0.0-18.0.0.sql
```diff
@@ -476,3 +476,67 @@ ALTER TABLE llx_actioncomm ADD INDEX idx_actioncomm_ref (ref, entity);
 -- Bump llx_reception.ref_supplier to allow up to 255 characters to match llx_commande_fournisseur.ref_supplier.
 -- See: https://github.com/Dolibarr/dolibarr/pull/25034
 ALTER TABLE llx_reception MODIFY COLUMN ref_supplier varchar(255);
+
+-- Use foreign key contraints for extrafields
+ALTER TABLE llx_actioncomm_extrafields DROP INDEX idx_actioncomm_extrafields; ALTER TABLE llx_actioncomm_extrafields ADD FOREIGN KEY idx_actioncomm_extrafields (fk_object) REFERENCES llx_actioncomm (id);
```
eldy, line 481: Index and foreign keys are 2 differents thinks. Index is adding data managed for performance (idx_...)
Forrign keys is a rule/constraint to check (called fk_...)
Note tha adding a foreign keys needs an index key on the parent field

mc2contributor: Thanks for pointing this out.  Will the following be good, or should I avoid the foreign key?

```
+ALTER TABLE llx_actioncomm_extrafields DROP INDEX idx_actioncomm_extrafields; ALTER TABLE llx_actioncomm_extrafields ADD UNIQUE INDEX uk_actioncomm_extrafields (fk_object); ALTER TABLE llx_actioncomm_extrafields ADD FOREIGN KEY fk_actioncomm_extrafields (fk_object) REFERENCES llx_actioncomm (id);
```

mc2contributor: There are two parts to this merge request:

1) Use of UNIQUE INDEX to ensure that there is no more than one extrafields row for every master table row.  Without this it is more work to prevent duplicate extrafields rows, where the authoritative value for a field could be uncertain.  This is important.  (First of two commits in this MR)

2) Use of FOREIGN KEY to ensure that every extrafields row corresponds to a master table row.  This is not critical and could be omitted from this merge request.  (Second of two commits in this MR)

mc2contributor:
There are two parts to this merge request:

1) Use of UNIQUE INDEX to ensure that there is no more than one extrafields row for every master table row.  Without this it is more work to prevent duplicate extrafields rows, where the authoritative value for a field could be uncertain.  This is important.  (First of two commits in this MR)

2) Use of FOREIGN KEY to ensure that every extrafields row corresponds to a master table row.  This is not critical and could be omitted from this merge request.  (Second of two commits in this MR)

eldy:
Moving idx_ key into unique key uk_ is possible but PR can't be qualified due to "ADD CONSTRAINT" addition.

PR #25129, opened 2023-06-16
Convenience methods for database access

mc2contributor:
# NEW Easy database query methods
Two methods I've found incredibly useful in other projects: `getVar()` to retrieve single values from the database, and `getCol()` to retrieve a single column.  Simplifies code and saves typing.

I was happy to find that `getRow()` and `getRows()` were already there, but disappointed that `getRows()` is marked as deprecated.  I realize the potential for memory issues with `getRows()`, but I think it's too useful for small requests to remove it entirely.  I'll be using all four of these in our system.

### DIFF
--- a/htdocs/core/db/DoliDB.class.php
+++ b/htdocs/core/db/DoliDB.class.php
@@ -363,9 +363,34 @@ abstract class DoliDB implements Database
 	}
 
 	/**
+	 * Return single field from query.
+	 * Only to be used with SELECT queries.
+	 * Don't add LIMIT to your query, it will be added by this method
+	 *
+	 * @param 	string 				$sql 	The sql query string
+	 * @return 	mixed    			False on failure, null on empty, field value on success (could be null)
+	 */
+	public function getVar($sql)
+	{
+		$sql .= ' LIMIT 1';
+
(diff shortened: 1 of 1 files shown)


htdocs/core/db/DoliDB.class.php
```diff
@@ -411,4 +436,27 @@ public function getRows($sql)
...
+	 * @param 	string 		$sql 		The sql query string
+	 * @return 	bool|array				Result
+	 */
+	public function getCol($sql)
+	{
+		$res = $this->query($sql);
+		if ($res) {
+			$results = array();
+			if ($this->num_rows($res) > 0) {
+				while ($arr = $this->fetch_row($res)) {
```
eldy, line 453: Using of fetch_row make unnamed access to data.
Column access must use a named access to avoid bug or bad data use when field orders differs from one database to another one.

mc2contributor: The idea here is to use with queries like:

```
SELECT nom FROM d_projet;
```

to get a list of names as an array, where only one column is being selected, so that field order is not applicable.  The same applies to both `getVar()` and `getCol()`.

PR #25133, opened 2023-06-18
FIX | Update fourn/facture/list.php

cmfpmatik:
# FIX|Fix #0000
Can apply filter when user config it on "/admin/defaultvalues.php?mode=filters"


htdocs/fourn/facture/list.php
```diff
@@ -113,6 +112,10 @@
 $search_categ_sup = trim(GETPOST("search_categ_sup", 'int'));
 $search_product_category = GETPOST('search_product_category', 'int');
 
+$search_status = GETPOST('search_status', 'alpha');
+$optioncss = GETPOST('optioncss', 'alpha');
+$object_statut = GETPOST('search_statut', 'alpha');
```
eldy, line 117: Why adding search_statut ?  Seems a duplicate of search_status.
Filter on search_status seems already working.

cmfpmatik: pl

eldy:

PR #25134, opened 2023-06-18
FIX | Update compta/facture/list.php

cmfpmatik:
FIX|
Can apply filter when user config it on "/admin/defaultvalues.php?mode=filters"


htdocs/compta/facture/list.php
```diff
@@ -145,6 +145,10 @@
 $search_btn = GETPOST('button_search', 'alpha');
 $search_remove_btn = GETPOST('button_removefilter', 'alpha');
 
+$search_status = GETPOST('search_status', 'alpha');
+$optioncss = GETPOST('optioncss', 'alpha');
+$object_statut = GETPOST('search_statut', 'alpha');
```
eldy, line 150: Why adding search_statut ? Seems a duplicate of search_status.
Filter on search_status seems already working.

eldy:

PR #25135, opened 2023-06-19
Fix:warehouses table name is entrepot which does not hold an fk_soc field, so _checkAccessToResource returns always false and must be disabled before fix.

bafbes:
Fix:warehouses table name is entrepot which does not hold an fk_soc field, so _checkAccessToResource returns always false and must be disabled before fix.

### DIFF
--- a/htdocs/product/stock/class/api_warehouses.class.php
+++ b/htdocs/product/stock/class/api_warehouses.class.php
@@ -71,7 +71,7 @@ class Warehouses extends DolibarrApi
 			throw new RestException(404, 'warehouse not found');
 		}
 
-		if (!DolibarrApi::_checkAccessToResource('stock', $this->warehouse->id,'entrepot')) {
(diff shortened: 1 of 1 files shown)


htdocs/product/stock/class/api_warehouses.class.php
```diff
@@ -71,10 +71,10 @@ public function get($id)
 			throw new RestException(404, 'warehouse not found');
 		}
 
-		if (!DolibarrApi::_checkAccessToResource('warehouse', $this->warehouse->id)) {
+/*		if (!DolibarrApi::_checkAccessToResource('warehouse', $this->warehouse->id)) {
```
eldy, line 74: This disable the security on warehouse objects.

bafbes: I think it is better to disable momentarily this check as it is bugged in respect of wirehouses. There remains a first security check to check the user read right on warehouses.

PR #25161, opened 2023-06-20
Edit the status prospect

KiVoilaJe:
Edit the status prospect icone when we change the status prospect
[Capture vidéo du 21-06-2023 01:47:17.webm](https://github.com/Dolibarr/dolibarr/assets/36703787/678544d1-7531-43bf-8a7f-51cc9d9fc130)


eldy:
Thanks to change the image to use a value return by img_action from the ajax component.
Add also the error case

PR #25163, opened 2023-06-21
Fix : image conversion to webp for websites

Hystepik:
Make image conversion to webp of website module overwrite if image already exist in webp format


eldy:
It doesn't seems to fix the case:
When you click on "Generate wepb" on a given png file, when the webp file already exists, the webp file is not regenerated.

Hystepik:
https://github.com/Dolibarr/dolibarr/assets/45882981/7d9eef11-259b-4b74-9a81-653545081694

PR #25167, opened 2023-06-21
NEW: Sort batch by dates when create a shipping (independently of the warehouse)

kkhelifa-opendsi:
Sort batch by dates when create a shipping (independently of the warehouse)

Before:
image

After:
image


eldy:
I don't see a sort order on date on the screenshot. I see date 22 then 11 then 14 for product tube.

eldy:

PR #25173, opened 2023-06-22
FIX : Link in knowledge records 2 

atm-adrien:
## FIX : Link in knowledge records

This PR refers to this one: https://github.com/Dolibarr/dolibarr/pull/25113

image


eldy:
Because ckeditor is reponsible of edit, it is also this component that must be responsible of output.
So i tried a fix by adding a ckeditor change into option for click on readonly view.

atm-adrien:
@eldy what is the point of having this textearea displayed in readonly with ckeditor ?

PR #25181, opened 2023-06-23
NEW #25171 : add a function in Societe class to retrieve contacts of …

### LINKED ISSUE #25171: no function to retrieve contacts of a societe
### Bug

I mark this as a bug even it's not a real one, even it seems to be

I wanted in one of my custom modules to retrieve all contacts of a societe
I naturaly tried to use
`$societe->list_contact(-1, "external', ...);`
but even if the societe do have some, no one has been retrieved.
The reason : for external contacts, `list_contact` is only looking in `llx_socpeople` table, but, only for societe contacts, these are stored in `llx_societe_contacts` table
`llx_societe_contacts` has very a similary structure as `llx_element_contacts`, except for the unique key 'element_id, fk_c_type_contact, fk_socpeople' which is absent of llx_societe_contacts (a same societe can have several contacts of same types).
And I looked every where in the code, there is no function to retrieve all contacts (by code) of a societe.
I suggest 2 ways to correct this :

- delete the unique key in llx_element_contacts, and store societe contacts in this table. If so, no need to modify generic list_contact method
- modify generic list_contact method, so that when searching for external societe contacts, it select in llx_societe_contacts table.

I'll happy to hear some comments about this !

### Environment Version

all

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

vmaury:
NEW #25171 : no function to retrieve contacts of a societe

Add method list_contact_societe in Societe class

### DIFF
--- a/htdocs/societe/class/societe.class.php
+++ b/htdocs/societe/class/societe.class.php
@@ -5292,7 +5292,7 @@ class Societe extends CommonObject
 	function liste_contact_societe($list = 0, $code = '', $element = '')
 	{
 		// phpcs:enable
-		global $langs, $db;
+		global $langs;
 
 		$tab = array();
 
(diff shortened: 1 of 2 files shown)


htdocs/societe/class/societe.class.php
```diff
@@ -5276,4 +5276,98 @@ public function getKanbanView($option = '', $arraydata = null)
...
+	 *    @param	int			socid		id of societe
+	 *    @param	int         $list       0:Return array contains all properties, 1:Return array contains just id
+	 *    @param    string      $code       Filter on this code of contact type ('SHIPPING', 'BILLING', ...)
+	 *	  @param    string      $element       Filter on this element of default contact type ('facture', 'propal', 'commande' ...)
+	 *    @return	array|int		        Array of contacts, -1 if error
+	 * 
+	 *
+	 * 
+	 */
+	function liste_contact_societe($list = 0, $code = '', $element = '')
```
eldy, line 5292: When we introduce a new method, we should use camel names.
I suggest:
getContacts(...)

vmaury: Ok venerated Yoda !
I change method's name, buy don't have right to change label ...

PR #25189, opened 2023-06-24
NEW  #25188 :  Add a status shipment in process

### LINKED ISSUE #25188: Jam in shipment and orders statuses workflow
### Bug

I activated workflow module, and the 2 options
WORKFLOW_ORDER_CLASSIFY_SHIPPED_SHIPPING and WORKFLOW_ORDER_CLASSIFY_SHIPPED_SHIPPING_CLOSED

But when I validate a shipment (of the total order products), the order's becomes "received" (=STATUS_CLOSED) instead of STATUS_SHIPMENTONPROCESS.
I found the bug (PR to come in few minutes) in /htdocs/core/triggers/interface_20_modWorkflow_WorkflowManager.class.php.

BUT

Even if you desactive WORKFLOW_ORDER_CLASSIFY_SHIPPED_SHIPPING and WORKFLOW_ORDER_CLASSIFY_SHIPPED_SHIPPING_CLOSED , the order status is (good) refreshed when shipment status is changed (may be not exactly as the same maner than with Workflow but more logicaly)

### Environment Version

all

### Environment OS

all

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

vmaury:
Actually, when an expedition is Validated, the linked order is closed (marked as delivered) instead of being marked as SHIPMENTONPROCESS

### DIFF
--- a/htdocs/core/triggers/interface_20_modWorkflow_WorkflowManager.class.php
+++ b/htdocs/core/triggers/interface_20_modWorkflow_WorkflowManager.class.php
@@ -370,7 +370,11 @@ class InterfaceWorkflowManager extends DolibarrTriggers
 				$diff_array = array_diff_assoc($qtyordred, $qtyshipped);
 				if (count($diff_array) == 0) {
 					//No diff => mean everythings is shipped
-					$ret = $order->setStatut(Commande::STATUS_CLOSED, $object->origin_id, $object->origin, 'ORDER_CLOSE');
+					if ($action == 'SHIPPING_VALIDATE') {
+						$ret = $order->setStatut(Commande::STATUS_SHIPMENTONPROCESS, $object->origin_id, $object->origin, 'ORDER_MODIFY');
+					} else { // $action == 'SHIPPING_CLOSED'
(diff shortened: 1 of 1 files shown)


atm-maxime:
I think this is the wanted behavior. When you validate a shipment and all the products from the order have been shipped, the order is closed and this means there's nothing to ship anymore.

eldy:
> I think this is the wanted behavior. When you validate a shipment and all the products from the order have been shipped, the order is closed and this means there's nothing to ship anymore.

I think the goal of pr is to change this behaviour to set status of order in "shipment in progress" (so order is still open but this is not a problem, if we try another shipment only diff missing are set by default, so nothing if all was already in first order).
And when the shipment is closed, status of order is set to closed.

There is however an old option WORKFLOW_BILL_ON_SHIPMENT, when this option is on, validating shipment means shipment processed, so we should close completely order also in this case (pr to enhance with this)

eldy:
The need seems already implemented (or i missed something ?), so i close the ticket.

PR #25192, opened 2023-06-25
Update Dutch lang.

Tim-VL:
Update writing for some words and lines.


priojk:
Hi and thanks for improving the Dutch translations. Unfortunately, all translations are managed via the tool/portal transifex.com and not directly in the sources. Here they would get lost with the next synch of transifex translations.

It's easy to set up a transifex account and if you don't mind, it would be great, if you could make your improvements there.

Some more information around Transifex can be found in the Dolibarr Wiki: [https://wiki.dolibarr.org/index.php?title=Translator_documentation](https://wiki.dolibarr.org/index.php?title=Translator_documentation)

Tim-VL:
Thanks, missed that part!

I will go ahead with Transifex.

fappels:
Transifex team join request accepted.
Welcome Tim

PR #25198, opened 2023-06-26
FIX : Make batch "datec" not visible and not settable

atm-adrien:
## FIX : Make batch "datec" not visible and not settable

Currently, when creating a batch/serial number, it is necessary to enter the creation date.
It's not done for any other module, so I don't understand why it's done here? Is it useful?

Moreover, even if I enter a creation date, it is the current date that is taken into account in the recording.


eldy:
Creation date for a lot is more important than for other fields. Because it says when a lot was initialized the first time and this is required in a lot of tracking activity. This is the reason why this is a field that can be entered by the user.
For other object, this is not required and date of creation is just for history log usage.
May be it could be defined to current date by default to save input time ?

atm-maxime:
I guess the batch object should have a datec (date of record) field working just like all other object, used for log purpose. If a batch needs an important information such as date creation (date of initialization) it should be in a dedicated field (date of production?)

eldy:
> I guess the batch object should have a datec (date of record) field working just like all other object, used for log purpose. If a batch needs an important information such as date creation (date of initialization) it should be in a dedicated field (date of production?)

Yes having a date_init is clearly the best choice. Until we have it i prefer keeping datec available in user input mode as it is used for both purpose due to lack of this date_init field.

atm-adrien:
I noticed that a "manufacturing_date" field already exists.
This field is subject to a hidden conf: PRODUCT_ENABLE_TRACEABILITY

Couldn't this field replace the creation date and become a standard field? It seems that this answers your traceability problem

eldy:
> manufacturing_date

good catch.  Yes it seems this field is the date_init we miss. You can make this field visible as standard and hide the creation date (Even if i don't like the name, date_for_first_use would be better). Sometime we plan to assign lot number several days before using them. But well, we can just put a comment into sql file to say "manufacturing_date" is the "date for first use of the lot".
In migration update this field manufacturing_date with date_creation when field manufacturing_date is empty.

eldy:
Replaced by #25381

PR #25203, opened 2023-06-26
NEW do not add default value in list

frederic34:
if `MAIN_DEFAULT_PAYMENT_TERM_ID` or `MAIN_DEFAULT_PAYMENT_TYPE_ID` is set, this add a default value in the select box in list


eldy:
Option MAIN_DEFAULT_PAYMENT_TERM_ID is an old option when default value feature was not yet available.
The doc says "Constant MAIN_DEFAULT_PAYMENT_TERM_ID can be used to set default value but scope is all application, probably not what you want."
and invite to use the default value feature. I think we should not make change on this option since it is deprecated now that we have a generic default feature option.

PR #25204, opened 2023-06-26
qodana fix 2 missing returns

oussamanhairech:
qodana fix 2 missing returns


oussamanhairech:
[qodana fix 2: missing returns]

htdocs/core/class/CSMSFile.class.php
```diff
@@ -82,7 +82,7 @@ public function __construct($to, $from, $msg, $deliveryreceipt = 0, $deferred =
 		// If ending method not defined
 		if (empty($conf->global->MAIN_SMS_SENDMODE)) {
 			$this->error = 'No SMS Engine defined';
-			return -1;
+			throw new Exception('No SMS Engine defined');
```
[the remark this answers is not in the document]
oussamanhairech: Hi, I added a try catch [9a31b20](https://github.com/Dolibarr/dolibarr/pull/25204/commits/9a31b207e9a023341541bc7a76e6933ae18c5214).

htdocs/core/class/html.formcontract.class.php
```diff
@@ -187,8 +187,9 @@ public function formSelectContract($page, $socid = -1, $selected = '', $htmlname
 		print '<form method="post" action="'.$page.'">';
 		print '<input type="hidden" name="action" value="setcontract">';
 		print '<input type="hidden" name="token" value="'.newToken().'">';
-		$this->select_contract($socid, $selected, $htmlname, $maxlength, $showempty, $showRef);
 		print '<input type="submit" class="button smallpaddingimp valignmiddle" value="'.$langs->trans("Modify").'">';
 		print '</form>';
+		return $this->select_contract($socid, $selected, $htmlname, $maxlength, $showempty, $showRef);
```
eldy, line 192: The method formSelectContract make print. So the goal is to output HTML code.
The select_contract() also output HTML code but if you move it inside the return, it means the code will be output out of the `<form> </form>`
Call of select_contract must be restored as its original place and result saved into a $result variable to be output into the return.

htdocs/core/class/html.formfile.class.php
```diff
@@ -1958,7 +1958,10 @@ public function list_of_autoecmfiles($upload_dir, $filearray, $modulepart, $para
 		if (!empty($addfilterfields)) {
 			print '</form>';
 		}
+		return count($filearray);
 		// Fin de zone
+
```
eldy, line 1963: Do not use empty space lines before a }.
Take a look at this doc, and enable the git hook like explained into documentation so each time you make such errors, it will be fixed automatically during the commit:  https://wiki.dolibarr.org/index.php?title=Outils_de_d%C3%A9veloppement#Installer_le_hook_git_Pre-commit

htdocs/core/class/translate.class.php
```diff
@@ -170,9 +170,17 @@ public function getDefaultLang($mode = 0)
 	 */
 	public function loadLangs($domains)
 	{
+		$loaded = 0;
 		foreach ($domains as $domain) {
-			$this->load($domain);
+			$result = $this->load($domain);
+			if( $result>0 ){
```
eldy, line 176: No space before the )
Need a space between the if and (
If you add the git hook as explained here https://wiki.dolibarr.org/index.php?title=Outils_de_d%C3%A9veloppement#Installer_le_hook_git_Pre-commit
such error will be fixed automatically

htdocs/core/lib/functionsnumtoword.lib.php
```diff
@@ -247,6 +247,7 @@ function dolNumberToWord($numero, $langs, $numorcurrency = 'number')
 		}
 		return $entexto;
 	}
+	return -1;
```
eldy, line 250: If you return -1 here (like it is done for some previous error check), you must also modify the comment doc of the method by changing 
`* @return string Text of the number`
into
`* @return int|string Text of the number`

htdocs/hrm/class/skill.class.php
```diff
@@ -417,7 +417,7 @@ public function fetchLines()
 			$this->error = $skilldet->error;
 			return $this->lines;
 		}
-		return -1;
+		return [];
```
eldy, line 420: In Dolibarr, we have a convention. On error, we return -1 so we can make a difference between, the result was an empty array (nothing found) and there was a technical error.
So keep here the return -1 and instead change the doc comment from
`* @return array <0 if KO, array of skill level found`
into
`* @return array|int    <0 if KO, array of skill level found`

htdocs/core/class/html.formcontract.class.php
```diff
@@ -182,13 +182,14 @@ public function select_contract($socid = -1, $selected = '', $htmlname = 'contra
 	public function formSelectContract($page, $socid = -1, $selected = '', $htmlname = 'contrattid', $maxlength = 16, $showempty = 1, $showRef = 0)
 	{
 		global $langs;
-
+		$result = $this->select_contract($socid, $selected, $htmlname, $maxlength, $showempty, $showRef);
```
eldy, line 185: To move at the good place

PR #25206, opened 2023-06-26
FIX : Can update extrafield of expedition det with lot

Kazimir42:
# FIX : Can update extrafield of expedition det with lot

Hi, if I want to update an extra field of expedition det with lot isn't possible, i have to change one other value who isn't an extra field cuz of the check on the if()

### DIFF
--- a/htdocs/expedition/card.php
+++ b/htdocs/expedition/card.php
@@ -661,7 +661,7 @@ if (empty($reshook)) {
 								}
 
 								$line->detail_batch->fk_origin_stock = $batch_id;
-								$line->detail_batch->batch = $lotStock->batch;a
+								$line->detail_batch->batch = $lotStock->batch;
 								$line->detail_batch->id = $detail_batch->id;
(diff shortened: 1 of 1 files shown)


htdocs/expedition/card.php
```diff
@@ -648,7 +648,7 @@
 						$qty = "qtyl".$detail_batch->fk_expeditiondet.'_'.$detail_batch->id;
 						$batch_id = GETPOST($batch, 'int');
 						$batch_qty = GETPOST($qty, 'int');
-						if (!empty($batch_id) && ($batch_id != $detail_batch->fk_origin_stock || $batch_qty != $detail_batch->qty)) {
+						if (!empty($batch_id) && ($batch_id != $detail_batch->fk_origin_stock || $batch_qty > 0)) {
```
eldy, line 651: What if you want to change qty in line of shipment detail from 5 to 0 ?

Kazimir42: You right, maybe should remove checks ? It's what i have done now

PR #25220, opened 2023-06-28
FIX check tva_tx before comparing price_min_ttc

altairis-noe:
# FIX check tva_tx before comparing price_min_ttc
When updating or adding a line in a propal, min_price is checked with both excluding and including VAT values. When we sell internationnaly, VAT value may not be default VAT set in product, so the comparison will fail. This checks if the line VAT is the same as the product VAT before checking price including VAT value.


htdocs/comm/propal/card.php
```diff
@@ -1267,7 +1267,7 @@
 					$mesg = $langs->trans("CantBeLessThanMinPrice", price(price2num($price_min, 'MU'), 0, $langs, 0, 0, -1, $conf->currency));
 					setEventMessages($mesg, null, 'errors');
 					$error++;
-				} elseif ($pu_equivalent_ttc && $price_min_ttc && ((price2num($pu_equivalent_ttc) * (1 - $remise_percent / 100)) < price2num($price_min_ttc))) {
+				} elseif ($tva_tx == $prod->tva_tx && $pu_equivalent_ttc && $price_min_ttc && ((price2num($pu_equivalent_ttc) * (1 - $remise_percent / 100)) < price2num($price_min_ttc))) {
```
eldy, line 1270: This means that if product that has a default vat of 20, when sold to a foreign country (where vat is 0), the test on minimum prices will be completely disabled ?

altairis-noe: The test on minimal price *including vat* is averted, but the test on the price *excluding vat* remains untouched

Whatever is decided here, it might be a good idea to also do it on commande and facture cards.

altairis-noe: Hey there @eldy, what do you think we should do here. I see several options:

- we can reapply the vat rate specific to this proposal to check the minimum price. Which would mean:
  - comparing `$tva_tx` to `$prod->tva_tx`
  - if `$tva_tx` is inferior, make `$pu_equivalent_ttc` the vat included price recalculated as if it had the `$prod->tva_tx` rate for VAT
- ignore `$pu_equivalent_ttc` altogether (I must admit that I fail to see why we compare both vat excluded and included price)
- use the solution in this PR
- the answer D

Thanks for your input here

eldy: If the price was set in HT, it means the reference price is the price without tax, son we should check the minimum sell price without tax only.
If the price was set including tax, then we should check the price with tax.

altairis-noe:
I think this is it.

In fact, the way it was : it had little to no chance to ever check the tax included price !

altairis-noe:
How do you see it, @eldy ? Does this work for you ?

altairis-noe:
Some new input here ?

altairis-noe:
Hi there, @eldy, where are we on this one?

eldy:
> Hi there, @eldy, where are we on this one?

Looks better.
Just one thing to fix
... price2num(GETPOST('price_ht') * (1 + $vat_rate), '', 2)...
Must be replaced with 
price2num(GETPOST('price_ht')) * ( 1 + $vat_rate)

Getpost return a non formated string with local separator. No operation is possible on it. You must first convert it into a real numeric with price2num before doing the *

altairis-noe:
I guess this is it!

altairis-noe:
@eldy are we good, here ?

PR #25225, opened 2023-06-29
float casting empty value

atm-greg:
# FIX price2num bad return
Fix case when price2num return an empty string => php8.2 through a fatal error when adding a string to an int

### DIFF
--- a/htdocs/core/triggers/interface_50_modBlockedlog_ActionsBlockedLog.class.php
+++ b/htdocs/core/triggers/interface_50_modBlockedlog_ActionsBlockedLog.class.php
@@ -119,7 +119,7 @@ class InterfaceActionsBlockedLog extends DolibarrTriggers
 			$amounts = 0;
 			if (!empty($object->amounts)) {
 				foreach ($object->amounts as $amount) {
-					$amounts += floatval(price2num($amount));
+					$amounts += (double) $amount;
 				}
 			} elseif (!empty($object->amount)) {
 				$amounts = $object->amount;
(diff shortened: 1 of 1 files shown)


htdocs/core/triggers/interface_50_modBlockedlog_ActionsBlockedLog.class.php
```diff
@@ -119,7 +119,7 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
 			$amounts = 0;
 			if (!empty($object->amounts)) {
 				foreach ($object->amounts as $amount) {
```
eldy, line 121: $amount here is a variable in memory. So its content should already be a valid amount.
So I think we should not have price2num() at all here.
We should just have 
$amounts += (double) amount;
like we have into other trigger codes of the same trigger.
What did you get into the $object->amount array ?

htdocs/core/triggers/interface_50_modBlockedlog_ActionsBlockedLog.class.php
```diff
@@ -119,7 +119,7 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
...
-					$amounts += price2num($amount);
+					$amounts += floatval(price2num($amount));
```
atm-greg, line 122:
```suggestion
					$amounts += (double) $amount;
```

atm-greg:
@eldy Change done !

PR #25226, opened 2023-06-29
# FIX|Fix dispatch correct status by ignoring lines with specific product type 9 

gdsoftdev:
Ignoring lines with specific product type 9 to calculate differences, remaining quantities and dispatch the correct status for orders


eldy:
Can you give more information. How can we reproduce the bug currently ?

gdsoftdev:
Make a supplier order with lines of text from the subtotal module. Create a receipt with all quantities received, the order changes to 'partially received' status instead of 'fully received'.
The quantity of 50 of the text line is taken into account to calculate if there are still products to be received

atm-maxime:
Hi @gdsoftdev ! Thanks for this contribution, can you please use different branches for each contribution?

gdsoftdev:
@atm-maxime. You're right, I forgot that I had this PR when I merged some new commits. 
So I close this PR and open a new one with a new branch

atm-maxime:
😉

PR #25228, opened 2023-06-29
add call_trigger variable

atm-greg:
# FIX missing $call_trigger variable

### DIFF
--- a/htdocs/contact/class/contact.class.php
+++ b/htdocs/contact/class/contact.class.php
@@ -433,10 +433,10 @@ class Contact extends CommonObject
 	 *  Add a contact into database
 	 *
 	 *  @param      User	$user           Object user that create
-     *  @param      int     $call_trigger   0=no, 1=yes
+     *  @param      int     $notrigger	    1=Does not execute triggers, 0= execute triggers
 	 *  @return     int      			    <0 if KO, >0 if OK
 	 */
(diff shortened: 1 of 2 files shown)


atm-maxime:
Have a look in the modulebuilder myobject.class.php, you'll see the standard variable used is "no trigger" set to false by default. 😉

atm-greg:
@atm-maxime I have done this like that because i found this variable on the update method of societe.class.php.

Then i realized that $call_trigger = 1 was more understandable than !$notrigger without changing the basic behaviour.
And the parameter wasn't present before.

htdocs/contact/class/contact.class.php
```diff
@@ -432,10 +432,11 @@ public function load_state_board()
...
 	 *  Add a contact into database
 	 *
-	 *  @param      User	$user       Object user that create
-	 *  @return     int      			<0 if KO, >0 if OK
+	 *  @param      User	$user           Object user that create
+     *  @param      int     $call_trigger   0=no, 1=yes
+	 *  @return     int      			    <0 if KO, >0 if OK
 	 */
-	public function create($user)
+	public function create($user, $call_trigger = 1)
```
eldy, line 439: To follow standard, using $notrigger = 0 would be a better choice.

atm-greg:
@eldy fixed

PR #25229, opened 2023-06-29
fix sql with left join added by hook and no rights to see all thirdparties

frederic34:
using printFieldListFrom to add a left join, if user hasnt right to see all thirdparties, the request fails


frederic34:
no feedback, closing

PR #25233, opened 2023-06-29
FIX #25143 Using default_values for extrafields by replacing GETPOSTISSET by GETPOST

### LINKED ISSUE #25143: Default values for extrafields does not work on lists
### Bug

I'm trying to use the default search filters on one of my list, and it seems to not be working for the extrafields of the object, mainly because of this line `$search_array_options = $extrafields->getOptionalsFromPost($object->table_element, '', 'search_')`

where the `getOptionalsFromPost()` function is using the `GETPOSTISSET()` function which doesn't check for values inside user->default_values unlike the `GETPOST()` function

Shall I do a PR updating the `getOptionalsFromPost` function with GETPOST instead of GETPOSTISSET (seems to work at least for the `type boolean` just tried in local) or is there a reason we're using the GETPOSTISSET function inside of it ? 

Thanks!

### Environment Version

_No response_

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

V-Py:
Replacing `GETPOSTISSET()` by `GETPOST()` in function `getOptionalsFromPost()` to use default_values

**Allows the use of extrafields for filters/sorting** in admin/defaultvalues.php by using GETPOST() instead of GETPOSTISSET(), because GETPOST() also check the `$user->default_values` array which contains the values for **filters** and **sorting** by default on the different lists.


ksar-ksar:
#25143

eldy:
@V-Py i think pr is ok. However can you help me fix the conflict ?

PR #25234, opened 2023-06-29
FIX: VTA rounding calc option must be the same on update line as on create line

FHenry:
On addline method the ronding method is set to auto, so on updateline it must be the same.

I have a case where
(PU HT = 24 €) X (Qty = 3) = 72 € - (Discount 16.66% = 11.9952 €) = 60,0048

VAT 5.5 of 60,0048 € = 3,300264 € => TOTAL VTA : rounding to 2 decimal 3.31 €
But
TOTAL HT = 60.00 €  => VTA 5.5 of 60.00 € = TOTAL VTA 3.30 €

OK, MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND is here for this case

it work nice on addline 
but If I click on update line to change the description and Save => the rule of MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND is not apply any more 
=> my VAT is back to 3.31 €

### DIFF
--- a/htdocs/comm/propal/class/propal.class.php
+++ b/htdocs/comm/propal/class/propal.class.php
@@ -955,7 +955,7 @@ class Propal extends CommonObject
 					$this->line_order(true, 'DESC');
 				}
 
-				$this->update_price(1);
+				$this->update_price(1, 'auto');
 
 				$this->fk_propal = $this->id;
 				$this->rowid = $rowid;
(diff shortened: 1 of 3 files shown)


Daviid-P:
What about `facture.class.php`

`update_percent`
`delete_line`

and other places in the class where `update_price` is used?

And `fournisseur.facture.class.php`?

FHenry:
@Daviid-P @eldy 
Good point ! 
So why not change default param on method call to auto ?
https://github.com/Dolibarr/dolibarr/blob/cf54e7879dfc3462d6a64e565d4a7e809c2d7144/htdocs/core/class/commonobject.class.php#L3513

PR #25245, opened 2023-06-30
FIX|Fix dispatch correct status by ignoring lines with specific product type 9 

gdsoftdev:
Ignoring lines with specific product type 9 to calculate differences, remaining quantities and dispatch the correct status for orders
To reproduce :
Make a supplier order with lines of text from the subtotal module. Create a receipt with all quantities received, the order changes to 'partially received' status instead of 'fully received'.
The quantity of 50 of the text line is taken into account to calculate if there are still products to be received


eldy:
Does this means that you have a quantity of 50 as qty into lines of text added by the subtotal module ?

gdsoftdev:
it depends on the type of line (title, subtotal, free text) but the
quantity can go from 1 to 99 for lines from the subtotal module.

eldy:
So you are probably using an external module that do a bad job. 
The field total_... is dedicated for accountancy (for aggregation on reports, on export, ...). A module should not insert a value into it if this value is not dedicated to be sum with other lines into accountancy reports. The module should :
- calculate dynamically the value if he can
- or store its data in its own table/field or at least into an extrafield, but must not use a column used for calculation and aggregation from different axis by different tools or pages.

You can use the module subtotal by atm that has not this bug, if you absolutely need a subtotal module.

Issue #25242, opened 2023-06-30
BOM - Calculate quantity on line of services with service data to keep the consistency of the data


lmag:
### Feature Request

Add field quantity calculed
Qté (unit) * Durée (Unit) = Quantity calculed
![image](https://github.com/Dolibarr/dolibarr/assets/3288514/d48b94a0-6791-4116-bf7b-6dbf063927f0)

### Use case

We find the same operation in the invoicing of time in the tasks
The operation is identical for the time calculated from the time of the tasks via a service
![image](https://github.com/Dolibarr/dolibarr/assets/3288514/101660ba-62ce-43d8-939f-fdb7b2e33799)
In invoice
![image](https://github.com/Dolibarr/dolibarr/assets/3288514/ca54b4d1-33f3-4c2b-b68c-ccd2f3e2e670)

lmag:
Sample of the new function
![image](https://github.com/Dolibarr/dolibarr/assets/3288514/f312f615-80cc-4456-a80c-40202f4986c2)

lmag:
@FHenry

PR #25248, opened 2023-06-30
feat : Real time calculation base on service in BOM

FHenry:
#25242 25242


htdocs/bom/tpl/objectline_view.tpl.php
```diff
@@ -149,19 +149,28 @@
 	echo $line->efficiency;
 	print '</td>';
 } else {
-	// Unit
-	print '<td class="linecolunit nowrap right">';
+	// Duration
+	print '<td class="linecolqty nowrap right">';
```
eldy, line 153: linecolduration instead of linecoqty ?

htdocs/bom/tpl/objectline_view.tpl.php
```diff
@@ -149,19 +149,28 @@
...
+			setEventMessages(null, $tmpproduct->errors, 'errors');
+		}
+		$prodDurationSecond = $prodDurationHours*3600;
+		print convertSecondToTime($line->qty*$prodDurationSecond, 'allhourmin');
 	}
 
 	print '</td>';
 
+	// Qty
+	print '<td class="linecolduration nowrap right">';
```
eldy, line 169: linecolqty instead of linecolduration ?

eldy:
Something is hurting me here.
When we define a BOM, for example, duration of a service for the BOM is 2 days., it must remains 2 days as long as we do not change this. Even if the unit of service is changed from 1 min into 1 hour, the BOM must keep its reference of 2 days. Here you remove the unit of BOM line, this means that if we change a product unit reference, it have effect on all BOM that use this product and qty means nothing anymore.
The unit should remain on BOM line (because it is a duration entered by the end user, so a use rinput), and it is when we convert the BOM into the MO that we can convert it into duration in seconds (because it is a calculated duration). We can use for this the unit of BOM + Unit of the service).

eldy:
So can you paste the screenshot of a BOM card after your change ?

FHenry:
@eldy 
BOM
image

OF
image

PR #25249, opened 2023-06-30
qodana fix : code smell, missing returns.

oussamanhairech:
qodana fix : code smell, missing returns.


oussamanhairech:
qodana fix : code smell, missing returns.

htdocs/comm/action/class/cactioncomm.class.php
```diff
@@ -341,5 +341,6 @@ public function getNomUrl($withpicto = 0)
 		if ($transcode != "Action".$this->code) {
 			return $transcode;
 		}
+		return -1;
```
eldy, line 344: So comment doc say "@return string Label of action type"
So you must choose between making a return '';
or a return -1 and you modify the doc to say  "* @return int|string Label of action type"

htdocs/conf/conf.phy~
```diff
@@ -0,0 +1,49 @@
+<?php
```
eldy, line 1: You must remove this file from your commit

htdocs/conf/conf.phz~
```diff
@@ -0,0 +1,49 @@
+<?php
```
eldy, line 1: You must remove this file from the PR

htdocs/core/class/fiscalyear.class.php
```diff
@@ -446,6 +446,8 @@ public function LibStatut($status, $mode = 0)
 				return $langs->trans($this->statuts_short[$status]).' '.img_picto($langs->trans($this->statuts_short[$status]), 'statut6');
 			}
 		}
+		return "";
+
```
eldy, line 450: No spaces before a }

htdocs/core/class/google.class.php
```diff
@@ -86,5 +86,7 @@ public function getGeoCoordinatesOfAddress($address)
 		//echo "Longitude: $longitude ',' Latitude: $latitude";
 
 		$i++;
+		return "Longitude: $longitude, Latitude: $latitude";
+
```
eldy, line 90: No spaces before a }

htdocs/comm/mailing/class/advtargetemailing.class.php
```diff
@@ -516,6 +516,7 @@ public function savequery($user, $arrayquery)
 				$this->create($user);
 			}
 		}
+		return -1;
```
eldy, line 519: When we reach this point, it seems this is the "standard" path. So we should not return error but ok, so
return 1

eldy: @oussamanhairech ping

PR #25263, opened 2023-07-04
Fix: translation missing for key 'Sens'

priojk:


priojk:
Mixup with other PR. Will reissue.

PR #25274, opened 2023-07-05
Develop4

oussamanhairech:
qodana 4 : code smells, missing returns


eldy:
Same than develop3 PR

PR #25287, opened 2023-07-06
Get infos from thirdparty

Quentin-Seekness:
Fix a bug when creating an order from scratch. Once the thirparty has been selected payment mode, terms of payment and bank account should appear


eldy:
I can't reproduce the trouble.
when i create a draft order from scratch, after selecting a thirdparty, the payment mode, term of payment and bank are already set with the value of the selected thirdparty.

PR #25292, opened 2023-07-06
FIX Supplier orders : delete dispatched lines on supplier order deletion

ibuiv:
# FIX|Fix #[Supplier orders : delete dispatched lines on supplier order deletion]
If stock increase is set up to dispatch to warehouse manually, the order deletion may cause stock issues. 
You can valid an order, dispatch some lines (increasing your stock) and delete the order. 
The lines are dispatched, the stock moves are written and you cannot access to the dispatch lines because the order has been removed from the database.
A warning has been added on deletion modal if there are existing dispatched lines.
If confirmed, for each dispatched line, we apply the same method used by clicking on the trash icon on a dispatch line. 
The label for the stock move generated will be trans('SupplierOrderDeletion') By default : Supplier order %s Deletion


eldy:
When product were shipped, and order is canceled after, the product are still shipped.
For this reason, an order deletion should not affect the stock. A deletion is an action to "clean database", not a business event. For a business event, the action "cancel" is better. There is a lot of use cas also we need to delete and without reverting stock change already done.
So adding a warning that there is some stock movement already recorded, during order deletion, is a very good thing. But the message must also say:
1) solution 1: explain that if a stock movement must be reverted, this must be done manually because not done by default
2) Solution 2: explain that if a stock movement were already recorded, and ask the user if this must be reverted or not (end user must decide).

ibuiv:
@eldy 
Ok solution 2 was applied 
image

Text is added to en_US trans only

htdocs/fourn/commande/card.php
```diff
@@ -1100,12 +1106,62 @@
...
-			exit;
-		} else {
-			setEventMessages($object->error, $object->errors, 'errors');
+		// Delete existing dispatched lines
+		$errOnDelete = 0;
+		if($stockDelete){
+			$dispatchedLines = $object->getDispachedLines();
+			if (!empty($dispatchedLines)) {
+				foreach ($dispatchedLines as $dispatchedLine) {
+					$db->begin();
```
eldy, line 1115: The begin end transaction must not be inside the loop. We must have 1 transaction for all the event so result in database will be everything done or nothing.

ibuiv: I modified code to locate correctly begin, rollback and commit instructions for db. 
I noticed that $commande->delete also have a begin, rollback and commit logic on $this->db in the commande.class.php file. 
If two transactions have begun in parallel, I prefered to finish all the logic on $db in the card.php code.

PR #25307, opened 2023-07-07
Fix Project stock movements only displayed one way

### LINKED ISSUE #20010: Project stock movements only displayed one way
### Bug

If you make a stock correction picking some products from a warehouse and affect them in/at a project, everything's ok: in stock movements, and in project overview (to see stock movements in project overview, you've to set the conf var STOCK_MOVEMENT_INTO_PROJECT_OVERVIEW=1)

But, if, for exemple you don't need all products in the project, and you make another stock movement in the reverse way : in stock movements it's ok, but this movement doesn't appear in project overview (and benefice is also false)

### Environment Version

all including trunk

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

To correct this
In /projet/class/project.class.php
-  method get_element_list of /project.class.php, delete "AND ms.type_mouvement = 1" at the end of the sql
In /projet/element.php : delete every occurence of "abs(qty)" in all lines concerning stock_movement

_No response_

### Attached files

_No response_

vmaury:
Fix #20010 
And fix another thing : stock mouvements can now directly be attached to a project with fk_project (instead of origin_type and fk_origin)

Add a new conf constant PROJECT_ALLOW_NEGATIVE_STOCK_MVT

### DIFF
--- a/htdocs/projet/class/project.class.php
+++ b/htdocs/projet/class/project.class.php
@@ -792,7 +792,11 @@ class Project extends CommonObject
 		} elseif ($type == 'element_time') {	// Case we want to duplicate line foreach user
 			$sql = "SELECT DISTINCT pt.rowid, ptt.fk_user FROM ".MAIN_DB_PREFIX."projet_task as pt, ".MAIN_DB_PREFIX."element_time as ptt WHERE pt.rowid = ptt.fk_element AND ptt.elementtype = 'task' AND pt.fk_projet IN (".$this->db->sanitize($ids).")";
 		} elseif ($type == 'stock_mouvement') {
-			$sql = "SELECT ms.rowid, ms.fk_user_author as fk_user FROM ".MAIN_DB_PREFIX."stock_mouvement as ms, ".MAIN_DB_PREFIX."entrepot as e WHERE e.rowid = ms.fk_entrepot AND e.entity IN (".getEntity('stock').") AND ms.origintype = 'project' AND ms.fk_origin IN (".$this->db->sanitize($ids).") AND ms.type_mouvement = 1";
+			if (!empty($conf->global->PROJECT_ALLOW_NEGATIVE_STOCK_MVT)) {
+				$sql = "SELECT ms.rowid, ms.fk_user_author as fk_user FROM ".MAIN_DB_PREFIX."stock_mouvement as ms, ".MAIN_DB_PREFIX."entrepot as e WHERE e.rowid = ms.fk_entrepot AND e.entity IN (".getEntity('stock').") AND ((ms.origintype = 'project' AND ms.fk_origin IN ({$parameters['ids']})) OR ms.fk_projet in ({$parameters['ids']})) ";
+			} else {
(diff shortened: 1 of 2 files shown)


eldy:
I don't understand. 
We can already make negative stock movement. A stock movement has a direction that is the "type_mouvement" for increase or decrease (manual entry = 0, manual exit = 1, automatic exit = 2, automatic entry = 3) and value into field "value" is stored as absolute value because the sign is already stored into the type.

If we want to take into account movement that are negative movement, we must use the sign of the type_mouvement.

Also the PR is not atomic, another change is introduced. Non atomic PR are automatically discarded. Thanks to split PR into 2.

vmaury:
Yes we can already make negative stock movement.
But (may be) only in project element page, only the "positive" stock movements are considered, ie the stock movements from a warehouse to a project.
In my situation, I'm working for a company who's building prototypes. Sometimes, somebody picks a part in the stock in a warehouse, but this part doesn't fit with the project, and he has to bring it back to the warehouse.
And actually, this last movement is not considered in the project overview (element.php), because of 2 things
- all the "abs" functions in element.php
- the "AND ms.type_mouvement = 1" in get_element_list function in project.class.php.

And when I analyzed this problem, I saw they are now 2 ways to link a stock movement to a project, with fk_project or with origin_type and fk_origin. 
Ok it's another problem, but very close to the first, and solved by modifying the same sql clause ...

Thank you for all your work ! ;-)

eldy:
> Yes we can already make negative stock movement. But (may be) only in project element page, only the "positive" stock movements are considered, ie the stock movements from a warehouse to a project. In my situation, I'm working for a company who's building prototypes. Sometimes, somebody picks a part in the stock in a warehouse, but this part doesn't fit with the project, and he has to bring it back to the warehouse. And actually, this last movement is not considered in the project overview (element.php), because of 2 things
> 
> * all the "abs" functions in element.php
> * the "AND ms.type_mouvement = 1" in get_element_list function in project.class.php.
> 
> And when I analyzed this problem, I saw they are now 2 ways to link a stock movement to a project, with fk_project or with origin_type and fk_origin. Ok it's another problem, but very close to the first, and solved by modifying the same sql clause ...
> 
> Thank you for all your work ! ;-)

Your decription of your need seems to be a problem in the report of stocks movements on the project view. So a fix for this must not change the way we store stock movement,but the way it is read to make the report. The trouble seems the "AND ms.type_mouvement = 1" into the code that report list of stock movement (I think we should report all movement if they are linked to project).

For the second point, because we have a dedicated field to store the link between stock movement and project we should not use the method using origin_type and fk_origin (this one should be used when there is no dedicated field,so for other cases)

vmaury:
Stock movements are deeply modified with advanced stock transfers, and I confess not well understanding how this 2 modules are interconnected :-\
It seems to be lot confusing at least in project overview (projet/element.php) betwwen this 2 modules (stock_mouvement and stocktransfer)

eldy:
It seems this PR was set with "Discussion" or "PR to fix" tag, but no correction was provided since a long time. So request is automatically closed.

PR #25309, opened 2023-07-08
FIX unexistant POST variables price_ht_devise and price_ttc_devise in supllier invoice card

hansemschnokeloch:
# FIX
The `$_POST['price_ht_devise']` and `$_POST['price_ttc_devise'] `don't exist. The correct variables to be used instead are `multicurrency_price_ht` and `multicurrency_price_ttc`

### DIFF
--- a/htdocs/fourn/facture/card.php
+++ b/htdocs/fourn/facture/card.php
@@ -1618,11 +1618,11 @@ if (empty($reshook)) {
 				$localtax2_tx = get_localtax($tva_tx, 2, $mysoc, $object->thirdparty, $tva_npr);
 
 				$type = $productsupplier->type;
-				if (GETPOST('price_ht') != '' || GETPOST('price_ht_devise') != '') {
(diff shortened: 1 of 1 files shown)


eldy:
Can you describe the process to reproduce the bug ?

hansemschnokeloch:
You can reproduce the bug with following sequence

1. Create new multicurrency supplier invoice
2. Add a new line and enter only an amount in foreign currency, leave default currency field empty
3. Save the new line

**Result** : added line with all amounts set to zero
**Expected result** : added line with default currency amount calculated with the actual conversion rate

Input name attribute is `multicurrency_price_ht` in the code, not `price_ht_devise`
https://github.com/Dolibarr/dolibarr/blob/671a534a1caba8b23a11c74189a816411ce1f2bb/htdocs/core/tpl/objectline_create.tpl.php#L411

Bug is most likely due to this commit 
https://github.com/Dolibarr/dolibarr/commit/8debb58f7c0610150244e2ba584deda128212ef6

PR #25310, opened 2023-07-08
NEW Add an api to retrieve contacts of projects

kadogo:
NEW Add an API to retrieve contacts of projects

I didn't find a way to retrieve the contacts from a project.
I added a new API endpoint to the project side that will return the contacts of the projects (internal and external).
Furthermore, I also added a user_type element to be able to easily different the internal and external contact in the response.

I only tested the code on our test instance 14.0.5, but it merges correctly in the develop branch.
You are welcome to modify the PR if needed.

Cheers.

### DIFF
--- a/htdocs/projet/class/api_projects.class.php
+++ b/htdocs/projet/class/api_projects.class.php
@@ -20,6 +20,8 @@
 
  require_once DOL_DOCUMENT_ROOT.'/projet/class/project.class.php';
  require_once DOL_DOCUMENT_ROOT.'/projet/class/task.class.php';
+ require_once DOL_DOCUMENT_ROOT.'/contact/class/contact.class.php';
+ require_once DOL_DOCUMENT_ROOT.'/user/class/api_users.class.php';
 
 /**
  * API class for projects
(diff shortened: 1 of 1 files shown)


htdocs/projet/class/api_projects.class.php
```diff
@@ -82,6 +84,64 @@ public function get($id)
...
+                $sql .= ' AND (fk_c_type_contact=160 OR fk_c_type_contact=161 OR fk_c_type_contact=170 OR fk_c_type_contact=171)';
+
+		dol_syslog("API Rest request");
+		$result = $this->db->query($sql);
+
+		if ($result) {
+                        foreach ($result as $key => $value) {
+                            // fk_c_type_contact = 160, 161 we use users object because they are internal contacts
+                            // fk_c_type_contact = 170, 171 we use contacts object because they are external contacts
+                            if ($value["fk_c_type_contact"] == "160" || $value["fk_c_type_contact"] == "161") {
```
eldy, line 121: Hard coded value 160 and 161 are not allowed.
Take a look at other API getContacts for other objects like api_orders.php to make a similar code...

kadogo: Hello @eldy 

I don't know the complete code of Dolibarr. If you mean something like that

```
		// Add external contacts ids
		$tmparray = $this->commande->liste_contact(-1, 'external', $contact_list);
		if (is_array($tmparray)) {
			$this->commande->contacts_ids = $tmparray;
		}
		$this->commande->fetchObjectLinked();
```

It would be faster to tell me what to change and how.

Regards,

eldy: Yes, i speak about this example. Just replace  $this->commande with this->project

kadogo: Hello @eldy 

I thought I would have time to make the needed change, but it's not the case...
It would also take more time because when I will have time, I would need to dig again in the code to make the change properly.

If you have spare to do it, I guess it will be faster.
If not, I keep it on my to-do list when I have time for it.

Cheers

kadogo: Hello @eldy 

Still no time to dig again where to add the example above, but in our use case, it is interesting to be able to see the `$object->array_options` so I think it would be ok to comment `unset($object->array_options);`.

I wanted to notify you in case you had time to make the update.

Cheers

Issue #25312, opened 2023-07-09
Support search on extrafields in selectForFormsList


kubajznik:
### Feature Request

It should be possible to use filters on extrafields when referencing an object as foreign key. 
Example: `'type' => 'integer:Contact:contact/class/contact.class.php:0:(extrafieldname:=:2)'`

Right now, only the table of the object itself is loaded.

kubajznik:
Proposed solution: [https://github.com/Dolibarr/dolibarr/pull/25313](https://github.com/Dolibarr/dolibarr/pull/25313)

PR #25313, opened 2023-07-09
NEW #25312 Support extrafields in selectForForms

kubajznik:
NEW *Support filter on extrafields in selectForForms*

It should be possible to use filters on extrafields when referencing an object as foreign key.
Example: 'type' => 'integer:Contact:contact/class/contact.class.php:0:(extrafieldname:=:2)'

Right now, only the table of the object itself is loaded.


htdocs/core/class/html.form.class.php
```diff
@@ -8030,6 +8030,7 @@ public function selectForFormsList($objecttmp, $htmlname, $preselectedvalue, $sh
 
 		// Search data
 		$sql = "SELECT t.rowid, " . $fieldstoshow . " FROM " . $this->db->prefix() . $objecttmp->table_element . " as t";
+		$sql .= " LEFT JOIN " . $this->db->prefix() . $objecttmp->table_element . "_extrafields as e ON t.rowid=e.fk_object";
```
eldy, line 8033: Not that some object has no extrafields so the table does not exists.
You may add a test on if ($objecttmp->isextrafieldmanaged) before adding this left join...

kubajznik: True, thank you!

kubajznik:
Hi @eldy, is there a reason why object Contact is not extrafieldmanaged? [$isextrafieldmanaged is missing in contact.class.php](https://github.com/Dolibarr/dolibarr/blob/91020eacfadf8fc162ed5f247f3d61e7c7add005/htdocs/contact/class/contact.class.php#L62)
I believe there is the native possibility for extrafields for contacts.

eldy:
> 

No reason. The property isextrafieldmanaged was introduced after extrafields was added to Contact. I think you can add it to 1 into Contact class. Can you do it into this PR ?

kubajznik:
I added the extrafieldmanaged var to the class files for which I use extrafields.

PR #25315, opened 2023-07-10
NEW | New product stock : show the buy price of each serialized product

gdsoftdev:
On a serialized product, we want to see the buy price of each product in stock and not only PMP

### DIFF
--- a/htdocs/product/stock/product.php
+++ b/htdocs/product/stock/product.php
@@ -1178,7 +1178,23 @@ if (!$variants) {
 							print '<td class="center">'.dol_print_date($pdluo->eatby, 'day').'</td>';
 						}
 						print '<td class="right" colspan="'.$colspan.'">'.$pdluo->qty.($pdluo->qty < 0 ? ' '.img_warning() : '').'</td>';
-						print '<td colspan="4"></td>';
+						$lignepmp = 0;
+						if ($usercancreadprice) {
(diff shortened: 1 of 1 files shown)


htdocs/product/stock/product.php
```diff
@@ -1178,7 +1178,23 @@
 							print '<td class="center">'.dol_print_date($pdluo->eatby, 'day').'</td>';
 						}
 						print '<td class="right" colspan="'.$colspan.'">'.$pdluo->qty.($pdluo->qty < 0 ? ' '.img_warning() : '').'</td>';
-						print '<td colspan="4"></td>';
+						$lignepmp = 0;
+						if ($usercancreadprice) {
+							$sql = "SELECT price FROM llx_stock_mouvement WHERE fk_product = $id AND batch = '$pdluo->batch' AND (type_mouvement = 3 OR (type_mouvement = 0 AND value > 0)) ORDER BY rowid DESC LIMIT 1";
```
eldy, line 1183: SQL strings must be escaped with $db->escape($pdluo->batch).
Also, the price into stock_mouvement is the buy price when making the movement used to update the AWP (PMP) price.
The result of the select can return hundred of line. It has no sense to show one value among all of them. This is not the buy price. But a buy price at a moment in the past.

gdsoftdev: I noted the point on escaping sql strings. I will modify.
The goal here is to display the buy price of serialized products when the serial numbers are unique per product. In this case, the movement corresponds well to the buy price of this unique product.

eldy:

PR #25326, opened 2023-07-11
Fix : bookcal database structure

Hystepik:
fix booking structure of data to permit the link between availabilities and their booking

### DIFF
--- a/htdocs/install/mysql/tables/llx_bookcal_booking-bookcal.sql
+++ b/htdocs/install/mysql/tables/llx_bookcal_booking-bookcal.sql
@@ -35,6 +35,7 @@ CREATE TABLE llx_bookcal_booking(
 	lastname varchar(128) NOT NULL, 
 	email varchar(128) NOT NULL, 
 	start datetime NOT NULL, 
-	duration integer NOT NULL
+	duration integer NOT NULL, 
+	fk_bookcal_availability integer NOT NULL
 	-- END MODULEBUILDER FIELDS
 ) ENGINE=innodb;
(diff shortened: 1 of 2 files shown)


htdocs/install/mysql/tables/llx_bookcal_booking-bookcal.sql
```diff
@@ -35,6 +35,7 @@ CREATE TABLE llx_bookcal_booking(
 	lastname varchar(128) NOT NULL, 
 	email varchar(128) NOT NULL, 
 	start datetime NOT NULL, 
-	duration integer NOT NULL
+	duration integer NOT NULL, 
+	fk_bookcal_availability integer NOT NULL
```
eldy, line 39: This field is an id of another table ? Foregeign key of each table ?

For me bookcal_bookeeping is table with list of all calendars for example "Calendar for commercial contact". So it is a parent table.
Then inside a calendar, there is n entries of availabilities so table llx_boockcal_availability

Hystepik: As the module is created it's the contrary.
llx_boockcal_availabilities is the table that is a list of all the calendars and llx_bookcal_booking is a table that permit to create appointment with someone

So fk_bookcal_availability is the link between the appointment and the calendar id present in llx_boockcal_availabilities

eldy:
@Hystepik 
We are missing one table called llx_bookcal_calendar.
llx_bookcal_calendar should be the name of a calendar example "Calendar for commercial campaign 1"
Then inside this calendar, we should have llx_bookcal_availabilities. So we should have a fk_bookcal_calender key in this table.
Then the llx_bookcal_booking is when a customer register for a meeting. So your addition of the key is good, but we are missing the main calendar table, parent of ll_booking_availabilities. 
The public url to register should use the ID of the calendar table and not the id of the availability, because 1 calendar/campagin can have n different availabilities.

PR #25328, opened 2023-07-11
Change SQL request for MAX and MIN

certigna:
FIX *Change SQL for MIN and MAX search id for rowid*
*When looking for the MAX and MIN id of the rowid field (the primary key), the SQL query is not optimized.
On a table with a million records, the query takes a few seconds to get a response. With correction, the request is instantaneous*

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -2047,7 +2047,12 @@ abstract class CommonObject
 		if ($restrictiononfksoc && !$user->rights->societe->client->voir && !$socid) {
 			$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."societe_commerciaux as sc ON ".$aliastablesociete.".rowid = sc.fk_soc";
 		}
-		$sql .= " WHERE te.".$fieldid." < '".$this->db->escape($fieldid == 'rowid' ? $this->id : $this->ref)."'"; // ->ref must always be defined (set to id if field does not exists)
+    if($fieldid == 'rowid') {
+		  $sql .= " WHERE 1=1";
+    }
+    else {
+		  $sql .= " WHERE te." . $fieldid . " < '" . $this->db->escape($this->ref) . "'"; // ->ref must always be defined (set to id if field does not exists)
+    }
 		if ($restrictiononfksoc == 1 && !$user->rights->societe->client->voir && !$socid) {
 			$sql .= " AND sc.fk_user = ".((int) $user->id);
 		}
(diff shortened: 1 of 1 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -2047,7 +2047,12 @@ public function load_previous_next_ref($filter, $fieldid, $nodbprefix = 0)
 		if ($restrictiononfksoc && !$user->rights->societe->client->voir && !$socid) {
 			$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."societe_commerciaux as sc ON ".$aliastablesociete.".rowid = sc.fk_soc";
 		}
-		$sql .= " WHERE te.".$fieldid." < '".$this->db->escape($fieldid == 'rowid' ? $this->id : $this->ref)."'"; // ->ref must always be defined (set to id if field does not exists)
+    if($fieldid == 'rowid') {
+		  $sql .= " WHERE 1=1";
```
eldy, line 2051: If the where is always true, it will not return then previous ref or id.

htdocs/core/class/commonobject.class.php
```diff
@@ -2117,7 +2122,12 @@ public function load_previous_next_ref($filter, $fieldid, $nodbprefix = 0)
 		if ($restrictiononfksoc && !$user->rights->societe->client->voir && !$socid) {
 			$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."societe_commerciaux as sc ON ".$aliastablesociete.".rowid = sc.fk_soc";
 		}
-		$sql .= " WHERE te.".$fieldid." > '".$this->db->escape($fieldid == 'rowid' ? $this->id : $this->ref)."'"; // ->ref must always be defined (set to id if field does not exists)
+    if($fieldid == 'rowid') {
+		  $sql .= " WHERE 1=1";
```
eldy, line 2126: If the where is always true, it will not return then next ref or id.

eldy:
rowid in table should be a primary key, so should be indexed.So result should be immediate.
Can you paste the sql plain you have.
Just do
explain xxx
where xxx is the sql request you have

certigna:
I try with this request : SELECT MAX(te.id) FROM llx_actioncomm as te WHERE te.id < '1720110' AND te.entity IN (1);

The processing time is more than 1000 ms.
If I remove AND te.entity IN (1) the request processing time is 0.
I look indexes and there is index on both fields
I dont explain this time.

eldy:
> I try with this request : SELECT MAX(te.id) FROM llx_actioncomm as te WHERE te.id < '1720110' AND te.entity IN (1);
> 
> The processing time is more than 1000 ms. If I remove AND te.entity IN (1) the request processing time is 0. I look indexes and there is index on both fields I dont explain this time.

Try this:
- try to run update of statitics with  
`OPTIMIZE TABLE llx_actioncomm;`
then retry test on performance for the request 
`RESET QUERY CACHE; SELECT MAX(te.id) FROM llx_actioncomm as te WHERE te.id < '1720110' AND te.entity IN (1);.`
If still more than 500 ms, try this
- create an index on (entity, id)  and check time  (so 1 index on the 2 combined fields, in this order: entity, id) with `ALTER TABLE llx_actioncomm ADD INDEX idx_actioncomm_entity_id (entity, id);`
then retry test on performance for
`RESET QUERY CACHE; SELECT MAX(te.id) FROM llx_actioncomm as te WHERE te.id < '1720110' AND te.entity IN (1);.`

@certigna feedback is welcome...

PR #25332, opened 2023-07-11
qodana fix 3 : missing returns

oussamanhairech:


htdocs/core/class/translate.class.php
```diff
@@ -170,9 +170,17 @@ public function getDefaultLang($mode = 0)
 	 */
 	public function loadLangs($domains)
 	{
+		$loaded = 0;
 		foreach ($domains as $domain) {
 			$this->load($domain);
+			$result = $this->load($domain);
```
eldy, line 176: The $this->load is called twice line 175 and 176

PR #25334, opened 2023-07-11
qodana fix 4: missing returns

oussamanhairech:


htdocs/compta/sociales/class/chargesociales.class.php
```diff
@@ -730,9 +730,12 @@ public function info($id)
 			}
 
 			$this->db->free($result);
+			return 1;
 		} else {
 			dol_print_error($this->db);
+			return -1;
 		}
+		return -1;
```
eldy, line 738: This line of code seems not possible to be reached...

htdocs/core/modules/bank/doc/pdf_ban.modules.php
```diff
@@ -388,6 +388,6 @@ protected function _pagefoot(&$pdf, $object, $outputlangs, $hidefreetext = 0)
 		global $conf;
 
 		$showdetails = empty($conf->global->MAIN_GENERATE_DOCUMENTS_SHOW_FOOT_DETAILS) ? 0 : $conf->global->MAIN_GENERATE_DOCUMENTS_SHOW_FOOT_DETAILS;
-		//return pdf_pagefoot($pdf,$outputlangs,'BANK_FREE_TEXT',$this->emetteur,$this->marge_basse,$this->marge_gauche,$this->page_hauteur,$object,$showdetails,$hidefreetext);
+		return pdf_pagefoot($pdf,$outputlangs,'BANK_FREE_TEXT',$this->emetteur,$this->marge_basse,$this->marge_gauche,$this->page_hauteur,$object,$showdetails,$hidefreetext);
```
eldy, line 391: This fix the warning but it also change the behaviour by calling a method that was not called.
Call is to remove warning only.
So i suggest instead to keep the comment and add a return 1

PR #25347, opened 2023-07-12
qodana fix 5 : missing returns

oussamanhairech:


htdocs/contrat/class/contrat.class.php
```diff
@@ -753,6 +753,7 @@ public function fetch($id, $ref = '', $ref_customer = '', $ref_supplier = '')
 			$this->error = $this->db->error();
 			return -1;
 		}
+		return -1;
```
eldy, line 756: The only case we can reach this point is when we reach line 691: $result = -2
All other cases, the return is inside the if or else. So i suggest to remove this added line and instead make a return 0 line 691. 
So we match the doc comment saying that return is 0 when not found or when we found more than 1 record.

htdocs/emailcollector/class/emailcollectorfilter.class.php
```diff
@@ -427,6 +427,7 @@ public function LibStatut($status, $mode = 0)
 				return $this->labelStatus[$status].' '.img_picto($this->labelStatus[$status], 'statut5', '', false, 0, 0, '', 'valignmiddle');
 			}
 		}
+		"";
```
eldy, line 430: What is this ?

PR #25351, opened 2023-07-13
harmonize hook name

defrance:
I went around all the calls to the inihook function and I spotted names by module that are not "harmonious"
my solution to rename all the hooks is probably brutal, we could probably also use both names for a certain period but I wonder how to indicate then that there is deprecation of the old name
__ 

j'ai fait le tour de toute les appels à la fonction inihook et j'ai repéré par module des noms qui ne sont pas "harmonieux"
ma solution de renommer tout les hook est sans doute brutale,  on pourrait sans doute aussi utiliser les deux noms durant une certaine période mais je me demande comment indiquer ensuite qu'il y a dépréciation de l'ancien nom

### DIFF
--- a/htdocs/compta/bank/card.php
+++ b/htdocs/compta/bank/card.php
@@ -65,7 +65,7 @@ $extrafields = new ExtraFields($db);
 $extrafields->fetch_name_optionals_label($object->table_element);
 
 // Initialize technical object to manage hooks of page. Note that conf->hooks_modules contains array of hook context
-$hookmanager->initHooks(array('bankcard', 'globalcard'));
+$hookmanager->initHooks(array('bankaccountcard', 'globalcard'));
 
 // Security check
 $id = GETPOST("id", 'int') ? GETPOST("id", 'int') : GETPOST('ref', 'alpha');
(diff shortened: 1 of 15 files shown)


htdocs/compta/bank/card.php
```diff
@@ -65,7 +65,7 @@
 $extrafields->fetch_name_optionals_label($object->table_element);
 
 // Initialize technical object to manage hooks of page. Note that conf->hooks_modules contains array of hook context
-$hookmanager->initHooks(array('bankcard', 'globalcard'));
+$hookmanager->initHooks(array('bankaccountcard', 'globalcard'));
```
eldy, line 68: To remove hardcoded value, context for hook on object card should now be
`$hookmanager->initHooks(array($object->element.'card', 'globalcard'));`

fappels:
For indicate deprecated, think at least a list of changed ones should be in changelog to have a quick check in module config to see if one is used.

eldy (commented):
The new name may also be not consistent.
With 18.0, hookname now use a common syntax that is
$object->element.'card'
Or
$object->element.'list'

If we make a change that will break compatibility, we should do it for the new  generic syntax and we must complete the changelog with list of hook name that was renamed (old name -> new name).
To keep a period of deprecation, i also suggest to keep old name at least 1 or 2 versions...

eldy:

PR #25372, opened 2023-07-17
FIX issue #25353 for get the correct value of product id

### LINKED ISSUE #25353: Import material line
### Bug

When you are importing BOM lines, it works wonders. But, when you're creating the MO, it doesn't detect correctly some of the IDs. I've found that most of the times the error is where the reference is number-based, if it has some text, it works fine.
image
****
image

### Environment Version

17.0.2

### Environment OS

Linux

### Environment Web server

Linux

### Environment PHP

7.4.33

### Environment Database

MySQLI

### Environment URL(s)

_No response_

It should be able to create the MO, but it is not.

_No response_

### Attached files

_No response_

lamrani002:
# FIX|Fix #25353 when try to add new MO the value of fk_product was incorrect

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -3126,7 +3126,7 @@ if ($action == 'create') {
 
 	// when bank account is empty (means not override by payment mode form a other object, like third-party), try to use default value
 	if ($socid > 0 && $fk_account) {	// A company has already been set and it has a default fk_account
-		$fk_account = GETPOSTISSET('fk_account') ? GETPOST("fk_account", 'int') : $fk_account;	// The GETPOST is used only if form was posted to avoid to take default value, because in such case, the default must be the one of the company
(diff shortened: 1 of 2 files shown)


htdocs/compta/facture/card.php
```diff
@@ -3126,7 +3126,7 @@
 
 	// when bank account is empty (means not override by payment mode form a other object, like third-party), try to use default value
 	if ($socid > 0 && $fk_account) {	// A company has already been set and it has a default fk_account
-		$fk_account = GETPOSTISSET('fk_account') ? GETPOST("fk_account", 'int') : $fk_account;	// The GETPOST is used only if form was posted to avoid to take default value, because in such case, the default must be the one of the company
+		$fk_account = GETPOSTISSET('fk_account') ? $fk_account : GETPOST("fk_account", 'int');	// The GETPOST is used only if form was posted to avoid to take default value, because in such case, the default must be the one of the company
```
eldy, line 3129: This change looks strange. Is it related with the bug number ? If not, can you make a PR with the bug fix and only the bug fix ?

lamrani002: my bad , i did a mistake

eldy:
Old code seems good to me. I comment with my hypothesis in the issue ticket...

PR #25375, opened 2023-07-17
New | Add AGENDA_SORT_EVENT_TYPE_BY_POSITION_FIRST hidden conf to change sort method of actions

hansemschnokeloch:
# New | Use position as main sort criterion
Currently the `position` defined in the dictionnary isn't the main sort criterion. This prevents you from ordering the actions as you wish.
This PR changes the sort criterions order, and defines the `position` as main sort criterion


eldy:
Yes goal of position is to sort record among a given type. Prority is on type because we always want to have actions of the same type together...

hansemschnokeloch:
I see your point.
But sadly this prevents you to put the most used items in top of the select when you create manually an event.

eldy:
> I see your point.
> But sadly this prevents you to put the most used items in top of the select when you create manually an event.
> 

I see. I suggest to introduce a hidden conf (no need for setup, just the test for one order by or thevother) to allow your behaviour for integrators.

hansemschnokeloch:
> I see. I suggest to introduce a hidden conf (no need for setup, just the test for one order by or thevother) to allow your behaviour for integrators.

I added a hidden conf. I didn't find any naming convention in the docs for such variables. I hope the name is fine, it should be self-explanatory enough.

hansemschnokeloch:
@eldy PR ready for review

PR #25377, opened 2023-07-17
FIX - V19 is not an SQL instruction ;)

altairis-noe:
I tried to migrate to the dev sql schema… it did not work 😇


frederic34:
you can also change `when current version is 18.0.0 or higher.` to 19

altairis-noe:
> you can also change `when current version is 18.0.0 or higher.` to 19

Good comment @frederic34. Done 🙂

PR #25384, opened 2023-07-18
FIX Fatal Error PHP 8 : division by 0

atm-lena:
# FIX|Fix #

Same fix of https://github.com/Dolibarr/dolibarr/pull/25373/files for v17

### DIFF
--- a/htdocs/bom/tpl/objectline_view.tpl.php
+++ b/htdocs/bom/tpl/objectline_view.tpl.php
@@ -332,9 +332,11 @@ if ($resql) {
 			$resql_supplier_price = $object->db->query($sql_supplier_price);
 			if ($resql_supplier_price) {
 				$obj = $object->db->fetch_object($resql_supplier_price);
-                if(!empty($obj->qty) && !empty($sub_bom_line->qty) && !empty($line->qty)) $line_cost = $obj->min_price/$obj->qty * $sub_bom_line->qty * $line->qty;
-                else $line_cost = $obj->min_price;
-
+                if (!empty($obj->qty) && !empty($sub_bom_line->qty) && !empty($line->qty)) {
+					$line_cost = $obj->min_price/$obj->qty * $sub_bom_line->qty * $line->qty;
+				} else {
+					$line_cost = $obj->min_price;
+				}
 				print '<td class="linecolcost nowrap right" id="sub_bom_cost_'.$sub_bom_line->id.'"><span class="amount">'.price2num($line_cost, 'MT').'</span></td>';
 				$total_cost+= $line_cost;
 			}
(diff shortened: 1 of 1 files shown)


htdocs/bom/tpl/objectline_view.tpl.php
```diff
@@ -332,8 +332,11 @@
 			$resql_supplier_price = $object->db->query($sql_supplier_price);
 			if ($resql_supplier_price) {
 				$obj = $object->db->fetch_object($resql_supplier_price);
-				$line_cost = $obj->min_price/$obj->qty * $sub_bom_line->qty * $line->qty;
-
+                if (!empty($obj->qty) && !empty($sub_bom_line->qty) && !empty($line->qty)) {
+					$line_cost = $obj->min_price/$obj->qty * $sub_bom_line->qty * $line->qty;
```
eldy, line 336: Don't you think we should, instead of just doing
`line_cost = $obj->min_price`
replace each qy with 1 if not defined. So we would get
`$line_cost = $obj->min_price / (empty($obj->qty) ? 1 : $obj->qty) * (empty($sub_bom_line->qty) ? 1 : $sub_bom_line->qty) * (empty($line->qty) ? 1 : $line->qty)`;
Because a not defined qty means 1 i think ?

hansemschnokeloch: What if a qty is defined as 0 (not quite sure if it's possible in this case) ?
But with the check on empty you will assume that is value is 1

eldy: > What if a qty is defined as 0 (not quite sure if it's possible in this case) ? But with the check on empty you will assume that his value is 1

On BOM, having a qty of 0 of a product should be same than not having the line at all. So this should not happen.

atm-thibaultf: I think it's important to allow to have a quantity of 0 for example for product for which we don't know the exact quantity we are going to use. We need to have it in the BOM but we don't know the qty so we can put 0 just to have the information that we need this product. Also it can be usefull for products (like consumables) we need only if something happens during the production, it's good to have it in the BOM so the operator can consume this product if needed.

eldy: > I think it's important to allow to have a quantity of 0 for example for product for which we don't know the exact quantity we are going to use. We need to have it in the BOM but we don't know the qty so we can put 0 just to have the information that we need this product. Also it can be usefull for products (like consumables) we need only if something happens during the production, it's good to have it in the BOM so the operator can consume this product if needed.

Good catch. Note that currently on v18, using a 0 qty is not possible. For your use case, we need to enter 1

PR #25396, opened 2023-07-19
FIX: Update mouvementstock.class.php

prietojc:
If first character changes to upper case, some external modules don't show its origin in mouvements.

### DIFF
--- a/htdocs/product/stock/class/mouvementstock.class.php
+++ b/htdocs/product/stock/class/mouvementstock.class.php
@@ -975,7 +975,7 @@ class MouvementStock extends CommonObject
 				if ($origin_type) {
 					// Separate originetype with "@" : left part is class name, right part is module name
 					$origin_type_array = explode('@', $origin_type);
-					$classname = ucfirst($origin_type_array[0]);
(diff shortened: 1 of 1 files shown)


htdocs/product/stock/class/mouvementstock.class.php
```diff
@@ -975,7 +975,7 @@ public function get_origin($origin_id, $origin_type)
 				if ($origin_type) {
 					// Separate originetype with "@" : left part is class name, right part is module name
 					$origin_type_array = explode('@', $origin_type);
-					$classname = ucfirst($origin_type_array[0]);
+					$classname = $origin_type_array[0];
```
eldy, line 978: Class name should start with uppercase.
Contact, Warehouse, User, ...
This may generates a lot of trouble if classname is only  contact, warehouse, user...

prietojc: All right I agree,   but uppercase is already done on line 983...
Module name and URL use to be in lowercase.
I propose a new commint more simple

PR #25398, opened 2023-07-19
NEW Upgrade in module builder in menu section

lamrani002:
# NEW|New upgrading on section menu for module builder by add a script for auto fill input field and new delete msg for confirming remove menu parent

### DIFF
--- a/htdocs/theme/eldy/global.inc.php
+++ b/htdocs/theme/eldy/global.inc.php
@@ -7994,10 +7994,6 @@ table.jPicker {
 	}
 }
 
-/* for hide a select */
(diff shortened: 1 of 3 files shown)


htdocs/theme/eldy/global.inc.php
```diff
@@ -7994,6 +7994,11 @@
 	}
 }
 
+/* for hide a select */
+.hideSelect {
```
eldy, line 7998: There is already a class called "hideobject" with the display: none. 
Can you use this one instead of adding a new one into the CSS ?

PR #25439, opened 2023-07-23
Create box_products.php

dolibit-ut:
box_products  copied  from  box_produit 

- new label for future standardized uniform english designations
- keep old file for a certain time for compatibility


eldy:
We can make transition with compatibility without loosing setup this way:
- remove also old file.
- add into the file htdocs/install/mysql/migration/18.0.0-19.0.0.sql, a sql line to rename box_produits.php into box_product.php
- update also into the file modProduct.class.php

eldy:

PR #25448, opened 2023-07-24
Fix : don't display birthdays when 0 years

atm-maxime:


htdocs/core/boxes/box_birthdays.php
```diff
@@ -90,12 +90,14 @@ public function loadBox($max = 20)
 			$sql .= " WHERE u.entity IN (".getEntity('user').")";
 			$sql .= " AND u.statut = ".User::STATUS_ENABLED;
 			$sql .= dolSqlDateFilter('u.birth', 0, $tmparray['mon'], 0);
+			$sql .= " AND YEAR(u.birth) != ".$tmparray['year'];
```
eldy, line 93: Use of SQL date function YEAR must be avoided. See why on https://wiki.dolibarr.org/index.php?title=Language_and_development_rules#SQL_Coding_rules
Instead we can use this
`$sql .= " AND u.birth >= '".$db->idate(dol_get_first_day($tmparray['year']))."'";`

atm-maxime: Sorry, I'm not up to date with SQL rules :)
Fix done

PR #25465, opened 2023-07-24
FIX Handle projects for the api documents listing route

boherm:
(I'm pretty glad to make my first contribution for Dolibarr after some years to work with!)

# Fix Handle projects for the api documents listing route

After some tests, I noticed that the api documents listing route `GET /api/index.php/documents` don't handle linked documents projects.

This PR fix this route and add the possibility to list all uploaded documents for one project given.

(I have target the `17.0` but I can rebase if needed! 😉)


boherm:
Thanks @eldy! 🙂

Issue #25370, opened 2023-07-17
PRODUCT_USE_UNITS is hidden


sonikf:
### Bug

In https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/product/admin/product.php

This feature is hidden for no apparent reason, i can't find the duplicated feature! and i can't find another way to enable units.

```
// Use units
/* Kept as hidden feature only. PRODUCT_USE_UNITS is hidden for the moment. Because it seems to be a duplicated feature with already existing field to store unit of product

print '<tr class="oddeven">';
print '<td>'.$langs->trans("UseUnits").'</td>';
print '<td class="right">';
print $form->selectyesno("activate_units",$conf->global->PRODUCT_USE_UNITS,1);
print '</td>';
print '</tr>';
*/
```
Also in https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/core/tpl/objectline_edit.tpl.php  line 265

`if (!empty($line->fk_unit) && empty($conf->global->MAIN_EDIT_LINE_ALLOW_ALL_UNIT_TYPE)) {`

@eldy should i enable PRODUCT_USE_UNITS and add MAIN_EDIT_LINE_ALLOW_ALL_UNIT_TYPE too?

ksar-ksar:
Hello,

These hidden variables can be activated through set-up other : https://wiki.dolibarr.org/index.php?title=Setup_Other

sonikf:
Hi @ksar-ksar,
Ι know this, my question is why are they hidden?
I'm working on merge requests to implement necessary changes for Dolibarr to work/comply with Greek standards  and  units is one of them.
Also MAIN_EDIT_LINE_ALLOW_ALL_UNIT_TYPE is not documented in https://wiki.dolibarr.org/index.php?title=Setup_Other. Should i add it there too?
Thanks

ksar-ksar:
Hello,

They are hidden because they are not stable or deprecated :` Some options are not visible in the interface (often because they are deprecated or are beta features). In most cases you do not need these functions and their use may make the application unstable. `

You can add any missing hidden variables in the wiki

sonikf:
I've read the wiki but this feature maybe not  fully implemented but it is stable. Greek users are using it for years now with out problems.

Daviid-P:
> I've read the wiki but this feature maybe not fully implemented but it is stable. Greek users are using it for years now with out problems.

"Hasn't broken yet" is not the same as "Stable"

[...]

Issue #25366, opened 2023-07-15
Add product units on Manufacturing order and Warehouse documents


vdegrandpre:
### Feature Request

This feature request is coming from a customer request.

Many documents implement the PRODUCT_USE_UNITS global variable to show the product units aside the quantities, but in some documents it is missing.

User story : As a user, I want to see the product units in Manufacturing orders and Warehouse documents so they can be clearly seen by our staff during operations.

### Use case

When PRODUCT_USE_UNITS is enabled, the system shall show the product units aside quantity on every documents.

PR #25489, opened 2023-07-25
Enable by default INVOICE_CREDIT_NOTE_STANDALONE and PRODUCT_USE_UNITS for Greece

sonikf:
Hi eldy 
Regarding PRODUCT_USE_UNITS and this issue #25370
The use case for Greece is to output the unit label in documents(mainly invoices and shipment sheets) and upload it as part of the invoice or shipment sheet data to the ministry db. 
We only use piece, kg and lt. 

From this 8 year old discussion nearly the same rules apply for Mexico
https://github.com/Dolibarr/dolibarr/commit/d843b1fbe442a198103cb9a66bcae4bd4fb08e35#diff-3a48cb158377e8951e07df9f6bb926c3932d307842720f379acdb0ebea36e3a0R436

So far this is the only implemented method for this use case(half baked as it is ) but it is hidden and not documented other than in [Setup_Other](https://wiki.dolibarr.org/index.php/Setup_Other) wiki page

Please instruct me how would you like this to be properly implemented so i can try and fix it

### DIFF
--- a/htdocs/master.inc.php
+++ b/htdocs/master.inc.php
@@ -241,6 +241,14 @@ if (!defined('NOREQUIREDB') && !defined('NOREQUIRESOC')) {
 		// The deposit invoice type is not allowed in Greece.
 		$conf->global->INVOICE_DISABLE_DEPOSIT = 1;
 	}
+	if ($mysoc->country_code == 'GR' && !isset($conf->global->INVOICE_CREDIT_NOTE_STANDALONE)) {
+		// Standalone credit note is compulsory in Greece.
+		$conf->global->INVOICE_CREDIT_NOTE_STANDALONE = 1;
+	}
+	if ($mysoc->country_code == 'GR' && !isset($conf->global->PRODUCT_USE_UNITS)) {
+		// Unit labels are required in Greek invoices.
+		$conf->global->PRODUCT_USE_UNITS = 1;
+	}
 	if (($mysoc->localtax1_assuj || $mysoc->localtax2_assuj) && !isset($conf->global->MAIN_NO_INPUT_PRICE_WITH_TAX)) {
 		// For countries using the 2nd or 3rd tax, we disable input/edit of lines using the price including tax (because 2nb and 3rd tax not yet taken into account).
 		// Work In Progress to support all taxes into unit price entry when MAIN_UNIT_PRICE_WITH_TAX_IS_FOR_ALL_TAXES is set.


eldy:
Using PRODUCT_USE_UNITS seems to be a bad practice. 
The unit of a product is specific to the product ref. A product should be defined with the lowest used unit. 
A product sell in meters should not be the same product than the one sell in cm or kg. Because quantities does not match.
A price for 1 meter will also differs of the price for 1 cm.
You have the same problem with price, size, volume and weight calculation. If product ref is defined with property price=10, weight= 1kg, volume = 1 m3, when qty is 2, price should be 20, weight should be 2kg and volume = 2 m3 (price, weight and volume are properties of a product ref). If when you make your proposal, you say qty 2, but with a unit of gram, you will still get a price of 20, weight calculated for shipment of 2 kg and a volume calculated for shipment 2 m3,  but for a 2 g line, all this values are wrong. You will also have trouble on process of discount per quantity that will be wrong, and more...
For this reason, unit should not change for 1 product ref. And the option PRODUCT_USE_UNITS that allow to show another unit is hidden because most feature can't work this way.
If you need to sell both in g and kg, unit should be g and when selling 2 kg, qty should be 2000.

sonikf:
@eldy 
I understand your reasoning and PRODUCT_USE_UNITS value should really be saved in a different field other than llx_product.fk_unit the way it is used

Based on that my take on this problem would be the use of 4 different unit fields in order to tackle all the use cases and conditions
1. main unit for accounting, stock  based on the lowest unit         e.g. beer unit=piece
2. secondary unit that would be a divisor or multiplier depending on the use case and maybe in a locked relation with main unit so that values are autoupdated autocalculated in screens, documents e.g. multiplier six packs, dozens(12 pieces)...or divisor for volume weight length...the user could choose that way what's best
3. purchase unit for use in supplier documents because we may buy a pallet containing 20 dozens of sodas 
4. sale unit for use in documents(this is actually the use case for this mr)

What do you think? 

Is it ok to  resubmit with only INVOICE_CREDIT_NOTE_STANDALONE?

Daviid-P:
@sonikf 

1. main unit for accounting, stock based on the lowest unit e.g. beer unit=piece
 - If you can buy by the can then you should have a product for a single can of beer
2. secondary unit that would be a divisor or multiplier depending on the use case and maybe in a locked relation with main unit so that values are autoupdated autocalculated in screens, documents e.g. multiplier six packs, dozens(12 pieces)...or divisor for volume weight length...the user could choose that way what's best
 - If you can also buy in packs of 6 then you should have a different product for the 6-pack
3. purchase unit for use in supplier documents because we may buy a pallet containing 20 dozens of sodas
 - If you can buy a pallet then you should have a 3rd product that's the pallet.
4. sale unit for use in documents(this is actually the use case for this mr)
 - Shouldn't sale unit be how many multiples of the product you sold? I don't think that should carry a unit, the unit should go in the product title/description.

chrome_OJHdvxyXBh

Or are you implying you'd like the red square column to say something like:
8x330ml = 2,640ml
10*6*330ml = 19,800ml
2*20*6*330ml = 79,200ml


I'm not usre I understand but this doesn't seem necessary.

sonikf:
@Daviid-P 
@eldy 

This is a thought exercise but I hope it leads to the start of a solution.
After reading all the issues and posts here and in all the forums, this picture from this post https://www.dolibarr.org/forum/t/units-implementation-in-dolibarr/22012/6 speaks a thousand words
c5fb44cee25699559c4f043eb2d273b0313e74db

Lets say for the sake of the discussion that I own a small market

1. main unit for accounting, stock based on the lowest unit e.g. beer unit=piece
-     If you can buy by the can then you should have a product for a single can of beer
My selling products are: single can of beer, six pack and box of dozen(which could be either 12 cans or 2 six packs) but not pallets because e.g. I don't have the space in my warehouse.
I can buy six packs, boxes of dozen or pallets from supplier but single cans are not sold by suppliers and or manufacturers

2. secondary unit that would be a divisor or multiplier depending on the use case and maybe in a locked relation with main unit so that values are autoupdated autocalculated(**EDIT: in DB**), screens, documents e.g. multiplier six packs, dozens(12 pieces)...or divisor for volume weight length(**EDIT: or vice versa**)...the user could choose that way what's best
-     If you can also buy in packs of 6 then you should have a different product for the 6-pack
I already have that product but at the end of the day,  I also care about the buying price of a single can so that i can calculate my selling price e.g. 1 can = 1 euro      1 six pack = 6 euro....

Maybe 3 and 4 can be skipped if PRODUCT_USE_UNITS was to be reworked somehow

3. purchase unit for use in supplier documents because we may buy a pallet containing 20 dozens of sodas
-     If you can buy a pallet then you should have a 3rd product that's the pallet.
For reasons stated above i also have a pallet product but for buy only, again i need to calculate my selling price and to quote eldy
from this post https://www.dolibarr.org/forum/t/units-implementation-in-dolibarr/22012
> But currently on purchase prices, we only record the quantity of a package without the label.

Label is a requirement too in some cases.

4. sale unit for use in documents(this is actually the use case for this mr)
-     Shouldn't sale unit be how many multiples of the product you sold? I don't think that should carry a unit, the unit should go in the product title/description.
 
That's right but the requirement per the ministry directive is for the unit type label to be part of the line in documents like 

| Description | Qty| Unit | U.P. (net |
| ----------------------  | ------------- | ------------- | ------------- |
| Beer can of 330 ml  | 5 | Pieces | 1,00 |

and also part of the xml body  e.g. `<measunit>'.$unit.'</measunit>`

sorry if I didn't communicate this more clearly.


As per this issue #25366 I discovered PRODUCT_USE_UNITS after working with pdf templates and i assumed it was the default way of showing product unit type in lines.

Obviously this use case affects at least 3 countries and right now it is used in the wild with implications if  mixed with other use cases.

PR #25492, opened 2023-07-26
Fix count cronjoblist

atm-quentin:
# FIX count cronjoblist
Numbers of lines sometimes differs from the nb count on cronjob list because of test column.
I move the test before request to solve this

### DIFF
--- a/htdocs/cron/list.php
+++ b/htdocs/cron/list.php
@@ -318,7 +318,6 @@ $sql .= $db->order($sortfield, $sortorder);
 $nbtotalofrecords = '';
 if (empty($conf->global->MAIN_DISABLE_FULL_SCANLIST)) {
 	$result = $db->query($sql);
-	print $sql;
 	$nbtotalofrecords = $db->num_rows($result);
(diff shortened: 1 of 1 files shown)


htdocs/cron/list.php
```diff
@@ -239,6 +239,18 @@
...
 llxHeader('', $pagetitle);
 
+$sqlTest = 'SELECT DISTINCT test FROM '.MAIN_DB_PREFIX.'cronjob';
+$resultTest = $db->query($sqlTest);
+$TTestNotAllowed = array();
+if($resultTest) {
+	while ($objTest = $db->fetch_object($resultTest)) {
+		$veriftest = verifCond($objTest->test);
+			if (!$veriftest) {
+				$TTestNotAllowed[] = $db->escape($objTest->test);
```
eldy, line 249: If goal is to exclude some record later, we should store an id into array instead of field test.

atm-quentin: Why do you prefer id ? If some lines have the same test we'll pass multiple time in verif cond with same test

eldy: Because later you use the array TTestNotAllowed to exclude some line. And what identify a line is its ID.

PR #25500, opened 2023-07-26
Fix inconsistent is-empty check for extrafields

mc2contributor:
# FIX inconsistent is-empty check for extrafields
Checks for empty extrafields values were in two different places, and behaved slightly differently.  This puts the code in one place to ensure consistent behaviour.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -6187,7 +6187,7 @@ abstract class CommonObject
 				// Similar code than into insertExtraFields
 				if ($attributeRequired) {
 					$v = $this->array_options[$key];
-					if (ExtraFields::isEmptyValue($v,$attributeType)) {
+					if (ExtraFields::isEmptyValue($v, $attributeType)) {
 						$langs->load("errors");
 						dol_syslog("Mandatory field '".$key."' is empty during create and set to required into definition of extrafields");
 						$this->errors[] = $langs->trans('ErrorFieldRequired', $attributeLabel);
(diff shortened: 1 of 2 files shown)


htdocs/core/class/extrafields.class.php
```diff
@@ -2343,4 +2342,14 @@ public function getOptionalsFromPost($extrafieldsobjectkey, $keyprefix = '', $ke
 
 		return 0;
 	}
+
+	public static function isEmptyValue($v, string $type) {
+		if ($v === null || $v === '') $isEmpty = true;
+		elseif (is_array($v) || $type == 'select') $isEmpty = empty($v);
+		elseif ($type == 'link') $isEmpty = $v == '-1';
```
eldy, line 2349: What mean $isEmpty = $v == '-1';
Can you inclue parenthesis so code would be clear ?

mc2contributor: Parentheses added for clarity.

htdocs/core/class/commonobject.class.php
```diff
@@ -6169,6 +6169,8 @@ public function insertExtraFields($trigger = '', $userused = null)
 				}
 			}
 
+			include_once DOL_DOCUMENT_ROOT.'/core/class/extrafields.class.php';
```
eldy, line 6172: This has been already added line 6158

hansemschnokeloch:
Hi,

Usage of elseif whithout curly braces {} adds confusion to the code - see [PSR-2](https://www.php-fig.org/psr/psr-2/#51-if-elseif-else).
Following code is simplier to read, without need of $isEmpty variable.
```php
	public static function isEmptyValue($v, string $type) {
		if ($v === null || $v === '') return true;
		if (is_array($v) || $type == 'select') return empty($v);
		if ($type == 'link') return ($v == '-1');
		if ($type == 'sellist') return ($v == '0');
		return (empty($v) && $v != '0');
	}

```

PR #25501, opened 2023-07-26
Fix default value for extrafields to avoid SQL 'cannot be null' error

mc2contributor:
# FIX default value for extrafields to avoid SQL 'cannot be null' error
Empty extrafields were being set to null, rather than using the `fielddefault` field from the `llx_extrafields` table.  This could result in a "Field 'xxx' cannot be null" error from MySQL.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -6169,6 +6169,7 @@ abstract class CommonObject
 				}
 			}
 
+			$attrs = $extrafields->attributes[$this->table_element];
 			foreach ($new_array_options as $key => $value) {
 				$attributeKey      = substr($key, 8); // Remove 'options_' prefix
 				$attributeType     = $extrafields->attributes[$this->table_element]['type'][$attributeKey];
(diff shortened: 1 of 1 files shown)


eldy:
Can you describe a step by step process to reproduce the bug ?

mc2contributor:
I added a field to the `llx_projet_task_extrafields` table:

```
taskType int(11) NOT NULL DEFAULT 0
```

with corresponding entry in `llx_extrafields`:

```
rowid          | 37
name           | taskType
entity         | 1
elementtype    | projet_task
label          | TaskType
type           | int
size           | 10
fieldcomputed  | <null>
fielddefault   | 0
fieldunique    | 0
fieldrequired  | 0
...
```

Then in the Gantt Pro Advanced module I change the task end date by dragging it on the Gantt chart.  This avoids the usual form entry page, and the value is submitted as an empty string, causing it to fall through to the null case at line 6367 of commonobject.class.php.  This results in a database error because NULL is not allowed.

The default value should not be hard-coded as `null` (line 6367), but should come from `llx_extrafields.fielddefault`.

eldy:
The default value of a database is used when a value is NOT provided, not to replace a value explicitly provided by the code. If value is provided to null into the sql there is a bug somewhere else.
Also if you set a field with a NOT NULL in database but the entry into llx_extrafields contains fieldrequired = 0 ? This means you database does not match the description.
How did you create the extrafields ? From the web interface to add extrafields ?

PR #25502, opened 2023-07-27
prevent error with php 8.2

atm-jpb:
# FIX prevent error with php 8.2

Fatal error: Uncaught TypeError: round(): Argument #1 ($num) must be of type int|float occur  with php 8.2

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="JavaScriptSettings">
+    <option name="languageLevel" value="ES6" />
+  </component>
+</project>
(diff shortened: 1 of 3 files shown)


hansemschnokeloch:
Why not update line 962 like this ?
`
$size = (float)$this->attributes[$extrafieldsobjectkey]['size'][$key];
`

hansemschnokeloch:
Please don't commit your **.idea** folder, [set up a global gitignore file](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer)

.idea/misc.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
eldy, line 1: Do we need this file or is it a commit error ?

PR #25503, opened 2023-07-27
Update interface_20_all_Logevents.class.php

dolibit-ut:
$langs->load("users");     only once at the start

### DIFF
--- a/htdocs/core/triggers/interface_20_all_Logevents.class.php
+++ b/htdocs/core/triggers/interface_20_all_Logevents.class.php
@@ -79,6 +79,7 @@ class InterfaceLogevents extends DolibarrTriggers
 
 		/* Actions */
 
+		$txt = $action;
 		$langs->load("users");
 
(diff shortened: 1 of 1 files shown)


htdocs/core/triggers/interface_20_all_Logevents.class.php
```diff
@@ -77,54 +77,58 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
 
 		$date = dol_now();
 
-		// Actions
+		/* Actions */
+
+		$langs->load("users");
```
eldy, line 82: We don't want to load the file if we don't need it. So if action is not a USER_... action, we should not load it.

dolibit-ut: Yes -  I agree,
but before the file was loaded for every action and it seems that it is always needed - if this module is active.
??

PR #25496, opened 2023-07-26
NEW LDAP Active Directory UserAccountControl

ibuiv:
# NEW|New [UserAccountControl can be set]
In the mode Dolibarr -> LDAP sync with activedirectory, there were some issues.

Now, in the LDAP Admin page, you can set a default userAccountControl for user creation (with examples)

The creation and the enable/disable action now updates this param while updating your ldap entry.

PR #25507, opened 2023-07-28
NEW LDAP Active Directory UserAccountControl

ibuiv:
In the mode Dolibarr -> LDAP sync with activedirectory, there were some issues.

Now, in the LDAP Admin page, you can set a default userAccountControl for user creation (with examples)

The creation and the enable/disable action now updates this param while updating your ldap entry.

#25496 on branch develop instead of 16.0


ibuiv:
@eldy
I pushed it on develop branch as you asked.

Issue #25335, opened 2023-07-11
Tools --> Email Templates section visible for all users


elpreto:
### Bug

All users can see the Tools --> Email Templates section.

The section is empty for the users with no authorization, but still visible, even with the setting to hide all sections with no authorizations.

### Environment Version

17.0.2

### Environment OS

Linux br-asc-web888.main-hosting.eu 4.18.0-425.3.1.lve.3.el8.x86_64

### Environment Web server

LiteSpeed

### Environment PHP

8.0.28

### Environment Database

MariaDB 10.6.12-MariaDB-cll-lve

### Environment URL(s)

admin/mails_templates.php?leftmenu=email_templates

### Expected and actual behavior

Section shuold be hidden, or have an option to hide it

### Steps to reproduce the behavior

Login with a user with no authorizations, and check the Tools --> Email Templates section.

### Attached files

![email templates](https://github.com/Dolibarr/dolibarr/assets/13504962/fdaf398c-fb71-42f3-8ba1-93e9fdba6491)

oussamanhairech:
This is not considered a bug since this menu is accessible without permission. Users can define their own email templates without needing any permission.

PR #25513, opened 2023-07-28
fix : #25335 Tools --> Email Templates section visible for all users

oussamanhairech:
In this fix, if the constant MAIN_BUTTON_HIDE_UNAUTHORIZED is activated the section email templates is hidden.


atm-maxime:
This menu is the one accessible with no specific permission. So with your fix, no one will be able to see this menu.
Users can define their own email templates with no permission needed.

PR #25517, opened 2023-07-29
Societe.class: refactor create() to early return on the error path 

alexandre-janniaux:
This is a small refactor PR to favour readability of the code by making the path returning an error closer to the call that generate the error, and only indent to handle the error and not in the case no error happened (early return).

It also removes a check on `$error` where it is only set after the condition itself, so the test was always valid.


eldy:

PR #25523, opened 2023-07-31
fix error seteventmessage on white screen

atm-greg:
# FIX Error seteventmessage on white screen on holiday creation

When the current user creates an holiday for someone else which id is not in $childids array and without type, the seteventmessage "fieldRequired Type" is displayed on a white screen...

All the display is in the else clause of the "if" on line 933, so there is no display

I've just moved the ending bracket of the else to diplay creation for in each case.


atm-greg:
@eldy it's ready to merge

PR #25525, opened 2023-07-31
Use SSL connection to MySQL server if server name starts with ssl://

### LINKED ISSUE #24404: remote mysql server with ssl transport ?
### Feature Request

Hello all,
i'm looking for a way to configure dolibarr with a remote mysql/mariadb server using ssl layer

> php mysqli will do it with https://www.php.net/manual/fr/mysqli.ssl-set.php

but i can't find any trace of ssl_set into dolibarr php code

may i make a PR for that or that's not important ?

### Use case

_No response_

### Suggested implementation

_No response_

### Suggested steps

_No response_

ol-fr:
CLOSE #24404 Support secure connections using SSL on MySQL server

When using MySQL, if server name starts with ssl://, connection to server will force use of SSL.

I had to extend mysqli class in order to implement this feature properly with ::init, ::ssl_set, ::real_connect.


hansemschnokeloch:
Why not use a configuration variable like `$dolibarr_main_db_ssl = true;` in the configuration file of dolibarr instead of prefixing the host with `ssl://` ?
The whole configuration is done this way, is there a specific reason why this shouldn’t be the case here ?

ol-fr:
That's was a choice.
Using `[protocol]://` before server name is very common. 
We could also get rid of `$dolibarr_main_db_port` as most of installations use default port (and localhost); and if the port is not set or equal to 0, mysqli and pgsql will use default one.
So with this behavior, I also make the economy of creating one more configuration variable which will need to be passed over several function parameters, and managed in install process and default conf files.

PR #25537, opened 2023-08-02
FIX_reload_linked_objects_on_propal_closeas

atm-thibaultf:
# FIX
Needed if object linked modified by trigger on propal closeas (because linked objects can't be fetched two times : linkedObjectsFullLoaded)

# USE CASE
We want to linked order with propal on propal signed trigger but the attribute linkedObjectsFullLoaded is already set before the trigger, so the linked objects by the trigger can't be reloaded by the fetchObjectLinked(). So we need to reload the page manually to show new linked objects.

# SECURITY ENHANCE
Furthermore, don't show arg in the URL is better for the final user (prevent doing action again on refresh page).


atm-arnaud:
👍 nice for users

eldy:
Note:
If a trigger modify content in database like object linked, it is also role of the trigger to modify the $object in memory (data in memory must reflect data in database).
To allow a reload of object linked of an object, you can call
$object->clearObjectLinkedCache()
before calling 
$object->fetchObjectLinked()

Reload to avoid action=... in URL however is a good reason

PR #25538, opened 2023-08-02
fix: on contact list if thirdparty is only supplier the link comm/card php is bad regarding perms tested and tabs preselected

FHenry:
On contact list if thirdparty is only supplier the url link comm/card.php  give and accessforbiden because
of 
https://github.com/Dolibarr/dolibarr/blob/3d65f576acb659ee77ea8efafa673977f1666501/htdocs/comm/card.php#L134

Antoher fix could be to change this line to 
`if ((!($object->client > 0) && !($object->fournisseur > 0) )|| !$user->hasRight('societe', 'lire')) {`

but in this case societe_prepare_head() do not selected the good first tabs (none selected...)

this fix seems to be the more harmless


frederic34:
@FHenry a new menu manager ? It looks like a module wrote some files into the core

FHenry:
@frederic34 No, my bad, I let some custom files introduce in the commit, from dropdrown menu module
https://github.com/FHenry/dolibarr_dropdownmenu.git

PR #25544, opened 2023-08-02
Update availabilities_contact.php

dolibit-ut:


dolibit-ut:
just see here:   $permission = $user->hasRight('bookcal', 'availabilities', 'write');

can we change  $permission  to  $permissiontowrite  or   $permissiontoadd  ??

eldy:
> $permission

Inside main .php file, yes $permission must be replaced with $permissiontoadd or $permissionnote (for page to show/edit notes).
inside .tpl file, we keep to use local var $permission that is set from $permissiontoadd.

PR #25550, opened 2023-08-02
Update booking_contact.php

dolibit-ut:


dolibit-ut:
here again i would change $permission  to  $permissiontoadd  ?

PR #25558, opened 2023-08-02
Update card.php

dolibit-ut:
edit require_once


dolibit-ut:
why this line on top:   `use Stripe\BankAccount;  `   ?
can we move / remove ?

eldy:
> why this line on top: `use Stripe\BankAccount; ` ? can we move / remove ?

Yes, this is an addition done automatically by the developer IDE. No reason to have this here... I removed it manually.

PR #25561, opened 2023-08-02
Update ChangeLog

dolibit-ut:
sorting 
and moved some line from users to devs/integrators


dolibit-ut:
is line 22 the sames as line 15 ?

PR #25563, opened 2023-08-02
Quick fix to allow JSON in extrafields POST parameters

mc2contributor:
# FIX Cannot use JSON for extrafields parameters
We have a complex field that needs to be submitted as JSON.  None of the available filters in `setOptionalsFromPost()` allow for this.  This is a quick fix which does no filtering at all.  Additional validation should probably be performed.

### DIFF
--- a/htdocs/core/class/extrafields.class.php
+++ b/htdocs/core/class/extrafields.class.php
@@ -2184,6 +2184,8 @@ class ExtraFields
 						$label_security_check = empty($conf->global->MAIN_SECURITY_ALLOW_UNSECURED_LABELS_WITH_HTML) ? 'alphanohtml' : 'restricthtml';
 					}
 					$value_key = GETPOST("options_".$key, $label_security_check);
+				} elseif (in_array($key_type, array('json'))) {
+					$value_key = GETPOST("options_".$key, 'none');
 				} else {
 					$value_key = GETPOST("options_".$key);
 					if (in_array($key_type, array('link')) && $value_key == '-1') {


eldy:
If no filter are available into GETPOST, it probably means that you are introducing a security vulnerability by passing data that can't be sanitized.
A JSON is often a full object. And transmetting a full object as a parameter is not a good practice. You should transfer only the ID of object or each value that will be sanitized independently.
Can you describe your user case with example ?

mc2contributor:
We purchased the Advanced Professional Gantt module and are adding the ability to add multiple task dependencies for a single task.  Each linked task has a link type (finish-to-start, start-to-start, start-to-finish, or finish-to-finish) and a lag time.  Here is our widget for editing this list of linked tasks:

Screenshot 2023-08-10 at 1 12 41 PM

The simplest way of submitting this to the server is as an array of objects, such as:

```json
[{"id":63,"type":"1","lag":"0"},{"id":66,"type":"3","lag":"19"}]
```

This merge request could add basic JSON validation with the following code:

```php
$value_key = GETPOST("options_".$key, 'none');
$value_key = json_encode(json_decode($value_key));
```

Further validation could be added with a hook, and that is the approach I started with.  I was going to use a separate table (`llx_projet_task_links`) to store these links, but I realized it would be adequate and much simpler to use Dolibarr's existing extrafields system with a JSON value.

If you see a better way to implement all of this please let me know.

eldy:
Using an extrafields seems a good choice.
On your example, we see that the data you need to provide and store is just an array with list of key and id.
So you can use and extrafield of text (that is protected with format alphanohtml).
And send data this way: 
63,1,0:66,3,19
It will need more work to parse the string to recreate the array, but it remains easy and is secured. I recommand this instead of introducing a format with sanitizing that is doubtful on a security concern.

PR #25566, opened 2023-08-03
NEW Add filter to restrict the data returned by API

hansemschnokeloch:
# NEW Filter data returned by the API

Currently the dolibarr API is very verbose if you get a list of objects, and you usually don't need all the data.
This PR introduce a new filter on the index method, that will only return the requested properties.


hansemschnokeloch:
@eldy could you give me your opinion on this new feature in the API? If accepted this will of course be done for each API endpoint.

Example of a result, id is always returned

Swagger-UI

eldy:
Good idea. You can submit PR
However for the part than clean the $object, instead of recreating a stdObj why not doing a
```
foreach (get_object_vars($object) as $key => $value) {
   if (!in_array($key, $properties)) { unset($object->$key); }
}
```
This initializes a second new object in memory, cleaning existing one may be better ?

hansemschnokeloch:
I agree, no need to create a stdObj, unsetting unwanted properties is a better way to do it.

eldy:
I let you resubmit the PR with the suggested change...

hansemschnokeloch:
@eldy PR is ready for review.

- Properties filter is added on all relevant api endpoints.
- Filter function is added in htdocs/api/class/api.class.php file.

eldy:
Thanks for this very valuable PR

PR #25574, opened 2023-08-04
NEW : Date field for shipment export

altairis-melina:
Add validation date field to shipment export
Capture d’écran 2023-08-04 095823


BenjaminFlr:
Hi @eldy, we get weird errors from Travis. If I understand correctly, there's no link with the PR, but only with testing methods from the job. Am I right?

PR #25584, opened 2023-08-07
FIX : float on size field to prevent error on php 8.2

atm-jpb:
FIX : float on size field to prevent error on php 8.2

### DIFF
--- a/htdocs/core/class/extrafields.class.php
+++ b/htdocs/core/class/extrafields.class.php
@@ -959,7 +959,7 @@ class ExtraFields
 
 		$label = $this->attributes[$extrafieldsobjectkey]['label'][$key];
 		$type = $this->attributes[$extrafieldsobjectkey]['type'][$key];
-		$size = $this->attributes[$extrafieldsobjectkey]['size'][$key];
+		$size = (float) $this->attributes[$extrafieldsobjectkey]['size'][$key];
 		$default = $this->attributes[$extrafieldsobjectkey]['default'][$key];
 		$computed = $this->attributes[$extrafieldsobjectkey]['computed'][$key];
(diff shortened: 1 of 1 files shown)


htdocs/core/class/extrafields.class.php
```diff
@@ -959,7 +959,7 @@ public function showInputField($key, $value, $moreparam = '', $keysuffix = '', $
 
 		$label = $this->attributes[$extrafieldsobjectkey]['label'][$key];
 		$type = $this->attributes[$extrafieldsobjectkey]['type'][$key];
-		$size = $this->attributes[$extrafieldsobjectkey]['size'][$key];
+		$size = (float) $this->attributes[$extrafieldsobjectkey]['size'][$key];
```
eldy, line 962: Size is a a string because it can be composed with 2 values, a first one for total digit and a second one for total nb of digit of decimal part.

atm-greg: The $size is used in round on line 1002 and 1004 that creates a fatal error in PHP 8.x
The test is a compare with integer so the second part of the size is useless.

This create a fatal error on extrafields display just for a dumb test and so we can't create invoices !!!!

eldy: > The $size is used in round on line 1002 and 1004 that creates a fatal error in PHP 8.x
> The test is a compare with integer so the second part of the size is useless.
> 
> This create a fatal error on extrafields display just for a dumb test and so we can't create invoices !!!!

I see. So i suggest to force the float only on line 1002 dand 1004, so when we need to use the value as a simple numeric, so we can keep the original value in memory.

eldy:
Seems fixed by another PR, so i close this one

PR #25586, opened 2023-08-07
Fix minimal price in order

ptibogxiv:
bad variable name

### DIFF
--- a/htdocs/commande/card.php
+++ b/htdocs/commande/card.php
@@ -1007,11 +1007,11 @@ if (empty($reshook)) {
 			$desc = dol_htmlcleanlastbr($desc);
 
 			if ($usermustrespectpricemin) {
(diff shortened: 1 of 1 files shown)


htdocs/commande/card.php
```diff
@@ -1007,11 +1007,11 @@
 			$desc = dol_htmlcleanlastbr($desc);
 
 			if ($usermustrespectpricemin) {
-				if ($pu_equivalent && $price_min && ((price2num($pu_equivalent) * (1 - $remise_percent / 100)) < price2num($price_min))) {
```
eldy, line 1010: At line 692, we have this comment
// Prepare a price equivalent for minimum price check
So it seems it is done on purpose to test on pu_equivalent. if i compare to the page of proposal, it seems the definition of pu_equivalent was not done a the good place. So i moved it at the same place than for proposal.
Can you check it fixes for your case ?

PR #25588, opened 2023-08-08
NEW no need to create invoice supplier object on supplier card for standalone credit note

lvessiller-opendsi:
NEW no need to create invoice supplier object on supplier card for standalone credit note

- no need to use constructor method of supplier invoice "new FactureFournisseur()" if its a standalone credit note
- "CorrectedInvoiceNotFound" was not found in lang files and I suggest to replace with "NoInvoiceToCorrect" (translated for "fr_FR" : Pas de facture à corriger)

### DIFF
--- a/htdocs/fourn/facture/card.php
+++ b/htdocs/fourn/facture/card.php
@@ -3100,7 +3100,7 @@ if ($action == 'create') {
 				$facusing->fetch($object->fk_facture_source);
 				print ' <span class="opacitymediumbycolor paddingleft">'.$langs->transnoentities("CorrectInvoice", $facusing->getNomUrl(1)).'</span>';
 			} else {
-				print ' <span class="opacitymediumbycolor paddingleft">'.$langs->transnoentities("NoInvoiceToCorrect").'</span>';
+				print ' <span class="opacitymediumbycolor paddingleft">'.$langs->transnoentities("CorrectedInvoiceNotFound").'</span>';
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


htdocs/fourn/facture/card.php
```diff
@@ -3095,12 +3095,12 @@ function setRadioForTypeOfInvoice() {
...
 		}
 		if ($object->type == FactureFournisseur::TYPE_CREDIT_NOTE) {
-			$facusing = new FactureFournisseur($db);
 			if ($object->fk_facture_source > 0) {
+				$facusing = new FactureFournisseur($db);
 				$facusing->fetch($object->fk_facture_source);
 				print ' <span class="opacitymediumbycolor paddingleft">'.$langs->transnoentities("CorrectInvoice", $facusing->getNomUrl(1)).'</span>';
 			} else {
-				print ' <span class="opacitymediumbycolor paddingleft">'.$langs->transnoentities("CorrectedInvoiceNotFound").'</span>';
+				print ' <span class="opacitymediumbycolor paddingleft">'.$langs->transnoentities("NoInvoiceToCorrect").'</span>';
```
eldy, line 3103: NoInvoiceToCorrect seems better on creation, but on edition (after deletion of source invoice for a special reason), old label seems better.

lvessiller-opendsi: @eldy 
It seems to be better but the translated key "CorrectedInvoiceNotFound" is missing in lang files.
What kind of translated value can be used for "CorrectedInvoiceNotFound" in "en_US" lang file ?

PR #25592, opened 2023-08-08
Add new dictionary for modulebuilder 

lamrani002:
# NEW|New function for create dictionnary by create table & write in module file


htdocs/modulebuilder/template/core/modules/modMyModule.class.php
```diff
@@ -229,7 +229,7 @@ public function __construct($db)
 			'tabhelp'=>array(array('code'=>$langs->trans('CodeTooltipHelp'), 'field2' => 'field2tooltip'), array('code'=>$langs->trans('CodeTooltipHelp'), 'field2' => 'field2tooltip'), ...),
 		);
 		*/
-
+		/*Begin dictionaries*/
```
eldy, line 232: Can you replace the /*Begin dictionaries*/
with 
/* BEGIN MODULEBUILDER DICTIONARIES */
and
/* END MODULEBUILDER DICTIONARIES */

So it follow same rules than other sections...

lamrani002: yes of course

PR #25593, opened 2023-08-08
Management Societe Address in propale

leninrivas:
# NEW|New *Address delivery  independent of contact people*
[*Delivery address independent of contact people*]

### DIFF
--- a/htdocs/comm/propal/card.php
+++ b/htdocs/comm/propal/card.php
@@ -83,6 +83,7 @@ $backtopage = GETPOST('backtopage', 'alpha'); // Go back to a dedicated page
 $lineid = GETPOST('lineid', 'int');
 $contactid = GETPOST('contactid', 'int');
 $address_id = GETPOST('address_id', 'int');
+$delivery_contactid = GETPOST('delivery_contactid', 'int');
 $projectid = GETPOST('projectid', 'int');
 $rank = (GETPOST('rank', 'int') > 0) ? GETPOST('rank', 'int') : -1;
 
@@ -420,7 +421,7 @@ if (empty($reshook)) {
 					$object->delivery_date = $date_delivery;
 					$object->availability_id = GETPOST('availability_id');
 					$object->demand_reason_id = GETPOST('demand_reason_id');
-					$object->fk_delivery_address = GETPOST('fk_address', 'int');
+					$object->fk_delivery_address = GETPOST('address_id', 'int');
 					$object->shipping_method_id = GETPOST('shipping_method_id', 'int');
 					$object->warehouse_id = GETPOST('warehouse_id', 'int');
 					$object->duree_validite = $duration;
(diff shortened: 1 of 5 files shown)


eldy:
The table socpeople is dedicated to stores alternative addresses. So for example a  delivery address. 

What is the goal of having a second solution to store delivery address of a proposal ?

leninrivas:
> The table socpeople is dedicated to stores alternative addresses. So for example a delivery address.
> 
> What is the goal of having a second solution to store delivery address of a proposal ?

Hello.
When a contact requests that a sale of products be made to an address other than the one they belong to. This address must be shown on the quote. It will also show your name and surname of the person requesting the budget

eldy:
> > The table socpeople is dedicated to stores alternative addresses. So for example a delivery address.
> > 
> > What is the goal of having a second solution to store delivery address of a proposal ?
> 
> Hello.
> When a contact requests that a sale of products be made to an address other than the one they belong to. This address must be shown on the quote. It will also show your name and surname of the person requesting the budget

I see. But this can already be achieved with the existing "socpeople" table. If a thirdparty wants a delivery at a different address, you can set on the object the delivery address on the tab Contact/Adress. If this is a different billing address, you can also force the villing address on the tab Contact/Adress

So we already have a place to store thisbin database (llx_socpeople) and a place in screen to set this address (in tab Contact/Address).

leninrivas:
> I see. But this can already be achieved with the existing "socpeople" table. If a thirdparty wants a delivery at a different address, you can set on the object the delivery address on the tab Contact/Adress. If this is a different billing address, you can also force the villing address on the tab Contact/Adress
> 
> So we already have a place to store thisbin database (llx_socpeople) and a place in screen to set this address (in tab Contact/Address).

In that case it's perfect. But there is another case. The contact you request has a different address. For example, the company Crash Bandicoot S.A. It has 4 locations. Lenin Rivas is a logistics supervisor and his office is at location 2. He requests that the products be sent to location 3, so the quote must show who requests and where it is delivered.

eldy:
> > I see. But this can already be achieved with the existing "socpeople" table. If a thirdparty wants a delivery at a different address, you can set on the object the delivery address on the tab Contact/Adress. If this is a different billing address, you can also force the villing address on the tab Contact/Adress
> > 
> > So we already have a place to store thisbin database (llx_socpeople) and a place in screen to set this address (in tab Contact/Address).
> 
> In that case it's perfect. But there is another case. The contact you request has a different address. For example, the company Crash Bandicoot S.A. It has 4 locations. Lenin Rivas is a logistics supervisor and his office is at location 2. He requests that the products be sent to location 3, so the quote must show who requests and where it is delivered.

So why don't you create all alternative addresses into the contact/addresses. You can create Lenin Rivas supervisor, Location 2, location 3. Then on invoice, you set all this address with the correct type, billing, delivery. The type of contact/adress is designed for this.
Then on pdf, you can have the main billing address but also the delivery address.

leninrivas:
> > > I see. But this can already be achieved with the existing "socpeople" table. If a thirdparty wants a delivery at a different address, you can set on the object the delivery address on the tab Contact/Adress. If this is a different billing address, you can also force the villing address on the tab Contact/Adress
> > > So we already have a place to store thisbin database (llx_socpeople) and a place in screen to set this address (in tab Contact/Address).
> > 
> > 
> > In that case it's perfect. But there is another case. The contact you request has a different address. For example, the company Crash Bandicoot S.A. It has 4 locations. Lenin Rivas is a logistics supervisor and his office is at location 2. He requests that the products be sent to location 3, so the quote must show who requests and where it is delivered.
> 
> So why don't you create all alternative addresses into the contact/addresses. You can create Lenin Rivas supervisor, Location 2, location 3. Then on invoice, you set all this address with the correct type, billing, delivery. The type of contact/adress is designed for this. Then on pdf, you can have the main billing address but also the delivery address.

I see. I doit but on pdf propal no shows delivery adresss

eldy:
You may check that option INVOICE_SHOW_SHIPPING_ADDRESS is on.
Shipping address should be a native feature now.

PR #25612, opened 2023-08-10
$object->entity not on product

defrance:
use $conf->entity instead


hregis:
@defrance heuuu, what is the problem ?

defrance:
oups, they append only when product are not initialised
i close it

PR #25614, opened 2023-08-11
FIX: warning when Workboard Responses display non numeric strings

atm-florianm:
# FIX home page warning when workboard responses have non numeric data
Some external modules use workboards to show data that is not exclusively numeric.

Here is a (fictitious) example:
image

This is a convenient way to display more than just an integer, but it causes warnings when Dolibarr sums overdue / pending elements.

We would like to keep this simple way of showing slightly more detailed information without impacting the overdue/pending calculation.

### DIFF
--- a/htdocs/index.php
+++ b/htdocs/index.php
@@ -437,7 +437,7 @@ if (empty($conf->global->MAIN_DISABLE_GLOBAL_WORKBOARD)) {
 
 	// We calculate $totallate. Must be defined before start of next loop because it is show in first fetch on next loop
 	foreach ($valid_dashboardlines as $board) {
-		if ($board->nbtodolate > 0) {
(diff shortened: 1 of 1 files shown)


htdocs/index.php
```diff
@@ -437,7 +437,7 @@
 
 	// We calculate $totallate. Must be defined before start of next loop because it is show in first fetch on next loop
 	foreach ($valid_dashboardlines as $board) {
-		if ($board->nbtodolate > 0) {
+		if ($board->nbtodolate > 0 && is_numeric($board->nbtodo) && is_numeric($board->nbtodolate)) {
```
eldy, line 440: So you should move the test $board->nbtodolate > 0 after the test is_numeric($board->nbtodolate) to avoid to make the comparison if value is not numeric

atm-florianm, line 440:
```suggestion
		if (is_numeric($board->nbtodo) && is_numeric($board->nbtodolate) && $board->nbtodolate > 0) {
```

Issue #24948, opened 2023-06-02
Dutch version of Belgian Journals


mschamp:
### Feature Request

In the accountancy module, only the French version of the journals are included: [url]/dolibarr/accountancy/admin/account.php
It would be a good option to include the dutch version also.
The translation is available here: https://www.cnc-cbn.be/fr/node/2250/multilingual-comparison/nl-fr-en 

### Use case

Allow Dutch speaking companies to use the accountacy module

### Suggested implementation

Add journal in htdocs\install\mysql\data\llx_accounting_abc.sql
Add translations of all lines to htdocs\install\mysql\data\llx_accounting_account_be.sql

First lines should look like:
INSERT INTO llx_accounting_account (entity, rowid, fk_pcg_version, pcg_type, account_number, account_parent, label, active) VALUES (__ENTITY__, ????1, 'MAR-MIN', 'KAPITAAL', '10',  '????', 'Kapitaal', 1);
INSERT INTO llx_accounting_account (entity, rowid, fk_pcg_version, pcg_type, account_number, account_parent, label, active) VALUES (__ENTITY__, ????, 'MAR-MIN', 'KAPITAAL', '100',  '????1', 'Geplaatst kapitaal', 1);

Is it correct that the ???? are 'random' numbers  and could be the numbers from the French entries, + 200000

### Suggested steps

- Add journal in htdocs\install\mysql\data\llx_accounting_abc.sql
- Add translations of all lines to htdocs\install\mysql\data\llx_accounting_account_be.sql

PR #25624, opened 2023-08-13
Be verkort

mschamp:
# Fix #[#24948 Dutch version of Belgian journals]
I tried to add the Dutch version of the journals.
Would it be possible to check, and if I made a mistake, gives some information how to fix it.
To be clear, I am not an accountant, only used information available on the net, maybe someone with a better knowlegde should verify it.


htdocs/install/mysql/data/llx_accounting_account_be.sql
```diff
@@ -945,3 +945,432 @@ INSERT INTO llx_accounting_account (entity, rowid, fk_pcg_version, pcg_type, acc
 INSERT INTO llx_accounting_account (entity, rowid, fk_pcg_version, pcg_type, account_number, account_parent, label, active) VALUES (__ENTITY__, 1355, 'PCMN-BASE', 'FINAN', '5', '0', 'Placement de trésorerie et de valeurs disponibles', 1);
 INSERT INTO llx_accounting_account (entity, rowid, fk_pcg_version, pcg_type, account_number, account_parent, label, active) VALUES (__ENTITY__, 1356, 'PCMN-BASE', 'EXPENSE', '6', '0', 'Charges', 1);
 INSERT INTO llx_accounting_account (entity, rowid, fk_pcg_version, pcg_type, account_number, account_parent, label, active) VALUES (__ENTITY__, 1357, 'PCMN-BASE', 'INCOME', '7', '0', 'Produits', 1);
+
+INSERT INTO llx_accounting_account (entity, rowid, fk_pcg_version, pcg_type, account_number, account_parent, label, active) VALUES (__ENTITY__, 200439, 'MAR-VERKORT', 'CAPIT', '1',  '0', 'Eigen vermogen, voorzieningen voor risico''s en kosten en schulden op meer dan één jaar', 1);
```
eldy, line 949: Looks ok. Just one thing to fix.
The ID to use (example here 200439) must be an ID lower than 99999.  The 200 000 is added automatically for belgium to avoid conflict with another country.
The previous plan end at 1357, so i suggest to start the new one for dutch for the new plan at 10000
So you can replace the string  
", 200"  into  ", 10"  
and  ",  '200"  into ",  '10"
So ID will be 10439, 10440, ...

PR #25631, opened 2023-08-15
Fix Computed Extrafields on list

ibuiv:
# FIX|Fix # Computed Extrafields on list
The computed extrafields were not calculated on list views. 
The formula asks to write $objectoffield to reference the actual object. Otherwise the template called by list.php has the object referenced in $object. 
I simply added a reference to $object name $objectoffield 
I removed debug commentaries also.


nat-itopie:
Works for bills but doesn't fix empty `$objectoffield` in list for thirdparties extrfields.

PR #25632, opened 2023-08-16
doc

frederic34:
@altairis-tof ->statut is deprecated, can you confirm that it is still working


altairis-tof:
Hello @frederic34 i have made this fix on v16 branch where status was still in the statut property ;)
i think it is a very big bug and i am surprised that it was merged only in develop branch...

PR #25635, opened 2023-08-16
NEW We added a hook to allow us to modify the Prospect Customer drop down…

thomasjulie:
We added a hook to allow us to modify the Prospect Customer drop down list in Third Party.


htdocs/societe/card.php
```diff
@@ -120,7 +120,7 @@
 $socialnetworks = getArrayOfSocialNetworks();
 
 // Initialize technical object to manage hooks of page. Note that conf->hooks_modules contains array of hook context
-$hookmanager->initHooks(array('thirdpartycard', 'globalcard'));
+$hookmanager->initHooks(array('thirdpartycard', 'globalcard', 'customizeprospectcustomerfield'));
```
eldy, line 123: Don't need another context here. The context is the "main page" where the hook will be called.
You already have thirdpartycard

htdocs/societe/card.php
```diff
@@ -1500,7 +1500,10 @@ function formatCustomerSelection (selection) {
 		print '<tr><td class="titlefieldcreate">'.$form->editfieldkey('ProspectCustomer', 'customerprospect', '', $object, 0, 'string', '', 1).'</td>';
 		print '<td class="maxwidthonsmartphone">';
 		$selected = (GETPOSTISSET('client') ?GETPOST('client', 'int') : $object->client);
-		print $formcompany->selectProspectCustomerType($selected);
+		$reshook = $hookmanager->executeHooks('prospectcustomerfield');
```
eldy, line 1503: Method in execute hook must be a method in camel expression.  Replace  
prospectcustomerfield 
into
selectProspectCustomerType

PR #25636, opened 2023-08-16
FIX: Quick search Intervention redirect to wrong page

FHenry:
FIX: Quick search Intervention redirect to wrong page when Ref exists


FHenry:
@eldy Could have a look please ?

PR #25648, opened 2023-08-19
FIX TakePOS receipt preview in admin

Max13:
Receipt preview in admin shows header and footer repeating themselves because in preview, `$constFreeText` equals `TAKEPOS_HEADER` (as there is no `$_SESSION['takeposterminal']` defined).

Capture d’écran 2023-08-19 à 12 18 33
Capture d’écran 2023-08-19 à 12 18 41

### DIFF
--- a/htdocs/takepos/receipt.php
+++ b/htdocs/takepos/receipt.php
@@ -111,7 +111,7 @@ if (!empty($hookmanager->resPrint)) {
 <br>
 <p class="left">
 <?php
-$constFreeText = 'TAKEPOS_HEADER'.$_SESSION['takeposterminal'];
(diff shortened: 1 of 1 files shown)


Max13:
@eldy Should I also PR the same fix for 16.0, 17.0 and develop ?

eldy:
When a fix is pushed for version x, it is also available for version x+1, x+2, ... and develop. Merges are done regularly to have fixes propagated.
So you can push it on v16 bit no need to do it for v17 .

Max13:
Ok, didn't know. I thought I had to take into account X - 1 and X - 2. You did merge 18.0, should I also PR for 16.0 to cover 17.0 ?

PR #25649, opened 2023-08-19
FIX TakePOS redundant front class

Max13:
Removed `centerinmiddle` class from categories and products buttons, as it's redundant with flexbox and moved the chevrons/arrows to the bottom of the buttons.

Capture d’écran 2023-08-19 à 17 56 09

### DIFF
--- a/htdocs/takepos/index.php
+++ b/htdocs/takepos/index.php
@@ -1404,10 +1404,10 @@ if (!empty($conf->global->TAKEPOS_WEIGHING_SCALE)) {
 				<?php
 				if ($count == ($MAXCATEG - 2)) {
 					//echo '<img class="imgwrapper" src="img/arrow-prev-top.png" height="100%" id="catimg'.$count.'" />';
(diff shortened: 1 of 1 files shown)


eldy:
On the screenshot, it seems the arrow is no more centered.

Max13:
Exactly. That's what I fix. Removing the class centers them back

eldy:
> Exactly. That's what I fix. Removing the class centers them back

And why is it a fix ? Seems strange after. What's wrong with a centered image ?

Max13:
There is nothing wrong, when I installed v18 from scratch, the chevrons appeared aligned at the bottom of the container and what I noticed is that 2 classes were redundant and resulted in the chevrons not centered vertically and horizontally but aligned at the bottom.

It's a fix because it puts back the chevrons at the center of the container, where they're supposed to be (where they was in v17)

eldy:
I think trouble was in v17.
V18 tried to provide nicer css rules, that is why you got a change. But i think the expected behaviour is to have arrows centered both vertically and horizontally.

PR #25650, opened 2023-08-19
FIX TakePOS redundant front class

Max13:
Removed `centerinmiddle` class from categories and products buttons, as it's redundant with flexbox and moved the chevrons/arrows to the bottom of the buttons.

See: #25649


eldy:
No need to push a pr on develop as a pr is already pushed in v18
All commit in v18 are already merged in develop automatically.

PR #25652, opened 2023-08-21
Replacement of 'llx_' by MAIN_DB_PREFIX in DolibarrModules.class.php

zahwyrix:
# NEW
Implementation of the user-defined constant as a prefix for database tables, rather than using "llx_" as a brute in the code.

### DIFF
--- a/htdocs/core/modules/DolibarrModules.class.php
+++ b/htdocs/core/modules/DolibarrModules.class.php
@@ -1156,7 +1156,7 @@ class DolibarrModules // Can not be abstract, because we need to instantiate it 
 					if (is_resource($handle)) {
 						$dirfound++;
 
-						// Run llx_mytable.sql files, then llx_mytable_*.sql
+						// Run MAIN_DB_PREFIX_mytable.sql files, then MAIN_DB_PREFIX_mytable_*.sql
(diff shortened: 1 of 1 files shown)


hregis:
@zephyriony this is not good !! this the files with prefix "llx_" not the tables

zahwyrix:
> @zephyriony this is not good !! this the files with prefix "llx_" not the tables

@hregis  Yes and precisely the name of the file is supposed to have the name of the table for a minimum of coherence. I end up with "dolibarr_" tables and "llx_" files, it's illogical.

hregis:
@zephyriony and if you would a prefix "ovni_" in database you rename all the file sql with "ovni_" ??? :-)

zahwyrix:
@hregis  Oh no, but I'm an idiot, I forgot that it was in the module section and that it therefore varied for each user, so much the better for me :)
In that case, why take the prefix into account? Why not add all sql files directly to the directory?
I'm a bit worried about the llx, though, as I've changed my prefix. Even if others who use my module will have llx.

hregis:
@zephyriony "llx_" for the name of the sql file is historical, the tables are declared with llx but renamed during installation if the prefix is not llx

PR #25653, opened 2023-08-21
add private note on Facture::TYPE_REPLACEMENT and Facture::TYPE_CREDI…

atm-jpb:
FIX :  add private note on Facture::TYPE_REPLACEMENT and Facture::TYPE_CREDI… 

This is a question in the form of a pr.
When creating a credit or replacement invoice, the public note is taken over but not the private note.
since these two fields come from two input textareas of the
 form create and do not come from a copy of the original invoice or the replacement invoice my question is :
 is there a business rule hidden behind this ? or my pr is valid ?


eldy:
Good question. I must analyze situation to see what is the expected generic behaviour...

eldy:
It seems the private note copy has been added in recent versions like we did for the public note.
So this question is now solved: We can do the copy of private note the same way we do for the public note. This is now the default behaviour

PR #25656, opened 2023-08-22
FIX Add substitution variables from project into propal

hansemschnokeloch:
# FIX Add missing project substitution variables in propal

When a propal is linked to a project, it is now possible to use the following substitution variables in the propal

- __PROJECT_REF__
- __PROJECT_NAME__
- __PROJECT_ID__

Issue is mentionned on the forum https://www.dolibarr.fr/forum/t/project-ref-non-reconnu-dans-propal/43978


htdocs/core/lib/functions.lib.php
```diff
@@ -8064,7 +8064,13 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 			if (is_object($object) && $object->element == 'project') {
 				$substitutionarray['__PROJECT_NAME__'] = $object->title;
 			}
-
+			if (is_object($object) && $object->element == 'propal' && !empty($object->fk_project)) {
```
eldy, line 8067: We should make fetch and database acces into this function.
If the property about proejct should be filled if property $object->projet is correctly filled.
The problem seems to be in the code that is calling getCommonSubstitutionArray()
The method   $object->fetch_project() is probably not called before calling the getCommonSubstitutionArray()

hansemschnokeloch: Thank you for the advice.
Changes on functions.lib are cancelled, and I added fetch_project() in propal.class.
This solves the issue in a better way

htdocs/comm/propal/class/propal.class.php
```diff
@@ -1693,6 +1693,9 @@ public function fetch($rowid, $ref = '', $ref_ext = '', $forceentity = 0)
 				$this->multicurrency_total_ht = $obj->multicurrency_total_ht;
 				$this->multicurrency_total_tva = $obj->multicurrency_total_tva;
 				$this->multicurrency_total_ttc = $obj->multicurrency_total_ttc;
+				
+				// Project
+				$this->fetch_project();
```
eldy, line 1698: I think it is still at wrong place: 
For standardization purpose, the fetch load only the object, the exteafields, and in some cases the lines of detail.
So can you move the fetch_project() into the caller of fetch() , just after the fetch() ?

hansemschnokeloch: Hi, fetch_project moved just after the fetch as you requested.

PR #25666, opened 2023-08-23
Adding foreninger as DK business entity types

JonBendtsen:
# NEW[*Add Forening as DK business entity type*]
The danish non profit organisation that I explore Dolibarr 18.0.0 for is a business entity type called *Forening* but this kind of business entity type was not listed, so I had to add it.

Source https://erhvervsstyrelsen.dk/regler-erhverv/virksomhedstyper/foreninger/generelt-om-foreninger


JonBendtsen:
@eldy When adding a thirdparty I noticed that I can not select any of these business entity types. Where would I go and add that change?

eldy:
> @eldy When adding a thirdparty I noticed that I can not select any of these business entity types. Where would I go and add that change?

Did the entry are visible in the dictionnary ?

JonBendtsen:
> 

I'm sorry @eldy  I do not understand what you mean, please elaborate

PR #25667, opened 2023-08-24
add email variable for shipping method

ultrasites:
# NEW|New Add email template variable __SHIPPINGMETHOD__
It adds a new template variable __SHIPPINGMETHOD__ for delviery emails to show the shipping method e.g. GPS ...


ultrasites:
@eldy

Issue #25678, opened 2023-08-25
add option in admin/mails.php - MAIN_EMAIL_SUPPORT_ACK


lmag:
### Feature Request

Add option Oui/Non
I did not find any informations about depriciation of this option by mailer ?

### Suggested implementation

![image](https://github.com/Dolibarr/dolibarr/assets/3288514/20310bee-6c98-41cf-af17-af8a42e247c7)

eldy:
See the analysis reported in PR #25679

PR #25679, opened 2023-08-25
feat: add option in email settings to retrieve acknowledgment feature

FHenry:
https://github.com/Dolibarr/dolibarr/issues/25678


htdocs/langs/fr_FR/admin.lang
```diff
@@ -302,6 +302,7 @@ MAIN_DISABLE_ALL_MAILS=Désactiver globalement tout envoi d'emails (pour mode te
 MAIN_MAIL_FORCE_SENDTO=Envoyer tous les emails à (au lieu des vrais destinataires, à des fins de test)
 MAIN_MAIL_ENABLED_USER_DEST_SELECT=Proposer les courriels des employés (si définis) dans la liste des destinataires prédéfinis lors de la rédaction d'un nouveau courriel
 MAIN_MAIL_NO_WITH_TO_SELECTED=Ne pas sélectionner de destinataire par défaut même si choix unique
+MAIN_EMAIL_SUPPORT_ACK=Ajouter la case à coché "Accusé de réception"
```
frederic34, line 305: @FHenry la case à cocher

eldy:
Adding a "visible" feature for "Acknowledgement of Receipt" in dolibarr let think that this feature can be used and is working.
But today this is no more true and AK is rarely sent even when requested (for both security and privacy reasons).

Example: 
- on apple Mail,  "The Mail application doesn't support read receipts by default" (https://discussions.apple.com/thread/4438975)
- gmail does not provide the feature anymore. The option "request an AK" is still supported only if you are using Google Workspace (formerly G Suite) as part of an organization and if it was enabled by an administrator (and it is not by default). it has been removed on personal gmail accounts (according to this source https://www.keeping.com/content/gmail-read-receipt/)
- always on gmail, when condition are met to accept to send AK (very rare, see previous point), the user must also validate the AK sending. More and more users just say no and disable it on this first popup.
- thunderbird does not send anymore the AK automatically by default. Instead a popup ask the user to confirm if the user is allowed to know when its email has been read. Like on gmail reader more and more users just say no and disable it on this first popup.
- email read from smartphone using an app implementing email access using POP or IMAP protocol does not send AK as they are just email reader and are not able to send.
- email read from integrated application (ERP, CMS, Chat services) using an email collector with POP or IMAP protocol does not send AK as they are just email collector and are not in charge to send email.
- AK system is also broken when target email is an alias and a transferred email like for a ML (due to email protocol, remember that AK system is not part of email protocol but a very old non official hack introduced by Microsoft Echange a very long time ago).
- emails in major companies does not support AK as mail software were designed to no do it for security purpose (fight against DOS done by AK bouncing email) and privacy purpose (a employee must not know when user read its email at home or on holiday). On large companies, when allowed, AK is allowed only among email in same domain (this restriction is now the default for most professional email software solution)
- there is a lot of other cases...

I don't know any mailer that answer accept AK now automatically by default.
So I made an experience among 500 emails from all the word and various mail provider (not a spam, but an important service email of my company) and i got only 3% of answer of the AK when i know that 99% has read it.

For this reason, i don't think we can let the user thinking Dolibarr provide a feature officially when this feature just does not work 97% of the time.

PR #25682, opened 2023-08-25
phpcs for only modified files

frederic34:


frederic34:
@eldy result after ~1min

image

image

PR #25683, opened 2023-08-25
NEW add simple order list as email variable

ultrasites:
# NEW|New simple order list as email variable
It renders a simple (position, description, amount, unit, price, total and discount) order list in an email.


htdocs/core/lib/functions.lib.php
```diff
@@ -10197,6 +10198,74 @@ function natural_search($fields, $value, $mode = 0, $nofirstand = 0)
...
+ * @return	string						template
+ */
+function showSimpleOrderTable($outputlangs, $object)
+{
+
+	global $conf;
+
+	$discountIsAvailable = false;
+	
+	foreach($object->lines as $order_position) {
```
eldy, line 10215: What if object has no ->lines property ?

ultrasites: @eldy i included a condition if lines property is empty, but why the test is failing? https://app.travis-ci.com/github/Dolibarr/dolibarr/builds/265505646 - can you explain it? Where is Task::$price? :)

ultrasites: @eldy any ideas?

eldy: > @eldy i included a condition if lines property is empty, but why the test is failing? https://app.travis-ci.com/github/Dolibarr/dolibarr/builds/265505646 - can you explain it? Where is Task::$price? :)

This seems introduced by the new function line 10245.
When object is not an Order but Task. The field ->line->price does not exists ...

ultrasites: It should be possible in both variants in my thoughts. I dont't use ODT-Files - it is explicit a digital process. Nothing to print and no attachments. It would be a very nice feature to list the order in an email. I have seen that as order confirmation on many online shops.

ultrasites: @eldy is it ok for you?

ultrasites: @eldy reminder :)

htdocs/core/lib/functions.lib.php
```diff
@@ -8261,6 +8261,7 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 			$already_payed_all = $object->sumpayed + $object->sumdeposit + $object->sumcreditnote;
 		}
 
+		$substitutionarray['__SIMPLE_ORDER_TABLE__'] = is_object($object) && !empty($object->lines) ? showSimpleOrderTable($outputlangs, $object) : "";
```
eldy, line 8264: It seems the call to showSimpleOrderTable can work only for some specific $object.
To avoid errors with object without the properties used by showSimpleOrderTable, can you add a test 
if (in_array($object->element, array('command', 'facture'...))) {
with list of relevent element code ...

htdocs/core/lib/functions.lib.php
```diff
@@ -10197,6 +10198,92 @@ function natural_search($fields, $value, $mode = 0, $nofirstand = 0)
...
+ */
+function showSimpleOrderTable($outputlangs, $object)
+{
+
+	global $conf;
+
+	$discountIsAvailable = false;
+	$orderPositionHasNoPrice = false;
+
+	if(!property_exists($object->lines[0], "remise_percent") ||
```
ultrasites, line 10216: @eldy here is my solution to ensure the right and "specific" object.

ultrasites:
@eldy are you resume the implementation?

eldy:
No plan to conitnue the implementation myself. But it could be enhnaced by renaming function to use it for all typeq of objects.

PR #25685, opened 2023-08-25
fix some warnings in dolibarr log

Hystepik:
fix warnings in dolibarr log

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -1392,7 +1392,7 @@ class Form
 	 * @param int 		$showcode 		Show code in list
 	 * @return array|string            	HTML string with
 	 */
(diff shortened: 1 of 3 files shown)


htdocs/core/class/html.form.class.php
```diff
@@ -1392,7 +1392,7 @@ public function select_company($selected = '', $htmlname = 'socid', $filter = ''
 	 * @param int 		$showcode 		Show code in list
 	 * @return array|string            	HTML string with
 	 */
-	public function select_thirdparty_list($selected = '', $htmlname = 'socid', $filter = '', $showempty = '', $showtype = 0, $forcecombo = 0, $events = array(), $filterkey = '', $outputmode = 0, $limit = 0, $morecss = 'minwidth100', $moreparam = '', $multiple = false, $excludeids = array(), $showcode = 0)
+	public function select_thirdparty_list($selected = '', $htmlname = 'socid', $filter = '(1:=:1)', $showempty = '', $showtype = 0, $forcecombo = 0, $events = array(), $filterkey = '', $outputmode = 0, $limit = 0, $morecss = 'minwidth100', $moreparam = '', $multiple = false, $excludeids = array(), $showcode = 0)
```
eldy, line 1395: (1:=:1) will mean "no filter"
But param $filter = '' also means "no filter".  
And empty string  '' seems more natural to say no filter provided.
So i pushed a better fix

PR #25686, opened 2023-08-25
fix phpstan

frederic34:


eldy:
thanks @frederic34 .
I started a tool to report all phpstan errors (low level) on a web page. It is available here (updated once per day):
https://doxygen.dolibarr.org/apstats/
Nb of phpstan error was 1500 few days ago and 1258 now, so we are on a good way to the 0 soon...

PR #25707, opened 2023-08-28
Update workstation_list.php

dolibit-ut:


htdocs/workstation/workstation_list.php
```diff
@@ -34,6 +34,8 @@
 // Load translation files required by the page
 $langs->loadLangs(array('mrp', 'other'));
 
+// Get parameters
+$id          = GETPOST('id', 'int');
```
eldy, line 38: Param specific to page should be after generic parameters.

PR #25709, opened 2023-08-28
Update /datapolicy/admin/setup.php

dolibit-ut:


dolibit-ut:
Is it useful to add for members a separate global option  
DATAPOLICY_USE_SPECIFIC_DELAY_FOR_MEMBER
to define its own time period ?

dolibit-ut:
Make it sense to define here an option:  if (isModEnabled('fournisseur') for Thirdparty & Contact ?

eldy:
> Is it useful to add for members a separate global option  
> DATAPOLICY_USE_SPECIFIC_DELAY_FOR_MEMBER
> to define its own time period ?

Why not. It is done for contact, why not for members.

PR #25711, opened 2023-08-28
Update htdocs/adherents/list.php

dolibit-ut:
action+display parameters


htdocs/adherents/list.php
```diff
@@ -41,6 +41,10 @@
 
 
 // Get parameters
+$catid 		= GETPOST("catid", 'int');
```
eldy, line 44: Most pages follow the example of modulbuilder that set for action+navigation parameters

PR #25718, opened 2023-08-28
Update 18.0.0-19.0.0.sql

dolibit-ut:
-- DROP COLUMN
-- maybe there is an option to copy into new fields or delete only if empty

### DIFF
--- a/htdocs/install/mysql/migration/18.0.0-19.0.0.sql
+++ b/htdocs/install/mysql/migration/18.0.0-19.0.0.sql
@@ -66,6 +66,11 @@ ALTER TABLE llx_societe_rib ADD COLUMN extraparams varchar(255);
 
 ALTER TABLE llx_c_type_container ADD COLUMN position integer DEFAULT 0;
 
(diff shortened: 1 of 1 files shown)


htdocs/install/mysql/migration/18.0.0-19.0.0.sql
```diff
@@ -66,6 +66,11 @@ ALTER TABLE llx_societe_rib ADD COLUMN extraparams varchar(255);
 
 ALTER TABLE llx_c_type_container ADD COLUMN position integer DEFAULT 0;
 
+ALTER TABLE llx_prelevement_demande ADD INDEX idx_prelevement_demande_ext_payment_id (ext_payment_id);
+
+
+-- DROP COLUMN
+-- maybe there is an option to copy into new fields or delete only if empty
```
eldy, line 73: No need to copy, this has already be done in previous migration.

dolibit-ut: okay

dolibit-ut:
close

PR #25719, opened 2023-08-28
Update conferenceorbooth_list.php

dolibit-ut:


htdocs/eventorganization/conferenceorbooth_list.php
```diff
@@ -42,21 +42,23 @@
 $langs->loadLangs(array("eventorganization", "other", "projects", "companies"));
 
 // Get Parameters
-$action     = GETPOST('action', 'aZ09') ? GETPOST('action', 'aZ09') : 'view'; // The action 'create'/'add', 'edit'/'update', 'view', ...
```
eldy, line 45: According to myobject_list.php (the module builder template), we first set common page, navigation parameters and then the specific param for page.

PR #25728, opened 2023-08-30
FIX Return right content type

avolani:
Revert content type to html (no json here, ajax calls fails)


htdocs/projet/ajax/projects.php
```diff
@@ -62,7 +62,7 @@
 
 require_once DOL_DOCUMENT_ROOT.'/core/class/html.formprojet.class.php';
 
-top_httphead('application/json');
+top_httphead();
```
eldy, line 65: Last line of file is print dol_json_encode, so content type should be json.

avolani: You're right but the function is never executed on mode 'gettask' since $formproject->selectTasks already prints the list and then return...

avolani: So it's better to convert the gettask to json in this case?

eldy: I see. Well we have 2 diffrrent types of return so yes we must add a condition and return 2 different types of content type depending on condition on gettasks.
Can you update pr like this ?

avolani: Yes, i will update the PR

PR #25740, opened 2023-08-30
NEW : hide external user in combobox

atm-thibaultf:
# NEW|New hide external user in combobox
Adding configuration to hide external user in combobox and select


htdocs/core/class/html.formother.class.php
```diff
@@ -513,6 +513,12 @@ public function select_salesrepresentatives($selected, $htmlname, $user, $showst
 		if (!empty($user->socid)) {
 			$sql_usr .= " AND u.fk_soc = ".((int) $user->socid);
 		}
+		if (!empty($conf->global->USER_HIDE_EXTERNAL_IN_COMBOBOX)) {
+			$sql_usr .= " AND u.employee <> 0";
+		}
+		if (!empty($conf->global->USER_HIDE_INACTIVE_IN_COMBOBOX)) {
```
atm-thibaultf, line 519: adding this in the same time because it was missing

htdocs/core/class/html.formother.class.php
```diff
@@ -513,6 +513,12 @@ public function select_salesrepresentatives($selected, $htmlname, $user, $showst
...
+			$sql_usr .= " AND u.statut <> 0";
+		}
```
atm-thibaultf, line 521: adding this in the same time because it was missing

atm-maxime:
In Dolibarr, an external user is attached to a third-party (fk_soc > 0). So there will be a confusion between external and non employee users

htdocs/comm/mailing/class/html.formadvtargetemailing.class.php
```diff
@@ -171,6 +171,9 @@ public function multiselectselectSalesRepresentatives($htmlname, $selected_array
 		if (!empty($conf->global->USER_HIDE_INACTIVE_IN_COMBOBOX)) {
 			$sql_usr .= " AND u2.statut<>0 ";
 		}
+		if (!empty($conf->global->USER_HIDE_EXTERNAL_IN_COMBOBOX)) {
+			$sql_usr .= " AND u2.employee<>0 ";
```
eldy, line 175: May be you mean  USER_HIDE_NONEMPLOYEE_IN_COMBOBOX  ?  (external term is used for external users so fk_soc > 0).
If you really mean hide external user, condition should be AND u2.fk_soc IS NULL

atm-thibaultf: Indeed ! In the end, I have added the two options for non employee and external user if it's fine for you @eldy

atm-thibaultf:
@eldy

PR #25742, opened 2023-08-30
llx_product_extrafields should probably also use fk_product

JonBendtsen:
# NEW [*llx_product_extrafields now uses fk_product just like other llx_product tables*]

[*llx_product_extrafields used fk_object, where as all these tables uses fk_product. llx_product_extrafields should probably also use fk_product just like these other tables below

llx_product_association
llx_product_attribute_combination
llx_product_attribute_combination_price_level
llx_product_attribute_value
llx_product_batch
llx_product_customer_price
llx_product_customer_price_log
llx_product_fournisseur_price
llx_product_fournisseur_price_log
llx_product_lang
llx_product_lot
llx_product_price
llx_product_price_by_qty
llx_product_stock
llx_product_warehouse_properties

These 2 tables also contains fk_object, but I have no data in them, so I am unsure if they are really the same data: llx_product_fournisseur_price_extrafields, llx_product_lot_extrafields*]

### DIFF
--- a/htdocs/install/mysql/migration/repair.sql
+++ b/htdocs/install/mysql/migration/repair.sql
@@ -129,7 +129,7 @@ update llx_commande set fk_user_valid = null where fk_user_valid not in (select 
 
 delete from llx_societe_extrafields where fk_object not in (select rowid from llx_societe);
 delete from llx_adherent_extrafields where fk_object not in (select rowid from llx_adherent);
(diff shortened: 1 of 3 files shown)


atm-maxime:
All extrafields tables are related to an object and have the same structure. That's why in all these tables the foreign key to the related object is called fk_object instead of fk_product or else.

PR #25744, opened 2023-08-30
Update llx_10_c_regions.sql

dolibit-ut:
Departements


hregis:
@dolibit-ut departments

hregis:
@dolibit-ut "county"is better

PR #25745, opened 2023-08-30
fix : translation was not working for defaultlang of thirdparty

grandoc:


eldy:

grandoc:
First you must have a product with translation for the description
Second use a thirdparty with a default language, same has the product
create a proposal for that thirdparty with a line with the product, and even if the language of the thirdparty is preselected, you will have only the label translated and the description still in your language

eldy:
> First you must have a product with translation for the description Second use a thirdparty with a default language, same has the product create a proposal for that thirdparty with a line with the product, and even if the language of the thirdparty is preselected, you will have only the label translated and the description still in your language

The description is a free text that user can modify by adding its own complementary information.
If a user has enter custom data here in the line of proposal or invoice, we must not drop this information.
For this reason, we can't and we should not translate a custom value of description (we don't have the translation).
So the translation is done only if current description has not been modified. In such a case, we know we can replace in PDF with the translated value. This is the reason of the test on textwasnotmodified that we must keep.

You can set the option MAIN_MULTILANG_TRANSLATE_EVEN_IF_MODIFIED is you want to drop the custom value and force translation with a generic text value in language of customer.

Currently, behaviour is the expected behaviour.

PR #25746, opened 2023-08-30
formprojet to formproject

dolibit-ut:
1. renamed all   html.formprojet.class.php  to  html.formproject.class.php
2. renamed all   FormProjets                      to  FormProjects


dolibit-ut:
something went wrong with git.dev

PR #25750, opened 2023-08-31
NEW Add subtotal for all types of linked objects and sort them by rowid (…

samifilali:
# Fix
Add subtotal for all types of linked objects and sort them by rowid (creation date)

### DIFF
--- a/htdocs/asset/tpl/linkedobjectblock.tpl.php
+++ b/htdocs/asset/tpl/linkedobjectblock.tpl.php
@@ -66,7 +66,7 @@ foreach ($linkedObjectBlock as $key => $objectlink) {
 	echo '</td>';
 	echo "</tr>\n";
 }
-if (count($linkedObjectBlock)) {
+if (count($linkedObjectBlock) > 1 || !empty($conf->global->LINKED_OBJECTS_HAVE_ALWAYS_SUBTOTAL)) {
 	echo '<tr class="liste_total '.(empty($noMoreLinkedObjectBlockAfter) ? 'liste_sub_total' : '').'">';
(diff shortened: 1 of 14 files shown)


eldy:
What is goal of adding a total when there is only 1 line ?

samifilali:
> What is goal of adding a total when there is only 1 line ?

I think it's cleaner to separate the different types of related objects when many are alone, logically and especially visually.

eldy:
The most common case is when there is only 1 line so with no need of subtotal. So it means we duplicate all lines with a subtotal just for tha rare case there is 2 lines of same type.
This create a visual load.
If separation between types an subtotal when there is is not clear, we should solve this with css for this rare case ( for example by adding a bolder separator line), instead of overloading screen for most common case.

samifilali:
I understand but in this case, should we apply the css on the <tr> tags and add a collapse attribute to the table, apply the css on each td tag or something else ?

atm-thibaultf:
@eldy

eldy:
> I understand but in this case, should we apply the css on the <tr> tags and add a collapse attribute to the table, apply the css on each td tag or something else ?
> 
> 

Using js to allow collapse can also be an interesting feature.
But my understanding here is that you want to show a total line, even when there is only 1 line. Am i right ? If yes, as this looks strange, i suggest first to add a hidden const to implement this behaviour so we won't loos time in pro or cons of this new behaviour, both are better and worse..

eldy:
Can you push this PR on develop branch.

samifilali:
cf. https://github.com/Dolibarr/dolibarr/pull/26566

PR #25755, opened 2023-08-31
Payment term is not required

defrance:


htdocs/compta/facture/card.php
```diff
@@ -3692,7 +3693,7 @@ function setRadioForTypeOfIncoice() {
 	}
 
 	// Payment term
-	print '<tr><td class="nowrap fieldrequired">'.$langs->trans('PaymentConditionsShort').'</td><td colspan="2">';
+	print '<tr><td class="nowrap ">'.$langs->trans('PaymentConditionsShort').'</td><td colspan="2">';
```
eldy, line 3696: Looks strange.
Payment term is a mandatory information on a invoice (in legal point of view).

defrance: functionally I agree with you on the mandatory aspect of this field but it is not mandatory in terms of the table constraints, hence my proposal to remove the indicators

eldy:
As the field is mandatory in a user point of view, the look and feel of the field must show this status mandatory.
Result on screen is as expected by GUI practices, so i close the PR.

Issue #25762, opened 2023-09-01
Avoid multiple DB calls on htdocs\accountancy\journal\purchasesjournal.php


Daviid-P:
### Feature Request

This change on https://github.com/Dolibarr/dolibarr/commit/767ad3e366d1cc7fc5cce75070b37bc87417a440 saves a lot of DB queries by getting the info_bits on the main SQL query

But on https://github.com/Dolibarr/dolibarr/blob/767ad3e366d1cc7fc5cce75070b37bc87417a440/htdocs/accountancy/journal/purchasesjournal.php#L192

we could still save even more with

```php
$tabother = array();
$vatdata_cache = array();

[...]

$tax_id = $obj->tva_tx . ($obj->vat_src_code ? ' (' . $obj->vat_src_code . ')' : '');
if (array_key_exists($tax_id, $vatdata_cache)) {
	$vatdata = $vatdata_cache[$tax_id];
} else {
	$vatdata = getTaxesFromId($tax_id, $mysoc, $mysoc, 0);
        $vatdata_cache[$tax_id] = $vatdata;
}
```

So we get the vat data for one ID only once from DB and all subsequent come from the stored value.

Daviid-P:
Also on https://github.com/Dolibarr/dolibarr/blob/767ad3e366d1cc7fc5cce75070b37bc87417a440/htdocs/accountancy/journal/purchasesjournal.php#L257-L273

we could use 
```php
$sql = "
SELECT
    fk_facture_fourn,
    COUNT(fd.rowid) as nb
FROM
    llx_facture_fourn_det as fd
WHERE
    fd.product_type <= 2
    AND fd.fk_code_ventilation <= 0
    AND fd.total_ttc <> 0
GROUP BY fk_facture_fourn
";
$resql = $db->query($sql);

$num = $db->num_rows($result);
$i = 0;
while ($i < $num) {
	$obj = $db->fetch_object($resql);
	if ($obj->nb > 0) {
		$errorforinvoice[$obj->fk_facture_fourn] = 'somelinesarenotbound';
	}
}
```

This way `$errorforinvoice` is always loaded with ALL unbound lines, i believe this is better than a querie for each line and it shouldn't take that long even with lots of lines (tested with 55 thousand invoices and 381 thousand invoice lines it took 0.03 seconds of query time).

Daviid-P:
Done

PR #25763, opened 2023-09-01
Remove unnecessary queries

Daviid-P:
# NEW|New Remove unnecessary queries in `accountancy/journal/purchasesjournal.php`
To address https://github.com/Dolibarr/dolibarr/issues/25762, for every invoice there were 2 unnecessary queries, 1 to get vat data with `getTaxesFromId()` and 1 to detect lines with not binding lines.

The first one is fixed by creating an array to store already retrieved vat data.
The second one is fixed by getting all invoices with unbound lines at once (should be faster than multiple queries) and populating the `$errorforinvoice` array with thos eunbound lines. As far as I can see `$errorforinvoice` is always accessed like `$errorforinvoice[$key]` inside a `foreach $array as $key => $value` loop so even if we add to `$errorforinvoice` invoices that are not shown the errors won't be shown either.

### DIFF
--- a/htdocs/accountancy/journal/purchasesjournal.php
+++ b/htdocs/accountancy/journal/purchasesjournal.php
@@ -325,6 +325,12 @@ if ($result) {
 	dol_print_error($db);
 }
 
+// Check for too many invoices first.
+if (count($tabfac) > 10000) { // Global config in htdocs/admin/const.php???
+	$error++;
+	setEventMessages("TooManyInvoicesToProcessPleaseUseAMoreSelectiveFilter", null, 'errors');
+}
+
 $errorforinvoice = array();
 
 /*
@@ -357,6 +363,7 @@ WHERE
     fd.product_type <= 2
     AND fd.fk_code_ventilation <= 0
     AND fd.total_ttc <> 0
+	AND fk_facture_fourn IN (".$db->sanitize(join(",", array_keys($tabfac))).")
 GROUP BY fk_facture_fourn
 ";
 $resql = $db->query($sql);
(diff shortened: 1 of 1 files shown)


htdocs/accountancy/journal/purchasesjournal.php
```diff
@@ -336,6 +345,31 @@
 		dol_print_error($db);
 	}
 }
+*/
+// New way, single query, load all unbound lines
```
eldy, line 349: Good idea.
However something is missing. In previous call, we fill the array $errorforinvoice only for invoices into the array $tabfac.
With the new code, we save ton of SQL request, but if the number of invoice in database is very large, we may return a  very large number of data, even if user has a filter to work with only 10 invoices. Time saved may be transformed into not enough memory error.
So I recommend to add a test in the SQL
`AND fk_facture_fourn IN ('.$db->sanitize(join(',', array_keys($tabfac))).')';`

A good practice is to not have too much elements inside a IN of a SQL (mysql and postgresql has no limit because they creates a temporary table and make a join when nb is > 999).
But i suggest to introduce a limit, no existing for the moment
```
if (count($tabfac) > 10000) {
$error++;
setEventMessages("TooManyInvoicesToProcessPleaseUseAMoreSelectiveFilter", null, 'errors');
}

```

Daviid-P: Committed the changes.
Should `if (count($tabfac) > 10000) {` go as a config variable from `htdocs/admin/const.php`??

eldy: This is a security against bad use.
100% of users should be ableto work with this limit of 10 000, so i see no reason for the moment to allow user to setup this.
A good thing would be however to reproduce the same optimization for the page salesjournal.php

PR #25768, opened 2023-09-01
Fix attempt to access stdClass::$* on query result

mc2contributor:
# FIX attempt to access stdClass::$* on query result
Existing code resulted in a PHP 'undefined property' warning, and was also inefficient, pulling the whole list from the database for PHP to count.  The change:
- warns if the `$field_select` argument to `getAllItemsLinkedByObjectID()` is invalid
- gets the database to count the records in this case and return a single result

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -4256,6 +4256,10 @@ abstract class CommonObject
 			return -1;
 		}
 
+		if (!preg_match('/^[_a-zA-Z0-9]+$/', $field_select)) {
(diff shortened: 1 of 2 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -4256,6 +4256,10 @@ public static function getAllItemsLinkedByObjectID($fk_object_where, $field_sele
 			return -1;
 		}
 
+		if (!preg_match('/^[_a-zA-Z0-9]+$/', $field_select)) {
+			error_log("ERROR: Invalid value '$field_select' for parameter \$field_select in call to getAllItemsLinkedByObjectID(). Must be a single field name.");
```
eldy, line 4260: error_log must not be used.
You can use instead 
dol_syslog(message, LOG_ERR);

PR #25782, opened 2023-09-04
do not display revenustamp in tooltip if zero

frederic34:


htdocs/compta/facture/class/facture.class.php
```diff
@@ -1982,7 +1982,7 @@ public function getTooltipContentArray($params)
 			if (!empty($this->total_tva)) {
 				$datas['amountvat'] = '<br><b>'.$langs->trans('AmountVAT').':</b> '.price($this->total_tva, 0, $langs, 0, -1, -1, $conf->currency);
 			}
-			if (!empty($this->revenuestamp)) {
+			if (!empty($this->revenuestamp) && $this->revenuestamp != 0) {
```
eldy, line 1985: If this->revenuestamp = 0, then the !empty should be already false ?

PR #25807, opened 2023-09-06
FIX avoid wrong backtopage url

hregis:


htdocs/comm/propal/card.php
```diff
@@ -1889,7 +1889,7 @@
 
 		$thirdparty = $soc;
 		$discount_type = 0;
-		$backtopage = urlencode($_SERVER["PHP_SELF"].'?socid='.$thirdparty->id.'&action='.$action.'&origin='.GETPOST('origin').'&originid='.GETPOST('originid'));
+		$backtopage = $_SERVER["PHP_SELF"].'?socid='.$thirdparty->id.'&action='.$action.'&origin='.GETPOST('origin').'&originid='.GETPOST('originid');
```
eldy, line 1892: urlencode must be added at least around the param origin and originid because this 2 params are coming from user input.

hregis: @eldy it's ok, thanks

PR #25811, opened 2023-09-06
fix: Optimize event per user view :  INNER JOIN always better to use index than WHERE (or LEFT JOIN with WHERE)

FHenry:
fix: Optimize event per user view :  INNER JOIN always better to use index than WHERE (or LEFT JOIN with WHERE)


htdocs/comm/action/peruser.php
```diff
@@ -645,19 +647,8 @@
 if ($status == 'todo') {
 	$sql .= " AND (a.percent >= 0 AND a.percent < 100)";
 }
-// We must filter on assignement table
-if ($filtert > 0 || $usergroup > 0) {
-	$sql .= " AND (";
-	if ($filtert > 0) {
-		$sql .= "ar.fk_element = ".$filtert;
```
eldy, line 652: Filter on user seems to have been lost.

FHenry: No because in new version there is a line 554
if ($filtert > 0) {
		$sql .= " AND ar.fk_element = ".((int) $filtert);
}

on the INNER JOIN directly no need in the WHERE clause (better for use index)

PR #25818, opened 2023-09-06
Fix add bookcal extrafields

Hystepik:
add extra fields tables for bookcal availabilities and  calendars


htdocs/install/mysql/tables/llx_bookcal_availabilities_extrafields.key.sql
```diff
@@ -0,0 +1,19 @@
+-- Copyright (C) 2023 Alice Adminson <aadminson@example.com>
```
eldy, line 1: The name of the table must be 
llx_bookcal_availabilities_extrafields-bookcal.key.sql

PR #25832, opened 2023-09-07
fix sql query for external users

atm-greg:
# FIX Sql query returning an error
When the user is an external user and the box is display, it returns an error because the query structure

Type gestionnaire de base de données: mysqli
Requête dernier accès en base en erreur: SELECT product.ref, COUNT(product.ref) as nb, SUM(tl.total_ht) as total, AVG(tl.total_ht) as avg FROM llx_commande as c, llx_commandedet as tl, llx_product as product INNER JOIN llx_societe_commerciaux as sc ON c.fk_soc = sc.fk_soc AND sc.fk_user = 66 WHERE c.entity IN (2,1) AND c.rowid = tl.fk_commande AND tl.fk_product = product.rowid AND c.date_commande BETWEEN '2023-01-01 00:00:00' AND '2023-12-31 23:59:59' GROUP BY product.ref ORDER BY nb DESC
Code retour dernier accès en base en erreur: DB_ERROR_NOSUCHFIELD
Information sur le dernier accès en base en erreur: Unknown column 'c.fk_soc' in 'on clause'

### DIFF
--- a/htdocs/commande/class/commandestats.class.php
+++ b/htdocs/commande/class/commandestats.class.php
@@ -245,13 +245,14 @@ class CommandeStats extends Stats
 		global $user;
 
 		$sql = "SELECT product.ref, COUNT(product.ref) as nb, SUM(tl.".$this->field_line.") as total, AVG(tl.".$this->field_line.") as avg";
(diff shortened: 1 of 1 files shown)


eldy (commented):
Because what we return is a list of product, using left join is not required (and not recommended)
Can you try to replace left join with inner join

atm-greg:
Update done

PR #25834, opened 2023-09-07
FIX payment : language is not propagated to following pages

thomas-Ngr:
Fix : on Stripe payment page, the &lang parameter is not correctly handled.

Context : the owner of the DLB instance is a french company (DLB default language is FR). It has some german customers thet pay with Stripe. But the Stripe portal is in French.

On payment page, a parameter &lang is used on the first page load but not on following screens.

first screen, german is OK : 

Capture d’écran_2023-09-07_17-02-09

second screen, german is broken : 

Capture d’écran_2023-09-07_17-03-04

This PR fixes this behavior by passing lang as a parameter in forms.


thomas-Ngr:
Close :
- base branch was wrong (14.0 instead of 5.3_rc)
- fix has been included into 14.0 AND 5.3_rc. PR is now useless.

PR #25840, opened 2023-09-07
Move common task card code into static method

mc2contributor:
# FIX Redundant code
On the Project / Tasks tab there are two different views, each provided by a separate page.  The buttons to access those pages are the same for both pages and should be in a common location.  This will simplify changes to the views and buttons (we want to add another view).


eldy:

PR #25849, opened 2023-09-08
NEW add order supplier submit notif

FlorentPoinsaut:
# NEW Add order supplier submit notif


eldy:
Do not update language files.
Only english files into en_US directory must be updated. All other files are overwritten by transifex translation sync.

FlorentPoinsaut:
OK @eldy, it's fixed.

PR #25858, opened 2023-09-10
NEW Accountancy - Option to choose length of lettering code

aspangaro:
image


htdocs/accountancy/class/lettering.class.php
```diff
@@ -288,7 +288,15 @@ public function letteringThirdparty($socid)
 	public function updateLettering($ids = array(), $notrigger = false)
 	{
 		$error = 0;
-		$lettre = 'AAA';
+		if (getDolGlobalInt('ACCOUNTING_LETTERING_NBLETTERS')) {
```
eldy, line 291: Why not replacing all this block with just

// Generate a string with n char A where n is ACCOUNTING_LETTERING_NBLETTERS (So 'AA', 'AAA', ...)
$lettre = str_pad("", getDolGlobalInt('ACCOUNTING_LETTERING_NBLETTERS', 3), "A");

aspangaro: Hi @eldy !

Thanks for your return. 
I use str_pad as you recommended.

According to what I can see on all the different software, we're almost always on 3 letters. I discovered on Quadra that it was limited to 2 letters, hence the suggested modification. 4 or even 5 letters is not impossible, but has never been seen before.

PR #25859, opened 2023-09-10
NEW Accountancy - Added management of tax period closures #FPC21

aspangaro:
image

image

image

image

image

image


eldy:
What is difference between "Close fiscal period" and "Locked record in a period" ? 

Can you explain the sentence "Add closing bookkeeping records" in the confirm box.   Do you mean  "Close (the verb) all bookkeeping record not yet closed" ?  So we check the box to make an action on database ? And what ill be this action ? setting a flag ? which one ?

aspangaro:
A few explanations: the difference between "Recording blocked in a period" & "Closing a fiscal period".

A/ For the first term, a blocked record in a period is an operation that cannot be modified or deleted, but records can be added to the same tax period.

B / For the second term, a fiscal period closing is a period containing only blocked records, which can no longer receive records.

For the sentence "Add closing bookkeeping records" in the confirm box. This means step by step:
Step 1: All records are blocked (See A above)

Step 2: (See B above) If all records are blocked, you can close the year. A year-end closing consists in blocking the fiscal year; it is no longer possible to add a record in the closed fiscal year. The status of the fiscal period is changed to "Closed".
If the box "Add closing bookkeeping records" in the confirm box is checked, we add a record at the first day of the next fiscal period. This is an "a-new" entry (https://www.compta-facile.com/reprise-des-a-nouveaux-en-comptabilite/) which carries forward balance sheet information from one fiscal year to the next.

Step 3 (Optional): We extinguish the so-called "inventory" transactions by taking the same transactions from this journal and reversing them on the 1st day of the new fiscal period.

eldy:
> Add closing bookkeeping records

Ok, so my understanding was correct but because the term differs from existing one, it creates a confusion.
"Lock existing bookkeeping record in the period to close"

Because we "Lock a line" and we "Close a period". Closing a period means : "Existing record are locked from any change or deletion and we can't add more into the period".

The term "Add" was also confusion, because Add means we will "create" new record. Here we just want to make an action on existing one. So what do you think with this term for the option:

"Lock/Close bookkeeping record that currently exist in the period to close ?"

It is longer, but such a term will avoid the 2 confusions I got.

aspangaro:
Hi @eldy !

I've reworked the texts and added a new option for subsidiary accounts at the close of the fiscal period.

The new windows:

image

I hope it's clearer !

eldy:
What happen if I uncheck the box "Generate has new": I imagine that lines in bookkeeping are just locked. 
And if I check ?  the lines are locked but what else ? Can you describe with an example ?

aspangaro:
If you uncheck the "generate a-new" box, the fiscal year will be closed (impossible to write in afterwards) and all accounting entries for the fiscal year will be blocked. There will simply be no a-new entries in the next fiscal year.

If you tick the "generate a-new" box, the fiscal year will be closed (impossible to write in afterwards) and the accounting entries for the fiscal year will all be blocked. There will be an a-new entry in the next fiscal year.

This is the basis for transmitting accounting information from one fiscal year to another. A basic rule of accounting is the separation of fiscal years.

Some reading (in french) : 
- https://www.compta-facile.com/reprise-des-a-nouveaux-en-comptabilite/ 
- https://www.sage.com/fr-fr/blog/glossaire/cloture-comptable-definition-de-la-cloture-de-lexercice/

htdocs/accountancy/class/bookkeeping.class.php
```diff
@@ -760,7 +789,7 @@ public function fetch($id, $ref = null, $mode = '')
 		$sql .= ' WHERE 1 = 1';
 		$sql .= " AND entity = " . ((int) $conf->entity); // Do not use getEntity for accounting features
 		if (null !== $ref) {
-			$sql .= " AND t.rowid = ".((int) $ref);
+			$sql .= " AND t.ref = '".$this->db->escape($ref)."'";
```
eldy, line 792: Currently, we have no field ref (we don't really need it), so we use the id as unique ref key. So old test looks as the good version

aspangaro: Ok, old method restored

PR #25866, opened 2023-09-11
FIX user notification for ticket assigned on creation

BenjaminFlr:
When an user is assigned to a ticket directly during the creation, there is no email notification whereas there is one when the user is assigned AFTER the creation.

I propose this little change in order to fix it and notify an user even when he's assigned during the ticket creation.


htdocs/ticket/class/ticket.class.php
```diff
@@ -568,6 +567,13 @@ public function create($user, $notrigger = 0)
 				$this->id = $this->db->last_insert_id(MAIN_DB_PREFIX."ticket");
 			}
 
+			// Assign user if defined
```
eldy, line 570: The create method is a low level CRUD method, so we should not have other business event and trigger inside the create.
Call of the assignUser should be done after calling the create by the parent page orchestrating the create call.

altairis-benjamin: Thanks for the feedback. I didn't know what was the best way to do this.

I'll make the required changes.

BenjaminFlr:
I made the changes.
@eldy : it should now do the trick.

BenjaminFlr:
Any update on here?

BenjaminFlr:
@eldy all the pre-commit tests are now ok :)

BenjaminFlr:
Little up here @eldy :)

htdocs/core/triggers/interface_50_modTicket_TicketEmail.class.php
```diff
@@ -190,6 +170,35 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
...
+							if (!empty($conf->global->TICKET_DISABLE_MAIL_AUTOCOPY_TO)) {
+								$old_MAIN_MAIL_AUTOCOPY_TO = $conf->global->MAIN_MAIL_AUTOCOPY_TO;
+								$conf->global->MAIN_MAIL_AUTOCOPY_TO = '';
+							}
+
+							if (!empty($sendto)) {
+								$this->composeAndSendAssigneeMessage($sendto, $subject_assignee, $body_assignee, $see_ticket_assignee, $object, $langs);
+							}
+
+							if (!empty($conf->global->TICKET_DISABLE_MAIL_AUTOCOPY_TO)) {
```
eldy, line 192: Use of $conf->global-> is forbidden.
Must use getDolGlobalString or getDolGlobalInt

htdocs/core/triggers/interface_50_modTicket_TicketEmail.class.php
```diff
@@ -461,4 +470,62 @@ private function composeAndSendCustomerMessage($sendto, $base_subject, $body, $s
...
+	{
+		global $conf, $user;
+
+		// Init to avoid errors
+		$filepath = array();
+		$filename = array();
+		$mimetype = array();
+
+		// Send email to assigned user
+		$subject = '['.$conf->global->MAIN_INFO_SOCIETE_NOM.'] '.$langs->transnoentities($base_subject);
```
eldy, line 495: Old code was using
'['.$mysoc->name.']'
why changing into 
$conf->global->MAIN_INFO_SOCIETE_NOM  ?

htdocs/core/triggers/interface_50_modTicket_TicketEmail.class.php
altairis-benjamin: It was used a few lines ahead so I didn't think to use the function. Now fixed.

htdocs/core/triggers/interface_50_modTicket_TicketEmail.class.php
altairis-benjamin: Because it wasn't passed as global, I just changed it.

W1W1-M:
Hello,
This fix would be a great "little" addition for ticketing system !

eldy:
Replaced by #31928 that has no conflict.

PR #25888, opened 2023-09-13
add comment

frederic34:
https://stackoverflow.com/questions/61597263/office-365-xoauth2-for-imap-and-smtp-authentication-fails

IMAP, SMTP scopes are targeted for Exchange resource and not Graph. Whereas User.Read, Mail.ReadWrite are meant for Graph resource.

We do not support generation of tokens that are meant for two resources. Hence the error "Provided value for the input parameter scope is not valid because it contains more than one resource."

You should generate two tokens separately by two calls to /token. 1. One with the IMAP, SMTP scopes generated for the Exchange resource. 2. The other with Graph scopes (User.Read, Mail.ReadWrite) meant for Graph resource.

### DIFF
--- a/htdocs/core/lib/oauth.lib.php
+++ b/htdocs/core/lib/oauth.lib.php
@@ -25,18 +25,66 @@
 
 // Supported OAUTH (a provider is supported when a file xxx_oauthcallback.php is available into htdocs/core/modules/oauth)
 $supportedoauth2array = array(
-	'OAUTH_GOOGLE_NAME'=>array('callbackfile' => 'google', 'picto' => 'google', 'urlforapp' => 'OAUTH_GOOGLE_DESC', 'name'=>'Google', 'urlforcredentials'=>'https://console.developers.google.com/',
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/oauth.lib.php
```diff
@@ -25,18 +25,66 @@
...
+	'returnurl' => '/core/modules/oauth/github_oauthcallback.php'
+);
+// See https://learn.microsoft.com/fr-fr/azure/active-directory/develop/quickstart-register-app#register-an-application
+$supportedoauth2array['OAUTH_MICROSOFT_NAME'] = array(
+	'callbackfile' => 'microsoft',
+	'picto' => 'microsoft',
+	'urlforapp' => 'OAUTH_MICROSOFT_DESC',
+	'name' => 'Microsoft',
+	'urlforcredentials' => 'https://portal.azure.com/',
+	// User.Read is a microsoftgraph scope, if it's not working, do not select it
```
frederic34, line 74: comment is here

PR #25895, opened 2023-09-13
fix sql

frederic34:


fappels:
@frederic34 https://github.com/Dolibarr/dolibarr/pull/25893

PR #25905, opened 2023-09-13
En dk

JonBendtsen:
# NEW [*Add en_DK language*]
[*Adding en_DK which is very identical to en_US, though with Formats and Separators from da_DK*]


eldy:
Thanks for your contribution
However changing file in git for language other than en_US is useless (your fix will be erased by transifex).  Translation is managed on transifex and you must fix translations on transifex directly :
https://explore.transifex.com/dolibarr-association/dolibarr/

JonBendtsen:
@eldy 

Okay, I looked at https://explore.transifex.com/dolibarr-association/dolibarr/ and found this:

"For languages that are a local variant a primary language (for example chile es_CL is a variant of es_ES, or fr_BE that is a variant of fr_FR), translation must be done into dolibarr github repository. The reason is that for such languages, you should not do all translation create only translation files that contains only the "delta" from primary language."

Reading that my understanding is that I need to trim down en_DK with only the differences from en_US, and submit that as a pull request. Do you agree?

eldy:
> @eldy
> 
> Okay, I looked at https://explore.transifex.com/dolibarr-association/dolibarr/ and found this:
> 
> "For languages that are a local variant a primary language (for example chile es_CL is a variant of es_ES, or fr_BE that is a variant of fr_FR), translation must be done into dolibarr github repository. The reason is that for such languages, you should not do all translation create only translation files that contains only the "delta" from primary language."
> 
> Reading that my understanding is that I need to trim down en_DK with only the differences from en_US, and submit that as a pull request. Do you agree?

Thanks for feedback and sorry for the confusion. This note was outdated. I have completely removed it.
Now 100% of translation is done on transifex, whatever is the language. Only the "source" language that is en_US must be commted into git. All other (da_DK for example) must be done in transifex. The sync script run every weeks do the job to integrate the change into git.
Just edit translation for da_DK on transifex and forget.

JonBendtsen:
@eldy I looked at transifex, I signed up, I looked around, I tried to get to review. It is not an easy interface :-( and I'm not gonna spend my time there.

PR #25913, opened 2023-09-14
FIX: Copy line label if there is no description given

noec764:
When we create a supplier Order from another object, lines have to be cloned, if there is no description on a parent line but a label, Label should be copied instead of leaving description empty.


eldy:
The label on lines of object is a deprecated field.
On a line we have  "product id"  (so we can guess "product label", in the correct language of user for multilanguage) and a free text "desc"
Field label is no more used, so should be empty. On which table do you have it filled ?

PR #25917, opened 2023-09-14
add VAT switch mode for customer invoices

samifilali:
# NEW

add VAT switch mode for customer invoices

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -2947,6 +2947,16 @@ if (empty($reshook)) {
 			} else {
 				dol_print_error($db);
 			}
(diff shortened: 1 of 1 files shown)


eldy:
We don't need a switch mode for customer invoice. As the seller, we control how the invoice is generated. So we can choose between round of total or total of round.
And to have a valid accountancy with must use total of round. round of vat was used when invoice was done iin paper and without computer, nowadays it should not be used, otherwise, transfer into accountancy or accounting reports will return results with total that will differ depending on the filter we select. This is no more accepted by modern accounting. So we need exact matching of accountancy report whatever is the way we make the calculation of total of round is the only mathematical solution that can achieve this goal.

PR #25927, opened 2023-09-15
WIP move hook at the right place

altairis-tof:
hook was not executed if previous hook return 1


altairis-tof:
@eldy this PR is a suggestion from @hregis

altairis-tof:
replaced by #25928

PR #25931, opened 2023-09-15
[modulebuilder] FIX handling of the $filter parameter in fetchAll() method

atm-florianm:
# FIX [modulebuilder] handling of `$filter` in method `fetchAll()`

## Rationale
Curently, the way `fetchAll()` handles `$filter` can result in invalid SQL.

The `fetchAll()` of class `KnowledgeRecord` offers a good example (there are minor differences between its implementation and the one you would get if you were to use ModuleBuilder in v18 but they do not impact this bug):

```php
/* not included in this example: inclusion of master.inc.php and knowledgerecord.class.php */
$test = new KnowledgeRecord($db);
echo $kr->fetchAll('', '', 1, 0, array('model_pdf' => 'mymodel')); // returns -1
echo $db->lastqueryerror(); // Unknown column 'mymodel' in 'where clause'
```
because `$value` is not quoted when it is injected (see query below).

```sql
SELECT  […] FROM llx_knowledgemanagement_knowledgerecord as t
WHERE t.entity IN (1) AND (model_pdf IN (mymodel)) LIMIT 1 -1
```

Another problem is a potential warning on this test when $key is not a field (or when it is a field but prefixed with the table alias):
```php
} elseif (in_array($this->fields[$key]['type'], array('date', 'datetime', 'timestamp'))) {
```

## This proposal
Just quoting the field and adding an additional `isset()` for the warning would have been enough, but I think it was also an opportunity to include an improvement proposal.

@eldy maybe you will find a better way to implement these filters with respect to the declared fields (I would have liked to make it consistent with how search works in lists, including `natural_search`, but I didn't exactly know how).

## Further actions
If this PR is merged, it could be interesting to update classes that use old versions of this function with no customizations.
Or maybe even implement a method on CommonObject to make this easier if you think it is worth the effort (a well-filtered fetchAll has use cases, but is currently not used a lot)?


eldy:
@atm-florianm Can you help me fix the conflict ?

atm-florianm:
> @atm-florianm Can you help me fix the conflict ?

I've merged develop into my FIX branch. I did have one little issue with a CodeSniffer auto-fix, which would have turned a PHP8 attribute (`#[AllowDynamicProperties]`) into a regular PHP comment (`// [AllowDynamicProperties]`).

It looks like the github-run CodeSniffer hook didn't mind that, any idea as to why? Did I misconfigure my pre-commit hook?

eldy:
> > @atm-florianm Can you help me fix the conflict ?
> 
> I've merged develop into my FIX branch. I did have one little issue with a CodeSniffer auto-fix, which would have turned a PHP8 attribute (`#[AllowDynamicProperties]`) into a regular PHP comment (`// [AllowDynamicProperties]`).
> 
> It looks like the github-run CodeSniffer hook didn't mind that, any idea as to why? Did I misconfigure my pre-commit hook?

I suppose that AllowDynamicProperties is is not supported by phpcs. However i see no reason why we should use this tag now. You can try to remove it completely.

atm-florianm:
I've removed the attribute completely as advised. However, I distinctly remember seeing things like `$conf->{$modulename}` in some places, but maybe these were from older Dolibarr versions?

If this is still used, we should transition to a different system to prepare for php 9 compliance.

htdocs/modulebuilder/template/class/myobject.class.php
```diff
@@ -460,16 +460,40 @@ public function fetchAll($sortorder = '', $sortfield = '', $limit = 0, $offset =
...
+				if ($key === 'customsql') {
+					// Never use 'customsql' with a value from user input since it is injected as is. The value must be hard coded.
+					$sqlwhere[] = $value;
+					continue;
+				} elseif (isset($this->fields[$columnName])) {
+					$type = $this->fields[$columnName]['type'];
+					if (preg_match('/^integer/', $type)) {
+						if (is_int($value)) {
+							// single value
+							$sqlwhere[] = $key . " = " . $value;
```
eldy, line 473: Cast $value into (int) so php unit test will not comolain about possible sql injection

htdocs/modulebuilder/template/class/myobject.class.php
```diff
@@ -460,16 +460,40 @@ public function fetchAll($sortorder = '', $sortfield = '', $limit = 0, $offset =
...
+						} elseif (is_array($value)) {
+							if (empty($value)) continue;
+							$sqlwhere[] = $key . ' IN (' . implode(',', array_map('intval', $value)) . ')';
```
eldy, line 476: When forging a IN sql, use db->sanitize around parameter to have phpunit not complain...

atm-florianm:
@eldy  I'm not sure why the phpcs check failed. Was it related to my changes?

PR #25944, opened 2023-09-17
fix translation

sonikf:


htdocs/langs/en_US/donations.lang
```diff
@@ -32,4 +32,4 @@ DONATION_ART238=Show article 238 from CGI if you are concerned
 DONATION_ART978=Show article 978 from CGI if you are concerned
 DonationPayment=Donation payment
 DonationValidated=Donation %s validated
-DonationUseThirdparties=Use an existing third party as coordinates of donors
+DonationUseThirdparties=Use an existing third party as coordinators of donors
```
eldy, line 35: I think good text should be
"Use the address of an existing thirdparty as the address of the donor"

sonikf: Yes much better, fixed

PR #25946, opened 2023-09-17
adding stock menu point to services menu

JonBendtsen:
# NEW [*Adding stock menu point to services menu*]
[*Now easier than ever to list stocks for services since the stocks menu point from the Products menu has been copied to services menu and modified to list services.*]


JonBendtsen:
I tried with `&& $conf->global->STOCK_SUPPORTS_SERVICES` such that the stock menu point would only list if `STOCK_SUPPORTS_SERVICES` was enabled in the stock module. But it showed up regardless, so I ended up commenting that piece of code with the hope that someone who knows PHP and Dolibarr a little better quickly can modify my pull request and get it to respect the `STOCK_SUPPORTS_SERVICES` slider.

If you do that, then I can even help testing it.

JonBendtsen:
adding_stocks_2_services_menu
This is where the pull request adds a menu point

htdocs/core/menus/standard/eldy.lib.php
```diff
@@ -2064,6 +2064,9 @@ function get_left_menu_products($mainmenu, &$newmenu, $usemenuhider = 1, $leftme
 			$newmenu->add("/product/index.php?leftmenu=service&amp;type=1", $langs->trans("Services"), 0, $user->hasRight('service',  'read'), '', $mainmenu, 'service', 0, '', '', '', img_picto('', 'service', 'class="pictofixedwidth"'));
 			$newmenu->add("/product/card.php?leftmenu=service&amp;action=create&amp;type=1", $langs->trans("NewService"), 1, $user->hasRight('service',  'creer'));
 			$newmenu->add("/product/list.php?leftmenu=service&amp;type=1", $langs->trans("List"), 1, $user->hasRight('service',  'read'));
+			if (isModEnabled('Stock') && $conf->global->STOCK_SUPPORTS_SERVICES) {
```
eldy, line 2067: Can you replace $conf->global->STOCK... with
getDolGlobalString('STOCK...')

JonBendtsen: @eldy  I think I made the change you requested.

FYI: I also did a grep for '$conf->global->STOCK_SUPPORTS_SERVICES' and it is mostly (37 files)
`    empty($conf->global->STOCK_SUPPORTS_SERVICES)
`
but also one like mine:
./htdocs/admin/stock.php:
`
print $form->selectarray("STOCK_SUPPORTS_SERVICES", $arrval, $conf->global->STOCK_SUPPORTS_SERVICES);
`
showed up in a total of 38 other files - maybe all those they should be fixed too?

eldy: Yes, we should have other ref fixed.

PR #25947, opened 2023-09-17
adding variants menu entry to services menu

JonBendtsen:
# NEW [*adding variants menu entry to services menu*]
[*Now variants are also easily accessible for services, previously the menu entry was only listed under products*]


JonBendtsen:
Both Product and Service Variants should probably check if user has permission to variants, but because Product Variants does not include a check for this, then I could not copy it to services.

Picture shows where the new `Variant` menu entry was added
adding_variants_2_services_menu

PR #25961, opened 2023-09-18
FIX Warehouse Parent error

samifilali:
# FIX
fix the problem of creating a shipment if it is parent and negative stock is authorized

### DIFF
--- a/htdocs/expedition/card.php
+++ b/htdocs/expedition/card.php
@@ -1247,7 +1247,7 @@ if ($action == 'create') {
 					}
 
 					$warehouseObject = null;
-					if (count($warehousePicking) == 1 || $conf->global->STOCK_ALLOW_NEGATIVE_TRANSFER || !($line->fk_product > 0) || !isModEnabled('stock')) {     // If warehouse was already selected or if product is not a predefined, we go into this part with no multiwarehouse selection
+					if (count($warehousePicking) == 1 || !empty($conf->global->STOCK_ALLOW_NEGATIVE_TRANSFER) || !($line->fk_product > 0) || !isModEnabled('stock')) {     // If warehouse was already selected or if product is not a predefined, we go into this part with no multiwarehouse selection
 						print '<!-- Case warehouse already known or product not a predefined product -->';
 						//ship from preselected location
 						$stock = + (isset($product->stock_warehouse[$warehouse_id]->real) ? $product->stock_warehouse[$warehouse_id]->real : 0); // Convert to number
--- a/htdocs/expedition/card.php
+++ b/htdocs/expedition/card.php
@@ -1247,7 +1247,7 @@ if ($action == 'create') {
 					}
 
 					$warehouseObject = null;
-					if (count($warehousePicking) == 1 || !($line->fk_product > 0) || !isModEnabled('stock')) {     // If warehouse was already selected or if product is not a predefined, we go into this part with no multiwarehouse selection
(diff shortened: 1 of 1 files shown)


eldy:
I don't understand the fix. 
The test is to know if a warehouse was already known or not. It should have no relation with the option to allow negative or not.
Can you describe the situation you face and how to reproduce the bug ?

samifilali:
When I attempt to ship, even though negative stock is allowed, if my selected warehouse is a parent warehouse, I cannot ship the out-of-stock items with it.

eldy:
> When I attempt to ship, even though negative stock is allowed, if my selected warehouse is a parent warehouse, I cannot ship the out-of-stock items with it.

Can you provide a screenshot with data (and full screen) ?

samifilali:
image

eldy:
> image

On previous step, you probably select the warehouse "Rack A" ? If yes, it mean you force the warehouse you want to use for stock movement so only stock into Rack A is allowed to fill shipment creation. Can you try without setting warehouse at previous step ?

samifilali:
The problem is precisely there. Indeed, I should be able to carry out stock movements by selecting a child warehouse if negative stock is authorized

eldy:
> The problem is precisely there. Indeed, I should be able to carry out stock movements by selecting a child warehouse if negative stock is authorized

You should be able if you are in the mode all warehouse can be used.
If you force a warehouse W, you can only pick in warehouse W, not into a child nor a parent of W.

samifilali:
yes, but if I select a warehouse W and my product A has no stock in this warehouse, I must be able to ship it anyway (if I authorise negative stock).

eldy:
> yes, but if I select a warehouse W and my product A has no stock in this warehouse, I must be able to ship it anyway (if I authorise negative stock). 
> 

It depends.
If you select the warehouse at the first step, it means "i want to ship only what isbinside this warehouse" so if there is not stock in it, you should not be able to ship it at all.
But if you do not force the warehouse when building the shipment from the order,  you should be able to ship it.

atm-thibaultf:
> It depends.
If you select the warehouse at the first step, it means "i want to ship only what isbinside this warehouse" so if there is not stock in it, you should not be able to ship it at all. 
But if you do not force the warehouse when building the shipment from the order, you should be able to ship it.

If I authorize stock to be negative, I should be able to ship from any warehouse no matter if I select one warehouse from the order or not.

Anyway, in fact, the only case when it works correctly is the case you talk about saying it shouldn't work (when I select a warehouse from the order). We have 3 cases explained below (assuming that I have allowed the negative stock and I don't have any stock in warehouses for my product) :

1. I select a warehouse **with no child warehouse** (it's important) from the order, I can ship my product from it : **=> Ok, it works correctly**
image

2. I select a warehouse **with child warehouse** from the order, I can't ship my product from it : **=> Not OK, I should be able to ship it like below**
image

3. I don't select any warehouse from the order, I can't ship my product : **=> Not OK, I should be able to ship it like below and like you said it**
image

samifilali:
@eldy ?

eldy:
@samifilali  I still not see what is the trouble.

Case A
If you select a warehouse at first step to say "I want to ship only products that are inside this warehouse", then you are not able to select a product that is not in warehouse. It does not depends on stock qty.

Case B
If you want to be able to ship from any warehouse, do not select a warehouse at first step. And then on next page you should be able to select any product (and if option "disallow negative stock is on", only products with enough in stock)

eldy:
Behaviour seems the one expected. Also a lot of change were done in this process since v17 so i think this PR may be deprecated, so i close it (i can't reproduce trouble on v22).

PR #25962, opened 2023-09-18
Fix #25860 not allow encrypt data if key and instance_unique_id are empty

### LINKED ISSUE #25860: Can't decrypt encrypted data if instance_unique_id is not defined
### Bug

[dolEncrypt](https://github.com/Dolibarr/dolibarr/blob/9c903ecf56d3650bcacce4b47eb0e4e26f8f3a1b/htdocs/core/lib/security.lib.php#L122C10-L122C20) function allows to encrypt data even if instance_unique_id is not defined.

But [dolDecrypt](https://github.com/Dolibarr/dolibarr/blob/9c903ecf56d3650bcacce4b47eb0e4e26f8f3a1b/htdocs/core/lib/security.lib.php#L177) function fails if it is empty.

https://github.com/Dolibarr/dolibarr/blob/9c903ecf56d3650bcacce4b47eb0e4e26f8f3a1b/htdocs/core/lib/security.lib.php#L193C4-L195

I'm not sure what is the best solution: allow to decrypt data even or not allow to encrypt data if instance_unique_id is empty.

### Environment Version

18

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

tuxgasy:
# Fix #25860
Not allow encrypt data if key and instance_unique_id are empty (as decrypt function) to avoid losing data into database configuration llx_const.

The config **instance_unique_id** is added automatically only for new installation since Dolibarr 10. Some users may not added it after an upgrade.

### DIFF
--- a/htdocs/core/lib/security.lib.php
+++ b/htdocs/core/lib/security.lib.php
@@ -144,6 +144,10 @@ function dolEncrypt($chain, $key = '', $ciphering = 'AES-256-CTR', $forceseed = 
 	$newchain = $chain;
 
 	if (function_exists('openssl_encrypt') && empty($dolibarr_disable_dolcrypt_for_debug)) {
+                if (empty($key)) {
+                        return 'Error dolEncrypt encrypt key is empty';
+                }
+
 		$ivlen = 16;
 		if (function_exists('openssl_cipher_iv_length')) {
 			$ivlen = openssl_cipher_iv_length($ciphering);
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/security.lib.php
```diff
@@ -144,6 +144,10 @@ function dolEncrypt($chain, $key = '', $ciphering = 'AES-256-CTR', $forceseed =
 	$newchain = $chain;
 
 	if (function_exists('openssl_encrypt') && empty($dolibarr_disable_dolcrypt_for_debug)) {
+                if (empty($key)) {
+                        return 'Error dolEncrypt encrypt key is empty';
```
eldy, line 148: Instead of return an error string, i would make
return $chain;
So encryption is disabled with nobrisknto corrupt data with an error string when we save the result in database.

PR #25965, opened 2023-09-19
FIX group by in count elements on invoice list

lvessiller-opendsi:
FIX group by in count elements on invoice list

When you have a sub-request with a "GROUP BY", the SQL request used to count elements remove all "GROUP BY".
So the sub-request is wrong (GROUP BY is missing) and the count request is also wrong.

**Example of subrequest from "printFieldListFrom" hook :**
LEFT JOIN (
	SELECT ee.fk_target as id_invoice, GROUP_CONCAT(DISTINCT link_src.rowid SEPARATOR ',') as link_propal
	FROM llx_element_element as ee 
	LEFT JOIN llx_propal as link_src ON link_src.rowid = ee.fk_source AND ee.sourcetype='propal'
	WHERE ee.targettype = 'facture'
	GROUP BY ee.fk_target
) as link_fac ON link_fac.id_invoice = f.rowid

**Module builder**
And fix on object list of module builder

### DIFF
--- a/htdocs/compta/facture/list.php
+++ b/htdocs/compta/facture/list.php
@@ -942,7 +942,6 @@ $nbtotalofrecords = '';
 if (!getDolGlobalInt('MAIN_DISABLE_FULL_SCANLIST')) {
 	/* The fast and low memory method to get and count full list converts the sql into a sql count */
 	$sqlforcount = preg_replace('/^'.preg_quote($sqlfields, '/').'/', 'SELECT COUNT(*) as nbtotalofrecords', $sql);
-	$sqlforcount = preg_replace('/GROUP BY .*$/', '', $sqlforcount);
 	$resql = $db->query($sqlforcount);
 	if ($resql) {
 		$objforcount = $db->fetch_object($resql);
--- a/htdocs/modulebuilder/template/myobject_list.php
+++ b/htdocs/modulebuilder/template/myobject_list.php
@@ -379,7 +379,6 @@ $nbtotalofrecords = '';
 if (!getDolGlobalInt('MAIN_DISABLE_FULL_SCANLIST')) {
 	/* The fast and low memory method to get and count full list converts the sql into a sql count */
 	$sqlforcount = preg_replace('/^'.preg_quote($sqlfields, '/').'/', 'SELECT COUNT(*) as nbtotalofrecords', $sql);
-	$sqlforcount = preg_replace('/GROUP BY .*$/', '', $sqlforcount);
 	$resql = $db->query($sqlforcount);
 	if ($resql) {
 		$objforcount = $db->fetch_object($resql);


eldy:
The pb looks more inside whatvyou do in your external moduel.
Here you add a left join on a subrequest.
And this subrequest is a full database scan on 2 tables with join with no restriction from parent request. Using a subrequest with an aggregate like group_contat inside a left join is clearly a bad practice.
Here due to the fact that your subrequest is a full database scan, it looks better to get the information you get by doing the subrequest inside the php loop than inside the main sql.
Removing the "group by" to make the count may generate errors for a lot of case. And the use case you present to justify the removal is clearly a use case that must not be implemented this way (it will hang on large database and looks not compatible with other database). Converting lines into column is a often a practice that generate troubles. Can you describe what you want to do (business need) so i can suggest a better alternative.

lvessiller-opendsi:
@eldy 
I understand your point of view but even if we don't scan all table with a WHERE condition, we can't use a sub-request with a "GROUP BY". And to remove all "group by" is not used in this list because there is no GROUP BY in SQL request and you have not hook to add it.

**The business need :** 
- the need is to have a new column in the invoice list with all proposals linked  (as source or target)

eldy:
> @eldy 
> I understand your point of view but even if we don't scan all table with a WHERE condition, we can't use a sub-request with a "GROUP BY". And to remove all "group by" is not used in this list because there is no GROUP BY in SQL request and you have not hook to add it.


Yes we can't, but we should never have to do this.
Because if we include a subrequest with a group by into a list of fields of a select, it means we are doing something that can be done more easily and with better performance by doing a php fetch by using hooks during the generation of the list.

> **The business need :** 
> - the need is to have a new column in the invoice list with all proposals linked  (as source or target)

So why don't you use hook printFieldListXxxx that are designed for that ? Inside this hook, you make the fetch to get the list and you output the content of this column.

lvessiller-opendsi:
@eldy 

> So why don't you use hook printFieldListXxxx that are designed for that ? Inside this hook, you make the fetch to get the list and you output the content of this column.

Yes you can, but if you want to filter on a proposal reference linked to the invoice of this list, How can I do this ?

eldy:
For filter purpose, we now must avoid to add join just to get data to use in the filter. 
Instead the recommended method that will be compliant with the future advanced search tool is
"1 search criteria" = "1 instruction added into the where and only to the where".
So for you need of filtering, it should be an addition into the WHERE using the printFieldListWhere()

For example, your hook should return a string like this :
"AND EXISTS (select rowid from llx_element_element as ee, llx_propal as p WHERE ee.targettype = 'facture' AND ee.fk_target = f.rowid AND ee.sourcetype='propal' AND ee.fk_source = p.rowid AND p....condition on the propal you search)"

I did not tested this example, I wrote it on the fly, this is to show you the idea... This is easier to understand, compliant with advanced filter that will use multicriteria with OR and AND combinaisons, and is compatible with postgres (GROUP_CONCAT seems not standard SQL) and should have a better execution plan (better performances if we have the good indexes)

lvessiller-opendsi:
Thanks a lot for your answer. 
It's a good way to do.
I didn't think to do like this.

PR #25969, opened 2023-09-19
New updating for edit NRC in skills tabs in user fiche

lamrani002:
# NEW|New updating in tab skills in user fiche for show & edit national registration number

### DIFF
--- a/htdocs/hrm/skill_tab.php
+++ b/htdocs/hrm/skill_tab.php
@@ -113,6 +113,15 @@ if (empty($reshook)) {
 		}
 	}
 
+	// update national_registration_number
+	if ($action == 'setnational_registration_number') {
+		$object->national_registration_number = (string) GETPOST('national_registration_number', 'alphanohtml');
+		$result = $object->update($user);
+		if ($result < 0) {
+			setEventMessages($object->error, $object->errors, 'errors');
+		}
+	}
+
 	if ($action == 'addSkill') {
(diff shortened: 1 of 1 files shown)


htdocs/hrm/skill_tab.php
```diff
@@ -314,9 +323,21 @@
...
+			$ret .= '<input type="hidden" name="action" value="setnational_registration_number">';
+			$ret .= '<input type="hidden" name="token" value="'.newToken().'">';
+			$ret .= '<input type="hidden" name="id" value="'.$object->id.'">';
+			$ret .= '<input type="text" name="national_registration_number" value="'.$object->national_registration_number.'">';
+			$ret .= '<input type="submit" class="button smallpaddingimp" name="modify" value="'.$langs->trans("Modify").'"> ';
+			$ret .= '<input type="submit" class="button smallpaddingimp button-cancel" name="cancel" value="'.$langs->trans("Cancel").'">';
+			$ret .= '</form>';
+			print $ret;
+		} else {
+			print $object->national_registration_number;
```
eldy, line 339: When not in esit mode, you can keep the use of 
                print showValueWithClipboardCPButton(!empty($object->national_registration_number) ? $object->national_registration_number : '');

So we son't loose the feature to copy oaste the number when in read mode.

lamrani002: it's done i fix it.

PR #25976, opened 2023-09-20
fix line rank in massaction supplier order to supplier invoice

frederic34:
also related to https://github.com/Dolibarr/dolibarr/pull/25950

### DIFF
--- a/htdocs/fourn/commande/list.php
+++ b/htdocs/fourn/commande/list.php
@@ -507,8 +507,8 @@ if (empty($reshook)) {
 								false,
 								$lines[$i]->array_options,
 								$lines[$i]->fk_unit,
-								// we use the id of each orderline, not the id of the first one stored in $objecttmp->origin_id
-								$lines[$i]->id,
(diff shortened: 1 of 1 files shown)


htdocs/fourn/commande/list.php
```diff
@@ -502,12 +502,13 @@
...
-								$lines[$i]->rang,
+								// we dont use the rank from orderline because we may have lines from several orders
+								-1,
 								false,
 								$lines[$i]->array_options,
 								$lines[$i]->fk_unit,
-								// we use the id of each order, not the id of the first one stored in $objecttmp->origin_id
-								$lines[$i]->fk_commande,
+								// we use the id of each orderline, not the id of the first one stored in $objecttmp->origin_id
+								$lines[$i]->id,
```
eldy, line 511: This value is the origin_id.
It  is used inside the addline method to insert it into database with
			$this->line->origin = $this->origin;
			$this->line->origin_id = $origin_id;

So the value must represent an id of the type in this->origin. Currently this->origin is 'order_supplier' so the id must be the purchase order id and not the order line id

PR #25979, opened 2023-09-20
NEW key to detect a new line for reception of serial numbers

lvessiller-opendsi:
NEW key to detect a new line for reception of serial numbers (barcode reader)
- add admin setup in reception module

1. can configure a key (ASCII code) to detect a new line for reception of serial numbers
image

- add javascript on each line of batch product if a key is set in setup of reception module

1. get focus on first serial input
2. a new line is added each time we read the ASCII code 
3. go on next serial input if found or go on create button

image


eldy:
Can you confirm me the goalofthis feature:
- If you set a key,when on shipment for a product that need a serial, the predefined quantity is no more auto-filled and is 0 instead. Right ?
- When you push the key on keyboard, it autofill the batch with the first serial value found, right ? In which warehouse is done the pick ?

lvessiller-opendsi:
> Can you confirm me the goalofthis feature:
> 
> * If you set a key,when on shipment for a product that need a serial, the predefined quantity is no more auto-filled and is 0 instead. Right ? Here, it's on a receipt of products (not on shipment) and the first serial input take the focus and when we scan a serial and the "key to detect" is detected, a new line is created and take the focus and so on ... and qty are all set to 1 at the end (uses dispatcher)
> * When you push the key on keyboard, it autofill the batch with the first serial value found, right ? In which warehouse is done the pick ? In reception you have new serials numbers and the barcode reader scan the serial of the product and when a key as enter the field is filled and create a new line and the focus is set on this new serial input. The warehouse will be select at the begining. The goal is to pick-up materials (using serial numbers) on the same warehouse with a barcode reader.

eldy:
I see. We are speaking about the same process to autofill serial with successive barcode scanning.
You are on the good page to implement this feature: fourn/commande/disparch.php
But such a similar feature was already introduced into the page to make inventories (see page product/inventory/inventory.php).
We should implement for reception the same way: with the link "complete quantity by scanning".
We clink on the link, this open the barcode scanning tool and when cliking on "Add", it edit the javascript lines to autofill.
Because there is several way to make input of serial: manual, successive barcode scanning, copy paste of a xls file list of barcode, and mix of this. And this tool was designed thinking in this.

This tool, currently implemented and working in inventory, was designed to be used also for reception and shipment disparching pages (may be dev is not complete, but we should go on this directy), and for the need you try to implement, we must rely on this same process/tool.

lvessiller-opendsi:
I barcode scan tool exists in develop

PR #25984, opened 2023-09-21
modify select_change_categorie to get the category hierarchy in massaction

iouston:
modify select_change_categorie to get the category hierarchy.
But in this case, we lose the category color marker.

Should we add a configuration variable to have one mode or the other? or add coloring according to the method's mode?

### DIFF
--- a/htdocs/core/tpl/massactions_pre.tpl.php
+++ b/htdocs/core/tpl/massactions_pre.tpl.php
@@ -73,7 +73,8 @@ if ($massaction == 'preaffecttag' && isModEnabled('category')) {
 	$formquestion = array();
 	if (!empty($categ_types)) {
 		foreach ($categ_types as $categ_type) {
-			$categ_arbo_tmp = $form->select_all_categories($categ_type['code'], null, 'parent', null, null, 2);
+			//$categ_arbo_tmp = $form->select_all_categories($categ_type['code'], null, 'parent', null, null, 2); //does not show the category hierarchy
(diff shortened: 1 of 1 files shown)


htdocs/core/tpl/massactions_pre.tpl.php
```diff
@@ -73,7 +73,8 @@
 	$formquestion = array();
 	if (!empty($categ_types)) {
 		foreach ($categ_types as $categ_type) {
-			$categ_arbo_tmp = $form->select_all_categories($categ_type['code'], null, 'parent', null, null, 2);
+			//$categ_arbo_tmp = $form->select_all_categories($categ_type['code'], null, 'parent', null, null, 2); //does not show the category hierarchy
+			$categ_arbo_tmp = $form->select_all_categories($categ_type['code'], '', 'parent', 64, 0, 1); //show the category hierarchy
```
eldy, line 77: Using the new param 2 instead of 1 seems to return a more completearray
Which data do you expect to have with param 1 you don't have with param 2 ?

eldy:
> modify select_change_categorie to get the category hierarchy. But in this case, we lose the category color marker.
> 
> Should we add a configuration variable to have one mode or the other? or add coloring according to the method's mode?

No need for a configuration variable when we can have the best of the 2 world without any configuration. I added an output mode = 3 into the method to list tags.

iouston:
Thank you for this enhancement, this was the best solution, sure !

iouston:
it's 8:49. your comment was 5 hours ago. When are you sleeping ?

PR #25987, opened 2023-09-21
FIX #25985 [Ticket] tzuser param to ticket index

### LINKED ISSUE #25985: [Bug] date_c need tzuser when we create a ticket
### Bug

The date of creation of a ticket is not set on the user timezone

### Environment Version

17 - 18

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

It should set the creation date of a ticket with "tzuser" argument to have the good timezone

_No response_

### Attached files

_No response_

evarisk-micka:


eldy:
A date of creation is not a data coming from a user. So there is no User timezone. It can be defined into a batch context. And result must be the same whatever is the location of the user. 
The result is to be storedina php memory data, so must be a GMT date.
dol_now looks correct. Trouble issurely some where else, probably when outputing the data to the user. The output must be done according to the user timezone.
Can you provide a screenshot where data is wrongly shown ?

PR #25989, opened 2023-09-21
NEW expedition serial for barcode reader

lvessiller-opendsi:
NEW key to detect a new line for shipping of serial numbers (barcode reader)
- add admin setup in shipping module

1. can configure a key (ASCII code) to detect a new line for shipping of serial numbers
image

- add javascript on each line of batch product if a key is set in setup of shipping module

1. set qty to 0 on each qty where the product has an unique serial number (not a product lot)
2. get focus on first serial input using to search a serial number
3. search qty for this serial number and set to 1
4. go on next serial input if found or go on create button

image

### DIFF
--- a/htdocs/admin/expedition.php
+++ b/htdocs/admin/expedition.php
@@ -54,6 +54,23 @@ if (empty($conf->global->EXPEDITION_ADDON_NUMBER)) {
 	$conf->global->EXPEDITION_ADDON_NUMBER = 'mod_expedition_safor';
 }
 
+// Set this to 1 to use the factory to manage constants. Warning, the generated module will be compatible with version v15+ only
+$useFormSetup = 1;
+
+if (!class_exists('FormSetup')) {
+	require_once DOL_DOCUMENT_ROOT . '/core/class/html.formsetup.class.php';
+}
+
+$formSetup = new FormSetup($db);
+
+// Enable access for the membership record
+$batchNewLineAsciiCodeList = array(
+	-1 => $langs->trans('AsciiKeyNone'),
+	13 => $langs->trans('AsciiKeyEnter'),
+	59 => $langs->trans('AsciiKeySemiColon'),
+);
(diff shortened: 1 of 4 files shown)


eldy:
My understanding of this change is:
- If a "key to detect" is set, then when creating a shipment, the line for product that are serial are no more autofilled by first serial found, right ?
- Instead it is 0 with a focus and typing enter take the first serial found, right ?
Can you confirm ?

lvessiller-opendsi:
> My understanding of this change is:
> 
> * If a "key to detect" is set, then when creating a shipment, the line for product that are serial are no more autofilled by first serial found, right ? Yes
> * Instead it is 0 with a focus and typing enter take the first serial found, right ?
>   Can you confirm ? Yes

eldy:
@lvessiller-opendsi 
I see. 

Let me give you a vision of a future and current work on shipment/reception.

All definition of lot/serial to fill a shipment, in a future, should be done into the second tab "Shipment of distribution" of a shipment.
This means that, in a future, the page you have modified "expedition/card.php" will be used only to define the quantity expected to be shipped. Only quantity, but no Lot or serial data.
Because some business need to fill the serial or lot when making the shipment, but some other need to fill the serial or lot when the shipment is really done, and not when just prepared (so only when the product are picked randomly from a warehouse). 
So having a shipment process in 2 steps (introduced in v18) was  the solution to match all needs:
- 1 step to generate a shipment
- 1 step (the new tab "Shipment distribution") to fill lot, serial and real warehouse to use with custom quantities.

The 2 steps with v19+ will  be very similar but with this difference **in bold**:
- 1 step to generate a shipment **with only the expected quantity (currently we fill the qty but also lot/serial "automatically" when we generate the shipment)**
- 1 step (the new tab "Shipment distribution") to fill lot, serial and real warehouse to use using the **scanning tool** introduced into v16 or v17.

Lucas will try to work on implementing the "Scanning tool" to fill lines from the page "expedition/dispatch.php?id="

This is to explain that your PR may become useless so why i keep it pending...

lvessiller-opendsi:
I barcode scan tool exists in develop

PR #26001, opened 2023-09-23
FIX missing entity filter to customize ticket dictionaries by entity

hregis:
FIX add field "entity" only in $tabfieldinsert

### DIFF
--- a/htdocs/ticket/class/ticket.class.php
+++ b/htdocs/ticket/class/ticket.class.php
@@ -1242,7 +1242,8 @@ class Ticket extends CommonObject
 
 		$sql = "SELECT rowid, code, label, use_default, pos, description";
 		$sql .= " FROM ".MAIN_DB_PREFIX."c_ticket_type";
-		$sql .= " WHERE active > 0";
+		$sql .= " WHERE entity = ".getEntity('c_ticket_type');
+		$sql .= " AND active > 0";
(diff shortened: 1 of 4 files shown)


htdocs/ticket/class/ticket.class.php
```diff
@@ -1242,7 +1242,8 @@ public function loadCacheTypesTickets()
 
 		$sql = "SELECT rowid, code, label, use_default, pos, description";
 		$sql .= " FROM ".MAIN_DB_PREFIX."c_ticket_type";
-		$sql .= " WHERE active > 0";
+		$sql .= " WHERE entity = ".getEntity('c_ticket_type');
```
eldy, line 1245: With getEntity in a SQL, you must use
...entity IN (".getEntity().")"

hregis: @eldy not for dictionary table, getEntity return 1 if all entities use the same dictionary or the current entity if the dictionary is personalized

eldy: So even in this case, we can use the IN.
For a more uniform situation, it should be better to have IN everywhere for getEntity, even if in execution th in become a IN (1)

hregis: @eldy ok it's fixed

PR #26004, opened 2023-09-23
FIX #25983

### LINKED ISSUE #25983: Bug - Cannot create payment mode dictionary entry - "Field 'entity' is required"
### Bug

Since 18.0.x it's no longer possible to create a payment mode (under `/admin/dict.php?id=13`). An error is shown: `Field 'entity' is required`

### Environment Version

18.0.x

### Environment OS

Debian 12.1

### Environment Web server

_No response_

### Environment PHP

7.4.33

### Environment Database

MySQL 8.0.33

### Environment URL(s)

_No response_

Expected: Payment mode is created

Actual: Payment mode cannot be created as Dolibarr is not populating the mandatory `entity` database field correctly.

In an environment with no additional modules (i.e. no multicompany support added):

1. Navigate to the "Payment modes" dictionary page (`/admin/dict.php?id=13`)
2. Create a new payment mode, i.e. code `TEST`, label `Test`, type `Payment Type - Customer and Vendor`, click 'ADD"
3. Observe: Error is shown: `Field 'entity' is required`

### Attached files

_No response_

tiaris:
# FIX|Fix #25983 Creation of paymentmethods in dictionnary]
An error is shown: Field 'entity' is required

Same error on other dictionnaries (payment terms ..) because of 
code change in function fieldList(
if ($value == 'entity' && isset($obj->$value)) {
			$withentity = $obj->$value;
			continue;
		}
When adding an item, $obj->entity is always empty so $withentity remain empty instead of sending the entity number

### DIFF
--- a/htdocs/admin/dict.php
+++ b/htdocs/admin/dict.php
@@ -1557,6 +1557,8 @@ if ($id > 0) {
 		print '<tr class="oddeven nodrag nodrop nohover">';
 
 		$obj = new stdClass();
+		$ent = getEntity($tabname[$id]); 
+		if (!empty($ent)) $obj->entity = $ent; // always needed
 		// If data was already input, we define them in obj to populate input fields.
 		if (GETPOST('actionadd')) {
 			foreach ($fieldlist as $key => $val) {
--- a/htdocs/admin/dict.php
+++ b/htdocs/admin/dict.php
@@ -1557,7 +1557,7 @@ if ($id > 0) {
 		print '<tr class="oddeven nodrag nodrop nohover">';
 
 		$obj = new stdClass();
-		$ent = getEntity($tabname[$id]); 
+		$ent = getEntity($tabname[$id]);
 		if (!empty($ent)) $obj->entity = $ent; // always needed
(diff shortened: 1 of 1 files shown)


hregis:
@tiaris i always send a fix with this PR https://github.com/Dolibarr/dolibarr/pull/26001

tiaris:
@hregis you should mark on the issue taht you are working on it ...

hregis:
@tiaris je n'avais pas vu l'issue, j'ai découvert le bug en corrigeant un autre bug 😅

eldy:
> @tiaris i always send a fix with this PR https://github.com/Dolibarr/dolibarr/pull/26001

@hregis Can you confirm your fix #26001 is enough and this pr become useless ?

hregis:
@eldy @tiaris 
yes, the problem came from the fact that you should not put the "entity" field in $tabfield because it adds an unnecessary "entity" field when adding a record.


Capture d’écran 2023-09-27 à 06 55 17


the "entity" field must be added only in $tabfieldinsert.

PR #26006, opened 2023-09-23
FIX #25973 false link for order widget

### LINKED ISSUE #25973: Orders to process main menu bug
### Bug

In the main menu, when I click on Orders (Commandes in French),
image
the link go to 
https://myserver/commande/list.php?search_status=-3&mainmenu=commercial&leftmenu=orders
instead of
https://myserver/commande/list.php?search_status=-2&mainmenu=commercial&leftmenu=orders
In the code, the search_status is set in hard.

### Environment Version

17.0.3

### Environment OS

_No response_

### Environment Web server

Hosted

### Environment PHP

_No response_

### Environment Database

MySQL

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

tiaris:
# FIX|Fix #25973 Linking error in order widget
widget is for orders that must be treated. inconsistant with the list pointed
The link is set with search_status = -3  wich will include also treated orders set it to -2
also changed the sql order wiich hast a facture=0 criteria, wich is not correct, because you can have paid orders to treat also. 
List pointed with link does not have a paid criteria.

### DIFF
--- a/htdocs/commande/class/commande.class.php
+++ b/htdocs/commande/class/commande.class.php
@@ -3578,7 +3578,7 @@ class Commande extends CommonOrder
 		}
 		$sql .= $clause." c.entity IN (".getEntity('commande').")";
 		//$sql.= " AND c.fk_statut IN (1,2,3) AND c.facture = 0";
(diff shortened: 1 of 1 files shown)


eldy:
As it is a bug, to validate the fix, can you describe how to reproduce the bug?

eldy:

PR #26036, opened 2023-09-26
add extrafields and defaulttpldir for hook parameters

samifilali:
# FIX

add extrafields and defaulttpldir for hook parameters

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -4909,10 +4909,10 @@ abstract class CommonObject
 			//if (is_object($hookmanager) && (($line->product_type == 9 && !empty($line->special_code)) || !empty($line->fk_parent_line)))
 			if (is_object($hookmanager)) {   // Old code is commented on preceding line.
 				if (empty($line->fk_parent_line)) {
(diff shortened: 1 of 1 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -4909,10 +4909,10 @@ public function printObjectLines($action, $seller, $buyer, $selected = 0, $dateS
 			//if (is_object($hookmanager) && (($line->product_type == 9 && !empty($line->special_code)) || !empty($line->fk_parent_line)))
 			if (is_object($hookmanager)) {   // Old code is commented on preceding line.
 				if (empty($line->fk_parent_line)) {
-					$parameters = array('line'=>$line, 'num'=>$num, 'i'=>$i, 'dateSelector'=>$dateSelector, 'seller'=>$seller, 'buyer'=>$buyer, 'selected'=>$selected, 'table_element_line'=>$line->table_element);
+					$parameters = array('line'=>$line, 'num'=>$num, 'i'=>$i, 'dateSelector'=>$dateSelector, 'seller'=>$seller, 'buyer'=>$buyer, 'selected'=>$selected, 'table_element_line'=>$line->table_element, 'extrafields'=>$extrafields, 'defaulttpldir'=>$defaulttpldir);
```
eldy, line 4912: $extrafields has been stated to be a global variable (search on "skeleton global variables" on wiki). Only one instance for all php execution is required. 
So code must access variable $extrafields with global $extrafields. So we must not add it as parameter.
This is not the case for $defaultdir. So can you remove extrafields from list ?

PR #26038, opened 2023-09-26
add a trigger BEFORE delete

rycks:
If we make some interactions with trigger, we can add something when a user is added to a ficheinter but we can't do the job in case of FICHINTER_DELETE_CONTACT.

Example with FICHINTER: 

- on FICHINTER_ADD_CONTACT i could send an email to the contact to explain what we must to do on that interventions
- but if sometime after we delete that contact i can't send a new email with new instructions like "please forget the previous mail about that intervention, it's not for you"

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -1170,6 +1170,11 @@ abstract class CommonObject
 		// phpcs:enable
 		global $user;
 
+		$result = $this->call_trigger(strtoupper($this->element).'_BEFORE_DELETE_CONTACT', $user);
+		if ($result < 0) {
+			$this->db->rollback();
+			return -1;
+		}
 
 		$this->db->begin();
 


htdocs/core/class/commonobject.class.php
```diff
@@ -1170,6 +1170,11 @@ public function delete_contact($rowid, $notrigger = 0)
 		// phpcs:enable
 		global $user;
 
+		$result = $this->call_trigger(strtoupper($this->element).'_BEFORE_DELETE_CONTACT', $user);
```
eldy, line 1173: A definition of a trigger is that action is inside the transaction so after the begin.
There is no need for _BEFORE_ trigger for a delete because for a delete action trigger must always be executed before the deletion of main object.

Creation:
- Create
- Trigger create

Deletion = this opossite order so:
- Trigger delete
- Delete

rycks: hum ok but please have a look to my example, how could i get the id of the contact who is deleted ?

yes main object (fichinter) is not yet deleted but what about the contact ? who could i get the deleted contact to send him a notification, mail or whatelse about the fact he is no more linked to that intervention ?

hregis:
@rycks @eldy a good fix here https://github.com/Dolibarr/dolibarr/pull/26039

eldy:
Closed, replaced with regis PR.

rycks:
> @rycks @eldy a good fix here #26039

hmmm no no no ... please have a look on my comment on #26039 

with regis commit we have a trigger before delete, nice but we don't know what contact is deleted ! that could be worse in fact !

the only way to know - with trigger - what contact is deleted is to make a diff between "before delete" and "after delete" trigger or to invent a new way to pass that information into trigger, something like that :

```$this->contact_deleted_was = $contactid (if trigger is called after like it was on old code)```

or

```$this->contact_delete_will_be = $contactid (if trigger is called before if it's with new "regis" code)```

then the object passed into trigger will contain information about the deleted contact, i think

**Comment by hregis** [comment]
@eldy @rycks 

maybe missing send the contactid when call "delete_contact" in card action ?
and call "fetch_contact" to have $object->contact->id in the trigger ?

before:
`$result = $object->delete_contact(GETPOST('lineid', 'int'));`

after:
```
$contactid = (GETPOST('userid', 'int') ? GETPOST('userid', 'int') : GETPOST('contactid', 'int'));
$object->fetch_contact($contactid);
$result = $object->delete_contact(GETPOST('lineid', 'int'));
```

eldy:
For trigger, the way to pass information is to use the variable array
$object->context
You are free to use the key of your choice,
for example
$object->context['contact_id'] = ...

It is the same role than the $parameter areay for hooks, but for triggers

Regis, can you complete ?

hregis:
@eldy @rycks it's ok for you ? https://github.com/Dolibarr/dolibarr/pull/26043

rycks:
ok, but please cancel your commit Regis, because people could have already implemented solutions with trigger called after delete !!!

just add  $object->context['contact_deleted_id'] = ...

thanks

hregis:
@rycks no ! the trigger delete is already before the object delete

eldy:
> > @rycks @eldy a good fix here #26039
> 
> hmmm no no no ... please have a look on my comment on #26039 
> 
> with regis commit we have a trigger before delete, nice but we don't know what contact is deleted ! that could be worse in fact !
> 
> the only way to know - with trigger - what contact is deleted is to make a diff between "before delete" and "after delete" trigger or to invent a new way to pass that information into trigger, something like that :
> 
> ```$this->contact_deleted_was = $contactid (if trigger is called after like it was on old code)```
> 
> or
> 
> ```$this->contact_delete_will_be = $contactid (if trigger is called before if it's with new "regis" code)```
> 
> then the object passed into trigger will contain information about the deleted contact, i think

We don't need to invent a new way ro pass information to trigger. A way already exists. It is the $object->context[...]

eldy:
@rycks is right. Even if it is a bug fix, it is a structural change that change "the behaviour". 
Such cases should be merged into develop and existing version should be stated with "bug won't be fixed in this branch". 
Never mind, it seems this trigger is not so often used by external modules so risk is limited. I hope the mention in changelog will be enough for modules that rely on the default to know they have to change.
This is my fault. I should have merge this into develop. Sorry for troubles.

PR #26039, opened 2023-09-26
FIX wrong place of trigger delete

hregis:
@eldy same problem with v15, v16, v17, v18, develop


rycks:
hello,
i'm sorry but "my" problem is the same (cf #26038) how could i know what contact will be deleted !

really, i repeat it but we need a trigger BEFORE AND another one AFTER delete to be able to make a "diff" and then deduce what contact is to be deleted !!!

no ?

PR #20918, opened 2022-05-20
NEW: Save invoice payment with a AJAX request so the user don't send many request at the same time

kkhelifa-opendsi:
NEW: Save invoice payment with a AJAX request so the user don't send many request at the same time (when the user refresh the page when the action take too many time for some reason, ...)

PR #26041, opened 2023-09-27
NEW: Save invoice payment with a AJAX request so the user don't send many request at the same time

aspangaro:
Report an old feature with corrections suggest, see discussion on #20918


htdocs/compta/ajax/addpayment.php
```diff
@@ -0,0 +1,277 @@
...
+
+$langs->loadLangs(array('banks', 'bills', 'companies', 'multicurrency'));
+$error = 0;
+$invoice_id = 0;
+$errors_msg = array();
+
+/*
+ * View
+ */
+top_httphead();
```
eldy, line 44: The page return a json file so the top_httphead must be
top_httphead('application/json')

aspangaro: Thanks @eldy for your return

Corrected !

eldy:
The payment is a post of a page that need to switch onto another page.
Ajax is a solution to use when we want to remain on the same page.
Don't understand the goal of this. PR title say "So the user don't send many request at the same time "
Did you know the process the user do to make "many request at the same time" ?

aspangaro:
Hi @eldy ,

The only "known" process we have is that impatient users making quick clicks and encountering internet outages manage to insert two or more payments on one or more invoices.

Since we implemented this patch on Easya 2022 (base v14.0.5), i.e. 1.5 years ago, we have simply not had to deal with this recurring problem, despite the other protections previously put in place against this kind of action.

We have several testimonials available on request. It's implemented on all our easya 2022 instances.

Thanks in advance

eldy:
Are you sure this problem is because user is "impatient".
I confirm i had the trouble in the past.
But I think this is a bug of old versions of dolibarr (14 and may be 15), may be when user use the feature "Go back" on the browser.

My first idea is that the post was not correctly implemented and after the post, there was no "header("Location..." instruction to switch on a GET url (we should do this for action=add or update). So this means if we use the Go back, then the post was submitted a second time.
But this has been fixed in recent versions as I received no report about this trouble on recent version since a long time.

Another point is that there is a protection into code. See line after the comment
`"code to disable button to avoid double clic"`
this should be already enough to avoid double clic and is really better than trying to replace/emulate a POST request by using Ajax services.

So the PR may fix a trouble, but it introduces a very important technical debt. Currently i do not succeed in making a double submission by being impatient. 
Do you use module that affect the payment process ? Can you check js is not disabled and that the protection is working on your instance ?

PR #26043, opened 2023-09-27
FIX missing contact_id for the trigger

hregis:


rycks:
nice, but as i mentioned in #26039 please revert your initial commit (trigger before) because people could have already implement solutions and your commit could break a lot of things (move trigger from "after delete" to "before delete" is a big break, my initial purpose #26038 was to ADD a new trigger before, not replace after by before).

hregis:
@rycks sorry but it's a bug, the "delete" trigger must always be before deleting the object in case the trigger is in error we must not delete the object!

rycks:
> @rycks sorry but it's a bug, the "delete" trigger must always be before deleting the object in case the trigger is in error we must not delete the object!

okaaaaaaaaaay, then i cross my finger for all others developpers, please add a big warning about that into changelog ...

PR #26047, opened 2023-09-27
FIX #26033 Stock Movement associated with project doesn't show up in Project / Overview window 

### LINKED ISSUE #26033: [BUG] Stock Movement associated with project doesn't show up in Project / Overview window
### Bug

Stock movements associated with a Project do not display, although the code appears to exist for it already in htdocs/projet/element.php

Likely this functionality was broken at some point in the past, but since I find a need for it I will see if I can resolve this and create a pull request.

### Environment Version

18.0.1

### Environment OS

Linux

### Environment Web server

Linux

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

Expected: Stock movements associated with a project to appear in the Projects / Overview Tab, but it is missing

image

Enable Projects, Stocks, Advanced Stock Transfer

Create new Project
Create new stock movement associated with project (Using the Advanced Stock Transfer)
The project is visible when viewing the stock transfers page, but is not visible from the overview page of the Project.

### Attached files
There is a check for the module around Line 88
```
if (isModEnabled('stock')) {
	require_once DOL_DOCUMENT_ROOT.'/product/stock/class/entrepot.class.php';
	require_once DOL_DOCUMENT_ROOT.'/product/stock/class/mouvementstock.class.php';
}
```

Around line 615 appears to encode this functionality
```
	'stock_mouvement'=>array(
		'name'=>"MouvementStockAssociated",
		'title'=>"ListMouvementStockProject",
		'class'=>'MouvementStock',
		'table'=>'stock_mouvement',
		'datefieldname'=>'datem',
		'margin'=>'minus',
		'disableamount'=>0,
		'test'=>isModEnabled('stock') && $user->hasRight('stock', 'mouvement', 'lire') && !empty($conf->global->STOCK_MOVEMENT_INTO_PROJECT_OVERVIEW)
	),
	```

### PR DESCRIPTION
**Pull request by TheRealBeef**
# FIX #26033 Stock Movement associated with project doesn't show up in Project / Overview window 

Updates the old functionality of stock transfers associated with projects to properly display the "Advanced Stock Transfers" associated with it. 

Existing functionality was depreciated as it's not possible to associate normal stock movements with projects in any method that I tried, which is likely why this was hidden behind the undocumented global config variable that must be added to Other Setup.

I have only tested the changes in my own Dolibarr installation running 18.0.1 at this time.

### DIFF
--- a/htdocs/projet/class/project.class.php
+++ b/htdocs/projet/class/project.class.php
@@ -790,8 +790,8 @@ class Project extends CommonObject
 			$sql = "SELECT DISTINCT pt.rowid FROM ".MAIN_DB_PREFIX."projet_task as pt WHERE pt.fk_projet IN (".$this->db->sanitize($ids).")";
 		} elseif ($type == 'element_time') {	// Case we want to duplicate line foreach user
 			$sql = "SELECT DISTINCT pt.rowid, ptt.fk_user FROM ".MAIN_DB_PREFIX."projet_task as pt, ".MAIN_DB_PREFIX."element_time as ptt WHERE pt.rowid = ptt.fk_element AND ptt.elementtype = 'task' AND pt.fk_projet IN (".$this->db->sanitize($ids).")";
-		} elseif ($type == 'stock_mouvement') {
-			$sql = "SELECT ms.rowid, ms.fk_user_author as fk_user FROM ".MAIN_DB_PREFIX."stock_mouvement as ms, ".MAIN_DB_PREFIX."entrepot as e WHERE e.rowid = ms.fk_entrepot AND e.entity IN (".getEntity('stock').") AND ms.origintype = 'project' AND ms.fk_origin IN (".$this->db->sanitize($ids).") AND ms.type_mouvement = 1";
+		} elseif ($type == 'stocktransfer_stocktransfer') {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: htdocs/projet/class/project.class.php [hunk 1]
```diff
@@ -805,8 +805,8 @@ public function get_element_list($type, $tablename, $datefieldname = '', $date_s
 			$sql = "SELECT DISTINCT pt.rowid FROM ".MAIN_DB_PREFIX."projet_task as pt WHERE pt.fk_projet IN (".$this->db->sanitize($ids).")";
 		} elseif ($type == 'element_time') {	// Case we want to duplicate line foreach user
 			$sql = "SELECT DISTINCT pt.rowid, ptt.fk_user FROM ".MAIN_DB_PREFIX."projet_task as pt, ".MAIN_DB_PREFIX."element_time as ptt WHERE pt.rowid = ptt.fk_element AND ptt.elementtype = 'task' AND pt.fk_projet IN (".$this->db->sanitize($ids).")";
-		} elseif ($type == 'stock_mouvement') {
```
eldy, line 808: Why removing this case ? Can't you add the new case without removing this one ?

TheRealBeef: It's my understanding from looking around that the stock_mouvement class in the Projects is depreciated for the following reasons:

1. Already, the only way to display stock transfers (of either class) is to set the undocumented global config variable STOCK_MOVEMENT_INTO_PROJECT_OVERVIEW. Once it has been set, nothing from stock_mouvement class can displays anyway, because:
2. To the best of my knowledge the stock movements in stock_mouvement does not have any way to associate with project in a stock Dolibarr install. It appears that this functionality has been moved into only advanced stock transfer unless manually adding extrafields/complementary attributes to stock_mouvment.

If I am wrong (or at least we cannot be certain that I am correct), I will happily make the alterations just-in-case, but I'm not sure if it makes sense to leave the old case if its depreciated.

PR #26053, opened 2023-09-27
Update ticket.class.php - SQL Syntax Error

carmelchas:
Function "loadCacheCategoriesTickets" query had two WHERE clauses.  Change the second "WHERE" to "AND".

# FIX|Fix #[26052]

Dolibarr V.18 introduces a new error in the Ticket module.

dolibarr/htdocs/ticket/class/ticket.class.php LINE: 1287
function loadCacheCategoriesTickets

Duplicate "WHERE" changed to "AND".


carmelchas:
Please let me know if there is anything more I need to do.  This is my first time attempting to contribute.  I won't be surprised if I missed some protocol.

PR #26057, opened 2023-09-28
NEW intervention with service or product lines

lvessiller-opendsi:
NEW intervention with service or product lines

- add / edit or delete a product or service on intervention card
- can select a product / service or free line
- can set a quantity
- SQL changes are in this request (it will be separate later)

image

@eldy 
I would like your point of view about the need to have the possibility to select services on intervention card.
Do you think it will be better to have a new tab with consumption of products / services ?


eldy:
> NEW intervention with service or product lines
> 
> - add / edit or delete a product or service on intervention card
> - can select a product / service or free line
> - can set a quantity
> - SQL changes are in this request (it will be separate later)
> 
> image
> 
> @eldy 
> I would like your point of view about the need to have the possibility to select services on intervention card.
> Do you think it will be better to have a new tab with consumption of products / services ?
> 
> 

I think it is a good direction because it is more standard way of use compared to other modules. However 
- compatibily must be managed. I imagine that the current intervention lines will become a line of a non predefined service ?
- database must be modified in a separate request first. It must extends the current table for lines of intervention.

However, why do you need to modify the sql structue ?

BurksEngineering:
Thank you @lvessiller-opendsi for all of this work! I am excited to see the feature in a new version of Dolibarr. 

From glancing over the current PR there are a few things that might be missing and leave the users a bit confused when working with templates, which already had product and service support:
* When creating a new intervention from a template, it looks like [the 'addline' function will pass all lines as a non-predefined service](https://github.com/Easya-Solutions/dolibarr/blob/567d016f889592552cd1ee578b34c02ff69ba046/htdocs/fichinter/card-rec.php#L222) (old method) instead of passing along any product information that the template already has
* Similarly, when creating a template from an existing intervention, [the product/service details of each line will be previewed as a non-predefined service (old method)](https://github.com/Easya-Solutions/dolibarr/blob/567d016f889592552cd1ee578b34c02ff69ba046/htdocs/fichinter/card-rec.php#L415) instead of passing along any product information that the intervention already has
    * Weirdly, I think creating a template from an existing intervention will work, [because code to copy product info has been there for 3 years](https://github.com/Easya-Solutions/dolibarr/blob/567d016f889592552cd1ee578b34c02ff69ba046/htdocs/fichinter/class/fichinterrec.class.php#L225)? I don't understand this but I guess its good news!

I am certainly not a PHP developer so I could definitely be mistaken. Please let me know what I can do to help, because this addition looks to be extremely useful and highly sought after!

mmir-work:
And what about PDF generation? As of right now, using this change does not address the fact that none of the Properties of a Product are displayed in the PDF. All it does is just show up as another service, which is not exactly what this feature is useful for. 

Curiously enough, I already know that I need to edit the pdf_soleil.modules.php file, but I am not exactly sure which changes need to be applied to differentiate between products and services, since I am a bit rusty regarding PHP. 

I would love some help and/or suggestions :)

lvessiller-opendsi:
I'm not sure it's a good way to do like this and this work is too time-consuming for this moment.
I can't finish it alone.

PR #26059, opened 2023-09-28
Options email attach all files

leninrivas:
# NEW|New Attach all files email
Constant for attach all files email


htdocs/core/lib/files.lib.php
```diff
@@ -2523,8 +2523,14 @@ function dol_compress_dir($inputdir, $outputfile, $mode = "zip", $excludefiles =
  */
 function dol_most_recent_file($dir, $regexfilter = '', $excludefilter = array('(\.meta|_preview.*\.png)$', '^\.'), $nohook = false, $mode = '')
 {
+	global $conf;
 	$tmparray = dol_dir_list($dir, 'files', 0, $regexfilter, $excludefilter, 'date', SORT_DESC, $mode, $nohook);
-	return isset($tmparray[0])?$tmparray[0]:null;
+	//return isset($tmparray[0])?$tmparray[0]:null;
+	if(!empty($conf->global->MAIN_EMAIL_ATTACH_ALL_FILES)) {
```
eldy, line 2529: To avoid noise due to code style difference, can you enable the precommit git hook as explained here in your IDE:
https://wiki.dolibarr.org/index.php?title=Environment_and_development_tools#Install_Pre-commit_git_hook
If you do this, such errors reported by our CTI will be automatically fixed during your local commit...

leninrivas: Hi. Fixed.

htdocs/core/lib/files.lib.php
```diff
@@ -2840,8 +2840,14 @@ function dol_compress_dir($inputdir, $outputfile, $mode = "zip", $excludefiles =
  */
 function dol_most_recent_file($dir, $regexfilter = '', $excludefilter = array('(\.meta|_preview.*\.png)$', '^\.'), $nohook = 0, $mode = 0)
 {
+	global $conf;
 	$tmparray = dol_dir_list($dir, 'files', 0, $regexfilter, $excludefilter, 'date', SORT_DESC, $mode, $nohook);
-	return isset($tmparray[0]) ? $tmparray[0] : null;
+	//return isset($tmparray[0])?$tmparray[0]:null;
+	if (!empty($conf->global->MAIN_EMAIL_ATTACH_ALL_FILES)) {
+		return isset($tmparray) ? $tmparray : null;
```
eldy, line 2847: dol_most_recent_file may be used on a lot of different use case.
If you change the behaviour according to one option (for email attachment), it will change everywhere including the case where the behaviour should no change.
It seems you try to change the behaviour of a function by setting a global parameter to retrieve it inside the function. Such a need should be addressed by propagating paramater controled by the caller and notby setting a global setup option.

PR #26064, opened 2023-09-28
Update notes.tpl.php

randallmoraes:
Remove permision when a user is created from member can display private notes


eldy:
When you create a user from a membership file, you have choice to create user as an internal or external user:
image

If the user of the member must not see the private notes reserved for internal user, why don't you create the user as an external user ?

PR #26065, opened 2023-09-28
Update notes.tpl.php

randallmoraes:
Remove permision when a user is created from member can display private notes


eldy:
See comment on https://github.com/Dolibarr/dolibarr/pull/26064

PR #26068, opened 2023-09-29
FIX : Inventory lines saving with big inventory

Kazimir42:
When the inventory is displayed for the user there are an 'order by' on the initial query.

But when we save the inventory lines, the same request is played without the 'order by' so there is a lots of chance with a big inventory that your edited lines were not saved because Dolibarr doesn't find them with the query.


Kazimir42:
I think that fix the real issue behind : https://github.com/Dolibarr/dolibarr/commit/d21daa267c707878682b4f2ae2cc10f52b661b04

Kazimir42:
Should be backported to 18 as well ?

eldy:
I made a cherry-pick to get the fix on v18 too.

PR #26071, opened 2023-09-29
fix: wrong permission test on payment_sc

FHenry:
feature to test for payment_sc is `charges` and not `chargesociales` like it is done restrictedArea :
https://github.com/Dolibarr/dolibarr/blob/9300411ce3f5fe175732ba7a1e4ccd1febc8219d/htdocs/core/lib/security.lib.php#L491
https://github.com/Dolibarr/dolibarr/blob/9300411ce3f5fe175732ba7a1e4ccd1febc8219d/htdocs/core/lib/security.lib.php#L684

### DIFF
--- a/htdocs/core/lib/security.lib.php
+++ b/htdocs/core/lib/security.lib.php
@@ -822,7 +822,7 @@ function checkUserAccessToObject($user, array $featuresarray, $object = 0, $tabl
 		}
 
 		if ($feature == 'payment_sc') {
-			$feature = "chargesociales";
+			$feature = "charges";
 		}
 		$checkonentitydone = 0;
 


htdocs/core/lib/security.lib.php
```diff
@@ -822,7 +822,7 @@ function checkUserAccessToObject($user, array $featuresarray, $object = 0, $tabl
 		}
 
 		if ($feature == 'payment_sc') {
-			$feature = "chargesociales";
+			$feature = "charges";
```
eldy, line 825: Few lines after, we use "chargesociales". We find also chargesociales for ->element and at different places.

If goal is to fix a problem in permission, i think this must be managed inside the 
user->hasRight method :
Its goal is to be able special values that are not normalized.

PR #26077, opened 2023-09-29
NEW contact type on auto add contributor

BenjaminFlr:
For now, if an user creates a new ticket and is automatically added as a contributor, it's as an internal contributor even if it's an external user.

I propose to make the difference between internal and external for automatic add.


htdocs/ticket/class/ticket.class.php
```diff
@@ -570,7 +571,18 @@ public function create($user, $notrigger = 0)
 
 			if (!$error && !empty($conf->global->TICKET_ADD_AUTHOR_AS_CONTACT)) {
 				// add creator as contributor
-				if ($this->add_contact($user->id, 'CONTRIBUTOR', 'internal') < 0) {
+
+				// We first check the type of contact (internal or external)
```
eldy, line 575: I suggest 2 behaviour :
If TICKET_ADD_AUTHOR_AS_CONTACT = 1, we do old method, with 2, we do the one you suggest...

altairis-benjamin: I just added the check on constant == 2 to do what I suggested.

BenjaminFlr:
@eldy : I made the change but it looks like Travis is unhappy for some code not related to my PR.

hregis:
@altairis-benjamin a "var_dump" line 87 in "/htdocs/core/ajax/bankconciliate.php"

BenjaminFlr:
> @altairis-benjamin a "var_dump" line 87 in "/htdocs/core/ajax/bankconciliate.php"

Oui mais il ne vient pas de chez moi, pour le coup ;) C'est pour ça que je ne l'ai pas touché. Je le supprime et je push à nouveau @hregis ?

hregis:
@altairis-benjamin fait un pull, Laurent a corrigé  l'erreur

BenjaminFlr:
I think I'll make new changes as the activation of TICKET_ADD_AUTHOR_AS_CONTACT prevent public interface from working with inexistant users like a simple contact without any access to the dolibarr.

What do you think about that @eldy ?

htdocs/ticket/class/ticket.class.php
```diff
@@ -570,7 +571,18 @@ public function create($user, $notrigger = 0)
...
+				if (!empty($user->socid) && !empty($user->contact_id) && $conf->global->TICKET_ADD_AUTHOR_AS_CONTACT == 2) {
```
eldy, line 576: Can you replace $conf->global->TICKET_ADD_AUTHOR_AS_CONTACT
with
getDolGlobalInt('TICKET_ADD_AUTHOR_AS_CONTACT')

altairis-benjamin: I'll do but I also have to fix some things here so I'll do it for the next commit.

BenjaminFlr:
Hello @eldy,
I just fixed my PR with the use of `GetDolGlobalInt`.
I added a context for tickets created from the public interface and defined it as a requirement to the constant `TICKET_ADD_AUTHOR_AS_CONTACT`. Indeed, without that check, it will block the ticket creation from public interface.

I also added the ticket module by default to the `MAIN_MODULES_FOR_EXTERNAL` constant, because I assume that it's quite standard.

BenjaminFlr:
Little up here @eldy ? :)

PR #26083, opened 2023-09-30
FIX #25127

### LINKED ISSUE #25127: Unable to use cancel action on form with mandatory fields defined in llx_default_values table
### Bug

If you set a field as mandatory on the /admin/defaultvalues.php page, you can't use the CANCEL button on the form without filling out the mandatory field.

Issue is due to the fact that the CANCEL button is a SUBMIT button with name set to cancel. So you have to submit the form if you want to cancel your action. And because the field has now the property "required", the browser will not allow you to submit the form.

Property required is set with the following JS code
https://github.com/Dolibarr/dolibarr/blob/128b31feccd9edc7c1087c55b9eba9e1800b2dad/htdocs/core/lib/functions.lib.php#L9834

### Environment Version

16, 17, 18

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

Actual behavior : unable to use the cancel button without filling out user configured mandatory fields

Expected behavior : ability to use the cancel button without filling out the form

1. Define the field `accountancy_code_sell` as mandatory on url `product/card.php` on the admin page `/admin/defaultvalues.php`
2. Create a new product
3. Clic on cancel button

### Attached files

_No response_

hansemschnokeloch:
FIX #25127

If we use the cancel button there is no reason to have required fields. This PR removes all custom required rules added by jQuery.


htdocs/core/lib/functions.lib.php
```diff
@@ -9999,6 +9999,8 @@ function printCommonFooter($zone = 'private')
 								// Add 'field required' class on closest td for all input elements : input, textarea and select
 								print 'jQuery(":input[name=\'' . $paramkey . '\']").closest("tr").find("td:first").addClass("fieldrequired");' . "\n";
 							}
+							// If we submit the cancel button we remove the required attributes
+							print 'jQuery("input[name=\'cancel\']").click(function() {jQuery("input, textarea, select").each(function(){this.removeAttribute(\'required\');});});'."\n";
```
eldy, line 10003: To help debug and understand what happen on any js code executed by a binded function, we must include a log into console. Can you add
console.log("We cick on cancel button so removed all required attribute")
just after 
`click(function() {`

hansemschnokeloch: Log added as requested

PR #26102, opened 2023-10-03
NEW add recursive deletion option for child m os

samifilali:
# NEW
- Added a method that allows you to retrieve all child MOs recursively
- Added preview of the number of child MOs and a checkbox to delete them or not in the formconfirm


htdocs/mrp/class/mo.class.php
```diff
@@ -1659,6 +1659,43 @@ public function getMoChilds()
 		}
 	}
 
+	/**
+	 * Function used to return all child MOs recursively
+	 *
+	 * @return Mo[]|int[]  array of MOs if OK, -1 if KO
+	 */
+	public function getAllMoChilds()
```
eldy, line 1667: Sometimes, the database is corrupted with child that has a child that is parent.
This lead to infinite loop. For this reason on Dolibarr, when we make a recursing method, we always add a counter to counter the depth and stop when depth is 1000 (just to avoid infinite loop).
Can you introduce a parameter $depth = 0 into method, and increase +1 at each recursive call and add a test to stop when value > 1000
(1000 can be hardcoded because it should not happen in common cases).

htdocs/mrp/mo_card.php
```diff
@@ -176,6 +176,66 @@
...
+		}
+
+		$error = 0;
+		$deleteChilds = GETPOST('deletechilds', 'boolean');
+
+		if ($deleteChilds === 'on') {
+			$TMoChildren = $object->getAllMoChilds();
+
+			// Supprimer les objets enfants en utilisant une transaction séparée
+			$db->begin();
```
eldy, line 192: The delete of the child and the delete of the parent (line 212) must be included into the same transaction.

htdocs/mrp/class/mo.class.php
```diff
@@ -1659,6 +1659,45 @@ public function getMoChilds()
 		}
 	}
 
+	/**
+	 * Function used to return all child MOs recursively
+	 *
+	 * @return Mo[]|int[]  array of MOs if OK, -1 if KO
+	 */
+	public function getAllMoChilds($depth = 0)
```
eldy, line 1667: parameter $depth must be added into comments with
`@param int $depth   Depth for recursing loop count`

eldy:
Can you merge the conflict with another dev related to cancellation of MO to make reversed stock movement.

PR #26105, opened 2023-10-03
fix #26104 : remove global object

### LINKED ISSUE #26104: dolibarr 15 / multicompany bad link for showOnlineSignatureUrl if not $object
### Bug

showOnlineSignatureUrl call getOnlineSignatureUrl to make links

but in dolibarr-15.0 there was a strange commit #ade854a : to be multicompany compatible links now use a global $object in getOnlineSignatureUrl function.

the big problem is that we can use other name of variables than $object and there is no mention about that need when we use showOnlineSignatureUrl for example

### Environment Version

_No response_

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

rycks:
# FIX26104 : remove global $object

i hope that PR will be clean enough, the main problem is with global $object variable needed to make links compatible with multicompany

### DIFF
--- a/htdocs/core/lib/signature.lib.php
+++ b/htdocs/core/lib/signature.lib.php
@@ -20,11 +20,12 @@
 /**
  * Return string with full online Url to accept and sign a quote
  *
- * @param   string	$type		Type of URL ('proposal', ...)
- * @param	string	$ref		Ref of object
+ * @param   string			$type		Type of URL ('proposal', ...)
+ * @param	string			$ref		Ref of object
+ * @param   CommonObject  	$obj  		object (needed to make multicompany good links)
  * @return	string				Url string
  */
-function showOnlineSignatureUrl($type, $ref)
+function showOnlineSignatureUrl($type, $ref, CommonObject $obj = null)
 {
 	global $langs;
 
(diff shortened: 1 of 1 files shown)


rycks:
If that's ok, i do the job for dolibarr-16 -17 -18 and develop branches, please let me know if you are ok for PR on that branches too ?

eldy:
Yes,removing global $object is a good thing. Butmust be done into develop branch.

rycks:
hmmm, that's not my point of view, that is a bug introduced into 15.0.

branch 14.0 does not have that problem

so i purpose a fix for

- 15.0 (https://github.com/rycks/dolibarr/tree/15.0_fix_global_object_var_for_multicompany_online_sign)
- 16.0 (https://github.com/rycks/dolibarr/tree/16.0_fix_global_object_var_for_multicompany_online_sign)
- 17.0 (https://github.com/rycks/dolibarr/tree/17.0_fix_global_object_var_for_multicompany_online_sign)
- 18.0 (https://github.com/rycks/dolibarr/tree/18.0_fix_global_object_var_for_multicompany_online_sign)
- and develop https://github.com/rycks/dolibarr/tree/develop_fix_global_object_var_for_multicompany_online_sign

with a syslog "deprecated" in case of somebody use it ... then if you are ok i will do PR for each branch one by one because fix on 15.0 will not be automatically mergeable into 16 and others +

PR #26109, opened 2023-10-04
fix #26104 : remove global object (18.0)

### LINKED ISSUE #26104: dolibarr 15 / multicompany bad link for showOnlineSignatureUrl if not $object

rycks:
fix #26104 : remove global object

Same as previous but for 18.0 branch

### DIFF
--- a/htdocs/core/lib/signature.lib.php
+++ b/htdocs/core/lib/signature.lib.php
@@ -20,11 +20,12 @@
 /**
  * Return string with full online Url to accept and sign a quote
  *
- * @param   string	$type		Type of URL ('proposal', ...)
- * @param	string	$ref		Ref of object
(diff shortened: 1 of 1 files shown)


eldy:
Already merged from 15.0

PR #26113, opened 2023-10-04
php warning and field not used

defrance:
variable not defined, so we use another
question why keep this update code if we don't pass on cause "&& 0"


eldy:
I think the && 0  is because the action=classify is not yet implemented into this page.
So feature to edit project is disabled.

PR #26117, opened 2023-10-04
FIX: modification of complementary attributes in commercial proposals

am97:
# FIX: modification of complementary attributes in commercial proposals

Fixes the same problem as #26115


eldy:
Already pushed into v17, so will be merged automatically into v18 during the upward merge process

PR #26122, opened 2023-10-06
NEW - Add edit line on MO

atm-lucasmantegari:
Addition a button for editing product and service lines to be consumed when the MO has draft status.

### DIFF
--- a/htdocs/mrp/mo_production.php
+++ b/htdocs/mrp/mo_production.php
@@ -1141,7 +1141,7 @@ if ($object->id > 0 && (empty($action) || ($action != 'edit' && $action != 'crea
 
 						// Action Edit line
 						if ($object->status == Mo::STATUS_DRAFT) {
-							$href = $_SERVER["PHP_SELF"] . '?id=' . ((int) $object->id) . '&action=editline&token=' . newToken() . '&lineid=' . ((int) $line->id);
+							$href = $_SERVER["PHP_SELF"] . '?id=' . ((int)$object->id) . '&action=editline&token=' . newToken() . '&lineid=' . ((int)$line->id);
 							print '<td class="center">';
 							print '<a class="reposition" href="' . $href . '">';
 							print img_picto($langs->trans('TooltipEditAndRevertStockMovement'), 'edit');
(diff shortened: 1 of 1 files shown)


htdocs/mrp/mo_production.php
```diff
@@ -434,6 +434,15 @@
 			setEventMessages($object->error, $object->errors, 'errors');
 		}
 	}
+
+	if ($action == 'confirm_editline' && $permissiontoadd) {
+		$moline = new MoLine($db);
+		$moline->fetch(GETPOST('lineid', 'int'));
+		$moline->qty = GETPOST('qty_lineProduce', 'int');
```
atm-jpb, line 441: test retour du fetch ?

htdocs/mrp/mo_production.php
```diff
@@ -434,6 +434,16 @@
...
 		}
 	}
+
+	if ($action == 'confirm_editline' && $permissiontoadd) {
+		$moline = new MoLine($db);
+		$res = $moline->fetch(GETPOST('lineid', 'int'));
+		if ($result > 0){
+			$moline->qty = GETPOST('qty_lineProduce', 'int');
+			$moline->update($user);
+			header("Location: ".$_SERVER["PHP_SELF"].'?id='.$object->id);
```
eldy, line 444: After a "header(Location...", we must always have an "exit;"

htdocs/mrp/mo_production.php
```diff
@@ -890,116 +906,186 @@
...
-						print '<td class="center">';
-						print '<a class="reposition" href="'.$href.'">';
-						print img_picto($langs->trans('TooltipDeleteAndRevertStockMovement'), 'delete');
-						print '</a>';
-						print '</td>';
-					}
+						// Action Edit line
+						if ($object->status == Mo::STATUS_DRAFT) {
+							$href = $_SERVER["PHP_SELF"] . '?id=' . ((int) $object->id) . '&action=editline&token=' . newToken() . '&lineid=' . ((int) $line->id);
+							print '<td class="center">';
```
eldy, line 1070: You add a new column with <td></td> section here.
but later for exemple into the loop line 1090, you don't have the new column added meaning that there is a bad balance of nb of column between lines.
To match Dolibarr behaviour, it is recommended to consume only 1 line for the column of edit and delete. So can you add the action to edit into the same <td></td> than the delete (at end of line) ? This will solve the bug of wrong nb of columns and standardize code with other tables.

PR #26126, opened 2023-10-06
New Add js scanner expedition repartition

Hystepik:
Add the js scanner module to expedition repartition page


htdocs/expedition/ajax/searchfrombarcode.php
```diff
@@ -0,0 +1,135 @@
...
+$fk_product = GETPOST("fk_product", "int");
+$reelqty = GETPOST("reelqty", "int");
+$batch = GETPOST("batch", "int");
+$mode = GETPOST("mode", "aZ");
+
+$warehousefound = 0;
+$warehouseid = 0;
+$objectreturn = array();
+
+
```
eldy, line 57: This ajax component is missing security test.

PR #26128, opened 2023-10-06
FIX: Use of line->insert instead of line->create

noec764:
When the function create_common copied line for example if called by a createFromClone method, an error occured because no object was implementing ObjectLine::create() but ObjectLine::insert()


eldy:
For which object do you have this trouble ?

PR #26131, opened 2023-10-07
FIX avoid warning + CSRF

hregis:


htdocs/admin/ticket_public.php
```diff
@@ -299,7 +299,7 @@
 		if (!empty($conf->use_javascript_ajax)) {
 			print ajax_constantonoff('MAIN_SECURITY_ENABLECAPTCHA_TICKET');
 		} else {
-			if (empty($conf->global->MAIN_SECURITY_ENABLECAPTCHA_TICKET)) {
+			if (getDolGlobalInt('MAIN_SECURITY_ENABLECAPTCHA_TICKET')) {
```
eldy, line 302: if !  
i think...

hregis: @eldy yes thank you

PR #26143, opened 2023-10-09
FIX remove DISTINCT when MULTICOMPANY_TRANSVERSE_MODE is used

hregis:


hregis:
@eldy Do you prefer to integrate gradually with several PRs?

eldy:
Yes several PR is always better as long as the PR does not generate broken code

PR #26149, opened 2023-10-09
[FIX] Warnings in ModuleBuilder Configuration Page

Daoud-mohamed:
# FIX
Fix warnings in ModuleBuilder configuration page due to PHP Version.

### DIFF
--- a/htdocs/modulebuilder/admin/setup.php
+++ b/htdocs/modulebuilder/admin/setup.php
@@ -137,14 +137,14 @@ if ($conf->global->MAIN_FEATURES_LEVEL >= 2) {
 print '<tr class="oddeven">';
 print '<td>'.$langs->trans("UseSpecificEditorName").'</td>';
 print '<td>';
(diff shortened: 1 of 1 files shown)


htdocs/modulebuilder/admin/setup.php
```diff
@@ -137,59 +137,59 @@
 print '<tr class="oddeven">';
 print '<td>'.$langs->trans("UseSpecificEditorName").'</td>';
 print '<td>';
-print '<input type="text" name="MODULEBUILDER_SPECIFIC_EDITOR_NAME" value="'.$conf->global->MODULEBUILDER_SPECIFIC_EDITOR_NAME.'">';
+print '<input type="text" name="MODULEBUILDER_SPECIFIC_EDITOR_NAME" value="'.(isset($conf->global->MODULEBUILDER_SPECIFIC_EDITOR_NAME) ? $conf->global->MODULEBUILDER_SPECIFIC_EDITOR_NAME : "").'">';
```
eldy, line 140: This fix the warning. Great. This is the solution that is used at different places.
This is however the good direction for a better solution:
Use 
`...value="'.getDolGlobalString('MODULEBUILDER_SPECIFIC_EDITOR_NAME').'" ...`

Daoud-mohamed: I apologize for not having noticed that, even though it was mentioned in the documentation.
I've made the suggested changes using `getDolGlobalString()`  and have repushed the PR. Thank you!

PR #26151, opened 2023-10-09
NEW better way to manage main.inc.php inclusion in modules generated …

altairis-noe:
NEW better way to manage main.inc.php inclusion in modules generated with module builder.
Right now, in order to include `main.inc.php` in files which need it, we use a test on a pretty obscure path string comparison, which looks like this :
```php
// Load Dolibarr environment
$res = 0;
// Try main.inc.php into web root known defined into CONTEXT_DOCUMENT_ROOT (not always defined)
if (!$res && !empty($_SERVER["CONTEXT_DOCUMENT_ROOT"])) {
	$res = @include $_SERVER["CONTEXT_DOCUMENT_ROOT"]."/main.inc.php";
}
// Try main.inc.php into web root detected using web root calculated from SCRIPT_FILENAME
$tmp = empty($_SERVER['SCRIPT_FILENAME']) ? '' : $_SERVER['SCRIPT_FILENAME']; $tmp2 = realpath(__FILE__); $i = strlen($tmp) - 1; $j = strlen($tmp2) - 1;
while ($i > 0 && $j > 0 && isset($tmp[$i]) && isset($tmp2[$j]) && $tmp[$i] == $tmp2[$j]) {
	$i--; $j--;
}
if (!$res && $i > 0 && file_exists(substr($tmp, 0, ($i + 1))."/main.inc.php")) {
	$res = @include substr($tmp, 0, ($i + 1))."/main.inc.php";
}
if (!$res && $i > 0 && file_exists(substr($tmp, 0, ($i + 1))."/../main.inc.php")) {
	$res = @include substr($tmp, 0, ($i + 1))."/../main.inc.php";
}
// Try main.inc.php using relative path
if (!$res && file_exists("../../main.inc.php")) {
	$res = @include "../../main.inc.php";
}
if (!$res && file_exists("../../../main.inc.php")) {
	$res = @include "../../../main.inc.php";
}
if (!$res) {
	die("Include of main fails");
}
```
It really is not readable and it seems to me that it is more complicated than it should.
The replacement that I suggest will simply explore recursively the parents folders in the paths of `$_SERVER['SCRIPT_FILENAME']` until it finds `main.inc.php`.
It works event if the module’s folder is a symlink, and fails if there is no `main.inc.php` in the path on the server.
Thanks for any feedback.

### DIFF
--- a/htdocs/modulebuilder/template/admin/myobject_extrafields.php
+++ b/htdocs/modulebuilder/template/admin/myobject_extrafields.php
@@ -29,6 +29,10 @@
 
 // Load Dolibarr environment
 $res = 0;
+// Try main.inc.php into web root defined in DOCUMENT_ROOT, very frequently defined
+if (!$res && !empty($_SERVER["DOCUMENT_ROOT"])) {
+	$res = @include $_SERVER["DOCUMENT_ROOT"]."/main.inc.php";
+}
 // Try main.inc.php into web root known defined into CONTEXT_DOCUMENT_ROOT (not always defined)
 if (!$res && !empty($_SERVER["CONTEXT_DOCUMENT_ROOT"])) {
 	$res = @include $_SERVER["CONTEXT_DOCUMENT_ROOT"]."/main.inc.php";
--- a/htdocs/modulebuilder/template/admin/myobject_extrafields.php
+++ b/htdocs/modulebuilder/template/admin/myobject_extrafields.php
@@ -43,7 +43,7 @@ if (!$res) {
 			break;
 		}
 	}
-	$res = @include($dolipath."/main.inc.php");
+	$res = @include $dolipath."/main.inc.php";
 }
 if (!$res) {
 	die("Include of main fails");
--- a/htdocs/modulebuilder/template/admin/myobject_extrafields.php
+++ b/htdocs/modulebuilder/template/admin/myobject_extrafields.php
@@ -5,6 +5,7 @@
  * Copyright (C) 2012		Regis Houssin			<regis.houssin@inodbox.com>
  * Copyright (C) 2014		Florian Henry			<florian.henry@open-concept.pro>
  * Copyright (C) 2015		Jean-François Ferry		<jfefe@aternatik.fr>
+ * Copyright (C) 2023      Noé Cendrier         <noe.cendrier@altairis.fr>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -32,23 +33,17 @@ $res = 0;
 if (!$res && !empty($_SERVER["CONTEXT_DOCUMENT_ROOT"])) {
 	$res = @include $_SERVER["CONTEXT_DOCUMENT_ROOT"]."/main.inc.php";
 }
-// Try main.inc.php into web root detected using web root calculated from SCRIPT_FILENAME
-$tmp = empty($_SERVER['SCRIPT_FILENAME']) ? '' : $_SERVER['SCRIPT_FILENAME']; $tmp2 = realpath(__FILE__); $i = strlen($tmp) - 1; $j = strlen($tmp2) - 1;
-while ($i > 0 && $j > 0 && isset($tmp[$i]) && isset($tmp2[$j]) && $tmp[$i] == $tmp2[$j]) {
-	$i--; $j--;
-}
-if (!$res && $i > 0 && file_exists(substr($tmp, 0, ($i + 1))."/main.inc.php")) {
-	$res = @include substr($tmp, 0, ($i + 1))."/main.inc.php";
-}
-if (!$res && $i > 0 && file_exists(dirname(substr($tmp, 0, ($i + 1)))."/main.inc.php")) {
-	$res = @include dirname(substr($tmp, 0, ($i + 1)))."/main.inc.php";
-}
-// Try main.inc.php using relative path
-if (!$res && file_exists("../../main.inc.php")) {
-	$res = @include "../../main.inc.php";
-}
-if (!$res && file_exists("../../../main.inc.php")) {
-	$res = @include "../../../main.inc.php";
+// Try main.inc.php in all parent folders (cope with symlinks)
+if (!$res) {
+	$dolipath = dirname($_SERVER['SCRIPT_FILENAME']);
+	while (!file_exists($dolipath."/main.inc.php")) {
+		$abspath = $dolipath;
+		$dolipath = dirname($dolipath);
+		if ($abspath == $dolipath) { // cope with no main.inc.php all the way to filesystem root
+			break;
+		}
+	}
+	$res = @include($dolipath."/main.inc.php");
 }
 if (!$res) {
 	die("Include of main fails");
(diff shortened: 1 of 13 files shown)


hregis:
@eldy it's very more simple and works in all cases

htdocs/modulebuilder/template/admin/myobject_extrafields.php
```diff
@@ -28,27 +29,25 @@
 
 // Load Dolibarr environment
 $res = 0;
+// Try main.inc.php into web root defined in DOCUMENT_ROOT, very frequently defined
```
eldy, line 32: DOCUMENT_ROOT is not defined with the realpath of files and can differs from the real path that we need to have into the conf.php file for different purposes.
This is why we prefer CONTEX_DOCUMENT_ROOT.
Not as often defined but when defined, value is always good. For DOCUMENT_ROOT, value can be good or wrong and we have no way to make the difference.

htdocs/modulebuilder/template/admin/myobject_extrafields.php
```diff
@@ -28,27 +29,25 @@
...
-}
-// Try main.inc.php using relative path
-if (!$res && file_exists("../../main.inc.php")) {
-	$res = @include "../../main.inc.php";
-}
-if (!$res && file_exists("../../../main.inc.php")) {
-	$res = @include "../../../main.inc.php";
+// Try main.inc.php in all parent folders (cope with symlinks)
+if (!$res) {
+	$dolipath = dirname($_SERVER['SCRIPT_FILENAME']);
```
eldy, line 42: Script_filename is not always defined, so we must have a fallback using the .. ../.. etc syntax.
Also script_filename and document_root are root if the server but does not take into account the case we are using an apache alias to access different web sites on the same virtual host but isolated with an alias.
There is a lot of cases to support:
- case web server is used throw a proxy
- case web server is using a cgi wrapper like fpm instead of a apache prefork wrapper
- case web server is used with a load repartition system layer
- case web server run with virtualhost on a different filesystem than the main apache biqnaries, when using symbolic links, when using hard links.
- case using some php security modules that hide some server variables
- case web server uses a special setup of some exotics panel managers like cpanel, plesk, ...
- case web server is using apache "directory alias" instead of virtual host to isolate servers...
And any combination of all of this !

Current handler to find the main.inc.php contains a lot of lines, yes, but is has 20 years of experience and has proved it works in all the situation (probably nearly hundred of different combination of situations).
We already experiment use of document_root and we know we must NOT use it.
Changing the main loader will surely break some cases, even if they are not frequent, when current is robust. It does not worth the gain.
Only the mutualization of the search using the .. then ../.. then ../../.. etc into a loop can be done at no risk.

hregis:
@eldy 
for me DOCUMENT_ROOT is always ok with Apache, Nginx and MAMP (Mac OS)
CONTEXT_DOCUMENT_ROOT is wrong with MAMP !

eldy:
> @eldy 
> for me DOCUMENT_ROOT is always ok with Apache, Nginx and MAMP (Mac OS)
> CONTEXT_DOCUMENT_ROOT is wrong with MAMP ! 
>  


The experience and hundred of use case tested in Dolibarr life has proved that DOCUMENT_ROOT can't be used. Don't forget the hundred of other uses cases i reported in previous message. You will see sometimes
DOCUMENT_ROOT is not available, but the main problem is that it is also wrongly defined, with no way to know it breaking completely important features !
CONTEXT_DOCUMENT_ROOT is often not defined but when defined, it is good. And we need reliable method. If a reliable method is not available, we fallback on another one. Only method that does not report bad value can be used. Not the most often available.
CONTEXT_DOCUMENT_ROOT has been introduced by Apache project in apache 2.4 because DOCUMENT_ROOT was not reliable to get the web app root dir, so we can use it. Other fallback tests are here to support when it is not available and it is working as expected.

altairis-noe:
I did not propose to ignore CONTEXT_DOCUMENT_ROOT but to use DOCUMENT_ROOT in complement.

I do understand the use of ../ in case $_SERVER is somehow broken, and will put it back, but I wish you could tell me how the string comparison is more effective than to recursively check the script path.

I have to admit that the argument of “20 years of experience” sound a little too much ”we’ve always done like that” to convince me…

eldy:
> I did not propose to ignore CONTEXT_DOCUMENT_ROOT but to use DOCUMENT_ROOT in complement.
> 
> I do understand the use of ../ in case $_SERVER is somehow broken, and will put it back, but I wish you could tell me how the string comparison is more effective than to recursively check the script path.
> 
> I have to admit that the argument of “20 years of experience” sound a little too much ”we’ve always done like that” to convince me…

I didn't said "we have always do like that", I say "we already try like you suggested in the past, by using DOCUMENT_ROOT, and the result was it is not working in a lot of use cases (see my previous message to see some examples of use cases that were tested). We **can't use DOCUMENT_ROOT** as a complement because the value is sometimes wrong in some of the advanced use cases and we got too many problem with this method when we did it. Problem is not that it is define or not but that it is wrongly defined, when we no experience of broken cases with current wrapper. 
30 lines of code working in 99.99% of use cases is better than 10 lines of code working in 99%.

altairis-noe:
I strongly disagree !
The purpose of this PR was NOT tu use DOCUMENT_ROOT. It was to replace the unreadable string comparison shenanigan, on which you did not provide any comment.
I added the DOCUMENT_ROOT element because @hregis told me it worked in his development environment and CONTEXT_DOCUMENT_ROOT did not.
I certainly could have done without it.

ksar-ksar:
Hello,

There is a need for sure, you can see another approach : https://www.dolibarr.org/forum/t/new-script-to-load-main-inc-php-on-third-party-modules-avoiding-include-of-main-fails/24296

caos30:
> Hello,
> 
> There is a need for sure, you can see another approach : https://www.dolibarr.org/forum/t/new-script-to-load-main-inc-php-on-third-party-modules-avoiding-include-of-main-fails/24296

Thank you, @ksar-ksar for mentioning my solution. I believe that my solution addresses:

A) The main concern of @altairis-noe regarding an "ugly & long" block of lines needed to load `main.inc.php`

But also, and I think more importantly:

B) It delegates this loading to a SINGLE php script focused on this task, which is easily "updatable" in our modules over time.

C) It offers users the possibility to directly provide the path of the `main.inc.php` file the FIRST time, after which it remembers the absolute path (saving it in a text file in the `/custom` directory), eliminating the need to search for it EACH TIME.

Furthermore:

D) This solution is open to evolution in the future, as we are all aware that the world of "cloud storage" and VMs is continuously evolving. In this regard, I must say to @eldy that the solutions we have utilized until now will probably not cover all new cases. Therefore, we need a system that is capable of EVOLVING without complicating the lives of module developers, Dolibarr distributors, or end-users.

Let me share the repository on Github of my proposed solution:

https://github.com/caos30/dolibarr_load_main_inc_php

altairis-noe:
Thank you a lot @ksar-ksar and @caos30, I really like this approach !

caos30:
Thanks to you, @altairis-noe for your support and enhancement of the BACKWARDS technique to search the `main.inc.php`, using an smart and "endless" recurring loop. Yeah!

That was my idea when publishing the script on Github: that together we can make it stronger and more powerful. And it is very easily updatable for our modules in Dolistore... you just have to replace that single script with its latest version. 😁

PR #26154, opened 2023-10-09
Feat: Add refactoring user permission

d3v2a:
# FIX Change all access to $user->rights to function $user->hasRight()

Changes were made with Rector


eldy:
Can you add a README into the firectory dev/tools/rector, to explain how to launch the correction program.
So we will be able to run it on small part first before on a large number of files...

The tool enhancements and the result of running the tool must be in 2 different PR.

d3v2a:
[New PR](https://github.com/Dolibarr/dolibarr/pull/26162) created

PR #26157, opened 2023-10-09
NEW : Adding default recipients to email forms

samifilali:
# NEW
On email sending forms, recipients are pre-filled with external contacts associated with the document in the document's "Contacts / Addresses" tab


samifilali:
Does this way of doing things suit you? if this is the case I can adapt it for all email sending forms

eldy (commented):
Here you decide to always autofill all the contact of the doculent as recipients of email by default ?
This is a very dangerous feature. Don't forget thatcemail can also be used to send email to internal only recipients, or unique recipients like a laywer, the bookkeeper, ...

eldy:
CI conflict since a too long time. Solving not available. So closed

PR #26164, opened 2023-10-09
NEW - Add culum Technical ID in liste det order

BB2A-Anthony:
# NEW|New [*Add culum Technical ID in liste det order*]
[*Long description*]


edupulpillo:
The option is not present yet in v 19.0.0

BB2A-Anthony:
Active MAIN_FEATURES_LEVEL = 2

PR #26168, opened 2023-10-10
FIX - Search product category on list det order

BB2A-Anthony:
# FIX|Fix #[*FIX - Search product category on list det order*]
[*Long description*]


htdocs/commande/list_det.php
```diff
@@ -407,7 +407,7 @@
 	$sql .= ' LEFT JOIN '.MAIN_DB_PREFIX.'commandedet as pd ON c.rowid=pd.fk_commande';
 }
 if ($search_product_category > 0) {
-	$sql .= ' LEFT JOIN '.MAIN_DB_PREFIX.'categorie_product as cp ON cp.fk_product=pd.fk_product';
+	$sql .= ' LEFT JOIN '.MAIN_DB_PREFIX.'categorie_product as cp ON cp.fk_product=cdet.fk_product';
```
eldy, line 410: Filter on categories must no more use the LEFT JOIN solution. It must use a solution without adding tables in the SELECT nor into the FROM.
All conditions must be into the WHERE. See example into projet/list.php after comment
// Search for tag/category ($searchCategoryProjectList is an array of ID)
Can you investigate to make changes into this direction (this is a goal to modify everywhere in dolibarr this way).

BB2A-Anthony: @eldy it's done

PR #26170, opened 2023-10-10
FIX - Clean label product on list det order

BB2A-Anthony:
# FIX|Fix #[*FIX - Clean label product on list det order*]
[*Long description*]

If free ligne print description

Result if dol_escape_htmltag

Capture d'écran 2023-10-09 202844

Result if dol_string_nohtmltag

Capture d'écran 2023-10-09 203025


htdocs/commande/list_det.php
```diff
@@ -1543,7 +1543,7 @@
 		if (!empty($arrayfields['pr.desc']['checked'])) {
 			// print '<td class="nowrap tdoverflowmax200">'.$obj->description.'</td>';
 			!empty($obj->product_label) ? $labelproduct = $obj->product_label : $labelproduct = $obj->description;
-			print '<td class="nowrap tdoverflowmax200">'.dol_escape_htmltag($labelproduct).'</td>';
+			print '<td class="nowrap tdoverflowmax200">'.dol_string_nohtmltag($labelproduct).'</td>';
```
eldy, line 1546: It seems we introduce a description (so rich html content) into label. This is why the dol_escape_htmltag() will generate a bad output.
According to experimentation done on tons of use case and reported into the comment of method dol_escape_htmltag(), the correct way to output when we are dealing with "rich text" is
`dol_escape_htmltag(dol_htmlwithnojs(dol_string_onlythesehtmltags(dol_htmlentitiesbr(  ....    ), 1, 1, 1), 1, 1)`
Just for the experience, what is the screenshot if using this "long" but "theoricaly always good sentence ? Can you confirm that the comment into dol_escape_htmltag is good ?

PR #26173, opened 2023-10-10
NEW Add a button to create a product or a service from an order or an invoice

FlorentPoinsaut:
# NEW Add a button to create a product or a service from an order or an invoice

Capture d’écran du 2023-10-10 13-44-51
Capture d’écran du 2023-10-10 13-45-07

### DIFF
--- a/htdocs/core/tpl/objectline_create.tpl.php
+++ b/htdocs/core/tpl/objectline_create.tpl.php
@@ -316,14 +316,15 @@ if ($nolinesbefore) {
 				}
 			}
 
-			$facid = GETPOST('facid', 'int');
+			$parentId = GETPOST('parentId', 'int');
+
 			echo '<div id="dropdownAddProductAndService" class="dropdown inline-block">';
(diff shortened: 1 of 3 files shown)


htdocs/core/tpl/objectline_create.tpl.php
```diff
@@ -315,6 +315,31 @@
 					<?php
 				}
 			}
+
+			$parentId = GETPOST('parentId', 'int');
+
+			echo '<div id="dropdownAddProductAndService" class="dropdown inline-block">';
+			echo '<a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" id="dropdownAddProductAndServiceLink" aria-haspopup="true" aria-expanded="false">';
+	        echo '<span class="fa fa-plus-circle valignmiddle paddingleft"> '.$langs->trans("Add").'</span>';
```
eldy, line 323: The picto may be enough (no need for the "Add" label)

PR #26175, opened 2023-10-10
fix: result checks for PHP 8.1

am97:
# fix: result checks for PHP 8.1

## Problem

In our Dolibarr 16.0.5 instance, overwriting of translation strings isn't working. For example, I go to `/admin/translation.php?mode=overwrite&transkey=Notifications` and I fill in the fields:

Screenshot 2023-10-10 at 19-25-22 Setup

Then, when I click `ADD`, the page is reloaded, and I still see the button, but the overwrite isn't added.

After debugging the problem, I found out that the transaction was rollback. Our instance uses PHP 8.1, which is listed as supported for 16.0.5 in https://wiki.dolibarr.org/index.php/Versions, however, some `if` conditions no longer work as intended. In PHP 8.0 and older, `pg_query` returns a `resource`, so the following works

```php
if ($result > 0) {
	$db->commit();
} else {
	$db->rollback();
}
``` 

As of PHP 8.1, `pg_query` returns a [PgSql\Result](https://www.php.net/manual/en/class.pgsql-result.php) object. The check no longer works, we hit the `else` case even with a successful query.

## Solution

This PR fixes the checks in the `htdocs/admin/translation.php` file. It fixes the issue in our instance. I verified that the `if ($result)` condition works correctly in the oldest PHP version supported by Dolibarr 16.0.5 (PHP 5.6).

### DIFF
--- a/htdocs/admin/translation.php
+++ b/htdocs/admin/translation.php
@@ -116,7 +116,7 @@ if ($action == 'update') {
 
 		$sql = "UPDATE ".MAIN_DB_PREFIX."overwrite_trans set transkey = '".$db->escape($transkey)."', transvalue = '".$db->escape($transvalue)."' WHERE rowid = ".((int) GETPOST('rowid', 'int'));
 		$result = $db->query($sql);
-		if ($result > 0) {
+		if ($result) {
 			$db->commit();
 			setEventMessages($langs->trans("RecordSaved"), null, 'mesgs');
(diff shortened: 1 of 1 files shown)


am97:
There are also some bad checks in branches 17, 18 and develop (but less than in branch 16). Should I submit a fix there ? Or the upward merge process will be sufficient ?

PR #26177, opened 2023-10-10
NEW - Row in list higher height

BB2A-Anthony:
# NEW|New [*Row in list higher height*]
[*Long description*]

To have better readability and give modernism
image


eldy:
Can you show a screen shot to see what is the diff on ?

BB2A-Anthony:
> Can you show a screen shot to see what is the diff on ?

image

eldy:
This is interesting for the first "wow-effect"  (Very important to adopt dolibarr), but may have bad effect for the day-to-day usage by reducing the number of lines visible on screen.
I think we can introduce an option into the display page to choose "Table line spacing" between "compact" or "large" ?
We can use the large by default (on first install for the wow-effect, but we allow end user to retrieve a compact mode for day to day usage)

BB2A-Anthony:
> This is interesting for the first "wow-effect" (Very important to adopt dolibarr), but may have bad effect for the day-to-day usage by reducing the number of lines visible on screen. I think we can introduce an option into the display page to choose "Table line spacing" between "compact" or "large" ? We can use the large by default (on first install but we allow end user to retrieve a compact mode for day to day usage)

@eldy it's done

PR #26180, opened 2023-10-11
FIX: modification of complementary attributes in invoices

am97:
# FIX: modification of complementary attributes in invoices

Similar to #26115 , this PR fixes the modification of complementary attributes in invoices, which was leading to a blank page with the following error:

```
PHP Fatal error:  Uncaught Exception: Serialization of PgSql\Connection is not allowed in /var/www/html/core/lib/functions.lib.php:1168
Stack trace:
#0 /var/www/html/core/lib/functions.lib.php(1168): serialize()
#1 /var/www/html/compta/facture/card.php(2798): dol_clone()
#2 {main}
```

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -2795,7 +2795,7 @@ if (empty($reshook)) {
 
 
 	if ($action == 'update_extras') {
-		$object->oldcopy = dol_clone($object);
+		$object->oldcopy = dol_clone($object, 2);
 
 		// Fill array 'array_options' with data from add form
 		$ret = $extrafields->setOptionalsFromPost(null, $object, GETPOST('attribute', 'restricthtml'));


eldy:
This looks strange because into the dol_clone method when using parameter zero, we have
`if (isset($object->db) && isset($object->db->db) && is_object($object->db->db) && get_class($object->db->db) == 'PgSql\Connection') {
			$tmpsavdb = $object->db;
			unset($object->db);		// Such property can not be serialized with pgsl (when object->db->db = 'PgSql\Connection')
`
So we should not experience this trouble. Can you check why we don't enter into this test ?

hregis:
@am97 what is the version of your php ?

am97:
> Can you check why we don't enter into this test ?

I made a dump of the object passed to `dol_clone`, there is another `PgSql\Connection` object (`lines[0]->db->db`). Here is the dump (I replaced some data with `**redacted**`): 

[php_object_dump.txt](https://github.com/Dolibarr/dolibarr/files/12868285/php_object_dump.txt)

> @am97 what is the version of your php ?

PHP 8.1.20 (we are using a Docker image based on `tuxgasy/dolibarr:16.0.5`)

PR #26181, opened 2023-10-11
FIX : HTML in ODT templates

atm-adrien:
## FIX : HTML in ODT templates

There is currently no interpretation of HTML in ODT type document templates.

Here is what we have in entry : 
`<p><strong><u>DESCRIPTION DES PRODUITS</u></strong><br />
Ceci est un produit &amp; un service<br />
Contr&ocirc;le des stock de ce produit&nbsp;<br />
V&eacute;rification des dispositifs d&#39;installation pour ce produit&nbsp;<br />
Vidange des filtres &agrave; carburant suppl&eacute;mentaire pour ce produit&nbsp;<br />
N&deg; de ce produit</p>`

We have the same thing written in our ODT document

### DIFF
--- a/htdocs/includes/odtphp/odf.php
+++ b/htdocs/includes/odtphp/odf.php
@@ -161,7 +161,8 @@ class Odf
 	 */
 	public function convertVarToOdf($value, $encode = true, $charset = 'ISO-8859')
 	{
-		$value = $encode ? htmlspecialchars($value) : $value;
+        $value = dol_html_entity_decode($value, ENT_COMPAT);
+//		$value = $encode ? htmlspecialchars($value) : $value;
 		$value = ($charset == 'ISO-8859') ? utf8_encode($value) : $value;
 		$convertedValue = $value;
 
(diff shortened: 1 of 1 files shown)


htdocs/includes/odtphp/odf.php
```diff
@@ -161,7 +161,8 @@ public function setVars($key, $value, $encode = true, $charset = 'ISO-8859')
 	 */
 	public function convertVarToOdf($value, $encode = true, $charset = 'ISO-8859')
 	{
-		$value = $encode ? htmlspecialchars($value) : $value;
+        $value = dol_html_entity_decode($value, ENT_COMPAT);
```
eldy, line 164: Can you confirm this:  This means the parameter $encode will not be used anymore ?

eldy:
It seems a different fix were done in v18.
Can you check to backport done into the v18 instead of this version for v17 ?

atm-adrien:
I copied what was done in V18, this does not seem to pose a compatibility problem

I also modified the regexp a little (it did not target all of the tags)

htdocs/includes/odtphp/odf.php
```diff
@@ -18,980 +18,1003 @@ class OdfException extends Exception
...
+                $handlein = popen($command, 'r');
+                while (!feof($handlein)) {
+                    $read = fgets($handlein);
+                    fwrite($handle, $read);
+                    $output_arr[]=$read;
+                }
+                pclose($handlein);
+                fclose($handle);
+            }
+            dolChmod($outputfile);
```
eldy, line 906: With v17, this line of code will hang. Keep v17 code as this has no relation with the bugfix.

atm-adrien: Done

PR #26184, opened 2023-10-11
NEW : Allow to have products not managed in stocks

thomas-Ngr:
# NEW : Allow to have products not managed in stocks

This PR:
- adds a field "Do not manage product in stocks" on product card (editable on update, not on create)
- hides the "stocks" tabs if this option is set
- does not manage stock movements for a product if this option is set.


thomas-Ngr:
Follow PR, invert naming of the field : https://github.com/Dolibarr/dolibarr/pull/26200

PR #26190, opened 2023-10-11
NEW : Allow to have products not managed in stocks - SQL Part

aspangaro:
Structure part of #26184


htdocs/install/mysql/migration/18.0.0-19.0.0.sql
```diff
@@ -123,3 +123,5 @@ ALTER TABLE llx_links ADD UNIQUE INDEX uk_links (objectid, objecttype,label);
 
 ALTER TABLE llx_c_invoice_subtype MODIFY COLUMN entity integer DEFAULT 1 NOT NULL;
 
+-- Product/service not managed in stock
+ALTER TABLE llx_product ADD COLUMN not_managed_in_stock integer DEFAULT NULL;
```
eldy, line 127: Using negation in name of var make bad readability of code.
To match other ERP, i suggest to name this field  stockable_product with value 1by default (because in dolibarr, default is to manage stock, at least when module stock is on).

So i suggest this to add this into migration file:
ALTER TABLE llx_product ADD COLUMN stockable_product integer DEFAULT 1 NOT NULL;
UPDATE llx_product set stockable_product = 0 WHERE type = 1;

aspangaro: Thanks @eldy  for your proposal. I'm Agree. Done

thomas-Ngr:
code part : #26200

PR #26193, opened 2023-10-12
Show product description

tgsw:
# NEW shows the product-description when hover product-ref in the list view of products
shows the product-description (only if not empty) when hover product-ref in the list view of products

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -5080,6 +5080,9 @@ class Product extends CommonObject
 		if (!empty($this->label)) {
 			$datas['label']= '<br><b>'.$langs->trans('ProductLabel').':</b> '.$this->label;
 		}
+		if (!empty($this->description)) {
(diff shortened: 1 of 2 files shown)


htdocs/product/class/product.class.php
```diff
@@ -5080,6 +5080,9 @@ public function getTooltipContentArray($params)
 		if (!empty($this->label)) {
 			$datas['label']= '<br><b>'.$langs->trans('ProductLabel').':</b> '.$this->label;
 		}
+		if (!empty($this->description)) {
+			$datas['description']= '<br><b>'.$langs->trans('ProductDescription').':</b> '.$this->description;
```
eldy, line 5084: What happen if product description is several page of HTML content ?
I suggest to add dolGetFirstLineOfText() around text to keep only the 5 first lines...

tgsw: Yes, that is indeed helpful.
I will modify that.

PR #26200, opened 2023-10-12

thomas-Ngr:

This PR:

Discussion started on #26184

SQL Part : #26190


htdocs/expedition/card.php
```diff
@@ -2300,6 +2339,7 @@
 				$product_static->surface_units = $lines[$i]->surface_units;
 				$product_static->volume = $lines[$i]->volume;
 				$product_static->volume_units = $lines[$i]->volume_units;
+				$product_static->stockable_product = $lines[$i]->stockable_product;
```
josett225, line 2342: Hi @thomas-Ngr 
I believe this line does not work as $lines[$i]->stockable_product does not exist.

htdocs/expedition/card.php
```diff
@@ -2468,9 +2508,10 @@
 				// Warehouse source
 				if (isModEnabled('stock')) {
 					print '<td class="linecolwarehousesource tdoverflowmax200">';
+          
 					if ($lines[$i]->product_type == Product::TYPE_SERVICE && !empty($conf->global->SHIPMENT_SUPPORTS_SERVICES)) {
 						print '<span class="opacitymedium">('.$langs->trans("Service").')</span>';
-					} elseif ($lines[$i]->entrepot_id > 0) {
+					} elseif ($lines[$i]->entrepot_id > 0 && $lines[$i]->product->stockable_product == Product::ENABLED_STOCK) {
```
josett225, line 2514: Hi @thomas-Ngr 
Same here; this $lines[$i]->product->stockable_product does not exist in the object $lines[$i]

PR #26209, opened 2023-10-13
NEW show VAT free amount on payment input close #26208

### LINKED ISSUE #26208: Add VAT free & VAT amounts to payment view
### Feature Request

On the payment input view, in the table that lists various payment information, it could be useful, in some cases, to also have VAT free (HT) and VAT (TVA) amounts. These 2 new columns would be just before the Amount including VAT column as in the suggested implementation below.
 
new-show-vat-free-amount-payment-input

In order to keep importance on the full amount on which we are checking and entering a payment for, we could use `font-weight: bold` on the _Amount (inc. tax)_ value.

**Remarks:** 
1. Nowadays lots of screens are at least HD, I believe we have enough space to easily add these two columns to the view without affecting readability.
2. Importance could also be put on the _"Remaining amount to take"_ as this is what we would probably be comparing the full amount with.

### Use case

In some of the views of various accounting programs, the importance is sometimes put on VAT free amounts. For the person in charge of manually entering payments into Dolibarr it can be useful to also see VAT free and VAT amounts in order to pinpoint the correct payment information. The objectif is to reduce human error and improve quality of life for the user.

### Suggested implementation

Cf. screenshot above

### Suggested steps

_No response_

W1W1-M:
# Show VAT free amount and VAT amount on user payment input view close #26208


eldy:
This generate a scrollbar on medium screen.
Why not showing information like we do on invoice like this :
image

W1W1-M:
> This generate a scrollbar on medium screen. Why not showing information like we do on invoice like this : image

I'll try implementing it in this way to see. Thanks for pointing this out.

hansemschnokeloch:
Hi,

if this PR fixes the issue that you mentionned in the title, can you please change the title of the PR ?
It should begin with FIX followed by the number of the issue. After that the relevant issue will be closed automatically if your PR is merged.

Thx

W1W1-M:
> Hi,
> 
> if this PR fixes the issue that you mentionned in the title, can you please change the title of the PR ? It should begin with FIX followed by the number of the issue. After that the relevant issue will be closed automatically if your PR is merged.
> 
> Thx

Thanks for tip. For reference: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

W1W1-M:
@eldy I've re-implemented this feature as recommended, I've also added the tooltip on the equivalent multicurrency amount. I hesitated in making a method in `paiement.class.php` as to not add too many lines to `paiement.php` but in this context the the payment hasn't been entered yet their is no `paiement` object to work with. At best in could be a static method. 

Also I'm not 100% sure if I've implemented the tooltip correctly, what are the best practices for tooltips in Dolibarr ?

Thanks for your input

PR #26217, opened 2023-10-13
FIX: Fix create shipping with product who have negative stocks on warehouse but the negative stock transfert is allowed

kkhelifa-opendsi:
Fix create shipping with product who have negative stocks on warehouse but the negative stock transfert is allowed


kkhelifa-opendsi:
@eldy did you see it ?

lvessiller-opendsi:
@eldy did you see it ?

PR #26223, opened 2023-10-15
FIX #26100 - Ticket - On edit, list of closed project must be excluded

### LINKED ISSUE #26100: Is not allowed to select closed projects at tickets creation
### Bug

Is not allowed to select closed projects in Tickets.
Even when a ticket is closed, remains under garantee, and new support tickets can arise over an alreaady closed project.

### Environment Version

ALl

### Environment OS

All

### Environment Web server

All

### Environment PHP

All

### Environment Database

All

### Environment URL(s)

https://myserver.com/ticket/card.php?action=create&idmenu=6596&mainmenu=ticket&leftmenu=

Now all the closed projects appears in grey color and is not possible to select it when a ticket is created.
This force the user to reopen the project, then assign the project to the ticket, and close again the project.

_No response_

### Attached files

_No response_

aspangaro:
Difference between creation and display/modification phases

On ticket create mode, closed project are not available.
On ticket view mode, closed project are available.

Propose to access to closed projects in ticket create mode

### DIFF
--- a/htdocs/core/class/html.formticket.class.php
+++ b/htdocs/core/class/html.formticket.class.php
@@ -437,7 +437,7 @@ class FormTicket
 			if (!empty($conf->projet->enabled) && !$this->ispublic) {
 				$formproject = new FormProjets($this->db);
 				print '<tr><td><label for="project"><span class="">'.$langs->trans("Project").'</span></label></td><td>';
-				print img_picto('', 'project').$formproject->select_projects(-1, GETPOST('projectid', 'int'), 'projectid', 0, 0, 1, 1, 0, 0, 0, '', 1, 0, 'maxwidth500');
+				print img_picto('', 'project').$formproject->select_projects(-1, GETPOST('projectid', 'int'), 'projectid', 0, 0, 1, 0, 0, 0, 0, '', 1, 0, 'maxwidth500');
 				print '</td></tr>';
 			}
 		}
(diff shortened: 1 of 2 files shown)


eldy:
The only goal of closing a project is to be sure they are not used anymore into other created object like tickets (closing project = "i don't want this project to be included into new object").
If we include then in creation of tickets, what is goal of closing a project ?

For edit, project must appear into list if it is already set, but we should not be able to modify to switch on a closed project (except if we open the project for that).

aspangaro:
Thanks @eldy 

In this case, compared to my first PR, I propose to invert how it works because for the moment, when you edit a project on ticket, you see closed projects:

image

I push a new PR for:
No access to close projects on creation
No access to close projects on edit

eldy:
I confirm that closed project must even not appear into combo list. Don't forget that nb of project increase with time. But list of values into combo must remain with a qty that is stable (only active project)

Can you confirm that if we edit an object (ticket) and the project is a closed project we can still keep the current value. Only changing for a closed project is not allowed. Can you confirm this is what you did ?

edupulpillo:
The closed proyects also should be availble when a ticket is created.
After a project is closed, a garantee about it can be stablished, and then still maintenance tickets can arise.
So maybe a special setup in the ticket module: Allow to select closed projects YES/NO, can be the best solution.

At the moment if a customer have garantee about a project already closed, when a ticket is created, force us to open the project, create a ticket, and close the project again. 

About the question of Eldy "If we include then in creation of tickets, what is goal of closing a project ?" the goal of closing a project is to establish that the project execution phase is end. This means can invoice it etc. But a project that is closed, can have an aftersales service. This is why still is needed to be able to select a closed project when a ticket is created.

PR #26228, opened 2023-10-16
FIX template invoice list extrafield filters (backport v18)

W1W1-M:
# FIX Broken extrafield filter on template invoices list (backport v18)
This PR includes `/core/tpl/extrafields_list_search_sql.tpl.php` in `compta/facture/invoicetemplate_list.php`


eldy:
Already pushed into lower version 17 so no need to push it in v18. Will appears after merge process.

PR #26234, opened 2023-10-17
FIX - Row in list higher height PR #26177

BB2A-Anthony:

As I explained to you following the PR #26177, I changed.

"height of column must use padding of td and not lineheight that has bad side effect"

Is it ok for you?
Is that what you are suggesting?


htdocs/core/lib/usergroups.lib.php
```diff
@@ -576,7 +576,6 @@ function showSkins($fuser, $edit = 0, $foruserprofile = false)
 	}
 
 	// Table line height
-	/* removed. height of column must use padding of td and not lineheight that has bad side effect
```
eldy, line 579: I may need a long time to detect side effect on CSS change. So can you keep option to end user not visible yet and have option hidden for the moment only. So we can switch test (with no effect on final version yet) as long as necessary...

BB2A-Anthony: @eldy I have use MAIN_FEATURES_LEVEL at 1

PR #26238, opened 2023-10-17
FIX project substitution

FlorentPoinsaut:
# FIX project substitution

The project attribute may be empty even if fk-project is not zero or null.


htdocs/core/lib/functions.lib.php
```diff
@@ -8190,7 +8190,7 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 				$substitutionarray['__ATTENDEE_LASTNAME__'] = isset($object->lastname) ? $object->lastname : '';
 			}
 
-			if (is_object($object->project)) {
+			if (is_object($object->project) || $object->fetch_project() > 0) {
```
eldy, line 8193: The method >fetch_project() is doing SQL request.
No SQL is allowed into a function designed to be run a very high number of time.
->fetch_project() must be called in the parent page

FlorentPoinsaut: OK. Thank you for your prevention. 
I will change this.

PR #26248, opened 2023-10-18
Fix::Correct full path tcpdf

Dupy007:
### FIX
When the absolute path is good it duplicates Document_Root


eldy:
Can you provide an example of content you have in the PDF that need this change ?

d3v2a:
We have a pdf generator which uses the absolute path of the file (files are in the document folder)  and this code add the document root to the path so if the path is /path/dolibarr/documents/file.ext, this code transforms it to /path/dolibarr/htdocs/path/dolibarr/documents/file.ext

d3v2a:
@eldy Did you need more information

eldy:
What value to you have for $imgsrc ?

Dupy007:
image

eldy:
It seems there is a pb inside the source of you PDF; 
To invclude an image from a local file you should have into tthe HTML conte

`<img src="file//var/www.."`

PR #26249, opened 2023-10-18
Fix: Correct full path tcpdf

Dupy007:
FIX

When the absolute path is good it duplicates Document_Root

### DIFF
--- a/htdocs/includes/tecnickcom/tcpdf/tcpdf.php
+++ b/htdocs/includes/tecnickcom/tcpdf/tcpdf.php
@@ -19243,7 +19243,7 @@ class TCPDF
 							// convert URL to server path
 							$imgsrc = str_replace(K_PATH_MAIN, K_PATH_URL, $imgsrc);
 						}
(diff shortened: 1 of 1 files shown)


htdocs/includes/tecnickcom/tcpdf/tcpdf.php
```diff
@@ -19243,7 +19243,7 @@ protected function openHTMLTagHandler($dom, $key, $cell)
 							// convert URL to server path
 							$imgsrc = str_replace(K_PATH_MAIN, K_PATH_URL, $imgsrc);
 						}
-					} elseif (($imgsrc[0] === '/') AND !empty($_SERVER['DOCUMENT_ROOT']) AND ($_SERVER['DOCUMENT_ROOT'] != '/')) {
+					} elseif (!file_exists($imgsrc) && ($imgsrc[0] === '/') AND !empty($_SERVER['DOCUMENT_ROOT']) AND ($_SERVER['DOCUMENT_ROOT'] != '/')) {
```
eldy, line 19246: $imgsrc seems to be an array so file_exists on an array parameter when expected a string is strange.

eldy, line 19246: Can you provide an example of value you have in the PDF that need this change ?

PR #26253, opened 2023-10-18
FIX : export separator config with multientity

Kazimir42:
I was trying to te set a default separator for export on another entity than the default one and i view it wasn't working.

After check the admin/import.php i notice some differences who shouldn't exist for me.

the changes come from this commit : 30b48a9ee3480fccf9e067ccd805874f695d6e22 (i don't really know why)

So i changed the file to be like before and like admin/import.php

--

Anyway if you prefere we can only fix the initial issue on change dolibarr_set_const() funciton like that :

`dolibarr_set_const($db, 'EXPORT_CSV_SEPARATOR_TO_USE', GETPOST('EXPORT_CSV_SEPARATOR_TO_USE', 'alphanohtml'), 'chaine', 0, '', $conf->entity);`

### DIFF
--- a/htdocs/admin/export.php
+++ b/htdocs/admin/export.php
@@ -40,15 +40,15 @@ if (!$user->admin) {
 }
 
 $action = GETPOST('action', 'aZ09');
-$value = GETPOST('value', 'alpha');
-$modulepart = GETPOST('modulepart', 'aZ09');	// Used by actions_setmoduleoptions.inc.php
 
 
 /*
  * Actions
  */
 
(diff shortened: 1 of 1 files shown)


eldy:
I don't know why change were done. I think for security purpose because the page was allowing a change any var, not only the CSV separator. Never mind, to match rest of code to save a variable that is different between entities, i suggest to fix the second way you suggest with
`dolibarr_set_const($db, 'EXPORT_CSV_SEPARATOR_TO_USE', GETPOST('EXPORT_CSV_SEPARATOR_TO_USE', 'alphanohtml'), 'chaine', 0, '', $conf->entity);
`

PR #26260, opened 2023-10-18
Fix #26259 : product's label in designatin column of MO Vinci template

### LINKED ISSUE #26259: Confusion between "description" and "designation" in Vinci MO pdf template
### Bug

In Vinci MO pdf template, the description of product is written in designation column, but it will be far more logical to have the product's label in this column 

### Environment Version

all

### Environment OS

all

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

vmaury:
# FIX#26259
Replace product description by product label in MO Vinci pdf template

### DIFF
--- a/htdocs/core/modules/mrp/doc/pdf_vinci.modules.php
+++ b/htdocs/core/modules/mrp/doc/pdf_vinci.modules.php
@@ -517,9 +517,9 @@ class pdf_vinci extends ModelePDFMo
 
 					if ($this->getColumnStatus('desc')) {
 						$pdf->startTransaction(); //description
-						$des = $prod -> description;
-						$descr = $des;//implode("<br>", $des);
-
+						/** $des = $prod -> description; it's far more logical to have the product's label instead of product's description
+						$descr = $des;//implode("<br>", $des); **/
+						$descr = $prod->label;
 						$this->printStdColumnContent($pdf, $curY, 'desc', $descr);
 
 						$pageposafter = $pdf->getPage();


htdocs/core/modules/mrp/doc/pdf_vinci.modules.php
```diff
@@ -517,9 +517,9 @@ public function write_file($object, $outputlangs = '', $srctemplatepath = '', $h
 
 					if ($this->getColumnStatus('desc')) {
 						$pdf->startTransaction(); //description
-						$des = $prod -> description;
-						$descr = $des;//implode("<br>", $des);
-
+						/** $des = $prod -> description; it's far more logical to have the product's label instead of product's description
+						$descr = $des;//implode("<br>", $des); **/
+						$descr = $prod->label;
```
eldy, line 522: So we should also change the key into getColumnStatus and into $this->cols['desc']

vmaury: I checked in some other modules (facture, order, propal), this col is named "desc", and it contains the product label (or the description if it's a free line)
So do you think it's realy necessary to change the col code from desc to label ?

vmaury: @eldy As we say in french "le mieux est l'ennemi du bien", and in this situation I really think it's better not to change the key in getColumnStatus 

- to keep this template similar to the others (facture, order, propal)
- to avoid problems to those have used the defineColumnField hook with desc key
- because changing this key could create some bugs

Issue #21425, opened 2022-07-01
When cloning a salary, after entering start of period, autofill in js also the field end of period (with end of month)


eldy:
### Feature Request

When cloning a salary, after entering start of period, autofill in js also the field end of period (with end of month compared to the entered start of period).

![image](https://github.com/Dolibarr/dolibarr/assets/883887/be0273be-d519-435a-939d-81c72c2ec2e4)

### Suggested steps

Add a link "Fill with end of month" after the field "Date fin" in the confirm popup. When clicking on the link, the Date is filled with the date of the last day of the same month than the date start.

PR #26263, opened 2023-10-18
FIX - Auto fill the field end of period with end of month when cloning a salary

Daoud-mohamed:
# Done
Adding button to auto fill the field end of period with end of month when cloning a salary.
Capture

- #21425

# Suggestion
I suggest removing this button and make it more easy by filling directely the field end date once the value of start date changes.

### DIFF
--- a/htdocs/salaries/card.php
+++ b/htdocs/salaries/card.php
@@ -741,17 +741,17 @@ if ($id > 0) {
 
 		//Add fill with end of month button
 		$formconfirm .= "<script>
-			$('#clone_date_end').after($('<button id=\"fill_end_of_month\" class=\"dpInvisibleButtons\" style=\"color: var(--colortextlink);font-size: 0.8em;opacity: 0.7;margin-left:4px;\" type=\"button\">". $langs->trans('FillEndOfMonth')."</button>'));
+			$('#clone_date_end').after($('<button id=\"fill_end_of_month\" class=\"dpInvisibleButtons\" style=\"color: var(--colortextlink);font-size: 0.8em;opacity: 0.7;margin-left:4px;\" type=\"button\">".$langs->trans('FillEndOfMonth')."</button>'));
 			$('#fill_end_of_month').click(function(){
 				var clone_date_startmonth = +$('#clone_date_startmonth').val();
 				var clone_date_startyear = +$('#clone_date_startyear').val();
 				var end_date = new Date(clone_date_startyear, clone_date_startmonth, 0);
 				end_date.setMonth(clone_date_startmonth - 1);
-				$('#clone_date_end').val( (new Intl.DateTimeFormat('".str_replace("_", "-", $langs->shortlang)."',{year: 'numeric',month: '2-digit',day: '2-digit'})).format(end_date));
+				$('#clone_date_end').val(formatDate(end_date,'".$langs->trans("FormatDateShortJavaInput")."'));
 				$('#clone_date_endday').val(end_date.getDate());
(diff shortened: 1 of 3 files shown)


aspangaro:
Can you provide a picture of the modification. Thanks in advance

Daoud-mohamed:
@aspangaro  Certainly! I've added a picture.

htdocs/salaries/card.php
```diff
@@ -738,6 +738,22 @@ function( data ) {
...
+
+		//Add fill with end of month button
+		$formconfirm .= "<script>
+			$('#clone_date_end').after($('<button id=\"fill_end_of_month\" class=\"dpInvisibleButtons\" style=\"color: var(--colortextlink);font-size: 0.8em;opacity: 0.7;margin-left:4px;\" type=\"button\">".$langs->trans('FillEndOfMonth')."</button>'));
+			$('#fill_end_of_month').click(function(){
+				var jsDate = $('#clone_date_start').datepicker('getDate');
+				if (jsDate !== null) {
+					jsDate instanceof Date;
+					var lastDate = new Date(jsDate.getFullYear(), jsDate.getMonth() + 1, 0);
+					$('#clone_date_endday').val(lastDate.getDate());
```
eldy, line 750: You get the date using 
var jsDate = $('#clone_date_start').datepicker('getDate');  then  lastDate.getDate() for the day, getMonth for the month, ...
But you should be able to get the value with 
$('#clone_date_startday').val()
$('#clone_date_startmonth').val()
...
Because the day, month and year are already always saved as structure data as hidden field when we have a date. 
The advantage is that your new features does not depends on datepicker. If we replace datepicker with another js component to fill date_start and its strcutured data, the code will continue to work.

Daoud-mohamed: @eldy 
I added the button without using the datepicker() library. 
I used the variable $langs->shortlang, which returns en_US or fr_FR .... 
I didn't find any variable that returns en-US instead of en_US.

htdocs/salaries/card.php
```diff
@@ -738,6 +738,21 @@ function( data ) {
...
+
+		//Add fill with end of month button
+		$formconfirm .= "<script>
+			$('#clone_date_end').after($('<button id=\"fill_end_of_month\" class=\"dpInvisibleButtons\" style=\"color: var(--colortextlink);font-size: 0.8em;opacity: 0.7;margin-left:4px;\" type=\"button\">". $langs->trans('FillEndOfMonth')."</button>'));
+			$('#fill_end_of_month').click(function(){
+				var clone_date_startmonth = +$('#clone_date_startmonth').val();
+				var clone_date_startyear = +$('#clone_date_startyear').val();
+				var end_date = new Date(clone_date_startyear, clone_date_startmonth, 0);
+				end_date.setMonth(clone_date_startmonth - 1);
+				$('#clone_date_end').val( (new Intl.DateTimeFormat('".str_replace("_", "-", $langs->shortlang)."',{year: 'numeric',month: '2-digit',day: '2-digit'})).format(end_date));
```
eldy, line 750: We are on the good way. Still one thing to change. Here you use the country of user to format the visible string of date.
A better solution is to use the string format that is related to the active user language and found with
$langs->trans("FormatDateShortJavaInput")
Just replace the string %Y with year on 4 char, %m with month on 2 char, %d with day on 2 char in this string and you will get the string to inject into $("#clone_date_end)
Si we are using the same method than into the lethod selectDate() used to output the date selector.

Daoud-mohamed: Good, I found a function formatDate(date,format) into core\js\lib_head.js.php it return the needed output format.

PR #26268, opened 2023-10-19
FIX in_numeric not exists

hregis:


mdeweerd:
This is also part of #26264 .

hregis:
@mdeweerd yes but my fixe is for v18

hregis:
@mdeweerd ok I close my PR thanks

mdeweerd:
@hregis I am not familiar with your flow, but you are correct to mention that this applies to several versions that are under maintenance.

PR #26274, opened 2023-10-19
Fix Reception: refactor code to simplify indentation

alexandre-janniaux:
# Fix Simplify the Reception class by erroring out faster 

This merge request refactor the Reception class in the following ways:
 - Early return is used when the error path is terminal for the function, moving the error handling closer to the call making the error and removing (a lot of) indentation level for the success path.
 -  Almost identical branches are refactored together with specific assignment to the parameter instead.
 - `$error` variable is removed when it's not used anymore.

In addition, some fixes are included:
 - Documentation fixes (**reviewer**: see `reception: fix typo in doc` to ensure the fix is correct).
 - Copy-paste typos when logging to the syslog.
 - `CommonObjet::setErrorsFromObject` is used instead of assigning `error` and `errors` manually.

Finally, a new test was added in previous MR #26310 to check that nothing was broken by the previous changes, except for the function `setBilled`, `setDeliveryDate` and `fetch_lines` which I haven't tested yet, but I can provide those tests before the merge if needed.

### DIFF
--- a/htdocs/reception/class/reception.class.php
+++ b/htdocs/reception/class/reception.class.php
@@ -1753,29 +1753,28 @@ class Reception extends CommonObject
 					$mouvS->origin = &$this;
 					$mouvS->setOrigin($this->element, $this->id);
 
-					if (empty($obj->batch)) {
-						// line without batch detail
+					$inventorycode = '';
+					$date_eatby = '';
+					$date_sellby = '';
+					$batch = '';
+					$fk_origin_stock = 0;
 
-						// We decrement stock of product (and sub-products) -> update table llx_product_stock (key of this table is fk_product+fk_entrepot) and add a movement record
-						$inventorycode = '';
-						$result = $mouvS->livraison($user, $obj->fk_product, $obj->fk_entrepot, $qty, $obj->cost_price, $langs->trans("ReceptionUnClassifyCloseddInDolibarr", $numref), '', '', '', '', 0, $inventorycode);
+					if (!empty($obj->batch))
+					{
+						$date_eatby = $this->db->jdate($obj->eatby);
+						$date_sellby = $this->db->jdate($obj->sellby);
+						$batch = $obj->batch;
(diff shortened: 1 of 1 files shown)


htdocs/reception/class/reception.class.php
```diff
@@ -137,8 +139,55 @@ class Reception extends CommonObject
 	// We can use this to know warehouse planned to be used for each lot.
 	public $detail_batch;
 
+	/**
+	 * Workflow of the status variables:
+	 *
+	 * A reception is first created as a draft. It can then be validated
+	 * to apply the changes in the stock, and the closed when the stock
```
eldy, line 146: With develop, it is the workflow module that decide when the stock is modified.
Can be when validating reception and can be when closing reception.

alexandre-janniaux: Thanks for the quick review and the pointer, I'll check the path more in details and address that in the week end.

eldy:
Also can you split or in 2. 1 for the addition of phpunit test and the other one for code quality ?

alexandre-janniaux:
It will require some work since the reindent changes are a pretty conflicting changes, but it might help unblocking the rest of my work too so that's a good idea. I'll do that.

htdocs/reception/class/reception.class.php
alexandre-janniaux: > With develop, it is the workflow module that decide when the stock is modified.

After some research, I'm not sure I understand this, the Reception class is not calling anything related to the Workflow module, and vice-versa, and it creates the MouvementStock entries. Is it related to future work?

eldy (commented):
There is never explicit call to the workflow. Call to the workflow module is done automatically because it is code implemented into a trigger

eldy:
Due to a too high number of other changes on the same file and conflict generated by that, i was not able to merge this PR, so i prefer to close it instead of keeping it open when unmergeable.

Issue #26142, opened 2023-10-09
Make product description available as a column


bos4711:
### Feature Request

Today, in v17.0.2, when listing products it is not possible to select Product Description column when browsing through the product list. You can select Label, Weight, Cost price, or whatever, but not Description. Having the Description visible when listing all products would really help when many products are similiar.

![2023-10-09_064235](https://github.com/Dolibarr/dolibarr/assets/11036018/043042c0-da63-4ef0-b2a6-da587f1c656c)

### Suggested implementation

Add relevant code in `/product/list.php`

ksar-ksar:
Hello,

It's the actual choice, because Description could be very big and not adapted to be displayed in List format.

bos4711:
I agree that for very big descriptions it will be a mess, but how about just using a short excerpt of the description (like first 200 chars) or similiar, so the list won't be ruined?

bos4711:
@ksar-ksar In PR #26193 @eldy suggested the use of `dolGetFirstLineOfText() ` which I think will solve the issue here as well.

bos4711:
Fixed by PR #26287

PR #26287, opened 2023-10-20
Update list.php

bos4711:
# FIX 26142
Add the Description column in the list of possible columns to select from when viewing Products.


bos4711:
FIX #26142

PR #26289, opened 2023-10-20
[Discount] fix SQL syntax error introduced in 9861b50

AmazingBeu:
This pull request just remove the parenthesis no more needed since the change in the commit 9861b50.
These parenthesis prevent to delete a discount


hansemschnokeloch:
Fixes also https://github.com/Dolibarr/dolibarr/issues/26619

PR #26299, opened 2023-10-20
Fix : Use explicit JOINS on lists instead of comma join

thomas-Ngr:
Fix error when using hook printFieldListFrom in a module.

One should not use comma-joins, as stated in the MySQL manual:
"The precedence of the comma operator is less than that of INNER JOIN, CROSS JOIN, LEFT JOIN, and so on. If you mix comma joins with the other join types when there is a join condition, an error of the form Unknown column 'col_name' in 'on clause' may occur. Information about dealing with this problem is given later in this section. " [source](https://dev.mysql.com/doc/refman/8.0/en/join.html)

In a PrintFieldListFrom hook on facture list, I add this statement : `LEFT JOIN llx_element_contact ec ON ec.element_id = f.rowid` and I get this error : `Unknown column 'f.rowid' in 'on clause'`

### DIFF
--- a/htdocs/comm/propal/list.php
+++ b/htdocs/comm/propal/list.php
@@ -533,7 +533,7 @@ $sql .= " LEFT JOIN ".MAIN_DB_PREFIX."c_departements as state on (state.rowid = 
 if (!empty($search_categ_cus) && $search_categ_cus != '-1') {
 	$sql .= ' LEFT JOIN '.MAIN_DB_PREFIX."categorie_societe as cc ON s.rowid = cc.fk_soc"; // We'll need this table joined to the select in order to filter by categ
 }
-$sql .= ', '.MAIN_DB_PREFIX.'propal as p';
+$sql .= ' INNER JOIN '.MAIN_DB_PREFIX.'propal as p ON p.fk_soc = s.rowid';
 if (is_array($extrafields->attributes[$object->table_element]['label']) && count($extrafields->attributes[$object->table_element]['label'])) {
 	$sql .= " LEFT JOIN ".MAIN_DB_PREFIX.$object->table_element."_extrafields as ef on (p.rowid = ef.fk_object)";
 }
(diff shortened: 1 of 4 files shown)


htdocs/comm/propal/list.php
```diff
@@ -548,22 +548,22 @@
 $sql .= " LEFT JOIN ".MAIN_DB_PREFIX."c_availability as ava on (ava.rowid = p.fk_availability)";
 // We'll need this table joined to the select in order to filter by sale
 if ($search_sale > 0 || (!$user->rights->societe->client->voir && !$socid)) {
-	$sql .= ", ".MAIN_DB_PREFIX."societe_commerciaux as sc";
+	$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."societe_commerciaux as sc ON s.rowid = sc.fk_soc";
```
eldy, line 551: Left join has not the same result than exact join (inner join)

htdocs/commande/list.php
```diff
@@ -449,7 +449,7 @@
 if (($search_categ_cus > 0) || ($search_categ_cus == -2)) {
 	$sql .= ' LEFT JOIN '.MAIN_DB_PREFIX."categorie_societe as cc ON s.rowid = cc.fk_soc"; // We'll need this table joined to the select in order to filter by categ
 }
-$sql .= ', '.MAIN_DB_PREFIX.'commande as c';
+$sql .= 'RIGHT JOIN '.MAIN_DB_PREFIX.'commande as c ON c.fk_soc = s.rowid';
```
eldy, line 452: We must avoid right join that generated a very difficult way to understand readability of requests. Only inner and left join should be used.

htdocs/commande/list.php
```diff
@@ -464,19 +464,19 @@
...
+	$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."c_type_contact as tc ON ec.fk_c_type_contact = tc.rowid AND tc.element='commande' AND tc.source='internal'";
 }
 
 // Add table from hooks
 $parameters = array();
 $reshook = $hookmanager->executeHooks('printFieldListFrom', $parameters, $object); // Note that $action and $object may have been modified by hook
 $sql .= $hookmanager->resPrint;
 
-$sql .= ' WHERE c.fk_soc = s.rowid';
+$sql .= ' WHERE c.rowid IS NOT NULL';
```
eldy, line 479: This generates very bad performance and can be avoid by using inner join instead of right join.

hregis:
@thomas-Ngr just for fun 😉

htdocs/comm/propal/list.php
thomas-Ngr: Yes, if you compare comma-join with left join without considering the sorting arguments.

 But in this case, the statement was:

```
... , llx_societe_commerciaux as sc
...
WHERE...
AND s.rowid = sc.fk_soc
```

Which has *exactly* the same result as, and is more understandeable than : 

```
LEFT JOIN llx_societe_commerciaux as sc ON s.rowid = sc.fk_soc
```

Repeating myself : Please note that **comma-join is NOT equivalent to inner-join"". "The precedence of the comma operator is less than that of INNER JOIN, CROSS JOIN, LEFT JOIN, and so on" [source on MySQL manual](https://dev.mysql.com/doc/refman/8.0/en/join.html)

htdocs/commande/list.php
thomas-Ngr: OK, I guess I will simply replace every comma-join by an explicit INNER JOIN.

thomas-Ngr:
@hregis, I don't see any comma-join in this graph, which is the point here.

thomas-Ngr:
@eldy OK, I reworked this PR using only INNER JOINs.

htdocs/commande/list.php
eldy: > OK, I guess I will simply replace every comma-join by an explicit INNER JOIN.

Yes,for similar result, comma-join can be replaced with inner join. It is the other way to write inner join.

eldy:
As it is a re-engeneering code, it must be pushed into develop.

thomas-Ngr:
@eldy I close this PR and open this one for following : #26530

PR #26302, opened 2023-10-20
NEW dol_debug function to send message to the debugbar

hansemschnokeloch:
# NEW dol_debug function : easily send a message to the debugbar

The debugbar is a powerfull tool that is not commonly used by developpers (see https://www.dolibarr.fr/forum/t/comment-dumper-une-variable-dans-la-debugbar/44491/3)
This new function will allow to send messages to the debugbar in a similar way that dol_syslog operates: dol_debug instructions can stay in the code and will only be triggered if the debugbar module is active.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -1910,6 +1910,25 @@ function dol_syslog($message, $level = LOG_INFO, $ident = 0, $suffixinfilename =
 }
 
 /**
+ *	Write messages into the debugbar
+ *
+ * 	@param  mixed		$message				Message to write into the debugbar. Can be text or any php object
+ * 	@param  string		$level					Log level
+ *
+ *  @return	void
+ */
+function dol_debug($message, $level = 'debug')
+{
+	global $debugbar;
+
+	// If debugbar module enabled
+	if (!isModEnabled('debugbar') || !isset($debugbar) || !isset($debugbar['messages'])) {
+		return;
+	}
+	$debugbar['messages']->addMessage($message, $level);
+}
+
+/**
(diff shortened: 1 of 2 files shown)


eldy:
Why not just using dol_syslog with LOG_DEBUG level ?

hansemschnokeloch:
If we use `dol_syslog` to send messages to the debugbar we have to activate the two modules `modSyslog` and `modDebug` in order to send messages to the debugbar. Therefore we loose the independancy of the two modules.

eldy:
Currently the debugbar already rely on the module syslog to show logs.
If module syslog is not enabled,  it means we want to track logs, so we want them both on file and browser side.
If module syslog is not enabled, it means we don't need to track what happe, so why having it on debug bar ?

What is goal of having 4 combinations of log :
- log in debug bar and not on file
- log in files and not in debug bar
- log in debugbar + file
- log nowhere
This means also that developers will have to  think of the 4 situations when inserting its tracking information (do i want it in log and not in debugbar, do i want it into debugbar and not into logfile). Why adding such a complexity ?

If an information is interesting for log tracking it should appear in all log tracking layers output or nowhere.

hansemschnokeloch:
I think there is a misunderstanding here. What I'm proposing is the possibility to use the "messages" component of the debugbar. This component allow us to dump any variable in the debugbar. That is not possible currently with dol_syslog.

For example
`dol_syslog($object, LOG_DEBUG)` ends up with an error as you can see in the debugbar

FA23-017267-Fiche(1)

`dol_debug($object)` uses the messages component of the debugbar and dumps the variable like this

FA23-017267-Fiche

PR #26303, opened 2023-10-20
FIX #26296 call trigger TICKET_ASSIGNED after creation of ticket, if a user got assigned

### LINKED ISSUE #26296: Trigger TICKET_ASSIGNED not called during creation of new ticket
### Bug

The trigger TICKET_ASSIGNED shall notify, whenever a user gets assigned to a ticket.

This is executed correctly, when for an already created ticket, a user gets assigned.

But when a ticket gets newly created and a user is entered as an assigned user, the trigger is not executed.

### Environment Version

17.0.3 and 18.0.2

The trigger should be executed every time, when an assignment of a user to a ticket happens.

Create a new ticket, enter an assigned user in the form, create the ticket: no trigger is executed.

priojk:
The trigger TICKET_ASSIGNED shall notify, whenever a user gets assigned to a ticket.

This is executed correctly when for an already created ticket a user gets assigned. 
But when a ticket gets newly created and a user is entered as an assigned user in the create page, the user is set as assigned, but the trigger is not called.

This PR calls the trigger TICKET_ASSIGNED once a new ticket has definitely been created (after the db->commit()). The trigger is called, when a user has been set from the drop-down-field or when the constant TICKET_AUTO_ASSIGN_USER_CREATE was set and the creator got automatically assigned.

Consistent notifications about assigns are required for a notification module, which informs users visually, when a new ticket got assigned to them.

### DIFF
--- a/htdocs/ticket/card.php
+++ b/htdocs/ticket/card.php
@@ -286,6 +286,11 @@ if (empty($reshook)) {
 			if (!$error) {
 				$db->commit();
 
+				// if a user got defined as an assigned user: run trigger
+				if (!empty($object->fk_user_assign)) {
+					$res = $object->call_trigger("TICKET_ASSIGNED", $user);
+				}
+
 				if (!empty($backtopage)) {
 					if (empty($id)) {
 						$url = $backtopage;


htdocs/ticket/card.php
```diff
@@ -286,6 +286,11 @@
 			if (!$error) {
 				$db->commit();
 
+				// if a user got defined as an assigned user: run trigger
+				if (!empty($object->fk_user_assign)) {
```
eldy, line 290: We already have a trigger for the create and it is job of the trigger to intercept also this case to scan if an assignement is done directly during creation and to execute its custom code.
Rule is to have only 1 trigger per business CRUD event, even when several actions are done in same transaction.

priojk: Having only one trigger call makes perfect sense, thanks for the explanation. I added this rule also to the wiki.

PR #26304, opened 2023-10-20
NEW add ORDER_SUPPLIER_APPROUVE2 trigger

FlorentPoinsaut:
# NEW add ORDER_SUPPLIER_APPROUVE2 trigger


htdocs/core/class/notify.class.php
```diff
@@ -79,6 +79,7 @@ class Notify
 		'FICHINTER_ADD_CONTACT',
 		'ORDER_SUPPLIER_VALIDATE',
 		'ORDER_SUPPLIER_APPROVE',
+		'ORDER_SUPPLIER_APPROVE2',
```
eldy, line 82: No need to duplicate the trigger code
You can reuse ORDER_SUPPLIER_APPROVE and set 
$this->context['secondlevel'] = 1;
before calling
$result = $this->call_trigger($trigger, $user)

FlorentPoinsaut: Hum ok but if I don't add a trigger code, how could I set a dedicated email template?

eldy: > Hum ok but if I don't add a trigger code, how could I set a dedicated email template? 

With tha var into the context that is different

FlorentPoinsaut: Oh I see, context was introduced after the release v16.
I put this PR in draft and I update when my client upgrade to the v18.

Issue #26272, opened 2023-10-19
FK violation when merging companies with Reception objects attached


alexandre-janniaux:
### Bug

When merging two companies with existing Reception object attached, the following error happens

```
ERROR: 23503: update or delete on table 'llx_societe' violates foreign key constraint 'fk_reception_fk_soc' on table 'llx_reception' DETAIL: Key (rowid)=(822) is still referenced from table 'llx_reception'. SCHEMA NAME: public TABLE NAME: llx_reception CONSTRAINT NAME: fk_reception_fk_soc LOCATION: ri_ReportViolation, ri_triggers.c:2797
Echec de la fusion de 2 enregistrements. Demande annulée.
```

Indeed the lists are not accounting Reception:
- https://github.com/Dolibarr/dolibarr/blob/268812fa3c014c3426842ac39984aa7f6e57ada1/htdocs/societe/card.php#L293C19-L293C19
- https://github.com/Dolibarr/dolibarr/blob/268812fa3c014c3426842ac39984aa7f6e57ada1/htdocs/societe/class/api_thirdparties.class.php#L437

And thus the associated reception are not handled.

### Environment Version

17.0.2

### Expected and actual behavior

Expected behaviour:

The companies should be merged and Reception objects from the deleted company should be reassigned to the resulting merged company.

Actual behaviour:

Red alert with the following error:

```
ERROR: 23503: update or delete on table 'llx_societe' violates foreign key constraint 'fk_reception_fk_soc' on table 'llx_reception' DETAIL: Key (rowid)=(822) is still referenced from table 'llx_reception'. SCHEMA NAME: public TABLE NAME: llx_reception CONSTRAINT NAME: fk_reception_fk_soc LOCATION: ri_ReportViolation, ri_triggers.c:2797
Echec de la fusion de 2 enregistrements. Demande annulée.
```

### Steps to reproduce the behavior

Create two companies, create a Reception objects assigned to the company that will be deleted, and try merging the companies.

alexandre-janniaux:
I have some refactors for Societe, moving the merge code to the Societe class, fixing this issue and adding a new test for that, can I be assigned to this issue? @eldy 

Also, should I send the fix after refactor and create a new MR for the backport, or should I just fix it before moving the mergeCompany code to Societe?

alexandre-janniaux:
[...]

PR #26310, opened 2023-10-21
FIX|Fix some minor issues on Reception and add a new test for it

alexandre-janniaux:
# FIX|Fix Simply Reception and add a new test for it

This merge request brings a new test along with some fixes that are blocking it:
 - `$this->status` is now correctly assigned the same value as the deprecated `$this->statut`, and is tested.
 - $weight is now defined on the class

The new test is added to check that nothing will be broken by future changes, except for the function `setBilled`, `setDeliveryDate` and `fetch_lines` which I haven't tested yet.

The test is a necessary base and work to send the test for the fix for #26272. 

Extracted from PR https://github.com/Dolibarr/dolibarr/pull/26274

### DIFF
--- a/htdocs/reception/class/reception.class.php
+++ b/htdocs/reception/class/reception.class.php
@@ -1959,6 +1959,7 @@ class Reception extends CommonObject
 
 			if (!$error) {
 				$this->statut = self::STATUS_DRAFT;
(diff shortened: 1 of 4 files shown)


alexandre-janniaux:
Not sure what happened on the CI. :/

PR #26327, opened 2023-10-23
FIX Max version of PHP

aspangaro:
For v18, this table in wiki define PHP8.2 as maximum version: https://wiki.dolibarr.org/index.php?title=Versions

Also this morning with an update test in 18.0.2 :

Clipboard - 23 octobre 2023 06 42


eldy:
8.2 compatibility is not complete in v18 it is better to fix doc to be more clear.

PR #26335, opened 2023-10-23
fix: if field is boolean in fileds array, it must be boolean as datatype for PgSQL

FHenry:
I didn't put migration script as this module is still in develop state in 18.0 and in develop branch

Else you've got this type of message

ERROR: 42804: column 'not_depreciated' is of type integer but expression is of type boolean LINE 1: ...tatus) VALUES ('qsdsqd', 'qsdqsd', NULL, 0, 0, 0, true, '202... ^ HINT: You will need to rewrite or cast the expression. LOCATION: transformAssignedExpr, parse_target.c:583

### DIFF
--- a/htdocs/install/mysql/tables/llx_asset-asset.sql
+++ b/htdocs/install/mysql/tables/llx_asset-asset.sql
@@ -43,13 +43,13 @@ CREATE TABLE llx_asset(
     acquisition_type        smallint        DEFAULT 0 NOT NULL,
     asset_type              smallint        DEFAULT 0 NOT NULL,
 
-    not_depreciated         integer         DEFAULT 0,
+    not_depreciated         boolean         DEFAULT false,
 
     disposal_date           date,
     disposal_amount_ht      double(24,8),
     fk_disposal_type        integer,
-    disposal_depreciated    integer         DEFAULT 0,
-    disposal_subject_to_vat integer         DEFAULT 0,
+    disposal_depreciated    boolean         DEFAULT false,
+    disposal_subject_to_vat boolean         DEFAULT false,
 
     note_public             text,
     note_private            text,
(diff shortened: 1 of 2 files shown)


eldy:
We should avoid boolean type in database because it can't evolve when we need to add a third value. We use smallint fir such needs.
Pb is surely in sql request on test that must be.  If(var = 1).  instead of  if (var)
Behaviour may also differs from database.

FHenry:
> We should avoid boolean type in database because it can't evolve when we need to add a third value. We use smallint fir such needs. Pb is surely in sql request on test that must be. If(var = 1). instead of if (var) Behaviour may also differs from database.

SQL query produce by in SQL INSERT or UPDATE query come from quote function (https://github.com/Dolibarr/dolibarr/blob/2c01a59e6039693cfa8c8f0298b2d1fb38a49076/htdocs/core/class/commonobject.class.php#L9407) from CommonObject call by createCommon or updateCommon, if field  type (from object->fields) is "boolean" then SQL value is "true" or "false", and with pgSQL set true of false for an integer is not allowed.  In MySQL/Marraidb boolean are SQL column type is an alias with tinyint(1).
But boolean field type (from object->fields or extrafields) is on form with a visual checkbox, that what is wanted in this module  (by it first design), so two solutions here :
 - Change field type with Interger and add with arrayvalue Yes/No : but this will display a select/option instead of a check box
 - accept that boolean as DB type and make modulebuilder create this SQL column as boolean with fields if typed like that in fields class (as it is in was created with extrafields).

Whatever the solution you prefer, let me know it, but for this class and this module (in develop state) the conversion with PgSql with "alter table/alter column" from int to boolean is not allowed... 

I'm agree that boolean SQL type is not ideal, but now it's in Dolibarr core and if we want to keep checkbox as visual input, what to do ?

eldy:
I see.
So lets take the choice 2 that seems to be what you suggest with your PR.

PR #26344, opened 2023-10-24
FIX : wrong decimals on dispatch page

atm-adrien:
## FIX : wrong decimals on dispatch page

When I am on the quantity entry page (for receipts), it appears that the quantities displayed have decimal places that are not supposed to exist (I think they are generated because of the SUM function of the query )

### DIFF
--- a/htdocs/fourn/commande/dispatch.php
+++ b/htdocs/fourn/commande/dispatch.php
@@ -644,7 +644,7 @@ if ($id > 0 || !empty($ref)) {
 
 		// Set $products_dispatched with qty dispatched for each product id
 		$products_dispatched = array();
-		$sql = "SELECT l.rowid, cfd.fk_product, round(sum(cfd.qty), 2) as qty";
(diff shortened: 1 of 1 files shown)


htdocs/fourn/commande/dispatch.php
```diff
@@ -644,7 +644,7 @@
 
 		// Set $products_dispatched with qty dispatched for each product id
 		$products_dispatched = array();
-		$sql = "SELECT l.rowid, cfd.fk_product, sum(cfd.qty) as qty";
+		$sql = "SELECT l.rowid, cfd.fk_product, round(sum(cfd.qty), 2) as qty";
```
eldy, line 647: rounding must never be done by the database because database has no knowledge of Dolibar rounding rules.
Instead the rounding must be done in PHP. It is done later by 
price2num($objd->qty, 5);
that should be
price2num($objd->qty, 'MS');

atm-adrien:
@eldy

atm-adrien:
@eldy

atm-adrien:
@eldy

Issue #25600, opened 2023-08-09
Add a PDF option to show the value (number) of the barcode of product (just under the ref) 


eldy:
### Feature Request

Add a PDF option to show the value (number) of the barcode of product (just under the ref) 
shipment documents
proposal
order
invoice
purchase proposal
purchase order
purchase invoice

Output on PDF is 
Ref - Label or product
Description

Must be 
Ref - Label or product
Barcode (if exists and if module is on)
Description

finalbeta:
After enabling the module. I noticed that I had no way to get the serial numbers on invoices. 

We continu to place serial numbers on invoices manually as a legal obligation for warranty exists, and invoices are the proof that the product with that serialnumber came from us. 
So this would definitly be a usefull feature.

PR #26345, opened 2023-10-24
New: Adding the PDF option displaying the product's barcode number

Daoud-mohamed:
# New

Adding the PDF option displaying the product's barcode number in the following documents:
Shipment documents
Purchase proposal
Purchase order
Vendor invoice
Customer invoice

Capture

- #25600


htdocs/core/lib/pdf.lib.php
```diff
@@ -1541,6 +1542,10 @@ function pdf_getlinedesc($object, $i, $outputlangs, $hideref = 0, $hidedesc = 0,
 		}
 	}
 
+	if (isModEnabled('barcode') && getDolGlobalString('MAIN_GENERATE_DOCUMENTS_SHOW_PRODUCT_BARCODE') && !empty($product_barcode)) {
```
eldy, line 1545: May be you can do 
$libelleproduitservice = dol_concatdesc($libelleproduitservice, $langs->trans("BarCode")." ".$product_barcode);
So it is clear for PDF reader that this number alone on the line is a barcode.

Daoud-mohamed: @eldy You're right! that will be more clear, I added the barcode label

PR #26347, opened 2023-10-24
Qual cleaner code for InterfaceLogevents

W1W1-M:
# Qual cleaner PHP code for InterfaceLogevents class

- Refactored code to make more readable and efficient
- Abstracted code for event data to a class method to eliminate repeated code
- Refactored to make use of parent class constructor
- Various fixes & optimizations to code & comments

The general idea was to make use of some of the best PHP OOP practices I know of. I'm open to any feedback.


htdocs/core/triggers/interface_20_all_Logevents.class.php
```diff
@@ -76,103 +80,70 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
...
 				break;
 			case 'USER_NEW_PASSWORD':
 				dol_syslog("Trigger '" . $this->name . "' for action '$action' launched by " . __FILE__ . ". id=" . $object->id);
-				// Initialize data (date,duree,text,desc)
-				$text = $langs->transnoentities("NewUserPassword", $object->login);
-				$desc = $langs->transnoentities("NewUserPassword", $object->login);
+				$this->initEventData("NewUserPassword", $object, $langs, true);
 				break;
-			case 'USER_ENABLEDISABLE':
+			case 'USER_ENABLEDISABLE': // TODO maybe divide this action into 2 separate actions for better traceability
```
eldy, line 109: We try to do the opposite:  reduce the number of trigger codes (in perfect word, we should have only trigger CREATE, UPDATE, DELETE and use the property of object to make the difference between actions.

htdocs/core/triggers/interface_20_all_Logevents.class.php
```diff
@@ -188,4 +159,24 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
...
+	 * @param	Object		$object				Object
+	 * @param	Translate	$langs				Object langs
+	 * @param	bool		$user_else_group	Bool to define if localized string param is Object login or name
+	 * @return	void
+	 */
+	private function initEventData(string $key, Object $object, Translate $langs, bool $user_else_group): void
+	{
+		$langs->load("users");
+		$this->event_date = dol_now();
+		$this->event_label = $this->event_desc = $langs->transnoentities($key, $user_else_group ? $object->login : $object->name);
```
eldy, line 176: The goal of this trigger is to report technical security information. Goal is to be analyzed for stats of security analysis.
If you translate the string with the language of the connected user, this make the use of the data difficult to exploit because we can have a ton of different values. We will get in the same column values like

User abc modified
Usuario def creation
Utilisateur ijk créé

This may make automatic and mass analysis difficult. So w should instead store into database with a structured string like
(Key,userorlogin)
Must not depend on the language.
Then, on page admin/tools/listevents.php to list security event, we make the translation on the fly in the language of reading user.

htdocs/core/triggers/interface_20_all_Logevents.class.php
W1W1-M: Ok, thanks for explaining this. I think this will need more substantial work and should be done separately.

htdocs/core/triggers/interface_20_all_Logevents.class.php
W1W1-M: Thanks for input 👍 I've refactored to follow your guidelines. Event description in database will now only be in English. In order for the translations to keep working and also conserve existing keys I've added an array to match between the action key and localization keys (It would be possible to just use action key but then all lang files would be out of sync). 

The localization is now done on frontend using a static class method to check if the event description holds an action key so it can be localized. It would have been possible to use the event type for this but that would go against previous comment on simplifying number of triggers. This could be improved on after working on triggers.

W1W1-M:
Code still needs some more cleaning ...

PR #26361, opened 2023-10-25
fix suppliers bills payment link

sonikf:
Hi @eldy 
This actually doesn't fix the menu.
What is the right way in situations like this should I add update entries in all relevant migration files, meaning 16.0.0-17.0.0.sql, 17.0.0-18.0.0.sql and 18.0.0-19.0.0.sql?


eldy:
Fix should fix for new installation 
For migration, yes please fix all migeation files in each version.

PR #26363, opened 2023-10-25
NEW Allow notification to attach PDF from an ODT template

### LINKED ISSUE #24869: The notification module should be able to send an attachment if it is generated by an ODT template.
### Feature Request

If a PDF is generated via an ODT template and transformed into a PDF, it will not be attached to an email notification even if the "joinfiles" field is at 1.

Indeed, when a PDF is generated via an ODT template, it suffixes the generated PDF with the name of the template preceded by the character `_`

Example for a supplier order:
- Name of the template: `example.odt`
- Ref of the supplier order: `SO290-1019`
=> Name of the PDF: `SO290-1019_example.pdf`

While the notification system is limited to search if `SO290-1019.pdf` exists.

### Suggested implementation

I wonder how to correct this.
In my humble opinion, it would be better to remove or make optional this suffix story when generating via ODT.

### Suggested steps

What do you think about it?

If you agree, I can take care of creating the fix pretty quickly.

FlorentPoinsaut:
# NEW Allow notification to attach PDF from an ODT template

This amendment harmonises the notification behaviour on the HTML form behaviour by attach the most recent file.

Closes #24869


htdocs/core/class/notify.class.php
```diff
@@ -651,15 +633,11 @@ public function send($notifcode, $object, $filename_list = array(), $mimetype_li
...
 
 						$ref = dol_sanitizeFileName($newref);
-						$pdf_path = $dir_output."/".$ref.".pdf";
-						if (!dol_is_file($pdf_path)||(is_object($arraydefaultmessage) && $arraydefaultmessage->id > 0 && !$arraydefaultmessage->joinfiles)) {
-							// We can't add PDF as it is not generated yet.
-							$filepdf = '';
-						} else {
-							$filepdf = $pdf_path;
+						if (is_object($arraydefaultmessage) && $arraydefaultmessage->id > 0 && !$arraydefaultmessage->joinfiles) {
+							$filepdf = (dol_most_recent_file($dir_output, preg_quote($ref, '/').'[^\-]+'))['fullname'];
```
eldy, line 637: The dol_most_recent_filecreturn the latest file linked to object. It can be any document, a legal private document for example.
On notification, we only want to send document that are the generated document. So document call ref.pdf it it exists.

FlorentPoinsaut: Yes, that's why there's a filter that limits to a list to the only file starting with the ref.
But if it's not limiting enough, how can I proceed?

eldy:
CI conflict open since a too long time. No way to solve/merge it so i close.

PR #26365, opened 2023-10-25
Workflow: code quality with early return

alexandre-janniaux:
# Qual Workflow: use early return to simplify code

Reduce the level of indentation in the Workflow code by using early return when possible.
Some conditions with early return were already present but some location weren't using it despite being really nested between loops and conditions.

No functional changes should be expected from those.


htdocs/core/triggers/interface_20_modWorkflow_WorkflowManager.class.php
```diff
@@ -80,24 +81,22 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
...
-						$this->error = $newobject->error;
-						$this->errors[] = $newobject->error;
-					}
+				$newobject->context['createfrompropal'] = 'createfrompropal';
+				$newobject->context['origin'] = $object->element;
+				$newobject->context['origin_id'] = $object->id;
 
-					$object->clearObjectLinkedCache();
+				$ret = $newobject->createFromProposal($object, $user);
+				if ($ret < 0)
```
eldy, line 94: Try to always use { on condition, even if one line

alexandre-janniaux: Ok no problem.

PR #26385, opened 2023-10-27
FIX select supplier order vat_rate 0%

General-Q:
# FIX|Fix #[*FIX select supplier order vat_rate 0%*]
[*Disables the selection of other values than 0% while creating or editing a supplier order when the supplier is not subject to vat*]


htdocs/core/class/html.form.class.php
```diff
@@ -6622,10 +6622,18 @@ public function load_tva($htmlname = 'tauxtva', $selectedrate = '', $societe_ven
 				$return .= '<select class="flat minwidth50imp maxwidth100" id="' . $htmlname . '" name="' . $htmlname . '"' . ($disabled ? ' disabled' : '') . $title . '>';
 			}
 
+			$fourn = boolval(0);
+			$assujtva = boolval(1);
+			if(!$societe_vendeuse->tva_assuj && $societe_vendeuse->fournisseur){
+				$fourn = 1;
+				$assujtva = 0;
+			}
+
```
eldy, line 6631: Few line before, we have
$defaulttx = get_default_tva($societe_vendeuse, (is_object($societe_acheteuse) ? $societe_acheteuse : $tmpthirdparty), $idprod);

And inside get_default_tva, we should have the test on $societe_vendeuse->tva_assuj that is more complete (value can be 0/1 or 'franchise'/'reel') to return 0. 
So VAT should already be 0 when seller is not using VAT.

If not, thanks to describe how to reproduce the bug.

eldy:

PR #26390, opened 2023-10-30
Fix : version 18.0 PHP compatibility is 8.2

atm-maxime:
https://wiki.dolibarr.org/index.php?title=Versions


eldy:
I have fixed the wiki instead. Use with PHP 8.2 has still too many problems.

atm-maxime:
Ok. We have several instances on v18.0 with PHP 8.2, we'll monitor them to check for issues. @atm-quentin

PR #26393, opened 2023-10-30
FIX Cast int $qty to fix dispatch fatal error

c3do:
Fix for: PHP Fatal error:  Uncaught TypeError: Unsupported operand types: string * string in /[...]/htdocs/product/stock/class/mouvementstock.class.php:524

### DIFF
--- a/htdocs/fourn/commande/dispatch.php
+++ b/htdocs/fourn/commande/dispatch.php
@@ -258,7 +258,7 @@ if ($action == 'dispatch' && $permissiontoreceive) {
 			}
 
 			// We ask to move a qty
-			if (GETPOST($qty) != 0) {
+			if (GETPOSTINT($qty) != 0) {
 				if (!(GETPOST($ent, 'int') > 0)) {
 					dol_syslog('No dispatch for line '.$key.' as no warehouse was chosen.');
 					$text = $langs->transnoentities('Warehouse').', '.$langs->transnoentities('Line').' '.($numline);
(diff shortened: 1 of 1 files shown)


htdocs/fourn/commande/dispatch.php
```diff
@@ -258,7 +258,7 @@
 			}
 
 			// We ask to move a qty
-			if (GETPOST($qty) != 0) {
+			if (GETPOSTINT($qty) != 0) {
```
eldy, line 261: GETPOST means the entry come from a parameter so from a user input. GETPOSTINT will not be enough
We must converte a text input into a numeric value.
So good fix should be
`if (price2num(GETPOST($qty), 'MS') != 0) {`

Can you try with this ?

htdocs/fourn/commande/dispatch.php
```diff
@@ -267,7 +267,7 @@
 				}
 
 				if (!$error) {
-					$result = $object->dispatchProduct($user, GETPOST($prod, 'int'), GETPOST($qty), GETPOST($ent, 'int'), GETPOST($pu), GETPOST('comment'), '', '', '', GETPOST($fk_commandefourndet, 'int'), $notrigger);
+					$result = $object->dispatchProduct($user, GETPOST($prod, 'int'), GETPOSTINT($qty), GETPOST($ent, 'int'), GETPOST($pu), GETPOST('comment'), '', '', '', GETPOST($fk_commandefourndet, 'int'), $notrigger);
```
eldy, line 270: Same here. Clear data with price2num method and second parameter  'MS' because value must be rounded for a stock.

Issue #26384, opened 2023-10-27
Feature Request - Quarantine warehouse


AshleyButcher:
### Feature Request

It would be useful to be able to mark a warehouse as a quarantine warehouse. This would allow stock to be recorded in inventory but be unavailable for order, without having to implement stock lotting and inventory quality recording.

A quarantine warehouse should function exactly the same as a normal warehouse, only with a boolean flag "Quarantine" set to "true" which indicates that this stock is not to be proposed on a sales order when preparing a shipment.

### Use case

Dolibarr doesn't have any quality control processes, which makes rigidly implementing GxP difficult, but not impossible.

Helpful would be the ability to control the quality of the stock. At the very least, damaged stock or returned stock should be in a quarantined status pending a quality control. To do this simply, stock can enter a quarantine warehouse where it is isolated from the main stock and cannot be sold by accident.

Implementing a quarantine warehouse is a simple but very effective way of adding a small level quality control.

### **Scenario 1:**
 - Customer returns an item they no longer want
 - Item should be returned to stock in the quarantine warehouse pending examination to answer if it can be resold, and if it can be resold as what quality (New/A-stock, Damaged/Used/B-stock, etc.)
 - If the item can be resold, it can be moved to a main warehouse

### **Scenario 2:**
 - Items have been damaged in the warehouse, i.e. water damage
 - Items should be then moved to the quarantine warehouse pending a quality examination and decision to destroy, gift, rework, repurpose, etc.

### **Scenario 3:**
 - Items have expired
 - To avoid accidentally shipping, items should be moved to a quarantine space pending destruction, donation, etc.

### Suggested implementation

The simplest implemention is as follows:

1. An additional "warehouse type" is added to the "Warehouse" settings with the possibility for two values, "Normal" (existing behaviour) and "Quarantine" (modified behavior)
2. When creating an order shipment, any stock in a warehouse marked as "quarantine" is not considered as available
3. Optional: If an order shipment contains a line with the warehouse set to a quarantine warehouse, display a warning that stock is emanating from a quarantine warehouse.

### Suggested steps

1. Add a "warehouse type" field to the warehouse record (`llx_entrepot`) of integer type with values `0` = "Normal" and `1` = "Quarantine" (maybe this can be extended in future with 'transit', but that's a user story for another day)
[...]

PR #26398, opened 2023-10-31
New allow modify stock_theorique, not only replace by hook.

fappels:
# New 
Not only replace virtual stock calculation but also allow modify virtual stock by some additional stat.
For example #26384 reduce by quarantine stock.


AshleyButcher:
I don't think virtual stock should be used for quarantine stock. This is a bad idea and breaks the inventory valuation model. This is not a good use case.

fappels:
@Ashley-Butcher Is only hook for external module. Made this PR after reading comments on your issue to see if possible to modify virtual stock calculation by hook. Not really related to your issue, only gave idea. Before hook replace all calculation, for me make no sense, is same as override load_virtual_stock function.

PR #26428, opened 2023-11-01
FIX #24367 - membership duration is the default duration adhtype

### LINKED ISSUE #24367: No automatic end date of membership
### Bug

The end date of membership is not calculated accordingly to the membership duration.

### Environment Version

16.0.5

### Environment OS

Linux 5.15.0-69 (Kubuntu)

### Environment Web server

Apache/2.4.52

### Environment PHP

8.1.2-1ubuntu2

### Environment Database

MariaDB 5.5.5-10.6.12-MariaDB-0ubuntu0.22.04.1

### Environment URL(s)

_No response_

In the member module, I created several member types : one for each payment frequency (monthly, quarterly, etc.), and I filled the _duration_ field accordingly.
When I register a contribution, the _end date of membership_ is supposed to be calculated automatically when left empty. But instead of this, it registers the membership for one year, no matter the membership duration.

Happens every time.

### Attached files

image

image

image

daraelmin:


daraelmin:
I did not provide a default end date in the "end date" box, because it is too complicated to modify it on the fly when the start date is changed.
However, the default duration defined in the membership type is applied if no end date is entered.
However, a little help text should probably be added so that the user knows that if they leave the box empty, the default duration applies.
To be done in a future version.

PR #26438, opened 2023-11-02
FIX trad menu left recruitment

atm-jonathan:
# FIX|Fix #[*add key trad*]
Add trad 'Recruitment' > 'Recrutement' menu left module recruitment


eldy:
Fix on translation mus be done in transifex.
Pr on language files not en_US are useless as erased by transifex sync.

Issue #19501, opened 2021-11-26
Add hook's in dolreceiptprinter.php


catrielr:
# Instructions
*The idea is to add 2 hooks to be able to add more fields and functions to the receiptsprinters.*

# Feature Request

in ../core/class/dolreceiptprinter.class.php: 
## Hook for ADD template variables

`		$this->template = str_replace('{dol_value_vendor_firstname}', $user->firstname, $this->template);
		$this->template = str_replace('{dol_value_vendor_lastname}', $user->lastname, $this->template);
		$this->template = str_replace('{dol_value_vendor_mail}', $user->email, $this->template);`
**-->>HOOK HERE**
		`// parse template
		$this->template = str_replace("{", "<", $this->template);
		$this->template = str_replace("}", ">", $this->template);`

## Hook for ADD template format and data:

`						default:
						$this->printer->text($vals[$tplline]['tag']);
						$this->printer->text($vals[$tplline]['value']);`	
**-->>HOOK HERE**	
`			                        $this->errors[] = 'UnknowTag: &lt;'.strtolower($vals[$tplline]['tag']).'&gt;';
						$error++;
						break;
`

catrielr:
I have problems with Github, I attach file with the modifications.
I tested it and it works perfect.

I add 2 hooks inside the function: **sendToPrinter()**
**sendToPrinterBefore**  and  **sendToPrinterAfter**


```
<?php
/* Copyright (C) 2015-2021  Frédéric France     <frederic.france@netlogic.fr>
 * Copyright (C) 2020       Andreu Bisquerra    <jove@bisquerra.com>
 *
 * This program is free software; you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation; either version 3 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program. If not, see <https://www.gnu.org/licenses/>.
 * or see https://www.gnu.org/
 */

/**
 *  \file           htdocs/core/class/dolreceiptprinter.class.php
 *  \brief          Print receipt ticket on various ESC/POS printer
 */

/*
 * Tags for ticket template
 *
[...]

PR #26439, opened 2023-11-02
NEW #19501 Add two hooks in dolreceiptprinter.php

Daoud-mohamed:
# NEW
Adding two hooks to be able to add more fields and functions to the receiptsprinters.

- #19501


catrielr:
The HOOK context needs to be initialized 
Ex:

```
	public function sendToPrinter($object, $templateid, $printerid)
	{
		global $conf, $mysoc, $langs, $user;
		global $hookmanager;
	
		$hookmanager->initHooks(array('receiptprinter'));

```

PR #26441, opened 2023-11-02
Fix trad hrm

atm-jonathan:
# FIX|Fix #[*Fix trad hrm*]
Add trad missing module hrm
image


htdocs/langs/fr_FR/hrm.lang
```diff
@@ -28,6 +28,9 @@ DateEval=Date d'évaluation
 JobCard=Fiche emploi
 JobPosition=Profil de poste
 JobsPosition=Profils de poste
+JobProfile = Profil de poste
```
eldy, line 31: Pr on translation files for language that are not en_US is useless because translation files are erased by transifex translation during transifex sync.

PR #26445, opened 2023-11-02
FIX substitute project variables in invoice documents

lvessiller-opendsi:
FIX substitute project variables in invoice documents
- see FIX Add substitution variables from project into propal (PR 25656)
- related to FIX substitute project variables in document with free text #23820

If invoice is not linked to a project  : 
- substitute __PROJECT_ID__ with empty values

Example : 
image


lvessiller-opendsi:
@eldy 
I'm not sure if it's the good way to have empty values when the invoice is not linked to a project ?

After thinking, I kept substitution keys if there is no link with project (we can see where there are keys to substitute)

PR #26451, opened 2023-11-03
NEW lazy load to substitute project variables

lvessiller-opendsi:
NEW lazy load to substitute project variables
- related to FIX substitute project variables in document with free text (lazy load) #23820


lvessiller-opendsi:
I kept substitution keys if there is no link with project (we can see where there ara keys to substitute)

Issue #26225, opened 2023-10-15
Accountancy report : "from date" after "to date" - wrong year


maxoghana:
### Bug

accountancy report > Net income by years (Résultat / exercice in french)

When I click an the amont on a specific month to see the detail (ie: month of december 2023)
![image](https://github.com/Dolibarr/dolibarr/assets/63582926/5860861f-3979-4858-b2d3-22d5d852a790)

THe from date is on next year ( ie : 01/12/2024 .  expected 01/12/2023)
![image](https://github.com/Dolibarr/dolibarr/assets/63582926/83643de3-8f1c-42ae-89b4-110e8f33beb8)

### Environment Version

18.0.2

### Environment OS

LINUX (RASPBIAN 11 bullseye)

### Environment Web server

APACHE 

### Environment PHP

PHP 8.1.20

### Environment Database

MARIADB 10.11

### Expected and actual behavior

see description

### Steps to reproduce the behavior

Comptabilité > rapport > Résultat/exercice 
Cliquer sur le montant afficher pour un mois.

PR #26456, opened 2023-11-03
Fix #26050 #26225 and fiscal year report

### LINKED ISSUE #26050: In Accounting > Reporting, if I click on a month expenses, the dates are incorrectly loaded
### Bug

I want to go Accounting > reporting and click on a month to see the details of that month.

It goes to this url: /compta/resultat/clientfourn.php?year=2023&month=6&modecompta=CREANCES-DETTES

But the first date of the reporting period is a year in the future and I don't see the data

### Environment Version

18.0.1

### Environment OS

CloudLinux 7.x Linux hl109.lucushost.org 3.10.0-962.3.2.lve1.5.77.el7.x86_64 #1 SMP Mon Dec 12 07:06:14 EST 2022 x86_64

### Environment Web server

LiteSpeed

### Environment PHP

7.4.33

### Environment Database

MySQL or MariaDB 10.6.11-MariaDB-cll-lve

### Environment URL(s)

/compta/resultat/clientfourn.php?year=2023&month=6&modecompta=CREANCES-DETTES

To load the detailed expenses for the month 6 of the year 2023

- Go to Accounting > Reporting
- Click on a month's detail

### Attached files

image

sonikf:


eldy:
If fiscal year start on june and we are in april, what is the suggested range after the pr ?
Same question if fiscal yearcstart on june and we are in october ?

sonikf:
Hi @eldy 
If i understood your questions correctly 
When SOCIETE_FISCAL_MONTH_START = 6
Setting $month_current = 4 range is 1/6/2022 to 31/5/2023
and when $month_current = 10 range is 1/6/2023 to 31/5/2024

Something else i noticed is highlighted menu entry, I think it should be renamed to "By fiscal year"
FISCAL

eldy:
> mething else i noticed is highlighted menu entry, I think it should be renamed to "By fiscal year"

The group by is "Predefined group" so the good menu label should be "By predefined group"

eldy:
I have fixed the label for menu. Good label is
"By predefined group"

PR #26457, opened 2023-11-03
QUAL Clean DOM -> Hide empty box on dashboard

kevin-reo:
QUAL
Hide empty div.box on dashboard for clean DOM
Empty boxes cause bugs when parent is display as a CSS flex container

### DIFF
--- a/htdocs/takepos/index.php
+++ b/htdocs/takepos/index.php
@@ -557,7 +557,7 @@ function Floors() {
 
 function FreeZone() {
 	console.log("Open box to enter a free product");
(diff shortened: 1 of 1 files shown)


eldy:
We can't remove the empty box. This is what we get if we remove them when the number of box does not match the number of columns:
image

Not that extra box has their own css so you can disable them if necessary by your module.

kevin-reo:
Ok sorry @eldy , I have not test the IHM for this :/
Can I adjust theme css directly in Eldy theme folder for Fix this ?

eldy:
If your change does not break any feature, you can suggest it.
But what bug do you want to fix. I do not experience trouble with current code

kevin-reo:
@eldy, thanks for your answer. 
To answer you I do not have big trouble also, my aim is to optimise and clean the DOM and the layout for modernize IHM, maybe I can suggest new layout system for boxes. 

here is an exemple (you can resize your window for play with boxes)
https://codepen.io/KRIKRAK/full/XWOMpPY

PR #26459, opened 2023-11-03
QUAL group all flag images into 1 sprite file.

frederic34:
only load one png instead ~250


htdocs/core/lib/functions.lib.php
```diff
@@ -9616,8 +9616,8 @@ function picto_from_langcode($codelang, $moreatt = '', $notitlealt = 0)
 		$tmparray = explode('_', $codelang);
 		$flagImage = empty($tmparray[1]) ? $tmparray[0] : $tmparray[1];
 	}
-
-	return img_picto_common($codelang, 'flags/'.strtolower($flagImage).'.png', $moreatt, 0, $notitlealt);
+	return '<span class="saturatemedium paddingrightonly flag-sprite '.strtolower($flagImage).'"></span>';
+	// return img_picto_common($codelang, 'flags/'.strtolower($flagImage).'.png', $moreatt, 0, $notitlealt);
```
frederic34, line 9620: need to adapt here...

hansemschnokeloch:
Hi,
we could use emoji flags instead of png images : 🇫🇷, 🇬🇧, ..., zero image to load !

The following php code converts iso code to corresponding emoji (php 7.4 min)
```php
    public function country2flag(string $countryCode): string
    {
        return (string) preg_replace_callback(
            '/./',
            static fn (array $letter) => mb_chr(ord($letter[0]) % 32 + 0x1F1E5),
            $countryCode
        );
    }
```

frederic34:
@hansemschnokeloch the result is depending of the os and browser with your solution, the sprite in this PR is generated from the png found in themes/common/flags

PR #26462, opened 2023-11-04
Fix Bad Behaviour using custom constant STOCK_UPDATE_AWP_EVEN_WHEN_ENTRY_PRICE_IS_NULL

ibuiv:
# FIX Bad Behaviour with custom constant STOCK_UPDATE_AWP_EVEN_WHEN_ENTRY_PRICE_IS_NULL

## Unwanted effect
The stock movements generated by reopen or set to draft a customer order/invoice (depending of stock configuration) cause the AWP to decrease tending to 0.

## Solution found
Added a filter to trigger the AWP calculation with price == 0.
Now, constant STOCK_UPDATE_AWP_EVEN_WHEN_ENTRY_PRICE_IS_NULL must be specified AND the origin_type for the stock_movement must be supplier_order or supplier_invoice. 

P.S. : I'll apply the same fix for the Multicompany PMP per entity Trigger

### DIFF
--- a/htdocs/product/stock/class/mouvementstock.class.php
+++ b/htdocs/product/stock/class/mouvementstock.class.php
@@ -516,7 +516,7 @@ class MouvementStock extends CommonObject
 					// After a stock increase
 					// Note: PMP is calculated on stock input only (type of movement = 0 or 3). If type == 0 or 3, qty should be > 0.
 					// Note: Price should always be >0 or 0. PMP should be always >0 (calculated on input)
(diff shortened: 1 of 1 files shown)


eldy:
Can you include change for trigger in the same PR (the calculation must be consistent between different path that make the calculation)

ibuiv:
Hi, 
The trigger modified is not present in Dolibarr but in the Multicompany module. 
That's why I cannot update it here. 

Here is a screenshot from Multicompany repo with my merge request

image

PR #26466, opened 2023-11-05
Fix some warnings -technical debt-

Daoud-mohamed:
Fix some warnings -technical debt-


htdocs/commande/class/commande.class.php
```diff
@@ -285,6 +285,8 @@ class Commande extends CommonOrder
 	 */
 	public $online_payment_url;
 
+	public $stocks = array();
```
eldy, line 288: It seems the property ->stocks is used only in method public function stock_array() of the object Commande.
And a scan of code show that this method is never called. So i suggest to not add this property and instead to completely comment the dead method...

htdocs/compta/bank/class/account.class.php
```diff
@@ -279,6 +279,11 @@ class Account extends CommonObject
 
 
 
+	/**
+	 * @var string Error message.
+	 */
+	public $error_message;
```
eldy, line 285: We already have 
->error (for simple string) and ->errors (for an array of several error message)
After scan on ->->error_message it seems we use this at 4 places. i suggest instead to rename 
->error_message into ->error.

htdocs/compta/bank/class/paymentvarious.class.php
```diff
@@ -125,6 +125,16 @@ class PaymentVarious extends CommonObject
 	 */
 	public $fk_user_modif;
 
+	/**
+	 * @var int ID
+	 */
+	public $user_modif;
```
eldy, line 131: We already have 
$user_modification_id
into commonobject. We should instead rename user_modif into user_modification_id

htdocs/compta/bank/class/paymentvarious.class.php
```diff
@@ -125,6 +125,16 @@ class PaymentVarious extends CommonObject
...
+
+	/**
+	 * @var int timestamp
+	 */
+	public $date_modif;
```
eldy, line 136: We already have 
$date_modification
into commonobject. We should instead rename date_modification into date_modif

htdocs/compta/facture/class/facture-rec.class.php
```diff
@@ -86,6 +86,20 @@ class FactureRec extends CommonInvoice
...
 
+	/**
+	 * @var int
+	 */
+	public $fk_soc;
+
+	/**
+	 * @var double
+	 */
+	public $remise_percent;
```
eldy, line 97: There is no column remise_percent in table facturerec, so there is no reason to have such a property on this object.

htdocs/compta/facture/class/facture-rec.class.php
```diff
@@ -86,6 +86,20 @@ class FactureRec extends CommonInvoice
 	 */
 	public $titre;
 
+	/**
+	 * @var int
+	 */
+	public $fk_soc;
```
eldy, line 92: We already have socid.

htdocs/compta/facture/class/facture.class.php
```diff
@@ -255,6 +255,23 @@ class Facture extends CommonInvoice
 	 */
 	public $retained_warranty_fk_cond_reglement;
 
+	/**
+	 * @var int availabilty ID
+	 */
+	public $availability_id;
+
+	public $date_closing;
```
eldy, line 263: We already have 
$date_cloture

htdocs/compta/facture/class/facture.class.php
```diff
@@ -255,6 +255,23 @@ class Facture extends CommonInvoice
...
+
+	/**
+	 * @var double
+	 */
+	public $remise_percent;
```
eldy, line 268: No reason to have this property.

PR #26473, opened 2023-11-06
FIX query syntax for extrafields

BenjaminFlr:
When we fetch optionals, or when we add an extrafield, we should use backticks to define the field.

Indeed, it prevents from getting errors when the field as the same name as an SQL instruction.
For example, on a product, if you have an extrafield 'option', you'll get an error with MySQL 8 since OPTION is now an SQL instruction.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -6077,7 +6077,7 @@ abstract class CommonObject
 			$sql = "SELECT rowid";
 			foreach ($optionsArray as $name => $label) {
 				if (empty($extrafields->attributes[$this->table_element]['type'][$name]) || $extrafields->attributes[$this->table_element]['type'][$name] != 'separate') {
-					$sql .= ", ".$name;
+					$sql .= ", `".$name."`";
 				}
 			}
 			$sql .= " FROM ".$this->db->prefix().$table_element."_extrafields";
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -6077,7 +6077,7 @@ abstract class CommonObject
 			$sql = "SELECT rowid";
 			foreach ($optionsArray as $name => $label) {
 				if (empty($extrafields->attributes[$this->table_element]['type'][$name]) || $extrafields->attributes[$this->table_element]['type'][$name] != 'separate') {
-					$sql .= ", `".$name."`";
+					$sql .= ", `".$this->db->escape($name)."`";
 				}
 			}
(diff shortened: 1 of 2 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -6077,7 +6077,7 @@ public function fetch_optionals($rowid = null, $optionsArray = null)
 			$sql = "SELECT rowid";
 			foreach ($optionsArray as $name => $label) {
 				if (empty($extrafields->attributes[$this->table_element]['type'][$name]) || $extrafields->attributes[$this->table_element]['type'][$name] != 'separate') {
-					$sql .= ", ".$name;
+					$sql .= ", `".$this->db->escape($name)."`";
```
eldy, line 6080: Backtrick is not allowed in sql standard (it is a mysql extension)
If you escape data for string with db->escape, what you must use is simple quote ' (so the char thatvis escaped by the db->escape).

altairis-benjamin: The problem is that when a society uses an extrafield since the beginning, like a few years or more, it can be hard to change the field value without breaking anything.
From my point of view, it's quite hard to simply forbid reserved words, as their list evolves at each new version.

I'll keep testing and try to find the best way to fix it.

altairis-benjamin: In fact, the only way to pass through a reserved word is the use of backticks here, as `", '". $this->db->escape($name)."'"` won't make the reserved word as a field, but still as a reserved word.

The only way I found to make it work is the use of backticks. I don't know how to make it working correctly.

What do you think about the add of backticks only if we're in a MySQL configuration ? I could test `$this->db`

hregis: Maybe you can add the table name, look

https://dev.mysql.com/doc/refman/8.0/en/keywords.html

BenjaminFlr:
@eldy and @hregis : I just removed the backticks from` commonobject.class.php` to allow it to remain standard, but didn't remove it from `mysqli.class.php` as it's a specificity from MySQL. From my point of view, it will make debug easier without the table name.

hregis:
@altairis-benjamin in my opinion, the best solution is to rename the field in your module or for your customer !

BenjaminFlr:
> @altairis-benjamin in my opinion, the best solution is to rename the field in your module or for your customer !

From my point of view, even if I rename the field for my customer, I'll only fix once. With this PR, if this field exists for somebody else, it will keep working.

eldy:
PR not qualified due to use ob backtrick and prefix of database that are not portable and standardized code.

PR #26476, opened 2023-11-06
Update card.php

dolibit-ut:


htdocs/adherents/card.php
```diff
@@ -1969,6 +1980,13 @@ function initfieldrequired() {
 					}
 				}
 
+				// Delete Member
```
eldy, line 1983: In Dolibarr look and feel convention, button action that close or delete are always at end of list

dolibit-ut: okay

eldy:

Issue #26262, opened 2023-10-18
Phan static checks - merits attention


mdeweerd:
### Bug

I've added phan analysis support in [mdeweerd/test/phan](https://github.com/mdeweerd/Dolibarr/tree/test/phan).

I think several messages merit attention:
[phan.log](https://github.com/Dolibarr/dolibarr/files/13031629/phan.log)

```text
htdocs\supplier_proposal\card.php:1342 PhanUndeclaredFunction Call to undeclared function \in_numeric() (Did you mean \is_numeric())
htdocs\public\onlinesign\newonlinesign.php:183 PhanUndeclaredMagicConstant Reference to magic constant __METHOD__ that is undeclared in the current scope: used outside of a functionlike
htdocs\product\stock\stocktransfer\class\stocktransfer.class.php:345 PhanStaticCallToNonStatic Static call to non-static method \StockTransfer::cmp defined at htdocs\product\stock\stocktransfer\class\stocktransfer.class.php:356. This is an Error in PHP 8.0+.
htdocs\comm\action\class\actioncomm.class.php:1337 PhanUndeclaredVariable Variable $db is undeclared (Did you mean $this->db)
htdocs\core\menus\standard\empty.php:108 PhanUndeclaredVariable Variable $atarget is undeclared (Did you mean $this->atarget)
htdocs\core\menus\standard\empty.php:128 PhanUndeclaredVariable Variable $atarget is undeclared (Did you mean $this->atarget)
htdocs\expedition\class\expedition.class.php:1595 PhanUndeclaredVariable Variable $line is undeclared (Did you mean $lineindex)
htdocs\expedition\class\expedition.class.php:1596 PhanUndeclaredVariable Variable $line is undeclared (Did you mean $lineindex)
htdocs\fourn\class\fournisseur.facture-rec.class.php:1081 PhanUndeclaredVariable Variable $pu_ttc is undeclared (Did you mean $pu_ht)
htdocs\fourn\class\fournisseur.facture-rec.class.php:1097 PhanUndeclaredVariable Variable $pu_ttc is undeclared (Did you mean $pu_ht)
htdocs\hrm\position.php:1328 PhanUndeclaredVariable Variable $permissiontoread is undeclared (Did you mean $permissiontoadd)
htdocs\mrp\class\mo.class.php:807 PhanUndeclaredVariable Variable $line is undeclared (Did you mean $this->line)
[...]

PR #26480, opened 2023-11-07
Qual Add phan flow

mdeweerd:
# QUAL Use phan to help detect new issues

Phan's baseline was set to ignore issues before #67f1d4345a8 .
However, new issues were introduced with latest changes (undefined variable, constant).

I suggest to add phan to the workflow to detect newly introduced issues.

See #26262 for initial reference to phan results.

### DIFF
--- a/htdocs/resource/class/dolresource.class.php
+++ b/htdocs/resource/class/dolresource.class.php
@@ -634,11 +634,14 @@ class Dolresource extends CommonObject
 		$error = 0;
 
 		// Clean parameters
+		if (!is_numeric($this->resource_id)) {
+			$this->resource_id = 0;
+		}
 		if (isset($this->resource_type)) {
 			$this->resource_type = trim($this->resource_type);
 		}
-		if (isset($this->element_id)) {
-			$this->element_id = trim($this->element_id);
+		if (!is_numeric($this->element_id)) {
+			$this->element_id = 0;
 		}
 		if (isset($this->element_type)) {
 			$this->element_type = trim($this->element_type);


mdeweerd:
The 18.0.3 preparation adds several new Phan detections:

```plaintext
htdocs/compta/facture/card-rec.php:1009 PhanParamTooMany Call with 6 arg(s) to \Form::getSelectInvoiceSubtype(int $selected = 0, string $htmlname = 'subtypeid', int $addempty = 0, int $noinfoadmin = 0, string $morecss = '') which only takes 5 arg(s) defined at htdocs/core/class/html.form.class.php:10865
htdocs/core/class/commonobject.class.php:10433 PhanUndeclaredVariable Variable $object is undeclared
htdocs/salaries/document.php:239 PhanPluginDuplicateExpressionAssignment Both sides of the assignment = are the same: $permissiontoadd
htdocs/salaries/virement_request.php:107 PhanUndeclaredConstantOfClass Reference to undeclared class constant \Salary::STATUS_DRAFT
```

mdeweerd:
I have now proposed fixes or ignore rules for most detections for the active phan checks.

Several cases related to "undefined constants" remain unfixed/not ignored (Exemple: Reference to undeclared class constant \Salary::STATUS_DRAFT) .
I opened 2 or 3 issues for some of the detections.

I also proposed changes for some cases that were integrated but possibly lost/undone.

The remaining issue could be ignored in the baseline so that Phan can be integrated in the ci flow to detect new issues on code changes.
Another PR could be created to remove ignored cases to run the checks and have the results available for future fixing.

eldy:
phan seems to return very interesting information.
Before adding it by default on github action, can you modify the script dev/tools/apstats.php to execute phan and output the brut list of error and warning into the html page generated by apstats.php ?
So we will have the report always available on url  https://doxygen.dolibarr.org/apstats

mdeweerd:
I find that phan and phpstan are complementary on several accounts, with some overlap, and phan being faster to run.

- apstats: this would be a second table below/above for phpstan?
- How should phan be installed for this? (What method do you use for phpstan here - some composer.json to update ?).
- Currently I kept the following, with is not really "Technical debt" IMHO:
  - Results that likely have a functional impact;
  - Rules that had only a few (and fixable) results;
  - Results that appeared after I created the baseline (.phan/baseline.txt).
- There is "Technical debt" in:
  - The results excluded by the baseline (.phan/baseline.txt);
  - The results excluded by excluding rules.

So there could be two extra tables:
1. A table with results that should be looked at (and fixed or excluded);
2. A table with what would be techical debt.

Also, phan could be included in the flow with a baseline that excludes all current results so that CI works.
After that another PR could be created where the baseline is restored to the current one (most are fixed but I think some fixes still need to ripple up to develop).

For apstats, we could start with the current baseline, and I could add another configuration with a few more rules and a fresh baseline.

I suppose that apstats.php is run from time to time to generate https://doxygen.dolibarr.org/apstats .

mdeweerd:
I enabled more checks on the project.

A lot of the extra messages are related to Dolibarr relying on dynamic properties (deprecated in PHP8.2) not declared in PHPDoc either.

CommonObject->$socid is an example.

mdeweerd:
With the current settings there are 7 errors and 20 warnings in the Phan report.

2 messages will be fixed by two of the outstanding PRs.

I leave the other messages to be fixed by developers taking a deeper part in the functional code.

If there is a message to ignore, let me know in a comment and I'll add the corresponding annotation to the code in the second commit of this PR.

The first commit of this PR is related to the phan configuration, the second part is related to the annotations in the code to ignore some false positives.

mdeweerd:
A rebase went wrong and closed this - I will open another one once the ci is ok again.

PR #26486, opened 2023-11-07
NEW order by batch number in product batch find all method

lvessiller-opendsi:
NEW order by batch number in product batch find all method
- when you have a lot a batches with the same warehouse and the same quantity, it's difficult to choose a specific batch without order
- add an hidden const "PRODUCTBATCH_ORDER_BY_BATCH_NUMBER" to order by batch number
- this order is after DLC and DLUO orders and after qty

@eldy
- I prefer to add an hidden const but I wonder if it's needed ?


eldy:
Seems you don't need the constant.
Hiwever, are you sure this new criteria can be usefull, because there is the criteria qty before that means this criteria will often be useless.

lvessiller-opendsi:
> Seems you don't need the constant. Hiwever, are you sure this new criteria can be usefull, because there is the criteria qty before that means this criteria will often be useless.

Yes when you have a lot a serial number in the same warehouse (all have 1 in quantity)

PR #26487, opened 2023-11-07
NEW : add hook tabContentEditThirdparty

altairis-tof:
add a hook to personnalize thirdparty card edit form while keeping actions buttons and actions management.
the use case is a company that want thirdparty card fields in different order than the core standard, for example, if he does not want to show all the fields or if he wants extrafields at other position than the end.


hregis:
@altairis-tof you can use "canvas" ! ;-)

altairis-tof:
> @altairis-tof you can use "canvas" ! ;-)

two different methods; both working ;-)

PR #26495, opened 2023-11-07
Delete check price min on ttc

atm-lena:
# FIX|Fix #

Check of the price compared with the minimum price of the product should not be donne on the price TTC.
A VAT different from the default VAT of the product can be used in the propal / order / invoice : for example a customer from another country.


eldy:
Removing the test break the feature.
What we want is: 
If the price of the product was set as a price without tax (price_base_type ='HT') what we must compare for minimum price is the price without tax 
If the price of the product was set as a price with a tax (price_base_type = 'TTC'), what we must compare is price including tax.

PR #26505, opened 2023-11-08
Fix #26504 Default value extrafield line object

### LINKED ISSUE #26504: Default value for extrafields of lines of object (propal, invoice, etc) doesn't work
### Bug

If I create an extrafield for a line of an object such as proposal, invoice, etc, and set up a default value, it doesn't work as intended.
For example : a boolean extrafield checked by default.

I'm creating this issue and I'll push my PR with the solution right away 

### Environment Version

18.0

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

V-Py:
Fix #26504 Default value extrafield line object
Handle the default value for the extrafields ofa line of an object such as proposal, invoice, etc
For example : a boolean extrafield checked by default.


PsyCrow-code (commented):
Only works for SelectList, numbers, text and boolean values.
Does not function for other type of fields

PR #26514, opened 2023-11-08
Fix some warnings -technical debt-

Daoud-mohamed:
Fix some warnings -technical debt-


htdocs/don/class/don.class.php
```diff
@@ -113,6 +113,18 @@ class Don extends CommonObject
 	public $phone;
 	public $phone_mobile;
 
+	public $country_olddata;
```
eldy, line 116: No need for this. You can remove the line 
`$this->country_olddata    = $obj->country_olddata; // deprecated`

htdocs/fichinter/class/fichinterrec.class.php
```diff
@@ -57,6 +57,13 @@ class FichinterRec extends Fichinter
 	 */
 	public $picto = 'intervention';
 
+	/**
+	 * @var string
+	 * @deprecated
+	 * @see $title
+	 */
+	public $titre;
```
eldy, line 65: Should be useless. Has been replaced with title.

PR #26516, opened 2023-11-09
Fix problems on bank transfer

SylvainLegrand:
Fix the choice problem when the supplier has more than one IBAN.
Fixes issue of multiple transfer requests from tab or mass actions.
Fix selection of bank transfer request before creating the xml file


eldy:
It seems this has been solve already on recent version.
Do you experience the troubles on v18 ?

PR #26518, opened 2023-11-09
"No_Email" Field Contact : visibility -1 to 0

atm-lena:
# FIX|Fix #
The field must not be visible : "-1" value bug => field is visible on list.


eldy:
Why this field must not be visible ? It is a property we see on the file of a contact: 
image
So why not on list ?

atm-lena:
> Why this field must not be visible ? It is a property we see on the file of a contact: image So why not on list ?

If "no_email" is not visible, field in list does not disappear : 
image

If "no_email" visible, there are 2 same column proposed on list : 
image

One works, the other doesn't.
The column that works : 
image

The column that doesn't works is "no_email" field. 

These fields are not very clear... but they create a duplicate in the list

eldy:
I see. I suggested a better fix by completely removing this field as it is no more used now. Because unsusbcribe is stored into the dedicated table llx_mailing_unsubscribe.

PR #26519, opened 2023-11-09
NEW shipment kits with dispatcher v2 (SQL)

lvessiller-opendsi:
NEW shipment kits with dispatcher v2 (SQL)
- repost #21034 for tables structure

- add "fk_product" in "expeditiondet" table (it exists since v20)

-  add "fk_parent" in "expeditiondet" table (it's for parent link of one product line to determine virtual product with children lines)


lvessiller-opendsi:
@eldy 
Do you think database structure is correct ?

eldy:
What is the goal of fk_parent for shipment ?
To manage the same feature that for order, proposal and invoice: grouping lines for subtotal ?
Or doing something else...

lvessiller-opendsi:
> What is the goal of fk_parent for shipment ? To manage the same feature that for order, proposal and invoice: grouping lines for subtotal ? Or doing something else...

The "fk_parent" is used for virtual product in a shipment line.
It's the id of parent line.
It's useful when you have virtual products and components in virtual product.
Such as : 
image

lvessiller-opendsi:
@eldy 
It's ready to merge this SQL feature

eldy:
I removed the migration line UPDATE llx_expeditiondet ...
Because when an expedition line is linked to a parent line (order line for example), the value of product to use must be the one of the parent. The fk_product can be used only if the line is linked to nothing.

eldy:
I replaced the update instruction in migration because update left join is not valid SQL. It may fails on some database or options.

@lvessiller-opendsi Thanks to check it works as expected. I replaced with 
UPDATE llx_expeditiondet as ed SET ed.fk_product = (SELECT cd.fk_product FROM llx_commandedet as cd WHERE cd.rowid = ed.fk_elementdet AND ed.element_type = 'commande') WHERE ed.fk_product IS NULL;

lvessiller-opendsi:
> I replaced the update instruction in migration because update left join is not valid SQL. It may fails on some database or options.
> 
> @lvessiller-opendsi Thanks to check it works as expected. I replaced with UPDATE llx_expeditiondet as ed SET ed.fk_product = (SELECT cd.fk_product FROM llx_commandedet as cd WHERE cd.rowid = ed.fk_elementdet AND ed.element_type = 'commande') WHERE ed.fk_product IS NULL;

For me this update is right.

PR #26525, opened 2023-11-09
FIX: more correctly parse the select part to be replaced in sql queries

marc-dll:
The way the `SELECT` part of list SQL queries are replaced to build the query that counts elements seems incomplete.

This is a problem because external modules can add more complex fields to the queries via hooks. Some do so by including columns that have calculations in their definitions. This causes fatal SQL errors.

In addition to harmonizing the existing occurrences, I propose to include in the list of "allowed" characters basic mathematical operators - + / * and ways to escape identifiers (\` in MySQL, " in PostgreSQL)

If this PR is merged, there will still be work to do in Dolibarr v16+, as there are new occurrences in those versions too (they can all be found by launching `grep --include=* -Rni '\^SELECT' [htdocs_folder]`.

### DIFF
--- a/htdocs/comm/action/list.php
+++ b/htdocs/comm/action/list.php
@@ -563,7 +563,7 @@ if (empty($conf->global->MAIN_DISABLE_FULL_SCANLIST)) {
 		$nbtotalofrecords++;
 	}*/
 	/* The fast and low memory method to get and count full list converts the sql into a sql count */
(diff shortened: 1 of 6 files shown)


eldy:
A better way to transform sql to make count has been fixed into v17. By using a var $sqlfields.
If we want to introduce a fix in lower version, it should be done by a backport of the same fix.

marc-dll:
@eldy : Done ;)

marc-dll:
Hi @eldy 

I did as you said three weeks ago, do you still have issues with this PR ?

PR #26530, opened 2023-11-10
 Fix : Use explicit JOINS on lists instead of comma join #26299 

thomas-Ngr:
Fix error when using hook printFieldListFrom in a module.

One should not use comma-joins, as stated in the MySQL manual:
"The precedence of the comma operator is less than that of INNER JOIN, CROSS JOIN, LEFT JOIN, and so on. If you mix comma joins with the other join types when there is a join condition, an error of the form Unknown column 'col_name' in 'on clause' may occur. Information about dealing with this problem is given later in this section. " [source](https://dev.mysql.com/doc/refman/8.0/en/join.html)

In a PrintFieldListFrom hook on facture list, I add this statement : `LEFT JOIN llx_element_contact ec ON ec.element_id = f.rowid` and I get this error : `Unknown column 'f.rowid' in 'on clause'`

Follow PR #26299


eldy:
Most links with table societe_commerciaux has been removed into v20 and replaced with a simple WHERE EXISTS so this PR is no more valid.

PR #26543, opened 2023-11-11
NEW: Add invoice subtype in customer template invoice

sonikf:


atm-maxime:
What is an invoice subtype ?

sonikf:
Hi @atm-maxime 

Thanks to eldy it's an umbrella term to extend Dolibarr standard invoice in order to cover more invoice types used in some countries
They are housed in this dictionary https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/install/mysql/data/llx_c_invoice_subtype.sql

For some context see [here ](https://github.com/Dolibarr/dolibarr/issues/23076) and also in my commit history

If you need more info I'll be happy to oblige
Asap I'll document this feature in the wiki

PR #26554, opened 2023-11-12
FIX #26553 Supplier invoice - Do not display the delete button for reconciled payment

### LINKED ISSUE #26553: Supplier invoice - Dolibarr shall not display a button to delete a payment when this one is reconciled
### Bug

If you reopen a paid supplier invoice, Dolibarr displays a button to delete the existing payment.

However a reconciled payment can't be deleted and wouldn't be: in that case Dolibarr would display an error message.

So maybe Dolibarr shall not display a delete button for reconciled payment in the first place.

### Environment Version

v18.0.3

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

thibdrev:
I've tried to keep it simple by creating a function that could be reused elsewhere. 

For example here:
https://github.com/Dolibarr/dolibarr/blob/b709de584c33b4562baf981ede40f2f1aeb096f9/htdocs/fourn/class/paiementfourn.class.php#L455-L465

Or here:
https://github.com/Dolibarr/dolibarr/blob/b709de584c33b4562baf981ede40f2f1aeb096f9/htdocs/fourn/paiement/card.php#L236-L241

However I'm not sure how to handle the potential errors in this function (for example if the `PaiementFourn` wasn't fetched first and `$this->bank_line` is still empty ; or if the `fetch()` on the `AccountLine` returns KO).

### DIFF
--- a/htdocs/fourn/class/paiementfourn.class.php
+++ b/htdocs/fourn/class/paiementfourn.class.php
@@ -917,4 +917,17 @@ class PaiementFourn extends Paiement
 
 		return parent::fetch_thirdparty($force_thirdparty_id);
 	}
+
+
+	/**
+	 *  Return if payment is reconciled
+	 *
+	 *  @return     boolean     True if payment is reconciled
+	 */
+	public function isReconciled()
+	{
+	    $accountline = new AccountLine($this->db);
+		$accountline->fetch($this->bank_line);
+		return $accountline->rappro;
+	}
 }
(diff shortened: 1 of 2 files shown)


eldy:
Note that we may have the same problem on customer invoice.

thibdrev:
> Note that we may have the same problem on customer invoice.

Correct. See PR #26568.

PR #26557, opened 2023-11-12
Enable by default standalone credit note and invoice subtype for Greece

sonikf:


htdocs/master.inc.php
```diff
@@ -215,33 +215,41 @@ function is_countable($c)
 
 	// We set some specific default values according to country
 
-	if ($mysoc->country_code == 'DE' && !isset($conf->global->MAIN_INVERT_SENDER_RECIPIENT)) {
+	if ($mysoc->country_code == 'DE' && !getDolGlobalInt('MAIN_INVERT_SENDER_RECIPIENT')) {
```
eldy, line 218: If MAIN_INVERT_SENDER_RECIPIENT was set to 0 into home - setup - other,the if will be true and the MAIN_INVERT_SENDER_RECIPIENT will be forced to 1... when we want it to 0.
Value must be forced only if not defined by setup. So it is a "default" value. If setup set value to 0 or 1, itmust remains to 0 or 1.

PR #26566, opened 2023-11-13
New add sort and separation for objects linked

samifilali:
# NEW
- add sort and separation for objects linked


samifilali:
@eldy

atm-thibaultf:
@eldy

atm-maxime:
@eldy

htdocs/bom/tpl/linkedobjectblock.tpl.php
```diff
@@ -77,6 +77,18 @@
 	}
 	echo '</td>';
 	echo "</tr>\n";
+
+	if (count($linkedObjectBlock) > 1 || !empty($conf->global->LINKED_OBJECTS_HAVE_ALWAYS_SUBTOTAL)) {
```
eldy, line 81: conf->global->xxx is now forbidden. Can you replace with getDolGlobalString('xxx') ?

Also it seems pr is chanching the user interface, can you provide a screenshot to show the result after the change ?

eldy (changes requested):

samifilali:
Before
Before

After
After

samifilali (commented):
Done

atm-thibaultf:
@eldy

PR #26573, opened 2023-11-13
FIX add display of an error when attempting to delete a committed transaction

samifilali:
# FIX
- add display of an error when attempting to delete a committed transaction


htdocs/accountancy/bookkeeping/listbyaccount.php
```diff
@@ -417,7 +417,7 @@
 		if (!$error) {
 			foreach ($toselect as $toselectid) {
 				$result = $object->fetch($toselectid);
-				if ($result > 0 && (!isset($object->date_validation) || $object->date_validation === '')) {
+				if ($result > 0 && !isset($object->date_validation)) {
```
eldy, line 420: Why removing the or on $object->date_validation === ''

PR #26588, opened 2023-11-14
QUAL: feature parity between vendor and customer template invoice

sonikf:


sonikf:
@eldy 
Can you review please? Left behind from v 19.0.0

Issue #26416, opened 2023-10-31
Bug: BOM_SUB_BOM blank page


dolibarr95:
Hello !
# Instructions
Try to use BOM_SUB_BOM blank page

# Bug
Upgrade to Dolibarr 18.0.2 use GPAO with BOM_SUB_BOM  create a bom then the fabrication order. 
Every thing is ok
![Capture](https://github.com/Dolibarr/dolibarr/assets/24292300/518a27a1-f0fa-4975-bf40-7d0c9dd5d45e)

validate and then blank page (url -> mo_card.php?id=0)
![Capture](https://github.com/Dolibarr/dolibarr/assets/24292300/eee5261c-6008-40f1-90e4-4a89f1acae87)

## Environment
- **Version**: 18.0.2
- **OS**: Linux
- **Web server**: Apache
- **PHP**: litespeed 7.2.34
- **Database**: MySQL or MariaDB 5.5.5-10.6.15-MariaDB
- **URL(s)**: /dolibarr18/mrp/mo_card.php?id=0

[*Files*]

lamrani002:
I reproduce the same process and it works for me

dolibarr95:
Hello
@lamrani002 tks for this feedback are you running Dolibarr 18.0.2 ?

lamrani002:
Hello, yes

dolibarr95:
Ok tks I will re-install a fresh 18.0.2 and test again
tks ;)

lamrani002:
no problem

dolibarr95:
Test on a new Dolibarr 18.0.2
BOM_SUB_BOM activated still same bug when I select "Générer OF enfant"
![Capture](https://github.com/Dolibarr/dolibarr/assets/24292300/702f4372-f6d4-4fd2-91b7-57ccca9a22a7)

dolibarr95:
here is my log 
```
2023-11-09 14:04:43 NOTICE  myip    --- Access to POST /18TST/mrp/mo_card.php - action=add, massaction=
2023-11-09 14:04:43 DEBUG   myip    sql=SELECT transkey, transvalue FROM llx_overwrite_trans where (lang='fr_FR' OR lang IS NULL) AND entity IN (0, 0,1) ORDER BY lang DESC
[...]

PR #26524, opened 2023-11-09
FIX header location when creating sub MO is not working

atm-thibaultf:
# FIX|Fix #26416 Bug: BOM_SUB_BOM blank page
Fix header location when creating sub MO is not working because $moline is an empty object, we need to use new $tmpmoline

PR #26590, opened 2023-11-14
FIX header location when creating sub MO is not working

atm-thibaultf:
FIX|Fix https://github.com/Dolibarr/dolibarr/issues/26416 Bug: BOM_SUB_BOM blank page
Fix header location when creating sub MO is not working because $moline is an empty object, we need to use new $tmpmoline (remplacement of #26524)

### DIFF
--- a/htdocs/mrp/mo_card.php
+++ b/htdocs/mrp/mo_card.php
@@ -168,7 +168,7 @@ if (empty($reshook)) {
 			$res = $object->add_object_linked('mo', $mo_parent->id);
 		}
 
-		header("Location: ".dol_buildpath('/mrp/mo_card.php?id='.((int) $moline->fk_mo), 1));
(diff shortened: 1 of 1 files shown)


htdocs/mrp/mo_card.php
```diff
@@ -168,7 +168,7 @@
 			$res = $object->add_object_linked('mo', $mo_parent->id);
 		}
 
-		header("Location: ".dol_buildpath('/mrp/mo_card.php?id='.((int) $moline->fk_mo), 1));
+		header("Location: ".dol_buildpath('/mrp/mo_card.php?id='.((int) $tmpmoline->fk_mo), 1));
```
eldy, line 171: $tmpmoline is a variable inside a loop. 
This var should not be used outside the loop.

PR #26604, opened 2023-11-15
Invoice - payment list - Manage Checkbox on left column

aspangaro:


altairis-tof:
@aspangaro i think you'd better use getDolGlobalInt() especially if you test it with not (!) like in the last occurence of your PR.

PR #26609, opened 2023-11-15
NEW Suggest root path configurations in conf.php.example

mdeweerd:
# New Suggest root path configurations in conf.php.example
The suggested dynamic expression makes moving the directory around easier and makes it easier to provide the root paths.

### DIFF
--- a/htdocs/conf/conf.php.example
+++ b/htdocs/conf/conf.php.example
@@ -36,15 +36,16 @@ $dolibarr_main_url_root='';
 // Examples:
 // $dolibarr_main_document_root='/var/www/dolibarr/htdocs';
 // $dolibarr_main_document_root='C:/My web sites/dolibarr/htdocs';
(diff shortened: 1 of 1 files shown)


htdocs/conf/conf.php.example
```diff
@@ -37,7 +37,7 @@ $dolibarr_main_url_root='';
 // $dolibarr_main_document_root='/var/www/dolibarr/htdocs';
 // $dolibarr_main_document_root='C:/My web sites/dolibarr/htdocs';
 //
-$dolibarr_main_document_root='';
+$dolibarr_main_document_root=realpath(__DIR__.'/..');
```
eldy, line 40: This let think that we can keep value realpath(__DIR__.'/..'); as value.
I am not sure that such value will work with some special setup (like when we use diralias of apache instead of virtual host).

mdeweerd: I moved these suggestions in the comments, and fixed some spelling and end of line whitespaces.

PR #26623, opened 2023-11-16
Solution to restore a massive stock movement error.

lamrani002:
# NEW|New functionality for reverse mouvement stock with massive actions and adding function for update label and code_inventory in table


jyhere:
Sorry to dig up this PR... We've had some feedback on this feature which, at first glance, seems to be buggy. However, looking at the code, I have the feeling this is the intended behavior : the action just tags selected movements with a “Annulation movement ID XX” label, but not add any movement to correct stock?
What's the point of this feature?
First you have to make all the corrective movements manually, then select them to classify them as cancellations?
If so, isn't the label wrong? The id used in the label should be the id of the initial movement, the ID of the correction movement, right ?

PR #26637, opened 2023-11-16
FIX MAIN_LANG_DEFAULT was no longer saved during installation

hregis:
[*Long description*]


hregis:
@eldy ping

PR #26642, opened 2023-11-16
increase phpstan level

frederic34:
0:  basic checks, unknown classes, unknown functions, unknown methods called on $this, wrong number of arguments passed to those methods and functions, always undefined variables

https://phpstan.org/user-guide/rule-levels

@eldy is is possible to run phpstan report more often these days

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,6 +1,6 @@
 parameters:
 	customRulesetUsed: true
-	level: 0
+	level: 1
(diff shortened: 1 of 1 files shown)


eldy:
No way to execute more often the tool during the devcamp (no access to the cron task from the hostel)

frederic34:
ok, I will see results tomorrow

PR #26646, opened 2023-11-16
fix undef line and action

rycks:
fix undef vars on mo


htdocs/mrp/class/mo.class.php
```diff
@@ -809,7 +809,7 @@ public function updateProduction(User $user, $notrigger = true)
 					if ($moLine->role == 'toconsume' || $moLine->role == 'toproduce') {
 						if (empty($moLine->qty_frozen)) {
 							$qty = $newQty * $moLine->qty / $oldQty;
-							$moLine->qty = price2num($qty * (!empty($line->efficiency) ? $line->efficiency : 1 ), 'MS'); // Calculate with Qty to produce and efficiency
+							$moLine->qty = price2num($qty * (!empty($obj->efficiency) ? $obj->efficiency : 1 ), 'MS'); // Calculate with Qty to produce and efficiency
```
eldy, line 812: $obj->efficiency does not exist.
$obj is result of a sql with only field rowid.
May be it is $moLine ->efficiency ?

PR #26647, opened 2023-11-16
QUAL: Phpstan github action

frederic34:


frederic34:
@eldy the check is a little too long but errors are reported in comment

eldy:
May be problem is because the phpstan is run on all the project. No way to run it only on modified files ?

mdeweerd:
You could try to enable the cache.

See https://staabm.github.io/2023/10/21/phpstan-result-cache-gotchas.html#github-actions for instance.


Also rename phpstan.neon to phpstan.neon.dist .

mdeweerd:
Possibly also reinstate:
```
        bootstrapFiles:
                - build/phpstan/bootstrap.php
```

Some extra constants need to be defined in there to allow the main.inc.php to complete.

frederic34:
@mdeweerd thank's for the tips

mdeweerd:
Currently your are testing the cache, so it's ok to use the PR index.

However, I'ld suggest that the final version restores the cache from the PR cache if it exists and otherwise from the develop branch.

Afterwards, if on the develop branch, store to the develop branches' cache, and otherwise to the PR cache in a pull-request.

mdeweerd:
Comme c'est vraiment très long, je suggère de limiter l'analyse à qqs fichiers seulement le temps de valider la mise en place du cache.

frederic34:
image

frederic34:
@mdeweerd tu es au devcamps ?

mdeweerd:
Non, je suis juste un peu l'amélioriation de la qualité.

mdeweerd:
D'ailleurs j'ai un PR pour phan en attente: #26480 .

frederic34:
Oui je l'ai vu

frederic34:
image

frederic34:
image

mdeweerd:
La mise en place du cache a bien avancé (récupération + sauvegarde), mais il y a toujours ce message de phpstan: "Result cache not used because the cache file does not exist." - donc il manque qqchse.

Perso l'expérience que j'ai c'est avec gitlab ou j'ai la tache suivante:

```yaml
phpstan:
  interruptible: true
  rules:
    - if: $CI_PIPELINE_SOURCE == "push" || $CI_PIPELINE_SOURCE == "trigger"
  stage: test
  image:
    name: ghcr.io/phpstan/phpstan
    entrypoint: [""]
  script:
    - phpstan analyse --no-progress --error-format gitlab | tee phpstan.json
    # - phpstan analyse -vvv
  cache:
    key: phpstan-$CI_COMMIT_REF_SLUG
    paths:
      - .tmp/
      - .cache/
      - vendor/
  artifacts:
    when: always
    reports:
      codequality: phpstan.json
```

Le cache de 'vendor' n'a pas d'utilité pour ce target de mon exemple, mais peut l'avoir quand on installe des modules pour analyse et évite de réinstaller à chaque fois.

Par contre je ne vois pas '.cache' dans la config pour dolibarr et aussi j'ai .tmp/ dans ma config, pas /tmp - le tmp n'est peut-être pas utile mais je n'ai pas cherché à optimiser ce point.

frederic34:
le problème de cache vient de là
image
j'ai eu des étapes où il le retrouvait

frederic34:
image
when i get resultCache.php present, it say that cache was generated in the lest run

htdocs/core/db/DoliDB.class.php
```diff
@@ -30,7 +30,7 @@
 abstract class DoliDB implements Database
 {
 	/** @var string Force subclass to implement VERSIONMIN */
-	const VERSIONMIN=self::VERSIONMIN;
```
frederic34, line 33: @mdeweerd @eldy this make phpstan crash

frederic34:
seems the cache running, ~1 minute to run
image

htdocs/core/db/DoliDB.class.php
mdeweerd: I opened an [issue with phpstan](https://github.com/phpstan/phpstan/issues/10155) .

I guess that for know this line has to be commented so that phpstan can pass.  It kept phan happy and also helps enforce that the constant is defined in derived classes though.

mdeweerd: I also found why the cache is not in '/tmp' in my project.

I have the [tmpDir option](https://phpstan.org/config-reference#caching) set:

```yaml
parameters:
  tmpDir: .cache/phpstan
```

If you rename phpstan.neon to phpstan.neon.dist, then you can set local parameters in phpstan.neon.  So for the github action you could create phpstan.neon with the specific location for the phpstan cache directory.
That helps avoid caching the entire /tmp directory which may not only hold phpstan cache.

frederic34:
with level 9
image

htdocs/core/db/DoliDB.class.php
frederic34: @mdeweerd phpstan bug has been fixed

frederic34:
@eldy what's missing to merge ?

build/phpstan/bootstrap.php
```diff
@@ -1,11 +1,13 @@
...
+define('DOL_URL_ROOT', '/');
+define('DOL_MAIN_URL_ROOT', '/');
+define('MAIN_DB_PREFIX', 'llx_');
 
 // Load the main.inc.php file to have functions env defined
 if (! defined("NOLOGIN")) define("NOLOGIN", '1');
 if (! defined("NOHTTPSREDIRECT")) define("NOHTTPSREDIRECT", '1');
 global $conf, $langs, $user, $db;
-include_once __DIR__ . '/../../htdocs/main.inc.php';
+// include_once __DIR__ . '/../../htdocs/main.inc.php';
```
eldy, line 13: What is gain of removing call to main.inc.php ? This does not break knowledge of type of conf, user and db variable ?

frederic34: Test will never fail, until we remove `continue-on-error: true` from phpstan.yml but we can see errors

Issue #26283, opened 2023-10-20
Contract list services filter error with "expired service"


aschio:
### Bug

![image](https://github.com/Dolibarr/dolibarr/assets/1215488/ca6f14f2-5b1f-40b3-858b-3142b99a5e4b)

With filter EXPIRED, in list we see also "NOT EXPIRED". 

I suppose that sql query is not correct.

### Environment Version

Dolibarr >= 18.x

### Environment OS

all version

### Environment Web server

all version

### Environment PHP

all version

### Environment Database

all version

### Environment URL(s)

/contrat/services_list.php?leftmenu=contracts&search_status=4&filter=expired

### Expected and actual behavior

Do not show "not expired" services.

### Steps to reproduce the behavior

Click on top menu "COMMERCIAL"
Click on left menu "Expired Services"

PascalB67:
Hello, I see the same thing in 18.0.1

W1W1-M:
Hi, in dev v19 switching between : 
- `running, not expired` and `running` 
- `running, expired` and `running `

does not show the correct results. This seems to be related to your bug. 

This could be fixed with updated code to PR #21647

PR #26664, opened 2023-11-17
FIX contract service list status filter

W1W1-M:
# FIX contract service list status filter
In the current implementation of the service list, a `$filter` variable is used in pair with the `$search_status` variable in order to filter the results. In some use cases the list isn't properly filtered and the results do not correspond to the filter requested by the user:
- Using the services submenu link **Expired services**
- Switching between `running, expired`, `running, not expired` & `running` states on the service **status** filter

The ideal solution will be to eventually add a new list filter as proposed in an older PR #21647
For the moment this fix removes the `$filter` variable and makes correct use of the `$search_status` variable for list filtering. EN & FR localizations have also been updated for clearer list titles.
This PR is related to issue #26283, a fix for v18 will also be needed.


eldy:
Seems already fixed in develop.
Can you check on your side it is ok on develop. If yes a backport of what was done in v19 is better than a new fix.

W1W1-M:
> Seems already fixed in develop. Can you check on your side it is ok on develop. If yes a backport of what was done in v19 is better than a new fix.

I'm using the develop branch but I've synced just in case. I can still repeat the problem, here are the steps to reproduce it :
1. Set up some services which are **running not expired** and also **running expired**.
2. Filter list to `running, expired` using column header
3. Try filtering list back to just `running`

Try switching between different states and you will also see that the **filter** variable in `$_POST ` is not always correctly set.

Please look just a little bit more, the bug is a little subtle but it's there.

W1W1-M:
Hello @eldy,
I've retested for this bug and kept this branch up to date as it is still present in v19. 

On the contracts list, when switching the status column filter between `running, expired` and `running` or between `not expired` & `running` the results aren't correct:

- The `running` state doesn't show all `running expired` and `running not expired` services
- The column filter doesn't update correctly when switching from `running not expired` to `running`
- The `filter` variable in $POST isn't correctly updating when changing list status filter.

To reproduce the bug, a contract with multiple services in different states needs to be set up.

Thanks for your input

W1W1-M:
@eldy I've synced this PR with dev branch (now for v20.0) and checked for any merge conflicts. I've retested and can confirm this bug still exists. Thanks for any feedback.

W1W1-M:
@eldy I've synced this PR with dev branch and checked for any merge conflicts. Thanks for any feedback.

W1W1-M:
Added with 5dbecd7

PR #26680, opened 2023-11-17
NEW include fontawesome-picker (needed on new extrafields type icone=

defrance:


defrance:
@eldy : please integrate it now

eldy:
Can you show with a screenshot what the feature brings ?

defrance:
image
image
image

hregis:
@defrance ce projet n'a pas de mise à jour depuis 7 ans et archivé sur github !

eldy:
This introduce a dependency for a very small usage.
This is interesting but I think we should be able to implement a Dolibarr combo/select component that may be more resistent to depreciation without using external javascript lib, with just full CSS and few embedded js code.

PR #26682, opened 2023-11-17
FIX : duplicate hook formObjectOptions expedition card

Thomas905:
The hook is duplicated in expedition card, as it is already in extrafields_view.tpl.php:55

### DIFF
--- a/htdocs/expedition/card.php
+++ b/htdocs/expedition/card.php
@@ -2104,11 +2104,6 @@ if ($action == 'create') {
 		print '</td></tr>';
 	}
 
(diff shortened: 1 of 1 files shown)


htdocs/expedition/card.php
```diff
@@ -2104,11 +2104,6 @@
 		print '</td></tr>';
 	}
 
-	// Other attributes
-	$parameters = array('colspan' => ' colspan="3"', 'cols' => '3');
```
eldy, line 2108: Duplicate with which line ?

PR #26683, opened 2023-11-17
Finishing auto TakePOS setup with default account

andreubisquerra:
To finish auto TakePOS setup for Dolibarr 19, I liked @eldy proposal in the comments of:
https://github.com/Dolibarr/dolibarr/pull/25037
We create an account automatically without asking the user when the TakePOS module is enabled and there is no account.
It is tested, but if it is approved I will test more and find TakePOS bugs during current Devcamp.

### DIFF
--- a/htdocs/core/modules/modTakePos.class.php
+++ b/htdocs/core/modules/modTakePos.class.php
@@ -267,7 +267,7 @@ class modTakePos extends DolibarrModules
 	 */
 	public function init($options = '')
 	{
-		global $conf, $db, $langs, $user;
+		global $conf, $db, $langs, $user, $mysoc;
 		$langs->load("cashdesk");
 
 		dolibarr_set_const($db, "TAKEPOS_PRINT_METHOD", "browser", 'chaine', 0, '', $conf->entity);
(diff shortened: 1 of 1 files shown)


htdocs/core/modules/modTakePos.class.php
```diff
@@ -323,6 +323,24 @@ public function init($options = '')
...
+		//Create cash account if not exists
+		if (!getDolGlobalInt('CASHDESK_ID_BANKACCOUNT_CASH1')) {
+			require_once DOL_DOCUMENT_ROOT.'/compta/bank/class/account.class.php';
+			$cashaccount = new Account($db);
+			$cashaccount->ref = "CASH-POS";
+			$cashaccount->label = $langs->trans("DefaultCashPOSLabel");
+			$cashaccount->courant = 2;
+			$cashaccount->country_id = $mysoc->country_id ? $mysoc->country_id : 1;
+			$cashaccount->date_solde = dol_now();
+			$result = $cashaccount->create($user);
```
eldy, line 335: Before creating the account, can you add a check before with a fetch() and create only if it does not exists. Like done for the default thirdparty.

andreubisquerra:
Cutomer logic copied, is now OK @eldy ?

PR #26684, opened 2023-11-17
Fix : revalidating a shipment was modifying the ref in EXP+id

### LINKED ISSUE #26468: Shipment Sheet Edit issue
### Bug

Regenerate issue: 
1. Create a shipment sheet  and validate the shipment sheet. 
2. Back to draft and reopen the shipment sheet
3. Edit the shipment sheet and validate 
4. the Shipment Sheet reference number changed from "SH2311-0378" to "EXP568", for example. 
5. As reference number is incorrect, so I delete this shipment sheet and go back to customer order and recreate shipment sheet
6. When click "validate", the following error occurs, 
"Duplicate entry 'expedition/sending/SH2311-0378-SH22311-0378.pdf-1' for key 'llx_ecm_files.uk_ecm_files' 

### Environment Version

18.0.2

### Environment OS

Linux - Centos 8 

### Environment Web server

Apache/2.4.37 (centos)

### Environment PHP

PHP 7.2.24 

### Environment Database

mysql  Ver 8.0.26

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

atm-maxime:
Fix #26468


eldy:
A fix wasintroduced by a backport of commit 75a7d0de9b7efb8999de58712226e150973b1b7f (because it was fixed by another PR into develop)

atm-maxime:
I want to reopen this PR because I disagree with the fix made by the commit you mention. My fix respect the way modulebuilder deals with the ref in the validate function. The one you mention does not.
https://github.com/Dolibarr/dolibarr/blob/18.0/htdocs/modulebuilder/template/class/myobject.class.php#L587

rycks:
seems ok for me, that's strange (at least) to have "EXP" prefix harcoded on validate object

lvessiller-opendsi:
@atm-maxime 
You rename "numref" variable with "num" and I think it's not necessary for this fix.
I suppose you do this for quality of code "QUAL".
In this case, you can do this on "v18" branch and it will automatically be in "develop" branch.
For me, you have just conflicts to resolve.

htdocs/expedition/class/expedition.class.php
```diff
@@ -732,17 +732,17 @@ public function valid($user, $notrigger = 0)
 
 		// Define new ref
 		if (!$error && (preg_match('/^[\(]?PROV/i', $this->ref) || empty($this->ref))) { // empty should not happened, but when it occurs, the test save life
```
lvessiller-opendsi, line 734: The condition $error is not necessary here (its value is 0)
And it makes a difficulty to understand "else" condition.
No correction is needed here but it's just a remark.

eldy:
> I want to reopen this PR because I disagree with the fix made by the commit you mention. My fix respect the way modulebuilder deals with the ref in the validate function. The one you mention does not. https://github.com/Dolibarr/dolibarr/blob/18.0/htdocs/modulebuilder/template/class/myobject.class.php#L587

Your fix respect the way modulebuilder work but not the paradigm "we must change the minimum and only necessary" for a maintenance fix. So we must "go toward modulbuilder" but if, to fix a thing, only the minimal part of change from modulbuier is required, we must limit to this minimal change.

Why ? Example: Here in your suggestion, i think you introduce a bug. Because modulebuilder suppose the numbering setup is always defined (this is the case when module is fully generated). But here in v18, there is a case where numbering was not correctly activated and the fallback to the else = EXP+id was introduced to manage this (in old version the numbering module did not exists and may still not be setup in migrated instances). This fallback is probably still required today.
So for me the current fix is ok (it is minimal, it fixed the bug and it go into direction of module builder event if it is not yet  completely to be equal to modulebuilder template).

Try to do more to reach too quickly modulebuilder can lead to some regression like i think we have by removing the else = EXP+id. There is always things we can't see, this is why the rule "minimal change" always has priority on code quality or standardization when we are in a maintenance release.

Do you still experience the bug with only the fix of https://github.com/Dolibarr/dolibarr/commit/75a7d0de9b7efb8999de58712226e150973b1b7f ?

eldy:
@atm-maxime Do you still experience the bug with only the fix of https://github.com/Dolibarr/dolibarr/commit/75a7d0de9b7efb8999de58712226e150973b1b7f

PR #26686, opened 2023-11-17
Fix : right to show suppliers count on thirdparty index

Kazimir42:
It should be this right i think

### DIFF
--- a/htdocs/societe/index.php
+++ b/htdocs/societe/index.php
@@ -130,7 +130,7 @@ if ($result) {
 		if (isModEnabled('societe') && $user->hasRight('societe', 'lire') && empty($conf->global->SOCIETE_DISABLE_CUSTOMERS) && empty($conf->global->SOCIETE_DISABLE_CUSTOMERS_STATS) && ($objp->client == 1 || $objp->client == 3)) {
 			$found = 1; $third['customer']++;
 		}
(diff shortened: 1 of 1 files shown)


htdocs/societe/index.php
```diff
@@ -130,7 +130,7 @@
 		if (isModEnabled('societe') && $user->hasRight('societe', 'lire') && empty($conf->global->SOCIETE_DISABLE_CUSTOMERS) && empty($conf->global->SOCIETE_DISABLE_CUSTOMERS_STATS) && ($objp->client == 1 || $objp->client == 3)) {
 			$found = 1; $third['customer']++;
 		}
-		if (((isModEnabled('fournisseur') && $user->hasRight('fournisseur', 'facture', 'lire') && empty($conf->global->MAIN_USE_NEW_SUPPLIERMOD)) || (isModEnabled('supplier_order') && $user->hasRight('supplier_order', 'lire')) || (isModEnabled('supplier_invoice') && $user->hasRight('supplier_invoice', 'lire'))) && empty($conf->global->SOCIETE_DISABLE_SUPPLIERS_STATS) && $objp->fournisseur) {
+		if (isModEnabled('fournisseur') && $user->hasRight('fournisseur', 'lire') && empty($conf->global->SOCIETE_DISABLE_SUPPLIERS) && empty($conf->global->SOCIETE_DISABLE_SUPPLIERS_STATS) && $objp->fournisseur) {
```
eldy, line 133: We try to split in a future the module "Supplier" into "Purchase Order" and "Purchase Invoice". This is why we have 2 cases in the test with a if on MAIN_USE_NEW_SUPPLIERMOD (0 = current mode, 1 = the future with 1 module supplier_order and 1 supplier_invoice).

PR #26703, opened 2023-11-17
cron job class refactoring

moreauftheobald:
# QUAL
cron job class refactoring following new class standard and using CommonObject methode


htdocs/cron/class/cronjob.class.php
```diff
@@ -42,38 +47,111 @@ class Cronjob extends CommonObject
...
+	public $rowid;
+	public $tms;
+	public $datec;
+	public $jobtype;
 	public $label;
-
-	/**
-	 * @var string Job command
-	 */
 	public $command;
```
frederic34, line 155: @moreauftheobald what is the reason to remove all phpdoc ?

moreauftheobald: i've done the refactoring using module builder witch dont provide phpdoc on this point. so i thout it was not mandatory .
if needed i'll put them back ....

just told me.

eldy:
Too many conflict with current version, so i close pr

PR #26713, opened 2023-11-18
Fix: php stan

FHenry:


frederic34:
https://github.com/Dolibarr/dolibarr/pull/26698

frederic34:
and in another PR https://github.com/Dolibarr/dolibarr/pull/26647

I have done something for listevents.php

PR #6871, opened 2017-05-21
NEW : Add information for VAT deductibility rate in supplier price

aspangaro:
No calcul for the moment - See in V7 but begin to integrate information.

Useful to calculate the partial recovery of the VAT on some fuel or other product (France or Belgium for example)

image

PR #26732, opened 2023-11-18
NEW [Accounting] Allow grouping taxes with primary line price

BurksEngineering:
# NEW [Accounting] Allow grouping taxes with primary line price
Provides the ability for taxes paid in expense reports (ER_JOURNAL_GROUP_TAX_WITH_LINES) or purchases (AC_JOURNAL_GROUP_TAX_WITH_LINES) to get grouped with the main expense/purchase line and associated with the accounting account for the item itself. 

In non-VAT countries like USA it is more common for sales tax to get grouped with the expense before being transferred to the proper accounting account. This is in contrast to the default behavior designed around VAT which splits off any taxes paid into a separate accounting account for each type of tax, which makes it easier to recoup the VAT later.

This is a temporary solution that I use to make double entry accounting work better in USA. However, I would prefer if a 'tax deductibility' attribute existed for each defined tax rate which determined how much of the tax is allocated to the original line account vs the special tax account. This is the original proposal from @eldy in #6871 , but it was never implemented by @aspangaro . I do not think that my PHP skills or knowledge of dolibarr is sufficient for me to implement the proposal from that ticket, but this simple option/setting works well for people doing business exclusively in the USA.

Examples:
1. In USA, you buy a ream of copy paper for $10 + 5% tax = $10.50.
    * Current system: debit from purchase $10.50, credit to office supplies account $10, credit to Tax account $0.50 
        * Problem: the $0.50 will never be recuperated (paper is being consumed not resold) and visibility into the source of the tax expense is lost
    * With option enabled: debit from purchase $10.50, credit to office supplies account $10.50
2. In USA, you buy a photocopy machine for $1,000 + 5% tax = $1050.
    * Current system: debit from purchase $1050, credit to depreciable asset account $1000, credit to tax account $50
        * Problem: the basis for depreciation of the copy machine should be $1050 not $1000. The $50 sales tax should be depreciated
    * With option enabled: debit from purchase $1050, credit to depreciable asset $1050.
3. In a VAT county, you buy a ream of copy paper for $10. VAT is included and no extra tax is shown on the purchase. The entire $10 is transferred to the office expense account.


eldy:
I don't understand your analysis: When you buy a ream of copy paper for $10 + 5% tax = $10.50, you have this in Dolibarr :
For the invoice,  we have debit/credit dispatched like this:
* 10.5 for the thirdparty
* 10 for the product
* and 0.5 for the tax 
Balance = 0
When you make the payment, you have 
* 10.5 for the thirdparty (that cancel the line 10.5 created by invoice)
* 10.5 for the bank
Balance = 0

So what do you mean with "visibility into the source of the tax expense is lost" ?

Also when you say "The $50 sales tax should be depreciated", depending of what you buy, you can decide to use a sale tax that is Sale tax for "depreciable asset " and set the same accounting account than the product. So for invoice:
* 1050 for the thirdparty
* 1000 for the product on account 123456
* and 50 for the tax. Do you mean that the tax, for this case, in USA must be dispatched also into the 123456 ?
This means you do no track at all the taxes into you accountancy (all amount go into 123456) in USA ?

BurksEngineering:
Thank you for taking the time to work through my example with me! 

> For the invoice, we have debit/credit dispatched like this:
10.5 for the thirdparty
10 for the product
and 0.5 for the tax
Balance = 0
When you make the payment, you have
10.5 for the thirdparty (that cancel the line 10.5 created by invoice)
10.5 for the bank
Balance = 0

What I desire is instead to have:
* 10.5 for the thirdparty
* 10.5 for the product (which includes the tax paid)
* 0 for the tax

> So what do you mean with "visibility into the source of the tax expense is lost" ?

What I mean is that an expense account titled 'office expenses' should represent the total cost of acquiring all of the office equipment in the given year. You cannot simply multiply the value by some default tax rate (5%) because each individual purchase can be taxed at different rates depending on what state it shipped from and to.

> Do you mean that the tax, for this case, in USA must be dispatched also into the 123456 ?
This means you do no track at all the taxes into you accountancy (all amount go into 123456) in USA ?

**Yes**. In the USA sales tax is depreciated in the same accounts as assets. [Here is an open source book from a university on accounting in the USA](https://web.ung.edu/media/university-press/Principles-of-Financial-Accounting.pdf?t=1542408454385) that I use as a reference. Here is a relevant expert from page 140:
image


In the USA, sales tax is only tracked on product sold, not products bought. On products sold the tax should be moved into a liability account to be paid out to the government on a regular basis (page 200), which I believe is supported well by dolibarr currently.
image


I understand that one functional way to achieve this is to have a different sales tax entry in the tax dictionary for every single expense account (could easily be 50 nearly identical entries because there are so many expense or asset accounts). However, this turns the sales tax entry into an *accounting* decision, not a *tax rate* decision which I think is very undesirable. There are a few ways I see to solve the problem:

1. global config (this PR) to group sales tax paid with the account of the item on each line. This works well for business done exclusively in USA, and is the easiest to implement.
2. per-tax-rate field to group or not group tax paid with the account of the item on each line (or group by a percent, which was the proposal from #6871 for partial VAT refunds). This probably best represents the problem and would be my preferred choice if I knew how to implement it.
3. Add a field to the journalization page to group or not group tax paid with the account when generating all of the potential lines to journalize (or as a percent). This gives the most flexibility and does not impact the data model, but it adds an extra step to the fairly regular journalization procedure.

Can you think of other alternate solutions? 
Would you prefer to pursue a more ideal solution (2 or 3) instead of the quick and easy proposal here (1)?

PR #26733, opened 2023-11-18
NEW [Accountancy] Associate expense report lines with user

BurksEngineering:
# NEW [Accountancy] Associate expense report lines with user

Currently, when journalizing expense reports, there is no credit entry (refunding the user) associated with the debit entry (usually an expense account) for each line of the expense report. The auto-linking logic falls through all of the if statements and defaults to 'NotDefined'. This requires manual intervention for every line on every expense report.

In this proposed solution, if there is a user associated with the expense report then the user will get linked to the credit from the transaction. Depending on other settings (default accounting account for that user, default accounting account for users in general, sub-ledgers) an accounting account will then be determined for the credit half of the journal entry.

ping @aspangaro to see if there is a special edge case I am missing, and to understand the history of the 'payment_expensereport' type having $tabpay[] entries but no $tabtp[] entries.


aspangaro:
Thanks @BurksEngineering , I proposed your PR on v18 by adding to it, as the problem is the same for salaries in v18. It used to work, so it's not a new feature, it's a bug.

eldy:
Replaced by #26743

BurksEngineering:
> Thanks @BurksEngineering , I proposed your PR on v18 by adding to it, as the problem is the same for salaries in v18. It used to work, so it's not a new feature, it's a bug.

Thanks, Glad I could help highlight the issue!

PR #26742, opened 2023-11-19
Closure period - Add information on index & warning on each  journal pages

aspangaro:
Following the integration of the annual accounting closing function, additional information has been added:

image

image


eldy:
i added the option ACCOUNTANCY_FISCAL_PERIOD_MODE so we are able to manage the 2 mode:
- dispatching allowed if a fiscal period is open 
- dispatching allowed if there is no closed fiscal period

Commit 6c0c1a988191c7aa090c5eee190098edc4acbfb8

PR #26743, opened 2023-11-19
FIX Accountancy - Missing subledger user during integration of salaries/expensereport

aspangaro:
Better fix for PR #26733 of @BurksEngineering 

It worked in the past, but not now

_Example on salary, same on expense report:_

image

image

### DIFF
--- a/htdocs/accountancy/journal/bankjournal.php
+++ b/htdocs/accountancy/journal/bankjournal.php
@@ -433,6 +433,9 @@ if ($result) {
 					$paymentsalstatic->label = $links[$key]['label'];
 					$tabpay[$obj->rowid]["lib"] .= ' '.$paymentsalstatic->getNomUrl(2);
 					$tabpay[$obj->rowid]["paymentsalid"] = $paymentsalstatic->id;
+					if ($compta_user) {
+						$tabtp[$obj->rowid][$compta_user] += $amounttouse;
+					}
 
 					// This part of code is no more required. it is here to solve case where a link were missing (ith v14.0.0) and keep writing in accountancy complete.
 					// Note: A better way to fix this is to delete payment of salary and recreate it, or to fix the bookkeeping table manually after.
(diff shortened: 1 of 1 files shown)


htdocs/accountancy/journal/bankjournal.php
```diff
@@ -433,6 +433,9 @@
 					$paymentsalstatic->label = $links[$key]['label'];
 					$tabpay[$obj->rowid]["lib"] .= ' '.$paymentsalstatic->getNomUrl(2);
 					$tabpay[$obj->rowid]["paymentsalid"] = $paymentsalstatic->id;
+					if ($compta_user) {
+						$tabtp[$obj->rowid][$compta_user] += $amounttouse;
```
eldy, line 437: The $tabtp should be already filled previously by line 359 in the

elseif ($links[$key]['type'] == 'user')

but may be we don't enter into it because of line 318 ?
of because into database we don't have the line for the user ?

aspangaro: Thanks @eldy for your return.

> _The $tabtp should be already filled previously by line 359_:

I'm ok with this but no longer applicable for unknown reasons

> _may be we don't enter into it because of line 318 ?_:

This line is here to exclude the user from social contribution, if present, otherwise adding a line we don't need

> _We don't have the line_:

No, the user line is there in llx_bank_url

image

eldy: So if line "user" is here, it means that after the PR, the amount will be added twice:
- once on line 373:   $tabtp[$obj->rowid][$compta_user] += $amounttouse;
- once on the new added line 437:   $tabtp[$obj->rowid][$compta_user] += $amounttouse;

same code, but executed twice because we loop on $links that is list of all linked lines. Can you check if $links is correctly loaded with the line 77 (and not just the line 76) ?

eldy: @aspangaro A feedback on my last comment ?

PR #26745, opened 2023-11-19
Fix : doxygen : Offset non-falsy-string on array{} in empty() does not exist.

john-botella:
https://doxygen.dolibarr.org/apstats/

| File                                  | Line | Type                                                          |
|---------------------------------------|------|---------------------------------------------------------------|
| htdocs/core/class/CMailFile.class.php | 1093 | Offset non-falsy-string on array{} in empty() does not exist. |

need @frederic34 rewiew and testing because I haven't OOauth smtp api service for testing.


frederic34:
@atm-john phpstan doesnt report error with this change but I can't test functionality today (on Sunday)

frederic34:
@atm-john @atm-maxime @eldy I made test today, it works
image

PR #26747, opened 2023-11-19
fix some php stan errors

Hystepik:
fix some php stan errors


htdocs/core/modules/dons/html_generic.modules.php
```diff
@@ -175,7 +175,7 @@ private function saveFile($path, $contents)
 		$handle = fopen($path, "w");
 		fwrite($handle, $contents);
 		fclose($handle);
-		dolChmod($file);
+		dolChmod($handle);
```
eldy, line 178: dolChmod accept a filename.

htdocs/core/modules/DolibarrModules.class.php
```diff
@@ -2138,7 +2138,7 @@ public function create_dirs()
 		// phpcs:enable
 		global $langs, $conf;
 
-		$err = 0;
+		$err = 0;$name='';
```
eldy, line 2141: Set var on next line

PR #26748, opened 2023-11-19
fix: phpstan

FHenry:


htdocs/fourn/class/fournisseur.facture-rec.class.php
```diff
@@ -586,6 +586,7 @@ public function fetch($rowid, $ref = '', $ref_ext = '')
 		$sql .= ', f.fk_multicurrency, f.multicurrency_code, f.multicurrency_tx, f.multicurrency_total_ht, f.multicurrency_total_tva, f.multicurrency_total_ttc';
 		$sql .= ', f.usenewprice, f.frequency, f.unit_frequency, f.date_when, f.date_last_gen, f.nb_gen_done, f.nb_gen_max, f.auto_validate';
 		$sql .= ', f.generate_pdf';
+		$sql .= ', f.titre'; // For Backward compatibility
```
eldy, line 589: Field already in list 
f.titre as title

FHenry: @eldy Yes but we have 
https://github.com/Dolibarr/dolibarr/blob/7ac05c7549e2f6dd9d3c41882f2f96d16516e2ea/htdocs/fourn/class/fournisseur.facture-rec.class.php#L603

and $keyforref='titre'

eldy: So why don't you replace
$this->ref = $obj->$keyforref;
with
$this->ref = $obj->title;

FHenry: Yes more simple, my bad, I fix it

PR #26762, opened 2023-11-20
Update README.md

dolibit-ut:
added some new available modules


dolibit-ut:
maybe should not run Travis CI at simple textfiles like readme or logs

PR #26771, opened 2023-11-20
fix fetch vat details not working for some use cases

sonikf:


htdocs/societe/card.php
```diff
@@ -3224,7 +3224,7 @@ function GRVAT(a, u, p, myafm) {
 
   $.ajax({
 	type: "GET",
-	url: "/societe/checkvat/checkVatGr.php",
+	url: './checkvat/checkVatGr.php',
```
eldy, line 3227: To avoidb troubles with path, it is recommended to use absolute path so here
DOL_URL_ROOT.'/societe/...

sonikf: Done

PR #26784, opened 2023-11-21
Debug v19

aspangaro:


htdocs/admin/company.php
```diff
@@ -51,6 +51,16 @@
 
 $error = 0;
 
+// Logo small - Variable by default
```
eldy, line 54: To get consistent values in all app for size of images, you can use 
getDefaultImageSizes()

aspangaro: Thanks @eldy, done. I didn't know this function

PR #26787, opened 2023-11-21
NEW : add titles for wysiwig editor in eldy theme

samifilali:
# FIX
- add titles for wysiwig editor in eldy theme

### DIFF
--- a/htdocs/theme/eldy/ckeditor/config.js
+++ b/htdocs/theme/eldy/ckeditor/config.js
@@ -74,7 +74,7 @@ CKEDITOR.editorConfig = function( config )
 	 	['Maximize'],
 	 	['SpellChecker', 'Scayt'],		// 'Cut','Copy','Paste','-', are useless, can be done with right click, even on smarpthone
 	 	['Undo','Redo','-','Find','Replace'],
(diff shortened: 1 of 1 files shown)


eldy:
Looks more as a not yet provided feature than a bug. 
As this add a new feature for end user, this must be pushed into develop. Also this option may generates troubles with rest of features (like pdf generator, or function to calculate size of note content), so i recommand to add this only if a hidden constant if set.

samifilali:
I can do it if necessary but this functionality was present until version 12 of dolibarr which is why I considered it as a fix

eldy:
Yes. It was removed because this options generates a lot of problem with no solution to fix them (like pdf generation that has bad pagination, function to calculate size of note content that is wrong). Using it inside notes and description of lines must be avoid. This is why i recommand to add this only if a hidden constant if set.

samifilali:
how can I do this? It's in a js file, there is no js context?

eldy:
> how can I do this? It's in a js file, there is no js context?

Right. Did not found solution to set ckeditor menu toolbar on the fly so it was reintroduced into v21.

Issue #26088, opened 2023-10-01
Deliveries are not visible out of the moment of their creation for shipments


bafbes:
### Bug

Recently, the handling of shipments deliveries has changed. Previously, for each shipment we had a “delivery” tab which listed its single associated delivery. The id associated with this tab was that of the shipment. With the probable aim of being able to associate several deliveries to a shipment, the current id of this tab is that of the delivery. The problem is that the delivery tab no longer appears, even when the shipment has several deliveries. There is also no isolated way to view deliveries out of shipments. This means that we can only see the deliveries at the time of their creation. It therefore becomes urgent to resolve this problem or to return back to the old mode of viewing shipment deliveries.

### Environment Version

18.0.1

eldy:
A fix was pushed in 18 branch 2 days ago...

Issue #26055, opened 2023-09-28
Bug: Delivery card is not associate in expedition


leninrivas:
# Bug
When you create the delivery, it is not associated with the shipment

## Environment
- **Version**: 18.0.1
- **OS**: Linux
- **Web server**: Apache
- **PHP**: litespeed 7.3.33
- **Database**: MySQL or MariaDB 5.5.5-10.3.23-MariaDB
- **URL(s)**: /expedition/card.php?id=130&save_lastsearch_values=1

eldy:
A fix was pushed in 18 branch 2 days ago...

PR #26789, opened 2023-11-21
NEW add tpl for delivery

frederic34:
@eldy tooltip is broken
image


frederic34:
perhaps it fix : 
https://github.com/Dolibarr/dolibarr/issues/26088
https://github.com/Dolibarr/dolibarr/issues/26055

PR #26795, opened 2023-11-21
fix: feature MAIN_MAIL_ADD_INLINE_IMAGES_IF_IN_MEDIAS was broken, no more image send in email body

FHenry:
in HTML of the body of an email

if MAIN_MAIL_ADD_INLINE_IMAGES_IF_IN_MEDIAS =1

 and there is <img src="/viewimage.php?modulepart... 
https://github.com/Dolibarr/dolibarr/blob/8eb2383003ced1a939c25ffb742a8cae23839fbd/htdocs/core/class/CMailFile.class.php#L265

We always go throught 
https://github.com/Dolibarr/dolibarr/blob/8eb2383003ced1a939c25ffb742a8cae23839fbd/htdocs/core/class/CMailFile.class.php#L268

as now MAIN_MAIL_ADD_INLINE_IMAGES_IF_DATA is defaulted et 1 in 
https://github.com/Dolibarr/dolibarr/blob/8eb2383003ced1a939c25ffb742a8cae23839fbd/htdocs/core/class/conf.class.php#L932

 and there is no image like <img src="data:image/ext;base64,..."
  then 
$findimg = 0

even if we found image with MAIN_MAIL_ADD_INLINE_IMAGES_IF_IN_MEDIAS


eldy:
Can you provide a step by step instruction to reproduce the bug ?

FHenry:
Set MAIN_MAIL_ADD_INLINE_IMAGES_IF_IN_MEDIAS = 1
in a mail add a image let's say image.png to the "media" with the WYSIWYG
put the image in the mail

The result HTML email should something like 

``

So when you send the email
```
			if (!empty($conf->global->MAIN_MAIL_ADD_INLINE_IMAGES_IF_IN_MEDIAS)) {	// Off by default
				// Search into the body for <img tags of links in medias files to replace them with an embedded file
				// Note because media links are public, this should be useless, except avoid blocking images with email browser.
				// This convert an embedd file with src="/viewimage.php?modulepart... into a cid link
				// TODO Exclude viewimage used for the read tracker ?
				$findimg = $this->findHtmlImages($dolibarr_main_data_root.'/medias');
			}
```

Here $findimg=1 and $this->html_images is populated

just after 

```
	if (!empty($conf->global->MAIN_MAIL_ADD_INLINE_IMAGES_IF_DATA)) {
				// Search into the body for <img src="data:image/ext;base64,..." to replace them with an embedded file
				// This convert an embedded file with src="data:image... into a cid link + attached file
				$findimg = $this->findHtmlImagesIsSrcData($upload_dir_tmp);
			}
```
As MAIN_MAIL_ADD_INLINE_IMAGES_IF_DATA is default to 1 in the conf.class.php 

In your HTML email you don't have <img src="data:image/ext;base64,..."  so 
$findimg = 0 

And after 

```
if ($findimg > 0) {
...
}
```

$filename_list is never populated 

so the picture is not in the resulted HTML of the email send

FHenry:
@eldy Could you have a look ?

eldy:
I see.
So i understand the change on line 272
`$findimg = $findimg + $this->findHtmlImagesIsSrcData($upload_dir_tmp);`

But i don't understand the change into findHtmlImages
$i = count($this->html_images);
because when calling findHtmlImages, ->html_images looks empty at first call, and the method is called only once.

Also, this fix has a risk of regression on common feature. Because it is a fix to fix a non official feature (MAIN_MAIL_ADD_INLINE_IMAGES_IF_IN_MEDIAS hidden and not yet stable, so in status "development still need to be finished"), we must push this fix into develop (in hope to have the feature stable for next version and to become official). 
We must not take risk on stable official feature when goal is to fix non official feature, so PR must be pushed on develop.

FHenry:
@eldy you'r right about
$i = count($this->html_images);
in findHtmlImages, there is no need for it as at this step the $this->html_images is empty

I repush it base on 17.0 branch

eldy (commented):
It would be great if you can paste the example of email content you used to test the fix is ok inside a phpunit file: Calling the constructor then testing the size of html_images...

FHenry:
@eldy test added and passed

FHenry:
@eldy PHPUnit test added and pass OK (both case src=/medi/... and src=base64)
image

FHenry:
The feature of MAIN_MAIL_ADD_INLINE_IMAGES_IF_IN_MEDIAS allow for exmeple, for user, when they put their in signature an image (upload in media WYSIWYG) to be send in email

FHenry:
@eldy Did you have time to check this PR ?

PR #26821, opened 2023-11-22
 compat php8.2 facture sponge

atm-saamiperdrix:
compat php8.2 facture sponge

### DIFF
--- a/htdocs/core/modules/facture/doc/pdf_sponge.modules.php
+++ b/htdocs/core/modules/facture/doc/pdf_sponge.modules.php
@@ -1640,7 +1640,7 @@ class pdf_sponge extends ModelePDFFactures
 		// Total remise
 		$total_line_remise = 0;
 		foreach ($object->lines as $i => $line) {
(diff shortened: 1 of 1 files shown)


htdocs/core/modules/facture/doc/pdf_sponge.modules.php
```diff
@@ -1640,7 +1640,7 @@ protected function drawTotalTable(&$pdf, $object, $deja_regle, $posy, $outputlan
 		// Total remise
 		$total_line_remise = 0;
 		foreach ($object->lines as $i => $line) {
-			$total_line_remise += pdfGetLineTotalDiscountAmount($object, $i, $outputlangs, 2); // TODO: add this method to core/lib/pdf.lib
+			$total_line_remise += floatval(pdfGetLineTotalDiscountAmount($object, $i, $outputlangs, 2)); // TODO: add this method to core/lib/pdf.lib
```
eldy, line 1643: The doc of pdfGetLineTotalDiscountAmount says the method already return float.
So this should be useless. May be the fix must be done into pdfGetLineTotalDiscountAmount to fix the case it does not return an int or float.

frederic34: it may return also string
image

PR #26825, opened 2023-11-23
FIX warning addLine Facture fk_prev_id and situation_percent

atm-jonathan:
# FIX|Fix #[proprety indefined]
Two parameters situation_percent and fk_prev_id of the addLine method are not defined
Except for the object  FactureLigne and SupplierInvoiceLine
In addLine() :
I've used a ternary to avoid warnings
```
$lines[$i]->situation_percent ?? '',
$lines[$i]->fk_prev_id ?? '',
```


htdocs/compta/facture/card.php
```diff
@@ -1790,8 +1790,8 @@
 											$lines[$i]->pa_ht,
 											$label,
 											$array_options,
-											$lines[$i]->situation_percent,
-											$lines[$i]->fk_prev_id,
+											$lines[$i]->situation_percent ?? '',
```
eldy, line 1793: Ok for the ??
But if not defined, you should use the "default" value expected by the function instead of ''.
So 100 for situation_percent and 0 for fk_prev_id

PR #26826, opened 2023-11-23
FIX warning param $lineID getSpecialCode is negatif

atm-jonathan:
# FIX|Fix #[value $idLine méthode getSpecialCode()]
Implement a condition to prevent the $lineId parameter from being negative in the getSpecialCode() method
The getSpecialCode() method is only executed if $object->lines[$i]->fk_parent_line is greater than 0


htdocs/core/lib/pdf.lib.php
```diff
@@ -1351,7 +1351,7 @@ function pdf_writelinedesc(&$pdf, $object, $i, $outputlangs, $w, $h, $posx, $pos
 	//if (is_object($hookmanager) && ( (isset($object->lines[$i]->product_type) && $object->lines[$i]->product_type == 9 && ! empty($object->lines[$i]->special_code)) || ! empty($object->lines[$i]->fk_parent_line) ) )
 	if (is_object($hookmanager)) {   // Old code is commented on preceding line. Reproduct this test in the pdf_xxx function if you don't want your hook to run
 		$special_code = empty($object->lines[$i]->special_code) ? '' : $object->lines[$i]->special_code;
-		if (!empty($object->lines[$i]->fk_parent_line)) {
+		if ($object->lines[$i]->fk_parent_line > 0) {
```
eldy, line 1354: In case of is not defined or null, both test are required to avoid warning with php 8.2
`if (!empty($object->lines[$i]->fk_parent_line) && $object->lines[$i]->fk_parent_line > 0) {`

PR #26849, opened 2023-11-24
QUAL PHP-CS-FIXER configuration and applied upgrade to PHP7.1 + Possiblity to fix PSR-12

mdeweerd:
# QUAL Use PHP-CS-FIXER for code upgrades

I propose a php-cs-fixer configuration which currently only upgrade the to PHP7.1 - and I disabled changes from array() to [].

I think this improves readability and performance (isset() replaced with ...??).

The configuration is contained in `.php-cs-fixer.dist.php` which is a configuration file detected by the tool so that fixing just requires:
```
.../php-cs-fixer fix
```

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -1,0 +1,30 @@
+<?php
+
+$finder = (new PhpCsFixer\Finder())
+	->in(__DIR__)
+	->exclude([
+		'htdocs/includes',
+	])
+	->notPath([
+		'vendor',
+	])
+;
+
+return (new PhpCsFixer\Config())
+	->setRules([
+		// Apply PSR-12 as per https://wiki.dolibarr.org/index.php?title=Langages_et_normes#PHP:~:text=utiliser%20est%20le-,PSR%2D12,-(https%3A//www
+		// '@PSR12' => true,  // Disabled for now to limit number of changes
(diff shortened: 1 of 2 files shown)


mdeweerd:
PHPCS failed which seems to be because the command line it generated is too long.

I tested locally, no issue:
```
 phpcbf --extensions=php --standard=dev/setup/codesniffer/ruleset.xml  htdocs/ scripts/
```

mdeweerd:
According to https://wiki.dolibarr.org/index.php?title=Langages_et_normes#PHP:~:text=pour%20configurer%20Eclipse.-,Note%204%3A,-Vous%20remarquerez%20que

The PSR12 ruleset should also be enabled and applied.
This implies even more updates, so for now this has been commented at this time.

eldy:
You must push the PR of the tool that make reengeneering into a different PR that the PR that is the result of running the tool.
Can you first push the PR making the reengeneering of code and only the tool for code fixes (not the code fix) ?

mdeweerd:
OK, I removed the file changes from the current commit.

I added a script to run it (documentation in the script).

We can add a word about it in some other location as well - you know best where that is.

mdeweerd:
Maybe it's also best to apply this from the oldest "minimum" version upwards to ensure that merges are still feasible without too much work.

PR #26852, opened 2023-11-25
Fix stock on disapproval

ibuiv:
# FIX Stock on supplier order disapproval
## Bug
If the stock configuration is set to increments on supplier order approval, no action was planned on disapproval. 
So stocks were incremented twice if the supplier order was reopened.

## Expected Behaviour
On reopen, with this stock configuration, stocks must be reversed.   

## Solution 
Check if concerned, and for each line of the suppplier_order, create a livraison which invert the recent receptions

### DIFF
--- a/htdocs/fourn/class/fournisseur.commande.class.php
+++ b/htdocs/fourn/class/fournisseur.commande.class.php
@@ -13,6 +13,7 @@
  * Copyright (C) 2018-2022  Ferran Marcet         	<fmarcet@2byte.es>
  * Copyright (C) 2021       Josep Lluís Amador      <joseplluis@lliuretic.cat>
  * Copyright (C) 2022       Gauthier VERDOL         <gauthier.verdol@atm-consulting.fr>
+ * Copyright (C) 2023       Antonin Marchal         <antonin@letempledujeu.fr>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


eldy:
Once an invoice is validated, you have 2 choices:
- approve (this may increase stock)
- disapprove. In such a case, no stock was increased, so why decreasing it ?

ibuiv:
It is not an invoice but a supplier order. 

You can validate and then approve (increasing this stock)
You can after that re open the order, because you detected a mistake for example, (setting its status to Draft directly) but the stock does not decrease with this action 
And when you'll approve it again (to set the status to approved), then this action increases the stock again.

To simplify : with this parameter set, click on REOPEN will create stocks errors.

I hope it's more clear

drissbenchekroun:
hi dear @ibuiv I am on the 18.0.3 version of Dolibarr and we have the same problem with the supplier invoice deletion, 
- if we have a supplier invoice at statute ''validated'' and we click delete, the invoice is deleted and the stock remain the same,

thanks for your support,

ibuiv:
Yes
I had several discussions with @eldy. He explained to me that the action of deleting an order/invoice is an action used to clear a record in the database. 
The best way to remove a such object is to cancel it.  

#25292 (discussion about that for example)

drissbenchekroun:
Thanks a lot dear for your relpy,

 

I checked the #25292 <https://github.com/Dolibarr/dolibarr/pull/25292>  (discussion about that for example)

All the topic is about the order deletion, but my case is about the supplier invoice deletion, (when stock is set to increase when supplier invioce is validated)

 

So, in that case the logic is :

-        If the supplier invoice is validated : stock should be increased

-        If the supplier invoice is devalidated : stock should be decreased

-        If the supplier invoice is deleted : stock should be  decreased (but in this case, at this version, nothing happens)

 

مع تحياتي
Thanks & Regards

 

Driss Benchekroun – PURE Solutions

General Manager & IT Consultant 

 ***@***.***> ***@***.***

 <http://www.puresolutions.ma/> www.puresolutions.ma

Office/Fax : +212 522 50 58 58

Mobile : +212 66 123 50 11

Skype : driss.puresolutions



Please consider the environment, don’t print if it’s not necessary,

 

 

De : antonin_tdj ***@***.*** 
Envoyé : lundi 27 novembre 2023 14:38
À : Dolibarr/dolibarr ***@***.***>
Cc : drissbenchekroun ***@***.***>; Comment ***@***.***>
Objet : Re: [Dolibarr/dolibarr] Fix stock on disapproval (PR #26852)

 

Yes
I had several discussions with @eldy <https://github.com/eldy> . He explained to me that the action of deleting an order/invoice is an action used to clear a record in the database.
The best way to remove a such object is to cancel it.

#25292 <https://github.com/Dolibarr/dolibarr/pull/25292>  (discussion about that for example)

ibuiv:
-        If the supplier invoice is deleted : stock should be  decreased (but in this case, at this version, nothing happens)

No, I think that you should never delete a validated invoice supplier. (You can remove this from your permissions)
and if you remove it when it is a draft, your stock will be ok.

drissbenchekroun:
Yes this is true, the question now is, how to remove this permission ? because, at the permission page, there is only right to remove invoice, (not remove validated invoice)

 

مع تحياتي
Thanks & Regards

 

Driss Benchekroun – PURE Solutions

General Manager & IT Consultant 

 ***@***.***> ***@***.***

 <http://www.puresolutions.ma/> www.puresolutions.ma

Office/Fax : +212 522 50 58 58

Mobile : +212 66 123 50 11

Skype : driss.puresolutions



Please consider the environment, don’t print if it’s not necessary,

 

 

De : antonin_tdj ***@***.*** 
Envoyé : mardi 28 novembre 2023 09:13
À : Dolibarr/dolibarr ***@***.***>
Cc : drissbenchekroun ***@***.***>; Comment ***@***.***>
Objet : Re: [Dolibarr/dolibarr] Fix stock on disapproval (PR #26852)

 

·           If the supplier invoice is deleted : stock should be  decreased (but in this case, at this version, nothing happens)

*	 

No, I think that you should never delete a validated invoice supplier. (You can remove this from your permissions)
and if you remove it when it is a draft, your stock will be ok.

eldy:
Permission "Delete invoice" means "Delete invoice that is validated".
If invoice is draft, no need for particular permission. Permission to create is enough.

PS: I confirm that "deleting" supplier invoice (like any action in dolibarr) is an action to "clean" the database so by default should not have effect on stock. In a common usage, we should not need to delete. We modify instead the status (cancel, abandon, ...).

drissbenchekroun:
Sorry to tell you that when I remove the permission of deleting the invoice the delete button disappears





 

And once I have the delete permission, I have the delete button :







 

So how can we deal with this problem J ?

 

 

مع تحياتي
Thanks & Regards

 

Driss Benchekroun – PURE Solutions

General Manager & IT Consultant 

 ***@***.***> ***@***.***

 <http://www.puresolutions.ma/> www.puresolutions.ma

Office/Fax : +212 522 50 58 58

Mobile : +212 66 123 50 11

Skype : driss.puresolutions



Please consider the environment, don’t print if it’s not necessary,

 

 

De : Laurent Destailleur ***@***.*** 
Envoyé : mardi 28 novembre 2023 11:37
À : Dolibarr/dolibarr ***@***.***>
Cc : drissbenchekroun ***@***.***>; Comment ***@***.***>
Objet : Re: [Dolibarr/dolibarr] Fix stock on disapproval (PR #26852)

 

Permission "Delete invoice" means "Delete invoice that is validated".
If invoice is draft, no need for particular permission. Permission to create is enough.

ibuiv:
Hi @drissbenchekroun 
I don't understand your question. Your description seems to be the correct expected behaviour.
@eldy was my previous explanation clear enough to apply this PR ?

eldy:
@drissbenchekroun ,not really related to PR, but the behavior to allow deletion of an invoice that is draft with only permission to create has been fixed with commit c534beecb81deb3280896fbe927aee0520d5f2d3

eldy:

PR #26854, opened 2023-11-25
New POS admin setting to disable sales to the default customer

ymollard:
# NEW POS admin setting to disable sales to the default customer
Added a new setting in the TakePOS admin section to disable sales to the default third party. If enabled, the POS requires the user to select a customer if (s)he did not.


ymollard:
Poke @eldy

Issue #24940, opened 2023-06-01
Error 400: invalid_request :: When trying to login with Google OAUTH


veronicacletizia:
### Bug

Hello community,
We are having an issue when trying to connect login of Doli 16.0.1 with Google.

We configured OAUTH modele as follows:
![image](https://github.com/Dolibarr/dolibarr/assets/74732581/9e1f8d02-f658-4a77-99b2-c207c22e0d52)

We configured Client ID and secret key in Google account as well as "Consent screen". We also added in the scope openid and email.
![image](https://github.com/Dolibarr/dolibarr/assets/74732581/f98f8a92-6569-49dd-a608-b48a00af960f)

We added in the conf.php
$dolibarr_main_authentication='openid,dolibarr';

We added in llx_const:
MAIN_AUTHENTICATION_OPENID_URL = https://accounts.google.com/o/oauth2/v2/auth

Then when trying to login in the application, we get the following error:
![image](https://github.com/Dolibarr/dolibarr/assets/74732581/c73b541b-17ef-428d-8816-81a464b47072)

"Error 400: invalid_request"

We tried everything but still no luck on logging in with Google. Any ideas of what we are doing wrong?

Many thanks in advance!
Vero

### Environment Version

16.0.1

### Environment OS

Ubuntu

### Environment Web server

Apache/2.4.41 (Ubuntu)

### Environment PHP

7.4.3-4ubuntu2

### Environment Database

5.5.5-10.5.16-MariaDB

PR #26858, opened 2023-11-26
FIX include oauth.lib.php, not require_once as it defines local varia…

mdeweerd:
# FIX include, not require_once a oath.lib.php which defines locals.

DOL_DOCUMENT_ROOT.'/core/lib/oauth.lib.php' defines locals.
"require_once" would make the code fail on multiple calls or when the file is included in multiple locations in the same execution.

So this must be included every time.

### DIFF
--- a/htdocs/admin/emailcollector_card.php
+++ b/htdocs/admin/emailcollector_card.php
@@ -403,7 +403,7 @@ if ($object->id > 0 && (empty($action) || ($action != 'edit' && $action != 'crea
 
 			if ($object->acces_type == 1) {
 				// Mode OAUth2 with PHP-IMAP
-				require_once DOL_DOCUMENT_ROOT.'/core/lib/oauth.lib.php'; // define $supportedoauth2array
+				include DOL_DOCUMENT_ROOT.'/core/lib/oauth.lib.php'; // define $supportedoauth2array
 				$keyforsupportedoauth2array = $object->oauth_service;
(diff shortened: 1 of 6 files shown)


mdeweerd:
May fix #24940 , ##24163 (not verified).

eldy:
Yes, it includes variable.
However it should be constant variable that should not be modified so no reason to call the file a second time.
Also it contains a function, so doing an include will generate fatal error.

mdeweerd:
Ok, I did not see the function, but this can be protected from relaod.

The value of the variable is overridden (set to empty) in several cases before the require_once.

Also this is used in different scopes, so there is the risk that this is not defined in the right execution scope.

Finally, this was also noticed by phan in a way and by changing this to include, the typing is recognized and some notifications disappear.
I'll protect the function.

eldy:
OAuth.lib is a lib.php
So it should not be used as if it was a .inc.php
So i pushed a better fix.

mdeweerd:
https://github.com/Dolibarr/dolibarr/blob/1d589fd74e29a1f872a120277b11194b59dc9e3c/htdocs/core/class/CMailFile.class.php#L925

https://github.com/Dolibarr/dolibarr/blob/1d589fd74e29a1f872a120277b11194b59dc9e3c/htdocs/core/class/CMailFile.class.php#L1080

https://github.com/Dolibarr/dolibarr/blob/1d589fd74e29a1f872a120277b11194b59dc9e3c/htdocs/emailcollector/class/emailcollector.class.php#L1135

PR #26860, opened 2023-11-26
Fix/tcpdf for PHP8+

mdeweerd:
# FIX for PHP8+

Updated tcpdf with maintenance fixes.


frederic34:
@mdeweerd do not forget https://github.com/Dolibarr/dolibarr/blob/develop/COPYRIGHT

mdeweerd:
OPen since a long time, closing.

PR #26873, opened 2023-11-27
FIX linked files when change ref issue #26844

### LINKED ISSUE #26844: BANK, Account statements : loose linked files if change Ref.
### Bug

If we change or correct the reference name of an Account statement on the Bank account, the linked files were not available anymore. But if we change back the ref for the previous name, the files will be refund.

### Environment Version

18.03

### Environment OS

_No response_

### Environment Web server

Debian GNU/Linux 8.11 (jessie)

### Environment PHP

8.1.22

### Environment Database

MariaDB 10.4.18

### Environment URL(s)

_No response_

We expect that the attached files were all the time available independently of the ref name.

1. On the 'Bank entries' make a 'reconcile' with few of your entries and use the value '20231113'
2. On the 'Account statement' open the statement '20231113' and upload an attached file ex. 'EXTRACT_20231112.pdf'
3. On the 'Account statement' edit the statement '20231113' and change the name of it for '20231112'

Now you can check back the attached file of the account statement '20231112' and there is no attached file..

But if you change back the ref for '20231113' and check again, the attached file will be back.

### Attached files

_No response_

lamrani002:
# FIX|Fix #26844 linked files when change ref by updating table


htdocs/compta/bank/line.php
```diff
@@ -257,9 +261,38 @@
...
+
+			if ($rslt) {
+				$obj = $db->fetch_object($rslt);
+				$fullpath = $obj->fullpath_orig;
+			}
+
+			$fullpathUpdated = str_replace($oldNum_rel, dol_sanitizeFileName($num_rel), $fullpath);
+
+			$sql = "UPDATE ".MAIN_DB_PREFIX."ecm_files";
+			$sql .= " SET filepath = '".($filepath)."'";
```
eldy, line 277: When adding a sql string, you must escape it 
`$sql .= " SET filepath = '".$db->escape($filepath)."'";`

PR #26878, opened 2023-11-28
Dev fix php stan commonobject

moreauftheobald:
# FIX
Fix PHP Stan Error on commonobject.class.php

Doxygene stat:
htdocs/core/class/commonobject.class.php	9569	Access to an undefined property CommonObject::$pass.
htdocs/core/class/commonobject.class.php	10121	Call to an undefined method CommonObject::fetch().
htdocs/core/class/commonobject.class.php	10129	Call to an undefined method CommonObject::delete().


htdocs/bom/class/bom.class.php
```diff
@@ -1806,6 +1785,7 @@ public function __construct(DoliDB $db)
 		global $conf, $langs;
 
 		$this->db = $db;
+		$this->setPicto('bomline');
```
eldy, line 1788: What is the goal to have this compared to old code ?

eldy:
The pr seems to contain a lot of change that is related to refactoring (not allowed during a beta) and not to fix warning (allowed during a beta)

htdocs/bom/class/bom.class.php
moreauftheobald: regarding to PHP 8.2 documentation it seems to be the best way to not overrid the property $picto of the common object line .
if in the __construct method you just call $this->picto = 'bomline' PHP consider this as on override.

moreauftheobald:
error from my self this PR mix to PR.... i close it and send a new one ;)

moreauftheobald:
See #26879

PR #26879, opened 2023-11-28
Dev fix phpstan commonobject

moreauftheobald:
# FIX| PHP Stan errors on commonobject.class.php

Doxygene Stats:
htdocs/core/class/commonobject.class.php	9569	Access to an undefined property CommonObject::$pass.
htdocs/core/class/commonobject.class.php	10121	Call to an undefined method CommonObject::fetch().
htdocs/core/class/commonobject.class.php	10129	Call to an undefined method CommonObject::delete().


htdocs/core/class/commonobject.class.php
```diff
@@ -10118,15 +10124,15 @@ public function deleteByParentField($parentId = 0, $parentField = '', $filter =
 				$error++;
 			} else {
 				while ($obj = $this->db->fetch_object($resql)) {
-					$result = $this->fetch($obj->rowid);
+					$result = $this->fetchCommon($obj->rowid);
```
eldy, line 10127: The method fetch does not exists because it is defined into parent. When a parent instance call deleteByParentField, we must delete the fetch of the parent instance.

PR #26886, opened 2023-11-28
Change lists generated by ModuleBuilder: loop on $arrayfields instead of $object->fields

atm-florianm:
# Rationale

Currently, `$arrayfields` is used to configure and select columns to be displayed on the list. However, lists generated by ModuleBuilder don't use this array to its full potential, making it harder to add columns to the list if they are not fields of the object.

This is a work in progress, an attempt to replace looping over `$object->fields` with loops over `$arrayfields` to make the two arrays look more and more similar and behave more and more similarly (of course `$object->fields` will still have specificities, the intent is only to standardize what is identical).

If my proposal is accepted, adding a new non-field column on the list could be as easy as:
```
$arrayfields['s.siren'] = [
    'label' => 'SIREN',
    'key' => 'siren',
    'checked' => 1,
    'enabled' => 1,
    'position' => 35,
    'help' => 'HelpForSirenColumn',
    'css' => 'col-siren'
];
```

I am not expecting immediate merge (I have conflicts because I started from Dolibarr 17 and even regardless of this, it would be a bad idea since I haven't tested all of my changes and may have forgotten a few things) but rather a discussion on how to do it best.


eldy:
Looping on $arrayfields instead of $object->fields seems the good idea. But this PR is too old and contains too much other change in template so it is no more possible to validate it.Can you try it but with ONLY the use of arrayfield as arry for the loop (no other refactoring, no code quality change, ...)

PR #26896, opened 2023-11-29
FIX : same displayed contact in thirdparty contact tab

atm-quentin:
# FIX : same displayed contact in thirdparty contact tab

### DIFF
--- a/htdocs/core/lib/company.lib.php
+++ b/htdocs/core/lib/company.lib.php
@@ -1117,7 +1117,7 @@ function show_contacts($conf, $langs, $db, $object, $backtopage = '', $showuserl
 	$extrafieldsobjectkey = $contactstatic->table_element;
 	include DOL_DOCUMENT_ROOT.'/core/tpl/extrafields_list_search_param.tpl.php';
 
-	$sql = "SELECT t.rowid, t.entity, t.lastname, t.firstname, t.fk_pays as country_id, t.civility, t.poste, t.phone as phone_pro, t.phone_mobile, t.phone_perso, t.fax, t.email, t.socialnetworks, t.statut, t.photo, t.fk_soc,";
+	$sql = "SELECT DISTINCT t.rowid, t.entity, t.lastname, t.firstname, t.fk_pays as country_id, t.civility, t.poste, t.phone as phone_pro, t.phone_mobile, t.phone_perso, t.fax, t.email, t.socialnetworks, t.statut, t.photo, t.fk_soc,";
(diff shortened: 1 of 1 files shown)


atm-maxime:
How do you have same contact displayed several times with this query ?

htdocs/core/lib/company.lib.php
```diff
@@ -1117,7 +1117,7 @@ function show_contacts($conf, $langs, $db, $object, $backtopage = '', $showuserl
 	$extrafieldsobjectkey = $contactstatic->table_element;
 	include DOL_DOCUMENT_ROOT.'/core/tpl/extrafields_list_search_param.tpl.php';
 
-	$sql = "SELECT t.rowid, t.entity, t.lastname, t.firstname, t.fk_pays as country_id, t.civility, t.poste, t.phone as phone_pro, t.phone_mobile, t.phone_perso, t.fax, t.email, t.socialnetworks, t.statut, t.photo, t.fk_soc,";
+	$sql = "SELECT DISTINCT t.rowid, t.entity, t.lastname, t.firstname, t.fk_pays as country_id, t.civility, t.poste, t.phone as phone_pro, t.phone_mobile, t.phone_perso, t.fax, t.email, t.socialnetworks, t.statut, t.photo, t.fk_soc,";
```
eldy, line 1120: There is no reason to have a DISTINCT here (like often in Dolibarr) since we use a select that include the t.towid and we want all lines in the table socpeople.
If you have a duplicated line, there is surely a trouble somewhere else. May be a corrupted database with duplicate data ? May be a bad join on a non 1-1 link.
What is the sql you have ?

mdeweerd: Just "scratching the surface":

I am not sure what "socpeople_extrafields", but I suppose that this is related to fields that are added to a user on top of those that are defined in Dolibarr.
So what if there is more than one extra field for a user?

The left join seems to be there to be able to filter on them:
https://github.com/Dolibarr/dolibarr/blob/e34940fca76290e099110ba4740e66a78ed8cb5b/htdocs/core/lib/company.lib.php#L1149-L1150

However, what if there is filter condition on several extra fields ?  So it seems that there needs to be a left join for each extrafield that must be filtered on and not just one left join.

mdeweerd: I was thinking that this could happen if there are two different extrafield types?

atm-quentin: no because we have a column by extrafield

PR #26899, opened 2023-11-30
php V8 warning

defrance:
appair when create from other element


frederic34:
already fixed here ? https://github.com/Dolibarr/dolibarr/pull/26712

eldy:
Yes seems duplicate so i keep first merged PR only.

PR #26911, opened 2023-12-01
Fix TakePOS SumUp connector

HLFH:
This PR fixes:
1) Add SumUp connector compatibility with SumUp app on iOS

**Explanation**

https://github.com/sumup/sumup-android-api
VS https://github.com/sumup/sumup-ios-url-scheme

The amount field is deprecated on the Android API but not on the iOS API. The iOS API does not support the total field, only the Android API does support the total field. Hopefully, the Android API supports as well the amount field that is only partially deprecated on the Android API. For maximum compatibility with both Android/iOS SumUp app, best to keep this field as "amount" rather than "total".

2) The second issue is the set amount was 0 instead of the real amount! It is now fixed.

3) TakePOS SumUp connector now has multicurrency support instead of only EUR that was hardcoded.

### DIFF
--- a/htdocs/takepos/pay.php
+++ b/htdocs/takepos/pay.php
@@ -472,6 +472,9 @@ if (!getDolGlobalInt("TAKEPOS_NUMPAD")) {
 		if (amountpayed > <?php echo $invoice->total_ttc; ?>) {
 			amountpayed = <?php echo $invoice->total_ttc; ?>;
 		}
+		if (amountpayed == 0) {
+            amountpayed = <?php echo $invoice->total_ttc; ?>;
+        }
 
 		// Starting sumup app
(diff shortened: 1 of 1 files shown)


HLFH:
And the currency should not be hardcoded to EUR.

Currencies supported by SumUp:

- EUR
- GBP
- USD
- BRL
- CLP
- PLN
- CHF
- SEK
- CZK
- NOK
- DKK
- HUF
- BGN

eldy:
@HLFH We have a sumup device at the Dolibarr foundation.
But we (and a lot of Dolibarr users) are missing documentation on how to use it with Dolibarr.
Would you be kind to create a wiki on https://wiki.dolibarr.org/ to explain how to setup and use Sumup ?

PR #26941, opened 2023-12-02
QUAL Fix several PSR-12 rules on files that can be merged up to develop without conflict

mdeweerd:
# QUAL Fix several PSR12 rules in mergable files

Applies several PSR12 rule while keeping the files mergable from 16.0 to develop.

Files that resulted in conflicts were excluded from this PR.

This PR is related to #26849

### DIFF
--- a/htdocs/accountancy/class/accountancyexport.class.php
+++ b/htdocs/accountancy/class/accountancyexport.class.php
@@ -1789,7 +1789,7 @@ class AccountancyExport
 					&& $objectLines[$aIndex + 1]->piece_num == $line->piece_num
 					&& $aIndex - 1 < $aSize
 					&& $objectLines[$aIndex - 1]->piece_num == $line->piece_num
-					) {
+			) {
 				$sammelBuchung = true;
 			}
 
--- a/htdocs/accountancy/class/accountingjournal.class.php
+++ b/htdocs/accountancy/class/accountingjournal.class.php
@@ -407,12 +407,12 @@ class AccountingJournal extends CommonObject
 				case 1: // Various Journal
 					$data = $this->getAssetData($user, $type, $date_start, $date_end, $in_bookkeeping);
 					break;
-				//              case 2: // Sells Journal
-				//              case 3: // Purchases Journal
-				//              case 4: // Bank Journal
-				//              case 5: // Expense reports Journal
-				//              case 8: // Inventory Journal
-				//              case 9: // hasnew Journal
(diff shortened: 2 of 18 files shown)


eldy:
On already releases version, no reengeering is allowed. Only PR for blocking or critical bugs are allowed as we are in a maintenance phase.
Can you push your PR in develop ?

mdeweerd:
I updated the PR to reference develop (merge, rebase).

I have kept the limitations for this PR: only the files that I was able to merge from 16.0 up to develop.

I can make another PR (after merging this one), where I apply the other rules as well (spaces around operators, upgrade to PHP7.1).  I can also do that in 2 steps: for instance, first upgrade to PHP7.1 and then apply the formatting rules.

mdeweerd:
I tried to improve the phpcs action which does not handle a lot of file changes very well - however it turns out that the action is using a library called php-codesniffer which passes the list of files to the CLI.

I did not find a similar action working only on changed/added files.

eldy:
I tried few weeks ago to find a solution to process only modified by a github action.
It was a headach.
I think i succeed in this goal into file .github/workflows/.phpcsfixer.yml.disabled
But i also tried to update the PR automatically, like stickler was doing before service was closed, but I was tired in testing new solutions to test so i abandoned...

I made a change in the tool to target v7.0 instead of 7.1. 
V18 targets 7.1 but it still works great with v7.0 and i prefer encouranging to switch to v7.1 instead of forcing.

eldy:
I tried few weeks ago to find a solution to process only modified by a github action.
It was a headach.
I think i succeed in this goal into file .github/workflows/.phpcsfixer.yml.disabled
But i also tried to update the PR automatically, like stickler was doing before service was closed, but I was tired in testing new solutions to test so i abandoned...

I made some change into php-cs-fixer so running it keep compatibility with 7.1 (if tool was ran with php8.2, rules breaking compatibility withnlower version were applied.
Files were applied.
Can you push the github action change on a different pr than the pr with modified files ?

mdeweerd:
I am testing the action on this case as I want to make sure it works with many changes.

I change it to draft.

mdeweerd:
I've rebased - the list has greatly reduced.
This can be merged - I will see to apply the fixer again on the entire codebase (not only the files that could be merged from 16.0 onwards) - and see if I can finish #26968 which is not too far from something useable I think.

mdeweerd:
Open since about 3 months, closing.

PR #26959, opened 2023-12-04
NEW : Added a configuration in workflows

samifilali:
# NEW
Added a configuration that allows an customer order to be placed in invoiced status if its total excluding tax is equal to the sum of the totals excluding tax of the linked invoices


samifilali:
@eldy

eldy:
What is difference with the existing option WORKFLOW_INVOICE_AMOUNT_CLASSIFY_BILLED_ORDER ?

samifilali:
The difference is that this configuration takes into account orders linked to several invoices (deposit invoice for example)

samifilali:
@eldy

PR #26960, opened 2023-12-04
added a missing translation that generates an error

samifilali:
# FIX
added a missing translation that generates an error when generating an enstein PDF template in Swiss German for orders


eldy (commented):
Prnto change language files is useless. It is erased by transifex sync. To get thz % symbol in a string, we must use %% in transifex translation.

PR #26968, opened 2023-12-04
Qual/new php cs workflow

mdeweerd:
I managed to setup this workflow that works on more changed files.

It still has to be validated when there is a PR from a fork.


mdeweerd:
The pre-commit workflow works pretty well, closing this.

PR #26969, opened 2023-12-04
Automatic php-cs-fixer with commit

mdeweerd:
This implements:
- php-cs-fixer, then
- phpcbf, then
- phpcs report, then
- commit.

### DIFF
--- a/.github/workflows/phpcsfixer.yaml
+++ b/.github/workflows/phpcsfixer.yaml
@@ -1,0 +1,110 @@
+---
+on: [pull_request, push]
+name: PHP_CS_FIXER
+jobs:
+  phan:
+    name: Run phpcs on modified or added files
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        name: Default checkout
+        with:
+          fetch-depth: 50
+      - name: Show variables
+        # if: false
+        env:
+          GITHUB_CONTEXT: ${{ toJSON(github) }}
+        run: |
+          echo "$GITHUB_CONTEXT"
+          echo "BASE_REF=$GITHUB_BASE_REF  HEAD_REF=$GITHUB_HEAD_REF GITHUB_REF=$GITHUB_REF GITHUB_REF_NAME=$GITHUB_REF_NAME"
+          env
+      - name: Fetch up to merge-base on push
+        uses: BioData/fetch-through-merge-base@v0
+        timeout-minutes: 3
+        if: github.event_name == 'push'
+        with:
+          base_ref: develop
+          head_ref: ${{ github.sha }}
+      - name: Sync upstream changes
+        id: sync
+        uses: aormsby/Fork-Sync-With-Upstream-action@v3.4
+        if: github.event_name == 'push'
+        with:
+          target_sync_branch: develop
+          # REQUIRED 'target_repo_token' exactly like this!
+          target_repo_token: ${{ secrets.GITHUB_TOKEN }}
+          upstream_sync_branch: develop
+          upstream_sync_repo: Dolibarr/dolibarr
+          upstream_repo_access_token: ${{ secrets.UPSTREAM_REPO_SECRET }}
+      - name: Fetch up to merge-base on pull
+        uses: BioData/fetch-through-merge-base@v0
+        timeout-minutes: 3
+        if: ${{ github.event_name == 'pull_request' && github.ref_name != 'develop' }}
+        with:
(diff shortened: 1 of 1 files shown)


mdeweerd:
It seems I can not test the PR from the fork in draft mode, moving to "Ready for Review" but not mergeable yet.

mdeweerd:
@eldy This workflow can commit on a pull action but not on a pull request action.

Maybe you have some idea to try, but at this time I do not see the way to get the information to be able to push to the source repository.

mdeweerd:
@eldy Inside the "github" context if found some "variables" that refer to the pulled repository (see "Show variables" step in the workflow log).

So there is a possibility (if the token works) to push changes back to the PR submitter branch.

Qiuestions are:
- Is that what you want to do?
- To push back, the corrections should not be applied on the merge IMHO, but on the source branch.
- To be milder, corrections could be applied only to the files that fail phpcs after the merge.
- To be even milder: corrections could be limited to phpcbf corrections, or the corrections made by php-cs-fixer could be limited to a subset that is just enough to keep phpcs happy as per the defined rules (and the fixes could be extended every time an automatic fix would help to comply with the rules set for phpcbf).


Other:
The BioData/fetch-through-merge-base action is actualy a fairly simple script that could be pasted almost as-is in a standard run step.  What do you prefer.

htdocs/user/class/user.class.php
```diff
@@ -372,16 +372,16 @@ class User extends CommonObject
...
+		'lastname' => array('type' => 'varchar(50)', 'label' => 'LastName', 'enabled' => 1, 'visible' => 1, 'notnull' => 1, 'showoncombobox' => 1, 'index' => 1, 'position' => 20, 'searchall' => 1),
+		'firstname' => array('type' => 'varchar(50)', 'label' => 'FirstName', 'enabled' => 1, 'visible' => 1, 'notnull' => 1, 'showoncombobox' => 1, 'index' => 1, 'position' => 10, 'searchall' => 1),
+		'ref_employee' => array('type' => 'varchar(50)', 'label' => 'RefEmployee', 'enabled' => 1, 'visible' => 1, 'notnull' => 1, 'showoncombobox' => 1, 'index' => 1, 'position' => 30, 'searchall' => 1),
+		'national_registration_number' => array('type' => 'varchar(50)', 'label' => 'NationalRegistrationNumber', 'enabled' => 1, 'visible' => 1, 'notnull' => 1, 'showoncombobox' => 1, 'index' => 1, 'position' => 40, 'searchall' => 1)
 	);
 
 
-	const STATUS_DISABLED = 0;
-	const STATUS_ENABLED = 1;
+	public const STATUS_DISABLED = 0;
```
eldy, line 383: This is not allowed in PHP7.0
Even if we target PHP 7.1+, there is currently nothing that force us to break compatibility with php 7.0 (Dolibarr works great with PHP 7.0).  So we have no reason to introduce compatibility breaking code for the moment.

eldy: Also PR the add dev tools to change code, must be separated from PR that is the result of code changed.

mdeweerd: See #27006 .  I'll rebase this PR afterwards, and keep the examples to show the test.

What is your feedback regarding the automatic commit?

eldy: I am fully fan of the automatic commit. We did it with stickler-ci few month ago before the service closed.
Just not succeed to do it easily for the moment with github action. 
If your PR can introduce such a feature can you push it into a separate PR:  PR for the CI engine change must be separated from the PR on code that results of running a CI or quality tool.

mdeweerd:
Theh modified PHP files were only included to test/demonstrate that the automatic correction works.

I have removed those from the PR and only the CI action.

mdeweerd:
Changed to draft to improved action on push

mdeweerd:
The action now updates the "develop" branch in the fork.

As I pull and rebase to develop locally, the github develop branch was not up-to-date and as a result the proper fork point was not found resulting in more identified differences than there really are.
So that is fixed by having the action update the branch.

Possibly I can avoid pushing it to the contributors develop branch.

In case the contributor's develop branch is not clean the develop branch will not be updated.

From what I notice, the CI will still be ok in the Dolibarr/dolibarr PR (if the CI checks pass).

The contributor can either pull to get the updates or make local changes and make a force push to his branch.

mdeweerd:
Open since a while, closing.

PR #26976, opened 2023-12-05
fix sql error in case of balance is empty

rycks:
# FIX create bank account with empty field on balance - sql error

image

Ui set default balance to 0 but in case of user remove 0 and keep empty field -> sql error.

One solution could be on card form (and balance as required field) but the bug will be back in case of other way to call -> create on that object ... then my fix is against the class.

### DIFF
--- a/htdocs/compta/bank/class/account.class.php
+++ b/htdocs/compta/bank/class/account.class.php
@@ -650,6 +650,14 @@ class Account extends CommonObject
 			return -1;
 		}
 
+		$balance = $this->balance;
+		if (empty($balance) && !empty($this->solde)) {
+			$balance = $this->solde;
+		}
+		if (empty($balance)) {
+			$balance = 0;
+		}
+
 		// Chargement librairie pour acces fonction controle RIB
 		require_once DOL_DOCUMENT_ROOT.'/core/lib/bank.lib.php';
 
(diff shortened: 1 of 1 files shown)


htdocs/compta/bank/class/account.class.php
```diff
@@ -650,6 +650,14 @@ public function create(User $user, $notrigger = 0)
 			return -1;
 		}
 
+		$balance = $this->balance;
+		if (empty($balance) && !empty($this->solde)) {
+			$balance = $this->solde;
+		}
+		if (empty($balance)) {
+			$balance = 0;
```
hansemschnokeloch, line 658: `$balance` is declared as float.
if we use `declare(strict_types=1);` in the future this code can raise an error
so you should use `$this->balance = 0.0;` instead

rycks: @hansemschnokeloch are you sure ?

```
echo (float)0.00 . "\n";
echo (float)0 . "\n";
```

hansemschnokeloch: My bad. I thought that we were in the opposite situation float assigned to int
```
declare(strict_types=1);
error_reporting(E_ALL);
class myClass
{
    public int $balance;
}

$c = new myClass();
$c->balance = 0.0; // Fatal error: Uncaught TypeError: Cannot assign float to property myClass::$balance of type int
```

PR #26980, opened 2023-12-05
FIX in_array() on key

atm-jonathan:
# FIX|Fix #[Display showdocuments() ]
$arrayofmassactions = 
array(8) { ["generate_doc"]=> string(93) "Re-générer le PDF" **["builddoc"]=> string(72)** "Fusion PDF" ["presend"]=> string(69) "Envoyer email" ["prevalidate"]=> string(79) "Valider" ["presign"]=> string(89) "Signer" ["nopresign"]=> string(90) "Refuser" ["setbilled"]=> string(112) "Classer Facturé" ["predelete"]=> string(68) "Supprimer" } 

**in_array('builddoc', $arrayofmassactions) return false** 
With this comparison I have no display 

**in_array('builddoc', array_map('trim', array_keys($arrayofmassactions)))**
With this modification, I'm doing the comparison on the keys instead of the value, to return true.
And I can access the display of showdocuments()

Before version 19, there was no condition for print showdocuments()

### DIFF
--- a/htdocs/adherents/list.php
+++ b/htdocs/adherents/list.php
@@ -1428,7 +1428,7 @@ print '</div>'."\n";
 
 print '</form>'."\n";
 
-if (in_array('builddoc', $arrayofmassactions) && ($nbtotalofrecords === '' || $nbtotalofrecords)) {
+if (in_array('builddoc', array_map('trim', array_keys($arrayofmassactions))) && ($nbtotalofrecords === '' || $nbtotalofrecords)) {
 	$hidegeneratedfilelistifempty = 1;
 	if ($massaction == 'builddoc' || $action == 'remove_file' || $show_files) {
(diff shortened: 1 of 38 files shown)


htdocs/adherents/list.php
```diff
@@ -1428,7 +1428,7 @@
 
 print '</form>'."\n";
 
-if (in_array('builddoc', $arrayofmassactions) && ($nbtotalofrecords === '' || $nbtotalofrecords)) {
+if (in_array('builddoc', array_map('trim', array_keys($arrayofmassactions))) && ($nbtotalofrecords === '' || $nbtotalofrecords)) {
```
eldy, line 1431: Why this ?
arrayofmassactions is an array filled by code. If there is an error into content, we must fix when we fill it at the source of the trouble and not later.

PR #26982, opened 2023-12-05
NEW Asset Module - make asset model creation in a single step

Daoud-mohamed:
# Make asset model creation in a single step

- Fix some bugs related to the PHP version
- Organize the process of creating and displaying an asset model on a single page
- Set the asset model to enabled by default

Navigate to “Accounting – Assets - Model assets – New”

View page:


eldy:
@aspangaro Your feedback is welcome.
If you have more small enhanced to do related to Asset module, you can ask @Daoud-mohamed

josett225:
Hi @Daoud-mohamed @aspangaro and @eldy 
Looking to dev v21, this code is not included in latest version. Does it mean it has been cancelled?

PR #26993, opened 2023-12-06
Fix reload page on customer and supplier change

SylvainLegrand:
Add the ability to preserve the project ID if found

### DIFF
--- a/htdocs/fourn/facture/card.php
+++ b/htdocs/fourn/facture/card.php
@@ -2287,9 +2287,15 @@ if ($action == 'create') {
 						$("#socid").change(function() {
 							console.log("We have changed the company - Reload page");
 							// reload page
-							$("input[name=action]").val("create");
-							$("input[name=changecompany]").val("1");
-							$("form[name=add]").submit();
+							//$("input[name=action]").val("create");
+							//$("input[name=changecompany]").val("1");
+							//$("form[name=add]").submit();
+							var socid = $(this).val();
+							var projectid = "";
+							if ($(\'#projectid\').val() > 0) {
+								projectid = "&projectid="+$(\'#projectid\').val();
+							}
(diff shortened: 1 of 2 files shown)


htdocs/fourn/facture/card.php
```diff
@@ -2287,9 +2287,15 @@
...
-							$("form[name=add]").submit();
+							//$("input[name=action]").val("create");
+							//$("input[name=changecompany]").val("1");
+							//$("form[name=add]").submit();
+							var socid = $(this).val();
+							var projectid = "";
+							if ($(\'#projectid\').val() > 0) {
+								projectid = "&projectid="+$(\'#projectid\').val();
+							}
+							window.location.href = "'.$_SERVER["PHP_SELF"].'?action=create"+projectid+"&socid="+socid;
```
eldy, line 2298: If you reload this way, you will make a GET when we need a POST to not lose all fields already modified.
What is wrong with old code ?

SylvainLegrand:
1. The old code doesn't work (payment term and method are not updated) !
2. If you create a new customer invoice from a open project (no specific customer), after the reload you will loose the project number !
3.  If you create a new supplier invoice from a project, after the reload you will loose the project number !

Tests made on fresh install.

eldy:
> 1. The old code doesn't work (payment term and method are not updated) !
> 2. If you create a new customer invoice from a open project (no specific customer), after the reload you will loose the project number !
> 3. If you create a new supplier invoice from a project, after the reload you will loose the project number !
> 
> Tests made on fresh install.

So there is a trouble in the code the generate the form. The code must retreive the lost parameter with GETPOST if GETPOSTISSET is defined. 

I did a test on develop branch and I do not experience trouble. The project ref is not lost on my side after the reload POST.  I click on the "Create invoice" button from a project. After selecting a thirdparty, the post is sent and all fields are kept (this is the goal of using a POST), project, but also payment term. Did i miss something in my test process ?

PR #26994, opened 2023-12-06
FIX missing accent characters in contact name (backport v17)

W1W1-M:
# FIX missing accent characters in contact name
On generation of various PDF (cyan, eratosthene, ...), if the contact name contains accents they aren't properly outputed in the sender part of the document. This fix adds the proper arguments to the `dol_escape_htmltag` method used in the `getFullName` method of the `CommonObject` in order to not escape special characters. 

This was already fixed in V18+. If I'm not mistaken the `getFullName` method was moved to the User class and stopped making use of the `dol_escape_htmltag` method.


W1W1-M:
@eldy Thanks 👍

PR #26998, opened 2023-12-06
fix : Fatal error: Uncaught TypeError: Unsupported operand types: str…

grandoc:
…ing + int in /home/httpd/vhosts/aflac.fr/domains/dev.aflac.fr/httpdocs/core/modules/DolibarrModules.class.php on line 879


eldy:
The PR remove the file completely...

PR #27005, opened 2023-12-06
Fatal error: Uncaught TypeError: Unsupported operand types: string + …

grandoc:
…int in /home/httpd/vhosts/aflac.fr/domains/dev.aflac.fr/httpdocs/core/modules/DolibarrModules.class.php on line 879


htdocs/core/modules/DolibarrModules.class.php
```diff
@@ -873,6 +873,7 @@ public function getModulePosition()
 		if (in_array($this->version, array('dolibarr', 'experimental', 'development'))) {	// core module
 			return $this->module_position;
 		} else {																			// external module
+			$position = (int) $this->module_position;
```
eldy, line 876: You add a variable but it is not used.

PR #27006, opened 2023-12-06
QUAL Fix php-cs-fixer config to avoid 'public const' not ok for PHP7.0

mdeweerd:
Do not create incompatibility with PHP7.0 with php-cs-fixer.

From #26969
> Also PR the add dev tools to change code, must be separated from PR that is the result of code changed.
I just added a few files here to test that the commit works.  It does on a push request, but not on a pull request - the latter might be possible by looking for the source repository and try to push the change there.

> Even if we target PHP 7.1+, there is currently nothing that force us to break compatibility with php 7.0

Ok, I'll remove this from the rules - I'll submit another PR to change the ruleset.

_Originally posted by @mdeweerd in https://github.com/Dolibarr/dolibarr/pull/26969#discussion_r1417490279_

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -41,6 +41,8 @@ return (new PhpCsFixer\Config())
 		// So we use target PHP70 for the moment.
 		'@PHP70Migration' => true,
 		//'@PHP71Migration' => true,
+                // Avoid adding public to const (incompatible with PHP 7.0):
+                'visibility_required' => ['elements'=>['property', 'method']],
 
 		//'strict_param' => true,
 		//'array_syntax' => ['syntax' => 'short'],


.php-cs-fixer.dist.php
```diff
@@ -41,6 +41,8 @@
 		// So we use target PHP70 for the moment.
 		'@PHP70Migration' => true,
 		//'@PHP71Migration' => true,
+                // Avoid adding public to const (incompatible with PHP 7.0):
+                'visibility_required' => ['elements'=>['property', 'method']],
```
eldy, line 45: This rule is not the only one to not execute.
The only solution to be sure to not execute rule that are not "minimum version" compliant is to install php-cs-fixer by using the good php version.  If you install php-cs-fixer and runit from php8, you will also break compatibility with 7.4
This is why php version is forced into file php-cs-fixer/run-php-cs-fixer.sh

mdeweerd: "Absolutely sure" - certainly.

However, the configuration also allows to exclude conversions - updating the configuration to not apply this change does not exclude the other approach to help avoid unwanted changes.

The line
```
                'visibility_required' => ['elements'=>['property', 'method']],
```
Overrides the configuration for the PSR12 rule that applies public on consts .

The tool should have included this conversion rule in `@PHP71Migration` - I filed an [issue in php-cs-fixer](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/7526)

PR #27009, opened 2023-12-06
QUAL phpstan github action

frederic34:


.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,60 @@
...
+      - uses: actions/checkout@v4
+        with:
+         fetch-depth: 0
+      - name: Setup PHP
+        id: setup-php
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: "${{ matrix.php-version }}"
+          tools: phpstan, cs2pr
+          extensions: json, imagick, gd, zip, mbstring, intl, opcache, imap, mysql, pgsql, sqlite3, ldap, xml
```
mdeweerd, line 38: Not sure which can be added as extensions but for phan verification I also added (as stubs):

PDO,brotli,calendar,curl,fileinfo,ftp,mcrypt,memcache,mysqli,pdo_cubrid,pdo_mysql,pdo_pgsql,pdo_sqlite,session,soap,sockets

.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,60 @@
...
+        uses: actions/cache/restore@v3
+        with:
+          path: ./tmp
+          key: "phpstan-cache-PR-${{ matrix.php-version }}-${{ github.run_id }}"
+          restore-keys: |
+            phpstan-cache-PR-${{ matrix.php-version }}-
+      - name: Debug
+        run: cd ./tmp && ls -al
+      - name: Run PHPSTAN
+        run: phpstan -vvv analyse --error-format=checkstyle --memory-limit 4G -c phpstan_action.neon htdocs/ | cs2pr --graceful-warnings
```
mdeweerd, line 53: Do we need to specify the path (htdocs/) here as the phpstan_action.neon already mentions two?

.github/workflows/phpstan.yml
frederic34: pdo is not used

PR #27011, opened 2023-12-06
Fix  #27010 Possible to bill several times lines of an order already billed …

### LINKED ISSUE #27010: Possible to bill several times lines of an order already billed
### Bug

This is between a bug and a feature request 
Example : you create an invoice from an order, but you bill only some lines
Some time after, you create an invoice from the same order : you'll see a warning that an invoice already exists for this order, but all the lines of the original order will be pre-selected, ignoring those that are already billed...
This is true with both client invoices and supplier_invoices
I will propose a PR to resolve this

### Environment Version

all

### Environment OS

All

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

vmaury:
# FIX #27010  Possible to bill several times lines of an order already billed …


vmaury:
@eldy : I suppose you're tuning phpcs, I don't understand the phpcs / travis errors ...
Have a nice sunday !

vmaury:
@eldy : I saw you made some changes on phpcs config, I merged by branch #27011 with develop this morning, but there's still an error I don't understand ...
If you can have a look ... thanks !

eldy:
The choice on lines and quantity must not be defined by hidden rules:
When we create an invoice from an order, we have the list of invoice with a checkbox so user can select which line we import. We should also have the qty into an input so user can also select quantity.
And the predefined value can be set according to rules like "not already invoiced quantities". We can show for example 1/3 if there is only 1 qty to invoice among 3. Things are clearer for end user.
So we must first enhance the interface to allow to choose the qty to use in order before being able to modulate the qty to use by default.
A end user visible solution is better than a silent hidden rule.

Before we have this implementation, we can rely on the warning: "An invoice was already done for this order"

vmaury:
I well understand what you mean ...
But for now, to achieve what you said, we're obliged to modify lot of things in commonobject, core/tpl and so on..
I don't feel experimented enough to do this alone on my side.
I suggest we code or discuss about this during devcamp in Montpellier, I'll be there !

vmaury:
As my final client asks me to have a solution NOW, I coded this 
After the suggested lines, if there are other invoices/lines that are concurrent they are displayed
Existing_facts

Do you thing it could be a temporary solution ?

eldy:
> As my final client asks me to have a solution NOW, I coded this After the suggested lines, if there are other invoices/lines that are concurrent they are displayed Existing_facts
> 
> Do you thing it could be a temporary solution ?

Why not.
However, to make screen lighter, i suggest to just show a text link under the list of product
"[Picto warning] Other invoices already exists. Click here to show them..."
When clicking a js code show the full part of "existing invoices" as e see it on your screenshot.

PR #27042, opened 2023-12-11
NEW Allow to set a default project for POS sales

ymollard:
# NEW Allow to set a default project for POS sales
The TakePOS module can be set to a default project in the admin settings, so that all future sales are affected to this project.


htdocs/takepos/invoice.php
```diff
@@ -149,6 +150,11 @@ function fail($message)
 	$soc->fetch(getDolGlobalString($constforcompanyid));
 }
 
+// Assign a default project, if relevant
+if (isModEnabled('project') && !empty(getDolGlobalString("CASHDESK_ID_PROJECT".$takeposterminal))) {
```
eldy, line 154: !empty(getDolGlobalString(...  is always true.
I think you want to do 
!getDolGlobalString(...

ymollard: I want to execute that piece of code only if the global string exists and is not an empty string.
*  If it's "some_string", then getDolGlobalString() will return that string, so empty() will be false, so !empty() will be true and will execute the code
* If it's "", then getDolGlobalString() will return "", so empty() will be true, so !empty() will be false and will not execute the code
* It it's not set, then getDolGlobalString() will return "", and the latest behaviour applies

Where am I mistaken?

eldy: I made a mistake. The !empty(getDolGlobalString() will not be always true.
Bit the call of empty is useless, this is why the cti return an error expecting a more simple condition.
Test 
If (getDolGlobalString())
Is enough.

ymollard: Fixed with getDolGlobalInt() instead

ymollard: Poke (bis) @eldy

ymollard:
Can this PR now be merged, @eldy?

PR #27044, opened 2023-12-11
NEW : add some bank info on odt

defrance:


htdocs/core/class/commondocgenerator.class.php
```diff
@@ -944,7 +948,7 @@ public function fill_substitutionarray_with_extrafields($object, $array_to_fill,
 		// phpcs:enable
 		global $conf;
 
-		if (is_array($extrafields->attributes[$object->table_element]['label'])) {
+		if ($extrafields->attributes[$object->table_element]['count'] > 0) {
```
eldy, line 951: What was wrongwith the test 
if (is_array($extrafields->attributes[$object->table_element]['label'])  ?

defrance: if we have no extrafields for this table_element, the label is not present
the good way to verify it is to test the 'count' value

PR #27049, opened 2023-12-11
CLOSE #27048 add form to globally set "VARIANT_ALLOW_STOCK_MOVEMENT_ON_VARIANT_PARENT" variable

### LINKED ISSUE #27048: Add Yes/No Form to Variations Module to globally set "VARIANT_ALLOW_STOCK_MOVEMENT_ON_VARIANT_PARENT" variable - concerns Issue #26965
### Feature Request

In Issue #26965 the missing check for the Variable "VARIANT_ALLOW_STOCK_MOVEMENT_ON_VARIANT_PARENT" was discussed and fixed in the PR #26967

Now I propose a change in the Variants module settings page to allow the user to set this global variable. This would also be the first and only way to set this variable via an option.

### Use case

When a user activates the variations module and checks the possible options inside the module the user will be able to see that there is a possibility to allow the stock movements for the variant parent products as well.

### Suggested implementation

Use the same syntax as with the "HideProductCombinations":

https://github.com/Dolibarr/dolibarr/blob/e4536e0a2ebb87a91a09bf595242c2181889c0b4/htdocs/variants/admin/admin.php#L75-L76

Generally add a form with a yes/no option to the options page
```php
$form->selectyesno()
```

Together with a similar syntax as in the Lines 40-56 one could implement and set the variable globally:

https://github.com/Dolibarr/dolibarr/blob/e4536e0a2ebb87a91a09bf595242c2181889c0b4/htdocs/variants/admin/admin.php#L40-L56

### Suggested steps

- add Option to variants/admin/admin.php via $form->selectyesno()
- add translationkey for new option to langs/en_US

DaBootO:
CLOSE #27048 add form to globally set "VARIANT_ALLOW_STOCK_MOVEMENT_ON_VARIANT_PARENT"

This commit adds another option to the variants module to set the global variable "VARIANT_ALLOW_STOCK_MOVEMENT_ON_VARIANT_PARENT".
This variable allows the variant parent product to be used in stock movements.

The changes in langs/en_US are needed as a new string will be used to explain to the user what this option does.

The changes in variants/admin/admin.php are split into setting the global variable via the GETPOST (if it was sent) and into another selectyesno() form.

Both changes are oriented toward the same syntax used in both cases.


eldy:
By definition, the parent of variants is not a real product. Can you provide a use case when we need to use the virtual parent product in stock change ?

DaBootO:
We are using the variants to keep track of products which do have different external part numbers. In essence the variant product is the same as the variant parent. So using the same BOM etc.

We could create another product with a new ref. Nr. etc. and assign the external ref. Nr. to it but we wanted to keep track of how many of our "base products" also have variants for specific customers and don't want to duplicate every BOM etc.

Therefore we sell our "variant parent" and the "variations" seperately (based on the customer) and need the stock movement on the variant parent. I Hope this made clear why allowing the stock movement on the variant parent can be useful (although I also see that our use of the variant module is atypical, but in our opinion a viable option)

Maybe to make things also a little clearer: Our products did have another attribute named "externalPN". Now we are using Variants for this. A product 1000001 can have the variant 1000001_1234567 when our customer has already created an internal product on their side they called "1234567" and give us orders for the "1234567". That way we can keep the external ref. Nr. assigned to the variant parent, bc they are the same product with different numbers.

P.S. Thank you for this awesome piece of open-source software!

Issue #14266, opened 2020-07-20
Sending email if there's error during cron processing


atm-quentin:
# Sending email if there's error during cron processing
When cron running every short period , we sometimes dont check if cron ran well, sending an email allow us to be warned

## Use case
You launch cron, it's getting error, you receive an email with the cron output

## Suggested implementation
If conf is activated and cron send error code then send mail to error mail

## Suggested steps
- Create conf
- Check if email is filled
- If cron send an error code, send email with output and in subject Error Cron "name of cron"

atm-quentin:
@atm-john @atm-maxime @ATM-Nicolas @atm-lena @atm-gauthier @atm-greg @atm-florianm

atm-maxime:
Still wanted 😊

atm-quentin:
@eldy

atm-maxime:
Still would be useful

PR #27050, opened 2023-12-11
New trigger CRONJOB_ERROR

hansemschnokeloch:
# NEW trigger CRONJOB_ERROR

This trigger is called when a cronjob fails.
Purpose of this trigger is to get a versatile method to send any type of message if a cron job fails.
This trigger is primarily intended to be used in combination with the recent [webhook module](https://wiki.dolibarr.org/index.php?title=Module_Webhook) in order to send a notification to [ntfy.sh](https://ntfy.sh/) or any other notification system.
This trigger can also be used in a custom module to send an email (this could satisfy this old request https://github.com/Dolibarr/dolibarr/issues/14266 )

### DIFF
--- a/htdocs/modulebuilder/template/core/triggers/interface_99_modMyModule_MyModuleTriggers.class.php
+++ b/htdocs/modulebuilder/template/core/triggers/interface_99_modMyModule_MyModuleTriggers.class.php
@@ -311,6 +311,9 @@ class InterfaceMyModuleTriggers extends DolibarrTriggers
 			//case 'SHIPPING_REOPEN':
 			//case 'SHIPPING_DELETE':
 
+			// Cron jobs
+			//case 'CRONJOB_ERROR':
+
 			// and more...
 
 			default:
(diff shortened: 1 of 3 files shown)


htdocs/modulebuilder/template/core/triggers/interface_99_modMyModule_MyModuleTriggers.class.php
```diff
@@ -311,6 +311,9 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
 			//case 'SHIPPING_REOPEN':
 			//case 'SHIPPING_DELETE':
 
+			// Cron jobs
+			//case 'CRONJOB_ERROR':
```
eldy, line 315: Triggers are reserved for CRUD methods of an object, inside the update transaction.
To intercept code out of transaction context (like in your cas after running the cron_rub_jobs), you must instead use hooks.

eldy:
Triggers are reserved for CRUD methods of an object, inside the update transaction.
To intercept code out of transaction context (like in your cas after running the cron_rub_jobs), you must instead use hooks.
See https://wiki.dolibarr.org/index.php/Hooks_system

PR #27052, opened 2023-12-11
FIX backtopage on comm action card create action

atm-saamiperdrix:
FIX backtopage on comm action card create action


htdocs/core/modules/modHRM.class.php
```diff
@@ -265,6 +265,14 @@ public function __construct($db)
 		$this->rights[$r][4] = 'write_personal_information';
 		$this->rights[$r][5] = 'write'; // In php code, permission will be checked by test if ($user->rights->hrm->write_personal_information->write)
 		$r++;
+
+		// Evaluation
+		$this->rights[$r][0] = 4033; // Permission id (must not be already used)
```
eldy, line 270: Adding permission is not allowed into a branch in maintenance status.

PR #27062, opened 2023-12-12
fix missing declaration

altairis-tof:
reduce technical debt 
image

### DIFF
--- a/htdocs/fourn/class/fournisseur.facture.class.php
+++ b/htdocs/fourn/class/fournisseur.facture.class.php
@@ -201,7 +201,6 @@ class FactureFournisseur extends CommonInvoice
 
 	// Warning: Do not set default value into property defintion. it must stay null.
 	// For example to avoid to have substition done when object is generic and not yet defined.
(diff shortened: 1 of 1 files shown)


htdocs/fourn/class/fournisseur.facture.class.php
```diff
@@ -201,6 +201,7 @@ class FactureFournisseur extends CommonInvoice
 
 	// Warning: Do not set default value into property defintion. it must stay null.
 	// For example to avoid to have substition done when object is generic and not yet defined.
+	public $amount_ht;
```
eldy, line 204: We alread have total_ht for amount. May be the problem is in the code rhat use amount_ht ? What will be the difference between amount_ht and total_ht ?

altairis-tof:
@eldy you are right; i removed the use of amount_ht here ; it was only used in the initAsSpecimen function and not in the cannelle pdf model.

altairis-tof:
@eldy seems to be a problem with phpcs

PR #27063, opened 2023-12-12
make field date_start accessable for GETPOST in proposal

cfoellmann:
makes the field `date_start` programmatically accessable on the proposal add form


htdocs/comm/propal/card.php
```diff
@@ -2008,7 +2008,8 @@
 			$sday = date("d", $tmpdte);
 			print $form->selectDate($syear."-".$smonth."-".$sday, 'date_livraison', '', '', '', "addprop");
 		} else {
-			print $form->selectDate(-1, 'date_livraison', '', '', '', "addprop", 1, 1);
+			$tmp_date_start = GETPOST('date_start') ? : -1;
+			print $form->selectDate($tmp_date_start, 'date_livraison', '', '', '', "addprop", 1, 1);
```
eldy, line 2012: Why using name date_start and not date_livraison (in english date_delivery) ?

cfoellmann: you are totally right.
I did this because we use it as a "start date" but it makes no sense in code. Force push is done

cfoellmann:
@eldy corrected and force pushed

cfoellmann:
can I change this to branch `19.0`

cfoellmann:
ready for 19/20

htdocs/comm/propal/card.php
```diff
@@ -2008,7 +2008,8 @@
 			$sday = date("d", $tmpdte);
 			print $form->selectDate($syear."-".$smonth."-".$sday, 'date_livraison', '', '', '', "addprop");
 		} else {
-			print $form->selectDate(-1, 'date_livraison', '', '', '', "addprop", 1, 1);
+			$tmp_date_delivery = GETPOST('date_delivery') ? : -1;
```
eldy, line 2011: Getpost key name is 'date_delivery'
But name of field 'date_livraison'
Should not be the same ?

cfoellmann: I would say that everything should be English and whenever possible. 
But if you insist on changing this I can

eldy: I think too that we must migrate step by step to english. But whatever is language, the name of variable must be consistent in all page.
So if the name of param is xxx for assignement, you should use xxx to read it. But here, there is 2 different name. Is it done on purpose ?

PR #27066, opened 2023-12-12
FIX default value for radio extrafield (v17+)

W1W1-M:
# FIX default value for radio extrafield
If a default value was defined for a radio extrafield, on creation forms the default input was never checked.


eldy:
Is it really a backport from v18 ?
I can't find such a fix in v18.

W1W1-M:
> Is it really a backport from v18 ? I can't find such a fix in v18.

Hi, 
Sorry I used the term _backport_ in the wrong manner. This also needs to be fixed in v18 & v19.

W1W1-M:
@eldy I've synced this PR with 17.0 branch and checked for merge conflicts. Thanks for any feedback.

Issue #26111, opened 2023-10-04
Add geolocation on most of dolibarr objects


rycks:
### Feature Request

There is a geo URI scheme: https://en.wikipedia.org/wiki/Geo_URI_scheme since 2010 ... and i make a proposition to add geo entries in most of dolibarr objects.

For example on socpeople (delivery address), entrepot (location), fichinter(where to go) and so on. Not only "postal address" like we have for the moment.

What about the datamodel ? 

MySQL/MariaDB/PostgreSQL seems to have a POINT datatype for that:
- https://mariadb.com/kb/en/point/
- https://www.postgresql.org/docs/current/datatype-geometric.html

Then it seems to be the most optimal way to store that data ... what do you think about that feature request ?

### Use case

Easy way to find all my customer near than a point.

More accurate data than "postal address" for parts of earth where "postal address" is a sort of graal.

Example:

`<a href="geo:37.786971,-122.399677">Direct GeoLocation</a>`

JoAnMaFe:
From my point of view, this is a good contribution and would significantly help the pick-up and delivery options, as well as the definition and location of customers and warehouses.

AshleyButcher:
If this is to be implemented, then whoever implements it needs to be conscious of different spacial reference systems and choose a common system for storing points in Dolibarr. It's important because while geospacial data is commonly using WGS-84 (like Google Maps or indeed the `geo` URI in the feature request), sometimes databases such as post code databases use some more national-specific coordinate system and requires conversion. 

Some obvious examples I've run into over the years beyond some paid national post code databases are couriers like UPS and FedEx who don't use WGS-84 for their interfaces.

It may as well be standardised as WGS-84 since it's so common, but it should be standardised for the project. Without it, it's like a date/time without a timezone.

hansemschnokeloch:
We should also add altitude. This would open the way to warehouse stock management with an indication of the storage location of a product (location with level on the storage rack).

JoAnMaFe:
OK.

AshleyButcher:
@hansemschnokeloch I don't think altitude is a useful measure for determining stock location. If you want accurate stock location, you should follow the standard Warehouse -> Aisle -> Rack -> Shelf -> Bin location used standard in the WMS world. I don't see how altitude would be more useful than just knowing what shelf it's on, and specifying a shelf is much simpler to implement and manage than altitude.

[...]

PR #27068, opened 2023-12-12
#26111 [DataModel] add: geolocation databases

nicolas-eoxia:


eldy:
Innmost cases, each object has 1 address, so 1 geolocation. So most needs are 1-1 relations. 
This means the 2 columns should directly be in the table of each object it is related to, to respect the normalization of database.
So lat and long should be in the same table where address, zip and country is stored.

d3v2a:
For GIS operations (e.g. finding all points within a perimeter), it would be simpler to check a table and retrieve elements than a union on each table containing the geolocation.

The implementation is simpler and more extensible for an external module.

Also, this data should be stored in a POINT-type field, to take advantage of the GIS features of SQL servers.

lmag:
https://www.dolibarr.fr/forum/t/idee-lancee-sur-le-github-de-dolibarr-ajouter-la-geolocalisation-des-objets-dolibarr/44366

PR #27082, opened 2023-12-13
ADD ticket subject on emails

BenjaminFlr:
When an email is sent to the customer or to the assignee, there's no information regarding the subject of the ticket.

I would like to add the information in the email subject.

So, instead of:
[MyBigCompany - Ticket TS2310-0006] New message

We could see:
[MyBigCompany - Ticket TS2310-0006: Ticket subject] New message


eldy:
Why not adding the topic to replace "New Message" ?
Such string is useless to say there is a new message. Receiving the email is enough to know that. So the value between [. ] remains a structure data and free text come after...

BenjaminFlr:
Hi @eldy, I just made the changes to replace the "New message" string by the ticket's subject. Does it look good?

BenjaminFlr:
Any update on here? :)

PR #27083, opened 2023-12-13
add hooks to actions

altairis-tof:


cfoellmann:
I like the idea for this hook.
But with your implementation you can only replace the complete "eventOptions". To no break the process you have to mirror all parameter set there.
Wouldnt it be good to give the option to only modify the variables with hook action?

altairis-tof:
> I like the idea for this hook. But with your implementation you can only replace the complete "eventOptions". To no break the process you have to mirror all parameter set there. Wouldnt it be good to give the option to only modify the variables with hook action?

i think we may go step by step ; merge this PR then do other one(s ?) to follow your suggestions.

htdocs/comm/action/index.php
```diff
@@ -968,6 +973,14 @@
 			//print 'Event '.$i.' id='.$event->id.' (start='.dol_print_date($event->datep).'-end='.dol_print_date($event->datef);
 			//print ' startincalendar='.dol_print_date($event->date_start_in_calendar).'-endincalendar='.dol_print_date($event->date_end_in_calendar).') was added in '.$j.' different index key of array<br>';
 		}
+
+		$parameters['obj'] = $obj;
+		$reshook = $hookmanager->executeHooks('eventElements', $parameters, $event, $action); // Note that $action and $object may have been modified by some hooks
```
eldy, line 978: For new hook, we should use a method name that start with 
"hook..."
so
hookEventElements

PR #27087, opened 2023-12-14
remove status select from application add form

cfoellmann:
Showing the status select is totally out of line from all other creation forms.
It also has the problem that it prevents all triggers from firing when the status changes on the element.

# QUAL
streamline the application add form by not showing the status select


cfoellmann:
This can also be pushed to 17/18 in my opinion. Would streamline the workflow for users

cfoellmann:
ready for 19.x

PR #27098, opened 2023-12-14
feature supplier order self-approval - 

cfoellmann:
# NEW
add three levels with permissions for supplier orders approvals. users can approve supplier orders up to the set netto amount without approve permission
Set a value in the setting for each level and give the user the permission.

@eldy what do you think? It is still a WIP. no UI strings but it is in internal use for a few months now without any functional issues.


eldy:
What is the difference with the existing option "3 step approval":
https://wiki.dolibarr.org/index.php?title=Module_Purchase_Orders#Configuration

cfoellmann:
supplier orders below this amount can be approved by users without the normal approve perm. One of the levels need to be given as perm to these users

eldy:
> supplier orders below this amount can be approved by users without the normal approve perm. One of the levels need to be given as perm to these users

So it means you want to have users allowed to approve the purhcase order without having the permission. So why not just giving the permission to all ?

cfoellmann:
> o it means you want to have users allowed to approve the purhcase order without having the permission. So why not just giving the permission to all ?

We give users the ability to order "without approval" if the amount is small.

htdocs/core/modules/modFournisseur.class.php
```diff
@@ -253,6 +254,30 @@ public function __construct($db)
 		$this->rights[$r][4] = 'commande';
 		$this->rights[$r][5] = 'export';
 
+		$r++;
+		$this->rights[$r][0] = 1192;
+		$this->rights[$r][1] = 'Self-Approve LVL1';
+		$this->rights[$r][2] = 'r';
+		$this->rights[$r][3] = 0;
+		$this->rights[$r][4] = 'commande';
+		$this->rights[$r][5] = 'selfapprovelvl1';
```
eldy, line 263: As this need an more complete setup in permission that standard rule (can read / update /delete), the new permissions must have key  
$this->rights[$r][5] = 'selfapprovelvl1_advance';
with _advance at end of key, so the permission will be visible when the option "Use advanced permission" is on.

eldy:

PR #27100, opened 2023-12-14
add rank to shipment lines display

altairis-tof:


hregis:
@altairis-tof you can't not use "position" instead "rang"?

altairis-tof:
> @altairis-tof you can't not use "position" instead "rang"?

@hregis position is not a field of commandedet

htdocs/expedition/shipment.php
```diff
@@ -623,6 +623,7 @@
 			$i = 0;
 			print '<thead>';
 			print '<tr class="liste_titre">';
+			print '<th>'.$langs->trans("Rank").'</th>';
```
eldy, line 626: To follow other document, can you add the condition 
` if (getDolGlobalString('MAIN_VIEW_LINE_NUMBER')) {`

altairis-tof: ok it is fixed

PR #27105, opened 2023-12-14
NEW Simplify the verification of context in hook

d3v2a:
# New

it add a method to simplify the verification of context.

usage:

```php
if ($this->isContext($parameters, 'propalcard')) {
```

```php
if ($this->isContext($parameters, ['propalcard', 'invoicecard', 'ticketcard'])) {
```

### DIFF
--- a/htdocs/core/class/commonhookactions.class.php
+++ b/htdocs/core/class/commonhookactions.class.php
@@ -36,4 +36,26 @@ abstract class CommonHookActions
 	 * @var array 	Array of results.
 	 */
 	public $results = array();
+
+	/**
+	 * Check context of hook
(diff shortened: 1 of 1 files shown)


htdocs/core/class/commonhookactions.class.php
```diff
@@ -36,4 +36,26 @@ abstract class CommonHookActions
...
+	 */
+	protected function isContext($parameters, $allContexts)
+	{
+		if (is_array($allContexts)) {
+			foreach ($allContexts as $context) {
+				if ($this->isContext($parameters, $context)) {
+					return true;
+				}
+			}
+		}
```
eldy, line 54: May be we should add return false here ?

PR #27107, opened 2023-12-14
Fix utf8_encoding/decoding (PHP8.2 deprecation)

mdeweerd:
Fix files that were missed in #27018

# Fix deprecation of utf_(en|de)code function ins PHP8.2


eldy:
Can you solve conflict and repush without any change into extrnal lib (files into includes dir).
such change must be done into separate pr

mdeweerd:
Ok, noted that changes to libraries need to be in separate PR.  However, it seems that now there are no more changes left to these.

This PR has been rebased.

eldy:
Note: Do you know we already have this in functions.lib.php

```
if (!function_exists('utf8_decode')) {
	/**
	 * Implement utf8_decode for PHP that does not support it.
	 *
	 * @param	mixed	$elements		PHP Object to json encode
	 * @return 	string					Json encoded string
	 */
	function utf8_decode($elements)
	{
		return mb_convert_encoding($elements, 'ISO-8859-1', 'UTF-8');
	}
}

```

mdeweerd:
Yes, I know that already is in the functions.lib.php - it will not avoid deprecation messages as the built-in php function takes precedence!

PR #27109, opened 2023-12-14
dont reopen billed orders

altairis-tof:
when an order is billed, it should not be modified anymore

### DIFF
--- a/htdocs/commande/card.php
+++ b/htdocs/commande/card.php
@@ -2908,7 +2908,7 @@ if ($action == 'create' && $usercancreate) {
 			$reshook = $hookmanager->executeHooks('addMoreActionsButtons', $parameters, $object, $action);
 			if (empty($reshook)) {
 				// Reopen a closed order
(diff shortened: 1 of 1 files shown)


eldy (commented):
Why not ? 
The status billed or not is independant if the workflow of staus of order. We can bill or not an order at begin of order status, even when not yet delivered, so going back should not depend on billed or not

altairis-tof:
Hello @eldy Maybe we could add a hidden constant for those who dont want their teams to do such things ?

eldy:
> Hello @eldy Maybe we could add a hidden constant for those who dont want their teams to do such things ?

Yes you can. Option to allow "to block" looks better than blocking by default (will match more use cases).

altairis-tof:
> > Hello @eldy Maybe we could add a hidden constant for those who dont want their teams to do such things ?
> 
> Yes you can. Option to allow "to block" looks better than blocking by default (will match more use cases).

Ok @eldy, we have new constant ORDER_DONT_REOPEN_BILLED

PR #27113, opened 2023-12-14
Fix deprecations for PHP 7.2 and up

mdeweerd:
# Fix PHP7.2 and up deprecations

The deprecations are found in branch develop and are can be fixed from vers 17.0 upwards/

Fix issue with nested ternary operators:
https://www.php.net/manual/en/migration74.deprecated.php

Fix use of deprecated each()
https://wiki.php.net/rfc/deprecations_php_7_2#each

Fix use of access to string acces using curly brackets
https://www.php.net/manual/en/migration74.deprecated.php

Note: bug in qrcode library was found because of "invalid" fix by rector.

### DIFF
--- a/htdocs/includes/nusoap/lib/Mail/mime.php
+++ b/htdocs/includes/nusoap/lib/Mail/mime.php
@@ -183,7 +183,7 @@ class Mail_mime
 
         // Update build parameters
         if (!empty($params) && is_array($params)) {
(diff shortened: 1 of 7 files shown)


eldy:
Because this is code of external lib, change for php compatibility purpose is done by upgrading the lib in the develop branch. It should has been now done in develop by upgrading nusoap to 0.9.16

PR #27118, opened 2023-12-15
fix phpdoc

frederic34:


htdocs/core/class/commonobject.class.php
```diff
@@ -93,7 +93,7 @@ abstract class CommonObject
 	public $element;
 
 	/**
-	 * @var int 		The related element
+	 * @var string    Fieldname with ID of parent key if this field has a parent
```
eldy, line 96: Sure ? Not the id of element here ?

frederic34:
```
        $sql = "SELECT count(rowid) FROM ".$this->db->prefix().$this->table_element_line;
        $sql .= " WHERE ".$this->fk_element." = ".((int) $this->id);
        if (!$renum) {
```

PR #27157, opened 2023-12-18
Fix Update PSWebServiceLibrary to fix deprecated function call libxml…

mdeweerd:
…_disable_entity_loader()

# Fix PSWebServiceLibrary.class.php: PhanDeprecatedFunctionInternal: Call to deprecated function \libxml_disable_entity_loader()

See https://github.com/Dolibarr/dolibarr/actions/runs/7242938198/job/19729044021#step:5:32

AFAICS, only the get method is used - it's API interface looks alike.

Source from https://github.com/PrestaShop/PrestaShop-webservice-lib .

### DIFF
--- a/htdocs/admin/dolistore/class/PSWebServiceLibrary.class.php
+++ b/htdocs/admin/dolistore/class/PSWebServiceLibrary.class.php
@@ -289,7 +289,13 @@ class PrestaShopWebservice
 		if ($response != '') {
 			libxml_clear_errors();
 			libxml_use_internal_errors(true);
-			$xml = simplexml_load_string(trim($response), 'SimpleXMLElement', LIBXML_NOCDATA);
+			if (LIBXML_VERSION < 20900) {
+				// Avoid load of external entities (security problem).
+				// Required only if LIBXML_VERSION < 20900
+				libxml_disable_entity_loader(true);
+			}
+
+			$xml = simplexml_load_string(trim($response), 'SimpleXMLElement', LIBXML_NOCDATA|LIBXML_NONET);
 			if (libxml_get_errors()) {
 				$msg = var_export(libxml_get_errors(), true);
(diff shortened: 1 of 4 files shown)


htdocs/admin/dolistore/class/PSWebServiceLibrary.class.php
```diff
@@ -222,26 +279,21 @@ public function getVersion()
...
 		if ($response != '') {
 			libxml_clear_errors();
 			libxml_use_internal_errors(true);
-			libxml_disable_entity_loader(true);	// Avoid load of external entities (security problem). Required only if LIBXML_VERSION < 20900
-
-			if (!function_exists('simplexml_load_string')) {
-				throw new PrestaShopWebserviceException('Method simplexml_load_string not available. Your PHP does not support xml.');
-			}
-			$xml = simplexml_load_string($response, 'SimpleXMLElement', LIBXML_NOCDATA|LIBXML_NONET);
+			$xml = simplexml_load_string(trim($response), 'SimpleXMLElement', LIBXML_NOCDATA);
```
eldy, line 292: This introduce a security hole. To fix it we can use both

`libxml_disable_entity_loader(true);`	// Avoid load of external entities (security problem). Required only if LIBXML_VERSION < 20900

and

`$xml = simplexml_load_string(trim($response), 'SimpleXMLElement', LIBXML_NOCDATA|LIBXML_NONET);`

mdeweerd:
Reintroduced the security measures - should I report this upstream?

eldy:
> Reintroduced the security measures - should I report this upstream?

It would be usefull. But in a lot of case, upstream project does not merge our report. So we can keep into file dolibarr_changes.txt, the list of changes we do in external libs to remind us to apply them on a next upgrade...

PR #27159, opened 2023-12-18
FIX Multilangs : PDF lines description

atm-lena:
# FIX|Fix #

**Steps to reproduce the behavior**

- Default Language to FR
- Activate MultiLangs 
- Use a personal language for user "English"
- Generate a propal PDF with a translatable product label in English

Result => product label isn't translated (display in fr)

**Bug**

fetch_lines() in fetch() function of propal class doesn't load translations of product if multilangs is activated
In pdf_getlinedesc() function, $label is so FR but $outputlangs->defaultlang and $langs->defaultlang is same so label is not translated

**Suggestion**
Maybe translations lines must be loaded for all object in fetch() function if multilangs is activated ? This PR manage only propal  and invoice case.

### DIFF
--- a/htdocs/comm/propal/class/propal.class.php
+++ b/htdocs/comm/propal/class/propal.class.php
@@ -1680,7 +1680,7 @@ class Propal extends CommonObject
 				$this->lines = array();
 
 				// Lines
-				$result = $this->fetch_lines(0, (getDolGlobalInt('MAIN_MULTILANGS') ? 1 : 0));
+				$result = $this->fetch_lines();
 				if ($result < 0) {
 					return -3;
 				}
--- a/htdocs/comm/propal/class/propal.class.php
+++ b/htdocs/comm/propal/class/propal.class.php
@@ -1680,7 +1680,7 @@ class Propal extends CommonObject
 				$this->lines = array();
 
 				// Lines
-				$result = $this->fetch_lines();
+				$result = $this->fetch_lines(0, (!empty(getDolGlobalInt('MAIN_MULTILANGS')) ? 1 : 0));
 				if ($result < 0) {
 					return -3;
 				}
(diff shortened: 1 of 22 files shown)


htdocs/comm/propal/class/propal.class.php
```diff
@@ -1680,7 +1680,7 @@ public function fetch($rowid, $ref = '', $ref_ext = '', $forceentity = 0)
 				$this->lines = array();
 
 				// Lines
-				$result = $this->fetch_lines();
+				$result = $this->fetch_lines(0, (getDolGlobalInt('MAIN_MULTILANGS') ? 1 : 0));
```
eldy, line 1683: For oerformance purpose, we should load translations only when needed. So load should be done into the code for odf generation only.

atm-lena: Like that ?

htdocs/core/lib/pdf.lib.php
```diff
@@ -1410,29 +1410,68 @@ function pdf_getlinedesc($object, $i, $outputlangs, $hideref = 0, $hidedesc = 0,
...
+        if($multilangsactive){
+            $prodser->getMultiLangs();
+            $object->lines[$i]->multilangs = $prodser->multilangs;
+        }
+    }
+	//label
+	if (!empty($object->lines[$i]->label)) {
+		$label = $object->lines[$i]->label;
+	} else {
+		if (!empty($object->lines[$i]->multilangs[$langs->defaultlang]['label']) && $multilangsactive) {
```
eldy, line 1441: For PDF,what we want to use is the language of recipient of the pdf, so $outputlangs, not $langs. Am i wrong ?

atm-lena: You're right : fixed !

atm-lena:
@eldy

PR #27161, opened 2023-12-18
FIX : Correction of the backtopage on job position création

atm-adrien:
## FIX : Correction of the backtopage on job position création

### DIFF
--- a/htdocs/hrm/position.php
+++ b/htdocs/hrm/position.php
@@ -906,7 +906,7 @@ if ($action == 'create') {
 	print '<input type="hidden" name="token" value="' . newToken() . '">';
 	print '<input type="hidden" name="action" value="add">';
 	if ($backtopage) {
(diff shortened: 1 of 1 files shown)


htdocs/hrm/position.php
```diff
@@ -906,7 +906,7 @@ function DisplayPositionList()
 	print '<input type="hidden" name="token" value="' . newToken() . '">';
 	print '<input type="hidden" name="action" value="add">';
 	if ($backtopage) {
-		print '<input type="hidden" name="backtopage" value="' . $backtopage . '">';
+        print '<input type="hidden" name="backtopage" value="' . $backtopage . '?fk_job='.$fk_job.'">';
```
eldy, line 909: What if $backtopage is someting like
url?param1=x
You will add 2 "?"
Pb is surely at the page that set the "backtopage" parameter. the fk_job link should be added in this link by the caller.

PR #27165, opened 2023-12-18
Php V8 warning

defrance:
FIX : append when create an order from a proposal

### DIFF
--- a/htdocs/commande/card.php
+++ b/htdocs/commande/card.php
@@ -166,7 +166,7 @@ if (empty($reshook)) {
 		}
 	}
 
-	$selectedLines = "";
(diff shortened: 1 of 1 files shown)


htdocs/commande/card.php
```diff
@@ -166,6 +166,8 @@
 		}
 	}
 
+	$selectedLines = "";
```
eldy, line 169: selectedLines should be an array.
Is   selectedLines = array()    solve the trouble ?

defrance: the main problem is the variable is not defined initialy. propose to define it as array if you prefer

Issue #27166, opened 2023-12-19
Attachment auto save issue in EmailCollector module


alsoft10:
### **EmailCollector module can't save the attachement though IMAP fetch**

When customer reply to the ticket conversation using support email with attachment, EmailCollector class can't save the attachment. Please see the attachment for more details.
![attachment-save -error](https://github.com/Dolibarr/dolibarr/assets/1249577/afa5b6d7-e2ef-420a-a7d5-df4a7e27354c)

### Environment Version

18.0.4

### Environment OS

Windows 10

### Environment Web server

Apache 2.4

### Environment PHP

8.1.12

### Environment Database

MySQL

### Expected and actual behavior

Attachment should be saved automatically with the ticket reply by the customer.

### Steps to reproduce the behavior

1. Create an email collector **Collect_Responses_In**
2. Configure it with your IMAP email server
3. Add Operations to do by collector as the following
4. Record an event in agenda (with type Email sent or received)
5. Save attached files into object documents if a ref of an object is found into email topic.
6. Now send an emaill with attachment to the IMAP email address you are using as a reply. Not the fisrt email. 

### Attached files

![FireShot Capture 001 - EmailCollector - localhost](https://github.com/Dolibarr/dolibarr/assets/1249577/3db2bbe5-811c-4307-ab97-87fcade48324)

alsoft10:
The issue is solved!

PR #27168, opened 2023-12-19
FIX#27166

alsoft10:
# Instructions
#27166 now Dolibarr email collector module can automatically save ticket reply with attachment via IMAP automation.


alsoft10:
@eldy IAMP attachemnt issue solved

PR #27169, opened 2023-12-19
FIX : search by thirdparty in don list

Thomas905:


htdocs/don/list.php
```diff
@@ -119,6 +124,9 @@
 if (trim($search_company) != '') {
 	$sql .= natural_search('d.societe', $search_company);
 }
+if (trim($search_thirdparty) != '') {
+	$sql .= " AND s.nom LIKE '%" . $db->escape($search_thirdparty) . "%'";
```
eldy, line 128: You should use
`$sql .= natural_search("s.nom", $search_thirdparty);`

PR #27175, opened 2023-12-19
QUAL fix hooks on contact.php files - fixes contexts

cfoellmann:
# QUAL
Some */contact.php pages are missing proper hooks and thereby context for externally defined actions.
Also doAction is missing in some.


hregis:
@cfoellmann can you use english for the context ? "invoicecontactcard", "ordercontactcard", etc... thank you

cfoellmann:
@hregis will do.
what about 'globalcard'? where should that be added? Only to the main card.php or to all cards like contact, note, agenda etc?

eldy:
'globalcard' is for all tabs.
Not sure it is used yet, but well, was introduced like that.

cfoellmann:
I think this PR is ready now.
I will go over the other cards files and clean those up over the holidays.

PR #27182, opened 2023-12-20
PERF : change default db host

defrance:
# Instructions
by default the dbhost is set to localhost, but we have a lost of perf comparing use of 127.0.0.1
propose to change the default value to give performance at beginner too

### DIFF
--- a/htdocs/install/fileconf.php
+++ b/htdocs/install/fileconf.php
@@ -285,7 +285,7 @@ if (!empty($force_install_noedit)) {
 
 	<?php
 	if (!isset($dolibarr_main_db_host)) {
-		$dolibarr_main_db_host = "localhost";
+		$dolibarr_main_db_host = "127.0.0.1";
 	}
 	?>
 	<tr>
--- a/htdocs/install/fileconf.php
+++ b/htdocs/install/fileconf.php
@@ -405,7 +405,7 @@ if (!empty($force_install_noedit)) {
 			<input type="text"
 				   id="db_host"
 				   name="db_host"
-				   value="<?php print(!empty($force_install_dbserver) ? $force_install_dbserver : (!empty($dolibarr_main_db_host) ? $dolibarr_main_db_host : 'localhost')); ?>"
+				   value="<?php print(!empty($force_install_dbserver) ? $force_install_dbserver : (!empty($dolibarr_main_db_host) ? $dolibarr_main_db_host : '127.0.0.1')); ?>"
 				<?php if ($force_install_noedit == 2 && $force_install_dbserver !== null) {
 					print ' disabled';
 				} ?>


htdocs/install/fileconf.php
```diff
@@ -285,7 +285,7 @@ class="minwidth300"
 
 	<?php
 	if (!isset($dolibarr_main_db_host)) {
-		$dolibarr_main_db_host = "localhost";
+		$dolibarr_main_db_host = "127.0.0.1";
```
eldy, line 288: localhost is not always resolved as 127.0.0.1, may be 127.0.0.2... depending on serveur setup and distribution.
Also if using a direct io, it gas effect on oermission in mysql (localhost and 127.0.0.1 are sozcial case in permission tables for host)
Then which bznchmark did you use to get differences. Because localhost is resolved with a local dns resolution based in a cache file, the cost happen only at first access (even if so low that we can't see it) and should be totaly invisible at all any other other.
Do tou have benchmark studies ?

defrance: First benchmark is on my own windows pc (2 second on loading page)
https://serverfault.com/questions/1040302/windows-10-localhost-resolution-slow

I agree that for dolibarr users hosted on Linux environments there should be no problems but under Windows it is a real performance problem, but these are the least competent users (and the most numerous. ..)

If I proposed this fix it is because I am thinking of the basic users of dolibarr, not the system expert

eldy: The link mention a trouble because the user has a difference when it should not. Having a difference between using localhost / 127.0.0.1, even on windows, means there is a trouble on the system that is not related with Dolibarr (may be a malware installed that hook network access, a dns component that has been removed, an ipv6 layer that is not correctly setup of not working, there is a lot of possible reason...).
Windows also use a dns resolution with a cache for "localhost", so result should be the same.
does the hosts file on windows contains its official content ("127.0.0.1 localhost") ?

Introducing 127.0.0.1 generates a lot of trouble for end users, above all on permissions or remote cloud installation. It has also a different meaning: localhost sometime means "use socket" instead of tcp/ip. Impact is important and performance change should be very low. "localhost" should be as fast as 127.0.0.1 and in some case should even be faster (because socket is faster that full tcp/ip on linux, same for windows that uses named pipes). It has also impact on security (not managed the same way by mariadb for example). 

Changing this will introduce troubles for end users. And we should not include changes because few users has instable environments that they should not have. First, finding why there is such bad behaviour to understand what's wrong must be done before making a change to compensate in Dolibarr.

I did some test with bat shell in a loop to make several launches of mysql command line and see no difference on a windows 10 (OEM). Which process did you used to get a difference ?

defrance: I have made 2 test on a V19 without external module
time 127 0 0 1
time localhost
my /windows/drivers/etc/hosts has no specifics setting
From my analysis it is the first connection to the base which takes so long.
After further research, it seems that this problem is linked to wampserver, I found messages on the forum also talking about slowness with doliwamp probably linked to this too.

what do you think about simply modifying the text next to the field during installation to specify that it is preferable to use an IP address?

eldy: Ouch !  2.1 s for a first connection !  Really something wrong in the server. No idea on my side. No trouble with doliwamp.

Completing the help text may be a good thing. We must make a difference between linux and windows. Most Linux distributions are designed to use localhost that is the recommended choice for local access (because it will use "socket" in priority before pure tcp/ip. Windows should also be "better" with localhost but if you and other experience troubles on doliwamp, we may help the user by adding a warning. For example we can complete the help message in installation with:
 
"Very slow database connection are experienced by some users when using 'localhost' on some Windows servers. You can try to replace with '127.0.0.1' if you experience such troubles (this can be done later by editing the conf.php file)"

PR #27193, opened 2023-12-21
Fix #27190 Set inventory to draft and label

### LINKED ISSUE #27190: Inventory : Setting it back to draft clears the inventory
### Bug

Bug:
- Setting an inventory back to "Draft" will empty the inventory without confirmation

I had problems with users inadvertently (or because they wanted to edit the label ) clicking the "Back to draft" button and therefore clearing the inventory ( and they had to start over the whole inventory). There is no confirmation, as soon as they click the button, the inventory is cleared.

Expected behavior: 
- We should be able to edit the label of the inventory without clearing it.
- Ask for confirmation before clearing

Version : 18.0.4
( I checked the code for the master branch, it does not appear to be fixed )

This file:
https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/product/inventory/card.php

that button:
```
			// Back to draft
			if ($object->status == $object::STATUS_VALIDATED) {
				if ($permissiontoadd) {
					print '<a class="butAction" href="'.$_SERVER['PHP_SELF'].'?id='.$object->id.'&action=confirm_setdraft&confirm=yes&token='.newToken().'">'.$langs->trans("SetToDraft").'</a>';
				}
			}
```

### Environment Version

18.0.4

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

Hystepik:
- Add a confirmation to set inventory to draft
-  Add possibility to edit label when inventory set to validated


Drinausaur:
Thank you,

I think the message in the confirmation form could be changed to add something like:

"The quantities currently set in the inventory will be reset."

In my opinion, this is something important as it could reset hours of work. People clicking on this button should be aware of that.

PR #27197, opened 2023-12-21
QUAL fix stock list table with proper header

cfoellmann:
# QUAL
streamline the table of stock in a warehouse.
Add a context `stocklist` to suggest a list on the page


htdocs/product/stock/card.php
```diff
@@ -608,8 +609,10 @@
 
 			// TODO Create $arrayfields with all fields to show
 
-			print '<table class="noborder centpercent">';
-			print "<tr class=\"liste_titre\">";
+			print '<div class="div-table-responsive">';
```
eldy, line 612: Can you move this div just under the print of title (1 line toward bottom) ?

cfoellmann: done.  and force push
I hope I understood you right.

PR #27200, opened 2023-12-21
QUAL professional identifiers - remove TODO by adding profid.lib.php

thibdrev:
This draft PR aims at answering a TODO found in `societe.class.php`:
https://github.com/Dolibarr/dolibarr/blob/b1fa6f596f8424993f02fcad55d32bd58bf67bc0/htdocs/societe/class/societe.class.php#L3834

For the moment I've simply:
- created a new library `/htdocs/core/lib/profid.lib.php`, and
- moved into it a function that checks whether a SIREN string is valid or not.

However I don't know if this is the right way to do 😇 .

So far I am thinking of:
- validating this first PR, then
- creating an additional PR for each function moved into this new library (to ease follow-up and code review), and
- after v19 beta freeze, eventually pushing other PR to improve each function if needed.

Can someone please confirm if this is the right way to go and or provide remarks / way ahead on how to do this properly?


eldy:
Looks good. You can continue to switch cide to validate prof id into the new .lib.php file.

PR #27201, opened 2023-12-21
next fix for  #17665 : autoloader collision

rycks:
#FIX 17665 : restler autoload and other autoloader

In case of other autoload classes it could happens a sort of collision with restler (when using api).

That patch fix the "PHP Fatal error:  Uncaught Error: Access to undeclared static property: $loader" problem when Restler try to manage classes autoloaded by other systems like composer.

Why do i put it in dolibarr 13.0 ? just because of one of my customer has that bug and uses that dolibarr version, please let me know if i have to make the job for newer versions of dolibarr.

thanks
Éric


rycks:
oooops forget that PR i have to do it againts 13.0 !

PR #27205, opened 2023-12-21
next fix for  #17665 : autoloader collision

rycks:
#FIX 17665 : restler autoload and other autoloader

In case of other autoload classes it could happens a sort of collision with restler (when using api).

That patch fix the "PHP Fatal error: Uncaught Error: Access to undeclared static property: $loader" problem when Restler try to manage classes autoloaded by other systems like composer.

Why do i put it in dolibarr 13.0 ? just because of one of my customer has that bug and uses that dolibarr version, please let me know if i have to make the job for newer versions of dolibarr.

thanks
Éric

### DIFF
--- a/dev/dolibarr_changes.txt
+++ b/dev/dolibarr_changes.txt
@@ -152,9 +152,9 @@ with
         elseif (strpos($filename, '://') === false)
 
 * To avoid to have QRcode changed because generated with a random mask, replace
-define('QR_FIND_FROM_RANDOM', 2);
-with
-define('QR_FIND_FROM_RANDOM', false);
+	define('QR_FIND_FROM_RANDOM', 2);
+	with:
+	define('QR_FIND_FROM_RANDOM', false);
 
 * Removed useless directories ("examples", "tools")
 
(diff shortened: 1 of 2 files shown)


frederic34:
https://github.com/Dolibarr/dolibarr/blob/63c38da901d11847100b65685b2be6e0f9b427fd/htdocs/includes/restler/framework/Luracast/Restler/AutoLoader.php#L68

autoloader is removing other autoloader

rycks:
@frederic34 yes i know that's the reason why i spend 4 hours on that question yesterday ... and some other experiences like https://www.dolibarr.fr/forum/t/lirremediable-collision-de-composer-dans-les-modules-complementaires/44936

Then my fix is a next step with that collision with restler autoloader and other like composer : i just add a test agains a property who may not exists and make error 500.

eldy:
It's a long time now we know that autoload is a big mess. The one who must decide which lib file to use in code is the developer and only the developer (and a simple include is enough for that), and not a random algorithm depending on  the environment setup.
So best thing is to avoid to use it to avoid conflicts, and use instead explicit declaration (explicit is better than implicit).
If we can't avoid it (because code rely on external lib you don't control), your fix can help, it can be merged in v13 too but you must also modify the file dolibarr_changes.txt that keep tracks of all changes done into external lib so when we update a lib, we can report it.

rycks:
Ok, i've just added that to .txt file and make a PR agains official Restler repository ...

But autoloader is not a closed road for dolibarr, lot of things to do before but one day i would became the solution i'm sure, like ld-loader had it before in c++ and all other languages :)

PR #27210, opened 2023-12-22
FIX - Workstation in BOM

atm-lucasmantegari:
# FIX - Workstation in BOM
The management of workstations has recently been modified, which creates a bug. To manage the services in the Bom you must use the product workstation and not the line.

### DIFF
--- a/htdocs/bom/class/bom.class.php
+++ b/htdocs/bom/class/bom.class.php
@@ -1475,9 +1475,9 @@ class BOM extends CommonObject
 					$unitforline = measuringUnitString($line->fk_unit, '', '', 1);
 					$qtyhourforline = convertDurationtoHour($line->qty, $unitforline);
 
-					if (isModEnabled('workstation') && !empty($line->fk_default_workstation)) {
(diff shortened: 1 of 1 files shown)


htdocs/bom/class/bom.class.php
```diff
@@ -1475,9 +1475,9 @@ public function calculateCosts()
 					$unitforline = measuringUnitString($line->fk_unit, '', '', 1);
 					$qtyhourforline = convertDurationtoHour($line->qty, $unitforline);
 
-					if (isModEnabled('workstation') && !empty($line->fk_default_workstation)) {
+					if (isModEnabled('workstation') && !empty($tmpproduct->fk_default_workstation)) {
```
eldy, line 1478: This looks strange because.
- the default workstation of a product is set at product level.
When we create a MO, we use the product and use this information to store it into the MO by default (but can be modified by end user), so when we change the product later, the workstation is kept as defined by user and not lost.
So using the workstation defined on the line of the MO looks what we expect to use (Note: with v19, this information, default workstation, is now visible into the line of MO).

PR #27211, opened 2023-12-22
fix aperçu/dl fichier quand il est dans {userid}/temp

atm-saamiperdrix:


eldy:
Can you describe process how to reproduce the bug ?

atm-saamiperdrix:
Ajouter une PJ manuellement puis essayer de la voir / télécharger

htdocs/document.php
```diff
@@ -207,19 +207,18 @@ function llxFooter()
...
+//$refname = basename(dirname($original_file)."/");
 
 // Security check
 if (empty($modulepart)) {
 	accessforbidden('Bad value for parameter modulepart');
 }
 
 // Check security and set return info with full path of file
-$check_access = dol_check_secure_access_document($modulepart, $original_file, $entity, $user, $refname);
+$check_access = dol_check_secure_access_document($modulepart, $original_file, $entity, $user, '');
```
eldy, line 218: This introduce a security hole by allowing to download file without checking on object itself (just on permission on module) so no check on entity become possible.

eldy:
On a particular object ? Can you reproduce the bug on v19 branch ?

atm-quentin:
For example, on a commercial proposal, when we attach a file and want to view it through the magnifying glass, if we don't have the "View other users, groups and their permissions" right, we won't be able to view it even though we've just attached it ourselves.

atm-quentin:
i reproduced on v19

atm-quentin:
in an email form

Issue #27215, opened 2023-12-23
Price HT becomes TTC when HT is empty in invoice/commande lines


UlusoftConsultancy:
### Bug

When I edit an order line and only fill the TTC value of the line and the HT value is left empty, the HT value is replaced by the TTC value. 

Steps to reproduce : add order line with a predefined product with a sell price. Click on edit line and remove the HT price. Click on save and watch how the HT is replaced by the TTC price.

<img width="794" alt="Screenshot 2023-12-23 140935" src="https://github.com/Dolibarr/dolibarr/assets/93868548/9211d6e7-9c5e-4f85-936a-7ac7368e0448">

<img width="788" alt="Screenshot 2023-12-23 141005" src="https://github.com/Dolibarr/dolibarr/assets/93868548/bf8f7aad-be24-4ebe-ab84-f3684a0683ec">

<img width="787" alt="Screenshot 2023-12-23 141034" src="https://github.com/Dolibarr/dolibarr/assets/93868548/ff0f0e76-de68-430f-a8b5-016a54c102d4">

In the code I saw that this is explicit behaviour, but I am almost certain this shouldn't work this way. If confirmed I will gladly remove this code.

<img width="254" alt="Screenshot 2023-12-23 141110" src="https://github.com/Dolibarr/dolibarr/assets/93868548/79b95547-b38d-4d68-b9d6-9434be20cb29">

<img width="223" alt="Screenshot 2023-12-23 141133" src="https://github.com/Dolibarr/dolibarr/assets/93868548/677eb2ca-f905-4c05-8b48-8d5dfec3b89a">

### Environment Version

18.0.1

atm-maxime:
I think this issue has been solved, I can't reproduce with 18.0.4. Can you confirm ?

UlusoftConsultancy:
Hi Maxime,

I can confirm this issue hasn't been fixed in version 18.0.4 see screenshots below:

![step1](https://github.com/Dolibarr/dolibarr/assets/93868548/0d371f71-828d-4605-857f-1dc5a7374efb)
![step2](https://github.com/Dolibarr/dolibarr/assets/93868548/c713f9e1-f6d0-44af-b043-1d26e42e2236)
![step3](https://github.com/Dolibarr/dolibarr/assets/93868548/b976c49b-518b-4b50-980d-a4b7ee6664cd)

[...]

PR #27216, opened 2023-12-23
# FIX [#27215 Deprecated subprice commande card]

UlusoftConsultancy:


eldy:
Waiting feedback on the issue #27215

UlusoftConsultancy:
> Waiting feedback on the issue #27215

Hi Eldy, I have the current PR running on my clients Dolibarr instance and it has been running without error there for months. This bug has been confirmed by other members also. Could you review and pull this changeset into core please?

sonikf:
@eldy fix is working for me
@UlusoftConsultancy can you target v. 18 instead of develop so it will be fixed in all consecutive versions?

PR #27228, opened 2023-12-27
Fix: Add not in operator

d3v2a:
# Fix


eldy:
@d3v2a Can you check to install the githook to clean your code before commiting it. See doc
https://wiki.dolibarr.org/index.php?title=Environment_and_development_tools#Install_Pre-commit_git_hook
This is to avoid a lot of errors reported by the CI.

Issue #27236, opened 2023-12-28
Missing field "manufacturer" on product/services


lmag:
### Feature Request

Missing manufacturer on product

### Use case

Missing manufacturer on product to know norme, type of fabrication, etc...
Sample an scaffolding should be manufacturer by
-Thirdparty 1
-Thirdparty 2
and sell by
-Thirdparty 3
On certain products we can have the name of the manufacturer for traceability

### Suggested implementation

Add table for link product/service to manufacturer
Add tab
Add norme, label, etc... for manufacer between date start and date end

FHenry:
Table schema suggested
new **Dictionnary of Relation type** between product/service and thridparty
> rowid
> code
> label
> active

new **Relation Table**
> rowid
> fk_product : required
> fk_soc  : required
> fk_relationtype  : required
> date_start
> date_end

Sample data can be
**Product 1** is **Manufacture** by **Soc1** from 01/01/2021 to 31/12/2022
**Product 1** is **certified ISO9011** by Soc1 from 01/01/2021 to 31/12/2021
**Product 1** is **Selling** by **Soc2** from 01/01/2022 to 31/12/2023

On Product or service list (or other dedicated list) it should be possible to filter by Relation type , date range start, date range end, Thirdparty

lmag:
@eldy any advice ?

eldy:
Looks good to me.
Can you update the migration 19 to 20 instead of 18 to 19 ?

FHenry:
Work in progress
https://github.com/FHenry/dolibarr/tree/dev_UI_product_thirdparty
IF anyone want to contribute, feel free

PR #27239, opened 2023-12-28
NEW: add table to store product relation with thirdparty

FHenry:
BDD change for #27236


lmag (approved):
First proposal for the management of product and service manufacturers

eldy:
As the new tables are for develop (v20), can you move migration lines into file 19.0.9-20.0.0.sql

FHenry:
> As the new tables are for develop (v20), can you move migration lines into file 19.0.9-20.0.0.sql

@eldy Done

PR #27240, opened 2023-12-28
NEW: Automatically fill matching extra fields of object on line creation.

ionutfinta:
# NEW|New [*Automatically fill matching extra fields of object on line creation.*]
This modification allows to automatically fill the matching extra field of an object (product/service) when both the object and the form (for example, when adding the object to a propale or invoice) have a same-named extra-field.

### DIFF
--- a/htdocs/core/tpl/objectline_create.tpl.php
+++ b/htdocs/core/tpl/objectline_create.tpl.php
@@ -821,13 +821,13 @@ if (!empty($usemargins) && $user->hasRight('margins', 'creer')) {
 
 						// Set values for any fields in the form options_SOMETHING
 						for (var key in data.array_options) {
-						  if (data.array_options.hasOwnProperty(key)) {
-						    var field = jQuery("#" + key);
-							if(field.length > 0){
-						    	console.log("objectline_create.tpl set content of options_" + key);
-								field.val(data.array_options[key]);
+							if (data.array_options.hasOwnProperty(key)) {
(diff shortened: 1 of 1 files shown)


eldy:
Can you provide an example with a screenshot of what the PR do ?

ionutfinta:
Ok, let's say you have a color attribute on a service:

image

... and you have the same attribute in your proposal:

image

Then, when you add the service to your proposal, it will automatically fill the attributes that have the same name on both your product/service and proposal:
image

BlueGarm:
Hello, 
Is there anything I can do to make it work?
I installed a test version with the V20 but for me it doesn't work, but maybe I missed a parameter.
Thanks

ionutfinta:
Hello @BlueGarm

I just tested it on a fresh install of the V20 and it works.

There is no parameter to set, you just have to define the same attribute for the object (Product/Service) and in the propale lines attributes.

In my example, I created a custom attribute "color" on the service and the same attribute in the Propale lines, then the value of the field is automatically copied on add.

For fields of type "List selection", we don't see the selected item directly, but the value is still copied and you can see it when validating the line or when you try to select a value.
I did not test for fields of type check boxes or radio buttons, but I don't think it will work for these.

Check your javascript console for eventual errors and give us more details regarding your problem if you need further assistance.

BlueGarm:
Thanks, i do some supplementary tries and yes it's working but not with suppliers orders.

PR #27245, opened 2023-12-29
FIX #24185: v18: display of the merged pdf lists

### LINKED ISSUE #24185: from release 16 onwards: pdf merge result for Intervention doesn't show up
### Bug

The merged file doesn't show up after performing a pdf merge for object type "Intervention".
The merged file is being created and can be found in the documents folder, but that's inconvenient.

I would expect the link to the merged file being displayed below the single files.

This has to be considered as a bug, because it worked up to release 15 and still works in current versions for all other relevant object's I've tested.

### Environment Version

16 and newer, including alpha of R18

### Environment OS

Linux Debian 5.10

### Environment Web server

Apache

### Environment PHP

8.1

### Environment Database

MariaDB 10.5.18

### Environment URL(s)

/htdocs/fichinter/list.php?leftmenu=ficheinter

I would expect the link to the merged file being displayed below the single files.
Instead the file is being created and can be found in the respective subfolder of the document folder, but there's no link displayed in the window.

### Module Commerce, section Interventions:
* Create more than one object of type interventions, including the validation, so that a pdf document is generated for each object
* Navigate to Interventions -> list
* Select at least two objects via the checkbox
* Select action -> pdf merge, confirm
* Realise the pop up message that a document has been created
* Realise that no result link is being displayed.
* Navigate to document subfolder and find the merged pdf

### Attached files

_No response_

routmoute:
Fix display of the merged pdf lists in some modules.

#24185

### DIFF
--- a/htdocs/adherents/list.php
+++ b/htdocs/adherents/list.php
@@ -1431,7 +1431,7 @@ print '</div>'."\n";
 
 print '</form>'."\n";
 
-if (array_key_exists('builddoc', $arrayofmassactions) && ($nbtotalofrecords === '' || $nbtotalofrecords)) {
+if (in_array('builddoc', array_keys($arrayofmassactions)) && ($nbtotalofrecords === '' || $nbtotalofrecords)) {
 	$hidegeneratedfilelistifempty = 1;
 	if ($massaction == 'builddoc' || $action == 'remove_file' || $show_files) {
 		$hidegeneratedfilelistifempty = 0;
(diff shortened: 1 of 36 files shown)


eldy:
As it is a backport, it must be fixed the same way it was in develop.
Otherwise, the merge of v18 into develop to retreive maintenance fixes will no more be possible (or become more difficult generating conflicts instead of automatic success).

routmoute:
I understand, Done.

PR #27248, opened 2023-12-29
25828 odt fix wrong styles

### LINKED ISSUE #25828: odt Generator Problem wrong colors and styles
### Bug

When generating odt it happens wrong and mixed styles and colors when using html code.

To Fix this we need to change line 299 ff.

from $key = floatval(str_replace('.', '', microtime(true))) + rand(0, 10);

$key = str_replace('.', '', microtime(true)) . uniqid() . rand(0, 10); // fixed by tronx2100 

now it will work perfect without errors.

### Environment Version

18.0.1

### Environment OS

linux debian bookworm

### Environment Web server

nginx

### Environment PHP

8.2.7

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

a-schild:
# FIX #25828 wrong style numbering
With php 8.2 and very very fast cpu's, we can still have non-unique styles
Adding a guuid solves this

### DIFF
--- a/htdocs/includes/odtphp/odf.php
+++ b/htdocs/includes/odtphp/odf.php
@@ -304,7 +304,7 @@ class Odf
 							}
 							if (strlen($odtStyles) > 0) {
 								// Generate a unique id for the style (using microtime and random because some CPUs are really fast...)
-								$key = str_replace('.', '', microtime(true)) . uniqid(mt_rand(), true) . rand(0, 10); // fixed by tronx2100 and aschild
+								$key = str_replace('.', '', (string) microtime(true)) . uniqid(mt_rand(), true);
 								$customStyles[$key] = $odtStyles;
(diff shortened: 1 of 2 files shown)


htdocs/includes/odtphp/odf.php
```diff
@@ -304,7 +304,7 @@ private function _replaceHtmlWithOdtTag($tags, &$customStyles, &$fontDeclaration
 							}
 							if (strlen($odtStyles) > 0) {
 								// Generate a unique id for the style (using microtime and random because some CPUs are really fast...)
-								$key = floatval(str_replace('.', '', microtime(true))) + rand(0, 10);
+								$key = str_replace('.', '', microtime(true)) . uniqid() . rand(0, 10); // fixed by tronx2100
```
eldy, line 307: It seems uniqid() is not unique when calling several time fast (see https://www.php.net/manual/fr/function.uniqid.php). 
May be you can call 
`uniqid(mt_rand(), false)`
we do this in a lot of other part of dolibarr code...

a-schild:
Is now updated as recommended. (But with true for more entropy in the uuid(...) call)

Issue #27249, opened 2023-12-29
Missing hook contact.php and contact.tpl


lmag:
### Feature Request

Missing hook contact.php and contact.tpl

### Use case

allow to target every line in php and js
![image](https://github.com/Dolibarr/dolibarr/assets/3288514/e3ddf1c9-9800-46a7-a9ad-216a20a1280f)

PR #27252, opened 2023-12-29
#27249 [TPL] add: JS element target for contact line in contact tpl

nicolas-eoxia:


lmag (approved):
https://github.com/Dolibarr/dolibarr/issues/27249

Issue #24180, opened 2023-03-10
Max decimals shown can't be less than Max decimals stored on DB


Daviid-P:
https://github.com/Dolibarr/dolibarr/blob/ae8b213d0497ed00371118c0c2b77b078fa8b151/htdocs/admin/limits.php#L96

Why?

I see the commit by @eldy "FIX Protection against bad value into accurancy setup" but why would showing less than what we store be bad?

Daviid-P:
Maybe it was supposed to be `>` instead of `<`?
```php
if (! $error && ((float) $valmainmaxdecimalsshown > $valmainmaxdecimalsunit || (float) $valmainmaxdecimalsshown > $valmainmaxdecimalstot)) {
```

eldy:
The question should be instead 
"Why showing on screen a value that is not the correct value" ?
This leads to serious bad experience: For example, users don't understand this: "Screen show me 3 lines of 1.23 and a total of the 3 lines of 3.70, I should see 3.69" !   But the 1.23 was in fact 1.234 and total 3.702. This is not a problem of rounding, but a bug of reality that is truncated.
If your monetary system needs 2 digits on total, so you must setup the accuracy of the total to 2 and not more (This is the rule in Euro lang:  total must be on 2 digits only). So Dolibarr will make the rounding to guarantee that total are always the total of lines (this is an important need for exact accountancy).

vdegrandpre:
@eldy there are times when calculated rounded totals do not match expected total.

In Québec, we have two taxes : 5% and 9,975%. because of that, when end-up having amounts that need to round on the third decimal. There are also times where amounts are spanned within more that two decimals.

Example :

We have an invoice that have the following definition
![image](https://github.com/user-attachments/assets/e2f807b5-35b4-44f7-91a2-325d20aeb00e)

But, as you do in France, we need to show the numbers only with 2 decimals. When configuring so, this happens:

![image](https://github.com/user-attachments/assets/9c1f35e0-8669-4344-8751-bdb03b889058)

Therefore difference of 0,01. This is problematic in many cases : 

- Invoices sent are not 100% correct
- To be correct, we'd need to display 3 decimals, which is incorrect by nature
- Syncing data to third-party accounting systems (QuickBooks Online) fails because of 0,01 $ difference
- Etc.

I've found a solution, making the following configuration : 

- In Limits and precision, filled in the following
[...]

PR #27266, opened 2024-01-01
Fix #27265 Limits and precision : bad control on MAIN_MAX_DECIMALS_SHOWN

### LINKED ISSUE #27265: Limits and precision : bad control on MAIN_MAX_DECIMALS_SHOWN
### Bug

In admin ->limit and precision, it's not allowed to set MAIN_MAX_DECIMALS_TOT>MAIN_MAX_DECIMALS_SHOWN
This throws an error and I don't understand why, I think the test is inverted
For example, you want the HT totals calculated with 5 digits (when applying VAT afterwards for example), but you want only 2 digits displayed; but that's now impossible

It would be more logical to throw and error if MAIN_MAX_DECIMALS_TOT < MAIN_MAX_DECIMALS_SHOWN . Or may be not comparing these 2 values at all 

### Environment Version

all (including develop branch)

### Environment OS

All

### Environment Web server

All

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

vmaury:
Fix #27265 Limits and precision : bad control on MAIN_MAX_DECIMALS_SHOWN

In admin ->limit and precision, it's not allowed to set MAIN_MAX_DECIMALS_TOT>MAIN_MAX_DECIMALS_SHOWN
This throws an error and I don't understand why, I think the test is inverted
For example, you want the HT totals calculated with 5 digits (when applying VAT afterwards for example), but you want only 2 digits displayed; but that's now impossible

It would be more logical to throw and error if MAIN_MAX_DECIMALS_TOT < MAIN_MAX_DECIMALS_SHOWN . Or may be **not comparing these 2 values at all : that's done in this PR**

### DIFF
--- a/htdocs/admin/limits.php
+++ b/htdocs/admin/limits.php
@@ -94,7 +94,7 @@ if ($action == 'update' && !$cancel) {
 		setEventMessages($langs->trans("ErrorValueCantBeNull", dol_trunc(dol_string_nohtmltag($langs->transnoentitiesnoconv("MAIN_MAX_DECIMALS_SHOWN")), 40)), null, 'errors');
 		$action = 'edit';
 	}
-	if (! $error && ((float) $valmainmaxdecimalsshown < $valmainmaxdecimalsunit) { // Fix #27265 : || (float) $valmainmaxdecimalsshown < $valmainmaxdecimalstot)
+	if (! $error && ((float) $valmainmaxdecimalsshown < $valmainmaxdecimalsunit)) { // Fix #27265 : || (float) $valmainmaxdecimalsshown < $valmainmaxdecimalstot)
 		$langs->load("errors");
 		$error++;
 		setEventMessages($langs->trans("ErrorValueForTooLow", dol_trunc(dol_string_nohtmltag($langs->transnoentitiesnoconv("MAIN_MAX_DECIMALS_SHOWN")), 40)), null, 'errors');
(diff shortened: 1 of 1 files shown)


vmaury:
Fix #24180 too !

eldy:
This is the expected behaviour.
If you want digit on 5 number for total, why do you want to show on 2 (it means you want to see on screen a value that is wrong).
Also, note that in europe, for total, the number of digit allowed must be 2 !
You probably want to have 5 digits on unit prices, not on total ?

vmaury:
I was obliged to set this directly in db (5 digits on totals, only 2 displayed) to solve my another problem of tva / ttc roundings (cf #27273) 
It doesn't seem anormal to me to calculate on 5 digits and display on 2, ok the values could be different, but not "wrong"
And the're 2 other people who noticed the same :
https://github.com/Dolibarr/dolibarr/issues/24180
https://github.com/Dolibarr/dolibarr/commit/509c85672e4f6510ff6dc6cc69146292280dcc2c#commitcomment-99154012

eldy:
> t doesn't seem anormal to me to calculate on 5 digits and display on 2,

This is already what we do: we calculate using the unit price accuracy (that is set by default on 5 digits) and we save the total in 2 digits (because this is the rule in europe that are using euro).
If you decide to bypass european rule by saving on 5 digits, you can, but in such a case, why showing a value that is the wrong value ? If decide to use rules to have total price on 5 digits like 1.23456 (some countries need 6 digits) then you must show 1.23456 not 1.23 that is the wrong value.
With this PR you introduce serious bugs in calculation and hidding them on screen is clearly not a solution.
Can you instead report the value you used and result on a screenshot to show that MAIN_MAX_DECIMALS_SHOWN does not do the job he should do ?

PR #27268, opened 2024-01-02
NEW : add constant TAKEPOS_CONFIRM_PAYMENT_CLOSE

altairis-tof:
add confirmation before payment close


htdocs/langs/en_US/cashdesk.lang
```diff
@@ -153,3 +153,4 @@ LineDiscount=Line discount
 LineDiscountShort=Line disc.
 InvoiceDiscount=Invoice discount
 InvoiceDiscountShort=Invoice disc.
+ConfirmClosePayment=Are you sure you want to delete this payment ?
```
eldy, line 156: The message ask to confirm to delete payment, but in code, the message is ask when we try to enter the payment popup so to create a payment (so the opposite). Did i miss something ?

altairis-tof: it replaces the colorbox.close function of the payment colorbox, so it it called when you close the payment box.

eldy:
So pr add a confirm popup showned when you close the payment box. It appears if you close the popup without trying to enter payments ? What is goal of adding a confirmation. If no payment entered, you can reopen it, can't you ?

Also it seems pr is not mergeable (conflict)

altairis-tof:
Hello, in my customer usecase we have an intermediate screen before payment to select equipments and warehouses (fork from charlene equipement module) so we dont want to lose those informations by error.

altairis-tof:
conflict solved

eldy:
So my understanding is this is a feature added to core to solve a problem due to a change in workflow after an external module addition. If an external module change a user interface workflow at a place and need to change it also at another place, this must be managed by the external module.
Usually, we can't add a too much business code into the core of Dolibarr if it has no sense for the Dolibarr itself. 
Here the change is light, so why not with a hidden option if change is generic code and we must find a name of the option that we are sure to understand what it does. 
For the moment i still don't understand the goal of this option for Dolibarr. the new code is executed when we Open the payment popup. But the hidden option is called "TAKEPOS_CONFIRM_PAYMENT_CLOSE" and it ask to confirm the closing of a popup we have just open. Also the message is "ConfirmClosePayment" ?
If need is purely for external module, a hook would be better...

altairis-tof:
ok i will try to manage it in the external module.

altairis-tof:
in fact i realize it was already done by my external module !

PR #27270, opened 2024-01-02
NEW : add constant VATUPDATE_NO_TRANSACTION

altairis-tof:
do not saturate server memory when vat update of thousands products


htdocs/product/admin/product_tools.php
```diff
@@ -65,7 +65,9 @@
 
 		$nbrecordsmodified = 0;
 
-		$db->begin();
+		if (!getDolGlobalInt('VATUPDATE_NO_TRANSACTION')) {
```
eldy, line 68: Why do you need this ?

PR #27272, opened 2024-01-02
NEW : add hook AFTER_VAT_UPDATE

altairis-tof:
allow external modules to do actions after mass vat update


htdocs/product/admin/product_tools.php
```diff
@@ -259,6 +259,14 @@
 			dol_print_error($db);
 		}
 
+		// add hook for external modules
+		$parameters = array('oldvatrate' => $oldvatrate, 'newvatrate' => $newvatrate);
+		$reshook = $hookmanager->executeHooks('AFTER_VAT_UPDATE', $parameters);
```
eldy, line 264: For Hook, the first param is a method. And we must follow the standard camelCase notation for methods, can you use 'afterVatUpdate' 
instead of 
'AFTER_VAT_UPDATE'

altairis-tof: ok i renamed the hook to follow standard.

PR #27274, opened 2024-01-02
Fix #27273 MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND doesn't work as expected

### LINKED ISSUE #27273: MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND doesn't work as expected
### Bug

 I wanted to use MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND=1 (tva calculated and rounded on general sums, cf "mode 2" described here https://wiki.dolibarr.org/index.php/R%C3%A8gles_de_param%C3%A9trage,_calcul_et_arrondi_de_TVA)

But it doesn't run as expected.

I examined the code of 
`function update_price(...)`
in commonobject.class.php

Even in mode 2 wih MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND=1, the sum of vat is calculated with sum of vat lines (rounded to 2 dec by default), and is eventualy corrected by $diff (see in the code).
That's seems to me unlogical.
I think the total(s) vat should be calculated at the end of the loop on lines using $total_ht_by_vats array * vat_rate

I suggest like in the followig PR to add another mode MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND=2 which use total_ht_by_vats array, so that it doesn't break existing configurations

Inspecting the code, I found in facture and in supplier facture some incoherencies : sometimes 
`update_price()`
is called
`update_price(1)`
some other times, it's called
`update_price(1, 'auto')`

As the default value for 2nd param of update_price is 'none', the invoice's total may vary when validated.

So I changed every call to 
`update_price(1, 'auto')`

### Environment Version

all (including develop branch)

### Environment OS

All

### Environment Web server

not applicable

### Environment PHP

all

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

vmaury:
# FIX #27273 MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND doesn't work as expected

Even in mode 2 wih MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND=1, the sum of vat is calculated with sum of vat lines (rounded to 2 dec by default), and is eventualy corrected by $diff (see in the code).
That's seems to me unlogical.
I think the total(s) vat should be calculated at the end of the loop on lines using $total_ht_by_vats array * vat_rate

I suggest like in the followig PR to add another mode MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND=2 which use total_ht_by_vats array, so that it doesn't break existing configurations

Inspecting the code, I found in facture and in supplier facture some incoherencies : sometimes
update_price()
is called
update_price(1)
some other times, it's called
update_price(1, 'auto')

As the default value for 2nd param of update_price is 'none', the invoice's total may vary when validated.

So I changed every call to
update_price(1, 'auto')

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -529,7 +529,7 @@ if (empty($reshook)) {
 		$object->fetch($id);
 		$object->revenuestamp = GETPOST('revenuestamp');
 		$result = $object->update($user);
-		$object->update_price(1);
+		$object->update_price(1, 'auto');
 		if ($result < 0) {
 			dol_print_error($db, $object->error);
 		} else {
@@ -1276,7 +1276,7 @@ if (empty($reshook)) {
 							}
 						}
 
-						$object->update_price(1);
+						$object->update_price(1, 'auto');
 					}
 				}
 
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -2646,7 +2646,7 @@ class Facture extends CommonInvoice
 
 			$lineid = $facligne->insert();
 			if ($lineid > 0) {
-				$result = $this->update_price(1);
+				$result = $this->update_price(1, 'auto');
 				if ($result > 0) {
 					// Create link between discount and invoice line
 					$result = $remise->link_to_invoice($lineid, 0);
@@ -3331,7 +3331,7 @@ class Facture extends CommonInvoice
 		$this->newref = dol_sanitizeFileName($num);
 
 		if ($num) {
-			$this->update_price(1);
+			$this->update_price(1, 'auto');
 
 			// Validate
 			$sql = 'UPDATE '.MAIN_DB_PREFIX.'facture';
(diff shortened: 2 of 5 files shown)


eldy:
If you calculate the vat at the end of the loop on lines using $total_ht_by_vats array * vat_rate, you will get the new value that may differ that first value calculated with method totalofround.
But if you just update the total with this value, you will introduce corrupted data, because the total of each line will not match anymore the total of the invoice.
So we use another way. When we found a difference, we update final total on invoice, but we also fix some lines because we need to always have total on each line that match final total (not doing this means it won't be possible to use accountancy module anymore due to data not consistent on each invoice).

vmaury:
Ok, ok.
It's a very tricky problem, in addition, I had it on an situation invoice. I'll close this issue, and wait to the new "situation invoice" feature to see what happens.
The initial problem came of incapable "fonctionnaires rond de cuir" who refuse invoices where total ht * vat_rate <> total_vat.
As I coded admin functionality to handle MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND in config screen "limits and precision" I'll make a new Feature Request and a new PR for this

PR #27279, opened 2024-01-02
Fix #27278 admin main roundoftotal not totalofround

### LINKED ISSUE #27278: Add admin feature to manage MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND
### Feature Request

Add an admin - config feature to handle MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND in config screen "limits and precision" 

### Use case

Now you're obliged to set the const directly in database, it would be convenient to manage this in config

### Suggested implementation

_No response_

### Suggested steps

_No response_

vmaury:
# FIX #27278 

Add an admin - config feature to handle MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND in config screen "limits and precision"

### DIFF
--- a/htdocs/admin/limits.php
+++ b/htdocs/admin/limits.php
@@ -108,7 +108,7 @@ if ($action == 'update' && !$cancel) {
 		dolibarr_set_const($db, $mainmaxdecimalsshown, $valmainmaxdecimalsshown, 'chaine', 0, '', $conf->entity);
 
 		dolibarr_set_const($db, $mainroundingruletot, $valmainroundingruletot, 'chaine', 0, '', $conf->entity);
-		
+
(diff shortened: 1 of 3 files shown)


vmaury:
RoundingMethod

eldy:
The MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND is an old option that is deprecated since several years because expected reliability in accountancy nowdays needs a rounding on line level not on final total. MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND must not be used anymore as with such an option it is not possible to have strict accounting. 
See page https://wiki.dolibarr.org/index.php?title=R%C3%A8gles_de_param%C3%A9trage,_calcul_et_arrondi_de_TVA#TVA_standard_et_r.C3.A8gles_d.27arrondi

PR #27284, opened 2024-01-03
Use unique keys to link extrafields tables, fixes #22117 and #22467

### LINKED ISSUE #22117: Indexes on extrafields should be UNIQUE

mc2contributor:
# CLOSE #22117 and #22467
It should not be possible to have more than one `extrafields` row associated with a row in the main table.

This pull request duplicates https://github.com/Dolibarr/dolibarr/pull/25116, but without the "ADD CONSTRAINT" commit that caused its rejection.  It has been rebased off of the latest `develop` branch.


eldy:
As this is a structure change (already freezed for v19 and lower), can you push the migration instructions into the file
19.0.0-20.0.0.sql  (i have just added it in develop branch) ?

mc2contributor:
Migration queries removed from 17.0.0-18.0.0.sql, added to 19.0.0-20.0.0.sql.
Also added unique index to `llx_mrp_production_extrafields`, which was added in 18.0.0.sql-19.0.0.sql.

PR #27285, opened 2024-01-03
NEW : allow search all to filter on extrafield on order list

atm-quentin:
# NEW allow search all to filter on extrafield on order list

new conf allowing user to search on extrafield during global search

### DIFF
--- a/htdocs/commande/list.php
+++ b/htdocs/commande/list.php
@@ -875,7 +875,7 @@ if (!empty($extrafields->attributes[$object->table_element]['label']) && is_arra
 if ($search_all) {
 	$sql .= ' LEFT JOIN '.MAIN_DB_PREFIX.'commandedet as pd ON c.rowid=pd.fk_commande';
 }
-if($search_all && getDolGlobalInt(strtoupper($object->table_element_line).'_SEARCHALL_EXTRAFIELDS_ENABLE') > 0) {
+if ($search_all && getDolGlobalInt(strtoupper($object->table_element_line).'_SEARCHALL_EXTRAFIELDS_ENABLE') > 0) {
 	$sql .= ' LEFT JOIN '.MAIN_DB_PREFIX.'commandedet_extrafields as efcd ON pd.rowid=efcd.fk_object';
 }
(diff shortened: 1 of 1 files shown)


htdocs/commande/list.php
```diff
@@ -163,6 +163,27 @@
 	$fieldstosearchall["c.note_private"] = "NotePrivate";
 }
 
+if (getDolGlobalInt(strtoupper($object->table_element).'_SEARCHALL_EXTRAFIELDS_ENABLE') > 0) {
```
eldy, line 166: According to the modulebuilder template we already have the extrafields for the table into the list of fields in the select. And the hook "completeFieldsToSearchAll" already exists to allow to complete the "$fieldstosearchall". It was just not added into list of order. I fix this.
Also the hook "printFieldListFrom" can add the  LEFT JOIN '.MAIN_DB_PREFIX.'commandedet_extrafields
So with hooks, we can already complete the search_all. Looks better and more generic than a hidden option _SEARCHALL_EXTRAFIELDS_ENABLE

atm-quentin: Yes, that seems like a better option, but to properly integrate it into the core, how would you proceed ?
I had considered using a tpl file, but based on your suggestion, that doesn't seem like the right solution.
There doesn't seem to be a hook file actions_commande.class.php.

eldy: Ther is nothing to do. The hook completeFieldsToSearchAll was added into commande/list.php 9 days ago.
So now you have all required hooks to have an external module add the fields of its choice (extra fields or fields from any other table) into the $fieldstosearchall, by using hooks.

atm-quentin: Ok, but you don't want to have this feature in dolibarr core ?

eldy: We may introduce such a feature in core. But if we all to use extrafields into the search all, we may decide which extrafields to use. Adding ALL may often make the feature useless.
Also, the order list is using the old school coding. If we introduce such a feature, it must first be added into the new school coding, so into modulebuilder template. Then and only after, we will see how to backport the generic code into old school pages.

PR #27287, opened 2024-01-03
fix warning php 8.2

john-botella:


eldy:
Note for @atm-john 
You should have a tool in the install/repair.php to upgrade the old path to the new path.

PR #27290, opened 2024-01-04
NEW allow [] in dol_eval

altairis-tof:
we need it to validate $objectoffield->array_options["options_myextrafield"] for instance to condition add tab in external module
ie : $this->tabs = array('order:-shipping::empty($object->array_options["options_rental_doc"])',);


htdocs/core/lib/functions.lib.php
```diff
@@ -9601,7 +9601,7 @@ function dol_eval($s, $returnvalue = 0, $hideerrors = 1, $onlysimplestring = '1'
 		if ($onlysimplestring == '1') {
 			// We must accept: '1 && getDolGlobalInt("doesnotexist1") && getDolGlobalString("MAIN_FEATURES_LEVEL")'
 			// We must accept: '$user->hasRight("cabinetmed", "read") && !$object->canvas=="patient@cabinetmed"'
-			if (preg_match('/[^a-z0-9\s'.preg_quote('^$_+-.*>&|=!?():"\',/@', '/').']/i', $s)) {
+			if (preg_match('/^a-z0-9\s'.preg_quote('^$_+-.*>&|=!?():"\',/@', '/').'/i', $s)) {
```
eldy, line 9604: If you remove the [ here, you change the regex
must not contains the char in list
into
must start with the a character and then must be followed by - then z, ...
Surely not what you want to do.

altairis-tof: Hello @eldy i think i need some help here; i've never understood how regexp work and i guess i never will...
Is it possible for you to fix the problem please ?

altairis-tof: ok i did it but i had to modify dol_eval call in verifCond function because it is the one called by complete_head_from_modules

altairis-tof: maybe best would be to add a parameter $onlysimplestring to verifCond with default '1' ?

eldy: Right, we should add $onlysimplestring with default to 1 (the more secured by default) to verifCond so we can call verifCond with param 2 for case we need it. Can you do it ?

altairis-tof: ok i've just done it; it is ready to merge. Thanks

PR #27315, opened 2024-01-06
Fix: mesgs to mesg variable naming for consistency.

mdeweerd:
# Fix Variablename to match function parameter.

Phan detects a potential bug because the variable name mismatches the parameter name

### DIFF
--- a/htdocs/compta/bank/transfer.php
+++ b/htdocs/compta/bank/transfer.php
@@ -189,13 +189,13 @@ if ($action == 'add' && $user->hasRight('banque', 'transfer')) {
 					$error++;
 				}
 				if (!$error) {
-					$mesgs = $langs->trans("TransferFromToDone", '{s1}', '{s2}', $amount[$n], $langs->transnoentitiesnoconv("Currency".$conf->currency));
-					$mesgs = str_replace('{s1}', '<a href="bankentries_list.php?id='.$tmpaccountfrom->id.'&sortfield=b.datev,b.dateo,b.rowid&sortorder=desc">'.$tmpaccountfrom->label.'</a>', $mesgs);
-					$mesgs = str_replace('{s2}', '<a href="bankentries_list.php?id='.$tmpaccountto->id.'">'.$tmpaccountto->label.'</a>', $mesgs);
-					setEventMessages($mesgs, null, 'mesgs');
+					$mesg = $langs->trans("TransferFromToDone", '{s1}', '{s2}', $amount[$n], $langs->transnoentitiesnoconv("Currency".$conf->currency));
+					$mesg = str_replace('{s1}', '<a href="bankentries_list.php?id='.$tmpaccountfrom->id.'&sortfield=b.datev,b.dateo,b.rowid&sortorder=desc">'.$tmpaccountfrom->label.'</a>', $mesg);
+					$mesg = str_replace('{s2}', '<a href="bankentries_list.php?id='.$tmpaccountto->id.'">'.$tmpaccountto->label.'</a>', $mesg);
+					setEventMessage($mesg);
 				} else {
 					$error++;
-					setEventMessages($tmpaccountfrom->error.' '.$tmpaccountto->error, null, 'errors');
+					setEventMessage($tmpaccountfrom->error.' '.$tmpaccountto->error, 'errors');
 				}
 			}
 		}


htdocs/compta/bank/transfer.php
```diff
@@ -189,10 +189,10 @@
 					$error++;
 				}
 				if (!$error) {
-					$mesgs = $langs->trans("TransferFromToDone", '{s1}', '{s2}', $amount[$n], $langs->transnoentitiesnoconv("Currency".$conf->currency));
-					$mesgs = str_replace('{s1}', '<a href="bankentries_list.php?id='.$tmpaccountfrom->id.'&sortfield=b.datev,b.dateo,b.rowid&sortorder=desc">'.$tmpaccountfrom->label.'</a>', $mesgs);
-					$mesgs = str_replace('{s2}', '<a href="bankentries_list.php?id='.$tmpaccountto->id.'">'.$tmpaccountto->label.'</a>', $mesgs);
-					setEventMessages($mesgs, null, 'mesgs');
```
frederic34, line 195: why not `setEventMessage($mesgs);`

htdocs/compta/bank/transfer.php
```diff
@@ -189,10 +189,10 @@
...
+					$mesg = $langs->trans("TransferFromToDone", '{s1}', '{s2}', $amount[$n], $langs->transnoentitiesnoconv("Currency".$conf->currency));
+					$mesg = str_replace('{s1}', '<a href="bankentries_list.php?id='.$tmpaccountfrom->id.'&sortfield=b.datev,b.dateo,b.rowid&sortorder=desc">'.$tmpaccountfrom->label.'</a>', $mesg);
+					$mesg = str_replace('{s2}', '<a href="bankentries_list.php?id='.$tmpaccountto->id.'">'.$tmpaccountto->label.'</a>', $mesg);
+					setEventMessages($mesg, null, 'mesg');
```
frederic34, line 195: image

mdeweerd: 1. The first argument as defined in the the method definition of setEventMessages is "$mesg".  Phan triggers when the variable provided as the argument in the function call is $mesgs which is the second argument of the method is "mesgs".  So it flags a suspicious arguments order or position.
2. I try to modify as little as possible - possibly setEventMessages($mesg) is ok, but the default style is 'mesgs', and the original call requests 'mesg' as the style.

frederic34: did you try setEventMessage without 's', default style is mesgs

mdeweerd: Ok, sorry, I see the issue now, my replacement also replaced the third parameter to 'mesg' while it was 'mesgs' which is the default.

So I further simplified the statement now as you suggested.

htdocs/compta/bank/transfer.php
```diff
@@ -189,10 +189,10 @@
...
 				}
 				if (!$error) {
-					$mesgs = $langs->trans("TransferFromToDone", '{s1}', '{s2}', $amount[$n], $langs->transnoentitiesnoconv("Currency".$conf->currency));
-					$mesgs = str_replace('{s1}', '<a href="bankentries_list.php?id='.$tmpaccountfrom->id.'&sortfield=b.datev,b.dateo,b.rowid&sortorder=desc">'.$tmpaccountfrom->label.'</a>', $mesgs);
-					$mesgs = str_replace('{s2}', '<a href="bankentries_list.php?id='.$tmpaccountto->id.'">'.$tmpaccountto->label.'</a>', $mesgs);
-					setEventMessages($mesgs, null, 'mesgs');
+					$mesg = $langs->trans("TransferFromToDone", '{s1}', '{s2}', $amount[$n], $langs->transnoentitiesnoconv("Currency".$conf->currency));
+					$mesg = str_replace('{s1}', '<a href="bankentries_list.php?id='.$tmpaccountfrom->id.'&sortfield=b.datev,b.dateo,b.rowid&sortorder=desc">'.$tmpaccountfrom->label.'</a>', $mesg);
+					$mesg = str_replace('{s2}', '<a href="bankentries_list.php?id='.$tmpaccountto->id.'">'.$tmpaccountto->label.'</a>', $mesg);
+					setEventMessages($mesg);
```
frederic34, line 195: if you use one parameter you need to use setEventMessage without s, not setEventMessages with s

mdeweerd: Ok, the original call was setEventMessages, so I changed it to setEventMessage and as I noticed that the error message was also a single string, I changed that to setEventMessage as well.

PR #27333, opened 2024-01-08
Fix: Change TCPDF::_out to public because of pdf.lib.php

mdeweerd:
# Fix: Change TCPDF::_out to public because of pdf.lib.php

TCPDF::_out is used 2 times as ->_out(...) in htdocs/core/lib/pdf.lib.php.
Changing php.lib.php requires more analysis.  Making TCPDF::_out public is
the quick solution.

### DIFF
--- a/dev/dolibarr_changes.txt
+++ b/dev/dolibarr_changes.txt
@@ -65,6 +65,20 @@ NUSOAP:
 
 TCPDF:
 ------
+* Modify in tcpdf.php: make TCPDF::_out public.
+  (htdocs/core/lib/pdf.lib.php uses it as a public method)
+
+  Change:
+      protected function _out($s)
+  to
+      public function _out($s)
+
+  Change in method's _out phpdoc:
+
+      * @protected
+  to
+      * @public
+
 * Replace in tcpdf.php:
             if (isset($this->imagekeys)) {
                 foreach($this->imagekeys as $file) {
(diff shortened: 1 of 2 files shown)


eldy:
This generates trouble ? Can you describe the use case that generates the bug ?

mdeweerd:
Here is a standalone example:

test.php:

```php
<?php

class TC1 {
    protected function HelloWorld() {
        print "Hello World!";
    }
}


$tc = new TC1();

$tc->HelloWorld();
```

PHP version:

```text
$ php --version
PHP 8.1.13 (cli) (built: Nov 22 2022 15:49:14) (ZTS Visual C++ 2019 x64)
Copyright (c) The PHP Group
Zend Engine v4.1.13, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.13, Copyright (c), by Zend Technologies
    with Xdebug v3.1.6, Copyright (c) 2002-2022, by Derick Rethans
```


```text
$ php test.php

Fatal error: Uncaught Error: Call to protected method TC1::HelloWorld() from global scope in .\dolibarr\test.php on line 12
```

To check if with "$conf->global->FACTURE_DRAFT_WATERMARK)" set and generating an invoice on PHP8.1  this does result in a Fatal Error.

eldy:
This happen only when option MAIN_DISABLE_TCPDI is on. In default mode, the protected function is not a problem as it is used by a parent class.

PR #27341, opened 2024-01-08
Fix: Add missing argument in getOnlineSignatureUrl

mdeweerd:
# Fix: Add missing argument in getOnlineSignatureUrl.

The function requires 5 arguments and only had 4.


eldy:
Note that such dev error does not generates "user visible bugs", so we must push this into develop branch as only visible bug fixes should be pushed in old maintenance versions.

PR #27343, opened 2024-01-08
Fix: Sitemap generation - fix invalid replacement regex

mdeweerd:
# Fix: Sitemap generation - fix invalid replacement regex

The regex is invalid because the delimiters are slashes which are also present in the text to match.
Fixed by changing the delimiters.


eldy:
As we are using the / for regex everywhere, i modified your fix to keep on using the / in this case too.

mdeweerd:
@eldy Just to note, '/' is not used "everywhere" - I stumbled on different ones here:

https://github.com/Dolibarr/dolibarr/blob/79e9caaf5daf3b1c416eaf4ca20237abfa7bd951/htdocs/core/class/extrafields.class.php#L1444-L1458

Issue #27064, opened 2023-12-12
New signed status for interventions


W1W1-M:
### Feature Request

Currently _Interventions_ can have **Draft**, **Validated** or **Done** status. They can be signed by the client (paper or online). In the same way quotes have a **signed** (or **unsigned**) status, interventions could too.

### Use case

For example: an intervention with a **signed** status gives extra information to the person in charge of billing or organizing team. This is an extra guarantee that the intervention is complete and can be properly finalized and billed. 

### Suggested implementation

Add an extra status with associated logic.

### Suggested steps

- [x] Add data structure
- [x] Add new signed status
- [x] Add _Set signed_ button (like Accepted button for quotes) & associated logic
- [x] Add notification for signed status
- [x] Add logic for status update with online signing
- [x] Update intervention module Wiki page: https://wiki.dolibarr.org/index.php?title=Module_Interventions

lmag:
- [ ] Add signed status company
- [ ] Add signed status thirdparty
- [ ] Add pdf filename sign company
- [ ] Add pdf filename thirdparty

FHenry:
suggest 3 value possible for signed field
1 => sign by senders(company)
2 => sign by receiver (thridparty)
3 => Sign by both

eldy:
> suggest 3 value possible for signed field 1 => sign by senders(company) 2 => sign by receiver (thridparty) 3 => Sign by both

Agree. Signed status must be a separated field. Ok for the 3 values.

W1W1-M:
Feel free to work on this. I haven't done anymore since #27856

W1W1-M:
Thanks to @eldy @evarisk-micka @FHenry @lmag @mdeweerd for all your input, work and help on this 😄

PR #27346, opened 2024-01-09
NEW intervention signed status

W1W1-M:
# NEW intervention signed status
This PR is for #27064, **Signed** status, button and associated logic have been added. Email notifications can be set. Localization is set up for translation.

However I can't seem to get the event to be recorded in the calendar. 
At the moment this code does not include updating the status when intervention is signed online.

Any feedback/help would be appreciated.

Thank you


eldy:
Do we really need a status for a signature. It appears that the signature can be done at different step, so should be an extra status "signed / not signed". In the workflow you entered, can we switch from VALIDATED to CLOSED directly (in a lot of cases, users won't need to have a sign step) ?

W1W1-M:
> Do we really need a status for a signature. It appears that the signature can be done at different step, so should be an extra status "signed / not signed".

I agree that **signed / not signed** is extra information to the workflow status. Maybe I can refactor to add this as a complementary information to **validated** or **done** statues. For example: **Validated (Signed)**.

> In the workflow you entere, can we switch from VALIDATED to CLOSED directly (in a lot of cases, users won't need to have a sign step) ?

Yes, nothing stops the user from directly using **classify done** to go from **validated** to **done** (closed).

eldy:
Status signed/Not signed must be managed by the new field "signed" into intervention table. So this PR is no more valid.

PR #27353, opened 2024-01-09
FIX : bad test on socid

defrance:
never work if no user id send (aka -1 on first param)


htdocs/core/class/html.formintervention.class.php
```diff
@@ -74,7 +74,7 @@ public function select_interventions($socid = -1, $selected = 0, $htmlname = 'in
 		$sql = "SELECT f.rowid, f.ref, f.fk_soc, f.fk_statut";
 		$sql .= " FROM ".$this->db->prefix()."fichinter as f";
 		$sql .= " WHERE f.entity = ".$conf->entity;
-		if ($socid != '') {
+		if ($socid  >0) {
```
eldy, line 77: Just after we have 
`if ($socid == '0') {`
that seems broken

htdocs/core/class/html.formintervention.class.php
```diff
@@ -74,12 +74,10 @@ public function select_interventions($socid = -1, $selected = 0, $htmlname = 'in
...
 		$sql .= " FROM ".$this->db->prefix()."fichinter as f";
 		$sql .= " WHERE f.entity = ".$conf->entity;
-		if ($socid != '') {
-			if ($socid == '0') {
-				$sql .= " AND (f.fk_soc = 0 OR f.fk_soc IS NULL)";
-			} else {
+		if ($socid  > 0) {
 				$sql .= " AND f.fk_soc = ".((int) $socid);
-			}
+		} else {
```
eldy, line 79: Thus seems to break the case when $socid = -1, where we should not have any test in such a case.

eldy:

Issue #26230, opened 2023-10-16
Resource module improvements 


W1W1-M:
### Feature Request

The resource module could use some improvements to make it more usable:

- The possibility to add a main **photo** for the resource (like the logo on third-party)
- Some resources such as rooms or venues have a generic **email**, **telephone number**  and/or **URL** not directly linked to a contact or third-party
- To know which resource to use it's good to know how many **people** it can hold or can use it simultaneously 
- To be able to use a resource it's important to know where it is : **address**
- It could be useful to be able to link an event(project) directly to a resource without using the agenda. This could maybe replace the location string on an event card.

The following fields would need to be added to the data structure of llx_resource:
- Address
- Postal code
- City
- Photo file name
- Number of places
- Telephone
- Email
- URL

Any feedback would be helpful in order to better qualify what is exactly expected of this module.

This also addresses issue #22121

Dolibarr developer and user documentation will also need to be updated.

### Use case

A resource can be used in many different cases so it's important to keep resources as generic as possible. 

### Suggested steps (Pull Requests)

- [x] Data structure
- [x] Full address (in progress)
- [ ] Photo
- [x] Resource details (places, telephone & email)
- [x] Resource URL
- [x] List ajustements for new fields
- [ ] Link to projects / events

W1W1-M:
Data structure will also need a foreign key to address state/province (département)

prietojc:
Could the status label be included in the cards?. The status field already exists in the structure. Thanks.

prietojc:
I have proposed PR  #33041

JonBendtsen:
Would be great if this was re-opened

JonBendtsen:
> It could be useful to be able to link an event(project) directly to a resource without using the agenda. This could maybe replace the location string on an event card.

I am not sure I would do that @W1W1-M  because then each event(project) can only have 1 resource and/or each resource can only have 1 event(project).

Wouldn't it be better to add something like this table for booking a resource?
```
CREATE TABLE llx_eventorganization_resource_booking (
    rowid integer AUTO_INCREMENT PRIMARY KEY,
    fk_resource integer NOT NULL,          -- Link to llx_resource (Room/Equip)
    fk_actioncomm integer NOT NULL,        -- Link to llx_actioncomm (Session)
    fk_project integer NOT NULL,
    date_start datetime NOT NULL,          -- Booking Start
    date_end datetime NOT NULL,            -- Booking End
[...]

PR #27355, opened 2024-01-09
NEW resource address

W1W1-M:
# NEW resource address
As proposed in #26230 this adds full address to a resource


htdocs/resource/class/dolresource.class.php
```diff
@@ -98,69 +110,65 @@ class Dolresource extends CommonObject
...
-	 *  @param	DoliDb		$db      Database handler
+	 *  @param DoliDb $db      Database handler
 	 */
-	public function __construct($db)
+	public function __construct(DoliDb $db)
 	{
 		$this->db = $db;
+		$this->element = 'dolresource';
+		$this->table_element = 'resource';
+		$this->picto = 'resource';
```
eldy, line 120: Why moving the value of final properties into the constructor ? They are properties that should not be modified, so no reason to need to instantiate an instance to get their value.

W1W1-M: I moved the properties to constructor to centralize variable initialization and make good use of the class constructor, just a coding preference. 

However if these properties should not be modified and are to be accessed without instantiating maybe the better solution is to declare them clearly as class constants ?

W1W1-M: But then I suppose they would have to be redefined as constants for CommonObject and all other child classes ...

eldy: Yes, they should have been set as const. But this was coded a long time ago and changing property to const now is too complex and not a priority. 
In a future we may use the declaration with final https://www.php.net/manual/fr/language.oop5.final.php
public final string $element = 'abcdef';
but we need to wai min php 8.1 for that. For the moment, we should keep them in declaration.

W1W1-M: Thanks

PR #27373, opened 2024-01-09
FIX: Update mod_codeclient_* constructor to have $db argument as in callers.

mdeweerd:
# FIX: Remove  parameter in call to mod_codeclient_* constructor.

None of the mod_codeclient_* constructors has an argument.
Reported by phan as `PhanParamTooMany: Call with 1 arg(s) to \mod_codeclient_leopard::__construct() which only takes 0 arg(s) defined at htdocs/core/modules/societe/mod_codeclient_leopard.php:69`
in all modified files.
One exception in the card.php file because the code has no explicit detail,
but it also uses a mod_codeclient_ constructor.

### DIFF
--- a/htdocs/core/modules/societe/mod_codeclient_elephant.php
+++ b/htdocs/core/modules/societe/mod_codeclient_elephant.php
@@ -89,8 +89,10 @@ class mod_codeclient_elephant extends ModeleThirdPartyCode
 
 	/**
 	 *	Constructor
+	 *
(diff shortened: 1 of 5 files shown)


eldy:
A better fix would be to modify the constructor to accept the $db as it is what we do for all submodule factory classes.

mdeweerd:
Ok, that was the other option - PR is modified.

PR #27382, opened 2024-01-10
Fis: dolibar -> dolibarr

mdeweerd:
# Fis: dolibar -> dolibarr

A tool should know how to write its own name 😉 .


mdeweerd:
@eldy 
I made requests on transifex to go and modify or suggest these changes.

However, bg_BG was refused, so I guess you have superpowers to update that.

> Your request to add Bulgarian to the project Dolibarr ERP &amp; CRM software was declined

Issue #27376, opened 2024-01-10
Missing '$sql =' before dangling filter condition in VAT filter for product_tools.php


mdeweerd:
### Bug

Phan reports:
```text
Unused result of a string literal " AND default_vat_code = IS NULL" near this line
```

The condition probably needs to be applied but it has not been since 6.0.2 (Fix #7473).

https://github.com/Dolibarr/dolibarr/blame/f45f5c3291b060bdd25df3777b5957a5e5e00317/htdocs/product/admin/product_tools.php#L96-L100

https://github.com/Dolibarr/dolibarr/blame/f45f5c3291b060bdd25df3777b5957a5e5e00317/htdocs/product/admin/product_tools.php#L198-L202

### Environment Version

6.0.2 to develop (since #7473)

altairis-tof:
hello, i've just commit a fix.

altairis-tof:
@mdeweerd you can close the issue please

mdeweerd:
Yes, I am surprised it was not closed automatically by the integration in develop.

PR #27417, opened 2024-01-11
fix old issue 

altairis-tof:
https://github.com/Dolibarr/dolibarr/issues/27376


mdeweerd:
Actually, there is still one of the two lines dangling.

I did and do not want to apply the apparent fix myself since it requires to know if this condition is needed as it was there since so many years.

PR #27424, opened 2024-01-11
FIX - Add condition to delete product

atm-lucasmantegari:
# FIX - *Add condition to delete product*
We can currently delete a product even though it is in a BOM.
Adding this check helps avoid an SQL error when using this product on a OF.


atm-lucasmantegari:
@eldy I think Travis has a bug. If not, please explain to me

atm-lucasmantegari:
@eldy

eldy:
> @eldy 

See my first comment..

fappels:
See https://github.com/Dolibarr/dolibarr/pull/30140/commits/f401ad5e9bcd9658b2a4ada71557dff4a303e013
Is for MO, also BOM need to be added.

atm-lucasmantegari:
@eldy its better like this ?

eldy:
> @eldy its better like this ?

Seems better. But have cti errors

PR #27427, opened 2024-01-11
Qual: Spelling htdocs/[d-k]*

mdeweerd:
# Qual: Fix spelling htdocs/[d-k]*

Highlights:

- htdocs/install/doctemplates/websites/website_template-restaurant/containers/page249.tpl.php
  Potential functional change (fix) because toekn was changed to token: `<input type=hidden name=toekn`

- htdocs/install/upgrade2.php $filles changed to $children

- htdocs/intracommreport/class/intracommreport.class.php
  Local $enveloppe changed to $envelope.


mdeweerd:
Note: phpcs fails because there are too many files for it in the current workflow action - it is ok when run locally.

htdocs/install/doctemplates/websites/website_template-homesubmenu/containers/page202.tpl.php
```diff
@@ -128,7 +128,7 @@ class="dropdown-menu"
                         <p>
                             Lorem ipsum dolor, sit amet consectetur adipisicing
                             elit. Ab fuga nobis omnis alias, aliquid iste cumque
-                            tempora nam reprehenderit quia itaque debitis,
+                            tempora name reprehenderit quia itaque debitis,
```
eldy, line 131: Possible to exclude this as this is latin and not english ?

mdeweerd: It's "Lorem Ipsum", not actually latin.

Codespell does not allow precise local exclusions: one needs to exclude the word or the file.

To me this was not important, and I prefer checking the rest of the file rather than excluding the file or the suggestion for something that could easily happen ("nam").

Another solution is to change "nam" to "nim".  Codepsell does not have a suggestion for that.

mdeweerd: I changed it to nim.  (+squash & rebase).

PR #27448, opened 2024-01-11
Qual: Spelling outside htdocs

mdeweerd:
# Qual: Fix spelling of files not in the htdocs subdirectory.

Highlight:
- Change in email test from unvalid to invalid that should give the same test result.

### DIFF
--- a/test/soapui/Dolibarr-soapui-project.xml
+++ b/test/soapui/Dolibarr-soapui-project.xml
@@ -3637,7 +3637,7 @@
         <xsd:all>
           <xsd:element name="type" type="xsd:string"/>
           <xsd:element name="priority" type="xsd:string"/>
-          <xsd:element name="connexion" type="xsd:string"/>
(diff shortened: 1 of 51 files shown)


test/soapui/Dolibarr-soapui-project.xml
```diff
@@ -3637,7 +3637,7 @@
         <xsd:all>
           <xsd:element name="type" type="xsd:string"/>
           <xsd:element name="priority" type="xsd:string"/>
-          <xsd:element name="connexion" type="xsd:string"/>
```
eldy, line 3640: This is a field name choosed in the protocol of this xml file so we should not modify it.

mdeweerd: 4 eyes are better than 2 ;-) .  I scrutenized the changes several times to revert those, but I missed this one.

Issue #27453, opened 2024-01-12
Improvement in the management of batch numbers and serial numbers


lmag:
### Feature Request

- [ ] add a check when a product is managed by batches in serial number management (UC1)
- [ ] Proposal allowing the numbers to be split into serial numbers if we wish to validate the change of status (UC1)
- [ ] Separation of batch number and series number fields(UC2)
- [ ] Add option to see all serial number of one lot (UC3)

### Use case 1
![image](https://github.com/Dolibarr/dolibarr/assets/3288514/53e886e4-9695-4b52-934a-eb932e7d6947)

### Use Case 2
Missing field for lot : 
![image](https://github.com/Dolibarr/dolibarr/assets/3288514/4a75f9d4-8975-419d-89ff-ebac2393b5ea)

Should be
![image](https://github.com/Dolibarr/dolibarr/assets/3288514/87d05770-dff9-4157-8e2e-cf811d8aa268)

### Use Case 3
![image](https://github.com/Dolibarr/dolibarr/assets/3288514/23317dc4-24b0-44ee-a377-c6c86ef02ebd)

lmag:
voir aussi : https://github.com/Dolibarr/dolibarr/issues/27456

PR #27454, opened 2024-01-12
feat: check number of batch number before change rule on items (setting option in batch setting module)

FHenry:
first point of #27453


eldy:
I don't think we need such an option :
There is already a protection to block any stock movement on a product in serial mode if "at end of any movement" the qty is not 1. So the user has no choice. He can switch the option and he has no choice to fix the qty just after. Be if he do an error during the switch (when seeing he must change the qty in stock, may be he will think oh my good, i made an error during the swith, it is not what i wanted), then he can switch back to lot.
So everything is already secured.

PR #27455, opened 2024-01-12
feat: prepare BDD to have batch number on product with serial management

FHenry:
#27456


fappels:
Think current 'no lot / lot / lot or serial' flag seems not so good solution. We should have had better a separate serial flag and have new object ProductSerial.

altairis-tof:
> Think current 'no lot / lot / lot or serial' flag seems not so good solution. We should have had better a separate serial flag and have new object ProductSerial.

yes there are several discussion on this subject : https://github.com/Dolibarr/dolibarr/issues/27456
and https://github.com/Dolibarr/dolibarr/issues/27453

fappels:
> > Think current 'no lot / lot / lot or serial' flag seems not so good solution. We should have had better a separate serial flag and have new object ProductSerial.
> 
> yes there are several discussion on this subject : #27456 and #27453

@altairis-tof thanks, deleted and moved to #27456. Also changed my mind.

PR #27457, opened 2024-01-12
FIX : fix pdf number of pages

altairis-tof:
it was printed 1/1 on first page, 2/2 on second, 3/3 on third and so on


eldy:
The opposite PR was done due to bug in getAliasNbPages();
So i restore and introduced a hidden const. I can't reproduce your trouble on standard version. Do you use external modules that manipulate PDF ?

altairis-tof:
> The opposite PR was done due to bug in getAliasNbPages(); So i restore and introduced a hidden const. I can't reproduce your trouble on standard version. Do you use external modules that manipulate PDF ?

yes

Issue #27458, opened 2024-01-12
Missing hook main.inc inf fonction top_htmlhead


lmag:
### Feature Request

Missing hook main.inc inf fonction top_htmlhead

altairis-tof:
top_htmlhead is called by main.inc.php so your issue is strange

altairis-tof:
in fact top_htmlhead is a function inside main.inc.php

altairis-tof:
@lmag there is already a hook addHtmlHeader at the end of the function; do you want another one at the beginnning to replace header ?

FHenry:
@altairis-tof Yes that it, for replace the current manifest.json force by main.inc.php

FHenry:
I close the issue as the PR was merged

PR #27459, opened 2024-01-12
feat: add hook in main.inc.php to override the manifest.json

FHenry:
In some case module need to change the PWA descriptor


htdocs/main.inc.php
```diff
@@ -1719,6 +1719,13 @@ function top_htmlhead($head, $title = '', $disablejs = 0, $disablehead = 0, $arr
 
 		// Mobile appli like icon
 		$manifest = DOL_URL_ROOT.'/theme/'.$conf->theme.'/manifest.json.php';
+		$parameters = array('manifest'=>$manifest);
+		$resHook = $hookmanager->executeHooks('setManifest', $parameters); // Note that $action and $object may have been modified by some hooks
```
eldy, line 1723: For new introduced hook we must now start the method name with hook... Can you rename into
hookSetManifest

FHenry:
fix #27458

PR #27461, opened 2024-01-12
Fix PHPStan

W1W1-M:
# Fix PHPStan
htdocs/resource/class/dolresource.class.php | 512 | Default value of the parameter 5 $filter (string) of method Dolresource::fetchAll() is incompatible with type array.

htdocs/resource/class/html.formresource.class.php | 82 | Default value of the parameter 1 $selected (string) of method FormResource::select_resource_list() is incompatible with type int.

### DIFF
--- a/htdocs/resource/class/html.formresource.class.php
+++ b/htdocs/resource/class/html.formresource.class.php
@@ -65,9 +65,9 @@ class FormResource
 	/**
 	 *  Output html form to select a resource
 	 *
-	 *	@param	int   	$selected       Preselected resource id
-	 *	@param  string	$htmlname       Name of field in form
-	 *  @param  string	$filter         Optionnal filters criteras (example: 's.rowid <> x')
+	 *	@param	int		$selected		Preselected resource id
+	 *	@param	string	$htmlname		Name of field in form
+	 *  @param	array	$filter			Optionnal filters criteras (example: 's.rowid <> x')
 	 *	@param	int		$showempty		Add an empty field
 	 * 	@param	int		$showtype		Show third party type in combolist (customer, prospect or supplier)
 	 * 	@param	int		$forcecombo		Force to use combo box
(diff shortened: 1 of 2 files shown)


frederic34:
image
$filtertype is a string, $filterarray is not used

W1W1-M:
> $filtertype is a string, $filterarray is not used

@frederic34 I agree, in the method **select_types_resource** it is a string but the correction is for **select_resource_list** method where filter needs to be an array for **fetchAll** method. The **resources_used** variable does however seem unused and could be removed, wasn't sure of its use.

frederic34:
It looks like a unfinished feature

htdocs/resource/class/html.formresource.class.php
```diff
@@ -76,10 +76,10 @@ public function __construct($db)
 	 *  @param	int		$outputmode		0=HTML select string, 1=Array, 2=without form tag
 	 *  @param	int		$limit			Limit number of answers
 	 *  @param	string	$morecss		More css
-	 * 	@param	bool	$multiple       add [] in the name of element and add 'multiple' attribut
+	 * 	@param	bool	$multiple		add [] in the name of element and add 'multiple' attribut
 	 * 	@return	string|array			HTML string with
 	 */
-	public function select_resource_list($selected = '', $htmlname = 'fk_resource', $filter = '', $showempty = 0, $showtype = 0, $forcecombo = 0, $event = array(), $filterkey = '', $outputmode = 0, $limit = 20, $morecss = '', $multiple = false)
+	public function select_resource_list($selected = '', $htmlname = 'fk_resource', $filter = [], $showempty = 0, $showtype = 0, $forcecombo = 0, $event = array(), $filterkey = '', $outputmode = 0, $limit = 20, $morecss = '', $multiple = false)
```
eldy, line 82: Correct fix seems 
`public function select_resource_list($selected = 0, $htmlname = 'fk_resource', $filter = '', $showempty = 0, $showtype = 0, $forcecombo = 0, $event = array(), $filterkey = '', $outputmode = 0, $limit = 20, $morecss = '', $multiple = false)`
phpstan error is reported for parameter $selected

PR #27469, opened 2024-01-12
fix phpdoc

frederic34:


eldy:
@frederic34 I have finished the migration of bool notrigger into int notrigger...

PR #27477, opened 2024-01-12
fix select_produits when $status_purchase == -1

rycks:
There is a "race condition" with new parameter introduced into dolibarr 16.0 with select_produits function.

Here is what we can read into documentation:

```
	 *  @param		int		$status_purchase	Purchase status -1=Return all products, 0=Products not on purchase, 1=Products on purchase
```

but in fact if $status_purchase=1 sql request does not returns all products !

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -2638,10 +2638,10 @@ class Form
 			$sql .= " AND p.finished = ".((int) $finished);
 		} elseif ($finished == 1) {
 			$sql .= " AND p.finished = ".((int) $finished);
-			if ($status >= 0) {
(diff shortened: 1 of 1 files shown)


eldy:
sell status and purchase status are 2 different status for product.

The good text for purchase_status parameter in doc should be  : 
Purchase status -1=No filter on purchase status, 0=Products not on purchase, 1=Products on purchase
so current code seems good. If value is -1 no filter on purchase status is used, but filter on other may apply

I tried a better fix. Can you test it on your side ?

rycks:
arf, i learn we can change documentation :-) let me try but it seems good (and better), please check/update documentation on all functions like select_produits and select_produits_list ...

PR #27486, opened 2024-01-13
Qual: Fix some more spelling in the translations

mdeweerd:
Fix some occurences of dolibarr misspellings that were not catched before.

I do not know if these are synced with transifex - if not somebody will need to update these 14 occurences there.


eldy:
PR on files into langs/xx_XX where xx_XX is not en_US is useless because files are erased by Transifex translation of en_US. Change on translation files must be done with PR for en_US only. For other languages (en_NZ, es_ES, ...), fix of translation must be done directly on https://www.transifex.com/dolibarr-association/dolibarr/

PR #27503, opened 2024-01-14
Try to move files from root in dedeicated dir

eldy:
Try to move files from root in dedeicated dir


mdeweerd:
It is better to have them in the root (or at least a version that includes them) because then it is possible to run the tools without specifying where the configuration is.

The namse can be:
- `.php-cs-fixer.dist.php`;
- `.phpcs.xml.dist`

For phpstan, I can't see that a configuration with a '.' is allowed, but the file in the root directory could include another one:
https://phpstan.org/config-reference#multiple-files

and it should be called `phpstan.neon.dist` so that the user can create `phpstan.neon` for local changes.

PR #27505, opened 2024-01-14
Test move option in the precommit config file

eldy:
Test move option in the precommit config file


mdeweerd:
@eldy 
The purpose of keeping the options as much as possible in pyproject.toml is that they are detected when running `codespell`from the command line.

However, there is the choice to use `.codespellrc`.

PR #27508, opened 2024-01-14
Timezone godthab rename

JonBendtsen:
# Qual - Timezone godthab rename

rename for deprecated timezone America/Godthab to America/Nuuk
https://data.iana.org/time-zones/tzdb/NEWS
https://www.php.net/manual/en/timezones.others.php


eldy:
I have removed the lib jzts that was not necessary anymore and i have upgrade the lib for Sabre so the PR is no more valid.
for a final fix, it would be better to push the change in timezone file to the sabre project:
https://github.com/sabre-io/vobject

PR #27522, opened 2024-01-14
add function isDolTms

frederic34:


htdocs/core/lib/functions.lib.php
```diff
@@ -234,6 +234,25 @@ function isModEnabled($module)
...
 
+/**
+ * isDolValidTms check if a timestamp is valid.
+ *
+ * @param  int|string|null $timestamp timestamp to check
+ * @return bool
+ */
+function isDolValidTms($timestamp)
+{
+	if ($timestamp == '') {
```
eldy, line 245: May be 
`if ($timestamp === '') {`
here ?

frederic34: I also use a shorter nam

PR #27537, opened 2024-01-15
NEW: cronjob: deny to launch a method from a deactivated module

libremaster:
A deactivated module should remove cronjobs but in some cases, a user can add by himself a cron calling a method from a class.

With this feature, there is a check if this method is from a deactivated module. In this case, a denied error is raised.


htdocs/cron/class/cronjob.class.php
```diff
@@ -1223,6 +1223,45 @@ public function run_jobs($userlogin)
 
 		// Run a method
 		if ($this->jobtype == 'method') {
+			// Deny to launch a method from a deactivated module
```
eldy, line 1226: On each job, there is a property "module_name" loaded with the fetch()
Knowing if the module is on can be done more easily with
`if (!empty($this->module_name) && isModEnabled($this->module_name)) {`

libremaster:
I will try but I see that isModEnabled return false and my module is enabled. I will check where is the problem.

PR #27538, opened 2024-01-15
FIX #27536 VAT of the buyer's country while SERVICE_ARE_ECOMMERCE_200238EC option is activated

### LINKED ISSUE #27536: VAT of the buyer's country not applied if it is an individual while the SERVICE_ARE_ECOMMERCE_200238EC option is activated
### Bug

VAT of the buyer's country not applied if it is an individual while the SERVICE_ARE_ECOMMERCE_200238EC option is activated
as it describe in https://wiki.dolibarr.org/index.php?title=Loi_europ%C3%A9enne_TVA_Vente_%C3%A0_Distance_VAD_et_e-commerce_BtoC

### Environment Version

18.0 19.0

### Environment OS

Linux Debian

### Environment Web server

Linux Debian

### Environment PHP

8.1

### Environment Database

5.7

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

comaiteseb:
VAT of the buyer's country not applied if it is an individual while the SERVICE_ARE_ECOMMERCE_200238EC option is activated
Just a logic issue line 6514 and reduce VAT list in document to the buyer's country VAT


htdocs/core/class/html.form.class.php
```diff
@@ -6460,7 +6460,7 @@ public function load_cache_vatrates($country_code)
 	 *                         Si vendeur non assujeti a TVA, TVA par defaut=0. Fin de regle.
 	 *                         Si le (pays vendeur = pays acheteur) alors la TVA par defaut=TVA du produit vendu. Fin de regle.
 	 *                         Si (vendeur et acheteur dans Communaute europeenne) et bien vendu = moyen de transports neuf (auto, bateau, avion), TVA par defaut=0 (La TVA doit etre paye par l'acheteur au centre d'impots de son pays et non au vendeur). Fin de regle.
-	 *                         Si vendeur et acheteur dans Communauté européenne et acheteur= particulier alors TVA par défaut=TVA du produit vendu. Fin de règle.
+	 *                         Si vendeur et acheteur dans Communauté européenne et acheteur= particulier alors TVA par défaut=TVA du pays acheteur. Fin de règle.
```
eldy, line 6463: I think the term "du produit vendu" was the good one so if we sell wine in a restaurant it may be vat 10% and if meet, it may be 20%).  We use the vat of selling country only if product is unknow.

htdocs/core/class/html.form.class.php
```diff
@@ -6511,19 +6511,19 @@ public function load_tva($htmlname = 'tauxtva', $selectedrate = '', $societe_ven
 		}
 		if (getDolGlobalString('SERVICE_ARE_ECOMMERCE_200238EC')) {    // If option to have vat for end customer for services is on
 			require_once DOL_DOCUMENT_ROOT . '/core/lib/company.lib.php';
-			if (!isInEEC($societe_vendeuse) && (!is_object($societe_acheteuse) || (isInEEC($societe_acheteuse) && !$societe_acheteuse->isACompany()))) {
+			if (isInEEC($societe_vendeuse) && isInEEC($societe_acheteuse) && !$societe_acheteuse->isACompany()) {
 				// We also add the buyer country code
 				if (is_numeric($type)) {
 					if ($type == 1) { // We know product is a service
-						$code_country .= ",'" . $societe_acheteuse->country_code . "'";
+						$code_country = "'" . $societe_acheteuse->country_code . "'";
```
eldy, line 6518: When SERVICE_ARE_ECOMMERCE_200238EC is set, we want to have ALSO in the combo list the vat rate of the purchaser country because there is a lot of exception and restricting the list to purchaser country only will block user if he need to fallback to its own vat rate. This is why we did a concat .=

htdocs/core/class/html.form.class.php
```diff
@@ -6511,19 +6511,19 @@ public function load_tva($htmlname = 'tauxtva', $selectedrate = '', $societe_ven
 		}
 		if (getDolGlobalString('SERVICE_ARE_ECOMMERCE_200238EC')) {    // If option to have vat for end customer for services is on
 			require_once DOL_DOCUMENT_ROOT . '/core/lib/company.lib.php';
-			if (!isInEEC($societe_vendeuse) && (!is_object($societe_acheteuse) || (isInEEC($societe_acheteuse) && !$societe_acheteuse->isACompany()))) {
+			if (isInEEC($societe_vendeuse) && isInEEC($societe_acheteuse) && !$societe_acheteuse->isACompany()) {
```
eldy, line 6514: This seems the fix and looks OK.

htdocs/core/class/html.form.class.php
comaiteseb: Ok, I think it's the right way. But where should I document this new value for SERVICE_ARE_ECOMMERCE_200238EC ?

htdocs/core/class/html.form.class.php
comaiteseb: I understand what you mean but, there is a problèm with the main reason why this rule was introduced.
If you sell meat to an EEC country, you have to apply the VAT of this country. 
This calls into question the structure of VAT rates because there is a lack of a notion of “VAT class” in order to be able to transpose VAT between different countries. ex: VAT 20% (FR) <=> VAT 19%(DE)

PR #27540, opened 2024-01-15
NEW: Dev documentation about setting up pre-commit locally

mdeweerd:
# NEW: Dev documentation about setting up pre-commit locally

A quick start guide about setting up pre-commit.

### DIFF
--- a/dev/setup/README.md
+++ b/dev/setup/README.md
@@ -1,0 +1,10 @@
+# Setting up for Dolibarr development
+
+Check out the
+[online documentation](https://wiki.dolibarr.org/index.php?title=Environment_and_development_tools)
+for recommendations (to be updated).
+
+If anything, install `pre-commit` - it will help run the tools to format and
+make some basic verifications on your submission before committing and pushing
+to github for a PR. See [a quick start guide](./pre-commit/README.md) in this
+Dolibarr repository.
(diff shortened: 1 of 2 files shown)


eldy:
May be rhe README file should be in a dev/setup/precommit dir ?

mdeweerd:
@eldy 
I noticed afterwards that there is a policy to add documentation in separate subdirectories, but I found that not to be very practicle: a developer needs to look into every subdirectory and every README to determine which tools to setup and which are useful. 

Other than the README there is no other file to add to the subdirectory and IMHO, pre-commit is something that should be recommended to the developer.
I think the README could be extended with a summary of what the other directories are for.

I do not determine the policy, but I do recommend to keep this README at least in the root directory of the setup directory.
OR, put a small README in the setup directory and the more detailed documentation in the subdirectory.

Tell me what you want and I'll comply to it.

eldy:
I prefer to have the README into the directory related to what it is speaking about. So into the subdirectory precommit, because we have tons of use case and tons of tools. The dev must choose according to its time.
To solve the pb you mention: "no directive for dev" (and i agree we should be stricter on this point), we can add a readme also into the dev/setup directory, but it would be better to point into this readme to the wiki developer https://wiki.dolibarr.org/index.php?title=Environment_and_development_tools
and then complete this wiki page with instructions recommended for each developers...

mdeweerd:
Done:
- README.md moved to subdirectory;
- Small README.md at the root for setup referencing the online URL.

CI reported spelling errors are in files not related to this PR.

PR #27544, opened 2024-01-15
fix phpstan

thibdrev:
Function dol_get_first_day() should return int but returns string. 
Function getWeekNumber() should return int but returns string.

### DIFF
--- a/htdocs/core/lib/date.lib.php
+++ b/htdocs/core/lib/date.lib.php
@@ -579,7 +579,7 @@ function dol_get_first_day($year, $month = 1, $gm = false)
 	if ($year > 9999) {
 		return '';
 	}
-	return dol_mktime(0, 0, 0, $month, 1, $year, $gm);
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/date.lib.php
```diff
@@ -579,7 +579,7 @@ function dol_get_first_day($year, $month = 1, $gm = false)
 	if ($year > 9999) {
 		return '';
 	}
-	return dol_mktime(0, 0, 0, $month, 1, $year, $gm);
+	return (int) dol_mktime(0, 0, 0, $month, 1, $year, $gm);
```
eldy, line 582: dol_mktime can return
int|string					Date as a timestamp, '' if error

We can modify this to return
int|null					Date as a timestamp, null if error

but the non int case must remain propagated.

thibdrev: Right, I missed that sorry.
I've changed it to `@return int|string` since we have a `return '';` at line 580 which I missed as well.

PR #27546, opened 2024-01-15

frederic34:
@mdeweerd can you check ?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,7 +49,7 @@ repos:
     hooks:
       - id: php-cbf
         files: \.(php)$
-        args: [-p]
+        args: [--standard=dev/setup/codesniffer/ruleset.xml -p]
       - id: php-cs
       - id: php-lint
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,7 +49,7 @@ repos:
     hooks:
       - id: php-cbf
         files: \.(php)$
-        args: [--standard=dev/setup/codesniffer/ruleset.xml -p]
+        args: [--standard=dev/setup/codesniffer/ruleset.xml, -p]
       - id: php-cs
       - id: php-lint
 


mdeweerd:
Technically this will work.

However, when `.phpcs.xml.dist` is in the root directory this is not needed - and it is preferable because the 'phpcbf' can be run directly from the CLI without needing any configuration.

I have the following php.xml in my local setup - it can be the `.php.xml.dist`:
```xml
<?xml version="1.0"?>
<ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="PHP_CodeSniffer" xsi:noNamespaceSchemaLocation="phpcs.xsd">
    <!--
      Documentation:
      https://github.com/squizlabs/PHP_CodeSniffer/wiki/Annotated-Ruleset
      -->

    <rule ref="dev/setup/codesniffer/ruleset.xml"/>

    <arg value="p"/>
    <arg name="colors"/>
    <arg name="extensions" value="php,inc" />
</ruleset>
```

So this still references the file in the common directory, but I can run phpcs and phpcbf from the CLI.
So just `phpcs FILE` or `phpcbf FILE` anywhere in the tree.

Similarly, I believe the codespell configuration was moved into the pre-commit configuration file, but this can break running codespell from the CLI if this eventually leads to removing pyproject.toml.
Currently codespell is the only tool in project.toml and one can choose to use .codespellrc as a configuration file as well.


I started using pre-commit a lot like this (having the configuration in the .pre-commit-config.yaml), but it is better to keep that as light as possible and move the confirmation into the designated configuration files for the tools, and preferably in files such as 'pyproject.toml' (the 'new' standard for python programs), 'setup.cfg' (the previous 'standard' for python).

I've seen several project move from a configuration in `.pre-commit-config.yaml` to the configuration files.

That avoids having to update multiple files when updating a configuration.

Finally, the phpstan configuration files should also be phpstan.neon.dist and 


`.php-cs-fixer.dist.php` is also a good name for a configuration file for php-cs-fixer .


So this will technically work, but I am not a fan of it.

mdeweerd:
Sorry, I missed the '-p' at the end, you need to add a ',' between the arguments.

.pre-commit-config.yaml
```diff
@@ -49,7 +49,7 @@ repos:
     hooks:
       - id: php-cbf
         files: \.(php)$
-        args: [-p]
+        args: [--standard=dev/setup/codesniffer/ruleset.xml -p]
```
mdeweerd, line 52:
```suggestion
        args: [--standard=dev/setup/codesniffer/ruleset.xml,-p]
```

PR #27552, opened 2024-01-15

frederic34:
tried locally : 

image


mdeweerd:
Do not forget to add the ',' between the arguments!

frederic34:
comma is needed if we escape with " like in the doc when having multiple standard
```
- repo: https://github.com/bolovsky/pre-commit-php.git
  rev: 1.4.0
  hooks:
  - id: php-cs
    files: \.(php)$
    args: ["--standard=PSR1/,path/to/ruleset.xml", "-p"]
```

PR #27563, opened 2024-01-15
Fix: PHP8.1 strftime deprecation -> dol_print_date

mdeweerd:
# Fix: PHP8.1 strftime deprecation -> dol_print_date

Change strftime to dol_print_date to work around deprecation


htdocs/salaries/card.php
```diff
@@ -496,7 +496,7 @@
 // Create
 if ($action == 'create' && $permissiontoadd) {
 	$year_current = dol_print_date(dol_now('gmt'), "%Y", 'gmt');
```
mdeweerd, line 498: I see that there is a mixture here: 
- $year_current was already using the 'gmt' option, but,
- $pastmonth was using strftime.

Is the correct approach to add $gmt here or not?

PR #27565, opened 2024-01-16
New getBankAccountPaymentMethod hook for paymentok.php

Hystepik:
Add a new hook in htdocs/public/payment/paymentok.php to get a bank account id for a custom module


eldy:
No need for such changes iinto paymentsKO ?

PR #27567, opened 2024-01-16
NEW can display view "conversation" for events of a user

lamrani002:
# NEW|New display for user events by create function show_actions_messaging_user()


htdocs/core/lib/usergroups.lib.php
```diff
@@ -1722,3 +1722,603 @@ function show_actions_done_user($conf, $langs, $db, $filterobj, $objcon = '', $n
...
+ * 		@param	Contact		       $objcon		   Filter on object contact to filter events on a contact
+ *      @param  int			       $noprint        Return string but does not output it
+ *      @param  string		       $actioncode     Filter on actioncode
+ *      @param  string             $donetodo       Filter on event 'done' or 'todo' or ''=nofilter (all).
+ *      @param  array              $filters        Filter on other fields
+ *      @param  string             $sortfield      Sort field
+ *      @param  string             $sortorder      Sort order
+ *      @return	string|void				           Return html part or void if noprint is 1
+ */
+function show_actions_messaging_user($conf, $langs, $db, $filterobj, $objcon = '', $noprint = 0, $actioncode = '', $donetodo = 'done', $filters = array(), $sortfield = 'a.datep,a.id', $sortorder = 'DESC')
```
eldy, line 1743: Instead of adding a new function show_actions_messaging_user(), can you instead extends the existing show_actions_messaging(). 
When more than 90% of the code will be the same, it is better to add a variant of the function (for example by adding parameters if necessary) instead of duplicating the existing function.

htdocs/user/messaging.php
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	print_barre_liste($titlelist, 0, $_SERVER["PHP_SELF"], '', $sortfield, $sortorder, '', 0, -1, '', 0, $morehtmlright, '', 0, 1, 0);
+
+	// List of all actions
+	$filters = array();
+	$filters['search_agenda_label'] = $search_agenda_label;
+	$filters['search_rowid'] = $search_rowid;
+
+	show_actions_messaging_user($conf, $langs, $db, $object, null, 0, $actioncode, '', $filters, $sortfield, $sortorder);
```
eldy, line 225: Try to reuse the existing  `show_actions_messaging()`

PR #27568, opened 2024-01-16
FIX #27536 VAT of the buyer's country not applied while SERVICE_ARE_ECOMMERCE_200238EC option is activated

### LINKED ISSUE #27536: VAT of the buyer's country not applied if it is an individual while the SERVICE_ARE_ECOMMERCE_200238EC option is activated

comaiteseb:
VAT of the buyer's country not applied if it is an individual while the SERVICE_ARE_ECOMMERCE_200238EC option is activated
Just a logic issue line 6514 and reduce VAT list in document to the buyer's country VAT
As @eldy tell in the previous PR 27538 (canceled cause there is to many conflicts), I introduce a new value for SERVICE_ARE_ECOMMERCE_200238EC parameters for adapt the vat combo with different vat rates.
If SERVICE_ARE_ECOMMERCE_200238EC=1 => the combo is fill with vat of buyer and seller countries
if SERVICE_ARE_ECOMMERCE_200238EC=2 => the combo is only fill with vat of the buyer country.

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -6509,14 +6509,31 @@ class Form
 		}
 		if (getDolGlobalString('SERVICE_ARE_ECOMMERCE_200238EC')) {    // If option to have vat for end customer for services is on
 			require_once DOL_DOCUMENT_ROOT . '/core/lib/company.lib.php';
+   			// If SERVICE_ARE_ECOMMERCE_200238EC=1 combo list vat rate of purchaser and seller countries
+			// If SERVICE_ARE_ECOMMERCE_200238EC=2 combo list only the vat rate of the purchaser country 
+			$selectVatComboMode = getDolGlobalString('SERVICE_ARE_ECOMMERCE_200238EC');
 			if (!isInEEC($societe_vendeuse) && (!is_object($societe_acheteuse) || (isInEEC($societe_acheteuse) && !$societe_acheteuse->isACompany()))) {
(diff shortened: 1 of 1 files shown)


priojk:
Thanks for this fix, which I just found while debugging a v18 setup regarding wrong VAT rates. However, I still have a concern regarding the applicable tax rate:

As far as I know, it is not the country of the buyer's primary address/invoice address that is relevant, but the country of the shipping address (if one is specified). An example is a private person as a customer with a main address in Germany, who buys goods from a German seller, but has them delivered directly to his holiday home in Spain. In this case, the Spanish VAT rate must be applied and not the German VAT rate. Although this may seem a bit exotic, I think we should add the additional check for the existence of a shipping address and its respective destination country. What do you think?

However, an additional problem may occur, if the line items are entered before the shipping address is added.... In this case, the VAT rates are just wrong, but this could be corrected manually by a knowledgeable user.

PR #27572, opened 2024-01-16
NEW  input in emailForm for put instructions

lamrani002:
# NEW|New line with input field by creating function to display it getHtmlForInstruction()


htdocs/core/class/html.formmail.class.php
```diff
@@ -755,7 +755,7 @@ public function get_form($addfileaction = 'addfile', $removefileaction = 'remove
 			if (!empty($this->withtoccuser) && is_array($this->withtoccuser) && getDolGlobalString('MAIN_MAIL_ENABLED_USER_DEST_SELECT')) {
 				$out .= '<tr><td>';
 				$out .= $langs->trans("MailToCCUsers");
-				$out .= '</td><td>';
+				$out .= '</td><>';
```
Hystepik, line 758: This seems to be an error because a empty html tag like this isn't normal and we close a td on line 770 with no new td open

PR #27576, opened 2024-01-16
clean code

frederic34:


htdocs/recruitment/admin/setup_candidatures.php
```diff
@@ -22,35 +22,9 @@
...
-if (!$res && file_exists("../../../main.inc.php")) {
-	$res = @include "../../../main.inc.php";
-}
-if (!$res) {
-	die("Include of main fails");
-}
+require '../../main.inc.php';
 
-global $langs, $user;
+global $conf, $langs, $user;
```
hregis, line 27: @frederic34 call "global" is not mandatory in card !

PR #27578, opened 2024-01-16
Draft: API: Link existing contact to thirdparty

romain-neil:
Hello,

This PR is aim to track the progress against the functionnality of link existing contact to multiple thirdparties.

Theoretical procedure :
- Create a pivot table which record which contact is linked with which thirdparty
- Add corresponding endpoint in api class
- (Should be a part of another PR) Add the "graphical" capacity to link an existing contact

### DIFF
--- a/htdocs/install/mysql/migration/19.0.0-20.0.0.sql
+++ b/htdocs/install/mysql/migration/19.0.0-20.0.0.sql
@@ -203,3 +203,10 @@ ALTER TABLE llx_c_tva ADD COLUMN type_vat smallint NOT NULL DEFAULT 0 AFTER fk_p
 
 ALTER TABLE llx_product DROP COLUMN onportal; 
 
+-- Draft : Create pivot table contact 0...N -> thirdparty
+CREATE TABLE llx_contact_thirdparty
+(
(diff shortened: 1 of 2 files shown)


htdocs/install/mysql/migration/19.0.0-20.0.0.sql
```diff
@@ -203,3 +203,10 @@ ALTER TABLE llx_c_tva ADD COLUMN type_vat smallint NOT NULL DEFAULT 0 AFTER fk_p
 
 ALTER TABLE llx_product DROP COLUMN onportal; 
 
+-- Draft : Create pivot table contact 0...N -> thirdparty
```
eldy, line 206: We already add a n-n link between contact and thirdparties.
This generates more problem than enhancement. End users were lost on how to use permissions for contacts that are users, conflict in when we show the contact of a thirdparty or not, problem when we want to change the phone of a contact in a company that is different into another company, import and export of contact that become no more possible due to storing in 2 different tables, .... Finaly we reduced a lot of trouble by keeping only a relation 1-n.
Providing such a feature brings so many troubles of any kindsthat it won't be included anymore into the core. This should be implemented by external module if you need it.

PR #27582, opened 2024-01-17
QUAL Send API response data before completing post-processing tasks

msoula:
This commit allows Rester to immediately send response data back to the client before completing some time consuming post-processing tasks (such as sending emails) in a termination callback if defined.

Termination callback should be created into the same class as the API method it terminates. Callback name must match the following format : `"_terminate_" + <API_METHOD_NAME> + "_" + <API_METHOD_RESPONSE_FORMAT_EXTENSION>`.

For instance, termination callback for orders _post_ method defined in `htdocs/commande/class/api_orders.class.php` must be named __terminate_post_json_.

### DIFF
--- a/htdocs/api/index.php
+++ b/htdocs/api/index.php
@@ -411,19 +411,19 @@ Luracast\Restler\Defaults::$returnResponse = $usecompression;
 
 // Call API (we suppose we found it).
 // The handle will use the file api/temp/routes.php to get data to run the API. If the file exists and the entry for API is not found, it will return 404.
-$result = $api->r->handle();
+$responsedata = $api->r->handle();
 
 if (Luracast\Restler\Defaults::$returnResponse) {
 	// We try to compress the data received data
 	if (strpos($_SERVER['HTTP_ACCEPT_ENCODING'], 'br') !== false && function_exists('brotli_compress') && defined('BROTLI_TEXT')) {
 		header('Content-Encoding: br');
-		$result = brotli_compress($result, 11, constant('BROTLI_TEXT'));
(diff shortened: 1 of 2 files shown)


htdocs/includes/restler/framework/Luracast/Restler/Restler.php
```diff
@@ -1686,4 +1686,27 @@ protected function postCall()
             ), $this->responseData);
         }
     }
+
```
eldy, line 1689: Restler is an external library, so not developped by the Dolibarr team.
So to add new features in it, you should first try to post the PR to the upstream project...

htdocs/api/index.php
```diff
@@ -434,4 +434,17 @@
 	echo $result;
 }
 
+// Now flush output buffers so that response data is sent to request client
+ob_end_flush();
+ob_flush();
```
eldy, line 439: If we have a ob_end_flush before, this ob_flush should be useless, shouldn't it ?

htdocs/api/index.php
```diff
@@ -434,4 +434,17 @@
...
+flush();
```
eldy, line 440: We did a ob_end_flush, so it means we have returned the output to the client and we now have time to do all the rest. So why doing a flush here. It forces to make IO when they can be delayed and managed by the system when he can.

msoula: Agreed

htdocs/api/index.php
msoula: Agreed

msoula:
I have made changes to the pull request in response to your comments. Thank you for your feedback.

PR #27586, opened 2024-01-17
Fix : in suppier order dispatch, if qty cleaned, bad error + clean code

vmaury:
# FIX in supplier order dispatch, if qty cleaned, bad error 

In supplier Order Dispatch, if you click on the rubber to clear qty and then you submit, you get a bad error :

> Incorrect double value: '' for column `erp-rec`.`llx_commande_fournisseur_dispatch`.`qty` at row 1

cf screeshot

Looking at the code, the qty are bad tested ... 

and It appears that all the GETPOST wasn't cleanly used, I tried to correct them ...


eldy:
I pushed a simpler fix at https://github.com/Dolibarr/dolibarr/commit/b7212cbac40a16e867436dec4353b31955ee8bd3
Does it fix the trouble on your side ?

vmaury:
Perfect if you prefer !
Have a nice day

PR #27587, opened 2024-01-17
17.0 fix replenish virtual stock by warehouse

atm-lena:
Replenish : stock virtual by warehouse no defined

Calcul : 

**if warhouse filter != default warehouse of product :**
real stock by warehouse + qty to product in mo with selected warehouse 

**if warhouse filter = default warehouse of product :**
real stock by warehouse + qty to product in mo with selected warehouse + qty in supplier order - (qty in order + qty to consume in mo)

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -3350,9 +3350,10 @@ class Product extends CommonObject
 	 * @param  string 	$filtrestatut    Id statut pour filtrer sur un statut
 	 * @param  int    	$forVirtualStock Ignore rights filter for virtual stock calculation.
 	 * @param	int		$dateofvirtualstock	Date of virtual stock
+	 * @param	int		$warehouseid	Filter by a warehouse
 	 * @return integer                 Array of stats in $this->stats_mrptoproduce (nb=nb of order, qty=qty ordered), <0 if ko or >0 if ok
 	 */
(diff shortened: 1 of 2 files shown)


eldy:
Yes virtual stock for a warehouse is not defined because we can't calculate a virtual stock per warehouse:
For example we don't know in which warehouse a purchase order will be received, from which warehouse we will take product to make a shipment, even for manufacturing the warehouse is often not know and decided once manufacturing is done.

You suggested some rules to try to guess a value for virtual stock for MO,but what about sale order, purchase order, etc.... I am afraid such a feature can generate confusion for users. So i suggest to add an option 'STOCK_ALLOW_VIRTUAL_STOCK_PER_WAREHOUSE' to enable this behaviour

Also because this is to implement a feature not yet available, it must be pushed into develop branch.

atm-lena:
> Yes virtual stock for a warehouse is not defined because we can't calculate a virtual stock per warehouse: For example we don't know in which warehouse a purchase order will be received, from which warehouse we will take product to make a shipment, even for manufacturing the warehouse is often not know and decided once manufacturing is done.
> 
> You suggested some rules to try to guess a value for virtual stock for MO,but what about sale order, purchase order, etc.... I am afraid such a feature can generate confusion for users. So i suggest to add an option 'STOCK_ALLOW_VIRTUAL_STOCK_PER_WAREHOUSE' to enable this behaviour
> 
> Also because this is to implement a feature not yet available, it must be pushed into develop branch.

I added a hidden option here : https://github.com/Dolibarr/dolibarr/pull/27668

PR #27590, opened 2024-01-17
remove test on WAREHOUSE_ASK_WAREHOUSE_DURING_PROJECT

defrance:
the activation of this value is disabled on stock admin page and if we associate a project on warehouse, we did'nt show the link on project...

### DIFF
--- a/htdocs/projet/element.php
+++ b/htdocs/projet/element.php
@@ -391,7 +391,7 @@ $listofreferent = array(
 		'buttonnew'=>'AddWarehouse',
 		'project_field'=>'fk_project',
 		'testnew'=>$user->hasRight('stock', 'creer'),
-		'test'=>isModEnabled('stock') && $user->hasRight('stock', 'lire') && getDolGlobalString('WAREHOUSE_ASK_WAREHOUSE_DURING_PROJECT')
+		'test'=>isModEnabled('stock') && $user->hasRight('stock', 'lire') // && getDolGlobalString('WAREHOUSE_ASK_WAREHOUSE_DURING_PROJECT')
 	),
 	'propal'=>array(
 		'name'=>"Proposals",


htdocs/projet/element.php
```diff
@@ -391,7 +391,7 @@
 		'buttonnew'=>'AddWarehouse',
 		'project_field'=>'fk_project',
 		'testnew'=>$user->hasRight('stock', 'creer'),
-		'test'=>isModEnabled('stock') && $user->hasRight('stock', 'lire') && getDolGlobalString('WAREHOUSE_ASK_WAREHOUSE_DURING_PROJECT')
+		'test'=>isModEnabled('stock') && $user->hasRight('stock', 'lire') // && getDolGlobalString('WAREHOUSE_ASK_WAREHOUSE_DURING_PROJECT')
```
eldy, line 394: When WAREHOUSE_ASK_WAREHOUSE_DURING_PROJECT is not set, there is no project information linked to a warehouse. So no way to have a report of list of warehouse inside a project. So old test seems good to me. We don't show a page/link if this page is not relevant...

defrance: Except that it is not possible to change this variable, which is only used in two places in dolibarr (the stock admin, the list of elements)
the main problem is that if we do not remove this variable, it is impossible to have the list of warehouses associated with the project...

image

defrance:
image

Except that it is not possible to change this variable, which is only used in two places in dolibarr (the stock admin, the list of elements)
the main problem is that if we do not remove this variable, it is impossible to have the list of warehouses associated with the project...

eldy:
Yes, this is the goal. This option is not yet very used and answers a not very common needs (a project is rarely dependant of a warehouse).
So without option, we don't see the feature (nowhere) to avoid options inflation, if hidden option is set, we show features related to it everywhere.

defrance:
in this case, we need to hide the project selector on warehouse

eldy:
> in this case, we need to hide the project selector on warehouse

Yes, if WAREHOUSE_ASK_WAREHOUSE_DURING_PROJECT, we should not see the project selector in a warehouse

eldy:

PR #27596, opened 2024-01-17
fix: exchange bad translation in simple accounting report

FHenry:
In all translation file 
 - key CalcModeDebt = "Analysis of known recorded documents"
 - key CalcModeEngagement="Analysis of known recorded payments"

It should be inverted in the translation file regarding key name
- "Debt"=Payment
- "Engagement"=Document

In all other use case of these translations key are "correctly inverted", I mean CREANCES-DETTES = CalcModeDebt and RECETTES-DEPENSES = CalcModeEngagement
htdocs/compta/resultat/result.php
htdocs/compta/stats/cabyuser.php
...

I suggest to keep this  key like it is but I'm not totally OK with this solution.


eldy:
Yes CREANCES-DETTES is CalcModeDebt but it also mean in french "Compta d'engagement".
The key CalcModeEngagement is wrong and it means nothing in english. So i replaced it into develop with a new key "CalcModePayment"

PR #27597, opened 2024-01-17
New: manage position for tags ans categories 

Daoud-mohamed:
# Manage position in categories
Add position in edit and show views for tags ans categories 

Capture


hregis:
@Daoud-mohamed the next and reorder position are managed? Reorder in list and not by integer in card! With click and slide in list?

eldy:
> @Daoud-mohamed the next and reorder position are managed? Reorder in list and not by integer in card! With click and slide in list? 

I don't think so. Not yet. Not planned. pr for this is welcome but difficult because tags are multi level...

PR #27602, opened 2024-01-18
stock transfert not stable : return to experimental

defrance:
I detected quite a few bugs, absence of token, error in class definition on certain page... I don't understand how we could declare this module stable overnight without having a minimum of testing it.

### DIFF
--- a/htdocs/core/modules/modStockTransfer.class.php
+++ b/htdocs/core/modules/modStockTransfer.class.php
@@ -63,7 +63,7 @@ class modStockTransfer extends DolibarrModules
 		// Used only if file README.md and README-LL.md not found.
 		$this->descriptionlong = "Advanced management of stock transfer orders with generation of stock transfer sheets";
 		// Possible values for version are: 'development', 'experimental', 'dolibarr', 'dolibarr_deprecated' or a version string like 'x.y.z'
-		$this->version = 'dolibarr';
+		$this->version = 'experimental';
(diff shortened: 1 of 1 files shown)


eldy:
As it was published a long time ago, your issues on module are welcome to make it more stable...

defrance:
long time ago???
image

Showing stable modules when they are not is a very bad advertisement for the overall quality of dolibarr...
It's all well and good to add multiple validation tests to the code if we don't check at least the functioning of the module developed with the generator without the slightest test

eldy:
Sorry i was thinking about the module Reception.
Can you paste at least one issue about the pb you have and ping @atm-thibaultf into its comment. He is the one who did the tests to switch the PR to "stable" status.

defrance:
image

atm-maxime:
Thanks for your feedback @defrance.

The main error (fatal on card view) has been introduced after module was declared stable by a commit from @eldy (https://github.com/Dolibarr/dolibarr/commit/d191220683c8d11478b40beb8d91f5a618eecc84#diff-56b94c8f67a0977e899bc74a512911e834928b3da6eef48507b37f5862454f2fR608). I'll push a fix for this part.

But you're right, the contact tab doesn't work, and the integration of the module is incomplete. @atm-thibaultf I think this should have been integrated as a module, but as a new object in the stock module. @eldy what do you think ?

eldy:
I pushed some fixes to fix the regressions

PR #27604, opened 2024-01-18
Fix phpstan dolresource

W1W1-M:
# Fix dolresource.class.php phpstan lvl3
htdocs/resource/card.php	102	Property Dolresource::$fk_code_type_resource (int) does not accept array|string.
htdocs/resource/class/dolresource.class.php	132	Property Dolresource::$fk_code_type_resource (int) does not accept string.
htdocs/resource/class/dolresource.class.php	286	Property Dolresource::$fk_code_type_resource (int) does not accept string.
htdocs/resource/class/dolresource.class.php	601	Property Dolresource::$resource_id (int) does not accept string.


htdocs/resource/class/dolresource.class.php
```diff
@@ -618,7 +615,7 @@ public function update_element_resource($user = null, $notrigger = 0)
 
 		// Update request
 		$sql = "UPDATE ".MAIN_DB_PREFIX."element_resources SET";
-		$sql .= " resource_id=".(isset($this->resource_id) ? "'".$this->db->escape($this->resource_id)."'" : "null").",";
+		$sql .= " resource_id=".(isset($this->resource_id) ? $this->resource_id : "null").",";
```
eldy, line 618: To avoid sql injection, we must force the cast to (int) during the sql forging when data is int (even if already cast to int previously).

W1W1-M: Thanks

PR #27607, opened 2024-01-18
fix some warnings in apstats

Hystepik:
Fix warnings in doxygen apstats


htdocs/accountancy/bookkeeping/card.php
```diff
@@ -127,17 +127,17 @@
 			$object->subledger_label = $subledger_label;
 			$object->label_compte = $accountingaccount_label;
 			$object->label_operation = $label_operation;
-			$object->debit = $debit;
-			$object->credit = $credit;
+			$object->debit = (float) $debit;
```
eldy, line 130: Cast must be done at the source...

PR #27609, opened 2024-01-18
FIX: Fix supplier invoice security check

creekorful:
The following API routes were not working :

- GET {id}/payments
- POST {id}/payments

Because the invoice object was being fetched after we checked whether the user has access to it or not.

### DIFF
--- a/htdocs/fourn/class/api_supplier_invoices.class.php
+++ b/htdocs/fourn/class/api_supplier_invoices.class.php
@@ -380,13 +380,13 @@ class SupplierInvoices extends DolibarrApi
 			throw new RestException(400, 'Invoice ID is mandatory');
 		}
 
-		if (!DolibarrApi::_checkAccessToResource('fournisseur', $this->invoice->id, 'facture_fourn', 'facture')) {
-			throw new RestException(401, 'Access not allowed for login '.DolibarrApiAccess::$user->login);
-		}
-
 		$result = $this->invoice->fetch($id);
 		if (!$result) {
 			throw new RestException(404, 'Invoice not found');
+		}
+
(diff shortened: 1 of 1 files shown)


htdocs/fourn/class/api_supplier_invoices.class.php
```diff
@@ -380,15 +380,15 @@ public function getPayments($id)
...
-		if (!DolibarrApi::_checkAccessToResource('fournisseur', $this->invoice->id, 'facture_fourn', 'facture')) {
-			throw new RestException(401, 'Access not allowed for login '.DolibarrApiAccess::$user->login);
-		}
-
 		$result = $this->invoice->fetch($id);
 		if (!$result) {
 			throw new RestException(404, 'Invoice not found');
 		}
 
+		if (!DolibarrApi::_checkAccessToResource('fournisseur', $this->invoice->id, 'facture_fourn', 'facture')) {
```
eldy, line 388: This generate a security pb. A user is able to know if an invoice exists or not without having permission on it.
I think a better fix is to replace 
```
if (!DolibarrApi::_checkAccessToResource('fournisseur', $this->invoice->id, 'facture_fourn', 'facture')) {
			throw new RestException(401, 'Access not allowed for login '.DolibarrApiAccess::$user->login);
		}

```with 
```
if (!DolibarrApi::_checkAccessToResource('fournisseur', $id, 'facture_fourn', 'facture')) {
			throw new RestException(401, 'Access not allowed for login '.DolibarrApiAccess::$user->login);
		}
```

We keep to show the error "Access not allowed" before the "Invoice not found"

eldy: I think so. Let's first fix the api to make it working...

creekorful: Thank you @eldy. Is thing going to be ported to 17.0/18.0 and develop as well?

Issue #27220, opened 2023-12-26
Homogenise extrafields fk_object uniqueness handling


creekorful:
Hello there,

Dolibarr extrafields fk_object handling is not consistent :

- There is an unique constraint on ``llx_societe_extrafields.fk_object``
- There is no unique constraint on ``llx_socpeople_extrafields.fk_object``.

In the module builder SQL template the fk_object column is not unique too.

See:

https://github.com/Dolibarr/dolibarr/blob/096a6251dd62a67a393b784f5b0a26079450df27/htdocs/install/mysql/tables/llx_societe_extrafields.key.sql#L20

https://github.com/Dolibarr/dolibarr/blob/096a6251dd62a67a393b784f5b0a26079450df27/htdocs/install/mysql/tables/llx_socpeople_extrafields.key.sql#L20

https://github.com/Dolibarr/dolibarr/blob/096a6251dd62a67a393b784f5b0a26079450df27/htdocs/modulebuilder/template/sql/llx_mymodule_myobject_extrafields.key.sql#L18

---

I did not have crawled through all the files but it already shows index management is no consistent with extrafields.

Do you recon we should do something about it?

Regards

atm-maxime:
Hi @creekorful !
That's definitely something we should improve to have a unique key on fk_object in *_extrafields tables. We should start with the modulebuilder if it's not already done.

creekorful:
Hi @atm-maxime I did open a PR for the ModuleBuilder.

Regarding the others tables I think another PR making sure that the fk_object column is created as an UNIQUE INDEX could be a good idea too.

Sadly this cannot be enforced for existing users by a migration or something else since we cannot predict whether some would have duplicate entries in the extrafields table.

Cheers

atm-maxime:
Great @creekorful !
Maybe we can also deal with cases where duplicates exist. We have to check which line is fetched when there are 2 lines in the extrafields table for the same object (I'm guessing the first). The end user only sees data of this line. And perform a SQL delete for duplicates keeping the lowest rowid.

What do you think?

creekorful:
While this will technically work I am not sure whether this would be a good idea. There could be business critical data on those "hidden" extrafields and deleting them would cause more harm than good IMO.

Maybe we could provide a repair step in `htdocs/install/repair.php` that does what you suggested and let the user execute it manually if wanted?

Either way I feel the only good option here is letting the human choose what to do about it.

Cheers

creekorful:
[...]

PR #27610, opened 2024-01-18
FIX: ModuleBuilder: make sure fk_object is unique on extrafields

creekorful:
See https://github.com/Dolibarr/dolibarr/issues/27220

### DIFF
--- a/htdocs/modulebuilder/template/sql/llx_mymodule_myobject_extrafields.key.sql
+++ b/htdocs/modulebuilder/template/sql/llx_mymodule_myobject_extrafields.key.sql
@@ -15,5 +15,5 @@
 
 
 -- BEGIN MODULEBUILDER INDEXES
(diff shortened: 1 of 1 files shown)


htdocs/modulebuilder/template/sql/llx_mymodule_myobject_extrafields.key.sql
```diff
@@ -15,5 +15,5 @@
 
 
 -- BEGIN MODULEBUILDER INDEXES
-ALTER TABLE llx_mymodule_myobject_extrafields ADD INDEX idx_myobject_fk_object(fk_object);
+ALTER TABLE llx_mymodule_myobject_extrafields ADD UNIQUE INDEX idx_myobject_fk_object(fk_object);
```
eldy, line 18: To match name rule for index, can you also replace the idx_ with uk_

PR #27650, opened 2024-01-19
FIX update price extrafield on propal card

lvessiller-opendsi:
FIX update price extrafield on propal card
- replace "insertExtraFields" with "updateExteField" method on update extrafield (seems to be more coherent)
- fix https://github.com/Dolibarr/dolibarr/issues/24298 No error or 0.00 instead of NULL in database anymore when emptying an extrafield of type price on a propal card
- using price as double type before updating the price extrafield on database

### DIFF
--- a/htdocs/comm/propal/card.php
+++ b/htdocs/comm/propal/card.php
@@ -1635,14 +1635,15 @@ if (empty($reshook)) {
 		$result = $object->setWarehouse(GETPOST('warehouse_id', 'int'));
 	} elseif ($action == 'update_extras') {
 		$object->oldcopy = dol_clone($object, 2);
+		$attribute_name = GETPOST('attribute', 'restricthtml');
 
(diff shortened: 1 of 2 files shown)


lvessiller-opendsi:
@eldy 
The fix was on "insertExtraFields" and not on "updateExtraField".
And according to me, when we update extrafield on card it's more logical to use "updateExtraField()" method instead of "insertExtraFields" method.
What do you think about it ?

eldy:
> @eldy 
> The fix was on "insertExtraFields" and not on "updateExtraField".
> And according to me, when we update extrafield on card it's more logical to use "updateExtraField()" method instead of "insertExtraFields" method.
> What do you think about it ?

Yes, for update we should use updateExtraField(). This method was not available in the past but i think we can now try to use it. Note updateExtraField() is designed to update 1 field.

PR #27654, opened 2024-01-19
qual: phpstan

thibdrev:
htdocs/comm/card.php	221	Property Societe::$tva_assuj (int) does not accept array|string. 
htdocs/comm/card.php	231	Property Societe::$fk_prospectlevel (int) does not accept array|string.

### DIFF
--- a/htdocs/comm/card.php
+++ b/htdocs/comm/card.php
@@ -228,7 +228,7 @@ if (empty($reshook)) {
 	// set prospect level
 	if ($action == 'setprospectlevel' && $user->hasRight('societe', 'creer')) {
 		$object->fetch($id);
-		$object->fk_prospectlevel = GETPOSTINT('prospect_level_id');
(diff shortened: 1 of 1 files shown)


htdocs/comm/card.php
```diff
@@ -228,7 +228,7 @@
 	// set prospect level
 	if ($action == 'setprospectlevel' && $user->hasRight('societe', 'creer')) {
 		$object->fetch($id);
-		$object->fk_prospectlevel = GETPOST('prospect_level_id', 'alpha');
+		$object->fk_prospectlevel = GETPOSTINT('prospect_level_id');
```
eldy, line 231: fk_prospectlevel is a "false friend". It real type is string. I have fixed that in develop.

thibdrev: hum, tricky one.
fixed now.

PR #27659, opened 2024-01-19
qual: phpstan

thibdrev:
htdocs/compta/facture/card.php	531	Property Facture::$revenuestamp (float) does not accept array|string. htdocs/compta/facture/card.php	1006	Property CommonInvoice::$socid (int) does not accept array|string. htdocs/compta/facture/card.php	1705	Property DiscountAbsolute::$fk_soc (int) does not accept array|string.

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -528,7 +528,7 @@ if (empty($reshook)) {
 		}
 	} elseif ($action == 'setrevenuestamp' && $usercancreate) {
 		$object->fetch($id);
-		$object->revenuestamp = (float) GETPOST('revenuestamp','int');
+		$object->revenuestamp = (float) GETPOST('revenuestamp', 'int');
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/card.php
```diff
@@ -528,7 +528,7 @@
 		}
 	} elseif ($action == 'setrevenuestamp' && $usercancreate) {
 		$object->fetch($id);
-		$object->revenuestamp = GETPOST('revenuestamp');
+		$object->revenuestamp = (float) GETPOST('revenuestamp', 'int');
```
eldy, line 531: When a data for an amount or qty is coming from a user, the good way to retreive it is:
`$object->revenuestamp = (float) price2num(GETPOST('revenuestamp'));`

thibdrev: OK noted.
Fixed.

PR #27672, opened 2024-01-19
If the product has no stock limit in the filter warehouse, then it is displayed

atm-lena:
# FIX|Fix #

**replenish.php : filter by warehouse**

If the product has no stock limit in the filter warehouse, then it is displayed. He should not.


eldy:
Why not. If there is no predefined value for the Expected stock for a warehouse, it means the system must not suggest any value to automatically fill the value to order. But end user may want to order the product despite of that. Default value should be empty (because there is no pre-instruction in setup to fill it), but why blocking ordering. My understanding is that if you use inner, it will not be possible to see and fill the product to order it at all ?

atm-lena:
> Why not. If there is no predefined value for the Expected stock for a warehouse, it means the system must not suggest any value to automatically fill the value to order. But end user may want to order the product despite of that. Default value should be empty (because there is no pre-instruction in setup to fill it), but why blocking ordering. My understanding is that if you use inner, it will not be possible to see and fill the product to order it at all ?

You're right. I close this PR.

PR #27674, opened 2024-01-19
fix phpdoc and default value

frederic34:
@mdeweerd your opinion ?

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -5346,13 +5346,13 @@ function info_admin($text, $infoonimgalt = 0, $nodiv = 0, $admin = '1', $morecss
  *  This function must be called when a blocking technical error is encountered.
  *  However, one must try to call it only within php pages, classes must return their error through their property "error".
  *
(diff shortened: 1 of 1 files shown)


mdeweerd:
# DoliDB|null

I prefer nullable.  FYI, you can also write ?DoliDB .
This will probably generate static check errors: fix them using rector or by searching with a regex . 

Impacts 140 files:
```bash
grep -lr -P 'dol_print_error\(\s*(""|'"''"')' htdocs/ | xargs -n 50 perl -p -i -e 's/(dol_print_error\(\s*)(""|'"''"')/$1null/;'
```
There is no result for
```bash
grep -lr -P 'dol_print_error\(.*,\s*\b(array\(\)|null)\s*)' htdocs
grep -lr -P 'dol_print_error\(.*,\s*null\s*)' htdocs
```

Also rather than using 'is_object($db)', it's better to use '$db instanceof DoliDB' .

# errors defaulting to null

I generally prefer defaulting to [] because then there is no type change.  On the other hand null avoids an array_merge operation.

So I agree with both.  Again, you can write `?string[]` .

PR #27678, opened 2024-01-19
qual: phpstan

thibdrev:
htdocs/loan/payment/payment.php	155	Property PaymentLoan::$fk_bank (int) does not accept array|string. 
htdocs/loan/payment/payment.php	157	Property PaymentLoan::$num_payment (int) does not accept array|string.
htdocs/loan/payment/payment.php	185	Property LoanSchedule::$fk_bank (int) does not accept array|string.

### DIFF
--- a/htdocs/loan/class/paymentloan.class.php
+++ b/htdocs/loan/class/paymentloan.class.php
@@ -77,7 +77,7 @@ class PaymentLoan extends CommonObject
 	public $fk_typepayment;
 
 	/**
-	 * @var int Payment ID
+	 * @var string Payment ID (ex: 12345-chq1)
 	 */
 	public $num_payment;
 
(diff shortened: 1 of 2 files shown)


htdocs/loan/payment/payment.php
```diff
@@ -152,9 +152,9 @@
 			$payment->amount_capital	= $pay_amount_capital;
 			$payment->amount_insurance	= $pay_amount_insurance;
 			$payment->amount_interest	= $pay_amount_interest;
-			$payment->fk_bank           = GETPOST('accountid', 'int');
+			$payment->fk_bank           = GETPOSTINT('accountid');
 			$payment->paymenttype       = GETPOST('paymenttype', 'int');
-			$payment->num_payment		= GETPOST('num_payment');
+			$payment->num_payment		= GETPOSTINT('num_payment');
```
eldy, line 157: num_payment is a reference of payment. It can contains chars. For example
12345-chq1
So should remain GETPOST(. , alphanohtml)

htdocs/loan/class/paymentloan.class.php
```diff
@@ -77,7 +77,7 @@ class PaymentLoan extends CommonObject
 	public $fk_typepayment;
 
 	/**
-	 * @var int Payment ID
+	 * @var string Payment ID (ex: 12345-chq1)
```
eldy, line 80: 👍🏻

PR #27691, opened 2024-01-20
qual: phsptan

thibdrev:
htdocs/core/login/functions_openid.php	62	Function check_user_password_openid() should return string but returns false. 
htdocs/core/login/functions_openid.php	64	Function check_user_password_openid() should return string but returns false. 
htdocs/core/login/functions_openid.php	103	Function check_user_password_openid() should return string but returns false. 
htdocs/core/login/functions_openid.php	107	Function check_user_password_openid() should return string but returns false. 
htdocs/core/login/functions_openid.php	112	Function check_user_password_openid() should return string but returns false.


htdocs/core/login/functions_openid.php
thibdrev, line 1: Apparently it should work: https://phpstan.org/writing-php-code/phpdoc-types

PR #27696, opened 2024-01-20
qual: phpstan

thibdrev:
htdocs/multicurrency/multicurrency_rate.php	149	Property CurrencyRate::$rate (float) does not accept string.
htdocs/multicurrency/multicurrency_rate.php	150	Property CurrencyRate::$rate_indirect (float) does not accept string.
htdocs/multicurrency/multicurrency_rate.php	170	Property CurrencyRate::$rate (float) does not accept string.


htdocs/multicurrency/multicurrency_rate.php
```diff
@@ -57,8 +57,8 @@
 $search_code		= GETPOST('search_code', 'alpha');
 $multicurrency_code = GETPOST('multicurrency_code', 'alpha');
 $dateinput 			= dol_mktime(0, 0, 0, GETPOST('dateinputmonth', 'int'), GETPOST('dateinputday', 'int'), GETPOST('dateinputyear', 'int'));
-$rateinput 			= price2num(GETPOST('rateinput', 'alpha'));
-$rateindirectinput 	= price2num(GETPOST('rateinidirectinput', 'alpha'));
+$rateinput 			= (float) price2num(GETPOST('rateinput', 'int'));
```
eldy, line 60: You must use 'alpha' because an amount or rate can be a string (the decimal separator is a special char).It is goal of price2num to convert it to a real number.

thibdrev: I see... `GETPOST($val, 'int')` returns a value only if is a numeric _correctly formatted_, because it ultimately checks the latter using `is_numeric()`.

However `is_numeric()` returns false for `"1 203.45"`, `"1 203,45"` and `"1203,45"`. It solely returns true for `is_numeric("1203.45")`.

To ease things, how about a function such as: 
```
function GETPOSTFLOAT($paramname, $method = 0) {
    return (float) price2num($GETPOST($paramname, 'alpha', $method, null, null, 0));
}
```
?

eldy: You must think user input, and using any language.
For example '1'234.5' or '1·234,5' is not a valid int but may be a valid user input. It depends on language file and countries. Only price2num is able to convert a string user input, that is not always an int, but a string with special char into a real number, taking into account the thousand and decimal separator.
Doing getpost(int) on '1'234.5' will not return 1234.5, price2num will do int.

PR #27700, opened 2024-01-21
Qual: Improve odt2pdf, add pandoc

mdeweerd:
# Qual: Improve odt2pdf, add pandoc

Fix shellcheck notices, add pandoc method.

### DIFF
--- a/scripts/odt2pdf/odt2pdf.sh
+++ b/scripts/odt2pdf/odt2pdf.sh
@@ -3,6 +3,7 @@
 # @copyright  GPL License 2013 - Florian HEnry - florian.henry@open-concept.pro
 # @copyright  GPL License 2017 - Laurent Destailleur - eldy@users.sourceforge.net
 # @copyright  GPL License 2019 - Camille Lafitte - cam.lafit@azerttyu.net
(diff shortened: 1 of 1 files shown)


scripts/odt2pdf/odt2pdf.sh
```diff
@@ -18,41 +19,61 @@
 # If you got this error: javaldx failed! Warning: failed to read path from javaldx with no return to prompt when running soffice --headless -env:UserInstallation=file:"/tmp" --convert-to pdf --outdir xxx ./yyy.odt,
 # check that directory defined into env:UserInstallation parameters exists and is writeable.
 
-# shellcheck disable=2046,2086,2268
-
-
-if [ "x$1" == "x" ]
+if [ "$1" = "" ] || [ "$2" = "" ]
```
eldy, line 22: What happen if you run the script without parameters ? Don't you have a syntax error ?
The x is here to avoid that...

mdeweerd: No, I do not have any script error.

Here's what shellcheck has to say about it:
https://www.shellcheck.net/wiki/SC2268

PR #27706, opened 2024-01-21
QUAL Use phan to help detect new issues

mdeweerd:
# QUAL Use phan to help detect new issues

- Add phan configuration (config, baseline, stubs);
- Add github action;
- Update apstats.php.

Continuation from PR #26480 that was closed following an error in a rebase action.

With the current settings there are 7 errors and 20 warnings in the Phan report.

I leave the other messages to be fixed by developers taking a deeper part in the functional code.

If there is a message to ignore, let me know in a comment and I'll add the corresponding annotation to the code in the second commit of this PR.

The first commit of this PR is related to the phan configuration, the second part is related to the annotations in the code to ignore some false positives.


eldy:
seems to have some false positive error. For example;
Error: upgrade.php: PhanUndeclaredConstantOfClass: Reference to undeclared class constant \DoliDB::VERSIONMIN

Warning: functions.lib.php: PhanPluginInvalidPregRegexReplacement: Call to \preg_replace was passed an invalid replacement reference '$1' to pattern '/^!?[a-zA-Z0-9_]+\\(/'

mdeweerd:
> Error: upgrade.php: PhanUndeclaredConstantOfClass: Reference to undeclared class constant \DoliDB::VERSIONMIN

I proposed a fix in the past that was integrated, then there was the phpstan bug which is fixed now, so I suggest to reintroduce the fix.
It's not a false positive in the sence that the parent class does not have this constant.

I propose it again in #27765 .

> Warning: functions.lib.php: PhanPluginInvalidPregRegexReplacement: Call to \preg_replace was passed an invalid replacement reference '$1' to pattern '/^!?[a-zA-Z0-9_]+\(/'

How is this a false positive? - there is no backref for group 1 .  A group needs parentheses.

eldy:
No way to have the phan files installed by CI instead of beeing commited into .phan dir ?

mdeweerd:
What is installed in the .phan dir are "stubs" that are not part of the phan installation itself.  There is no official source for it and some were generated by me (one of them is a stub for a Dolibarr module for converting numbers to letters).

To be closer to the project's organisation, I can put the files in a directory in dev/tools/phan and keep only a "wrapper" inside .phan so that the configuration file can be automatically found when running it.
However I prefer to do that when you're about to integrate this in the develop branch.

To integrate it in the develop branch sooner, I can have the remaining notices ignored in the baseline (currently .phan/baseline.txt) so that the ci does not break and will report only future exceptions.
Once integrated, I'ld create another PR with the current baseline so that the current state of principal issues can be observed (there are more detections in phan, but we can call that technical debt).

mdeweerd:
(Also see comment above).

I have reorganised the files and ignore *all* the notices so that ci passes (except for the error that was introduced in 'develop').

I also created #27831 where the baseline was restored and where the notices that were seen here will be reported in the phan ci workflow.

eldy:
136 files added to the official sources to just run 1 external looks looks strange to me.
We can have one 1 config file, 1 doc, but we should no have so many files commited for just one purpose.

mdeweerd:
As explained, all these files correspond to stubs which represent the interfaces to libraries.

For php modules (binary), one could rely on phan extracting the information from run-time installed modules, but then these modules need to be installed every time.

For php libraries, this avoids installing (compose) and analyzing them.  Some are even unavailable for installation in ci.

I probably can reduce the number of files by excluding php libraries that are not used in the current configuration, and I can combine all the stub files into one big stub file - it will be less clear which stub came from where, but it's doable and I will insert comments to indicate start and end.

I do not know what you are comfortable with.

EDIT: I removed the stubs that were not referenced in the configuration.

EDIT2: I updated the baseline further - #27831 has the less forgiving baseline.

.github/workflows/phan.yml
```diff
@@ -0,0 +1,40 @@
+---
```
eldy, line 1: Before adding a new too in CI, we must first have the tool working from command line.

Can you make instad oa PR to add files (README) into dev/tools/phan to explain how to install and run phan from command line ?

mdeweerd: SCRIPT addeed, README.md added.

htdocs/variants/tpl/productattributevalueline_view.tpl.php
```diff
@@ -38,6 +38,9 @@
 	exit;
 }
 
+'@phan-var-force CommonObject $this
```
eldy, line 41: Having this into code looks strange.
Should be into comment ?

mdeweerd: These are implemented as string literals due to the limitation of php-ast being unable to parse inline doc comments.
From the docs:
https://github.com/phan/phan/wiki/Annotating-Your-Source-Code#inline-type-checks-via-string-literals

Mostly needed in view files and some rare cases in code.

mdeweerd:
@eldy I updated apstats.php as requested in one of the comments on one of the PRs for phan (https://github.com/Dolibarr/dolibarr/pull/26480#issuecomment-1889966096 ).

eldy:
Do we need to embed all the php stubs ? They are not part of phpstan product ?

mdeweerd:
The stubs are not part of a phan product (there a plans to include stubs with phan, but it's WIP).
As I explained in https://github.com/Dolibarr/dolibarr/pull/27706#issuecomment-1911708363 - I could join them all together.  Either in one big file, or in a few files according to their nature (some correspond to binary libraries, others correspond to php libraries).
I kept only those that are needed for the project.

Also note that number of stubs is 28 files.

htdocs/variants/tpl/productattributevalueline_view.tpl.php
eldy: However on file supplier_proposal.class.php, we have
// @phan-suppress-next-line PhanPluginSuspiciousParamOrde
So it seems possible to have phan tags in doc.

PR #27712, opened 2024-01-21
Qual: Fix/extend language scripts, integrate in pre-commit

mdeweerd:
# Qual: Fix language scripts, integrate in pre-commit

Updated some scripts to comply with shellscript.
Also updated one script to allow annotating the source code in the PR.
Integrated one in pre-commit the seems useful.

This detects duplicates in language files that need separate fixing.


mdeweerd:
I modified dev/translation/strip_language_file.php to generate messages compatible with logToCs.py.

Running 'fix-alt-languages' in ci: seem a bit of an overhead, but I'll leave it for now so that the issues are identifiable.
After that 'fix-alt-languages' can be added to "SKIP" - and it will run locally only when one of the relevant files changed.

mdeweerd:
I added a script 'dev/translation/sanity_check_trans_missing_unused.sh' to quickly find missing and unused translations.
The set of unused translations is larger than the real set because some keys use an composed key (variable+ fixed string).
Those cases could be added as exceptions (regex or file).

I did not add this script to the ci.

There seem to be about 1000 missing translations!

mdeweerd:
I removed the remaining duplicates in the same lang files.

I am not sure if the changes proposes by the fix-alt-languages script must be applied (there are not duplicate keys in the same file).  (fix-alt-languages is the only hook/script preventing the action from finishing with success now).

And, considering #27726, it does not seem needed to keep these pre-commit hooks in ci, I'ld advise to skip them and run only once a day.

PR #27726, opened 2024-01-22
Qual: Remove exceptions for missing/unused/duplicate translations

mdeweerd:
# Qual: Remove exceptions for missing/unused/duplicate translations

This removes the exceptions for the detection of issues in the
translations.  The goal is to incrementally fix these issues until the
PR can be integrated with a "green" ci.

Updated so that #28610 can be safely integrated while this PR continues to report the exceptions "as is".


mdeweerd:
I include the script in the pre-commit configuration as I see that @eldy made good progress on the duplicate strings.
This also allows testing the annotation of the issues in regression.

I think that in the end, only this step is needed in the continuous integration workflow, the other translations checks could be executed in a daily run for instance (same, yaml, different trigger, extra step for the different trigger).

Maybe there are too many warnings here - extra steps can be taken to ignore "issues" (for instance: configuration file with translation keys to ignore for warnings).

mdeweerd:
@eldy Your feedback?

I added 'dev/translation/ignore_translation_keys.lst' to ignore translations keys so that they are not reported as missing or unused keys (to allow ci to work without error and detect only new cases).

However, I did not do anything for the duplicates.
Most concern expenses and trips that share quite a few keys.

I can add the duplicates to dev/translation/ignore_translation_keys.lst as well so that they are ignored for the ci reports.

Goal: run the script in CI to detect future issues on keys, and fix the existing translation keys later.  Most or all "in-file" duplicates were fixed by you.

Edit: I noticed in a comment that the language file for expenses is a copy of the one from trips, so I excluded the keys for Expenses from the report.

mdeweerd:
I created a separate exclusion file for duplicates currently containing the duplicates found on strip and expenses.

Let me know if the remaining duplicates are to be ignored as well (if yes, I'll add them to a section I'll call "Other").

The keys WebPortalSetupPage and WebPortalCSS seem to be new keys as there were not in the exclusion list - maybe you want to check with the author to have them added in the appropriate sections.

This check is actually turning out faster than I expected which means that a lot of time was spent in identifying the locations for the exceptions, not finding the keys that have exceptions.

eldy:
I don't understand result. For exemple we've got
menu.tpl.php: Missing translation; 'name' => $langs->trans('WebPortalInvoiceListMenu'),

Do you know why we have this ?

mdeweerd:
~~I think the script currently relies on~~

```plaintext
WebPortalInvoiceListMenu = Invoices
```

~~to be:~~

```plaintext
WebPortalInvoiceListMenu=Invoices
```

~~If the first case is accepted, I need to update the script.~~

EDIT:
The script seems fine and this PR needed rebasing for your observation/question.
WebPortalInvoiceListMenu is not in the result after rebasing on the develop branch.  I checked the script, it allows a space before the equal sign.

EDIT2:
To explain the message
`menu.tpl.php: Missing translation; 'name' => $langs->trans('WebPortalInvoiceListMenu'),` 
It shows the line where WebPortalInvoiceListMenu is used while the key WebPortalInvoiceListMenu is not found in the en_US translation files.
(The line is available in the pre-commit output, the cs2pr output omits the line number, but the first 10 messages of each type can be found as annotations in the "Files Changed" tab or in the summary of the workflow action - where one can click on the message to go to the relevant annotation in the "Files Changed" tab).

EDIT3:
I checked why FromServerPointOfView was in the missing list, it's only in the fr_FR translation file (in English).

mdeweerd:
I made an improvement to handle translations that use a parameter.

EDIT:
- The script now also includes transnoentities and transnoentitiesnoconv

EDIT2:
- I also added `formSetup->newItem` + made it easier to add such prefixes.
- I wonder if I should add dolGetGlobalString().

mdeweerd:
Do you find this useful, should it be amended, should I delete it ?

eldy:
> Do you find this useful, should it be amended, should I delete it ?

Yes. PR is usefull. I try to fix step by step the report so i can validate the PR and keep CTI green. But there is still a lot of work to do to clean language files...

mdeweerd:
OK, that's good.
Know that there are files that define the exceptions.
We could add the current cases to those files so that the tool can be integrated with a green CI and detect new cases.
And another PR could have the files without those cases added to the exception lists so that they can be fixed gradually.

EDIT: I prepared that PR: #28610 .

PR #27727, opened 2024-01-22
FIX : llx_prospectlevel doesn't have rowid

samifilali:
# FIX
- llx_prospectlevel doesn't have rowid

$newid = ($obj->newid + 1);
=> This generates a fatal error in php 8 because we are trying to add a string (the code field) and an int

image

### DIFF
--- a/htdocs/install/mysql/tables/llx_c_prospectlevel.sql
+++ b/htdocs/install/mysql/tables/llx_c_prospectlevel.sql
@@ -19,6 +19,7 @@
 
 create table llx_c_prospectlevel
 (
+  rowid           integer AUTO_INCREMENT PRIMARY KEY NOT NULL,
   code            varchar(12) PRIMARY KEY,
   label           varchar(128),
   sortorder       smallint,


eldy:
I have to revert due to critical regression. The table already has a primary key that is code so adding a new one will fails.

When this code is executed on table llx_c_prospectlevel ?

samifilali:
The code is executed when a new line is inserted into this dictionary. As the table does not contain a rowid, we use the code. Except that since php 8, the addition of a string and an int generates a fatal result (in any case it didn't make sense to do that). Either we add a specific condition for this table, or we add a rowid to conform to the standard.

Furthermore, I noticed that almost all dictionaries have an auto-increment rowid/id. What is the point of manually calculating and inserting the id/rowid?

samifilali:
@eldy

eldy:
> The code is executed when a new line is inserted into this dictionary. As the table does not contain a rowid, we use the code. Except that since php 8, the addition of a string and an int generates a fatal result (in any case it didn't make sense to do that). Either we add a specific condition for this table, or we add a rowid to conform to the standard.
> 
> Furthermore, I noticed that almost all dictionaries have an auto-increment rowid/id. What is the point of manually calculating and inserting the id/rowid?


I see no reason to have sql error.
With dictionary llx_c_prospectlevel, we should never enter the portion of code in the screenshot that make the select max.
Because there is no rowid, we don't need to make the select max. I tested on develop and we do not execute this code, so i have no trouble with mysql 8.

This seems tobhave been fixed by commit #7939e9... in v19.

PR #27736, opened 2024-01-22
new: add GETPOSTFLOAT() function

thibdrev:
This function is introduced following phpstan errors which required corrections using long code such as 
```
$val = (float) price2num(GETPOST('name','alpha'));
```

@eldy: I kept this new function short and simple but you might want to change it if needed.

Especially:
- I did not mention `$check='alpha'`because `GETPOST()` uses `'alphanohtml'` by default. https://github.com/Dolibarr/dolibarr/blob/6c6adc5333790b1e3d90a5b195ee14a86877a892/htdocs/core/lib/functions.lib.php#L640
    and
https://github.com/Dolibarr/dolibarr/blob/6c6adc5333790b1e3d90a5b195ee14a86877a892/htdocs/core/lib/functions.lib.php#L623
- I did include the parameter `$rounding` from `price2num()`, since it might be useful
- I didn't include the `$method` parameter from `GETPOST()`
    - So it's a bit different from the existing similar function: https://github.com/Dolibarr/dolibarr/blob/6c6adc5333790b1e3d90a5b195ee14a86877a892/htdocs/core/lib/functions.lib.php#L957

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -946,17 +946,31 @@ function GETPOST($paramname, $check = 'alphanohtml', $method = 0, $filter = null
 }
 
 /**
- *  Return value of a param into GET or POST supervariable.
+ *  Return the value of a $_GET or $_POST supervariable, converted into integer.
  *  Use the property $user->default_values[path]['creatform'] and/or $user->default_values[path]['filters'] and/or $user->default_values[path]['sortorder']
  *  Note: The property $user->default_values is loaded by main.php when loading the user.
  *
- *  @param  string  $paramname   Name of parameter to found
- *  @param	int		$method	     Type of method (0 = get then post, 1 = only get, 2 = only post, 3 = post then get)
- *  @return int                  Value found (int)
+ *  @param  string  $paramname   Name of the $_GET or $_POST parameter
+ *  @param  int     $method      Type of method (0 = $_GET then $_POST, 1 = only $_GET, 2 = only $_POST, 3 = $_POST then $_GET)
+ *  @return int                  Value converted into integer
  */
 function GETPOSTINT($paramname, $method = 0)
 {
 	return (int) GETPOST($paramname, 'int', $method, null, null, 0);
+}
+
+
+/**
+ *  Return the value of a $_GET or $_POST supervariable, converted into float.
+ *
+ *  @param  string          $paramname      Name of the $_GET or $_POST parameter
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -946,20 +946,34 @@ function GETPOST($paramname, $check = 'alphanohtml', $method = 0, $filter = null
...
 
+/**
+ *  Return the value of a $_GET or $_POST supervariable, converted into float.
+ *
+ *  @param  string          $paramname      Name of the $_GET or $_POST parameter
+ *  @param  string|int      $rounding       Type of rounding ('', 'MU', 'MT, 'MS', 'CU', 'CT', integer) {@see price2num()}
+ *  @return float                           Value converted into float
+ */
+function GETPOSTFLOAT($paramname, $rounding = '')
+{
```
mdeweerd, line 971: IHMO, the method signature should be similar to the other GETPOSTXXXX methods, so the second argument should be $method and the third could be $rounding.

thibdrev: I hesitated. 
Because on one hand the signature is a bit different, but on the other we might use `$rounding` more often than `$method`.
So we would end up writing `$val = GETPOSTFLOAT($paramname, 0, 'MU')` for example?

mdeweerd: I think that `GETPOSTFLOAT($paramname)` is more frequent than anything else, but the frequency of occurence does not matter here.

If we dive deeper, GETPOSTFLOAT is a bad name because it suggests "POST" while it can also be "GET" (as the "GET" here refers to "getter" for a developer, not to $_GET).  In another framework, I have `getParam($name, $default=null)` when the source can be GET or POST  (in that order).  and `getPost()` or `getQuery()` with the same arguments then specifically targetting the $_POST or $_GET values.

What will happen more often (which is kind of happening now), is that GETPOST is replace with GETPOSTINT which could get replaced with GETPOSTFLOAT which could be replaced with GETPOST* depending on the evolutions of the code or mistakes that were made.
So it should just be as simple as replacing the name of the function.

If the function name suggest it is providing the same functionnality but slightly different than another function, then it's signature should be as compatible as possible.  It's less exceptions to handle for a developer and therefore less bugs that are introduced.

In the framework functions I refered to, the choice was made to differentiate GET/POST/GETorPOST through different function names, not by adding a parameter.  The fact that it is a parameter, now implies (for me) that this second parameter needs to be defined on all similar functions;

PR #27750, opened 2024-01-23
NEW filter on sale representative in contact list

lvessiller-opendsi:
NEW filter on sale representative in contact list
- remove "LEFT JOIN" on commercial user table
- add a filter to select a sale representative (third-party of contact)


eldy:
Thanks for the removal of the left join ;-)

PR #27751, opened 2024-01-23
FIX: use the currency for propal signature page

Slordef:
# FIX Signature of propal now use the propal currency

From a client point of view, it is more pleasant to have the currency of the proposal which automatically comes from the client's currency during creation.

Of course, even if we do not activate "multicurrency" in the modules, the `multicurrency_code` and `multicurrency_total_ttc` fields are correctly used

### DIFF
--- a/htdocs/public/onlinesign/newonlinesign.php
+++ b/htdocs/public/onlinesign/newonlinesign.php
@@ -350,7 +350,7 @@ if ($source == 'proposal') {
 	// Amount
 	$amount = '<tr class="CTableRow2"><td class="CTableRow2">'.$langs->trans("Amount");
 	$amount .= '</td><td class="CTableRow2">';
-	$amount .= '<b>'.price($object->total_ttc, 0, $langs, 1, -1, -1, $conf->currency).'</b>';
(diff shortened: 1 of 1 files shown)


eldy:
I think we should show both when object is multicurrency object.
The main currency (because this is the value that will be requested at the bank and used into accountancy). And if the proposal is mutlicurrency, we show the amount in parenthesis. For example
Amount:   30 €    ($28)
Can you enhance in this direction ?

Slordef:
@eldy sounds good

Slordef:
@eldy After discuss with @Amael-PE 
We came to the conclusion that it is still preferable to only show the customer's currency since it is only him who will have this document.
As for the accounting data, that's another part.

Should we ?

Amael-PE:
Hey here,
Indeed, our use case is the following: Albert is working in a large French company and can accept/sign up to 5000€.

- I sell Albert a service S, whose price is 3000 CHF. As Albert is in France, I offer 3100 €
- Once th offer is done I want to send the link to sign to Albert... But I need him to see the 3100 €, cause he is French and doesn't care about our Swiss prices.

If we show a CHF price (main currency), I will only confuse Albert. He will go to the management etc and it will only make thnigs longer... Better send an email then

I hope our use case is clear !
An thanks for the amazing work on Dolibarr !

PR #27757, opened 2024-01-23
Revert "FIX : llx_prospectlevel doesn't have rowid"

eldy:
Reverts Dolibarr/dolibarr#27727

Generate critical regression. Table already has a primary key.


samifilali:
@eldy How could I handle the case where we try to increment newid (here the code) in dict.php? Because this generates a fatal error in php 8

$newid = ($obj->newid + 1);
=> This generates a fatal error in php 8 because we are trying to add a string (the code field) and an int

image

PR #27770, opened 2024-01-23
Qual: Fix translation keys.

mdeweerd:
# Qual: Fix translation keys

Modify some translation keys so that they correspond to each other, use an existing key or fix the contents as it's not translated yet.

This PR is anticipated in #27726.

EDIT: made a functional change to an error message report, and some minor changes to use newer methods.


htdocs/contrat/class/contrat.class.php
```diff
@@ -1157,7 +1158,8 @@ public function create($user)
 				return -2;
 			}
 		} else {
-			$this->error = $langs->trans("UnknownError: ".$this->db->error()." -", LOG_DEBUG);
+			$this->error = $langs->trans("UnknownError").": ".$this->db->error();
+			dol_syslog(get_class($this)."::create - 10 - ".$this->error, LOG_ERR);
```
eldy, line 1162: Note that when an error occurs after a sql request, it is already logged automatically with a dol_syslog.

PR #27772, opened 2024-01-24
NEW: separation of expense report from salary accounting code

Slordef:
# NEW Ability to choose to separate the accounting code for expense reports and salaries 

Currently, in accounting, expense reports are directly recorded under the same accounting code as salaries.

However, in our case, we need this separation.

I hope that for this change, I haven't forgotten anything
During my tests, the result is as expected

@Amael-PE if you want add some more explanations


eldy:
For new feature, you must push the PR in develop branch

PR #27778, opened 2024-01-24
FIX: update currency by selected supplier

Slordef:
# FIX In supplier invoice, update currency by the selected one

in the creation of a supplier invoice, when you select the supplier, the page reloads with the information but does not change the currency with it.

This behavior is annoying with multi-currencies.

However, having made the change, the page reload no longer takes the user's selection since it systematically takes that of the provider.

Is this a viable behavior?


eldy:
Both solution are not optimal. After your fix, if we change the currency and submit the form and an error occured (for example a mandatory field not filled), then the user selection is lost. There is a solution to get the best of the 2.
When we change the company, a parameter "changecompany" is also set to 1 into the posted form with Dolibarr 19.0.

So i suggest to keep old code, and reuse the currency of company but only if "changecompany" is 1. This PR can be done only in branch v19 because changecompany is not set in v18.

Slordef:
I close this one while another is open for branch 19

PR #27780, opened 2024-01-24
remove code deprecated

frederic34:


htdocs/adherents/card.php
```diff
@@ -334,7 +334,6 @@
 			}
 
 			// Get status and public property
-			$object->statut = GETPOSTINT("statut");
```
eldy, line 337: I suggest to migrate in 2 steps. It may generate a lot of trouble so doing this slowly this way seems better:
- First we move statut into status, EXCEPT for assignement and we keep
```
->statut = ...
->status = ...
```
for compatibility.
Once we use everywhere ->status, we will be able to remove the double assigment in a next version.

frederic34: @eldy can you check change in commonobject class
I put properties private to access them win `__get()` and `__set()`

PR #27784, opened 2024-01-24
qual: phpstan for htdocs/compta/facture/class/facture.class.php

thibdrev:
htdocs/compta/facture/class/facture.class.php	4468	Property Facture::$remise_percent (int) does not accept string.
htdocs/compta/facture/class/facture.class.php	6299	Method FactureLigne::insert() should return int but returns float.
htdocs/compta/facture/class/facture.class.php	6534	Method FactureLigne::update() should return int but returns float.
htdocs/compta/facture/class/facture.class.php	6780	Method FactureLigne::get_prev_progress() should return int but returns float.

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -4433,7 +4433,7 @@ class Facture extends CommonInvoice
 	{
 		// Clean parameters
 		if (empty($remise)) {
-			$remise = 0.0;
+			$remise = 0;
 		}
 
 		if ($user->hasRight('facture', 'creer')) {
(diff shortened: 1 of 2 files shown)


htdocs/compta/facture/class/facture.class.php
```diff
@@ -4433,18 +4433,18 @@ public function setDiscount($user, $remise, $notrigger = 0)
...
-			$remise = price2num($remise, 2);
+			$remise = (float) price2num($remise, 2);
 
 			$error = 0;
 
 			$this->db->begin();
 
 			$sql = 'UPDATE '.MAIN_DB_PREFIX.'facture';
-			$sql .= ' SET remise_percent = '.((float) $remise);
+			$sql .= ' SET remise_percent = '.$remise;
```
eldy, line 4447: When we forge a sql request, we must always for the escaping or the cast for numeric value.

PR #27799, opened 2024-01-25
FIX avoid from re-initializing result on nested hook getEntity

lvessiller-opendsi:
FIX avoid from re-initializing result on nested hook getEntity
- when you have a first hook such as "formBuilddocOptions" calling "getEntity()" function
- then "getEntity()" function executes the hook "hookGetEntity" (nested hook) and results of the previous hook are re-initialized

**For example**
We have two module : Module 1 and Module 2
1. Module 1 : hook "formBuilddocOptions" is adding some  (using "resPrint" of hook manager object)
2. Module 2 : hook "formBuilddocOptions" is calling getEntity() and then the global "hookmanager" is re-initialized and "resPrint" value is now empty
3. So the options before generating a document are lost (resPrint is empty)

**Before**
image

**After**
image

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -309,15 +309,13 @@ function getEntity($element, $shared = 1, $currentobject = null)
 		'currentobject' => $currentobject,
 		'out' => $out
 	);
-	// avoid from re-initializing results of previous hook (nested hooks)
-	$hookmanager2 = clone $hookmanager;
-	$reshook = $hookmanager2->executeHooks('hookGetEntity', $parameters, $currentobject, $action); // Note that $action and $object may have been modified by some hooks
+	$reshook = $hookmanager->executeHooks('hookGetEntity', $parameters, $currentobject, $action); // Note that $action and $object may have been modified by some hooks
 
 	if (is_numeric($reshook)) {
-		if ($reshook == 0 && !empty($hookmanager2->resPrint)) {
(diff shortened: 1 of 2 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -309,13 +309,15 @@ function getEntity($element, $shared = 1, $currentobject = null)
 		'currentobject' => $currentobject,
 		'out' => $out
 	);
-	$reshook = $hookmanager->executeHooks('hookGetEntity', $parameters, $currentobject, $action); // Note that $action and $object may have been modified by some hooks
+	// avoid from re-initializing results of previous hook (nested hooks)
+	$hookmanager2 = clone $hookmanager;
```
eldy, line 313: The idea of having $hookmanager as a global variable is to have only 1 hook manager instance in memory for all php session.
Here by doing cloning inside getEntity (a very often called method), you may highly increase the memory use.
Can you try a different way to fix:
Inside the executeHooks method, replace the
`$this->resPrint [.]= $actionclassinstance->resprints;`
by
`$localResPrint [.]= $actionclassinstance->resprints;`

and at the end of the method executeHooks:
$this->resPrint = $localResPrint

Does it fix the bug you have ?

lvessiller-opendsi:
@eldy
You're right. It's a better way to resolve this problem of nested hooks.

Could you check why I have this Travis error ?
- Checking PHP syntax errors (only 1 version to not overload travis and avoid duplicate tests)
-  parallel-lint: command not found

PR #27801, opened 2024-01-25
FIX : missing hook initialization (IMPORTANT)

altairis-tof:
getSellPrice hook was not triggered because of missing initialization

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -1814,6 +1814,7 @@ class Product extends CommonObject
 
 		// Call hook if any
 		if (is_object($hookmanager)) {
+			$hookmanager->initHooks(array('productdao'));
 			$parameters = array('thirdparty_seller'=>$thirdparty_seller, 'thirdparty_buyer' => $thirdparty_buyer, 'pqp' => $pqp);
 			// Note that $action and $object may have been modified by some hooks
 			$reshook = $hookmanager->executeHooks('getSellPrice', $parameters, $this, $action);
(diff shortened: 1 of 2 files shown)


htdocs/product/class/product.class.php
```diff
@@ -1814,6 +1814,7 @@ public function getSellPrice($thirdparty_seller, $thirdparty_buyer, $pqp = 0)
 
 		// Call hook if any
 		if (is_object($hookmanager)) {
+			$hookmanager->initHooks(array('productdao'));
```
eldy, line 1817: A hook must not be intialized by classes. Hooks must be initialized in main cases by the main parent page or the main CLI script.
Module using the hook must use the context of the page.

eldy, line 1817: We have some initHook(array('objectdao')) inside some method of class.
But this is a bad practice. A context of use should not be defined into a class. It must be defined by the parent page.
If a module need to have its hook executed in all cases, he can use the context 'all' or 'main' (all web pages).

altairis-tof: ok @eldy i have update my PR and i initialize the hook in takepos/invoice.php (https://github.com/Dolibarr/dolibarr/pull/27801/commits/f69ed514cd95d7b766cb4b1bc5aa4620d50123e5)
(there are other hook init in product.class.php it is why i did like that)

altairis-tof:
forgot it, i will test the takeposinvoice context in my module !
sorry for the inconvenience

PR #27802, opened 2024-01-25
Qual phpstan societe class 

W1W1-M:
# Qual societe.class.php phpstan lvl3
htdocs/societe/class/societe.class.php	316	PHPDoc type int of property Societe::$region_code is not covariant with PHPDoc type string of overridden property CommonObject::$region_code.
htdocs/societe/class/societe.class.php	1346	Property CommonObject::$state_id (int) does not accept string.
htdocs/societe/class/societe.class.php	1369	Property Societe::$tva_assuj (int) does not accept string.
htdocs/societe/class/societe.class.php	1371	Property Societe::$vat_reverse_charge (int) does not accept string.
htdocs/societe/class/societe.class.php	1392	Property Societe::$capital (float) does not accept string.
htdocs/societe/class/societe.class.php	1395	Property Societe::$capital (float) does not accept string.
htdocs/societe/class/societe.class.php	4020	Method Societe::isACompany() should return bool but returns int.
htdocs/societe/class/societe.class.php	4241	Property Societe::$code_fournisseur (string) does not accept int.
htdocs/societe/class/societe.class.php	4308	Property Societe::$region_code (int) does not accept string.
htdocs/societe/class/societe.class.php	4330	Property Societe::$country_id (int) does not accept string.
htdocs/societe/class/societe.class.php	4401	Property Societe::$capital (float) does not accept string.
htdocs/societe/class/societe.class.php	4413	Property Societe::$tva_assuj (int) does not accept string.

For `region_code`: in db it's an `integer` but seems to be used as `string` in various parts of Dolibarr code.


htdocs/societe/class/societe.class.php
```diff
@@ -1388,12 +1388,7 @@ public function update($id, $user = '', $call_trigger = 1, $allowmodcodeclient =
 		$this->localtax1_value = trim($this->localtax1_value);
 		$this->localtax2_value = trim($this->localtax2_value);
 
-		if ($this->capital != '') {
-			$this->capital = price2num(trim($this->capital));
-		}
-		if (!is_numeric($this->capital)) {
-			$this->capital = ''; // '' = undef
-		}
+		$this->capital = (!empty($this->capital)) ? (float) price2num(trim($this->capital)) : 0;
```
eldy, line 1391: We must be able to set nothing (so '' in input = null in databse) OR 0 OR a value.
It seems this break this feature...

htdocs/societe/class/societe.class.php
```diff
@@ -1550,7 +1545,7 @@ public function update($id, $user = '', $call_trigger = 1, $allowmodcodeclient =
 				$sql .= ",localtax2_value =0.000";
 			}
 
-			$sql .= ",capital = ".($this->capital == '' ? "null" : $this->capital);
+			$sql .= ",capital = ".($this->capital == 0 ? "null" : $this->capital);
```
eldy, line 1548: If 0, we must store 0, if null, we must store "null".

htdocs/societe/class/societe.class.php
```diff
@@ -3994,30 +3989,27 @@ public function info($id)
...
-	 *  @return    boolean     true=is a company, false=a and user
+	 *  @return		boolean		if a company: true || if a user: false
 	 */
 	public function isACompany()
 	{
-		global $conf;
-
 		// Define if third party is treated as company (or not) when nature is unknown
-		$isacompany = !getDolGlobalString('MAIN_UNKNOWN_CUSTOMERS_ARE_COMPANIES') ? 0 : 1; // 0 by default
+		$isACompany = !getDolGlobalString('MAIN_UNKNOWN_CUSTOMERS_ARE_COMPANIES') ? false : true;
```
eldy, line 3999: As most option, sometimes it may have 3 values or more.
So we must keep a type int and not boolean.

htdocs/societe/class/societe.class.php
```diff
@@ -4410,7 +4403,7 @@ public function setMysoc(Conf $conf)
 		$this->logo_squarred_mini = getDolGlobalString('MAIN_INFO_SOCIETE_LOGO_SQUARRED_MINI');
 
 		// Define if company use vat or not
-		$this->tva_assuj = getDolGlobalString('FACTURE_TVAOPTION');
+		$this->tva_assuj = (is_numeric(getDolGlobalString('FACTURE_TVAOPTION'))) ? (int) getDolGlobalString('FACTURE_TVAOPTION') : 0;
```
eldy, line 4406: I think this option FACTURE_TVAOPTION is 0 or 1 so you may replace with
`$this->tva_assuj = getDolGlobalInt('FACTURE_TVAOPTION');`

htdocs/societe/class/societe.class.php
W1W1-M: Ok, I will keep this in mind. This method historically returns a boolean so I will still need to transform final result to a bool.

htdocs/societe/class/societe.class.php
W1W1-M: I checked and FACTURE_TVAOPTION is effectively only 0 or 1. Thanks.

htdocs/societe/class/societe.class.php
W1W1-M: Ok, you are right 👍

htdocs/societe/class/societe.class.php
W1W1-M: Yes I agree. I will refactor.

W1W1-M (commented):
Have refactored according to comments.

PR #27808, opened 2024-01-25
FIX: propal use devise changes

Slordef:
# FIX On propal, when use product, ability to redefine price

Currently, when creating a propal, if you use a registered product but choose to redefine the price in multicurrency, it remains unchanged at the time of addition


eldy:
We may have the same problem with sale orders. Can you check and aply the same fix if this is true ?

PR #27817, opened 2024-01-26
FIX Https detection

mdeweerd:
# Fix Https detection

theme/eldy/manifest.json.php was returning http:// links instead of the https link defined in the configuration.
I found that this is because I am behide a proxy and that isHTTPS does the detection as expected.

Therefore that function is used to detect https which results in the correct urls.

Detected on my 18.0.2 installation

### DIFF
--- a/htdocs/main.inc.php
+++ b/htdocs/main.inc.php
@@ -432,7 +432,7 @@ if (GETPOST('textbrowser', 'int') || (!empty($conf->browser->name) && $conf->bro
 
 // Force HTTPS if required ($conf->file->main_force_https is 0/1 or 'https dolibarr root url')
 // $_SERVER["HTTPS"] is 'on' when link is https, otherwise $_SERVER["HTTPS"] is empty or 'off'
-if (!empty($conf->file->main_force_https) && (empty($_SERVER["HTTPS"]) || $_SERVER["HTTPS"] != 'on') && !defined('NOHTTPSREDIRECT')) {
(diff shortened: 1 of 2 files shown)


htdocs/main.inc.php
```diff
@@ -461,15 +461,16 @@ function analyseVarsForSqlAndScriptsInjection(&$var, $type)
 if (!defined('NOLOGIN') && !defined('NOIPCHECK') && !empty($dolibarr_main_restrict_ip)) {
 	$listofip = explode(',', $dolibarr_main_restrict_ip);
 	$found = false;
+	$user_ip = getUserRemoteIP();
 	foreach ($listofip as $ip) {
 		$ip = trim($ip);
-		if ($ip == $_SERVER['REMOTE_ADDR']) {
+		if ($ip == $user_ip) {
```
eldy, line 467: This break the security:
getUserRemoteIP return a non reliable IP for security (try to guess the end user IP but can be spoofed by another one by a hacker just with a simple http header). 
If we restrict acces on some IP, test must no be done on a value forged by the end user or a intermediate server but on the REAL IP that connect to the server. So it must be $_SERVER['REMOTE_ADDR'] (the only value that is the real ip viewed by the server).
If you need to connect using a proxy, you must allow the IP of the proxy. This is the goal of the protection to not allow a proxy if ip is another one.

mdeweerd: Ok for the explication, so I reverted that.
However, with regards to security, when IP verification is done, I've seen it mostly done for the entire proxy chain.  So maybe there is an improvement to do on this check.

PR #27825, opened 2024-01-26
Minor fixes, typo or descriptions

libremaster:
These are small and trivial fixes.


libremaster:
I don't know what is is this new function : getDolGlobalConst but PHPStan failed because this.

PR #27831, opened 2024-01-26
Qual: Restore phan baseline not ignoring *all* notices

mdeweerd:
# Qual: Restore baseline not ignoring *all* notices

This restores the baseline for phan to observe the more important
phan notices so that they can be fixed.

The goal is to integrate #27706 which ignores all existing phan
warnings and use this PR to check the issues that still need fixing.

### DIFF
--- a/dev/tools/phan/baseline.txt
+++ b/dev/tools/phan/baseline.txt
@@ -24,15 +24,9 @@ return [
 
     // Currently, file_suppressions and directory_suppressions are the only supported suppressions
     'file_suppressions' => [
-        'htdocs/adherents/type.php' => ['PhanPluginDuplicateExpressionBinaryOp'],
         'htdocs/admin/receiptprinter.php' => ['PhanRedefineFunctionInternal'],
-        'htdocs/api/class/api_documents.class.php' => ['PhanPluginDuplicateExpressionBinaryOp'],
-        'htdocs/barcode/printsheet.php' => ['PhanPluginDuplicateExpressionBinaryOp'],
-        'htdocs/categories/class/api_categories.class.php' => ['PhanAccessMethodProtected'],
-        'htdocs/categories/viewcat.php' => ['PhanPluginDuplicateExpressionBinaryOp'],
         'htdocs/collab/index.php' => ['PhanParamTooMany'],
         'htdocs/comm/mailing/card.php' => ['PhanPluginSuspiciousParamPosition'],
-        'htdocs/compta/cashcontrol/cashcontrol_card.php' => ['PhanPluginDuplicateExpressionBinaryOp'],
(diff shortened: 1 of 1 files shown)


eldy:
The git dir of dolibarr should not contains external project tools for development but a script or doc, for example a README, to install them. Only setup specific to dolibarr should be invluded into project.

mdeweerd:
Can you explain what you are referring to exactly?

eldy:
> Can you explain what you are referring to exactly?

I was thinking that you commit some files that were part of phan project. But it appears no. So forgot my comment...

eldy:
PR #27706 become ok. So i merged it. Does this PR is still usefull ? Looks a duplicate ?

mdeweerd:
```
$a=2;
$b=3;
print ($a & 1 == 1) ? 1 : 0;
print PHP_EOL;
print ($a & 1 == 0) ? 1 : 0;
print PHP_EOL;
print ($b & 1 == 1) ? 1 : 0;
print PHP_EOL;
print ($b & 1 == 0) ? 1 : 0;
print PHP_EOL;
```

gives

```
0
0
1
0
```

so priority is indeed on the '=='.

mdeweerd:
> PR #27706 become ok. So i merged it. Does hits PR is still usefull ? Looks a duplicate ?

This PR uses another configuration file to show more errors (less exclusions).

So it's usefull to compute the messages based on config_extended.php and once fixed, update config.php and allow more messages by updating config_extended.php according to priority on the messages to fix.

I can also enable all the rules in a PR for config.php but exclude all existing code that results in a detection by phan.  Then only the new code will trigger notifications.  (The above observation remains valid though).

mdeweerd:
+ You can also enable phan for apstats now that the configuration is in the 'develop' branch.

mdeweerd:
I fixed some Phan notices and the ones remaining here are for repeated conditions.

That is surely not intentional and some maybe a change that should be made to the second condition rather than just deleting it.
I am not familiar with the rights or with the post variables usage in dolibarr, so I think it's best that someone with more knowledge about that has a look.

Once these are fixed, this can be merged.  I can open a new PR with an updated list of rules that are checked/not skipped.

eldy:
> + You can also enable phan for apstats now that the configuration is in the 'develop' branch.

This was difficult because our sercer was ubuntub 20.94 and phan fails with dependency problems. It was upgraded to 22.04 this week end. So phan is now available on apstats. A new url has been introduced:  https://cti.dolibarr.org

mdeweerd:
These issue are not complex to solve but personally I am not sure if these are copy/paste issues that still need an update, or if they or simply forgotten duplicate tests that can be deleted.

So I am not fixing these minor issues myself.

mdeweerd:
Method change with config_extended.

PR #27832, opened 2024-01-26
NEW rename const on third-party card

lvessiller-opendsi:
NEW rename const on third-party card
- see #27673


lvessiller-opendsi:
@eldy 
Could you check why "getDolGlobalConst" was introduced in "htdocs/core/class/html.formmail.class.php" at line 892 ?
Is it a new function not implemented yet ?

PR #27846, opened 2024-01-28
New branch 28 01 2024

grandoc:
Fix : Warning: Undefined variable $tmpfieldlist in /home/httpd/vhosts/aflac.fr/domains/dev.aflac.fr/httpdocs/admin/mails_templates.php on line 1099
This should preserve the logic of code for adding the 'nobottom' class while using the new method to generate the ID of the <tr>.

### DIFF
--- a/htdocs/admin/mails_templates.php
+++ b/htdocs/admin/mails_templates.php
@@ -1096,10 +1096,16 @@ if ($num) {
 				}
 				print "</tr>\n";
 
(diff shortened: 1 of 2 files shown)


eldy:
This looks to be quality code change, so this must be pushed on develop.
If it is a bug fix, can you explain how to reproduce the bug and use FIx in or label

PR #27848, opened 2024-01-28
Fix pre commit (add --files)

mdeweerd:
# Fix pre commit (add --files)

The '--files' option was forgotten before fhe filelist.


mdeweerd:
Added in #27847, closing this PR.

PR #27852, opened 2024-01-29
fix phpstan in website.class.php

frederic34:
image


thibdrev:
Hi @frederic34 , 
I just had a look on this one before work.
Shouldn't it return `true|-2  True if the line has been replaced, -2 if error` ?
Also I think we're checking twice if `posix_getpwuid()` exists, but not if `posix_geteuid()` exists: https://github.com/Dolibarr/dolibarr/blob/f61dee7456ea130b2700b559de1666b79c128092/htdocs/website/class/website.class.php#L2080-L2082

PR #27855, opened 2024-01-29
NEW supplier order submit method message

W1W1-M:
# NEW supplier order submit method message
This PR adds the supplier order submission method (email, phone, etc.) to the end of the event message.


htdocs/core/triggers/interface_50_modAgenda_ActionsAuto.class.php
```diff
@@ -853,7 +854,7 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
 
 			if (empty($object->actionmsg2)) {
 				if (empty($object->context['actionmsg2'])) {
-					$object->actionmsg2 = $langs->transnoentities("SupplierOrderSubmitedInDolibarr", ($object->newref ? $object->newref : $object->ref));
+					$object->actionmsg2 = $langs->transnoentities("SupplierOrderSubmitedInDolibarr", ($object->newref ?: $object->ref), ($object->getInputMethod()));
```
eldy, line 857: Because the input method is a custom label for user, inside the sentence it may have no sence. So i suggest to keep the old text Purchase Order %s submitted and complete in a more neutral translation way (with parenthesis for example) like this
$object->actionmsg2 = $langs->transnoentities("SupplierOrderSubmitedInDolibarr", ($object->newref ?: $object->ref))."  (".$object->getInputMethod().")";

W1W1-M: Good idea. I propose to add the parenthesis in the lang file in order to keep the PHP code more readable. In any case the localization will need to be updated for other langs.

W1W1-M (commented):
Code adjusted

PR #27856, opened 2024-01-29
New intervention signed status data structure

W1W1-M:
# New intervention signed status data structure
In order to rework PR #27346 a signed bool is necessary in the intervention table. This PR adds the field as well as signed status action trigger. Feature request #27064


htdocs/install/mysql/data/llx_c_action_trigger.sql
```diff
@@ -109,6 +109,7 @@ insert into llx_c_action_trigger (code,label,description,elementtype,rang) value
 insert into llx_c_action_trigger (code,label,description,elementtype,rang) values ('MEMBER_EXCLUDE','Member excluded','Executed when a member is excluded','member',27);
 insert into llx_c_action_trigger (code,label,description,elementtype,rang) values ('FICHINTER_VALIDATE','Intervention validated','Executed when a intervention is validated','ficheinter',30);
 insert into llx_c_action_trigger (code,label,description,elementtype,rang) values ('FICHINTER_MODIFY','Intervention modify','Executed when a intervention is modify','ficheinter',30);
+insert into llx_c_action_trigger (code,label,description,elementtype,rang) values ('FICHINTER_SIGN','Intervention signed','Executed when a intervention is signed','ficheinter',31);
```
eldy, line 112: Ok to add the "signed" column

However, for trigger, we must try to reduce number of triggers to CRUD methods only.
So signing will be setting a flag on intervention so updating the intervention. The trigger FICHINTER_MODIFY (the U of CRUD) should be used for that. 
We may add fill the $object->context to show we are in a signing process.

htdocs/install/mysql/tables/llx_fichinter.sql
```diff
@@ -39,6 +39,7 @@ create table llx_fichinter
   datee				date,						-- date de fin d'intervention
   datet				date,						-- date de terminaison de l'intervention
   duree				real,                       -- duree totale de l'intervention
+  signed            boolean  DEFAULT NULL,      -- signed status
```
eldy, line 42: Can you use the type int default 0 ?
Rhevtype boolen is a pain when we need to introduce a third possible value...

W1W1-M: I had thought of using the following :

- default **NULL** : unknown/empty sign status for not showing anything for interventions pre feature
- **FALSE** : not signed
- **TRUE** : signed

I will update and propose to use int as following :

- default **-1** or **NULL** : unknown/empty sign status for not showing anything for interventions pre feature
- **0** : not signed
- **1** : signed (by client)

If needed futur values could be :

- **2** : signed by organisation/company
- **3** : signed by both parties

PR #27861, opened 2024-01-29
Close #27777 : new massaction subscription creation

### LINKED ISSUE #27777: mass action to create member subscriptions
### Feature Request

in member list select members to make mass action create new subscription

### Use case

_No response_

### Suggested implementation

_No response_

### Suggested steps

_No response_

Hystepik:
Create a new massaction in htdocs/adherents/list.php to generate subscriptions of multiple members


eldy:
When we create a subscription manually, depending on a lot of other modules, we may have a lot of other information to enter:
image

What about this field if doing this in a mass action ?

Hystepik:
> When we create a subscription manually, depending on a lot of other modules, we may have a lot of other information to enter: image
> 
> What about this field if doing this in a mass action ?

Hystepik:
At the moment the code makes it that the date is dol_now() and the amount is like in newpayment public page for complementary action after save it is set to none

eldy:
Add warning in confirm to inform user about what the action will do...

Hystepik:
image

PR #27864, opened 2024-01-29
Qual clean dolresource

W1W1-M:
# Qual clean dolresource class code
Refactored code & comments to make more readable, efficient & safe


htdocs/resource/class/dolresource.class.php
```diff
@@ -460,24 +490,24 @@ public function fetchElementResource($id)
...
+		global $user, $conf;
 		require_once DOL_DOCUMENT_ROOT.'/core/lib/files.lib.php';
 
 		$error = 0;
 
 		$this->db->begin();
 
 		$sql = "DELETE FROM ".MAIN_DB_PREFIX.$this->table_element;
-		$sql .= " WHERE rowid = ".((int) $rowid);
+		$sql .= " WHERE rowid = ".($rowid);
```
eldy, line 505: We must force the int when forging sql request. Even if we are sure that it is an int, so it will avoid false positive in our tool to dectect sql injection that analyze the sql syntax.

W1W1-M: I think for phpstan it's ok. Thanks for explanation about sql injection. I will adjust code.

PR #27876, opened 2024-01-29
FIX: Show tickets created by current user

ksar-ksar:
List should shows tickets created by or flagged for the current user Currently, only shows flagged for the current user not created by See : https://www.dolibarr.org/forum/t/hortcut-to-check-my-tickets/24877


htdocs/ticket/list.php
```diff
@@ -448,6 +448,7 @@
 
 if (!$user->socid && ($mode == "mine" || (!$user->admin && $conf->global->TICKET_LIMIT_VIEW_ASSIGNED_ONLY))) {
 	$sql .= " AND (t.fk_user_assign = ".((int) $user->id);
+	$sql .= " OR t.fk_user_create = ".((int) $user->id);
```
eldy, line 451: This condition is already present 2 lines after.

ksar-ksar:
Sorry, I should have checked.

PR #27884, opened 2024-01-30
Fix some phpstan warnings

Hystepik:


htdocs/adherents/class/adherent.class.php
```diff
@@ -1689,7 +1689,7 @@ public function subscriptionComplementaryActions($subscriptionid, $option, $acco
 
 		$error = 0;
 
-		$this->invoice = null; // This will contains invoice if an invoice is created
+		$this->invoice = new Facture($this->db); // This will contains invoice if an invoice is created
```
eldy, line 1692: According to comment, we should have here an invoice but when the invoice will be created (later). If not, we should keep null to say "invoice not yet created"

Hystepik: The problem is that setting null to property that is of type object Invoice is a bad way to proceed.
 Maybe this could be a better way ?
```suggestion
		if (isset($this->invoice)){ // This will contains invoice if an invoice is created
		     unset($this->invoice);
		} 
```

eldy: What's wrong ? In PHP, a vriable can have several types. For example, it can be null (null is a type) or Facture.
This can be declared this way
`* @var Facture|null    To store the created invoice into subscriptionComplementaryActions()`

PR #27887, opened 2024-01-30
FIX : Adding hooks init

atm-adrien:
## FIX : Adding hooks init

 Adding hooks init

### DIFF
--- a/htdocs/commande/document.php
+++ b/htdocs/commande/document.php
@@ -82,7 +82,7 @@ if ($user->socid) {
 }
 
 // Initialize technical object to manage hooks of page. Note that conf->hooks_modules contains array of hook context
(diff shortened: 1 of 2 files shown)


htdocs/commande/document.php
```diff
@@ -80,6 +80,10 @@
 if ($user->socid) {
 	$socid = $user->socid;
 }
+
+// Initialize technical object to manage hooks of page. Note that conf->hooks_modules contains array of hook context
+$hookmanager->initHooks(array('ordercarddocument'));
```
eldy, line 85: according to modulebuilder, we should have here
`$hookmanager->initHooks(array($object->element.'document', 'globalcard')); // Note that conf->hooks_modules contains array
`

PR #27888, opened 2024-01-30
NEW get html template exemple for each model

lamrani002:
# NEW|New file for display html template email
dd


eldy:
@mdeweerd is it possible to have the phpcs inside the pre-commit to support warning instead of just error ?
So it will follow same report than PHPCS. Here when a line is longer than 500 PHPCS report a warning, but phpcs by pre-commit report error.
Also, we should keep only 1 of two to save CPU usage of CI. What is the best candidate if we imagine that the next step is not just "dectecting" , but also "fixing" by submiting a PR after fix (done by phpcbf for example) ?

eldy:
@lamrani002 Can you paste a screenshot of the new feature...

mdeweerd:
Regarding pre-commit:
- #27857 (now merged) improves the extraction of the messages which allows the identication as 'warnings' - the messages are reported in emacs format which make them extractable;
- With the current pre-commit action, the php-cs checks are limited to the changed files - so the impact is limited to these files;
- I will check if pre-commit emits another error code when the php-cs task fails with a warning and ignore that code if possible.
- The other options is to add an option to the pre-commit php-cs step to not report these.
- Next step can be to activate phpcbf which is usefull only if this is followed by a commit.

Edit:
- IMHO, the pre-commit workflow is best if we want to commit changed;
- pre-commit can execute php-cbf on the changed files only;
- pre-commit runs faster than php-cs workflow on everything;
- Commit could be executed on the changed files.
- Recommendation is to setup pre-commit locally as well.

Edit2:
- The idea is to observe that when the current phpcs action fails that the pre-commit action also fails.

mdeweerd:
I looked at the pre-commit hook that is used to run php-cs - it's intent is to ignore warnings, but it is flawed.  I'll suggest an update.

PR #27891, opened 2024-01-30
FIX 17.0 - php8 warnings: test for $field existence before checking is_null

atm-florianm:
# FIX php8 warnings in CommonObject

```php
is_null($this_variable_is_not_set);
```
In php7 ⇒ notice. In php8 ⇒ warning.

Maybe this fix could be refactored by moving the `isset()` test to the beginning of the loop. I felt more comfortable doing it inside each test that has the warning because with less analysis comes less risk for misanalysis.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -9086,7 +9086,7 @@ abstract class CommonObject
 
 		foreach ($this->fields as $field => $info) {
 			if ($this->isDate($info)) {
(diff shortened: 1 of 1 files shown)


eldy:
> # FIX php8 warnings in CommonObject
> ```
> is_null($this_variable_is_not_set);
> ```
> 
> In php7 ⇒ notice. In php8 ⇒ warning.
> 
> Maybe this fix could be refactored by moving the `isset()` test to the beginning of the loop. I felt more comfortable doing it inside each test that has the warning because with less analysis comes less risk for misanalysis.

You're right. Always the minimal and less risky when changing code into maintenance version. Code quality is for develop.

PR #27898, opened 2024-01-30
FIX: Add info Tab. Update invoice.lib.php

prietojc:
FIX: Add info Tab

The info Tab at invoices has disappeared in version 18, why?

### DIFF
--- a/htdocs/core/lib/invoice.lib.php
+++ b/htdocs/core/lib/invoice.lib.php
@@ -152,6 +152,11 @@ function facture_prepare_head($object)
 	$head[$h][2] = 'agenda';
 	$h++;
 
+	$head[$h][0] = DOL_URL_ROOT.'/compta/facture/info.php?facid='.$object->id;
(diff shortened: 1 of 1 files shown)


prietojc:
And... in clean installation of V.18 the file _**/compta/facture/info.php**_ is missing

eldy:
The tab has been removed becasue another tab has been introduced called "Events/Agenda", file agenda.php. It contains all the data of the old tab Info but with more info complete list of event.

prietojc:
Ok. Honestly,  I never thought it would have moved there. Log tab still appears on supplier invoices, it's a bit confusing.

eldy:
> Ok. Honestly,  I never thought it would have moved there. Log tab still appears on supplier invoices, it's a bit confusing.

Yes, the tab Events has not yet been implemented everywhere, only few places like thirdparty, project, ticket. I hope to see more implementation on more objects

PR #27902, opened 2024-01-30
fix comment about datehour -> datetime

rycks:
Just a basic fix about comments for datehour -> datetime


rycks:
sorry -> i will change that PR to 14.0 branch

PR #27904, opened 2024-01-30
fix #27883 : zero could be a value stored

### LINKED ISSUE #27883: getDolGlobalString and string "0" returns '' or default value
### Bug

Hello
i think there is a big fail on getDolGlobalString function in race condition where we store "0" string as value ...

```
	return (string) (empty($conf->global->$key) ? $default : $conf->global->$key);
```

-> empty(0) return true then the function return default value, not "0"

just try that php code:

```
echo 'empty("0") returns ' . empty("0") . "\n";
echo 'empty(0) returns ' . empty(0) . "\n";
echo "empty('0') returns " . empty('0') . "\n";
```

results:

```
empty("0") returns 1
empty(0) returns 1
empty('0') returns 1
```

### Environment Version

_No response_

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

rycks:
If you want - for example - store a key with Product or Service code into a configuration key for your module:

```
dolibarr_set_const($db, 'ONEKEY', Product::TYPE_PRODUCT, 'chaine', 0, '', $conf->entity);
dolibarr_set_const($db, 'OTHERKEY', Product::TYPE_SERVICE, 'chaine', 0, '', $conf->entity);
```

then you store "0" as value in database for ONEKEY but getDolGlobalString will return '', not "0" ...

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -51,7 +51,7 @@ function getDolGlobalString($key)
 {
 	global $conf;
 	// return $conf->global->$key ?? '';
-	return (string) (empty($conf->global->$key) ? '' : $conf->global->$key);
+	return (string) (isset($conf->global->$key) ? $conf->global->$key : '');
 }
 
 /**
@@ -63,7 +63,7 @@ function getDolGlobalInt($key)
 {
 	global $conf;
 	// return $conf->global->$key ?? 0;
-	return (int) (empty($conf->global->$key) ? 0 : $conf->global->$key);
+	return (int) (isset($conf->global->$key) ? $conf->global->$key : 0);
 }
 
 /**


eldy:
Note: The choice was done to avoid a too large setup table parsed at each page access, to never store line when value is 0 or '0'. The setup page remove the line when we try to set a value that is '0'
If value should be a number, using getDolGlobalInt() would return 0 when value is not present into database. May be the problem is because you use getDolGlobalString instead of getDolGlobalInt() ?

rycks:
@eldy as it's explained in https://github.com/Dolibarr/dolibarr/issues/27883 that fix is already done in dolibarr 18.0 with exactly same code and my "fix" is for 14 -> 18 versions for people who make "late" upgrades ...

htdocs/core/lib/functions.lib.php
```diff
@@ -51,7 +51,7 @@ function getDolGlobalString($key)
 {
 	global $conf;
 	// return $conf->global->$key ?? '';
-	return (string) (empty($conf->global->$key) ? '' : $conf->global->$key);
+	return (string) (isset($conf->global->$key) ? $conf->global->$key : '');
```
altairis-tof, line 54: Hello, this is already fixed since v18; i dont know if @eldy still apply fixes on v14.

eldy:
> @eldy as it's explained in #27883 that fix is already done in dolibarr 18.0 with exactly same code and my "fix" is for 14 -> 18 versions for people who make "late" upgrades ...

Great. When making a backport of bug, you can mention "Backport" in the comment or better "Backport of #..." as we know the change was already validated and tested.

PR #27927, opened 2024-01-31
Simpler version of getRootURLFromURL

libremaster:
Will do the same replacement with only one line of code. I corrected the comments.


htdocs/core/lib/geturl.lib.php
```diff
@@ -390,23 +390,15 @@ function getDomainFromURL($url, $mode = 0)
...
-	$tmpurl = $url;
-	$reg = null;
-	if (preg_match('/^(https?:\/\/)/i', $tmpurl, $reg)) {
-		$prefix = $reg[1];
-	}
-	$tmpurl = preg_replace('/^https?:\/\//i', '', $tmpurl); // Remove http(s)://
-	$tmpurl = preg_replace('/\/.*$/i', '', $tmpurl); // Remove part after domain
-
-	return $prefix.$tmpurl;
+	return preg_replace('/(.*:\/\/[^\/]*).*/', '$1', $url);
```
eldy, line 401: Do we need the ending .* ?
What about 
`return preg_replace('/(.*:\/\/[^\/]*)/', '$1', $url);`

libremaster: Yes I confirm we need the ".*".
Validation is coming.

libremaster:
Hello, could you validate ?

My version works, it's a replace, not a match. The .* corresponds to the path to remove if it exists.

Thank you.

PR #27929, opened 2024-01-31
FIX #26920 use status instead of fk_statut

### LINKED ISSUE #26920: Workflow for project and task
Proposal step to follow by @eldy in (https://github.com/Dolibarr/dolibarr/pull/27929)

- [ ] Step 1 - Manage in php code the both field fk_status and status and set old one as deprecated. So existing modules are compatible and can be updated using the new field name being also compatible.

- https://github.com/Dolibarr/dolibarr/pull/27930

- [ ] Step 2 - when we are sure that step 1 us completed (we can use ->oldname and ->newname in code, for both create,update and read, with the same result), then in a next version n+x, we update field in databse with the new name and update sql requests in core code.

- https://github.com/Dolibarr/dolibarr/pull/27929

- [ ] Step 3 - Then in a next version n+y, we remove the duplicated code related to fk_status.

Old proposal by @lmag and FHenry

- PR1 : Update fkstatut in fkstatus + drop colum (UPDATE llx_project__task SET status=fk_statut; )
- PR2 : Class task.class.php replace statut by status (https://github.com/Dolibarr/dolibarr/pull/27929)
- PR3 : Replace task fkstatut by status
- PR4 : DROP COLUMN fk_statut FROM llx_projet task (Via deprecated script Dolibarr v25) (/**
	 * @var int ID
	 * @deprecated use status instead
	 */
	public $fk_statut; )
- PR6 : Take all standard class from module builder

evarisk-micka:
FIX https://github.com/Dolibarr/dolibarr/issues/26920
Update SQL task field fk_statut in status
First pull request concerning issue #26920 , others will follow to complete the work (but better go step by step)


FHenry (approved):

evarisk-micka:
@eldy can you give a review please ?
There is an inconsistency at this level and we need the status field in contract (this pr was merged https://github.com/Dolibarr/dolibarr/pull/27930 but not this one)

eldy:
Duplicating fields is never a good idea. It means we will get data in conflict.

For a transition to rename a field we must first :

Step 1 - Manage in php code the both property field ->fk_status and ->status and set old one as deprecated.
So existing modules are compatible and can be updated using the new field name being also compatible.
We can also use alias a "as status" into sql to have directly the new name as result of sql $obj->status

Step 2 - when we are sure that step 1 is completed (we can use ->oldname and ->newname in code, for both create,update and read, with the same result), then in a next version n+x, we update field in database with the new name and update sql requests in core code.

Step 3 - Then in a next version n+y, we remove the duplicated code related to fk_status.


Currentlty, there is still work to do on step 1

lmag:
I update https://github.com/Dolibarr/dolibarr/issues/26920 we check tomorrow step 1 and 2

PR #27930, opened 2024-01-31
FIX #26920 replace fk_statut by status

### LINKED ISSUE #26920: Workflow for project and task

evarisk-micka:
2nd pull request concerning issue #26920


htdocs/projet/class/task.class.php
```diff
@@ -335,7 +370,7 @@ public function fetch($id, $ref = '', $loadparentdata = 0)
 		$sql .= " t.datee as date_end,";
 		$sql .= " t.fk_user_creat,";
 		$sql .= " t.fk_user_valid,";
-		$sql .= " t.fk_statut,";
+		$sql .= " t.status,";
```
eldy, line 373: We already have a column that is used at different place.
To transition onto another name of column, we must first:
Change read sql by adding an alias
`$sql .= " t.fk_status as status,";`
everywhere the t.fk_status is in select.
Then into assignement in memory, duplicate line with old and new variable
```
$this->fk_status = $obj->status;
$this->status = $obj->status;
```

This will be the content of one major version.

In the following major version, we can then rename the field with an alter table change column oldname newname.

But first step to do is the transitionnal step by adding aliases in every fetch and having assignement duplicated.

PR #27940, opened 2024-02-01
Qual ldap class phpcs

W1W1-M:
# Qual fix phpcs ldap class methods
This PR renames **connect_bind** & **dump_content** methods to remove disabled PHPCS warnings.
I've tested this with our LDAP server.


eldy:
As this may break compatibility,can you also add a mention into the ChangeLog in the section WARNING: to alert developer that a method has been renamed ?

PR #27943, opened 2024-02-01
NEW api_thirdparties add and delete representatives

frederic34:


htdocs/societe/class/api_thirdparties.class.php
```diff
@@ -467,6 +467,68 @@ public function setThirdpartyPriceLevel($id, $priceLevel)
...
+	 * @param int		$id					Id of thirdparty
+	 * @param int       $representative_id	Id of representative
+	 * @return Object|void
+	 *
+	 * @url DELETE {id}/representative/{representative_id}
+	 */
+	public function deleteRepresentative($id, $representative_id)
+	{
+		if (!DolibarrApiAccess::$user->hasRight('societe', 'supprimer')) {
+			throw new RestException(401);
```
eldy, line 513: I think we should use 403 for permission pb.
401 for bad password, 403 for no permission
https://www.permit.io/blog/401-vs-403-error-whats-the-difference

frederic34: image

eldy: Yes, we have to fix this to switch old code from 401 to 403.

PR #27947, opened 2024-02-01
Close  #27774 new contact in project

### LINKED ISSUE #27774: Project contacts add UserGroup
### Feature Request

add the possibility to add usergroups as contacts to projects

### Use case

you can add a usergroup (with all members) in one step 
and do not have to add the members of this individually

### Suggested implementation

1. insert a line to select groups   at  /projet/contact.php

2. after selection & adding:  i see two options to handle
   A)  copy all members of the group into contacts  
       or
   B)  keep the group and link always to actual group members

### Suggested steps

_No response_

Hystepik:
New feature to have a new form contact with user groups
image

### DIFF
--- a/htdocs/core/tpl/contacts.tpl.php
+++ b/htdocs/core/tpl/contacts.tpl.php
@@ -128,7 +128,7 @@ if ($permission) {
 		<?php
 	}
 
-	if ($module == 'project') {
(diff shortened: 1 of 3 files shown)


eldy:
Merge group and user combo on same line. No default value for group.

PR #27967, opened 2024-02-01
QUAL Useless code

c3do:
Hello,

This piece of code seems completely unnecessary. However, it is used on most controllers.

if ($user->socid) {
    $socid = $user->socid;
}

Additionally, in this file, the $socid variable is already initialized.

Please confirm its uselessness

### DIFF
--- a/htdocs/contact/card.php
+++ b/htdocs/contact/card.php
@@ -99,13 +99,10 @@ $triggermodname = 'CONTACT_MODIFY';
 $permissiontoadd = $user->hasRight('societe', 'contact', 'creer');
 
 // Security check
-if ($user->socid) {
-	$socid = $user->socid;
-}
 if ($object->priv && $object->user_creation->id != $user->id) {
 	accessforbidden();
 }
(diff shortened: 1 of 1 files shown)


mdeweerd:
Here my interpretation is that by default the socid is taken from the POST variable, but when the user has a socid, then this overrides what is in the post.
It seems to prevents undesired effects in case the post request is forged with a socid that is different from the user's socid.

eldy:
> Here my interpretation is that by default the socid is taken from the POST variable, but when the user has a socid, then this overrides what is in the post. It seems to prevents undesired effects in case the post request is forged with a socid that is different from the user's socid.

I confirm. Test to overwrite $socid when $user->socid is set is important for security purpose: An external user of a company X should not have code executed with socid that has another value than X

c3do:
@eldy I just realized I was wrong. I was just going to cancel the PR, because I just understood what it was for.
However, I just gave all possible permissions to a contact.
The latter can create new third parties, but cannot do anything with them.
He can't even consult them.
What is the intention?
Should we allow an external user to create third parties?

eldy:
An external user must always be limited to its own thirdparty (in view and in edit mode if he has permission to write).
So en external user should not have permission to create a thirdparty. Surely a bug. Can you fix it if he can ?

c3do:
@eldy

PR #27968, opened 2024-02-02
Qual clean dolibarrtriggers

W1W1-M:
# Qual cleaner PHP code for dolibarrtriggers class

- Refactored code to make more readable and efficient
- Added VERSIONS array dictionary constant & depreciated individual VERSION_XXX constants
- Moved property init to constructor so that child classes can follow parent constructor
- Various fixes & optimizations to code & comments


htdocs/core/triggers/dolibarrtriggers.class.php
```diff
@@ -124,31 +143,29 @@ public function getDesc()
 	 *
 	 * @return string Version of trigger file
 	 */
-	public function getVersion()
+	public function getVersion(): string
```
eldy, line 146: As the transition from PHP7.0 to PHP7.1+ is not yet complete by a lot of users, for the moment, we must avoid the use of the type in the function
"function(): type"
We will continue to wait a little bit that the nb of users with php 7.0 is lower (A lot of users first upgrade their Dolibarr to then upgrade the PHP according to Dolibarr requirement, so having a Dolibarr not hanging on fatal error with PHP 7.0 is still important to allow migration when this direction is used) .

W1W1-M: Thanks for explanation, I'll keep this in mind.

W1W1-M: For reference https://www.php.net/manual/en/migration70.new-features.php

htdocs/core/triggers/dolibarrtriggers.class.php
```diff
@@ -30,61 +31,76 @@ abstract class DolibarrTriggers
...
 	 * @var string module is dolibarr ready
+	 * @deprecated Use self::VERSIONS
+	 * @see self::VERSIONS
 	 */
 	const VERSION_DOLIBARR = 'dolibarr';
 
+	/**
+	 * @var array dictionary of possible module states
+	 */
+	const VERSIONS = [
```
eldy, line 98: What is the benefit of using an array of const instead of const directly ?

W1W1-M: The benefit is limited to code readability. Grouping the 3 `const VERSION_XXX` consts under one array seemed more convenient and discoverable by developers, it also gives a stronger limit to the scope of values that `$version` can take.

PR #27992, opened 2024-02-04
NEW Various payment - Use list of account when edit

aspangaro:
When we edit accounting account in various payment card, it's a simple input. Possible to define an account who doesn't exist in chart of account. Modify to show list.

Before:

302082551-38c6e125-7423-4a6b-992e-8afad8fef7f0

After:

image

image


eldy:
We already have this option ACCOUNTANCY_COMBO_FOR_AUX in accountancy setup:
image

So we should reuse the same option to define if auxiliary account must use a combo list or not, like done on the filter on auxiliary account into bookeepinglist.php.

aspangaro:
Hello @eldy

I'm taking this option into consideration on line 677 of the accounting/bank/various_payment/card.php file in order to revert to a simple input if the user chooses not to use a sub-account selection list.

Do you want me to change anything else, I don't understand your request?

eldy:
> Hello @eldy
> 
> I'm taking this option into consideration on line 677 of the accounting/bank/various_payment/card.php file in order to revert to a simple input if the user chooses not to use a sub-account selection list.
> 
> Do you want me to change anything else, I don't understand your request?

I just did not found the if. Was looking for getDolGlobalString. Can you just replace the
`if (!empty($conf->global->ACCOUNTANCY_COMBO_FOR_AUX)) {`
with
`if (getDolGlobalString('ACCOUNTANCY_COMBO_FOR_AUX')) {`
The old syntax with conf->global->xxx must be abandonned;

aspangaro:
Correction done. Sorry, bad habit.

PR #27997, opened 2024-02-04
Update dbtable dump

frederic34:


htdocs/admin/system/dbtable.php
```diff
@@ -85,19 +85,20 @@
...
-		//  var_dump($link);
-
 		print '<table class="noborder">';
 		print '<tr class="liste_titre">';
-		print '<td>'.$langs->trans("Fields").'</td><td>'.$langs->trans("Type").'</td><td>'.$langs->trans("Index").'</td>';
-		print '<td></td>';
-		print '<td></td>';
-		print '<td></td>';
-		print '<td></td>';
+		print '<td>'.$langs->trans("Fields").'</td>';
```
eldy, line 90: When using pgsql, the meaning of column differs. This is why no title were output.
May be you can add a test as it is done on line 50 to support 2 case ?

PR #28000, opened 2024-02-04
FIX #27942 Accounting files export - Wrong document return for salaries

### LINKED ISSUE #27942: Wrong Salary folder/files in Export Accounting
### Bug

When exporting source document files linked to PaymentSalary are invalid when ID of PaymentSalary is different of ID of Salary

In file /compta/accounting-files.php around line 342, $subdir is made with $objd->id
`$subdir .= ($subdir ? '/' : '').dol_sanitizeFileName($objd->id);`
This is wrong because $subdir should use ID of Salary and in $objd->id we got ID of PaymentSalary

### Environment Version

>=18 for sure

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

Files displayed should be those linked to Salary of PaymentSalary

Example configuration :
Salary ID 1 with PayementSalary ID 2
Salary ID 2 with file linked
Export source document with payments salary
In the result table you will have Payment 2 (linked to Salary 1) displayed with file of Salary 2 (which is wrong)

### Attached files

image

image

image

aspangaro:
See #27942 for explanation / pictures

### DIFF
--- a/htdocs/compta/accounting-files.php
+++ b/htdocs/compta/accounting-files.php
@@ -214,9 +214,9 @@ if (($action == 'searchfiles' || $action == 'dl')) {
 			if (!empty($sql)) {
 				$sql .= " UNION ALL";
 			}
-			$sql .= " SELECT t.rowid as id, t.entity, t.label as ref, 1 as paid, amount as total_ht, amount as total_ttc, 0 as total_vat, '".$db->escape($conf->currency)."' as currency, t.fk_user as fk_soc, t.datep as date, t.dateep as date_due, 'SalaryPayment' as item, CONCAT(CONCAT(u.lastname, ' '), u.firstname)  as thirdparty_name, '' as thirdparty_code, c.code as country_code, '' as vatnum, ".PAY_DEBIT." as sens";
-			$sql .= " FROM ".MAIN_DB_PREFIX."payment_salary as t LEFT JOIN ".MAIN_DB_PREFIX."user as u ON u.rowid = t.fk_user LEFT JOIN ".MAIN_DB_PREFIX."c_country as c ON c.rowid = u.fk_country";
-			$sql .= " WHERE datep between ".$wheretail;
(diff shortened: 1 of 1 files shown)


htdocs/compta/accounting-files.php
```diff
@@ -214,9 +214,9 @@
...
-			$sql .= " FROM ".MAIN_DB_PREFIX."payment_salary as t LEFT JOIN ".MAIN_DB_PREFIX."user as u ON u.rowid = t.fk_user LEFT JOIN ".MAIN_DB_PREFIX."c_country as c ON c.rowid = u.fk_country";
-			$sql .= " WHERE datep between ".$wheretail;
+			$sql .= " SELECT t.rowid as id, t.entity, t.label as ref, paye as paid, amount as total_ht, amount as total_ttc, 0 as total_vat, '".$db->escape($conf->currency)."' as currency, t.fk_user as fk_soc, t.datesp as date, t.dateep as date_due, 'SalaryPayment' as item, CONCAT(CONCAT(u.lastname, ' '), u.firstname)  as thirdparty_name, '' as thirdparty_code, c.code as country_code, '' as vatnum, ".PAY_DEBIT." as sens";
+			$sql .= " FROM ".MAIN_DB_PREFIX."salary as t LEFT JOIN ".MAIN_DB_PREFIX."user as u ON u.rowid = t.fk_user LEFT JOIN ".MAIN_DB_PREFIX."c_country as c ON c.rowid = u.fk_country";
```
eldy, line 218: A salary can be payed in 2 times. First an advance on day x in cash then the rest on day y in check, for example.
So 1 salary can have 2 payments at 2 different dates.

So the question is what do we want to export ?

A) Here the current design is to export ithe detail of payments. So with example, it will return 2 lines.
The table for payment of salaries is payment_salary and if we change the table, it means we do not extract the same thing, we will get one line that is the salary. Why not, it is a different feature.
If what we want is this case A, if there is a problem into the link to salary or payment record, this can be fixed by getting the good id of salary from the payment.

B) But if we want to export the salary (1 line) and not the payments (2 lines), it is a different feature. In such a case, we must also modify the label of the export. Instead of calling it "Payments of salary", it must be called "Salaries". We must also know that in this case, the date and the amount in salary is not the real amount. It is just the date of start of period for the salary and the expected amount to pay. 

So:
- If what we want in the file is to list of paymet (so the case A), we must fix by adding code to get the ID of salary from the ID of salary to show correcly the good link to document (the original bug).
- if what we want if the file is the list of expected payment of salary (so the case B) , we must also fix this everywhere by changing also the label. The key 'selectpaymentsofsalaries' should also be named 'selectsalaries' instead to avoid confusion.

aspangaro: Interesting return. I'll think about it and get back to you.

PR #28003, opened 2024-02-04
FIX Accounting files export - Use th instead of td on all title columns

aspangaro:
Also remove duplicate code and use class instead of align


eldy:
As this is change related to quality or refactoring, it would be better to have pushed this into develop.

PR #28012, opened 2024-02-05
fix: auto generated user login 'undefined' appears on login input)[fi…

yacineahmaich:
fix: auto generated user login 'undefined' appears on login input)[fix: auto generated user login 'undefined' appears on login input

When We enter a lastname then we enter a firstname the generated login looks fine
Example:
    - lastname = Yacine
    - firstname = Ahmaich
    - login = yahmaich
But if we clear the firstname or the lastname we got
    - lastname = Yacine
    - firstname =  // we clear the value of firstname
    - login = yundefined

I fix the issue and also add a dot in the generated login because 'y.ahmaich' as login looks better than 'yahmaich'

### DIFF
--- a/htdocs/user/card.php
+++ b/htdocs/user/card.php
@@ -948,7 +948,7 @@ if ($action == 'create' || $action == 'adduserldap') {
 					login = "";
 					if(lastname) {
 						if(firstname) {
-							login = firstname + ".";
+							login = firstname + "'. getDolGlobalString("MAIN_USER_SEPARATOR_CHAR_FOR_GENERATED_LOGIN", "").'";
 						}
 						login += lastname;					
 					}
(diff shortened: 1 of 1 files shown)


htdocs/user/card.php
```diff
@@ -941,15 +941,19 @@
...
-						firstname = firstname.toLowerCase();
-						firstname = firstname[0];
+					lastname = $("#lastname").val().toLowerCase();
+					firstname = $("#firstname").val().toLowerCase()[0];
+
+					console.log("We generate login intell we have a valid lastname");
+					login = "";
+					if(lastname) {
+						if(firstname) {
+							login = firstname + ".";
```
eldy, line 951: Because the . introduce a behaviour change and may generate trouble for some users, can you replace the 
+ '.' 
with
+ "<php echo getDolGlobalString("MAIN_USER_SEPARATOR_CHAR_FOR_GENERATED_LOGIN"); ?>"
So default doe snot change and end user can decide the separator if he want by hidding the constant MAIN_USER_SEPARATOR_CHAR_FOR_GENERATED_LOGIN to the good char.

PR #28026, opened 2024-02-05
QUAL fix lines' title height calculation

ariel-bentu:
QUAL fix lines' title height calculation #28026

When the lines' title is higher than the default 5 (e.g., because the translated title is longer than the allocated width and wraps to the next) the line under the title is correctly placed, but the y-position for the actual lines is still at the default 5.


eldy:
Can you paste a screenshot before + after ?

ariel-bentu:
[before.pdf](https://github.com/Dolibarr/dolibarr/files/14200244/before.pdf)
[after.pdf](https://github.com/Dolibarr/dolibarr/files/14200243/after.pdf)

PR #28033, opened 2024-02-05
qual: phpstan - last errors stating "getNextValue() should return string but returns int."

thibdrev:


thibdrev:
@eldy : the above commits have been merged but the changes do not appear in Dolibarr:develop. Any idea why? 

At first I thought it might be linked to the force-pushed I did, but I see the same behaviour with PR #27990.

Any chance to get this back on track easily or should I recreate a PR?

Indeed the errors `Method mod_xxxx.php::getNextValue() should return string but returns int.` are still listed on [https://doxygen.dolibarr.org/apstats/](https://doxygen.dolibarr.org/apstats/)

thibdrev:
Ah, I think I've just found the reason why :smile: => #28071

PR #28036, opened 2024-02-06
NEW Add DolGetStatus Hook

EchoLoGeek:
NEW Add DolGetStatus Hook


htdocs/core/lib/functions.lib.php
```diff
@@ -11465,82 +11465,95 @@ function dolGetBadge($label, $html = '', $type = 'primary', $mode = '', $url = '
  */
 function dolGetStatus($statusLabel = '', $statusLabelShort = '', $html = '', $statusType = 'status0', $displayMode = 0, $url = '', $params = array())
 {
-	global $conf;
+	global $conf, $object, $action, $hookmanager;
```
eldy, line 11468: $object should no more be used as a global variable. It generates a lot of bad side effect.
You can pass the $object into $params['object']

PR #28039, opened 2024-02-06
Fix show desact projects in os card using small penci

vmaury:
#FIX When creating or modifying order supplier with complete form, only active projects (if activated) are displayed in select (ajax or not)

But when using the small pencil, desactivated project are displayed

### DIFF
--- a/htdocs/fourn/commande/card.php
+++ b/htdocs/fourn/commande/card.php
@@ -2086,7 +2086,7 @@ if ($action == 'create') {
 			if ($action != 'classify' && $caneditproject) {
 				$morehtmlref .= '<a class="editfielda" href="'.$_SERVER['PHP_SELF'].'?action=classify&token='.newToken().'&id='.$object->id.'">'.img_edit($langs->transnoentitiesnoconv('SetProject')).'</a> ';
 			}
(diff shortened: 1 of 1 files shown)


eldy:
Disabled project are included in edit because if the order has a closed project, when we edit, the combo list must remain on the project of the order.
So in create mode we should exclude open project but, when editing, the closed project must also be included into combo list so the current project of the order does not disappear

vmaury:
If you watch the code of the project select, even if closed project are hidden ($discard_closed = 1), the current project - even if it's closed - is displayed 

Cf line ~225 of     dolibarr/htdocs/core/class/html.formprojet.class.php, function FormProjets.select_projects_list

`if ($discard_closed == 1 && $obj->fk_statut == 2 && $obj->rowid != $selected) { // We discard closed except if selected`

eldy:
> If you watch the code of the project select, even if closed project are hidden ($discard_closed = 1), the current project - even if it's closed - is displayed
> 
> Cf line ~225 of dolibarr/htdocs/core/class/html.formprojet.class.php, function FormProjets.select_projects_list
> 
> `if ($discard_closed == 1 && $obj->fk_statut == 2 && $obj->rowid != $selected) { // We discard closed except if selected`

Good catch 👍🏻

PR #28042, opened 2024-02-06
FIX:  extrafields linked object type on external module object

atm-jpb:
# FIX : test linkedObject extrafield type

When adding an extrafields of type linkedObject to an object from an external module, on a record of the object to which this extrafields is added and on the list, errors may occur sometimes, especially if ajax_autocompleter is called.

In selectobject.php, the test is only performed on objectfield, which is the calling object, without checking for the presence of objectdesc, which is the linked object type.

I have added this test in this file as well as in html.form.class.php.

In the call of the selectForForms function, I pass the complete array so that it remains operational in the following functions.

### DIFF
--- a/htdocs/core/ajax/selectobject.php
+++ b/htdocs/core/ajax/selectobject.php
@@ -53,7 +53,7 @@ if (empty($htmlname)) {
 	httponly_accessforbidden('Bad value for param htmlname');
 }
 
-if (!empty($objectfield)) {
+if (!empty($objectfield && empty($objectdesc))) {
 	// Recommended method to call selectobject.
 	// $objectfield is Object:Field that contains the definition (in table $fields or extrafield). Example: 'Societe:t.ddd' or 'Societe:options_xxx'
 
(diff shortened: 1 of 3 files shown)


htdocs/core/ajax/selectobject.php
```diff
@@ -53,7 +53,7 @@
 	httponly_accessforbidden('Bad value for param htmlname');
 }
 
-if (!empty($objectfield)) {
+if (!empty($objectfield && empty($objectdesc))) {
```
eldy, line 56: Pb in parenthesis

htdocs/core/ajax/selectobject.php
```diff
@@ -53,7 +53,7 @@
 	httponly_accessforbidden('Bad value for param htmlname');
 }
 
-if (!empty($objectfield)) {
+if (!empty($objectfield) && empty($objectdesc)) {
```
eldy, line 56: I don't understand why you need this. Allowing a user to force himself the value of objectdesc as parater is a security hole.
When we call ajax/selectobject, the main parameter is objectfield. And from it, we should be able to retreive all information of objectdesc.
The objectdesc should not be provided as parameter (a end user must not be able to bypass definition nown by dolibarr).
What value do you have for $objectfield ?
And what value do you have in the definition table llx_extrafields of the extrafield mentionned into $objectfield ?

PR #28055, opened 2024-02-07
fix missing customer and vendor name from title

sonikf:
Errors:
Undefined variable $soc
Attempt to read property "name" on null


htdocs/compta/facture/list.php
```diff
@@ -582,6 +582,8 @@
 $companyparent = new Societe($db);
 $company_url_list = array();
 $subtypearray = $object->getArrayOfInvoiceSubtypes(0);
+$soc = new Societe($db);
+$soc->fetch($socid);
```
eldy, line 586: For the list $socid is most often empty.

sonikf: It's about the thirdparty own list

PR #28058, opened 2024-02-07
fix warning

frederic34:
image


htdocs/core/class/commoninvoice.class.php
```diff
@@ -1999,4 +1999,5 @@ abstract class CommonInvoiceLine extends CommonObjectLine
 	public $fk_user_modif;
 
 	public $fk_accounting_account;
+	public $code_ventilation;
```
eldy, line 2002: This seems deprecated. 
It has been replaced with fk_accounting_account.

frederic34: image

PR #28079, opened 2024-02-08
Qual: ci: Run pre-commit/php-cs with cache

mdeweerd:
This adds a hook to .pre-commit-config.yaml and updates the workflow to run php-cs with cache when it is run for all files. When running on changed files only, the cache is not useful.

The php-codesniffer ruleset.xml was cleaned up (duplicates removal/formatted)

### DIFF
--- a/dev/setup/codesniffer/ruleset.xml
+++ b/dev/setup/codesniffer/ruleset.xml
@@ -3,6 +3,7 @@
 <ruleset name="Dolibarr">
 	<description>Dolibarr coding standard.</description>
 	<arg name="tab-width" value="4"/>
+	<arg name="extensions" value="php" />
 
 	<exclude-pattern type="relative">build/html</exclude-pattern>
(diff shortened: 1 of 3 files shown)


dev/setup/codesniffer/ruleset.xml
```diff
@@ -17,8 +18,7 @@
 	<exclude-pattern type="relative">*/nltechno*</exclude-pattern>
 	<exclude-pattern type="relative">source</exclude-pattern>
 	<exclude-pattern type="relative">.git</exclude-pattern>
-	<exclude-pattern>htdocs/includes</exclude-pattern>
```
eldy, line 20: This directory contains code that we do not control. They are external sources.
So if we apoly the dolibarr rule checks on it, we should get a lot of errors, shouldn't we ?

mdeweerd: I agree about the directory but it's already excluded in a rule above.

Issue #28070, opened 2024-02-08
Missing Mobile Phone field on thirdParty


lmag:
### Feature Request

Missing Mobile Phone field on thirdParty

bos4711:
Bug confirmed for v17.0.2. The field "Mobile Phone" can be found under a Thirdparty contact, but not under the Thirdparty itself (which OP is reporting about).

JonBendtsen:
18.0.4 - there is a phone field on thirdparties, and I just use that for my mobilePhone.

PR #28110, opened 2024-02-10
Attempt at fixing #28070 with mobile phone to a thirdparty

JonBendtsen:
NEW #28070 Adding mobile phone to thirdparty
Attempt at adding mobile phone to thirdparty. Most code is a direct copy of how phone and fax was setup for a thirdparty just modified to be about mobile phone.

Parts with images are inspired from phone_mobile under contacts.


JonBendtsen:
I can't seem to find the error for why Tavis CI build check fails?

PR #28119, opened 2024-02-11
qual: phpstan for htdocs/loan/class/loanschedule.class.php

thibdrev:
htdocs/loan/class/loanschedule.class.php	308	Property LoanSchedule::$num_payment (int) does not accept string.
htdocs/loan/class/loanschedule.class.php	317	Property LoanSchedule::$fk_bank (int) does not accept string.

### DIFF
--- a/htdocs/loan/class/loanschedule.class.php
+++ b/htdocs/loan/class/loanschedule.class.php
@@ -305,7 +305,7 @@ class LoanSchedule extends CommonObject
 			$this->amount_interest = trim($this->amount_interest);
 		}
 		if (isset($this->num_payment)) {
-			$this->num_payment = trim($this->num_payment);
+			$this->num_payment = (int) $this->num_payment;
 		}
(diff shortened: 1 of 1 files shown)


htdocs/loan/class/loanschedule.class.php
```diff
@@ -305,7 +305,7 @@ public function update($user = 0, $notrigger = 0)
 			$this->amount_interest = trim($this->amount_interest);
 		}
 		if (isset($this->num_payment)) {
-			$this->num_payment = trim($this->num_payment);
+			$this->num_payment = (int) $this->num_payment;
```
eldy, line 308: I think num_payment can be ABC123, so string

thibdrev: Correct. This is indeed something we already saw in the past: https://github.com/Dolibarr/dolibarr/commit/aafe309b5814de27a6aced88a507747463b25878.

Now corrected.

PR #28127, opened 2024-02-11
Qual: Fix apstats & add phan to apstats.

mdeweerd:
# Qual: Fix apstats & add phan to apstats

I fixed several issues that I ran into when running apstats.php locally to test the phan integration.

So this commit only proposes the updates to apstats.php.

I added a test that disables phan when its configurations is not found so it should be fine to integration in develop.


mdeweerd:
Here's a sample result - the table is searchable:
[phan_apstats.zip](https://github.com/Dolibarr/dolibarr/files/14234414/phan_apstats.zip)

PR #28132, opened 2024-02-12
outputlangs must not be a string as default value like others generat…

rycks:
# FIX default value for outputlangs is set to ''

That makes a bug in case of calling generateDocument without parameters. There is only 3 functions of that sort in all dolibarr code, all of others don't have default values for outputlangs then that PR is just a "uniformisation" of code.

### DIFF
--- a/htdocs/delivery/class/delivery.class.php
+++ b/htdocs/delivery/class/delivery.class.php
@@ -1044,7 +1044,7 @@ class Delivery extends CommonObject
 	 *  @param     int			$hideref        Hide ref
 	 *  @return    int             				0 if KO, 1 if OK
 	 */
-	public function generateDocument($modele, $outputlangs = '', $hidedetails = 0, $hidedesc = 0, $hideref = 0)
+	public function generateDocument($modele, $outputlangs, $hidedetails = 0, $hidedesc = 0, $hideref = 0)
 	{
 		global $conf, $user, $langs;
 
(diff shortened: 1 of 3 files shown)


htdocs/delivery/class/delivery.class.php
```diff
@@ -1044,7 +1044,7 @@ public function setDeliveryDate($user, $delivery_date)
 	 *  @param     int			$hideref        Hide ref
 	 *  @return    int             				0 if KO, 1 if OK
 	 */
-	public function generateDocument($modele, $outputlangs = '', $hidedetails = 0, $hidedesc = 0, $hideref = 0)
+	public function generateDocument($modele, $outputlangs, $hidedetails = 0, $hidedesc = 0, $hideref = 0)
```
eldy, line 1047: If you remove the default value, this will break compatibility (fatal error on code that call the method without the outputlang defined).
What about using `  $outputlangs = null`  instead ?

rycks: impossible to break compatibilty because if somebody use  default parameter that will bug due to 3 lines after

`
		$outputlangs->load("something");
`

And all others function generateDocument in all dolibarr sources are without default value :)

eldy: If an external module do
generateDocument($modele);
I've got a fatal error.

PR #28135, opened 2024-02-12
Qual: Correct spelling errors

mdeweerd:
# Qual: Correct spelling errors

Fix new error in Changelog


mdeweerd:
Note: when using "pre-commit" locally, you can avoid these errors.
When commiting regularly to the main branches, I suggest to set "SKIP=no-commit-to-branch" in shell configuration scripts.

PR #28137, opened 2024-02-12
18.0 FIX error message for stock disable consumption

samifilali:
# FIX
- error message for stock disable consumption

To merge after : https://github.com/Dolibarr/dolibarr/pull/28140

### DIFF
--- a/htdocs/mrp/class/mo.class.php
+++ b/htdocs/mrp/class/mo.class.php
@@ -373,6 +373,13 @@ class Mo extends CommonObject
 		unset($object->fk_user_creat);
 		unset($object->import_key);
 
+		// Remove produced and consumed lines
+		foreach ($object->lines as $key => $line) {
+			if (in_array($line->role, array('consumed', 'produced'))) {
+				unset($object->lines[$key]);
+			}
+		}
+
 		// Clear fields
 		$object->ref = empty($this->fields['ref']['default']) ? "copy_of_".$object->ref : $this->fields['ref']['default'];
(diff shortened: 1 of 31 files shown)


htdocs/mrp/class/mo.class.php
```diff
@@ -1861,6 +1861,10 @@ public function create(User $user, $notrigger = false)
 			$this->error = 'BadValueForQty';
 			return -1;
 		}
+		if (empty($this->fk_stock_movement) && $this->fields['disable_stock_change']['enabled']) {
```
eldy, line 1864: Don't understand this addition.
mrp_production is the table that contains lines for product to consume, product consumed, product to produce and product produced.
We may not know the stock we will use before consumption or production is used. So we $this->fk_stock_movement may be empty and we should be able to insert line in mrp_production to prepare the MO.

htdocs/mrp/class/mo.class.php
```diff
@@ -1869,6 +1869,11 @@ public function create(User $user, $notrigger = false)
 			return -1;
 		}
 
+		if (empty($this->fk_stock_movement) && $this->fields['disable_stock_change']['enabled'] && $this->role=='consumed') {
+			$this->error = 'StockChangeDisable';
```
eldy, line 1873: It seems there is a misunderstanting of what mean disable_stock_change
It means that we won't update the stock during the fabrication. It is not a lock.

eldy:
No feedback on last comment.
Problem seems misunderstanding of a feature, so i close ticket.

PR #28138, opened 2024-02-12
Fix some php warnings and errors

Hystepik:
Fix an SQL error and some warnings in php code


htdocs/core/class/commonobject.class.php
```diff
@@ -8075,7 +8075,7 @@ public function showOutputField($val, $key, $value, $moreparam = '', $keysuffix
 				}
 			}
 		} elseif ($type == 'chkbxlst') {
-			$value_arr = explode(',', $value);
+			$value_arr = (isset($value) ? explode(',', $value) : array(''));
```
eldy, line 8078: array() instead of areay('') would be better i think

PR #28143, opened 2024-02-12
Fix: PhanParamSignatureMismatch

mdeweerd:
# Fix: PhanParamSignatureMismatch

PhanDoc was updated to require specific types which changes the method signature versus the parent.
This results in a signature mismatch.
Updated to maintain the method signature but dynamically check that the object type is correct.

Maybe get_substitutionarray_object should call the parent calss if the object type does not match the expected one.

### DIFF
--- a/htdocs/core/modules/member/doc/doc_generic_member_odt.class.php
+++ b/htdocs/core/modules/member/doc/doc_generic_member_odt.class.php
@@ -2,6 +2,7 @@
 /* Copyright (C) 2010-2012 	Laurent Destailleur <eldy@users.sourceforge.net>
  * Copyright (C) 2012		Juanjo Menent		<jmenent@2byte.es>
  * Copyright (C) 2018-2021  Frédéric France         <frederic.france@netlogic.fr>
+ * Copyright (C) 2024		MDW							<mdeweerd@users.noreply.github.com>
 *
 * This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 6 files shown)


htdocs/core/modules/member/doc/doc_generic_member_odt.class.php
```diff
@@ -413,7 +414,6 @@ public function write_file($object, $outputlangs, $srctemplatepath, $mode = 'mem
 		return -1;
 	}
 
-	// phpcs:disable PEAR.NamingConventions.ValidFunctionName.ScopeNotCamelCaps
```
eldy, line 416: Removing this generate errors with phpcs run by Travis. This is the expected behaviour.
Do yo know why it does not report error with phpcs run by pre-commit ?

mdeweerd: I have made some changes to the phpcs github action but I want to test it better.
The core issue is that phpcs returns exit code 1 even for errors, so I updated the test in the github action but did not succeed  yet in adding a testcase for the github action to validate it.

mdeweerd: I validated the updated php-cs pre-commit hook with this PR (I attached the updated configuration before fixing the issues).

mdeweerd:
The attached pre-commit hook configuration fixes the absence of report in php-cs.

PR #28154, opened 2024-02-12
Qual: Apply automatic phan fixes (deprecations, unneeded imports)

mdeweerd:
# Qual: Apply automatic phan fixes (deprecations, unneeded imports)
    
This applies automatic fixes by phan for deprecated functions, unneeded imports).


mdeweerd:
One commit/file which may help in case of merge conflicts.

PR #28170, opened 2024-02-14
NEW: Add global STOCK_MOVEMENT_FORCE_DO_NOT_CLEAN_EMPTY_LINES to force do not clean empty lines on movement stock

kkhelifa-opendsi:
Add global STOCK_MOVEMENT_FORCE_DO_NOT_CLEAN_EMPTY_LINES to force do not clean empty lines on movement stock.

Because customer need to see the warehouse when the product previously even if the stock is at 0.

And when in the reassort page, the product with a stock at 0 don't display and block the repleinishment of the stock of the product.

### DIFF
--- a/htdocs/product/stock/class/mouvementstock.class.php
+++ b/htdocs/product/stock/class/mouvementstock.class.php
@@ -611,7 +611,7 @@ class MouvementStock extends CommonObject
 				}
 			}
 
-			if (empty($donotcleanemptylines)) {
+			if (empty($donotcleanemptylines) && !getDolGlobalInt('STOCK_MOVEMENT_FORCE_DO_NOT_CLEAN_EMPTY_LINES')) {
 				// If stock is now 0, we can remove entry into llx_product_stock, but only if there is no child lines into llx_product_batch (detail of batch, because we can imagine
 				// having a lot1/qty=X and lot2/qty=-X, so 0 but we must not loose repartition of different lot.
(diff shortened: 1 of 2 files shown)


eldy:
You say "And when in the reassort page, the product with a stock at 0 don't display and block the repleinishment of the stock of the product."
If a value for "Desired stock" is defined, even if stock is empty, you should see the product in reassort. Don't you ?

kkhelifa-opendsi:
Without the globale STOCK_MOVEMENT_FORCE_DO_NOT_CLEAN_EMPTY_LINES and also with PRODUCT_STOCK_LIST_SHOW_VIRTUAL_WITH_NO_PHYSICAL set to 1, we have :
Capture d’écran_2024-02-28_09-47-48
Capture d’écran_2024-02-28_09-50-55

With STOCK_MOVEMENT_FORCE_DO_NOT_CLEAN_EMPTY_LINES set to 1, we have:
Capture d’écran_2024-02-28_09-51-51
Capture d’écran_2024-02-28_09-52-13
Capture d’écran_2024-02-28_09-52-37

eldy:
On the first screenshot, you have  "Stock désiré optimal" to 0. So it means you want no stock at all for this product. So it is not suggested in the replenishment. This looks the expected behaviour.
If you want to not have it excluded, set the "Stock désiré optimal" to a value or keep it empty (not defined means no instruction to exclude it so it should appear, shouldn't it ?).

PR #28171, opened 2024-02-14
NEW: Add global PRODUCT_STOCK_LIST_SHOW_WITH_COMPILED_PHYSICAL_STOCK for this is based on compiled physical stock

kkhelifa-opendsi:
Add global PRODUCT_STOCK_LIST_SHOW_WITH_COMPILED_PHYSICAL_STOCK for this is based on compiled physical stock, not the sum of product stocks on existing links.

Because products without a warehouse no longer appear.


eldy:
I suppose this is for perf purpose ?
Note: i have renamed the var into PRODUCT_STOCK_LIST_SHOW_WITH_PRECALCULATED_DENORMALIZED_PHYSICAL_STOCK
because i know we ill search on "denormalized" or "precalculated" in a future to refind it...

Issue #28168, opened 2024-02-14
SecurityTest::testGetURLContent fails - ftp connection is allowed despite default options


mdeweerd:
### Bug

The test case wants to validate the an ftp request is not possible, however it gets a "Connection refused back" with a port number.
To make sure that an attempt to connect was made, I added a valid ftp server and got the following reply:

```
1) SecurityTest::testGetURLContent
Did not get 'not supported' in Access denied: 530
Failed asserting that 0 is true.
```

The reason this is possible is not obvious from the code.

### Environment Version

develop/8592f6ba492/ed1e818ada9

### Environment OS

Windows 10

### Environment PHP

c:\wamp64\bin\php\php7.4.33\
phpunit 9.5

### Steps to reproduce the behavior

Run SecurityTest with phpunit (on windows).

PR #28172, opened 2024-02-14
SEC: #28168 Correct protocol limitations (PHP7.4/Win)

mdeweerd:
# SEC: #28168 Correct protocol limitations (PHP7.4/Win)

Protocol limitation was not active during test on windows platform.
Moving the application of the limitation just before the curl_exec
instruction made the limitation effective.

Also extended the code to enable allowing ftp and ftps and extended
the code for [CURLOPT_REDIR_PROTOCOLS_STR](https://www.php.net/manual/en/curl.constants.php#constant.curlopt-redir-protocols-str).


mdeweerd:
PHPstan does not recognize the new constant yet.  I let "you" add the exception to the phan configuration for that.

PR #28174, opened 2024-02-14
FIX: Shipment closing action has wrong value

n15c:
Change the value of the action parameter from "close" to "classifyclosed" to make the button work again.


JonBendtsen:
There are 14 other files that also contain `action=close&token=` I wonder if they should also be changed to classifyclosed?

Further more, not all the lines are formatted the same way, some use `dolGetButtonAction` others `<a class="butActionDelete"` and a `<div class="inline-block divButAction"><a class="butAction`

Some of the `dolGetButtonAction` has a `dolGetButtonAction('', $langs->trans`  others `dolGetButtonAction($langs->trans(`

PR #28175, opened 2024-02-14
fix warning

frederic34:
image
@mdeweerd

### DIFF
--- a/htdocs/ticket/index.php
+++ b/htdocs/ticket/index.php
@@ -1,7 +1,7 @@
 <?php
-/* Copyright (C) - 2013-2016    Jean-François FERRY     <hello@librethic.io>
- * Copyright (C) - 2019         Nicolas ZABOURI         <info@inovea-conseil.com>
- * Copyright (C) 2021		Frédéric France				<frederic.france@netlogic.fr>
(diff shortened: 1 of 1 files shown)


htdocs/ticket/index.php
```diff
@@ -376,7 +376,6 @@
 				$object->id = $objp->rowid;
 				$object->ref = $objp->ref;
 				$object->track_id = $objp->track_id;
-				$object->fk_statut = $objp->status;
```
mdeweerd, line 377: Attention, this should only be done when the DolDeprecatedHandler is integrated & when fk_statut is set as a deprecated property and mapped to status.

fk_status is not in the deprecatedProperties function in commonobject.class, but one can extend that method in a subclass.

```php
    protected function deprecatedProperties()
    {
        return ["fk_statut"=>"status"] + parent::deprecatedProperties();
    }
```

Maybe I should just add it to my branch.

frederic34: it is just for display tooltip

mdeweerd: Ok, that's why I do not touch this type of code as I do not have enough context 😉

PR #28179, opened 2024-02-14
use browser cache for pictures

frederic34:


frederic34:
@mdeweerd what do you think of this param for chackout: [https://github.com/marketplace/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit](url)

mdeweerd:
> https://github.com/marketplace/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit

In my opinon this is not useful - the HEAD commit could be further than the current commit.  I think that it could cause confusions.
What did you have in mind as a benefit?  Maybe I miss something.


The original fetches:
```
 /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=1 origin +4a307c0a92cdd76342ab62d308d38d39bdf3aec6:refs/remotes/pull/28179/merge
  From https://github.com/Dolibarr/dolibarr
   * [new ref]         4a307c0a92cdd76342ab62d308d38d39bdf3aec6 -> pull/28179/merge
``` 

The one referencing the head fetches:

```
  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune --no-recurse-submodules --depth=1 origin +0ca824e9d2b477c5a33bd4dabc02e3d2bc6111e7:refs/remotes/origin/patch-12
  From https://github.com/frederic34/dolibarr
   * [new ref]         0ca824e9d2b477c5a33bd4dabc02e3d2bc6111e7 -> origin/patch-12
```

When I look at the history, I can find the second hash, but not the first hash.
(I make a new comment for my thought).

```
| * 0ca824e9d2b (frederic34/patch-12) test
| * ba3174d0329 use browser cache for pictures
| | *   fb0c1f52e77 (frederic34/patch-9) Merge branch 'develop' into patch-9
| | |\  
| | |/  
| |/|   
| | * feaef602ddf fix
| | * cddc7263af4 phpdoc
| | | *   b3d8bbfdd01 (frederic34/patch-8) Merge branch 'develop' into patch-8
```

frederic34:
I'm looking why we have 
image

mdeweerd:
The merge hash is not found in the history because that islikely a merge between the target branch and the source branch.  So it's a mix of the target and the source.

The HEAD commit is on the source branch and is probably further away from a merge between the source branch and the target branch.

So if my source branch was built on 17.0 and the PR is towards develop, the workflow using the HEAD commit would run on code that is using the 17.0 codebase, while the workflow using the merge comit would run on code using the merge between develop and the source branch.

So we want the merge commit and not the HEAD commit.

mdeweerd:
> I'm looking why we have image

In pdf_standard, the page_head module is void and elsewhere it returns the offset.  And that offset is used in certain locations - phpstan seems to randomly decide on using void or the offset (int or float).

PR #28181, opened 2024-02-14
New substitution variables for invoice

SylvainLegrand:


htdocs/core/lib/functions.lib.php
```diff
@@ -8633,6 +8633,9 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 				}
 				if (is_object($object) && $object->element == 'facture') {
 					$substitutionarray['__URL_INVOICE__'] = DOL_MAIN_URL_ROOT."/compta/facture/card.php?id=".$object->id;
+					$substitutionarray['__FACDATE__'] = isset($object->date) ? dol_print_date($object->date, 'daytext', 0, $outputlangs) : '';
```
eldy, line 8636: We should already have this. See
https://wiki.dolibarr.org/index.php?title=Variable_substitution_system#Objects_with_amounts

htdocs/core/lib/functions.lib.php
```diff
@@ -8633,6 +8633,9 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
...
+					$substitutionarray['__FACDATELIMREG__'] = isset($object->date_lim_reglement) ? dol_print_date($object->date_lim_reglement, 'daytext', 0, $outputlangs) : '';
```
eldy, line 8637: Same, we already have __DATE_DUE__ 
https://wiki.dolibarr.org/index.php?title=Variable_substitution_system#Objects_with_amounts

SylvainLegrand: @eldy 

you are almost right, the date format is not the same: day versus daytext.

frederic34: why not __DATE_DUE_DAYTEXT__ and __DATE_DAYTEXT__ ?

SylvainLegrand: Why not ?
It's ok for me if we just have to change the variables names

PR #28184, opened 2024-02-15
Add dolistore_id on module

defrance:
the goal is to verify module version directly on dolistore

### DIFF
--- a/htdocs/core/modules/DolibarrModules.class.php
+++ b/htdocs/core/modules/DolibarrModules.class.php
@@ -48,6 +48,11 @@ class DolibarrModules // Can not be abstract, because we need to instantiate it 
 	public $numero;
 
 	/**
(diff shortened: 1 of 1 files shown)


eldy:
Dolistore is not the only platform that will store external modules. Also when moving dolistore to another system, id may change. So i suggest to be ready for future need, to not store this as an id but as an array. For example:

$ids_on_external_stores = array('dolistore'=>....,  'dolistorev2'=>...., 'otherstore'=>....)

rycks:
Maybe a list of (tiny)url, not just id ?

defrance:
maybe on this case, add some other link like github, X etc,
so the ID is not really pertinent

eldy:
So the var could be named
$url_on_external_systems

PR #28187, opened 2024-02-15
Qual: Run all tests + ignore some exceptions for Windows CI

mdeweerd:
# Qual: Run all tests + ignore some exceptions for Windows CI

## Catch exceptions when loading tests with false Security Issue.
    
When main.inc.php is loaded a security exception is detected on  the windows ci platform.
This update aims to avoid not running the other tests.

## Do not fail early when running tests.
    
Finishing the phpunit tests does not require a lot of extra time so rather than quiting early, run them all.

Also, when needed this option can be enabled on the CLI, but it can not be disabled.  So when desired, it can still be added in a ci-flow or local script.


mdeweerd:
The testcases now run (with errors) and are therefore more useful.

I can probably also re-enable the open_basedir restriction - either before or after these testcases pass on windows.

Note: SecurityTests and possibly WebsiteTests are skipped because a security issue is reported when those class files are loaded when 'main.inc.php' is included.

mdeweerd:
The win-ci workflow now attaches an artefact containing the files `documents/dolibarr.log` and the output from php_unit as `phpunit_tests.log`.

We now have the following results with the base code (excluding tests that are skipped because of main.inc.php):
```plaintext
Tests: 366, Assertions: 77417, Errors: 7, Failures: 12, Skipped: 15.
```
EDIT: I implemented a workaround for the false security issue.  Almost 300 extra assertions in 16 extra tests, but no new failures.

```
Tests: 382, Assertions: 77598, Errors: 7, Failures: 12, Warnings: 1, Skipped: 15.
Currently:
Tests: 382, Assertions: 78138, Errors: 6, Failures: 12, Skipped: 15.
```

mdeweerd:
The PHPUNIT errors from the Windows workflow are now converted into GitHub annotations.

mdeweerd:
After setting up the built-in php server, the test status is now:

```
Tests: 385, Assertions: 78228, Errors: 4, Failures: 3, Skipped: 5.
```

The 4 errors are related to open_base dir handling.
2 of the 3 failures are already "under discussion" in 2 PR.
The 3rd failure is due to the fact that a PHP script is trying to be run using it's shebang which does not work on windows - the test would need to be updated to execute it using the php executable (if the desire is to test the script on windows), or skip it.

I will open PRs for the issues that were fixed but not yet opened as a PR.

mdeweerd:
Not jsst cygwin:
````
docker run -v $(cygpath -w $PWD):/root --workdir=/root debian bash -c "chmod +x bash_unit && tests/test_doc.sh"
cat: README.adoc: No such file or directory
```

For you this would be 
```
docker run -v $PWD):--workdir=/root debian bash -c "chmod +x bash_unit && tests/test_doc.sh"
```

PR #28213, opened 2024-02-16
Fix null dir_output

W1W1-M:
# Fix null dir_output 
Following warnings on agenda event cards & resource cards :
Undefined property: Conf::$unknown (/var/www/dolibarr/20.0/htdocs/core/class/commonobject.class.php:9058)
Attempt to read property "dir_output" on null (/var/www/dolibarr/20.0/htdocs/core/class/commonobject.class.php:9058)

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -9055,8 +9055,7 @@ abstract class CommonObject
 		}
 
 		$id = $this->id;
-
-		$dir = $conf->$newmodulepart->dir_output;
+		$dir = $conf->$newmodulepart->dir_output ?? '';
 		if (!empty($this->photo)) {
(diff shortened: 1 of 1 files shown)


mdeweerd:
Also some issues with dir_output here (in the first case a nul value may not be very nice, the second case is a duplicate if condition so different)

```plaintext
htdocs\compta\paiement\cheque\class\remisecheque.class.php:629 PhanPluginUnknownObjectMethodCall Phan could not infer any class/interface types for the object of the method call $docmodel->write_file($this, ($conf->bank->dir_output . '/checkdeposits'), $this->ref, $outputlangs) - inferred a type of object
htdocs\core\lib\files.lib.php:3148 PhanPluginDuplicateIfCondition Saw the same condition (($modulepart == 'actions') && !empty($conf->agenda->dir_output)) in an earlier if/elseif statement
```

PR #28214, opened 2024-02-16
Qual: Add test case demonstrating the security checks for phpunit on win-ci

mdeweerd:
# Qual: Add test case demonstrating/testing the security checks for phpunit

When testing in Windows CI the tests including main.inc.php fail because of a security check.

The 2nd test added demonstrates when this happens

@eldy I let you implement (i.e. choose & code) the method to not have this false positive or bypass it in tests.

EDIT: The error 'Undefined array key "REQUEST_URI"' is another sideeffect which does not occur on my local machine.  However, REQUEST_URI is not set when running PHPUNIT and I suppose this happens when the report message is created from the injection checking code.

### DIFF
--- a/test/phpunit/SecurityTest.php
+++ b/test/phpunit/SecurityTest.php
@@ -1,6 +1,7 @@
 <?php
 /* Copyright (C) 2010 Laurent Destailleur  <eldy@users.sourceforge.net>
  * Copyright (C) 2023 Alexandre Janniaux   <alexandre.janniaux@gmail.com>
+ * Copyright (C) 2024		MDW							<mdeweerd@users.noreply.github.com>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


test/phpunit/SecurityTest.php
```diff
@@ -224,62 +225,71 @@ public function testSqlAndScriptInjectWithPHPUnit()
...
+		$test = "/dolibarr/htdocs/index.php/".chr('246')."abc";	// Add the char %F6 into the variable
+		$result = testSqlAndScriptInject($test, 2);
 		//print "test=".$test." result=".$result."\n";
 		$this->assertGreaterThanOrEqual($expectedresult, $result, 'Error on testSqlAndScriptInject with a non valid UTF8 char');
+
+		$vars = array("PHP_SELF" => '/c/phptools/bin/phpunit');
+		$result = analyseVarsForSqlAndScriptsInjection($vars, 2);
+		$this->assertTrue($result, "Error on testSqlAndScriptInject for {$vars['PHP_SELF']} that should be accepted");
+
+		$vars = array("PHP_SELF" => 'c:\tools\bin\phpunit');
```
eldy, line 289: According to PHP doc PHP_SELF should be the relative path from the web server. 
So having a content like c:\tools\bin\phpunit is not an expected content.
What value do you have for $_SERVER["PHP_SELF"]  ?

mdeweerd: That is exactly the value I have.

eldy: > That is exactly the value I have.

Looks strange. Which webserver are you using to have this value ?

mdeweerd: No webserver - just phpunit.
See #28187 https://github.com/Dolibarr/dolibarr/actions/runs/7969576072 .
And now also #28301 https://github.com/Dolibarr/dolibarr/actions/runs/7969600927 .

eldy:
PR #28313 seems better to me so i close this one

mdeweerd:
#28313 is a band-aid solution IMHO and only applied when running AllTests, so it could still pose issues when running tests individually.

PR #28226, opened 2024-02-16
Fix: Fix var_dump checker

mdeweerd:
# Fix: Fix var_dump checker
   
The core issue was that in PHP whitespace includes newlines by default, the m modifier
is needed to not match multilines.

# Discussion:
On windows the test found a file with an unauthorized var_dump.  Turns out that this was due the file filter not working because the separator is \ on windows and / is being matched.

I update the test, but the real question is: do the functions in files.lib.php need to be updated knowing that in php the '/' eventually convert to the system's directory separator when accessing files (I did not formally check that that is true, but I suppose it is).
I also notice that the $path in dol_dir_list is cleaned, but it supposes that it's a unix-like path.

To ensure that the regexes work, there should also be some files inside a test related directory where the tests can find positive matches.  The test should then exclude the match from the errors, but increment a counter for the false matches to be certain that the expected count of false matches was found.

### DIFF
--- a/test/phpunit/CodingPhpTest.php
+++ b/test/phpunit/CodingPhpTest.php
@@ -599,17 +599,68 @@ class CodingPhpTest extends CommonClassTest
 		//    - Any character not / or *
 		//    - Any / not preceded with / and not followed by / or *
 		//    - Any * not preceded with /
-		preg_match_all('{(?:^[ \t]*|(?:(?:[^*/]|(?<![^/])/(?![*/])|(?!/)\*)(\S)))\bvar_dump\(}m', $filecontent, $matches, PREG_SET_ORDER);
+		preg_match_all('{^(?:^|^(?:[ \t]*|(?:(?:[^*/]|(?<![^/])/(?![*/])|(?!/)\*)(\S))))\bvar_dump\(}m', $filecontent, $matches, PREG_SET_ORDER);
 		$failing_string = "";
-		//var_dump($matches);
 		foreach ($matches as $key => $val) {
-			if ($val[1] != '/' && $val[1] != '*') {
+			if (!isset($val[1]) || $val[1] != '/' && $val[1] != '*') {
 				$ok = false;
 				$failing_string = $val[0];
 				break;
 			}
 		}
 		$this->assertTrue($ok, "Found string var_dump that is not just after /* or // in '$filename': $failing_string");
+	}
+
+
+	/**
+	 * Provide test data for testing the method detecting var_dump presence.
+	 *
+	 * @return array<string,array{0:string,1:bool}> Test sets
+	 */
+	public function vardumpTesterProvider()
+	{
+		return [
+			 'var_dump at start of file' => ["var_dump(\$help)\n", true],
+			 'var_dump at start of line' => ["\nvar_dump(\$help)\n", true],
+			 'var_dump after comment next line' => ["/* Hello */\nvar_dump(\$help)\n", true],
+			 'var_dump with space' => [" var_dump(\$help)\n", true],
+			 'var_dump after comment' => [" // var_dump(\$help)\n", false],
+			 '2 var_dumps after comment' => [" // var_dump(\$help); var_dump(\$help)\n", false],
+			 'var_dump before and after comment' => [" var_dump(\$help); // var_dump(\$help)\n", true],
+		];
+	}
+
+	/**
+	 * Test that verifyNoActiveVardump generates a notification
+	 *
+	 * @param string $filecontent Fake file content
+	 * @param bool   $hasVardump  When true, expect var_dump detection
+	 *
+	 * @return void
+	 *
+	 * @dataProvider vardumpTesterProvider
+	 */
+	public function testVerifyNoActiveVardump(&$filecontent, $hasVardump)
+	{
+		$this->nbLinesToShow = 1;
+		// Create some dummy file content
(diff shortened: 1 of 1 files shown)


mdeweerd:
I updated the test to add the full path and verified in the Windows CI workflow where the test reports:

```plaintext
2) CodingPhpTest::testPHP
Found string var_dump that is not just after /* or // in includes/nusoap/lib/nusoap.php(D:\a\dolibarr\dolibarr\htdocs/includes/nusoap/lib/nusoap.php)
```

So this confirms the backslaches in the path. nusoap.php is inside the includes path and would have been excluded if htdocs were preceeded with a slash, but it isn't.

This can be considered a bug in the files.lib.php as the method listing all the files is supposed to be returning only paths with slashes.
This can be fixed in this test by replacing the relevant fields using `$var[$key]=strtr($var[$key], DIRECTORY_SEPARATOR, '/');`

The shortterm quick approach is to add the workaround in the test.
To update dol_dir_list, tests need to be added to demonstrate the issues, then fix them, and agree on the other changes as well.

(win-ci does not fail here because non of the workaround for the $_SERVER['PHP_SELF'] issue is implemented so no real test is running.

eldy:
> CodingPhpTest



> I updated the test to add the full path and verified in the Windows CI workflow where the test reports:
> 
> ```
> 2) CodingPhpTest::testPHP
> Found string var_dump that is not just after /* or // in includes/nusoap/lib/nusoap.php(D:\a\dolibarr\dolibarr\htdocs/includes/nusoap/lib/nusoap.php)
> Failed asserting that false is true.
> ```
> 
> So this confirms the backslaches in the path. nusoap.php is inside the includes path and would have been excluded if htdocs were preceeded with a slash, but it isn't.
> 
> This can be considered a bug in the files.lib.php as the method listing all the files is supposed to be returning only paths with slashes. This can be fixed in this test by replacing the relevant fields using `$var[$key]=strtr($var[$key], DIRECTORY_SEPARATOR, '/');`
> 
> The shortterm quick approach is to add the workaround in the test. To update dol_dir_list, tests need to be added to demonstrate the issues, then fix them, and agree on the other changes as well.
> 
> (win-ci does not fail here because non of the workaround for the $_SERVER['PHP_SELF'] issue is implemented so no real test is running.

This test should have been excluded with the `if (preg_match('/\/(htdocs|html)\/includes\//', $file['fullname'])) {`
What value do you have for $file['fullname'] ?

mdeweerd:
> This test should have been excluded with the `if (preg_match('/\/(htdocs|html)\/includes\//', $file['fullname'])) {`
> What value do you have for $file['fullname'] ?

It has something like '\htdocs/restofpath'  - the match expects '/htdocs/restofpath'.
dol_dir_list does not fix "$path".

eldy:
May be we can fix this in the phpunit code to exclude path with both / or \
preg_match('/ [ \ / \ \ ] htdocs ..

mdeweerd:
> May be we can fix this in the phpunit code to exclude path with both / or preg_match('/[/\]htdocs ..

I did not fix it in the test because I think it is better to fix dol_dir_list .  dol_dir_list should return paths with '/' as the separator IMHO.

That is the main raison why this is in draft.

mdeweerd:
I propose to integrate this PR as a fix of the var_dump checker that did not work for a case like this:

```php
    /* comment
     */
    var_dump($var);
```

The file separator issue is different and can be handled with the files.lib.php improvements.

test/phpunit/CodingPhpTest.php
```diff
@@ -188,10 +187,10 @@ public function testPHP()
 						break;
 					}
 					//print __METHOD__." Result for checking we don't have non escaped string in sql requests for file ".$file."\n";
-					$this->assertTrue($ok, 'Found string GETPOST into a .class.php file in '.$file['relativename'].'.');
+					//					$this->assertTrue($ok, 'Found string GETPOST in a .class.php file in '.$file['relativename'].'.');
```
eldy, line 190: Why disabling this assert  ?

mdeweerd: I probably disabled it during debug and forgot to remove the comment.  Restored it.

eldy:
We have now a test for var_dump into the pre-commit. So may be we can comment completely the test into the phpunit ?

mdeweerd:
> We have now a test for var_dump into the pre-commit. So may be we can comment completely the test into the phpunit ?

I am not aware of a test for var_dump in the pre-commit configuration.  A hook exists (but is not defined for dolibarr) to check for debug statements in python code.  But I do not believe it exists for php code.

I think several checks that are done in CodingPhpTest would be a good job for a plugin for Phan, with var_dump being one of them - possibly one could start from the deprecated functions checker to build a new plugin.

mdeweerd:
As the CodingPhpTest does not just fail on the first failing file, we now see 69 cases where var_dump is preceded with a ';' and probably preceeded with a '// ... ;'.  I'll update the check to avoid updating 69 occurences.

mdeweerd:
Reduced to 7 cases that are in multiline /* ... */ comments.

mdeweerd:
I added a function to remove the php comments (/* */ and //, not #) from the loaded source before submitting it to the tests.
This removes the var_dump statements that are in comments from the source directly.
This also helps with the other noticies- e.g. some of the isModEnabled() statements are commented and should not result in any notice anymore.

PR #28228, opened 2024-02-16
Qual: Refactor for distinction between encodings (STEP6)

mdeweerd:
# Qual: Refactor for distinction between encodings

Renamed path/file variable to equivalents with prefix os_ and utf8.
Also added $utf8_fullpath to optimize.

### DIFF
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -148,7 +148,7 @@ function dol_dir_list($path, $types = "all", $recursive = 0, $filter = "", $excl
 				}
 				// Check if file is qualified
 				foreach ($excludefilterarray as $filt) {
-					if (preg_match('/'.$filt.'/i', $file) || preg_match('/'.$filt.'/i', $fullpathfile)) {
+					if (preg_match('{'.$filt.'}i', $file) || preg_match('{'.$filt.'}i', $fullpathfile)) {
 						$qualified = 0;
 						break;
 					}
@@ -171,7 +171,7 @@ function dol_dir_list($path, $types = "all", $recursive = 0, $filter = "", $excl
 								$fileperm = dol_fileperm($path."/".$file);
 							}
 
-							if (!$filter || preg_match('/'.$filter.'/i', $file)) {	// We do not search key $filter into all $path, only into $file part
+							if (!$filter || preg_match('{'.$filter.'}i', $file)) {	// We do not search key $filter into all $path, only into $file part
 								$reg = array();
 								preg_match('/([^\/]+)\/[^\/]+$/', $path.'/'.$file, $reg);
 								$level1name = (isset($reg[1]) ? $reg[1] : '');
@@ -205,7 +205,7 @@ function dol_dir_list($path, $types = "all", $recursive = 0, $filter = "", $excl
 							$filesize = dol_filesize($path."/".$file);
 						}
 
-						if (!$filter || preg_match('/'.$filter.'/i', $file)) {	// We do not search key $filter into $path, only into $file
+						if (!$filter || preg_match('{'.$filter.'}i', $file)) {	// We do not search key $filter into $path, only into $file
 							if (empty($nbsecondsold) || $filedate <= ($now - $nbsecondsold)) {
 								preg_match('/([^\/]+)\/[^\/]+$/', $path.'/'.$file, $reg);
 								$level1name = (isset($reg[1]) ? $reg[1] : '');
@@ -3494,7 +3494,7 @@ function getFilesUpdated(&$file_list, SimpleXMLElement $dir, $path = '', $pathre
 		$expectedsize = (empty($file['size']) ? '' : $file['size']);
 		$expectedmd5 = (string) $file;
 
-		//if (preg_match('#'.$exclude.'#', $filename)) continue;
+		//if (preg_match('{'.$exclude.'}', $filename)) continue;
 
 		if (!file_exists($pathref.'/'.$filename)) {
 			$file_list['missing'][] = array('filename' => $filename, 'expectedmd5' => $expectedmd5, 'expectedsize' => $expectedsize);
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -521,20 +521,17 @@ function dol_is_link($pathoffile)
 }
 
 /**
- * Return if path is an URL
+ * Return if path is an URI (the name of the method is misleading).
  *
- * @param   string		$url	Url
- * @return  boolean      	   	True or false
+ * URLs are addresses for websites, URI refer to online resources.
+ *
+ * @param   string		$uri	URI to test
+ * @return  boolean      	   	True if the path looks like a URI, else false.
  */
-function dol_is_url($url)
+function dol_is_url($uri)
 {
-	$tmpprot = array('file', 'http', 'https', 'ftp', 'zlib', 'data', 'ssh', 'ssh2', 'ogg', 'expect');
-	foreach ($tmpprot as $prot) {
-		if (preg_match('/^'.$prot.':/i', $url)) {
-			return true;
-		}
-	}
-	return false;
+	$prots = array('file', 'http', 'https', 'ftp', 'zlib', 'data', 'ssh', 'ssh2', 'ogg', 'expect');
+	return false !== preg_match('/^('.implode('|', $prots).'):/i', $uri);
 }
 
 /**
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -540,13 +540,13 @@ function dol_is_url($uri)
  * 	@param	string	$folder		Name of folder
  * 	@return boolean				True if dir is empty or non-existing, False if it contains files
  */
-function dol_dir_is_emtpy($folder)
+function dol_dir_is_empty($folder)
 {
 	$newfolder = dol_osencode($folder);
 	if (is_dir($newfolder)) {
 		$handle = opendir($newfolder);
 		$folder_content = '';
-		$name_array = [];
+		$name_array = array();
 		while ((gettype($name = readdir($handle)) != "boolean")) {
 			$name_array[] = $name;
 		}
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -131,14 +131,18 @@ function dol_dir_list($utf8_path, $types = "all", $recursive = 0, $filter = "", 
 			$filedate = '';
 			$filesize = '';
 			$fileperm = '';
-			while (false !== ($utf8_file = readdir($dir))) {        // $utf8_file is always a basename (in directory $os_path)
-				if (!utf8_check($utf8_file)) {
-					$utf8_file = mb_convert_encoding($utf8_file, 'UTF-8', 'ISO-8859-1'); // Make sure data is stored in utf8 in memory
+			while (false !== ($os_file = readdir($dir))) {        // $utf8_file is always a basename (in directory $os_path)
+				$os_fullpathfile = ($os_path ? $os_path.'/' : '').$os_file;
+
+				if (!utf8_check($os_file)) {
+					$utf8_file = mb_convert_encoding($os_file, 'UTF-8', 'ISO-8859-1'); // Make sure data is stored in utf8 in memory
+				} else {
+					$utf8_file = $os_file;
 				}
-				$mixed_fullpathfile = ($os_path ? $os_path.'/' : '').$utf8_file;  // TODO: Suspicious! Combining system encoding with UTF-8 encoding ...
 
 				$qualified = 1;
 
+				$utf8_fullpathfile = "$utf8_path/$utf8_file";  // Temp variable for speed
 				// Define excludefilterarray
 				$excludefilterarray = array('^\.');
 				if (is_array($excludefilter)) {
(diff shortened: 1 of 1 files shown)


mdeweerd:
@eldy Do not merge yet.
Your opinion?

According to me the following line needs to be moved before the conversion of file to utf8:
```php
$fullpathfile = ($newpath ? $newpath.'/' : '').$file;
```

here:

https://github.com/Dolibarr/dolibarr/blob/df1c989491bcaf2b9415c8fa3039732a9feae126/htdocs/core/lib/files.lib.php#L115-L119

As is, the os_encoded path is mixed with an utf8 file...

Do I move it up ?

EDIT: Also see code that is mentioned in review annotation where IMHO the test should be done against $utf8_fullpath (I renamed the variables to make it easier to verify and modify).

(Note: I am analysing this method because it does not seem to return all the files in the CodingPhpTest)

htdocs/core/lib/files.lib.php
```diff
@@ -129,7 +129,7 @@ function dol_dir_list($path, $types = "all", $recursive = 0, $filter = "", $excl
 				}
 				// Check if file is qualified
 				foreach ($excludefilterarray as $filt) {
-					if (preg_match('/'.$filt.'/i', $file) || preg_match('/'.$filt.'/i', $fullpathfile)) {
+					if (preg_match('{'.$filt.'}i', $file) || preg_match('{'.$filt.'}i', $fullpathfile)) {
```
eldy, line 132: Everywhere in dolibarr, we are using the / as regex start-stop operator.
We should keep this or develloper will not know how to escape the start stop operator if it is not always the same.

htdocs/core/lib/files.lib.php
```diff
@@ -129,39 +129,40 @@ function dol_dir_list($path, $types = "all", $recursive = 0, $filter = "", $excl
 				}
 				// Check if file is qualified
 				foreach ($excludefilterarray as $filt) {
-					if (preg_match('/'.$filt.'/i', $file) || preg_match('/'.$filt.'/i', $fullpathfile)) {
+					if (preg_match('{'.$filt.'}i', $utf8_file) || preg_match('{'.$filt.'}i', $mixed_fullpathfile)) {
```
mdeweerd, line 132: The renamed variable makes it clear that the matching is not ok here / on the left side with match the regex with an utf-8 encoded filed, and on the write the directory is in native encoding while the basename is in utf8.

htdocs/core/lib/files.lib.php
mdeweerd: I already showed in another commit that this is not true, there are locations where other delimiters are used.

Escaping the '/' still works, but it is more secure this way.
The variable is provided by the caller, nothing was specified about the delimiters and the caller of the function should not have to worry about that.
So if you insist that the code uses '/' as the delimiters, then the code should escape the '/' delimiters in place of the caller.  However, as the caller is supposedly already doing so for the old code, it is harder (or impossible) to distinguish between the arguments that are already escaped for this and which ones are not.

Anywhere where the delimiters are enclosing "external" variables the "{}" are a good choice and "[]" is another possible choice.  In a regex, "{}" and "[]" would have to be paired or escaped, but maybe less so the '{}' tokens.

I'll perform a regex on the code to see if there are other locations where external variables are inserted like that. 
It's quite dangerous to use the '//' delimiters as that may accidentally introduce a regex modifier that the developer did not think about - {} or [] avoids that.

mdeweerd: These are the other locations that I found where a variable is used.  Some may be for 'internal' variables, others for external ones - all could be change to use the same more protective method.

```php
./htdocs/accountancy/class/accountancyexport.class.php:                 $tab[] = preg_match('/'.$separator.'/', $line->label_operation) ? "'".$line->label_operation."'" : $line->label_operation;
./htdocs/admin/translation.php:                 if ($transkey && !preg_match('/'.$regexstring.'/i', $key)) {
./htdocs/categories/class/categorie.class.php:                          $test = (preg_match('/'.$keyfilter1.'/', $val['fullpath']) || preg_match('/'.$keyfilter2.'/', $val['fullpath'])
./htdocs/categories/class/categorie.class.php:                                  || preg_match('/'.$keyfilter3.'/', $val['fullpath']) || preg_match('/'.$keyfilter4.'/', $val['fullpath']));
./htdocs/compta/bank/bankentries_list.php:              if (preg_match('/'.$search_description.'/i', $translatedlabel)) {
./htdocs/core/ajax/ajaxdirtree.php:             } elseif (preg_match('/'.$val['fullpath'].'_/i', $fullpathselected.'_')) {
./htdocs/core/class/html.formfile.class.php:                            if ($filter && !preg_match('/'.$filter.'/i', $file["name"])) {
./htdocs/core/class/lessc.class.php:                                    if (preg_match('/'.$min[0].'.*?(?<!\\\\)'.$min[0].'/', $text, $m, 0, $count)) {
./htdocs/core/class/utils.class.php:                                    preg_match('/'.$filter.'/i', $gzfile['name'], $tabmatches);
./htdocs/core/filemanagerdol/connectors/php/connector.lib.php:  while (preg_match('/'.$regex.'/S', $string, $matches)) {
./htdocs/core/lib/files.lib.php:                                        if (preg_match('/'.$filt.'/i', $file) || preg_match('/'.$filt.'/i', $fullpathfile)) {
./htdocs/core/lib/files.lib.php:                                                        if (!$filter || preg_match('/'.$filter.'/i', $file)) {     // We do not search key $filter into all $path, only into $file part
./htdocs/core/lib/files.lib.php:                                                if (!$filter || preg_match('/'.$filter.'/i', $file)) {     // We do not search key $filter into $path, only into $file
./htdocs/core/lib/files.lib.php:                //if (preg_match('#'.$exclude.'#', $filename)) continue;
./htdocs/core/lib/functions2.lib.php:   if (preg_match('/'.$urlregex.'/i', $url)) {
./htdocs/core/lib/website2.lib.php:             if (preg_match('/'.$forbiddenphpcommand.'\s*\(/ms', $phpfullcodestring)) {
./htdocs/core/lib/website2.lib.php:             if (preg_match('/'.$forbiddenphpchar.'/ms', $phpfullcodestring)) {
./htdocs/core/modules/asset/mod_asset_standard.php:             if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/bom/mod_bom_standard.php:         if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/cheque/mod_chequereceipt_mint.php:                if ($payyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $payyymm)) {
./htdocs/core/modules/commande/mod_commande_marbre.php:         if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/contract/mod_contract_serpis.php:         if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/delivery/mod_delivery_jade.php:           if ($fayymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $fayymm)) {
./htdocs/core/modules/expedition/mod_expedition_safor.php:              if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/expensereport/mod_expensereport_jade.php:         if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/export/exportcsv.class.php:               if (preg_match('/'.$this->separator.'/', $newvalue)) {
./htdocs/core/modules/export/export_tsv.modules.php:            if (preg_match('/'.$this->separator.'/', $newvalue)) {
./htdocs/core/modules/facture/mod_facture_mars.php:             if ($fayymm && !preg_match('/'.$this->prefixinvoice.'[0-9][0-9][0-9][0-9]/i', $fayymm)) {
./htdocs/core/modules/facture/mod_facture_mars.php:             if ($fayymm && !preg_match('/'.$this->prefixcreditnote.'[0-9][0-9][0-9][0-9]/i', $fayymm)) {
./htdocs/core/modules/facture/mod_facture_terre.php:            if ($fayymm && !preg_match('/'.$this->prefixinvoice.'[0-9][0-9][0-9][0-9]/i', $fayymm)) {
./htdocs/core/modules/facture/mod_facture_terre.php:            if ($fayymm && !preg_match('/'.$this->prefixcreditnote.'[0-9][0-9][0-9][0-9]/i', $fayymm)) {
./htdocs/core/modules/facture/mod_facture_terre.php:            if ($fayymm && !preg_match('/'.$this->prefixdeposit.'[0-9][0-9][0-9][0-9]/i', $fayymm)) {
./htdocs/core/modules/fichinter/mod_pacific.php:                if (!$fayymm || preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $fayymm)) {
./htdocs/core/modules/holiday/mod_holiday_madonna.php:          if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/hrm/mod_evaluation_standard.php:          if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/import/import_csv.modules.php:                                              } elseif (!preg_match('/'.$objimport->array_import_regex[0][$val].'/i', $newval)) {
./htdocs/core/modules/member/mod_member_advanced.php:           if (!$coyymm || preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/mrp/mod_mo_standard.php:          if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/payment/mod_payment_cicada.php:           if ($payyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $payyymm)) {
./htdocs/core/modules/product_batch/mod_lot_standard.php:               if ($max && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $max)) {
./htdocs/core/modules/product_batch/mod_sn_standard.php:                if ($max && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $max)) {
./htdocs/core/modules/project/mod_project_simple.php:           if (!$coyymm || preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/project/task/mod_task_simple.php:         if (!$coyymm || preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/propale/mod_propale_marbre.php:           if (!$pryymm || preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $pryymm)) {
./htdocs/core/modules/reception/mod_reception_beryl.php:                if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/stocktransfer/mod_stocktransfer_standard.php:             if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/supplier_invoice/mod_facture_fournisseur_cactus.php:              if ($siyymm && !preg_match('/'.$this->prefixinvoice.'[0-9][0-9][0-9][0-9]/i', $siyymm)) {
./htdocs/core/modules/supplier_invoice/mod_facture_fournisseur_cactus.php:              if ($siyymm && !preg_match('/'.$this->prefixcreditnote.'[0-9][0-9][0-9][0-9]/i', $siyymm)) {
./htdocs/core/modules/supplier_invoice/mod_facture_fournisseur_cactus.php:              if ($siyymm && !preg_match('/'.$this->prefixdeposit.'[0-9][0-9][0-9][0-9]/i', $siyymm)) {
./htdocs/core/modules/supplier_order/mod_commande_fournisseur_muguet.php:               if (!$coyymm || preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/supplier_payment/mod_supplier_payment_bronan.php:         if ($payyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $payyymm)) {
./htdocs/core/modules/supplier_proposal/mod_supplier_proposal_marbre.php:               if (!$pryymm || preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $pryymm)) {
./htdocs/core/modules/takepos/mod_takepos_ref_simple.php:               if (!$pryymm || preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $pryymm)) {
./htdocs/core/modules/ticket/mod_ticket_simple.php:             if (!$coyymm || preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/core/modules/workstation/mod_workstation_standard.php:         /*if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/emailcollector/class/emailcollector.class.php:                                         if (preg_match('/'.$regexstring.'/'.$regexoptions, $sourcestring, $regforval)) {
./htdocs/emailcollector/class/emailcollector.class.php:                                           if (preg_match('/'.$regexstring.'/ms', $sourcestring, $regforval)) {
./htdocs/includes/tecnickcom/tcpdf/include/tcpdf_static.php:                                      } elseif (preg_match('/'.$val.'[\-]{1}/i', $dom[$key]['attribute'][$att]) > 0) {
./htdocs/install/upgrade.php:                   if (preg_match('/'.$from.'\-/i', $file)) {
./htdocs/knowledgemanagement/core/modules/knowledgemanagement/mod_knowledgerecord_standard.php:   if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/modulebuilder/template/core/modules/mymodule/mod_myobject_standard.php:                if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/partnership/core/modules/partnership/mod_partnership_standard.php:             if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/recruitment/core/modules/recruitment/mod_recruitmentcandidature_standard.php:          if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/recruitment/core/modules/recruitment/mod_recruitmentjobposition_standard.php:          if ($coyymm && !preg_match('/'.$this->prefix.'[0-9][0-9][0-9][0-9]/i', $coyymm)) {
./htdocs/user/class/user.class.php:                             if (preg_match('/'.$keyfilter1.'/', $val['fullpath']) || preg_match('/'.$keyfilter2.'/', $val['fullpath'])
./htdocs/user/class/user.class.php:                                     || preg_match('/'.$keyfilter3.'/', $val['fullpath']) || preg_match('/'.$keyfilter4.'/', $val['fullpath'])) {
```

mdeweerd:
As a reminder, an update would be needed for the following as well:

```plaintext
2) CodingPhpTest::testPHP
Found string var_dump that is not just after /* or // in includes/nusoap/lib/nusoap.php(D:\a\dolibarr\dolibarr\htdocs/includes/nusoap/lib/nusoap.php)
```
(See https://github.com/Dolibarr/dolibarr/pull/28226#issuecomment-1949624405).

eldy:
> @eldy Do not merge yet.
> Your opinion?
> 
> According to me the following line needs to be moved before the conversion of file to utf8:
> ```php
> $fullpathfile = ($newpath ? $newpath.'/' : '').$file;
> ```
> 
> here:
> 
> https://github.com/Dolibarr/dolibarr/blob/df1c989491bcaf2b9415c8fa3039732a9feae126/htdocs/core/lib/files.lib.php#L115-L119
> 
> As is, the os_encoded path is mixed with an utf8 file...
> 
> Do I move it up ?
> 
> EDIT: Also see code that is mentioned in review annotation where IMHO the test should be done against $utf8_fullpath (I renamed the variables to make it easier to verify and modify).
> 
> (Note: I am analysing this method because it does not seem to return all the files in the CodingPhpTest)

I agree, the concat must be done before the utf8 conversion so we concat iso with iso or utf with utf.
But i think we should also force the convert into utf8 just after for $fullpath, like we do for $file, because goal is to use $fullpath and $file to make regex test. And we need to have var in memory always in utf8 for this.

mdeweerd:
> force the convert into utf8 just after for $fullpath, like we do for $file, because goal is to use $fullpath and $file to make regex test.

The conversion is already in place, but the wrong variables are used in the current code - and the regex is compared against the mixed encoding filename, so that would need to be the utf8 filename.

https://github.com/Dolibarr/dolibarr/blob/1c4b799d76692f9ce5073a5fc2e7c6f4e327b592/htdocs/core/lib/files.lib.php#L115-L139

I'll wait for the integration of the unescaped forward slash checks.

eldy:
Sorry i am completely lost with this PR. It contains several topic in the same:
- renaming vars to better see if var is utf or not
- changing regexp separator
- moving concat of variables to avoid mixing utf8 and not utf8. Can you split it in 3 ?

mdeweerd:
> Sorry i am completely lost with this PR. It contains several topic in the same:
> 
> * renaming vars to better see if var is utf or not
> * changing regexp separator
> * moving concat of variables to avoid mixing utf8 and not utf8. Can you split it in 3 ?

You can go through the commits individually, I tried to do them in order.  If it's not clear enough by going through the commits (there is a github function for going from one to the next), I'll cherry-pick them.

mdeweerd:
I split this up in 6 PRs (5 extra PRs) that need to be done in order to be certain to avoid conflicts.

This PR becaome STEP6.

The PRs can be rebase one after the other to see the changes of each PR.

mdeweerd:
I already had to rebase all 6 steps because the other PR with the change in files.lib.php was merged in the meanwhile.

I added another commit on top of this which is the move of the statement to fix the mixing of the encodings + I then reuse the non-utf8 strings and avoid a os_encode step.

eldy:
I close this PR. Splitted into several other.

mdeweerd:
@eldy 6his PR contained STEP nbr 6 !

PR #28236, opened 2024-02-17
Qual: Improve rendering of phpstan technical debt table in apstats.

mdeweerd:
# Qual: Improve rendering of phpstan technical debt table in apstats.

Apply the DataTable rendering to the phpstan technical debt (same as applied to phan).

Note: there are several issues with apstats - for instance, the issue descriptions are not html encoded, there are duplicate ids, etc.

Checkout the [W3 validation for apstats](https://validator.w3.org/nu/?doc=https%3A%2F%2Fdoxygen.dolibarr.org%2Fapstats%2F).


mdeweerd:
The  [W3 validation for apstats](https://validator.w3.org/nu/?doc=https%3A%2F%2Fdoxygen.dolibarr.org%2Fapstats%2F). is much cleaner now and the phpstan table is searchable 😉

PR #28249, opened 2024-02-18
Improve file preview

SylvainLegrand:
Better user experience with long file name!

improve file preview - before

improve file preview - after


eldy:
On which url do you have this. On v19 and develop, i have the picto that is correctly in its own column
image

PR #28253, opened 2024-02-18
New Add alias field to socpeople table

Daoud-mohamed:
# Add alias field to socpeople table
In certain cases, we need an alias, first name, and last name for addresses.


htdocs/install/mysql/migration/19.0.0-20.0.0.sql
```diff
@@ -235,3 +235,5 @@ ALTER TABLE llx_socpeople ADD COLUMN geolat double(24,8) DEFAULT NULL;
 ALTER TABLE llx_socpeople ADD COLUMN geolong double(24,8) DEFAULT NULL;
 ALTER TABLE llx_socpeople ADD COLUMN geopoint point DEFAULT NULL;
 ALTER TABLE llx_socpeople ADD COLUMN georesultcode varchar(16) NULL;
+
+ALTER TABLE llx_socpeople ADD COLUMN alias varchar(255) NULL;
```
eldy, line 239: To match the name used into llx_thirdparty table, can you rename the field into "name_alias" instead of "alias"

Daoud-mohamed: Ok, done.

PR #28260, opened 2024-02-18
Fix: Use pattern delimiter that is less likely to interfere.(STEP3)

mdeweerd:
# Fix: Use pattern delimiter that is less likely to interfere.

Forward slashes are likely to interfere with the provided pattern, so using {}
delimiters for externally provided patterns is safer.

### DIFF
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -152,7 +152,7 @@ function dol_dir_list($utf8_path, $types = "all", $recursive = 0, $filter = "", 
 
 				// Check if file is qualified
 				foreach ($excludefilterarray as $filt) {
-					if (preg_match('/'.$filt.'/i', $utf8_file) || preg_match('/'.$filt.'/i', $utf8_fullpathfile)) {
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/files.lib.php
```diff
@@ -152,7 +152,7 @@ function dol_dir_list($utf8_path, $types = "all", $recursive = 0, $filter = "",
 
 				// Check if file is qualified
 				foreach ($excludefilterarray as $filt) {
-					if (preg_match('/'.$filt.'/i', $utf8_file) || preg_match('/'.$filt.'/i', $utf8_fullpathfile)) {
+					if (preg_match('{'.$filt.'}i', $utf8_file) || preg_match('{'.$filt.'}i', $utf8_fullpathfile)) {
```
eldy, line 155: This will break compatibility with current format of the parameter $excludefilter that expect to have string escaped for "/" regex separator. It is commonly used this way by a lot of external modules (and may be dolibarr itself), so i prefer to cancel PR. We have now a protection that may be enough.

PR #28261, opened 2024-02-18
Qual: Optimize dol_is_url & improve doc (STEP4)

mdeweerd:
# Qual: Optimize dol_is_url & improve doc

A single regex match will be faster.  Explained that it's a URI tester.


mdeweerd:
@eldy ??

Strange, the system indicates this as 'closed' but it's merged.

PR #28264, opened 2024-02-19
Fix dol_dir_move for windows platform

mdeweerd:
# Fix dol_dir_move for windows platform

'rename' works differently on linux and windows.  This fixes that and the test result:
1) FilesLibTest::testDolDirMove
move of directory on existing directory with empty directory Failed asserting that false is true.

### DIFF
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -1129,6 +1129,100 @@ function dol_move($srcfile, $destfile, $newmask = '0', $overwriteifexists = 1, $
 	return $result;
 }
 
+
+/**
+ * Rename a file or a name, with overwrite.
+ *
+ * The PHP rename method works differently between Linux and Windows platforms.
+ * This method aims to match the linux implementation
+ *
+ * @param	string  $oldname            Source file or directory
+ * @param   string	$newname            Destination file or directory
+ * @return  bool    true if success
+ */
+function dol_rename($oldname, $newname)
+{
+	if (strtoupper(substr(PHP_OS, 0, 3)) === 'WIN') {
+		// On Windows, use custom function to rename directories
+		if (is_dir($oldname)) {
+			return dol_win_dir_rename($oldname, $newname);
+		}
+	}
+	// By default use the rename method directly
+	return (bool) @rename($oldname, $newname);
+}
+
+
+/**
+ * Rename a directory on windows platform similar to Linux
+ *
+ * On Linux a rename to an empty directory works, but not on windows.
+ * This is a replacement for rename to achieve the same behavior on windows.
+ *
+ * @param	string  $oldname            Source directory
+ * @param   string	$newname            Destination directory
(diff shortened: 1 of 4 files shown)


htdocs/core/lib/files.lib.php
```diff
@@ -1122,6 +1122,117 @@ function dol_move($srcfile, $destfile, $newmask = '0', $overwriteifexists = 1, $
...
+
+
+/**
+ * Rename a directory on windows platform
+ *
+ * @param	string  $oldname            Source directory
+ * @param   string	$newname            Destination directory
+ * @return  bool    true if success
+ */
+function dol_win_dir_rename($oldname, $newname)
```
eldy, line 1159: The rename on windows should work like on linux. If you need a copy and delete, it probably means you make the move on a different file system ?

mdeweerd: The PHPUNIT test tries to do it on the same file system (in the documents directory) and that fails.

You can check on [php.net's rename documentation page](https://www.php.net/manual/en/function.rename.php) that the behavior is not the same for all systems.

eldy: copy and delete is a different operation than rename. If file is large, difference is very important.
I do not experience trouble on windows with rename, both on FAT or NTFS. So we should have no reason to have a special rename function for windows.
May be we can add a test inside dol_move_dir(() to be sure to return the same result. My understanding is that it works the same, but return code differs ?

mdeweerd: The function '@rename' fails on a windows in a testcase where the destination directory is empty - the result on linux is different.  The function can be updated to move the files rather than copy/delete them.

eldy: It means you can't move a file called /aaa/filename  into a directory /bbb if /bbb is empty on windows ?

mdeweerd:
On Windows, `dol_rename` now matches the behavior on \*nix.

This is validated by phpunit tests.

The following test now passes on Windows:
```
1) FilesLibTest::testDolDirMove
move of directory on existing directory with empty directory

D:\a\dolibarr\dolibarr\test\phpunit\FilesLibTest.php:516
```

The only remaining failing tests on the windows platform are related to open_basedir restrictions - and the ScriptTests that are fixed in #28565 .

mdeweerd:
@eldy This fix and #28598 are the only remaining PR's to fix 'windows specific' issues.
Integrating those will no longer require that the windows-ci failures are ignored.
That in turn will allow earlier identification of phpunit run issues (when the travis queue is filled) and provide Github Annotations for the failing phpunit tests.

I do not know what is still blocking - 'dol_rename' is a replacement for 'rename' which behaves as it rename does on a unix system - 'rename' itself behaved differently on windows and unix.  The test case for dol_rename demonstrates that dol_rename behaves the same (because it passes on windows and on travis).

PR #28291, opened 2024-02-19
qual: phpstan for htdocs/salaries/class/paymentsalary.class.php

thibdrev:
htdocs/salaries/class/paymentsalary.class.php	205	Property PaymentSalary::$amount (float) does not accept string.

htdocs/salaries/class/paymentsalary.class.php	292	Property PaymentSalary::$amount (float) does not accept string.

htdocs/salaries/class/paymentsalary.class.php	389	Property PaymentSalary::$amount (float) does not accept string.

htdocs/salaries/class/paymentsalary.class.php	565	Property PaymentSalary::$datec (int) does not accept string.

### DIFF
--- a/htdocs/salaries/class/paymentsalary.class.php
+++ b/htdocs/salaries/class/paymentsalary.class.php
@@ -202,7 +202,7 @@ class PaymentSalary extends CommonObject
 			$this->fk_salary = (int) $this->fk_salary;
 		}
 		if (isset($this->amount)) {
-			$this->amount = trim($this->amount);
+			$this->amount = (float) $this->amount;
 		}
 		if (isset($this->fk_typepayment)) {
 			$this->fk_typepayment = (int) $this->fk_typepayment;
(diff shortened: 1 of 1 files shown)


htdocs/salaries/class/paymentsalary.class.php
```diff
@@ -228,11 +228,10 @@ public function create($user, $closepaidcontrib = 0)
 
 		$totalamount = 0;
 		foreach ($this->amounts as $key => $value) {  // How payment is dispatch
-			$newvalue = price2num($value, 'MT');
+			$newvalue = (float) price2num($value, 'MT');
 			$this->amounts[$key] = $newvalue;
 			$totalamount += $newvalue;
 		}
-		$totalamount = price2num($totalamount);
```
eldy, line 235: totalamount is result of += of a float value.
So result is a float. So may be a 0 may be stored as 0.0000000000001
This means the test just after 
`if ($totalamount == 0) 
`will fails when it should be true.
A price2num should solve this but it will work if there is a rounding param. So i suggest to add here
`$totalamount = price2num($totalamount, 'MT');`
instead of removing it completely...

PR #28292, opened 2024-02-19
Fix: Deprecation message for backupGlobalsBlackList

mdeweerd:
# Fix: Deprecation message for backupGlobalsBlackList

Fix the deprecation messages during the tests:

1) AdminLibTest::testVersionCompare
PHPUnit\Framework\TestCase::$backupGlobalsBlacklist is deprecated and will be removed in PHPUnit 10. Please use PHPUnit\Framework\TestCase::$backupGlobalsExcludesList instead.

### DIFF
--- a/test/phpunit/AdminLibTest.php
+++ b/test/phpunit/AdminLibTest.php
@@ -1,6 +1,7 @@
 <?php
 /* Copyright (C) 2010 Laurent Destailleur  <eldy@users.sourceforge.net>
  * Copyright (C) 2023 Alexandre Janniaux   <alexandre.janniaux@gmail.com>
+ * Copyright (C) 2024		MDW							<mdeweerd@users.noreply.github.com>
  *
  * This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


test/phpunit/AdminLibTest.php
```diff
@@ -48,7 +49,23 @@
  */
 class AdminLibTest extends CommonClassTest
 {
-	protected $backupGlobalsBlacklist = array('conf', 'user', 'langs', 'db');
```
eldy, line 51: I don't see reason of having this line of code.
May be we can remove it completely ?

mdeweerd: [I can only think that the author of the test ](https://docs.phpunit.de/en/9.6/fixtures.html?highlight=backupGlobalsExcludeList#global-state)

I think that excluding the variables from backup/restore implies that if the test modifies the values that their values are maintained.

This is the commit where it was introduced:
https://github.com/Dolibarr/dolibarr/commit/599aa6009a6

PR #28293, opened 2024-02-19
Fix: Deprecation message for assertRegExp

mdeweerd:
# Fix: Deprecation message for assertRegExp

Fix the deprecation messages during the tests:

assertRegExp() is deprecated and will be removed in PHPUnit 10. Refactor your code to use assertMatchesRegularExpression() instead.

### DIFF
--- a/test/phpunit/SecurityTest.php
+++ b/test/phpunit/SecurityTest.php
@@ -1,6 +1,7 @@
 <?php
 /* Copyright (C) 2010 Laurent Destailleur  <eldy@users.sourceforge.net>
  * Copyright (C) 2023 Alexandre Janniaux   <alexandre.janniaux@gmail.com>
+ * Copyright (C) 2024		MDW							<mdeweerd@users.noreply.github.com>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


test/phpunit/SecurityTest.php
```diff
@@ -839,7 +840,15 @@ public function testGetURLContent()
 		$url = 'ftp://mydomain.com';
 		$tmp = getURLContent($url);
 		print __METHOD__." url=".$url."\n";
-		$this->assertRegExp("/not supported/", $tmp['curl_error_msg'], "Should disable ftp connection");	// Test error if return does not contains 'not supported'
```
eldy, line 842: I think a better way to fix this is to make a
_$tmpvar = preg_match('/not supported/', ...)_
then
_this->assertTrue($tmpvar);_

We have so few use of assertRegExp that removing it completely seems better for future.

mdeweerd: That is not better IMHO as the message that is reported in case of failure does not report the regex it is trying to match, while using the phpunit methods for regex shows this.  So it's easier to debug.

Or the other method is to add a method in the Common*Test class that now exists and do these conversions there: implement assertRegex and call the appropriate parent method according to the version of PHPUNIT.
When I wrote this there was no Common*Test class.

PR #28294, opened 2024-02-19
Fix: Improve test configuration

mdeweerd:
# Fix: Improve test configuration

## Do not fail early when running tests (stopOnFailure->false)

Finishing the phpunits tests does not require a lot of extra time so rather than quiting early, run them all.

Also, when needed this option can be enabled on the CLI, but it can not be disabled.  So when desired, it can still be added in a ci-flow or local script.

## Improve reporting

We want to report Notices and Warnings and with enough detail to help during debug.  These options are explicitly activated.

### DIFF
--- a/test/phpunit/phpunittest.xml
+++ b/test/phpunit/phpunittest.xml
@@ -1,45 +1,53 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE phpunit SYSTEM "phpunit.dtd" >
 <phpunit
-	backupGlobals="false"
-	backupStaticAttributes="true"
-	convertErrorsToExceptions="true"
-	convertNoticesToExceptions="true"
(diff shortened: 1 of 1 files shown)


test/phpunit/phpunittest.xml
```diff
@@ -1,45 +1,53 @@
...
-	  </whitelist>
-	</filter>
+  backupGlobals="false"
+  backupStaticAttributes="true"
+  convertErrorsToExceptions="true"
+  convertNoticesToExceptions="true"
+  convertWarningsToExceptions="true"
+  colors="true"
+  processIsolation="false"
+  stopOnFailure="false">
```
eldy, line 11: As we have to pay for CPU used by travis, we don't want to continue if an error is found.
Knowing there is an error is enough to stop consumption.

PR #28295, opened 2024-02-19
Fix: CodingPHPTest - Fix var_dump checker + improvements

mdeweerd:
# Fix: Fix var_dump checker

The core issue was that in PHP whitespace includes newlines by default, the m modifier
is needed to not match multilines.

# Qual: Remove unnecessary exception

The exception 'class.nusoap_base.php' no longer exists.

Qual: For var_dump report, report also the full path
    
# Qual: For var_dump report, report also the full path

In win-ci var_dump was detected in a file that should normally be excluded.
So add the full part in the error message as well


mdeweerd:
Duplicate

PR #28297, opened 2024-02-19
Fix: Allow new & old names to check if module is loaded.

mdeweerd:
# Fix: Allow new & old names to check if module is loaded.

The method 'isModEnabled()' was already checking some modules for their new name, but after transitioning Conf this was not working for 'member'.  Now testing for both names as 'Conf' can handle both and deprecation will be reported.

Several issues were discovered after removing deprecated values.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -226,7 +226,7 @@ function isModEnabled($module)
 		'expedition' => 'delivery_note',
 		'facture' => 'invoice',
 		'projet' => 'project',
-		'propal' => 'propale',
+		'propale' => 'propal',
 	);
 
 	if (!getDolGlobalString('MAIN_USE_NEW_SUPPLIERMOD')) {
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -218,21 +218,27 @@ function isModEnabled($module)
...
 	}
+
+	$module_alt = $module;
 	if (!empty($arrayconv[$module])) {
-		$module = $arrayconv[$module];
+		$module_alt = $arrayconv[$module];
 	}
 
-	return !empty($conf->modules[$module]);
+	return !empty($conf->modules[$module]) || !empty($conf->modules[$module_alt]);
```
eldy, line 241: Imagine entry of function is 'delivery_note'.
It seems old code result will differ from new code result.

mdeweerd: OK, I can see the table is inconsistent.
I added conversions as needed (e.g. 'facture' is still used while 'invoice' is the actual module name), considered that the 'old' names must be mappable to the new names, but not the inverse - but he old entries map new to old (while the old names should no longer be the active ones in the code).

So I change the remaining mappings and if anything fails, then I guess the code needs updates elsewhere.

PR #28305, opened 2024-02-20
Fix tva_tx on deposit invoice

vmaury:
When we create a deposit and select "invoiceAvoirWithPaymentRestAmount" checkbox, the vat rate is always set to zero

This fix set the vat rate according to seller and buyers societys

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -1287,13 +1287,20 @@ if (empty($reshook)) {
 						$totaldeposits = $facture_source->getSumDepositsUsed();
 						$remain_to_pay = abs($facture_source->total_ttc - $totalpaid - $totalcreditnotes - $totaldeposits);
 
-						if ((empty($soc) || !is_object($soc) || get_class($soc) != 'Societe') && $socid >0) {
-							$soc = new Societe($db);
-							$soc->fetch($socid);
+						if (!empty($conf->global->INVOICE_VAT_TO_USE_ON_CREDIT_NOTE_WHEN_GENERATED_FROM_REMAIN_TO_PAY) && $conf->global->INVOICE_VAT_TO_USE_ON_CREDIT_NOTE_WHEN_GENERATED_FROM_REMAIN_TO_PAY == 'default') {
+							if ((empty($object->thirdparty) || !is_object($object->thirdparty) || get_class($object->thirdparty) != 'Societe')) {
+								$object->fetch_thirdparty();
+							}
+							if (!empty($object->thirdparty) && is_object($object->thirdparty) && get_class($object->thirdparty) == 'Societe') {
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/card.php
```diff
@@ -1287,7 +1287,15 @@
 						$totaldeposits = $facture_source->getSumDepositsUsed();
 						$remain_to_pay = abs($facture_source->total_ttc - $totalpaid - $totalcreditnotes - $totaldeposits);
 
-						$object->addline($langs->trans('invoiceAvoirLineWithPaymentRestAmount'), $remain_to_pay, 1, 0, 0, 0, 0, 0, '', '', 'TTC');
+						if ((empty($soc) || !is_object($soc) || get_class($soc) != 'Societe') && $socid >0) {
+							$soc = new Societe($db);
+							$soc->fetch($socid);
```
eldy, line 1292: Don't you already have the thirdparty into $object->thirdparty ?

htdocs/compta/facture/card.php
```diff
@@ -1287,7 +1287,15 @@
...
+						}
+						if (!empty($soc) && is_object($soc) && get_class($soc) == 'Societe') {
+							$tva_tx = get_default_tva($mysoc, $soc);
+						} else $tva_tx = 0;
+
+						$object->addline($langs->trans('invoiceAvoirLineWithPaymentRestAmount'), $remain_to_pay, 1, $tva_tx, 0, 0, 0, 0, '', '', 'TTC');
```
eldy, line 1298: If you have an invoice with product 1 of 50 euros VAT 10% + product 2 of 50 euros VAT 20%, so 100 euros without tax and (5+10 = 15) euros of tax, total is 115 euros.
You receive a payment of 100 euros. When you make a credit note using the remain to pay, you have no way to defined what is the good vat. Did the 100 euros was used to pay the product 1 or product 2, or both partially, or to pay just the product without vat or to pay product 1 + its vat and partially product 2. It is not possible to defined this. Do we have to use vat 0 on credit note (if we suppose vat was already covered, or vat 10 or vat 20 ? There is no real rule. Using the default vat of company is not always the best choice. 

So i suggest to introduce a hidden option
INVOICE_VAT_TO_USE_ON_CREDIT_NOTE_WHEN_GENERATED_FROM_REMAIN_TO_PAY, with possible values:
- not defined, we keep vat to 0 (current code):
- if value is string 'default', we do what you suggest, we use the default vat for company (in most cases it will be the higher vat rate in dictionary).
- if value is a float, we use this float value.

htdocs/compta/facture/card.php
vmaury: sorry no, but I'll use a more elegant method ..

PR #28308, opened 2024-02-20
Qual clean dolresource class

W1W1-M:
# Qual clean dolresource class

- Make use of properties from `CommonPeople` trait
- Make use of properties inherited from `CommonObject`


htdocs/resource/class/dolresource.class.php
```diff
@@ -171,7 +131,8 @@ public function __construct(DoliDb $db)
 	{
 		$this->db = $db;
 		$this->status = 0;
-
+		$this->element = 'dolresource';
```
eldy, line 134: This properties are pseud constant. There is no reason to have them modified so they can be set statically and not intothe constructor.

PR #28312, opened 2024-02-20
Fix: Add facture->invoice modulepart translation

mdeweerd:
# Fix: Add facture->invoice modulepart translation

Issue occurred when detecting use of deprecated properties in tests when DolDeprecatedHandler was added.

The comment on line 512 relates to another issue I encounter with open_basedir set, I removed the '@' that fixed that to keep issue resolutions separate.

### DIFF
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -509,7 +509,7 @@ function dol_is_dir_empty($dir)
 function dol_is_file($pathoffile)
 {
 	$newpathoffile = dol_osencode($pathoffile);
-	return is_file($newpathoffile);
+	return is_file($newpathoffile);  // Throws error when outside open_basedir
 }
 
 /**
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/files.lib.php
```diff
@@ -2660,14 +2660,14 @@ function dol_check_secure_access_document($modulepart, $original_file, $entity,
 		}
 	}
 	// Fix modulepart for backward compatibility
-	if ($modulepart == 'users') {
+	if ($modulepart == 'facture') {
+		$modulepart = 'invoice';
```
eldy, line 2664: Too many impact for the moment.
Such standardization would be try when all
* all module->enabled will be replaced by isModEnabled
* all user->rights-> will be replaced with user->hasRight
* all conf->xxx->dir_output will be replaced with ->getElementProperties()
...
So impact of changing modulepart will be reduced enough to start to play with this...

PR #28313, opened 2024-02-20
Fix: Workaround for false security issue in tests

mdeweerd:
# Fix: Workaround for false security issue in tests

This is a quick workaround for the false security issues in tests by setting the ['SELF'] variable to something that is considered harmless


test/phpunit/AllTests.php
```diff
@@ -26,6 +27,9 @@
 print "PHP Version: ".phpversion()."\n";
 print "Memory limit: ". ini_get('memory_limit')."\n";
 
+// Workaround for false security issue with main.inc.php in tests:
+$_SERVER['PHP_SELF'] = "phpunit";
```
eldy, line 31: We should not force this value. This break test that must fails when value is wrong.
I see no reason to force his value to a dummy value. Dummy value may result in dummy test too.

mdeweerd: You can unforce it later once the security test issue is fixed - see #28214 .
This helps unblock that testing is failing on windows-ci.

mdeweerd: Regarding the regex: it's likely that but I did not examine the exact line of code.
The regex could be adapted to allow the case where we have a windows path : `/((?<!^[A-Za-z]):|[;"\'<>\?\(\){}\$%]/`

mdeweerd:
I just tried to run apstats to see why the data is stale, and I got:

```plaintext
In main.inc.php line 263:
                                                                                                                                                             
  [Exception]                                                                                                                                                
  Security injection exception: Access refused to unknown by SQL or Script injection protection in main.inc.php:analyseVarsForSqlAndScriptsInjection type=2
```

Running the CLI tools from a windows environment may be challenging for the end user...

eldy:
I pushed this commit
https://github.com/Dolibarr/dolibarr/pull/28313/commits/36b9806cf3700b41f08e37a220230879d4abcbe1
in your PR
I hope to have this fixing the trouble.

mdeweerd:
It will fix for the phpunit tests on the windows ci platform, but not for the cas I encountered on the CLI (phpstan including the bootstrap), and possibly also other CLI tools.

PR #28315, opened 2024-02-20
Qual: Re-enable open_basedir restriction in tests

mdeweerd:
# Qual: Re-enable open_basedir restriction in tests

Using open_basedir restriction is closer to the production reality
and seemed to generate some issues on a windows development machine.

Note: this applies after fixing test errors reported in #28187 .


eldy:
The return of error
_pagehead() (void)
But there is no more such function. All _pagehead are now return float|int

PR #28317, opened 2024-02-20
fix entity if interventions are shared between entities

rycks:
Fix a bug : impossible to validate an intervention if there configuration is set to "share interventions between entities"... @hregis please confirm / have a look to that PR
thanks !

### DIFF
--- a/htdocs/fichinter/class/fichinter.class.php
+++ b/htdocs/fichinter/class/fichinter.class.php
@@ -575,7 +575,8 @@ class Fichinter extends CommonObject
 			$sql .= ", date_valid = '".$this->db->idate($now)."'";
 			$sql .= ", fk_user_valid = ".((int) $user->id);
 			$sql .= " WHERE rowid = ".((int) $this->id);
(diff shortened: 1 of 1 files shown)


rycks:
@eldy, @hregis explain to me that getEntity should be used only on SELECT requests, not on UPDATE ... then "the good" patch seems to be AND entity = ". ((int) $this->entity); (replace $conf->entity by $this->entity) ... let me check and tests that on my dolibarr dev setup

eldy:
> @eldy, @hregis explain to me that getEntity should be used only on SELECT requests, not on UPDATE ... then "the good" patch seems to be AND entity = ". ((int) $this->entity); (replace $conf->entity by $this->entity) ... let me check and tests that on my dolibarr dev setup

I see. Let me know...

rycks:
hmmm @eldy  as you do the merge i have to make a new PR ?

eldy:
> hmmm @eldy  as you do the merge i have to make a new PR ?

I'm afraid yes. Sorry...

PR #28321, opened 2024-02-21
Fix supplier order default payment mode v18+

W1W1-M:
# Fix supplier order default payment mode v18+
On supplier order form default payment mode is not set when selecting supplier. 
No problem with payment terms.


eldy:
A fi must never fix refactoring of code and fix.
Fix on maintenance branch must be atomic and contains the minimal changes to have the bug fixed for end user event if change is not clean, minimum is priority in old branch. Quality must or refactoring must be done into develop branch.

W1W1-M:
> A fi must never fix refactoring of code and fix. Fix on maintenance branch must be atomic and contains the minimal changes to have the bug fixed for end user event if change is not clean, minimum is priority in old branch. Quality must or refactoring must be done into develop branch.

Oups, ok will keep that in mind.

Issue #28303, opened 2024-02-20
Solved or closed ticket should automatically open by Email Collecter module upon further email from customer


alsoft10:
### Feature Request

Email collector module doesn't re-open after getting further reply via email from customer email.

### Use case

Generally ticket are closed by organizational employee after the problem is solved. Now the same ticket can be opened by the customer after reply on the same ticket after few days. Dolibarr adds the new message using the help of Email Collector Class to closed ticket, but don't re-open the ticket again automatically.

### Suggested implementation

Upon collecting response by **Collect_Responses_In**, closed or solved ticket should be re-opened. Other wise employee would not know about the new reply. If the ticket is not assigned, then the re-opened ticket status should be **Unread**. And if the ticket is assigned, then the re-opened ticket status should be **Assigned**.

alsoft10:
Feature is implemented.

PR #28323, opened 2024-02-21
Dolibarr can now re-open ticket upon customer replay on the same ticket via email#28303

alsoft10:
# Instructions
 [#28303 ](https://help.github.com/articles/getting-started-with-writing-and-formatting-on-github/) according to this feature, Dolibarr will re-open the ticket upon getting reply via email. Dolibarr email collector class will re-open the tick and make the status **Unread**, if the ticket is solved or closed and it is not assigned to any user. It will make the ticket status **Assigned**, if the ticket is closed or solved and it is assigned to a user already in past.


alsoft10:
@eldy ticket status will be changed for solved or closed tickets upon further customer reply via email.

PR #28324, opened 2024-02-21
Dolibarr can now re-open ticket upon customer replay on the same ticket via email#28303 

alsoft10:
# Instructions
 [#28303 ](https://help.github.com/articles/getting-started-with-writing-and-formatting-on-github/) according to this feature, Dolibarr will re-open the ticket upon getting reply via email. Dolibarr email collector class will re-open the tick and make the status **Unread**, if the ticket is solved or closed and it is not assigned to any user. It will make the ticket status **Assigned**, if the ticket is closed or solved and it is assigned to a user already in past.


alsoft10:
@eldy ticket status will be changed for solved or closed tickets upon further customer reply via email.

htdocs/emailcollector/class/emailcollector.class.php
```diff
@@ -2601,8 +2603,24 @@ public function doCollectOneCollector($mode = 0)
 										$errorforactions++;
 										$this->errors = $actioncomm->errors;
 									} else {
-										$operationslog .= '<br>Event created -> id='.dol_escape_htmltag($actioncomm->id);
+										if ($objectemail->status == Ticket::STATUS_CLOSED || $objectemail->status == Ticket::STATUS_CANCELED) {
```
eldy, line 2606: When we are here, email is not always related to a ticket.
You should be able to add a check on $fk_element_type to fix this

Also, as this change the behaviour, can you push this on develop branch.

eldy: The pull request system is done this way.
If you made some change in the same local branch you develop the first initial PR, by just pushing the code on your own fork should also update the PR automatically.
So just try to fix on you local branch called 18.0 (using a new name of branch before pushing is a better practice) and it should be enough...

alsoft10: As per your suggestion, I have updated the pull request with a Ticket check.

eldy:
Replaced by #28451

PR #28326, opened 2024-02-21
Fix supplier invoice default payment v18

W1W1-M:
# Fix supplier invoice default payment v18
On supplier invoice form default payment terms & mode is not set when selecting supplier.


eldy:
This problem has been fixed in v19.
Fixing the way you do will also lost the data that is preselected by user or by a default value when an error occured (of a mandatory field is not defined). It is not possible to have a good fix in v18 because we must be able to make a difference between the form was posted to submit it and the form was posted because we changed the supplier.
This is possible in v19 and the correct fix should be available in v19 only. 
So we won't fix it in v18 as it break other cases.

PR #28335, opened 2024-02-21
NEW multiselect with checkbox in categories/tags search for thirdparty list

joseplluis:
NEW multiselect with checkbox in categories/tags search for thirdparty list


htdocs/societe/list.php
```diff
@@ -370,6 +390,10 @@
 		$search_alias = '';
 		$search_categ_cus = 0;
 		$search_categ_sup = 0;
+		$searchCategoryCustomerOperator = 0;
```
eldy, line 393: can you resuse the variable $search_categ_cus and $search_categ_sup ?
We need a var named 'search_...' as the use of name starting with search_ has a meaning, it is used to remenber and restore filters when changing page.

PR #28343, opened 2024-02-22
datenextrun is mandatory as dolibarr 16.0

rycks:
In dolibarr 16.0 and + datenextrun is mandatory ...


rycks:
aaaaaaaaa error sorry

PR #28369, opened 2024-02-23
FIX: Paiment list does not display right amount

noec764:
Better Display should be like

### DIFF
--- a/htdocs/compta/paiement/list.php
+++ b/htdocs/compta/paiement/list.php
@@ -197,7 +197,7 @@ if (GETPOST("orphelins", "alpha")) {
 	$reshook = $hookmanager->executeHooks('printFieldListWhere', $parameters); // Note that $action and $object may have been modified by hook
 	$sql .= $hookmanager->resPrint;
 } else {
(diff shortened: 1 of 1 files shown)


htdocs/compta/paiement/list.php
```diff
@@ -197,7 +197,7 @@
 	$reshook = $hookmanager->executeHooks('printFieldListWhere', $parameters); // Note that $action and $object may have been modified by hook
 	$sql .= $hookmanager->resPrint;
 } else {
-	$sql = "SELECT p.rowid, p.ref, p.datep, p.fk_bank, p.amount, p.statut, p.num_paiement";
+	$sql = "SELECT p.rowid, p.ref, p.datep, p.fk_bank, pf.amount, p.statut, p.num_paiement";
```
eldy, line 200: We show the data of table llx_paiement table p, so we need the p.amount.
If you add a table from a subtable (here table pf that is detil of 1 paiement dispatched on n invoice), you also need to add an aggregate function (for example SUM).

PR #28373, opened 2024-02-23
Qual: Fix/complete/refactor ModuleTest case.

mdeweerd:
# Fix: Correct CommonClassTest's constructor

Dataproviders did not work because of the issue with the constructor
    
# Qual: Refactor module name list, add mapping to class name

Based on ModuleTest and search for modules in the code, complete the
list of modules and map to the class names.

This will allow reuse in the ModuleTest.

# Qual: Refactor ModulesInit test

Use the updated common module mapping list, more complete than
the original list.
Also refactor the test to use a data provider.

# Note

Module list needs review/completion.


test/phpunit/ModulesTest.php
```diff
@@ -38,52 +39,65 @@
...
-		'Categorie','ClickToDial','Collab','Commande','Comptabilite','Contrat','Cron','DataPolicy','Dav','DebugBar','Deplacement','DocumentGeneration','Don','DynamicPrices',
-		'ECM','EmailCollector','EventOrganization','Expedition','ExpenseReport','Export','ExternalRss','ExternalSite',
-		'Facture','Fckeditor','Ficheinter','Fournisseur','FTP','GeoIPMaxmind','Gravatar','Holiday','HRM','Import','Incoterm','Intracommreport',
-		'KnowledgeManagement','Label','Ldap','Loan',
-		'Mailing','MailmanSpip','Margin','ModuleBuilder','Mrp','MultiCurrency',
-		'Notification','Oauth','OpenSurvey','Paybox','PaymentByBankTransfer','Paypal','Prelevement','Printing','Product','ProductBatch','Projet','Propale',
-		'ReceiptPrinter','Reception','Recruitment','Resource',
-		'Salaries','Service','SocialNetworks','Societe','Stock','Stripe','SupplierProposal','Syslog',
-		'TakePos','Tax','Ticket','User','Variants','WebServices','WebServicesClient','Website','Workflow','Workstation','Zapier');
-		foreach ($modulelist as $modlabel) {
```
eldy, line 73: I don't understand where is the foreach after the change ?

PR #28376, opened 2024-02-23
NEW Filter shipments by selecting multiple shipping methods

c3do:
This pull request modifies the shipping list to allow the selection of multiple shipping methods.
This modification will be especially useful for mass actions.

Not being used to the "Dolibarr approach" on how to create forms, do not hesitate to ask me to correct my errors.


c3do:
I disagree with Travis Ci that my variable would not be sanitized... It is sanitized twice, by GETPOST, and by the array_filter function

PR #28385, opened 2024-02-23
Qual: Review pre-commit configure - re-enable no-commit-to-branch

mdeweerd:
# Qual: Review pre-commit configure - re-enable no-commit-to-branch

This reviews some of the documentation inside the pre-commit-config.yaml file. It also re-enables the no-commit-to-branch hook.

@eldy Just add
```bash
export SKIP=no-commit-to-branch   # In your .bashrc or session.
```

I've been commiting to develop several times now by mistake which could have been avoided with this hook.
It should be enabled for all of us, you're the exception (and you should be working in branches for some stuff).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,22 +5,30 @@ repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.5.0
     hooks:
-      # This hook test the name of the branch and return error if the name is 'develop' or an official version 'x.y'
-      # so you can't commit if your branch name is not a custom branch name. Using a custom branch name is good
-      # practice that make easier to manage PR. You can enable/disable this if you want to commit or not on branch
-      # with common names.
-      #- id: no-commit-to-branch
-      #  args: [--branch, develop, --pattern, \d+.0]
-      # This check format of xml files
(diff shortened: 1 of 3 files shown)


eldy:
What's wrong with same name ?
A lot of contributors are doing a fork, then commit on their fork (in the cloned branch, so v18, v19 or develop), then push the PR.
Exemple: https://github.com/Dolibarr/dolibarr/pull/28194
Even if using a different name is better (it allows to work on several different PR), using a branch name into the fork that is similar to the one of the main stream is frequent by a lot of contributors. This will make the CTI fails for all of them, right ?

mdeweerd:
New contributors possibly will not install pre-commit.
If they do, then they likely know git and will create a branch.
Committing to develop/19.0/... in a fork ends up being a mess when synchronising with upstream and having local changes - and possibly also a mess when proposing PRs that then include changes that were already proposed and changed.
So keeping a clean fork with a main/develop that does not diverge from the upstream is a good thing.
When you're not very familiar with git it is at least cumbersome to recover from a poluted main/develop branch in a fork.

The Continuous Integration does not fail because this rule was explicitly excluded already in CI - otherwise CI would fail on the main branch.

PR #28388, opened 2024-02-23
Qual: Refactor CodingPhpTest / CommonClassTest - less verbosity, but better when error

mdeweerd:
# Qual: Refactor CodingPhpTest

- Use dataprovider (better progress report, better errors, better continuation)
- Use dol_dir_list's exclude_filter capability (do not traverse the excluded dirs)
- Reduce debug output from dolibarr.log (not really relevant for these tests).

# Qual: CommonClassTest - less verbosity, but better when error
    
Report the test method and parameters in case of error.
Less verbosity about setup.

### DIFF
--- a/test/phpunit/CommonClassTest.class.php
+++ b/test/phpunit/CommonClassTest.class.php
@@ -79,7 +79,9 @@ abstract class CommonClassTest extends TestCase
 		$this->savlangs = $langs;
 		$this->savdb = $db;
 
-		print __METHOD__." db->type=".$db->type." user->id=".$user->id;
+		if ((int) getenv('PHPUNIT_DEBUG') > 0) {
+			print get_called_class()." db->type=".$db->type." user->id=".$user->id;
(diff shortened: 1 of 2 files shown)


test/phpunit/CommonClassTest.class.php
```diff
@@ -99,7 +101,8 @@ public static function setUpBeforeClass(): void
 			die(1);
 		}
 
-		if (isset($_ENV['PHPUNIT_DEBUG'])) {
+		if ((int) getenv('PHPUNIT_DEBUG') > 0) {
+			print get_called_class()."::".__FUNCTION__.PHP_EOL;
 			print get_called_class().PHP_EOL;
```
eldy, line 106: We may remove this line if you add the other one just before ?

eldy:
We have now a test for var_dump into the pre-commit. So may be we can comment completely the test into the phpunit ?

PR #28389, opened 2024-02-23
Enable open_basedir restriction during CI setup on windows

mdeweerd:
# Enable open_basedir restriction during CI setup on windows

The condition is that the 'documents' (data) directory exists before running PHP code that tries to access it - PHP can no longer create the directory itself when open_basedir restriction is in effect with the documents directory as a root path.
This may need to be documented towards end-users.

Currently there are some notifications regarding nu_soap during the DB setup process showing that there is a relative include that needs to be fixed.

### DIFF
--- a/.github/workflows/windows-ci.yaml
+++ b/.github/workflows/windows-ci.yaml
@@ -69,10 +69,11 @@ jobs:
             ${{ matrix.os }}-${{ env.ckey }}-${{ matrix.php_version }}-
       - name: Create local php.ini
         shell: cmd
-        if: false
+        # if: false
         # Note: Setting the basedir before initialising the database results in open_basedir errors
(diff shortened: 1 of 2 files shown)


mdeweerd:
@eldy
I think I found why there is an open_basedir violation during the database setup process.
In htdocs.install/upgrade2.php, we have `include_once 'inc.php';` before reading the configuration file and in 'inc.php' the DOCUMENT_ROOT is set to '..' .
This is also true in 'upgrade.php' but we do not notice it there because there is no attempt to erase files.

I'll let you choose how you fix that.

Also: you can merge this PR - I simplified the windows-ci.yaml file as I found what was preventing the use of the open_basedir restriction during the setup.sh script (setup of DB).

PR #28424, opened 2024-02-25
Qual: New Phan plugin for testing that argument matches regex

mdeweerd:
# Qual: New Phan plugin for testing that argument matches regex

This Plugin - currently applied to GETPOST - allows verifying that a selected argument of a function matches a regular expression.

There are several cases where the argument is invalid.  I let the codeowners correct those.

Also added a test for isModEnabled - it detects 3 cases where the value may be invalid - maybe the best solution is to add a '@phan-ignore... ParamMatchRegexError' annotation for these cases.


mdeweerd:
I added annotations to ignore the remaining cases - I could not find the reason for the reports.

PR #28425, opened 2024-02-26
Input select from foraign keys

leninrivas:
# NEW|New [*input select from foraign key*]
[*Module builder create dictionaries from foraign keys with select of array.*]

### DIFF
--- a/htdocs/admin/dict.php
+++ b/htdocs/admin/dict.php
@@ -2621,13 +2621,13 @@ function fieldList($fieldlist, $obj = null, $tabname = '', $context = '')
 				$if = 0;
 				while ($if < $numf) {
 					$objf = $db->fetch_object($resqlf);
-				
+
 					if ($objf) {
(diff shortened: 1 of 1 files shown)


htdocs/admin/dict.php
```diff
@@ -1758,6 +1759,13 @@
 				$showfield = 0;
 			}
 
+			// Other fk_s
+			$isforaign = false;
```
mdeweerd, line 1763: Spelling: isforeign

leninrivas: Fixed, Thanks

eldy:
Can you paste a screenshot of what the feature brings ?

leninrivas:
> Can you paste a screenshot of what the feature brings ?

This is.
dictionaryfk

eldy (changes requested):

eldy:

PR #28428, opened 2024-02-26
add arrayfields in parameters for hook printFieldPreListTitle

frederic34:
image


eldy:
We can add filters for any fields, and I think we can't add a filter for any field we want to filter on when field is not selected.
I will make this invoice in pending unti we have the "generic filter tool"...

frederic34:
@eldy I removed the search

PR #28439, opened 2024-02-26
fix #28431

### LINKED ISSUE #28431: Access a third party from the contact list
### Bug

In the contact list: /contact/list.php

If we display the Third Party column.

If you choose a prospect/client company: no mistake.
If you choose a company, neither prospect nor customer: error.

I noticed that each time, we are redirected to a link:
/comm/card.php?socid=12345

There is no error when using the following link:
/societe/card.php?socid=12345

So, if the company is not a customer, the link must be /societe/card.php

### Environment Version

18/19

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

8.2

### Environment Database

_No response_

### Environment URL(s)

/contact/list.php

_No response_

_No response_

### Attached files

image

frederic34:


sonikf:
Ηι @frederic34 
Issue reporter is using a thirdparty that is not customer, prospect or vendor.
Is there a valid use case for this?

frederic34:
image

image
image

with the fix we have a better link

sonikf:
Sorry, I'm not questioning your fix, my question is what is the purpose/use case of a thirdparty that is not customer, prospect or vendor and should this be allowed?

frederic34:
it just correct the link, create a thirdparty with no type is already possible

kuscoo:
> Sorry, I'm not questioning your fix, my question is what is the purpose/use case of a thirdparty that is not customer, prospect or vendor and should this be allowed?

You can use third parties as an address book, without commercial vocation (for example for institutions, etc.)

PR #28447, opened 2024-02-27
New: Phan Plugin (with fixer) for GETPOST->GETPOSTINT

mdeweerd:
# New: Phan Plugin (with fixer) for GETPOST->GETPOSTINT

This detects GETPOST calls with second parameter 'int' and can convert to GETPOSTINT

(Not added to default checks (yet))


mdeweerd:
Added fixer for deprecated module names.

eldy:
I was just thinking of using rector to do this...  ;-)

mdeweerd:
I think phan is faster, both probably have a similar learning curve.

However, phan can use the rule to continue to monitor any new occurrences!  (After the the other changes are integrate to avoid extra messages).

PR #28452, opened 2024-02-27
Fix: Replace deprecated modulename with new in isModEnabled

mdeweerd:
# Fix: Replace deprecated modulename with new in isModEnabled
    
Replacement of old module names performed with dedicated detection and fixer in phan.


fappels:
This PR removed shipments from menu. Shipments now only displayed if isModEnabled('delivery_note') should be 'expedition'. Think not only menu affected see also in showLinkedObjectBlock. Not sure replace to delivery_note is individual commit to revert.

mdeweerd:
I'll check the deprecated modules table.

fappels:
See also more affected 'commande' -> 'order' , 'facture' -> 'invoice' ...

mdeweerd:
I can revert changes with the inverse operation, but these module names are in principle deprecated.  So the issue may need a different fix.

mdeweerd:
@eldy I think we need your view on this.

In htdocs/core/modules, we find the following subdirectories:

```action/ asset/ bank/ barcode/ bom/ cheque/ commande/ contract/ delivery/ dons/ expedition/ expensereport/ export/ facture/ fichinter/ holiday/ hrm/ import/ mailings/ member/ movement/ mrp/ oauth/ payment/ printing/ printsheet/ product/ product_batch/ project/ propale/ rapport/ reception/ security/ societe/ stock/ stocktransfer/ supplier_invoice/ supplier_order/ supplier_payment/ supplier_proposal/ syslog/ takepos/ ticket/ user/ usergroup/ workstation/```

Thes are at least the system directories - we can see that bank and member are there, meaning that banque and adherent have been renamed on disk.

However, we still see facture, fichinter (not ficheinter as suggested in a recent change), et.

So the question is: should the directories be renamed (git mv) to the new names or (and) should we revert to the old names.

We can also consider that only the renamed directories are deprecated.

The Dolibarr code has distributed locations where the same issue of module naming is treated, so this should be centralized.  There are also several locations (at least 2) with different methods for finding a module.

Personally I think that it's best to revert first except the cases where the directory names match the new names.
Then update the codebase to centralize the module handling functions.
Then consider deprecating other names.

eldy:
Yes please can you make a revert ?
We will address this in slower steps...

PR #28463, opened 2024-02-27
NEW : Statistic page donation

Thomas905:
Complete statistics page for donations


htdocs/don/class/donstats.class.php
```diff
@@ -76,17 +81,31 @@ public function __construct($db, $socid, $mode, $userid = 0)
 		$this->socid = ($socid > 0 ? $socid : 0);
 		$this->userid = $userid;
 		$this->cachefilesuffix = $mode;
+		$this->join = '';
+
+		if ($status === '0' || $status === '1' || $status === '2') {
```
mdeweerd, line 86: Doing a strict string comparisson on $status that is a string will not yield the expected result when it is indeed an int.

htdocs/don/stats/index.php
```diff
@@ -29,56 +29,62 @@
 require_once DOL_DOCUMENT_ROOT.'/don/class/don.class.php';
 require_once DOL_DOCUMENT_ROOT.'/don/class/donstats.class.php';
 require_once DOL_DOCUMENT_ROOT.'/core/class/dolgraph.class.php';
-
+require_once DOL_DOCUMENT_ROOT.'/core/class/html.formcompany.class.php';
+if (isModEnabled('categorie')) {
```
mdeweerd, line 33: CI will soon report a notification for this: it should be category now.

htdocs/langs/fr_FR/donations.lang
```diff
@@ -35,3 +35,4 @@ DonationPayments=Paiement du don
 DonationValidated=Don %s validé
 DonationUseThirdparties=Utiliser un tiers existant comme coordonnées du donateur
 DonationsStatistics=Statistiques des dons
+NbOfDonations=Nombre de dons
```
mdeweerd, line 38: FYI: non en_US translations are managed on Transifex.

htdocs/don/stats/index.php
```diff
@@ -29,56 +29,62 @@
...
 $WIDTH = DolGraph::getDefaultGraphSizeForStats('width');
 $HEIGHT = DolGraph::getDefaultGraphSizeForStats('height');
 
-$userid = GETPOSTINT('userid');
-$socid = GETPOSTINT('socid');
+// Load translation files required by the page
+$langs->loadLangs("donations");
+
+$userid = GETPOST('userid', 'int');
+$socid = GETPOST('socid', 'int');
```
mdeweerd, line 43: Use GETPOSTINT for those (will also be suggested by CI soon).

htdocs/don/stats/index.php
```diff
@@ -39,8 +39,8 @@
 // Load translation files required by the page
 $langs->loadLangs("donations");
 
-$userid = GETPOST('userid', 'int');
-$socid = GETPOST('socid', 'int');
+$userid = GETPOSTINT('userid', 'int');
+$socid = GETPOSTINT('socid', 'int');
```
mdeweerd, line 43: GETPOSTINT does not need 'int' as a second parameter.

htdocs/don/class/donstats.class.php
```diff
@@ -83,11 +83,11 @@ public function __construct($db, $socid, $mode, $userid = 0, $typentid = 0, $sta
 		$this->cachefilesuffix = $mode;
 		$this->join = '';
 
-		if ($status === '0' || $status === '1' || $status === '2') {
+		if ($status == '0' || $status == '1' || $status == '2') {
```
mdeweerd, line 86: As $status is an int, comparing to ints is faster - otherwise PHP needs to convert the constant string to an int 😉

Thomas905: Except that it currently send string

mdeweerd: It is a string in practice because you have:

https://github.com/Dolibarr/dolibarr/blob/7ca1a9b5f0ea57b6b8e3d9ebb88b19331519e674/htdocs/don/stats/index.php#L50

I do not know if you need 'intcomma', but if 'int' is enough you can use GETPOSTINT which returns an int.  If not, you should cast the result of GETPOST to int possibly after checks (GETPOSTISSET, etc), but the expected type is int once you feed it further down function calls, etc.

eldy:
Can you paste a screenshot of the new donation home page ?

Thomas905:
image

PR #28465, opened 2024-02-27
fix isModEnabled for all case

frederic34:


htdocs/core/class/conf.class.php
```diff
@@ -394,6 +394,9 @@ public function setValues($db)
 								if ($modulename == 'propale') {
 									$modulename = 'propal';
 								}
+								if ($modulename == 'projet') {
```
mdeweerd, line 397: I suggest to move to the usage of https://github.com/Dolibarr/dolibarr/blob/55e201f63fb3bcaeced946bf8e4f3f59d004ce51/htdocs/core/lib/functions.lib.php#L209-L229 .

Also, this is another location (besides API) where I see supplierproposal so it may need to be added as a deprecated module name as well.

mdeweerd: I wam thinking of :

```php
if (isset(DEPRECATED_MODULE_MAPPING[$modulename]) {
   $modulename=DEPRECATED_MODULE_MAPPING[$modulename]);
}
```

isModEnabled already does that, but $modulename is also use to check the path for the modules.


https://github.com/Dolibarr/dolibarr/blob/55e201f63fb3bcaeced946bf8e4f3f59d004ce51/htdocs/core/lib/functions.lib.php#L237-L257

frederic34: checking for `project` must check `projet` ?

projects are back

frederic34: same pb with contract

htdocs/core/lib/functions.lib.php
```diff
@@ -222,7 +222,7 @@ function getDolUserInt($key, $default = 0, $tmpuser = null)
 		'fichinter' => 'intervention',
 		'product_fournisseur_price' => 'productsupplierprice',
 		'product_price' => 'productprice',
-		'projet'  => 'project',
+		'project'  => 'projet',
```
mdeweerd, line 225: The key is the old module name!

mdeweerd: This looks like the better way to do "it" - maybe refactor this to a common function :
https://github.com/Dolibarr/dolibarr/blob/55e201f63fb3bcaeced946bf8e4f3f59d004ce51/htdocs/core/lib/admin.lib.php#L1122-L1134

htdocs/core/class/conf.class.php
mdeweerd: Apparently the mapping old/new is not fully completed.

```
grep -rP 'class .* extends DolibarrModules' htdocs 
```

I implemented a mapping table in the test, maybe it's best to make this table available to production code and use that to determine the class for a module name (if it's in the table), and the search method if it's not in the table.

https://github.com/Dolibarr/dolibarr/blob/55e201f63fb3bcaeced946bf8e4f3f59d004ce51/test/phpunit/CommonClassTest.class.php#L301-L304

mdeweerd: Using 'array_flip' the deprecated table can be converted to a table that can convert the 'new' names to the 'old' names that still seem to be used as class names.

Anyway, rather than copying the same conversions "all over the place", there should be (a) common method(s).

frederic34:
Without change

image

After change, by example project, categories are back

image

frederic34:
image

image

mdeweerd:
Following #28452 I prepare a PR that reverts the module names that do not correspond to their directory counterpart.
We should then check to see if(/which) other names also need to be reverted.

PR #28468, opened 2024-02-27
FIX column when update validation and verify date and time

lamrani002:
# FIX|Fix update availibility and add two verification for date and time

### DIFF
--- a/htdocs/bookcal/availabilities_card.php
+++ b/htdocs/bookcal/availabilities_card.php
@@ -136,14 +136,24 @@ if (empty($reshook)) {
 
 	$triggermodname = 'BOOKCAL_AVAILABILITIES_MODIFY'; // Name of trigger action code to execute when we modify record
 
-	$dateStartTimestamp = dol_stringtotime(GETPOST('start'));
-	$dateEndTimestamp = dol_stringtotime(GETPOST('end'));
 
-	$startHour = intval(GETPOST('startHour'));
-	$endHour = intval(GETPOST('endHour'));
+	$startday = GETPOST('startday', 'int');
+	$startmonth = GETPOST('startmonth', 'int');
+	$startyear = GETPOST('startyear', 'int');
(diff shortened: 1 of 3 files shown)


htdocs/bookcal/availabilities_card.php
```diff
@@ -202,8 +227,7 @@
 	}
 
 	print load_fiche_titre($langs->trans("NewObject", $langs->transnoentitiesnoconv("Availabilities")), '', 'object_'.$object->picto);
-
-	print '<form method="POST" action="'.$_SERVER["PHP_SELF"].'">';
+	print '<form method="POST" action="'.($error == 0 ? $_SERVER["PHP_SELF"] : $_SERVER["PHP_SELF"].'?action=create').'">';
```
eldy, line 230: Do not use param action=create on url.
Into a form, you must use the param action that is a sensible parameter) as a <input> field so param does not appear into end user url submited url

htdocs/bookcal/availabilities_card.php
```diff
@@ -132,8 +136,29 @@
 
 	$triggermodname = 'BOOKCAL_AVAILABILITIES_MODIFY'; // Name of trigger action code to execute when we modify record
 
-	// Actions cancel, add, update, update_extras, confirm_validate, confirm_delete, confirm_deleteline, confirm_clone, confirm_close, confirm_setdraft, confirm_reopen
-	include DOL_DOCUMENT_ROOT.'/core/actions_addupdatedelete.inc.php';
+	$dateStartTimestamp = dol_stringtotime(GETPOST('start'));
```
eldy, line 139: In dolibarr dates are not passed as a parameter into only one field. There are propaged into 3 or 6 fields (one field for each component). and then converted into a date using the 
dol_mktime(...the 6 components...). Date into the string has a format that depends on user language and is not predicable on some languages

You should ignore GETPOST('start') and use GETPSOT('startday'),  GETPSOT("sartmonth')...

PR #28469, opened 2024-02-27
FIX subscription must be editable as long as accounting isn't reconciled

daraelmin:


daraelmin:
see https://www.dolibarr.fr/forum/t/impossible-de-modifier-les-cotisations/44368/3

PR #28471, opened 2024-02-27
FIX return method DeleteLine in api_boms file

lamrani002:
return method must be an integer and not an array


eldy:
Can you take a looknat the delete method into the file modulebyilder/templates/class/api_mymodule.class.php
and use a return value the same way it is done in template ?

PR #28474, opened 2024-02-27
Fix type of variable in api_invoices

lamrani002:


htdocs/compta/facture/class/api_invoices.class.php
```diff
@@ -1492,7 +1492,7 @@ public function addPayment($id, $datepaye, $paymentid, $closepaidinvoices, $acco
 
 		// Creation of payment line
 		$paymentobj = new Paiement($this->db);
-		$paymentobj->datepaye     = $datepaye;
+		$paymentobj->datepaye     = strtotime($datepaye);
```
eldy, line 1495: Use of strtotime is forbidden as it does not manage correctly time zone.
You can use instead  dol_stringtotime()

PR #28482, opened 2024-02-27
TODO incompatibility with socket connection

hregis:


htdocs/filefunc.inc.php
```diff
@@ -144,6 +144,7 @@
 $dolibarr_main_document_root = trim($dolibarr_main_document_root);
 $dolibarr_main_document_root_alt = (empty($dolibarr_main_document_root_alt) ? '' : trim($dolibarr_main_document_root_alt));
 
+// TODO incompatibility with socket connection
```
eldy, line 147: Why not replacing 
if !empty()
with
if isset()
Does this work with socket for you ?

Issue #28481, opened 2024-02-27
BUG: REST API "/mos/{id}/produceandconsume" not working and have a confuse behavior


Humml87:
### Bug

Hi,.
either I don't understand the API function or it doesn't work.
I tried now a few hours...

id: 21
request_data:
`{
"inventorylabel": "Produce and consume using API",
"inventorycode": "PRODUCEAPI-YY-MM-DD",
"autoclose": 1,
"arraytoconsume": [{
"objectid": "203",
"fk_warehouse": "677",
"qty": "1"
}],
"arraytoproduce": []
}`

The data origin:
![image](https://github.com/Dolibarr/dolibarr/assets/78662388/ead6cc0c-b020-4eec-b2bb-cfe66d7ae7d5)

**I hope anyone can help me, its very import for my project.**
Thank you together!

Best regards 
Christian

### Environment Version

17...19...

### Environment OS

Ubuntu

### Environment PHP

8.2

### Environment Database

mariaDB

### Environment URL(s)

http://localhost/dolibarr/htdocs/api/index.php/mos/{ID}/produceandconsume

JonBendtsen:
What happens when you send the json to the API, and which of the sub parts of the API are you using? Would you mind posting a picture?

Humml87:
Hallo JonBendtsen,

the result is:
`{
  "error": {
    "code": 500,
    "message": "Internal Server Error"
  },
  "debug": {
    "source": "api_mos.class.php:544 at call stage",
    "stages": {
      "success": [
        "get",
        "route",
        "negotiate",
        "authenticate",
        "validate"
      ],
      "failure": [
        "call",
        "message"
      ]
    }
  }
}`

![image](https://github.com/Dolibarr/dolibarr/assets/78662388/bce9f869-50c3-45b2-a8fd-0d273840de57)

I found some logical errors:
![image](https://github.com/Dolibarr/dolibarr/assets/78662388/ed80cc9d-c2e3-46d3-89fc-ada422f71070)
... if a MoLine with already consumed Products are deleted, sometimes the consumed Products don't delete from the DB.

I think its necessary to refactor the ConsumAndProduce:
- mo_production.php => Move the actions for ConsumAndProduce to a new function in "MoLine" Class.
- api_mos.class.php => then refactor...

@mdeweerd Hi, you make a PR #26805 at 23.11.2023. You know how the API-function work?
[...]

PR #28484, opened 2024-02-28
Qual: Add AbstractRestApiTest class, refactor RestApiUserTest, add RestApiMosTest

mdeweerd:
# Qual: Add AbstractRestApiTest class, refactor RestApiUserTest, add RestApiMosTest

Built start of Mos API test case following #28481.

Creasted AbstractRestApiTest to share Api related methods/initialisations.
Refactored RestApiUserTest to use it.
Created initial RestApiMosTest.


mdeweerd:
I propose to add this to develop as a baseline - RestAPIUserTest is refactored and a common class is created.
The RestAPIMosTest is not active in regression and I leave it for extension with regards to the referenced PR for instance.

PR #28499, opened 2024-02-28
Fix return type string instead of int

lamrani002:


htdocs/core/class/html.formintervention.class.php
```diff
@@ -59,7 +59,7 @@ public function __construct($db)
 	 *	@param	int		$maxlength	Maximum length of label
 	 *	@param	int		$showempty	Show empty line ('1' or string to show for empty line)
 	 *	@param	bool	$draftonly	Show only drafts intervention
-	 *	@return int         		Nbre of project if OK, <0 if KO
+	 *	@return string         		Nbre of project if OK, <0 if KO
```
eldy, line 62: Type does not match the comment.

PR #28501, opened 2024-02-28
Fix return type when output is 1

lamrani002:


htdocs/core/class/html.formcontract.class.php
```diff
@@ -61,7 +61,7 @@ public function __construct($db)
 	 *	@param	int		$showRef	Show customer and supplier reference on each contract (when found)
 	 *  @param	int		$noouput	1=Return the output instead of display
 	 *  @param	string	$morecss	More CSS
-	 *	@return int         		Nbr of contract if OK, <0 if KO
+	 *	@return int|string         		Nbr of contract if OK, <0 if KO
```
eldy, line 64: Can  you also update to doc comment.
So you can for example update  doc comment with
If nooutput = 0: Nbr of contract if OK, <0 if KO
If nooutput = 1: The HTML select string

PR #28509, opened 2024-02-28
Fix variable by adding other type

lamrani002:


htdocs/comm/action/class/actioncomm.class.php
```diff
@@ -316,7 +316,7 @@ class ActionComm extends CommonObject
 	public $icalcolor;
 
 	/**
-	 * @var string Extraparam
+	 * @var string|array Extraparam
```
eldy, line 319: We should not have array in this field as this is the property to store content in database.
What is the point/warning that you try to solve?

lamrani002: PHPDoc type string of property ActionComm::$extraparams is not covariant with PHPDoc type array of overridden property CommonObject::$extraparams.

eldy: > extraparams

I see. It seems $extraparams is sometime string and sometimes array. So i have modified instead the CommonObject class by setting phpdoc to
array| string

PR #28511, opened 2024-02-28
Fix: Modify module mapping principle

mdeweerd:
My changes "conflict" with other integrated at the same time.

I changed the principle of the mapping: map used names to internal names, regardless of deprecation.

I've also updated the CodingPhp test to check only the really deprecated modules.

Another PR is under preparation to revert the module names and the phan checks to not report on them.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -206,25 +206,40 @@ function getDolUserInt($key, $default = 0, $tmpuser = null)
 	return (int) (empty($tmpuser->conf->$key) ? $default : $tmpuser->conf->$key);
 }
 
+
+/**
+ * This mapping defines the conversion to the current internal
+ * names from the alternative allowed names (including effectively deprecated
+ * and future new names (not yet used as internal names).
+ *
+ * This allows to map any temporary or future name to the effective internal name.
+ *
+ * The value is typically the name of module's root directory.
+ */
 define(
-	'DEPRECATED_MODULE_MAPPING',
+	'MODULE_MAPPING',
 	array(
-		'actioncomm' => 'agenda',
-		'adherent' => 'member',
-		'adherent_type' => 'member_type',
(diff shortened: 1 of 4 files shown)


frederic34:
@mdeweerd lost intervention again, modulename is ficheinter (even directory is fichinter) 🌵 

image

mdeweerd:
I kept "ficheinter" in the list of module names that are mapped.

But we have to map fichinter to ficheinter ... and use 'ficheinter' everywhere else?

frederic34:
isModEnabled is checking $conf->modules array and in this one you will find a key named ficheinter, you are looking for fichinter which is never present

mdeweerd:
OK, what determines that it is 'ficheinter' in that array ?
We could change that so that the module_name === directory_name .

frederic34:
image

modFicheinter => ficheinter

image

mdeweerd:
Ok, so class names rule internal names but are still controlled in the module.
I'll update the used names to ficheinter.

frederic34:
image

mdeweerd:
I added a test to ModuleTest to compare the name generated by the module with the name hard coded in the test.
The indicate that most of the old internal names are still in use, and the mapping therefore looks like this (once I update it):

```php
array(
        'actioncomm' => 'agenda',
        'adherent_type' => 'member_type',
        'entrepot' => 'stock',
        'product_fournisseur_price' => 'productsupplierprice',
        'product_price' => 'productprice',
        'socpeople' => 'contact',
        // Prefer internal names over new names
        'bank' => 'banque',
        'category' => 'categorie',
        'contract' => 'contrat',
        'intervention' => 'ficherinter',
        'invoice' => 'member',
        'member' => 'adherent',
        'order' => 'commande',
        'project'  => 'projet',
        'propal' => 'propale',
        'shipping' => 'expedition',
        'supplier_proposal' => 'supplierproposal',
    );
```

PR #28512, opened 2024-02-28
Fix: Revert module naming

mdeweerd:
# Fix: Revert module naming

Revert module naming, update checks.

The module names have been changed to the names as defined in the 'modMODLABEL' classes.
The ModuleTest has been extended to compare the name set by the module itself against the static table defined in CommonClassTest which has been updated to match those.

The Phan configuration has been changed to generate 3 differen message keys : deprecated module names, unknown module names, the use of new names where the old names are preferred for now.

The module name replacements are in the oldest commits.

The functional changes are in the recent commits.


eldy:
We experience too many regression with so large PR. So i won't validate them until we have restore a stable situation.

I suggest instead to make smaller steps. For exemple by making change on 1 module at a time.
For exemple, normalizing code to use "intervention" instead of "fichinter" or "ficheinter" only, into one PR or several PR, and do other PRs for other modules.

mdeweerd:
> We experience too many regression with so large PR. So i won't validate them until we have restore a stable situation.
> 
> I suggest instead to make smaller steps. For exemple by making change on module at a time. For exemple, normalizing code to use "intervention" instead of "fichinter" or "ficheinter" only into one PR or severla PR, other PR for other modules.

This PR is to come back to using the names defined in the module classes which is the original situation.

Currently the "modern names" are in effect.

Comming back to the previous situation in a lot of different PRs while maintaining a correct CI is going to need some scripting to update the configuration files and CI tests module by module.

mdeweerd:
There was a conflig with the develop branch which I fixed by choosing the develop branch solution, while I belive the approache should be different.

I'll let you choose a method and then we stick to it and fix the other implications.
I prefer the rule that the module property "name" defined in the module is the target name.
So if that name is not the desired name then the module name in the module should be updated.

https://github.com/Dolibarr/dolibarr/blob/c0962889cee9a72d5e47c41b71b8a26dde7b68f4/htdocs/core/lib/functions.lib.php#L222-L265

mdeweerd:
It seems to me that the ranaming is acted and that the code was updated to cope with that.
I change this PR to draft and will apply the improvements made to the phan configuration & fixers are ported to the develop branch while having a configuration compatible with the develop branch.

mdeweerd:
I think the project move past reverting the naming.

PR #28515, opened 2024-02-28
New: Add REST API functions for MRP-Module

Humml87:
solution for #28481

Add REST API functions:
image

Some refactoring was necessary to get a safe result.


Humml87:
PR conflict: #28564

Humml87:
Hallo @eldy ,
before i spend time again to refactor the PR, i like to get your mind for this:
IMG_20240307_115633

Which variant is the way to go?
- copy
- move 
- some else...

Best regards
Christian

eldy:
The pattern used by dolibarr for business logic is "Active Record" (See https://wiki.dolibarr.org/index.php?title=Language_and_development_rules#Patterns_of_code_organization)
So if we need to share a "business logic" between the web service controller and the gui page, we have 2 choices:
- Add a business method that call the low level CRUD method and call this method in both the webservice and the gui pages (what you called "move" on your graph i think)
- Or copy the business logic from Gui page into the webservice (what you called "copy" on your graph). This remains possible but is good only if the business sequence of code if just few lines.  If more complex, sharing a business common method (inside the CRUD class), as suggested by the Active Record pattern, is better . So the "move" option looks better.

eldy:
PR seems locked in edition so no way for us to solve conflict.
Also a lot of code may have been validated since the PR and the PR may be obsolete, or too difficult to merge due to other changes. Best choice is too cancel it to avoid to keep PR with too low chances to be validated.

PR #28563, opened 2024-03-01
FIX return in method put on api_ticket

lamrani002:


htdocs/ticket/class/api_tickets.class.php
```diff
@@ -384,9 +384,9 @@ public function put($id, $request_data = null)
 
 		if ($this->ticket->update(DolibarrApiAccess::$user)) {
 			return $this->get($id);
+		} else {
+			throw new RestException(304, $this->ticket->error);
```
eldy, line 388: As this is a technical error, you can use http 500 instead of 304

PR #28565, opened 2024-03-01
Fix: Script tests on Windows where PHP executable is not in path.

mdeweerd:
# Fix: Script tests (on Windows)

- Set `exit(255)` instead of `exit(-1)` in `scripts/bank/export-bank-receipts.php`. On *nix, the -1 converts into an exist code of 255 but on windows it is -1, making the test fail.
- Add a helper method to execute php scripts using the php binary used to run phpunit.


scripts/bank/export-bank-receipts.php
```diff
@@ -34,7 +34,7 @@
 // Test if batch mode
 if (substr($sapi_type, 0, 3) == 'cgi') {
 	echo "Error: You are using PHP for CGI. To execute ".$script_file." from command line, you must use PHP for CLI mode.\n";
-	exit(-1);
+	exit(255);
```
eldy, line 37: Currently all scripts, templates are using -1, -2, ... for error code. So for consistency we should keep this.
What is the test that fails on windows ?

PR #28567, opened 2024-03-01
fix phpstan

frederic34:


compgeniuses:
why so many PR for the same thing?

PR #28580, opened 2024-03-02
FIX: Count of virtual stock at Services and MoLine with disabled stock change

### LINKED ISSUE #28560: Bug: Virtual stock calculation is wrong if a MoLine was marked as "Stock changed disabled"
### Bug

If a MoLine is marked with "Stock changed disabled", the "Virtual stock" at the product is also reduced. The "Virtual stock" must untouched in this case.
Mo:
image

Product:
image

### Environment Version

19

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

Humml87:
view the following Issue:
Close #28560


Humml87:
I don't like to fix the problem of the test pre-commit / pre-commit (pull_request):
image

It's already fixed in the open PR #28515. I like to avoid a conflict.

PR #28585, opened 2024-03-02
Qual: FactureTest - use assertions rather than dying in the setup.

mdeweerd:
# Qual: FactureTest - use assertions rather than dying in the setup.

The 'die()' instructions have been replaced - not all tests should fail because the preconditions for a test case are not met.

### DIFF
--- a/test/phpunit/FactureTest.php
+++ b/test/phpunit/FactureTest.php
@@ -56,20 +56,9 @@ class FactureTest extends CommonClassTest
 	 */
 	public static function setUpBeforeClass(): void
 	{
-		global $conf,$user,$langs,$db;
-
-		if (!isModEnabled('facture')) {
-			print __METHOD__." module customer invoice must be enabled.\n";
-			die(1);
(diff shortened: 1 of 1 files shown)


test/phpunit/FactureTest.php
```diff
@@ -56,20 +56,9 @@ class FactureTest extends CommonClassTest
...
-
-		if (!isModEnabled('facture')) {
-			print __METHOD__." module customer invoice must be enabled.\n";
-			die(1);
-		}
-		if (isModEnabled('ecotaxdeee')) {
-			print __METHOD__." ecotaxdeee module must not be enabled.\n";
-			die(1);
-		}
-
```
eldy, line 69: Why removing the begin of transaction. We want to have nothing modified in database when test is launched manually for 1 file only.

mdeweerd: `$db->begin()` is already in the parent class `CommonClassTest`.

Issue #23022, opened 2022-11-28
Translation inconsistent


John-H-Smith:
### Bug

Translations in the companies are inconsistent.
I'm using de_DE as language.
If navigating to /societe/card.php, you can (with module `notifications` enabled) see a tab 'Benachrichtigungen'.
But while accessing `societe/consumption.php` or `societe/paymentmodes.php`, it uses the default fallback 'Notifications'.
I followed the paths in the code to `core/class/translate.class.php` (line 612, function `trans($key....)`.
The associative array `tab_translate` in the `translate.class.php` is different for the three frontend pages (`card.php`, `consumption.php` and `paymentmodes.php`).
For `card.php`, the key 'Notifications' is set. But the key isn't present in the both other cases.
As I first would have to read into the docs and the translation mechanism, I assume it would be much quicker if you just check it yourself.

P.S.: In the `langs/de_DE/companies.lang` file is no 'Notifications' key set either.

### Environment Version

16.0.2

### Environment OS

AlmaLinux release 8.6

### Environment Web server

Apache/2.4.37

### Environment PHP

PHP 7.4.33

### Environment Database

10.6.10-MariaDB

### Attached files

![image](https://user-images.githubusercontent.com/18233867/204256836-76c95634-c923-41fb-9582-c34cb1516457.png)

![image](https://user-images.githubusercontent.com/18233867/204256972-875ab6c9-1d2b-41b6-8186-3345591c05c7.png)

![image](https://user-images.githubusercontent.com/18233867/204256909-9fdda1ef-dff6-4f8b-8601-727e7c7372ec.png)

ksar-ksar:
Hello,

It is still the case with V18 ?

John-H-Smith:
Hi @ksar-ksar ,
yes it is (Dolibarr v18.0.0).

PR #28590, opened 2024-03-02
FIX issue #23022

c3do:
FIX for https://github.com/Dolibarr/dolibarr/issues/23022


c3do:
@eldy when fixing a bug, is it better to correct it on develop or on the oldest version possible? and if version, which version?

eldy:
When fixing a bug, if you can fix it into old version, it is better on old version. It will be merged automatically on higher version.
Question is on which version. Answer is generic: The lower that you can keeping a risk of automatic merge pb lower. In real life , fixers fix bug into version n-4, -3, -2 or just -1 with n = current version.

PR #28598, opened 2024-03-03
Fix: Ignore errors (related to open_basedir restrictions

mdeweerd:
# Fix: Ignore errors (related to open_basedir restrictions

Used '@' on specific checks on the file system that failed with open_basedir restrictions.  With these changes several phpunit tests now pass with open_basedir restriction in effect

### DIFF
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -586,11 +586,7 @@ function dol_compare_file($a, $b)
 function dol_is_dir($folder)
 {
 	$newfolder = dol_osencode($folder);
-	if (is_dir($newfolder)) {
-		return true;
-	} else {
-		return false;
-	}
+	return (bool) @is_dir($newfolder);
 }
 
 /**
@@ -616,7 +612,7 @@ function dol_is_dir_empty($dir)
 function dol_is_file($pathoffile)
 {
 	$newpathoffile = dol_osencode($pathoffile);
-	return is_file($newpathoffile);
+	return (bool) @is_file($newpathoffile);
 }
 
 /**
@@ -885,7 +881,7 @@ function dol_copy($srcfile, $destfile, $newmask = '0', $overwriteifexists = 1, $
 		dol_syslog("files.lib.php::dol_copy failed Permission denied to overwrite target file", LOG_WARNING);
 		return -1;
 	}
-	if (!is_writable($newdirdestfile)) {
+	if (!(bool) @is_writable($newdirdestfile)) {
 		dol_syslog("files.lib.php::dol_copy failed Permission denied to write into target directory ".$newdirdestfile, LOG_WARNING);
 		return -2;
 	}


htdocs/core/lib/files.lib.php
```diff
@@ -479,11 +479,7 @@ function dol_compare_file($a, $b)
 function dol_is_dir($folder)
 {
 	$newfolder = dol_osencode($folder);
-	if (is_dir($newfolder)) {
-		return true;
-	} else {
-		return false;
-	}
+	return (bool) @is_dir($newfolder);
```
eldy, line 482: This may hide errors that we would like to see. We should not have open basedir errors if code is correctly designed and setup correct. I would prefer to see first if we can fix this cases first...

mdeweerd: The test cases explicitly go to non existing paths.

To fix this while still logging open_basedir access issues, one would need to do something like https://stackoverflow.com/a/36690291/6378634 .  I.e., replace the error handler, execute the test, restore the handler, check if an error occured.  If yes, report the error.  And return the boolean result of the test.

eldy: I think that Dolibarr should not try to go out of non existing path, if the setup is correct.
If this is affirmation true, we have no reason to hide warnings because when it happens, we want to see warnings.
In which use case (except phpunit), do you have access out of the open_basedir area ?

mdeweerd:
> In which use case (except phpunit), do you have access out of the open_basedir area ? 
- I had to add '/tmp' in my 18.0.2 production installation after enabling the open_basedir restriction as recommended on the security page - I did not examine in detail what was causing it exactly as it's my production installation and I do not want to mess that up in case there are sideeffects.  I believe that I had issues uploading files for instance.
- When enableing open_basedir restrictions during the BDD installation/upgrades in windows-ci, this step fails.  I think this was also true on my production platform whan I upgraded (but I am not sure).


> I think that Dolibarr should not try to go out of non existing path, if the setup is correct.

I agree, but I also believe in "gracefull degradation".  The code could be amended to use the "@" prefixes, but handle the case a bit differently in case the return is "void".  I.e., the method calls will not return a value true/false/0/1 so that can be detected and used to decide to log a message/eventually generate a popup.

> If this is affirmation true, we have no reason to hide warnings because when it happens, we want to see warnings.

We are not getting warnings as far as I remember, these are fatal errors. EDIT: there are quite a few open_basedir notifiations that are "only" warnings, I did not observa a fatal stop recently.

mdeweerd:
@eldy 
Complement to my previous reply:

I managed to identify the main reason the DB setup process was failing: when the documents directory does not exist yet, the open_basedir restriction does not allow it to be created (probably because the documents directory is a basedir itself).  By creating it before executing the setup, this greatly improves.

During setup there are essentially issues with a relative path check for the nusoap directory.  Look for "is not within the allowed" in the logs for win-ci in #28389 .

There is a new issue: the tests on windows abruptly stop during a builddoctest.  There is no error reported in CI, but there is no test summary.  This means that an explicit test for the test summary is needed, but also that there is another issue.

htdocs/core/lib/files.lib.php
```diff
@@ -508,11 +508,7 @@ function dol_compare_file($a, $b)
...
```
eldy, line 511: I still don't understand the need to add the @ here.
With or without, the return will be the same, right ? If not, in which case the result differs ?
So difference will be in having or not warning in log, right ?
And in a common usage, we should not have test done out of allowed dir, if we have, we want to see them in log because it means we must make probably a code fix somewhere. The case where we really need to remove warning must be low compared to the one we want the warning.

mdeweerd: Transcript of CLI on windows in cygwin bash.  Paths provided to PHP are windows paths.
The transcript shows the differences.

```console
# 
# CONTENT OF CONFIGUARTION FILE FOR PHP (openbase_dir)
# 
$ cat INI_PHPUNIT/dolibarr.ini 
[php] 
open_basedir="D:\mdeweerd\workspace\dolibarr\\htdocs;D:\mdeweerd\workspace\dolibarr\\documents;D:\mdeweerd\workspace\dolibarr\\test;D:\mdeweerd\workspace\dolibarr\\dev\initdemo;D:\mdeweerd\workspace\dolibarr\phpunit-9.5.phar"

# 
# SET ENVIRONMENT VARIABLE TO USE THIS
# 
$ export PHP_INI_SCAN_DIR=$(cygpath -w $PWD/INI_PHPUNIT)

# 
# Check that the ini file is used
# 
$ php --ini
Configuration File (php.ini) Path: 
Loaded Configuration File:         C:\wamp64\bin\php\php8.2.12\php.ini
Scan for additional .ini files in: D:\mdeweerd\workspace\dolibarr\INI_PHPUNIT
Additional .ini files parsed:      D:\mdeweerd\workspace\dolibarr\INI_PHPUNIT\dolibarr.ini

# 
# Access directory outside allowed paths
# 
$ php -r 'var_dump(is_dir("Q:AEZIFRQ"));'

Warning: is_dir(): open_basedir restriction in effect. File(Q:AEZIFRQ) is not within the allowed path(s): (D:\mdeweerd\workspace\dolibarr\htdocs;D:\mdeweerd\workspace\dolibarr\documents;D:\mdeweerd\workspace\dolibarr\test;D:\mdeweerd\workspace\dolibarr\dev\initdemo;D:\mdeweerd\workspace\dolibarr\phpunit-9.5.phar) in Command line code on line 1

Call Stack:
    0.0001     419816   1. {main}() Command line code:0
    0.0001     419816   2. is_dir($filename = 'Q:AEZIFRQ') Command line code:1

Command line code:1:
bool(false)

# 
# Same access with @ before is_dir .
# 
$ php -r 'var_dump(@is_dir("Q:AEZIFRQ"));'
Command line code:1:
bool(false)

# 
# CLI command with more configuration if needed to show effect
# 
$ php -d error_reporting=32767 -d display_errors=on -r 'var_dump(is_dir("Q:AEZIFRQ"));'
```

If we want something in the log, then this needs to be catched to report it and not interrupt execution for generating a response to the browser.

mdeweerd: @eldy 
In complement to the above, if you want to log something, this could be the approach.
The error_handler needs to be set every time, or one needs to be set once in dolibarr.

```php
function my_is_dir($dir)
{
    $error = null;
    set_error_handler(
        function ($errno, $errstr, $errfile, $errline) use (&$error) {
            if (0 === error_reporting()) {
                return false;
            }

           // Method 1:  Throw an exception but need to catch it.
           //    throw new ErrorException($errstr, 0, $errno, $errfile, $errline); 
           // Method2: set a string, available in the setter of the handler
            $error = "$errfile:$errline:$errno:$errstr";
            // Method3: report with dol_syslog
            dol_syslog($error, LOG_ERR);
            return false;
        }, E_WARNING
    );
    $isDir = @is_dir($dir);
    restore_error_handler();

    print "Provided Error '$error'".PHP_EOL;
    return $isDir;
}

// Test I performed
var_dump(my_is_dir("Q:NOTAUTHORIZED"));
```

mdeweerd:
Closing because this is stalled.

PR #28601, opened 2024-03-03
Fix: Skip UtilsTest on windows where ls is not available

mdeweerd:
# Fix: Skip UtilsTest on windows where ls is not available

The test in question uses 'ls' which is not available on windows. This change limits the execution to non windows systems.

### DIFF
--- a/test/phpunit/UtilsTest.php
+++ b/test/phpunit/UtilsTest.php
@@ -55,6 +55,11 @@ class UtilsTest extends CommonClassTest
 	 */
 	public function testExecuteCLI()
 	{
+		// Needs ls. Skip test if not running on *nix system.
+		if ($this->fakeAssertIfNotUnix(__METHOD__." only works on *nix")) {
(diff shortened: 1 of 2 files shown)


test/phpunit/UtilsTest.php
```diff
@@ -55,6 +55,11 @@ class UtilsTest extends CommonClassTest
 	 */
 	public function testExecuteCLI()
 	{
+		// Needs ls. Skip test if not running on *nix system.
```
eldy, line 58: May be you can jsut replace the "ls" with "dir" for windows ? So the executCLI is also executed in Windows ?

mdeweerd: It's more complex than that - the path would be invalid as it contains forward slashes.
What is the default method to convert a path with mixed '\` and '/' to a path compatible with the windows system.  I did not notice a function doing that.

PR #28610, opened 2024-03-03
NEW: Script for fast identification of missing/unused/duplicate translations

mdeweerd:
# NEW: Script for fast identification of missing/unused/duplicate translations

Fast analysis of files to identify missing, unused and duplicate translations. Fast = less than 3 seconds when there are no exceptions (on my development machine).

Note: #28611 fixes the issue reported for the translations in this PR.


mdeweerd:
I improved the script to cope with cases revealed in new code.

mdeweerd:
New untranslated keys were added since I created the exceptions.
The goal of this PR is to detect them as they appear.

eldy:
The key AI_API_CHATGPT_* is used but with a $key that is dynamically generated.
The other will be used in a future.
Can we exclude this keys ?

mdeweerd:
Yes, these can be excluded - I updated the files.

Note that AI_KEY_API_CHATGPT still gives an issue because it is not defined in a en_US translation file.
Also note that AI_KEY_API_CHATGPT is mostly in admin.lang, and one time in ai.lang (french translation).
So that error needs to be fixed.

*About the ignore files:*
Dynamics keys are added to dev/translation/dynamic_translation_keys.lst which means that their translation is required.

Keys to be ignored (missing or available but not used yet) can be added to:
dev/translation/ignore_translation_keys.lst .

Duplicates that are to be ignored are added to dev/translation/duplicate_translation_keys.lst .

mdeweerd:
AI_KEY_API_CHATGPT is still missing in en_US/admin.lang (since french ai.lang was removed).

eldy:
About the error :  "Note that AI_KEY_API_CHATGPT still gives an issue because it is not defined in a en_US translation file."

There is no need to track such errors. Because when we make a transifex sync, all non en_US languages are erased and replaced with files that contains all keys and only all keys of the en_US version. Can you remove this case ?

mdeweerd:
> There is no need to track such errors. Because when we make a transifex sync, all non en_US languages are erased and replaced with files that contains all keys and only all keys of the en_US version. Can you remove this case ?

I do not fully understand what you mean with regards to the impact for 'AI_KEY_API_CHATGPT' .
AI_KEY_API_CHATGPT  is not in an 'en_US' language file which is used as the base for the translations.  So sure, if it is missing then the other language files will be updated so that it becomes missing there too.

The process only checks the en_US language files to find required keys and the dynamic_translation_keys.lst file for required keys.   Any key in the latter is considered to be a key that can not be extracted from the code but that is required in the base translation.  That also avoids that keys that are present are not listed as unused keys.

So I removed AI_KEY_API_CHATGPT from dynamic_translation_keys.lst - it was not selected because of the non en_US files.

PR #28617, opened 2024-03-04
Fix the default email template is not selected

SylvainLegrand:
No email template is ever selected even if there is only one configured and validated.

### DIFF
--- a/htdocs/core/class/html.formmail.class.php
+++ b/htdocs/core/class/html.formmail.class.php
@@ -420,7 +420,7 @@ class FormMail extends Form
 					$model_id = $this->param["models_id"];
 				}
 
-				$arraydefaultmessage = $this->getEMailTemplate($this->db, $this->param["models"], $user, $outputlangs, $model_id, 1, '', ($model_id > 0 ? -1 : 1)); // If $model_id is empty, preselect the first one
+				$arraydefaultmessage = $this->getEMailTemplate($this->db, $this->param["models"], $user, $outputlangs, $model_id, 1, '', ($model_id > 0 ? 1 : -1)); // If $model_id is empty, preselect the first one
 			}
 
 			// Define list of attached files


htdocs/core/class/html.formmail.class.php
```diff
@@ -420,7 +420,7 @@ public function get_form($addfileaction = 'addfile', $removefileaction = 'remove
 					$model_id = $this->param["models_id"];
 				}
 
-				$arraydefaultmessage = $this->getEMailTemplate($this->db, $this->param["models"], $user, $outputlangs, $model_id, 1, '', ($model_id > 0 ? -1 : 1)); // If $model_id is empty, preselect the first one
+				$arraydefaultmessage = $this->getEMailTemplate($this->db, $this->param["models"], $user, $outputlangs, $model_id, 1, '', ($model_id > 0 ? 1 : -1)); // If $model_id is empty, preselect the first one
```
eldy, line 423: I don't understand.
If $model_id is set for a default selection (it means we want a given $model_id selected), then we must use -1 in $defaultfortype because we don't want to filter on default, we want to filter on $model_id.
if $model_id is not set, then we must use 1 to get the default value.

lvessiller-opendsi: @SylvainLegrand 
Maybe, give us more details in email templates parameters (some screenshots).
I tried with only one email template and is not selected by default but I forgot to set it as default.
image
Did you try setting the email template as default ?

rycks:
Hello,
we made a new test : just one model for email on invoices not checked "by default":

image


then when i go to a invoice card and click on "send by email", that template is not used

image


but if i set that model as "default" the problem is not present and default invoice model is applied ...


@SylvainLegrand your idea is to be "more simple for user" is that true ? to imagine that if there is only one model THEN it's by default ?

PR #28618, opened 2024-03-04
FIX cloned supplier order creation date as now

FlorentPoinsaut:
# FIX cloned supplier order creation date as now

When a supplier order with a order date set is cloned the new creation date is the order date instead of now.

### DIFF
--- a/htdocs/fourn/class/fournisseur.commande.class.php
+++ b/htdocs/fourn/class/fournisseur.commande.class.php
@@ -5,14 +5,15 @@
  * Copyright (C) 2007		Franky Van Liedekerke	<franky.van.liedekerke@telenet.be>
  * Copyright (C) 2010-2020	Juanjo Menent			<jmenent@2byte.es>
  * Copyright (C) 2010-2018	Philippe Grand			<philippe.grand@atoo-net.com>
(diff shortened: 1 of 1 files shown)


htdocs/fourn/class/fournisseur.commande.class.php
```diff
@@ -1573,7 +1574,7 @@ public function create($user, $notrigger = 0)
 		$sql .= ", ".setEntity($this);
 		$sql .= ", ".((int) $this->socid);
 		$sql .= ", ".($this->fk_project > 0 ? ((int) $this->fk_project) : "null");
-		$sql .= ", '".$this->db->idate($date)."'";
+		$sql .= ", '".$this->db->idate($now)."'";
```
eldy, line 1577: This means we can't create anymore an order with the date of our choice.
The use of $now was already managed on line 1522
```
if (empty($date)) {
			$date = $now;
		}
```
So to have the current date when cloning we must set ->date = dol_now() into the clone method.

FlorentPoinsaut: Thank you for you feedback.
I fix this now. :)

PR #28623, opened 2024-03-04
FIX: Verification of variables

### LINKED ISSUE #28622: Undefined array key 0
### Bug

Undefined array key 0 in htdocs/core/class/CMailFile.class.php on line 317

### Dolibarr Version

19.0

### Environment PHP

8.1

### Environment Database

MySQL

When the user wants to send a document by an e-mail (bill, proposal...).

### Attached files

_No response_

JeromeBARBOTIN:
Verification of the variables setting before its use.
Fix for #28622


htdocs/comm/propal/class/propal.class.php
```diff
@@ -2816,7 +2816,7 @@ public function classifyBilled(User $user, $notrigger = 0, $note = '')
 		}
 
 		if (!$error) {
-			$modelpdf = $conf->global->PROPALE_ADDON_PDF_ODT_CLOSED ? $conf->global->PROPALE_ADDON_PDF_ODT_CLOSED : $this->model_pdf;
+			$modelpdf = isset($conf->global->PROPALE_ADDON_PDF_ODT_CLOSED) ? $conf->global->PROPALE_ADDON_PDF_ODT_CLOSED : $this->model_pdf;
```
frederic34, line 2819: you can write `$modelpdf = getDolUserString('PROPALE_ADDON_PDF_ODT_CLOSED', $this->model_pdf);`

htdocs/core/actions_sendmails.inc.php
```diff
@@ -434,10 +434,10 @@
 					setEventMessages($mesg, null, 'mesgs');
 
 					$moreparam = '';
-					if (isset($paramname2) || isset($paramval2)) {
+					if (isset($paramname2) && isset($paramval2)) {
```
eldy, line 437: Why changing the || into &&  ?

If $paramval2 is set and not $paramname2, we should get '&mid='.$paramval2

JeromeBARBOTIN: wit the `||`, you allow to call `$paramname2` even if it is not defined. This lead to a php error (version 19.0.0):
screenshot

We could also remove  `|| isset($paramval2)`

eldy: So this means we must replace
`($paramname2 ? $paramname2 : 'mid')`
with
`(!empty($paramname2) ? $paramname2 : 'mid')`

JeromeBARBOTIN: ok, if `empty($paramname2)`does not return an error when `$paramname2`is not set.
This is one of the main reasons I have a lot of crash with Dolibarr 19.0.0 with php 8.1. 
[https://wiki.php.net/rfc/undefined_property_error_promotion](php documentation)

JeromeBARBOTIN: I replaced with `$paramname ?? 'id'`. I tested and it worked.

PR #28643, opened 2024-03-04
FIX wrong GETPOST check

hregis:


hregis:
@eldy why GETPOST("search_type", "int") is not the same result of GETPOSTINT("search_type") ?

GETPOSTINT is designed to always return an int. It is not able to support the value "unknown". Getpost can make a difference between no filter set, or filter set to 0, 1 2...

PR #28646, opened 2024-03-05
Fix: Deposit lines are not created correctly

fmarcet:
If the MAIN_DEPOSIT_MULTI_TVA option is active, it doubles the amount of the deposit line, since the value of $object->total_ttc is not populated and then doubles the amount of the line.


htdocs/compta/facture/card.php
```diff
@@ -1596,7 +1596,7 @@
 									null,
 									0,
 									'',
-									1
+									0
```
eldy, line 1599: So we do not update the total.
Not a problem because later, we make an update if MAIN_DEPOSIT_MULTI_TVA is set.

But what if MAIN_DEPOSIT_MULTI_TVA is off ? May be we can disable only if option is on ?

fmarcet: It's just the opposite. If the parameter is 1, the total is not updated (that's how it was until now). And if the parameter is 0, this is when the total is updated (necessary for the operation done on line 1063 and used if the MAIN_DEPOSIT_MULTI_TVA option is activated).

But I can modify it to be set to 0 or 1 depending on whether the MAIN_DEPOSIT_MULTI_TVA option is activated.

PR #28659, opened 2024-03-06
fix delete line by line

frederic34:


htdocs/accountancy/bookkeeping/listbyaccount.php
```diff
@@ -435,7 +435,9 @@
 			foreach ($toselect as $toselectid) {
 				$result = $object->fetch($toselectid);
 				if ($result > 0 && (!isset($object->date_validation) || $object->date_validation === '')) {
-					$result = $object->deleteMvtNum($object->piece_num);
+					// this function deletes many lines at the same time
+					// $result = $object->deleteMvtNum($object->piece_num);
+					$result = $object->delete($user);
```
eldy, line 440: If we use the delete, it will delete the line only, but when we delete a line of a transaction, we must delete all the lines of the same transaction. For example if you delete the line that add 10 euro in account for bank, we must also delete the line that remove the 10 from the invoice.
I think this is why we used deleteMvtNum().

frederic34: if you select all the lines of a same transactions, the fetch for the second line fails and rollback, you need to select only the first of a transaction to avoid fail

eldy: I see. But if we select only 1 line among the 2 of the transaction, only 1 line will be deleted.
If i am right, i suggest to enhance the error management and return error if fetch is < 0, if 0 we do nothing.

frederic34: if we select one line, two lines are deleted

eldy: I can't see where is the deletion of the second line when you select only 1 line.
The $object->delete() delete only the line with rowid = $toselectedid. Can you point me where the other line with the same value for $piece_num (so the same accountingtransaction) are deleted ?

frederic34: before using $object->deleteMvtNum it fetch the line with rowid to find `piece_num` and then  in deleteMvtNum, it delete whith `piece_num` found in the fetch, not the rowid selected
https://github.com/Dolibarr/dolibarr/blob/c2379bfdf542b54dd09b01d65864612c8a29749f/htdocs/accountancy/class/bookkeeping.class.php#L1650-L1655

eldy: The code mentionned is the code inside deleteMvtNum(), but your PR replace the call of deleteMvtNum() with a simple delete() using the id.

frederic34: if you have three lines rowid 1,2,3 and piece_num 5,5,5
you fetch the first line (rowid 1) it say piece_num = 5 so you call deleteMvtNum(5) and he delete three lines...
you fetch the second line(rowid 2), but she have been deleted so it gives error and rollback and exit

eldy: Sorry, i still miss something: you say
You call deleteMvtNum(5)
But this is the point. According to github diff, your PR is removing the call to deleteMvtNum(). Your PR replaces it with a simple delete(). What is the line of call of deleteMvtNum after your PR ?

frederic34:
I think i undestood my mistake

PR #28661, opened 2024-03-06
Fix resource list

W1W1-M:
# FIX resource list
Commit c544efe seems to break resource list because of the way this list handles extrafields filters. Alos, compared to other lists the SQL SELECT request is in the the resource class method `fetchAll` whereas in general it is found in `list.php`

This PR refactors the resources list to standardize it with other lists, also adding massactions & hooks.


W1W1-M:
#26230

PR #28678, opened 2024-03-07
Update functions.lib.php

### LINKED ISSUE #20209: __DATE_DUE_YMD__ always give date format %d/%m/%Y no matter what main.lang defines it as
### Bug

v15.0.0

The code in `htdocs/core/lib/functions.lib.php` somehow always returns in DMY-format even when our main.lang formats are defined as this:

`
FormatDateShort=%Y-%m-%d
FormatDateShortInput=%Y-%m-%d
FormatDateShortJava=yyyy-MM-dd
FormatDateShortJavaInput=yyyy-MM-dd
FormatDateShortJQuery=yy-mm-dd
FormatDateShortJQueryInput=yy-mm-dd
FormatHourShortJQuery=tt:mm
FormatHourShort=%H:%M
FormatHourShortDuration=%H:%M
FormatDateTextShort=%Y-%m-%d
FormatDateText=%Y-%m-%d
FormatDateHourShort=%Y-%m-%d %H:%M
FormatDateHourSecShort=%Y-%m-%d %I:%M:%S %p
FormatDateHourTextShort=%d %b %Y, %H:%M
FormatDateHourText=%d %B %Y, %H:%M
`

We use __DATE_DUE_YMD__ in email templates for invoices, and for today's email it contains "10/02/2022" when it should be "2022-02-10".

Reproducable every time.

The email template looks like this:

`
Vi saknar betalning för faktura __REF__ med förfallodatum __DATE_DUE_YMD__. Kopia på fakturan finns bifogad.
`

### Environment Version

15.0.0

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

It should respect the date-format specified in main.lang

_No response_

### Attached files

_No response_

bos4711:
Fix #20209

When using multiple languages Dolibarr sometimes switches languages depending on settings. By not loading the "main.lang" file, the proper date formats will not be used and this leads to inconsistency.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -2987,6 +2987,9 @@ function dol_print_date($time, $format = '', $tzoutput = 'auto', $outputlangs = 
 		$format = 'daytextshort';
 	}
 
+	// Make sure the format strings are loaded/updated (#20209)
+	$outputlangs->load("main");
+
 	// Do we have to reduce the length of date (year on 2 chars) to save space.
 	// Note: dayinputnoreduce is same than day but no reduction of year length will be done
 	$reduceformat = (!empty($conf->dol_optimize_smallscreen) && in_array($format, array('day', 'dayhour', 'dayhoursec'))) ? 1 : 0;	// Test on original $format param.


htdocs/core/lib/functions.lib.php
```diff
@@ -2987,6 +2987,9 @@ function dol_print_date($time, $format = '', $tzoutput = 'auto', $outputlangs =
 		$format = 'daytextshort';
 	}
 
+	// Make sure the format strings are loaded/updated (#20209)
+	$outputlangs->load("main");
```
eldy, line 2991: dol_print_date is a low level memory function called a lot of time.
We should not have instruction into it that make IO like loading file from disk.
The main language file should be loaded by the caller that has initialized the $outputlangs object.

bos4711: You are right. I just found a better solution and will make new PR.

PR #28700, opened 2024-03-08
Fix: Ensure the argument type for urlencode is string

mdeweerd:
# Fix: Ensure the argument type for urlencode is string

This ensures that the arguments presented to urlencode have the string type.

### DIFF
--- a/htdocs/accountancy/admin/accountmodel.php
+++ b/htdocs/accountancy/admin/accountmodel.php
@@ -129,7 +129,7 @@ $tabrowid[31] = "";
 
 // List of help for fields
 $tabhelp = array();
-$tabhelp[31] = array('pcg_version'=>$langs->trans("EnterAnyCode"));
+$tabhelp[31] = array('pcg_version' => $langs->trans("EnterAnyCode"));
 
 
 // Define elementList and sourceList (used for dictionary type of contacts "llx_c_type_contact")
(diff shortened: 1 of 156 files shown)


htdocs/accountancy/admin/accountmodel.php
```diff
@@ -353,7 +353,7 @@
 
 // Confirmation de la suppression de la ligne
 if ($action == 'delete') {
-	print $form->formconfirm($_SERVER["PHP_SELF"].'?'.($page ? 'page='.urlencode($page).'&' : '').'sortfield='.urlencode($sortfield).'&sortorder='.urlencode($sortorder).'&rowid='.urlencode($rowid).'&code='.urlencode($code).'&id='.urlencode($id), $langs->trans('DeleteLine'), $langs->trans('ConfirmDeleteLine'), 'confirm_delete', '', 0, 1);
+	print $form->formconfirm($_SERVER["PHP_SELF"].'?'.($page ? 'page='.urlencode((string) ($page)).'&' : '').'sortfield='.urlencode((string) ($sortfield)).'&sortorder='.urlencode((string) ($sortorder)).'&rowid='.urlencode((string) ($rowid)).'&code='.urlencode((string) ($code)).'&id='.urlencode((string) ($id)), $langs->trans('DeleteLine'), $langs->trans('ConfirmDeleteLine'), 'confirm_delete', '', 0, 1);
```
eldy, line 356: What is the goal of this ?
If it is an int or float, a cast will be done.
If not a scalar, it will generate an error, that is the expected behaviour. This may hide bugs by removing a warning we would expect to have.

mdeweerd: Phan reports +700 type mismatches for urlencode().
Automatic type conversions seems to be phased out progressively and if you want to have useful reports from phan for new cases, then one way is to fix the old cases (in stead of adding exceptions which are file based if we want to do it fast, so ignoring new cases).

Therefore given that the conversion to string is currently implicit, casting explicitly to string does not change the behavior but fixes the phan notifications and helps pave the way to allow notifications for future forgotten casts or useless url_encodes.

eldy: A reason for 
`$page    ->    (string) ($page)`
instead of
`$page    ->    (string) $page`

mdeweerd: Yes, because I used a fixer I wanted to ensure that cases such as `1 + 2` are properly converted to strings - the easiest way to do that was to enclose the original expression in parentheses and cast that.

That avoids `(string) 1 + 2` by ensuring `(string) (1 + 2)` .

(Where 1 and 2 could be expressions).

Issue #28710, opened 2024-03-08
date in buy price update doesn't update on changes !


le-reparateur:
### Bug

As you change a purchase price, the new date to apply on this price is not updated in the product card -> buy price

### Environment PHP

8.2.7

### Environment Database

mariadb 10.11.4

### Steps to reproduce the behavior and expected behavior

update a buy price in product card

le-reparateur:
solved by :
$sql .= ", datec = '".$this->db->idate($now)."'"; // mise à jour de la date
added line 421 in /fourn/class/fournisseur.product.class.php

JonBendtsen:
> solved by : $sql .= ", datec = '".$this->db->idate($now)."'"; // mise à jour de la date added line 421 in /fourn/class/fournisseur.product.class.php

Good, may I suggest that you edit the file here https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/fourn/class/fournisseur.product.class.php by clicking the little pencil icon. When you do so, please add #28710 in the commit message

Then you submit a Pull Request and add #28710 in the Pull Request message.

JonBendtsen:
@le-reparateur

eldy:
There is 2 dates on each dolibarr tables: 1 for the creation date (set once, never changed then) and 1 for the update date (modified each time we do an update).
The update date (column "tms" often) should be updated automatically by the database system. What is your database ? mysql or postgres.
May be the problem is that we show the creation data instead of the update date...

ksar-ksar:
🎉

PR #28729, opened 2024-03-09
Update fournisseur.product.class.php

le-reparateur:
#28710
date in buy price update doesn't update on changes !


eldy:
There is 2 dates: 1 for the creation date and 1 for the update date.
The update date should be updated automatically. What is your database ? mysql or postgres.
May be the problem is that we show the creation data instead of the update date...

PR #28734, opened 2024-03-09
Qual: Fix and enable PhanTypeMismatchDimAssignment

mdeweerd:
# Qual: Fix and enable PhanTypeMismatchDimAssignment

This PR fixes and enables PhanTypeMismatchDimAssignment.
This is mostly done by adding more specific array type hinting.


mdeweerd:
@eldy The variable `$code` is used outside the loop where it is a loop variable.

I prefer that you have a look to properly fix it:

https://github.com/Dolibarr/dolibarr/blob/7e31aedf94d3b71ca7b76635a99a49f98945416c/htdocs/core/modules/export/export_excel2007.modules.php#L289-L291

Other question:
IMHO '$val' is supposed to be a string (it comes from extrafields->attributes[key]['label]['otherkey']).
So is this valid code ? (it raises an issue with Phan):

https://github.com/Dolibarr/dolibarr/blob/b1a5c46981b628f4626623c3fe09dcb4d4f2ebe3/htdocs/core/tpl/extrafields_list_print_fields.tpl.php#L74

mdeweerd:
FYI: In my experience phan does not recognize/propagate "very well" that 'empty' is also true for 'null'.

I have at least 2 types of workarounds for that:
- Explicitly test for null equality (which is generally the only test I need);
- Cast the values (here we could cast to string), so that phan sees it as a string and not as a potential null.

eldy:
> FYI: In my experience phan does not recognize/propagate "very well" that 'empty' is also true for 'null'.
> 
> I have at least 2 types of workarounds for that:
> 
> * Explicitly test for null equality (which is generally the only test I need);
> * Cast the values (here we could cast to string), so that phan sees it as a string and not as a potential null.

Don't mind.
So let's explicitly test for null.

PR #28741, opened 2024-03-10
Fix phpstan workflow file

mdeweerd:
# Fix phpstan workflow file

Line was split in two, joined


mdeweerd:
For some reason the phpstan analys stops after 3 minutes.
I am commenting the "stub" files to see if that might have anything to do with it.

mdeweerd:
Removing the stubFiles from the configuration which was updated in other ways during the process.

PR #28743, opened 2024-03-10
Debug phanworkflow

mdeweerd:
# Fix phpstan

phpstan stops analysis after 3 minutes, this PR enables --debug to help understand the reason.


mdeweerd:
Finished debugging - using the stubfiles is not working and is the cause of the failure after 3 minutes (also tried just one stubfile, different settings, etc).
So removing the stubfiles.

PR #28750, opened 2024-03-10
Fix: Resolve several PhanPluginSuspiciousParamPosition cases

mdeweerd:
# Fix: Resolve several PhanPluginSuspiciousParamPosition cases

Resolved several PhanPluginSuspiciousParamPosition cases.


mdeweerd:
@eldy Merge conflict resolved.

htdocs/comm/mailing/card.php
```diff
@@ -1036,7 +1036,7 @@
 				$morehtmlright .= ') &nbsp; ';
 			}
 
-			dol_banner_tab($object, 'id', $linkback, 1, 'rowid', 'ref', $morehtmlref, '', 0, '', $morehtmlright);
+			dol_banner_tab($object, 'id', $linkback, 1, 'rowid', 'ref', $morehtmlref, '', 0, '', '', 0, $morehtmlright);
```
eldy, line 1039: Like all other cases in the PR, here the trouble is not in the position of parameters. 
But the name of the string. It is currently named $morehtmlright when it should be $morehtmlstatus. But position and behaviour is good.

mdeweerd: Ok, I will rename the string then.

mdeweerd: Renamed !

PR #28751, opened 2024-03-11
Fix some cases of PhanTypeSuspiciousStringExpression

mdeweerd:
# Fix some cases of PhanTypeSuspiciousStringExpression

PhanTypeSuspiciousStringExpression is generally about using a boolean in "concatenation" to a string.
Fixed mostly by converting to a string value.  There were a few cases with a misplaced parentheses in a (cond?X:Y) expression.

### DIFF
--- a/htdocs/core/modules/dons/html_cerfafr.modules.php
+++ b/htdocs/core/modules/dons/html_cerfafr.modules.php
@@ -297,6 +297,10 @@ class html_cerfafr extends ModeleDon
 		$prim = array();
 		$secon = array();
 		$trio = array();
+		'@phan-var string[] $prim
+		 @phan-var string[] $secon
+		 @phan-var string[] $trio
+        ';
 		for ($i = 1; $i <= 3; $i++) {
 			$prim[$i] = '';
 			$secon[$i] = '';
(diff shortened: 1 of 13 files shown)


htdocs/core/modules/dons/html_cerfafr.modules.php
```diff
@@ -297,6 +297,10 @@ private function amountToLetters($montant, $devise1 = '', $devise2 = '')
 		$prim = array();
 		$secon = array();
 		$trio = array();
+		'@phan-var string[] $prim
```
eldy, line 300: Did you try to use
`/** var string[] $prim */`
before the $prim = array(); 
So this avoid to introduce phan specific code. Dos this trick works ?

mdeweerd: I use phan in other projects - as far as I am used to using PHPDoc does not work here for phan, so I do not try '/** @var ... */ anymore when the case corresponds to something where that would not work.

The documentation explains that this is a limition of AST:
https://github.com/phan/phan/wiki/Annotating-Your-Source-Code#inline-type-checks-via-string-literals

I try to avoid specific code as much as possible myself too.

PR #28756, opened 2024-03-11
Fix: 'count' on scalar zip->numFiles

mdeweerd:
# Fix: 'count' on scalar zip->numFiles

numFiles is already the number of files.


mdeweerd:

PR #28757, opened 2024-03-11
Fix: getChildrenOfLine's use of includealltree as array

mdeweerd:
# Fix: getChildrenOfLine's use of includealltree as array

includealltree of getChildrenOfLine was used as an array, fixed


mdeweerd:

PR #28758, opened 2024-03-11
Qual: Precise typing for Societeaccount's fields property

mdeweerd:
# Qual: Precise typing for Societeaccount's fields property

Better typing helps static analysis and fixes the message: PhanTypeMismatchArgumentInternal Argument 2 ($array) is $object->fields['site']['arrayofkeyval'] of type null but \array_key_exists() takes array

## Discussion

$field is redefined in the child classes of CommonObject, but IMHO these assignments should be made in the constructor of the child classes, not by redefining the fields.

For 'fields', to enhance readability, there could be a class constant for the initialiser that is used in the constructor for initialisation.

As the typing in CommonObject is different from the types specified in the subclasses, phpstan complains about this difference in level 3.  The solution is not to remove the specific typing from CommonObject, as phpstan will also "need" this more specific typing to avoid other false positives in it's higher verification levels.  And the goal is to help the static tools detect bugs, not avoid that they highlight (potential) issues.


mdeweerd:

PR #28767, opened 2024-03-11
update code and better compliance with php 8

grandoc:


htdocs/core/modules/commande/doc/pdf_eratosthene.modules.php
```diff
@@ -1658,7 +1658,7 @@ protected function _pagehead(&$pdf, $object, $showaddress, $outputlangs, $output
 			}
 
 			// Show recipient frame
-			if (!getDolGlobalString('MAIN_PDF_NO_RECIPENT_FRAME')) {
+			if (!getDolGlobalString('MAIN_PDF_NO_RECIPIENT_FRAME')) {
```
frederic34, line 1661: https://github.com/Dolibarr/dolibarr/blob/7226c7c25a7ad40128fb90e50ce4b514228451c4/htdocs/admin/pdf.php#L96-L98

htdocs/core/modules/contract/doc/pdf_strato.modules.php
```diff
@@ -740,7 +740,7 @@ protected function _pagehead(&$pdf, $object, $showaddress, $outputlangs, $output
 			$this->recipient = $object->thirdparty;
 
 			// Recipient name
-			if ($usecontact && ($object->contact->socid != $object->thirdparty->id && (!isset($conf->global->MAIN_USE_COMPANY_NAME_OF_CONTACT) || getDolGlobalString('MAIN_USE_COMPANY_NAME_OF_CONTACT')))) {
+			if ($usecontact && $object->contact->socid != $object->thirdparty->id && getDolGlobalInt('MAIN_USE_COMPANY_NAME_OF_CONTACT')) {
```
eldy, line 743: The 2 conditions are not the same.
If option never set, we want to act as if szt to 1 (default value)

Issue #27998, opened 2024-02-04
[Discussion] Continuous Integration Failures on 'develop' Branch Impacting Pull Reques


mdeweerd:
### Problem Description
The 'develop' branch frequently experiences continuous integration (CI) failures. Consequently, any Pull Requests (PRs) based on the develop branch either have a failing CI or become failing when rebasing the PR to the develop branch. This results in unnecessary checks to ensure that the CI does not fail due to changes made in the PR.

### Proposed Solution
I suggest adopting a process where all code changes are introduced through PRs, and the develop branch only accepts PRs that have a passing CI. Exceptions to this rule could be *urgent* updates necessary to fix the develop branch, either because the CI only fails in the develop branch and not in the PR, or because the CI in the older version did not detect the quality issue (e.g., changes in version 16.0 merged into the upper versions).

### Rationale
Maintaining a clean 'develop' branch, free from CI failures, should be given priority as a courtesy to other developers. This ensures that PRs are based on a stable foundation, reducing the likelihood of CI issues not related to the work on the PR.

This proposal aims to enhance the development process and improve the overall stability of the codebase.

mdeweerd:
To give a concrete example:

develop branch:
![image](https://github.com/Dolibarr/dolibarr/assets/1504752/776e9e9e-702a-4348-bd11-a33cc0c2126c)

Impact on PRs:
![image](https://github.com/Dolibarr/dolibarr/assets/1504752/3763abcc-6595-4442-90b2-368bf6440e52)

And travis: the majority of builds are failing:
![image](https://github.com/Dolibarr/dolibarr/assets/1504752/e074f73c-d617-40b8-8fb4-22590b09fe0b)

mdeweerd:
This has been documented, closing IMHO the discussion.

https://github.com/Dolibarr/dolibarr/blob/6ce09ce68d2d9279e68cc80dfded6ab7ba03bc30/.github/CONTRIBUTING.md?plain=1#L55

PR #28778, opened 2024-03-12
Fix shipping is not a delivery_note (is sub module of shipping)

fappels:
Related to #28452, #28512 and #28511


mdeweerd:
To fix phan, update the list of modules in dev/tools/phan/config_*.php .
To fix phpunit, update the list of modules in test/phpunit/CommonClass.class.php .

fappels:
For info screen dump of a 'delivery note'

2024-03-12 - 11_47_50 - Delivery

fappels:
@mdeweerd not sure why phan still failing

mdeweerd:
The remaining issue was introduced in develop - probably a "use Stripe..." in conf.class.php that is not useful to the code in conf.class.php.  #27998 .

PR #28781, opened 2024-03-12
fix default current user as commercial for the thirdparty

rycks:
That is a very very old bug (introduced into dolibarr 8.0 !) current user is not used as default commercial contact for a new thirdparty.

That fix is dedicated to @le-reparateur 

I really don't like ternary and ternary into ternary is a mess to read and understand !

For more details please have a look here forum https://www.dolibarr.fr/forum/t/constante-affecter-un-commercial-user-pour-creation-des-tiers/45868/11

### DIFF
--- a/htdocs/societe/card.php
+++ b/htdocs/societe/card.php
@@ -1652,7 +1652,7 @@ if (is_object($objcanvas) && $objcanvas->displayCanvasExists($action)) {
 		print '<td colspan="3" class="maxwidthonsmartphone">';
 		$userlist = $form->select_dolusers('', '', 0, null, 0, '', '', 0, 0, 0, 'AND u.statut = 1', 0, '', '', 0, 1);
 		// Note: If user has no right to "see all thirdparties", we force selection of sale representative to him, so after creation he can see the record.
-		$selected = (count(GETPOST('commercial', 'array')) > 0 ? GETPOST('commercial', 'array') : (GETPOST('commercial', 'int') > 0 ? array(GETPOST('commercial', 'int')) : (!empty($user->rights->societe->client->voir) ? array($user->id) : array())));
+		$selected = (count(GETPOST('commercial', 'array')) > 0 ? GETPOST('commercial', 'array') : (GETPOST('commercial', 'int') > 0 ? array(GETPOST('commercial', 'int')) : array($user->id)));
 		print img_picto('', 'user').$form->multiselectarray('commercial', $userlist, $selected, null, null, 'quatrevingtpercent widthcentpercentminusx', 0, 0);
 		print '</td></tr>';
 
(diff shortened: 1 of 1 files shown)


htdocs/societe/card.php
```diff
@@ -1652,7 +1652,7 @@ function init_check_no_email(input) {
 		print '<td colspan="3" class="maxwidthonsmartphone">';
 		$userlist = $form->select_dolusers('', '', 0, null, 0, '', '', 0, 0, 0, 'AND u.statut = 1', 0, '', '', 0, 1);
 		// Note: If user has no right to "see all thirdparties", we force selection of sale representative to him, so after creation he can see the record.
-		$selected = (count(GETPOST('commercial', 'array')) > 0 ? GETPOST('commercial', 'array') : (GETPOST('commercial', 'int') > 0 ? array(GETPOST('commercial', 'int')) : (empty($user->rights->societe->client->voir) ? array($user->id) : array())));
+		$selected = (count(GETPOST('commercial', 'array')) > 0 ? GETPOST('commercial', 'array') : (GETPOST('commercial', 'int') > 0 ? array(GETPOST('commercial', 'int')) : (!empty($user->rights->societe->client->voir) ? array($user->id) : array())));
```
eldy, line 1655: If the user has no permission to read all thirdparties, then filter must be selected on himself (as he is not allowed to see others, preselected user must be forced to him). The change seems to revert this.
I think se should removed the end test
`(!empty($user->rights->societe->client->voir) ? array($user->id) : array()))`
So when no value are coming from POST, then default value should be set to current user (whatever is permission), right ?

rycks: that code :
`(empty($user->rights->societe->client->voir) ? array($user->id) : array()`

means, if that user does not have rights to see then affect user->id, else empty array ... that's just the opposite of what we want :  

i think we want  "if that user have rights then return user id else return empty array"

no ?

really ternary into ternary into ternary ... that's not fun

rycks: and the use of empty function make the opposite of what we read ...

rycks: in dolibarr 7.0 the use case was : a user who make a thirdpard is linked to that thirdpart

in dolibarr 8.0 that code was replaced but as default entry it could be normal that current user is default contact for the thirdpart he create ...

eldy: "i think we want "if that user have rights then return user id else return empty array"

For the case user has the permission to see all, yes, we want user id. This is also what i sai (case A)
But i also say that we ALSO want user id if user does NOT have permission. Because if user has no permission to see all, he must see only himself (case B). The fix in 
The pr fix the case A (good), but break the case B.
The fix in v18 has fixed B but has broken A. Your fix fix Z but break B.
I think we should ALWAYS have user id, whatever is the permission...

le-reparateur: Hi
Yes for : I think we should ALWAYS have user id by default, whatever is the permission... And never empty array, so both A and B are fixed in the same time.
That was the philosophy at the beggining (before V8)

rycks: hum ... that's a very strange situation: one user can CREATE thirdpart and NOT have read access ?

You mean that i can check the second but not the first here ?

image

le-reparateur: Yes it's a nonsense.
create an modify include read...
Then the first toggle is unnecessary !

PR #28784, opened 2024-03-12
FIX variable module_parts

lamrani002:


mdeweerd:
I believe that this definition for module_parts was the reason that I made all fields optional in:
https://github.com/Dolibarr/dolibarr/blob/16edaf92089ec3b7201b066c4330e213295addea/htdocs/core/modules/DolibarrModules.class.php#L144

If there are mandatory fields, the '?' after the respective fields should be removed there for proper 'type' checking by phpstan and phan.

PR #28785, opened 2024-03-12
Qual: Fix PhanTypeExpectedObjectOrClassName by testing on classname

mdeweerd:
# Qual: Fix PhanTypeExpectedObjectOrClassName by testing on classname

PhanTypeExpectedObjectOrClassName should be fixed by testing that the classname is not
empty.

### DIFF
--- a/dev/tools/phan/config.php
+++ b/dev/tools/phan/config.php
@@ -497,7 +497,7 @@ return [
 		// 'PhanPluginDescriptionlessCommentOnProtectedMethod',
 		// 'PhanPluginPrintfIncompatibleArgumentTypeWeak',
 		'PhanUndeclaredVariableAssignOp',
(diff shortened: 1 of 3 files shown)


mdeweerd:
@eldy
Should there not be stricter checks for `$tmpobjectkey = GETPOST('object');` as it leads to following line where it can load any class simply based on a GET/PORT varaiable ?  (To fix the phan notice a simple check for an empty string would be enough, but I think that the value should be checked for contents (alpha/...) and possibly even class restrictions).
 
https://github.com/Dolibarr/dolibarr/pull/28785/files#diff-5772981a00f29ef99fa4827509f55a1d60ed7bbf566576756eb62c0efd941afaR109

mdeweerd:
:warning: @eldy TRAVIS CI is failing at this moment even if it looks ok see #28904 .

PR #28786, opened 2024-03-12
Fix: In PHP 8 we need to return 0 instead of '' for integers

bafbes:
Fix: In PHP 8 ''<0 and is not converted to 0 so we should return 0 expressly for intreger values


mdeweerd:
GETPOST does not only return integers, but GETPOSTINT does.  So IMHO the caller should either use GETPOSTINT or cast the return value, not GETPOST.

eldy:
> GETPOST does not only return integers, but GETPOSTINT does.  So IMHO the caller should either use GETPOSTINT or cast the return value, not GETPOST.

Confirmed.
Do not fix in GETPOST that must return a string, but fix at kocation of warning with cast or GETPOSTINT.

PR #28805, opened 2024-03-13
# Fix & enable PhanPluginPrintfNotPercent

mdeweerd:
Fix PhanPluginPrintfNotPercent by setting format length to non 0 value.

Example of Notification message:

    'Format string "%0.10F" contains something that is not a percent sign, it will be treated as a format string '%0.10F' with padding of "0" and alignment of '' but no width. Use %% for a literal percent sign, or '%1$.10F' to be less ambiguous' .


eldy:
OK. test result seems similar.
image

PR #28821, opened 2024-03-13
Fix & enable PhanTypeMismatchPropertyProbablyReal

mdeweerd:
# Fix & enable PhanTypeMismatchPropertyProbablyReal

Mostly fixed by better typing


mdeweerd:
@eldy 
There is an issue - Account changes the type of status to an array which is different from CommonObject and CompanyBankAccount is a child class and htdocs\societe\paymentmodes.php supposes it is an int again :
https://github.com/Dolibarr/dolibarr/blob/45a68ba612d3d4bf8b442605a784428690e72c1e/htdocs/compta/bank/class/account.class.php#L203-L209

I will ignore that to allow phan to work with PhanTypeMismatchPropertyProbablyReal (don't worry, I do not make it a 'hard exception', I add it to phan's baseline).

Issue #26549, opened 2023-11-12
API show me thirdparty selected contact notifications


JonBendtsen:
### Feature Request

Under detailed view of a thirdparty I can click the Notifications tab, and under this segment: `LIST OF ALL ACTIVE SUBSCRIPTIONS (TARGETS/EVENTS) FOR AUTOMATIC EMAIL NOTIFICATION` I can add a contact that would get some automatic notifications - (unfortunately I can only add one Event at a time, selection of multiple at once would be cool).

I wish that the API system would allow me to at least see these data, or even better allow me to change this data and add notifications. But it does not. I tried using both
`/dolibarr/api/index.php/thirdparties/<id number>` and `/dolibarr/api/index.php/contacts/<contact id>?includecount=1&includeroles=1` and neither would show any changes from before adding these notifications using the GUI to after adding them.

![jons_notifications](https://github.com/Dolibarr/dolibarr/assets/2551481/b0ca1ab4-ea11-4176-876f-526607fccce2)

### Use case

Automation of who gets what kinds of notifications

### Suggested implementation

expand API under thirdparties and add an API endpoint to get the notification settings for a thirdparty

Like we have `/dolibarr/api/index.php/thirdparties/<thirdparty id>/bankaccounts` or `/dolibarr/api/index.php/thirdparties/<thirdparty id>/categories`

it would be cool to have an API endpoint like `/dolibarr/api/index.php/thirdparties/<thirdparty id>/notifications`

To begin with I could live with GET so I would know which thirdparties to edit in hand in the GUI, but setting them would be fantastic :-)

PR #28823, opened 2024-03-13
Thirdparty notification api GET, CREATE, DELETE & UPDATE

JonBendtsen:
NEW thirdparty notification api GET, CREATE, DELETE & UPDATE
Extends the thirdparty API functions with 4 new functions:
GET which shows you all notifications for a given thirdparty.
DELETE which deletes the specified notification when notification id belongs to supplied thirdparty id
UPDATE which updates the specified notification when conditions are met
CREATE which creates the specified notification


JonBendtsen:
update works, but it produces strange output?

```
{
  "id": "12",
  "type": "email",
  "fk_user": null,
  "email": null,
  "threshold": null,
  "socid": "2",
  "event": "15",
  "target": "1",
  "datec": "2024-03-15",
  "datem": "2024-03-15 23:47:12",
  "ref": null,
  "date": null,
  "duree": null,
  "fk_project": null
}
```

I think that the ref, date and duree field comes from htdocs/core/class/notify.class.php 
```
112:	public $ref;
113:	public $date;
114:	public $duree;
```

but others are not mentioned like
```
115:	public $note;
120:	public $fk_project;
```

even though they are still defined in the same file htdocs/core/class/notify.class.php ?

JonBendtsen:
during editing `notify.class.php` I did not insert references to calling triggers because dolibarr doesn't have a trigger for company notifications. Dolibarr does have modify company, but I saw build errors when I left it in, so out it vent from my code.

```
				if (!$notrigger) {
					// Call trigger
					$result = $this->call_trigger('COMPANY_RIB_CREATE', $user);
					if ($result < 0) {
						$error++;
					}
					// End call triggers
				}
```

JonBendtsen:
during editing `api_thirdparties.class.php` I did not insert code like this which might have been there for some kind of circular protection? I didn't remove it from other functions, but my new functions does not have it.

```
		foreach ($request_data as $field => $value) {
			if ($field === 'caller') {
				// Add a mention of caller so on trigger called after action, we can filter to avoid a loop if we try to sync back again with the caller
				$this->company->context['caller'] = $request_data['caller'];
				continue;
			}

			$this->company->$field = $this->_checkValForAPI($field, $value, $this->company);
		}
```

JonBendtsen:
Create produces the same output that update does, and the same fields which is not in the database but which does exist in the `core/class/notify.class.php file`

```
{
  "id": 16,
  "type": "email",
  "fk_user": null,
  "email": null,
  "threshold": null,
  "socid": 2,
  "event": "8",
  "target": "1",
  "datec": 1710578900,
  "datem": null,
  "ref": null,
  "date": null,
  "duree": null,
  "fk_project": null
}
```

I don't know if it is an issue?

JonBendtsen:
All these notifications seems to be email oriented. What would it take to use a different kind of notification? Like SMS? like possibly sending the recipient a social media message?

JonBendtsen:
Will fix #26549

PR #28827, opened 2024-03-14
UPDATE: add a conf param to disable extrafields code autofill

jyhere:
I find this feature pretty annoying, so i added a misc conf global param to disable the feature.


htdocs/core/tpl/admin_extrafields_add.tpl.php
```diff
@@ -137,10 +137,12 @@ function init_typeoffields(type)
 		});
 
 		/* Autofill the code with label */
+		<?php if (empty($conf->global->MAIN_EXTRAFIELDS_CODE_AUTOFILL_DISABLED)) : ?>
```
eldy, line 140: Can you use instead
`if (!getDolGlobalString('MAIN_EXTRAFIELDS_CODE_AUTOFILL_DISABLED')) {`

PR #28831, opened 2024-03-14
NEW function for removing company contacts

lamrani002:


htdocs/societe/contact.php
```diff
@@ -133,6 +133,27 @@
 	include DOL_DOCUMENT_ROOT.'/core/actions_changeselectedfields.inc.php';
 }
 
+if ($action == 'confirm_delete') {
```
eldy, line 136: If we have a test on action, we must also have a test on a permission here
`if ($action = '...' && ...permission...)`

htdocs/societe/contact.php
```diff
@@ -133,6 +133,27 @@
 	include DOL_DOCUMENT_ROOT.'/core/actions_changeselectedfields.inc.php';
 }
 
+if ($action == 'confirm_delete' && $user->hasRight('societe', 'contact', 'delete')) {
+	$id = GETPOST('id', 'int');
```
eldy, line 137: Missing begin and rollback.

PR #28839, opened 2024-03-14
FIX PHPStan warnings

lamrani002:


htdocs/core/modules/member/doc/pdf_standard.class.php
```diff
@@ -263,7 +263,7 @@ public function Add_PDF_card(&$pdf, $textleft, $header, $footer, $outputlangs, $
 	/**
 	 *	Function to build PDF on disk, then output on HTTP stream.
 	 *
-	 *	@param	Adherent	$object		        Member object. Old usage: Array of record information (array('textleft'=>,'textheader'=>, ...'id'=>,'photo'=>)
+	 *	@param	mixed	    $object		        Member object. Old usage: Array of record information (array('textleft'=>,'textheader'=>, ...'id'=>,'photo'=>)
```
eldy, line 266: What is the other types required in addition to Adherent ?

eldy: Better to set 
`Adherent|array`

PR #28841, opened 2024-03-14
FIX variable by adding a type BOMLine (PHPStan warnings)

lamrani002:


htdocs/mrp/class/mo.class.php
```diff
@@ -231,7 +231,7 @@ class Mo extends CommonObject
 	protected $childtablesoncascade = array('mrp_production');
 
 	/**
-	 * @var MoLine[]     Array of subtable lines
+	 * @var MoLine[]|BOMLine[]     Array of subtable lines
```
eldy, line 234: Looks strange. i see no reason to have the type BOMLine here. The problem may be at a different place.

PR #28845, opened 2024-03-14
FIX variable by casting it with int (PHPStan warnings)

lamrani002:


htdocs/product/price.php
```diff
@@ -292,7 +292,7 @@
 			$newlocaltax2_type = GETPOST('localtax2_type', 'array');
 
 			//Shall we generate prices using price rules?
-			$object->price_autogen = GETPOST('usePriceRules') == 'on';
+			$object->price_autogen = (int) GETPOST('usePriceRules') == 'on';
```
mdeweerd, line 295: Should this not be (int) (GETPOST() == 'on') ?

eldy: @lamrani002 i think @mdeweerd is right...

PR #28853, opened 2024-03-14
FIX variables in project class (PHPStan Warnings)

lamrani002:


htdocs/projet/class/project.class.php
```diff
@@ -2670,10 +2670,10 @@ public function createWeeklyReport()
 					$actioncomm->email_sender = '';
 					$actioncomm->email_to = $to;
 
-					$actioncomm->errors_to = $errors_to;
+					$actioncomm->errors_to = (string) $errors_to;
```
eldy, line 2673: var $errors_to is already declared as string so cast should be useless.
i found a $errors_to = 0 line of code. This is surely the line that make the CTI saying there is a problem. I fixed it for $errors_to but other cases are probably similar...

PR #28878, opened 2024-03-15
Remove PhanTypeMismatchPropertyProbablyReal exceptions in baseline

mdeweerd:
# Remove PhanTypeMismatchPropertyProbablyReal exceptions in baseline

Removing the exceptions as actions were take to remove some.


mdeweerd:
@eldy I removed all the exceptions for PhanTypeMismatchPropertyProbablyReal as you made a fix for at least one of them.
This PR reports the one that is remaining: ignore for the moment or leave the PR open so that you can fix it?

mdeweerd:
All fixed, so the notification can be enabled.

PR #28880, opened 2024-03-15
QUAL use real time when creating action from calendar index

W1W1-M:
# QUAL use real time when creating action from calendar index
Currently, when creating action from calendar index default time is 10am GMT.


eldy:
A better fix that support correctly timezone has been pushed.
0617d4346046b5ec659f474decfc4b1719642268

PR #28883, opened 2024-03-15
FIX problem for display content in editor

lamrani002:


htdocs/core/class/html.formmail.class.php
```diff
@@ -1004,6 +1004,8 @@ public function get_form($addfileaction = 'addfile', $removefileaction = 'remove
 				$out .= '</td>';
 				$out .= '<td>';
 
+				$this->withaiprompt = 'html';
```
eldy, line 1007: This must not be always set.

It is the caller that make the new FormMail() that will decide if he want to set ->withaiprompt or ->withlayout.
So move this at the parent caller place...

htdocs/core/class/html.formmail.class.php
```diff
@@ -1473,8 +1476,8 @@ public function getSectionForAIPrompt($format = '')
 					$('#ai_status_message').show();
 
 					//editor on readonly
-        			if (CKEDITOR.instances.message) {
-						CKEDITOR.instances.message.setReadOnly(1);
+        			if (CKEDITOR.instances.".$htmlContent.") {
```
eldy, line 1479: As we are using a variable to forge javascript content, you must use 
dol_escape_js($htmlContent)
to invoid js injection through this variable.

PR #28884, opened 2024-03-15
NEW animation for display spinner when waiting response

lamrani002:
# NEW|New animation for display spinner 
fod


frederic34:
@lamrani002 why not something like `<i class="fa fa-spinner fa-spin fa-3x fa-fw"></i>` with img_picto

lamrani002:
thanks @frederic34 it's better

PR #28885, opened 2024-03-15
Fix some phan notifications with low occurrence

mdeweerd:
# Fix some phan notifications with low occurrence

Fix some phan notifications that have only one or two occurrences.


mdeweerd:
@eldy I want to highlight the next NOOP code - what is the correct fix for this?

https://github.com/Dolibarr/dolibarr/blob/a1b4bcbaa9739ee049ded6896cdf4697860043eb/htdocs/core/login/functions_dolibarr.php#L147-L149

eldy:
> @eldy I want to highlight the next NOOP code - what is the correct fix for this?
> 
> https://github.com/Dolibarr/dolibarr/blob/a1b4bcbaa9739ee049ded6896cdf4697860043eb/htdocs/core/login/functions_dolibarr.php#L147-L149

I think the noop line                                                        
 !isModEnabled('multicompany'); // Global not available, disable $conf->multicompany->enabled for safety
Must be replaced with
unset($conf->multicompany ->enabled);

PR #28886, opened 2024-03-15
Fix Deprecated Function Calls

mdeweerd:
# Fix Deprecated Function Calls

This replaces several function calls with their non deprecated equivalent.

The calls to deprecated Stripe functions were not replaced as that is more complex and risky.

However, I think there is a `@deprecated` tag that should not be a tag of that kind resulting in notifications like this:

`PhanDeprecatedFunction Call to deprecated function \Commande::setDiscount() defined at htdocs\commande\class\commande.class.php:2450 (Deprecated because: remise_percent is a deprecated field for object parent)`

@eldy Do you confirm that the following method is not deprecated?

https://github.com/Dolibarr/dolibarr/blob/a1b4bcbaa9739ee049ded6896cdf4697860043eb/htdocs/commande/class/commande.class.php#L2441-L2450


eldy:
> 
> @eldy Do you confirm that the following method is not deprecated?
> 

I confirm. This is due to the @deprecated comment related to the field and should not appear into the function definition.
i removed it.

PR #28887, opened 2024-03-15
Fix delete company bank account

### LINKED ISSUE #28822: API will delete a different thirdparties bankaccount
### Bug

Using API DELETE /thirdparties/{id}/bankaccounts/{bankaccount_id} filling out with the thirdparty ID of a different thirdparty I got it to delete a bank account from a unrelated thirdparty.

I think the bug is in 	`public function deleteCompanyBankAccount($id, $bankaccount_id` in the file `htdocs/societe/class/api_thirdparties.class.php` but it could also be in `DOL_DOCUMENT_ROOT.'/societe/class/companybankaccount.class.php';`

### Dolibarr Version

19.0.0 docker.io/tuxgasy/dolibarr    19.0.0-php8.2

### Environment PHP

8.2.7

### Environment Database

11.3.2-MariaDB-1:11.3.2+maria~ubu2204

While developing API to handle thirdparty notifications that is based on extending the thirdparties bankaccounts I noticed that with only 2 thirdparties, 1 == takepos, 2 == the one I created, even if I wrote 1 in the thirdparty field it would still delete notifications from my 2. thirdparty. So I also tested it with bank accounts, and it deletes the bank account there as well. I only have bank accounts on my 2. thirdparty.

I expected it to reject my deletion saying that there is no such bank account for that thirdparty.

### Attached files

_No response_

JonBendtsen:
FIX #28822 API will delete a different thirdparties bankaccount
Issue fixed by only deleting bank account if there is an exactly match between supplied thirdparty id number and the id number of thirdparty in the supplied bank account number.

Error codes:
401 when API user has no permission in Dolibarr
416 Range Not Satisfiable when number mismatch


JonBendtsen:
To anyone reading, maybe also read and/or join the HTTP error code discussion here in the issue:
  `Fix delete company bank account` #28887

htdocs/societe/class/api_thirdparties.class.php
```diff
@@ -1266,18 +1266,20 @@ public function updateCompanyBankAccount($id, $bankaccount_id, $request_data = n
 	public function deleteCompanyBankAccount($id, $bankaccount_id)
 	{
 		if (!DolibarrApiAccess::$user->hasRight('societe', 'creer')) {
-			throw new RestException(403);
+			throw new RestException(401);
```
eldy, line 1269: 401 must be used when user failed to login.
403 when loggin is ok but permission are wrong.
So here, value should be 403.

htdocs/societe/class/api_thirdparties.class.php
```diff
@@ -1273,11 +1273,13 @@ public function deleteCompanyBankAccount($id, $bankaccount_id)
...
 
-		if (!$account->socid == $id) {
-			throw new RestException(403);
-		}
+		$socid = (int) $account->socid;
 
-		return $account->delete(DolibarrApiAccess::$user);
+		if ($socid == $id) {
+			return $account->delete(DolibarrApiAccess::$user);
+		} else {
```
eldy, line 1280: 416 is reserved for the use of range header.
For this use case, best cide seems 403 that may means not allowed (not because peymissions are not granted, but because value of data to delete are not enough consistent to allow access).
A message can be added in the through to help user understand

throw new RestException(403, "Not allowed due to bad consistency of input data");

JonBendtsen: ok

PR #28889, opened 2024-03-15
Fix & enable PhanTypeMismatchArgumentInternalReal

mdeweerd:
# Fix & enable PhanTypeMismatchArgumentInternalReal

This revealed several real issues with 28269b1a0 for #28785 and therefore has a functional impact (which should be positive).


mdeweerd:
@eldy 
Should I add $db to the constructor for KnowledgeManagement?

image

mdeweerd:
Phpstan does not complain about KnowledegeManagment being constructed with a parameter anymore (because I use the value for 'class' from the array and not a 'ucfirst(...)' to use the proper class case).  However that issue still needs consideration.

But as all phpstan and phan issues are fixed, I am making this PR ready for review.

PR #28890, opened 2024-03-15
Proposed function for records that can't be found

zahwyrix:
# NEW [Function for records that can't be found]
I propose a recordNotFound() function on the same model as accessforbidden() to display the record not found message within the Dolibarr interface and customize this page using hooks.


eldy:
Can you include in the same PR an example where this function will be called ?

zahwyrix:
it's done!

PR #28891, opened 2024-03-16
Fix & enable PhanTypeSuspiciousStringExpression

mdeweerd:
# Fix & enable PhanTypeSuspiciousStringExpression


mdeweerd:
@eldy 
global $website is a Webclass instance, but in the lines below it is used as a string and I do not see a __toString  implementation so it is justified that phan notifies about this not being convertable to a string.

What do you suggest?
https://github.com/Dolibarr/dolibarr/blob/5e458bbda801862158ca436c656865a1794abbeb/htdocs/core/filemanagerdol/connectors/php/config.inc.php#L52-L53

mdeweerd:
@eldy 
Another issue.  In the following code it is allowed that $params['filename'] is empty, but then it is being used for file_name_complement on which pathinfo analysis is performed.
That does not seem appropriate when $params['filename'] is empty.

Maybe I am missing something but I thinkt that is not ok.

https://github.com/Dolibarr/dolibarr/blob/5e458bbda801862158ca436c656865a1794abbeb/htdocs/emailcollector/class/emailcollector.class.php#L3512-L3528

mdeweerd:
@eldy 
I suppose the next line needs to oerate on $child and not on $item:

https://github.com/Dolibarr/dolibarr/blob/48d051aba7c366ac1d6acc40cc170591a1397cbb/htdocs/public/webportal/lib/webportal.lib.php#L63

eldy:
> globel $website is a Webclass instance, but in the lines below it is used as a string and I do not see a __toString implementation so it is justified that phan notifies about this not being convertable to a string.

this file is included by files into the /ecm directory. Such files get $website from URL param. Example dir_card.php
$website = GETPOST('website', 'alpha');
Inside the module website, $website is also a string that is propagated into GET or POST submission.

This 2 cases, it would have been better to have named the parameter $websiteref to avoid confusion with the $website as a class instance used at different places. Currently, i don't think there is conflict. Both features works correctly.

eldy:
> @eldy I suppose the next line needs to oerate on $child and not on $item:
> 
> https://github.com/Dolibarr/dolibarr/blob/48d051aba7c366ac1d6acc40cc170591a1397cbb/htdocs/public/webportal/lib/webportal.lib.php#L63

@lvessiller-opendsi Are you able to answer this question of mdeweerd ?

mdeweerd:
> This 2 cases, it would have been better to have named the parameter $websiteref to avoid confusion with the $website as a class instance used at different places. Currently, i don't think there is conflict. Both features works correctly.

@eldy
I agree that the variable is better renamed.  This will help static analysis (which helps developers write correct code).

However I think that the reported issue is different.  In `htdocs/core/filemanagerdol/connectors/php/config.inc.php`, $website is from 'global $website' (inside that file) which is a Website instance.  Maybe global $website is wrong and the variable is provided because config.inc.php is included, or the code using $website is wrong.

In either case, it is clear that the plurality of $website accross files is confusing.

EDIT: I re-enabled the PhanTypeSuspiciousStringExpression on the config.inc.php file to show where the issue occure (after the global $website statement).

eldy:
> However I think that the reported issue is different. In `htdocs/core/filemanagerdol/connectors/php/config.inc.php`, $website is from 'global $website' (inside that file) which is a Website instance. Maybe global $website is wrong and the variable is provided because config.inc.php is included, or the code using $website is wrong.

For me var in this file is a string not a Website instance. What let you think it is a instance of Website  ?

mdeweerd:
I think I saw a 'global $website' and a '$website = new Website($db)' in the same block, but I did a careful analysis and I can't find that in the current code.

However the assumed typing of $website as a global is at least dubious, dol_syslog only tests if it is an object and from my search in the current code I can not determine where it is set as an object.
It uses $websitekey (and not $website) if it is not an object.

```
        global $website, $websitekey;
        if (is_object($website) && !empty($website->ref)) {
            $suffixinfilename .= '_website_'.$website->ref;
        } elseif (!empty($websitekey)) {
            $suffixinfilename .= '_website_'.$websitekey;
        }
```


And now, even after indicating 'string' as the type of the global '$website', Phan is still notifying the same message - so it seems there is something suggesting to Phan that it is a Website object.

mdeweerd:
@eldy I found my mistake with regards to forcing the $website global - phan now passes.

PR #28904, opened 2024-03-16
Fix use  array instead of fileinfo

mdeweerd:
# Fix use  array instead of fileinfo

Detected with PhanTypeSuspiciousStringExpression, seperate PR because of functional impact (to check)


mdeweerd:
I am releasing this because only the windows ci test is failing because it abruptly stop during the builddoctest - which is something I noticed already in another PR discussion.

@eldy 
When running on my PC, I see the message:
"Error, template page can't be called as URL"

And in those cases the program exits with "exit", which of course aborts the tests:

```
// Protection to avoid direct call of template
if (empty($object) || !is_object($object)) {
    print "Error, template page can't be called as URL";
    exit;
}
```

In fact, Travis also fails like this!!!!

So there should be an additional check: no pure exit should exist in the program, they should all have a non 0 exit code at least, and using a function like 'dol_exit' that would behave differently under test and in production would also be recommended.

eldy:
> I am releasing this because only the windows ci test is failing because it abruptly stop during the builddoctest - which is something I noticed already in another PR discussion.
> 
> @eldy When running on my PC, I see the message: "Error, template page can't be called as URL"
> 
> And in those cases the program exits with "exit", which of course aborts the tests:
> 
> ```
> // Protection to avoid direct call of template
> if (empty($object) || !is_object($object)) {
>     print "Error, template page can't be called as URL";
>     exit;
> }
> ```
> 
> In fact, Travis also fails like this!!!!
> 
> So there should be an additional check: no pure exit should exist in the program, they should all have a non 0 exit code at least, and using a function like 'dol_exit' that would behave differently under test and in production would also be recommended.

Here the test exists for security purpose. I think when we deal with security, when we detect a hack attempt, we should not continue to try to go onto another branch of code and should not have to check any condition to stop. Instead i think we must stop as fast as we can. So i think the exit is ok here. But we may have exit(1) instead of exit(0).
The question is why you enter this part of code for business tests when we should never reach it...

mdeweerd:
It was introduced to the develop branch, not for this evolution specifically!

I understand the security but by having a dol_exit() function, the behavior could depend on whether it is in PHPUNIT or not.

Currently it looks like CI is passing but it is not, not even in TRAVIS CI.

I have not done a "git bisect" on this to find the reason - I am focussing on the getting rid of the SuspisciousString notifications.

eldy:
I have modified the exit into exit(1);
Also i have fixed the cause of the travis error d3262ace71c14931e673445183242a8bd21b5a1b

mdeweerd:
The issue was introduced in 38e7d389a92bcc8be948a6d184375260f2c70de4

PR #28921, opened 2024-03-17
Update main.inc.php with new hooks

zahwyrix:
# NEW [Generic hook]
Creation of beforeBodyOpen hook to enable generic actions on every page without exception

### DIFF
--- a/htdocs/main.inc.php
+++ b/htdocs/main.inc.php
@@ -1606,7 +1606,7 @@ if (!function_exists("llxHeader")) {
  */
 function top_httphead($contenttype = 'text/html', $forcenocache = 0)
 {
-	global $db, $conf, $hookmanager;
+	global $db, $conf, $user, $hookmanager;
 
 	if ($contenttype == 'text/html') {
 		header("Content-Type: text/html; charset=".$conf->file->character_set_client);
--- a/htdocs/main.inc.php
+++ b/htdocs/main.inc.php
@@ -1694,7 +1694,7 @@ function top_httphead($contenttype = 'text/html', $forcenocache = 0)
 			$hookmanager = new HookManager($db);
 		}
 		$hookmanager->initHooks(array("main"));
-
+		$error = 0;
 		$result = $user->call_trigger('BEFORE_PAGE_LOAD', $user);
 		if ($result < 0) {
 			$error++;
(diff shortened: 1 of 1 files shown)


htdocs/main.inc.php
```diff
@@ -1694,6 +1694,11 @@ function top_httphead($contenttype = 'text/html', $forcenocache = 0)
 			$hookmanager = new HookManager($db);
 		}
 		$hookmanager->initHooks(array("main"));
+		$error = 0;
+		$result = $user->call_trigger('BEFORE_PAGE_LOAD', $user);
```
eldy, line 1698: Triggers are reserved for use inside a transaction and on business CRUD event.
To include code here, hooks are a better choice.
If you miss the hook, you can add it, but it must be a hook, not a trigger.

htdocs/main.inc.php
```diff
@@ -3853,6 +3858,12 @@ function llxFooter($comment = '', $zone = 'private', $disabledoutputofmessages =
 			print $hookmanager->resPrint;
 		}
 
+		$result = $user->call_trigger('AFTER_PAGE_LOAD', $user);
```
eldy, line 3861: Triggers are reserved for use inside a transaction and on business CRUD event.
To include code here, hooks are a better choice.
If you miss the hook, you can add it, but it must be a hook, not a trigger.

htdocs/main.inc.php
zephyriony: > Triggers are reserved for use inside a transaction and on business CRUD event. To include code here, hooks are a better choice. If you miss the hook, you can add it, but it must be a hook, not a trigger.

Thanks for the reply, I'll see about making the changes, but I think a generic hook is missing, I don't think there's a hook that loads regardless of the content.
I'm glad I've learned the difference between hooks and triggers, because I was having a hard time figuring things out.

htdocs/main.inc.php
```diff
@@ -1694,6 +1694,11 @@ function top_httphead($contenttype = 'text/html', $forcenocache = 0)
 			$hookmanager = new HookManager($db);
 		}
 		$hookmanager->initHooks(array("main"));
+		$parameters = array();  // @phan-suppress-current-line PhanPluginRedundantAssignment
+		$reshook = $hookmanager->executeHooks('beforeBodyOpen', $parameters); // Note that $action and $object may have been modified by some hooks
```
eldy, line 1698: If you want to intercept or execute hook here with the same condtion (so only in some cases of setup), you have already a hook 2 lines after.
I think what you want to do is having a hook global for the top_httphead() that is always called (before outputing the http header or after outputing the http header). 
If i am right, you should set the hoop at begin or at end of top_htthead() and not in the middle inside a line executed if a condition is matched.

PR #28924, opened 2024-03-17
Fix trailing whitespace + format code + minor spelling + enable checks

mdeweerd:
# Fix trailing whitespace + format code + minor spelling + enable checks

This fixes some trailing whitespace issues detected by phan.
Enabled the phan rule and qlso the pre-commit hook in CI.

### DIFF
--- a/dev/tools/phan/config.php
+++ b/dev/tools/phan/config.php
@@ -318,6 +318,7 @@ return [
 	//	to `exclude_analysis_directory_list`.
 	"exclude_analysis_directory_list" => [
 		'htdocs/includes/',
+		'htdocs/install/doctemplates/websites/',
 		'htdocs/core/class/lessc.class.php', // External library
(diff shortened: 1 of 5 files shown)


htdocs/install/doctemplates/websites/website_template-homesubmenu/containers/styles.css.php
eldy, line 1: Can you exclude files into install/doctemplates ?
Because such files are not code produced by developer but are code generated by a feature to "export a website into git to allow to save sources of a template that are inside database with other source code files"
Files may be regenerated and we need to have the minimal change. Until the tool to export can generated clean code, it is better to exclude all files into install/doctemplates.

mdeweerd: I updated the configuration and removed the changes to that directory tree's files.

PR #28925, opened 2024-03-17
Establish a single entry point

rsanjoseo:
Establish a single entry point
Scrutinizer and Composer Installation

# NEW|New [Establish a single entry point]
It uses .htaccess to intercept the current url and always redirect the call to htdocs/index.php.
From index.php the requested file is loaded according to the written url without requiring further changes to the configuration.

To improve security, it also loads the Composer autoload, with the idea of keeping libraries updated outside of htdocs, and moving towards the use of namespaces.

The constants DOL_DOCUMENT_ROOT and DOL_URL_ROOT are now defined directly in index.php, being available from any point.

This not only adds security, but also makes it easier to load additional files by always having the base path available.


eldy:
> To improve security, it also loads the Composer autoload, with the idea of keeping libraries updated outside of htdocs, and moving towards the use of namespaces.

This means the library executed by the application is not the one validated by the users but another version depending on an environment, excluding the fix on external lib bringing by the official version. 
This generates a lot of troubles meaning that the behaviour of application is no more what the developer wanted and what was tested to allow the release but another version and behaviour that is no more under control of the development team.  Past experience shows that this brings more trouble than it solves solution.

Also embedding a .htaccess inside the core will break the own security and setup of the hoster.  Use of .htaccess is not supported by all web servers, the same way and need to have this feature enabled on apache (and it is not the case on heavy loaded hosting server as each url access need an IO to this file).

Removing the line
define('DOL_DOCUMENT_ROOT', $dolibarr_main_document_root); //
also introduce a lot of bug as the var is frequently used after (in web or cli context).

The suggested changes seems also to introduce a lot of way to bypass the security (using a parameter to define which url is called offer a various way to hack the system).
The use of chdir bring also some trouble (on secured systems, this method may simply be disabled).

DOL_DOCUMENT_ROOT must not be defined into index.php but before because it is a variable required even when there is no call of pages (like when using API or when using CLI scripts).

The autodetection of URL seems also broken on some special setup using complex setup of proxy in front of the server web.

Note: 
If you want to update lib using composer, you can just run the composer update command (the vendor dir is currently named /includes, may be renamed to traditional /vendor in a future). But don't forget that this will download different version of external libraries that are not compatible with the rest of Dolibarr,  for example because it does not includes the fixes done by the Dolibarr team to make external libs compatible togethers and with Dolibarr (so application will be unstable).

PR #28926, opened 2024-03-17
fix phpstan

frederic34:


htdocs/fourn/facture/list.php
```diff
@@ -1595,7 +1595,6 @@
 	$facturestatic->alreadypaid = ($paiement ? $paiement : 0);
 
 	$facturestatic->paye = $obj->paye;
-	$facturestatic->socid = $thirdparty->getNomUrl(1, 'supplier', 3);
```
frederic34, line 1598: this is no longer used in kanban view

PR #28930, opened 2024-03-17
NEW API creation of subscription can now link to existing invoice

JonBendtsen:
NEW API creation of subscription can now link to existing invoice
Before this change one could create a subscription using the API, but it would not include a link to an existing invoice. With the changes in this pull request one can submit json like this and the existing invoice and the new subscription are linked together.

```
{
  "linkedObjectsIds": {
    "facture": "6"  
  },
  "note_public": "api Contribution 2024",
  "datec": 1710699079,
  "datem": 1710699079,
  "dateh": 1710509000,
  "datef": 1744999200,
  "fk_type": "1",
  "fk_adherent": "1",
  "amount": "100.00000000"
}
```

Usecase:
We mostly sell subscriptions while our members register for dance classes, meaning we already have a different invoice and we would prefer to not create an invoice just for subscription because it would mean that the member would have to make 2 payments.

Using this API change we can only create the subscription once the member has actually paid the invoice. Granted that was possible before, but with this change we can super easily create a link to the invoice, a link that even shows up in the Dolibarr GUI and is clickable and such.

This change makes it much easier to keep track of membership status and automate the generation of membership contributions.


JonBendtsen:
All code is a direct copy from `htdocs/fourn/class/fournisseur.facture-rec.class.php` with the only change being indentation

PR #28936, opened 2024-03-18
Update commande.class.php trigger name correction according to standard

zahwyrix:
# QUAL [Correction of trigger names]
ORDER_CLASSIFY_BILLED becomes ORDER_BILLED
ORDER_CLASSIFY_UNBILLED becomes ORDER_UNBILLED
following the module_action standard

### DIFF
--- a/htdocs/commande/class/commande.class.php
+++ b/htdocs/commande/class/commande.class.php
@@ -2985,7 +2985,7 @@ class Commande extends CommonOrder
 
 			if (!$notrigger && empty($error)) {
 				// Call trigger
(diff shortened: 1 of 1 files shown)


zahwyrix:
Here's an example of where this is a problem in my module : 
```
	private function extractModuleNameFromAction($action) {
		$parts = explode('_', $action);
		array_pop($parts);
		return strtolower(implode('_', $parts));
	}
```
This function will return order_classify as the module name instead of order

eldy:
Having the suggested code would be better if we start fromscratch.
But the change will create a lot of regression in a lot of external modules for a low gain.
So i suggest instead to enhance your method by removing "_classify" before returning the string.

PR #28949, opened 2024-03-18
NEW functionality for salaries with massaction (bank-transfer)

lamrani002:
# NEW|New functionality in list of salaries for make a credit transfer request with massaction 

gogo


htdocs/salaries/list.php
```diff
@@ -166,6 +166,93 @@
 	$massaction = '';
 }
 
+//credit transfer request
+if ($massaction == 'withdrawrequest') {
+	$langs->load("withdrawals");
```
eldy, line 171: To avoid confusion,can you rename 'withdrawrequest' into 'creditransfer'

htdocs/salaries/list.php
```diff
@@ -166,6 +166,93 @@
 	$massaction = '';
 }
 
+//credit transfer request
+if ($massaction == 'withdrawrequest') {
```
eldy, line 170: To avoid confusion,can you rename massaction = 'withdrawrequest' into 'creditransfer'

PR #28952, opened 2024-03-18
Fix Thirdparty Notification number persist after Contact deletion

### LINKED ISSUE #28220: BUG: Thirdparty Notification number persist after Contact deletion
### Bug

BUG: Thirdparty Notification number persist after Contact deletion.

3_notifications_despite_no_notifications

### Environment Version

docker.io/tuxgasy/dolibarr:18.0.4

### Environment OS

docker.io/tuxgasy/dolibarr:18.0.4

### Environment Web server

Probably apache inside the container, but I have a nginx as a proxy infront

### Environment PHP

 8.1.20

### Environment Database

MySQL or MariaDB 11.2.2-MariaDB-1:11.2.2+maria~ubu2204

### Environment URL(s)

/societe/notify/card.php?socid=007 ;-)

Expected: The displayed notification number would be deleted if the contact is deleted
Actual: It seems like it keeps showing the number even if the contact is deleted

Create thirdparty, created thirdparty contact, create some notifications, delete contact

### Attached files

_No response_

JonBendtsen:
FIX #28220 Thirdparty Notification number correct after Contact deletion
When Thirdparty contacts are deleted the data remains in llx_notify_def, so when the code only counted all Notifications belonging to a Thirdparty, it would include counts with deleted Contacts and show a higher number than what was actually enabled.

This Pull Request not only adjusts the counting method such that the numbers only show notifications belonging to actual existing Contacts, it also deletes notifications belonging to a contact whenever that contact is deleted.

correct_notification_count


JonBendtsen:
this commit `actually delete notifications belonging to a contact when a contact is deleted`

should fix the issue permanently, and also keep the database clean by removing notifications whenever a contact is deleted.

**PRE DELETION**
pre2_delete_contact_notifications

**POST DELETION**
post_delete_contact_notifications

PR #28968, opened 2024-03-19
Fix: Update typing for qty as float

mdeweerd:
# Fix: Update typing for qty as float

Update the typing hints for qty to float (instead of int).
No functionnal issue was found in the review (i.e. conversion to int of qty) - the only outlier seems to be that qty is typed to int in a recruitement table which may need extra care.

### DIFF
--- a/htdocs/bom/class/bom.class.php
+++ b/htdocs/bom/class/bom.class.php
@@ -580,7 +580,7 @@ class BOM extends CommonObject
 	 *
 	 * @param	int		$fk_product				Id of product
 	 * @param	float	$qty					Quantity
-	 * @param	float	$qty_frozen				Frozen quantity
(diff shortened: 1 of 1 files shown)


htdocs/bom/class/bom.class.php
```diff
@@ -580,7 +580,7 @@ public function delete(User $user, $notrigger = 1)
 	 *
 	 * @param	int		$fk_product				Id of product
 	 * @param	float	$qty					Quantity
-	 * @param	int		$qty_frozen				Frozen quantity
+	 * @param	float	$qty_frozen				Frozen quantity
```
eldy, line 583: qty_frozen is not a qty but a status flag (0, 1, ...) to say if we want to have the qty modified depending on qty produced or a constant value.
So should be an int.

PR #28973, opened 2024-03-19
FIX avoid php warning

hregis:


hregis:
@eldy i just move the code in `if (!empty($listinsetup)) { }`

PR #28975, opened 2024-03-19
Create a shipment without documents

zahwyrix:
# NEW [Possibility of creating a shipment without documents or from all documents (genericity)]
At the moment, there's a big restriction on shipping: you can only ship from a propal or an order. Sometimes, however, we just need to make an internal transfer, or we don't yet have the customer order.
I therefore propose to make this form active:
expedition/card.php?action=create
from expedition -> create without having to go through a document by setting product as origin.
I'd have to rework the shipping page, but this would allow you to either directly retrieve a product, or retrieve the line from a document.
So here's my first proposal along these lines: record the origin of the shipment: propal or order for the moment, and we can imagine product in the future, for example for a shipment without documents.

### DIFF
--- a/htdocs/install/mysql/tables/llx_expeditiondet.sql
+++ b/htdocs/install/mysql/tables/llx_expeditiondet.sql
@@ -23,7 +23,7 @@ create table llx_expeditiondet
 (
   rowid             integer AUTO_INCREMENT PRIMARY KEY,
   fk_expedition     integer NOT NULL,
-  fk_origin_line    integer,           -- should be renamed into fk_elementdet. The ID of line of source object (proposal, sale order)
+  fk_elementdet    integer,           -- should be renamed into fk_elementdet. The ID of line of source object (proposal, sale order)
   element_type   	varchar(50) DEFAULT 'order' NOT NULL,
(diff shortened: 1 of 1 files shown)


htdocs/install/mysql/tables/llx_expeditiondet.sql
```diff
@@ -24,6 +24,7 @@ create table llx_expeditiondet
   rowid             integer AUTO_INCREMENT PRIMARY KEY,
   fk_expedition     integer NOT NULL,
   fk_origin_line    integer,           -- Correspondance de la ligne avec le document d'origine (propal, commande)
+  origin_type		varchar(255),		-- Dans le but de pouvoir créer une expédition sans document lié avec une valeur à product par exemple
```
eldy, line 27: To follow the same name used by reception, can you make a PR with only the database enhancement, that will use add the field

`element_type   varchar(50) DEFAULT 'order' NOT NULL,`

eldy:
I also added field fk_element to better match the reception side.
Next step is to fill the value for element_type and fk_element when we add a new value into llx_expeditiondet.

PR #28978, opened 2024-03-19
FIX it creates side effects after logout with takepos

hregis:


htdocs/takepos/index.php
```diff
@@ -73,7 +73,8 @@
 	// We will recalculate amount for foreign currency at next call of invoice.php when $_SESSION["takeposcustomercurrency"] differs from invoice->multicurrency_code.
 }
 
-$_SESSION["urlfrom"] = '/takepos/index.php';
+// TODO it creates side effects
+//$_SESSION["urlfrom"] = '/takepos/index.php';
```
eldy, line 77: This seems to introduce a problem with the logout from takepos. 
After this change when we logout from Takepos, we go back on login (good), but when we login, we do not reache the takepos page again. 
Instead we reach the backoffice (with sometime access error when users are allowed to use the takepos only).

What was wrong with old code ?

eldy:
I pushed a fix that replace the session with a param in url in develop and have cherry pickedninto v19 and 18. Can you test it's ok for you in your v18 ?

PR #28983, opened 2024-03-20
Dolibarr customer area

zahwyrix:
# NEW [Customer area]
An accessible customer space for third-party dolibarr contacts. Nothing's been done yet, but I think it's the first brick in the road.


eldy:
Can you give more information of what you plan to have behind this page.
My understanding is that you plan to have tool for contacts ? To see their proposal, invoice, but accessible with a login/pass owned by the contact ?
If yes, this feature is already in development inside the module "webportal" (currently into develop branch). 
Can you have a look at it ? You can create 1 or several logins for your thirdparties and then they can access the webportal to see all information related to them. Signing a proposal could be done in a future from this area...

zahwyrix:
Yes, thank you!

PR #28996, opened 2024-03-20
Update typing for database functions

mdeweerd:
# Update typing for database functions

This aims to fix several array access notices for db fields arrays (static analysis).

@eldy 
For the $field_desc parameter, I supposed that this is the same as the $fields value in the commonclass object and children.

However, this leads to the report of missing keys such as 'value', 'attribute', 'null', 'extra' .

Should I add those to the typing?


eldy:
@mdeweerd Travis is slow and cost us money. Precommit and phan are fast and cost nothing (except for the planet).
Do you know if we can add a condition to start Travis, PHPStan, Win CI only if precommit and phan are complete and OK ?

mdeweerd:
@eldy 
Making github actions depend on other github actions is relatively easy.

From what I've seen for travis, it's a special case: it needs to be launched differently from within a github workflow I think, and then it can depend on the rest.

The question is also: (how much) do you want to maintain (in) travis?  What is done in travis could move to github actions:
- PHPCS is already done in a github workflow - if you want to have it done with multiple versions, that's possible;
- PHPUNIT tests could be setup on unix platforms, using the same methodology as for windows (i.e.: use a setup script, and probably have the sql cache really work because actions/cache does not work at this time on windows, but it does on Linux).  We could start with mysql (and keep postgresql on travis for the moment).
- php linting on travis: could probably be migrated, but there is linting in a pre-commit hook as well (only executed on changed files at this time).
- ... (I do not have all the checks in travis in mind).

So I'll look into making depend win-ci and phpstan on pre-commit and phan, and later making travis depend on that.


EDIT: FYI I found an action to start travis: https://github.com/marketplace/actions/run-travis-yml  so if that could be set up rather than the current method of starting travis, then making travis depend on the other jobs should be "easy".

eldy:
> @eldy Making github actions depend on other github actions is relatively easy.
> 
> From what I've seen for travis, it's a special case: it needs to be launched differently from within a github workflow I think, and then it can depend on the rest.
> 
> The question is also: (how much) do you want to maintain (in) travis? What is done in travis could move to github actions:
> 
> * PHPCS is already done in a github workflow - if you want to have it done with multiple versions, that's possible;
> * PHPUNIT tests could be setup on unix platforms, using the same methodology as for windows (i.e.: use a setup script, and probably have the sql cache really work because actions/cache does not work at this time on windows, but it does on Linux).  We could start with mysql (and keep postgresql on travis for the moment).
> * php linting on travis: could probably be migrated, but there is linting in a pre-commit hook as well (only executed on changed files at this time).
> * ... (I do not have all the checks in travis in mind).
> 
> So I'll look into making depend win-ci and phpstan on pre-commit and phan, and later making travis depend on that.
> 
> EDIT: FYI I found an action to start travis: https://github.com/marketplace/actions/run-travis-yml so if that could be set up rather than the current method of starting travis, then making travis depend on the other jobs should be "easy".

We still don't know if we abandon travis or not (he foundation is still ok to pay for it). I know that we can use github action to do the same. The script for using github instead of Travis was already implemented and used successfully. They are saved as    
.github/workflows/github_ci_php71_mysql.yml.disabled
.github/workflows/github_ci_php81_mysql.yml.disabled

But making a full CTI script is complexe and the way to do it may differs on platform. Relying too much on github may also be dangerous. so Travis is for us a security to have a backup CTI (Travis vs Github). I know one day, github will become a paying subscription for us. It is also possible that we use our own server for this too. Don't know yet. So for the moment, we can continue to use Travis. But i don't know how this will evolve... So for the moment, i just want to reduce CPU needs...

mdeweerd:
I agree (even in keeping travis active).

Regarding travis: you could run it nightly or some other schedule & for releases to verify/know that the travis flow is still ok, but run the PR and merge executions on github.

Travis has a caching feature as well - running the steps to setup the database is long and avoiding that saves considerable CPU.  By setting a limited cache validity duration + the proper hash on files, the process should still run regularly and also when needed.

PR #28998, opened 2024-03-21
NEW preview exemple page in website when create page

lamrani002:
# NEW|New function for display content page in editor when we create page for website
dolo


eldy:
The message "Fill message with..."  should be replaced when we use it in a website page context with 
"Fill page with..."

Also CTI has error due to a vardump and exist forgotten...

htdocs/website/index.php
```diff
@@ -4521,6 +4580,13 @@ function switchEditorOnline(forceenable)
 		print '</td></tr>';
 	}
 
+	print '<tr id="pageContent" class="hideobject"><td>';
```
eldy, line 4583: Replace <td> with <td class="tdtop"> so the title will be at top.

htdocs/website/index.php
```diff
@@ -4521,6 +4580,13 @@ function switchEditorOnline(forceenable)
...
+	print $langs->trans('PreviewPageContent');
+	print '</td><td>';
+	$doleditor = new DolEditor('content', GETPOST('content', 'restricthtmlallowunvalid'), '', 600, 'dolibarr_mailings', '', true, true, getDolGlobalInt('FCKEDITOR_ENABLE_MAILING'), 20, '90%');
```
eldy, line 4586: Use height of 200 instead of 600, we need a small preview but we won't edit the page content immediatly (hits is generaly done at next step), so no need for a high area.

htdocs/website/index.php
```diff
@@ -4288,10 +4289,68 @@ function switchEditorOnline(forceenable)
 
 	// Example/templates of page
 	if ($action == 'createcontainer') {
+		$formmail = new FormMail($db);
+		$formmail->withaiprompt = 'html';
+
+
 		print '<tr><td class="titlefield fieldrequired">';
```
eldy, line 4296: Add class "tdtop" so title will remain at top, even when the line increase after selecting the Fille layaout or Fill with AI.

PR #28999, opened 2024-03-21
Fix return method deleteLine (PHPStan warnings

lamrani002:


htdocs/expedition/class/api_shipments.class.php
```diff
@@ -410,7 +410,12 @@ public function deleteLine($id, $lineid)
 
 		$updateRes = $this->shipment->deleteLine(DolibarrApiAccess::$user, $lineid);
 		if ($updateRes > 0) {
-			return $this->get($id);
+			return array(
```
eldy, line 413: You must also replace the  
* @return int
with
* @return array

PR #29006, opened 2024-03-21
Fix/phan type invalid left & right operand of add

mdeweerd:
# Fix PhanTypeInvalid{Left,Right}OperandOfAdd

Cast or PHPdoc updates to fix $var1 + $var2 where $var1 is not known to be numeric.


mdeweerd:
@eldy I have really only done the first case out of about 200 I believe (in about 50 files).  Not ready for merge yet.

mdeweerd:
@eldy 
As I added typing on $objectpageold, phan detects a foreach on aliasalt which is a string.

Line 1095 in website.class.php:
```php
                foreach ($objectpageold->aliasalt as $tmpaliasalt) {
                    dol_delete_file($conf->website->dir_temp.'/'.$website->ref.'/containers/'.$tmpaliasalt.'.php', 0, 0, 0, null, false, 0);
                    dol_delete_file($conf->website->dir_temp.'/'.$website->ref.'/containers/*/'.$tmpaliasalt.'.php', 0, 0, 0, null, false, 0);
                }
```

I'll let you fix this as you know the intent (explode aliasalt/other ?).
https://github.com/Dolibarr/dolibarr/pull/29006/files#diff-84f5a811db1ef0dc5e8a7c459e31207ed79d85a76c7beac0c3a4bb9f74b139a6L1095

mdeweerd:
Already merged because the next PR was merged.
Not sure why rank changed to string though...

PR #29007, opened 2024-03-21
FIX : TakePOS | Add product / Delete line of existing invoice

Kazimir42:
Hello,

I don't know if it's the right way to do this fix, but it's not possible to add product or delete a line of a selected invoice in TakePOS. By doing this, it looks like fixed. What do you think?

### DIFF
--- a/htdocs/takepos/index.php
+++ b/htdocs/takepos/index.php
@@ -529,7 +529,7 @@ function ChangeThirdparty(idcustomer) {
 function deleteline() {
 	console.log("Delete line");
 	invoiceid = $("#invoiceid").val();
-	$("#poslines").load("invoice.php?action=deleteline&token=<?php echo newToken(); ?>&place="+place+"&idline="+selectedline+"invoiceid="+invoiceid, function() {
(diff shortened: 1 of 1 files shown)


htdocs/takepos/index.php
```diff
@@ -527,7 +528,8 @@ function ChangeThirdparty(idcustomer) {
 
 function deleteline() {
 	console.log("Delete line");
-	$("#poslines").load("invoice.php?action=deleteline&token=<?php echo newToken(); ?>&place="+place+"&idline="+selectedline, function() {
+	invoiceid = $("#invoiceid").val();
+	$("#poslines").load("invoice.php?action=deleteline&token=<?php echo newToken(); ?>&place="+place+"&idline="+selectedline+"invoiceid="+invoiceid, function() {
```
eldy, line 532: `...line="+selectedline+"invoiceid="+inv...`
looks stange. May be 
`...line="+selectedline+"&invoiceid="+inv...`

Kazimir42:
Yes,

So on takePos when i click on the "Historique" button and i choose an invoice to update 
image

It's adding the lines nicely
On takpose :
image

On document :
image

But when i try to add more product, "kiwi" here i got this :
image

The kiwi as been added on the last invoice opened with takepose terminal "PROV-POS1-0" here or it create a new invoice if no invoice  is already on the current takepos terminal :
image

But it should be on my selected invocie "PROV239" here

PR #29009, opened 2024-03-21
Fix condition in  createDictionary for modulebuilder

lamrani002:
# FIX|Fix in method createNewDictionary by replacing sql request by function in dolibar for check if table exist


htdocs/core/lib/modulebuilder.lib.php
```diff
@@ -1288,7 +1288,7 @@ function createNewDictionnary($modulename, $file, $namedic, $dictionnaires = nul
 	}
 
 	$columns = array(
-		'rowid' => array('type' => 'integer', 'value' => 11, 'extra' => 'AUTO_INCREMENT PRIMARY KEY'),
+		'rowid' => array('type' => 'integer', 'value' => 11, 'extra' => 'AUTO_INCREMENT'),
```
eldy, line 1291: If we remove the "PRIMARY KEY", how is defined the primary key ?
Looking at scripts to create tables in dolibarr we have for exemple 
```
create table llx_c_input_method
(
  rowid    integer AUTO_INCREMENT PRIMARY KEY,
  code     varchar(30),
  libelle  varchar(128),
  active   tinyint default 1  NOT NULL,
  module   varchar(32) NULL
)ENGINE=innodb;
```

How is defined the PRiMARY KEY ?

lamrani002: $primaryKey = 'rowid';
	foreach ($columns as $key => $value) {
		if ($key === 'rowid') {
			$primaryKey = 'rowid';
			break;
		}
		if (!array_key_exists('rowid', $columns)) {
			$primaryKey = array_key_first($columns);
			break;
		}
	}
we check it after declare table of columns

hregis:
@lamrani002 and if one day “rowid” is no longer called “rowid” what do we do? ;-)

htdocs/core/lib/modulebuilder.lib.php
hregis: @lamrani002 and if one day “rowid” is no longer called “rowid” what do we do? ;-)

lamrani002: @hregis thats why we have the seconde check ,it verify if no column named rowid than we take the first column and add primary key

hregis: @lamrani002 heuuu sorry but where is your second check ?

lamrani002: sorry i thought i added it i'll fix it

PR #29014, opened 2024-03-21
Qual: Add dependencies between workflows

mdeweerd:
# Qual: Add dependencies between workflows

Run costly workflows only after a few workflows succeed (pre-commit, phan).

@eldy 
Here you have it!  Dependencies between workflows.

In practice it has to be one workflow, but I created 'ci.yml' to refer to the others.

ci.yml defines the dependencies as well.

gh-travis.yml tries to run the job on a github runner - it starts but fails at the point where it touches a file:
```
: touch(): Unable to create file /home/runner/.config/composer/config.json because Permission denied
```

Maybe you know why - I disable it for now.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,16 @@
+# Workflow run order
+
+To reduce run minutes, the following order is put in place:
+
+On PR & Merge, always run:
+
+- pre-commit;
+- phan.
+
+When both succeed, start:
+
+- phpstan;
+- Windows-ci;
+- travis.
+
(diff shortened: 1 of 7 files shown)


eldy:
>Maybe you know why - I disable it for now.

No, a file already exists ? Permission on parent dirs ?

mdeweerd:
> > Maybe you know why - I disable it for now.
> 
> No, a file already exists ? Permission on parent dirs ?

Ok, I may retry this at a later time & read the script in more detail;

PR #29015, opened 2024-03-21
new: upgrade minimum version of dolibarr for module builder

FHenry:
Module builder use feature than are not available in version 11. Seems reasonable to upgrade it at least to version 17 of dolibarr


sonikf:
Hi @FHenry 
I've just noticed that code generated with module builder in v.19 breaks in v18
e.g. in actions_mymodule.class.php

```
require_once DOL_DOCUMENT_ROOT.'/core/class/commonhookactions.class.php';

/**
 * Class ActionsMyModule
 */
class ActionsMyModule extends CommonHookActions
```

FHenry:
> Hi @FHenry I've just noticed that code generated with module builder in v.19 breaks in v18 e.g. in actions_mymodule.class.php
> 
> ```
> require_once DOL_DOCUMENT_ROOT.'/core/class/commonhookactions.class.php';
> 
> /**
>  * Class ActionsMyModule
>  */
> class ActionsMyModule extends CommonHookActions
> ```

Yes , that a fact with the module builder, each version is compatible only with the version of dolibarr design for.
In version 19, there is a new abstract class CommonHookActions that do not exists previous version.
The version comptability with module builder is better from previous version to upper, I mean make a module with module builder in version 17, it will still works in version 19.

PR #29016, opened 2024-03-22
FIX Add CSV download link to stock at date table

bradley-jarvis:
This update adds a csv download link to the stock at date list

### DIFF
--- a/htdocs/product/stock/.htaccess
+++ b/htdocs/product/stock/.htaccess
@@ -1,7 +1,0 @@
-RewriteEngine On
-
-#RewriteCond %{HTTP:X-Forwarded-Proto} !https
-#RewriteRule .* https://%{HTTP_HOST}%{REQUEST_URI} [R,L]
-
-#RewriteRUle ^product/stock/stockatdate\.csv$ product/stock/stockatdate.php  [L]
-RewriteRUle ^stockatdate\.csv$ stockatdate.php  [L]
--- a/htdocs/product/stock/.htaccess
+++ b/htdocs/product/stock/.htaccess
@@ -1,0 +1,7 @@
+RewriteEngine On
+
+#RewriteCond %{HTTP:X-Forwarded-Proto} !https
+#RewriteRule .* https://%{HTTP_HOST}%{REQUEST_URI} [R,L]
+
+#RewriteRUle ^product/stock/stockatdate\.csv$ product/stock/stockatdate.php  [L]
+RewriteRUle ^stockatdate\.csv$ stockatdate.php  [L]
--- a/htdocs/product/stock/stockatdate.php
+++ b/htdocs/product/stock/stockatdate.php
@@ -247,8 +247,8 @@ if ($date && $dateIsValid) {
 			$movements_prod_warehouse_nb[$fk_product][$fk_entrepot] = $nbofmovement;
 
 			// Pour llx_product.stock
-			$movements_prod[$fk_product] += $stock;
-			$movements_prod_nb[$fk_product] += $nbofmovement;
+			$movements_prod[$fk_product] = $stock + (array_key_exists($fk_product, $movements_prod)?$movements_prod[$fk_product]:0);
+			$movements_prod_nb[$fk_product] = $nbofmovement + (array_key_exists($fk_product, $movements_prod_nb)?$movements_prod_nb[$fk_product]:0);
 
 			$i++;
 		}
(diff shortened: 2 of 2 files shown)


htdocs/product/stock/.htaccess
```diff
@@ -0,0 +1,7 @@
+RewriteEngine On
```
eldy, line 1: Adding  files with specific setup to an external software and environment variable is not portable.
Can you remove this file.

htdocs/product/stock/stockatdate.php
```diff
@@ -353,164 +359,187 @@
...
+	if (!empty($search_fk_warehouse)) {
+		foreach ($search_fk_warehouse as $val) {
+			$param_warehouse .= '&search_fk_warehouse[]='.$val;
+		}
+		$param .= $param_warehouse;
+	}
+	if ($productid > 0) {
+		$param .= '&productid='.(int) $productid;
+	}
+	if (GETPOST('dateday', 'int') > 0) {
```
eldy, line 454: Why replacing GETPOSTINT int GETPOST(... , 'int') ?

htdocs/product/stock/stockatdate.php
```diff
@@ -353,164 +359,187 @@
...
-print '<td class="liste_titre"></td>';
-print '<td class="liste_titre"></td>';
-print '<td class="liste_titre"></td>';
-if ($mode == 'future') {
+	// TODO Move this into the title line ?
+	print_barre_liste('', $page, $_SERVER["PHP_SELF"], $param, $sortfield, $sortorder, '', $num, $nbtotalofrecords, 'stock', 0, '', '', $limit, 0, 0, 1);
+
+	print '<div class="div-table-responsive">'; // You can use div-table-responsive-no-min if you don't need reserved height for your table
+	if ($num) {
+		print "<p><a href=\"stockatdate.csv?sortfield=$sortfield&sortorder=$sortorder&type=$type&mode=$mode".
```
eldy, line 469: A dynamic page must end with .php, using another extension will generate troubles when using proxy or cache server.
You can add the parameter  &output=file  to provide the information you want a download.

bradley-jarvis:
That now makes sense why you don't use mod rewrite anywhere else, updated code as requested. The GETPOSTINT/GETPOST revert was an accident and also now fixed.

htdocs/product/stock/stockatdate.php
```diff
@@ -353,164 +359,187 @@
...
+		}
+		$param .= $param_warehouse;
+	}
+	if ($productid > 0) {
+		$param .= '&productid='.(int) $productid;
+	}
+	if (GETPOSTINT('dateday') > 0) {
+		$param .= '&dateday='.GETPOSTINT('dateday');
+	}
+	if (GETPOSTINT('datemonth', 'int') > 0) {
```
eldy, line 457: With GETPOSTINT, there is no need for the second parameter 'int'.

htdocs/product/stock/stockatdate.php
```diff
@@ -582,135 +611,178 @@
...
+		print '<tr><td colspan="'.$colspan.'"><span class="opacitymedium">'.$langs->trans("EnterADateCriteria").'</span></td></tr>';
 	} else {
+		print '<tr class="liste_total">';
+		print '<td>'.$langs->trans("Totalforthispage").'</td>';
 		print '<td></td>';
-		print '<td class="right">'.price(price2num($totalbuyingprice, 'MT')).'</td>';
-		if (!getDolGlobalString('PRODUIT_MULTIPRICES')) {
-			print '<td class="right">'.price(price2num($totalsellingprice, 'MT')).'</td>';
+		if ($mode == 'future') {
+			print '<td class="right">'.$totalcurrentstock.'</td>';
```
eldy, line 755: The $totalcurrentstock is a result of operation between float value so result is a float with a non exact decimal part. For this reason me must use price(price2num($totalcurrentstock, 'MS'))
price2num to trunc the decimal part that is not relevant and introduced by float number calculation. Then price to format the number into a number using the language of the user.

bradley-jarvis: This would be the same for the totalvirtualstock value as well?

eldy: > This would be the same for the totalvirtualstock value as well? what is the difference beween the 'MS' and 'MT' value?

Yes same.MT is when we need to trunc decimal part for an amount, MS is used for a stock.

bradley-jarvis: That makes sense, I have now wrapped all the stock values after initially missing some. Hopefully this sorts it out

bradley-jarvis:
I have made the changes requested, so this should be all good to go now.

PR #29021, opened 2024-03-22
Fix: Incorrect label in Gantt

fmarcet:
The label name was not displayed correctly and the text 'undefined' was displayed


hregis:
@fmarcet it's not possible to change elsewhere for to have "duration" and "resource" instead "dur" and "res" ?

hregis:
@fmarcet it would be more meaningful no ?

fmarcet:
@hregis the translations are in that same file. I thought it was easier to change the two places where the variable is used than to change the text in the translations of all the languages. I also wasn't sure if it was used somewhere else and could cause something else to stop working. But if you consider it appropriate, I can make the change as you say.

hregis:
@fmarcet it's just to better to understand, "duration" is more explicit than "dur"... jsgantt.js is an external library? you have to propose the modifications to the developers, right?

fmarcet:
@hregis I have already made the change. But I don't know how the issue of external libraries is managed in Dolibarr, or if I should notify the original developers of the library. If you consider it appropriate, you can do it yourself.

eldy:
We upgrade sometime the lib from the upstream source.
I have downloaded the last version. It seems your bug fix is included. thanks to check...

PR #29022, opened 2024-03-22
New msg attachment linked to event

MaximilienR-easya:
# New
Now the linked files inside ticket messages are linked to the event of the message and not to the ticket.
With this change, we can see the attached files to the message they are linked inside the message history

test

This PR is needed for another PR [*#24872*]


lvessiller-opendsi:
@eldy 
Can you have a look please ?

PR #29030, opened 2024-03-22
FIX  #29029 Impossible to delete an order line

### LINKED ISSUE #29029: Impossible to delete an order line
### Bug

In an order when you try to delete a line, you go through the normal Dolibarr process with the form confirm but the line is never deleted

### Dolibarr Version

18

### Environment PHP

7.4

### Environment Database

_No response_

Add a line in an order and try to delete it

### Attached files

_No response_

Delthair:
FIX #29029  Impossible to delete a line because the code lead to add an error if the request is valid therefore leading to a rollback of the request


bradley-jarvis:
This needs to be applied to the develop branch too

hregis:
@bradley-jarvis please be patient ;-)

PR #29034, opened 2024-03-23
Fix & enable PhanTypeInvalidRightOperandOfNumericOp

mdeweerd:
# Fix & enable PhanTypeInvalidRightOperandOfNumericOp

This fixes almost all the PhanTypeInvalidRightOperandOfNumericOp noticies and several PhanTypeInvalidLeftOperandOfNumericOp notices.

This is built on #29006 which also fixes some of the notices (side effects/fixes at the same time).

A few notices were left unfixed to leave them for further investigation.
In particular Cronjob->$unitfrequency is defined as a string and treated as a string (escaped, etc), but used as a number in formulas.  I think that should be changed to be a number.

I can add them to the phan baseline (to ignore them for now).


mdeweerd:
@eldy I think that you should create a PR to deprecate the fetch method and fix all exceptions there before applying it on the main branch.
It's hard to see if all exceptions for the other PRs are fixed.

eldy:
> @eldy I think that you should create a PR to deprecate the fetch method and fix all exceptions there before applying it on the main branch. It's hard to see if all exceptions for the other PRs are fixed.

Sorry, i wasn't aware that it was no more possible to depreciate a method without triggering errors. It seems we must also add a phan comment at each call of the deprecated function when we depreciate a new method ? No way to depreciate in an easier manner ?

mdeweerd:
> Sorry, i wasn't aware that it was no more possible to depreciate a method without triggering errors. It seems we must also add a phan comment at each call of the deprecated function when we depreciate a new method ? No way to depreciate in an easier manner ?

There are three methods to disable these notices (other than fixing them):
1. Disable the detection for all occurrences;
2. Disable the detection on a file per file basis (through the baseline (baseline.txt).
3. As you indicate: add an exception for each violation.

The problem with 1. and 2. is that that his disables potentially valid notices for deprecations that were already active and fixed.  All occurences will still be detected through the analysis without the baseline (as available on the cti page).

If you want to introduce deprecation in a "mild" way, then maybe a change to the code before running phan & phpstan - for instance, the comment could read "deprecation-planned" and that could be replaced with "@deprecated" just for the cti report generation.

Issue #25738, opened 2023-08-30
Notifications are generated before the triggering process is completed


jacotec:
### Bug

The "processing hook" of the module seems to be too early:

* When an invoice is validated, the checkbox "append file" in the email template does nothing. The notification seems to be generated before the final invoice is completely generated.

* The Reference still gives back the "PROV" number, not the final invoice number

* When a notification is triggered by a completed payment, the attached document is still the original, old invoice - not the new version with the payment remark.

* When a subscription triggers an automatic validated invoice, there is also no document attached. In this case, even the public download link placeholder is empty - so in this case no invoice at all can be sent to the customer automatically.

As a conclusion: The notification should be processed after all the things which had triggered the notification are completely processed. Or the notification should be in general delayed by a few seconds, so it can be assured that all the needed processes are completed before the notification is generated.

### Environment Version

18.0.0

### Environment OS

Ubuntu 22.04

### Environment Web server

NGINX

### Environment PHP

8.1

### Environment Database

MariaDB 11

### Expected and actual behavior

See above. The notification should be generated after all underlying processes which are triggering the notification are completed, so that it can be assured that all variables, links and attached files are already up-to-date.

Maybe a solution is to add a delay of a few seconds before the notification task is triggered, so that it can be assured that all underlying variables and documents are processed and up-to-date before generating the notification.

### Steps to reproduce the behavior

Follow the procedure in the description and i.e. validate an invoice which sends a notification to the customer.

The reference is still the previois "PROV" number, the file is not attached (as the invoice is not yet finally generated when the notification is triggered).

narkin:
I can confirm that this is still happening in 18.0.2

JonBendtsen:
This issue just bite me when it comes to stripe payments because the automatic notification sends this url:
```
/dolibarr/public/payment/newpayment.php?source=order&ref=%28PROV301%29&securekey=...
```

where as if I send the email manually then the url are:
```
/dolibarr/public/payment/newpayment.php?source=order&ref=SO2311-0065&securekey=...
```

I wish this bug would be prioritized higher, as it seems to affect more than just one place, this issue is most likely related
https://github.com/Dolibarr/dolibarr/issues/26396

![broken_payment](https://github.com/Dolibarr/dolibarr/assets/2551481/bc00128f-9f3b-40b6-b83c-e9234dd813f0)
[...]

PR #29041, opened 2024-03-23
New notification trigger for post order validation

JonBendtsen:
NEW notification trigger ORDER_POST_VALIDATE for post order validation notifications

The still existing trigger ORDER_VALIDATE is triggered pretty early in the validation process. This caused issues in the emails in that some substitution variables would not have content or contain the wrong content like the `__REF__` which would still have the old name. Also, the PDF was not available for attachment.

This pull request introduces a new notification trigger called ORDER_POST_VALIDATE which is executed after the order actually has both been validated and committed to the database. This means that the substitution variable `__REF__` & others now contains the correct name and that the file is actually attached in the email. 

Known issues:
The document still needs to be re-generated before sending.


JonBendtsen:
This pull request is not a complete fix of #25738 because this pull request only applies to orders, not proposals, invoices, ... or anything else that can be validated.

Also, the document attached should be regnerated.

However, if accepted, I'll probably slowly introduce similar pull requests for proposals, invoices and possible members - because I use those modules.

JonBendtsen:
Here is the new message in Dolibarr when validating an order

pre_n_post_validation_emails

JonBendtsen:
The emails looks like this:

mutt_2_emails

ORDER_VALIDATE
mutt_order_validate

ORDER_POST_VALIDATE
mutt_order_post_validate

JonBendtsen:
This works like any other notifications, I need to enable them before usage

thirdparty_notifications


And I need to configure some email templates
email_templates

Plus I need to select those templates in the notification module configuration
notification_module

htdocs/commande/class/commande.class.php
```diff
@@ -596,6 +596,14 @@ public function valid($user, $idwarehouse = 0, $notrigger = 0)
 
 		if (!$error) {
 			$this->db->commit();
+			if (!$error && !$notrigger) {
+				// Call trigger
+				$result = $this->call_trigger('ORDER_POST_VALIDATE', $user);
```
eldy, line 601: The goal of a trigger is to be inside a transaction so it can be canceled by the rollback so no trigger must be after commit()

JonBendtsen: > The goal of a trigger is to be inside a transaction so it can be canceled by the rollback, so, no trigger must be after commit(). Never. 

ok. What if you want to be "notified", informed, messaged, ... about an after post transaction fact? Sort of like "heads up, this thing happened, deal with it"

Sometimes there is a value in getting a post happening "call to action". In this case it can be emailing the customer with new information about their purchase status. And it certainly before emailing it could be to regenerate the document.

> You can retrieve the new ref into property this->newref Generation of links that are using the old ref should be instead enhanced to user ->newref instead of ->ref when it is set. Can you work on a fix going in this direction ?

Perhaps, I'll have to look at it. Initially I don't really like that situation of just using different variables, and I don't think it makes good sense to move the post trigger to JUST before the commit, because it makes little sense to email the customer "accepted" when some other thing that listens to this trigger can cancel and then everything has been rolled back.


So how would you bring about a post transaction "trigger"/information/... ? (It could be named something like: Alleviate, Assauge, Commence, Debrief, Decommision, Mitigate, Mollify, Palliate, Placate, ...)

Would you make a whole duplication of the trigger system? but only for things that happened after a transaction has been committed?

JonBendtsen: I am not testing in latest development commit, it is a tuxgasy v19 container, but here's my comments on that

> You can retrieve the new ref into property this->newref Generation of links that are using the old ref should be instead enhanced to user ->newref instead of ->ref when it is set. Can you work on a fix going in this direction ?

I just looked at what substitution variables is available when manually sending an order email, and here `__NEWREF__` is not available, only `__REF__` so I'll have to need 2 different email templates? one for manual sending and one for automatic notifications?

I also looked at this order using the API, and here the json does not contain any `newref`, only `ref` - and `"oldref": null,`


So those links that you say have to be updated to use `->newref instead of ->ref` - can I trust that ->newref is always available? What about substitution variables in email templates, will I have to use different variables here too depending if I'm sending manually or sending using the automatic notification system?


Sorry sir, but I don't really like this "solution" either. I wish that the only difference between manually send emails and automatic send emails would be the sender, I don't wish to have multiple templates or to use different variables in the json depending on it's status if it is draft or validated - or has been validated.

Also, there's the whole issue of potentially having 2 or more different subsystems that listens to the ORDER_VALIDATE trigger, does something based on that - like reserving a product/resource for something or emails the customer, ... and then the last thing that listens to the ORDER_VALIDATE trigger says no, because it is my understanding that they can say no and interrupt the action about to happen. Then what? automatic notifications already send the email.

eldy: > What if you want to be "notified", informed, messaged, ... about an after post transaction fact? Sort of like "heads up, this thing happened, deal with it"

The trigger ORDER_VALIDATE is already executed at the end of the transaction. It is the last business action that has effect into database. So you should be able to use it for post action. All actions after are technical change that should not be modified by external code, like updating cache table.

the ->newref exists during the life of the trigger. So you can get it only inside code executed by a trigger.
When using an api, you will get only ref that is final ref.

You don't need several template. A template should use __REF__ . If the __REF__ does not contains the new ref when making the substitution, it is was i said when i say "Generation of links that are using the old ref should be instead enhanced to use ->newref instead of ->ref when it is set." 
It means there is something missing in the existing code in the core that make the substitution of __REF__. The current code does not take into account the ->newref field when it should.

JonBendtsen: Hi

Do you have 2 code examples? One where the code uses ->newref and one where it does not use ->newref but it should?

> The ->newref exists on object during the life of the trigger. So you can get it only inside code executed by a triggers.

How does the code know if it is being executed inside a trigger?  Could you provide an example? then I can see if I can understand it.


> You don't need several template. A template should use `__REF__` (You should never have `__NEWREF__` inside a template, this key has no sense as a template is a user data and old ref / new ref is something managed on developer side). If the `__REF__` does not contains the new ref when making the substitution of this key, it is was i said when i say "Generation of links that are using the old ref should be instead enhanced to use ->newref instead of ->ref when it is set.". I mean this: There is something missing in the existing code in the core that make the substitution of `__REF__`. The current code does not take into account the ->newref field when it should.

If ->newref is supposed to be hidden, how come it showed up in the list of substitution variables in the emails I attached pictures of above?

eldy: Where the code make the substitution __REF__ with ->ref, the code must instead do __REF__ replaced with ->newref (if ->newref is set).

JonBendtsen: > Where the code make the substitution `__REF__` with ->ref, the code must instead do `__REF__` replaced with ->newref (if ->newref is set).

yeah? but where? I'm leaning towards closing this pull request and letting someone else fix the ref/newref quest.

RAF0846:
Hello,
First at all, sorry for my english!
I have the same issue with automatic email notification on 20.0.1 version.
The PDF order confirmation sent to customer is not the validated one.
It is the draft one.

PR #29042, opened 2024-03-24
FIX update bom product/service tree

bradley-jarvis:
This patch fixes the BOM view to include all nested Sub BOMs in the bom_card Product BOM and Service BOM lists, not just the immediate children. It includes a significant rewrite of the objectline_view.tpl.php which includes a recursive function to work through the nested BOM's.

The function takes better advantage of the calculateCosts function output and in particluar the $line->childBom variable(this variable was incorrectly being defined as an array of BOM but my understanding is that only one BOM can be assigned to a product), the costs displayed in the template are all calculated in that function instead of within the template as the old code did.

This update also fixes the quantity and price amounts displayed in the list, correctly taking into account the quantity of product created in the sub BOM.

Also if no cost value is added to a service the best buy price is used, this is the same as for product.

I also want to add a piece unit value to service along with all the time units. The reason is that we make a product that we own the stock used but pay another company to mould the part. This is tracked as a service that we buy and the price we pay is a per item piece quantity not a time.


htdocs/bom/class/bom.class.php
```diff
@@ -586,7 +585,7 @@ public function delete(User $user, $notrigger = 1)
 	 * @param	int		$position				Position of BOM-Line in BOM-Lines
 	 * @param	int		$fk_bom_child			Id of BOM Child
 	 * @param	string	$import_key				Import Key
-	 * @param	int 	$fk_unit				Unit
+	 * @param	int		$fk_unit				Unit
```
mdeweerd, line 588: This should actually be int<0,1> as in addLine().

htdocs/bom/class/bom.class.php
```diff
@@ -689,7 +688,7 @@ public function addLine($fk_product, $qty, $qty_frozen = 0, $disable_stock_chang
 	 *
 	 * @param 	int		$rowid					Id of line to update
 	 * @param	float	$qty					Quantity
-	 * @param	float	$qty_frozen				Frozen quantity
+	 * @param	flloat	$qty_frozen				Frozen quantity
```
mdeweerd, line 691: This should actually be int<0,1> as in addLine()

htdocs/bom/class/bom.class.php
```diff
@@ -1789,24 +1789,25 @@ class BOMLine extends CommonObjectLine
...
 	/**
-	 * @var array     array of Bom in line
+	 * @var BOM				 Reference to Bom in line
 	 */
-	public $childBom = array();
+	public $childBom = null;
 
 	/**
-	 * @var int|null                ID of the unit of measurement (rowid in llx_c_units table)
+	 * * @var int|null		ID of the unit of measurement (rowid in llx_c_units table)
```
mdeweerd, line 1807: One '*' too many.

htdocs/bom/tpl/objectline_view.tpl.php
```diff
@@ -77,300 +75,139 @@
...
-					print $langs->trans($label);
-				}
-				print '</td>';
+		// Cost
+		$column[] = '<td id="costline_'.$data->id.'" class="linecolcost nowrap right"><span class="amount">'.
+			price(price2num($data->total_cost*$quantity, 'MT')).'</span></td>';
+
+		if ($data->status == 0 && ($object_rights->write) && $action != 'selectlines') {
+			$column[] = '<td class="linecoledit center">'.
+				(($line->info_bits & 2) == 2 || !empty($disableedit)?'':
```
mdeweerd, line 163: If $line is provided from the function that includes the template (tpl), then you can add
```php
'@phan-var-force BOMLine $line';
```
somewhere near the head of the file (BOMLine being replace with the appropriate type of course).

bradley-jarvis: Thanks for having a look, I actually copy and pasted that back in but because it is inside the function I renamed $line to $data which in hindsight wasn't the best idea, I have commited a fix which corrects the $line to $data. I did test and it appears to work correctly.

htdocs/bom/class/bom.class.php
bradley-jarvis: Yep I saw that one, I did start on this patch a week or so ago, before your commit came in, it was a bit of a pain to re-integrate and I was doing a bit of copy/paste. It also messed around with the tabs/spaces.

mdeweerd: The easiest way to handle tabs/spaces is to setup phpcbf which is done automatically when you setup pre-commit.

pip install pre-commit
Then in the local clone: pre-commit install

And pre-commit will run on every commit doing some checks and making some style corrections.

bradley-jarvis:
I am nut sure why PHAN is complaining about the latest commit, are you able to shed some light on why it has an issue

Below is how I am calling print_line and below that is the definition of the function and parameters. The code is working fine and correctly prints out multiple levels of BOM recursion.

print implode('', print_line($line, 1, 0, $this));

/**
* Recursively loop through and print BOM lines
*
* @param  CommonObjectLine $data	BOMLine to print on row
* @param  float $quantity			Quantity modifier for sub BOM
* @param  int $level				Level of recursion
* @param  CommonObject $parent		Parent BOMLine ID, used for show/hide/edit/delete
* @return array					Return array of html rows
*/
function print_line($data, $quantity, $level, $parent)

mdeweerd:
> I am nut sure why PHAN is complaining about the latest commit, are you able to shed some light on why it has an issue
> * @param  CommonObjectLine $data	BOMLine to print on row
> * @param  float $quantity			Quantity modifier for sub BOM

The code is: 

```
print_line($child, ($quantity * $data->qty) / (($bom->qty??1) * $data->efficiency), $level + 1, $data)
```

Phan notices the use of $data in argument 2 and/or 4 while the function's signature uses $data in argument 1.

The technique I prefer is to rename my local variables or to add intermediate variables.
For instance:
$quantity_per_line = (FORMULA);
$parent = $data;

or, rename $data so that it is more meaningful (we do not know what data is just by looking at the line.

This Phan rule generally helps me when I forget a parameter on the line, mixed some parameters up, or changed from one function to another not noticing that their is a change in the signature.

(Another technique is to add a @phan-suppress-next-line ... ).

bradley-jarvis:
Ok all sorted now, the current PHAN warning comes from commit [4efa503](https://github.com/Dolibarr/dolibarr/commit/4efa50340024d2ff859cc9767164764e484daebe), specifically the line 

$ret = str_replace('__USER_ID__', (int) $user->id, $ret);

which is not related to this PR

htdocs/bom/class/bom.class.php
bradley-jarvis: Thanks for this, I now have pre-commit  and phpcs/phpcbf

htdocs/bom/class/bom.class.php
```diff
@@ -486,13 +486,13 @@ public function fetchLinesbytypeproduct($typeproduct = 0)
...
-	 * @param  string   		$filter       Filter USF
-	 * @param  string      		$filtermode   Filter mode (AND or OR)
-	 * @return array|int        			  int <0 if KO, array of pages if OK
+	 * @param  string               $sortorder    Sort Order
+	 * @param  string               $sortfield    Sort field
+	 * @param  int                  $limit        Limit
+	 * @param  int                  $offset       Offset
+	 * @param  string               $filter       Filter USF
+	 * @param  string               $filtermode   Filter mode (AND or OR)
+	 * @return array|int                              int <0 if KO, array of pages if OK
```
mdeweerd, line 495: For static analysis, we need to specify the fields of the array return value.
I suspect it it the following
```
* @param int<-1,-1>|BOM[]
```

htdocs/bom/class/bom.class.php
```diff
@@ -1324,7 +1323,7 @@ public function generateDocument($modele, $outputlangs, $hidedetails = 0, $hided
 	}
 
 	// phpcs:disable PEAR.NamingConventions.ValidFunctionName.ScopeNotCamelCaps
-	/**
+	 /**
```
mdeweerd, line 1326: I am surprised about the spacing.

htdocs/bom/class/bom.class.php
```diff
@@ -1470,6 +1465,17 @@ public function calculateCosts()
 					$unitforline = measuringUnitString($line->fk_unit, '', '', 1);
 					$qtyhourforline = convertDurationtoHour($line->qty, $unitforline);
 
+					$line->unit_cost = price2num((!empty($tmpproduct->cost_price)) ? $tmpproduct->cost_price : $tmpproduct->pmp);
+					if (empty($line->unit_cost)) {
+						if ($productFournisseur->find_min_price_product_fournisseur($line->fk_product) > 0) {
+							if ($productFournisseur->fourn_remise_percent != "0") {
+								$line->unit_cost = $productFournisseur->fourn_unitprice_with_discount;
```
mdeweerd, line 1472: I did not check, but if fourn_unitprice_with_discount is up-to-date even when there is no discount (i.e., fourn_unit_price == fourn_unitprice_with_discount when there is no discount), then unit_price can always use fourn_unitprice_with_discount without any specific test.

htdocs/bom/class/bom.class.php
```diff
@@ -1789,24 +1789,25 @@ class BOMLine extends CommonObjectLine
...
 
 	/**
-	 * @var float		Line unit cost based on product cost price or pmp
+	 * @var float			Line unit cost based on product cost price or pmp
 	 */
 	public $unit_cost = 0;
 
 	/**
-	 * @var array     array of Bom in line
+	 * @var BOM				 Reference to Bom in line
```
mdeweerd, line 1802: Should be ?BOM as the value can be null.

htdocs/bom/class/bom.class.php
```diff
@@ -1789,24 +1789,25 @@ class BOMLine extends CommonObjectLine
...
 
 	/**
-	 * @var float		Calculated cost for the BOM line
+	 * @var float			Calculated cost for the BOM line
 	 */
 	public $total_cost = 0;
 
 	/**
-	 * @var float		Line unit cost based on product cost price or pmp
+	 * @var float			Line unit cost based on product cost price or pmp
```
mdeweerd, line 1797: Should be 'float|int' or initialised to 0.0;

mdeweerd (approved):
I do not see anything critical ("level 1" review=no algorithm review).
I made some comments that I would make.

htdocs/bom/class/bom.class.php
bradley-jarvis: Ok, for I actually just copy and pasted from the product cost part of code from the block above. The problem I was having with this was that the service cost was 0 because I use buying price (thirdparty supplies the service). I will dig a bit deeper and check

htdocs/bom/class/bom.class.php
bradley-jarvis: I didn't change that one ;) I'm guessing the comment above confused it

htdocs/bom/class/bom.class.php
bradley-jarvis: This is not part of the PR, I am guessing that phpcs/phpcbf made this change to the spacing for some reason... tabs to spaces. But your analysis looks right.

mdeweerd: Sometime phpcbf is not doing the job as expected.
If you update it it could show other notices in other code leading to more changes to "fix" phan.

bradley-jarvis:
Just some context on this update and what it fixes, the top image is current develop branch and the bottom image is with patch applied (plus an extra for being able to set the service units to quantity, we pay an external company per unit for moulding)

This is what the current develop branch looks like when I view a BOM with Sub BOM, notice that the quantities of the sub BOM do not take into account the efficiency. The cost is way out, we are a plastics moulding company and we measure the plastic additives in parts or percentage. So the TPU-WT bom includes the white additive at 1% to 100% TPU with Dark Laser Additive  Which is also a sub BOM (note that this bom makes 101kg of material, 1kg colour and 100kg tpu mix). The TPU with Laser additive has 100kg (or %)), 4.5kg additiva A, 3kg additive B and 1kg additive C to make a total BOM quantity of 108.5kg. The current code does not use the unit amount of a sub bom, but the total, which is why we end up with a cost of $822,009 for essentially 5.3kg (8.7kg if mfg efficiency is taken into account)

Screenshot from 2024-03-29 15-00-16

This patch fixes the use of units of sub bom and also lists the price based on manufacturing efficiency because that is importand in calculating actual margin (and will be great info to have to justify recycling of waste product)

Screenshot from 2024-03-29 15-19-16

bradley-jarvis:
I have been working on adding proper sub BOM support into the MO module and in the process found a possible issue with the way unit translations were being done. The issue became apparent when I added qty as an option unit and 'Piece' was being translated into Accounting Doc. which is defined in lang/compta.php (or something like).

This update hopefully sorts out some of the format specified above.

I also am not sure about the correct way to go with this but at the moment I have updated the code to use the sub BOM unit price when calculating costs only when BOM_USE_PRICE_FOR_COSTING is set to 1 as a quick fix so that the default pricing works as is (when BOM_USE_PRICE_FOR_COSTING not set). I think it might be worthwhile for the user to be able to prioritise pricing used or choose highest or lowest.

When checking the calculateCost function i notice that at the end the unit price was being updated different for -ve quantity. I cannot for the life of me think of why you would have a negative quantity on a BOM production, and that the -ve quantity would make the unit cost a multiple of the total cost??? I removed this and added a check in create and update on BOM and BOMLine that will not allow a <=0 quantity.

mdeweerd:
@bradley-jarvis FYI: I am not a member of the core team, just decided to help out a bit recently mostly with helping improve quality.

I can think of a reason for negative qty: reusable "scrap"/"waste"/replaced components.
For instance, I could buy an apparatus that is needed to make an assemble and it comes with a power adapter.  However the assembly requires a more powerful adapter or there is already one from another device.  Therefore, after assembly I have one unused adapter that could serve another purpose: after sales, other assemblies, ... .
Another example would be joints.

bradley-jarvis:
@mdeweerd  No problem, I have appreciated the comments. Hopefully someone in the core team have another look. It does fix a core issue with sub boms and there usefulness.

There is a dissassembly BOM option which consumes one product to make multiple. And the unit price = total_price * quatity makes no sense at all (note that the unit price also becomes negative, where-as the total price is still positive). I think at the lease that line of code needs a comment.

eldy:
Do you know you alread have a tree view into tab "BOM net needs", click on the picto tree view.

bradley-jarvis:
Yep @eldy , but it only displays the direct childs of the BOM, this update recursively displays the child->child->child and so on for as many levels as you have. You can see this in the pictures above, the TPU-DARK is a child BOM of a child BOM (TPU-WT). In the original it does not recurse into this, but in the updated code it does. This patch also fixes the child BOM quantities, you can also see this from the example pictures above where the original code uses the total quantity and not the unit quantity, which is in my opinion a bug.

The updated code also uses the calculateCost function for generating the product costs in the values displayed in the list template, the original code recalculates the child BOM costs used in the template, so leads to a greater chance of discrepancies.

This fix is the first part of another fix that I am currently testing in a live system where the MO also recurses through the child BOM's and adds them as product to consume without having to make MO's for all the child products which we would consider virtual product.

As a side note, I did notice that I have to fix the <=0 quantity update that I last committed to this PR, it is currently checking quantity and it should be qty.

PR #29047, opened 2024-03-24
Fixes for PhanPluginUnknownObjectMethodCall

mdeweerd:
# Fixes for PhanPluginUnknownObjectMethodCall

This fixes several PhanPluginUnknownObjectMethodCall notices which transform most of the time ito a PhanUndeclaredMethod because the parent class does not declare the abstract method that is supposed to be available on the child classes (write_file, getNextValue, getExample, ... .

The issue for  member.php is real - the related getToolTip method only takes 2 arguments
```
PhanParamTooMany: Call with 3 arg(s) to \ModeleNumRefMembers::getToolTip(\Translate $langs, \Societe $soc) which only takes 2 arg(s) defined at htdocs/core/modules/member/modules_member.class.php:74
```

@eldy: you know that in the CTI log there are notifications for htdocs/custom/mymodule which are not part of this repository?!

### DIFF
--- a/dev/tools/phan/baseline.txt
+++ b/dev/tools/phan/baseline.txt
@@ -24,6 +24,7 @@ return [
 
     // Currently, file_suppressions and directory_suppressions are the only supported suppressions
     'file_suppressions' => [
+        'htdocs/adherents/admin/member.php' => ['PhanParamTooMany'],
         'htdocs/adherents/type.php' => ['PhanPluginDuplicateExpressionBinaryOp'],
         'htdocs/admin/receiptprinter.php' => ['PhanRedefineFunctionInternal'],
         'htdocs/api/class/api_documents.class.php' => ['PhanPluginDuplicateExpressionBinaryOp'],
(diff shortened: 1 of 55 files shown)


eldy:
> # Fixes for PhanPluginUnknownObjectMethodCall
> 
> This fixes several PhanPluginUnknownObjectMethodCall notices which transform most of the time ito a PhanUndeclaredMethod because the parent class does not declare the abstract method that is supposed to be available on the child classes (write_file, getNextValue, getExample, ... .
> 
> The issue for  member.php is real - the related getToolTip method only takes 2 arguments
> ```
> PhanParamTooMany: Call with 3 arg(s) to \ModeleNumRefMembers::getToolTip(\Translate $langs, \Societe $soc) which only takes 2 arg(s) defined at htdocs/core/modules/member/modules_member.class.php:74
> ```
> 
> @eldy: you know that in the CTI log there are notifications for htdocs/custom/mymodule which are not part of this repository?!

No, wasn't aware. The .gitignore should have avoid this. Will check what happened...

PR #29050, opened 2024-03-25
FIX: last arg is for showempty, not for a translation / label ...

rycks:
That function is

```
public function select_categories($type, $selected = 0, $htmlname = 'search_categ', $nocateg = 0, $showempty = 1, $morecss = '')
```

Then i don't know why showempty was set from a translation ...


rycks:
arg ! phpcs / something appends with spaces / tabs / i will change the commit please wait a little

PR #29056, opened 2024-03-25

thomas-Ngr:

This PR:

SQL Part : #26190


htdocs/product/class/product.class.php
```diff
@@ -620,6 +627,21 @@ class Product extends CommonObject
 	 * Service
 	 */
 	const TYPE_SERVICE = 1;
+	/**
+	 * Advanced feature: assembly kit
+	 */
+	const TYPE_ASSEMBLYKIT = 2;
```
eldy, line 633: It seems this is not related to the feature stockable or not.
Thanks to split this other feature into a different PR.

thomas-Ngr:
@eldy I have removed the only faulty line, can you have a look ?

thomas-Ngr:
follow-up in #30230

PR #29071, opened 2024-03-26
NEW Constant to select if typent helps define whether the thirdparty is a company

aspangaro:
When you create an european customer and if you define in type company it's a company... But the customer has not provided you with its intra-Community VAT number.

You need to apply normal VAT rate, no exemption because you have define it's a company in a simple list.

image

I introduce a constant DEFINE_CUSTOMERS_ARE_COMPANIES_BY_TYPE_COMPANY to avoid this behavior

### DIFF
--- a/htdocs/societe/class/societe.class.php
+++ b/htdocs/societe/class/societe.class.php
@@ -4083,7 +4083,7 @@ class Societe extends CommonObject
 		} elseif (!empty($this->idprof1) || !empty($this->idprof2) || !empty($this->idprof3) || !empty($this->idprof4) || !empty($this->idprof5) || !empty($this->idprof6)) {
 			$isacompany = 1;
 		} else {
-			if (!empty($conf->global->DEFINE_CUSTOMERS_ARE_COMPANIES_BY_TYPE_COMPANY)) {
+			if (getDolGlobalString('DEFINE_CUSTOMERS_ARE_COMPANIES_BY_TYPE_COMPANY')) {
 				// TODO Add a field is_a_company into dictionary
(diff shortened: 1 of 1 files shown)


htdocs/societe/class/societe.class.php
```diff
@@ -4082,12 +4082,16 @@ public function isACompany()
...
 		} elseif (!empty($this->idprof1) || !empty($this->idprof2) || !empty($this->idprof3) || !empty($this->idprof4) || !empty($this->idprof5) || !empty($this->idprof6)) {
 			$isacompany = 1;
-		} elseif (!empty($this->typent_code) && $this->typent_code != 'TE_UNKNOWN') {
-			// TODO Add a field is_a_company into dictionary
-			if (preg_match('/^TE_PRIVATE/', $this->typent_code)) {
-				$isacompany = 0;
+		} else {
+			if (getDolGlobalString('DEFINE_CUSTOMERS_ARE_COMPANIES_BY_TYPE_COMPANY')) {
+				// TODO Add a field is_a_company into dictionary
+				if (preg_match('/^TE_PRIVATE/', $this->typent_code)) {
```
eldy, line 4088: Whats if typent_code is not defined (null) it may generzte an error ?
We may add before
If !empty() && ...

thomas-Ngr: Passing null as a $subject for preg_match() does not raise an error. It returns 0. I tested it.

```
echo '<?= var_dump(preg_match("/truc/", null)) ?>' > /tmp/test.php ; php /tmp/test.php
```

And $this->typent_code is always set (even if null) since it is a property of class Societe, so php 8.X won't cry either.

Issue #26032, opened 2023-09-26
Bug :  Not possible to filter by Products tags in Supplier invoices listing


hasse-basse:
# Bug
When trying to filter on product tags/categories in the table listing of Suppliers invoices, it is not possible to do, it does not show up any result at all, while it works fine in Client invoices and in Supplier order.

## Environment
- **Version**: 18.0.1
- **OS**: Linux
- **Web server**: Apache
- **PHP**: fpm-fcgi 8.1.23
- **Database**: MySQL or MariaDB 10.5.19-MariaDB-0+deb11u2
- **URL(s)**: /azert/htdocs/fourn/facture/list.php

hasse-basse:
This issue is still present in 19.0.1

hregis:
@hasse-basse try with this fix https://github.com/Dolibarr/dolibarr/pull/29074

hasse-basse:
Wonderful, it works !
Thanks !

PR #29074, opened 2024-03-26
FIX wrong sql request with product category filter

hregis:
@eldy for fix https://github.com/Dolibarr/dolibarr/issues/26032


hasse-basse:
Works fine for me in 19.0.1

PR #29077, opened 2024-03-26
FIX use $conf->browser object + clean code

hregis:


hregis:
@eldy phan (club) errors are fixed

PR #29082, opened 2024-03-27
Fix row order

bradley-jarvis:
Fixes being able to reorder lines in commande/card.php and I am guessing all other items that have a list. The 'fk_element' was being validated as an int, whereas from what I can tell it is infact a string label of the sql table column name


bradley-jarvis:
Please disregard all the noise on this pull request, it is a simple one liner in core/ajax/row.php on the FIX row order change, I accidentality added it onto the  [fix-update-bom-productservice](https://github.com/Dolibarr/dolibarr/pull/29082/commits/82760db88e17d8708e1992d96d17e2d0108e4944)

PR #29126, opened 2024-03-29
FIX: $conf must reflect enabled/disabled state of modules

atm-florianm:
# FIX $conf does not always reflect enabled/disabled state of modules
The issue is quite minor and doesn't occur often, but in some scripts we have something like:
```php
unactivateModule($modulename, 0);
activateModule($modulename, 0);
```
To reload a module.

Today, the script will not work as intended because both `unactivateModule` and `activateModule` check in `$conf` if the module is already enabled or disabled, but `$conf` is not updated or reloaded by the functions, so what happens is one of two possibilities:

1. the module is initially enabled => `unactivateModule()` will work, but `activateModule()` won't because `$conf->global->MAIN_MODULE_THEMODULENAME` hasn't been unset.
2. the module is initially disabled => this part of the script will work as intended, but if the script is longer and uses `$conf->{'themodule'}->enabled` later, there will be problems.


eldy:
The travis CI error has been fixed. Can you update your PR with last dev branch so CTI can be executed completely, i can't do it myself...

PR #29153, opened 2024-03-31
FIX correction coquilles sur PCG14

Vaadasch:
Correction de quelques coquilles sur le PCG14.

Une ou deux inversions de chiffres dans un nombre.

Des relations aux parents qui n'étaient pas exactes.


hregis:
@Vaadasch and for the PCG14 in production ?

Vaadasch:
Those things are not critical, and if it is, an accountancy expert probably has already corrected the errors in production.

I'm not sure about changing values that are inside the database of those who have already installed Dolibarr.
Or do you mean about the master branch of the repository ?
If you speak about the "DEV" in the "PCG14-DEV", it is because it is the developped version of the PCG, with all its values. There is "base" and "abbreviated" (abrégé) too

hregis:
@Vaadasch pas de soucis ! ;-) la compta je perso m'en fout en fait ! ;-))

Vaadasch:
You're right to ask, and it is legitimate. It's good to document and clarify those things :-)

PR #29163, opened 2024-04-01
 FIX Dolibarr#29028 Md-theme - Login is on other icons

aspangaro:
Without patch

image

With patch

image

### DIFF
--- a/htdocs/main.inc.php
+++ b/htdocs/main.inc.php
@@ -2522,10 +2522,19 @@ function top_menu_user($hideloginname = 0, $urllogout = '')
 	}
 
 	if (!getDolGlobalString('MAIN_OPTIMIZEFORTEXTBROWSER')) {
+		if (getDolGlobalString('MAIN_THEME') != 'md') {
+			if (empty($user->photo)) {
+				$userLogin = '<span class="hidden-xs maxwidth200 atoploginusername hideonsmartphone paddingleft">' . dol_trunc($user->firstname ? $user->firstname : $user->login, 10) . '</span>';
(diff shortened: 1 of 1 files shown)


eldy:
I do not experience trouble with current v19
image

Did your clear CSS cache of your browser ?

aspangaro:
You don't see the problem because you have a photo or you use the gravatar module.

The problem is only when we have no photo (empty($user->photo))

eldy:
With no photo and no gravatar.
image

also i see no diff in code to explain what you got.

I will test with your change to see what happen on my side...

aspangaro:
In the linked issue, several user are the same problem. I'm going to try again with a new clean installation

sonikf:
I can confirm the bug exists also in a clean 20.0.0-alpha installation and occurs when username is visible.
Tested in Ubuntu 22.04
Screen resolution 1920x1080, 1024x768
Latest Google Chrome and Firefox
Cleaning browser cache doesn't fix this
@aspangaro fix works

eldy:
> and occurs when username is visible.

What do you mean with "and occurs when username is visible". Do you use a hidden option to have the user name visible ?

eldy:
Can you try commit 330d20813ebd7c462f48a796293cd0efdde33740 instead.
I think it is a better fix because trouble is in CSS and not in HTML: HTML must be the same for eldy and md theme and difference must be into CSS file.
Also the trouble existed also in eldy theme when user has no photo and username is long.

sonikf:
It works!

PR #29167, opened 2024-04-01
allow updating a single Complementary Attribute 

JonBendtsen:
NEW: Allow API change a proposals single Complementary Attribute 
Previously to update a single Complementary Attribute for a proposal through the API one would have to include the full list of all complementary attributes in the `array_options` segment of the json, but with this change it is possible to only include the single attribute one wants to change.


JonBendtsen:
This fix is Similar to issue # 29115 but just for proposals - I hope to let valtbern submit the one for products.

JonBendtsen:
I'll submit a few more fixes similar to this one, but will only reference this merge request.

PR #29168, opened 2024-04-01
allow updating orders single Complementary Attribute

JonBendtsen:
NEW: Allow API change a orders single Complementary Attribute
Previously to update a single Complementary Attribute for an order through the API one would have to include the full list of all complementary attributes in the array_options segment of the json, but with this change it is possible to only include the single attribute one wants to change.


JonBendtsen:
similar to pull request #29167

JonBendtsen:
No need to specify all options_* just to keep them unchanged when updating 1 or more other Complementary attributes

```
{    "array_options": {
      "options_one": "09:46",
      "options_two": "09:46"
    }}
```

PR #29179, opened 2024-04-01
Fix php error in cron.class.php

ptibogxiv:


htdocs/core/modules/expedition/doc/pdf_espadon.modules.php
```diff
@@ -996,12 +996,6 @@ protected function _pagehead(&$pdf, $object, $showaddress, $outputlangs)
 		}
 
 		$pdf->SetDrawColor(128, 128, 128);
-		if (isModEnabled('barcode')) {
-			// TODO Build code bar with function writeBarCode of barcode module for sending ref $object->ref
```
eldy, line 1000: TODO are here to alert developers there is a task to do. We should not remove it. What's wrong with this part of code ? Which warning do you get ?

ptibogxiv: i dont see it, my PR add some modified files.... i will send a new one

PR #29185, opened 2024-04-01
remove ref_ext because my llx_projet table has no ref_ext column

JonBendtsen:
During development of getByRef expansion of API for project I also initially created a getByExtRef, but that newer gave an error with no project found. Turns out that there is no column ext_ref in my llx_projet table, only id, ref and email_msgid.

Those changes will be submitted in a separate pull request.


eldy:
I added the field ref_ext into table llx_projet so table is aligned with others main tables

PR #29186, opened 2024-04-01
NEW: API GET projects by REF and by email_msgid

JonBendtsen:
NEW: API GET projects by REF and by email_msgid
The API for Projects only had the possibility of using GET by id, but I needed REF for my own usage. I also created get by email_msgid because that seems to exist in the llx_projet table and in the fetch function for projects.

The fetch function for projects also listed an ext_ref which I initially also created an API GET for - but using it never failed when projects did not exist. Possibly because llx_projet table does not contain any ext_ref column. Removal of ext_ref will be supplied in a different pull request.

### DIFF
--- a/htdocs/projet/class/api_projects.class.php
+++ b/htdocs/projet/class/api_projects.class.php
@@ -77,7 +77,7 @@ class Projects extends DolibarrApi
 
 		$result = $this->project->fetch($id);
 		if (!$result) {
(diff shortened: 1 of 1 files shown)


JonBendtsen:
IF #29185 is approved then this line has to be adjusted to only 3 arguments, not 4 arguments.

`		$result = $this->project->fetch('','','',$email_msgid);`

PR #29203, opened 2024-04-02
move BILLREC_CREATEBILL to facture because of ui actions

rycks:
user can create an invoice from recurring invoices then the trigger must not be only into cron code ...

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -1092,8 +1092,17 @@ class Facture extends CommonInvoice
 						if ($result < 0) {
 							$error++;
 						}
+
+						//special case for recurring invoices, moved from facture-rec.class.php
+						if (!$error && $this->fac_rec > 0) {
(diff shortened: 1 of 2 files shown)


htdocs/compta/facture/class/facture.class.php
```diff
@@ -1092,9 +1092,18 @@ public function create(User $user, $notrigger = 0, $forceduedate = 0)
 						if ($result < 0) {
 							$error++;
 						}
+
+						//special case for recurring invoices, moved from facture-rec.class.php
+						if (!$error && $this->fac_rec > 0) {
+							$result = $this->call_trigger('BILLREC_CREATEBILL', $user);
```
eldy, line 1098: A trigger should reflect a CRUD action. So it is good to remove it from createRecurringInvoices, because this method is an orchestration method. No reason to use a trigger inside this method.
Having it at this new place is better. However, we see that the trigger is now a duplicate of BILL_CREATE that can already know that execution is done to create an invoice from a recurring invoice because ->fac_rec will be set when creation is done from a recurring invoice.

PR #29210, opened 2024-04-03
bad module name on activity trim

defrance:


eldy:
We are starting to standardize name of module. For that a compatibility later has been included into isModEnabled. Using the key invoice should have the same effect than facture. Do you experience a bug ?

PR #29211, opened 2024-04-03
FIX round capital societe

atm-jonathan:
# Instructions
$societe->capital 
Does not respect "Limites et précisions"

# FIX|Fix #[*rounding capital*]
add price() for display capital
before 2000000.00000000 (double24.8)
after fix 2 000 000,00 with price()


htdocs/core/class/commondocgenerator.class.php
```diff
@@ -323,7 +323,7 @@ public function get_substitutionarray_thirdparty($object, $outputlangs, $array_k
 			'company_supplieraccountancycode'=>$object->code_compta_fournisseur,
 			'company_juridicalstatus'=>$object->forme_juridique,
 			'company_outstanding_limit'=>$object->outstanding_limit,
-			'company_capital'=>$object->capital,
+			'company_capital'=> price($object->capital, 0, '', 1, -1),
```
eldy, line 326: The substitution array must contains brut data, so a number must have a php number format in memory. And price format a data for output so data is no more a number but a string with special chars. A data can't be reused once formated with price.
May be you want to use price2num function ?

eldy (commented):
The substitution array must contains brut data, so a number must have a php number format in memory. And price format a data for output so data is no more a number but a string with special chars. A data can't be reused once formated with price.
May be you want to use price2num function ?

htdocs/core/class/commondocgenerator.class.php
atm-quentin: ok so can we have a new array key like company_capital_formated ? we have things like that for total fields of customer documents

eldy: Yes, i think it is a better idea to have company_capital_formated if you want to use price.
company_capital = brut data
company_capital_formated = data formated accroding to language, country, currency and rounding setup.

atm-jonathan: @eldy I've made the change

PR #29214, opened 2024-04-03
NEW feature to merge 2 tasks

Hystepik:
Add a function to merge 2 tasks based on thirdparty merge


htdocs/projet/class/task.class.php
```diff
@@ -1329,6 +1329,79 @@ public function getListContactId($source = 'internal')
...
+
+			$this->timespent_date = $timespent->timespent_line_date;
+			$this->timespent_datehour = $timespent->timespent_line_datehour;
+			$this->timespent_withhour = $timespent->timespent_line_withhour;
+			$this->timespent_duration = $timespent->timespent_line_duration;
+			$this->timespent_fk_user = $timespent->timespent_line_fk_user;
+			$this->timespent_fk_product = $timespent->fk_project;
+			$this->timespent_note = $timespent->timespent_line_note;
+
+			$result = $this->addTimeSpent($user);
```
eldy, line 1393: Adding and deleting has a lot of bad effect. Risk to forgot a field if a new field is added in future, mode code. More insert, update.
Why not just updating the task id from old value to new one with a simple update ?

htdocs/core/class/html.form.class.php
JonBendtsen, line 1: purpose in changing htdocs/core/class/html.form.class.php?

htdocs/projet/class/task.class.php
JonBendtsen: @eldy  would the same apply to merging contacts and members?

eldy: > @eldy would the same apply to merging contacts and members?

Currently, we can merge only thirdparties. But being able to merge a contact with another contact or a memeber with another one, would be very interesting... Code must follow same method than merging thirdparties.

PR #29219, opened 2024-04-03
FIX phan error function isModEnabled can have value "proposal"

hregis:


hregis:
@mdeweerd 'proposal' isn't better for an english man instead of 'propal' or 'propale' ?

hregis:
english men or women, by the way! parity ! ;-)

mdeweerd:
@hregis
- I did not decide on what made it into a renamed module or not.
- My mother tongue is neither french nor english.

If it were up to me, the better wording is "quote" or "quotation" or "offer" with regards to how this is used in Dolibarr.  A "proposal" does not seem to be something legaly binding.
In an English business setting we often use RFQ not RFP nor RFF.

https://www.sumup.com/en-gb/invoices/dictionary/quotation/

EDIT: Also found the following - a proposal can be vague, a quote indicates how much you will charge to fullful the order.
https://www.lawbite.co.uk/resources/blog/what-is-the-difference-between-rfi-rfq-rft-rfp

hregis:
@mdeweerd ok ! 'proposal' or 'propal' or 'propale' is just for me the key of an element, an object ! :-) no problem for me... maybe in the futur to replace "element" with an "Object::PROPOSAL_ELEMENT_NAME" in order to free ourselves from names ! ;-)

`isModEnabled(Object::PROPOSAL_ELEMENT_NAME)`

mdeweerd:
> ok ! 'proposal' or 'propal' or 'propale' is just for me the key of an element, an object ! :-)
> Object::PROPOSAL_ELEMENT_NAME" in order to free ourselves from names ! ;-)

I believe that the use of constants is so much better because there is less room for error (even if phan & CodingPhpTest check for the module names in this case).

One could have both PROPOSAL_ELEMENT_NAME and QUOTATION_ELEMENT_NAME with the first being deprecated, and using name mapping as already done to allow non-core modules to continue to work with multiple dolibarr versions, including of course the new ones.

hregis:
> > ok ! 'proposal' or 'propal' or 'propale' is just for me the key of an element, an object ! :-)
> > Object::PROPOSAL_ELEMENT_NAME" in order to free ourselves from names ! ;-)
> 
> I believe that the use of constants is so much better because there is less room for error (even if phan & CodingPhpTest check for the module names in this case).
> 
> One could have both PROPOSAL_ELEMENT_NAME and QUOTATION_ELEMENT_NAME with the first being deprecated, and using name mapping as already done to allow non-core modules to continue to work with multiple dolibarr versions, including of course the new ones.

no need to have PROPOSAL_ELEMENT_NAME... QUOTATION_ELEMENT_NAME is just the only one... this is a key... with a key i can adjust a value in the key and if the code use this key, the value is ok with the key ! ;-)

hregis:
@mdeweerd @eldy "isModEnabled(Object::PROPOSAL_ELEMENT_NAME)" is a good solution to avoid a problem with element/object name

hregis:
if all call of element name use "Object::PROPOSAL_ELEMENT_NAME" i can rename "Object::PROPOSAL_ELEMENT_NAME" with no problem !

eldy (approved):

PR #29231, opened 2024-04-04
FIX avoid to delete "lock" and "unlock" files

hregis:


htdocs/core/class/utils.class.php
```diff
@@ -96,14 +96,14 @@ public function purgeFiles($choices = 'tempfilesold+logfiles', $nbsecondsold = 8
 			if ($choice == 'allfiles') {
 				// Delete all files (except install.lock, do not follow symbolic links)
 				if ($dolibarr_main_data_root) {
-					$filesarray = dol_dir_list($dolibarr_main_data_root, "all", 0, '', 'install\.lock$', 'name', SORT_ASC, 0, 0, '', 1);
+					$filesarray = dol_dir_list($dolibarr_main_data_root, "all", 0, '', '[\.lock|\.unlock]$', 'name', SORT_ASC, 0, 0, '', 1);
```
eldy, line 99: I think it should be 
'(\.lock|\.unlock)$'
instead of
'[\.lock|\.unlock]$'

the [ is used for a list of char, here we want the sequence of char .lock but .col should be not excluded.

hregis: @eldy same result but fixed thanks

hregis: > the [ is used for a list of char, here we want the sequence of char .lock but .col should be not excluded.

yes thanks :-)

PR #29237, opened 2024-04-04
Fix #29115 API PUT call issue (product extra fields)

### LINKED ISSUE #29115: API PUT call issue (product extra fields)
### Bug

Unexpected behavior of PUT method upon updating product's extra fields

### Dolibarr Version

19.0.1

### Environment PHP

8.1.27

### Environment Database

MariaDB 10.6.17-MariaDB-cll-lve

Trying to call a product update through PUT method to update an extra field of the product:

1. Extrafield (checkbox) underlined: API must uncheck the box

image

2. API request body:

image

3. After the call checkbox is unchecked which is right,
however, previously set values for extra fields are now gone.
NOTE: some of the extra fields didn't lose their values and it seems that those fields having _default value_ defined in extrafield settings don't lose the values, in fact, they **change** to default value from settings, regardless of what was set before or defined in API schema.

image

Obviously, standard fields are not affected, they're not changing (unless defined in schema) or loosing the values.

### Attached files

_No response_

JonBendtsen:
FIX #29115 API PUT call issue (product extra fields)
Allow API Put call to update just a single extra field / complementary attribute / array_options / options_ field without having to supply all the other extra fields which previously would be reset to null.

This commit should close #29115

### DIFF
--- a/htdocs/product/class/api_products.class.php
+++ b/htdocs/product/class/api_products.class.php
@@ -368,7 +368,12 @@ class Products extends DolibarrApi
 				$this->product->context['caller'] = sanitizeVal($request_data['caller'], 'aZ09');
 				continue;
 			}
-
+			if ($field == 'array_options' && is_array($value)) {
+				foreach ($value as $index => $val) {
+					$this->product->array_options[$index] = $val;
+				}
+				continue;
+			}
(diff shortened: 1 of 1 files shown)


JonBendtsen:
API_GET_PRODUCT_extra_fields_no_overwrite

PR #29240, opened 2024-04-05
FIX #28541 trunc barcode to avoid trespass on fichhalfright

### LINKED ISSUE #28541: Gestion de la longueur des données dans le champs QRCode
### Bug

image

### Environment Version

_No response_

### Environment OS

_No response_

### Environment Web server

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

### Environment URL(s)

_No response_

_No response_

_No response_

### Attached files

_No response_

evarisk-micka:
# FIX #28541
Truncate after 13 character to avoid value trespassing on fichhalfright

### DIFF
--- a/htdocs/product/card.php
+++ b/htdocs/product/card.php
@@ -2510,7 +2510,7 @@ if (is_object($objcanvas) && $objcanvas->displayCanvasExists($canvasdisplayactio
 						print '&nbsp;<input type="submit" class="button smallpaddingimp" value="'.$langs->trans("Modify").'">';
 						print '</form>';
 					} else {
-						print showValueWithClipboardCPButton($object->barcode);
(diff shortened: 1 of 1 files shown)


frederic34:
@evarisk-micka why not `print '</td><td class="tdoverflow">';` at line 2498

edit : no more clipbutton...

eldy:
I confirm feedback of frederic34
Truncation must no more be managed on PHP side but on client side by the CSS style.
you can use
td class="tdoverflow" (truncation will be done by browser)
You can enhance it with
td class="tdoverflow" title="valueverylong"  (so full value will be visible when mouse is over).

or

td class="wordbreak"  (line will be breaked in 2 to fit the cell).

For list, you should one first solution.
For the card page, you can use the second one.

PR #29245, opened 2024-04-06
fix phpdoc for ismultientitymanaged

frederic34:


hregis:
@frederic34 thanks, but "$ismultientitymanaged" si only an integer ! entiy is an integer ! nerver a string...

hregis:
if "entity" is a string... this is a bug !

frederic34:
image

PR #29251, opened 2024-04-06
Add multicompany support to bookcal (+title typo)

G-ladal:
Fix typo for "new availability" (trans index)
Add multicompany support to bookcal

@hregis : Requires adding following to last released MC actions_multicompany.class.php (in $this->sharingnamecompatibility) ... i don't know how to do PR on MC (or access MC repo) :
```
 'calendar'                  => 'agenda',
'bookcal_calendar'          => 'agenda',
 ```

Note/help appreciated :
- there's unsolved "required or not" comments, let me know best practices
- i couldn't find the "auto display entity in list" and "display entity name instead of number" tricks


hregis:
@G-ladal the better... give ma the goal of you want and i added this in the multicompany module ! ;-)

G-ladal:
with this PR, there is possibility to have bookcal ("réservation en ligne") for each entity (and share it via agenda perms).

This module create a new object type "calendar" which works as an agenda (that's why i bind it to agenda perms). If you want to create a specific sharing line for it it's ok, but I did it the easiest way, as it's +/- the same kind of things (and because you need agenda perms to be able to see bookcal menu ...).

G-ladal:
@hregis here is some screens of it working already :
conf : ent2 share agenda/bookcal with ent5 (ent5 doesn't share)
from ent 5 we can see all, and add availability to any.
Screenshot 2024-04-06 173811
Screenshot 2024-04-06 173735
Screenshot 2024-04-06 173437

eldy:
> ...  i don't know how to do PR on MC (or access MC repo) :

This is managed by telling what is parent table that contains the entity. I fixed this way in the child table:

`public $ismultientitymanaged = 'fk_bookcal_calendar@bookcal_calendar';`

G-ladal:
thanks, I didn't know about this trick yet.
(I'm digging into Dolibarr sourcecode for just a few days/weeks)

PR #29254, opened 2024-04-06
use MAIN_DB_PREFIX which seems to be the current standard

JonBendtsen:
Qual replace $this->db->prefix() with MAIN_DB_PREFIX which seems to be the current standard
While reading the code in this file I noticed that it did not seem to follow the same standard as other files when referencing the database tables.

### DIFF
--- a/htdocs/core/class/extrafields.class.php
+++ b/htdocs/core/class/extrafields.class.php
@@ -260,10 +260,10 @@ class ExtraFields
 				'default' => $default_value
 			);
 
-			$result = $this->db->DDLAddField($this->db->prefix().$table, $attrname, $field_desc);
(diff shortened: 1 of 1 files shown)


frederic34:
@JonBendtsen I think MAIN_DB_PREFIX is the old way, `$this->db->prefix()` has been introduced after

JonBendtsen:
> @JonBendtsen I think MAIN_DB_PREFIX is the old way, `$this->db->prefix()` has been introduced after

oh

mdeweerd:
My suggestion is to either use DoliDB::prefix() or to deprecate it or move to a function such as 'get_db_table($table_name, $quote=true)' which would prefix and  quote the db table for the db request (i.e. `` `prefix_table_name` `` for mysql or `"pref_table_name"` for sqlite3).

PR #29255, opened 2024-04-06
Qual: Enable Phan Rules and update baseline

mdeweerd:
# Qual: Enable Phan Rules and update baseline

This enables all of the rules that are currently also
enabled for cti, but sets the baseline to ignore the
exceptions already present.
This will help detect exceptions added by code modifications
if the rule is not ignored for the given file.

I also added a stub for nusoap (gives better and the same results in ci and locally) and fixed some minor phan notices that were reported in ci but not locally.


eldy:
@mdeweerd Do you know this event: https://www.dolibarr.org/dolibarr-devcamp-summer-montpellier-2024.php
It would be great if we can see you...

PR #29271, opened 2024-04-07
Cleanaccount

frederic34:


htdocs/societe/class/companybankaccount.class.php
```diff
@@ -193,17 +187,6 @@ class CompanyBankAccount extends Account
 
 	public $country_code;
 
-
-	/**
-	 * @var string owner
-	 */
-	public $proprio;
```
eldy, line 200: To have transition version, we are keeping $this->proprio assignation, for example line 520.
Don't we must keep also the declaration here and add the flag
`@deprecated use $owner`

frederic34: $this->proprio is in account.class.php and is set to deprecated

frederic34: https://github.com/Dolibarr/dolibarr/blob/2e938fe9ef32a464aeb0c7f782ec1d261da7f0f5/htdocs/compta/bank/class/account.class.php#L397-L406

frederic34: if you try to access proprio (private) it will return owner_name

PR #29275, opened 2024-04-08
CLOSE|Close #29274 Add possibility to have different mail for autocopy for holiday

### LINKED ISSUE #29274: Add possibility to have several mail for autocopy
### Feature Request

Add possibility to have several mail for autocopy, 
Let me explain myself, it's currently possible to define a mail added in hidden copy each time a mail is send via dolibarr in Home > Setup > Mails > Others options. What I want to add is the possibility to change this email according to the situation.

### Use case

Here is an example to better understand:
Until now, this email was set up on commercial@my-compagny.com, but I recently added the holiday module manage leaves and my emails were polluted by leave requests and validation, so I need to be able to set up many emails.

This feature is necessary for large companies that have different services to manage the commercial side, human resource management... and where each service has its own email address

### Suggested implementation

Add a parameter ($autocopyTo) to __construct function of CMailFile class (/core/class/CMailFile.class.php).
 
This param is set to main by default (so if not specified, it just work as always)

Further in the function, just before the value of MAIN_MAIL_AUTOCOPY_TO is added to the list, we get the corresponding [$autocopyTo]_MAIL_AUTOCOPY_TO

Then we can add option of configuration in Setup > Mails > Others options (Otherwise it's possible to add your own option in Setup > Other Setup)

### Suggested steps

_No response_

lunoshist:
Add possibility to have several mail for autocopy,
Let me explain myself, it's currently possible to define a mail added in hidden copy each time a mail is send via dolibarr in Home > Setup > Mails > Others options. What I want to add is the possibility to change this email according to the situation.
Use case

Here is an example to better understand:
Until now, this email was set up on [commercial@my-compagny.com](mailto:commercial@my-compagny.com), but I recently added the holiday module manage leaves and my emails were polluted by leave requests and validation, so I need to be able to set up many emails.

This feature is necessary for large companies that have different services to manage the commercial side, human resource management... and where each service has its own email address


eldy:
I seems there is already a feature to allow MAIN_MAIL_AUTOCOPY according to module
MAIN_MAIL_AUTOCOPY_TO is for everywhere
MAIN_MAIL_AUTOCOPY_XXX_TO is for module XXX

For example 
MAIN_MAIL_AUTOCOPY_INVOICE_TO
MAIN_MAIL_AUTOCOPY_ORDER_TO
MAIN_MAIL_AUTOCOPY_SHIPMENT_TO
MAIN_MAIL_AUTOCOPY_PROJECT_TO
...

You should have a look at how it is done for existing case and see how to reproduce the same mecanism for holiday...
It seems the autocopy is propagated to CMailfile into $sendtobcc with this line of code by the caller

```
			// Autocomplete the $sendtobcc
			// $autocopy can be MAIN_MAIL_AUTOCOPY_PROPOSAL_TO, MAIN_MAIL_AUTOCOPY_ORDER_TO, MAIN_MAIL_AUTOCOPY_INVOICE_TO, MAIN_MAIL_AUTOCOPY_SUPPLIER_PROPOSAL_TO...
			if (!empty($autocopy)) {
				$sendtobcc .= (getDolGlobalString($autocopy) ? (($sendtobcc ? ", " : "") . getDolGlobalString($autocopy)) : '');
			}
```

So can you do the same for holiday ?

lunoshist:
You're right, what I had done was complex for nothing simply add getDolGlobalString('MAIN_MAIL_AUTOCOPY_HOLIDAY_TO') in the function call was enough

PR #29280, opened 2024-04-08
Deposit not taken into account when displaying the amount of the supplier invoice.

atm-jpb:
# FIX|Fix # Deposit not taken into account when displaying the amount of the supplier invoice.

When we create a deposit for a supplier invoice, it is not considered by the getSumDepositUsed function, which results in not updating the amount on the supplier invoice if we apply this account.

@eldy : Is there a particular rule that I might have missed or misunderstood for which this test is applied here? Or is this fix applicable for supplier invoices?
i'm aware of the presence in v19 of this update but i ask if we can set this part of code in v15.

### DIFF
--- a/htdocs/core/class/commoninvoice.class.php
+++ b/htdocs/core/class/commoninvoice.class.php
@@ -159,10 +159,6 @@ abstract class CommonInvoice extends CommonObject
 	 */
 	public function getSumDepositsUsed($multicurrency = 0)
 	{
-		//if ($this->element == 'facture_fourn' || $this->element == 'invoice_supplier') {
-			// TODO
-		//	return 0.0;
-		//}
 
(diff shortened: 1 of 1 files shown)


eldy:
> # FIX|Fix # Deposit not taken into account when displaying the amount of the supplier invoice.
> 
> When we create a deposit for a supplier invoice, it is not considered by the getSumDepositUsed function, which results in not updating the amount on the supplier invoice if we apply this account.
> 
> 
> @eldy : Is there a particular rule that I might have missed or misunderstood for which this test is applied here? Or is this fix applicable for supplier invoices?
> 
> 
> 
> 
> 

In old version, the deposits for supplier invoice was not implmented.
So part of code was commented. 
I think it should have been uncommented with recent versions, once the feature was stable enough.

PR #29283, opened 2024-04-08
FIX phan error

hregis:


mdeweerd:
@hregis It's having trouble to download the sql installation file.  It's not systématic.
However it seems that this is not the case in older windows version.  I prepare a PR.

PR #29286, opened 2024-04-08
Fix #29279 Not admin that can create group have disabled button for group creation

### LINKED ISSUE #29279: Not admin that can create group have disabled button for group creation
### Bug

As not admin, I have a greyed "New group" button :
image

But, the top right button stay available to click:
image

And going to this page doesn't return 403. It seems I should have the permission to do it:
image

So why is this button greyed ?

### Dolibarr Version

18.0.5

### Environment PHP

8.1.27

### Environment Database

MariaDB 10.11.6

- Login with not admin account
- Clic on "Users and groups"
- See disabled button

### Attached files

_No response_

JonBendtsen:
Fix #29279 Grey/Disabled button for group creation
Menu code did not check for the correct permissions, maybe some permissions have changed because the old code does have:

`$user->hasRight("user", "user", "create")`

which is very similar to the change in this pull request.


hregis:
@JonBendtsen yes this is `$user->hasRight('user', 'user', 'creer')`

JonBendtsen:
@hregis Do you think it would be possible for the build system to detect if non-existent permissions are used? like in this case where it was supposed to be `creer` not `create`

hregis:
@JonBendtsen sorry... This is $user->hasRight('user', 'user', 'write') no?

JonBendtsen:
> @JonBendtsen sorry... This is $user->hasRight('user', 'user', 'write') no?

@hregis 

the old code had this

`  $user->hasRight("user", "user", "create"  `

but it should be

`  $user->hasRight("user", "user", "creer"  `

So I was thinking if the build system could detect when some code tries to use a non-existent permission - just like it detects using non-existent variables, functions, …

PR #29287, opened 2024-04-08
FIX multiple problems when MULTICOMPANY_TRANSVERSE_MODE

hregis:


hregis:
@mdeweerd  ?? "CI / windows-ci" error ? ;-)

mdeweerd:
@hregis It's having trouble to download the sql installation file. It's not systématic.
However it seems that this is not the case in older windows version. I prepared a PR.

hregis:
@eldy it's ready for me, in transverse mode, a basic user with the permissions of create a user or group can not create users and groups and buttons/menus for to create are grey

hregis:
@eldy in the futur, with the field "enabled" in the llx_rights_def, we will have the possibility to deactivate the possibility of assigning a permission under certain conditions

hregis:
@eldy this is a fault positive phan error :-) (an open wound)

mdeweerd:
~Almost or~ already resolved in #29298 .

PR #29289, opened 2024-04-09
NEW possibility to define a completely inactive module

hregis:


hregis:
@mdeweerd @eldy bon je dois dire que "phan" me soule grave ! tout ça pour du commentaire mal commenté ! est-ce qu'il y a une doc ?

hregis:
je lâche l'affaire ! il faut nous donner des bonnes pratiques !

hregis:
tu mets une modif dans un fichier il te met des dizaines d'erreurs "phan" dans pleins de fichiers qui n'ont rien à voir !

hregis:
@mdeweerd what is the goal of "phan" ? if a variable is used incorrectly we will have an error, why try to create errors before there are any? ;-)

hregis:
je me vois bien mettre ça en commentaire ! ;-)

`array|array<string,array<string,associative-array<int,mixed>>>|array<string,array<string,associative-array<mixed,string>>>|array<string,array<string,float>>|array<string,array<string,int>>|array<string,array<string,mixed>>|array<string,array<string,string>>`

franchement ça rime à quoi ? alors si on a une variable de type "array" qui peut contenir des valeurs indéfinies nous seront obligés de mettre en commentaire ce type de déclaration pour éviter une "phan" erreur ?

hregis:
il n'y a pas une balise "nophan" ?  :-)

```
/**
 * __stop__the__phan__and__go__home__
 * @var array List of hooks called during this request
 * __i__am__the__fan__of_phan__
 */
 public $hooksHistory = [];
```

hregis:
```
/**
 * @var array|array<string,array<string,associative-array<int,mixed>>>|array<string,array<string,associative-array<mixed,string>>>|array<string,array<string,float>>|array<string,array<string,int>>|array<string,array<string,mixed>>|array<string,array<string,string>> List of hooks called during this request
 */
 public $hooksHistory = [];
```

mdeweerd:
> @mdeweerd @eldy bon je dois dire que "phan" me soule grave ! tout ça pour du commentaire mal commenté ! est-ce qu'il y a une doc ?

https://github.com/phan/phan/blob/v5/internal/Issue-Types-Caught-by-Phan.md


> @mdeweerd what is the goal of "phan" ? if a variable is used incorrectly we will have an error, why try to create errors before there are any? ;-)

"Phan attempts to prove incorrectness rather than correctnes."
Incorrectness implies things that lead to bugs.

Phan & phpstan (!) are static analysis tools that help avoid bugs:
- By bringing "strong" typing to php;
- By identifying statements/structures that are likely indicative of bugs.

Phan does not as such create errors before there are any however in the CI flow they are called errors.  By paying attention to the phan and phpstan messages and avoid them (either by coding differently or ignoring the annotation), bugs are detected before they are actually detected by a user's use case;

The phpdoc comments are no longer just indicative comments, they are binding comments at analysis time, not at run time.
If you declare an argument as an int but then initialise it to a string, it's normal that you get a notice about that.  There are plenty of languages where your code would just not compile.

Static tools can notify about things that aren't bugs and the overhead that generates is not "fun", however overall they help avoid bugs that require more effort to identify and fix.

IMHO the static checks are important for Dolibarr as the automated tests only cover a small part of the functionnality and newly added functionnality is not required to have any associated tests to be integrated - there are several opensource projects (not the majority) where a functionnality is not integrated if there are no associated tests.

hregis:
@mdeweerd @eldy I see the principle clearly, I got a little angry because we don't know if we should correct it in our PR, or in a separate PR, and how? Is there a dictionary of errors that would allow us to guide ourselves in resolving these "errors"? Thank you for your work and your explanations.

mdeweerd:
@hregis The closes to a documentation of the issue is https://github.com/phan/phan/blob/v5/internal/Issue-Types-Caught-by-Phan.md .
However that only shows examples that trigger the given exception, not what the typical fix is.

In this PR, the issues in the hookmanager were added this morning and are related to #27998 .
To allow the static tools to do their job better, the array types need to be more specific.  If its not added at the time of writing it's harder to do by some future developer, incorrect arguments would slip through, etc.

```
commit b172640ae507f1f97a0d88176d2cff4cd4d2433a
Date:   Tue Apr 9 03:38:38 2024 +0200

    Debug v20

commit 4880ec1b771f3d264f9b922d69bdbf0d2061fe98
Date:   Tue Apr 9 03:09:10 2024 +0200

    NEW: Hooks tab in debugbar (#24992)
```

hregis:
thank you @mdeweerd, at the end of the day I know it comes from another PR... but it's painful that it fell on me... should I not be angry and wait patiently for @eldy to validate my PR without have you corrected this error which is not my fault?

on the other hand I can't find this error in the list you just gave me? "PhanPluginUnknownArrayPropertyType"

hregis:
@mdeweerd phan phan phan... You want to drive me crazyyyyy ? ;-))

hregis:
@mdeweerd No, it’s already been done for a long time! ;-))

mdeweerd:
I agree that PhanPluginUnknownArrayPropertyType is not documented on that page and when you look it up using a major search engine, Dolibarr is the first result !

eldy:
> I agree that PhanPluginUnknownArrayPropertyType is not documented on that page and when you look it up using a major search engine, Dolibarr is the first result !

:-)
@mdeweerd 
May be we can disable it for cti and enable it only with apstats ? Is this possible ?

eldy (approved):

PR #29293, opened 2024-04-09
fix regl_mode on new command not set

vmaury:
FIX : When creating a new supplier order, even if the default reglement_mode is set for this supplier, the reglement_mode for the new order is not set. As It is for the cond_reglement
I don't understand why my update is ok, but IT IS !


vmaury:
Sorry I dont understand the problem with CI/phan 
I added only 3 letters in one line :-\

Issue #28296, opened 2024-02-19
Warning: Undefined array key "totalbalance" in htdocs/accountancy/bookkeeping/listbyaccount.php on line 1278


AWeerWolf:
### Bug

Warning: Undefined array key "totalbalance" in htdocs/accountancy/bookkeeping/listbyaccount.php on line 1278

![Screenshot 2024-02-20 001131](https://github.com/Dolibarr/dolibarr/assets/74211972/7621f68f-7ecb-45e4-bfd8-23d757995f94)

AWeerWolf:
Only with PHP 8.0
No warning with PHP 7.4

sonikf:
No errors in v 19.0.0 PHP 7.4, 8.0, 8.1, 8.2

lunoshist:
I have the same warning, 
I'm in v19.0.1 and PHP 8.1.25, and above all I'm developing my own module so I activated 'display_errors=On' in my php.ini and xdebug.

I personnaly removed the warning with this simple change : 
```
$totalarray['val']['totalbalance'] += $line->debit - $line->credit;
```
into
```
isset($totalarray['val']['totalbalance']) ? $totalarray['val']['totalbalance'] += $line->debit - $line->credit : $totalarray['val']['totalbalance'] = $line->debit - $line->credit;
```

PR #29295, opened 2024-04-09
FIX|Fix #29294 and #28296 Remove warning

### LINKED ISSUE #29294: Warning in accountancy journals
### Bug

There are multiple warning in sells, purchses, or bankjournal page of accountancy module.
I see them when option 'display_errors=On' is active in my php.ini file, it's basically undefined array key, it's still working correctly but I think it deserves to be fix.

### Dolibarr Version

19.0.1

### Environment PHP

8.1.25

### Environment Database

_No response_

_No response_

### Attached files

Screenshot from 2024-04-09 13-08-07

lunoshist:
Remove multiple warning due to undefined array key in sells, purchses, bankjournal, and listbyaccount page of accountancy module.


eldy:
@mdeweerd Do you know how to fix the phan error on this pr ?

mdeweerd:
Adding multiple `'@phan-debug-var $totalarray';` helped to locate the issue.

I added a cast to a few lines such as `+= (float) $line->credit` which seemed to fix a warning - maybe it's not needed - I did not remove it.

The main issue is located in the 'isset()' test with embedded assignments.  But even moving the isset() inside an 'if' condition did not help - as the 'totalcredit' and 'totaldebit' values are already initialised to 0 there is no need to test if they are set, so I removed the test.  Phan does not raise an issue.

IMHO phan does not always do a great job with these array indexes and it seems easier in several cases to just add a `@phan-suppress-next-line` comment.  That implies that the notification was examined and considered false "by a human".  So if the notification is real, then it is at least hightlighted and fixable.

I proposed the fix in https://github.com/mdeweerd/dolibarr/tree/pr29295 and made a PR to lunoshist:Warning .

PR #29304, opened 2024-04-10
FIX : Error when updating user

atm-adrien:
## FIX : Error when updating user

I created an extrafield that is unique
When i update an user to test the unique propertie of the new extrafield, the wrong error is displayed.

Moreover, it seems that this error is already manage in the update function (user.class.php)

### DIFF
--- a/htdocs/user/card.php
+++ b/htdocs/user/card.php
@@ -548,13 +548,8 @@ if (empty($reshook)) {
 					$ret = $object->update($user);		// This may include call to setPassword if password has changed
 					if ($ret < 0) {
 						$error++;
-						if ($db->errno() == 'DB_ERROR_RECORD_ALREADY_EXISTS') {
-							$langs->load("errors");
-							setEventMessages($langs->trans("ErrorLoginAlreadyExists", $object->login), null, 'errors');
-						} else {
-							setEventMessages($object->error, $object->errors, 'errors');
-							$action = 'edit';
-						}
+						setEventMessages($object->error, $object->errors, 'errors');
+						$action = 'edit';
 					}
 				}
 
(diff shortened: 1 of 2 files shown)


htdocs/user/card.php
```diff
@@ -548,13 +548,8 @@
...
 					if ($ret < 0) {
 						$error++;
-						if ($db->errno() == 'DB_ERROR_RECORD_ALREADY_EXISTS') {
-							$langs->load("errors");
-							setEventMessages($langs->trans("ErrorLoginAlreadyExists", $object->login), null, 'errors');
-						} else {
-							setEventMessages($object->error, $object->errors, 'errors');
-							$action = 'edit';
-						}
+						setEventMessages($object->error, $object->errors, 'errors');
```
eldy, line 551: object->error will contains a technical error message. When we have a technical error, we have no choice than showing this. But in the case of an error because of a duplicate value not allowed, we can (so we should) show a message the end user understand.
A solution would be to replace in olde code
setEventMessages($langs->trans("ErrorLoginAlreadyExists", $object->login), null, 'errors');
into
setEventMessages($langs->trans("ErrorUpdateCanceledDueToDuplicatedUniqueValue", $object->login), null, 'errors');   // duplicate value can be a login already exists or a unique extrafield exists.

with ErrorUpdateCanceledDueToDuplicatedUniqueValue=Error, The update was canceled because one of the pieces of data that should be unique already exists

htdocs/user/card.php
```diff
@@ -548,7 +548,7 @@
 					$ret = $object->update($user);		// This may include call to setPassword if password has changed
 					if ($ret < 0) {
 						$error++;
-						if ($db->errno() == 'DB_ERROR_RECORD_ALREADY_EXISTS') {
+						if ($db->errno() == 'ErrorUpdateCanceledDueToDuplicatedUniqueValue') {
```
eldy, line 551: This is surely not the error code. Should be the error message...

eldy (changes requested):
This is surely not the error code. Should be the error message...

htdocs/user/card.php
atm-adrien: sorry, too fast on the correction :)

PR #29307, opened 2024-04-10
17.0 Backport autocomplete in dialog box

MaximilienR-easya:
# Fix
Fix of the autocomplete list inside a dialog box

Before the fix:
Screenshot from 2024-04-10 14-00-39

the list is unusable in this state

After the fix:
after_fix

I also added the possibility to resize the dialog box in case there is not enough space for the list


eldy:
I have got no trouble on develop
image

Can you check to fix the same way (by backport) ?

MaximilienR-easya:
@eldy the backport is done

PR #29322, opened 2024-04-11
FIX multicurrency syncRates disable condition

AlbanDurrheimer:
# FIX multicurrency syncRates disable condition  
The condition to disable currency synchronization seems to have been reversed.  
Currently, it is necessary to set `MULTICURRENCY_DISABLE_SYNC_CURRENCYLAYER` to `1` for synchronization to work. However, this also has the effect of hiding the API configuration menu.


edupulpillo:
The call to curencylayer works succesfully, but the values are not added to the table llx_multicurrency_rate

PR #29327, opened 2024-04-11
FIX : Scale for the "second" unit

atm-adrien:
## FIX : Scale for the "second" unit

I can't do operations/conversion with seconds because the system constantly returns 0 to me
I think the error comes from the fact that the seconds scale should be 1 and not 0

image

image

Maybe there is a better solution than this?

### DIFF
--- a/htdocs/install/mysql/migration/19.0.0-20.0.0.sql
+++ b/htdocs/install/mysql/migration/19.0.0-20.0.0.sql
@@ -1,0 +1,1 @@
+UPDATE llx_c_units SET scale = 1 WHERE code = "S";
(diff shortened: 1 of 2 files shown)


htdocs/install/mysql/migration/19.0.0-20.0.0.sql
```diff
@@ -0,0 +1 @@
+UPDATE llx_c_units SET scale = 1 WHERE code = "S";
```
eldy, line 1: SQL stringmust use simple quotes so   = 'S';
This will solve the CTI error about SQL syntax...

atm-adrien: Done

PR #29341, opened 2024-04-13
Fix API GET extrafield returns 404 when not found. In general better (error) descriptions.

JonBendtsen:
Fix API GET extrafield returns 404 when not found. In general better (error) descriptions.
GET extrafield now returns 404 when not found, before it was just []
Using elementtype in the getallextrafields so all API methods uses the same name for elementtype.
Better error description when POST and PUT can not find obligatory fields in request data at the root level.

**WARNING!**
How does multi entity work when it comes to extrafields? Because in my previous pull requests for now API endpoints I let people specify the entity, which I fear may have allowed different entities to POST and PUT extrafields for other entities.

**FIX**
This pull request should fix that by asking Dolibarr for the entity, not trusting the user input.


JonBendtsen:
@eldy yes I know there is a build error, but it does not at all seem related to my code

htdocs/api/class/api_setup.class.php
```diff
@@ -1341,17 +1343,15 @@ public function postExtrafields($attrname, $elementtype, $request_data = null)
 			$extrafields->$field = $this->_checkValForAPI($field, $value, $extrafields);
 		}
 
+		$entity = getEntity('extrafields');
```
eldy, line 1346: getEntity can be used to return the list of allowed entities. It can be '1,4,9'
Here you need the entity where the post is addressed. You are right, providing it as parameter is not enough.
I think you can try use $conf->entity
to get the entity the code is running into (solution 1) but a http header can be used in the POST to work with another value. Value is checked before setting it into conf->entity.
Another solution (solution 2) may be to allow parameter entity and check that value is among the list of values returned by getEntity.
Sol 1 looks better. Thanks to try it.

eldy (changes requested):

htdocs/api/class/api_setup.class.php
JonBendtsen: `$entity = DolibarrApiAccess::$user->entity;`

found in api_documents.class.php seems to work.

JonBendtsen: except that database contents are different than those edited inside Dolibarr

inside Dolibarr is created with 1 as entity, where as using the line above uses 0 for entity :-( and $conf->entity did not work

JonBendtsen: https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/api/class/api_documents.class.php#L677-L680

works

JonBendtsen:
@eldy  does this pull request still have a conflict or something that I need to fix?

PR #29342, opened 2024-04-13
FIX - Update product.class.php

vanyolai:
Field and value order mismatch


htdocs/product/class/product.class.php
```diff
@@ -825,10 +825,10 @@ public function create($user, $notrigger = 0)
 					$sql .= ", label";
 					$sql .= ", fk_user_author";
 					$sql .= ", fk_product_type";
+					$sql .= ", price_label";
```
eldy, line 828: Old code seems good.
In values we have
```
$sql .= ", ".((int) $user->id);
					$sql .= ", ".((int) $this->type);
					$sql .= ", ".price2num($price_ht, 'MT');
					$sql .= ", ".price2num($price_ttc, 'MT');
```
so after fk_product_type, we should have price.

vanyolai: Now it's  working, looks like it's been solved right before I posted the pull request.

PR #29344, opened 2024-04-14
Qual: Fix or ignore functions.lib.php phan notices

mdeweerd:
# Qual: Fix or ignore functions.lib.php phan notices

No more 'general' exceptions are needed in phan's baseline for function.lib.php .


htdocs/main.inc.php
```diff
@@ -443,7 +443,7 @@ function analyseVarsForSqlAndScriptsInjection(&$var, $type, $stopcode = 1)
 	global $debugbar;
 	include_once DOL_DOCUMENT_ROOT.'/debugbar/class/DebugBar.php';
 	$debugbar = new DolibarrDebugBar();
-	$renderer = $debugbar->getRenderer();
+	$renderer = $debugbar->getJavascriptRenderer();
```
eldy, line 446: Do we really need to replace getRenderer with getJavascriptRenderer() ?
What about the call of disableVendor() that is inside getRender() that is lost ?

mdeweerd: This change is because of
https://github.com/Dolibarr/dolibarr/blob/64ecea7c83db5e4a433b07fc94295e9f3e240028/htdocs/core/lib/functions.lib.php#L10800-L10804

That code supposes that the DebugBar has a getRenderer) method which it does not have.

The DolibarrDebugBar was updated such that a call to the getJavascriptRenderer without arguments does exactly the same as the original getRenderer() method which is maintained as a deprecated method:

https://github.com/Dolibarr/dolibarr/blob/4547c8306077e9ca7e3e811d834200240f7b48d5/htdocs/debugbar/class/DebugBar.php#L77-L89

mdeweerd:
@eldy I see that you moved back to allow 'small' on a function.

It's better to write:
```php
* @param int<0,1>|''|'small' $size 0 or ''=original size, 1 or 'small'=use thumbnail if possible
```
so that the static checks only accept the documented argument values and report about misspellings.

mdeweerd:
@eldy 

FYI, the change of int to int|string needs to propagate to other methods as well (partial list, issued from a difference between reports):

```
@@ -5910,3 +5892,3 @@ htdocs\core\lib\functions.lib.php:2129 PhanUndeclaredMethod Call to undeclared m
htdocs\core\lib\functions.lib.php:2578 PhanTypeMismatchArgumentProbablyReal Argument 3 ($size) is 'small' of type 'small' but \Categorie::show_photos() takes int (no real type) defined at htdocs\core\class\commonobject.class.php:[-9175-]{+9183+} (the inferred real argument type has nothing in common with the parameter's phpdoc type)
htdocs\core\lib\functions.lib.php:2599 PhanTypeMismatchArgumentProbablyReal Argument 3 ($size) is 'small' of type 'small' but \BOM::show_photos() takes int (no real type) defined at htdocs\core\class\commonobject.class.php:[-9175-]{+9183+} (the inferred real argument type has nothing in common with the parameter's phpdoc type)
htdocs\core\lib\functions.lib.php:2621 PhanTypeMismatchArgumentProbablyReal Argument 3 ($size) is 'small' of type 'small' but \Ticket::show_photos() takes int (no real type) defined at htdocs\core\class\commonobject.class.php:[-9175-]{+9183+} (the inferred real argument type has nothing in common with the parameter's phpdoc type)
@@ -13430,2 +13412,2 @@ htdocs\takepos\ajax\ajax.php:129 PhanTypeMismatchArgumentProbablyReal Argument 1
htdocs\takepos\ajax\ajax.php:223 PhanTypeMismatchArgumentProbablyReal Argument 3 ($size) is 'small' of type 'small' but \Product::show_photos() takes int (no real type) defined at htdocs\core\class\commonobject.class.php:[-9175-]{+9183+} (the inferred real argument type has nothing in common with the parameter's phpdoc type)
htdocs\takepos\ajax\ajax.php:341 PhanTypeMismatchArgumentProbablyReal Argument 3 ($size) is 'small' of type 'small' but \Product::show_photos() takes int (no real type) defined at htdocs\core\class\commonobject.class.php:[-9175-]{+9183+} (the inferred real argument type has nothing in common with the parameter's phpdoc type)
@@ -13435 +13417 @@ htdocs\takepos\freezone.php:68 PhanTypeMismatchArgumentProbablyReal Argument 1 (
htdocs\takepos\genimg\index.php:86 PhanTypeMismatchArgumentProbablyReal Argument 3 ($size) is 'small' of type 'small' but \Product::show_photos() takes int (no real type) defined at htdocs\core\class\commonobject.class.php:[-9175-]{+9183+} (the inferred real argument type has nothing in common with the parameter's phpdoc type)
```

PR #29345, opened 2024-04-14
Update bonprelevement.class.php

developmentOYR:
Problem

When creating the file sepa ( Direct Debit File) if the amount ends in 0, such as 50.60, the program generates the file with rounding 50.6.

Some banks, for example, La Caixa (CaixaBank) do not accept the direct debit file if the decimal number does not have two digits.

Solution:

What I propose is to create a configuration variable called SEPA_ROUND_TWO_ZERO that allows for a two-digit rounding even if it ends in 0. That is, if the amount is 50.6, the file will be generated with the amount 50.60.

# Instructions
*Please:*
- *only keep the "FIX", "CLOSE", "NEW", "PERF" or "QUAL" section* (use uppercase to have the PR appears into the ChangeLog, lowercase will not appears)

[*Long description*]

[*Long description*]

[*Long description*]

[*Long description*]

[*Long description*]


htdocs/compta/prelevement/class/bonprelevement.class.php
```diff
@@ -1790,7 +1790,9 @@ public function generate($format = 'ALL', $executiondate = 0, $type = 'direct-de
 				if ($result != -2) {
 					$fileEmetteurSection .= $this->EnregEmetteurSEPA($conf, $date_actu, $nbtotalDrctDbtTxInf, $this->total, $CrLf, $format, $type, $fk_bank_account);
 				}
-
+				if($conf->global->SEPA_ROUND_TWO_ZERO){
+					$this->total=number_format($this->total,2,".","");
```
eldy, line 1794: We already have a functions to format price using the number if decimal if the country.
Can you try to use price2num with param 'MT' that can be used to force the rounding .

eldy (changes requested):

developmentOYR (commented):
As you have told me, I have used the price2num function with the 'MT' parameter but I have also had to use the number_format function since this function, when passing a value like 6.50, returns 6.5 and I need it to return 6.50 as i explained to you at the beginning of the request

htdocs/compta/prelevement/class/bonprelevement.class.php
developmentOYR: As you have told me, I have used the price2num function with the 'MT' parameter but I have also had to use the number_format function since this function, when passing a value like 6.50, returns 6.5 and I need it to return 6.50 as i explained to you at the beginning of the request

eldy:
> As you have told me, I have used the price2num function with the 'MT' parameter but I have also had to use the number_format function since this function, when passing a value like 6.50, returns 6.5 and I need it to return 6.50 as i explained to you at the beginning of the request

Ok, so lets stay on using  "number_format"
However, there is still a lot of errors in code syntax style. I recommend you to install phpcs and phpcbf tool with the git hook pre-commit to have them run and fix your code syntax automatically at each git commit.
See: https://wiki.dolibarr.org/index.php?title=Environment_and_development_tools#Install_Pre-commit_githook

developmentOYR:
After several attempts (I couldn't find the key) I managed to get it to pass the tests. There is one that does not pass but fails in a file that I have not modified (htdocs/core/modules/export/export_excel2007.modules.php).
I hope it's enough.

All the best

PR #29348, opened 2024-04-15
NEW Add option to change all service dates at once

SylvainLegrand:
Add an icon to the description title
Description-1

This icon open a form
Description-2

Before
Before

After
After


SylvainLegrand:
@eldy 

The run Phan analysis shows an error which comes from the original Dolibarr code.
Can we correct this?

eldy:
> @eldy 
> 
> The run Phan analysis shows an error which comes from the original Dolibarr code.
> Can we correct this?

Yes, you must pull the last version of branch and solve conflict to have such errors fixed

htdocs/core/tpl/objectline_title.tpl.php
```diff
@@ -56,7 +56,24 @@
 }
 
 // Description
-print '<th class="linecoldescription">'.$langs->trans('Description').'</th>';
+print '<th class="linecoldescription">'.$langs->trans('Description');
+// @phan-suppress-next-line PhanUndeclaredConstantOfClass
```
eldy, line 60: Can you remove this comment line ? May hide a bug.

SylvainLegrand: The huge problem is that if I remove this line, the CI/phan check fails.
The second point is that I copy this solution from line 91!

????? What should we do ????? ;)

eldy: Remove the line. If CI/phan check fails after, it means we found a bug to fix. So remove it so we can see what is the error message to have a clue of the bug...

htdocs/core/tpl/objectline_title.tpl.php
```diff
@@ -56,7 +56,23 @@
 }
 
 // Description
-print '<th class="linecoldescription">'.$langs->trans('Description').'</th>';
+print '<th class="linecoldescription">'.$langs->trans('Description');
+if (in_array($object->element, array('propal', 'commande', 'facture', 'order_supplier', 'invoice_supplier')) && $object->status == $object::STATUS_DRAFT) {
```
eldy, line 60: The problem is due to the line 
@phan-var-force CommonObject
few line before. It says to phan than $object is CommonObject but it is not, it is Facture or Propal or Commande...

Can you try this workaround:
`$constant = get_class($object)."::STATUS_DRAFT";
if (in_array($object->element, array('propal', 'commande', 'facture', 'order_supplier', 'invoice_supplier')) && defined($constant) && $object->status == constant($constant)) {
`

eldy:
How do you make your push request ? Did you make your commit on an IDE and the create the PR from github or do you edit code from github web ?

SylvainLegrand:
@eldy 

The last change you asked me to make I did by editing the file from the github web.

SylvainLegrand:
@eldy 

Now it's ready to merge

PR #29350, opened 2024-04-15
NEW update price with auto to use const for rounding prices

lvessiller-opendsi:
NEW update price with auto to use MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND const for rounding prices
- add a const MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND_SUPPLIER to have the same way to compute total amounts (round of total or total of round) but for supplier
- use the argument "auto" by default instead of "none" in "update_price()" method in order to have the same computed total amounts for each action : add line, update line, delete line, insert a discount, create from origin, etc

**Example**
This customer invoice with MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND = 0
image
- the VAT amount is 610,56

This same customer invoice with MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND = 1
image
- the VAT amount is 610,55

And when you add, update, delete a line or insert a discount you keep the same method to compute all total amounts

See #25234

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -3763,7 +3763,7 @@ abstract class CommonObject
 	 *  @param	Societe	$seller				If roundingadjust is '0' or '1' or maybe 'auto', it means we recalculate total for lines before calculating total for object and for this, we need seller object (used to analyze lines to check corrupted data).
 	 *	@return	int    			           	Return integer <0 if KO, >0 if OK
 	 */
-	public function update_price($exclspec = 0, $roundingadjust = 'none', $nodatabaseupdate = 0, $seller = null)
+	public function update_price($exclspec = 0, $roundingadjust = 'auto', $nodatabaseupdate = 0, $seller = null)
 	{
 		// phpcs:enable
 		global $conf, $hookmanager, $action;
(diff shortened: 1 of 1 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -3898,7 +3906,7 @@ public function update_price($exclspec = 0, $roundingadjust = 'none', $nodatabas
 						}
 						$obj->total_tva = $tmpcal[1];
 						$obj->total_ttc = $tmpcal[2];
-					} elseif ($diff_when_using_price_ht && $roundingadjust == '0') {
+					} elseif ($diff_when_using_price_ht) {
```
eldy, line 3909: Why removing the test ?
If we call the method with 'none', it means we don't want to make data cleaning. 
By removing the test, we will execute this even when using 'none'

lvessiller-opendsi: @eldy
"roundingajust" was not the good variable to use here and I replaced this condition with `$forcedroundingmode == '0'` but it was already checked at line 3889 : `if (empty($reshook) && $forcedroundingmode == '0') {`
Otherwise when you got "auto" in roudingadjust parameter, total amounts weren't updated even if a difference was detected.
So if we have `roundingajust = 'none'` as parameter, it will never goes to this part of code because forcedroundingmode was set to 'none" at line 3816 : `$forcedroundingmode = $roundingadjust"`

PR #29352, opened 2024-04-15
FIX: absolute discount set from client tab

altairis-noe:
FIX: absolute discount setting in the client thirdparty tab (htdocs/comm/card) used the "create" action, while the action in remx.php was create_remise, so it was not taken in consideration.

Additionnaly, I suggest using the MAIN_VAT_DEFAULT_IF_AUTODETECT_FAILS global setting to set the default vat value for the discount.

### DIFF
--- a/htdocs/comm/remx.php
+++ b/htdocs/comm/remx.php
@@ -378,7 +378,7 @@ if ($socid > 0) {
 			// VAT
 			print '<tr><td>'.$langs->trans("VAT").'</td>';
 			print '<td>';
-			print $form->load_tva('tva_tx', GETPOSTISSET('tva_tx') ? GETPOST('tva_tx', 'alpha') : 0, $mysoc, $object, 0, 0, '', 0, 1);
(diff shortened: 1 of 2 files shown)


htdocs/comm/remx.php
```diff
@@ -378,7 +378,7 @@
 			// VAT
 			print '<tr><td>'.$langs->trans("VAT").'</td>';
 			print '<td>';
-			print $form->load_tva('tva_tx', GETPOSTISSET('tva_tx') ? GETPOST('tva_tx', 'alpha') : 0, $mysoc, $object, 0, 0, '', 0, 1);
+			print $form->load_tva('tva_tx', (GETPOSTISSET('tva_tx') ? GETPOST('tva_tx', 'alpha') : (!empty(getDolGlobalString('MAIN_VAT_DEFAULT_IF_AUTODETECT_FAILS')) ? getDolGlobalString('MAIN_VAT_DEFAULT_IF_AUTODETECT_FAILS') : 0)), $mysoc, $object, 0, 0, '', 0, 1);
```
eldy, line 381: Do not use !empty() on getDolGlobalString(), it has becale useless.
You can replace:
(!empty(....  ...0))
With just
getDolGlobalString('MAIN_VAT_DEFAULT_IF_AUTODETECT_FAILS', '0')

eldy (changes requested):

altairis-noe:
Hello @eldy,
Thanks for the tip, it should be all good now

PR #29354, opened 2024-04-15
Qual: Fix Phan notices in CommonObject

mdeweerd:
# Qual: Fix Phan notices in CommonObject

Fix several phan notices in CommonObject and update baseline.


mdeweerd:
@eldy I will re-update this branch to fix the "new" issues but I do not have the time to finish this (it was ok at some point, but some new stuff slipped in);

PR #29362, opened 2024-04-16
fix #22744: aux account for salary

### LINKED ISSUE #22744: Bank Journal - Auxiliary account - "NotDefined" is returned instead of a clear value

rycks:
# FIX #22744


lvessiller-opendsi:
@rycks 
I can't determine if it's the good value for this subledger account.
@aspangaro, could you help us and check it  ?

You fixed for payment salary, nethertheless it seems to remain a similar fix for expense report (see comment on Oct 20 2023 of thibdrev).

htdocs/accountancy/journal/bankjournal.php
```diff
@@ -1315,6 +1315,9 @@ function writebookkeeping() {
 								} else {
 									$accounttoshowsubledger = '<span class="error small">'.$langs->trans("ThirdpartyAccountNotDefinedOrThirdPartyUnknownBlocking").'</span>';
 								}
+								if (in_array($tabtype[$key], array('payment_salary'))) {
```
eldy, line 1318: This fix is suspicious for me because the idea of the page is this:
- First we analyse database to fille php memory array $tabpay (lien for payment) and tabtp (line for the thirdparty or user).
- Then we show on screen the content of this arrays so user what will be done during transfer and when we do the transfer, we will use the same array to fill the accountancy.

On this fix, we fix the output shown to the user but but without changing the tabpay and tabtp. So it means when we will transfer in accountancy, we wont transfer the same that what we show ?

Because of this doubt, i decided to do the following thing: 
- I tried to reproduce the bug on v18 -> confirmed.
- i tried to reproduce the bug on v20 -> fixed.

So it appears this was already fixed. In such a case, we must take a look at the diff and backport the same fix.

Solution seems in this part of code (added into v20 and not present into v18)
```
if ($v['type'] == 'payment_expensereport') {
					$is_expensereport = true;
					break;
				}
```

and later on test (in v20)
`if ($links[$key]['type'] == 'user' && !$is_sc && !$is_salary && !$is_expensereport) {`
instead of (in v18)
`if ($links[$key]['type'] == 'user' && !$is_sc) {`

rycks: @eldy after re-reading initial bug details, then @lvessiller-opendsi  remarks ...  that bug concern two things

1. salary, what that patch try to fix (and as i remember the problem was only on "user display" but i have to re re re re read initial bug report to be really sure)
2. expenses reports 

yes you tests against v 20.0 for expense reports, an other test should be done for salary

eldy:
I removed the tag priority high because bug seems fixed in v20.
Waiting a backport of v20 fix into v18 to merge.

eldy:
Closed as PR should be added in v18 by a backport of fix in v20, however, it seems a lof ot change was necessary into v20 to fix this and backport seems not easy meaning that we can't fix this one into v18.
In such a case, it is better to say that this need v20+ if you don't find workaround on this bug, and close ticket as "Won't be fixed in v18".

PR #29367, opened 2024-04-17
FIX : Added missing hook to handle load_fiche_titre in showactions fu…

atm-irvine:
…nction

# FIX|Fix FIX : Added missing hook to handle load_fiche_titre in showactions function
[*Long description*]
This commit allows us to handle the function load_fiche_titre in showactions function


eldy:
As this is a new feature (adding a hook is an additional feature not yet implemented) can you push it into develop ?

PR #29370, opened 2024-04-17
FIX : Ergonomic fail

john-botella:
The road to hell is paved with good intentions :-) 

The show_actions_messaging in dialog version display behavior have change introducing bad user expérience reported by all employee from ATM after V19 Update.

Read message history is became more difficult.
And in some cases, we've even experienced a lack of understanding between users in the ticket messaging system. Because in, the users doesn't saw the 3 little dots.

Remember the function of message view (1) is to allow user to read an history fill like a chat or forum, if the user need a list he can use the list view (2)
image

# FIX

- [x] 1 - First step backward to origin display  
  Done by this pull request

- [x] 2 - Create a new PR to introduce a mor #29373e ergonomic read more behavior, (with configuration for user préférences).  
   see  #29373

image

image


eldy:
So a button Show more... will be usefull...

john-botella:
> So a button Show more... will be usefull...

yes that's why I created PR https://github.com/Dolibarr/dolibarr/pull/29373

PR #29373, opened 2024-04-17
FIX : Ergonomic fail  part02

john-botella:
follow #29370 

# FIX

Add hidden conf  `MAIN_TRUNCATE_TIMELINE_MESSAGE` to allow admin to set or disable truncate behavior

image


john-botella:
@eldy conflict managed

PR #29374, opened 2024-04-17
FIX - Search user in where propal and order

BB2A-Anthony:
# FIX|Fix #[*Search user in where*]
[*Long description*]


htdocs/commande/list.php
```diff
@@ -868,8 +868,10 @@
 $sql .= " LEFT JOIN ".MAIN_DB_PREFIX."projet as p ON p.rowid = c.fk_projet";
 $sql .= ' LEFT JOIN '.MAIN_DB_PREFIX.'user as u ON c.fk_user_author = u.rowid';
 if ($search_user > 0) {
-	$sql .= ", ".MAIN_DB_PREFIX."element_contact as ec";
-	$sql .= ", ".MAIN_DB_PREFIX."c_type_contact as tc";
+	$sql .= " INNER JOIN ".MAIN_DB_PREFIX."element_contact as ec";
```
eldy, line 871: We must abandon the use of join to make filter on data that is not part of the main table. We must instead have all the logic of the filter into the WHERE with IN (subselect)

eldy (changes requested):

BB2A-Anthony (commented):
> We must abandon the use of join to make filter on data that is not part of the main table. We must instead have all the logic of the filter into the WHERE with IN (subselect) 

@eldy it's done

PR #29377, opened 2024-04-17
FIX : No decimals in result.php

atm-adrien:
## FIX : No decimals in result.php

Continuation of this PR : https://github.com/Dolibarr/dolibarr/pull/28310

### DIFF
--- a/htdocs/compta/resultat/result.php
+++ b/htdocs/compta/resultat/result.php
@@ -354,7 +354,11 @@ if ($modecompta == 'CREANCES-DETTES') {
 					//$r = $AccCat->calculate($result);
 					$r = (float) dol_eval($result, 1, 1, '1');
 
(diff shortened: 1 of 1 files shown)


htdocs/compta/resultat/result.php
```diff
@@ -354,7 +354,11 @@
 					//$r = $AccCat->calculate($result);
 					$r = (float) dol_eval($result, 1, 1, '1');
 
-					print '<td class="liste_total right"><span class="amount">'.price($r).'</span></td>';
```
eldy, line 357: No need for if !empty when usin getDol...
Use if getDol...

eldy (changes requested):

htdocs/compta/resultat/result.php
atm-adrien: Done

Issue #29306, opened 2024-04-10
If VATIsNotUsed = true, Add value to CustomerAbsoluteDiscount gives error.


mikko80:
### Bug

If VATIsNotUsed = true, Adding value to CustomerAbsoluteDiscount not working. It gives error.
Fatal error: Uncaught TypeError: Unsupported operand types: string * string in /societe/class/societe.class.php:2435 Stack trace: #0 /comm/remx.php(182): Societe->set_remise_except('80', Object(User), 'Testi', '', '0', 'HT') #1 {main} thrown in /societe/class/societe.class.php on line 2435

### Dolibarr Version

19.0.1

### Environment PHP

8.2.16

### Environment Database

10.3.39-MariaDB, mysqli (mysqlnd 8.2.16)

### Steps to reproduce the behavior and expected behavior

In invoices or third party try to add absolute discount values.

mikko80:
ADD: societe.class.php:
// Check parameters
.....
line 2412->
if (empty($vatrate)) {
     $vatrate = '0 (0%)';
}

JonBendtsen:
> ADD: societe.class.php: // Check parameters ..... line 2412-> if (empty($vatrate)) { $vatrate = '0 (0%)'; }

that is one method. I was thinking of, this change

```
-       public function set_remise_except($remise, User $user, $desc, $vatrate = '', $discount_type = 0, $price_base_type = 'HT')
+       public function set_remise_except($remise, User $user, $desc, $vatrate = 0, $discount_type = 0, $price_base_type = 'HT')
```

but I don't know which is correct?

Which appears to be working?
![image](https://github.com/Dolibarr/dolibarr/assets/2551481/c7dda2e5-68c3-4db6-9ac6-bc13dcfda18c)

JonBendtsen:
seems to work irregardless of choosing with or without tax (and when sales tax set to `* not used` in the configuration

![image](https://github.com/Dolibarr/dolibarr/assets/2551481/438b8418-756e-4a24-becd-8c0edc9f5800)

![image](https://github.com/Dolibarr/dolibarr/assets/2551481/367d4c40-63bb-45ba-b752-9767eccfdcae)

JonBendtsen:
this works as well

```
-       public function set_remise_except($remise, User $user, $desc, $vatrate = '', $discount_type = 0, $price_base_type = 'HT')
[...]

PR #29328, opened 2024-04-12
NEW: Invoice Creditor Reference and banking barcode (FI)

mikko80:
NEW|New [Invoice Creditor Reference and banking barcode (FI)]

PR #29383, opened 2024-04-18
Mikko80 fix societe discount 1

mikko80:
# FIX|Fix #[#29306 VATIsNotUsed = true, Add value to CustomerAbsoluteDiscount gives error]
[IF vat empty set to 0]

# NEW|New [Invoice Payment reference]
[Calculate Creditor Reference RF / FI Bank payment reference number , Calculate payment Barcode data with FI/RF bank payment reference number ]


eldy:
Seems a duplicate of #29328 already merged. So i close the PR.

PR #29388, opened 2024-04-18
FIX not redirect when error occurs on updating card

lvessiller-opendsi:
FIX not redirect when error occurs on updating card
- if an error occurs when you update some information in a contact card, you are redirected in view mode of this contact card

This fix show an error message and keep the edit mode when an error occurs on updating a contact card.

### DIFF
--- a/htdocs/contact/card.php
+++ b/htdocs/contact/card.php
@@ -477,6 +477,7 @@ if (empty($reshook)) {
 
 					$action = 'view';
 				} else {
+					$error++;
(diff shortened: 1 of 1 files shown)


lvessiller-opendsi:
@eldy 
Do you think I can candidate for becoming a Jedi ? and how ?

eldy:
> @eldy 
> Do you think I can candidate for becoming a Jedi ? and how ?

No doubt, you can ! There is so few candidate, that there is no process, except asking, no matter the channel.
I suggest to make the switch during the devcamp.

PR #29393, opened 2024-04-18
Fix emailcollector ticket tracking

Hystepik:
Fix email collector to parmit the tracking of emails for ticket


htdocs/emailcollector/class/emailcollector.class.php
```diff
@@ -2063,7 +2063,7 @@ public function doCollectOneCollector($mode = 0)
 						} elseif (preg_match('/<(.*@.*)>/', $reference, $reg)) {
 							// This is an external reference, we check if we have it in our database
 							if (!is_object($objectemail)) {
-								$sql = "SELECT rowid FROM ".MAIN_DB_PREFIX."ticket where email_msgid = '".$this->db->escape($reg[1])."'";
+								$sql = "SELECT rowid FROM ".MAIN_DB_PREFIX."ticket where email_msgid = '".$this->db->escape($reg[1])."' OR origin_references like '%".$this->db->escape($reg[1])."%'";
```
eldy, line 2066: You must also include this->db->escapeforlike into the escape().
Escapeforlike is escaping the % and _ char 
Escape is escaping the '

eldy (changes requested):

htdocs/emailcollector/class/emailcollector.class.php
```diff
@@ -2136,6 +2136,25 @@ public function doCollectOneCollector($mode = 0)
 								} else {
 									$projectid = $objectemail->id;
 								}
+
+								if (get_class($objectemail) == 'Ticket') {
+									$changeonticket_references = false;
+									$trackid = $objectemail->track_id;
```
eldy, line 2142: Do we need this line ?

Hystepik: You we do need this line because if the track id isn't found in the mail, then it recreate a ticket after the mail was collected even if a ticket was already created with an earlier email

htdocs/emailcollector/class/emailcollector.class.php
```diff
@@ -3043,7 +3063,7 @@ public function doCollectOneCollector($mode = 0)
 							// Create ticket
 							$tickettocreate = new Ticket($this->db);
 
-							$alreadycreated = $tickettocreate->fetch(0, '', '', $msgid);
+							$alreadycreated = $tickettocreate->fetch(0, '', $trackid, $msgid);
```
eldy, line 3066: Do we really need to add $trackid ?
$msgid is already unique so if we found a ticket with the msgid, we know it is already recorded in database.

Hystepik: Yes but the $msgid is never the same between the mails, so if a ticket is already created and we receive a mail that responds to the mail that had created the ticket then the new mail needs to not create a ticket and be linked to the ticket (Done with another operation in collector config)

eldy: > Yes but the $msgid is never the same between the mails, so if a ticket is already created and we receive a mail that responds to the mail that had created the ticket then the new mail needs to not create a ticket and be linked to the ticket (Done with another operation in collector config)

I see.

Note that if the ticket was created with the msgid and trackid was not set, if we collect another email for the same trackid, we will not find the email with the fetch() because into the fetch, the test on trackid and msgid is a AND, and we need a OR.
I think we must do a
$alreadycreated1 = $tickettocreate->fetch(0, '', $trackid);
$alreadycreated2 = $tickettocreate->fetch(0, '', '', $msgid);
$alreadycreated = $alreadycreated1 + $alreadycreated2;



Bur we need OR.

PR #29359, opened 2024-04-16
NEW: new const MAIN_ORDERLIST_CREATEBILLS_ONEBYTHIRD

altairis-noe:
NEW: new const to allow preselecting yes in the bill orders massaction from order list

PR #29394, opened 2024-04-18
Fix #29359 Add codespell exception for contributor email

mdeweerd:
Fix #29359 Add codespell exception for contributor email


altairis-noe:
Thanks a lot @mdeweerd !

PR #29395, opened 2024-04-18
NEW: Add shipments and reception stats on product stat tab

jyhere:


htdocs/core/lib/product.lib.php
```diff
@@ -571,6 +572,46 @@ function show_stats_for_company($product, $socid)
 		print '</tr>';
 	}
 
+	// Shipments
+	if (isModEnabled('expedition') && $user->hasRight('expedition', 'lire')) {
```
eldy, line 576: You shoul be able to use the english name now
 isModEnabled('shipment')

eldy (changes requested):

PR #29396, opened 2024-04-18
Fix: replenish.php: no virtual stock shown for warehouses that have no physical stock

atm-florianm:
## Issue

Currently, the replenish page can be filtered by warehouse. If you use this filter in the "virtual stock" mode, it shows a column with the virtual stock of every product for the selected warehouse.

However, when the product doesn't currently have any physical stock in that warehouse, the virtual stock is not computed even though it might be non-zero (for instance if the warehouse is the default warehouse for this product and there are supplier orders for that product, or if you allow negative stock and there are future shipments with lines referring to this warehouse and this product).

This is because `load_virtual_stock()` loops over keys of `$this->stock_warehouses` (an array indexed by warehouse), whose keys are initialized when computing the physical stock. Therefore, warehouses with no physical stock don't have a key in this array and are not considered for virtual stock computation.

## Proposed solution

This fix initializes the array `$this->stock_warehouses` with empty objects for every warehouse.

An alternative solution would be, in `load_virtual_stock()`, to loop over all warehouses instead of those listed as keys of `$this->stock_warehouses`. It would have a smaller memory footprint (no need to keep an array of mostly empty values for possibly thousands of warehouses) but this might be harder to read.

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -5830,6 +5830,22 @@ class Product extends CommonObject
 			}
 		}
 
+		if (! preg_match('/novirtual/', $option)) {
+			// initialize $this->stock_warehouse for all warehouses, even those with no current physical stock because
+			// they may still have a non-zero virtual stock
+			$sql = 'SELECT rowid FROM ' . $this->db->prefix() . 'entrepot WHERE entity IN (' . getEntity('stock') . ')';
+			$result = $this->db->query($sql);
+			if ($result) {
+				while ($obj = $this->db->fetch_object($result)) {
+					$stockWarehouse = $this->stock_warehouse[$obj->rowid] = new stdClass();
+					$stockWarehouse->id = 0;
+					$stockWarehouse->real = 0;
+					$stockWarehouse->virtual = 0;
+					$stockWarehouse->detail_batch = array();
+				}
+			}
+		}
+
 		$sql = "SELECT ps.rowid, ps.reel, ps.fk_entrepot";
 		$sql .= " FROM ".$this->db->prefix()."product_stock as ps";
 		$sql .= ", ".$this->db->prefix()."entrepot as w";
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -5833,7 +5833,7 @@ class Product extends CommonObject
 		if (! preg_match('/novirtual/', $option)) {
 			// initialize $this->stock_warehouse for all warehouses, even those with no current physical stock because
 			// they may still have a non-zero virtual stock
-			$sql = 'SELECT rowid FROM ' . $this->db->prefix() . 'entrepot WHERE entity IN (' . getEntity('stock') . ')';
+			$sql = "SELECT rowid FROM " . $this->db->prefix() . "entrepot WHERE entity IN (" . getEntity('stock') . ")";
 			$result = $this->db->query($sql);
 			if ($result) {
 				while ($obj = $this->db->fetch_object($result)) {
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -5830,19 +5830,19 @@ class Product extends CommonObject
 			}
 		}
 
-	// initialize $this->stock_warehouse for all warehouses, even those with no current physical stock because
-	// they may still have a non-zero virtual stock
-	$sql = "SELECT rowid FROM " . $this->db->prefix() . "entrepot WHERE entity IN (" . getEntity('stock') . ")";
-	$result = $this->db->query($sql);
-	if ($result) {
-		while ($obj = $this->db->fetch_object($result)) {
-			$stockWarehouse = $this->stock_warehouse[$obj->rowid] = new stdClass();
-			$stockWarehouse->id = 0;
-			$stockWarehouse->real = 0;
-			$stockWarehouse->virtual = 0;
-			$stockWarehouse->detail_batch = array();
+		// initialize $this->stock_warehouse for all warehouses, even those with no current physical stock because
+		// they may still have a non-zero virtual stock
+		$sql = "SELECT rowid FROM " . $this->db->prefix() . "entrepot WHERE entity IN (" . getEntity('stock') . ")";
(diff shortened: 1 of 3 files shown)


eldy:
The solution hurt me because it change code into product.class.php
when it should fix by changing code into the incriminated page so product/stock/replenish.php

Can't you try to add a fix inside this page ?

atm-florianm:
Sorry for not replying for a long time. I haven't spent time on that issue for weeks, so I forgot many of the specificities.

I will try to book some time in the next few weeks to help move this PR forwards.

atm-florianm:
@eldy this PR was brought to my attention again because we encountered a case that looks totally different on the outside, but traces back to the same underlying problem this PR attempts to address.

This more recent case was found in v20.0, with the conf `STOCK_ALLOW_NEGATIVE_TRANSFER` disabled:
* our customer creates an order with two identical lines having a qty of 1 and the same product (which is not serialized — I didn't try it with serialized products)
* the customer only has one warehouse and that product has a physical stock of 1 in that warehouse
* the customer creates a shipment from the order, validates it, then closes it.
* expected behavior: error message (like “*cannot create stock movement because negative physical stock is not allowed*”)
* actual behavior: the shipment is closed and the physical stock is decreased to -1

In my investigation, I found that the class `MouvementStock` does have the desired physical stock checks. However, the check fails to work as expected when the current stock for a given warehouse is 0.

Here is the test:
```php
if (isset($product->stock_warehouse[$entrepot_id]) && (empty($product->stock_warehouse[$entrepot_id]->real) || $product->stock_warehouse[$entrepot_id]->real < abs($qty))) {
```
IMO, the test itself is OK, but the data in `$product->stock_warehouse` is incomplete.

In our customer's case, `$product->stock_warehouse[$entrepot_id]` is not set because of how `Product::load_stock` works: it queries `llx_product_stock` for the product, then loops over the results to populate `$product->stock_warehouse`. By doing this, it only creates keys for warehouses where this product is present and this causes the test cited above to fail.

By populating the keys of `$product-stock_warehouse` for all warehouses and not just those where the product has physical stock, it would solve both the initial problem and this new problem.

Of course, this is but one solution.

Another solution would be to change the failing tests to no longer assume that keys are populated (instead, we would assume that missing key = empty warehouse for the product). I think this would create more work and required more testing because there might be other causes (causes I have not yet thought of) for a missing warehouse key.

htdocs/product/class/product.class.php
```diff
@@ -5830,6 +5830,22 @@ public function load_stock($option = '', $includedraftpoforvirtual = null, $date
 			}
 		}
 
+		if (! preg_match('/novirtual/', $option)) {
```
atm-florianm, line 5833:
```suggestion
```

htdocs/product/class/product.class.php
```diff
@@ -5830,6 +5830,22 @@ public function load_stock($option = '', $includedraftpoforvirtual = null, $date
...
+			if ($result) {
+				while ($obj = $this->db->fetch_object($result)) {
+					$stockWarehouse = $this->stock_warehouse[$obj->rowid] = new stdClass();
+					$stockWarehouse->id = 0;
+					$stockWarehouse->real = 0;
+					$stockWarehouse->virtual = 0;
+					$stockWarehouse->detail_batch = array();
+				}
+			}
+		}
```
atm-florianm, line 5847:
```suggestion
```

htdocs/product/class/product.class.php
```diff
@@ -5830,6 +5830,22 @@ public function load_stock($option = '', $includedraftpoforvirtual = null, $date
...
+
```
atm-florianm, line 5848: Actually, for the second problem, it would be better to do this initialization regardless of whether virtual stock is requested or not.

```suggestion
	// initialize $this->stock_warehouse for all warehouses, even those with no current physical stock because
	// they may still have a non-zero virtual stock
	$sql = "SELECT rowid FROM " . $this->db->prefix() . "entrepot WHERE entity IN (" . getEntity('stock') . ")";
	$result = $this->db->query($sql);
	if ($result) {
		while ($obj = $this->db->fetch_object($result)) {
			$stockWarehouse = $this->stock_warehouse[$obj->rowid] = new stdClass();
			$stockWarehouse->id = 0;
			$stockWarehouse->real = 0;
			$stockWarehouse->virtual = 0;
			$stockWarehouse->detail_batch = array();
		}
	}
```

eldy:
You point the bug. It is on the line
`if (isset($product->stock_warehouse[$entrepot_id]) && (empty($product->stock_warehouse[$entrepot_id]->real) || $product->stock_warehouse[$entrepot_id]->real < abs($qty))) {`

The isset was introduced in v20 to avoid a warning on test $product->stock_warehouse[$entrepot_id]->real when $product->stock_warehouse[$entrepot_id] is not set.

The bug is a side effect of a bad addition.

Can you test if this work:
`if (empty($product->stock_warehouse[$entrepot_id]) || empty($product->stock_warehouse[$entrepot_id]->real) || $product->stock_warehouse[$entrepot_id]->real < abs($qty)) {
`

It assumes that non populated keys are 0 (what is the current behaviour).

eldy:
This is the expected behaviour to have entry into associative array not defined when there is not stock. Same than defined to 0.
If we have 1 million of product references and only 10 in stocks, the array should contains 10 entries and not 1 million.

For the second point i tried a fix with ec1ea71175025787170d4ce0b4dbb99a91b4609d

eldy:
> This is the expected behaviour to have entry into associative array not defined when there is not stock. Same than defined to 0.
> If we have 1 million of product references and only 10 in stocks, the array should contains 10 entries and not 1 million.
> 
> For the second point i tried a fix with ec1ea71175025787170d4ce0b4dbb99a91b4609d

Can this ticket be claused now this patch has been introduced
https://github.com/Dolibarr/dolibarr/commit/ec1ea71175025787170d4ce0b4dbb99a91b4609d

atm-florianm:
I haven't been assigned to a project that uses stock in ages, so I can't retest, but the fix you presented should work at least for the cases we discussed. I am closing the PR.

PR #29408, opened 2024-04-19
NEW Option to transfer only reconciliated lines from bank

rycks:
That PR add a checkbox on bank import to be able to not import non-reconciliated lines

image


rycks:
Add default option and better choice with a drop down

image

eldy:
For term consistency, can you rename Import ...
 Into 
Transfer ...

rycks:
for sure !

rycks:
@eldy done

PR #29409, opened 2024-04-19
add new column to store shipping date from warehouse

frederic34:


eldy:
Don't we already jave the date of stock movement into lllx_stock_mouvement ?

eldy:
My understanding is that you need is the "date planned for shipment" (when your products leave your warehouses).
Currently we have date_delivery that is "date when product arrives" at destination. So a different date.

i think we already have a field "date_expedition" that seems to be dedicated for that (date when product leave). But it is commented currently into code as "not used (deprecated)". But i think it is the field you are looking for. So instead of adding a field, we may start to use the existing field "date_expedition" ?
So we can replace the "not used (deprecated)" by "Used for the date planned when product must leave the warehouse".

frederic34:
Thanks for the advice, I'll look into how to use the existing field and if it's not already being used by external modules. I'll open another pull request with an improved solution once I have one.

PR #29415, opened 2024-04-19
New: Mail template for contacts

fmarcet:
Add the possibility of having email templates for sending to contacts.

I have also modified the .lang file of the en_US language but I don't know if it is the correct procedure.


eldy:
> Add the possibility of having email templates for sending to contacts.
> 
> I have also modified the .lang file of the en_US language but I don't know if it is the correct procedure.

It is the good process for en_US lang.

PR #29416, opened 2024-04-19
FIX: It was not possible to mass edit VAT to 0% on a supplier invoice

spooky360:
On a supplier invoice, when you want to set all lines VAT to 0, the check performed in card.php fails and an error occurs. The 0 value was interpreted as false so we never enter in the "if" condition. We now check that the value is not an empty string.


spooky360:
changed branch

Issue #29421, opened 2024-04-20
BUG: API GET /members/types/{id} says it deletes member type


JonBendtsen:
### Bug

BUG: API GET /members/types/{id} says it deletes member type

{
  "success": {
    "code": 200,
    "message": "Member type deleted"
  }
}

but GET should never do that, only DELETE

### Dolibarr Version

20.0.0-alpha

### Environment PHP

 8.2.7

### Environment Database

10.11.7-MariaDB-1:10.11.7+maria~ubu2204

### Steps to reproduce the behavior and expected behavior

enable API module, create API key for user, visit the API explorer, scroll down and find GET /members/types/{id}, fill in a number, any number, even 89327434 and it says

{
  "success": {
    "code": 200,
    "message": "Member type deleted"
  }
}

### Attached files

![get_member_types_says_delete_membeR_types](https://github.com/Dolibarr/dolibarr/assets/2551481/0cb13c2d-7408-4a73-a7a4-4c45c77f9b64)

PR #29423, opened 2024-04-20
FIX: API GET /members/types/{id} says it deletes member type #29421

JonBendtsen:
FIX: API GET /members/types/{id} says it deletes member type #29421
Fix also includes a fix for giving an error message when trying to delete a member type that does not exist.

### DIFF
--- a/htdocs/adherents/class/adherent_type.class.php
+++ b/htdocs/adherents/class/adherent_type.class.php
@@ -522,9 +522,10 @@ class AdherentType extends CommonObject
 
 				// fetch optionals attributes and labels
 				$this->fetch_optionals();
+				return $this->id;
+			} else {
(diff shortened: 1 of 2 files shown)


JonBendtsen:
get_member_types_fixed

JonBendtsen:
API GET member types now actually gets member types

get_member_types_fixed

JonBendtsen:
API GET member types now actually gets member types


delete_member_types_fixed

JonBendtsen:
During fixing this bug I discovered another - DELETE member types always reports success

delete_member_types_always reports_success

JonBendtsen:
how it actually reports membertype not found when trying to delete one which does not exists

delete_not_found

PR #29431, opened 2024-04-22
fix phpstan

frederic34:
Parameter `#2` $socid (int) of method CompanyBankAccount::fetch() should be compatible with parameter $ref (string) of method Account::fetch()


eldy:
As this will create regression in external module using the fetch of CompanyBankAccount class, can you add a mention into the ChangeLog file, section "WARNING".

PR #29435, opened 2024-04-22
FIX: missing comma in query

mc2rcanarslan:


fappels:
Seems to be regression down to 14.0

PR #29441, opened 2024-04-23
Fix18.0 29424 member api

### LINKED ISSUE #29424: BUG: in API GET member categories
### Bug

If you in dolibarr version 20-alpha tries to use the API to get member categories it does not respect when you supply a non existing member id, where as in version 18.0.5 it does respect and gives an error.

### Dolibarr Version

20.0.0-alpha

### Environment PHP

8.2.7

### Environment Database

10.11.7-MariaDB-1:10.11.7+maria~ubu2204

create some member categories and assign some members to them. use the API to get the member categories for a non existing member

expected is an error better than the one in 18.0.5, because in 18.0.5 it says that the category does not exist, but it is the member that does not exist.

### Attached files

v20 ignores non existing members

get_member_categories_v20_ignores_nonexisting_member

GregoryBlemand:
# FIX #29424 part 18.0
Fix member and thirdparty fetch and bad error messages


eldy:
The PR contains both a fix and refactoring. 
Fix must contains the MINIMAL changes to fix the bug. Because it constains refactoring i pushed the PR into develop branch. But there is a lot of error reported by the CTI (like identation broken).

JonBendtsen:
I think this PR can be closed since the bug  seems to have been fixed

https://github.com/Dolibarr/dolibarr/issues/29424

PR #29462, opened 2024-04-24
fix: when invoice is created by WORKFLOW_ORDER_AUTOCREATE_INVOICE on ORDER_CLOSE, the invoice must have the default bank account of the thirdparty is it's empty on order

FHenry:


eldy:
If the bank account is empty on order, it is probable because the user has removed the value on the order (otherwise it is set with the default value of the thirdparty on order).
So if the user has removed the default value on order, it is surely because it don't want to have a bank account forced on order and on invoice.
If a default bank account has been defined on thirdparty, it will be set on order and already propagated on invoice.
And if empty on order, it is because user has removed it on purpose so don't want to have it forced on PDF. So no reason to have it forced on PDF. Di i miss something ?

FHenry:
I will be agree if on clone order on another thirdparty that default bank account on the clones one where be taken from the  "another thirdparty"
and but bank account on order is optionnal (const BANK_ASK_PAYMENT_BANK_DURING_ORDER not enabled by default)

So may be If I test this BANK_ASK_PAYMENT_BANK_DURING_ORDER is empty, it means user cannot set bank on order, then it have to come from thirdparty ?

PR #29471, opened 2024-04-25
FIX #29470 relate format

### LINKED ISSUE #29470: We can no longer relate if the previously free format had not been initiated in YYYMM
### Bug

see that : https://github.com/Dolibarr/dolibarr/commit/65b0086cfb7b5cd666a1f523397f89bd9aac27ff

@klugfr said : 

Could it be possible to have a way to disable/enable this from the admin?

We've been naming our bank receipt according to the bank for years.
For this month we have for example: BPOP202310, SG202310, CIC202310, etc.

Forcing this change means changing our naming convention and that all bank receipts for the same month will have the same name (YYYYMM), less easy to read/check.

### Dolibarr Version

19

### Environment PHP

_No response_

### Environment Database

_No response_

_No response_

### Attached files

_No response_

cmfpmatik:
FIX #[29470](https://github.com/Dolibarr/dolibarr/issues/29470)

### DIFF
--- a/htdocs/compta/bank/bankentries_list.php
+++ b/htdocs/compta/bank/bankentries_list.php
@@ -822,7 +822,7 @@ if ($resql) {
 	// Form to reconcile
 	if ($user->hasRight('banque', 'consolidate') && $action == 'reconcile') {
 		print '<div class="valignmiddle inline-block" style="padding-right: 20px;">';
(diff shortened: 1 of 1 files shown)


akhamar:
This fix is not yet available in 19.0.3
Do we know when this will be available in dolibarr, and on which version ?

akhamar:
Only available on 20.x, can it be ported to 19.x ?

Issue #29458, opened 2024-04-24
incorrect page numbering in PDF


GildDart:
### Bug

In v18.0.5 (php 8.0.30) the page numbering in the Propale pdf is incorrect. It is in the form: 1/2, 2/3, 3/4, 4/4 instead of 1/4, 2/4, 3/4, 4/4 (azure and cyan models).

On the same server and configuration it is correct for v17.0.3

### Dolibarr Version

18.0.5

### Environment PHP

8.0.3

### Environment Database

	10.6.17-MariaDB

sonikf:
Hi @GildDart 
Can you check if PDF_USE_GETALIASNBPAGE_FOR_TOTAL set to 1 fixes your problem and if numbering is in correct position?

GildDart:
Great it works !!!
Thank you sonikf.

sonikf:
Hi @GildDart 
Can you reopen? This is actually a bug and is reproducible in 19.0.1 and 20.0.0-alpha

GildDart:
Done...

fappels:
Have put this on high priority, had more complaints on this. Some Authorities do not accept invoices with incorrect page numbering.

sonikf:
Hi @fappels,
Can we collaborate on this?
Problem is here
https://github.com/Dolibarr/dolibarr/blob/e2eb4e9168fc61de192f96122d1ad87541b8cee2/htdocs/core/lib/pdf.lib.php#L1355-L1365
For countries using Freemono or DejaVuSans  instead of default Helvetica font only const PDF_USE_GETALIASNBPAGE_FOR_TOTAL works but renders wrong 

![2024-04-25_11-44](https://github.com/Dolibarr/dolibarr/assets/93765174/c03503d9-cdb3-46c5-ad4d-b884062327a0)
line 1365 should change from
`$pdf->MultiCell(18, 2, $pagination, 0, 'R', 0);`
to
`$pdf->MultiCell(30, 2, $pagination, 0, 'R', 0);`

fappels:
Maybe make PR for V18 like:
```
if (in_array(pdf_getPDFFont($outputlangs), array('Freemono',  'DejaVuSans'))) {
    $pdf->MultiCell(30, 2, $pagination, 0, 'R', 0);
} else {
    $pdf->MultiCell(18, 2, $pagination, 0, 'R', 0);
}
```
[...]

PR #29476, opened 2024-04-25
FIX incorrect page numbering in PDF #29458

fappels:


eldy:
Can you provide a process to reproduce the error before the fix ?

PR #29477, opened 2024-04-25
FIX #23185 translation of permissions

### LINKED ISSUE #23185: Read Permissions BOM and MRP
### Feature Request

Currently, it's not possible to give a user READ permissions to a BOM or MRP. 
As for all other Dolibarr modules this functionality is available, it would be useful for BOM and MRP to be inline.

BOM-MRP-Read-permissions

### Use case

Sales people who want to know when their customer production order is ready. 
Junior production people who want to read BOM but are not authorized to make changes. 

### Suggested implementation

_No response_

### Suggested steps

_No response_

AlbanDurrheimer:
# FIX #23185 translation of permissions
The commit 6c521073ea9215fb7baddca5e48345b20b1ab5a4 has changed the permission id but has not updated the translation.  
This problem exists in all versions since 14.0.


eldy:
Because we can only commit change into files for en_US language i did the marge manually in develop.

Other files will be translated by transifex.

PR #29485, opened 2024-04-26
Prevents PHP Warnings on multicurrency_rate.php

PsyCrow-code:
# FIX PHP Warnings on multicurrency_rate.php
Prevents PHP Warnings about undefined variables
$type
$texte
$newcardbutton
and Undefined array key "nbfield"


htdocs/multicurrency/multicurrency_rate.php
```diff
@@ -64,6 +64,10 @@
 $sortfield 			= GETPOST('sortfield', 'aZ09comma');
 $sortorder 			= GETPOST('sortorder', 'aZ09comma');
 $page = (GETPOST("page", 'int') ? GETPOST("page", 'int') : 0);
+// Prevents PHP Warnings
+$type = ''; 			// PHP Warning:  Undefined variable $type on line 440
```
eldy, line 68: Can you remove all comment // PHP Warning ...
This is already provided into the PR title and we can't add such comment at each init of a variable (init should be the standard behaviour so no comment required).

PsyCrow-code (commented):
Unnecessary inline comments removed

PR #29488, opened 2024-04-26
FIX: PHP Warning:  Undefined array key "totalam" on line 1890

PsyCrow-code:
# FIX PHP Warning:  Undefined array key "totalam" on line 1890

Avoid PHP Warning:  Undefined array key "totalam" on line 1890


eldy:
Duplicate of #29489

PR #29497, opened 2024-04-26
NEW : Edit ticket

atm-irvine:
# NEW|New Edit ticket
This commit allows you to edit a ticket at once !

image
image


zahwyrix:
Great feature, I wonder why it was done differently on the tickets.
On the other hand, if I may give my opinion, I suggest you remove the pencil, which would :
1 - lighten the interface
2 - lighten the code by removing the form processing in particular
3 - Reduce redundancy
In the final analysis, I see very little point in these pencils; for me, they should remain specific to document lines.

PR #29498, opened 2024-04-26
FIX:  made invalid code shape error more user friendly

YannisHoareau:
When a barcode had an invalid shape, the page would die, now we use the error, catch it and send a pop up the the user so that he is told about what could have caused this error.

### DIFF
--- a/htdocs/barcode/printsheet.php
+++ b/htdocs/barcode/printsheet.php
@@ -255,10 +255,16 @@ if ($action == 'builddoc') {
 
 			if (!$mesg) {
 				$outputlangs = $langs;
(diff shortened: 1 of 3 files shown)


htdocs/barcode/printsheet.php
```diff
@@ -231,7 +231,7 @@
 						'code'=>$code,
 						'encoding'=>$encoding,
 						'is2d'=>$is2d,
-						'photo'=>$barcodeimage	// Photo must be a file that exists with format supported by TCPDF
+						'photo'=>!empty($barcodeimage) ??  ''	// Photo must be a file that exists with format supported by TCPDF
```
eldy, line 234: What   !empty($barcodeimage) ?? ''  means ?

YannisHoareau:
I changed this line (I changed it again because I made a mistake) because before, the barcode pdf would never be loaded because of an undefined variable. It was fixed in 18.0 version but not in 17.0. Also, my first fixing goal here was the not user friendly logic error when the barcode had an invalid shape (it was dying and throwing outside dolibarr app without saying what could have happened)

PR #29507, opened 2024-04-27
NEW: Added hook for printing banner in card

kubajznik:
NEW
To be able to print custom information on top of cards. In our company we use it for related links and back-to-links which are different than list.

### DIFF
--- a/htdocs/commande/card.php
+++ b/htdocs/commande/card.php
@@ -2485,8 +2485,27 @@ if ($action == 'create' && $usercancreate) {
 		}
 		$morehtmlref .= '</div>';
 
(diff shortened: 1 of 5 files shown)


eldy:
It seems you want to modify morehtmlref, morehtmlstatus, ... before calling dol_banner that call showrefnav.
I pushed a commit here https://github.com/Dolibarr/dolibarr/commit/5ff2138b17688ab06080748b51ab5aa9cb5f6f88
line 2851.
With this you can use the existing hook  "formDolBanner" inside dol_banner to get and modify all values of morehtmlref, morehtmlstatus, ...
This avoid to add a new hook and new code on each page.

Can you try to reach your goal with this patch instead ?

eldy:
Closed as an already existing hooks can now be used.

PR #29509, opened 2024-04-27
How many errors remain before reaching the next phpstan analysis level?

frederic34:


eldy:
Note @frederic34 : You can get the info here:   cti.dolibarr.org

frederic34:
@eldy yes I know, but I find more errors in local

PR #29510, opened 2024-04-27
FIX : Display the date according to user language on subsitutions

ksar-ksar:
See : https://www.dolibarr.fr/forum/t/format-date-dune-variable-de-substitution-dans-un-modele-de-courriel/40705

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -7780,12 +7780,12 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 				$substitutionarray['__MEMBER_PHONEPRO__'] = (isset($object->phone_perso) ? dol_print_phone($object->phone_perso) : '');
 				$substitutionarray['__MEMBER_PHONEMOBILE__'] = (isset($object->phone_mobile) ? dol_print_phone($object->phone_mobile) : '');
 				$substitutionarray['__MEMBER_TYPE__'] = (isset($object->type) ? $object->type : '');
-				$substitutionarray['__MEMBER_FIRST_SUBSCRIPTION_DATE__']       = dol_print_date($object->first_subscription_date, 'dayrfc');
-				$substitutionarray['__MEMBER_FIRST_SUBSCRIPTION_DATE_START__'] = (isset($object->first_subscription_date_start) ? dol_print_date($object->first_subscription_date_start, 'dayrfc') : '');
-				$substitutionarray['__MEMBER_FIRST_SUBSCRIPTION_DATE_END__']   = (isset($object->first_subscription_date_end) ? dol_print_date($object->first_subscription_date_end, 'dayrfc') : '');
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -7780,12 +7780,12 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
...
 				$substitutionarray['__MEMBER_TYPE__'] = (isset($object->type) ? $object->type : '');
-				$substitutionarray['__MEMBER_FIRST_SUBSCRIPTION_DATE__']       = dol_print_date($object->first_subscription_date, 'dayrfc');
-				$substitutionarray['__MEMBER_FIRST_SUBSCRIPTION_DATE_START__'] = (isset($object->first_subscription_date_start) ? dol_print_date($object->first_subscription_date_start, 'dayrfc') : '');
-				$substitutionarray['__MEMBER_FIRST_SUBSCRIPTION_DATE_END__']   = (isset($object->first_subscription_date_end) ? dol_print_date($object->first_subscription_date_end, 'dayrfc') : '');
-				$substitutionarray['__MEMBER_LAST_SUBSCRIPTION_DATE__']        = dol_print_date($object->last_subscription_date, 'dayrfc');
-				$substitutionarray['__MEMBER_LAST_SUBSCRIPTION_DATE_START__']  = dol_print_date($object->last_subscription_date_start, 'dayrfc');
-				$substitutionarray['__MEMBER_LAST_SUBSCRIPTION_DATE_END__']    = dol_print_date($object->last_subscription_date_end, 'dayrfc');
+				$substitutionarray['__MEMBER_FIRST_SUBSCRIPTION_DATE__']       = dol_print_date($object->first_subscription_date, 'day');
```
eldy, line 7783: By definition, substitution array is an array in memory. So data must be saved into standard date format.
To save a value into end user local format, you can user _LOCALE_ at end of substitution key. In suh a case, the format can be 'day' and the timezone must be 'tzserver'

frederic34:
@mdeweerd is it possible that codespell checks commit message to avoid errors in changelog?

mdeweerd:
@frederic34 I think that should be possible with the commit-msg hook.

[[pre-commit](https://pre-commit.com/#pre-commit)](https://pre-commit.com/#commit-msg)

However, just adding the following does not work, so a "local" hook may be needed:

```yaml
      - id: codespell
        stages: [commit-msg]
        name: codespell-commit-msg
```
I found another project where it is configured that way so I wonder if it just does not work not for me (my setup being a bit "speciam") - see https://forge.softwareheritage.org/differential/changeset/?ref=117189 .


I do not know when/how the Changelog is generated but if it is ensured that pre-commit is run after generating it then the Changelog contents will be checked.

frederic34:
@mdeweerd thank's, for the changelog see here https://github.com/Dolibarr/dolibarr/commit/f12378d732ff3f35ff0a5e5c8467fe522f7385ba

PR #29517, opened 2024-04-28
New - updated discount splitting function in Remx.php

henrynopo:
# NEW
The existing discount splitting function needs to manually input the two smaller amounts in main currency. When user changes one of the amounts, the other one is not auto calculated.

The proposed change allows key in only one amount, either in main currency or original currency and the discount will be split into two discounts accordingly.

The splitting icon will be shown besides the TTC amount in main currency and that in original currency, respectively to different the splitting in either main currency or original currency.


henrynopo:
Relocate the split icon to besides the amounts (both main currency and original currency)
image

The split confirmation will pop up and user can input the first value he wants, the second value will be calculated automatically. 
image
image

PR #29522, opened 2024-04-28
QUAL Normalize inclusion of main.inc.php with require_once __DIR__

Vaadasch:
# QUAL|Qual Normalization of calls to main.inc.php
In the event of a page called by another page, the relative path used to include/require the main.inc.php is broken.

As it is mainly require that is used, the error is interruptive.

Modification of all the files of the project to normalize the use of a "require_once" and the use of `__DIR__` as a base on which use the "relative" path to find main.inc.php.

# moduleBuilder
Inside this PR, modification of the template behavior to find the main.inc.php from the moduleBuilder templates.
If a main.inc.php at the closest path of the file (current dir) is present, tries to include (require_once) it.
And if not present, search inside parent folder.
And if not present, search inside parent's parent folder.
And so on until finding one. Ultimately the `/var/www/htdocs/main.inc.php`. 
To prevent infinite while loop, if it doesn't find it, a counter of 15.

One can have its own main.inc.php inside its module, as long as it include the same code block in order to include the "main" main.inc.php.
Probably useless but that come with the new behavior.

PS : QUAL does mean quality, right ?


htdocs/accountancy/admin/account.php
```diff
@@ -25,7 +25,7 @@
  */
 
 // Load Dolibarr environment
-require '../../main.inc.php';
+require_once __DIR__.'/../../main.inc.php';
```
eldy, line 28: the main.inc.php must never be included several times, it is the init of env that must be done once and only once.
So we must use the require and not require_once

Vaadasch: I don't understand what you're saying.
>the main.inc.php must never be included several times SO we use require instead of require_once

Isn't that the very purpose of require_once ?

Vaadasch: Hi, @eldy, 
>A page should never include another page (this will generates conflict in mime type, duplicate http header, content policies rules, and more evil side effects.

Have you encounter these problems or is that a general assumption ? Is that not dependant on how you code your page ?

As I finally succeed to make work with hooks, can you confirm me that is the way I search ?

Did you happen to see the modifications that i've done inside the templates of modulebuilder ? I find this more elegant. Will that be interesting to include that in a specific PR ?

Vaadasch:
Just using hooks. That is what they are for.

PR #29527, opened 2024-04-29
FIX : Default action type code when send mail.

atm-irvine:
# FIX|Fix Default action type code when send mail. Changed from AC_OTH_AUTO to AC_EMAIL
By default, the action type code when you send an email is AC_OTH_AUTO. I think it is more relevant to set it to AC_EMAIL.

### DIFF
--- a/htdocs/core/actions_sendmails.inc.php
+++ b/htdocs/core/actions_sendmails.inc.php
@@ -382,7 +382,7 @@ if (($action == 'send' || $action == 'relance') && !GETPOST('addfile') && !GETPO
 					// Initialisation of datas of object to call trigger
 					if (is_object($object)) {
 						if (empty($actiontypecode)) {
-							$actiontypecode = 'AC_OTH_AUTO'; // Event insert into agenda automatically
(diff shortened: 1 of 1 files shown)


htdocs/core/actions_sendmails.inc.php
```diff
@@ -382,7 +382,7 @@
 					// Initialisation of datas of object to call trigger
 					if (is_object($object)) {
 						if (empty($actiontypecode)) {
-							$actiontypecode = 'AC_OTH_AUTO'; // Event insert into agenda automatically
+							$actiontypecode = 'AC_EMAIL'; // Event insert into agenda automatically
```
eldy, line 385: There is 2 types of events:
Event enter manually by the user into the agenda.
Events created automatically by the system for tracking purposes.

An email send from a screen is the second one. So it must not appears into the agenda of the user among its rendez vous but into the agenda of tracking events. So $actiontypecode (the type auto / non auto) must be 'AC_OTH_AUTO'.
There is a column "code" into the table llx_actioncomm that details what king of action has been recorded. Emails are codes like 'AC_..._SENTBYMAIL'

PR #29533, opened 2024-04-29
FIX display massaction for bank entries list

lamrani002:
# FIX|Fix display massaction for bank entries
releve

### DIFF
--- a/htdocs/compta/bank/bankentries_list.php
+++ b/htdocs/compta/bank/bankentries_list.php
@@ -1458,10 +1458,6 @@ if ($resql) {
 		if (getDolGlobalString('MAIN_CHECKBOX_LEFT_COLUMN')) {
 			print '<td>';
 			if (!$objp->conciliated && $action == 'reconcile') {
-				if ($objp->num_releve) {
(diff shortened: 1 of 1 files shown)


eldy:
Don't forget to push a screenshot when you submit a PR that modifies the GUI.

htdocs/compta/bank/bankentries_list.php
```diff
@@ -1462,12 +1457,12 @@
 		// Action column
 		if (getDolGlobalString('MAIN_CHECKBOX_LEFT_COLUMN')) {
 			print '<td>';
-			if ($massactionbutton || $massaction) {   // If we are in select mode (massactionbutton defined) or if we have already selected and sent an action ($massaction) defined
-				$selected = 0;
-				if (in_array($obj->rowid, $arrayofselected)) {
-					$selected = 1;
+			if (!$objp->conciliated && $action == 'reconcile') {
+				if ($objp->num_releve) {
```
eldy, line 1461: We can remove this 3 lines (it was used before to ad a space between the checkbox and the number. They are on 2 different columns now.

Issue #29544, opened 2024-04-30
The company name is empty in the shipping list even though it was entered.


dolibarr95:
# Bug
The company name is not filled even it was informed with param socid.

## Environment
- **Version**: 17.0.2
- **OS**: Linux
- **Web server**: Apache
- **PHP**: litespeed 7.2.34
- **Database**: MySQL or MariaDB 5.5.5-10.6.17-MariaDB
- **URL(s)**: /17/expedition/list.php?sortfield=e.date_delivery&sortorder=asc&begin=&contextpage=shipmentlist&search_company=THE+COMPANY+EXAMPLE

## Expected and actual behavior

1. Go to THE COMPANY EXAMPLE card
2. Click on all the shipping of this company
3. Reorder with an other filter (eg shipping date). At this step the company name shoud be filled in company filter
4. the company name is 'lost'

Can be fixed with  
```php

if ($socid > 0) {
    $soc = new Societe($db);
    $soc->fetch($socid);
    if (empty($search_company)) {
        $search_company = $soc->name;
    }
}
```

Tks for reading

PR #29545, opened 2024-04-30
Fix : Company name should not be empty if socid param declared

dolibarr95:
https://github.com/Dolibarr/dolibarr/issues/29544


ksar-ksar:
Hello,

This would be interested to be merged on V17/18/19 also no ?

PR #29546, opened 2024-04-30
NEW : Add option position['find'] to DolEditor

Vaadasch:
By adding the possibility to find a word inside a content opened by DolEditor, it allow to go to the first line matching the word.

It is usefull for long and diverses files, like the descriptor file of a module.
The module builder was adapted to use this parameter and open the descriptor file directly to the flagged line.

By "word", it allow the possibility to use textual flags inside a document.


htdocs/modulebuilder/template/core/modules/modMyModule.class.php
```diff
@@ -179,8 +182,7 @@ public function __construct($db)
 			$conf->mymodule->enabled = 0;
 		}
 
-		// Array to add new pages in new tabs
-		$this->tabs = array();
```
eldy, line 183: Removing this is required ?

Vaadasch: Like for Permissions, I've moved it after the examples, because i thought seeing some pattern but it seems i've only seen it for Dictionnaries

Vaadasch:
I've manually "revert" the template descriptor file to what it was before my PR, only adding some new flags (HOOKSCONTEXTS, DESCRIPTION_FLAG and TABS).

PR #29557, opened 2024-04-30
19.0 fix search all hidden input

samifilali:
# FIX

The filters applied during the global search are retained even after loading the corresponding object list (example: tier, customer order, etc.)

This poses a problem when you want to apply a new filter on the list (coming from the global search) => the global search filter is kept and cannot be deleted without refresh the page

Can this code be deleted ?


eldy:
What do you mean with "The filters applied during the global search are retained even after loading the corresponding object list " ? Can you provide a screenshot and a process to reproduce ?

samifilali:
When I perform a global search in third parties (1st screenshot), I obtain 7 results (2nd screenshot). However, when I perform a new search in the list, the 'test' filter that I had entered for my global search remains present and is added with the other filters as long as I do not refresh the page (3rd screenshot).

eldy:
I see.
However, it seems your change will break another expected feature. If you make a global search. And just after, you try to to execute a mass action after having selected the first 2 lines for exemple. The the initial filter will be lost and the page that ask the confirmation will be different...
I see no easy way to fix this, except this way:  adding a jquery onchange trigger on fields with name=search_* and when we change a value on such a field, we reset the field search_all.

eldy:

PR #29561, opened 2024-04-30
add column to expedition table

lamrani002:
# NEW|New column for check if expedition was signed


htdocs/install/mysql/migration/19.0.0-20.0.0.sql
```diff
@@ -316,3 +316,5 @@ DELETE FROM llx_c_action_trigger WHERE code = 'BILLREC_AUTOCREATEBILL';
 -- element_element, see https://github.com/Dolibarr/dolibarr/pull/29329
 
 ALTER TABLE element_element ADD COLUMN relationtype	varchar(64) DEFAULT NULL AFTER targettype;
+
+ALTER TABLE llx_expedition ADD COLUMN signed_status smallint DEFAULT NULL AFTER billed;
```
eldy, line 320: Can you push in the same PR the change into the llx_expedition.sql file used for new installation.

PR #29562, opened 2024-04-30
qual: fix phpstan for ProductCombination.class.php

thibdrev:
Construct empty() is not allowed. Use more strict comparison. 

Property Product::$oldcopy (Product) in empty() is not falsy.


htdocs/variants/class/ProductCombination.class.php
```diff
@@ -527,10 +527,10 @@ public function updateProperties(Product $parent, User $user)
 		$child->price_autogen = $parent->price_autogen;
 		$child->weight = $parent->weight;
 		// Only when Parent Status are updated
-		if (!empty($parent->oldcopy) && ($parent->status != $parent->oldcopy->status)) {
+		if (($parent->oldcopy !== null) && ($parent->status != $parent->oldcopy->status)) {
```
eldy, line 530: Can't see why we can't use 
!empty($parent->oldcopy)
Whch value do you have for ->oldcopy that return a different value with !== null ?

thibdrev: Indeed it's a recommendation from PHPStan not to use empty() here.

If you try this simple code on https://phpstan.org/try it will throw a PHPStan error:
```
<?php
class Product { public Product $oldcopy; }
$product = new Product();
$product->oldcopy = $product;
echo empty($product->oldcopy);
```
&rarr; Line: 5. Error: "Property Product::$oldcopy (Product) in empty() is not falsy."    [empty.property](https://phpstan.org/error-identifiers#gbi-empty.property)

And if you also tick the box "Strict rules", PHPStan will throw a second error:
&rarr; Line: 5. Error: Construct empty() is not allowed. Use more strict comparison.     [empty.notAllowed](https://phpstan.org/error-identifiers#gbi-empty.notallowed)


PHPStan probably recommends that to minimize the side effects with `empty()`.
I've check on the Internet and found this one described [here](https://localheinz.com/articles/2023/05/10/avoiding-empty-in-php/#content-inaccessible-instance-property) where the magic method `__isset()` is defined for the object ; and the property is either protected or private.

For example the code below returns `true`:
```
<?php
class Product { 
	protected int $num = 1;
	public function __isset(string $name):bool { return true; }
}
$product = new Product();
var_dump(empty($product->num));
```

There might be other side effects.

Bottom line is: I think PHPStan's recommendation is to be cristal clear on what we're testing.
But PHPStan might be as well overreacting.

@mdeweerd : any idea?

mdeweerd:
The equivalent of empty() needs a correction: the comparison to false is not a strict comparison (see https://www.php.net/manual/en/function.empty.php#refsect1-function.empty-parameters ) .

I made the 'oldcopy' nullable so that 'isempty()' can be true - otherwise the typing suggests that it is always an object.

https://github.com/mdeweerd/dolibarr/blob/1b435259f58eb210062f67d18c760b0be5e1800e/htdocs/product/class/product.class.php#L74-L77

Locally I did not get the phpstan notice that I stil get for the following:
```plaintext
 ------ -----------------------------------------------------------------------------
  ^[[32mLine^[[39m   ^[[32mhtdocs\adherents\class\adherent.class.php^[[39m             
 ------ -----------------------------------------------------------------------------
  824    Property CommonObject::$oldcopy (static(Adherent)) in empty() is not falsy.
  2789   Property CommonObject::$oldcopy (static(Adherent)) in empty() is not falsy.
 ------ -----------------------------------------------------------------------------
```

eldy:
So i pushed the change suggested by mdeweerd.

Issue #29565, opened 2024-04-30
Script to calculate invoice totals from invoice lines


vdegrandpre:
### Feature Request

When importing data from legacy system, I want to have a tool to calculate the facture totals after importing facture and facuredet objects to avoid complexity of ETL scripting.

### Use case

Using an ETL tool to import data  exported in CSV from legacy system to Dolibarr, import process leaves the facture with totals 0, because facturedet created after. 

### Suggested implementation

A successful scenario is to run a Dolibar script that calculate total amounts and taxes on facture that have invoice line totalizing amounts greater than zero.

### Suggested steps

Done.

vdegrandpre:
Closed as merged.

PR #29566, opened 2024-04-30
#29565 added script to calculate facture totals from child items

vdegrandpre:
# NEW|New [*CLI script to calculate facture totals from line items*]
[*Calculate missing (zeroed) invoice totals when there should be non-zero totals. Useful after importation with an external ETL script tool.*]


vdegrandpre:
Hourra!

PR #29568, opened 2024-05-01
Repair function to calculate invoice total from line items

vdegrandpre:
# NEW|New #29565
Invoice script to calculate invoce totals when the are zero and line items exists.
Add missing data after importing invoices with ETL tool.


scripts/invoices/calculate_total_and_taxes.php
```diff
@@ -0,0 +1,118 @@
...
+				FROM
+					".MAIN_DB_PREFIX."facturedet fd
+				WHERE 
+					fd.fk_facture = $obj->rowid";
+			$ressql_calculs = $db->query($sql_calculs);
+			while ($obj_calcul = $db->fetch_object($ressql_calculs)) {
+				$sql_maj = "
+					UPDATE ".MAIN_DB_PREFIX."facture 
+					SET 
+						total_ht = ".($obj_calcul->total_ht ? round($obj_calcul->total_ht, 2) : 0).",
```
eldy, line 98: round on 2 digits will be wrong with some setup in some countries
You can use   
`price2num($obj_calcul->total_ht, 'MT')`

vdegrandpre: I added price2num usage, moved to repair.php and added a moyen to not grab facture that are zero buy have their line's total zero or no line at all (refind SQL query for facture qualification).

PR #29580, opened 2024-05-02
FIX: notification: error 500 in fixed emails due to a bad copy/paste

marc-dll:
The fixed e-mails are not translated but #29407 copy/pasted code from per-user notifications, that do use variable `$outputlangs`


htdocs/core/class/notify.class.php
```diff
@@ -855,20 +855,13 @@ public function send($notifcode, $object, $filename_list = array(), $mimetype_li
 				if (!empty($mailTemplateLabel)) {
 					include_once DOL_DOCUMENT_ROOT.'/core/class/html.formmail.class.php';
 					$formmail = new FormMail($this->db);
-					$emailTemplate = $formmail->getEMailTemplate($this->db, $object_type.'_send', $user, $outputlangs, 0, 1, $labeltouse);
+					$emailTemplate = $formmail->getEMailTemplate($this->db, $object_type.'_send', $user, $langs, 0, 1, $labeltouse);
 				}
 				if (!empty($mailTemplateLabel) && is_object($emailTemplate) && $emailTemplate->id > 0) {
-					// Set output language
-					$outputlangs = $langs;
-					if ($obj->default_lang && $obj->default_lang != $langs->defaultlang) {
```
eldy, line 863: Why removing the use of the targeted user language ?

PR #29584, opened 2024-05-02
NEW: new status shipment in progress + comments

rycks:
We need a new status for expedition to be able to display "expedition is in progress"

I've just added some comments around others status to be more clear because it's sometimes very difficult to associate a status and what dolibarr means with it ...


frederic34:
@rycks I want to add a field to store the shipping date (not the shipping date creation in dolibarr)
https://github.com/Dolibarr/dolibarr/pull/29409

rycks:
@frederic34 nice !

PR #29595, opened 2024-05-03
Fix : Fatal error: Uncaught TypeError: abs(): Argument #1 ($num) must be of type int|float, string given in /home/httpd/vhosts/aflac.fr/domains/dol190.aflac.fr/httpdocs/core/tpl/commonfields_add.tpl.php on line 39

grandoc:


eldy:
Pb in tab/space

PR #29604, opened 2024-05-04
NEW Add option to open files in new tab

Vaadasch:
This is a feature that is recurringly requested along with the modification of "MAIN_DISABLE_FORCE_SAVEAS"

See for exemple : 
https://www.dolibarr.org/forum/t/option-to-open-pdf-in-new-window/8389
https://www.dolibarr.org/forum/t/open-links-in-new-tab/14791
https://www.dolibarr.fr/forum/t/ouverture-des-pdf-dans-une-nouvelle-page/30766
https://www.dolibarr.fr/forum/t/pdf-dans-une-autre-fenetre/8058
https://www.dolibarr.fr/forum/t/ouverture-des-pdf/9447

In this PR, it is adding the option "MAIN_FILES_OPEN_NEWTAB" to open most files inside a newtab if it is not empty.

I will update the wiki once/if this PR is merge.

I suggest is to make this behavior as defaut, as it is the main behavior of all the main tools on the web. 
In compensation, we could have an option "MAIN_FILES_OPEN_SAMETAB".
But it require consent from the experts who prefer otherwise, because the consensus will be status quo.

### DIFF
--- a/htdocs/core/class/html.formfile.class.php
+++ b/htdocs/core/class/html.formfile.class.php
@@ -891,7 +891,7 @@ class FormFile
 						$out .= '<span class="spanoverflow">';
 					}
 					$out .= '<a class="documentdownload paddingright" ';
-					if (getDolGlobalString('MAIN_FILES_OPEN_NEWTAB')) {
+					if (getDolGlobalString('MAIN_FILES_OPEN_NEWTAB') && getDolGlobalString('MAIN_DISABLE_FORCE_SAVEAS')) {
 						print 'target="_blank" ';
 					}
 					print 'href="'.$documenturl.'?modulepart='.$modulepart.'&file='.urlencode($relativepath).($param ? '&'.$param : '').'"';
@@ -1134,7 +1134,7 @@ class FormFile
 
 				// Download
 				$tmpout .= '<li class="nowrap"><a class="pictopreview nowrap" ';
-				if (getDolGlobalString('MAIN_FILES_OPEN_NEWTAB')) {
+				if (getDolGlobalString('MAIN_FILES_OPEN_NEWTAB') && getDolGlobalString('MAIN_DISABLE_FORCE_SAVEAS')) {
 						$tmpout .= 'target="_blank" ';
 				}
 				$tmpout .= 'href="'.DOL_URL_ROOT.'/document.php?modulepart='.$modulepart.'&amp;entity='.$entity.'&amp;file='.urlencode($relativepath).'"';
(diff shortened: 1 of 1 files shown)


Vaadasch:
I've added the condition on MAIN_DISABLE_FORCE_SAVEAS so those who do not have it won't open a new tab when then have the popup to download the file.

I don't know how to solve the missing check

htdocs/core/class/html.formfile.class.php
```diff
@@ -890,7 +890,11 @@ public function showdocuments($modulepart, $modulesubdir, $filedir, $urlsource,
 					} else {
 						$out .= '<span class="spanoverflow">';
 					}
-					$out .= '<a class="documentdownload paddingright" href="'.$documenturl.'?modulepart='.$modulepart.'&file='.urlencode($relativepath).($param ? '&'.$param : '').'"';
+					$out .= '<a class="documentdownload paddingright" ';
+					if (getDolGlobalString('MAIN_FILES_OPEN_NEWTAB') && getDolGlobalString('MAIN_DISABLE_FORCE_SAVEAS')) {
+						print 'target="_blank" ';
```
hregis, line 895: @Vaadasch use `$out.=` instead "print"

htdocs/core/class/html.formfile.class.php
```diff
@@ -890,7 +890,11 @@ public function showdocuments($modulepart, $modulesubdir, $filedir, $urlsource,
...
+					}
+					print 'href="'.$documenturl.'?modulepart='.$modulepart.'&file='.urlencode($relativepath).($param ? '&'.$param : '').'"';
```
hregis, line 897: @Vaadasch use `$out.=` instead "print"

Vaadasch:
Thanks @hregis , i hadn't test enough :disappointed:

eldy:
I don't think we must introduce a second option. Because the feature is dependent on MAIN_DISABLE_FORCE_SAVEAS activation or not, i think it must be a third value of MAIN_DISABLE_FORCE_SAVEAS
MAIN_DISABLE_FORCE_SAVEAS=0   :  Default     (file is auto downloaded)
MAIN_DISABLE_FORCE_SAVEAS=1   :  For to output the file into the browser tab
MAIN_DISABLE_FORCE_SAVEAS=2   :  For to output the file into a browser tab open with target_blank

Because having MAIN_DISABLE_FORCE_SAVEAS off and MAIN_FILES_OPEN_SAMETAB on has no sense.
Don't you think so. if you agree, can you simplify the code to just test on MAIN_DISABLE_FORCE_SAVEAS=2 to add the target_blank ?

Vaadasch:
Of course, you're perfectly right.
It's done

PR #29605, opened 2024-05-04
document tabs display added and link document on two columns

defrance:
please consider new sample display
image


eldy:
A different look and feel enhancement has been introduced in v21, so i close this PR.

defrance:
Seven month later  ...

defrance:
I will still propose a new PR with the positioning on two columns which seems to me more ergonomic and consistent with the majority of Dolibarr pages on two columns too.
We have Screens that are wider and wider and less and less high

Issue #24271, opened 2023-03-19
ECM / GED : work on base tree ?


rycks:
### Feature Request

I would like to be able to have a trashbin and a file versioning solution into dolibarr ECM/GED.

Forget dolibarr objects, i'm really only on GED/ECM part, only thinking about file life cycle.

- File created
- File modifications
- File deleted

For the moment we only have a documents directory where we store files.

One of the main free software project who works on files (nextcloud) have good ideas (i think) : one main directory to store files, one directory to store trash files and an other directory where all file versions are stored. trashed files are suffixed with .d+timestamp and versions files are suffixed with .v+timestamp

Admin sys can choose to put main directory on a fast file system and trash / versions on a slow (cheep) one.

### Use case

Here is what we have on nextcloud:

```
├── files
│       └── projets
│           └── mydoc.md
├── files_trashbin
│   └── files
│       └── otherfile.md.d1666805941
└── files_versions
        └── projets
            ├── mydoc.md.v1666805931
            ├── mydoc.md.v1668361351
            ├── mydoc.md.v1678984851
            ├── mydoc.md.v1678985045
            ├── mydoc.md.v1678985199
            └── mydoc.md.v1678985387
```

### Suggested implementation

Are you ok to add two root files tree:

- documents_versions
- documents_trashbin

If you says "ok" i could try to do the job and make a PR for that.

### Suggested steps

in htdocs/conf/conf.php

```
$dolibarr_main_data_root='/var/www/dolibarr/documents';
$dolibarr_main_data_trash='/var/www/dolibarr/documents_trashbin';
$dolibarr_main_data_versions='/var/www/dolibarr/documents_versions';
```

Then we could update methods delete from ecmfiles.class.php to move file into same path but in $dolibarr_main_data_trash base dir.

And add some methods to ecmfiles.class.php like "archive" for example then dolibarr developpers could make an archive of a file when they need / want / wish.

And some others files to navigate into files history like we can have on a GED/ECM system.

eldy:
What do you think of this

/documents/.trashbin
/documents/.versions
 
Having other directory out of /documents has big impact on existing security code.
We don't need $dolibarr_main_data_trash or $dolibarr_main_data_version. Allowing custom name introduces also more difficulties. We are still able to move this directory on an external support/drive by a symbolic link for example or a filesystem mount. Name should be fixed so we can easily implement.
[...]

PR #29616, opened 2024-05-06
New functionality for save 5 oldest versions of pages

lamrani002:
# NEW|New functionality for save 5 oldest web page by updating dolSavePageContent() function to handle it.


rycks:
Hello all,
could we make a little brainstorm for that sort of things ?

Make a .old.x. ... file ... ok why not, but for example nextcloud has already made a choice for that sort or need, maybe we could do the same choice to be able (one day) to "natively" store dolibarr data into a nextcloud ?

Near than one year before i open a message on french dolibarr forum about that problem:
https://www.dolibarr.fr/forum/t/ged-dolibarr-17-suffixe-des-fichiers-pdf-pseudo-signes-et/

Even if you're working on website part i mean we could use same solution for all needs of "keep archives" of dolibarr documents no ?

eldy:
> ...but for example nextcloud has already made a choice for that sort or need, maybe we could do the same choice...

Sure, why not. 
Because most of us does not know what is the choice of nextcloud, can you describes what is the choice of nextcloud ?
I did not find any doc about this. Asked to chatgpt give me this:

"A previous version of a file named "document.txt" might be stored as something like ".document.txt.v1.timestamp", where "v1" represents the version number and "timestamp" represents the date and time when the version was saved"

Is this good ?
Tihs means archives files are hidden on disk (due to the starting .)  Right ?  Is there any draft of spec about this topic on dolibarr or nextcloud github repositories ?

rycks:
Excuse me dolibarr forum was not the only and the right link, here it is

https://github.com/Dolibarr/dolibarr/issues/24271

rycks:
And the art of my brain last year ... in french

https://inligit.fr/cap-rel/dolibarr/plugin-archivespdf/-/wikis/R%C3%A9flexion%20globale%20GED%20de%20dolibarr

eldy:
@lamrani002 Can you take a look at the summary i added on issue #24271, and try to suggest a rotation of archive files as described into part 1 of the summary to match nexcloud (you can forget part 2 for the moment as Dolibarr is not ready yet).

lamrani002:
i update the code to get like you said
files

PR #29617, opened 2024-05-06
Fix deposit from element with discount

SylvainLegrand:
When you create a deposit (amount type) from an element with more than one VAT rate and a discount line the calculation doesn't take in account the discount line.

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -1508,7 +1508,7 @@ if (empty($reshook)) {
 
 								$TTotalByTva = array();
 								foreach ($srcobject->lines as &$line) {
-									if (!empty($line->special_code)) {
+									if (!empty($line->special_code) && empty($line->info_bits)) {
 										continue;
 									}
 									$TTotalByTva[$line->tva_tx] += $line->total_ttc;


eldy:
Can you provide a screenshot of the web page with the error with steps to reproduce ?

SylvainLegrand:
2024-05-07 - 05_18_59 - DE24-02_0036 - Fiche

2024-05-07 - 05_20_36 - Nouvelle facture


2024-05-07 - 05_21_30 - (PROV1853) - Fiche

htdocs/compta/facture/card.php
```diff
@@ -1508,7 +1508,7 @@
 
 								$TTotalByTva = array();
 								foreach ($srcobject->lines as &$line) {
-									if (!empty($line->special_code)) {
+									if (!empty($line->special_code) && empty($line->info_bits)) {
```
lvessiller-opendsi, line 1511: @SylvainLegrand 
Is this fix only for external modules ? (special_code is not empty)
Maybe, you can fix it by using the "insert_discount" method of "facture" class in the external module or by adding a line without a special code ?

SylvainLegrand: Yes this PR is for lines with special code AND info-bit !

lvessiller-opendsi: ok for me @rycks

rycks: ok for me too ! sorry i have to make an effort to understand the problem :) and an other effort thanks to lvessiller to understand the use of info-bits

rycks (approved):

lvessiller-opendsi (approved):

lvessiller-opendsi:
@eldy 
Everybody is ok for this PR.
What about you ?

eldy:
> @eldy Everybody is ok for this PR. What about you ?

I am not ok, this is why :

The PR is changing a test on "special_code", in case the unstable option MAIN_DEPOSIT_MULTI_TVA is on, to complete it with another test. So, the first thing is to have a look at what is the goal of special code field. 
The wiki doc + code comments in source + parsing dolibarr source shows that the special code can have this values :
`
- 0=default, 
- 1=seems no more used (may be used in past for transport/shipping fees), 
- 2=used by WEEE modules to define if line as a WEEE line, 
- 3=option line (price must not included in total), 
- 4=Sent for cooking (POS modules for restaurant), 
- module id (>= 5)=value for a custom need. 

This value is an additional information and has no effect in most cases in code, except value 3 that means "do not take into account amount, it is an option" (so price must not included in total). 
`
So the only case when we can use "special_code" into a if to know if we can take into account a vat or not is:
- A) when value is 3 according to the definition of special_code
- B) or, like explained line 609 that process the case option MAIN_DEPOSIT_MULTI_TVA is off, when we want to "discard any special lines like shipping fees, ecotax and options" (comment found into line 609) to decide which vat to apply on the discount line.

I don't know yet if the good rule must be A or B (this is a question that we must clarify). The PR after the change, does not respect A rule and B rule. It introduces another rule (called C) taking into account the field info_bits. So:

**If rule must be A:** 
I see nothing in relation with a test on value "3" in the code before (so there is surely a bug if rule must be A before the PR). But i see nothing after the fix, so there is still the bug if the rule must be A after the PR. Note that if MAIN_DEPOSIT_MULTI_TVA is off, there is a test on "if (empty($lines[$i]->qty)) {" that is similar to making a test on special_code = 3. So rule A seems applied only when MAIN_DEPOSIT_MULTI_TVA is off.

**If rule must be B:**
The current code seems not consistent for MAIN_DEPOSIT_MULTI_TVA on and off, because the case of "option" (special code is 3 or qty is 0) is not taken into account.

**If rule must be C:**
The PR changes the rule in case of MAIN_DEPOSIT_MULTI_TVA is on, but not in case MAIN_DEPOSIT_MULTI_TVA is off. So this introduce a bug.

So what to do:

This bug is related to a hidden option (option MAIN_DEPOSIT_MULTI_TVA, so not an official and not yet stable option), this explains that few users should have the trouble and this also means we are free to decide what we want (we not have to maintain compatibility on not yet stable options).  So this is a good situation to clarify the use case.

So the first thing to do is to clarify what must be the business rule to decide if we must exclude or not a line to make the calculation of the VAT... document this into code comment and align the code in both MAIN_DEPOSIT_MULTI_TVA on and off with this rule.

How to decide A, B or C ? and to decide if C, we must understand what is goal if info_bit.

rycks:
@eldy ok so could we please clarify `info_bits` field too because for me that is a really dark part of dolibarr code ? 

sometimes that field is "Miscellaneous informations" or "Bits of type of lines" or "VAT npr or not ?" (in source code) ... and "TVA NPR or not" (in wiki) but this one seems good :

```
	public $info_bits = 0; // Some other info:
	// Bit 0: 	0 si TVA normal - 1 si TVA NPR
	// Bit 1:	0 ligne normale - 1 si ligne de remise fixe
```

Then, as sylvain fix concern info_bits (add a test on empty info bits) it seems the good place to clarify that no ?

eldy:
> @eldy ok so could we please clarify `info_bits` field too because for me that is a really dark part of dolibarr code ?
> 
> sometimes that field is "Miscellaneous informations" or "Bits of type of lines" or "VAT npr or not ?" (in source code) ... and "TVA NPR or not" (in wiki) but this one seems good :
> 
> ```
> 	public $info_bits = 0; // Some other info:
> 	// Bit 0: 	0 si TVA normal - 1 si TVA NPR
> 	// Bit 1:	0 ligne normale - 1 si ligne de remise fixe
> ```
> 
> Then, as sylvain fix concern info_bits (add a test on empty info bits) it seems the good place to clarify that no ?

Yes, this is necessarly to clarify info_bit to decide if rule must be B or C. For the moment my answer is just to say Yes or No to the PR. We have enough to say No, but not enought to say what is the good fix. I will take a look at this using my knowledge of the history of this field and try to document this a little bit in wiki...

eldy:
> > @eldy ok so could we please clarify `info_bits` field too because for me that is a really dark part of dolibarr code ?
> > sometimes that field is "Miscellaneous informations" or "Bits of type of lines" or "VAT npr or not ?" (in source code) ... and "TVA NPR or not" (in wiki) but this one seems good :
> > ```
> > 	public $info_bits = 0; // Some other info:
> > 	// Bit 0: 	0 si TVA normal - 1 si TVA NPR
> > 	// Bit 1:	0 ligne normale - 1 si ligne de remise fixe
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then, as sylvain fix concern info_bits (add a test on empty info bits) it seems the good place to clarify that no ?
> 
> Yes, this is necessarly to clarify info_bit to decide if rule must be B or C. For the moment my answer is just to say Yes or No to the PR. We have enough to say No, but not enought to say what is the good fix. 

For the moment i have just push this commit 43e9b00b3c9aae23d79e120471377662074170ca to have option B consistent in the unstable mode with the official stable mode. So B is no more bugged in unstable mode for this point. But we still have to decide if we must switch in mode C...

I will take a look at this my my knowledge of the history of this field and try to document this a little bit in wiki...

eldy:
> > @eldy ok so could we please clarify `info_bits` field too because for me that is a really dark part of dolibarr code ?
 > > sometimes that field is "Miscellaneous informations" or "Bits of type of lines" or "VAT npr or not ?" (in source code) ... and "TVA NPR or not" (in wiki) but this one seems good :
 > > ```
 > > 	public $info_bits = 0; // Some other info:
 > > 	// Bit 0: 	0 si TVA normal - 1 si TVA NPR
 > > 	// Bit 1:	0 ligne normale - 1 si ligne de remise fixe
 > > ```
 > > 
 > > 
 > >     
 > >       
 > >     
 > > 
 > >       
 > >     
 > > 
 > >     
 > >   
 > > Then, as sylvain fix concern info_bits (add a test on empty info bits) it seems the good place to clarify that no ?
 > 
 > 
 > Yes, this is necessarly to clarify info_bit to decide if rule must be B or C. For the moment my answer is just to say Yes or No to the PR. We have enough to say No, but not enought to say what is the good fix.
 
For the moment i have just push this commit [43e9b00b3c9aae23d79e120471377662074170ca](https://github.com/Dolibarr/dolibarr/commit/43e9b00b3c9aae23d79e120471377662074170ca) to have option B consistent in the unstable mode with the official stable mode. So B is no more bugged in unstable mode for this point. But i think thsi will not change anything for Sylvain trouble. So we still have to decide if we must switch in mode C...  @SylvainLegrand what value do you have for special_code and info_bits ?
 
I will take a look at this using my knowledge of the history of the field info_bit and try to document this a little bit in wiki...

PR #29621, opened 2024-05-06
Product/Stock Hook formObjectOptions can replace

ibuiv:
# NEW Hook formObjectOptions can replace in product/stock.php
This hook did not permit to replace the standard output. Just a check for the return value to give the possibility of replace the output text.

### DIFF
--- a/htdocs/product/stock/product.php
+++ b/htdocs/product/stock/product.php
@@ -812,7 +812,7 @@ if ($id > 0 || $ref) {
 					$result = $object->load_stats_sending(0, '2', 1, $filterShipmentStatus);
 					$helpondiff .= $langs->trans("ProductQtyInShipmentAlreadySent").': '.$object->stats_expedition['qty'];
 				}
-	
+
 				// Number of supplier order running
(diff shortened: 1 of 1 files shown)


htdocs/product/stock/product.php
```diff
@@ -725,178 +725,179 @@
 			$parameters = array();
 			$reshook = $hookmanager->executeHooks('formObjectOptions', $parameters, $object, $action); // Note that $action and $object may have been modified by hook
```
eldy, line 726: The hook here is designed to add lines, inside the first columns.
My understanding is that you want to completely removed the second column ?
The pb with your PR is that it generates not correctly balanced open/close object. In one case the table is closed into a given level and into another one, it is closed into another one. We should avoid such cases as it makes html templates reading difficult. If goal is to hide completely the right column, why don't you add a simple CSS entry into the custom CSS provided by your module ?

ibuiv: OK. 
I understand these remarks. 
I close the PR. Thanks for the time you spent to this issue.

PR #29623, opened 2024-05-06
NEW functionality for display codebar on pdf files 

lamrani002:
# NEW|New functionality for display qrcode if this option is checked on pdf config
qrcode


htdocs/core/modules/expedition/doc/pdf_espadon.modules.php
```diff
@@ -992,9 +992,47 @@ protected function _pagehead(&$pdf, $object, $showaddress, $outputlangs)
...
+					if ($num > 0) {
+						$obj = $db->fetch_object($resql);
+						$encoding = $obj->code;
+						$classname = "mod".ucfirst($obj->coder);
+					}
+
+					if (class_exists($classname)) {
+						$module = new $classname($db);
+						if ($module->encodingIsSupported($obj->code)) {
+							$result = $module->writeBarCode($object->ref, $encoding);
```
eldy, line 1015: To represent url, it is better to use a QR code, you can get it by replacing param $encoding into 'QRCODE'.

htdocs/core/modules/expedition/doc/pdf_espadon.modules.php
```diff
@@ -992,9 +992,47 @@ protected function _pagehead(&$pdf, $object, $showaddress, $outputlangs)
...
-			//$pdf->Image($logo,10, 5, 0, 24);
+			require_once DOL_DOCUMENT_ROOT.'/core/modules/barcode/doc/tcpdfbarcode.modules.php';
+			require_once DOL_DOCUMENT_ROOT.'/core/modules/barcode/doc/phpbarcode.modules.php';
+
+			$encoding ='';
+			if (getDolGlobalString('GENBARCODE_BARCODETYPE_THIRDPARTY')) {
+				$sql = 'SELECT rowid, code,coder FROM '.MAIN_DB_PREFIX.'c_barcode_type';
+				$sql .= ' WHERE rowid='.((int) (getDolGlobalString('GENBARCODE_BARCODETYPE_THIRDPARTY')));
+				$sql .= ' AND entity = 1';
+				$resql = $db->query($sql);
```
eldy, line 1003: Inside a class, you should use $this->db instead of $db.

htdocs/core/modules/expedition/doc/pdf_espadon.modules.php
lamrani002: ok, done

htdocs/core/modules/expedition/doc/pdf_espadon.modules.php
```diff
@@ -991,10 +991,27 @@ protected function _pagehead(&$pdf, $object, $showaddress, $outputlangs)
...
+				// get path of qrcode image
+				$newcode = $object->ref;
+				if (!preg_match('/^\w+$/', $newcode) || dol_strlen($newcode) > 32) {
+					$newcode = dol_hash($newcode, 'md5');
+				}
+				$barcode_path = $conf->barcode->dir_temp . '/barcode_' . $newcode . '_' . $encoding . '.png';
+			}
+
+			if ($result > 0) {
+				$pdf->Image($barcode_path, $this->marge_gauche,  $this->marge_haute +80 -5, 20, 20);
```
fappels, line 1011: Why not $pdf->write2DBarcode ?

eldy: > Why not $pdf->write2DBarcode ?

We want to get barcode using an intermediate image using >writeBarCode() because >writeBarCode() is an abstraction class that is common to different barcode generator (phpbarcode and tcpdfbarcode and more in future if we need).

fappels: Made the remark because seems more complicated to make first png using TCPDF and then use image iso of using TCPDF to write barcode in pdf. For object->ref not big issue, but if it is a big QR, QR quality coud be less by the image resize.

eldy: > Made the remark because seems more complicated to make first png using TCPDF and then use image iso of using TCPDF to write barcode in pdf. For object->ref not big issue, but if it is a big QR, QR quality coud be less by the image resize.

Not the same if you use $pdf->write2DBarcode() ? I suppose that tcpdf also generated an intermediate image or the barcode is directly saved as pdf instruction ?

fappels: $pdf->write2DBarcode() draws barcode pixel on pdf: https://github.com/Dolibarr/dolibarr/blob/dc8b79e0d4e3a9ac1eef56c1a5b8bb407d825cae/htdocs/includes/tecnickcom/tcpdf/tcpdf.php#L16071
No intermediate image involved.
$modTcpdfbarcode->writeBarCode() is more for showing barcode on screen.

eldy: You're right. We should enhance this by adding a method
`function pdf_barcode(&$pdf, ...)`
into file **pdf.lib.php**
with all content inside the 
`if (isModEnabled('barcode') && getDolGlobalString('BARCODE_RECEPTION_INVOICE')) {
...
}`

So we will be able to replace current code with a more opmitised version

PR #29631, opened 2024-05-07
FIX update price HT on expense report card

lvessiller-opendsi:
FIX update price HT on expense report card
- for expense report the unit price is in "TTC" base price type


lvessiller-opendsi:
@eldy 
See fix #29634 for "phan" errors but "adherent" or "membership" is not authorized in module name

PR #29683, opened 2024-05-14
fix document model activation

sonikf:


htdocs/modulebuilder/template/admin/setup.php
```diff
@@ -527,13 +527,13 @@
 										// Active
 										if (in_array($name, $def)) {
 											print '<td class="center">'."\n";
-											print '<a href="'.$_SERVER["PHP_SELF"].'?action=del&token='.newToken().'&value='.urlencode($name).'">';
+											print '<a href="'.$_SERVER["PHP_SELF"].'?action=del&amp;token='.newToken().'&amp;value='.urlencode($name).'&amp;type='.urlencode($type).'">';
```
eldy, line 530: Don't understand why this fix something. Other pages are using the &token , &value...
Can you descibre how to reproduce the bug

PR #29689, opened 2024-05-15
init array_options in current object   if we don't have record for it

atm-jpb:
# NEW|New 
We are in a situation where the current object has no values in its extra fields.
We want to initialize all the values to null so that the array_option is accessible in other contexts (especially in document generation).

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -6388,7 +6388,7 @@ abstract class CommonObject
 				} else {
 					/**
 					We are in a situation where the current object has no values in its extra fields.
-					We want to initialize all the values to empty so that the array_option is accessible in other contexts (especially in document generation).
+					We want to initialize all the values to null so that the array_option is accessible in other contexts (especially in document generation).
 					 **/
 					if (is_array($extrafields->attributes[$this->table_element]['label'])) {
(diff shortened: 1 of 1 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -6385,6 +6385,16 @@ public function fetch_optionals($rowid = null, $optionsArray = null)
...
 						}
 					}
+				}else{
+					/**
+					We are in a situation where the current object has no values in its extra fields.
+					We want to initialize all the values to empty so that the array_option is accessible in other contexts (especially in document generation).
+					 **/
+					if (is_array($extrafields->attributes[$this->table_element]['label'])) {
+						foreach ($extrafields->attributes[$this->table_element]['label'] as $key => $val) {
+							$this->array_options['options_' . $key] = '';
```
eldy, line 6395: $this->array_options['options_' . $key] to null may be semantically more true

atm-jpb: agree with that

eldy: No advice about that because i don't see what you mean. Without code exemple, i am just lost.
So may be you can choose what you think the best, and submit a PR the way you think it is best.
With code exemple, i can understand and tell if it's ok or not or if there is a better solution.

PR #29694, opened 2024-05-15
NEW Adds payment terms and procentual discount to the creation of an invoice from time entries in projects

UBuschler:
# NEW Adds payment terms and procentual discount to the creation of an invoice from time entries in projects

when an invoice is created from time entries in projects, a payment term is only transferred if the customer has a customer-specific payment term. If the customer does not have this, the invoice is created without a payment term.
This code inserts two additional lines in the pre-diagram for invoice creation. These lines show the payment terms and the percentage discount - pre-filled with the customer's values if available otherwise the first payment term and 0% discount.
It is possible to customize the values here. The invoice is then created with the specified payment terms and each invoice line with the specified percentage discount.
If an existing invoice is selected instead of a new one, the line for the payment term is hidden.

Greenshot 2024-05-15 16 37 30

### DIFF
--- a/htdocs/projet/tasks/time.php
+++ b/htdocs/projet/tasks/time.php
@@ -419,8 +419,7 @@ if ($action == 'confirm_generateinvoice') {
 		$tmpproduct = new Product($db);
 		$fuser = new User($db);
 		$remiseproject = GETPOST('remiseproject', 'int');
-        $condidproject = GETPOST('condidproject', 'int');
-		
+		$condidproject = GETPOST('condidproject', 'int');		
 		$db->begin();
 		$idprod = GETPOSTINT('productid');
 		$generateinvoicemode = GETPOST('generateinvoicemode', 'alphanohtml');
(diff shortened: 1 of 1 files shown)


htdocs/projet/tasks/time.php
```diff
@@ -418,7 +418,8 @@
 		$tmptimespent = new Task($db);
 		$tmpproduct = new Product($db);
 		$fuser = new User($db);
-
+		$remiseproject = GETPOST('remiseproject', 'int');
+		$condidproject = GETPOST('condidproject', 'int');
```
eldy, line 422: GETPOSTINT here

htdocs/projet/tasks/time.php
```diff
@@ -418,7 +418,8 @@
 		$tmptimespent = new Task($db);
 		$tmpproduct = new Product($db);
 		$fuser = new User($db);
-
+		$remiseproject = GETPOST('remiseproject', 'int');
```
eldy, line 421: The data is a string used by user to enter a numeric value;
So you must use here
$remiseproject = price2num(GETPOST('remiseproject', 'int'));

htdocs/projet/tasks/time.php
lumb2019: Done! Thank you

htdocs/projet/tasks/time.php
lumb2019: Done. Is it possible to combine both of your comments and change this to:
$remiseproject = price2num(GETPOSTINT('remiseproject'));?

Thanks for your answers. It is my first Request. :-)

eldy: GETPOSTINT will always return an int, so no need to use price2num.
But an amount or rate is not an int, it is a float with a possible decimalseparator. So you must use GETPOST and then convert it into a float according to the currency format of the user with price2num()

htdocs/projet/tasks/time.php
```diff
@@ -418,7 +418,8 @@
 		$tmptimespent = new Task($db);
 		$tmpproduct = new Product($db);
 		$fuser = new User($db);
-
+		$remiseproject = price2num(GETPOST('remiseproject', 'int'));
```
eldy, line 421: Must use GETPOST('remiseprojet', 'alphanohtml') (see my answer to your question)

htdocs/projet/tasks/time.php
lumb2019: OK. Done. Thanks a lot

PR #29703, opened 2024-05-16
Research for the public note field in propals

zahwyrix:
# NEW Research for the public note field in propals


htdocs/comm/propal/list.php
```diff
@@ -718,6 +719,10 @@
 if ($search_date_signature_end) {
 	$sql .= " AND p.date_signature <= '".$db->idate($search_date_signature_end)."'";
 }
+
+if ($search_note_public) {
+	$sql .= " AND p.note_public LIKE '%".$db->escape($search_note_public)."%'";
```
eldy, line 724: For a search with like you must do
`$sql .= " AND p.note_public LIKE '%".$db->escape($db->escapeforlike($search_note_public))."%'";`

PR #29707, opened 2024-05-17
DOL_DOCUMENT_ROOT now needed for analyseVarsForSqlAndScriptsInjection

fappels:


htdocs/main.inc.php
```diff
@@ -310,6 +310,14 @@ function analyseVarsForSqlAndScriptsInjection(&$var, $type, $stopcode = 1)
 	exit;
 }
 
+// This is to make Dolibarr working with Plesk
+if (!empty($_SERVER['DOCUMENT_ROOT']) && substr($_SERVER['DOCUMENT_ROOT'], -6) !== 'htdocs') {
+	set_include_path($_SERVER['DOCUMENT_ROOT'].'/htdocs');
+}
+
+// Include the conf.php and functions.lib.php and security.lib.php. This defined the constants like DOL_DOCUMENT_ROOT, DOL_DATA_ROOT, DOL_URL_ROOT...
+require_once 'filefunc.inc.php';
```
eldy, line 319: Security test on input variables should be executed before any manipulation of data.
So filefunc.inc.php should be after the analyseVarsForSqlAndScriptsInjection
What needs to move this before ?

fappels: I came in some situation with fatal error, think was website module. Can't reproduce. When I was in situation and debug I saw DOL_DOCUMENT_ROOT not defined here: https://github.com/Dolibarr/dolibarr/blob/0fe6380480ab9bc230225478420fed82d3a8351e/htdocs/main.inc.php#L81

Issue introduced here: https://github.com/dolibarr/dolibarr/commit/1b2bad3f428d82a4de17d940ad529392749ef711

PR #29711, opened 2024-05-17
NEW : Added menu entry tickets assigned to me

atm-irvine:
# NEW|New Added menu entry tickets assigned to me
I added a new menu entry for tickets assigned to me into the descriptor. It filters assign to you directly


eldy:
There is already a menu entry "My tickets"
image

If you need a default filter, you can use the feature "Default filters" in setup to force filter to be on current user.

PR #29728, opened 2024-05-21
FIX: unitialised var

BenjaminFlr:
` $numlines` is never initialised if `!empty(reshook)` for tabContentCreateOrder.

### DIFF
--- a/htdocs/commande/card.php
+++ b/htdocs/commande/card.php
@@ -2917,7 +2917,7 @@ if ($action == 'create' && $usercancreate) {
 			// Note that $action and $object may be modified by hook
 			$reshook = $hookmanager->executeHooks('addMoreActionsButtons', $parameters, $object, $action);
 			if (empty($reshook)) {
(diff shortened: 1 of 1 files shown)


htdocs/commande/card.php
```diff
@@ -2919,6 +2917,8 @@
 			// Note that $action and $object may be modified by hook
 			$reshook = $hookmanager->executeHooks('addMoreActionsButtons', $parameters, $object, $action);
 			if (empty($reshook)) {
+				$numlines = !empty($numlines) ? $numlines : count($object->lines);
```
eldy, line 2920: $numlines seems never defined before. So why not just 
$numlines = count($object->lines);

altairis-benjamin: I assume that it can be defined in a hook, am I wrong ?

altairis-benjamin: You're right. I made that choice before deleting the previous $numlines initialisation.

PR #29739, opened 2024-05-22
Fix: call inithooks ('receiptPrinter', 'globalcard') before using hooks in sendToPrinter

bafbes:
sendToPrinterBefore and sendToPrinterAfter may not be executed if inithooks is not called


htdocs/core/class/dolreceiptprinter.class.php
```diff
@@ -633,6 +635,7 @@ public function sendToPrinter($object, $templateid, $printerid)
 	{
 		global $mysoc, $langs, $user;
 		global $hookmanager;
+		$hookmanager->initHooks(array('receiptPrinter', 'globalcard'));
```
eldy, line 638: The context of a hook must be initialized into the calling main page.

bafbes: This is really a mistake as the calling page contains a context. So, I abandon the pull request. Thanks.

PR #29744, opened 2024-05-23
NEW mass facturation from expedition

YannisHoareau:
This features allow you to use a new mass action to create several factures at once from expedition. It's the same as orders and you also can create one by third party and validate it.


htdocs/expedition/list.php
```diff
@@ -604,7 +982,10 @@
 	'classifyclose' => img_picto('', 'stop-circle', 'class="pictofixedwidth"').$langs->trans("Close"),
 	'presend'  => img_picto('', 'email', 'class="pictofixedwidth"').$langs->trans("SendByMail"),
 );
-if (in_array($massaction, array('presend'))) {
+if ($user->rights->facture->creer) {
```
eldy, line 985: Must use now
if ($user->hasRight('facture', 'creer')) {

YannisHoareau: Fixed

PR #29747, opened 2024-05-23
FIX :bad management right on holiday approval

defrance:
FIX : the rights for approval of leave is not granted to superiors, it is a separate right it is also necessary to envisage the case of a person responsible for authorizations who is not the person responsible for the person

### DIFF
--- a/htdocs/holiday/card.php
+++ b/htdocs/holiday/card.php
@@ -1566,7 +1566,7 @@ if ((empty($id) && empty($ref)) || $action == 'create' || $action == 'add') {
 
 					if ($object->status == Holiday::STATUS_VALIDATED) {	// If validated
 						// Button Approve / Refuse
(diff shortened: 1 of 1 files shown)


htdocs/holiday/card.php
```diff
@@ -1565,7 +1566,7 @@
 
 					if ($object->status == Holiday::STATUS_VALIDATED) {	// If validated
 						// Button Approve / Refuse
-						if ($user->id == $object->fk_validator) {
+						if (($user->id == $object->fk_validator && $user->rights->holiday->approve) || ($user->rights->holiday->writeall && $user->rights->holiday->approve)) {
```
eldy, line 1569: Must now use 
$user->hasRight('holiday', 'approve')

atm-maxime:
@defrance don't forget to change the permission test in the controller part when you change it in the view part to display action button

PR #29750, opened 2024-05-24
FIX: objects lists from a thirdparty do not keep socid when using a mass action

jyhere:
When a list page has been reached from a thirdparty card, using a mass action lose the socid param.


eldy:
Can you provide an example (process to reproduce the bug) ?

jyhere:
> Can you provide an example (process to reproduce the bug) ?

Yup sorry, the bug is only present in V19, the thirdparty name is reported in the thirdparty filter in V20...:) 
(What if many thirdparties have the same name.. :D)

PR #29758, opened 2024-05-24
Update Ticket card.php Issue #29757

### LINKED ISSUE #29757: Ticket Ref Collisions
### Bug

In v18.0 the ticket form preemptively fills out a "Ref" in the create ticket form.   If more than one user opens a new ticket form at the same time they have the same "Ref".  When the forms are submitted, the later person's ticket inherits extra field values from the other person's ticket.
 
I have identified the problem in v18.0.  The "$object" variable is used to set and save posted values.  At the same time the "$object" variable is used to "fetch" data using the posted "ref" to check if the ref is already used.  The "$object->fetch" call populates the object with data values for the ticket found.   The extra field values from the ticket found overrides the extra field values submitted.

I would attempt to create a pull request to prevent this problem, but I see that the develop version contains major refactoring of the logic in this code-block.

https://github.com/Dolibarr/dolibarr/blob/18.0/htdocs/ticket/card.php#L206

============
FUTURE CONCERN
============

In the develop branch the problem appears to be exacerbated.  If I am reading the impending code-logic correctly, the future code will be changing the ticket "add" action to an "update".    This means that when two people are creating a ticket at the same time and somehow start with the same "ref" in their form the second person who submits their form will be updating the other person's ticket instead of creating a new ticket as expected.

https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/ticket/card.php#L211C1-L240C17

### Dolibarr Version

18.0, develop

### Environment PHP

PHP 7.x

### Environment Database

10.11.3-MariaDB-1 - Debian 12

1. Open a new ticket form in one browser tab.
2. Open a new ticket form in a different browser tab.
3. Submit form from step 1 setting a value for extra field X.
4. Submit form from step 2 WITHOUT setting a value for extra field X.

The event messages report that the Ref was already used and has been replaced.  However, the second ticket now has the same value for extra field X that was specified on the first ticket.

### Attached files

Screenshot_2024_05_24_at_11 16 52_AM

carmelchas:
Bug-Fix Issue #29757  Prevent the check for a duplicate ref from populating the new ticket object fields with values from the conflicting ticket.

# FIX|Fix #29757
In v18.0 the ticket form preemptively fills out a "Ref" in the create ticket form. If more than one user opens a new ticket form at the same time they have the same "Ref". When the forms are submitted, the later person's ticket inherits extra field values from the other person's ticket.

I have identified the problem in v18.0. The "$object" variable is used to set and save posted values. At the same time the "$object" variable is used to "fetch" data using the posted "ref" to check if the ref is already used. The "$object->fetch" call populates the object with data values for the ticket found. The extra field values from the ticket found overrides the extra field values submitted.

### DIFF
--- a/htdocs/ticket/card.php
+++ b/htdocs/ticket/card.php
@@ -212,7 +212,8 @@ if (empty($reshook)) {
 			$db->begin();
 
 			$getRef = GETPOST("ref", 'alphanohtml');
-			if ($object->fetch('', $getRef) > 0) {
+			$refcheck_object = new Ticket($db);
+			if ($refcheck_object->fetch('', $getRef) > 0) {
 				$object->ref = $object->getDefaultRef();
 				$object->track_id = null;
 				setEventMessage($langs->trans('TicketRefAlreadyUsed', $getRef, $object->ref));


htdocs/ticket/card.php
```diff
@@ -212,7 +212,8 @@
 			$db->begin();
 
 			$getRef = GETPOST("ref", 'alphanohtml');
-			if ($object->fetch('', $getRef) > 0) {
+			$refcheck_object = new Ticket($db);
+			if ($refcheck_object->fetch('', $getRef) > 0) {
```
eldy, line 216: This looks strange.
Doing a fetch from $object that is a Ticket object or from $refcheck_object that is also a Ticket should have the same result,n whatever is the status of $object before the fetch.

PR #29759, opened 2024-05-24
18 fix default autoselect mailto

rycks:
# Fix default contact is not used on mail action

That bug is far away ... but that is really a bug : for example on an invoice if you add somebody as "billing contact" then clic on "send invoice by mail" that contact is not pre-selected !

Same problem if there is no contact linked to that invoice, default contact on thirdpart for invoices are not pre-selected too ... 

More details on french forum : 
https://www.dolibarr.fr/forum/t/selection-automatique-du-contact-mail-probleme-recurrent/45657

### DIFF
--- a/htdocs/core/class/html.formmail.class.php
+++ b/htdocs/core/class/html.formmail.class.php
@@ -1048,7 +1048,7 @@ class FormMail extends Form
 	 */
 	public function getHtmlForTo()
 	{
-		global $langs, $form;
+		global $langs, $form, $object;
 		$out = '<tr><td class="fieldrequired">';
 		if ($this->withtofree) {
 			$out .= $form->textwithpicto($langs->trans("MailTo"), $langs->trans("YouCanUseCommaSeparatorForSeveralRecipients"));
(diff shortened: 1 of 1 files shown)


lvessiller-opendsi:
@rycks 
It's difficult to analyze with "pre-commit" errors.
Could you fix it and add a commit ?

rycks:
@lvessiller-opendsi done, i hope it's more clear now :)

lvessiller-opendsi:
> @lvessiller-opendsi done, i hope it's more clear now :)

It's an interesting PR but according to me it's not a FIX but a NEW.
You change the behaviour of the default values in "mailto" select for mail form.

I will request @eldy to know his feeling about it ?

rycks:
Hello Lionel,
I try to answer by mail ... That is a bug for user point of view : if I choose a contact as billing contact and that contact is not pre selected when I try to send an email ... That is a bug (functional bug)

Éric

Le 14 juin 2024 16:52:30 GMT+02:00, lvessiller-opendsi ***@***.***> a écrit :
>> @lvessiller-opendsi done, i hope it's more clear now :)
>
>It's an interesting PR but according to me it's not a FIX but a NEW.
>You change the behaviour of the default values in "mailto" select for mail form.
>
>I will request @eldy to know his feeling about it ?
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/Dolibarr/dolibarr/pull/29759#issuecomment-2168210880
>You are receiving this because you were mentioned.
>
>Message ID: ***@***.***>
-
Envoyé depuis un mobile
Expertise Réseau, Linux - https://cap-rel.fr

eldy:
I agree with @lvessiller-opendsi. The point of view of user is not always enough to say it is a bug to include in a maintenance release (in user point of view, nearly everyting is bug). If the behaviour of the application change, it is (often) an enhancement.
On old version, the billing contact is just an information, on new version it is an information, reused to save time and error when sending an invoice by email.
But, warning when changing behaviour: what can we say to users that will say us this later: 
"Hey, if i set the contact customer as billing contact for my information, now, each time i want to send the invoice by email to my lawyer or to my book keeper, the customer is automatically selected and i made mistake by sending critical message to the wrong person. This is a critical security bug introduced in the maintenance version..."
And in another user point of view, it's true. We can't guess all the way an app is used.

So in my point of view, it is a "feature adding preselection of data, so enhancement". And because enhancement must be avoid into a maintenance version (except if very critical), it should be added into develop.
Not that some fix may also not be elligible for maintenance, when they does not change behavior, for example when it needs a lot of change. 
I will think to make a decision map graph on wiki to help to qualifiy if it is a bug for maintenance, versus a bug or feature for develop, because the deicision is often "borderline". I must think of this for future to help to decide with clear rules...

For this ticket, i will suggest to author to push or on develop...

PR #29766, opened 2024-05-27
FIX #29734

### LINKED ISSUE #29734: Projects -> Time Spent does not save Note of the entry
### Bug

I have an existing project where I want to add "time spent" records.

Creating a new "Time spent" Entry does not save the Note/Description of the effort:

image

### Dolibarr Version

develop

### Environment PHP

8.2.18

### Environment Database

	MySQL or MariaDB 10.5.23-MariaDB-0+deb11u1

Navigate to an existing Project:
Projects -> List open Projects -> select Project -> Time Spent -> press +

### Attached files

_No response_

lamrani002:
# FIX|Fix #29734 by update function addTimeSpent for take a note


mwelters:
I tested manually and saving Notes in Tasks work now again !
Thank you !

htdocs/projet/class/task.class.php
```diff
@@ -1505,6 +1505,14 @@ public function addTimeSpent($user, $notrigger = 0)
 				$this->error = $this->db->lasterror();
 				$ret = -2;
 			}
+
+			// update note in table
+			$nsql = "UPDATE ".MAIN_DB_PREFIX."element_time SET note = '".$this->db->escape($timespent->note)."'";
```
eldy, line 1510: This looks strange because saving the $timespent->note should be already done on line 1501 with
$timespent->update()
What sql do yo have when the timespent->update is done ?

lamrani002: the problem is the commonobject update function used to initialize the other columns to null, which causes the value of the “note” field to be lost.

eldy: Line 1510 is inside $timespent->create().
It used createCommon()
But before calling ->create(), we have
$timespent->note = $this->timespent_note;
So the createCommon should create record with note = value defined into line
$timespent->note = $this->timespent_note;

But the update should also do update of the note from the value 
$timespent->note

So the problem is surely why the note is set to null. it should not because the property $timespent->note should be set with value to use for the update.

eldy:
A different fix was done.

PR #29773, opened 2024-05-28
NEW funclionality to Clone user with permissions and categories

lamrani002:
# NEW|New functionality to clone user


htdocs/user/class/user.class.php
```diff
@@ -4203,4 +4202,87 @@ public function findUserIdByEmail($email)
...
+		// Construction de la requête d'insertion
+		$sql = "INSERT INTO ".$this->db->prefix()."user_rights (entity, fk_user, fk_id)";
+		$sql .= " SELECT entity, ".$this->db->escape($toId).", fk_id";
+		$sql .= " FROM ".$this->db->prefix()."user_rights src";
+		$sql .= " WHERE fk_user = ".((int) $fromId);
+		$sql .= " AND NOT EXISTS (";
+		$sql .= "   SELECT 1";
+		$sql .= "   FROM ".$this->db->prefix()."user_rights dest";
+		$sql .= "   WHERE dest.entity = src.entity";
+		$sql .= "   AND dest.fk_user = ".$this->db->escape($toId);
```
eldy, line 4240: It is an int so must use
$sql .= " AND dest.fk_user = ".((int) $toId);

htdocs/user/class/user.class.php
```diff
@@ -4203,4 +4202,87 @@ public function findUserIdByEmail($email)
...
+		dol_syslog(get_class($this)."::clone_rights (delete default permissions)", LOG_DEBUG);
+
+		if (!$this->db->query($sqlDelete)) {
+			$this->db->rollback();
+			return -1;
+		}
+
+		// Construction de la requête d'insertion
+		$sql = "INSERT INTO ".$this->db->prefix()."user_rights (entity, fk_user, fk_id)";
+		$sql .= " SELECT entity, ".$this->db->escape($toId).", fk_id";
```
eldy, line 4233: It is an int so must use
$sql .= " SELECT entity, ".((int) $toId).", fk_id";

hregis:
@lamrani002 it will probably be necessary to adapt the cloning functions when Multicompany uses user centralization on the main entity!

eldy:
> @lamrani002 it will probably be necessary to adapt the cloning functions when Multicompany uses user centralization on the main entity!

An easy to do first step can be to show the clone feature only if user is in same entity than current entity. So cloning is kept always in same entity.

PR #29777, opened 2024-05-28
Fix #28481 API /mos/{id}/produceandconsume

### LINKED ISSUE #28481: BUG: REST API "/mos/{id}/produceandconsume" not working and have a confuse behavior

Humml87:
Fix #28481 API functionality.
The old code can't work with the current versions.
Now the code API-Call represented the function from the Webpage.

Best regards
Christian

### DIFF
--- a/htdocs/mrp/class/api_mos.class.php
+++ b/htdocs/mrp/class/api_mos.class.php
@@ -395,7 +395,6 @@ class Mos extends DolibarrApi
 		$arrayofarrayname = array("arraytoconsume","arraytoproduce");
 		foreach ($arrayofarrayname as $arrayname) {
 			foreach (${$arrayname} as $value) {
-
 				if (empty($value["objectid"])) {
 					$error++;
 					throw new RestException(500, "Field objectid required in " . $arrayname);
@@ -491,12 +490,10 @@ class Mos extends DolibarrApi
 		}
 
 		if (!$error) {
-
 			$consumptioncomplete = true;
 			$productioncomplete = true;
 
 			if ($autoclose > 0) {
-
 				// Refresh Lines after consumptions.
 				$this->mo->fetchLines();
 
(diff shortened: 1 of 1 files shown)


htdocs/mrp/class/api_mos.class.php
```diff
@@ -368,297 +381,177 @@ public function produceAndConsume($id, $request_data = null)
 		}
 
 		if (empty($labelmovement)) {
+			$error++;
```
eldy, line 384: This line should be useless because there is a throw just after, so we live the function

eldy:
i think i understand how the API was supposed to work.
I added some comment to explain it.

The way to use it is:
	 * - If arraytoconsume and arraytoproduce are both filled, this fill an empty MO with the lines to consume and produce and record the consumption and production.
	 * - If arraytoconsume and arraytoproduce are not provided, it consumes and produces all existing lines.

This means we can't use it to decide which line we consume or not. We can't use it to make partial consumption or production.
So, the conclusion is that old API can't work. So i fixed the bug and i renamed it into produceAndConsumeAll (to not lose code that may be usefull once fixed).

so i merged your PR manually to add it as the code for produceAndConsume()

Thanks to check...

PR #29789, opened 2024-05-30
FIX : Extrafield intshowzero in list

atm-jpb:
# FIX|Fix Extrafield intshowzero in list

When we have an extrafield of type int and we enter the value zero, this value does not appear in the list but it does appear on the card.


atm-arnaud:
@eldy The problem is when we want to write "0" in an extrafield INT. If we enter this value (before this PR), 0 will be save in bdd but not be in edit card or list. We will only show an empty field. It's so disapointed. This fix correct that

PR #29790, opened 2024-05-30
CHANGE: add a notice  in getRows() instead of returning false

jyhere:
Returning false if no limit clause seems a pretty extreme modification to me, developers should be aware by a notice instead of empty results if their module use this method in a bad way.


eldy:
Why returning a warning instead of error.
We don't want to have this method used a bad way. Never. So why allowing it if we want to be sure it is not badly used ?

jyhere:
> Why returning awarning instead of error. We don't want to have this method used a bad way. Never. So why allowing it if we want to be sure it is not badly used ?

Better ? :D

eldy:
> > Why returning awarning instead of error. We don't want to have this method used a bad way. Never. So why allowing it if we want to be sure it is not badly used ?
> 
> Better ? :D 

The "return" to be sure that code is not executed (DOS vulnerability) is still missing.

jyhere:
> > > Why returning awarning instead of error. We don't want to have this method used a bad way. Never. So why allowing it if we want to be sure it is not badly used ?
> > 
> > 
> > Better ? :D
> 
> The "return" to be sure that code is not executed (DOS vulnerability) is still missing.

Sorry I don't understand what you want, the return statement will never be reached after a fatal error.

image


i was thinking the trigger_error did just output an error log. But with E_USER_ERROR, it stops, so PR is ok.

PR #29791, opened 2024-05-30
FIX: default element_type should be 'commande' (e.g Commande::$element)

jyhere:


jyhere:
Sorry, I missed this PR...
Fix new expedition stats for already V20 migrated data](https://github.com/Dolibarr/dolibarr/pull/29432/files)

jyhere:
I already changed this, it was revert by this commit : 
https://github.com/Dolibarr/dolibarr/pull/29432/files
I don't understand why, there's no point to specify a default value which is not an actual element type of a Dolibarr object !
E.g : Commande::$element = 'commande'

PR #29804, opened 2024-05-31
Fix incorrect translation code and missing one

MaximilienR-easya:
# Fix
The translation code was incorrect and corresponded to nothing inside the language files and one of the translation was missing


frederic34:
missing translation need to be added in en_US

MaximilienR-easya:
@frederic34 added

PR #29806, opened 2024-05-31
fix lost socid in massactions from list

frederic34:


frederic34:
image

you will have shippings from different company with the same name

Issue #29241, opened 2024-04-05
Taking away space when pasting SIRET number into Client


hasse-basse:
### Feature Request

In the French version of 19.0.1,  Numéro de SIRET is used for clients. I either have it from my clients or I find it on Infogreffe or similar sites and it is almost always written in the format "123 123 123 12345", but when I copy and paste it into the field for the SIRET number in /htdocs/societe/card.php  the result will be "123 123 123 12" so I have to edit it manually.

I think it must be possible to add a function that takes away the space, or? It would be very helpfull.

JonBendtsen:
this field is probably used in other countries which may or may not require spaces.

This page writes the format with spaces and even letters.
https://www.startbusinessinfrance.com/blog/post/tip-what-is-a-siren-or-siret-number
`RCS PARIS 453 983 245`
`RC NANTES 234 987 456`

What if the field is just made longer? Wouldn't that solve your problems?

hasse-basse:
That is not the same, France is confusing. there are 3 different systems in place, in parallel. What I talk about is Numéro de SIRET, and it is only numbers, and it is 123 123 123 12345

- Id. prof. 1 (SIREN)	123 123 123 (this is the company ID without telling the address. 9 figures and no letters
- Id. prof. 2 (SIRET)     123 123 123 12345 (this one is based on the on above but the last 5 figures is unique for each address of the company)
- Id. prof. 3 (NAF-APE)	This one tell what kind of activity it is about Letters and figures
- Id. prof. 4 (RCS/RM)	This is the one you brought up as an example. It is on the way out, but is still around. It is the same as SIREN but also indicates where the registration has been done (Paris, Nantes ...)	
- Numéro de TVA : Vat number. The same as SIREN but FRxx in the beginning

The field for  the Numéro de SIREN does automatically take away the space when you paste into the filed, but the field for the SIRET does not. There is a function in PHP that can do this, it is just to add it, but I don't know how. The Numéro de SIRET is always 14 figures long and there are never letters and we should not change this, because then errors is more likely to happen.

JonBendtsen:
@hasse-basse  can you provide a picture of where it is shown inside dolibarr?

hasse-basse:
In

/htdocs/societe/card.php

when you create a company situated in France. So just create a new client and set its address to France.

JonBendtsen:
> /htdocs/fourn/commande/card.php

vendor order? how is that a client?

hasse-basse:
[...]

PR #29818, opened 2024-06-02
NEW Remove spaces when copying and pasting into idprof fields

zahwyrix:
# NEW
#29241


zahwyrix:
@eldy to remember this

eldy:
Some users may need spaces.
Also removing spaces may be expected on other fields 
So i suggest to address this like we addressed the "mandatory field", the "default value", the "default filter", the "focus field".
We can add a tab "copy/paste field cleaner" to allow to define a regex. We can then have js that clean char of field according to this regex.

zahwyrix:
This pull request is a response to the max length attribute, which is very frustrating when copying from a site that presents the SIREN with spaces, for example.

`<input type="text" class="maxwidth200" name="idprof1" id="idprof1" maxlength="9" value="">`

The user may indeed want spaces. Personally, I find it easier to read with spaces, but this is not currently possible with maxlength.

But in fact, your solution is more viable in the long term, plus we could adapt the maxlength with regexes, so I'll keep this pull request aside and try to look into it.

eldy:
> This pull request is a response to the max length attribute, which is very frustrating when copying from a site that presents the SIREN with spaces, for example.
> 
> `<input type="text" class="maxwidth200" name="idprof1" id="idprof1" maxlength="9" value="">`
> 
> The user may indeed want spaces. Personally, I find it easier to read with spaces, but this is not currently possible with maxlength.
> 
> But in fact, your solution is more viable in the long term, plus we could adapt the maxlength with regexes, so I'll keep this pull request aside and try to look into it.

I see.
So i can suggest an intermediate solution before a generic one is available.
When we paste, if the size of content that we try to paste is higher than the maxsize, then we remove spaces, otherwsie we keep value unchanged. We should get the best of the 2 worlds like this.

PR #29820, opened 2024-06-03
Fix multi input workflow for text type

Hystepik:
Fix workflow of multi input  for text type to be a simple javascript component


htdocs/core/class/commonobject.class.php
```diff
@@ -7329,15 +7329,6 @@ public function showInputField($val, $key, $value, $moreparam = '', $keysuffix =
 			// Special case that prevent to force $type to have multiple input
 			if (empty($this->fields[$key]['multiinput'])) {
```
eldy, line 7330: @Hystepik 
Do we still need this test on $this->fields[$key]['multiinput']   ?

PR #29821, opened 2024-06-03
replace fetch_projet() and fetch_project() by fetchProject()

frederic34:


frederic34:
@eldy still postponed ?

PR #29830, opened 2024-06-04
FIX: value mismatch when the entity of a FormSetupItem is not the current entity

atm-florianm:
## Rationale
A FormSetupItem can have an `entity` that is different from `$conf->entity` (usually, when it is different, its value is `0` to make it a cross-entity conf). Currently, saving the conf is correctly implemented (saves to the specified entity), but loading the conf isn't (it takes the value from the current entity).

I made this fix weeks ago, while working on a feature that, ultimately, didn't require FormSetup at all, so I probably haven't tested it, or not extensively if I did.

### DIFF
--- a/htdocs/core/class/html.formsetup.class.php
+++ b/htdocs/core/class/html.formsetup.class.php
@@ -663,9 +663,7 @@ class FormSetupItem
 	 */
 	public function loadValueFromConf()
 	{
-		global $conf;
-		if (isset($conf->global->{$this->confKey})) {
-			$this->fieldValue = $conf->global->{$this->confKey};
+		if ($this->fieldValue = dolibarr_get_const($this->db, $this->confKey, $this->entity)) {
 			return true;
 		} else {
 			$this->fieldValue = null;


htdocs/core/class/html.formsetup.class.php
```diff
@@ -663,9 +663,7 @@ public function __construct($confKey)
 	 */
 	public function loadValueFromConf()
 	{
-		global $conf;
-		if (isset($conf->global->{$this->confKey})) {
-			$this->fieldValue = $conf->global->{$this->confKey};
+		if ($this->fieldValue = dolibarr_get_const($this->db, $this->confKey, $this->entity)) {
```
eldy, line 666: dolibarr_get_const()  is doing SQL request in database. 
We should already have all values loaded in $conf (values with current entity and entity = 0).

Why entity of a formsetupitem is no current entity or 0 ?  I don't think this can happen. Edit setup when on entity x should be done in entity X (or 0).

atm-florianm: I can think of some weird cases in which we might edit the conf of an entity from another, but this is extremely client-specific and doesn't really justify  

> dolibarr_get_const() is doing SQL request in database. We should already have all values loaded in $conf (values with current entity and entity = 0).

Thanks ! I had not thought of this, but you are right. The correct value should already be loaded.

> Why entity of a formsetupitem is no current entity or 0 ? I don't think this can happen. Edit setup when on entity x should be done in entity X (or 0). Can you describe how to reproduce the bug ?

I can think of some client-specific cases in which we could edit the conf of an entity from another, but these fringe cases don't justify changing the behaviour of Dolibarr. I am closing the PR.

If needed, I could always re-open it if better arguments pop up.

PR #29831, opened 2024-06-04
width td description line contrat/card.php

atm-jonathan:
Width problem on the service <td>
image

By adding the width a 85%. 
The cells adapt correctly
image

### DIFF
--- a/htdocs/contrat/card.php
+++ b/htdocs/contrat/card.php
@@ -1736,7 +1736,7 @@ if ($action == 'create') {
 					// Line in mode update
 					// Ligne carac
 					print '<tr class="oddeven">';
-					print '<td>';
+					print '<td width="85%">';
 					if ($objp->fk_product > 0) {
 						$canchangeproduct = 1;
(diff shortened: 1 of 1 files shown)


htdocs/contrat/card.php
```diff
@@ -1612,7 +1612,7 @@
 						$productstatic->status_buy = $objp->tobuy;
 						$productstatic->status_batch = $objp->tobatch;
 
-						print '<td>';
+						print '<td width="85%">';
```
eldy, line 1615: There is not reason to have td with width forced to a small value to render better than td with automatic rendering.
This may works for on use case, but will we worse for others.

atm-jonathan: Ok, finally it's the line extrafields that changed the display
Class "titlefieldcreate" causes conflict
I've added a parameter to the showOptionals method, to have class 'left' instead of  'titlefieldcreate'

atm-jonathan:
see FIX_WIDTH_TD_CARD_CONTRAT_2

PR #29856, opened 2024-06-06
NEW convenience methods for getting / setting the value of an extrafield

atm-florianm:
# NEW convenience methods

The basic idea is just to simplify extrafield value retrieval:

```php
if ($shipment->array_options['options_customs_clearance'] ?? null == 'XYZ') {
    // do something
}
```

Could be written more concisely and more clearly as:
```php
if ($shipment->getExtraField('customs_clearance') === 'XYZ') {
    // do something
}
```

Beginners (and even sometimes seasoned community members) often forget the `options_` prefix to the array keys because it is not obvious and feels redundant as `options` is already included in the array's name. There might be other added benefits (to debugging for instance) once their use is widely adopted.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -7789,7 +7789,7 @@ abstract class CommonObject
 				$keyList = (empty($InfoFieldList[2]) ? 'rowid' : $InfoFieldList[2].' as rowid');
 
 				if (count($InfoFieldList) > 3 && !empty($InfoFieldList[3])) {
-					list ($parentName, $parentField) = explode('|', $InfoFieldList[3]);
+					list($parentName, $parentField) = explode('|', $InfoFieldList[3]);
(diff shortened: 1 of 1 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -7766,7 +7789,7 @@ function handlemultiinputdisabling(htmlname){
 				$keyList = (empty($InfoFieldList[2]) ? 'rowid' : $InfoFieldList[2].' as rowid');
 
 				if (count($InfoFieldList) > 3 && !empty($InfoFieldList[3])) {
-					list($parentName, $parentField) = explode('|', $InfoFieldList[3]);
+					list ($parentName, $parentField) = explode('|', $InfoFieldList[3]);
```
atm-florianm, line 7792:
```suggestion
					list($parentName, $parentField) = explode('|', $InfoFieldList[3]);
```

PR #29857, opened 2024-06-06
FIX show generated accountancy reports in user timezone

lvessiller-opendsi:
FIX show generated accountancy reports in user timezone
In all accountancy reports, the generated date is in server timezone.
And if user has a different timezone the generated date as few hours difference.
For example without this fix the user see "06/06/2024 13:10" instead of "06/06/2024 15:10" (now) in Generated date field.

image

### DIFF
--- a/htdocs/core/lib/report.lib.php
+++ b/htdocs/core/lib/report.lib.php
@@ -115,7 +115,7 @@ function report_header($reportname, $notused, $period, $periodlink, $description
 	print '<tr>';
 	print '<td>'.$langs->trans("GeneratedOn").'</td>';
 	print '<td>';
-	print dol_print_date($builddate, 'dayhour');
+	print dol_print_date($builddate, 'dayhour', 'tzuserrel');
 	print '</td>';
 	if ($variante) {
 		print '<td>'.($exportlink ? $langs->trans("Export").': '.$exportlink : '').'</td>';


eldy:
Using the server timezone was done on purpose. The reason is this one :
The report must depends on "the company" not on user. So if user ask accounting report, he must have the same result than the result of user 2.
We can have only 1 version on accountancy. For this reason, we use the same reference date for everybody.

lvessiller-opendsi:
> Using the server timezone was done on purpose. The reason is this one : The report must depends on "the company" not on user. So if user ask accounting report, he must have the same result than the result of user 2. We can have only 1 version on accountancy. For this reason, we use the same reference date for everybody.

Maybe, I don't understand but if the user refresh the page, it will be display "13:10" and in user timezone it's "15:10" so the user will say that the page wasn't refreshed (there is a trouble ...)

eldy:
Are you speaking about the date used into the filter to find data into the sql orthe value show on screen after the line "Généré le" ?

lvessiller-opendsi:
> Are you speaking about the date used into the filter to find data into the sql orthe value show on screen after the line "Généré le" ?
the value show on screen after the line "Généré le"

lvessiller-opendsi:
@eldy 
It's for the value show on screen "Généré le".
I didn't understand the user interface of theses accountancy reports.
For me, the word "generated date" is inappropriate because nothing is generated but only filters are applied and refreshed.
You're right theses reports are for company accounting and the report date should be in "company" timezone.
But how can we do it ?

rycks (approved):
ok for me ... because user can't trust on information displayed on screen if that information is not in his local timezone

rycks:
> ok for me ... because user can't trust on information displayed on screen if that information is not in his local timezone

to be more clear: if my local time is 15:30 and i click on a button then the software display "document généré le  ... 11:30" i will make at least one more click on the button to get the up to date document to 15:30 ...

lvessiller-opendsi:
It's seems not to be the good way to do.

PR #29859, opened 2024-06-06
fix/feat: ProductLot document folder have to be the rowid

FHenry:
Problem with not implement stritly with ID, when user create ref lot that can be look like id
Exemple 
 - lot ref : 4883 (rowid : 5300)
 - lot ref : 5300 (rowid : 5538)
 
On lot ref 5300, upload a file, the file is store in "document"/productlot/5538/file.test
=> Directory name is rowid of lot ref( "document"/productlot/5538/file.test), that what we want as lot ref is not unique (unique key on ref lot(column batch)/product) 

On lot ref 4883, upload a file, the file is store in "document"/productlot/5300/file.test1

Go back on lot ref 5300, you see the file /productlot/5300/file.test1 instead of 5538/file.test


htdocs/product/stock/productlot_card.php
```diff
@@ -596,7 +596,7 @@
 	// Documents
 	if ($includedocgeneration) {
 		$objref = dol_sanitizeFileName($object->ref);
-		$relativepath = $objref.'/'.$objref.'.pdf';
+		$relativepath = $object->id.'/'.$objref.'.pdf';
```
eldy, line 599: Using id may be a problem when doing export/import
Id may change. Restoring directory of a backup for exemple 
will be  not sync with restore 
 This is why we use ref.
Possible to use productref_batchnumber instead of batch number ? (I think it is the real key of a batch number)

Possible to use

PR #29866, opened 2024-06-07
remove warning phpstan

iouston:
Property ActionComm::$libelle is unused.
Is it deliberate to keep this property depreciated?

### DIFF
--- a/htdocs/comm/action/class/actioncomm.class.php
+++ b/htdocs/comm/action/class/actioncomm.class.php
@@ -121,12 +121,6 @@ class ActionComm extends CommonObject
 	public $label;
 
 	/**
-	 * @var string Agenda event label
-	 * @deprecated Use $label
-	 */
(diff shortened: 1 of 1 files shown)


frederic34:
you can see that deprecated properties are private, if you try to access this property, the value will be returned with DeprecatedHandler

https://github.com/Dolibarr/dolibarr/blob/60c8470d49a58cf9a1857e14cb8aad86069ae46c/htdocs/comm/action/class/actioncomm.class.php#L417-L422

it you try to access libelle, it will return label (old code, old modules, etc...)

frederic34:
you can add at the end of line ` // @phpstan-ignore-line`

iouston:
Thank you, I didn't know about this possibility

frederic34:
https://github.com/Dolibarr/dolibarr/pull/29873

Issue #29369, opened 2024-04-17
ERROR : selectForForms: Error bad setup of field objectdescorig=User, objectfield=webpassword_webpassword:fk_element


john-botella:
@eldy I have this error message on V19 of Dolibarr : 
`selectForForms: Error bad setup of field objectdescorig=User, objectfield=webpassword_webpassword:fk_element`
This error appear on a lot of our Modules and also on fresh module created with ModuleBuilder 

https://github.com/Dolibarr/dolibarr/blob/5278cb07a834a1cc628b8fe184ecd37a0f4d26fa/htdocs/core/class/commonobject.class.php#L7726-L7727

Backward to like V17 resolve my problème. 
```php 
	$objectfield = null;
	$out = $form->selectForForms($param_list[0], $keyprefix.$key.$keysuffix, $value, $showempty, '', '', $morecss, $moreparam, 0, (empty($val['disabled']) ? 0 : 1), '', $objectfield);
```

But why `$objectfield` was added ?
Before I create a fix I need to understand why `$objectfield` was populated.

eldy:
When we link to a field of an object, we must know a lot of information about the fied we link to. The class and path TARGET was provided into objectdesc.

But sometimes we need more. it was not enough. We have all information, already defined into the SOURCE object into the table $fields OR into $extrafields definition. So with objectfields, we can just provide the name of the field in source object or extrafield of source object, and system will retreive the information of the target with it.

This second method is used only if param objectfield is provided. When provided, we can retreive the $objectdesc. If not, we backfall on old code for compatibility.
If objectfield is provided, objectdesc is not used.
First method is deprecated because less complete (was using a hardcoded solution when we can retreived info dynamically). It was also not compatible with extrafields.

If you have objectfield=webpassword_webpassword:fk_element then you must have into the
Webpassword->$fields an entry 'fk_element' with 'type'='integer:User...'

PS: i do not experience troubles with new generated object with modulebuilder.

john-botella:
Ok thanks, I'll look into it.

soehlrich:
After upgrading to V19.01 a lot of the complementary fields that use links to objects have stopped working. I have a Delivery Address field in Purchase orders that is a link to a warehouse (entrepot) object. 

I have tried to see what the issue is, but am not familiar enough with the system. I did manually put your (Eldy ) correction into the functions.lib.php code 
	_} elseif ($element_type == 'stock' || $element_type == 'entrepot') {
		$module='stock';
[...]

PR #29870, opened 2024-06-07
FIX : Missing hook in getElementProperties

atm-irvine:
#FIX|Fix Missing hook in getElementProperties
I added the missing hook in getElementProperties to handle $element_properties array because of some errors for permissions

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -12166,7 +12166,7 @@ function getElementProperties($element_type)
 		'dir_output' => $dir_output
 	);
 
-	$reshook = $hookmanager->executeHooks('changeElementProperties', $parameters);
+	$reshook = $hookmanager->executeHooks('getElementProperties', $parameters);
 	if ($reshook < 0) {
 		setEventMessages($hookmanager->error, $hookmanager->errors, 'errors');
 	} elseif ($reshook > 0) {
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -12154,7 +12166,15 @@ function getElementProperties($element_type)
 		'dir_output' => $dir_output
 	);
 
-	//var_dump($element_properties);
+	$reshook = $hookmanager->executeHooks('changeElementProperties', $parameters);
```
eldy, line 12169: Can youbrename trigger function in getElementProperties to match same name than dolibarr function.

eldy (changes requested):

john-botella:
@eldy this PR allow fix some troubles for modules discribed in [#29369 ](https://github.com/Dolibarr/dolibarr/issues/29369)

htdocs/core/lib/functions.lib.php
```diff
@@ -12154,7 +12166,15 @@ function getElementProperties($element_type)
 		'dir_output' => $dir_output
 	);
 
-	//var_dump($element_properties);
+	$reshook = $hookmanager->executeHooks('getElementProperties', $parameters);
+	if ($reshook < 0) {
+		setEventMessages($hookmanager->error, $hookmanager->errors, 'errors');
```
eldy, line 12171: The method is getElementProperties is low level method that can be run out of user/web context (in cron job for exemple).
So we should not have setEventMessage inside.

eldy (changes requested):

john-botella:
close this PR and use #30086 instead

PR #29872, opened 2024-06-07
Fix check if object is null

lamrani002:
# FIX|Fix phan warning

### DIFF
--- a/htdocs/core/class/commondocgenerator.class.php
+++ b/htdocs/core/class/commondocgenerator.class.php
@@ -1368,7 +1368,7 @@ abstract class CommonDocGenerator
 
 
 		// Load extra fields if they haven't been loaded already.
-		if (empty($this->extrafieldsCache)) {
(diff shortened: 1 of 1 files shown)


eldy:
What was the warning code ?

lamrani002:
Property CommonDocGenerator::$extrafieldsCache (ExtraFields) in empty() is not falsy.

htdocs/core/class/commondocgenerator.class.php
```diff
@@ -1368,7 +1368,7 @@ public function getExtrafieldContent($object, $extrafieldKey, $outputlangs = nul
 
 
 		// Load extra fields if they haven't been loaded already.
-		if (empty($this->extrafieldsCache)) {
```
eldy, line 1371: Try with if (!is_object...

eldy (changes requested):

eldy:
Pr has meege conflict, surely because of fix already meeged differently. So i close pr.

Issue #29402, opened 2024-04-19
Extend Pagination Data Feature to All Listing APIs (e.g., /thirdparties, /invoices...)


Kazimir42:
### Feature Request

Hello,

While using the REST API for Dolibarr, we noticed that the /product endpoint includes a nullable pagination_data object, which returns pagination information (total, page, page_count, and limit) when set to true. This feature appears to be exclusive to this endpoint. Why not implement it across all listing endpoints?

The intial issue from /products : https://github.com/Dolibarr/dolibarr/issues/17484

### Suggested implementation

The same as index() from htdocs/product/class/api_products.class.php

JonBendtsen:
@Kazimir42  I suppose it is this one you are talking about?

![product_pagination](https://github.com/Dolibarr/dolibarr/assets/2551481/d668ab40-15ad-4db8-a732-024d47c284a7)

JonBendtsen:
@Kazimir42  it seems like a good option, and a rather trivial change, so I suggest that you try implementing it for proposals - because there is likely to be many of those. 

Look at this Pull request for inspiration, but copy from the latest version of the file in this pull request, make it work for proposals, then porting to orders and invoices should be really trivial.

https://github.com/Dolibarr/dolibarr/pull/17485/files

PR #29895, opened 2024-06-07
NEW : Add pagination data to some api routes

Kazimir42:
Hello,

Based on [my issue](https://github.com/Dolibarr/dolibarr/issues/29402),

When using the product API, I noticed there is pagination data that can be useful for other endpoints. Therefore, I added it by simply copying and pasting the code already implemented in the product API [here](https://github.com/Dolibarr/dolibarr/pull/17485).

Would you prefer if we implemented a new function in the DolibarrApi class instead?

### DIFF
--- a/htdocs/comm/propal/class/api_proposals.class.php
+++ b/htdocs/comm/propal/class/api_proposals.class.php
@@ -159,7 +159,7 @@ class Proposals extends DolibarrApi
 	 * @param string    $sqlfilters         Other criteria to filter answers separated by a comma. Syntax example "(t.ref:like:'SO-%') and (t.datec:<:'2016-01-01')"
 	 * @param string    $properties	        Restrict the data returned to these properties. Ignored if empty. Comma separated list of properties names
 	 * @param bool      $pagination_data    If this parameter is set to true the response will include pagination data. Default value is false. Page starts from 0*
-     * @return  array                       Array of order objects
+	 * @return  array                       Array of order objects
 	 */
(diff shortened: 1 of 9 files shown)


htdocs/comm/propal/class/api_proposals.class.php
```diff
@@ -157,10 +157,11 @@ private function _fetch($id, $ref = '', $ref_ext = '', $contact_list = 1)
 	 * @param int		$page				Page number
 	 * @param string	$thirdparty_ids		Thirdparty ids to filter commercial proposals (example '1' or '1,2,3') {@pattern /^[0-9,]*$/i}
 	 * @param string    $sqlfilters         Other criteria to filter answers separated by a comma. Syntax example "(t.ref:like:'SO-%') and (t.datec:<:'2016-01-01')"
-	 * @param string    $properties	Restrict the data returned to these properties. Ignored if empty. Comma separated list of properties names
+	 * @param string    $properties	        Restrict the data returned to these properties. Ignored if empty. Comma separated list of properties names
+	 * @param bool      $pagination_data    If this parameter is set to true the response will include pagination data. Default value is false. Page starts from 0*
 	 * @return  array                       Array of order objects
 	 */
-	public function index($sortfield = "t.rowid", $sortorder = 'ASC', $limit = 100, $page = 0, $thirdparty_ids = '', $sqlfilters = '', $properties = '')
+	public function index($sortfield = "t.rowid", $sortorder = 'ASC', $limit = 100, $page = 0, $thirdparty_ids = '', $sqlfilters = '', $properties = '', $pagination_data = false)
```
eldy, line 164: There is already pagination available with parameter "$limit" and "$page", so what is goal of this one ?

PR #29905, opened 2024-06-07
fix: Mass stock transfert display error message when something goes wrong 

FHenry:
just UI Test

### DIFF
--- a/htdocs/product/stock/class/mouvementstock.class.php
+++ b/htdocs/product/stock/class/mouvementstock.class.php
@@ -312,6 +312,7 @@ class MouvementStock extends CommonObject
 				return -2;
 			}
 
+			// Check unicity for serial numbered equipment once all movement were done.
 			if ($product->status_batch==2 && abs($qty)>1) {
 				$this->errors[] = $langs->transnoentities('TooManyQtyForSerialNumber', $product->ref, $batch);
 				dol_syslog($langs->transnoentities('TooManyQtyForSerialNumber', $product->ref, $batch), LOG_ERR);
@@ -645,7 +646,6 @@ class MouvementStock extends CommonObject
 				$error++;
 			}
 			// End call triggers
-
 			// Check unicity for serial numbered equipment once all movement were done.
 			if (!$error && isModEnabled('productbatch') && $product->hasbatch() && !$skip_batch) {
 				if ($product->status_batch == 2 && $qty > 0) {	// We check only if we increased qty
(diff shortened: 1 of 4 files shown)


htdocs/product/stock/class/mouvementstock.class.php
```diff
@@ -306,12 +306,21 @@ public function _create($user, $fk_product, $entrepot_id, $qty, $type, $price =
...
 				$this->errors[] = $langs->transnoentitiesnoconv("ErrorTryToMakeMoveOnProductRequiringBatchData", $product->ref);
-				dol_syslog("Try to make a movement of a product with status_batch on without any batch data");
+				dol_syslog("Try to make a movement of a product with status_batch on without any batch data", LOG_ERR);
 
 				$this->db->rollback();
 				return -2;
 			}
 
+			// Check unicity for serial numbered equipment once all movement were done.
+			if ($product->status_batch==2 && abs($qty)>1) {
```
eldy, line 316: Even if status_batch is 2, we must be allowed to make a move on qty different than 1. For exemple you have a product with lot and qty of 5. You modify the product to be a product with unique serial. Qty remains 5 so dolibarr show a warning to ask to fix qty. You must be able to register a move stock of -4.
The check for uniciry must be done at znd of movement (after movement has been recorded) and not inthe delta we want to move. And we should arleady have this test, so this one seems useless (a simple comment to explain that may be enough)

eldy (changes requested):

htdocs/product/stock/class/mouvementstock.class.php
FHenry: That's another problem, You MUST not be able to change to "serial mode" if you have more than 1 in stock.
Actually it is a "Warning" in the UX, you've already rejected a PR that prevail this.

In your use case : First things to to is to split you'r lot into serial   and then change your batch type from lot to serial.

ping @lmag

eldy: We should be able. Not being able is a "dead lock" feature.
What is the problem ? Currently code manage correctly this. If you try to make a change in stock to get qty > 1 and product must be serial unique, you have a error message. So goal seems achieved ?

FHenry: Ok I unstand your point.I remove this change in mvtstock Class from this PR, because it's another point. Let this PR just deal with Mass Tranfert feature

PR #29907, opened 2024-06-07
Fix warnings technical debt

Daoud-mohamed:
Call to function is_array() with (int|string) will always evaluate to false.

### DIFF
--- a/htdocs/comm/action/class/actioncomm.class.php
+++ b/htdocs/comm/action/class/actioncomm.class.php
@@ -1274,8 +1274,10 @@ class ActionComm extends CommonObject
 
 				if (!empty($this->socpeopleassigned)) {
 					$already_inserted = array();
-					foreach (array_keys($this->socpeopleassigned) as $val) {
-						$val = array('id' => $val);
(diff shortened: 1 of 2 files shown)


htdocs/comm/action/class/actioncomm.class.php
```diff
@@ -1274,10 +1274,8 @@ public function update(User $user, $notrigger = 0)
 
 				if (!empty($this->socpeopleassigned)) {
 					$already_inserted = array();
-					foreach (array_keys($this->socpeopleassigned) as $key => $val) {
-						if (!is_array($val)) {	// For backward compatibility when val=id
-							$val = array('id' => $val);
-						}
+					foreach (array_keys($this->socpeopleassigned) as $val) {
+						$val = array('id' => $val);
```
eldy, line 1278: $val can be an id xxx (old versions) but can also be array('id' => xxx)
If it is already an array, the PR will transform it into array('id'=>array('id'=>...

eldy (changes requested):

PR #29910, opened 2024-06-07
fix: #29909

### LINKED ISSUE #29909: Problème de conservation des filtres dans la liste des propal depuis un tiers
### Bug

J’ai besoin d’accéder à mes PR ouvertes sur un tiers mais je me retrouve avec une PR qui n’a rien à voir dans ma recherche. 
Voici mon parcours : 

Image

Image

Image

Image

### Dolibarr Version

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

_No response_

### Attached files

_No response_

FHenry:
fix: #29909

### DIFF
--- a/htdocs/comm/propal/card.php
+++ b/htdocs/comm/propal/card.php
@@ -2849,8 +2849,8 @@ if ($action == 'create') {
 		}
 
 		$tmparray = $object->getTotalWeightVolume();
-		$totalWeight = $tmparray['weight'];
-		$totalVolume = $tmparray['volume'];
+		$totalWeight = isset($tmparray['weight']) ?:0;
+		$totalVolume = isset($tmparray['volume']) ?:0;
 		if ($totalWeight) {
 			print '<tr><td>'.$langs->trans("CalculatedWeight").'</td>';
 			print '<td class="valuefield">';
--- a/htdocs/comm/propal/card.php
+++ b/htdocs/comm/propal/card.php
@@ -2849,8 +2849,8 @@ if ($action == 'create') {
 		}
 
 		$tmparray = $object->getTotalWeightVolume();
-		$totalWeight = isset($tmparray['weight']) ?:0;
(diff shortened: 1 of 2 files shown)


htdocs/comm/propal/list.php
```diff
@@ -1230,7 +1232,7 @@
 }
 if (!empty($arrayfields['s.nom']['checked'])) {
 	print '<td class="liste_titre" align="left">';
-	print '<input class="flat maxwidth100" type="text" name="search_societe" value="'.dol_escape_htmltag($search_societe).'">';
+	print '<input class="flat maxwidth100" type="text" name="search_societe" value="'.dol_escape_htmltag($search_societe).'"'.($socid > 0 ? " disabled" : "").'>';
```
eldy, line 1235: If field is disabled, the new search/submit of form will lost the filter (field will not be propagated).

htdocs/comm/propal/card.php
```diff
@@ -2849,8 +2849,8 @@
 		}
 
 		$tmparray = $object->getTotalWeightVolume();
-		$totalWeight = $tmparray['weight'];
-		$totalVolume = $tmparray['volume'];
+		$totalWeight = isset($tmparray['weight']) ?:0;
```
eldy, line 2852: Using isset with ?: is strange. It means it will return true if isset true instead of $weight

eldy (changes requested):

htdocs/comm/propal/list.php
FHenry: Yes exactly ,as it is done in facture/list.php.

When "filter" is send in QueryString with socid attribute, it is priorized with socid (the search_societe is displayed for the UX but not used anymore)

htdocs/comm/propal/card.php
FHenry: Yes, error fixed

PR #29913, opened 2024-06-08
FIX #29901 supporting non-invoiceable task time in invoice generation process

### LINKED ISSUE #29901: User add non-billable time to invoices without noticing when generating invoices from project's time spent list
### Bug

Generating an Invoice from a Project's Time Spent List includes time that are not billable

The Project's TsksList show all the Times spent on the Project. The user can select Time that is not billed (attached to an Invoice) to generate an Invoice.

The List does not show if the Task Time is billable or not, thereforre, non-billable time is shown to the user, without feedback of wether the time is billable or not.

User add non-billable time to invoices without noticing when generating invoices from project's time.

### Dolibarr Version

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

1. Produce some Time associated to Project
2. GoTo Project's Tasks List
3. Filter by 'Facturable'
4. Select All
5. In the Actions Drop-down, ... Generate new Invoice
6. Select Anything, ... Generate.
7. GoTo generated Invoice
8. Non-billable time is added to the Invoice

### Attached files

_No response_

vdegrandpre:
REPLACED BY #29964 

Fixes an issue where the user add non-invoiceable time to an invoice by generating it from the time spent on project tasks list.
At the same time, add a safeguard to disable the select checkbox in two conditions : task is not invoiceable ou task time already invoiced. The checkbox can be re-enabled by double-clicking very close outside the desired checkbox to enable iut again.


htdocs/projet/tasks/time.php
```diff
@@ -1566,13 +1566,22 @@ function setDetailVisibility() {
...
 		$sql = "SELECT t.rowid, t.fk_element, t.element_date, t.element_datehour, t.element_date_withhour, t.element_duration, t.fk_user, t.note, t.thm,";
 		$sql .= " t.fk_product,";
 		$sql .= " pt.ref, pt.label, pt.fk_projet,";
 		$sql .= " u.lastname, u.firstname, u.login, u.photo, u.gender, u.statut as user_status,";
 		$sql .= " il.fk_facture as invoice_id, inv.fk_statut,";
 		$sql .= " p.fk_soc,s.name_alias,";
-		$sql .= " t.invoice_line_id";
+		$sql .= " t.invoice_line_id,";
+		if ($invoiceable) {
+			$sql .= " efpt.invoiceable as options_invoiceable";
```
eldy, line 1583: This field does not exist by default in dolibarr so should not be used in default code.

eldy:
Right. Adding the field "billable" into the table project_task is a better place...

PR #29915, opened 2024-06-08
Regression : no more categ's path in categories multi select 

vmaury:
Since at least v18, the path of cats (like catParent >> catChild) are no more displayed in categories select (for example used in list files)
In v16
select in v16

In V18
select in v18

With this fix (in v19)
select in v19 fixed


JoAnMaFe:
This functionality was great, it was lost from v17 I think. It was very interesting in products and projects.

PR #29934, opened 2024-06-08
Allow tweaking of the index that MySQL use when searching products

creekorful:
Sometimes when working with large dataset with many entities the MySQL engine does not pick-up the right index. This is especially true when doing a search over the reference because the LIKE '%%' prevent the uk_product_label from being chosen.

This PR aims to improve this situation by doing the following :

- Adding a simple index on entity column.
- Adding an hidden constant to force the index being used when searching products.

Performances over a 8M products dataset shows a reduction of 2/3 execution time.

PR based on the work done by @lemathou


htdocs/core/class/html.form.class.php
```diff
@@ -2874,6 +2874,7 @@ public function select_produits_list($selected = 0, $htmlname = 'productid', $fi
 		// phpcs:enable
 		global $langs;
 		global $hookmanager;
+		global $conf;
```
eldy, line 2877: No more need. See next comment

htdocs/core/class/html.form.class.php
```diff
@@ -2961,6 +2962,11 @@ public function select_produits_list($selected = 0, $htmlname = 'productid', $fi
 		}
 
 		$sql .= " FROM ".$this->db->prefix()."product as p";
+
+		if (getDolGlobalString($conf->global->MAIN_SEARCH_PRODUCT_FORCE_INDEX)) {
```
eldy, line 2966: getDolGlobalStrin('MAIN_...')

eldy (changes requested):

htdocs/core/class/html.form.class.php
creekorful: Thank you. I forgot to read the manual. :(

eldy (changes requested):

PR #29936, opened 2024-06-08
WIP Add option to search in sub-categories of the selection in lists

vmaury:
And begin to refactor the code, no more code duplicated for sql search for categories in list
Create a new method getSqlSearch in categorie.class.php called in lists, instead of forging the where by hand each time


vmaury:
This PR is not really ready to merge : 
I made changes in category.class.php to manage the new filters and improve the existings
I report changes only in product/list.php, societe/list.php (customer and suppliers) and project/list.php  : 
- subcategory search is available, but checkbox now only appears if categories deep is >=1 (at least 1 categ has a child)
- all the sql generated to filter by categ is now generic and moved in category.class.php (method ->getSqlSearch(..))

If you think it's ok, I'll report the same changes in ticket, knowledgemnagement, and if time enough in every other lists (member contact, user, bank_*, actioncomm,website_page,warehouse)

vmaury:
@eldy and @atm-maxime if you can have a look on this, and say if I'm in the right direction, I'll continue or change ...
Cheers !

eldy:
i don't think you are on good direction.
To make a search to know if a category C is in path of at least one category of an object, you must
- get the full path of categories with get_full_arbo
- to take all ID of categories that are child of C by a loop on the array return by get_full_arbo (so memory parsing only)
- generate the filter sql with EXIST category in the list of id generated at previous step.
- not mixing refactoring (renaming in same pr).

i close the PR as it should be wrote differently.

PR #29940, opened 2024-06-09
Fix ajax requests on commonobjectline

lemathou:
# FIX
All objects extended from CommonObjectLine, as OrderLine, ExpeditionLigne, FactureLigne, PropaleLigne, etc. cannot be requested using ajax selectobject (lack of attributes in object classes, etc.).


lemathou:
bad base branch...

PR #29951, opened 2024-06-09
Support Docker install in production environment

yledoare:
- Enable dolibarr_main_url_root writable in forced mode (needed for reverse proxy)
- Add install.forced.docker.php
- Move old docker files in docker-dev folder
- Fix else if CodeSniffer warning


htdocs/install/fileconf.php
```diff
@@ -221,7 +221,8 @@ class="minwidth300"
 				   name="main_url"
 				   value="<?php print $dolibarr_main_url_root; ?> "
 <?php if (!empty($force_install_noedit)) {
-	print ' disabled';
+	/* Behind a reverse proxy with https, we have to keep it writable */
```
eldy, line 224: I suggest to replace test 
if !empty(force_install_no_edit)
With 
If !empty(force_install_no_edit) && force_install_no_edit != 3

So you can use value 3 in your config file...

eldy (changes requested):

yledoare (commented):
revert on main_url editable

PR #29966, opened 2024-06-10
FIX width td desc line contrat/card.php view and edit

atm-jonathan:
Width problem on the service <td>
image

By adding the width a 85%. 
The cells adapt correctly
image

### DIFF
--- a/htdocs/contrat/card.php
+++ b/htdocs/contrat/card.php
@@ -1731,7 +1731,7 @@ if ($action == 'create') {
 						$line = new ContratLigne($db);
 						$line->id = $objp->rowid;
 						$line->fetch_optionals();
-						print $line->showOptionals($extrafields, 'view', array('class' => 'oddeven', 'style' => $moreparam, 'colspan' => $colspan), '', '', 1);
+						print $line->showOptionals($extrafields, 'view', array('class'=>'oddeven', 'style'=>$moreparam, 'colspan'=>$colspan, 'tdclass' => 'left'), '', '', 1);
 					}
 				} else {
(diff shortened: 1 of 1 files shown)


eldy:
Space should be computed correctly automatically by the browser.
Can you paste the example you used into service description to reproduce the bug ? You can do it on demo.dolibarr.org

atm-jonathan:
The display bug occurs when there are one or more contract line extrafields
I can't add extrafield on  "demo.dolibarr.org"

atm-jonathan:
@eldy 
Can I have a varchar extrafield on a contract line?
On demo.dolibarr.org
So that I can demonstrate the display problem
thank you

eldy:
> @eldy 
> Can I have a varchar extrafield on a contract line?
> On demo.dolibarr.org
> So that I can demonstrate the display problem
> thank you

I added field Extracontline...

atm-jonathan:
image
@eldy 
https://demo.dolibarr.org/contrat/card.php?id=563#403

Thanks for adding 

As you can see from the image above, the width of the td changes when an extrafield is used.

eldy:
The trouble seems to come from the balance of fields. As we can see in html source, on a tr we have 6 columns (colspan=6) and the next one we have a td + a td of colspan=6 so 7 columns.
image

atm-jonathan:
Hello !
No solutions found by modifying the colspan
Can we modify showOptionals() to make title and content <span> in the td that takes colspan=6?

to get this?
Capture d’écran du 2024-07-01 15-44-56

eldy:

PR #29990, opened 2024-06-11
NEW : option to allow freezing qty in BOM service's line

vmaury:
Some manufacturing needs to freeze quantities for services to : in printing industries, for example, the machine needs time to be tuned before starting the manufacturing, and this time is not dependant of the number of pieces printed.

Freezing quantity is for the moment only possible for products, but not for services

This PR allow to unlock this

In BOM config
Screenshot 2024-06-11 at 16-24-20 Configuration du module BOM multicompany

In a BOM card
Screenshot 2024-06-11 at 16-25-44 Nomenclature multicompany

### DIFF
--- a/htdocs/admin/bom.php
+++ b/htdocs/admin/bom.php
@@ -445,21 +445,6 @@ print '<td class="center" width="60">'.$langs->trans("Value").'</td>';
 print "<td>&nbsp;</td>\n";
 print "</tr>\n";
 
-// possibility to freeze qty in services to
-print '<tr class="oddeven">';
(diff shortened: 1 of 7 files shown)


vmaury:
And this frozen services are ok when production is started/done

(in the above examble, FRAIS_POSTE service is frozen)
Screenshot 2024-06-11 at 16-45-39 Mo

eldy:
There is no need to add an option to allow to use frozen qty on services.
This can be chosen for product when making the BOM, so it should be choosable for services too when doing the BOM.
So having the feature always on is ok.

vmaury:
Do and undo is always working (I'm not sure of the translation of "faire et défaire c'est toujours travailler") ;-)
I added this to avoid side effects (for example in gpaoplus module...), but I agree with you that this option is not necessary

PR #29996, opened 2024-06-12
fetchObjectByElement could return int error code

rycks:
There is a lot of case in dolibarr code where there is no tests on values returned by fetchObjectByElement function.

That function COULD return an object (most of the time when there is no problem) but sometimes it's 0 or -1 (for example) then we don't have to use properties on it !

### DIFF
--- a/htdocs/core/ajax/ajaxtooltip.php
+++ b/htdocs/core/ajax/ajaxtooltip.php
@@ -52,7 +52,7 @@ if (GETPOSTISSET('option')) {
 
 // Load object according to $element
 $object = fetchObjectByElement($id, $objecttype);
-if (empty($object->element)) {
+if (is_numeric($object) || empty($object->element)) {
 	httponly_accessforbidden('Failed to get object with fetchObjectByElement(id='.$id.', objecttype='.$objecttype.')');
 }
 
--- a/htdocs/core/ajax/extraparams.php
+++ b/htdocs/core/ajax/extraparams.php
@@ -46,7 +46,9 @@ $type = GETPOST('type', 'alpha');
 
 // Load object according to $id and $element
 $object = fetchObjectByElement($id, $element);
-
+if (is_numeric($object) || empty($object->element)) {
+	httponly_accessforbidden('Failed to get object with fetchObjectByElement(id='.$id.', objecttype='.$objecttype.')');
+}
 $module = $object->module;
 $element = $object->element;
 $usesublevelpermission = ($module != $element ? $element : '');
--- a/htdocs/core/ajax/fileupload.php
+++ b/htdocs/core/ajax/fileupload.php
@@ -45,6 +45,9 @@ $elementupload = $element;
 
 // Load object according to $id and $element
 $object = fetchObjectByElement($id, $element);
+if (is_numeric($object) || empty($object->element)) {
+	httponly_accessforbidden('Failed to get object with fetchObjectByElement(id='.$id.', objecttype='.$objecttype.')');
+}
 
 $module = $object->module;
 $element = $object->element;
--- a/htdocs/core/ajax/loadinplace.php
+++ b/htdocs/core/ajax/loadinplace.php
@@ -45,6 +45,9 @@ $id = $fk_element;
 
 // Load object according to $id and $element
 $object = fetchObjectByElement($id, $element);
+if (is_numeric($object) || empty($object->element)) {
+	httponly_accessforbidden('Failed to get object with fetchObjectByElement(id='.$id.', objecttype='.$objecttype.')');
+}
 
 $module = $object->module;
 $element = $object->element;
--- a/htdocs/core/ajax/objectonoff.php
+++ b/htdocs/core/ajax/objectonoff.php
@@ -52,7 +52,7 @@ $format = 'int';
 
 // Load object according to $id and $element
 $object = fetchObjectByElement($id, $element);
-if (!is_object($object)) {
+if (is_numeric($object) || empty($object->element)) {
 	httponly_accessforbidden("Bad value for combination of parameters element/field: Object not found.");	// This includes the exit.
 }
 
--- a/htdocs/core/ajax/saveinplace.php
+++ b/htdocs/core/ajax/saveinplace.php
@@ -57,6 +57,9 @@ savemethodname:
 
 // Load object according to $id and $element
 $object = fetchObjectByElement($id, $element);
+if (is_numeric($object) || empty($object->element)) {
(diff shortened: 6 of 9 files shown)


eldy:
Is it a reproductible bug ? 
If yes can you describe the process to reproduce the bug ?

rycks:
hello @eldy 
i'm sorry i forget to paste the definition of function (i'm sure it will be enough)

```
/**
 * Fetch an object from its id and element_type
 * Inclusion of classes is automatic
 *
 * @param	int     	$element_id 	Element id
 * @param	string  	$element_type 	Element type ('module' or 'myobject@mymodule' or 'mymodule_myobject')
 * @param	string     	$element_ref 	Element ref (Use this or element_id but not both)
 * @return 	int|object 					object || 0 || <0 if error
 */
function fetchObjectByElement($element_id, $element_type, $element_ref = '')
```

then due to the fact that this function MAY / COULD return a int 0 or <0 we can't trust on it and make some tests like that

```
$object = fetchObjectByElement($id, $element);
.../...
$object->property
```

because $object could be 0 or integer < 0 ...

but for sure result of empty($object->property) could be not affected by that because of specificity of empty but i think it's a good practice to do that to avoid biggest errors like there is in htdocs/resource/element_resource.php line 123

```
$objstat->element = $element;
```

maybe it will be better to split that PR in two parts: one really a bug with other use than with "empty" and an other one for something like "code quality" for all use with empty function ? or just remove all of commit concern the use with empty function ?

please let me know

john-botella:
I just read fetchObjectByElement function and this is my observation : 

I think this function still does too many things at all and need to be factor by split in two. 

- newObjectByElement
  -> Return false or an object if object exist

- keep fetchObjectByElement or rename It in loadObjectByElement to keep compatibility with old returned values  (and deprecate old name) .
It's role is to fetch element so, the result need to be clear, a good way is to return false for error, null for not found and an object on success. this kind of usage allow us to write simple tests like (!empty($res)) and/or deepen error result with a ($res !== false) test.

results like 0, -1 or +x are good for results on fetch but not on getting something it could be confusing.


I hope this helps, if you like this suggestion I'll do the refactoring in develop.

eldy:
> Is it a reproductible bug ?
> If yes can you describe the process to reproduce the bug ?



> hello @eldy i'm sorry i forget to paste the definition of function (i'm sure it will be enough)
> 
> ```
> /**
>  * Fetch an object from its id and element_type
>  * Inclusion of classes is automatic
>  *
>  * @param	int     	$element_id 	Element id
>  * @param	string  	$element_type 	Element type ('module' or 'myobject@mymodule' or 'mymodule_myobject')
>  * @param	string     	$element_ref 	Element ref (Use this or element_id but not both)
>  * @return 	int|object 					object || 0 || <0 if error
>  */
> function fetchObjectByElement($element_id, $element_type, $element_ref = '')
> ```
> 
> then due to the fact that this function MAY / COULD return a int 0 or <0 we can't trust on it and make some tests like that
> 
> ```
> $object = fetchObjectByElement($id, $element);
> .../...
> $object->property
> ```
> 
> because $object could be 0 or integer < 0 ...
> 
> but for sure result of empty($object->property) could be not affected by that because of specificity of empty but i think it's a good practice to do that to avoid biggest errors like there is in htdocs/resource/element_resource.php line 123
> 
> ```
> $objstat->element = $element;
> ```
> 
> maybe it will be better to split that PR in two parts: one really a bug with other use than with "empty" and an other one for something like "code quality" for all use with empty function ? or just remove all of commit concern the use with empty function ?
> 
> please let me know

I am sorry, but with the answer, i am still not able to reproduce any bug.
If there is a bug, can you describe the process to reproduce it ?  If there is no process a user can do to reproduce it, this is not a bug for a maintenance version, but a quality concern question and this should be pushed into develop.

rycks:
OK i'm on vacation for one week and afk please let me some time to send you some examples. But you can follow the link to French forum : there is some external modules where that bug is visible.

Eric

john-botella:
@rycks take a look at #30036 I have factor fetchObjectByElement and split into too new functions : newObjectByElement and loadObjectByElement 


The loadObjectByElement goal is to keep simple tests in code and to be sure the function have only one behavior according to input as fail or success to find element in base and avoid security issue...

rycks:
> @rycks take a look at #30036 I have factor fetchObjectByElement and split into too new functions : newObjectByElement and loadObjectByElement

@atm-john ok but that will not fix the bug for "old" versions of dolibarr ...

john-botella:
@rycks Yes, that's true, but it's a necessary step. Because of the sensitivity and criticality of these parts of the code, I start in the develop branch to be on the safe side. 

And then I'll see how to make a clean backport or just fix it while allowing the transition to cleaner code from V19 to V20.
(and save Laurent a headache when he has to pull up the fix from V19 to V20.)

lvessiller-opendsi:
> @rycks Yes, that's true, but it's a necessary step. Because of the sensitivity and criticality of these parts of the code, I start in the develop branch to be on the safe side.
> 
> And then I'll see how to make a clean backport or just fix it while allowing the transition to cleaner code from V19 to V20. (and save Laurent a headache when he has to pull up the fix from V19 to V20.)

I understand your point of view, but it seems to be a bug with PHP warnings and I think we should correct it also in V18. Why only fix on V19 and V20 ?

john-botella:
@lvessiller-opendsi  V18 too (just mistake on version number) ;-)

eldy:
If there is a user bug, the process to reproduce it must be described (in another issue or in this one).

For @atm-john, to answer your question, a backport of your current refactoring in an old version will be necessary ONLY if it fixes a proven bug in this old version. No need in other cases...

john-botella:
@eldy for all my fix I have found a dirty way to fix and you merged it, My PRs of Quality in develop are here to allow me create final cleans fix. So bugs are already proven bug by the way.

eldy:
> @eldy for all my fix I have found a dirty way to fix and you merged it, My PRs of Quality in develop are here to allow me create final cleans fix. So bugs are already proven bug by the way.
>

Sorry i didn't understand your comment @atm-john 
- Are you speaking about this PR or about another one on the same topic that replace this one ?
- Does this PR fix a bug existing in current v18 ? If yes, can you describe how to reproduce it so we can add a phpunit test ?  
- If not, it is change for quality of code ? If yes, it must be submitted in develop.

rycks:
i think i could close this one and switch to john stuff because of he says that his fix include 18.0 version

https://github.com/Dolibarr/dolibarr/pull/30036

PR #30000, opened 2024-06-12
NEW : Duplicate ref_client from order origin to intervention.

dolibarr95:
# NEW[Duplicate ref_client order origin to intervention]
When creating an intervention from an order, duplicate the ref_client (as is already done for public and private notes).


dolibarr95:
🥳 https://www.dolibarr.fr/forum/t/et-le-prix-de-la-30000eme-pr-revient-a/46690

htdocs/fichinter/card.php
```diff
@@ -882,6 +882,7 @@
 
 			$soc = $objectsrc->thirdparty;
 
+			$ref_client = (!empty($objectsrc->ref_client) ? $objectsrc->ref_client : GETPOST('ref_client', 'alpha'));
```
eldy, line 885: Imagine you crane an intervention from a proposal.
The proposal as the ref PRABC for you and the proposal as the ref PR456 on customer side (so ref_client is PR456.

If you copy the ref of proposal into the intervention, you will got the ref of intervention that is
IN123 and the reference of this intervention into the customer system of the client will ba saved as PR456.
But this is a reference of a proposal not of an intervention. You are mixing apple and pear here. So i suggest you introduce a hidden option
INTERVENTION_SET_CUSTOMER_PROPOSAL_REF_WITH_CUSTOMER_INTERVENTION_REF

dolibarr95:
Hello
@eldy tks for reading and feedback.

PR #30004, opened 2024-06-12
FIX : issue to get the right files exported in Quadratrus export.php

josett225:
# FIX 
The export of files does not work correctly when using Accounting option export Quadratus with file export. The files are not attached to the zip until we uncomment the 3 lines of this PR.

### DIFF
--- a/htdocs/accountancy/bookkeeping/export.php
+++ b/htdocs/accountancy/bookkeeping/export.php
@@ -1190,7 +1190,6 @@ while ($i < min($num, $limit)) {
 			require_once DOL_DOCUMENT_ROOT.'/compta/facture/class/facture.class.php';
 			$objectstatic = new Facture($db);
 			$objectstatic->fetch($line->fk_doc);
-			$modulepart = 'facture'; // or use $objectstatic->element
 
 			$filename = dol_sanitizeFileName($line->doc_ref);
 			$filedir = $conf->facture->dir_output.'/'.dol_sanitizeFileName($line->doc_ref);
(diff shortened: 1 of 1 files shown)


htdocs/accountancy/bookkeeping/export.php
```diff
@@ -1190,7 +1190,7 @@
 			require_once DOL_DOCUMENT_ROOT.'/compta/facture/class/facture.class.php';
 			$objectstatic = new Facture($db);
 			$objectstatic->fetch($line->fk_doc);
-			//$modulepart = 'facture';
+			$modulepart = 'facture'; // or use $objectstatic->element
```
eldy, line 1193: The variable $modulepart is defined into a if but not used inside this if. This make me think it is defined at wrong place.
Where do you need to use it ?

rycks: @eldy maybe because of a `global $modulepart;` into completeFileArrayWithDatabaseInfo function used to build the zip file i think but it's too hard to follow all stack call to be sure

rycks: so @josett225 could you please help us to find why that variable set will change the result ?

maybe a dolibarr in debug mode, collect the log with $modulepart in comments as official dolibarr sources, then apply your patch and redo the export, compare the log and send us more informations

thanks

josett225: I will do.

josett225: Hi @eldy and @rycks 
After diffing more in several versions, I found out we could remove $modulepart and replace it $objectstatic->element everywhere.
I changed the code accordingly and remove any mention of $modulepart

Regards
Jose

PR #30005, opened 2024-06-12
# NEW : add ActionComm categories accessible thru API

josett225:
# NEW : Add Actioncomm Categories management thru API 

The main goal is to manage ActionComm Categories thru API
All the changes are done in api_categories.class.php file.


josett225 (commented):
Change Actioncomm to ActionComm

PR #30008, opened 2024-06-12
2 Hooks, 1 fix to correct calculation (remise was + instead of -), 1 …

vsatmydynipnet:
I asked in the forum how to get hooks i need into core and I was adviced to create a PR here.

This PR adds 2 hooks that makes it possible to add e.g. discount_percentage 2-5, as used in food industry, to handle up to 5 discounts. The module will be published later.

Furthermore there is one fix where remise was added and not subtracted

I also changed the price display to a function  to be able to use the hook there too

All Changes are commented with VS....

Thank you for adding the hooks

Greetings

Re.: this is my first PR, if I have done something wrong, just let me know.

### DIFF
--- a/htdocs/fourn/class/fournisseur.product.class.php
+++ b/htdocs/fourn/class/fournisseur.product.class.php
@@ -40,6 +40,11 @@ require_once DOL_DOCUMENT_ROOT.'/product/class/productfournisseurprice.class.php
 class ProductFournisseur extends Product
 {
 	/**
+	 * @var string ID to identify managed object
+	 */
+	public $element = 'fournisseur_product';
+
+	/**
 	 * @var DoliDB Database handler.
 	 */
 	public $db;
@@ -802,7 +807,7 @@ class ProductFournisseur extends Product
 	public function find_min_price_product_fournisseur($prodid, $qty = 0, $socid = 0)
 	{
 		// phpcs:enable
-		global $conf;
+		global $conf, $hookmanager;
 
 		if (empty($prodid)) {
 			dol_syslog("Warning function find_min_price_product_fournisseur were called with prodid empty. May be a bug.", LOG_WARNING);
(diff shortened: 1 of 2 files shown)


frederic34:
@vsatmydynipnet use only tabs to indent, not spaces

htdocs/fourn/class/fournisseur.product.class.php
```diff
@@ -889,6 +894,18 @@ public function find_min_price_product_fournisseur($prodid, $qty = 0, $socid = 0
 							$fourn_unitprice_with_discount = $fourn_unitprice * (1 - $record["remise_percent"] / 100);
 						}
 					}
+
+                                        /* VSPR let us recalculate the objects values. Based on the Modul Extra Discounts adding 4 more discounts to supplier*/
+                                        global $action;
+                                        $hookmanager->initHooks(array($this->element . 'minprice'));
```
eldy, line 900: inithook is a method called by the main page or cron launcher. 
Should not be done inside a low level method.

eldy (changes requested):

htdocs/product/list.php
```diff
@@ -1972,7 +1972,9 @@
 					if ($product_fourn->product_fourn_price_id > 0) {
 						if ((isModEnabled("fournisseur") && $user->hasRight('fournisseur', 'lire') && !getDolGlobalString('MAIN_USE_NEW_SUPPLIERMOD')) || (isModEnabled("supplier_order") && $user->hasRight('supplier_order', 'lire')) || (isModEnabled("supplier_invoice") && $user->hasRight('supplier_invoice', 'lire'))) {
 							$htmltext = $product_fourn->display_price_product_fournisseur(1, 1, 0, 1);
-							print '<span class="amount">'.$form->textwithpicto(price($product_fourn->fourn_unitprice * (1 - $product_fourn->fourn_remise_percent / 100) - $product_fourn->fourn_remise).' '.$langs->trans("HT"), $htmltext).'</span>';
+// VSPR fix: this should not be hardcoded // price should be created by function so we changed the function too. 2 now comes back with the number only
```
eldy, line 1975: Do not mix fix and adding new feature in same pr.

eldy (changes requested):

htdocs/fourn/class/fournisseur.product.class.php
vsatmydynipnet: I am not a real developer so I searched, but in fournisseur.product.class.php no other initHook was found. Is there a way to see the initialized hook at this position?

htdocs/product/list.php
vsatmydynipnet: ok, that was also a new one because the value now comes from the function and the function reacts to value showunitprice "2" it its call. I use that one in my hook to calculate the 5 discounts.

vsatmydynipnet:
I will create a new one.

PR #30011, opened 2024-06-13
FIX #30010 : Use conf TICKET_MESSAGE_MAIL_INTRO instead of translation key

### LINKED ISSUE #30010: When sending ticket by mail, TicketMessageMailIntroText is forced in the message
### Bug

When you notify a user by mail when adding a new message in ticket, the message sent has the translation of TicketMessageMailIntroText on top of your message everytime. It doesn't take the const TICKET_MESSAGE_MAIL_INTRO into account.

### Dolibarr Version

17 - develop

### Environment PHP

_No response_

### Environment Database

_No response_

_No response_

### Attached files

_No response_

evarisk-micka:
Can it be added to version 18.0 too ?


eldy:
To get a fix being in v18, you must push your PR into v18 branch instead of develop...

PR #30013, opened 2024-06-13
fk_product could be = -1 then line added is not linked to selected pr…

rycks:
Situation i check some entries from project to make a new invoice from time spent on tasks and i choose a service in dropdown (adminsys80)

image

And here is the result: one line added to invoice (ok) but the service i selected in previous step is not used and by default that is a product (bad consequences for accounting)

image

that is because of fk_product = -1 in my situation ... that's why i purpose to add a condition in if

note: some lines before in the else i have to add a test on idprod <= 0, idprod is the value from the dropdown


rycks:
hmmm that's not so basic ... please wait

rycks:
ok, no more draft status for that but i don't know how to change the status @eldy maybe some non-jedi rights limitations ?

eldy:
The draft status is set by the user that submit the PR.
Such PR must be ignored asnlong as the PR has the draft status. It means "i pushed the PR but i have not finished the job".
I think only the submiter can change the status.

PR #30036, opened 2024-06-14
QUAL : Factoring fetchObjectByElement

john-botella:
# see  #29996

- [x] Create function newObjectByElement
   Used to only create a new instance of an object 
- [x] Create function loadObjectByElement  
   Do the same as  fetchObjectByElement by use a different kind of return
   The loadObjectByElement goal is to keep simple tests in code and to be sure the function have only one behavior according to input as fail or success to find element in base and avoid security issue...

- [x] factoring fetchObjectByElement
- [x] Full Tests

my comment pasted from #29996

> I just read fetchObjectByElement function and this is my observation :
> 
> I think this function still does too many things at all and need to be factor by split in two.
> 
> newObjectByElement
> -> Return false or an object if object exist
> 
> keep fetchObjectByElement or rename It in loadObjectByElement to keep compatibility with old returned values (and deprecate old name) .
> It's role is to fetch element so, the result need to be clear, a good way is to return false for error, null for not found and an object on success. this kind of usage allow us to write simple tests like (!empty($res)) and/or deepen error result with a ($res !== false) test.
> 
> results like 0, -1 or +x are good for results on fetch but not on getting something it could be confusing.
> 
> I hope this helps, if you like this suggestion I'll do the refactoring in develop.
> 
> 
> 
>


john-botella:
Add also error code to global error message to simplify debug
image

john-botella:
@eldy I don't understand the "CI / phan / Run phan" error

htdocs/core/lib/functions.lib.php
```diff
@@ -12822,56 +12857,96 @@ function fetchObjectByElement($element_id, $element_type, $element_ref = '', $us
...
+ * @param	string  	$element_type 		Element type ('module' or 'myobject@mymodule' or 'mymodule_myobject')
+ * @param	string     	$element_ref 		Element ref (Use this or element_id but not both. If id and ref are empty, object with no fetch is returned)
+ * @param	int<0,2>	$useCache 			If you want to store object in cache or get it from cache 0 => no use cache , 1 use cache, 2 force reload  cache
+ * @param	int			$maxCacheByType 	Number of object in cache for this element type
+ * @param	string		$errMsgCode 		Error message
+ * @return null|false|object 				object || null on not found || false if error
+ * @see getElementProperties()
+ * @see newObjectByElement()
+ */
+function loadObjectByElement($element_id, $element_type, $element_ref = '', $useCache = 0, $maxCacheByType = 10, &$errMsgCode = '')
```
eldy, line 12898: There is already a method to get an object from its id and type.
It is "fetchObjectByElement"

eldy:
Refactoring and optimization is not a fix. 
Thanks to rename PR with NEW or QUAL or PERF
if it is a fix, thanks to describe how to reproduce.

john-botella:
@eldy ready

I rename it QUAL

yes this is not a fix but needed for my next real FIX PR to correct some usage of  fetchObjectByElement and i will replace it by newObjectByElement or loadObjectByElement (and I try making atomic changes each PR)

This is linked to #30086, #29369, I work on a fix for develop to allow getElementProperties to add more information for  rights tests because in somme cases like the read test based on an element fail and create strange behavior. 
This is due to security improvement on selectforform and other stuff but V19 wasn't ready enought 
For V19, I limited the correction for external modules to a simple backport hook, but for V20 we need a more effective fix

htdocs/core/lib/functions.lib.php
```diff
@@ -12822,56 +12857,96 @@ function fetchObjectByElement($element_id, $element_type, $element_ref = '', $us
...
 
-					return $objecttmp;
-				}
-			} else {
-				return $objecttmp;	// returned an object without fetch
-			}
-		} else {
-			return -1;
+/**
+ * Fetch an object from its id and element_type
```
atm-john, line 12885:
```suggestion
 * Load and Fetch an object from its id and element_type
```

PR #26084, opened 2023-09-30
FIX #26017 ASSET: annual depreciation starting year

BurksEngineering:
FIX #26017 ASSET: annual depreciation starting year

When using annual depreciation of an Asset, the depreciation table now begins with the fiscal year when the asset was put in service or acquired (whichever occurred first) instead of the fiscal year of the current (server time) moment. This allows you to regenerate the depreciation table properly if you need to modify the acquisition or in-service date, and aligns with the way monthly and daily depreciation tables are generated.

The choice of using the earliest of acquisition date or in-service date is based on the calculation of $depreciation_date_start later in the function.

PR #30040, opened 2024-06-14
FIX ASSET: annual depreciation starting year (Again ;-)) #26084

josett225:
FIX Update asset.class.php
This is an old issue that has been already fixed by  https://github.com/Dolibarr/dolibarr/pull/26084

I tested it several times and it works and it required to close this Asset feature.

### DIFF
--- a/htdocs/asset/class/asset.class.php
+++ b/htdocs/asset/class/asset.class.php
@@ -863,7 +863,7 @@ class Asset extends CommonObject
 			// Get fiscal period
 			require_once DOL_DOCUMENT_ROOT . '/core/lib/date.lib.php';
 			require_once DOL_DOCUMENT_ROOT . '/core/lib/accounting.lib.php';
-			$dates = getDefaultDatesForTransfer();
+			$dates = getCurrentPeriodofFiscalYear($this->db, $conf, $this->date_start > $this->date_acquisition ? $this->date_start : $this->date_acquisition);
 			$init_fiscal_period_start = $dates['date_start'];
 			$init_fiscal_period_end = $dates['date_end'];
 			if (empty($init_fiscal_period_start) || empty($init_fiscal_period_end)) {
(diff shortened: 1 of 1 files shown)


eldy:
@aspangaro is it ok for you ?

josett225:
Hi @eldy and @aspangaro 
Not sure what has to be done here to get this approved.

Without this, nobody is able at least to test this module.

So even if it could potentially not be perfect, it is much better than the impossibility to move forward in testing and finish this module. 

In my side I tested it in 3 customers and it works. 

Thanks

eldy:
> Hi @eldy and @aspangaro Not sure what has to be done here to get this approved.
> 
> Without this, nobody is able at least to test this module.
> 
> So even if it could potentially not be perfect, it is much better than the impossibility to move forward in testing and finish this module.
> 
> In my side I tested it in 3 customers and it works.
> 
> Thanks

@josett225 Can you describe the process to reproduce the bug ?

@aspangaro The fix is just one line, it appears good to me but don't know how to test it. Can you give your point of view ?

htdocs/asset/class/asset.class.php
```diff
@@ -863,7 +863,7 @@ public function calculationDepreciation()
 			// Get fiscal period
 			require_once DOL_DOCUMENT_ROOT . '/core/lib/date.lib.php';
 			require_once DOL_DOCUMENT_ROOT . '/core/lib/accounting.lib.php';
-			$dates = getDefaultDatesForTransfer();
+			$dates = getCurrentPeriodofFiscalYear($this->db, $conf, $this->date_start > $this->date_acquisition ? $this->date_start : $this->date_acquisition);
```
aspangaro, line 866: Typo in function Name : getCurrentPeriod**O**fFiscalYear(

aspangaro:
@eldy Ok for me with typo fix.

When using annual depreciation of an asset, the depreciation table now begins with the fiscal year when the asset was put in service or acquired (whichever occurred first) instead of the fiscal year of the current (server time) moment. This allows you to regenerate the depreciation table properly if you need to modify the acquisition or in-service date, and aligns with the way monthly and daily depreciation tables are generated.

The choice of using the earliest of acquisition date or in-service date is based on the calculation of $depreciation_date_start later in the function.

PR #30051, opened 2024-06-17
FIX rounding update price for expense report lines

lvessiller-opendsi:
FIX rounding update price for expense report lines
- when you set "MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND" const to 1 and when there is a difference between two methods to compute "Total VAT", in some cases total amount (TTC) is different of unit price amount  (TTC) even if quantity is set to 1.

In expense report, if you add a new line of an existing line with "MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND" set to 1 : 
image
You got a non coherent "total amount" (TTC) : 6,21 instead of 6,20.
Here the quantity is 1, so the total amount should be the same as unit price value (TTC)

**Before**
What you got before the fix :
image

**After**
What you get after the fix :
image
image
You remark you get "5,17" in HT amount for line 1 and "5,16" in HT amount for line 2 but it's a correction value to have the HT total amount to "10,33" (in this case it's better to set quantity to 2)

And an other example with a positive diff between vat amounts :
image
image

 thanks @kkhelifa-opendsi for your help

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -3962,7 +3962,11 @@ abstract class CommonObject
 				$total_ttc_by_vats[$obj->vatrate] += $obj->total_ttc;
 
 				if ($forcedroundingmode == '1') {	// Check if we need adjustment onto line for vat. TODO This works on the company currency but not on foreign currency
-					$tmpvat = price2num($total_ht_by_vats[$obj->vatrate] * $obj->vatrate / 100, 'MT', 1);
+					if ($base_price_type == 'TTC') {
+						$tmpvat = price2num($total_ttc_by_vats[$obj->vatrate] - $total_ht_by_vats[$obj->vatrate], 'MT', 1);
+					} else {
+						$tmpvat = price2num($total_ht_by_vats[$obj->vatrate] * $obj->vatrate / 100, 'MT', 1);
+					}
 					$diff = price2num($total_tva_by_vats[$obj->vatrate] - (float) $tmpvat, 'MT', 1);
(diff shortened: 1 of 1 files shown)


eldy:
I have added a small popup when mouse is on the VAT rate so we can see the detail of total in database. So we can see this in your example (before your PR):
image

After your PR, what is the value inside this popup ?

lvessiller-opendsi:
> I have added a small popup when mouse is on the VAT rate so we can see the detail of total in database. So we can see this in your example (before your PR): image
> 
> After your PR, what is the value inside this popup ?

After my PR,
Total HT=5,16<br>Total TVA=1,04<br>Total TTC=6,20
image

eldy:
> > I have added a small popup when mouse is on the VAT rate so we can see the detail of total in database. So we can see this in your example (before your PR): image
> > After your PR, what is the value inside this popup ?
> 
> After my PR, Total HT=5,16Total TVA=1,04Total TTC=6,20 image

And what is the popup on first line (the one with Total HT = 5.17) ?

lvessiller-opendsi:
> > > I have added a small popup when mouse is on the VAT rate so we can see the detail of total in database. So we can see this in your example (before your PR): image
> > > After your PR, what is the value inside this popup ?
> > 
> > 
> > After my PR, Total HT=5,16Total TVA=1,04Total TTC=6,20 image
> 
> And what is the popup on first line (the one with Total HT = 5.17) ?

After PR, on the first line you have : 
Total HT=5,17
Total TVA=1,03
Total TTC=6,20

PR #30053, opened 2024-06-17
fix : mismatch of variables when changing there names

grandoc:
% per month and total was no more shown


grandoc:
Why did you closed it ?
It was fixing a regression

PR #30063, opened 2024-06-18
20.0 FIX warning: undefined variable $total_ht

atm-florianm:
# FIX warning

The variable `$total_ht` is not always defined, but it is passed to `price` and `price2num`. This just is a quickfix: maybe with more analysis there is a better fix.

### DIFF
--- a/htdocs/societe/consumption.php
+++ b/htdocs/societe/consumption.php
@@ -737,8 +737,8 @@ if ($sql_select) {
 		print '<td></td>';
 	}
 	print '<td class="right">'.$total_qty.'</td>';
-	print '<td class="right">'.price($total_ht).'</td>';
-	print '<td class="right">'.price(price2num($total_ht / (empty($total_qty) ? 1 : $total_qty), 'MU')).'</td>';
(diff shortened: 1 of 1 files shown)


htdocs/societe/consumption.php
```diff
@@ -737,8 +737,8 @@
 		print '<td></td>';
 	}
 	print '<td class="right">'.$total_qty.'</td>';
-	print '<td class="right">'.price($total_ht).'</td>';
-	print '<td class="right">'.price(price2num($total_ht / (empty($total_qty) ? 1 : $total_qty), 'MU')).'</td>';
+	print '<td class="right">'.price($total_ht ?? 0).'</td>';
+	print '<td class="right">'.price(price2num($total_ht ?? 0 / (empty($total_qty) ? 1 : $total_qty), 'MU')).'</td>';
```
atm-florianm, line 741:
```suggestion
	print '<td class="right">'.price(price2num(($total_ht ?? 0) / (empty($total_qty) ? 1 : $total_qty), 'MU')).'</td>';
```

PR #30070, opened 2024-06-18
2 new hooks to be able to calculate more then one discount

eushopnet:
# Instructions

This adds 2 Hooks and changes the price display to come from the function.

### DIFF
--- a/htdocs/fourn/class/fournisseur.product.class.php
+++ b/htdocs/fourn/class/fournisseur.product.class.php
@@ -149,7 +149,6 @@ class ProductFournisseur extends Product
 	const STATUS_OPEN = 1;
 	const STATUS_CANCELED = 0;
 
-
 	/**
 	 *	Constructor
 	 *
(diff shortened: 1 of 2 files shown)


frederic34:
new features need to be added in develop branch

eushopnet:
redone as https://github.com/Dolibarr/dolibarr/pull/30082

and added a question there

Unsure if this one https://github.com/Dolibarr/dolibarr/pull/30066 to 19.0, which was used, was wrong too? Will this be in future versions too?

eldy:
> redone as #30082
> 
> and added a question there
> 
> Unsure if this one #30066 to 19.0, which was used, was wrong too? Will this be in future versions too?

PR #30066 was accepted in v19 as it was a hidden option not enabled by default (and just few chars) so addition as no effect (in perfect word, it would have be better to push it on develop). But it is now in v19, so will be in v20, vdevelop...
For this one #30070, you must push on develop branch.

PR #30071, opened 2024-06-18
NEW #30044

### LINKED ISSUE #30044: When billing multiple orders, all order numbers are implicitly added in Public Note without the ability to edit/remove
### Bug

As the title says. If we select sales orders SO0001, SO0002, SO0003 and bill them as "one invoice for al third party", there is added a box on the invoice containing the text SO0001 (date), SO0002 and so on, but this information is added on-the-fly in `core/lib/pdf.lib.php`, with the side effect that we cannot remove it and there is no setting to not do this.

Suggestion: instead of letting `pdf_getLinkedObject()` handle this on the fly, make an actual Public Note immediately when clicking "Bill orders". That way, we can edit the note to whatever we want (in this case we want to delete all the information).

### Dolibarr Version

19.0, probably all versions

### Environment PHP

_No response_

### Environment Database

_No response_

1. Go to sales orders and choose minimum 2 from the same third party
2. Click Bill orders in top menu
3. Select "Yes" on "Create one invoice per third party"

For the new invoice draft, generate the pdf-file and see the sales order info in the box, while Public Note is empty (i.e. you cannot edit it).

zahwyrix:
Close #30044


htdocs/core/lib/pdf.lib.php
```diff
@@ -2500,13 +2500,7 @@ function pdf_getLinkedObjects(&$object, $outputlangs)
 		} elseif ($objecttype == 'commande' || $objecttype == 'supplier_order') {
 			$outputlangs->load('orders');
 
-			if (count($objects) > 1 && count($objects) <= (getDolGlobalInt("MAXREFONDOC") ? getDolGlobalInt("MAXREFONDOC") : 10)) {
```
eldy, line 2503: Removing generation from pdf_getLinkedObjects() to replace it by public_note seems a good idea.
However, the option getDolGlobalInt("MAXREFONDOC") must still be used to limit size of the public_note.

htdocs/commande/list.php
```diff
@@ -553,6 +559,11 @@
 				}
 			}
 
+			if ($currentIndex <= (getDolGlobalInt("MAXREFONDOC") ? getDolGlobalInt("MAXREFONDOC") : 10)) {
+				$objecttmp->note_public = dol_concatdesc($objecttmp->note_public, $langs->transnoentities($cmd->ref).(empty($cmd->ref_client) ? '' : ' ('.$cmd->ref_client.')'));
+				$objecttmp->update($user);
+			}
+
```
eldy, line 566: To avoid such errors, you should think to activate the "pre-commit" hooks in your development environment:
See https://wiki.dolibarr.org/index.php?title=Environment_and_development_tools#Install_Pre-commit_githook

PR #30088, opened 2024-06-19
NEW Add option to clone its default BOM when cloning a product

vmaury:
This option is available if 

- BOM module is enabled (elementary my dear Watson)
- product has a default BOM
- user has right to create BOM
clone BOM


eldy:
What is the default value of the new checkbox ? Should be "not" checked by default.

vmaury:
Ok. Now :
- option "clone default BOM" not checked by default ( can be pre-checked with constant BOM_CLONE_DEFBOM)
- and I add another option to clone all BOMs linked to a product
clone BOM2

PR #30090, opened 2024-06-19
2 new hooks to be able to calculate more then one discount (Dev Branche)

eushopnet:
Instructions

This adds 2 Hooks and changes the price display to come from the function.

Now to develop branch as requested in https://github.com/Dolibarr/dolibarr/pull/30070

Unsure if this one https://github.com/Dolibarr/dolibarr/pull/30066 to 19.0, which was used, was wrong too? Will this be in future versions too?


eushopnet:
can somebody help with this phan error?
not even a fresh fork does work....

PR #30092, opened 2024-06-19
NEW Implementing Billable tasks on projects using new attribute "billable"

vdegrandpre:
Functionality with attribute added in #30014.

Realize #29913.


vdegrandpre:
Will fix issues in test cases.

vdegrandpre:
@eldy is it possible that windows-ci test runner do not apply 20.0.0-21.0.0.sql migration file?

eldy:
> @eldy is it possible that windows-ci test runner do not apply 20.0.0-21.0.0.sql migration file?

yes, it is. Fixed.

htdocs/langs/de_DE/projects.lang
```diff
@@ -302,3 +302,4 @@ NewLeadbyWeb=Ihre Nachricht bzw. Anfrage wurde erfasst. Wir werden uns so bald w
 NewLeadForm=Neues Kontaktformular
 LeadFromPublicForm=Online-Interessent per öffentlichem Formular
 ExportAccountingReportButtonLabel=Bericht erhalten
+Billable = Verrechenbar
```
eldy, line 305: Update language file is useless. Only en_US must be filled. Other files are erased by transifex sync.

htdocs/projet/tasks.php
```diff
@@ -171,6 +173,7 @@
 	't.progress_summary' => array('label' => "TaskProgressSummary", 'checked' => 1, 'position' => 10),
 	't.budget_amount' => array('label' => "Budget", 'checked' => 0, 'position' => 11),
 	'c.assigned' => array('label' => "TaskRessourceLinks", 'checked' => 1, 'position' => 12),
+	't.billable' => array('label' => "Billable", 'checked' => 1, 'position' => 13),
```
eldy, line 176: This property must be moved 2 lines after, inside the test 
`if ($object->usage_bill_time) {`
So it will appear only if the project has the usage "Bill tasks"

htdocs/langs/de_DE/projects.lang
vdegrandpre: Good to know. 
Do I have to revert the change to language files ?

htdocs/projet/tasks.php
vdegrandpre: Still ok, pushing change.

htdocs/projet/class/task.class.php
```diff
@@ -174,7 +175,8 @@ class Task extends CommonObjectLine
 	public $task_parent_ref;
 	public $task_parent_position;
 
-
+	// Billable state (boolean, default to 1)
+	public $billable = 1;
```
frederic34, line 179: @vdegrandpre  int or boolean ? if default to 1, it's int. Can you add phpdoc ?

htdocs/langs/de_DE/projects.lang
eldy: Better but not mandatory...

PR #30099, opened 2024-06-20
FIX #30076

### LINKED ISSUE #30076: ajax autocomplete crash on 50000 products
### Bug

Enhanced Autocomplete Query Handling for Large Datasets
We've updated the ajax_autocompleter function to efficiently handle large datasets, even with more than 100,000 items. The new implementation limits the number of results returned by the autocomplete query, enhancing performance and usability. This limit can be dynamically set using the global configuration AJAX_LIMIT_AUTOCOMPLETE_RESULTS.

Here's the key change in the function:

Before: in ajax.lib.php on line 121 .. look for 

source: function(request, response) {
    $.get("'.$url.($urloption ? '?'.$urloption : '').'", { "'.str_replace('.', '_', $htmlname).'": request.term }, function(data){
        if (data != null) {
            response($.map(data, function(item) {
                if (autoselect == 1 && data.length == 1) {
                    $("#search_'.$htmlnamejquery.'").val(item.value);
                    $("#'.$htmlnamejquery.'").val(item.key).trigger("change");
                }
                // ...
            }));
        }
        // ...
    }, "json");
}

and change it to 

source: function(request, response) {
    var maxResults = '.(getDolGlobalString('AJAX_LIMIT_AUTOCOMPLETE_RESULTS') ? getDolGlobalString('AJAX_LIMIT_AUTOCOMPLETE_RESULTS') : '10').';  // Dynamically set maxResults
    $.get("'.$url.($urloption ? '?'.$urloption : '').'", { "'.str_replace('.', '_', $htmlname).'": request.term }, function(data){
        if (data != null) {
            response($.map(data.slice(0, maxResults), function(item) {

this will fix the problem and dolibarr never hangs anymore.

Greetings

### Dolibarr Version

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

_No response_

### Attached files

_No response_

zahwyrix:
FIX #30076


Humml87:
Hi @zephyriony,
thanks for the PR
We have this problem also in v19.
=> can you make the PR into the v19 branch please?
Best regards
Christian

htdocs/core/lib/ajax.lib.php
```diff
@@ -121,11 +121,11 @@ function ajax_autocompleter($selected, $htmlname, $url, $urloption = '', $minLen
 
 					// Activate the autocomplete to execute the GET
     				$("input#search_'.$htmlnamejquery.'").autocomplete({
-    					source: function( request, response ) {
-    						$.get("'.$url.($urloption ? '?'.$urloption : '').'", { "'.str_replace('.', '_', $htmlname).'": request.term }, function(data){
-								if (data != null)
-								{
-									response($.map( data, function(item) {
+						source: function(request, response) {
+							var maxResults = '.(getDolGlobalString('AJAX_LIMIT_AUTOCOMPLETE_RESULTS') ? getDolGlobalString('AJAX_LIMIT_AUTOCOMPLETE_RESULTS') : '10').'; // Dynamically set maxResults
```
eldy, line 125: Setting a limit without user consent lead to user experience trouble: They don't understand why they can't see some results and generates support ticket.
So default value must be 0. And if value is 0, we do no execute data.slice.

PR #30103, opened 2024-06-20
FIX: MRP cost is wrong when the associated BOM is using a service with duration

jyhere:
To reproduce the bug : 

1. create a service with a duration of 1 hour and a cost price of 100€

image1
image2

2. create a BOM for 1 quantity, with a single service line (our service), and a duration of 5 minutes. the line total cost is OK (100 / 12 = 8.33)

image5

3. Create a MO to produce 1 quantity, validate, click on Consume And Produce All : the end product unit cost price is wrong (5 x 100 = 500€ !)

image4

Result with the fix : 

image3

### DIFF
--- a/htdocs/mrp/mo_production.php
+++ b/htdocs/mrp/mo_production.php
@@ -990,8 +990,31 @@ if ($object->id > 0 && (empty($action) || ($action != 'edit' && $action != 'crea
 								$costprice = 0;
 							}
 						}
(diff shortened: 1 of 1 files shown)


eldy:
@jyhere i made some enhancement to your FIX. Can you check it still work after my changes ?

jyhere:
> @jyhere i made some enhancement to your FIX. Can you check it still work after my changes ?

Yes it's still working :+1:

PR #30123, opened 2024-06-21
Finished event end time

MaximilienR-easya:
# New
When you set an event to finished, it set to now the end time if it was undefined


MaximilienR-easya:
@eldy PR is fixed, can you review it ?

PR #30134, opened 2024-06-22
FIX #30132

### LINKED ISSUE #30132: Variables not taken into account for email models for Tickets
### Bug

If you configure an email model associated with a ticket, the variables are not taken into acount for the subject while it is ok in the email body.

### Dolibarr Version

18.0.5 and 19.0.2 at least

### Environment PHP

7.4.33

### Environment Database

MySQL or MariaDB 5.5.5-10.5.23-MariaDB-0+deb11u1

Create an email model associated with a ticket :
image

Try to send an email using this model :
image

### Attached files

_No response_

zahwyrix:
FIX #30132


htdocs/core/class/html.formticket.class.php
```diff
@@ -1567,7 +1567,7 @@ public function showMessageForm($width = '40%')
 			if (empty($topic)) {
 				print '<td><input type="text" class="text minwidth500" name="subject" value="['.getDolGlobalString('MAIN_INFO_SOCIETE_NOM').' - '.$langs->trans("Ticket").' '.$ticketstat->ref.'] '. $ticketstat->subject .'" />';
 			} else {
-				print '<td><input type="text" class="text minwidth500" name="subject" value="['.getDolGlobalString('MAIN_INFO_SOCIETE_NOM').' - '.$langs->trans("Ticket").' '.$ticketstat->ref.'] '.$topic.'" />';
+				print '<td><input type="text" class="text minwidth500" name="subject" value="'.make_substitutions($topic, $this->substit).'" />';
```
eldy, line 1570: Why removing the prefix 
[Thecompany - Ticket 123] ...
in default email topic ?

zephyriony: If the user has defined a subject by default, he can add it to the templates himself; on the other hand, if he has left it in, he can't remove it, which can be painful for the user experience.

PR #30143, opened 2024-06-24
use object contact as priority in MAIN_MAIL_NO_WITH_TO_SELECTED

rycks:
See  #29759 : mail to is not auto selected, that is a "user bug" not a "code bug" -> develop ... maybe in 20.0 (or too late) ?


htdocs/core/class/html.formmail.class.php
```diff
@@ -1096,7 +1097,7 @@ public function get_form($addfileaction = 'addfile', $removefileaction = 'remove
 	 */
 	public function getHtmlForTo()
 	{
-		global $langs, $form;
+		global $langs, $form, $object;
```
eldy, line 1100: Using $object as global variable has unpredicable results. ($object is often used by different purposes).
Also later having the selection of $withtoselected done inside the method breaks the idea the component html.formmail.class.

So I suggest to add a proterty ->withtoselected into the component html.formmail.class that will be used to set $withtoselected if defined. 
So the content of the preselected email can be set by the caller by setting ->withtoselected.
The code added into this method to set $withtoselected can then be moved into the caller to fill ->withtoselected.

rycks: ok, i use global to be as close as possible with other dolibarr code, but that's a good news to avoid global object :)

rycks:
@eldy done

rycks:
@eldy i'm in trouble that pr seems to stay not merged no ?

PR #30145, opened 2024-06-24
fix: bad errors variable for error return

FHenry:
Bad variable affection on error in insertExtrafield in contract addline method


rycks (approved):
ok for me

lvessiller-opendsi:
Already merged :+1:

PR #30150, opened 2024-06-25
Fix: Different behavior on extrafield updates.

Humml87:
Use "updateExtraField" instead of "insertExtraFields" to update extrafields to havethe same and correct behavior equals propal card as example.

### DIFF
--- a/htdocs/commande/card.php
+++ b/htdocs/commande/card.php
@@ -1482,16 +1482,17 @@ if (empty($reshook)) {
 
 	if ($action == 'update_extras') {
 		$object->oldcopy = dol_clone($object, 2);
+		$attribute_name = GETPOST('attribute', 'restricthtml');
 
 		// Fill array 'array_options' with data from update form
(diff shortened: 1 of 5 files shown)


Humml87:
@eldy Hi, you switch the branch too "develop". This is a problem for us. It is possible to integrate this in v20?
Then i have a chance to update our company.

eldy:
> @eldy Hi, you switch the branch too "develop". This is a problem for us. It is possible to integrate this in v20?
> Then i have a chance to update our company.

The PR appeared to me as a quality change. You feedback let me think it is a bug. Can you describe the bug if it is and how to reproduce it ?

Humml87:
@eldy 
> The PR appeared to me as a quality change. You feedback let me think it is a bug. Can you describe the bug if it is and how to reproduce it ?

We have a module that hook into the "updateExtraFieldBeforeCommit" (commonobject.class.php line 7068).

In the old code some extrafields are not update with the function "updateExtraField" from the commonobject.class.php, rather with the wrong function "insertExtraFields".
So we don't can hook in, if a extrafield is updated.

I hope you can retrace our problem.

PR #30155, opened 2024-06-25
New updating for preview php content

lamrani002:
# NEW|New updating for preview php content on Ckeditor 
for exemple preview news : 
gogog


htdocs/core/lib/process_template.lib.php
```diff
@@ -0,0 +1,58 @@
+<?php
```
eldy, line 1: This is the ajax service i think ?
If i am right, move the file into directory htdocs/core/ajax/emailingemplates.php.

htdocs/core/lib/process_template.lib.php
```diff
@@ -0,0 +1,58 @@
...
+if (!defined('NOREQUIRESOC')) {
+	define('NOREQUIRESOC', '1');
+}
+require_once '../../main.inc.php';
+require_once './files.lib.php';
+
+top_httphead();
+
+if ($_SERVER['REQUEST_METHOD'] == 'POST' && GETPOSTISSET('content')) {
+	$content = $_POST['content'];
```
eldy, line 27: $_POST use is forbidden.
You must use GETPOST()

htdocs/core/lib/process_template.lib.php
```diff
@@ -0,0 +1,58 @@
...
+	$content = str_replace('<!-- PHP_END -->', ' ?>', $content);
+
+	$directory = DOL_DATA_ROOT . '/mailing/email_template';
+	if (!is_dir($directory)) {
+		dol_mkdir($directory);
+	}
+
+	$i = 0;
+	do {
+		$filePath = $directory . '/template_' . $i++ . '.php';
```
eldy, line 40: The $directory = DOL_DATA_ROOT . '/mailing/email_template'; 
is inside the root where binary files are stored. It means the directory should be in readonly mode and is not writable by the web server on a secured installation.
Any temporaty files must be stored into a temp directory.
For example, you can use  $conf->admin->dir_temp

lamrani002: ok it s solved

PR #30164, opened 2024-06-26
NEW_CARE_GRANDCHILD

IC-Mathieu:
I added a new constant to be able to include children in the sort.
What allows children to be above parent lines

### DIFF
--- a/htdocs/core/class/html.formwebsite.class.php
+++ b/htdocs/core/class/html.formwebsite.class.php
@@ -369,7 +369,7 @@ class FormWebsite
 
 		// We disable some not ready templates
 		unset($arrayofsamples['dynamiccontent']);
-		unset($arrayofsamples['news']);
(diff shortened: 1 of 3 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -3325,13 +3325,15 @@ public function line_order($renum = false, $rowidorder = 'ASC', $fk_parent_line
 
 			dol_syslog(get_class($this)."::line_order search all parent lines", LOG_DEBUG);
 			$resql = $this->db->query($sql);
+			$grandchild = getDolGlobalInt('CARE_GRANDCHILD');
+			$grandchild = empty($grandchild) ? 0 : getDolGlobalInt('CARE_GRANDCHILD');
```
eldy, line 3329: Why second line 3329, the 3328 already return the value or 0.

PR #30170, opened 2024-06-26
Update list.php

TechServiceConseil:
Add user as default value for $search_project_user & $search_task_user in tasklist

### DIFF
--- a/htdocs/projet/tasks/list.php
+++ b/htdocs/projet/tasks/list.php
@@ -75,8 +75,8 @@ $searchCategoryCustomerList = GETPOST('search_category_customer_list', 'array');
 
 // Set search_project_user and search_task_user to current user if not already set
 if (empty($search_project_user) && empty($search_task_user)) {
-    $search_project_user = $user->id;
-    $search_task_user = $user->id;
+	$search_project_user = $user->id;
+	$search_task_user = $user->id;
 }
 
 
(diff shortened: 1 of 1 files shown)


TechServiceConseil:
Hi, 

If someone can help me to understand how to pass the autocheck, i take.

Thx

aspangaro:
> Hi,
> 
> If someone can help me to understand how to pass the autocheck, i take.
> 
> Thx

The problem don't coming from your PR. See #30169

eldy:
Can you provide a screenshot of the page concerned by the fix, after the fix.

TechServiceConseil:
Of course !


default_user

eldy:
There is already a feature to define the default filter of any list. Menu setup - default value.
Forcing to a value will have 2 bad effects:

- the permission system to read project - task already need to be assigned to project to see task, so filter is useless, except if user has permissuon to see all project he is not assigned to (but in such a case, limiting to user is rarely the expected behaviour).

- it breaks the default generic feature

- if user has selected the filter "all projects" in the home page "Projects" or "Tasks", then its choice is lost (ps: this was bugged but i fixed this in v20). With this feature fixed, selecting it user as filter is kept among other pages making this pr useless.

TechServiceConseil:
OK, i thought the filter above the table doesn't work in the default feature. If someone go here just know it's because the feature doesn't work in the same way. 
Instead of taking the "name" attribute of the html field, for example "search_task_ref"
You have to take the "id" attribute of the html field, for example, "search_project_user_container" BUT you have to REMOVE "container" to get it works.

I have create a PR to improve the helptext
https://github.com/Dolibarr/dolibarr/pull/30216

PR #30173, opened 2024-06-27
Correct CI on Workstation API

aspangaro:


frederic34:
https://github.com/Dolibarr/dolibarr/pull/30169

PR #30179, opened 2024-06-27
FIX orders to bill menu

lvessiller-opendsi:
FIX orders to bill menu
- when you access from "orders to bill" menu, the filter "billed" is not set to 0.
- so the list contains billed orders

I keep the compatibility with the old url path (billed=0) in menu and I modified in "Eldy" menu for upper versions ("search_billed" is used in filter)


lvessiller-opendsi:
But there is a problem if we set default value "search_billed" to 0 or 1 for this order list page.
The GETPOSTISSET don't check the default value for the page.
So I decided only to keep the fix in url menu.

lvessiller-opendsi:
The same problem exists for supplier orders

PR #30182, opened 2024-06-27
NEW: add trigger for linking and unlinking invoices

altairis-noe:
# NEW add trigger for linking and unlinking invoices
Some external modules might benefit from being able to act upon discount being added to (or removed from) invoices


htdocs/core/class/discount.class.php
```diff
@@ -4,6 +4,7 @@
  * Copyright (C) 2024      Alexandre Janniaux   <alexandre.janniaux@gmail.com>
  * Copyright (C) 2024       Frédéric France             <frederic.france@free.fr>
  * Copyright (C) 2024		MDW							<mdeweerd@users.noreply.github.com>
+ * Copyright (C) 2024		Noé Cendrier		<noe.cendrier@altairis.fr>
```
altairis-noe, line 7: @eldy My email address is repelled by by pre-commit, and I don’t remember how we addressed this last the last time 😅

eldy:
I added "noe" int the exception of spellchecker

htdocs/core/class/discount.class.php
```diff
@@ -470,6 +473,14 @@ public function link_to_invoice($rowidline, $rowidinvoice)
 				$this->fk_facture_line = $rowidline;
 				$this->fk_facture = $rowidinvoice;
 			}
+			if (!$notrigger) {
+				// Call trigger
+				$result = $this->call_trigger('DISCOUNT_LINK_TO_INVOICE', $user);
```
eldy, line 478: Trigger codes must be limited to the 4 CRUD events. Here we modify the discount line, so you can use the code
DISCOUNT_MODIFY

altairis-noe:
> I added "noe" int the exception of spellchecker

Thanks!

htdocs/core/class/discount.class.php
altairis-noe: Thanks for the correction!

PR #30184, opened 2024-06-27
FIX: pos: invoice date incorrectly set beacause of timezome mismatches

marc-dll:
This reverts commit 36e91da that does not seem to work.

Tickets created from 10 PM French time (during DST) have their date set to the next day.

The server date and timezones are correctly set.

### DIFF
--- a/htdocs/takepos/invoice.php
+++ b/htdocs/takepos/invoice.php
@@ -473,7 +473,7 @@ if (empty($reshook)) {
 	// If we add a line and no invoice yet, we create the invoice
 	if (($action == "addline" || $action == "freezone") && $placeid == 0) {
 		$invoice->socid = getDolGlobalString($constforcompanyid);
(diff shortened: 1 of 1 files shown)


eldy:
@marc-dll 
Can you tell be if this works:

```
include_once DOL_DOCUMENT_ROOT.'/core/lib/date.lib.php';
$invoice->date = dol_get_first_hour(dol_now('tzuserrel'));		// Invoice::create() needs a date with no hours

```

marc-dll:
Hi @eldy ,

Unfortunately this doesn't work, from 10 PM to midnight, the invoice is dated to the next day.

eldy:
> Hi @eldy ,
> 
> Unfortunately this doesn't work, from 10 PM to midnight, the invoice is dated to the next day.

In your test, what was:
The timezone of user (web browser) 
The timezone of the php.

PR #30185, opened 2024-06-27
NEW Preview product list when we choose model layout Commerce

lamrani002:
# NEW|New functionality for display products on Ckeditor 
prof


htdocs/core/lib/process_template.lib.php
```diff
@@ -0,0 +1,55 @@
+<?php
```
eldy, line 1: This file seems to be a service and not a library.
So it must be moved into dir
htdocs/core/ajax/process_template.php

PR #30190, opened 2024-06-28
Fix: BOM columns if "Stock management supports Services" is active

### LINKED ISSUE #30187: BUG: Missing columns in BOM-Card if stockmanagement for services is active
### Bug

If the stockmanagement for services is active we can't configure the BOM correct:
image

### Dolibarr Version

19.0

### Environment PHP

_No response_

### Environment Database

_No response_

1. Activate Moduls: Services, Stock
2. Switch "Stock management supports Services" in the stock module to active.
image

3. Create a BOM => you can't set the Frozen Qty, Stock change disabled and Manufacturing efficiency (no columns visible).

### Attached files

_No response_

Humml87:
Fix #30187.
Now the Columns: Frozen Qty, Stock change disabled and Manufacturing efficiency shows if the option "Stock management supports Services" is active.
image

### DIFF
--- a/htdocs/bom/tpl/objectline_create.tpl.php
+++ b/htdocs/bom/tpl/objectline_create.tpl.php
@@ -47,11 +47,6 @@ if (empty($forceall)) {
 if (empty($filtertype)) {
 	$filtertype = 0;
 }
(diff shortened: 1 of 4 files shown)


eldy:
The feature was not completely implemented in v19. Finishing implementation must be done in develop branch.

PR #30192, opened 2024-06-28
NEW checkbox to choose thirdparty nature

Hystepik:
Change the way nature of thirdparty is chosen to be more easy for the user
Before:
image
After :
image


eldy:
Can you merge the 3 choices into the same td with colspan=3 so all fields will be together, seoarated with a space.
Also the text of label Prospect, Customer, Supplier must be in white and background use the same color than the color visible when in view mode in field "Nature of thirdparty"

PR #30200, opened 2024-06-30
Better comments

aspangaro:
Apply suggested corrections on comments


htdocs/langs/en_US/eventorganization.lang
```diff
@@ -50,6 +50,7 @@ DateOfRegistration = Date of registration
 ConferenceOrBoothAttendee = Conference Or Booth Attendee
 ApplicantOrVisitor=Applicant or visitor
 Speaker=Speaker
+EventRegistrationAreClosed=Event registration is closed
```
frederic34, line 53: it is present in develop but missing in 20.0
https://github.com/Dolibarr/dolibarr/pull/30198

PR #30211, opened 2024-06-30
NEW: customer code module gazel

rycks:
# NEW customer code system

when we do migration from other ERP to dolibarr we have always to do with old customer numbers and most of the time that customer numbering is bad but we have to do handle it because all of other systems (and most of the time paperworks) of the companie use that numbers.

For example customer code C01LK (created 20 years ago when it was from a DOS system) is used in paper archiving system and when they change for a new ERP in 1998 new customer have new numbers like CL199803442 but old customers keep C01LK/* numbers ... and in 2024 when they switch to dolibarr we have to do with history.

Then in that situation the only solution is to use "Leopard" numbering system but people have to manually choose customer numbers for new customers ... that's impossible and the reason why i purpose that "gazel" code number.

Gazel is Elephant but with a "from started date"  : all customer before that date is in mode "no verification, code is totaly open" and elephent code number is mandatory from that date so new customers would get automatically good numbers.

Is that clear enougth ?

Note: i don't know where to put translations strings, in english lang file ? other ?

### DIFF
--- a/htdocs/core/modules/societe/mod_codeclient_gazel.php
+++ b/htdocs/core/modules/societe/mod_codeclient_gazel.php
@@ -1,0 +1,367 @@
+<?php
+/* Copyright (C) 2004      Rodolphe Quiedeville <rodolphe@quiedeville.org>
+ * Copyright (C) 2006-2009 Laurent Destailleur  <eldy@users.sourceforge.net>
+ * Copyright (C) 2007-2012 Regis Houssin        <regis.houssin@inodbox.com>
+ * Copyright (C) 2011      Juanjo Menent	    <jmenent@2byte.es>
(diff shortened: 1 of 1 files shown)


aspangaro:
Interesting ! For the language keys, in my mind, put them in admin.lang

eldy:
Why duplicating code by creating a new numbering module.
A comprison with elephant show this add only a dozen of line (So only a 3% change in code and 97% are similar). Why not adding the new test into mod_codeclient_elephant.php file  ? (so if the new option is off, the code will remains the same)

rycks:
lol, because i was certain you will ask me "why do you modify elephant with a new option, what about history and people who chooses elephant blablablabla" ... ok let me close that pr and make an option to elephant

PR #30214, opened 2024-07-01
Fix filter bug, overdue invoices from home page

zahwyrix:
# FIX filter bug, overdue invoices from home page


htdocs/compta/facture/list.php
```diff
@@ -147,8 +147,8 @@
 $search_product_category = GETPOST('search_product_category', 'intcomma');
 $search_fac_rec_source_title = GETPOST("search_fac_rec_source_title", 'alpha');
 
-$search_late = GETPOST('search_late');
-if ($search_late == 'late') {
+$search_option = GETPOST('search_option');
```
eldy, line 150: Can you describe why changing name of var search_late into search_option will fix a bug ?

PR #30224, opened 2024-07-01
FIX: Error mesg show untranslated extrafield name

noec764:
Before
image

After
image

### DIFF
--- a/htdocs/core/class/extrafields.class.php
+++ b/htdocs/core/class/extrafields.class.php
@@ -833,7 +833,7 @@ class ExtraFields
 	public function fetch_name_optionals_label($elementtype, $forceload = false)
 	{
 		// phpcs:enable
-		global $conf;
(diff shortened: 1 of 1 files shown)


htdocs/core/class/extrafields.class.php
```diff
@@ -883,7 +883,7 @@ public function fetch_name_optionals_label($elementtype, $forceload = false)
 					}
 
 					$this->attributes[$tab->elementtype]['type'][$tab->name] = $tab->type;
-					$this->attributes[$tab->elementtype]['label'][$tab->name] = $tab->label;
+					$this->attributes[$tab->elementtype]['label'][$tab->name] = $langs->transnoentities($tab->label);
```
eldy, line 886: This denaturate the content on array  $this->attributes that should contains the database content.
Can't you make the translation later when we forge the error message ?

PR #30225, opened 2024-07-01
Additional values in website language file

Daoud-mohamed:
Additional values in website language file


eldy:
Can you try to move all those entries into the lang file of the marketplace module instead (file marketplace.lang) ?
You will just have to make  $weblang->load("marketplace@markeplace")  to load lang file;

Goal is to have trnaslation key used only into template of website provided by marketplace module inside the lang file of module too.

PR #30230, opened 2024-07-01

thomas-Ngr:

This PR:

SQL Part : #26190


eldy:
Change into language file  htdocs/langs/fr_FR/products.lang   is not allowed  (only en_US is allowed for language files)

htdocs/expedition/card.php
```diff
@@ -343,6 +344,30 @@
...
 			} else {
+				$p = new Product($db);
+				$res = $p->fetch($objectsrc->lines[$i]->fk_product);
+				if ($res > 0) {
+					if (GETPOST('entrepot_id', 'int') == -1) {
+						$qty .= '_'.$j;
+					}
+
+					if ($p->stockable_product == Product::DISABLED_STOCK) {
+						$w = new Entrepot($db);
```
eldy, line 355: If product is not stockable, we have no reason to manage some warehouses.

htdocs/expedition/card.php
```diff
@@ -343,6 +344,30 @@
...
+						$Tw = $w->list_array();
+						if (count($Tw) > 0) {
+							$w_Id = array_keys($Tw);
+							$stockLine[$i][$j]['qty'] = GETPOST($qty, 'int');
+
+							// lorsque que l'on a le stock désactivé sur un produit/service
```
eldy, line 361: English comment not allowed

htdocs/expedition/class/expedition.class.php
```diff
@@ -2829,6 +2833,13 @@ class ExpeditionLigne extends CommonObjectLine
 	public $volume;
 	public $volume_units;
 
+	/**
+	 * 0=This service or product is not managed in stock, 1=This service or product is managed in stock
+	 *
+	 * @var boolean
```
eldy, line 2839: @var int

htdocs/expedition/class/expedition.class.php
```diff
@@ -2829,6 +2833,13 @@ class ExpeditionLigne extends CommonObjectLine
...
+	 */
+	public $stockable_product = true;
```
eldy, line 2841: public $stockable_product = 1;

thomas-Ngr:
@eldy I guess all threads can be resolved, and the checks pass. Can you have another look at this PR ? Thanks :+1:

thomas-Ngr:
closing and restarting from branch Develop since file expedition.class.php has been split and there is no easy way to resolve conflicts. see #30996

PR #30240, opened 2024-07-02
ADD_HOOK_IN_SITUATION

IC-Mathieu:
NEW|New add hook "createFrom" in Situation
When you create Invoice situation from other object the hook "createFrom" is not executed

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -2048,6 +2048,47 @@ if (empty($reshook)) {
 					if ($ret > 0) {
 						$nextSituationInvoice->insertExtraFields();
 					}
+
+					// Hooks
+					$regs = array();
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/card.php
```diff
@@ -2048,6 +2048,47 @@
 					if ($ret > 0) {
 						$nextSituationInvoice->insertExtraFields();
 					}
+
+					// Hooks
+					$regs = array();
```
eldy, line 2053: this line until 2085 should be done inside the hook.
No reason to overload code with this when there is no hook...

IC-Mathieu: I used the same code as the other invoices (line 1907 for the hook and lines 1507 to 1562 for the creation of $srcobject)

PR #30244, opened 2024-07-02
NEW : use specific landing page to home menu entry

altairis-tof:
if specific landing page is specified, use it as dolibarr homepage menu entry link

### DIFF
--- a/htdocs/core/menus/standard/eldy.lib.php
+++ b/htdocs/core/menus/standard/eldy.lib.php
@@ -79,7 +79,7 @@ function print_eldy_menu($db, $atarget, $type_user, &$tabMenu, &$menu, $noout = 
 	if (!empty($landingpage)) {
 		$landingpage = dol_buildpath($landingpage, 1);
 	} else {
-		$landingpage = '/index.php?mainmenu=home&amp;leftmenu=home';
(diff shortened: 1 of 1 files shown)


htdocs/core/menus/standard/eldy.lib.php
```diff
@@ -75,9 +75,15 @@ function print_eldy_menu($db, $atarget, $type_user, &$tabMenu, &$menu, $noout =
 	$menu_arr = array();
 
 	// Home
+	$landingpage = (empty($user->conf->MAIN_LANDING_PAGE) ? (!getDolGlobalString('MAIN_LANDING_PAGE') ? '' : $conf->global->MAIN_LANDING_PAGE) : $user->conf->MAIN_LANDING_PAGE);
```
eldy, line 78: Use of conf->global->xxx or user->conf->xxx is noore allowed.

eldy (changes requested):

htdocs/core/menus/standard/eldy.lib.php
altairis-tof: sorry it was a copy/paste from main.inc.php :wink:

eldy: $user->conf-> is also not allowed.
use getDolUserString

PR #30251, opened 2024-07-03
fix #2810

### LINKED ISSUE #2810: Show warning sign in homescreen if dolibarr database version differs from actual dolibarr version
Because you cannot notice this until you access "Info Dolibarr" page, it would be _very_ nice to warn the user in the homescreen when the dolibarr database version differs from the version of the code.

captura de pantalla 2015-05-11 a las 12 30 43

hansemschnokeloch:
Fix #2810

Time to fix oldest open issue :)


htdocs/index.php
```diff
@@ -98,6 +98,16 @@
 
 print load_fiche_titre('&nbsp;', $resultboxes['selectboxlist'], '', 0, '', 'titleforhome');
 
+if (DOL_VERSION  != $conf->global->MAIN_VERSION_LAST_UPGRADE) {
```
eldy, line 101: This should not be required anymore.
We have this in current main.inc.php

```
// If an upgrade process is required, we call the install page.
$checkifupgraderequired = false;
if (getDolGlobalString('MAIN_VERSION_LAST_UPGRADE') && getDolGlobalString('MAIN_VERSION_LAST_UPGRADE') != DOL_VERSION) {
	$checkifupgraderequired = true;
}
if (!getDolGlobalString('MAIN_VERSION_LAST_UPGRADE') && getDolGlobalString('MAIN_VERSION_LAST_INSTALL') && getDolGlobalString('MAIN_VERSION_LAST_INSTALL') != DOL_VERSION) {
	$checkifupgraderequired = true;
}
if ($checkifupgraderequired) {...
```

and if upgrade is required, a redirect to the upgrade process is automatically done. No choice.

hansemschnokeloch: Yes, that doesn't seem necessary any more. The issue should have been closed a long time ago.

PR #30256, opened 2024-07-03
FIX: social charges accountancy: link with employee should be ignored

marc-dll:
Social charges payment bank url is both linked with the social charge itself and with the user.

The link with the user should be ignored as the payment of social charges is not made to the employee but to a public establishment.

This causes an imbalance as it makes two debit counterparts, one with the social charge type accountancy code and the other with the employee accountancy code.

This bug is still present in develop.

### DIFF
--- a/htdocs/accountancy/journal/bankjournal.php
+++ b/htdocs/accountancy/journal/bankjournal.php
@@ -314,7 +314,9 @@ if ($result) {
 			}
 			// Now loop on each link of record in bank (code similar to bankentries_list.php)
 			foreach ($links as $key => $val) {
-				if ($links[$key]['type'] == 'user' && !$is_sc) continue;
+				if ($links[$key]['type'] == 'user' && $is_sc) {
+					continue;
+				}
 				if (in_array($links[$key]['type'], array('sc', 'payment_sc', 'payment', 'payment_supplier', 'payment_vat', 'payment_expensereport', 'banktransfert', 'payment_donation', 'member', 'payment_loan', 'payment_salary', 'payment_various'))) {
 					// So we excluded 'company' and 'user' here. We want only payment lines
 


htdocs/accountancy/journal/bankjournal.php
```diff
@@ -314,7 +314,9 @@
 			}
 			// Now loop on each link of record in bank (code similar to bankentries_list.php)
 			foreach ($links as $key => $val) {
-				if ($links[$key]['type'] == 'user' && !$is_sc) continue;
+				if ($links[$key]['type'] == 'user' && $is_sc) {
```
eldy, line 317: Note: There is no screen to transfer social contribution in a different step than payment of social contribution like we have for invoice. So for social contribution, it seems only payment is transferred.

Pb is here:
image

When a social contribution is linked to a user, if user has an aux account, the aux account is used as a different line.

@aspangaro 
If we have a social contribution SC with account code SC, paid to user U with account code AUXU, paid in bank B with account code B what should we get in transfer ?

This ?
```
Account    Aux account      Deb   Cred
----------------------------------------------------
B                                                         X
SC               AUXU                X    
```

aspangaro: I confirm. For social contribution, only payment are transfered.

User was added to social contribution for statistic but it should have no influence on the generation of the accounting entry. Whether the user is defined or not, he/she should not proceed with the addition of a line containing a subsidiary accounting account.

@eldy  We'll always have:
`Account    Aux account         Deb   Cred
----------------------------------------------------
B                                                        X
SC                                           X`

For me PR is validated

Example:

Without correction

image

With correction

image

aspangaro (approved):
Fix ok for me

eldy:
I pushed a different way to fix. Becasue i think the change 
`if ($links[$key]['type'] == 'user' && !$is_sc) continue;`
into
`if ($links[$key]['type'] == 'user' && $is_sc) {`
is ok for social contribution but may introduce effect on non social contribution.

So i pushed a fix that has effect on social contribution only and also it keeps the info on link to user on screen.
Thanks to check...

PR #30257, opened 2024-07-03
FIX - Unknown Character on HTML

atm-lucasmantegari:
# FIX - *Unknown Character on HTML*

Character ZWSP (ZeroWidthSpace) which are unknown in PDFs, this is automatically replaced by '????' which is problematic.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -7717,6 +7717,9 @@ function dol_htmlwithnojs($stringtoencode, $nouseofiframesandbox = 0, $check = '
 				}
 			}
 
+			//Clear ZERO WIDTH NO-BREAK SPACE, ZERO WIDTH SPACE, ZERO WIDTH JOINER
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -7821,6 +7821,7 @@ function dol_htmlentitiesbr($stringtoencode, $nl2brmode = 0, $pagecodefrom = 'UT
 		if ($removelasteolbr) {
 			$newstring = preg_replace('/<br>$/i', '', $newstring); // Remove last <br> (remove only last one)
 		}
+		$newstring = preg_replace('/[\x{200B}-\x{200D}\x{FEFF}]/u', '', $newstring);
```
eldy, line 7824: This looks to be a range of several utf8 char.
Or is it just the zero width space char ?

atm-lucasmantegari: its for this char : 
\x{FEFF}    ZERO WIDTH NO-BREAK SPACE
\x{200B}    ZERO WIDTH SPACE
\x{200D}    ZERO WIDTH JOINER

PR #30259, opened 2024-07-03
fix: on edit line contrat, if failed at the end, stay on edit mode

FHenry:
If update failed for whatever error, it should stay on edit mode as it is done on first date check

As it is done in 
https://github.com/Dolibarr/dolibarr/blob/29dd8ade0164640289b69ef46d4b7e310c9289d3/htdocs/contrat/card.php#L697

### DIFF
--- a/htdocs/contrat/card.php
+++ b/htdocs/contrat/card.php
@@ -802,6 +802,8 @@ if (empty($reshook)) {
 			$result = $objectline->update($user);
 			if ($result < 0) {
 				$error++;
+				$action = 'editline';
+				$_GET['rowid'] = GETPOST('elrowid');
 				setEventMessages($objectline->error, $objectline->errors, 'errors');
 			}
 		}


htdocs/contrat/card.php
```diff
@@ -802,6 +802,8 @@
 			$result = $objectline->update($user);
 			if ($result < 0) {
 				$error++;
+				$action = 'editline';
+				$_GET['rowid'] = GETPOST('elrowid');
```
rycks, line 806: @FHenry : are you sure to set _GET php magic variable ? i will have a look at the other part of the code but that's strange ...

rycks: ok that's the same code as for example line 698 ...

rycks (approved):
ok for me

htdocs/contrat/card.php
FHenry: Yes exactly the same code on failure as in 
```
elseif ($action == 'updateline' && $user->hasRight('contrat', 'creer') && !GETPOST('cancel', 'alpha')) {
		$error = 0;
		$predef = '';

		if (!empty($date_start_update) && !empty($date_end_update) && $date_start_update > $date_end_update) {
			setEventMessages($langs->trans("Error").': '.$langs->trans("DateStartPlanned").' > '.$langs->trans("DateEndPlanned"), null, 'errors');
			$action = 'editline';
			$_GET['rowid'] = GETPOST('elrowid');
			$error++;
		}

```

I Think the $_GET['rowid'] affectation is to scroll to line and/or set hidden input
I've just use the same code in case of failed update

eldy: > ok that's the same code as for example line 698 ...

Yes, it's true that code is strange. It is a bad practice. But it can be accepted in maintenance branch because it is already done this way in samefile few lines before.
This bad coding behaviour has been fixed into the develop branch.

eldy:
@lvessiller-opendsi your point of view ?

lvessiller-opendsi:
> @lvessiller-opendsi your point of view ?
It's ok for me (it's a fix)

lvessiller-opendsi (approved):

FHenry:
@eldy as @lvessiller-opendsi is OK to merge this in 18.0, and PR in develop merged, could you are you ok to accept the PR ?

PR #30263, opened 2024-07-03
NEW  Thirdparty - Can upload a file with drag and drop

aspangaro:
image

### DIFF
--- a/htdocs/societe/card.php
+++ b/htdocs/societe/card.php
@@ -6,17 +6,17 @@
  * Copyright (C) 2005-2017  Regis Houssin           <regis.houssin@inodbox.com>
  * Copyright (C) 2008       Patrick Raguin          <patrick.raguin@auguria.net>
  * Copyright (C) 2010-2020  Juanjo Menent           <jmenent@2byte.es>
- * Copyright (C) 2011-2023  Alexandre Spangaro      <aspangaro@open-dsi.fr>
+ * Copyright (C) 2011-2024  Alexandre Spangaro      <alexandre@inovea-conseil.com>
  * Copyright (C) 2015       Jean-François Ferry     <jfefe@aternatik.fr>
  * Copyright (C) 2015       Marcos García           <marcosgdf@gmail.com>
  * Copyright (C) 2015       Raphaël Doursenaud      <rdoursenaud@gpcsolutions.fr>
- * Copyright (C) 2018       Nicolas ZABOURI	        <info@inovea-conseil.com>
- * Copyright (C) 2018       Ferran Marcet		    <fmarcet@2byte.es.com>
+ * Copyright (C) 2018       Nicolas ZABOURI         <info@inovea-conseil.com>
+ * Copyright (C) 2018       Ferran Marcet           <fmarcet@2byte.es.com>
  * Copyright (C) 2018-2022  Frédéric France         <frederic.france@netlogic.fr>
- * Copyright (C) 2022-2023  George Gkantinas	    <info@geowv.eu>
+ * Copyright (C) 2022-2023  George Gkantinas        <info@geowv.eu>
  * Copyright (C) 2023       Nick Fragoulis
  * Copyright (C) 2023       Alexandre Janniaux      <alexandre.janniaux@gmail.com>
- * Copyright (C) 2024		MDW							<mdeweerd@users.noreply.github.com>
+ * Copyright (C) 2024       MDW                     <mdeweerd@users.noreply.github.com>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


hregis:
@aspangaro how are you Alexandre ?

aspangaro:
Fine @hregis ans you ?

hregis:
@aspangaro 
I live what I have to live and when I come back to reality I realize that Dolibarr's code is getting better and better... and sometimes becoming nonsense! We accept crap code that creates... crap!
But hey... one person cannot judge the impact of a change in the code... ;-)
And as long as it's like that we'll have shitty code!
(my words are with respect) ;-)

hregis:
@aspangaro @eldy 
not enough time between versions, not enough time when a function has changed name, sometimes no time at all, and sometimes from one version to another a radical change that breaks the functioning of our modules.. .
No time to integrate changes for Multicompany between versions! I have them in my head, I have to add them, but no time! too much work in the meantime and when the freeze version arrives it's too late! ;-)

hregis:
@aspangaro @eldy 

6 months for a major release is too short ! again and again...

hregis:
@aspangaro @eldy 

especially to provide a shaky version after 6 months... 
you might as well take the time to provide a stable version! 
even if it means making intermediate versions!

hregis:
@aspangaro @eldy 

customers or others sick of the latest version want to update their Dolibarr... 
I tell them that they must at least wait for one or two releases...
a release would have to be stable! 
otherwise it's a pre-release version!

hregis:
@aspangaro @eldy 

there are too many codes that are not thoroughly tested and it creates side effects without us understanding why, and we have to search for hours to find where it comes from! It's tiring !

hregis:
@eldy 

we need more people who check and test the code before integrating it... which necessarily requires more time... which means one major version per year and not every 6 months... we are not not pizza sellers!
It’s business management!!! ;-)

hregis:
@eldy 

I don't know what world you live in, but take a step back and consider those who can't handle this very short period of time...
i love you ! ;-)

hregis:
@aspangaro @eldy 

a dolibarr member pays contributions, apart from the reduction on the purchase of a module, he also expects stability and professionalism on the application... bring this professionalism! that's the lesser of it !

sorry @aspangaro These words are directed towards Eldy, I took advantage of this opening to vent my anger! ;-)

i love you all ;-)

hregis:
@aspangaro @eldy 

eldy too ! i love you too ;-)

aspangaro:
It's clear that it's a never-ending battle: we're not done playing with one version until another comes out, and we have to support it through the various modules we can edit. You also need to have tested the version in question from top to bottom to detect undesirable effects and respond to customers who want more and more.

On the one hand, it's burdensome, but on the other, it's a big part of Dolibarr's dynamism.

Code quality has made great strides with all the CI tools, and in time this will enable us to concentrate on other points.
As @eldy explained at devcamp, we're almost moving straight to v21 for this v20, given the review that's been done.

After that, moving to a more complete version every year wouldn't be bad, but it would bring other problems and in the end, it would be a bigger version to stabilize, so it would take more time and resources, and external modules wouldn't be up to date at the time of release.

The difficulty you face is that multicompany is ultra-central to the deployment of Dolibarr. Whether or not you use your module is a matter for the early stages of a project, as it has a major impact on the management of a company, which is why you need to keep up with this crazy pace. After that, if you decided to slow down by adapting multicompany every two versions to give yourself time, it wouldn't be incomprehensible and people would probably follow you. There's also the much-discussed notion of LTS, which is every 4 versions (v14 / v18 ...). We're talking about ERP, not just software that works on a single subject.

Otherwise, you should try to integrate more multicompany functions directly into the core (and I'm only talking about this module).

Another solution is to work as a team! This allows you to divide up the work.

Take care of yourself ;)

hregis:
@aspangaro @eldy 

one version per year would allow more time to resolve bugs that have been hanging around for years and no longer have to tell customers in a hurry not to upgrade to a new version because it will inevitably be bugged... it's too short to take the time to add or improve the code! customers go with one version and when we tell them that they will have to move to another version because the additions will not be present, it exasperates them! So yes we can make a git branch for them... but hey! let's stop making two major releases per year!!! Where are the Release versions? alpha, beta.. but release!
a year for a version, damn it!!! not two versions per year!
I kiss your molar...

hregis:
@aspangaro @eldy 

une version par an permettrait d'avoir plus de temps pour résoudre des bugs qui trainent depuis des années et ne plus avoir à dire au clients pressés de ne pas passser sur une nouvelle version car forcément elle sera buggée... c'est trop court pour prendre le temps d'ajouter ou améliorer le code ! les clients ils partent sur une version et quand on leurs dit qu'il va falloir passer sur une autre version car les ajouts ne seront pas présent, ça les exaspèrent ! Alors ouii on peut faire une branche git pour eux... mais bon ! arrêtons de faire deux versions majeurs par an !!! Elles sont où les versions  Release ? alpha, beta.. mais release !
une année pour une version, bordel !!! pas deux versions par an !
je vous kiss la molaire...

PR #30271, opened 2024-07-04
replace old syntax, rename pdf models with same class name

frederic34:


htdocs/install/mysql/migration/20.0.0-21.0.0.sql
```diff
@@ -40,6 +40,10 @@ ALTER TABLE llx_projet_task ADD COLUMN billable smallint DEFAULT 1;
 
 ALTER TABLE llx_inventory DROP COLUMN datec;
 
+DELETE FROM llx_document_model WHERE nom='standard' AND type='expensereport';
```
eldy, line 43: Delete or update with the new name ?

frederic34: I tried first with rename, but on migration module reload create the new entry

PR #30272, opened 2024-07-04
Exit code must be negative then external scripts can detect an error

libremaster:
Hello,

I made a script but i can't detect an error with database connexion without a negative exit code.
Exit code should be negative.

Thank you.


htdocs/master.inc.php
```diff
@@ -161,7 +161,7 @@ function is_countable($c)
 			exit;
 		}
 		dol_print_error($db, "host=".$conf->db->host.", port=".$conf->db->port.", user=".$conf->db->user.", databasename=".$conf->db->name.", ".$db->error);
-		exit;
+		exit(-1);
```
eldy, line 164: With Dolibarr v20, negatives codes are now forbidden (not portable on some OS that convert negative coe into positive value).
Rule is return code 0 for no error and >= 1 when error.

So i switched your PR to exit(1)

PR #30273, opened 2024-07-04
NEW More option for elephant numbering rule

rycks:
All details are on https://github.com/Dolibarr/dolibarr/pull/30211

when we do migration from other ERP to dolibarr we have always to do with old customer numbers and most of the time that customer numbering is bad but we have to do handle it because all of other systems (and most of the time paperworks) of the companie use that numbers.

For example customer code C01LK (created 20 years ago when it was from a DOS system) is used in paper archiving system and when they change for a new ERP in 1998 new customer have new numbers like CL199803442 but old customers keep C01LK/* numbers ... and in 2024 when they switch to dolibarr we have to do with history.

Then in that situation the only solution is to use "Leopard" numbering system but people have to manually choose customer numbers for new customers ... that's impossible and the reason why i purpose that "gazel" code number.

So there is a new option to Elephant number : "choose from started date" : all customer before that date is in mode "no verification, code is totaly open" and elephant code number is mandatory from that date so new customers would get automatically good numbers.


htdocs/core/modules/societe/mod_codeclient_elephant.php
```diff
@@ -114,6 +117,33 @@ public function info($langs)
...
+			}
+		}
+		if (empty($dateinput)) {
+			$dateinput = dol_now();
+		}
+		$texte .= '<tr><td>'.$langs->trans("DateStartThatModel").' ('.$langs->trans("DateStartThatModelHelp").'):</td>';
+		$texte .= '<td class="" style="flex-direction: row;display: flex;">';
+		$texte .= '<input type="checkbox" onclick="let d=document.getElementById(\'elephantchoosedate\'); if(this.checked){d.style.cssText = \'display: block;\'}else{{d.style.cssText = \'display: none;\'}}" name="value4" value="1" style="float: left;"/>';
+		$texte .= '<div style="display: none;" id="elephantchoosedate" class="right">';
+		$texte .= $form->selectDate($dateinput, 'value3', 0, 0, 1, '', 1, 1, $disabled);
```
frederic34, line 142: `		$selectdisabled = ((!empty($mc->sharings['referent']) && $mc->sharings['referent'] != $conf->entity) ? 1 : 0);
		$texte .= $form->selectDate($dateinput, 'value3', 0, 0, 1, '', 1, 1, $selectdisabled);
`

rycks: @frederic34 ... i really can't see that your code could do the same thing as mine ?

my objective is:
- add a checkbox and if user check it, then display the input field, uncheck that checkbox, hide the text input field
- on submit we could get value if checkbox is checked or no and then delete / update value

that is an elegant way to handle situation where people want to "remove" data: with classic input field if the field is empty on POST dolibarr do not "delete" / "set to empty" value (i think, that was the situation last time i check)...

frederic34: $disabled returns a string but selectDate wait for an int

eldy: You probably want to do
$disabled ? 1 : 0

PR #30277, opened 2024-07-05
New jnotify for login errors

Hystepik:
Use jnotify to display error of login page if $conf->use_javascript_ajax is set to 1
Before:
image
After:
image


eldy:
@Hystepik 
Can you try to put the message in the center (but only for the login page) ?

PR #30298, opened 2024-07-05
NEW : add option to add external billing contact on the supplier_proposal pdf

samifilali:
# FIX
- add external contact with billing code in supplier_proposal pdf


htdocs/core/modules/supplier_proposal/doc/pdf_zenith.modules.php
```diff
@@ -1375,7 +1375,7 @@ protected function _pagehead(&$pdf, $object, $showaddress, $outputlangs)
 
 			// If CUSTOMER contact defined on proposal, we use it. Note: Even if this is a supplier object, the code for external contat that follow order is 'CUSTOMER'
 			$usecontact = false;
-			$arrayidcontact = $object->getIdContact('external', 'CUSTOMER');
+			$arrayidcontact = array_merge($object->getIdContact('external', 'CUSTOMER'), $object->getIdContact('external', 'BILLING'));
```
eldy, line 1378: In most common cases, the contact for billing must appears when billing only but not on the proposal (where the contact is the sale representative). So to not disturb this behaviour that works this way for other documents (like customer proposal), i recommend to add a hidden option
SUPPLIER_PROPOSAL_ADD_BILLING_CONTACT_ON_PDF
to enable the second contact on PDF.

Also, if you change in one template, you must change into the other template of the supplier_poposal.

Issue #30252, opened 2024-07-03
llx_ecm_files error when editing a product, changing only the case of the product ref


AlbanDurrheimer:
### Bug

Hello,  
When I edit the product references by only changing the case, it creates a `DB_ERROR_RECORD_ALREADY_EXISTS` error in the `llx_ecm_files` table with conflit on `uk_ecm_files` key.  
After some research, it appears that the issue is due to the fact that unique constraints are [case-insensitive by default in MySQL](https://stackoverflow.com/a/6448861).

### Dolibarr Version

19.0.2

### Environment PHP

8.2.7

### Environment Database

MariaDB 11.4.2

### Steps to reproduce the behavior and expected behavior

1. Open a product/service with attached files ;  
2. Edit the product reference by only changing the case (ex: `p00001` -> `P00001`) ;

### Attached files

![screenshot](https://github.com/Dolibarr/dolibarr/assets/21126730/852e33ec-882d-4eb4-9d97-d063b82dae55)

PR #30299, opened 2024-07-05
update ecmfile on product update

frederic34:


frederic34:
https://github.com/Dolibarr/dolibarr/issues/30252

htdocs/product/class/product.class.php
```diff
@@ -1454,6 +1454,9 @@ public function update($id, $user, $notrigger = 0, $action = 'update', $updatety
 								$langs->load("errors");
 								$this->error = $langs->trans('ErrorFailToRenameDir', $olddir, $newdir);
 								$error++;
+							} else {
+								require_once DOL_DOCUMENT_ROOT . '/ecm/class/ecmfiles.class.php';
```
eldy, line 1458: May be we should call ths only if
$this->oldcopy->ref is defined and different than $this->ref   ?

frederic34: image

PR #30300, opened 2024-07-05
Fix:Non tag special chars should be escaped before calling xml_parse_into_struct

bafbes:
Fix:Non tag special chars should be escaped before calling xml_parse_into_struct

### DIFF
--- a/htdocs/core/class/dolreceiptprinter.class.php
+++ b/htdocs/core/class/dolreceiptprinter.class.php
@@ -1,7 +1,8 @@
 <?php
 /* Copyright (C) 2015-2024  Frédéric France     <frederic.france@free.fr>
  * Copyright (C) 2020       Andreu Bisquerra    <jove@bisquerra.com>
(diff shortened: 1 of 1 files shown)


eldy:
Can you provide an example of string that generate the bug with old code ?

bafbes:
The chars '"&<> are rontrol characters for xml_parse_into_struct and must be escaped before to be used in this function. As an example xml_parse_into_struct($p,'<tag>young & young',$vals) produces 
$vals=Array
(
    [0] => Array
        (
            [tag] => TAG
            [type] => open
            [level] => 1
            [value] => young 
        )
)
Where the second part of the string is ignored while xml_parse_into_struct($p,'<tag>young &amp ; young',$vals) produces a correct response (The space is introduced to prevent html decoding)
$vals=Array
(
    [0] => Array
        (
            [tag] => TAG
            [type] => open
            [level] => 1
            [value] => young & young
        )
)

PR #30301, opened 2024-07-05
FIX #10789 Add a warning if we add a contact that has a user

### LINKED ISSUE #10789: External users cannot see projects
# Bug
External users cannot see or access projects when assigned as contact.

## Environment
- **Version**: [9.0.1]
- **PHP**: [7.0]

External users should see projects they are assigned as contact

- Create a contact
- Create an external user for this contact
- Create a project
- Assign contact on project

## Fix ?
It seems to be an error in htdocs/projet/class/project.class.php lines 1239 & 1247.
`$sql.= " AND ec.fk_socpeople = ".$user->id.")";`
Maybe use $user->contactid instead ?
fk_socpeople seems to be sometime a user id and sometime a contact id, this can lead to security issues.

The line 701 in `htdocs/projet/list.php` prevents the project list to be shown.

FlorentPoinsaut:
# FIX #10789 Add a warning if we add a contact that has a user


eldy:
Can you check the user of the external user is correctly assigned to the project ? If not, this is the trouble.

Current rule to manage permission on project is:
Your user is assigned to the project, you can see the project.
Your user is not, you can't (except if you have the permission see all project even if not assigned).

So permission is managed by the user and only the users. Trying to introduce permissions based on other element than users is dangerous. it may generates a lose of control of the security. All the security layer must be based on the user and only the user.
So current behaviour is the expected feature.

it is clearthat being able to assign a contact to a project may let think we grant permission.  But permission system is not handled by contacts, but by users and only users. So i recommend instead of this PR, to add a picto "warning, you must assign the user to allow access..." on the page to assign contact when we assign a contact (that is just an information) of an external user instead of assigning the user (that give the permission).

FlorentPoinsaut:
OK @eldy  I add a warning message :)

eldy:
@FlorentPoinsaut 
Branch needs to be updated (merge) with last develop to fix CTI errors. But i have no permission on your branches to do it (i can do it on other PR, don't know hy i can't with yours). Can you do it ?

FlorentPoinsaut:
It's done @eldy

PR #30312, opened 2024-07-08
fix phpstan expects int, null ...

frederic34:


frederic34:
@eldy conflicts solved

PR #30317, opened 2024-07-08
NEW Module to setup OpenID Connect module

MaximilienR-easya:
# New
OpenID Connect as a standard module


htdocs/user/logout.php
```diff
@@ -99,6 +99,15 @@
 	$url .= (preg_match('/\?/', $url) ? '&' : '?').'dol_use_jmobile=1';
 }
 
+// Logout openid_connect sessions using OIDC logout URL if defined
+if (getDolGlobalInt('MAIN_MODULE_OPENIDCONNECT', 0) > 0 && !empty($_SESSION['OPENID_CONNECT']) && !empty($conf->global->MAIN_AUTHENTICATION_OIDC_LOGOUT_URL)) {
```
eldy, line 103: !empty($conf->global->MAIN_AUTHENTICATION_OIDC_LOGOUT_URL)
is not allowed.
Can you replace with
getDolGlobalString('MAIN_AUTHENTICATION_OIDC_LOGOUT_URL') ?

htdocs/core/tpl/login.tpl.php
```diff
@@ -335,15 +358,19 @@
...
+	dol_include_once('/core/lib/openid_connect.lib.php');
 	$langs->load("users");
 
 	//if (!empty($conf->global->MAIN_OPENIDURL_PERUSER)) $url=
 	print '<div class="center" style="margin-top: 20px; margin-bottom: 10px">';
 	print '<div class="loginbuttonexternal">';
 
-	$state = hash('sha256', session_id());
-	$url = getDolGlobalString('MAIN_AUTHENTICATION_OPENID_URL').'&state='.$state;
+	if (empty($conf->global->MAIN_AUTHENTICATION_OPENID_URL)) {
```
eldy, line 369: must use
if (!getDolGlobalString('MAIN_AUTHENTICATION_OPENID_URL')

htdocs/main.inc.php
```diff
@@ -887,7 +887,7 @@ function analyseVarsForSqlAndScriptsInjection(&$var, $type, $stopcode = 1)
 		if (GETPOST("username", "alpha", $allowedmethodtopostusername)) {	// For posting the login form
 			$goontestloop = true;
 		}
-		if (GETPOST('openid_mode', 'alpha', 1)) {	// For openid_connect ?
+		if (GETPOST('openid_mode', 'alpha')) {	// For openid_connect ?
```
eldy, line 890: Can you confirm we must remove the parameter ",1" here ?

MaximilienR-easya: This is to allow POST, this is used on the form inside 'callback.php'

We want to prevent the argument to stay in the URL when a user doesn't change page after login. When a new OpenID authentication is needed and this parameter is still present in the URL this cause an error since it's awaiting the code parameter if this parameter is present and the user not logged.

MaximilienR-easya: @eldy

PR #30320, opened 2024-07-08
FIX #30274 : Add the include before executing dolibarr_set_const

### LINKED ISSUE #30274: before the ajax line the reference to the error function is missing from dolibarr 13 to current
https://github.com/Dolibarr/dolibarr/blob/ac136ec1aeee17edf8ab5f02e567fabdbb320600/htdocs/main.inc.php#L1814

current
	if (GETPOST('dol_resetcache')) {

			dolibarr_set_const($db, "MAIN_IHM_PARAMS_REV", getDolGlobalInt('MAIN_IHM_PARAMS_REV') + 1, 'chaine', 0, '', $conf->entity);
		}

correction
	if (GETPOST('dol_resetcache')) {
			include_once DOL_DOCUMENT_ROOT.'/core/lib/admin.lib.php';
			dolibarr_set_const($db, "MAIN_IHM_PARAMS_REV", getDolGlobalInt('MAIN_IHM_PARAMS_REV') + 1, 'chaine', 0, '', $conf->entity);
		}

ksar-ksar:
FIX #30274

### DIFF
--- a/htdocs/main.inc.php
+++ b/htdocs/main.inc.php
@@ -1582,6 +1582,7 @@ function top_htmlhead($head, $title = '', $disablejs = 0, $disablehead = 0, $arr
 		}
 		// Refresh value of MAIN_IHM_PARAMS_REV before forging the parameter line.
 		if (GETPOST('dol_resetcache')) {
+			include_once DOL_DOCUMENT_ROOT.'/core/lib/admin.lib.php';
 			dolibarr_set_const($db, "MAIN_IHM_PARAMS_REV", ((int) $conf->global->MAIN_IHM_PARAMS_REV) + 1, 'chaine', 0, '', $conf->entity);
 		}
 


rycks:
ok for me :)

eldy:
@rycks 

Do you have permission to set the Requesters ?
So when such a ticket appears, you can set yourself and lionel and me as requested approvers.

Better, if we can have a guthub action that do this automatically each time a pr is submitted, this will save time and allows us to follow who said yes and no.

And finally, we could have a github actuon that merge automatically the pr when all requested approvers have said ok.

This should be more efficient that sending each time a ping to the other manually...

rycks:
@eldy i only have Reviewers, Assignees, Labels, Projects, Milestone and Notifications like you can see here :

(sorry for the huge picture, i retry)

menu

eldy:
I was speaking about Reviewers. 
You have a wheel on the screenshot on right of Reviewers so i think you have permission to set yourself and lionel as reviewer.

rycks:
yes of course as reviewer but as that bug was not on 18.0 i was ... hmmm how could i say that, i was not sure to be "in my good job" for that, i realize it after sending "ok for me" ...

eldy:
Right.
I did the same error seeing the message "ok for me"...

mstoluca:
That error is constantly present. Since version 13, when I started working with Dolibarr (Unboxerp), and in every version. I have patched it locally, but since I noticed it hadn't been detected, I sent the message!

https://github.com/Dolibarr/dolibarr/issues/30274

PR #30327, opened 2024-07-09
WIP : Create packing.php file

dolibarr95:
Creation of the packing.php page to manage the packages (dimensions, weight, etc.) of a shipment.

- [ ] packing.php
- [ ] expedition_packing.class.php
- [ ] CRUD
- [ ] Translations
- [ ] Dictionary

### DIFF
--- a/htdocs/expedition/packing.php
+++ b/htdocs/expedition/packing.php
@@ -82,3 +82,19 @@ if ($user->socid) {
 	$socid = $user->socid;
 }
 $result = restrictedArea($user, 'expedition', $object->id, '');
+
+/*
+ * Actions
+ */
+
+/*
+ * View
+ */
+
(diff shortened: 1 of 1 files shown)


dolibarr95:
advice welcome 😂

eldy:
I recommend to describe what you plan to do (what will be screen content, where to you plan to store data, ...)

fappels:
@dolibarr95 This module https://github.com/fappels/dolibarr-shipmentpackage already does something like you need and is already based on the [llx_expedition_package](https://github.com/fappels/dolibarr-shipmentpackage/blob/develop/sql/llx_expedition_package.sql) table.
@eldy Do you think it's a module that can be added to core?

eldy:

PR #30331, opened 2024-07-09
FIX: Always re-instantiate object when doing mass PDF generation

creekorful:
Sometime when mass generating PDFs some strange bugs can happen because the object being rendered as PDF are not always "clean".

For example this happen in supplier invoice because in the method fetch (https://github.com/Dolibarr/dolibarr/blob/2f351ae4dc80c5e899333be26a633f5e1c67843f/htdocs/fourn/class/fournisseur.facture.class.php#L863C9-L863C24) we only reset "basic" properties and not the **thirdparty** property. Therefore if a module use the **thirdparty** magic property and does not reload it manually by calling **fetch_thirdparty** then we may end up with information from the first generated invoice where the thirdparty was initially loaded.

I think it would be cleaner to just reallocate the object each time before the PDF generation occurs to ensure up-to-date information. The impact caused by reallocating the object every-time in the loop would be less problematic than the current situation.

What do you think?

### DIFF
--- a/htdocs/core/actions_massactions.inc.php
+++ b/htdocs/core/actions_massactions.inc.php
@@ -1183,7 +1183,7 @@ if (!$error && $massaction == 'generate_doc' && $permissiontoread) {
 				$nbok++;
 			}
 
-            unset($objecttmp);
+			unset($objecttmp);
 		} else {
 			setEventMessages($objecttmp->error, $objecttmp->errors, 'errors');
 			$error++;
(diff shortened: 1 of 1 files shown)


eldy:
There is already a fetch(), so a database access is already done inside the loop. So doing a renew should have no effect on performance.
But may be on memory usage ? 
What happen on memory if you do   $a = new ... severaltimes.
Does the memory used by theprevious $a is reused ? Or memory of old $a is just in a garbage.
PHP is not clear with that. Do you know ?
@creekorful  In doubt, may be we can use unset($a) at end of the loop ?

creekorful:
@eldy that's the beauty with garbage collected language... We don't known when exactly. PHP will for sure perform some cleanup but it depends on many parameters. I am not sure whether this PR will have a considerable impact on resources but in any case I added an unset() so that it may help the GC reuse the memory location (not exactly sure because we would need to do some benchmark here).

Note that we can have simply reset the reference by setting the value to null, this will definitely help clean up memory faster but this will have an impact on the CPU cycles (and thus increase execution time). Not wanted here.

PR #30345, opened 2024-07-10
FIX: please compare strings with strings

altairis-noe:
# FIX please compare strings with strings
PHP 7 considers true `in_array("my string", array(0, 1, 2))`, because it gives a numeric value of 1 to a non empty string.
PHP 8 however, considers that "my string" != 1 (and I tend to agree).

This searches for the field value in an array which contains fields values and not numeric keys, it works better !


htdocs/imports/import.php
```diff
@@ -1716,7 +1716,7 @@
 		print ' &nbsp; <a href="'.$_SERVER["PHP_SELF"].'?step=5'.$param.'">'.$langs->trans("Modify").'</a>';
 	} else {
 		if (is_array($objimport->array_import_updatekeys[0]) && count($objimport->array_import_updatekeys[0])) {   //TODO dropdown UL is created inside nested SPANS
-			print $form->multiselectarray('updatekeys', $objimport->array_import_updatekeys[0], $updatekeys, 0, 0, '', 1, '80%');
+			print $form->multiselectarray('updatekeys', $objimport->array_import_updatekeys[0], $objimport->array_import_preselected_updatekeys[0], 0, 0, '', 1, '80%');
```
eldy, line 1719: I don't understand
$updatekeys is the array with the list of all preselected value of the combo box, so we should use this array as parameter for default value.
May be a screenhsot may help to understand ?

altairis-noe: $updatekeys does not list values, if empty in GETPOST it’s initiated on line 158/160 : 
```php
if (empty($updatekeys) && !empty($objimport->array_import_preselected_updatekeys[0])) {
	$updatekeys = array_keys($objimport->array_import_preselected_updatekeys[0]);
}
```
at this point `$objimport->array_import_preselected_updatekeys[0])` contains (0 => 'my_first_key', 1 => 'my_second_key', …)
so $updatekeys contains… numeric keys and not values. What we want to compare are values from $objimport->array_import_preselected_updatekeys[0]

I’m afraid a screenshot would only show that my preselected keys are preselected with php7 and not preselected with php8, and, with this modification, it works with both 🙂

eldy: I see.
So imagine we change the list of update keys and we submit the form, if there is an error on submit, can you confirm that the update keys is restired on the selection of user and not on the default value retreived from the moduke descriptor file ?

altairis-noe: I get your point. `$updatekeys` probably is the good var, but it has the wrong value. I’ll check where else it is used and amend the PR accordingly

altairis-noe:
@eldy, I tested it this way, and it seems to be correctly working, what do you think?

htdocs/imports/import.php
altairis-noe: After further testing, I have come to the conclusion that the error here was using `array_keys` in the first place : I removed it and it works like a charm

altairis-noe:
After further testing, I have come to the conclusion that the error here was using `array_keys` in the first place : I removed it and it works like a charm. What do you think, @eldy ?

PR #30346, opened 2024-07-10
fix: When update failed (extrafield check failed or 'date end' before 'date start' values inputed are lost

FHenry:
Following #30259


FHenry:
Additionnaly  a missing `?>`  php close tag before javascript...

PR #30348, opened 2024-07-11
20.0 better solution to minimize consequences on fetchAll function

rycks:
# QUAL other solution for fetchAll recent modification

## Situation

the new fetchAll function present in dolibarr 20 impose to all external modules to make changes like that to be compatible with dolibarr 20:

old syntax was

```
$SkillrecordsForActiveUser = $sk->fetchAll('ASC', 'fk_skill', 0, 0, array("customsql"=>"fk_object = ".$object->fk_user ." AND objecttype ='".SkillRank::SKILLRANK_TYPE_USER."'"), 'AND');
```

new syntax is

```
$SkillrecordsForActiveUser = $sk->fetchAll('ASC', 'fk_skill', 0, 0, "(fk_object:=:".((int) $object->fk_user).") AND (objecttype:=:'".$db->escape(SkillRank::SKILLRANK_TYPE_USER)."')", 'AND');
```

(note: i'm working on a documentation for developpers https://wiki.dolibarr.org/index.php?title=Dolibarr_20.0#fetchAll_and_sqlfilters)

but with that modification the module becomes incompatible with dolibarr version < 20 because in < 20 filter must be an array then developpers must do something like that

```
if (floatval(DOL_VERSION) < 20.0) {
  $filter = array("customsql"=>"fk_object = ".$object->fk_user ." AND objecttype ='".SkillRank::SKILLRANK_TYPE_USER."'");
} else {
  $filter = "(fk_object:=:".((int) $object->fk_user).") AND (objecttype:=:'".$db->escape(SkillRank::SKILLRANK_TYPE_USER)."')";
}
$SkillrecordsForActiveUser = $sk->fetchAll('ASC', 'fk_skill', 0, 0, $filter, 'AND');

```

## That is an other solution

After some hours spending on that problem, the solution included in current PR has a different approach, then external developpers have to do that to be compatible dolibarr < 20 AND >= 20

```
$filter = array(
    "customsql"=>"fk_object = ".$object->fk_user ." AND objecttype ='".SkillRank::SKILLRANK_TYPE_USER."'",
    "uss" => "(fk_object:=:".((int) $object->fk_user).") AND (objecttype:=:'".$db->escape(SkillRank::SKILLRANK_TYPE_USER)."')"
);
$SkillrecordsForActiveUser = $sk->fetchAll('ASC', 'fk_skill', 0, 0, $filter, 'AND');
```

Maybe you will find that's not better than the previous one ?

But that's some more thing to consider:

1. into dolibarr the module builder produce code and it makes fetchAll functions for the new module objects.

```
class MyObject extends CommonObject {
.../...
	public function fetchAll($sortorder = '', $sortfield = '', $limit = 1000, $offset = 0, string $filter = '', $filtermode = 'AND')
```

With your solution, objects issues from module builder < 20 will have internal fetchAll functions with array and new modules build with dolibarr >= 20 will have fetchAll with string.

Then for the same name function sometimes we will have strings, sometimes array. 

2. for human people around us who makes code into dolibarr modules

Yes i know there is a lot of surprises like that into dolibarr core but it could be an occasion to change and stop that sort of things. Why ? 

Because of human people like me ... when i read dolibarr sources i must do a more important cerebral effort than the other time for other projects, that's really not fun and contribute to make dolibarr hard for understand.

I'm ok to change french words for example but when we could avoid changing function signature i can't understand why we do not ?

### DIFF
--- a/htdocs/accountancy/class/bookkeeping.class.php
+++ b/htdocs/accountancy/class/bookkeeping.class.php
@@ -1098,6 +1098,8 @@ class BookKeeping extends CommonObject
 
 		// Manage filter
 		if (is_array($filter)) {	// deprecated, use $filter = USF syntax
+			dol_syslog(__METHOD__ . "Using deprecated filter with old array data, please update to Universal Search string syntax", LOG_NOTICE);
+
 			$sqlwhere = array();
 			if (count($filter) > 0) {
 				foreach ($filter as $key => $value) {
--- a/htdocs/accountancy/class/bookkeeping.class.php
+++ b/htdocs/accountancy/class/bookkeeping.class.php
@@ -1050,6 +1050,11 @@ class BookKeeping extends CommonObject
 		if (isset($filter['customsql'])) {
 			trigger_error(__CLASS__ .'::'.__FUNCTION__.' customsql in filter is now forbidden, please use $filter["uss"]="xx:yy:zz" with Universal Search String instead', E_USER_ERROR);
 		}
+		if (isset($filter['uss'])) {
+			$filter = $filter['uss'];
+		} else {
+			$filter = '';
+		}
 
 
 		dol_syslog(__METHOD__, LOG_DEBUG);
--- a/htdocs/accountancy/class/bookkeeping.class.php
+++ b/htdocs/accountancy/class/bookkeeping.class.php
@@ -1038,7 +1038,8 @@ class BookKeeping extends CommonObject
 	 * @param string 		$sortfield                      Sort field
 	 * @param int 			$limit                          Limit
 	 * @param int 			$offset                         Offset limit
-	 * @param string|array 	$filter                         Filter array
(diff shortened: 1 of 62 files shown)


htdocs/accountancy/class/bookkeeping.class.php
```diff
@@ -1038,15 +1038,26 @@ public function fetchAllByAccount($sortorder = '', $sortfield = '', $limit = 0,
 	 * @param string 		$sortfield                      Sort field
 	 * @param int 			$limit                          Limit
 	 * @param int 			$offset                         Offset limit
-	 * @param string|array 	$filter                         Filter array
+	 * @param array			$filter       					Filter as an Universal Search string.
```
eldy, line 1041: To make the transition with $filter that is a string, we must introduce the string here.
array will be abandonned after several versions of transitions.

PR #30361, opened 2024-07-12
New input to stock blacklist on ai prompt

lamrani002:
# NEW|New input for stock blacklist in ai module


eldy:
I made modification in develop branch so we can add and edit the blacklist the same way we edit the post and pre prompt.

PR #30367, opened 2024-07-13
Increase precision in calculation of totals

leninrivas:
# NEW|New [*Increase precision in calculation of totals*]
[*The current precision is currently limited when there are many lines with many quantities of items and many lines.
By having more decimals in the totals per line (for example 3 decimals or more) and separately controlling the decimal of the total of the parent object (for example 2 decimals). The rounding will be more precise on the total of the parent object.*]


eldy:
Doing this will generates big troubles in your accountancy.
Imagine the law of your country says you need 2 decimals in your accountancy, you must have both 2 decimals on the total of object but also on the total of lines, because accountancy is done per lines, not per object.
If law is you need 3 decimals, then you must have 3 decimals on total of lines. 
And a total of object must always be the total of lines, so if accuracy on lines is 2, total for object will be 2 and if accuracy expected is 3 decimals, then total per object should be 3 too.

If you have an accuracy of 3 on total of lines and only 2 on total of object, it means you make a truncation when showing the total of object loosing information. The total of object must always be the sum of total on lines and accuracy must be the one expected by the accounting rules of your country. Having differences means you will get different value when doing a sum on lines that what you have when reading value on object. And of course this is not allowed in an accountancy point of view.

leninrivas:
> Doing this will generates big troubles in your accountancy. Imagine the law of your country says you need 2 decimals in your accountancy, you must have both 2 decimals on the total of object but also on the total of lines, because accountancy is done per lines, not per object. If law is you need 3 decimals, then you must have 3 decimals on total of lines. And a total of object must always be the total of lines, so if accuracy on lines is 2, total for object will be 2 and if accuracy expected is 3 decimals, then total per object should be 3 too.
> 
> If you have an accuracy of 3 on total of lines and only 2 on total of object, it means you make a truncation when showing the total of object loosing information. The total of object must always be the sum of total on lines and accuracy must be the one expected by the accounting rules of your country. Having differences means you will get different value when doing a sum on lines that what you have when reading value on object. And of course this is not allowed in an accountancy point of view.

Hello.
But the idea is to improve the precision of the object. In the case of some countries, the decimals of lines are 10 and that of the object is 2. In the end it will be the administrator's decision if the total of line and object is the same depending on the case.
For example, in this case, if you have a total of 2 lines, the total of the object is also 2, it is not as precise as one requires.

Price this lines:
64.05
36.60
27.45
9.15
54.90
9.15
36.60
36.60
36.60
54.90
45.75
45.75
45.75
45.75
64.05
18.30

Case decimales totals: 2
vat lines:
11.53
6.59
4.94
1.65
9.88
1.65
6.59
6.59
6.59
9.88
8.24
8.24
8.24
8.24
11.53
3.29
Total vat: 113.67

Case decimales totals: 3
vat lines:
11.529
6.588
4.941
1.647
9.882
1.647
6.588
6.588
6.588
9.882
8.235
8.235
8.235
8.235
11.529
3.294
Total vat: 113.643
With rount total parent in 2: 113.64 <-- OK

leninrivas:
> > Doing this will generates big troubles in your accountancy. Imagine the law of your country says you need 2 decimals in your accountancy, you must have both 2 decimals on the total of object but also on the total of lines, because accountancy is done per lines, not per object. If law is you need 3 decimals, then you must have 3 decimals on total of lines. And a total of object must always be the total of lines, so if accuracy on lines is 2, total for object will be 2 and if accuracy expected is 3 decimals, then total per object should be 3 too.
> > If you have an accuracy of 3 on total of lines and only 2 on total of object, it means you make a truncation when showing the total of object loosing information. The total of object must always be the sum of total on lines and accuracy must be the one expected by the accounting rules of your country. Having differences means you will get different value when doing a sum on lines that what you have when reading value on object. And of course this is not allowed in an accountancy point of view.
> 
> Hello. But the idea is to improve the precision of the object. In the case of some countries, the decimals of lines are 10 and that of the object is 2. In the end it will be the administrator's decision if the total of line and object is the same depending on the case. For example, in this case, if you have a total of 2 lines, the total of the object is also 2, it is not as precise as one requires.
> 
> Price this lines: 64.05 36.60 27.45 9.15 54.90 9.15 36.60 36.60 36.60 54.90 45.75 45.75 45.75 45.75 64.05 18.30
> 
> Case decimales totals: 2 vat lines: 11.53 6.59 4.94 1.65 9.88 1.65 6.59 6.59 6.59 9.88 8.24 8.24 8.24 8.24 11.53 3.29 Total vat: 113.67
> 
> Case decimales totals: 3 vat lines: 11.529 6.588 4.941 1.647 9.882 1.647 6.588 6.588 6.588 9.882 8.235 8.235 8.235 8.235 11.529 3.294 Total vat: 113.643 With rount total parent in 2: 113.64 <-- OK

Hi.
So, in this case, to improve precision, would the number of decimal places for unit prices need to be increased from 8 to 10 or more?

PR #30369, opened 2024-07-13
CLOSE #30347 Add functionality to select parts via lot and serial number in shipment creation

### LINKED ISSUE #30347: Allow for selection via serial numbers in delivery
### Feature Request

When creating a delivery from a sales order after validating the delivery will automatically select N-amount of serial numbers as defined in the sales order. Sometimes there is a much bigger serial number pool than needed and it is very annoying to use the search function of the browser to find all the serial numbers you picked out for an order and manually updating the "Qty to ship".

It would be beneficial to have the possibility to use a scanner (as we have barcodes with the S/N) to input all the S/N which are part of that delivery.

### Use case

At our company we have much higher stocks than the volume of orders. Therefore it happens that the selected S/N in the delivery view is not what we picked out from our warehouse.

### Suggested implementation

1. text input to manually - or with a scanner - write the picked S/N
2. "submit" button
3. JS Script which catches the button press (or a newline character from a scanner) and automatically increments the Qty to ship for a S/N
4. reset the text input to empty and focus on it again for a seamless next scanning action

Idea for UI:
image

### Suggested steps

- add buttons and text input
- write JS Script

DaBootO:
This commit adds the functionality to increment the qty to ship when creating a shipment via input of a lot or serial number. This option has to be activated in the shipments module. It works by enclosing the batch detail in a span with the id equal to the batch number. The JS script uses JQuery to relatively select the input box for "qty to ship" and increments the value.


eldy:
Can you provide a screenshot of the full page after the change that show where is the combo ?

DaBootO:
image

This shows where the input field and button are located after activating this feature in the shipping module.

PR #30374, opened 2024-07-14
Fix: position of select key is 2 not 4

bafbes:
Fix: position of select key is 2 not 4


eldy:

bafbes:
The bug is easy to understand in the code. We test for the existence of the fourth element in order to take the value from the 2nd element.
In my case, it generated an error with the following object field element generated by modulebuilder :"type"=>"sellist:c_paiement:libelle:id". The bug does not allow using "id" instead of "rowid" when selecting an element from the llx_c_paiement table.

PR #30398, opened 2024-07-16
FIX #30397: for direct debit payments, BIC is optional

### LINKED ISSUE #30397: SWIFT/BIC should be optional for direct debit payments according to regulations
### Bug

**Background:**
After importing customer payment details from another system, no direct debit payments could be generated, because Dolibarr expects not only the IBAN but also the SWIFT/BIC code as mandatory information, but the BIC data was not present.

**Regulatory Status**
There has been a change in the SEPA direct debit regulations from February 2016(!), which made the BIC optional for direct debit transactions within the EU (Switzerland, Monaco and San Marino still require a BIC code), see: https://www.sparkasse.de/pk/ratgeber/finanzplanung/banking-tipps/bic.html (in German).

**Fix:**
Will add a hidden variable WITHDRAWAL_WITHOUT_BIC which, if set to 1, makes the BIC optional (if BIC is empty, this is ok, if any data is entered for BIC, the existing structural check will still be applied).

priojk:
According to SWIFT regulations for direct debit payments, IBAN is sufficient information and the BIC code is optional. But Dolibarr is rejecting the creation of direct debit payments without BIC code (see Issue #30397).

The code adjusts the behaviour of Dolibarr and allows the creation of direct debit payment xml-files based only on the customer IBAN (for the time being, only if the variable WITHDRAWALS_WITHOUT_BIC is set, in order not to interfere with existing installations).


htdocs/societe/paymentmodes.php
```diff
@@ -136,11 +136,11 @@
 
 	if ($action == 'update') {
 		// Update the bank account
-		if (!GETPOST('label', 'alpha') || !GETPOST('bank', 'alpha')) {
+		if (!GETPOST('label', 'alpha') || !(GETPOST('bank', 'alpha') or (getDolGlobalInt('WITHDRAWAL_WITHOUT_BIC')!=0))) {
```
eldy, line 139: use || instead of "or"

htdocs/societe/paymentmodes.php
```diff
@@ -2178,7 +2179,7 @@
 			$name = 'bic';
 			$size = 12;
 			$content = $companybankaccount->bic;
-			if ($companybankaccount->needIBAN()) {
+			if ($companybankaccount->needIBAN() and (getDolGlobalInt('WITHDRAWAL_WITHOUT_BIC')==0)) {
```
eldy, line 2182: use && instead of "and"

PR #30399, opened 2024-07-16
FIX backport Conflict with autoload to 18.0

fcharlaix-opendsi:
Backport the fix of Restler autoload conflict to the 18.0
[original commit](https://github.com/Dolibarr/dolibarr/commit/0dcc75318bda2466f83da360d02062a3702a3b15)

## The problem
When using the API, if a module use composer is invoked, this will enter an infinite loop in the `AutoLoader.php`.
This can be a module API including its composer or a hook of a module using composer.

### Module API example
1. Add a module that's use composer inside it's API file
2. Query the explorer
3. Wait for the API JSON fetch to fail

### Hook example
1. Add a module that's use composer and add a hook used by an API path (example FacturX)
4. Query an API path that's invoke the hook at some point (with FacturX its `POST /invoices/{id}/payment`)
5. Wait for PHP to run out of memory or timeout (if Xdebug is enabled it will stop and say it's an infinite loop)

## Remaining questions
I got some environments without the infinite loop, I wasn't able to find why

### DIFF
--- a/htdocs/includes/restler/framework/Luracast/Restler/AutoLoader.php
+++ b/htdocs/includes/restler/framework/Luracast/Restler/AutoLoader.php
@@ -268,8 +268,8 @@ namespace Luracast\Restler {
 				return $this->loadThisLoader($className, $loader);
 			}
 			foreach ($loaders as $loader)
-			if (false !== $file = $this->loadThisLoader($className, $loader))
-				return $file;
+				if (false !== $file = $this->loadThisLoader($className, $loader))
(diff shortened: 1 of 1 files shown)


eldy:
> ## Remaining questions
> I got some environments without the infinite loop, I wasn't able to find why

Yes i got the infinite loop too when using code using autoload inside API (managed by autoload too).
Don't know why also, but by commenting the commented code it worked for my test case. It seems for you too ;-)

PR #30404, opened 2024-07-16
Fix #30298 - billing contact in supplier proposal

samifilali:
sorry, I just realized that I reversed the condition : 			
`if (getDolGlobalInt('SUPPLIER_PROPOSAL_ADD_BILLING_CONTACT'))`

it must be
`if (!getDolGlobalInt('SUPPLIER_PROPOSAL_ADD_BILLING_CONTACT'))`


samifilali:
@eldy

PR #30412, opened 2024-07-17
New functionality for create and delete socialnetwordFait divers

lamrani002:
# NEW|New functionality for module social networks


htdocs/admin/dict.php
```diff
@@ -1261,6 +1262,10 @@
 
 llxHeader('', $title, '', '', 0, 0, '', '', '', 'mod-admin page-dict');
 
+$head = socialnetwork_prepare_head();
+
+print dol_get_fiche_head($head, 'dict', $langs->trans('MenuDict'), -1, 'user');
```
eldy, line 1267: If you add a dol_get_fiche_head, you must also add a dol_get_fiche_end().

Also adding the tab must be introduce only when you open the page from the page of the module social network.
When this page is open from menu Home - dictionnary, we must keep only the simple table view with no tabs.

htdocs/admin/faitdivers.php
```diff
@@ -0,0 +1,290 @@
+<?php
```
eldy, line 1: faitdiverse  must be renamed into fediverse:  https://fr.wikipedia.org/wiki/Fediverse

htdocs/admin/dict.php
lamrani002: ok i see i will fix it

eldy:
Duplicate

PR #30417, opened 2024-07-19
New select thirdparty filter by categ

atm-lena:
# NEW|New 
Add filter by category in select_thirdparty_list() function


htdocs/core/class/html.form.class.php
```diff
@@ -1553,6 +1553,12 @@ public function select_thirdparty_list($selected = '', $htmlname = 'socid', $fil
 		if (!$user->hasRight('societe', 'client', 'voir')) {
 			$sql .= ", " . $this->db->prefix() . "societe_commerciaux as sc";
 		}
+		if (isModEnabled('category')) {
+			$sql .= " LEFT JOIN " . $this->db->prefix() . "categorie_societe as cats ON cats.fk_soc = s.rowid";
```
eldy, line 1557: Filter on category must now no more use a left join.
Filter must be implemented by adding code in where and only in WHERE.
You can take a look how the filter is implemented on project list into project/list.php, you should use AND EXISTS or NOT EXISTS depending if you need a filter on a category or "no link to category".

atm-lena: I get it ! 
I have to use hook ? Or can I suggest a new param in function select_thirdparty_list() ? But it's strange to add a parameter dependent on the activation of another module... What do you think ?

eldy: No, ne need of hook.
Just a different writing of the sql to use NOT EXISTS in the where instead of JOIN LEFT...

eldy:
Seems a need for an external module.
Completing the sql for external module could be done withhook selectThirdpartyListWhere() so i close this ticket.

PR #30423, opened 2024-07-19
Fix line_order when we have children with children

MaximilienR-easya:
# Fix

Fix case when we have children with children that are not taken into account and cause a problem with the rank of the lines


eldy:
What was the trouble ? Can you provide an example ?

MaximilienR-easya:
In a case that we have a product (Test0) who have a child (Test00) who also have child (test01, test02, test03, test04).
Screenshot from 2024-07-22 10-34-19

Without the Fix we obtain that
Screenshot from 2024-07-22 10-33-44
But if we check children recursively, we obtain that

Screenshot from 2024-07-22 10-35-45

If we don't check the children recursively, we can have case like this one where everything is mixed up and not ordered

MaximilienR-easya:
@eldy do you need more exemple ?

MaximilienR-easya:
@eldy is there a problem ?

MaximilienR-easya:
Was finally merged here #31314

PR #30428, opened 2024-07-20
(really) Fix #20527 Accountancy Unbalanced entry proposed when an employee are declared on social contribution

### LINKED ISSUE #20527: Social charge with user related create unbalanced entry in accounting breakdown
### Bug

**Description**
A paid social charge with user related create unbalanced entry in accounting breakdown.
3 accounts are modified instead of 2, and the resulting entry is balanced.
See bellow for details.

**Bypass**
I 
- deleted the bank entry corresponding to the charge
- marked the charge unpayed again
- unlink the user by modifying the charge
- register the payment again
This second time, that is without user related, the accounting breakdown shown a balanced entry, as expected.

### Environment Version

14.0.4

### Environment OS

Linux 5.4.0-88-generic #99-Ubuntu SMP Thu Sep 23 17:29:00 UTC 2021 x86_64

### Environment Web server

Apache

### Environment PHP

 7.4.3

### Environment Database

MySQL or MariaDB 5.5.5-10.3.34-MariaDB-0ubuntu0.20.04.1

### Environment URL(s)

https://mydolibarr.com/htdocs/accountancy/journal/bankjournal.php?mainmenu=accountancy&leftmenu=accountancy_journal&id_journal=3

**Actual behaviour**
When a social charge is payed, with a user related, the bank inscription relate to user and charge (as can be expected).
However, when coming to accounting breakdown ("ventilation" in French), things goes wrong: the line is accounted twice, one from the social charge account and one from user account. 
Then, in the Journal the line is : 
> Desc                            | Account                                                                      | Debit | Credit
> BankId - Social Charge | Bank (1010 in CH accounting plan)                             | 0       | 100
> BankId - Social Charge | Social charge debt (2270 in CH accounting plan)         | 100   | 0
> BankId - Social Charge | User personnal account (2000 in CH accounting plan) | 100   | 0
Then, the line in unbalanced and when trying to write it in the "Grand Livre", an error is raised indicated that this is probably a bug. (And it is)

**Expected behaviour**
Even when indicating a user linked to a social tax, the accounting entry should stay balanced, that is no entry should be writen on user account.

**Steps to reproduce**
- Create a tax/social charge, and link a user to it
- Mark the social charge as "Paid", by registring the payment
- Go to Accounting > Accounting Breakdown (_Ventilation_ in French) and select the right period to make the charge appears (Bank Journal).
- If bug reproduce, you should have un unbalanced entry. Try to save and rigth it into the accounting ledger (_Grand Livre_) and the error message should appear.

### Attached files

_No response_

t3cneo:
# FIX #20527 

Merged PR #21193 doesn't fix the issue at least in latest v19.0.2

### DIFF
--- a/htdocs/accountancy/journal/bankjournal.php
+++ b/htdocs/accountancy/journal/bankjournal.php
@@ -358,7 +358,8 @@ if ($result) {
 
 			// Now loop on each link of record in bank (code similar to bankentries_list.php)
 			foreach ($links as $key => $val) {
-				if ($links[$key]['type'] == 'user' && !$is_sc && !$is_salary && !$is_expensereport) {
+				// Skip user link when processing social charge to avoid unbalanced entry
+				if ($links[$key]['type'] == 'user' && $is_sc) {
 					// We must avoid as much as possible this "continue". If we want to jump to next loop, it means we don't want to process
 					// the case the link is user (often because managed by hard coded code into another link), and we must avoid this.
 					continue;


eldy:
It seems another fix was applied with this patch  ca39809658f116d974b76436761e38060c68f731 in v18 (and merged into upstream).
Can you describe how to reproduce the bug if you still have it in develop despit this patch ?

t3cneo:
I'm experiencing the issue described in #20527 in v 19.0.2 : A social charge attached to a user will result in unbalanced entry.

To my (very little and possibly wrong) understanding, merged PR #21193 was "flawed" : the patch adds a test to detect a social charge and intend to skip user related journal entry but it doesn't because `!$is_sc` is used instead of `$is_sc`

I guess that is why [someone was surprised](https://github.com/Dolibarr/dolibarr/issues/20527#issuecomment-1625037929) the issue was closed back then.

This has been kept since and even more tests have been added from [line 339 ](https://github.com/Dolibarr/dolibarr/blob/32015628a9c769d50f17050edb44f5666ec37010/htdocs/accountancy/journal/bankjournal.php#L339) to [line 356](https://github.com/Dolibarr/dolibarr/blob/32015628a9c769d50f17050edb44f5666ec37010/htdocs/accountancy/journal/bankjournal.php#L356) making the condition in [line 361](https://github.com/Dolibarr/dolibarr/blob/32015628a9c769d50f17050edb44f5666ec37010/htdocs/accountancy/journal/bankjournal.php#L361C43-L361C50) impossible to meet.

eldy:
When line is a user line, when for social contribution, expense report or salary payment, the line is supported by the rest of code, thus must din't need to ignore line type user. Management is not implemented for other type this is why there is a continue for any other cases.
Also after fixes merged, it was no more possible to reproduce the bug (meaning bug was fixed).
If there is still a use case known to reproduce the bug, this use case must be described. Currently, with all merged last fix, i did not succeed in reproducing the bug.

t3cneo:
> When line is a user line, when for social contribution, expense report or salary payment

This is not what I understand from [line 361](https://github.com/Dolibarr/dolibarr/blob/32015628a9c769d50f17050edb44f5666ec37010/htdocs/accountancy/journal/bankjournal.php#L361C43-L361C50)

`if ($links[$key]['type'] == 'user' && !$is_sc && !$is_salary && !$is_expensereport) {`

I understand when line is a user line AND when NOT for social contribution AND NOT expense report AND NOT salary payment which is basically never...

Please correct me if I'm wrong, I'm just a user not a dev.





Anyway, when I create a social charge and attach it to a user :

Copie d'écran_20240721_113717

Here is what I get in accountancy :

Copie d'écran_20240721_113935


This PR fixes this.

It removes expense report and salary payment as they are correctly handled.

EDIT :

Oh sorry I just tried with latest patch from develop and it is indeed fixed by [ca39809](https://github.com/Dolibarr/dolibarr/commit/ca39809658f116d974b76436761e38060c68f731) 
I therefore think [line 361](https://github.com/Dolibarr/dolibarr/blob/1bb6be158b837dd0b76e3b786207ccc170f1b98f/htdocs/accountancy/journal/bankjournal.php#L361) to [line 365](https://github.com/Dolibarr/dolibarr/blob/1bb6be158b837dd0b76e3b786207ccc170f1b98f/htdocs/accountancy/journal/bankjournal.php#L365) can be removed as it is useless.

eldy:
i will keep line 361 to 365 for few version waiting to be sure everything is correct and stable before trying to remove it to avoid to increase risk of regression...
Thanks for you feedback on the very latest version so i can close the ticket.

PR #30436, opened 2024-07-23
fix imports

altairis-tof:
keyfield may content any field defined in import_fieldshidden_array so the present code dont do updates !

### DIFF
--- a/htdocs/core/modules/import/import_csv.modules.php
+++ b/htdocs/core/modules/import/import_csv.modules.php
@@ -929,9 +929,9 @@ class ImportCsv extends ModeleImports
 								// Note: For extrafield tablename, we have in importfieldshidden_array an enty 'extra.fk_object'=>'lastrowid-tableparent' so $keyfield is 'fk_object'
 								$sqlSelect = "SELECT rowid FROM ".$tablename;
 
-								if (empty($keyfield)) {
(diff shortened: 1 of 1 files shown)


rycks:
@altairis-tof i need more informations too ... 

for information in 20.0 you seems to made a different fix : https://github.com/Dolibarr/dolibarr/pull/29553 no ?

and in case of fix for that bug, i think xls import will be concerned too

@eldy could a backport of 20.0 fix be a better solution in same situation ?

eldy:
> @altairis-tof i need more informations too ...
> 
> for information in 20.0 you seems to made a different fix : #29553 no ?
> 
> and in case of fix for that bug, i think xls import will be concerned too
> 
> @eldy could a backport of 20.0 fix be a better solution in same situation ?

@rycks Yes, if a bug is already fixed in a higher version, a fix for a backport should use the same way to fix than the fix in higher version. It is always a better solution, as long as it is possible. This may not be possible if backport/cherry-pick is not possible because the fix in recent version is based on other changes done by other PR available only into the recent version.

rycks:
@altairis-tof what do you think about that ? different solution with https://github.com/Dolibarr/dolibarr/pull/29553 looks strange and make me trouble ...

eldy:
I close the ticket. It seems a fix was done differently in develop so backport shoud fix the same way.

PR #30437, opened 2024-07-23
FIX: manage to find elements linked for elements belonging to modules

thomas-Ngr:
# FIX "linked to object block" : manage to find elements linked for elements belonging to modules

Problem : I have linked an invoice to a 'session' object (specific to my module).
But the invoices do not appear on the 'session' card on linked to objects block.

before PR : 
image

after PR : 
image

explanation : 

- CommonObject::add_object_linked() sets the targettype field to `<module>_<element>`
- but CommonObject::fetchObjectLinked() and CommonObject::deleteObjectLinked() search tarbettype only by `<element>`

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -3682,11 +3682,7 @@ abstract class CommonObject
 			$origin = 'order_supplier';
 		}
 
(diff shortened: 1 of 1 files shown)


eldy:
Do you have the same trouble with v20 branch ?

PR #30441, opened 2024-07-23
FIX project: contact type for project creator

Slordef:
# FIX project: ability to select contact type for project creator

Currently, when we create a project, the user who creates it is automatically assigned as a contact with type "PROJECTLEADER".

In this PR, I make sure that the user can choose their type with a default selection of "PROJECTLEADER"


eldy:
As this is a change in the behavior, this must be pushed as an enhancement into develop branch...

PR #30444, opened 2024-07-24
NEW Can set the contact type for the autoassigned user that creates a project

Slordef:
# FIX project: ability to select contact type for project creator

Currently, when we create a project, the user who creates it is automatically assigned as a contact with type "PROJECTLEADER".

In this PR, I make sure that the user can choose their type with a default selection of "PROJECTLEADER"


eldy:
Any change on screen must provide a screenshot of the interface change after the PR to be validated. Can you paste it ?

Slordef:
image
image

eldy:
My understanding seeing the screenshot is that there is a new combo to select the type of contact that will be used during the assignation of the user on the project. User that create the project is not necessarly the "project manager", this is why we set a contact type, to know the type of contact.

Are you ok if the label before the combo is
"Assign me as contact with the type"
?

Slordef:
Yes of course

PR #30450, opened 2024-07-24
Qual: add fields array property  into categorie class

FHenry:
like other class, Categorie class deserve her fields property array


eldy:
Why in v20 and not develop ?

FHenry:
@eldy 
Not mandatory in 20.  I've change the base branch.

FHenry:
@eldy 
What's missing to merge this PR in develop ?

PR #30466, opened 2024-07-25
Fix 20 missing entity filter

hregis:


hregis:
@eldy see you why the entity field is not created in llx_c_holiday_types ? (Travis error)

hregis:
@eldy ok the wrong table name was fixed in develop but not in branch 20... it's fixed

PR #30467, opened 2024-07-25
Fix uniformize shipment PDF for sales name

atm-thibaultf:
# FIX|Fix # Correction of ":" in shipment PDFs
In shipment PDF we have some ":" on internal contact which double this mark because it's already in the lang files for all langages (not for de_CH, that's why I have added it in this PR). By the way, it feels strange to me because for me the ":" should be displayed into the code and the lang files should contain only "Name" without the ":". The point is, if we want to do this we have to change all the lang files...


eldy:
I confirm that a better fix should be to remove the : from the language file.
So i pushed a fix in main languages.  As it is fixed into en_US language, fix should be propagated to other language in transifex.

PR #30473, opened 2024-07-26
FIX: multicurrency unit price on supplier order and invoice

### LINKED ISSUE #30452: Error on supplier order with multi-currency
### Bug

When we add a product with a buying price in another currency to a supplier order, the price in original currency is wrongly calcultated in most cases.

### Dolibarr Version

18.0 and more (at least)

### Environment PHP

7.4 and 8.1

### Environment Database

MariaDB 11.4.2

- Create a new product from a supplier whose currency is not yours.

- Add a price to this product in the supplier's currency with a conversion rate.

- Add the product to an order to this supplier, with an important quantity.

- Check the unit product in the supplier's currency, and the total amount in the same currency.

On my side, I created a supplier whose currency is Japan Yen with a rate defined at 167.
I created a product called Issue with a price with this supplier defined at 23000 Japan Yen (first and second screenshot)

I created a new supplier order in Japan Yen, and added this product with a quantity of 15000.

The result should be the following :

- Unit price Excl. vat: 23000/167 = 137,724550898
- Total price Excl. vat: 23000*15000/167 = 2065868,263473054
- Unit price Excl. vat in original currency: 23000
- Total price Excl.vat in original currency: 23000*15000 = 345000000

But, with rounded and current calculation method, we get respectively:
137,72455
2 065 868,25
22 999,99985
344 999 997,75

(Thrid screenshot)

It's a small difference here, but in fact it's due to the original currency price calculation: we start from the unit price in euros here, we multiply it by the conversion rate, and because of rounded values, we don't get the good result.
Because of that, supplier orders are wrong, and can't be used to send the order to the related supplier.

I just opened this issue because I wanted to give the possibility to discuss that point, but I'm ALREADY working on that point, so could you please assign myself to this issue while I can't do it on my own?

### Attached files

image
image
image

BenjaminFlr:
Aimed to fix #30452


eldy:
It seems a fix was already provided into v20. Can you check ?
If true, fix into 18 or 19 must be done propagated using backport.

BenjaminFlr:
You're right, it has been fixed by [this PR](https://github.com/Dolibarr/dolibarr/pull/28021).
How do we backport it? Do I have something to do?

eldy:
Two methods: 

The best one and fast one:
Once you have found the commit (or the list of commits), just do a "git cherry-pick commitid" with git starting from the older one. With luck, you will get no conflict.

The more complex (but also required if conflicts with the cherry-pick are too complex to solve:
See manually what is different and copy manually the change.

BenjaminFlr:
Will open a new one because I failed with git

Issue #19916, opened 2022-01-24
Mass stock transfer Batch/Series drop-down list


DATUC:
Mass stock transfer Batch/Series drop-down list
Hi all,
![image](https://user-images.githubusercontent.com/26550824/150879580-63a8ed96-b246-468a-8cad-9d2af911da38.png)

## Use case
[*Batch/Series drop-down list*]

## Suggested implementation
[*when mass transferring products with Lot/Series numbers, is it possible to scroll down a list from which the user can choose the Lot/Series to transfer??*]

Dominatorr007:
and please also add pdf document generation for the movement

DATUC:
> Mass stock transfer Batch/Series drop-down list Hi all, ![image](https://user-images.githubusercontent.com/26550824/150879580-63a8ed96-b246-468a-8cad-9d2af911da38.png)
> 
> ## Use case
> [_Batch/Series drop-down list_]
> 
> ## Suggested implementation
> [_when mass transferring products with Lot/Series numbers, is it possible to scroll down a list from which the user can choose the Lot/Series to transfer??_]

Any update on this request please

Dominatorr007:
this issue is not resolved and even in advanced product transfer still exists, but there it will not allow movement of stock ib lot/ serial do not match, drop down will be far more efficient

lmag:
Step 1 : Filter product by content of Warehouse https://github.com/Dolibarr/dolibarr/issues/30474
Filter content of product (Ajax) or reload page @FHenry 
Step 2 : Add list of serial/lot of product (Select Product Lot) @evarisk-charles 
Uses "datalist" like in production

PR #30477, opened 2024-07-26
NEW: add warehouseId parameters in select_produit to limit result to existing product in select

FHenry:
In order to prepare for Stock Transfert page a product list limited by existing product in source warehouse, a new parameters $warehouseId to product_list is added

Related to #19916


htdocs/core/class/html.form.class.php
```diff
@@ -2974,10 +2980,19 @@ public function select_produits_list($selected = 0, $htmlname = 'productid', $fi
 		$reshook = $hookmanager->executeHooks('selectProductsListFrom', $parameters); // Note that $action and $object may have been modified by hook
 		$sql .= $hookmanager->resPrint;
 
-		if (count($warehouseStatusArray)) {
-			$sql .= " LEFT JOIN " . $this->db->prefix() . "product_stock as ps on ps.fk_product = p.rowid";
-			$sql .= " LEFT JOIN " . $this->db->prefix() . "entrepot as e on ps.fk_entrepot = e.rowid AND e.entity IN (" . getEntity('stock') . ")";
-			$sql .= ' AND e.statut IN (' . $this->db->sanitize($this->db->escape(implode(',', $warehouseStatusArray))) . ')'; // Return line if product is inside the selected stock. If not, an empty line will be returned so we will count 0.
+		if (count($warehouseStatusArray) || ((int) $warehouseId) > 0 ) {
```
eldy, line 2983: As this is a filter, can you implement it into the WHERE only, using
WHERE EXISTS (....)

FHenry: Not the same result

From sample DB of Dolibarr dolibarr/dev/initdemo/mysqldump_dolibarr_20.0.0.sql

```
SELECT p.rowid,
       p.ref,
       p.label,
       p.description,
       p.barcode,
       p.fk_country,
       p.fk_product_type,
       p.price,
       p.price_ttc,
       p.price_base_type,
       p.tva_tx,
       p.default_vat_code,
       p.duration,
       p.fk_price_expression,
       (CASE WHEN p.stock IS NULL THEN 0 ELSE p.stock END) AS stock,
       pl.label                                            as label_translated,
       pl.description                                      as description_translated
FROM llx_product as p
         LEFT JOIN llx_product_lang as pl ON pl.fk_product = p.rowid AND pl.lang = 'fr_FR'
WHERE p.entity IN (1)
  AND p.fk_product_type = 0
  AND EXISTS (SELECT psw.fk_product FROM llx_product_stock as psw WHERE psw.reel>0 AND psw.fk_entrepot=2) AND ((p.ref LIKE '%COMP%' OR p.label LIKE '%COMP%' OR pl.label LIKE '%COMP%') OR p.barcode LIKE '%COMP%')
ORDER BY p.ref ASC
```

return rows 2 rows
10|COMP-XP4523|Computer XP4523|A powerfull computer XP4523&nbsp;|123456789055|11|0|100|110|HT|10||""||112||
13|COMP-XP4548|Computer XP4523|A powerfull computer XP4523&nbsp;||11|0|100|110|HT|10||""||0|Computer XP4523|A powerfull computer XP4523&nbsp;

But the last row  COMP-XP4548 is not in WAREHOUSEPARIS (id 2)

Can be done with existing SQL or with 
p.rowid IN (....)
if you want in in a where but it's not the better execution paln

eldy: I disagree. All what you can do with a left join can be done with an EXISTS and/or NOT EXISTS.

Also in your sample you use the filter on p inside the EXISTS instead of into the main WHERE.
The EXISTS or NOT EXISTS must only contains the condition on the table product_stock.

Third point, there is no condition into your sub request EXISTS that is in relation with parent table product so you join is not a join but a cartesian produt.
(SELECT psw.fk_product FROM llx_product_stock as psw WHERE psw.reel>0 AND psw.fk_entrepot=2)
should be 
(SELECT psw.fk_product FROM llx_product_stock as psw WHERE psw.reel>0 AND psw.fk_entrepot=2 AND psw.fk_product = p.rowid)

What about this
SELECT p.rowid,
       p.ref,
       p.label,
       p.description,
       p.barcode,
       p.fk_country,
       p.fk_product_type,
       p.price,
       p.price_ttc,
       p.price_base_type,
       p.tva_tx,
       p.default_vat_code,
       p.duration,
       p.fk_price_expression,
       (CASE WHEN p.stock IS NULL THEN 0 ELSE p.stock END) AS stock,
       pl.label                                            as label_translated,
       pl.description                                      as description_translated
FROM llx_product as p
         LEFT JOIN llx_product_lang as pl ON pl.fk_product = p.rowid AND pl.lang = 'fr_FR'
WHERE p.entity IN (1)
  AND p.fk_product_type = 0 
  AND ((p.ref LIKE '%COMP%' OR p.label LIKE '%COMP%' OR pl.label LIKE '%COMP%') OR p.barcode LIKE '%COMP%')
  AND EXISTS (SELECT psw.fk_product FROM llx_product_stock as psw WHERE psw.reel>0 AND psw.fk_entrepot=2 AND psw.fk_product = p.rowid) 
ORDER BY p.ref ASC

The last AND is a condition to say "product must have stock > 0 into warehouse 2"

FHenry: Of course, I forget the last check on product...

I've update the code, it's more readable like that, just a new `where clause`

FHenry:
@eldy 
Every things seems to be OK to merge in develop, what I'm missing ?

Issue #27130, opened 2023-12-15
Issue with Virtual Stock, partial reception using SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK


josett225:
### Bug

When using SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK with values above 5 is generating wrong virtual values when doing partial receptions.

### Environment Version

any version starting from 14.0.5

### Environment PHP

7.4 and above

### Expected and actual behavior

Due to our business in the bike industry, we need to calculate our virtual stock without supplier orders and without open receipts until supplier order is fully closed (all products received STATUS_RECEIVED_COMPLETELY = 5).
The reason is that our lead times are sometimes above several months (up to 48 months), which is not compatible with sales where the maximum waiting time is a week as maximum.

To prevent this, I've used a constant, SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK, with a value of 8 that isn't use for now in Dolibarr, hence it will never occur therefore this excludes counting supplier orders products qty. It could work also with 5 (STATUS_RECEIVED_COMPLETELY = 5).
However the Virtual Stock calculation does not work as soon as we start reception as soon as it is validated. It is even worst with partial receptions. We got immediately a negative virtual stock value as reception is linked with the supplier order status value 4 (STATUS_RECEIVED_PARTIALLY = 4).
As we set a value of 5 and above with SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK, we lost the sync between Supplier Order calculation and Reception generating a big mess. The problem is even more pronounced with partial receipts because the closure of the supplier order came at the end of the chain.

Workaround requiring your approval:
There's a line in the product.class.php code for calculating supplier orders that takes into account this variable:
$filterStatus = empty($conf->global->SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK) ? '3,4' : $conf->global->SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK;
[[Link to the code line on GitHub]](https://github.com/Dolibarr/dolibarr/blob/58eca0ac04047c61b42a306a068d510c450575fd/htdocs/product/class/product.class.php#L5792C9-L5792C9)

To fix the issue,  I modified the management of $filterStatus for receipts as follows:
[...]

PR #30481, opened 2024-07-27
FIX virtual stock if const SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK is…

josett225:
When using SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK const, the function load_virtual_stock (in product.class.php) has to used this CONST to calculate correctly the virtual stock.
Today the default value used is 4 for supplier orders or invoice. 
This has to be changed by using SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -5293,7 +5293,7 @@ class Product extends CommonObject
 		}
 		if (((!empty($conf->fournisseur->enabled) && empty($conf->global->MAIN_USE_NEW_SUPPLIERMOD)) || !empty($conf->supplier_order->enabled) || !empty($conf->supplier_invoice->enabled)) && empty($conf->reception->enabled)) {
 			// Case module reception is not used
-			$filterStatus = '4';
+			$filterStatus = empty($conf->global->SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK) ? '4' : $conf->global->SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK;
 			if (isset($includedraftpoforvirtual)) {
 				$filterStatus = '0,'.$filterStatus;
 			}
@@ -5305,7 +5305,7 @@ class Product extends CommonObject
 		}
 		if (((!empty($conf->fournisseur->enabled) && empty($conf->global->MAIN_USE_NEW_SUPPLIERMOD)) || !empty($conf->supplier_order->enabled) || !empty($conf->supplier_invoice->enabled)) && !empty($conf->reception->enabled)) {
 			// Case module reception is used
-			$filterStatus = '4';
+			$filterStatus = empty($conf->global->SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK) ? '4' : $conf->global->SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK;
 			if (isset($includedraftpoforvirtual)) {
 				$filterStatus = '0,'.$filterStatus;
 			}


htdocs/product/class/product.class.php
```diff
@@ -5293,7 +5293,7 @@ public function load_virtual_stock($includedraftpoforvirtual = null)
 		}
 		if (((!empty($conf->fournisseur->enabled) && empty($conf->global->MAIN_USE_NEW_SUPPLIERMOD)) || !empty($conf->supplier_order->enabled) || !empty($conf->supplier_invoice->enabled)) && empty($conf->reception->enabled)) {
 			// Case module reception is not used
-			$filterStatus = '4';
+			$filterStatus = empty($conf->global->SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK) ? '4' : $conf->global->SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK;
```
eldy, line 5296: As the name says the option "SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK" is to define the status into table SUPPLIER_ORDER to take into account.
Here, we will analyse the table llx_receptiondet_batch (the name in v20 or llx_commande_fournisseur_dispatch in v14). This is a different table with different status. So using the same variable name is not possible.

Pichinov-Jose: Hi @eldy 
I reviewed this issue deeply and need your advice on how to fix it as there are several solutions.
What I discovered so far is :
when we do partial receptions depending if stock module is available and how the configuration of reception are done (either on validation of closed), we have to be careful as there is a stock movement done or not.
As today the load_stats_reception is not taking in consideration the stock configuration. 
This is generating errors in calculating the virtual stock.

Example with stock movement based on validated reception
Product A Supplier Order : 200 
Received 50 with Validation (Supplier Order Partial received = 4)(Stock Movement =+50)
Function load_stats_reception result = 50. The result has to be normally 0 as the real stock has already the right value (+50).

2 solutions here:

- either use field status of llx_receptiondet_batch with a specific value to mention that a stock movement has been done. This could avoid future issues if the stock movement configuration is changing from validation to closing  
- or extending to the SQL request if stock module is actived (and based on the Stock configuration getDolGlobalString('STOCK_CALCULATE_ON_RECEPTION') or getDolGlobalString('STOCK_CALCULATE_ON_RECEPTION_CLOSE') the right calculation based on the fk_status of the Reception Object (llx_reception) with the following statuses :
	const STATUS_DRAFT = 0;
	const STATUS_VALIDATED = 1;
	const STATUS_CLOSED = 2;

Let me know if I am wrong or right.

Best Regards

Jose

Pichinov-Jose:
Superseded by #39674, which targets `develop`.

This PR was opened against the `14.0` branch, which is now end of life, so it could not be merged as it stands. The new PR contains the same one-line change, rebased on `develop` (where the supplier-order side already reads the option, so only the reception side remains hardcoded).

On the point raised in the 2024 review — that receptions are read from a different table with different statuses: `load_stats_reception()` does read the *quantities* from `llx_receptiondet_batch` (`llx_commande_fournisseur_dispatch` in v14), but its status filter is applied to `cf.fk_statut`, i.e. the status of the supplier order in `llx_commande_fournisseur`. That is exactly what `SUPPLIER_ORDER_STATUS_FOR_VIRTUAL_STOCK` defines; the reception rows only provide the quantities. This was already the case on the `14.0` branch of this PR.

The deeper question I raised in January — whether `load_stats_reception()` should also take `STOCK_CALCULATE_ON_RECEPTION` / `_CLOSE` into account, since a stock movement may already have been made — is deliberately left out of the new PR and stays tracked in #27130, so that the simple consistency fix can move forward on its own.

Closing in favour of #39674.

PR #30497, opened 2024-07-31
fix phpstan warnings

Hystepik:
fix some warnings of phpstan


mdeweerd:
It looks like 'null' needs to be dded to the list:

https://github.com/phpstan/phpstan/issues/7627 .

mdeweerd:
Adding 'null' to the type list did not help - I proposed other fixes in #30509 .

PR #30518, opened 2024-08-01
fix: add missing hook in propal and invoice (contact and document tabs)

FHenry:
- `htdocs/comm/propal/contact.php` hookmanager is corrrectly implemented (with doAction) in 19.0. I suggest to add only $hookmanger->init in 18.0 (to at least be able use llxfooter hook).
- `compta/facture/contact.php` hookmanager is corrrectly implemented (with doAction) in 19.0. I suggest to add only $hookmanger->init in 18.0 (to at least be able use llxfooter hook).
- `htdocs/comm/propal/document.php` hookmanager->init is missing even in newer version, implement the same way as it's already done in htdocs/commande/document.php
- `htdocs/compta/facture/document.php` hookmanager->init is missing even in newer version, implement the same way as it's already done in  htdocs/commande/document.php

### DIFF
--- a/htdocs/comm/propal/contact.php
+++ b/htdocs/comm/propal/contact.php
@@ -26,12 +26,12 @@
 
 // Load Dolibarr environment
 require '../../main.inc.php';
-require_once DOL_DOCUMENT_ROOT.'/comm/propal/class/propal.class.php';
-require_once DOL_DOCUMENT_ROOT.'/contact/class/contact.class.php';
-require_once DOL_DOCUMENT_ROOT.'/core/lib/propal.lib.php';
-require_once DOL_DOCUMENT_ROOT.'/core/class/html.formother.class.php';
-require_once DOL_DOCUMENT_ROOT.'/core/class/html.formcompany.class.php';
-require_once DOL_DOCUMENT_ROOT.'/projet/class/project.class.php';
+require_once DOL_DOCUMENT_ROOT . '/comm/propal/class/propal.class.php';
+require_once DOL_DOCUMENT_ROOT . '/contact/class/contact.class.php';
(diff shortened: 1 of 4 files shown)


lvessiller-opendsi:
I think it's a good idea to report theses hooks on 18.0.
However, I wonder why sometimes the init hook is before the "restrictedArea()" function and sometimes after.
If we analyze the "restrictedArea()" function, we can see a call to "executeHooks" method.
For me, the "initHooks" method must be before.
@rycks, what do you think ?

FHenry:
@lvessiller-opendsi I was facing the same question when looking at these inithook and restrictedArea().
I'm agree that it should be before restrictedArea(). If it's OK it will be a new huge PR in develop branch

lvessiller-opendsi:
I think it should be interesting of having the advice of @eldy ?

eldy:
> restrictedArea

i would say that when we face 2 different way of developping, we must align in the direction we have into the modulebuilder template. This template is supposed to be the one the most advanced and the less bugged. So good way would be to have the initHook before the restrictArea.
In the past there was no rule. With modulebuilder enhancement, code is now aligned to have the restrictArea call after all initialization.

FHenry:
@lvessiller-opendsi 
I've apply requested review comment to change execution order of initHook() and restrictedArea() and create another PR for develop to address this point. Are you agree to remove label "waiting review" and add label "ready to merge"

lvessiller-opendsi (approved):

rycks (approved):
ok for me but is that a fix or a "new" ... (then in develop ?)

FHenry:
> ok for me but is that a fix or a "new" ... (then in develop ?)

That's the goal  : backport just this init hook in 18.0, and make something proper in develop, already done a'd merge

PR #30522, opened 2024-08-01
fix: catch error extrafields delete

FHenry:
method extrafields->delete() do not handle error correctly


lvessiller-opendsi (approved):

lvessiller-opendsi:
@rycks 
It's ok for me

FHenry:
@eldy 
Is it in contribution guidelines to allow this kind of changes in 18.0 branch or should I change the PR base branch to develop ?

I've suggested this PR in 18.0 branch because I've encounter "not displayed error messages at all and failed action" when I try to delete extra-fields with foreign key constraint added by external module in 18.0 branch, but that's obviously an edged use case. 

Would it be a better way to change base merge branch (or create a new PR on develop branch) ?

eldy:
> @eldy 
> Is it in contribution guidelines to allow this kind of changes in 18.0 branch or should I change the PR base branch to develop ?
> 
> I've suggested this PR in 18.0 branch because I've encounter "not displayed error messages at all and failed action" when I try to delete extra-fields with foreign key constraint added by external module in 18.0 branch, but that's obviously an edged use case. 
> 
> Would it be a better way to change base merge branch (or create a new PR on develop branch) ?

No guideline available for the moment. I must write it. Yes it is at the edge of the limit. It may changes behaviour so should be in develop, but it also avoids an error visibility so may be small bug. If t's 50/50.

FHenry:
@eldy 
I've change to develop branch

PR #22849, opened 2022-11-15
NEW Add order tags

FlorentPoinsaut:
# CLOSE|Close #22848

# NEW|Add order tags

PR #22734, opened 2022-11-02
NEW Add supplier invoice tags

FlorentPoinsaut:
# CLOSE|Close #22733

# NEW|Add supplier invoice tags

PR #22694, opened 2022-10-28
NEW Add invoice tags

FlorentPoinsaut:
# CLOSE|Close #22693

# NEW|New Add invoice tags

PR #30526, opened 2024-08-01
Add new tags in database

FlorentPoinsaut:
# NEW tags in database

Creation tags in database to allow this features (as asked by @eldy):

- https://github.com/Dolibarr/dolibarr/pull/22849
- https://github.com/Dolibarr/dolibarr/pull/22750
- https://github.com/Dolibarr/dolibarr/pull/22734
- https://github.com/Dolibarr/dolibarr/pull/22694

### DIFF
--- a/htdocs/install/mysql/tables/llx_categorie_invoice.key.sql
+++ b/htdocs/install/mysql/tables/llx_categorie_invoice.key.sql
@@ -1,0 +1,25 @@
+-- ============================================================================
+-- Copyright (C) 2007 Patrick Raguin     <patrick.raguin@gmail.com>
+-- Copyright (C) 2022 Solution Libre SAS <contact@solution-libre.fr>
+--
+-- This program is free software; you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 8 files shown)


hregis:
@eldy There was no development to start changing the tables "llx_categorie_xxxx" to "llx_element_categorie"? (which should be called llx_element_category" by the way!)

eldy:
> @eldy There was no development to start changing the tables "llx_categorie_xxxx" to "llx_element_categorie"? (which should be called llx_element_category" by the way!)

For the moment no. 
In v21, there is some work to standardize the architecture of tables for categories by using the same naming convention everywhere (for exemple for the categories of bank that is currently using different naming convention in v20). 

Also for performance purpose i am not sure we should use llc_element_categorie for main tables (thirdparties, actioncomm...). May be using llx_element_categorie must be common  for uncommon elements only... I still have doubt with that...

hregis:
@eldy and what is this table for? https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/install/mysql/tables/llx_element_categorie.sql

PR #30533, opened 2024-08-02
Fix const INVOICE_USE_RETAINED_WARRANTY

thomas-Ngr:
FIX setup of const INVOICE_USE_RETAINED_WARRANTY

When modifying conf for situation invoice (admin/invoice_situation.php), the const INVOICE_USE_RETAINED_WARRANTY is set to -1 when it is left empty.

This creates issues in DLB on code using getDolGlobalInt('INVOICE_USE_RETAINED_WARRANTY').
___
edit:

Another issue is the display of chosen value right after the update, as demonstrated in this gif : 
situation_invoice

This PR fixes both issues.


htdocs/admin/invoice_situation.php
```diff
@@ -79,12 +79,13 @@
 $item->nameText = $langs->trans('AllowedInvoiceForRetainedWarranty');
 
 $arrayAvailableType = array(
+	0 => "", // don't use $showempty in selectArray since it sets a value of -1
```
eldy, line 82: We must use showempty into select component when we want empty value. Empty value generates troubles depending on js code used to beautify the combi list (like select2).
So fix must be managed during or before  the insert. For example if value is -1, no update should be done.

eldy (changes requested):

htdocs/admin/invoice_situation.php
thomas-Ngr: I changed the PR to solve a related issue on display of chosen value (see initial message edit).

PR #30538, opened 2024-08-02
NEW API /setup endpoint for getting action triggers

JonBendtsen:
#NEW API /setup endpoint for getting action triggers
When using the API [POST] /thirdparties/{id}/notifications call for automatic enabling of notifications it is easier to understand what the code is doing if I am using trigger codes like: PROPAL_VALIDATE, ORDER_VALIDATE and BILL_PAYED rather than numbers like 8, 15 and 24.


JonBendtsen:

PR #30539, opened 2024-08-02
FIX api call Get Notifications output

JonBendtsen:
FIX api call Get Notifications to make output more in line with what post notifications expects as input. This also removes duplicated fields.

New output looks like:
```
  {
    "id": "2",
    "event": "8",
    "socid": "2",
    "contact_id": "2",
    "type": "email",
    "datec": "2024-08-02",
    "tms": "2024-08-02 17:34:24"
  }
  ```
  
  Old output looked like:
  ```
    {
    "id": "2",
    "fk_action": "8",
    "event": "8",
    "fk_soc": "2",
    "socid": "2",
    "fk_contact": "2",
    "target": "2",
    "type": "email",
    "datec": "2024-08-02",
    "tms": "2024-08-02 17:34:24"
  }
  ```
  
  event and fk_action are the same
  fk_soc and socid are the same
  fk_contact and target are the same


JonBendtsen:

PR #30541, opened 2024-08-02
NEW: Create thirdparty notification by action code

JonBendtsen:
#NEW Create thirdparty notification by action code
Creates one thirdparty notification based on supplying the action code in text and not a number.
Existing create thirdparty notification on id numbers expanded with checks for duplications and error message when event key is missing from request_data


JonBendtsen:

PR #30542, opened 2024-08-02
Qual: Fix PhanTypeMismatchProperty and some other phan notices

mdeweerd:
# Qual: Fix PhanTypeMismatchProperty and some other phan notices

Fixes several PhanTypeMismatchProperty notices and some other notices.


mdeweerd:
@eldy https://github.com/Dolibarr/dolibarr/pull/30542/files#annotation_24940857472 seems to be accessing Adherent as an array.

PR #30543, opened 2024-08-03
Fix: 🛟Fix getrights usage & Update the documentation to suggest to use loadRights

mdeweerd:
# Fix: Fix getrights usage & Update the documentation to suggest to use loadRights

Given the pending issues, update the documention regarding getrights to point to loadRights.
Also fixes the issues introduced in the user api class.

Given the pending issues, update the documention regarding getrights to point to loadRights.
Also fixes the issues introduced in the user api class.

### DIFF
--- a/htdocs/core/customreports.php
+++ b/htdocs/core/customreports.php
@@ -80,7 +80,7 @@ if (!defined('USE_CUSTOM_REPORT_AS_INCLUDE')) {
 	// $search_measures, $search_xaxis or $search_yaxis may have been defined by the parent.
 
 	if (empty($user) || empty($user->id)) {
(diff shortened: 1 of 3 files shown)


mdeweerd:
@eldy : Merge this, it will fix CI.

Maybe a special symbol should be chosen for the title if a PR fixes CI.
With GH UI code (exemple `:warning`) : ⚠️ 🥇 🆘 🚑 🛟 ⛳ 🩹

eldy:
Why not. Not sure to know how to find symbols in my Ide (eclipse)

eldy (approved):

mdeweerd:
On windows WIN+;  opens a dialog to find an emoji - I suppose there is one in Linux based GUI environments as well.

However, I'ld be good to settle on one, and document it somewhere.

PR #30557, opened 2024-08-05
Qual: Fix PhanTypeInvalidLeftOperandOfNumericOp

mdeweerd:
# Qual: Fix PhanTypeInvalidLeftOperandOfNumericOp

These changes ensure that the left operand is a numeric value


mdeweerd:
@eldy  This PR is 'ready' except for the notifications left in 
`htdocs/takepos/invoice.php`.

https://github.com/Dolibarr/dolibarr/pull/30557/files#annotation_24741053698

Either  `$number` is the invoice number and it can be a string in general IMHO, or for POS it's forced to an integer.

If it's an integer, then GETPOST should be updated, if it can be a string, then the code needs more updates.

eldy:
> @eldy This PR is 'ready' except for the notifications left in `htdocs/takepos/invoice.php`.
> 
> https://github.com/Dolibarr/dolibarr/pull/30557/files#annotation_24741053698
> 
> Either `$number` is the invoice number and it can be a string in general IMHO, or for POS it's forced to an integer.
> 
> If it's an integer, then GETPOST should be updated, if it can be a string, then the code needs more updates.

I pushed a fix for this. Can you merge it with your PR ?

htdocs/takepos/invoice.php
```diff
@@ -937,7 +937,7 @@ function fail($message)
 
 				$pu_ht = price2num($number / (1 + ((float) $vatratecleaned / 100)), 'MU');
```
mdeweerd: @eldy I think that this expression is not exactly what is needed - $number is interpreted as a number in the expression here - I suppose it should be treated by price2num first - I casted it for now .

eldy: You're right
$number must be replaced into this line with 
price2num($number, 'MU');

mdeweerd: I changed it.

PR #30559, opened 2024-08-06
Add endpoint /contract/expired to get expired contracts

kadogo:
# NEW|New Adding /contracts/expired endpoint

The interface let us see the contracts that will be expiring based on a configuration value. We wanted a way to get them through the web service. It was easier to just add an endpoint for it than doing multiple search.

Our instance is pretty old (14.0.5) but the patch applied properly on the develop branch.

Don't hesitate to update the code if there is a better way to do.

Cheers

### DIFF
--- a/htdocs/contrat/class/api_contracts.class.php
+++ b/htdocs/contrat/class/api_contracts.class.php
@@ -244,6 +244,95 @@ class Contracts extends DolibarrApi
 	}
 
 	/**
+	 * Get all contracts that will expiring soon
+	 *
+	 * @url	GET /expired
+	 *
+	 * @return array
+	 */
+	public function getExpired()
+	{
+		global $conf;
+
+		if (!DolibarrApiAccess::$user->rights->contrat->lire) {
+			throw new RestException(401);
+		}
+
+		$obj_ret = array();
+
+		// case of external user, $thirdparty_ids param is ignored and replaced by user's socid
+		$socids = DolibarrApiAccess::$user->socid ? DolibarrApiAccess::$user->socid : $thirdparty_ids;
+
+		// If the internal user must only see his customers, force searching by him
+		$search_sale = 0;
+		if (!DolibarrApiAccess::$user->rights->societe->client->voir && !$socids) {
+			$search_sale = DolibarrApiAccess::$user->id;
+		}
+
+		$sql = "SELECT c.rowid as contract_id, cd.rowid, cd.date_fin_validite as datefin";
+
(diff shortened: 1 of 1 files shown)


kadogo:
Hello @eldy 

Does it mean I need to fix the spaces / tabs issues mentionned here https://github.com/Dolibarr/dolibarr/actions/runs/10267943407/job/28409725198?pr=30559 and that the PR itself will be good to be merged ?

eldy:
PR are analysed only once there is no CTI errors.
Also what is goal of endpoint ? Because what is "expired" is not a contract but lines of a contract.
So it would be better to add an endpoint called /lines/expired if goal is to return all lines of contract that are expired.

kadogo:
Hello @eldy 

The PR is to have a way to get the expiration like we can see in the interface.

image

image

It really looks like it is the contract that is expired based on the number of days that we configure in the backend configuration. Our Dolibarr version is pretty old, so maybe there are some differences with the stable version.

Regards

eldy:
> Hello @eldy
> 
> The PR is to have a way to get the expiration like we can see in the interface.
> 
> image
> 
> image
> 
> It really looks like it is the contract that is expired based on the number of days that we configure in the backend configuration. Our Dolibarr version is pretty old, so maybe there are some differences with the stable version.
> 
> Regards

No it is not the contract, it is the line of contract (called "service lines"). On screenshot we see the date of end of line. But each line is inside one and only contract, so we can show the contract once we have an expired line. Note that we can have several times the same contract if a contract has 2 lines expired.

kadogo:
Hello @eldy 
Thanks for your answer.

In the interface it really looks like it is the contract, but I don't mind changing it as contract/lines/expired if you prefer. But the return is a contract object and not just the expired lines. Is that alright?

Is there a notion of begin / end of a contract that is stored when the contract is expired?

eldy:
> Is there a notion of begin / end of a contract that is stored when the contract is expired?

No. There is no notion of begin / end of a contract stored, expiration is stored into each lines.

htdocs/contrat/class/api_contracts.class.php
```diff
@@ -243,6 +243,95 @@ public function getLines($id)
 		return $result;
 	}
 
+	/**
+	 * Get all contracts that will expiring soon
+	 *
+	 * @url	GET /expired
```
eldy: I suggest to rename the endpoint  /lines  and add a parameter expired = 1 to get only expired lines
Also a parameter 	 
`* @param string           $sqlfilters          Other criteria to filter answers separated by a comma. Syntax example "(t.ref:like:'SO-%') and (t.date_creation:<:'20160101')"
`Can be added

PR #30561, opened 2024-08-06
NEW display for box fediverse

lamrani002:
# NEW|New functionnality for get fediverse post by creating a widget,  create class box_fediverse, fediverseparser 
mastodon


htdocs/admin/dict.php
```diff
@@ -2620,7 +2627,7 @@
 }
 
 print '<br>';
-
+print dol_get_fiche_end();
```
eldy, line 2630: There is  test on line 1265 to execute dol_get_fiche_head so we shuld have same test to do dol_get_fiche_end().
Otherwise code is not correclty balanced.

lamrani002: yes, i forgot it

eldy:
It seems a similar PR was already merged.
And after some enhancement were done. So this PR has a lot of conflict. Does this PR still has things not merged ?

lamrani002:
yeah i think i will close this PR because we had already all we need

PR #30563, opened 2024-08-07
Qual: Fix PhanPluginUnknownObjectMethodCall ("part 1")

mdeweerd:
# Qual: Fix PhanPluginUnknownObjectMethodCall ("part 1")

This fixes about 20% of the PhanPluginUnknownObjectMethodCall mosty by type hinting.

There are some issues that remain, some reveal bugs IMHO.

Regarding 'getExample()': a few classes take parameters for this.  I'll be changing that to rename those to 'getExampleFromData()' and create a new getExample() that takes not arguments + update the callers to the original getExample (to use the renamed method).

However, I prefer that you look at these cases:

```plaintext
htdocs\admin\reception_setup.php:246 PhanUndeclaredProperty Reference to undeclared property \ModelNumRefReception->scandir
htdocs\admin\supplier_payment.php:273 PhanUndeclaredProperty Reference to undeclared property \ModeleNumRefSupplierPayments->scandir


htdocs\knowledgemanagement\core\modules\knowledgemanagement\mod_knowledgerecord_advanced.php:145 PhanUndeclaredProperty Reference to undeclared property \KnowledgeRecord->date (Did you mean expr->date_creation or expr->date_validation or expr->date_modification or expr->date_cloture)

# $type seems to have multiple, different uses - just changing from static to dynamic is  likely nok.
htdocs\intracommreport\list.php:572 PhanAccessPropertyStaticAsNonStatic Accessing static property \IntracommReport::$type as non static
```

Note: the 'write_file' notices are likely handled in #30511 - I'll rebase this PR when that one is integrated.


mdeweerd:
This PR is ready for review (not ready for merge because there are some open issues to attend to, and another PR #30511 needs to be integrated first).

mdeweerd:
EDIT: I found the information in the mean time.

@eldy About 'public $version'.

Should it be kept in the range `''|'development'|'dolibarr'|'experimental'`, or maybe yes, except for the case where the version is already fixed?

Also, class  pdf_octopus sets it to `'disabled'` - Add it to the list?.

Example:
```
export_csviso.modules.php: PhanTypeMismatchPropertyProbablyReal: Assigning '1.32' of type '1.32' to property but \ExportCsvIso-&gt;version is ''|'development'|'dolibarr'|'experimental' (no real type) (the inferred real assigned type has nothing in common with the declared phpdoc property type)
```

eldy:
Version can be

dolibarr, dolibarr_deprecated, experimental, development, disabled    or    x.y.z

PR #30565, opened 2024-08-07
fix phpstan level 5

frederic34:


mdeweerd:
@frederic34 FYI: I have some pending PRs that fix phpstan issues.

Example: #30536 and I believe I have another fix in one of the commits of another outstanding PR.  (I can not find it so don't care about the second one).

frederic34:
I am going to resolve :
image

and also remove not matched errors

image

PR #30567, opened 2024-08-07
NEW more complete stock product API call

ptibogxiv:


htdocs/product/class/api_products.class.php
```diff
@@ -1995,15 +1995,18 @@ public function getStock($id, $selected_warehouse_id = null)
...
 				if ($warehouse_id != $selected_warehouse_id) {
 					unset($stockData[$warehouse_id]);
 				}
 			}
 		}
+		$obj_ret = $this->_filterObjectProperties($this->_cleanObjectDatas($product_model), 'stock_warehouses,stock_reel,stock_theorique');
+		$obj_ret->stock_warehouses = $stockData;
 
-		return array('stock_warehouses'=>$stockData);
+		return $obj_ret;
```
eldy, line 2009: Does tthis keep the output format ?
What is the json string after the pr compared to before ?

ptibogxiv: This PR is fully compatible with old version of dolibarr, just add more informations of stock reel and theorique
before
```
{
  "stock_warehouses": {
    "1": {
      "real": "3",
      "id": "11"
    }
  }
}
```

after
```
{
  "stock_reel": 3,
  "stock_theorique": -1,
  "stock_warehouses": {
    "1": {
      "real": "3",
      "id": "11"
    }
  }
}
```

PR #30581, opened 2024-08-08
FIX| clearsession extrafleds

ATM-NicolasV:
# FIX|Fix 
  clearsession extrafleds when extrafiels is mendatory but empty


eldy:

PR #30582, opened 2024-08-08
NEW Add caldav on fichinter

defrance:
The goal is to access on fichinterdet line from a caldav to follow interventional
this is based on actioncomm caldav


htdocs/admin/fichinter_xcal.php
```diff
@@ -0,0 +1,244 @@
...
+	}
+}
+
+
+
+/**
+ * View
+ */
+
+if (!isset($conf->global->MAIN_FICHINTER_EXPORT_PAST_DELAY)) {
```
eldy, line 87: Use conf->global is now forbiden. Can you replace with getDolGlobalString ?

eldy (changes requested):

htdocs/admin/fichinter_xcal.php
defrance: change made

PR #30585, opened 2024-08-08
Fix Preview present even when there is no corresponding file

MaximilienR-easya:
# Fix
We got case where in the client page, in the list of propale, contract, etc, case where one of them don't have a PDF but still have a preview link.
We add a check to verify if the file is existing and if it's non-existent, we don't show a preview link.

### DIFF
--- a/htdocs/comm/card.php
+++ b/htdocs/comm/card.php
@@ -868,9 +868,9 @@ if ($object->id > 0) {
 				$filedir = $conf->propal->multidir_output[$objp->entity].'/'.dol_sanitizeFileName($objp->ref);
 				$file_list = null;
 				if (!empty($filedir)) {
-					$file_list = dol_dir_list($filedir, 'files', 0, '', '(\.meta|_preview.*.*\.png)$', 'date', SORT_DESC);
+					$file_list = dol_dir_list($filedir, 'files', 0, dol_sanitizeFileName($objp->ref).'.pdf', '(\.meta|_preview.*.*\.png)$', 'date', SORT_DESC);
 				}
-				if (is_array($file_list)) {
+				if (is_array($file_list) && !empty($file_list)) {
 					// Defined relative dir to DOL_DATA_ROOT
 					$relativedir = '';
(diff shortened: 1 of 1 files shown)


htdocs/comm/card.php
```diff
@@ -887,7 +887,8 @@
 						}
 					}
 					$relativepath = dol_sanitizeFileName($objp->ref).'/'.dol_sanitizeFileName($objp->ref).'.pdf';
-					print $formfile->showPreview($file_list, $propal_static->element, $relativepath, 0);
+					if (file_exists($conf->propal->multidir_output[$objp->entity].'/'.$relativepath))
```
eldy, line 890: Because wec' are in a loop, doing a filexist generate a lot of io.
Do you know in which you case you have the file that does not exists ?

MaximilienR-easya: It can happen when the file was deleted by a user for any reason, or the file was not generated.
I don't see another way to test if the file is existing

rycks:
hello @MaximilienR-easya i do not have the bug on my dolibarr ...

on /comm/index.php?mainmenu=commercial&leftmenu= there is now preview icon:

image

could you please explain to us where is the bug ?

thanks,
Éric

MaximilienR-easya:
> hello @MaximilienR-easya i do not have the bug on my dolibarr ...
> 
> on /comm/index.php?mainmenu=commercial&leftmenu= there is now preview icon:
> 
> image
> 
> could you please explain to us where is the bug ?
> 
> thanks, Éric

The problem is not on comm/index.php but in comm/card.php.
The problem also exists when there is a file but is not named as the reference.

htdocs/comm/card.php
MaximilienR-easya: I have modified the PR to not use file_exists()

rycks:
Here is the bug!

image

image

rycks (approved):
ok for me with that solution

lvessiller-opendsi (approved):

PR #30589, opened 2024-08-08
Qual replacing $conf->global-> with getDolGlobalString because phan complained

JonBendtsen:
#Qual replacing $conf->global-> with getDolGlobalString because phan complained in a different commit
replacing $conf->global-> with getDolGlobalString because phan complained in this commit
https://github.com/Dolibarr/dolibarr/pull/30562/commits/b236d606fef7e5e62c86c8bd807ae9fd57371a84

Code copied from the very same file.

### DIFF
--- a/htdocs/api/class/api_setup.class.php
+++ b/htdocs/api/class/api_setup.class.php
@@ -2259,7 +2259,7 @@ class Setup extends DolibarrApi
 		global $conf, $mysoc;
 
 		if (!DolibarrApiAccess::$user->admin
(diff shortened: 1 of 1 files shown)


JonBendtsen:
Not tested with any API_LOGINS_ALLOWED_FOR_GET because I don't know how.

I have tested these API endpoints with the token of the superadmin user, and they did work, except for
```
    "message": "Internal Server Error: Xml Integrity File of application not found: https://www.dolibarr.org/files/stable/signatures/filelist-21.0.0-alpha.xml - 404  "

```

mdeweerd:
FYI: the root cause is that `global $conf` was missing, but this change implements the recommended approach to access the configuration.

JonBendtsen:
> FYI: the root cause is that `global $conf` was missing, but this change implements the recommended approach to access the configuration.

Thanks for the hint @mdeweerd  I used it elsewhere

Issue #30482, opened 2024-07-27
REST API :: PUT /users/{id} :: command for regenerate/change password


adriansev:
### Feature Request

It would be really useful to have a programmatic way to regenerate and send user password.

### Use case

For programmatic creation of users, an addition of regenerate and send password would complete the user addition procedure.
Moreover, for security reasons, would be useful to have this functionality as part of REST API.

JonBendtsen:
Some notes for changing the password 

https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/user/card.php#L679
https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/user/card.php#L686


Seems like the entire segment from line 673 to like 695 handles password changes and if needed sending it to the user.

JonBendtsen:
The 2 functions called above are found here: 

public function setPassword:
https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/user/class/user.class.php#L2394

public function send_password
https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/user/class/user.class.php#L2536

JonBendtsen:
What kind of http method would be a good choice here?

GET, PUT, POST?

GET - no, because we are not getting any data
PUT - no because we are not putting any data either, the password is generated and then sent, we do not supply the password
POST - same reasons as PUT

JonBendtsen:
basing my attempt on:

```
	 * @url	GET {id}/setGroup/{group}
	 */
	public function setGroup($id, $group, $entity = 1)
```

JonBendtsen:
I'll sprinkle in some code from `public function put($id, $request_data = null)` as well

JonBendtsen:
It changes the password, but at the moment I can not get it to send, and I do not like some variables I had to set

![image](https://github.com/user-attachments/assets/f1651c24-8429-4c38-bd70-b653963fc862)

PR #30590, opened 2024-08-08
NEW API call to re-generate a users password.

JonBendtsen:
# NEW API call to re-generate a users password 
Adds a API endpoint GET /users/{user}/setPassword with option for sending the new password to user.
Partial fix of #30482.

### DIFF
--- a/htdocs/user/class/api_users.class.php
+++ b/htdocs/user/class/api_users.class.php
@@ -472,7 +472,7 @@ class Users extends DolibarrApi
 		// I do not like that the first parameter has to be an empty string :-()
 		// in user/card.php it is called like this setPassword($user, '')
 		// and my development server does not actually send email, so I can't really test this
-		$newpassword = $this->useraccount->setPassword($user, '');	// This will generate a new password
+		$newpassword = $this->useraccount->setPassword(global $user, '');	// This will generate a new password
 		if (is_int($newpassword) && $newpassword < 0) {
 			throw new RestException(520, 'ErrorFailedToSetNewPassword'.$this->useraccount->error);
 		} else {
@@ -480,7 +480,7 @@ class Users extends DolibarrApi
 			if ($send_password) {
 				// I do not like that the first parameter has to be an empty string :-(
 				// in user/card.php it is called like this send_password($user, $newpassword)
-				if ($this->useraccount->send_password($user, $newpassword) > 0) {
+				if ($this->useraccount->send_password(global $user, $newpassword) > 0) {
 					return 2;
 				} else {
 					throw new RestException(530, 'ErrorFailedSendingNewPassword'.$this->useraccount->error);
--- a/htdocs/user/class/api_users.class.php
+++ b/htdocs/user/class/api_users.class.php
@@ -430,6 +430,65 @@ class Users extends DolibarrApi
 		}
 	}
 
+	/**
+	 * Update a users password
+	 *
+	 * @param   int     $id        User ID
+	 * @param	int		$send_password		Set this to 1 to have thenew password sent to the user
+	 * @return  int                1 if password changed, 2 if password changed and sent
+	 *
+	 * @throws RestException 403 Not allowed
+	 * @throws RestException 404 User not found
+	 * @throws RestException 500 System error
+	 *
+	 * @url	GET {id}/setPassword
+	 */
+	public function setPassword($id, $send_password = 0, $entity = 1)
+	{
+		global $conf;
+
+		if (!DolibarrApiAccess::$user->hasRight('user', 'user', 'creer') && empty(DolibarrApiAccess::$user->admin)) {
+			throw new RestException(403, "setPassword on user not allowed");
+		}
+
+		$result = $this->useraccount->fetch($id);
+		if (!$result) {
+			throw new RestException(404, 'User not found, no password changed');
+		}
+
+		if (!DolibarrApi::_checkAccessToResource('user', $this->useraccount->id, 'user')) {
+			throw new RestException(403, 'User not allowed for login '.DolibarrApiAccess::$user->login);
+		}
+
+		if (isModEnabled('multicompany') && getDolGlobalString('MULTICOMPANY_TRANSVERSE_MODE') && !empty(DolibarrApiAccess::$user->admin) && empty(DolibarrApiAccess::$user->entity)) {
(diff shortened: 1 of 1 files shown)


JonBendtsen:
Not yet ready. I can re-generate a password, but I can not send it to the recipient - maybe because I don't have email sending in my development dolibarr?

I get this error if I try:
```
PHP Fatal error:  Uncaught Error: Call to a member function getFullName() on string in
/var/www/html/user/class/user.class.php:2580
Stack trace:
#0 /var/www/html/user/class/api_users.class.php(477): User->send_password()
#1 [internal function]: Users->setPassword()
#2 /var/www/html/includes/restler/framework/Luracast/Restler/Restler.php(1068): call_user_func_array()
#3 /var/www/html/includes/restler/framework/Luracast/Restler/Restler.php(308): Luracast\\Restler\\Restler->call()
#4 /var/www/html/api/index.php(421): Luracast\\Restler\\Restler->handle()
#5 {main}
  thrown in /var/www/html/user/class/user.class.php on line 2580, referer: http://localhost/api/index.php/explorer/
```

JonBendtsen:
image

JonBendtsen:
line 2580 is this line:

```
$mesg .= $user->getFullName($outputlangs); // Username that send the email (not the user for who we want to reset password)
```

JonBendtsen:
latest error message

```
PHP Fatal error:  Uncaught Error: Call to a member function getFullName() on null in /var/www/html/user/class/user.class.php:2580
Stack trace:
#0 /var/www/html/user/class/api_users.class.php(483): User->send_password()
#1 [internal function]: Users->setPassword()
#2 /var/www/html/includes/restler/framework/Luracast/Restler/Restler.php(1068): call_user_func_array()
#3 /var/www/html/includes/restler/framework/Luracast/Restler/Restler.php(308): Luracast\Restler\Restler->call()
#4 /var/www/html/api/index.php(421): Luracast\Restler\Restler->handle()
#5 {main}
  thrown in /var/www/html/user/class/user.class.php on line 2580, referer: http://localhost/api/index.php/explorer/
 ```
 
 and this is with this call:
 
 ```
 				if ($this->useraccount->send_password($user, $newpassword) > 0) {
```

JonBendtsen:
Seems like it works now, I don't get any more errors in the apache log or in dolibarr.log, even when sending passwords.

The only error I get is that it can not send email - which is expected because I have not configured email sending for my development server.


```
{
  "error": {
    "code": 530,
    "message": "ErrorFailedSendingNewPasswordFailed to send password Failed to send mail with php mail to HOST=localhost, PORT=25<br>Recipient address '' invalid"
  },
  "debug": {
    "source": "api_users.class.php:487 at call stage",
    "stages": {
      "success": [
        "get",
        "route",
        "negotiate",
        "authenticate",
        "validate"
      ],
      "failure": [
        "call",
        "message"
      ]
    }
  }
}
```

JonBendtsen:
Thanks to @mdeweerd for the global $user hint, I used it elsewhere

mdeweerd:
> Thanks to @mdeweerd for the global $user hint, I used it elsewhere

It was `global $conf`, but I believe there is $user as well.   The global $conf is best avoided, and global $user is normal usage.

htdocs/user/class/api_users.class.php
```diff
@@ -430,6 +430,66 @@ public function put($id, $request_data = null)
...
+		} else {
+			// When using API, action is done on entity of logged user because a user of entity X with permission to create user should not be able to
+			// hack the security by giving himself permissions on another entity.
+			$entity = (DolibarrApiAccess::$user->entity > 0 ? DolibarrApiAccess::$user->entity : $conf->entity);
+		}
+
+		// I do not like that the first parameter has to be an empty string :-()
+		// in user/card.php it is called like this setPassword($user, '')
+		// and my development server does not actually send email, so I can't really test this
+		$newpassword = $this->useraccount->setPassword(global $user, '');	// This will generate a new password
```
mdeweerd, line 475: `global $user;` needs to be on a separte line.

(Also, make sure that you are referring to the user you want to change the password for the user you want to because the code above seems to be checking access to the provides user "ressource", and here you seem to be updating the password of the user itself).

JonBendtsen: thanks @mdeweerd, I think I forgot to push the commit I made locally this morning before going to work, latest commits gives no error other than it can not send email because I don't have email configured on my development dolibarr.

htdocs/user/class/api_users.class.php
```diff
@@ -470,18 +470,13 @@ public function setPassword($id, $send_password = 0, $entity = 1)
...
 		$newpassword = $this->useraccount->setPassword($user, '');	// This will generate a new password
 		if (is_int($newpassword) && $newpassword < 0) {
 			throw new RestException(500, 'ErrorFailedToSetNewPassword'.$this->useraccount->error);
 		} else {
 			// Success
 			if ($send_password) {
-				// I do not like that the first parameter has to be an empty string :-(
-				// in user/card.php it is called like this send_password($user, $newpassword)
-				if ($this->useraccount->send_password( $user, $newpassword) > 0) {
+				if ($this->useraccount->send_password($user, $newpassword) > 0) {
```
eldy, line 479: In API context, the $user may be defined as global variable but should be empty.
Can you try to replace with 
DolibarrApiAccess::$user

That is the $user that called the API.

JonBendtsen: @eldy  it was $this->useraccount which was needed and all checks seems to be successful.

I've seen no error with send empty, send false or send true. I can not send though as I have not configured email in my development server.

Also, password does change.

hregis:
@JonBendtsen @eldy the $entity parameter and this part of the code are useless, you have to use "DOLAPIENTITY" in the curl header instead.

```
if (isModEnabled('multicompany') && getDolGlobalString('MULTICOMPANY_TRANSVERSE_MODE') && !empty(DolibarrApiAccess::$user->admin) && empty(DolibarrApiAccess::$user->entity)) {
			$entity = (!empty($entity) ? $entity : $conf->entity);
		} else {
			// When using API, action is done on entity of logged user because a user of entity X with permission to create user should not be able to
			// hack the security by giving himself permissions on another entity.
			$entity = (DolibarrApiAccess::$user->entity > 0 ? DolibarrApiAccess::$user->entity : $conf->entity);
		}
```

PR #30593, opened 2024-08-09
Fix error due to mismatch type for duration value

MaximilienR-easya:
# Fix

Fix error due to mismatch type for duration value. It should be an INT as it is everywhere in the code compared to INT value and passed as an INT to function or used in INT operation.
Keeping as a string can cause in example this error `error evaluating code: Unsupported operand types: int + string`
since PHP8.0.

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -2541,7 +2541,7 @@ class Product extends CommonObject
 				$this->fk_default_bom = $obj->fk_default_bom;
 
 				$this->duration = $obj->duration;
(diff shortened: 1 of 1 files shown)


mdeweerd:
@eldy This is where fixing the "Could not infer type" issues is useful - I saw the same on duration_value in another file thanks to fixing some of these with phan:

In PR #30591:
95304013630a or line: 752fde11bdded236589d8ff8157e353eeb4e48ebc5f84cbcd8ac2f1d60536a45R508

And was reported by phan here: https://github.com/Dolibarr/dolibarr/actions/runs/10311736712/job/28545831211#step:5:620 .

PR #30595, opened 2024-08-09
NEW translation with action triggers API

ptibogxiv:


ptibogxiv:
@eldy  i don't see something to fix in this PR

eldy:
I don't think it is important because api files can't be used by other code directly, and params of api are not ordered but by principle, when we add a new param in a function (here $lang), we should add it at end.

ptibogxiv:
> I don't think it is important because api files can't be used by other code directly, and params of api are not ordered but by principle, when we add a new param in a function (here $lang), we should add it at end.

i know but i respect the same order that in other object and API. It seems  to be good that API are the same everywhere.It's a new api not publish yet in a stable version so i think this change can be ok (function introduce last week)

eldy:
> > I don't think it is important because api files can't be used by other code directly, and params of api are not ordered but by principle, when we add a new param in a function (here $lang), we should add it at end.
> 
> i know but i respect the same order that in other object and API. It seems to be good that API are the same everywhere.It's a new api not publish yet in a stable version so i think this change can be ok (function introduce last week)

That is a valid reason...

PR #30599, opened 2024-08-10
FIX compatibility with MULTICOMPANY_TRANSVERSE_MODE

hregis:


hregis:
@eldy where are "$bookmarkList" and "$searchForm" used? In DoliDroid?

PR #30604, opened 2024-08-11
Qual: Phan fixes (mainly Initial type cannot be inferred) in popular changed classes

mdeweerd:
# Qual: Phan fixes (mainly Initial type cannot be inferred) in popular changed classes

This PR concentrates on Societe, Facture and Product that are amongst the most changed classes in 1 year.  Issues that appeared following the fixes or notices during the process were also adressed.


htdocs/core/modules/barcode/mod_barcode_thirdparty_standard.php
```diff
@@ -231,17 +231,22 @@ public function getNextValue($objthirdparty = null, $type = '')
 	 *
 	 *	@param	DoliDB		$db					Database handler
 	 *	@param	string		$code				Code to check/correct
-	 *	@param	Societe		$thirdparty			Object third-party
-	 *  @param  int		  	$thirdparty_type   	0 = customer/prospect , 1 = supplier
+	 *	@param	Societe|Product	$thirdparty	Object third party
```
eldy, line 234: I see no reason to allow "Product" here in documentation. In this file method should be called only for Societe.
Why allowing Products here ?

htdocs/core/modules/barcode/mod_barcode_product_standard.php
```diff
@@ -224,18 +224,22 @@ public function getNextValue($objproduct = null, $type = '')
 	 *
 	 *	@param	DoliDB		$db					Database handler
 	 *	@param	string		$code				Code to check/correct
-	 *	@param	Product		$product			Object product
-	 *  @param  int		  	$thirdparty_type   	0 = customer/prospect , 1 = supplier
+	 *	@param	Product|Societe	$product			Object product
```
eldy, line 227: I see no reason to allow "Societe" here in documentation. In this file method should be called only for Product or Service.
Why allowing Societe here ?

mdeweerd: We have mod_barcode_product.... and mod_barcode_societe... deriving from the same parent class.  The method signatures need to be the same.

Another class could be created, one for barcode_product and one for barcode_thirdparty, but I am focussing on applying typing hints for what they are rather than adding structural changes to fix the notificitions.

In general it would be better to test the class type (instanceof) rather than relying solely on the type/element text field or filename.

htdocs/core/modules/barcode/mod_barcode_thirdparty_standard.php
mdeweerd: Same anwser as for Societe that is allowed:

We have mod_barcode_product.... and mod_barcode_societe... deriving from the same parent class.  The method signatures need to be the same.

Another class could be created, one for barcode_product and one for barcode_thirdparty, but I am focussing on applying typing hints for what they are rather than adding structural changes to fix the notificitions.

In general it would be better to test the class type (instanceof) rather than relying solely on the type/element text field or filename.

PR #30607, opened 2024-08-12
restored value __REF__

atm-jonathan:
# Instructions
FIX To restore the value of the __REF__ key, the user can still use __REF_CLIENT__ in an email template.

### DIFF
--- a/htdocs/core/tpl/card_presend.tpl.php
+++ b/htdocs/core/tpl/card_presend.tpl.php
@@ -85,10 +85,8 @@ if ($action == 'presend') {
 	}
 
 	$topicmail = '';
-	if (empty($object->ref_client)) {
+	if (!empty($object->ref)) {
 		$topicmail = $outputlangs->trans($defaulttopic, '__REF__');
-	} elseif (!empty($object->ref_client)) {
-		$topicmail = $outputlangs->trans($defaulttopic, '__REF__ (__REF_CLIENT__)');
 	}
 
 	// Build document if it not exists


htdocs/core/tpl/card_presend.tpl.php
```diff
@@ -85,10 +85,8 @@
 	}
 
 	$topicmail = '';
-	if (empty($object->ref_client)) {
+	if (!empty($object->ref)) {
 		$topicmail = $outputlangs->trans($defaulttopic, '__REF__');
-	} elseif (!empty($object->ref_client)) {
-		$topicmail = $outputlangs->trans($defaulttopic, '__REF__ (__REF_CLIENT__)');
```
eldy, line 91: Why removing this ?

atm-jonathan: There are no bugs
When you use the subtitution key ‘__REF__’ and you have filled in the ‘ref_client’ field in the interface, this field is concatenated with __REF__ in the email subject, which gives :
__REF__ (__REF_CLIENT__)
But at no time in the email template did we call the subtitution key __REF_CLIENT__.
**Concatenation is imposed by the code.**
If the user absolutely wanted to see the ref_client field in the subject line of their email, they could always add the __REF_CLIENT__ subtitution key in the subject line of their email template.
I just want that when we use __REF__ we display what corresponds to the __REF__ and don't concatenate anything.

atm-jonathan: To reproduce this behaviour, simply use an email template with the key `__REF__` in the subject line.

eldy: > To reproduce this behaviour, simply use an email template with the key `__REF__` in the subject line.

I do not succeed in reproducing the problem. I created an email template with __REF__ as topic. When I send an email from a proposal P1 and ref_client PC1, after selecting the email template, the topic is set to P1 as expected.

atm-jonathan: I made a mistake.
To reproduce, you don't need to use an email template.

eldy: > I made a mistake. To reproduce, you don't need to use an email template.

So how did you do ?

atm-jonathan: I have a customer who wants only their reference in the subject, without including the customer reference, even if the customer reference is set. How can we properly fix this?

eldy: If use don't want the default email autofill behaviour, he can set its own email template and set it as default.

atm-jonathan: OK, so that's how we're going to do it
Thanks anyway

PR #30608, opened 2024-08-12
FIX selectForFormsListWhere should getEntity from object element

fappels:
selectForFormsListWhere should getEntity from object element and not from object table_element

Is old issue, not sure if backport needed.

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -8512,7 +8512,7 @@ class Form
 			$sql .= " WHERE 1=1";
 			if (isset($objecttmp->ismultientitymanaged)) {
 				if ($objecttmp->ismultientitymanaged == 1) {
-					$sql .= " AND t.entity IN (" . getEntity($objecttmp->table_element) . ")";
+					$sql .= " AND t.entity IN (" . getEntity($objecttmp->element) . ")";
 				}
 				if (!is_numeric($objecttmp->ismultientitymanaged)) {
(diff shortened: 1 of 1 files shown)


fappels:
@hregis Can you verify this. The issue causes trouble for external module (made with module builder) using multi-entity object sharing. Maybe fix this in V18 branch.

hregis:
@fappels first param of getEntity is $object->element. 

just for dictionary the first param is table name without prefix.

fappels:
@hregis So no danger to have this in V18?

hregis:
@fappels no, by default this is $object->element for the historical module do not have the same value between element and table_element

JamBalaya56562:
This looks already fixed on `develop` by #36405 ("FIX getEntity use element instead table_element", commit 7d403a2fff8c, merged into 22.0 on 2025-11-24 and merged forward to 23.0 and develop).

That PR applies the exact same one-line change as this one, in `selectForFormsList()`:

```diff
-  $sql .= " AND t.entity IN (" . getEntity($objecttmp->table_element) . ")";
+  $sql .= " AND t.entity IN (" . getEntity($objecttmp->element) . ")";
```

Current `develop`, `23.0` and `22.0` already use `getEntity($objecttmp->element)`, so this PR is now redundant. It can probably be closed as fixed. (Only `21.0` still has `table_element`, if a backport there is wanted.)

fappels:
@JamBalaya56562  Thanks for pointing this out. Backporting is not needed.

PR #30610, opened 2024-08-12
FIX - bank account proprio into owner_name

IC-Safa:
FIX the owner_name in bank account, there was an issue preventing the name of the owner to be saved


aspangaro:
In v20, it is currently not possible to register the account owner (Proprio) as the field has started to be renamed to owner_name but only partially. This is to finish the job.

PR #30613, opened 2024-08-13
Qual: Backport phan to 18.0

mdeweerd:
# Qual: Backport phan to 18.0

This PR backports phan to 18.0.
It must be merge after the backport for 19.0, which is after a "backport" to 20.0 and an "empty" PR to develop which needs to be integrated first.


mdeweerd:
I think that the outcome of travis is some hazard (no code change in the tests or htdocs).  There's been some trouble with travis I believe these last few days.

eldy:

PR #30614, opened 2024-08-13
Qual: Qual: "Backport phan to 18.0" Prepare 19.0 branch

mdeweerd:
# Qual: Qual: "Backport phan to 18.0" Prepare 19.0 branch

This PR backports phan to 19.0 / enables it in ci.
It must be merge after the "backport" to 20.0 and an "empty" PR to develop which needs to be integrated first.


eldy:

PR #30616, opened 2024-08-13
Qual: "Backport phan to 18.0" prepare develop branch

mdeweerd:
# Qual: "Backport phan to 18.0" prepare develop branch

The goal is integration in 18.0 and 19.0 as phan is already setup in 20.0, but this PR is to make it "seamless" to backport to 18.0 and 19.0.

It must be merged after the "empty" PR to develop which needs to be integrated first.


eldy:

mdeweerd:
@eldy 
The principle is to avoid the introduction of bugs in fixes while they can be avoided while the PR is being prepared and not when the PR has been propagated to the develop version.

This is not about "fixing" what is already not ok in the older versions, it's about fixing what is added in changes.

Once in the develop version, ci can be stuck for quite a long time.

All the existing notices are excluded, so only new ones are excluded.

N-2 is the recommended bug fix release, so that is a logical backport for the flow.

> we must reduce as much as possible the printing of any change on maintenance versions

So other verifications could also be inhibited: spelling checks for instance.

But I do not believe that inhibiting is the best approach.

PR #30618, opened 2024-08-13
qual: avoid double negations by introducing getDolGlobalBool()

thibdrev:
Avoid double <strike>negatives</strike> negations which appeared with PR #30537 :wink:.

There's probably a PHAN rule for that to detect the potential other cases present in the code.

### DIFF
--- a/htdocs/api/admin/index.php
+++ b/htdocs/api/admin/index.php
@@ -117,7 +117,7 @@ print "</tr>";
 
 print '<tr class="oddeven">';
 print '<td>'.$langs->trans("ApiProductionMode").'</td>';
-$production_mode = getDolGlobalString('API_PRODUCTION_MODE');
+$production_mode = getDolGlobalBool('API_PRODUCTION_MODE');
 if ($production_mode) {
 	print '<td><a class="reposition" href="'.$_SERVER['PHP_SELF'].'?action=setproductionmode&token='.newToken().'&status=0">';
 	print img_picto($langs->trans("Activated"), 'switch_on');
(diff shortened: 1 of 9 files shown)


mdeweerd:
@thibdrev 
Do note that `!(!getDolGlobalString('API_PRODUCTION_MODE'));` is not the same as `getDolGlobalString('API_PRODUCTION_MODE'));`.

There are cases where the assigned variable needs to be a boolean.

In the first case the end result is a bool, in the second a string.

Another method is to cast: `(bool) get...()`.


Before that code was something like `(!get...()) !== false` for which the result was a bool as well.

The SimplifyExpression rule suggest `!(!(...))` because it checks the upper level.  Changing like that was less errorprone than removing parentheses and adding (bool) everywhere.

thibdrev:
@mdeweerd, @eldy :
Ah ah. The devil hides in the details...

Then how about introducing getDolGlobalBool() to be cristal clear?

This would add yet another function. But its comprehension would be straightforward (for both human and PHAN). Avoiding commenting each line of ``!(!(...))`` and/or future dev to propose similar bad fix in the future.

Should the function look like this?
```
function getDolGlobalBool($key, $default = false)
{
	global $conf;
	return isset($conf->global->$key) ? (bool) $conf->global->$key : $default;
}
```

mdeweerd:
Introducing getDolGlobalBool is a good approach.  I would cast the end result so that if a developer provides a non-bool as the default that the return value is a bool.

```php
return (bool) (isset($conf->global->$key) ? $conf->global->$key : $default);
// or
return (bool) ($conf->global->$key ?? $default);
```

mdeweerd:
@thibdrev Also, FYI, when I read 'double negatives', I thought of (- (- number))``.  In fact the word to use here is `negations`.

thibdrev:
Thanks @mdeweerd. 
I've introduced getDolGlobalBool() and pushed the necessary updates. 
All checks have passed.
Should be ok now.

PR #30621, opened 2024-08-13
Align and padding in ref column

joseplluis:
# FIX Align left and padding in ref column on MRP vinci document
Add in ref column the same align (left) and padding than description column

### DIFF
--- a/htdocs/core/modules/mrp/doc/pdf_vinci.modules.php
+++ b/htdocs/core/modules/mrp/doc/pdf_vinci.modules.php
@@ -5,8 +5,9 @@
  * Copyright (C) 2010-2014 Juanjo Menent         <jmenent@2byte.es>
  * Copyright (C) 2015      Marcos García         <marcosgdf@gmail.com>
  * Copyright (C) 2017      Ferran Marcet         <fmarcet@2byte.es>
- * Copyright (C) 2018-2024  Frédéric France         <frederic.france@free.fr>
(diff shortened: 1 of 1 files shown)


htdocs/core/modules/mrp/doc/pdf_vinci.modules.php
```diff
@@ -1334,9 +1335,15 @@ public function defineColumnField($object, $outputlangs, $hidedetails = 0, $hide
 			'status' => true,
 			'width' => 35, // in mm
 			'title' => array(
-				'textkey' => 'Ref'
+				'textkey' => 'Ref',
+				'align' => 'L',
```
eldy, line 1339: Do you really need align => L ? (It should be default value, and tobe compatible with Right to Left languages, having no direction forced is better).

joseplluis: The default value is C for titles and R for values, and I suggest in code the same align and padding as used in description

 ```
		// Default field style for content
		$this->defaultContentsFieldsStyle = array(
			'align' => 'R', // R,C,L
			'padding' => array(1, 0.5, 1, 0.5), // Like css 0 => top , 1 => right, 2 => bottom, 3 => left
		);
		// Default field style for content
		$this->defaultTitlesFieldsStyle = array(
			'align' => 'C', // R,C,L
			'padding' => array(0.5, 0, 0.5, 0), // Like css 0 => top , 1 => right, 2 => bottom, 3 => left
		);
```

PR #30622, opened 2024-08-13
Fix default payment term with deposit

TitCaz:
# FIX|Fix
If default payment term has a deposit associated, it should be initialized also

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -4611,7 +4611,7 @@ class Form
 							}
 
 							return true;
-						});
+						}).trigger("change");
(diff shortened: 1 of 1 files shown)


eldy:

Marukome0743:
@eldy This PR has been superseded by #39248, which was merged into `develop` today. #39248 includes the original one-line change from this PR plus the reproduction steps that were requested above, and two additional robustness fixes (guarding options without `data-deposit_percent`, and keeping an already customized deposit percent on load).

Since the fix is now in `develop`, this PR should be safe to close.

Thanks @TitCaz for the original fix.

Issue #29353, opened 2024-04-15
Hidden extrafields of type boolean are set to 0 on update


glu000:
### Bug

Hidden extrafields of type boolean are set to 0 on update. Can only happen with extrafields in positions of commercial elements (facture, propal, commande, etc)

### Dolibarr Version

18, 19, develop

### Environment PHP

na

### Environment Database

na

### Steps to reproduce the behavior and expected behavior

1. Create an extrafield of type boolean for an invoice or proposal or order on the lines (DB: ....det_extrafields) and set visibility to 1
2. Create a new line (position) of an element where you have created the extrafield and enable the checkbox
3. Set the visibility of the extrafield created in step 1 to 0
4. Edit the line which has been created in step 2. Doesn't matter what is changed, just important to push the "save" button
5. Set the visibility of the extrafield created in step 1 to 1
6. View the line created in step 2 - Checkbox is not set (value 0)

JonBendtsen:
I see the same on Tuxgasy development Dolibarr container image.

Dolibarr Version: 20.0.0-alpha
PHP version: 8.2.7
DB version: 10.11.7-MariaDB-1:10.11.7+maria~ubu2204
OS version: Linux dolipod_develop 6.7.9-200.fc39.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Mar 6 19:35:04 UTC 2024 x86_64

json from API on the extrafield for propal lines
```
{
  "propaldet": {
    "glu000": {
      "id": "5",
      "type": "boolean",
      "label": "glu000",
      "size": "",
      "elementtype": "propaldet",
      "default": null,
      "computed": null,
      "unique": "0",
      "required": "0",
      "param": {
        "options": {
          "": null
        }
      },
      "pos": "100",
      "alwayseditable": "1",
      "perms": null,
      "list": "1",
      "printable": "1",
      "totalizable": "0",
      "langs": null,
      "help": null,
      "css": null,
      "cssview": null,
      "csslist": null,
      "fk_user_author": "1",
      "fk_user_modif": "1",
      "datec": "2024-04-15 23:17:58",
      "tms": "2024-04-15 23:17:58"
    }
  }
}
```

the proposal Json before ticking that Complementary Attribute
```
{
  "module": null,
  "id": "1",
  "newid": null,
  "entity": "1",
  "import_key": null,
  "array_options": [],
  "array_languages": null,
  "contacts_ids": [],
  "linked_objects": null,
[...]

PR #30625, opened 2024-08-13
FIX keep hidden optionals values in form not to reset extra field

lvessiller-opendsi:
FIX keep hidden optionals values in form not to reset extra field

**To reproduce**
- create an extrafield on proposal line and set visibility as 0
- create a new proposal
- add a line and set the hidden extrafield to a non empty value (such as 5)
- edit the line and save it
So the extrafield is updated and is set now to NULL

**Fix**
This fix add a hidden input field in the form on edit mode and so the optional value is kept and not reset to NULL in the database.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -8205,10 +8205,12 @@ abstract class CommonObject
 						$perms = dol_eval($extrafields->attributes[$this->table_element]['perms'][$key], 1, 1, '2');
 					}
 
-					if (($mode == 'create') && abs($visibility) != 1 && abs($visibility) != 3) {
+					$hidden = 0; // label and form input are not hidden by default
+					if (($mode == 'create') && !in_array(abs($visibility), array(1, 3))) {
 						continue; // <> -1 and <> 1 and <> 3 = not visible on forms, only on list
-					} elseif (($mode == 'edit') && abs($visibility) != 1 && abs($visibility) != 3 && abs($visibility) != 4) {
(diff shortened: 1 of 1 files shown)


eldy:
Don't understand your process to reproduce you say:
To reproduce

- create an extrafield on proposal line and set visibility as 0
- create a new proposal
- add a line and set the hidden extrafield to a non empty value (such as 5)
=> But if visibility is 0, you can't set value. Did i miss something ?

lvessiller-opendsi:
> Don't understand your process to reproduce you say: To reproduce
> 
> * create an extrafield on proposal line and set visibility as 0
> * create a new proposal
> * add a line and set the hidden extrafield to a non empty value (such as 5)
>   => But if visibility is 0, you can't set value. Did i miss something ?

Sorry, I didn't explain all.
The extra field is set by an external module using a trigger on create.
But when the user wants to modify one line, the value of the extra field is reset to NULL, but I want to keep the value set in database.

priojk:
This seems to be the same issue, raised by @glu000, with a detailed description of how to reproduce the problem: #29353

eldy:
#29353 seems different in a sense it is on "boolean" option that is managed differently than other fields.
@lvessiller-opendsi i did a test into v20 with an extrafields of type varchar, and i do not experience the trouble. The extrafields with visilibility 0 is not lost when editing a proposal line in v20.
Is your problem only on boolean fields ? or only in v18 ?

lvessiller-opendsi:
@eldy
You're right, it wasn't the origin of this trouble.
It was due to an external module.

PR #30626, opened 2024-08-13
FIX avoid from re-initializing array result on nested hook getEntity

kkhelifa-opendsi:
FIX avoid from re-initializing result on nested hook getEntity for also resArray

See already merged #27799 who manage resPrint only


lvessiller-opendsi (approved):

lvessiller-opendsi:
@rycks 
it's ok for me.

rycks (approved):
ok for me, i have to manage that strange thing into my module

Issue #28871, opened 2024-03-15
New signed status for sign all elements (Order, Intervention, Etc...)


lmag:
### Feature Request

- [x] Contract See : https://github.com/Dolibarr/dolibarr/issues/28203
- [x] #35271
- [x] Order Customer See: https://github.com/Dolibarr/dolibarr/pull/30407
- [ ] Order supplier
- [ ] Intervention
- [ ] Ticket
- [ ] Etc...

### Use case

Sign all element

### Other related issues

https://github.com/Dolibarr/dolibarr/issues/28203
https://github.com/Dolibarr/dolibarr/issues/34766
https://github.com/Dolibarr/dolibarr/issues/34040

lmag:
For " Order Customer" @evarisk-micka 

- [ ] Only add column "signed_status"
- [ ] Use only llx_onlinesignatures with a object
- [ ] Do class for add onlinesignature
- [ ] Do case for Order

evarisk-micka:
- [x] Order DATAMODEL => #30359
- [ ] Order Signature => https://github.com/Dolibarr/dolibarr/pull/30407

PR #30629, opened 2024-08-13
NEW intervention signed status update

W1W1-M:
# NEW intervention signed status update
This PR adds signed status : button, badge label, list search, ~~action trigger~~, notification, agenda event, & associated EN localization. This code is continued work for #27064 #28871 #28874

~~The new action trigger is `FICHINTER_SIGN`, this trigger doesn't make a difference between the different signed statuses.~~ A future iteration could be written for each explicit signed status (maybe using trigger context ?https://wiki.dolibarr.org/index.php/Triggers#Manage_and_create_a_new_trigger.27s_action).

On the card the signed status is integrated into the standard label badge through the `getLibStatut` method. It didn't seem resonable to add a second label badge.
On the list the signed status column uses a new method `getLibSignedStatus` in order to give user a dedicated label badge.

The `fichinter` class use constants for the different signed statues. In order to centralise the code for label localization and also to confine the different signed statuses I propose to use a `SIGNED_STATUSES` array dictionary and remove the individual constants ~~(constants not removed yet in this PR)~~.

~~This PR doesn't add any logic concerning online signing and associated status changes.~~

Card :
image

List:
new-intervention-list-signed-status

~~Notification :~~

image


eldy:
The current workflow (for status) is described in this graph:
https://wiki.dolibarr.org/index.php?title=Module_Interventions#Default_setup

We see that having the intervention signed or not is an independent status. This is above all true because we can have several level of signature. This can be done before or after intervention.
It seems it is what you have implemented by adding the column signed_status. Good.

However, to follow contributing rules, can you send all addition done into the database structure into an independent PR ?

W1W1-M:
> The current workflow (for status) is described in this graph: https://wiki.dolibarr.org/index.php?title=Module_Interventions#Default_setup
> 
> We see that having the intervention signed or not is an independent status. This is above all true because we can have several level of signature. This can be done before or after intervention. It seems it is what you have implemented by adding the column signed_status. Good.

OK, will update PR to include the possibility to set signed status when closed. It seems logical that the signed status can't be set when intervention is still a draft.
> 
> However, to follow contributing rules, can you send all addition done into the database structure into an independent PR ?

Thanks again @eldy I forgot to separate concernes. Will do another PR for action trigger db structure.

W1W1-M:
Also added signed status update when intervention is signed online. Not sure what is the best way to set the `$user` for action trigger seeing as it is a public user ? This currently sets an empty user in agenda auto action event.

eldy:
> Also added signed status update when intervention is signed online. Not sure what is the best way to set the `$user` for action trigger seeing as it is a public user ? This currently sets an empty user in agenda auto action event.

yes, for public/anonymous events, we set the create/update user as empty.

W1W1-M:
@eldy I've refactored to use `FICHINTER_MODIFY` action trigger with context information. Unneeded database structure changes have been removed. As mentioned before I've also removed individual signed status class constants in favor of an array dictionary `SIGNED_STATUSES`.

When using `FICHINTER_MODIFY` action trigger, I had to add an extra signed status: `STATUS_SIGNED_RECEIVER_ONLINE` in order to differentiate between manual(in Dolibarr) and online third party signature, for the automatic event action messages. This seemed like the cleanest way.

image

For email notification, I'm not sure how to correctly customize the `FICHINTER_MODIFY` message. I know I can add an extra string parameter `%s` in the `notify.class.php` `send` method but I didn't manage to get this to work properly. I think this is fired early in the process, before the new signed status is actually recorded or maybe not ? If action triggers are limited to CRUD operations the notification email needs to customized in ordre to prevent it for being too generic. I'd appreciate some help to better understand this please. 

Dolibarr documentation wiki has also been updated to reflect these changes.

eldy:
Lets first validate the structure change. I did not find the PR for structure change. it seems you closed it ?

W1W1-M:
> Lets first validate the structure change. I did not find the PR for structure change. it seems you closed it ?

Yes I closed #30639, it is no longer needed because it is for unnecessary action trigger. We agree to use existing `FICHINTER_MODIFY`
Other necessary structure changes have already been merged with #28874

eldy:
Ok, so let's see what is your last point.
For custom message, users can set a template. Templates are defined into home - tools - email templates.
And you decide which template to use for each trigger event is on setup of module notifications. Each selection of email template is store into TRIGGER_CODE_TEMPLATE constant.
This means, for the moment, we are able only to manage "full trigger" code, without context, but we should be able to enhance this.
So for example, we can save the constant into
INTERVENTION_MODIFY_TEMPLATE-context_signature
to offer to save email template for triggers for a limited case of trigger when context['signature'] = 1

W1W1-M:
> Ok, so let's see what is your last point. For custom message, users can set a template. Templates are defined into home - tools - email templates. And you decide which template to use for each trigger event is on setup of module notifications. Each selection of email template is store into TRIGGER_CODE_TEMPLATE constant. This means, for the moment, we are able only to manage "full trigger" code, without context, but we should be able to enhance this. So for example, we can save the constant into INTERVENTION_MODIFY_TEMPLATE-context_signature to offer to save email template for triggers for a limited case of trigger when context['signature'] = 1

Sounds good. I'll try this and if not too much change I'll update this PR accordingly.

W1W1-M:
I tried implementing a solution using your suggestion with `INTERVENTION_MODIFY_TEMPLATE-context_signature` and the notification email templates. It seems that notifications are intimately linked to the predefined action triggers and the necessary code changes to get this to work are out of scope/too much for this PR. 

I reworked my original solution using your idea with a condition in the object context `object->context['signature']`. I've added an extra string parameter to the notification message based on this condition. The signed context information only appears in the case of a signature.

Here is an example of an email notification received after an online signature by public user:

> This event is the following:
> 
> The intervention FI2408-0003 has been modified. Signed by third party

I've retested all of this code, I believe it's in a good state for merge.

htdocs/core/class/notify.class.php
```diff
@@ -1105,6 +1113,13 @@ public function send($notifcode, $object, $filename_list = array(), $mimetype_li
 						$object_type = 'ficheinter';
 						$mesg = $langs->transnoentitiesnoconv("EMailTextInterventionValidated", $link);
 						break;
+					case 'FICHINTER_MODIFY':
+						$link = '<a href="'.$urlwithroot.'/fichinter/card.php?id='.$object->id.'&entity='.$object->entity.'">'.$newref.'</a>';
+						$context_info = array_key_exists('signature', $object->context) ? $object->getLibSignedStatus() : '';
+						$dir_output = $conf->ficheinter->dir_output;
```
mdeweerd, line 1119: @hregis Recommends multidir_output

W1W1-M: Changing `$dir_output` to 
`$dir_output = $conf->ficheinter->multidir_output[$object->entity]."/".get_exdir(0, 0, 0, 1, $object);`
doesn't directly work. The PDF is no longer attached to the email notification. Maybe my code is not quite right or maybe this needs refactoring elsewhere to work ?

It seems other notifications could be refactored to make use of `multidir_output` as well. As per contributions guidelines, I think it would be best not to overload this PR, and rework your recommendation in another PR. 

Thank you for input.

mdeweerd: @hregis is in a better position than me to recommend the details of the how-to.  What I understood from his feedback in another PR was that a simple replacement as you did should work.
Best guess is that the entity property is not set, I've seen some code where a test is made to see if it is set and use 1 if it is not.

eldy: For information, multidir_output as introduce to be able to manage multicompany (1 directory for each company).
But we have now a function called getMultidirOutput() thatmust be used to replace any call of ->dir_output and ->multidir_output.
The function will return the multidir_output with more check on compatibilty.

htdocs/core/class/commonobject.class.php
```diff
@@ -10733,7 +10734,6 @@ public function setSignedStatusCommon($user, $status, $notrigger = 0, $triggerco
 			}
 
 			if (!$error) {
-				$this->status = $status;
```
eldy, line 10736: Should not be removed but replace by 
$this->signed_status = 1;

htdocs/core/class/notify.class.php
W1W1-M: Thanks will keep this in mind for future PR

htdocs/core/class/commonobject.class.php
W1W1-M: I removed this and added it to the `setSignedStatus` method of `fichinter.class.php` because the `$signed_status` property is defined in that class and not in `commonobject.class.php`. Setting it here means accessing it through magic method. If that isn't a problem I can add it back here. Maybe moving `signed_status` to `CommonObject` class could be discussed in another PR ?

PR #30634, opened 2024-08-14
FIX: PHPDoc wrong return value

altairis-noe:
# FIX: PHPDoc wrong return value
Sorry, my IDE kept underlining in red if I tried to `foreach` the result 😂


mdeweerd:
That's nice, but `Categorie[]`is better ;-).  It helps static analysis.

altairis-noe:
I beg to differ, @mdeweerd : the result is an array of category *ids*, so an int[]
I corrected the `@return` description accordingly.

eldy:
The doc in $c->containing says it return Categorie[]|int so doc in "containing" is wrong ?

altairis-noe:
@eldy : the Categorie::containing() has 'id' as third argument, which means we’ll get an array of Categorie ids, not full Categorie objects (see line 1573 of the same file)

eldy:
> @eldy : the Categorie::containing() has 'id' as third argument, which means we’ll get an array of Categorie ids, not full Categorie objects (see line 1573 of the same file)

So doc was not complete. result can be both Categorie[] and int[] (it depends on mode). I pushed a fix for this.

altairis-noe:
> The doc in $c->containing says it return Categorie[]|int so doc in "containing" is wrong ?

I think it is indeed incomplete, I’ll correct!

mdeweerd:
Yes, this happens during the process of typing php items, and in the end the documentation and checks improve and future development is faster with less errors.

mdeweerd:
Looking at the code, it can also be a list of strings.

altairis-noe:
> > @eldy : the Categorie::containing() has 'id' as third argument, which means we’ll get an array of Categorie ids, not full Categorie objects (see line 1573 of the same file)
> 
> So doc was not complete. result can be both Categorie[] and int[] (it depends on mode). I pushed a fix for this.

I added string[] in case mode is label ;)

PR #30635, opened 2024-08-14
Update ticket.lang Finnish lang Add missed "Message"

mikko80:
# FIX|Fix #[Add missed "Message"]

### DIFF
--- a/htdocs/langs/fi_FI/ticket.lang
+++ b/htdocs/langs/fi_FI/ticket.lang
@@ -285,6 +285,7 @@ OneOfTicketTrackId=Syötä yksi seurantatunnuksestasi
 ErrorTicketNotFound=Tikettiä seurantatunnuksella %s ei löytynyt!
 Subject=Aihe
 Message=Viesti
+Progression=Edistyminen
 ViewTicket=Näytä tiketti
 ViewMyTicketList=Katso omat tiketit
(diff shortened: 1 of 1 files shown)


mdeweerd:
FYI, languages need to be updated on transifex.

frederic34:
you can ask to register here https://explore.transifex.com/dolibarr-association/dolibarr/

mikko80:
I already use transifex. But those keywords missing -> not possible set translations.

mdeweerd:
The en_US files are the reference so you need to add the keys there.

I can see that it is used in ticket/card.  It's strange that it is not reported as a missing translation.

mdeweerd:
It's not reported because Progression is listed in 
dev/translation/ignore_translation_keys.lst in the section of translations ignored for ci (because this check was added while translations were already missing).

So you can update the appropriate language file under langs/en_US and remove the keys that you are adding in dev/translation/ignore_translation_keys.lst .

mdeweerd:
Also, a tip: you can remove the keys that you want to translate from `dev/translation/ignore_translation_keys.lst`, push that in your PR and you can check in the Action results or the Files Changes tab which files/locations are related to the translation so that you start to get an idea of the context.  (The ci flow annotates these issues).

eldy:
@mikko80 - I added the entry into the main.lang so it should appear in 7 days on transifex

PR #30638, opened 2024-08-14
NEW autogenerate project reference in API

anthonyvdberghe:
When posting to projects via the API it was not possible to auto-assign a reference like for third parties. 

This commit uses the assigned numbering module like in /public/project/new.php and assigns it when ref = "auto" like in the third parties API class.


eldy:
Using value -1 and auto is the good choice.
Need however to fix cti errors.

anthonyvdberghe:
Thanks, errors have been fixed.

PR #30649, opened 2024-08-16
fix ignored phpstan (most of expects string, int given)

frederic34:


htdocs/accountancy/admin/account.php
```diff
@@ -191,7 +192,7 @@
 				}
 			}
 
-			if (!dolibarr_set_const($db, 'CHARTOFACCOUNTS', $chartofaccounts, 'chaine', 0, '', $conf->entity)) {
+			if (!dolibarr_set_const($db, 'CHARTOFACCOUNTS', (string) $chartofaccounts, 'chaine', 0, '', $conf->entity)) {
```
eldy, line 195: Why not just setting dolibarr_set_const to accept both string and int ?
Wanting to store an option that is an int as does not hurt me (even if function convert it into string)...

PR #30652, opened 2024-08-16
Qual: Enable phan on modulebuilder

mdeweerd:
# Qual: Enable phan on modulebuilder
    
Most files of the modulebuilder template can be checked as regular files.
It will help improve the template


mdeweerd:
There is one remaining issue that I leave up to the merger ;-).

https://github.com/Dolibarr/dolibarr/blob/71c3c4d4c785b693e639d0b65b15401d2236bce3/htdocs/modulebuilder/index.php#L2148-L2153

mdeweerd:
@eldy
With the changes to the develop branch this is kind of a moving target, but now it's back downto just the case mentioned above that you probably instantly know the fix for https://github.com/Dolibarr/dolibarr/pull/30652#issuecomment-2294465769 .

PR #30659, opened 2024-08-17
NEW: API close proposal now supports both private and public note

JonBendtsen:
# NEW API close proposal now supports both private and public note
Adding public note to API call to close a proposal. I'm using it when my dance class registrants close their own registration.

image


JonBendtsen:
FYI: I haven't planned to add more changes, I only left it in draft because it didn't build, I wanted to later merge develop in here occasionally until it would build, and then I'd flip the ready for review.

PR #30661, opened 2024-08-17
Qual: Fix PhanUndeclaredMethod

mdeweerd:
# Qual: Fix PhanUndeclaredMethod

Fix notices PhanUndeclaredMethod


mdeweerd:
@eldy 
I think there is an error in the logic here.  It seems to me that the code in the `if ($generic != 'tcpdfbarcode')` is assuming that the opposite is true because the methods only exist on that class.

https://github.com/Dolibarr/dolibarr/blob/906c6b437fa651d1756077114581769ba3b4fe11/htdocs/barcode/printsheet.php#L182-L188

eldy:
> @eldy I think there is an error in the logic here. It seems to me that the code in the `if ($generic != 'tcpdfbarcode')` is assuming that the opposite is true because the methods only exist on that class.
> 
> https://github.com/Dolibarr/dolibarr/blob/906c6b437fa651d1756077114581769ba3b4fe11/htdocs/barcode/printsheet.php#L182-L188

I don't think there is error.
Test is to enter if generator is 'phpbarcode'. For 'tcpdfbarcode' generator, we don't need to enter.

mdeweerd:
I updated the typing hint like this:
```php
            '@phan-var-force modTcpdfbarcode|modPhpbarcode $module';
```

PR #30662, opened 2024-08-18
chore: tell about delivery objecttype tabs

lippoliv:
I just needed to lookup the code to add a tab for `delivery` view, so I added it to the comment.

This probably should also be reflected on https://wiki.dolibarr.org/index.php?title=Tabs_system?!


lippoliv:
I can offer to take some more time to research other calls in code and bring the list up to date on current developer branch? May that makes more sense?

eldy:
> I can offer to take some more time to research other calls in code and bring the list up to date on current developer branch? May that makes more sense?

Why not.
A better thing would be to add in doc/comment a cli command find/grep that can be used to get all possible values.

PR #30667, opened 2024-08-18
Qual: Fix phan notifications for compta/facture/card

mdeweerd:
# Qual: Fix phan notifications for compta/facture/card

Disable ignores for compta/facture/card.php and fix notices.

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -3228,7 +3228,6 @@ if ($action == 'create') {
 	// Load objectsrc
 	$objectsrc = null;  // Initialise
 	//$remise_absolue = 0;
-	$objectsrc = null;
 	if (!empty($origin) && !empty($originid)) {
 		// Parse element/subelement (ex: project_task)
 		$element = $subelement = $origin;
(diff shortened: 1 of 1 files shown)


mdeweerd:
@eldy 
Only a few things left.

The main question about these lines where $remise_percent  and  $dateinvoice are undefined is: should the be fetched with GETPOST or not?

https://github.com/Dolibarr/dolibarr/blob/fcfd39b4ce9c36c65a3b3d214903442c8908bf41/htdocs/compta/facture/card.php#L3240-L3246

mdeweerd:
@eldy I've ignored the PhanUndeclaredProperty notice again so that this PR can be integrated - I think that the remaining ones were introduced since I started this PR.

htdocs/compta/facture/card.php
```diff
@@ -3297,7 +3302,7 @@
 				$classname = ucfirst($subelem);
 
 				$expesrc = new $classname($db);
-				'@phan-var-force CommonObject $expesrc';
+				'@phan-var-force Facture|Commande $expesrc';
```
eldy, line 3305: I think the type of object in this case is Expedition

eldy (changes requested):

mdeweerd:
@eldy 
I added the dol_syslog statement to the code to log the class type so that you can check it yourself by visiting `compta/facture/card.php?action=create&origin=shipping&originid=2&socid=18` from the test database an check the log file contents.

Issue #30670, opened 2024-08-19
Too late invoice filter is not working in version 19.0.02 


BitKFu:
### Bug

The filtering for too late invoices is not working anymore.

Example
The start page, shows that there are 20 too late invoices.
But when I click onto the 20, all invoices are showed, that have been ever entered to the system.

### Dolibarr Version

19.0.2

### Environment Database

mysql

### Attached files

![invoiceProblem](https://github.com/user-attachments/assets/c876ea36-5eaf-4676-bdf5-2821ad771b83)

Issue #26056, opened 2023-09-28
Alert filter in more lists


leonbrehme:
### Feature Request

Currently only the invoices (for customer and form supplier) have this option. It would be nice to filter also for alerts in other lists where the alter flag is implemented (e.g. projects).

Show Case on Invoices:
![image](https://github.com/Dolibarr/dolibarr/assets/115941074/66c05c05-8ae3-412c-98c5-d51e8ce9e3d8)

Missing option on projects:
![image](https://github.com/Dolibarr/dolibarr/assets/115941074/7a7b8893-3986-4e02-82f3-0c4beb94be49)

### Use case

Show all items with the alert flag in the list.

### Suggested implementation

Add the filter option in the following lists:

- [ ] projects
- [ ] proposals
- [ ] sales orders
- [ ] purchase orders
- [ ] Contracts / Subscriptions

lists with out the alter (but I would make sense):
- [ ] vendor proposals (altert when delivery date is in the past)

Issue #25963, opened 2023-09-19
Filter for late proposals ,projects etc. is not working properly 


pmr94209:
### Bug

If you click at the mainpage in the different boxes shwoing the number of items for example Proposals on the yellow marked box for late items the filter won´t work except for suplier invoices and invoices

leonbrehme:
Hi, thanks for the issues but could you provide at least the version of your Dolibarr installation?

pmr94209:
Yes , sorry the version is 18.0.1.

leonbrehme:
I'm sorry. I tried to reproduce the error, but it was not possible. The filters are applied correctly.

pmr94209:
If you go on the dolibarr demo at dolibarr.org https://demo.dolibarr.org/public/demo/ an login and the click on the startpage the late projects, which are curretly 92 and the click on the number 92 the opcoming page shows 540 projects and not just the 92 which are overdue , see enclosed pics 
<img width="2353" alt="Bildschirmfoto 2023-09-22 um 16 23 28" src="https://github.com/Dolibarr/dolibarr/assets/38521120/7dbebc66-046c-4318-bf63-5d8042899db4">
<img width="2345" alt="Bildschirmfoto 2023-09-22 um 16 23 18" src="https://github.com/Dolibarr/dolibarr/assets/38521120/3f1194a0-4769-43fc-bbb3-353f215f06d5">
<img width="2353" alt="Bildschirmfoto 2023-09-22 um 16 23 04" src="https://github.com/Dolibarr/dolibarr/assets/38521120/62d51127-f4cd-4f45-9e0a-efbf0ac068b0">

leonbrehme:
I've investigated: your were right. It's not working for projects because they don't have the option to filter for overdue (today after enddate). For the invoices there is a filter option. So maybe this needs to be feature request.

![Screenshot 2023-09-22 16 50 28](https://github.com/Dolibarr/dolibarr/assets/115941074/dcbfbe93-fb75-41a7-a0a3-16c3e0957ef3)

pmr94209:
In the past versions 11.x it did work for example for quotations or orders  which are overdue. Now it does only work for invoices anymore

leonbrehme:
Do you know at witch version it stops? As fare as I now, there was a change in the way there filters were built. Maybe this feature was lost during the harmonization. Right know I don't have any other idea.

[...]

PR #30687, opened 2024-08-20
FIX late propal search option (v18+)

W1W1-M:
# FIX late propal search option (v18+)
Like invoices #30670, the home page link to late propal does not work correctly. Currently the link will open the propal list with all **open** propals, the late alert duration is not used.
This PR adds a `search_option` post variable for the home page link as well as a alert checkbox on the propal list.

fix-late-propal-list
fix-late-propal-list-alert

Associated issues: #25963 #26056

Merge tip: from v20+ `object_statut` no longer exists

### DIFF
--- a/htdocs/comm/propal/class/propal.class.php
+++ b/htdocs/comm/propal/class/propal.class.php
@@ -3517,7 +3517,7 @@ class Propal extends CommonObject
 			$response->label = $label;
 			$response->labelShort = $labelShort;
 			$response->url = DOL_URL_ROOT.'/comm/propal/list.php?search_status='.$status.'&mainmenu=commercial&leftmenu=propals';
-			$response->url_late = DOL_URL_ROOT.'/comm/propal/list.php?search_status='.$status.'&mainmenu=commercial&leftmenu=propals&sortfield=p.datep&sortorder=asc';
+			$response->url_late = DOL_URL_ROOT.'/comm/propal/list.php?search_option=late&mainmenu=commercial&leftmenu=propals&sortfield=p.datep&sortorder=asc';
 			$response->img = img_object('', "propal");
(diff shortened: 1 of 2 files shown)


rycks:
hello @W1W1-M your PR seems really good but i think (@eldy could be more precise) there is a problem with dolibarr policy : please split your PR in two parts:

1. one PR with part to fix the bug (modify the url filters to display -> 18.0 impact is confirmed)
2. an other PR with alert checkbox (this is a new feature) -> that PR will be made on develop dolibarr branch

Like you did it https://github.com/Dolibarr/dolibarr/issues/25963 https://github.com/Dolibarr/dolibarr/issues/26056 for alert new feature

thanks,

W1W1-M:
> hello @W1W1-M your PR seems really good but i think (@eldy could be more precise) there is a problem with dolibarr policy : please split your PR in two parts:
> 
> 1. one PR with part to fix the bug (modify the url filters to display -> 18.0 impact is confirmed)
> 2. an other PR with alert checkbox (this is a new feature) -> that PR will be made on develop dolibarr branch
> 
> Like you did it #25963 #26056 for alert new feature
> 
> thanks,

Hello,
Thanks for feedback. I agree with Dolibarr policy, however the alert checkbox feels to me to be intrinsically linked to the warning on dashboard: 

If a user opens the warning on the dashboard to get to the list of concerned propals, they will also need a way to un-filter the list. I suppose using the reset filters would be enough, the added logic in the list is still required. The added checkbox is only a few lines of code more. I can remove it, please confirm if : "This is the way" ⭐ ⚔️

lvessiller-opendsi (approved):

rycks (approved):
ok for me

PR #28157, opened 2024-02-13
Fix: Add array access to CommonObject

mdeweerd:
# Fix: Add array access to CommonObject

Phan detects array access to subclasses of CommonObject (for instance: Societe). This adds array access to properties of CommonObjects.

PR #30691, opened 2024-08-20
Qual: Bugs/Remove exceptions on phan rules disabled for ci

mdeweerd:
# Qual: Bugs/Remove exceptions on phan rules disabled for ci

Some phan exceptions were added to allow several pull requests to be merged without breaking the flow.
However, they hide real issues that can not just be 'fixed' by type hinting or small rewrites.

This PR is there to highlight them to the core team for fixing.

Summary:
- Looping over Adherent as an array while #28157 was rejected;
- Numbering class for Availabilities/Calendar not defined but code is expecting one;
- DolibarrModules->id expected, but not existing;
- (pdf_eagle.modules.php -> Disabled PhanUndeclaredProperty again: requires more in-depth effort to fix.  For instance, proprerties from Expedition/Reception are expected on StockTransfer, and others on StockTransferLines but do not exist).


mdeweerd:
It looks like blockedlog is accessing CommonObject properties as if CommonObject is an array.  Possibly silently ignored on execution.

I disabled the ignore for those cases as well.

PR #30692, opened 2024-08-20
FIX late order search option (v18+)

W1W1-M:
# FIX late order search option (v18+)
Like invoices #30670, the home page link to late orders does not work correctly. Currently the link will open the order list with all validated+in process orders, the late alert duration is not used.
This PR adds a `search_option` post variable & `url_late` for the home page link as well as a alert checkbox on the order list.

fix-late-order-list
fix-late-order-list-alert

Associated issues: #25963 #26056


lvessiller-opendsi:
Could you separate the FIX content and NEW content into 2 PRs ? (one on 18.0 and one on develop)

lvessiller-opendsi (approved):

lvessiller-opendsi:
@rycks 
it's ok for me

rycks (approved):
ok for me

PR #30694, opened 2024-08-20
FIX : attached file on first page load

atm-jpb:
## Bug Description

When creating a new ticket, if the user adds an attachment without creating the ticket, leaves the page, and then returns to the ticket creation page, the previously added attachment is still present. This behavior is undesirable as it can lead to confusion for the user and potentially result in the addition of irrelevant attachments to a new ticket.

## Proposed Solution

To resolve this issue, we will check if the page is loaded for the first time by testing for the presence of a token in the $_POST. If this is the case, we will clean up the temporary attachments associated with the user's session.

### DIFF
--- a/htdocs/ticket/card.php
+++ b/htdocs/ticket/card.php
@@ -738,7 +738,21 @@ llxHeader('', $title, $help_url);
 
 if ($action == 'create' || $action == 'presend') {
 	$formticket = new FormTicket($db);
+	if (!GETPOSTISSET('token')){
+		// we are on create mode for the first time
+		//  we need to remove any attached files in session
+		if (!empty($_SESSION["listofpaths"])) {
+			unset($_SESSION["listofpaths"]);
+		}
 
+		if (!empty($_SESSION["listofnames"])) {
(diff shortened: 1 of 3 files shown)


htdocs/ticket/card.php
```diff
@@ -738,7 +738,21 @@
 
 if ($action == 'create' || $action == 'presend') {
 	$formticket = new FormTicket($db);
+	if (!GETPOSTISSET('token')) {
```
eldy, line 741: In other part of code, we used a different solution:
The parameter mode=init was added on the link with action=presend
And we reset dir and session when mode=init. For eample in htdocs/admin/mails.php
```
		// Init list of files
		if (GETPOST("mode", "aZ09") == 'init') {
			$formmail->clear_attached_files();
		}

```
When mode is not init, we untouch dir

atm-jpb: I have taken your comment into account to use mode=init as suggested. This results in a modification of modTiket and an SQL update. Let me know if this works for you.

atm-jpb: @eldy

PR #30696, opened 2024-08-20
fix cast

frederic34:


htdocs/core/tpl/extrafields_list_print_fields.tpl.php
```diff
@@ -43,7 +43,7 @@
 				// If field is a computed field, we make computation to get value
 				if ($extrafields->attributes[$extrafieldsobjectkey]['computed'][$key]) {
 					$objectoffield = $object; //For compatibility with the computed formula
-					$value = dol_eval((int) $extrafields->attributes[$extrafieldsobjectkey]['computed'][$key], 1, 1, '2');
+					$value = (int) dol_eval((string) $extrafields->attributes[$extrafieldsobjectkey]['computed'][$key], 1, 1, '2');
```
eldy, line 46: May be we must remove completely the int ? What if the computed field is to forge a string for a varchar extrafield ?

PR #30697, opened 2024-08-20
Fix #30695

### LINKED ISSUE #30695: Disabled php functions are not detected if there are spaces in php.ini
### Bug

If php.ini directive `disable_functions` contains spaces, Dolibarr admin security page `admin/system/security.php` will not detect that the functions are disabled

### Dolibarr Version

18, 19, 20, develop

### Environment PHP

_No response_

### Environment Database

_No response_

Add spaces after commas in php.ini `disable_functions` list and display `admin/system/security.php page`

### Attached files

_No response_

hansemschnokeloch:
# FIX #30695
To avoid incorrect detection of disabled php functions, we use php native `function_exists`.

### DIFF
--- a/htdocs/admin/system/security.php
+++ b/htdocs/admin/system/security.php
@@ -149,7 +149,7 @@ print "<br>\n";
 $todisabletext = '';
 $i = 0;
 foreach ($arrayoffunctionstodisable as $functiontodisable) {
-	if (! in_array($functiontodisable, $arrayoffunctionsdisabled)) {
+	if (\function_exists($functiontodisable)) {
 		if ($i > 0) {
 			$todisabletext .= ', ';
 		}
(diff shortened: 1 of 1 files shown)


eldy:
Can you provide an example on how to reproduce the trouble ?

hregis:
@hansemschnokeloch it's getting really complicated for me to understand why a "function_exists" could become more efficient than an "in_array" in a list of names? Have you taken drugs too? ;-)

hansemschnokeloch:
> Can you provide an example on how to reproduce the trouble ?

If you use spaces like below for more readability in php.ini
`disable_functions = "pcntl_alarm, pcntl_fork, pcntl_waitpid, pcntl_wait"`
instead of
`disable_functions = "pcntl_alarm,pcntl_fork,pcntl_waitpid,pcntl_wait"`
you will end with spaces at the begining of the names or the names of the functions in the array `$arrayoffunctionsdisabled` after using `explode`

hansemschnokeloch:
> @hansemschnokeloch it's getting really complicated for me to understand why a "function_exists" could become more efficient than an "in_array" in a list of names? Have you taken drugs too? ;-)

With the in_array method you're checking what is declared in php.ini, you're not checking if a function exists or not.
For example pcntl_* functions don't exist on windows, so you don't need to disable them in php.ini. If you check only what's in php.ini you will then have a false positive.
You can also use a php flavor compiled without certains functions, and checking what's in php.ini is then not accurate.

PR #30701, opened 2024-08-20
Fix website htmlheader content may contain more tags

Daoud-mohamed:
# Fix
Must accept tags like '<script>' and '<link.>'

### DIFF
--- a/htdocs/website/index.php
+++ b/htdocs/website/index.php
@@ -1594,7 +1594,7 @@ if ($action == 'updatecss' && $usercanedit) {
 			}
 
 
-			$dataposted = trim(GETPOST('WEBSITE_HTML_HEADER', 'restricthtmlallowunvalid'));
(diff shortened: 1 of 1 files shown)


htdocs/website/index.php
```diff
@@ -1594,7 +1594,7 @@
 			}
 
 
-			$dataposted = trim(GETPOST('WEBSITE_HTML_HEADER', 'restricthtmlallowunvalid'));
+			$dataposted = trim(GETPOST('WEBSITE_HTML_HEADER', 'none'));		// Must accept tags like '<script>' and '<link>'
```
eldy, line 1597: Because the none is not clear of what we can sanitize or not, i have introduced a new possible parameter 'restricthtmlallowlinkscript' ?
Is it ok with it indtead of none on your side ?

Daoud-mohamed: Yes, it works well. I sent a new PR to apply the 'restrictHtmlAllowLinkScript' parameter.

PR #30705, opened 2024-08-21
Qual: Fix caching for windows workflow

mdeweerd:
# Qual: Fix caching for windows workflow

actions/cache/v3 and v4 do not work on windows, but there are new suggestions to try to make it work.
The main approach is to prefix the cached paths with './' and have exactly the same content for 'path:'.
See https://github.com/actions/cache/issues/1361#issuecomment-2015044001

Success: the cache was successfully restored in this PR after pushing the rebased branch - about 4 minutes gained for the windows action!

### DIFF
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -65,8 +65,8 @@ jobs:
           # See https://github.com/actions/cache/issues/1275#issuecomment-1925217178
           enableCrossOsArchive: true 
           path: |
-            db_init.sql
-            db_init.sql.md5
+            ./db_init.sql
+            ./db_init.sql.md5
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-ci.yml
```diff
@@ -61,19 +61,22 @@ jobs:
       - name: Restore cache
         id: cache
         uses: actions/cache/restore@v4
+        env:
+          HASH: ${{ hashFiles('htdocs/instll/**') }}
```
eldy, line 65: Do you mean htdocs/install ?

mdeweerd: Yes.

mdeweerd: I pushed an update with the fix.

PR #30711, opened 2024-08-21
qual: PHPStan for tva.class.php

thibdrev:
htdocs/compta/tva/class/tva.class.php	57	Property Tva::$total has no type specified.

htdocs/compta/tva/class/tva.class.php	59	Property Tva::$datep has no type specified.

htdocs/compta/tva/class/tva.class.php	60	Property Tva::$datev has no type specified.

htdocs/compta/tva/class/tva.class.php	61	Property Tva::$amount has no type specified.

htdocs/compta/tva/class/tva.class.php	62	Property Tva::$type_payment has no type specified.

htdocs/compta/tva/class/tva.class.php	924	Method Tva::getKanbanView() has parameter $arraydata with no value type specified in iterable type array.

### DIFF
--- a/htdocs/compta/tva/class/tva.class.php
+++ b/htdocs/compta/tva/class/tva.class.php
@@ -69,7 +69,7 @@ class Tva extends CommonObject
 	public $datev;
 
 	/**
-	 * @var float VAT amount
+	 * @var float|string VAT amount
 	 */
 	public $amount;
 
@@ -163,7 +163,7 @@ class Tva extends CommonObject
 		$now = dol_now();
 
 		// Clean parameters
-		$this->amount = (float) $this->amount;
+		$this->amount = (float) price2num($this->amount);
 		$this->label = trim($this->label);
 		$this->type_payment = (int) $this->type_payment;
 		$this->note = trim($this->note);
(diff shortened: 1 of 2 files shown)


htdocs/compta/tva/class/tva.class.php
```diff
@@ -163,7 +163,7 @@ public function create($user)
 		$now = dol_now();
 
 		// Clean parameters
-		$this->amount = trim($this->amount);
+		$this->amount = (float) $this->amount;
```
mdeweerd, line 166: I've been "PHPDoc typing" in several cases that monetary values can be either string or float because price2num will ensure that they have the desired count of digits after the decimal dot.

thibdrev: Yes, that's what I figured out with the phan warnings.
So I've updated to:
```
	/**
	 * @var float|string VAT amount
	 */
	public $amount;
```
and
``` 
// Clean parameters
$this->amount = (float) price2num($this->amount);
```

mdeweerd: Actually, you can keep 
```
$this->amount = price2num($this->amount);
```

And whenever $this->amount is used in a computation, cast it with (float).

It would be better to have a class represent monetary and quantity values IMHO, but that's a task in itself.  That class would be able to handle all the operators +, -, /, ... .

PR #30720, opened 2024-08-22
Fix #29292

### LINKED ISSUE #29292: The colon char is breaking sqlfilters search
### Bug

If you try :
```
http://my-dolibarr.local/api/index.php/orders?sortfield=t.rowid&sortorder=ASC&limit=100&sqlfilters=t.ref_client%3A%3D%3A'a%3Ab'
```
With sqlfilters which is `t.ref_client:=:'a:b'`, the `:` make everything breaks. With this filter, it find order with those client ref: "LAvionJaune", "test", "htrh", "Mathis55109". I don't understand why or how it works, but it's very strange.

### Dolibarr Version

18.0.5

### Environment PHP

8.1.27

### Environment Database

MariaDB 10.11.6

Call API with this : `api/index.php/orders?sortfield=t.rowid&sortorder=ASC&limit=100&sqlfilters=t.ref_client%3A%3D%3A'a%3Ab'`, in general with a `:` in the ref client

### Attached files

_No response_

hansemschnokeloch:
# FIX #29292
Allowing colon char `":"` in sqlfilters is done by limiting results from explode

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -13591,7 +13591,7 @@ function dolForgeCriteriaCallback($matches)
 	if (empty($matches[1])) {
 		return '';
 	}
-	$tmp = explode(':', $matches[1]);
+	$tmp = explode(':', $matches[1], 3);
 	if (count($tmp) < 3) {
 		return '';
 	}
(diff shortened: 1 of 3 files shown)


hansemschnokeloch:
@eldy is this phpunit test correct ?
https://github.com/Dolibarr/dolibarr/blob/3dbd79dad83bf9a0e3150be15957666122fe2b24/test/phpunit/FunctionsLibTest.php#L239-L242

Why should `(t.date_creation:<:'2016-01-01 12:30:00')` be transformed into `(t.date_creation < 0)` ?
What is the meaning of a negative date ?

hansemschnokeloch:
I added a fix for the phpunit test

PR #30724, opened 2024-08-22
#FIX : datenextrun for CRON

atm-GregM:
#FIX : datenextrun for CRON

Fix on new entry CRON, the next run is not included.


htdocs/core/modules/DolibarrModules.class.php
```diff
@@ -1572,6 +1576,9 @@ public function insert_cronjobs()
 							if (is_int($priority)) {
 								$sql .= "'".$this->db->escape($priority)."', ";
 							}
+							if (!empty($datenextrun)) {
+								$sql .= "'".$this->db->escape($datenextrun)."', ";
```
eldy, line 1580: For a date, we must use
$this->db->idate()

atm-GregM: Replace by idate() :+1:

PR #30733, opened 2024-08-22
Qual: Fix some phan typing in actions_massactions.inc.php

mdeweerd:
# Qual: Fix some phan typing in actions_massactions.inc.php

Fix phan in actions_massactions.inc.php

### DIFF
--- a/htdocs/core/actions_massactions.inc.php
+++ b/htdocs/core/actions_massactions.inc.php
@@ -49,6 +49,18 @@
 @phan-var-force ?string $deliveryreceipt
 ';
 
+'
+@phan-var-force ?string $permissiontoread
+@phan-var-force ?string $permissiontodelete
+@phan-var-force ?string $permissiontoclose
+@phan-var-force ?string $permissiontoapprove
+@phan-var-force ?int[] $toselect
+@phan-var-force ?string $diroutputmassaction
+@phan-var-force ?string $objectlabel
+@phan-var-force ?string $option
+@phan-var-force ?string $deliveryreceipt
+';
+
 
 // Protection
 if (empty($objectclass) || empty($uploaddir)) {
--- a/dev/tools/phan/baseline.txt
+++ b/dev/tools/phan/baseline.txt
@@ -248,7 +248,6 @@ return [
         'htdocs/core/actions_extrafields.inc.php' => ['PhanTypeMismatchArgumentProbablyReal'],
         'htdocs/core/actions_lineupdown.inc.php' => ['PhanPossiblyUndeclaredGlobalVariable', 'PhanUndeclaredGlobalVariable'],
         'htdocs/core/actions_linkedfiles.inc.php' => ['PhanPossiblyUndeclaredGlobalVariable', 'PhanUndeclaredGlobalVariable'],
-        'htdocs/core/actions_massactions.inc.php' => ['PhanPossiblyUndeclaredGlobalVariable', 'PhanUndeclaredProperty'],
         'htdocs/core/actions_printing.inc.php' => ['PhanUndeclaredProperty'],
         'htdocs/core/actions_sendmails.inc.php' => ['PhanPluginUndeclaredVariableIsset', 'PhanPossiblyUndeclaredGlobalVariable', 'PhanUndeclaredGlobalVariable', 'PhanUndeclaredProperty'],
         'htdocs/core/actions_setmoduleoptions.inc.php' => ['PhanPluginUnknownObjectMethodCall', 'PhanPossiblyUndeclaredGlobalVariable'],


htdocs/core/actions_massactions.inc.php
```diff
@@ -47,6 +47,13 @@
 }
 $error = 0;
 
+// Note: list of strings for objectclass could be extended to accepted/expected classes
+'
+@phan-var-force "Societe"|"Contact"|"ExpenseReport"|"Partnership"|"Holiday"|"ConferenceOrBoothAttendee"|"Facture"|"CommandeFournisseur" $objectclass
```
eldy, line 52: $objectclass can be any Dolibarr business object here.
Is it possible to have instead a generic object class, may be CommonObject class ?

mdeweerd: I find often that putting CommonObject alone generates issues with the typing as some fields are missing and expected by the algorithm.

I propose to "add" CommonObject and also leave the other options that are tested for in the code.  In case there is a typing issue it will indicate which classes are not compatible.

Ideally - in some future - the code would do a test using "instanceof" rather than relying on the 'element' field and the typing hints.
That implies risks of course: the "instanceof" condition could be too strict vs. what the code was actually expecting and introduce some bugs in the process.

eldy: > I propose to "add" CommonObject and also leave the other options that are tested for in the code. In case there is a typing issue it will indicate which classes are not compatible.

You mean having this
@phan-var-force CommonObject|"Societe"|"Contact"|"ExpenseReport"|"Partnership"|"Holiday"|"ConferenceOrBoothAttendee"|"Facture"|"CommandeFournisseur" $objectclass   ?

mdeweerd: Yes.

FYI, phan has more than 50 notices for this file.

For instance, we have:
```
htdocs\core\actions_massactions.inc.php:1288 PhanUndeclaredProperty Reference to undeclared property \CommonObject->price
htdocs\core\actions_massactions.inc.php:1289 PhanUndeclaredProperty Reference to undeclared property \CommonObject->price_min
```

There is no check on the object type: it is implied by the 'updateprice' action.

mdeweerd: That is because we have the following in the configuration for phan:

```
dev/tools/phan/config.php:              'obj' => '\CommonObject',     // Deprecated
```` 

We could check what the result is if this is commented in a separate PR (I guess there will be some notifications that pop up).

**Comment by mdeweerd** [comment]
I've now extended to fixing all the notices in `htdocs/core/actions_massactions.inc.php`.

Some observations:
- The file arguments to `pdftk` use '/' which is not system dependent.  Maybe this works for `pdftk` on Windows but technically it's better that this is the path according to the system rules;
- I'ld rather have a `dol_exec` rather than the native `exec` to allow the addition of security checks.  `dol_exec` would take an `string[][]` as the argument with each item being a command to execute in the format `array{command,arg1,arg2,...}`.
 `dol_exec` would take care of escaping those and executing them in order (either by separating them with ';' or executing one after the other.
Also - not verified in detail - according to a summary for the following information, `escapeshellarg` was designed for bash/... not a windows shell, so that's a potential security issue: https://gist.github.com/Zenexer/40d02da5e07f151adeaeeaa11af9ab36 .  And therefore a good reason to have a dol_exec taking care of escaping (or refusing execution) .

Regarding the issues:
- fk_soc and socid on common object not defined:  not fixed because when that happens the type of CommonObject does not seem to be limited a a specific set of child classes;
- $pagecount possibly undefined - not fixed because the use seems invalid in case there is more than one document - it uses the pagecount of the last document;
- Issue regarding ModeleNumRefTask: not for me to decide;
- Other cases: ~I am looking into them.~ @eldy These are now fixed - your turn to "play".

**Comment by mdeweerd** [comment]
I've reduced this PR to only the changes in actions_massactions.inc.php .   The rest of the PR was moved to #30747

File: htdocs/core/actions_massactions.inc.php [hunk 2]
```diff
@@ -49,6 +49,18 @@
 @phan-var-force ?string $deliveryreceipt
 ';
 
+'
+@phan-var-force ?string $permissiontoread
```
eldy, line 53: Seems duplicate of line 41 to 49

eldy (changes requested):

htdocs/core/actions_massactions.inc.php
mdeweerd: Possibly because this PR and changes still pending here are "old" and repeated in more recent PRs.  I'll clean that up.

The main reason this PR is still open is for these phan notices:
```plaintext
Warning: actions_massactions.inc.php: PhanUndeclaredProperty: Reference to undeclared property \CommonObject-&gt;fk_soc
Warning: actions_massactions.inc.php: PhanUndeclaredProperty: Reference to undeclared property \User-&gt;email_aliases
Warning: actions_massactions.inc.php: PhanPossiblyUndeclaredGlobalVariable: Global variable $pagecount is possibly undeclared
```

I am tending to think that I should simply extract the code changes to a separate PR and close this one - the real technical debt is likely not going to be fixed until almost all of the pending phpdoc changes and false positives are identified.

PR #30744, opened 2024-08-23
Add includetimespent in index method for tasks

kadogo:
# NEW|New Add includetimespent in index method for tasks

There were no time spent option when using the /tasks endpoint. It is not possible to use the option 0 or 1.
I have tested it only on 14.0.5 because it is our actual instance.

Regards

### DIFF
--- a/htdocs/projet/class/api_tasks.class.php
+++ b/htdocs/projet/class/api_tasks.class.php
@@ -103,9 +103,10 @@ class Tasks extends DolibarrApi
 	 * @param int			   $page				Page number
 	 * @param string           $sqlfilters          Other criteria to filter answers separated by a comma. Syntax example "(t.ref:like:'SO-%') and (t.date_creation:<:'20160101')"
 	 * @param string    $properties	Restrict the data returned to these properties. Ignored if empty. Comma separated list of properties names
+   * @param   int         $includetimespent       0=Return only task. 1=Include a summary of time spent, 2=Include details of time spent lines (2 is no implemented yet)
 	 * @return  array                               Array of project objects
 	 */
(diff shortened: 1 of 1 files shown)


kadogo:
Hello @eldy 

Is there a way to easily fix the space / tabs issue because all other projects I know are using spaces, so I will not change my editor only for Dolibarr, sorry.

If there is a kind of command, or something like that, I would be more than happy to run it.

Regards

frederic34:
@kadogo what is your editor ? you can use plugin editorconfig if your editor support it.

kadogo:
Hello @frederic34 

I use geany but configured to convert tabs in 4 spaces.

frederic34:
image
with vscode and this plugin, editor is auto configured to use .editorconfig file at the root of the project

frederic34:
https://github.com/editorconfig/editorconfig-geany

kadogo:
Thanks, I will take a look to install it so I will avoid these CI issues.

JamBalaya56562:
For information: I have opened #39441, which implements this feature on current develop.

It follows the same approach as here — `includetimespent` on `TaskApi::index()` — implemented like the existing `get()`: `1` calls `Task::getSummaryOfTimeSpent(0)` (fills the `timespent_*` fields) and `2` calls `Task::fetchTimeSpentOnTask()` (fills `lines`). The parameter is appended at the end of the signature and defaults to `0`, so existing callers are unaffected.

Credit for the idea and the original implementation goes to @kadogo — this PR is mentioned in #39441.

If #39441 is merged, this one can be closed.

eldy:
Replaced by #39441

PR #30746, opened 2024-08-23
# NEW|New UTF-8 General CI Collation for Database Tables

ems-co:
## Description
In an effort to standardize the database collation and improve compatibility, the following changes have been made:
- Updated the collation of tables in the database to `utf8_general_ci`.
- Ensured that all text columns use the `utf8_general_ci` collation.

These changes are intended to ensure consistent handling of UTF-8 characters throughout the Dolibarr application and to enhance compatibility with a wide range of languages and symbols, including Arabic and Kurdish. Previously, text written in these languages in the “Notes Public” and “Notes Private” fields was being recorded as question marks (???????) in the database. By updating the collation to utf8_general_ci, this issue will be resolved, allowing for accurate representation and storage of multilingual content.

This update aligns with best practices for character encoding and collation settings in modern applications.

## Testing
Please review the changes to ensure that the collation update does not affect existing functionality.


sonikf:
Hi @ems-co
 
My language is Greek, which falls in the same category as yours. 

Dolibarr already supports utf8 by default and utf8_unicode_ci, until recently, the recommended settings  for character encoding and collation.

Any problem you may have encountered is in your underlying database(also take a look at the wysiwyg module settings)
My recommendation  to you is to use utf8mb4_unicode_ci as it is the new recommended character encoding and collation settings.

I have been using it for 3years now and solved all my problems with ????s stored in dbs plus emojis and mathematical symbols cause it allows 4-byte UTF-8 input to be stored correctly. 

see
https://dev.mysql.com/doc/refman/8.4/en/charset-unicode-sets.html

@eldy should we make utf8mb4_unicode_ci default?

hregis:
@ems-co I don't understand your choice! "unicode" is actually better for special characters! Do a search for "unicode vs general" and you'll get the answer.

frederic34:
leaving empty will use the default collation of the db which is the better choice and let users able to choice for what they want (utf8mb4_unicode_ci...)

image

PR #30747, opened 2024-08-23
Qual: Fix phan issues related to getNextValue() notices

mdeweerd:
# Qual: Fix phan issues related to getNextValue() notices

Separated from #30733 (changes to actions_massactions.inc.php removed).


htdocs/projet/contact.php
```diff
@@ -105,7 +105,7 @@
 			foreach ($task_array as $task) {
 				$task_already_affected = false;
 				$personsLinked = $task->liste_contact(-1, $source);
-				if (!is_array($personsLinked) && count($personsLinked) < 0) {
+				if (!is_array($personsLinked) || count($personsLinked) == 0) {
```
eldy, line 108: I think what we wanted to do is:
If not array && < 0
Not array test is to avoid to compare array with int and < 0 is to detect if liste_,contact was in error.
Empty array is not an error case because object can have no contact at all.

mdeweerd: OK, I'll remove count() then.

mdeweerd: @eldy I modified the code.

PR #30757, opened 2024-08-25
fix miscellaneous payment stays in draft status

sonikf:
before

after

### DIFF
--- a/htdocs/compta/bank/class/paymentvarious.class.php
+++ b/htdocs/compta/bank/class/paymentvarious.class.php
@@ -179,7 +179,12 @@ class PaymentVarious extends CommonObject
 	/**
 	 * Draft status
 	 */
-	const STATUS_DRAFT = 0;
+	//const STATUS_DRAFT = 0;
+
+	/**
+	 * Paid status
+	 */
+	const STATUS_PAID = 1;
 
 
 	/**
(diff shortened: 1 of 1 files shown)


htdocs/compta/bank/class/paymentvarious.class.php
```diff
@@ -631,14 +636,16 @@ public function LibStatut($status, $mode = 0)
 		global $langs;
 
 		if (empty($status)) {
-			$status = 0;
+			$status = 1;
```
eldy, line 639: Looks strange.
If value is 0 the you set it to 1 ?

eldy (changes requested):

htdocs/compta/bank/class/paymentvarious.class.php
```diff
@@ -179,7 +179,12 @@ class PaymentVarious extends CommonObject
 	/**
 	 * Draft status
 	 */
-	const STATUS_DRAFT = 0;
+	//const STATUS_DRAFT = 0;
```
eldy, line 182: Why removing the status 0 / Draft ?

eldy:
I pushed a different fix. The pb is that paylents have no status. It just exists or not, so i made the same behaviour than invoice payment: no status visible.

PR #30763, opened 2024-08-26
Fix board info box list links menus

W1W1-M:
# Fix board info box list links menus
This PR fixes the main & left menus shown when opening the info boxes links on home page board. Currently when opening each list from the badge on board info boxes will leave the main & left menus as default (home). The benefit for end user is improved navigation experience.

### DIFF
--- a/htdocs/core/boxes/box_clients.php
+++ b/htdocs/core/boxes/box_clients.php
@@ -77,7 +77,7 @@ class box_clients extends ModeleBoxes
 		$thirdpartystatic = new Client($this->db);
 
 		$this->info_box_head = array(
-			'text' => $langs->trans("BoxTitleLastModifiedCustomers", $max).'<a class="paddingleft" href="'.DOL_URL_ROOT.'/societe/list.php?type=c&sortfield=s.tms&sortorder=DESC"><span class="badge">...</span></a>',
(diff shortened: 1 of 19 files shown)


htdocs/core/boxes/box_clients.php
```diff
@@ -77,7 +77,7 @@ public function loadBox($max = 5)
 		$thirdpartystatic = new Client($this->db);
 
 		$this->info_box_head = array(
-			'text' => $langs->trans("BoxTitleLastModifiedCustomers", $max).'<a class="paddingleft" href="'.DOL_URL_ROOT.'/societe/list.php?type=c&sortfield=s.tms&sortorder=DESC"><span class="badge">...</span></a>',
+			'text' => $langs->trans("BoxTitleLastModifiedCustomers", $max).'<a class="paddingleft" href="'.DOL_URL_ROOT.'/societe/list.php?mainmenu=companies&leftmenu=customers&type=c&sortfield=s.tms&sortorder=DESC"><span class="badge">...</span></a>',
```
eldy, line 80: Boxes can be used on any area page, in future we may see boxes into home page if stock or memership deoending on user choice. It means user has already chosen a menu context ti work in, when boxes is shown. Adding such param will break the environment where user was. Changing an envirinment of menu lust 'ot becsometimes whenbyser ask it with lenu and sometimes automatic without user consent. Thus is an awful user experience. So we have choice of doing always or never. It appears with user feedback over time that never was a preferred choice. So we must keep this rule. Changing context if work must be fone explicitely by me'u selection. Don't forget that clink on link is often to see temporary information and user the do back once info has been read.
If we change context without user consent, we break also this common behaviour.

PR #30765, opened 2024-08-26
QUAL Update invoice status in API to be consistent with invoice class

hansemschnokeloch:
# QUAL 
Invoice status in REST API is not consistent with invoice class.
With this PR we now use the constant names used in the invoice class to be more consistent.

Warning : this PR will certainly break applications that use this api feature (to be discussed)


eldy:
Looks to be a false good idea for invoice.
For any object, changing status can be done at no risk with no need of rules to allow this or not.
For invoices, there is some legal rules to validate invoice and has an important effect. Also fir common object this is a reversible action. Fir invouce validation is defrnitive and not reversible (due to law requirements).
Si validation should be done with dedicated api for invoice.

PR #30774, opened 2024-08-27
Qual: 🛟 Fix CI Issues ($db to $this->db / LDAP typing / Win-ci cache md5sum)

mdeweerd:
# Qual: 🛟 Fix CI Issues ($db to $this->db / LDAP typing / Win-ci cache md5sum)

- $db was undeclared, use $this->db.
- Also some typing with LDAP (initialize/reset error number, ignore notice wrt 'resource')
- Update the type of email_date for tickets;
- Correct md5sum calculation for reusing sql cache (work around issue with `find` command and environment limit on windows).


mdeweerd:
Updated base as the ci error was introduced in 20.0.

mdeweerd:
I suspect that the categorytest is failing because the sql database is V21 because the md5sum may be calculated on too few files.
```
2024-08-27T21:13:00.6465132Z ++ find D:/a/dolibarr/dolibarr/htdocs/install -type f -exec md5sum '{}' +
2024-08-27T21:13:00.6521836Z ++ LC_ALL=C
2024-08-27T21:13:00.6522257Z ++ sort
2024-08-27T21:13:00.6583011Z ++ md5sum
2024-08-27T21:13:00.7929648Z find: The environment is too large for exec().
```

I modify setup_conf.sh to compute the sum slightly differently and with extra output to show the number of files used.

EDIT:
The root cause is confirmed, the updated setup_conf.sh works correctly on the windows environment.

mdeweerd:
@eldy Please integrate this PR in 20.0 to fix the CI issues!!!

PR #30778, opened 2024-08-27
FIX - Substitution error in ticket emails in the subject

atm-lucasmantegari:
# FIX - *Substitution error in ticket emails in the subject*
*When applying an email template to a ticket email. Keys are not substituted in the subject.*

### DIFF
--- a/htdocs/core/class/html.formticket.class.php
+++ b/htdocs/core/class/html.formticket.class.php
@@ -1492,7 +1492,7 @@ class FormTicket
 			// Subject/topic
 			$topic = "";
 			foreach ($formmail->lines_model as $line) {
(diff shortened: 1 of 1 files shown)


htdocs/core/class/html.formticket.class.php
```diff
@@ -1493,7 +1493,7 @@ public function showMessageForm($width = '40%')
 			$topic = "";
 			foreach ($formmail->lines_model as $line) {
 				if ($this->param['models_id'] == $line->id) {
-					$topic = $line->topic;
+					$topic = make_substitutions($line->topic, $this->substit);
```
eldy, line 1496: this->substit seems not initialized in function. What happen if var is null or unset ?

Issue #28203, opened 2024-02-16
New signed status for contract


lmag:
### Feature Request

Feature Request
Currently Contract can have Draft, Validated or Done status. They can be signed by the client (paper or online). In the same way quotes have a signed (or unsigned) status, Contract could too.

Use case
For example: an Contract  with a signed status gives extra information to the person in charge of billing or organizing team. This is an extra guarantee that the Contract is complete and can be properly finalized and billed.

Suggested implementation
Add an extra status with associated logic.

Suggested steps

- [x]  Add data structure https://github.com/Dolibarr/dolibarr/pull/28204
- [ ]  Add new signed status
- [ ]  Add Set signed button (like Accepted button for quotes) & associated logic
- [ ]  Add notification for signed status
- [ ]  Add logic for status update with online signing
- [ ]  Update Contract module Wiki page
Thank to https://github.com/W1W1-M :)
See : https://github.com/Dolibarr/dolibarr/issues/27064

lmag:
Add new signed status @evarisk-micka 

- [ ] Class Ajouter dans les fields de la class contract
- [ ] Class Ajouter CRUD
- [ ] Ajouter dans la bannière getlibstatus
- [ ] Card ajouter les implémentations nécessaires (Add signed :0 etc...)
- [ ] List tag sign
- [ ] Pdf Sign ?
- [ ] API : CRUD

PR #30779, opened 2024-08-27
NEW contract signed status update

W1W1-M:
# NEW contract signed status update
This PR adds signed status : button, badge label, list search, notification, agenda event, & associated EN localization like #30629  This code is continued work for #28203 #28874 #28871 #23080

new-contract-list-signed-status
new-contract-signed-status

On the card, the signed status is integrated into the standard label badge through the `getLibStatut` method. It didn't seem reasonable to add a second label badge.
On the list, the signed status column uses a new method `getLibSignedStatus` in order to give user a dedicated label badge.

The action trigger used is `CONTRACT_MODIFY`, this trigger doesn't make a difference between the different signed statuses, it just shows it through the object context.

The `contrat` class use constants for the different signed statues. In order to centralise the code for label localization and also to confine the different signed statuses they have been moved to `SIGNED_STATUSES` array dictionary, like in #30629. 

In order to optimise code base, `SIGNED_STATUSES` + associated class methods could be moved to `CommonObject` but this then seems less optimal because not all objects will need this. A intermediate object (or PHP trait?) for commercial documents and similar objects might be a good compromise ?


W1W1-M:
@lmag any input for this PR ? Thanks

lmag:
I will check-it on this friday with @FHenry It seems excellent that a very good news for Dolibarr users !

PR #30784, opened 2024-08-28
Qual isolate FichinterLigne class to separate file

W1W1-M:
# Qual isolate FichinterLigne class to separate file
In order to make code more readable by reducing file length, "line" classes could be separated to individual files. This should improve developer experience by making navigation better and mitigating overscroll into line class when working on initial class. This is also improves on PSR-1 compliance: https://www.php-fig.org/psr/psr-1/

I'm not sure how much of a negative impact this would have on loading times, if at all ?

In this PR, the `Fichinter` class file length is reduced by 19% about 400 lines, from 2100 to 1700 lines.

If this is validated, other classes could easily follow (propal, commande, contrat, etc.)


htdocs/fichinter/class/fichinterligne.class.php
```diff
@@ -0,0 +1,392 @@
...
+	public $date; 		// Date intervention
+	/**
+	 * @var int Date of intervention
+	 * @deprecated
+	 */
+	public $datei; 		// Date intervention
+
+	public $duration; 	// Duration of intervention
+	public $rang = 0;
+	public $tva_tx;
```
mdeweerd, line 69: These phan notifications appear because there are no exceptions for the new file.
To fix the PhanUnknownPropertyType, a phpdoc should be added.

Here (from CommonInvoice) that could be:
```
    /**
     * VAT %  Vat rate can be like "21.30 (CODE)"
     * @var string|float
     */
    public $tva_tx;
```

These also appear with PHPSTAN as "Property FichinterLigne::$tva_tx has no type specified.".

You can filter these notices on https://cti.dolibarr.org/ by looking for  `FichinterLigne`  or `fichinter.class`

W1W1-M: Thanks for help

PR #30787, opened 2024-08-28
Qual ticket class

W1W1-M:
# Qual ticket class

- Updated PHPDoc
- Cleaned code(unnecessary logic, variables)


mdeweerd:
I can only add messages for lines that changed, so I'll comment here to help fix phan notices:

One technique is to cast the value:
```php
strtolower((string) $info_sendto['libelle'])
```
Phan does not keep track that $info_sendto['email'] is empty, so my technique is to assign that to a variable and test the variable.

Here we already tested the value is not '' in the previous if.

```php
$email=$info_sendto['email']

if ($email != null) { // or !== null

   $array[$email] = ...
```

W1W1-M:
@mdeweerd thanks for hep 😃

Issue #26596, opened 2023-11-14
Signature is not apply correctly on document


homer8173:
Signature doesn't care about PROPOSAL_FREE_TEXT

https://github.com/Dolibarr/dolibarr/blob/213ebc5e26a353efcd4eed25512f7eadba1c858d/htdocs/public/onlinesign/newonlinesign.php#L392 

you can have PDF with a signature on the legals

PascalB67:
Hi, ((in Dolibarr 18.03)
I take advantage of the post, 
The signature is at the bottom of the contract page (On 2 pages it's OK).
Only:
-the customer signature is at the bottom left and not on the right.
-the entered name and time information are not included.
Best regards

RANDOMLY-OBFUSCATED:
Hello!

Seen on Dolibarr `18.0.3` and `18.0.4`

I confirm what @PascalB67  wrote, but I'll add to it:
* I never seen in change logs an item about the ability to sign on line for the two parts. I discover it when I saw the customer signature on my box
* I asked again to the customer to sign after I suppressed the pdf and sign it by myself before him. He received a pdf with mine but my sign was replaced (I have now a pdf for each signature).
* After the second signing it's not locked. The customer can sign multiple times and Dolibarr add a pdf each time
* There's no notification item on signing event. Like for proposal, it will be really useful.

Best regards,

PascalB67:
Thank you for your clarification, I modified the location of the signature in the code, on occasion I will look to override this.

Yes, on the other hand, as a development, a message warning of the signature would be a good idea, even essential.

Best regards

RANDOMLY-OBFUSCATED:
Thank you too.
Could you write here when the PR will be integrated please?
(It will be useful for me to know in which version it will be corrected.)

Best regards,

RANDOMLY-OBFUSCATED:
> Yes, on the other hand, as a development, a message warning of the signature would be a good idea, even essential.

We have already an issue arround this: [no email sent when proposal is signed online #20204](https://github.com/Dolibarr/dolibarr/issues/20204)

PascalB67:
Thank you but nothing in 18.03 maybe en 18.04 ... ?

RANDOMLY-OBFUSCATED:
Welcome. No change with the .4

simogeo:
Hi ,I'm using dolibarr 19 with same issue here .... any news ? thanks

RANDOMLY-OBFUSCATED:
Hello,
[...]

PR #30788, opened 2024-08-28
FIX contract online signature image position

W1W1-M:
# FIX contract online signature image position
In current implementation, the online signature link generates image of signature in the company box (left) instead of third party box (right). This PR moves the position of signature image to box for third party on the right, like on interventions, proposals, etc.

fix-contract-online-signature-image-position

Associated issue : #26596


htdocs/core/ajax/onlineSign.php
```diff
@@ -410,9 +410,9 @@
 								// A signature image file is 720 x 180 (ratio 1/4) but we use only the size into PDF
 								// TODO Get position of box from PDF template
 
-								$param['xforimgstart'] = 10;
-								$param['yforimgstart'] = (empty($s['h']) ? 240 : $s['h'] - 65);
-								$param['wforimg'] = $s['w'] / 2 - $param['xforimgstart'];
+								$param['xforimgstart'] = 110;
```
eldy, line 413: The width of pdf may vary according to page format. So to have the start on right side we should use size of pdf /2 + offset.
It seems this is what old code did (good). But new one is using hard cided value 110 (wrong)

W1W1-M: Thanks for help

PR #30813, opened 2024-08-30
Add token oauth for social network

lamrani002:
# NEW|New Functionality to connect with token oauth when create fediverse 
fo
wi


htdocs/admin/fediverse.php
```diff
@@ -252,6 +273,53 @@
...
+	print ' &nbsp; <a href="'.DOL_URL_ROOT.'/admin/oauth.php">'.$langs->trans("SetupModuleX", " OAuth").'</a>';
+}
+print '</td>';
+print '</tr>';
+
+print '<tr class="oddeven" id="oauth_service_div"  style="display: none;">';
+print '<td>'.$langs->trans("OAuthService").'</td>';
+print '<td>';
+if (!isModEnabled('multicompany') || ($user->admin && !$user->entity)) {
+	print $form->selectarray('OAUTH_SERVICE_SOCIAL_NETWORK', $oauthservices, $conf->global->OAUTH_SERVICE_SOCIAL_NETWORK);
```
mdeweerd, line 310: I suppose that getDolGlobalString() will help for phan, and also typing $oauthservices for phan.
```
'@phan-var-force array{field1:type,field2:type,...} $oauthservices';
```
 just after the loop where it is dynamically defined.

lamrani002: ok i see, thanks

hregis: @lamrani002 why you check if multicompany is not enabled ? Wht is the problem ?

`if (!isModEnabled('multicompany') || ($user->admin && !$user->entity)) {`

htdocs/admin/fediverse.php
```diff
@@ -252,6 +273,56 @@
...
+        if ($("#radio_oauth").is(":checked")) {
+            $("#oauth_service_div").show();
+        } else {
+            $("#oauth_service_div").hide();
+        }
+    }
+
+    toggleOAuthServiceDisplay();
+
+    $("input[name=\"'.$vartosmtpstype.'\"]").change(function() {
```
eldy, line 288: var not defined ?

htdocs/admin/fediverse.php
mdeweerd: @lamrani002 As far as I can tell, after adding the typing hint, the phan notice is now there because one of the items in the array is an int and 'selectarray' is said to expect string[] (when it's an array).  So either convert all items to string, or update the selectarray() param type to 'int|string|array<int|string>'

PR #30827, opened 2024-09-02
NEW: Add a new hook into CommonObject#commonReplaceThirdparty

creekorful:
Sometime it is useful to complete or override the existing commonReplaceThirdparty behaviour. This can be the case if we need to modify/delete database entries during the replacement and we need to do so during the transaction.

The existing replaceThirdparty hook does not work for such use cases because the previous database transaction (during which replaceThirdParty/commonReplaceThirdParty are executed) may fail. If it's the case then the whole database transaction is going for a rollback rendering any actions done in the existing replaceThirdparty hooks useless.


creekorful:
@eldy rebased!

PR #30835, opened 2024-09-03
FIX: selection of parentholder for button dropdown

kubajznik:
# FIX
Multiple actions buttons with dropdown can be used on one page and won't open simultaneously.


eldy:
Can you provide more information on the bug that this PR fix ? May be a screenshot may help ?

kubajznik:
@eldy I attached a video. If you use two drop-down action buttons on one page, then both are opened at the same time.
https://github.com/user-attachments/assets/c5e73048-6753-47a6-87be-d79663b31c20

Here a screenshot of how it looks when I click on either one of the buttons:
image

PR #30840, opened 2024-09-03
Qual isolate OrderLine class to separate file

W1W1-M:
# Qual isolate OrderLine class to separate file
Like #30784 this moves the line class to a separate file.

In this PR, the Commande class file length is reduced by 14% about 700 lines, from 4900 to 4200 lines.

Also includes PHPDoc updates.


fappels:
Indeed this will be more clean, but will give trouble for merging fixes done in a lower version.
If you want to avoid scrolling, most IDE have some breadcrumb feature to go immediately to the line class.

PR #30841, opened 2024-09-03
FIX: Error message overwrited when a error occurs during update of product multilangs

kkhelifa-opendsi:
FIX: Error message overwrited when a error occurs during update of product multilangs


lvessiller-opendsi (approved):

rycks (approved):
ok for me, this error is not due to local sql error but setMultiLangs call and errors are localy handled into setMultiLangs

rycks:
@eldy ok for us, do we have to change tag or add a new one ?

eldy:
> @eldy ok for us, do we have to change tag or add a new one ?

No, just set the approved status so i know i can check on my side.

rycks (approved):

PR #30843, opened 2024-09-03
FIX : Wrong default PDF model when creating the second situation invoice

atm-adrien:


htdocs/compta/facture/card.php
```diff
@@ -3929,8 +3929,9 @@ function setRadioForTypeOfInvoice() {
 		include_once DOL_DOCUMENT_ROOT.'/core/modules/facture/modules_facture.php';
 		$liste = ModelePDFFactures::liste_modeles($db);
 		if (getDolGlobalString('INVOICE_USE_DEFAULT_DOCUMENT')) {
+			$type = GETPOSTINT('type') ? GETPOSTINT('type') : $object->type;
```
eldy, line 3932: May be 
$type = GETPOSTISSET('type') ? GETPOSTINT('type') : $object->type;
is better. Should work also if $_POST['type'] is '0'

atm-adrien: Done

Issue #30810, opened 2024-08-30
Allow to choose IBAN on Direct debit order and Credit transfert


atm-thibaultf:
### Feature Request

The new feature consists to allow to select which IBAN we would like to use on "Direct debit orders" for customer invoices and "Credit transfert" for vendor invoice

### Use case

For now on, when you create an invoice  with the payment method "Debit payment order", in the "Direct debit orders", it will automatically propose to make the requets on the defautl IBAN of the customer. If the customer has several IBAN, you can't choose the one you want. The feature would be to allow on this tab to choose an IBAN from the list of the customer IBANs, this field should be filled by default with the default IBAN of the client.

This behaviour is the same on supplier invoice for credit transfert

### Suggested implementation

We would like to remove the actual field "IBAN of customer" which is only for display information. We would like to add a new input "IBAN of customer" filled by default by the defautl IBAN of the customer. This field would be a select list with all the IBAN of the invoice customer and would allow to change this IBAN. When you are going to click on the request button, the selected IBAN will be used to be associated to the debit order (or the  credit transfert).

We are going to need a new column in the prelevement_demande table. This column called fk_societe_rib should stock the foreign key ok the societe RIB used for the request. We can also remove the column below from this table :
- code_banque
- code_guichet
- number
- cle_rib

PR #30850, opened 2024-09-04
NEW: added column default_rib in llx_user_rib

atm-thomasb:
# NEW|New column default_rib in llx_user_rib as requested in a TODO

In htdocs/compta/prelevement/class/bonprelevement.class.php, a TODO requested the following (see image):
image

In order to do that, we need to add a "default_rib" column to llx_user.

THIS COLUMN IS DEFAULTED AS 1 TO ALLOW RETROCOMPATIBILITY. The column isn't used yet so  it was set to default at 1 to make sure the newly added condition doesn't mess with existing processes.

This PR is made to prepare a future PR for this feature request about RIB : https://github.com/Dolibarr/dolibarr/issues/30810

### DIFF
--- a/htdocs/install/mysql/migration/20.0.0-21.0.0.sql
+++ b/htdocs/install/mysql/migration/20.0.0-21.0.0.sql
@@ -87,6 +87,7 @@ ALTER TABLE llx_c_hrm_public_holiday ADD UNIQUE INDEX uk_c_hrm_public_holiday(en
 ALTER TABLE llx_c_hrm_public_holiday ADD UNIQUE INDEX uk_c_hrm_public_holiday2(entity, fk_country, dayrule, day, month, year);
 
 ALTER TABLE llx_societe_account ADD COLUMN date_last_reset_password datetime after date_previous_login;
+ALTER TABLE llx_user_rib ADD COLUMN default_rib smallint NOT NULL DEFAULT 1;
 
 -- Rename of bank table
 ALTER TABLE llx_bank_categ RENAME TO llx_category_bank;		-- TODO Move content into llx_categorie instead of renaming it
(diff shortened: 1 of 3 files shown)


htdocs/install/mysql/migration/20.0.0-21.0.0.sql
```diff
@@ -87,6 +87,7 @@ ALTER TABLE llx_c_hrm_public_holiday ADD UNIQUE INDEX uk_c_hrm_public_holiday(en
 ALTER TABLE llx_c_hrm_public_holiday ADD UNIQUE INDEX uk_c_hrm_public_holiday2(entity, fk_country, dayrule, day, month, year);
 
 ALTER TABLE llx_societe_account ADD COLUMN date_last_reset_password datetime after date_previous_login;
+ALTER TABLE llx_user_rib ADD COLUMN default_rib smallint NOT NULL DEFAULT 1;
```
eldy, line 90: If you set DEFAULT 1 in sql, it means when we will be able to insert new lines, they will have default set automatically.
May be you can set "DEFAULT NULL".
Then run UPDATE set default_rib = 1 where default_rib IS NULL 
so all existing lines, at migration time, will have default_rib to 1, and then later, when we will enter a new bank, it will not be set as the default.

atm-thomasb: Thank you for your time!

After discussing with the team and our client we realized that issue https://github.com/Dolibarr/dolibarr/issues/30810 isn't about **user** ribs, so doing this is bonus.
With that in mind and since I'm running out of time this sprint, I won't be able to properly develop a work logic around **user** ribs soon. (ie: setting which ribs are to be used by default for each user, allowing to change it, etc.)

We then decided to do the following:
- I reverted the code NOT to use "default =1" in the sql request
- I STILL create the column and set the default to 0 (see below for explanation).
- I updated the TODO comment to say that the column has been created if someone wants to do the work logic.
I may end up doing it in the future, but I don't want to promise it so at least it's one step done


**Why not setting default to NULL ?**
--> because it's supposed to be working the same way as default_rib from societe_rib, which is NOT NULL, and by default 0

**Why not using UPDATE to set 1 to all existing so we can still add the sql "AND sr.default_rib = 1"?**
--> It would work, but only at first. Since we have no work logic for users yet, creating a new user would set its value to 0, while having every old user to 1, with no way to modify it. In the sql request it wouldn't be able to select the new user.

PR #30853, opened 2024-09-05
FIX shipment closing date not saved

zahwyrix:
FIX shipment closing date not saved


htdocs/expedition/class/expedition.class.php
```diff
@@ -2252,7 +2252,7 @@ public function setClosed()
 
 		$this->db->begin();
 
-		$sql = "UPDATE ".MAIN_DB_PREFIX."expedition SET fk_statut = ".self::STATUS_CLOSED;
+		$sql = "UPDATE ".MAIN_DB_PREFIX."expedition SET fk_statut = ".self::STATUS_CLOSED.", date_expedition = NOW()";
```
eldy, line 2255: NOW is forbidden. Does not take into account timezone setup of dolibarr.
You can use
date_expedition = '".$this->db->idate(dol_now())."'";

PR #30855, opened 2024-09-05
NEW - Add an advanced permission to validate knowledge

BB2A-Anthony:
# NEW|New [*Add perms validate knowladge*]
[*Long description*]


htdocs/core/modules/modKnowledgeManagement.class.php
```diff
@@ -251,6 +251,11 @@ public function __construct($db)
 		$this->rights[$r][5] = 'write'; // In php code, permission will be checked by test if ($user->rights->knowledgemanagement->level1->level2)
 		$r++;
 		$this->rights[$r][0] = $this->numero + $r + 1; // Permission id (must not be already used)
+		$this->rights[$r][1] = 'Validate articles'; // Permission label
```
eldy, line 254: We try to standardize the permission by default to just
read, write, delete
when this is possible.
And when we need another level, the permission must be introduce as an advanced permission.

Can you replace 
$this->rights[$r][4] = 'knowledgerecord';
into
$this->rights[$r][4] = 'knowledgerecord_advance';

then you must test permission to validate in 2 case like we do for orders:
```
if (!((!getDolGlobalString('MAIN_USE_ADVANCED_PERMS') && $user->hasRight('commande', 'creer'))
			|| (getDolGlobalString('MAIN_USE_ADVANCED_PERMS') && $user->hasRight('commande', 'order_advance', 'validate')))) 
```

BB2A-Anthony: @eldy
it's done

PR #30859, opened 2024-09-05
Qual: Improve typing in Export class

mdeweerd:
# Qual: Improve typing in Export class

Improve typing hints in Export class


mdeweerd:
@eldy 
What do you want to do about:
```plaintext
Error: Default value of the parameter #4 $downloadMode (0) of method AccountancyExport::export() is incompatible with type 1.
Error: Default value of the parameter #6 $noouput (0) of method AccountancyExport::export() is incompatible with type 1.
```

The default values for the arguments are the deprecated values.
I have set the typing for the arguments to the "recommended" values to detect any occurence where the deprecated values are used.

eldy:
> @eldy What do you want to do about:
> 
> ```
> Error: Default value of the parameter #4 $downloadMode (0) of method AccountancyExport::export() is incompatible with type 1.
> Error: Default value of the parameter #6 $noouput (0) of method AccountancyExport::export() is incompatible with type 1.
> ```
> 
> The default values for the arguments are the deprecated values. I have set the typing for the arguments to the "recommended" values to detect any occurence where the deprecated values are used.

Seems good to me to set the doc type to the default value so we have alert for those who use the deprecated value.

mdeweerd:
@eldy 
This is what I already did:
> Seems good to me to set the doc type to the default value so we have alert for those who use the deprecated value.

However, do I update the default values in the function argument defaults/initializers or do you want to keep backward compatibility for those defaults (in which case the phpstan notices need to be ignored).

eldy:
> @eldy This is what I already did:
> 
> > Seems good to me to set the doc type to the default value so we have alert for those who use the deprecated value.
> 
> However, do I update the default values in the function argument defaults/initializers or do you want to keep backward compatibility for those defaults (in which case the phpstan notices need to be ignored).

You can change the default value.
This method is at a very very low probability to be used by external module.

PR #30872, opened 2024-09-06
NEW display mark rates in objectLineCreate

atm-jonathan:
# Instructions
Display the input for the markup rate in the line addition form if the "DISPLAY_MARK_RATES" configuration of the margin module is enabled.


htdocs/core/tpl/objectline_create.tpl.php
```diff
@@ -1295,7 +1295,9 @@ function setforpredef() {
 		<?php } ?>
 		/* jQuery("#tva_tx, #title_vat").hide(); */
 		/* jQuery("#title_fourn_ref").hide(); */
-		jQuery("#np_marginRate, #np_markRate, .np_marginRate, .np_markRate, #units, #title_units").hide();
+		<?php if (!getDolGlobalString('DISPLAY_MARK_RATES')) { ?>
```
eldy, line 1298: There is two vars
DISPLAY_MARGIN_RATES
and
DISPLAY_MARK_RATES

May be you can make a difference with the two ?

PR #30877, opened 2024-09-06
Qual isolate ExpeditionLigne class to separate file

W1W1-M:
# Qual isolate ExpeditionLigne class to separate file
Like #30784 this moves the line class to a separate file.

In this PR, the Expedition class file length is reduced by 18% about 600 lines, from 3200 to 2600 lines.


W1W1-M:
@mdeweerd any idea for fixing this Phan type warning ? or does it need some sort of exception ? Thanks

mdeweerd:
> any idea for fixing this Phan type warning ? or does it need some sort of exception ? Thanks

I think we can agree that this is a bug in the code.

`$this->errors[]` is string[] and we are appending a `string[]` item to it, which makes it a `array<string|string[]>` .

So these solutions exist:
- Append the list of error to the pre-existing list `$this->errors = array_merge($this->errors, $lot->errors);` (possibly ensuring first that $this->errors is indeed an array).
- Change the type for $this->errors to `array<string|string[]>` ;
- Substitute $this->errors: `$this->errors = $lot->errors`.

I suppose that the first option is the correct one.

W1W1-M:
> > any idea for fixing this Phan type warning ? or does it need some sort of exception ? Thanks
> 
> I think we can agree that this is a bug in the code.
> 
> `$this->errors[]` is string[] and we are appending a `string[]` item to it, which makes it a `array<string|string[]>` .
> 
> So these solutions exist:
> 
> * Append the list of error to the pre-existing list `$this->errors = array_merge($this->errors, $lot->errors);` (possibly ensuring first that $this->errors is indeed an array).
> * Change the type for $this->errors to `array<string|string[]>` ;
> * Substitute $this->errors: `$this->errors = $lot->errors`.
> 
> I suppose that the first option is the correct one.

Thanks for help. I wasn't seeing it was the `errors` array 😅 Would there be any way to improve the Phan output on Github ? It seems like there are special chars not being formatted correctly, like `Assigning ($lot-&gt;errors as a field)` or am I misreading this ?
image

mdeweerd:
Concerning the representation of the messages: I think an option needs to be added to https://github.com/mdeweerd/logtocheckstyle to decode the url encoded messages or to not encode them (or maybe automatic with possibility to force it).

mdeweerd:
@W1W1-M I examined the possibility to parse the url encoded message, but it is [cs2pr](https://github.com/staabm/annotate-pull-request-from-checkstyle) that is used to generate the github annotations directly from the checkstyle compatible xml generated by phan.
I thought it was using logToCheckStyle to do this from a textual output.
So either phan or cs2pr would need to evolve.

W1W1-M:
> @W1W1-M I examined the possibility to parse the url encoded message, but it is [cs2pr](https://github.com/staabm/annotate-pull-request-from-checkstyle) that is used to generate the github annotations directly from the checkstyle compatible xml generated by phan. I thought it was using logToCheckStyle to do this from a textual output. So either phan or cs2pr would need to evolve.

Thanks for looking into this ! It would be nicer for dev experience. Not sure how much work is required but it would be good to get this done at some point.

PR #30883, opened 2024-09-06
fix typo actions_massactions.inc.php

frederic34:


mdeweerd:
If we're fixing the translation key, I'ld suggest: "ErrorIInsufficientPermission**s**" or "ErrorNoPermission" 😉

Issue #30790, opened 2024-08-28
Intervention list has duplicates


AMG-dot:
### Bug

![image](https://github.com/user-attachments/assets/1ca0cc3d-9b3f-42c3-9701-2175a7324ebf)

Intervention list has duplicates

### Dolibarr Version

19.0.3

### Environment PHP

8.2.16

### Environment Database

MySQL or MariaDB 10.11.8-MariaDB-cll-lve

W1W1-M:
Hello,
If an intervention has multiple lines attached to it they will show in the list. This may not be a bug. Do your duplicate interventions have one line or multiple lines ?

AMG-dot:
Multiple Lines but how is that useful to show same intervention because it has many lines

I didnt get the purpose of could u give me some idea?

W1W1-M:
> Multiple Lines but how is that useful to show same intervention because it has many lines
> 
> 
> 
> I didnt get the purpose of could u give me some idea?

I think if extra columns are shown it could be useful to see start & end dates for certain use cases ...

Maybe a new option to show only one line per intervention would be a good compromise.

Issue #25190, opened 2023-06-24
in fichinter we see double lines which confuse


tronx2100:
to limit the output in the fichinter list it should be enough to have one line per rowid
to limit the output to one line per rowid we should consider to add this block

         if (empty($conf->global->MAIN_FICHINTER_DISABLE_MULTI_LINES)){
                 if ($lastid == $objectstatic->id) {
                         continue;
                         }
                $lastid=$objectstatic->id;
         }

at line 697 in  /htdocs/fichinter/list.php

so with a switch everybody can easy restore original function with multiple output lines

Issue #7951, opened 2017-12-13
Feature request: Group intervention list by rowid


dolibarr95:
Hi

I think it could be great if in fichinter/list.php display list of intervention grouped by references as in orders. Currently the list displays all the lines of all the interventions. No very easy to understand/read.

## Screencapture
![screencapture](https://user-images.githubusercontent.com/24292300/33931708-18fa92b8-dff1-11e7-8d64-ae5db790847f.png)

# Feature Request
Group intervention by id in fichinter/list.php

Tks for reading

dolibarr95:
Hi!

@eldy Can you please give your opinion ?
I would like to create a PR but no sure if users agree.
The goal of this PR is to list the intervention as it's done for the orders, proposals, bills...
Tks for reading.

dolibarr95:
FICHINTER_DISABLE_DETAILS

dolibarr95:
Hi
This constant disable detais in  fichinter/list.php  but also in fichinter/card.php
@eldy I think we should add a new constant or only use this constant for the list.php page.

Tks for reading

dolibarr95:
Hi @eldy !
Can you please give help or advice ?
What about something like this :
```php
//FICHINTER_GROUP_LIST constant to group by lines in the list.php page (only)
//FICHINTER_DISABLE_DETAILS constant to disable details in Fichinter module

if (empty($conf->global->FICHINTER_DISABLE_DETAILS) || empty($conf->global->FICHINTER_GROUP_LIST )) $sql.= " fd.description as descriptiondetail, fd.date as dp, fd.duree,";

etc.
```
in the file 
```
fichinter/list.php
```

dolibarr95:
done in v9
https://github.com/Dolibarr/dolibarr/commit/b83ff06db80028467f70954e93eead9013e3bb99

PR #30884, opened 2024-09-06
NEW fichinter list disable details option

W1W1-M:
# NEW fichinter list disable details option
Currently `FICHINTER_DISABLE_DETAILS` option will disable details for list view and card views of interventions. In some cases (like #30790 #25190 #7951) it is necessary to have intervention detail lines but not on the list view: only one intervention per line. This PR adds an explicit `FICHINTER_DISABLE_DETAILS_ON_LIST` option which replaces `FICHINTER_DISABLE_DETAILS` for the list view only. This means users can have the best of both depending on their needs and without having to fight with column settings.

If this PR is merged, intervention admin page could be updated to add these options to administration. Dolibarr wiki will also need updating.


eldy:
> FICHINTER_DISABLE_DETAILS_ON_LIST

Instead of adding new constant, i would prefer that we play on the value. For example:
FICHINTER_DISABLE_DETAILS = '0'    // not disabled (like not set)
FICHINTER_DISABLE_DETAILS = '1'    // lines disabled
FICHINTER_DISABLE_DETAILS = 'listonly'        // Lines disabled on list only

Avoid to be lost when sharing all options, and avoid the "optionflation".

W1W1-M:
> > FICHINTER_DISABLE_DETAILS_ON_LIST
> 
> Instead of adding new constant, i would prefer that we play on the value. For example: FICHINTER_DISABLE_DETAILS = '0' // not disabled (like not set) FICHINTER_DISABLE_DETAILS = '1' // lines disabled FICHINTER_DISABLE_DETAILS = 'listonly' // Lines disabled on list only
> 
> Avoid to be lost when sharing all options, and avoid the "optionflation".

Thanks for input. I've also updated Dolibarr wiki accordingly : https://wiki.dolibarr.org/index.php?title=Setup_Other#Interventions_Module

silvianionescu:
In "fichinter/list", in practice it is easier to follow interventions, the details of interventions should appear in "fichinter/card", every detail of the intervention displayed in the lists creates confusion, especially when there are many interventions

PR #30885, opened 2024-09-06
Fix : update line order with TTC price leads to bad unit price

lemathou:
# FIX
When updating an order line with a TTC price, the pu_ht is stored as TTC and the TTC has double TTC, and so, when creating and invoice, the prices are also different
So, I disabled the bad variable affectation.

### DIFF
--- a/htdocs/commande/class/commande.class.php
+++ b/htdocs/commande/class/commande.class.php
@@ -3182,7 +3182,8 @@ class Commande extends CommonOrder
 
 			// Anciens indicateurs: $price, $subprice (a ne plus utiliser)
 			$price = $pu_ht;
-			if ($price_base_type == 'TTC') {
+			// disabled, because $tabprice returns already the right prices, so subprice should always be equal to $pu_ht
+			if (false && $price_base_type == 'TTC') {
 				$subprice = $pu_ttc;
 			} else {
 				$subprice = $pu_ht;
--- a/htdocs/commande/class/commande.class.php
+++ b/htdocs/commande/class/commande.class.php
@@ -3182,8 +3182,7 @@ class Commande extends CommonOrder
 
 			// Anciens indicateurs: $price, $subprice (a ne plus utiliser)
 			$price = $pu_ht;
-			// disabled, because $tabprice returns already the right prices, so subprice should always be equal to $pu_ht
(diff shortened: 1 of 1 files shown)


hregis:
@lemathou i don't have this problem ! you have external modules ?

lemathou:
I have modules, I will try with a fresh install, but I see that the order::lineupdate is the only case (of line update) where the pu_ttc is put into the suprice, instead of the the pu_ht, as it is in the invoices and propales.

lemathou:
When I update an order product line and I put 10€ in the TTC unit price, with qty=1, it stores 10€ in the HT unit price, 12€ in the TTC unit price, and 10€ in the total... this is not consistent

lemathou:
> @lemathou i don't have this problem ! you have external modules ?

I confirm I have the problem without external modules. With v.19.0.2.
@hregis Can you check again and confirm ?
I should suggest to merge it to v19. Maybe an older version...

hregis:
@lemathou yes the problem exists in v19 ! but not in develop branch... 
`$subprice = $pu_ttc;` is ok if price_base_type is TTC... the problem is not your fix !

lemathou:
> @lemathou yes the problem exists in v19 ! but not in develop branch... `$subprice = $pu_ttc;` is ok if price_base_type is TTC... the problem is not your fix !

@hregis OK I switched to v19. But, I don't understand ... 

> `$subprice = $pu_ttc;` is ok if price_base_type is TTC... the problem is not your fix !

@hregis I am not sure to understand your answear correctly, but I see that :

in the v.19 we have :

```
$subprice = $pu_ttc;
...
$this->line->subprice       = $subprice;
```

and this is not working even if price_base_type is TTC

otherwise, in the develop, we have :
```
$this->line->subprice       = $pu_ht;
```
and this is working

So, doing `$subprice = $pu_ht;` in the v19 fixes the bug, non ?

hregis:
@lemathou ok i understand, your rights... you can see this PR, same fix from develop branch : https://github.com/Dolibarr/dolibarr/pull/30887

lemathou:
@hregis OK Thanks, I simplified the PR as it is in the develop, as you did !
I am quite a newbie in PRs...

PR #30892, opened 2024-09-07
Update api_projects.class.php with pagination

ptibogxiv:


hregis:
Hello @ptibogxiv is this to avoid recalculating for the application that queries the API? does it not slow down processing if there are a lot of lines in the queried database?

ptibogxiv:
> Hello @ptibogxiv is this to avoid recalculating for the application that queries the API? does it not slow down processing if there are a lot of lines in the queried database?

Indeed, its better that calculate pages and datas with the external application
it's more efficient with all my tests and deployment, i introduce this optional pagination in all objects

PR #30913, opened 2024-09-10
Qual common signed object trait

W1W1-M:
# Qual common signed object trait
In order to make new signed status code reusable for other objects, this PR moves concerned class properties & methods to a new class trait: `CommonSignedObject`

Any input on trait name, properties, methods, architecture, etc ... is welcome


htdocs/core/class/commonsignedobject.php
```diff
@@ -0,0 +1,203 @@
...
+		$sql .= " WHERE rowid = ".((int) $this->id);
+
+		if ($this->db->query($sql)) {
+			if (!$error) {
+				$this->oldcopy = clone $this;
+			}
+
+			if (!$error && !$notrigger) {
+				// Call trigger
+				$result = $this->call_trigger($triggercode, $user);
```
mdeweerd, line 166: This supposes that the Trait is used with a CommonObject and the Trait can not itself be restricted to certain classes language-wise.

There are two solutions:
- Test that $this is a CommonObject (`if ($this instanceof CommonObject`) and preferable add some error handling;
- Add phpdoc suggesting that the method exists (which can also serve as documentation to remind that the method is expected on class the trait is used with).
```
Methods that should exist in the host class:
@method call_trigger() <returntype> call_trigger(<param doc>)

W1W1-M: > This supposes that the Trait is used with a CommonObject and the Trait can not itself be restricted to certain classes language-wise.
> 
> There are two solutions:
> 
> * Test that $this is a CommonObject (`if ($this instanceof CommonObject`) and preferable add some error handling;
> * Add phpdoc suggesting that the method exists (which can also serve as documentation to remind that the method is expected on class the trait is used with).
> 
> ```
> /**
>  * Methods that should exist in the host class:
>  * @method call_trigger() <returntype> call_trigger(<param doc>) 
>  *
>  * To detect all properties that are supposed to exist in the host class:
>  * @phan-forbid-undeclared-magic-properties 
>  */
> ```
> 
> Regarding the properties, you can suggest them with `@property` in the class documentation.

Thanks for advice !

PR #30915, opened 2024-09-10
Backport #yogosha18281

MaximilienR-easya:
# Backport yogosha [*4b214b4*]
backport that fix a case when preview of ticket attached file on the card contain ' and break the preview link

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -10358,7 +10358,7 @@ function getAdvancedPreviewUrl($modulepart, $relativepath, $alldata = 0, $param 
 	// old behavior, return a string
 	if ($isAllowedForPreview) {
 		$tmpurl = DOL_URL_ROOT.'/document.php?modulepart='.urlencode($modulepart).'&attachment=0&file='.urlencode($relativepath).($param ? '&'.$param : '');
-		$title = $langs->trans("Preview");
+		$title = $langs->transnoentities("Preview");
 		//$title = '%27-alert(document.domain)-%27';
 		//$tmpurl = 'file='.urlencode("'-alert(document.domain)-'_small.jpg");
 
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -10364,7 +10364,7 @@ function getAdvancedPreviewUrl($modulepart, $relativepath, $alldata = 0, $param 
 
 		// We need to urlencode the parameter after the dol_escape_js($tmpurl) because  $tmpurl may contain n url with param file=abc%27def if file has a ' inside.
 		// and when we click on href with this javascript string, a urlcode is done by browser, converted the %27 of file param
-		return 'javascript:document_preview(\''.urlencode(dol_escape_js($tmpurl)).'\', \''.urlencode(dol_mimetype($relativepath)).'\', \''.urlencode(dol_escape_js($title)).'\')';
+		return 'javascript:document_preview(\''.urlencode(dol_escape_js($tmpurl)).'\', \''.urlencode(dol_mimetype($relativepath)).'\', \''.rawurlencode(dol_escape_js($title)).'\')';
 	} else {
 		return '';
 	}
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -10349,7 +10349,7 @@ function getAdvancedPreviewUrl($modulepart, $relativepath, $alldata = 0, $param 
 
 	if ($alldata == 1) {
 		if ($isAllowedForPreview) {
-			return array('target'=>'_blank', 'css'=>'documentpreview', 'url'=>DOL_URL_ROOT.'/document.php?modulepart='.$modulepart.'&attachment=0&file='.urlencode($relativepath).($param ? '&'.$param : ''), 'mime'=>dol_mimetype($relativepath));
+			return array('target'=>'_blank', 'css'=>'documentpreview', 'url'=>DOL_URL_ROOT.'/document.php?modulepart='.urlencode($modulepart).'&attachment=0&file='.urlencode($relativepath).($param ? '&'.$param : ''), 'mime'=>dol_mimetype($relativepath));
 		} else {
 			return array();
 		}
@@ -10357,7 +10357,14 @@ function getAdvancedPreviewUrl($modulepart, $relativepath, $alldata = 0, $param 
 
 	// old behavior, return a string
 	if ($isAllowedForPreview) {
-		return 'javascript:document_preview(\''.dol_escape_js(DOL_URL_ROOT.'/document.php?modulepart='.$modulepart.'&attachment=0&file='.urlencode($relativepath).($param ? '&'.$param : '')).'\', \''.dol_mimetype($relativepath).'\', \''.dol_escape_js($langs->trans('Preview')).'\')';
+		$tmpurl = DOL_URL_ROOT.'/document.php?modulepart='.urlencode($modulepart).'&attachment=0&file='.urlencode($relativepath).($param ? '&'.$param : '');
+		$title = $langs->trans("Preview");
+		//$title = '%27-alert(document.domain)-%27';
+		//$tmpurl = 'file='.urlencode("'-alert(document.domain)-'_small.jpg");
+
+		// We need to urlencode the parameter after the dol_escape_js($tmpurl) because  $tmpurl may contain n url with param file=abc%27def if file has a ' inside.
+		// and when we click on href with this javascript string, a urlcode is done by browser, converted the %27 of file param
+		return 'javascript:document_preview(\''.urlencode(dol_escape_js($tmpurl)).'\', \''.urlencode(dol_mimetype($relativepath)).'\', \''.urlencode(dol_escape_js($title)).'\')';
 	} else {
 		return '';
 	}
--- a/htdocs/core/class/translate.class.php
+++ b/htdocs/core/class/translate.class.php
@@ -580,7 +580,7 @@ class Translate
 	 */
 	private function getTradFromKey($key)
 	{
-		global $conf, $db;
+		global $db;
 
 		if (!is_string($key)) {
 			//xdebug_print_function_stack('ErrorBadValueForParamNotAString');
@@ -660,7 +660,7 @@ class Translate
 				}
 			}
 
-			// Crypt string into HTML
+			// Encode string into HTML
 			$str = htmlentities($str, ENT_COMPAT, $this->charset_output); // Do not convert simple quotes in translation (strings in html are embraced by "). Use dol_escape_htmltag around text in HTML content
 
 			// Restore reliable HTML tags into original translation string
@@ -669,6 +669,10 @@ class Translate
 				array('"', '<b>', '</b>', '<u>', '</u>', '<i', '</i>', '<center>', '</center>', '<a ', '</a>', '<br>', '<span', '</span>', '< ', '>'),
 				$str
 			);
+
+			// Remove dangerous sequence we should never have. Not needed into a translated response.
+			// %27 is entity code for ' and is replaced by browser automatically when translation is inside a javascript code called by a click like on a href link.
+			$str = str_replace(array('%27', '&#39'), '', $str);
 
 			if ($maxsize) {
 				$str = dol_trunc($str, $maxsize);
@@ -738,6 +742,10 @@ class Translate
 				//print $str;
 				$str = sprintf($str, $param1, $param2, $param3, $param4, $param5); // Replace %s and %d except for FormatXXX strings.
 			}
+
+			// Remove dangerous sequence we should never have. Not needed into a translated response.
+			// %27 is entity code for ' and is replaced by browser automatically when translation is inside a javascript code called by a click like on a href link.
+			$str = str_replace(array('%27', '&#39'), '', $str);
 
 			return $str;
 		} else {
--- a/htdocs/core/js/lib_head.js.php
+++ b/htdocs/core/js/lib_head.js.php
@@ -996,7 +996,7 @@ function document_preview(file, type, title)
 	var ValidImageTypes = ["image/gif", "image/jpeg", "image/png", "image/webp"];
 	var showOriginalSizeButton = false;
 
-	console.log("document_preview A click was done. file="+file+", type="+type+", title="+title);
+	console.log("document_preview A click was done: file="+file+", type="+type+", title="+title);
 
 	if ($.inArray(type, ValidImageTypes) < 0) {
 		/* Not an image */


rycks (approved):
ok for me (security backport), thanks for 18 LTS :-)

htdocs/core/lib/functions.lib.php
```diff
@@ -10349,15 +10349,22 @@ function getAdvancedPreviewUrl($modulepart, $relativepath, $alldata = 0, $param
...
 	if ($isAllowedForPreview) {
-		return 'javascript:document_preview(\''.dol_escape_js(DOL_URL_ROOT.'/document.php?modulepart='.$modulepart.'&attachment=0&file='.urlencode($relativepath).($param ? '&'.$param : '')).'\', \''.dol_mimetype($relativepath).'\', \''.dol_escape_js($langs->trans('Preview')).'\')';
+		$tmpurl = DOL_URL_ROOT.'/document.php?modulepart='.urlencode($modulepart).'&attachment=0&file='.urlencode($relativepath).($param ? '&'.$param : '');
+		$title = $langs->trans("Preview");
+		//$title = '%27-alert(document.domain)-%27';
+		//$tmpurl = 'file='.urlencode("'-alert(document.domain)-'_small.jpg");
+
+		// We need to urlencode the parameter after the dol_escape_js($tmpurl) because  $tmpurl may contain n url with param file=abc%27def if file has a ' inside.
+		// and when we click on href with this javascript string, a urlcode is done by browser, converted the %27 of file param
+		return 'javascript:document_preview(\''.urlencode(dol_escape_js($tmpurl)).'\', \''.urlencode(dol_mimetype($relativepath)).'\', \''.urlencode(dol_escape_js($title)).'\')';
```
rycks, line 10367: @lvessiller-opendsi show that's a strange thing with urlencode on mimetype and title

here is the result of my test in javascript console with manual data set with urlencode:

image

the result is a popup with no pdf displayed and a title with raw url encoded string :

image

without urlencode on mimetype and title everything seems good:

image
 

image

rycks (changes requested):

htdocs/core/lib/functions.lib.php
rycks: here is the code you can type into debug js console

```
javascript:document_preview('/document.php?modulepart=propal&attachment=0&file=PR2409-0059%2FPR2409-0059.pdf&&entity=1', 'application/pdf', 'test éric / fichier js'); 
```

and

```
javascript:document_preview('/document.php?modulepart=propal&attachment=0&file=PR2409-0059%2FPR2409-0059.pdf&&entity=1', 'application%2fpdf', 'test+%C3%A9ric+%2F+fichier+js'); 
```

@eldy original fix seems to be concerned by the same error ...

eldy: I don't understand how to reproduce your point @rycks. 
You just upload a file with name "test éric / fichier js.pdf" on a proposal ?

lvessiller-opendsi (approved):

htdocs/core/lib/functions.lib.php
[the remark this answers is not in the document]
lvessiller-opendsi: @rycks 
It works fine with urlencode.

MaximilienR-easya:
New commit to backport #29888

lvessiller-opendsi (approved):

MaximilienR-easya:
@eldy 
I have case when using urlencode on title with space character, exemple :
with_urlencode

We can solve this with using rawurlencode which replace the '+' of the urlencode by '%20' which is working with js parameter :
with_rawurlencode

or by droping the urlencode :
without_urlencode

Which of the solution do you think is the more appropriate approach ?
When I have your answer, I'm going to make a 2nd PR for develop to also change the urlencode on this parameter

eldy:
> I have case when using urlencode on title with space character, exemple

This is on developer side. But how, as a user, do you reproduce the bug ?

MaximilienR-easya:
> > I have case when using urlencode on title with space character, exemple
> 
> This is on developer side. But how, as a user, do you reproduce the bug ?

It was user side, I just changed the translation for preview

rycks:
ok, thanks to @MaximilienR-easya rawurlencode works better than urlencode for that sort of situations then a fix with rawurlencode is better than urlencode for js context

@eldy it could be a good idea to make a global review of all javascript calls where urlencode is used ?  >> after a speed review there is no other part of dolibarr where there is javascript call with urlencode

eldy:
> ok, thanks to @MaximilienR-easya rawurlencode works better than urlencode for that sort of situations then a fix with rawurlencode is better than urlencode for js context
> 
> @eldy it could be a good idea to make a global review of all javascript calls where urlencode is used ?  >> after a speed review there is no other part of dolibarr where there is javascript call with urlencode

Using rawurlencode should not be necessary for part of query string (urlencode should work)
rawurlencoee should be necessary for url path only, but in the use case of dolibarr we have url path into DOL_URL_ROOT, so i see no reason to have rawurlencode in dolibarr.
When we want to escape an url, we should use urlencode, when we need to escape javascript, we should use dol_escape_js.
This is why i would like to know how to reproduce the bug because a fix using rawurlencode is suspicious.

eldy:
> > > I have case when using urlencode on title with space character, exemple
> > 
> > 
> > This is on developer side. But how, as a user, do you reproduce the bug ?
> 
> It was user side, I just changed the translation for preview

Sorry, the process to reproduce the bug is still not clear. Can you describe what you do to generate the bug (in a user point of view, only end user action) ?

MaximilienR-easya:
> > > > I have case when using urlencode on title with space character, exemple
> > > 
> > > 
> > > This is on developer side. But how, as a user, do you reproduce the bug ?
> > 
> > 
> > It was user side, I just changed the translation for preview
> 
> Sorry, the process to reproduce the bug is still not clear. Can you describe what you do to generate the bug (in a user point of view, only end user action) ?


To reproduce it with a user point of view, only end user action : 

1. go inside the ticket to see the preview title default traduction
Screenshot from 2024-09-20 14-10-27

2. go inside the parameter in the traduction
Traduction

3. Replace the traduction for the key “Preview” with the string of your choice
Screenshot from 2024-09-20 14-08-51

4. Return to your ticket to see the preview
Screenshot from 2024-09-20 14-11-11

For the third point, you can retrieve the traduction key simply by research
Screenshot from 2024-09-20 14-24-02

rycks:
@eldy could you please explain to me why ?

> a fix using rawurlencode is suspicious

because https://www.php.net/rawurlencode and https://www.php.net/urlencode explains the diff:

> urlencode Returns a string in which all non-alphanumeric characters except -_. have been replaced with a percent (%) sign followed by two hex digits and spaces encoded as plus (+) signs. It is encoded the same way that the posted data from a WWW form is encoded, that is the same way as in application/x-www-form-urlencoded media type. This differs from the [» RFC 3986](http://www.faqs.org/rfcs/rfc3986) encoding (see [rawurlencode()](https://www.php.net/manual/en/function.rawurlencode.php)) in that for historical reasons, spaces are encoded as plus (+) signs.

then rawurlencode is better than urlencode in such situation where a js interpreter could be present (like in web browser on javascript://url) because rawurlencode implements RFC 3986 ... then i really can't understand why using rawurlencode is suspicious ?

eldy:
> @eldy could you please explain to me why ?
> 
> > a fix using rawurlencode is suspicious
> 
> because https://www.php.net/rawurlencode and https://www.php.net/urlencode explains the diff:
> 
> > urlencode Returns a string in which all non-alphanumeric characters except -_. have been replaced with a percent (%) sign followed by two hex digits and spaces encoded as plus (+) signs. It is encoded the same way that the posted data from a WWW form is encoded, that is the same way as in application/x-www-form-urlencoded media type. This differs from the [» RFC 3986](http://www.faqs.org/rfcs/rfc3986) encoding (see [rawurlencode()](https://www.php.net/manual/en/function.rawurlencode.php)) in that for historical reasons, spaces are encoded as plus (+) signs.
> 
> then rawurlencode is better than urlencode in such situation where a js interpreter could be present (like in web browser on javascript://url) because rawurlencode implements RFC 3986 ... then i really can't understand why using rawurlencode is suspicious ?

rawurlencode was designed to encode url part before the parameters, like the web domain name. In Dolibarr we should never have the part before the DOL_URL_ROOT appearing in html or js because we use relative path. So no need for rawurlencode. The other case is the rest of url berweend domain name and parameters, but for such part, the urlencode is recommanded. Also for filename part into an url that is under control of user, Dolibarr should have sanitized the filename before writing it on disk. So if we need rawurlencode, it probably mean we miss something else somewhere else.

But as long as there is no description on how to reproduce the bug i can't provide more information. So I insist another time, if there is a bug, providing a description on how to reproduce it is as important as providing a solution.

For the moment ticket is still not clear.
Is the bug appears before or after the suggested PR ?

So my final question remain the same: How can we reproduce it ? (It is necessary to know that to validate the change, and, if necessary, to implement the phpunit to avoid to have this bug back again).

hregis:
@MaximilienR-easya @rycks i don't have your problem ? 

Capture d’écran 2024-09-28 à 08 44 51

MaximilienR-easya:
> @MaximilienR-easya @rycks i don't have your problem ?
> 
> Capture d’écran 2024-09-28 à 08 44 51

@hregis I think you used the wrong link to open the preview:


how to reproduce

The problem is different with the preview next to the file, it's not the same code. for this one the code is in htdocs/core/js/lib_foot.js.php

```
if ($conf->browser->layout != 'phone') {
	print "\n/* JS CODE TO ENABLE document_preview */\n"; // Function document_preview is into header
	print '		jQuery(document).ready(function () {
			        jQuery(".documentpreview").click(function () {
            		    console.log("We click on preview for element with href="+$(this).attr(\'href\')+" mime="+$(this).attr(\'mime\'));
            		    document_preview($(this).attr(\'href\'), $(this).attr(\'mime\'), \''.dol_escape_js($langs->transnoentities("Preview")).'\');
                		return false;
        			});
		});'."\n";
}
```
Where the title is not encoded at all.

hregis:
@MaximilienR-easya I have the same result with this link, I don't have your error, with the current v18 branch.

MaximilienR-easya:
> @MaximilienR-easya I have the same result with this link, I don't have your error, with the current v18 branch.

The current v18 don't have the backport yogosha with urlencode at all. This is why you can't reproduce it with the current v18.

The discussion is about a modification added by the backport.
If we backport the same as in develop, we should use urlencode (which cause the '+' character to appear).
That is why I suggest modifying urlencode to rawurlencode inside the backport and once it is validated, apply the modification apported to the backport to the next version.

MaximilienR-easya:
> @MaximilienR-easya I have the same result with this link, I don't have your error, with the current v18 branch.

@hregis 
If you want to test it with the backport with urlencode and not rawurlencode, this is this commit 1ee3c5e5a078f090f314e5731dd7efd930082e43

hregis:
@MaximilienR-easya yes ok i see, thanks

MaximilienR-easya:
> For the moment ticket is still not clear.
> Is the bug appears before or after the suggested PR ?

@eldy You have an aswer to that question here


> > @MaximilienR-easya I have the same result with this link, I don't have your error, with the current v18 branch.
> 
> The current v18 don't have the backport yogosha with urlencode at all. This is why you can't reproduce it with the current v18.
> 
> The discussion is about a modification added by the backport. If we backport the same as in develop, we should use urlencode (which cause the '+' character to appear). That is why I suggest modifying urlencode to rawurlencode inside the backport and once it is validated, apply the modification apported to the backport to the next version.



> > @MaximilienR-easya I have the same result with this link, I don't have your error, with the current v18 branch.
> 
> @hregis If you want to test it with the backport with urlencode and not rawurlencode, this is this commit [1ee3c5e](https://github.com/Dolibarr/dolibarr/commit/1ee3c5e5a078f090f314e5731dd7efd930082e43)

eldy:
> > For the moment ticket is still not clear.
> > Is the bug appears before or after the suggested PR ?
> 
> @eldy You have an aswer to that question here
> 
> 
> > > @MaximilienR-easya I have the same result with this link, I don't have your error, with the current v18 branch.
> > 
> > The current v18 don't have the backport yogosha with urlencode at all. This is why you can't reproduce it with the current v18.
> > 
> > The discussion is about a modification added by the backport. If we backport the same as in develop, we should use urlencode (which cause the '+' character to appear). That is why I suggest modifying urlencode to rawurlencode inside the backport and once it is validated, apply the modification apported to the backport to the next version.
> 
> 
> 
> > > @MaximilienR-easya I have the same result with this link, I don't have your error, with the current v18 branch.
> > 
> > @hregis If you want to test it with the backport with urlencode and not rawurlencode, this is this commit [1ee3c5e](https://github.com/Dolibarr/dolibarr/commit/1ee3c5e5a078f090f314e5731dd7efd930082e43)
> 

Things are still suspicious to me.
Change to use rawurlcode is required if we backport a security fix. On this case because the security fix is already in v20, we should also need rawurlcode in v20. If  we have this in v20 can you point me when rawurlencode is used in v20 ? And if we have not, it should means there is a bug in v20 too, so we should be able to reproduce but i did not find how.

eldy:
> > For the moment ticket is still not clear.
> > Is the bug appears before or after the suggested PR ?
> 
> @eldy You have an aswer to that question here
> 
> 
> > > @MaximilienR-easya I have the same result with this link, I don't have your error, with the current v18 branch.
> > 
> > The current v18 don't have the backport yogosha with urlencode at all. This is why you can't reproduce it with the current v18.
> > 
> > The discussion is about a modification added by the backport. If we backport the same as in develop, we should use urlencode (which cause the '+' character to appear). That is why I suggest modifying urlencode to rawurlencode inside the backport and once it is validated, apply the modification apported to the backport to the next version.
> 
> 
> 
> > > @MaximilienR-easya I have the same result with this link, I don't have your error, with the current v18 branch.
> > 
> > @hregis If you want to test it with the backport with urlencode and not rawurlencode, this is this commit [1ee3c5e](https://github.com/Dolibarr/dolibarr/commit/1ee3c5e5a078f090f314e5731dd7efd930082e43)
> 

Things are still suspicious to me.
Change to use rawurlcode is required if we backport a security fix. On this case because the security fix is already in v20, we should also need rawurlcode in v20. If  we have this in v20 can you point me when rawurlencode is used in v20 ? And if we have not, it should means there is a bug in v20 too, so we should be able to reproduce but i did not find how.

lvessiller-opendsi:
@eldy 
The problem still exists on v20 and on develop.
This problem is only on preview when you click on picture on the top banner of the ticket card : 
image

**To reproduce**
1. Change the translate of "Preview" from "Traduction" menu : 
image

2. Insert a "PNG" picture on ticket card
image

3. Click on picture on top banner of ticket card
image

4. You got a translated title containing "+" instead of having "spaces" on the opened dialog for preview :
image

Here we see two problems : 
- "spaces" are replaced by "+"
- "ç" is replaced by HTML code : `&ccedil;`

If we apply the rawurlencode() for title of preview : 
`return 'javascript:document_preview(\''.urlencode(dol_escape_js($tmpurl)).'\', \''.urlencode(dol_mimetype($relativepath)).'\', \''.rawurlencode(dol_escape_js($title)).'\')';`

-> it fixes the first problem of "spaces" replaced by "+"
image

However, it still remains a problem : "ç" is replaced by `&ccedil;` but it's seem to be a wrong way to save translated words in database.
If you change directly in the code : `$langs->transnoenttites("Aperçu d'images")`, you don't have this problem.

So I suggest to begin with the approval of this FIX and then see why the "ç" is encoded in database.

eldy:
> @eldy 
> The problem still exists on v20 and on develop.
> This problem is only on preview when you click on picture on the top banner of the ticket card : 
> image
> 
> **To reproduce**
> 1. Change the translate of "Preview" from "Traduction" menu : 
> image
> 
> 2. Insert a "PNG" picture on ticket card
> image
> 
> 3. Click on picture on top banner of ticket card
> image
> 
> 4. You got a translated title containing "+" instead of having "spaces" on the opened dialog for preview :
> image
> 
> Here we see two problems : 
> - "spaces" are replaced by "+"
> - "ç" is replaced by HTML code : `&ccedil;`
> 
> If we apply the rawurlencode() for title of preview : 
> `return 'javascript:document_preview(\''.urlencode(dol_escape_js($tmpurl)).'\', \''.urlencode(dol_mimetype($relativepath)).'\', \''.rawurlencode(dol_escape_js($title)).'\')';`
> 
> -> it fixes the first problem of "spaces" replaced by "+"
> image
> 
> However, it still remains a problem : "ç" is replaced by `&ccedil;` but it's seem to be a wrong way to save translated words in database.
> If you change directly in the code : `$langs->transnoenttites("Aperçu d'images")`, you don't have this problem.
> 
> So I suggest to begin with the approval of this FIX and then see why the "ç" is encoded in database.
> 
> 
> 
> 

Thanks for giving the step to reproduce the bug in b20. This help me to understand what was wrong. So i made a more complete and commented fix in v20. V20 was corrextly secured but encoding was done on url param instead of the full uri.

This PR on v18 can be validated, even if correction not completely good. Still missing the last changement i did on v20...

PR #30918, opened 2024-09-10
FIX ajaxik URL in ExpenseReport to load coef calculation

### LINKED ISSUE #30917: In expense report coef calculation is not loaded
### Bug

When `MAIN_USE_EXPENSE_IK` set to 1, in the expense report after selecting a vehicule category, the coef calculation is not loaded. The URL for the Ajax call gives a 404 error.

### Dolibarr Version

19.0

### Environment PHP

_No response_

### Environment Database

_No response_

On a expense report, add an expense with type `Mileage costs` (`fk_c_type_fees` = `4`) and in `Vehicle category` `6 CV` for example.
After change the column `U.P. (net)` should be updated.

### Attached files

_No response_

zebastiane:
# FIX #30917 load coef calculation in expense report

This patch fixes a regression in Dolibarr v19.0 and revert the line like it was in Dolibarr v18.0.

### DIFF
--- a/htdocs/expensereport/card.php
+++ b/htdocs/expensereport/card.php
@@ -2615,7 +2615,7 @@ if ($action == 'create') {
                     let tva = jQuery("#vatrate").find(":selected").val();
                     let qty = jQuery(".input_qty").val();
 
-					let path = "'.DOL_DOCUMENT_ROOT.'/expensereport/ajax/ajaxik.php";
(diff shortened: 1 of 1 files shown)


eldy:
Note that in v20, we try a better code:
let path = "'.DOL_URL_ROOT.'/expensereport/ajax/ajaxik.php";
(no need for dynamic guess of path)

PR #30930, opened 2024-09-11
Qual: Add type hints for Stripe objects

mdeweerd:
# Qual: Add type hints for Stripe objects
    
Suggest the more specific implementation of the StripeObject.


mdeweerd:
Note: I will rebase this once the previous branches are integrated.
Not started from develop to avoid merge conflicts regarding the phan baseline.

PR #30939, opened 2024-09-12
FIX V20 Update societe.class.php FIX#29306

mikko80:
FIX#29306

### DIFF
--- a/htdocs/societe/class/societe.class.php
+++ b/htdocs/societe/class/societe.class.php
@@ -2492,6 +2492,8 @@ class Societe extends CommonObject
 			if (preg_match('/\((.*)\)/', $vatrate, $reg)) {
 				$vat_src_code = $reg[1];
 				$vatrate = preg_replace('/\s*\(.*\)/', '', $vatrate); // Remove code into vatrate.
+			} else {
+				$vatrate = 0;
 			}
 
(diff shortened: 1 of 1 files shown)


htdocs/societe/class/societe.class.php
```diff
@@ -2492,6 +2492,8 @@ public function set_remise_except($remise, User $user, $desc, $vatrate = '', $di
 			if (preg_match('/\((.*)\)/', $vatrate, $reg)) {
```
eldy, line 2492: If $vatrate was provided to 1.23, it seems that after your change $vatrate = 0

htdocs/societe/class/societe.class.php
```diff
@@ -2492,6 +2492,8 @@ public function set_remise_except($remise, User $user, $desc, $vatrate = '', $di
...
 				$vat_src_code = $reg[1];
 				$vatrate = preg_replace('/\s*\(.*\)/', '', $vatrate); // Remove code into vatrate.
+			} else {
+				$vatrate = 0;
```
eldy, line 2496: If $vatrate was provided to 1.23, it seems that after your change $vatrate = 0

mikko80:
Good point. This seems to work without any other changes. This can be cancelled/closed.

PR #30943, opened 2024-09-13
NEW - Add pre-calculated duration of the ticket (when a intervention line is added/modified/deleted) into the field duration of the ticket table

kkhelifa-opendsi:
Add pre-calculated duration of the ticket (when a intervention line is added/modified/deleted) into the field duration of the ticket table and show this on card and list (also public with option)


htdocs/fichinter/class/fichinterligne.class.php
```diff
@@ -337,17 +337,49 @@ public function update_total()
...
+
+			if (!$error && isModEnabled('ticket')) {
+				// Get linked tickets
+				require_once DOL_DOCUMENT_ROOT . '/fichinter/class/fichinter.class.php';
+				$intervention = new Fichinter($this->db);
+				$intervention->id = $this->fk_fichinter;
+				$intervention->fetchObjectLinked(null, "ticket", null, '', 'OR', 1, 'sourcetype', 0);
+				if (!empty($intervention->linkedObjectsIds["ticket"])) {
+					// Update tickets duration
+					$sql = "UPDATE llx_ticket AS t1";
```
eldy, line 353: Can you confirm this king of request is compatible with pgsql ?
May be a loop in PHP would make code easier to understand and more compatible ?

eldy:
Can you first push a PR with structure addition ?

PR #30948, opened 2024-09-13
FIX select list for contacts with very long third party names

W1W1-M:
# FIX select list for contacts with very long third party names
In some cases when third party name is very long the other columns are squashed, especially contact role. This is visible on contacts tab for propals, orders, etc. This fix is a backport from v20.


lvessiller-opendsi (approved):

rycks (approved):
if backported from v20, ok for me

PR #30952, opened 2024-09-13
Update html.form.class.php

fbosman:
# NEW|New [missing fk_soc in hook 'selectProduitsFournisseurListOption']

When using the hook 'selectProduitsFournisseurListOption' and there is no supplier-price found the fk_soc will be empty so in the hook-code you don't know which supplier the order is for. This fix will add the fk_soc to the parameters


htdocs/core/class/html.form.class.php
```diff
@@ -4055,6 +4055,9 @@ public function select_produits_fournisseurs_list($socid, $selected = '', $htmln
 					'outvallabel' => &$outvallabel,
 					'outarrayentry' => &$outarrayentry
 				);
+				if (is_null($objp->fk_soc)) {
```
eldy, line 4058: Obj var represent result of a sql, content must not be modified. To propagate an information to the trigger, you can complete the $parameters array by adding an entry 'socid' => $socid for example

eldy (changes requested):

Issue #29331, opened 2024-04-12
add picture for each user for document signing (ex contract or propal)


RANDOMLY-OBFUSCATED:
### Feature Request

Have the hability to add a personnal signature picture and automatically use it on documents sended to customers for signing 

### Use case

In my profile, I use a file upload button to store my signature which will be added for every documents I will send for signing.
Mine will be add first before sending. This way, the customers will be able to save it just after he signs.

### Suggested implementation

@simogeo made a first shot w/o the html part and users management.
It's here : [issue 26596 (Signature is not apply correctly on document)](https://github.com/Dolibarr/dolibarr/issues/26596#issuecomment-2016508227)

Issue #26916, opened 2023-12-01
Missing signature and name of the intervenant


lmag:
### Feature Request

![image](https://github.com/Dolibarr/dolibarr/assets/3288514/0faa216f-6420-486f-b1d5-5c36505ed7cd)

PR #30954, opened 2024-09-13
NEW intervention online sign for internal

W1W1-M:
# NEW intervention online sign for internal
This PR adds possibility for intervenant (internal user) to use online signing to sign document. This uses `$last_main_doc` as source document.

Currently second signature gets pushed onto next page. Not sure how to fix this yet ...

Associated issues : #26916 #29331


eldy:

Issue #30976, opened 2024-09-16
ODT to PDF Invoice having full file path in their Filename


accellier-dev:
### Bug

When I'm generating the PDF Invoice From ODT template, The file name I'm getting the full path of the file. I'm using **MAIN_DISABLE_PDF_AUTOUPDATE** configuration in other setup.  As you can see in the Screenshot. 
![image](https://github.com/user-attachments/assets/878351db-8117-4176-ba9d-033713ca6189)

### Dolibarr Version

19.0

PR #30979, opened 2024-09-16
Fixed: #30976

accellier-dev:
Bug Fixed: #30976, ODT to PDF Invoice having full file path in their Filename.

### DIFF
--- a/htdocs/includes/odtphp/odf.php
+++ b/htdocs/includes/odtphp/odf.php
@@ -923,9 +923,14 @@ IMG;
 				}
 
 				if (!empty($conf->global->MAIN_DISABLE_PDF_AUTOUPDATE)) {
+					if (preg_match('/[^\/]+\.od(x|t)$/', $name, $matches)) {
+						$fname = $matches[0];
+						$fname = preg_replace('/\.od(x|t)/i', '', $fname);
+					}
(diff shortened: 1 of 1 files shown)


htdocs/includes/odtphp/odf.php
```diff
@@ -923,9 +923,14 @@ public function exportAsAttachedPDF($name = "")
...
 				if (!empty($conf->global->MAIN_DISABLE_PDF_AUTOUPDATE)) {
+					if (preg_match('/[^\/]+\.od(x|t)$/', $name, $matches)) {
+						$fname = $matches[0];
+						$fname = preg_replace('/\.od(x|t)/i', '', $fname);
+					}
 					$name=preg_replace('/\.od(x|t)/i', '', $name);
+					if( $fname == "" ) $fname = $name;
 					header('Content-type: application/pdf');
-					header('Content-Disposition: attachment; filename="'.$name.'.pdf"');
+					header('Content-Disposition: attachment; filename="'.$fname.'.pdf"');
```
eldy, line 933: Why not just doing
header('Content-Disposition: attachment; filename="'.basename($name).'.pdf"');

accellier-dev: Yep, It'll be quick fix, Thanks

PR #30981, opened 2024-09-16
FIX reset date on search filter in list of WebPortal

lvessiller-opendsi:
FIX reset date on search filter in list of WebPortal

**To reproduce**
- filter on date in proposal list in WepPortal
- remove all filters (click on "X" button)
=> So filter date is still here

### DIFF
--- a/htdocs/webportal/class/html.formlistwebportal.class.php
+++ b/htdocs/webportal/class/html.formlistwebportal.class.php
@@ -423,9 +423,8 @@ class FormListWebPortal
 					}
 				}
 			} elseif (preg_match('/(_dtstart|_dtend)$/', $key) && !empty($val)) {
-				$param .= '&search_' . $key . 'month=' . (GETPOSTINT('search_' . $key . 'month'));
-				$param .= '&search_' . $key . 'day=' . (GETPOSTINT('search_' . $key . 'day'));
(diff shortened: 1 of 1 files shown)


htdocs/webportal/class/html.formlistwebportal.class.php
```diff
@@ -487,14 +486,14 @@ public function elementList($context)
 				if (!empty($val['arrayofkeyval']) && is_array($val['arrayofkeyval'])) {
 					$html .= $this->form->selectarray('search_' . $key, $val['arrayofkeyval'], (isset($search[$key]) ? $search[$key] : ''), $val['notnull'], 0, 0, '', 1, 0, 0, '', '');
 				} elseif (preg_match('/^(date|timestamp|datetime)/', $val['type'])) {
-					$postDateStart = GETPOST('search_' . $key . '_dtstart', 'alphanohtml');
-					$postDateEnd = GETPOST('search_' . $key . '_dtend', 'alphanohtml');
+					$postDateStart = empty($search[$key.'_dtstart']) ? '' : GETPOST('search_' . $key . '_dtstart', 'alphanohtml'); // remove search date if empty
```
eldy, line 489: The use of a simple text to manage date fields could be ok on simple form on a single timezone with a single date format string always the same as the value of use browser but can't work into an international context where setup may differ.
So i pushed modification to use the same component as dolibarr.

PR #30996, opened 2024-09-17

thomas-Ngr:

This PR:

SQL Part : #26190

follow-up on #30230 (closed) because file expedition.class has been split into two files.


htdocs/product/class/product.class.php
```diff
@@ -767,6 +767,17 @@ class Product extends CommonObject
 	 */
 	public $mandatory_period;
 
+	/**
+	 * @var int
+	 */
+	public $no_button_copy;
```
eldy, line 773: This property is no more used in develop. Can you remove  it?

htdocs/societe/class/societe.class.php
```diff
@@ -873,6 +873,11 @@ class Societe extends CommonObject
 	 */
 	public $bank_account;
 
+	/**
+	 * @var int
+	 */
+	public $no_button_copy;
```
eldy, line 879: This property is no more used in develop. Can you remove  it?

thomas-Ngr: @eldy I just did it.

htdocs/product/card.php
```diff
@@ -813,6 +816,9 @@
 					$object->fk_default_bom = 0;
 				}
 
+				// managed_in_stock
+				$object->stockable_product   = GETPOSTISSET('stockable_product');
```
eldy, line 820: This line looks strange.
if stockable_product = 0,
you wil get
$object->stockable_product = true.

i think what you want to do is 
$object->stockable_product = GETPOSTINT('stockable_product');

htdocs/product/class/product.class.php
```diff
@@ -769,6 +769,12 @@ class Product extends CommonObject
 	 */
 	public $mandatory_period;
 
+	/**
+	 * 0=This service or product is not managed in stock, 1=This service or product is managed in stock
+	 *
+	 * @var boolean
+	 */
+	public $stockable_product = true;
```
eldy, line 777: Using boolean must be avoid as it can't be used to evoluate.
It is better to use int so
/* @var integer */
public $stockable_product = 1;

htdocs/expedition/card.php
```diff
@@ -362,6 +363,30 @@
...
 				}
 			} else {
+				$p = new Product($db);
+				$res = $p->fetch($objectsrc->lines[$i]->fk_product);
+				if ($res > 0) {
+					if (GETPOST('entrepot_id', 'int') == -1) {
+						$qty .= '_'.$j;
+					}
+
+					if ($p->stockable_product == Product::DISABLED_STOCK) {
```
eldy, line 373: This part of code from line 366 to 389 looks strange:
Currently, default is to have all product managed as stockable.
So we should have code added to avoid to do something when product is not stock managed. But here yo add code to do more things when stock is not managed. 

I think we should not have this part of code from line 366 to 389 and we should have the code to do or not action depending of stockable status somewhere later.
If we remove this part of code, where does the program hangs ?

thomas-Ngr: You are right. I changed the test in Expedition::addline() that forbid to add a line if warehouse was empty. You can see it in commit 2407fe070da8d79a9952e0e0a8a0de2423fe57ef

In case of product that need a batch or a serial, stock management seems necessary.

htdocs/product/card.php
thomas-Ngr: No, since stockable_product is a checkbox. 
- When not checked, $_POST['stockable_product'] is not set
- When checked, $_POST['stockable_product'] is equal to string 'on'. As a consequence, `GETPOSTINT('stockable_product') == 0`.

The exact way to deal with it is `$object->stockable_product = (int) GETPOST('stockable_product') == 'on'`. But it makes useless operations and it does not harm security to `$object->stockable_product   = GETPOSTISSET('stockable_product');` .

I will cast it to (int).

htdocs/product/class/product.class.php
thomas-Ngr: ok

htdocs/expedition/class/expedition.class.php
```diff
@@ -1724,12 +1724,12 @@ public function fetch_lines()
 
 			$shipmentlinebatch = new ExpeditionLineBatch($this->db);
 
+			$line = new ExpeditionLigne($this->db); // always set $line for PHAN analyser. @phan-var-force muse be used after an assignation, and there is no assignation for $line.
```
thomas-Ngr, line 1727: @phan-var-force (l.1732) must be used after an assignation, and we don't have any assignation for $line at the beginning of the loop.

This is the solution I found to pass Phan tests.

source : https://github.com/phan/phan/wiki/Annotating-Your-Source-Code#inline-type-checks-via-string-literals

htdocs/expedition/class/expedition.class.php
```diff
@@ -1724,12 +1724,12 @@ public function fetch_lines()
...
 			while ($i < $num) {
 				$obj = $this->db->fetch_object($resql);
 
-
 				if ($originline > 0 && $originline == $obj->fk_elementdet) {
-					'@phan-var-force ExpeditionLigne $line';  // $line from previous loop
+					// '@phan-var-force ExpeditionLigne $line'; // $line from previous loop
```
thomas-Ngr, line 1732: @phan-var-force (l.1732) must be used [after an assignation](https://github.com/phan/phan/wiki/Annotating-Your-Source-Code#inline-type-checks-via-string-literals), and we don't have any assignation for $line at the beginning of the loop.

Declaring $line before starting the loop is the solution I found to pass Phan tests.

PR #31003, opened 2024-09-17
New unbilled btn for supplier order

thersane-john:
# NEW

New unbilled button  for  supplier order.

@eldy do we need this?
I use it for myself, but I don't know if the community needs it.


eldy:
It seems there is already a button Reopen.
What difference between the button Reopen and the button Unbill ?

thersane-john:
@eldy yes there is a reopen button but, when we trigger it change status and **"billed"** flag.

See there is a button for classify billed and for reopen
image

But when the order is billed there is no button for "classify unbilled"
image

In some cases in the warehouse, employees need to reopen the order but Dolibarr has lost the invoiced flag... and for these employees, managing invoiced or uninvoiced is not in their scope of work. As a result, it can give headaches to the accounting manager.


PS : @ htdocs/fourn/commande/card.php line 374 there is this comment : 
`// Currently the "Re-open" also remove the billed flag because there is no button "Set unpaid" yet.`
I remove it with this PR

eldy:
I see. Yes we should have a separation between status of order and status of invoicing. It is now 2 different fields in db and different workflows.
Reopen should not change bill status, only status of order.
Let'go with your PR...

PR #31005, opened 2024-09-17
fix phan

frederic34:


mdeweerd:
@frederic34 Thanks for fixing the notices.
@eldy:
The code copied from propal references a property no_button_copy that does not exist on Societe (and not on Product either).  This PR fixes the notice by ignoring the error in the baseline.txt file, which IMHO suggests that it is to be fixed later, otherwise we could ignore the specific cans in the source file itself.

If this is a valid property under certain conditions, I think it has to be documented in the code.

Personally I think that (apparent) bugs introduced by new code should be properly documented and locally ignored, or fixed when they are added to the codebase.

eldy:
> @frederic34 Thanks for fixing the notices. @eldy: The code copied from propal references a property no_button_copy that does not exist on Societe (and not on Product either). This PR fixes the notice by ignoring the error in the baseline.txt file, which IMHO suggests that it is to be fixed later, otherwise we could ignore the specific cans in the source file itself.
> 
> If this is a valid property under certain conditions, I think it has to be documented in the code.
> 
> Personally I think that (apparent) bugs introduced by new code should be properly documented and locally ignored, or fixed when they are added to the codebase.

@mdeweerd I agree we must avoid to add exception. I fixed the use of ->no_button_...  so no new exception is required.

PR #31012, opened 2024-09-18
Update objectline_create.tpl.php extra hooks for extra fields

fbosman:
# NEW Add hooks to objectline_create
In order to act on the selection of the product I added two hooks: one on the event of the product selected and one on the event of the extra-field selected. 
This way we are able to add extra functionality on the selection of product cq the extrafield.


htdocs/core/tpl/objectline_create.tpl.php
```diff
@@ -1073,6 +1073,8 @@ function(data) {
 			} else { jQuery("#pbq").val(''); }
 			<?php
 		}
+		$reshook = $hookmanager->executeHooks('objectlinecreatesupplierorderproductselected', $parameters, $object, $action);
```
eldy, line 1076: As the name of hook is also a php function, can you use the camelCase syntax for the name ofthe hook...

fbosman: done

PR #31031, opened 2024-09-19
FIX : max product combo limit

thersane-john:
# Instructions
product combo search have no limit....
So combo can timeout when your database return more than 100000 products


eldy:
Product combo already has a limit.
It is defined into the setup if module product.
Set to 1000 by defaut if option is preload all products.

Do you mean there is no limit for the case we are using the option "show list after typing n characters" ?

thersane-john:
> Do you mean there is no limit for the case we are using the option "show list after typing n characters" ?

@eldy yes, that's it !

eldy:
> > Do you mean there is no limit for the case we are using the option "show list after typing n characters" ?
> 
> @eldy yes, that's it !

So, if we introdice the limit also for ajax combo, we should be able to set it into setup. Currently, if we are in mode combo with ajax, the limit parameter is hidden.
Can you make it visible...

thersane-john:
image

eldy:
> image

On your screenshot, change the setup of the firs line
"Attendre que vous ayez..." to the value "No".
You see that a new line will appears to be able to define the limit of product.
This setup must remain always active if we use it whetever is choice of first line.

thersane-john:
@eldy Done !

PR #31048, opened 2024-09-19
Fix substitution variables in notification quote signed by customer

MathildeMar:
Add fetch_thirdparty before getCommonSubstitutionArray to initialize substitution variables of type __THIRDPARTY_XXX__ in notification mails


htdocs/core/class/notify.class.php
```diff
@@ -938,6 +938,9 @@ public function send($notifcode, $object, $filename_list = array(), $mimetype_li
 							$arraydefaultmessage = $formmail->getEMailTemplate($this->db, $object_type.'_send', $user, $outputlangs, 0, 1, $labeltouse);
 						}
 						if (!empty($labeltouse) && is_object($arraydefaultmessage) && $arraydefaultmessage->id > 0) {
+							if (empty($object->thirdparty)) {
+								$object->fetch_thirdparty();
```
eldy, line 942: The method may not exists for some objects

eldy (changes requested):

lainwir3d:
Now that this PR has been fixed as per your comment @eldy, can this be merged please?

This would allow BILL_PAYED /  ORDER_CLOSE and probably other notifications  to use _THIRDPARTY_XXXXXXX_ substitution family.

Cheers

PR #31052, opened 2024-09-20
FIX : FATAL ERROR abusively triggered due to incomplete regex

thersane-john:
# FIX 
FATAL ERROR abusively triggered due to incomplete regex

`SELECT [...] WHERE [...] LIMIT 50`  **OK**
`SELECT [...] WHERE [...] LIMIT 50  ` **FAIL**
`SELECT [...] WHERE [...] LIMIT 50 ;` **FAIL**
`SELECT [...] WHERE [...] LIMIT 25,50;` **FAIL**
`SELECT [...] WHERE [...] LIMIT 25,50` **FAIL**

After:

`SELECT [...] WHERE [...] LIMIT 50`  **OK** 
`SELECT [...] WHERE [...] LIMIT 50  ` **OK** 
`SELECT [...] WHERE [...] LIMIT 50 ;` **OK** 
`SELECT [...] WHERE [...] LIMIT 25,50;` **OK** 
`SELECT [...] WHERE [...] LIMIT 25,50` **OK** 
`SELECT [...] WHERE [...] LIMIT` **FAIL** 
`SELECT [...] WHERE [...] `  **FAIL**

### DIFF
--- a/htdocs/core/db/DoliDB.class.php
+++ b/htdocs/core/db/DoliDB.class.php
@@ -421,7 +421,7 @@ abstract class DoliDB implements Database
 	 */
 	public function getRows($sql)
 	{
-		if (!preg_match('/LIMIT \d+,?\d*\ *;?$/', $sql)) {
+		if (!preg_match('/LIMIT \d+(?:(?:,\ *\d*)|(?:\ +OFFSET\ +\d*))?\ *;?$/', $sql)) {
 			trigger_error(__CLASS__ .'::'.__FUNCTION__.'() query must have a LIMIT clause', E_USER_ERROR);
 		}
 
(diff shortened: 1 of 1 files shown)


eldy:
SELECT [...] WHERE [...] LIMIT 25,50; is a wrong syntax. A limit accept only an int parameter.
So it should fails like it does currently.
Only the addition of ;? seems good to me.

hregis:
> SELECT [...] WHERE [...] LIMIT 25,50; is a wrong syntax. A limit accept only an int parameter. So it should fails like it does currently. Only the addition of ;? seems good to me.

@eldy you can use : 

`LIMIT offset, row_count;`

or 

`LIMIT row_count OFFSET offset`

thersane-john:
Add OFFSET 
image

PR #31053, opened 2024-09-20
FIX : To have only the sender's entity when sending mail

Thomas905:


lvessiller-opendsi:
Thanks for your contribution.
First, you should fix the indentation code and commit again to resolve the pre-commit errors.
For me, it's not obvious to understand this FIX without description.
Could you give us more details and how to reproduce ?

Concerning the filter for user entity in SQL request, it's recommended  to use "getEntity('user')" instead of "(0, $conf->entity)".
@hregis, is it right ?

hregis:
@lvessiller-opendsi @Thomas905 this is a dictionary

getEntity('c_email_senderprofile')

rycks (changes requested):
waiting for same solution as @hregis and @lvessiller-opendsi  :-)
i discover that getentity can work with dictionnary

please update your commit :)

rycks:
@Thomas905 could you please update your PR ?

eldy:
@rycks @lvessiller-opendsi When a PR has no answer like this one after a long time and when the change to do is minor, a good way to progress is to do the change ourself from the github interface directly by using the "Edit file" available on tab "Files changed":
See on right of this screenshot.
image

We know here we just have to replace spaces by tab and we got the line to change. So it takes less time to do it than to send reminder.
Do you have permission to use this feature ? Can one of you try it ?

rycks:
Hello @eldy 
i don't have that super power :-)

image

lvessiller-opendsi:
@eldy 
I have the same rights as @rycks and I can't edit the file.

eldy:
It seems that github has recently modified the permissions to do the edit. 
@rycks, @lvessiller-opendsi  Still not able to make the change from github Edit link ?

lvessiller-opendsi (approved):

PR #31062, opened 2024-09-20
NEW: Public and private note options in thirdparty list

sonikf:


htdocs/societe/list.php
```diff
@@ -2157,6 +2179,24 @@
...
+			print dolPrintHTML($obj->note_public);
+			print '</td>';
+			if (!$i) {
+				$totalarray['nbfield']++;
+			}
+		}
+		// Note private
+		if (!empty($arrayfields['s.note_private']['checked'])) {
+			print '<td class="flat maxwidth250imp">';
+			print dolPrintHTML($obj->note_private);
```
eldy, line 2194: Because content of notes can be very large html content, i recommand to do
dolPrintHtml(dolGetFirstLineOfText(note, 5))
Can you try this ?

eldy (changes requested):

htdocs/societe/list.php
sonikf: Done

PR #31068, opened 2024-09-20
Qual deprecated line->pu_ttc

W1W1-M:
# Qual deprecated line->pu_ttc
This fixes some deprecated warnings about using `$pu_ttc` property on line objects. From what I understand `$total_ttc` is to be used, also with object lines.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -5327,7 +5327,7 @@ abstract class CommonObject
 				$description .= (getDolGlobalInt('PRODUIT_DESC_IN_FORM_ACCORDING_TO_DEVICE') ? '' : (!empty($line->description) ? dol_htmlentitiesbr($line->description) : '')); // Description is what to show on popup. We shown nothing if already into desc.
 			}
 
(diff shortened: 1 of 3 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -5327,7 +5327,7 @@ public function printObjectLine($action, $line, $var, $num, $i, $dateSelector, $
 				$description .= (getDolGlobalInt('PRODUIT_DESC_IN_FORM_ACCORDING_TO_DEVICE') ? '' : (!empty($line->description) ? dol_htmlentitiesbr($line->description) : '')); // Description is what to show on popup. We shown nothing if already into desc.
 			}
 
-			$line->pu_ttc = price2num((!empty($line->subprice) ? $line->subprice : 0) * (1 + ((!empty($line->tva_tx) ? $line->tva_tx : 0) / 100)), 'MU');
+			$line->total_ttc = price2num((!empty($line->subprice) ? $line->subprice : 0) * (1 + ((!empty($line->tva_tx) ? $line->tva_tx : 0) / 100)), 'MU');
```
eldy, line 5330: total_ttc  contains up * qty
Here deprecation is to say we must avoid use the up when we need to calculate the total_ttc and reuse the already calculated total_ttc.
But this is the good variable here.

PR #31072, opened 2024-09-20
Qual: Fix copyright fixer

mdeweerd:
# Qual: Fix copyright fixer

Enable actual file write, disable debug message


frederic34:
@mdeweerd can you check what happen if (C) 2018-2023 is updated (it give 2023-2024... instead 2018-2024)

mdeweerd:
@frederic34 Thank you for the observation & that's fixed now.

PR #31074, opened 2024-09-21
Qual: Fix pdf.lib.php, project.lib.php and related files

mdeweerd:
# Qual: Fix pdf.lib.php, project.lib.php and related files

Fix most phan notices for pdf.lib.php and project.lib.php.
And fixed some items in other files in the process


mdeweerd:
Merged with other PR.

Issue #31054, opened 2024-09-20
Feature : Add third-party name in tooltip of linked object lines


lvessiller-opendsi:
### Feature Request

Add third-party name in tooltip of linked object lines
- on each card you have an linked object block
- this block show all linked object and when you put your mouse over the reference, you got an "tootltip" with some information of this object (like ref) and it would be interesting of having third-party name in this "tooltip"

### Use case

**In develop**
In the last version of Dolibarr, we have the linked object block : 
![image](https://github.com/user-attachments/assets/ee67b720-d215-4890-8461-b8f524e7ab8e)

With this "tooltip" when you put your mouse over the "ref" of the object : 
![image](https://github.com/user-attachments/assets/c2c28aca-6952-439e-bfda-f782d89d2804)

**Feature**
The "tooltip" will look like :
![image](https://github.com/user-attachments/assets/75a94eaf-37fb-4ac2-9c2c-d7f4b3aac544)

### Suggested implementation

The block to show linked object call a template for each object.
Each template are in the element directory of the object and are named "linkedobjectblock.tpl" (such as : "expedition/tpl/linkedobjectblock.tpl.php").
Then, the linked object call "getNomUrl()" to show the reference as label and add a link to access to its card.
And the "getNomUrl()" method calls "getTooltipContentArray()" method to show the tooltip when you put your mouse over.
Finally, the "getTooltipContentArray()" was called with an array of params : 
- id
- objecttype
- option
- nofetch

I suggest to complete the "getNomUrl()" method in order to add "with_third_name" in last argument (default value is 0) :
- so "params" array will be used in each template "linkedobjectblock.tpl" and the "nofetch" option will be set to 0
- so the third-party with name will be shown in the tooltip

PR #31100, opened 2024-09-23
NEW third-party name in tooltip of linkedobject templates

lvessiller-opendsi:
NEW third-party name in tooltip of linkedobject templates
- see features https://github.com/Dolibarr/dolibarr/issues/31054


eldy:
Because the load of thirdparty may need a fetch and because the getNomUrl is a low level function that can be called frequently inside a loop, so the load of the thirdparty must not be done inside the getNomUrl (except if function is called to show the tooltip using ajax).
Why not just showing the thirdparty if the $this->thirdparty object is defined.
It will be the decision to the caller to load or not the thirdparty()

lvessiller-opendsi:
I would have liked to have the third-party name in linked objects, but it seems not the good way to do.
If third-party is not already fetched, it won"t be show in the tooltip.
So I will change the way to do it.

PR #31113, opened 2024-09-24
Qual: More specific ignoreErrors for phpstan

mdeweerd:
# Qual: More specific ignoreErrors for phpstan

More specific ignore expressions helps identify new exceptions that would have matched the more generic ignore expressions.

It also helps to progressively solve phpstan notices in smaller steps by modifying the expressions


htdocs/admin/dict.php
```diff
@@ -1775,7 +1775,7 @@
 				}
 
 				if (empty($reshook)) {
-					fieldList($fieldlist, $obj, $tabname[$id], 'add');
+					dict_fieldList($fieldlist, $obj, $tabname[$id], 'add');
```
eldy, line 1778: If we rename a function, we must use a camelcase format function name
For example dictFieldList()

mdeweerd: @eldy Good to know, I changed to camelcase.

PR #31117, opened 2024-09-25
Fix wrong invoice status when invoice has discount

UltraViolet33:
$object->totalpaid in compta/facture/card is set with CommonInvoice::getSommePaiement() method.
So $object->totalpaid has only the paiements from the payment table.
Payments done using discounts, credit notes, etc are not included in $object->totalpaid.

So when we add a discount to an invoice, the status stays 'unpaid' instead of 'started' because $object->totalpaid stays 0.
And when we want to close this invoice as 'partially paid', the status is 'closed (unpaid)' instead of 'partially paid'.

This PR fixe that so the invoice status is 'started' when we add a discount and 'partially paid' when we close this invoice as 'partially paid'


htdocs/compta/facture/card.php
```diff
@@ -4657,8 +4657,7 @@ function setRadioForTypeOfInvoice() {
 	}
 	$morehtmlref .= '</div>';
 
-	$object->totalpaid = $totalpaid; // To give a chance to dol_banner_tab to use already paid amount to show correct status
-
+	$object->totalpaid = $totalpaid + $totalcreditnotes + $totaldeposits; // To give a chance to dol_banner_tab to use already paid amount to show correct status
```
eldy, line 4660: For consistency of meaning of variable, totalpaid must remain to amount of payments only.
However, I initialize the other properties ->totalcreditnote, ->totaldeposits and ->remaintopay...
Is it enough ?

UltraViolet33: I used these properties in CommonInvoice::getLibStatut() to get the good invoice status. Is it OK ?

eldy: Things are not clear at a lot of place of code.
To I pushed a change to standardize the code. The amount already paid in parameter of getLibStatus should be already consolidated (amount of payment + credit notes + deposit). it was sometimes and sometimes not.

UltraViolet33: I create another PR: https://github.com/Dolibarr/dolibarr/pull/31300.
Your commit does not fix the status problem when the status is displayed with dol_banner_tab() function. My PR fix that.

PR #31119, opened 2024-09-25
NEW Do not include cancelled orders in the billable filter

zahwyrix:
# NEW Do not include cancelled orders in the billable filter


htdocs/commande/list.php
```diff
@@ -941,6 +941,9 @@
 }
 if ($search_billed != '' && $search_billed >= 0) {
 	$sql .= ' AND c.facture = '.((int) $search_billed);
+	if (getDolGlobalString('HIDE_CANCELLED_ORDERS_BILLABLE') == 1) {
```
eldy, line 944: If the user need both a filter on 
billed
+
status of order that is not canceled.

The user can set the two filters to match its need. Why adding an option to replace silently a filter that should be already available to end user.

zephyriony: Logically, by definition, a cancelled order is not billable.
I had the same reaction as you when this remark was made to me in my company, but the secretary told me that it was a pain to have to select both the invoiced and the three statuses.
Especially as Dolibarr gives the wrong information, for me it's almost a bug fix, it's cancelled and can't be billed.

zephyriony: image
image

There are other orders that I have removed for reasons of confidentiality.

Currently, as you suggested, you can use the “Status” filter, but it's a pain to select all three open statuses.
This seems to me to be a bug, since I think that in the majority of companies, a cancelled order is not billable.
The total amount is therefore not correct, the two cancelled orders need to be removed. Indeed, as you say, we could filter by status, but that would take longer and be illogical, I insist, there's a lack of consistency, in most cases we don't invoice a cancelled order.

eldy: I see the filter "Billed" is already set correctly.
To reach this page, where does he/she click ?  From the entry "Billlable orders" in the left menu ?
I check with last version of v19, v20 and the filter "Status" that exclude draft order is correctly set automatically.
What is the version used ?

PR #31120, opened 2024-09-25
Fix Ticket message file

MaximilienR-easya:
# Fix

Fix the file linked to ticket message on the document page.

Now we have a separate list for the files linked to the messages for more visibility, and we can modify or delete them without problem.


eldy:
How do you reproduce the bug ?

MaximilienR-easya:
> How do you reproduce the bug ?

It's in accord with this PR #30554 : 

> Futhermore, edit and delete files doesn't work due to the wrong module part passed into list of documents method.
> The module part is "actions" for the attached files in a message sent by mail.
> The module part "ticket" is correct when you join a file from the ticket card.
> ...
> In the case of an attached file in an event, the module part is "actions" (not "ticket") and I prevent the user from
> edit / delete theses files directly from the ticket card. He can do it from the linked event.

I fixed it to be able to do modify /delete without problem the files linked to the messages

eldy:
We experienced a lot of problem in v20 due to storing files of ticket into a different directory than tickets. 
Pb when using api, when using webdav, pb in update and delete files, pb with thumbs generation, pb with permissions and more...

So i had to revert all the feature in v20.0.1 to restore stability of v20 branch.

After analysis, i think we made a conception error when implementing this feature. Reason of so many trouble is that, in Dolibarr philosophy, files are attached to business objects (framework is based on this) but agenda is not a business object but a transverse object (related to any business object) so we must implement a transverse solution too.

The conclusion is that we must implement this feature differently (working for any case, not only ticket with specific code).
So i added a column agenda_id in develop into table llx_ecmfiles.
When we upload a file to a business object, we can also store the id of event into this table too. So we have our link to know which file is linked to which event.

I commented code that was specific to ticket so we can now implement the new way of working. All places to implement the new feature can be found by searching on agenda->dir_output. If we have code commented with a mention "disabled...", it means we can implement the new way of working.
Sorry to force you to change this but old one was a wrong decision (my fault) and to avoid to have it used with no way to change it later, i prefer to revert it immediately before it is too late to fix it later.

PR #31127, opened 2024-09-25
Qual: Fix type errors discovered by CI

mdeweerd:
# Qual: Fix type errors discovered by CI

This fixes the current CI errors, mostly related to typing, two because of a missing call to trans.

### DIFF
--- a/htdocs/compta/prelevement/factures.php
+++ b/htdocs/compta/prelevement/factures.php
@@ -3,6 +3,7 @@
  * Copyright (C) 2005-2017 Laurent Destailleur  <eldy@users.sourceforge.net>
  * Copyright (C) 2005-2009 Regis Houssin        <regis.houssin@inodbox.com>
  * Copyright (C) 2010-2012 Juanjo Menent        <jmenent@2byte.es>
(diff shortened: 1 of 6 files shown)


mdeweerd:
@eldy 
I think that the following is wrong as $projectsListId is a string that can correspond to multiple project ids seperated with ',' in print_projecttasks_array or project.lib.php .
```php
    $sql .= " AND p.rowid IN (".$db->sanitize($projectsListId).")";
```

Edit:
Also sanitized like that in projet/index.php

Edit2:
I did not fix these notices regarding projectsListId - I do not know how lists as strings are properly sanitized in Dolibarr.

eldy:
> @eldy 
> I think that the following is wrong as $projectsListId is a string that can correspond to multiple project ids seperated with ',' in print_projecttasks_array or project.lib.php .
> ```php
>     $sql .= " AND p.rowid IN (".$db->sanitize($projectsListId).")";
> ```
> 
> Edit:
> Also sanitized like that in projet/index.php
> 
> Edit2:
> I did not fix these notices regarding projectsListId - I do not know how lists as strings are properly sanitized in Dolibarr.

List string of id are sanitized with db->sanitize(list string separated by comma)

Issue #31131, opened 2024-09-26
Custom theme JS file rendering issue


alsoft10:
### Bug

As there is no instruction to the browser that the ALLOW_THEME_JS file is JavaScript file. Because there is no type attribute in the line `print '<script nonce="'.getNonce().'" src="'.DOL_URL_ROOT.'/theme/'.$conf->theme.'/'.$conf->theme.'.js'.($ext ? '?'.$ext : '').'"></script>'."\n";` This would be normal Dolibarr behavior if you you would render the file with `.php` extension. 

### Dolibarr Version

20.0

### Environment PHP

8.1.25

alsoft10:
@eldy still this bug is present at Dolibarr 20.0.1

eldy:
Pb seems into the js file you call.
Can youbcheck this file is sending a mime type ?

alsoft10:
@eldy I checked it once again carefully. And it is solved. Thank you.

PR #31132, opened 2024-09-26
Fix for issue #31131

alsoft10:
#31131 has been fixed.


alsoft10:
@eldy #31131 has fixed

eldy:
As this change an old behavior, can you push it into develop branch.

alsoft10:
@eldy If that is the case then how about doing this, `<script type="text/javascript"` instead adding the `.php` extension. That would not change old behaviour, I hope.

Issue #31056, opened 2024-09-20
Table names in object.class.php


dolibarr95:
### Feature Request

Hello

Just a question or a misunderstanding :
 Why in object classes we declare the name of the table.
 For example:

```php
/**
 * @var string Name of table without prefix where object is stored
 */
public $table_element = 'commande';
```
But in the queries we write the name of the table
```php
$sql = 'UPDATE '.MAIN_DB_PREFIX.'commande SET facture = 1';
$sql .= " WHERE rowid = ".((int) $this->id).' AND fk_statut > '.self::STATUS_DRAFT;
```
and not:
```php
$sql = 'UPDATE '.MAIN_DB_PREFIX.$this->table_element .' SET facture = 1';
$sql .= " WHERE rowid = ".((int) $this->id).' AND fk_statut > '.self::STATUS_DRAFT;
```
Tks for reading.

zahwyrix:
Probably because it was done after the fact, or because the developer wasn't aware that this property existed, you can make a pull request. In my opinion, you should actually do this

PR #31137, opened 2024-09-26
Update sql queries

dolibarr95:
Replace the values : 
commande : $this->commande
table_element : $this->table_element_line
in the sql queries.
see https://github.com/Dolibarr/dolibarr/issues/31056


dolibarr95:
Hello
@eldy I will do the other classes

Issue #31103, opened 2024-09-24
Auto assign user at creation


MaximilienR-easya:
### Feature Request

Auto assign the user at creation if no user was set assigned at creation.
Actually the option that exist TICKET_AUTO_ASSIGN_USER_CREATE force the user even if you select someone else as the assigned user at creation.

### Use case

You want to create a ticket and be sure that someone is assigned even if you forget to fill the assigned user field but have the possibility to assign whoever you want at creation

### Suggested implementation

Either :
1. Modify TICKET_AUTO_ASSIGN_USER_CREATE to check if a user is already assigned.
2. Have 2 options : 

- TICKET_AUTO_ASSIGN_USER_CREATE : Auto assign the user at creation if no user was set assigned at creation.
- TICKET_FORCE_ASSIGN_USER_CREATE : Force the user as the assigned user (actual TICKET_AUTO_ASSIGN_USER_CREATE)

PR #31140, opened 2024-09-26
NEW #31103

zahwyrix:
#New #31103


eldy:
Because the 2 options are exclusive, we don't need another variable. We just need another value. So
TICKET_AUTO_ASSIGN_USER_CREATE = 0   No auto assignement
TICKET_AUTO_ASSIGN_USER_CREATE = 1   Auto assign current user if no user was set
TICKET_AUTO_ASSIGN_USER_CREATE = 2   Auto assign current user even if another user was set.

PR #31142, opened 2024-09-26
FIX : Wrong price for BOM with warkstation

atm-adrien:
## FIX : Wrong price for BOM with warkstation

### DIFF
--- a/htdocs/bom/tpl/objectline_view.tpl.php
+++ b/htdocs/bom/tpl/objectline_view.tpl.php
@@ -177,13 +177,19 @@ if ($filtertype != 1) {
 	print '</td>';
 
 	// Work station
-	if (isModEnabled('workstation')) {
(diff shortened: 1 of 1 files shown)


htdocs/bom/tpl/objectline_view.tpl.php
```diff
@@ -177,13 +177,19 @@
 	print '</td>';
 
 	// Work station
-	if (isModEnabled('workstation')) {
+	if (isModEnabled('workstation') && !empty($line->fk_default_workstation)) {
```
eldy, line 180: If there is 2 lines, one with default workstation and one without,
we will get a <td> for one and not for the other.
This introduces surely a bug.

atm-adrien: Done

Issue #31163, opened 2024-09-27
Table llx_product_perentity is not created at the time of Dolibarr fresh install


alsoft10:
### Bug
While you install Dolibarr first time, then the table `llx_product_perentity` is not created in both Windows and Linux OS. This is due to inappropriate filename `llx_product_perentity-multicompany.sql` and `llx_product_perentity-multicompany.key.sql`. As there is a DASH character in the filename OS is not give those filename to the installation script.

### Dolibarr Version

19.0, 20.0

### Environment PHP

8.1.25

### Environment Database

MySQL

hregis:
@alsoft10 this table is created if you have and activated the Multicompany module...

hregis:
@alsoft10 Do you have the Multicompany module? If not, do you still have errors concerning this table?

alsoft10:
I don't have Multicompany module. But it will create problem at the time upgrading from Dolibarr 17.0 to any next version as that table will not be created at first install for Dolibarr version 16.0, 17.0, 18.0, 19.0 and 20.0.

hregis:
@alsoft10 i don't have error if table not exists !! what is your error ?

alsoft10:
You will get error while you will upgrade and your first install version was (16.0, 17.0, 18.0 or 19.0). Any user who installed Dolibarr in these four version they will get error at the time of upgrade.

hregis:
@alsoft10 can you send a screenshot of error please, thanks

eldy:
The table wil be created when you will enable the moduke multicompany.
During a migration,
If table exists, it will be upgraded.
If not, nothing will be done (upgrade should fails with trapped errors that should not appear to end user). And this is not a problem because as soon as you need it, when module multicompany will be enabled, it will be created with the last version.

alsoft10:
@eldy thanks for your reply. It make sense now. Sorry for the inconvenience caused!

PR #31164, opened 2024-09-27
FIX#31163

alsoft10:
Bug #31163 has been fixed


hregis:
@alsoft10 sorry it's not good, i don't accept this rename

hregis:
@alsoft10 the name of files are goods

PR #31172, opened 2024-09-27
NEW: PDF Add option to print rounded corner frames

sonikf:
Hi @eldy 
Do you approve so i can continue with all other templates?

εικόνα

corner radious=1

corner radious=2

corner radious=3


eldy (approved):

htdocs/core/class/commondocgenerator.class.php
```diff
@@ -98,7 +99,7 @@ abstract class CommonDocGenerator
 	public $marge_droite;
 	public $marge_haute;
 	public $marge_basse;
-
+	public $corner_radius;
```
frederic34, line 102: @sonikf can you add phpdoc for the property ?

sonikf: Done

PR #31200, opened 2024-09-30
NEW Accountancy - Option to select the label operation

aspangaro:
Add an option to generate label operation before transfer in accountancy

Capture d’écran du 2024-09-30 09-41-46

image


htdocs/accountancy/journal/expensereportsjournal.php
```diff
@@ -539,7 +553,7 @@
 				print '"'.$date.'"'.$sep;
 				print '"'.$val["ref"].'"'.$sep;
 				print '"'.length_accountg(html_entity_decode($k)).'"'.$sep;
-				print '"'.dol_trunc($accountingaccount->label, 32).'"'.$sep;
+				print '"'.mb_convert_encoding($bookkeepingstatic->accountingLabelForOperation($userstatic->name, '', $accountingaccount->label), 'ISO-8859-1').'"'.$sep;
```
eldy, line 556: To be compatible with the rest of export that need to manage CSV export, i have moved the method csvClean into the file functions2.lib.php
You can use it instead of mb_convert_encoding with
print '"'.csvClean($bookkeepingstatic->accountingLabelForOperation($userstatic->name, '', $accountingaccount->label), 'ISO-8859-1', $sep).'"'.$sep;

So the encoding into ISO will be done but will follow to the setup of user for other country that does not use ISO for CSV.

aspangaro: Thanks @eldy 

I push a commit with use of the csvClean() function

Issue #31206, opened 2024-09-30
Ticket markAsRead function does not update the status to STATUS_READ


cbeasley-sbb:
### Bug

Clicking the "Read" status button does not update the object's fk_statut attribute resulting in the object's status remaining the same as before the action.

The ticket.class.php 'markAsRead' function executes a database update for the field.  However, it does not set '$this->fk_statut' to Ticket::STATUS_READ.  As a result, if there are triggers that update the object the status is reverted back to the original status.

If the following is added somewhere before or after the database update the problem is resolved.

`$this->fk_status = Ticket::STATUS_READ;`

### Dolibarr Version

18.0

### Environment PHP

PHP 7.x

### Environment Database

10.11.3-MariaDB-1 - Debian 12

### Steps to reproduce the behavior and expected behavior

1. Open a ticket card.
2. Click the 'Read' status button.
3. Observe the status did not change.

rycks:
hello @cbeasley-sbb i can't reproduce that bug, here is the result when i click on "read" (sorry for the screenshot in french) on dolibarr 18

![image](https://github.com/user-attachments/assets/38fdc330-9f62-4278-92f3-0e3adb0a5834)

rycks:
That bug seems to be present only if there is a trigger is that true ?

What could we do to reproduce the bug ? what trigger is called from what module ?

Maybe you could put dolibarr as debug mode, collect the log, clean private part and publish it here to help us ?

Thanks

rycks:
i think with dolibarr debug we could read something like that:

(dolibarr core) update database with status read (but do not update object attribute status)
(trigger) update database with all current object attributes -> then sql will make update with "old status" (so not read in that case)


i search into other part of dolibarr and for example there is a solution like your for setDraft on invoices (facture.class.php) but it seems to be different in commande.class.php where attribute status is set at the end of function after trigger ...

so @eldy i think we need your expertise for that sort of things because there is different solutions on different objects, then it could be better to make a choice and publish a note with "good solution is ..." and apply it in all dolibarr objects

eldy:
The good solution is the one suggested by cases of modulebuiler: When we update a status or a property, we first use the this->oldcopy = dol_clone($object, 2) to make a copy.
Then we set status and properties and persist them in database.
if error, we revert properties in memoery by using the value into the backup.

[...]

PR #31207, opened 2024-09-30
Bug-Fix 31206 -Update ticket.class.php markAsRead function

cbeasley-sbb:
Fix Issue #31206  

Sets the object's  "fk_statut" to Ticket::STATUS_READ. In the event of rollback reverts the object's "fk_statut" to its old value.

### NOTE:  I need this fix to be applied to v18.x if and when approved.

### DIFF
--- a/htdocs/ticket/class/ticket.class.php
+++ b/htdocs/ticket/class/ticket.class.php
@@ -1706,8 +1706,10 @@ class Ticket extends CommonObject
 
 		if ($this->status != self::STATUS_CANCELED) { // no closed
 			$this->oldcopy = dol_clone($this, 2);
-
 			$this->db->begin();
+
+			$oldStatus = $this->fk_statut;
+			$this->fk_statut = Ticket::STATUS_READ;
 
 			$sql = "UPDATE ".MAIN_DB_PREFIX."ticket";
 			$sql .= " SET fk_statut = ".Ticket::STATUS_READ.", date_read = '".$this->db->idate(dol_now())."'";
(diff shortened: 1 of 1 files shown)


lvessiller-opendsi:
Thanks for your remarks
Why not only set the status before db commit : (line 1735)
```
if (!$error) {
        $this->fk_statut = Ticket::STATUS_READ;
	$this->db->commit();
	return 1;
}
```

cbeasley-sbb:
> Thanks for your remarks Why not only set the status before db commit : (line 1735)
> 
> ```
> if (!$error) {
>         $this->fk_statut = Ticket::STATUS_READ;
> 	$this->db->commit();
> 	return 1;
> }
> ```

That could work.  However, in other trigger situations the primary transaction for the event has already been executed but not committed.  This way is more consistent with that pattern.  With the approach you suggest $object->old_copy will be identical to the object that has supposedly changed during trigger function execution(s).

1. change executed
2. run_triggers called
3. Trigger functions evaluate the transition with custom logic; Presumably deltas can be identified by comparing $object->X to $object->old_copy->X
4.  main function commits the changes if no errors.

eldy:
A fix was pushed into v18. When merging to higher branch, upstream code was modified to use the ->oldcopy instead of $oldStatus var.
16ab86a9aedb2b1ed97588d9f039dd4b05306b03

@lvessiller-opendsi "Why not only set the status before db commit : (line 1735)"
The set is not required just before the commit because we must do it (the set of values) at begin to have all the new value loaded into the object for the manipulation that will come later (update, trigger, ...)
So setting at end in case of commit should be useless, however, we must revert value in case of rollback using the old value defined into ->oldcopy.

PR #31215, opened 2024-10-01
Fix $ in key of button action

thersane-john:
# FIX
Fix $ in key of button action

I don't undestand why there is $ in keys


htdocs/core/lib/functions.lib.php
```diff
@@ -12309,9 +12309,9 @@ function dolGetButtonAction($label, $text = '', $actionType = 'default', $url =
 				$text = $button['text'] ?? '';
 				$actionType = $button['actionType'] ?? '';
 				$tmpUrl = DOL_URL_ROOT.$button['url'].(empty($params['backtopage']) ? '' : '&amp;backtopage='.urlencode($params['backtopage']));
```
thersane-john, line 12311: Question ?
Why ?
$tmpUrl = DOL_URL_ROOT.$button['url'].(empty($params['backtopage']) ? '' : '&amp;backtopage='.urlencode($params['backtopage']));

thersane-john: I created a fix see #31217

PR #31217, opened 2024-10-01
Fix 20.0 button dropdown part02

thersane-john:
need #31215 PR before

# FIX Ajax loaded button or if button was added after dom ready by js
see comment

# FIX Abusive Url completion 
Dropdown url doesn't work for non core url or javascript url

I have created an anonymous function to factor code during the fix

# FIX inconsistent case  isDropDown => isDropdown
keep isDropdown not isDropDown because Dropdown is used as a word 

@eldy Sorry I can't split this one... the 3 problems are in same part of code.


htdocs/core/js/lib_head.js.php
```diff
@@ -1321,7 +1321,7 @@ function price2numjs(amount) {
 
 
 jQuery(document).ready(function() {
-	jQuery(".butAction.dropdown-toggle").on("click", function(event) {
+	jQuery(document).on("click", ".butAction.dropdown-toggle", function(event) {
```
thersane-john, line 1324: This this ajax load button.

eldy:
I merge manually the fix for 2 of 3 points.
For the last one, the pb to use the function for external modules links, i pushed a different fixes.
Module can provide an url build with buildpath('/monmondule/mapage.php', 1) into 'urlroot' instead of 'url'
This looks better than compensating with js code to fix a bug in php.

htdocs/core/js/lib_head.js.php
thersane-john: > This looks better than compensating with js code to fix a bug in php.

@eldy this javascript is an other fix (the 4th) ... sorry I haven't explain enouth this one, it's for buttons created by js or ajax loaded

I updated the fix description

htdocs/core/lib/functions.lib.php
```diff
@@ -12291,6 +12291,39 @@ function dolGetButtonAction($label, $text = '', $actionType = 'default', $url =
...
+				}
+			}
+
+			// TODO : this is a first fix of button behavior : external module will needs to use http:// or https:// prefix on there Urls
+			if (!isset($parsedUrl['scheme'])) {
+				$url = DOL_URL_ROOT.$url;
+			}
+
+			return $url;
+		};
```
thersane-john, line 12324: @eldy I think you've misunderstood what I mean with this function.
In some case if href attribute of link start with javascript: because a link can have differrent sheme like

- http:
- https:
- javascript:
- maito:
- #

but in all cases the code force to add `&backtopage=` to url. Also  the actual dolibarr code doesn't check if query already contain  backtopage param.

I will recreate 2 PRs 
- first to fix ajax loaded BTN
- second to allow desabling `&backtopage=`

PR #31221, opened 2024-10-01
add db tables for product attribute variant

frederic34:


htdocs/install/mysql/tables/llx_product_attribute_extrafields.key.sql
```diff
@@ -0,0 +1,19 @@
+-- ===================================================================
```
eldy, line 1: I suggest to rename the table into
llx_product_attribute_extrafields so the table will be create only if we activate the module variant.

frederic34: rename the table or the file ?
llx_product_attribute_extrafields.sql to llx_product_attribute_extrafields-variants.sql

eldy: > rename the table or the file ? llx_product_attribute_extrafields.sql to llx_product_attribute_extrafields-variants.sql

You're right, i mean rename the "file" only.

frederic34: Does I need to remove migrations ?

PR #31222, opened 2024-10-01
Qual: Add codespell entries to catch some css typos

mdeweerd:
# Qual: Add codespell entries to catch some css typos

Added some codespell entries based on the observed typos in css tags

### DIFF
--- a/htdocs/accountancy/admin/journals_list.php
+++ b/htdocs/accountancy/admin/journals_list.php
@@ -438,7 +438,7 @@ if ($id) {
 		print '</tr>';
 
 		// Line to enter new values
-		print '<tr class="oddeven nodrag nodrap nohover">';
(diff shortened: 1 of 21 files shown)


htdocs/accountancy/admin/journals_list.php
```diff
@@ -438,7 +438,7 @@
 		print '</tr>';
 
 		// Line to enter new values
-		print '<tr class="oddeven nodrag nodrap nohover">';
```
frederic34, line 441: nodrag nodrop

mdeweerd: Thanks for the hint - I noticed the nodrap case when reviewing the changes (done with codespell ... -w).

I added the minwith100 and minwith200

PR #31229, opened 2024-10-02
NEW - replenishment if the warehouse is set to the default selected user

BB2A-Anthony:
# NEW|New [*replenishment if the warehouse is set to the default selected user*]
[*Long description*]


htdocs/product/stock/replenish.php
```diff
@@ -84,6 +84,10 @@
 if ($count == 1 && (empty($fk_entrepot) || $fk_entrepot <= 0) && getDolGlobalString('MULTICOMPANY_PRODUCT_SHARING_ENABLED')) {
 	$fk_entrepot = $lastWarehouseID;
 }
+//If the warehouse is set to the default selected user
+if ((empty($fk_entrepot) || $fk_entrepot <= 0) && getDolGlobalString('MAIN_DEFAULT_WAREHOUSE_USER')) {
```
eldy, line 88: Does this means the use can't set warehouse to another value?
Mays be it is better to do 
`if (!GETPOSTISSET('fk_warehouse') && (empty($fk_entrepot) || $fk_entrepot <= 0) && getDolGlobalString('MAIN_DEFAULT_WAREHOUSE_USER')) {`
So user can still change value if default is not the good one ?

BB2A-Anthony: @eldy it's done

PR #31232, opened 2024-10-02
Fix Problem with search engine and ckeditor config

### LINKED ISSUE #31231: Problem with search engine and ckeditor config
### Bug

By default, when rich text editor ckeditor is activated, it converts accentuated characters into html entities (é => &eacute;)
But doing so, the words witch accentuated characters are'nt found when searched with global search

Tested for example with "description" field of project

### Dolibarr Version

all

### Environment PHP

_No response_

### Environment Database

_No response_

_No response_

### Attached files

_No response_

vmaury:
# FIX #31231

A so-so fix that's change ckeditor conf so that it doesn't convert accentuated chars in html_entities but let them as is (in utf8)

May be it would be better change this in /htdocs/includes/ckeditor/ckeditor/build-config.js and re-build ckeditor but I dn't know how to ..

### DIFF
--- a/htdocs/theme/eldy/ckeditor/config.js
+++ b/htdocs/theme/eldy/ckeditor/config.js
@@ -32,6 +32,7 @@ CKEDITOR.editorConfig = function( config )
 	//config.autoParagraph = false;
 	//config.removeFormatTags = 'b,big,code,del,dfn,em,font,i,ins,kbd';		// See also rules on this.dataProcessor.writer.setRules
 	//config.forcePasteAsPlainText = true;
+	config.entities_latin = false;
 
 	config.toolbar_Full =
 	[


htdocs/theme/eldy/ckeditor/config.js
```diff
@@ -32,6 +32,7 @@ CKEDITOR.editorConfig = function( config )
 	//config.autoParagraph = false;
 	//config.removeFormatTags = 'b,big,code,del,dfn,em,font,i,ins,kbd';		// See also rules on this.dataProcessor.writer.setRules
 	//config.forcePasteAsPlainText = true;
+	config.entities_latin = false;
```
eldy, line 35: A better choice would be to add it into doleditor.class.php after line
/* property:xxx is same than CKEDITOR.config.property = xxx */
so we can set the value to true of flase depending if $conf->db->dolibarr_main_db_collationcontains the key 'latin'or not.
Can you try this ?

vmaury: I'm not sure it depends on the database collation
the screen charset is I think always utf-8 
`<meta charset="utf-8">`
I tested on 2 instances
the first dolibarr 16 
`$dolibarr_main_db_character_set='latin1';`
`$dolibarr_main_db_collation='latin1_swedish_ci';`

the second dolibarr 21 alpha
`$dolibarr_main_db_character_set='utf8';`
`$dolibarr_main_db_collation='utf8_unicode_ci';`

In both cases, 
`config.entities_latin = false`
Is ok.

I think in ckeditor, _config.entities_latin_ is not the right word, It should be _config.html_entities_

eldy (approved):

PR #31234, opened 2024-10-02
NEW : add fichinter category - define new table

defrance:


htdocs/install/mysql/tables/llx_categorie_fichinter.key.sql
```diff
@@ -0,0 +1,26 @@
+-- ============================================================================
```
eldy, line 1: Can you rename the file llx_categorie_fichinter-fichinter so table is created only when module fichinter is enabled.

htdocs/install/mysql/tables/llx_categorie_fichinter.sql
```diff
@@ -0,0 +1,27 @@
+-- ============================================================================
```
eldy, line 1: Can you rename the file llx_categorie_fichinter-fichinter so table is created only when module fichinter is enabled.

eldy (changes requested):

defrance:
@eldy what is the good way to inform you when i made change asked? I have the impression that the information is not reaching you and that my PRs then remain indefinitely forgotten.

eldy:
> @eldy what is the good way to inform you when i made change asked? I have the impression that the information is not reaching you and that my PRs then remain indefinitely forgotten.

No need to inform.
PR are stacked in the pull of open PR to be processed every month.

Note that due to a high number of PR of new features merged compared to a very very low number of PR contributions to stabilize application and fix bugs, PR for new features have low priority.

Also not that due to also a very high number of pr merged, there is a high probability of conflict making or not  mergeable (like this one). Any PR with conflicts not solved by the submiter are not processed and may stay in pending status for several month until fixed.

PR #31244, opened 2024-10-03
fix: php warning 8.0

FHenry:


mdeweerd:
FYI: This shows up in phan as "PossiblyUndeclaredVariable".

Search term on https://cti.dolibarr.org for this case: "possiblyundeclaredV admin.lib"

Personnally most of the time I initialise the variable to an appropriate value in an earlier location.  Here I would have done `$objMod = null;` near the start of the function where the other initializers are.

PR #31246, opened 2024-10-03
fix precommit Ci no langs for key

Hystepik:
Add trans key to a specific key used in htdocs/core/triggers/interface_50_modAgenda_ActionsAuto.class.php


frederic34:
already done here... https://github.com/Dolibarr/dolibarr/pull/31214

PR #31263, opened 2024-10-03
(Re)store mariadb setup executable from/to cache

mdeweerd:
# (Re)store mariadb setup executable from/to cache

After adding cachability for setup-mariadb on the fork of that action, use the feature


frederic34:
@mdeweerd restore cache is done after setup mariadb ?

PR #31266, opened 2024-10-04
Fix 20.0 Action buttion url autocomplete (button dropdown part04)

thersane-john:
See #31217

# FIX : Action buttion url autocomplete

In some case if href attribute of link start with javascript: because a link can have differrent sheme like

- http:
- https:
- javascript:
- maito: 
- tel:
- \#  
but in all cases the code force to add `&backtopage=` to url. Also  the actual dolibarr code doesn't check if query already contain  backtopage param.


htdocs/core/lib/functions.lib.php
```diff
@@ -12292,6 +12293,39 @@ function dolGetButtonAction($label, $text = '', $actionType = 'default', $url =
 
 	// If $url is an array, we must build a dropdown button or recursively iterate over each value
 	if (is_array($url)) {
+
+		/**
+		 * An anonymous function to complete dropdown url
+		 * @param string $url
+		 * @return string
+		 */
+		$completeUrl = function ($url, $addDolUrlRoot = true) use ($params) {
```
eldy, line 12302: what is the goal of using here an anonymous function instead of dolCompletUrlForDropdownButton() for example ?

thersane-john: try to not pollute functions.lib.php with function only used in one context..

If we want to add this function to lib, we will have to think about something more generic. Create a new functions  like :
```php

/**
* A generic function to complete url with query
* @param string $url
* @param array $query ex: ['queryKey'=>'yourData']
* @param bool $replace deault false , set to true to replace $query values if $query keys already exist
* @return string
*/
function completteUrlQuery(string $url, array $query, bool $replace = false){

}
```

but in this case too the anonymous function will be used to call `completteUrlQuery` lib because of context to prepare and pass to function.

thersane-john: Anonymous functions are very useful when you want to start factoring some of your code. They make it easier to extract and create generic functions.

eldy: Very usefull for hacker too !

eldy: Not example yet in real life, but experience shows that theory always find its real life exemple one day.

For example, we can image an extrafield where user enter a $completeUrl="exec" as a condition or computed field (so executed) and the button is called somewhere in a hook of dolGetButtonAction (so after the first definition of the anonymous function).
So when anonymous function is called, it call the exec (when such function is not allowed into extrafields or dynamic condition)
I don't think it works, this will probably need to bypass other protection but the fact that i got the idea to test this without searching means there is a risk we must avoid.

PR #31267, opened 2024-10-04
NEW : Use ContratLigne object instead of Contrat into API Contracts a…

atm-irvine:
# NEW|New Use ContratLigne object instead of Contrat into API Contracts

Introduced a new ContratLigne instance ($contractLine) in the Contracts class to manage contract lines more efficiently.

Replaced direct calls to the contract object for contract line operations (addline, updateline, deleteline) with methods from ContratLigne (insert, update, delete, fetch, active_line, close_line).

Added new delete function in ContratLigne class


atm-irvine:
@eldy Is it clear for you?

eldy:

PR #31273, opened 2024-10-04
FIX : DEVELOP pdf space lost with images at bottom of Cyan PDF

thersane-john:
# FIX

When the MAIN_GENERATE_PROPOSALS_WITH_PICTURE conf is enabled in your Dolibarr, the page break behavior is "broken" and page breaks are too large. 

I also fix the sequence of images between each line. So they don't overlap if the description is lower than the image.

Before PATCH
github screen 001

After PATCH
github screen 002


htdocs/core/modules/propale/doc/pdf_cyan.modules.php
```diff
@@ -563,43 +563,53 @@ public function write_file($object, $outputlangs, $srctemplatepath = '', $hidede
...
 
 					if ($this->getColumnStatus('position')) {
 						$this->printStdColumnContent($pdf, $curY, 'position', $i + 1);
 					}
 
+					$newPageAddedForPhoto = false;
 					if ($this->getColumnStatus('photo')) {
 						// We start with Photo of product line
-						if (isset($imglinesize['width']) && isset($imglinesize['height']) && ($curY + $imglinesize['height']) > ($this->page_hauteur - ($heightforfooter + $heightforfreetext + $heightforsignature + $heightforinfotot))) {	// If photo too high, we moved completely on new page
+						if (isset($imglinesize['width']) && isset($imglinesize['height']) && ($curY + $imglinesize['height']) > ($this->page_hauteur - ($heightforfooter + $heightforfreetext))) {	// If photo too high, we moved completely on new page
```
eldy, line 581: This seems to fix for some case but generate trouble on other.
Imagine the line that generate a break page with old code so output on second page and that, with new code, can be output on the first page. If this line is the last one of the proposal, then no break will be done.
But because ti is the last page, then the signature and the section for total is output. In such a case, this section overwrite the last line.
This is why the test to know if there is enough room was done using 
($heightforfooter + $heightforfreetext + $heightforsignature + $heightforinfotot)

Do yo understand what i mean. Did you test this particular case ?

sonikf: Hi @eldy , @thersane-john 
I did some testing with 40+ lines with or without images and both implementations(current and @thersane-john ) still break things
This line causes problems
https://github.com/Dolibarr/dolibarr/blob/7faa9bad6777310993bbcf9f29fb2e66d5530c83/htdocs/core/modules/propale/doc/pdf_cyan.modules.php#L1908
setting `'status' => false,` partially fixes things cause it solves the random page breaks but not initial problem of empty space in each page
What is the purpose of position?

sonikf: Do you mean the template(s) or generated PDF?

thersane-john: @sonikf your generated PDF with  cyan tpl

thersane-john: After a big headache trying to fit cubes into round holes. I finally know why we're having so much trouble solving these page break problems:
The code structure is coded backwards (for real, it's not an expression).

note: I distinguish between the page footer (which is displayed on all pages) and the document footer (which is only displayed on the last page, e.g. signature frame).

instead of generating the lines as if there were no document footer and determining at the end whether or not to add a page for the document footer, the code does things this way:

it adds the photo and description columns to the PDF as if the document footer were present, and if it overprints the document footer (for each cols), then it adds a page to push the document footer back. 

This makes adding pages and displaying frames complex for nothing.

thersane-john: @eldy : I think I found a way to solve all pages break issues by rebuild the line display loop, 
Take a look at my last commit.
Page break management is now easier than ever.

I think this fix need to be moved in develop but before spend energy to update orders and invoices PDF, I want your feed back if it's possible.

thersane-john: @eldy je viens de voir que tu as mergé cette PR du coup je présume que le code te vas, du coup avant que je le fasse aussi pour les commandes et factures à tu repéré un quelconque bug lié à cette refonte ?

sonikf: Ηι @thersane-john 

Sorry I couldn't send original pdf(made in backup instance with real data, cumbersome to anonymize)

I tested again with new changes you made(Great work) and things are way better than current implementation with a  gain of 1 page less in many line PDF's

I don't know if they can considered as edge cases or not but i manage to slightly break things....

See  PDF'S with  lines_mixed in name

[20_lines_with_pic.pdf](https://github.com/user-attachments/files/17408527/20_lines_with_pic.pdf)
[30_lines_with_pic.pdf](https://github.com/user-attachments/files/17408529/30_lines_with_pic.pdf)
[30_lines_without_pic.pdf](https://github.com/user-attachments/files/17408530/30_lines_without_pic.pdf)
[40_lines_without_pic.pdf](https://github.com/user-attachments/files/17408531/40_lines_without_pic.pdf)
[40_lines_with_pic.pdf](https://github.com/user-attachments/files/17408532/40_lines_with_pic.pdf)
[42_lines_mixed.pdf](https://github.com/user-attachments/files/17408533/42_lines_mixed.pdf)
[50_lines_mixed.pdf](https://github.com/user-attachments/files/17408534/50_lines_mixed.pdf)

sonikf (approved):
Works great now!

htdocs/core/modules/propale/doc/pdf_cyan.modules.php
```diff
@@ -609,6 +609,8 @@ public function write_file($object, $outputlangs, $srctemplatepath = '', $hidede
 						}
 					}
 
+					// restore Page orientation for text
+					$pdf->setPageOrientation('', 1, $heightforfooter); // The only function to edit the bottom margin of current page to set it.
```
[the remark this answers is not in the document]
thersane-john: Nope ;-) that's just the trick, taking the problem in reverse simplifies everything.

thersane-john:
@eldy can you check please ?

PR #31274, opened 2024-10-04
Flow: Setup mariadb after restoring cache

mdeweerd:
# Flow: Setup mariadb after restoring cache

To use the cache, it has to be restored first.  The original order of actions did not enable the reuse of the mariadb executable from cache

### DIFF
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -36,11 +36,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


mdeweerd:
For some reason the 'msi' installation file is not store in cache (the size of the stored cash is too small) - this needs further investigation.

mdeweerd:
Storing and usage of mariadb installer is confirmed - saved about 11s for the installation step + download from remote server.
(The setup-mariadb action needed an update use the work directory).

PR #31285, opened 2024-10-05
Fix: Number of decimals for stock values ignores configuration

priojk:
In the select field for adding a new line for a proposal/invoice/... product as a line, the stock level is always shown with exactly two decimals. However, the number of decimals should be controlled by the existing global variable MAIN_MAX_DECIMALS_STOCK, but it is not taken into account:

image

Building the stock level with the function price from the function price2num, which uses the configured number of decimals, always leads to exactly two decimals.

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -3334,7 +3334,7 @@ class Form
 
 		if (isModEnabled('stock') && isset($objp->stock) && ($objp->fk_product_type == Product::TYPE_PRODUCT || getDolGlobalString('STOCK_SUPPORTS_SERVICES'))) {
 			if ($user->hasRight('stock', 'lire')) {
-				$opt .= ' - ' . $langs->trans("Stock") . ': ' . price2num($objp->stock, 'MS');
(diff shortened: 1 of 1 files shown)


htdocs/core/class/html.form.class.php
```diff
@@ -3334,7 +3334,7 @@ protected function constructProductListOption(&$objp, &$opt, &$optJson, $price_l
 
 		if (isModEnabled('stock') && isset($objp->stock) && ($objp->fk_product_type == Product::TYPE_PRODUCT || getDolGlobalString('STOCK_SUPPORTS_SERVICES'))) {
 			if ($user->hasRight('stock', 'lire')) {
-				$opt .= ' - ' . $langs->trans("Stock") . ': ' . price(price2num($objp->stock, 'MS'));
+				$opt .= ' - ' . $langs->trans("Stock") . ': ' . price2num($objp->stock, 'MS');
```
eldy, line 3337: Can you try 
`price(price2num($objp->stock, 'MS'), 0, $langs, 0, 0)`
So the format of number will be in the user language format instead of php international format.

priojk: Your suggestion works perfectly:
- respects the setting of MAIN_MAX_DECIMALS_STOCK
- never prints more decimals than required, i.e. 0 instead of 0.00

image

PR #31292, opened 2024-10-06
NEW Insert overwrite translation

leninrivas:
# NEW|New [*Insert overwrite translation in modules*]
[*Insert overwrite translation*]


hregis:
@leninrivas maybe it's better to define an array with keys (lang, transkey, value, entity) ?

leninrivas:
> @leninrivas maybe it's better to define an array with keys (lang, transkey, value, entity) ?

Hi.
Yes, but the structure was previously defined in this way: 
`$this->overwrite_translation = array(
	'en_US:ParentCompany'=>'Parent company or reseller',
	'fr_FR:ParentCompany'=>'Maison mère ou revendeur'
)`

PR #31294, opened 2024-10-06
NEW Accountancy - Add general account on thirdparty - Code customer part

aspangaro:
Customer party with general account
image

Customer thirdparty without general account
image

Result:
image


htdocs/comm/card.php
```diff
@@ -362,6 +372,18 @@
 		}
 		print '</td></tr>';
 
+		if (!empty($conf->accounting->enabled)) {
```
eldy, line 375: Thanks to use the new syntax
 if isModEnabled('accountancy')

aspangaro: Oups ! Thanks @eldy , I pushed a patch just now

PR #31298, opened 2024-10-07
WAIT : FIX : v20 pdf space lost with images eratosthene

thersane-john:
Need PR #31273 to be merged before

# FIX  pdf space lost with images on eratosthene PDF
like #31273

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,56 @@
 English Dolibarr ChangeLog
 --------------------------------------------------------------
 
+***** ChangeLog for 20.0.1 compared to 20.0.0 *****
+
+FIX: #30960 show and search extrafields (#31026)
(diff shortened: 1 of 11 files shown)


eldy:
This seems to fix for some case but generate trouble on other.
Imagine the line that generate a break page with old code so output on second page and that, with new code, can be output on the first page. If this line is the last one of the proposal, then no break will be done.
But because ti is the last page, then the signature and the section for total is output. In such a case, this section overwrite the last line.
This is why the test to know if there is enough room was done using
($heightforfooter + $heightforfreetext + $heightforsignature + $heightforinfotot)

Do yo understand what i mean. Did you test this particular case ?

thersane-john:
I close this one i will open a new PR based on https://github.com/Dolibarr/dolibarr/pull/31273

PR #31309, opened 2024-10-07
NEW Share msg "not found" for widgets and add a button to add new record when none exists

lamrani002:
# NEW|New msg when no records found on widget and add button for create a new one
widgetNo


htdocs/core/boxes/modules_boxes.php
```diff
@@ -429,6 +429,67 @@ public function showBox($head = null, $contents = null, $nooutput = 0)
 						$out .= "</tr>\n";
 					}
 				}
+			} else {
+				if (!empty($head['text']) || !empty($head['sublink']) || !empty($head['subpicto']) || $nblines) {
+					$out .= '<tr><td colspan="2" class="center"><span class="opacitymedium">'.$langs->trans("NoRecordFound").' </span>';
+					if (!empty($this->depends)) {
+						if (count($this->depends) > 1) {
+							foreach ($this->depends as $key => $val) {
```
eldy, line 437: I don't understand why you use ->depends to define  the link.
For example the module Order depends on Thirdparty but the "+" link must be to add a new order.
I suggest to introduce a new property 
->urltoaddentry
into each widget
If ->urltoaddentry  is defined into the widget file, we can show the "+" link, if not, we do nothing.

lamrani002: yes i see this is much better

PR #31311, opened 2024-10-08
NEW Accountancy - Add the possibility to have multiple personalized reports - SQL Part

aspangaro:
Preparing the ground to manage multiple personalized reports in accountancy.


htdocs/install/mysql/data/llx_accounting_category.sql
```diff
@@ -1,5 +1,5 @@
--- Copyright (C) 2016 		Alexandre Spangaro   <aspangaro@open-dsi.fr>
--- Copyright (C) 2017 		Laurent Destailleur  <eldy@users.sourceforge.net>
+-- Copyright (C) 2016-2024	Alexandre Spangaro			<alexandre@inovea-conseil.com>
```
eldy, line 1: Can you rename the file llx_accounting_category into
llx_accounting_category-accounting.sql
So table will be created when module accounitngis enabled.
We must start to have table name suffixed with module name after the "-".

htdocs/install/mysql/tables/llx_accounting_category_link.key.sql
```diff
@@ -0,0 +1,26 @@
+-- ===================================================================
```
eldy, line 1: For table that are link tables, we used to use the name of both joined tables into the name of table link.
So can you name it 
llx_accounting_category_account-accounting.sql
so
llx_modulename_table1_table2-modulename.sql

htdocs/install/mysql/data/llx_accounting_category.sql
aspangaro: Done. Thanks @eldy

htdocs/install/mysql/tables/llx_accounting_category_link.key.sql
aspangaro: Done. Thanks @eldy

PR #31315, opened 2024-10-08
NEW - Add a new hook on order list

atm-lucasmantegari:
# NEW - *Add a new hook on order list*
Currently, it is impossible to modify the content displayed when hovering over the shippable picto on the order list.

### DIFF
--- a/htdocs/commande/list.php
+++ b/htdocs/commande/list.php
@@ -2566,6 +2566,21 @@ if ($resql) {
 							}
 						}
 					}
+
+					// Call Hook modifyTextInfo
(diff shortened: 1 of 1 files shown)


eldy:
Because this introduce a new hook, you must also implement it into the template of modulebuilder and push it into develop

atm-florianm:
> implement it into the template of modulebuilder and push it into develop

@eldy What do you have in mind? To me, it looks like Lucas' hook is very specific to the order list (in relation with stock management), so I don't know how it could be added to `myobject_list.php`, which doesn't have a `$text_info` variable.

PR #31335, opened 2024-10-09
FIX 21: HTML Markup no longer allowed in $title parameter of print_barre_liste

atm-florianm:
# FIX
In Dolibarr 21, when you go to a company's Related Items tab, you get this:
image

After my PR it would look like this:
image


eldy:
I reverted the dolPrintLabel on the title of print_barre_liste() because we may have several cases that want to use riche html content into the title line.

PR #31336, opened 2024-10-09
Fix css

frederic34:
BEFORE
image

AFTER
image

since this commit https://github.com/Dolibarr/dolibarr/commit/e1f000fbf3c23671dadb175d6258e6953ee88ced


frederic34:
in theme md it is right
image

PR #31341, opened 2024-10-10
NEW: setDefaultLangs('auto') now check mysoc and conf

atm-thomasb:
# NEW|New setDefaultLangs now checks MAIN_LANG_DEFAULT and mysoc lang if 'auto'

Currently, setDefaultLangs with 'auto' argument sets the lang to the server's. 
I've added so that it will beforehand check if $mysoc->defaultlang is defined and if yes it will take its lang,
otherwise if 'MAIN_LANG_DEFAULT' is defined it takes its lang.

Otherwise it does the same thing as before (server's lang).

### DIFF
--- a/htdocs/core/class/translate.class.php
+++ b/htdocs/core/class/translate.class.php
@@ -93,7 +93,7 @@ class Translate
 		if ($srclang == 'auto') {
 			if (!empty($mysoc->defaultlang)) {
 				$srclang = $mysoc->defaultlang;
-			} else if (!empty(getDolGlobalString('MAIN_LANG_DEFAULT', 'auto'))) {
+			} else {
 				$srclang = getDolGlobalString('MAIN_LANG_DEFAULT', 'auto');
 			}
 		}
--- a/htdocs/core/class/translate.class.php
+++ b/htdocs/core/class/translate.class.php
@@ -70,7 +70,7 @@ class Translate
 	 */
 	public function setDefaultLang($srclang = 'en_US')
 	{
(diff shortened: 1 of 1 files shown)


atm-thomasb:
Travis currently returns an error because it expects an output that changed with my PR:

$_SERVER['HTTP_ACCEPT_LANGUAGE'] = "' malicious text with quote";
		$tmplangs->setDefaultLang('auto');
		print __METHOD__.' $tmplangs->defaultlang='.$tmplangs->defaultlang."\n";
		$this->assertEquals($tmplangs->defaultlang, 'malicioustextwithquote_MALICIOUSTEXTWITHQUOTE');

With the new system when given 'auto' it returns "en_US" instead. So it's as it's supposed to be, but is different from before.

htdocs/core/class/translate.class.php
```diff
@@ -90,6 +90,14 @@ public function setDefaultLang($srclang = 'en_US')
 
 		$this->origlang = $srclang;
 
+		if ($srclang == 'auto') {
+			if (!empty($mysoc->defaultlang)) {
+				$srclang = $mysoc->defaultlang;
```
eldy, line 95: This will create a lot of confusion for user.
a german user with setup to "auto" will get the spanish language if company is spanish when he want german. Not what we want.
Also any change in behaviour must be pushed into develpo branch.

atm-thomasb: Understandable, thank you for your answer.

PR #31346, opened 2024-10-10
NEW: tables for a new module, supplier contract (same a contrat but for supplier)

kkhelifa-opendsi:
Tables for a new module, supplier contract (same a contrat but for supplier)

I renamed the fields in french to english and remove obsolete fields
I plan to used template for the product line for customer contract et supplier contract same as template for proposal, orders, invoices but specific for contract.

Is it ok ?

### DIFF
--- a/htdocs/install/mysql/tables/llx_supplier_contract.key.sql
+++ b/htdocs/install/mysql/tables/llx_supplier_contract.key.sql
@@ -1,0 +1,31 @@
+-- ============================================================================
+-- Copyright (C) 2024      Support              <support@easya.solutions>
+--
+-- This program is free software; you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as published by
+-- the Free Software Foundation; either version 3 of the License, or
+-- (at your option) any later version.
+--
+-- This program is distributed in the hope that it will be useful,
+-- but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- GNU General Public License for more details.
+--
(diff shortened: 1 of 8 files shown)


eldy:
Currently, the module contract was designed to work on both side for customer and supplier.
Can we  support this with just a flag 'Customer / Supplier' on existing contract ?

kkhelifa-opendsi:
ok

kkhelifa-opendsi:
In Dolibarr, modules are now split between customer and supplier.
This makes it possible to have different sets of document templates, separate additional fields, separate reference, add lines specific to purchases and sales, etc...
Now this can be done in a single module, but not in the same way as other modules where the customer and supplier parts have been separated.

philippe-opendsi:
we need to start developing. @eldy Are you sure we're starting with a single module ?

eldy:
> we need to start developing. @eldy Are you sure we're starting with a single module ?

No, not sure. There is pro and cons.

So i I procrastinated by notice you the fact that we can already do it in a simple manner. I would just like to be sure that before starting a full module you were aware that we can already manage supplier contract in a simple way. And if this is enough for short time, may be priority for this enhancement is no more a priority and we can merge other PR before...

pscoffoni-opendsi:
@eldy Concerning our customer's request, the module as it stands doesn't meet his requirements. He already uses the module for the customer part, and would like to be able to clearly differentiate between the two types of contract with additional fields specific to customers and suppliers. The same applies to document templates.
Last but not least, the lines need to retrieve purchase prices for products/services, rather than sales prices as at present.
That's why we're moving towards a complete, separate module.
If you think this is premature, we can do it as an external module and push it later into the core.

pscoffoni-opendsi:
This will be an external module.

PR #31347, opened 2024-10-10
NEW : Addition of contact substitution keys for sending emails from massactions

samifilali:


htdocs/core/actions_massactions.inc.php
```diff
@@ -488,6 +488,20 @@
 
 					complete_substitutions_array($substitutionarray, $langs, $objecttmp, $parameters);
 
+					$contactarr = $objecttmp->liste_contact(-1, 'external', 0, '', 1);
```
eldy, line 491: $objecttmp may be empty.
Also method liste_contact may not exists.

atm-irvine: I have taken your suggestion into account.
If there is 2 different for the same type/code, then we use the first one.

PR #31349, opened 2024-10-10
Fix: code_compta aliased as code_compta_client

mdeweerd:
# Fix: code_compta aliased as code_compta_client

Fix warning regarding non-existing field on sql result

image


mdeweerd:
I also added a mapping for 'ode_compta' to 'code_compta_client' for the deprecationhandler implying that the code_compta property should no longer be declared.

PR #31361, opened 2024-10-11
fix: add rounding for multicurrency unit price

kreativekrise:
In **Dolibarr**, when using **foreign currencies**, the wrong value was used for the unit price. Instead of using the correctly rounded value for the unit price, `$pu_devise` (the foreign currency value) was directly used as `$pu`. This resulted in the **unit price being stored in the database with more than two decimal places** and displayed incorrectly in documents like PDFs, leading to inconsistent price representations.

### Cause

The issue was caused by using **`$pu_devise`** (unit price in foreign currency) as the value for **`$pu`** (unit price), without applying the necessary rounding to two decimal places. As a result, the unit price for foreign currencies was not correctly rounded or stored.

### Solution

I modified the code so that:
- **`$pu`** is now passed as **`0`** when calling **`calcul_price_total()`** for foreign currencies.
- **`$pu_devise`** is now passed instead of **`0`** when calling **`calcul_price_total()`** for foreign currencies.
- This ensures that the correct unit price calculation logic is triggered.

### Impact

This fix ensures that:
- Unit prices for foreign currencies, are **stored in the database with two decimal places**, bringing consistency across the system.

### DIFF
--- a/htdocs/core/lib/price.lib.php
+++ b/htdocs/core/lib/price.lib.php
@@ -174,7 +174,7 @@ function calcul_price_total($qty, $pu, $remise_percent_ligne, $txtva, $uselocalt
 	// pu calculation from pu_devise if pu empty
 	if (empty($pu) && !empty($pu_devise)) {
 		if (!empty($multicurrency_tx)) {
-			$pu = price2num($pu_devise / $multicurrency_tx, 'MT');
+			$pu = price2num($pu_devise / $multicurrency_tx, 'MU');
 		} else {
 			dol_syslog('Price.lib::calcul_price_total function called with bad parameters combination (multicurrency_tx empty when pu_devise not) ', LOG_ERR);
 			return array();
--- a/htdocs/core/lib/price.lib.php
+++ b/htdocs/core/lib/price.lib.php
@@ -174,7 +174,7 @@ function calcul_price_total($qty, $pu, $remise_percent_ligne, $txtva, $uselocalt
 	// pu calculation from pu_devise if pu empty
 	if (empty($pu) && !empty($pu_devise)) {
 		if (!empty($multicurrency_tx)) {
-			$pu = $pu_devise / $multicurrency_tx;
+			$pu = price2num($pu_devise / $multicurrency_tx, 'MT');
 		} else {
 			dol_syslog('Price.lib::calcul_price_total function called with bad parameters combination (multicurrency_tx empty when pu_devise not) ', LOG_ERR);
 			return array();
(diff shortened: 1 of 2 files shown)


htdocs/core/lib/price.lib.php
```diff
@@ -174,7 +174,7 @@ function calcul_price_total($qty, $pu, $remise_percent_ligne, $txtva, $uselocalt
 	// pu calculation from pu_devise if pu empty
 	if (empty($pu) && !empty($pu_devise)) {
 		if (!empty($multicurrency_tx)) {
-			$pu = $pu_devise / $multicurrency_tx;
+			$pu = price2num($pu_devise / $multicurrency_tx, 'MT');
```
eldy, line 177: $pu is a unit price, also it is a foreign currency, so rounding must use 'UC' instead of 'MT'

test/phpunit/PricesTest.php
```diff
@@ -89,15 +89,15 @@ public function testCalculPriceTotal()
 		$result1 = calcul_price_total(2, 8.56, 0, 10, 0, 0, 0, 'HT', 0, 0, '', '', 100, 1.09205);
 		print __METHOD__." result1=".join(', ', $result1)."\n";
 		// result[0,1,2,3,4,5,6,7,8]	(total_ht, total_vat, total_ttc, pu_ht, pu_tva, pu_ttc, total_ht_without_discount, total_vat_without_discount, total_ttc_without_discount)
-		$this->assertEquals(array(17.12, 1.71, 18.83, 8.56, 0.856, 9.416, 17.12, 1.71, 18.83, 0, 0, 0, 0, 0, 0, 0, 18.7, 1.87, 20.57, 9.34795, 0.93479, 10.28274, 18.7, 1.87, 20.57, 0, 0), $result1, 'Test1b FR');
```
eldy, line 92: Why changing the phpunit test ?
Goal of unit test is to validate that the code after give same result than before with same dataset.

htdocs/core/lib/price.lib.php
KreativeKrise: You are right, 'MT' is wrong. But you probably mean 'CU' instead of 'UC', because my system does not recognize 'UC'. 

I have tested it with 'CU'.
The problem with that is that then the value for “Total (excl. in currency)” is wrong by 0.01.



Then I tested it with 'MU'.
Here the value for “Total (excl. in currency)” is calculated correctly. However, the value for “Remaining unpaid, original currency” is not correct because it is calculated separately from the original currency and the exchange rate.



I don't know if and how this can be solved in a good way.

test/phpunit/PricesTest.php
KreativeKrise: I changed the phpunit test because the rounding of `$pu` results in different values.

kreativekrise:
Since most total values are calculated on the fly using the original currency and do not take the unit price into account, there will always be differences due to different rounding.

To solve this it probably requires a much more extensive change, which is probably not worthwhile. But I find it strange to have a unit price with more than 2 decimal places in an invoice.

test/phpunit/PricesTest.php
eldy: If it result in different values, it surely means the PR introduce bugs.

eldy:
> Since most total values are calculated on the fly using the original currency and do not take the unit price into account, there will always be differences due to different rounding.
> 
> To solve this it probably requires a much more extensive change, which is probably not worthwhile. But I find it strange to have a unit price with more than 2 decimal places in an invoice.

A unit price is necessarly using more decimals than the totalprice, otherwise when qty is high (1000), you will get a high difference when calculating total. In most countries unit price is 5 or 6 decimals when decimals for total is 2 (like for dollar or euro). For example in Europe, when you buy petrol,the unitprice has often 3 decimal when total has 2 decimal.

htdocs/core/lib/price.lib.php
eldy: I doubt this can always be fixed. Main currency and foreign currency are using different accuracy, rounding rules, so it is not possible to have always no rounding error in both main currency and foreign currency.

eldy:
I am still not able to reproduce the problem.
Can you provide the data you use as input for a test that prove something is wrong, the data you got in database and the data you expect to have ?

PR #31363, opened 2024-10-11
New datastructure for webhok_target for new status

Hystepik:
Add a new column to wabhook_target database to save a Json array for manual trigger status target

### Workflow
When the code is trigger save a new object in $trigger_stack json array with trigger_code, element_type, element_id


mdeweerd:
The sqlfluff tool is correct, a comma is missing between these fields:

```
     trigger_codes text NULL                 -- list of selected trigger that must call the webhook
    trigger_stack text,  
```

PR #31368, opened 2024-10-12
Qual: fix phan notices

mdeweerd:
# Qual: fix phan notices

Fix phan/phpstan notices, mostly in html,formmail


mdeweerd:
@eldy 
I leave it to you to fix this phpstan notice.

https://github.com/Dolibarr/dolibarr/blob/235d23a2b2161679c9791deaeca4aa7bdc8a414b/htdocs/core/class/html.formmail.class.php#L957-L959

```plaintext
Parameter #2 $substitutionarray of function make_substitutions expects array<string, string>, string given.
```

EDIT: I added an exception to phan so that this notice is ignored.

htdocs/comm/mailing/card.php
```diff
@@ -1241,7 +1241,7 @@
 				$formmail->withfrom = 0;
 				$formmail->withto = $user->email ? $user->email : 1;
 				$formmail->withtocc = 0;
-				$formmail->withtoccc = getDolGlobalString('MAIN_EMAIL_USECCC');
+				$formmail->withtoccc = getDolGlobalInt('MAIN_EMAIL_USECCC');
```
eldy, line 1244: Option MAIN_EMAIL_USECCC so getString looks better.

PR #31371, opened 2024-10-12
Bug fix #31369 date_solde not taken into account when sent through API

### LINKED ISSUE #31369: Bug: date_solde not taken when creating a bank account via API
### Bug

Version: 20.0.1

Hello,

I'm able to create a bank account via API but although initialising a solde value works, initialising a date_solde doesn't.
This results in having the solde initial with the date of today, always.
image

I looked at the code in
`/htdocs/compta/bank/class/api_bankaccounts.class.php`

And at line 172 I have
`$account->date_solde = time();`

I checked that hardcoding setting $account->date_solde to some value have the desired effect.

So the problem is that instead of retrieving date_solde from the API call, it always set $account->date_solde to the date of right now

I replaced
`$account->date_solde = time();`
by
```
if (isset($request_data['date_solde'])) { $account->date_solde = $this->_checkValForAPI('date_solde', $request_data['date_solde'], $account); }
else { $account->date_solde = time(); }
```
Then it works properly

### Dolibarr Version

20.0.1

### Environment PHP

_No response_

### Environment Database

_No response_

_No response_

### Attached files

_No response_

mikygee:
# Fix #[31369 date_solde not taken into account when sent through API]
date_solde is not taken into account when sent through API. This fix resolves this issue.

### DIFF
--- a/dev/translation/ignore_translation_keys.lst
+++ b/dev/translation/ignore_translation_keys.lst
@@ -1004,7 +1004,6 @@ VideoGeneration
 #####################################################################
 #####################################################################
 AddService
-AnHelpMessage
 Analytic
 AnalyticLabel
 BankId
@@ -1036,6 +1035,7 @@ FailedToLogin
 First
 GroupCreate
 GroupUpdate
+HelpMessage
 INVOICE_RETAINED_WARRANTY_LIMITED_TO_FINAL_SITUATION
 INVOICE_RETAINED_WARRANTY_LIMITED_TO_SITUATION
 INVOICE_SITUATION_DEFAULT_RETAINED_WARRANTY_COND_ID
--- a/htdocs/core/class/html.formsetup.class.php
+++ b/htdocs/core/class/html.formsetup.class.php
@@ -945,7 +945,7 @@ class FormSetupItem
 	}
 
 	/**
-	 * generatec default input field
+	 * Generate default input field
 	 *
 	 * @return string
 	 */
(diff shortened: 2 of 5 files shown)


htdocs/compta/bank/class/api_bankaccounts.class.php
```diff
@@ -170,6 +170,8 @@ public function post($request_data = null)
 		}
 		// Date of the initial balance (required to create an account).
 		$account->date_solde = time();
+		// Use the value of date_solde sent through API otherwise let date_solde value to now which was initialised just before
```
eldy, line 173: Can you test this other fix:  Just move the line
$account->date_solde = time();
at line 162 before the foreach ?

mikygee: Hello Laurent,

Indeed, that worked as you said

```
                // Date of the initial balance (required to create an account).
                $account->date_solde = time();
                foreach ($request_data as $field => $value) {
                        if ($field === 'caller') {
                                // Add a mention of caller so on trigger called after action, we can filter to avoid a loop if we try to sync back again with the caller
                                $account->context['caller'] = sanitizeVal($request_data['caller'], 'aZ09');
                                continue;
                        }

                        $account->$field = $this->_checkValForAPI($field, $value, $account);
                }
```

eldy: Yes, you must get your branch, edit code to follow the change suggested (so just moving the line with no need to add the 
if (isset($request_data['date_solde'])) { ...
And push. This should update the PR automatically if you take the same branch than the one you used for your first push...

mikygee: Hello Laurent,
Thank you for your message.
I think I managed to do the change.

eldy: I don't think so. PR still has old code. Did you repush after the change ?
You can also cancel pr and restart it from scratch as it is just 1 line move ...

PR #31376, opened 2024-10-13
NEW Invoice - Generate payment information - Structured communication

aspangaro:
Add payment information "structured communication" on invoice. 
See this discussion on french forum : https://www.dolibarr.fr/forum/t/communication-structuree-europeenne/24399/15

Use in other country than France / Germany / Netherlands
In line with ISO 20022 (SEPA) standard

image

image

image

_Next step: use this structured number in prelevement module_


aspangaro:
I don't understand the CI error...

frederic34:
@aspangaro create table is invalid see https://github.com/Dolibarr/dolibarr/pull/31364

htdocs/admin/invoice.php
```diff
@@ -758,6 +759,17 @@
 print ajax_combobox("chq", array(), 0, 0, 'resolve', '-2');
 
 print "</td></tr>";
+
+// Structured communication
+// France / Germany / Netherlands do not use this structured system of invoice's payment
+if ($mysoc->country_code != 'FR' && $mysoc->country_code != 'DE' && $mysoc->country_code != 'NL') {
```
sonikf, line 765: Hi @aspangaro could you change this line to 

`if ($mysoc->country_code = 'BE') {`

as it only applies to Belgium for the moment

aspangaro: I've read that this doesn't apply in France, Germany and the Netherlands (cf link to ISO 20022 standard from the forum), so the activation option will be available for Belgium and other countries, but not for the three mentioned above.

sonikf: The use of  ISO 20022 standard will be mandatory for all SEPA countries (including France, Germany and the Netherlands)  by November 2025 and changes will have to be made to Dolibarr as well, but it's not the issue here.

https://www.securities-services.societegenerale.com/en/insights/views/news/adoption-of-the-swift-iso-20022-standard-for-corporate-actions/

https://www.ubaf.fr/en/news/iso-20022-adoption-and-migration-cross-border-payments-and-reporting-plus

As stated above the structured communication payment reference only applies to Belgium for the moment and is sort of the equivalent to Finland's RF payment reference which is already implemented

https://github.com/Dolibarr/dolibarr/blob/26bb977006ce803363dcbc8c529bce61461f49f9/htdocs/core/lib/functions_fi.lib.php#L37

aspangaro: Thanks for all the information @sonikf I move the function in specific file for Belgium.

Also, in admin, the option is only seen by our Belgian friends

PR #31377, opened 2024-10-13
fix translation

frederic34:


frederic34:
@mdeweerd windows ci fails in branch 20.0 since version 20.0.1 (version programs != version db)

mdeweerd:
> @mdeweerd windows ci fails in branch 20.0 since version 20.0.1 (version programs != version db)

ll31344 was pending integration .

PR #31382, opened 2024-10-13
Qual: Suppress PhanTypeSuspiciousNonTraversableForeach

mdeweerd:
# Qual: Ignore PhanTypeSuspiciousNonTraversableForeach

PhanTypeSuspiciousNonTraversableForeach reports on the use of
`foreach ($object as $key => value)` which is valid accross
tested php versions (at least since PHP7.0).

So suppressing this notice.

### DIFF
--- a/dev/tools/phan/baseline.txt
+++ b/dev/tools/phan/baseline.txt
@@ -26,7 +26,6 @@ return [
     // PhanRedefineFunction : 50+ occurrences
     // PhanPluginEmptyStatementIf : 45+ occurrences
     // PhanTypeMismatchDimFetch : 40+ occurrences
-    // PhanTypeSuspiciousNonTraversableForeach : 30+ occurrences
     // PhanTypeExpectedObjectPropAccess : 25+ occurrences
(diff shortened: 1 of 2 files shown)


htdocs/blockedlog/class/blockedlog.class.php
```diff
@@ -478,26 +478,30 @@ public function setObjectData(&$object, $action, $amounts, $fuser = null)
...
+					if (!is_object($value) && !is_null($value) && $value !== '') {
+						$this->object_data->thirdparty->$key = $value;
+					}
 				}
 			}
 		}
 
 		// Add company info
-		if (!empty($mysoc)) {
+		if (!empty($mysoc) && (is_array($mysoc) || ($mysoc instanceof Traversable))) {
```
eldy, line 504: It seems that a PHP object will return false with 
$object instanceof Traversable.
And we want to do the foreach($object as $key => $value) to loop on each of its property.
So the test will break this.

mdeweerd: Ok, I assumed that this was not possible but it is.

So I suppress the notice.

PR #31386, opened 2024-10-14
Fix: CommonClassTest: fix type on onNotSuccessfulTest

alexandre-janniaux:
# Fix CommonClassTest: fix type on onNotSuccessfulTest

TestCase::onNotSuccessfulTest returns never and not void, which also enforce calling a function returning never (ie. in this case or most other cases, parent::onNotSuccessfulTest) or throwing.


alexandre-janniaux:
It seems that I'm still a bit in advance on phpunit again... :(

test/phpunit/CommonClassTest.class.php
```diff
@@ -137,9 +137,9 @@ public static function setUpBeforeClass(): void
 	 *	This method is called when a test fails
 	 *
 	 *  @param	Throwable	$t		Throwable object
-	 *  @return void
+	 *  @return never
```
eldy: This seems available with php 8.1 only.
Currently our CI and php range is Min 7.2  -  Max 8.5

alexandre-janniaux: Yes unfortunately :/ I'm not sure how to handle this properly across version, but I also haven't touched this part for two years.

PR #31390, opened 2024-10-14
FIX: accountancysystem: fix object create() method

alexandre-janniaux:
# FIX: accountancysystem: fix object create() method

The accountancysystem::create() method is not used in the code, and thus was not updated, especially after the introduction of the whole new accountancy system. The method is at least useful for the tests so fix it to create values like in every other tests.


alexandre-janniaux:
@eldy I didn't though it would be merged that fast with failing CI, but it was actually depending on https://github.com/Dolibarr/dolibarr/pull/31389. No worry since it was broken before anyway.

I have a test suite coming right after those patches (my end goal is BookKeepingTest).

eldy:
> @eldy I didn't though it would be merged that fast with failing CI, but it was actually depending on https://github.com/Dolibarr/dolibarr/pull/31389. No worry since it was broken before anyway.
> 
> I have a test suite coming right after those patches (my end goal is BookKeepingTest).

Yes when ci is wrong i analyze the cause and ignore it when i think risk is liw and if trouble comes from somewhere else (or i fix ci myself to rerun the ci)

PR #31391, opened 2024-10-14
QUAL: AccountancySystem: maintenance work to uniformize and clean up

alexandre-janniaux:
# QUAL: AccountancySystem: maintenance work to uniformize and clean up

This MR provides a few improvement to the AccountancySystem class, which was pretty much left unmaintained for a while in the new accountancy system. A few fields have never been used, and error reporting to the caller was not setup before this MR.


alexandre-janniaux:
Depends on https://github.com/Dolibarr/dolibarr/pull/31390 since numero was incorrectly used before.

alexandre-janniaux:
https://github.com/Dolibarr/dolibarr/pull/31390 was merged :)

PR #31387, opened 2024-10-14
Qual: societe: fix warning during tests

alexandre-janniaux:
# Qual societe: fix warning during tests

    1) /home/alexandre/workspace/dolibarr/htdocs/societe/class/societe.class.php:1299
    Attempt to read property "country_id" on null

    Triggered by:

    * BookKeepingTest::testBookKeepingCreate (6 times) /home/alexandre/workspace/dolibarr/test/phpunit/BookKeepingTest.php:53

    * SocieteTest::testSocieteCreate (6 times) /home/alexandre/workspace/dolibarr/test/phpunit/SocieteTest.php:87

    * SocieteTest::testSocieteFetch (6 times) /home/alexandre/workspace/dolibarr/test/phpunit/SocieteTest.php:114

    * SocieteTest::testSocieteMerge (12 times) /home/alexandre/workspace/dolibarr/test/phpunit/SocieteTest.php:450

    * SocieteTest::testSocieteUpdate (6 times) /home/alexandre/workspace/dolibarr/test/phpunit/SocieteTest.php:143

PR #31393, opened 2024-10-14
Qual: Nullable mysoc hint

mdeweerd:
# Qual: Nullable mysoc hint

$mysoc can be nullable resulting in issues (#31387) so it is best to detect these cases with static analysis


alexandre-janniaux:
In particular, it can happen during tests scenarios, and maybe during incomplete installation. :)

mdeweerd:
Unfortunately the update did not reveal the issue fixed in #31387 but it did mostly in the document generators where I have added a null test and a log message.

PR #31313, opened 2024-10-08
NEW: Management of extrafields on customer prices, level prices and default prices

altairis-noe:
# NEW: Management of extrafields on customer prices, level prices and default prices

## Add configuration of extrafields on customer prices and price level for products

image

## Add extrafields on customer prices in the product Sales price tab and for the default price of the product

image

## Add of extrafields on customer prices on the thirdparty Customer prices tab

image

Extrafields are managed while creating and updating customer prices, price level and default prices.

PR #31395, opened 2024-10-14
NEW: tables for product customer prices extrafields management

altairis-noe:
# NEW: tables for product customer prices extrafields management
As asked in PR #31313 new tables should be in dedicated PR!


eldy:
Can you include in this same PR  the creation of this tables and index into the migration file 20.0.0-21.0.0.sql too ?

altairis-noe:
You asked nicely, so I did!

PR #31397, opened 2024-10-14
FIX Warning message

zahwyrix:
# FIX 
To avoid the warning “Error Failed to load doc generator with modelpaths=core/modules/mrp/doc/ - modele=alpha”.

### DIFF
--- a/htdocs/mrp/class/mo.class.php
+++ b/htdocs/mrp/class/mo.class.php
@@ -485,7 +485,9 @@ class Mo extends CommonObject
 		}
 
 		$this->socid = $this->fk_soc;
-
+		if(empty($this->model_pdf)){
+			$this->model_pdf = 'vinci';
+		}
 		return $result;
 	}
 
--- a/htdocs/mrp/class/mo.class.php
+++ b/htdocs/mrp/class/mo.class.php
@@ -485,9 +485,7 @@ class Mo extends CommonObject
 		}
 
 		$this->socid = $this->fk_soc;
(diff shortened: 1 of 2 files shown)


htdocs/mrp/class/mo.class.php
```diff
@@ -485,7 +485,9 @@ public function fetch($id, $ref = null)
 		}
 
 		$this->socid = $this->fk_soc;
-
+		if (empty($this->model_pdf)) {
```
eldy, line 488: If there is no default value for PDF to generate, behavior is to have no document generation done.
If we change this, i think we will always have a doc generated in Vinci model instead of no doc.
Can't we fix the warning by disabling generation try if template not defined ?

zephyriony: I understand what you're saying, but the default behavior is that there's no template defined in the object when it's validated, which is a bit strange since for all documents there is a template and when you validate the pdf is automatically generated.

eldy: Not for all. For example on contract we can decide to not have document generated, same for supplier invoice.
Should be sale for mrp: if no default doc set, we do not generate doc.

zephyriony: In fact, the problem isn't with the code, which is fine, but with the constant in the MRP_MO_ADDON_PDF database, which is set to alpha (which doesn't exist) instead of vinci. I've reactivated the definition of vinci in the constructor for this constant because I can't find where the alpha comes from.

PR #31404, opened 2024-10-15
New  tab Note for facture_rec

lamrani002:
# NEW|New Tab note for customer invoice model
fect


htdocs/compta/facture/note-rec.php
```diff
@@ -0,0 +1,170 @@
...
+
+$objecttype = 'facture_rec';
+if ($action == "create" || $action == "add") {
+	$objecttype = '';
+}
+
+$object = new FactureRec($db);
+// Load object
+if ($id > 0 || !empty($ref)) {
+	$object->fetch($id, $ref, '', '', (getDolGlobalString('INVOICE_USE_SITUATION') ? $conf->global->INVOICE_USE_SITUATION : 0));
```
frederic34, line 55: @lamrani002 I think you can keep only `$object->fetch($id, $ref);` as there are no sition in invoice models

lamrani002: yes i agree thanks

eldy:
On previous version, after the label "Note private" or "Note public", we got a picto "i" that show some information about substitution keys we can use into text.
You must report also this on the new page.

PR #31405, opened 2024-10-15
Add tab Note for fournisseur facture_rec

lamrani002:
# NEW|New tab note for fournisseur invoice _rec
fectFou


eldy:
On previous version, after the label "Note private" or "Note public", we got a picto "i" that show some information about substitution keys we can use into text.
You must report also this on the new page.

PR #31409, opened 2024-10-15
NEW include custom tpl path for note

Jonathanbtq:
I included the path to use custom templates for notes.tpl.php.

* Update commande/notes.php

### DIFF
--- a/htdocs/core/class/commondocgenerator.class.php
+++ b/htdocs/core/class/commondocgenerator.class.php
@@ -912,6 +912,9 @@ abstract class CommonDocGenerator
 			array('line_date_start', 'date_start', 'day', 'auto', null),
 			array('line_date_start_locale', 'date_start', 'day', 'tzserver', $outputlangs),
 			array('line_date_start_rfc', 'date_start', 'dayrfc', 'auto', null),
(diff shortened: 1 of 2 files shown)


htdocs/commande/note.php
```diff
@@ -137,7 +137,13 @@
 
 
 	$cssclass = "titlefield";
-	include DOL_DOCUMENT_ROOT.'/core/tpl/notes.tpl.php';
+	$dirtpls = array_merge($conf->modules_parts['tpl'], array('/core/tpl'));
```
eldy, line 140: If $conf->modules_parts['tpl'] is empty,what if value for $dirtpls ?
May be this can avoidtrouble ?
$dirtpls = array('/core/tpl');
if ($conf->modules_parts['tpl']) {
     $dirtpls = array_merge($conf->modules_parts['tpl'], $dirtpls);
}

Jonathanbtq: Yes, this makes sense. Initializing $dirtpls with /core/tpl as a default ensures there's always a valid path, even if $conf->modules_parts['tpl'] is empty.
I make a update, thanks

PR #31410, opened 2024-10-15
use constant

frederic34:


htdocs/expensereport/card.php
```diff
@@ -1847,7 +1847,7 @@
 				print '</tr>';
 			}
 
-			if ($object->status == 99 || !empty($object->detail_refuse)) {
+			if ($object->status == ExpenseReport::STATUS_CANCELED || !empty($object->detail_refuse)) {
```
eldy, line 1850: STATUS_CANCELED is 4.
Should be STATUS_REFUSED ?

frederic34:
@eldy any update ?

PR #31414, opened 2024-10-15
Qual: Fix phan notices in admin.lib and tax.lib + related

mdeweerd:
# Qual: Fix phan notices in admin.lib and tax.lib + related

PHPdoc typing, and casts to fix typing.


mdeweerd:
The fixes/improvements to the typing created new phan notices about invalid property access to an array which were confirmed using `compta/localtax/clients.php` with the test database:
{4BD6B0F8-A02E-4D43-9763-2BF2A5DF3281}

mdeweerd:
Similar on `compta/tva/quadri_detail.php`.

{FED04E5B-00C5-403D-BF75-9EC985AB59D4}

mdeweerd:
It looks like the tva_intra field used in files such as htdocs\compta\tva\quadri_detail.php was renamed to company_tva_intra, and also that pstatus and pstatusbuy are no longer made available in the tax_by_* results, but are still supposed to exist.

This PR does not fix that - the goal is to bring the nubmer of notifications down with 'minimal' code impact.

PR #31422, opened 2024-10-16
FIX - Correction of new hook PR

atm-lucasmantegari:
# FIX - *Correction of new hook PR*
This PR corrects the [PR for adding a new hook](https://github.com/Dolibarr/dolibarr/pull/31315). I made the mistake of setting the text_info replacement to $reshook == 0, which is the default value for reshook, which always replaced the given with empty when the hook was not in use.

### DIFF
--- a/htdocs/commande/list.php
+++ b/htdocs/commande/list.php
@@ -2573,7 +2573,7 @@ if ($resql) {
 					if ($reshook == 1) {
 						// for add information
 						$text_info .= $hookmanager->resPrint;
-					} elseif ($reshook == 0) {
+					} elseif ($reshook == 2) {
 						// for replace information
 						$text_info = $hookmanager->resPrint;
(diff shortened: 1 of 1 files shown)


htdocs/commande/list.php
```diff
@@ -2573,7 +2573,7 @@
 					if ($reshook == 1) {
 						// for add information
 						$text_info .= $hookmanager->resPrint;
-					} elseif ($reshook == 0) {
```
eldy, line 2576: i made an error by validating a new hook into an already released branch. i revert it.
You can submit the hook into develop branch.
Also the fix is not correct. We should have
```

if ($reshook == 1) {
						// for add information
						$text_info = $hookmanager->resPrint;
					} elseif ($reshook == 0) {
						$text_info .= $hookmanager->resPrint;

```

eldy:

PR #31423, opened 2024-10-16
NEW Authorize modification of a partially shipped order

zahwyrix:
# NEW
Authorize modification of a partially shipped order

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -5221,6 +5221,10 @@ abstract class CommonObject
 		}
 		$extrafields->fetch_name_optionals_label($this->table_element_line);
 
+		if (method_exists($this, 'loadExpeditions')) {
+			$this->loadExpeditions();
+		}
+		
 		$parameters = array('num' => $num, 'dateSelector' => $dateSelector, 'seller' => $seller, 'buyer' => $buyer, 'selected' => $selected, 'table_element_line' => $this->table_element_line);
 		$reshook = $hookmanager->executeHooks('printObjectLineTitle', $parameters, $this, $action); // Note that $action and $object may have been modified by some hooks
 		if (empty($reshook)) {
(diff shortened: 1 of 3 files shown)


htdocs/core/class/commonobject.class.php
```diff
@@ -5350,6 +5354,13 @@ public function printObjectLine($action, $line, $var, $num, $i, $dateSelector, $
 			// Output template part (modules that overwrite templates must declare this into descriptor)
 			// Use global variables + $dateSelector + $seller and $buyer
 			// Note: This is deprecated. If you need to overwrite the tpl file, use instead the hook printObjectLine and printObjectSubLine.
+
+			$qty_shipped = 0;
+			if (isset($this->expeditions[$line->id])) {
+				$qty_shipped = $this->expeditions[$line->id];
+			}
+			$disableedit = $qty_shipped >= $line->qty;
```
eldy, line 5362: May be the disableedit should be one line higher in the if ?

eldy:
My understanding by reading code is that you want to allow to edit order, with a protection to be sure the edition is to change the qty ordered to a value always higher than qty shipped. You confirm ?

zahwyrix:
> My understanding by reading code is that you want to allow to edit order, with a protection to be sure the edition is to change the qty ordered to a value always higher than qty shipped. You confirm ?

Exactly, that's it, today dolibarr is very restrictive by completely blocking the modification of a partially shipped order but for some reason the other part of the order could be cancelled.
So the idea is to put a pencil only on the lines that have not been dispatched or have only been partially dispatched, and at validation check that the quantity dispatched is less than or equal to the quantity ordered, to avoid inconsistencies.

I'd also like to point out that the fact that it's generic with commonObject is great, but it's also very annoying, as in this case where I was just trying to target orders. I got away with if (method_exists($this, 'loadExpeditions')) but I don't find this method very convincing. I'd be more inclined to use methods specific to each object calling the parent method. I don't know, it's just a suggestion.

htdocs/core/class/commonobject.class.php
zephyriony: I've rethought the logic and I don't think you should put it in the if, you do set qty_shipped to 0 outside the if and each line object has a quantity so everything is well defined, you have to recalculate $disableedit each time because it's a loop and each line is different.
Then there's a second solution where you actually put it in the loop and delete qty_shipped before the if, but to do this you have to redefine disableedit at each iteration, otherwise if it doesn't work in the if, it will take the disableedit value of the previous line.

PR #31437, opened 2024-10-17
Fix:edition rapide de la date de livraison et propagation vers la commande

ATM-NicolasV:
# FIX
Pouvoir modifier la date de livraison via le alwayseditable au statut brouillon et propagation de la valeur lors de la création d'une commande

### DIFF
--- a/htdocs/supplier_proposal/card.php
+++ b/htdocs/supplier_proposal/card.php
@@ -1692,7 +1692,8 @@ if ($action == 'create') {
 		print '<table class="nobordernopadding" width="100%"><tr><td>';
 		print $langs->trans('DeliveryDate');
 		print '</td>';
-		if ($action != 'editdate_livraison' && $object->statut == SupplierProposal::STATUS_VALIDATED) {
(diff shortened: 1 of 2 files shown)


htdocs/supplier_proposal/card.php
```diff
@@ -1692,7 +1692,8 @@
 		print '<table class="nobordernopadding" width="100%"><tr><td>';
 		print $langs->trans('DeliveryDate');
 		print '</td>';
-		if ($action != 'editdate_livraison' && $object->statut == SupplierProposal::STATUS_VALIDATED) {
+		// Pourquoi avoir une régle diffrente des autres champs pour l'edition de la date ?
```
eldy, line 1695: I see no reason for this test. Test on not signed as suggested seems better.

Can you remive the comment so i can validate pr. Question on code should be added in github comment, not into code itself.

eldy (changes requested):

htdocs/supplier_proposal/card.php
ATM-NicolasV: yes it's a change i notice for me and i forgot to remove

PR #31438, opened 2024-10-17
FIX : Fix SQL request for fix duplicate accounting account and remove wrong and obsolete SQL request who break the accounting account when we launch repair with multi-entities

kkhelifa-opendsi:
Fix SQL request for fix duplicate accounting account and remove wrong and obsolete SQL request who break the accounting account when we launch repair with multi-entities

### DIFF
--- a/htdocs/install/mysql/migration/repair.sql
+++ b/htdocs/install/mysql/migration/repair.sql
@@ -307,15 +307,6 @@ update llx_societe set barcode = null where (rowid, barcode) in (select max_rowi
 drop table tmp_societe_double;
 
 
--- Sequence to removed duplicated values of llx_accounting_account. Run several times if you still have duplicate.
-drop table tmp_accounting_account_double;
---select account_number, fk_pcg_version, max(rowid) as max_rowid, count(rowid) as count_rowid from llx_accounting_account where label is not null group by account_number, fk_pcg_version having count(rowid) >= 2;
-create table tmp_accounting_account_double as (select account_number, fk_pcg_version, max(rowid) as max_rowid, count(rowid) as count_rowid from llx_accounting_account where label is not null group by account_number, fk_pcg_version having count(rowid) >= 2);
---select * from tmp_accounting_account_double;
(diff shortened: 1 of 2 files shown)


htdocs/install/mysql/migration/repair.sql
```diff
@@ -416,10 +407,33 @@ drop table tmp_bank_url_expense_user;
...
+    GROUP BY account_number, entity, fk_pcg_version
+    HAVING COUNT(*) >= 2
+) AS aad ON
+            aa.account_number = aad.account_number
+        AND aa.entity = aad.entity
+        AND aa.fk_pcg_version = aad.fk_pcg_version
+        AND aa.rowid != aad.min;
+-- SELECT * from tmp_llx_accouting_account;
+-- If there is record in tmp_llx_accouting_account, make a look on each line to do
+UPDATE llx_facturedet        AS t LEFT JOIN tmp_llx_accouting_account AS taa ON taa.rowid = t.fk_code_ventilation SET t.fk_code_ventilation = taa.min WHERE taa.rowid IS NOT NULL;
```
eldy, line 432: I am not sure this kind of syntax Update xxx left join yyy is standard sql accepted by pgsql. Did you test it in postgres ?

kkhelifa-opendsi: it's in /mysql/migration/repair.sql so is it not executed only for mysql ?

eldy: No, it is common to all databases.
Mysql is the database to use for main sql code (so contains allnpart common)  and other dir for specific. Repair is currently common so found into the mysql dir.

rycks: for me https://github.com/Dolibarr/dolibarr/commit/d9d0618eec42f549e82c21d3178f8221a8077550 seems to be ok to fix that bug because entity is added into group by but did somebody have a sql dump with the bug and could you please send me by private mail to run and tests the repair process with pgsql and mariadb/mysql if you can't do it by yourself @kkhelifa-opendsi ?

eldy:

PR #31440, opened 2024-10-17
FIX: Replenishment lib class does not extract data from already received supplier orders

omogenot:
In stock replenishment, the computation of supplier ordered products is not made as the column name in received orders is wrong. Therefore, the number of needed products is wrongly set to 0 with a total number of already ordered (and received) products displayed in the green box.
On the top of that the global variable `STOCK_CALCULATE_ON_SUPPLIER_VALIDATE_ORDER` seems not to be defined and generates an error as it is accessed directly from the $conf array, rather than using the error free getDolGlobalString function.

* Fix wrong table column name from `fk_command` to `fk_element` for table receptiondet_batch
* use error free  getDolGlobalString to access global variable to replace $conf->global access.


eldy:
I pushed a more complete fix. Can you check it is ok for you ?

PR #31441, opened 2024-10-17
FIX #31305 Translation of INVOICE_ADD_EPC_QR_CodePay moved from admin to bills

### LINKED ISSUE #31305: [EPC-QR, Sponge]: No translation of "INVOICE_ADD_EPC_QR_CODEPay", Runtime error/misplaced text when INVOICE_ADD_EPC_QR_CODE==bottom
### Bug

Hi,

While playing with the account name issue in EPC QR, I also noticed that there is no translation of "INVOICE_ADD_EPC_QR_CODEPay" due to a missing domain in loadlangs.

When placing the QR code at the bottom with "INVOICE_ADD_EPC_QR_CODE"==bottom I got a runtime error because $langs was not available there (outputlangs is) and the descriptory text was placed within the QR code.

The attached patch fixes all issues for me. Maybe it's of some use.

Cheers,
Michael

### Dolibarr Version

20.0

### Environment PHP

8.2.20-1~deb12u1

### Environment Database

PostgreSQL 15.8-0+deb12u1

_No response_

### Attached files

Patch fixing translation, QR Text placement (bottom) and runtime error (bottom)
[epc-translation.patch.txt](https://github.com/user-attachments/files/17281419/epc-translation.patch.txt)

shasta123:
Instead of loading the admin domain in lang, I followed your way of moving
the translation of INVOICE_ADD_EPC_QR_CodePay from admin.lang to bills.lang
I used a small script to perform this step on all language directories.


hregis:
@shasta123 you can only modify the en_US, Transifex will modify other languages

shasta123:
Oh, that's even simpler, thanks for the hint. Sorry for the noise.

PR #31442, opened 2024-10-17
Fix DEVELOP pdf space lost with images eratosthene

thersane-john:
# Instructions

NEED #31273 to be merged before Or merge this one to merge both

# FIX : Eratosthene PDF when use images


thersane-john:
@eldy can you check please ?

eldy:
Hi @thersane-john 
I made PR related to PDF generation and pagebreak trouble in standby because before validating it, i want to be able to "see" and "check" easily there is no regression.
It appears to me that result and trouble for page break depends on the number of line in the PDF.
However, the "preview" feature does not allow to choose the number of line making not possible to test the template with a given number of line.
So i would like first to introduce a "test" feature: Being able to add, into the url to generate the preview PDF, a parameter "nblines=x" that will have effect into the function initAsSpecimen() so we will be able easily to test a major change for different number of lines.

So i pushed this commit: 71d988b015677851cc3848736cce7300bc5da332

thersane-john:
@eldy Can you also delay the merges of other PRs (from other people) that would modify the 3 PDFs in question? This is to avoid conflicting merges that would be complicated to manage, given that the code modification is structural.
It will be easier to integrate new fixes and other improvements afterwards, I think.

PR #31443, opened 2024-10-17
Update change select to button on/off

lamrani002:
# NEW|New update in virtual card for user by changing select yes/no  to button on/off
ff


htdocs/user/virtualcard.php
```diff
@@ -70,19 +70,18 @@
...
-	$tmparray['USER_PUBLIC_SHOW_ADDRESS'] = (GETPOST('USER_PUBLIC_SHOW_ADDRESS') ? 1 : 0);
-	$tmparray['USER_PUBLIC_HIDE_COMPANY'] = (GETPOST('USER_PUBLIC_HIDE_COMPANY') ? 1 : 0);
-	$tmparray['USER_PUBLIC_HIDE_JOBPOSITION'] = (GETPOST('USER_PUBLIC_HIDE_JOBPOSITION') ? 1 : 0);
-	$tmparray['USER_PUBLIC_HIDE_OFFICE_PHONE'] = (GETPOST('USER_PUBLIC_HIDE_OFFICE_PHONE') ? 1 : 0);
-	$tmparray['USER_PUBLIC_HIDE_OFFICE_FAX'] = (GETPOST('USER_PUBLIC_HIDE_OFFICE_FAX') ? 1 : 0);
-	$tmparray['USER_PUBLIC_HIDE_USER_MOBILE'] = (GETPOST('USER_PUBLIC_HIDE_USER_MOBILE') ? 1 : 0);
-	$tmparray['USER_PUBLIC_HIDE_SOCIALNETWORKS_BUSINESS'] = (GETPOST('USER_PUBLIC_HIDE_SOCIALNETWORKS_BUSINESS') ? 1 : 0);
-	$tmparray['USER_PUBLIC_MORE'] = (GETPOST('USER_PUBLIC_MORE') ? GETPOST('USER_PUBLIC_MORE') : '');
-
+	$tmparray['USER_PUBLIC_HIDE_PHOTO'] = (getDolGlobalString('USER_PUBLIC_HIDE_PHOTO') ? 1 : 0);
```
eldy, line 73: Because all option ate now set by ajax, we don't need any more this part of cor. Except for the "note". For it, we can keep old code to have the note saved from the user choice (so using GETPOST).

eldy (changes requested):

htdocs/user/virtualcard.php
lamrani002: ok i fix it and i complete the process

htdocs/core/lib/functions.lib.php
```diff
@@ -14850,3 +14850,42 @@ function recordNotFound($message = '', $printheader = 1, $printfooter = 1, $show
 	}
 	exit(0);
 }
+
+/**
+ * Get socal network for user or societe
+ * @param   int      $userId    user Id
+ * @return  int|array    -1 if KO, array if OK
```
mdeweerd, line 14857: array<array{name:string,url:string}>|int<-1,-1> is better/needed.

htdocs/core/lib/functions.lib.php
```diff
@@ -14850,3 +14850,42 @@ function recordNotFound($message = '', $printheader = 1, $printfooter = 1, $show
 	}
 	exit(0);
 }
+
+/**
+ * Get socal network for user or societe
```
mdeweerd, line 14855: Small typo to fix: 'social'

htdocs/core/lib/functions.lib.php
mdeweerd: It looks like the type of the User class also needs an update

User::$socialnetworks

htdocs/public/users/view.php
```diff
@@ -341,11 +341,15 @@
 }
 
 // Social networks
+if (empty($object->socialnetworks)) {
```
eldy, line 344: if $object->socialnetworks is empty then getSocialNetworks($object->id) will return empty too.
Because $object->socialnetworks is loaded by the fetch of user the same way the getSocialNetworks do it.

PR #31460, opened 2024-10-18
FIX: Fix return value of hook sendMail when hook return -1 who must be return false in sendfile() function

kkhelifa-opendsi:
Fix return value of hook sendMail when hook return -1 who must be return false in sendfile() function


kkhelifa-opendsi:
Fixed in develop

rycks:
ok, maybe enhance that PR : please have a look at line 1209 and 1226 of the same file with returns other things that a boolean :)

rycks (changes requested):
ok, maybe enhance that PR : please have a look at line 1209 and 1226 of the same function with returns other things that a boolean :)

lvessiller-opendsi (commented):
Maybe complete with https://github.com/Dolibarr/dolibarr/commit/8bf54dee9ea7acb16a6d6c6140516caf0655d429

PR #31462, opened 2024-10-18
Fix v20 linked doc note info repeated

thersane-john:
# Instructions
If you generate several PDF models with the same object, the public notes data are duplicated as many times as there are PDF models to generate.

Need PR #31459 to be merged first

Note : In this case, use `$object->note_public` isn't a good way to add data to PDF


eldy:
I merged 31459 but have now conflict. Can solve them ?

thersane-john:
@eldy done

PR #31483, opened 2024-10-20
 NEW: PDF add discount total if line discount exists

sonikf:


eldy:
Can you provide a screenshot or pdf of section before and after change ?

sonikf:
Before

b

After (with next pr preview frameless)

a

PR #31489, opened 2024-10-20
Qual: Fix phan/stan notices for facture-rec.class

mdeweerd:
# Qual: Fix phan/stan notices for facture-rec.class

Fix phan/phpstan notices in facture-rec.class


mdeweerd:
@eldy 
This PR is now limited to the facture-rec.class file with your changes.

PR #31491, opened 2024-10-20
fix phpstan

frederic34:


mdeweerd:
FYI, there is some overlap with #31485 .

PR #31494, opened 2024-10-20
NEW Accountancy report, grouped by project

ymollard:
# NEW Accountancy report, grouped by project

This new report derived from `clientfourn.php` can be used either in income/expenses mode or in document mode.
It presents a report when all of them are grouped by project.

Bookkeeping mode not implemented (so far).


eldy:
Due to error in code syntax, i merged the report page manually.

PR #31517, opened 2024-10-22
FIX - Wrong Typing in DataBase

atm-lucasmantegari:
# FIX - *Wrong Typing in DataBase*
To follow this hidden conf: 
“MAIL_MAX_NB_OF_RECIPIENTS_IN_SAME_EMAIL”
It is important to change the type of the email_to, email_tocc and email_tobcc columns. They must be changed from varchar(255) to TEXT. This change allows you to accommodate more data and avoids an event creation error when sending mail. Currently, it's possible to change the conf value to 100 (for example). The e-mail is sent to 100 people, but the event is not created because the type is not capable of handling that much data.


eldy:
Change o database structure is not possible on already released version.
Can you push it on develop ?

PR #31524, opened 2024-10-23
FIX#31237

### LINKED ISSUE #31237: Wrong thrird-party linked by EmailCollector class on precssing customer emails
### Bug

When EmailCollector class is enables and Collect_Ticket_Requests option is enabled, after processing the customer emails, if the Third-party with that email address is already present, then EmailCollector class selecting wrong thirdparty. You can see the EmailCollector class settings below.
EmailCollector class settings

You can also check the EmailCollector class output below.
EmailCollector class output

### Dolibarr Version

20.0

### Environment PHP

8.1.25

### Environment Database

MySQL

1. Create a Third-party with and email address at Dolibarr
2. Setup EmailCollector class with your IMAP email
3. Send email from that Third-party email to your support IMAP email
4. Run EmailCollector class

### Attached files

_No response_

alsoft10:
# FIX|Fix #31237 
[*Long description*]

Wrong third-party linking by Emailcollector class has been fixed more accurate way.

### DIFF
--- a/htdocs/societe/class/societe.class.php
+++ b/htdocs/societe/class/societe.class.php
@@ -2136,7 +2136,12 @@ class Societe extends CommonObject
 		$tmpthirdparty = new Societe($this->db);
 		$result = $tmpthirdparty->fetch($rowid, $ref, $ref_ext, $barcode, $idprof1, $idprof2, $idprof3, $idprof4, $idprof5, $idprof6, $email, $ref_alias, $is_client, $is_supplier);
 
-		return $result;
+		//If the row is found by previous operation return it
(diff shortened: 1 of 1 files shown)


alsoft10:
@eldy issue #31237 is fixed.

eldy:
Thanks for pointing to the solution. In dolibarr dev rules says that a fetch should return the id found when a record is found.
So can you check @alsoft10 if this other fix is ok: https://github.com/Dolibarr/dolibarr/commit/58c0b97300d7663ecadd8e8e10302c9b9b38fedd

PR #31535, opened 2024-10-24
Fix PDF watermark alpha transparency

thersane-john:
# Instructions
The watermark uses an opaque rose instead of a transparent red.
With this fix it is now possible to read the text below the watermark.


pcouy:
Funny, I just did this on my own instance and was about to create almost the exact same PR. Good thing I checked for duplicates before opening it ;)

PR #31536, opened 2024-10-24
NEW: Use multiple element types to attach extra fields to

omogenot:
# NEW Allow multiple element types to attach extra fields to

The proposal is to allow modules to define an extra field common to several element types at the same time so that this field would be propagated throughout descendant objects.
The `$elementype` parameter can contain several types comma separated.
```
Ex.:

include_once DOL_DOCUMENT_ROOT.'/core/class/extrafields.class.php';
$extrafields = new ExtraFields($this->db);
$result=$extrafields->addExtraField('myextrafield', 'extrafield', 'text', 1, 128, 'propal,commande,expedition,facture', 0, 1, '', '', 0, "", -1, 0, '', '', 'mymodule@mymodule', 'isModEnabled("mymodule")');
```

In this case, an extra field named `myextrafield` is shown in each card of element type : `propal`, `commande`, `expedition` and `facture` 
Not only this extra field has the same definition (text, 128 characters), but when the user creates a `commande` object form a `propal` object, the newly created `commande` object extra field is automatically populated with the `propal` extra field content. (I must admit that this was the expected effect, but I don't know why it occurred without any code change for that 😄 )

This upgrade does not prevent the original way of working of the extra fields. And the user can even change the extra field definition thought the regular module setup `extra fields` tab. However, this multi-element type extra field feature can only be created  programmatically. It looks difficult, and useless, to propose to create such extra fields using the regular user's interface (where would you put it? ).

Feel free to complete or comment this proposal to make it evolve if need be.

Note: sorry for the multiple commits due to a global copy/paste from eclipse to GitHub => TABs were a bit messy.


htdocs/core/actions_extrafields.inc.php
```diff
@@ -386,18 +386,25 @@
 				// Example: is_object($object) ? ($object->id < 10 ? round($object->id / 2, 2) : (2 * $user->id) * (int) substr($mysoc->zip, 1, 2)) : 'objnotdefined'
 				$computedvalue = GETPOST('computed_value', 'nohtml');
 
+				// Load $extrafields->attributes
+				$extrafields->fetch_name_optionals_label($elementtype);
+				$attrname = GETPOST('attrname', 'aZ09');
+				$perms = $extrafields->attributes[$elementtype]['perms'][$attrname] ?? '';
+				if ((isset($_GET['perms'])) || (isset($_POST['perms']))) {
```
eldy, line 393: For this line, you can use
`if (GETPOSTISSET('perms')) {`

eldy:
The current existing rule in application is this one: When we clone on object, or create an object from another one, we propagate value of extrafields if the code of extrafield is similar and only in this case.
This means we can already manage extrafields that are common to several objects. however, the GUI is not firendly for this (it need to be created or modified several time on each object).
Adding a better GUI is possible, however, i see no reason to store the extrafield definition differently than currently. We are already "technically" able to define an extrafield, to detect it as existing onto other object and to modify it. 
So the need seems on interface only.
And when we need to change the interface, it is better to suggest screenshots to understand how it will work for user.

omogenot:
Hi @eldy 

Thanks for the explanation. I must admit that I did not find this definition in the description of the extra fields in the documentation (or I did not see it).
Therefore, this proposal can be considered as useless and you can just close it if you consider that there is no need for such a feature. I will revert the code to the original in my system for the time being.
I needed this feature whilst I was developing a module to allow the use of a kind of generic empty `thirdparty`, without any contact, for which the user can in each card creation freely and manually set the contact information that will be then shown on each document (propal, commande, expedition, invoice). A bit like the POS works with a special thirdparty record to attach all activities to.
Knowing what you explained, I just have to use 4 times the same extra field name for each element. I wrongly thought that the extra field name had to be unique in the db, whereas this is true for the association extra field name + associated element.
Out of this work, however, some of the code could be adapted, if need be, to allow the function `fetch_name_optionals_label` to accept a comma separated list of elements to look for (like 'propal,commande' for instance). This will allow the function call to be more specific and avoid the usage of 'all' too many times, that could be time and resources consuming. You let me know if this is of any interest, I can do another PR for that if it is of any interest (just let me know).
In any case, thanks for your help.
Olivier.

eldy:
Yes the behavior of propagation of extrafield is documented nowhere, just implemented like this. So i added a tooltip in this commit 060720124df9c52a3d6b21f3c1a6421ea578b74e

About adding a param to allow to call fetch_name_optionals_label with elem1+elem2 in one step to avoid to call
fetch_name_optionals_label with elem1 and then fetch_name_optionals_label with elem2.
This is possible, but effect will be low. In such case calling with "all" is not really an extraload because the quantity of record to describe all extrafields is always low (often lower than 100) and done into table with content that rarely change, so allcontent is alreay in memory cache of the database. Getting all has same cost that getting elem1+elem2. So we could call it with 'all' when we need.

So for performance, it will have no effect. But for clarity of code, it may be interesting to allow a param separated with a coma to accept n values of elem...

omogenot:
@eldy ,

OK I close this PR and eventually open a new one just for the multiple type element in `fetch_name_optionals_label` 

Bye for now.

PR #31549, opened 2024-10-25
NEW Add thirdparty state in substitution array

joseplluis:
# NEW Add thirdparty state in substitution array
Add thirdparty state in substitution array, and also mysoc state and the member state

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -8906,7 +8906,7 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 			'__MYCOMPANY_ADDRESS__' => $mysoc->address,
 			'__MYCOMPANY_ZIP__'     => $mysoc->zip,
 			'__MYCOMPANY_TOWN__'    => $mysoc->town,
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -8905,6 +8906,7 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 			'__MYCOMPANY_ADDRESS__' => $mysoc->address,
 			'__MYCOMPANY_ZIP__'     => $mysoc->zip,
 			'__MYCOMPANY_TOWN__'    => $mysoc->town,
+			'__MYCOMPANY_STATE__'    => getState($mysoc->state_id),
```
eldy, line 8909: Simple substition key `__XXX__` must never call methods that do SQL request.
If we need this we must use lazy load substitution key with syntax 
`__XXX__@lazyload` 
See https://wiki.dolibarr.org/index.php?title=Variable_substitution_system#How_it_work

But in your case, may be you can use directly
`'__MYCOMPANY_STATE__' => $mysoc->state,`

joseplluis: Perfect, I've already applied the change. This is easier than the initial proposal. Thank you!

Issue #31378, opened 2024-10-13
Add payment reference number and types


sonikf:
### Feature Request

Payment reference number mainly helps companies identify payments. 

### Use case

Using payment reference number in invoices and PDF(invoice, proposal, order) makes it easy for debtor to quickly pay and for creditor upon receipt of the payment, to link the payment to their sales invoice.

### International
Customer code, contract ref, invoice ref or a combination can be used as a unique payment id.

### Europe
Through Single Euro Payments Area (SEPA) banks already implement various methods.

1. Belgium is using the so called Structured communication based on ISO 20022 standard.
https://support.nickel.eu/en-be/how-do-i-create-structured-communication-when-making-transfer

2. Finland is using its own implementation of Structured Creditor Reference(RF), based on
 ISO 11649 with a string of 4-20 characters based on the invoice number.

3. Germany, Greece, Switzerland and possibly other countries are using Structured Creditor Reference(RF),  an international business standard based also on  ISO 11649 implementation with an alphanumeric string, up to 25 characters long which adds more flexibility as it can also be used per third party.

relevant link
https://www.dolibarr.org/forum/t/how-are-you-using-dolibarr-without-being-able-to-issue-payment-reference-numbers/24929/22
https://pbs.twimg.com/media/ECkcCF-U4AAoJ0c.jpg:large

### Suggested implementation

1.
- [x] Add field payment_reference in llx_facture table
- [x] Add field tp_payment_reference in llx_societe table

2.
- [ ] Add method to create user configurable payment_reference using Customer code, contract ref, invoice ref or a combination.
- [x] Commit https://github.com/Dolibarr/dolibarr/pull/31376 for Belgian type
- [ ] Add QR code for Finnish RF
- [ ] Add Finnish legacy payment reference
- [ ] Add Finnish legacy payment barcode (PDF) / virtual barcode (email)
- [ ] Add methods for "European" RF generation and QR code support
- [x] Structured communication in sepa xml https://github.com/Dolibarr/dolibarr/pull/31383
- [ ] Add RF in sepa xml
- [ ] Add FI/RF in sepa xml 
- [ ] Add payment_reference in EPC QR code and Swiss QR-Bill 

3.
- [ ] Add option for user to select payment reference  type 
- [ ] Add per third party RF generation options like customer/vendor code in societe/card.php
- [ ] Show selected type generated field in validated invoice, proposal, sales cards
- [ ] Add payment_reference in relevant PDF templates

[...]

PR #31553, opened 2024-10-25
NEW Add payment reference - SQL Part

sonikf:
See https://github.com/Dolibarr/dolibarr/issues/31378

### DIFF
--- a/htdocs/install/mysql/migration/20.0.0-21.0.0.sql
+++ b/htdocs/install/mysql/migration/20.0.0-21.0.0.sql
@@ -282,3 +282,5 @@ CREATE TABLE llx_product_customer_price_extrafields (
 ) ENGINE=innodb;
 
 ALTER TABLE llx_product_customer_price_extrafields ADD UNIQUE INDEX uk_product_customer_price_extrafields (fk_object);
+ALTER TABLE llx_facture ADD COLUMN payment_reference varchar(25) AFTER date_lim_reglement;
+ALTER TABLE llx_societe ADD COLUMN tp_payment_reference varchar(25) AFTER code_fournisseur;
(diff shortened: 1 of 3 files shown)


htdocs/install/mysql/migration/20.0.0-21.0.0.sql
```diff
@@ -282,3 +282,5 @@ CREATE TABLE llx_product_customer_price_extrafields (
 ) ENGINE=innodb;
 
 ALTER TABLE llx_product_customer_price_extrafields ADD UNIQUE INDEX uk_product_customer_price_extrafields (fk_object);
+ALTER TABLE llx_facture ADD COLUMN payment_reference varchar(25) AFTER date_lim_reglement;
```
eldy, line 285: An invoice can be paid by several payment. So data for payments should be store into the table llx_payment and table for invoice.
What is the use case of this field ?

htdocs/install/mysql/migration/20.0.0-21.0.0.sql
```diff
@@ -282,3 +282,5 @@ CREATE TABLE llx_product_customer_price_extrafields (
...
+ALTER TABLE llx_societe ADD COLUMN tp_payment_reference varchar(25) AFTER code_fournisseur;
```
eldy, line 286: Why adding another field. Why not using the already existing field customer code ?

mikko80: Payment reference is calculated data for bank payment and sepa payments.

sonikf: Hi,
There are 2 ways of use, per invoice and per customer
In Greece per customer(we use RF) is widely used from companies issuing recurring bills e.g.: electricity, gas, telephony-internet etc
This way every customer has to know only his/her unique payment reference and amount to pay the bills each month
Per invoice is usually used in  no recurring bills
Payment reference are then included in the bank statements received by vendor who can automatically process payments - reconcile

sonikf: Some more important info...

Structured payment references like international  RF main difference with a custom reference like customer code you mentioned is that they **must** be used through your cooperating payment service provider(Banks, electronic payments providers etc).

You have to contact your bank first to let you use it(Banks need to know to whose account e.g every RF corresponds)

So you can't just generate them and use them in the payment info field like you do with the customer code, invoice ref or any other way any company uses.

PR #31556, opened 2024-10-25
FIX add a control before difining "DOL_DOCUMENT_ROOT" in filefunc.php

thomas-Ngr:
# FIX add a control before difining "DOL_DOCUMENT_ROOT" in filefunc.php

 It may generate errors otherwise on module activation.

'PHP Warning:  Constant DOL_DOCUMENT_ROOT already defined in /var/www/dolitest/htdocs/[filefunc.inc](http://filefunc.inc/).php on line 245'


eldy:
if you got this warning, it measn you run twice the filefunc.in.php and you probably have a more serious other problem.
What use case do you do to get this warning ?

eldy:

PR #31558, opened 2024-10-25
Fix a missing css fontawesome

MaximilienR-easya:
# Fix

Add a CSS removed from a previous PR #26040 that break some fontawesome icon
in example the icon fa-refresh

### DIFF
--- a/htdocs/main.inc.php
+++ b/htdocs/main.inc.php
@@ -1958,6 +1958,7 @@ function top_htmlhead($head, $title = '', $disablejs = 0, $disablehead = 0, $arr
 			print '<!-- Includes CSS for font awesome -->'."\n";
 			$fontawesome_directory = getDolGlobalString('MAIN_FONTAWESOME_DIRECTORY', '/theme/common/fontawesome-5');
 			print '<link rel="stylesheet" type="text/css" href="'.DOL_URL_ROOT.$fontawesome_directory.'/css/all.min.css'.($ext ? '?'.$ext : '').'">'."\n";
+			print '<link rel="stylesheet" type="text/css" href="'.DOL_URL_ROOT.$fontawesome_directory.'/css/v4-shims.min.css'.($ext ? '?'.$ext : '').'">'."\n";
 		}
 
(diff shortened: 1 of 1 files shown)


eldy:
We should be able to get the refresh picto without adding this old file.
If you correctly use the img_picto('', 'refresh') then the font awesome for refresh is output as 'fa-redo' and shown correctly.

How do you reproduce the bug ? Which value do you use into img_picto to get the fa-refresh or where do you find a fa-refresh hardcoded in dolibarr ?

MaximilienR-easya:
We are using fa-refresh inside one of our module, and we detected the broken picto when we were testing for compatibility with the version 20 of Dolibarr.
We found that the change that broke the compatibility of our module was this PR #26040
This PR was to fix what looks like a file removed accidentally.

eldy:
The file was not removed accidently.
This file "v4-shims.min.css" was introduced into v12 (or before) when upgrading fontawesome from v4 to v5, to keep immediate compatibility in v12 at no cost for Dolibarr main code and external modules that were using fontawesome v4 to be compatible with the new font awesome v5, to allow time to make the transition of Dolibarr code and external module to use the new font awesome v5.
However after 10 major versions, the authors of modules should have enough time to fix their modules to work natively with the v5 version of font awesome, this is why there is no more reason to keep this legacy code in v20.
If you are still using legacy code of v4 fontawesome in modules, you must upgrade your module because we can't cumulate indefinitely the compatibility layers.

MaximilienR-easya:
> However after 10 major versions, the authors of modules should have enough time to fix their modules to work natively with the v5 version of font awesome, this is why there is no more reason to keep this legacy code in v20. If you are still using legacy code of v4 fontawesome in modules, you must upgrade your module because we can't cumulate indefinitely the compatibility layers.

We are not on fontawesome-4 but in fontawesome-6 and fa-refresh is present on it. It was more on a backward compatibility with the basic installation of Dolibarr which is on fontawesome-5 that this file was useful to us.
We are going to alter our CSS on our module to be backward compatible with fontawesome-5 and closing this PR.

PR #31568, opened 2024-10-26
Qual: phpstan api_*/Ignore 'no value type specified in iterable type …

mdeweerd:
…array'

# Qual: phpstan api_*/Ignore 'no value type specified in iterable type array'

This phpstan notice seems to be manually added as an exception and creates conflicts when merging.

So this updates phpstan.neon.dist for all these notices in the api_* files.


mdeweerd:
@frederic34 I think this is what you are trying to achieve.

PR #31587, opened 2024-10-27
fix phpstan phpdoc

frederic34:


htdocs/compta/sociales/class/chargesociales.class.php
```diff
@@ -39,6 +39,11 @@ class ChargeSociales extends CommonObject
 	 */
 	public $element = 'chargesociales';
 
+	/**
+	 * @var string Name of table without prefix where object is stored
+	 * @deprecated
```
mdeweerd, line 44: I recommend to add "Use $...." after `@deprecated` because this shows up in the phan reports.

frederic34: is it good now ?

mdeweerd: Yes, that will show up when derpacation notices are enabled.

PR #31593, opened 2024-10-27
NEW: add duration extrafield

benoitpascal:
# NEW|New Add duration extrafield

Add a extrafield with type duration that react has task duration type.


frederic34:
you need to add new features only in develop branch

benoitpascal:
> you need to add new features only in develop branch

Sorry, I will fix my commit then :wink:

eldy:
> > you need to add new features only in develop branch
> 
> Sorry, I will fix my commit then :wink: 

I recommand you install git precommit hook. See https://wiki.dolibarr.org/index.php?title=Environment_and_development_tools

So your code syntax will be automatically fixed and this will avoid ct/ci errors.

If not possible, think to setup your ide to use tabs instead of space for tabs.

benoitpascal:
The other PR is the cleaned one. 

I close this PR to avoid merging the wrong one.

PR #31595, opened 2024-10-27
NEW: add duration extrafields

benoitpascal:
# NEW duration extrafields
Add extrafields to deal with durations on the same pattern as task duration fields.

Duration is registered in seconds into database.
Value is displayed with two inputs 'hour' and 'minute'

I had to bypass @GETPOSTISSET control because value is splitted in two inputs and none has 'options_$key' key.

### Extrafield selector
Capture d’écran du 2024-10-27 22-41-11

### Input
Capture d’écran du 2024-10-27 22-42-04

### Output
Capture d’écran du 2024-10-27 22-42-18

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,7 +43,7 @@ repos:
       - id: end-of-file-fixer
       # Check that there are no completely merged file conflicts
       - id: check-merge-conflict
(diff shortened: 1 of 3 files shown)


eldy:
Duplicate of #31593 ?

benoitpascal:
> Duplicate of #31593 ?

Yes. This one is the cleaned commit. I close the other one.

PR #31601, opened 2024-10-28
fix const  WORKFLOW_RECEPTION_CLASSIFY_CLOSED_INVOICE

thomas-Ngr:
# FIX rename const WORKFLOW_EXPEDITION_CLASSIFY_CLOSED_INVOICE to WORKFLOW_RECEPTION_CLASSIFY_CLOSED_INVOICE

The constant had been renamed in logic code on commit #82f69c6 .

But changes had not been done to config page and lang files.

This PR also fixes the comparison between floats in this workflow since floating-point comparison should not be done directly.

image


eldy:

PR #31603, opened 2024-10-28
FIX php-stan incorrect case newtoken

lvessiller-opendsi:
FIX php-stan incorrect case newtoken

PR #31604, opened 2024-10-28
Update ignore_translation_keys.lst

frederic34:


frederic34:
need also this https://github.com/Dolibarr/dolibarr/pull/31603

PR #31609, opened 2024-10-28
FIX: changing page with the pageplusone input is not working

jyhere:
The bug can be reproduced on supplier order / supplier proposal lists, but the same code is found on many other pages, not sure if the fix is OK to deploy on all pages...

Reproducing the bug : 
Go to the supplier order list, enter a new page in the input field, press enter, the page input field is still on 1.
image


eldy:
The good fix is the one suggested in modulebuilder/template/myobject_list.php

PR #31617, opened 2024-10-29
Qual: Fix phan/phpstan notices in class files (part 5)

mdeweerd:
# Qual: Fix phan/phpstan notices in class files (part 5)

Fix phan and phpstan notices in .class files and related notices/code.

### DIFF
--- a/htdocs/expedition/class/expeditionligne.class.php
+++ b/htdocs/expedition/class/expeditionligne.class.php
@@ -511,8 +511,8 @@ class ExpeditionLigne extends CommonObjectLine
 		$qty = price2num($this->qty);
 		$remainingQty = 0;
 		$batch = null;
-		$batch_id = null;
-		$expedition_batch_id = null;
+		$batch_id = 0;
+		$expedition_batch_id = 0;
 		if (is_array($this->detail_batch)) { 	// array of ExpeditionLineBatch
 			if (count($this->detail_batch) > 1) {
 				dol_syslog(get_class($this).'::update only possible for one batch', LOG_ERR);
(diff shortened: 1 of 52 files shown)


mdeweerd:
Prefious parts:
- #31578 
- #31598

htdocs/expedition/class/expeditionligne.class.php
```diff
@@ -605,7 +605,7 @@ public function update($user = null, $notrigger = 0)
 						$shipmentLot->sellby = $lot->sellby;
 						$shipmentLot->entrepot_id = $this->detail_batch->entrepot_id;
 						$shipmentLot->qty = $this->detail_batch->qty;
-						$shipmentLot->fk_origin_stock = $batch_id;
+						$shipmentLot->fk_origin_stock = (int) $batch_id;
```
eldy, line 608: This means we will store 0 in database instead of null currently ?

mdeweerd: Ok, this looks bothersome as it says fk_origin_stock is not null (in v20 sqldump of demo):
```plaintext
dev/initdemo/mysqldump_dolibarr_20.0.0.sql:5712:4:  `fk_origin_stock` int(11) NOT NULL,
```

Also confirmed in my local database.


I have updated the code where a line is added to the DB - I did not find anything that updates a line.

However, should we maintain null not be compliant with the DB structure?

eldy: You're right. What rules is the file into /install/mysql dir and we have 
fk_origin_stock integer NOT NULL
so it's ok for forcing to int so to 0 if not defined.

PR #31621, opened 2024-10-29
Prefer using given field properties instead of object fields in Commo…

atm-etienne:
# New - Prefer to use given field properties instead of object ones
In showInputField method of the CommonObject class, prefer to use the given field properties to show if not empty, instead of force replacing it by the object ones


atm-etienne:
In case we want to add new field in a common object list, from the hook, if the field is not from the object itself, it won't be displayed

PR #31627, opened 2024-10-29
NEW VAT by state (SQL)

lvessiller-opendsi:
NEW VAT by state (SQL)
- the SQL structure to add VAT rate in departments dictionary


eldy:
In nearly all countries, there is several vat rates. Si having one field into a table of department will lead to a blocked situation.
For this reason, i think it is better to add a column fk_department into llx_c_tva. We can already store n vat rates into this table and flag them per country, so we can also flag them per department too (default will be 0 = all department) by introducing the field into llx_c_tva.

lvessiller-opendsi:
> In nearly all countries, there is several vat rates. Si having one field into a table of department will lead to a blocked situation. For this reason, i think it is better to add a column fk_department into llx_c_tva. We can already store n vat rates into this table and flag them per country, so we can also flag them per department too (default will be 0 = all department) by introducing the field into llx_c_tva.

I understand what you want to do.
However, if I have several departments in the same country with the same VAT rate, I should insert several lines with the same VAT rate. One line for each department (with the same VAT rate). Is it correct ?

**For example**
Let's suppose 
- "85GUA" VAT code is for "Guadeloupe" department
- "85MAR" VAT code is for "Martinique" department
image
So here if the department of the buyer is "Guadeloupe" the VAT rate selected by default will be "85GUA". 
And if department of the buyer is "Martinique" the VAT rate selected by default wil be "85MAR".

eldy:
> > In nearly all countries, there is several vat rates. Si having one field into a table of department will lead to a blocked situation. For this reason, i think it is better to add a column fk_department into llx_c_tva. We can already store n vat rates into this table and flag them per country, so we can also flag them per department too (default will be 0 = all department) by introducing the field into llx_c_tva.
> 
> I understand what you want to do. However, if I have several departments in the same country with the same VAT rate, I should insert several lines with the same VAT rate. One line for each department (with the same VAT rate). Is it correct ?
> 
> **For example** Let's suppose
> 
> * "85GUA" VAT code is for "Guadeloupe" department
> * "85MAR" VAT code is for "Martinique" department
>   image
>   So here if the department of the buyer is "Guadeloupe" the VAT rate selected by default will be "85GUA".
>   And if department of the buyer is "Martinique" the VAT rate selected by default wil be "85MAR".

Yes, if you want to have a management of vat rate per state, if you have to enter several time the vat rate for each state, even if value is same. 
Whatever is the side of the relation 1-n, we may experience duplicates on the side n. The only solution to avoid absolutely no duplicate at all is to have a relation n-n but clearly here, we must avoid this as this increase complexity just to avoid few duplicates not so annoying.
Also don't forget that if the numeric value is same, the other properties (like the accountancy code of vat rate) may differ (for example the accounting account for Guadeloupe may differ of Martinique for accounting analytic purpose).
So having the n on the side of the vat table seems a better choice (we already have chosen this side for the country of seller, we must remain on the same side for the state of the buyer). I would also use a clear column name to be clear that it is the state of buyer so field could be fk_department_buyer

lvessiller-opendsi:
@eldy 
I changed the way to set vat rate by department.
Now the department is in the vat rate dictionary.
I only get some "PHPStan" errors on unchanged files.
Could you have a check ?

PR #31628, opened 2024-10-29
NEW VAT by state

lvessiller-opendsi:
NEW VAT by state
- for SQL structure see #31627

**1. Define the VAT default rate for a state**
image

Here you can select a VAT rate from dictionary (the empty value is needed to reset the VAT rate to NULL).
If department is NULL, it will keep the default vat rules for the country.

**2. Rules for VAT rates are :** 
By default when creating prospects, invoices, orders etc. the Sales Tax rate follows the active standard rule:
- RULE 1 : If the seller is not subject to Sales tax, then Sales tax defaults to 0. End of rule.
- RULE 2 : If buyer department has a VAT rule from vat rates dictionary then it's the default VAT rate. End of rule.
- RULE 3 : If the (seller's country = buyer's country), then the Sales tax by default equals the Sales tax of the product in the seller's country. End of rule.
- RULE 4 : If the seller and buyer are both in the European Community and goods are transport-related products (haulage, shipping, airline), the default VAT is 0. This rule is dependant on the seller's country - please consult with your accountant. The VAT should be paid by the buyer to the customs office in their country and not to the seller. End of rule.
- RULE 5 : If the seller and buyer are both in the European Community and the buyer is not a company (with a registered intra-Community VAT number) then the VAT defaults to the VAT rate of the seller's country. End of rule.
- RULE 6 : If the seller and buyer are both in the European Community and the buyer is a company (with a registered intra-Community VAT number), then the VAT is 0 by default. End of rule.
- RULE 7 : In any other case the proposed default is Sales tax=0. End of rule.

**3. Create your proposal, order or invoice**
The default VAT is set to the buyer state VAT if exists
image

If there is not VAT set for the buyer state, the previous rules are kept.
image


eldy:
As explained into #31627, link between vat rate and department should be into llx_c_tva instead of llx_department.

lvessiller-opendsi:
@eldy 
I put link between vat rate and department into "llx_c_tva" table.
Could you have a look ?

PR #31631, opened 2024-10-29
NEW Add Filter Assistance for Filling Filters Field on Report Page

Daoud-mohamed:
# Filter Assistance for Filling Filters Field on Report Page

This PR introduces a filter assistance feature to the report page, enhancing user experience by guiding users in selecting appropriate filters. The assistance dynamically suggests options based on object and the selected field type, improving filter accuracy and ease of use.

image

image

image


htdocs/core/customreports.php
```diff
@@ -887,7 +926,26 @@
 	$sqlfilters = $search_component_params_hidden;
 	if ($sqlfilters) {
 		$sql .= forgeSQLFromUniversalSearchCriteria($sqlfilters, $errormessage, 0, 0, 1);
+
+		// Replace year, month, day with %Y, %Y-%m, %Y-%m-%d. For example, (t.datef-year := '2024') becomes WHERE DATE_FORMAT(t.datef, '%Y') = '2024'
```
eldy, line 930: We must not use DATE_FORMAT to filter on date. This is not enable to use the index on date.
Instead you can replace all this block to use the function
`dolSqlDateFilter(...)`

htdocs/core/lib/functions.lib.php
```diff
@@ -13702,7 +13702,7 @@ function forgeSQLFromUniversalSearchCriteria($filter, &$errorstr = '', $noand =
 		$filter = '(' . $filter . ')';
 	}
 
-	$regexstring = '\(([a-zA-Z0-9_\.]+:[<>!=insotlke]+:[^\(\)]+)\)';	// Must be  (aaa:bbb:...) with aaa is a field name (with alias or not) and bbb is one of this operator '=', '<', '>', '<=', '>=', '!=', 'in', 'notin', 'like', 'notlike', 'is', 'isnot'
+	$regexstring = '\(([a-zA-Z0-9_\.\-]+:[<>!=insotlke]+:[^\(\)]+)\)';	// Must be  (aaa:bbb:...) with aaa is a field name (with alias or not) and bbb is one of this operator '=', '<', '>', '<=', '>=', '!=', 'in', 'notin', 'like', 'notlike', 'is', 'isnot'
```
eldy, line 13705: I think you introduce the "-" char to allow fields like t.datef-year
I think for filter, we should not have the date-year or date-month  fields, only the date. This variant were introduced for the X axis but is not interesting for filter because to get a good SQL syntax it is better to do a date > YYYY0101 and date < YYYY1231 instead of dateformat. So lets just ask date with YYYY MM DD

In a next next we will allow to enter year > ... but we must forge date >  YYYY0101

PR #31634, opened 2024-10-29
Qual: Fix phan/phpstan notices in class files (part 6)

mdeweerd:
# Qual: Fix phan/phpstan notices in class files (part 6)

Fix phan/phpstan notices in class files and some related files (final batch of 'bulk' fixes for .class.php files).


mdeweerd:
Prefious parts:
- #31598
- #31617

PR #31208, opened 2024-09-30
Update ticket.class.php Bug-fix  Issue #31206

cbeasley-sbb:
Fix Issue #31206 for the v.18 branch because unrelated tests failing on develop branch.
Fix Issue #31206  

Sets the object's  "fk_statut" to Ticket::STATUS_READ. In the event of rollback reverts the object's "fk_statut" to its old value.

### NOTE:  I need this fix to be applied to v18.x if and when approved.

PR #31642, opened 2024-10-30
FIX update status on create supplier order for trigger

lvessiller-opendsi:
FIX update status on create supplier order for trigger
- it's similar as PR #31208
- the status is not set in the trigger and if the external module can do an update of the object and the status is set to NULL in database

**Remarks**
- dol_clone() function uses "2" as second argument to keep the same code for upper versions of Dolibarr
Should I backport the full function in Dolibarr v14 ?

- status attribute is used in commonobject
- statut attribute is used in this supplier order object (see update method)

### DIFF
--- a/htdocs/fourn/class/fournisseur.commande.class.php
+++ b/htdocs/fourn/class/fournisseur.commande.class.php
@@ -1291,6 +1291,8 @@ class CommandeFournisseur extends CommonOrder
 	{
 		global $langs, $conf, $hookmanager;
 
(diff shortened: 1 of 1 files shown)


lvessiller-opendsi:
@eldy 
I got a Tavis error but I don't know why.
It not seems to be link with this PR.

eldy:
When error is in travis at the very begin of process,you can try to restart travis. If you don't have permission on restart button for travis, it can be doneby repushing i think.

PR #31673, opened 2024-11-03
NEW Custom accounting reports

aspangaro:
MAIN_FEATURES_LEVEL = 2


aspangaro:
Problem with travis... No error proposed

PR #31676, opened 2024-11-03
NEW Accountancy - Add general account on thirdparty - Code user part

aspangaro:


htdocs/accountancy/journal/bankjournal.php
```diff
@@ -1044,20 +1050,21 @@
 					print '"'.$key.'"'.$sep;
 					print '"'.$date.'"'.$sep;
 					print '"'.$val["type_payment"].'"'.$sep;
-					print '"'.length_accountg(html_entity_decode($k)).'"'.$sep;
```
eldy, line 1047: Line 1047 to 1060 was outputing 2 columns (1 column by line 1047 and 1 by lines 1048 to 1060).
The new code replace this with ouput of 1 column only. Do you confirm ?

aspangaro: You're right, my code creates a column offset. I've pushed a new patch.

We have a single general-account column responding to length_accountg() and another sub-account column using the length_accounta() function.

Before :
image

After : 
image

PR #31682, opened 2024-11-04
fix trim null value on invoice creation

uvaldenaire-opendsi:
FIx trim() with null value on invoice creation when $this->note_private is null.


frederic34:
in version develop : 
image

lvessiller-opendsi (approved):

rycks (approved):
ok for me, maybe same fix for note_public (next line) ?

PR #31683, opened 2024-11-04
FIX : parent status of company in card company

Thomas905:
We add a parent to the child, then we close the child tier and in the list of subsidiaries we have the status at hover which is not good the company status is will always open.

AFTER :
image


lvessiller-opendsi (approved):

rycks (approved):
ok for me, thanks !

Issue #31015, opened 2024-09-18
Can't add payment through API since upgrade to 20.0


ro78:
### Bug

Hello,

Since upgrade to version 20.0, I can't add a payment to an invoice through the API.
I'm getting this error:

`{
    "error": {
        "code": 400,
        "message": "Bad Request: Payment error : Incorrect datetime value: '' for column `dolibarr`.`llx_paiement`.`datep` at row 1"
    },
    "debug": {
        "source": "api_invoices.class.php:1516 at call stage",
        "stages": {
            "success": [
                "get",
                "route",
                "negotiate",
                "authenticate",
                "validate"
            ],
            "failure": [
                "call",
                "message"
            ]
        }
    }
}`

The request has not changed since months:
`CallDOAPI("POST", "/invoices/$facid/payments", json_encode(array("datepaye"=>time(), "paymentid"=>6, "closepaidinvoices"=>"yes", "accountid"=>5)));`

time() returns the expected value (tried with another value and I'm getting the "Bad Request: Invalid value specified for `datepaye`. Expecting unix timestamp, such as 1726652385" error).

The function called is:
`$paymentobj->datepaye     = dol_stringtotime($datepaye);`

In version 19.0.3, it was:
`$paymentobj->datepaye     = $datepaye;`

### Dolibarr Version

20.0.0

### Environment PHP

8.2.23

### Environment Database

MariaDB 10.11.6

### Steps to reproduce the behavior and expected behavior

See above

ro78:
I did this change and the API call is working again:
`//$paymentobj->datepaye     = dol_stringtotime($datepaye); --> $paymentobj->datepaye     = $datepaye;`

ro78:
@eldy @hansemschnokeloch is it normal to not have this fix included in 20.0.1?

PR #31703, opened 2024-11-05
FIX api paiement datepaye type, change from timestamp to datetime

marsender:
Database field type datec of llx_paiement has changed from timestamp to datetime but this change has not been passed on the paiement api class.


eldy:
> Database field type datec of llx_paiement has changed from timestamp to datetime but this change has not been passed on the paiement api class.

I am not sure that the fix in #31015 was a good idea. i modified so we can send both an 
int (timestamp) or a date string ('YYY-MM-DD')
by doing
```
		if (is_numeric($datepaye)) {
			$paymentobj->datepaye = $datepaye;
		} else {
			$paymentobj->datepaye = dol_stringtotime($datepaye);
		}
```
Is this ok for you ?

PR #31704, opened 2024-11-05
New : Mass action for closing cash controls

bafbes:
New : Mass action for closing cash controls


htdocs/core/actions_massactions.inc.php
```diff
@@ -1133,6 +1134,30 @@
 	//var_dump($listofobjectthirdparties);exit;
 }
 
+// close cashpoint from mass action
```
eldy, line 1137: if you need to add a mass action that is specific to a page/object, you must add it into the file compta/cashcontrol/cashcontrol_list.php
You can add this test just before the include of the generic mass action

bafbes: Modifications done

eldy:
Mass action for delete has been implemented.
Closing need interaction (selection of period, of amountin cashdesk) so can't be implemented as massaction.

PR #31706, opened 2024-11-05
Fix: unit output (use_short_label)

Humml87:
Now the same short Labels are used.

image

P.S. in transiflex no resources exist for ....Short, which is generated from 'measuringUnitString'.


eldy:
We use the param 1 because we are in a table where we must save spaces, so we ask to use the very short label. So for "piece", we should get "p".
this is strange, old code 
`$unitforline = measuringUnitString($line->fk_unit, '', '', 1);`
seems to return correctly  "p"
If using 2, we will get PieceShort. So 1 seems the correct value. Did i miss something ?

Humml87:
other pr

PR #31707, opened 2024-11-06
FIX|Fix # Change Class Name of Mark Rates on Objectlines_view tpl

atm-corentin:
# FIX|Fix # class name of MarginRate <td> tag 
I noticed that when I added Margin rates and Mark rates on a proposal, both td tags had the same class name, which is incorrect. The td tag for Margin had the class name linecolmargin2, which is correct, but the td tag for Mark rates also had linecolmargin2, which is incorrect and should be linecolmark1


eldy:
If you change it in this filed, you should probably also fix the same way on other files.
Also as this does not fix directly a known user bug, you should push the PR in develop.

PR #31709, opened 2024-11-06
FIX ticket list : type_code multiselect search can be an empty value.…

thomas-Ngr:
# Fix multiselect on ticket list would allow empty value for type_code

On ticket list, the search for type_code is a multiselect. It can receive an empty value that would mean 'type_code is not set'.
Since type_code is a mandatory value, a search on value 'is not set' should not be allowed.

Capture d’écran_2024-11-06_11-49-56

### DIFF
--- a/htdocs/ticket/list.php
+++ b/htdocs/ticket/list.php
@@ -402,6 +402,12 @@ foreach ($search as $key => $val) {
 		continue;
 	} elseif ($key == 'type_code') {
 		$newarrayoftypecodes = is_array($search[$key]) ? $search[$key] : (!empty($search[$key]) ? explode(',', $search[$key]) : array());
+		$newarrayoftypecodes = array_filter(
+			$newarrayoftypecodes,
+			function ($var) {
+				return !(empty($var) && $var !== '0');
+			}
+		);
 		if (count($newarrayoftypecodes)) {
 			$sql .= natural_search($key, join(',', $newarrayoftypecodes), 3);
 		}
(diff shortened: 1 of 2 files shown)


eldy:
The empty value into the select list means "We don't want any filter". And we should be able to not set any filter on this field so the empty choice should remain into the combo, but no filter should be done in this case.

thomas-Ngr:
> The empty value into the select list means "We don't want any filter". And we should be able to not set any filter on this field so the empty choice should remain into the combo, but no filter should be done in this case.

If you chose the empty value, you get a nice error page (here an abstract)

image


with this nice request

> SELECT t.rowid,t.entity,t.ref,t.track_id,t.fk_user_create,t.origin_email,t.subject,t.type_code,t.category_code,t.severity_code,t.fk_soc,t.notify_tiers_at_create,t.fk_project,t.datec,t.date_read,t.date_last_msg_sent,t.fk_user_assign,t.date_close,t.tms,t.message,t.email_msgid,t.progress,t.email_date,t.resolution,t.fk_statut,t.import_key FROM llx_ticket as t LEFT JOIN llx_societe as s ON (t.fk_soc = s.rowid) WHERE t.entity IN (1) AND (type_code IN ()) ORDER BY t.datec DESC LIMIT 21

containing this nice WITH clause : 

>  AND (type_code IN ())

It seems that asking 'no filter' is different from asking 'filtering on empty value', both on UI/UX and on filtering logic behind.

no filter:
image

filter empty value (leads to SQL error) : 
image

thomas-Ngr:
You are right. Problem is, if we chose this empty value we get a nice error page (here an abstract). Tested on a fresh V17 install without any module.

image

with this nice request

> SELECT t.rowid,t.entity,t.ref,t.track_id,t.fk_user_create,t.origin_email,t.subject,t.type_code,t.category_code,t.severity_code,t.fk_soc,t.notify_tiers_at_create,t.fk_project,t.datec,t.date_read,t.date_last_msg_sent,t.fk_user_assign,t.date_close,t.tms,t.message,t.email_msgid,t.progress,t.email_date,t.resolution,t.fk_statut,t.import_key FROM llx_ticket as t LEFT JOIN llx_societe as s ON (t.fk_soc = s.rowid) WHERE t.entity IN (1) AND (type_code IN ()) ORDER BY t.datec DESC LIMIT 21

containing this nice WITH clause : 

>  AND (type_code IN ())

___

The solution is to remove empty values from the search array. User may still be confused by being able to chose empty value, but at least the search will work.

eldy:
Solution is to fix the part of code that generates the SQL by adding a test so if no filter is provided, the SQL must not include the "AND (type_code IN ..." at all.

htdocs/ticket/list.php
```diff
@@ -402,6 +402,12 @@
 		continue;
 	} elseif ($key == 'type_code') {
 		$newarrayoftypecodes = is_array($search[$key]) ? $search[$key] : (!empty($search[$key]) ? explode(',', $search[$key]) : array());
+		$newarrayoftypecodes = array_filter(
```
eldy, line 405: What is value of $newarrayoftypecodes here ?

eldy: So we must found when the entry 0 => '' is added to disabled it.

PR #31710, opened 2024-11-06
NEW|add member civility to the labels

AlainRnet:
# NEW|New add member civility to the labels
as described on https://www.dolibarr.fr/forum/t/modification-carte-ou-licence-adherent/36013/15
But I don't know how to do to add extrafields. If anybody has an idea to get me on the track...

Note that the global variable ADHERENT_ETIQUETTE_TEXT does not work. \n appear in text instead of a return line, only those of this file return a new line, as described on https://www.dolibarr.fr/forum/t/personnaliser-champs-a-imprimer-sur-etiquettes-pdf-des-adherents/47668


hregis:
@AlainRnet request civility code instead the label and use langs->trans with civility code

AlainRnet:
You’re absolutely right, thank you for the tip!
Do you have a lead for the extrafields, which are "options" based on what I see in the MySQL rendering: options_". $key ??

hregis:
@AlainRnet try the tag `__OPTIONS_XXXX__` XXX is the $key in upper case

AlainRnet:
Thank you @hregis this niew code works for me.

To understand: in the next line, what is the function `complete_substitutions_array($substitutionarray, $langs, $adherentstatic);`? Strangely, if I comment on this line, it does not change the rendering?!?

hregis:
@AlainRnet "complete_substitutions_array" is for external module for completed/personnalized the substitution

PR #31716, opened 2024-11-06
FIX: Corrected typo preventing data from being saved

atm-quentin:
# FIX Corrected typo preventing data from being saved

This Pull Request addresses a bug caused by a typo in the code, which prevented data from being saved properly. The issue was tracked down to a small but critical error in one of the variables, which caused the save functionality to fail. The typo has been fixed, ensuring that data can now be saved as expected without any errors.


frederic34:
@atm-quentin also at line 376

atm-quentin:
> @atm-quentin also at line 376

thanks

PR #31717, opened 2024-11-06
add phpdoc and clean phpstan baseline

frederic34:


htdocs/accountancy/class/accountingjournal.class.php
```diff
@@ -211,7 +211,7 @@ public function getNomUrl($withpicto = 0, $withlabel = 0, $nourl = 0, $moretitle
 		}
 		if ($withlabel == 2 && !empty($this->nature)) {
 			$key = $langs->trans("AccountingJournalType".$this->nature);
-			$transferlabel = ($this->nature && $key != "AccountingJournalType".strtoupper($langs->trans($this->nature)) ? $key : $this->label);
+			$transferlabel = ($key != "AccountingJournalType".strtoupper($langs->trans($this->nature)) ? $key : $this->label);
```
eldy, line 214: Can you confirm we can remove the test on  $this->nature
The test was here to avoid error when doing $langs->trans($this->nature) when ->nature is not set or empty.

frederic34: There is already a test at line 212

PR #31728, opened 2024-11-08
Fix function num_between_day in date.lib.php

josett225:
When debugging Asset Module, I discovered that function num_between_day is returning nothing when  $timestampStart equals (==) $timestampEnd
Normally it has to return 
0 if $lastday = 0
1 if $lastday = 1


josett225 (commented):
Clean code

josett225:
I will do another as I believe there are conflicts between my patch-7 that was supposed to change version 12 and this one that is supposed to changed dev version.

frederic34:
@josett225 perhaps you need to update your fork

image

you can easily do that here

image

PR #31755, opened 2024-11-09
fix calculate costs bom

frederic34:
image


htdocs/bom/class/bom.class.php
```diff
@@ -1471,8 +1471,11 @@ public function calculateCosts()
 					}
 				} else {
 					// Convert qty of line into hours
-					$unitforline = measuringUnitString($line->fk_unit, '', '', 1);
-					$qtyhourforline = convertDurationtoHour($line->qty, $unitforline);
+					require_once DOL_DOCUMENT_ROOT.'/core/class/cunits.class.php';
+					$measuringUnits = new CUnits($this->db);
+					$measuringUnits->fetch($line->fk_unit);
+					// $unitforline = measuringUnitString($line->fk_unit, '', '', 1);
+					$qtyhourforline = $line->qty * $measuringUnits->scale;
```
eldy, line 1478: The scale is the power of 10 between a unit and reference unit. For exemple 3 for kg means that unit is 10^3 = 1000 * g
So this line is suspicious. Didn't find what was wrong with old code. What was the value for fk_unit and unitforline you got beforefix ?

frederic34: unit is time here

eldy: You're right, the nature of unit is "time", so the scale is directly the factor.
However, the variable $qtyhourforline, according to its name is the amount of hours, not of second.
And $qtyhourforline = $line->qty * $measuringUnits->scale will give a number of second.
May be we should do:

```
$qtysecforline = $line->qty * $measuringUnits->scale;
$qtyhourforline = convertDurationtoHour($qtysecforline, 's');
```

lmag:
Month and year can not a time unit !
image

PR #31763, opened 2024-11-10
fix phpstan

frederic34:


.github/workflows/phpstan.yml
```diff
@@ -44,7 +44,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php-version }}
-          tools: phpstan, cs2pr
+          tools: phpstan:1.12, cs2pr
```
frederic34, line 47: @mdeweerd to stay on 1.12 before switching on v2

mdeweerd: Thanks for the head's up.

PR #31701, opened 2024-11-05
NEW: use Expedition::$date_shipping property in card and list

altairis-noe:
# NEW use Expedition::$date_shipping property in card and list
Currently, Expedition object does have a `$date_shipping` property, stored as `date_expedition` in the database.

But there is no way in Dolibarr interface to set or display its value.

If your customer wants to know when he will receive the goods, you might want to know when it is gone from your facilities. Here is my proposition.

image

image

image

Issue #31652, opened 2024-10-31
Digital/Electronic shipment sheet part of E-Reporting


sonikf:
### Feature Request

According to EU proposal

https://taxation-customs.ec.europa.eu/taxation/vat/vat-digital-age-vida_en

VAT in the Digital Age (ViDA) will eventualy replace Vies.
E-invoicing and e-reporting will be mandatory across Europe by 2030

Some more info and timeline

https://www.grantthornton.ie/insights/factsheets/vat-in-the-digital-age-vida-approaches/
https://www.vatcalc.com/eu/eu-vat-in-the-digital-age-vida-adopted-by-ec/

Top five implementations based on results are from Greece, Italy, Poland, Portugal and  Spain and reports for these member states will act as reference for the rest.

**Contributors from above countries please advise on following feature request**

EU member states have already started ahead of schedule.
In Greece starting from 1/4/2025 Digital/Electronic shipment sheet will be mandatory for all legal entities
Starting from 1/7/2025 E- invoicing will be mandatory for all legal entities(another issue will follow) 

Based on Greek implementation api, new fields(SQL part) should be added to core to also facilitate future core or module implementations

### Use case

Implementation for simple shipment
Process made easy through scanning single QR code
![00](https://github.com/user-attachments/assets/888e6c5f-fae4-422c-bb72-ffb1e7bd546e)

For ease of use an invoice can act also as shipment so new fields needed  in both tables
French Chorus pro Greek equivalent screenshot

![0](https://github.com/user-attachments/assets/464b3c8a-554d-4392-aac7-109123874b36)

### Suggested implementation

|Field|Type|Description|Dolibarr|
|-|-|-|-|
|IssueDate|date|shipment issue date|llx_facture     OK                                                         llx_expedition KO|
|correlatedInvoices|BigInt|correlated invoices|llx_facture    KO                                                  llx_expedition KO|
|qrCodeUrl|string |response received after issue of invoice|llx_facture    KO                                                  llx_expedition KO|
|dispatchDate|date|shipment start date|llx_facture    KO                                                  llx_expedition KO|
|dispatchTime|time|shipment start time(when vehicle is loaded and ready to begin)|llx_facture    KO                                                    llx_expedition KO|
|vehicleNumber|string|licence plate of vehicle|llx_facture    KO                                         llx_expedition KO|
|movePurpose|int|Purpose of movement from list/dictionary of values..|llx_facture     KO                                           llx_expedition KO|
[...]

PR #31773, opened 2024-11-11
New: prepare for e-shipment (SQL part)

sonikf:
Related to #31652

Hi @eldy 

My guess is that by 1/4/2025(compulsory date for Greece) v 22.0.0  will not have  been released yet, so i kindly request  for the inclusion to v. 21.0.0 of the minimum changes needed for e-shipment to work meaning this pr + #31701 first + a pr following to show fields in modules card.php

A few remarks
- General rule is that for every item exiting business facilities a shipment must be issued and there are many use cases not supported such as:

|Type|use case examples|Dolibarr|
|-|-|-|
|internal movements of assets(not financial)|move a pc from one building office to another building office |ko|
|internal movements of  stock|from main warehouse to secondary warehouse, warehouse to store etc |ko|
|repair - maintenance|faulty equipment, motherboard fails,equipment service etc|ko|
|Sampling - Exhibition and/or Free distribution|event organization-conference also involves an equipment movement|ko|
|Return|damaged goods from vendor|ko|
|Recycling or Destruction of waste material |manufacturing|ko|

- Right now shipment is  treated as a secondary document by being dependant to sales orders.

- Following comment seems wrong, warehouse is used for stock management and not for per product departure this is not how logistics work.
As per above examples a warehouse can be a starting point or a destination of a shipment
https://github.com/Dolibarr/dolibarr/blob/40aca2803f1b329bb95d8d784d0c87eba2353362/htdocs/install/mysql/tables/llx_expeditiondet.sql#L31

- Shipment address is used through contacts but is not recorded to shipment or invoice as it should. Same would apply to loading address

If you agree i will work for independent shipment creation for v 22.0.0  plus support for above use cases


eldy:
I validated PR but :
- I remove field datee: seems duplicate of the field date_expedition. Or i miss something ?
- I renamed  is_delivery_note into is_also_delivery_note because if my understanding is good, the invoice will be both a real invoice (that affect the turnover and go into accounting) and is also used as a shipment document to avoid to duplicate documents. This is not invoice XOR shipment but Invoice alone, or invoice + shipment, right ?

sonikf:
Ηι @eldy 
Thanks for  validation 
We also need datee as issue date(shipment has no issue date because it depends on order)  + https://github.com/Dolibarr/dolibarr/pull/31701 because as @altairis-noe explains we need the date shipment actually starts/gone  meaning when loaded to a company truck for example or received from a shipment/delivery service.

workflow example:
today 12/11/2024 i issue Proposal or order with planned date(date_delivery=18/11/2024) 
Friday 15/11/2024 i issue Invoice or shipment with issue date (datef, datee=15/11/2024) and also actual starting date (date_expedition=Tuesday 19/11/2024). 

In future this date(date_expedition) will be filled by e.g. shipment/delivery service driver by scanning QR CODE and transmitted to all related parties (The way this is designed based on EU VIDA proposal is that  info provided to both vendor and mainly client will be in near real time so there will be no need for tracking number in future. Authorities need to fight illegal trade and vat fraud)

eldy:
> Ηι @eldy 
> Thanks for  validation 
> We also need datee as issue date(shipment has no issue date because it depends on order)  + https://github.com/Dolibarr/dolibarr/pull/31701 because as @altairis-noe explains we need the date shipment actually starts/gone  meaning when loaded to a company truck for example or received from a shipment/delivery service.
> 
> workflow example:
> today 12/11/2024 i issue Proposal or order with planned date(date_delivery=18/11/2024) 
> Friday 15/11/2024 i issue Invoice or shipment with issue date (datef, datee=15/11/2024) and also actual starting date (date_expedition=Tuesday 19/11/2024). 
> 
> In future this date(date_expedition) will be filled by e.g. shipment/delivery service driver by scanning QR CODE and transmitted to all related parties (The way this is designed based on EU VIDA proposal is that  info provided to both vendor and mainly client will be in near real time so there will be no need for tracking number in future. Authorities need to fight illegal trade and vat fraud)

So the issue date in your exemple seems to be the date of creation of shipment ?

sonikf:
Yes ok we can use date_creation, thanks

PR #31775, opened 2024-11-12
Fix:hook for adding params for credit notes creation

bafbes:
Fix:hook for adding params for credit notes creation.
The hook should provide by js url params in creditNoteParams variable


bafbes:
Hi, just following up on this PR.  
Tests are passing — is there anything I should change or improve before it can be merged?  
Thanks for your feedback.

PR #31778, opened 2024-11-12
[FIX] Colspan error in card.php

cmfpmatik:
The replacement of the code is necessary to address a layout issue in the table caused by incorrect column handling, which can affect the proper display of data.

Explanation of the Problem

In the original code:
    The table displays the HT (pre-tax) value in one column, and if the MAIN_SHOW_PRICE_WITH_TAX_IN_SUMMARIES variable is enabled, an additional column is added for the TTC (tax-included) value.
    This means that the number of columns in the table changes depending on the configuration: an extra column appears when MAIN_SHOW_PRICE_WITH_TAX_IN_SUMMARIES is active.
    This variation in the number of columns can cause issues with the table layout, especially if other cells (like headers or summary rows) use a colspan attribute that does not match the total number of columns, resulting in misalignment.

Why Replace with the New Code?

The new code solves this issue by displaying both the HT and TTC values in the same column, separated by a visual divider, when MAIN_SHOW_PRICE_WITH_TAX_IN_SUMMARIES is enabled:

If MAIN_SHOW_PRICE_WITH_TAX_IN_SUMMARIES is enabled: The HT value is followed by the TTC value within the same cell, maintaining a consistent column count.
 Otherwise: Only the HT value is displayed, still within a single cell.

Conclusion :

Consistent Column Count: The number of columns remains constant, regardless of the MAIN_SHOW_PRICE_WITH_TAX_IN_SUMMARIES setting. This ensures that the colspan attribute in other parts of the table remains accurate, preserving the layout.

The code modification ensures that the table layout is correct and that columns remain alwys properly aligned, preventing layout issues that could occur if the number of columns.


htdocs/comm/card.php
```diff
@@ -1490,13 +1490,13 @@
...
 					}
 				}
-				print '<td class="right nowraponall">';
-				print price($objp->total_ht);
-				print '</td>';
-
-				if (getDolGlobalString('MAIN_SHOW_PRICE_WITH_TAX_IN_SUMMARIES')) {
-					print '<td class="right nowraponall">';
-					print price($objp->total_ttc);
+				if (!empty($conf->global->MAIN_SHOW_PRICE_WITH_TAX_IN_SUMMARIES)) {
```
eldy, line 1493: Must use 
if (getDolGlobalString('MAIN_SHOW_PRICE_WITH_TAX_IN_SUMMARIES')) {

eldy:
The goal of option MAIN_SHOW_PRICE_WITH_TAX_IN_SUMMARIES is to add one column.
Each information must be into its dedicated column.
Can't reproduce trouble with old code. Can you describe how to reproduce the bug(with screenshot)

cmfpmatik:
> The goal of option MAIN_SHOW_PRICE_WITH_TAX_IN_SUMMARIES is to add one column.

Okay, but you have to increase the colspan first in this case.

> Each information must be into its dedicated column. Can't reproduce trouble with old code. Can you describe how to reproduce the bug(with screenshot)

Why not! But in merge #24518, I preferred this approach for many reasons, but you mentioned : "_Both fields represent the same information, only the unit differs. So i suggest to keep only one line: The original line and you just add the value in the second unit on the same line_"

cmfpmatik:
up

PR #31796, opened 2024-11-13
ODF : fix : escape specials chars in IF ELSE ENDIF $key

thomas-Ngr:
# FIX|Fix Use internal constants in ODT files

The [wiki](https://wiki.dolibarr.org/index.php?title=Create_an_ODT_or_ODS_document_template#Conditional_substitution) states that ODT substitution allows using internal constants :

> `{__[XXX]__} = Value of the constant XXX (into setup)`

However, this does not work since the chars `[` and `]` are special chars char used in PCRE expressions.

These keys should be escaped.

## Example

here, the ODT template : 

image

MY_VERY_SPECIAL_CONST = 1

The generated ODT is as so : 

image

after applying the fix, the generated ODT is right:

image

## Configuration

Bug observed on DLB14, DLB18, develop.

## Implementation

Note that the first IF is correctly deleted since the function used is str_replace. The ELSE and ENDIF are not deleted since they are searched using preg_replace function.

```php
odtphp/segment.php l.118
                // Remove the IF tag
                $this->xml = str_replace('[!-- IF '.$key.' --]', '', $this->xml);
                // Remove everything between the ELSE tag (if it exists) and the ENDIF tag
                $reg = '@(\[!--\sELSE\s' . $key . '\s--\](.*))?\[!--\sENDIF\s' . $key . '\s--\]@smU'; // U modifier = all quantifiers are non-greedy
                $this->xml = preg_replace($reg, '', $this->xml);
```
At this stage, `$key == '[MY_VERY_SPECIAL_CONST]'` so the regex is not what we mean.

The implementation fixes this bug by applying `preg_quote($key)` when using `$key` in regex.

### DIFF
--- a/htdocs/includes/odtphp/Segment.php
+++ b/htdocs/includes/odtphp/Segment.php
@@ -117,14 +117,14 @@ class Segment implements IteratorAggregate, Countable
                 // Remove the IF tag
                 $this->xml = str_replace('[!-- IF '.$key.' --]', '', $this->xml);
                 // Remove everything between the ELSE tag (if it exists) and the ENDIF tag
-                $reg = '@(\[!--\sELSE\s' . preg_quote($key) . '\s--\](.*))?\[!--\sENDIF\s' . preg_quote($key) . '\s--\]@smU'; // U modifier = all quantifiers are non-greedy
+                $reg = '@(\[!--\sELSE\s' . preg_quote($key, '@') . '\s--\](.*))?\[!--\sENDIF\s' . preg_quote($key, '@') . '\s--\]@smU'; // U modifier = all quantifiers are non-greedy
                 $this->xml = preg_replace($reg, '', $this->xml);
             }
             // Else the value is false, then two cases: no ELSE and we're done, or there is at least one place where there is an ELSE clause, then we replace it
             else
             {
                 // Find all conditional blocks for this variable: from IF to ELSE and to ENDIF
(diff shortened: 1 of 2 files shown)


htdocs/includes/odtphp/Segment.php
```diff
@@ -117,14 +117,14 @@ public function merge()
                 // Remove the IF tag
                 $this->xml = str_replace('[!-- IF '.$key.' --]', '', $this->xml);
                 // Remove everything between the ELSE tag (if it exists) and the ENDIF tag
-                $reg = '@(\[!--\sELSE\s' . $key . '\s--\](.*))?\[!--\sENDIF\s' . $key . '\s--\]@smU'; // U modifier = all quantifiers are non-greedy
+                $reg = '@(\[!--\sELSE\s' . preg_quote($key) . '\s--\](.*))?\[!--\sENDIF\s' . preg_quote($key) . '\s--\]@smU'; // U modifier = all quantifiers are non-greedy
```
eldy, line 120: If you use preg_quote, you must use the second parameter to be sure result is the one expected.
Can you replace with
`preg_quote($key, '@') `
and tell if solution still works...

thomas-Ngr: I tried, it works. And it would also work if the constant name contains a `@`. Thanks for the reminder.

PR #31797, opened 2024-11-13
fix link if subdirectory to remove contains numbers

frederic34:


htdocs/ecm/class/ecmfiles.class.php
```diff
@@ -1007,7 +1007,7 @@ public function getNomUrl($withpicto = 0, $option = '', $notooltip = 0, $maxlen
 		}
 
 		if ($option) {
-			$url = DOL_URL_ROOT.'/document.php?modulepart='.$option.'&file='.urlencode(preg_replace('/[a-zA-Z]+\//', '', $this->filepath).'/'.$this->filename).'&entity='.$this->entity;
+			$url = DOL_URL_ROOT.'/document.php?modulepart='.$option.'&file='.urlencode(preg_replace('/[a-zA-Z0-9]+\//', '', $this->filepath).'/'.$this->filename).'&entity='.$this->entity;
```
eldy, line 1010: my understanding is that you remove the first level "directory" name into filepath.
So why not 
`preg_replace('/^[^\/]+\//', '', $this->filepath)`
to remove the starting part ".../" into the filepath.

frederic34: image
I want to remove produit/ on this example, but if your module is called my123module, it only remove module/ from my123module/REF456

PR #31805, opened 2024-11-13
fix phpstan

frederic34:


htdocs/accountancy/bookkeeping/export.php
```diff
@@ -487,9 +487,9 @@
 	// Mass actions
 	$objectclass = 'Bookkeeping';
 	$objectlabel = 'Bookkeeping';
-	$permissiontoread = $user->hasRight('societe', 'lire');
-	$permissiontodelete = $user->hasRight('societe', 'supprimer');
-	$permissiontoadd = $user->hasRight('societe', 'creer');
+	$permissiontoread = ($user->hasRight('societe', 'lire') == 1);
```
eldy, line 490: Does not look a good idea to do
$permissiontoread = ($user->hasRight('societe', 'lire') == 1);
everywhere when this is more natural :
$permissiontoread = $user->hasRight('societe', 'lire');

What's wrong with having 
$permissiontoread = 1 ?
May be can set such var to be both boolean and int ?

frederic34: image
`if ($permissiontoread) ` with $permissiontoread is int phpstan can't determine if test if right

eldy: Stupid phpstan. So what do you think of:
`$permissiontoread = (bool) $user->hasRight('societe', 'lire');`
instead of 
`$permissiontoread = ($user->hasRight('societe', 'lire') == 1);`

frederic34: I tried...

PR #31818, opened 2024-11-14
fix add v20 in script for contributors

Hystepik:
add v20 in script dev/tools/github_commits_byversion.sh


mdeweerd:
Good but maybe you could reuse what I did in the setup script so that the script works for *all* versions:

https://github.com/Dolibarr/dolibarr/blob/1d1c381e98606f22dc4fdea9e05077b57b8b949c/dev/setup/phpunit/setup_conf.sh#L223-L226

https://github.com/Dolibarr/dolibarr/blob/1d1c381e98606f22dc4fdea9e05077b57b8b949c/dev/setup/phpunit/setup_conf.sh#L233-L264

Issue #31741, opened 2024-11-08
Emailing module throws error in PHP8.2 but not in PHP7.4


hasse-basse:
### Bug

When trying to set up a massmailing and trying to send a Test mail I fail with an error when using PHP8.2, but all is fine with PHP7.4. I use Password and Swift Mailer socket library (smtp.office365.com)

AH01071: Got error 'PHP message: PHP Fatal error:  Uncaught TypeError: preg_split(): Argument #2 ($subject) must be of type string, array given in /htdocs/includes/swiftmailer/lib/classes/Swift/Mime/Headers/AbstractHeader.php:312\nStack trace:\n#0 /htdocs/includes/swiftmailer/lib/classes/Swift/Mime/Headers/AbstractHeader.php(312): preg_split()\n#1 /htdocs/includes/swiftmailer/lib/classes/Swift/Mime/Headers/AbstractHeader.php(259): Swift_Mime_Headers_AbstractHeader->getEncodableWordTokens()\n#2 /htdocs/includes/swiftmailer/lib/classes/Swift/Mime/Headers/UnstructuredHeader.php(103): Swift_Mime_Headers_AbstractHeader->encodeWords()\n#3 /htdocs/includes/swiftmailer/lib/classes/Swift/Mime/SimpleHeaderSet.php(309): Swift_Mime_Headers_UnstructuredHeader->getFieldBody()\n#4 /htdocs/...', referer: https://WWW.XXX/XXX/htdocs/comm/mailing/card.php?action=test&token=c9911ad4d077eabb07f5597a92641508&id=15

### Dolibarr Version

19.0.3

### Environment PHP

PHP 8.2.15

### Environment Database

Maria

marc-dll:
Hello,

I cannot reproduce with the latest Dolibarr (19.0.4 from Github), can you test with the latest code please ?

Greetings.

hasse-basse:
Still same problem and now I'm on 19.0.4. To be sure we talk about the same thing, pls see the below image of where in Dolibarr I am.
![Capture d’écran 2024-11-14 232348](https://github.com/user-attachments/assets/f4d0bef5-d9c8-49b9-8e59-81fe410afb5c)

marc-dll:
Thank you for your quick reply, I reproduced by setting the field "Errors to/Erreurs à" field that was not set in my first test. I'll work on a fix

hasse-basse:
Great !

PR #31825, opened 2024-11-14
FIX: swiftmailer: correctly set errors-to header (FIX #

marc-dll:
Swiftmailer's `addTextHeader` method does not handle array parameters like `setTo`, `setCC`, etc. So we format the `Errors-To` mail header as a `string` by hand.

Fixes issue #31741


marc-dll:
Close because bug exists in 18.0

PR #31835, opened 2024-11-15
Fix dol_dir_list_in_database to use filters for db

Hystepik:
Fix dol_dir_list_in_database to use filters as Universal Search string to filter sql request


eldy:
Note: Looks to be a first step to fix a bug into another ticket. So ok as fix but mentioning the target ticket in comment will be welcome.

PR #31841, opened 2024-11-15
fix: serialize arrays to persist them

benoitpascal:
Array attributes are not serialized and save as 'Array'.

When a field was created as an array, it was registered in database with wrong value : 'Array'
Capture d’écran du 2024-11-15 16-14-57

To avoid breaking functionnalities and loosing data, I serialized before saving in database and I unserialize to use typed attribute later
Capture d’écran du 2024-11-15 16-17-11


eldy:
Can you describe the goal of the change.
As it is flagged as a Fix, can you describe how to reproduce the bug ?

eldy:
It is strange to have Array in database. How do you reproduce this ?

eldy:

Issue #30203, opened 2024-06-30
TakePOS (Barcode rule to insert product) is not working


ems-co:
### Bug

i have a product which i want to extract the amount from the generated barcode, for example a product from a weighing scale, the scale is generating barcodes and the weight of the product embedded e.g 2112345000001 , the zeros are used for the weight, i have enabled Barcode rule to insert product (ref:7+qu:2+qd:3+other:1) but when i scan the item with the barcode e.g 2112345012401 , the part 2112345 is finding the product reference but the quantity and decimal quantity is not added to the POS product lines.

### Dolibarr Version

19.0.2

### Environment PHP

8.2

### Environment Database

Mysql

### Steps to reproduce the behavior and expected behavior

-Create a product with reference e.g 2112345
-Enable POS module and activate barcode rule to insert products (ref:7+qu:2+qd:3+other:1)
-Write an example barcode e.g 2112345012401 which is a product of quantity 1.240
-it wont be added to the POS line items

lvessiller-opendsi:
@ems-co 
I will check it and see how to fix it.

lvessiller-opendsi:
Fix in PR #31857

PR #31857, opened 2024-11-15
FIX TakePos barcode rule

lvessiller-opendsi:
FIX TakePos barcode rule
Issue https://github.com/Dolibarr/dolibarr/issues/30203

### DIFF
--- a/htdocs/takepos/ajax/ajax.php
+++ b/htdocs/takepos/ajax/ajax.php
@@ -212,7 +212,7 @@ if ($action == 'getProducts') {
 							if (isset($barcode_value_list['qd'])) {
 								$qty_str .= '.' . $barcode_value_list['qd'];
 							}
-							$qty = floatval($qty_str);
+							$qty = (float) $qty_str;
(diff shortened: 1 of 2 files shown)


rycks:
Hello,
i'm sorry but i'm in trouble because two years ago we did the opposite commit, please have a look here:

https://github.com/Dolibarr/dolibarr/commit/cad8b1ee7e84f704b3f66f3bda54940ee31e6cec "Fix autoadd in basket after a search (only if search on barcode)"

what do you think about that ?

maybe i need more time to test and set up a dolibarr with that use case

lvessiller-opendsi:
@rycks 
You're right I will see an another solution to do it.

rycks:
note about quality : $term is most of the time used for session terminal number

image

then it could be a quality fix to change term= by something like search= or barcode= .... and a new variable $search or $barcode then $term will be only for terminal number

rycks (approved):
ok for me (and "quality" note is a "global" remark)

Issue #31856, opened 2024-11-15
Assets: error 500 when creating asset with no depreciation


marc-dll:
### Bug

Division by zero when creating an asset with no depreciation
![Uploading Capture d’écran de 2024-11-15 15-20-14.png…]()

### Dolibarr Version

20.0

### Environment PHP

8.2.x

### Environment Database

MariaDB

FHenry:
fix by https://github.com/Dolibarr/dolibarr/pull/31858

PR #31858, opened 2024-11-15
FIX: assets: division by zero when trying to calculate depreciation on assets that don't have it

marc-dll:


marc-dll:
Fixes #31856

Issue #19493, opened 2021-11-26
Wrong number of files in the electronic document management system


mister-pc:
# Bug
The number of uploaded can be wrong if we upload many files at one time : The number indicated reflects the number of uploads, not the number of files.

## Environment
- **Version**: 14.0.1
- **OS**: Debian
- **Web server**: Apache
- **PHP**: 7.4
- **Database**: MySQL
- **URL(s)**: /ecm/index.php

## Steps to reproduce the behavior
In a new folder under the EDM tab, when browsing for uploading files, just select many ones with the Ctrl-key (or the Shift-key) down. The displayed number of files uploaded is 1, instead of the real number.

## Attached files
[*Files*
![2021-11-26_14-16-01](https://user-images.githubusercontent.com/25563509/143587000-d9d9a6f7-ab8b-4485-a44c-34d74a3f5258.png)
]

edupulpillo:
this problem exist even after a "refresh"

edupulpillo:
The problem persist in 17.0.2

edupulpillo:
nobody can reproduce the same problem?

sonikf:
Hi @edupulpillo 
It seems that this is a feature(for performance reasons) and not a bug.
You have to press the edit button for it to refresh.
https://github.com/Dolibarr/dolibarr/blob/57d832052e39f8884feebd885c42ab1895dfd122/htdocs/ecm/index.php#L191-L195

PR #31866, opened 2024-11-15
FIX : calculate the files count on ECM when add multiple files in one time

kkhelifa-opendsi:
Calculate the files count on ECM when add multiple files in one time #19493

image
image

### DIFF
--- a/htdocs/ecm/class/ecmdirectory.class.php
+++ b/htdocs/ecm/class/ecmdirectory.class.php
@@ -296,12 +296,37 @@ class EcmDirectory extends CommonObject
 	 */
 	public function changeNbOfFiles($value)
 	{
+		global $conf;
+
+		if ($value == '+') {
+			$relativepath = $conf->ecm->dir_output . '/' .$this->getRelativePath(); // Ex: dir1/dir2/dir3/
+			$relativepath = preg_replace('/^'.preg_quote(DOL_DATA_ROOT, '/').'/', '', $relativepath);
+			$relativepath = trim($relativepath, '/');
+
+			// Get nb file in relative path
+			$sql = "SELECT COUNT(*) AS nb";
+			$sql .= " FROM " . $this->db->prefix() . "ecm_files";
+			$sql .= " WHERE filepath = '" . $this->db->escape($relativepath) . "'";
+
+			dol_syslog(get_class($this) . "::changeNbOfFiles - Get nb file in relative path", LOG_DEBUG);
+			$resql = $this->db->query($sql);
+			if (!$resql) {
+				$this->error = "Error " . $this->db->lasterror();
+				return -1;
+			} else {
+				$value = 0;
+				if ($obj = $this->db->fetch_object($resql)) {
+					$value = (int)$obj->nb;
+				}
+			}
+		}
+
 		// Update request
 		$sql = "UPDATE ".MAIN_DB_PREFIX."ecm_directories SET";
 		if (preg_match('/[0-9]+/', $value)) {
 			$sql .= " cachenbofdoc = ".(int) $value;
 		} else {
-			$sql .= " cachenbofdoc = cachenbofdoc ".$value." 1";
+			$sql .= " cachenbofdoc = cachenbofdoc " . preg_replace('/[^\-]/', '', $value) . " 1";
 		}
 		$sql .= " WHERE rowid = ".((int) $this->id);
 
--- a/htdocs/ecm/class/ecmdirectory.class.php
+++ b/htdocs/ecm/class/ecmdirectory.class.php
@@ -316,7 +316,7 @@ class EcmDirectory extends CommonObject
 			} else {
 				$value = 0;
 				if ($obj = $this->db->fetch_object($resql)) {
-					$value = (int)$obj->nb;
+					$value = (int) $obj->nb;
 				}
 			}
 		}
--- a/htdocs/ecm/class/ecmdirectory.class.php
+++ b/htdocs/ecm/class/ecmdirectory.class.php
@@ -299,8 +299,8 @@ class EcmDirectory extends CommonObject
 		global $conf;
 
 		if ($value == '+') {
-			$relativepath = $conf->ecm->dir_output . '/' .$this->getRelativePath(); // Ex: dir1/dir2/dir3/
-			$relativepath = preg_replace('/^'.preg_quote(DOL_DATA_ROOT, '/').'/', '', $relativepath);
+			$relativepath = $conf->ecm->dir_output . '/' . $this->getRelativePath(); // Ex: dir1/dir2/dir3/
+			$relativepath = preg_replace('/^' . preg_quote(DOL_DATA_ROOT, '/') . '/', '', $relativepath);
 			$relativepath = trim($relativepath, '/');
 
 			// Get nb file in relative path


rycks (dismissed):
Hello,
ok for me as a fix bug it could be easyear to understand if we call  $ecmdir->changeNbOfFiles($res) into ecm/index.php (show line 138) ... but dol_add_file_process do not return number of success ...

so the quality fix could be against dol_add_file_process to return nb success and use that value to call changeNbOfFiles

rycks:
@eldy what do you think about that ?

a) that is a fix -> ok

b) better solution will be to change dol_add_file_process returned value

 * @return	int                             <=0 if KO, >0 if OK

return number of changed files in case of success then we could use that value into that PR

rycks:
> @eldy what do you think about that ?
> b) better solution will be to change dol_add_file_process returned value

see [ce0fe7103e2cf12d0a08175fa734a4104c454e11](https://github.com/Dolibarr/dolibarr/commit/ce0fe7103e2cf12d0a08175fa734a4104c454e11)

eldy:
> @eldy what do you think about that ?
> 
> a) that is a fix -> ok
> 
> b) better solution will be to change dol_add_file_process returned value
> 
> * @return	int                             <=0 if KO, >0 if OK
> 
> return number of changed files in case of success then we could use that value into that PR

According to inline documentation the parameter value must be the nb of doc to add.
`* @param string $value '+' or '-' or new number`
The number should be the quantity to store in database. So bug should be to fix the propagated parameter instead of doing a database SQL to recount.
So good solution is for me the one that has been suggested here:  [ce0fe7103e2cf12d0a08175fa734a4104c454e11] for this reason:
1-it fixes the problem at the source and its does not break behaviour (documentation of the  dol_add_file_process already says that "return is >0 if OK"). So i think we can go
2-fix match the doc of the expected value
3-a fix with the low printing in change is better than the one with more line code modified

i did not test, but i recommend to try the ce0fe7103e2cf12d0a08175fa734a4104c454e11 as fix for v18 instead of this version of PR.

lvessiller-opendsi:
@eldy @rycks 
We can change the method **changeNbOfFiles** like that to accept "**+NB of files**" or "**-Nb or files**" :
```
* 	@param	string	$value		'+' or '-' or new number or '-N' or '+N' (N is number to remove or add)
 *  @return int		         	Return integer <0 if KO, >0 if OK
 */
	public function changeNbOfFiles($value)
	{
		// Update request
		$sql = "UPDATE ".MAIN_DB_PREFIX."ecm_directories SET";
		if (preg_match('/^[\-\+][0-9]+$/', $value)) {
			$sql .= " cachenbofdoc = cachenbofdoc".$value;
		} elseif (preg_match('/[0-9]+/', $value)) {
			$sql .= " cachenbofdoc = ".(int) $value;
		} else {
			$sql .= " cachenbofdoc = cachenbofdoc ".preg_replace('/[^\-\+]/', '', $value)." 1";
		}
		$sql .= " WHERE rowid = ".((int) $this->id);

```

And int he file "htdocs/ecm/index.php" to add the "+nb" when we upload several files : 
$res = dol_add_file_process($upload_dir, $overwritefile, 1, 'userfile', '', null, '', $generatethumbs);
		if ($res > 0) {
			$result = $ecmdir->**changeNbOfFiles('+'.$res);**
		}


But how do you know if you overwrite some files ? (the nb of uploaded files is not the good number to add)
image

eldy:
> @eldy @rycks 
> We can change the method **changeNbOfFiles** like that to accept "**+NB of files**" or "**-Nb or files**" :
> ```
> * 	@param	string	$value		'+' or '-' or new number or '-N' or '+N' (N is number to remove or add)
>  *  @return int		         	Return integer <0 if KO, >0 if OK
>  */
> 	public function changeNbOfFiles($value)
> 	{
> 		// Update request
> 		$sql = "UPDATE ".MAIN_DB_PREFIX."ecm_directories SET";
> 		if (preg_match('/^[\-\+][0-9]+$/', $value)) {
> 			$sql .= " cachenbofdoc = cachenbofdoc".$value;
> 		} elseif (preg_match('/[0-9]+/', $value)) {
> 			$sql .= " cachenbofdoc = ".(int) $value;
> 		} else {
> 			$sql .= " cachenbofdoc = cachenbofdoc ".preg_replace('/[^\-\+]/', '', $value)." 1";
> 		}
> 		$sql .= " WHERE rowid = ".((int) $this->id);
> 
> ```
> 
> And int he file "htdocs/ecm/index.php" to add the "+nb" when we upload several files : 
> $res = dol_add_file_process($upload_dir, $overwritefile, 1, 'userfile', '', null, '', $generatethumbs);
> 		if ($res > 0) {
> 			$result = $ecmdir->**changeNbOfFiles('+'.$res);**
> 		}
> 
> 
> But how do you know if you overwrite some files ? (the nb of uploaded files is not the good number to add)
> image
> 

You found a point. We must know the exact number to update the qty and some files may erase other and some not.
So there is no other solution than "recalculate". This is what the PR tried to do by doing a sql select. 
However you must know that the information in database is just an "index" to store complementary information of a file on disk. It is not reliable. The reliable information can be obtained by as scan of the directory (for example when dir is updated by a webdav transfer, index is not immediatly updated but done later when Dolibarr detect database is not synch with files, another example is after restoring a disk of data, a ftp synch, etc...). So the re-count must be done on disk with dol_dir_scan and not with a select that is not reliable.
The rest of code is already doing dol_dir_scan and use the ecmfile table only for additionnal information.

lvessiller-opendsi:
@eldy @rycks 
Here this is a fix to determine what we see in GED and it's a good way to use SQL request and not to scan dicrectory physically for this action of uploading several files.
The "refresh" action should be implemented when you click on "image" button. And in this case, the directory will be scan with dol_dir_scan.
So it's ok for me to integrate this FIX on v18 .

lvessiller-opendsi (dismissed):

eldy:
> @eldy @rycks 
> Here this is a fix to determine what we see in GED and it's a good way to use SQL request and not to scan dicrectory physically for this action of uploading several files.
> The "refresh" action should be implemented when you click on "image" button. And in this case, the directory will be scan with dol_dir_scan.
> So it's ok for me to integrate this FIX on v18 .
> 

The value "+" like "-" is already implemented to make a +1 or -1, so i recommend to use another value to update with a scan of database, for example we can use value "database" as value for parameter to say we want to update nb with the count from database.

@kkhelifa-opendsi it seems you PR is pushed without permission to be edited so we have no way to fix this ourself (i just want to replace your new addition of the use of "+" into "database" to avoid conflict with old value and provide a easier way to understand how to use the parameters with a low risk of breaking compatibility or existing use. Can you fix this ?

rycks:
@kkhelifa-opendsi old PR ... but as @eldy says "you PR is pushed without permission to be edited so we have no way to fix this ourself (i just want to replace your new addition of the use of "+" into "database" to avoid conflict with old value and provide a easier way to understand how to use the parameters with a low risk of breaking compatibility or existing use. Can you fix this ?"

thanks

eldy:
@kkhelifa-opendsi can you replace the 
`if ($value == '+') {`
into
`if ($value == 'database') {`
The PR is not open to external contributions so github does not allow us to do it ourself.

rycks:
@eldy as @kkhelifa-opendsi pr is on a read-only are you ok if i merge that PR and just after merge make a new PR to replace "+" by "database" ?

eldy:
> @eldy as @kkhelifa-opendsi pr is on a read-only are you ok if i merge that PR and just after merge make a new PR to replace "+" by "database" ?
> 

Yes. When a pr remains too long with status pending and with no answer of submitter, if you know how to fix it yourself, you can do it (by editing the pr if pr is open, or by creating a new one if not).
And when you push or commit your own change, just include the #close idprtoreplace in your prnor commit label soit will close the old one and keep a link with. 
This is how i proceed on develop.

PR #31873, opened 2024-11-15
FIX : Default values feature is broken on units field on product create

jyhere:
This fixes the not working default values feature on units fields (weight_units, size_units, etc) when creating a new product

How to reproduce the bug :

Set a new default value (Setup > Default values/filters/sorting) with the following setup :
relative URL : product/card.php
field : weight_units
value : -6
Go to the new product creation page, Weight should be "mg" but is still "kg".

### DIFF
--- a/htdocs/product/card.php
+++ b/htdocs/product/card.php
@@ -1651,7 +1651,7 @@ if (is_object($objcanvas) && $objcanvas->displayCanvasExists($canvasdisplayactio
 					print '<tr><td>'.$langs->trans("Weight").'</td><td>';
 					print img_picto('', 'fa-balance-scale', 'class="pictofixedwidth"');
 					print '<input name="weight" size="4" value="'.GETPOST('weight').'">';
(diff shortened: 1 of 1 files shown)


htdocs/product/card.php
```diff
@@ -1662,29 +1662,29 @@
 					print '<input name="size" class="width50" value="'.GETPOST('size').'"> x ';
 					print '<input name="sizewidth" class="width50" value="'.GETPOST('sizewidth').'"> x ';
 					print '<input name="sizeheight" class="width50" value="'.GETPOST('sizeheight').'">';
-					print $formproduct->selectMeasuringUnits("size_units", "size", GETPOSTISSET('size_units') ? GETPOST('size_units', 'alpha') : '0', 0, 2);
+					print $formproduct->selectMeasuringUnits("size_units", "size", GETPOSTISSET('size_units') ? GETPOST('size_units', 'alpha') : (GETPOST('size_units', 'alpha') ?: '0'), 0, 2);
```
eldy, line 1665: To reach the GETPOST you added, we must have the GETPOSTISSET that is false. And if value is false, your GETPOST would always return empty. Did i liss something ?

erikvanberkum:
I am on version 20.0.2 when I do modify and save. I loose the default setting kg it becomes /null in the database and when the screen is reloaded the kg is gone. When i set the value to gram its all working fine just the 0 value is not. 
I have applied this patch but no improvement on this issue.

PR #31875, opened 2024-11-15
FIX FIX_Extrafields_type_Select_list_dependent_other_in_card #30801

### LINKED ISSUE #30801: Extrafields of type "Select list" which are dependent on other lists are not working
### Bug

Extarfield type “Select list” that are dependent on another list do not work

Bug was reported already in 2020 (#14736) 

### Dolibarr Version

12.0 till 19.0

### Environment PHP

7.2 till 8.2

### Environment Database

n/a

1. Create an extrafield with type "Select list" with some entries (see screenshot 1)
2. Create a second extrafield with type "Select list" with some entries which are dependent on the first list (see screenshot 2)
3. Edit the second extrafield - all entries are shown, but only entries of the parent list should be shown (see screenshot 3)

### Attached files

screenshot1
screenshot2
screenshot3

IC-Mathieu:
# FIX|Fix #30801
Extrafields of type "Select list" which are dependent on other lists are not working after object  creation


JamBalaya56562:
#39488 has been merged into develop (9b9e0889), so the problem reported here is fixed:

- `ExtraFields::showInputField()` now takes a `$filteronparentvalue` parameter, and `core/tpl/extrafields_view.tpl.php` passes it when a field is edited alone from the pencil, so a dependent *Select list* is filtered on the parent value already saved;
- the create and edit forms are unchanged and keep the javascript filtering;
- `test/phpunit/ExtraFieldsTest.php` was added to cover it.

Thanks @IC-Mathieu for reporting and investigating this one — your PR is credited in #39488. This PR can now be closed.

PR #31876, opened 2024-11-15
Fix #30766 Extra field Can Always Be Edited

### LINKED ISSUE #30766: Extra field “Can Always Be Edited” Setting Not Working as Expected
Screenshot 2024-08-26 at 5 11 27 PM
Screenshot 2024-08-26 at 5 11 03 PM
Screenshot 2024-08-26 at 5 10 10 PM
### Bug

I’m experiencing an issue with the “Can always be edited” setting for extra fields in Dolibarr. Despite this option being unchecked, I am still able to edit the values of these extra fields even when the status is not set to “Draft.” This behavior is inconsistent with the expected functionality, where the extra field should only be editable when the status is “Draft” and the “Can always be edited” option is unchecked.

### Dolibarr Version

19.0.2

### Environment PHP

_No response_

### Environment Database

Mysql

1. Go to the module where extra fields are used (e.g., Knowledge management).
2. Create an extra field and make sure the “Can always be edited” option is unchecked.
3. Create a record in the knowledge management and validate it
4. Try editing the extra field value without setting the record back to draft
5. its editable!

### Attached files

_No response_

Daoud-mohamed:
# Fix #30766 Extra field Can Always Be Edited

#30766


eldy:
Must use object $object instead of id.

Daoud-mohamed:
> Must use object $object instead of id.
@eldy Done!

PR #31877, opened 2024-11-15
FIX: All contacts were loaded even if no thirdparty was selected

noec764:
This bug is only problematic with a huge amount of contacts. 

It only append on an existing actioncomm with no contact or thirdparty attributed. When the user chose to modify it, all the contacts were loaded into the contact selector, resulting in a very slow-loading or even broken page.

I decided to do as it is now on actioncomm creation: the thirdparty needs to be selected before contact. 

This behavior can be deactivated with conf `MAIN_ACTIONCOM_CAN_ADD_ANY_CONTACT`

### DIFF
--- a/htdocs/comm/action/card.php
+++ b/htdocs/comm/action/card.php
@@ -2067,7 +2067,30 @@ if ($id > 0) {
 			// related contact
 			print '<tr><td>'.$langs->trans("ActionOnContact").'</td><td>';
 			print '<div class="maxwidth200onsmartphone">';
-			print img_picto('', 'contact', 'class="paddingrightonly"').$form->selectcontacts(!getDolGlobalString('MAIN_ACTIONCOM_CAN_ADD_ANY_CONTACT') ? $object->socid : 0, array_keys($object->socpeopleassigned), 'socpeopleassigned[]', 1, '', '', 1, 'minwidth300 widthcentpercentminusx', false, 0, 0, array(), 'multiple', 'contactid');
+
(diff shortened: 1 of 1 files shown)


htdocs/comm/action/card.php
```diff
@@ -2067,7 +2067,30 @@ function setdatefields()
 			// related contact
 			print '<tr><td>'.$langs->trans("ActionOnContact").'</td><td>';
 			print '<div class="maxwidth200onsmartphone">';
-			print img_picto('', 'contact', 'class="paddingrightonly"').$form->selectcontacts(!getDolGlobalString('MAIN_ACTIONCOM_CAN_ADD_ANY_CONTACT') ? $object->socid : 0, array_keys($object->socpeopleassigned), 'socpeopleassigned[]', 1, '', '', 1, 'minwidth300 widthcentpercentminusx', false, 0, 0, array(), 'multiple', 'contactid');
+
+			$searchSocid = $object->socid > 0 ? $object->socid : -1;
+			if ($searchSocid <= 0 && getDolGlobalString('MAIN_ACTIONCOM_CAN_ADD_ANY_CONTACT')) {
+				$searchSocid = 0;
+			}
+
```
hansemschnokeloch, line 2075: You have redondancy with your multiple if statements, you can simplify with
```php
$searchSocid = ($object->socid > 0) ? $object->socid : (getDolGlobalString('MAIN_ACTIONCOM_CAN_ADD_ANY_CONTACT') ? 0 : -1);
```

Issue #26718, opened 2023-11-18
API GET ORDER sqlfilter seems to be ignored or not working


JonBendtsen:
### Bug

It seems to me like when I use the API to get selected orders, then the sqlfilter option is either ignored or plain out isn't working.

### Environment Version

18.0.1

### Environment OS

upshift dolibarr docker image 18.0.1

### Environment Web server

upshift dolibarr docker image 18.0.1

### Environment PHP

PHP 8.2.10 (cli) (built: Aug 31 2023 20:34:52) (NTS)

### Environment Database

1:11.0.3+maria~ubu2204

### Environment URL(s)

/dolibarr/api/index.php/orders

### Expected and actual behavior

Expected behaviour is that I would expect the SQL filter to work
Actual behaviour is that the SQL filter is either ignored or is not working.

(t.date_creation:>:1699724452) is ignored, I get all 3 orders this thirdparty have made, and I should get 0. If I use the value of 1697094397 I still get all 3 orders and I should get 1. 
(t.date_creation:=:1697094397) is not working, I get no orders this thirparty have made, and I should get exactly one.
(t.date_creation:<:1697094397) is not working, I get no orders this thirparty have made, and I should get exactly one.

### Steps to reproduce the behavior

I have a thirdparty with 3 orders. I use that thirdparty's id to limit the search to only that thirdparty. 

This thirdparty have 3 orders with date_creation as:  1696767386	1697094397	1699724452

### Attached files

![order_sqlfilter_bigger](https://github.com/Dolibarr/dolibarr/assets/2551481/f8e6658f-27b0-4ce5-b07c-02386039abed)
![order_sqlfilter_equal](https://github.com/Dolibarr/dolibarr/assets/2551481/f4e0e329-42ba-48bd-8cc0-937c8d827b01)
![order_sqlfilter_less](https://github.com/Dolibarr/dolibarr/assets/2551481/801388e9-7737-4d25-a728-8c7af61dc59a)

hansemschnokeloch:
Did you try following syntax with sqlfilters ? 

`(t.date_creation:>:'2023-11-22')`

I'm not sure timestamp date format is the right syntax to use here.

JonBendtsen:
> `(t.date_creation:>:'2023-11-22')`
> 
> I'm not sure timestamp date format is the right syntax to use here.

This is more correct. I just fed the same value as in the json into the format.

I can now use:
(t.date_creation:>:'2024-02-21') and get 0 orders
[...]

PR #31903, opened 2024-11-15
NEW : Filter API on DATE(t.field) if  equal 'YYYY-MM-DD'

kkhelifa-opendsi:
Filter API on DATE(t.field) if  equal 'YYYY-MM-DD' #26718

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -14016,6 +14016,11 @@ function dolForgeCriteriaCallback($matches)
 		}
 	}
 
+	// Filter on DATE(t.field) if $tmpescaped equal 'YYYY-MM-DD'
+	if (strpos($operand, 'date') !== false && preg_match("/^'\d{4}\-\d{2}\-\d{2}'\$/", $tmpescaped)) {
+		$operand = 'DATE('.$operand.')';
+	}
+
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -14016,6 +14016,11 @@ function dolForgeCriteriaCallback($matches)
 		}
 	}
 
+	// Filter on DATE(t.field) if $tmpescaped equal 'YYYY-MM-DD'
+	if (strpos($operand, 'date') !== false && preg_match("/^'\d{4}\-\d{2}\-\d{2}'\$/", $tmpescaped)) {
+		$operand = 'DATE('.$operand.')';
```
eldy, line 14021: Use of DATE function is forbidden. See https://wiki.dolibarr.org/index.php?title=Language_and_development_rules#SQL_Coding_rules

Try to use sqlDateFilter() instead.

JonBendtsen: @kkhelifa-opendsi  I wish this PR would get updated

JonBendtsen:
I think this PR can be closed because Apparently one can just use both a > and a < and then you only get on that date

(t.date_creation:>:'2025-08-11' and t.date_creation:<:'2025-08-12')

which works fine :-)

PR #31910, opened 2024-11-15
FIX: ajax tooltip not closing if mouse leave the element before ajax response

jyhere:
Fix the annoying ajax tooltip bug (tooltip not closing if moving the mouse too quickly over a .classfortooltip element )

Also, javascript code is now "human readable"... :D


eldy:
As this is a bug fix, can you repush the PR with only the MINIMAL change required to fix the bug ? Refactoring or code style change must be excluded from a bug fix PR (or must be into another dedicated PR for code style change).

eldy:
Done by frederic24

PR #31916, opened 2024-11-15
FIX Takepos barcode rule with float qty

lvessiller-opendsi:
FIX Takepos barcode rule with float qty


htdocs/takepos/ajax/ajax.php
```diff
@@ -220,7 +220,7 @@
 							if (isset($barcode_value_list['qd'])) {
 								$qty_str .= '.' . $barcode_value_list['qd'];
 							}
-							$qty = (float) $qty_str;
+							$qty = floatval($qty_str);
```
eldy, line 223: What is difference ?

lvessiller-opendsi: It's just a mistake :)

htdocs/takepos/index.php
```diff
@@ -817,7 +817,7 @@ function Search2(keyCodeForEnter, moreorless) {
 						console.log("There is only 1 answer with barcode matching the search, so we change the thirdparty "+data[0]['rowid']);
 						ChangeThirdparty(data[0]['rowid']);
 					}
-					else if ($('#search').val() == data[0]['barcode'] && 'product' == data[0]['object']) {
+					else if ('product' == data[0]['object']) {
```
eldy, line 820: It seems the old test was wrong.
According to rhe coment, the old test was doing:
- if we made a search ok a string and result is a product with the barcode searched, we know we did a search to find 1 particular product, in goalntobadd it to basket, so we add it automatically.
- but if we made a search, but not on a barcode value, for example a search on "abc" to see all products that contains abc, we show the result but we do not add the product automatically in basket, even if there is 1 answer ok only (because we want to see answers of search and not necessarly select it, it may depends on result)

And this is the behaviour that we want.
If you change this test, this behaviour will be broken, and if you try to search a product just to see what can be answer on the search string, product will be also added with no chance to check result before.

Am i right ?

lvessiller-opendsi: You're right and I understand your point of view.
However, in the case of you have a bar-code rule and some products with the reference of the product in the 8 first digits, the full bar-code is not equal to the product reference and the 5 others digits are here to determine the quantity and the product must be selected automatically with this quantity.
I believed that if there was only one product found, it will be automatically inserted but I see it not necessary the case (it was the default behavour in Dolibarr v14)
Maybe I have to use another condition for bar-code rule case ?

ems-co:
Thank you so much! now the barcode rule is finally working! 
i tried it with barcode rule (ref:5+qu:2+qd:2+other:0)

and product reference was 12345 (without barcode in the product card)
then i added the product 123452522 to the POS and it added it automatically as expected

Screenshot 2024-11-16 at 5 16 49 PM

@eldy if we specify the Barcode rule and the TakePOS does not add the quantity and quantity decimal to the POS card automatically then what will be the use of the Barcode rule to insert product? 

Screenshot 2024-11-16 at 5 22 21 PM

lvessiller-opendsi:
FIX Issue https://github.com/Dolibarr/dolibarr/issues/30203

lvessiller-opendsi:
@eldy 
I keep last behavour and condition of barcode when only one product is found.
I only change the result of barcode found (in ajax) when the barcode rule is correct and there is only one product found with this barcode rule.
The CI errors are on unchanged files and not concern this PR.

Issue #31880, opened 2024-11-15
Serious error in ver 20*


ghonztrail:
### Bug

purchase order reception ignore qty if it in decimal only, i tested this error in fresh installation for version 20.0.1 and 20.0.2 and found both version have the same error which not exist in version 19* which work prefect 
i attached 5 screen shots for the order qty and receiving screen one by one, you will find that the product 3 with qty 0.444 disappeared in final reception screen 

### Dolibarr Version

20.0.2

### Environment PHP

PHP 8.1.2-1ubuntu2.19

### Environment Database

mariadb  Ver 15.1 Distrib 10.6.18-MariaDB, for debian-linux-gnu (x86_64) using  EditLine wrapper

### Steps to reproduce the behavior and expected behavior

1- create purchase order with item qty in only decimal like 0.700 
2- complete the reception steps 
3- you will find empty reception to validate in the end of reception process 

### Attached files

<img width="1173" alt="Order qty" src="https://github.com/user-attachments/assets/2708284c-6ca2-412d-b570-a4c1b03a7963">
<img width="1173" alt="create reception screen 1" src="https://github.com/user-attachments/assets/96bcacf0-e1d2-4cec-b5ed-65c75e8f8c63">
<img width="1173" alt="create reception screen 2" src="https://github.com/user-attachments/assets/f1c0aacb-ddf5-4529-b895-bb1bd38ef823">
<img width="1173" alt="validate reception" src="https://github.com/user-attachments/assets/286052e3-1f29-4532-99ef-693922acec02">
<img width="1173" alt="Reception Validated" src="https://github.com/user-attachments/assets/bcbdd4e1-5dfe-4f8d-82dc-68140c52015e">

PR #31920, opened 2024-11-15
Fix reception lines when qte is strictly between 0 and 1

lemathou:
Fix Issue https://github.com/Dolibarr/dolibarr/issues/31880


lemathou:
Duplicate... I was not enough fast !

PR #31922, opened 2024-11-15
FIX Status not correct in public ticket list

fappels:


htdocs/public/ticket/list.php
```diff
@@ -702,7 +702,10 @@
 					// Statut
 					if (!empty($arrayfields['t.fk_statut']['checked'])) {
 						print '<td class="nowraponall">';
-						$object->fk_statut = $obj->fk_statut;
+						$object->status = $obj->fk_statut;
+						if (getDolGlobalString('TICKET_SHOW_PROGRESSION')) {
```
eldy, line 706: Why not having the set all the time ?

fappels: To avoid missing object property exception because progress not set in $obj if show progress to public disabled.

PR #31928, opened 2024-11-15
FIX send email to assigned user on ticket create

ghost:
# FIX send email to assigned user on ticket create
Fixed send email to assigned user. Refactored some code into new method `composeAndSendAssigneeMessage`
Code inspired from PR #25866


eldy:
Does it replace https://github.com/Dolibarr/dolibarr/pull/25866 ?

ghost:
> Does it replace #25866 ?

Yes exactly ! Credit to @BenjaminFlr 
Thanks eldy

PR #31929, opened 2024-11-15
fix PR 31910

frederic34:
merge first https://github.com/Dolibarr/dolibarr/pull/31910


eldy:
As this is a bug fix, can you repush the PR with only the minimal change required to fix the bug ? Refactoring or code style change must be excluded from a bug fix PR (or must be into another dedicated PR for code style change).

frederic34:
https://github.com/Dolibarr/dolibarr/pull/31933

PR #31931, opened 2024-11-15
test github token

frederic34:
@eldy I reactivate action on pull request
this action calculate how many errors are ignored, if the number decrease it update the baseline

### DIFF
--- a/.github/workflows/phpstan_baseline.yml
+++ b/.github/workflows/phpstan_baseline.yml
@@ -20,6 +20,7 @@ permissions: {} # none
 
 env:
   PHP_VERSION: '8.2'
+  GH_TOKEN: ${{ github.token }}
   gh_event: ${{ inputs.gh_event || github.event_name }}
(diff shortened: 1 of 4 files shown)


frederic34:
image

frederic34:
if it works, you can also run the workflow manually to clean the phpstan baseline

frederic34:
action to clean caches is also in error
image

image

eldy:
> image
> 

Why having the pho baseline run at each pr ? Why not doing it only once a week or day to avoid overload of github action on our repo ?

eldy:
> image
> 

Why having the pho baseline run at each pr ? Why not doing it only once a week or day to avoid overload of github action on our repo ?

frederic34:
@eldy this task is also sheduled, here you can see it don't work https://github.com/Dolibarr/dolibarr/actions/runs/11831647690/job/32967144876
ok to not run on pull_request, i have no other way to test

PR #31946, opened 2024-11-16
Fix SQL missing constraint in user::load_previous_next_ref() with MULTICOMPANY_TRANSVERSE_MODE

lemathou:
Fix bad SQL query in user::load_previous_next_ref().

A cartesian product is done, without constraint, I replaced by a INNER JOIN with integrated constraint in ON.
Without fix, it can take a took a long time to display informations when we have a lot of users/groups.


hregis:
@lemathou why change the 13 branch ??

lemathou:
Because a customer (and Dolibarr partner : Code42) needs a fix, he has a database with thousands of user, still in v13, and the user card takes seconds to display.

PR #31949, opened 2024-11-16
New: fields for expedition

nicolas-eoxia:


eldy:
@frederic34 @mdeweerd Can you help me to find what's wrong with this error phan and phpstan. What is the clue to find ?

mdeweerd:
The phpstan report should show which field value is not matching the specification.
Phan does not make a suggestion.

With phpstan, the full detailed log will be more "verbose".
I am working on generating the report locally (I do not get a notice for this with my local setup).

mdeweerd:
~I did not reproduce it, but visual inscpection indicates phpdoc is missing the 'css' field.~

~This is to be used instead:~
```php
       * @var array<string,array{type:string,label:string,enabled:int<0,2>|string,position:int,notnull?:int,visible:int<-2,5>|string,noteditable?:int<0,1>,default?:string,index?:int,foreignkey?:string,searchall?:int<0,1>,isameasure?:int<0,1>,css?:string,csslist?:string,help?:string,showoncombobox?:int<0,2>,disabled?:int<0,1>,arrayofkeyval?:array<int|string,string>,comment?:string,validate?:int<0,1>}>  Array with all fields and their property. Do not use it as a static var. It may be modified by constructor.
```

EDIT: css is present in the phpdoc, my grep was wrong.

mdeweerd:
Got it: "Offset 'default' (string) does not accept type int."

mdeweerd:
The easiest way to solve this is to make the values for 'default' strings - these values are used to build SQL expressions, considered strings, possibly sanitized and casted.   Changing the PHPDoc requires to update them for all $fields.  

https://github.com/Dolibarr/dolibarr/blob/b2bad15a046d13594efd463fcf10fe899e2eb736/htdocs/core/db/mysqli.class.php#L868-L876

https://github.com/Dolibarr/dolibarr/blob/b2bad15a046d13594efd463fcf10fe899e2eb736/htdocs/core/db/mysqli.class.php#L1027-L1033

htdocs/expedition/class/expedition.class.php
```diff
@@ -84,11 +84,84 @@ class Expedition extends CommonObject
...
+		'note_private'       => ['type' => 'text',         'label' => 'NotePrivate',    'enabled' => 1,                          'position' => 270, 'notnull' => 0, 'visible' => -1],
+		'note_public'        => ['type' => 'text',         'label' => 'NotePublic',     'enabled' => 1,                          'position' => 280, 'notnull' => 0, 'visible' => -1],
+		'model_pdf'          => ['type' => 'varchar(255)', 'label' => 'PDFTemplate',    'enabled' => 1,                          'position' => 290, 'notnull' => 0, 'visible' => -1],
+		'last_main_doc'      => ['type' => 'varchar(255)', 'label' => 'LastMainDoc',    'enabled' => 1,                          'position' => 300, 'notnull' => 0, 'visible' => -1],
+		'fk_incoterms'       => ['type' => 'integer',      'label' => 'IncotermCode',   'enabled' => '$conf->incoterm->enabled', 'position' => 310, 'notnull' => 0, 'visible' => -1],
+		'location_incoterms' => ['type' => 'varchar(255)', 'label' => 'IncotermLabel',  'enabled' => '$conf->incoterm->enabled', 'position' => 320, 'notnull' => 0, 'visible' => -1],
```
frederic34, line 161: make use of isModEnabled instead $conf->incoterm->enabled

frederic34: and with the fix suggested by @mdeweerd 
image

Issue #31950, opened 2024-11-16
Bug on ticket stats


nicolas-eoxia:
### Bug

Stats are wrong due to JOIN on : 
c_ticket_type
c_ticket_severity
c_ticket_category

### Dolibarr Version

19.0.4 - 21.0.0-beta

### Environment PHP

7.4.33

### Attached files

![image](https://github.com/user-attachments/assets/6cff7a03-787d-4c76-a1e8-f034f7317eeb)

PR #31951, opened 2024-11-16
#31950 [Ticket] fix: unwanted join for sql request ticket stats

nicolas-eoxia:


htdocs/public/ticket/list.php
```diff
@@ -374,18 +374,14 @@
 		$sql .= " t.datec,";
 		$sql .= " t.date_read,";
 		$sql .= " t.date_close,";
-		$sql .= " t.tms,";
-		$sql .= " type.label as type_label, category.label as category_label, severity.label as severity_label";
```
eldy, line 378: Line 638,we have 
$obj->type_label
If you remove this, this property will be empty...

PR #31952, opened 2024-11-16
Fix:hook for adding params for payments/CloseBill action

bafbes:
Fix:hook for adding params for payments/CloseBill action


bafbes:
closed because of bugs/errors

PR #31953, opened 2024-11-16
feat: add entity in custom translation with multicompany

benoitpascal:
Allow setting entity to apply custom translation when multicompany is enabled.

### View
image

### Edit
image

## Without Multicompany
image


hregis:
@benoitpascal great but a simple drop-down list with 2 values is better :

0 => all entities
1 => current entity (selected by default)

if "1" you use conf->entity and if "0" you use 0 value

PR #31965, opened 2024-11-17
19.0 fix missing hook for ticket message form

samifilali:


eldy:
Why fix ?
A not implemented feature is not a fix if the feature was never announced into the Changelog. Thanks to push in develop.

PR #31976, opened 2024-11-18
CHANGE : reformat lib_foot.js code to be more readable by IDEs

jyhere:


htdocs/core/js/lib_foot.js.php
```diff
@@ -101,320 +98,313 @@
...
-							  //$(".dropdown dt a").parents(\'form:first\').submit();
-							  //$(".dropdown dt a").closest("form").submit();
-							  lastopendropdown.closest("form").submit();
-					      }
-					  }
-                  });
-              });
-           ';
-}
+<?php if (!defined('JS_JQUERY_DISABLE_DROPDOWN')) : ?>
```
eldy, line 154: The use of 
```
if ... :
endif
```
 is not allowed by the php syntax rules, must follow the PSR-12 rules (section 5.1) so the { and }

jyhere: This notation is used when inserting PHP code into HTML / JS files, so I'm not sure that this PSR-12 recommendation applies in this case. Using classic notation in js / html files makes them less readable, whereas we're looking for a shortened “template engine” type of writing in this case.

jyhere: Also, the notation <?= ?> is apparently forbidden in the Dolibarr code, but this notation is PSR-12 friendly, and is always available (independant from short_open_tag : https://www.php.net/manual/en/ini.core.php#ini.short-open-tag). It should be authorized, so we can work on real html/php template-like files

PR #31979, opened 2024-11-18
NEW constants EMAIL

leninrivas:
NEW __MYCOMPANY_TVAINTRA__
NEW __AMOUNT_MULTICURRENCY_FORMATED__
NEW  function dol_set_value($selector, $value)
NEW rounding MTL


htdocs/core/lib/functions.lib.php
```diff
@@ -6822,6 +6823,8 @@ function price($amount, $form = 0, $outlangs = '', $trunc = 1, $rounding = -1, $
 			$nbdecimal = getDolGlobalInt('MAIN_MAX_DECIMALS_UNIT');
 		} elseif ($forcerounding === 'MT') {
 			$nbdecimal = getDolGlobalInt('MAIN_MAX_DECIMALS_TOT');
+		} elseif ($forcerounding === 'MTL') {
+			$nbdecimal = getDolGlobalInt('MAIN_MAX_DECIMALS_TOT_LINE'); // TODO Future feature  to improve accuracy
```
eldy, line 6827: Total of lines are used in accountancy to calculate the total of document. For legal reason, total of lines must match total on document. 
For this reason the accuracy of total of lines must always be the same than the accuracy of total of documents, so nb of decimals for total of lines must use MAIN_MAX_DECIMALS_TOT, otherwise, accountancy will always be wrong and total of lines will never match amount in document.

leninrivas: I understand, so is there a way to improve the precision perhaps by increasing the decimals of the unit prices to 10 or 12?

PR #31986, opened 2024-11-18
Update oldcopy in task

### LINKED ISSUE #31985: oldcopy not set
### Bug

In all objects, there's a property "oldcopy" supposed to be object's clone before its modification.
But oldcopy is nearly never correctly set (see [issue #25956 ](https://github.com/Dolibarr/dolibarr/issues/25956), closed by bot

Yet It would be very usefull : when using update trigger on an object, we can't know if a particular property has changed or not !
I'm working on such a use case : we wan't to send a mail when start date of a task has changed. I cannot achieve this with a trigger till I have no solution to retrieve the old value

### Dolibarr Version

all

### Environment PHP

all

### Environment Database

all

_No response_

### Attached files

_No response_

vmaury:
FIX #31985 oldcopy not set

Example of updating oldcopy when fetching task object.

It would be very usefull to update oldcopy for all objects

### DIFF
--- a/htdocs/projet/class/task.class.php
+++ b/htdocs/projet/class/task.class.php
@@ -587,6 +587,7 @@ class Task extends CommonObjectLine
 			$this->db->free($resql);
 
 			if ($num_rows) {
+				$this->oldcopy = clone $this;
 				return 1;
 			} else {
 				return 0;


hregis:
@vmaury and what will happen if you have several triggers or hooks which modify the object and which do a fetch!! the original oldcopy will no longer be valid!! the oldcopy can't be in fetch function !!

vmaury:
@see last comments in #31985

PR #31988, opened 2024-11-18
fix phpstan

frederic34:


frederic34:
```
$permissiontoadd = $user->hasRight('accounting', 'chartofaccount');

if (!$permissiontoadd) {
	accessforbidden();
}
// now $permissiontoadd or $user->hasRight('accounting', 'chartofaccount') are always equal to 1
```

after the `if (!$permissiontoadd) {` the only value possible is 1

PR #32017, opened 2024-11-20
Fix not third,email reminder error

Expresion:


htdocs/comm/action/class/actioncomm.class.php
```diff
@@ -2702,7 +2702,6 @@ public function sendEmailsReminder()
 
 					// Load event
 					$res = $this->fetch($actionCommReminder->fk_actioncomm);
-					if ($res > 0) $res = $this->fetch_thirdparty();
```
eldy, line 2705: Why removing this call ?
This allow to load the ->thirdpardy object that can be used later into trigger or getCommonSubstituionArray().

Expresion: Because if the third party is not in the event, the fetch fails.
To reproduce the error:
Create an event without a third party and run a send Emails Reminder cron job.

eldy: If the event has no thirdparty, then the fetch_tirdparty will return 0.
The pb is on test > 0

I tried to push a better fix on 0d07520e02c8156834093e2a9fdce45198b2f7ae.
Thanks to check.

PR #32018, opened 2024-11-20
FIX: use tax with code on supplier order line give tax code missing in supplier invoice

FHenry:
Actually

image
image

Idem on create invoice from supplier order list

I report code use in customer invoice card
https://github.com/Dolibarr/dolibarr/blob/ac6c57e90e129bed83ca032165f0100a85f61f66/htdocs/compta/facture/card.php#L1740


rycks (approved):
ok for me

lvessiller-opendsi (approved):

PR #32019, opened 2024-11-20
set object entity at create

frederic34:
if you create a supplier invoice from a supplier order, object entity is not set (but saved in db) and if process continue, by example it can't find `$conf->mycompany->multidir_output[$object->entity]`


hregis:
@frederic34 this check is already in "setEntity()" function :` if (empty($this->entity))`
and you broken other functionnality of "setEntity()" function of multicompany module

frederic34:
can you explain the problem ?

hregis:
@frederic34 the "setEntity()" function inside Multicompany module check if the element is sharing for avoid have a wrong value in $this->entity, 
eg if we are in entity 1 and you create an object with "$this->entity = 2" "setEntity()" check if the element is sharing, if not the function return value 1 instead 2. 
if you add `if (empty($this->entity))` you bypass this check
you see ?

eldy:
> @frederic34 the "setEntity()" function inside Multicompany module check if the element is sharing for avoid have a wrong value in $this->entity, eg if we are in entity 1 and you create an object with "$this->entity = 2" "setEntity()" check if the element is sharing, if not the function return value 1 instead 2. if you add `if (empty($this->entity))` you bypass this check you see ?

To complete, the setEntity has rules managed by multicompany than can set/overwrite the entity in creation to a different value than the one into this->entity. So the test if (empty($this->entity)) remove this feature.
I think a solution is to do
`$this->entity = setEntity($this); `
all the time. If there is no overwrite to do, this will do $this->entity = $this->entity so not changing anything. And if ->entity not defined, then the setEntity will return a default value.

frederic34:
I think I understood, we always ignore object->entity set before calling create()

hregis:
@frederic34 $object->entity is not ignore, if exist, setEntity check if the value is authorized

frederic34:
in any way we must do the setEntity thing even entity is set or not

hregis:
@frederic34 yes if the entity is not defined, setEntity return the current entity, and if entity is defined, the function check if the element is sharing for this entity value

PR #32028, opened 2024-11-21
fix: buyer and sellers in supplier invoice and supplier order list are not corectly defined

FHenry:
In supplier invoice addline method 
we use wrongly  $this->thirdparty as buyer insted of $mysoc
and in 
calcul_price_total in case of create supplier invoice from supplier order list $this->thirdparty is never defined

### DIFF
--- a/htdocs/fourn/class/fournisseur.facture.class.php
+++ b/htdocs/fourn/class/fournisseur.facture.class.php
@@ -2144,7 +2144,7 @@ class FactureFournisseur extends CommonInvoice
 				$pu = 0;
 			}
 
-			$localtaxes_type = getLocalTaxesFromRate($txtva, 0, $mysoc, $mysoc);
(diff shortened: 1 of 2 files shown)


htdocs/fourn/class/fournisseur.facture.class.php
```diff
@@ -2144,7 +2144,7 @@ public function addline($desc, $pu, $txtva, $txlocaltax1, $txlocaltax2, $qty, $f
 				$pu = 0;
 			}
 
-			$localtaxes_type = getLocalTaxesFromRate($txtva, 0, $mysoc, $this->thirdparty);
+			$localtaxes_type = getLocalTaxesFromRate($txtva, 0, $mysoc, $mysoc);
```
eldy, line 2147: what is wrong with old code ?
Param 3 must be the buyer, and param 4 must be the seller (so the thirdparty).

FHenry: Yes, I was wrong, I miss read $myssoc as first parameters

PR #32041, opened 2024-11-21
Update tax.lib.php

le-reparateur:
When expensereport for a month and for a tax rate contains more than 1 line (n lines), the tax report contains n identical lines and total are impacted with those n values.
In this exemple ER2409-0045 contains 7 different payments (2 payments with 20% tva, 2 payments with 10% tva and 3 payments with 5.5% tva -> the picture).
The lines are copied 7 times for the same value in the report.

Capture d'écran 2024-11-21 152725

Add GROUP BY in SELECT sql request solve the problem.

Francis

### DIFF
--- a/htdocs/core/lib/tax.lib.php
+++ b/htdocs/core/lib/tax.lib.php
@@ -1177,6 +1177,7 @@ function tax_by_rate($type, $db, $y, $q, $date_start, $date_end, $modetax, $dire
 		if (!empty($conf->global->MAIN_NOT_INCLUDE_ZERO_VAT_IN_REPORTS)) {
 			$sql .= " AND (d.".$f_rate." <> 0 OR d.total_tva <> 0)";
 		}
+		$sql .= " GROUP BY d.rowid";
 		$sql .= " ORDER BY e.rowid";
 
 		if (!$sql) {


eldy:
What is the url or menu to use to reach the page with the screenshot ?

le-reparateur:
url : .../htdocs/compta/tva/quadri_detail.php
Then left menu -> month report -> select a period -> select a month -> click on a rate in supplier list to obtain all lines...
Capture du 2024-11-21 23-24-44

eldy:
This does not look to be a bug: The goal of report is to show the detail of vat for each payment part. So if there is 7 payments, you should have 7 lines, each lines 

For example, a report for an expense report with 2 payments show the detail of vat part of each couple expensereport line/payment. On v20, we see the amount of expense report (8.33 euros) but also the amount of payment and the part of vat for this related payment.
image

for example on last line, amount if 8.33 HT (so 10 TTC), the payment is 1 euros, so 10% of total, so the part paid for this line is 0.833 and the VAT part among this part paid is 0.17. Payment may have been done at different date, so we have no other choice to calculate using a detail for each payment.
A think that could be interesting to avoid confusion is to add the date of payment (with hour min sec) so we can see that lines are not duplicates but a detailed list

le-reparateur:
Problem is not in that way... sorry !

If you have an expense report (for august a real one!) with 8 different lines (different kind of expenses) and 7 paiements.

Capture d'écran 2024-11-22 162000

The 8 lines of expense have different tax rate : 3 for 20%, 2 for 10% and 3 for 5.5%.

When we look at the tax report for august, then we have 

Capture d'écran 2024-11-22 162250

For the rate 20% we can see 7 identical lines for the 3 expenses realised at a rate of 20%

The same for 10% 2 lines repeated 7 times each...

Capture d'écran 2024-11-22 162314

And for 5.5% 3 lines repeated 7 times.

Capture d'écran 2024-11-22 162353

And for each total corresponding to a rate, the result is bad. And of course total of total !

So with my proposal, same screnshots give 

Capture d'écran 2024-11-22 162605

Capture d'écran 2024-11-22 162623

Capture d'écran 2024-11-22 162637

And totals are good and my VAT declaration too !
Francis

eldy:
I think the bug is not in duplicate.
If we have 7 payments on an expense report with 8 lines, we should get 56 lines in the detail.
The pb here seems in total. The first colum of amount in your screenshot is the total of line so because we show detail of vat payment part for each line, leach line is outpu 7 times, so we must not make sum on this column, or we will count 7 times the same. Sum must be visible only for the column of the part that is the detail for couple expense line/paylent. It seems this is correctly done on v20.so i suggest to get a fix on this branch to try a backport of code in v20 branch instead.

le-reparateur:
On a specific rate we have the list of all lines (supplier invoices, expense report) which impact the total of VAT for that rate.
We need the exact total of that sum to declare the VAT. I do that way because I don't use accounting part for that details.
For my use, I solve the problem this way. I check that this function is only used for the vat report, no other impact on my Dolibarr.
In the near future I will test that in my V20 ... I have now included that point in my checklist !
I can also test that point on a V19... I quickly have a look on V18 and it seems we have the same problem...

le-reparateur:
Hi
After testing my V20, I confirm we have the same problem, wich I solved with the same solution.

eldy:
> If we have 7 payments on an expense report with 8 lines, we should get 56 lines in the detail.

Can you enter the case that show the error on this test instance:   
https://preview.dolibarr.org   (admin / admin/admin)

Remind, if we have 7 payments on an expense report with 8 lines, the expected result is to have 56 lines in the line of detail. 
But sum must be visible only for the column of the part that is the detail for couple expense line/paylent

le-reparateur:
Hi
I am disagree with 7 payments with 8 lines give 56 lines in details report !

First payment cover a product n°1 with 20% tax and another product n°2 with 5.5% tax.
Second payment cover a product n°3 with 5.5% tax.
Then you have 2 payments BUT you must have in the report only 3 lines : one for product n°1 for 20% rate et 2 lines for other products for 5.5% rate.

I will test in the instance.

le-reparateur:
Impossible to connect to instance...
Sorry !

eldy:
> Hi
> I am disagree with 7 payments with 8 lines give 56 lines in details report !
> 
> First payment cover a product n°1 with 20% tax and another product n°2 with 5.5% tax.
> Second payment cover a product n°3 with 5.5% tax.
> Then you have 2 payments BUT you must have in the report only 3 lines : one for product n°1 for 20% rate et 2 lines for other products for 5.5% rate.
> 
> I will test in the instance.
> 

You disagree because you try/expect to use the detail kines to get a report of all payments. But it was not the goal of this feature. Goal was to see the repartition of vat for each couple (line of expense report / line if payment) so to see the vat part for the 56 possible couples.
Another feature or screen is surely necessary for your need to get a more summarized report.
Current detail works as expected but i understand that you would prefer an intermediate more condensed report.

le-reparateur:
Hi
Sorry i don't anderstand your point of view.
Anyway, in the reports VAT totals are wrong, even if nobody use them !
If you have time I can open a user for you on my Dolibarr to see the problem during a rapid phone call...

PR #32047, opened 2024-11-21
FIX: backport - flip supplier proposal and orders to show up as negative in the project Profit overview

JonBendtsen:
FIX: backport - flip supplier proposal and orders to show up as negative in the project Profit overview
In the project overview, supplier proposals and orders was shown as positive, where as the supplier invoices was shown as negative. This backport MR fixes the presentation so they show up as negative.


eldy:
As you did a PR for v18, it is not necessary to do a PR for v19, nor v20, nor develop.
Any PR merged into a branch n is also few days later automatically merged into n+1, n+2...until develop

Issue #32056, opened 2024-11-22
[DATAMODEL] - Correction for time unit in database - minutes is "min"


lmag:
### Feature Request

![image](https://github.com/user-attachments/assets/8f9f6415-7757-41b0-bc02-991106569019)

https://github.com/Dolibarr/dolibarr/issues/31740

PR #32058, opened 2024-11-22
fix: #32056

FHenry:
rename unit short label as in #32056


htdocs/core/lib/date.lib.php
```diff
@@ -130,7 +130,7 @@ function dol_time_plus_duree($time, $duration_value, $duration_unit, $ruleforend
 	if ($duration_unit == 's') {
 		return $time + (int) ($duration_value);
 	}
-	if ($duration_unit == 'i' || $duration_unit == 'mn' || $duration_unit == 'min') {
+	if ($duration_unit == 'i' || $duration_unit == 'min') {
```
eldy, line 133: For compatibility with external modules, we should keep the compatiblity test
$duration_unit == 'mn'
too

FHenry: There is update in 20.0.0-21.0.0.sql that change the code, the code "mn" is from 20.0

Issue #32057, opened 2024-11-22
Standardization of time units in the bom - Delete years and month


lmag:
### Bug

Year and Month are not time unit

![image](https://github.com/user-attachments/assets/2280654a-f4f3-4bea-bfe8-d0612467fc45)

### Dolibarr Version

20.0.0

https://github.com/Dolibarr/dolibarr/issues/31740

PR #32059, opened 2024-11-22
fix: #32057

FHenry:


FHenry:
May be to return to Draft state on this PR , I can't do it

eldy:
I suggest to close and reopen. Did not find either how to go back on draft.

mdeweerd:
To go back to draft - in the right sidebar of my github ui:
{9B983D67-ACA8-48D6-94DE-8958FA2B84D1}

FHenry:
@mdeweerd Thank's I was Sure it was possible I just forget where and a little link like taht is not easy to find

PR #32062, opened 2024-11-22
Fix : load data in extrafields

ATM-NicolasV:
When  we call gettasksArray, by default loadextras is set to 0 but we need it now to load data for extrafields : 
```
if ($loadextras) {	// TODO Replace this with a fetch_optionnal() on the project after the fetch_object of line.
			if (!empty($extrafields->attributes['projet']['label'])) {
				foreach ($extrafields->attributes['projet']['label'] as $key => $val) {
					$sql .= ($extrafields->attributes['projet']['type'][$key] != 'separate' ? ",efp.".$key." as options_".$key : '');
				}
			}
			if (!empty($extrafields->attributes['projet_task']['label'])) {
				foreach ($extrafields->attributes['projet_task']['label'] as $key => $val) {
					$sql .= ($extrafields->attributes['projet_task']['type'][$key] != 'separate' ? ",efpt.".$key." as options_".$key : '');
				}
			}
		}
```

### DIFF
--- a/htdocs/projet/activity/perday.php
+++ b/htdocs/projet/activity/perday.php
@@ -398,7 +398,7 @@ $search_options_pattern = 'search_task_options_';
 $extrafieldsobjectkey = 'projet_task';
 include DOL_DOCUMENT_ROOT.'/core/tpl/extrafields_list_search_sql.tpl.php';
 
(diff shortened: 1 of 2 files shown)


eldy:
When you made your PR, did you check the checkbox "Allow edit ..." ?
It seems i can't modify the PR. Need to fix param "", "" that should be 0, array()

ATM-NicolasV:
No, I don’t think I checked it. I’ll fix the pr.

PR #32064, opened 2024-11-22
FIX: reload $object after updating lines

altairis-noe:
# FIX: reload $object after updating lines
current Propal object should be fetched again when its lines are all updated, in order to reflect differences in the margin table
image


frederic34:
can you compare with https://github.com/Dolibarr/dolibarr/pull/32081

altairis-noe:
> can you compare with #32081

I kind of like your approach. What does @eldy think about it?

altairis-noe:
Hey there, @eldy
Where are we on this issue ?

eldy:
> Hey there, @eldy
> Where are we on this issue ?

I gave priority to #32081 so fix will benefit to other use cases not yet identified.

altairis-noe:
OK, thanks!

PR #32075, opened 2024-11-23
fix phpstan

frederic34:


htdocs/adherents/subscription.php
```diff
@@ -179,35 +179,33 @@
...
+			if ($result < 0) {
+				dol_print_error(null, $object->error);
 			}
+			$action = '';
 		}
 	}
 }
 
-if ($user->hasRight('adherent', 'cotisation', 'creer') && $action == 'subscription' && !$cancel) {
+if (empty($reshook) && $user->hasRight('adherent', 'cotisation', 'creer') && $action == 'subscription' && !$cancel) {
```
eldy, line 208: Why removong the empty($reshook) ?
Seem it can be 0 or 1 depending on hook result at begin.

frederic34: I'm not removing, I'm adding  it 👓

PR #32076, opened 2024-11-23
fix phpstan doc ui

frederic34:


frederic34:
image

AFTER

image

htdocs/admin/tools/ui/components/test_arrays.php
```diff
@@ -1,25 +1,47 @@
...
  */
 
 
 // Load Dolibarr environment
-@include '../../main.inc.php';
+require '../../../../main.inc.php';
 
-// Security
-if ($dolibarr_main_prod) {
-	accessforbidden('Access forbidden when $dolibarr_main_prod is set to 1');
```
eldy, line 11: Why removing this ? As it is not a fearure for end users, but for dev only, this protection avoid any vulnerability in this feature ok n priduction.

frederic34: oh sorry, too large copy paste

PR #32083, opened 2024-11-24
test multiline phpdoc

frederic34:
@mdeweerd do you know if phan will support multiline doc ? Or is there a way to do that ?


mdeweerd:
@frederic34 
IHMO it's not yet available in phan.
It [this issue](https://github.com/phan/phan/issues/1488) there is a reference to a PR that should fix it but it's pending since January.

frederic34:
@mdeweerd ok just have to wait...

frederic34:
have to wait for phan v6

PR #32090, opened 2024-11-25
FIX correct supplier price ID retrieval logic

IC-faycal:
# FIX| Supplier price ID retrieval issue

Resolved an issue with the retrieval of supplier price IDs for orders, invoices, and credit notes:
- Adjusted the query logic to ensure correct price ID fetching.
- Added validation checks to avoid incorrect mappings.
- Improved error handling for edge cases.

### DIFF
--- a/htdocs/commande/card.php
+++ b/htdocs/commande/card.php
@@ -986,7 +986,7 @@ if (empty($reshook)) {
 			$localtax2_tx = get_localtax($tva_tx, 2, $object->thirdparty);
 
 			// Margin
-			$fournprice = price2num(GETPOST('fournprice'.$predef) ? GETPOST('fournprice'.$predef) : '');
+			$fournprice = price2num(GETPOST('fournprice_predef') ? GETPOST('fournprice_predef') : '');
 			$buyingprice = price2num(GETPOST('buying_price'.$predef) != '' ? GETPOST('buying_price'.$predef) : ''); // If buying_price is '0', we muste keep this value
 
 			// Prepare a price equivalent for minimum price check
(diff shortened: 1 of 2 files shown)


IC-faycal:
@eldy it's ok for you ?

eldy:
> @eldy it's ok for you ?

As it is flagged as a fix pr, can you describe how to reproduce the bug ?

IC-faycal:
The issue is that the supplier price ID (fournprice_predef) is not being retrieved correctly because the key name in the getPost method is incorrect. Simply correcting the key name will allow the value to be retrieved properly.

eldy:
> The issue is that the supplier price ID (fournprice_predef) is not being retrieved correctly because the key name in the getPost method is incorrect. Simply correcting the key name will allow the value to be retrieved properly.

Ok, but can you describe how i can reproduce the bug (where screen do i have to go, what value do i have to enter/select and which part of screen read to see the trouble ?

IC-faycal:
PR

When we place a supplier order and select a product with a supplier purchase price (as shown in the first case, left image), the supplier is automatically retrieved by Dolibarr when the order is validated. However, in the other case, the supplier is not automatically retrieved.

eldy:
So fix is really suspicious 
* it modifies page of customer order and invoice but screenshot show a change in page for supplier orders.
 * Also i suppose that the good one according to you is the first screenshot ? If yes this is also strange. There is no reason to show the name of a supplier in a product line for a purchase order because all product line are necessarly for this supplier (defined under the ref of the purchase order). So good screenshot should be the second one and it is what we have currenctly in both v17 and develop.
 * third point, the description exaplin it fixes 3 points: Adjusted the query logic to ensure correct price ID fetching, Added validation checks to avoid incorrect mappings, Improved error handling for edge cases. But i see only one of this in the code change.

PR #32095, opened 2024-11-25
FIX executeHooks Function To Accumulate Returns

Daoud-mohamed:
# FIX executeHooks Function To Accumulate returns

Reverse the `IF` logical in executeHooks function to ensure proper accumulation of return values.

### DIFF
--- a/htdocs/core/class/hookmanager.class.php
+++ b/htdocs/core/class/hookmanager.class.php
@@ -330,17 +330,17 @@ class HookManager
 
 						if (isset($actionclassinstance->results) && is_array($actionclassinstance->results)) {
 							if ($resactiontmp > 0) {
-								$localResArray = $actionclassinstance->results;
-							} else {
 								$localResArray = array_merge_recursive($localResArray, $actionclassinstance->results);
+							} else {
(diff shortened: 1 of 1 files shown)


htdocs/core/class/hookmanager.class.php
```diff
@@ -330,17 +330,17 @@ public function executeHooks($method, $parameters = array(), &$object = null, &$
 
 						if (isset($actionclassinstance->results) && is_array($actionclassinstance->results)) {
 							if ($resactiontmp > 0) {
-								$localResArray = $actionclassinstance->results;
-							} else {
 								$localResArray = array_merge_recursive($localResArray, $actionclassinstance->results);
+							} else {
```
eldy, line 334: I think it should be the oposite: the merge when result is 0, and other value erased and not merged when result is > 0

eldy (changes requested):

htdocs/core/class/hookmanager.class.php
Daoud-mohamed: You're right. Thank you, I will close the PR.

PR #32102, opened 2024-11-26
FIX : Hidden dropdown download link in project

atm-irvine:
# FIX|Fix Hidden dropdown download link in project

I removed the tdoverflowmax class because it contains an "overflow: hidden;" that hide the dropdown of the download button and prevent access to menu when you want to download or preview the PDF.

Before : 
image

Now:
image


htdocs/projet/element.php
```diff
@@ -1288,7 +1288,7 @@
 				print "</td>\n";
 
 				// Ref
-				print '<td class="left nowraponall tdoverflowmax250">';
+				print '<td class="left nowraponall">';
```
eldy, line 1291: We need the tdoverflowmax250 for the case of object that do have a download arrwo and output a very long ref including a text (like loan , tasks). Don't know the liste exact so i suggest to remove it for all excet if 
$element instanceof Task is true or $element instanceof Loan is true

atm-irvine: I have put a table inside the td to prevent adding another td because the table head would not be right.

atm-irvine:
@eldy Is it ok for you?

PR #32105, opened 2024-11-26
Reload the page if you selected the thirsparty "Create contact"

Thomas905:
When you create a new contact, and select the thirdparty, the page"refresh"  here's a screen
image

### DIFF
--- a/htdocs/contact/card.php
+++ b/htdocs/contact/card.php
@@ -646,6 +646,16 @@ if (is_object($objcanvas) && $objcanvas->displayCanvasExists($action)) {
 
 			if ($conf->use_javascript_ajax) {
 				print "\n".'<script type="text/javascript">'."\n";
+				//reload the page, if selected thirdparty
+				print '
+				$(function() {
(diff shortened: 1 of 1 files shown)


dolibarr95:
Hello
I think : when you create from third party it's preload the address

Thomas905:
> Hello I think : when you create from third party it's preload the address

Yes but, if you create a contact with no thirdparty linked, you don't have a refresh page (lock select thirdparty) exemple as propal

lvessiller-opendsi:
It's seems to be a good idea but I don't think it's a FIX but a NEW.
@eldy
What's your advice about it ?

eldy:
> It's seems to be a good idea but I don't think it's a FIX but a NEW.
> @eldy
> What's your advice about it ?

I confirm that if this fixes a bug it is more for convenient needs than to fix a critical bug. A NEW tag should be used and change should be pushed into develop

eldy:
CLosed, to be pushed in develop.

PR #32108, opened 2024-11-26
Fix double hook printFieldPreListTitle in command list

lemathou:
Already done in v20 (#30001) but the bug is present since v18.


hregis:
@lemathou can you remove the 3 others files please, thanks

lemathou:
Oups sorry... my repo was not clean... remove done ! thx !

PR #32109, opened 2024-11-26
Fix handling of boolean extrafields in list filters

lemathou:
Cherry-pick commits done on earlier versions.


lemathou:
I reverted a commit who was in my branch cause I started from the wrong branch... (from a fix branch I did juste before)

eldy:
Do you mean there is a trouble in v16 ? And what about v20 for example, is it the same ?

lemathou:
It was fixed in v20, but I (and 3 customers) are still in v19 with the bug. I a using a lot of boolean extrafields, and I cannot sort columns without loosing filters with those fields.

So, I cherry-picked the fixes from v20 in v16, because I think it was here that the bug started.

lemathou:
I wrote "earlier version" but I though "more recent version" ... My English is bad sorry

PR #32118, opened 2024-11-27
FIX Linked elements not up to date

zahwyrix:
# FIX
If you add a linked element, it won't appear without this correction, and if you delete an element, it won't disappear. You're forced to refresh the page to have the links updated, which is rather unpleasant.


htdocs/core/actions_dellink.inc.php
```diff
@@ -76,6 +77,7 @@
 			setEventMessage($langs->trans('ErrorRecordNotFound'), 'errors');
 		}
 	}
+	$object->clearObjectLinkedCache();
```
eldy, line 80: May be we can add this before line 73 instead of here ?

PR #32119, opened 2024-11-27
FIX Bug page_y url

zahwyrix:
# FIX
In some cases, 'reposition' modified the url by assigning variables other than page_y, for example :  https://XXX/ticket/card.php?id=2158&action=dellink&token=e88288a7e4488ebafb81a80fa5f58913&dellinkid=20308.60009765625.60009765625&page_y=2113.60009765625

Here, for example, we were unable to delete a linked element, which is logical when you see the contents of dellinkid

### DIFF
--- a/htdocs/core/js/lib_foot.js.php
+++ b/htdocs/core/js/lib_foot.js.php
@@ -243,21 +243,27 @@ if ($conf->browser->layout != 'phone') {
 print "\n/* JS CODE TO ENABLE reposition management (does not work if a redirect is done after action of submission) */\n";
 print '
     jQuery(document).ready(function() {
-        page_y = getParameterByName("page_y", 0);
-        if (page_y == 0) page_y = jQuery("#page_y").text();
+		/* If page_y set, we set scrollbar with it */
+        page_y = getParameterByName("page_y", 0);				/* search in GET parameter */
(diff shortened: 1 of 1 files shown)


htdocs/core/js/lib_foot.js.php
```diff
@@ -242,38 +242,28 @@
 // Code to manage reposition
 print "\n/* JS CODE TO ENABLE reposition management (does not work if a redirect is done after action of submission) */\n";
 print '
-	jQuery(document).ready(function() {
-				/* If page_y set, we set scrollbar with it */
```
eldy, line 246: Why removing all the comments that help to understand the code ?

htdocs/core/js/lib_foot.js.php
```diff
@@ -242,38 +242,28 @@
...
-				page_y=getParameterByName(\'page_y\', 0);				/* search in GET parameter */
-				if (page_y == 0) page_y = jQuery("#page_y").text();		/* search in POST parameter that is filed at bottom of page */
-				if (page_y > 0)
-				{
-					console.log("page_y found is "+page_y);
```
eldy, line 251: Same question for console.logthat are necessary to debug.

Issue #31619, opened 2024-10-29
On Ticket Module filters are prefilled with "another" value 


Timoune974:
### Bug

On Ticket Module filters are prefilled with "another" value and don't let user to change it 

### Dolibarr Version

20.0.1

### Environment PHP

8.1.30

### Environment Database

MySQL or MariaDB 10.5.26-MariaDB-0+deb11u2

### Steps to reproduce the behavior and expected behavior

1) click on the ticket “list” submenu ==> instead of only displaying open tickets it displays all tickets 
2) filters are pre-filled with values
Type ==> Other 
Ticket group ==> Other 
Severity ==> normal 
if you try to empty one of these fields the filter will automatically take the "other" value

ghost:
Hello, bug confirmed on standard v20 & develop.

ghost:
If default value is selected for ticket severity, groupe or type in dictionnaries this is "normal" functionnality

Timoune974:
Hello, 
it's not "normal", the default values in dictionnaries must be use only when you create the ticket to fill mandatory fields, not when you call the ticket list. 
By default filters must be empty when you click on list sub menu in ticket, this new mode of operation make the screen less user friendly because if you just want refresh the page by clicking on the search button or if you want just search with one critera, you must clean all fields before

and another modification in this version is when you click on list sub menu, before you had only ticket in "open" status now you have all tickets, let me know if we need to open a new ticket for this.
Regards,

PR #32120, opened 2024-11-27
18.0 fix select type on ticket list

thomas-Ngr:
# FIX Select type in ticket list

see issue https://github.com/Dolibarr/dolibarr/issues/31619 (issue mentions 20.0 ; I confirm on 18.0 and develop)

We have a problem on ticket type selection :
- on creation form, default value should be set and empty value should be forbidden.
- on ticket list search field, default value should not be set and empty value should be allowed. 

The function selectTypesTickets deals with many cases when `$selected` is empty :

- the field can be empty, and we don't want to select the default value (this is the **case in ticket list search field**, which is a multiselect)
- the field can be empty, and we want to select the default value (theoritical case, I will not deal with it)
- the field should not be empty, and we want to select the default value (this is the **case in ticket creation card**)
- the field should not be empty, and there is no default value (this **may be the case in ticket creation card**) -> this means the select will be selected on the first entry. The function should display an entry that is obviously wrong, like 'Please select type.'

To illustrate the problem on ticket list search field, when I load ticket/list.php (or do a search without setting the field for ticket type), the search field for type is set to default by DLB : 

image

This means that I have to remove this field each time I do a new search.

This PR :

- modifies selectTypesTickets to match these requirements
- changes the call to selectTypesTickets in creation form to forbid empty value.

As a consequence : 
- on ticket list, the search field for type remains empty
- on creation card, the default value is selected
- on creation card, **if there is no default value**, the value '-- Select type --' is the first of the list and is used.
image

### DIFF
--- a/htdocs/core/class/html.formticket.class.php
+++ b/htdocs/core/class/html.formticket.class.php
@@ -716,7 +716,7 @@ class FormTicket
 		$ticketstat->loadCacheTypesTickets();
 
 		print '<select id="select'.$htmlname.'" class="flat minwidth100'.($morecss ? ' '.$morecss : '').'" name="'.$htmlname.($multiselect?'[]':'').'"'.($multiselect?' multiple':'').'>';
-		if ($empty) {
+		if ($empty && !$multiselect) {
 			print '<option value="">&nbsp;</option>';
 		}
 
@@ -753,7 +753,7 @@ class FormTicket
 					print ' selected="selected"';
 				} elseif (in_array($id, $selected)) {
 					print ' selected="selected"';
-				} elseif ($arraytypes['use_default'] == "1" && empty($selected)) {
+				} elseif ($arraytypes['use_default'] == "1" && empty($selected) && !$multiselect) {
 					print ' selected="selected"';
 				}
 


rycks:
i will try your fix on my dolibarr :)

rycks (approved):
ok for me

lvessiller-opendsi (approved):

lvessiller-opendsi:
Sorry but I have changed my point of view and I revert my review.
There still exist a problem when you edit the type field, you can reset the value if you select "-- Select type --" value.
However this field must be mandatory.

htdocs/core/class/html.formticket.class.php
```diff
@@ -753,7 +755,7 @@ public function selectTypesTickets($selected = '', $htmlname = 'tickettype', $fi
 					print ' selected="selected"';
 				} elseif (in_array($id, $selected)) {
 					print ' selected="selected"';
-				} elseif ($arraytypes['use_default'] == "1" && empty($selected)) {
+				} elseif ($arraytypes['use_default'] == "1" && empty($selected) && !$empty) {
 					print ' selected="selected"';
```
lvessiller-opendsi, line 759: Maybe the test !$empty should be replace with a specific list mode using multiselect

lvessiller-opendsi (changes requested):

thomas-Ngr:
> Sorry but I have changed my point of view and I revert my review. There still exist a problem when you edit the type field, you can reset the value if you select "-- Select type --" value. However this field must be mandatory.

This is due to the fact that these fields are not checked before updating.

The right way to fix this would be to use a function Ticket::validate() in create() and update(). But this function does not exist in DLB18 (I did not look after).

So I implemented a control in line with the code for ticket creation.

htdocs/ticket/card.php
```diff
@@ -679,19 +679,27 @@
 	if ($action == "change_property" && GETPOST('btn_update_ticket_prop', 'alpha') && $permissiontoadd) {
 		$object->fetch(GETPOST('id', 'int'), '', GETPOST('track_id', 'alpha'));
 
-		$object->type_code = GETPOST('update_value_type', 'aZ09');
-		$object->severity_code = GETPOST('update_value_severity', 'aZ09');
-		$object->category_code = GETPOST('update_value_category', 'aZ09');
+		if (!GETPOST("update_value_type", 'alpha')) {
```
thomas-Ngr, line 682: > Sorry but I have changed my point of view and I revert my review. There still exist a problem when you edit the type field, you can reset the value if you select "-- Select type --" value. However this field must be mandatory.

This is due to the fact that these fields are not checked before updating.

The right way to fix this would be to use a function Ticket::validate() in create() and update(). But this function does not exist in DLB18 (I did not look after).

So I implemented a control in line with the code for ticket creation.

PR #32129, opened 2024-11-28
NEW : Billing Term field database

atm-irvine:
# NEW|New Billing term field for model invoice
First PR for new field billing term in database : this field will allow to choice prepaid or postpaid for model invoice.


htdocs/install/mysql/tables/llx_facture_rec.sql
```diff
@@ -68,6 +68,7 @@ create table llx_facture_rec
   usenewprice        integer DEFAULT 0,			-- update invoice with current price of product instead of recorded price
   frequency          integer,						-- frequency (for example: 3 for every 3 month)
   unit_frequency     varchar(2) DEFAULT 'm',		-- 'm' for month (date_when must be a day <= 28), 'y' for year, ...
+  billing_term       integer DEFAULT 0,
```
eldy, line 71: Can you provide more information on the differznce between this field and the existing one fk_cond_reglement ?
May be provide 2 use case for each value ?

atm-irvine: I have introduced a new field called billing_term to the model, with two possible values: postpaid or prepaid. By default, models are set to prepaid, meaning that invoice lines will reflect the dates of the current billing period when the invoice is generated. If the billing_term is set to postpaid, the invoice lines will instead correspond to the previous billing period. For instance, in a monthly billing cycle, a prepaid invoice for January would reflect January's dates, while a postpaid invoice generated in January would reflect December's dates. Similarly, for a quarterly cycle, a postpaid invoice generated for Q2 would cover the dates for Q1.

This differs from the fk_cond_reglement field, which governs the payment terms, such as whether payment is due upon receipt, within 30 days end-of-month, or within 10 days. While billing_term determines the period referenced in the invoice lines (current or prior), fk_cond_reglement defines when payment is expected, addressing distinct parts of the invoicing process.

Use Case Example: Consider a subscription service. For a prepaid model, the customer is billed at the start of January for services to be delivered in January. In contrast, for a postpaid model, the customer is billed at the start of January for services delivered in December. This allows flexibility in how the company aligns its invoicing practices with its operational or contractual obligations.

atm-maxime:
@eldy is the explanation ok for you ?

htdocs/install/mysql/tables/llx_facture_rec.sql
eldy: I see so the new field is not related how the payment term will be filled but to how the field start/end will be filled. You suggest 2 different rules but ibthink in a future we may have more.
So i suggest to rename the field into another name like
rule_for_lines_dates
Value can be 'prepaid or 'postpaid'.

htdocs/install/mysql/migration/20.0.0-21.0.0.sql
```diff
@@ -382,6 +382,7 @@ INSERT INTO llx_c_type_contact (element, source, code, libelle, active ) values
 INSERT INTO llx_c_type_contact (element, source, code, libelle, active ) values ('expedition', 'external', 'SHIPPING',      'Customer shipping contact', 1);
 
 ALTER TABLE llx_facture_rec ADD COLUMN fk_societe_rib integer DEFAULT NULL;
+ALTER TABLE llx_facture_rec ADD COLUMN rule_for_lines_dates varchar(255) DEFAULT 'prepaid';
```
eldy, line 385: Can you move the addition of this column into file 21.0.0-22.0.0.sql

eldy (changes requested):

htdocs/install/mysql/migration/20.0.0-21.0.0.sql
atm-irvine: The file has been moved

atm-irvine:
@eldy I've split SQL into this new PR : https://github.com/Dolibarr/dolibarr/pull/33237

eldy:
SQL change in PR https://github.com/Dolibarr/dolibarr/pull/33237 has been validated.
May be you can close this PR to push the new one or try to fix this one to use the new name. 
You decide.

atm-irvine:
@eldy I'm about to close this PR, you can find the implementation here : https://github.com/Dolibarr/dolibarr/pull/33252

PR #32142, opened 2024-11-29
FIX select on ticket list for fields group and severity (issue #31619)

thomas-Ngr:
# FIX select group and severity search fields on ticket list

issue #31619 (indicates 20.0, but I encountered it also on develop)
see also PR #32120 , which is similar but for ticket type, and brings more details.

group and severity must be set when creating/updating a ticket, and have a default value.
However, on search fields, we don't want to set the default value.

This PR sets the default value only if `$empty` is not set.
These are the changes brought by this PR : 

- when $empty is set (case search lists), the empty value is selected
- when $empty value is not set, the default value is selected. If there is no default value, the firs value of the list is selected.

### DIFF
--- a/htdocs/core/class/html.formticket.class.php
+++ b/htdocs/core/class/html.formticket.class.php
@@ -1203,7 +1203,7 @@ class FormTicket
 					print ' selected="selected"';
 				} elseif ($arrayseverities['use_default'] == "1" && empty($selected)) {
 					print ' selected="selected"';
-				} elseif (count($conf->cache['severity_tickets']) == 1 && (!$empty || $empty == 'ifone')) {	// If only 1 choice, we autoselect it
+				} elseif (count($ticketstat->cache_severity_tickets) == 1 && (!$empty || $empty == 'ifone')) {	// If only 1 choice, we autoselect it
 					print ' selected="selected"';
 				}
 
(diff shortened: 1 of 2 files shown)


lvessiller-opendsi (approved):

eldy:
I disagree with the change for the following reason:
This pr fix the trouble by introducing a way of working that differs on how other combo list had adressed the same trouble.
The way a lot of other combolist works was to use the string "Select a new elem" into the param $empty of the function of the component (so propagated from the caller) instead of being hard coded into the component itself.
Also when we want an auto select of a value in combo when combo has only 1 element, we can use the value 'ifone' for the param $empty.
So, i suggest insteaf to try a fix by a backport of this commit b56ac120931ed46f14617b9c19a7eaf6f229273f

thomas-Ngr:
@eldy @lvessiller-opendsi @rycks I force-pushed a new version : 
- backport https://github.com/Dolibarr/dolibarr/commit/b56ac120931ed46f14617b9c19a7eaf6f229273f
- set `&nbsp;` as default empty value on the list.php for both functions `selectGroupTickets` and `selectSeveritiesTickets`

NB - the bug exists also on develop despite commit b56ac120931ed46f14617b9c19a7eaf6f229273f, so this fix may be ported upwards up to develop branch. here the ticket list on develop, showing default values instead of empty boxes : 

image

thomas-Ngr:
I had to restart the PR on #https://github.com/Dolibarr/dolibarr/pull/32411 because the branch name created an issue in precommit. (it contained "18.0")

PR #32144, opened 2024-11-29
FIX data too long for column signature

atm-GregM:
FIX data too long for column signature

Some users cannot add their signature banners correctly.
Error message: Data too long for column 'signature' at row 1

Changing TEXT to LONGTEXT for the signature field on llx_user table


hregis:
@atm-GregM change in llx_user.sql file too , thanks

PR #32148, opened 2024-11-29
FIX #31890 storing empty extrafield

### LINKED ISSUE #31890: v19.0.4 line extrafield in proposal cannot be saved, if empty
### Bug

Proposal line extrafield with contents: contents cannot be completely deleted - will keep the original contents after save. However, changes of the contents get successfully stored.

### Dolibarr Version

Confirmed at least on v19.0.2, v19.0.3 and v19.0.4

### Environment PHP

8.1

### Environment Database

MariaDB

1. create extrafield of type String for proposal line
2. create a proposal with one line with text in extrafield 
3. click on pencil, delete the text of the extrafield and press save

The extrafield now still contains the original text, log has a SQL error

priojk:
Trying to delete an existing contents of an extrafield was ignored, the original contents was still present after the edit. Empty input needs explicit handling.

### DIFF
--- a/htdocs/comm/propal/card.php
+++ b/htdocs/comm/propal/card.php
@@ -1011,10 +1011,13 @@ if (empty($reshook)) {
 		// Extrafields
 		$extralabelsline = $extrafields->fetch_name_optionals_label($object->table_element_line);
 		$array_options = $extrafields->getOptionalsFromPost($object->table_element_line, $predef);
-		// Unset extrafield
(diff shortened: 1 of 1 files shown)


JonBendtsen:
devcamp munich 2024

priojk:
This would only apply to line extrafields of propal. The problem is with all line extrafields. Will submit a more general fix.

PR #32150, opened 2024-11-29
FIX better error when api tempdir is not writeable

JonBendtsen:
FIX better error when api tempdir is not writeable
When api/temp folder does not exist or isn't writeable dol_syslog what the folder is and print where to find that information

devcamp munich 2024


JonBendtsen:
I am running this image

which I think is listed using this id on hub.docker.com    51ab7349fbc8

docker.io/dolibarr/dolibarr   latest            8c51fea87df4  6 days ago     825 MB

PR #32163, opened 2024-11-30
Fix #32161 - Payment terms are not taken from thirdparty when proposal is created from project

### LINKED ISSUE #32161: Payment terms are not taken from thirdparty when proposal is created from project
### Bug

Payment terms are not taken from thirdparty when proposal is created from project

### Dolibarr Version

19.0

### Environment PHP

n/a

### Environment Database

n/a

1. Create a new Thirdparty and set payment terms to any value
2. Create a new project with this thirdparty
3. Create a new proposal from the project

Bug: Payment terms are empty and not taken from thirdparty.

### Attached files

_No response_

glu000:
FIX#32161
Payment terms are not taken from thirdparty when proposal is created from project

### DIFF
--- a/htdocs/comm/propal/card.php
+++ b/htdocs/comm/propal/card.php
@@ -1766,6 +1766,7 @@ if ($action == 'create') {
 	$fk_account = GETPOST('fk_account', 'int');
 
 	// Load objectsrc
+	$soc = NULL;
 	if (!empty($origin) && !empty($originid)) {
 		// Parse element/subelement (ex: project_task)
 		$element = $subelement = $origin;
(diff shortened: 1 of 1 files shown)


htdocs/comm/propal/card.php
```diff
@@ -1766,6 +1766,7 @@
 	$fk_account = GETPOST('fk_account', 'int');
 
 	// Load objectsrc
+	$soc = null;
```
eldy, line 1769: $soc is already defined line 1757.
An later the added test 
if (!empty($soc)) {
would become useless.

glu000: Thanks, corrected it.

Issue #32134, opened 2024-11-28
Issue Detected in Dolibarr API 20.0.2: Updating Invoice Notes


natalieboucher:
### Bug

Context
In Dolibarr version 20.0.2, we have encountered an issue when attempting to update the public and private notes of an invoice using the API endpoint PUT /invoice/{id}. This behavior differs from the previous version, where this functionality worked correctly.

### Dolibarr Version

20.0.1 ; 20.0.2

### Environment PHP

7.3.33

### Environment Database

MySQL or MariaDB 8.0.37

### Steps to reproduce the behavior and expected behavior

Problem Description

Use Case:

We attempt to update the public and private notes of an invoice by sending an HTTP PUT request to the API endpoint PUT /invoice/{id}.
The payload sent includes the fields for the notes. For example:

[...]

PR #32189, opened 2024-12-01
Updating invoice notes 32134

JonBendtsen:
FIX #32134 Updating Invoice Notes
throw RestException(500) if update of invoice fails

PR #32190, opened 2024-12-01
throw RestException(500) if update of invoice fails #32134

JonBendtsen:
FIX https://github.com/Dolibarr/dolibarr/issues/32134 Updating Invoice Notes
throw RestException(500) if update of invoice fails


JonBendtsen:
Even though @natalieboucher said it worked in version 19, I decided to make invoice API use same structure as order API, and from v18 because v18 are still released as LTS.

JonBendtsen:
devcamp munich 2024

frederic34:
@JonBendtsen it will throw a phan error in develop like in your closed PR
https://github.com/Dolibarr/dolibarr/pull/32189/files

JonBendtsen:
> it will throw a phan error in develop like in your closed PR

Are orders any different than invoices?

**commande/class/api_orders.class.php**
```
		if ($this->commande->update(DolibarrApiAccess::$user) > 0) {
			return $this->get($id);
		} else {
			throw new RestException(500, $this->commande->error);
		}
```

**compta/facture/class/api_invoices.class.php**
```
		if ($this->invoice->update(DolibarrApiAccess::$user) > 0) {
			return $this->get($id);
		} else {
			throw new RestException(500, $this->invoice->error);
		}
```

Because all I am doing is making the code pretty identical

The public function update( seems pretty similar **between orders**
https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/compta/facture/class/facture.class.php#L2500

**and invoices**
https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/commande/class/commande.class.php#L3349


Am I wrong in expecting that phan would also throw an error if the change was in the existing order code?

I also actually tested this code, and with it I do get an error in the API, where as without it I get http code 200 as if no errors

htdocs/compta/facture/class/api_invoices.class.php
```diff
@@ -632,8 +632,10 @@ public function put($id, $request_data = null)
 			}
 		}
 
-		if ($this->invoice->update(DolibarrApiAccess::$user)) {
+		if ($this->invoice->update(DolibarrApiAccess::$user) > 0) {
 			return $this->get($id);
+		} else {
+			throw new RestException(500, $this->invoice->error);
```
eldy, line 638: If you add a thro, you must remove the return false after, because this code become unreachable.

JonBendtsen: Okay, that makes sense

lvessiller-opendsi (approved):

rycks (approved):
ok for me thanks !

PR #32194, opened 2024-12-01
FIX Accounting Closure Duplicates, Subledger accounts, Account Labels and more... Update bookkeeping.class.php

josett225:
FIX I did further testing and investigation and I fixed the following issues that stop doing a full closure without duplicate lines generated by an unclean database :

    - different label_compte with same account number
    - removing label_compte is raising an issue and the code in the line around 2770 $bookkeeping->label_compte = $obj->label_compte;
    - different subledger_label with same subledger_account
    - empty versus null values for subledger_label and subledger_account
    - opening_balance is 0 as it creates a bookkeeping entry for now.  

See other comments here on this draft: https://github.com/Dolibarr/dolibarr/pull/30653
See also Fix required here : FIX - Update Accounting closure with missing too many A-Nouveau #30039)](https://github.com/Dolibarr/dolibarr/pull/30039)

### DIFF
--- a/htdocs/accountancy/class/bookkeeping.class.php
+++ b/htdocs/accountancy/class/bookkeeping.class.php
@@ -2692,10 +2692,8 @@ class BookKeeping extends CommonObject
 
 				$sql = 'SELECT';
 				$sql .= " t.numero_compte,";
-				$sql .= " t.label_compte,";
 				if ($separate_auxiliary_account) {
-					$sql .= " t.subledger_account,";
-					$sql .= " t.subledger_label,";
+					$sql .= " NULLIF(t.subledger_account, '') as subledger_account,"; // fix db issues with Null or "" values
 				}
 				$sql .= " aa.pcg_type,";
 				$sql .= " (SUM(t.credit) - SUM(t.debit)) as opening_balance";
@@ -2707,10 +2705,11 @@ class BookKeeping extends CommonObject
 				$sql .= ' AND aa.pcg_type IN (' . $this->db->sanitize(implode(',', $pcg_type_filter), 1) . ')';
 				$sql .= " AND DATE(t.doc_date) >= '" . $this->db->idate($fiscal_period->date_start) . "'";
 				$sql .= " AND DATE(t.doc_date) <= '" . $this->db->idate($fiscal_period->date_end) . "'";
-				$sql .= ' GROUP BY t.numero_compte, t.label_compte, aa.pcg_type';
+				$sql .= ' GROUP BY t.numero_compte, aa.pcg_type';
 				if ($separate_auxiliary_account) {
-					$sql .= ' ,t.subledger_account, t.subledger_label';
+					$sql .= " , NULLIF(t.subledger_account, '')";
 				}
+				$sql .= ' HAVING (SUM(t.credit) - SUM(t.debit)) != 0 '; // Exclude rows with opening_balance = 0
 				$sql .= $this->db->order("t.numero_compte", "ASC");
 
 				$resql = $this->db->query($sql);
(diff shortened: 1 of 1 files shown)


htdocs/accountancy/class/bookkeeping.class.php
```diff
@@ -2707,10 +2706,11 @@ public function closeFiscalPeriod($fiscal_period_id, $new_fiscal_period_id, $sep
 				$sql .= ' AND aa.pcg_type IN (' . $this->db->sanitize(implode(',', $pcg_type_filter), 1) . ')';
 				$sql .= " AND DATE(t.doc_date) >= '" . $this->db->idate($fiscal_period->date_start) . "'";
 				$sql .= " AND DATE(t.doc_date) <= '" . $this->db->idate($fiscal_period->date_end) . "'";
-				$sql .= ' GROUP BY t.numero_compte, t.label_compte, aa.pcg_type';
+				$sql .= ' GROUP BY t.numero_compte, aa.pcg_type';
 				if ($separate_auxiliary_account) {
-					$sql .= ' ,t.subledger_account, t.subledger_label';
+					$sql .= " , NULLIF(t.subledger_account, '')";
```
eldy, line 2711: Adding a function call into a group by is strange. DoesYour fix work if you replace this with
`$sql .= ", subledger_account";`

josett225: Hi @eldy

It does not work if you remove the NULLIF function. We still get duplicates.
Here are some screenshots with and without:

SQLDIFFWITHOUTISNULLINGROUP
SQLDIFFWITHISNULLINGROUP

josett225:
Hi @eldy 
Just answered in your comment.

eldy:
@aspangaro is this patch looks ok for you ?

aspangaro:
It looks pretty good to me, and completes my patch for the sub-accounts section.

I'll do a few tests and confirm that.

josett225:
Hi @aspangaro and @eldy 
To work correctly the PR https://github.com/Dolibarr/dolibarr/pull/30039 has to be validated also.
Could you please validate it or do you want me to add it to this PR and I cancel  the other one

eldy:
> Hi @aspangaro and @eldy 
> To work correctly the PR https://github.com/Dolibarr/dolibarr/pull/30039 has to be validated also.
> Could you please validate it or do you want me to add it to this PR and I cancel  the other one

Yes please, close old one and merge it into one request (this pr), so it will be easier to follow all the changes of the fix...

htdocs/accountancy/class/bookkeeping.class.php
```diff
@@ -2732,23 +2736,38 @@ public function closeFiscalPeriod($fiscal_period_id, $new_fiscal_period_id, $sep
 							$bookkeeping = new BookKeeping($this->db);
 							$bookkeeping->doc_date = $new_fiscal_period->date_start;
 							$bookkeeping->date_lim_reglement = '';
-							$bookkeeping->doc_ref = $new_fiscal_period->label;
+							$bookkeeping->doc_ref = $fiscal_period->label;
```
eldy, line 2739: I still have an interrogation with this replacement.
Why $fiscal_period->label instead of $new_fiscal_period->label  ? Done on prupose ?

alexandre-janniaux:
Just so that I follow correctly, this is still about fixing the option where the details of the RAN are merged in a single transaction per subaccount, not yet the new feature where the details of the RAN would be integrated one by one into the AN journal (which would actually duplicates the operation label/account I guess) ?

josett225:
Yes this is correct for now this is a balance mode with subledgers. Doing it in a more detailed view requires a bit more of work as there are several issues as taking in consideration the 'lettrage' for e.g. and some other aspects. I am working right now with 2 accounting experts as my first version I did was a bit incomplete. I used the label and amount to avoid duplicates but it was not perfect at all. So this is requiring a bit more of work.

Issue #32207, opened 2024-12-03
Bug: In Multi currency Supplier Order Multi currency buying price is coming as NaN


support-accellier:
### Bug

In Multi currency Supplier Order Multi currency buying price is coming as NaN when selecting the line item for which Vendor buying price is defined in a currency other than Dolibarr's base currency

### Dolibarr Version

20.0

### Environment PHP

8.2

### Steps to reproduce the behavior and expected behavior

1. Create a Supplier Order in currency other than base currency set in Dolibarr
2. Add the line item for which the vendor buying price is defined in currency other than base currency set in Dolibarr
3. Multi currency price comes as NaN

eldy:
@support-accellier it seems it has been fixed with last version ni branch v20. Can you check ?

support-accellier:
Thanks this issue has been fixed now.

Issue #32206, opened 2024-12-03
Bug: Incorrect Discounts or Credit is applied in a multicurrency Invoice


support-accellier:
### Bug

If an invoice is created in a currency other than Dolibarr installation's default currency and then when we apply the available discount or credit in this invoice then the discount line is inserted in this invoice without converting the discount/credit to the invoice currency

### Dolibarr Version

20.0

### Environment PHP

8.2

### Steps to reproduce the behavior and expected behavior

1. Dolibarr default currency is USD
2.  Create a customer with currency GBP
3.  Add Absolute discount to this customer created in step 2. Absolute discount is created in USD as that is the only option in Dolibarr (e.g., 100 USD)
4. Create invoice for this customer in GBP (127 GBP). Conversion between USD to GBP is defined as 1.27 in the system.
5. Apply the discount in the invoice. Invoice will still show the unpaid amount

### Attached files

![image](https://github.com/user-attachments/assets/16903c3e-47a1-4e4d-95a7-64dfcc2596ee)

PR #32209, opened 2024-12-03
Bugfix for #32206 and #32207

support-accellier:
# FIX
Bug: Incorrect Discounts or Credit is applied in a multicurrency Invoice #32206
Bug: In Multi currency Supplier Order Multi currency buying price is coming as NaN #32207

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -2638,6 +2638,13 @@ class Facture extends CommonInvoice
 			$facligne->multicurrency_total_tva = -$remise->multicurrency_amount_tva;
 			$facligne->multicurrency_total_ttc = -$remise->multicurrency_amount_ttc;
 
+			//Bugfix#32206 - Incorrect Discounts or Credit is applied in a multicurrency Invoice 
+			if ($this->multicurrency_tx > 0) {
+				$facligne->multicurrency_subprice = round(-$remise->multicurrency_subprice * $this->multicurrency_tx,2);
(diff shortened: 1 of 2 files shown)


support-accellier:
#32206 and #32207 has been fixed in this pull request

htdocs/compta/facture/class/facture.class.php
```diff
@@ -2638,6 +2638,13 @@ public function insert_discount($idremise)
 			$facligne->multicurrency_total_tva = -$remise->multicurrency_amount_tva;
 			$facligne->multicurrency_total_ttc = -$remise->multicurrency_amount_ttc;
 
+			//Bugfix#32206 - Incorrect Discounts or Credit is applied in a multicurrency Invoice 
+			if ($this->multicurrency_tx > 0) {
```
eldy, line 2642: When multicurrency_tx is set, all other fields multicurrency should already contains a value include the rate. So bug shoul be at a higher place.

pratushraj: Yes You are right, but here the calculation is Without Multicurreny, Because it'll directly calculate with the Invoice multicurrency Rate.

pratushraj: Yes you are right, I have created a pull request for this.
https://github.com/Dolibarr/dolibarr/pull/32872

pratushraj: Also we have to migrate the previous data as well.

eldy:
According to answer another PR https://github.com/Dolibarr/dolibarr/pull/32872 has been created, so i close this one.

PR #32210, opened 2024-12-03
NEW : auto-detect country from code in creation

atm-maxime:
When using API to create a thirdparty, we currently must give the country_id.
To have the country_id we must use the API setup/dict/countries
This PR simplifies this by retrieving the country id from the code if the id is empty and code is filled.
This allows to call the API POST /thirdparties with the country_code in order to create a new thirdparty.


atm-maxime:
@eldy should be good now, sorry for the phpstan error

PR #32215, opened 2024-12-03
NEW : Stock at date column is sortable / filterable in stockatdate.php page

jyhere:
For the needs of a customer, I reworked the logical part (sql & php) of the stockatdate.php page to be able to sort / filter on the Stock at date column, in particular to be able to output all stock at date greater than 0.
Tested on fairly large volumes (+22000 products, +40000 stock movements), performance was similar to the previous code. (sql part is about 0.25s on by dev laptop for 500 results per page)

image


jyhere:
This is a false positive: $stock_at_date is always defined because the condition in which it is found is always the same. But I've modified the code to overwrite the $objp->stock_at_date property instead, which is a useless property in future mode.

eldy:
Lets summarize.
The algorithm of old code:
- We take all product and their current stock into an array.
- We take all past stock movement for each product into an array
We combine the two array to get the result.

The algorithm of the new code.
- We get the list of all product with a join on a subquery to get all stock and a join on a subquery to get all stock movement. Substraction is done inside the main SQL allowing to use the HAVING instruction.

Looking at code only, your second method seems to be slower compared to old one. Testing on less than 100 000 record is not significant. We know a lot of users of Dolibarr have more than 500 000 reference. I think problem may appears only with such a volume. But may be not.

Can someone with a very large database of product/stock/movement can take the code of the new page and run it to say if its ok for them ?

PR #32216, opened 2024-12-03
Ticket : don't autofill search field for ticket type on multiselect

thomas-Ngr:
# FIX Select type on ticket list
 See issue https://github.com/Dolibarr/dolibarr/issues/31619

Problem 1 : when going to the ticket list, the search field for ticket type is filled with the default value, even if we do not want it. So we have to remove it manually for each search.
image

Problem 2 : the empty value is available for the multiselect. Is should not be (moreover, it triggers an SQL error when chosen).
image

This PR fixes both problems.


rycks (approved):
thanks !

lvessiller-opendsi (approved):

PR #32217, opened 2024-12-03
FIX Missing trigger on situation invoice

atm-thibaultf:
# FIX | Missing trigger on invoice situation creation
Missing trigger on invoice situation creation for allow action on extrafields

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -2050,7 +2050,7 @@ if (empty($reshook)) {
 					$extrafields->fetch_name_optionals_label($nextSituationInvoice->table_element);
 					$ret = $extrafields->setOptionalsFromPost(null, $nextSituationInvoice);
 					if ($ret > 0) {
-						$nextSituationInvoice->insertExtraFields();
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/card.php
```diff
@@ -2050,7 +2050,7 @@
 					$extrafields->fetch_name_optionals_label($nextSituationInvoice->table_element);
 					$ret = $extrafields->setOptionalsFromPost(null, $nextSituationInvoice);
 					if ($ret > 0) {
-						$nextSituationInvoice->insertExtraFields();
+						$nextSituationInvoice->insertExtraFields('BILL_CREATE');
```
eldy, line 2053: Problem here is that the trigger BILL_CREATE was already called into createFromCurrent.
I suggest to try this. Just before the 
`$object->createFromCurrent()`
do
```
$extrafields->fetch_name_optionals_label($$object->table_element);
$ret = $extrafields->setOptionalsFromPost(null, $object);
```
so the extrafields are already set when calling the createFromCurrent.Does this work for you ?

PR #32226, opened 2024-12-04
Fix 20.0 header call for product/stats/supplier_proposal.php

thersane-john:
# FIX header call for product/stats/supplier_proposal.php

param offset in function call...


eldy:
When you made your pr, did you check the checkkox "allow edit".
Because i can't use thz button Update on it to merge last branch version.

thersane-john:
@eldy you can retry

PR #32228, opened 2024-12-04
FIX wrong param entity value

hregis:


hregis:
@eldy it's ok for you ?

eldy (approved):

PR #32235, opened 2024-12-04
fix: cannot update password if dolibarr_main_authentication="mc" (multicompany)

FHenry:
If $dolibarr_main_authentication=mc impossible to udate password because filed readonly is displayed

### DIFF
--- a/htdocs/fourn/commande/list.php
+++ b/htdocs/fourn/commande/list.php
@@ -511,7 +511,7 @@ if (empty($reshook)) {
 							}
 
 							$tva_tx = $lines[$i]->tva_tx;
-							if (!empty($lines[$i]->vat_src_code) && !preg_match('/\(/', $tva_tx)) {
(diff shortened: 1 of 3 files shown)


eldy:
@hregis is it ok for you ?

hregis:
@FHenry @eldy
 it's a Multicompany bug since v19, I didn't see that the `$caneditpassword` variable was changed to `$caneditpasswordandsee`
I made a correction on v19: https://github.com/Dolibarr/dolibarr/pull/32246
and I will correct Multicompany.
after if we can add a test on the "mc" authentication it will save me some hooks! :-)

hregis:
@FHenry @eldy maye add this in develop branch to avoid conflict

FHenry:
We can close this PR, because use hook  in multicompany is the real fix, as $dolibarr_main_authentication can be provided by external module, all related test on this in Dolibarr Core have to be overridable by a hook.
I've just check in 20.0 the occurence of $dolibarr_main_authentication and I didn't find any other test like the one in user/card.php

@hregis I've just check our fix, it works, thank's

PR #32237, opened 2024-12-04
# NEW| Added a configuration option for the propagation to variants

EnjoyFelix:
The propagation of modifications to variant products is automatic, but there are legit cases where you don't want to propagation to occur.
> Example: The status of a parent product should be "not for sale" while the variants are "for sale"

This adds an option in the configuration of the module to prevent the propagation from occurring.

### DIFF
--- a/htdocs/langs/en_US/products.lang
+++ b/htdocs/langs/en_US/products.lang
@@ -432,3 +432,4 @@ ConfirmEditExtrafield = Select the extrafield you want modify
 ConfirmEditExtrafieldQuestion = Are you sure you want to modify this extrafield?
 ModifyValueExtrafields = Modify value of an extrafield
 OrProductsWithCategories=Or products with tags/categories
+AlwaysPropagateToVariants = Always propagate the modifications to the variants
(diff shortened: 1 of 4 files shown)


eldy:
I don't think having a global option is enough.
Each decision to propagate or not will depend on what data is changed. So a better change would be to have a checbox before the button Save to ask 
"Propagate change on variant" so user can decide or not depending on its change to prooagate on not.

daxid:
Hi @eldy, thanks for your answer,

I work with @EnjoyFelix and some context is missing to this PR.

We are developing a Dolibarr module to improve attribute propagation to variants. The default module only propagates "base" attributes to the variants. We needed to propagate also client and provider prices, as well as user defined attributes. We have detailed preferences in the module to know which attributes to propagate and ask for user confirmation before propagation.

To do so, we need to be able de deactivate the default propagation in the core "variant" module (I don't remember the english name...). I think our module could be included in the core module eventually.

@eldy What do you think ? 

Also for our client, asking each time which attribute to propagate on save will be an overhead.

eldy:
> Hi @eldy, thanks for your answer,
> 
> I work with @EnjoyFelix and some context is missing to this PR.
> 
> We are developing a Dolibarr module to improve attribute propagation to variants. The default module only propagates "base" attributes to the variants. We needed to propagate also client and provider prices, as well as user defined attributes. We have detailed preferences in the module to know which attributes to propagate and ask for user confirmation before propagation.
> 
> To do so, we need to be able de deactivate the default propagation in the core "variant" module (I don't remember the english name...). I think our module could be included in the core module eventually.
> 
> @eldy What do you think ? 
> 
> Also for our client, asking each time which attribute to propagate on save will be an overhead.
> 
> 

So if need is to disable a feature to open the way for addons, it is better to change test into
If ! getDolGlobaString("PRODUIT_ATTRIBUTE_DISABLE_PROPAGATION")
Also this must remain a hidden option with no access for user. If your module disable optuon and user enable option form setup, result will not be as expected.

eldy:

PR #32240, opened 2024-12-04
NEW Enhance selectDate Function to Support Icon Customization

Daoud-mohamed:
# Enhance selectDate Function to Support Icon Customization

- Provide the ability to customize the icon used in the component.
- Replace DOL_URL_ROOT with DOL_MAIN_URL_ROOT to enable the selectDate component to be used from a different host

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -7158,10 +7158,11 @@ class Form
 	 * @param string 				$labeladddateof Label to use for the $adddateof parameter. Deprecated. Used only when $adddateof is not an array.
 	 * @param string 				$placeholder 	Placeholder
 	 * @param 'auto'|'gmt'|'tzserver'|'tzuserrel'	$gm 	'auto' (for backward compatibility, avoid this), 'gmt' or 'tzserver' or 'tzuserrel'
+	 * @param string				$calendarpicto 	URL of the icon/image used to display the calendar
 	 * @return string               	         	Html for selectDate
 	 * @see    form_date(), select_month(), select_year(), select_dayofweek()
 	 */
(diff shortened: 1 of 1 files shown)


hregis:
@Daoud-mohamed What is the problem with DOL_URL_ROOT ?

Daoud-mohamed:
@hregis In some cases, such as when using the selectDate component in a website template encapsulated in a virtual host, the selectDate component generates a relative path for the icon using DOL_URL_ROOT, which can create a conflict. Using DOL_MAIN_URL_ROOT provides the full URL and avoids this conflict.

htdocs/core/class/html.form.class.php
```diff
@@ -7312,9 +7313,10 @@ public function selectDate($set_time = '', $prefix = 're', $h = 0, $m = 0, $empt
 						}
 						// Note: We don't need monthNames, monthNamesShort, dayNames, dayNamesShort, dayNamesMin, they are set globally on datepicker component in lib_head.js.php
 						if (!getDolGlobalString('MAIN_POPUP_CALENDAR_ON_FOCUS')) {
+							$buttonImage = $calendarpicto ?: DOL_MAIN_URL_ROOT . "/theme/" . dol_escape_js($conf->theme) . "/img/object_calendarday.png";
```
eldy, line 7316: Using DOL_MAIN_URL_ROOT here will generate pb: the url used may differe when using the compononent from backoffice.
Why not keeping DOL_URL_ROOT ? If we need to use the component from another context like a website, we can use the new parameter. Did i miss something ?

Daoud-mohamed: Yes, using the personalized icon will allow the component to be utilized in an external context. I updated the PR and reverted to using DOL_URL_ROOT

PR #32242, opened 2024-12-04
Fix hook in facture/document

Expresion:


htdocs/compta/facture/document.php
```diff
@@ -86,11 +86,19 @@
 
 
 /*
- * Actions
- */
+* Actions
+*/
 
 include DOL_DOCUMENT_ROOT.'/core/actions_linkedfiles.inc.php';
```
eldy, line 92: You can remove this line, already on new line 99

htdocs/comm/action/class/actioncomm.class.php
```diff
@@ -2702,7 +2702,6 @@ public function sendEmailsReminder()
 
 					// Load event
 					$res = $this->fetch($actionCommReminder->fk_actioncomm);
-					if ($res > 0) $res = $this->fetch_thirdparty();
```
eldy, line 2705: Why removing this line ?

Expresion (commented):
Solved

PR #32249, opened 2024-12-05
fix:  FATAL ERROR (php 8, invoice situation=2) Invoice line Save after edit (even if invoice is STANDARD)

FHenry:
On standard invoice

Edit invoice lline on std invoice trigger Fatal Error if INVOICE_USE_SITUATION=2 (as the invoice is not a situation invoice)

as $fullprogress = price2num() return '' 

image

image

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -2742,7 +2742,7 @@ if (empty($reshook)) {
 		// Invoice situation
 		if (getDolGlobalInt('INVOICE_USE_SITUATION') == 2) {
 			$previousprogress = $line->get_allprev_progress($line->fk_facture);
-			$fullprogress = price2num(GETPOST('progress', 'alpha'));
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/card.php
```diff
@@ -2742,7 +2742,7 @@
 		// Invoice situation
 		if (getDolGlobalInt('INVOICE_USE_SITUATION') == 2) {
 			$previousprogress = $line->get_allprev_progress($line->fk_facture);
-			$fullprogress = price2num(GETPOST('progress', 'alpha'));
+			$fullprogress = price2num(GETPOST('progress', 'alpha'), 'MU');
```
eldy, line 2745: The 'MU' is related to amount, so varies with setup and may be 0.
For consistent result, i recommend for percentage to use 2 instead of 'MU'

FHenry: Done

PR #32254, opened 2024-12-05
Fix setup color in display mode

thersane-john:
# FIX setup color in display mode

color item display is broken 

this patch restore old V16 display method but using V20 type of color data.

## Before patch 
image
image

## After patch
image
image


eldy:
Using type=color seems to generate  troubles:
It is no more possible to define an empty value. Empty means " nit forced used value of template.

thersane-john:
@eldy 

> sing type=color seems to generate troubles:
It is no more possible to define an empty value. Empty means " nit forced used value of template.

Yes, but here we're not in edit mode, I can replace the entry with a span and a background color... so this unnamed input also works

PR #32267, opened 2024-12-06
FIX(invoice): mutlicurrency_tx correct value

### LINKED ISSUE #32266: Incorrect multicurrency_tx when create invoice from other object
### Bug

When creating invoice from command for example
Multicurrency_tx is not correctly applied. If i set it to 1,08, it will be floor to 1 by int conversion

### Dolibarr Version

20.0.2

### Environment PHP

_No response_

### Environment Database

_No response_

- Set Multicurrency to 1,08 or whatever with some digits after comma
- Create a command for a Client with a price at 1080 ot this currency
- From this command, create directly an invoice
- that show you that multicurrency is applied, but when create
- Multicurrency tx is reduce to 1 and invoice prices are reset to 1000, all

### Attached files

_No response_

Slordef:
# FIX #32266 
Instead of taking `originmulticurrency_tx` as an integer, take it from any kind of value to match it correctly

- Fixes #32266

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -1093,7 +1093,7 @@ if (empty($reshook)) {
 				$object->fk_incoterms       = GETPOSTINT('incoterm_id');
 				$object->location_incoterms = GETPOST('location_incoterms', 'alpha');
 				$object->multicurrency_code = GETPOST('multicurrency_code', 'alpha');
-				$object->multicurrency_tx   = GETPOST('originmulticurrency_tx');
+				$object->multicurrency_tx   = GETPOSTFLOAT('originmulticurrency_tx');
 
 				// Special properties of replacement invoice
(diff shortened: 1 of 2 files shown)


htdocs/compta/facture/card.php
```diff
@@ -1093,7 +1093,7 @@
 				$object->fk_incoterms       = GETPOSTINT('incoterm_id');
 				$object->location_incoterms = GETPOST('location_incoterms', 'alpha');
 				$object->multicurrency_code = GETPOST('multicurrency_code', 'alpha');
-				$object->multicurrency_tx   = GETPOSTINT('originmulticurrency_tx');
```
eldy, line 1096: Can you use instead GETPOSTFLOAT ?

Slordef: Ok, i see that some others values are not right due to "integer" conversion too
I will fix that too in this pr

PR #32287, opened 2024-12-09
Fix/130/remove dol syslog throws

atm-etienne:
# Qual | Dol_Syslog must not throws error 

Currently, the dol_syslog function can throw errors if the provided log level is invalid. Since this function is itself called in case of an error to populate error and debug logs, it should not be capable of raising errors, as this could lead to blocking situations.

I have therefore modified the behavior of the throw to retrieve a default log level from the database via the `ErrorBadLogLevel` configuration if it exists, or use the `LOG_ERR` level otherwise.

This way, the log can still be displayed, and the user will be informed that the initial log level is incorrect through a recursive call to the function itself to convey this information.


eldy:
Did you check the checkbox "Allow edit " when you pushed your PR. It seems you PR can't be modified, so updated to be merged.

atm-etienne:
I don't see any checkbox on my PR page to modify it. I also tried to recreate it, but there's still no checkbox. What should I do?

atm-etienne:
Create new PR from my own fork, with checking "Allow edit" - https://github.com/Dolibarr/dolibarr/pull/32383

PR #32291, opened 2024-12-09
FIX: deleting a reception do not create reverse stock movements

jyhere:
How to reproduce the bug : 
- Activate stock movement on reception (eg. on validation)
- Create a supplier reception with a single product, validate the reception, and delete the reception
- check the product stock movements : only the stock movement of the reception validation is visible, not the corrective stock movement of the reception deletion.

A condition was simply inverted to fix this.

Also, eatby and sellby date has also been fixed in this part of the code. This dates must be timestamp, not sql date, otherwise this comparison will fail : 
https://github.com/Dolibarr/dolibarr/blob/8ca831ff885c1a979f93b1973f91be1183688980/htdocs/product/stock/class/mouvementstock.class.php#L393-L420


htdocs/reception/class/reception.class.php
```diff
@@ -1116,7 +1116,7 @@ public function delete(User $user)
 		$this->db->begin();
 
 		// Stock control
-		if (isModEnabled('stock') && !getDolGlobalInt('STOCK_CALCULATE_ON_RECEPTION') && $this->statut > 0) {
+		if (isModEnabled('stock') && getDolGlobalInt('STOCK_CALCULATE_ON_RECEPTION') && $this->statut > 0) {
```
eldy, line 1119: I think the test should be more complete. When do you think of
```
if (isModEnabled('stock') && ((getDolGlobalInt('STOCK_CALCULATE_ON_RECEPTION') && $this->statut > 0)
|| (getDolGlobalInt('STOCK_CALCULATE_ON_RECEPTION_CLOSE') && $this->statut == $this::STATUS_CLOSED))) {
```

jyhere: Exact ! :) :+1:

PR #32294, opened 2024-12-09
Fix hook result usage

thersane-john:
# Fix

The result of the hook pdf_getLinkedObjects is being misused. If your hook returns 1, but the resArray is cleared (as this is precisely what your use case requires), the hook ends up being ignored.

### DIFF
--- a/htdocs/core/lib/pdf.lib.php
+++ b/htdocs/core/lib/pdf.lib.php
@@ -2618,8 +2618,9 @@ function pdf_getLinkedObjects(&$object, $outputlangs)
 	if (is_object($hookmanager)) {
 		$parameters = array('linkedobjects' => $linkedobjects, 'outputlangs' => $outputlangs);
 		$action = '';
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/pdf.lib.php
```diff
@@ -2618,8 +2618,9 @@ function pdf_getLinkedObjects(&$object, $outputlangs)
 	if (is_object($hookmanager)) {
 		$parameters = array('linkedobjects' => $linkedobjects, 'outputlangs' => $outputlangs);
 		$action = '';
-		$hookmanager->executeHooks('pdf_getLinkedObjects', $parameters, $object, $action); // Note that $action and $object may have been modified by some hooks
-		if (!empty($hookmanager->resArray)) {
+		$reshook = $hookmanager->executeHooks('pdf_getLinkedObjects', $parameters, $object, $action); // Note that $action and $object may have been modified by some hooks
+		if ($reshook>0) {
```
eldy, line 2622: If reshook is 0, it means the hook ask to concat, then the resArray must be concatenate to linkedobjects.
If reshook is 1, it means the hook asked to replace completely.
so i think we are missing the cas reshook == 0 that was correctly supported before.

PR #32301, opened 2024-12-09
SQL data model

SylvainLegrand:
Prepare improvements regarding SEPA transfers.

1 - Add dictionaries for "category purpose" and "community instrument".

Users should be able to choose the value of these two parameters for each bank account as they may be different from one bank to another.

### DIFF
--- a/htdocs/install/mysql/data/llx_c_sepa_category_purpose.sql
+++ b/htdocs/install/mysql/data/llx_c_sepa_category_purpose.sql
@@ -1,0 +1,32 @@
+-- Copyright (C) 2024 		Sylvain Legrand           <contact@infras.fr>
+--
+-- This program is free software; you can redistribute it and/or modify
+-- it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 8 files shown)


hregis:
@SylvainLegrand 
are these dictionaries that can be customized by entity? (enable/disable values ​​and/or add/customize values)
if yes, an "entity" field should be added to the dictionaries. Thanks

SylvainLegrand:
@hregis 

No need to manage entities on these dictionaries.
They just need to contain the list of abbreviations needed to create SEPA files (depending on the bank's requirements).

It's like postal codes, the list remains the same regardless of the entity.

eldy:
No need for a transverse dictionarie. The possible values are technical "codes" defined by the SEPA protocol and not custom end user choice. Also the quantity of different values is very low and dos not need to consume 2 tables for that (note: Minimum size on disk to store a table is 96 Ko with innodb engine).
A list of all possible codes into the setup page of module SEPA (direct debit module and/or credit transfer) is enough.

PR #32308, opened 2024-12-09
FIX Asset Module Update to use gmt (versus tz) everywhere to avoid time/calculation issues

josett225:
FIX Asset Module issue as already discussed longely in this closed PR:
https://github.com/Dolibarr/dolibarr/pull/31434

card.php and /class/assetdepreciationoptions.class.php

This fix is to avoid issue when using asset starting date that equal to either first or last day of fiscal year.

The solution is to get all the asset module using GMT (versus TZ) everywhere to avoid creating a new function to reset time as the following:
/**
 * Function to reset the time of a date to 00:00:00
 */
private function resetTimeToMidnight($date)
{
    if (!empty($date)) {
        // Assuming $date is a timestamp or string in format compatible with strtotime
        return dol_mktime(0, 0, 0, dol_print_date($date, '%m'), dol_print_date($date, '%d'), dol_print_date($date, '%Y'),'gmt');
    }
    return $date; // Return unchanged if date is empty or null
}

### DIFF
--- a/htdocs/asset/card.php
+++ b/htdocs/asset/card.php
@@ -132,7 +132,7 @@ if (empty($reshook)) {
 
 	// Action dispose object
 	if ($action == 'confirm_disposal' && $confirm == 'yes' && $permissiontoadd) {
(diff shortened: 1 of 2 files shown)


josett225:
Hi @eldy 
The errors are neither in Asset Module nor in my code.

PR #32313, opened 2024-12-10
FIX Force the sales representative if they don't have permissions

joseplluis:
# FIX Force the sales representative if they don't have permissions


hregis:
@joseplluis If you want to limit to third parties to which the user is linked as a sales representative, you can disable this right: 
`"Extend access to all third parties AND their objects (not only third parties for which the user is linked as a sales representative)"`

if a user does not have the right to see third parties he does not see third parties!

joseplluis:
@hregis yes, and this is the current bug, users with this permission disabled can see all thirdparties
In v19 it was working fine, but now is buggy since v20 yes
In this correction we are using !$user->hasRight('societe', 'client', 'voir') that is different from $user->hasRight('societe', 'lire')

PR #32322, opened 2024-12-11
new: if default template for ticket exist we use it avec if there is email template 'all'

FHenry:
Actually if there is email tample type 'all' they are use as default in email ticket mail template even if there is default=1 for email template type ticket_send

the remove 1611 to 1617 is because the fetch al load is already done before, no need to make it twice


htdocs/core/class/html.formticket.class.php
```diff
@@ -1608,13 +1612,6 @@ public function showMessageForm($width = '40%')
 			print '<input type="hidden" name="'.$key.'" value="'.$value.'">';
 		}
 
-		// Get message template
```
eldy, line 1611: Are you sure the getEMailTemplate is always already done ? Because line 1500 the call has the condition
`if (isset($this->param['models']) && $this->param['models'] != 'none') {`

FHenry: I think it was the case $this->param['models_id'] have to be set before from the caller htdocs/ticket/card.php
https://github.com/Dolibarr/dolibarr/blob/07c2feb69caf380ba1c07600fbf9f50011a4bbe5/htdocs/ticket/card.php#L1574
But may be, may be this method can be called from other pages from external module. 
So I re-add this getEMailTemplate with the same defaultfortype parameters to avoid to return email template with "All" type

PR #32326, opened 2024-12-11

thomas-Ngr:

This PR:

SQL Part : #26190

follow-up on #30996 (closed) and previously #30230 (closed because file expedition.class has been split into two files)


thomas-Ngr:
@eldy is there anything to add/modify to this PR ?

thomas-Ngr:
@eldy do you plan to merge this feature ?

PR #32333, opened 2024-12-12
FIX avoid php warnings

hregis:


htdocs/compta/facture/list.php
```diff
@@ -174,6 +174,13 @@
 $pageprev = $page - 1;
 $pagenext = $page + 1;
 
+// Security check
```
eldy, line 177: According to template in modulebuilder, the test on restrictArea is done at end of header.
The reason is that some tests on permission need the variable to be initialized before being checked.
On which field is the warning ?

PR #32340, opened 2024-12-12
FIX to have two lists with total on same screen

altairis-tof:


hregis:
@altairis-tof @altairis-noe where is ths function `getLabelSpecialCode` ?

altairis-noe:
@hregis in `functions.lib.php` (line 11854)

PR #32346, opened 2024-12-13
FIX #32339 Delete a loan settlement is partial

### LINKED ISSUE #32339: Incoherence in loan payment management
### Bug

On module loan (htdocs/loan/list.php) , deleting a payment causes issues :

1. The bank line still exists after deleting the payment. Should it still exist, or should it be deleted with the payment loan ?
2. The bank line is not linked to the payment anymore, but is still linked to the loan. As a consequence :
    - on the loan screen, the payment is not shown
    - but, if the line is reconciliated, the loan can't be deleted. Moreover, no error is displayed (see PR #32338). The user has no clue on what is going on. Even if there was an error, the user can't navigate from the loan to the faulty bank line, so they can't fix the problem.

### Dolibarr Version

18.0, develop

### Environment PHP

any

### Environment Database

any

## To reproduce

- create a loan
- create a payment for this loan.
    - Observe that this has created a bank line. (expected behavior)
- delete the payment
- navigate to the bank line.
    - Observe that it is still linked to the loan
- navigate to the loan.
    - Observe that there is no payment shown, so the user can't navigate to the related bank line.
- Try to delete it.
    - Observe that it can't be deleted.

## Expected behavior

I think that, when we delete a payment loan, the related bank line should be also deleted. And if the bank line is reconciliated, one should not be able to delete the payment.

### Attached files

_No response_

aspangaro:
If you delete a payment of a loan, bank part of the payment is still present but not the loan part payment...

### DIFF
--- a/htdocs/loan/class/paymentloan.class.php
+++ b/htdocs/loan/class/paymentloan.class.php
@@ -369,7 +369,6 @@ class PaymentLoan extends CommonObject
 			}
 			$this->db->rollback();
 			return -1 * $error;
-
 		} else {
 			$this->db->commit();
 			return 1;
--- a/htdocs/loan/class/paymentloan.class.php
+++ b/htdocs/loan/class/paymentloan.class.php
@@ -1,7 +1,7 @@
 <?php
-/* Copyright (C) 2014-2018  Alexandre Spangaro   <aspangaro@open-dsi.fr>
- * Copyright (C) 2015-2018  Frederic France      <frederic.france@netlogic.fr>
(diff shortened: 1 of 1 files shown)


thomas-Ngr:
I confirm that thix fixes issue https://github.com/Dolibarr/dolibarr/issues/32339. Thanks @aspangaro !

aspangaro:
@lvessiller-opendsi / @rycks PR to merge, thanks !

lvessiller-opendsi (approved):

PR #32382, opened 2024-12-16
FIX broken feature, wrong GETPOSTINT parameter

hregis:


htdocs/core/lib/files.lib.php
```diff
@@ -2747,12 +2747,9 @@ function dol_check_secure_access_document($modulepart, $original_file, $entity,
...
 		return 'ErrorBadParameter';
 	}
+	// entity can't be empty
 	if (empty($entity)) {
-		if (!isModEnabled('multicompany')) {
-			$entity = 1;
-		} else {
-			$entity = 0;
-		}
+		$entity = $conf->entity;
```
eldy, line 2752: This looks to introduce bugs.
When modulepart is "medias", by definition it means it is the "public" directory (access by public with no login, no session and public for everybody). So there is only one "medias" directory (same for all entities). So for this special value of modulepart=medias, $entity must always be 1 and not $conf->entity.

Also when multicompany is on and a document is shared between entity 2 and 3 (for example a proposal), the PDF was createdand stored into entity 2. When we are in entity 3, we see the document and the PDF of entity 2 (because shared), but when we click on the link to PDF, it will try to download the file from the directory of entity 3 instead of 2. Did i miss something ?

hregis: @eldy ok, the problem was elsewhere, look my last fix, thanks

eldy:
@hregis i have complete with this
$entity = GETPOSTINT('entity') ? GETPOSTINT('entity') : $conf->entity;

still ok for you ?

hregis:
@eldy maybe you lose the possibility to have "entity = 0" in "dol_check_secure_access_document" no ?

eldy:
> @eldy maybe you lose the possibility to have "entity = 0" in "dol_check_secure_access_document" no ?

I think you are right, i reverted my addition.

PR #32414, opened 2024-12-18
FIX: status ticket update for new message

altairis-noe:
# FIX: status ticket update for new message
There is a parameter in Ticket module that allows to set a status when we send a message from ticket. However, this shouldn’t apply when said message is private.
The current test to determine whether our message is private… is always false, because the private message option is a checkbox, an not a text content. This should fix the issue.


lvessiller-opendsi (approved):

altairis-noe:
Hey there @eldy, is there a problem that would prevent merging this ?

rycks (approved):
sorry for the deleay, @altairis-noe  i think that's the reason for the delay !

ok for me !

altairis-noe:
> sorry for the deleay, @altairis-noe i think that's the reason for the delay !
> 
> ok for me !

\o/, thx @rycks and @eldy !

PR #32415, opened 2024-12-18
FIX : Removed display of login input field when main_authentication == "googleoauth"

Gouttfi:
# FIX : Removed display of login input field when main_authentication == "googleoauth"

When the user edit `htdocs/conf/conf.php` and set variable `$dolibarr_main_authentication` to `googleoauth`, all fields relative to login/password, forgotten password and login button should disapear, but actually, the `username` field remain visible and alone
image

This commit fix this behavior and hide the `username` field with the other fields relative to default Dolibarr authentication system, only if `$dolibarr_main_authentication == 'googleoauth'`


hregis:
@Gouttfi @eldy Dolibarr can use different method of authentication ! 

eg `$dolibarr_main_authentication='googleoauth,ldap,dolibarr';`

Maybe use a constant instead ?

eldy:
> @Gouttfi @eldy Dolibarr can use different method of authentication !
> 
> eg `$dolibarr_main_authentication='googleoauth,ldap,dolibarr';`
> 
> Maybe use a constant instead ?

@hregis this should not be a problem the test is on == "googleoauth", so hide the login only when using only oauth. When we use other method, the field remains visible. What we want.

PR #32421, opened 2024-12-19
FIX #32402 Social Contribution - Update - Drop the attached employee

### LINKED ISSUE #32402: Employee is dropped on social update
### Bug

When I update a social expense, at the moment I saved, the employee input is blanked.

### Dolibarr Version

20.0.0

### Environment PHP

8.2.25

### Environment Database

PostgreSQL

Create a social expense with an employee, clik save.
Then edit the social expense and save.

### Attached files

_No response_

aspangaro:
Create a social contribution with an employee attached.

Update this social contribution & you dropped the employee attached. You need to redefined employee previously attached.

Solution : remove the fk_user update in the update action
The user can nevertheless be updated with the dedicated setfk_user

image


eldy:
Can you confirm you can still modify the user ?

aspangaro:
Yes, with my fix, you can always defined the user with the method setfk_user

PR #32423, opened 2024-12-19
FIX : ref should be UNIQUE for dolibarr objects

altairis-tof:


htdocs/core/lib/modulebuilder.lib.php
```diff
@@ -411,7 +411,7 @@ function rebuildObjectSql($destdir, $module, $objectname, $newmask, $readdir = '
 		foreach ($object->fields as $key => $val) {
 			$i++;
 			if (!empty($val['index'])) {
-				$texttoinsert .= "ALTER TABLE llx_".strtolower($module).'_'.strtolower($objectname)." ADD INDEX idx_".strtolower($module).'_'.strtolower($objectname)."_".$key." (".$key.");";
+				$texttoinsert .= "ALTER TABLE llx_" . strtolower($module) . '_' . strtolower($objectname) . " ADD " . ($key == 'ref' ? "UNIQUE " : "") .  "INDEX idx_" . strtolower($module) . '_' . strtolower($objectname) . "_" . $key . " (" . $key . ($key == 'ref' && array_key_exists('entity', $object->fields) ? ", entity" : "") . ");";
```
eldy, line 414: Can you also rename the idx into uk ? when unique key we use a different prefix for name of indexes.

altairis-tof: @eldy ok it is done.

htdocs/core/class/commonobject.class.php
```diff
@@ -10378,6 +10378,15 @@ public function createCommon(User $user, $notrigger = 0)
 			}
 			$key_fields = $this->fields[$key];
 
+			// Check ref
+			if ($key == 'ref' && !empty($values['ref'])) {	// We check that ref is not already used
```
eldy, line 10382: Surely a bug here: We add a { and we add another one line 10384 but there is only one closing added.

PR #32431, opened 2024-12-19
FIX Bad GETPOSTINT for document preview

atm-lucasmantegari:
# FIX - Misuse GETPOSTINT param 
There was a misuse of the second parameter of a GETPOSTINT, here you put $conf->entity as the second parameter, which changed the method used depending on the entity you were on. which made


hregis:
@atm-lucasmantegari thanks but it's already in this PR : https://github.com/Dolibarr/dolibarr/pull/32382

PR #32439, opened 2024-12-20
New : Add hook

thersane-john:
# NEW hooks

Note that for theLineView_BeforeProduct object, there is no replacement option, because at this stage, it can be considered more of a tpl substitution and more appropriate .


frederic34:
for phpstan errors https://github.com/Dolibarr/dolibarr/pull/32440

PR #32456, opened 2024-12-23
Fix email collector datetimelast ok and email fetching

Hystepik:
Fix a bug where email collector does not update datelastok to the last email date when there are still some email to get.
Fix a bug where email was fetch in the bad order which make a broken timeline if a email chain is bigger then the number of mail to fetch


htdocs/emailcollector/class/emailcollector.class.php
```diff
@@ -1690,7 +1691,7 @@ public function doCollectOneCollector($mode = 0)
 				if ($mode > 0) {
 					$Query->leaveUnread();
 				}
-				$arrayofemail = $Query->limit($this->maxemailpercollect)->setFetchOrder("asc")->get();
+				$arrayofemail = $Query->limit($this->maxemailpercollect)->setFetchOrder("desc")->get();
```
eldy, line 1694: We must keep the order "asc" because we must process the email from the older one to the new one. In same order than they are received.

Hystepik: Yes as said in the description if we set setFetchOrder as asc the mail are not proceed the same way.
For exmple we want 2 mails to bee proceed a mail chain  of 10 mails is sent
* if we keep asc the last 2 mails of the chain mail are proceed first because in this case 'asc' means we start at the end of the mail inbox
* if we set to desc the 2 first mails are proceed because desc means we start from top of inbox.

And if we proceed the 10 emails of the mail chain then if we set desc or asc is the same

eldy: Strange. I have got the opposite :

I have 1 email on 19 dec and another one of 20 dec.

If i set limit to 10, the result is the same, i agree.

But if i set limit to 1, i get the email of the 19 dec when sort is asc, and i get the 20 dec if sort is asc.
So on my side, I need asc.

According to this https://github.com/Webklex/php-imap/issues/71, the sort is to define which bloc to receive when the limit is spliting the result into several blocks. According to exemple in this issue, when whe use the sort desc, we got the last recent block. And we want the older block.

So my result matches the explanation and the example of weklex. If max limit is higher than the number of results, then the sort has no effect. But if the max limit is lower, then we should use asc to get the block that is lower so older emails (the date that is the lower).

What weblex and its example says is that, inside a bloc, the sorting is always asc (great this is what we want), so trouble is only to get the oldest bloc when number of record is higher than limit.

We may see each other if

PR #32481, opened 2024-12-29
NEW : hook for adding params for credit notes creation

Bahfirilyes:
Fix:hook for adding params for credit notes creation


eldy:
Adding a hook is not a fix but an additon of a new possible baheviour. You can prefix your PR label with "NEW: ..." instead of "FIX: ..."

PR #32483, opened 2024-12-29
Fix: add view code in case of hook invocation

Bahfirilyes:
Fix: add view code in case of hook invocation


eldy:
We can't output code at this position because we are not in the view part but into the action part (so only non output code can be executed).

PR #32484, opened 2024-12-29
NEW : hook for payment actions

Bahfirilyes:
NEW : hook for payment actions


eldy:
Adding a hook is not a fix but an additon of a new possible behaviour. You can prefix your PR label with "NEW: ..." instead of "FIX: ..."

PR #32485, opened 2024-12-29
NEW : hook for pay.php actions

Bahfirilyes:
NEW : hook for pay.php actions


eldy:
Adding a hook is not a fix but an additon of a new possible baheviour. You can prefix your PR label with "NEW: ..." instead of "FIX: ..."

PR #32488, opened 2024-12-30
# FIX|Fix # Special Character Not Allowed for Extrafields

atm-corentin:
# FIX|Fix # Special Character Not Allowed for Extrafields

I got this error message when I added this extrafield:

 **'Bad string syntax to evaluate (found chars that are not chars for a simple clean eval string): $object->array_options['options_anciennete'] % 72 <= 74 ? 'Yes' : 'No'** 

This is because the character '<' was not listed in the allowed special characters, so I added it


eldy:
The character < is voluntarly exlcuded because it introduces a lot of security vulnerability.
You can invert the 2 parameters to use >= instead of <=

The message to suggest this when we found < can be enhanced in this direction.

PR #32489, opened 2024-12-30
Fix phpstan for color on assigned event

lvessiller-opendsi:
Fix phpstan for color on assigned event


lvessiller-opendsi:
@mdeweerd 
I updated from develop.
Could you help me, the error "Property DolEditor::$content (string) on left side of ?? is not nullable." is still here

mdeweerd:
> I updated from develop. Could you help me, the error "Property DolEditor::$content (string) on left side of ?? is not nullable." is still here
I was looking for this comment in the other PR where I replied...

PR #32490, opened 2024-12-30
Fix phpstan doleditor content cannot be nullable

lvessiller-opendsi:
Fix phpstan doleditor content cannot be nullable


hregis:
@lvessiller-opendsi nullable or not nullable, that is the question !

`htmlspecialchars(): Passing null to parameter #1 ($string) of type string is deprecated`

`Property DolEditor::$content (string) on left side of ?? is not nullable.`

hregis:
phpstan renvoi parfois des incohérences je trouve !!!

mdeweerd:
> phpstan renvoi parfois des incohérences je trouve !!!

Le PHPdoc dit que cela ne peut pas être null, mais en même temps la valeur n'est pas initialisé avec certitude à qqchse de non-null.
Perso je prend ces messages comme indicateur: donc je lis qu'il faut soit initialiser $this->content à '' par exemple, soit il faut dire que c'est un `?string` dans son PHPdoc.  Souvent rendre le code cohérent c'est impacter positivement l'analyse statique et ici éviter tout simplement le besoin de convertir un null possible en string.

lvessiller-opendsi:
According to me, even if the "content" is initialized to empty string '', the "content" variable can be reset to NULL value (it's a public attribute of "DolEditor" class). This attribute mustn't be NULL and set by default to empty string ''.
PHP Doc is good but it misses a (string) cast if there is an unset of content variable (set to NULL). So the cast into (string) do this job and the first parameter of "htmlspecialchars" is now a string. In all cases, the "content" variable is now a string.

mdeweerd:
The static analysis is there to detect that the attribute is not set to null elsewhere.

So if the assumption has to be that the attribute can be set to null in some other code, the phpdoc type must be updated to reflect this.

By doing so, the static analysis will flag any location where this assumption is not handled by a proper cast or test.

lvessiller-opendsi:
> The static analysis is there to detect that the attribute is not set to null elsewhere.
> 
> So if the assumption has to be that the attribute can be set to null in some other code, the phpdoc type must be updated to reflect this.
> 
> By doing so, the static analysis will flag any location where this assumption is not handled by a proper cast or test.

The PHP Doc for "content" attribute is set to string. 
So what do you suggest to pass CI error ? (Property DolEditor::$content (string) on left side of ?? is not nullable)
Should I remove the cast to string line ?

mdeweerd:
> The PHP Doc for "content" attribute is set to string.
> So what do you suggest to pass CI error ? (Property DolEditor::$content (string) on left side of ?? is not nullable)
> Should I remove the cast to string line ?

In this case the PHPdoc could be '?string' indicating that the value is nullable.
This may reveal code locations where the code supposes the value is a string and where additional casts or tests are needed.

lvessiller-opendsi:
> > The PHP Doc for "content" attribute is set to string.
> > So what do you suggest to pass CI error ? (Property DolEditor::$content (string) on left side of ?? is not nullable)
> > Should I remove the cast to string line ?
> 
> In this case the PHPdoc could be '?string' indicating that the value is nullable. This may reveal code locations where the code supposes the value is a string and where additional casts or tests are needed.

Please, could you explain to me, how you removed this check without changing the "PHP Doc" of the attribute "content" in DolEditor class file ?

mdeweerd:
> Please, could you explain to me, how you removed this check without changing the "PHP Doc" of the attribute "content" in DolEditor class file ?

Apparently the notice is gone with '?string':
{EA8941CD-D35E-424D-A205-64682B9DD882}

lvessiller-opendsi:
> > Please, could you explain to me, how you removed this check without changing the "PHP Doc" of the attribute "content" in DolEditor class file ?
> 
> Apparently the notice is gone with '?string': {EA8941CD-D35E-424D-A205-64682B9DD882}


Thanks for your help

PR #32493, opened 2024-12-30
Qual: Fix technical debt creep and update baseline

mdeweerd:
# Qual: Fix technical debt creep and update baseline

This fixes most of the new technical debt that was added with regards to my latest baseline.

It also updates the baseline.


mdeweerd:
@eldy This also fixes some of the notices in develop.  #32490 and #32489 should fix a few more.

PR #32494, opened 2024-12-31
FIX glob is better for search files with wilcard + avoid warning

hregis:


htdocs/install/upgrade2.php
```diff
@@ -4133,7 +4133,7 @@ function migrate_delete_old_files($db, $langs, $conf)
 
 	foreach ($filetodeletearray as $filetodelete) {
 		//print '<b>'DOL_DOCUMENT_ROOT.$filetodelete."</b><br>\n";
-		if (file_exists(DOL_DOCUMENT_ROOT.$filetodelete) || preg_match('/\*/', $filetodelete)) {
+		if (count(glob(DOL_DOCUMENT_ROOT.$filetodelete)) > 0) {
```
eldy, line 4136: Glob may have trouble on windows. 
For example c:\mydir is not found because \ is used as escapement.
May be we can invert the 2 tests ?

hregis: @eldy arf... Windows ! ;-) it's ok, thanks

PR #32496, opened 2024-12-31
FIX units used scale and scale is an integer

hregis:


erikvanberkum:
@hregis this is not fixing it for me. The fix needs to be applied to: public function update($id, $user, $notrigger = 0, $action = 'update', $updatetype = false) your fix is applied to the public function fetch($id = 0, $ref = '', $ref_ext = '', $barcode = '', $ignore_expression = 0, $ignore_price_load = 0, $ignore_lang_load = 0)

Can you check

hregis:
@erikvanberkum @eldy This first fix allows not to lose the value selected in the "selectMeasuringUnits()" function when the value is empty ("scale" field of the measurement units dictionary)
A deep correction would be needed because the processing of these measurement units were treated as strings while they are integers.

PR #32505, opened 2025-01-02
FIX send email to assigned user on ticket create

ghost:
# FIX send email to assigned user on ticket create
Fixed send email to assigned user. This is a backport of #31928


rycks (approved):
ok for me, thanks for the contrib, that's in prod on my dolibarr :)

PR #29212, opened 2024-04-03
Add entity to bookcal

G-ladal:
# New prepare DB for bookcal compatibility with multicompany
Core database miss "entity" for bookcal tables, it will allow compatibility with MC (in following pushes).

PR #32513, opened 2025-01-03
FIX: allow bookcal display on calendar list

G-ladal:
disable checks on "filtert" (default does not allow bookcal from other entities)
fix : add check on entites for bookcals

# FIX bookcal : FIX : check on entities + not displaying calendar if multicompany ON
In multicompany context, we have to check on entities so only display shared agenda/bookcal AND have to disable checks on "filtert" (default only allow bookcal from same user, but is not altered by bookcal)

### DIFF
--- a/htdocs/bookcal/availabilities_list.php
+++ b/htdocs/bookcal/availabilities_list.php
@@ -263,6 +263,9 @@ $sql = preg_replace('/,\s*$/', '', $sql);
 $sqlfields = $sql; // $sql fields to remove for count total
 
 $sql .= " FROM ".MAIN_DB_PREFIX.$object->table_element." as t";
+if ($object->ismultientitymanaged == 1  || $object->ismultientitymanaged != '') { // value is fk_bookcal_calendar@bookcal_calendar
+ 	$sql .= ", ".MAIN_DB_PREFIX."bookcal_calendar as bc";
+}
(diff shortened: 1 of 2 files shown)


G-ladal:
phpstan error on code i didn't change, I can correct it if someone can pin to a standard correction (cast?)

hregis:
Module builder add more and more code with no sense !! getEntity($object->element) doesn't make sense, object is "availabilities" and "bookcal_availabilities" table has no entity field !!

hregis:
@eldy ce n'est que mon opinion, mais un module fait avec Module Builder et qu'on veut intégrer a Dolibarr devrait être nettoyé, voir réécrit correctement !

G-ladal:
> @eldy ce n'est que mon opinion, mais un module fait avec Module Builder et qu'on veut intégrer a Dolibarr devrait être nettoyé, voir réécrit correctement !

i just don't have time to clean/recreate it from scratch, so try fixing what i can for my own needs ! (working in production env)

G-ladal:
> Module builder add more and more code with no sense !! getEntity($object->element) doesn't make sense, object is "availabilities" and "bookcal_availabilities" table has no entity field !!

it doesn't but bookcal_calendar have it, see my old PR (and Eldy comments) on it : #29212

G-ladal:
fixed : $object->element to 'calendar', in getEntity

PR #32515, opened 2025-01-03
FIX ODT substitution when many HTML tags in string

thomas-Ngr:
# FIX ODT substitution : if a string contains many times the same HTML tag, only the first opening tag and the last closing tag are used.

## To reproduce : 

- create a translation string containing two times the same HTML tag, : `I <b>contain</b> two <b>bold</b> tags`
two_strong_tags_string

- create an ODT template with this translation string and upload it for any object
two_strong_tags_tempalte

- generate the ODT
- see that the string is incorrectly rendered
two_strong_tags_result

This PR fixes this behavior and adds a test to lock the correct behavior.

As a side enhancement, the PR reverses the orders of parameters in test assertEquals() method since they were not in the right order.

### DIFF
--- a/htdocs/includes/odtphp/odf.php
+++ b/htdocs/includes/odtphp/odf.php
@@ -46,8 +46,8 @@ class Odf
 	public $userdefined=array();
 
 	const PIXEL_TO_CM = 0.026458333;
-	const FIND_TAGS_REGEX = '/<([A-Za-z0-9]+)(?:\s([A-Za-z]+(?:\-[A-Za-z]+)?(?:=(?:".*?")|(?:[0-9]+))))*(?:(?:\s\/>)|(?:>(.*)<\/\1>))/s';
-	const FIND_ENCODED_TAGS_REGEX = '/&lt;([A-Za-z]+)(?:\s([A-Za-z]+(?:\-[A-Za-z]+)?(?:=(?:".*?")|(?:[0-9]+))))*(?:(?:\s\/&gt;)|(?:&gt;(.*)&lt;\/\1&gt;))/';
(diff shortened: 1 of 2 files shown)


htdocs/includes/odtphp/odf.php
```diff
@@ -46,8 +46,8 @@ class Odf
 	public $userdefined=array();
 
 	const PIXEL_TO_CM = 0.026458333;
-	const FIND_TAGS_REGEX = '/<([A-Za-z0-9]+)(?:\s([A-Za-z]+(?:\-[A-Za-z]+)?(?:=(?:".*?")|(?:[0-9]+))))*(?:(?:\s\/>)|(?:>(.*)<\/\1>))/s';
-	const FIND_ENCODED_TAGS_REGEX = '/&lt;([A-Za-z]+)(?:\s([A-Za-z]+(?:\-[A-Za-z]+)?(?:=(?:".*?")|(?:[0-9]+))))*(?:(?:\s\/&gt;)|(?:&gt;(.*)&lt;\/\1&gt;))/';
+	const FIND_TAGS_REGEX = '/<([A-Za-z0-9]+)(?:\s([A-Za-z]+(?:\-[A-Za-z]+)?(?:=(?:".*?")|(?:[0-9]+))))*(?:(?:\s\/>)|(?:>(((?!<\1>).)*)<\/\1>))/s';
+	const FIND_ENCODED_TAGS_REGEX = '/&lt;([A-Za-z]+)(?:\s([A-Za-z]+(?:\-[A-Za-z]+)?(?:=(?:".*?")|(?:[0-9]+))))*(?:(?:\s\/&gt;)|(?:&gt;(((?!&lt;\1&gt;).)*)&lt;\/\1&gt;))/';
```
rycks, line 50: note, FIND_ENCODED_TAGS_REGEX seems to be never used into all dolibarr source code, maybe a code cleanup @eldy to simplify the code ?

lvessiller-opendsi (approved):

htdocs/includes/odtphp/odf.php
eldy: cleaning code is a question related to develop branch (may be a cleanup is ok for develop, i don't know) but question is not relevant for v18.

PR #32520, opened 2025-01-04
FIX clean unique extrafields when create product combination

hregis:


hregis:
@eldy @mdeweerd maybe exclude "includes" directory for this pre-commit check ? this is external libraries

frederic34:
https://github.com/Dolibarr/dolibarr/pull/32522

hregis:
@frederic34 thank you !

mdeweerd:
Ignoring errors is not my recommended approach, especially in this case.

The issue occurs in pre-commit because it happens during the linting step which is the 'php' executable checking the php files for syntax.

{0D329E1E-711D-4C29-B0FB-F1EC540518C7}

So it's not about higher level incompatibilities but about language incompatibilities.

Therefore, we should be thankful for this notice and not ignore it, but fix it.  Either by adjusting the code or by upgrading the module.

Ignoring all the includes here is preventing the flow from detecting that the includes are really outdated to comply with the range of PHP versions that dolibarr wants to support.

PR #32522, opened 2025-01-04

frederic34:
it seems we can't choose to execlude `includes` dir

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -154,6 +154,7 @@ repos:
           ]
         pass_filenames: false # Run on all files
       - id: php-lint
+        stages: [manual]
       - id: php-stan
         stages: [manual]
         files: \.(php)$
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -154,7 +154,8 @@ repos:
           ]
         pass_filenames: false # Run on all files
       - id: php-lint
-        stages: [manual]
+        exclude:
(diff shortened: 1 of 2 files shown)


mdeweerd:
As explained in https://github.com/Dolibarr/dolibarr/pull/32520#issuecomment-2571283835 , I do not recommend ignoring the the lint step.

EDIT: and the php-lint step does allow to exclude specific files or even directories - exclusions are specified as a regex - there are other examples in the pre-commit configuration;

frederic34:
I agree, it was to avoid CI fails for other PR

frederic34:
for curly brace in tcpdf it's fixed since Feb 2020 in official lib 

https://github.com/tecnickcom/TCPDF/commit/c09bef348d0d6e853b473b76f5dd2d00f07b2488#diff-43dfa6c00b99a4d77d8408e159baef42325b773421d4aa45066b57c329727c13

frederic34:
in branch 20.0 it's 6.3.2 with curly braces... in develop it's 6.7.5
need to check var_dumper lib

mdeweerd:
Ok, I thought this was about the dev branch.
This probably occurred after the automatic change of the runner image to ubuntu24.04.1 - maybe it should be forced to the previous LTS or we should setup the desired php version in the flow.

frederic34:
the problem occur in dev branch vwith symfony/var-dumper

mdeweerd:
This was only for 20.0 but propagated to develop!

PR #32523, opened 2025-01-04
FIX $this->origin_object can not be instanceof CommandeFournisseur if it is already an instanceof CommonObject

hregis:


htdocs/reception/class/reception.class.php
```diff
@@ -729,11 +729,7 @@ public function getStatusDispatch()
 				$this->fetch_origin();
 				if ($this->origin_object instanceof CommonObject && empty($this->origin_object->lines)) {
 					$res = $this->origin_object->fetch_lines();
-					if ($this->origin_object instanceof CommandeFournisseur) {
-						$this->commandeFournisseur = $this->origin_object;	// deprecated
-					} else {
-						$this->commandeFournisseur = null;	// deprecated
-					}
+					$this->commandeFournisseur = null;	// deprecated
```
eldy, line 732: What if $this->origin_object is a CommandeFournisseur object ?

hregis:
@eldy ready

PR #32536, opened 2025-01-05
Qual: Fix phan notices

mdeweerd:
# Qual: Fix phan notices

Fix a set of phan notices.

### DIFF
--- a/htdocs/core/customreports.php
+++ b/htdocs/core/customreports.php
@@ -1,6 +1,6 @@
 <?php
 /* Copyright (C) 2020-2024 Laurent Destailleur  <eldy@users.sourceforge.net>
- * Copyright (C) 2024		MDW							<mdeweerd@users.noreply.github.com>
+ * Copyright (C) 2024-2025	MDW						<mdeweerd@users.noreply.github.com>
  * Copyright (C) 2024		Frédéric France			<frederic.france@free.fr>
  *
  * This program is free software; you can redistribute it and/or modify
@@ -40,6 +40,11 @@
  * @var User $user
  */
 
+// Initialise values
+$search_xaxis = array();
+$search_groupby = array();
+$tabfamily = null;
+
 if (!defined('USE_CUSTOM_REPORT_AS_INCLUDE')) {
 	require '../main.inc.php';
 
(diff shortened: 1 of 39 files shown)


htdocs/core/customreports.php
```diff
@@ -56,14 +61,10 @@
 	//$search_xaxis = GETPOST('search_xaxis', 'array');
 	if (GETPOST('search_xaxis', 'alpha') && GETPOST('search_xaxis', 'alpha') != '-1') {
 		$search_xaxis = array(GETPOST('search_xaxis', 'alpha'));
-	} else {
```
eldy, line 59: Why removing this else ?

eldy: When the customreport.php page is called as a standalone page, i do not see at wich line the $search_xaxis is already initialized. 
Can you point where the init is done ?

mdeweerd: In the "else" branch where this value is not set, we have explicitly this comment:
```
    // $search_measures, $search_xaxis or $search_yaxis may have been defined by the parent.
```

It says 'may be', so the code the follows needs to be protected from unset variables.

We can observe this protection:
https://github.com/Dolibarr/dolibarr/blob/28a955204f963054a93e39f32f8e3329efd72d30/htdocs/core/customreports.php#L410-L413

https://github.com/Dolibarr/dolibarr/blob/28a955204f963054a93e39f32f8e3329efd72d30/htdocs/core/customreports.php#L410-L413

Therefore it is useless to initialize variables to "empty values" that are protected with isset anyway.

In case the protection is unsufficient, the static analysis should kick in and report it.

eldy: I think you are right.
It's true. value can be set by parent when the page is called as an include. So should not be erased in this case.
When called as a standalone page, it may be initialized but if there is a protection later, we can avoid it to simplify code.
Will wait CTI green to merge with your removal.

mdeweerd:
@eldy ci has been green several times now and I have a few PRs to submit pending that fix notices for another 125 files.
I am continuously rebasing this PR due to the LIFO approach to PRs, and stalling the 3 pending ones for the same reason.

If you prefer, I can make about 164 PRs for all the changes I have pending which I am sure is more effort on my end, but also on your end.

What is the best way to proceed?

eldy:
> @eldy ci has been green several times now and I have a few PRs to submit pending that fix notices for another 125 files.
> I am continuously rebasing this PR due to the LIFO approach to PRs, and stalling the 3 pending ones for the same reason.
> 
> If you prefer, I can make about 164 PRs for all the changes I have pending which I am sure is more effort on my end, but also on your end.
> 
> What is the best way to proceed?

It is not more effort on my side to have more smaller PR, it is same for me (even a little bit easier for me).
But what is more important is Your time, not mine. Because on ci and quality improvement, there is too few dev working on this, so i try to give priority on them but i can't postponed too much other pr either.
165 pr may be too much for you, but 1 may make a too large pr too. May be you can split in 10-15 pr ?
As your pr are always related to 1 topic (code quality), it may be ok to have larger pr for that to reduce nb. Just do your best to have smaller "as possible" but "as possible" means also "by keeping your time optimized". Difficult for me to provide another direction than this....

mdeweerd:
@eldy Thank your for your feedback.
I try to limit the size of the PRs to about 30 files/200-250 phan notices, but some fixes drag in other files so that varies.
I then run phan locally to ensure no new notices appeared and that the files I worked on are as clean as my goal was.

Smaller means more local phan runs, baselines to update and merge, ... .

I'll think a bit more about this to better optimize for you too.  Faster integration is also less work for me.

PR #32542, opened 2025-01-06
FIX #21294 Stock import sql query

### LINKED ISSUE #21294: Import stock fails
### Bug

When running a stock import via the import module the following error is thrown up.

Error running final request: UPDATE llx_product as p SET p.stock

This bug only seems to be present when running a postgres database as I had tested on a mariaDB install and this worked well. 

I managed to remove the error by commenting out line 417 of modStock.class.php or removing the p. prefix from stock - see below.

Before:
'UPDATE '.MAIN_DB_PREFIX.'product as p SET p.stock = (SELECT SUM(ps.reel) FROM '.MAIN_DB_PREFIX.'product_stock ps WHERE ps.fk_product = p.rowid);'

After:
Change to 'UPDATE '.MAIN_DB_PREFIX.'product as p SET stock = (SELECT SUM(ps.reel) FROM '.MAIN_DB_PREFIX.'product_stock ps WHERE ps.fk_product = p.rowid);'

### Environment Version

15.02

### Environment OS

Linux dolibarr 5.10.0-14-amd64 #1 SMP Debian 5.10.113-1 (2022-04-29) x86_64

### Environment Web server

Apache/2.4.53 (Debian)

### Environment PHP

7.4.28

### Environment Database

11.8 (Debian 11.8-1.pgdg100+1)

### Environment URL(s)

_No response_

Importing a CSV via the import Stocks and location (warehouse) of products should allow me to import stocks.

It throws an error that "p" is not a valid column name. Error running final request: UPDATE llx_product as p SET p.stock

When line 417 of modStock.class.php = 
'UPDATE '.MAIN_DB_PREFIX.'product as p SET p.stock = (SELECT SUM(ps.reel) FROM '.MAIN_DB_PREFIX.'product_stock ps WHERE ps.fk_product = p.rowid);'

Stock import fails.

When changed to 

'UPDATE '.MAIN_DB_PREFIX.'product as p SET stock = (SELECT SUM(ps.reel) FROM '.MAIN_DB_PREFIX.'product_stock ps WHERE ps.fk_product = p.rowid);'

I expect the stock import feature to import stock - this only seems to affect postgres databases.

### Attached files

_No response_

Al-thi:
# FIX #21294 Import stock fails

To respect the PostgreSQL update statement syntax, cf. https://www.postgresql.org/docs/17/sql-update.html

```
column_name
    The name of a column in the table named by table_name. The column name can
    be qualified with a subfield name or array subscript, if needed. Do not
    include the table's name in the specification of a target column — for
    example, UPDATE table_name SET table_name.col = 1 is invalid.
```

### DIFF
--- a/htdocs/core/modules/modStock.class.php
+++ b/htdocs/core/modules/modStock.class.php
@@ -429,7 +429,7 @@ class modStock extends DolibarrModules
 		);
 		$this->import_updatekeys_array[$r] = array('ps.fk_product'=>'Product', 'ps.fk_entrepot'=>"Warehouse");
 		$this->import_run_sql_after_array[$r] = array(    // Because we may change data that are denormalized, we must update dernormalized data after.
-			'UPDATE '.MAIN_DB_PREFIX.'product as p SET p.stock = (SELECT SUM(ps.reel) FROM '.MAIN_DB_PREFIX.'product_stock ps WHERE ps.fk_product = p.rowid);'
+			'UPDATE '.MAIN_DB_PREFIX.'product as p SET stock = (SELECT SUM(ps.reel) FROM '.MAIN_DB_PREFIX.'product_stock ps WHERE ps.fk_product = p.rowid);'
 		);
 	}
 


rycks (approved):
ok for me, checked with mysql too, there is no impact

`
UPDATE llx_product as p SET stock = (SELECT SUM(ps.reel) FROM llx_product_stock ps WHERE ps.fk_product = p.rowid)
`

lvessiller-opendsi (approved):

PR #32543, opened 2025-01-06
Backport Add filter from date to date into action/evt list

uvaldenaire-opendsi:
Backport of https://github.com/Dolibarr/dolibarr/pull/18917/files

Currently on comm/action/list.php with list view we can't filter actions by date.


eldy:
Only bug fix are allowed into backport (security / stability / integrity).
Backport of added features is not allowed.
"we can't" does not mean it is a bug. 
It means the feature was not implementedin this version and we need a higher version to get the new behaviour.

PR #32552, opened 2025-01-07
FIX|Fix # Update TimeSpent Issue

atm-corentin:
# FIX|Fix #Issue when you try to update the TimeSpent

I noticed that when I try to update the time on a task, it deletes the current timespent line and then adds a new one. Instead of doing that, we can use the appropriate method, updateTimeSpent, which makes more sense.

### DIFF
--- a/htdocs/projet/tasks/time.php
+++ b/htdocs/projet/tasks/time.php
@@ -286,9 +286,6 @@ if (($action == 'updateline' || $action == 'updatesplitline') && !$cancel && $us
 			$object->fetchTimeSpent(GETPOST('lineid', 'int'));
 
 			$result = 0;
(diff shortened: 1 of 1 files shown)


eldy:
Does it work also when you use the action "updatesplitline" ?

atm-corentin:
@eldy I tried the action 'updatesplitline.' It seems it doesn't work properly without my fix on V21.

eldy:
> @eldy I tried the action 'updatesplitline.' It seems it doesn't work properly without my fix on V21.
> 
> 

My question was "does updatesplitline works after your fix" ?
So i understand that yes ?

atm-corentin:
@eldy yes, it works

PR #32557, opened 2025-01-07
fix join societe_commerciaux in societe list

uvaldenaire-opendsi:
In societe/list.php, when we filter the list of tiers with the sale representatives and when the hook printFieldListFrom is used, we get "no such field xxx in on clause" error.

This is the same problem as here: https://github.com/Dolibarr/dolibarr/pull/23941

To fix it i used INNER JOIN instead of comma join.


rycks:
that is a bug in 18.0 but in 20.0 the solution was different with an EXIST and a subselect request because we don't use data from the join ...

rycks (approved):
ok for me but will not applyed into dolibarr 20.0+ because solution was really different into 20.0

lvessiller-opendsi (approved):

PR #32563, opened 2025-01-07
FIX: product variants copy: also copy multiprice variations

marc-dll:
When conjointly using product variants et multiple price levels, when copying variants from a product A to a product B, the multiprice impact was not copied

Original product variant :
Capture d’écran de 2025-01-07 17-14-02

Copied product variant :
Capture d’écran de 2025-01-07 17-14-05


rycks:
PR code seems ok for me but i have to set up a dolibarr with product variants et multiple price levels to test it !

lvessiller-opendsi (approved):

PR #32578, opened 2025-01-08
Fix modulebuilder backward compatibility.

fappels:
To have a V21 build module also working on v20

~~On second thought, the return string to array change in V21 will cause trouble running v20- build modules on v21+.~~

@eldy Forgot to remove the nooutput parameter for testing.

Working on other backward compatibility solution.


fappels:
@eldy  Think current patch is most clean solution

PR #32579, opened 2025-01-08
Fix fatal error on sprintf on translate

Hystepik:
When we translate a string with '%)' in it the sprintf in trans function end with a fatal error saying that ')' isn't a valid specifier.

To prevent that  a test on this string is made and replace by __percent__ and reintegrated after the sprintf


htdocs/core/class/translate.class.php
```diff
@@ -699,6 +699,10 @@ public function trans($key, $param1 = '', $param2 = '', $param3 = '', $param4 =
 				$str
 			);
 
+			if (preg_match('/%\)/', $str)) {
+				$str = preg_replace('/%\)/', '__percent__)', $str);
+			}
+
```
mdeweerd, line 705: As for the reverse operation you could just do the str_replace directly without the preg_match.
The performance of str_replace is likely a bit better too.

Hystepik: thx it's done

htdocs/core/class/translate.class.php
```diff
@@ -708,6 +710,8 @@ public function trans($key, $param1 = '', $param2 = '', $param3 = '', $param4 =
 				}
 			}
 
+			$str = str_replace('__percent__', '%', $str);
```
eldy, line 713: May be we should restore the parenthesis ?

mdeweerd:
Not superimportant, but parenthesis if with an 'h' (https://dictionary.cambridge.org/fr/dictionnaire/anglais-francais/parenthesis) .

PR #32580, opened 2025-01-08
fix(ecm-files): missing src_object_type and src_object_id

atm-irvine:
# FIX|Fix Missing src_object_type and src_object_id
When I create an invoice and the PDF is automatically generated, in the ecm_files table it doesn't fill src_object_type and src_object_id.

### DIFF
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -313,14 +313,13 @@ function dol_dir_list_in_database($path, $filter = "", $excludefilter = null, $s
  */
 function completeFileArrayWithDatabaseInfo(&$filearray, $relativedir)
 {
-	global $conf, $db, $user;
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/files.lib.php
```diff
@@ -382,6 +381,8 @@ function completeFileArrayWithDatabaseInfo(&$filearray, $relativedir)
 				$ecmfile->label = md5_file(dol_osencode($filearray[$key]['fullname'])); // $destfile is a full path to file
 				$ecmfile->fullpath_orig = $filearray[$key]['fullname'];
 				$ecmfile->gen_or_uploaded = 'unknown';
+				$ecmfile->src_object_type = $object->element;
```
eldy, line 384: What happen if $object is not set or null  or an object with these properties ?

PR #32582, opened 2025-01-08
FIX compatibility with xdebug 3

hregis:


hregis:
@mdeweerd @frederic34 should we rename the` dolPrintHTMLForAttribute` function to `dolPrintHtmlForAttribute`?

frederic34:
it is to the commiter of https://github.com/Dolibarr/dolibarr/commit/c0250e4c9106b5c889e512a4771f0205d4f99b99 to fix that

mdeweerd:
I actually already prepared the branch, the pre-commit step was running while checking & updating the style.

I found pre-existing calls with HTML so change everything to that in #32583.

hregis:
@mdeweerd @frederic34 ok thanks

PR #32588, opened 2025-01-09
NEW add hook 'addMoreObjectBlock'

hregis:
FIX avoid to show html entities


hregis:
@mdeweerd heuuuu why ? `htdocs/user/card.php:2091: informations ==> information`

mdeweerd:
Informations does not exist in English - it's a french thing.  It's invariable in English: 'information'.

It's like toilets, in France we say that we go to the toilets, but elsewhere we go to the toilet.

hregis:
@mdeweerd Thank you for this English lesson 😉

hregis:
@eldy green green green... super green !!! 😄

htdocs/user/card.php
```diff
@@ -2043,9 +2043,9 @@
 			// Other form for user password
 			$parameters = array('valuetoshow' => $valuetoshow, 'caneditpasswordandsee' => $permissiontoeditpasswordandsee, 'caneditpasswordandsend' => $permissiontoeditpasswordandsend);
 			$reshook = $hookmanager->executeHooks('printUserPasswordField', $parameters, $object, $action); // Note that $action and $object may have been modified by hook
-			if ($reshook > 0) {
+			if ($reshook === 1) {
```
eldy, line 2046: The dev rules for hook are
Return 0 means the hook add action
Return > 0 means replace action
So we should have > 0 here

hregis: @eldy ok it's removed

PR #32604, opened 2025-01-09
Try to sync the pending changes

mikee2:
# FIX|Fix #2
This is a try to, again, sync the changes made years ago that never went into the main branch for whatever reason.

Pull request #1 is still ending and I do not know how to fix that. Please help with instructions.

Thanks.


eldy:
Did you send your changes into a PR ? Do you have the number ? 
You mention an issue # 1 and # 2 in your PR but they are not related to your code.

phpstan.neon.dist
```diff
@@ -15,19 +15,21 @@ parameters:
 	fileExtensions:
 		- php
 	paths:
-		- htdocs
-		- scripts
+		- dolibarr/htdocs
+		- dolibarr/scripts
+        - dolibarr/htdocs/core
```
eldy, line 20: You should not add this dir here, it is already include into dolibarr/htdocs.

mikee2: Those were suggestions by Git Copilot because I am were not able for many tries to make the pull request to get through the phpStan check. As I am not a very advanced user in Git, I did apply them blindly even though I also though that they were redundant.

I am trying to update the former code that is now not only obsolete, but wrong, and that I have to manually change every new version of Dolibarr is published making painful to apply updates. I know there are other users out there that also had issues with this code and I hope that, by applying the suggested changes, it will help them too.

mikee2:
> Did you send your changes into a PR ? Do you have the number ? You mention an issue # 1 and # 2 in your PR but they are not related to your code.

I am not quite sure. Maybe this is the number: 32604?

This what it show at the top:

Try to sync the pending changes #32604
[mikee2]wants to merge 21 commits into [Dolibarr:develop] from [mikee2:develop]

 As told I am not quite good at managing this git thing. But I am learning. Sorry.

eldy:
So is it a fix or enhancement ?
If fix , thanks to set description to describe how to reproduce the bug
If enhancement, thanks to set description of the enhancement.

mikee2:
It's a fix. The actual code that is distributed with the application does not work well. This fixes the existing issues and use the correct method of making substitutions as you suggested.

PR #32609, opened 2025-01-10
FIX #32477 Loan - Insurance amount need decimals

### LINKED ISSUE #32477: Bug: Inssurance of a loan not properly taken when there are decimals - Loan
### Bug

Version: 20.0.2

Hello,

I have created a loan and I had forgotten to add the price of the insurance of the loan.
So I edit the loan and change the value of the insurance by 12,70
image

After validation of the change, the insurance is 12,00€ instead of 12,70€
image

Also when I edit the loan there are two small problems:
- A dot is used to separate the decimals but when I'm not in edit mode it's a coma (coma is what I want in both cases)
- There are too many zeros for decimals, you can see on the screeshot 75989.000000, but 75989,00 should be expected.

Regards

### Dolibarr Version

20.0.2

### Environment PHP

_No response_

### Environment Database

_No response_

_No response_

### Attached files

_No response_

aspangaro:
#32477 

Without this fix, you can only specify a insurance amount without decimals. It's problematic

### DIFF
--- a/htdocs/loan/card.php
+++ b/htdocs/loan/card.php
@@ -106,7 +106,7 @@ if (empty($reshook)) {
 		if (!$cancel) {
 			$datestart = dol_mktime(12, 0, 0, GETPOSTINT('startmonth'), GETPOSTINT('startday'), GETPOSTINT('startyear'));
 			$dateend = dol_mktime(12, 0, 0, GETPOSTINT('endmonth'), GETPOSTINT('endday'), GETPOSTINT('endyear'));
-			$capital = price2num(GETPOSTFLOAT('capital'));
+			$capital = GETPOSTFLOAT('capital');
 			$rate = price2num(GETPOST('rate'));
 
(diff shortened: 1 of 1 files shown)


htdocs/loan/card.php
```diff
@@ -192,7 +192,7 @@
 				$object->capital = $capital;
 				$object->nbterm = GETPOSTINT("nbterm");
 				$object->rate = price2num(GETPOST("rate", 'alpha'));
-				$object->insurance_amount = price2num(GETPOSTINT('insurance_amount'));
+				$object->insurance_amount = price2num(GETPOSTFLOAT('insurance_amount'));
```
eldy, line 195: When using GETPOSTFLOAT, the price2num is not necessary. included into GETPOSTFLOAT.

aspangaro: Correct @eldy ! Thanks

PR #32611, opened 2025-01-10
FIX an admin can read all

hregis:


htdocs/user/bank.php
```diff
@@ -97,7 +97,7 @@
 // Ok if user->hasRight('salaries', 'readall') or user->hasRight('hrm', 'read')
 //$result = restrictedArea($user, 'salaries|hrm', $object->id, 'user&user', $feature2);
 $ok = false;
-if ($user->id == $id) {
+if ($user->id == $object->id || !empty($user->admin)) {
```
eldy, line 100: Admin means has permission to manage permission.
He should not be able to read salaries if he does not have permission to. Yes he can set permission to himself and remove them after because he is admin but this will be logged.

i think fix should be only 
if ($user->id == $object->id) {

hregis: @eldy yes but if "hrm", "salaries" and "expensereport" modules is not enabled, an admin can not access to this tabs !

hregis:
@eldy yes but if "hrm", "salaries" and "expensereport" modules is not enabled, an admin can not access to this tabs !

htdocs/user/bank.php
eldy: Yes, if module hrm, salaries and expenserport are disabled, an admin should not be able to access this tab. Like other.
For which use cas do you need to access the tab if the 3 modules are disabled ?

hregis: @eldy and what happen if the boss is an admin? 😅

hregis: @eldy if "hrm", "salaries" and "expensereport" is disable but if "holiday" is enable, the tab is visible but the page is "access denied"

eldy: > @eldy if "hrm", "salaries" and "expensereport" is disable but if "holiday" is enable, the tab is visible but the page is "access denied"

This is a different case. Bad consistency between test to show tab and test to access tab.
I pushed a fix.

hregis: Thanks

PR #32614, opened 2025-01-10
Replace js with css for third party types and membres

atm-etienne:
# Qual #[*Remove JS to add styles on third parties and membres types*]
As CSS is powerful, we should not use JS to add or remove styles according to a checked value of an input.
In this way, I removed JQuery scripts in third parties and members page, to replace it with some CSS.

The only difference is that I hid the default checkox, since the button colours are significant enough, and it gives a more modern look

image

### DIFF
--- a/htdocs/adherents/card.php
+++ b/htdocs/adherents/card.php
@@ -1030,40 +1030,18 @@ if (is_object($objcanvas) && $objcanvas->displayCanvasExists($action)) {
 		print "</td>\n";
 
 		// Morphy
-		$morphys = array();
-		$morphys["phy"] = $langs->trans("Physical");
-		$morphys["mor"] = $langs->trans("Moral");
-		$checkednature = (GETPOSTISSET("morphy") ? GETPOST("morphy", 'alpha') : $object->morphy);
+		$checkedNature = (GETPOSTISSET("morphy") ? GETPOST("morphy", 'alpha') : $object->morphy);
 		print '<tr><td class="fieldrequired">'.$langs->trans("MemberNature")."</td><td>\n";
-		print '<span id="spannature1" class="nonature-back spannature paddinglarge marginrightonly"><label for="phisicalinput" class="valignmiddle">'.$morphys["phy"].'<input id="phisicalinput" class="flat checkforselect marginleftonly valignmiddle" type="radio" name="morphy" value="phy"'.($checkednature == "phy" ? ' checked="checked"' : '').'></label></span>';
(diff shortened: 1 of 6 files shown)


htdocs/adherents/card.php
```diff
@@ -1030,40 +1030,18 @@ function initfieldrequired() {
...
-				</script>';
-		}
-
+		?>
+		<span class="nature paddinglarge marginrightonly">
+			<input type="radio" id="physical-input" name="morphy" value="phy" <?= $checkedNature == "phy" ? 'checked="checked"' : '' ?>>
+			<label for="physical-input"><?= $langs->trans("Physical") ?></label>
+		</span>
+		<span class="nature paddinglarge marginrightonly">
+			<input type="radio" id="moral-input" name="morphy" value="mor" <?= $checkedNature == "mor" ? 'checked="checked"' : '' ?>>
```
eldy, line 1041: Can you replace <?= with <?php echo
The tag <?= is often disabled on some secured php.

atm-etienne: The short echo tag `<?= ... ?>` can't be disabled 
What is disabled in the secured PHP is the short open tag `<? ... ?>`, which has not effect on the short echo tag

Mentioned in the php.ini 
```ini
; This directive determines whether or not PHP will recognize code between
; <? and ?> tags as PHP source which should be processed as such. It is
; generally recommended that <?php and ?> should be used and that this feature
; should be disabled, as enabling it may result in issues when generating XML
; documents, however this remains supported for backward compatibility reasons.
; Note that this directive does not control the <?= shorthand tag, which can be
; used regardless of this directive.
; Default Value: On
; Development Value: Off
; Production Value: Off
; https://php.net/short-open-tag
short_open_tag = Off
```
> ; Note that this directive does not control the <?= shorthand tag, which can be used regardless of this directive.

I can still replace it, if you prefer, but I think the short echo tag is easier to read und understand, while printing lot of HTML code

hregis:
@atm-etienne @eldy my PR change codesniffer rule to find short open tag "<?" and exclude short echo tag "<?=" : https://github.com/Dolibarr/dolibarr/pull/32634

htdocs/adherents/card.php
atm-etienne: @eldy have you  had time to review the last comment?

eldy: Don' find easier.
The fact is that we got problem in the past with it so we must avoid them.

PR #32629, opened 2025-01-11
QUAL upgrade jQuery Jeditable to 2.0.21

hregis:


hregis:
@eldy @mdeweerd 
i don't understand this error : 

`Short PHP opening tag used with echo; expected "<?php echo DOL_URL_ROOT ..." but found "<?= DOL_URL_ROOT`

https://www.php.net/manual/en/language.basic-syntax.phptags.php

hregis:
short echo tag "<?=" is not a short open tag "<?"

hregis:
@eldy @mdeweerd the new codesniffer rule found short open tag "<?" and exclude short echo tag "<?="

In php to allow <?= It seems it is the same option than to allow short tag <?
And php recommand to didable <? So mechanically we must avoid <?=
You can replace with <?php echo

mdeweerd:
As you indicated this is probably configurable as a PHP Codesniffer option.

However the question is what the rule is for Dolibarr.  There was a document explaining that PSR-2 or PSR-12 or so was applicable with a list of the exceptions (not the phpcs configuration itself).
So it depends on the policy for Dolibarr.

hregis:
jQuery Jeditable n'est plus maintenu et va être remplacé par Malle qui ne dépendra plus de jQuery

PR #32634, opened 2025-01-11
FIX allow short echo tag "<?=" and not short open tag "<?"

hregis:


eldy:
Why allowing <?= 

Php recommand to not allow it and reolace it with <?php echo

hregis:
@eldy 
"short_open_tag = off" is for "<?" and your rights its not recommanded.
but the short echo tag "<?=" is an alias of "<?php echo"

```
"You can use the short echo tag to <?= 'print this string' ?>.
It's equivalent to <?php echo 'print this string' ?>."
```
https://www.php.net/manual/en/language.basic-syntax.phptags.php

eldy:
> @eldy 
> "short_open_tag = off" is for "<?" and your rights its not recommanded.
> but the short echo tag "<?=" is an alias of "<?php echo"
> 
> ```
> "You can use the short echo tag to <?= 'print this string' ?>.
> It's equivalent to <?php echo 'print this string' ?>."
> ```
> https://www.php.net/manual/en/language.basic-syntax.phptags.php
> 

Yes but it seems that enabling <?= is same setup than enabling <?.
So <?= does not work on all php configurations.
This is for this reason i suppose that using <?= is disallowed by ci rules and not recommanded to be used by php.

hregis:
@eldy no read the php docs, "short_open_tag = off" disallow only "<?"

Note:

`As short tags can be disabled it is recommended to only use the normal tags (<?php ?> and <?= ?>) to maximise compatibility`

eldy:
> @eldy no read the php docs, "short_open_tag = off" disallow only "<?"
> 
> Note:
> 
> `As short tags can be disabled it is recommended to only use the normal tags (<?php ?> and <?= ?>) to maximise compatibility`
> 
> 

Yes but i found other mention that say opposite. Also this generates pb with ci tools. It seems that it depends on php version too. Forums are full of you can, you shouldn't.
May questuon is why using 2 different syntax when one is ok. Is saving 5 chars sometimes enough to justify this ?
Note also that using <?= is also used to bypass security protection of php injection, external waf but also dolibarr waf so we had to add protection into function like dolKeepOnlyPhpCode or dol_eval.
As we have no use of <?= currently in dolibarr, in the doubt, there is no reason to introduce a controversal practice.

PR #32640, opened 2025-01-13
QUAL remove unused jQuery-Jeditable

hregis:


hregis:
@mdeweerd @frederic34 can you see this phan and php-stan errors ? thanks

mdeweerd:
These are fixed in #32638

eldy:
And after the removal if option MAIN_USE_EDIT_IN_PLACE is on, can you paste a screenshot to show what is result ?

hregis:
@eldy for the moment I just renamed the constant, we need to make a point with @lmag and @thersane-john in order to define the best way to integrate an "edit in place" in relation to this feature request : https://github.com/Dolibarr/dolibarr/issues/32620

thersane-john:
There is currently some important groundwork being done around the Dolibarr JS context to properly prepare the arrival of inline editing, meaning the ability to edit object fields directly through Ajax without reloading the page.

However, before achieving something truly maintainable and generic, the JavaScript layer behind fields needs to become much more dynamic. Today, a significant part of the JS is still generated directly by PHP and often injected inline. The issue is that when content is reloaded through Ajax, this JavaScript does not properly reinitialize and often requires manual re-execution.

One of the first steps is therefore to progressively move away from the current PHP/JS mixing approach and stop generating JavaScript directly from field rendering. Instead, the idea would be to rely on dedicated JS libraries and let the DOM drive component behavior through attributes, JS hooks, or clearly defined focus areas.

The goal is that when an HTML block is dynamically injected through Ajax, the related components can automatically reinitialize themselves without requiring custom re-execution logic everywhere.

I am currently thinking about implementing a JS library autoloader integrated into the Dolibarr JS context. The idea would be to load libraries only when a component actually requires them. If a library, class, or module is not yet loaded, the system could automatically fetch it through a central registry or directory mechanism. This would avoid loading the entire JS stack upfront while also improving frontend modularity and resilience.

I would like to discuss this topic during the DevCamp to explore the different possible approaches and determine what kind of long-term architecture would make the most sense for Dolibarr.

PR #32644, opened 2025-01-13
Show only open projects on short list

rycks:
With a standard dolibarr setup here is what you can see:

image

most of them are not valided ... that PR add an option to make that result:

image

### DIFF
--- a/htdocs/core/lib/security.lib.php
+++ b/htdocs/core/lib/security.lib.php
@@ -245,7 +245,7 @@ function dolDecrypt($chain, $key = '')
  * 										Use 'md5' if hash is not needed for security purpose. For security need, prefer 'auto'.
  * 	@param 		int 		$nosalt		Do not include any salt
  *  @param		int			$mode		0=Return encoded password, 1=Return array with encoding password + encoding algorithm
- * 	@return		string|array<pass_encrypted:string,pass_encoding:string>			Hash of string or array with pass_encrypted and pass_encoding
+ * 	@return		string|array(pass_encrypted:string,pass_encoding:string)			Hash of string or array with pass_encrypted and pass_encoding
  *  @see getRandomPassword(), dol_verifyHash()
  */
 function dol_hash($chain, $type = '0', $nosalt = 0, $mode = 0)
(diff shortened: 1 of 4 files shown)


rycks:
note: i forget all the time where / what i have to do for new translations strings ... so please refresh my memory about that point, just to add PROJECT_SHOW_ONLY_VALIDATED_ON_LATEST_UPDATE_help and PROJECT_SHOW_ONLY_VALIDATED_ON_LATEST_UPDATE strings !

eldy:
> note: i forget all the time where / what i have to do for new translations strings ... so please refresh my memory about that point, just to add PROJECT_SHOW_ONLY_VALIDATED_ON_LATEST_UPDATE_help and PROJECT_SHOW_ONLY_VALIDATED_ON_LATEST_UPDATE strings !

Just add the 2 entries into a htdocs/langs/en_US/xxx.lang file.

rycks:
ok, that's done on lang us but i don't know what i have to do with phan / php-stan errors ...

htdocs/core/lib/security.lib.php
```diff
@@ -245,7 +245,7 @@ function dolDecrypt($chain, $key = '')
  * 										Use 'md5' if hash is not needed for security purpose. For security need, prefer 'auto'.
  * 	@param 		int 		$nosalt		Do not include any salt
  *  @param		int			$mode		0=Return encoded password, 1=Return array with encoding password + encoding algorithm
- * 	@return		string|array<pass_encrypted:string,pass_encoding:string>			Hash of string or array with pass_encrypted and pass_encoding
+ * 	@return		string|array(pass_encrypted:string,pass_encoding:string)			Hash of string or array with pass_encrypted and pass_encoding
```
eldy, line 248: Why changing this ?
Before the change, the syntax was correct.

rycks: sorry that's a try to be clean against phpcs/phan/other rules who seems to refuse that PR due to line 248 syntax ...

rycks: image

hregis: @rycks merge the branch develop, i think is fixed

rycks:
@eldy up ?

PR #32645, opened 2025-01-13
fix no contact in select when COMPANY_USE_SEARCH_TO_SELECT is not empty

uvaldenaire-opendsi:
When COMPANY_USE_SEARCH_TO_SELECT is not empty, it was not possible to select a contact after a company was selected.
 
Also, when editing a event, if object->socid is empty, the selectcontacts method was loading all the contacts in database


eldy:
Is it a backport from a higher version. If yes, which one ?

uvaldenaire-opendsi:
It was not a backport, but i found one that fix the problem on Dolibarr 18. I close this PR and open another with the backport.

PR #32649, opened 2025-01-14
FIX split data and rename files (Multicompany compatibility)

hregis:
@eldy @aspangaro no errors during a fresh install


aspangaro:
Thank you @hregis 

This PR follows an exchange between the two of us where it is necessary to separate the Journals / Chart of Accounts data files for the accounting in order to facilitate entity deployments.

hregis:
je met en brouillon, je vais tester avec Postgresql

aspangaro (approved):

hregis:
@eldy @aspangaro no problem with postgresql

PR #32651, opened 2025-01-14
Apply suggestion by @mdeweerd to ensure pre-commit will pass.

atm-florianm:
# FIX pre-commit failing on a debug bar class

As discussed in #32621, there are some code standards issues with the DebugBar in v17, which cause the pre-commit hook to fail even though these files are entirely unaffected by the PR.

I am opening this PR instead to separate this issue from the original one.

### DIFF
--- a/htdocs/includes/maximebf/debugbar/src/DebugBar/DataCollector/PDO/TraceablePDO.php
+++ b/htdocs/includes/maximebf/debugbar/src/DebugBar/DataCollector/PDO/TraceablePDO.php
@@ -23,13 +23,13 @@ class TraceablePDO extends PDO
         $this->pdo->setAttribute(PDO::ATTR_STATEMENT_CLASS, [TraceablePDOStatement::class, [$this]]);
     }
 
-	/**
-	 * Initiates a transaction
-         *
-	 * @link   http://php.net/manual/en/pdo.begintransaction.php
-	 * @return bool TRUE on success or FALSE on failure.
-	 */
-    public function beginTransaction()
+   /**
+    * Initiates a transaction
+    *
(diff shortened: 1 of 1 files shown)


atm-florianm:
> `PHP Fatal error:  $GLOBALS can only be modified using the $GLOBALS[$name] = $value syntax in htdocs/includes/symfony/var-dumper/Tests/CliDumperTest.php on line 512`

@eldy Should I attempt to fix that as well, or is it better to leave it alone for the time being?

mdeweerd:
```
PHP Fatal error:  Array and string offset access syntax with curly braces is no longer supported in htdocs/includes/tecnickcom/tcpdf/include/barcodes/pdf417.php on line 881
Errors parsing htdocs/includes/tecnickcom/tcpdf/include/barcodes/pdf417.php
```

We either ignore this like so:
https://github.com/Dolibarr/dolibarr/blob/54477083d4f03a58bb495a16dc0c72a35d8fb5e7/.pre-commit-config.yaml#L89-L91

Or we ensure that an older php version is installed in the pre-commit setup (corresponding to the maximum version supported by the corresponding dollibarr version).
Get the version by getting it from htdocs/install/check.php - currently in develop: `"$arrayphpmaxversionwarning = array(8, 3, 0);"`

mdeweerd:
@atm-florianm I made a PR to your branch which installs the maximum supported PHP version before running pre-commit.

This way ci work and will only show issues in the appropriate dolibarr branches where external libraries may already have been updated.

EDIT:
Or, I can make the PR directly to dolibarr as my change fixes the issue by itself in 17.0.
EDIT2:
I went ahead and created #32653 .

PR #32652, opened 2025-01-14
fix: add entity in where for fetch when fetch by ref

FHenry:
With multicompany, when try to diplay PDF of exepnese report
document.php=>dol_check_secure_access_document=>$tmpexpensereport->fetch('', $refname);
if there is other exepnese report with the same ref in other entity the test done is not correct (bad fk_user_create)

### DIFF
--- a/htdocs/expensereport/class/expensereport.class.php
+++ b/htdocs/expensereport/class/expensereport.class.php
@@ -619,6 +619,7 @@ class ExpenseReport extends CommonObject
 		$sql .= " FROM ".MAIN_DB_PREFIX.$this->table_element." as d";
 		if ($ref) {
 			$sql .= " WHERE d.ref = '".$this->db->escape($ref)."'";
+			$sql .= " WHERE d.entity IN (".getEntity('expensereport').")";
 		} else {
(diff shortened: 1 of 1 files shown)


htdocs/expensereport/class/expensereport.class.php
```diff
@@ -619,6 +619,7 @@ public function fetch($id, $ref = '')
 		$sql .= " FROM ".MAIN_DB_PREFIX.$this->table_element." as d";
 		if ($ref) {
 			$sql .= " WHERE d.ref = '".$this->db->escape($ref)."'";
+			$sql .= " WHERE d.entity IN (".getEntity('expensereport').")";
```
eldy, line 622: I think you want to do 
AND
and not
WHERE

eldy (changes requested):

htdocs/expensereport/class/expensereport.class.php
FHenry: Damn.... Of course

rycks (approved):

lvessiller-opendsi (approved):

PR #32661, opened 2025-01-15
Backport FIX manage event in select_company with AJAX search option on

uvaldenaire-opendsi:
Backport of https://github.com/Dolibarr/dolibarr/pull/18253

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -1281,6 +1281,9 @@ class Form
 			if ($hidelabel == 3) {
 				$out .= img_picto($langs->trans("Search"), 'search');
 			}
+
+			$out .= ajax_event($htmlname, $events);
+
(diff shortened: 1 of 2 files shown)


eldy:
Can you remind me how to reproduce the bug ?

uvaldenaire-opendsi:
In third parties config page put COMPANY_USE_SEARCH_TO_SELECT to at least 1.
This is the field: "Wait until a key is pressed before loading content of Third Parties combo list".

Then go to create new event. Select a third party (Related company) . 

The third party contacts should now be in "Related contact" select field. But they are not.

Also, if you set CONTACT_USE_SEARCH_TO_SELECT to at least one in the third parties config page,
(The field "Wait until a key is pressed before loading content of Contact combo list.")
and type the begining of a contact name, the contact don't appear in the options.

PR #32668, opened 2025-01-15
Qual: Fix ci (pre-commit) Exclude/fix shellcheck notices

mdeweerd:
# Qual: Fix ci (pre-commit) Exclude/fix shellcheck notices

This fixes ci complaining about script inside pre-commit.yml.  Will likely also need to backport this to older branch.


mdeweerd:
@eldy This is something that you want to merge asap.

PR #32680, opened 2025-01-16
FIX Udefined property $read, $upload, $setup

PsyCrow-code:
FIX Udefined property $read, $upload, $setup
FIX PHP Warning: undefined property on eval()


PsyCrow-code (commented):
changing old style for acces to user rights

PR #32684, opened 2025-01-16
ENH: accessibility on ticket admin page

BenjaminFlr:
Little PR to check if it could be integrated in far more pages.
I just added a label for input fields in order to focus on the input if we click on the parameter `<td>`.

For example, if we click on the `<td>` containing `E-mail de l'expéditeur pour la notification des réponses`, it will automatically focus on the input.

Capture d’écran 2025-01-16 à 12 11 39

What do you think of this just starting work to improve Dolibarr accessibility ?
If ok for you, I'll make more PRs (and bigger PRs) in a close future.


eldy:
Why not...

PR #32688, opened 2025-01-16
Fix ci: travis - Better test that phpunit really completed

mdeweerd:
# Fix ci: travis - Better test that phpunit really completed

Check that phpunit completed by looking for the test summary.


mdeweerd:
@eldy This makes travis fail if phpunit does not have a test conclusion.

In this PR it fails because the exit is still present in the SecurityTest.php.  So this is proof that the extra code works when it fails.
I'll let you update the SecurityTest.php here so that you can observe for yourself that travis no longer fails when that is fixed.

Later this can be backported to an older branch if you want.

mdeweerd:
@eldy Further improved + showing message with result code instead of abruptly exiting.

mdeweerd:
Already merged in 21.0 because merge in 19.0 or 20.0.

PR #32691, opened 2025-01-17
TakePos Once payment is complet, a click on a product should start a new sale

dbeniamine:
# New Start a new sale on click product on TakePOS

We upgraded a customer from Dolibarr 17.x to 20.x and they add an issue : on Dolibarr 17.x on TakePOS when we click on a product after a sale is complete it tries to update the sell instead of creating a new one.

This patchs allows to start a new sell by clicking a product

### DIFF
--- a/htdocs/takepos/index.php
+++ b/htdocs/takepos/index.php
@@ -527,6 +527,10 @@ function MoreProducts(moreorless) {
 
 function ClickProduct(position, qty = 1) {
 	console.log("ClickProduct at position"+position);
+	if ($('#invoiceid').val() == "") {
+		invoiceid = $('#invoiceid').val();
+		Refresh();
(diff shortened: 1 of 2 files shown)


eldy:
What if payment is partial ? 
For example invoice is 100, payment recorded is 80, then we add a product to complete the invoice.
Is a new invoice created or product added on existing invoice ?

dbeniamine:
> What if payment is partial ? For example invoice is 100, payment recorded is 80, then we add a product to complete the invoice. Is a new invoice created or product added on existing invoice ?

As expected, in this use cas you get a warning that the operation is not possible (adding a product to a partially payed bill) : 

image
image

Et then I still can finish the payment and the next click on product will trigger a new sale

PR #32714, opened 2025-01-19
NEW add credit note ids in API invoice return

hregis:


htdocs/fourn/class/fournisseur.facture.class.php
```diff
@@ -1013,6 +1013,10 @@ public function fetch($id = 0, $ref = '', $ref_ext = '')
 
 				$this->socid  = $obj->socid;
 
+				if ($this->type != self::TYPE_CREDIT_NOTE) {
+					$this->creditnotelist = $this->getListIdAvoirFromInvoice();
```
eldy, line 1017: A fetch is a CRUD method to get a record of the table, Except for extrafields and lines, we should not have sub request into a fetch.
Doing subrequest must be done only when necessary.

PR #32722, opened 2025-01-20
Qual: Fix ci for new api_memberstypes.class + 2 other fixes

mdeweerd:
# Qual: Fix ci for new api_memberstypes.class + 2 other fixes

This fixes the notifications for the new api_memberstypes.class


mdeweerd:
@hregis Should we change this to `membertypes.class.php` instead or is it too late to do that?

hregis:
@mdeweerd you can rename the file and the class

mdeweerd:
> @mdeweerd you can rename the file and the class

@eldy I see that this is a rathrer old function - should I rename it, in which version ?
(That should not be a limitation for merging this ci-fix, I can do another PR for the rename in the target version).

hregis:
@mdeweerd oh sorry, you were talking about adherent_type.class.php ?

mdeweerd:
@hregis NO, MembersTypes class in memberstypes.class.php.  In english that would be have to be MemberTypes.
I thought this was new as the phan notices appeared just now, but from the git log that seems to exist since quite a few versions already.

PR #32724, opened 2025-01-20
fix: allow list mass action update price when price mode is multiprice price

FHenry:
On the contrary of https://github.com/Dolibarr/dolibarr/pull/32564 a better solution is to update all multiprice (and display inforamtion message).

As this feature was present, the old fix to avoid it was not ideal. Thie PR allow the feature again in this case and do the correct update.

### DIFF
--- a/htdocs/product/list.php
+++ b/htdocs/product/list.php
@@ -829,8 +829,11 @@ $arrayofmassactions = array(
 	//'presend'=>img_picto('', 'email', 'class="pictofixedwidth"').$langs->trans("SendByMail"),
 );
 if ($user->hasRight($rightskey, 'creer')) {
-	$arrayofmassactions['preupdateprice'] = img_picto('', 'edit', 'class="pictofixedwidth"').$langs->trans("UpdatePrice");
-
+	if (getDolGlobalString('PRODUCT_PRICE_UNIQ')
+		|| getDolGlobalString('PRODUIT_CUSTOMER_PRICES')
+		|| getDolGlobalString('PRODUIT_MULTIPRICES')) {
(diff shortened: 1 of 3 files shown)


htdocs/product/list.php
```diff
@@ -829,9 +829,8 @@
 	//'presend'=>img_picto('', 'email', 'class="pictofixedwidth"').$langs->trans("SendByMail"),
 );
 if ($user->hasRight($rightskey, 'creer')) {
-	if (getDolGlobalString('PRODUCT_PRICE_UNIQ') || getDolGlobalString('PRODUIT_CUSTOMER_PRICES')) {
```
eldy, line 832: It seems there is other cases.
When in dev mode, we see 6 different values in combo for parameter to set the for rules to define price.
The update price was implemented for 2, to if you remove this test, it means you implement it for the 6 cases now, so 4 more cases.
However in htdocs/core/actions_massactions.inc.php, i see you add only one if for case  PRODUIT_MULTIPRICES
The initial fix was a protection to avoid bad effect on not yet implement cases, removing it to support more case would be better in develop.

FHenry: Ok I understand, So I suggest in this new commit only allow 
`getDolGlobalString('PRODUCT_PRICE_UNIQ') || getDolGlobalString('PRODUIT_CUSTOMER_PRICES') || getDolGlobalString('PRODUIT_MULTIPRICES')`
and manage in actions_massactions.inc.php only this cases.

I agree that I preffers to have "all cases" senario, and I will provide a PR in develop branch, in case of PRODUIT_MULTIPRICES, to choose witch price level have to be concern about price changes.

In this version of the PR, only managed cases are correctly calculated and annonced to users

PR #32731, opened 2025-01-21
NEW default values for filters that are multiselect array are now supported

frederic34:
image

image


htdocs/core/lib/functions.lib.php
```diff
@@ -1065,7 +1065,7 @@ function GETPOST($paramname, $check = 'alphanohtml', $method = 0, $filter = null
 	// Check type of variable and make sanitization according to this
 	if (preg_match('/^array/', $check)) {	// If 'array' or 'array:restricthtml' or 'array:aZ09' or 'array:intcomma'
 		if (!is_array($out) || empty($out)) {
-			$out = array();
+			$out = explode(',', $out);
```
eldy, line 1068: Other cases include a sanitizeVal on the value coming from the database.
Can you include someting like 

foreach ($out as $outkey => $outval) {
				$out[$outkey] = sanitizeVal($outval, 'alphanohtml', $filter, $options);
			}

PR #32741, opened 2025-01-22
Make table responsive in eldy

ALYMBEK-KG:
Added following styles:
----------------------------------
display: grid;
grid-template-columns: minmax(0, 1fr);


eldy:
Can you provide a screenshot that show the difference ?

ALYMBEK-KG:
image
image

eldy:
Currently we want to avoid this.
Imagine you have 20 lines instead of 3 in the list: 
When you want to scroll in right or left it becomes no more possible. You need to first scroll down ti find how to scroll left, right 
For this reason we must first wait to have the feature of v21 that allows to have the max nb if lines that is set automatically according to height of page si the bottom of table will be always visible before adding this...

PR #32743, opened 2025-01-22
 FIX the sum of warehouse totals

djmai:
Remove this line, since if the variable $totalarray is left active it is reset and does not correctly sum the totals in the warehouse list table

Preview before fixing the error
image

Preview bug already fixed
image


hregis:
@djmai i move your fix in 20 branch, thanks : https://github.com/Dolibarr/dolibarr/pull/32744

PR #32761, opened 2025-01-23
NEW: Link contract and fichinter

noec764:
When the contract module is activated on an intervention card, it is possible to link the intervention with a contract. But from the contract form, there's currently no way of finding out which intervention the contract is linked to.

So I've added a link in the llx_element_element table when the contract is selected on the job. In this way, the intervention can be found from the contract card, thanks to the linked_object block.

I've also added a button on the contract card that lets you create an intervention from within the contract.


eldy:
The problem is that we introduce duplication of information. A link between a contract and an intervention is now materialized by 2 different information.
What if you have the link with field fk_contrat filled but not the link with table llx_elem ? And what if the opposite ?

Because the linke is saved into field fk_contrat, the screen pages must also use this field to show the information.
Trying to reuse an existing feature may be a good idea but only if it does not generate unstability into the database.

So, because other objects are using the link with the table llx_element_element, it is a good idea to switch to this mode too. However, it means we must abandon the other way (in both the use of the database but also in the code). This will need a transition time so i suggest to introduce a hidden constant
CONTRACT_INTERVENTION_LINK_USING_TABLE_ELEMENT
When set, you can introduce code that is using the llx_element_element table. If unset, the link is the fk_contrat of the table llx_fichinter like currently.

noec764:
@eldy Thanks for your answer,
However, I have a question. You say that fk_contract should be phased out, does this mean that when the new hidden const is defined, fichinter's fk_contract should be left empty? 

Not necessarly. Because if the new hidden const is defined, the field fk_contract will not be "read" anymore. So we don't mind if value is empty or not. Wecan continue to save it so we can switch to the other mode with no risk to loose data.

PR #32770, opened 2025-01-24
FIX broken feature, show open tickets by default

hregis:


hregis:
@eldy better fix here : https://github.com/Dolibarr/dolibarr/pull/32776

PR #32774, opened 2025-01-24
Bug in tva.lib.php when using FACTURE_DEPOSITS_ARE_JUST_PAYMENTS hidden feature and vat setup on payment date

jyhere:
**How to reproduce the bug :**
- on the VAT admin setup page, use the following setup (or test with the standard setup and a service) :  
image

- set the misc param FACTURE_DEPOSITS_ARE_JUST_PAYMENTS to 1
- Create an order with a 20% vat product
image

- Create a deposit invoice from the order
- Add a payment on the deposit invoice
- Create the order invoice, and pay this invoice using the deposit invoice. A discount is created.
- Apply the discount on the invoice
image
image

- Go to /compta/tva/quadri_detail.php page, filter on the current date : 
image

Our deposit vat should be visible.

### DIFF
--- a/htdocs/core/lib/tax.lib.php
+++ b/htdocs/core/lib/tax.lib.php
@@ -438,19 +438,7 @@ function tax_by_thirdparty($type, $db, $y, $date_start, $date_end, $modetax, $di
 		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."product as p on d.fk_product = p.rowid";
 		$sql .= " WHERE f.entity IN (".getEntity($invoicetable).")";
 		$sql .= " AND f.fk_statut in (1,2)"; // Paid (partially or completely)
-		if ($direction == 'buy') {
-			if (getDolGlobalString('FACTURE_SUPPLIER_DEPOSITS_ARE_JUST_PAYMENTS')) {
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/tax.lib.php
```diff
@@ -438,19 +438,7 @@ function tax_by_thirdparty($type, $db, $y, $date_start, $date_end, $modetax, $di
...
-				$sql .= " AND f.type IN (0,1,2,3,5)";
-			}
-		} else {
-			if (getDolGlobalString('FACTURE_DEPOSITS_ARE_JUST_PAYMENTS')) {
-				$sql .= " AND f.type IN (0,1,2,5)";
-			} else {
-				$sql .= " AND f.type IN (0,1,2,3,5)";
-			}
-		}
+		$sql .= " AND f.type IN (0,1,2,3,5)";
```
eldy, line 441: The down/deposit invoices were excluded when using the option _ARE_JUST_PAYMENTS because when we calculate vat on invoice generation, we need to exclude them (since they are no more invoice when hidden option is on).
So you can enable the filter for type = 3 like you did, but only if vat calculation is done on payment and not on invoice.

PR #32779, opened 2025-01-24
Fix ci phan/phpstan notices

mdeweerd:
# Fix ci phan/phpstan notices

Fix most notices, except undefined fk_website in SocieteAccount as it is not clear if that is supposed to be a property there (more specifically, that it is set).


frederic34:
https://github.com/Dolibarr/dolibarr/pull/32758 done some fix in branch 21

mdeweerd:
> #32758 done some fix in branch 21

Thanks, I update the resolution in note.php to match yours, cherry picked the resolution for societeaccount, but kept a resolution in conf.class.php as I restrict the values to `int<0,1>` (which will be in conflict).

mdeweerd:
Fixed some new things, remaining issues need functional analysis.

PR #32785, opened 2025-01-25
Add WEBSITE_PHP_ALLOW_DYNAMIC_FUNCTIONS passby

defrance:
we need it for marketplace install


eldy:
> we need it for marketplace install
> 

We should not need it, in any case. Which error do you have and how do you reproduce ?

defrance:
They append here
image

Need to add this too : 
WEBSITE_PHP_ALLOW_EXEC 
WEBSITE_PHP_ALLOW_WRITE

eldy:
This looks strange. I do not have this.
Can you try this. With last recent version, enable the module and check that WEBSITE_PHP_ALLOW_WRITE is set.

The WEBSITE_PHP_ALLOW_EXEC is not required. We don't have it in production.

eldy:

PR #32789, opened 2025-01-25
FIX: Disable Multicompany module in demo set

BenjaminFlr:
If enabled by default but not installed, it will cause many errors in various pages.


hregis:
@BenjaminFlr you can remove all for avoid 404 error with css or others

BenjaminFlr:
@hregis also triggers and hook? I thought it wouldn't break anything but you're right.

hregis:
@BenjaminFlr if no module... no constants ! 😄

BenjaminFlr:
Done.

PR #32805, opened 2025-01-27
FIX avoid php8 warning

hregis:


hregis:
@eldy GETPOST() can return `array(0 => '')` and this is not empty and count > 0 but the value is empty and create warnings in many cases

htdocs/core/lib/functions.lib.php
```diff
@@ -1075,8 +1075,14 @@ function GETPOST($paramname, $check = 'alphanohtml', $method = 0, $filter = null
 				$tmpcheck = $tmparray[1];
 			}
 		}
-		foreach ($out as $outkey => $outval) {
-			$out[$outkey] = sanitizeVal($outval, $tmpcheck, $filter, $options);
+		// avoid warning when the return is array(0 => '')
```
eldy, line 1078: I can't see how you can get array(0 => '') here ?
Which value of $out in input do you have for this ?

hregis:
@eldy by example here, with "show_contacts() function  : 

Capture d’écran 2025-01-27 à 14 42 24

the sql error : `WHERE sc.fk_c_type_contact IN ()`

Capture d’écran 2025-01-27 à 14 42 52

in the code : 

Capture d’écran 2025-01-27 à 14 45 26


`$search_roles = GETPOST("search_roles", 'array');`


return an array : array(0 => '')

et ça arrive souvent !! 😄

hregis:
@eldy ça doit surement venir de cette partie de code dans GETPOST mais comme je ne sais pas le pourquoi du comment il y a ça, je préfère traiter le problème après coup : 

```
// Check type of variable and make sanitization according to this
	if (preg_match('/^array/', $check)) {	// If 'array' or 'array:restricthtml' or 'array:aZ09' or 'array:intcomma'
		if (!is_array($out) || empty($out)) {
			$out = explode(',', $out);
			$tmpcheck = 'alphanohtml';
		} else {
			$tmparray = explode(':', $check);
			if (!empty($tmparray[1])) {
				$tmpcheck = $tmparray[1];
			} else {
				$tmpcheck = 'alphanohtml';
			}
		}
		foreach ($out as $outkey => $outval) {
			$out[$outkey] = sanitizeVal($outval, $tmpcheck, $filter, $options);
		}
	}
```

eldy:
> search_roles

I think so too.This is why i ask the question, because i like to fight the trouble at the first apparition. later is good too, it is compensation, but when doing compensation it hides the first apparition of trouble. And i prefer hunting at the source...

eldy:
Hum, it seems i have already fixed this, accidentily in this commit  a99b132c346e91892973bcd95d389fe39210b05f

hregis:
@eldy yes, you are right ! the miracle of computing !! 😄 Thank you God !! ❤️

PR #32810, opened 2025-01-27
FIX : Mass file area rights with "objects"

Kazimir42:
Hello,

In my module, the right to view an object is defined as follows:

`$user->rights->mymodule->myobject->read`

However, the current code does not handle "myobject" properly; it only checks the `$lire` variable instead of `$read`.

So, I simply copied and pasted the code below from the else block.

### DIFF
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -3564,9 +3564,19 @@ function dol_check_secure_access_document($modulepart, $original_file, $entity, 
 				dol_print_error(null, 'Error call dol_check_secure_access_document with not supported value for modulepart parameter ('.$modulepart.')');
 				exit;
 			}
-			if ($fuser->hasRight($tmpmodule, $lire) || preg_match('/^specimen/i', $original_file)) {
-				$accessallowed = 1;
-			}
+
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/files.lib.php
```diff
@@ -3564,9 +3564,19 @@ function dol_check_secure_access_document($modulepart, $original_file, $entity,
 				dol_print_error(null, 'Error call dol_check_secure_access_document with not supported value for modulepart parameter ('.$modulepart.')');
 				exit;
 			}
-			if ($fuser->hasRight($tmpmodule, $lire) || preg_match('/^specimen/i', $original_file)) {
+
```
eldy, line 3567: the method >hasRight($tmpmodule, $lire) will test both on "read" and "lire" if value of $lire is "lire" or "read" (we don't mind, both are tested inside method).
Your problem is surely not on the read/lire, but only the fact you made a permission on 3 levels:
module->object->lire/read
where default code searches permission on 2 level only
module->lire/read
for the massaction massfilesarea_
So can you remove the duplicate test on "lire"/ "read" (one is enough), and keep only the part that do the test

Kazimir42: Done :)

PR #32811, opened 2025-01-27
FIX can not delete files in task card

hregis:


rycks:
mmmm i think that PR does not work, sorry @hregis but let me try with a file linked to a task wich name is not "tast ref name" based for example

PR #32830, opened 2025-01-28
FIX compatibility with multicompany

hregis:


hregis:
@eldy @aspangaro records appear duplicate with multicompany (v17 to develop)

aspangaro (commented):
Ok for me. This only applies to accounting categories, not to the accounting itself

PR #32839, opened 2025-01-28
Qual: Fix ci - &quot; replacement and expected test result

mdeweerd:
# Qual: Fix ci - &quot; replacement and expected test result

@eldy This fixes one of the issues in phpunit ci in 21.0

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -2220,8 +2220,8 @@ function dol_escape_htmltag($stringtoescape, $keepb = 0, $keepn = 0, $noescapeta
 				} while ($diff);
 			}
 
-			$tmp = str_ireplace('&quot', '__DOUBLEQUOTE', $tmp);
-			$tmp = str_ireplace('&lt', '__LESSTAN', $tmp);
+			$tmp = str_ireplace('&quot;', '__DOUBLEQUOTE', $tmp);
+			$tmp = str_ireplace('&lt', '__LESSTHAN', $tmp);
 			$tmp = str_ireplace('&gt', '__GREATERTHAN', $tmp);
 		}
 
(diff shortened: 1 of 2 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -2220,8 +2220,8 @@ function dol_escape_htmltag($stringtoescape, $keepb = 0, $keepn = 0, $noescapeta
 				} while ($diff);
 			}
 
-			$tmp = str_ireplace('&quot', '__DOUBLEQUOTE', $tmp);
-			$tmp = str_ireplace('&lt', '__LESSTAN', $tmp);
+			$tmp = str_ireplace('&quot;', '__DOUBLEQUOTE', $tmp);
```
eldy, line 2223: We don't want the semicolon because in hacking world the &xxx is often used without the semicolin. It works same way often and it bypass sanitization.

test/phpunit/SecurityTest.php
```diff
@@ -794,7 +795,7 @@ public function testDolPrintHTMLAndDolPrintHtmlForAttribute()
 
 		// For a string that is already HTML (contains HTML tags) with special tags but badly formatted
 		$stringtotest = "&quot; &gt; &lt; <b>bold</b>";
-		$stringfixed = "&quot; &gt; &lt; <b>bold</b>";
+		$stringfixed = "\" &gt; &lt; <b>bold</b>";
```
eldy, line 798: We want here &quot.
Why replacing with " ?

htdocs/core/lib/functions.lib.php
mdeweerd: Ok, the issue is then that:
`"` is replaced with __DOUBLEQUOTE
and
`&quot` is also replaced with __DOUBLEQUOTE which then becomes `";` .

So this means that we either need two different names for DOUBLEQUOTE or we need to convert `"` to "__DOUBLEQUOTE;" so that it becomes `&quot;` in the end.

I would think we cant the first approach, unless the `"` tokens need to be escaped which you did not implement (so therefore I think it's the first option (same as keeping " or &quot; in the end according to what was in the original.

eldy: i see. I pushed a fix for this point into 2483c4ec58bc452d747796d41269b9a379d4fb0f

PR #32844, opened 2025-01-29
FIX wrong table name

hregis:


aspangaro (commented):
I confirm the problem.

Without fix
image

PR #32847, opened 2025-01-29
fix phan

frederic34:


mdeweerd:
As I supposed these notices appeared because of fixes I made, I solved the in #32841 or #32842 (I believe the latter).

PR #32866, opened 2025-01-29
FIX add DDLDropIndex function for Mysql compatibility

hregis:


hregis:
@FHenry it's ok for you ?

eldy:
For my understanding, what was wrong with the code done just after this PR #32848.
The old code was doing a simple 
DROP INDEX without the IF EXISTS that was removed.
so it should works in all databases with PR #32848. Did i miss something ?

hregis:
@eldy yes you are right, on the other hand for my part, I will put back the display of errors because if we modify an extrafield by putting that it must be unique and that this extrafield already has duplicated data, we do not see that the creation of the unique key is in error.

PR #32868, opened 2025-01-30
Auto-Generation of Lot/Serial Numbers in Manufacturing Orders and Reception Modules

pratushraj:
*Currently, users must manually input Lot/Serial numbers for products during Manufacturing Orders or Reception processes. This can be time-consuming, especially in large-scale operations or scenarios where unique Lot/Serial tracking is critical for compliance or inventory management.*

When the Sr.no/lot no. Auto generation enabled, if the user leaves the Lot/Serial number field blank, Dolibarr should automatically generate a unique identifier based on a pre-defined format or sequence .

# Expected Benefits:
Improved efficiency by reducing manual data entry during Manufacturing and Reception processes.

### DIFF
--- a/htdocs/core/lib/functions2.lib.php
+++ b/htdocs/core/lib/functions2.lib.php
@@ -1312,7 +1312,14 @@ function get_next_value($db, $mask, $table, $field, $where = '', $objsoc = '', $
 
 		$numFinal = $ref;
 	} elseif ($mode == 'next') {
-		$counter++;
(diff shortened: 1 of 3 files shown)


eldy:
Hi @pratushraj .
Not sure, but when you pushed your PR, it seems you didn't check the box "Can by modified", so we are not able to update the PR to use the last version of code that contains the bug fixes for the CI.
Can you try to repush it by checking the checkbox ?

pratushraj:
I have repush this.

eldy:

PR #32869, opened 2025-01-30
FIX add constant to avoid price error when TTC is price base type

hregis:


hregis:
@eldy maybe remove the possibility to set content of TTC field with ajax return ?

hregis:
https://forum.dolibarr.de/forum/t/problem-mit-netto-rechnung-fuer-eu-kunde/9733

sonikf:
Hi @hregis @eldy
The client in this case is from an EU country so this is a reverse charge and vat must always be 0.
Imo setting TTC is not a problem since ht=TTC and PDF lines do not show TTC but changing vat percent other than 0 for EU customers shouldn't be allowed.

hregis:
@eldy @sonikf here the problem is that if you have a product with a price defined in TTC base with a VAT rate of 20% and you sell to an EU or non-EU customer, the HT is not taken into account, the line is recorded in TTC! That's why I tell myself that when we select a defined product, whether it is in TTC or HT base we should always have the HT informed during the selection (ajax return)

hregis:
@eldy @sonikf look if you have a product with price base type in TTC

Capture d’écran 2025-01-30 à 00 21 01

and if you add a line for an EU customer with VAT intracom

Capture d’écran 2025-01-30 à 00 22 49

Capture d’écran 2025-01-30 à 00 23 07

the real price is 125€ !!! not 150€

hregis:
the ajax return should only provide the HT field and not the TTC field
with my fix the HT field is 125€

sonikf:
You are right. Sorry i was lost in translation seeing forum post and didn't realize the price was incl tax thought was different products!!!
Do you agree though that vat percent change in reverse charge cases should be disabled?

hregis:
@sonikf yes if there is no exception in the rule 😄

zeug-it:
Hi, 
there are two variables possible with this fix:

MAIN_NO_INPUT_PRICE_WITH_TAX
MAIN_SET_CONTENT_INPUT_PRICE_WITH_TAX_DISABLED

which one should be set or not set?

with MAIN_NO_INPUT_PRICE_WITH_TAX I get only NET price displayed - but get NET and with Tax price on in the PDF 

This works for B2B but I need the option like before to change to NET + TAX (Brutto) price for private customers

eldy:
When a price has been defined to be a price including tax, the prixe must be recorded including tax. And it is the prices without tax that will vary depeding on we add tax or not.

I think the trouble is only in edit. When we edit, if the price on product is tax included, the suggested price in form must be with tax, if not,the suggested price must be price without. But currently, we do not save the unit price including tax into the field subprice_ttc (we only fill subprice). So we always suggest the price exluding tax when we edit. This feature was never implemented.

So first thing to do is to update code in develop to store the unit price including tax into  subprice_ttc when we enter a price including tax. So we are able to autofill the field price including taxt when we edit the line.

hregis:
@eldy 

The problem is that there can be two visions of things when we define a price on a tax-inclusive basis.
Either we want this final price whatever the value of the VAT and in this case the VAT-inclusive becomes excluding VAT in the documents.

Product Price: 150 including VAT with a rate of 20%
Sold: 150 excluding VAT with a rate of 0%

Or we can think that when we sell a product with a tax-inclusive base with a VAT rate of 0, then we will have the price excluding VAT in the documents.

Product Price: 150 including VAT with a rate of 20%
Sold: 125 excluding VAT with a rate of 0%

hregis:
@eldy with my PR you can manage the 2 cases : 

without MAIN_SET_CONTENT_INPUT_PRICE_WITH_TAX_DISABLED = case 1
with MAIN_SET_CONTENT_INPUT_PRICE_WITH_TAX_DISABLED = case 2

zeug-it:
From my Point of View: 
We import the selling Prices from Priceslists where we get the RRP (with Tax) from the Distributor and import this as selling price. On the other side we import the NET buying price. So the net selling price is no option in retail.

hregis:
@zeug-it what is the price base type of your customer prices?

zeug-it:
@hregis  The base price is the with tax Brutto price 
image

hregis:
@zeug-it and if you create an invoice for an UE or not UE with 0% you would brutto or netto prices?

zeug-it:
@hregis  for a company customer in EU with TAX ID we create a netto price invoice, because the customer has to pay the tax in his own country.  0% netto price

 for a company customer in EU without TAX ID we create a brutto price invoice (German Tax), because the customer has to pay the tax in his own country. 19% brutto price

invoice for non EU Countries 0% nettp price

hregis:
@eldy ok donc il y a un soucis quand un produit a une base de prix ttc et qu'on vend à TVA 0%
Ou alors il faut prendre en compte les deux cas de figure

eldy:
Yes, there is a trouble:
"Currently, we do not save the unit price including tax into the field subprice_ttc (we only fill subprice). So we always suggest the price excluding tax when we edit."

So to fix this, we must first 
- implement storing into the database field subprice_ttc when we enter a line with a price including tax.
- then it could be possible to fix correctly this ticket.

hregis:
Closing given the discussion below (never reached a conclusion) and since the code has moved on since:

- The line this PR touched in `objectline_create.tpl.php` doesn't exist anymore, the file was refactored to build its JS config from a `jsConf.conf` object.
- The prerequisite you asked for ("we must first implement storing into subprice_ttc when we enter a line with a price including tax") has been done since — `subprice_ttc` is now stored to preserve the TTC entry mode.
- But the actual inconsistency causing the original bug report (150€ charged instead of 125€ HT for an EU reverse-charge customer) is still there today, and it's not in the template: it's in `product/ajax/products.php`. When the `addalsovatforthirdpartyid` block recomputes the VAT rate applicable to the buyer (0% for intracom), it only updates `tva_tx`/`default_vat_code` — it never recomputes `price_ht`/`price_ttc` to stay consistent, so the ajax response can return an inconsistent triple (e.g. HT=125, TTC=150, VAT=0%).

Fixing that needs picking one of the two philosophies from the thread (final TTC price fixed vs HT fixed, VAT applied on top), which was never settled. Rather than hardcoding either, would you be open to a hidden constant on `product/ajax/products.php`'s recompute — undefined keeps today's behavior (TTC-fixed, current triple as-is), and when set, use the HT-fixed method instead (recompute price_ttc from price_ht and the buyer's actual VAT rate)? That would let each install pick, without forcing a global behavior change, and without touching the template this PR was on.

PR #32872, opened 2025-01-30
Bug Fix: #32206

pratushraj:


mdeweerd:
FYI: This PR change is what is causing the DiscountTest::testDiscountCreate test to fail (in ci).

hregis:
fixed : https://github.com/Dolibarr/dolibarr/pull/32984

pratushraj:
@eldy It is a bug fixed, So it'll also merged in 20.0

PR #32878, opened 2025-01-30
Qual: ci 21.0 - Improve lang test + message in case of test exception

mdeweerd:
# Qual: Improve lang test + message in case of test exception

This fixes the language test enabling the error message itself. I also improved the Exception handling so that a backtrace is shown to help find where it happened (here the error message was empty which was strange and made it harder to find - the (empty(val)?a:b) was lacking parentheses for proper concatenation


mdeweerd:
@eldy This is a partial fix for the language test case in ci.
It fixes the issue with showing what the language file issue is about.
However, the language files still need to be fixed to have ci pass.

mdeweerd:
The issues in the language files are all the same:

Found a sequence tag 'notranslate' in the translation file ca_ES/website.lang

PR #32880, opened 2025-01-30
FIX If color is light, we force color to dark

hregis:


hregis:
@eldy with the eldy theme (background is white) : 

Capture d’écran 2025-01-30 à 20 11 59

hregis:
without this fix : 

Capture d’écran 2025-01-30 à 19 29 01

hregis:
@eldy but for me the better is : 

```
// Check contrast with background and correct text color
$forced_color = 'categtextblack'; // default with eldy theme (eldy theme is default theme !), the background is white
if (!empty($this->color)) {
	if (!colorIsLight($this->color)) {	// If color is not light, we force color to light
		$forced_color = 'categtextlight';
	}
}
```

eldy:
On which screen do you have this ?
Because by default on most screen with eldy them, when you have a tag output you should get this:
image

The rule for tag rendering is to have it with the picto "tag" before and tag is inside a backgrounded bar.

hregis:
@eldy my PR is for an extrafield "link" with value : 

`Categorie:categories/class/categorie.class.php:0:(t.type:=:0)`

without my PR : 

Capture d’écran 2025-01-30 à 19 29 01

with my PR :

Capture d’écran 2025-01-30 à 19 31 09

maybe a better fix ?

hregis:
@eldy another "problem", don't fixed, with an extrafield "Liste issue d'une table (n choix)", 
by example with value : "`categorie:label:rowid::`"

Capture d’écran 2025-02-02 à 06 46 07

eldy:
I fixed the trouble into v20.

Issue #32835, opened 2025-01-28
Possible deletion of the parent product while deletion of children is impossible


lmag:
### Bug

Deletion of lot/serial impossible

![Image](https://github.com/user-attachments/assets/d1b06923-6279-4ba3-a7e2-762fdfa5cef2)
Because of stock and mouvement, that is normal for me

But you can delete the parent product and all information, stock, lot/serial etc...

https://github.com/user-attachments/assets/14c9ce6a-9d25-4d7f-a7e9-66726c21a566

### Dolibarr Version

19.0.3

### Environment PHP

7.44

eldy:
Why not deleting a product ?
If you say "i don't want to manage this product any more", it means "i don't want to see this product and i don't want to see its stock (i don't mind stock of this product anymore), and i don't mind to see what was its different possible lot/serial.
So why not allowing this ?
You can't delete a lot if there is some lot currently used.
But when you delete a product, there is no more lot used for this product, since we want to delete it completely.

In most cases, you don't want to delete a projet, you want to disable it because you want to keep the history. But when you say "i want to delete it so to remove it completely from database, it also means you don't want to follow its stock anymore", no ?

PR #32884, opened 2025-01-31
fix: #32835, should not be able to delete product with stock (aka batch lot)

FHenry:
As describe in issue #32835, the delete product when there is stock.


lvessiller-opendsi:
I'm ok to add a protection before deleting a product if there is stock movement

**Before**
- we can delete the product (manage with batches) even if there was some stock movement for its batches -> **it's dangerous**
- we can't delete batches if there was some stock movement

**After**
- we can't delete the product if there is movement on batches --> **it's a protection**
so the delete button is disabled on product card
image

lvessiller-opendsi (approved):

lmag (approved):

rycks (approved):
ok for me one more protection is better

(i explain to my users to not "delete a product" but disable it ...)

PR #32889, opened 2025-01-31
Fix #32888

### LINKED ISSUE #32888: Ticket list shows all tickets instead of open tickets only when selected from left menu
### Bug

If you open the ticket list from left menu all tickets are shwon instead of open tickets only

### Dolibarr Version

20.0

### Environment PHP

PHP 8.x

### Environment Database

n/a

Enable Ticket module
Select Top Menu Ticket
Select Ticket List from left menu

All tickets are shown instead of open tickets only

### Attached files

_No response_

glu000:
FIX #32888
Ticket list shows all tickets instead of open tickets only when selected from left menu

### DIFF
--- a/htdocs/ticket/list.php
+++ b/htdocs/ticket/list.php
@@ -108,9 +108,9 @@ if (!$sortorder) {
 	$sortorder = "DESC";
 }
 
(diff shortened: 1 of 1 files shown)


hregis:
@glu000 it's not a bug, you can define "openall" in the default values

PR #32898, opened 2025-01-31
add test on null for situation_percent

rycks:
there is a race condition with situation_percent var : it could contain "null" value ...


hregis:
Hello @rycks it's not more simple ?

`if (empty($situation_percent) || $situation_percent > 100) {`

PR #32901, opened 2025-02-01
Qual: Fix notices, update baseline, enable PhanTypeMismatchArgument

mdeweerd:
# Qual: Fix notices, update baseline, enable PhanTypeMismatchArgument

- Fix phan notices mostly in relation with supplier invoices,
- Fix phan notices in relation with formquestion (phpdoc update, casts).Update the baseline for fixes.
- Enable PhanTypeMismatchArgument as this is being fixed and to avoid
that new cases occur in files where these are fixed.
  Only about 3800 cases remain, was more than 12000 when starting with Phan.


mdeweerd:
Note: this PR will be split in 2 and is based on #32842 so will convert this as "Ready" when that is integrated.
In practice this PR is ready.

PR #32907, opened 2025-02-02
FIX problem of duplicate data with Multicompany

hregis:
@eldy why to move my PR to develop ? https://github.com/Dolibarr/dolibarr/pull/32870

This PR fixes the filtering of exchange rates by entity

### DIFF
--- a/htdocs/admin/multicurrency.php
+++ b/htdocs/admin/multicurrency.php
@@ -130,9 +130,9 @@ if ($action == 'add_currency') {
 	}
 } elseif ($action == 'setapilayer') {
 	if (GETPOSTISSET('modify_apilayer')) {
-		dolibarr_set_const($db, 'MULTICURRENCY_APP_ID', GETPOST('MULTICURRENCY_APP_ID', 'alpha'));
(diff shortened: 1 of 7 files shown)


eldy:
We must make a difference between a fix of a standard feature that is available by default or that has worked one day and become broken, and a change to make something working that has never work correctly.
In the second case, this is not a maintenance bug fix of a broken feature, but a change to finish the development of an implementation that was not yet completed. In such a case, there is 2 direction: If change is very light and with no risk to impact something else (for eemple 1 or 2 lines of code changed), we can accept to merge it into an already released maintenance branch. In any other case, we must push in develop. 
The exception is if we announce the feature as fully available into the Changelog of the version, in this case, even if the implementation was never completed, we must consider it as a bug fix qualified for the maintenance branch.
All things that are not yet working in a maintenance branch are not necessarly maintenance fix, above all changes that are planned to be done on several years.

hregis:
@eldy without this PR we cannot configure the update of exchange rates in the child entities with Multicompany. Moreover, once configured, we saw the exchange rates in duplicate for all entities. These modifications only filter by entity, it does not change the initial operation in any way.
For the rest, I just put the code in order and put double quotes instead of simple quotes

hregis:
@eldy c'était juste une correction !!!! ça ne changeait en rien le fonctionnement initial !!

PR #32918, opened 2025-02-03
NEW: Add a cancel status on interventions

noec764:
The canceled status has been added to the interventions. 
Indeed, I've added a user permission to cancel an intervention, the mechanics allowing cancellation and reopening, as well as list filters.


htdocs/fichinter/class/fichinter.class.php
```diff
@@ -712,6 +717,111 @@ public function setValid($user, $notrigger = 0)
...
+		}
+	}
+
+
+	/**
+	 *	Tag the fichinter as validated (opened)
+	 *	Function used when fichinter is reopend after being closed.
+	 *
+	 *	@param      User	$user       Object user that change status
+	 *	@return     int         		<0 if KO, 0 if nothing is done, >0 if OK
```
frederic34, line 775: il faut ajouter du texte devant le < 0 if KO, par exemple Return integer < 0 if KO etc

PR #32932, opened 2025-02-04
FIX wrong file path + avoid warning

hregis:


htdocs/install/upgrade2.php
```diff
@@ -4135,7 +4135,7 @@ function migrate_delete_old_files($db, $langs, $conf)
 		//print '<b>'DOL_DOCUMENT_ROOT.$filetodelete."</b><br>\n";
 		if (preg_match('/\*/', $filetodelete) || file_exists(DOL_DOCUMENT_ROOT.$filetodelete)) {
 			//print "Process file ".$filetodelete."\n";
-			$result = dol_delete_file(DOL_DOCUMENT_ROOT.$filetodelete, 0, 0, 0, null, true, false);
+			$result = dol_delete_file(DOL_DOCUMENT_ROOT.$filetodelete, 0, 1, 0, null, true, false); // nophperrors to avoid false positive with asterisk
```
eldy, line 4138: If we change this, if deletion fails because of permission for example, we won't we aware. 
And we must be aware here if a deletion fails to show the error to ask the user to delete the file manually.

PR #32934, opened 2025-02-04
FIX delete file with mask

lvessiller-opendsi:
FIX delete file with mask
- I have an error when I want to upgrade the new version of Dolibarr (develop only)
image

If we check the upgrade2.php script at line 4123, we see the script has a mask to delete several files :
- /includes/nusoap/lib/class.*

And the function "dol_delete_file()" can have several files as mask in the first parameter (see in "files.lib.php")

I don't know if it's the good way to fix it but I wonder why if no files found there is an error on line 1591 of "file.lib.php" : 
```
$ok = false; // to avoid false positive
dol_syslog("No files to delete found", LOG_DEBUG);
```

and the syslog in not a Warning or Error.

### DIFF
--- a/htdocs/core/lib/files.lib.php
+++ b/htdocs/core/lib/files.lib.php
@@ -1588,7 +1588,7 @@ function dol_delete_file($file, $disableglob = 0, $nophperrors = 0, $nohook = 0,
 					}
 				}
 			} else {
-				$ok = false; // to avoid false positive
+				$ok = true;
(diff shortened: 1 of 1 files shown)


eldy:
A better fix was submitted and merged.

hregis:
`$ok = false; // to avoid false positive`

@lvessiller-opendsi 
I added this because if a file path is not correct it said the file was deleted even if it couldn't find the file: https://github.com/Dolibarr/dolibarr/pull/32811

PR #32939, opened 2025-02-04
remove superfluous characters

frederic34:


hregis:
@frederic34 heuuuu ! 😄 

`(GETPOSTISSET('usage_task') ? (GETPOST('usage_task', 'alpha') ? ' checked="checked"' : '') : ' checked="checked"')`

it's always `checked="checked"` no ?

frederic34:
yes it will be another correction 🪜

PR #32940, opened 2025-02-04
NEW: takepos cash control: add vat taxless total

marc-dll:
Before :

Capture d’écran de 2025-02-04 16-14-35

After :

Capture d’écran de 2025-02-04 16-14-43


eldy:
How does the ticket look if there is both a vat rate 20% and another vat rate 10% ?

marc-dll:
@eldy Here it is with multiple VAT rates :

Capture d’écran de 2025-02-06 15-07-19

PR #32931, opened 2025-02-04
Qual: Fix phan notices (intracomm, knowledgemgmt, loan,main.inc,master.inc,eldy.lib)

mdeweerd:
# Qual: Fix phan notices (intracomm, knowledgemgmt, loan,main.inc,master.inc,eldy.lib)

PR #32942, opened 2025-02-05
Qual: Fix argument types in calls to some PDF functions

mdeweerd:
# Qual: Fix argument types in calls to some PDF functions

Fixed several argument typs in PDF function calls.
Change 0/1 into false/true with fixers using phan (~1000 phan ParamType mismatch notices solved).


mdeweerd:
Steps leading up to this PR (included):

- #32917
- #32921
- #32931
- #32942

PR #32954, opened 2025-02-05
FIX Clone propale with update prices

MathildeMar:
# FIX|Fix #[*Prices don't update when clone a propale with update prices*]
Prices didn't update when clone a propale with update prices :
Fixed test on update_prices and update_desc parameters when calling CreateFromClone() function.


hregis:
@MathildeMar i send a backport to 20 branch, thanks : https://github.com/Dolibarr/dolibarr/pull/32955

PR #32955, opened 2025-02-05
FIX Prices didn't update when clone a propale with update prices

hregis:


MathildeMar:
Hello,
It's not the good way to get the value of update_prices and update_desc. 
Sorry for the mistake. It would be :
$result = $object->createFromClone($user, $socid, (GETPOSTISSET('entity') ? GETPOSTINT('entity') : null), (GETPOST('update_prices')=='on' ? true : false), (GETPOST('update_desc')=='on' ? true : false));

hregis:
@MathildeMar yes you're right, thank you

PR #32959, opened 2025-02-05
Qual: Fix phan notices (modulebuilder)

mdeweerd:
# Qual: Fix phan notices (modulebuilder)

Note @eldy 
`get_substitutionarray_each_var_object()`  is for an array of string-indexed objects in the first argument.
But in the code it is often (always?) used with object being a child of CommonObject (not an array).

What to do?
- Delete this line (here, and elsewhere)?
- Convert it to an array?  Which index key?
- ... .

Example:
https://github.com/Dolibarr/dolibarr/blob/7e3c11c285e4fb2c6612bacc45d274daf42aba12/htdocs/modulebuilder/template/core/modules/mymodule/doc/doc_generic_myobject_odt.modules.php#L399


eldy:
I have fixed the bug in get_substitutionarray_each_var_object() that was introduced into v21.
We should accept both array and object.

mdeweerd:
Ok, I updated the baseline but there are some things to fix in the code in ci.

PR #32961, opened 2025-02-06
fix(facture): removed useless fetch product

atm-irvine:
# FIX|Fix removed useless fetch product
I remove this piece of code because it fetches a product for nothing and didn't handle after


atm-irvine:
@eldy I moved this PR here : https://github.com/Dolibarr/dolibarr/pull/33238

PR #32962, opened 2025-02-06
fix missing translations

frederic34:


htdocs/website/index.php
```diff
@@ -64,7 +64,7 @@
  */
 
 // Load translation files required by the page
-$langs->loadLangs(array("admin", "other", "website", "errors"));
+$langs->loadLangs(array("admin", "other", "website", "main", "errors"));
```
eldy, line 67: The "main" should already be loaded by the require '../main.inc.php'

frederic34:
image

PR #32964, opened 2025-02-06
FIX max file size to download not needed for WebPortal

lvessiller-opendsi:
FIX max file size to download not needed for WebPortal
- the file size in not for downloading PDF documents (no limit)

With this FIX it's the same behavior when you download PDF documents from the back office.

### DIFF
--- a/htdocs/webportal/controllers/document.controller.class.php
+++ b/htdocs/webportal/controllers/document.controller.class.php
@@ -212,15 +212,6 @@ class DocumentController extends Controller
 			exit;
 		}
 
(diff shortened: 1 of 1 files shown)


eldy:
The parameter MAIN_SECURITY_MAXFILESIZE_DOWNLOADED is desgined for download from any public pages or portal (so like the webportal).
It is to avoid some king of cyber attack. The problem is not the option, but the fact that the default value is wrong. No test must be done if option is set to 0.
I pushed a better fix.

PR #32966, opened 2025-02-06
Add class for each option value of an additional select extrafield on project list

iouston:
# NEW|New Add class for each option value of an additional select extrafield on project list


eldy:
To match consistency with fields in module builder and to follow the use of css of extrafields, a different R has been merged.
We have 3 CSS already. One to output a variable in view mode, one for create/update and one for the td column.
We must not use the one for td column for the view mode, so i pushed a different PR.
Thanks to check it matches your need.

iouston:
I understand. This is ok for me

PR #32976, opened 2025-02-07
NEW: add const to not include default lib_head.js in top_htmlhead with DISABLE_LIB_HEAD_JS

### LINKED ISSUE #32975: Missing only "disable lib_head" on main.inc.php - Add variable DISABLE_LIB_HEAD
### Feature Request

Allow to disable lib_head

### Use case

Disable theme CSS on main.inc.php - lib_head

### Suggested implementation

- [ ] Add variable : lib_head
- [ ] Add use in main.inc.php

FHenry:
fix: #32975 

When custom module add page, it would be nice to not always include lib_head.js  (may be confilct with other UI js lib in module)


htdocs/main.inc.php
```diff
@@ -2216,7 +2216,9 @@ function top_htmlhead($head, $title = '', $disablejs = 0, $disablehead = 0, $arr
 
 			// Global js function
 			print '<!-- Includes JS of Dolibarr -->'."\n";
-			print '<script nonce="'.getNonce().'" src="'.DOL_URL_ROOT.'/core/js/lib_head.js.php?lang='.$langs->defaultlang.($ext ? '&amp;'.$ext : '').'"></script>'."\n";
+			if (!defined('DISABLE_LIB_HEAD_JS')) {
```
eldy, line 2219: If you disable this lib, all the rest of application will hang due to js error.
What is the goal of disabling it. As it is just js function, you should not experience conflict ?

nicolas-eoxia: for custom page like public interface you don't always want JS function in lib_head like date

PR #32979, opened 2025-02-07
fk_user_author instead of user_author into update

rycks:
user_author is deprecated and fk_user_author is used everywhere into the class BUT not in sql update statement !

then sometimes you could have that sort of bug, described into french forum https://www.dolibarr.fr/forum/t/bug-profil-creation-facture/48191/14

image

and 

image


mdeweerd:
I created the DeprecationHandler for handling this type of situations, but it needs to be adopted....

https://github.com/Dolibarr/dolibarr/blob/c316d1e5f38cd1c89c8736c2c5a4b31bd7ba16c3/htdocs/core/class/doldeprecationhandler.class.php#L27

fk_user_author is also deprecated accourding to its phpdoc... .

PR #32983, opened 2025-02-07
FIX #32981 bad dispatched quantities for batches on shipment card

### LINKED ISSUE #32981: Bad dispatched quantities for batches on shipment card
### Bug

Bad dispatched quantities for batches on shipment card

### Dolibarr Version

18.0 and develop

### Environment PHP

PHP 7.4

### Environment Database

Mysql

1. Create an order with a batch product (serial number only)
Image

2. Put stock for batch product number on different warehouses
Image

3.  Create an order with this batch product and set quantity to 3 : 
Image

4. Create a shipment from batch (without selected a warehouse)
Image

5. Then go to the dipatch tab of this new shipment
Image

The quantities of dispatched batches are not the same as step 4. when you choose each batch product for each warehouse.
The batches quantities are wrong on dispatched tab of this shipment card.

### Attached files

_No response_

lvessiller-opendsi:
FIX #32981 bad dispatched quantities for batches on shipment card


eldy:
Fix seems already embeded into v22, provided by commit c85cc3061bd52ae71dbf0a2f3a4a339f5c680d7e of PR https://github.com/Dolibarr/dolibarr/pull/33750

PR #32990, opened 2025-02-08
 Qual: Fix phan notices (multicurrency,opcache,opensurvey,partnership)

mdeweerd:
# Qual: Fix phan notices (multicurrency,opcache,opensurvey,partnership)


htdocs/opensurvey/class/opensurveysondage.class.php
```diff
@@ -60,6 +60,13 @@ class Opensurveysondage extends CommonObject
 	public $date_m;
 
 
+	/**
+	 * Lines of the survey - Note: the type differs from CommonObjectLine[] !
+	 *
```
mdeweerd, line 65: @eldy OpenSurvey does not return a list of objects in its $lines property.

This the $lines property does not seem to be used at the moment, so there is a possibility to change the name of the property.
If not, a new subclass is needed.

I can add an entry in the phpstan baseline to ignore this for now.

PR #33003, opened 2025-02-10
FIX: Gravatar picture don't appear on top right

### LINKED ISSUE #32970: Gravatar picture - Don't appear on top right
### Bug

Hello,

I notice my gravatar image appears on my user page, on the event line on different pages 
... but not on the image on the top right. 

### Dolibarr Version

20.0.3

### Environment PHP

8.3.15

### Environment Database

MySQL or MariaDB 10.6.20-MariaDB-cll-lve

Activate Gravatar module.

Also, I tried to manually upload a picture.
And to delete the picture after.
At this moment, the image on right top seems used my gravatar image.

And soon after, this right top image return to default image.

### Attached files

_No response_

NefiteTifall:
# FIX|Fix #32970
Add Gravatar image detection, if the user has a Gravatar image and they have enabled the module then we display it, otherwise we fall back to nophoto URLs

🩵@defrance

### DIFF
--- a/htdocs/main.inc.php
+++ b/htdocs/main.inc.php
@@ -2608,7 +2608,6 @@ function top_menu_user($hideloginname = 0, $urllogout = '')
 		if ($user->gender == 'woman') {
 			$nophoto = '/public/theme/common/user_woman.png';
 		}
-		
 		if (isModEnabled('gravatar')) {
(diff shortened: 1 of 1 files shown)


eldy:
Thanks for pointing the bug and solution. 
I pushed a better fix. Thanks to check it's ok for you.

PR #33007, opened 2025-02-10
Fix: Prevent warnings and fatal error in admin/mrp.php

AnthonyVallenet:
- Fixed "Undefined property: stdClass::$MRP_MO_ADDON_PDF" by checking existence with !empty().
- Fixed "Class 'Product' not found" by adding require_once for product.class.php.
- Improves stability and prevents unexpected crashes in MRP module.

❤️ @defrance


htdocs/admin/mrp.php
```diff
@@ -397,7 +398,7 @@
 
 								// Default
 								print '<td class="center">';
-								if ($conf->global->MRP_MO_ADDON_PDF == $name) {
+								if (!empty($conf->global->MRP_MO_ADDON_PDF) && $conf->global->MRP_MO_ADDON_PDF == $name) {
```
defrance, line 401: better way is to use getDolGlobalString("MRP_MO_ADDON_PDF")

PR #33011, opened 2025-02-10
Update html.formmargin.class.php

homer8173:
Under PHP 8 , can't add string and float on following lines. So i propose to convert to avoid errors

But i don't understand why database double(24,8) are fetched as string

### DIFF
--- a/htdocs/core/class/html.formmargin.class.php
+++ b/htdocs/core/class/html.formmargin.class.php
@@ -99,7 +99,7 @@ class FormMargin
 				$line->pa_ht = $line->subprice * (1 - ($line->remise_percent / 100));
 			}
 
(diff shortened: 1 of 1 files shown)


eldy:
> Under PHP 8 , can't add string and float on following lines. So i propose to convert to avoid errors
> 
> But i don't understand why database double(24,8) are fetched as string

Data are retreived doing SQL request. A SQL request is a full text protocol. Question is full text and answer is full text. So all is text, even dates. This is goal of the requester to then convert them into the correct type (for example with db->jdate() for date or casting to int or float for numeric. However sometime the developer did not add this into the code after its fetch and format in memory is still text instead of numeric or date...

PR #33025, opened 2025-02-11
FIX: Time spent for timespent PDF

altairis-noe:
# FIX: Time spent for timespent PDF

Since V18 in the function getTasksArray() `$tasks[$i]->duration = $obj->duration_effective` is replaced by `$tasks[$i]->duration_effective = $obj->duration_effective`. We must therefore refer to `$object->lines[$i]->duration_effective` in the timespent PDF model

Before:
image

After:
image


mdeweerd:
https://cti.dolibarr.org - Maybe you can fix a similar issue in odt outputs.

{48595BF7-115D-4986-A84A-F02432D2AF02}

altairis-noe:
That should do the trick, @mdeweerd!

PR #33028, opened 2025-02-11
 Qual: Fix phan notices (public-1)

mdeweerd:
#  Qual: Fix phan notices (public-1)


mdeweerd:
@eldy Not sure why you closed this, but it's included in #33029 .

PR #33040, opened 2025-02-12
Fix ci: Fix RedefineFunction in newonlinesign

mdeweerd:
# Fix ci: Fix RedefineFunction in newonlinesign

This is in complement to other fixes in #33034.
Not everything is fixed here.

Only fixing so that it is inline with #33028 and #33029


frederic34:
i pushed other fix in branch 21

PR #33041, opened 2025-02-12
NEW: Add status in module Resources

prietojc:
Add Status into card and list of module Resources


hregis:
@prietojc move your PR in develop branch...

prietojc:
Checks fails in constants, there is a conflict with develop branch.... but, why that?

JonBendtsen:
Was the changes in this PR ever completed? perhaps in another PR?

JonBendtsen:
@eldy  I think it would be nice if we could have this status field

PR #33046, opened 2025-02-13
fix: Remove undefined $buttonreconcile

RemiLecas:
image

Remove undefined $buttonreconcile

@defrance

### DIFF
--- a/htdocs/compta/bank/releve.php
+++ b/htdocs/compta/bank/releve.php
@@ -310,8 +310,6 @@ if (empty($numref)) {
 
 		print dol_get_fiche_end();
 		
-		$buttonreconcile = '';
-
 		/* Moved as a tab
 		if ($object->canBeConciliated() > 0) {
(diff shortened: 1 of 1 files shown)


htdocs/compta/bank/releve.php
```diff
@@ -352,7 +354,7 @@
 		$massactionbutton = $form->selectMassAction('', $arrayofmassactions);
 
 		$morehtml = '';
-		if ($action != 'addline' && $action != 'reconcile') {
+		if ($action != 'addline' && $action != 'reconcile' && $buttonreconcile) {
```
mdeweerd, line 357: I would just move this part into the code section that was commented out.
That is what made this error appear and now makes this code "useless".
However, move $morehtml to the location where `$buttonreconcile = ''` that no longer needs to be added.

aspangaro:
Button reconcile as moved in other tab in recent version

PR #33048, opened 2025-02-13
FIX: handle undefined cookie value for virtual card expansion, refact…

Maxenor:
Fix "Undefined array key" warning and remove debug output in virtualcard.php

Summary

Replaced direct access to $_COOKIE['virtualcard_expand'] with the null coalescing operator to avoid the "Undefined array key" warning.

Why

Undefined array key warning: Accessing $_COOKIE['virtualcard_expand'] directly triggered a warning if the cookie didn’t exist.

@defrance

### DIFF
--- a/htdocs/user/virtualcard.php
+++ b/htdocs/user/virtualcard.php
@@ -52,7 +52,7 @@ if (empty($id) && empty($ref)) {
 	$id = $user->id;
 }
 
-$expand = $_COOKIE['virtualcard_expand'];
+$expand = $_COOKIE['virtualcard_expand'] ?? null;
 
 $object = new User($db);
 if ($id > 0 || !empty($ref)) {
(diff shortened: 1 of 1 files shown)


htdocs/user/virtualcard.php
```diff
@@ -17,6 +17,7 @@
  * along with this program. If not, see <https://www.gnu.org/licenses/>.
  */
 
+ob_start();
```
eldy, line 20: Use of ob_start lead to difficulty to understand and debug.
We should not need this for common pages.
Can you describe why you need this ?

Maxenor: Hello, there were multiple errors displaying on the virtual card page with headers already sent by the main.inc.php file. It helps prevent "headers already sent" errors by delaying the actual sending of output until the page's ready, allowing to modify headers at any point before the flush.

hregis: @Maxenor and if multiple errors... fix this errors ! ;-)

PR #33058, opened 2025-02-13
Fix: Hide empty categories table

RemiLecas:
image

Hide empty categories table when $cats < 0

@defrance


eldy:
To follow the GUI rules, when nothing is found, it is better to show the table with a line
print $langs->trans("NoRecordFound") instead of hiding completely.

aspangaro:
The interface has changed; this block no longer exists. This PR is no longer needed.

PR #33062, opened 2025-02-13
[FIX] Remove xdebug_print_function_stack() call to suppress verbose XDebug output in non-production environments

kbegot:
Removes the verbose xdebug_print_function_stack() call in non-production environments. Now, only targeted debug info (file, line, function) is output when applicable.
@defrance


hregis:
@kbegot why remove this ?

PR #33064, opened 2025-02-13
fix: cast timestamp to int

kbegot:
Ensure that $time is cast to an integer before calling DateTime::setTimestamp to avoid a TypeError.

@defrance


htdocs/core/lib/date.lib.php
```diff
@@ -166,7 +166,7 @@ function dol_time_plus_duree($time, $duration_value, $duration_unit, $ruleforend
 	if (getDolGlobalString('MAIN_DATE_IN_MEMORY_ARE_GMT')) {
 		$date->setTimezone(new DateTimeZone('UTC'));
 	}
-	$date->setTimestamp($time);
+	$date->setTimestamp((time) $time);
```
frederic34, line 169: @kbegot `$date->setTimestamp((int) $time);`

PR #33066, opened 2025-02-13
Fix: Memory exhaustion issue in dol_htmlentitiesbr()

GregoireIvars:
FIx: Memory exhaustion when comment in a ticket is too long #33065
## Description : 
When writing a message or comment in a ticket, if it turns out to be too long, a memory error occurs. I have implemented a fix for this case

This PR fixes a memory exhaustion issue in the dol_htmlentitiesbr() function.

 - Increased memory limit to prevent crashes.
 - Added input size verification to avoid excessive memory usage.
 - Logged input size for debugging large strings.
@defrance


htdocs/core/lib/functions.lib.php
```diff
@@ -8723,6 +8723,9 @@ function dol_htmlentitiesbr($stringtoencode, $nl2brmode = 0, $pagecodefrom = 'UT
 		$newstring = strtr($newstring, array('&' => '__PROTECTand__', '<' => '__PROTECTlt__', '>' => '__PROTECTgt__', '"' => '__PROTECTdquot__'));
 		$newstring = dol_htmlentities($newstring, ENT_COMPAT, $pagecodefrom); // Make entity encoding
 		$newstring = strtr($newstring, array('__PROTECTand__' => '&', '__PROTECTlt__' => '<', '__PROTECTgt__' => '>', '__PROTECTdquot__' => '"'));
+		if (strlen($stringtoencode) > 1000000) {// limit, to avoid memory problem
+			return htmlentities(substr($stringtoencode, 0, 1000000)) . '...';
```
eldy, line 8727: Using simple htmlentities will lead to security vulnerability.

Also the only thing we have after is return, so I don't see why returning a string instead of another (both are already in memory) generates a memory error

This means the $newstring is too large, so we must investigate what happen to have this memory error.

hregis:
@GregoireIvars @defrance it's not a limitation of your server ? Maybe add a constant for this ?

hregis:
@eldy @GregoireIvars @defrance je ne comprend pas cette limitation ?! est-ce que qu'au fond ça ne dépend pas des limitation du serveur ? est-ce que ceci ne va pas limiter ceux qui ont la puissance de gérer ceci ? (je dis ça je dis rien !) ;-)

eldy:
> @eldy @GregoireIvars @defrance je ne comprend pas cette limitation ?! est-ce que qu'au fond ça ne dépend pas des limitation du serveur ? est-ce que ceci ne va pas limiter ceux qui ont la puissance de gérer ceci ? (je dis ça je dis rien !) ;-)

Yes this looks strange. When there is a memory limit reached, there is a php fatal error, not an infinite delay with no answer.
@GregoireIvars do you know what was size of the string before truncation ? And what is your php max size of memory ?

JamBalaya56562:
The root cause behind #33065 has been identified and fixed in #39260, which was merged into develop.

As eldy suspected above, this was **not** a memory-exhaustion problem — that is exactly why the symptom was an endless hang rather than a PHP fatal error. The real cause was a quadratic (`O(n²)`) loop in `dol_escape_htmltag()`, which burns CPU until `max_execution_time` is reached while memory usage stays flat. #39260 replaces that loop with a single `preg_replace_callback()` pass per tag (identical output, ~7–69× faster depending on size).

Because the slowdown does not come from `dol_htmlentitiesbr()` (which is linear), the 1,000,000-character truncation proposed here should no longer be needed. The truncation would also only cover the `dol_textishtml()` branch and would affect all ~240 call sites of `dol_htmlentitiesbr()`, including PDF and e-mail generation.

So I think this PR can likely be closed in favour of #39260. Thanks @GregoireIvars for the original report that surfaced the issue!

PR #33069, opened 2025-02-13
Fix : delete logo from disk

KenjiLAO:
@defrance

### DIFF
--- a/htdocs/societe/card.php
+++ b/htdocs/societe/card.php
@@ -750,59 +750,86 @@ if (empty($reshook)) {
 				}
 
 				// Logo/Photo save
-				$dir     = $conf->societe->multidir_output[$object->entity]."/".$object->id."/logos";
-				$file_OK = is_uploaded_file($_FILES['photo']['tmp_name']);
-				if (GETPOST('deletephoto') && $object->logo) {
-					$fileimg = $dir.'/'.$object->logo;
-					$dirthumbs = $dir.'/thumbs';
-					dol_delete_file($fileimg);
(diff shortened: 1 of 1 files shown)


KenjiLAO:
Add code to delete logo from third place from disk

hregis:
@defrance ça serait bien de faire vos tests sur un autre dépôt svp, merci

JamBalaya56562:
This looks already fixed on develop.

The bug — the logo file staying on disk after ticking "Delete" — came from the condition testing `$object->logo`, which is emptied earlier in the same request:

```php
if (GETPOST('deletephoto')) {
    $current_logo = $object->logo;
    $object->logo = '';
}
```

so `if (GETPOST('deletephoto') && $object->logo)` was never true and nothing was removed from disk.

#36197 ("FIX member photo and thirdparty logo not deleted in documents directory", commit a00ff66f45bd, merged 2025-11-11) fixes exactly that, by using the saved name instead:

```diff
-				if (GETPOST('deletephoto') && $object->logo) {
-					$fileimg = $dir.'/'.$object->logo;
+				if (GETPOST('deletephoto') && $current_logo) {
+					$fileimg = $dir.'/'.$current_logo;
 					$dirthumbs = $dir.'/thumbs';
```

On develop today, `societe/card.php` removes the logo file and its `thumbs` directory through `$current_logo`, and the same commit also cleans the related entries in the ECM/database file index.

So the problem this PR targets is covered, and this PR can probably be closed (it is also currently conflicting).

PR #33074, opened 2025-02-13
New/documentation UI/inputs

atm-lena:
# NEW|New 

Issue https://github.com/Dolibarr/dolibarr/issues/30294


eldy:
@frederic34 any idea of what can be wrong with phpstan ?

frederic34:
@eldy phpstan version update ? it's fixed on 2.0.x trying 2.1.x 
https://github.com/Dolibarr/dolibarr/pull/33080

eldy:
> @eldy phpstan version update ? it's fixed on 2.0.x trying 2.1.x #33080

I think i have the point.
The file  install/inc.php is analyzed before the bootstrap_action.php is executed, so the variable DOL_DOCUMENT_ROOT is defined to ".." (by the install/inc.php file) instead of the value into the bootsreap_action.php

The problem is the same when run from command line with
PHP runtime version: 8.3.16
PHP version for analysis: 8.3.16 (from runtime)
PHPStan version: 2.1.5

I succeed in running localy after adding an exclude 
`- htdocs/install/inc.php   into the phpstan.neon.dist file`
like this 01b9ff7f00b1682b8d2e768ef496a05e2c043c3a

I finally disabled all the /install directory to remove completely the error because not loading the install/inc.php generatess error in all files into install/*
and phpstan is now green. Wemay need to find a better solution if we need to restore the scan of /install...

PR #33094, opened 2025-02-16
NEW: Option to set Third Party Customer Code equal to Member Reference

iLLixM:
when using the function of creating third parties from members, you are able to set the customer code equal to the existing member reference by setting the parameter THIRDPARTY_CUSTOMERCODE_EQUALS_MEMBERREF


hregis:
@iLLixM @eldy hum... there will be no conflict with the accounting accounts?

iLLixM:
@hregis Please help me understand your concerns.
In principle, any authorized user can modify the customer code of a third-party object via the web front end (/societe/card.php?socid=x&action=edit&token=xxxx). I assume that this has no effect on the accounting accounts.

The addition I made to the code takes effect when a new customer is created as a third party based on a member's master data. As far as I know, this function can be triggered in two ways
a) directly using the manual "Create a third party" button on the member card (/adherents/card.php?rowid=x&action=create_thirdparty&token=xxxx)
b) if the system parameter ADHERENT_DEFAULT_CREATE_THIRDPARTY is set (the use case I use)

A collision when executing the additional function can occur if an auto-incremented member reference for a new member has already been manually assigned as a customer code for a completely different third party. However, the error message "Customer code already used" is then displayed. A new and linked third party for the newly created member is then not created.

hregis:
It's just a question I was wondering, I'm not an accounting expert 😄 
@aspangaro ?

PR #33096, opened 2025-02-17
dol_eval_new

c3do:
Hello,
Short description for those who have not seen the previous PR.
Rewriting of dol_eval allowing a better compatibility with PHP syntax.
Please read the comments on filters by function name.
Seems fully compatible with v20.
Normally similar to dol_eval on the security side.
To be tested.


c3do:
I didn't do it for dol_eval_new, to clearly show that the function needs it, but it is possible to put the "defined...." lines outside the function to avoid executing them at each call, and improve execution speed.

hregis:
Yes, it is important to gain micro-nano-seconds on increasingly powerful servers.

PR #33106, opened 2025-02-17
FIX - Filling an idate function parameter 

atm-lucasmantegari:
# FIX - filling an idate function parameter 
*It's important to fill in this data because if you don't, the base GMT time zone will be used. For me, this means a 1-hour time difference because I'm in France.*


htdocs/comm/action/class/actioncomm.class.php
```diff
@@ -1187,8 +1187,8 @@ public function update(User $user, $notrigger = 0)
 		$sql .= ", fk_action = ".(int) $this->type_id;
 		$sql .= ", code = " . ($code ? "'".$this->db->escape($code)."'" : "null");
 		$sql .= ", label = ".($this->label ? "'".$this->db->escape($this->label)."'" : "null");
-		$sql .= ", datep = ".(strval($this->datep) != '' ? "'".$this->db->idate($this->datep)."'" : 'null');
-		$sql .= ", datep2 = ".(strval($this->datef) != '' ? "'".$this->db->idate($this->datef)."'" : 'null');
+		$sql .= ", datep = ".(strval($this->datep) != '' ? "'".$this->db->idate($this->datep, 'tzuser')."'" : 'null');
```
eldy, line 1190: We should not have tzuser here (don't forget that in a CLI context a idate with tzuser will return a different value than when used on a web context.
Pb is surely when setting the $this->datep parameter.
Can you describe how to reproduce the bug.

Issue #30742, opened 2024-08-23
Title and Subtotal (In Dolibarr Core) - Specification


FHenry:
### English

**1 - Objectives**
The goal of this feature is to integrate functionalities into the Dolibarr Core for certain documents that manage lines with the following concepts:

    Title
    Subtotal
    Comment

Initial scope:

    Commercial Proposal
    Customer Order
    Customer Invoice
    Supplier Commercial Proposal
    Supplier Order
    Shipment
    Supplier Reception

Secondary scope:

    Refactoring and standardization of Dolibarr's code across all Core objects that manage lines, including:
        Intervention
        Expense Report
        BOM (Bill of Materials)
        Work Order (WO)
    Default integration in the "boilerplate" offered by the Module Builder.

**2 - Technical and Functional Specification**

**2.1 Module Configuration**

The "Detailed Presentation" module has a configuration page.
On this page, it is possible to activate the functionality for:

    Commercial Proposal
    Customer Order
    Customer Invoice
    Supplier Commercial Proposal
    Supplier Order
    Shipment
    Supplier Reception

For each document type:

    Title
    Subtotal
    Comment

Presentation Suggestion

![image](https://github.com/user-attachments/assets/b65b3613-5451-41b3-9f52-46af9c0378e1)

The list of supported documents can be extended via a hook.
The "Number of Levels" column:

    Defaults to 2
    Is empty and non-editable if Title = No and Subtotal = No

PDF Rendering Details

![image](https://github.com/user-attachments/assets/31847d48-45be-439d-b81b-b40fe56ad43b)

**2.2 UI/UX**

Based on the technical and functional specifications:

    Commercial Proposal
    Customer Order
    Customer Invoice
    Supplier Commercial Proposal
    Supplier Order

The line entry allows the user to choose the type or nature of the line.

![image](https://github.com/user-attachments/assets/0d22f7dd-6155-4bb2-92bb-890ccdaffcb9)

The "Subtotal" option only appears if a "Title" exists.

Title, Subtotal, and Comment lines are never transmitted in advanced accounting operations (linking), nor accounted for in reports based on Documents, nor accounted for in Referencing Objects (product/service/third parties).

**2.2.1 Title**

For a line of type "Title":

**2.2.1.1 Creation and Modification**

Presentation Suggestion in Creation and Modification Mode

![image](https://github.com/user-attachments/assets/6a832de0-570b-41a1-810b-d2022131e22a)

Fields:

    Title: Text
[...]

PR #33109, opened 2025-02-18
NEW(subtotals) : subtotals fields

YannisHoareau:
# NEW|New subtotals fields
This PR is to add new fields for the new subtotals module. The new module will use these fields to know when generating a PDF if a line has an option (page break for example).
Ref : https://github.com/Dolibarr/dolibarr/issues/30742


hregis:
Hello @YannisHoareau 
where is this module ?

YannisHoareau:
Hello @hregis this is a new module I'm working on. This PR is to validate that my database changes suits Dolibarr.

htdocs/install/mysql/migration/21.0.0-22.0.0.sql
```diff
@@ -131,3 +131,7 @@ ALTER TABLE llx_categorie_supplier_invoice ADD INDEX idx_categorie_supplier_invo
 
 ALTER TABLE llx_categorie_supplier_invoice ADD CONSTRAINT fk_categorie_supplier_invoice_categorie_rowid FOREIGN KEY (fk_categorie) REFERENCES llx_categorie (rowid);
 ALTER TABLE llx_categorie_supplier_invoice ADD CONSTRAINT fk_categorie_supplier_invoice_fk_supplier_invoice_rowid FOREIGN KEY (fk_supplier_invoice) REFERENCES llx_facture_fourn (rowid);
+
+ALTER TABLE llx_propaldet ADD COLUMN subtotal_options text;
```
eldy, line 135: Can you describe the goal of the field ?

There is already exisitng subtotal modules that works correctly (for example subtotal of ATM) using the existing fields. It should be a good idea to integrate them the same way of working first and then add features.

YannisHoareau:
Hello @eldy. I've reviewed how ATM's module works and I don't think the way the options are managed is really good. I think theses options should be stored with the line they belong to. If we add these fields, the options will be serialized and then stored into the database.

hregis:
@YannisHoareau why not use the "extraparams" field of object to store lines parameters in json format?

marc-dll:
Hi,

@eldy In ATM's subtotal module, those options are stored as extrafields, we deemed that solution inelegant, as extrafields are designed to be used by external modules.

@hregis There are no extraparams on lines. These options are related to lines of objects and not on the objects themselves.

Greetings.

hregis:
@YannisHoareau yes but it's possible to add in this object field the `(array("typeofparams") => (array("lineid" => array("options")))` in json format

or add "extraparams" field in line ! 
for me "subtotal_options" is a field name too specific

eldy:
> @YannisHoareau yes but it's possible to add in this object field the `(array("typeofparams") => (array("lineid" => array("options")))` in json format
> 
> or add "extraparams" field in line ! 
> for me "subtotal_options" is a field name too specific
> 

Yes, if we don't have the field extraparams yet on line, we should (but this is not necessarly related to subtotal need)
For the moment i don't know what kind of value you plan to store in this field, so go with extraparams, when i will see next prs and examples of content, i will tell if we need a dedicated field...

hregis:
@YannisHoareau and maybe use json format instead serialize format ?

YannisHoareau:
@eldy so I will use extraparams that I will add in another PR.

PR #33129, opened 2025-02-19
New add new setup in ai setup

Hystepik:
Add setup of model for each feature by AI type.
Like this we can choose which model to use for each feature
image


eldy:
Because the choice of a model depends on the "feature", it would be better to have the setup of models on the page custom_prompts (Custom prompt and models).

PR #33131, opened 2025-02-19
FIX invoice creation : use dol_include_once instead of require_once to allow external modules.

thomas-Ngr:
# Fix : Impossible to create a supplier invoice from an object introduced by an external module

Loading the class of the origin element should be done through dol_include_once and not require_once, to allow use from external modules.

This is right on most cards:
- compta/facture/card.php
- comm/propal/card.php
- commande/card.php
- fourn/commande/card.php
- fourn/propal/card.php

.. but not on fourn/facture/card.php

### DIFF
--- a/htdocs/fourn/facture/card.php
+++ b/htdocs/fourn/facture/card.php
@@ -1042,7 +1042,7 @@ if (empty($reshook)) {
 					$object->origin_id = GETPOST('originid', 'int');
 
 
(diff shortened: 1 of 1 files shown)


lvessiller-opendsi (approved):

rycks (approved):
as it's already done on other cards it's ok for me

PR #33141, opened 2025-02-19
QUAL setup disable stock calculate on rules

lvessiller-opendsi:
QUAL setup disable stock calculate on rules
- prepare kit feature

### DIFF
--- a/htdocs/admin/stock.php
+++ b/htdocs/admin/stock.php
@@ -200,6 +200,8 @@ if (isModEnabled('productbatch')) {
 	$langs->load("productbatch");
 	$disableStockCalculateOn[] = 'BILL';
 	$disableStockCalculateOn[] = 'VALIDATE_ORDER';
+	$disableStockCalculateOn[] = 'SUPPLIER_BILL';
+	$disableStockCalculateOn[] = 'SUPPLIER_VALIDATE_ORDER';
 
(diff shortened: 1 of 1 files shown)


htdocs/admin/stock.php
```diff
@@ -335,7 +343,7 @@
 print '<td class="right">';
 if (isModEnabled("supplier_order") || isModEnabled("supplier_invoice")) {
 	if ($conf->use_javascript_ajax) {
-		if ($disabled) {
+		if (in_array('BILL', $disableStockCalculateOn)) {
```
eldy, line 346: I think you want to do here ?
`if (in_array('SUPPLIER_BILL', $disableStockCalculateOn)) {`

htdocs/admin/stock.php
```diff
@@ -357,7 +365,7 @@
 print '<td class="right">';
 if (isModEnabled("supplier_order") || isModEnabled("supplier_invoice")) {
 	if ($conf->use_javascript_ajax) {
-		if ($disabled) {
+		if (in_array('VALIDATE_ORDER', $disableStockCalculateOn)) {
```
eldy, line 368: I think you want to do here ?
if (in_array('SUPPLIER_VALIDATE_ORDER', $disableStockCalculateOn)) {

lvessiller-opendsi: Yes, you're right. 
I have done it.

PR #33156, opened 2025-02-20
NEW Init externalModules Class to Fetch Modules from Multiple Sources

Daoud-mohamed:
This PR initializes the externalModules class to fetch modules from multiple sources and adapts it to the new Dolistore API

### DIFF
--- a/htdocs/admin/remotestore/class/externalModules.class.php
+++ b/htdocs/admin/remotestore/class/externalModules.class.php
@@ -67,8 +67,9 @@ class ExternalModules
 	 * @var string
 	 */
 	public $dolistore_api_key;
+
 	/**
-	 * @var array|null
+	 * @var array<int, mixed>|null
 	 */
 	public $products;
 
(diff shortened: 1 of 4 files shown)


htdocs/admin/remotestore/class/externalModules.class.php
```diff
@@ -0,0 +1,526 @@
...
+		// If status code is 200 or 201, return an empty string
+		if ($request['status_code'] === 200 || $request['status_code'] === 201) {
+			return '';
+		}
+
+		// Get the predefined error message or use a default one
+		$error_message = $error_messages[$request['status_code']] ?? 'Unexpected HTTP status: ' . $request['status_code'];
+
+		// Append error details if available
+		if (!empty($request['response']) && isset($request['response']['errors']) && is_array($request['response']['errors'])) {
```
mdeweerd, line 517: It is indicated that 'response' is a string.  So to fix that you need to suggest it is `response:?array{errors:string[]}` or `response:null|string|array{errors:string[]}` if it can also be a string.

Daoud-mohamed:
Thanks @mdeweerd !

PR #33158, opened 2025-02-20
Fix extrafield filter with $SEL$ ... IN ...  statement

### LINKED ISSUE #33153: Filter on extrafields not working correctly for IN
### Bug

When using something like `table:ref:rowid::season:IN:$SEL$ rowid FROM llx_season WHERE season_start => CURRENT_DATE AND AND season_end <= CURRENT_DATE` as filter for extrafields the code is. not working properly. I think it's line 1361 in extrafields.class.php

### Dolibarr Version

20.0.1

### Environment PHP

8.1

### Environment Database

_No response_

_No response_

### Attached files

_No response_

kubajznik:
When using something like table:ref:rowid::season:IN:$SEL$ rowid FROM llx_season WHERE season_start => CURRENT_DATE AND AND season_end <= CURRENT_DATE as filter for extrafields the code is. not working properly. I think it's line 1361 in extrafields.class.php

# FIX #33153

### DIFF
--- a/htdocs/core/class/extrafields.class.php
+++ b/htdocs/core/class/extrafields.class.php
@@ -1433,7 +1433,7 @@ class ExtraFields
 						//var_dump($InfoFieldList[4].' -> '.$pos); var_dump($tmpafter);
 						$InfoFieldList[4] = $tmpbefore;
 						if ($tmpafter !== '') {
-							$InfoFieldList = array_merge($InfoFieldList, explode(':', $tmpafter));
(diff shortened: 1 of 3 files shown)


eldy:
It is on purpose that using a SQL request at the place of the data for the list on value in the IN has been sanitized. But:
- Sanitizing was not complete, it was still possible to the subrequest SQL with some requests different than the one in your example
- For cases like yours the <= were lost.

Using your PR, I have fixed the 2 points differently into v21.

eldy:
Can you try with last version in develop branch if it is ok now  for your use case ?

PR #33172, opened 2025-02-21
Add original object for hook selectForFormsListWhere

BenjaminFlr:
When we want to use `selectForFormsListWhere` hook, the only way to get information about the original object, from a card for example, is using a global variable.

Maybe it could be interesting to get this object directly in `$parameters` array.


eldy:
The code called by an external module may also need a lot of other information. We can't provide on any dolibarr function all the object than an external modulemay need just to avoid this external module to get the value using a global $parametertoget.
This will overload the code of dolibarr for a need that is not a dolibarr need but a need of an external module. We must do it when there is no solution, but here the use of global is possible and even if it is "cleaner" if we take into consideration both dolibarr+modules, in a Dolibarr only context it adds a technical debt. So i prefer to discard it for the moment.

BenjaminFlr:
Thanks for your answer @eldy . But in fact, I'm not sure to understand what can be done with this hook if we don't get anything but the object from which this function is called. For sure I've got the list of objects, but I can't filter anything without more information related to the current page.

Of course I can use a "global". But I've got to check if we only get the context.

PR #33175, opened 2025-02-21
Fix formai class to make ai prompt code to be more general

Hystepik:
+ Move of ajax function in a specific php function to have a generic function
+ set timeout of display ai prompt call to 30 seconds


htdocs/core/class/html.formai.class.php
```diff
@@ -131,112 +134,130 @@ public function getSectionForAIPrompt($function = 'textgeneration', $format = ''
...
-								if (timeoutfinished) {
-									$('#ai_status_message').hide();
-								}
-							},
-							error: function(xhr, status, error) {
-								alert(error);
-								console.error('error ajax', status, error);
-								$('#ai_status_message').hide();
+						$('#ai_status_message".$htmlContent."').hide();
+					}, 30000);
```
eldy, line 146: In a next PR, i recommand to move this value 30000 into a parameter "Time out delay for AI response" in the first tab of setup of AI.

PR #33183, opened 2025-02-21
Qual: Fix phan notices (user)

mdeweerd:
# Qual: Fix phan notices (user)


htdocs/user/class/user.class.php
```diff
@@ -1182,7 +1182,7 @@ public function delrights($rid, $allmodule = '', $allperms = '', $entity = 0, $n
 			// les charactis for the module, permissions and sub-permission of this permission.
 			$sql = "SELECT module, perms, subperms";
 			$sql .= " FROM ".$this->db->prefix()."rights_def";
-			$sql .= " WHERE id = '".$this->db->escape($rid)."'";
+			$sql .= " WHERE id = '".$this->db->escape((string) $rid)."'";
```
eldy, line 1185: As this is an id, you can replace with 
$sql .= " WHERE id = ".((int) $rid);
so no cast is required by database engine

mdeweerd: @eldy For fixing notices, adding `(string) ` to the existing code is faster and safer (and avoids deep analysis):
- Faster because there is no need to remove characters, and I can just repeat my last insert with a "." (in vim);
- Faster also because I had some case where this resulted in more notices (I dont't remember the reason) and I ended up using a simple string cast;
- Safer because there are cases where the variable is incorrectly typed as an int and therefore casting it to an int would result in a bug.

The main goal is to fix the "false" positives so that the rule is no longer ignored for the file (but not "fixing" real positives by simply "hiding" notices).

Casting int to string does not harm in these cases.

I'll fix this one differently.

BTW, this not what is causing the PHPUnit CodingTest failure.

mdeweerd:
@eldy

I updated the CodingTest case so that we get more information about where this occurs in the logs, and it should also be picked up by logToCs so that it is annotated in the `Files changed` tab.
I can make a separate PR for the test update, but it's useful to show the effect here.

I believe this is the case that gave me troubhe in the past, "NOT IN" does not accept int casts.
Which requires to also update the NOT IN query, etc with high risk of introducing bugs.

```plaintext
htdocs/user/class/user.class.php:2177: NotSanitizedString in IN/NOT IN sql query `((int) $this->entity).", 0) AND login = '".$this->db->escape($this->login)."'";...`)
htdocs/user/class/user.class.php:2191: NotSanitizedString in IN/NOT IN sql query `((int) $this->entity).", 0) AND email = '".$this->db->escape($this->email)."'";...`)
```

Do want to update the CodingTest, or not.  As-is it seems useful to remind that the "IN/NOT IN" test is not of any use meaning that there is a copy/paste error and that the lilst to test on was not properly introduced.

Edit: win-ci is disabled, so the impact on the annotations can not be checked in this PR.

PR #33187, opened 2025-02-22
Fix ci - customreports, list_print_total

mdeweerd:
# Fix ci - customreports, list_print_total


mdeweerd:
@eldy This should fix the ci.

Issue #33190, opened 2025-02-23
"Pending" amount not updating after "convert excess received into available credit"


AurelienBISOTTI:
### Bug

When a customer create an excess paiement, the Dolibarr user can create an available credit.

The invoice is then considered as paid.

There should not be "pending" amount in the invoices list.

Bug : The "pending" amount is still shown as if there was not available credit created.

### Dolibarr Version

18, 19, 20

### Environment PHP

Any

### Environment Database

Any

### Steps to reproduce the behavior and expected behavior

Do an excess payment on an invoice

Convert into available credit

Check "pending" amount on invoices list

![Image](https://github.com/user-attachments/assets/c7204007-d762-4bc6-89ef-c212e30dc13f)

![Image](https://github.com/user-attachments/assets/08bcfdd8-3657-42f3-a188-76963d0504a5)

![Image](https://github.com/user-attachments/assets/f3a962e1-ab03-4cb4-adad-9fc0d55287e8)

![Image](https://github.com/user-attachments/assets/bff7dee6-6a2c-43b0-af9e-667cfdb2d6d7)

hregis:
@AurelienBISOTTI this is with situation invoices?

AurelienBISOTTI:
With classic invoices

hregis:
@AurelienBISOTTI i send a fix, thank you : https://github.com/Dolibarr/dolibarr/pull/33191

rycks:
@hregis i dont have the time to try to fix such bug, please don't assign bugs to people like that !

rycks:
@AurelienBISOTTI there is a lot of messages on the PR linked to that bug, could you please participate here : https://github.com/Dolibarr/dolibarr/pull/33191 that's old but that's a real problem and we could find a solution i'm sure !

eldy:
Closed as duplicate of #33191

PR #33191, opened 2025-02-23
FIX remain to pay is 0 if invoice has paid status

hregis:
@eldy same fix for 19, 20, 21, develop


htdocs/compta/facture/list.php
```diff
@@ -2400,6 +2400,9 @@
 				// Pending amount
 				if (!empty($arrayfields['rtp']['checked'])) {
 					print '<td class="right nowraponall amount">';
+					if (!empty($obj->paye)) {
+						$remaintopay = 0;
```
eldy, line 2404: If remain to pay is 2 euros and we decide to set invoice to status Paid, why hiding the information of remain to pay is 2 ?
We can keep the information of balance on screen because we must know that we set to pay even if balance was not null.
What do you think of keeping the 2 but striked (barré) so we understand balance was not zero and it is no more exoected to be paid.

hregis: @eldy same sentence in "/compta/facture/card.php" in v18 line 4099 ?

```
if ($object->paye) {
	$resteapayer = 0;
}
```

hregis: @eldy the balance to pay may be less or more! when you create a credit for an excess payment the invoice is declared paid, we should not see the negative balance if a credit has been created no? Do you have a solution to manage all the cases?

rycks:
Hello @hregis what is the use case where you find that as a problem ?

There is a lot of different situations so i can't say 'ok' or 'not ok' for such request and i see different situations into dolibarr code (sometimes we "hide" that information, for example on pdf files where 

```
	if ($object->paye) {
		$resteapayer = 0;
	}
	$resteapayeraffiche = $resteapayer;
```

but some other part of code that information is used ... that's not clear ...

htdocs/compta/facture/list.php
eldy: > @eldy the balance to pay may be less or more! when you create a credit for an excess payment the invoice is declared paid, we should not see the negative balance if a credit has been created no? Do you have a solution to manage all the cases?

Yes, a good evolution is to not force to 0 the value but showbthe value striked.
I think having on screen a remain to pay = 2 with the 2 striked, we understand that it is not really 2 but 0.

hregis:
@eldy @rycks i change my PR, it's ok for you ?

Capture d’écran 2025-02-28 à 08 41 02

rycks:
> @eldy @rycks i change my PR, it's ok for you ?

hmmmm that's sort of use case ... i just imagine there is 1 or 2€ remains to pay and then you don't want to get that invoice into "unpaid" list and my accounter says "ok to push 0€ as remain to pay" that's reallly a common situation at the end of the fiscal year (errors on payment or such situations)

BUT your screenshot is really different, in such situation with > 100€ (it depends on annual business) as remain to pay i think we can't put "paid" invoice, we have to make an credit note and use it to put the invoice "paid"

hregis:
@rycks je te laisse gérer le ticket hein ? bonne journée https://github.com/Dolibarr/dolibarr/issues/33190

rycks:
@hregis why did you close ? we are talking about a problem don't close it, please !

i cant' see the amount, is "990" striked  or "-990" ?

the link to #33190 is new for me, excuse-me but that's the first time you make a link or an information that this PR will fix that bug !!!!

hregis:
@eldy @rycks Basically it's when we convert the overpayments and the invoice is changed to paid. Can I filter just on negative amounts?

hregis:
> @hregis why did you close ? we are talking about a problem don't close it, please !
> 
> i cant' see the amount, is "990" striked or "-990" ?
> 
> the link to #33190 is new for me, excuse-me but that's the first time you make a link or an information that this PR will fix that bug !!!!

this is a negative amount -990

rycks:
> @eldy @rycks Basically it's when we convert the overpayments and the invoice is changed to paid. Can I filter just on negative amounts?

ok then the bug is not on list of invoices but on overpayment ...

then when a customer makes me 990€ of overpayment i do a bank transfert of 990 and on dolibarr i with to put that information on the invoice

for example here is what i do:

image

then

image

...

that is what "i" do, but i read #33190 bug and i will take a moment to have a look at it please wait :)

rycks:
ok i understand, the bug is on clic on "Convertir le trop-perçu en crédit disponible" button

my suggestion is to add a line into llx_paiement_facture or something like that to get the same result as i have with my "remboursement via virement pour trop percu" is that a good idea ?

hregis:
@rycks yes but here it's when you use this button for convert the overpayment

Capture d’écran 2025-02-28 à 09 47 15

Capture d’écran 2025-02-28 à 09 49 04

rycks:
then the result could become

image

rycks:
And the benefit was on PDF TOO : we have the information displayed (with current bug on convert to credit the pdf remains with bad informations)

image

hregis:
you can not create a negative payment and after use this amount in deduction of other invoice !!?

rycks:
with the bug PDF will stays displaying that:

image

i think we have to go deeper on what "convert to credit" button do and add a new line into payment history, but we have to double check all consequences ...

rycks:
so i'm sorry but that's not a trivial bug for me we have to take time and i'm sorry i don't have more time now to do it :-(

hregis:
me too ! 😄

eldy:
> me too ! 😄 

I assign the ticket to myself, I will have a look...

hregis:
> > me too ! 😄
> 
> I assign the ticket to myself, I will have a look...

thank you grand master Yoda ! 😄

rycks:
so @eldy what could we do for that PR ?

eldy:
This ticket seems the same problem than the one described here #33747 
So i cut/paste the anser:  
On recent version, the remain to pay is calculated with
```
$totalpaid = $object->getSommePaiement();
$totaldeposits = $object->getSumDepositsUsed();
$resteapayer = price2num($object->total_ttc - $totalpaid - $totalcreditnotes - $totaldeposits, 'MT');
```
And the part of amount of excess received seems already included into the getSumCreditNotesUsed(). So don't need to use the getAvailableDiscounts(). So i surely miss something, so the most important is to be able to find a use case to reproduce the bug. I failed to find one. So if you can describe a proces to reproduct it we may progress on this.

JamBalaya56562:
Here is a reproducible case (the one that was missing), verified on a fresh **21.0.4** install, plus a measurement of every layer so we can see exactly where the inconsistency is.

### Steps to reproduce
1. Create and validate a standard customer invoice, total 500.
2. Register a payment of 1000 on it (overpayment).
3. On the invoice card, click **"Convert excess received to discount"** (`action=converttoreduc`).

Resulting DB state (confirmed):
- `llx_facture`: `paye=1`, `fk_statut=2` (paid/closed), `close_code=NULL`.
- `llx_societe_remise_except`: one row, `(EXCESS RECEIVED)`, `amount_ttc=500`, `fk_facture_source=<invoice>`, `fk_facture=NULL`.
- `llx_paiement_facture`: still 1000 — **the conversion does not touch the payment**, so `getSommePaiement()` keeps returning 1000.

So the raw balance is `total_ttc - getSommePaiement() - getSumCreditNotesUsed() - getSumDepositsUsed() = 500 - 1000 - 0 - 0 = -500`. (`getSumCreditNotesUsed()` returns 0 here because it counts the discounts *consumed by* this invoice, not the one *sourced from* it.)

### What each layer shows for this paid invoice
| Layer | remain-to-pay shown | masks `paye`? |
|---|---|---|
| Invoice **card** (`compta/facture/card.php`) | **0.00** | yes — `if ($object->paye) { $resteapayer = 0; }` (card.php ~4343) |
| **PDF** (`pdf_sponge`) | **0.00** | yes — `if (!empty($object->paye)) { $resteapayer = 0; }` (pdf_sponge ~2091; present on 18.0→develop) |
| Invoice **list** (`compta/facture/list.php`) | **-500.00** (and the column total becomes -500.00) | **no** |

So the card and the PDF already zero the balance when the invoice is `paye`; **`list.php` is the only place that still shows the raw -500** for a standard invoice. It only has that mask for credit notes (`type == TYPE_CREDIT_NOTE && paye == 1`), not for standard invoices converted via the excess-received button.

### With this PR applied
I applied this PR's change locally on the same data: the list cell then shows `-500.00` **struck through / greyed** (`colorgrey strikefordisabled`) and the column **total becomes 0.00**. That matches @eldy's request to keep the value visible-but-struck rather than hiding it.

<details>
<summary>Screenshots (before / after this PR, and the card for the same invoice)</summary>

**Invoice list, current code** — the invoice is `Paid` but "Remaining unpaid" shows `-500.00` and the column total is `-500.00`:

image

**Invoice list, with this PR** — value is struck through / greyed and the column total is `0.00`:

image

**Invoice card, same invoice, current code** — "Remaining unpaid" is already `0.00` here:

image

</details>

### Note on @rycks's concern
The PDF does **not** stay wrong: `pdf_sponge` already applies the same `if (!empty($object->paye)) $resteapayer = 0;` (verified on 18.0, 19.0, 21.0, 22.0, 23.0 and develop), so it prints 0. So aligning `list.php` with the rule already used by both the card and the PDF is consistent with the rest of the code — a deeper change to what the convert button writes into `paiement_facture` is not required just to make the list correct.

(Reproduced on a local Docker install; I can provide the seed SQL / more detail if useful.)

PR #33201, opened 2025-02-24
Fix selectcontacts where GETPOST('socid') is empty

uvaldenaire-opendsi:
When GETPOST('socid') is an empty string, all the contacts were loaded in $form->selectcontacts() even if MAIN_ACTIONCOM_CAN_ADD_ANY_CONTACT is not set to avoid to load all contacts.


eldy:
And if we want all the contact by providing socid=0 or socid='' ?  (This may be the expected behaviour on some pages)

PR #33211, opened 2025-02-24
Qual: Fix phan notices in functions.lib.php

mdeweerd:
# Qual: Fix phan notices in functions.lib.php

Fixes the phan notices in functions.lib.php, but I did not fix those where an amount is passed into a field that does not expect this for Product::getLibStatut.


mdeweerd:
@eldy I am releasing this with the two remaining phan notices - I think it's better to fix them to detect future noticies of these types on functions.lib.php

mdeweerd:
@eldy The last notices are now fixed with some type hinting.

PR #33222, opened 2025-02-25
feat(contract-line): created new endpoint paginated with limit and page

atm-irvine:
# NEW|New endpoint paginated with limit and page
I've been added a new endpoint paginated for contract lines


htdocs/contrat/class/api_contracts.class.php
```diff
@@ -262,6 +262,52 @@ public function getLines($id)
...
+	 *
+	 * @param $id Id of contract
+	 * @param $limit Limit of lines to return
+	 * @param $page Page number
+	 *
+	 * @url GET {id}/lines/paginated
+	 *
+	 * @return array
+	 */
+	public function getLinesPaginated($id, $limit = 10, $page = 0)
```
eldy, line 277: In doc coment, you must add the type of var, for example:
@param int    $id     Id of contract

atm-irvine: Corrected

eldy:
Instead of a adding a new url endpoint, can you instead follow the same solution used for other api than need pagination:
Add the field 
	 * @param int			   $limit				Limit for list
	 * @param int			   $page				Page number
	 * @param bool             $pagination_data     If this parameter is set to true the response will include pagination data. Default value is false. Page starts from 0*

Return old format when pagination_data is not set and return with pagination information when set.

See example into api that Get a list of orders. We should use the same way for api to get lines.

atm-irvine:
@eldy Is it better now ?

atm-maxime:
@eldy can you give us feedback on this one too ?

PR #33223, opened 2025-02-25
NEW: Add management of date begin/end and discount for a customer price (SQL)

kkhelifa-opendsi:
Add management of date begin/end and discount for a customer price (SQL)


htdocs/install/mysql/tables/llx_product_customer_price.sql
```diff
@@ -41,6 +43,7 @@ create table llx_product_customer_price
   localtax1_type        varchar(10)  NOT NULL DEFAULT '0',
   localtax2_tx			double(7,4)  DEFAULT 0,         -- Other local VAT 2
   localtax2_type        varchar(10)  NOT NULL DEFAULT '0',
+  remise_percent		real DEFAULT 0,
```
eldy, line 46: Can you use name discount_percent instead of remise_percent ?

Renaming remise_percent into discount_percent into other table is not yet planned, but we should have to do it one day...

PR #33227, opened 2025-02-25
Fix Situation Percent with USE_INVOICE_SITUATION = 2

rycks:
Backport of #33174 thanks to @EchoLoGeek fix but for 20.0 branch ... is that acceptable due to the fact that situation invoices is experimental feature ? i hope because without that fix "new" situations invoices in 20 is unusable


eldy:
Yes, as it is experimental feature, it is code that is not "officially" available to end user (user use it as its own risk). So if change is "inside" code commented by a 
if (experimentaloption is on)
it can be included into released version (it does not mean it will always, may depend on rules not clear, depends on good sense only).

Here it seems it is correctly into a test if option on, so merged.

PR #33238, opened 2025-02-26
fix(facture): removed useless fetch product

atm-irvine:
# FIX|Fix removed useless fetch product
I remove this piece of code because it fetches a product for nothing and didn't handle after


eldy:
As we are in a frozen version, changefor cosmetic or code quality is not allowed. 
Only change to fix visible bugs with the lower printing is allowed for bug fix maintenance
Can you push in develop ?

atm-irvine:
@eldy You can find here the PR to develop : https://github.com/Dolibarr/dolibarr/pull/33253

PR #33241, opened 2025-02-26
fix: incorrect string value when emoji in body

atm-irvine:
# FIX|Fix incorrect string value when emoji in body
I add a SET NAMES request set to utf8mb4 because when a message has emoji, it throws a sql error incorrect string value


hregis:
@atm-irvine it's not compatible with all mysql/mariadb and with postgresql

PR #33252, opened 2025-02-27
feat(invoice-rec): new rule_for_lines_dates field - Impl

atm-irvine:
# NEW|New rule_for_lines_dates field - Impl
This is the implementation of the DB PR : https://github.com/Dolibarr/dolibarr/pull/33237


htdocs/compta/facture/card-rec.php
```diff
@@ -1159,6 +1164,11 @@ function showselectfksocieterib(){
 			print "</td></tr>";
 		}
 
+		// Rule for lines dates
+		print "<tr><td>".$langs->trans("RuleForLinesDates")."</td><td>";
+		print $form->getSelectRuleForLinesDates(GETPOSTISSET('rule_for_lines_dates') ? GETPOST('rule_for_lines_dates', 'alpha') : $object->rule_for_lines_dates);
```
eldy, line 1169: The property 
rule_for_lines_dates
must be declared among the list of properties into the class Facture.

htdocs/core/class/html.form.class.php
```diff
@@ -4753,6 +4774,46 @@ public function getSelectConditionsPaiements($selected = 0, $htmlname = 'condid'
 	}
 
 
+	/**
+	 * Returns select with rule for lines dates
+	 *
+	 * @param $selected Selected value
```
eldy, line 4780: Bad declaration of parameter.Must be
@param   type    $varname    Descritpion

htdocs/compta/facture/card-rec.php
atm-irvine: I don't understand why I have to declare this field in the Facture class, as it's a field in the FactureRec class.

eldy: So there is a bug introduced  here because according to phan the $object here seems to be a Facture and not FactureRec.

atm-irvine: Yes I see.. How should we do it ?

eldy: I depends of what you plan to do.
May be line 
$object = new Facture($db); // Source invoice
we should use another name for variable than $object (it erase here the $object of the recurring invoice).

PR #33257, opened 2025-02-27
New/group fetch and fetch optionals

atm-irvine:
# NEW|New  group classic fetch and fetch optionals depending on conf
I've grouped classic fetch and fetch optionals to prevent doing multiple sql requests for the same result. For invoice and its lines. All of this depends on MAIN_DO_FETCH_IN_ONE_SQL_REQUEST conf.
Once this PR is merged, I'll do it for every document.


hregis:
@eldy @atm-irvine the sql function "ST_AsWKT()" is not compatible with Postgresql no ?  
the use of this sql function is also in "fetch_optionals()" function of commonobject.class.php" !

atm-irvine:
@hregis This function is available if you activate an option on PostgreSQL, apparently. Since it was already in the core, I took the liberty of taking it.

hregis:
@atm-irvine are you talking about PostGIS?

https://postgis.net/

If yes the name of the function is not the same :

https://postgis.net/docs/ST_AsEWKT.html

`ST_AsEWKT` Instead of `ST_AsWKT`

Maybe the better is to use ST_AsText

https://postgis.net/docs/ST_AsText.html

`ST_AsWKT` is an alias of `ST_AsText` with Mysql/MariaDB

And what happens if PostGIS can not be installed?

Thanks

atm-irvine:
@eldy What do you think about that ?

atm-irvine:
@hregis Tbh I don't know. I just took the code.. Eldy will tell us what to do

hregis:
@eldy it's possible to check if PostGIS is enabled with "PostGIS_Version()"
https://postgis.net/docs/PostGIS_Version.html

maybe add a specific function in sql drivers for this to avoid errors ?

atm-maxime:
Hi @eldy , can you give us feedback ?

eldy:
The ST_AsWKT was introduced in core and is used as experimental only when we are using extafields 'point', 'multipont', ...
Goal was to experiment compatiility issue. So in this PR, we can keep it to match the same function already introduced.

What we want is a full text representation. ST_AsWKT that means "Well Known Text" so a text.

ST_AsWKT ans ST_asText are alias in both pgsql and mysql.  ST_AsWKT was chosen because the term "WKT" is the technical  admited term in GEO manipulation. asText would be good too because we understand it is a "full text" output. This is not really important as they are both alias.
In a future, we can add an abstraction method into the db driver ->geoAsText() that will use the correct function but, because the ST_AsWKT was already introdcued, this should be in a different PR.

Same for testing if GIS features are available. We should add this test to be allowed to create a geo field only if the geoip feature is available.

For the moment all geoip are not yet stabilized, and such enhancement should be done before feature become official. For the moment, it is an experimental feature that need the hidden option MAIN_USE_GEOPHP to be on. But this should be part of other different PRs.

htdocs/compta/facture/class/facture.class.php
```diff
@@ -2436,7 +2574,47 @@ public function fetch_lines($only_product = 0, $loadalsotranslation = 0)
...
+							} else {
+								$this->array_options['options_' . $key] = $rawval;
+							}
+						}
+					}
+
+					// Champs "computed"
+					foreach ($extrafields->attributes[$this->table_element_line]['label'] as $key => $val) {
+						if (!empty($extrafields->attributes[$this->table_element_line]['computed'][$key])) {
+							if (empty($conf->disable_compute)) {
```
eldy, line 2605: @atm-irvine I don't understand this test.
Is this related to fetch change or part of another feature ?

atm-irvine: @eldy I have backported this code from "fetch_optionals" method. I didn't change anything.

PR #33265, opened 2025-02-27
Fix Excel Export of zero values

omogenot:
# FIX Excel Export for zero values

When exporting a using Excel file format, if an exported value is zero, the corresponding cell in the exported excel file will be empty. This creates a problem when exporting flag type of value (ex. customer type 0 or 1) and trying to import the resulting Excel file back which will emit the 'no default value for field xyz' error message and force the user to edit the Excel file to set these empty cells with a value of 0.

### DIFF
--- a/htdocs/core/modules/export/export_excel2007.modules.php
+++ b/htdocs/core/modules/export/export_excel2007.modules.php
@@ -336,9 +336,9 @@ class ExportExcel2007 extends ModeleExports
 				continue;
 			}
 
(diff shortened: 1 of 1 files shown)


eldy:
Did you check that for fields that are text, if content is empty, we continue to have empty space and not 0 ?

omogenot:
> Did you check that for fields that are text, if content is empty, we continue to have empty space and not 0 ?

I added a `is_numeric` condition so that the value is set to '' if it is empty and not numeric (i.e. not 0)

PR #33277, opened 2025-02-28
FIX missing check constant PROJECT_CAN_ALWAYS_LINK_TO_ALL_SUPPLIERS

hregis:


htdocs/fourn/facture/card-rec.php
```diff
@@ -1158,7 +1158,7 @@
 					$morehtmlref .= '<input type="submit" class="button valignmiddle" value="' . $langs->trans("Modify") . '">';
 					$morehtmlref .= '</form>';
 				} else {
-					$morehtmlref .= $form->form_project($_SERVER['PHP_SELF'] . '?id=' . $object->id, $object->socid, $object->fk_project, 'none', 0, 0, 0, 1, '', 'maxwidth300');
+					$morehtmlref .= $form->form_project($_SERVER['PHP_SELF'] . '?id=' . $object->id, (empty($conf->global->PROJECT_CAN_ALWAYS_LINK_TO_ALL_SUPPLIERS) ? $object->socid : -1), $object->fk_project, 'none', 0, 0, 0, 1, '', 'maxwidth300');
```
eldy, line 1161: Can you use 
(getDolGlobalString('PROJECT_...
Instead of
If (empty(...
v18 accepts the 2 ways but higher version requires thebsecond one to it will avoid trouble when change will be merged upstream.

hregis: @eldy in v18 all the call of "PROJECT_CAN_ALWAYS_LINK_TO_ALL_SUPPLIERS" are with $conf ! 
i change all with getDolGlobalString() ?!

eldy: Only new introduced lines, old one should be updated by higher version...

rycks:
ok @eldy job is done there is getDolGlobalString everywhere for that PR, could you please have a look at that PR again ?

rycks (approved):
seems good for me

PR #33288, opened 2025-03-01
rename llx_societe.nom into llx_societe.name

frederic34:


hregis:
@eldy @frederic34 
à chaque changement important il faut faire un appel important ! 
sinon ça risque de créer des problèmes importants !

eldy:
Such change will need more steps to be achieve.
For exemple replacing
```
$sql = "SELECT code_client, fk_forme_juridique, nom, address, zip, town, fk_pays, phone, siret FROM ".MAIN_DB_PREFIX."societe";
...$obj->nom...
```
into
```
$sql = "SELECT code_client, fk_forme_juridique, name, address, zip, town, fk_pays, phone, siret FROM ".MAIN_DB_PREFIX."societe";
...$obj->name...
```

Should have first a step
```
$sql = "SELECT code_client, fk_forme_juridique, nom as name, address, zip, town, fk_pays, phone, siret FROM ".MAIN_DB_PREFIX."societe";
...$obj->name...
```

So we can change the database only once the code (internal and external is ready with the new name).
A one step massive change will generate too much troubles.

PR #33299, opened 2025-03-03
Fix dolGetButtonAction Tooltip

thersane-john:
# Fix 
Small Fix for Button Tooltips

When I coded the dolGetButtonAction function, I used dol_escape_htmltag instead of dol_htmlentities.
As a result, the tooltip does not support HTML formatting, preventing the use of data-* attributes containing HTML especially for JavaScript interactions.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -12344,6 +12344,7 @@ function dolGetStatus($statusLabel = '', $statusLabelShort = '', $html = '', $st
  *                                      'xxxxx' => '', // your xxxxx attribute you want
  *                                      'class' => 'reposition', // to add more css class to the button class attribute
  *                                      'classOverride' => '' // to replace class attribute of the button
+ *                                      'use_unsecured_unescapedattr' => [] | bool // on true will use unsecured fonction to escape all attributes, if use array it will use unsecured fonction to escape attribute key present in this array
  *                                      ],
  *                                      'confirm' => [
  *                                      'url' => 'http://', // Override Url to go when user click on action btn, if empty default url is $url.?confirm=yes, for no js compatibility use $url for fallback confirm.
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -12520,7 +12520,7 @@ function dolGetButtonAction($label, $text = '', $actionType = 'default', $url =
 	}
 
 	// escape all attribute
-	$attr = array_map('dol_escape_htmltag', $attr);
+	$attr = array_map('dol_htmlentities', $attr);
```
eldy, line 12523: And if you use the function   dolPrintHtmlForAttribute  (designed to allow html but more secured than dol_htmlentities).
Does it works as expected ?

thersane-john: @eldy : I have tested and it fail

thersane-john: @eldy : dolPrintHtmlForAttribute disable the following tags:
hr, strong, div
but the b tag works and so do the p tags with class

## dol_htmlentities :
Capture d’écran du 2025-03-04 08-26-15


## dolPrintHtmlForAttribute : 
Capture d’écran du 2025-03-04 08-26-51

thersane-john: > If not, can you describe the use case to reproduce (we really need to replace dol_htmlentities by dolPrintHtmlForAttribute that was too easily bypassed for js and html injection by yogosha searcher)

Okay, but here if a hacker has access to $attr he's probably got access to $text or $label vars and there are no restrictions for that one, so what do we do?
It's a display function, a deveoppeur must be able to put in the data we want, and data verification must be done beforehand.
if not, we need to factor and make 2 functions:
dolGetButtonActionRaw: a new function copying the current dolGetButtonAction function but with dol_htmlentities

and dolGetButtonAction (the current function), which uses dolGetButtonActionRaw after cleansing the data sent.

the problem I often see with functions that are too restrictive is that module devs won't actually use them, which is generally worse...
I'd rather have 2 functions, one of which is raw but contains warnings and examples of “good practice” in the description, and which in particular encourages use of the other function.

eldy: > > If not, can you describe the use case to reproduce (we really need to replace dol_htmlentities by dolPrintHtmlForAttribute that was too easily bypassed for js and html injection by yogosha searcher)
> 
> Okay, but here if a hacker has access to $attr he's probably got access to $text or $label vars and there are no restrictions for that one, so what do we do? It's a display function, a deveoppeur must be able to put in the data we want, and data verification must be done beforehand. if not, we need to factor and make 2 functions: dolGetButtonActionRaw: a new function copying the current dolGetButtonAction function but with dol_htmlentities
> 

This is why we have  dolPintHTML() function for text and label.

> It's a display function, a deveoppeur must be able to put in the data we want, and data verification must be done beforehand.

Relying on sanitizing by each caller is not reliable and never secured. We should never trust incoming external data. 

> the problem I often see with functions that are too restrictive is that module devs won't actually use them, 
which is generally worse...

Sure, but here this should not be the case, the developer is still able to provide the string for attribute he want. Which king of value do you fail to send as attribute when using dolPrintHtmlForAttribute ? Do you have an example ?

thersane-john: I've made a few changes

PR #33302, opened 2025-03-03
Add missing hook to print common object lines

atm-etienne:
# FIX - Add missing hook to replace default lines list on Common Object card

On common object cards, there is a list of object lines (invoice lines, for example), which can be modified with a hook for the title line of the list, or a hook executed on each line.

However, there is no hook to completely replace this list.
Some objects have several hundred or even thousands of lines, which prevents the object's card from loading. We therefore need to be able to replace this list to enable the card to be displayed, and custom the list.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -4951,57 +4951,61 @@ abstract class CommonObject
 		}
 		$extrafields->fetch_name_optionals_label($this->table_element_line);
 
-		$parameters = array('num'=>$num, 'dateSelector'=>$dateSelector, 'seller'=>$seller, 'buyer'=>$buyer, 'selected'=>$selected, 'table_element_line'=>$this->table_element_line);
-		$reshook = $hookmanager->executeHooks('printObjectLineTitle', $parameters, $this, $action); // Note that $action and $object may have been modified by some hooks
+		$parameters = array();
+		$reshook = $hookmanager->executeHooks('printObjectLinesList', $parameters, $this, $action);
 		if (empty($reshook)) {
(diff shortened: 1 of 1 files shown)


eldy:
Such hook "printObjectLinesList" was never introduced in v17 (can't find in v21 either)
If this is a new need, thanks to push in develop.

atm-etienne:
@eldy I see you changed the base branche from 17.0 to develop, then you revert it. 
What should I do so?

eldy:
> @eldy I see you changed the base branche from 17.0 to develop, then you revert it. 
> What should I do so?

Repush in develop if not already present (changing myself generates too conflict).
Or close the pr if already in develop.

htdocs/core/class/commonobject.class.php
```diff
@@ -4951,57 +4951,61 @@ public function printObjectLines($action, $seller, $buyer, $selected = 0, $dateS
 		}
 		$extrafields->fetch_name_optionals_label($this->table_element_line);
 
-		$parameters = array('num'=>$num, 'dateSelector'=>$dateSelector, 'seller'=>$seller, 'buyer'=>$buyer, 'selected'=>$selected, 'table_element_line'=>$this->table_element_line);
-		$reshook = $hookmanager->executeHooks('printObjectLineTitle', $parameters, $this, $action); // Note that $action and $object may have been modified by some hooks
+		$parameters = array();
```
altairis-tof, line 4954: hello, maybe you can pass some parameters to the hook (seller, buyer...) ?

Vaadasch:
PR pulled anew in  #33641 ? 
If confirmed, can this one be closed ?

eldy:
Yes, replaced by the new PR

PR #33308, opened 2025-03-03
add status class

iouston:
# NEW
add a status class on invoice list


eldy:
Can you provide a screenshot of the change ?

PR #33311, opened 2025-03-03
Qual: Fix phan notices in api classes

mdeweerd:
# Qual: Fix phan notices in api classes

Fixes most notices in API classes - the remaining notices require more analysis.


mdeweerd:
With this PR all files have been reviewed for notices visible in cti.dolibarr.org .
That means that they need better insight in Dolibarr than I am targetting.

The PhanTypeMismatchArgument and PhanTypeMismatchArgumentNull notices still have about 1200 and 300 cases to review in about 400 files.  However files from htdocs/hrm onwards (in the alphabet) are already reviewed for these (33 PhanTypeMismatchArgument\* notices remaining).

PR #33315, opened 2025-03-04
fix special case INVOICE_USE_SITUATION type 2 and get_prev_progress call

rycks:
Next step as #33227 for situation invoices type 2 : each call to get_prev_progress must be only for situation invoices type 1

note: code could be factorized with test into get_prev_progress function but as all of other parts of dolibarr makes tests before function call i do the same


rycks:
@EchoLoGeek please have a look to that PR

rycks:
next step is to fix that on sponge pdf : 

image


in that case it's final situation invoice so 100% is required not 25%

htdocs/compta/facture/card.php
```diff
@@ -2668,7 +2668,11 @@
 
 		$line = new FactureLigne($db);
 		$line->fetch(GETPOSTINT('lineid'));
-		$percent = $line->get_prev_progress($object->id);
+		if (getDolGlobalInt('INVOICE_USE_SITUATION') == 2) {
+			$percent = 0; // New situation percent must be 0 (No cumulative)
+		} else {
```
EchoLoGeek, line 2673: Percent is calculated after ($addprogress on line 2764)
The problem is on the updateline() method.

```
// Line 4227
if (!isset($situation_percent) || $situation_percent > 100 || (string) $situation_percent == '' || $situation_percent == null) {
	$situation_percent = 100;
}
```

The fix must be the same as the addline method
```
if (!isset($situation_percent) || $situation_percent > 100 || (string) $situation_percent == '' || $situation_percent == null) {
	// INVOICE_USE_SITUATION = 2 - Lines situation percent on new lines must be 0 (No cumulative)
	if ($this->isSituationInvoice() && getDolGlobalInt('INVOICE_USE_SITUATION') == 2) {
		$situation_percent = 0;
	} else {
		$situation_percent = 100;
	}
}
```

EchoLoGeek (changes requested):
The code on the card.php file is not operational. It is the updateline method that must be adapted.

The rest seems ok, we don't need to retrieve the old progress because we only need to take into account the current line

EchoLoGeek:
I can't make a pull request on your branch, so I made a PR #33319 for the updateline method.
I think you can remove your changes from compta/facture/card.php and keep the rest of the files on your PR for compta and PDF files.
Thanks

rycks:
> I can't make a pull request on your branch, so I made a PR #33319 for the updateline method. I think you can remove your changes from compta/facture/card.php and keep the rest of the files on your PR for compta and PDF files. Thanks

yes, thanks ! 

as @eldy merged your fix, i just do a merge agains 20.0 to my branch, i think it's enough no ?

htdocs/compta/facture/card.php
rycks: but as i merge your fix i don't understand why your code is not "visible here" ...

rycks: ok i understand, you put a comment on card.php and the fix is on the other file .class.php ...

rycks:
note: i'm so sorry i just realize i made my branch from develop not from 20.0 !!!!

rycks:
Close this one and switch to https://github.com/Dolibarr/dolibarr/pull/33321, so sorry for that garbage shame on me !

PR #33321, opened 2025-03-04
FIX situation invoices progress + global percent sponge

rycks:
Here is a fix for dolibarr 20.0 and will replace https://github.com/Dolibarr/dolibarr/pull/33315

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -2668,7 +2668,11 @@ if (empty($reshook)) {
 
 		$line = new FactureLigne($db);
 		$line->fetch(GETPOSTINT('lineid'));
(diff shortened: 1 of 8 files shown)


rycks:
Then @EchoLoGeek could you please check that PR one more time :pray: ? thanks !

EchoLoGeek:
I think it's all right but I think there is one last little problem on Octopus model. 
Amount is good compared to the invoices I made, but remain to pay should be positive:
image

Is it normal for the amount to be reversed?

rycks:
that's not the only think to correct on octopus, i will do it tomorrow (i hope)

eldy:
The PR and way to fix seemsdeprecated, it has been fixed differently in upstream, so i close this PR.

PR #33323, opened 2025-03-04
New Ai translate feature in mail form

Hystepik:
Add a new AI feature to translate message field in mail form
image


htdocs/core/class/html.formai.class.php
```diff
@@ -106,12 +106,12 @@ public function getSectionForAIPrompt($function = 'textgeneration', $format = ''
...
 
-		$out = '<div id="ai_input'.$htmlContent.'" class="ai_input'.$htmlContent.' hidden paddingtop paddingbottom">';
-		$out .= '<input type="text" class="quatrevingtpercent" id="ai_instructions'.$htmlContent.'" name="instruction" placeholder="'.$langs->trans("EnterYourAIPromptHere").'..." />';
-		$out .= '<input id="generate_button'.$htmlContent.'" type="button" class="button smallpaddingimp"  value="'.$langs->trans('Generate').'"/>';
-		$out .= '<div id="ai_status_message'.$htmlContent.'" class="fieldrequired hideobject marginrightonly margintoponly">';
+		$out = '<div id="ai_status_message'.$htmlContent.'" class="fieldrequired hideobject marginrightonly margintoponly">';
 		$out .= '<i class="fa fa-spinner fa-spin fa-2x fa-fw valignmiddle marginrightonly"></i>'.$langs->trans("AIProcessingPleaseWait", getDolGlobalString('AI_API_SERVICE', 'chatgpt'));
 		$out .= '</div>';
+		$out .= '<div id="ai_input'.$htmlContent.'" class="ai_input'.$htmlContent.' hidden paddingtop paddingbottom">';
+		$out .= '<input type="text" class="quatrevingtpercent" id="ai_instructions'.$htmlContent.'" name="instruction" placeholder="'.$langs->trans("EnterYourAIPromptHere").'..." />';
```
eldy, line 113: Why moving the input "EnterYourAIPromptHere" after the div used for "AIProcessingPleaseWait" ?
It means that after entering a prompt, the message "wait will appear" and the prompt will be shifted lower.
The message "wait" shouldappear just under the prompt.

Hystepik: That was to prevent the use of multiple "AIProcessingPleaseWait" message. Because the message is inside the div of "EnterYourAIPromptHere" input this mean I cannot use it for text editorial features to tell to user that AI is working.

Do having 2 same "AIProcessingPleaseWait" message one for Prompt and one for text editorial feature is a better workaround ?

eldy: I see.
Yes we can share the same Wait message.

I would like to havethis king of look and feel (so all feature to support AI are in the same link called 
"AI enhancement" 
that will repace the link 
"[Remplir le message avec du contenu IA...]"

Can you evolve in the direction of the screenshot.
image

PR #33324, opened 2025-03-04
allow spaces on natural search

rycks:
Origin of the bug is a bank journal with spaces like that:

image

Then that journal code is displayed into filters and list with empty filter displays lines 

image

But as soon as i choose that journal and apply filter, list remains empty

and debug log display this SQL request 

```
sql=SELECT COUNT(*) as nbtotalofrecords FROM llx_accounting_bookkeeping as t WHERE t.entity IN (1) AND (t.code_journal IN ('ENC') AND t.code_journal IN ('SGAPSO'))
```

That's because of the use of natural_search on t.code_journal

### DIFF
--- a/htdocs/core/db/DoliDB.class.php
+++ b/htdocs/core/db/DoliDB.class.php
@@ -121,24 +121,17 @@ abstract class DoliDB implements Database
 	}
 
 	/**
-	 * Sanitize a string for SQL forging
-	 *
-	 * @param   string 	$stringtosanitize 	String to escape
-	 * @param   int		$allowsimplequote 	1=Allow simple quotes in string. When string is used as a list of SQL string ('aa', 'bb', ...)
-	 * @param   int		$allowspaces 		1=Allow spaces in string. When string is a sentense or a composed word
-	 * @return  string                      String escaped
-	 */
-	public function sanitize($stringtosanitize, $allowsimplequote = 0, $allowspaces = 0)
+		 * Sanitize a string for SQL forging
+		 *
+		 * @param   string 	$stringtosanitize 	String to escape
+		 * @param   int		$allowsimplequote 	1=Allow simple quotes in string. When string is used as a list of SQL string ('aa', 'bb', ...)
+		 * @param	int		$allowsequals		1=Allow equals sign
+		 * @param	int		$allowsspace		1=Allow space char
+		 * @return  string                      String escaped
+		 */
+	public function sanitize($stringtosanitize, $allowsimplequote = 0, $allowsequals = 0, $allowsspace = 0)
 	{
-		$regex = "^a-z0-9_\-\.,";
-		if ($allowspaces) {
-			$regex .= " ";
-		}
-		if ($allowsimplequote) {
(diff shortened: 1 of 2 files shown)


eldy:
This is the expected value we should have into phpunit (and that we don't):

		$s = natural_search("t.field", "abc def");
		$this->assertEquals($s, " AND (t.field LIKE '%abc%' AND t.field LIKE '%def%')");

		$s = natural_search("t.field", "abc def,ghi", 3);
		$this->assertEquals($s, " AND (t.field IN ('abc def', 'ghi'))");

The second test, for mode 3, is to use a list of code separated by a coma. Bug is confirmed because we do not get the expected 
` " AND (t.field IN ('abc def', 'ghi'))"`

I agree with the fix suggested in natural_search but i see not reason to change ->sanitize. I did not find case that does not return the expected value.

So I have pushed the fix in develop, i let you see how to backport it.
Note that among the 3 new phpunit test added in develop, only 2 are relavant for older version because the second one is a new feature of the future v22.

rycks:
@eldy as sanitize remove spaces i don't know how you could have "abc def" it will become "abcdef" ...

rycks:
> @eldy as sanitize remove spaces i don't know how you could have "abc def" it will become "abcdef" ...

ok you did the same update to allow spaces into sanitize function cf 97ae1142651965ef1568ceaccff5d22e6ce5ba5b

htdocs/core/db/DoliDB.class.php
```diff
@@ -125,15 +125,20 @@ public function lasterrno()
 	 *
 	 * @param   string 	$stringtosanitize 	String to escape
 	 * @param   int		$allowsimplequote 	1=Allow simple quotes in string. When string is used as a list of SQL string ('aa', 'bb', ...)
+	 * @param   int		$allowspaces 		1=Allow spaces in string. When string is a sentense or a composed word
 	 * @return  string                      String escaped
 	 */
-	public function sanitize($stringtosanitize, $allowsimplequote = 0)
+	public function sanitize($stringtosanitize, $allowsimplequote = 0, $allowspaces = 0)
```
eldy, line 131: The parameter $allowsspace was introduced in version v19 with signature
$allowsspace = 0

As it is a fix already fixed into a higher version, it must be fixed by doing a backport, so it means the fix must b done in a similar way: so parameter and var name must be same at same place (so must be $allowsspace and must be at position 4 in function signature)
So signature must be
```
public function sanitize($stringtosanitize, $allowsimplequote = 0, $allowsequals = 0, $allowsspace = 0)
```
even if param allowsequals is not used.

Because if you do not follow same signature, if you introduce the function differently, imagine someone, in a module is using it when he will upgrade to a newer version of dolibarr, his module will be broken.

rycks: for sure !

rycks: @eldy job is done : 

- sanitize function backported from dolibarr 19.0
- change call of db->sanitize used with right param orders

PR #33333, opened 2025-03-04
Qual: Fix phpstan by reverting config to 2.0

mdeweerd:
# Qual: Fix phpstan by reverting config to 2.0

Recent updated removed the ':2.0' modifier on the phpstan tool.


mdeweerd:
@eldy This resolves the issues with the "Unable to resolve the template type T in call to function dol_sort_array" notices, but reveils other notices - I guess related to code changes.

PR #33339, opened 2025-03-05
FIX: dolPrintHTML double htmlentities

jyhere:
Not sure this is the right way to fix the bug, dolPrintHTML() do not like ampersand because of double htmlentities (one in dol_escape_htmltag(), one in dol_htmlentitiesbr()). I add a optionnal $nohtmlentities parameter to the dol_escape_htmltag() function to bypass htmlentities() call.

Tested on export page.

Before : 
image

After : 
image

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -2082,7 +2082,7 @@ function dolPrintText($s)
  */
 function dolPrintHTML($s, $allowiframe = 0)
 {
-	return dol_escape_htmltag(dol_htmlwithnojs(dol_string_onlythesehtmltags(dol_htmlentitiesbr($s), 1, 1, 1, $allowiframe)), 1, 1, 'common', 0, 1);
+	return dol_escape_htmltag(dol_htmlwithnojs(dol_string_onlythesehtmltags(dol_htmlentitiesbr($s), 1, 1, 1, $allowiframe)), 1, 1, 'common', 0, 1, true);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


eldy:
We should not introduce parameters that bypass security sanitization.
If value is encoded twice it may need that we ask to print a value that was already encoded. In such a case it must be encoded twice.
I am on develop branch and i do not experience the trouble:
image

what is the language code you use ?

jyhere:
I'm using the demo db with French language (fr_FR). 
Tested on two differents environnements (local docker and production server, both with php 7.4).

eldy:
> I'm using the demo db with French language (fr_FR). Tested on two differents environnements (local docker and production server, both with php 7.4).

Hum strange, the text "Utilisateurs & groupes" does not depend on data but on the language file only. And I can't reproduce your bug with v20, v21 either develop.

Do you have a forced specific value for this hidden cosntant:
MAIN_RESTRICTHTML_ONLY_VALID_HTML
MAIN_RESTRICTHTML_ONLY_VALID_HTML_TIDY
MAIN_RESTRICTHTML_REMOVE_ALSO_BAD_ATTRIBUTES
?
Can you also confirm the url of the page ?

jyhere:
I can reproduce the bug with V21,but dolPrintHTML()  was not used here in V20 and prior.
Frederic reproduced the bug with v22 / PHP 8.2 

image

Theses screens comes from exports/export.php?leftmenu=export page

I checked the constants
MAIN_RESTRICTHTML_ONLY_VALID_HTML
MAIN_RESTRICTHTML_ONLY_VALID_HTML_TIDY
MAIN_RESTRICTHTML_REMOVE_ALSO_BAD_ATTRIBUTES
None is set.

The bug seems inevitable, because of htmlentities() called twice on the same string, I don't know how you don't get the bug. :D
L8838 (in dol_htmlentities() function)
L2263 (in dol_escape_htmltag() function)


Maybe we should set the following parameter to 1 from dol_escape_htmltag ?
 @param		int			$escapeonlyhtmltags		1=Escape only html tags, not the special chars like accents.

eldy:
Found the difference. Problem disapper when option 

MAIN_RESTRICTHTML_REMOVE_ALSO_BAD_ATTRIBUTES=0
AND
MAIN_RESTRICTHTML_ONLY_VALID_HTML_TIDY=1

Need to find why result is not the same...

eldy:
I pushed a fix into develop branch.
However, because it is not a critcal bug, but the fix is critical and with a high risk of regression, it will remains fixed into develop only.
Commit 202ffe732e09dba631aa3e12322cd53d7cf1597d

Thanks to check on develop branch if it's ok for you.

PR #33343, opened 2025-03-05
FIX #33169 Anyone can cancel vacation requests

### LINKED ISSUE #33169: Anyone can cancel vacation requests?
### Bug

A user with rights to read all vacation requests can cancel requests of other users.

### Dolibarr Version

20.0.4

### Environment PHP

7.3

### Environment Database

MariaDB 10.3.39

Here are the rights of my user :
Image

And here is a vacation request for someone that is not me, and for whom I am not the hierarchical superior :
Image

I tested this with other profiles (with the same rights) and it's the same.
I also tested on an old version (18.0.4) and it's the same.

Is this intended? Is there any other right that might have an impact on this?

### Attached files

_No response_

Daoud-mohamed:
FIX #33169
This PR fixes an issue where any user could cancel vacation requests


htdocs/holiday/card.php
```diff
@@ -1598,7 +1598,7 @@
 
 							// Button Cancel (because we can't approve)
 							if ($permissiontoadd || $permissiontoaddall) {
-								if (($object->date_fin > dol_now()) || !empty($user->admin)) {
+								if (($object->date_fin > dol_now() && $user->id == $object->fk_user) || !empty($user->admin)) {
```
eldy, line 1601: The permission should already be tested by the 
if ($permissiontoadd || $permissiontoaddall) {

Daoud-mohamed: In the following condition:
`if user->hasRight('holiday', 'write) and in_array($fuserid, $childids)) {`

The expression `in_array($fuserid, $childids)` will always evaluate to true in the cancel action case because:

Parameter `fuserid` does not exist so `$fuserid = (GETPOSTINT('fuserid') ? GETPOSTINT('fuserid') : $user->id);` will take current user ID
also `$childids = $user->getAllChildIds(1); ` with parameter 1 will include current user ID 
so I used `$object->fk_user` instead of `$fuserid` since we are in edit mode.
To fix this, I replaced `$fuserid` with `$object->fk_user` because in the cancel action case `$fuserid` will be always current user ID.

eldy: $fuserid is the variable that represent the ID of the user we are trying to edit/or cancel.
So for me the bug is when we set $fuserid.

Your fix will work for the case of cancel. But it meas the $permissiontoadd is wrong, so it means there is other bugs at other place.
Can you try this fix ?

`$fuserid = (GETPOSTINT('fuserid') ? GETPOSTINT('fuserid') : ($action == 'create' ? $user->id : 0));`

The user we want to edit is the one in parameter, and we set it to ourself (the default value) only if we enter the page to create a new leave. In other cases, we should retreived the edited user from parameter.

PR #33353, opened 2025-03-06
FIX sales representatives can affect contacts from thirdparties they are not allowed to see

### LINKED ISSUE #33350: "Contacts" tabs : User can affect contacts from thirdparties they are not allowed to see
### Bug

When a user does not have the right to see all thirdparties, they can still view all thirdparties in "Contacts" tab, and affect contacts from thirdparties they should not be allowed to manage.

"does not have the right to see all thirdparties" means that user does not have right `societe, client, voir`

Observed on a DLB14 and on develop branch. However, since this fix would bring a change in behavior, it would better be included in develop.

To reproduce :

- your instance should have at least two thirdparties, each with at least one contact. Here we have Alice and Chapelier fou.

- create a user which does not have `societe, client, voir`. Here Côme Hercial.
- allow this user to view, create, modify commercial proposals (for the example)

- affect the user to one thirdparty as a commercial.
- create a proposal for the same thirdparty.
- go to the "Contacts" tab
- observe you can select thirdparty "Alice" and affect contacts from Alice to this proposal.

Expected behavior :
- on the front, the user should not be able to select a thirdparty if the user is not affected to the thirdparty.
- on the back, when a user adds/removes a user from any object, we shoudl verify that they can make changes for this contact.

Proposed implementation :

1. introduce a method `User::getAffectedThirdparties($only_ids = 1)` that would return all thirdparties on which the user is a commercial. Returns an array of IDs if param $only_ids is set, else returns an array of Thirdparties.

2. in `core/tpl/contacts.tpl.php`, just before `selectCompaniesForNewContact`, depending on the users right, call User::getAffectedThirdparties(1), and pass the result as the `$limitto` param of `selectCompaniesForNewContact`. This should solve the problem on the front.

3. in CommonObject::add_contact(), call `$allowed_thirdparties = $user->getAffectedThirdparties(1)`, then do a SQL request to verify that `$fk_socpeople` belongs to one of `$allowed_thirdparties`. Returns with an error if it is the case.

### Dolibarr Version

observed on V14, confirmed on V18 and develop. I guess each version  between 14 and develop. However, since the fix would introduce a change in behavior, I think it should be implemented in develop.

### Environment PHP

any

### Environment Database

any

- your DLB instance should have at least two thirdparties, each with at least one contact. Here we have Alice and Chapelier fou.
- create a user which does not have `societe, client, voir`. Here Côme Hercial. Allow this user to view, create, modify commercial proposals (for the example)

Image

- affect the user to one thirdparty as a commercial. (here Chapelier Fou)
- make sure the user is **not** affected to the other thirdparty (here Alice)

Image

- create a proposal for the same thirdparty (here Chapelier Fou).
- go to the "Contacts" tab
- observe you can select thirdparty "Alice" and affect contacts from Alice to this proposal.

Image

Image

### Expected behavior :

- on the front, the user should not be able to select a thirdparty if the user is not affected to the thirdparty.
- on the back, when a user adds/removes a user from any object, we shoudl verify that they can make changes for this contact.

### Suggested fix :

1. introduce a method `User::getAffectedThirdparties($only_ids = 1)` that would return all thirdparties on which the user is a commercial. Returns an array of IDs if param $only_ids is set, else returns an array of Thirdparties.

2. in `core/tpl/contacts.tpl.php`, just before `selectCompaniesForNewContact`, depending on the users right, call User::getAffectedThirdparties(1), and pass the result as the `$limitto` param of `selectCompaniesForNewContact`. This should solve the problem on the front.

3. in CommonObject::add_contact(), call `$allowed_thirdparties = $user->getAffectedThirdparties(1)`, then do a SQL request to verify that `$fk_socpeople` belongs to one of `$allowed_thirdparties`. Returns with an error if it is the case.

### Attached files

_No response_

thomas-Ngr:
# FIX #33350 : sales representatives can affect contacts from companies they are not allowed to see

On contact tabs for propal/facture/..., users could 
- see thirdparties they wer not supposed to see
- affect contacts to the object from these thidparties

more details in the issue.


htdocs/core/tpl/contacts.tpl.php
```diff
@@ -181,7 +183,18 @@
...
+			if ($object->restrictiononfksoc && ! $permission_view_all_thirdparties) {
+				$allowed_thirdparties = $user->getAffectedThirdparties(1);
+				if (is_int($allowed_thirdparties)) {
+					print $user->errorsToString();
+					$allowed_thirdparties = [$object->socid];
+				} elseif (empty($allowed_thirdparties)) {
+					$allowed_thirdparties = [$object->socid];
+				}
+			}
+			$selectedCompany = $formcompany->selectCompaniesForNewContact($object, 'id', $selectedCompany, 'newcompany', $allowed_thirdparties, 0, '', 'minwidth300imp maxwidth400 widthcentpercentminusx');	// This also print the select component
```
eldy, line 196: The test should be included into the selectCompaniesForNewContact for 2 reasons:
- it is a security condition, so should never be bypassed so should not be decided by each caller of selectCompaniesForNewContact but by selectCompaniesForNewContact itself.
- second reason is that getAffectedThirdparties can return millions of lines making memory allocation for million of php object. Will generate a memory error on large database. Filter must be directly inside the SELECT inside the selectCompaniesForNewContact.

htdocs/core/class/commonobject.class.php
```diff
@@ -1262,6 +1262,20 @@ public function add_contact($fk_socpeople, $type_contact, $source = 'external',
 			return -2;
 		}
 
+		if ($this->restrictiononfksoc && ! $user->hasRight('societe', 'client', 'voir')) {
+			$allowed_thirdparties = $user->getAffectedThirdparties(1);
+			$sql_allowed_contacts = 'SELECT rowid FROM '.$this->db->prefix().'socpeople';
+			$sql_allowed_contacts.= ' WHERE fk_soc IN ('.$this->db->sanitize(implode(',', $allowed_thirdparties)).')';
```
eldy, line 1268: The number of element into a IN has a limit (generally errors appears when we are over max_allowed_packet) into size of the subrequest.

Instead of dowing a getAffectedThirdparties to get all thirdparties allowed by a user and then a select to get all socpeople for thes thirdparies,  why not just doing a 

SELECT COUNT(*) FROM llx_societe as s JOIN llx_socpeople as sp...

htdocs/core/tpl/contacts.tpl.php
thomas-Ngr: Thanke. I have implemented things your way.

htdocs/core/class/commonobject.class.php
thomas-Ngr: You are right, I did it this way.

PR #33360, opened 2025-03-06
FIX make ToS concat to pdf actually work

t3cneo:
# FIX
Just updated to 21 to finally be able to merge ToS to quotes and invoices and noticed it is not working. This is my attempt to make it actually work.

### DIFF
--- a/htdocs/core/modules/commande/doc/pdf_eratosthene.modules.php
+++ b/htdocs/core/modules/commande/doc/pdf_eratosthene.modules.php
@@ -864,9 +864,9 @@ class pdf_eratosthene extends ModelePDFCommandes
 				}
 
 				// Add terms to sale
-				if (getDolGlobalString("MAIN_INFO_ORDER_TERMSOFSALE") && getDolGlobalInt('MAIN_PDF_ADD_TERMSOFSALE_ORDER')) {
-					$termsofsalefilename = getDolGlobalString('MAIN_INFO_ORDER_TERMSOFSALE');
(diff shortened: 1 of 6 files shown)


htdocs/core/modules/commande/doc/pdf_eratosthene.modules.php
```diff
@@ -864,9 +864,9 @@ public function write_file($object, $outputlangs, $srctemplatepath = '', $hidede
 				}
 
 				// Add terms to sale
-				if (getDolGlobalString("MAIN_INFO_ORDER_TERMSOFSALE") && getDolGlobalInt('MAIN_PDF_ADD_TERMSOFSALE_ORDER')) {
-					$termsofsalefilename = getDolGlobalString('MAIN_INFO_ORDER_TERMSOFSALE');
-					$termsofsale = $conf->order->dir_output.'/'.$termsofsalefilename;
+				if (getDolGlobalString("MAIN_INFO_SOCIETE_TERMSOFSALE") && getDolGlobalInt('MAIN_PDF_ADD_TERMSOFSALE_ORDER')) {
+					$termsofsalefilename = getDolGlobalString('MAIN_INFO_SOCIETE_TERMSOFSALE');
```
eldy, line 868: On devleop branch, the feature to add term of sale allows to define a different file for each type of object.
Si we must have here MAIN_INFO_ORDER_TERMSOFSALE for order, MAIN_INFO_PROPAL_TERMSOFSALE for proposal, etc...

PR #33373, opened 2025-03-08
NEW On list, see complete ref customer/supplier with constant MAIN_SHOW_GLOBAL_REF_CUSTOMER_SUPPLIER

aspangaro:
The goal of this new constant is to show complete ref customer/supplier directly in list without "css truncate".

If ok, I will extended to other list


htdocs/compta/facture/list.php
```diff
@@ -2236,7 +2236,7 @@
 
 			// Customer ref
 			if (!empty($arrayfields['f.ref_client']['checked'])) {
-				print '<td class="nowrap tdoverflowmax200">';
+				print '<td class="nowrap '.(getDolGlobalInt('MAIN_SHOW_GLOBAL_REF_CUSTOMER_SUPPLIER') ? '' : 'tdoverflowmax200').'">';
```
eldy, line 2239: The good prectice is too add first a
`title="'.dolPrintHtmlForAttribut($obj->ref_client);'"`
 on the td line
So if ref is very long, we can get the value by setting the mouse on ref. This is implemented this way everywhere else.

We can also add MAIN_SHOW_GLOBAL_REF_CUSTOMER_SUPPLIER but this break a lot of things (css, responsive) and should be done only if the alt is not enough.
Can you add the title too (or just the title if it is enough) ?

aspangaro:
New version with title & modifiy css

Example with 255 caracters:

Without: 
image

With:
image

Issue #33362, opened 2025-03-07
Can't close a year with "À-nouveaux" => BookkeepingRecordAlreadyExists


Liammer:
### Bug

Hi there!
I tried to close the year 2024 but I always have the "BookkeepingRecordAlreadyExists" error when choosing the "À-nouveaux" (on 21.0.0 but it did the same on 20.0.4). It works well when I simply close the year.
I tried to understand why! There's an sql request that is repeated (I've done a simple 'echo' of the $sql request in bookkeeping.class.php', around line 350): 
```SELECT count(*) as nb FROM llx_accounting_bookkeeping WHERE doc_type = 'closure' AND fk_doc = 1 AND numero_compte = '120' AND label_operation = '2025' AND entity = 1```
is the first request sent. I assume that if $row->nb == 0 it will be written in the table (and deleted later if there's an error).
In the 17th request, I find exactly the same request! Here is what the echo shows (with result of $row->nb)
```
SELECT count(*) as nb FROM llx_accounting_bookkeeping WHERE doc_type = 'closure' AND fk_doc = 1 AND numero_compte = '120' AND label_operation = '2025' AND entity = 1
0
SELECT count(*) as nb FROM llx_accounting_bookkeeping WHERE doc_type = 'closure' AND fk_doc = 1 AND numero_compte = '129' AND label_operation = '2025' AND entity = 1
0
SELECT count(*) as nb FROM llx_accounting_bookkeeping WHERE doc_type = 'closure' AND fk_doc = 1 AND numero_compte = '271' AND label_operation = '2025' AND entity = 1
0
SELECT count(*) as nb FROM llx_accounting_bookkeeping WHERE doc_type = 'closure' AND fk_doc = 1 AND numero_compte = '401' AND label_operation = '2025' AND entity = 1
0
SELECT count(*) as nb FROM llx_accounting_bookkeeping WHERE doc_type = 'closure' AND fk_doc = 1 AND numero_compte = '411' AND label_operation = '2025' AND entity = 1
0
SELECT count(*) as nb FROM llx_accounting_bookkeeping WHERE doc_type = 'closure' AND fk_doc = 1 AND numero_compte = '431' AND label_operation = '2025' AND entity = 1
0
SELECT count(*) as nb FROM llx_accounting_bookkeeping WHERE doc_type = 'closure' AND fk_doc = 1 AND numero_compte = '4417002' AND label_operation = '2025' AND entity = 1
0
[...]

PR #33384, opened 2025-03-09
NEW Accountancy - Add centralized account (SQL & migration part)

aspangaro:
#33362

See this discussion (in french) for an explanation of this functionality.

https://www.dolibarr.fr/forum/t/cloture-dannee-bookkeepingrecordalreadyexists/48658/23?u=aspangaro-inovea

------------------------------------------------
Discussion is about accountancy annual closure

For reasons of compatibility with other software, you must avoid empty auxiliary accounts in what are known as centralizing accounts, which group together employees (421xxx), customers (411xxx), suppliers (401xxx) and expense accounts since v20 (425xxx).

However, we don't identify these centralizing accounts in Dolibarr, as they are accounts that require an attached auxiliary. If they were identified with a simple option, scripts could be adapted to check whether an auxiliary account has been declared if a centralizing account is used. This applies to both manual and automatic entries.

Another problem is that our sub-ledger accounts don't know their centralizer and are therefore transverse. If we use the 401 account and the 9ASSODOLIBARR sub-ledger account, we can use it on the 411 account and it won't make any difference; when we view it in the general ledger, we'll see the two operations in the same sub-ledger account, even though they have nothing to do with each other.

Closing with sub-ledger details either can't work without this, or is complex to correct.

A final point that has long been a problem: if Dolibarr doesn't have an auxiliary on a transaction, it reuses the G/L account as an auxiliary. You end up with 411 / 411... An auxiliary account that doesn't exist...
If the accounting numerator for third parties is not set up, it's a disaster! In most cases, this means a complete overhaul of the accounting integration


alexandre-janniaux:
Looks good to me for this MR, but either some form of the repair.php part might become needed at some point, which I find it hard to do (modifying accounting data that might be locked), or we need to also have an additional field on fiscal_period to indicate whether we want to apply the (future) rule or not.

aspangaro:
I'm sorry @alexandre-janniaux , but I don't understand why I'm going to change the anteriority of the data. As far as I'm concerned, this applies from the new version (v22 if PR accepted) and in terms of code, you'll have to make a page listing the problems, centralizing accounts that don't have a declared auxiliary or auxiliaries that don't have the right centralizer.

htdocs/install/mysql/migration/21.0.0-22.0.0.sql
```diff
@@ -194,3 +194,6 @@ ALTER TABLE llx_product_customer_price ADD CONSTRAINT fk_product_customer_price_
 ALTER TABLE llx_product_customer_price ADD CONSTRAINT fk_product_customer_price_fk_soc FOREIGN KEY (fk_soc) REFERENCES llx_societe(rowid);
 UPDATE llx_product_customer_price SET date_begin = datec WHERE date_begin IS NULL;
 UPDATE llx_product_customer_price_log SET date_begin = datec WHERE date_begin IS NULL;
+
+ALTER TABLE llx_accounting_account ADD COLUMN centralized tinyint DEFAULT 0 NOT NULL AFTER active;
+UPDATE llx_accounting_account AS acc JOIN llx_const AS cst ON acc.account_number = cst.value SET acc.centralized = 1 WHERE cst.name in (__ENCRYPT('ACCOUNTING_ACCOUNT_CUSTOMER')__,__ENCRYPT('ACCOUNTING_ACCOUNT_SUPPLIER')__,__ENCRYPT('SALARIES_ACCOUNTING_ACCOUNT_PAYMENT')__);
```
eldy, line 199: Don't think this request is valid for pgsql.
do you want to do something like this ?

`UPDATE llx_accounting_account as acc SET acc.centralized = 1 WHERE acc.account_number in (SELECT value  FROM llx_const WHERE name IN (__ENCRYPT('ACCOUNTING_ACCOUNT_CUSTOMER')__,__ENCRYPT('ACCOUNTING_ACCOUNT_SUPPLIER')__,__ENCRYPT('SALARIES_ACCOUNTING_ACCOUNT_PAYMENT')__));
`

aspangaro: Thanks @eldy. You are right, my request works on mariadb but on pgsql...

PR corrected!

PR #33390, opened 2025-03-10
FIX : PAIEMENT Wrong field displayed for DateChequeReceived

warnerbryce:
FIX

On compta/paiement/cheque/card.php - New Deposit slip page
There is a mismatch on the date you are looking for Checks Deposit and the date you see in the list.

By reading the code line 444 to 464 You can see that $search_date_start has something to deal with date (of the payment).
Whereas in line 547, the part where the list is built, the $value["date"] refers to Creation date (of the payment) (cf line 471) which is not the same.

My fix try to apply the "right" date on the list.
I mean right because we need to use the date of the payment and not its creation date.
We can create a payment the 2nd of Febrary, but we want to deposit the check on the 2nd of march.
In order to make a deposit slip at the good date, we need to use the date of the payment, not the creation date.

### DIFF
--- a/htdocs/compta/paiement/cheque/card.php
+++ b/htdocs/compta/paiement/cheque/card.php
@@ -544,7 +544,7 @@ if ($action == 'new') {
 		if (count($lines[$bid])) {
 			foreach ($lines[$bid] as $lid => $value) {
 				print '<tr class="oddeven">';
-				print '<td>'.dol_print_date($value["date"], 'day').'</td>';
+				print '<td>'.dol_print_date($value["paymentdate"], 'day').'</td>';
 				print '<td>'.$value["numero"]."</td>\n";
 				print '<td>'.$value["emetteur"]."</td>\n";
 				print '<td>'.$value["banque"]."</td>\n";


htdocs/compta/paiement/cheque/card.php
```diff
@@ -544,7 +544,7 @@
 		if (count($lines[$bid])) {
 			foreach ($lines[$bid] as $lid => $value) {
 				print '<tr class="oddeven">';
-				print '<td>'.dol_print_date($value["date"], 'day').'</td>';
+				print '<td>'.dol_print_date($value["paymentdate"], 'day').'</td>';
```
eldy, line 547: It seems the label depends on $type
```
if ($type == 'CHQ') {
			print $langs->trans("DateChequeReceived");
		} else {
			print $langs->trans("DatePaymentForDepositReceived", $type);
		}
```
So may be the value should also depend on $type ?

warnerbryce: Hello @eldy 
The label depends on $type it's true. Can be CHQ or FAC or GIR etc...

But the problem stays the same, whatever the type is. 
At the moment, the list displays $value["date"] => $obj->datec . We need to see the payment date. 

I've run some tests with $type == 'FAC' and 'LIQ' with $value["paymentdate"] it's working, no needs to do a IF statement. The correct value is written in the database always at the same spot llx_bank -> dateo .

By the way DatePaymentForDepositReceived don't have translation at all.

philippe-opendsi (approved):

PR #33392, opened 2025-03-10
Fix hook completefieldstosearchall with sellyoursaas

atm-jonathan:
# Instructions
**/societe/list.php?search_all=mycompany**
Problem setting up the “completeFieldsToSearchAll” hook in the list of third parties
in the sql query, the fields **username_os, database_db, username_db** will create an error because they don't exist in the llx_societe_extrafield table 

This problem is not encountered in the list of contracts, as the extrafields username_os, database_db, username_db have been added by SellYourSaas.

hook declaration in action_sellyoursaas.class.php
```
	public function completeFieldsToSearchAll($parameters, &$action, $hookmanager)
	{
		$this->results['fieldstosearchall']['username_os'] = 'Username OS';
		$this->results['fieldstosearchall']['database_db'] = 'Database DB';
		$this->results['fieldstosearchall']['username_db'] = 'Username DB';
		return 0;
	}
```
What do we do ?
delete the hook call like on my PR?
add the extrafields username_os, database_db and username_db on llx_societe_extrafields as for contracts?
Or do you find a solution in societe/list.php, but how do you avoid adding SellYourSaas-specific information?
Or maybe change the hook name?


eldy:
Bug looks into the sellyoursaas hook no into dolibarr. So i close this ticket.
i pushed a fix in sellyoursaas.

	public function completeFieldsToSearchAll($parameters, $object, $action, $aa)
	{
		if ($object->element == 'societe') {
			$this->results['fieldstosearchall']['ef.firstname'] = 'Firstname';
			$this->results['fieldstosearchall']['ef.lastname'] = 'Lastname';
		}
		if ($object->element == 'contrat') {
			$this->results['fieldstosearchall']['username_os'] = 'Username OS';
			$this->results['fieldstosearchall']['database_db'] = 'Database DB';
			$this->results['fieldstosearchall']['username_db'] = 'Username DB';
		}

		return 0;
	}

PR #33393, opened 2025-03-10
fix: thirdparty substitutions var ignored on automatic event

### LINKED ISSUE #33295: Thirdparty variables substitution ignored when paid invoice notification is sent
### Bug

I'm trying to setup automatic notification when customer invoice is paid.  I would need to customize email template and add thirdparty information. But none of thirdparty variable (__THIRDPARTY_XX__) variables are replaced.

### Dolibarr Version

20.0.3

### Environment PHP

8.2.27

### Environment Database

Mariadb 11.7.2

1. Setup an email template as followed

> Dear __THIRDPARTY_NAME__,
> 
> Here will be the invoice ref:  __REF__.
> 

2. Configure as template for customer paid invoice notification
3. Create an invoice an mark it as paid

Here is what I received

> Dear **__THIRDPARTY_NAME__**,
> 
> Here will be the invoice ref:  FA2502-0007.

Here is what I would expect

> Dear **MyThirdpartyName**,
> 
> Here will be the invoice ref:  FA2502-0007.

### Attached files

_No response_

atm-irvine:
# FIX|Fix #33295 Thirdparty variables substitution ignored
Thirdparty substitutions var ignored because thirdparty object has never been fetched on automatic event

### DIFF
--- a/htdocs/core/class/notify.class.php
+++ b/htdocs/core/class/notify.class.php
@@ -1219,7 +1219,9 @@ class Notify
 				}
 				if (!empty($mailTemplateLabel) && is_object($emailTemplate) && $emailTemplate->id > 0) {
 					if (property_exists($object, 'thirdparty')) {
-						$object->fetch_thirdparty();
+						if (!($object->thirdparty instanceof Societe)) {
+							$object->fetch_thirdparty();
+						}
 
 						if ($object->thirdparty instanceof Societe && $object->thirdparty->default_lang && $object->thirdparty->default_lang != $langs->defaultlang) {
 							$outputlangs = new Translate('', $conf);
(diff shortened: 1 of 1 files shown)


htdocs/core/class/notify.class.php
```diff
@@ -1218,10 +1218,14 @@ public function send($notifcode, $object, $filename_list = array(), $mimetype_li
 					$emailTemplate = $formmail->getEMailTemplate($this->db, $object_type.'_send', $user, $outputlangs, 0, 1, $mailTemplateLabel);
 				}
 				if (!empty($mailTemplateLabel) && is_object($emailTemplate) && $emailTemplate->id > 0) {
-					if (property_exists($object, 'thirdparty') && $object->thirdparty instanceof Societe && $object->thirdparty->default_lang && $object->thirdparty->default_lang != $langs->defaultlang) {
-						$outputlangs = new Translate('', $conf);
-						$outputlangs->setDefaultLang($object->thirdparty->default_lang);
-						$outputlangs->loadLangs(array('main', 'other'));
+					if (property_exists($object, 'thirdparty')) {
+						$object->fetch_thirdparty();
```
eldy, line 1222: Weshould avoid this if $object->thirdparty is already loaded (already instanceof Societe).

atm-irvine:
@eldy It's corrected

PR #33399, opened 2025-03-10
Fix CI: Do not require some fields in the 'title' definition of $cols

mdeweerd:
# Fix CI: Do not require some fields in the 'title' definition of $cols


mdeweerd:
@eldy Fixed ci by adapting phpdoc to new usage, fixing one global definition fetch, the missing rename of $attrV, and also fixed a MismatchArgument in the commondocgenerator to clean that file from its last phan notice as well.

PR #33402, opened 2025-03-11
Fix Supplier order tags

aspangaro:
Fix error on tags


josett225:
Hi @aspangaro 
We want to get also tags on proposals, did you start the work or do you know if someone has started it? If not based on your work we can do it.

PR #33404, opened 2025-03-11
Fix #33403: E-Mailcollector doesn't find Thirdparty by E-Mail Address of Contact

### LINKED ISSUE #33403: E-Mailcollector doesn't find Thirdparty by E-Mail Address of Contact
### Bug

When using action "loadthirdparty" there are different attempts to find the right thirdparty. if parameter "email" is specified, also contacts of thirdparties are searched using the method findNearest. This method declares a local instance of the Contact object and does not return the ID found

### Dolibarr Version

20.0

### Environment PHP

na

### Environment Database

na

_No response_

### Attached files

_No response_

glu000:
FIX #33403
E-Mailcollector doesn't find Thirdparty by E-Mail Address of Contact

### DIFF
--- a/htdocs/contact/class/contact.class.php
+++ b/htdocs/contact/class/contact.class.php
@@ -1189,8 +1189,7 @@ class Contact extends CommonObject
 		// Then search on email
 		// Then search on lastname + firstname
 		// Then search ref_ext or alias with a OR
-		$tmpcontact = new Contact($this->db);
-		$result = $tmpcontact->fetch($id, null, $ref_ext, $email, 0, $socid);
+		$result = $this->fetch($id, null, $ref_ext, $email, 0, $socid);
 
 		return $result;
 	}


htdocs/contact/class/contact.class.php
```diff
@@ -1189,8 +1189,7 @@ public function findNearest($id = 0, $lastname = '', $firstname = '', $ref_ext =
 		// Then search on email
 		// Then search on lastname + firstname
 		// Then search ref_ext or alias with a OR
-		$tmpcontact = new Contact($this->db);
-		$result = $tmpcontact->fetch($id, null, $ref_ext, $email, 0, $socid);
+		$result = $this->fetch($id, null, $ref_ext, $email, 0, $socid);
```
eldy, line 1192: This looks strange because the function should return only the ID found with no need to load data.
Loading of data is done later with $thirdpartystatic->fetch($result); line 2672 of emailcollector.class.php

glu000:
The fix is not working, see my comment here: https://github.com/Dolibarr/dolibarr/commit/47af73d891d84a7750ea87165758fc0e4b36a61a#comments

eldy:
> The fix is not working, see my comment here: [47af73d#comments](https://github.com/Dolibarr/dolibarr/commit/47af73d891d84a7750ea87165758fc0e4b36a61a#comments)

This is where is the trouble: A fetch must return the ID when it found.
Most dolibarr code rely on this.
So i pushed a fix to return the id when found. Is it ok with that ?

glu000:
Thanks @eldy , now it looks great 👍

PR #33418, opened 2025-03-11
Fix ci: Use table_element to get table name for class

mdeweerd:
# Fix ci: Use table_element to get table name for class

Export was referencing lowercase of class name, not table_element.

### DIFF
--- a/htdocs/core/modules/modCategorie.class.php
+++ b/htdocs/core/modules/modCategorie.class.php
@@ -4,7 +4,7 @@
  * Copyright (C) 2012-2016	Juanjo Menent			<jmenent@2byte.es>
  * Copyright (C) 2020		Stéphane Lesage			<stephane.lesage@ateis.com>
  * Copyright (C) 2022-2025	Solution Libre SAS		<contact@solution-libre.fr>
- * Copyright (C) 2024		MDW						<mdeweerd@users.noreply.github.com>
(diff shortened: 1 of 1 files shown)


mdeweerd:
@eldy This is the fix for the failing PHPUNIT test.

htdocs/core/modules/modCategorie.class.php
```diff
@@ -821,14 +821,15 @@ protected function exportTagLinks(int $r, string $categcode, string $class, stri
 		$keyforselect = $class;
 		$keyforelement = $class;
 		$keyforaliasextra = 'extra';
+		$class_table = (new $class($db))->table_element;
```
eldy, line 824: I cant find wherecis the include that guarantee the new will not fails ?

mdeweerd: Either it is up to the caller,
or there a way load all classes,
or there is a way to match "CommandeFournisseur" to "htdocs/fourn/class/fournisseur.commande.class.php"
or there is a way to match "CommandeFournisseur" to "commande_fournisseur" for the table name
or files needs to be "renamed" (while keeping backward compatibility)
and/or there should be a function "dol_include_class" that would include the proper file.


getElementProperties does not help.

So, the class file needs to be included by the caller which is luckily only found in the same file.

The code is updated accordingly.

PR #33428, opened 2025-03-12
FIX: bookkeeping/export error when not reconciled option is checked

jyhere:
How to reproduce the bug : 
- Activate lettering feature (Accountancy Configuration)
- Go to Accountancy > Export Accountancy (accountancy/bookkeeping/export.php)
- Check Not reconciled option and refresh your list
- Export to file =>  the following error will occur : 
image


htdocs/accountancy/class/bookkeeping.class.php
```diff
@@ -1153,6 +1153,8 @@ public function fetchAll($sortorder = '', $sortfield = '', $limit = 0, $offset =
 						} else {
 							$sqlwhere[] = natural_search("t.code_journal", $value, 3, 1);
 						}
+					} elseif ($key == 'customsql') {
```
eldy, line 1156: Use of "customsql" has been deprecated and is now forbidden to forge filters.

jyhere: Damn, that's what I'd done originally, but I'd found it “cleaner” to use a customsql option in the class... I didn't know it was deprecated. :) I close and reopen a PR with the right fix.

PR #33431, opened 2025-03-12
FIX: remove code that does nothing

atm-florianm:
# FIX
Remove some dead code


eldy:
Is there a known visible bug ?
If no, PR must be pushed into develop.

eldy:
It seems this cleaning has already be done into develop branch so i close PR.

PR #33435, opened 2025-03-13
#FIX revertonoff and forcereload error

mapiolca:
Fix revertonoff and forcereload error


eldy:
What is the effect of adding
`$item->fieldParams = array('revertonoff' => '', 'forcereload' => '') ;` ?
Is there a change compared to not having the line at all ?

mapiolca:
Without this line, Dolibarr show an error : 
image

eldy:
So i pushed a fix that should solve the problem at the root: https://github.com/Dolibarr/dolibarr/commit/8449961d9e98bb18ef3dd40674075989c3443022

PR #33437, opened 2025-03-13
FIX revertonoff & forcereload error

mapiolca:


eldy:
With fix https://github.com/Dolibarr/dolibarr/commit/8449961d9e98bb18ef3dd40674075989c3443022, PR should become useless.

PR #33439, opened 2025-03-13
FIX: Update import_xlsx.modules.php

prietojc:
$keyfield must be always 'rowid' because we compare with $lastinsertid

### DIFF
--- a/htdocs/core/modules/import/import_csv.modules.php
+++ b/htdocs/core/modules/import/import_csv.modules.php
@@ -962,7 +962,7 @@ class ImportCsv extends ModeleImports
 								if (empty($keyfield)) {
 									$keyfield = 'rowid';
 								}
-								
+
 								$sqlSelect .= " WHERE ".$keyfield." = ".((int) $lastinsertid);
 
 								if (!empty($tablewithentity_cache[$tablename])) {
--- a/htdocs/core/modules/import/import_csv.modules.php
+++ b/htdocs/core/modules/import/import_csv.modules.php
@@ -959,9 +959,9 @@ class ImportCsv extends ModeleImports
 								// may already exists. So we rescan the extrafield table to know if record exists or not for the rowid.
 								// Note: For extrafield tablename, we have in importfieldshidden_array an entry 'extra.fk_object'=>'lastrowid-tableparent' so $keyfield is 'fk_object'
 								$sqlSelect = "SELECT rowid FROM ".$tablename;
-
-								
+								if (empty($keyfield)) {
 									$keyfield = 'rowid';
+								}
 								
 								$sqlSelect .= " WHERE ".$keyfield." = ".((int) $lastinsertid);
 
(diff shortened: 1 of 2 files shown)


eldy:
Can you describe how to reproduce the bug ,

prietojc:
When defining a new import with hidden fields ​​and using a file with an existing reference.

The **`$keyfield`** variable is used in the last hidden fields loop, so that when the SQL query is executed, it always takes the last value from the hidden fields array (`import_fieldshidden_array[]`). However, this is incorrect, the WHERE clause compares with `$lastinsertid`, the query should always be: **`WHERE rowid = $lastinsertid`**

htdocs/core/modules/import/import_csv.modules.php
```diff
@@ -960,9 +960,9 @@ public function import_insert($arrayrecord, $array_match_file_to_database, $obji
 								// Note: For extrafield tablename, we have in importfieldshidden_array an entry 'extra.fk_object'=>'lastrowid-tableparent' so $keyfield is 'fk_object'
 								$sqlSelect = "SELECT rowid FROM ".$tablename;
 
-								if (empty($keyfield)) {
+								
 									$keyfield = 'rowid';
-								}
+								
 								$sqlSelect .= " WHERE ".$keyfield." = ".((int) $lastinsertid);
```
eldy, line 966: This looks strange.
 You do 
select rowid where rowid = xxx
to get the value xxx.

eldy:
> When defining a new import with hidden fields ​​and using a file with an existing reference.
> 
> The **`$keyfield`** variable is used in the last hidden fields loop, so that when the SQL query is executed, it always takes the last value from the hidden fields array (`import_fieldshidden_array[]`). However, this is incorrect, the WHERE clause compares with `$lastinsertid`, the query should always be: **`WHERE rowid = $lastinsertid`**

This seems to break import when there is extafields where value $keyfield is 'fk_object'
For example for this case in module modProduct.class.php

`$this->import_fieldshidden_array[$r] = array('extra.fk_object' => 'lastrowid-'.MAIN_DB_PREFIX.'product'); // aliastable.field => ('user->id' or 'lastrowid-'.tableparent)`

prietojc:
Sorry, I explained it poorly and made a mistake:
The problem appears in the update query; the select query shouldn't be touched. (I'll correct the files)

I use a file where the invoice reference is the key field to update:
```
$this->import_updatekeys_array[$r]=array('f.ref'=>'Invoice Ref');
$this->import_fieldshidden_array[$r]=array('f.libelle'=>'const-'.date('ymdHi'));
```

eldy:
Sorry i am still not able to reproduce the bug.
Can you provide the full import profile you use. My understanding is that your trouble is when doiing a custom import from your own module ?

eldy:
@prietojc
Before i validate, can you confirm me this :
After your fix, if you try to run the import in update mode for a file of thirdparties, including extrafields, do you succeed in doing the import in update mode (does the extra fields are modified correctly) ?

prietojc:
No, it doesn't work...
After several hours, I found the reason:
There is one query for main fields (lines 944-976) and another for extrafields (lines 1009-1021), but only one update query.

My change works for the main fields but not for the extrafields.

I'm making a new commit with the change, which I believe will fix both situations.  I've tested it on thirdparties and invoices.

PR #33443, opened 2025-03-13
13.0 fix migration missing tms auto update

rycks:
some surprises today for a customer, there is no data on his "last updated products" ... hum, the migration dolibarr 12 -> 13 missing all of that fields according to single sql tables

so sorry !

@eldy do i have to do the same for all versions or only for "LTS" ones ? then all people applying updates will be catched by the LTS migration one ?


eldy:
i think lts + last stable may be enough.

PR #33449, opened 2025-03-13
Fix /* TODO This does not work for number with thousand separator that is , */

Expresion:


eldy:
Can you describe the use case to reproduce the bug ?

Expresion:
Calculate selling price from purchase price or pmp with decimals plus margin over cost or margin over sale

eldy:
This does not help me to reproduce the current bug. Can you describe step by step by using the dolibarr screen how we can reproduce it ?
Which value to enter, on wich page ? Which button to click ?

Expresion:
When you enter a value it works fine. The error occurs when you select cost price or PMP with decimals from the dropdown.

To reproduce the error:
- Enter cost price in the product card.
- Add the product to quotation, order, invoice...
- Select that cost price and Margin rate or Mark rate
- Check that the selling price is not correct.

eldy:
I pushed this fix https://github.com/Dolibarr/dolibarr/commit/fa5c98824fbfdfcc3b08445a54b839528bef0676 + https://github.com/Dolibarr/dolibarr/commit/6af829873657de48c6f44535c69b830db5607f9c
Do you still have the trouble after that ?
If yes, can you give the string of the value you enter to get a wrong value after recording ?

Expresion:
It seems to be working, but new problem:
In orders and invoice, error if you edit a line and change the percentage.
In quotations, if the product has a purchase price with decimals defined, although the result of adding a percentage is correct, sale price is shown xx,00 and therefore the percentage entered also changes.

PR #33451, opened 2025-03-13
fix missing alter table for tms auto update timestamp

rycks:
Same as #33443 but for people going to 18.0


rycks:
set as draft because i have to track tables renamed between 13.0 and 18.0 ... :-(

PR #33457, opened 2025-03-14
FIX Reverse const STOCK_ALLOW_NEGATIVE_TRANSFER to STOCK_DISALLOW_NEGATIVE_TRANSFER

thomas-Ngr:
# FIX Reverse const STOCK_ALLOW_NEGATIVE_TRANSFER to STOCK_DISALLOW_NEGATIVE_TRANSFER

It is part of fix for issue #23023
Followin discussion on PR https://github.com/Dolibarr/dolibarr/pull/23050 , we want to reverse constant STOCK_ALLOW_NEGATIVE_TRANSFER. 

The reason is that :
- STOCK_DISALLOW_NEGATIVE_TRANSFER is set to 1 by default in `htdocs/core/modules/modStock.class.php`
- as a consequence, if a user sets this value to 0, the line is deleted in llx_const
- as a consequence, if the module stock is reactivated, the value is set to 1 anew : the user setting is overridden.

___

NB - same previous PR https://github.com/Dolibarr/dolibarr/pull/33444 has been merged, then reverted. Is there anything wrong ?


eldy:
PR was reverted but then the revert was reverted, so PR #33444 is included.
I have just after changed manually the label of button to we still have a positive label "Stock can be negative", event if constant has been switched into STOCK_DISALLOW_NEGATIVE_TRANSFER.  I used an option of ajax_constant_onoff calle revert for that.

PR #33478, opened 2025-03-15
FIX ci: Replace compromised tj-actions/changed-files

mdeweerd:
# FIX: Replace compromised tj-actions/changed-files

See https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised .

Use local script instead.


mdeweerd:
@eldy This fixes the CI flow (the repository for the compromised action was deleted) by adding a local script.

I added to PHP changes of an upcoming PR to verify that the script works propery.

To backport we just need to cherrypick the first commit.

EDIT: The phpstan/phan notices are unrelated to this PR.

PR #33507, opened 2025-03-18
FIX: Fix download shared document between shared entity

kkhelifa-opendsi:
Fix download shared document when for exemple
the thirdparty is shared between entity 1 and 2
and we create the thirdpaty on entity 1 but generate and share a document from entity 2

Dolibarr try to download the file on entity 2 but it's created on entity 1 (the entity where the thirdparty is created)


eldy:
The source problem is when we store the value of entity in ecmfile, it should use the same than the one of the object and currently we store the entity where we are.
This fix compensate this.

PR #33508, opened 2025-03-18
CLOSE #33506 add a constant holding the ratio to resize images in ODT…

### LINKED ISSUE #33506: ODT logo size : be able to accommodate the DPI of the source image
### Feature Request

Hello,

When using the logo within a ODT template the odf.php sets a constat for the pixel to cm ratio assuming 96 DPI.
This leads to a poor image quality in the resulting file, since printed mostly in 300DPI on printers.
Would it be possible to have a Dolibarr Const that add another value the resulting image width and height is multiplied with ?
Since the odf.php is outside of the dolibarr scope such a dolibarr constant can not be retrieved there. The setImage method is called from each doc_generic_*_odt_modules.php so it requires a lot of files to change.

### Use case

_No response_

### Suggested implementation

I'm not a professional PHP developer so i do not know if the suggested approach is good but i tried to come up with something that i could test with success :)

### Suggested steps

modify the setImage method in odf.php to have one more parameter called ratio
`public function setImage($key, $value, $ratio=1)`
Use this ratio to compute the image size
```
$width *= self::PIXEL_TO_CM * $ratio;
$height *= self::PIXEL_TO_CM * $ratio;
```
Change all setImage calls in the doc_generic_*_odt_modules.php files after getting the Dolibarr constant with the defined ratio like
```
if (getDolGlobalString('MAIN_DOC_ODT_IMAGE_RATIO')) {
    $ratio = getDolGlobalString('MAIN_DOC_ODT_IMAGE_RATIO') * 1;
} else {
    $ratio = 1;
}
$odfHandler->setImage($key, $value, $ratio);
```

Movix:
… templates

A new Dolibarr constant MAIN_DOC_ODT_IMAGE_RATIO is taken in consideration for ODT templates.
This constant holds a number that is multiplied when computing the size of images in ODT templates. All doc_generic_*_odt_modules.php files have been updated to call the setImage Method with a new parameter $ratio The method setImage in the file odt.php has been modified to have a new parameter $ratio and use it to compute the image size

### DIFF
--- a/htdocs/core/modules/asset/doc/doc_generic_asset_odt.modules.php
+++ b/htdocs/core/modules/asset/doc/doc_generic_asset_odt.modules.php
@@ -386,12 +386,19 @@ class doc_generic_asset_odt extends ModelePDFAsset
 				$parameters = array('odfHandler'=>&$odfHandler, 'file'=>$file, 'object'=>$object, 'outputlangs'=>$outputlangs, 'substitutionarray'=>&$tmparray);
 				$reshook = $hookmanager->executeHooks('ODTSubstitution', $parameters, $this, $action); // Note that $action and $object may have been modified by some hooks
 
+				// retrieve the constant to apply a ratio for image size or set the ratio to 1
+				if (getDolGlobalString('MAIN_DOC_ODT_IMAGE_RATIO')) {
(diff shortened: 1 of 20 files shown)


eldy:
As this is a new feature, thanks to push it into develop branch.

PR #33509, opened 2025-03-18
FIX: products with a single price by qty may have empty price when adding it in a commercial piece

jyhere:
How to reproduce the bug : 
- set MAIN_FEATURE_LEVEL = 2
- Activate Price by quantity in Product module settings
- activate the ajax search product feature in Product module settings
- Activate Price by quantity in the product sell tab, add only one price by quantity (e.g for quantity 1).
- Create an order. In the product search box, search for the full product reference to get only 1 result (=> no option is shown is the select2 component). 
- The price is 0.


htdocs/core/lib/ajax.lib.php
```diff
@@ -130,6 +130,14 @@ function ajax_autocompleter($selected, $htmlname, $url, $urloption = '', $minLen
 										console.log("Received answer from ajax GET, we populate array to return to the jquery autocomplete");
 										if (autoselect == 1 && data.length == 1) {
 											$("#search_'.$htmlnamejquery.'").val(item.value);
+											if (item.pbq) {
```
eldy, line 133: On line 216, we have a similar part of code but under a condition
```
if (getDolGlobalString('PRODUIT_CUSTOMER_PRICES_BY_QTY') || getDolGlobalString('PRODUIT_CUSTOMER_PRICES_BY_QTY_MULTIPRICES')) {
		$script .= '
							// For customer price when PRODUIT_CUSTOMER_PRICES_BY_QTY is on
							console.log("PRODUIT_CUSTOMER_PRICES_BY_QTY is on, so we propagate also prices by quantity into data-pbqxxx properties");
```

May be we should have the same condition for this part of code too ?

jyhere: Yes, you're right, I just replaced the JS condition with the PHP condition used in other places in the code.
Tested and OK on products with no price by qty, products with 1 price by quantity, products with many prices by qty.

PR #33511, opened 2025-03-18
Add param to disable the auto download after auto convert odt to pdf

kkhelifa-opendsi:
Add globale MAIN_DISABLE_PDF_AUTO_DOWNLOAD_AFTER_CONVERT_FROM_ODT for disable the auto download after auto convert odt to pdf


eldy:
I wasn't aware that there was an autodownload.
i think we should never have this.

May be we can add a parameter into exportAsAttachedPDF to it is the caller that decide if there is a download or not.

PR #33512, opened 2025-03-18
FIX: Always show message if accountancy hidden options are activated to help to resolve some problems

kkhelifa-opendsi:
Always show message if accountancy hidden options are activated to help to resolve some problems


aspangaro (approved):
Ok for me. More information to solve problems is always nice

PR #33514, opened 2025-03-18
FIX: Set to empty accountancy code on company table if MAIN_COMPANY_PERENTITY_SHARED is activated (when update company)

kkhelifa-opendsi:
Set to empty accountancy code on company table if MAIN_COMPANY_PERENTITY_SHARED is activated (when update company)


aspangaro:
Hello @kkhelifa-opendsi,  

Why force third-party sales/purchases accounting accounts (level 3) in the llx_societe table to 'empty' if the llx_societe_perentity sub-table is used, which will cause problems later on for links? It favors the llx_societe table even if the MAIN_COMPANY_PERENTITY_SHARED constant is activated?

kkhelifa-opendsi:
hello @aspangaro 
i don't known, it's fix you have commit the May 12, 2023 on easya 2022.5.3 who have not been put on develop branch

eldy:
Don't see a reason to purge data in the common columns if we decide to manage per entity (at least, the common column should not be used, but if can be retreive if we switch option). So i close PR.

PR #33517, opened 2025-03-18
Develop fix STOCK_MUST_BE_ENOUGH_FOR_(PROPAL|ORDER|INVOICE)

thomas-Ngr:
# Fix Constants STOCK_MUST_BE_ENOUGH_FOR_(PROPAL|ORDER|INVOICE) would not work as expected for products not managed in stock

On products where `$product->stockable_product == 0`, we would expect that STOCK_MUST_BE_ENOUGH_FOR_XXX don't have any effect.

However, we would get errors when adding a line on propal, order, invoices, even if stock management is not enabled on a product, if STOCK_MUST_BE_ENOUGH_FOR_(PROPAL|ORDER|INVOICE) is set.

This PR fixes this behavior.


htdocs/comm/propal/class/propal.class.php
```diff
@@ -694,7 +694,7 @@ public function addline($desc, $pu_ht, $qty, $txtva, $txlocaltax1 = 0.0, $txloca
 				$result = $product->fetch($fk_product);
 				$product_type = $product->type;
 
-				if (getDolGlobalString('STOCK_MUST_BE_ENOUGH_FOR_PROPOSAL') && $product_type == 0 && $product->stock_reel < $qty) {
+				if ($product->stockable_product && $product->stock_reel < $qty && getDolGlobalString('STOCK_MUST_BE_ENOUGH_FOR_PROPOSAL') && ($product_type == Product::TYPE_PRODUCT || ($product_type == Product::TYPE_SERVICE && getDolGlobalString('STOCK_SUPPORTS_SERVICES')))) {
```
eldy, line 697: I suggest to replace with
`if ($product->isStockManaged() && $product->stock_reel < $qty && getDolGlobalString('STOCK_MUST_BE_ENOUGH_FOR_PROPOSAL')) {`

thomas-Ngr: You are right, I changed the PR this way.

thomas-Ngr:
@eldy do you have another question ?

PR #33519, opened 2025-03-18
NEW const PREVIEW_PICTO_ON_LEFT_OF_NAME for icon to preview documents

lvessiller-opendsi:
NEW const PREVIEW_PICTO_ON_LEFT_OF_NAME for icon to preview documents
- so the preview icon can be on left of document name : 
image


eldy:
What is goal compare to this one ?
On left, but why having on left instead of on right ?
image

aspangaro:
In essence, yes, that's the only difference. It's an old legacy from Easya 2020 (Dolibarr v10 LTS) when the search icon could be hidden if the file name was too long. Customers have gotten used to using it.

lvessiller-opendsi:
> What is goal compare to this one ? On left, but why having on left instead of on right ? image

It's just a setting to choose between having the preview icon on left (If PREVIEW_ON_LEFT is set) or on right (by default).
It's similar as const "MAIN_CHECKBOX_LEFT_COLUMN".
Maybe we can overload CSS but I don't know why.

eldy:
Can you rename into 
PREVIEW_PICTO_ON_LEFT_OF_NAME ?

lvessiller-opendsi:
> Can you rename into PREVIEW_PICTO_ON_LEFT_OF_NAME ?

Ok the const has been renamed

eldy:
It seems the problem to update or edit a PR is due to the fact the fork was done into an organization repo. Problem seems not present when using a personal repo.
Do you know if there is particular option on your organization repo that lock PR for update ?

lvessiller-opendsi:
> It seems the problem to update or edit a PR is due to the fact the fork was done into an organization repo. Problem seems not present when using a personal repo. Do you know if there is particular option on your organization repo that lock PR for update ?

@eldy
I don't know why.
I will see with the admin sys.
Maybe there is a box to check before to create PR.
What do you want to modify in this PR ? Or it was to update from develop ?

PR #33520, opened 2025-03-18
NEW : Add globale ACCOUNTANCY_SELL_JOURNAL_FIX_PRECISION_AND_GAP for fix amount precision and gap on accountancy sell journal page

kkhelifa-opendsi:
Add globale ACCOUNTANCY_SELL_JOURNAL_FIX_PRECISION_AND_GAP for fix amount precision and gap on accountancy sell journal page


eldy:
It is not clear of what are king of price fix done by this feature.
Can you provide examples with value ?

eldy:

Issue #33378, opened 2025-03-09
Add open protocol based social networks


strk:
### Feature Request

It would be nice to be able to add Mastodon / Pixelfed / Mobilizon / Peertube etc. accounts in the "Social Networks" page for a Company/Organization, rather than limiting that to the usual big players on the market.

![Image](https://github.com/user-attachments/assets/57b73fa5-4a45-488e-a359-1ed8769472fb)

frederic34:
it's in a dictionary, check for socialnetworks module

strk:
I would also like to be able to remove the references to proprietary platforms. I don't want to push my customers to those places...

PR #33525, opened 2025-03-19
Add mastodon icon and pixelfed record in social networks

strk:
See GH-33378


sonikf:
Ηι @strk
You have to push db changes also to https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/install/mysql/migration/21.0.0-22.0.0.sql for those who upgrade

strk:
> You have to push db changes also to https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/install/mysql/migration/21.0.0-22.0.0.sql for those who upgrade

Uhm... are these per-database ?
My need is to use Dolibarr with a PostgreSQL backend, is that not tested by CI ?

sonikf:
It affects both databases, nothing to do with CI, leaving it the way it is changes will be visible only to new installations.

strk:
@eldy isn't the Mastodon record also missing from migrations ?

eldy:
> @eldy isn't the Mastodon record also missing from migrations ?

I added the insert by may be ii shloud have add an update.

PR #33526, opened 2025-03-19
NEW: Add global STOCK_MOVEMENT_FORCE_DO_NOT_CLEAN_EMPTY_LINES to force do not clean empty lines on movement stock

kkhelifa-opendsi:
Add global STOCK_MOVEMENT_FORCE_DO_NOT_CLEAN_EMPTY_LINES to force do not clean empty lines on movement stock

Because customer can need to see the warehouse when the product previously have stock and now has stock at 0.
Customer case when manage a lot of warehouses and keep specifique warehouses for a product, don't have to remember what warehouses is attibuted to the product.


eldy:

PR #33539, opened 2025-03-19
FIX Sort and search Ref Project column was missing

mapiolca:
This list was the only one where it didn't have a column for projects ref sort and search.


mapiolca (commented):
Fix white space

PR #33543, opened 2025-03-20
NEW #33506 Add a constant to apply a ratio to resize images in ODT templates

Movix:
Added a new Dolibarr constant MAIN_DOC_ODT_IMAGE_RATIO.
This constant holds a number that is multiplied when computing the size of images in ODT templates.
All doc_generic_*_odt_modules.php files have been updated to call the setImage Method with a new parameter $ratio.
The method setImage in the file odt.php and segment.php has been modified to have a new parameter $ratio and use it to compute the image size

### DIFF
--- a/htdocs/includes/odtphp/odf.php
+++ b/htdocs/includes/odtphp/odf.php
@@ -530,7 +530,7 @@ class Odf
 	 * @throws OdfException
 	 * @return odf
 	 */
-	public function setImage($key, $value, $ratio=1)
+	public function setImage($key, $value, float $ratio=1)
 	{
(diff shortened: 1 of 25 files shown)


htdocs/includes/odtphp/odf.php
```diff
@@ -530,7 +530,7 @@ public function preOdfToOdf($value)
 	 * @throws OdfException
 	 * @return odf
 	 */
-	public function setImage($key, $value)
+	public function setImage($key, $value, float $ratio=1)
```
eldy, line 533: You must add the parameter $ratio into the doc comment.
`@param float $ratio   Ratio for image`

PR #33544, opened 2025-03-20
FIX: Fix the construct of SQL request when key field parent for dependent list is provided on extrafields and object fields

kkhelifa-opendsi:
Fix the construct of SQL request when key field parent for dependent list is provided on extrafields and object fields


eldy:
The use of this feature was always obscur to me. This is a good occasion to ask you if you can provide an example.
Which value do you use for options of chckboxlist ?

eldy:
Fixed by #37706 ?

PR #33547, opened 2025-03-20
FIX: The select value is not only integer value so we accept string value for index into ajax_combobox() function

kkhelifa-opendsi:
The select value is not only integer value so we accept string value for index into ajax_combobox() function

Example: If the empty value index is empty and we have a option with value index at 0 with a label, the label is set to empty

### DIFF
--- a/htdocs/core/lib/ajax.lib.php
+++ b/htdocs/core/lib/ajax.lib.php
@@ -517,7 +517,7 @@ function ajax_combobox($htmlname, $events = array(), $minLengthToAutocomplete = 
 	 					/* Code to add class of origin OPTION propagated to the new select2 <li> tag */
 						if (data.element) { $(container).addClass($(data.element).attr("class")); }
 						//console.log("data html is "+$(data.element).attr("data-html"));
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/ajax.lib.php
```diff
@@ -517,7 +517,7 @@ function ajax_combobox($htmlname, $events = array(), $minLengthToAutocomplete =
 	 					/* Code to add class of origin OPTION propagated to the new select2 <li> tag */
 						if (data.element) { $(container).addClass($(data.element).attr("class")); }
 						//console.log("data html is "+$(data.element).attr("data-html"));
-						if (data.id == '.((int) $idforemptyvalue).' && $(data.element).attr("data-html") == undefined) {
+						if (data.id == "'.dol_escape_js($idforemptyvalue, 2).'" && $(data.element).attr("data-html") == undefined) {
```
eldy, line 520: The char to use for js string must be ' (this is the char escaped by dol_escape_js).
To should be
`if (data.id == \''.dol_escape_js($idforemptyvalue, 2).'\' && $(data.element).attr("data-html") == undefined) {`
Can you test this ?

PR #33550, opened 2025-03-20
Fix : MAIN_SEARCH_PRODUCT_BY_FOURN_REF search critera

thersane-john:
# Fix

When use MAIN_SEARCH_PRODUCT_BY_FOURN_REF conf search not use correct operator

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -3089,7 +3089,7 @@ class Form
 
 				// include search in supplier ref
 				if (getDolGlobalString('MAIN_SEARCH_PRODUCT_BY_FOURN_REF')) {
-					$sqlSupplierSearch .= !empty($sqlSupplierSearch) ? ' OR ':'';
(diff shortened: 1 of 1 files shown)


htdocs/core/class/html.form.class.php
```diff
@@ -3089,7 +3089,7 @@ public function select_produits_list($selected = 0, $htmlname = 'productid', $fi
 
 				// include search in supplier ref
 				if (getDolGlobalString('MAIN_SEARCH_PRODUCT_BY_FOURN_REF')) {
-					$sqlSupplierSearch .= !empty($sqlSupplierSearch) ? ' OR ':'';
+					$sqlSupplierSearch .= !empty($sqlSupplierSearch) ? ' AND ':'';
```
eldy, line 3092: Use of AND here looks strange.
If you search string "abc" you wan to get the line if ab is inside customer ref OR inside supplier ref.
Also having a AND will generate a incorrect condition 
a OR b OR c OR d AND ref_supplier
what is priority between OR and AND ?

PR #33565, opened 2025-03-21
FIX: Incorrect entity used when downloading a document generated in another entity

atm-quentin:
# FIX Incorrect entity used when downloading a document generated in another entity
In Dolibarr version 21, an inconsistency in entity management prevents downloading documents like proposals (propales), while this works correctly in version 20.
Identified Issue:

    The function dol_dir_list_in_database uses the current entity ($conf->entity) in its SQL query to retrieve a file.
    If the file was generated under a different entity than the active one, it is not found.
    As a result, since $found remains 0, a new ECM entry is created with the current entity, causing an inconsistency: the file appears to be stored under the wrong entity.

Proposed Solution:

    Modify dol_dir_list_in_database to consider the source object's entity instead of the current entity.
    Add an entity parameter to allow passing the object's entity instead of the default ($conf->entity).

This fix ensures that documents are correctly retrieved from the appropriate entity and prevents erroneous ECM entries from being created.


htdocs/core/class/commonobject.class.php
```diff
@@ -9797,7 +9797,7 @@ public function show_photos($modulepart, $sdir, $size = 0, $nbmax = 0, $nbbyrow
 		 $filearray=array_merge($filearray, $filearrayold);
 		 }*/
 
-		completeFileArrayWithDatabaseInfo($filearray, $relativedir);
```
eldy, line 9800: Why not adding $this as parameter (so we provide all the object). We wil be able to retreive $objectfordbinfo->entity and we will have all other data of object for future need.

atm-quentin: i did the modification, is it ok for you ?

PR #33598, opened 2025-03-24
NEW : add new permissions to read supplier prices

altairis-tof:
Some companies dont want some collaborators to read supplier prices but to be allowed to see customer prices.
@eldy do you think we replace the current read_prices permission by a read_supplier_prices and a read_customer_prices ?
or do we only add this new read_supplier_prices permission ?


eldy:
If you add read_supplier_prices, it will be better to have read_prices for customer price only.
But this can be done in a second steps if you prefer

PR #33599, opened 2025-03-24
NEW input reason for customer invoice

lvessiller-opendsi:
NEW input reason for customer invoice
- see SQL implementation #33569

Add an input reason in customer invoice
- this input reason can be propagated from customer proposal and order
- this input reason is set to origin input reason when the customer invoice is created from an other card (proposal, order or shipment)
- the input reason is the same as customer order when it's created from mass action of order list : "createbills"


lvessiller-opendsi:
@eldy 
I fixed all CI errors and updated from develop branch.
All others CI errors seem not concern this pull request.

PR #33602, opened 2025-03-25
FIX: we must retrieve linked order_supplier and no other object

altairis-noe:
# FIX we must retrieve linked order_supplier and no other object
If we want to link objects different than a CommandeFournisseur to a Reception object, we get an ugly error, because fetch->origin will try to fetch… well, not the actual origin!
image

With this correction: here we go!
image

### DIFF
--- a/htdocs/reception/class/reception.class.php
+++ b/htdocs/reception/class/reception.class.php
@@ -378,7 +378,7 @@ class Reception extends CommonObject
 		$sql .= ', e.fk_incoterms, e.location_incoterms';
 		$sql .= ', i.libelle as label_incoterms';
 		$sql .= " FROM ".MAIN_DB_PREFIX."reception as e";
-		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."element_element as el ON el.fk_target = e.rowid AND el.targettype = '".$this->db->escape($this->element)."' AND el.sourcetype = order_supplier";
+		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."element_element as el ON el.fk_target = e.rowid AND el.targettype = '".$this->db->escape($this->element)."' AND el.sourcetype = 'order_supplier'";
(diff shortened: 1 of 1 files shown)


altairis-noe:
Hey there, @eldy (or any Travis CI expert). I fail to understand why Travis can’t build. The error states that when fetching a Reception object with id = 1, it was not found (return of the fetch method = 0), which seems fine to me, as I don’t know that there should be such a reception in db.

Is that related to my modification, and in which way ?

altairis-noe:
Maybe @frederic34 would know why Travis is so mean to me !

frederic34:
@altairis-noe on your example can you remove link to order_supplier and see what happen with your change

altairis-noe:
Good point ! As Reception are created from a Purchase order, I would not have thought about what would happen in case they don’t have an originating one… Thx @frederic34

htdocs/reception/class/reception.class.php
```diff
@@ -378,7 +378,7 @@ public function fetch($id, $ref = '', $ref_ext = '')
 		$sql .= ', e.fk_incoterms, e.location_incoterms';
 		$sql .= ', i.libelle as label_incoterms';
 		$sql .= " FROM ".MAIN_DB_PREFIX."reception as e";
-		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."element_element as el ON el.fk_target = e.rowid AND el.targettype = '".$this->db->escape($this->element)."'";
+		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."element_element as el ON el.fk_target = e.rowid AND el.targettype = '".$this->db->escape($this->element)."' AND el.sourcetype = order_supplier";
```
frederic34, line 381: `= 'order_supplier'"` ?

altairis-noe: Pardon my french but : “M’enfin!”
I cleaned my glasses, it should go better 😒

PR #33603, opened 2025-03-25
Fix removed unnecessary tests

YannisHoareau:


alexandre-janniaux:
Why just commenting it instead of really removing it?

YannisHoareau:
@alexandre-janniaux I could have done that but I rather follow what is already done and when looking at the code, these tests were commented. I only added missing ones.

Issue #33575, opened 2025-03-22
Add Doc of how create Doc in UXUI doc


john-botella:
### Feature Request

Hi, 

This is an issue for Giff UX UI project

in admin/tools/ui/

we need a new section fo how use Doc add create menu, item etc...

lmag:
Add in Dolibarr 22
/admin/tools/ui/content/

![Image](https://github.com/user-attachments/assets/433cc21f-ef2e-4f5b-8860-c7d3bc49fb0e)

Sample for table

![Image](https://github.com/user-attachments/assets/d60bbf68-445a-4b51-b67a-bc6c1fc17ac5)

PR #33604, opened 2025-03-25
UIUX: Update documentation interface

EchoLoGeek:
- Update documentation interface
- Add contributing page ( See #33575 )
- Same code for each group index file


EchoLoGeek:
Pre commit says that several lines are indented incorrectly, but they are indented to match the html.
Any idea how to fix this while keeping the indentation?

eldy:
precommit si rright.
Some lines are indented with spaces instead of tab. So indentation will be wrong depending on setup of user.
Just replace space by tabs to fix this.

EchoLoGeek:
> precommit si rright. Some lines are indented with spaces instead of tab. So indentation will be wrong depending on setup of user. Just replace space by tabs to fix this.

I can't find indentation by spaces in code. All indentation are converted to tabs.
Due to maintaining HTML indentation, there are two levels of indentation for PHP code. I'll remove that.
image

PR #33605, opened 2025-03-25
FIX|Fix # SQL error when filtering by draft status in list

atm-corentin:
When applying a filter on the draft status in a list view, an SQL error occurred due to incorrect handling of the filter logic. The query builder did not account for specific edge cases where the draft status caused the generated SQL to become invalid or malformed

The filter attempted to include or exclude draft statuses without properly handling how these statuses interact with other conditions in the SQL WHERE clause. This led to broken SQL syntax or unexpected behavior in the results

This fix adjusts the filter logic to safely handle the draft status when building the SQL query. It ensures that the query remains valid and that draft-related filters do not break the request. The logic now checks for the presence of draft status more explicitly and avoids injecting invalid conditions into the query


eldy:
Can you provide an example of the input value that make the result wrong.
This is a typical case we must add into the php unit test.

PR #33606, opened 2025-03-25
Unset fk_user_cloture upon opening/reopening a contract line

atm-florianm:
# Rationale
Currently, `fk_user_cloture` is still set even when a line has been reopened. This can be counter-intuitive and cause people to believe that the line is still closed.

One argument against unsetting `fk_user_cloture` could be that this information can be useful (after reopening a line, you may want to check who closed it last), but if so, why not keep `date_cloture` as well for consistency?

I think agenda events would be a cleaner way to keep track of when and by whom a line was closed / opened etc.


atm-florianm:
I pulled develop into my branch (I can't unlock it because it is owned by ATM-Consulting).

PR #33622, opened 2025-03-26
FIX v22 warnings and remove seemingly useless code

atm-florianm:
I came across these warnings while testing for #33616. I believe the lines I deleted were never going to be used since in this context $line doesn't have an `element` property and, additionally, we already know we're dealing with contract lines and not lines of another object type.


eldy:
Your branch is still locked.
Did you check the checkbox "allow edit" when you make your PR ?

atm-florianm:
> Your branch is still locked. Did you check the checkbox "allow edit" when you make your PR ?

I didn't see the checkbox. I just tried creating another dummy PR to see if I can find the checkbox, but it wasn't there.

After browsing the [doc](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), it looks like it is available for user-owned forks only, but not for organization-owned forks like ATM-Consulting:Dolibarr.

eldy:
What is strange is that i do not experience trouble of PR locked for update with some other ATM users.
For exemple with atm-irvine, #3325
image

And with yours
image

As you can see i don't have the "Update" button required to fix myslef CI troubles or other on your PR but i have iton atm-irvine. Can you investigate if the fork is done the same way ?
It seems irvine is doing the fork in its dir, and you into dir "ATM-Consulting". How do you do your initial fork ? Both users are users in same organization i suppose. So do you know where come the difference ?

atm-florianm:
> It seems Irvine is doing the fork in its dir, and you into dir "ATM-Consulting". How do you do your initial fork ? Both users are users in same organization i suppose. So do you know where come the difference ?

This is precisely what you pointed out: Irvine pushes to his own fork (as do all students at ATM) while employees push their branches to the organization's fork (because it is more convenient for us, especially if someone has to take over someone else's work). We could change that policy of course, nothing is set in stone. I've already brought up the subject with my colleagues this morning.

If I understand correctly, the « Update branch » button allows you to merge the target branch into the source branch to make sure they are in sync? Do you require all source branches to have that option? If not, we could also sync our branches manually.

eldy:
Being able to make the sync myself is not the only difference.
The PR is also editable or not editable (See github menu i have on your PR)
image

When branch is locked for edit, it means also there is no way to solve a conflit (for me but any other contributor, it is more and more frequent that contributors correct or enhance a PR when it is in status "PR waiting for merging" mode). It also means it is not possible to fix a CI error that apears in PR, forcing us to ask you to make the change yourself, one feedback by one (when the merger or any guy working on the CI can do it themself immediatly with non need of round tripwith the other mode).
This increase highly delay of merging and amount of work (mostly on your side, but also the side of the merger that must frequently make feed back on the PR to have it repushed with last version of the branch, and reanalyze later once CI is green.
It looks strange that an organization branch has less capabilities than a personal branch. May be there is an option somewhere in the organisation setup to restore the same behavior than for personnal fork ?
1 or 2 years ago, this problem was never meet, so it may be a change in default github policy for security reason.

for this particularPR, can you do the merge with version of the develop branch.

PR #33647, opened 2025-03-28
NEW Can link member subscription/contribution with proposal, order and invoice

JonBendtsen:
# NEW [*linking member subscription contribution <-> proposal, order and invoice*]
This PR allows linking member subscription contribution to/from proposals, orders and invoices.

This can both be done manually from an existing proposal/order/invoice to a subscription created without creating an invoice. I use this when we sell memberships as part of registration for our membership classes.

The link will also be shown on the invoice that can be created as part of the subscription contribution creation.

These links show up in the json from the API on proposals, orders and invoices. The subscription does not show anything. It is currently unknown if the API can do this link or it will have to be done manually in the GUI.


JonBendtsen:
If I hand link from proposal, order or invoice to a subscription created by hand without that subscription creating an invoice it looks like this:

handlinking_proposal_start
handlinking_proposal_to_list_of_subscriptions
handmade_invoice_linked_to_subscription
handmade_order_linked_to_subscription
handmade_propal_linked_to_subscription

JonBendtsen:
On the subscription above the links looks like

contribution_links

JonBendtsen:
If i add a subscription and create an invoice, it will look like this from the invoice

subscription_created_invoice

JonBendtsen:
in the API, the jsons for subscriptions has no changes.

However the proposals, orders and invoices does show the link in the json

htdocs/core/class/html.form.class.php
JonBendtsen, line 1: @eldy Do you want changes like this done every place? is it safe enough just to do a search'n'replace?

eldy: If you are speaking about replacing
isModEnabled('adherent'),
into
isModEnabled('member'),

Yes replacement is safe, but there is no need. It was already done.Now the CIrefuse code that use 
isModEnabled('adherent'),

eldy:
Can you pus your PR in develop ?.
I tried to switch it but got conflicts...

PR #33655, opened 2025-03-29
UIUX : Giff UX Doc add experiment Tooltip freeze

thersane-john:
# UIUX
Add UX experimental section in UX Doc
This section gathers UX experiments that are not yet integrated into Dolibarr but are available only on these dedicated pages. They allow testing different approaches before a possible integration into the develop branch.

To make a demonstration of usage of experimental section I add also an expériment UX feature : Tooltip freeze

image

image


EchoLoGeek (approved):
Great work !

PR #33680, opened 2025-04-01
shipping can manage its contacts

frederic34:


htdocs/core/lib/sendings.lib.php
```diff
@@ -71,7 +72,7 @@ function shipping_prepare_head($object)
 
 	if (!getDolGlobalString('MAIN_DISABLE_CONTACTS_TAB')) {
 		$objectsrc = $object;
-		if ($object->origin == 'commande' && $object->origin_id > 0) {
+		if (getDolGlobalInt('SHIPPING_USE_ORDER_CONTACTS') && $object->origin_type == 'commande' && $object->origin_id > 0) {
```
eldy, line 75: Old code was in mode:  "shipping is using the contact of parent object (order, propal;;.)"
So to have a transition step toward "shipping is using its onw contact", i suggest to rename the variable
SHIPPING_USE_ORDER_CONTACTS into
SHIPPING_USE_ITS_OWN_CONTACTS
and invert the test.

PR #33685, opened 2025-04-01
Add quote arround value in extra field sql where

atm-etienne:
# Fix - Set extra field search value as string in dolForgeCriteriaCallback
Between 20.0.2 and 20.0.5 a new bug appears on the extra fields. String value for direct search are not parse as string, so SQL throws an error.


eldy:

PR #33687, opened 2025-04-01
NEW(project, admin): Display or discard closed projecs in select list

BenjaminFlr:
Allow user to define the value of PROJECT_ALWAYS_DISCARD_CLOSED_PROJECTS_IN_SELECT directly in the project admin interface.
Capture d’écran 2025-04-01 à 15 19 14


eldy:
what is the rule if the option is not checked ? The cosed product are always shown ?
i think it depends on screen ?

BenjaminFlr:
For the tests I made in various objects' card, if not checked, closed project are not displayed.
If checked, closed projects are displayed.

In fact it's an existing constant since v20 if I remember correctly. I juste moved it to the admin interface.

eldy:
> For the tests I made in various objects' card, if not checked, closed project are not displayed. If checked, closed projects are displayed.
> 
> In fact it's an existing constant since v20 if I remember correctly. I juste moved it to the admin interface.

The constant was introduced to do "hide always" when before we were doing "show always".
But since, we should have introduced a more clever way of working. Now it shows closed project only "when required" so option should become useless.
On which page for example do you have a not optimal behaviour shoing project when it should not ?

BenjaminFlr:
> On which page for example do you have a not optimal behaviour shoing project when it should not ?

I just tried it when creating an invoice. If I want to link it to a closed project, I have to enable this option.

Also, on some cards (tickets for example), and it's something I will open an issue for and work on it, project select list are not built the same way when action = create than when action = edit.

eldy:
> I just tried it when creating an invoice. If I want to link it to a closed project, I have to enable this option.

Yes this is the expected behaviour. A close project menas project should not be linked to any invoice. In this situation, user must reopen the project. This is the definition of a closed project.
Yes, on create, the list is only open project.
But on edit mode, the list is (i mean should be) open projects + the current assigned project. So when we edit the project, we can keep old closed project.

BenjaminFlr:
So it adds few steps for something that is quite common in a company daily workflow.

In fact, I just wanted to make this possibility easier to find, but it's not something we need to lose time on. If it's not pertinent, we can just close the PR ;)

eldy:
> So it adds few steps for something that is quite common in a company daily workflow.
> 
> In fact, I just wanted to make this possibility easier to find, but it's not something we need to lose time on. If it's not pertinent, we can just close the PR ;)

It may be pertinent, but i must be sure to understand because in past we receive a lot of complain because project was not visible from list and project list was too long. This is why we have now a mode "it depends on the situation".
If the few step added is to reopen the proejct, yes it adds. But this is what we want. Closing aproject means we don't want to use it anymore WITHOUT an explicit step. So it looks to have a sens.

PR #33690, opened 2025-04-01
NEW : Rounding unit price - line discount before applying quantity - option MAIN_APPLY_DISCOUNT_ON_UNIT_PRICE_THEN_ROUND_BEFORE_MULTIPLICATION_BY_QTY

omogenot:
On some online store applications, the line total price is calculated based on the discount applied on the unit price first, giving a new rounded unit price, and then multiplied by the quantity. Whereas the current way Dolibarr acts is to apply the discount on the subtotal (unit price * quantity). This could lead to discrepancy between what the customer has seen on the merchant web site and what Dolibarr finally issues as an invoice.
Using an extra OTHER SETUP constant 'MAIN_ROUND_UNIT_PRICE_BEFORE_QTY', we can choose between the two ways. Leaving this option to 0 makes Dolibarr act as usual. A value of 1 make Dolibarr round discounted unit price first.


eldy:
I renamed constant.
Also i replaced the 
round(... , 2)  with
price2num(.. ,   constant)
So you can decide on how many decimal you round by setting the constant to the nb of decimal instead of 1, so you can match the rule of the other system even it if round on 5 decimals for example.

PR #33699, opened 2025-04-02
FIX: sql error when upgrading to 22 if session table is already in use

jyhere:
If upgrading Dolibarr from a previous version, and sessions are already stored in db, an error will occurs : 

image


eldy:
I try a different fix because we must avoid the use of CURRENT_TIMESTAMP that give date that are wrong because not using the correct timezone (using the TZ of database instead of the PHP server and it may differ).

PR #33701, opened 2025-04-02
NEW #32706 - Treasury accountancy

lvessiller-opendsi:
NEW #32706 - Treasury accountancy


aspangaro (approved):
Already seems ok for me. Just resolve CI problem

lvessiller-opendsi:
A new option "ACCOUNTING_USE_TREASURY" is defined in main accounting setup : 
image
- now the main options use the new items created with FormSetup object

When this option is enabled, the left menu of accounting is like that : 
image
- there is only journal whose nature is bank

When we click on "Treasury journal" link of this above menu, we go to this ventilation page : 
image

And then you can click on the save button to insert theses lines in the bookkeeping.

lvessiller-opendsi:
@eldy 
I've solved every CI errors I could.
The last ones are on unchanged files and seem to not concern this pull request.
Could you have a check ?

eldy:
My understanding is that the new option
ACCOUNTING_USE_TREASURY
is to make "comptabilité de trésorerie" instead of "comptabilité d'engagement" (i don't know the name in english).

But we already have a variable for that. It is called ACCOUNTING_MODE. When value is "RECETTES-DEPENSES", we are in the mode "comptabilité de trésorerie".
I will make a PR to merge the two vars...

PR #33706, opened 2025-04-03
FIX $object not initialised in extrafields_list_print

Vaadasch:
When using a computed int extrafield in factures and propals, i've came accross this bug.

`$object->id == 0` inside `extrafields_list_print_fields.tpl.php`, when it is called from **propal/list.php**

So i've done a quick and clean fix where i was able to : i didn't found where the $object was fetch inside the facture/list.php to reproduce it on the propal/list.php. But my fix is a good workaround : it only fetch anew $object if its "id" is different from the $obj->rowid. So if $object was correctly set it (and maybe have some modifications) it doesn't change it.
If this bug appear on other list.php, it should work too.

To test it, simply create an Int extrafield with computed value of `$object->id` :
Setup_test_33706

# Before
Propal list (bug) : 
Before_propal_33706
Factures list (working) :
Before_factures_33706

# After 
Propal list (working) : 
After_propal_33706
Factures list (working) :
After_factures_33706

## My case (not important for the PR, i've managed to enlight the bug more simply) :
If you want to test it on your side the before/after : my extrafield is "How many emails had been sent" : 
For propals: 
```( ($resql = $db->query(implode(' ', array("SELECT id FROM llx_actioncomm WHERE code = 'AC_PROPAL_SENTBYMAIL' AND fk_element = ", $object->id))) && $nb = $db->num_rows($resql)) ? $nb : $nb)```
For factures:
```( ($resql = $db->query(implode(' ', array("SELECT id FROM llx_actioncomm WHERE code = 'AC_BILL_SENTBYMAIL' AND fk_element = ", $object->id))) && $nb = $db->num_rows($resql)) ? $nb : $nb)```
Beware of the removing of the `"` characters when creating the extrafield, need modification to write them again. So create then modify.


Vaadasch:
It seems that i'm not able to find the php-stan check on the number of `Variable \$obj might not be defined` to change it from 3 to 5 to match my fix.

Vaadasch:
I believed that a warning log was in caused by these, but in fact, no

josett225:
Hi @Vaadasch 
Thanks to bring this workaround. It works fine.
However I believe we have to find where the issue is coming from as I am not sure this PR will be validated as it is.

Thanks anyway.

Vaadasch:
@josett225 
Thanks !
Do you have this warning too ?
Do you have it because of this PR or did you have it before ?

josett225:
Hi @Vaadasch 
The main difference between Invoice (=facture) is the following line in list.php:
$object = $facturestatic; 

The issue I am facing is when adding the same following line in propal list.php, it does not work. 
I think there are some fields missing after fetching $obj and assigning some values to $objectstatic.
$object = $objectstatic;
For now there is still the need of adding:
$object->fetch($objectstatic->id);

No need to change the extrafields_list_print file

Still investigating...

eldy:
According to documentation into the tooltip of the computed formula, we must use 
$objectoffield
inside formula to get property of an object.

"... You can enter here a formula using other properties of object or any PHP coding to get a dynamic computed value. You can use any PHP compatible formulas including the "?" condition operator, and following global object: $db, $conf, $langs, $mysoc, $user, $objectoffield.
WARNING: If you need properties of an object not loaded, just fetch yourself the object into your formula like in the second example. ....

((($reloadedobj = new Societe($db)) && $reloadedobj->fetchNoCompute($objectoffield->id) > 0) ? $reloadedobj->array_options['options_extrafieldkey'] * $reloadedobj->capital / 5: '-1')
"

Can you try with this ?

Vaadasch:
@eldy 
$objectoffield is naturally populated with $object in : 
https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/core/tpl/extrafields_list_print_fields.tpl.php#L72
`$objectoffield = $object; `

This bug is only in propal

I've managed to make a develop environment on my side (was only testing and contributing on v20)
I'll dig in.


I confirm that $objectoffield->id doen't work too, and the bug is present in the develop branch .

image

image

josett225:
Hi @eldy and @Vaadasch 
The issue is not in using the $object or $objectoffield.

Vaadasch:
I don't even understand the need to have a static object that is an instance of Facture or Propal and have all its needed properties set from $obj. Is that not the purpose of fetch ?
It seems to me that, as it has all the needed properties, executing a method of $objectstatic will have exactly the same results as doing it on a fetched $object->fetch($id).
So it does not seems to prevent a unwilling modification

For me, we could (mentally) remove every static object and replace it by $object->fetch($id).
( Edit : no, for performances, we should prefer keeping only the needed fields)

Hint : 
$obj is the row extracted from $db->fetch_object($resql);


I confirm that there is no reference of $objectstatic in none of the include or require_once files of propal/list.php

Vaadasch:
$objectoffield is never populated through CommonObject::fetch_optionals() in facture/list.php nor propal/list.php
That's why the $objectoffield = $object was present in the extrafields_list_print_fields.php that i've mentioned.

Vaadasch:
@eldy , @josett225 
See #33753 for my suggestion.

PR #33710, opened 2025-04-03
New add sumarize function to mail Ai Enhancements

Hystepik:
Add sumarize function to mail Ai Enhancements

Add the possibility to sumarize in 20, 50, 100 or 200 words and 1 or 2 paragraphs
image

After AI generation the response fill the mail message field


eldy:
Can you also add the choice 25%, 50% and 75% ?

PR #33711, opened 2025-04-03
fix(evaluation): on list, removed getNomUrl for label column

atm-irvine:
# FIX|Fix removed getNomUrl for label column
I've removed the getNomUrl for the label field as there is already an existing field that need to be seen.


eldy:
Can you provide a screenshot that show the trouble or change ?

atm-irvine:
@eldy Sure.

Before : 
image

After:
image

PR #33737, opened 2025-04-05
NEW: classmap autoloading

marc-dll:
# NEW: classmap autoloading

Dolibarr doesn't provide autoloading mechanisms. This means that in order to use a `class`, `interface`, `trait`, or `enum`, we need to `include` or `require` the source.

This can be error-prone, as sometimes badly tested code where such inclusion has been forgotten can get merged (I know for a fact that it can be done, as I already did it myself some years ago!). And it's without saying that it needlessly increases the codebase size.

The *de facto* standard for autoloading in current PHP state of the art is [PSR-4 autoloading](https://www.php-fig.org/psr/psr-4/), but as it is based on a filesystem hierarchy, implementing it in Dolibarr would require a major refactoring. For the same reason, [PSR-0 autloading](https://www.php-fig.org/psr/psr-0/) cannot be considered, and is deprecated since 2021 anyway.

Therefore, I propose a classmap based mechanism and a script to generate such classmap.

The new mechanism happens inside `filefunc.inc.php` and is therefore available near the very beginning of any page or script. It loads a new `classmap.inc.php` file, and leverages standard PHP function [`spl_autoload_register()`](https://www.php.net/manual/fr/function.spl-autoload-register.php), so that the callback provided to it is called when a `class` (or equivalent) is not loaded.

The generator script analyzes all PHP files inside `/htdocs`, except inside :
* `/htdocs/includes` : for the most part, those external libraries have their own autoloading mechanisms
* `/htdocs/core/menus` : at the moment, each menu manager must define its own version of a class named `MenuManager`, so they can't be autoloaded yet
* all configured external modules locations (`$dolibarr_main_document_root_alt` inside `conf.php`)

As a proof of concept, I removed most of includes and requires from many of the most used modules : directories `admin`, `comm`, `commande`, `compta`, `contact`, `core`, `fourn` `install`, `modulebuilder`, `product`, `projet`, `public`, `societe`, and `user` directories inside `htdocs`, and also the `/scripts` directory.

As a bonus, in the future, this can permit us putting all Dolibarr classes inside a `namespace` without breaking compatibility with external modules, by declaring non-namespaced children of each loaded class.


marc-dll:
Hi @eldy, branch is unlocked. Thank you.

marc-dll:
Hi @eldy,

This PR is now conflict-free once again, can you please make a feedback on it ?

I understand making feedback is difficult, as the basics of the changes are hidden under the bulk of the cleaning. The changes that make the system work are only on the following files : 
* dev/tools/dolibarr-generate-classmap.php
* htdocs/classmap.inc.php
* htdocs/filefunc.inc.php

The rest is just cleaning code accordingly.

Greetings,
Marc.

eldy:
For the moment i experience a lot of pb with external modules that use autoload too and we are too short of the freeze to take risks for v22.
So i will investigate later.

For the moment, i suggest to split the pr in 2:
Can you push a pr with just the tool/script that transform the code  but with no change of code (so without running this script).
So testing and investigating on trouble the would be possible by other dev just by launching the tool that transform the code when moment will be better.

PR #33739, opened 2025-04-06
FIX public ticket layout for Safari on macOS (v18)

W1W1-M:
# FIX public ticket layout for Safari on macOS (v18)
This is a backport for #33728 & #33742

### DIFF
--- a/htdocs/public/ticket/create_ticket.php
+++ b/htdocs/public/ticket/create_ticket.php
@@ -508,7 +508,7 @@ $arrayofcss = array('/opensurvey/css/style.css', getDolGlobalString('TICKET_URL_
 llxHeaderTicket($langs->trans("CreateTicket"), "", 0, 0, $arrayofjs, $arrayofcss);
 
 
-print '<div class="ticketpublicarea ticketlargemargin centpercent">';
+print '<div class="ticketpublicarea ticketlargemargin">';
 
 if ($action != "infos_success") {
 	$formticket->withfromsocid = isset($socid) ? $socid : $user->socid;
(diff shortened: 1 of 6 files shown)


eldy:
Can you check this PR instead of #33742
https://github.com/Dolibarr/dolibarr/commit/d77898ff3dba5e7f968e7a097ded4b0d7c7cc343

W1W1-M:
@eldy Retested this PR on Safari, MS Edge/Chrome & Firefox. Index, view, list & create layouts are centered and taking up normal width.

eldy:
> @eldy Retested this PR on Safari, MS Edge/Chrome & Firefox. Index, view, list & create layouts are centered and taking up normal width.

And on smartphone ?
And what about my suggested fix ?

W1W1-M:
> > @eldy Retested this PR on Safari, MS Edge/Chrome & Firefox. Index, view, list & create layouts are centered and taking up normal width.
> 
> And on smartphone ? And what about my suggested fix ?

Removing `flexcontainer` doesn't help for Safari and seems to also break things for chrome and firefox. I looked closer at the `flexcontainer` CSS class and noticed Safari was missing from list of browsers used for `display: inline-flex`. 

I've reverted my older fixes and just added Safari to list of browser for `flexcontainer` and everything is rendering correctly.

As for iPad & iPhone, I still have the problem. I propose to make a separate PR for this (maybe not the same user agent and also need to check `getBrowserInfo` function).

If this fix is merged, then 72be18345e49d133c656847a48e7379a8354b4f3 is wrong and needs to be reverted (sorry 🙈)

W1W1-M:
Better fix #33933

PR #33747, opened 2025-04-07
fix(invoice): bad computation of rest column for invoices on list

atm-irvine:
# FIX|Fix bad computation of rest column for invoices on list
I've added getAvailableDiscounts method to get the good rest


htdocs/compta/facture/list.php
```diff
@@ -1910,8 +1910,7 @@
 			$multicurrency_totalcreditnotes = $facturestatic->getSumCreditNotesUsed(1);
 			$multicurrency_totaldeposits = $facturestatic->getSumDepositsUsed(1);
 
-			$totalallpayments = $paiement + $totalcreditnotes + $totaldeposits;
-			$remaintopay = $obj->total_ttc - $totalallpayments;
+			$remaintopay = -$discount->getAvailableDiscounts($companystatic, '', 'fk_facture_source=' . $facturestatic->id);
```
eldy, line 1913: The getAvailableDiscounts seems to return the amount of available discount générated by an invoice, not the remain to pay of an existing invoice.
The remain to pay is the total to pay minus all things that can reduce the amount to pay, so the credit notes, the deposits, the payments already done.
Old code seems correct.

atm-irvine: @eldy The problem arises when you declare an overpayment, for example. 
Example:
Invoice of €500, payment received of €1000 => "Créance" field indicated in the invoice list = -€500
If the overpayment is transformed into a discount (button for this on the invoice) and then used in part to pay another invoice (breakdown of the discount on the customer file, customer tab, link to fixed discounts), say €400 is used to pay another invoice, leaving €100 of overpayment => In this case, the "Créance" on the original invoice is still -€500, which is not right, it should show -€100.

eldy:
The valu shown must follow the same business rule than the business rule defined to show the invoice (compta/facture/card.php)

The current rule in compta/facture/card.php to define amount paid and remain to pay is:
```
	$totalpaid = $object->getSommePaiement();
	$totaldeposits = $object->getSumDepositsUsed();
	//print "totalpaid=".$totalpaid." totalcreditnotes=".$totalcreditnotes." totaldeposts=".$totaldeposits."
	// selleruserrevenuestamp=".$selleruserevenustamp;

	// We can also use bcadd to avoid pb with floating points
	// For example print 239.2 - 229.3 - 9.9; does not return 0.
	$resteapayer = price2num($object->total_ttc - $totalpaid - $totalcreditnotes - $totaldeposits, 'MT');

	// Multicurrency
	if (isModEnabled('multicurrency')) {
		$multicurrency_totalpaid = $object->getSommePaiement(1);
		$multicurrency_totalcreditnotes = $object->getSumCreditNotesUsed(1);
		$multicurrency_totaldeposits = $object->getSumDepositsUsed(1);
		$multicurrency_resteapayer = price2num($object->multicurrency_total_ttc - $multicurrency_totalpaid - $multicurrency_totalcreditnotes - $multicurrency_totaldeposits, 'MT');
		// Code to fix case of corrupted data
		// TODO We should not need this. Also data comes from a not reliable value of $object->multicurrency_total_ttc that may be wrong if it was
		// calculated by summing lines that were in a currency for some of them and into another for others (lines from discount/down payment into another currency for example)
		if ($resteapayer == 0 && $multicurrency_resteapayer != 0 && $object->multicurrency_code != $conf->currency) {
			$resteapayer = price2num((float) $multicurrency_resteapayer / $object->multicurrency_tx, 'MT');
		}
	}

	if ($object->paye) {
		$resteapayer = 0;
	}
```

So the difference with the list is that we have at end 
```
if ($object->paye) {
		$resteapayer = 0;
}
```
 in the card that we don't have.
So i think the fix should be to add this test only.
Can you check if it works as expected for you with this simple change ?

atm-irvine:
@eldy The correct amount is on the comm/remx.php page. In the section "Réductions ou crédits disponibles", we have a list of overpayment invoice with discount available. The amount of discount available is correct in this section.

eldy:
> @eldy The correct amount is on the comm/remx.php page. In the section "Réductions ou crédits disponibles", we have a list of overpayment invoice with discount available. The amount of discount available is correct in this section.

The value into remx are the discount available for consumption into other invoices, not the remain to pay of an invoice.

If you have an invoice of 100, then you pay 20, then you consume an old credit not of 30, and you apply a commercial discount of 10, the remain to pay is 40.
You won't fin 'd any entry into comx with a value of 40. You will just find an entryvfor the 30 of the old credit not to consume it and an entry of 10 for the commercial disount to consume.

I think the trouble of overpayment has a different origin. The calculation for column value must be the same than when reading the card. I pushed recently a fix on develop to align the code on list with the one on code so you should find a remain to pay of 0 when there is an overpayment now.
Fix may be backported.

rycks:
> I pushed recently a fix on develop to align the code on list with the one on code so you should find a remain to pay of 0 when there is an overpayment now.

@eldy did you have the commit link of that for the backport ?

or at least the file name impacted by your fix ?

eldy:
Sorry it was a long time ago, and i can't retreive where the fix was. I think it was "several fixes" to fix the bad consistency.

Nevermind, on recent version, the remain to pay is calculated with
```
$totalpaid = $object->getSommePaiement();
$totaldeposits = $object->getSumDepositsUsed();
$resteapayer = price2num($object->total_ttc - $totalpaid - $totalcreditnotes - $totaldeposits, 'MT');
```
And the part of amount of excess received seems already included into the getSumCreditNotesUsed(). So don't need to use the getAvailableDiscounts(). So i surely miss something, so the most important is to be able to find a use case to reproduce the bug. I failed to find one. So if you can describe a proces to reproduct it we may progress on this.

Seems duplicate of need for #33191

JamBalaya56562:
This looks resolved now.

As discussed in this thread, the invoice **list** was aligned with the invoice **card** (`compta/facture/card.php`): in `list.php`, a closed/paid invoice now gets `$remaintopay = 0` via

```php
if ($facturestatic->status == Facture::STATUS_CLOSED) {
    $remaintopay = 0;
    $multicurrency_remaintopay = 0;
}
```

which is the "remain to pay of 0 when there is an overpayment" @eldy mentioned above.

In addition, #33191 has just been merged (into 18.0, so it will be merged-forward to the maintained branches): for a paid invoice it strikes through the remain-to-pay value in the list and removes it from the column total — covering this overpayment case, following the card's business rule rather than the `getAvailableDiscounts()` value (which, as @eldy noted, is the discount available for consumption in other invoices, not the invoice's remain to pay).

So the underlying issue looks fixed and this PR can probably be closed.

PR #33749, opened 2025-04-07
FIX: exclude lines with type 9 from qties calculation in receptions …

jyhere:
Exclude lines with type 9 from qties calculation in reception /shipment workflow

How to reproduce the bug : 
- set the increase stock on reception validate
- set STOCK_SUPPORTS_SERVICES = 1
- Activate the Workflow module, with WORKFLOW_ORDER_CLASSIFY_RECEIVED_RECEPTION  = 1
- use a free text line module like Subtotal
-  Create a supplier order with a product line and a title line
- Create the reception from the order, validate the reception. The order won't be classified.

### DIFF
--- a/htdocs/core/triggers/interface_20_modWorkflow_WorkflowManager.class.php
+++ b/htdocs/core/triggers/interface_20_modWorkflow_WorkflowManager.class.php
@@ -450,7 +450,8 @@ class InterfaceWorkflowManager extends DolibarrTriggers
 					if (is_array($order->lines) && count($order->lines) > 0) {
 						foreach ($order->lines as $orderline) {
 							// Exclude lines not qualified for shipment, similar code is found into calcAndSetStatusDispatch() for vendors
-							if (!getDolGlobalString('STOCK_SUPPORTS_SERVICES') && $orderline->product_type > 0) {
+							if ((!getDolGlobalString('STOCK_SUPPORTS_SERVICES') && $orderline->product_type > 0) ||
+								$orderline->product_type == 9) {
 								continue;
 							}
 							if (isset($qtyordred[$orderline->fk_product])) {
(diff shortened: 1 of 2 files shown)


JamBalaya56562:
**Update (2026-08-19): this is now fixed on `22.0`.**

I had opened #39531 for it, but the same commit was also sitting on the branch of #39532, so it went in with that merge instead (merge commit 1459d36). #39531 is therefore closed as already-applied. `interface_20_modWorkflow_WorkflowManager.class.php` on `22.0` now skips `product_type == 9` in both loops, the shipment one and the reception one.

I could reproduce your scenario exactly: with `STOCK_SUPPORTS_SERVICES = 1`, the title line adds an entry `$qtyordred['']` (the line has no product), which no received quantity can ever match, so `array_diff_assoc()` never becomes empty and the order stays unclassified. It even shows in the log as `Undefined array key ""` on that line.

Same diagnosis as yours. Two details differ from your patch:

- the exclusion is a separate guard next to the existing condition rather than an extension of it;
- `CommandeFournisseur::calcAndSetStatusDispatch()` is left alone, because on `24.0` and `develop` the condition is already `$line->product_type > 0 && (!getDolGlobalString('STOCK_SUPPORTS_SERVICES') || empty($line->stockable_product))`, which already skips title lines, and changing it on `22.0` would conflict when merging forward.

Measured before/after on a local install (supplier order with a product line qty 2 + a title line, reception of the whole quantity): order status stays at 1 before the fix, and becomes 5 (received completely) after it, while an order without a title line is unaffected.

Status per branch as of today: `22.0` fixed, `23.0` / `24.0` / `develop` not yet — the change should reach them through the usual forward merges from `22.0`. If it does not land on `develop`, I will open a dedicated PR there.

Credit for the report and the diagnosis goes to @jyhere.

---

@eldy I think this PR is safe to close now: the `WorkflowManager` part is applied on `22.0` (1459d36) and the `fournisseur.commande.class.php` part is already covered on `develop` by the `empty($line->stockable_product)` condition, so there is nothing left to merge from here. Closing it as superseded looks right to me — but happy to leave it open if you prefer to wait until the forward merge actually reaches `develop`.

PR #27570, opened 2024-01-16
NEW shipment kits with dispatcher v2

lvessiller-opendsi:
NEW shipment kits with dispatcher v2
- see https://github.com/Dolibarr/dolibarr/pull/26519 for SQL structure

### Example
**1. Create a virtual product**
Virtual product '"PANIER1" contains sub-products  : 
- 1 x PANIERVIDE
- 3 x POMME1
- 2 x ANANAS1
- **MINIPANIERGARNI** (virtual product)
image

and **MINIPANIERGARNI** it's a sub virtual product and contains sub-products : 
- 1 x POMME1
- 1 x ANANAS1
image

**2. Create an order**
image

**3. Create a shipment**
image

Then choose your quantities on each product line to prepare the shipment (see 5. to dispatch all shipment lines and children products of a virtual product)
image
image

**4. Shipment**
Here you can see all lines of the shipment
image

and the details for each warehouse and quantities of each sub-product  as tooltip : 
_POMME1 P1 : ENTREPOT2 (Qty: 4)
POMME1 P1 : STOCKP2 (Qty: 4)
ANANAS1 Ananas : ENTREPOT2 (Qty: 3)
ANANAS1 Ananas : STOCKP2 (Qty : 3)_

and we can edit a virtual product
- link to "Dispatch shipment" tab (see 5. above)

and we can delete a virtual-product
- all children are deleted (by reccurence)

**5.Dispatch shipment**
Here you can select each warehouse and the quantities for each line of products and sub-products
image

And then you can save all dispatched lines and validate your shipment.

**5.Stock**
For virtual product
- check on all children if the checkbox "Increase or dscrease stock" (incdec) is checked (see 1.) and then dicrease the stock with the quantity dispatched for each warehouse

The stock for virtual product is not managed : movements are only for sub-products (if "incdec" is checked)
- for "POMME1"
image
- for "ANANAS"
image

The movements are only when we validate or close a shipment (see parameters of stock module)

And when we cancel or delete a shipments, all inverse movements are done in for stock of each sub-product (not virtual product)
- example for sub-product "POMME1"
image

- example for sub-product "ANANAS1"
image

**Kit with batches as sub-components**
In shipment details you can choose serial numbers for sub-components
image

PR #33750, opened 2025-04-07
NEW shipment kits with dispatcher v2

lvessiller-opendsi:
NEW shipment kits with dispatcher v2
- see https://github.com/Dolibarr/dolibarr/pull/26519 for SQL structure
- repost #27570

### DIFF
--- a/htdocs/admin/expedition.php
+++ b/htdocs/admin/expedition.php
@@ -456,6 +456,13 @@ print '<td>';
 print ajax_constantonoff('SHIPPING_DISPLAY_STOCK_ENTRY_DATE');
 print '</td></tr>';
 
+print '<tr class="oddeven">';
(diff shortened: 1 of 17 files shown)


peyo0:
Hello, are you considering the fact that we need this feature on receptions as well ?

lvessiller-opendsi:
> Hello, are you considering the fact that we need this feature on receptions as well ?

For me, the need is not the same for receptions : we only receipt components of virtual product.

peyo0:
Well, not in the case you import products, then you receive packages of products. In this case, we need to handle those packages first before handling the products.

lvessiller-opendsi:
Maybe you could have an use case on supplier order and receipts of virtual products but it"s not implemented here.
You could suggest it in a feature request.

PR #33758, opened 2025-04-07
FIX avoid sql error when a comment is after without comma

hregis:


hregis:
@eldy what is ERLT ??

`$buf = preg_replace('/([,;ERLT\s+\)])\s*--.*$/i', '\1', $buf); `

htdocs/core/lib/admin.lib.php
```diff
@@ -242,7 +242,7 @@ function run_sql($sqlfile, $silent = 1, $entity = 0, $usesavepoint = 1, $handler
 			// Add line buf to buffer if not a comment
 			if ($nocommentremoval || !preg_match('/^\s*--/', $buf)) {
 				if (empty($nocommentremoval)) {
-					$buf = preg_replace('/([,;ERLT\)])\s*--.*$/i', '\1', $buf); //remove comment from a line that not start with -- before add it to the buffer
+					$buf = preg_replace('/([,;ERLT\s+\)])\s*--.*$/i', '\1', $buf); //remove comment from a line that not start with -- before add it to the buffer
```
eldy, line 245: Looks strange to add \s+ because we are inside [ and ] and it means any char in this list before the \s*. I think the chard , ; E R L T were added as the last character of a sql string before the comment to avoid to remove the -- when it is not used as a comment.

eldy: So you just need \s and not \s+ inside the [ ].
According to regex syntax, using \s+ inside [ ] has no sense.
Do you want to do :
`preg_replace('/([,;ERLT\s\)])\s*--.*$/i', '\1', $buf);`
or
`preg_replace('/([,;ERLT\s\+\)])\s*--.*$/i', '\1', $buf);`

PR #33763, opened 2025-04-08
fix(mail): lot of 0 after amount of total_ht and total_ttc

atm-irvine:
# FIX|Fix lot of 0 after amount of total_ht and total_ttc
I've added update_price before substitution because there was a lot of 0 after the price in the email.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -9344,6 +9344,7 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 
 		// For backward compatibility
 		if ($onlykey != 2) {
+			if (is_object($object)) $object->update_price();
 			$substitutionarray['__TOTAL_TTC__']    = is_object($object) ? $object->total_ttc : '';
 			$substitutionarray['__TOTAL_HT__']     = is_object($object) ? $object->total_ht : '';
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -9344,6 +9344,7 @@ function getCommonSubstitutionArray($outputlangs, $onlykey = 0, $exclude = null,
 
 		// For backward compatibility
 		if ($onlykey != 2) {
+			if (is_object($object)) $object->update_price();
```
eldy, line 9347: We should not have any SQL action inside the getCommonSubstitutionArray that should remain a memory only function.
Can you describe how to reproduce the problem ?

atm-irvine: @eldy A customer reported an issue with the __TOTAL_TTC__ variable. It displays too many zeros at the end, even though the configuration limits it to two decimal places. I don't understand why.

eldy: First thing to do is to look into database. If data was inserted directly in database or with old version the value may have not be rounded by the rounding limit.
Note: We round during the insert, never when showing on screen. Data must have correct rounding in database. if not, database is corrupted.
if yes, rounding on screen is useless.

atm-irvine: Ok, I see your point. I'm going to check this right now, thanks.

PR #33764, opened 2025-04-08
NEW Add contacts and addresses to templace invoices / factures recurrentes with MAIN_PROPAGATE_CONTACTS_FROM_ORIGIN

Vaadasch:
# NEW Add contacts and addresses to templace invoices / factures recurrentes
With this PR, it will allow to add contacts and addresses to recurring invoices / templates invoices / facture-rec

On invoice generation, contacts will be propagated, only with the use of the MAIN_PROPAGATE_CONTACTS_FROM_ORIGIN constant.
Need to discuss about the point of having MAIN_PROPAGATE_CONTACTS_FROM_ORIGIN for invoice-rec. If we set contacts in invoice-rec, it is not for the template but for the generated invoices. So we might want to have it by default, without this const.

We have the problematic about the duplication of entering twice the Dict->Contact Type, one for invoice and one for invoice-rec.
I didn't found any automation allowing to enter it only once. I don't think that it is really a problem for users willing to use this functionnality to think to enter a new type both in invoice and invoice-rec. It is only once in many days, and if it is not working, it will think of that or quickly find the solution in a forum post / wiki / other.
Automation of that could come in the future.

This propagation is working ( from template to invoice ) and ( from invoice to template )

This PR add the link between generated invoices and their template, without the link to the other generated invoices.

I took the liberty to remove a `if (0) {...}` block that was lurking since v17

### DIFF
--- a/htdocs/admin/dict.php
+++ b/htdocs/admin/dict.php
@@ -718,6 +718,7 @@ if ($id == DICT_TYPE_CONTACT) {
 		'commande' => img_picto('', 'order', 'class="pictofixedwidth"').$langs->trans('Order'),
 		'shipping' => img_picto('', 'dolly', 'class="pictofixedwidth"') . $langs->trans('Shipment'),
 		'facture' => img_picto('', 'bill', 'class="pictofixedwidth"').$langs->trans('Bill'),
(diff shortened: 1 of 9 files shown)


Vaadasch:
Too much changes

PR #33768, opened 2025-04-08
New Add CCI in Soc Rib

leninrivas:
# NEW|New *Add CCI in Soc Rib*
*Interbank code to make a transfer between accounts at different banks*


htdocs/societe/paymentmodes.php
```diff
@@ -1579,6 +1582,7 @@
 		print_liste_field_titre($form->textwithpicto($langs->trans('ExternalSystemID'), $langs->trans("IDOfPaymentInAnExternalSystem")));		// external system ID
 		//print_liste_field_titre("Bank");
 		print_liste_field_titre("RIB");
+		print_liste_field_titre("CCI");
```
eldy, line 1585: I think the code CCI is used by Peru only, right ?
If yest, can you add a condition
`if ($mysoc->country_code == 'PE) {`
'

htdocs/install/mysql/tables/llx_societe_rib.sql
```diff
@@ -36,6 +37,7 @@ create table llx_societe_rib
   code_banque						varchar(128),  -- bank code
   code_guichet						varchar(6),    -- desk code
   number							varchar(255),  -- account number
+  cci								varchar(32),   -- Interbank code to make a transfer between accounts at different banks
```
eldy, line 40: If you add the field cci into table llx_societe_rib, you must also add it into migration script.
I pushed a PR to do it.

PR #33769, opened 2025-04-08
NEW Add contacts and addresses to templace invoices / factures recurrentes with MAIN_PROPAGATE_CONTACTS_FROM_ORIGIN

Vaadasch:
# NEW Add contacts and addresses to templace invoices / factures recurrentes with MAIN_PROPAGATE_CONTACTS_FROM_ORIGIN
(Try 2 : see #33764 ; there are code review comments in this previous PR )
With this PR, it will allow to add contacts and addresses to recurring invoices / templates invoices / facture-rec

On invoice generation, contacts will be propagated, only with the use of the MAIN_PROPAGATE_CONTACTS_FROM_ORIGIN constant.
Need to discuss about the point of having MAIN_PROPAGATE_CONTACTS_FROM_ORIGIN for invoice-rec. If we set contacts in invoice-rec, it is not for the template but for the generated invoices. So we might want to have it by default, without this const.

We have the problematic about the duplication of entering twice the Dict->Contact Type, one for invoice and one for invoice-rec.
I didn't found any automation allowing to enter it only once. I don't think that it is really a problem for users willing to use this functionnality to think to enter a new type both in invoice and invoice-rec. It is only once in many days, and if it is not working, it will think of that or quickly find the solution in a forum post / wiki / other.
Automation of that could come in the future.

This propagation is working ( from template to invoice ) and ( from invoice to template )

This PR add the link between generated invoices and their template, without the link to the other generated invoices.

I took the liberty to remove a `if (0) {...}` block that was lurking since v17


Vaadasch:
For the php-stan errors :
`Warning: custom_prompt.php: PhanUndeclaredVariableAssignOp: Variable $out was undeclared, but it is being used as the left-hand side of an assignment operation` 
I didn't modify custom_prompt.php so i'm not sure what to do about this.

`Warning: contact.php: PhanTypeMismatchArgument: Argument 5 ($nolines) is getDolGlobalBool('INVOICE_USE_SITUATION') of type bool but \FactureRec::fetch() takes int defined at htdocs/compta/facture/class/facture-rec.class.php:650` 
Yeah, it seem not to understand that $object will be used differently given the $object_rec value.
I'm not sure either how to arrange the code to pass this test.

htdocs/install/mysql/migration/21.0.0-22.0.0.sql
```diff
@@ -246,3 +246,9 @@ ALTER TABLE llx_contratdet DROP COLUMN remise;
 ALTER TABLE llx_extrafields ADD COLUMN aiprompt text;
 
 ALTER TABLE llx_menu ADD COLUMN showtopmenuinframe integer DEFAULT 0;
+
+-- Add contacts to reccurent / templates invoices / factures
```
eldy, line 250: When you need to change the database structure, you must first push an independent PT with just the database structure addition, with no code change.

PR #33770, opened 2025-04-08
add param nocreateEcmFile

atm-jonathan:
When the list_of_documents() method is called to display.
And one of the files has already been saved, we get a warning `DB_ERROR_RECORD_ALREADY_EXISTS `
With this additional` $noCreateEcmFile` parameter, we prevent the `$ecmfile->create() `of the `EcmFiles` object from being launched.


eldy:
i made a mistake by validating a new feature into an old version.
Of course it generates trouble when merging with recent version that already used the added parameter for another need.
So i had to revert it.
Thanks to redo the PR in develop branch.

PR #33779, opened 2025-04-08
Add hook before each generated invoice by createRecurringInvoices

Vaadasch:
The need is to be able to adapt to specifics :

Like to set the facturation date long in the future (3m) and keep the invoice in draft until then.
Unable to do so without this hook or without making code modifications too important for this simple need.


Vaadasch:
Very wrong line modification

Vaadasch:
Replaced by #33780

PR #33780, opened 2025-04-08
Add hook before each generated invoice by createRecurringInvoices

Vaadasch:
The need is to be able to adapt to specifics :

Like to set the facturation date long in the future (3m) and keep the invoice in draft until then.
Unable to do so without this hook or without making code modifications too important for this simple need.


Vaadasch:
Well, how can I respect the parameters of the executeHooks functions and still pass 2 objects ($facture ans $facturerec) to the hook without creating 2 hooks ?

htdocs/compta/facture/class/facture-rec.class.php
```diff
@@ -1455,6 +1455,8 @@ public function createRecurringInvoices($restrictioninvoiceid = 0, $forcevalidat
 						}
 					}
 
+					$reshook = $hookmanager->executeHooks('beforeCreationOfEachRecurringInvoice', $parameters, $facturerec, $facture); // note that $facturerec or $facture might be modified by hooks
```
eldy, line 1458: Signatrue of executeHooks must always be
$hookmanager->executeHooks('myHookName', $parameters, $object, $action);
but you can provide more parameter inside the $parameters array
$parameters = array('myparam' => &$valueofmyparam);

Also you alreaydy have a hook 
beforeCreationOfRecurringInvoices

Why not using it ?
Also do you know you can also use the trigger on invoice create ?

Vaadasch: Hi @eldy ,
My need is to create invoices kept in draft until the next generation, so the facturation date of these is the next date of generation. If i do it inside the recurring invoice generation, I can use the $facturerec->getNextDate(), all following calculations are done inside $facture->create(), it is as simple as : 
``` php
public function beforeCreationOfEachRecurringInvoice($parameters, $facturerec, &$facture) {
    $facture->date = $facturerec->getNextDate();
}
``` 
It allows me to have not only the facturation date at the date i want, but the start and end date of services lines too.

The only thing i have with beforeCreationOfRecurringInvoices is the sql request, not even its result, which i would have willing to modify for each invoice. But maybe i can make date calculations inside the sql to have different values for each result line, i'm note sure.
``` php
$reshook = $hookmanager->executeHooks('beforeCreationOfRecurringInvoices', $parameters, $sql); // note that $sql might be modified by hooks
```
With this way, the dates of the service lines are auto-adapting to the value of the facturation date.

Using triggers will need me to retrieve the liked invoice model, take its frequency, calcul anew the facturation date, apply this new date, apply it to the start of each service line, calcul anew the end of service line and apply it.
I can do it with trigger but it will be long and hard.

Thanks for your tip about including the objects in the params, i'll modify my code.

Vaadasch:
As executeHooks need its 3rd parameter to be object, i choose to send it $facture.
For consistency, I've add both the $facture and the $facturerec in the $parameters, but $facture is not needed here, as it is send in the 3rd parameter.

My bad concerning the update of the service lines dates, it is based on a new FactureRec($db)->date_when.
I really think that the dates of the services do not need to be based on the generation of the invoice but on the billing date of the invoice, but that is for another PR #33794

Vaadasch:
#33794

eldy:
I removed the facturerec from parameter because alreay into $object param (that is already passed by reference)

PR #33784, opened 2025-04-09
FIX - $error is never set 

atm-lucasmantegari:
# FIX - $error is never set
*The error variable is never set, so if it is filled during recursive creation, the error always remains set.*

### DIFF
--- a/htdocs/compta/facture/class/facture-rec.class.php
+++ b/htdocs/compta/facture/class/facture-rec.class.php
@@ -1318,6 +1318,7 @@ class FactureRec extends CommonInvoice
 				$this->db->begin();
 
 				$invoiceidgenerated = 0;
+				$error = 0;
 
 				$facture = null;
 				$facturerec = new FactureRec($this->db);
@@ -1426,7 +1427,7 @@ class FactureRec extends CommonInvoice
 
 		$this->output = trim($this->output);
 
-		return $error ? $error : 0;
+		return count($this->errors);
 	}
 
 	/**


htdocs/compta/facture/class/facture-rec.class.php
```diff
@@ -1426,7 +1427,7 @@ public function createRecurringInvoices($restrictioninvoiceid = 0, $forcevalidat
 
 		$this->output = trim($this->output);
 
-		return $error ? $error : 0;
+		return count($this->errors);
```
eldy, line 1430: this->errors may not be always set, sometimes it is ->error that is set only.

If you move $error = 0; before the loop, is it ok ?

frederic34:
image
it is already set ?

eldy:
> image it is already set ?

True, so i don't see in which case this change fix something.
If this->errors is set, $error should also be set.

atm-lucasmantegari:
The problem is that it's outside the loop. Because it's never reset when I loop, so if I have 50 invoices and an error occurs at invoice 2, the other 48 invoices will have the same $error.

eldy:
> The problem is that it's outside the loop. Because it's never reset when I loop, so if I have 50 invoices and an error occurs at invoice 2, the other 48 invoices will have the same $error.

Because, there is only 1 error code return, for 50 invoices, the error code is global and mean "there is ate least one error". 
So currently the return is the number of invoice in error only. 

The problem is that we used it also inside the loop to track error on the current processed invoice. 2 need but 1 variable. This is the trouble. So inside the loop, we should use another error code for the given current processed invoice. On other part of code, we are using a variable $errorforline or $errorforinvoice. So i pushed a fix using this method.

PR #33794, opened 2025-04-09
FIX Update recurring invoice generation to calculate services date based on billing date not generation date

Vaadasch:
FIX Update recurring invoice generation to calculate services date based on billing date not generation date
On recurring invoice generation with a freqency, dates in services lines are based on the generation date instead of the billing date.

The generation date has nothing to do with the start and end of the services, which need to be based on the billing date.
A user might want to generate its invoice weeks or even months in advance (or rarely after) the billing date, and that need not to be linked to the actual billing date.

This exemple enlight a billing date different of the creation date.
For the same setting of recurring invoice model :
image
Generation of invoice planned for 01/04/2025 with a billing date at 01/08/2025 and prepaid mode.
The problem affect invoices with billing date in the past too, but for clarity of the exemple, here the billing date is set in the futur. 
Setting a billing date in the future need INVOICE_MAX_FUTURE_DELAY 
# Before 
image
# After 
image

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -547,9 +547,11 @@ class Facture extends CommonInvoice
 
 		$this->db->begin();
 
-		$originaldatewhen = null;
-		$nextdatewhen = null;
-		$previousdaynextdatewhen = null;
+		$originalGenDateWhen = null;
+		$nextGenDateWhen = null;
+		$previousDayNextGenDateWhen = null;
+		$startOfServices = null;
+		$endOfServices = null;
 
 		$_facrec = null;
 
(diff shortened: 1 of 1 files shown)


eldy:
Can you confirm me this:
If param INVOICE_MAX_FUTURE_DELAY is not set, is there a difference between before and after your change ?

Vaadasch:
INVOICE_MAX_FUTURE_DELAY only prevent to create invoices in the future.
The change base the start date and end date of services on the billing date instead of creation date.
It is not in relation

The exemple is just more clear with a facturation in the future, it does not affect the behavior modified with the PR.
I've rephrased the first exemple.

Without INVOICE_MAX_FUTURE_DELAY, the difference would be the following (with exactly the same settings of the model) : 
Generation of invoice planned for 01/04/2025 with a billing date at 01/01/2024 (in the past) and prepaid mode.
# Before
image
# After 
image

Note : if the billing date is exactly the same as the creation date, like it is the case for automatic generation with cron (without hook), there is absolutely no difference of behavior.
If the new hook beforeCreationOfEachRecurringInvoice change the billing date, the dates of the services will now change, as it will not be dependant of the generation date.
INVOICE_MAX_FUTURE_DELAY is not needed when creating recurring invoices with a billing date in the future with cron, it is only needed with manual generation.

PR #33806, opened 2025-04-09
Add STATUS_CLOSED on interventional graph

defrance:


frederic34:
https://github.com/Dolibarr/dolibarr/pull/33869

PR #33809, opened 2025-04-10
NEW hidden conf for api contrat putLine endpoint to only return the line

atm-florianm:
# NEW hidden conf to enable sparse output for Contracts::putLine endpoint

When contracts have thousands of lines, the putLine endpoint becomes unwieldy because it returns the whole contract with every call, which is not a problem when you have smaller contracts.

This PR introduces a hidden conf to change this behavior and return only the modified line (and also remove attributes that are not set on the ContratLigne object when fetching it).

#### Off-topic
Unfortunately, I was unable to run the pre-commit hook because of unsatisfied dependencies; I am going to fix this, but I need to upgrade my distro first; in the meantime, I'm happy to fix problems with this PR manually.

### DIFF
--- a/htdocs/contrat/class/api_contracts.class.php
+++ b/htdocs/contrat/class/api_contracts.class.php
@@ -450,8 +450,81 @@ class Contracts extends DolibarrApi
 		);
 
 		if ($updateRes > 0) {
-			$result = $this->get($id);
-			unset($result->line);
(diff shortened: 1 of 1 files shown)


atm-florianm:
I cherry-picked my commit from a commit designed for branch 17.0, but I'm realizing there are structural differences that make my commit not suitable for develop (as pointed out by the phan inspection).

[EDIT] ☑ I fixed the problem; the PR should be good now (just waiting for the Travis build to complete to remove the label).

PR #33817, opened 2025-04-11
return empty array instead -1 if fail

defrance:
the function is used directly on loop

ex : 		foreach ($this->fetchAllByFkProductParent($fk_product_parent) as $prodcomb) {

so if we have a no value we fail...

### DIFF
--- a/htdocs/variants/class/ProductCombination.class.php
+++ b/htdocs/variants/class/ProductCombination.class.php
@@ -302,12 +302,10 @@ class ProductCombination
 	 *
 	 * @param	int							$fk_product_parent	Rowid of parent product
 	 * @param	bool						$sort_by_ref		Sort result by product child reference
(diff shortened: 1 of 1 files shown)


htdocs/variants/class/ProductCombination.class.php
```diff
@@ -302,12 +302,10 @@ public function fetchByFkProductChild($productid, $donotloadpricelevel = 0)
 	 *
 	 * @param	int							$fk_product_parent	Rowid of parent product
 	 * @param	bool						$sort_by_ref		Sort result by product child reference
-	 * @return	int|ProductCombination[]						Return integer <0 KO
+	 * @return	ProductCombination[]						Return empty array if ko
```
eldy, line 305: How do we make a difference between technical error and no error but nothing found ?

PR #33818, opened 2025-04-11
add param nocreateEcmFile

atm-jonathan:
When the list_of_documents() method is called to display.
And one of the files has already been saved, we get a warning DB_ERROR_RECORD_ALREADY_EXISTS 
With this additional $noCreateEcmFile parameter, we prevent the $ecmfile->create() of the EcmFiles object from being launched.

see
[](https://github.com/Dolibarr/dolibarr/pull/33770)


frederic34:
this can happen when changing ref by example from aaa to AAA, ecmfiles may be not correctly updated, is it a product ?

Issue #33617, opened 2025-03-26
Variable not interpreted


fred-gb:
### Bug

Hi,

Following this post on dolibarr forum:
https://www.dolibarr.fr/forum/t/variables-de-substitutions-non-interpretees/48905

Variable `__THIRDPARTY_NAME__` was not interpreted in automatic mail after events like invoice paid. 

All others variables are interpreted. For example, in received mail:
```
Bonjour,

Client: __THIRDPARTY_NAME__
Facture FA2501-2948 a été réglée
Montant: 132,00 € TTC ( 110,00 € HT )
Fiche facture: [FA2501-XXXX](https://XXXXXXXXXfacture/card.php?id=XXXX)
```

### Dolibarr Version

20.0.4

### Environment PHP

8.2.27

### Environment Database

MariaDB 10.11.7-MariaDB-1:10.11.7+maria~ubu2204-log

### Steps to reproduce the behavior and expected behavior

Create a model email `invoice paid`like:
```
Bonjour,<br />
<br />
Client: __THIRDPARTY_NAME__  
<br />
Facture __REF__ a été réglée
<br />
Montant: __AMOUNT_FORMATED__ TTC ( __AMOUNT_EXCL_TAX_FORMATED__ HT )
<br />
Fiche facture: <a href="__URL_INVOICE__">__REF__</a><br />
<br />
```

In notification module. Put for event `Customer invoice paid` the email model.

Vaadasch:
Hi @fred-gb ,

Following and outputting debug inside the email, i've managed to track the problem.

It's because $object->thirdparty is not set when calling getCommonSubstitutionArray

You can add those two lines inside `core/class/notify.class.php` : 
```
function send( ... ) {
       // Check notification per third party
        if (!empty($object->socid) && $object->socid > 0) {
            // ADD THESE TWO LINES
            $object->thirdparty = new Societe($this->db);
            $object->thirdparty->fetch($object->socid);
            // Just before the SQL request construction
            $sql .= '...';
```

Can you manage to add those lines inside the develop branch of the github and create the PR to the FIX ?

PR #33820, opened 2025-04-11
Fix: Add missing thirdparty variable substitution in notify class

fred-gb:
Bonjour 👋🏻 

# FIX #[[*33617*](https://github.com/Dolibarr/dolibarr/issues/33617)]

Thanks @Vaadasch


Vaadasch:
Well, i've might have overlooked a little too much on this fix 😅
Following other interactions i've had with eldy, it might be a little overloading, loading a new Thirdparty over almost every notification.
And it seems that the gardian phan doesn't like that either.

Vaadasch:
Well, i've tracked a way to have a more lazy load, and found parts in the notifications.class.php where it is `$object->fetch_thirdparty()` that is called... only to make inside this more tests and then do a Societe::fetch()...

fred-gb:
OK. But I don't know dev in PHP. So... I don't understand, sorry 😅

Vaadasch:
No problem, @fred-gb , it was more a comment for the great Merger :-)
Thanks for doing this, it help

htdocs/core/class/notify.class.php
```diff
@@ -679,6 +679,8 @@ public function send($notifcode, $object, $filename_list = array(), $mimetype_li
 
 		// Check notification per third party
 		if (!empty($object->socid) && $object->socid > 0) {
+			$object->thirdparty = new Societe($this->db);
```
eldy, line 682: This code is inside the if but the $object->thirdparty seems never used inside this if. So i suppose this code must be at a higher level.

Also in dev, we already have it line 955
```
if (!($object->thirdparty instanceof Societe)) {
							$object->fetch_thirdparty();
						}
```

So i think the trouble is alreayd fixed in develop. If not can you decribe how to reprouce on develop ?

Vaadasch: Yes, thanks @eldy, it is indeed already fixed in the develop branch.
Btw, i didn't succeed in sending email notification with the current version, it was poping an error message : 
```
Error [120]: Ran SMTP into problems sending Mail. Response: 501 Syntax error in parameters or arguments Error [120]: Ran SMTP into problems sending Mail. Response: 554 For explanation visit https://www.ionos.com/help/index.php?id=2425&ip=<MY_PUBLIC_IP>&c=hd&r=1MYvTs-1tYr181Svw-00JGmE
```
(I've masked <MY_PUBLIC_IP>)

I've managed to track the error message to the /core/class/smtps.class.php and put a var_dump($this) just before to confirm the substitution of this PR / bug, so i can confirm that __THIRDPARTY_NAME__ is properly substituted.

@fred-gb, it was a nice try but we learn something today : we need to test a bug with the develop branch before submitting a fix ^^

Issue #33829, opened 2025-04-13
Implement graceful handling of "21.500,00" (French, ...)


mdeweerd:
Whenever I copy/paste a value similar to 21.500,00 into a field of dolibarr, it is quite annoying that after validation this resolves to "21,5" instead of "21 500,00".

Yes, even French invoices or bank statements like to show the 'dot' as a separator.

However, the thousands separator is a space:
```plaintext
main.lang:15:1:SeparatorDecimal=,
main.lang:16:1:SeparatorThousand=Space
```

Strictly speaking, we could state that interpreting 21.500,00 as 21.5 is strictly invalid as the decimal separator for french is ',' and not the dot !

IMHO it is a nice to have to have some smarter handling of such an input.

If we have '.' and ',' in the same value, then Dolibarr could use the latter character as the decimal separator and the first character as the thousands separator.  I would be ok if the thousands separator would only be space or '.' in french because a thousands separator being a ',' is not used.

So:
- 10.0 -> 10,0
- 1.324,32 -> 1 324,32
- 1324,32 -> 1 324,32
- 1 324.32 -> 1 324,32
- 1.324,323.32 -> Invalid

I would suggest this for all languages where a space is accepted.

Maybe this could be reflected in the translation file.  Then is adjustable by language, and I suppose also by user preferences.

```plaintext
main.lang:15:1:SeparatorDecimal=,
main.lang:16:1:SeparatorThousand=SpaceOrDot
```

I added a test case to dolibarr to demonstrate this #33830 .

PR #33830, opened 2025-04-13
UIUX: Interpret first token as thousands separator 21.500,00 (when French, ...).

mdeweerd:
# UIUX Interpret first token as thousands separator 21.500,00 (when French, ...).

This demonstrates #33829 in a test case adjustment and provides a code modification to implement it.


mdeweerd:
Closing because phpstan does not support the branch name, replaced with #33832

PR #33832, opened 2025-04-13
UIUX: Interpret first token as thousands separator 21.500,00 (when French, ...).

mdeweerd:
# UIUX Interpret first token as thousands separator 21.500,00 (when French, ...).

This demonstrates #33829 in a test case adjustment and provides a code modification to implement it.


htdocs/core/lib/functions.lib.php
```diff
@@ -7260,7 +7260,20 @@ function price2num($amount, $rounding = '', $option = 0)
...
+			if (preg_match('/^(?<int>[^,]*,|[^.]*\.)(?<dec>[^.,]*)$/u', $amount, $matches)) {
+				$intPart = $matches['int'];
+				$decPart = $matches['dec'];
+
+				// Remove all commas and dots from intPart
+				$intPart = str_replace(['.', ','], '', $intPart);
+
+				// Combine intPart and decPart with a dot
+				$amount = $intPart . $dec . $decPart;
+			}
```
eldy, line 7276: Reading regex rule is never easy. For security purpose may be we can add a else here to execute old code in case of regex test is false ?

mdeweerd: ## Summary

IHMO the old method is insecure because it allows improbable numbers like '100.023,210.00`.

But if you still want me to add an else condition to execute the old replacement.  Reminder: it was already an exceptional handling "only for french".

More desired outcomes for special inputs can be added as test cases and if these require the old method as the "alternative" then that is a clear motivation for it.


## Discussion:

The old code replaced all the ',' with '.'.  That is in itself not very secure...
```
$amount = str_replace(',', '.', $amount); // To accept 2 notations for french users
```

The new code does not make blind replacements.  It will only replace if the last first n dots or commas are all different from the apparently "last decimal point".
For instance, it will not make replacements in '100.023,210.00` will not receive replacements in the regex.

If we want security, then there would be a check at the end verifying the there is at most a single dot in the number.

The regex  `/^(?<int>[^,]*,|[^.]*\.)(?<dec>[^.,]*)$/u` can read as follows, and broken up in these parts:
- `(?<int>[^,]*,|[^.]*\.)` is any accepted sequence up to the last potential decimal point '.' or ',' and named `int`.  It covers two cases:
  - `[^,]*,`:  Any sequence of characters that is not ',' with ',' accepted as the decimal point (from start of string because of earlier `^`);
  - `[^.]*\.`:  Any sequence of characters that is not a '.' with '.' accepted as the decimal point (from start of string.
- `(?<dec>[^.,]*)`: The sequence after the character accepted as the decimal point, not including it.

We then remove all dots and commas from the integral part.  That works for all cases.
And we recompose the number using the decimal point expected by PHP.

IHMO, the test cases speak for themselves and we can add more of them.
I added '1.023,210.00' and '1,023.210,00' so that you can indicate what the desired results for these would be.

PR #33837, opened 2025-04-14
FIX compute pending amount on debit or credit request

lvessiller-opendsi:
FIX compute pending amount on debit or credit request

On debit request the computed pending amount was incorrect.

**Before**
The calculation of the amount took into account unprocessed requests.
The "unprocessed" status (pdf.traite=0) is only when the request in not in the "llx_demande_lignes" table.
The debit request line is not created in database but some debit request are always in progress (status = 0).

**After**
The calculation of the amount took into account unprocessed and processed requests.
The "unprocessed" status in not the good way to take into account debit request are in progress : a line exist in "llx_demande_lignes" whose status is 0.
When the status is "2" (paid) or "3" (rejected), there is a payment in the linked invoice, so we don't need and the amount is removed from the calculation of remain to pay.

**=> This fix is also for customer credit requests**

### DIFF
--- a/htdocs/compta/facture/prelevement.php
+++ b/htdocs/compta/facture/prelevement.php
@@ -718,13 +718,14 @@ if ($object->id > 0) {
 	// For which amount ?
 
 	$sql = "SELECT SUM(pfd.amount) as amount";
-	$sql .= " FROM ".MAIN_DB_PREFIX."prelevement_demande as pfd";
+	$sql .= " FROM ".$db->prefix()."prelevement_demande as pfd";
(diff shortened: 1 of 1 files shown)


eldy:
Fix is not enough. Because there is 2 cases to manage.
A pending transfer has been request but no  lines exits yet in llx_prelevement_demande_lignes (we must not request another one) + a transfer has been requested and is on the road to the bank ( a line exist with status 0).
I have pushed a more complete fix here in develop b27c3b692e699b16264a7f707977159262941c56
This is a dirty fix because we need to fix this at 3 different places and i just did a copy paste.

lvessiller-opendsi:
Not the good way to fix it.
A better one was proposed on develop branch

PR #33847, opened 2025-04-15
Review of module asset

aspangaro:
Reminder: module in experimental status

- Add 'VALIDATED' status
- Modify process to use available numerator
- Small ajustement (title, agenda, translation)


htdocs/asset/class/asset.class.php
```diff
@@ -1637,4 +1655,139 @@ public function getNextNumRef()
 			return "";
 		}
 	}
+
+	/**
+	 *	Validate bom
```
josett225, line 1660: Hello @aspangaro 
Small string change here bom to asset

PR #33852, opened 2025-04-15
New webhook history database structure

Hystepik:
Database structure for new webhook history


htdocs/install/mysql/tables/llx_webhook_history_extrafields-webhook.sql
```diff
@@ -0,0 +1,22 @@
+-- Copyright (C) 2025		Alice Adminson				<myemail@mycompany.com>
```
eldy, line 1: This table will be filled by system, not by end user, so the extrafields table to add custim field should be useless.
You can remove it.

PR #33854, opened 2025-04-15
FIX: Task time update without author column selected

BenjaminFlr:
When updating a task time, if the column 'author' is not selected, it's impossible to update it because the field is required. But while the author can't be changed on editing it from `time.php`, we should be able to edit without this column.

### DIFF
--- a/htdocs/projet/tasks/time.php
+++ b/htdocs/projet/tasks/time.php
@@ -10,6 +10,7 @@
  * Copyright (C) 2024		MDW						<mdeweerd@users.noreply.github.com>
  * Copyright (C) 2024		Vincent de Grandpré		<vincent@de-grandpre.quebec>
  * Copyright (C) 2024		Solution Libre SAS		<contact@solution-libre.fr>
+ * Copyright (C) 2025		Benjamin Falière		<benjamin@faliere.com>
  *
  * This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


htdocs/projet/tasks/time.php
```diff
@@ -2127,6 +2128,9 @@ function setDetailVisibility() {
 				print '<td class="center nowraponall">';
 				if (($action == 'editline' || $action == 'splitline') && GETPOSTINT('lineid') == $task_time->rowid) {
 					print '<input type="hidden" name="lineid" value="' . GETPOSTINT('lineid') . '">';
+					if (!empty($task_time->fk_user)) {
```
eldy, line 2131: If yoy add this, it mens you accept that any user is allowed to modify any user (by editing htm content).
But rule to define who can modify who is inside the controller part. So a change should be done in controller part to be able to correctly manage the uqe cas when author is not provided.

BenjaminFlr: Ok I'll improve this PR with changes on controller side.

BenjaminFlr:
@eldy I made some new tests, in fact my PR doesn't affect the security more than before while this parameter is only passed when clicking on "save". It's not displayed in the address bar and this parameter can't be changed more easily than before.
Assuming that we sould make a new PR to improve the rights check, which would be another fix. What do you think?

PR #33855, opened 2025-04-15
FIX: partial result on timespent list

BenjaminFlr:
If the column `author` isn't checked, there is only a partial answer while only the current user is selected by default and can't be changed. I propose to fix this by removing this filter if the column isn't checked.


eldy:
This may to introduce a security vulnerability.
can yoy provide a screenshot of what user see when he has just low level permission ?

BenjaminFlr:
I created an user with only `Read projects and tasks (shared projects and projects of which I am a contact).` checked in his rights list.

Currently, this user can only see his list of timespent.

With my code, there is no change, other users' times are not displayed even if there is none selected.

Here is what I see with admin account (alice adminson):
Capture d’écran 2025-04-18 à 09 48 44
Capture d’écran 2025-04-18 à 09 48 58

Here is what I see with the user created without rights to see others' information:
Capture d’écran 2025-04-18 à 09 48 33

PR #33859, opened 2025-04-15
FIX: accountancy bank journal: handle payments made on behalf of sister companies

marc-dll:
# FIX: accountancy bank journal: handle payments made on behalf of sister companies

When the `FACTURE_PAYMENTS_ON_DIFFERENT_THIRDPARTIES_BILLS` setting is enabled (which is not a hidden setting by the way, it can be toggled on `/htdocs/admin/payment.php`, Dolibarr can record payments made by a different company than the one to which the invoice is attributed.

This case is not handled in accoutancy by Dolibarr's bank journal : the amounts are deduplicated among the related thirdparties (in table `llx_bank_url`).

Moreover, the thirdparty shown for the counterparts is always the one doing or receiving the payment, and not the one on the invoice (while the subledger account is correctly).

The only solution is to link with the payment and filter out the superfluous ones. Handling both supplier and customer invoices is done with SQL `IF` statements.

There is one small regression as a result that is not simple to solve in my opinion : there is no more link to the thirdparty in the lines linked to the bank account (those with account 5121 in the exemple below), but this piece of info is not totally relevant in this context and the link will be present in the counterpart lines.

On the following example, we recieve a payment of €225 from `Filiale 2`, €150 for customer invoice `FA2504-0005` made to `Filiale 2` and €75 for customer invoice `FA2504-0004` made to `Filiale 1`, and we make a "normal" payment of €480 to `Fournisseur n°1` for supplier invoice `SI2504-0002` to test if the general case and the supplier case is still working.

## Reproducing case

Before :
Capture d’écran de 2025-04-17 11-53-44

After :
Capture d’écran de 2025-04-17 11-53-35


marc-dll:
Hi @eldy @rycks @lvessiller-opendsi,

I guess you have been on holiday, and I hope you had a good one :)

Can you please make a feedback on this PR ?

Greetings,
Marc.

lvessiller-opendsi:
Could have a check @SylvainLegrand ?

lvessiller-opendsi (dismissed):

marc-dll:
Hi @eldy @rycks,

Sorry to be a bother, but can I please have feedback on this one ?

Greetings,
Marc.

rycks:
@marc-dll i'm sorry but that point is really hyper specific and i don't have any customer with such situation so i have to set up a dolibarr with mother company and reproduce your bug to *understand* what your PR makes ... and i really don't have the time for that :-( 

i could read the code but that's not enough to make sense i mean

hregis:
@marc-dll @rycks
Cette PR me fait peur !
Et dès que je vois un getEntity virgule zéro je me dis que la seule solution pour me calmer c'est de me servir un whisky sec...

FHenry:
@hregis @rycks 
This is not a topic about multi-company, but just about mother third-party (société mére), a core Dolibarr feature.
Please take time to read the test case and reproduce.

rycks (dismissed):
ok

eldy:
My understandingis that this is a fix to return different numbers when option FACTURE_PAYMENTS_ON_DIFFERENT_THIRDPARTIES_BILLS is on.
But i can't see any if (FACTURE_PAYMENTS_ON_DIFFERENT_THIRDPARTIES_BILLS) in code added to manage this option.
As the SQL is complexified for the need of this use case, i think it would be better to add a test 
`if (getDolGlobalString('FACTURE_PAYMENTS_ON_DIFFERENT_THIRDPARTIES_BILLS')`
to enable the new code but only when we are in this use case and keep the "simplest current code" for standard case. The main reason is that this code introduce new JOINS that add a risk of dudplicate lines in asnwer and also a GROUP BY that has a high cost in term of performance and we don't need to pay this price for common use.

lvessiller-opendsi:
> My understandingis that this is a fix to return different numbers when option FACTURE_PAYMENTS_ON_DIFFERENT_THIRDPARTIES_BILLS is on. But i can't see any if (FACTURE_PAYMENTS_ON_DIFFERENT_THIRDPARTIES_BILLS) in code added to manage this option. As the SQL is complexified for the need of this use case, i think it would be better to add a test `if (getDolGlobalString('FACTURE_PAYMENTS_ON_DIFFERENT_THIRDPARTIES_BILLS')` to enable the new code but only when we are in this use case and keep the "simplest current code" for standard case. The main reason is that this code introduce new JOINS that add a risk of dudplicate lines in asnwer and also a GROUP BY that has a high cost in term of performance and we don't need to pay this price for common use.

I understand your point of view but if we put a test with this const in each line where there is a change, the code wil be understandable.

eldy:
> > My understandingis that this is a fix to return different numbers when option FACTURE_PAYMENTS_ON_DIFFERENT_THIRDPARTIES_BILLS is on. But i can't see any if (FACTURE_PAYMENTS_ON_DIFFERENT_THIRDPARTIES_BILLS) in code added to manage this option. As the SQL is complexified for the need of this use case, i think it would be better to add a test `if (getDolGlobalString('FACTURE_PAYMENTS_ON_DIFFERENT_THIRDPARTIES_BILLS')` to enable the new code but only when we are in this use case and keep the "simplest current code" for standard case. The main reason is that this code introduce new JOINS that add a risk of dudplicate lines in asnwer and also a GROUP BY that has a high cost in term of performance and we don't need to pay this price for common use.
> 
> I understand your point of view but if we put a test with this const in each line where there is a change, the code wil be understandable.

Not on each line, but on each portion of code this option has an impact.
In my point of view,  it does not make code less understandable. It is the opposite. It become clear in code which part is common feature and which part of code is the special complex feature. The good example is this pr itself. I am honestly not able to guarantee that the change will not have a regression in default usage as I don t understand the complexity of the new suggested code.
If code is separated in 2, the code when option is off should be same than before so it become easy to see that nothing was changed so easy to guarantee there is no regression. And for the part specific to option on, as the feature seems bugged, it is less important if fix is not completely under control because it will mean we replace a bugged code by another one. When we try to fix if we fail, it is acceptable. But trying to fix and breaking the main use case is not.
Also having separation of use case, even if it is at a lot of different places in code is something that make usually maintenance easier. Think to situation invoice feature. There is a lot of "if" at different places. But this save our life when we need years to make the feature stable, because it guarantee there is no impact or the rest. And wecan make impact analysis on a given option faster too. A search on option loclize all part of code that has special code for it.

So because the change is to fix a hidden option, not yet stable, the fix can be validated only in a maintenance release if we see that code changed is inside a if (option) scope.

PR #33864, opened 2025-04-16
Error urlencode(): Passing null to parameter #1 ($string) of type string is deprecated

### LINKED ISSUE #33863: Error urlencode(): Passing null to parameter #1 ($string) of type string is deprecated (/core/tpl/extrafields_list_search_param.tpl.php:53)
### Bug

PHP reports an error on 
/core/tpl/extrafields_list_search_param.tpl.php:53
when $val is null

### Dolibarr Version

21.0.1

### Environment PHP

8.2

### Environment Database

mariaDB 11.4

- Create a 'Complementary attribute' of String (1 Line) type.
- On the list view, activate de complementary attibute.

Now when a registrer has its complementary attribute empty, the error is logged on DebugBar

You must activate the DegubBar and DebugLogs modules to see the log

### Attached files

_No response_

PsyCrow-code:
FIX #33863 
When $var is null; the evaluation of urlencode($var) generates an error log because of ($var !== '') is true


htdocs/core/tpl/extrafields_list_search_param.tpl.php
```diff
@@ -33,7 +33,7 @@
 			$param .= '&'.$search_options_pattern.$tmpkey.'_endmin='.dol_print_date($val['end'], '%M');
 			$val = '';
 		}
-		if ($val !== '') {
+		if (!empty($val)) {
```
eldy, line 36: What if $val is '0' instead of '' ?
For this case we need to have test true.

PsyCrow-code: So, the correct condition whold be 
($val !== null && $val !== '' && $val !== [])
 to filter null, empty string and empty array

PsyCrow-code (commented):
the complete condiciontion should be 
($val !== null && $val !== '' && $val !== [])

PsyCrow-code:
fixed at pull request #34033

PR #33868, opened 2025-04-16
UIUX Improved ergonomics through UI design

zahwyrix:
# UIUX
In IT conventions, the Cancel button is on the left and the Create button on the right. On Dolibarr, it's currently the other way round, which is rather counter-intuitive when you reflexively click. Correction to improve the GUI while respecting cultural norms.

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -12200,9 +12200,17 @@ class Form
 		if (!empty($morebuttons)) {
 			$buttons[] = $morebuttons;
 		}
+		$mainBtnFirst = getDolGlobalInt(MAIN_BUTTON_POSITION_FIRST_OR_LEFT);
+		
+		if(!$mainBtnFirst){
+			!empty($save_label) ? $buttons[] = $save : '';
+		}
 
 		!empty($cancel_label) ? $buttons[] = $cancel : '';
-		!empty($save_label) ? $buttons[] = $save : '';
+
+		if($mainBtnFirst){
+			!empty($save_label) ? $buttons[] = $save : '';
+		}
 
 		$retstring = $withoutdiv ? '' : '<div class="center">';
 
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -12197,18 +12197,18 @@ class Form
 			'label_key' => 'Cancel',
 		);
 
-		if (!empty($morebuttons)) {
-			$buttons[] = $morebuttons;
-		}
-		$mainBtnFirst = getDolGlobalInt('MAIN_BUTTON_POSITION_FIRST_OR_LEFT');
-
-		if ($mainBtnFirst) {
+		// If MAIN_BUTTON_POSITION_FIRST_OR_LEFT not set, default is to have main action first, then complementary, then cancel at end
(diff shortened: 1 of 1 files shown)


thomas-Ngr:
> In IT conventions, the Cancel button is on the left 

No, there is no convention. Or please provide a source. [This is my source](https://www.nngroup.com/articles/ok-cancel-or-cancel-ok/), in which you can read for example : 

> Sadly, the Windows User Experience Guidelines differ from the Apple Human Interface Guidelines when it comes to the sequence of OK / Cancel buttons:
> 
> - Windows puts OK first
> - Apple puts OK last

zahwyrix:
Apple screen (2025-01-27) :
macos-catalina-restart-modal

Although not all windows are like that, I grant you.
In fact, the majority of users are Windows users.

**There is, of course, a perhaps unofficial but cultural convention (learned in any good computer training course in basic human-machine interface). In Western countries, we read from left to right, cancel left, confirm right.**
Whether or not Dolibarr wants to do this is another matter, but the basic rule is this.

[https://www.nngroup.com/articles/ok-cancel-or-cancel-ok/](url)
[https://medium.com/design-bootcamp/confirmation-button-placement-c32292fab848](url)

ChatGPT :
Habitudes culturelles et de lecture : Dans les cultures où l'on lit de gauche à droite (comme en Europe ou en Amérique du Nord), l'utilisateur s'attend généralement à ce que l'option la plus "négative" ou "réversible" (comme "Annuler") se trouve sur la gauche, tandis que l'option qui conduit à une action positive ou irréversible (comme "Suivant") se trouve sur la droite.

Conventions d'interface : Sur des systèmes comme Windows et macOS, il est courant de voir "Annuler" à gauche et "Suivant" ou "OK" à droite dans les boîtes de dialogue. Cela est aligné avec la tendance à rendre l'interaction aussi intuitive que possible, en exploitant les attentes naturelles des utilisateurs.

Accessibilité et ergonomie : Il est également question de rendre l'interface plus facile à utiliser pour les utilisateurs en fonction de leur orientation mentale (action positive à droite, annulation à gauche). Les boutons sont disposés de manière à minimiser les erreurs, par exemple en réduisant les risques de cliquer accidentellement sur "Annuler" plutôt que sur "Suivant".

eldy:
It is a point of view. Currently Dolibarr follow this convention:
Main action first, alternative or dangerous action after. 
So for languages that are left to right, the cancel or delete action are at end (not priority).
Giving priority on cancel or delete by having them at end looks strange to me. First (as it is currenctly) looks more logic;
Any other point of view ?
May be we can add a hidden option MAIN_BUTTON_POSITION_FIRST_OR_LEFT for this ?

zahwyrix:
I'm going to do it with a constant that will settle all disagreements :)
I understand your point of view but we're not talking about “giving priority”.
In my university course, I learned about order, for the good reason that if you imagine the “course of action”, you go to the right. A bit complicated to explain, but that's the idea. Just as you'll never have a red validation button in France. You can play on the word convention, but it's mostly cultural and therefore has a lot to do with the user experience.

I've just realized that it's not the same in all parts of Windows, two developers don't agree x) 
Maybe it's just me, but when I first started using Dolibarr, I was always being fooled by the cancel button because I had the “Next” reflex on the right of Windows popups (well, I'm also a bit stupid for not reading XD).
I'm going to make do with a constant.

PR #33870, opened 2025-04-16
 NEW Supplier order - Introduce constant ORDER_SUPPLIER_METHOD_BY_DEFAULT to define order method

thomas-Ngr:
# NEW introduce const ORDER_SUPPLIER_METHOD_BY_DEFAULT on supplier order card

On supplier order, there is a select with the order method. There is no way to set a default order method.
However, users may want to have a default method if they always send the order in a certain way.

This PR introduces constant ORDER_SUPPLIER_METHOD_BY_DEFAULT to solve this case.

The value for ORDER_SUPPLIER_METHOD_BY_DEFAULT must be the ID of the method to use in the dict of order methods.

### DIFF
--- a/htdocs/fourn/commande/card.php
+++ b/htdocs/fourn/commande/card.php
@@ -2790,7 +2790,7 @@ if ($action == 'create') {
 
 			// Force mandatory order method
 			print '<tr><td class="fieldrequired">'.$langs->trans("OrderMode").'</td><td>';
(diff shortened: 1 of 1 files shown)


thomas-Ngr:
@eldy I rebased the PR and now CI pass. Can you have another look ? Thanks !

eldy:
We are abandoning step by step the ton of option related to default values (because their are on a "per field" basis, so may need too much options), to replace them with the generic feature "Default values" in home setup, because it is a genetic solution working for all fields.
Can't you try to use it and make a feedback ?

thomas-Ngr:
You are right, following config works : 

image

PR #33874, opened 2025-04-16
NEW - Add ability to display order in proposal list

omogenot:
Add the ability to show a customer order column in the proposal list in order to check that each signed proposal has been converted to a customer order. This would help to find missed proposal conversions.


htdocs/comm/propal/list.php
```diff
@@ -53,6 +53,9 @@
 require_once DOL_DOCUMENT_ROOT.'/core/lib/company.lib.php';
 require_once DOL_DOCUMENT_ROOT.'/comm/propal/class/propal.class.php';
 require_once DOL_DOCUMENT_ROOT.'/projet/class/project.class.php';
+if (isModEnabled('commande')) {
```
frederic34, line 56: `isModEnabled('order')`

htdocs/comm/propal/list.php
```diff
@@ -1097,6 +1116,9 @@
 if ($search_import_key != '') {
 	$param .= '&search_import_key='.urlencode($search_import_key);
 }
+if ($search_commande) {
+	$param .= '&serach_commande='.urlencode((string) ($search_commande));
+}
```
frederic34, line 1121: search instead serach

htdocs/comm/propal/list.php
omogenot: The phan error message is misleading:
```
Warning: list.php: DeprecatedModuleName: Argument 0 function isModEnabled can't have the value "commande" that does not match the expected pattern '/^(?!(?:actioncomm|adherent|adherent_type|banque|categorie|commande|contrat|entrepot|expedition|facture|ficheinter|product_fournisseur_price|product_price|projet|propale|socpeople)$).*/'
```

From what I read 'commande' is part of what is expected. I'm OK to change it to 'order' however, I would like to be sure that it works...

eldy:
Can you provide a screenshot of what happen if there is one order linked to 2 proposals and 2 orders linked to the same proposal ?

omogenot:
> Can you provide a screenshot of what happen if there is one order linked to 2 proposals and 2 orders linked to the same proposal ?

image

Multiple orders linked to a single proposal creates a duplicate proposal lines per order in table list, whereas same order is shown if part of multiple proposals.

However, hopefully, this case shall not occur in normal life 😀

eldy:
> > Can you provide a screenshot of what happen if there is one order linked to 2 proposals and 2 orders linked to the same proposal ?
> 
> image
> Multiple orders linked to a single proposal creates a duplicate proposal lines per order in table list, whereas same order is shown if part of multiple proposals.
> 
> However, hopefully, this case shall not occur in normal life 😀

This occurs in real life i'm afraid, more often that we can think, for several reason (for example a proposal can be done for the year but then it must be converted into 3 orders, one of each coming month in the quarter. There is ton of other reason, like for example because some products will be sent by a different transporter or because the customer ask to split it in 2, half in 2024 and half in 2025 for financial need, and more...). 
With this PR the proposal is shown twice instead of one, and total amount is wrong.
So such a feature must be implemented differently. For exemple if there is several order for one proposal, all orders must be shown separated by a "," in the column order. If there is more than x order (x to be determined), we show instead the number "5 orders...".  This is done for example to show the assigned users of projects in list of projects.

omogenot:
> > > Can you provide a screenshot of what happen if there is one order linked to 2 proposals and 2 orders linked to the same proposal ?
> > 
> > Multiple orders linked to a single proposal creates a duplicate proposal lines per order in table list, whereas same order is shown if part of multiple proposals.
> > However, hopefully, this case shall not occur in normal life 😀
> 
> This occurs in real life i'm afraid, more often that we can think, for several reason (for example a proposal can be done for the year but then it must be converted into 3 orders, one of each coming month in the quarter. There is ton of other reason, like for example because some products will be sent by a different transporter or because the customer ask to split it in 2, half in 2024 and half in 2025 for financial need, and more...). With this PR the proposal is shown twice instead of one, and total amount is wrong. So such a feature must be implemented differently. For exemple if there is several order for one proposal, all ordersmust be show spearated by a "," in the column order. If there is more than x order (x to be determined), we show instead the number "5 orders...". This is done for example to show the assigned users of projects in list of projects.

@eldy 
It is clear that it exists thousands of use cases... I understand your concern(s) and therefore I propose to change the way this would be done. I suggest to show a check mark for the "order" column as it is done for the "billed" column in the orders list. By the way, why not show the same check mark to a new "billed" column if a bill is associated with this proposal.
If you approve this functionality extension, I propose to handle it as I have some short bandwidth this week by creating another PR.

On the other hand I must admit that I do not really understand the logic of the proposal generating a bill, I saw that it exists a `PROPOSAL_ARE_NOT_BILLABLE` constant to hide the "Create Bill" button, in which case the Dashboard page shall not show a "billable" line in the proposal box as this is kind of misleading for the user (as well as its status of 'Billed'). On the other hand, as far as I understand the logic, a proposal that is signed shall be 'converted' in a (or some) customer order(s) and the fact that a signed proposal is not yet converted as a customer order shall be pointed out (hence my original PR proposal) so that no accepted proposal is missed by the sales assistants. I understand it's a different use case. Does it make sense and is there an interest that I try to handle that "Billed" vs. "Ordered" status depending on the `PROPOSAL_ARE_NOT_BILLABLE` constant value? If yes, I would then create yet another PR for that in the coming weeks (depending on my workload) or if someone else wants to tackle it.
Do not hesitate to contact me to let me know.

Olivier.

Issue #33901, opened 2025-04-18
Error filter project for users add bad filter


lmag:
### Bug

Bad filter for project list @hregis 

### Dolibarr Version

21

### Steps to reproduce the behavior and expected behavior

Step 1 : add right on users

![Image](https://github.com/user-attachments/assets/99450046-7af2-4f6c-8ca2-0ea4efe4c219)

Step 2 : add project for this users

![Image](https://github.com/user-attachments/assets/b7d429aa-0c0c-4e67-838d-0af5a4e5adf3)

Step 3 : Dashboard is ok i can see the project

![Image](https://github.com/user-attachments/assets/7c80542d-c584-46a8-9f63-ed039c0b6a73)

Step 4 : Clic on Liste

![Image](https://github.com/user-attachments/assets/b4a1740b-487a-4079-829f-a63effe765fd)

Step 5 : Bad filter

![Image](https://github.com/user-attachments/assets/33defde6-2e94-4395-a2b7-5bd96561a1a6)

Step 6 : Close with the cross and it will be ok

![Image](https://github.com/user-attachments/assets/9a1a17cc-d95e-4b26-94a1-b47ef1d5226b)

lw365:
I tried to reproduce the error, but in my case it worked fine. The project showed up in the user interface. Seems to be already fixed by fix: Dolibarr#33901.  e38b045f1b187ca8e8ba321fc647f5ad3c82bc68

PR #33903, opened 2025-04-18
fix: #33901

FHenry:
To discuss about entity management in list preset


eldy:
@hregis Is it ok for you ?

Issue #33905, opened 2025-04-18
Missing autoreload in field for external contact


lmag:
### Bug

![Image](https://github.com/user-attachments/assets/cef2cd41-7154-44d2-a976-dc9ce1687fd1)

### Dolibarr Version

21

Do like in event

![Image](https://github.com/user-attachments/assets/5775e25a-4be7-4b2a-87eb-e3a246419f38)

PR #33907, opened 2025-04-18
Fix: On User edit page when Field External User : select company => Pre-filter contact list

FHenry:
fix: #33905


eldy:
@FHenry Are you sure on the number of the fix. Seems not related ?

FHenry:
> @FHenry Are you sure on the number of the fix. Seems not related ?

I've just updated the issue number (invert 9 and 0)

PR #33909, opened 2025-04-18
fix: User Card  - UI/UX - Move field contact under type to be more logical

### LINKED ISSUE #33906: UI/UX - Move field contact under type to be more logical
### Feature Request

Image

### Use case

_No response_

### Suggested implementation

_No response_

### Suggested steps

_No response_

FHenry:
fix: #33906


eldy:
Not necessarly more logic.
Because all fields "Link to" are at end of the card on other card like member card.
Must think of it...

PR #33919, opened 2025-04-18
FIX intervention API get contacts ids

W1W1-M:
# FIX intervention API get contacts ids
contacts_ids are empty when getting interventions from API
This fix also adds a parameter to indicate if we want external, internal or third party ids to be returned
Also fixed some PHPDoc

Is this fix considered NEW or can it backported to v18 ?


eldy:
Due to the message 
"also adds a parameter"
it is in develop.

htdocs/fichinter/class/api_interventions.class.php
```diff
@@ -93,6 +94,7 @@ public function get($id)
 		}
 
 		$this->fichinter->fetchObjectLinked();
+		$this->fichinter->contacts_ids = $this->fichinter->liste_contact(-1, $contact_type, 1);
```
eldy, line 97: Becausecthe contacts needs a subquery it should be done onlybif expected. So i suggest to have this parameter set to empty string '' by default. The load should be done only if param is provided.

W1W1-M: Done

W1W1-M:
> Due to the message "also adds a parameter" it is in develop.

OK

Code updated to make default empty string

PR #33927, opened 2025-04-19
Fix the `accountancy report by project`

ymollard:
This PR follows https://github.com/Dolibarr/dolibarr/pull/31494 that has been manually merged.

# FIX|Fix Show accountancy-by-project report into menus
That was forgotten from the original PR https://github.com/Dolibarr/dolibarr/pull/31494
As a consequence, the report could not be accessed.

# FIX|Fix Restore the normal URL link for details
@eldy, you have commented part of the code here https://github.com/ymollard/dolibarr/commit/66f0ffd278abc48e196fa4846bfa465a886103de and that broke the URLs that now lead to a 404.
What's the reason for these comments? How can we restore the normal URLs?


eldy:
Don't remember why i comment this line but i remember why menu was not added:
I experienced several bugs with this report.
Yesterday i made several fixes in the develop branch to fix them including the link.
Can you check they works for you ?

ymollard:
Yes it works (no observable difference in the output for my usecase)

eldy:
So i close this ticket.

PR #33938, opened 2025-04-21
if localtaxes_type is empty no array return

defrance:
fail on create contract line (generate warning)


defrance:
I précise, it's from apî create call....

PR #33939, opened 2025-04-22
NEW get contacts for interventions API

W1W1-M:
# NEW get contacts for interventions API
Added new endpoint to get an intervention's full list of contacts 

new intervention api get contacts


htdocs/fichinter/class/api_interventions.class.php
```diff
@@ -534,4 +534,39 @@ private function _validateLine($data)
...
+			throw new RestException(404, 'Intervention not found');
+		}
+		if (!DolibarrApi::_checkAccessToResource('fichinter', $this->fichinter->id)) {
+			throw new RestException(403, 'Access not allowed for login '.DolibarrApiAccess::$user->login);
+		}
+		$contacts = $this->fichinter->liste_contact(-1, $type, 0, $code, $status);
+		if (empty($contacts)) {
+			throw new RestException(404, 'No intervention contacts found');
+		} else {
+			return $contacts;
```
eldy, line 569: Can you try
return $this->_cleanObjectDatas($contacts);
to have similar code than for sale orders.

W1W1-M: Code updated

eldy:
In develop, the api get already has been modified to match other get api on other object. It means now you can get the contact by calling the get api and adding the param $contact_list to 1 making this PR obsolete.

W1W1-M:
> In develop, the api get already has been modified to match other get api on other object. It means now you can get the contact by calling the get api and adding the param $contact_list to 1 making this PR obsolete.

OK, thanks for input.

PR #33947, opened 2025-04-22
Fix monthly and period total calculations in result.php

ATM-NicolasV:
# FIX|Fix #
Fix monthly and period total calculations in result.php

Corrected the calculation of monthly turnover ($resultM) by using month-specific date ranges ($month_start, $month_end) in getSumDebitCredit, ensuring distinct monthly results. Fixed period total ($resultN) to only sum $resultM for months within the user-selected date range ($date_start to $date_end). Added debug logs to verify date ranges and results.

before : 
image

after : 
image


JamBalaya56562:
This looks already fixed on develop.

The monthly and period totals were corrected by #35083 ("FIX: Correct the calculation of the amount of the current period between the period provided", commit a69ba8885713, merged 2025-09-04 and then merged forward to develop). It adds, in `result.php`, the period restriction this PR is aiming for:

```php
if (($yeartoprocess == $start_year && ($k + 1) >= $date_startmonth && $k < $date_endmonth) ||
    ($yeartoprocess == $start_year + 1 && ($k + 1) < $date_startmonth)
) {
    ...
    $resultN += $resultM;
}
```

so `$resultN` only sums the months inside the selected period (the goal of the `if ($month_start >= $date_start ...)` guard in this PR).

The per-month `$resultM` is also already correct without the `$month_start`/`$month_end` change: `AccountancyCategory::getSumDebitCredit()` already filters on its `$month`/`$year` arguments, which are "stronger than filter date_start/date_end" (see the comment in that method), so it returns the month-specific sum.

Finally, the `(int)` cast this PR removes from `getSumDebitCredit($cpt['account_number'], ...)` has since been removed on develop as well (via #35658 / #39208) — which is why this PR now shows as conflicting.

So this PR looks superseded and can probably be closed.

PR #33955, opened 2025-04-23
FIX: In the caluculation of the structured communication, the modulo must have 2 digits

kkhelifa-opendsi:
In the caluculation of the structured communication, the modulo must have 2 digits

+++3digits/4digits/5digits+++
+++202/5030/01408+++ and not +++202/5030/0148+++


eldy:
I am still not able to edit/fix the PR.

eldy:

PR #33956, opened 2025-04-23
20.0 gpaoplus bom of extrafields

ATM-NicolasV:
Ajout de la gestion du transfert des champs complémentaires d’une BOM vers un OF.


eldy:
This PR still has CI + conflicts errors and is not open to be fixed by third party developers so we can't fix it. A lot of code may have been validated since the PR and the PR may be obsolete, or too difficult to merge due to other changes. Best choice is too cancel it to avoid to keep PR with too low chances to be validated.

PR #33958, opened 2025-04-23
FIX dashboard: restrict proposal stats visibility for external users

f-hoedl:
External users (with socid set) were previously able to see all commercial proposals in the home dashboard blocks (e.g., open/signed proposals), ignoring entity-based visibility restrictions. This commit ensures that only proposals linked to the user’s associated third party are counted and shown.

Also applies proper filtering for internal users without global 'read all clients' rights.

Fixed view:
image

### DIFF
--- a/htdocs/comm/propal/class/propal.class.php
+++ b/htdocs/comm/propal/class/propal.class.php
@@ -3485,7 +3485,12 @@ class Propal extends CommonObject
 		$clause = " WHERE";
 
 		$sql = "SELECT p.rowid, p.ref, p.datec as datec, p.fin_validite as datefin, p.total_ht";
-		$sql .= " FROM ".MAIN_DB_PREFIX.$this->table_element." as p";
+		if (empty($user->socid) && !$user->hasRight('societe', 'client', 'voir')) {
+			$sql .= " FROM ".MAIN_DB_PREFIX.$this->table_element." as p";
+			$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."societe_commerciaux as sc ON p.fk_soc = sc.fk_soc";
+		} else {
+			$sql .= " FROM ".MAIN_DB_PREFIX."propal as p";
+		}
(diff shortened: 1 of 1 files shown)


htdocs/comm/propal/class/propal.class.php
```diff
@@ -3485,14 +3485,27 @@ public function load_board($user, $mode)
 		$clause = " WHERE";
 
 		$sql = "SELECT p.rowid, p.ref, p.datec as datec, p.fin_validite as datefin, p.total_ht";
-		$sql .= " FROM ".MAIN_DB_PREFIX.$this->table_element." as p";
+		if (empty($user->socid) && !$user->hasRight('societe', 'client', 'voir')) {
+			$sql .= " FROM ".MAIN_DB_PREFIX.$this->table_element." as p";
```
eldy, line 3489: Condition on societe_commerciaux must be added into the WHERE by doing a
		$sql .= " AND EXISTS (SELECT sc.fk_soc FROM ".MAIN_DB_PREFIX."societe_commerciaux as sc WHERE sc.fk_soc = p.fk_soc AND sc.fk_user = ".((int) $user->id).")";

htdocs/comm/propal/class/propal.class.php
```diff
@@ -3485,14 +3485,27 @@ public function load_board($user, $mode)
...
+		}
 		$sql .= $clause." p.entity IN (".getEntity('propal').")";
 		if ($mode == 'opened') {
 			$sql .= " AND p.fk_statut = ".self::STATUS_VALIDATED;
 		}
 		if ($mode == 'signed') {
 			$sql .= " AND p.fk_statut = ".self::STATUS_SIGNED;
 		}
+
+		if (empty($user->socid) && !$user->hasRight('societe', 'client', 'voir')) {
```
eldy, line 3502: This condition seems already added lin 3511

PR #33965, opened 2025-04-24
height tabs

zahwyrix:
I tried to remove this line on my Dolibarr and it didn't make any difference, it doesn't seem useful.
What's more, it causes a bug on some pages (notably the supplier price) where the tabs take up the whole page (100%) and the banner comes underneath but you can't even see it without scrolling.


eldy:
Can you provide a screenshot of the result when keeping the line ?

zahwyrix:
The banner is below by scrolling
In V21 and after validating a cost price
image

PR #33973, opened 2025-04-24
NEW - Add orders and invoices columns in Proposal list

omogenot:
Add the ability to set Orders and Invoices columns in the proposal list. If a proposal is linked to one or more orders, a check mark is displayed and a tooltip window will popup with the list of related orders. Same applies for Invoices column.


omogenot:
Useless...

PR #33979, opened 2025-04-25
New AI extrafield filler

Hystepik:
Add a new AI feature to fill extrafield  wit AI generated values.

ai/admin/custom_prompt.php display
image

Exemple of a extrafield fille by AI
image


eldy:
Expected feature is not exactly this but we are near the goal.
The next step is now to move the AI prompt into the setup of the extrafields instead of popup :
On page to setup an extrafield we must be able to enter an AI prompt like this
image

The textarea for the AI prompt must be saved into column "aiprompt" of table llx_extrafields (field is already present in database but not used yet).

And then on the page of thirdparty card, on the magic picto, we don't need anymore to have the popup to show the prompt. 
We can directly have into the getSectionForAIEnhancement()
a handler "onlick on magic picto" that call the AI with the instruction that is the prompt from extrafield (we can add a parameter $prompt to getSectionForAIEnhancement() when the prompt is already known to give it after variable substitution when we make the rendering of the page.
In this case, when we click on picto, we don't have to show the popup and we can make the call directly with the prompt.
For example, when on card of thirdparty "Customer in Bordeaux" we knwo the promptfor the field will be 
`"What is the distance of Paris from __THRIDPARTY_TOWN__"`
that will be converted by the php page into
`"What is the distance of Paris from Bordeaux".`

This prompt will be provided to getSectionForAIEnhancement() so it can make the AI call directly with ne need of popup.

Another exemple of use case:
"What is the SIRET of company __THIRDPARTY_NAME__ ?"

PR #33981, opened 2025-04-25
FIX#33963 - Enabling EmailCollector foreign key field to allow mandatory parameter saving

### LINKED ISSUE #33963: Bug: Email-Collector Filters POSTing 'Null' Foreign Key at admin/emailcollector_card.php on Clean install
https://github.com/Dolibarr/dolibarr/blame/ec2a1928bab0a8c830815046a5da25a7b83ff049/htdocs/emailcollector/class/emailcollectorfilter.class.php#L61C5-L61C5

On trying to modify filters on an email-collector, the following foreign key assosciation is failing and so the module is unable to function.

`Cannot add or update a child row: a foreign key constraint fails (`digitbk1_doli753`.`llxtj_emailcollector_emailcollectorfilter`, CONSTRAINT `fk_emailcollectorfilter_fk_emailcollector` FOREIGN KEY (`fk_emailcollector`) REFERENCES `llxtj_emailcollector_emailcollector` (`rowid`))`

The current workaround is to create the records in the database directly.

vdegrandpre:
# FIX|Fix #33963 Exception when adding filter and actions to EmailCollector

There was an issue adding related objects to EmailCollector. Because of disabled fields, CommonObject::setSaveQuery() do not get mandatory (non null) fk_emailcollector table row data. This fix enables both for EmailCollector Filter and Actions objects.


zacharymelo:
I'm not seeing this fix in either of the subsequent updates. Is there something missing before it can be included? I had Vincent prepare this fix so that our 20.0.1 install could continue to have a functional email collector. 

Am I missing something?

PR #33987, opened 2025-04-26
NEW: attendee status "Showed Up" - to ensure that only one attendee gets in on a ticket

JonBendtsen:
# NEW attendee status Used - to ensure that only one attendee gets in on a ticket
This PR is a subset of #33894 and provides a new status used for event attendees. The reason goal is to make it easier to ensure that any attendee is only let in once by marking their ticket used right after checkin.

Once this PR https://github.com/Dolibarr/dolibarr/pull/33983 has been merged, there should only be the new status USED


JonBendtsen:
looks like this 
image

Issue #23763, opened 2023-02-03
Notes are not stored in EMailing module


edupulpillo:
### Feature Request

It will be very nice to have NOTES available in each emailing. Now available in v19.0.2
Will be very useful for remindings about each emailing, and to take some notes if any of the contacts call perform an easy following notes, etc

![image](https://user-images.githubusercontent.com/7486607/216566064-e5a19908-6416-4daa-beb3-a9c9d4b91622.png)

### Use case

Every time a emailing is setup.

### Suggested implementation

Add the NOTES sheet to each emailing recordf.

edupulpillo:
This feature really is interesting to complete the functionality of emailing s, and take notes about the following of it

edupulpillo:
The functionality is included and present in v19.0.2, but there is a bug and the notes are not stored, or not visible as is explained in this issue #31047

edupulpillo:
The problem is fixed in V22

PR #33994, opened 2025-04-27
Fix #31047 #23763 mailing.class.php do not retrieve the public and private notes

### LINKED ISSUE #31047: The notes of mass emailing tool are not retrieved (Undefined array key "" in /comm/mailing/note.php )
### Bug

The mass emailing notes are not stored.
Is possible to edit the notes and included text in the notes filed, but after record it, changing the option in dolibarr and returning to the same note, the notes field appears empty.

error.log PHP Warning:  Undefined array key "" in /public_html/comm/mailing/note.php on line 52

### Dolibarr Version

19.0.2, 20.0.4

### Environment PHP

8.1.29

### Environment Database

MySQL or MariaDB 10.6.19-MariaDB

1. Select in Dolibarr one emailing in the menu Tools/Emailing List
2. Edit  "Note" of the emailing and add some content and store it
3. Select the "Emailing" Sheet of the emailing
4. Return to the same "note" sheet of the emailing selected and observe that is empty, instead with the content stored as should be

error.log PHP Warning:  Undefined array key "" in /public_html/comm/mailing/note.php on line 52

### Attached files

_No response_

edupulpillo:
# FIX #31047 #23763 mailing.class.php do not retrieve the public and private notes
The public and private notes of the emailing tool are stored but not retrieved in the fields note_public & note_private of the table llx_mailing of the database because are missing in the class /comm/mailing/class/mailing.class.php 
After this commit adding both fields to the class everything works like a charm.

### DIFF
--- a/htdocs/comm/mailing/class/mailing.class.php
+++ b/htdocs/comm/mailing/class/mailing.class.php
@@ -73,12 +73,12 @@ class Mailing extends CommonObject
 	public $evenunsubscribe;
 
 	/**
-	 * @var	string|null Text content for public notes
+	 * @var	string Text content for public notes
 	 */
 	public $note_public;
 
 	/**
-	 * @var	string|null Text content for private notes
+	 * @var	string Text content for private notes
 	 */
 	public $note_private;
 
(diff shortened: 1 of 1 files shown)


edupulpillo:
I suggest you squash the 8 commits before merge it. 
Sorry but it seems that i do not have the rights to do it myself, i do not find the option to do it myself.

PR #33995, opened 2025-04-27
shipping date_create not well used 

defrance:
use date_creation fields on create instead now()
display creation date on stat instead of date validation


defrance:
Update PR, I have added a select type of date to allow display info on specific date (creation, valid, shipping ...)

PR #33998, opened 2025-04-27
Fix : wrong access check on api to add categorIE

atm-maxime:
In contact and thirdparty api to add a category, the access check to resource was done for feature "**category**" but the **checkUserAccessToObject** function is checking feature "**categorie**".

When multicompany is not active, the check was never done, otherwise the check was returning false because llx_category does not exist.

### DIFF
--- a/htdocs/societe/class/api_contacts.class.php
+++ b/htdocs/societe/class/api_contacts.class.php
@@ -465,7 +465,7 @@ class Contacts extends DolibarrApi
 		if (!DolibarrApi::_checkAccessToResource('contact', $this->contact->id)) {
 			throw new RestException(401, 'Access not allowed for login '.DolibarrApiAccess::$user->login);
 		}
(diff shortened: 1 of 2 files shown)


htdocs/societe/class/api_contacts.class.php
```diff
@@ -465,7 +465,7 @@ public function addCategory($id, $category_id)
 		if (!DolibarrApi::_checkAccessToResource('contact', $this->contact->id)) {
 			throw new RestException(401, 'Access not allowed for login '.DolibarrApiAccess::$user->login);
 		}
-		if (!DolibarrApi::_checkAccessToResource('category', $category->id)) {
+		if (!DolibarrApi::_checkAccessToResource('categorie', $category->id)) {
```
eldy, line 468: Because future and followikg version are using "category", we must keep to use this key zs it is the target to switch to english keys.
Instead we must backport the same way to fix that done into recent version, so adding a compatibility test inside the function.

PR #34037, opened 2025-04-30
NEW : add api to list enable/disable modules

defrance:
This feature allows you to list the modules present in a Dolibarr application and their status, and to enable/disable them via APIs.

My goal is to be able to configure and populate a Dolibarr application with data after installing it, so that it can then be tested without having to create the necessary amount of data.


JonBendtsen:
revived here https://github.com/Dolibarr/dolibarr/pull/35718

JonBendtsen:
Given that  https://github.com/Dolibarr/dolibarr/pull/35718 has been merged, this MR can be closed @defrance

PR #34090, opened 2025-05-10
NEW Add option to create simple shipment of non origin

### LINKED ISSUE #22943: Standalone "Shipments" to create quick shipment sheets not linked to any order
### Feature Request

A Standalone "Shipments" to create quick shipment sheets not linked to any order is missing. 

### Use case

It is in the nature of business that we need to ship stuff not related to any order. Now we need to create a shipment sheet using Word or similar. This creates an overhead as Dolibarr users would like to use Dolibarr as the only tool.

### Suggested implementation

From the "Shipments" menu add "Quick Shipments (or similar)" menu 

### Suggested steps

_No response_

sonikf:
related to #31652
fix #22943


sonikf:
Hi @eldy 
Just realized you have authored beta.
I'm sorry i am late for this but  we really need this for v. 22
Unfortunately no API was given for e- invoicing like in our MyData module and sadly many users migrated to proprietary platforms so in order not to have to abandon Dolibarr, 2 months ago 15 forum users we created a group and a 'GIFF' as you say, to contract MyData  dev also part of the group to build a module bridge with a Greek certified peppol provider.
I've created this for v. 21 because mandate for e- invoicing starts 1/7/2025 and we will have to work with modified core sources anyway...

sonikf:
Hi @eldy
Can you please tell me if I'm on the right path with this?
If not how would you like to approach this?
Right now we are acting(Greece) as a test bed for things to come in the rest of EU
Next is e-reception which is also developed same way as shipments, catering one use case strictly dependent on order but not inline with rest core modules.

eldy:
> Hi @eldy
> Can you please tell me if I'm on the right path with this?
> If not how would you like to approach this?
> Right now we are acting(Greece) as a test bed for things to come in the rest of EU
> Next is e-reception which is also developed same way as shipments, catering one use case strictly dependent on order but not inline with rest core modules.

I think it looks good, but because develop is frozen for v22, i will be able to analyze more seriously for merge after release only.

sonikf:
Great, thank you very much that's all I needed for now.
I'll try with reception next, early this time so hopefully both make it for v. 23.

sonikf:
Hi @eldy 
Can you review please? 
We really need this for v. 23, e-shipment mandate has already started and will be compulsory for everyone in Greece from 1/12/2025.
I would like also to submit pr for standalone reception next, built the same way as this, and i think it will be better to have your feedback on this first.
Thank you

eldy:
> Hi @eldy 
> Can you review please? 
> We really need this for v. 23, e-shipment mandate has already started and will be compulsory for everyone in Greece from 1/12/2025.
> I would like also to submit pr for standalone reception next, built the same way as this, and i think it will be better to have your feedback on this first.
> Thank you

PR is in stand by because it is too large making difficult to analyse to know if there is no regression with other works. Can you split it into several pr smaller so we can progress on this ?

sonikf:
Yes of course, from where should i start?

PR #34095, opened 2025-05-11
adding API endpoint GET categories/label/{label}

JonBendtsen:
# NEW *API endpoint GET categories/label/*
This PR adds a new endpoint which can get a category based on the label

label_categories_not_id_get


eldy:
As label is not unique, API on categories must use the ID.

PR #34096, opened 2025-05-11
new API endpoint put categories/label/{label}

JonBendtsen:
# NEW API endpoint PUT categories/label/
This PR adds a new endpoint which can change a category based on the label - you can even change the label.

label_categories_not_id_put_top

label_categories_not_id_put_bottom


eldy:
As label is not unique, API on categories must use the ID.

PR #34100, opened 2025-05-11
adding API endpoint DELETE categories/label/{label}

JonBendtsen:
# NEW API endpoint DELETE categories/label/
This PR adds a new endpoint which can delete a category based on the label.


eldy:
Adding api that allows to make actions based on a key that lay nit be unique is not a good idea.
It is better to call an api to get the id and then call the api with the unique id if answer is unique.

Delete api must always be called by id to avoid trouble with duplicate for different categories or duplicate because of multicompany use.

JonBendtsen:
> # NEW API endpoint DELETE categories/label/
> This PR adds a new endpoint which can delete a category based on the label.

yeah I can certainly see that @eldy. I kind of expected that the label was unique **AND** I when saw this line in the fetch function I kind of expected that it was planned to one day code fetch based on a label - so I coded it, because I kind of need it for my event configuration file which has a category by label, not by id.

https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/categories/class/categorie.class.php#L444

```
	public function fetch($id, $label = '', $type = null, $ref_ext = '')
```

PR #34112, opened 2025-05-12
NEW: add hooks on societe consumption page

jyhere:
Add all necessary hooks to modify query and add more columns on the societe consumption page


hregis:
@jyhere great but don't let `$parameters` with no values between each call !

hregis:
@jyhere great for parameters.... But... Look around the hook... No need to have a value of a parameter for this hook? 😉😘

PR #34117, opened 2025-05-12
adding API endpoint GET categories/label/{label}/objects

JonBendtsen:
# NEW API endpoint GET categories/label/{label}/objects
This PR adds a new endpoint which will get objects in a category based on the label of the category


eldy:
Same trouble than for delete. Key for targeted category must be id of category.

PR #34131, opened 2025-05-13
FIX update line minimum price in propale

leninrivas:
# FIX|Fix #[*FIX update line minimum price in propale*]
[*FIX Variable $price_base_type in minimum price in propale*]


htdocs/comm/propal/card.php
```diff
@@ -1931,7 +1931,7 @@
 			$res = $product->fetch($productid);
 
 			$type = $product->type;
-			$$price_base_type = $product->price_base_type;
+			$price_base_type = $product->price_base_type;
```
eldy, line 1934: I pushed a fix recently that set $price_base_type at line 1920.
It should be enough to fix your trouble.
Can you confirm that with it we can close this PR (It is better to detect price base on what field user has fill instead of default value of product. default value of product is used to fill user, but then we must use the field fill by user).

leninrivas: Alright

PR #34151, opened 2025-05-14
FIX : usage of public member registration form (Multiadmin context)

geourjoa:
Hi ! 

I discovered problem in the member registration form in multicompany context.

For example I have entity 1 and 2 :

If I'm not connected and go to **url/public/members/new.php?entity=2**. No problem
If I'm connected on entity 1 and try to go to  **url/public/members/new.php?entity=2**, form is generated with entity 1 config and not 2 as expected.


geourjoa:
Thanks @eldy for the simplification

PR #34155, opened 2025-05-15
add invoice pie chart (like other page)

defrance:
# Instructions
*Please:*

[*Long description*]

[*Long description*]

[*Long description*]

[*Long description*]

[*Long description*]

[*Long description*]

### DIFF
--- a/htdocs/compta/facture/index.php
+++ b/htdocs/compta/facture/index.php
@@ -72,6 +72,13 @@ if ($tmp) {
 	print $tmp;
 	print '<br>';
 }
+
+$tmp =	print getCustomerInvoicePieChart();
+if ($tmp) {
+	print $tmp;
+	print '<br>';
+}
+
 $tmp = getCustomerInvoiceDraftTable($maxDraftCount, $socid);
 if ($tmp) {
 	print $tmp;
--- a/htdocs/compta/index.php
+++ b/htdocs/compta/index.php
@@ -125,6 +125,8 @@ if (isModEnabled('invoice')) {
 if (isModEnabled('fournisseur') || isModEnabled('supplier_invoice')) {
 	print getNumberInvoicesPieChart('suppliers');
 	print '<br>';
+	print getCustomerInvoicePieChart();
+	print '<br>';
 }
 
 if (isModEnabled('invoice')) {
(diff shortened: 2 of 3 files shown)


eldy:
This chart "count of invoices per status open/close" was removed few version agao and was replaced with a chart "count per delay invoices of open invoices" because the chart per status has no sense : When time increase, the number of invoice closed will always increased indefinitely, so nb in chart of open invoice will finally hide completely the number of draft or open invoices hidden because values will be too low compared to closed one, and graphic chart will be same for everybody, a simple round with closed invoices and a simple bar for draft and open invoices.

Pb is relatively similar with other object, but for other object, we have more satus, so it was possible to remove the status "Closed" that has no sense, and we can show only open status keeping more than 1 value, this is why graph was not changed for other objects.

defrance:
I quite agree with your analysis, but this view introduces an inconsistency.
Other similar charts for orders, proposals, contracts and interventions have exactly the same problem, namely a status that ends up being a "garbage collector."

A solution has been found for tickets, where a year filter can resolve this issue.

That's why I suggest you first accept my PR, then update all the charts to reflect the ticket charts and achieve both visual and functional consistency.

--- 
Je suis assez d'accord avec ton analyse mais cette vision apporte une incohérence 
Les autres graphiques similaire au niveau des commandes, proposition, contrat et interventions ont exactement le meme soucis à savoir un statut qui finit par etre un "garbage collector".

Une solution a été apporté au niveau des ticket ou un filtre sur l'année permet de régler ce soucis.

C'est pour cela que je te propose d'accepter dans un premier temps mon PR puis de faire évoluer l'ensemble des graphiques dans le sens de celui des tickets et avoir un cohérence a la fois visuelle et fonctionnelle

defrance:
Another note: we have a toggle for statuses that allows you to hide them with a single click.
I suggest we leave this PR pending and discuss at the next devcamp the best way to manage the display of this information harmoniously.

eldy:
If you still need this graph i suggest to introduce it as a hidden feature.
Bad consistency is a concern if we can have consistency, but here we are speaking of 2 situations that are different : For proposal, order there is several "OPEN"status, so we can report a chart per status (a fixh shoul dbe done to not have the "CLOSED" status checked by default. For invoice, there is just one "OPEN" status, so chart per status has no sense.

PR #34159, opened 2025-05-15
FIX|fix Webportal remain to pay column

Expresion:


eldy:
I have modified your suggestion to keep the rule, entity must be 1 at anytime if unknown or can't be guess.
Thanks to check it is still ok after.

Expresion:
Yes, Work

PR #34168, opened 2025-05-16
Coreenabled doesn't enable everything globally.

### LINKED ISSUE #34167: Enabling a "core_enabled" module always inserts the "module parts" using the current entity.
### Bug

My understanding is that setting the "core_enabled" attribute of a Module class to true should enable the module globally. However, this is not the case (or at least is very inconsistent), as the "module parts" are still inserted using the current entity.

This issue is similar to [Dolibarr##24375](https://github.com/Dolibarr/dolibarr/issues/24375).

### Dolibarr Version

Since #91bf93a6s (2009)

### Environment PHP

_No response_

### Environment Database

_No response_

1. Create a module
3. In your module class, define `public $core_enabled = 1;`
5. Enable the module

The module should now be enabled globally (entity 0) whereas the modules parts are enabled locally.
Image

### Attached files

_No response_

EnjoyFelix:
# FIX|Fix [#34167](https://github.com/Dolibarr/dolibarr/issues/34167) - Enabling a "core_enabled" module inserts the "module parts" using the current entity

See [Dolibarr#34167](https://github.com/Dolibarr/dolibarr/issues/34167)

### DIFF
--- a/htdocs/core/modules/DolibarrModules.class.php
+++ b/htdocs/core/modules/DolibarrModules.class.php
@@ -2483,7 +2483,7 @@ class DolibarrModules // Can not be abstract, because we need to instantiate it 
 					}
 				}
 
-				$entity = $conf->entity; // Reset the current entity
+				$entity = ((!empty($this->always_enabled) || !empty($this->core_enabled)) ? 0 : $conf->entity); // Reset the current entity
 				$newvalue = $value;
 
 				// Serialize array parameters
@@ -2565,7 +2565,7 @@ class DolibarrModules // Can not be abstract, because we need to instantiate it 
 				if (is_array($value) && isset($value['entity'])) {
 					$entity = $value['entity'];
 				} else {
-					$entity = $conf->entity;
+					$entity = ((!empty($this->always_enabled) || !empty($this->core_enabled)) ? 0 : $conf->entity);
 				}
 
 				$sql = "DELETE FROM ".MAIN_DB_PREFIX."const";


hregis:
@EnjoyFelix entity = 0 can be define with "data" array !!!

```
// Serialize array parameters
				if (is_array($value)) {
					// Can defined other parameters
					// Example when $key='hooks', then $value is an array('data'=>array('hookcontext1','hookcontext2'), 'entity'=>X)
					if (isset($value['data']) && is_array($value['data'])) {
						$newvalue = json_encode($value['data']);
						if (isset($value['entity'])) {
							$entity = $value['entity'];
						}
					} elseif (isset($value['data']) && !is_array($value['data'])) {
						$newvalue = $value['data'];
						if (isset($value['entity'])) {
							$entity = $value['entity'];
						}
					} else { // when hook is declared with syntax 'hook'=>array('hookcontext1','hookcontext2',...)
						$newvalue = json_encode($value);
					}
				}
```

hregis:
@EnjoyFelix you can not force entity = 0 here !!!!

hregis:
@EnjoyFelix 

example of use for "$this->module_parts" : 

```
'css' => array(
    'data' => '/multicompany/css/multicompany.css.php',
    'entity' => '0'
),
```

EnjoyFelix:
I was aware of this and still made the request. When you define "core_enabled", you expect the default value to be 0. The fact that it reverts to the current entity is inconsistent.

hregis:
@EnjoyFelix it is better to set entity 0 in the settings than to force everything to 0 because there may be cases where entity 0 is not desirable for some settings

PR #34169, opened 2025-05-16
QUAL harmonization between statistical sections

zahwyrix:


htdocs/compta/facture/stats/index.php
```diff
@@ -414,9 +414,9 @@
 	print '<td align="center"><a href="'.$_SERVER["PHP_SELF"].'?year='.$year.'&amp;mode='.$mode.($socid > 0 ? '&socid='.$socid : '').($userid > 0 ? '&userid='.$userid : '').'">'.$year.'</a></td>';
 	print '<td class="right">'.$val['nb'].'</td>';
 	print '<td class="right opacitylow" style="'.($greennb ? 'color: green;' : 'color: red;').'">'.(!empty($val['nb_diff']) && $val['nb_diff'] < 0 ? '' : '+').round(!empty($val['nb_diff']) ? $val['nb_diff'] : 0).'%</td>';
-	print '<td class="right"><span class="amount">'.price(price2num($val['total'], 'MT'), 1).'</span></td>';
+	print '<td class="right">'.price(price2num($val['total'], 'MT'), 1).'</td>';
```
eldy, line 417: As the css "amount" was introduce to be the syle for any total amount in chart or graph, we should keep it and add it when it is not present instead of removing it.

PR #34187, opened 2025-05-18
New : create bill from contract use only enabled lines

defrance:
# New : create bill from contract use only enabled lines
The purpose of this feature is to avoid billing for services that are not active at the contract level.
We use a contract parameter.


hregis:
@defrance why not use just ? : 

`print ajax_constantonoff('CONTRACT_BILL_ONLY_ACTIVE_LINE');`

and

`if (getDolGlobalInt('CONTRACT_BILL_ONLY_ACTIVE_LINE')) {`

Issue #34197, opened 2025-05-19
Add hook 'createFrom' for supplier invoices


glu000:
### Feature Request

Add hook 'createFrom' for supplier invoices

PR #34198, opened 2025-05-19
Add hook 'createFrom' for supplier invoice #34197

glu000:
NEW #34197 Add hook 'createFrom' for supplier invoice


glu000:
@eldy it seems that phan / phpstan is broken as it shows an error in a file I haven't modified

PR #34204, opened 2025-05-20
NEW add totalization on ticket graph

defrance:


defrance (commented):
move 		$totalnb = 0; upper

PR #34206, opened 2025-05-20
FIX|Fix # Ticket module: display issue with HTML links in header/footer configuration

atm-corentin:
Problem encountered (Ticket module)

In the ticket module, when HTML links are used in the message header or footer configuration fields, a display issue occurs for private messages and for email sending above the ticket form.

If these fields contain an HTML link, and the value is injected into a `<input type="hidden">`, the presence of the HTML tag breaks the markup: everything after the link is incorrectly displayed above the form, resulting in misplaced content and incorrect rendering (truncated links, shifted display, broken HTML structure, etc.).

Solution implemented

The implemented solution is to properly decode HTML entities (using ) when retrieving the signature from the hidden input in the ticket module.

This now allows the original HTML content (including links) to be correctly rendered in both forms and emails, without disrupting the page’s structure or display `htmlspecialchars_decode`

### DIFF
--- a/htdocs/admin/ticket.php
+++ b/htdocs/admin/ticket.php
@@ -231,7 +231,7 @@ if ($action == 'updateMask') {
 		$error++;
 	}
 
-	$mail_signature = GETPOST('TICKET_MESSAGE_MAIL_SIGNATURE', 'restricthtml');
+	$mail_signature = htmlspecialchars_decode(GETPOST('TICKET_MESSAGE_MAIL_SIGNATURE', 'restricthtml'));
 	$signature_description = "Signature of ticket replies sent from Dolibarr";
 	if (!empty($mail_signature)) {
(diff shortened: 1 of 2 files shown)


eldy:
We should never have to use the function html_specialchar_decode as we have  more powerfull function for escaping in dolibarr lib. Also it is strang me to have to "unescape". 
Using it means there is a security hole somewhere else.
Can you provide an example of value used to reproduce the bug (screenshot of setup) ?

atm-corentin:
This the setup page of Tickets module

image

And here is what we see when you try to add a private message to a ticket

Capture d’écran du 2025-05-21 16-20-08

As you can see, if you put a link into the header or footer, the text message before the link will be displayed above the form

eldy:
> This the setup page of Tickets module
> 
> image
> 
> And here is what we see when you try to add a private message to a ticket
> 
> Capture d’écran du 2025-05-21 16-20-08
> 
> As you can see, if you put a link into the header or footer, the text message before the link will be displayed above the form

I pushed a different fix, thanks to check if it's ok for you.

PR #34210, opened 2025-05-20
FIX: pgsql: cast string_agg arguments

marc-dll:
# FIX: pgsql: cast string_agg arguments

When converting MySQL/MariaDB `GROUP_CONCAT` calls to PostgreSQL `STRING_AGG`, an error occurs because of incorrect argument types (the literal `','` that is added if no `SEPARATOR` is given in the original call is of type `unknown`, for example).

`STRING_AGG` takes either `(TEXT, TEXT)` or `(BYTEA, BYTEA)` arguments (see https://www.postgresql.org/docs/12/functions-aggregate.html), therefore, I propose casting to `TEXT`.


marc-dll:
I didn't see at first that a phpunit test existed, I modified it accordingly.

PR #34221, opened 2025-05-21
FIX: projects: go back to edit view after cloning

marc-dll:
# FIX: projects: go back to edit view after cloning

Since commit https://github.com/Dolibarr/dolibarr/commit/ea7e7456a5884471c13e1c0c2c9b538bc84969f0, the screen we go back to after cloning a project is the project visualization.

This is impractical, this makes to user do many inputs if they want to edit the cloned project, which is the majoritary use case (to remove the prefix "Copy of", etc.). It is even more notable when we clone a closed project, as we must put it back to draft or validated before modifying it.

As changing back the original line doesn't work (I guess because the `edit` action is handled after `confirm_clone`), I propose making a redirection to the edit view.

### DIFF
--- a/htdocs/projet/card.php
+++ b/htdocs/projet/card.php
@@ -499,7 +499,7 @@ if (empty($reshook)) {
 			$newobject->fetch_optionals();
 			$newobject->fetch_thirdparty(); // Load new object
 			$object = $newobject;
-			$action = 'view';
+			$action = 'edit';
 			$comefromclone = true;
 
 			setEventMessages($langs->trans("ProjectCreatedInDolibarr", $newobject->ref), "", 'mesgs');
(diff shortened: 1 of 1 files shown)


lvessiller-opendsi (approved):

htdocs/projet/card.php
```diff
@@ -496,14 +496,11 @@
 			// Load new object
 			$newobject = new Project($db);
 			$newobject->fetch($result);
-			$newobject->fetch_optionals();
-			$newobject->fetch_thirdparty(); // Load new object
-			$object = $newobject;
-			$action = 'view';
```
eldy, line 502: And what if you just set $action='edit' here to avoid the redirect ?

marc-dll: @eldy it doens't work : the form is displayed but if you cancel or leave the page, the cloned project is not created at all.

PR #34222, opened 2025-05-21
FIX Bug Menu Icon

zahwyrix:
# FIX
Currently, if I want to use a png logo for my module and an awesome font in the top menu, it's not possible. What's more, the image name isn't right, it's missing object_.

### DIFF
--- a/htdocs/theme/eldy/global.inc.php
+++ b/htdocs/theme/eldy/global.inc.php
@@ -3286,7 +3286,7 @@ li.tmenu:hover .tmenuimage:not(.menuhider), li.tmenu:hover .tmenuimage:not(.menu
 			}
 		}
 
-		 if ($menuEntry && !empty($menuEntry['prefix'])) {
+		if ($menuEntry && !empty($menuEntry['prefix'])) {
 			print "/* mainmenu $val uses a custom picto in prefix */\n";
 			continue;
 		}
--- a/htdocs/theme/eldy/global.inc.php
+++ b/htdocs/theme/eldy/global.inc.php
@@ -3278,6 +3278,19 @@ li.tmenu:hover .tmenuimage:not(.menuhider), li.tmenu:hover .tmenuimage:not(.menu
 			continue;
 		}
 
+		$menuEntry = null;
+		foreach ($menumanager->tabMenu as $entry) {
+			if (!empty($entry['mainmenu']) && $entry['mainmenu'] === $val && $entry['type'] === 'top') {
+				$menuEntry = $entry;
+				break;
+			}
+		}
+
+		 if ($menuEntry && !empty($menuEntry['prefix'])) {
+			print "/* mainmenu $val uses a custom picto in prefix */\n";
+			continue;
+		}
+
 		$found = 0;
 		$url = '';
 		$constformoduleicon = 'MAIN_MODULE_'.strtoupper($val).'_ICON';
@@ -3292,8 +3305,8 @@ li.tmenu:hover .tmenuimage:not(.menuhider), li.tmenu:hover .tmenuimage:not(.menu
 		} else {
 			// Search img file in module dir
 			foreach ($conf->file->dol_document_root as $dirroot) {
-				if (file_exists($dirroot."/".$val."/img/".$val.".png")) {
-					$url = dol_buildpath('/'.$val.'/img/'.$val.'.png', 1);
+				if (file_exists($dirroot."/".$val."/img/object_".$val.".png")) {
+					$url = dol_buildpath('/'.$val.'/img/object_'.$val.'.png', 1);
 					$found = 1;
 					break;
 				}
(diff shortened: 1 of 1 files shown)


eldy:
I do not experience troubles in mixing png file and fontawesome image. Did you try to name your image file nameofimage.png instead of object_nameofimage.png ?

zahwyrix:
If I am not mistaken, it must be called object so that it is the picto in the description with the name module@module

At first I tried to call them module.png in V21 and I didn't have the icon in the module page, with object_module.png I have the icon

eldy:
The convention object_module.png was used in past because itwas no possible to define the name of image.
Now we can set the name of picto to whatever value we want so we must now use the name of the picto as image file name.
For transition, i recommend to store both object_file.png and file.png to be compatible with old code and current one.

zahwyrix:
> The convention object_module.png was used in past because itwas no possible to define the name of image. Now we can set the name of picto to whatever value we want so we must now use the name of the picto as image file name. For transition, i recommend to store both object_file.png and file.png to be compatible with old code and current one.


It gives me this, indeed the font awesome is displayed well but the png appears in the background, it should be hidden if there is a font awesome

image

What's strange is that we never look at the characteristics of the menu entry (its icon, or its image) we just look at the characteristics of the module. That's what surprises me a bit, and that's why I made this PR.


Here is my menu:


```
$this->menu[$r++] = array(
			'fk_menu' => '', // Will be stored into mainmenu + leftmenu. Use '' if this is a top menu. For left menu, use 'fk_mainmenu=xxx' or 'fk_mainmenu=xxx,fk_leftmenu=yyy' where xxx is mainmenucode and yyy is a leftmenucode
			'type' => 'top', // This is a Top menu entry
			'titre' => 'ModulePriceName',
			'prefix' => img_picto('', 'fa-dollar-sign', 'class="fa-fw pictofixedwidth"'),
			'mainmenu' => 'price',
			'leftmenu' => '',
			'url' => '/custom/price/card.php',
			'langs' => 'price@price', // Lang file to use (without .lang) by module. File must be in langs/code_CODE/ directory.
			'position' => 1000 + $r,
			'enabled' => 'isModEnabled("price")', // Define condition to show or hide menu entry. Use 'isModEnabled("price")' if entry must be visible if module is enabled.
			'perms' => '1', // Use 'perms'=>'$user->hasRight("price", "myobject", "read")' if you want your menu with a permission rules
			'target' => '',
			'user' => 2, // 0=Menu for internal users, 1=external users, 2=both
		);
```

And my png picto :
`$this->picto = 'price@price';`

zahwyrix:
The night brought me advice, problem solved this morning with a css :

```
div.mainmenu.price {
	background-image: none;
}
```

But I really think we should discuss this PR, it's not normal to have two icons appearing on top of each other, or if it's intended, for the user experience, we should rather by default disable this and give the option to activate it via css.

eldy:
Strange.
This is menu entry i have for the top menu
```
$this->picto = "abc@abc";

$this->menu[$r]=array(	'fk_menu'=>0,
								'type'=>'top',
								'titre'=>'ABC',
								'prefix' => img_picto('', $this->picto, 'class="paddingright2imp pictofixedwidth valignmiddle"'),
								'mainmenu'=>'abc',
								'url'=>'/abc/index.php',
								'langs'=>'',
								'position'=>200,
								'enabled'=>'isModEnabled("abc")',
								'perms'=>'$user->hasRight("abc", "read")',
								'target'=>'',
								'user'=>0);
```

I have a file abc.png into dir img.  But i do not experience double menu entry. Only the png one is shown.
May be this is because you used a name that is also a name of a fontawesome key ?

zahwyrix:
> Strange. This is menu entry i have for the top menu
> 
> ```
> $this->picto = "abc@abc";
> 
> $this->menu[$r]=array(	'fk_menu'=>0,
> 								'type'=>'top',
> 								'titre'=>'ABC',
> 								'prefix' => img_picto('', $this->picto, 'class="paddingright2imp pictofixedwidth valignmiddle"'),
> 								'mainmenu'=>'abc',
> 								'url'=>'/abc/index.php',
> 								'langs'=>'',
> 								'position'=>200,
> 								'enabled'=>'isModEnabled("abc")',
> 								'perms'=>'$user->hasRight("abc", "read")',
> 								'target'=>'',
> 								'user'=>0);
> ```
> 
> I have a file abc.png into dir img. But i do not experience double menu entry. Only the png one is shown. May be this is because you used a name that is also a name of a fontawesome key ?

no I have 2 different names, the name of my module is not a font awesome  for the png. But for the icon of the module I want the icon dollars font awesome so it's normal that it appears what I don't understand that's why the logo of the module appears in background. 
I believe that the conflict comes from the fact that I have a png logo of the module and that in the menu I want a different picto font awesome.

eldy:
Closed. Use of image with name object_xxx has been abandonned in favor of name myfile.png

PR #34228, opened 2025-05-21
FIX: mandatory extrafelds on lines are not checked

FHenry:
If you set **mandatory** extrafields on lines of 

- Propal
- Invoice
- Supllier Order
- Supplier Invoice 

then Dolibarr let users add the line (or update) the line without error (or without displaying error message) when the extrafield is empty

On Order Lines, it is actually the only case where this is correctly managed

I've used the code of OrderLine->update and OrderLine->insert to apply the same check and rerror return and displayed on Propal Line, Invoice Line, Supplier Order line, Supplier Invoice line

### DIFF
--- a/htdocs/comm/propal/class/propal.class.php
+++ b/htdocs/comm/propal/class/propal.class.php
@@ -4443,8 +4443,17 @@ class PropaleLigne extends CommonObjectLine
 				// End call triggers
 			}
 
-			$this->db->commit();
-			return 1;
+			if (!$error) {
+				$this->db->commit();
+				return $this->id;
+			}
+
+			foreach ($this->errors as $errmsg) {
+				dol_syslog(get_class($this)."::insert ".$errmsg, LOG_ERR);
+				$this->error .= ($this->error ? ', '.$errmsg : $errmsg);
+			}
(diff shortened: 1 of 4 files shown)


htdocs/comm/propal/class/propal.class.php
```diff
@@ -4443,8 +4443,17 @@ public function insert($notrigger = 0)
 				// End call triggers
 			}
 
-			$this->db->commit();
-			return 1;
+			if (!$error) {
+				$this->db->commit();
+				return $this->id;
```
lvessiller-opendsi, line 4448: Here the return value was modified.
Before you got "return 1" and then you get "return $this->id" and it can be null

lvessiller-opendsi (changes requested):

htdocs/comm/propal/class/propal.class.php
FHenry: haa... you've spotted it.
In all other 'Line" class the insert method already return the rowid of the new line created, so it should be the same here, but I underantd that it is a change not a fix. I revert it and made another PR for develop

lvessiller-opendsi (approved):

PR #34250, opened 2025-05-22
FIX: SQL query too heavy 

noec764:
In the case of a very large number of users shared by many entities, the SQL query defined in User::fetchAll causes the page to stop loading. 
Adding a DISTINCT significantly improves performance.

Example of usage in action of s'ending mail from a document with email conf MAIN_MAIL_ENABLED_USER_DEST_SELECT.


htdocs/user/class/user.class.php
```diff
@@ -3949,7 +3949,7 @@ public function fetchAll($sortorder = '', $sortfield = '', $limit = 0, $offset =
 	{
 		global $conf, $user;
 
-		$sql = "SELECT t.rowid";
+		$sql = "SELECT DISTINCT t.rowid";
```
eldy, line 3952: We should never have to use distinct in SQL
Can you describe the final SQL you have ?

noec764: What do you recommend instead ? A group by t.rowid ? 
The original SQL query is 
```SQL 
SELECT t.rowid FROM llx_user as t ,llx_usergroup_user as ug WHERE ((ug.fk_user = t.rowid AND ug.entity IN (0,1)) OR t.entity = 0) AND (t.statut=1 AND t.employee=1 AND t.email IS NOT NULL AND t.email <> '') ORDER BY t.lastname ASC
```

rycks:
@noec764 "very large number of users shared by many entities" could you please give us more information about "very large number" ? 1000, 10.000, 100.000 users ? and how much entities ? that's just to try to make a local setup with same amount of data than you

thanks

noec764:
I have a configuration of 475 users shared by 48 entities, totaling 8,637 rows in llx_usergroup_user. It's not much, but it's enough to freeze the mariadb server when the request is executed.

htdocs/user/class/user.class.php
eldy: > ```sql
> SELECT t.rowid FROM llx_user as t ,llx_usergroup_user as ug WHERE ((ug.fk_user = t.rowid AND ug.entity IN (0,1)) OR t.entity = 0) AND (t.statut=1 AND t.employee=1 AND t.email IS NOT NULL AND t.email <> '') ORDER BY t.lastname ASC
> ```

Can you try to rewrite the request this way:

SELECT t.rowid FROM llx_user as t WHERE t.statut=1 AND t.employee=1 AND t.email IS NOT NULL AND t.email <> '' AND (t.entity = 0 OR EXISTS (SELECT ug.rowid FROM llx_usergroup as ug WHERE ug.fk_user = t.rowid AND ug.entity IN (0,1) )) ORDER BY t.lastname ASC

This is the SQL implementation of what we want.

First check you get the expected result, then when done check the performance of request. It still may be long but with this new request, without distinct, it become possible to add index that are missing on table t for this use case.  
For exemple 
ALTER TABLE llx_user ADD INDEX idx_user_status_employee(statut, employee);
If it is still too long, give me the result of the 
EXPLAIN SELECT ...

noec764:
@eldy 
I've updated the SQL query as requested and it works.
But adding an index to the llx_user table isn't considered a fix but a new feature, isn't it?
I've updated llx_user.key.sql, but I think I need to add this SQL query to repair.sql, I'm not sure how to do that.

eldy:
> @eldy 
> I've updated the SQL query as requested and it works.
> But adding an index to the llx_user table isn't considered a fix but a new feature, isn't it?
> I've updated llx_user.key.sql, but I think I need to add this SQL query to repair.sql, I'm not sure how to do that.

Adding index may be a fix or not: After changing the sql to get a simple sql with no distinct, what is the time to execute the sql without thevl addition of index and what is time with ?

lvessiller-opendsi (approved):

rycks:
ok for me
in case of optimization i think we could imagine an other one on email field because of

>  AND t.email IS NOT NULL AND t.email <> '' 

what do you think about that idea :

do not allow null on email field (`ALTER TABLE llx_user CHANGE email email VARCHAR(255) NOT NULL DEFAULT ''; `) and set default value to '' then the SQL request could be simplified with only one exclude filter on email ?

>  AND t.email <> ''

eldy:
Disallowing null will have effect in size of data stored. Null does not use space on disk, emptystring need 1 or 2 bytes. But this is not important so we can ignore this side effect.

However, there is another effect that we can't ignore:  disallowing the null have a large effect on indexes. Null is never stored into index, so if you have 90% of record with the email column that is empty, then the index is composed by 90% of "" (number of entry into index is multiplicated by 10 for 90% of empty, moving from 10% of data indexes to 100%) and the analysis of the selectivity of the use of index is also wrongly calculated: The plan analyzer will say, atfer reading statistics of column usage, that there is similar data for 90% of record so do not use index to filter on this column because the selectivity is not good enough.  This often break the usability of the index.

Also note that doing :
"AND t.email IS NOT NULL AND t.email <> ''"
is the same than
"AND t.email <> ''
because any comparison with NULL return false (except test "IS NULL"), so YES, t.email <> '' is false when t.email is NULL (even if this is counter-intuitive).
So having one test on <> '' is already enough. But for developers, the readbility is better with the 2 tests, this is why we add "AND t.email IS NOT NULL", but database already should ignore the instuction due to the second test.

So doing what you suggest will have
- very small degradation of size of storage for table (not important)
- large degration of performance and size of storage for index (very important)

A general rule in database management is: if you can store null instead of '', use null, but do not the opposite.

@noec764 To progress i suggest first to submit the PR with the SQL rewritten without the DISTINCT. We will see the rest later.

noec764:
https://github.com/Dolibarr/dolibarr/pull/34371

hregis:
> I have a configuration of 475 users shared by 48 entities, totaling 8,637 rows in llx_usergroup_user. It's not much, but it's enough to freeze the mariadb server when the request is executed.

sorry i don't see before !!! :-) 
48 entities !!
est-ce que tu as activé le cache memcahe de multicompany ?

PR #34258, opened 2025-05-23
DRAFT : Problem with HTML encoding on MultiSelectArrayWithCheckbox

warnerbryce:
There is a problem on Dolibarr 22.0-beta on MultiSelectArrayWithCheckbox. 
This is what i see : 
image
So bad isn't it ? 
I saw some diff between Dolibarr 21.0 and Develop on function multiSelectArrayWithCheckbox in core/class/html.form.class.php 

Dolibarr 21 => `$listoffieldsforselection .= '<label for="checkbox' . $key . '">'.dol_string_nohtmltag($langs->trans($val['label'])) . '</label></li>';`

Dolibarr 22 =>` $listoffieldsforselection .= dolPrintHTML(dol_string_nohtmltag($langs->trans($val['label'])));
`
Function dolPrintHTML get inserted. 

When i go into the details of the function i get a diff in treatment on function dol_htmlwithnojs in core/lib/function.lib.php
the function preg_replace gave two different output between Dolibar 21 and 22: 

Dolibarr 21 $out
`"<?xml encoding="UTF-8"><div class="tricktoremove">R&eacute;f.</div>" `will become Réf. 

Dolibarr 22 $out 
`"<?xml encoding="UTF-8"><div class="tricktoremove">R&eacute;f.</div>" ` will become `"R&eacute;f.</div>" 
`

If i remove the function dolPrintHTML in $listoffieldsforselection i got a correct printing.
image

The problem seems to be into the function dol_htmlwithnojs but i am not skilled enought to find why.
If someone know the fix to that.

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -9687,7 +9687,7 @@ class Form
 				// Note: $val['checked'] <> 0 means we must show the field into the combo list  @phan-suppress-next-line PhanTypePossiblyInvalidDimOffset
 				$listoffieldsforselection .= '<li><input type="checkbox" id="checkbox' . $key . '" value="' . $key . '"' . ((!array_key_exists('checked', $val) || empty($val['checked']) || $val['checked'] == '-1') ? '' : ' checked="checked"') . ' data-position="'.(empty($val['position']) ? '' : $val['position']).'" />';
 				$listoffieldsforselection .= '<label for="checkbox' . $key . '" class="paddingleft">';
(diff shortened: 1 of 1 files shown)


warnerbryce:
I got the same problematic print on societe/card.php 
image

dolPrintHTML is involved in societe/card.php  line 3304

warnerbryce:
In fact i have this problem because getDolGlobalString('MAIN_RESTRICTHTML_ONLY_VALID_HTML') = 1 
But in llx_const MAIN_RESTRICTHTML_ONLY_VALID_HTML isn't set. 
I will have to look further why. 
But the problem is not on dolPrintHTML side.

PR #34286, opened 2025-05-26
FIX: Use highest VAT rate from dictionary when buyer's state/province…

Expresion:
When the buyer's state or province has a VAT rule defined in the VAT rates dictionary, Dolibarr should apply the highest applicable rate as the default. This fix ensures that the correct VAT is selected in such cases.

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -6996,7 +6996,7 @@ class Form
 		$sql .= " AND t.active > 0";
 		$sql .= " AND t.entity IN (".getEntity('c_tva').")";
 		$sql .= " AND c.code IN (" . $this->db->sanitize($country_code, 1) . ")";
-		$sql .= " ORDER BY t.code ASC, t.taux ASC, t.recuperableonly ASC";
+		$sql .= " ORDER BY t.code ASC, t.taux DESC, t.recuperableonly ASC";
 
 		$resql = $this->db->query($sql);
 		if ($resql) {


eldy:
Dolibarr use the vat rate withvstatus Default when there is several vat rate in the list. Did you set the vat rate you want to "Default" in the vat dictionary ?

JamBalaya56562:
This does not look needed anymore, and the change would have a side effect.

**The buyer state/province case is already handled by `get_default_tva()`.** Since 21.0, VATRULE 2 selects the VAT rate attached to the buyer's department (`c_tva.fk_department_buyer`) with

```sql
ORDER BY t.use_default DESC, t.taux DESC, t.code ASC, t.recuperableonly ASC
```

so the rate flagged as *Default* wins first and, failing that, the highest rate — which is what @eldy described above. `load_tva()` calls `get_default_tva()` to compute the preselected rate, so when the buyer's state/province has a rule in the dictionary the right rate is already preselected, whatever order `load_cache_vatrates()` uses.

**Changing the sort in `load_cache_vatrates()` would break the documented fallback in `load_tva()`:**

```php
// If we fails to find a default vat rate, we take the last one in list
// Because they are sorted in ascending order, the last one will be the higher one (we suppose the higher one is the current rate)
if ($defaulttx < 0 || dol_strlen($defaulttx) == 0) {
    if (!getDolGlobalString('MAIN_VAT_DEFAULT_IF_AUTODETECT_FAILS')) {
        // We take the last one found in list
        $defaulttx = $arrayofvatrates[$num - 1]['txtva'];
```

That fallback takes the **last** entry precisely because the list is sorted ascending. With `t.taux DESC` the last entry becomes the **lowest** rate, so whenever auto-detection fails Dolibarr would preselect the lowest VAT rate instead of the highest — for every country, not only for the buyer-state case. It would also change the display order of every VAT combo.

So the case this PR targets is covered by VATRULE 2 today, and the proposed sort change would introduce a regression. This PR can probably be closed.

PR #34291, opened 2025-05-27
Handling of selected customer invoice to create a credit note if not in the list restricted by secret parameter

splohmer:
# NEW|New Handling of selected customer invoice to create a credit note if not in the list restricted by secret parameter
Previously, the parameter LIST_OF_QUALIFIED_INVOCIES_LIMIT_DEFINED could limit the number of customer invoices shown in the dropdown when creating a credit note. In some cases—especially when using the "Create Credit Note" button from within an invoice—the original invoice was not included in the list due to this restriction.
With this update, the originating invoice will now always be included in the dropdown, regardless of the defined limit.

### DIFF
--- a/htdocs/compta/facture/class/facture.class.php
+++ b/htdocs/compta/facture/class/facture.class.php
@@ -5382,11 +5382,13 @@ class Facture extends CommonInvoice
 		if ($socid > 0) {
 			$sql .= " AND f.fk_soc = ".((int) $socid);
 		}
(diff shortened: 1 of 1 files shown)


splohmer:
Hi, this is my first commit for Dolibarr. I saw that 2 pipelines where failing, but all the things mentioned are not changed by my code or even in the corresponding class. Do I need to fix that or how is the process here?

Thanks and best regards,
Sven

PR #34309, opened 2025-05-28
fix flowjs-server to have specific upoload dir in tempdir

Hystepik:
Fix the impossibility to choose the upload folder in the module temp folder with flow upload process


htdocs/core/ajax/flowjs-server.php
```diff
@@ -73,6 +74,9 @@
 
 if (!empty($conf->$module->dir_temp)) {
 	$upload_dir = $conf->$module->dir_temp;
+	if (!empty($uploaddirname)) {
+		$upload_dir .= "/".$uploaddirname;
```
eldy, line 78: Must sanitize the uploaddirname with dol_sanitizeFileName().
Can be done line 61 just after getting paramvalue.

PR #34312, opened 2025-05-28
fix: ref_client GETPOST facture/card.php should be alpha

FHenry:
following https://github.com/Dolibarr/dolibarr/pull/34311 for 20.0


eldy:
Already submited in v18. Closed

Issue #33911, opened 2025-04-18
Active services are reported as expired on the homepage


ro78:
### Bug

I have 23 active services

![Image](https://github.com/user-attachments/assets/ec47b935-c687-484f-8bf3-4922d6a17631)

I don't have any of them expired

![Image](https://github.com/user-attachments/assets/63be325f-6ad2-4488-bcc5-e3635b81a593)

Still on the dashboard/home page, they are reported under "expired" instead of "active".

![Image](https://github.com/user-attachments/assets/59e0296b-aa24-4246-a95c-c5afff32ea12)

### Dolibarr Version

21.0.1

### Environment PHP

8.2.28

### Environment Database

MariaDB 10.11.11

ro78:
@sonikf maybe?

ro78:
@sonikf in htdocs/contrat/class/contrat.class.php on line 2343, you just need to replace
`			} elseif ($mode == 'active') {
`
by
`			} elseif ($mode == 'expired') {
`

It was OK in branch 20, broken in branch 21. #

sonikf:
Hi @ro78 
Can you post a screenshot of list of contracts or post the status of those 23 contacts?

ro78:
@sonikf I don't have any expired services

![Image](https://github.com/user-attachments/assets/26587a66-aac8-461f-b983-6843b5aaec69)

but I do have 23 active contracts 

![Image](https://github.com/user-attachments/assets/401b3d65-c9ab-441e-8af3-7cdb0350d191)

If you compare the line I shared between version 21.0 and 20.0, you'll see that there was a change in the order of if/else, and it makes the dashboard shows them as "expired" while they are active.

sonikf:
I need to see the status or screenshot of status column in contrat/list.php cause  i don't use contracts currently
I think the change was made to fix the list.php and broke the info box...

ro78:
@sonikf they are all active or closed. None of them are expired.

![Image](https://github.com/user-attachments/assets/d6efcdca-859a-4d1c-8cc2-ac4ae263251f)

sonikf:
Hi @ro78 
I can confirm your fix.
Would you like to submit the fix?
It is a very simple process from your browser, you just have to fork Dolibarr with all branches then choose and submit fix in your 21 branch. 
Just be careful to submit back to  21 and not develop see:

![Image](https://github.com/user-attachments/assets/01e8164d-9ad6-43d6-b9c4-c6084d065a68)

[...]

PR #34317, opened 2025-05-28
FIX: #33911 active services are reported as expired on the homepage

ro78:
# FIX|Fix #[*33911 active services are reported as expired on the homepage*]
In version 21.0, a change was introduced, making active contracts counted as expired contracts on the dashboard.
This fix will address this issue, correctly showing them as active on the dashboard.


ro78:
@eldy any idea in which version it will be included? It wasn't included in 21.0.2

eldy:
> @eldy any idea in which version it will be included? It wasn't included in 21.0.2

Pr has been merged into v21 branch so it should be available in v21.0.2. If not 21.0.3, except if another pr has reversed the change. Check with branch v21.

ro78:
@eldy upgraded to 22.0 and still not fixed

PR #34318, opened 2025-05-29
NEW Use arrays in default values (for tags in particular)

zahwyrix:


eldy:
Can you provide example of why you need this ?

zahwyrix:
In any list if you want to filter by default category, for example.
For example, I have 0€ products that only interest me in certain cases, and I'd like to be able to filter by default all categories except those.
I'll add a condition in my PR code to authorize arrays only if the category module is activated.

My product category at 0€ at id 3, so I put in default values for the list [1,2] and it works well.

eldy:
> In any list if you want to filter by default category, for example. For example, I have 0€ products that only interest me in certain cases, and I'd like to be able to filter by default all categories except those. I'll add a condition in my PR code to authorize arrays only if the category module is activated.
> 
> My product category at 0€ at id 3, so I put in default values for the list [1,2] and it works well.

Sorry, it is not clear. Do you speak about the feature "default value" in create form ? Can you provide a screenshot  of setup used to set [1,3] ?

zahwyrix:
> > In any list if you want to filter by default category, for example. For example, I have 0€ products that only interest me in certain cases, and I'd like to be able to filter by default all categories except those. I'll add a condition in my PR code to authorize arrays only if the category module is activated.
> > My product category at 0€ at id 3, so I put in default values for the list [1,2] and it works well.
> 
> Sorry, it is not clear. Do you speak about the feature "default value" in create form ? Can you provide a screenshot of setup used to set [1,3] ?

image
image

[2,3] => Identifiers of the two categories I see by default, so I exclude products in the category with id 1 when the page opens.

Issue #34315, opened 2025-05-28
BUG: API get stockmovements sqlfilter - Unknown column 't.product_id'


JonBendtsen:
### Bug

BUG: when I try to use the API to get stockmovements for a product the text says:

```
Other criteria to filter answers separated by a comma. Syntax example "(t.product_id:=:1) and (t.date_creation:
```

but when I actually use it with this sqlfilter `(t.product_id:=:128)`

I get the error message
```
{
  "error": {
    "code": 503,
    "message": "Service Unavailable: Error when retrieve stock movement list : Unknown column 't.product_id' in 'WHERE'"
  }
}
```

### Dolibarr Version

21.0.1

### Environment PHP

8.2.28

### Environment Database

mariadb latest

JonBendtsen:
somehow fixed in 22.0.1

PR #34320, opened 2025-05-29
attempt at fixing #34315 BUG: API get stockmovements sqlfilter

JonBendtsen:
# FIX BUG: API get stockmovements sqlfilter - Unknown column 't.product_id'
This PR changes product_id to fk_product because that is what is used in the database

image

After these changes I can then use the API GET stockmovements for just that product and get a good response

api_get_stockmovements


eldy:
You may note that the field warehouse_id does not exists too.
The name of field in API are defined to be standard (they does not necessarly match field in database, but try to match the class property that we try to standardize step by step).

So for me only the example for the filter must be fixed (becasue for the filter, it must match the database field that may differ from the class property name, there is no mapping layer yet).

JonBendtsen:
someone has fixed it in 22.0.1

PR #34330, opened 2025-06-02
NEW Accountancy - Add a protection on various payment for auxiliary account on general account not centralized

aspangaro:
# NEW|New [Add a protection on various payment for auxiliary account on general account not centralized]

Today, too many mistakes are being made. Many users specify sub-ledger accounts on G/L accounts that are not centralized (e.g. 411/401/421). This leads to difficulties in importing to accounting software and problems in understanding the accounting.

Solution: Add a message to prevent and not recorded auxiliary account. It's not an error, we continued !

Example: Create a various payment with an auxiliary account and a general account not centralized 

image

Capture d’écran du 2025-06-02 04-13-19

Capture d’écran du 2025-06-02 04-20-36

Capture d’écran du 2025-06-02 04-15-03


aspangaro:
@philippe-opendsi @kkhelifa-opendsi Maybe use this method to prevent error on various payment creation with module "Suivi bancaire"

PR #34331, opened 2025-06-02
FIX : bad value for company ref crashes canvas feature

altairis-tof:
company ref was loaded with rowid that cause crashes in canvas assign_values (fetch was done on rowid=2 and name=2 for example). 
ref should be the company name like in the info() function


eldy:
In Dolibarr, a ref is something that must be unique. This is why the ref (that does not exist in company) is using the ID.

PR #34339, opened 2025-06-02
FIX: accountancy sales/purchases journal: sql error logged when no invoice in journal

marc-dll:
# FIX: accountancy sales/purchases journal: sql error logged when no invoice in journal

In purchases and sales journal, when checking if some lines are not bound in the selected invoices, an SQL error occurs if no invoices are selected (this generates an SQL statement with `AND fk_facture IN ()`).

A simple check on array `$tabfac` being empty clears the error and has the added benefit of suppressing one superfluous SQL query.

And while fixing that, I discovered the check was not working for sales invoice as a copy/paste made in so that an incorrect sql result column was used (`$obj->fk_facture_fourn` instead of `$obj->fk_facture)`.

### DIFF
--- a/htdocs/accountancy/journal/purchasesjournal.php
+++ b/htdocs/accountancy/journal/purchasesjournal.php
@@ -361,29 +361,32 @@ foreach ($tabfac as $key => $val) {		// Loop on each invoice
 }
 */
 // New way, single query, load all unbound lines
-$sql = "
-SELECT
-    fk_facture_fourn,
-    COUNT(fd.rowid) as nb
-FROM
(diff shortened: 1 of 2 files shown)


alexandre-janniaux:
Hi, sorry I made the changes targetting only develop back then in https://github.com/Dolibarr/dolibarr/pull/32573 and https://github.com/Dolibarr/dolibarr/pull/32574 (but didn't made the $obj->fk_facture_fourn change). You might also be interested in https://github.com/Dolibarr/dolibarr/pull/32675 ;)

PR #34344, opened 2025-06-03
NEW  - hide objectlinked in public note of pdf

atm-corentin:
# NEW|New Hide linked object detail in public note in pdf

I noticed there is no way to hide the linked object details in the PDF, as you can see in the picture.

image
 

To fix this, I propose adding a hidden configuration to hide them. What do you think?

### DIFF
--- a/htdocs/core/lib/pdf.lib.php
+++ b/htdocs/core/lib/pdf.lib.php
@@ -2593,7 +2593,7 @@ function pdf_getLinkedObjects(&$object, $outputlangs)
 			$outputlangs->load('orders');
 
 			if (count($objects) > 1 && count($objects) <= (getDolGlobalInt("MAXREFONDOC") ? getDolGlobalInt("MAXREFONDOC") : 10)) {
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/pdf.lib.php
```diff
@@ -2593,7 +2593,7 @@ function pdf_getLinkedObjects(&$object, $outputlangs)
 			$outputlangs->load('orders');
 
 			if (count($objects) > 1 && count($objects) <= (getDolGlobalInt("MAXREFONDOC") ? getDolGlobalInt("MAXREFONDOC") : 10)) {
-				if (empty($object->context['DolPublicNoteAppendedGetLinkedObjects'])) { // Check if already appended before add to avoid repeat data
+				if (empty($object->context['DolPublicNoteAppendedGetLinkedObjects']) && !getDolGlobalString("hideLinkedObjectInPublicNotePDF")) { // Check if already appended before add to avoid repeat data
```
eldy, line 2596: Name of options must always be in uppercase and underscore. Can you rename it ?

eldy (changes requested):

PR #34350, opened 2025-06-04
# FIX|Fix in list, removed useless search_entity

atm-irvine:
# FIX|Fix in list, removed useless search_entity
Search_entity was never used in project list.

### DIFF
--- a/htdocs/projet/list.php
+++ b/htdocs/projet/list.php
@@ -105,7 +105,6 @@ $pageprev = $page - 1;
 $pagenext = $page + 1;
 
 $search_all = GETPOST('search_all', 'alphanohtml');
-$search_entity = ($user->entity > 0 ? $user->entity : GETPOSTINT('search_entity'));
 
 $search_ref = GETPOST("search_ref", 'alpha');
 $search_label = GETPOST("search_label", 'alpha');
@@ -432,7 +431,6 @@ if (empty($reshook)) {
 		$search_price_booth = '';
 		$search_login = '';
 		$search_import_key = '';
-		$search_entity = '';
 		$toselect = array();
 		$search_array_options = array();
 		$search_category_array = array();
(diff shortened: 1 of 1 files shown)


hregis (approved):

eldy:
I don't understand how you can filter on entity if you remove this.
Like on this page (column Environment added by the multicompany module):
image

atm-irvine:
The problem is that projects are displayed even though we don't have access to them because they are on different entities. On the other objects, we have the objects of the current entity even if the user comes from a different entity.
What's more, I can't find an environment column in the list of projects. Is there a configuration to enter?

eldy:
> The problem is that projects are displayed even though we don't have access to them because they are on different entities. On the other objects, we have the objects of the current entity even if the user comes from a different entity. What's more, I can't find an environment column in the list of projects. Is there a configuration to enter?

I don't think this need a setup. May be this depends on the multicompany module version ?

Poject should not appear if that are on an entity that is not "shared", thanks to line
`$sql .= " WHERE p.entity IN (".getEntity('project', (GETPOSTINT('search_current_entity') ? 0 : 1)).')';
`

So if no specific filter is set, the filter is done on IN getEntity('project'). This looks to be the expected behaviour. If you don't want to see project of other entity, you must not share the project on several entities.

hregis:
@atm-irvine @eldy the search of entity is already define with hook in multicompany module !

atm-irvine:
We'll never go here : `$sql .= " WHERE p.entity IN (".getEntity('project', (GETPOSTINT('search_current_entity') ? 0 : 1)).')';`

Because search_entity is always greater than 0 because it takes the user's entity : 
`$search_entity = ($user->entity > 0 ? $user->entity : GETPOSTINT('search_entity'));`

So we always have other project from different entities

hregis:
@eldy i agree with @atm-irvine
This code is already in multicompany module with the hooks...

atm-irvine:
@eldy Is it ok for you?

eldy:
> @eldy Is it ok for you?

This seems to break filter on entity when using filter string on link or old version of multicompany.
As this look more a quality pr and not a fix pr (no process to reproduce a big) i think we can take no risk on this and keep  a status quo on current code.

hregis:
@eldy this filter exists in long time in Multicompany module. I don't understand why this is included in dolibarr?

hregis:
@eldy @atm-irvine
Can we find out who added this and ask them why?

eldy:
> @eldy @atm-irvine
> Can we find out who added this and ask them why?

May be me, to solve troubles for some multicompany users.

hregis:
@atm-irvine @eldy this fix can be it's the best ! https://github.com/Dolibarr/dolibarr/pull/34714

PR #34355, opened 2025-06-04
# NEW|New - add conf to hide linked object in public note of pdf

atm-corentin:
# NEW|New I noticed there is no way to hide the linked object details in the PDF

To fix this, I propose adding a hidden configuration to hide them.
Or I have another suggestion: instead of adding a hidden configuration, should I add a configuration inside the setup of the Invoices module?
Which one do you choose ?


eldy:
Feature was added manually to solve conflict.
Constant has been named PDF_HIDE_LINKED_OBJECT_IN_PUBLIC_NOTE

PR #34359, opened 2025-06-05
FIX: warnings when printing start/end date on line tpl

marc-dll:
# FIX: warnings when printing start/end date on line tpl

Some supplier classes don't use `DoliDB::jdate()` after fetching lines from database, like it is done elsewhere. This causes numerous warnings that flood the `dolibarr.log` file, as `dol_print_date()` (called by get_date_range() inside the lines templates) deprecates being called with `YYYY-MM-DD (HH:MM:SS)` values.


rycks (approved):
ok for me

PR #34360, opened 2025-06-05
FIX: correctly calculate prices in supplier context + remove warnings from calcul_price_total()

marc-dll:
# FIX correctly calculate prices in supplier context + remove warnings from calcul_price_total()

In many Dolibarr instances we host, the `dolibarr.log` file is flooded with warnings related to price calculation :

```
Price.lib::calcul_price_total Warning: function is called with parameter seller that is missing
```

To lighten the log file, I searched all instances of `calcul_price_total()` (itself being called by `CommonObject::update_price()`), and explicitly passed the `$seller` argument.

This warning was effective as in most of the supplier part of Dolibarr, it was incorrectly assumed to by `$mysoc` (which can lead to incorrect local tax amounts if I remember correctly), so I passed the supplier instead.

For expense reports, I passed a modified version of `$mysoc` as was already done in some places of the `ExpenseReport` class.

A further step in the future would be to make the `$seller` argument mandatory in both `CommonObject::update_price()` and `calcul_price_total()`.


eldy:
As this PR mostly fixes hidden warning and not visible bugs, it should be pushed in develop.

rycks:
> As this PR mostly fixes hidden warning and not visible bugs, it should be pushed in develop.

not "visible" for who ? as marc i'm in the same situation : we are flooded with tons of lines in log then we make a "grep -v" when we have to analyse logs but that's a bug in adminsys point of view and really visible ... and log file size could become a problem, for me that's clearly a target of a long term support version

marc-dll:
@eldy Well, those warnings are so much well "hidden" that they are everywhere in the logs of most of our customers! I'm really surprised nobody catched them before, they come from the core for the most part, as our customers don't use much external modules. The PRs I suggested this week all come from a simple task I had never done before in 8 years of working with Dolibarr : monitor log file size, read heavy ones and see what you can find there.

Ans so here we are, I mean, I use the word "flooded" properly : in the state of the code before this PR and all others of the same kind (which you merged, by the way), we are speaking about **megabytes** of output in `dolibarr.log` in `LOG_WARNING` level in just several days of time. About this particular PR, any CRUD operation on customer or supplier document lines (or expense report lines) trigger it, and they happen everyday even for small Dolibarr instances.

As such, I'm convinced it unnecessary heavy logs prevent us developpers from catching others bugs, that are less log-intrusive but potentially more impactful, of instances of the same bug but coming from external modules.

This PR doesn't add anything new and/or structural, it doesn't need going through the upgrade process, it doesn't modify existing documents, it doesn't discriminate instances based on the hosting conditions, modules used, etc. It's only simple fixes, pushing towards better code quality, cleaner instance management, easier bug discovery, so it fully has its place in a stable version. Especially version 18.0, which is engaged in a long-term support process.

Above all, it contains a fix about an issue in tax calculation, which is in itself a very serious topic. I would not want for our community to be liable for any financial damage done to a fiscal administration or a Dolibarr user due to this oversight.

eldy:
> Above all, it contains a fix about an issue in tax calculation, which is in itself a very serious topic. I would not want for our community to be liable for any financial damage done to a fiscal administration or a Dolibarr user due to this oversight

If a PR contains a fix, the PR must contains the very minimal change to fix the trouble. As stated in the CONTRIBUTING.md

> "* A PR must be atomic. It means it must contains the lower possible changes for 1 need (1 bug fix or 1 new feature) without breaking usability of code. If a PR can be split into several PRs, it often means your PR is not atomic. For PR on maintenance versions, the modified code must be as low as possible and only the absolutely necessary changes are allowed (You must have the number of lines modified as low as possible to have the visible bug fixed (no refactoring, no perf, no quality, no renaming must be joined). This allow to reduce to the maximum the number of future conflicts (so the number of regressions) when merging fixes into higher version to reduce the time of any future impact analysis."

rycks:
@marc-dll it could be easyer to understand if you add a global $mysoc into update_price function no ?
and your patch / PR will be more tiny in that case i think.

rycks:
> @marc-dll it could be easyer to understand if you add a global $mysoc into update_price function no ? and your patch / PR will be more tiny in that case i think.

ok i understand that fix has two targets, first was log flood but the most important i think is the second part "fix about an issue in tax calculation"

so how could we fix that ?

i think we need to make a example of the problem @marc-dll, is that for supplier invoices for example ?

PR #34364, opened 2025-06-05
add param to filter product on formsetup item

ATM-NicolasV:
# Instructions

[*Long description*]
Add a filter to select in a formsetup select the type of product


eldy:
This PR still has CI errors and conflicts and is not open to fix by third party developers so we can't fix it. A lot of code may have been validated since the PR and the PR may be obsolete, or too difficult to merge due to other changes. Best choice is too cancel it to avoid to keep PR with too low chances to be validated.

PR #34369, opened 2025-06-06
New: Add function to make substitutions on Takepos receipt

fmarcet:
The complete_substitutions_array() function was not called, which meant that external module substitutions were not made on the TakePOS ticket.


jseborga:
ok

PR #34373, opened 2025-06-07
QUAL: stop support of run_trigger

marc-dll:
# QUAL: stop support of run_trigger

In trigger implementations, use of method `run_trigger` instead of `runTrigger` has been deprecated for 11 years now (see commit https://github.com/Dolibarr/dolibarr/commit/cfc337a79a52fba9e0dd19a7881c14356ac55174#diff-b5224739ad0df49fba209a47b98fc8409f21320b8607bf08d183a016069b4bb5), it's high time to cut the cord!


eldy:
Can you also add a mentuon ijto ChangeLog file to warn ti rename methid in its modules.

PR #34376, opened 2025-06-08
PERF: remove useless calls to count() in core classes, tpls and modulebuilder template

marc-dll:
# PERF: remove useless calls to count() in core classes, tpls and modulebuilder template

Dolibarr is filled with calls to function `count()` whose result is simply compared to 0 (to see if the passed `array` is empty or not) or below a certain value. This loses a lot of time because under the hood, time is spent needlessly to really return the number of elements.

I searched for any call to count in main core Dolibarr classes, tpls, and in the modulebuilder template, and replaced those calls (mostly using `empty()` or replacing `for` with `foreach`) where necessary.

### DIFF
--- a/htdocs/core/class/cgenericdic.class.php
+++ b/htdocs/core/class/cgenericdic.class.php
@@ -255,13 +255,11 @@ class CGenericDic
 
 		// Manage filter
 		$sqlwhere = array();
-		if (count($filter) > 0) {
-			foreach ($filter as $key => $value) {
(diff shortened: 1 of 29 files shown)


eldy:
As this is refactoring, this must be pushed into develop branch.
Only bug fixes can be pushed in already released version.

eldy:
PR could not be merged into develop branch, so i close it.

PR #34397, opened 2025-06-11
do not add connected user by default to all new customers as representative

frederic34:


htdocs/societe/card.php
```diff
@@ -1971,8 +1971,11 @@ function manageprospectcustomer(element) {
 			print '<td colspan="3" class="maxwidthonsmartphone">';
 			// TODO Use select_doluser in multiselect mode
 			$userlist = $form->select_dolusers($selected, '', 0, null, 0, '', '', '0', 0, 0, 'u.statut:=:1', 0, '', '', 0, 2);
-			// Note: If user has no right to "see all thirdparties", we force selection of sale representative to him, so after creation he can see the record.
-			$selected = (GETPOSTISARRAY('commercial') ? GETPOST('commercial', 'array:int') : (GETPOSTINT('commercial') > 0 ? array(GETPOSTINT('commercial')) : array($user->id)));
+			$selected = (GETPOSTISARRAY('commercial') ? GETPOST('commercial', 'array:int') : (GETPOSTINT('commercial') > 0 ? array(GETPOSTINT('commercial')) : []));
+			if (!$user->hasRight('societe', 'client', 'voir')) {
```
eldy, line 1975: May be we can add 
and !getDolGlobalConst('THIRDPARTY_DO_NOT_ASSIGN_POWERFUL_AUTHOR_AS_SALE_REPRESENTATIVE')
 in this test so user that used to have it auto assigned would be able to continuebtobhave this feature.

eldy (changes requested):

PR #34402, opened 2025-06-12
NEW: Support non standard names for PDF massaction mailing and merging

altairis-noe:
# NEW: Support non standard names for PDF massaction mailing and merging
Avoids those errors
image
image

When there is a last_main_doc for the object, but not a commonly named PDF


eldy:
Imagine you have 2 templates of pdf, one to build the maik pdf id the invoice to the customer, another one to build a pdf to make complain related to the thirparty invoice to your attorney service.

When generating the doc it may mix the 2 in same pdf mixing 2 features.

For the reason the mass merging feature must only merge the main doc of document because we can't select for the moment wich template to use.

altairis-noe:
I would argue that’s why CommonObjects in Dolibarr have a last_main_doc field which should be able to choose the right file, or have we understood that wrong?

altairis-noe:
Hey there, @eldy, do you have any suggestion on how to address this issue, if last_main_doc is not appropriate?

altairis-noe:
I really wish I shouldn’t bypass dolibarr methods in order to be able to send documents to multiple clients at once. Do you think we can find a reasonable solution, @eldy

eldy:
> I really wish I shouldn’t bypass dolibarr methods in order to be able to send documents to multiple clients at once. Do you think we can find a reasonable solution, @eldy 

But which document. If you have 3 templates, and you want to send to n customers, the odf, which pdf ? I think the only solution is to have a selectionnof the template into the confirm popup so we can force it into the generation if it differs from last_main_doc

altairis-noe:
Here is the use case :
We generate the PDFs with an external module which names the files `mycompany_REF.PDF` (or even differently).
Those PDFs actually *are* last_main_doc, but as they are not named `REF.PDF`, they can’t be sent as « main document » in the current state of dolibarr code.

That’s what we are trying to fix here: f there is a last_main_doc, use it, even if it’s named funny!

altairis-noe:
> > I really wish I shouldn’t bypass dolibarr methods in order to be able to send documents to multiple clients at once. Do you think we can find a reasonable solution, @eldy
> 
> But which document. If you have 3 templates, and you want to send to n customers, the odf, which pdf ? I think the only solution is to have a selectionnof the template into the confirm popup so we can force it into the generation if it differs from last_main_doc

Hello again, @eldy 
To be more precise: what we want here is not generating a new doc, as your answer suggests, but getting the previously generated doc and add it to the mail to be sent.
In the current state, the mail won’t be sent because the file name doesn’t suit Dolibarr…

eldy:
Bur if you take the last generated without forcing generation, it means you take a doc randomly (it may be any template, so it may be any kind of document, the one for the lawyer, the one for the boss, the one for the customer, ...). It is a very strange use case.

altairis-noe:
> Bur if you take the last generated without forcing generation, it means you take a doc randomly (it may be any template, so it may be any kind of document, the one for the layser, the one for the boss, the one for the customer, ...). It is a very strange use case.

Right now, it doesn’t ask any questions and just joins the PDF with the same name as the object, or, if it doesn’t find it, we get an error message…
Regenerating the PDF we want before sending is a good idea, though, we could alter the presend massaction in that perspective.

eldy:
> > Bur if you take the last generated without forcing generation, it means you take a doc randomly (it may be any template, so it may be any kind of document, the one for the layser, the one for the boss, the one for the customer, ...). It is a very strange use case.
> 
> Right now, it doesn’t ask any questions and just joins the PDF with the same name as the object, or, if it doesn’t find it, we get an error message…
> Regenerating the PDF we want before sending is a good idea, though, we could alter the presend massaction in that perspective.
> 

Currently the pdf generated with main default template with has only the name of ref, and is always the same nature document, it is the main document template (so the same for every object) so this is not a random selection of the document. The user know which document is used because any other custom templates generates a document that is not the name of ref but something else like ref_suffix.odt (example, si tu active le template odt de commande template_order.odt, on ne connait pas la nature du doc, donc le doc aura un nom qui ne sera pas la simple ref mais ref_suffix, ici ref_order.odt)

Not sure this is a rigourous rule but it is currently how we avoid trouble when doing mass merging: we only meegz file with ref.pdf so we are sure we merge the main doc related to object. If we wan to mergz fules that match a different naming convention, so a different nature of document, we should ask it into the main confirm popup so we can find the correct files to merge (and regenerate if necessary).

altairis-noe:
> Currently the pdf generated with main default template with has only the name of ref, and is always the same nature document, it is the main document template (so the same for every object) so this is not a random selection of the document. The user know which document is used because any other custom templates generates a document that is not the name of ref but something else like ref_suffix.odt (example, si tu active le template odt de commande template_order.odt, on ne connait pas la nature du doc, donc le doc aura un nom qui ne sera pas la simple ref mais ref_suffix, ici ref_order.odt)
> 
> Not sure this is a rigourous rule but it is currently how we avoid trouble when doing mass merging: we only meegz file with ref.pdf so we are sure we merge the main doc related to object. If we wan to mergz fules that match a different naming convention, so a different nature of document, we should ask it into the main confirm popup so we can find the correct files to merge (and regenerate if necessary).

OK, I think I get it. So, I suggest we *remove* from this PR the part concerning merging PDFs.

And we condition the possibility to use a custom named last_main_doc to a new constant MAIN_PRESEND_JOIN_CUSTOM_LAST_MAIN_DOC.

And, with a second constant MAIN_PRESEND_JOIN_REGENERATE_LAST_MAIN_DOC, add a model selector, from which we would regenerate the PDFs before sending them.

What would you think of that approach?

altairis-noe:
OK,
sorry about the noise, we finally found out that the MAIL_MASS_ACTION_ADD_LAST_IF_MAIN_DOC_NOT_FOUND covered our needs for the mail part. And we will open another PR that would permit, with a hook, to change the name template for file merging.
I think it will be better handled that way.

PR #34410, opened 2025-06-12
Fix: Initialize $groupby variable to avoid PHPStan error

grandoc:
# FIX|Fix #PHPStan reports "Variable $groupby might not be defined".his commit explicitly initializes the `$groupby` variable to ensure it is defined before usage.


eldy:
Pb already solved

PR #34421, opened 2025-06-12
QUAL: remove deprecated module webservices/soap

marc-dll:
# QUAL: remove deprecated module webservices/soap

Self-explanatory, this module has been deprecated


eldy:
Module still used by many external modules. Deprecation should be extended for a longer period...

PR #34422, opened 2025-06-12
QUAL: OpenID Setup fused in Security Setup + Module removed -- WIP

moreauftheobald:
# QUAL OpenID Setup fused in Security Setup + Module removed -- WIP
As requested By Eldy at Valence Devcamp, 
Open Id Module fusion in Security Setup
Module removed from Core Module
Misc translations


JonBendtsen:
Hi @moreauftheobald  and @eldy 

The consequence of this PR is this `BUG: OpenIDConnect module can not be searched in 23.0.0-alpha` #36048

So why do I want to search it? Well, the dolibarr.log file says

```
2025-11-01 09:46:43 INFO    192.168.127.1        58     33 functions_openid_connect::check_user_password_openid_connect Module disabled
```

When I specify this configuration
`$dolibarr_main_authentication='openid_connect,dolibarr'`

and try to log in, thus I try to search for it to enable it.

JonBendtsen:
here is an attempt to fix the bug #36051

PR #34424, opened 2025-06-13
FIX BankTransfer link in rejets.php

josett225:
FIX BankTransfer link in rejets.php

This is to fix a wrong action link in rejets.php.
The link was sending to a WithDrawals screen creating confusion.


eldy:
@lvessiller-opendsi @rycks 
This PR is strange. It contains change to add the changelog of v18.0.7.  But the v18.0.7 was released but this change was not merged into branch v18 before (sorry i missed that. I focus on the tg content and did not check the branch v18).
It seems you have worked and created a duplicate branch for 18. I delete it because we must have only one branch for v18. We must use tags and not branch to flag minor versions.

Can you restart the building of the v18.0.7 ? It seems all steps defined into the makepack-howto.pl was not done into the branch v18 to build the package v18.0.7.  You probably did it into another branch so all changes were lost.

josett225:
Hi @eldy @lvessiller-opendsi @rycks 
Do I need to do something in my side?

PR #34446, opened 2025-06-13
QUAL replace $conf->XXX->dir_output by getMultidiroutput

Chloe-Moreau-2009:


eldy:
Don't forget to set the name of your PR to start with 
QUAL
FIX
NEW
I have modified to add "QUAL".

PR #34455, opened 2025-06-13
QUAL replace $conf->XXX->dir_output by getMultidiroutput

Chloe-Moreau-2009:
Bom module


hregis:
@eldy @Chloe-Moreau-2009

Massaction directory is ok for a list but doesn't make sense for a card! No?

PR #34476, opened 2025-06-13
NEW webportal : add number of element in top menu

defrance:


defrance (commented):
ok

defrance:
seems present in V23 so i close it

Issue #33966, opened 2025-04-24
Bad displaying of supplier price list


mapiolca:
### Bug

Between v20.0.4 and v21.0.1, a bug appears in the getNomUrl function of supplier price. 

### Dolibarr Version

21.0.1

### Attached files
v21.0.1
![Image](https://github.com/user-attachments/assets/5330ef5e-dcdb-4595-818d-2e428b04ecee)

v20.0.4
![Image](https://github.com/user-attachments/assets/d9771dfe-30d1-46d9-b6b6-93362f4c31f3)

PR #34514, opened 2025-06-14
fix: bad display in tooltips of supplier price history

FHenry:
fix: #33966


eldy:
Fixed by 79a6825e6fc37c5055d887601dea9730454869fe

PR #34518, opened 2025-06-14
$forceFullTextIndexation not initialised

defrance:


htdocs/core/actions_linkedfiles.inc.php
```diff
@@ -97,6 +97,7 @@
 				$generatethumbs = 0;
 			}
 			$allowoverwrite = (GETPOSTINT('overwritefile') ? 1 : 0);
+			$forceFullTextIndexation = '';
```
eldy, line 100: As this is an include file some variables are defined by the caller.
It is the case into upload_page.php for example.

PR #34524, opened 2025-06-14
NEW: rest api: add database structure for multiple tokens with limited rights

marc-dll:
# NEW: rest api: add database structure for multiple tokens with limited rights

Following today's rest API debate, here is a proposal for REST API tokens with limited rights.

Each user would be able to have 0,n tokens, each with specific rights (obvious limited to the user rights).

Ping @lmag (who told me @W1W1-M would also be interested).


eldy:
A table was already introduce to manage this. It is called
llx_oauth_token
Thanks to check you have all fields required to store your n tokens.

PR #34575, opened 2025-06-19
Prepare 18.0.7

lvessiller-opendsi:
Prepare 18.0.7

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,74 @@
 English Dolibarr ChangeLog
 --------------------------------------------------------------
 
+***** ChangeLog for 18.0.7 compared to 18.0.6 *****
+138 files changed, 1622 insertions(+), 530 deletions(-)
+
+FIX: 17.0 API endpoints "PUT": prevent overwriting all extrafields if only some are supplied in the request
+FIX: 17.0 API endpoints "PUT": prevent overwriting all extrafields if only some are supplied in the request cf. PR #29237
+FIX: 17.0 - collisions in cache for dol_getIdFromCode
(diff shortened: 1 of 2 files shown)


rycks (approved):
ok

eldy:
ko.  See CI feedbacks into "Files changed"

lvessiller-opendsi:
> ko. See CI feedbacks into "Files changed"

Yes, but It's on Changelog.md and packages of this realease are already on SourceForge and on Dolibarr server with theses mistakes.
What can I do ?

eldy:
> > ko. See CI feedbacks into "Files changed"
> 
> Yes, but It's on Changelog.md and packages of this realease are already on SourceForge and on Dolibarr server with theses mistakes. What can I do ?

Never mind, we should never have a package that is released and contains CI errors.
It is strange that this not appears before (may be because package was done on another branch,).
Having a red CI on the tag that identify a released LTS version is clearly not what users expect on a maintenance version, so we should fix it and rebuild the package.

lvessiller-opendsi:
@eldy 
All checks have passed.
What is the next step ? Rebuild all packages with "makepack-dolibarr.pl" script and send one on Dolibarr asso (option 98) and one to SourceForge (option 99) ?

eldy:
> @eldy 
> All checks have passed.
> What is the next step ? Rebuild all packages with "makepack-dolibarr.pl" script and send one on Dolibarr asso (option 98) and one to SourceForge (option 99) ?

yes, i would rebuild the package (98) and erase the tag (99). We should not erase a tag usually but here the tag 18.0.7 is linked to files that does not exist in v18 branch so it is better to redo it. No better idea to restore the situation.

PR #34582, opened 2025-06-19
FIX : Missing translation for en-US in user page

Gouttfi:
FIX : Added two missing translation keys for a user's page.
image
image


eldy:
Entry already exists but into file ticket.lang
So i moved them into main.lang file

PR #34588, opened 2025-06-19
NEW -  can set project with origin on object creation

kkhelifa-opendsi:
Fix missing functionality when the projet of the origin of the movement stock is set in the movement

See -> https://github.com/Dolibarr/dolibarr/commit/7863642daaae1fecf02a123b02900f7d9173246a


kkhelifa-opendsi:
@eldy The PHAN error is only a warning and the paramaters send to the method is good

PR #34594, opened 2025-06-20
Fix #34593

### LINKED ISSUE #34593: PHP Error when setting a proposal to "ordered" with ODT files
### Bug

PHP Error when setting a proposal to "ordered" with ODT files

### Dolibarr Version

21

### Environment PHP

8.x

### Environment Database

n/a

Activate the workflow module and enable automatic creation of CO when proposal is ordered
Create an ODT template for proposal with a mandatory extrafield
Create a proposal using the ODT template and set the order to "ordered"

### Attached files

_No response_

glu000:
FIX #34593
In some cases $this->errors is NULL. With PHP 7 this generated a PHP warning. Since PHP 8 it creates a fatal error

### DIFF
--- a/htdocs/core/triggers/dolibarrtriggers.class.php
+++ b/htdocs/core/triggers/dolibarrtriggers.class.php
@@ -168,7 +168,7 @@ abstract class DolibarrTriggers
 	 */
 	public function setErrorsFromObject(CommonObject $object)
 	{
-		$this->errors = $this->errors ?? [];
+		isset($this->errors) ?: $this->errors = [];
 
 		if (!empty($object->error)) {
 			$this->errors = array_merge($this->errors, array($object->error));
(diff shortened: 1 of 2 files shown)


glu000:
Again a PR which fails because of problems in a file where I made no changes. Can someone (maybe @eldy ) support me how to fix it? It would be also great to get some information how to prevent such problems in future.

eldy:
> Again a PR which fails because of problems in a file where I made no changes. Can someone (maybe @eldy ) support me how to fix it? It would be also great to get some information how to prevent such problems in future. 

You can't yourslef prevent this, as trouble is not in your file but into other pr except byvasking other users to stop to contribute. 

If you did not forget to check the box "allow othercto contribute", you have nothing to do. Other pr will be fixed and once an Update branch is run, the problem will disappear.

If you dudn't check the checkbox, you must make the update of branch yourself. 
But best is to chrck the chrckbox so all you have to do is nothing.For thus case it serls you didn't check the box, so you must do the pull/push yourself.

PR #34607, opened 2025-06-23
signed interventional fail if no status selected (-1)

defrance:


htdocs/core/class/commonsignedobject.class.php
```diff
@@ -90,6 +90,7 @@ public function setSignedStatus(User $user, int $status = 0, int $notrigger = 0,
 	{
 		global $langs;
 		$langs->loadLangs(array('commercial'));
+		$status == -1 ? $status = 0 : $status;
```
eldy, line 93: If setSignedStatus is called with status = 0, it means no status was defined so we can't know which one set.
It means there is a bug before. surely because be call setSignedStatus instead of reporting an error to say that function is called with bad parameters.

defrance: hello, another solution is to add 'select_show_empty' => 0, on confirm select signed
but if somebody valid the confirm form without select a signed statut we have a fail and the only way to correct the problem is this fix (or add on upgrade.sql an update query to change -1 to 0 on llx_fichinter table

PR #34619, opened 2025-06-24
FIX: migrate boolean fields

marc-dll:
# FIX: migrate boolean fields

Objects with fields typed as `'boolean'` (checkboxes in modulebuilder-generated modules) are handled with database type `BOOLEAN`. This database type in not handled very well with the PHP PostgreSQL driver : `pg_fetch_*()` functions return `string` values `'t'` or `'f'` respectively for true or false, which are both 'truthy' values, so the checkboxes are always considered checked. Moreover, it complicates table create/update queries as we have to specifically pass `TRUE` or `FALSE` in SQL queries to have cross-compatible behaviour.

I therefore suggest migrating all those fields from `BOOLEAN` to `TINYINT` (which will be rewritten to `SMALLINT` in PostgreSQL), migrating existing fields (in v18, only field `totalizable` from table `llx_extrafields` is defined as such), and getting modulebuilder SQL-generation feature to output `SMALLINT` instead.

When this PR is merged, another will follow in v19 where more fields use the `BOOLEAN` database type.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -9249,9 +9249,9 @@ abstract class CommonObject
 			return (int) $value;
 		} elseif ($fieldsentry['type'] == 'boolean') {
 			if ($value) {
-				return 'true';
+				return '1';
 			} else {
-				return 'false';
+				return '0';
 			}
 		} else {
 			return "'".$this->db->escape($value)."'";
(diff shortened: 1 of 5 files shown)


eldy:
As this change the database structure, it must be pushed into develop

rycks:
hello @marc-dll 

just an other idea, i don't know if it could be done ... 

there is a DoliDBPgsql class and i imagine we could fix that "bug" with a sort of converter for values then "t" could be converted to "true" and "f" to false for the 18.0 branch ?

and then your patch will be send to develop as @eldy says ?

eldy:
It seems this PR was set with a status asking a change with no answer nor correction was provided since a long time. So request is automatically closed.

PR #34620, opened 2025-06-24
# NEW|New - add a conf to manage picture by their position on product object

atm-corentin:
# NEW|New Add configuration setting to control product photo sorting order
I noticed that when a product has a certain number of associated photos—in my case, 12 photos—if I changed the order of the images, the photos displayed on the product card did not match the order shown in the list of attached files. After investigation, I found that the image processing logic did not take the "position" field . I have therefore added this processing.


atm-corentin:
@eldy can you check pls ?

atm-corentin:
@eldy the branch is merged with the current develop and all checks are green.

I also fixed two things in my own code while doing that:

- `core/class/html.formfile.class.php`: the sort was applied to `$filearray`, which does not exist in that method — the local variable is `$file_list`, so the sort was silently dropped in the document list.
- the setup constant was named `OrderPhotoByPosition`, it is now `MAIN_SORT_PHOTO_BY_POSITION` to follow the usual naming of setup constants. The translation key `OrderPhotoByPosition` is unchanged.

Can you have a look please?

htdocs/core/class/commonobject.class.php
```diff
@@ -10420,7 +10420,11 @@ public function show_photos($modulepart, $sdir, $size = 0, $nbmax = 0, $nbbyrow
 
 		if (count($filearray)) {
 			if ($sortfield && $sortorder) {
-				$filearray = dol_sort_array($filearray, $sortfield, $sortorder);
+				if (getDolGlobalInt('MAIN_SORT_PHOTO_BY_POSITION')) {
+					$filearray = dol_sort_array($filearray, 'position', $sortorder);
```
eldy: This looks strange because 2 line alive, we already do a sorting on $sortfield.
And sortfield is set by default to position_name so it should be sorted on position, then on filename.

atm-corentin: @eldy `$sortfield` is indeed hardcoded to `position_name` a few lines above, but that key does not sort by position the way you describe, for two reasons.

1. `position_name` is a concatenated **string**, built in `files.lib.php` as `<position>_<name>` with no padding, and `dol_sort_array()` is called here with `$natsort = 0`, so the values end up in `asort()` and compare lexicographically. Past nine photos the order breaks: position `10` sorts before position `2`.
2. `position_name` is only assigned when the file is matched in `ecm_files`. A photo present on disk but absent from that table never gets the key, `dol_sort_array()` maps it to `0`, and it jumps to the front whatever its position.

Sorting on the numeric `position` avoids both, since two numeric strings compare numerically.

That said, I can go the other way if you prefer: zero-pad the position when `position_name` is built in `files.lib.php` and drop the conf entirely. That repairs the default for every caller instead of adding an option, but it changes the sort for every `dol_dir_list()` consumer using `position_name`, which is a wider blast radius than this PR.

atm-corentin:
@eldy  can you check pls

PR #34623, opened 2025-06-24
FIX: Update document.php

prietojc:
In the Resources module, uploaded files are stored in folders with the resource reference instead of its ID, which causes all documents to disappear when the  reference is changed.

This fix need a script to rename folders.

### DIFF
--- a/htdocs/resource/document.php
+++ b/htdocs/resource/document.php
@@ -84,7 +84,7 @@ $object = new Dolresource($db);
 // Load object
 include DOL_DOCUMENT_ROOT.'/core/actions_fetchobject.inc.php'; // Must be 'include', not 'include_once'.
 
-$upload_dir = $conf->resource->dir_output.'/'.dol_sanitizeFileName($object->ref);
+$upload_dir = $conf->resource->dir_output.'/'.dol_sanitizeFileName($object->id);
 $modulepart = 'resource';
 
 $result = restrictedArea($user, 'resource', $object->id, 'resource');


eldy:
> In the Resources module, uploaded files are stored in folders with the resource reference instead of its ID, which causes all documents to disappear when the reference is changed.
> 
> This fix need a script to rename folders.

Using the Ref in directory name is the most common behaviour for most object. 
Most other object have the ref locked so there is no trouble. 
For project, what we do is that we rename the directory after a ref renaming. 
A better fix would be to rename the directory when we rename the ref, like we do for projects.
We want to use ref into directory because it allows easier management and maintenance when the user manipulate the directory/files directly on disk (the id is not a user known information).

prietojc:
Okay, understood. I thought it was a mistake. However, it has a side effect that unfortunately happened to me: a user left the reference blank and then deleted the resource. **All the folders were deleted!** Because not including a reference is equivalent to deleting the parent folder.

In invoices or orders, changing the reference is not common, even prohibited, but it is often done in resources. Perhaps it would be useful to add a `Label` field and block the reference.

eldy:
> Okay, understood. I thought it was a mistake. However, it has a side effect that unfortunately happened to me: a user left the reference blank and then deleted the resource. **All the folders were deleted!** Because not including a reference is equivalent to deleting the parent folder.
> 
> In invoices or orders, changing the reference is not common, even prohibited, but it is often done in resources. Perhaps it would be useful to add a `Label` field and block the reference.

On invoice, renaming ref is not possible, renaming is only done when we change from PROV (when invoice is not yet finished to be edited) to final ref when invoice is definitely validated. So ref is already blocked for invoice.
Seems the same for order. No way to change it once validated.

Only project and resources allows renaming their ref, this is why only this one should renamed the dir.

prietojc:
> a user left the reference blank and then deleted the resource. **All the folders were deleted!** Because not including a reference is equivalent to deleting the parent folder.

I think the solution might be to not allow any references to be left blank.

eldy:
> > a user left the reference blank and then deleted the resource. **All the folders were deleted!** Because not including a reference is equivalent to deleting the parent folder.
> 
> I think the solution might be to not allow any references to be left blank.

Strange. For me blank reference is already forbidden. May be a bug. I Did not succeed in creating a resource with a blank reference. May be it was created using API ?

prietojc:
@eldy , I found the bug: create a new resource, type letters (no numbers) in name field, then resource will be created with blank reference.

Line 211, in function create() from class dolresource:
`(isset($value) && $value > 0)`

Only numeric reference is allowed....

**Post Data: This happened in version 20, solved in 21 and later.**

PR #34634, opened 2025-06-25
NEW Add thirdparty search on api list 

defrance:
Add llx_societe on api list element call
like this we can filter by name of the company (or some other societe fields)

### DIFF
--- a/htdocs/projet/class/api_projects.class.php
+++ b/htdocs/projet/class/api_projects.class.php
@@ -219,6 +219,7 @@ class Projects extends DolibarrApi
 
 		$sql = "SELECT t.rowid";
 		$sql .= " FROM ".MAIN_DB_PREFIX."projet as t";
+		$sql .= " INNER JOIN ".MAIN_DB_PREFIX."societe AS s ON (s.rowid = t.fk_soc)";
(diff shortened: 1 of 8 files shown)


htdocs/projet/class/api_projects.class.php
```diff
@@ -219,6 +219,7 @@ public function index($sortfield = "t.rowid", $sortorder = 'ASC', $limit = 100,
 
 		$sql = "SELECT t.rowid";
 		$sql .= " FROM ".MAIN_DB_PREFIX."projet as t";
+		$sql .= " INNER JOIN ".MAIN_DB_PREFIX."societe AS s ON (s.rowid = t.fk_soc)";
```
eldy, line 222: A thirdparty is not mandatory so we must use a left join instead of inner join.

htdocs/ticket/class/api_tickets.class.php
```diff
@@ -224,6 +224,7 @@ public function index($socid = 0, $sortfield = "t.rowid", $sortorder = "ASC", $l
 
 		$sql = "SELECT t.rowid";
 		$sql .= " FROM ".MAIN_DB_PREFIX."ticket AS t";
+		$sql .= " INNER JOIN ".MAIN_DB_PREFIX."societe AS s ON (s.rowid = t.fk_soc)";
```
eldy, line 227: A thirdparty is not mandatory so we must use a left join instead of inner join.

PR #34636, opened 2025-06-25
FIX 21.0: $height and $width can be ints, but also 'auto'

atm-florianm:
# FIX
Formconfirm dialogs take a `$height` argument. Before v21.0, the value `auto` was accepted (and frequently used in external modules), but in v21.0, a typecast to int was introduced, which turns `auto` into `0` and makes some of the dialogs unusable.


eldy:
As we know that height and width are not complex structures but should be a scalar (int) or 'auto' i tried to restore the use of the ' and dol_escape_js to sanitize the js code.
Is it ok ?

atm-florianm:
Absolutely fine by me, thanks!

PR #34643, opened 2025-06-26
NEW: Add property 'donthavesocparent' on CommonObject class

kkhelifa-opendsi:
Add property 'donthavefksoc' on CommonObject class for manage the case of a fields of type 'link' to a class who dont have a field fk_soc when the multi entity is activated and we connect with a external user (when the function selectForFormsList() is used for display the select input on list and card)


eldy:
Due to other bug, i pushed a commit to fix them that is not compatible with your change (an i think your change is no more required).
What i had to do is to have a different test for testing entity and testing users that are external users. Both condition are now managed independently.
Can you check your problem is fixed with last branch wihtout your PR ?

eldy:

PR #34648, opened 2025-06-26
FIX|Fix  - Duplicate bommline when i refresh my page

atm-corentin:
# FIX|Fix #Duplicate BOMLine when i refresh BOM page

I noticed that when you add a BOM line to a BOM, and then refresh the page, the line you just added is duplicated.
To fix this and secure this type of operation, I added this code


atm-corentin:
@eldy

eldy:
Can't reproduce.
How do you refresh the page ? Just hitting F5 keyboard key ?

PR #34651, opened 2025-06-27
New: Being able to pay credit notes from TakePOS and having their own numbering mask

fmarcet:
Until now, when creating a credit note from TakePOS, it was created and validated directly, but it wasn't possible to enter the payment/refund or issue a partial refund for only certain lines. Now, the credit note is created as a draft, allowing lines to be edited, and then the refund is entered from the payment screen.

In addition, the option for credit notes to have a different numbering series than sales invoices has been added, which is mandatory in many countries.


fmarcet:
Hi @eldy, I think now is ok.

htdocs/takepos/invoice.php
```diff
@@ -309,7 +310,7 @@ function fail($message)
 				$payment->amounts[$invoice->id] = $amountofpayment;
 				// If user has not used change control, add total invoice payment
 				// Or if user has used change control and the amount of payment is higher than remain to pay, add the remain to pay
-				if ($amountofpayment <= 0 || $amountofpayment > $remaintopay) {
+				if (abs($amountofpayment) <= 0 || abs($amountofpayment) > abs($remaintopay)) {
```
eldy, line 313: abs(x) is <= 0 only if amountofpayment is 0.
Surely not what the result you expect.

PR #34663, opened 2025-06-29
UX Repair install script

aspangaro:
# UIUX|Uiux
Review of repair install script on UX

If modification is Ok, use language key rather than hard-coded

Before (no link, not very readable)

image

After
image


aspangaro:
@lmag Other idea ?

PR #34667, opened 2025-06-30
FIX: restrictedArea from ajaxtooltip return 0 if my class is not multi entity managed

jyhere:
Tested with a custom module / object, with no entity management, on a Dolibarr with multicompany module enabled.

A check on $object->ismultientitymanaged should be added, otherwise the sql query will fail.

Not sure if the fix is sufficient ? (tested with other objects, seems to be ok)


htdocs/core/lib/security.lib.php
```diff
@@ -1176,7 +1176,7 @@ function checkUserAccessToObject($user, array $featuresarray, $object = 0, $tabl
 					$sql .= " AND dbt.entity IN (".getEntity($sharedelement, 1).")";
 					$sql .= " AND (sc.fk_user = ".((int) $user->id)." OR sc.fk_user IS NULL)";
 				}
-			} elseif (isModEnabled('multicompany')) {
+			} elseif (isModEnabled('multicompany') && !empty($object->ismultientitymanaged)) {
```
eldy, line 1179: For better compatibility may be we can do
`elseif (isModEnabled('multicompany') && (!empty($object->ismultientitymanaged) || !isset($object->ismultientitymanaged)) {
`

JamBalaya56562:
Fixed by #39246, which has now been merged.

The fix guards the `multicompany` entity-check branch in `checkUserAccessToObject()` so that it only runs when the object is actually multi-entity managed:

```diff
-			} elseif (isModEnabled('multicompany')) {
+			} elseif (isModEnabled('multicompany') && (!empty($object->ismultientitymanaged) || !isset($object->ismultientitymanaged))) {
```

This follows exactly what you suggested here (checking `$object->ismultientitymanaged` before running the entity query) and the condition @eldy requested in review — the entity check is skipped only when `ismultientitymanaged` is explicitly `0`, and is kept for legacy classes that never set the property (backward compatible).

Merged to `21.0` and propagates upward to `develop` via the normal merge-up flow.

Thanks for the report — this can be closed. 🙏

PR #34674, opened 2025-07-01
FIX : module builder objects not directly extending CommonObject

jyhere:
Changed the class detection in the objects tab of Module Builder : 
- exclude abstract classes.
- remove CommonObject in the the regex
- check for CommonObject inheritance using reflection.

This allows classes that are not direct children of CommonObject to be properly displayed

Example : 
Class A extends Class B
Abstract Class B extends CommonObject

Before : 
- Class A not displayed in module builder
- Abstract Class B is displayed in module builder (and generate errors)

 After :  
- Class A is displayed in module builder.
- Abstract Class B no more displayed in moduleBuilder, no more errors


eldy:

PR #34676, opened 2025-07-01
Added Phone Capture to Ticket module

burimikuqi:
# NEW #Added phone capture to ticket module

[New Feature Added to the Ticket Module: Phone Capture for Attachments
A new functionality has been integrated into the ticket module that allows users to capture photos directly using their mobile device and upload them as attachments to a ticket. This feature streamlines the process of reporting issues or providing visual context by enabling:
- Direct access to the phone camera from the ticket interface
- Instant photo capture without needing to browse files manually
- Automatic upload and attachment of the captured image to the corresponding ticket
This enhancement improves usability, especially for mobile users in the field, and ensures faster, more accurate reporting.]

### DIFF
--- a/htdocs/langs/en_US/ticket.lang
+++ b/htdocs/langs/en_US/ticket.lang
@@ -1,3 +1,4 @@
+
 # en_US lang file for module ticket
 # Copyright (C) 2013  Jean-François FERRY <hello@librethic.io>
 #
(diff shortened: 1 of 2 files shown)


htdocs/langs/en_US/ticket.lang
```diff
@@ -178,6 +179,7 @@ TicketHistory=Ticket history
 AssignUser=Assign to user
 TicketAssigned=Ticket is now assigned
 TicketChangeType=Change type
+ThirtParty=Thirt Party
```
eldy, line 182: You probably want to use ThirdParty ? (this entry is already defined).

eldy:
I don't understand. All this features seems already available with current version.
Tested with develop branch, i already have the camera selection automatically avaible with direct access after clicking the + to add a file or image, and files joined from ticket page.
What is difference with current behaviour ?
May be you can add sceenshots of what has changed ?

burimikuqi:
Hi, till this develop was not available not from iOS or android any way version that I'm using is 20.00 and till there was not available, and as you mentioned that is available now I've deleted forked repository, version mod that I had made was also to include date and time stamp (on file name)  on each photo "only photo" captured via phone.

In any way thank you for your prompt response.

PR #34677, opened 2025-07-02
# FIX|Fix re-use of the THIRDPARTY_CUSTOMERTYPE_BY_DEFAULT variable when creating a new third party

AlainRnet:
Since the improvement of the Prospect/Client/Supplier selection, the THIRDPARTY_CUSTOMERTYPE_BY_DEFAULT variable is no longer taken into account. This code restores this functionality when creating a new third party.


AlainRnet:
Sorry, replaced by #34702

PR #34686, opened 2025-07-03
18.0 create SOURCES dir if not exists

rycks:
On @lvessiller-opendsi  setup rpmbuild/SOURCES directory does not exists and makepack.pl fails due to that


rycks:
forget that PR i mix two commit shame on me :-(

PR #34690, opened 2025-07-04
Replaces useless calls to dol_now() and dol_now('gmt') with the native PHP function time()

loxK:
Because `dol_now()` and `dol_now('gmt')` do exactly the same thing as `time()`, thus returning a Unix timestamp that, by definition, is time zone independent.

**Note :** we should ultimately get rid of the dol_now() function that has the ability of returning "timezoned" timestamps which is a nonsense. Timezone work should be done when processing timestamps (for display, database insert...)


eldy:
time() function must never be used as it is not able to manage all timezone cases we need.
For this reason, we must use dol_now() everywhere instead with a parameter depending of what we need.
dol_now depends on server timezone (all the code currently is aligned with server timezone)
dol_now gmt does not depend on any timezone.
See page Language and development rules — Dolibarr ERP CRM Wiki https://share.google/i4603tKcYTlJb6ZTp about timezone.

loxK:
You are wrong, a timestamp is timezone independent : "Unix time is currently defined as the number of non-leap seconds which have passed since 00:00:00 **UTC** on Thursday, 1 January 1970"

Also, `dol_now()` and `dol_now('gmt')` [are both returning time()](https://github.com/Dolibarr/dolibarr/blob/dff214402ec71a3e145ca72bf3d1a0f7442813ea/htdocs/core/lib/functions.lib.php#L4002). 

How is it a good idea to do something like that, instead of using directly  the native function  :

```
<?php

function dol_now() 
{
 return time()
}
```

Because this is basically what calling `dol_now()` and `dol_now('gmt')` [is doing](https://github.com/Dolibarr/dolibarr/blob/dff214402ec71a3e145ca72bf3d1a0f7442813ea/htdocs/core/lib/functions.lib.php#L4002).

PR #34714, opened 2025-07-08
FIX wrong multicompany filter

hregis:
@eldy testé et approuvé ! ;-)

### DIFF
--- a/htdocs/projet/list.php
+++ b/htdocs/projet/list.php
@@ -105,8 +105,7 @@ $pageprev = $page - 1;
 $pagenext = $page + 1;
 
 $search_all = GETPOST('search_all', 'alphanohtml');
-$search_entity = ($user->entity > 0 ? $user->entity : GETPOSTINT('search_entity'));
-
+$search_entity = GETPOSTINT('search_entity');
 $search_ref = GETPOST("search_ref", 'alpha');
 $search_label = GETPOST("search_label", 'alpha');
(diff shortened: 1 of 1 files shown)


hregis:
@eldy un petit dernier fix et il faut sortir la 21.0.2 !! ça urge :-)

hregis:
@eldy in child entities, the entity's projects are not displayed in the list when using cross-entity project sharing.

without the fix : 

Capture d’écran 2025-07-11 à 08 50 53


with the fix : 

Capture d’écran 2025-07-11 à 08 51 40

eldy:
Can you confirm that the change on line 108 is enough to fix the bug ? I see no relation with feature on line 597.

eldy:
I am still not able to reproduce the problem. It seems there is another bug.
For example on your first screenshot, if we execute the code, we reach the line 598 with search_entity = 2 for exemple if user has entity 2.  So the where show only the project in entity 2 and if there is none we have nothing found (your second screenshot). 
But if search_entity is 2 line 598, the combo list "Environment" should be preselected on the entity 2 to show that the list is empty after a filter on entity 2 and this is not the case on your second screenshot.
So the bugs here seems also into how the combo list is selected. So it seems there is 2 bugs on your side. 

But on my side, i have none on them. 
Strange. The fix seems ok too (everything still works on my side), but you sureley have another trouble somewhere else.

PR #34717, opened 2025-07-09
FIX: Add security test for show terminal selection if no terminal selected when invoice.php is call

kkhelifa-opendsi:
Add security test for show terminal selection if no terminal selected when invoice.php is call


rycks (approved):
ok

eldy:
@kkhelifa-opendsi  For my information, how do you reproduce the bug (the information is required to validate the fix) ?

lvessiller-opendsi:
> @kkhelifa-opendsi For my information, how do you reproduce the bug (the information is required to validate the fix) ?

To reproduce:
- Terminal remains open with an empty basket
- Session expires
- Add a product to the basket
- Invoice.php is called and a draft invoice is created. Since the session variable is no longer set, the invoice is created as a draft on terminal 0, which does not exist.

lvessiller-opendsi (approved):

PR #34719, opened 2025-07-10
NEW: Management for REST API tokens

YannisHoareau:
This PR changes how current REST API tokens are managed. Currently, the user only has one token defined on him and stored in user table. Now we can enable/disable the API for the user and manage his tokens in a new tab. The user will be able to have multiple tokens, stored in oauth_token, and to choose a set of permission for each one. It is not possible for the token to get a permission that the user does not have, and the token will lose the permission if the user loses it (himself or through a group permission). This follows #34671 that migrates the existing tokens in user into oauth_token.


eldy:
Migrating the api into llx_oauth_token is ok.
However the management of permission per token must be managed in a second step to make transition easier. Also for a better architecture i think permissions on token must be a second layer of permission over the existing one that must remain untouched. So we can add the permission layer in a second step.
So i will start to modify the PR, in this direction, to remove all part related to the permission and keep the feature at the same level than now but with a management of n token per user.

eldy:
@YannisHoareau  Is PR complete ? What is still missing to be merged ?

PR #34723, opened 2025-07-10
Fix(core): Ensure tms field is always updated in updateCommon

atm-quentin:
#### The Problem

When updating an object via the `updateCommon` method, the `tms` (modification timestamp) field was not being consistently updated with the current date and time. The existing value stored in the object was often reused in the `UPDATE` query, which prevented accurately tracking the last modification date.

This behavior was noticed on several objects, for instance, the `Candidature` (Application) object.

#### The Solution

This change aligns the behavior of the `tms` field with that of `date_modification`, which is already correctly updated.

I have added a condition to the `updateCommon` method that checks for the existence of the `tms` field in the object being modified. If it exists, its value is now systematically replaced with the current timestamp (`dol_now()`) before the SQL query is built.

```php
if (array_key_exists('tms', $fieldvalues)) {
   $fieldvalues['tms'] = $this->db->idate($now);
}
```

#### Impact

This fix ensures that the modification timestamp (`tms`) is now reliable for all objects that use the `updateCommon` method. This improves data consistency and the traceability of modifications across the entire application.


htdocs/core/class/commonobject.class.php
```diff
@@ -10776,7 +10776,7 @@ public function updateCommon(User $user, $notrigger = 0)
 
 		$error = 0;
 
-		$now = dol_now();
+		$now = dol_now('tzuser');
```
eldy, line 10779: Not so easy, because currently the tms field is updated by a database trigger (a bad thing). This means the field is the only field to be updated using the timezone of the database that is not under control of the developer (very bad). Only this field is in this situation.

Having it updated during the update is a better way, but if we fix this definitely (and we should), it means the timezone may differ and we may experience few troubles during transition.

So i suggest to introduce a hidden constant
if (getDolGlobalString('MAIN_DISABLE_AUTO_UPDATE_OF_TMS_FIELDS')) {

So we will be able to start to migrate the code to definitely abandon the "database trigger" for the insinde SQL update because this trouble exists also on old school objects and we must migrate them too.

Can you add this ? We will be able to switch to new mode once it will be implemented in every hard coded updates.

PR #34724, opened 2025-07-10
NEW hook allowing external modules to replace the behavior of fetchOb…

atm-florianm:
# NEW Hook for fetchObjectLinked
As it stands, there is no obvious way for external modules to alter the fetching of linked object when it is performed by Dolibarr.

Our problem is with huge objects linked to many huge objects. The object's card calls `Form::showLinkedObjectBlock()`, which calls `$object->fetchObjectLinked()` (before the existing hook, so not modifiable) in order to perform `getNomUrl()` on each object.

When every object is big (many objects with many lines each), this easily exceeds performance limits.

With this hook, we can selectively choose to load less data / load it more effectively depending on the call context.

Note on design: I considered using the hookmanager's results array, but decided it was more convenient to let the implementer manipulate `$object->linkedObjectsIds`, `$object->linkedObjects` and `$object->linkedObjectsFullLoaded` directly, since those are public. 

[edit] *I just noticed that `linkedObjectsFullLoaded` is, in fact, private, so this will be a limitation of the hook.*

Note on overload: currently, the method `fetchObjectLinked` is very rarely overloaded in the core, but adding this hook means that 

Don't hesitate to give me feedback if you think this design was better.

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -4381,6 +4381,22 @@ abstract class CommonObject
 		$this->linkedObjectsIds = array();
 		$this->linkedObjects = array();
 
+		// Hook for allowing modules to completely alter the behavior of the method
+		$parameters = array(
+			'sourceid' => $sourceid,
+			'sourcetype' => $sourcetype,
+			'targetid' => $targetid,
+			'targettype' => $targettype,
+			'clause' => $clause,
+			'alsosametype' => $alsosametype,
+			'orderby' => $orderby,
+			'loadalsoobjects' => $loadalsoobjects
+		);
+		$reshook = $hookmanager->executeHooks('fetchObjectLinked', $parameters, $this, $action); // Note that $action and $object may have been modified by some hooks
+		if ($reshook > 0) {
+			return $reshook;
+		}
+
 		$justsource = false;
 		$justtarget = false;
 		$withtargettype = false;


eldy:
The fetchObjectLinked() should not be a load trouble if we can call it with parameter  loadalsoobject = 0    (the recommanded method)

So i tried to check if we can use 0 instead of 1. But I didn't see why we need to call fetchObjectLinked() to execute the getNomUrl (i see no reason why we do this if we do it, so trouble may be before). On which class/object for example do you have this ?

atm-florianm:
@eldy 

> The fetchObjectLinked() should not be a load trouble if we can call it with parameter loadalsoobject = 0 (the recommanded method)


The method `Form::showLinkedObjectBlock()` is called by almost every card.php in Dolibarr.

To summarize how it is implemented:
```php
	public function showLinkedObjectBlock($object, $morehtmlright = '', $compatibleImportElementsList = false, $title = 'RelatedObjects')
	{
		global $conf, $langs, $hookmanager;
		global $bc, $action;

		$object->fetchObjectLinked();
// […]
		foreach ($object->linkedObjects as $objecttype => $objects) {
 			// […]
 			global $linkedObjectBlock;
 			$linkedObjectBlock = $objects;
			$res = @include dol_buildpath($reldir.'/'.$tplname.'.tpl.php');
 			// […]
		}
```

And in the tpl (I didn't check every tpl, but they are often similar):
```php
foreach ($linkedObjectBlock as $key => $objectlink) {
	// […]
	print '<td class="linkedcol-name tdoverflowmax150">'.$objectlink->getNomUrl(1).'</td>';
	// […]
}
```

So `fetchObjectLinked()` is not called by my module, it is called by Dolibarr. And if it were called with `loadalsoobject = 0`, by definition, the linked objects wouldn't be fetched. This would in turn cause `getNomUrl()` to fail because it requires at least some attributes to be set, like `ref`, `total_ht` etc., depending on the element type.

I do agree, however, that, for most objects, `getNomUrl()` doesn't need the object to be fully fetched (with all its lines and all its attributes). That's the basis of my client-specific workaround.

> On which class/object for example do you have this

We encountered critical performance issues on:
- invoice cards (our client's invoices can be linked to a big number of contracts, and each contract can have hundreds of lines)
- contract cards (contracts can be linked to many invoices, and some invoices can have up to 20000 lines, and each line has a handful of extrafields as well).

We also had some other performance issues with extrafields that are links to objects when shown on lists (for the same reasons: for each line shown having this link extrafield, the target object needs to be fetched for `getNomUrl()`. Fetching always fetches all lines, all extrafields and all lines' extrafields, even though this data is (mostly) unnecessary just to show a link.

However, I expect identical problems with other object types as long as there are lots of links and lots of lines involved.

atm-florianm:
Hi @eldy, have you been able to review my answer?

eldy:
> Hi @eldy, have you been able to review my answer?

Not yet. I am still focusing on debugging v22.

PR #34725, opened 2025-07-10
FIX: asset: could not select invoice in disposal pop-in

marc-dll:
# FIX: asset: could not select invoice in disposal pop-in

When disposing of an asset, the combobox to select a customer invoice cannot be displayed :
* simulating a field by editing `$fields` does not work as the object is instantiated from scratch in `CommonObject`
* the universal filter syntax is not correct anymore anyway

To fix that, instead of simulating a field to generate the combo, we make the corresponding call in `Form` class, with the correct universal filter.

This bugfix is valid up to `develop` branch.


aspangaro:
I confirm this bug. See the result below

Capture d’écran du 2025-07-19 07-58-28

lvessiller-opendsi (approved):

PR #34731, opened 2025-07-11
FIX - The test for fetch the third party is not necessary for send the reminder by email to the user

kkhelifa-opendsi:
The test for fetch the third party is not necessary because the thirdparty is not required on a event and it's not required for send the reminder by email to the user.


lvessiller-opendsi (approved):

eldy:
@lvessiller-opendsi As this is a pure quality/performance fix (no bug visible by end users seems corrected), we should avoid to add it into a maintenance branch. I have moved it into develop

If we follow the diagram of choice, https://cloud.dolibarr.org/s/QAd4dsbpQyZcazZ, we see the case use the path straigth down on the left path until the section "Not a bug for amaintenance branch" then as we are removing code, we can imagine that after a merge on different version, removed code may have an effet if code in higher version rely on it, so path bring us to the "To merge in develop".

PR #34732, opened 2025-07-11
feat(extrafields): add extrafields hook points for enhanced customization

atm-quentin:
# New add extrafields hook points for enhanced customization
    - Introduced multiple new hook points (`printFieldListTitle`, `printFieldListValue`, `formObjectOptions`, `loadExtrafieldsAttributes`, etc.) to allow enhanced customization and extension of extrafields functionalities.
    - Updated templates and extrafields class to execute hooks for additional flexibility in rendering and data manipulation.
    - Ensured backward compatibility with existing extrafields implementations.

### DIFF
--- a/htdocs/core/class/extrafields.class.php
+++ b/htdocs/core/class/extrafields.class.php
@@ -82,6 +82,7 @@ class ExtraFields
 		'double' => 'Float',
 		'date' => 'Date',
 		'datetime' => 'DateAndTime',
+		'datetime' => 'DateAndTime',
 		'duration' => 'Duration',
 		//'datetimegmt'=>'DateAndTimeUTC',
 		'boolean' => 'Boolean',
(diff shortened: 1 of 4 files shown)


htdocs/core/class/extrafields.class.php
```diff
@@ -1036,14 +1037,21 @@ public function fetch_name_optionals_label($elementtype, $forceload = false, $at
 		// We should not have several time this request. If we have, there is some optimization to do by calling a simple $extrafields->fetch_optionals() in top of code and not into subcode
 		$sql = "SELECT rowid, name, label, type, size, elementtype, fieldunique, fieldrequired, param, pos, alwayseditable, perms, langs, list, printable, totalizable, fielddefault, fieldcomputed, entity, enabled, help, aiprompt,";
 		$sql .= " css, cssview, csslist";
+		$parameters = ['elementtype' => $elementtype, 'forceload' => $forceload, 'attrname' => $attrname];
+		$hookmanager->executeHooks('printExtrafieldsSelect', $parameters, $this, $action);
```
eldy, line 1041: The goal of fetch_name_optionals_label is to load the definition of the system table extrafields.
There is no customization of this system table, because this table contains data to customize the business object but is not customizable itself.

Can you describe your use case why you need extrafields on extrafields table ?

atm-quentin: for a customer we want to add specific fields on extrafields
we made a specific join table, but we need to load thoses fields in extrafield list / view

eldy: Sorry i don't understand what you mean with adding extrafields on extrafields.
Can you describe your use case why you need extrafields on extrafields table ?

PR #34733, opened 2025-07-11
FIX: Mo::deleteLine() - stock movement correction

altairis-noe:
# FIX: Mo::deleteLine() - stock movement correction should not rely on linked lines

When we have to delete an MoLine from the mo_production page, we might remove a line with the `toconsume` role (which seems like a terrible idea to me), in which case we want to restock the `consumed` role lines which are linked. Hence the `$this->fetchLinesLinked`

But we can also remove a `consumed` role line (if we made an error in qty, typically)

image

In which case, there are no line linked. Currently, that means there is no correcting stock movement generated, which is an obvious problem.

I deleted several `consumed` lines here

image

But before I made the modifications here proposed, no increasing stock movement where created, until after I changed the code

image

As we can see, only two products out of eight were restocked.

### DIFF
--- a/htdocs/mrp/class/mo.class.php
+++ b/htdocs/mrp/class/mo.class.php
@@ -1,8 +1,9 @@
 <?php
-/* Copyright (C) 2017  Laurent Destailleur <eldy@users.sourceforge.net>
- * Copyright (C) 2020  Lenin Rivas		   <lenin@leninrivas.com>
- * Copyright (C) 2023-2024  Frédéric France     <frederic.france@free.fr>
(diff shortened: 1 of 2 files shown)


altairis-noe:
Whoops, indeed!

I don’t even know why I missed your feedback…

eldy:
What do you think of this better fix:
If a line "toconsume" has at least 1 line linked "consumed", the picto "delete line" should not appear or must be greyed.
I think the feature for heare for the case of production has not yet started.
Is it what you did ?

altairis-noe:
Good thinking, I updated the production card, so now we have : 
image

PR #34746, opened 2025-07-15
fix translation for non latin languages

sonikf:


htdocs/admin/translation.php
```diff
@@ -649,7 +649,7 @@
 			$titleforvalue .= '<span class="opacitymedium">'.$langs->trans("None").'</span>';
 		}
 		print '<span title="'.dolPrintHTMLForAttribute($titleforvalue).'" class="classfortooltip">';
-		print dolPrintHTML($val);
+		print dol_escape_htmltag($val);
```
eldy, line 652: A we are outputing a string on HTML, we must use 
dolPrintHTML and not dol_escape_htmltag (dol_escape_htmltag is an old one not secured).
Can you paste the value of $val you have that is wrong with dolPrintHTML ?

PR #34748, opened 2025-07-16
Fix webhook trigger history implementation

Hystepik:
Implement webhook trigger history.

This history add the possibility to record triggered webhooks.
If target is of  "blocking" type then only Successful triggered webhooks will be saved
If target isn't of "blocking" type then Successful and Unsuccessful triggered webhooks will be saved


hregis:
@Hystepik @eldy missing the management of the field "entity" in "llx_webhook_history" ! thank you

PR #34752, opened 2025-07-16
FIX Issue #34750

### LINKED ISSUE #34750: Minimal Price of products not verified
### Bug

As said in title, when I try to ADD a price under the min price in a proposal, Dolibarr verify and reject if the price is under the limit. but when I MODIFY the price on a line already add, I can enter a price under the min price, and Dolibarr let me do.

@eldy 

### Dolibarr Version

21.0.1

### Environment PHP

8.2

### Environment Database

mysql

1 - Create a draft proposal with a restricted access user.
2 - Add a line with a preselected product which have a minimal price (ex. 100€) with a price under the limit.
3 - Dolibarr reject your request
4 - Add a line with a preselected product which have a minimal price (ex. 100€) with a price upper or equal the limit.
5 - Dolibarr lets you do.
6 - Modify the previous added line with a price under the limit.
7 - Dolibarr lets you do. (Not normal)

### Attached files

_No response_

mapiolca:
Fix an issue with the "Check Minimum Price" function in updateline section.

# FIX|Fix #34750 

# CLOSE|Close #34750


mapiolca:
Hello @eldy, there you will have to help me because I think I understand that the error comes from the fact that I removed the keywords "(float) in front of the variables, but on the other hand, the problem also came from said keywords... So I don't really know what to do...

eldy:
There is no reason that the (float) before the price2num generate a trouble, so no reason to remove it.
What is the value you have as result of price2num that can't be converted into a float ?

mapiolca:
Because if a user, who has the right to override the minimum price, changes the price of the line and if a user who does not have the right to override the minimum price changes the price, this generates an error without giving a message. So this message alerts about this situation.Le 26 juil. 2025 à 14:55, Laurent Destailleur ***@***.***> a écrit :﻿
@eldy commented on this pull request.



In htdocs/comm/propal/card.php:
> @@ -1636,6 +1641,9 @@
 					$error++;
 					$action = 'editline';
+				} else {
+					setEventMessages($langs->trans("ErrorMinPriceCheck", $langs->transnoentitiesnoconv("Type")), null, 'errors');

Why do you output an error message here ?
If there is an error message here, it means you always report an error message (because all the cases of the if containes an error message output).

eldy:
> Because if a user, who has the right to override the minimum price, changes the price of the line and if a user who does not have the right to override the minimum price changes the price, this generates an error without giving a message. So this message alerts about this situation.Le 26 juil. 2025 à 14:55, Laurent Destailleur ***@***.***> a écrit :﻿
> @eldy commented on this pull request.
> 
> 
> 
> In htdocs/comm/propal/card.php:
> > @@ -1636,6 +1641,9 @@
>  					setEventMessages($mesg, null, 'errors');
>  					$error++;
>  					$action = 'editline';
> +				} else {
> +					setEventMessages($langs->trans("ErrorMinPriceCheck", $langs->transnoentitiesnoconv("Type")), null, 'errors');
> 
> Why do you output an error message here ?
> If there is an error message here, it means you always report an error message (because all the cases of the if containes an error message output).
> 
> —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

But it seems after your change the appli will always return an error message even if no error (the setEventLessage do output in all if cases)
Can you describe the process to reproduce the bug ?

mapiolca:
For sure :

Step 1: Create a sales proposal and add a product configured with a sale price and a minimum sale price.

Step 2: With User A having permission to ignore the minimum sale price, set a price lower than it.

Step 3: Change User B to one without permission to ignore the minimum sale price.

Step 4: Try changing the price.
Two scenarios arise:
- If the price entered by User B is lower than the minimum sale price, the standard error message will appear.
- If the price entered by User B is higher than the minimum sale price, the application rejects the request silently.

The patch I'm submitting addresses this issue. Only a user with permission to ignore the minimum sale price can edit the line.

eldy:
i tested the process in v22 and i can't reproduce.
Are you able to check in v22 ?

mapiolca:
Yes, it seems resolved in v22, may be you could patch only v21.

mapiolca:
@eldy ? What do we do for 21.0 ?

eldy:
> @eldy ? What do we do for 21.0 ?
Can the bug be reproduced in v22 or is it only in v21 ?

mapiolca:
No, as said higher, it seems this bug is only on 21.0

PR #34760, opened 2025-07-17
Correction for the generation of the bankjournal for a paiement for different third parties

kollibar:
In the case of use the option  "Allow payments on different third parties bills but same parent company", a payment can concern different third parties.
In this case, the actual bank journal generation is wrong because the total amount for the payment is used for each line. So if there is 2 lines, the payment line is doubled and each company line is quadrupled. If the payment concern 3 third parties, the payment line is tripled....

I've made a bug issue about it : https://github.com/Dolibarr/dolibarr/issues/34755

I propose here a correction for the amount of each line. The titled of each line should be also corrected but I didn't already do it

### DIFF
--- a/htdocs/accountancy/journal/bankjournal.php
+++ b/htdocs/accountancy/journal/bankjournal.php
@@ -298,9 +298,6 @@ if ($result) {
 		// Set accountancy code for thirdparty (example: '411CU...' or '411' if no subledger account defined on customer)
 		$compta_soc = 'NotDefined';
 		$accountancy_code_general = 'NotDefined';
-
-		$socid = $obj->socid;	// the socid concerned by the current row
-
 		if ($lineisapurchase > 0) {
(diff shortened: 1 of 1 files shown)


eldy:
PR seems empty.
Yu may try to make another commit to sugest another fix or close the PR.

kollibar:
Sorry for the different commit I'm begginer using git and the code control

eldy:
No pb with different commit, they will be all merged in one at the final merge into code. Don't worry for that.

It seems the bug is only when option FACTURE_PAYMENTS_ON_DIFFERENT_THIRDPARTIES_BILLS is on.
This option is still a touchy feature; To help impact analysis of this function, it would be better if the code change for your fix can also be executed only when option FACTURE_PAYMENTS_ON_DIFFERENT_THIRDPARTIES_BILLS is on.
Can you try to modify your PR so when option is off, old code is unchanged and when option is on, your new code is executed ?

kollibar:
OK, I have modified the code as asked !

PR #34798, opened 2025-07-24
NEW Add tpl loader for discounts.tpl.php

comaiteseb:
QUAL .tpl.php loader in /compta/facture/card.php to permit overload by a module and respect the fact that all /core/tpl can be overload by a module


comaiteseb:
Hi @eldy ,

I don't understand why my PR failed the phan test. The only warnings are for files missing from my PR.

Issue #34781, opened 2025-07-22
Add response header in getURLContent function


atm-maxime:
### Feature Request

The getURLContent function in https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/core/lib/geturl.lib.php allow us to call external URL such as APIs. Sometimes the URL we call returns informations in the response header.
Currently the getURLContent function doesn't read the response header.
The goal is to read it and return it in $rep['header'] just as the $rep['content'] is returned

eldy:
Should be possible by adding curl_setopt($ch, CURLOPT_HEADER, true); We do it when we want to pass a header to the caller, we should do it every time if an option to have the header in response is on.

atm-maxime:
@kaantopkaya who'll be the first to assign this issue ? 😉

PR #34824, opened 2025-07-29
NEW : add hidden const to get response header in geturlcontent function (#34781)

atm-maxime:
@kaantopkaya can you test this code on your side ?
@eldy is it what you had in mind ?

I don't know if when the curl request is in error we get headers in the response nevertheless...


kaantopkaya:
@atm-maxime i'm getting properly the getURLContent(...)['header'] in result, even if the curl request is in error

atm-maxime:
> @atm-maxime i'm getting properly the getURLContent(...)['header'] in result, even if the curl request is in error

Are you sure ? I didn't set the 'header' when curl_errno($ch) > 0

kaantopkaya:
> > @atm-maxime i'm getting properly the getURLContent(...)['header'] in result, even if the curl request is in error
> 
> Are you sure ? I didn't set the 'header' when curl_errno($ch) > 0

It's strange because i broke url and the headers that i'm sending and i receive a getURLContent(...)['header'] 

I have NULL if i put the code in the condition `if (curl_errno($ch)) {`

PR #34825, opened 2025-07-29
Fixed the display order of product/service images.

SylvainLegrand:
The name of the sorting field does not exist!


htdocs/core/class/commonobject.class.php
```diff
@@ -8706,7 +8706,7 @@ public function show_photos($modulepart, $sdir, $size = 0, $nbmax = 0, $nbbyrow
 		include_once DOL_DOCUMENT_ROOT.'/core/lib/files.lib.php';
 		include_once DOL_DOCUMENT_ROOT.'/core/lib/images.lib.php';
 
-		$sortfield = 'position_name';
+		$sortfield = 'position';
```
eldy, line 8709: position_name means sort on position then name.

rycks: @eldy why did you say "position then name" ?

i can't read something like explode by "_" then apply sort by each part of that keyword ...

eldy: If $sortfield = 'position' at line 8709, then 
later at line 8740, when we call dol_dir_list() with this value as sort criteria, we should get a result that is sorted in a predictive order. With just 'position', the result will be random when 2 files are at same position in llx_ecm_files table.

SylvainLegrand: @eldy 
As @rycks points out, there is no provision for multiple sorting using the "_" character as a condition separator.
Neither in the dol_dir_list() nor in the dol_sort_array() function, which are called for processing.

eldy: Sorry, the use of position_name is not into dir_list but line 8752 into 
$filearray = dol_sort_array($filearray, $sortfield, $sortorder);
And the property is not filled by dol_dir_list (because list on dir can't have duplicate position, there is no position know yet) but by method dol_dir_list_in_database()   called by completeFileArrayWithDatabaseInfo line  8748.

This is what do the function.
It list al files with dol_dir_list.
It complete with info not into the filesystem but into the file indexing table (so it complete the position, the keyword, and other meta information from file index table)
Then we sort on position then name to get a predictive order.

rycks:
Hello all,
i think there is a use case of the problem (maybe, @SylvainLegrand please ask me if that's the use case of the bug) ?

after long time with @lvessiller-opendsi the "problem" (for human) could be due to file name case (lower case in first, then a >> z >> A >> Z) ?

like that:

image

and please have a look on ecmfile database : is there a position entry ? and is your file into database or not ?

PR #34839, opened 2025-07-30
Add precautionnary verification

Quentin-Seekness:
Prevent the case (had one just today) where a discount except would be created without multicurrency amounts in db, wich ends up to a bad "remaintopay" calculation on the invoice list (and maybe other side effects).


eldy:

PR #34840, opened 2025-07-30
FIX English names and add missing sepa countries

sonikf:


hregis:
@sonikf @eldy  I'm not sure Monaco should be removed from the Eec!? 

https://en.gouv.mc/Policy-Practice/Monaco-Worldwide/Monaco-and-the-European-Union#:~:text=The%20Principality%20of%20Monaco%20is,Ambassador%20to%20Brussels%20in%201999.

htdocs/install/mysql/data/llx_00_c_country.sql
```diff
@@ -530,9 +531,9 @@ UPDATE llx_c_country SET numeric_code = '894' WHERE code_iso = 'ZMB';
 UPDATE llx_c_country SET numeric_code = '716' WHERE code_iso = 'ZWE';
 
 
-
+-- Source: https://www.europeanpaymentscouncil.eu/document-library/other/epc-list-sepa-scheme-countries
```
sonikf, line 534: Hi @hregis 
I provided source..


Also from your link

hregis:
@sonikf @aspangaro @eldy 

This will create side effects with the "isInEEC()" function which allows you to check the presence of the intra-community VAT number in certain cases for Monaco !

sonikf:
Hi,

I did some research and according to EU directive last version of 14/4/2025
https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX:02006L0112-20240101#B-6
in Article 7 and tables in https://taxation-customs.ec.europa.eu/taxation/vat/vat-directive/how-does-vat-work/territorial-scope_en

transactions originating in or intended for the Principality of Monaco are treated as transactions originating in or intended for France which means that for transactions in and out of Monaco with an EU country a French vat id and French vat rates must be used so @hregis you are right and thanks about "isInEEC()" function and intra-community VAT for Monaco, **changes have been reverted**

* Above  also apply to Sovereign Base Areas of Akrotiri and Dhekelia in Cyprus where Cyprus vat id and vat rate must be used
* Article 6 excludes some territories some with iso code and some without.

European Economic Community (EEC) does not exist anymore as it was a former iteration of European Economic Area (EEA) so relevant docs must be updated to sth like  
`Return if for a country of an object intra-community VAT rules apply`

https://github.com/Dolibarr/dolibarr/blob/499dd4a15392d14f6f301fe1929785d30ab0937e/htdocs/core/lib/company.lib.php#L861-L866

PR #34849, opened 2025-07-31
# FIX - Warning syntaxe in console

atm-lucasmantegari:
# FIX - Warning syntaxe in console
*This change in syntax avoids an error in the console: there was a problem with quotes and double quotes.*


htdocs/core/lib/ajax.lib.php
```diff
@@ -533,7 +532,7 @@ function ajax_combobox($htmlname, $events = array(), $minLengthToAutocomplete =
 						return data.text;
 					},
 					templateSelection: function (selection) {		/* Format visible output of selected value */
-						if (selection.id == '.((int) $idforemptyvalue).') return \'<span class="placeholder">\'+selection.text+\'</span>\';
+						if (selection.id == '.((int) $idforemptyvalue).') return "<span class=\"placeholder\">"+selection.text+"</span>";
```
eldy, line 535: There is no reason to have a change here. The char to use for js string should be ' and not " so old code seems good.
Can you describe a way to reproduce the bug ?

PR #34854, opened 2025-08-01
NEW : add company date birth (SQL structure)

atm-maxime:
# NEW : add company date birth (SQL structure)


eldy:
The field name for llx_user is "birrh" for both the sqlncolumn and thebproperry in php fike.
So may be we should use same name for llx_societe

atm-maxime:
@eldy ok i renamed the column. FYI :

image

Should we rename all fields into "birth" ? Or "birth_date" to be consistent with "birth_place" ?

eldy:
> @eldy ok i renamed the column. FYI :
> 
> image
> Should we rename all fields into "birth" ? Or "birth_date" to be consistent with "birth_place" ?

Yes, as we have a majority of couple keys:
birth (for dateà / birth_place (for town)
we can standardize to this 2 column name everywhere.

PR #34856, opened 2025-08-01
FIX : Allow - or _ as delimeter for pdfs when builddoc (MergePDF) is used

warnerbryce:
# FIX 
When i add files into an invoice for example, Dolibarr will rename the files as follows : 
REF of the invoice then a dash then the name of the file
Example : FA2507-0075-Annexe 036.pdf

Then if this invoice is selected on a massaction for Merging PDF i won't find this additional file because of two things : 

1. After the REF i get a dash, not an underscore
2. My additional file have a space in his name.

This PR correct that.

The better use case of this is on Suppliers Invoices. 
Most users don't use Dolibarr PDF models for this, they add as attachment the original PDF invoice from the supplier itself.
If the accountant wants all the suppliers invoice, you can now select them and merge them easily.


eldy:
This is the expexted behaviour: the mass action is designed to merge only fikes that we can trust so file generated by dolibarr, because we any other fukes we may experience error fir different reasons:
Other files are not under control. They might be protected by password so nit mergeable,, they may be in a different sizet format, they may be confidential documents that an external user should absolutely not see. If you want to provide all pdf of suppliers to a nook keeper, even if theyvare not thebmayn generated doculent, why not using the "export source document" feature of moduke accounting ?

warnerbryce:
@eldy alright i will dig this way.

PR #34865, opened 2025-08-03
NEW : update country list

atm-maxime:


sonikf:
Hi @atm-maxime  @eldy
According to this: https://taxation-customs.ec.europa.eu/taxation/vat/vat-directive/how-does-vat-work/territorial-scope_en
vat rules do not apply for Guadeloupe, French Guiana, Martinique, Réunion, Mayotte, Saint-Martin

eldy:
> Hi @atm-maxime @eldy According to this: https://taxation-customs.ec.europa.eu/taxation/vat/vat-directive/how-does-vat-work/territorial-scope_en vat rules do not apply for Guadeloupe, French Guiana, Martinique, Réunion, Mayotte, Saint-Martin

Yes, this is a special case. We have 2 way to manage this. On old version, special french department was managed by adding a country so we can manage the vat that is different for this region than the vat of the french "metropole".
Another way of management of special french state is to have them as state (the have their own zip code and except vat, the rest of management is similar to other states). This method was sometimes used, sometimes not because we were not able to have the correct vat whe using this method. However, from v22, we are now able to have vat rates that depends on the state.

The eec column is currently most used to define if we apply rule for intravat code, so i think we should not have the flag eec to 1 for the special french region as intravat rule does not apply so i made a fix so we can validate the rest of the PR for property sepa.

PR #34866, opened 2025-08-03
NEW : add extrafield option "empty on clone"

### LINKED ISSUE #34735: Extrafields : add option to clear an extrafield on clone
### Feature Request

On an extrafield definition, add a checkbox "Clear on clone" and when the object is created from a clone, empty the value of the extrafield.

### Use case

We might want to have empty extrafields values on an object clone, i.e. I have a "Date sent" on the invoice object. If I clone an invoice, I don't want its "Date sent" copied as well.

### Suggested implementation

_No response_

### Suggested steps

_No response_

atm-maxime:
CLOSE #34735

Also fixes the extrafields declared as unique that were not emptied on clone...

### DIFF
--- a/htdocs/core/class/extrafields.class.php
+++ b/htdocs/core/class/extrafields.class.php
@@ -270,13 +270,13 @@ class ExtraFields
 		// Create field into database except for separator type which is not stored in database
 		if ($type != 'separate') {
 			dol_syslog(get_class($this).'::thisupdate', LOG_DEBUG);
-			$result = $this->update($attrname, $label, $type, $size, $elementtype, $unique, $required, $pos, $param, $alwayseditable, $perms, $list, $help, $default_value, $computed, $entity, $langfile, $enabled, $totalizable, $printable, $moreparams, $emptyonclone);
+			$result = $this->update($attrname, $label, $type, $size, $elementtype, $unique, $required, $pos, $param, $alwayseditable, $perms, $list, $help, $default_value, $computed, $entity, $langfile, $enabled, $totalizable, $printable, $moreparams, '', $emptyonclone);
 		}
 		$err1 = $this->errno;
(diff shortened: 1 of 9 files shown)


htdocs/core/class/extrafields.class.php
```diff
@@ -268,13 +270,13 @@ public function updateExtraField($attrname, $label, $type, $pos, $size, $element
 		// Create field into database except for separator type which is not stored in database
 		if ($type != 'separate') {
 			dol_syslog(get_class($this).'::thisupdate', LOG_DEBUG);
-			$result = $this->update($attrname, $label, $type, $size, $elementtype, $unique, $required, $pos, $param, $alwayseditable, $perms, $list, $help, $default_value, $computed, $entity, $langfile, $enabled, $totalizable, $printable, $moreparams);
+			$result = $this->update($attrname, $label, $type, $size, $elementtype, $unique, $required, $pos, $param, $alwayseditable, $perms, $list, $help, $default_value, $computed, $entity, $langfile, $enabled, $totalizable, $printable, $moreparams, $emptyonclone);
```
eldy, line 273: The parameter emptyonclone is at position of the other parameter aiprompt.

atm-maxime: Ok I fixed it. Don't know why "aiprompt" is here but ok.

JonBendtsen:
@atm-maxime  good idea, thank you for doing it. I only have a tiny wish, and that is that you would also have added this to the API functions for extrafields. (if you have then it probably is not in v23.0.0)

atm-maxime:
> @atm-maxime good idea, thank you for doing it. I only have a tiny wish, and that is that you would also have added this to the API functions for extrafields. (if you have then it probably is not in v23.0.0)

Hi ! Thanks, but you're right, I did'nt apply this new feature in the extrafields API... I keep that in mind and do it as soon as can be !

PR #34878, opened 2025-08-05
NEW : Change the path for the mailing files

kkhelifa-opendsi:
Change the path for the mailing files because if you have more than 100 mailing you have the files of the previous mailing show in the newest.

Only activate if the global MAILING_USE_NEW_PATH_FOR_FILES is set to 1.
Add a script for move files to new path.


scripts/emailings/migrate_files_path.php
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env php
```
eldy, line 1: A lot of user does not have access to the command line (most mutualized hosting and cloud services)
This is why such features are centralized into the page 
/install/repair.php

Can you move the call of the function into the repair.php page ?

thomas-Ngr:
@eldy I made the changes you requested.

PR #34883, opened 2025-08-06
NEW(propal): Add messaging and agenda features to proposals

Pichinov-Valentin:
**### NEW|New Extend messaging & agenda features to Proposals**

This PR extends the messaging functionality to propasals.

Specifically, this introduces:

A 'Messaging' tab on the Commercial Proposal card.
And modifications to the 'Agenda' tab on the Commercial Proposal card.

The goal is to centralize communication and improve event tracking for these documents, just as it was implemented for customer orders and shipments. This also includes minor copyright header updates in the modified files.

**How to test manually:**

For Commercial Proposals:

Navigate to any Commercial Proposal card.
Verify that a new tabs, 'Messaging' is present and you can access it from agenda.
Click on each tab and check that their respective interfaces display correctly.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -8509,8 +8509,8 @@ function get_default_tva(Societe $thirdparty_seller, Societe $thirdparty_buyer, 
 
 	// If the (seller country = buyer country) then the default VAT = VAT of the product sold. End of rule.
 	if (($seller_country_code == $buyer_country_code)
-	|| (in_array($seller_country_code, array('FR', 'MC')) && in_array($buyer_country_code, array('FR', 'MC')))
-	|| (in_array($seller_country_code, array('MQ', 'GP')) && in_array($buyer_country_code, array('MQ', 'GP')))	// We should be able to manage the case of MQ, GP, ... with a deicated vat rate at previous step.
+		|| (in_array($seller_country_code, array('FR', 'MC')) && in_array($buyer_country_code, array('FR', 'MC')))
+		|| (in_array($seller_country_code, array('MQ', 'GP')) && in_array($buyer_country_code, array('MQ', 'GP')))	// We should be able to manage the case of MQ, GP, ... with a deicated vat rate at previous step.
 	) { // Warning ->country_code not always defined
 		//print 'VATRULE 3';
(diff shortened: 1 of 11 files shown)


htdocs/core/lib/functions.lib.php
```diff
@@ -15764,6 +15766,8 @@ function show_actions_messaging($conf, $langs, $db, $filterobj, $objcon = null,
 			$sql .= " LEFT JOIN " . MAIN_DB_PREFIX . "commande as o ON a.fk_element = o.rowid AND a.elementtype = 'order'";
 		} elseif (is_object($filterobj) && get_class($filterobj) == 'Expedition') {
 			$sql .= " LEFT JOIN " . MAIN_DB_PREFIX . "expedition as s ON a.fk_element = s.rowid AND a.elementtype = 'shipping'"; // CHANGED: Table 'expedition' et alias 's', elementtype 'shipping'
+		} elseif (is_object($filterobj) && get_class($filterobj) == 'Propal') {
+			$sql .= " LEFT JOIN " . MAIN_DB_PREFIX . "propal as o ON a.fk_element = o.rowid AND a.elementtype = 'propal'";
```
eldy, line 15770: You should use an exact join because we want the actioncomm that are linked to the propal. If not we don't want them, so a left join is not necessary here, a
" INNER JOIN " 
should give the same result with better performances.

Pichinov-Valentin:
Hi @eldy, could i use this syntax instead :

} elseif (is_object($filterobj) && get_class($filterobj) == 'Propal') {
			$sql .= ", " . MAIN_DB_PREFIX . "propal as o";
}


this is the old way of doing inner join, and to follow the actual syntax of the other elements.

eldy:
> Hi @eldy, could i use this syntax instead :
> 
> } elseif (is_object($filterobj) && get_class($filterobj) == 'Propal') {
> 			$sql .= ", " . MAIN_DB_PREFIX . "propal as o";
> }
> 
> 
> this is the old way of doing inner join, and to follow the actual syntax of the other elements.

If you prefer, why not. The important is to avoid left join when we can.

Pichinov-Valentin:
I already did it in the new commit with all the others fixes.

PR #34897, opened 2025-08-08
FIX #34893

### LINKED ISSUE #34893: Change password without rights
### Bug

If a user don't have rights to change his password, he can bypass this right with the "password lost" feature.

### Dolibarr Version

all versions

### Environment PHP

any

### Environment Database

_No response_

just logout and ask for a new password

### Attached files

_No response_

hregis:
@eldy same bug in all versions

### DIFF
--- a/htdocs/user/passwordforgotten.php
+++ b/htdocs/user/passwordforgotten.php
@@ -136,7 +136,7 @@ if (empty($reshook)) {
 				$edituser->getrights('user');
 				if (!$edituser->hasRight('user', 'self', 'password')) {
 					$result = 0;
-					$edituser->error = 'USERNOTFOUND';
+					$edituser->error = 'USERNOTALLOWEDTOCHANGEPASS';
 				}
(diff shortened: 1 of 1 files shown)


htdocs/user/passwordforgotten.php
```diff
@@ -131,6 +131,15 @@
 				$result = $edituser->fetch('', '', '', 1, $conf->entity, $username);
 			}
 
+			// If the user does not have the right to change his own password
+			if ($result > 0) {
+				$edituser->getrights('user');
+				if (!$edituser->hasRight('user', 'self', 'password')) {
+					$result = 0;
+					$edituser->error = 'USERNOTFOUND';
```
eldy, line 139: May be we can use a different error code like 'USERNOTALLOWEDTOCHANGEPASS'  ?

hregis: @eldy ready

PR #34908, opened 2025-08-10
New : Separate permissions for Submit and delete documents in GED module

bafbes:
New : Separate permissions for Submit and delete documents in GED module


bafbes:
Bonjour,
Je constate que l’erreur de traduction détectée concerne un fichier non modifié dans cette PR.
Il me semble que la correction doit être effectuée ailleurs ?

eldy:
Imagine a user that has permission "upload" and not the permission "delete".
He can still delete the file by uploading a file with same name and a dummy content.
So do we really need this permission if we can rely on ? What is the use case that need this ?

bafbes:
We should then not allow uploading of existing files if delete permission is not granted. What do you think about it ?

eldy:
> We should then not allow uploading of existing files if delete permission is not granted. What do you think about it ?

So if you want to also not allow the upload whenever the user has not the permission to delete, it is the same that giving it only the permission to read.
So it seems we only need 1 permission for both upload and delete, and this is what we have currently.

bafbes:
No. A user with upload permission but without delete permission should only be allowed to upload new files, not overwrite existing ones. He must also have delete permission to be allowed to replace files by uploading. This way, the distinction between these two privileges would be truly meaningful.

eldy:
> No. A user with upload permission but without delete permission should only be allowed to upload new files, not overwrite existing ones. He must also have delete permission to be allowed to replace files by uploading. This way, the distinction between these two privileges would be truly meaningful.

Ok so to summarize, currently we have permission
"upload":   Allow to upload and overwrite and delete files in ECM. It could be named "Manage document upload/remove"

What you want is 2 different permissions:
"upload": Allow to upload only but not overwrite and not delete
"delete": Allow to overwite in upload and to delete

Am i right ? If yes, it is ok. Can you just add a test to block the overwrite when user has no permission to delete ? Because if we offer a "can't delete" feature, we should not be able to delete if by overwritting with an empty file.

PR #34912, opened 2025-08-11
Custom login template from custom module

hermans:
This commit will set custom login tamplate base on custom tpl, custom login tpl as first priority.


hregis:
@hermans @eldy this code break the search of other path in "$conf->modules_parts['tpl']" !
You must use an external module name and not directly "htdocs/custom/core/tpl/login.tpl.php"

hermans:
yes, your code is work, but i have custom custom login custom_login.php that will open database after logged. but when login first loaded it just connect to default db. do you have suggestion for this case. ?

hregis:
@hermans you can create a "fake" module:

add a directory "/htdocs/custom/myfakemodule/core/tpl/" with your file "login.tpl.php"
add a constant "MAIN_MODULE_MYFAKEMODULE_TPL" with value 1

hregis:
@hermans maybe add another constant "MAIN_MODULE_MYFAKEMODULE" with value 1

replace "myfakemodule" with the name you would

PR #34923, opened 2025-08-11
can set extrafields in project elements

frederic34:


htdocs/core/class/extrafields.class.php
```diff
@@ -2681,7 +2681,7 @@ public function showSeparator($key, $object, $colspan = 2, $display_type = 'card
 	 * @param	int			$todefaultifmissing 1=Set value to the default value in database if value is mandatory and missing
 	 * @return	int								1 if array_options set, 0 if no value, -1 if error (field required missing for example)
 	 */
-	public function setOptionalsFromPost($extralabels, &$object, $onlykey = '', $todefaultifmissing = 0)
+	public function setOptionalsFromPost($extralabels, $object, $onlykey = '', $todefaultifmissing = 0)
```
eldy, line 2684: Line 2827, the code change a value of $object->array_options to have $object modified after function.
Does this still works if you remove the & here ?

frederic34: yes it still works
image
changing value from aaa to bbb
image

PR #34924, opened 2025-08-11
fix #34920 remove hidden const

### LINKED ISSUE #34920: V22 Missing tags/Categories menu entry in products, customers, suppliers, contacts, and members
### Bug

After update from v21 to v22 there are missing the tags/Categories menu entries in in products, customers, suppliers, contacts, and members.
Only is present  the tags/Categories menu in Tickets section.

### Dolibarr Version

v22

### Environment PHP

8.1.33

### Environment Database

MySQL or MariaDB 10.6.23-MariaDB   mysqli (mysqlnd 8.1.33)

1. Just click on the top menu "Third parties" and observe that the menu item Tags/categories is missing in Customers, vendors and contacts.  
2. Just click  on the top menu "products/services" and observe that the menu item Tags/categories is missing in products and services

### Attached files

_No response_

sonikf:
Users expect menu to create modify tags/Categories as was the case up to v. 21. Use of constant is counter intuitive. User is now forced to open a customer card, edit it then press + icon...

### DIFF
--- a/htdocs/core/menus/standard/eldy.lib.php
+++ b/htdocs/core/menus/standard/eldy.lib.php
@@ -1264,7 +1264,7 @@ function get_left_menu_home($mainmenu, &$newmenu, $usemenuhider = 1, $leftmenu =
 				$newmenu->add("/user/card.php?leftmenu=users&action=create", $langs->trans("NewUser"), 2, (int) (($user->hasRight("user", "user", "write") || $user->admin) && !(isModEnabled('multicompany') && !empty($user->entity) && getDolGlobalString('MULTICOMPANY_TRANSVERSE_MODE'))), '', 'home');
 				$newmenu->add("/user/list.php?leftmenu=users", $langs->trans("ListOfUsers"), 2, (int) ($user->hasRight('user', 'user', 'read') || $user->admin));
 				$newmenu->add("/user/hierarchy.php?leftmenu=users", $langs->trans("HierarchicView"), 2, (int) ($user->hasRight('user', 'user', 'read') || $user->admin));
-				if (isModEnabled('category') && getDolGlobalString('CATEGORY_EDIT_IN_MENU_NOT_IN_POPUP')) {
+				if (isModEnabled('category')) {
 					$langs->load("categories");
 					$newmenu->add("/categories/categorie_list.php?leftmenu=users&type=7", $langs->trans("UsersCategoriesShort"), 2, $user->hasRight('categorie', 'read'), '', $mainmenu, 'cat');
 				}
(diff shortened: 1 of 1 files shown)


eldy:
No, using the open customer card, edit it... is the process to add a category "on the fly" when creating a thirdparty.
If you want to create some categories before creating your thirdparties, you can find all the entry menus to add a new cateogry into menu Home - Tools - Categories. Feature to manage categories was just centralized. 
Adding a category on the fly is an addition, not a replacement.

bos4711:
@eldy While I appreciate the "+" on category alteration directly on cards, removing all category management from the submenus to the Tools menu broke the work process we have been using since V12. Thankfully adding CATEGORY_EDIT_IN_MENU_NOT_IN_POPUP=1 restored it, but I would have preferred this option (or atleast some information) would be inside the settings of the Tags module.

eldy:
> @eldy While I appreciate the "+" on category alteration directly on cards, removing all category management from the submenus to the Tools menu broke the work process we have been using since V12. Thankfully adding CATEGORY_EDIT_IN_MENU_NOT_IN_POPUP=1 restored it, but I would have preferred this option (or atleast some information) would be inside the settings of the Tags module.

This a mostly a trouble due to an habit.
For a new user, the new way is more convenient, and we should avoid to suggest option to use old habit to avoid both the optionflation in default mode (increase of number of option so much that user is lost in setup) and too different way of using the interface.

bos4711:
@eldy Is there a good reason the search box (for tags) was removed? We used that frequently since we have over 300 tags, many which are nested. Now when the search box is gone it's very difficult to use.

eldy:
> @eldy Is there a good reason the search box (for tags) was removed? We used that frequently since we have over 300 tags, many which are nested. Now when the search box is gone it's very difficult to use. 

Sorry, I don't see what you are talking about. 
All categories are listed in list of categories, with a filter available.
May be you can you provide a screenshot ?

bos4711:
> > @eldy Is there a good reason the search box (for tags) was removed? We used that frequently since we have over 300 tags, many which are nested. Now when the search box is gone it's very difficult to use.
> 
> Sorry, I don't see what you are talking about. All categories are listed in list of categories, with a filter available. May be you can you provide a screenshot ?

This is from our non-production Dolibarr v16. We have used the tag search since V12 and it's very important since we are heavy reliant on the tag system.

foo

bos4711:
In v22 i don't find a search box for tags.

v22

hregis:
@bos4711 click on "view list"

Capture d’écran 2025-11-01 à 09 47 19

bos4711:
> @bos4711 click on "view list"
> 
> Capture d’écran 2025-11-01 à 09 47 19
> 

I did that, but searching for reference brings up results split in different pages instead of showing everything on the same screen like it has been doing for 8 years.

I wish i had known this before upgrading from v20 to v22. This change in the tag system completely broke our workflow 😔

eldy:
> In v22 i don't find a search box for tags.
> 
> v22
> 

I see. The code to manage categories was rewritten to match the rest of application using same template. In current template, the view by hierarchy (tree) is not able to make filter. If we search categ b but b is child of a and has a child c, what do we have to show? A, B,C ? We can t just show b because this is a tree view. This is why filter is only in the flat list view.

PR #34925, opened 2025-08-11
FIx virtual stock when stock decrease on invoice validate.

fappels:
If product in invoice but not in linked order count should not be used.

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -3659,7 +3659,7 @@ class Product extends CommonObject
 			}
 
 			// If stock decrease is on invoice validation, the theoretical stock continue to
-			// count the orders lines in theoretical stock when some are already removed by invoice validation.
+			// count the orders lines containing product in theoretical stock when some are already removed by invoice validation.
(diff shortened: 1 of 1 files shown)


htdocs/product/class/product.class.php
```diff
@@ -3668,6 +3668,7 @@ public function load_stats_commande($socid = 0, $filtrestatut = '', $forVirtualS
 					$sql .= " JOIN ".$this->db->prefix()."facture as f ON fd.fk_facture = f.rowid";
 					$sql .= " JOIN ".$this->db->prefix()."element_element as el ON ((el.fk_target = f.rowid AND el.targettype = 'facture' AND sourcetype = 'commande') OR (el.fk_source = f.rowid AND el.targettype = 'commande' AND sourcetype = 'facture'))";
 					$sql .= " JOIN ".$this->db->prefix()."commande as c ON el.fk_source = c.rowid";
+					$sql .= " JOIN ".$this->db->prefix()."commandedet as cd ON cd.fk_product = fd.fk_product AND cd.fk_commande = c.rowid";
```
eldy, line 3671: Why adding a link on this table "commandedet as cd", it seems this table is never used.

PR #34930, opened 2025-08-12
Update bankentries_list.php

homer8173:
Add invoice list linked to bank entries
Main feature required by clients


eldy:
It seems you have a problem into the seyup of your IDE. It does not remove automatically the ending whitespace.
It also doe not replace automatically 4 spaces with tab.
You should find such option on your IDE.

homer8173:
Thanks for your tip, I'll take care of this in future.

eldy:

PR #34933, opened 2025-08-12
FIX - Removal of accents in searches via comboxbox

atm-lucasmantegari:
# FIX - Removal of accents in searches via comboxbox
When searching in a combo box, it would be useful not to differentiate between words with and without accents for the sake of speed (for example, if I have a user named “José,” I would like to be able to find them by simply typing “jose”).


eldy:
As you said, "it would be useful", but "useful" is not a bug so should not be pushed into a maintenance version.
Thinks also to introduce a hidden option to support this mode, to avoid to break the current default behaviour that may be prefer (some users may prefer to not have the sentence with "sur" key found when a seach is done on "sûr" for example, and some language may also have word with complete different meaning with/without accent). So having an option to swith in the variant search mode is a better choice.

PR #34936, opened 2025-08-12
FIX / when we close shipping we change delivery date, not shipping date

defrance:
date_expedition is the date when we expediate the shipping date_delivery is the date when the customer receive the shipping

closing the shipping corresponding of the delivery, not the expedition


defrance:
@eldy , your PR don't change date_expodition, to date_delivery

eldy:
> @eldy , your PR don't change date_expodition, to date_delivery

Yes the date delivery is the date of reception. The date of shipment is date_expedition. When we close the shipment, the shipment is done so date_expedition must be filled.
date_delivery is filled when using the feature of delivery receipt. When feature is off (most common case), this information is never shown, so not filled
We have no reason to fill this field. It is the date planned of delivery of shipment. It can remain unknow if itnwas never planned.
The date_delivery in shipment is the planed value, the date_delivery in Delivery is the real date of reception (only available if reception management is on)

PR #34942, opened 2025-08-13
Update facture.class.php

Codered741:
Fix sendEmailsRemindersOnInvoiceDueDate to process all invoices independently, ensuring accurate email send count by continuing on template or fetch errors instead of early return.


htdocs/compta/facture/class/facture.class.php
```diff
@@ -6238,8 +6238,10 @@ public function sendEmailsRemindersOnInvoiceDueDate($nbdays = 0, $paymentmode =
 					$arraymessage = $formmail->getEMailTemplate($this->db, 'facture_send', $user, $outputlangs, (is_numeric($template) ? $template : 0), 1, (is_numeric($template) ? '' : $template));
 					if (is_numeric($arraymessage) && $arraymessage <= 0) {
 						$langs->load("errors");
-						$this->output .= $langs->trans('ErrorFailedToFindEmailTemplate', $template);
-						return 0;
+						$errormesg = $langs->trans('ErrorFailedToFindEmailTemplate', $template) . ' for invoice id=' . $tmpinvoice->id;
```
eldy, line 6241: What is trange to me is that this error does  not depend on the email to send.
So if you reach it, you should get it for every invoice. This is why we stop immediatly because we should always fails. In which use cas do you have an error in this loop at pass n and not at pass m ?

Codered741: The root issue I was solving was the results of calling the function in scheduled jobs would return "0" emails sent, when one or more emails were in fact sent.

eldy: But how can you have this ?
If it fails at first pass when trying to get template, no email scan be have sent (there is a return 0).
So you should not have no mail sent a tall ?
What was the output ?

PR #34944, opened 2025-08-13
Qual: Update phan baseline

mdeweerd:
# Qual: Update phan baseline

Remove solved exceptions from the phan baseline.


mdeweerd:
@eldy Note that phan is now at version 5.5.1 - I suppose that the server generating cti.dolibarr.org is not automatically kept up to date.

eldy:
Note: I added the version used of phan in apstats report.

PR #34946, opened 2025-08-14
fix: with MAIN_MAIL_ADD_INLINE_IMAGES_IF_IN_MEDIAS and multicompany image are not find on send mail

FHenry:
Following #34894, missing catch to (int) for $conf->entity. Seems to not always be an int (but can be string)....


FHenry:
@eldy Sorry I miss to set correctly the target branch, I reset all to 19.0

PR #34950, opened 2025-08-14
add doc

frederic34:


htdocs/core/tpl/notes.tpl.php
```diff
@@ -47,6 +47,8 @@
 $note_private = 'note_private';
 
 if ($module == "product") {
+	'@phan-var-force Product $object';
+	/** @var Product $object */
```
eldy, line 51: The line @var id not enough for phan ?

frederic34: I'm not sure so I put the 2 versions

eldy: > I'm not sure so I put the 2 versions

It could be great to know, so we can abandon specific syntax for phan.
@mdeweerd do you know the answer ?

mdeweerd: '@phan-var-force Product $object'; is needed for phan because of limitation in the AST parser - according to https://github.com/phan/phan/issues/313 the '@var' directives are only available for certain nodes.

A common approach is to use 'instanceof' where possible.  So something like 'if ($object instanceof Product)' .

PR #34970, opened 2025-08-15
NEW: Add prepare() method to DoliDB class to support prepared SQL statements

R0ck3n:
# NEW: Add prepare() method to DoliDB class

This PR adds support for prepared SQL statements in Dolibarr by implementing a `prepare()` method in the `DoliDB` abstract class.

## Purpose
Prepared statements help mitigate SQL injection risks and can improve performance when executing similar queries multiple times.

## Implementation
- Added a `prepare($sql)` method in `DoliDB`, which delegates to the native `prepare()` method of the underlying DB handler (e.g., MySQLi).
- Handles error assignment (`$this->lasterror`) if the statement fails.

## Usage example
```php
global $db;

$sql = "SELECT rowid, login, lastname, firstname FROM " . MAIN_DB_PREFIX . "user WHERE rowid = ?";

// Préparation
$stmt = $db->prepare($sql);

if ($stmt === false) {
    echo "Erreur de préparation : " . $db->lasterror;
    exit;
}

// Lier le paramètre (ici 'i' pour int, et on passe 1 pour rowid)
$rowid = 1;
$stmt->bind_param("i", $rowid);

// Exécution
if (!$stmt->execute()) {
    echo "Erreur à l'exécution : " . $stmt->error;
    exit;
}

// Récupération du résultat
$result = $stmt->get_result();

if ($result && $result->num_rows > 0) {
    $user = $result->fetch_assoc(); // ou fetch_object() si tu préfères un objet
    var_dump($user);
} else {
    echo "Aucun utilisateur trouvé avec rowid = 1.";
}

// Nettoyage
$stmt->close();
```


htdocs/core/db/DoliDB.class.php
```diff
@@ -459,4 +459,25 @@ public function getRows($sql)
...
+	 * @return false|mysqli_stmt
+	 */
+	public function prepare($sql)
+	{
+		if (!$this->connected) {
+			$this->lasterror = 'Not connected to database';
+			return false;
+		}
+
+		$stmt = $this->db->prepare($sql);
```
eldy, line 476: I think the implementation with
$stmt = $this->db->prepare($sql);

should not be into the DoliDB.class.php but into the DoliDBMysqli class. Because method ->prepare does not exists in postgresql.
So we can add the 
public function prepare($sql)
with empty content into DoliDB.class.php
and the complete implementation into class  DoliDBMysqli  and DoliDBPgsql

hregis:
@R0ck3n rebuild your PR with a clean develop branch please, thanks.

R0ck3n:
@hregis thx really better.
done here : https://github.com/Dolibarr/dolibarr/pull/35249

PR #34975, opened 2025-08-16
Bug Fix: Livraison date not store when call funciton Livraison on htdocs/fourn/class/fournisseur.commande.class.php

hermans:
Currenty we have: public function Livraison($user, $date, $type, $comment) on "htdocs/fourn/class/fournisseur.commande.class.php"

https://github.com/Dolibarr/dolibarr/blob/1fb37db6f6005cd7c76abefcd7560eb05c671ca2/htdocs/fourn/class/fournisseur.commande.class.php#L2636

but $date is unused so we don't know when the good is full received, so this patch to fix it so we can have date livration data.
this will update field: llx_commande_fournisseur->date_livraison

### DIFF
--- a/htdocs/fourn/class/fournisseur.commande.class.php
+++ b/htdocs/fourn/class/fournisseur.commande.class.php
@@ -2697,6 +2697,7 @@ class CommandeFournisseur extends CommonOrder
 
 				$sql = "UPDATE ".$this->db->prefix()."commande_fournisseur";
 				$sql .= " SET fk_statut = ".((int) $statut);
+				$sql .= ",date_livraison = ".($date ? "'".$this->db->idate($date)."'" : 'null');
 				$sql .= " WHERE rowid = ".((int) $this->id);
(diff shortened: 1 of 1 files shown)


htdocs/fourn/class/fournisseur.commande.class.php
```diff
@@ -2697,6 +2697,7 @@ public function Livraison($user, $date, $type, $comment)
 
 				$sql = "UPDATE ".$this->db->prefix()."commande_fournisseur";
 				$sql .= " SET fk_statut = ".((int) $statut);
+				$sql .= ",date_livraison = ".($date ? "'".$this->db->idate($date)."'" : 'null');
```
eldy, line 2700: The field date_livraison is the "date planed for delivery".
If i t was alreasy set, we should not erase it to not loose information.
The date of real reception is not stored for the moment (except into the list of agenda events) or if you use the module reception.
If not using this module, we could store it into a field "date_reception" in llx_commande_fournisseur but another PR to fist add the field, and then update/manage it should be done differently in 2 different PR.

PR #34978, opened 2025-08-16
NEW: add pdf_ttc_sponge invoice model (TTC-only)

Th9boss:
# NEW| Add pdf_ttc_sponge invoice model (TTC-only)

This PR introduces a new invoice PDF model `pdf_ttc_sponge` based on the existing Sponge template.

**Features:**
- Shows only amounts including tax (TTC).
- Hides HT and VAT columns/totals.
- Keeps layout similar to Sponge for consistency.
- Multilingual description added (EN, FR).

**Use case:**
For companies that exclusively manage prices including tax, this model provides a simpler and clearer invoice layout, avoiding confusion with HT/TVA columns.

**Tests done:**
- Invoices with multiple VAT rates.
- Invoices with 0% VAT.
- Invoices with discounts.
- Totals display only the TTC line.

_No impact on existing models (Sponge remains unchanged)._
[SPECIMEN-2.pdf](https://github.com/user-attachments/files/21812023/SPECIMEN-2.pdf)
SCR-20250816-kxym

### DIFF
--- a/htdocs/core/modules/facture/doc/pdf_ttc_sponge.modules.php
+++ b/htdocs/core/modules/facture/doc/pdf_ttc_sponge.modules.php
@@ -111,8 +111,8 @@ class pdf_ttc_sponge extends ModelePDFFactures
 	/**
  * @var array<string,array{
  *   rank:int,
- *   width:float|int|false,
- *   status?:bool|int<0,1>,
+ *   width:float|false,
+ *   status:bool|int<0,1>,
  *   'border-left'?:bool,
  *   title:array{
  *     textkey:string,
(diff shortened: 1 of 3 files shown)


eldy:
We could not add a new template for each variant of information we want to show. Instead we must implement the variant/option into the existing model.
For exemple, in sponge template, we already have a lot of option to decide wich field to show.
See PDF_INVOICE_HIDE_PRICE_EXCL_TAX or PDF_INVOICE_SHOW_PRICE_INCL_TAX (that work for both sale and puchrase invoices)

I have just pushed a commit to add option like PDF_PURCHASE_INVOICE_HIDE_VAT, PDF_PURCHASE_INVOICE_SHOW_PRICE_INCL_TAX and PDF_PURCHASE_INVOICE_HIDE_PRICE_EXCL_TAX to hide this column on puchase invoice instead of all documents to show you the example on how to do.

PR #34986, opened 2025-08-18
Fix bug unit prefix on Unit field in product

hermans:
When we enable PRODUCT_USE_UNITS global setting, we allow to select unit on the product. but the problem when display data always have prefix 'unit', this patch will fix it.

image

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -6762,7 +6762,7 @@ class Product extends CommonObject
 			dol_syslog(get_class($this)."::getLabelOfUnit Error ".$this->error, LOG_ERR);
 			return -1;
 		} elseif ($this->db->num_rows($resql) > 0 && $res = $this->db->fetch_array($resql)) {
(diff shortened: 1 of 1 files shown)


eldy:
I can't reproduce the bug on last v21.*
Which language code do you use (en_US, ... ?)

hermans:
en_US

hermans:
In my testing it happend on 21 and 22.

image

eldy:
Ok i see. Pb is when you use a custom unit with nobtranslation key.
I pushed a more complzte fix.

hermans:
@eldy 
product/price_suppliers.php broken after update:
image

image

$result no result

eldy:
This does not seems to be related to the fix. This part of code testing a not define var is in code since several years.
On my side var is always 1 because defined into main.inc.php, by chance. This is why it was working. You surely have another context that make it empty.
I see no reason to have this test. Is it ok if you remove completely the if $result ?

hermans:
yes, $result is not needed

PR #34994, opened 2025-08-19
Fix facture header title on dolibar-22

hermans:
Similar case with this: https://github.com/Dolibarr/dolibarr/pull/34993


eldy:
@hermans 
Because the PR was already submited in v21, it is useless to submit it to v22. It will be merged automatically every week.

PR #35010, opened 2025-08-19
FIX #35007

### LINKED ISSUE #35007: The hook addMoreActionsButtons in TakePOS looks unusable to add more action buttons
### Bug

Hello,

I stumbled upon a strange thing while reading the TakePOS file `pay.php`.

I am looking forward to implement a custom payment button to handle specific system for a customer, but the hook inside `pay.php`, at line 751 does not seems to pass the correct variable : 

```php
$parameters = array();
$reshook = $hookmanager->executeHooks('addMoreActionsButtons', $parameters, $invoice, $action); // Note that $action and $object may have been modified by hook
if ($reshook < 0) {
}
```

The hook is receiving the variable `$action` which looks not initialized in this file, then the following code is using the variable `$action_buttons` to add the buttons : 
```php
foreach ($action_buttons as $button) {
	$newclass = $class.($button["class"] ? " ".$button["class"] : "");
	print '<button type="button" class="'.$newclass.'" onclick="'.$button["function"].'"><span '.$button["span"].'>'.$button["text"].'</span></button>';
}
```

So I guess the hook should receive `$action_buttons` and not `$action` to actually add a button ?

Anyway, if I understand the naming conventions correctly, `$action` should contain the GETPOST value of $action (which is not the case in `pay.php`), this variable seems off and in any case, I would be surprised if this hook can actually add more buttons ?

Or maybe I am reading all of this incorrectly and there is a correct way to add buttons without changing this code ?

Thanks in advance for your insights on this issue.

mpellegrin:
Proposal to fix #35007


mpellegrin:
Hello,

Is there a mistake in the merge direction or should something be fixed before merging this PR ?

PR #35021, opened 2025-08-20
Update some ID Language translation

hermans:
Update ID Lang translations


eldy:
PR on files into langs/xx_XX where xx_XX is not en_US is useless because files are erased by Transifex translation of en_US. Change on translation files must be done with PR for en_US only. For other languages (en_NZ, es_ES, ...), fix of translation must be done directly on 
https://app.transifex.com/dolibarr-association/dolibarr/

eldy:
> PR on files into langs/xx_XX where xx_XX is not en_US is useless because files are erased by Transifex translation of en_US. Change on translation files must be done with PR for en_US only. For other languages (en_NZ, es_ES, ...), fix of translation must be done directly on 
> https://app.transifex.com/dolibarr-association/dolibarr/
> 

You can also ping me here if you need your transifex account to be validated to modify files for your language. I can grant you lead permission on language.

PR #35030, opened 2025-08-21
add CATEGORY_SHOW_MENU and CATEGORY_HIDE_EDIT_IN_POPUP on setting

defrance:
Propose 2 settings for new categorie feature instead one and allow to setup it on category admin page


eldy:
Currently there is one option that allow the state: show categorie menu under each object (old place) or into the menu Tools + popup (new place).
My understanding is that you want to manage a third state to have them both in menu Tools + popup (new place) and new place ? 
Did i understood correctly ?

defrance:
My idea is to have one setting to control the position of the category menus, and another setting to enable the pop-ups.  These are two separate functions, and I think it's unfortunate that we're forced to choose between having the category menus or the pop-ups.
This allows for 4 possible choices (not two or three)

PR #35033, opened 2025-08-21
UIUX : Auto select member's nature depending on member's type

Thatoo:
# UIUX|Uiux [*Auto select member's nature*]
[*Auto select member's nature depending on chosen member's type in new member form and auto-suscription form*]

I tried my best to adapt to dolibarr's code but I think my PR needs review and improuvent.
I hope you'll like the idea though.

### DIFF
--- a/htdocs/adherents/class/adherent_type.class.php
+++ b/htdocs/adherents/class/adherent_type.class.php
@@ -710,6 +710,81 @@ class AdherentType extends CommonObject
 	}
 
 	/**
+	 *  Return the array of morphy per membership type id
+	 *
+	 *  @param	int		$status			Filter on status of type
+	 *  @return array<int,string>		Array of membership type
+	 */
+	public function morphyByType($status = null)
+	{
+		$morphybytype = array();
+
+		$sql = "SELECT rowid, morphy";
+		$sql .= " FROM ".MAIN_DB_PREFIX."adherent_type";
+		$sql .= " WHERE entity IN (".getEntity('member_type').")";
+		if ($status !== null) {
+			$sql .= " AND statut = ".((int) $status);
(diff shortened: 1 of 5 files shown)


htdocs/adherents/class/adherent_type.class.php
```diff
@@ -662,51 +665,22 @@ public function liste_array($status = -1)
...
-					$amountbytype[$obj->rowid] = $obj->amount;
-					$i++;
-				}
-			}
-		} else {
-			print $this->db->error();
-		}
-
-		return $amountbytype;
+		return array(
```
eldy, line 678: This is a first suggestion you must apply to no break compatibility (the method may be used by external modules).
You add aparameter $mode = 0 as a second parameter of liste_array
When value is 0, the default, the return is the same than currently. No compatibility break or risk.
If value is 1, you return an array.
The array returned should be (the id first and the the property label, ...)
`array(id1 => array('label' => ..., 'morhpy' => ... ), id2 => array('label' => ..., 'morhpy' => ... )`
instead of 
`array('label' => array(id1 => value, id2 => value2), 'morhpy' => array(id1 => valuem, id2 => valuem2)`
This match better dolibarr current dev pratices.

Thatoo: Then, I guess I should not delete `public function amountByType ()` neither and instead of modifying 	`public function liste_array($status = -1)` I should create two more functions, one for `morphy` and one for `caneditamount`.
That way, we'll have 4 function to do the same than one but it will assure ther isn't any compatibility break.

What do you think @eldy ?

Thatoo:
@eldy what do you think of my PR now?

PR #35041, opened 2025-08-22
FIX set global mysoc and load langs in API access

lvessiller-opendsi:
FIX set global mysoc and load langs in API access

### DIFF
--- a/htdocs/api/class/api_access.class.php
+++ b/htdocs/api/class/api_access.class.php
@@ -81,7 +81,7 @@ class DolibarrApiAccess implements iAuthenticate
 	public function __isAllowed()
 	{
 		// phpcs:enable
-		global $conf, $db, $user;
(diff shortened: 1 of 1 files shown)


htdocs/api/class/api_access.class.php
```diff
@@ -194,6 +194,56 @@ public function __isAllowed()
 			// Set also the global variable $user to the $user of API
 			$user = $fuser;
 
+			// set global mysoc after setting conf entity (the entity can be changed with the user logged)
+			// see master.inc.php
+			require_once DOL_DOCUMENT_ROOT.'/societe/class/societe.class.php';
+
+			$fmysoc = new Societe($db);
+			$fmysoc->setMysoc($conf);
```
eldy, line 202: Resetting $mysoc with setMysoc should be done only if we have switched into another entity.
So i think code added should be only just after line 134 and should not be called in any other cases.

PR #35071, opened 2025-08-25
update parent table tms when updating extrafields

frederic34:
For example, when updating a product's extrafield, it doesn't appear updated.


htdocs/core/class/commonobject.class.php
```diff
@@ -7036,6 +7036,10 @@ public function insertExtraFields($trigger = '', $userused = null)
 		}
 
 		if (!$error && $trigger) {
+			if (!empty($this->table_element) && !empty($this->fields['rowid']) && !empty($this->fields['tms'])) {
```
eldy, line 7039: 2 points:
Update of tms is done automatically by database. However, there is 2 problems with that:
- it does not manage correctly the timezone of use
- it needs special macro with postgresql.
So we have started to move the autoupdate into explicit update by SQL
Such explicit update, like you do must be inside condition
if (getDolGlobalString('MAIN_DISABLE_AUTO_UPDATE_OF_TMS_FIELDS')
so we can manage the transition

Second point:
We should not do a second SQL update. Update of tms field should be done at same time than updating other fields.

frederic34:
https://github.com/Dolibarr/dolibarr/pull/35111

PR #35078, opened 2025-08-25
Prepare SEPA payments for supplier invoices via Stripe

Daoud-mohamed:


eldy:
To redo by moving the if supplier/customer inside the stripe cynematic

PR #35081, opened 2025-08-25
FIX #34857 Structured communication number checksum is now always correct

### LINKED ISSUE #34857: [BE] Missing digit in structured communication when control key is single digit
## Bug Description
The function `dolBECalculateStructuredCommunication()` in `/htdocs/core/lib/functions_be.lib.php` generates incorrect structured communications when the modulo 97 control key is a single digit.

## Steps to Reproduce
1. Use invoice number "it-comp-25057" 
2. Call `dolBECalculateStructuredCommunication("it-meal-25057", "0")`
3. Expected: `+++200/0025/05702+++` (12 digits total)
4. Actual: `+++200/0025/0572+++` (11 digits total - missing leading zero in control key)

## Root Cause
The control key calculation returns a single digit (2) but should be zero-padded to always be 2 digits (02).

## Proposed Fix
this line should be added on line 67
```php
$controlKey = sprintf('%02d', $controlKey);
```

juste before
```php
 $invoice_number .= $controlKey;

// Format reference as XXX/XXXX/XXXXX
$part1 = '+++'.substr($invoice_number, 0, 3);
$part2 = substr($invoice_number, 3, 4);
$part3 = substr($invoice_number, 7, 5).'+++'; // Includes last 3 digits + 2 check digits
```

### Dolibarr Version

21.0.1

### Environment PHP

8.1

### Environment Database

MariaDb 10.4.17

_No response_

### Attached files

_No response_

bchanudet:
FIX #34857 Structured communication number checksum is now always correct

This is a PR to merge the fix proposed by @vlandru in #34857. Upon verification and reproduction on my side, I could confirm the code is bad when the result of the modulo operation is less than 10. Adding the `sprintf()` makes sure the result number is always two digits long.

I also created a phpunit test file to make sure the bug can be caught earlier next time. As the 21.0 branch is supposedly frozen, I don't know if that's the good thing to do. Let me know if you want me to do another PR to add the unit test file only in develop.

### DIFF
--- a/htdocs/core/lib/functions_be.lib.php
+++ b/htdocs/core/lib/functions_be.lib.php
@@ -64,7 +64,7 @@ function dolBECalculateStructuredCommunication($invoice_number, $invoice_type)
 	$controlKey = ($mod97 === 0) ? 97 : $mod97;
 
 	// Add the check digit at the end of the reference
-	$invoice_number .= sprintf("%02d",$controlKey);
+	$invoice_number .= sprintf("%02d", $controlKey);
 
 	// Format reference as XXX/XXXX/XXXXX
 	$part1 = '+++'.substr($invoice_number, 0, 3);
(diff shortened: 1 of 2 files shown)


eldy:
Perfect.

PR #35083, opened 2025-08-26
FIX: Correct the calculation of the amount of the current period between the period provided

kkhelifa-opendsi:
Correct the calculation of the sum amount of the current period between the period provided

page /htdocs/compta/resultat/result.php when filtering not all the year (ex: january to june)

### DIFF
--- a/htdocs/compta/resultat/result.php
+++ b/htdocs/compta/resultat/result.php
@@ -505,33 +505,37 @@ if ($modecompta == 'CREANCES-DETTES') {
 							$yeartoprocess++;
 						}
 
-						//var_dump($monthtoprocess.'_'.$yeartoprocess);
-						if (isset($cpt['account_number'])) {
-							$return = $AccCat->getSumDebitCredit($cpt['account_number'], $date_start, $date_end, empty($cat['dc']) ? 0 : $cat['dc'], 'nofilter', $monthtoprocess, $yeartoprocess);
-							if ($return < 0) {
-								setEventMessages(null, $AccCat->errors, 'errors');
-								$resultM = 0;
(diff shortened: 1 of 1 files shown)


rycks:
Here is the bug:

full year:

image


but with half a year total is the same

image

rycks:
Ok for me but that interface is really not clear : what is the "previous period" ?

I think it could be better to replace start date / end date by a drop down with only accounting periods then people can' choose specific date like 1/1/2019 -> 1/4/2020 for example (and in that case what could be "previous period" ?).

for example it will be better (i think) with that list as period user can choose

image

eldy:
In the past, a lot of accounting reports was using just a limited list (year n, year n-1, quarter x), but its was not possible to make report on custom dates, what a lot of users expected to do. 
So we switched these accounting report to use a day start and end. And for consistency, using a day start and end was introduced everywhere, on all reports.
I think filtering on the custom choice is a must have feature (when doing analysis, we often need to use custom periods). But of course we should have calculated values that are correct with the selected range.
The side effect is that it makes the column "previous period" not clear as we don't know what it is. An easy solution that will continue to bring the feature it was designed for when the column was introduced, and that will remove the doubt, is to show the column "previous period" only when selection is a full year. (Previous period column was designed to show result of the previous 12month before the selected 12 month, so we can compare).  This way, we keep the feature and information. And when "previous period" has no sense (or else than full year), we don't show it to avoid questions. But this is another point compared to the fix that is just a fix of calculation of the selected period.

lvessiller-opendsi (approved):

rycks (approved):

PR #35098, opened 2025-08-27
FIX CI reception card

aspangaro:


eldy:
As a different fix was done in develop branch, i prefer to make a cherry-pick from the dev branch.

PR #35115, opened 2025-08-28
FIX 21.0 regression: loss of time of day information when editing agenda events

atm-florianm:
## Bug description
This bug popped up when we upgraded to Dolibarr 21: when you edit an agenda event (especially if it is a message on a ticket), the time of day of that event changes (is set to 00:00).

I didn't investigate the history of the bug, but I think this is because the inputs (`<select>` tags) for hour/minute/second are disabled. When they are, the browser doesn't send their values, so `GETPOSTINT('xyz_hour')` returns 0.

## The fix
I've rewritten the part that reads the time of day to make use of `GETPOSTDATE()`.
When time of day information is not sent with the request, it is then taken from the object itself.

I've attempted to preserve the different processing for full-day events and time zone management. However, there might be some cases that I haven't tested.

### DIFF
--- a/htdocs/comm/action/card.php
+++ b/htdocs/comm/action/card.php
@@ -836,41 +836,29 @@ if (empty($reshook) && $action == 'add' && $usercancreate) {
 if (empty($reshook) && $action == 'update' && $usercancreate) {
 	if (empty($cancel)) {
 		$fulldayevent = GETPOST('fullday');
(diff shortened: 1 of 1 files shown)


eldy:
Yes this is because the inputs (<select> tags) for hour/minute/second are disabled. 
So trouble is surely at everyplace we use the select component.
So i pushed a different fix directly into component.

PR #35130, opened 2025-08-29
NEW: add combining characters (accents, cedilla...) codes in dol_string_unaccent()

jyhere:
How to reproduce the problem : 
- upload a file on dolibarr (e.g on a order) named with a combining character (if needed, use this string : MÉLANIE.txt)
- the uploaded file still has the accented character

There's a lot of combining characters, I only added a few ones (U+0300 to U+0311, and the cedilla).
Full list here :  
[Combining characters](https://www.compart.com/en/unicode/search?q=Combining#characters)


eldy:
Here you replace the char with empty string. May be you can replace with the equivalent string, for example É become E, like we do for lower case strings.

frederic34:
@eldy the combining character is in reality two characters, you can copy paste the example and see what happen when deleting the É (two character E followed by the combining character) with back delete, he is different from É (one character).
It happens with MAC APPLE users...
https://github.com/Dolibarr/dolibarr/pull/35153

jyhere:
A quick demo of the combining character in PHPStorm: pressing Backspace once removes only the accent.
Pressing Delete before the É removes the E, and the accent moves to the M !

Peek 01-09-2025 10-15

PR #35143, opened 2025-08-29
Fix 35142

### LINKED ISSUE #35142: Setting product on/off sell or on/off buy with small buttons top right of the card doesn't create an event in product card
### Bug

After looking in DB, it appears that event is logged, but elementtype is 'product@product' instead of 'product', and because of this, action is not displayed in product's events list in card

### Dolibarr Version

at least >=20

### Environment PHP

_No response_

### Environment Database

_No response_

_No response_

### Attached files

_No response_

vmaury:
# FIX #35142 
[Setting product on/off sell or on/off buy with small buttons top right of the card doesn't create an event in product card]

This solution works, but I don't know if it's the good approach to solve the bug ....

### DIFF
--- a/htdocs/core/ajax/objectonoff.php
+++ b/htdocs/core/ajax/objectonoff.php
@@ -122,6 +122,9 @@ if (($action == 'set') && !empty($id)) {	// Test on permission already done in h
 	if ($triggerkey == 'PRODUCT_UPDATE') {
 		$triggerkey = 'PRODUCT_MODIFY';
 	}
+    if ($triggerkey == 'PRODUCT_MODIFY') {
+        unset($object->module);
+    }
 
 	$result = $object->setValueFrom($field, $value, $object->table_element, $id, $format, '', $user, $triggerkey);
 


htdocs/core/ajax/objectonoff.php
```diff
@@ -122,6 +122,9 @@
 	if ($triggerkey == 'PRODUCT_UPDATE') {
 		$triggerkey = 'PRODUCT_MODIFY';
 	}
+    if ($triggerkey == 'PRODUCT_MODIFY') {
```
eldy, line 125: If $triggerkey = 'PRODUCT_MODIFIY', then the trigger PRODUCT_MODIFIY is executed inside the method setValueFrom() just after.
So i see no reason tohave this working better than now.
Can you describe why you try to unset $object->module and where in code this change something ?

hregis: @vmaury you use the "canvas" system for product card ?

vmaury: No idea, my dolibarr is "stock"

Issue #35116, opened 2025-08-28
Third Party Logo and Entity Type Not Working


pln3to:
### Bug

1. Logo not working

After upload and save a logo, file exists in documents/societe but not showing in Third Party card.

2. Entity Type not working

After create a Third Party and select a entity type and save, when edit or view Third Party card, entity types dropdown is empty.

P.S. It's working well in version 21.0

### Dolibarr Version

22.0

### Environment PHP

8.3.19

### Environment Database

MySQL

edupulpillo:
The part of the problem about the logo of this bug is already mentioned here https://github.com/Dolibarr/dolibarr/issues/34922 and has been fixed in commit https://github.com/Dolibarr/dolibarr/commit/a9c07569fb30a2212b122e0034b0822b90b38eb2 and works fine in V22 with that commit.

The part of the problem about the entity type still is present.

gschratzer:
Has been fixed in commit https://github.com/Dolibarr/dolibarr/commit/a9c07569fb30a2212b122e0034b0822b90b38eb2
in #35152

PR #35152, opened 2025-08-31
FIX wrong country code in edit mode

hregis:


hregis:
@eldy FIX for #35116 thirdparty entity type is empty in edit mode

PR #35156, opened 2025-08-31
FIX #35154 - Massaction createbills in reception list don't mind about order and/or thirdparty currency

### LINKED ISSUE #35154: Massaction createbills in reception list don't mind about order and/or thirdparty currency
### Bug

The massaction createbills in reception list never take care about the multicurrency data.

### Dolibarr Version

20.0

### Environment PHP

8.1

### Environment Database

8.0

Create an supplier order in USD (for example if your Dolibarr instance is in EUR).
Create a reception from this supplier order
In this reception if you create a invoice, this one is in the right currency (the same as the supplier order in my example USD)
If you try to create the invoice from the reception list with the mass action, the currency is set to EUR in the invoice.

### Attached files

_No response_

comaiteseb:
FIX #35154 - Massaction createbills in reception list don't mind about order and/or thirdparty currency

Check in the loop if all reception in a invoice are in the same currency and raise an error if not.
The error message is the translate key: ReceptionMustBeInTheSameCurrencyThanOther

### DIFF
--- a/htdocs/reception/list.php
+++ b/htdocs/reception/list.php
@@ -265,7 +265,7 @@ if (empty($reshook)) {
 				$mode_reglement_id = 0;
 				$fk_account = 0;
 				$transport_mode_id = 0;
-				$mutlicurrency_code = null;
+				$multicurrency_code = null;
 
 				if (!empty($rcp->cond_reglement_id)) {
 					$cond_reglement_id = $rcp->cond_reglement_id;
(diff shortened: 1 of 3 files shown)


htdocs/reception/list.php
```diff
@@ -384,6 +397,13 @@
 			}
 
 			if ($objecttmp->id > 0) {
+				if (!isset($rcp->origin_object)) {
+					$rcp->fetch_origin();
+				}
+				if ($rcp->origin_object->multicurrency_code != $objecttmp->multicurrency_code) {
```
eldy, line 403: There is no amount and no currency on a reception so i don't understand why testing currency.
What is important is only qty.
Can you provide a screenshot of what's wrong ?

PR #35178, opened 2025-09-02
extra rows in the entity rights management

atm-jpb:
# FIX #[  extra rows in the entity rights management ]
Fix an issue where adding an empty  div introduces extra rows in the entity rights management pop-in even if we have no value in it.


atm-jpb:
@eldy probably a bad log on CI-PULL-REQUEST ?

PR #35180, opened 2025-09-02
clean code

frederic34:


htdocs/expensereport/class/expensereport.class.php
```diff
@@ -2852,21 +2852,19 @@ public function getKanbanView($option = '', $arraydata = null)
...
-		$return .= '<span class="info-box-ref inline-block tdoverflowmax150 valignmiddle">'.(method_exists($this, 'getNomUrl') ? $this->getNomUrl(1) : $this->ref).'</span>';
+		$return .= '<span class="info-box-ref inline-block tdoverflowmax150 valignmiddle">' . $this->getNomUrl(1) . '</span>';
 		if ($selected >= 0) {
 			$return .= '<input id="cb'.$this->id.'" class="flat checkforselect fright" type="checkbox" name="toselect[]" value="'.$this->id.'"'.($selected ? ' checked="checked"' : '').'>';
 		}
 		if (array_key_exists('userauthor', $arraydata) && $arraydata['userauthor'] instanceof User) {
 			$return .= '<br><span class="info-box-label">'.$arraydata['userauthor']->getNomUrl(-1).'</span>';
 		}
-		if (property_exists($this, 'date_debut') && property_exists($this, 'date_fin')) {
+		if (isDolTms($this->date_debut) && isDolTms($this->date_fin)) {
```
eldy, line 2862: I think it is not excaclyt the same than old code. If we have juste the date_fin set and not the date_start, on old code, we show the info we have so date_fin alone, when on new code we show nothing, right ?

PR #35190, opened 2025-09-03
Provide direct access to country data

bafbes:
With Providing direct access to country data, we wont need to explode $this->global->MAIN_INFO_SOCIETE_COUNTRY at each time to get country info.


bafbes:
This is right.

Issue #35197, opened 2025-09-03
Auto generation of SN-Lot in MO and Reception if user leaves SN and Lot field blank


accellier-dev:
### Bug

Auto generation of SN-Lot in MO and Reception if user leaves SN and Lot field blank

### Dolibarr Version

20.0

edupulpillo:
Related also to #23957

PR #35198, opened 2025-09-03
Auto generation of SN-Lot in MO and Reception if user leaves SN and Lot field blank #35197

pratushraj:
#35197


support-accellier:
Superseding this with a different approach: instead of silently auto-filling SN/Lot server-side when the field is left blank, the replacement adds an explicit "Generate" button (client-side, via a new AJAX endpoint), so nothing is filled without the user asking for it - avoids surprising behavior and is easier to review/test. It also targets develop directly rather than 20.0. Closing this one in favor of the new PR against #35197.

PR #35199, opened 2025-09-03
Fix impossibility to validate objects

Hystepik:
Fix a bug that make impossible to validate object when you have a core/module/<ObjectClass> directory in custom folder

To prevent this bug we try to find a file in the directory.

### Step to reproduce
We will try with invoice object but works on all objects

1.  set $dolibarr_main_document_root_alt = <pathtodolibarr>/custom
2.  create an empty directories with this patch  in custom directory: core/modules/facture

Now if you go in invoice configuration page there is no numbering files
If you try to create a new draft invoice and try to validate it. It create a fatal error.

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -1443,6 +1443,7 @@ function dol_include_once($relpath, $classname = '')
 function dol_buildpath($path, $type = 0, $returnemptyifnotfound = 0)
 {
 	global $conf;
+	require_once DOL_DOCUMENT_ROOT.'/core/lib/files.lib.php';
 
 	$path = preg_replace('/^\//', '', $path);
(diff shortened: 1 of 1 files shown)


sonikf:
Hi @Hystepik 
Can you describe a use case for having a  standalone core/module/ directory in custom folder?

Hystepik:
Hi sonikf

For exemple if you have custom docs for a invoice module you will have custom/core/modules/facture/docs/modele_pdf.php

Like that facture directory will be empty and will create the fatal error

sonikf:
Thanks, Haven't thought of that I always create modules...

Hystepik:
Closed to do PR in Dev with eldy advices  https://github.com/Dolibarr/dolibarr/pull/35202

PR #35219, opened 2025-09-04
Fix discount applied 2 times on contract

SylvainLegrand:
on update line the discount was apply 2 times !
no problem on insert line.

Already solved on v22

### DIFF
--- a/htdocs/contrat/card.php
+++ b/htdocs/contrat/card.php
@@ -756,10 +756,11 @@ if (empty($reshook)) {
 
 			$price_ht =  price2num(GETPOST('elprice'), 'MU');
 			$remise_percent = price2num(GETPOST('elremise_percent'), 2);
-			if ($remise_percent > 0) {
+			// Discount applied 2 times => see line 803
+			/*if ($remise_percent > 0) {
 				$remise = round(($price_ht * $remise_percent / 100), 2);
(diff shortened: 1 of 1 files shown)


rycks:
Just for info & tracking here is the patch on v22 but it's linked with bigger changes against llx_contratdet table, thanks to @atm-florianm comment :

> [breaking change] Remove deprecated columns `price_ht` and `remise` from llx_contratdet and corresponding class properties:

https://github.com/Dolibarr/dolibarr/commit/c9ecc5dfadfdd20a9136b1f2182a1c8e66d46a40

rycks:
@eldy what about that sort of fix considering others versions like 19.0 ... ? 

could it be better to remove the code but not cherry picking v22 fixe due to the sql changes in that version ?

lvessiller-opendsi (approved):

eldy:
It is clear that we can't do the automatic cherry-pick. But we can do a manual cherry-pick to take only the code line changed to match the same change as much as possible.

So what did Sylvain seems good.

Then when mergind from v18 to higher version, when a merging of v18 will be done on v19, v20, ... a conflict will surely appear when we reach the version that has a change for the samebug if change was done differently of inside anohter change. In this case a manuel correction of the conflict, expected by the system, will solve this. And all other following merge will then be successfull because conflict was already solved.

PR #35220, opened 2025-09-04
FIX or UPDATE Order Status linked with Invoice Billed status

Pichinov-Jose:
FIX or UPDATE commande.class.php Function LibStatut to revert previous link with OrderBilled status

For some reasons I could not find or understand, starting V19, the status displayed in order list or order card have been simplified creating confusion and less feature when working on mass order management for all our customers.
I added a CONST ORDER_STATUS_LINK_BILLED to included previous code status management back.
As a reminder a order that was shipped but not invoiced was Green and tag shipped. If Shipped and Invoiced an order was tagged Gray and Processed.
I put this behavior back starting V19 as it was already in code v18 and below.

### DIFF
--- a/htdocs/commande/class/commande.class.php
+++ b/htdocs/commande/class/commande.class.php
@@ -3706,9 +3706,21 @@ class Commande extends CommonOrder
 				$labelTooltip .= ' - '.$langs->transnoentitiesnoconv("DateDeliveryPlanned").dol_print_date($this->delivery_date, 'day').$billedtext;
 			}
 			$statusType = 'status4';
-		} elseif ($status == self::STATUS_CLOSED) {
+		} elseif ($status == self::STATUS_CLOSED && empty(getDolGlobalString('ORDER_STATUS_LINK_BILLED'))) {
 			$labelStatus = $langs->transnoentitiesnoconv('StatusOrderDelivered').$billedtext;
 			$labelStatusShort = $langs->transnoentitiesnoconv('StatusOrderDeliveredShort').$billedtext;
+			$statusType = 'status6';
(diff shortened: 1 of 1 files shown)


eldy:
The status "billed" or not a status of order but a second status "billed" or not that is different of "Order processed" or not"
This is why now we have the info "Billed" or not inside the label
We must have all combinations or all the 2 status So it seems we still have different status badge for all combination of cases, so we should have 6 cases:

For exmple on a validated order, we have
image
or
image

For order status "In process"
image
or
image

For example when order is processed
image
or
image

The status billed or not is not correlated with status of order, it is another workflow that can be started before, during or after the order process.
This is what we have currently so i don't understand the goal of your PR.
What would be the badge for the 6 cases in your variant mode ?

Pichinov-Jose:
Hi @eldy 

Ok I understood now reviewing deeper the code the new behavior.
The issue is that it does not show Billed in order list in the Status Column. 
This is unclear as the list does not work int the same way as before. I know there is the Column Billed but our customers were used to be focusing on one column only (Status One) with 2 green colors for Shipped and In process and Grey if Shipped and Billed.

Therefore we have to get them work either 
* as they were used in the past (This is the purpose of this PR)
* or give them option to display in lists the Billed Value to keep all data on same column. A new small PR with option to change $generic_commande->LibStatut($obj->fk_statut, $obj->billed, 5, **1**) by $generic_commande->LibStatut($obj->fk_statut, $obj->billed, 5, **0**)
* or to let them work by new design.

I will check further tomorrow with several customers to explain them deeper.

Pichinov-Jose:
Hi @eldy 
I finally used a hook to display in list Livré - Facturé on status to keep customer productivity. I am closing for now this topic. I will come back with CONST if someone needs it.

Issue #35232, opened 2025-09-05
New PDF mode Delivery note in Purchase Order


pratushraj:
### Feature Request

New PDF mode Delivery note in Purchase Order

PR #35233, opened 2025-09-05
#35232

pratushraj:


eldy:
Can you describe the difference of this template compared to default template of order ?
Or better, a screenshot of the generated PDF

eldy:

PR #35250, opened 2025-09-08
NEW Accountancy - Add a protection on various payment for auxiliary account on general account not centralized

aspangaro:
Today, too many mistakes are being made. Many users specify sub-ledger accounts on G/L accounts that are not centralized (e.g. 411/401/421). This leads to difficulties in importing to accounting software and problems in understanding the accounting.

Solution: Add a message to prevent and not recorded auxiliary account. It's not an error, we continued !

Example: Create a various payment with an auxiliary account and a general account not centralized

image

image

image

image


eldy:
Why recording with reseting the value.
it looks more natural to me to report the error without recording like we do in other forms when data are not consistent. We can complete the error message with "If you want to use this general account, keep the subsidiary account empty, or if you want to set a sbusidirary account, use a general account that is a centralized account".
So we remain on the same page to allow the user the god way to fix its form and he have to submit the form once fixed.
It is more consistent with other form and allow the user to fix in 2 directions (by removing subsidiary or by fixing the general account, both arror are possible).

PR #35273, opened 2025-09-09
#35271 [SQL] add: mandatory signature fields

nicolas-eoxia:
Data model for add feature we want add checkbox on for tell who can sign in contact
Ex: On contract card we want 2/3 signatories
-- Later we add feature for link external signature link with mandatory signature fields on contact.


lmag (approved):
I love it

PR #35284, opened 2025-09-10
fix fatal error

frederic34:
If an external module updates the object in a trigger, the 'oldcopy' is a stdClass.

### DIFF
--- a/htdocs/product/class/product.class.php
+++ b/htdocs/product/class/product.class.php
@@ -1493,8 +1493,8 @@ class Product extends CommonObject
 
 		if ($result >= 0) {
 			// $this->oldcopy should have been set by the caller of update (here properties were already modified)
-			if (is_null($this->oldcopy) || (is_object($this->oldcopy) && $this->oldcopy->isEmpty())) {
-				$this->oldcopy = dol_clone($this, 1);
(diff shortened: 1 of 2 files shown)


htdocs/product/class/product.class.php
```diff
@@ -1493,8 +1493,8 @@ public function update($id, $user, $notrigger = 0, $action = 'update', $updatety
 
 		if ($result >= 0) {
 			// $this->oldcopy should have been set by the caller of update (here properties were already modified)
-			if (is_null($this->oldcopy) || (is_object($this->oldcopy) && $this->oldcopy->isEmpty())) {
-				$this->oldcopy = dol_clone($this, 1);
+			if (is_null($this->oldcopy) || (is_object($this->oldcopy) && $this->oldcopy->isEmpty()) || !method_exists($this->oldcopy, 'hasbatch')) {
+				$this->oldcopy = dol_clone($this, 1); // 1 to clone with methods to avoid fatal error with $this->oldcopy->hasbatch()
```
eldy, line 1497: Here $this->propertyx  may have been modified. So if we erase the ->oldcopy, the ->oldcopy will get the value of the new instance making trouble with some triggers later for example.

Do you remember the process that make the error. May be we should fix the caller that  call the ->update() and make before a dol_clone with bad mode ?

PR #35288, opened 2025-09-10
fix: with PRODUCT_USE_UNIT preset the unit in document

FHenry:
# Instructions
- Set PRODUCT_USE_UNIT to 1
 - On a product set m2 as default unit

create a Proposal, select the product, the unit is not defaulted with the product unit selected

# PR Detail
Add default unit as data provided by htdocs/product/ajax/products.php

Use it in template htdocs/core/tpl/objectline_create.tpl.php if needed


glu000:
This bug exists in V21 too

PR #35290, opened 2025-09-10
feat: Subtotal module version Dolibarr and fr translation

FHenry:


frederic34:
https://github.com/Dolibarr/dolibarr/pull/35296

PR #35293, opened 2025-09-10
UIUX adds a new permission that restricts the user to seeing their salaries and only those.

mapiolca:
Even if a user is the immediate superior, they may not have the right to view their subordinates' salaries. This patch therefore adds an even lower level of permission to allow for this level of privacy.


mapiolca:
@eldy, do you want I create the PR for v22 ? or Develop ?

eldy:
For develop.
v22 is also frozen and released so no change in behaviour or features is allowed.

mapiolca (commented):
Fix

mapiolca (commented):
add space

mapiolca:
What I need to do to resolve the last failing test ?

htdocs/salaries/list.php
```diff
@@ -361,9 +361,13 @@
 $sql .= " ".MAIN_DB_PREFIX."user as u";
 $sql .= " WHERE u.rowid = s.fk_user";
 $sql .= " AND s.entity IN (".getEntity('salaries').")";
+if (!$user->hasRight('salaries', 'readchild')) {
+	$sql .= " AND s.fk_user = ".$user->id;
```
eldy, line 365: Try to force the cast into int of the user->id

mapiolca (commented):
force $user->id to (int)

mapiolca (commented):
add spaces

mapiolca:
It seems the error come from a file that I don't touch ? right ? What's the process to solve yhis issue ?

mapiolca (commented):
update

mapiolca (commented):
force int

mapiolca:
Youhouuuu !! Please @eldy, merge quickly before a new update of develop branch 😉

PR #35299, opened 2025-09-11
Prepare 18.0.8

rycks:
Due to last CVE fix it's time to make a new 18.0 release

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,36 @@
 English Dolibarr ChangeLog
 --------------------------------------------------------------
 
+***** ChangeLog for 18.0.8 compared to 18.0.7 *****
+35 files changed, 647 insertions(+), 298 deletions(-)
+
+FIX: #34746 - More complete fix for CVE-2024-40137
+FIX: Correct the calculation of the amount of the current period between the period provided (#35083)
+FIX: Add security test for show terminal selection if no terminal selected when invoice.php is call (#34717)
+FIX: Add security test for show terminal selection if no terminal selected when invoice.php is call
+FIX: missing quick edit for extrafields (baclport commit 4fc66c6) (#35160)
+FIX: Missing sentence part (#35144)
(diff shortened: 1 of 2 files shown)


eldy:
I modified the define('DOL_VERSION', '18.0.8-LTS'); into define('DOL_VERSION', '18.0.8');
because the variable DOL_VERSION is a technical information to allow comparison with other for some features, so should not contains other text information. Also it creates a confusion of what is a LTS : What is LTS is the "major" version so the branch version "18" not the maintenance tag 18.0.8. The 18.0.x version are very short time version, thay are time limited versions because they are few weeks after replaced with 18.0.(x+1), when a LTS last several years...

rycks:
@eldy ok for the "LTS" mention into version but the message just on the right of the number version make me trouble with your message :

```// a.b.c-alpha, a.b.c-beta, a.b.c-rcX or a.b.c```

so if "-string" is forbiden like you explain may we change the comment to remove -alpha, -beta, -rc and keep only a.b.c ?

eldy:
> @eldy ok for the "LTS" mention into version but the message just on the right of the number version make me trouble with your message :
> 
> `// a.b.c-alpha, a.b.c-beta, a.b.c-rcX or a.b.c`
> 
> so if "-string" is forbiden like you explain may we change the comment to remove -alpha, -beta, -rc and keep only a.b.c ?

alpha, beta are string similar to the third number to get a sorting number of version x.y.z-alpha is lower than x.y.z-beta that is lower than x.y.z that is lower than x.y.z+1
It is managed by replacing the alpha with -3, the beta with -2 number so we gomt a sorting number.
It allows to follow if we install a version lower or higher of current version to know if we must force an upgrade process or not for example or if we were using an unstable version (with a negative 4th number) to show warning. They is other use at different places of this version number that must match a syntax that is sortable with other value to not break Features.

Issue #35291, opened 2025-09-10
Add "__EVENT_DESCRIPTION__" variable


Liammer:
### Feature Request

It would be great to add an "__EVENT_DESCRIPTION__" variable to put the event's description in email when sending a reminder.

### Suggested implementation

Add a line in core/lib/functions.lib.php (at line 9655): 

`$substitutionarray['__EVENT_DESCRIPTION__'] = $object->note;`

PR #35302, opened 2025-09-11
Add "__EVENT_DESCRIPTION__" variable

Liammer:
To be able to add the event's description in reminder emails.


emheyarssi:
#35291

PR #35313, opened 2025-09-11
exclude escpos-php/doc from filelist_xml

rycks:
sorry for that PR on 18.0 branch but here is the result of integrity test of 18.0.8

image

/includes/mike42/escpos-php/doc/FAQ.md is missing because of the main pl script who makes a rm -rf of that directory ... but the file is included into the php script who make integrity list so i have to make a choice


eldy:
Yes, the scope of the makepack must be the same than the scope of the generate_xml file.
If removed in first, must be excluded in the second... but also into api_setup.class.php so i pushed a commit to complete tihs PR.

Issue #35330, opened 2025-09-12
Extrafieds bug in supplier invoices list


vmaury:
### Bug

in table header, instead of extrafield filters, this message appears
ERROR : selectForForms: Error bad setup of field objectdescorig=

### Dolibarr Version

21

### Environment PHP

any

vmaury:
<img width="5448" height="2997" alt="Image" src="https://github.com/user-attachments/assets/c2ee31ce-4550-494a-97a1-01ff0d7ca95b" />

PR #35331, opened 2025-09-12
Fix bug #35330

vmaury:
# FIX#35330 
in table header, instead of extrafield filters, this message appears
ERROR : selectForForms: Error bad setup of field objectdescorig=


vmaury:
I tested just now, this bug is only on v21, and no more in v22/develop

PR #35332, opened 2025-09-12
add check in dolibarr phpinfo page

frederic34:


htdocs/admin/system/phpinfo.php
```diff
@@ -235,6 +235,20 @@
...
 
+$functions = array();
+$name = "bz2";
+print "<tr>";
+print "<td>".$name."</td>";
+print getResultColumn($name, $activatedExtensions, $loadedExtensions, $functions);
+print "</tr>";
+
+$functions = array();
+$name = "bcmath";
```
eldy, line 246: Do we need this php extension "bcmath" ?

frederic34: bcmath may be used in swiftmailer
image
I also seen the need in some external module...

PR #35356, opened 2025-09-15
FIX #34678 Expense Reports now don't use VAT when Organization don't use Sales tax

### LINKED ISSUE #34678: [EXPENSE REPORT] In expense report, when I add a lign, the VAT is 20% by default, while I expect 0% (association)
### Bug

In expense report, when I add a lign, the VAT is 20% by default, while I expect 0% (association)

### Dolibarr Version

21.0.1

### Environment PHP

 8.1.2-1ubuntu2 

### Environment Database

MySQL or MariaDB 5.5.5-10.6.18-MariaDB-0ubuntu0.22.04.1 - mysqli (mysqlnd 8.1.2-1ubuntu2.20)

- I create an expense report ;
- I want to add a lign with an expense. I see the VAT with the value 20% by default, while I expect 0% because the configuration is for an association (well configured in Dolibarr parameters) ;
- if I add the lign with this default VAT, then I modify the lign, the VAT is modified automatically with 0% (before my action and validation).

Tested on my instance (version 21.0.1) and a Dolicloud test version (version 21.0.1 too)

### Attached files
Image
Image
Image

bchanudet:
# FIX #34678 Expense Reports now don't use VAT when Organization don't use Sales tax

`htdocs/expensereport/card.php` uses the function `load_tva()` twice in the file: 
- first when editing an already-existing line;
- second when adding a new line in the expense report.

The second one, as reported in #34678, displays all VAT rates even if the organization set in Dolibarr doesn't use Sales taxes. The problem becomes even more obvious where you try editing the line you just added: there, the VAT rates are not displayed, and if you entered one at the beginning, now you are losing the information.

I found out that the call to `load_tva()` doesn't use the same parameters in the two cases. On the correct one, `$mysoc` is used to fill in the `$societe_acheteuse` parameter, whereas `null` is passed in the buggy case. Passing `$mysoc` makes `load_tva()` not displaying any VAT rates as intended.

Bug was reproduced on 21, 22, and develop. I haven't checked on older versions.

### DIFF
--- a/htdocs/expensereport/card.php
+++ b/htdocs/expensereport/card.php
@@ -2634,7 +2634,7 @@ if ($action == 'create') {
 					// If option to have no default VAT on expense report is on, we force MAIN_VAT_DEFAULT_IF_AUTODETECT_FAILS
 					$conf->global->MAIN_VAT_DEFAULT_IF_AUTODETECT_FAILS = 'none';
 				}
-				print $form->load_tva('vatrate', (!empty($vatrate) ? $vatrate : $defaultvat), null, null, 0, 0, '', false, 1);
+				print $form->load_tva('vatrate', (!empty($vatrate) ? $vatrate : $defaultvat), $mysoc, null, 0, 0, '', false, 1);
 				print '</td>';
 
 				// Unit price net


htdocs/expensereport/card.php
```diff
@@ -2634,7 +2634,7 @@
 					// If option to have no default VAT on expense report is on, we force MAIN_VAT_DEFAULT_IF_AUTODETECT_FAILS
 					$conf->global->MAIN_VAT_DEFAULT_IF_AUTODETECT_FAILS = 'none';
 				}
-				print $form->load_tva('vatrate', (!empty($vatrate) ? $vatrate : $defaultvat), null, null, 0, 0, '', false, 1);
+				print $form->load_tva('vatrate', (!empty($vatrate) ? $vatrate : $defaultvat), $mysoc, null, 0, 0, '', false, 1);
```
eldy, line 2637: When you enter an expense report, it is the content of what an employee bought to a seller.
The seller is usinga vat when he do an invoice (whatever is you vat status).
So by default we should report this vat one expense report.
So null seems the good value because when a company is selling something to your employee, your employee pay full price including the vat and he is not $mysoc.

bchanudet: To be honest, I'm not knowledgeable on how expense reports are handled in "non-sales-taxes" organizations such as associations and liberal professions.

In any case, the current state is still bad: if you add a line with 20% VAT, but you try to edit it right after, the VAT selector is disabled, shows 0% with no way of changing it. So you lose the VAT information. 

This PR is making sure that the behavior is the same both when creating and editing a line. The changes could be "reversed" by removing the `$mysoc` used on [the other call to `load_tva()`](https://github.com/Dolibarr/dolibarr/blob/ea7152a818e7512c77b79be28ac3a0bd2ac0e24c/htdocs/expensereport/card.php#L2472), so the behavior with VAT is the same in all cases.

PR #35365, opened 2025-09-16
clean code

frederic34:


dev/setup/codesniffer/ruleset.xml
```diff
@@ -134,7 +134,7 @@
 
 	<rule ref="Generic.Formatting.SpaceAfterCast" />
 
-	<rule ref="Generic.Functions.CallTimePassByReference" />
+	<!-- <rule ref="Generic.Functions.CallTimePassByReference" /> -->
```
eldy, line 137: What's wrong with this rule ?

PR #35369, opened 2025-09-17
Fix:prevent handling update action by  generic include of actions_addupdatedelete.inc.php

bafbes:
Fix:
To prevent the generic post-update action in inventories (/core/actions_addupdatedelete.inc.php) from handling the update action again, the $action variable must be unset after the first handling; otherwise, the $object fields will be overwritten with empty values.

### DIFF
--- a/htdocs/product/inventory/inventory.php
+++ b/htdocs/product/inventory/inventory.php
@@ -282,10 +282,10 @@ if (empty($reshook)) {
 
 		if (! $error) {
 			$db->commit();
-			$action = '';
 		} else {
 			$db->rollback();
 		}
+		$action = '';
 	}
 
 	// Save quantity found during inventory (when we click on Save button on inventory page)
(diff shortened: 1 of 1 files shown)


pixodeo:
Could you point at a concrete repro? I have walked `product/inventory/card.php` and the include only handles `action=update` once (`core/actions_addupdatedelete.inc.php:258`). On success it resets `$action = 'view'` (line 388); on error it sets `$action = 'edit'` (line 399 / 402). There is no second handler in inventory/card.php that picks up `update` again after the include.

If the symptom you saw is fields ending up empty, the more likely cause is the include's foreach loop dispatching one of the typed branches (sellist, integer, html, etc.) with a missing POST key. Could you share:

1. Which inventory field gets blanked.
2. The relevant `type` value on `Inventory::$fields[<field>]` (from `product/inventory/class/inventory.class.php`).
3. The POST keys actually present when the update form is submitted (browser devtools form data).

That will tell us whether the fix belongs in the include or on the inventory class field definition.

PR #35381, opened 2025-09-17
Repair recalculateinvoicetotal

vdegrandpre:
# FIX|Fix #[*29565 Script to calculate invoice totals from invoice lines*]
This corrects invoice total calculation when many taxes are applixed. Because of decimal precision, an over of 0,01 appears in the sum.


htdocs/install/repair.php
```diff
@@ -2016,14 +2019,23 @@
...
-							total_tva = ".($obj_calcul->total_tva ? price2num($obj_calcul->total_tva, 'MT') : 0).",
-							localtax1 = ".($obj_calcul->localtax1 ? price2num($obj_calcul->localtax1, 'MT') : 0).",
-							localtax2 = ".($obj_calcul->localtax2 ? price2num($obj_calcul->localtax2, 'MT') : 0).",
-							total_ttc = ".($obj_calcul->total_ttc ? price2num($obj_calcul->total_ttc, 'MT') : 0)."
+							total_ht = $totHt,
+							total_tva = $totTva,
+							localtax1 = $totLocal1,
+							localtax2 = $totLocal2,
+							total_ttc = $totTtc,
+							fk_statut = ".($totTtc == price2num($montantPaiements, 'MT') ? 2 : 1 )."
```
eldy, line 2038: Not that situation should not happen because total should be rounded in database so total of all taxes + total_ht should match total_ttc.
Did you know how to reproduce the corruption of data ?

PR #35384, opened 2025-09-18
Fix localtax cumulative precision errors impact on calculated total

### LINKED ISSUE #35382: Wrong invoice local tax total sometimes
### Bug

Sometimes, two invoices with same total (without taxes) will have different localtax total calculated, depending on the number of lines. 

This happen even with MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND = 1 in global configuration.

### Dolibarr Version

? ... 19.0.2 - 23.0.0-alpha

### Environment PHP

8.2.1

### Environment Database

mariadb 10.6

1. Create two invoices that have the same total without taxes, one with minimal lines, other with many more lines;
2. Compare calculated totals; localtax1 totals are not the same.

### Attached files

Image
Image

vdegrandpre:
# FIX|Fix #35382 Wrong invoice local tax total sometimes

Add +2 precision decimals to localtax numbers of object lines to avoid situation where multiple lines exists and rounding errors cumulate. 

Example with localtax 9.975 %

Legend: 

Rx : Real number calculated = Amount * 9.975 % (in R with not decimal limits)
𝚺Ri : Cumulated sum of Rx up to i (line number)
Error : Rounded float (system current in R with 2 decimals limit)
𝚺Error : Sum of rounded with cumulated error

| Line No.  | Amount | Rx | 𝚺Ri | Error | 𝚺 Error 
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| 1  | 240  | 29.94  | 23.94  | 23.94  | 23.94  |
| 2  | 175  | 17.4563  | 41.3963  | 17.46  | 41.4  |
| 3 | 175  | 17.4563  | 58.8525  | 17.46  | 58.86  |
| 4  | 175  | 17.4563  | 76.3088  | 17.46  | 76.32  |
| 5  | 175  | 17.4563  | 93.765  | 17.46  | 93.78 |
| 6  | 175  | 17.4563  | 111.221  | 17.46  | 111.24  |
| 7  | 175  | 17.4563  | 128.678  | 17.46  | 128.7  |
| 8  | 175  | 17.4563  | 146.134  | 17.46  | 146.16  |
| 9  | 175  | 17.4563  | 163.59  | 17.46  | 163.62  |
| 10  | 175  | 17.4563  | 181.046 | 17.46  | 181.08  |

After 10 lines of items, there is a cumulated error of 0,03 (𝚺Ri rounded to 2 decimals).

With this fix, the sum of the tax amounts are accounted so that no more rounding errors cumulate.


eldy:
Your fix seems to generate a lot of troubles and calculation errors at different place of the application.
Yes rounding change may cumulate, but this is the expected value because the accountancy must be done "by line" and not on a final total of an object. 
On a given line, the total MUST match the exact value stored into database, with the rounding defined by the currency (2 digits for final total, 5 digit for unit total by default but may be you need of more important number of digit on unit amount). We must not have a different accuracy than the setup in value store in database and returned by calcul_price. Also we must not have a difference when doing the total on each tax on a given line, we should get the exact total saved on the line.
The total of rounded line will always differ of the round of the total and what modern accountancy need nowadays is the total of rounded lines.
Can you point me an example of a line that has a wrong total on a given line ?  (i don't understand your table, can you show detail of what you get in database of the line that has trouble:  total, total_ht, total_ttc, localtax1 and localtax2) ?

vdegrandpre:
Here you go, I've generated 3 CSVs and included 2 screenshots that reproduces the problem. Can't we put line totals with more decimals so that rounding errors are not a problem anymore without affecting the system?

CSV1 : Current system calculation without this fix
CSV2 : Current system calculation with minimized lines (this is not an option for the user, but just to show that for same total_ht, calculation give two total_ttc amounts)
CSV3 : System calculation with this fix

[CSV3-llx_facturedet-avec-35384-sans-erreur-cumulatives.csv](https://github.com/user-attachments/files/22409301/CSV3-llx_facturedet-avec-35384-sans-erreur-cumulatives.csv)
[CSV2-llx_facturedet-version-compacte-avec-lignes-minimales-sans-erreurs-car-sans-accumulation-derreurs.csv](https://github.com/user-attachments/files/22409302/CSV2-llx_facturedet-version-compacte-avec-lignes-minimales-sans-erreurs-car-sans-accumulation-derreurs.csv)
[CSV1-llx_facturedet-sans-35384-avec-erreurs-cumulatives.csv](https://github.com/user-attachments/files/22409303/CSV1-llx_facturedet-sans-35384-avec-erreurs-cumulatives.csv)

And screenshots to illustrate the situation (same as in #35382) 

Capture d’écran du 2025-09-17 17-27-14
Capture d’écran du 2025-09-17 17-25-54

Thanks

vdegrandpre:
added legend and conclusion in description

PR #35388, opened 2025-09-18
NEW add filter by invoice categories in the statistics module

developmentOYR:
# NEW
Add filter by invoice categories in the statistics module.

image


eldy:
Same on order and proposal would be welcome  ;-)

developmentOYR:
ok give me some time

PR #35392, opened 2025-09-18
fix CI :bug:

frederic34:


htdocs/core/lib/company.lib.php
```diff
@@ -954,7 +954,7 @@ function show_projects($conf, $langs, $db, $object, $backtopage = '', $nocreatel
 		$langs->load("projects");
 
 		$newcardbutton = '';
-		if (isModEnabled('project') && $user->hasRight('projet', 'creer') && empty($nocreatelink)) {
```
frederic34, line 957: already checked at line 953

htdocs/core/lib/company.lib.php
```diff
@@ -1133,7 +1133,7 @@ function show_projects($conf, $langs, $db, $object, $backtopage = '', $nocreatel
 						// To verify role of users
 						$userAccess = $projecttmp->restrictedProjectArea($user);
 
-						if ($user->hasRight('projet', 'lire') && $userAccess > 0) {
```
frederic34, line 1136: already checked at line 953

PR #35422, opened 2025-09-21
ticket create event no save  fk_task

cehojac:
// Special handling for fk_task to ensure it's properly saved
if ($field === 'fk_task') {
    $this->actioncomm->fk_task = (int) $value;
    continue;
}

// Auto-assign fk_task from elementid if fk_task is empty and elementtype is ticket
if (empty($this->actioncomm->fk_task) && !empty($request_data['elementid']) && !empty($request_data['elementtype']) && $request_data['elementtype'] === 'ticket') {
    $this->actioncomm->fk_task = (int) $request_data['elementid'];
}


htdocs/comm/action/class/api_agendaevents.class.php
```diff
@@ -250,8 +250,19 @@ public function post($request_data = null)
 				continue;
 			}
 
+			// Special handling for fk_task to ensure it's properly saved
```
eldy, line 253: This part of code seems to do the same things than the line 259
`$this->actioncomm->$field = $this->_checkValForAPI($field, $value, $this->actioncomm);`

do you really need it. Can you test without ?

PR #35429, opened 2025-09-22
Fix 35428

### LINKED ISSUE #35428: retained_warranty_date_limit not recorded
### Bug

When changing retained_warranty_date_limit (in situation invoice) the date is not recorded

### Dolibarr Version

>=21

### Environment PHP

_No response_

### Environment Database

_No response_

_No response_

### Attached files

_No response_

vmaury:
# Fix #35428

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -471,9 +471,14 @@ if (empty($reshook)) {
 		}
 	} elseif ($action == 'setretainedwarrantydatelimit' && $usercancreate) {
 		$object->fetch($id);
-		$result = $object->setRetainedWarrantyDateLimit(GETPOSTFLOAT('retained_warranty_date_limit'));
-		if ($result < 0) {
-			dol_print_error($db, $object->error);
+		$retainedWarrantyDateLimit = GETPOST('retained_warranty_date_limit');
+		if (preg_match('/^\d{4}-\d{2}-\d{2}$/', $retainedWarrantyDateLimit)) {
+			$result = $object->setRetainedWarrantyDateLimit($retainedWarrantyDateLimit);
+			if ($result < 0) {
+				dol_print_error($db, $object->error);
+			}
+		} else {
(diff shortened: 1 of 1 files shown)


htdocs/compta/facture/card.php
```diff
@@ -471,9 +471,14 @@
 		}
 	} elseif ($action == 'setretainedwarrantydatelimit' && $usercancreate) {
 		$object->fetch($id);
-		$result = $object->setRetainedWarrantyDateLimit(GETPOSTFLOAT('retained_warranty_date_limit'));
-		if ($result < 0) {
-			dol_print_error($db, $object->error);
+		$retainedWarrantyDateLimit = GETPOST('retained_warranty_date_limit');
+		if (preg_match('/^\d{4}-\d{2}-\d{2}$/', $retainedWarrantyDateLimit)) {
```
eldy, line 475: We should never used such things when reading date, format vary on language..
Can you try
$retainedWarrantyDateLimit = GETPOSTDATE('retained_warranty_date_limit');

and test if value is empty or not

vmaury:
in this case GETPOSTDATE is not working. I used same tests for same var in same file :-)

eldy:
i see no reason why GETPOSTDATE would not work.
What do you mean with "it is not working" ? What's wrong ?

vmaury:
The reason is that the date which is posted by the form has a yyyy-mm-dd format only (not 3 int values that can be received by GETPOSTDATE)
And I'm working in french config (the date posted is NOT dd/mm/yyyy)

eldy:
The date should be posted in 3 fields: year, month, adn day,  the date in the string field must not be used as format wil always be different.
If the 3 int values to use by GETPOSTDATE are not provided, the bug is here. The 3 values must always be provided for a date field.
The field in form is probably not an input date field implemented with select_date and is instead of string field. For a date input field, we must always use the date selector and not a free text string that will always be wrong (pb of language syntax, but also problem of timezone).

PR #35438, opened 2025-09-22
Add hook on expense api

defrance:
Like on interventional


defrance (commented):
global $hookmanager added

defrance (commented):
suppr space

defrance (commented):
x

defrance (commented):
ok

PR #35466, opened 2025-09-24
NEW API User - Generate Api Key

aspangaro:
User Endpoint to generate API key for an user. (only user admin)

### DIFF
--- a/htdocs/user/class/api_users.class.php
+++ b/htdocs/user/class/api_users.class.php
@@ -1088,6 +1088,47 @@ class Users extends DolibarrApi
 		return $this->_cleanObjectDatas($notification);
 	}
 
+	/**
+	 * Generate an API key for a user (only admin)
+	 *
+	 * @since    23.0.0    Initial implementation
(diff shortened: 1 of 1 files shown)


eldy:
There is already an api to generate the api key: It is the /login message with param reset=1 (this one generate an api key so do the same job than your PR but in a more secured manner because there is more protection around it)

Note: generating an api key from api is someting that should never be done as it means your API are not secured, because by definition the security on api is based on "having the token" and only on having the token (a user/login must never be used in an api called). If you don't have ithe token, you should never be able to run any api. 

This is why in future, the api /login/token generation will be blocked by default with constant API_DISABLE_LOGIN_API. This constant is already implemented but is off by default and not yet available into api setup page. In future it will be on by default, so the generation of a token from api (a commonly admitted bad practice) will be possible only if the admin user has clearly allowed it.

aspangaro:
Thanks @eldy for your return. Very explicit. I closed the PR.

PR #35491, opened 2025-09-26
QUAL  : Doc for the new intuitive table line selection

thersane-john:
# QUAL
linked to #35444 
Move experimental doc to content doc below the table doc and use native js include in dolibarr 

image


thersane-john:
linked to #35444

PR #35522, opened 2025-09-27
Fix #35520

### LINKED ISSUE #35520: SQL problem with extrafields when using SQL functions like concat_ws
### Bug

When using extrafields it oftne makes sense to have a combination of 2 DB-fields as label. Till V22 it was possible to use a string using SQL functions. For example you can have a Extrafield of type Dropdownlist/Select from DB

`ticket:concat_ws (' - ',ref,subject):rowid::(fk_statut:<:8)`

In this example, the label is the combination of ticket number AND ticket subject as label of the extrafield. This is used very often to help the user to select the correct item. 

In V22 this throws an error:

`Error in request SELECT rowid as rowid, concat_ws (' - ',ref,subject) FROM llx_ticket WHERE ((fk_statut < 8)) ORDER BY concat_ws ASC, ref ASC, subject ASC LIMIT 1000 Unknown column 'concat_ws' in 'order clause'. Check setup of extra parameters.`

### Dolibarr Version

22.0

### Environment PHP

na

### Environment Database

na

see above

### Attached files

_No response_

glu000:
FIX #35520
SQL problem with extrafields when using SQL functions like concat_ws


eldy:
Removing the ->order is annoying because it is an important part of the anti SQL injection protection.
So i pushed a better PR to this fix by enhancing the ->order method.

Note that you can also use this syntax to concat fields :
ticket:ref|subject:rowid::(fk_statut:<:8)
It was already implemented in v22 to avoid to use function, but when using | , fields will appears separated by a space in combo list.

or you can use now with my suggested fix a function like the one found into issue, but in this case, you must add a "as xxx" so it become possible to sort on this concatenated field
ticket:concat_ws(' - ', ref, subject) as labelconcatenated:rowid::(fk_statut:<:8)

glu000:
Thanks, @eldy - thant's a perfect solution!

PR #35523, opened 2025-09-27
Fix #35521 - properly display embedded images in mailings

### LINKED ISSUE #35521: Embedded images in mailings not showing properly
### Bug

Embedded images in Dolibarr mailings were showing as placeholders (X icons) instead of displaying properly. Additionally,The issue was caused by two problems in core/class/CMailFile.class.php:

Images with CIDs were being added to emails twice - once as attachments and once as inline images
Images with type cidfromdata (base64 embedded images) were not being properly added to the $this->images_encoded array for inline display these images appeared as email attachments after sending test emails.

### Dolibarr Version

20.0

priojk:
Embedded images display properly in emails instead of showing as placeholders
Embedded images no longer appear as email attachments
The email MIME structure is correctly formed with inline images in the multipart/related section


priojk:
Probably I should have explained the purpose of this proposed pull request better:

For e-mail campaigns/mass mailings, images can  get placed in the editor like so, where images can get placed via drag and drop:

image

In the current implementation, the recipient receives an e-mail with a body where the image is shown as a broken link and the embedded image is instead sent as an attachment. This is due to problems with the proper encoding of the body contents.

The fix takes care of this, the recipient will see an email with the correctly embedded image and no additional attachment.

eldy:
This looks strange as i don't have trouble on v22 nor develop.
The image is correctly inserted into the core of email.

Is it a backport of v22 ?

priojk:
Thanks for looking into this! Previously I did not check with v22 but just with v20 where I fixed the problem. But I re-checked with v22 right now and the problem seems to be there as well:

In v22 I created this in the mass mailings editor, where I placed the image file via drag & drop:

image

-------------------------------------------------------------------


When this gets sent out, the body of the received E-Mail looks like below. As you can see, the embedded image is showing as a broken link and the image itself is delivered as an attachment. The purpose of my fix is to deliver the image as part of the body contents as a user would expect it.


image

eldy:
Which sending method do you use ?   phpmail, smtps, ... ?
Which email client do you use ?

priojk:
In v22 I used SMTP for sending, in v20 phpmail, email client is Mozilla Thunderbird.

eldy:
I confirm the trouble : It happens when reading emails with thunderbird client (all Dolibarr versions with mode php "mail" and "smtps"). 
No trouble with Gmail client.

htdocs/core/class/CMailFile.class.php
```diff
@@ -585,7 +586,9 @@ public function __construct($subject, $to, $from, $msg, $filename_list = array()
 			if (!empty($this->atleastonefile)) {
 				foreach ($filename_list as $i => $val) {
 					$content = file_get_contents($filename_list[$i]);
-					$smtps->setAttachment($content, $mimefilename_list[$i], $mimetype_list[$i], (empty($cid_list[$i]) ? '' : $cid_list[$i]));
+					if (empty($cid_list[$i])) {
```
eldy, line 589: @priojk Can you look at this line i have added to your path.
Your patch fix the trouble of inline image no shown on thunderbird, however, when using smtps mode, the inline files was also added as an attachment visible both in gmail and thunderbird). 
With this line, it seems completely ok (image is visible inline only on gmail and thunderbird). Is it ok for you ?

The line does not have effect of method 'mail',

priojk: Your proposed change is ok for me as well - seems like the issue is fixed. Just tested it in v22.

PR #35549, opened 2025-09-29
add dolBuildUrl with examples of use

frederic34:


hregis:
@frederic34 why don't use camel case? dolBuildUrl()

hregis:
@frederic34 pourquoi tu ne mets pas DOL_URL_ROOT par défaut ?

frederic34:
> @frederic34 pourquoi tu ne mets pas DOL_URL_ROOT par défaut ?

because I use dol_buildpath(xxx, 1) (no more)

frederic34:
@hregis l'autre avantage c'est que si dans un onglet de ton navigateur tu es dans "Adhérents" et que dans un autre onglet tu vas dans "Commerce', quand tu reviens dans le premier onglet et que tu clique ur un lien (généré par dolBuilUrl) tu reste dans "Adhérents"

htdocs/core/lib/functions.lib.php
```diff
@@ -1649,6 +1649,55 @@ function dol_buildpath($path, $type = 0, $returnemptyifnotfound = 0)
...
+	// if (isModEnabled('multicompany') && !isset($params['entity'])) {
+	// 	$params = array_merge($params, ['entity' => $conf->entity]);
+	// }
+	$parameters = [
+		'path' => &$path,
+		'params' => &$params,
+		'addtoken' => &$addtoken,
+	];
+	$hookmanager->executeHooks('buildurl', $parameters);
+	$url = dol_buildpath($path, 1);
```
eldy, line 1688: This has 2 very bad side effects:
- First, the path may not be the one we want because dol_buildpath is searching for alternative path into modules when we want to be sure to use a core file when doing a DOL_URL_ROOT.'/thepath'
This is due to the part of code inside dol_buildpath inside loop 
`foreach ($conf->file->dol_document_root as $key => $dirroot) {`
- The second pb is that we do a file_exist() at every call of dolBuildUrl on a file that often does not eists (so with no cache managed by dolibarr). This is consumming a very high IO.

We can solve both case by replacing this line
`$url = dol_buildpath($path, 1);`
with
```
$tmppath = preg_replace('/^\//', '', $path);
$url = DOL_URL_ROOT.'/'.$tmppath;
```

frederic34: to avoid multiple parameters, I keep the original DOL_URL_ROOT or dol_builpath if present

PR #35555, opened 2025-09-29
NEW add specific parameters in getEntity for "user" element

hregis:
@eldy 

For example, will allow this to be replaced :

```
if (isModEnabled('multicompany') && getDolGlobalInt('MULTICOMPANY_TRANSVERSE_MODE')) {

    $sql .= " WHERE u.rowid IN (SELECT ug.fk_user FROM ".$this->db->prefix()."usergroup_user as ug WHERE ug.entity IN (".getEntity('usergroup')."))";

} else {

    $sql .= " WHERE u.entity IN (" . getEntity('user') . ")";

}
```

with this :

`$sql .= " WHERE u.entity IN (" . getEntity('user', 2) . ")";`


frederic34:
@hregis `int<min,max>` so `int<0,2>` to accept 0, 1 and 2

hregis:
@frederic34 ok thanks

Issue #25361, opened 2023-07-14
Default customer payment terms not taken when creating an invoice


mokacit:
### Bug

The default payment terms, method and Banck Account values set for the customer are not taken when I create an invoice and select the customer from the dropdown list.

### Environment Version

17.0.2

### Environment OS

Apache

### Environment PHP

8.1.21

### Environment Database

MySQL or MariaDB 10.5.19-MariaDB-0+deb11u2

### Environment URL(s)

/compta/facture/card.php?action=create&leftmenu=

### Expected and actual behavior

Payment Terms, Payment Method and Bank Account should be set in the form according to the values specified in the Customer card.

### Steps to reproduce the behavior

Create a customer and set the Payment Terms, Payment Method and Bank Account
Click on New Invoce within the menu "Billing | Payment menu/Customer invoices/"
Select a third party from the drop down menu.

### Attached files

Default values are taken if I comment the lines 3221 and 3222 of /compta/facture/card.php and add the lines of code taken from a previous version of dolibarr as follows:

					// jQuery("#formtocreateaction").val("create");
					// jQuery("#formtocreate").submit();
					var socid = $(this).val();
			                var fac_rec = $(\'#fac_rec\').val();
        			        window.location.href = "'.$_SERVER["PHP_SELF"].'?action=create&socid="+socid+"&fac_rec="+fac_rec;

mokacit:
Thanks, it works.
Perhaps the same approach could be used for 
/comm/propal/card.php
/commande/card.php
/fourn/commande/card.php
all of which suffer from the same problem that is reported for the creation of a new invoice.

mokacit:
The proposed fix introduces an issue when creating the invoice from an order. In this case, the payment fields defined in the order are not taken in the invoice.

eldy:
Seems fixed in v18 for proposal/card.php
Must align invoice in v18 with this.
Rules acted:
If a value were already set BEFORE selecting the thirdparty, the value must be replace by the one of the thirdparty. If no value is set on thirdparty, it must remains as it was set.

PR #35558, opened 2025-09-30
Improve and fix behavior using RELOAD_PAGE_ON_CUSTOMER_CHANGE

SylvainLegrand:
The bank account isn't refresh (fix from v19)


rycks:
@SylvainLegrand here it the full commit where that part of code is concerned

https://github.com/Dolibarr/dolibarr/commit/40fb7c5fa7ce338511b9eb79907d61401616f85b

but your "backport" is not exactly the same ... so i'm in trouble ...

and https://github.com/Dolibarr/dolibarr/issues/25361

eldy:
> @SylvainLegrand here it the full commit where that part of code is concerned
> 
> [40fb7c5](https://github.com/Dolibarr/dolibarr/commit/40fb7c5fa7ce338511b9eb79907d61401616f85b)
> 
> but your "backport" is not exactly the same ... so i'm in trouble ...
> 
> and #25361

I confirm that the fix is more complex than what we think (as we can see in commit id reported by rycks). We had to introduce a var "changecompany" to correctly fix this point. And if the suggested change fix a trouble in some case, it may also introduce troubles in other case. So i suggest to keep the status quo in v18 to avoid to take risk, above all because this is related to a hidden (so not supported) constant.

SylvainLegrand:
The fix made works for my customers.
@eldy The question is whether it is better to let a function requested by customers malfunction (or produce only a partial result) or whether we push the tests further...

eldy:
> The fix made works for my customers. @eldy The question is whether it is better to let a function requested by customers malfunction (or produce only a partial result) or whether we push the tests further...

As RELOAD_PAGE_ON_CUSTOMER_CHANGE is an experimental feature (no with dev not yet finished and planned to be finished in higher version), user reporting of malfunction is not the point(they should not have to use it, if they do it is at their own risk).  Using experimental not yet complete feature is not part of old version maintenance and fix or finishing the dev of experimental feature is always done in develop branch if not yet completed in a higher version).

See "merged process "(develop or maintenance)" in https://wiki.dolibarr.org/index.php/Category:RoadMap

PR #35559, opened 2025-09-30
Fix from V22

SylvainLegrand:
Using a packaging with a float value less than 1 (0.1, 0.5, etc.) generates a 500 error (division by 0)


htdocs/fourn/class/fournisseur.commande.class.php
```diff
@@ -1976,16 +1976,16 @@ public function addline($desc, $pu_ht, $qty, $txtva, $txlocaltax1 = 0.0, $txloca
...
+					$prod->get_buyprice($fk_prod_fourn_price, (float) $qty, $fk_product, 'none', (empty($this->fk_soc) ? $this->socid : $this->fk_soc));
 
 					if ($qty < $prod->packaging) {
-						$qty = $prod->packaging;
+						$qty = (float) $prod->packaging;
 					} else {
-						if (!empty($prod->packaging) && ($qty % $prod->packaging) > 0) {
-							$coeff = intval($qty / $prod->packaging) + 1;
-							$qty = $prod->packaging * $coeff;
+						if (!empty($prod->packaging) && (fmod((float) $qty, (float) $prod->packaging)  > 0.000001)) {
```
eldy, line 1986: We should not need to use this 0.000001
because we are using float that are already truncated  on the number of digits (by price2num before being inserted into database) so are exact float number.

rycks:
Then @eldy here is the trace of the commit into develop in Aug 15, 2024 by mdeweerd, here is the first commit where 0.000001 is used:

https://github.com/Dolibarr/dolibarr/pull/30557/commits/75ed0d3f877fcba06438c36a8613db79e429b80b


maybe https://www.php.net/manual/en/reserved.constants.php#constant.php-float-epsilon ?

rycks:
Thanks to @lvessiller-opendsi here is a example of the bug (we "hope")

setup of module product/service



then supplier price of product with "conditionnement des quantités" (packaging ? sorry)

image


then it makes a "error 500" on adding a new line on a supplier order

image

in error.log the error 500 is

"divisionbyzeroerror: modulo by zero in .../fournisseur.commande.class.php line 1989"

trace commande/card.php line 605

rycks:
@SylvainLegrand could you please make a step by step use case like we try to do with lionel ? because our test today seems to be outside of the scope ...

lvessiller-opendsi (dismissed):

lvessiller-opendsi:
> @SylvainLegrand could you please make a step by step use case like we try to do with lionel ? because our test today seems to be outside of the scope ...

You can reproduce with this packaging of product : 
image

Then you add this product in a supplier order : 
image

The packagin is computed and updated : 
image

eldy:
i succed in reproducing the bug with lionel steps.
The test on > 0.00001 in code hurts me but it was fixed this way in v22, so we can accept it as it is a backport.

rycks (approved):

lvessiller-opendsi (approved):

PR #35564, opened 2025-09-30
NEW Paymentok validate invoice if not already done

Daoud-mohamed:
This can happen for automatically generated invoices with a free amount

### DIFF
--- a/htdocs/public/payment/paymentok.php
+++ b/htdocs/public/payment/paymentok.php
@@ -1073,6 +1073,18 @@ if ($ispaymentok) {
 				$paiement->ext_payment_site = $service;
 
 				if (!$error) {
+					// Validate invoice if not already validated (this can happen for automatically generated invoices with a free amount)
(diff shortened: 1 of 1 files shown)


htdocs/public/payment/paymentok.php
```diff
@@ -1073,6 +1073,18 @@
...
+						if ($result < 0) {
+							$postactionmessages[] = $object->error;
+							$ispostactionok = -1;
+							$error++;
+						} else {
+							$postactionmessages[] = 'Invoice validated';
+							$ispostactionok = 1;
+						}
+					}
 					$paiement_id = $paiement->create($user, 1); // This include closing invoices and regenerating documents
```
eldy, line 1088: If we got an error at line 1082, i think we must not try to create the payement.
so we should add a test here 
`if (!$error)`

Daoud-mohamed: @eldy Done !

PR #35566, opened 2025-09-30
Use trigger prefix product also for supplier product.

fappels:
Because parent methods have triggers with the parent prefix I also set same prefix for supplier product.
For this I had to rename some triggers.
This was most simple solution I found.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -11,6 +11,7 @@ For users:
 
 For developers:
 ---------------
+* Introducing the TRIGGER_PREFIX property to force developers to use unique triggerkey per business object, limit also code to business CRUD events and identity when it is not.
 
 
 
@@ -28,7 +29,7 @@ The following changes may create regressions for some external modules, but were
   same path than the core path without using a mymodule directory is now forbidden.
 * The directory theme/common/octicons has been removed
 * The library timepicker.js has been removed. Was not used by Dolibarr.
-* Because of new TRIGGER_PREFIX property triggers SUPPLIER_PRODUCT_BUYPRICE_XXX are renamed to PRODUCT_BUYPRICE_XXX
+* Because of new TRIGGER_PREFIX property triggers SUPPLIER_PRODUCT_BUYPRICE_XXX are renamed to PRODUCT_BUYPRICE_XXX.
 
 
 ***** ChangeLog for 22.0.2 compared to 22.0.1 *****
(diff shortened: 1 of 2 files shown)


frederic34:
Do you have an example of something that doesn’t work?

fappels:
@frederic34 When I do this:
class MyProduct extends ProductFournisseur 

$myProduct = new MyProduct($db);

$myProduct->fetch($id);

$myProduct->description = 'new description';

$myProduct->update($user); 
->crash on product class here: https://github.com/Dolibarr/dolibarr/blob/af371b828f297c3cedcbc121451f68866c05e3bc/htdocs/product/class/product.class.php#L2099 caused by this check: https://github.com/Dolibarr/dolibarr/blob/af371b828f297c3cedcbc121451f68866c05e3bc/htdocs/core/class/commontrigger.class.php#L89

Maybe it's the check that needs to be corrected @eldy ?

frederic34:
if you change `$result = $this->call_trigger('PRODUCT_SET_MULTILANGS', $user);` to `$result = $this->call_trigger($this->TRIGGER_PREFIX.'_SET_MULTILANGS', $user);` ?

fappels:
Then we can have both PRODUCT_SET_MULTILANGS and SUPPLIER_PRODUCT_SET_MULTILANGS triggers for the same trigger action -> confusing. Also the PRODUCT_SET_MULTILANGS is only an example there are also PRODUCT_PRICE_MODIFY, PRODUCT_SUBPRODUCT... Even think
 PRODUCT_MODIFY which comes after PRODUCT_SET_MULTILANGS  will crash.

frederic34:
replace https://github.com/Dolibarr/dolibarr/blob/af371b828f297c3cedcbc121451f68866c05e3bc/htdocs/core/class/commontrigger.class.php#L90-L91 by dol_syslog for some versions

fappels:
Yes also think the exit; is too drastic, should be removed for at least Dolibarr V23 and V24 @eldy

eldy:
The goal of introducing the TRIGGER_PREFIX is to start to force dev to use unique triggerkey per business object, and limit also code to business CRUD events, and identity when it is not.
In dolibarr, the product and supplier product is the same object, same table, only some properties differs, so i have no problem to see the same trigger code. Trigger can identify if product is a supplier product or a customer product with property of object. And it can be both so having 1 code only for all products is not so strange.  

So your fix seems ok for me. Does it fix all the trouble you identified ?

fappels:
@eldy  Thanks for explaining TRIGGER_PREFIX, I added part of your explication to developers readme.
My fix is ok for me, have run some unit test on the example I described for @frederic34  and was ok with this fix.

PR #35570, opened 2025-09-30
check if file exists before include

frederic34:


htdocs/bookcal/availabilities_contact.php
```diff
@@ -190,8 +190,12 @@
 	$dirtpls = array_merge($conf->modules_parts['tpl'], array('/core/tpl'));
 	foreach ($dirtpls as $reldir) {
 		$res = @include dol_buildpath($reldir.'/contacts.tpl.php');
-		if ($res) {
-			break;
+		$file = dol_buildpath($reldir.'/contacts.tpl.php');
+		if (file_exists($file)) {
```
eldy, line 194: I think you want to remove also the line 192 becasue now done in line 195 ?

PR #35573, opened 2025-10-01
fix: preload of commande extrafields in delivery note and fix of para…

### LINKED ISSUE #35241: V22.0.1 Extrafield in expedition does not work anymore
### Bug

in V22.0.0, when we create an expedition, we can use this extrafield :
Image

It is setup like this :
Image

After update to V22.0.1 there is a problem with this extrafield ; when I wanted to manage it, blank page appears.

It showing this :
Image

I return to V22.0.0 and it seems that the defaut value is not take into account.

### Dolibarr Version

22.0.1

### Environment PHP

8.2.28

### Environment Database

MariaDB 10.11.6

_No response_

### Attached files

_No response_

splohmer:
# FIX|Fix #35241 load of order extrafields in delivery note broken
preload of commande extrafields in delivery note and fix of parameters for hook formObjectOptions

### DIFF
--- a/htdocs/expedition/card.php
+++ b/htdocs/expedition/card.php
@@ -1206,7 +1206,8 @@ if ($action == 'create') {
 			// $objectsrc is Commande|Facture
 			$objectsav = $object;	// Because Expedition is $expe and not $object that is wrongly a duplicate of $objectsrc.
 			$object = $expe;
-			$parameters = array('objectsrc' => isset($objectsrc) ? $objectsrc : '', 'cols' => '3', 'socid' => $socid);
+			$object->array_options = array_merge($object->array_options, $objectsav->array_options);
+			$parameters = array('objectsrc' => isset($objectsav) ? $objectsav : '', 'cols' => '3', 'socid' => $socid);
 			include DOL_DOCUMENT_ROOT.'/core/tpl/extrafields_add.tpl.php';
 			$object = $objectsav;
 


splohmer:
A bit of explaination:
in the $parameters before it was $objectsrc which is not used in this case but $object to load the order. Also missing was the part of preloading extrafields from order to shipping, which is not part otf the extrafields_add.tpl.php.

hregis:
@splohmer 
If you merge extrafields, you risk having extrafields that do not exist in an expedition.
`$object->array_options = array_merge($object->array_options, $objectsav->array_options);`

hregis:
@splohmer 

$objectsrc is the object source original (not modified) and is used with hook !!!

splohmer:
@hregis Thanks for your feedback. Let's clarify the two points:

1. Merge Extrafields: In previous versions of dolibarr the extrafields where copied from order to shipping before. This is currently not working. In Version 21 the array_merge mechanism was used as well. Additionally the call of $object->showOptional function in extrafields_add.tpl.php makes sure that only the ones are shown which are also in the shipping object. Do you have a better approach?
2. if I got $objectsrc correctly it is the original object the shipping is created from, right? In this flow the $objectsrc is just null  as the corresponding originate object (order) is loaded into $object in line 1028.

Regards,
Sven

hregis:
@splohmer Yes, you're right... we should revise this page in "develop" to replace "$exp" with "$object" so that there is no more confusion. Thanks

hregis (approved):

eldy:
I tried to create an extrafield on the shipment like explaine on issue #35241
I do not experience any trouble with v22.0.2
When i create a shipment, the extrafield is visible
image

If the goal of the PR is to fill the value with the value of the extrafields that has the same name than one on the origin sale order, then fix should contains a loop like this (i did not test)
```
			// Propagate extrafieldsvalue from source object to shipment object
			foreach ($extrafields->attributes[$object->table_element]['label'] as $key => $val) {
				if (array_key_exists('options_'.$key, $objectsav->array_options)) {  // We take value from order only if extrafield has the same name/key.
					$object->array_options['options_'.$key] = $objectsav->array_options['options_'.$key];
				}
			}
```

I pushed a better commit.

PR #35591, opened 2025-10-02
QUAL Remove duplicate code block

thomas-Ngr:
# QUAL Ticket creation trigger : remove duplicate code block to send twice the email to assignee

The right code block is just above the one deleted (line 178-199)


thomas-Ngr:
Php-stan fails on a file that is not changed.

PR #35592, opened 2025-10-02
Update html.formmail.class.php

vanyolai:
There was an error in email sending.

The php function was inside the SQL string.


eldy:
Fixed by another fix b1d57a6bd1a1c5d86cfc6ca33ec0c13c1481cd53

PR #35601, opened 2025-10-02
Enable filtering by supplier order in `/receptions` API endpoint

FabianVolkers:
# NEW Enable filtering by supplier order in `/receptions` API endpoint
Add a `LEFT JOIN` for table `llx_element_element` in the index API function to enable sqlfilters on `el.fk_origin`. This enables users to filter receptions by supplier_order id.


htdocs/reception/class/api_receptions.class.php
```diff
@@ -123,6 +123,7 @@ public function index($sortfield = "t.rowid", $sortorder = 'ASC', $limit = 100,
 		$sql = "SELECT t.rowid";
 		$sql .= " FROM ".MAIN_DB_PREFIX."reception AS t";
 		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."reception_extrafields AS ef ON (ef.fk_object = t.rowid)"; // Modification VMR Global Solutions to include extrafields as search parameters in the API GET call, so we will be able to filter on extrafields
+		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."element_element as el ON el.fk_target = t.rowid AND el.targettype = 'reception' AND el.sourcetype = 'order_supplier'";
```
eldy, line 126: If not filter (el.fk_origin:=:123) is provided, this will generate a duplication of lines
So this must be conditionned on having a filter on el field.

FabianVolkers:
Thanks for your review. I could not replicate the issue but have added a condition to check for a filter on any of the el columns.

The only way this would break is if someone used a filter like this `(t.desc:=:'hello (el.fk_source:!=:1)')`. Seems unlikely but should be fixed. I will look into this again the following days. Or do you know of a better way to check for a filter on the el columns?

eldy:
> Thanks for your review. I could not replicate the issue but have added a condition to check for a filter on any of the el columns.
> 
> The only way this would break is if someone used a filter like this `(t.desc:=:'hello (el.fk_source:!=:1)')`. Seems unlikely but should be fixed. I will look into this again the following days. Or do you know of a better way to check for a filter on the el columns?

The link with the sub table llx_element_element is done on the 3 fields
`"el.fk_target = t.rowid AND el.targettype = 'reception' AND el.sourcetype = 'order_supplier"`
And to avoid to generate duplicate lines when doing the request, a link on a sub table must be done on a unique key. The unique key on table ll_element_element is on the 4 columns fk_target, targettype, sourcetype and fk_source.
So the only field we miss is the fk_source.
This means the only filter that can be accepted to add a link with the subtable ll_element_element is a filter 
el.fk_source=xxx
So i made your test to detect if we must add the subtable llx_element_element or not more simple.

@FabianVolkers 
Thanks to check that code is still working after my change (we can still hack using the filter into the desc but this is not a common case and if dev ask strange things, he will get strange this, this does not hurt me).

FabianVolkers:
Thanks for your elaboration and code change. Your change looks good to me. Filtering by supplier order id still works as expected. If you say it's fine that the edge case is not being caught, I say we can merge.

PR #35604, opened 2025-10-02
NEW Create simple shipment of non origin

### LINKED ISSUE #22943: Standalone "Shipments" to create quick shipment sheets not linked to any order

sonikf:
Step 1
New functions and minor changes to allow standalone shipment while retaining existing functionality

related to https://github.com/Dolibarr/dolibarr/issues/31652
fix https://github.com/Dolibarr/dolibarr/issues/22943


htdocs/expedition/class/expedition.class.php
```diff
@@ -648,6 +649,17 @@ public function create($user, $notrigger = 0)
 					}
 				}
 
+				// update ref
+				$initialref = '(PROV'.$this->id.')';
+				if (!empty($this->ref)) {
+					$initialref = $this->ref;
+				}
+
+				$sql = 'UPDATE '.MAIN_DB_PREFIX.$this->table_element." SET ref='".$this->db->escape($initialref)."' WHERE rowid=".((int) $this->id);
```
eldy, line 658: Why not putting this (using $this->ref instead of (PROVx) directly at line 497 (to avoid 2 updates ?

edupulpillo:
How to access to this feature?
In 23.0.3 still there is no option in the menu to create a shipment without a Sales order associated.

sonikf:
Hi @edupulpillo 
Add in Home->Setup->Other Setup   SHIPMENT_STANDALONE 1
It is documented in the wiki https://wiki.dolibarr.org/index.php/Setup_Other#:~:text=top%29-,Shipments

edupulpillo:
Thank you !

PR #35612, opened 2025-10-03
NEW: import subscriptions

altairis-tof:
add import option for subscriptions


htdocs/adherents/class/adherent.class.php
```diff
@@ -314,6 +314,11 @@ class Adherent extends CommonObject
 	 */
 	public $invoice;
 
+	/**
+	 * @var string 	Import ID
+	 */
+	public $import_key;
```
eldy, line 320: Declare property with 
* @var ?string Import ID
to have it null possible.

But best is to remove declaration as it is already declared into CommonObject.class.php

altairis-noe: Good thinking, I removed it!

PR #35617, opened 2025-10-03
new: add Contact sign role

### LINKED ISSUE #35613: Missing contact client signataire on proposal, order
### Feature Request

Image

### Use case

_No response_

### Suggested implementation

_No response_

### Suggested steps

_No response_

FHenry:
fix: https://github.com/Dolibarr/dolibarr/issues/35613


htdocs/install/mysql/data/llx_c_type_contact.sql
```diff
@@ -48,6 +48,7 @@ insert into llx_c_type_contact (element, source, code, libelle, active ) values
 insert into llx_c_type_contact (element, source, code, libelle, active ) values ('propal',  'external', 'BILLING',       'Contact client facturation propale', 1);
 insert into llx_c_type_contact (element, source, code, libelle, active ) values ('propal',  'external', 'CUSTOMER',      'Contact client suivi propale', 1);
 insert into llx_c_type_contact (element, source, code, libelle, active ) values ('propal',  'external', 'SHIPPING',      'Contact client livraison propale', 1);
+insert into llx_c_type_contact (element, source, code, libelle, active ) values ('propal', 'external',  'SALESREPSIGN',   'Contact client signataire propale', 1);
```
eldy, line 51: The customer contact that is responsible to accept or not (so "sign" the proposal) is already the entry 
'propal', 'external', 'CUSTOMER'
We may want to make a difference between the guy that accept/refuse and the guy that sign but this looksmore an advanced use case (in most case, it will be the same).
So i suggest to add this new type of contact ("signature only") as not active by default and rename it into "Contact client signataire obligatoire propale" . The company that need a different explicit signer will be able to enable it from dictionnary.

Issue #33741, opened 2025-04-06
Mass action "Reverse stock movements" has no effect and seems incomplete


priojk:
### Bug

In the list of stock movements, I can check checkboxes of movement lines and than select a mass action to revert the selected movements.

What happens is a change in the movement code and the title of the stock movement to describe it as reverted/annulated - but nothing else.

I would have expected a similar behavior to canceling an accounting line: creation of an additional line with the same quantity but opposite sign (+/-) or at least a change of the quantity in the reverted line to 0.

Currently, there is no effect on the stock level of a product, when I revert the movement.

### Dolibarr Version

v20, v21

### Steps to reproduce the behavior and expected behavior

- when you have a product with a stock movement, check the stock level
- got to list of stock movements, select the movement of the product, select mass action "revert" and confirm
- check again stock level of product: no change although the movement got reverted

bradley-jarvis:
I can confirm this,  the issue appears to be in product/stock/class/mouvementstock.class.php in the function reverseMouvement which only updates the label and code, it should also update the product_stock table as it does in _create. There have been a few instances where this would have been useful. Looking at the latest version shows that this has not been addressed, I am going to have a go at fixing tomorrow

PR #35632, opened 2025-10-04
Fix 33741 Mass action "Reverse stock movements" has no effect and see…

vmaury:
…ms incomplete

I didn't reverse take product's batch in account as I dont know if it's usefull in this case ... and because I don't know exactly where batches are usefull

### DIFF
--- a/htdocs/product/stock/class/mouvementstock.class.php
+++ b/htdocs/product/stock/class/mouvementstock.class.php
@@ -1353,7 +1353,8 @@ class MouvementStock extends CommonObject
 			dol_print_error(null, "Failed to fetch product");
 			return -1;
 		}
-
+		$oldqtywarehouse = 0;
+		$fk_product_stock = 0;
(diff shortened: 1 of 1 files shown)


eldy:

vmaury:
Sorry I didn't put the link to the issue https://github.com/Dolibarr/dolibarr/issues/33741  
Before this fix, when you make a reverse stock mouvement (in mass actions), it only affects the label of the movement, but it does nothing

eldy:
> Sorry I didn't put the link to the issue #33741 Before this fix, when you make a reverse stock mouvement (in mass actions), it only affects the label of the movement, but it does nothing

Thanks for providing information.
I pushed a better fix for the following reason
- Any change in stock table must be done using he low level method _create() in stock movement class (direct update in product_stock is not allowed)
- Missing db transaction
- Label of movement must be translated
- Code must manage correctly batch number
- We must be able to revert a revert.

PR #35643, opened 2025-10-05
Adding hurl tests to Dolibarr

JonBendtsen:
# NEW Adding hurl tests to Dolibarr
This PR adds hurl tests to Dolibarr. Hurl is an Open Source CLI tool that can make HTTP requests, and thus be used to make automatic tests of the API, GUI as well as public/ http endpoints/pages of Dolibarr.

The goal is to make it easy to automatically detect unintended consequences of changes and additions to Dolibarr.


JonBendtsen:
@eldy Here's the PR with test/hurl/ - yes it doesn't build currently, but that does not seem to be due to any of my files

PR #35668, opened 2025-10-06
Refactor contact deletion logic in API proposals

defrance:
the main problem is the list_contact function is call without parameters, so we return only external contact
so we can't delete link of internal user

this is the same thing on deletecontact function of invoice and order and supplier order

I precise that deletecontact api call is not present on many element


defrance (commented):
suppress blank caracters

PR #35684, opened 2025-10-08
FIX Module-builder modMyModule.class.php

kuscoo:
FIX an error that generates errors in the logs.
dol_eval try/catch error : Undefined constant "mymodule"


kuscoo (commented):
Update following @eldy's comment

PR #35686, opened 2025-10-08
Fix accountancy pdfs

sonikf:
- Added missing translation strings

- Added new translation for C:Credit and D:Debit

- Fixed translation for non latin languages

- Added missing separator lines


sonikf:
@eldy, @aspangaro 
I think pdfs would look better in landscape mode(more space).
What do you think?

sonikf:
@eldy we need a space because this does not look good



we need this 




Can i try this instead?

`$langs->transnoentitiesnoconv('Total') . ' ' . sprintf('%s %s', $langs->trans('AccountancyGroup'), $accountingAccount->pcg_type),`

eldy:
The key 
AccountancyGroupXXX is as a parameter of trans is used to get the translation key AccountancyGroupXXX  (without pace).
If you add the key AccountancyGroup, you should manage the 2 cases
First test if translation AccountancyGroupXXX exists ( 'AccountancyGroupXXX' != $langs->trans("AccountancyGroupXXX")
And if not, you can output $langs->trans("AccountancyGroup").' '.XXX   (but the space must be outside of the trans.

PR #35688, opened 2025-10-08
NEW SQL structure for reception lines

sonikf:


eldy:
On shipment, the line for reception and the detail for each lot/serial number was savec into 2 different tables.
But is was not a good practice and for reception, we have fixed this to use just one table.
The table llx_receptiondet_batch is same than llx_receptiondet but with another optional field that is the batch number if we want to use it.
Why do you need a new structure ?

sonikf:
Hi @eldy
I didn't know that and just copied expedition structure. I started working on standalone reception too!
Fields are missing from llx_receptiondet_batch though so i will make a new pr.
Big problem for both modules in their current state is many common use cases are not supported.

PR #35690, opened 2025-10-09
Fix : Display supplier prices for shared products in multicompany context #35600

atm-corentin:
# FIX|Fix #*Display supplier prices for shared products in multicompany context*

@eldy  , It is a backport from https://github.com/Dolibarr/dolibarr/pull/35600 

In a multi-company setup, the buying prices for shared products do not appear on the supplier's product list. The database query is too restrictive and fails to link a shared product to its price when they are located in different entities.

Solution
The solution is to make the database query more flexible. The query is modified to search for the price across all entities the user can access, instead of being limited to the product's original entity. This allows the shared product's price to be linked and displayed correctly.


hregis:
@atm-corentin can you change with :

getEntity('productsupplierprice')

atm-corentin:
> @atm-corentin can you change with :
> 
> getEntity('productsupplierprice')

i applied the suggestion

PR #35709, opened 2025-10-10
NEW #35700 : Throw an error when validating a propal, order, supplier with a product no more in sale/purchase…

### LINKED ISSUE #35700: Valid a propal with products no more for salel is possible
### Bug

Imagine the following use case (that occurs to one of my client, I took some time to understand the scenario)
- somebody makes a propal (or order) with products : these are obliged  to be for sale to be selected.
The propal stays draft during some time for some reason
During this time, one of the products listed in the propal is set to "not for sale"
We return to the propal/order and valid it, no problem ... but in fact we've valid an object containing elements that are no more for sale, which seems to be abnormal
Same scenario with supplier 
I suggest (I'll make a PR) to add code to verify theses things before validation

### Dolibarr Version

all

### Environment PHP

_No response_

### Environment Database

_No response_

_No response_

### Attached files

_No response_

vmaury:
Fix #35700 Valid a propal with products no more for sale is possible


hregis:
@vmaury product and service 

`if (isModEnabled('product') || isModEnabled('service')) {`

hregis:
@vmaury if you have "$this->lines" reload the fetch is useless because it is already done before the action is called (at the top of the page or before calling the "valid()" function)

hregis:
@vmaury the check is ok during the "valid" action, but after, during the "closeas" action ?

hregis:
@vmaury and what happens if the product is no longer for sale but there was 1 left at the time of creating the commercial proposal or order?

hregis:
@vmaury for example if a function or a module makes the product "out of sale" when we add the last product in stock, when we validate we will get an error?

vmaury:
@hregis 
Ok for 
if (isModEnabled('product') || isModEnabled('service')) {

For 
 _if you have "$this->lines" reload the fetch is useless because it is already done before the action is called (at the top of the page or before calling the "valid()" function)_
I don't agree.
I don't see somewhere where $line->status or $line->status_buy is set
And I made some debug on an order->valid or supplier_proposal->valid(), I confirm $line->status is not fetched.
So we're obliged to fetch the product

hregis:
@vmaury yes you are right, sorry I confused it with the lines of the object

hregis:
@vmaury you can move this lines outside the loop

```
include_once DOL_DOCUMENT_ROOT.'/product/class/product.class.php';
$langs->load('products');
```

vmaury:
> @vmaury you can move this lines outside the loop
> 
> ```
> include_once DOL_DOCUMENT_ROOT.'/product/class/product.class.php';
> $langs->load('products');
> ```

I put that in the loop to minimize the memory used, so that if there is no fk_product in no line, we make no product include et no lang load.
But If you think it's better to put the lines outside the loop for "elegancy" no problem :-)

hregis:
@vmaury This is to avoid loading these files as many times as there are lines.

hregis:
@eldy @vmaury it's not better to invert the constant to avoid side effect? 

`XXXXX_CHECK_ONSALE_PRODUCTS_ONVALID`
instead of
`XXXXX_NOCHECK_ONSALE_PRODUCTS_ONVALID`

eldy:
> @eldy @vmaury it's not better to invert the constant to avoid side effect?
> 
> `XXXXX_CHECK_ONSALE_PRODUCTS_ONVALID` instead of `XXXXX_NOCHECK_ONSALE_PRODUCTS_ONVALID`

May be, but long term target should be to no allow it by default, so, as we are in a develop branch, we can take the risk.

hregis:
@eldy It's a bit nonsensical! :-)
You agree to it, knowing it could cause problems!
Laurent!! Are you being shameless?! 😄

eldy:
> It's a bit nonsensical! :-)

Any change is a risk. Even the one that seems not risky. So we must not simply think "there is a risk so PR is not accepted", or "there is no risk, we can accept PR".
But we must think of the Ratio :     (Benefit or need) / (Probability of risk * Gravity if risk appear)
Here I think  B / (P * G)  is high enough (Probability is low, Gravity is very low and Benefit is medium as it avoid to record data we should not), so we should go forward with this PR.

PR #35714, opened 2025-10-11
New module_part for controllers to allow modules to define and overri…

bafbes:
New module_part for controllers to allow modules to define and override predefined controllers


bafbes:
solved ci errors

bafbes:
solved conflicts

PR #35725, opened 2025-10-12
Add shipping own contacts type configuration option

defrance:
# Instructions
*Please:*
- ***in particular, in case of a bugfix, please check that you are targetting the branch corresponding to the oldest version in which the bug occurs***

[*Long description*]

[*Long description*]

[*Long description*]

[*Long description*]

[*Long description*]

[*Long description*]

### DIFF
--- a/htdocs/admin/expedition.php
+++ b/htdocs/admin/expedition.php
@@ -490,6 +490,14 @@ print '</td><td>';
 print '<input class="flat minwidth200" type="text" name="SHIPPING_DRAFT_WATERMARK" value="'.dol_escape_htmltag(getDolGlobalString('SHIPPING_DRAFT_WATERMARK')).'">';
 print "</td></tr>\n";
 
+// shipping use its own contacts
+print '<tr class="oddeven">';
+print '<td>'.$langs->trans("ShippingUseItsOwnContacts");
+print '</td>';
+print '<td>';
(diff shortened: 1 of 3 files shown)


eldy:
I don t understand the goal of the activated hidden option. What means its own type of contact ?
Can you explain it with/without option.

defrance:
When this setting is not enabled, shipments use the contact types from the orders.

When the setting is enabled, shipments use their own defined contacts.

Personally, I think this latter mode should be the default, but it would significantly change functionality for existing users.

eldy:
It seems shipping already used their own contact. They have their own tab "Contact/address" like any other object.
image

So this option is still not clear of what it does.

defrance:
I create an order with one associated contact.
I create a shipment from this order, it inherits the contact from the order.
I add the second contact to the shipment; it is also added to the order.
Once the setting is enabled, the shipment has its own contacts, independent of the order.
Here is a series of screenshots illustrating this, both with and without the setting in question.

Capture d’écran 2026-09-23 à 18 21 58
Capture d’écran 2026-09-23 à 18 22 10
Capture d’écran 2026-09-23 à 18 22 20
Capture d’écran 2026-09-23 à 18 22 26
Capture d’écran 2026-09-23 à 18 23 43

PR #35726, opened 2025-10-12
Modify getListOfContactTypes for expedition type

defrance:
Adjust the contact type based on shipping settings.

The contact types for shipments are based on those for orders unless the new setting is enabled since V22


hregis:
@defrance d’où vient cette manie d'inverser l’existent au risque de créer des effets de bords ?

eldy:
> @defrance d’où vient cette manie d'inverser l’existent au risque de créer des effets de bords ?

The management of contact on shipment is not consistent with the rest of other object. So introducing code with a behaviour is not a bad thing. Also the risk to break something is under control as the new behaviour is conditionned by a hidden option. So we can "test" and "wait" everything is ok to decide later if we must switch or not to standard behaviour.

PR #35738, opened 2025-10-13
FIX: Email template fetching

noec764:
When ref was not given in this fetch function, fetch sql request was:
 ```SQL
SELECT t.rowid,t.module,t.type_template,t.lang,t.private,t.fk_user,t.datec,t.tms,t.label,t.position,t.active,t.topic,t.content,t.content_lines,t.enabled,t.joinfiles,t.email_from,t.email_to,t.email_tocc,t.email_tobcc,t.defaultfortype 
FROM llx_c_email_templates as t 
WHERE t.rowid = XXX AND t.label = '' ;
```
The label condition was therefore blocking


hregis:
@noec764 

`$result = $this->fetchCommon($id, '', (empty($ref) ? '' : " AND t.label = '".$this->db->escape($ref)."'"), $noextrafields);`

PR #35747, opened 2025-10-13
QUAL: PHPstan Variable $compatible might not be defined.

JonBendtsen:
# qual PHPstan Variable $compatible might not be defined.
This PR is trying to get the build system back in working order

### DIFF
--- a/htdocs/admin/remotestore/class/externalModules.class.php
+++ b/htdocs/admin/remotestore/class/externalModules.class.php
@@ -500,6 +500,8 @@ class ExternalModules
 					).'</span>';
 					$compatible = 'NotCompatible';
 				}
+				// Output the line
+				$html .= '<tr class="app oddeven nohover '.dol_escape_htmltag($compatible).'">';
 			} else {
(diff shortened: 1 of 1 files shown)


frederic34:
also in branch 22 https://github.com/Dolibarr/dolibarr/pull/35737

JonBendtsen:
> #35737

@frederic34  what if the $compatible variable was defined inside the if, not before? then you would only have 1 assignment.

Also, I am not sure I like the value '' because that value is also used when it is actually compatible, then I'd rather have something else, perhaps 'NotYetAvailable' ?

JonBendtsen:
> also in branch 22 #35737

@frederic34  if it is fixed in v22, then it would also automatically be forward into development?

(if you had written a longer title than FIX CI then I might have found it, but at least now I will search for CI as well)

frederic34:
> 
> @frederic34 if it is fixed in v22, then it would also automatically be forward into development?

yes

> 
> (if you had written a longer title than FIX CI then I might have found it, but at least now I will search for CI as well)

sorry, but every day the builds are broken by direct push...

JonBendtsen:
duplicate of @frederic34's work

PR #35750, opened 2025-10-13
FIX #35318 better array check fields_label

### LINKED ISSUE #35318: Better array check in commonobject.class.php / extrafields.class.php
### Bug

I had problems with extrafields.class.php , it gave me an SQL error  because of added ', ' to SQL statement 
line1523 & 2192
			$fields_label = explode('|', $InfoFieldList[1]);
			if (is_array($fields_label)) {
				$keyList .= ', ';
				$keyList .= implode(', ', $fields_label);
			}

It will always be true even if the array is empty !!? And add a ', ' to SQL statement. 
A better solution could be:
			isset($InfoFieldList[1]) ? $fields_label = explode('|', $InfoFieldList[1]) : $fields_label=array() ; 
			if (!empty($fields_label)) {
				$keyList .= ', ';
				$keyList .= implode(', ', $fields_label);
			}

And I see that core/ajax/ajaxextrafield.php also have the same setup, but with a count() it will always be true because there will be one empty array !!

Also core/class/commonobject.class.php  have it

### Dolibarr Version

23

### Environment PHP

8.4

### Environment Database

mariadb 12.0.2

_No response_

### Attached files

_No response_

JonBendtsen:
# FIX #35318 better array check in categories, extrafields and commonobjects
This PR is an attempt to fix the bug reported in #35318 with some better checking that hopefully does not result in SQL errors.

### DIFF
--- a/htdocs/core/ajax/ajaxextrafield.php
+++ b/htdocs/core/ajax/ajaxextrafield.php
@@ -180,8 +180,8 @@ if ($object instanceof CommonObject) {
 		}
 
 		if (!$filter_categorie) {
-			$fields_label = explode('|', $InfoFieldList[1]);
-			if (count($fields_label) > 0) {
+			isset($InfoFieldList[1]) ? $fields_label = explode('|', $InfoFieldList[1]) : $fields_label=array();
+			if (!empty($fields_label)) {
 				$keyList .= ', ';
 				$keyList .= implode(', ', $fields_label);
 			}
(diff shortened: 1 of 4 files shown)


hregis:
@JonBendtsen 

when you feel that what will be the most common you have to put forward what will be the fastest

$fields_label = (!isset($InfoFieldList[1]) ? array() : explode('|', $InfoFieldList[1]));

JonBendtsen:
> @JonBendtsen
> 
> when you feel that what will be the most common you have to put forward what will be the fastest
> 
> $fields_label = (!isset($InfoFieldList[1]) ? array() : explode('|', $InfoFieldList[1]));

Good point. I do not feel what will be most common, I don't have that knowledge yet

PR #35785, opened 2025-10-16
FIX: Performance Problem on load stats command

noec764:
In the Command List, when shippable columns were selected, a timeout occurred. I found out that OR condition in query was leading to a bad index use. Instead of that, I found that having Two queries with union fix this problem.


hregis:
@noec764 @eldy je vais revoir ceci car non compatible avec Multicompany

PR #35788, opened 2025-10-16
FIX : remove useless condition to create credit on situation invoice …

### LINKED ISSUE #35786: Situation invoice : Condition to allow creation of credit invoices is wrong
### Bug

On situation invoices, user should be able to create a credit note if excess payment has been received. (as stated by a comment in the code)
- this is not the case : on the contrary, creation of credit notes is available if invoice has not been paid entirely.
- I don't see why this restriction is set (a customer of ours creates credits on situations not totally paid, using the current behavior.)

## Details

On compta/facture/card.php, we have this condition : 

```php
// For situation invoice with excess received
if ($object->status > Facture::STATUS_DRAFT
	&& $object->isSituationInvoice()
	&& ($object->total_ttc - $totalpaid - $totalcreditnotes - $totaldeposits) > 0
	// ...
) {
    // Display button to create credit invoice
}
```

I see two issues : 

### Condition is wrong

`($object->total_ttc - $totalpaid - $totalcreditnotes - $totaldeposits) > 0` will resolve to `false` if excess payment has been received. The comparison operator should be `<` instead of `>`.

### Condition is useless

I don't see why a credit invoice should be available on situation invoices only when excess payment has been received.

I suppose that this has not been detected since its implementation in DLB8.0 precisely because it gives user the ability to create credit invoices on situation invoices before it has been totally paid (or paid in excess).

Moreover, since V19.0, another buttons appears to "convert excess received to avalable credit".

## Suggested fix : 

- change comment ` For situation invoice with excess received` to ` For situation invoice`
- remove the wrong condition

### Dolibarr Version

8.0 to develop

### Environment PHP

any

### Environment Database

any

- prepare
  - create and validate an invoice (ex. total TTC = 50)
  - create a fix discount for the customer (ex. 10)
  - affect the fix discount to the invoice -> remain to pay = 40
- not completely paid : you see the button
  - make a payment lower than the invoice total (ex. 10)
  - you have the button "create credit invoice" available
- completely paid : no button
  - make a payment that fills the invoice total (ex. 30)
  - you don't have the button "create credit invoice" available
- excess payment : no button
  - delete the last payment
  - make a payment that  creates an excess payment (ex. 40)
  - you don't have the button "create credit invoice" available but (since V19) you have a new button "Convert excess to available credit"

### Attached files

_No response_

thomas-Ngr:
# FIX #35786 : wrong condition to allow the creation of credit invoices for situation invoices

The condition to allow creation of credit invoice only if excess payment was received is wrong, and (to my mind) useless.

I suggest to remove it.

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -5796,10 +5796,9 @@ if ($action == 'create') {
 				}
 			}
 
-			// For situation invoice with excess received
+			// For situation invoice
 			if ($object->statut > Facture::STATUS_DRAFT
 				&& $object->type == Facture::TYPE_SITUATION
-				&& ($object->total_ttc - $totalpaid - $totalcreditnotes - $totaldeposits) > 0
 				&& $usercancreate
 				&& !$objectidnext
 				&& $object->is_last_in_cycle()


rycks:
on a full paid cycle thanks to your patch we can make credit invoice:

image

@thomas-Ngr but with your patch there is a race condition if customer has made over payment there is now two buttons who (i think) made the same thing:

image

see line 5733 ...

thomas-Ngr:
Hi !

> with your patch there is a race condition if customer has made over payment there is now two buttons who (i think) made the same thing

These two buttons don't do the same thing. 

"Convertir le trop-perçu en crédit disponible" will convert **overpayment** to **available credit**. In your example, it will be an available credit of 100.
"Créer facture avoir" will only create a credit invoice. In your example it would **create a credit invoice** of 8000 HT (or 1600, I don't understand if your total is 1600 or 8000).

"Créer facture avoir" will allow to create a credit invoice for the last invoice in cycle, convert it to available credit, and then remove the last invoice from cycle. "Convertir le trop-perçu en crédit disponible" would not allow this.

I would understand that you tell me "this is a change of behavior, it should go in develop". I think it is a bug, but making a button appear seems like a new feature.

rycks:
for sure you are right ! people can be in the situation of overpayment and they want to do an "avoir" ... your are right !

rycks (approved):

lvessiller-opendsi (approved):

hregis:
@rycks faire un merge sans avoir l'aval du grand chef c'est risqué ! 😄

hregis:
@rycks @lvessiller-opendsi et c'est quand la prochaine version LTS ? parce que bon la v18... on attend la v30 ? 😄

hregis:
@rycks @lvessiller-opendsi sinon on ne devrait plus sortir de nouvelles versions et rester sur une v18 !

v18.0.15 
v18.5.32
v18.150.45

PR #35792, opened 2025-10-16
FIX : Loss of line breaks in the private and public note of a busines…

NasserNgandu:
Go to a business proposal

Edit the private or public note

Type a sentence, then press Enter to go to the next line

Save

Edit the note again by adding one or two lines below

Save

Note that line breaks are no longer respected

Here is the link to the issue: https://github.com/Dolibarr/dolibarr/issues/34940

Here is the PR for version 22: https://github.com/Dolibarr/dolibarr/pull/35792


hregis:
@NasserNgandu can you test your fix when the wysiwyg is activated at public/private note ?

hregis:
@NasserNgandu after test it's ok for me, can you propose this PR for 22 branch instead develop ?

hregis:
@NasserNgandu @eldy 
for me the "`<br>`" should not be visible in edit mode when not using WYSIWYG. (that's what we had before!)

Capture d’écran 2025-10-16 à 17 30 45

eldy:
> @NasserNgandu @eldy 
> for me the "`<br>`" should not be visible in edit mode when not using WYSIWYG. (that's what we had before!)
> 
> Capture d’écran 2025-10-16 à 17 30 45
> 

It depends.
Having no ckedtior does not mean we dont want to enter a content into html.
So there is 2 cases: if user has typed the br we must keep them.
If user has enter lines with enter/new lines, then text is full text (not html content) and if we edit the zone we should not see the br.

So if wysywig is off does not means we dont want to see br. It means we dont want to use a js editor to enter html content. But we must still be allowed to enter html. Html must remains html but html must not appear if we just enter text with new lines.

hregis:
@eldy Basically, WYSIWYG is not activated on the notes and I enter text with line breaks and when I edit the note again I have the "br"

eldy:
> @eldy Basically, WYSIWYG is not activated on the notes and I enter text with line breaks and when I edit the note again I have the "br"

Yes, the bug is confirmed, but fix is more complex than we it appears.
Pb is in tidy lib that make html sanitization: when content is not html, only lines with new lines, tidy corrupt the data by replacing the new lines with br when it should not.
I try to push a better fix thant the one of NasserNgandu 
@NasserNgandu can you check if my counter proposal work for you ?
I pushed it in v22 branch because tidy sanitization was introduced in v22.

NasserNgandu:
> @NasserNgandu can you test your fix when the wysiwyg is activated at public/private note ?

Yes, I tested it with WYSIWYG activated and it works well.

NasserNgandu:
> @NasserNgandu after test it's ok for me, can you propose this PR for 22 branch instead develop ?

@hregis Here is the PR link for version 22: https://github.com/Dolibarr/dolibarr/pull/35820

eldy:
Fixed by [5d3c80e](https://github.com/Dolibarr/dolibarr/commit/5d3c80e0afcb22c4be5fad3adda32ab0f0e2a185) so closed

PR #35803, opened 2025-10-17
FIX extra-field list depend on parent extra-field list on direct edit

lvessiller-opendsi:
FIX extra-field list depend on parent extra-field list on direct edit
- it's similar of PR #35165 but when you edit a field directly from card (live edit)
- for exemple when theses extrafields are always editables and you are on proposal card

**Before**
image

**After**
image

I only disabled options because if we hide it doesn't work with jQuery Select2 plugin

### DIFF
--- a/htdocs/core/js/lib_head.js.php
+++ b/htdocs/core/js/lib_head.js.php
@@ -1173,4 +1173,39 @@ $(document).ready(function() {
 	}
 });
 
+
+// Code to manage the js for combo list with dependencies (called by extrafields_view.tpl.php)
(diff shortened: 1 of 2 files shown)


hregis:
@lvessiller-opendsi Dolibarr 14 ! wouhouh... 😄 il va falloir mettre à jour ! hahaha

lvessiller-opendsi:
> @lvessiller-opendsi Dolibarr 14 ! wouhouh... 😄 il va falloir mettre à jour ! hahaha

Yes but it doesn't work from 14.0 until develop 😄

PR #35804, opened 2025-10-17
NEW get api jwt token

frederic34:
php-jwt version is 6.10.0 because it's the last version that support php 7.4
- [x] generate token
- [x] generate refresh token
- [x] make token usable by api
- [x] can refresh token using refresh token
- [x] waiting for comments


.gitleaks.toml
```diff
@@ -0,0 +1,5 @@
+[allowlist]
+    # Exclure un fichier spécifique par son chemin
+    paths = [
+        '''^htdocs/includes/php-jwt/README.md$''',
```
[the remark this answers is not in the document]
mdeweerd: In short:
This looks ok but the following is somewhat more generic
'''(?:^|/)htdocs/includes/php-jwt/README\ .md$'''`,` 

And `[[allowlists]]` would be the appropriate section now.
allowlist(s) do seem to be cumulative (they add to the existing list),

I would do a small local test on a branch with a secret in a file that should still be checked.

## Discussion:

The analysis of https://github.com/gitleaks/gitleaks/blob/master/config/gitleaks.toml shows these are indeed regexes.

The main question is : de .gitleaks.toml extend this list or replace it.
Possibly a '.gitleaksignore` file is more appropriate.

The readme of gitleaks indicates:
`# ⚠️ In v8.25.0 `[allowlist]` was replaced with '[[allowlists]]'.`


My guess is that `[[allowlists]]` can be specified multiple times and `[allowlist]` only once.  Therefore the latter might replace the default list, while the former probably adds to the list.   But the gitleaks readme indicates `Allowlist arrays are appended and can contain duplicates.`




Other observations:
So dots must be escaped.

At the start of the path they use `(?:^|/)` - I do not think this is needed here (the example is a generic configuration).

.gitleaks.toml
```diff
@@ -0,0 +1,5 @@
+[[allowlists]]
+    # Exclude a specific file
+    paths = [
+        '''(?:^|/)htdocs/includes/php-jwt/README\ .md$''',
```
mdeweerd: The `\` is for the dot, the space should be removed.

frederic34: aïe

hregis:
Hello @frederic34 

how to test with Multicompany ?

frederic34:
@hregis you can find the route in explorer (without DOLAPIKEY of course)
image

eldy:
> @hregis you can find the route in explorer (without DOLAPIKEY of course) image

By definition, API access is done using a token. So we should not have any feature to get a token from api because if we don't have token we should not be able to use API.
Token should be generated from backoffice, otherwise it is a big security hole. We added an exception to get the token from the login/pass but for convenient test, not for production usage and with this warning in api comment :
```
WARNING: You should NEVER use this API, like you should never use the similar API that uses the POST method. This will expose your password.
 * To use the APIs, you should instead set an API token to the user you want to allow to use API (This API token called DOLAPIKEY can be found/set on the user page) and use this token as credential for any API call.
 * From the API explorer, you can enter directly the "DOLAPIKEY" into the field at the top right of the page to get access to any allowed APIs.
```
Currently, it is recommanded to block this with the option "API_DISABLE_LOGIN_API", that can can block this unsecured use for production. 
But this option will be reverted soon to match security compliance rules, into something like "API_ALLOW_UNSECURED_TOKEN_ADMIN_FROM_LOGINPASS"
Because a login and password should never be used into an API calls.
So offering this new feature should be condtionned with this similar future option API_ALLOW_UNSECURED_TOKEN_ADMIN_FROM_LOGINPASS.

Also, do we really need to add a lib to manage tokens ?
I ask this because there is a current PR to move the current api token into the table of tokens llx_oauth_token (The PR to migrate data is already merged, but PR with code to use the new table not yet, target to have it merged is v23). So in a future, it will be possible to have several token with expiration date, additionnal permission, etc... for the same user. But for this need, it seems there is no need to introduce a new PHP dependency (that is a pain to manage). 
So can you describe the value added of jwt library, because a jwt is just a json string (with more information) used inside the bearer token instead of a simple string so i think in a long term vision, it costs less to forge it ourself that managing dependcy update, security troubles, php version compatibility pb, etc...) ?

frederic34:
I have changed the option to alow token from login/pass

Also we can reinvent the wheel by just decoding the token
From the doc of php-jwt :
```
// Decode headers from the JWT string WITHOUT validation
// **IMPORTANT**: This operation is vulnerable to attacks, as the JWT has not yet been verified.
// These headers could be any value sent by an attacker.
list($headersB64, $payloadB64, $sig) = explode('.', $jwt);
$decoded = json_decode(base64_decode($headersB64), true);
```
but I was thinking it was better to use a library

PR #35809, opened 2025-10-17
FIX: expense report card: do not show bank account if user cannot see them

marc-dll:
# FIX: expense report card: do not show bank account if user cannot see them

A user that has no rights on the bank module can see bank account ref + popin in the expense report card where a payment has been created.


rycks (approved):
ok

rycks (approved):

lvessiller-opendsi (approved):

PR #35812, opened 2025-10-17
FIX remove situation invoice from cycle

### LINKED ISSUE #35811: Situation invoice : can't remove a situation from cycle even if the credit has been applied
### Bug

When using credit on situation invoices, one can't remove the invoice from the situation cycle even if the credit has been used to "pay" the invoice.

### Dolibarr Version

Bug seen on 18.0 and develop.

### Environment PHP

any

### Environment Database

any

- create first situation
  - navigate to the "customer" tab of a thirdparty and "Create invoice"
  - create a first situation invoice S1
  - add 1 product line with progress = 30%
  - validate the invoice
- create the situation that has to be removed
  - create next situation S2
  - set the progress to 60%
  - validate
  - notice that button to remove from cycle is disabled : OK. we must create a credit invoice and apply the credit.
- try to remove it
  - create a credit invoice AS2 with same lines
  - validate AS2
  - set as available credit
  - navigate to S2
  - notice that button to remove from cycle is disabled : OK. we must apply the credit.
  - apply the relevant credit
  - notice that button to remove from cycle is disabled : **KO**. It should be enabled.

### Attached files

_No response_

thomas-Ngr:
# FIX #35811: you can't remove a situation from cycle even if available credit "pays" for the invoice

The bug is due to the fact that `$totalcreditnotes` is set at line 4784 and used at line 6714 :

```php
// l.4784

// l.6714
if (($object->total_ttc - $totalcreditnotes) == 0) {
```

But inbetween, we have a loop on next invoices in cycle that effectively overrides the value of `$totalcreditnotes` and sets it to `null`: 

```php
//l.5957
				foreach ($object->tab_next_situation_invoice as $next_invoice) {
					$totalpaid = $next_invoice->getSommePaiement(0);
					$totalcreditnotes = $next_invoice->getSumCreditNotesUsed(0);
```

### DIFF
--- a/htdocs/compta/facture/card.php
+++ b/htdocs/compta/facture/card.php
@@ -5128,9 +5128,9 @@ if ($action == 'create') {
 			$total_next_ht = $total_next_ttc = 0;
 
 			foreach ($object->tab_next_situation_invoice as $next_invoice) {
-				$totalpaid = $next_invoice->getSommePaiement(0);
-				$totalcreditnotes = $next_invoice->getSumCreditNotesUsed(0);
-				$totaldeposits = $next_invoice->getSumDepositsUsed(0);
+				$next_invoice_total_paid = $next_invoice->getSommePaiement(0);
+				$next_invoice_totalcreditnotes = $next_invoice->getSumCreditNotesUsed(0);
+				$next_invoice_totaldeposits = $next_invoice->getSumDepositsUsed(0);
 				$total_next_ht += $next_invoice->total_ht;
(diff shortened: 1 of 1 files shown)


rycks (approved):
ok for me, on a 18.0 
line 4080 there is an affectation

and into the foreach same var is override line 5132
$totalcreditnotes = $next_invoice->getSumCreditNotesUsed(0);

then last invoice of cycle override the value ...

lvessiller-opendsi (approved):

PR #35817, opened 2025-10-17
#Fix repeated types when multicompany is active

mapiolca:
When using Multicompany, each type and tickets are repeated as many times as the are entities.

### DIFF
--- a/.github/workflows/ci-on-pull_request.yml
+++ b/.github/workflows/ci-on-pull_request.yml
@@ -25,15 +25,15 @@ jobs:
   #  uses: ./.github/workflows/windows-ci.yml
   #  with:
   #    gh_event: ${{ github.event_name }}
-  gh-travis:  # Runs travis script on github runner (not on travis)
-    if: false
-    # needs: [pre-commit, phan]
-    # needs: [windows-ci]
-    secrets: inherit
-    needs: [pre-commit, phan, phpstan]
-    uses: ./.github/workflows/gh-travis.yml
-    with:
-      gh_event: ${{ github.event_name }}
+  #gh-travis:  # Runs travis script on github runner (not on travis)
+  #  if: false
+  #  # needs: [pre-commit, phan]
+  #  # needs: [windows-ci]
+  #  secrets: inherit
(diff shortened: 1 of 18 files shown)


hregis:
@mapiolca replace with : 

`$sql .= " AND ctt.entity IN (".getEntity('c_ticket_type').")";`

by default an entity use the dictionary of entity 1, but it's possible to have a dictionary by entity

hregis:
@mapiolca i work to have 3 possibilities : 

1 - use the dictionary of entity 1
2 - use the dictionary of current entity
3 - use the dictionary of entity 1 completed with current entity

mapiolca:
I’ve tried with your proposition, but it don’t work so when I use my solution, it’s work.

hregis:
@mapiolca je le fais en français...

si tu as activé cette option dans le module Multicompany et que tu as des soucis il faut relancer l'install de dolibarr et faire à nouveau la dernière mise à jour de la base. Multicompany va corriger les dictionnaires.

`Activer la personnalisation des types de ticket par entité (llx_c_ticket_type)`

pour rappel, les options concernant les dictionnaires dans le module ne sont pas des partages, mais des personnalisations par entité... si l'option n'est pas activée cela veut dire que toutes les entités utilisent le dictionnaire en question de l'entité principale.

normalement le code que je t'ai proposé doit fonctionner !

hregis:
@mapiolca i send a new PR for this : https://github.com/Dolibarr/dolibarr/pull/35873

mapiolca:
@hregis Great ! I will test it when merged. 

Next, I will try to fix showing files joined when an object is shared.

hregis:
@mapiolca check if the "documents" path used is "`->multi_output[$object->entity]`"

mapiolca:
@hregis OK, I'll check it.  Do you have a solution to resolve this alert ? It's happened when we are uploading a file in an entity that isn't the owner of the object. And the message appears until we display the object in the correct entity.

Capture d’écran 2025-10-22 à 20 20 44

hregis:
@mapiolca try to delete this record in llx_ecm_files and try again

PR #35819, opened 2025-10-17
Remove db attributes from webhook body

TitCaz:
# FIX|Fix #[*35818 Webhook body too long*]
Remove db attributes from objects being serialized to body of webhooks

### DIFF
--- a/htdocs/core/triggers/interface_95_modWebhook_WebhookTriggers.class.php
+++ b/htdocs/core/triggers/interface_95_modWebhook_WebhookTriggers.class.php
@@ -119,7 +119,7 @@ class InterfaceWebhookTriggers extends DolibarrTriggers
 				if (property_exists($resobject->object, 'errors')) {
 					unset($resobject->object->errors);
 				}
-				if(property_exists($resobject->object, 'db')) {
+				if (property_exists($resobject->object, 'db')) {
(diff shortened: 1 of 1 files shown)


hregis:
@TitCaz @eldy maybe it's better to use "dol_clone()" function ?

TitCaz:
Unfortunately, the problem is coming from dol_clone itself.
Because it's cloning everything which is a scalar, or an array; But an object can be hidden in an array...

Multiple solutions :
- Delete db field manually from linkedObjects
- On clone, iterate over elements of array to find Objects and dol_clone them
- On clone, iterate over elements of array and remove Objects

What do you think is the best ?
This raises the fact that dol_clone can keep some Objects in the clone.

hregis:
@TitCaz can you try this : 

```
if (property_exists($resobject->object, 'linkedObjects')) {
	foreach ($resobject->object->linkedObjects as $key => $linkedObject) {
		foreach ($linkedObject as $key2 => $obj) {
 			$resobject->object->linkedObjects[$key][$key2] = dol_clone($obj, 2);
		}
	}
}
```

eldy:
I solved the trouble by introducing a new method called: dol_clone_in_array()  that do recusing access to process all $db even for childs.

PR #35820, opened 2025-10-17
FIX : Loss of line breaks in the private and public note of a business proposal.

NasserNgandu:
Go to a business proposal

Edit the private or public note

Type a sentence, then press Enter to go to the next line

Save

Edit the note again by adding one or two lines below

Save

Note that line breaks are no longer respected

Here is the link to the issue: https://github.com/Dolibarr/dolibarr/issues/34940


eldy:
Should be fied with 5d3c80e0afcb22c4be5fad3adda32ab0f0e2a185, so i close ticket

PR #35824, opened 2025-10-18
NEW Accountancy - Transaction - Add verification on centralized account

aspangaro:
Verify if subsidiary account is autorized on centralized account

image


Pichinov-Jose:
Hi @aspangaro 
What do you think if we mask (or greyed) the subledger fields if the account is not a a centralized one? This can push all to configured correctly their accounts?
Thoughts?

aspangaro:
Hi @Pichinov-Jose See #35842 ;)

Need to update to the entry of accounting records now... Work differently

htdocs/accountancy/bookkeeping/card.php
```diff
@@ -198,6 +198,11 @@
 			setEventMessages($langs->trans('ErrorFieldRequired', $langs->transnoentitiesnoconv("AccountAccountingShort")), null, 'errors');
 			$action = '';
 		}
+		$subledger_account_str = is_array($subledger_account) ? reset($subledger_account) : (string) $subledger_account;
+		if (!checkGeneralAccountAllowsAuxiliary($db, $accountingaccount_number, $subledger_account)) {
```
eldy: May be you want to use $subledger_account_str instead of $subledger_account here ?

PR #35835, opened 2025-10-18
fix tva can be sometimes string instead float

frederic34:
@PsyCrow-code can you check, it's for branch 22


PsyCrow-code:
> @PsyCrow-code can you check, it's for branch 22

Yes, it is for branch 22. Issue detected after upgrade to v22

PR #35888, opened 2025-10-22
Fix - Task event linked from task and event

bradley-jarvis:
NEW This patch fixes task events so that they are correctly displayed in the task event list and that in the event the Linked Object refers back to the task.

The issue was that the elementtype value in the database was being referenced as both 'project_task' and 'task' by different parts of code

The only thing that this patch does not fix is that in the database table actioncomm the value in elementtype is both 'project_task' and 'task'. This patch should also update the database to rename 'task' to 'project_task' but I am not sure how to do this as part of the patch


eldy:
To complete migration i added this
UPDATE llx_actioncomm SET elementtype = 'project_task' WHERE elementtype = 'task';
into the migration script.

PR #35892, opened 2025-10-22
FIX: actions_addUpdateDelete password checked

RyAbn:
# FIX: actions_addUpdateDelete password check

Fields of type 'password' was checked by default with 'alphanohtml' which erase special chars.
Password with special chars was saved with wrong value.

### DIFF
--- a/htdocs/core/actions_addupdatedelete.inc.php
+++ b/htdocs/core/actions_addupdatedelete.inc.php
@@ -256,7 +256,11 @@ if ($action == 'update' && !empty($permissiontoadd)) {
 				$value = implode(',', $values_arr);
 			}
 		} elseif ($object->fields[$key]['type'] === 'password') {
-			$value = GETPOST($key, 'none');
+			/*
+			 * GETPOST with check 'None' is unauthorized
+			 * Bypass: check 'Cutom' with filter FILTER_UNSAFE_RAW which does nothing
+			 */
+			$value = GETPOST($key, 'custom', 0, FILTER_UNSAFE_RAW, 0);
 		} else {
 			if ($key == 'lang') {
 				$value = GETPOST($key, 'aZ09');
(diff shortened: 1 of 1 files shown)


htdocs/core/actions_addupdatedelete.inc.php
```diff
@@ -255,6 +255,12 @@
 			if (!empty($values_arr)) {
 				$value = implode(',', $values_arr);
 			}
+		} elseif ($object->fields[$key]['type'] === 'password') {
+			/*
+			 * GETPOST with check 'None' is unauthorized
```
rycks: @RyAbn why did you say that "none" is unauthorized ?

we try to follow that in 18.0 sources but "none" seems to be well handled, maybe you found a race condition, so could you please send us a little use-case ? (for example a "password" who does not works ?)

note: in develop a new "password" is comming ...

lvessiller-opendsi: @eldy 
What is your opinion about it ?

RyAbn: @rycks

There is a unit test that fails when we pass "none" as the $method argument in the GETPOST() function.

I ran some tests using the following password: 'password"test'

- GETPOST('password') => 'passwordtest' (Incorrect, it removes the double quotes)

- GETPOST('password', 'none') => 'password"test' (Correct, but unauthorized by the unit test)

- GETPOST('password', 'custom', 0, FILTER_UNSAFE_RAW, 0) => 'password"test' (Correct and works properly)

eldy: This is the label 
"PR to fix - see feedback in comment"
or
"PR to fix - Conflict or CI error"

eldy:
Note: a fix has been pushed into develop, based on using GETPOST(..., 'password')
#40702

PR #35893, opened 2025-10-22
FIX: Email Collector Module

RyAbn:
#FIX: Email Collector Module

Email Collector Module: manage error when imap_fetchstructure() return false. Previously it generated warning and fatal error because the returned value was not of type class.


lvessiller-opendsi:
Thanks for your contribution, it seems ok for me

lvessiller-opendsi (approved):

rycks (approved):

PR #35900, opened 2025-10-22
NEW : add conditional supplier price display

defrance:
conditional display of supplier price information based on advanced user permissions.

We have many other page with the same problem, propose a way to correct it before made it on other page


defrance (commented):
del spaces

PR #35902, opened 2025-10-22
NEW Added yearly and monthly breakdown totals to invoice template lists

Movix:
NEW Added yearly and monthly breakdown totals to invoice template lists


eldy:
Can you provide a screenshot of result after ?

Movix:
here a screenshot
image

PR #35907, opened 2025-10-23
NEW: add indexes on all entity columns

jyhere:
I recently worked on a script to merge all Dolibarr entities into one entity, which made me notice that many tables do not have an index on this column, slowing down queries that condition on this column.

I didn't go into detail, but all tables with the entity column now have an index on it.


hregis:
@jyhere great, but can you restore `llx_product_customer_price_log.sql` please, thanks

jyhere:
Whoops, sorry, I PR a fixed branch, and closed this PR.

PR #35909, opened 2025-10-23
NEW: add indexes on all entity columns

jyhere:
I recently worked on a script to merge all Dolibarr entities into one entity, which made me notice that many tables do not have an index on this column, slowing down queries that condition on this column.

I didn't go into detail, but all tables with the entity column now have an index on it.


eldy:
The entity field is not a selective field (if we have 2 entites, we can find only 2 different values, even if there is 1 million of record). This is why there is no index because having it may result into counterproductive performance when using dolibarr in a standard way.
Do you have an exemple of request that is slower/faster with index ? (with time for each of them) ?

jyhere:
Some requests without entity index and with entity index : 

 
image

eldy:
> Some requests without entity index and with entity index :
> 
> image

It seems your use case (your SQL request provided as exemple) is very specific to a mass migration of data.

For other use case (common use of dolibarr), the index will not be used but it will however be maintained and updated slowing down the application instead of speeding it.  

For such use case of migration or mass data update process, it is recommanded that the indexes are created dynamically by the migration tool/process and then deleted at end of migration.
So  i suggest to move your migration sql script into the repository 
https://github.com/Dolibarr/dolibarr-integration-resources 
into a directory called **entitymerger** where you can store both the sql to create all entity indexes and the tool you have created (whatever is its language and maturity). This repo is dedicated for that, to not lose a started work for a special migration use case.

PR #35914, opened 2025-10-23
NEW Add contact tab on product service

defrance:


defrance (commented):
oki

bos4711:
What's the point of this new tab in Products? I see no use case at all, and the inability to remove/hide it without altering the code seems impossible.

defrance:
In no particular order, the product manager, the manufacturing manager, the people authorized to perform the service, etc.

bos4711:
Maybe that's a normal situation to you but to me it sounds like an extreme corner case and I cannot anticipate a use case for it in our system.

Therefore I would suggest an option to turn off the tab in the Products view, preferably using a switch in the Products module.

bos4711:
Also, adding a feature like this - with absolutely no description, or reason, or explanation at all - is in my opinion not ideal.


@eldy Can you share your opinions about this feature?

defrance:
Do I need to remind you that this is an open-source project?
If you want a feature, nothing prevents you from developing it.

bos4711:
> Do I need to remind you that this is an open-source project? If you want a feature, nothing prevents you from developing it.

You could have just mentioned that this feature could be disabled with MAIN_DISABLE_CONTACTS_TAB instead of just sitting on the high horse.

PR #35929, opened 2025-10-24
QUAL Support user-based holiday calculation in num_open_day

rpopeescu:
# QUAL|Qual Extended num_open_day() function to support user-specific calculations

Extended the `num_open_day()` function to accept an optional `$fk_user` parameter, enabling user-specific open day calculation logic based on individual user calendars and country codes.

## Changes made:
- Modified `num_open_day()` function signature to accept `$fk_user` parameter
- Updated `actions_massactions.inc.php` to propagate user ID when approving leave requests
- Improved leave balance computation accuracy for users with distinct calendars or country configurations

## Benefits:
- More accurate leave balance calculations for multi-user environments
- Better support for organizations with users in different countries or with custom calendars
- Maintains backward compatibility with existing code that doesn't specify a user

This enhancement improves the precision of holiday/leave management calculations, particularly important for organizations with diverse user configurations.

### DIFF
--- a/htdocs/core/actions_massactions.inc.php
+++ b/htdocs/core/actions_massactions.inc.php
@@ -1742,7 +1742,7 @@ if (!$error && ($massaction == 'approveleave' || ($action == 'approveleave' && $
 				// If no SQL error, we redirect to the request form
 				if (!$error) {
 					// Calculate number of days consumed
-					$nbopenedday = num_open_day($objecttmp->date_debut_gmt, $objecttmp->date_fin_gmt, 0, 1, $objecttmp->halfday);
+					$nbopenedday = num_open_day($objecttmp->date_debut_gmt, $objecttmp->date_fin_gmt, 0, 1, $objecttmp->halfday, '', $objecttmp->fk_user);
 					$soldeActuel = $objecttmp->getCpforUser($objecttmp->fk_user, $objecttmp->fk_type);
 					$newSolde = ($soldeActuel - $nbopenedday);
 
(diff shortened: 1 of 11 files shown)


htdocs/core/actions_massactions.inc.php
```diff
@@ -1742,7 +1742,7 @@
 				// If no SQL error, we redirect to the request form
 				if (!$error) {
 					// Calculate number of days consumed
-					$nbopenedday = num_open_day($objecttmp->date_debut_gmt, $objecttmp->date_fin_gmt, 0, 1, $objecttmp->halfday);
+					$nbopenedday = num_open_day($objecttmp->date_debut_gmt, $objecttmp->date_fin_gmt, 0, 1, $objecttmp->halfday, '', $objecttmp->fk_user);
```
eldy: A more efficient solution would be to propagate into existing field country the $objecttmp->country_id  of the user.
I have modified the method num_open_day() so we can give directly the country id when we have it.

htdocs/core/lib/date.lib.php
```diff
@@ -1315,12 +1315,30 @@ function num_between_day($timestampStart, $timestampEnd, $lastday = 0)
  *	@param		int			$lastday            We include last day, 0: no, 1:yes
  *  @param		int			$halfday			Tag to define half day when holiday start and end
  *  @param      string		$country_code       Country code (company country code if not defined)
+ *  @param      int			$fk_user       		ID of the user for whom we are calculating
  *	@return    	int|string						Number of days or hours or string if error
  *  @seealso num_between_day(), num_public_holiday()
  */
-function num_open_day($timestampStart, $timestampEnd, $inhour = 0, $lastday = 0, $halfday = 0, $country_code = '')
+function num_open_day($timestampStart, $timestampEnd, $inhour = 0, $lastday = 0, $halfday = 0, $country_code = '', $fk_user = null)
```
eldy: On last develop, we can already propagate directly the counrty id of user so we should be able to avoid to propagate the user->id.

Issue #20240, opened 2022-03-01
Partially wrong permission test in viewcat.php


JohnXLivingston:
Hello,

In the following line, `restrictedArea` is called with the `$id` parameter:
https://github.com/Dolibarr/dolibarr/blob/8c94c2f98cbb881914923b28db57a35cf20926e3/htdocs/categories/viewcat.php#L73

But as you can see [here](https://github.com/Dolibarr/dolibarr/blob/8c94c2f98cbb881914923b28db57a35cf20926e3/htdocs/categories/viewcat.php#L67), sometimes we don't get `$id` but `$label` instead (for example from the [previous/next links](https://github.com/Dolibarr/dolibarr/blob/8c94c2f98cbb881914923b28db57a35cf20926e3/htdocs/categories/viewcat.php#L266))

JohnXLivingston:
This bug is still relevant.

ksar-ksar:
Hello,

Is it still the case on V18 ?

JohnXLivingston:
As far as i know, this bug is still present.

hregis:
fixed here : https://github.com/Dolibarr/dolibarr/pull/35933

PR #35930, opened 2025-10-24
FIX: access problem when label is used with next/prev

NasserNgandu:
In the following line, restrictedArea is called with the $id parameter:

[dolibarr/htdocs/categories/viewcat.php](https://github.com/Dolibarr/dolibarr/blob/8c94c2f98cbb881914923b28db57a35cf20926e3/htdocs/categories/viewcat.php#L73)

Line 73

 $result = restrictedArea($user, 'categorie', $id, '&category'); 

But as you can see [here](https://github.com/Dolibarr/dolibarr/blob/8c94c2f98cbb881914923b28db57a35cf20926e3/htdocs/categories/viewcat.php#L67), sometimes we don't get $id but $label instead (for example from the [previous/next links](https://github.com/Dolibarr/dolibarr/blob/8c94c2f98cbb881914923b28db57a35cf20926e3/htdocs/categories/viewcat.php#L266))

Here is the link to the issue: https://github.com/Dolibarr/dolibarr/issues/20240

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -2254,7 +2254,7 @@ abstract class CommonObject
 	public function getLineIdByLabel($label)
 	{
 		$sql = "SELECT rowid FROM ".$this->db->prefix().$this->table_element;
-		$sql .= " WHERE label = '" . $label . "'";
+		$sql .= " WHERE label = '" . $this->db->escape($label) . "'";
 		$resql = $this->db->query($sql);
 		if ($resql) {
 			$row = $this->db->fetch_row($resql);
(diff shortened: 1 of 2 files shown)


hregis:
@NasserNgandu i send a fix : https://github.com/Dolibarr/dolibarr/pull/35933

hregis:
@NasserNgandu some tables do not have a "label" field and your function risks creating side effects

NasserNgandu:
> @NasserNgandu some tables do not have a "label" field and your function risks creating side effects

I modified the method

hregis:
@NasserNgandu you can not modify showrefnav function with a specific function of category... See my fix : https://github.com/Dolibarr/dolibarr/pull/35933

PR #35952, opened 2025-10-25
clean code class commondict

frederic34:


eldy:
I don t want to have fetchAll in template interface because the use of a fetchAll is a often practice (generating a lot of trouble of performance when loading all data instead of managing a loop). If we add it interface it means it will be implemented and if implemented it will be use when we should.

frederic34:
ok removed

PR #35966, opened 2025-10-27
Add tpl files for standalone reception

sonikf:


hregis:
@sonikf it's not possible to use the globals "objectline_xxxx" ?

sonikf:
Hi @hregis 
Maybe but my aim is to try not to break existing functionality as reception like expedition is tied to orders.
Any help would be appreciated as we really need this for e-reporting in Greece.

hregis:
@sonikf can you add snapshots of your templates, what is different of global "objectline" templates ? thanks

hregis:
@sonikf ok, I didn't see that the expedition module had its own templates... forget what I said 😄

sonikf:
@hregis 
Any idea for the phan error? Code is the same in expedition and passed.

hregis:
@sonikf try with : 

`if ($cUnit->fetch((int) $line->fk_unit) > 0) {`

PR #35977, opened 2025-10-28
NEW Accountancy - Add accounting for discounts

aspangaro:
In the invoice or customer invoice modules, when a customer or supplier makes a partial payment or is partially paid, it is possible to supplement the payments with discounts ('escompte' in french).

Until now, these transactions were not recorded in the accounts. This PR uses the new page for generating miscellaneous transaction entries (introduced via the fixed asset module).

See the result:

image


hregis:
@aspangaro @eldy je vais déroger sur le principe de parler la langue de Shakespeare... je vais modifier la restriction actuelle qui veut que la comptabilité soit restreinte à l'entité courante : " = $conf->entity"
j'ai beaucoup de demandes dans un sens de partage... non seulement avec les sociétés qui se servent de Multicompany avec une seule entité comptable... mais aussi les sociétés qui gèrent des clients en portage ou autres sociétés de regroupements de facturations, etc..., etc.... il faut un fonctionnement plus souple niveau d'accès compta !

hregis:
@eldy @aspangaro ça peut être que de la visu de chiffres ou de l'inclusion des chiffres dans l'entité en question.... le tout c'est d'avoir un "getEntity()" dans les requêtes niveau compta au lieu d'une restriction à l'entité courante.

hregis:
you understand me @eldy  ? 😄

eldy:
> you understand me @eldy  ? 😄 

There is surely a bad understanding of what customer ask. Accounting dispatching of a company must never be done into another company. This will lead to critical problem and probably also non legal features.
If you customer want to manage accounting into one company only when using several it means he need a consolidated vision of accounting of several companies into one, but details must remains separated in each of them.
The need may be also a need in interface to be able to make dispatching once instead of having to do it in each company. But even in this case, data must be isolated and dispatched once and never mixed when we store them in database. But some external tool may try to make report mixing them, but storing (in database point of view) an accounting piece of entity x into entity y should not be done. However a screen or user point of view may be different than the database point of view and it is possible to let think that there is one dispatching for end user on screen when there is n in database storing.

PR #35989, opened 2025-10-29
New Hooks propals/orders

atm-quentin:
**PR Title:** `Feat: Add hooks for SQL list filtering (Propal, Order, etc.)`

## 💎 Context and Objective

This PR aims to add a series of new hooks throughout the Dolibarr core, mainly focused on proposals, orders, and product statistics.

The goal is to allow external modules (e.g., an entity/multicompany management module) to **inject custom `WHERE` conditions** into the SQL queries that generate lists, statistics, exports, and API responses.

Without these hooks, modules are forced to monkey-patch core files, which is a major source of technical debt and update conflicts.

## 🔬 Proposed Changes

This PR introduces new hook points in numerous files, including:

  * **Lists and Cards:** `htdocs/comm/card.php`, `htdocs/comm/index.php`, `htdocs/propal/index.php`, `htdocs/commande/index.php`, `htdocs/expedition/index.php`, etc.
  * **Classes (Stats, Boards):** `propal.class.php`, `commande.class.php`, `product.class.php`, `societe.class.php`, `propalestats.class.php`, `commandestats.class.php`.
  * **Boxes and Libs:** `core/boxes/box_commandes.php`, `core/boxes/box_propales.php`, `core/lib/order.lib.php`, `core/lib/propal.lib.php`.
  * **API:** `api_proposals.class.php`, `api_orders.class.php`.
  * **Exports:** `core/modules/modCommande.class.php`, `core/modules/modPropale.class.php`.

The main hook used is `printFieldListWhere`, which receives `$parameters` and `$object` to allow the module to determine the context.

-----

## 🤔 Open Question for @eldy

I have a major question about the **architectural approach** of this PR, and I would like your opinion before going further.

The `printFieldListWhere` hook is very **generic**. It is called from dozens of different places (a proposal list, a stats chart, a box, an API, etc.).

The consequence is that the module implementing this hook (e.g., `multicompany`) will have to contain a massive "God function" (`public function printFieldListWhere(...)`). This function will need to:

1.  Analyze the `$parameters` (e.g., `boxcode`, `type_element`) or the `$object->element` to guess *where* the hook is being called from.
2.  Contain a `switch` statement or a long series of `if/elseif` to return the correct SQL snippet (e.g., `AND (p.rowid ...)` or `AND (c.rowid ...)`).

**My question is:** Is this the validated way of doing things in Dolibarr?

I feel like we are only **moving the complexity from the Core to the Module**, not reducing it. The `multicompany` module (which will implement these hooks) will become very complex to maintain due to this manual routing.

### 💡 Alternative Suggestion

Wouldn't it be cleaner and more maintainable to have **more specific hooks**?

For example, instead of a generic `printFieldListWhere`:

  * In `htdocs/comm/index.php`, call `printFieldListWhere_commercialindex`
  * In `htdocs/core/boxes/box_commandes.php`, call `printFieldListWhere_boxcommandes`
  * In `htdocs/product/class/product.class.php` (for `load_stats_commande`), call `printFieldListWhere_productstatsorder`

This way, the module would implement small, atomic, and dedicated functions (SRP), making maintenance much simpler:

  * `public function printFieldListWhere_commercialindex(...)`
  * `public function printFieldListWhere_boxcommandes(...)`
  * `public function printFieldListWhere_productstatsorder(...)`
  
  I'm waiting for your suggestion

Thank you for your feedback on this approach.

### DIFF
--- a/htdocs/comm/card.php
+++ b/htdocs/comm/card.php
@@ -910,7 +910,6 @@ if ($object->id > 0) {
 
 		$resql = $db->query($sql);
 		if ($resql) {
-
 			$num = $db->num_rows($resql);
 			if ($num > 0) {
 				print '<div class="div-table-responsive-no-min">';
@@ -1018,7 +1017,6 @@ if ($object->id > 0) {
 
 		$resql = $db->query($sql);
 		if ($resql) {
-
 			$num = $db->num_rows($resql);
 			if ($num > 0) {
 				// Check if there are orders billable
--- a/htdocs/comm/card.php
+++ b/htdocs/comm/card.php
@@ -891,6 +891,7 @@ if ($object->id > 0) {
 	 * Latest proposals
 	 */
 	if (isModEnabled("propal") && $user->hasRight('propal', 'lire')) {
+		$propal_static = new Propal($db);
 		$langs->load("propal");
 
 		$sql = "SELECT s.nom, s.rowid, p.rowid as propalid, p.fk_projet, p.fk_statut, p.total_ht";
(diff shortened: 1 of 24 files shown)


defrance:
hello @atm-quentin 
maybe consider this PR as a better template of this problem : 
https://github.com/Dolibarr/dolibarr/pull/35203/files

eldy:
Duplicating the hook name to avoid to do a if into a common hook name will not change a lot.
Instead og doing a switch or if to manage the n different cases, the module will need to manage  n diffrent functions.
This is not really better than having a switch. It is even worse because the module will need to analyze dolibarr code to find all existing cases, when using a switch on context (or on a tag in $parameters), allow developers to guess the context and use case with a simple debug log line.

Having 100 function 
`public function printFieldListWhere_xxx`
where developper need to know the name/syntax of each of them seems worse than having 1 function
`public function printFieldListWhere`
called with $parameters['context'] to allow developper to decide in which case to change code.
If you really want to have n function, you can still do it into your module by doing a call of the n function by calling function with a dynamic name "function_ $parameters['context']" 

Currently,this is the expected behaviour.

PR #35991, opened 2025-10-29
NEW / DEVELOPP add display url bank card banner

atm-GregM:
NEW / DEVELOPP add display url bank card banner


eldy:
As it is not so easy to see where addition appears by reading code, can you paste a screenshot ?

PR #35996, opened 2025-10-29
Fix : Add 'printFieldListGroupBy' hook to stockatdate.php for module flexibility

atm-corentin:
# FIX|Fix #*Add 'printFieldListGroupBy' hook to stockatdate.php for module flexibility*

This PR is a follow-up to the discussion in #35895.

**Context:**
The original PR #35895 tried to fix an incorrect WAP (PMP) calculation in the "Stock at date" report for modules that manage WAP per entity (like Multicompany). The feedback was to use hooks instead of modifying the core query.

**Problem:**
After analysis, the existing hooks are insufficient. The `printFieldListWhere` hook is executed *after* the `GROUP BY` clause is already added to the query. This makes it impossible to correctly modify the grouping logic (e.g., to add `ppe.pmp` for per-entity WAP) without using a fragile `str_replace` hack.

**Solution:**
This PR introduces a new, flexible hook named `printFieldListGroupBy`.

* It is placed just *before* the `GROUP BY` clause is added.
* It passes the default `GROUP BY` string as a variable (`$sqlGroupBy`) in the `$parameters` array.
* It allows modules to cleanly **replace** this clause by setting `$this->resArray['sqlGroupBy']`.

This provides the necessary hook for modules to integrate properly and fixes the underlying issue from #35895 in a clean, robust, and future-proof way.

### DIFF
--- a/htdocs/product/stock/stockatdate.php
+++ b/htdocs/product/stock/stockatdate.php
@@ -287,7 +287,7 @@ if (!empty($search_fk_warehouse)) {
 }
 // Add fields from hooks
 $parameters = array();
-$reshook = $hookmanager->executeHooks('printFieldListSelect', $parameters); // Note that $action and $object may have been modified by hook
+$reshook = $hookmanager->executeHooks('printFieldListSelect', $parameters, $object, $action); // Note that $action and $object may have been modified by hook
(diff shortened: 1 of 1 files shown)


atm-corentin:
@eldy

PR #35997, opened 2025-10-29
fix date start and date end in massaction

ATM-NicolasV:
Bugfix: Robust Date Comparison for MO (Manufacturing Order) Planned Dates

Problem

The existing validation logic for modifying the planned start and end dates of a Manufacturing Order (MO) fails when one of the opposing dates is not yet defined (stored as an empty string "").

This issue is caused by PHP's loose comparison :

The positive timestamp $newDate is compared against the empty string "".

PHP implicitly converts "" to the integer 0.

The comparison (e.g., timestamp < 0) evaluates to FALSE, prematurely executing the error block (setEventMessages) and preventing the user from successfully setting the initial date.

Solution

The fix standardizes the object's date attributes ($objMo->date_end_planned, etc.) into reliable integer timestamps before performing validation.

We use is_numeric() to detect existing timestamps (positive integers).

We use strtotime() to safely convert empty strings ("" becomes 0) or date strings to integers.

The validation condition is updated to explicitly allow the modification if the opposing date is undefined (value $\le 0$ after conversion).

### DIFF
--- a/htdocs/mrp/mo_list.php
+++ b/htdocs/mrp/mo_list.php
@@ -255,15 +255,17 @@ if (empty($reshook)) {
 					if ($objMo->fetch($idMo)) {
 						if ($objMo->status == Mo::STATUS_DRAFT) {
 							if (!empty($changeDate)) {
-								if ($action == 'changedatestart_confirm') {			// Test on permission not required
-									if ($newDate < $objMo->date_end_planned) {
+								if ($action == 'changedatestart_confirm') {
+									// La date de début peut être définie SI (la date de fin est vide OU la nouvelle date est AVANT la date de fin existante).
+									if (empty($objMo->date_end_planned) || $newDate < $objMo->date_end_planned) {
 										$objMo->date_start_planned = $newDate;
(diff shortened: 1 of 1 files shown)


ATM-NicolasV (commented):
I have make change for comments

htdocs/mrp/mo_list.php
```diff
@@ -253,34 +253,32 @@
...
-									}
-								} elseif ($action == 'changedateend_confirm') {		// Test on permission not required
-									if ($newDate > $objMo->date_start_planned) {
-										$objMo->date_end_planned = $newDate;
-									} else {
-										setEventMessages($langs->trans('ErrorModifyMoDateEnd', $objMo->ref), null, 'errors');
-										break;
-									}
+						if (!empty($changeDate)) {
+							if ($action == 'changedatestart_confirm') { 		// Test on permission
```
eldy: Must be   // Test on permission not required

atm-corentin:
@eldy can you check pls ?

PR #36003, opened 2025-10-29
TakePOS Cash Control report: Fix incorrect float approximation and string/float mix on test

mpellegrin:
The test used in the Cash Control report of TakePOS was using float approximation between equality tests, leading to incoherent information shown in cash report. On the first test, string/float mix was also used.

This commit changes the test to use the displayed value for testing, to show the "Declared" red label only if a difference is actually visible to the user.

image


htdocs/compta/cashcontrol/report.php
```diff
@@ -406,7 +406,7 @@
 		print '<div class="inline-block amount width100"></div>';
 		print '<div class="inline-block amount width100">'.price($cash).'</div>';
 	}
-	if (!$summaryonly && $object->status == $object::STATUS_VALIDATED && $newcash != $object->cash) {
+	if (!$summaryonly && $object->status == $object::STATUS_VALIDATED && price($newcash) != price($object->cash)) {
```
eldy: price is a text formating function.
For comparison of amount, it should be smantically better to do 
`prcie2num( ) != price2num()`
Can you check this way ?
Of you can also try
`if (float) $newcash != (float) $object->cash`

PR #36004, opened 2025-10-29
FIX NewOnlineSign response format

jp-clerc:
Hi. I identified a problem using the onlinesign feature.
Once signed on the browser, I get the response **"Error on calling the core/ajax/onlineSign.php. See console log."** but the proposal status goes to "Signed" and the dolibarr.log showed nothing wrong.
The response from the console.log was **" response: '{"status":"ok"}success' "** so I guessed it was better to use "includes" instead of ".trim" to get the "sucess" keyword and validate the signature.


eldy:
There is no reason to have a string 
'{"status":"ok"}success' 
as an answer. We should get the string 'success'.
I can't reproduce.
Do you have an external module activated ?

jp-clerc:
I only have a Custom Webhook module triggered by 'PROPAL_CLOSE_SIGNED' that fetch the Company name and the total amount.

eldy:
> I only have a Custom Webhook module triggered by 'PROPAL_CLOSE_SIGNED' that fetch the Company name and the total amount.

And what if you disable/remove it ? Can you also ive on which version do you experience trouble.

eldy:

PR #36028, opened 2025-10-31
NEW Show cron last result and output in info

vmaury:
#NEW From now, last output and last result of a cron are only displayed in cron's list, and output isn't very convenient to read

This PR adds last output and last result in cron's info page
image

### DIFF
--- a/htdocs/cron/info.php
+++ b/htdocs/cron/info.php
@@ -78,7 +78,10 @@ print '<br>';
 
 print '<table width="100%"><tr><td>';
 dol_print_object_info($object);
-print '</td></tr></table>';
+print '</td></tr>';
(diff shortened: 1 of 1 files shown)


htdocs/cron/info.php
```diff
@@ -78,7 +78,10 @@
 
 print '<table width="100%"><tr><td>';
 dol_print_object_info($object);
-print '</td></tr></table>';
+print '</td></tr>';
+print '<tr><td>'.$langs->trans('CronLastResult').' : '.$object->lastresult.'</td></tr>';
```
eldy: lastresult and lastoutput are data coming from outside. So it must be sanitized when outpout.
Can you try  dolPrintHTML($object->lastoutput);

PR #36033, opened 2025-10-31
NEW add the option to not synchronize thirdparty <--> member

hregis:


hregis:
@eldy phan and php-stan errors already fixed here : https://github.com/Dolibarr/dolibarr/pull/36026

htdocs/adherents/card.php
```diff
@@ -416,7 +416,9 @@
 			}
 
 			if (!$error) {
-				$result = $object->update($user, 0, $nosyncuser, $nosyncuserpass);
+				$nosyncthirdparty = (!getDolGlobalString('MEMBER_NO_SYNC_LINKED_THIRDPARTY') ? 0 : 1);
```
eldy: Wy not just 
`$nosyncthirdparty = getDolGlobalInt('MEMBER_NO_SYNC_LINKED_THIRDPARTY');`

hregis: @eldy Yes, you're right, it's an old reflex 😄

iLLixM:
@hregis @eldy
Thank you very much for this special functional enhancement. I've noticed in various places that the mix of English and French in Dolibarr development is always a challenge. In this particular PR, I find this mix almost bizarre. The new system parameters are: MEMBER_NO_SYNC_LINKED_THIRDPARTY and SOCIETE_NO_SYNC_LINKED_MEMBER

Especially in the second case, the mix of the French "SOCIETE" and the English "MEMBER" is incomprehensible to me. Regarding the first parameter, I would have expected the following: THIRDPARTY_NO_SYNC_LINKED_MEMBER

According to the wiki, new parameters are usually introduced exclusively in English.
https://wiki.dolibarr.org/index.php/Setup_Other#Third_Parties_Module
Best regards

eldy:
> @hregis @eldy
> Thank you very much for this special functional enhancement. I've noticed in various places that the mix of English and French in Dolibarr development is always a challenge. In this particular PR, I find this mix almost bizarre. The new system parameters are: MEMBER_NO_SYNC_LINKED_THIRDPARTY and SOCIETE_NO_SYNC_LINKED_MEMBER
> 
> Especially in the second case, the mix of the French "SOCIETE" and the English "MEMBER" is incomprehensible to me. Regarding the first parameter, I would have expected the following: THIRDPARTY_NO_SYNC_LINKED_MEMBER
> 
> According to the wiki, new parameters are usually introduced exclusively in English.
> https://wiki.dolibarr.org/index.php/Setup_Other#Third_Parties_Module
> Best regards

The reason is that the begin of a new parameter must follow the technical name of the module he is related to. 
Currently it is SOCIETE. Renaming module technical key is not yet possible, we need to progress on other tasks first.

iLLixM:
> The reason is that the begin of a new parameter must follow the technical name of the module he is related to. Currently it is SOCIETE. Renaming module technical key is not yet possible, we need to progress on other tasks first.

Thank you for your reply @eldy .
Unfortunately, this statement, considering the naming of the system parameter MEMBER_NO_SYNC_LINKED_THIRDPARTY and many other parameters on the page https://wiki.dolibarr.org/index.php/Setup_Other, is causing even more confusion.

If the name of the affected module is decisive for the naming of system parameters, then many parameters recently added to the source code must have been named differently.
Especially with the Member or rather the Adherent module, there's a lot of confusion between the two terms.

Is it named the Member or the Adherent module?
Where is the definitive source for clearly identifying a module's title?
Can I see the title of a module in the name of the module folder in htdocs, or is that defined elsewhere?

Sorry for all the questions. I just want to know the correct procedure for any furture code additions.
Thank you

PR #36044, opened 2025-10-31
Fix #36019 : bank account count was wrong on thirdparty card

### LINKED ISSUE #36019: Wrong count of payment mode because of either NULL or empty varchar
### Bug

If I add a bank account on a thirdparty, the count is still 0 : 

Image

But if I add it by CSV import, the count is correct : 

Image

The difference is in database : 
On the first case, llx_societe_rib.stripe_card_ref = '' (empty varchar)
On the second case, llx_societe_rib.stripe_card_ref = NULL

Because the code in htdocs/core/lib/company.lib.php counts only bank accounts where stripe_card_ref = NULL : 

```
		$sql = "SELECT COUNT(n.rowid) as nb";
		$sql .= " FROM ".MAIN_DB_PREFIX."societe_rib as n";
		$sql .= " WHERE n.fk_soc = ".((int) $object->id);
		if (!isModEnabled('stripe')) {
			$sql .= " AND n.stripe_card_ref IS NULL";
		} else {
			$sql .= " AND (n.stripe_card_ref IS NULL OR (n.stripe_card_ref IS NOT NULL AND n.status = ".((int) $servicestatus)."))";
		}

		$resql = $db->query($sql);
		if ($resql) {
			$nbBankAccount = $obj->nb;
		} else {
			dol_print_error($db);
		}
```

These fields are either NULL or '' (empty varchar) depending the creation method : 
`code_banque, code_guichet, number, cle_rib, domiciliation, proprio, owner_address, stripe_card_ref, stripe_account`

So I think there is 2 solutions : 
a) Store NULL in all these fields instead of '' (empty varchar)
b) Replace
```
        if (!isModEnabled('stripe')) {
            $sql .= " AND n.stripe_card_ref IS NULL";
        } else {
```
by 

```
        if (!isModEnabled('stripe')) {
            $sql .= " AND (n.stripe_card_ref IS NULL OR n.stripe_card_ref = '')";
        } else {
```

### Dolibarr Version

20.0.4

### Environment PHP

8.2

### Environment Database

PostgreSQL

_No response_

### Attached files

_No response_

atm-maxime:
@eldy is this the right way to fix it ? Or should we force the stripe_card_ref to be null on UI creation instead of empty ?

### DIFF
--- a/htdocs/core/lib/company.lib.php
+++ b/htdocs/core/lib/company.lib.php
@@ -216,7 +216,7 @@ function societe_prepare_head(Societe $object)
 		$sql .= " FROM ".MAIN_DB_PREFIX."societe_rib as n";
 		$sql .= " WHERE n.fk_soc = ".((int) $object->id);
 		if (!isModEnabled('stripe')) {
-			$sql .= " AND n.stripe_card_ref IS NULL";
+			$sql .= " AND (n.stripe_card_ref IS NULL OR n.stripe_card_ref ='')";
 		} else {
 			$sql .= " AND (n.stripe_card_ref IS NULL OR (n.stripe_card_ref IS NOT NULL AND n.status = ".((int) $servicestatus)."))";
 		}


eldy:
> @eldy is this the right way to fix it ? Or should we force the stripe_card_ref to be null on UI creation instead of empty ?
 
Both would be better, but this one is enough as it is the most secured solution.

PR #36051, opened 2025-11-01
Remove obsolete check for MAIN_MODULE_OPENIDCONNECT during login

### LINKED ISSUE #36048: BUG: OpenIDConnect Module disabled - and it can not be searched in 23.0.0-alpha
### Bug

When searching for openid in the Modules/Applications in version 23.0.0-alpha, then no openid shows up. However if I do the same search in v22.0.1 (my staging) or v21.0.1 (my production), then I do get a module I can enable.

Any other search in v23.0.0-alpha does seem to produce a result?

I am trying to search because dolibarr.log says

```
2025-11-01 09:46:43 INFO    192.168.127.1        58     33 functions_openid_connect::check_user_password_openid_connect Module disabled
```

When I specify this configuration
`$dolibarr_main_authentication='openid_connect,dolibarr'`

and try to log in, thus I try to search for it to enable it.

### Dolibarr Version

23.0.0-alpha

### Environment PHP

dolibarr docker image v22

### Environment Database

mariadb latest

_No response_

### Attached files

# v. 23.0.0-alpha
Image

# v. 22.0.1
Image

# v. 21.0.1
Image

JonBendtsen:
# FIX #36048 replace var MAIN_MODULE_OPENIDCONNECT with MAIN_AUTHENTICATION_OIDC_ON because openid_connect is integrated into the core
This PR is an attempt to fix #36048, which happens when you set 
```
$dolibarr_main_authentication='openid_connect,dolibarr';
```
and have configured the openid_connect settings in `/admin/openid_connect.php`

### DIFF
--- a/htdocs/core/login/functions_openid_connect.php
+++ b/htdocs/core/login/functions_openid_connect.php
@@ -42,7 +42,7 @@ function check_user_password_openid_connect($usertotest, $passwordtotest, $entit
 {
 	global $db;
 
-	if (getDolGlobalInt('MAIN_MODULE_OPENIDCONNECT', 0) <= 0) {
+	if (getDolGlobalInt('MAIN_AUTHENTICATION_OIDC_ON', 0) <= 0) {
 		$_SESSION["dol_loginmesg"] = "OpenID Connect is disabled";
 		dol_syslog("functions_openid_connect::check_user_password_openid_connect Module disabled");
 		return false;
--- a/htdocs/core/tpl/login.tpl.php
+++ b/htdocs/core/tpl/login.tpl.php
@@ -173,7 +173,7 @@ if (getDolGlobalString('MAIN_OPTIMIZEFORTEXTBROWSER')) {
 }
 
 // If OpenID Connect is set as an authentication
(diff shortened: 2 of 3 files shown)


htdocs/core/login/functions_openid_connect.php
```diff
@@ -42,12 +42,6 @@ function check_user_password_openid_connect($usertotest, $passwordtotest, $entit
 {
 	global $db;
 
-	if (getDolGlobalInt('MAIN_MODULE_OPENIDCONNECT', 0) <= 0) {
```
eldy: I have pushed a change into develop to introduce an option MAIN_AUTHENTICATION_OIDC_ON so we can enable/disable OIDC globaly from setup.
Can you replace MAIN_MODULE_OPENIDCONNECT with MAIN_AUTHENTICATION_OIDC_ON instead of removing ?

JonBendtsen: @eldy no :-( this is my first entry into openid, and I run my own authentik where I think I still have configuration issues.

It might not help that I run both my Dolibarr and my Authentik inside separate podman pods, only exposing the ports that should be exposed - and further more both run on my mac laptop in podman desktop.

PR #36061, opened 2025-11-01
Fix: Polish (pl_PL) time format - use 24h instead of AM/PM

nyupyu:
# FIX Polish (pl_PL) time format - use 24h instead of AM/PM

Fixed time format in Polish language file. Poland uses 24-hour format, not AM/PM.

Changes:
- Changed %I (12h) to %H (24h)
- Removed %p (AM/PM indicator)  
- Changed date format from MM/DD/YYYY to DD/MM/YYYY (Polish standard)
- Removed comma from date-time text formats for consistency

Tested on Dolibarr 22.0.2 with pl_PL locale.


frederic34:
@nyupyu Translations for languages other than en_US are managed in Transifex. Request access if you need to update the Polish translation

nyupyu:
> @nyupyu Translations for languages other than en_US are managed in Transifex. Request access if you need to update the Polish translation
> 
> https://www.transifex.com/dolibarr-association/dolibarr/

I will, thanks for the info.

eldy:
According to this site https://lh.2xlibre.net/locale/pl_PL/
The format of pl_PL is using . between dat and mont, year.
So i have updated transife to match this.

PR #36067, opened 2025-11-02
FIX Accountancy - Problem with general setup

aspangaro:
Lot of problem setup since option are added (export & numbering).

Redefine all "form" to be able to save all options without any problems

image


frederic34:
image

the test is useless

aspangaro:
> image
> the test is useless

You're right. I remove this useless test

PR #36072, opened 2025-11-02
clean code

frederic34:
$modecompta not empty here


hregis:
@frederic34 this a new rules to send an empty parameter ?

frederic34:
image

what is the value of $modecompta ?

hregis:
@frederic34 arf ! CREANCES-DETTES ! :-)

PR #36074, opened 2025-11-03
FIX 16.0: extrafield of type link to category causes SQL error in selectForFormsList()

atm-florianm:
# FIX link-to-category extrafield causes SQL error
## How to reproduce?
In Dolibarr 17, 18 or 19:

1. Enable modules: project and categories.
2. Navigate to the setup page of the Projects / Opportunities module
3. Create an extrafield of type "link to an object" with value: `Categorie:categories/class/categorie.class.php`
4. Navigate to the list of projects

→ there is a non-blocking SQL error when displaying the filter for the column of the extrafield: 
```
Unknown column 't.ref' in 'SELECT'
```

## Versions of Dolibarr
The bug is present in Dolibarr 17 to 19. [edit] maybe in Dolibarr 16 too but I didn't test my fix there

In Dolibarr 20 and higher, the Categorie class has a `fields` property that is properly set up, which fixes this problem in a much cleaner way, so no need keep this fix in v20 and higher.

## This fix
This just checks whether the object element is `categorie` and, if so, forces `label` instead of `ref` as the field to select.

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -7698,7 +7698,7 @@ class Form
 			if ($tmpfieldstoshow) {
 				$fieldstoshow = $tmpfieldstoshow;
 			}
-		} else if ($objecttmp === 'categorie') {
+		} else if ($objecttmp->element === 'category') {
 			$fieldstoshow = 't.label';
 		} else {
 			// For backward compatibility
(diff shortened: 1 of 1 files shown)


htdocs/core/class/html.form.class.php
```diff
@@ -7821,6 +7698,8 @@ public function selectForFormsList($objecttmp, $htmlname, $preselectedvalue, $sh
 			if ($tmpfieldstoshow) {
 				$fieldstoshow = $tmpfieldstoshow;
 			}
+		} else if ($objecttmp === 'categorie') {
```
atm-florianm: I made ~~a copy-paste mistake~~ two copy-paste mistakes.
```suggestion
		} else if ($objecttmp->element === 'category') {
```

PR #36076, opened 2025-11-03
NEW: Rework of the management of the card and fields on the web portal

kkhelifa-opendsi:
Rework of the management of the card and fields on the web portal


kkhelifa-opendsi:
Error precommit and travis from external commit

htdocs/core/ajax/ajaxfield.php
```diff
@@ -0,0 +1,150 @@
+<?php
```
eldy: Is it a file for the webportal ? If yes, we should find it into public/webportal/... otherwise, when webportal will be behind its own virtual host name, the file will not be reachable as on file into subdir public/webportal are files the browser can connect to.

Also, as it is a page, we shoul dhave a protection to disable it when module webportal is off.

The include should be webportal.main.inc.php because this main may use a different policy for cookie and security http headers. The main.inc.php is reserved for the security inside the backoffice that is a different environment than the one of the webportal.

kkhelifa-opendsi:
Error phan and phpstan from external commit

htdocs/core/class/html.form.class.php
```diff
@@ -12847,4 +12847,417 @@ public function getSearchFilterToolInput($dataTarget, $htmlName = 'search-tools-
...
+	 * @param	int				$notooltip				No tooltip
+	 * @param  	string  		$morecss            	''=Add more css on link
+	 * @param	int				$add_label				0=Default, 1=Add label into string, >1=Add first chars into string
+	 * @param	string			$sep					' - '=Separator between ref and label if option 'add_label' is set
+	 * @return	string									String with URL
+	 */
+	public function getNomUrl(&$object, $withpicto = 0, $option = '', $maxlength = 0, $save_lastsearch_value = -1, $notooltip = 0, $morecss = '', $add_label = 0, $sep = ' - ')
+	{
+		if (is_object($object) && method_exists($object, 'getNomUrl')) {
+			$out = $object->getNomUrl($withpicto, $option, $maxlength, $save_lastsearch_value, $notooltip, $morecss, $add_label, $sep);
```
mdeweerd: @kkhelifa-opendsi @eldy 

My local phan run identifies the following issues here which do not seem to be false positives to me.
```
htdocs\core\class\html.form.class.php:13268 PhanParamTooMany Call with 8 arg(s) to \Categorie::getNomUrl(int $withpicto = 0, string $option = '', int $maxlength = 0, string $moreparam = '', int $notooltip = 0, string $morecss = '', int $save_lastsearch_value = 0) which only takes 7 arg(s) defined at htdocs\categories\class\categorie.class.php:1888
htdocs\core\class\html.form.class.php:13268 PhanPluginSuspiciousParamPosition Suspicious order for argument save_lastsearch_value - This is getting passed to parameter #7 (int $save_lastsearch_value) of \Categorie::getNomUrl(int $withpicto = 0, string $option = '', int $maxlength = 0, string $moreparam = '', int $notooltip = 0, string $morecss = '', int $save_lastsearch_value = 0) defined at htdocs\categories\class\categorie.class.php:1888
htdocs\core\class\html.form.class.php:13268 PhanTypeMismatchArgument Argument 4 ($moreparam) is $save_lastsearch_value of type int but \Categorie::getNomUrl() takes string defined at htdocs\categories\class\categorie.class.php:1888
```

Can also be found on https://cti.dolibarr.org/ .

PR #36078, opened 2025-11-03
fix CI V21

frederic34:


htdocs/compta/facture/class/facture.class.php
```diff
@@ -3953,7 +3953,7 @@ public function addline(
 			if (empty($fk_prev_id)) {
 				$fk_prev_id = 'null';
 			}
-			if (!isset($situation_percent) || $situation_percent > 100 || (string) $situation_percent == '' || $situation_percent === null) {
+			if (!isset($situation_percent) || $situation_percent > 100 || (string) $situation_percent == '' || $situation_percent === null) { // @phpstan-ignore-line
```
mdeweerd: I do not know what the notice was, but testing $situation_percent as a number before testing it is empty and before null is not ok.

`!isset($situation_percent) || (string) $situation_percent == '' || $situation_percent > 100 )`
is better - the null test is implied.

frederic34: It was  `Strict comparison using === between mixed and null will always evaluate to false.`

PR #36087, opened 2025-11-04
FIX: MAIN_AUTOFILL_DATE in supplier invoice display 01/01/1970

FHenry:
With option : MAIN_AUTOFILL_DATE on create supplier Invoice from scratch

the cast in (int) of date is not needed (never done elsewhere in Dolibarr Core)

before :
image

after

image

Also in this PR fix php warning

### DIFF
--- a/htdocs/fourn/facture/card.php
+++ b/htdocs/fourn/facture/card.php
@@ -2744,7 +2744,7 @@ if ($action == 'create') {
 		// Date invoice
 		print '<tr><td class="fieldrequired">'.$langs->trans('DateInvoice').'</td><td>';
 		print img_picto('', 'action', 'class="pictofixedwidth"');
-		print $form->selectDate($dateinvoice, '', 0, 0, 0, "add", 1, 1);
+		print $form->selectDate($dateinvoice ? (int) $dateinvoice : '', '', 0, 0, 0, "add", 1, 1);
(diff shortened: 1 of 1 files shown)


htdocs/fourn/facture/card.php
```diff
@@ -2740,10 +2740,11 @@ function setRadioForTypeOfInvoice() {
 		// Label
 		print '<tr><td>'.$langs->trans('Label').'</td><td><input class="minwidth200" name="label" value="'.dol_escape_htmltag(GETPOST('label')).'" type="text"></td></tr>';
 
+
 		// Date invoice
 		print '<tr><td class="fieldrequired">'.$langs->trans('DateInvoice').'</td><td>';
 		print img_picto('', 'action', 'class="pictofixedwidth"');
-		print $form->selectDate((int) $dateinvoice, '', 0, 0, 0, "add", 1, 1);
+		print $form->selectDate($dateinvoice, '', 0, 0, 0, "add", 1, 1);
```
eldy: To avoid CI error, i suggest
print $form->selectDate($dateinvoice ? (int) $dateinvoice : '', '', 0, 0, 0, "add", 1, 1);

FHenry: Right and done

PR #36106, opened 2025-11-05
FIX : Wrong cast on TVA rate when updating supplier order lines

atm-adrien:
## FIX : Wrong cast on TVA rate when updating supplier order lines


frederic34:
https://github.com/Dolibarr/dolibarr/pull/36109

PR #36107, opened 2025-11-05
setErrors does not exists, same solution as dolibarr v19.0

rycks:
There is no setErrors on that object
I apply same solution as we can find in dolibarr v19.0


lvessiller-opendsi (approved):

lvessiller-opendsi:
@eldy ok for me
Be careful, there is a problem (CI error) with set reviewer when the reviewer is himself.

lvessiller-opendsi (approved):

lvessiller-opendsi (approved):

lvessiller-opendsi (approved):

PR #36117, opened 2025-11-05
FIX: fetch LDAP members in LDAP sync sync_groups_ldap2dolibarr.php

labiloute:
# FIX|Fix

When you run the script "scripts/user/sync_groups_ldap2dolibarr.php", it doesn't fetch the members of the group.
The method "fetch" for UserGroup object require a boolean as third arg to do so.

The consequences were :
 - the users were always recreated (at every sync)
 - the users weren't deleted when removed from the LDAP group, because the comparison between existing members returns nothing.

I think this occur since a looooooong time ago :), but don't know how long.
I PR on 20.0 and I don't know if I need to PR on 21.0, 22.0, etc.


eldy:
> # FIX|Fix
> When you run the script "scripts/user/sync_groups_ldap2dolibarr.php", it doesn't fetch the members of the group. The method "fetch" for UserGroup object require a boolean as third arg to do so.
> 
> The consequences were :
> 
> * the users were always recreated (at every sync)
> * the users weren't deleted when removed from the LDAP group, because the comparison between existing members returns nothing.
> 
> I think this occur since a looooooong time ago :), but don't know how long. I PR on 20.0 and I don't know if I need to PR on 21.0, 22.0, etc.

If you PR on v20, no need to PR on v21 or v22. There regularly merge from downstream to upstream.

PR #36118, opened 2025-11-05
FIX Select correct approver when making a leave request for someone else

bchanudet:
# FIX Select correct approver when making a leave request for someone else

As long as you have the correct permissions, you can create a leave request for a colleague. You can go to `/htdocs/holiday/list.php?id={id_colleague}`, and there will be a button there, "Make a leave request". You are then redirected to the "new leave request" form, with the colleague preselected as user. The problem is, your own supervisor is also preselected, not theirs. It could lead to a leave request sent to the wrong approver. 

This PR fixes the problem by looking if we are in this case, and if yes, select the colleague's supervisor (or the designated approver) instead of the connected user's supervisor.

Steps to reproduce:

Given the following hierachy:
image

BEFORE: 
When I create a leave request for my colleague, it's my own supervisor that is selected by default
image

AFTER:
Now, their supervisor is selected by default, even if it's not my supervisor
image

### DIFF
--- a/htdocs/holiday/card.php
+++ b/htdocs/holiday/card.php
@@ -1164,6 +1164,12 @@ if ((empty($id) && empty($ref)) || $action == 'create' || $action == 'add') {
 			// Defined default approver (the forced approved of user or the supervisor if no forced value defined)
 			// Note: This use will be set only if the deinfed approvr has permission to approve so is inside include_users
 			$defaultselectuser = (empty($user->fk_user_holiday_validator) ? $user->fk_user : $user->fk_user_holiday_validator);
+			if ($fuserid != $user->id) {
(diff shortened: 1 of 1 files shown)


hregis:
@bchanudet can you replace "$target" with "$fuser"... Thanks

eldy:
@mdeweerd can you help me with the error phan ?

mdeweerd:
> @mdeweerd can you help me with the error phan ?
@eldy Yes, I propose #36147 - ~we'll see if it passes the CI.~ It passes CI, but there is some other issue in TRAVIS.
Also I se that this PR is for Dolibarrr:20.0, so I am updating the target branch.

PR #36120, opened 2025-11-06
Cast variable $s to string in dol_eval function

PsyCrow-code:
FIX So many warnings on passing null parameter on functions like preg_match(), str_replace(), preg_replace().
Passing null to a string parameter is deprecated on PHP 8.1+ and will be a TypeError on PHP 9.0+
explicit cast (string)$var return '' on null and is compatible since PHP 4


PsyCrow-code (commented):
fix space after string casting

hregis:
@PsyCrow-code define  the type in parameters of function and modify "dol_eval()" instead

`function dol_eval(string $s, int $returnvalue = 1, int $hideerrors = 1, string $onlysimplestring = '1')`

PsyCrow-code:
> @PsyCrow-code define the type in parameters of function and modify "dol_eval()" instead
> 
> `function dol_eval(string $s, int $returnvalue = 1, int $hideerrors = 1, string $onlysimplestring = '1')`

@hregis If we use that approach produces **TypeError**: dol_eval(): Argument #1 ($s) must be of type string, null given

the change could be in:
function dol_eval($s, $returnvalue = 1, $hideerrors = 1, $onlysimplestring = '1')
{
    $s = (string)$s;  **<-- CHANGE**

hregis:
@PsyCrow-code and where is you have a null value?

hregis:
@PsyCrow-code it's the goal to have a type error, where is you have this error?

PsyCrow-code:
> @PsyCrow-code it's the goal to have a type error, where is you have this error?

@hregis the issue rise on salary list

hregis:
@PsyCrow-code you have extrafields in user module ?

hregis:
@PsyCrow-code for me the better is to define the type before, i send a fix : https://github.com/Dolibarr/dolibarr/pull/36126

PsyCrow-code:
> @PsyCrow-code you have extrafields in user module ?

Yes, there are a native dolibarr extrafields,

PsyCrow-code:
> @PsyCrow-code for me the better is to define the type before, i send a fix : #36126

ok, thanks

hregis:
@PsyCrow-code my fix is merged, can you try with the branch 22 of github ?

PsyCrow-code:
Yes, I can try tomorrow

hregis:
@eldy @PsyCrow-code with this param, "$s" can be only a string no ? (phan or phpstan error ????)

`@param 	string		$s					String to evaluate`

eldy:
> @eldy @PsyCrow-code with this param, "$s" can be only a string no ? (phan or phpstan error ????)
> 
> `@param string $s String to evaluate`

Yes it should be a string as mentionned by doc.
But if we have in code
if (is_array($s) || $s === 'Array') {
i suppose that there is some case that use an array (even if this is not expected, it may happen sometimes).

hregis:
@eldy yes i see, if an external module use this function...

PsyCrow-code:
> > @eldy @PsyCrow-code with this param, "$s" can be only a string no ? (phan or phpstan error ????)
> > `@param string $s String to evaluate`
> 
> Yes it should be a string as mentionned by doc. But if we have in code if (is_array($s) || $s === 'Array') { i suppose that there is some case that use an array (even if this is not expected, it may happen sometimes).

I think the PR #36126 should be re-evaluated to considere the array case for $s
it is ok in case of $s is null, number or string,  but in case of array will be a missfunction

PR #36131, opened 2025-11-06
QUAL Stock movement API harmonize POST with GET

W1W1-M:
# QUAL Stock movement API harmonize POST with GET
Stock movement API POST method input parameter names have been harmonized with GET output variable names.
For example : the GET returns `inventorycode` where as the POST input for same data is `movementcode`.

This change is ultimately to help developers better apprehend the API in the long term.

Cleaned PHPDoc. 
Changelog has been updated to inform on changes.


eldy:
i think a better change would be to align in the direction of movementcode and movement label instead of inventorycode.
Because each stock movement is a movement but is not always a movement for an inventory. Inventory is also stock movement but a particular case, when we generted the movement by doing an inventory process.
Most stock movement are not inventory movement when all inventorymovement are stockmovement.

W1W1-M:
@eldy I've tested code and this is working. Changes are atomic and limited to API scope without affecting rest of codebase. Thank you for helping out with this.

PR #36135, opened 2025-11-06
Fix: sql error if LIST_OF_QUALIFIED_INVOICES_LIMIT_DEFINED is used

splohmer:
# FIX|Fix sql error if LIST_OF_QUALIFIED_INVOICES_LIMIT_DEFINED is used


htdocs/compta/facture/class/facture.class.php
```diff
@@ -5372,7 +5372,7 @@ public function list_qualified_avoir_invoices($socid = 0)
 		}
 
 		if (getDolGlobalInt('LIST_OF_QUALIFIED_INVOICES_LIMIT_DEFINED') > 0) {
-			$sql .= " ORDER BY CASE WHEN f.rowid = ".((int) GETPOST('fac_avoir'))."' THEN 0 ELSE 1 END, f.ref";
+			$sql .= " ORDER BY CASE WHEN f.rowid = '".((int) GETPOST('fac_avoir'))."' THEN 0 ELSE 1 END, f.ref";
```
eldy: f.rowid is an int so omparison should be = x and not = 'x' so i fixed by removing the (

PR #36136, opened 2025-11-06
FIX: dol_escape_htmltag in extrafields_view

RyAbn:
# FIX: Use dol_escape_htmltag() in extrafields_view

$value may contain double quotes or other HTML special characters which can break input fields by closing the value attribute prematurely and corrupt the DOM display. 

Fix: escape $value using dol_escape_htmltag() to ensure safe HTML output.


FHenry:
@RyAbn 
To check https://github.com/Dolibarr/dolibarr/pull/36127

eldy:
> @RyAbn To check #36127

@FHenry the dolPrintHTMLForAttribute does not exist yet in v18, so dol_escapte_htmltag is the only solution for v18

PR #36144, opened 2025-11-06
FIX : Can approuve holidays when negative balance

atm-adrien:
## FIX : Can approuve holidays when negative balance

Currently, it is possible to approve a leave request even if it causes the user's balance to go negative.
This should not be allowed.

### DIFF
--- a/htdocs/holiday/class/holiday.class.php
+++ b/htdocs/holiday/class/holiday.class.php
@@ -898,13 +898,8 @@ class Holiday extends CommonObject
 		if ($checkBalance > 0) {
 			$balance = $this->getCPforUser($this->fk_user, $this->fk_type);
 
-			$startDdatetime = new DateTime("@$this->date_debut");
-			$endDatetime = new DateTime("@$this->date_fin");
-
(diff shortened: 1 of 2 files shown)


htdocs/holiday/class/holiday.class.php
```diff
@@ -898,7 +898,13 @@ public function approve($user = null, $notrigger = 0)
...
 			$balance = $this->getCPforUser($this->fk_user, $this->fk_type);
 
-			if ($balance < 0) {
+			$startDdatetime = new DateTime("@$this->date_debut");
+			$endDatetime = new DateTime("@$this->date_fin");
+
+			$interval = $startDdatetime->diff($endDatetime);
+			$days = $interval->days;
+
+			if ($balance - $days < 0) {
```
eldy: I suggest to add a test on a getDolGlobalString('HOLIDAY_ALLOW_NEGATIVE_BALANCE')  for the case when company need to allow this (a lof of company does not manage the balance but need to follow leave requests.

eldy:
As we should not introduce change in business rules and behaviour into a stable version i have reerted the option so default is current behaviour.

PR #36147, opened 2025-11-06
Qual: Fix false phan positive in export.php

mdeweerd:
# Qual: Fix false phan positive in export.php

empty() has an undesired side effect, fix by casting bool on defined variable


mdeweerd:
The core issue was that GETPOST() returns `string|array` .

PR #36160, opened 2025-11-07
Fix 35017 accounts for local taxes 2

### LINKED ISSUE #35017: Add accounting ledger accounts for every taxes used in tax rates with local taxes
### Feature Request

When applicable, set two pair of ledger accounts to a tax rate that implies more than one tax rate.

### Use case

In Québec, it is required to account taxes rates individually (TVA and localtax1 must each have two accounts, one for sale other purchase).

Currently, Dolibarr only allow for one pair of account for a tax rate that have more than one tax applied.

Image

### Suggested implementation

@vdegrandpre will implement a fix based on discussions with other devs on #chat-devs-fr on Discord and document this FR.

### Suggested steps

1.  Add an attribute on the tax item to consider individual tax account for included tax rates, upgrade database as needed
2.  Add an option in the tax dictionnary line to allow selection of individual tax account for local taxes on the line
3.  When the option is activated by the user, show more select lists to the user to choose ledger tax accounts for sale and purchase
4. Update accounting reporting and tax payment consider individual taxes when applicable

vdegrandpre:
# FIX|Fix #35017 add accounting ledgers by default for local taxes
This change allow to set default account for localtaxes and take them in account when writing to accounting.

image

image

image

image

image


htdocs/accountancy/journal/purchasesjournal.php
```diff
@@ -257,13 +264,15 @@
...
-		$compta_localtax1 = (!empty($vatdata['accountancy_code_buy']) ? $vatdata['accountancy_code_buy'] : $cpttva);
-		$compta_localtax2 = (!empty($vatdata['accountancy_code_buy']) ? $vatdata['accountancy_code_buy'] : $cpttva);
+		$compta_localtax1 = (!empty($vatdata['accountancy_code_buy']) ? $vatdata['accountancy_code_buy'] : $cptlocaltax1);
+		$compta_localtax2 = (!empty($vatdata['accountancy_code_buy']) ? $vatdata['accountancy_code_buy'] : $cptlocaltax2);
 		$compta_counterpart_tva_npr = getDolGlobalString('ACCOUNTING_COUNTERPART_VAT_NPR', 'NotDefined');
 
 		// Define an array to display all VAT rates that use this accounting account $compta_tva
 		if (price2num($obj->tva_tx) || !empty($obj->vat_src_code)) {
 			$def_tva[$obj->rowid][$compta_tva][vatrate($obj->tva_tx).($obj->vat_src_code ? ' ('.$obj->vat_src_code.')' : '')] = (vatrate($obj->tva_tx).($obj->vat_src_code ? ' ('.$obj->vat_src_code.')' : ''));
+			$def_tva[$obj->rowid][$compta_localtax1][vatrate($obj->localtax1_tx).($obj->vat_src_code ? ' ('.$obj->vat_src_code.')' : '')] = (vatrate($obj->localtax1_tx).($obj->vat_src_code ? ' ('.$obj->vat_src_code.')' : ''));
```
eldy: Can you confirm that we can use the same array def_tva for localtax ?
If compta_localtax1 = compta_tva, we may erase the value in dev_tva[compta_tva].
May be we should user  def_tva_localtax1 and def_tva_localtax2 ?
We can stil merge them later if we need.

PR #36173, opened 2025-11-08
Add sortorder parameter to selectForFormsList method

defrance:
there are a problem on many extrafields list : we can't sort list by decrease and sort from a specific fields since we have add an sql prevention on filter based on : separator (who is used to separate parameters)

@eldy  i propose to remplace on extrafields parameters the : in the sql injection prevention by another caracters, by exemple the ; or the § , and replace it after having explode all the parameters


eldy:
Sorting list for extrafield with type "Select list" is not possible because feature was not implemented yet.
I dont see the relation with the separator character.
May be you can provide an example of what you mean ?

defrance:
in commonobject.class.php line 7944 we speak about sortfields for list
and line 8389 for the link extrafield type : 
			// $param_list='ObjectName:classPath[:AddCreateButtonOrNot[:Filter[:Sortfield]]]'
			// Filter can contains some ':' inside.
			$param_list = array_keys($param['options']);
			$param_list_array = explode(':', $param_list[0], 4);

but we have a limited explode trick to not use sortfield parameter (becose filter can contains ':' as explain.
so i propose to use another caracters than ':' on filter to allow following parameters like sorting

PR #36177, opened 2025-11-08
NEW Parent project column in list of projects

JonBendtsen:
# NEW Parent project column in list of projects
This PR adds the possibility of showing a parent column in the list of projects as well as making it searchable on both ref and label among parent project. If you don't have access to the parent project it will write that in the column rather than revealing the parent project ref.

This PR has also introduced a lot of id="search_..." on all the other columns to make it easier to identify the columns when you use the browsers inspect function.

### DIFF
--- a/htdocs/projet/list.php
+++ b/htdocs/projet/list.php
@@ -119,7 +119,8 @@ $search_opp_status = GETPOST("search_opp_status", 'alpha');
 $search_opp_percent = GETPOST("search_opp_percent", 'alpha');
 $search_opp_amount = GETPOST("search_opp_amount", 'alpha');
 $search_budget_amount = GETPOST("search_budget_amount", 'alpha');
-$search_parent = GETPOST('search_parent', 'alpha');
+$search_parent_ref = GETPOST('search_parent_ref', 'alpha');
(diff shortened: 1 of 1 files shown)


JonBendtsen:
You enable the column just like all the other columns
enable_parent_column

JonBendtsen:
You can search on both ref and label
project_list_search_parent_ref_and_label

JonBendtsen:
If you don't have access to the parent project, then you will see that there is a parent project, but also that you don't have access
project_list_with_parent_but_access_denied

If you click on the link you can see the project ID (in the url) and you can share that with your boss who should have access, but you are denied access
project_list_with_parent_but_access_denied

I choose this rather than revealing the parent project ref - which may have a meaning that you may or may not need to know.

JonBendtsen:
If you do have access, the project ref is listed as the parent project, and regular mouse over works and will reveal a little bit of information.
project_list_with_parent_with_mouse_over

JonBendtsen:
> This PR has also introduced a lot of id="search_..." on all the other columns to make it easier to identify the columns when you use the browsers inspect function.

This looks like this. Not all columns has an id because it would not work with them. Also the date fields does not have the correct id, with one for both start and end, but only part of the search_...

image

PR #36179, opened 2025-11-08
clean code

frederic34:


htdocs/core/actions_massactions.inc.php
```diff
@@ -72,6 +72,7 @@
  * @var ?int $hidedesc
  * @var ?int $hideref
  * @var ?array<string,mixed> $moreparams
+ * @var ?string $from
```
mdeweerd: Defining $from as a predefined value does not seem correct.
$from may not be defined because we have https://github.com/Dolibarr/dolibarr/blob/cec3c6238d144f36c2fc7c74792d0c7c8002db5a/htdocs/core/actions_massactions.inc.php#L501-L503 (`if ($obj) { $from = ... }`) in one of the conditions and then $from is not defined when using it.

mdeweerd: > :confused:
Bigger picture:
https://github.com/Dolibarr/dolibarr/blob/cec3c6238d144f36c2fc7c74792d0c7c8002db5a/htdocs/core/actions_massactions.inc.php#L487-L508

After the if/elseif sequence, `$from` may not be defined.
One method would be:

```php
$from = null;
if (...) ... else if() .. .
if ($from === null) {
    $from = GETPOST('fromname').' <'.GETPOST('frommail').'>';
}
```
or
```php
$from = GETPOST('fromname').' <'.GETPOST('frommail').'>';
if (...) ... else if() .. .
```

And remove
https://github.com/Dolibarr/dolibarr/blob/cec3c6238d144f36c2fc7c74792d0c7c8002db5a/htdocs/core/actions_massactions.inc.php#L75

PR #36180, opened 2025-11-08
clean code

frederic34:


LePat:
Side effect of the line 141 `$result = 0;` : it blocks the possibility to send mail in the "Emails setup", as there is no $object for the test mail it goes strait to the error ErrorFailedToReadObject. Removing the line `$result = 0;` fixes the bug because `$result = 1` following the inclusion of `main.inc.php` in `mails.php`, although it's clearly a good idea to set `$result` to 0.

PR #36182, opened 2025-11-08
Adding current timestamp to datec column in llx_c_email_templates

JonBendtsen:
# New Alter table llx_c_email_templates datec current timestamp
This PR alters table llx_c_email_templates setting the datec column to have a default value of CURRENT TIMESTAMP. This way if no timetamp is provided it will take the current one.

This is apparently something newish? in mariadb?
https://dev.mysql.com/doc/refman/8.0/en/timestamp-initialization.html

we probably should change more datec to this so we don't have to code it.

Even if I simulate dolibarr inserting a new using a value or the API doing it then it respects the datec value
```
INSERT INTO llx_c_email_templates (label,lang,type_template,fk_user,private,position,topic,email_from,joinfiles,defaultfortype,content,entity, active, enabled, datec) VALUES('sdfdfd', '', 'all', null, 0, 1, 'ddfd', null, 0, 0, 'dfdfd', 1, 1, 1, '2025-11-08 21:53:41')
```

### DIFF
--- a/dev/build/phpstan/phpstan-baseline.neon
+++ b/dev/build/phpstan/phpstan-baseline.neon
@@ -4183,12 +4183,6 @@ parameters:
 			path: ../../../htdocs/contrat/card.php
 
 		-
-			message: '#^Variable \$tmpcode might not be defined\.$#'
-			identifier: variable.undefined
-			count: 1
-			path: ../../../htdocs/contrat/card.php
-
-		-
 			message: '#^Variable \$mesg might not be defined\.$#'
 			identifier: variable.undefined
 			count: 1
--- a/htdocs/contrat/card.php
+++ b/htdocs/contrat/card.php
@@ -36,6 +36,15 @@
  */
 
 require "../main.inc.php";
+/**
+ * @var Conf $conf
+ * @var DoliDB $db
+ * @var HookManager $hookmanager
+ * @var Societe $mysoc
+ * @var Translate $langs
+ * @var User $user
+ */
+
 require_once DOL_DOCUMENT_ROOT.'/core/lib/date.lib.php';
 require_once DOL_DOCUMENT_ROOT.'/core/lib/price.lib.php';
 require_once DOL_DOCUMENT_ROOT.'/core/lib/contract.lib.php';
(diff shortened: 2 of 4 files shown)


JonBendtsen:
image

JonBendtsen:
Why is it though, that when we __construct any object, and that object has a public $datec, that we do not assign a value to that $datec?

# not done in variable definition
https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/core/class/cemailtemplate.class.php#L84

# not done in the __construct function
https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/core/class/cemailtemplate.class.php#L183

# not done during create
https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/core/class/cemailtemplate.class.php#L277

# but there is an unused dol_now() ?
https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/core/class/cemailtemplate.class.php#L250

hregis:
@JonBendtsen maybe the datec can be different of timestamp of sql server!

htdocs/install/mysql/migration/22.0.0-23.0.0.sql
```diff
@@ -339,4 +339,5 @@ INSERT INTO llx_c_currencies ( code_iso, unicode, active, label ) VALUES ( 'CDF'
 
 ALTER TABLE llx_societe MODIFY COLUMN mode_reglement integer;
 
+ALTER TABLE llx_c_email_templates CHANGE COLUMN datec datec timestamp DEFAULT CURRENT_TIMESTAMP;
```
eldy: We must abandon the use of current timestamp on fields in favor of using dol_now in php side. Reason is that current timestamp is done on database side, so using the timezone of the database that is completely out of control of the developer, leading to unexpected result.
You can instead add a check in create or update method of crud class to force value to dol_now().

eldy (changes requested):

htdocs/install/mysql/migration/22.0.0-23.0.0.sql
JonBendtsen: @eldy  makes good sense, and it is probably database agnostic as well

PR #36183, opened 2025-11-08
Fix:hook for adding params for adding payments/CloseBill action

bafbes:
Fix:hook for adding params for adding payments/CloseBill action


htdocs/takepos/invoice.php
```diff
@@ -1350,6 +1350,8 @@ function fail($message)
 		if ($remaintopay <= 0 && getDolGlobalString('TAKEPOS_AUTO_PRINT_TICKETS') && $action != "history") {
 			$sectionwithinvoicelink .= '<script type="text/javascript">console.log("Emulate click on #buttonprint"); $("#buttonprint").click();</script>';
 		}
+	} else {
+		$sectionwithinvoicelink = $hookmanager->resPrint;
```
eldy: This is the result of which hook ?

bafbes:
I’m sorry. The mistake happened when merging from an old version of Dolibarr.
I’ve just committed a fix.

PR #36187, opened 2025-11-09
Event Calendar

frederic34:
https://github.com/vkurko/calendar
https://vkurko.github.io/calendar/


frederic34:
image

frederic34:
Actual 
image

eldy:
Note that this is based on a JS library, so that has probably compatibility issues in future (there have already been attempts and rollbacks) because, like all JS libraries, it's no longer maintained or is poorly maintained few years after beiing introduced. 
It also doesn't handle timezones correctly because it usually relies on the Navigater's timezone and not the Dolibarr configuration. It also handles Dolibarr's own translation poorly, using its own translations, and has conflicts problems with certain other JS libraries. In short, lots of problems. But by going step by step, for example, with a hidden option initially to enable it and allow everyone to test, stabilize, and correct the mentioned problems, may be (low probability but no zero). But to reach a standard/stable version, it will need real added value that we can't provide in the standard version because external libraries are a nightmare (and often the death of a feature of a projects). 

Currently it seems that the value added is having the rendering grouped on hours shown on left. Such a feature could be introduced in our current calendar. Also all existing features (information on thumbs, filters, drag and drop, adding leaves in view, adding birthdays, adding external agenda), would have to be introduced.

Relying on a lib that may be abandonned or not compatible with rest of js code in few years is a high risk. And there is still a high number of problem to solve to have replacement in standard version to replace current calendar (clickable link in event, drag and drop, color manager according to type or user, including leaves en employees, birthdays, external agendas with ics links are all features that we support natively, etc... only group of thumb per hours seems not yet implemented). This is why, for the moment, such feature is better as an external module until the level of feature reach same level of current feature.

hregis:
@frederic34 @eldy It's either we create an input/output framework that allows us to switch libraries effortlessly... or we develop our own JavaScript libraries: "Vanilla JavaScript"

PR #36193, opened 2025-11-09
NEW stock API GET movement

W1W1-M:
# NEW stock API GET movement
Get an individual stock movement using ID

image


defrance:
Hello,
we can use the native stockmovement call with the filter associated but some new api is welcome ;-)

W1W1-M:
> Hello,
> 
> we can use the native stockmovement call with the filter associated but some new api is welcome ;-) 

Thank you for suggestion/solution. Using sql filter in this way seems a bit like a workaround but I agree it is possible.

Adding this individual GET was also to align/harmonize the stock movement API with other API objects.

PR #36194, opened 2025-11-09
FIX wrong path of libraries

hregis:


hregis:
@eldy  i send a fix for the travis error : https://github.com/Dolibarr/dolibarr/pull/36195

hregis:
@eldy  why not my fix ? https://github.com/Dolibarr/dolibarr/pull/36195

PR #36195, opened 2025-11-09
FIX travis error

hregis:


hregis:
@eldy it's not good ?

eldy:
yes it is. But conflict with another fix already done and merged. So this one became deprecated.

PR #36203, opened 2025-11-10
FIX re-create API temp dir after purging temp files

lvessiller-opendsi:
FIX re-create API temp dir after purging temp files

**Before**
When purge files is launched and Restler API is enabled and multicompany is enabled :
- API temp directory is deleted for all entities and only re-created for main entity [entity=1]
So when we try to create a third-party for an another entity, the request return a "Not allowed" response.
However the third-party is created and the result of request is not the "id" of created third-party.

The problem is the "routes.php" file for Restler API is not created for the entity (example : "2/api/temp" is not created for entity 2).

**To reproduce**
Enable multicompany and Rest API in entity 2.
Delete "2/api/temp" directory using purge files process or manually.

Try to create a third-party on entity 2.
curl --location --request POST 'DOLIBARR_URL/api/index.php/thirdparties' \
--header 'DOLAPIKEY: USER_API_KEY' \
--header 'DOLAPIENTITY: ENTITY' \
--header 'Content-Type: application/json' \
--data '{
  "entity": ENTITY,
  "name": "TEST Toto",
  "name_alias": "",
  "email": "test@test.test",
  "phone": "",
  "tva_assuj": "1",
  "tva_intra": "",
  "typent_id": "8",
  "lastname": "TEST",
  "firstname": "Toto",
  "address": "Rue du test",
  "zip": "00001",
  "town": "AIN",
  "country_id": "1",
  "country_code": "FR"
}'

**After**
The purge files re-create all API (if API is enabled and multicompany is enabled) temp directories after puging all temp directories and files.

So the "routes.php" file can be regenerated in API temp directory when we call the API with requests.

### DIFF
--- a/htdocs/api/class/api.class.php
+++ b/htdocs/api/class/api.class.php
@@ -58,6 +58,18 @@ class DolibarrApi
 
 		$this->db = $db;
 		$production_mode = (empty($conf->global->API_PRODUCTION_MODE) ? false : true);
+
+		if ($production_mode) {
+			// Create the directory Defaults::$cacheDirectory if it does not exist. If dir does not exist, using production_mode generates an error 500.
+			include_once DOL_DOCUMENT_ROOT.'/core/lib/files.lib.php';
+			if (!dol_is_dir(Defaults::$cacheDirectory)) {
+				dol_mkdir(Defaults::$cacheDirectory, DOL_DATA_ROOT);
+			}
+			if (getDolGlobalString('MAIN_API_DEBUG')) {
+				dol_syslog("Debug API construct::cacheDirectory=".Defaults::$cacheDirectory, LOG_DEBUG, 0, '_api');
+			}
+		}
+
 		$this->r = new Restler($production_mode, $refreshCache);
 
 		$urlwithouturlroot = preg_replace('/'.preg_quote(DOL_URL_ROOT, '/').'$/i', '', trim($dolibarr_main_url_root));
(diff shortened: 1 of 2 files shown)


hregis:
@lvessiller-opendsi Perhaps the REST API module should be made a core module? It should only be activated on the master entity, and if activated, then it should be activated for all entities?

lvessiller-opendsi:
> @lvessiller-opendsi Perhaps the REST API module should be made a core module? It should only be activated on the master entity, and if activated, then it should be activated for all entities?

I don't know if it will be a good pratice to enable the API for all entities and it's not the problem here.
The problem here is the purge files process and delete all temp API directories and only re-create for the main entity (not for all entities).

eldy:
A recent fix was done into develop branch to create the temp dir automatically when the code need it so there is no more need to create them.
I think you could backport it instead.

See line 63 to 70 of api.class.php

lvessiller-opendsi:
> A recent fix was done into develop branch to create the temp dir automatically when the code need it so there is no more need to create them. I think you could backport it instead.
> 
> See line 63 to 70 of api.class.php

I did it.
There is Travis error but I don't know why.

eldy:
I don't know either
I will wait few days to rerun travis later. May be pb is on travis tool

lvessiller-opendsi:
> I don't know either I will wait few days to rerun travis later. May be pb is on travis tool

CI checks succeeded

Issue #36201, opened 2025-11-10
Photo of member not show in web portal


hregis:
### Bug

<img width="1110" height="236" alt="Image" src="https://github.com/user-attachments/assets/5ac7791d-d227-441f-810e-65768ec7b532" />

### Dolibarr Version

22.0.x

hregis:
@eldy @frederic34 maybe record the photo in ECM table with a share field ?

PR #36205, opened 2025-11-10
display member photo with hash

frederic34:
https://github.com/Dolibarr/dolibarr/issues/36201

alse need fix https://github.com/Dolibarr/dolibarr/commit/9d70bcbf247e69dcb9c8b64c36cb6bec019a1d26

image

### DIFF
--- a/htdocs/adherents/card.php
+++ b/htdocs/adherents/card.php
@@ -444,6 +444,7 @@ if (empty($reshook)) {
 								} else {
 									// Create thumbs
 									$object->addThumbs($newfile);
+									addFileIntoDatabaseIndex($dir, $object->photo, '', 'uploaded', 1, $object);
 								}
 							}
 						} else {
--- a/htdocs/adherents/card.php
+++ b/htdocs/adherents/card.php
@@ -444,7 +444,6 @@ if (empty($reshook)) {
 								} else {
 									// Create thumbs
 									$object->addThumbs($newfile);
-									addFileIntoDatabaseIndex($dir, $object->photo, '', 'uploaded', 1, $object);
 								}
 							}
 						} else {
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -46,7 +46,6 @@
  * \brief      File of class with all html predefined components
  */
 
-use Artis\Dolibarr\Models\EcmFile;
 
 /**
  * Class to manage generation of HTML components
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -46,6 +46,7 @@
  * \brief      File of class with all html predefined components
  */
 
+use Artis\Dolibarr\Models\EcmFile;
 
 /**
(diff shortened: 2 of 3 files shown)


hregis:
@frederic34 can you remove the modification of member card, my fix uniformize the code https://github.com/Dolibarr/dolibarr/pull/36197/files

hregis:
@frederic34 @eldy petite question, on met l'image d'origine avec un "share" dans ECM, mais doit on référencer aussi les "thumbs" mini et small afin d'y avoir accès en share aussi ?

eldy:
> @frederic34 @eldy petite question, on met l'image d'origine avec un "share" dans ECM, mais doit on référencer aussi les "thumbs" mini et small afin d'y avoir accès en share aussi ?

Good question.
Having access to a small image and not the original or the opposite has no sense in a security point of view. This is why we must index only the main image and sharing status is always on main mage too. We should not have thumbs inside the llx_ecm_files.
If main image is shared, we can access all versions. If not, none.

hregis:
@frederic34 getAdvancedPreviewUrl() not manage the share (hashp) url

hregis:
> > @frederic34 @eldy petite question, on met l'image d'origine avec un "share" dans ECM, mais doit on référencer aussi les "thumbs" mini et small afin d'y avoir accès en share aussi ?
> 
> Good question. Having access to a small image and not the original or the opposite has no sense in a security point of view. This is why we must index only the main image and sharing status is always on main mage too. We should not have thumbs inside the llx_ecm_files. If main image is shared, we can access all versions. If not, none.

@eldy Yes, but for the web portal, for example, we have to use a hash to access the member's photo. So, we need to modify the displayed size using CSS?

eldy:
> > > @frederic34 @eldy petite question, on met l'image d'origine avec un "share" dans ECM, mais doit on référencer aussi les "thumbs" mini et small afin d'y avoir accès en share aussi ?
> > 
> > Good question. Having access to a small image and not the original or the opposite has no sense in a security point of view. This is why we must index only the main image and sharing status is always on main mage too. We should not have thumbs inside the llx_ecm_files. If main image is shared, we can access all versions. If not, none.
> 
> @eldy Yes, but for the web portal, for example, we have to use a hash to access the member's photo. So, we need to modify the displayed size using CSS?

A hash must always be the one of the main image and the wrapper must be able to check in main image even if the url is the one of a thumbs.

But here problem is not on sharing or not: The main pb is that the webportal is usung the wrapper viewimage of the backoffice when the webportal should work when hosted on a different virtual host, so should use its own viewimage.php file in dir public/webportal. And this file does not exist ! This is a missing or conception pb. This wrapper dedicated to webportal, if it exists, could replace the wrapper of backoffice to make the security test according to webportal setup. So if feature to show member card on webportal is on, and logged user is the member, viewimage of webportal should accept to serve the image. This wrapper should not care about the sharing or not. 
So fixing the trouble of image for member should be done by adding an image wrapper for webportal.
Note that this should be true for all files too. Using same wrapper (for image or documents) for webportal and backoffice is a security misconception error.

Issue #34138, opened 2025-05-13
Feature Request - display the label instead of the id for the fiscal year - Accounting module


mikygee:
### Feature Request

Version: 21.0.1

Hello,

When I create a new période fiscale / fiscal year in the accounting module

![Image](https://github.com/user-attachments/assets/aad14cfe-3ded-440f-bd2e-d5bbb2938e5d)

The label should be displayed instead of the id

![Image](https://github.com/user-attachments/assets/11e38653-ad40-4303-bd96-e0d801fa01a2)

I know it's cosmetic but that can be improved.

Also, I notice that one element is not translated in edition mode

![Image](https://github.com/user-attachments/assets/a3671166-f5bf-42c6-8925-fd4764d4eb61)

And in display mode

![Image](https://github.com/user-attachments/assets/e927cd39-56fe-42c0-94e1-e46916ba763b)

mikygee:
The request is still relevant

aspangaro:
Already correct in v23

PR #36210, opened 2025-11-10
[UIUX] Displaying the label instead of the id for the fiscal year

marcellintacite:
# UIUX| #34138  Display fiscal year label instead of ID in Accounting module
This Pull Request addresses a user experience issue within the Accounting module where the Fiscal Year was being displayed using its internal database ID, which is not user-friendly.


aspangaro:
Okay for me, but why add the ID display to the card then? It's not necessary, you're right to see id of tbe fiscal year

PR #36215, opened 2025-11-11
Update redirect URL in openid_connect_get_redirect_url

kubajznik:
# FIX

Redirect URL for openIDConnect in step 2 (code to token exchange) must match the one used in initial request.


eldy:
I check on v21 and develop, the url for redirect is 
image
So it is defined to the answer of openid_connect_get_redirect_url().
When i am on logon page, the link OpenId connect contains the redirect_uri that is also the content return by openid_connect_get_redirect_url()

So i don't see where is the URL "?openid_mode=true". If we have it somewhere, a better fix would be to replace this link with the value of openid_connect_get_redirect_url() so it will always match ?

kubajznik:
I am sorry. I realized I still had set the MAIN_AUTHENTICATION_OPENID_URL from previous dolibarr version. That was overriding the auto-generated. Wokring now.

PR #36219, opened 2025-11-11
FIX use $this->db instead $db

hregis:


htdocs/core/class/html.form.class.php
```diff
@@ -11051,7 +11050,7 @@ public function showbarcode(&$object, $width = 100, $morecss = '')
 	 */
 	public static function showphoto($modulepart, $object, $width = 100, $height = 0, $caneditfield = 0, $cssclass = 'photowithmargin', $imagesize = '', $addlinktofullsize = 1, $cache = 0, $forcecapture = '', $noexternsourceoverwrite = 0)
```
frederic34: static

hregis: @frederic34 arf ! sorry , thanks

PR #36225, opened 2025-11-11
FIX conflict between $user->id and $object->id

hregis:


htdocs/user/messaging.php
```diff
@@ -86,18 +86,13 @@
 // Initialize technical object to manage hooks of page. Note that conf->hooks_modules contains array of hook context
 $hookmanager->initHooks(array('agendathirdparty', 'globalcard'));
 
-// Security check
-if ($user->id) {
```
eldy: I don't think we must remove this but move it before the 
```
// Initialize technical objects
$object = new User($db);
if ($id > 0 || !empty($ref)) {
```

So an external user will always be forced to himself

hregis:
@eldy try in user "event/agenda" card to switch between "/user/agenda.php" and "/user/messaging.php" before and after my fix

with a login user from an another user card

hregis:
@eldy ok i see the problem... the link to see the messaging have just "userid" and not the current id of user

hregis:
@eldy cann you check this new fix

PR #36240, opened 2025-11-13
add gps position for files

rycks:
That PR add geo position for files


frederic34:
why not using extrafields ?

PR #36250, opened 2025-11-14
FIX $param is the 5th parameter of print_liste_field_titre ($moreparam)

hregis:


hregis:
@frederic34 @mdeweerd it's possible to check this with phan or other ?

mdeweerd:
> @hregis d it's possible to check this with phan or other ?

phan does make a check to verify that if the variable is named '$moreparams' that it is in the correct position.
However here, the parameter variable is '$params' and the functions parameter name is '$moreparams'.

Naming both $params or $moreparams would result in a notice from Phan if the position is incorrect.

If you see that most implementations use '$params' as the variable name, it may be a good idea toname the parameter $param in the function 'print_liste_field_titre'

mdeweerd:
@hregis 
Comme il y en a un qui a été oublié, j'ai tenté la modification de la signature de la fonction print_liste_field_titre .
Ce qui relève quelques cas - voir #36261 .

PR #36274, opened 2025-11-15
Fix hurl assert for API warehouses to new reply from Dolibarr

JonBendtsen:
This PR fixes hurl assert for API warehouses to new reply from Dolibarr


JonBendtsen:
@mdeweerd  the latest commit here uses jsonpath to test the output and I think that can handle when there is a "debug" in the output

PR #36285, opened 2025-11-17
Accountancy - Resolve some problem on new function "Discount in accountancy"

aspangaro:
Fix three points:
- Discount use wrong function (getSumOfCreditNotesUsed > getSumCreditNotesUsed) & problem HTML injection on label
- Correct already / not yet function - Piece_num is not enough strong and can be edited !
- Use closing date rather than the invoice date


htdocs/accountancy/class/accountingjournal.class.php
```diff
@@ -1430,7 +1438,7 @@ public function writeIntoBookkeeping(User $user, &$journal_data = array(), $max_
 							$bookkeeping->subledger_label = $line['subledger_label'];
 							$bookkeeping->numero_compte = $line['numero_compte'];
 							$bookkeeping->label_compte = $line['label_compte'];
-							$bookkeeping->label_operation = $line['label_operation'];
+							$bookkeeping->label_operation = dol_string_nohtmltag($line['label_operation']);
```
eldy: $bookkeeping->label_operation is a data in memory. We should not have a formatting function on it.
If we use the data later as html output, the html escaping must be done when the rendering is done, not in assignement in memory.

PR #36308, opened 2025-11-18
FIX Move 'holiday' from old path array to new one

developmentOYR:
Issue Number: [*issue_number]

Description of the Issue
When a file is uploaded to a newly created day-off (holiday) record, that file is incorrectly displayed in other day-off records where the last two digits of the ID are identical to the record where the file was uploaded.

Root Cause Analysis (Current Behavior)
The current file system structure uses a hierarchy based on the ID, which causes collisions.

1. Current Folder Structure (Based on ID Digits):

When a day-off is created with ID 645 and a temporary reference (prov)445, the folder structure is: documents/holiday/<ID-second-to-last-digit>/<ID-last-digit>/<file-name>

Example:

documents/holiday/5/4/(prov)445-test.txt
The 5 and 4 come from ID 645.

2. The Collision:

If a user later accesses a day-off with ID 545, the code retrieves files from the folder path: documents/holiday/5/4/. Since the folder is based on the last two digits (45), it incorrectly shows the file uploaded to the record with ID 645.

Comparison with Other Objects (Invoice Example)
Other objects successfully use the full reference in the file path, avoiding ID-based collisions.

Supplier Invoice (ID 192, Reference SI2511-5736):

documents/facture/2/9/SI2511-5736/<files-related-to-invoice>
Proposed Solution (Using Reference in Folder Name)
To resolve this, we must ensure the folder structure includes the unique reference, not just the ID digits.

1. Proposed Folder Structure:

If we wish to keep the ID digits in the path, we must add a folder for the unique reference:

documents/holiday/5/4/(prov)445/(prov)445-test.txt
2. Renaming on Validation:

When the day-off is validated, the temporary reference folder should be renamed to the final, permanent reference.

Current Renaming Bug:

The current code attempts to rename directly in the path:

documents/holiday/(PROV446)/(PROV446)-test.text  
...and replaces it with the final reference:

documents/holiday/HL2512-0398/HL2512-0398-test.text

Implemented Decision
Due to the complexity and existing bugs in the ID-based structure and the renaming process, the definitive system adopted is the simplified, reference-based file system:

documents/holiday/ref
This ensures files are uniquely separated by their final day-off reference.


eldy:
Your description of PR is on holiday,but the PR make a change on supplier invoice. Can you check ?

developmentOYR:
Sorry Eldy, I was confused. You are absolutely right, it is already corrected.

hregis:
@developmentOYR i think you can remove "supplier_invoice", it's not used with get_exdir()
just "invoice_supplier" is used

developmentOYR:
> @developmentOYRCreo que puedes eliminar "supplier_invoice", ya que no se usa con get_exdir(), solo se usa "invoice_supplier".

Let's see if Eldy confirms it to us

PR #36311, opened 2025-11-18
NEW|New -  add drag-and-drop functionality for managing draft expedition lines

atm-corentin:
# NEW|New *add drag-and-drop functionality for managing draft expedition lines*

We keep the standard drag-and-drop workflow (ajaxrow.tpl, tdlineupdown column, /core/ajax/row.php), so the user experience matches what you get on quotes/orders/invoices.                                                      
                                                                                                                                                                                          

1.    Shipments have one peculiarity: a single visible line can represent several llx_expeditiondet rows (multi-warehouse/lot). We therefore override Expedition so that reordering a line updates every sub-line sharing the same    fk_origin_line, and fetch_lines() now sorts by ed.rang to mirror the stored order. row.php instantiates Expedition whenever table_element_line=expeditiondet, letting the existing AJAX flow trigger these overrides without any frontend changes.


atm-corentin:
@Eldy, I don't know how to fix that because it's an override and the default method doesn't have a type

hregis:
@atm-corentin try to replace

`@param array $rows`

by

`@param int[] $rows`

atm-corentin:
@eldy thanks it's good now

atm-quentin:
i made a modification @eldy : fixes shipment line updates for origin-based shipments and aligns line extrafield rendering with the standard order/proposal layout.

  Main changes:

  - render shipment line extrafields inline in the description cell, like orders/proposals
  - fix line update flow on shipment cards created from source documents by restoring the correct origin context from the loaded shipment (origin_type / origin_id)
  - replace a hardcoded warning message with translated language keys
  - fix a generic core bug in geo extrafields persistence where SQL used options_* instead of the actual column name for geodata fields

  Result:

  - quantity and line extrafields now save correctly on shipment lines
  - geo extrafields also persist correctly
  - shipment line display is now consistent with the standard Dolibarr UX

PR #36316, opened 2025-11-18
enhance actioncomm reminder

frederic34:


eldy:
Can you describe what is goal of the 2 new columns ?

frederic34:
The llx_actionreminder table in Dolibarr is used to manage reminders for actions or events. By adding the foreign keys fk_soc (company ID) and fk_contact (contact ID), you enable the system to associate reminders not only with users but also with specific companies (société) or contacts linked to the event. This enhancement allows for more granular and targeted reminder configurations.

PR #36322, opened 2025-11-19
NEW Allow creating contact via api with ISO code

kubajznik:
# NEW ISO Country Code in contact creation API

External api users do not know internal dolibarr country id. Better to let them use ISO country code.


htdocs/societe/class/api_contacts.class.php
```diff
@@ -343,6 +343,20 @@ public function post($request_data = null)
 		// Check mandatory fields
 		$result = $this->_validate($request_data);
 
+		// External api user does not know internal country ID
+		if (!isset($request_data['country_id']) && isset($request_data['country_code'])) {
+			$country_code = $this->db->escape($request_data['country_code']);
```
eldy: You should use dol_getIdFromCode() to get the country id from the ise code.

Issue #30465, opened 2024-07-25
Can't create user without admin account (v15)


Hortiks:
### Bug

There is a problem with the creation of users using the api of Dolibarr v15 (Code error 401 with no admin account, BUT with permissions)
The problem comes from the function that checks users authorizations on api_users.class.php.

The function is
`// Check user authorization
		if (empty(DolibarrApiAccess::$user->rights->user->creer) && empty(DolibarrApiAccess::$user->admin)) {
			throw new RestException(401, "User creation not allowed for login ".DolibarrApiAccess::$user->login);
		}`
		
and should be
`// Check user authorization
		if (empty(DolibarrApiAccess::$user->rights->user->user->creer) && empty(DolibarrApiAccess::$user->admin)) {
			throw new RestException(401, "User creation not allowed for login ".DolibarrApiAccess::$user->login);
		}`
		
Thanks to TC_L from the forum Dolibarr that found the solution :
link to the french topic : [https://www.dolibarr.fr/forum/t/pourquoi-je-ne-peux-pas-creer-dutilisateur-malgre-les-droits/46995](url)

### Dolibarr Version

v15.0.0

### Environment PHP

7.3.12

### Environment Database

Mariadb 10.4.10

### Steps to reproduce the behavior and expected behavior

Try to create a new user using api with a non admin account.
Even if your non admin account has permissions, you'll get Code error 401.

williamdes:
Fixed in 78f90982c475e6cd2d8a1cb94f7e5b7911da456f by #36337

I ran `sed -i "s/hasRight('user', 'creer')/hasRight('user', 'user', 'creer')/" user/class/api_users.class.php` in Dolibarr+Docker v22.0.3

williamdes:
@eldy what is the LTS policy  ?
If I read https://wiki.dolibarr.org/index.php/Versions correctly it would be v18
Should I backport the permissions fix back to the v18 branch so you merge it up to v22 ?

williamdes:
Hello @rycks 

Here is the issue that can be backported
See my comment above for the commit to cherry-pick: https://github.com/Dolibarr/dolibarr/issues/30465#issuecomment-3794772888

rycks:
@williamdes yes if you make a PR on dolibarr 18.0 then this one will be merged into 19....21

williamdes:
[...]

PR #36337, opened 2025-11-19
Fix: API users/groups + fix hurl tests for those

JonBendtsen:
# FIX API users/groups + fix hurl tests for those 2 API
This PR fixes some permissions that had changed names. It also enables a check if the post has a login.

The hurl test has been modified in one test such that it can possibly handle if there is a "debug" in the json output.

```
JonSweet16:hurl jonbendtsen$ ./run.sh users

----- Run hurl test on APIs ---
::notice::1. Running tests (API,GUI,public) that do not require authentication
::notice::Using existing DOLAPIKEY.
::notice::2.a. Running API tests that do require authentication
Success api/users/10_groups.hurl (15 request(s) in 1837 ms)
Success api/users/10_users.hurl (17 request(s) in 1876 ms)
--------------------------------------------------------------------------------
Executed files:    2
Executed requests: 32 (17.0/s)
Succeeded files:   2 (100.0%)
Failed files:      0 (0.0%)
Duration:          1880 ms
```

### DIFF
--- a/htdocs/user/class/api_users.class.php
+++ b/htdocs/user/class/api_users.class.php
@@ -349,10 +349,10 @@ class Users extends DolibarrApi
 		// check mandatory fields
 		if (!isset($request_data["login"]))
 			throw new RestException(400, "login field missing");
-		if (!isset($request_data["password"]))
+		/*if (!isset($request_data["password"]))
 			throw new RestException(400, "password field missing");
 		if (!isset($request_data["lastname"]))
-			 throw new RestException(400, "lastname field missing");
+			 throw new RestException(400, "lastname field missing");*/
 
 		//assign field values
 		foreach ($request_data as $field => $value) {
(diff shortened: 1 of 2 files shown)


hregis:
@JonBendtsen @mdeweerd The API should not allow the creation of an object for one entity from another entity. `DOLAPIENTITY` must define the current entity. If you want to create an object for a different entity, you must modify `DOLAPIENTITY` on the client querying the API.

hregis:
@JonBendtsen @mdeweerd send an "entity" parameter in API function is not recommended ! current entity id is defined by DOLAPIENTITY in header curl request...
only API for users and groups can accept "entity" parameter to manage the transverse mode, when the users are managed in master entity, but only for the superadmin. superadmin = ($user->admin == 1 && $user->entity == 0)

williamdes:
Fixes https://github.com/Dolibarr/dolibarr/issues/30465

PR #36345, opened 2025-11-20
NEW Add country ID resolution from country code for thirdparty api

kubajznik:
NEW Add country ID resolution from country code for thirdparty api
NEW add extrafield support when creating thirdparty via API


kubajznik:
Not sure, why php-stan and php-phan are still failing.

PR #36346, opened 2025-11-20
# FIX|Fix - Update to the MRP Efficiency Loss Calculation

atm-corentin:
# FIX|Fix #*Update to the MRP Efficiency Loss Calculation*

**Explanation of the issue and the proposed fix**

In the MRP module, the current calculation of raw-material consumption when a production efficiency lower than 1 is defined uses the formula:
BOM_qty × quantity_to_produce / efficiency.
This implicitly interprets efficiency as output = input × efficiency, which generates non-rounded and unintuitive consumption values (e.g., 53,333.33 µL for a 10% loss). As a consequence, stock levels on consumed lots rarely reach zero, creating discrepancies and making inventory harder to manage.

In real manufacturing practice, users commonly interpret efficiency as a loss rate: an efficiency of 0.9 means a 10% loss, so the consumed quantity should increase by 10%. The correct expected formula is therefore:
BOM_qty × quantity_to_produce × (2 − efficiency)
For example, 48,000 µL × (2 − 0.9) = 52,800 µL, which accurately reflects a 10% loss added on top of the nominal required quantity.

The proposed patch replaces the division-based calculation with this loss-factor approach, resulting in more realistic consumption quantities and cleaner stock movements, avoiding artificial leftovers. The related UI display has also been updated to clearly show the new calculation logic to users.

### DIFF
--- a/htdocs/mrp/class/mo.class.php
+++ b/htdocs/mrp/class/mo.class.php
@@ -763,7 +763,10 @@ class Mo extends CommonObject
 							if ($line->qty_frozen) {
 								$moline->qty = $line->qty; // Qty to consume does not depends on quantity to produce
 							} else {
-								$moline->qty = (float) price2num(($line->qty / (!empty($bom->qty) ? $bom->qty : 1)) * $this->qty / (!empty($line->efficiency) ? $line->efficiency : 1), 'MS'); // Calculate with Qty to produce and  more presition
+								$efficiency = !empty($line->efficiency) ? $line->efficiency : 1;
+								$lossfactor = max(0, 2 - $efficiency); // Factor is 1 when efficiency is 1, >1 when efficiency <1
+								$qtyperbom = $line->qty / (!empty($bom->qty) ? $bom->qty : 1);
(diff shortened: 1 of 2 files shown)


htdocs/mrp/class/mo.class.php
```diff
@@ -763,7 +763,10 @@ public function createProduction(User $user, $notrigger = 0)
 							if ($line->qty_frozen) {
 								$moline->qty = $line->qty; // Qty to consume does not depends on quantity to produce
 							} else {
-								$moline->qty = (float) price2num(($line->qty / (!empty($bom->qty) ? $bom->qty : 1)) * $this->qty / (!empty($line->efficiency) ? $line->efficiency : 1), 'MS'); // Calculate with Qty to produce and  more presition
+								$efficiency = !empty($line->efficiency) ? $line->efficiency : 1;
+								$lossfactor = max(0, 2 - $efficiency); // Factor is 1 when efficiency is 1, >1 when efficiency <1
```
eldy: If efficiency is 0.1 (10%), it means we must have a quantity of 10 if we cant to get 1.
This is what we got with old code.

With new code, if we have an efficicency of 10% only, the planned value to consume to get 1 is 1.9 ???

Before, making a change on this, thanks to introduce a phpunit test with sample of value with expected result that show that result is wrong.

PR #36372, opened 2025-11-22
Qual: detect malformed DOLAPIKEY when run.sh hurl tests

JonBendtsen:
# Qual: detect malformed DOLAPIKEY when run.sh hurl tests
This PR detects a malformed DOLAPIKEY environment variable as well as when none is set.

```
  549  test/hurl/run.sh 
  550  test/hurl/run.sh --apikey=************************
  552   DOLAPIKEY=************************ test/hurl/run.sh 
  553  test/hurl/run.sh --apikey=************************
  554   DOLAPIKEY='DOLAPIKEY: ************************' test/hurl/run.sh
  561   DOLAPIKEY='' test/hurl/run.sh
  567  test/hurl/run.sh --apikey='DOLAPIKEY: ************************'
```

`run.sh` no longer tries to run API tests requiring a key when
- key has the wrong format
- no/empty key and no login/password is available

And `run.sh` does run API tests requiring a key when
- key is supplied as just key
- key is supplied as the whole header to sent
- key is supplied using environment variable in the correct format

Empty variable results in
`INFO: DOLAPIKEY bash variable is unset/empty, no API tests that require authentication`

Wrong format variable results in
```
ERROR: Environment variable DOLAPIKEY has the wrong format: 'sdssd'
ERROR: should perhaps be: 'DOLAPIKEY: sdssd'
``´

### DIFF
--- a/test/hurl/run.sh
+++ b/test/hurl/run.sh
@@ -233,9 +233,14 @@ get_dolibarr_api_key() {
 
 	# If key is already set, return it
 	if [ -n "${current_key}" ]; then
-		print_info "Using existing DOLAPIKEY."
-		echo "${current_key}"
-		return 0
+		if [[ "${DOLAPIKEY}" != *": "* ]]; then
+			print_error "Environment variable DOLAPIKEY has the wrong format: '${DOLAPIKEY}'"
+			print_error "should perhaps be: 'DOLAPIKEY: ${DOLAPIKEY}'"
+		else
+			print_info "Using existing DOLAPIKEY."
+			echo "${current_key}"
+			return 0
+		fi
 	fi
 
 	# Check if credentials are available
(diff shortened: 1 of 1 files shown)


JonBendtsen:
@mdeweerd FYI

mdeweerd:
I had added code to fetch a key when the username and password is provided.  I think this is still the case, so possibly the description is like:

`run.sh` does not run API tests requiring an API key when
- key has the wrong format;
- no/empty key and no login/password is available.

JonBendtsen:
> I had added code to fetch a key when the username and password is provided. I think this is still the case, so possibly the description is like:
> 
> `run.sh` does not run API tests requiring an API key when
> 
> * key has the wrong format;
> * no/empty key and no login/password is available.

yes, your code is still there, I will change the description

PR #36378, opened 2025-11-23
FIX Invoice - Problem with deposit line on Type of operation

aspangaro:
Ignore "Deposit" line to calculate type of operation.
It often incorrectly displays “mixed” when you are on a service-type invoice with a deposit.

Before
image

After
image


rycks:
@eldy may i do a backport into 18.0 for that fix ?

eldy:
> @eldy may i do a backport into 18.0 for that fix ?

You can.
The best method is to try to do a cherry-pick of the commit into v18.

eldy:
> @eldy may i do a backport into 18.0 for that fix ?

You can.
The best method is to try to do a cherry-pick of the commit into v18.

rycks:
> You can. The best method is to try to do a cherry-pick of the commit into v18.

ok thanks :)

i try to do it but octopus does not exists in 18.0 and cherry pick fails, so i do a copy/paste on files ... here is the PR : https://github.com/Dolibarr/dolibarr/pull/36541

Issue #34852, opened 2025-07-31
User need permission 'extended to all third obects' to read his own event


superdupont-fr:
### Bug

An user (internal) can add event, assigned to him, but can't see this event in the calendar without the right "_Extend access to all third parties AND their objects (not only third parties for which the user is linked as a sale representative)._"

If the user have all but this right, he can't see his own event.
It's a problem, because I don't want some user can see event assigned to other users.

### Dolibarr Version

21.0.1

### Environment PHP

php8*

### Environment Database

MariaDB

### Steps to reproduce the behavior and expected behavior

- Connect to the user, open calendar page.
- Add an event, and validate

==> The event isn't visible.

- With the admin user, allow the user or his group the permission "_Extend access to all third parties AND their objects (not only third parties for which the user is linked as a sale representative)._"
- The user can reload the page

==> Now, his event is visible, but all other event too. (of course, this is the permission for this)

JonBendtsen:
v22.0.3

Shows up on the front page

<img width="718" height="423" alt="Image" src="https://github.com/user-attachments/assets/d5108f49-e9af-4c00-bf90-e644467e7c64" />

but not here `comm/action/list.php`

<img width="1422" height="531" alt="Image" src="https://github.com/user-attachments/assets/ef4b9d55-8479-40cb-8c9d-322fa88aeec5" />

JonBendtsen:
seems like version 18 is working :-) (though with a databased restored from a v20)

<img width="666" height="358" alt="Image" src="https://github.com/user-attachments/assets/af3b2070-1013-40df-9dab-ff22eddeb6d4" />

JonBendtsen:
v19 seems to work as well (with database restored from v20 backup)

<img width="285" height="127" alt="Image" src="https://github.com/user-attachments/assets/a1c49f1b-411f-4b4e-9d75-2c744dc79a9a" />

JonBendtsen:
v20 fails, can not see anything

<img width="1376" height="383" alt="Image" src="https://github.com/user-attachments/assets/7869a8e2-0201-42ec-a258-3b5556ba2db5" />

JonBendtsen:
@eldy  I think your commit here https://github.com/Dolibarr/dolibarr/commit/910a541c60bd56268d92ea78ebda031e569fe499 introduced the check that causes this error.

JonBendtsen:
[...]

PR #36385, opened 2025-11-23
FIX: User need perm. 'extended to all third objects' to read own event #34852

JonBendtsen:
# FIX User need perm. 'extended to all third objects' to read own event
This PR is a partial fix to #34852 by including agenda events which does not have any thirdparty assigned.

The user can still see the thirdparty in the card view if they know the direct link. Perhaps this should be fixed

### DIFF
--- a/htdocs/comm/action/list.php
+++ b/htdocs/comm/action/list.php
@@ -513,7 +513,7 @@ if ($search_sale && $search_sale != '-1') {
 	if ($search_sale == -2) {
 		$sql .= " AND NOT EXISTS (SELECT sc.fk_soc FROM ".MAIN_DB_PREFIX."societe_commerciaux as sc WHERE sc.fk_soc = a.fk_soc)";
 	} elseif ($search_sale > 0) {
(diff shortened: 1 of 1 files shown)


JonBendtsen:
# no thirdparty assigned to these events

v20_fix_agenda_view

JonBendtsen:
partially fixed, as one can still see a card in card view even though a thirdparty has been assigned to it

image

JonBendtsen:
# agenda with a thirdparty not assigned to this user not shown

agenda with thirdparty not shown

JonBendtsen:
# if I assign the user as sales representative, then the user can see the agenda - as expected

agenda view with user assigned to thirdparty

JonBendtsen:
here's what branch 20.0 your small fix looks like

v20_with_small_fix_agenda

I think we should close this debate now with your small fix - and I think that long term we should some how start being able to hide thirdparties that users are not being allowed to see. This includes not just the orders, address, contacts, ... but any information.

Issue #36396, opened 2025-11-24
Add ability to list template invoices in REST API


LeGmask:
### Feature Request

I would like to be able to list template invoices using the REST API.

### Use case

I use the REST API to integrate with a custom dashboard for our members, template invoices allow me to see what service a user is subscribed at. But since I can't list them, I cannot do that.

JonBendtsen:
There is an endpoint `GET /invoices/templates/{id}' in v22.0.3 + some extra commits that has not been released yet.

 True there is no `GET /invoices/templates`

I would suggest that you copy the `public function index` and modify it so it would work for template invoices.
https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/compta/facture/class/api_invoices.class.php#L209 

you can see in this `function fetch` which fields are available.
https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/compta/facture/class/facture-rec.class.php#L2382

Fork the dolibarr repo to your own github account, clone your dolibarr repo, create a branch, copy the index function, could be indexTemplates? and then you modify it to work on template invoices. Final step is to push your changes to github and create a Pull Request with your changes.

LeGmask:
> There is an endpoint `GET /invoices/templates/{id}' in v22.0.3 + some extra commits that has not been released yet.
> 
> True there is no `GET /invoices/templates`
> 
> I would suggest that you copy the `public function index` and modify it so it would work for template invoices. [`develop`/htdocs/compta/facture/class/api_invoices.class.php#L209](https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/compta/facture/class/api_invoices.class.php?rgh-link-date=2025-12-04T20%3A48%3A30.000Z#L209)
> 
> you can see in this `function fetch` which fields are available. [`develop`/htdocs/compta/facture/class/facture-rec.class.php#L2382](https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/compta/facture/class/facture-rec.class.php?rgh-link-date=2025-12-04T20%3A48%3A30.000Z#L2382)
> 
> Fork the dolibarr repo to your own github account, clone your dolibarr repo, create a branch, copy the index function, could be indexTemplates? and then you modify it to work on template invoices. Final step is to push your changes to github and create a Pull Request with your changes.

Thanks for theses hints, but it's already done see #36397, I've just seen that I forget about a close keyword in the pull request...

JonBendtsen:
[...]

PR #36397, opened 2025-11-24
REST API support for listing template invoices

LeGmask:
# NEW|New add REST API support for listing template invoices (#36396)

This pull request adds a new REST API route: `GET /invoices/templates` that list template invoices. 
But overall, it feels like the template invoice is lacking API support, so we maybe can open another issue to implement the other feature that are missing.

It is my first contribution to Dolibarr, feel free to guide or correct me if you think it can be improved.

Thank you again for this great tool you've built :) !

### DIFF
--- a/htdocs/compta/facture/class/api_invoices.class.php
+++ b/htdocs/compta/facture/class/api_invoices.class.php
@@ -1952,6 +1952,146 @@ class Invoices extends DolibarrApi
 		return $this->_fetchTemplateInvoice($id, '', '', $contact_list);
 	}
 
+
+	/**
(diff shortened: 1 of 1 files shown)


LeGmask:
@eldy, I've seen you've triggered the remainings pipelines, the phan one is failing but as far as I understand the error log it don't seem linked to my changes  ?

eldy:
> @eldy, I've seen you've triggered the remainings pipelines, the phan one is failing but as far as I understand the error log it don't seem linked to my changes  ?

It is not. I will try to rerun it...

PR #36398, opened 2025-11-24
FIX undefined variables on create invoice from shipment card (backport from v20)

lvessiller-opendsi:
FIX undefined variables on create invoice from shipment card (backport from v20)
- see commit c623558 and PR #28629
- some variables are not defined when you create an invoice from a shipment card and causes somes errors in PHP 8.2


rycks (approved):
ok

rycks (approved):
ok

PR #36400, opened 2025-11-24
Fix: Allow password reset link when CSRF protection is strict

atm-jonathan:
# Instructions

When MAIN_SECURITY_CSRF_WITH_TOKEN is set to 2 or 3 (checking GET requests), the validation link fails because:
The user comes from an email, so the request is a GET.
The user is not logged in yet, so there is no active session matching the browser context initially.
The URL generated in the email cannot contain a valid CSRF token.

# FIX|Fix 
Added define("NOCSRFCHECK", 1); in passwordforgotten.php before including main.inc.php. This explicitly tells the main controller to bypass the generic token check for this specific script.

This change is safe because passwordforgotten.php implements its own verification mechanism later in the code. It verifies the unique and time-sensitive passworduidhash passed in the URL. The generic CSRF token is redundant and technically impossible to provide in this specific "anonymous" workflow.

### DIFF
--- a/htdocs/main.inc.php
+++ b/htdocs/main.inc.php
@@ -337,7 +337,7 @@ if ((!defined('NOCSRFCHECK') && empty($dolibarr_nocsrfcheck) && getDolGlobalInt(
 	if ((GETPOSTISSET('massaction') || $tmpaction) && getDolGlobalInt('MAIN_SECURITY_CSRF_WITH_TOKEN') >= 3) {
 		// All GET actions (except the listed exceptions that are usually post for pre-actions and not real action) and mass actions are processed as sensitive.
 		// We exclude some action that are not sensitive so legitimate
-		if (GETPOSTISSET('massaction') || (strpos($tmpaction, 'display') !== 0 && !in_array($tmpaction, array('create', 'create2', 'createsite', 'createcard', 'edit', 'editcontract', 'editvalidator', 'file_manager', 'presend', 'presend_addmessage', 'preview', 'reconcile', 'specimen')))) {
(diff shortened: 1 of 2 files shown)


eldy:
It is safe to disabke CSRF but only when  $action == 'validatenewpassword' because the check on the passworduidhash exists only for this case.
So to be more secured, I suggest to modify yur fi wiith 
```
if ($_GET['action'] == 'validatenewpassword') {
define("NOCSRFCHECK", 1);
}
```
another solution is to exclude the action code validatenewpassword from CSRF from the main.inc.php page:
by changing
`		if (GETPOSTISSET('massaction') || (strpos($tmpaction, 'display') !== 0 && !in_array($tmpaction, array('create', 'create2', 'createsite', 'createcard', 'edit', 'editcontract', 'editvalidator', 'file_manager', 'presend', 'presend_addmessage', 'preview', 'reconcile', 'specimen')))) {`
into
`		if (GETPOSTISSET('massaction') || (strpos($tmpaction, 'display') !== 0 && !in_array($tmpaction, array('create', 'create2', 'createsite', 'createcard', 'edit', 'editcontract', 'editvalidator', 'file_manager', 'presend', 'presend_addmessage', 'preview', 'reconcile', 'specimen', 'validatenewpassword')))) {
`

i prefer this second solution so we keep a clear vision of all action code that are excluded from CSRF check.

What is strange to me is that you have the trouve with MAIN_SECURITY_CSRF_WITH_TOKEN=2when you should have it only when MAIN_SECURITY_CSRF_WITH_TOKEN=3.
Can you paste an exemple of link you have with MAIN_SECURITY_CSRF_WITH_TOKEN=2 ?

atm-jonathan:
You are absolutely right, I double-checked and indeed, the blocking behavior occurs strictly at Level 3 (MAIN_SECURITY_CSRF_WITH_TOKEN=3), not Level 2. My mistake on the initial report.

However, to ensure compatibility with high-security configurations (Level 3), I have applied your suggested fix:

Reverted changes in passwordforgotten.php.

Added 'validatenewpassword' to the exception array in main.inc.php.

PR #36402, opened 2025-11-24
NEW : api setup - Add dictionnary holiday lists

defrance:
list of defined holiday and list of public holiday

### DIFF
--- a/htdocs/api/class/api_setup.class.php
+++ b/htdocs/api/class/api_setup.class.php
@@ -977,6 +977,10 @@ class Setup extends DolibarrApi
 	{
 		$list = array();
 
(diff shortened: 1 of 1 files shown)


defrance (commented):
del white spaces

eldy:
Whe should have (like we should for all endpoints for holiday) one test on permission 

For example, in the api file for dictionary of proposal, we have

```
if (!DolibarrApiAccess::$user->hasRight('propal', 'lire') && !DolibarrApiAccess::$user->hasRight('commande', 'lire') && !DolibarrApiAccess::$user->hasRight('facture', 'lire')) {
			throw new RestException(403);
}
```

defrance (commented):
added on  expensereport dictionary api too

PR #36024, opened 2025-10-30
FIX #35655 Check if API user has rights to see all thirdparties

JonBendtsen:
# FIX #35655 Check if API user has rights to see all thirdparties
This PR is a fix for #35655, checking if the api user/key has access to the thirdparty used in the contract

PR #36408, opened 2025-11-24
Qual: DENY access to Thirdparty when module is enabled, but no permissions at all

JonBendtsen:
# QUAL DENY access to Thirdparty when module is enabled, but no permissions at all
This PR will catch the hopefully rare case where the Thirdparty module has been enabled but no permissions of any kind has been granted on Thirdparties.

This PR was discovered during exploration of this bug 35655 in this MR #36024

### DIFF
--- a/htdocs/core/lib/security.lib.php
+++ b/htdocs/core/lib/security.lib.php
@@ -1052,6 +1052,9 @@ function checkUserAccessToObject($user, array $featuresarray, $object = 0, $tabl
 				if ($user->socid != $objectid) {
 					return false;
 				}
+			} elseif (isModEnabled('societe') && !($user->hasRight('societe', 'lire') && !$user->hasRight('societe', 'client', 'voir'))) {
+				dol_syslog("security.lib.php::checkUserAccessToObject True: (isModEnabled('societe') && !(user->hasRight('societe', 'lire') && !user->hasRight('societe', 'client', 'voir')))", LOG_DEBUG);
+				return false;
 			} elseif (isModEnabled("societe") && ($user->hasRight('societe', 'lire') && !$user->hasRight('societe', 'client', 'voir'))) {
 				// If internal user: Check permission for internal users that are restricted on their objects
 				$sql = "SELECT COUNT(sc.fk_soc) as nb";
--- a/htdocs/core/lib/security.lib.php
+++ b/htdocs/core/lib/security.lib.php
@@ -1266,7 +1266,7 @@ function checkUserAccessToObject($user, array $featuresarray, $object = 0, $tabl
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


htdocs/core/lib/security.lib.php
```diff
@@ -1052,6 +1052,9 @@ function checkUserAccessToObject($user, array $featuresarray, $object = 0, $tabl
 				if ($user->socid != $objectid) {
 					return false;
 				}
+			} elseif (isModEnabled('societe') && !($user->hasRight('societe', 'lire') && !$user->hasRight('societe', 'client', 'voir'))) {
```
hregis: @JonBendtsen for me it's not good, try this instead

`   } elseif (isModEnabled('societe') && !$user->hasRight('societe', 'lire') && !$user->hasRight('societe', 'client', 'voir')) {`

eldy: I think what you want to do is:
`} elseif (isModEnabled('societe') && !$user->hasRight('societe', 'lire') && !$user->hasRight('societe', 'client', 'voir')) {`
So: if thirdparty module is used, and user has no permisstion to read and no permission to read all (what you call "no permission of any kind", the permission check is false.
If you keep the parenthesis, the test become if has no permission to read OR can read all.
!( a and b)   is   (!a or !b)

hregis:
@eldy maybe add an extra for superadmin if muticompany !? a superadmin can always... an admin can always just without multicompany...maybe add a function for this ?

hregis:
@eldy for you it's possible to defined  a superadmin status ? if "admin" parameter = 1 and "entity" = 0 ... this is a superadmin... for me a superadmin can have all access ! without this rights are in rights definitions in database !

JonBendtsen:
> @eldy for you it's possible to defined a superadmin status ? if "admin" parameter = 1 and "entity" = 0 ... this is a superadmin... for me a superadmin can have all access ! without this rights are in rights definitions in database !

@hregis @eldy 

The API files usually has 2 checks:

1. `if (!DolibarrApiAccess::$user->hasRight('contrat', 'creer')) {`
2. `if (!DolibarrApi::_checkAccessToResource('contrat', $this->contract->id)) {`

So unless you define a superadmin check in the hasRight function then it will not matter if it is defined in the _checkAccessToResource function because it will never get there.

Also I think that while your idea is good, I think it should be a separate commit in the development branch, where as this commit is just a small fix for a bug.

If it is already in the hasRight function, then it should just be a small copy and I think that would be reasonable to add.

Given that this function ends with a `return true` I would prefer if the whole file could be re-designed in v24 or v25 such that the default is to deny access. It just feels more secure. I think it is too late for v23.

PR #36411, opened 2025-11-25
FIX avoid php warnings (or not ?)

hregis:


htdocs/core/class/dolgraph.class.php
```diff
@@ -732,7 +732,9 @@ public function GetMaxValueInData()
 		foreach ($this->data as $x) {	// Loop on each x
 			for ($i = 0; $i < $nbseries; $i++) {	// Loop on each series
 				if (is_null($max)) {
-					$max = $x[$i + 1];		// $i+1 because the index 0 is the legend
+					if (isset($x[$i])) {
```
eldy: Looks strange.
Do you mean 
`if (isset($x[$i + 1]))` 
?

hregis: @eldy arf yes you right, i fix my fix !

PR #36412, opened 2025-11-25
FIX do not force rowid to int

kubajznik:
FIX When string is field (e.g. ref) is used instead of rowid in checkboxlist, we can't force rowid to be int. Was changed in line 8801 commit [c62e015410e489b63bf3bdd58dfd93f38214e5f1](https://github.com/Dolibarr/dolibarr/commit/c62e015410e489b63bf3bdd58dfd93f38214e5f1#diff-cf800c5395ad947cf09307e873b8450cea7733e464cf7c9b553f03b7781fa4b8)

### DIFF
--- a/htdocs/core/class/commonobject.class.php
+++ b/htdocs/core/class/commonobject.class.php
@@ -8369,7 +8369,7 @@ abstract class CommonObject
 									$isDependList = 1;
 								}
 
-								$data[(int) $obj->rowid] = $labeltoshow;
+								$data[$obj->rowid] = $labeltoshow;
 							}
 
 							$i++;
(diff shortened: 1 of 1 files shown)


hregis:
@kubajznik strange ! a rowid it's an integer... have you an example ?

hregis:
@kubajznik i can't reproduce your problem

kubajznik:
> @kubajznik strange ! a rowid it's an integer... have you an example ?

@hregis for a custom field, you can set `'sites' => array('type' => 'chkbxlst:sites_table:label:ref', 'label' => ...`
In this case, the select statement is `SELECT ref as rowid, label FROM llx_sites_table WHERE 1=1`, so the code then uses `$obj->rowid` ... but rowid in that case is the ref from that table, and not the actual rowid.

hregis:
@kubajznik ok i see, thanks

PR #36414, opened 2025-11-25
FIX multicompany dictionary sharing

hregis:
@defrance It would be nice if my comments were taken into account when I request a correction. thank you : https://github.com/Dolibarr/dolibarr/pull/36402


defrance:
hello @hregis 
not see your comment on the mentionned PR
maybe on another one?

hregis:
@eldy @frederic34 phan it's not my PR

PR #36415, opened 2025-11-25
New : add api call on contact and user to get list of associated element

defrance:


hregis:
@defrance not muticompany compatibility for the moment here ! ❤️ 👍 😄

PR #36422, opened 2025-11-25
FIX: Resolve JS error on country change in public member form

marcellintacite:
# FIX : JS error on country change in public member form

On the public member subscription page http://localhost:8888/dolibarr/htdocs/public/members/new.php, changing the country dropdown triggers a form submission to reload the page (specifically to update the state/province list).

The JavaScript handler for the country change event was throwing the following error in the browser console:
Uncaught TypeError: Cannot read properties of undefined (reading 'action')

This occurred because the legacy DOM access method (document.newmember...) was failing to resolve the form object correctly in the current execution context. This prevented the form from reloading, making it impossible to select states/provinces for the chosen country.

### Solution
Updated the JavaScript code to use jQuery selectors to target the form and the action input. This is more robust and consistent with the rest of the file, ensuring the elements are correctly resolved before attempting to modify the value or submit the form.
Capture d’écran 2025-11-26 à 00 46 13

### DIFF
--- a/htdocs/public/members/new.php
+++ b/htdocs/public/members/new.php
@@ -683,8 +683,8 @@ if (getDolGlobalString('MEMBER_SKIP_TABLE') || getDolGlobalString('MEMBER_NEWFOR
 		print "\n".'<script type="text/javascript">'."\n";
 		print 'jQuery(document).ready(function () {
 			jQuery("#selectcountry_id").change(function() {
-				document.formsoc.action.value="create";
-				document.formsoc.submit();
+				document.newmember.action.value="create";
+				document.newmember.submit();
 			});
 			function initfieldrequired() {
 				jQuery("#tdcompany").removeClass("fieldrequired");


marclaporte:
Thank you @eldy!

So this was noticed here: https://admin.dolibarr.org/public/members/new.php

Should @marcellintacite propose PRs to backport to previous branches? If so, which ones?

eldy:
> Thank you @eldy!
> 
> So this was noticed here: https://admin.dolibarr.org/public/members/new.php
> 
> Should @marcellintacite propose PRs to backport to previous branches? If so, which ones?

Code seems different in previous versions. Not sure it is present. But if you find it in older version, yes you can suggest a backport. You can target v19 or v20 for example, it will be merged in all higher versions.

marclaporte:
@marcellintacite The issue persists on https://admin.dolibarr.org/public/members/new.php

What is the status for backports?

marcellintacite:
> @marcellintacite The issue persists on https://admin.dolibarr.org/public/members/new.php
> 
> What is the status for backports?

Sorry for the late delay, i investigated other versions and found that no backports are needed. The bug I fixed was introduced exclusively in the develop branch via PR #35033 and never existed in versions 19.0 through 22.0. Those older branches already use the correct document.newmember code and work perfectly, which I verified locally

marclaporte:
@eldy How can we check which version https://admin.dolibarr.org/public/members/new.php is running?

eldy:
> @eldy How can we check which version https://admin.dolibarr.org/public/members/new.php is running?

May be by reading the source file of html page (in the header) ?

marcellintacite:
> > @eldy How can we check which version https://admin.dolibarr.org/public/members/new.php is running?
> 
> May be by reading the source file of html page (in the header) ?

Thanks @eldy , 
the live website is running 24.0.0-beta which already contains the JavaScript fix. The reason the State/Province dropdown remains empty on the live site is a database configuration issue i think. 
Could you confirm if the states for Canada are activated in the Dictionary settings on the live server? Because if i choose another country, i can see states

PR #36429, opened 2025-11-26
FIX : Implementation of multi-company compatibility with inventory/warehouse management

atm-adrien:
## FIX : Implementation of multi-company compatibility with inventory/warehouse management

It seems that stock management is not fully compatible with the multicompany module. Indeed, in the current state, if I don’t share stock management between two entities, I still have the possibility to access the inventories and warehouses of entity 1 from entity 2.
Is this intended? I noticed that the “stock” feature was added to the $nocheck array.

@eldy @hregis

### DIFF
--- a/htdocs/core/lib/security.lib.php
+++ b/htdocs/core/lib/security.lib.php
@@ -902,14 +902,14 @@ function checkUserAccessToObject($user, array $featuresarray, $object = 0, $tabl
 		$checkonentitydone = 0;
 
 		// Array to define rules of checks to do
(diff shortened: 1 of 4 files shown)


rycks (approved):
ok

rycks (approved):

rycks:
je fais appel à un amis :) @hregis ho grand manitou du "MoutiComp" peux tu nous confirmer que c'est bon pour toi ?

hregis:
@rycks attend je regarde dans ma boule de cristal... oui c'est good... MoultiPass 👍

PR #36435, opened 2025-11-26
FIX generateInputFieldPassword max length

Daoud-mohamed:
Fix generateInputFieldPassword max length.


htdocs/core/class/html.formsetup.class.php
```diff
@@ -1139,7 +1139,7 @@ public function generateInputFieldPassword($type = 'generic')
 			require_once DOL_DOCUMENT_ROOT."/core/modules/security/generate/".$nomfichier;
 			$genhandler = new $nomclass($this->db, $conf, $langs, $user);
 			$min = $genhandler->length;
-			$max = $genhandler->length2;
+			// $max = $genhandler->length2; // length2 is not used for max length
```
eldy: Don't understand.
length2 is 0 with modGeneratePassNone, but set to 12 with modGeneratePassStandard and set to x with modGeneratePassPerso.
What was wrong with this line ?

Daoud-mohamed: In modGeneratePassStandard, the value of length and length2 is 12
In modGeneratePassPerso, only length2 is used to define the minimum length

With modGeneratePassStandard, generated field has minlength="12" and maxlength="12", so longer passwords can’t be entered or pasted (they get truncated).

Removing this line restores the default maxlength of 50 for both modes.

Alternatively, if we want a configurable maxlength, we need to update both modGeneratePassStandard and modGeneratePassPerso so they use length and length2 consistently.

PR #36438, opened 2025-11-27
NEW Add ACCOUNTANCY_EXPORT_DISABLE_COMPANY_SYNC flag

fbagnol:
# NEW Add ACCOUNTANCY_EXPORT_DISABLE_COMPANY_SYNC flag
Add Add ACCOUNTANCY_EXPORT_DISABLE_COMPANY_SYNC flag to disable syncing company data in accountancy export.
Work only or Quadratus right now

### DIFF
--- a/htdocs/accountancy/class/accountancyexport.class.php
+++ b/htdocs/accountancy/class/accountancyexport.class.php
@@ -925,43 +925,45 @@ class AccountancyExport
 
 			$tab = array();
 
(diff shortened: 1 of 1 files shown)


aspangaro:
Hello,

Interesting, but rather than deleting the entire line of type C, why not just leave position 217 blank so that it is not updated?

image

We agree that this only concerns Quadra as it stands.

PR #36442, opened 2025-11-27
add modulepart for direct print capabilities

frederic34:


hregis:
@frederic34 Wouldn't it be better to implement a parameter in the "module_parts" declaration in a module that would be registered in llx_const instead of hard-coding the module names in the code?

hregis:
@frederic34 This would avoid having to correct the code every time a module is added, and most importantly, it would be compatible with all added externals modules!

hregis:
@frederic34 stop hard coded please !

PR #36448, opened 2025-11-27
Insert comment before first line of sending in invoice

le-reparateur:
Hi
Create invoice from validated sendinds need to insert comment in the first line before insert lines of a sending (even if there is only on sending to invoice).
Added comment on first line before sending do not work.

Added conditional logic to insert a comment before the first line of sending, including order and shipment references.

Francis

### DIFF
--- a/htdocs/expedition/list.php
+++ b/htdocs/expedition/list.php
@@ -429,15 +429,14 @@ if (empty($reshook)) {
 							}
 							// if we insert first line of sending we must insert comment before first line
 							if ($i == 0) {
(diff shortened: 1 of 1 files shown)


le-reparateur (commented):
correction of syntax

le-reparateur (commented):
correction

le-reparateur:
Need to be more efficient with Github...
Sorry !

eldy:
Can you provide a screenshot of result you have after pr ?

le-reparateur:
Hi
With one sending per invoice
Ecran2
With 2 sendings for 1 invoice
ecran1

Francis

le-reparateur:
And with Sponge model
Sponge
Wich is not very nice...

PR #36456, opened 2025-11-28
family assignment logic on user perms of external modules

defrance:
Updated family assignment logic to default to 'other' if family does not exist in $familyinfo.


defrance (commented):
space ok

defrance (commented):
ok

defrance:
It's a main problem on V23 release if we have other family modules

Capture d’écran 2026-01-16 à 12 20 46

defrance:
@eldy , could you please look at this PR first? Otherwise, we'll have a problem with all external modules that aren't part of an existing module family.

eldy (approved):

PR #36479, opened 2025-11-29
subtotal on interventional: display bottom line

defrance:
Rest for the feature
- edit subtotal line
- display on pdf


defrance (commented):
oki

PR #36480, opened 2025-11-29
FIX timesheet date comparison

atm-jonathan:
# FIX Fix timesheet restriction date comparison

**Problem:**
When the constant `PROJECT_TIMESHEET_PREVENT_AFTER_MONTHS` is set, the check to prevent insertion, modification, or deletion of old timesheets was failing.
The code was comparing `$this->timespent_date` (a string, e.g., '2023-01-01') directly with `$restrictBefore` (an integer timestamp). This type mismatch caused the condition to evaluate incorrectly (often returning false), allowing changes even when they should have been restricted.

**Solution:**
Added `strtotime()` to convert `$this->timespent_date` into a Unix timestamp before comparing it with the `$restrictBefore` timestamp.

**Affected methods in `projet/class/task.class.php`:**
- `addTimeSpent`
- `updateTimeSpent`
- `delTimeSpent`


pixodeo:
PR #38921 opened against 18.0. The three comparison sites in task.class.php now coerce $this->timespent_date with is_numeric/(int)strtotime before comparing against the dol_time_plus_duree timestamp, so the PROJECT_TIMESHEET_PREVENT_AFTER_MONTHS guard catches both the date-string and the int-timestamp callers.

PR #36485, opened 2025-11-29
FIX #33899 API Timespent Service Field

### LINKED ISSUE #33899: [API] Timespent Service Field
### Feature Request

Hi guys,

I am trying to automate my timespent logging and use an XLS sheet to pass the data to the timespent endpoint - but the service field is missing - I would love to choose a service like I can do in the GUI.

In the API it could just be referenced by ID.
Currently, you need to edit every single line and choose the correct service via GUI.

I would love if you could implement this in the API

### Use case

Automation of timespent and invoicing

### Suggested implementation

_No response_

### Suggested steps

_No response_

JonBendtsen:
# FIX #33899 API Timespent Service Field
This PT is an attempt to fix the bug in #33899 and related bugs found during the investigation.

# TO DECIDE
- Add 'projet_task' to call of `_checkAccessToResource` in api_tasks.class.php OR add it in function `checkUserAccessToObject` in file security.lib.php? @eldy (like `$dbtablename = 'actioncomm';` in line 982)

### DIFF
--- a/htdocs/core/lib/security.lib.php
+++ b/htdocs/core/lib/security.lib.php
@@ -976,6 +976,7 @@ function checkUserAccessToObject($user, array $featuresarray, $object = 0, $tabl
 		}
 		if ($feature == 'task' || $feature == 'projet_task') {
 			$feature = 'project_task';
+			$dbtablename = 'projet_task';
 		}
 		if ($feature == 'eventorganization') {
 			$feature = 'agenda';
--- a/htdocs/projet/class/api_tasks.class.php
+++ b/htdocs/projet/class/api_tasks.class.php
@@ -78,7 +78,7 @@ class Tasks extends DolibarrApi
 			throw new RestException(404, 'Task not found');
 		}
 
-		if (!DolibarrApi::_checkAccessToResource('task', $this->task->id)) {
+		if (!DolibarrApi::_checkAccessToResource('task', $this->task->id, 'projet_task')) {
 			throw new RestException(403, 'Access not allowed for login '.DolibarrApiAccess::$user->login);
 		}
 
--- a/htdocs/projet/class/api_tasks.class.php
+++ b/htdocs/projet/class/api_tasks.class.php
@@ -78,7 +78,7 @@ class Tasks extends DolibarrApi
 			throw new RestException(404, 'Task not found');
 		}
 
-		if (!DolibarrApi::_checkAccessToResource('task', $this->task->id, 'projet_task')) {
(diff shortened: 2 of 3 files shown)


eldy:
> # FIX #33899 API Timespent Service Field
> This PT is an attempt to fix the bug in #33899 and related bugs found during the investigation.
> 
> # TO DECIDE
> - Add 'projet_task' to call of `_checkAccessToResource` in api_tasks.class.php OR add it in function `checkUserAccessToObject` in file security.lib.php? @eldy (like `$dbtablename = 'actioncomm';` in line 982)
> 

It seems that you choosed first option.
Both are good so keep the one you choosed to submit pr.

JonBendtsen:
> > # FIX #33899 API Timespent Service Field
> > This PT is an attempt to fix the bug in #33899 and related bugs found during the investigation.
> > # TO DECIDE
> > 
> > * Add 'projet_task' to call of `_checkAccessToResource` in api_tasks.class.php OR add it in function `checkUserAccessToObject` in file security.lib.php? @eldy (like `$dbtablename = 'actioncomm';` in line 982)
> 
> It seems that you choosed first option. Both are good so keep the one you choosed to submit pr.

@eldy  that was just a quick proof of concept - getting something that worked.

I am still not sure because: while this fix is tiny and precise, it does not help anywhere else that uses the checkUserAccessToObject function to check access to a task, but that would be a bigger possibly more interruptive change.

I am leaning towards using method b in the function checkUserAccessToObject and will submit a new PR with that. (once tested)

JonBendtsen:
@eldy  okay, so the more general fix is in now.

But there is another thing that worries me a little more.

When I call this endpoint `/api/index.php/tasks/timespentrecordchecks/6?timespent_id=3` then it seems like the API makes a call to this `protected function timespentRecordChecks($id, $timespent_id)`
https://github.com/Dolibarr/dolibarr/blob/22.0/htdocs/projet/class/api_tasks.class.php#L692

I did not expect a call to anything expect for a public function, and I don't like that the API just calls some more or less random function that has the right number and type of arguments. I would expect that the API **KNEW** which public function an API endpoint would correlate to.

eldy (approved):

JonBendtsen:
@eldy  I think you were a little too fast to merge this, I was working on fixing the call to the protected function when trying to getTimeSpent.

Also I am unsure if the bug is actually fixed, as I have not yet actually worked on the put/post timespent API endpoints that was reported in #33899

JonBendtsen:
continued here https://github.com/Dolibarr/dolibarr/pull/36486

PR #36486, opened 2025-11-30
FIX: API task getTimeSpent to actually return data

### LINKED ISSUE #33899: [API] Timespent Service Field

JonBendtsen:
# FIX #33899 API GET/POST/PUT Timespent Service/product Field
This PR will close #33899. The Task API can now GET a timespent record and it can POST and PUT a timespent record with a product_id.

### DIFF
--- a/htdocs/core/class/timespent.class.php
+++ b/htdocs/core/class/timespent.class.php
@@ -105,6 +105,7 @@ class TimeSpent extends CommonObject
 	 */
 	public $fields = array(
 		'rowid' => array('type' => 'integer', 'label' => 'TechnicalID', 'enabled' => 1, 'position' => 1, 'notnull' => 1, 'visible' => 0, 'noteditable' => 1, 'index' => 1, 'css' => 'left', 'comment' => "Id"),
+		'ref_ext' => array('type' => 'varchar(255)', 'label' => 'RefExt', 'enabled' => 1, 'visible' => 0, 'position' => 20),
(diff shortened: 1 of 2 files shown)


JonBendtsen:
@eldy I'm reading https://www.dolibarr.org/feedback-of-the-devcamp_angouleme-2025_whats-new-in-v23.php and the list of tickets created and closed. I notice that this ticket is on the list, and I wonder why it is on that list?

eldy:
> @eldy I'm reading https://www.dolibarr.org/feedback-of-the-devcamp_angouleme-2025_whats-new-in-v23.php and the list of tickets created and closed. I notice that this ticket is on the list, and I wonder why it is on that list?

I just massively flagged all tickets submitted between start and end of the event.

eldy:
> @eldy I'm reading https://www.dolibarr.org/feedback-of-the-devcamp_angouleme-2025_whats-new-in-v23.php and the list of tickets created and closed. I notice that this ticket is on the list, and I wonder why it is on that list?

I just massively flagged all tickets submitted between start and end of the event.

PR #36487, opened 2025-11-30
Add default supplier type handling in card.php in creation mode

le-reparateur:
Hi
By default supplier is a professional in creation mode.
Francis


le-reparateur (commented):
Syntax

PR #36494, opened 2025-12-01
FIX 18.0: `GETPOSTDATE()` and `buildParamDate()` assumed wrong HTTP param names

atm-florianm:
# FIX
When I introduced `GETPOSTDATE` three years ago (#23483), I didn't use the correct http parameter names for minutes and seconds. Usage of the function doesn't seem to have spread much since, but I recently had to use GETPOSTDATE to get a timestamp from a form that did use time of day information. I then noticed that minutes and seconds were set to 0 instead of what was actually sent.


lvessiller-opendsi (approved):

rycks (approved):
ok for me

Issue #36406, opened 2025-11-24
Bug: missing Key (column) to use for updating existing data in "Vendors/Purchase order lines" import tool


christophemoille:
# Bug
rang (llx_commande_fournisseurdet) is missing from Key (column) list to use for updating existing data in "Vendors/Purchase order lines" import tool.

## Environment
- **Version**: 22.0.2
- **OS**: Linux
- **Web server**: Apache/2.4.65 (Debian)
- **PHP**: fpm-fcgi 8.4.14
- **Database**: MySQL or MariaDB 11.8.3-MariaDB-0+deb13u1 from Debian
- **URL(s)**: /imports/import.php?step=5&format=csv&datatoimport=commande_fournisseurdet_4&filetoimport=20251124171311-lignes_commande_fournisseur_generique.csv&excludefirstline=2&separator=%2C&enclosure=%22&filetoimport=20251124171311-lignes_commande_fournisseur_generique.csv

## Expected and actual behavior
Should be able to update Vendors order lines by importing adequate data.
But as there is not rang (llx_commande_fournisseurdet) in Key (column) list to use for updating existing data, I can not import data.

If I force import, got an error : "Multiple records have been found with these filters"

## Steps to reproduce the behavior
Importing existing data for a purchase with multiple order lines.

christophemoille:
Another problem with "Vendors/Purchase order lines" import tool. 
If I try to import non existent purchase with multiple line, only one line (last in import) will be added to the order.

Danelif:
@christophemoille i am taking this issue

marclaporte:
@Danelif @christophemoille Any next steps? :-)

eldy:
Should be available with v23, so I close ticket.

PR #36501, opened 2025-12-02
improve purchase order line import by adding rang and flexible key se…

Danelif:
Users can now select rang (llx_commande_fournisseurdet) in the “Key (column) list to use for updating existing data” dropdown when importing vendor/purchase order lines.
This allows updates to existing lines without triggering the “Multiple records” error, ensuring more reliable and accurate import operations.

[See the issue](https://github.com/Dolibarr/dolibarr/issues/36406)


Danelif:
@eldy seems you tried to merge this but since then it is still open

PR #36503, opened 2025-12-02
FIX: when INVOICE_USE_SITUATION === 2, some libs still use the old algorithms

atm-florianm:
# FIX
## Background
I am not very familiar with situation invoices.

For reference: [GIFF Facture de situation (fr)](https://www.dolibarr.fr/forum/t/gif-facture-de-situation/41868/1)

In earlier Dolibarr versions, `INVOICE_USE_SITUATION` was functionally boolean (on/off). The column `llx_facturedet.situation_percent` contained a **state**: the overall percentage of progress for the line over the whole cycle.

Now situation invoices have been better integrated within Dolibarr's core, `INVOICE_USE_SITUATION` can have 3 values, depending on which the column `llx_facturedet.situation_percent` will represent different things:

| `INVOICE_USE_SITUATION` | `situation_percent` | Example / explanations |
| ---- | ---- | ---- |
| 0 | disabled | N/A |
| 1 = legacy mode | cumulative state of progress in % within the cycle | 80% = when that situation invoice was issued, the work was 80% complete |
| 2 = new mode | Δ<sub>progress</sub> in percentage points | 10% = between the previous situation and this situation, the work has progressed by 10% |

The same is true for other fields of the line (the amounts etc.).

## Rationale
The code was updated to accommodate for the new mode, but not everywhere. One of our customer found big errors (negative ratios leading to negative amounts) in the PDF generated using the new mode, because the legacy algorithm was used regardless of `INVOICE_USE_SITUATION`.

## This fix / caveat
As stated initially, I am not comfortable with situation invoicing. I looked for places in the code where the value of `INVOICE_USE_SITUATION` was not checked before performing the calculations and then attempted to fix them to the best of my knowledge, but, due to time constraints, I only tested the PDF fix using my customer's settings.

[edit]
I realized there were other functions forcing the legacy calculation mode, so I introduced a new `FactureLigne` method to get the progress ratio in one place rather than everywhere it is needed (it might be useful elsewhere as well).

### DIFF
--- a/htdocs/compta/facture/class/factureligne.class.php
+++ b/htdocs/compta/facture/class/factureligne.class.php
@@ -995,7 +995,12 @@ class FactureLigne extends CommonInvoiceLine
 			// state at the previous situation (if applicable).
 			$prevProgress = $this->get_prev_progress($this->fk_facture);
 
-			return ($this->situation_percent - $prevProgress) / 100;
+			if ($this->situation_percent == 0) {
+				// should not happen
+				return 0;
+			}
+
+			return ($this->situation_percent - $prevProgress) / $this->situation_percent;
 		}
 		// new mode (INVOICE_USE_SITUATION == 2):
 		// no ratio needed (data stored on line is already a delta)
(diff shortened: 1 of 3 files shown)


eldy:
As this is a big change and may have side effect on a feature that is not yet officially supported, I suggest to push this into develop branch.
We dont have to maintain experimental features into old brnach and both option 1 and 2 are still experimental features so i merged it into develop

PR #36506, opened 2025-12-02
FIX add constante to force to hide product not for sale/purchase

hregis:


hregis:
@eldy is for to have the possibility to hide the "desactivated" products (no to buy and/or not to purchase) in the list of choose product in stock and inventory or others !

hregis:
@eldy if other PR is better... je suis preneur ! 😄 
car voir des produits "désactivés" visibles lorsqu'on veut modifier un stock dans un entrepôt ou choisir un produit dans un inventaire est à mon sens un fonctionnement qui n'a pas de sens !

hregis:
@eldy voir dans l'absolu pourquoi pas désactiver l'onglet "stock" sur une fiche produit lorsqu'un produit est "désactivé" ?

htdocs/core/class/html.form.class.php
```diff
@@ -3083,10 +3083,14 @@ public function select_produits_list($selected = 0, $htmlname = 'productid', $fi
 		} elseif ($finished == 1) {
 			$sql .= " AND p.finished = ".((int) $finished);
 		}
-		if ($status >= 0) {
+		if (getDolGlobalString('PRODUIT_FORCE_HIDE_NOT_FOR_SALE')) {
```
eldy: There is already a param $status and $status_purchase so it is possible to decide if we want on purchase and/or on sale products depending on the context.

Adding option here will overwrite any context and break the feature to have it when we need, so it is better to set the value of status by the caller depending on context (only the caller can know if he want/need a status or not have them when he don't need/want).

hregis:
You're right, closing this. Looking closer, every stock/inventory/MRP product picker in the codebase (massstockmove.php, stockatdate.php, inventory.php, stocktransfer.tpl.php, stockcorrection.tpl.php, stocktransfer_card.php x2, mo_production.php x2) consistently passes `status=-1` — it's a deliberate, module-wide choice so stock can still be corrected/transferred/inventoried for a product that has since been deactivated for sale or purchase. Forcing `tosell`/`tobuy` to 1 on any of these callers would break that for the ones I'd change and be inconsistent with the rest of the module. The actual ask (hide products that are neither sellable nor purchasable at all) would need an OR-based filter that $status/$status_purchase can't express as independent AND conditions, so it's not just a matter of setting them from the caller — not worth the added complexity for what's normal behavior in this module.

PR #36518, opened 2025-12-03
Fix broken commit 9801f02 mentioned in issue #36507

### LINKED ISSUE #36507: BUG commit 9801f022f96342e20c77dc87e5c64233deac4e6b API explorer gives http code 500
### Bug

BUG commit 9801f022f96342e20c77dc87e5c64233deac4e6b API explorer gives http code 500

in branch 22

f7fee10e234aeb8a4a4bf921296066c2435bb9ab seems to work

# webserver log
```
[Tue Dec 02 18:33:52.357154 2025] [php:error] [pid 136:tid 136] [client 192.168.127.1:40167] PHP Fatal error:  Allowed memory size of 268435456 bytes exhausted (tried to allocate 4096 bytes) in /var/www/html/includes/restler/framework/Luracast/Restler/CommentParser.php on line 503, referer: http://localhost:22080/api/index.php/explorer/
192.168.127.1 - - [02/Dec/2025:18:33:50 +0100] "GET /api/index.php/explorer/swagger.json HTTP/1.1" 500 363 "http://localhost:22080/api/index.php/explorer/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/140.0.0.0 Safari/537.36"
```

### Dolibarr Version

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

_No response_

### Attached files

_No response_

JonBendtsen:
# Fix #36507 broken commit 9801f02 mentioned in issue
This PR is a fix for issue #36507 which was broken in commit 9801f02. I can now once again use the API explorer.

### DIFF
--- a/htdocs/projet/class/api_tasks.class.php
+++ b/htdocs/projet/class/api_tasks.class.php
@@ -536,7 +536,7 @@ class Tasks extends DolibarrApi
 	 *
 	 * @url	GET {id}/getTimeSpent/{timespent_id}
 	 *
-	 * @return	array|mixed                         data without useless information
+	 * @return	Object                      data without useless information
 	 *
 	 * @throws	RestException
 	 */
--- a/htdocs/projet/class/api_tasks.class.php
+++ b/htdocs/projet/class/api_tasks.class.php
@@ -536,7 +536,7 @@ class Tasks extends DolibarrApi
 	 *
 	 * @url	GET {id}/getTimeSpent/{timespent_id}
 	 *
-	 * @return  TimeSpent
+	 * @return	array|mixed                         data without useless information
 	 *
 	 * @throws	RestException
 	 */
@@ -945,8 +945,4 @@ class Tasks extends DolibarrApi
 		}
 		return $object;
 	}
-
-
-	// \todo
-	// getSummaryOfTimeSpent
 }


Hystepik:
I think a return of type Object is a better way to fix because the return is 
``
return $object;
``

JonBendtsen:
> I think a return of type Object is a better way to fix because the return is 
> 
> ``
> 
> return $object;
> 
> ``

Why would that work? It goes from being a TimeSpent object to just being a generic Object

hregis:
@JonBendtsen 

`return $this->_cleanTimeSpentObjectDatas($timespent);`

return an object, not an array

Hystepik:
Also TimeSpent is the specific type but Object is the generic type

hregis:
@JonBendtsen and "_cleanTimeSpentObjectDatas($timespent)" call

`protected function _cleanObjectDatas($object)`

hregis:
@JonBendtsen an "Object" is better of "TimeSpent"
the mystery of the galaxy... 😄

JonBendtsen:
> @JonBendtsen 
> 
> 
> 
> `return $this->_cleanTimeSpentObjectDatas($timespent);`
> 
> 
> 
> return an object, not an array

Yes @hregis, but the consequence was that without the change in this PR is that the API explorer fails with an error described in the bug report.

And last I checked this PR builds cleanly - just like the other one that caused the bug

What do you want me to do? I do not know why api explorer fails when the @return value was defined as TimeSpent but it works when it is an array

JonBendtsen:
> @JonBendtsen and "_cleanTimeSpentObjectDatas($timespent)" call
> 
> 
> 
> `protected function _cleanObjectDatas($object)`

No, because that function is made for tasks, and Task and TimeSpent objects are different, so different things should be cleaned

Hystepik:
The problem comes from an infinite loop when parsing TimeSpent class. If we use Object the class parse is not trigger and the explorer continue the execution without a problem.

Here is the problem :
Timespent as User object has param of a  function which has UserGroup object as param of a function which has User object as param of a function ...

hregis:
@Hystepik you are the best ! 😄

JonBendtsen:
> The problem comes from an infinite loop when parsing TimeSpent class. If we use Object the class parse is not trigger and the explorer continue the execution without a problem.
> 
> 
> 
> Here is the problem :
> 
> Timespent as User object has param of a  function which has UserGroup object as param of a function which has User object as param of a function ...
> 
> 


Great. I'll try with Object later and see if the API explorer still works


Would it be possible to make the build system detect that this would result in an infinite loop? Such that others are not hit by the same issue?


JonB

hregis:
@mdeweerd @frederic34

Je ne suis qu'un développeur avec des idées pré-déterminées... Je suis à la base un discident depuis 20 ans donc je n'ai plus aucun poids dans les décisions ! Je ne ne peux que vous aider pour vous sortir des méandres de Dolibarr.

JonBendtsen:
@Hystepik  it worked :-)

PR #36520, opened 2025-12-04
bug when use api supplierorder with productype blank and use pgsql, casting use string shoud be to int: pg_strtoint32_safe

hermans:
bug when use api supplierorder with productype blank and use pgsql, casting use string shoud be to int: pg_strtoint32_safe


eldy:
I pushed a better fix with commit https://github.com/Dolibarr/dolibarr/commit/193a42cd4756f2965c601d4f765767f2f7550d22

PR #36522, opened 2025-12-04
do not create accelerated deprecation lines if disabled

lippoliv:
# FIX|Fix accelerated deprecation lines created even when disabled
When creating / updating an asset and having "accelerated deprecations" disabled, still the book entries would be created.
As far as I understand this would end up in this asset being "double-deprecated" (normal + accelerated).
Assisted by AI I identified the issue and created a fix for it.

Now as long as the accelerated deprecations checkbox is disabled, the particular lines wouldn't be created. I also updated the UI to prevent an error because of an missing array element.

I tried to go with the least-possible set of changes.

### DIFF
--- a/htdocs/asset/class/assetdepreciationoptions.class.php
+++ b/htdocs/asset/class/assetdepreciationoptions.class.php
@@ -316,8 +316,8 @@ class AssetDepreciationOptions extends CommonObject
 		foreach ($this->deprecation_options_fields as $mode_key => $mode_info) {
 			if (!empty($mode_info['enabled_field'])) {
 				$info = explode(':', $mode_info['enabled_field']);
-				if (!empty($this->deprecation_options[$info[0]][$info[1]]) && $deprecation_options[$info[0]][$info[1]] != $info[2]) {
-					unset($deprecation_options[$info[0]][$info[1]]);
+				if (isset($deprecation_options[$info[0]][$info[1]]) && $deprecation_options[$info[0]][$info[1]] != $info[2]) {
+					unset($deprecation_options[$mode_key]);
 				}
 			}
 		}
@@ -389,8 +389,8 @@ class AssetDepreciationOptions extends CommonObject
 		foreach ($this->deprecation_options_fields as $mode_key => $mode_info) {
 			if (!empty($mode_info['enabled_field'])) {
 				$info = explode(':', $mode_info['enabled_field']);
-				if (isset($deprecation_options[$info[0]][$info[1]]) && $deprecation_options[$info[0]][$info[1]] != $info[2]) {
-					unset($deprecation_options[$info[0]][$info[1]]);
+				if (!empty($deprecation_options[$info[0]][$info[1]]) && $deprecation_options[$info[0]][$info[1]] != $info[2]) {
+					unset($deprecation_options[$mode_key]);
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


hregis:
@lippoliv why to have 2 loop (foreach) for "$this->deprecation_options_fields" ?

lippoliv:
@hregis interesting spot. I didn't add it, just changed it. Seems like the logic was duplicated in `\AssetDepreciationOptions::setDeprecationOptionsFromPost` and `\AssetDepreciationOptions::fetchDeprecationOptions`.

I could extract a function from it, but I guess that probably could be better done by trying to bring improved architecture to the module?

_I tried to focus on the issue, rather than to refactor the code, where I have not much time yet spend reviewing. I don't want to increase maintainers workloads by commiting broken code._

hregis:
@lippoliv If the modification doesn't change the result, I think this code needs to be optimized.

lippoliv:
@hregis I contribute a fix towards v22 but you suggest to rewrite the logic, right? While I agree that the code could/should be refactored, that would better fit into `development` branch then, am I right? I could give it a try but I can't promise.

Or did you just mean to at least extract that logic into a standalone function so it's not repeated?

eldy:
> @hregis I contribute a fix towards v22 but you suggest to rewrite the logic, right? While I agree that the code could/should be refactored, that would better fit into `development` branch then, am I right? I could give it a try but I can't promise.
> 
> Or did you just mean to at least extract that logic into a standalone function so it's not repeated?

Yes refactoring must be done into develop branch. Into stable, only the lowest change to make a fix must de done.

@aspangaro do you validate the fix ?

hregis:
@lippoliv @eldy @aspangaro

Maybe before... It's possible to unified the loop?

hregis:
@lippoliv why to have 2 loop (foreach) for "$this->deprecation_options_fields" ?

aspangaro:
Hello,

Sorry, I don't have time to look at it today, I'll do it tomorrow, but yes, I can confirm that this is a problem that has been reported to me.
I had another fix to test that was similar but worked differently.
After that, I'll leave you to debate who has the best code...

I'll test it and get back to you.

The module is experimental, so I don't mind updating it to 22 as long as we fix it and don't introduce any new bugs.

There is also a date issue for annual depreciation; the allocation is carried over to the following fiscal year, for example.

I'll get back to you.
Have a great day!

eldy:
> @lippoliv why to have 2 loop (foreach) for "$this->deprecation_options_fields" ?

@hregis Here PR is a bug fix for the stable version v22. We must never introduce quality code enhancement when we try to fix bug. Such change are elligible for develop branch only. For fix, we must restrict the change to the minimum required to solve the fix, even if there is a cleaner solution (that could be done in develop branch).
See chapter https://wiki.dolibarr.org/index.php/Category:RoadMap#Merging_PR_process_.28develop_or_maintenance_.3F.29

hregis:
@eldy 
How do you manage to have such a broad perspective? 
I respect you for that... I'm going to follow the white rabbit...

hregis:
@eldy @aspangaro 
j'ai mis les premiers arobases que j'ai trouvés... 
je voulais en mettre d'autres... désolé !
j'espère que ces mots seront vus par le plus grand nombres...

Ça fait plus de 20 ans que j'ai rejoint Dolibarr avec Laurent et Rodolphe...
Quand je suis arrivé dans le projet je me suis dit... bordel ! il y a du boulot ! 😄 
c'était un trou de sécurité et pour ajouter des modules c'était... chiant ! comme Prestashop ! 😄 
J'ai à mon sens apporté pleins de choses... je n'ai pas envie de détailler sinon vous allez me poser des questions !  😄 
Bref... tout ça n'est pas important...
ce qui est important c'est vous !!
merci d'être là, merci d'y croire !
quand j'ai bossé la première fois sur Dolibarr on en était à la version 1.x (je ne sais même plus !)
je me souviens que Laurent avait fait une annonce incroyable pour la version 2.x mais je ne sais plus pourquoi, surement un fonctionnement incroyabale...
Bref... Merci à tout ceux qui croit en Dolibarr !
MERCI !

lippoliv:
@hregis thank god there's deepl :) Nice words, I can't really imagine how it feeld to work on such an amazing project for a decade (or two), must be strong 👍 

@aspangaro if you have a fix pending already, I can test at least my scenario with it, once you push it somewhere public. I'd probably prefer to go with your fix then, since you have more experience in the module than I (I'm just started using it a few days ago -.-).

@aspangaro @eldy @hregis thanks for the contribution here. I didn't thought it will take so much time. I really appreciate your involvement even if I don't have much stake in the game (yet) 🤟

aspangaro:
@eldy I confirm that this fix works correctly. It's perfect as it is. The data is correct in the database and responds correctly depending on whether the tax exemption option is checked or not. 

Thanks @lippoliv !

Issue #36514, opened 2025-12-03
Incorrect HT/TTC recalculation when editing proposal lines (Dolibarr 22.0.3)


badrdigital9:
### Bug

Hello,

I am experiencing an issue in Dolibarr 22.0.3 when editing lines in a commercial proposal.

Whenever I edit a line (Quantity), **the initial HT price replace The initial TTC price**, and Dolibarr recalculates a new HT value based on that incorrect number. This results in wrong totals.

Example
Original: 20,000.00 TTC → **16,666.67 HT**

After editing the line quantity or discount:
→ **16,666.67** becomes TTC
→ HT recalculated to 13,888.89 (incorrect)

How can i Fix this please ?

Thanks

### Dolibarr Version

22.0.3

### Environment PHP

8.2

### Environment Database

MariaDB 11.8.3

### Steps to reproduce the behavior and expected behavior

<img width="1605" height="388" alt="Image" src="https://github.com/user-attachments/assets/6b0cf945-508f-4912-b3f9-9f495f872442" />

BenjaminFlr:
Currently working on it.

badrdigital9:
Thank you

PR #36525, opened 2025-12-04
FIX(propal): Incorrect HT/TTC recalculation when editing proposal lines #36514

BenjaminFlr:
Related to https://github.com/Dolibarr/dolibarr/issues/36514

When we edit a line in propal, even if price base type was defined to TTC, we used the ht value to update it. The ht value bacame the TTC one after editing.

### DIFF
--- a/htdocs/comm/propal/card.php
+++ b/htdocs/comm/propal/card.php
@@ -19,6 +19,7 @@
  * Copyright (C) 2023		William Mead				<william.mead@manchenumerique.fr>
  * Copyright (C) 2024-2025	MDW							<mdeweerd@users.noreply.github.com>
  * Copyright (C) 2024		Alexandre Spangaro			<alexandre@inovea-conseil.com>
+ * Copyright (C) 2025		Benjamin Falière			<benjamin@faliere.com>
  *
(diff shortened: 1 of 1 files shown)


hregis:
@BenjaminFlr maybe it's ok with propal card... but with order, invoice or others ?

BenjaminFlr:
@hregis not sure to understand, here the only card concerned by the issue is the propal card. And I didn't notice the same behaviour somewhere else.

My change has no effect on other cards.

hregis:
@BenjaminFlr ok if this bug is in only in Propal no problem... Je reprend mon cerf volant et je choppe le vent... 😘

badrdigital9:
Thank you very much for your work and the quick fix.
I have applied the modification to the propal/card.php file on my instance and I confirm that the issue is now fully resolved on my side.

I really appreciate the reactivity and the clear explanation.

JonBendtsen:
> Thank you very much for your work and the quick fix. I have applied the modification to the propal/card.php file on my instance and I confirm that the issue is now fully resolved on my side.
> 
> I really appreciate the reactivity and the clear explanation.

then please close the bug @badrdigital9

PR #36526, opened 2025-12-04
time to init modules

frederic34:


hregis:
@frederic34 for you all

$xxxxx = array();

replace with

$xxxxx = [];

??

PR #36531, opened 2025-12-04
FIX : WebPortal responsive menu dropdown

thersane-john:
# FIX WebPortal responsive menu dropdown

On smart phone or tabs, menu drop down don't work, and responsible is "Claqué au sol""

The bug : 
image

This patch do this 

image
image

For PC 

image


BebZ:
a small correction : ...and responsive is "fucked up" :)

Thanks for the tip !

thersane-john:
@eldy why closed ?

eldy:
> @eldy why closed ?

Don t know why. There is no tag so i think it is an error.

PR #36540, opened 2025-12-05
FIX Accountancy - Reverse accounting account in depreciation expense entries for fixed assets

aspangaro:
Before:
2xxxxx at debit
6xxxxx at credit

Result : 

image


BurksEngineering:
Thank you! I just noticed this today.

PR #36553, opened 2025-12-05
Fix: Margin rounding error and division by zero

noec764:
In case of margin computing for an object with down payment on it (line type 9 not a project or service and negative amount). I had a problem off quasi division by zero due to a rounding error.
  
image

I fixed that in rounding pa and pv before testing if it was strictly superior to zero.

### DIFF
--- a/htdocs/core/class/html.formmargin.class.php
+++ b/htdocs/core/class/html.formmargin.class.php
@@ -171,12 +171,13 @@ class FormMargin
 		}
 
 		// Prevent rounding error and division by zero
-		$marginInfos['pa_products'] = (float) price2num($marginInfos['pa_products'], 'MT');
-		$marginInfos['pv_products'] = (float) price2num($marginInfos['pv_products'], 'MT');
-		$marginInfos['pa_services'] = (float) price2num($marginInfos['pa_services'], 'MT');
-		$marginInfos['pv_services'] = (float) price2num($marginInfos['pv_services'], 'MT');
-		$marginInfos['pa_total'] = (float) price2num($marginInfos['pa_total'], 'MT');
-		$marginInfos['pv_total'] = (float) price2num($marginInfos['pv_total'], 'MT');
+		$precision = getDolGlobalInt('MAIN_MAX_DECIMALS_SHOWN', 8);
+		$marginInfos['pa_products'] = (float) price2num($marginInfos['pa_products'], $precision);
+		$marginInfos['pv_products'] = (float) price2num($marginInfos['pv_products'], $precision);
+		$marginInfos['pa_services'] = (float) price2num($marginInfos['pa_services'], $precision);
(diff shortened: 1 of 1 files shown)


rycks:
@SylvainLegrand seems to be close to the bug you identified during devcamp no ?

lvessiller-opendsi (commented):
Could you check the total amount of margin (computed) ?
The value seems to be incorrect : you got 0.0000000 ? (see on your picture)
And the mlargin of product is 0 € and 536,80 € for services : so if you add it you will get 536,80 €.

Furthermore you got 0€ in cost price of your down payment line. For me, you will have a positive value.

noec764:
The problem you notice is not related to what I want to fix in this PR.

> The value seems to be incorrect : you got 0.0000000 ? (see on your picture)

Yes, it depends on how you decide to compute the margin, I didn't change the calculus of margin on this PR. I've only fix the division by zero error. 

As down payment line is neither a product nor service, it is not computed in total margin product, and total margin services, but it is in total, so it is the normal behavior and I don't want to change it for now.

lvessiller-opendsi:
> The problem you notice is not related to what I want to fix in this PR.
> 
> > The value seems to be incorrect : you got 0.0000000 ? (see on your picture)
> 
> Yes, it depends on how you decide to compute the margin, I didn't change the calculus of margin on this PR. I've only fix the division by zero error.
> 
> As down payment line is neither a product nor service, it is not computed in total margin product, and total margin services, but it is in total, so it is the normal behavior and I don't want to change it for now.

It's not obvious because the code to compute magins is also used when you insert discount and the buying price come from this method of "getMarginInfosArray". And the precision with "MT" decimals could be too small for buying price when you insert a discount line in an invoice.
Could you check if the round precision can be only put in margin total ?

`$marginInfos['total_margin'] = (float) price2num($marginInfos['pv_total'], 'MT') - (float) price2num($marginInfos['pa_total'], 'MT');
`

noec764:
```
		if ($marginInfos['pa_products'] > 0) {
			$marginInfos['margin_rate_products'] = 100 * $marginInfos['margin_on_products'] / $marginInfos['pa_products'];
		}
		if ($marginInfos['pv_products'] > 0) {
			$marginInfos['mark_rate_products'] = 100 * $marginInfos['margin_on_products'] / $marginInfos['pv_products'];
		}

		if ($marginInfos['pa_services'] > 0) {
			$marginInfos['margin_rate_services'] = 100 * $marginInfos['margin_on_services'] / $marginInfos['pa_services'];
		}
		if ($marginInfos['pv_services'] > 0) {
			$marginInfos['mark_rate_services'] = 100 * $marginInfos['margin_on_services'] / $marginInfos['pv_services'];
		}
```
As there is a division for all, we have to check for all of these fields. But maybe we could set to check with more decimals, I agree.

PR #36554, opened 2025-12-05
NEW: UX file drag and drop regression and remove click factory

thersane-john:
# Instructions

The new default behavior of not displaying the file form in the object file tab is inconvenient for users who need to upload many files daily.  

Clicking the "Add file" button currently triggers a click event on the input, which breaks the drag-and-drop functionality and forces the user to close the prompt window before they can drag and drop files.

# Fix 
- check if file is present or not before hide or display btn
- Remove auto click trigger 
- Add hidden conf to disable behavior

# Suggestion for improvement:
add drag-and-drop detection to automatically open .divattachnewfile.
**But here it's a fix not a new so for next time**


thersane-john:
@eldy c'est pas un new, tu veux une vidéo pour voir le problème ?

eldy:
> @eldy c'est pas un new, tu veux une vidéo pour voir le problème ?

The drag and drop is supported on the main tab, not on the document tab. The drag and drop for this page is not implemented (yet)
An implementation must be introduced to work as the main tab.
Yes, a video may be welcome...

thersane-john:
I'm not talking about the same thing. I'm talking about the default drag and drop behavior in browsers when dealing with file input fields.

see your discord private message for  the vidéo

eldy:
Weare speaking of the same 

> I'm not talking about the same thing. I'm talking about the default drag and drop behavior in browsers when dealing with file input fields.
> 
> see your discord private message for the vidéo

I was talking of the same thing.
This is not a bug but a change into the behaviour on how you can upload the files. Implementation in v22 was not yet complete due to lack of time because feature was working correctly, even if it was not the final version, but was completely finished in v23.

With the finished implementation, you can drag the file anywhere on the tab. You don't need to open the input box. This will automatically upload all the files with no click at all, so it works as it works on the main tab. See the video as example that upload 2 files at once with a drag and drop of the 2 files onto the tab "Linked files" of a product.

Note that this target implementation is not yet finished anywhere but is complete on products, this is why the example is provided for products.

https://github.com/user-attachments/assets/ae6e7473-1b9d-446e-99e7-f7572e2f1de0

thersane-john:
Ok what's why it was wierd

PR #36568, opened 2025-12-07
FIX #36312 Prevent circular reference in category parent assignment

### LINKED ISSUE #36312: llx_categorie: category can reference itself as its own parent (fk_parent = rowid)
### Bug

In the table llx_categorie, it is currently possible to set the field fk_parent with the same value as rowid, meaning a category can become its own parent.

### Dolibarr Version

20.0

### Environment PHP

8

### Environment Database

_No response_

UPDATE llx_categorie
SET fk_parent = rowid
WHERE rowid = 1;

### Attached files

_No response_

Steph501:
## Summary
- Add `isDescendant()` method to Categorie class to check if a category is a descendant of the current one
- Add validation in `create()` and `update()` methods to prevent setting a descendant as parent
- Add translation keys for error message (EN/FR)

Fixes #36312

## Test plan
- [x] Normal category creation/update works without errors
- [x] API call trying to set a descendant as parent returns error 500 with message "ErrorCategoryParentIsDescendant"
- [x] Interface already filters the selector, this fix adds server-side protection for API and imports


Steph501:
The PHPStan failure is not related to this PR.

The errors are:
- `Property CommonObject::$shipping_method_id (int) in isset() is not nullable.`
- `Property Propal::$availability_id (int) in isset() is not nullable.`

These are pre-existing issues in `CommonObject` and `Propal` classes, not in the category code modified by this PR.

Issue #26864, opened 2023-11-27
Link member to third party - no check / no feedback


dolibit-ut:
### Bug

If you link a (second) member to a third party you get no warning, that you changed an existing link to another member.

I see three possible solutions here:
A) No new link possible & a warning message that there is an existing link and therefore no new one can be created.
B) A new link is created, but you receive a message that an existing link has been deleted.
or best way
C) You get displayed that a link exists and you are asked whether it should remain or be overwritten.

### Environment Version

18

JonBendtsen:
Solution C would not work through the API, and that is how I link members and Thirdparties.

dolibit-ut:
> Solution C would not work through the API, and that is how I link members and Thirdparties.

Not really. 
You can define that API has two options - overwrite or not
and this is a question you can create at GUI.

JonBendtsen:
> > Solution C would not work through the API, and that is how I link members and Thirdparties.
> 
> Not really. You can define that API has two options - overwrite or not and this is a question you can create at GUI.

True, that would work

JonBendtsen:
This issue is still needed

JonBendtsen:
Issue also applies to dolibarr v20.0.0, no warning is given, and the link is just overwritten

JonBendtsen:
API does "check", it says:

```
{
  "error": {
    "code": 500,
    "message": "Internal Server Error: Error when updating member: Duplicate entry '453' for key 'uk_adherent_fk_soc'"
  }
}
```

JonBendtsen:
We should re-test if this is still an issue

JonBendtsen:
It is still an issue in 22.0.3

# PUT non existing socid to existing member
```
{  "socid": "22"}
```
## result 
```
{
  "error": {
    "code": 500,
    "message": "Internal Server Error: Error when updating member: Cannot add or update a child row: a foreign key constraint fails (`dolidb`.`llx_adherent`, CONSTRAINT `adherent_fk_soc` FOREIGN KEY (`fk_soc`) REFERENCES `llx_societe` (`rowid`))"
  }
}
```

We should probably test if the socid exists before using

# Put with same socid to same socid. (with full permissions to all thirdparties)
Works fine

# BAD - should be fixed

Put to existing member with an existing link to a thirdparty with another existing thirdparty, but a thirdparty that the user has no rights to, and that user has no `Extend to all thirdparties` permissions just changes the member link 🚱

JonBendtsen:
[...]

PR #36570, opened 2025-12-07
FIX Link member to third party - no check, no feedback 26864

JonBendtsen:
# FIX Link member to third party - no check, no feedback 26864 - BREAKING CHANGE
This PR is a partial fix of bug 26864, it only handles the API. The bug was that one could silently overwrite an existing link between memberA and a thirdpartyA by specifying a new link to thirdpartyB. This PR introduces bans direct thirdparty links when there is an existing link and instead reports back that the user has to use a 2 step process by first setting the link to 0, aka remove the link, and then introduce the new link. 

# Breaking Change
You can no longer just update socid using the API, you have to use a 2 step process by first setting it to 0, and then setting it to the new value

### DIFF
--- a/htdocs/adherents/class/adherent.class.php
+++ b/htdocs/adherents/class/adherent.class.php
@@ -1004,15 +1004,26 @@ class Adherent extends CommonObject
 						$lthirdparty->default_lang = $this->default_lang;
 
 						$result = $lthirdparty->update($this->fk_soc, $user, 0, 1, 1, 'update'); // Use sync to 0 to avoid cyclic updates
-
 						if ($result < 0) {
-							$this->error = $lthirdparty->error;
+							if ($lthirdparty->error) {
+								$this->error = $lthirdparty->error;
+								dol_syslog(get_class($this)."::update::error=".$this->error, LOG_ERR);
+							} else {
+								$this->error = join(',', $lthirdparty->errors);
+								dol_syslog(get_class($this)."::update::errors=".$this->error, LOG_ERR);
+							}
 							$this->errors = $lthirdparty->errors;
-							dol_syslog(get_class($this)."::update ".$this->error, LOG_ERR);
+							dol_syslog(get_class($this)."::update::lthirdparty->errors=".join(',', $this->errors), LOG_ERR);
 							$error++;
 						}
 					} elseif ($result < 0) {
-						$this->error = $lthirdparty->error;
+						if ($lthirdparty->error) {
+							$this->error = $lthirdparty->error;
+							dol_syslog(get_class($this)."::update::error=".$this->error, LOG_ERR);
+						} else {
+							$this->error = join(',', $this->errors);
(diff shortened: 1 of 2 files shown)


JonBendtsen:
I've run into some questions that needs to be determined.
 
# First question
I'm adding a check for users access to the thirdparties included, and that is always granted. I think this is due to this check missing in v18, that I only added in v22. https://github.com/Dolibarr/dolibarr/pull/36408/files

Do we want to backport that fix to v18, do we want to ignore it? Or do we want to fix is from some other version between v18 and v22?

# Second question
At first If I try to break the thirdparty link with a PUT with a socid=0, then nothing happens, most likely this right here:
https://github.com/Dolibarr/dolibarr/blob/18.0/htdocs/adherents/class/adherent.class.php#L823

Do we want to support the API breaking a link to a thirdparty? Either using the PUT or do we really want to make a special endpoint just for that? So far I "fixed" it in PUT with using $member->setThirdPartyId(0) but we could probably also remove that check in Line 823

## sub decision
I decided to ban direct from thirdpartyA to thirdpartyB updates, forcing a 2 step process by first resetting and removing the link with socid=0, and then the 2. step would be to update again with the new thirdparty value.

# test cases
memberA and thirdpartyA not linked both with emailA + thirdpartyB with different email

## works
from any socid, update memberA with socid=0
from socid=0, update memberA with thirdpartyA or thirdpartyB works fine

## fails
from socid=thirdpartyA, update memberA with thirdpartyB fails, stating use socid=0
from socid=thirdpartyB, update memberA with thirdpartyA fails, stating use socid=0

## work arround
first set it to socid=0 and then to new socid

rycks (commented):
hello @JonBendtsen 
i'm sorry but i'm quite sure that such PR could not be integrated into 18.0 because as you says it introduce a break ... i share your point of view "that's a bug" but as a merger 
we can't apply it i think ... but maybe @eldy will have an other point of view ?

https://wiki.dolibarr.org/images/d/d9/Chart_of_decision.png

JonBendtsen:
> hello @JonBendtsen i'm sorry but i'm quite sure that such PR could not be integrated into 18.0 because as you says it introduce a break ... i share your point of view "that's a bug" but as a merger we can't apply it i think ... but maybe @eldy will have an other point of view ?
> 
> https://wiki.dolibarr.org/images/d/d9/Chart_of_decision.png

Hey @rycks. I appreciate that you are concerned about breaking changes, I am too, which is why I mentioned that this was a breaking change. I'll wait for @eldy's input before making any changes.

JonBendtsen:
Hey @eldy 

Can we get a resolution here?

1. Do you disagree that https://github.com/Dolibarr/dolibarr/issues/26864 is a bug?
2. Do you disagree that it should be fixed?
3. Do you disagree that it should be fixed in v18?
4. Do you think that my proposed fix is not good enough?

PR #36572, opened 2025-12-08
FIX Accountancy - Problem calculating fiscal year with different periods on fixed assets

aspangaro:
List of fiscal year:

Capture d’écran du 2025-12-07 12-53-37

On asset:

Capture d’écran du 2025-12-07 12-48-58

After correction:
image

**### Additional tests in progress !** Need the PR #36522 to test with the tax exemption option is checked or not.

All test are ok for my part

### DIFF
--- a/htdocs/asset/class/asset.class.php
+++ b/htdocs/asset/class/asset.class.php
@@ -1184,13 +1184,30 @@ class Asset extends CommonObject
 
 					// Next fiscal period (+1 day/month/year)
 					$fiscal_period_start = dol_time_plus_duree($fiscal_period_end, 1, 'd');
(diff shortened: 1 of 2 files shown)


josett225:
Hi @aspangaro 
Very strange as I went into this issue some time ago that I fixed partially. 
I will try your new code. 

I believe my previous code worked well only when the previous FY (and first one) was based in 12 month.

aspangaro:
Hi @josett225,

The problem only occurs in annual mode. In monthly mode, there is no problem, as far as I can tell.

Have a great day !

aspangaro:
All test are ok for my part

PR #36579, opened 2025-12-08
CLOSE #35561 End-of-month calculation functionality for recurring invoices

### LINKED ISSUE #35561: End-of-month and end-of-year calculation functionality for recurring invoices
### Bug

With reference to #32060 . A fix was posted to achieve this but it doesn't work as intended, i.e. end-of-month dates are not calculated.

### Dolibarr Version

22.0.2

### Environment PHP

8.4.11

### Environment Database

MariaDB 11.8.3

See below example for a recurring invoice:

Image

If the last generated date was 30/09/2025, the next one should be on 31/10/2025. The only workaround at the moment is to set the generation data as 28th of the month and then come back each real last day of the month to correct the invoice.

### Attached files

_No response_

apio-sys:
# CLOSE #35561 End-of-month calculation functionality for recurring invoices

If you set a day >= 28, the invoice will be generated on the last day of each month (end-of-month mode). For example, setting the 31st will generate invoices on Jan 31, Feb 28/29, Mar 31, Apr 30, etc.

This modification will make sure that if a day for a recurring monthly invoice is set >= 28, it is considered to be for the end of each coming month and schedules the next generation accordingly. This could be subject to discussion of course for say someone that always wants to invoice on day 28. I think this is the simplest compromise rather than having a separate month-end function/option in the recurring invoice template. Some help text have been added to make the user aware of this functionality.

Screenshot from 2025-12-08 21-09-01


apio-sys:
Can this be reopened Laurent @eldy please? I'm not able to from my end (there is already an open pull request from apio-sys:develop to Dolibarr:develop).

eldy:
> Can this be reopened Laurent @eldy please? I'm not able to from my end (there is already an open pull request from apio-sys:develop to Dolibarr:develop).

The PR is empty, so it is better to make another one.

apio-sys:
I have re-created a PR for this @eldy here : https://github.com/Dolibarr/dolibarr/pull/37234 .

PR #36583, opened 2025-12-09
FIX Accountancy - No limit on chart of account to define category on account

aspangaro:
I have two accounts define in category account in accountancy

Capture d’écran du 2025-12-09 06-39-26

I enter to see the list and the result is 12 accounts:

Capture d’écran du 2025-12-09 06-38-35

in my database:
Capture d’écran du 2025-12-09 06-38-54

Need to limit the list of category display only on the chart of account defined

### DIFF
--- a/htdocs/accountancy/class/accountancycategory.class.php
+++ b/htdocs/accountancy/class/accountancycategory.class.php
@@ -443,9 +443,8 @@ class AccountancyCategory // extends CommonObject
 
 		$sql = "SELECT t.rowid, t.account_number, t.label";
 		$sql .= " FROM ".$this->db->prefix()."accounting_account as t";
-		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."accounting_system as asy ON t.fk_pcg_version = asy.pcg_version AND t.entity = ".((int) $conf->entity);
-		$sql .= " WHERE asy.rowid = ".((int) $pcgver);
-		$sql .= " AND t.fk_accounting_category = ".((int) $id);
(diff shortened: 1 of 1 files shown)


htdocs/accountancy/class/accountancycategory.class.php
```diff
@@ -438,9 +438,14 @@ public function delete($user, $notrigger = 0)
...
 	{
 		global $conf;
+
+		$pcgver = getDolGlobalInt('CHARTOFACCOUNTS');
+
 		$sql = "SELECT t.rowid, t.account_number, t.label";
 		$sql .= " FROM ".$this->db->prefix()."accounting_account as t";
-		$sql .= " WHERE t.fk_accounting_category = ".((int) $id);
+		$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."accounting_system as asy ON t.fk_pcg_version = asy.pcg_version AND t.entity = ".((int) $conf->entity);
+		$sql .= " WHERE asy.rowid = ".((int) $pcgver);
```
eldy: IF you have a WHERE, the LEFT JOIN is useless and can be replaced with INNER JOIN
The where here must be a AND in the previous line of ON ... 
Also test on t.entity must not be into the condition of the LEFT JOIN but into the WHERE.
so
```
$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."accounting_system as asy ON t.fk_pcg_version = asy.pcg_version  AND asy.rowid = ".((int) $pcgver);
$sql .= " WHERE t.entity = ".((int) $conf->entity) 
...
```

PR #36585, opened 2025-12-09
FIX undefined variables on create invoice card from order (backport f…

lvessiller-opendsi:
FIX undefined variables on create invoice card from order (backport from v19)
- see PR #26825
- see commit 69f66f2

Create a customer invoice from a customer order and you will get a lot of warnings and then an error :
- upstream sent too big header while reading response header from upstream,


rycks (approved):

rycks:
as it's the solution in dolibarr 19.0 i apply that PR :)

PR #36602, opened 2025-12-10
NEW (QoL): better default value (copy-paste ready) for $pathtoscript in CRON setup page

atm-florianm:
# NEW

Currently, when you set up scheduled jobs, you see this, unless you have set up the hidden conf `MAIN_DOL_SCRIPTS_ROOT`.
> 
image

If this PR is accepted:
- users with a Dolibarr install that reflects the original directory structure from the repository (`dolibarr/htdocs`) ⇒ the CRON line will now be copy-paste ready even without `MAIN_DOL_SCRIPTS_ROOT`.
- users with a Dolibarr deployed differently (e.g. deb package) ⇒ no difference


htdocs/core/lib/cron.lib.php
```diff
@@ -130,7 +130,10 @@ function dol_print_cron_urls()
 
 	print '<u>'.$langs->trans("FileToLaunchCronJobs").':</u><br>';
 
-	$pathtoscript = '/pathtoscript';
+	$pathtoscript = realpath(dirname(__DIR__, 3).'/scripts/cron');
```
eldy: This expose to web user the path structure on server where only admin sys know where the script dir is stored.
This is why we show /pathtoscript.

PR #36603, opened 2025-12-10
FIX - False Def Bom on product clone

atm-lucasmantegari:
# FIX - False Def Bom on product clone
Currently, regardless of whether I check the default clone BOM or not, it is always retrieved during confirm_clone.
The behavior with this change becomes:
I clone without checking => creation of a new  BOM identical to the default one for the cloned product but not the same (to avoid confusion). Is this the expected behavior?

### DIFF
--- a/htdocs/product/card.php
+++ b/htdocs/product/card.php
@@ -994,7 +994,7 @@ if (empty($reshook)) {
 						if (!$error && isModEnabled('bom') && $user->hasRight('bom', 'write')) {
 							$defbomidac = 0; // to avoid cloning same BOM twice
 							$clone_defbom = GETPOSTISSET('clone_defbom') ? 1 : 0;
-							if($clone_defbom && $object->fk_default_bom > 0) {
+							if ($clone_defbom && $object->fk_default_bom > 0) {
 								$bomstatic = new BOM($db);
 								$bomclone = $bomstatic->createFromClone($user, $object->fk_default_bom);
 								if ((int) $bomclone < 0) {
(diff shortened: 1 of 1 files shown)


atm-lucasmantegari:
@eldy it's better ?

atm-lucasmantegari:
@eldy 

You are right that for values '', '0' and '1' the behavior is the same.

The change is only meant to avoid PHP truthy values (like 'on', 'true' or any non-empty string) from always triggering the clone logic.

Casting the input to int makes the intent explicit and ensures the checkbox only enables cloning when a valid numeric value is provided.

atm-lucasmantegari:
@eldy

eldy:
I see no difference in your new code and the old one.

Both are testing the same var.
If request to clone bom is on, we should, clone the bom, set it as default bomnok product and set the new product as product to produce on the new bom.
If not checked and the product must have no default bom.

Note: If value of input fieldb is on off, the getpost will return the same effect. This is why value must be 1 or 0 and parm should be retreived with GETPOSTINT.

atm-lucasmantegari:
@eldy 
The problem was not only related to cloning the default bill of materials, but also to copying the link to the default bill of materials when cloning the product.
Previously, even when the box was not checked, the cloned product could retain the fk_default_bom nomenclature of the source product.
This change makes the behaviour explicit:
the default nomenclature is only cloned if the box is checked
otherwise, the fk_default_bom of the cloned product is reset to 0
I have also made the checkbox handling more robust with the confirmation form (hidden value 0 + ‘on’) and replaced clone_otherboms with GETPOSTINT() as requested for consistency.
This ensures that the cloned product never retains a default BOM unless explicitly requested.

atm-lucasmantegari:
@eldy

atm-lucasmantegari:
@eldy

atm-lucasmantegari:
@eldy Sorry for the spam

PR #36605, opened 2025-12-10
FIX - Holidays negative balance

atm-lucasmantegari:
# FIX - Holidays negative balance
We block everything with "HOLIDAY_DISALLOW_NEGATIVE_BALANCE" OR we block according to the dictionary.


eldy:
Close conflict

Issue #34050, opened 2025-05-02
TraceableDB: Add backtrace report in case of database query failure


alexandre-janniaux:
### Feature Request

Currently, a database error will produce this kind of output:

<img width="1769" alt="Image" src="https://github.com/user-attachments/assets/610c4340-1001-4749-8625-c21e3d57f51e" />

But then, we need to research which file and line produced the query.

The goal would be to be able to attach backtrace directly to records done from TraceableDB, and then being able to display them from the debugbar there:

<img width="1769" alt="Image" src="https://github.com/user-attachments/assets/5238ace7-e43a-4385-80fc-446d6be3af73" />

### Use case

I'm doing a lot of mysql to postgres changes to make some modules and core part compatible with postgresql, but I struggle a lot to quickly find the incriminating code generating the query. I also do some module development and it would make module development a lot smoother.

### Suggested implementation

```
protected function startTracing()
	{
		$this->startTime   = microtime(true);
		$this->startMemory = memory_get_usage(true);
	}
```

We can already record the backtrace there using debug_backtrace()

alexandre-janniaux:
Pinged on the merge request.

PR #36612, opened 2025-12-11
NEW debugbar: Add backtrace capture for database query failures

alexandre-janniaux:
# NEW debugbar: Add backtrace capture for database query failures #34050
Implements backtrace forwarding to capture and display backtraces when
database queries fail, making it easier for developers to identify the
source of database errors.

Backtrace are captured in TraceableDB::endTracing() when queries fail
using debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS) and stored to be
fowarded to the DolQueryCollector that will expose it to the debugbar
integration.

Because of that, the backtrace is now captured at the correct location,
ie. when the query fails, rather than when the error is retrieved,
ensuring it shows the actual application code that triggered the
problematic query.

image

This is particularly useful for debugging MySQL to PostgreSQL
compatibility issues, module development and maintenance, and quickly
identifying query origins during development. In particular, paired with
the list and time profiling of each request, it will allow finding the
location where queries are not properly optimized.

Before this patch, it was possible to try and find snippets from the
queries but because queries are generated at runtime, it was tedious and
could lead to the wrong location.

Related to #34050

### DIFF
--- a/htdocs/debugbar/class/DataCollector/DolQueryCollector.php
+++ b/htdocs/debugbar/class/DataCollector/DolQueryCollector.php
@@ -71,7 +71,8 @@ class DolQueryCollector extends DataCollector implements Renderable, AssetProvid
 				'memory' => $query['memory_usage'],
 				'is_success' => $query['is_success'],
 				'error_code' => $query['error_code'],
-				'error_message' => $query['error_message']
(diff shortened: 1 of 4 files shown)


alexandre-janniaux:
Ping @eldy :)

PR #36642, opened 2025-12-14
Problem on request for category on account

aspangaro:
Refer to this fix #36583 , your proposition not working @eldy 

Modify request sql for working. Place the condition in the WHERE clause rather than in the LEFT JOIN.

With your proposition:

image

With fix:
image


eldy:
So if you need to move a criteria on a table that is left join into the where, it means what you want is NOT lines with link on criteria + line with no links (case a), but what you want is only lines with links on criteria (case b).
In this case, it means you must not use a left join but a inner join ( my suggestion was to answer need a, but if good screenshot is second one it means you need b so no left join at all).
You must choose left join and all condition on fields of tables on the the left join line (a)  Xor a inner join (b).
So try b. You pr is mix between a and b.

aspangaro:
@eldy thanks I used solution b

PR #36654, opened 2025-12-15
FIX missing search tasks extrafields input and title

hregis:


hregis:
@eldy travis error is not for my PR

PR #36656, opened 2025-12-15
NEW product barcode label print options

W1W1-M:
# NEW product barcode label print options
If a product is selected for barcode generation, 2 extra options will be available for the user:
- Add product reference to barcode label header
- Add product label to barcode label footer

If selected these options override the `BARCODE_LABEL_HEADER_TEXT` or `BARCODE_LABEL_FOOTER_TEXT` constants.

If needed user can: 
- manually input text to use custom ref or label
- modify ref or label fetched from product to use customized text

Print barcodes view with new options:
image

Example of generated barcode with these options:
image


W1W1-M:

PR #36660, opened 2025-12-15
NEW action button to print product barcodes

W1W1-M:
# NEW action button to print product barcodes

If barcode module is activated, user can open the barcode print sheet directly from product card. Print sheet is preselected with product.

Capture d’écran 2025-12-15 à 12 09 02


eldy:
My undestanding is that button is not a workflow action on status but a link to anohter page that is preselected.
So i think it would be better to replace the button, with a small link "Print barcode" just after the "Barcode value" like we have on this screenshot for the "Stock at date" or "Virtual stock" just after the value of stock :
image

W1W1-M:
> My undestanding is that button is not a workflow action on status but a link to anohter page that is preselected. So i think it would be better to replace the button, with a small link "Print barcode" just after the "Barcode value" like we have on this screenshot for the "Stock at date" or "Virtual stock" just after the value of stock : image

@eldy Thanks for feedback. Refactored action button to link on barcode value row:
Capture d’écran 2025-12-15 à 17 32 28

eldy:
Thanks. Pr is postponed for the moment, will be validated once branch v23 has been created (next step will be to add a larger space and add the picto used for links for the look and feel v24)...

W1W1-M:

eldy:
> @eldy  any chance this can be merged into dev branch now ? Thanks

As this is not a workflow action, buttonnmist not be withvaction button. Instead can you add a link "Print barcode sheet" at the right of the barcode field ?
See example on tab stock for the link to last stock movement.

PR #36661, opened 2025-12-15
FIX - Can Validated Holidays When Negative Balance

atm-lucasmantegari:
# FIX - Can Validated Holidays When Negative Balance
Exactly the same [PR](https://github.com/Dolibarr/dolibarr/pull/36144) but for validate


eldy:
the pr #36144 was fixed to use 
num_open_day instead of num_between_day
i suppose you should do the same here.

parenthesis on test are also expected to avoid ci alerts

atm-lucasmantegari:
@eldy Fix OK !

PR #36667, opened 2025-12-16
FIX - Add mapping for 'produit' to 'product' in isModEnabled function

omogenot:
This allows 'old' product related widget boxes to show up.

FIX - Convert old name 'produit' to 'product'

Some widget box code still contain the 'produit' word and would not show up in widget list nor main page.

ex.: `box_produit_alerte_stock.php`

### DIFF
--- a/htdocs/core/lib/functions.lib.php
+++ b/htdocs/core/lib/functions.lib.php
@@ -385,6 +385,7 @@ define(
 		'propale' => 'propal', // Has old directory
 		'socpeople' => 'contact', // Has old directory
 		'fournisseur' => 'supplier',  // Has old directory
+		'produit' => 'product',	// Prevented boxes to display
 
 		'actioncomm' => 'agenda',  // NO module directory (public dir agenda)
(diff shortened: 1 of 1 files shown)


eldy:
It seems that all mention to produit has already been removed into the widget, making the addition of the exception useless now. Can you point me to the line the still refer to "produit" that need your suggested change ?

omogenot:
> It seems that all mention to produit has already been removed into the widget, making the addition of the exception useless now. Can you point me to the line the still refer to "produit" that need your suggested change ?

The widget file refers to produit: `box_produit_alerte_stock.php` in the beginning of the file @ line 42:
```
	public $depends = array("produit");

```
Anyway, if you perform the modification I propose, this widget (and some others which probably have the same issue) shows up correctly, otherwise it does not appear in the Configuration - Widgets  list, nor in the Main window widget listbox.

eldy:
Thanks for feedback.
Because the goal is to remove exceptions and standardize name to english version, i have pushed a different fix that avoid to increase the list of exceptionto manage legacy names.

PR #36670, opened 2025-12-16
NEW - Add substitution for user warehouse

BB2A-Anthony:
# NEW|New [*Add substitution for user warehouse*]
[*Long description*]


BB2A-Anthony:
@eldy Ok for you?

BB2A-Anthony (commented):
> Some code or existing module may use fk_warehouse. So can you keep the property to avoid errors for them
> You can add the comment:
> 
>  @deprecated use $warehouse_id 


@eldy it's done

mdeweerd:
> > Some code or existing module may use fk_warehouse. So can you keep the property to avoid errors for them
> > You can add the comment:
> > @deprecated use $warehouse_id
> 
> @eldy it's done

FYI: IMHO it's better to use the DolDeprecationHandler for this by extending the attribute mapping.
In case somebody forgets to also assign warehouse_id, this is not an issue because both will refer to the same  property.

```php
    /**
     * Override the method to include all parent properties PLUS the new warehouse mapping.
     *
     * @return array<string,string>
     */
    protected function deprecatedProperties()
    {
          $extraMappings = array(
              'warehouse_id' => 'fk_warehouse',
          );
          return parent::deprecatedProperties() + $extraMappings;
    }
```

PR #36672, opened 2025-12-16
NEW: UIUX|Uiux Drag & drop ordering on project referrers

braito4:
his PR adds drag & drop reordering on the Project “Linked elements” overview (proposals, orders, invoices, supplier documents, shipments, etc.) and on project tasks (where supported).

Introduces a new table llx_projet_elementorder to persist a per-project ordering of linked elements.
Adds an Ajax endpoint htdocs/projet/ajax/reorderreferrers.php to save the new order securely (token required).
Updates the project overview page (htdocs/projet/element.php) to:
enable tableDnD when the user has project write rights and not on phone layout,
apply the saved order when displaying linked elements,
prevent subtotal/footer/header rows from being draggable.
Adds the DB migration for the new table in htdocs/install/mysql/migration/22.0.0-23.0.0.sql.

### DIFF
--- a/htdocs/install/mysql/migration/22.0.0-23.0.0.sql
+++ b/htdocs/install/mysql/migration/22.0.0-23.0.0.sql
@@ -390,6 +390,22 @@ CREATE TABLE llx_expensereport_det_extrafields
 	import_key                varchar(14)
 ) ENGINE=innodb;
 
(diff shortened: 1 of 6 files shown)


eldy:
Before submitting new code when there is SQL change or addition in the structure, a PR with just the SQL change and migration script must be submitted as a standalone PR, so the database structure will be validated first.

PR #36677, opened 2025-12-16
NEW proj: add table to persist project referrers ordering

braito4:
This PR introduces a new database table to persist a user-defined ordering for the different elements linked to a project (proposals, orders, invoices, supplier documents, etc.) as shown in the Project module “Referrers/Elements” view.

Because these elements belong to many different modules/tables and are only aggregated inside the Project module (they don’t share a common “position/rank” column), we can’t store the ordering directly on each source table. To support drag & drop ordering consistently across all project-linked element types, we add an auxiliary table (llx_projet_elementorder) that stores the rank per project and per element type, keyed by (entity, fk_projet, elementtype, fk_element), with an index to efficiently retrieve ordered lists.


eldy:
What do you want to order :
Having the proposal block before the invoice section before the order section, ... or is it to order lines inside a given block?

braito4:
With this PR I want to make each of the listings (invoices, orders, etc.) manually sortable.

In my workflow, I use the “items” tab to check the status of a project.
Supplier orders, for example, are started based on priority: if a supplier says no, when I add the next order it goes to the bottom of the list, often mixed with orders that are at a much later stage in the process.

This PR allows modifying the list by moving orders around however you want (as already happens with quotes).

eldy:
> With this PR I want to make each of the listings (invoices, orders, etc.) manually sortable.
> 
> In my workflow, I use the “items” tab to check the status of a project. Supplier orders, for example, are started based on priority: if a supplier says no, when I add the next order it goes to the bottom of the list, often mixed with orders that are at a much later stage in the process.
> 
> This PR allows modifying the list by moving orders around however you want (as already happens with quotes).

I have understood that goal is to change list but it is not clear list of what ?
List of sections (section invoice before/after the proposal section before/after the order section),
or list of the lines inside a given section ?

PR #36695, opened 2025-12-18
FIX: global search, list view new search is not working

splohmer:
FIX: global search, list view new search is not working
If you enter something in the global search with an input for example for products where no results are shown, you are in some kind of deadlock. You can enter new search values in the list view, but in the background search_all is added as hidden parameter which prevents the new search.

### DIFF
--- a/htdocs/core/tpl/massactions_pre.tpl.php
+++ b/htdocs/core/tpl/massactions_pre.tpl.php
@@ -67,12 +67,6 @@
 @phan-var-force string $sendto
 ';
 
-if (!empty($sall) || !empty($search_all)) {
-	$search_all = empty($sall) ? $search_all : $sall;
-
-	print '<input type="hidden" name="search_all" value="'.$search_all.'">';
-}
-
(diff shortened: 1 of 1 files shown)


splohmer:
Additional information: this was introduced in Version 19.
Commit was https://github.com/Dolibarr/dolibarr/commit/9beb210ead2367f9d986082b78b9d461609766b6

Currently it is not clear to me why this was added.

eldy:
This seems to be the expected behaviour
If you made a search on global filter with abc, and after you want to increase the filter by adding another term, the global filter is kept as shown into the sentence "Search criteria 'abc' into fields ..." to remind the additional filter.
If you want to make another search from scratch, just remove the global search by resetting filters on clicking on the cross.

splohmer:
Hi @eldy,
thanks for pointing this out. It was not really clear to me what the intension of this code change was to previous dolibarr verions. I understood that this is for further drilling down the search if coming from the global search. Therefore it makes defintiv sense to keep this as it is. No code change needed.

Regards,
Sven

PR #36699, opened 2025-12-18
FIX document API agenda event list and download

W1W1-M:
# FIX document API agenda event list and download
Documents uploaded to agenda events objects historically have `actioncomm` value for `src_object_type`. When trying to download these documents through API it is necessary for the `dol_check_secure_access_document` method to check the modulepart against `actioncomm`.

Also fixed directory of agenda event documents

### DIFF
--- a/htdocs/api/class/api_documents.class.php
+++ b/htdocs/api/class/api_documents.class.php
@@ -601,7 +601,6 @@ class Documents extends DolibarrApi
 				throw new RestException(404, 'Event not found');
 			}
 
-			$upload_dir = getMultidirOutput($object) . '/'.dol_sanitizeFileName($object->ref);
 			$upload_dir = $conf->agenda->multidir_output[$object->entity ?? $conf->entity].'/'.get_exdir(0, 0, 0, 1, $object, 'agenda');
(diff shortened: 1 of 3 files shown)


htdocs/api/class/api_documents.class.php
```diff
@@ -601,7 +601,7 @@ public function getDocumentsListByElement($modulepart, $id = 0, $ref = '', $sort
 				throw new RestException(404, 'Event not found');
 			}
 
-			$upload_dir = getMultidirOutput($object) . '/'.dol_sanitizeFileName($object->ref);
+			$upload_dir = $conf->agenda->multidir_output[$object->entity ?? $conf->entity].'/'.get_exdir(0, 0, 0, 1, $object, 'agenda');
```
eldy: Is it possible to fix inside the getMultidirOutput ?
Because future is to replace all things like 
`$conf->agenda->multidir_output[$object->entity ?? $conf->entity].'/'.get_exdir(0, 0, 0, 1, $object, 'agenda');`
into
`getMultidirOutput($object) . '/'.dol_sanitizeFileName($object->ref);`

PR #36707, opened 2025-12-19
FIX|Fix # Add fr langs trans

atm-corentin:
# FIX|Fix #* Add fr langs trans*

Some language translations were declared in the US languages but not in the French ones.


atm-corentin:
@eldy can you check pls ?

PR #36714, opened 2025-12-19
Fix: from_template sender without id on first email form display

jeremy-my:
# Fix from_template sender without id on first email form display

On the first display of the email send form, the variable modelmailselected is not yet posted.
The sender profile key was built using GETPOST('modelmailselected'), which could result in a from_template_ key without an identifier.

This prevents the mail sending logic from correctly resolving the email_from value defined in the email template, as the sending code expects a from_template_<id> format.

This fix replaces the usage of POST data with the already computed $model_mail_selected_id, which is the actual source of truth for the selected email template.
This ensures that a valid from_template_<id> key is always generated, even on the first display of the form.

Manual tests performed:
- Created an email template with the email_from field set
- Opened the email send form without clicking “Apply”
- Verified that the sender is correctly selected and the email is sent using the template sender


eldy:
A fix in conflict with this one was already merged few days ago.
Can you check everything is ok without your fix now ?

jeremy-my:
My mistake on my side my fork wasn’t up to date. It’s also possible the changes hadn’t been pushed yet at the time. In any case, I’ve tested it now and it works perfectly :)

PR #36716, opened 2025-12-19
NEW Adds the option to hide payment with BAN on the Sponge invoice template

nat-itopie:
NEW Adds the option to hide payment with BAN on the Sponge invoice template even if the payment method is set to bank transfer.


eldy:
The ban is shown if in setup of module invoice, you set the option "show ban number"
You should already be able to not show any ban if you change the setup in invoice module.

nat-itopie:
> The ban is shown if in setup of module invoice, you set the option "show ban number" You should already be able to not show any ban if you change the setup in invoice module.

Actually, my comment wasn't exhaustive.
The setting you mention and FACTURE_RIB_NUMBER allow the account number to be displayed even if no payment method is specified above.
However, if the payment method is set to bank transfer, which we need to do for another module that generates Swiss QR slips for transfers with a unique reference, then the account is automatically displayed at the bottom of the table. However, we do not want to display it because the IBAN is only valid for payments by QR slip with a unique reference and not for “standard” transfers (in Switzerland, the IBAN is not the same for the same account depending on the situation).

PR #36717, opened 2025-12-19
FIX|Fix reassort page colspan and category/tag text color when filtering

Boushabazakaria:
# FIX|Fix reassort page colspan and category/tag text color when filtering
When searching by category or tag on the reassort page, two issues occurred:

- If the search returned no rows, the table `colspan` was incorrect, causing layout issues
- Category and tag labels were forced to white text, making them unreadable depending on the theme background

This fix corrects the colspan calculation when no result rows are displayed
and removes the forced white text color for category/tag labels, allowing
theme CSS to handle colors properly and ensure readability.


Boushabazakaria:
Closing this PR because it was based on develop.
I will reopen a new PR based on 22.0

PR #36725, opened 2025-12-20
FIX - Standardization with other Dolibarr objects

BB2A-Anthony:
# FIX|Fix #[*Standardization with other Dolibarr objects*]
[*Modification of $morehtmlref in only one place*]


BB2A-Anthony:
@eldy I solved ci errors other is not in my files mofied

BB2A-Anthony:
@eldy it's good for you?

PR #36726, opened 2025-12-20
NEW - Look and Feel add title on fiche object

BB2A-Anthony:
# NEW|New [*Look and Feel add title on fiche object*]
[*Long description*]


eldy:
The goal of dol_get_fiche_load is to print the tabs
Printitng title is already provided by function  print load_fiche_titre

As they are 2 rendering function that should act on the same level, we should not have one of them included into the other one. If a page want bothe of them, the page must call the 2.

PR #36731, opened 2025-12-21
refacto initdemo.sh

fpiccinali:
# FIX|Fix initdemo.sh: dialog about documents failed
inidemo.sh failed after loading a new database. 

At almost the end, DIALOG ask if we want to reset document's folder and it is not working. 

This fix does:
* litle refactor to remove duplicated code
* replace DIALOG with a nicer YESNO DIALOG

To test : 
```
cd dev/initdemo
sudo ./initdemo.sh confirm mysqldump_dolibarr_22.0.0.sql dolidemo 3306 dolidemo dolidemo
```
 
----
Another point: if i launch this script twice, the second time i have this error : 
```
ERROR 3730 (HY000) at line 1: Cannot drop table 'llx_accounting_account' referenced by a foreign key constraint 'fk_accounting_category_account_fk_accounting_account' on table 'llx_accounting_category_account'.
```
Do you want me to add and ignore foreign key at the beginning ? Or do we need to empty also llx_accounting_category_account ? Or do we need to simply remove this DROP ?

### DIFF
--- a/dev/initdemo/initdemo.sh
+++ b/dev/initdemo/initdemo.sh
@@ -46,6 +46,8 @@ passwd=$6
 if [ "$confirm" != "confirm" ]
 then
 	echo "----- $0 -----"
+	echo "Usage: initdemo.sh confirm "
+	echo " or"
 	echo "Usage: initdemo.sh confirm [mysqldump_dolibarr_x.x.x.sql database port login pass]"
 	exit
 fi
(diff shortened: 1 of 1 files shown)


fpiccinali:
This is my first PR. I do not know if i need to sync from branch 22.0 or from branch develop for this kind of issue.

PR #36745, opened 2025-12-23
FIX - Select warehouse on propal create

BB2A-Anthony:
# FIX|Fix Select warehouse on propal create
[*Long description*]


eldy:
Can you describe how to reprouce the bug ?

BB2A-Anthony:
@eldy Y Solve conflic

$warehouse_id is -1 because the selectWarehouses function does not accept null or empty

In /htdocs/produc/class/html.fromproduct.class.php

`
L280:	 *  @param  ''|'-1'|'ifone'|'ifonenodefault'|'-2'|string|int|int[]  $selected     Id of preselected warehouse ('' or '-1' for no value, 'ifone' and 'ifonenodefault' = select value if one value otherwise no value, '-2' to use the default value from setup)
`

PR #36746, opened 2025-12-23
FIX - Select warehouse on propal list

BB2A-Anthony:
# FIX|Fix *Select warehouse on propal list*
[*If not empty*]


BB2A-Anthony:
@eldy Y Solve conflic

PR #36748, opened 2025-12-23
FIX : Reception::reOpen() misleading status display /missing error message when rollback 

jyhere:
# Instructions
How to reproduce : 
- Set stock increase on reception close
- Create a reception containing a product with batch enabled. Set a DLC / DLUO.
- Validate and close the reception (stock movements are created)
- Go to the corresponding productlot_card, and change a date (e.g DLUO)
- Reopen the reception => reverting movement stock will fail because of the date inconsistency. No error message, and status is "Validated" on the card... (but not in db because of the rollback)
- Since the status is actually wrong on the card, the user have access to the Set to draft action. Then he can valid and close again, causing movement stocks to be recreated, screwing available quantities.


jyhere:
Apparently i've been drinking a lot before christmas, I don't know what the **** i was committing... I'll open a new PR later.

PR #36751, opened 2025-12-23
Multiprices on product list multicompany compatibility

ibuiv:
getEntity added

# FIX compatibility multicompany on product list with multiprices

### DIFF
--- a/htdocs/product/list.php
+++ b/htdocs/product/list.php
@@ -2043,6 +2043,7 @@ while ($i < $imaxinloop) {
 				$sqlp = "SELECT p.rowid, p.fk_product, p.price, p.price_ttc, p.price_level, p.date_price, p.price_base_type";
 				$sqlp .= " FROM ".MAIN_DB_PREFIX."product_price as p";
 				$sqlp .= " WHERE fk_product = ".((int) $obj->rowid);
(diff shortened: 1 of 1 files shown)


hregis:
@ibuiv sorry but the share "productprice" is not for "multiprice" !!!! 
it's not the same !

hregis:
@ibuiv sorry i close before send mon opinion...
"produductprice" sharing it's only for the same product price for all entities...
multiprices sharing need a reflexion... an entity can use a price and multiprice for one and more entity... and another entity can use a price but another multiprice for  others entities !

hregis:
@ibuiv @eldy 

ça fait 20 ans que que je connais Dolibarr et ça fait 20 ans que Laurent nous dit qu'il faut avoir un fonctionnement souple... alors si nous sommes devenu rigide je souhaiterai être au courant !!! ;-))

hregis:
@ibuiv sorry... this PR is not valid

hregis:
@ibuiv send your request with more screenshots please

ibuiv:
@hregis Thanks for your opinion
I must admit I did not spend a lot of time for thinking further my own configuration (multicompany with multiprices on each with specific sharings)
I will do another PR after several tests and time for think this issue

PR #36764, opened 2025-12-27
NEW: Supplier payment mail/PDF improvements

braito4:
- Keep mail form open after applying an email template (presend).

- Pick a PDF model on presend so the receipt can be generated/attached.

- Add supplier payment substitution tags (amount/date/type/bank/refs).

- Enhance standard supplier payment PDF with supplier refs + IBAN/account info.

# Instructions


eldy:
Can you split PR in several PR to follow the atomic rule of PR.

PR #36775, opened 2025-12-30
FIX height of shipping address box in proposal PDF model

lvessiller-opendsi:
FIX height of shipping adress box in proposal PDF model

**To reproduce**
1. Enable the shipping address in proposal PDF model : 
image

2. Enable hidden const "MAIN_PDF_ADDALSOTARGETDETAILS" : set to 1.

3. Generate a proposal  using cyan PDF model :
image

=> The shipping box is too small

**The fix**
Keep the same size as recipient box by removing the line where the height of the box is reduced.
image


lvessiller-opendsi:
@eldy 
I still get CI errors, but they don't seem to affect the modified files.
I tried to update branch but I got the same errors.

lvessiller-opendsi:
@eldy
If you keep the line :
$hautcadre -= 5;

You still have this problem when you have more details, like the address with 2 lines and the VAT number :
image

PR #36781, opened 2025-12-31
feat(core): allow negative quantities on invoice and order lines

Boushabazakaria:
# NEW|New Allow negative quantities on invoice and order lines

This PR adds core support for negative quantities on invoice and order
document lines.

The feature is disabled by default and can be enabled by setting the
following constants:
- COMMANDE_ENABLE_NEGATIVE_QTY = 1
- FACTURE_ENABLE_NEGATIVE_QTY = 1

Allowing negative quantities enables additional business use cases such as:
- adjustments and corrections directly on documents
- discounts represented as document lines
- financial compensations without using credit notes

### DIFF
--- a/htdocs/commande/card.php
+++ b/htdocs/commande/card.php
@@ -1055,7 +1055,9 @@ if (empty($reshook)) {
 			}
 		}
 
-		if (!$error && ($qty >= 0) && (!empty($line_desc) || (!empty($idprod) && $idprod > 0))) {
+		$commande_qty_requirement = !empty(getDolGlobalString('COMMANDE_ENABLE_NEGATIVE_QTY')) ? ($qty >= 0 || $qty <= 0) : $qty >= 0;
+
+		if (!$error && $commande_qty_requirement && (!empty($line_desc) || (!empty($idprod) && $idprod > 0))) {
 			// Clean parameters
 			$date_start = dol_mktime(GETPOSTINT('date_start' . $predef . 'hour'), GETPOSTINT('date_start' . $predef . 'min'), GETPOSTINT('date_start' . $predef . 'sec'), GETPOSTINT('date_start' . $predef . 'month'), GETPOSTINT('date_start' . $predef . 'day'), GETPOSTINT('date_start' . $predef . 'year'));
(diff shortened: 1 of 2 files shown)


eldy:
Using negative qty without using the native "discount" feature has very bad side effect.Why not using a thirdparty commercial discount to add negative line ?
Is "financial compensations without using credit note" legal ?
In most european country it is not. For which country do you need this ?

creekorful:
> Using negative qty without using the native "discount" feature has very bad side effect.Why not using a thirdparty commercial discount to add negative line ? Is "financial compensations without using credit note" legal ? In most european country it is not. For which country do you need this ?

Hi @eldy, the main use case here is to be able to do amendment, for example when we want to switch a product for a customer. In this case we will add two lines in the propal:

- One line with the previous product and negative qty to simulate the return
- One line with the new product and positive qty

It is done like this to allow to perform a valuation of the product exchange and have a signed document (the propal) to acknowledge it.

eldy:
> > Using negative qty without using the native "discount" feature has very bad side effect.Why not using a thirdparty commercial discount to add negative line ? Is "financial compensations without using credit note" legal ? In most european country it is not. For which country do you need this ?
> 
> Hi @eldy, the main use case here is to be able to do amendment, for example when we want to switch a product for a customer. In this case we will add two lines in the propal:
> 
> - One line with the previous product and negative qty to simulate the return
> - One line with the new product and positive qty
> 
> It is done like this to allow to perform a valuation of the product exchange and have a signed document (the propal) to acknowledge it.

In proposal I think it is already allowed, but in invoice, it should not.

creekorful:
> > > Using negative qty without using the native "discount" feature has very bad side effect.Why not using a thirdparty commercial discount to add negative line ? Is "financial compensations without using credit note" legal ? In most european country it is not. For which country do you need this ?
> > 
> > 
> > Hi @eldy, the main use case here is to be able to do amendment, for example when we want to switch a product for a customer. In this case we will add two lines in the propal:
> > 
> > * One line with the previous product and negative qty to simulate the return
> > * One line with the new product and positive qty
> > 
> > It is done like this to allow to perform a valuation of the product exchange and have a signed document (the propal) to acknowledge it.
> 
> In proposal I think it is already allowed, but in invoice, it should not.

Hi @eldy, would that be ok if we push a PR to include negative qty on order as well?
For the invoice we will remove it from the PR for now and discuss it later on

Cheers

eldy:
The feature to allow negative qty was already present in proposal in past. For proposal, you must add option PROPAL_ENABLE_NEGATIVE_QTY.
I did some tests in v23 and it was broken so i fixed it.
It was also possible for order but only in edit mode. So i fixed this too and added
ORDER_ENABLE_NEGATIVE_QTY to have a similar behavior.
So with v23, you should be able to do it.

creekorful:
Thanks @eldy !

PR #36783, opened 2025-12-31
Fix email subject

SylvainLegrand:
Correction to avoid displaying ">" instead of ">" in email subject lines

### DIFF
--- a/htdocs/core/actions_sendmails.inc.php
+++ b/htdocs/core/actions_sendmails.inc.php
@@ -352,6 +352,8 @@ if (($action == 'send' || $action == 'relance') && !GETPOST('addfile') && !GETPO
 			complete_substitutions_array($substitutionarray, $langs, $object, $parameters);
 
 			$subject = make_substitutions($subject, $substitutionarray);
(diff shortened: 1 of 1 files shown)


rycks:
> Correction to avoid displaying ">" instead of ">" in email subject lines

@SylvainLegrand did you have a real use case of that problem ?

Éric

rycks:
image

doest not seems to be buggy :

image

rycks:
new test with direct write a &lt; in subject ...

image

email in my inbox

image

... so what it the real case problem @SylvainLegrand ? please help us to understand your fix...

emheyarssi:
Maybe linked with #38919.
Ping @SylvainLegrand

SylvainLegrand:
@rycks' screenshots speak for themselves: when the user uses the characters > or < in the subject of their email, they want these characters to appear exactly the same way in the recipient's email client and not transformed into &lt; or other translations that are unreadable for the recipient and stressful because they are associated by the average person with a fraudulent or dangerous email.

PR #36786, opened 2026-01-02
Add new permissions for batchlot management

defrance:
Currently, a person with product rights can access batches and even delete a batch.
The batchlot module must also have its own access rights.


defrance (commented):
ok whitespace

PR #36790, opened 2026-01-02
FIX - CronJob default is entity 1

BB2A-Anthony:
# FIX|Fix CronJob default is entity 1
If the tasks are not on entity 1, certain globo parameters are not available because by default they are on entity 1. OR [#36791](https://github.com/Dolibarr/dolibarr/pull/36791)


hregis:
@BB2A-Anthony for this cronjob this is entity 0
The 0 is for all entities.

hregis:
@BB2A-Anthony can you send an example of the problem?

BB2A-Anthony:
> @BB2A-Anthony can you send an example of the problem?

I understand that in this case, certain parameters, such as SYSTEMTOOLS_MYSQLDUMP or SYSTEMTOOLS_POSTGRESQLDUMP, must also be associated with entity 0; otherwise, they are not accessible for tasks scheduled with entity 0 or other entities, even though these are server parameters.

This is not the case as I propose here. #36791

eldy:
So we must choose between https://github.com/Dolibarr/dolibarr/pull/36791 xor https://github.com/Dolibarr/dolibarr/pull/36790
right ?
I think the #36791 looks better. Do you agree @hregis ?

BB2A-Anthony:
> So we must choose between #36791 xor #36790 right ? I think the #36791 looks better. Do you agree @hregis ?

For me it is the best choice

When creating a planned task the default entity should be 0 and the choice of entity is left if multicompany is activated

eldy:
Closed by #36791

PR #36791, opened 2026-01-02
FIX - Save param SYSTEMTOOLS_MYSQLDUMP and SYSTEMTOOLS_POSTGRESQLDUMP

BB2A-Anthony:
# FIX|Fix Save param SYSTEMTOOLS_MYSQLDUMP and SYSTEMTOOLS_POSTGRESQLDUMP
The parameter is valid for all entities since it is a server configuration parameter. OR #36790

### DIFF
--- a/htdocs/admin/tools/export.php
+++ b/htdocs/admin/tools/export.php
@@ -5,7 +5,7 @@
  * Copyright (C) 2021		Regis Houssin		<regis.houssin@inodbox.com>
  * Copyright (C) 2024		Frédéric France		<frederic.france@free.fr>
  * Copyright (C) 2025		MDW					<mdeweerd@users.noreply.github.com>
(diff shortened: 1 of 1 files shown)


eldy:
So we must choose between https://github.com/Dolibarr/dolibarr/pull/36791 xor https://github.com/Dolibarr/dolibarr/pull/36790
right ?
I think the https://github.com/Dolibarr/dolibarr/pull/36791 looks better. Do you agree @hregis ?

BB2A-Anthony:
> So we must choose between #36791 xor #36790 right ? I think the #36791 looks better. Do you agree @hregis ?

For me it is the best choice

When creating a planned task the default entity should be 0 and the choice of entity is left if multicompany is activated

BB2A-Anthony:
Up @eldy @hregis

PR #36796, opened 2026-01-04
NEW : add skill list on user skill tabs

defrance:
The goal is to know on the user skill tabs the current skill of the user without go on every evaluation element

Capture d’écran 2026-01-04 à 15 52 00


eldy:
This make me confused.
If a user is in company since 3 years and passed 3 evalution, 2 for the profile "Dev PHP" and 1 for the profile "Project leader", what are its "skills" ?
Is it "cumulative skills" found for all evaluation (and if there is several one with a scoere of 2 and another one with a score of 1, what is the score ?
Or is it the the "expected skills" of the last evaluation ?

defrance:
In my view, this allows us to quickly visualize a user's skills as well as their development over time. We have both a list of their skills and, if they have passed multiple ealuations, whether they have progressed.

PR #36804, opened 2026-01-05
FIX: inventory qty is not saved if page > 1

thomas-Ngr:
# FIX: On a long inventory, the quantities were not saved on page > 1

To reproduce :
- create an inventory with > 10 products
- validate inventory
- set pagination to 10
- navigate to page 2
- modify qty for a line, then click "Save"
- navigate to page 1
- navigate to page 2 : your quantity has not been saved !

Observed from branch 21.0 to develop.

This is due to the fact that the SQL request to fetch lines to update (l.291) and the one to fetch the lines to display (l.1045) have the same LIMIT params, but not the same ORDER params.

### DIFF
--- a/htdocs/product/inventory/inventory.php
+++ b/htdocs/product/inventory/inventory.php
@@ -89,6 +89,10 @@ if (!$sortorder) {
 	$sortorder = "ASC";
 }
 
+// Sort by warehouse/product or product/warehouse
+$sortfield .= ',' . ($sortfield == 'e.ref' ? 'p.ref' : 'e.ref');
+$sortorder .= ',' . $sortorder;
+
 // Fetch optionals attributes and labels
 $extrafields->fetch_name_optionals_label($object->table_element);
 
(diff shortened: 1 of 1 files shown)


htdocs/product/inventory/inventory.php
```diff
@@ -291,8 +295,10 @@
 		$sql = 'SELECT id.rowid, id.datec as date_creation, id.tms as date_modification, id.fk_inventory, id.fk_warehouse,';
 		$sql .= ' id.fk_product, id.batch, id.qty_stock, id.qty_view, id.qty_regulated';
 		$sql .= ' FROM '.MAIN_DB_PREFIX.'inventorydet as id';
+		$sql .= ' LEFT JOIN ' . $db->prefix() . 'product as p ON id.fk_product = p.rowid';
+		$sql .= ' LEFT JOIN ' . $db->prefix() . 'entrepot as e ON id.fk_warehouse = e.rowid';
 		$sql .= ' WHERE id.fk_inventory = '.((int) $object->id);
-		$sql .= $db->order('id.rowid', 'ASC');
+		$sql .= $db->order($sortfield, $sortorder);
```
eldy: If sort field is on a ref, when there is 2 lines with same ref, the order is not predicable and may vary.
May be we should add an rowid as a last sort parameter to avoid this ?

PR #36807, opened 2026-01-06
NEW Accountancy - Add transaction model (SQL Part)

aspangaro:
SQL Part

Propose to integrate transaction model for manual entry on accounting.

Example : 

image

image

### DIFF
--- a/htdocs/install/mysql/migration/23.0.0-24.0.0.sql
+++ b/htdocs/install/mysql/migration/23.0.0-24.0.0.sql
@@ -1,0 +1,75 @@
+--
+-- This file is executed by calling /install/index.php page
+-- when current version is higher than the name of this file.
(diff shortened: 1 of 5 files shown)


htdocs/install/mysql/migration/23.0.0-24.0.0.sql
```diff
@@ -0,0 +1,75 @@
...
+
+
+
+
+-- V24 migration
+CREATE TABLE llx_accounting_transaction_template (
+	rowid			integer AUTO_INCREMENT PRIMARY KEY NOT NULL,
+	entity          integer DEFAULT 1 NOT NULL,
+	code			varchar(128) NOT NULL,
+	description		varchar(300),
```
eldy: Do we really need 300 ?
Can 255 be enough ? It may be better if in future we need an index on field.

PR #36808, opened 2026-01-06
FIX: Correct due date calculation for payment term 45J FDM on 15th

Boushabazakaria:
# FIX|Fix Due date calculation for payment term 45 J FDM le 15

When using the payment term **"45 J FDM le 15"**, the due date is incorrectly calculated when the invoice date falls on the **last day of the month**.

image

### Expected behavior
- Invoice date: 2025-03-31  
- Payment term: 45 J FDM le 15 
- Due date: **2025-05-15**

### Actual behavior
- Invoice date: 2025-03-31  
- Payment term: 45 J FDM le 15  
- Due date: **2025-06-15**

An extra month is incorrectly added in this specific case.

### Root cause
The current due date calculation logic does not correctly handle end-of-month invoice dates when applying the **FDM (end of month)** rule combined with a fixed day (15th).  
This results in advancing the due date by one additional month.

### Fix
- Correct the calculation logic to properly handle invoice dates at the end of the month.
- Ensure that the **45 J FDM le 15** payment term always resolves to the correct month.
- Preserve existing behavior for all other payment terms.

### Tests
- Invoice date: 2025-03-30 → Due date: 2025-05-15
- Invoice date: 2025-03-31 → Due date: 2025-05-15

### DIFF
--- a/htdocs/core/class/commoninvoice.class.php
+++ b/htdocs/core/class/commoninvoice.class.php
@@ -1159,7 +1159,7 @@ abstract class CommonInvoice extends CommonObject
 
 			$diff = $date_piece - $date_lim_current;
 
-			if ($diff < 0) {
+			if ($diff <= 0) {
 				$datelim = $date_lim_current;
 			} else {
 				$datelim = $date_lim_next;


htdocs/core/class/commoninvoice.class.php
```diff
@@ -1159,7 +1159,7 @@ public function calculate_date_lim_reglement($cond_reglement = 0)
 
 			$diff = $date_piece - $date_lim_current;
 
-			if ($diff < 0) {
+			if ($diff <= 0) {
```
eldy: Can you try instead to change th eline 1158
$date_lim_next = dol_time_plus_duree((int) $date_lim_current, 1, 'm');
into
$date_lim_next = dol_time_plus_duree((int) $date_lim_current, 1, 'm', 1);

Because when we add 1 month and we are the 31/03 we become 01/05 (so 2 month of offset instead of 1) when not using the param 1 as last param. With param 1 we should get 30/04. Does it work ?

Boushabazakaria: I tried adding the last parameter 1 to dol_time_plus_duree, but the issue still persists.

With:

- invoice date = 31/03/2025
- payment conditions = “45 J FDM le 15”

I get $diff = 0.

In this case, I think we should keep:

`$datelim = $date_lim_current;`


That’s why I changed the condition from `if ($diff < 0)` to `if ($diff <= 0)`, to also handle the case where $diff is zero.

eldy: I don't understand your setup.
On your screennshot, i see the string "Current/next".
But you should get "At end of month" or "A given day in month". What is your language ?
image

I thnk current/next is old value for "a given day in month".
I try to implement a phpunit test to reproduce the error before the fix so we will be protected of a future regression on this.

eldy:
I added more phpunit test into CommonInvoiceTest.php to validate the fix.
d175489fde6bef48a41e1157546aad45a0c6421a

PR #36812, opened 2026-01-06
NEW : allow to save stocks as 0 on a stock movement.

thomas-Ngr:
# NEW hidden const STOCK_MOVEMENT_FORCE_DO_NOT_CLEAN_EMPTY_LINES

When this const is set, if a stock movement puts the stock value to zero, the line will not be erased from DB, but will be saved as "0 in stock".

The use case are : 
- one of our customer has multiple entities, but uses same stocks. They want to know that they have 5 products P on warehouse W on entity 1, and 0 product P on same warehouse W on entity 2
- more generally, one may have "reserved" warehouses for products, and want to see that there is 0 of product P on warehouse W.

(this is a restart for PR https://github.com/Dolibarr/dolibarr/pull/33526)

## screenshots 

There is 0 of stock in warehouse 'A-Concarneau' after a stock movement.

before : No mention of the warehouse

Capture d’écran_2024-02-28_09-47-48 

Capture d’écran_2024-02-28_09-50-55

after : we see stock 0 in warehouse

Capture d’écran_2024-02-28_09-52-13 

Capture d’écran_2024-02-28_09-52-37

### DIFF
--- a/htdocs/product/stock/class/mouvementstock.class.php
+++ b/htdocs/product/stock/class/mouvementstock.class.php
@@ -656,7 +656,7 @@ class MouvementStock extends CommonObject
 				}
 			}
 
-			if (empty($donotcleanemptylines)) {
+			if (empty($donotcleanemptylines) && !getDolGlobalInt('STOCK_MOVEMENT_FORCE_DO_NOT_CLEAN_EMPTY_LINES')) {
 				// If stock is now 0, we can remove entry into llx_product_stock, but only if there is no child lines into llx_product_batch (detail of batch, because we can imagine
 				// having a lot1/qty=X and lot2/qty=-X, so 0 but we must not loose repartition of different lot.
 				$sql = "DELETE FROM ".$this->db->prefix()."product_stock WHERE reel = 0 AND rowid NOT IN (SELECT fk_product_stock FROM ".$this->db->prefix()."product_batch as pb)";
--- a/htdocs/product/stock/product.php
+++ b/htdocs/product/stock/product.php
@@ -1052,8 +1052,10 @@ if (!$variants || getDolGlobalString('VARIANT_ALLOW_STOCK_MOVEMENT_ON_VARIANT_PA
 	$sql .= " FROM ".MAIN_DB_PREFIX."entrepot as e,";
 	$sql .= " ".MAIN_DB_PREFIX."product_stock as ps";
 	$sql .= " LEFT JOIN ".MAIN_DB_PREFIX."product as p ON p.rowid = ps.fk_product";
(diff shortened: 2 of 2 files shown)


eldy:
For the 2 needs you want, it must be solved by request and not by corrupting data in database. If there is 0 in stock a whatever is entity, there is no reason to have a line with 0 in database. No line already means 0.

thomas-Ngr:
>  it must be solved by request

Do you mean that we should have : 
```php
// dolibarr/htdocs/product/stock/product.php l.1051
$sql = "SELECT ... COALESCE (ps.reel,0) ...";
```
In that case, we would have "0" on display for _each_ warehouse where there is no stock, and not on each warehouse *that has contained the product* and has no stock now.

...

This need might be too specific and we may implement it in a specific module. Is it OK to add hooks printFieldListXXX for this SQL request ?

eldy:
> > it must be solved by request
> 
> Do you mean that we should have :
> 
> ```
> // dolibarr/htdocs/product/stock/product.php l.1051
> $sql = "SELECT ... COALESCE (ps.reel,0) ...";
> ```

May be or may not. i did no think on how should be the request and the code of page, i just think on database integrity and if there is no stock, we should not have line. Adding line with 0 to make request easier may works for one case but will also introduce bug as data is no more correct (for exemple a count will fails, a left join that should be empty will return a linge instead of nothing, export will increase number of exported lines when it should not, etc...).

> 
> In that case, we would have "0" on display for _each_ warehouse where there is no stock, and not on each warehouse _that has contained the product_ and has no stock now.
> 
> ...
> 
> This need might be too specific and we may implement it in a specific module. Is it OK to add hooks printFieldListXXX for this SQL request ?

Yes, i think your need seems specifi to your customer and above all on a specifi way to use multicomapny, so solving this inside a hook is surely the best thing. Hooks are always welcome as they does not affect existing code and stability. It is better that allowing (even if option is hidden) a practice that will introduce dangerous data alteration.

BebZ:
Hello, So how is it handled finally ? 
With a hidden const to keep displaying 0 stock when there was an already existing stock ?
or by another way because this hidden const may corrupt DB integrity ?
Thanks

thomas-Ngr:
@BebZ 

Eldy has merged and reverted the very same day. I guess the merge was an error. He was pretty clear that hooks would be favored over a hidden const.

I will introduce hooks at some point, but since [open-dsi ](https://www.opendsi.fr/) maintains its own version of Dolibarr named Osden, and Osden2026 contains the present feature, this is low on our priority list. We restart from DLB branch every two years.

PR #36815, opened 2026-01-06
FIX : remove stock correctly when reception is deleted (like 82e092f)

lemathou:
# FIX|Fix same as 82e092f but in v18

It is already done in https://github.com/Dolibarr/dolibarr/pull/32291 from v21, but I have customers that use older versions, so I dit it from v18.


lvessiller-opendsi (approved):

rycks (approved):
ok for me, just a remark about the pr linked for 21.0 : what about line 1145 with jdate ?

https://github.com/Dolibarr/dolibarr/pull/32291/files

lemathou:
I forgot to check this case, using batches... I do it today and I'll make another PR if needed.

lemathou:
There was an error using batches, I commited the fix in the same branch, is it possible to merge it again using the same PR or do I need to create another PR ?

PR #36821, opened 2026-01-07
Fix : PR #36804 includes a regression that blocks navigation between …

thomas-Ngr:
# Fix: PR #36804 includes a bug that blocks navigation between pages of the inventory

Due to the fact that `$sortfield` was expanded before building `$paramwithsearch` .

### DIFF
--- a/htdocs/product/inventory/inventory.php
+++ b/htdocs/product/inventory/inventory.php
@@ -89,10 +89,6 @@ if (!$sortorder) {
 	$sortorder = "ASC";
 }
 
-// Sort by warehouse/product or product/warehouse, then by batch.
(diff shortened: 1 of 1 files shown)


htdocs/product/inventory/inventory.php
```diff
@@ -126,6 +122,9 @@
 	$paramwithsearch .= '&limit='.((int) $limit);
 }
 
+// Sort by warehouse/product or product/warehouse
+$sortfield .= ',' . ($sortfield == 'e.ref' ? 'p.ref' : 'e.ref');
```
eldy: So we remove criteria batch + rowid ?

thomas-Ngr:
@eldy I fixed your point, can you have another look ? Thanks !

thomas-Ngr:
@eldy I fixed your point, can you have another look ? Thanks !

PR #36835, opened 2026-01-09
FIX extrafields label display when label is a CONCAT

thomas-Ngr:
# FIX extrafields label display when label is a CONCAT

When the label part of a selectlist extrafield is computed, the values are not shown. ex. with `CONCAT(wt.nom,' - ', m.ref) as labelcause`.

This bug has been seen on 22.0 and develop. Might exist before 22.0.

options of the selectlist extrafield :

```
commande m INNER JOIN llx_societe wt ON wt.rowid = m.fk_soc:CONCAT(wt.nom,' - ', m.ref) as labelcause:m.rowid::
```

display : 

**no display on card**

display-ef-1

**no display on list**

display-ef-2

This PR solves this.

### DIFF
--- a/htdocs/core/class/extrafields.class.php
+++ b/htdocs/core/class/extrafields.class.php
@@ -2223,7 +2223,7 @@ class ExtraFields
 
 					// Several field into label (eq table:code|label:rowid)
 					$fields_label = explode('|', $InfoFieldList[1]);
-					$fields_label_names = array_map([$this, 'extractSQLAlias'], $fields_label);	
+					$fields_label_names = array_map([$this, 'extractSQLAlias'], $fields_label);
 
 					if (is_array($fields_label_names) && count($fields_label_names) > 1) {
 						foreach ($fields_label_names as $field_toshow) {
--- a/htdocs/core/class/extrafields.class.php
+++ b/htdocs/core/class/extrafields.class.php
@@ -139,13 +139,13 @@ class ExtraFields
 
 	/**
 	 * Extract SQL aliases
-	 * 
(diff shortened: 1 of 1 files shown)


eldy:
I think this has been removed for security purpose. Can you provide the screenshot of the extrafield setup page you use ?

thomas-Ngr:
It is possible for tickets on branch in develop (commit a83378 of January 5th.) : 

image

Also possible for invoices, with the very same behavior.

eldy:
I confirm that the example you use was not an expected feature but a bypass of an existing vulnerability. The expected behaviour is was the one described into the tooltip help introduced in v23 (rule was correct in v22 but tooltip was not implemented) :

"
Liste des valeurs provenant d'une table
Syntaxe : nom_table:nom_champ:champ_id::filtersql
Exemple : c_typent:libelle:id::filtersql
- nom_champ est le nom du champ dans la base de données, ou peut être plusieurs champs séparés par |
- id_field est nécessairement une clé primaire de type int
- filtersql est une condition. Elle doit utiliser la syntaxe USF. Exemple : (active:=:1) pour afficher uniquement les valeurs actives"

The vulnerability (that allow any sql injection) was fixed in v22.
Concat (that can be used to objuscate any evil sql injection) has been relaced but the use of the | char between each field name. Can you try this ?

thomas-Ngr:
> Can you try this ?

Sure. Thanks for the hint.

I tested it on branch develop, commit 2c5be59 of January 15th.

This syntax works (request on a single table):
```
societe:nom|rowid:rowid::
```

This one does not work (request with a join) : 
```
commande c INNER JOIN llx_societe s ON s.rowid = c.fk_soc:s.nom|c.ref:c.rowid::
```
Peek 15-01-2026 10-39

PR #36860, opened 2026-01-12
FIX supplier order delivery date from order

W1W1-M:
# FIX supplier order delivery date from order
In the case a supplier order is created from a customer order, the delivery date was automatically copied even if a delivery date was specified in supplier order create form. The delivery date was also empty in the supplier order create form regardless of the one copied for customer order.

This PR correctly updates the delivery date select field when creating supplier order from customer order. This gives user feedback that the selected date is the one coming from the customer order and lets them decide to use it or change it.


W1W1-M:

PR #36877, opened 2026-01-14
Fix SMTP OAuth refresh scopes and error details dol 22

braito4:
# Instructions
*Please:*
- ***in particular, in case of a bugfix, please check that you are targetting the branch corresponding to the oldest version in which the bug occurs***

[*Long description*]

[*Long description*]

[*Long description*]

[*Long description*]

[*Long description*]

[*Long description*]


eldy:
What happen if we include the change for the refresh oftoken but not the addition of error management into OAuth lib ?
I ask this because OAUht lib is an externallib and we never change such libe because it is lost at the first lib upgrade. So what will happen when we will upgrade the OAuth lib ? Does the change for the refresh token will still works ?

braito4:
The PR has 2 independent parts:
(1) The functional fix is in CMailFile.class.php: when refreshing, we recreate the OAuth service with the configured scopes (instead of an empty scopes array) and we keep the previous refresh_token if the provider doesn’t return a new one. This is what fixes the Microsoft v2 refresh 400.
(2) The change in StreamClient.php is only for better diagnostics: it allows reading the HTTP response body on 4xx/5xx and including error_description in the exception.
If you merge only (1) and skip the OAuth lib change, refresh will still work; we just lose detailed error messages on failure.
On OAuth lib upgrades, (1) will still be present since it’s in Dolibarr core; only the extra error detail improvement would be lost. If preferred, we can move the error-detail logic outside the lib by injecting a custom HTTP client via ServiceFactory::setHttpClient().

eldy:
If we just lose detailed error messages, it is ok, however as itbis 2 different features/fixes and we may need in a future to make cherry pick to restore one the one lost if we update the voucher lib, can you split the PR into 2 different PR ?

Issue #36880, opened 2026-01-14
Extrafield cascading selects conflict between setListDependencies() and setListDependencies_extra() on line items


playcock:
### Related Issue
[FIX extra-field list depend on parent extra-field list on direct edit](https://github.com/Dolibarr/dolibarr/pull/35803)

### Bug

This was observed when editing line items on Purchase Orders with "Select from table" extrafields that have parent child dependencies, changing the parent field causes child options to become greyed out/disabled and unselectable.

This is related to commit [5de1fcca2f9 ](https://github.com/Dolibarr/dolibarr/commit/5de1fcca2f936c7d53fcd451be91081a5ced545d) where setListDependencies() was moved to lib_head.js.php.

### Dolibarr Version

18.0

### Environment PHP

issue is not PHP version specific - this is a JavaScript bug

### Environment Database

issue is not database specific - this is a JavaScript bug

### Steps to reproduce the behavior and expected behavior

1. Create two "Select from table" extrafields on a line item (e.g., commande_fournisseurdet):
        Parent field: `parent_field` - Select from table
        Create a Purchase Order and add a line item, selecting values for both extrafields
2. Create a Purchase Order and add a line item, selecting values for both extrafields
3. Save the line item
4. Edit the line item and change the parent field to a different value
5. Observe the child field - options become greyed out/disabled

## Expected behavior:
When changing the parent field, the child field should show only the matching options as enabled/selectable.

## Actual behavior:
Child options become greyed out and cannot be selected. The workaround without a code fix is to clear the parent field, save, then re-edit and select the new parent value.

## Additional details
Commit [5de1fcca2f9 ](https://github.com/Dolibarr/dolibarr/commit/5de1fcca2f936c7d53fcd451be91081a5ced545d) moved `setListDependencies()` to `lib_head.js.php`. This function now binds change handlers to ALL select elements with parent-attributed options, including line item extrafields.      

However, line item extrafields also have their own handler `setListDependencies_extra()` generated by `getJSListDependancies()` in `commonobject.class.php`.

When both handlers bind to the same elements:
- `showOptions()` uses `prop("disabled", true/false)`
- `showOptions_extra()` uses `clone()/remove()/append()`

These conflicting approaches cause the options to become permanently disabled.

## JavaScript fix
[...]

PR #36882, opened 2026-01-14
FIX cascading extrafields on line items being greyed out and unselectable

playcock:
Fixes editing line items with parent-child 'Select from table' extrafields, when changing the parent field causes child options to become greyed out.

FIX skip selects inside extrafield_lines_area containers in setListDependencies() as they are handled by setListDependencies_extra().

### DIFF
--- a/htdocs/core/js/lib_head.js.php
+++ b/htdocs/core/js/lib_head.js.php
@@ -1194,6 +1194,10 @@ function showOptions(child_list, parent_list) {
 function setListDependencies() {
 	console.log("setListDependencies");
 	jQuery("select option[parent]").parent().each(function() {
+		// Skip selects inside line item extrafield areas (handled by setListDependencies_extra)
+		if ($(this).closest('[id^="extrafield_lines_area"]').length > 0) {
+			return;
+		}
 		var child_list = $(this).attr("name");
 		var parent = $(this).find("option[parent]:first").attr("parent");
 		var infos = parent.split(":");


playcock:
@eldy Will you apply this to branches 14.0 to 18.0, please?

https://github.com/Dolibarr/dolibarr/issues/36880

eldy:
Is this a bug present on v22 too ?

playcock:
@eldy I apologize for not knowing. We are still running 18.0.8 and have not upgraded past 18.0 yet.

playcock:
> Is this a bug present on v22 too ?

Does the bug need to be present on v22 in order to roll out the fix for v14 - v18?

playcock:
@eldy Could I have a review, please?

eldy:
> @eldy Could I have a review, please?

Is this a bug present on v22 too ?

Answer is important because as youvn pushed the fix in a maintenance branch, if fix was already fixed in v22, the fix in older version mist be done by doing a cherry pick.
And i am surprised that such fix is not already fixed in recent version.

playcock:
> > @eldy Could I have a review, please?
> 
> Is this a bug present on v22 too ?
> 
> Answer is important because as youvn pushed the fix in a maintenance branch, if fix was already fixed in v22, the fix in older version mist be done by doing a cherry pick. And i am surprised that such fix is not already fixed in recent version.

@eldy 

I stood up a docker version of Dolibarr v22 and the issue does not exist. What is protocol for doing the cherry pick to resolve the issue in v18?

eldy:
cherry-picking must be done by finding the commit id that fi in higher version, thein going on older branch and lancuhing git cherry-pick idofcommit to retreive.

PR #36897, opened 2026-01-16
fix french doc

frederic34:


frederic34:
@eldy

frederic34:
@eldy Could you please review this to reduce the number of missing translations?

PR #36900, opened 2026-01-16
FIX: Add missing event counter on Agenda tab

noec764:
image
Some event counters related to different objects were missing.


eldy:
This looks more than a not yet implemented feature than a technical bug.

noec764:
At least i've tried

PR #36901, opened 2026-01-16
set progress value to 0 instead of null

atm-jpb:
# FIX|Fix : set progress value to 0 instead of null
we need to see a 0 % progression  in Declared actual progress like on consumption progres


eldy:
If progession is not defined, we must see empty.
If progression has been defined to 0, we must see "0"
It seems the modified code already apply this rule.
Can you describe the use cas you do to enter a 0 progression and not show it ?

PR #36909, opened 2026-01-17
New : add timespent made by socpeople

defrance:
Currently, only users are authorized to enter time entries for project tasks.

However, it can happen that a collaborator from an external service provider works on a client's project without their work time being recorded.

The purpose of adding fk_socpeople is to add this capability and also to be able to track the time spent by external service providers.


eldy:
All security on dolibarr is managed by "users" and it must remains managed by users (for time spent it is for users assigned to project and tasks).

Allowing contact to directly enter time bypassing the security layer of users will break all security and all security features related to project and task in Dolibarr.

If a thirdparty contact need to enter time, he should be able to do it using its user login (you should be able to create a user login from the contact card).

defrance:
@eldy , there's a misunderstanding in the title of my PR; I've just corrected it.

The purpose of this change isn't to allow third-party contacts to record time, but rather to allow third-party contacts to record time spent on a task (in concrete terms, external contractors working on another client's project).

PR #36934, opened 2026-01-20
# FIX|Fix #*Robust Date Comparison for MO (Manufacturing Order) Planned Dates*

atm-corentin:
# FIX|Fix #[*Robust Date Comparison for MO (Manufacturing Order) Planned Dates*]

Bugfix: Robust Date Comparison for MO (Manufacturing Order) Planned Dates

Problem

The existing validation logic for modifying the planned start and end dates of a Manufacturing Order (MO) fails when one of the opposing dates is not yet defined (stored as an empty string "").

This issue is caused by PHP's loose comparison :

The positive timestamp $newDate is compared against the empty string "".

PHP implicitly converts "" to the integer 0.

The comparison (e.g., timestamp < 0) evaluates to FALSE, prematurely executing the error block (setEventMessages) and preventing the user from successfully setting the initial date.

Solution

The fix standardizes the object's date attributes ($objMo->date_end_planned, etc.) into reliable integer timestamps before performing validation.

We use is_numeric() to detect existing timestamps (positive integers).

We use strtotime() to safely convert empty strings ("" becomes 0) or date strings to integers.

The validation condition is updated to explicitly allow the modification if the opposing date is undefined (value ≤ 0 after conversion).

copy from PR : #36827


atm-corentin:
@eldy can you check pls ?

atm-corentin:
@eldy can you check pls ?

PR #36946, opened 2026-01-21
NEW - VAT Report - Add button to collapse / expand all rates

aspangaro:
Add a button to expand all rates or collapse all (the button disappears in print mode).

image

image


eldy:
The use of a button action (used for worklow) is not compliant with dolibarr UI. Can you replace with a link "Expand all" like we have for example on the list of permission of a user.

See bottom of screenshot
image

aspangaro:
Thanks for your return @eldy

I will try to do as you requested, but I have some concerns about performance. The default collapse mode that was already in place before my pull request was there to solve the problem because if a customer had a huge number of invoices over a month (>800 per day if I remember correctly in my case), we didn't load the data to avoid timeouts. That's why it seems complicated to me to do a “JavaScript” collapse, if I understand correctly.

eldy:
> Thanks for your return @eldy
> 
> I will try to do as you requested, but I have some concerns about performance. The default collapse mode that was already in place before my pull request was there to solve the problem because if a customer had a huge number of invoices over a month (>800 per day if I remember correctly in my case), we didn't load the data to avoid timeouts. That's why it seems complicated to me to do a “JavaScript” collapse, if I understand correctly.
> 

You don't need to do a js collapse. You just need to change the look of button to match look of the sample page. Just add the picto and change css on button should be enough.

eldy:
> Thanks for your return @eldy
> 
> I will try to do as you requested, but I have some concerns about performance. The default collapse mode that was already in place before my pull request was there to solve the problem because if a customer had a huge number of invoices over a month (>800 per day if I remember correctly in my case), we didn't load the data to avoid timeouts. That's why it seems complicated to me to do a “JavaScript” collapse, if I understand correctly.
> 

You don't need to do a js collapse. You just need to change the look of button to match look of the sample page. Just add the picto and change css on button should be enough.

eldy:
> Thanks for your return @eldy
> 
> I will try to do as you requested, but I have some concerns about performance. The default collapse mode that was already in place before my pull request was there to solve the problem because if a customer had a huge number of invoices over a month (>800 per day if I remember correctly in my case), we didn't load the data to avoid timeouts. That's why it seems complicated to me to do a “JavaScript” collapse, if I understand correctly.
> 

You don't need to do a js collapse. You just need to change the look of button to match look of the sample page. Just add the picto and change css on button should be enough.

aspangaro:
Done @eldy 

Thanks for your return.

image

PR #36957, opened 2026-01-22
FIX NEW option to add product parent lines for variants

W1W1-M:
# FIX NEW option to add product parent lines for variants

In some use cases, it is necessary to use a product parent and not variant when adding line to order, invoice, etc. For example : the parent product is a generic/base coffee mug with no artwork, paint, etc, but it also has variants : mugs that have been painted, have stickers, logos, etc.

This PR adds an option `PRODUCT_ALLOW_VARIANT_PARENT_LINE` to allow adding of parent products if the user does select an option in the drop down list.

Capture d’écran 2026-01-22 à 09 51 45


W1W1-M:

PR #36962, opened 2026-01-22
FIX #36960 - User can access some data (product price, invoice amount) without having the respective rights

### LINKED ISSUE #36960: User can access some data (product price, invocie amount) without having the respective rights
### Bug

A user has just the right to enter spent time on project tasks (automatically having the right to read projects). But no other rights. This is relevant, when a company wants employees to record their project hours but does not want to reveal how much was charged to the customer.

On the page to manage times spent for project tasks (projet/tasks/time.php), there are two instances where privileged data is shown without the proper rights:

Issue 1:

Product/service prices - user has no right to read product data. It is possible to assign a product to the time entry (which is fine) but in the select list each product is showing the product price, which in some cases shall not get revealed to the user. This happens in create mode and in edit mode:

Image

Issue 2:

When spent time gets billed, both, the invoice number and the amount is shown to the user, even if he has no rights to read invoice data. In this case, there should be just a yes/no entry, if the position got billed:

Image

### Dolibarr Version

22.0

priojk:
See detailed description in issue #36960

Product price and invoice amount was shown to users without checking right to read.

### DIFF
--- a/htdocs/projet/tasks/time.php
+++ b/htdocs/projet/tasks/time.php
@@ -2499,9 +2499,9 @@ if (($id > 0 || !empty($ref)) || $projectidforalltimes > 0 || $allprojectforuser
 											}
 										}
 									}
-								} else {                                                                                                                                                   
-						              print $langs->trans("Yes");                                                                                                                            
-								}  
+								} else {
+									print $langs->trans("Yes");
+								}
 							}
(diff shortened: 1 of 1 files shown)


priojk:
@eldy failed check on CMailFile.class.php seems unrelated to this PR

Issue #36493, opened 2025-12-01
No stock movement recording for virtual/set products since 22.0.0 (production price is now lost)


lippoliv:
### Bug

#33750 introduced changes regarding stock movements for virtual/set products. Previously, the "parent" had a stock movement as well, now it's "just" the children.

I'm pretty sure there's reasons for that change, but I wonder why users couldn't decide, similar to product variants? 

Right now there's loss of information (see steps below).

### Dolibarr Version

22.0.3

### Environment PHP

8.2

### Environment Database

mariadb:latest

### Steps to reproduce the behavior and expected behavior

1. Configure stock movements on "invoice closed"
2. Create PROD1 (set production costs to $1)
3. Create PROD2 (set production costs to $1)
4. Create PROD3 as a set using 1x PROD1 and 1x PROD2 (set production costs to $3)
5. Create invoice for PROD3
6. Validate the invoice
7. Check stock movments

Findings:
1. In stock movements you'll just see PROD1 and PROD2 decreasing, not PROD3.
2. The movements for PROD1 and PROD2 are not for $1, the price simply is not set.

### Context

I'm affected because of a custom module that I use to document used shipping material and that uses the production cost information. For parent producs that information was set, now since the parent is gone, "shipping costs" always is $0 now...

chbacher:
Hi,

I am encountering exactly the same issue after updating to version 22.0.4.

The Setup: I have a Virtual Product (Kit/Package) that consists of 3 "real" sub-components.

The Problem: Since the update, it is no longer possible to change the stock levels for the parent Package/Virtual Product correctly.

Steps to Reproduce / Observed Behavior:

Current State: I have 28 units of the Package in stock. I received 40 new units.

Scenario A (Standard Increase): I attempt to "Add" 40 units to the stock of the Package.

Result: The 40 sub-components are correctly added to their respective stocks, BUT the stock level of the parent Package remains at 28.

Scenario B (With "Disable stock change for sub-products"): I check the option "Disable stock change for all sub-products of this set during this movement" and attempt the update.

Result: Nothing happens at all. No stock is changed.

Workaround found: The stock movement for the Package only functions if I remove all sub-components from the kit first.

Question: Is there a new hidden variable or a specific setting that needs to be toggled in version 22.x to allow manual stock corrections for Kits without breaking the link to sub-products? Or is this a regression in the latest version?

Any help would be greatly appreciated.

Environment:

Dolibarr Version: 22.0.4

[...]

Issue #36853, opened 2026-01-11
Bug: Product stock is not modified when product has kits


evansgl:
### Bug

This was working perfectly in the past but broke since 22.0
When a product has kits it does not change its quantity when you make an order or invoice.
Also count of units remain same even if you correct stock from Physical Stock tab.

### Dolibarr Version

>= 22.0

### Environment PHP

8.3.26

### Environment Database

mariadb-10.11.15-1.el10.x86_64

### Steps to reproduce the behavior and expected behavior

- Make a product 
- Add some quantity
- Enable kit and add a couple of other products to this product
- Try to  increase or decrease quantity.
- Quantity remains same.

This happens ONLY if product has a kit. If there is no kit. stock increase or decrease works fine

evansgl:
Can be reproduced in latest release 22.0.4

evansgl:
Any update on this please?

josett225:
Could you please share your stock configuration?

evansgl:
<img width="1706" height="1413" alt="Image" src="https://github.com/user-attachments/assets/7683bf9e-c6c3-4fba-a35f-d52f18817629" />

evansgl:
And in products: **(Enable Kits)**

<img width="1699" height="1291" alt="Image" src="https://github.com/user-attachments/assets/791bd49b-9043-40ce-8e66-60077f73d063" />

evansgl:
@josett225

chbacher:
Is identical to the reported bug here (including how to reproduce)
https://github.com/Dolibarr/dolibarr/issues/36493

But no solution or fix given

KimWittkowski1975:
Hi there, 
in the German Forum the same bug was reported. The possible solution: 

~\htdocs\product\stock\class\mouvementstock.class.php ab Zeile 321 following is in the code: 

	// Define if we must make the stock change (If product type is a service or if stock is used also for services)
	// Only record into stock tables will be disabled by this (the rest like writing into lot table or movement of subproucts are done)
	$movestock = 0;
	$productChildrenNb = 0;
	if (getDolGlobalInt('PRODUIT_SOUSPRODUITS')) {
		$productChildrenNb = $product->hasFatherOrChild(1);
	}
	if (($product->type != Product::TYPE_SERVICE || getDolGlobalString('STOCK_SUPPORTS_SERVICES')) && $productChildrenNb == 0) {
		$movestock = 1;
	}

the line: 

    $movestock = 0;

[...]

PR #36970, opened 2026-01-23
FIX: #36493 (#36853) for having parent kit stock changes you need to set PRODUIT_SOUSPRODUITS_ALSO_ENABLE_PARENT_STOCK_MOVE constant.

fappels:


eldy:
In v23, such an option was already introduced but is called PRODUIT_SOUSPRODUITS_ALSO_ENABLE_PARENT_STOCK_MOVE and must be set to allow stock change of parent kit. So for v22, i made the fix manually to reuse the same constant.
But this option is not yet complete. Stock movement on parent kit is not implemented at a lot of place, so feature must remain hidden until correctly supported everywhere.


Note: There is already an option "Stock is managed (or not)" and i think we should use this option (on a per product basis instead of global PRODUIT_SOUSPRODUITS_ALSO_ENABLE_PARENT_STOCK_MOVE) to decide if stock is changed or not so in v23 i also added a test on product_stockable = 1 (the hidden option PRODUIT_SOUSPRODUITS_ALSO_ENABLE_PARENT_STOCK_MOVE should become useless when option product_stockable=1 will be supported on parent kit).


Thanks to check if PRODUIT_SOUSPRODUITS_ALSO_ENABLE_PARENT_STOCK_MOVE is currently enough for your need.

fappels:
Did not see the V23 fix, Ok to backport to V22. But still some user who do future upgrades from lower V22 will have the issue and will post bug report (This is why I used a 'DISABLE' constant). So at least in the user changelog this new constant should be mentioned. 
Think the option "Stock is managed (or not)" is not usable, because even if 'no stock move' mode is set, you need to be able to do stock manipulations to trigger the child stock changes, am I correct @lvessiller-opendsi ?  

I pushed a white space commit to have title in changelog.

eldy:
> Did not see the V23 fix, 

Try to take the very last version of v23 branch. Do a search on PRODUIT_SOUSPRODUITS_ALSO_ENABLE_PARENT_STOCK_MOVE. You should find it at 3 different places.

> Ok to backport to V22. But still some user who do future upgrades from lower V22 will have the issue and will post bug report (This is why I used a 'DISABLE' constant). So at least in the user changelog this new constant should be mentioned. Think the option "Stock is managed (or not)" is not usable, because even if 'no stock move' mode is set, you need to be able to do stock manipulations to trigger the child stock changes, am I correct @lvessiller-opendsi ?
> I pushed a white space commit to have title in changelog.

Currently v22 like v23 is not able to manage correctly the stock of a parent kit. There is a lot of other places where code must be implemenented for this. This is currently a "not supported feature" (not a bug). In v23 things are cleaner, there is a message near the check box "Manage stock" to say "Not supported on parent kit"(this explain to user why it does not work). So there is still some work to do if we want to have it supported fully. The backport of option PRODUIT_SOUSPRODUITS_ALSO_ENABLE_PARENT_STOCK_MOVE in v22 is possible because it is a hidden option so has no effect on current features, but i can't guarantee that it will be enough and i am nearly sure it is enough.

fappels:
> > Did not see the V23 fix,
> 
> Try to take the very last version of v23 branch. Do a search on PRODUIT_SOUSPRODUITS_ALSO_ENABLE_PARENT_STOCK_MOVE. You should find it at 3 different places.

> 
> > Ok to backport to V22. But still some user who do future upgrades from lower V22 will have the issue and will post bug report (This is why I used a 'DISABLE' constant). So at least in the user changelog this new constant should be mentioned. Think the option "Stock is managed (or not)" is not usable, because even if 'no stock move' mode is set, you need to be able to do stock manipulations to trigger the child stock changes, am I correct @lvessiller-opendsi ?
> > I pushed a white space commit to have title in changelog.
> 
> Currently v22 like v23 is not able to manage correctly the stock of a parent kit. There is a lot of other places where code must be implemenented for this. This is currently a "not supported feature" (not a bug). In v23 things are cleaner, there is a message near the check box "Manage stock" to say "Not supported on parent kit"(this explain to user why it does not work). So there is still some work to do if we want to have it supported fully. The backport of option PRODUIT_SOUSPRODUITS_ALSO_ENABLE_PARENT_STOCK_MOVE in v22 is possible because it is a hidden option so has no effect on current features, but i can't guarantee that it will be enough and i am nearly sure it is enough.

Thanks to explain
With 'did not see' I meant, if I had seen before I would not have made a different approach. 

Will see to contribute in V24 on this,

Should I close or will you merge the whitespace to have the title in changelog?

I have put the whitespace close the fix so the issue creators can already see where and how to fix the issue.

PR #36971, opened 2026-01-23
add api for resources (take care that resources is a swagger reserved word

defrance:


eldy:
You should activate option to replace automatically spaces (4) with tab char on your IDE.

defrance (commented):
replace space by tabs

defrance:
seems ok so i close

PR #36992, opened 2026-01-26
Fix public note octopus 35394

vmaury:
Display public note (an incoterms and sales rep signat) only on real first page

Now the public note is displayed in middle of second page


vmaury:
Bad commit with to much changes (2 bugs)

PR #36650, opened 2025-12-14
NEW Minimum amount subscription and subscription's amount formula

Thatoo:
# NEW|New [*Minimum amount subscription and subscription's amount formula*]

[*Allow to set a minimum amount subscription by adherent type and display a comment to explain how the member is supposed to calculate the amount of his/her subscription such as "20€ per family member" for example. I also add some security check about the global minimum amount and this new specific adherent type minimum amount, both server side (php) and client side (jquery) *]

PR #36995, opened 2026-01-26
Minimum amount and subscription's amount formula description

Thatoo:


Thatoo:
I need to test it but it was working on develop before 23.0 got its own branch so there is not reason it should not work...

Thatoo:
I've tested it and it works as expected.
I really hope that could find its way into v 23.0 to complete that work https://github.com/Dolibarr/dolibarr/pull/35033

Thatoo:
I don't understand the phan error, I don't know where the $customprinterallowed is used

eldy:
You must push the pr on develop branch.

Thatoo:
develop isn't dedicated to v 24.0?
There is something that I don't understand. I did this one first https://github.com/Dolibarr/dolibarr/pull/36650 but then develop became v 24 instead of v 23 so i did this new PR to v 23...

Thatoo:
Here it is, https://github.com/Dolibarr/dolibarr/pull/37006 but then it can't be merged as it's behind v 24 as it is based on v23.
Sorry if I'm not clear, I'm quiet confused.

eldy:
Replaced with #37006

PR #36999, opened 2026-01-27
Export Filter on Extrafields of type list

Quentin-Seekness:
Filtering on a List (sellist) extrafields is imcompatible as it is, the 'WHERE' of the buil_filterField() 's method is only used for entity.

We should use the extrafields's showInputField instead.


eldy:
Can you describe more why you need to use showInputField() with an example ?

PR #37003, opened 2026-01-27
FIX Add table alias prefix in selectForFormsList() to prevent SQL ambiguity errors

aspangaro:
Bug see from v18 (maybe under) to develop

Problem detected in one of our modules: 

Capture d’écran du 2026-01-27 16-41-14

**_Problem_**
When using Universal Search syntax (e.g., (entity:=:1) AND (status:=:1)) in field definitions, the selectForFormsList() method generates SQL queries without table alias prefixes for common field names. This causes DB_ERROR_1052 "Column 'entity' in WHERE clause is ambiguous" when the query involves JOINs with tables that share the same column names.

Example of problematic filter:

php
```
'fk_soc' => array(
    'type' => 'integer:Societe:societe/class/societe.class.php:1:entity=__SHARED_ENTITIES__ AND status=1',
    // ...
)
```
Generated SQL (broken):

sql
```
SELECT t.rowid, t.nom, t.name_alias FROM llx_societe as t 
WHERE 1=1 AND t.entity IN (1) AND entity = 1 AND status = 1
```
The fields entity and status lose the t. prefix, causing ambiguity when JOINs are present.

**_Solution_**
This PR adds automatic table alias prefix (t.) to common field names that frequently cause ambiguity in SQL queries. The fix handles both:

Universal Search syntax: (entity:=:1) → (t.entity:=:1)

Standard SQL syntax: entity = 1 → t.entity = 1

Common fields covered:
entity, status, rowid, ref, label, fk_soc, date_creation, tms, nom, name

Generated SQL (fixed):

sql
```
SELECT t.rowid, t.nom, t.name_alias FROM llx_societe as t 
WHERE 1=1 AND t.entity IN (1) AND t.entity = 1 AND t.status = 1
```

**_Changes_**

Modified htdocs/core/class/html.form.class.php

Function: selectForFormsList()

Added regex-based field name prefixing before calling forgeSQLFromUniversalSearchCriteria()

### DIFF
--- a/htdocs/core/class/html.form.class.php
+++ b/htdocs/core/class/html.form.class.php
@@ -15,7 +15,7 @@
  * Copyright (C) 2012-2016  Marcos García           <marcosgdf@gmail.com>
  * Copyright (C) 2012       Cedric Salvador         <csalvador@gpcsolutions.fr>
  * Copyright (C) 2012-2015  Raphaël Doursenaud      <rdoursenaud@gpcsolutions.fr>
- * Copyright (C) 2014-2020  Alexandre Spangaro      <aspangaro@open-dsi.fr>
+ * Copyright (C) 2014-2026  Alexandre Spangaro      <alexandre@inovea-conseil.com>
  * Copyright (C) 2018-2022  Ferran Marcet           <fmarcet@2byte.es>
  * Copyright (C) 2018-2021  Frédéric France         <frederic.france@netlogic.fr>
  * Copyright (C) 2018       Nicolas ZABOURI	        <info@inovea-conseil.com>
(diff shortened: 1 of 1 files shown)


rycks:
@aspangaro could you please make a copy & paste of the error because your screenshot is hard to read ...

lvessiller-opendsi:
@aspangaro 
Maybe, could you add alias filter directly in your field ? 

'fk_soc' => array(
    'type' => 'integer:Societe:societe/class/societe.class.php:1:t.entity=__SHARED_ENTITIES__ AND t.status=1',
    // ...
)

For me you have not to use USF syntax in Dolibarr v18. You should have two versions of filters. One  using USF filter and one without for Dolibarr v18.

aspangaro:
Hello,

@rycks Sorry, it's a picture coming from my customer.

Column 'status' in where clause is ambigious

SELECT t.rowid, t.nom, t.name_alias 
FROM llx_societe AS t
LEFT JOIN llx_societe_extrafields AS e ON t.rowid = e.fk_object 
WHERE 1=1 
  AND t.entity IN (1) 
  AND ((entity = 1) AND (status = '1')) 
  ORDER BY t.nom ASC, t.name_alias ASC

@lvessiller-opendsi 
I understand the solution but how do you know when to use a ‘t’?

rycks:
@aspangaro we have a dream, heu no, an idea could be enough ... some of time there is a very hard job for us to analyze the problem and set up a dolibarr with same situation

> https://wiki.dolibarr.org/index.php/LTS_18_ACTIVITY_2026

so could you please make a new instance on your saas with your module where we could see / test the bug without your fix ? then we could try to apply your patch and check consequences / tests ?

that's an "new" idea to try some new solutions to help everybody (you and us) when PR seems to be stucked in a blocky situation

what did you think about that ? good or bad idea ?

PR #37006, opened 2026-01-27
NEW Minimum amount and subscription's amount formula description per membership type

Thatoo:

It's a new attempt of
https://github.com/Dolibarr/dolibarr/pull/36995
and
https://github.com/Dolibarr/dolibarr/pull/36650


Thatoo:
I don't know what I should do now? Update the branch? Fix the CI even though it isn't related to this PR?

eldy:
According to CONTRIBUTING file, before submitting change on code, you must first submit PR with only the change in the database structure. So you must split your PR in 2.

Thatoo:
Here is the PR with only the change in the database structure : https://github.com/Dolibarr/dolibarr/pull/37021

Thatoo:
Awesome. Thank you @eldy for checking and merging my work. Could it find its way into v 23.0 to complete that work https://github.com/Dolibarr/dolibarr/pull/35033 ?

eldy:
> Awesome. Thank you @eldy for checking and merging my work. Could it find its way into v 23.0 to complete that work https://github.com/Dolibarr/dolibarr/pull/35033 ?

I m afraid no. V23 was already frozen in november. No way to add new features in v23 now.

PR #37015, opened 2026-01-28
fix(emailcollector): persist and dedupe attachments

braito4:
## Summary
- Reset per-message IMAP globals to prevent attachment/body leakage between emails.
- Persist supplier-order/ticket attachments via a helper that dedupes by content hash and returns metadata.
- Enforce EMAILCOLLECTOR_ALLOWED_ATTACHMENT_EXTENSIONS across attachment paths and log skipped items.
- Remove extra chmod calls when saving attachments.

### DIFF
--- a/htdocs/emailcollector/class/emailcollector.class.php
+++ b/htdocs/emailcollector/class/emailcollector.class.php
@@ -2555,13 +2555,16 @@ class EmailCollector extends CommonObject
 								// Verify if ticket already exists to fall back on the right operation
 								$tickettocreate = new Ticket($this->db);
 							$errorfetchticket = 0;
-							$alreadycreated = 0;
-							if ($ticketid > 0) {
-								$alreadycreated = $tickettocreate->fetch($ticketid);
-							}
-							if ($alreadycreated == 0 && !empty($trackid)) {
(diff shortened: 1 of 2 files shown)


htdocs/emailcollector/class/emailcollector.class.php
```diff
@@ -4064,15 +4193,12 @@ private function saveAttachment($destdir, $filename, $content)
 		$quality = $tmparraysize['quality'];
 
 		file_put_contents($destdir.'/'.$filename, $content);
-		dolChmod($destdir.'/'.$filename);
```
eldy: Why removing this ?

htdocs/emailcollector/lib/emailcollector.lib.php
```diff
@@ -190,8 +190,6 @@ function saveAttachment($path, $filename, $data)
 		$i++;
 	}
 	file_put_contents($filepath, $data);
-	dolChmod($filepath);
```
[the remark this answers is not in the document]
braito4: restored

htdocs/emailcollector/class/emailcollector.class.php
braito4: restored

PR #37021, opened 2026-01-28
[database] Minimum amount and subscription's amount formula description

Thatoo:


Thatoo:
@eldy I don't find these two lines we have merged
```
ALTER TABLE llx_adherent_type ADD COLUMN minimumamount double(24,8) DEFAULT NULL AFTER caneditamount;
ALTER TABLE llx_adherent_type ADD COLUMN amountformuladescription text AFTER minimumamount;
```
neither in 
htdocs/install/mysql/migration/22.0.0-23.0.0.sql
nor in
htdocs/install/mysql/migration/23.0.0-24.0.0.sql
which will cause an issue when updating from v23 to v24 as these two new column are needed for this PR that has been merged https://github.com/Dolibarr/dolibarr/pull/37006
What should we do? a new PR adding only
```
ALTER TABLE llx_adherent_type ADD COLUMN minimumamount double(24,8) DEFAULT NULL AFTER caneditamount;
ALTER TABLE llx_adherent_type ADD COLUMN amountformuladescription text AFTER minimumamount;
```
to  htdocs/install/mysql/migration/23.0.0-24.0.0.sql ?

eldy:
The pr to create files were published into develop (good).
However, the file modified was v22 to v23 that must no be modified. May be a revert was done to fix this. You can resubmit a pr for the database change on sql file v23 to v24...

Thatoo:
here it is : https://github.com/Dolibarr/dolibarr/pull/37824

PR #37022, opened 2026-01-28
Fix: Pass complete object description to selectForForms in showInputF…

Boushabazakaria:
# FIX|Fix selectForForms receives incomplete object description in showInputField

## Problem
In `CommonObject::showInputField()` , only the class name (`$param_list_array[0]`) is passed to `selectForForms()` instead of the full object description (`$param_list[0]`).

**Change:**
```php
// Before (bug)
$out = $form->selectForForms($param_list_array[0], ...);

// After (fix)
$out = $form->selectForForms($param_list[0], ...);
```

## Impact
- Filters in link field definitions are lost
- AJAX search doesn't work properly for link fields
- Class path information is missing

## Testing
Tested with custom modules using link fields with filters and AJAX search enabled.


eldy:
Can you provide an example on how you reproduce the bug ?  What is you fields definition ?

Boushabazakaria:
## Debug Evidence

Screenshot 2026-01-29 at 13 45 15

Screenshot 2026-01-29 at 13 49 47


I added debug lines in `CommonObject::showInputField()`:
```php
var_dump($param_list_array[0], $param_list[0]);
```

**Result (see attached screenshot):**

**Current code (bug):**
- `$param_list_array[0]` = `"Device"` ← Only class name
- `$param_list[0]` = `"Device:intervold/class/device.class.php:1:(status:=:1) AND (entity:IN:__SHARED_ENTITIES__)"` ← Complete description

When `selectForForms()` receives only `"Device"`, it cannot:
- Load the correct class (missing path `intervold/class/device.class.php`)
- Apply filters `(status:=:1) AND (entity:IN:__SHARED_ENTITIES__)`
- Enable AJAX search properly

**Fix:**
Simply use `$param_list[0]` instead of `$param_list_array[0]` to pass the complete object description.

## Field Definition Used
```php
'fk_device' => array(
    'type' => 'integer:Device:intervold/class/device.class.php:1:(status:=:1) AND (entity:IN:__SHARED_ENTITIES__)',
    'label' => 'Device',
    'enabled' => 1,
    'visible' => 1,
),
```

> Can you provide an example on how you reproduce the bug ? What is you fields definition ?

eldy:
> ## Debug Evidence
> Screenshot 2026-01-29 at 13 45 15 Screenshot 2026-01-29 at 13 49 47
> I added debug lines in `CommonObject::showInputField()`:
> 
> ```
> var_dump($param_list_array[0], $param_list[0]);
> ```

Having just "Device" here is the old behaviour that should not be used anymore because replaced by using objectfield when we want to use a link with filter.

What do you have for
 
var_dump($objectfield, $keyprefix, $key, $keysuffix)

Boushabazakaria:
## Debug Output

As requested, here is the output of `var_dump($objectfield, $keyprefix, $key, $keysuffix)`:
```
array(7) {
  ["type"]=> string(98) "integer:Device:intervold/class/device.class.php:1:(status:=:1) AND (entity:IN:__SHARED_ENTITIES__)"
  ["label"]=> string(15) "InterVOLDDevice"
  ["enabled"]=> string(1) "1"
  ["position"]=> int(34)
  ["notnull"]=> int(0)
  ["visible"]=> int(1)
  ["index"]=> int(1)
}
$keyprefix = ""
$key = "fk_device"
$keysuffix = ""
```

## Analysis

`$objectfield` is correctly an **array** containing the full field definition with type, filters, etc.

You're right that using `$objectfield` array is the preferred method. However, there's an implementation issue.

### The Problem

In `showInputField()` :
```php
$param_list_array = explode(':', $param_list[0], 4);
$out = $form->selectForForms($param_list_array[0], ...); // Passes only "Device"
```

Even though `$objectfield` is passed correctly, the **first parameter** `$objectdesc` receives only `"Device"` instead of the complete description.

### Why This Matters

In `selectForForms()`:

1. ** `$objectdescorig = $objectdesc;` (saved BEFORE processing `$objectfield`)
2. ** `$urloption = '...' . urlencode($objectdescorig) . '...'`

The Ajax URL gets `objectdesc=Device` instead of `objectdesc=Device:intervold/class/device.class.php:1:(status:=:1)...`, causing filters to be lost in Ajax calls.

## The Fix

In `showInputField()`:
```php
// Before (bug)
$out = $form->selectForForms($param_list_array[0], $keyprefix.$key.$keysuffix, (int) $value, $showempty, '', '', $morecss, $moreparam, 0, (empty($val['disabled']) ? 0 : 1), '', $objectfield);

// After (fixed)
$out = $form->selectForForms($param_list[0], $keyprefix.$key.$keysuffix, (int) $value, $showempty, '', '', $morecss, $moreparam, 0, (empty($val['disabled']) ? 0 : 1), '', $objectfield);
```

**Result:** 
- AJAX search works correctly
- Filters are applied properly
- Class path is preserved

This ensures `$objectdescorig` contains the complete description for Ajax calls, working in harmony with the `$objectfield` array.

creekorful:
Hello @eldy, any news on this? Does this fix looks right to you or should we seek for something else?

PR #37024, opened 2026-01-29
MCP server sql part

sonikf:
Hi @eldy
I' ve seen your wish for a first example of a dirty implementation of an MCP server and since no one else volunteering I would like to offer my playground proof of concept to get this started.

If you approve, can I still target 23 or should i close this and target develop?

This is needed both for audit trail and debugging purposes.


eldy:
V23 has already entered the freeze period in november, so no change is allowed into structure for new feature. 
Can you push it for develop (v24) ?

sonikf:
Yes!

PR #37045, opened 2026-01-31
Add thumbnails for avif

frederic34:


htdocs/core/lib/images.lib.php
```diff
@@ -240,13 +240,13 @@ function dol_imageResizeOrCrop($file, $mode, $newWidth, $newHeight, $src_x = 0,
 	// Test function to read source image exists
 	$imgfonction = '';
 	switch ($infoImg[2]) {
-		case 1:	// IMG_GIF
+		case IMAGETYPE_GIF:	// IMG_GIF
```
eldy: I think we were using the num instead of the constant because the constant may be undefined if exit module (or gd i don't remeber) is not loaded.

PR #37046, opened 2026-01-31
FIX divisa rate indirect to direct because Dolibarr use indirect for default

leninrivas:
# FIX|Fix #[*divisa rate indirect to direct  because Dolibarr use indirect for default*]
[*
Direct	: 	1 Divisa Currency = X Currency Main.
Indirect: 	1 Currency Main = X Divisa Currency.
Then for Dolibarr use is Indirect for default
*]


htdocs/install/mysql/tables/llx_multicurrency_rate.sql
```diff
@@ -22,7 +22,7 @@ CREATE TABLE llx_multicurrency_rate
 	rowid 				integer AUTO_INCREMENT PRIMARY KEY,
 	date_sync 			datetime DEFAULT NULL,
 	rate 				double NOT NULL DEFAULT 0,
-	rate_indirect 		double DEFAULT 0,
+	rate_direct 		double DEFAULT 0,
```
eldy: If you change the structure for first intall, you must also add a ligne into the ligration 23 to 24 script .sql

PR #37055, opened 2026-02-02
ENH rename llx_societe field euid to idprof7

thomas-Ngr:
# QUAL Rename llx_societe field euid to idprof7

PR #36997 added a new field `euid` for European companies.
This PR suggests a mors generic approach by naming this field idprof7.

NB - idprof6 might also be used, but I don't know if european companies already have to use idprof6.

@aspangaro

### DIFF
--- a/htdocs/install/mysql/migration/23.0.0-24.0.0.sql
+++ b/htdocs/install/mysql/migration/23.0.0-24.0.0.sql
@@ -107,6 +107,7 @@ ALTER TABLE llx_facture_rec ADD COLUMN fk_email_template integer DEFAULT NULL;
 ALTER TABLE llx_holiday_users ADD COLUMN import_key varchar(14);
 
 ALTER TABLE llx_societe ADD COLUMN euid varchar (64);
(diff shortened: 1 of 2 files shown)


thomas-Ngr:
@aspangaro has a good point. idprofX may vary across european countries (idprof6 in FR, idprof3 in BE), so it should not be used for an information that should not vary accross countries.
He made this change like the field `tva_intra`, also specific to EU.

eldy:
> @aspangaro has a good point. idprofX may vary across european countries (idprof6 in FR, idprof3 in BE), so it should not be used for an information that should not vary accross countries. He made this change like the field `tva_intra`, also specific to EU.

I confirm that idprofX are data that depends on countries and variation is managed by the language file.
But data that are common to world (or at least europe) can keep their real name and should not use the dynamic generic feature (For example tva_num is in this case).
So i think euid should remain the name to use as it is common to enough countries and must not use a dynamic definition.

thomas-Ngr:
OK, I have no more argument so euid will stay as it is. Thanks !

PR #37062, opened 2026-02-02
FIX #36843 Update card-rec.php

### LINKED ISSUE #36843: Duplicate product description when adding a product on a template invoice
### Bug

Duplicate product/service description when adding a product/service on a template invoice.

### Dolibarr Version

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

- Go to a template invoice and add a product/service that currently has a description. That description is autofilled and you can edit it before clicking the "ADD" button.
- Click the "ADD" button
- The line with the product/service is added but the description is duplicated

This does not happen with standard invoices only with template invoices.

Not sure when this bug was introduced, I think before v17, can't test that back with my setup.

### Attached files

_No response_

AWeerWolf:
#Fix #36843 Item desc appreas twice on Sales Invoice template

The item desc is already added into the line when selecting the Item. No Need to concat it again when clicking the "add" button.

### DIFF
--- a/htdocs/compta/facture/card-rec.php
+++ b/htdocs/compta/facture/card-rec.php
@@ -688,7 +688,7 @@ if (empty($reshook)) {
 					$desc = $prod->description;
 				}
 
-				$desc = dol_concatdesc($desc, $product_desc);
+				//Fix #36843 $desc = dol_concatdesc($desc, $product_desc);
 
 				// Add custom code and origin country into description
 				if (!getDolGlobalString('MAIN_PRODUCT_DISABLE_CUSTOMCOUNTRYCODE') && (!empty($prod->customcode) || !empty($prod->country_code))) {


eldy:
On setup of module Product, what value do you have for "How to use the description of the products when adding a product as a line of a document" ?

AWeerWolf:
**Without the patch I made.**
For the value of  "How to use the description of the products when adding a product as a line of a document" in differents cases : 

_Sale Invoice :_
- "Do not autofill ... Produc desc will be concatenated.." : Product desc is only added when clicking "ADD" button => 👍
- "Auto fill ... with the description of the product" : Product desc is added when selecting the product and NOT duplicated when clicking the "ADD" button => 👍
- "Do not autofill ... Do not auto concat ..." : Poduct desc is never added => 👍

_Sales Invoice Template_
- "Do not autofill ... Produc desc will be concatenated.." : Product desc is only added when clicking "ADD" button => 👍
- "Auto fill ... with the description of the product" : Product desc is added when selecting the product and ALSO when clicking the "ADD" button => **Duplicate description**
- "Do not autofill ... Do not auto concat ..." : Poduct desc is not added at selection time BUT it is added when clicking "ADD" button => **Product description appears but it should not**

PR #37072, opened 2026-02-03
Fix: Add permission check before downloading documents

Boushabazakaria:
# FIX|Fix Missing permission check on document download

## Problem
Document download endpoints were not properly checking user permissions before allowing file access. Users without the "read document" permission could still download files by directly accessing the download URL, even though the UI correctly displayed "Access denied" on hover.

This security issue affected all modules using the generic `document.php` endpoint (orders, invoices, proposals, contracts, projects, etc.).

## Root Cause
The permission check was only implemented on the frontend (tooltip display) but was missing on the backend download endpoint in `htdocs/document.php`.

## Solution
Added proper permission verification in `document.php` before serving any document. The script now checks if the user has the appropriate read permission for the requested module before allowing download.

If permission is missing, the user receives a 403 Forbidden response with the translated message "NotEnoughPermissions".

## Changes
- Modified `htdocs/document.php` to add permission checks for all module parts
- Uses existing translation key `NotEnoughPermissions`

## Testing
Tested with a user account without document read permissions:
- Download link now returns 403 Forbidden
- With permissions granted, download works correctly

## Security Impact
This fix closes a security vulnerability where unauthorized users could download documents they shouldn't have access to.

### DIFF
--- a/htdocs/document.php
+++ b/htdocs/document.php
@@ -145,7 +145,7 @@ if (empty($modulepart) && empty($hashp)) {
 if (empty($original_file) && empty($hashp)) {
 	httponly_accessforbidden('Bad link. Missing identification to find file (original_file or hashp)', 400);
 }
-if (!$user->hasRight('ecm', 'read')) {
+if (isModEnabled('ecm') && !$user->hasRight('ecm', 'read')) {
 	httponly_accessforbidden($langs->trans("NotEnoughPermissions"), 403, 1);
 }
 if ($modulepart == 'fckeditor') {
--- a/htdocs/document.php
+++ b/htdocs/document.php
@@ -145,6 +145,9 @@ if (empty($modulepart) && empty($hashp)) {
 if (empty($original_file) && empty($hashp)) {
 	httponly_accessforbidden('Bad link. Missing identification to find file (original_file or hashp)', 400);
 }
+if (!$user->hasRight('ecm', 'read')) {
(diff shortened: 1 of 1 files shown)


htdocs/document.php
```diff
@@ -145,6 +145,9 @@ function llxFooter($comment = '', $zone = 'private', $disabledoutputofmessages =
 if (empty($original_file) && empty($hashp)) {
 	httponly_accessforbidden('Bad link. Missing identification to find file (original_file or hashp)', 400);
 }
+if (!$user->hasRight('ecm', 'read')) {
```
eldy: The wrapper documents is used for a lot of other use case than ecm.
It is also used for most object when ecm module does not exists. 

And protection is implemented by a check on the hash (for public ccess of files using the "sharing key" hash) or by this line :

$check_access = dol_check_secure_access_document($modulepart, $original_file, (int) $entity, $user, '');

If user try to download an invoice without having the permission to read invoice, the line will block download.
The permission ecm->read is used only for the "manual ECM" part.

So i think there is no error.
Note that files in "medias" dir are public so not protected.
If you think there is, thanks to provide an example of full url with param you can use that bypass permissions.

eldy: I confirm your report. There is inconsistency in UI.
I pushed a fix. [4da516b](https://github.com/Dolibarr/dolibarr/pull/37072/commits/4da516bf557ca3b96d221143bc682fa23dd1901f)

Issue #36944, opened 2026-01-21
No lot available on father kits for delivery after the migration from v20 to v21


syrinx75:
### Bug

At the during delivery page, for the products in kits, only the father of the kits, their lot numbers are not available. Dolibarr says 'no product to deliver has been found'.

This problem appears after the migration from v20 to v21. I also tested from v19 to v24.

During the migration to v21, the following errors are visible:

`Database Migration de la base (données)
Migrate Product Batch files path

Warning: Undefined property: stdClass::$id in C:\xampp\htdocs\doli\install\upgrade2.php on line 4492

Warning: Undefined property: stdClass::$id in C:\xampp\htdocs\doli\install\upgrade2.php on line 4492

Warning: Undefined property: stdClass::$id in C:\xampp\htdocs\doli\install\upgrade2.php on line 4492

Warning: Undefined property: stdClass::$id in C:\xampp\htdocs\doli\install\upgrade2.php on line 4492

Warning: Undefined property: stdClass::$id in C:\xampp\htdocs\doli\install\upgrade2.php on line 4492`

It's about this line:
`$lot->ref = $obj->id;		// No ref for the moment`

<img width="690" height="454" alt="Image" src="https://github.com/user-attachments/assets/3c5820be-1b91-4f65-b07f-f064d5fbb439" />

### Dolibarr Version

21.0.4

### Environment PHP

8.2.12 also earlier

### Environment Database

XAMPP and DoliWamp

### Steps to reproduce the behavior and expected behavior

To migrate from earlier version (v19, v20) to v21.
To make an order with kit products, the father kits products.
To start the delivery, to choose the warehouse, to select the lots available.

### Attached files

https://www.dolibarr.fr/forum/t/dolibarr-22-0-4-expedition-de-kits-avec-lots-affichee-sans-stock/50590/2

chbacher:
In my humble opinion that is a huge breaking issue and goes along with [https://github.com/Dolibarr/dolibarr/issues/36493](url)

syrinx75:
I performed another migration from v19 to v23 with the same result as in v21.

I compared the data in the databases linked to "product", "lot", and "batch", and I didn't see any differences within them. So the databases migration seems to be working correctly from one version to the next.

I think the code (since v21) is faulty, the shipping page for selecting batches doesn't recognise the batches of the "parents" products in the kits.

However, it's surprising that I'm the only one affected and that this has been going on for three versions of Dolibarr.

PR #37089, opened 2026-02-04
FIX: #36944 father kits with lot

fappels:
New hidden constant PRODUIT_SOUSPRODUITS_ALSO_ENABLE_PARENT_STOCK_MOVE need to be checked on more places with a father is a lot type product


fappels:
@eldy Still see some issue on this. Working on it. If you plan to make a 22.0.5, please wait.

PR #37094, opened 2026-02-04
FIX - knowledge no detail lines

BB2A-Anthony:
# FIX|Fix #[*knowledge no detail lines*]
[*knowledge has no detail lines so no restriction is needed on this to validate it*]


eldy:

PR #37117, opened 2026-02-05
set oldCopy and Status

atm-vincent-p:
# FIX|Fix Set oldcopy and status in the Ticket close and markAsRead functions


atm-vincent-p:
I created a module using the Module Builder.
I need to intercept all ticket status changes.

The issue is that for some status changes (read, assigned, canceled and closed), a trigger is called (TICKET_MODIFY, TICKET_ASSIGNED, TICKET_CLOSE).
However, for the other status changes, the status is updated via the setStatut function without any associated trigger key.
Even if we modify the core to add a trigkey, it doesn't change the fact that in the setStatut function, the object's oldcopy and status are not set.

What would be the best approach to follow the core standards ?

atm-vincent-p:
@eldy Can you check please ?

PR #37125, opened 2026-02-06
Improve user card performance by avoid mysql OR

dbeniamine:
# PERF user card loading can be slow

With Mysql backend, the `OR` statement forbids using indexes (as explained [here](https://stackoverflow.com/a/40074524), this is a rewrite of a slow query replacing a `OR` by an `UNION` to improve performance.

For a customer, loading a user card was approximately 40 seconds, is now instantaneous


htdocs/comm/action/class/actioncomm.class.php
```diff
@@ -1441,10 +1348,12 @@ public function getActions($socid = 0, $fk_element = 0, $elementtype = '', $filt
 				$sql .= " (SELECT r.rowid FROM ".MAIN_DB_PREFIX."actioncomm_resources as r WHERE";
 				$sql .= " r.element_type = 'socpeople' AND r.fk_element = ".((int) $fk_element).' AND r.fk_actioncomm = a.id)';
 			} elseif ($elementtype == 'user') {
-				$sql .= " AND (a.fk_user_action = ".((int) $fk_element)." OR EXISTS";
-				$sql .= " (SELECT r.rowid FROM ".MAIN_DB_PREFIX."actioncomm_resources as r WHERE";
+				$sql1 = $sql;
+				$sql .= " AND a.fk_user_action = ".((int) $fk_element);
+				$sql = "SELECT a.id FROM ((" . $sql . ") UNION (" . $sql1;
+				$sql .= " AND EXISTS (SELECT r.rowid FROM ".MAIN_DB_PREFIX."actioncomm_resources as r WHERE";
```
eldy: In the secodn request the test
`AND a.fk_user_action = ".((int) $fk_element)`
must be excluded with
`AND a.fk_user_action <> ".((int) $fk_element)`
If not a record that is both ok with fk_user_action and link to actioncomm_resources will returned twice.

PR #37130, opened 2026-02-06
Update api_thirdparties.class.php

SylvainLegrand:
$this->company->fetch($id); return an integer

### DIFF
--- a/htdocs/societe/class/api_thirdparties.class.php
+++ b/htdocs/societe/class/api_thirdparties.class.php
@@ -1060,12 +1060,14 @@ class Thirdparties extends DolibarrApi
 		}
 
 		$result = $this->company->fetch($id);
-		if (!is_array($result)) {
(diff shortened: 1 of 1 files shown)


htdocs/societe/class/api_thirdparties.class.php
```diff
@@ -1060,12 +1060,14 @@ public function getSalesRepresentatives($id, $mode = 0)
...
 
 		$result = $this->company->fetch($id);
-		if (!is_array($result)) {
+		if (!$result) {
 			throw new RestException(404, 'Thirdparty not found');
 		}
 
 		$result = $this->company->getSalesRepresentatives(DolibarrApiAccess::$user, $mode);
-
+		if (!is_array($result)) {
```
eldy: Why removing the test to detect if a technical error occured (in this case, return is a negetive int instead of an array) ?

PR #37154, opened 2026-02-08
Develop add group parent manage code

rycks:
and now here is the code for groups rights inheritance

then i could make a group "salary" with some rights, and an other group "commercials" with more rights and a third one "director" with more more rights ...


hregis:
@rycks tested with transverse mode ? (All users are managed in Master entité)

PR #37155, opened 2026-02-08
datamodel for user change password next time

rycks:
The target of this new field is to be able to add a "force change password on next login" then admin could set a password for a user or send new password via dolibarr process (email, so unsecure by default). Then we should "force" user to create a new password by himself.

Or after a new big update, or a sort of dataleak or ... 

Solution with massaction will come after that PR if you are agree with such change

### DIFF
--- a/htdocs/install/mysql/migration/23.0.0-24.0.0.sql
+++ b/htdocs/install/mysql/migration/23.0.0-24.0.0.sql
@@ -137,4 +137,7 @@ ALTER TABLE llx_usergroup ADD COLUMN fk_parent integer DEFAULT NULL AFTER entity
 ALTER TABLE llx_usergroup ADD INDEX idx_usergroup_fk_parent (fk_parent);
 ALTER TABLE llx_usergroup ADD CONSTRAINT fk_usergroup_parent FOREIGN KEY (fk_parent) REFERENCES llx_usergroup (rowid);
 
+-- Force change password next time
+ALTER TABLE llx_user ADD COLUMN force_pass_change TINYINT DEFAULT 0 AFTER pass_temp;
+
 -- end of migration
--- a/htdocs/install/mysql/tables/llx_user.sql
+++ b/htdocs/install/mysql/tables/llx_user.sql
@@ -41,6 +41,7 @@ create table llx_user
   pass                varchar(128),
   pass_crypted        varchar(128),
   pass_temp           varchar(128),			                  -- temporary password when asked for forget password or 'hashtoallowreset:YYYMMDDHHMMSS' (where date is max date of validity)
+  force_pass_change   tinyint       DEFAULT 0,                -- 1 if user must change password at next login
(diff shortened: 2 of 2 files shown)


hansemschnokeloch:
Why not use `llx_user.datelastpassvalidation` and add a configurable delay after which each user must change their password again?

rycks:
@hansemschnokeloch maybe because there is no informations about that field and the target of that field :)

in french i could translate with "date last pass validation" with "date de dernière validation" and it does not means the same as "end of life of password" or "user must change his password on date" but if that's the target of datelastpassvalidation i will be happy to use it and add some comments on @var field definition

```
	/**
	 * @var int|string
	 */
	public $datelastpassvalidation;
```

but really i think that's not the target for that property, please explain to me regarding 

```
			$sql = "UPDATE ".$this->db->prefix()."user";
			$sql .= " SET pass_crypted = '".$this->db->escape($password_crypted)."',";
			$sql .= " datelastpassvalidation = '".$this->db->idate(dol_now())."',";
			$sql .= " pass_temp = null,";
			$sql .= " force_pass_change = 0";
```

and in box_lastlogin

```
		if ($user->datelastpassvalidation) {
			$tmp = dol_print_date($user->datelastpassvalidation, "dayhour", 'tzuserrel');
		} else {
			$tmp = '<span class="opacitymedium">'.$langs->trans("Unknown").'</span>';
		}
```

rycks:
sorry my prev message was sent to speed, the second part of your message was important 

> "add a configurable delay after which each user must change their password again"

because that delay should be global ? and all users will be concerned ? in my point of view there is a mass action on users list then you could check all users but admin (for example) and put the flag "change password next time"

hansemschnokeloch:
`datelastpassvalidation` is defined here with a comment (I honnestly don't know what a disconnect all means)

https://github.com/Dolibarr/dolibarr/blob/a2822764e50240b8acbaca5f8918f5945eaa863c/htdocs/install/mysql/tables/llx_user.sql#L89

For the delay it doesn't have to be mandatory, but you can imagine it global to force regularly password updates, or define it for example to -1 to force password change only at first login.

hansemschnokeloch:
`datelastpassvalidation` is definitely used for the last password change in this commit

https://github.com/Dolibarr/dolibarr/commit/8e23f51959b992b9867bff970f7671494e8f123c

eldy:
> datelastpassvalidation

datelastpassvalidation is used to store the last date we change password of a user. It canbe used to manage a feature "disconnect all session".
This date is used this way:
All open sessions that are open before the value of datelastpassvalidation are automatically set as unvalid. So when a user change a password, this date is set, and all already open sessions will be automatically disconnected because the date of start of session will appear before the date of last change. This is to be sure that when you change a pass, no already open session remain valid. 
This looks a different need.

hansemschnokeloch:
Thx for the explanation for datelastpassvalidation

Issue #37185, opened 2026-02-10
Ticket - Adding the workflow to the tickets - Assigning the ticket changes the ticket to read


lmag:
### Feature Request

When a staff member is assigned to a ticket, it implies the content has been reviewed. This ensures accountability by signaling that the ticket is no longer unread or ignored.

PR #37187, opened 2026-02-10
#37185 [Workflow] add: ticket workflow read ticket when assign someone

evarisk-kilyan:


htdocs/admin/workflow.php
```diff
@@ -207,6 +207,13 @@
 		'picto' => 'shipment'
 	),
 
+	'WORKFLOW_TICKET_CLASSIFY_READ_ASSIGN' => array (
```
eldy: The module workflow is dedicated to make a change into the life cycle of an object in a module when we do something in another.  Think "workflow" module as "cross-module workflow".
Here action remains in ticket module, so it should be an option into the setup of module ticket.

PR #37196, opened 2026-02-11
NEW Can request and force user to change its password

rycks:
Update user card: add checkbox to handle force update user password

And handle force change password on login via a redirect to a dedicated page (to become on next PR) or is there a better solution for that point ?


hansemschnokeloch:
Hi,
did you test this in case of an openid connexion ?
Your changes main.inc.php can force the user to change his Dolibarr password even if he is connected via openid.
There will be some confusion for the user in this case.

hregis:
@rycks and ldap -> dolibarr

rycks:
@hansemschnokeloch as i don't have any dolibarr with such setup could you please give me some more details about that ?

are you in such situation :

`
$dolibarr_authentication_mode="openid_connect,dolibarr"
`

rycks:
@hregis and @hansemschnokeloch i've just added a test against auth context to apply that only if auth mode is dolibarr core

hregis:
@rycks for dol_authmode you can add : 

ldap : for dolibarr2ldap (not ldap2dolibarr) (internal module)

twofactor : for twofactor authentication (external module)
mc : for multicompany (external module)

or maybe add a method for an external module to add a value for "dol_authmode" ?

rycks:
@hregis for the moment i will only do it for core dolibarr normal auth mode, when it will work we could make other PR for orther auth mode : step by step and as little PR as i can to be easy to read, understand and merge

hansemschnokeloch:
@rycks there is also a constant `NOLOGIN`, I'm not sure if we can use it in this context to determine how the user was identified.

PR #37200, opened 2026-02-11
Qual: Fix ambigious redirect error on Phan workflow

mdeweerd:
# Qual: Fix ambigious redirect error on Phan workflow

Rewrote the shell command that is supposed to suppress a file contents but is flagged by the environment.


mdeweerd:
The issue was a typo in the definition of the variable/macro.

PR #37199, opened 2026-02-11
Qual: Update spelling

mdeweerd:
# Qual: Update spelling

Several fixes following codespell 2.4.2.dev run.

PR #37207, opened 2026-02-12
Qual: Update PHPStan workflow to run on all files in integration

mdeweerd:
# Qual: Update PHPStan workflow to run on all files in integration

The PHPStan workflow has been updated to run on all files in integration branches.


mdeweerd:
In #37199 a phpstan notice surfaced which did not occur in the develop branch in a file I changed only for comment spellings.

I found that the phpstan action only runs on changed files i nthe integration branch, so this update ensures that phpstan runs on all files in the integration branch but only on changed files in a pull request.

PR #37208, opened 2026-02-12
Reactivating the delivery time and reputation for supplier prices

splohmer:
# FIX Reactivating the delivery time and reputation for supplier prices
In previous versions of dolibarr the delivery time and reputation was shown in the list view. Currently it can be entered if creating or editing but the values are not shown anymore in the list. This fixes it.


splohmer:
Hi @eldy,

I saw that you made this change in https://github.com/Dolibarr/dolibarr/commit/148c285614f50bc922d3e5fd8e7cfba1dc432cac
Was there a reason to deactivate this?

eldy:
> Hi @eldy,
> 
> I saw that you made this change in [148c285](https://github.com/Dolibarr/dolibarr/commit/148c285614f50bc922d3e5fd8e7cfba1dc432cac) Was there a reason to deactivate this?

They were not disabled.
As you can see in current v22, we have them.
image
They are just not set as always visible by default. Only the most important field must be visible by default in list (we can't show all fields into table). It is the job of user to decide if he want or not some fields and which one. As we must select a limited list by default, only mandatory fields are selected by default, optional one are also optional in this list (-1 in checked parameter means just optional in list, user must choose to show them).

splohmer:
Hi @eldy,

thanks for clarifying this. I completely oversaw this - sorry. This is a really nice feature.

PR #37210, opened 2026-02-12
Fix api product update

vmaury:
Fix : If product price is set to price by customer, product sell price is never updated by api

### DIFF
--- a/htdocs/product/class/api_products.class.php
+++ b/htdocs/product/class/api_products.class.php
@@ -377,8 +377,8 @@ class Products extends DolibarrApi
 
 		$result = $this->product->update($id, DolibarrApiAccess::$user, 1, 'update', $updatetype);
 
-		// If price mode is 1 price per product
-		if ($result > 0 && !empty($conf->global->PRODUCT_PRICE_UNIQ)) {
+		// If price mode is 1 price per product or price by customer
+		if ($result > 0 && (!empty($conf->global->PRODUCT_PRICE_UNIQ) || !empty($conf->global->PRODUIT_CUSTOMER_PRICES))) {
 			// We update price only if it was changed
 			$pricemodified = false;
 			if ($this->product->price_base_type != $oldproduct->price_base_type) {


rycks (approved):
ok for me ... but after intensive review by @lvessiller-opendsi ... that's not so simple ...

lvessiller-opendsi:
@vmaury 
Be careful, for me, you will set product sell price but not product sell price by customer.
Could you give us a full example ?

vmaury:
Yes I set only set sell price, but it's better than nothing ...
And even when you're in "sell price by customer" mode, if no specific sell price is set for a customer, the "global" sell price is used, and is now not updated during a PUT.

As I saw in the code, there's no possibility to set a customer sell price
There is a getCustomerPricesPerCustomer method, but no setCustomerPricesPerCustomer

rycks:
@vmaury may we try a new idea with you ? because some of time there is a very hard job for us to analyse the problem and set up a dolibarr with same situation

> https://wiki.dolibarr.org/index.php/LTS_18_ACTIVITY_2026

we set up a dolibarr 18.0.8 LTS instance on one saas provider, so could you please reply the bug on that instance where we could replay the bug and the solution ?

i will send you the admin password via direct mail

what did you think about that idea ?

rycks:
It seems to be 4 prices modes for dolibarr:

- PRODUCT_PRICE_UNIQ : 
- PRODUIT_MULTIPRICES
- PRODUIT_CUSTOMER_PRICES
- and PRODUIT_CUSTOMER_PRICES_BY_QTY

see admin/product.php

```
			if ($princingrules == 'PRODUCT_PRICE_UNIQ') { // For this case, we disable entries manually
				$res = dolibarr_set_const($db, 'PRODUIT_MULTIPRICES', 0, 'chaine', 0, '', $conf->entity);
				$res = dolibarr_set_const($db, 'PRODUIT_CUSTOMER_PRICES_BY_QTY', 0, 'chaine', 0, '', $conf->entity);
				$res = dolibarr_set_const($db, 'PRODUIT_CUSTOMER_PRICES', 0, 'chaine', 0, '', $conf->entity);
```

with that PR there is only 2 cases covered ... that's the reason why we are in trouble ...

maybe we could invert the idea : cover specific cases before and at the end put default price

no ?

lvessiller-opendsi:
@vmaury 
Sorry but it's not forecast in the API endpoint of PUT /product/{id}
image

I suggest to add another API product endpoint [PUT /products/{id}/selling_multiprices/per_customer] and implement "setCustomerPricesPerCustomer" or "updateCustomerPricesPerCustomer".

lvessiller-opendsi (commented):
It's seems to be a NEW for develop

vmaury:
> @vmaury Sorry but it's not forecast in the API endpoint of PUT /product/{id} 
> 
> I suggest to add another API product endpoint [PUT /products/{id}/selling_multiprices/per_customer] and implement "setCustomerPricesPerCustomer" or "updateCustomerPricesPerCustomer".

For me it's not usefull, and may be to complex.

I suggest to take another way : not speaking about API, when a product public price is manually updated, customer prices are unchanged, which in most cases is not convenient as it's obliged to change them one by none by hand, or mostly with an dirty sql update (I do this at least every year for one of my client which is in PRODUIT_CUSTOMER_PRICES config ...)
I suggest to add an option "UPDATE_CUSTOMER_PRICES_RATIO_PUBLIC_PRICE"
If this option is set, if a public price is for example increased by 10% all depending prices (PRODUIT_MULTIPRICES or
PRODUIT_CUSTOMER_PRICES or PRODUIT_CUSTOMER_PRICES_BY_QTY) will be augmented with the same %
In real life, it's very logical

and our problem with the API would be solved to :-)

What do you think about this idea ?

rycks:
> What do you think about this idea ?

nice i like that idea, this is near than "Changement TVA en masse" we have in dolibarr but @lvessiller-opendsi says "that is a massaction on products" and this one already exists ... but it seems to work only on single prices (our tests seems to reveal a but in case of multiprices)

lvessiller-opendsi:
To post on develop branch

